PR #4323, opened 2017-12-19
doc: Remove mentions to deprecated options (both are removed at 1.3)

tk0miya:


stephenfin (approved):
To be specific, these were removed in commit 8c2fabe117227ea3f1c233d1196a3c87dc947b1c. You should include this in your commit message.

tk0miya:
Thank you for comment. Updated.

PR #4325, opened 2017-12-19
autodoc: Use logging module to emit warnings

tk0miya:
As a second step of refactoring autodoc, I propose that replacing `self.directive.warn()` to logging module.
As a side effect, the warning messages are not appeared on generated document when users specifies `keep_warnings = True` (see below).


stephenfin:
> As a side effect, the warning messages are not appeared on generated document when users specifies keep_warnings = True (see below).

The image and this statement make conflicting statements. Are you saying that warnings *are* shown when the option is enabled, or that they *are not* shown despite the option being enabled. What was the previous behaviour and what is the behaviour now?

sphinx/ext/autodoc/__init__.py
```diff
@@ -1542,15 +1540,14 @@ class AutoDirective(Directive):
 
     def warn(self, msg):
         # type: (unicode) -> None
-        self.warnings.append(self.reporter.warning(msg, line=self.lineno))
+        logger.warning(msg, line=self.lineno)
```
stephenfin: I'm assuming this where is the issue lies. Why can't we call warnings and still return this?

stephenfin (changes requested):
I'm not entirely sure why we're doing this when we actually lose functionality. Surely we can figure out a way to emit warnings to output and _still_ use logging for `stdout`?

tk0miya:
>The image and this statement make conflicting statements. Are you saying that warnings are shown when the option is enabled, or that they are not shown despite the option being enabled. What was the previous behaviour and what is the behaviour now?

Sorry for my poor description. I meant latter one.

* Current behavior (default): warnings are emitted to stderr only
* Current behavior (with `keep_warnings = True`): warnings are emitted to stderr and shown on the document
* New behavior: warnings are emitted to stderr only (even if `keep_warnings = True`)

The screenshot describes the second case.

>I'm not entirely sure why we're doing this when we actually lose functionality. Surely we can figure out a way to emit warnings to output and still use logging for stdout?

My final goal is simplify the implementation whole of autodoc. It is very complicated and strong-coupled. so it is very hard to understand and modify. To maintain the extension, I'd like to refactor it because there are no active maintainer for it. (In other word, we have to make it maintainable)

So this PR's goal is making the modules less-coupled. To do that, it uses `sphinx.util.logging` as a logger. I know this change make the functionality lost. But I think maintainability is more important for us.

In addition, I believe this feature is not important. because it is hard to notice the warnings on checking output on each build. For the purpose, the warnings on stderr is better. We also have `-W` (warn-is-error) option to notice them.

stephenfin:
Ok, thanks for the info. It seems a shame to lose functionality but I understand that the maintainability of this module is really important. I could volunteer to maintain it at some point in the future, but first I need to work on the other items I have.

I'm going to look through this one more time tomorrow to see what I think and then leave my final comments. It looks pretty good so far though 😄

shimizukawa (approved):
LGTM. 

I guess most users are unaware of this feature and it will not be necessary.

tk0miya:
@stephenfin It seems you're requesting changes. Is this intended?

stephenfin:
@tk0miya I wanted to review it once again given the additional information you provided. However, I have not had time and @shimizukawa has already reviewed it. I'll remove my review.

stephenfin (approved):
I can't see anything obvious and we can re-add the missing feature in the future if anyone misses it.

tk0miya:
Thanks.
I feel `keep_warnings` option is no longer needed. But, as you said, we can re-add it if anyone wants it.

I'll merge this after Travis says ok. (in this week we gets many timeouts from travis...)

stephenfin:
> I'll merge this after Travis says ok. (in this week we gets many timeouts from travis...)

Blame Spectre/Meltdown 😞

tk0miya:
Oh, I understand. That's too bad.

tk0miya:
Merged! Thank you for reviewing :-)

PR #4328, opened 2017-12-19
Fix #947: Adding autodoc option to ignore a module's __all__

### LINKED ISSUE #947: automodule: add option to ignore __all__
I'm dealing with a large set of Python modules and would like a way to see all the code in the generated HTML documentation.  This would be a conveniently way to search for duplicated code and such.

However, currently the automodule extension honors the `__all__` defined in modules, which is nice.  But it would also be nice to have an option that would ignore the `__all__` so that all the functions and classes in a module are pulled out and documentation generated.

I propose adding a 'hidden-members' that would effectively ignore the `__all__` variable, generating documentation for all vars, methods & classes, including those not specified in the `__all__`.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/947
- Originally reported by: Nick Hilton
- Originally created at: 2012-06-12T21:20:55.983

KevKeating:
### Feature or Bugfix
- Feature

### Purpose

This PR adds an ignore-dunder-all flag to autodoc that causes a module's `__all__`  to be ignored.  Without this flag, no documentation is generated for any class or variable that doesn't appear in `__all__`.  This new flag can be given in autodoc_default_flags in conf.py.

### Relates
- #947: automodule: add option to ignore __all__

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -50,6 +50,7 @@ Features added
 * #3160: html: Use ``<kdb>`` to represent ``:kbd:`` role
 * #4212: autosummary: catch all exceptions when importing modules
 * #3991, #4080: Add :confval:`math_numfig` for equation numbering by section
+* #947: autodoc now supports ignore-module-all to ignore a module's ``__all__``
 
 
 Features removed
--- a/doc/ext/autodoc.rst
+++ b/doc/ext/autodoc.rst
@@ -103,8 +103,10 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
      will document all non-private member functions and properties (that is,
      those whose name doesn't start with ``_``).
 
-     For modules, ``__all__`` will be respected when looking for members; the
-     order of the members will also be the order in ``__all__``.
+     For modules, ``__all__`` will be respected when looking for members unless
+     you give the ``ignore-module-all`` flag option.  Without
+     ``ignore-module-all``, the order of the members will also be the order in
+     ``__all__``.
 
      You can also give an explicit list of members; only these will then be
      documented::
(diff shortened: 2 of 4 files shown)


CHANGES
```diff
@@ -49,6 +49,7 @@ Features added
 * HTML themes can set up default sidebars through ``theme.conf``
 * #3160: html: Use ``<kdb>`` to represent ``:kbd:`` role
 * #4212: autosummary: catch all exceptions when importing modules
+* #947: autodoc now supports ignore-dunder-all to ignore a module's ``__all__``
```
stephenfin: What's the rationale for the name? `dunder` has no meaning for me, at least

KevKeating: It's a vaguely common-ish shortening (for certain values of "common") of "double underscore" to refer to \_\_variables\_\_ and \_\_methods\_\_.  (See [here](https://dbader.org/blog/python-dunder-methods), [here](https://stackoverflow.com/questions/33030434/dunder-methods-verus-builtin-methods), or [here](https://pythonconquerstheuniverse.wordpress.com/2012/03/09/pythons-magic-methods/) for a few random Google hits.)  I'd be happy to switch to a better name for the flag.  #947 suggests "hidden-members."  I don't really think of classes not in \_\_all\_\_ as "hidden," but I could switch to that if it's preferred.  Maybe "ignore-module-all" or "ignore-all-list"?

tk0miya: I know the keyword, but I don't know it is common sense or not. So +1 for `ignore-module-all`.

KevKeating: I've now switched from `ignore-dunder-all` to `ignore-module-all`.  Let me know if I should merge in changes in master to fix the conflict in CHANGES and/or squash this branch down to a single commit.

stephenfin: Thanks for that. I'd personally just squash and rebase, if that's ok by you.

KevKeating: Sure, that's no problem at all.  Just re-pushed.

tk0miya:

PR #4329, opened 2017-12-19
Add "Arcade" library documentation to examples

pvcraven:
Subject: Add "Arcade" library documentation to examples

### Feature or Bugfix
- Feature

### Purpose
- Give another example to people looking to use Sphinx for their documentation.

### DIFF
--- a/EXAMPLES
+++ b/EXAMPLES
@@ -111,7 +111,7 @@ Documentation using the sphinxdoc theme
 Documentation using another builtin theme
 -----------------------------------------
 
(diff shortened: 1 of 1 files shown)


tk0miya:
my browser says the site is insecure. I think it is good to use non-HTTPS url instead. What do you think?

pvcraven:
Sorry, one line and I didn't get it right. http it is.

tk0miya:
Thanks!

PR #4311, opened 2017-12-16
Let LaTeX obey :confval:`numfig_secnum_depth` for figures, tables, ...

jfbu:
... and code-blocks.

This is new feature and a bit of bugfix.

I I noticed in passing that prior to this PR, there is some non-coherent behaviour of code-block counter compared to figure and table, particularly when docclass is ``'howto'``, and I need to document the change. Documentation not yet added.

The aim is that figure, table, and code-block numbering will be the same in LaTeX as in HTML when ``numfig`` is activated. Thus it is breaking change. Possibly I could add an option to keep LaTeX ignore the ``numfig_secnum_depth`` setting.

It is independent of PR #4166, but motivated by it. Next step will be to get equation numbering obey the new html feature from #4166.

PR #4166, opened 2017-10-18
Number equations across pages and by section (fix #3991, #4080)

jahn:
### Feature or Bugfix
- Feature

### Purpose
Make equation numbers in html output unique across pages.

### Detail
This uses numfig for numbering and is only in effect if `numfig` and the new confval `math_numfig` are set.  Numbering is by section if `numfig_secnum_depth` is set.  The functionality is tested for each of the math backends.
This fix is admittedly rather hacky, but doing it properly seems to involve much deeper changes (see https://github.com/sphinx-doc/sphinx/pull/3676#issuecomment-298872853), and I feel a fair number of people would appreciate having this addressed (in particular preventing duplicate equation numbers on different pages), so I think it is still worth including.

### Relates
- #3991
- #4080 

Many thanks to @edoddridge for valuable input.

PR #4332, opened 2017-12-21
Let LaTeX obey :confval:`math_numfig` for equation numbering

jfbu:
Implements #4166 for LaTeX PDF output. This is continuation of #4311.

Note that #4311 added no latex config value to say not to obey config setting `numfig` it it is True. Thus, here I don't either add some extra config value which would allow to obey `math_numfig` for html like outputs but not for LaTeX. If activated for html, it affects LaTeX too.

### DIFF
--- a/doc/ext/math.rst
+++ b/doc/ext/math.rst
@@ -46,17 +46,12 @@ or use Python raw strings (``r"raw"``).
 
 .. confval:: math_numfig
 
-   If ``True``, displayed math equations are numbered across pages in html and
-   related (epub, ...) output (also with latex) when :confval:`numfig` is enabled.
-   :confval:`numfig_secnum_depth` is respected.  The :rst:role:`eq` role must
-   be used to reference equation numbers, not the :rst:role:`numref` role.
-   Default is ``True``.
+   If ``True``, displayed math equations are numbered across pages when
+   :confval:`numfig` is enabled.  The :confval:`numfig_secnum_depth` setting
+   is respected.  The :rst:role:`eq`, not :rst:role:`numref`, role
+   must be used to reference equation numbers.  Default is ``True``.
 
    .. versionadded:: 1.7
-      If ``False``, latex behaves as in former releases i.e. for ``'manual'``
-      docclass (and ``'howto'`` for Japanese) it resets by default the
-      equation numbers per each toplevel section, and for ``'howto'`` it uses
-      continuous numbering across entire PDF.
 
 :mod:`.mathbase` defines these new markup elements:
 
--- a/doc/ext/math.rst
+++ b/doc/ext/math.rst
@@ -47,12 +47,16 @@ or use Python raw strings (``r"raw"``).
 .. confval:: math_numfig
 
    If ``True``, displayed math equations are numbered across pages in html and
-   related (epub, ...) output when :confval:`numfig` is enabled.
+   related (epub, ...) output (also with latex) when :confval:`numfig` is enabled.
    :confval:`numfig_secnum_depth` is respected.  The :rst:role:`eq` role must
    be used to reference equation numbers, not the :rst:role:`numref` role.
    Default is ``True``.
 
    .. versionadded:: 1.7
+      If ``False``, latex behaves as in former releases i.e. for ``'manual'``
+      docclass (and ``'howto'`` for Japanese) it resets by default the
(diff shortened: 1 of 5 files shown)


doc/ext/math.rst
```diff
@@ -47,12 +47,16 @@ or use Python raw strings (``r"raw"``).
...
+   related (epub, ...) output (also with latex) when :confval:`numfig` is enabled.
    :confval:`numfig_secnum_depth` is respected.  The :rst:role:`eq` role must
    be used to reference equation numbers, not the :rst:role:`numref` role.
    Default is ``True``.
 
    .. versionadded:: 1.7
+      If ``False``, latex behaves as in former releases i.e. for ``'manual'``
+      docclass (and ``'howto'`` for Japanese) it resets by default the
+      equation numbers per each toplevel section, and for ``'howto'`` it uses
+      continuous numbering across entire PDF.
```
jfbu: Perhaps these explanations add noise to doc and I should not have added them?

sphinx/writers/latex.py
```diff
@@ -601,6 +601,11 @@ def __init__(self, document, builder):
                     (',numfigreset=%s' % self.numfig_secnum_depth)
             else:
                 self.elements['sphinxpkgoptions'] += ',nonumfigreset'
+            try:
+                if builder.config.math_numfig:
+                    self.elements['sphinxpkgoptions'] += ',mathnumfig'
+            except:
+                pass
```
jfbu: @tk0miya can you please check that. There is probably better way to access `math_numfig`, but I ended up using the try clause as a workaround.

jfbu: I have corrected the ``except``. But I would prefer more direct way to access this config value from the mathbase extension, if possible.

sphinx/writers/latex.py
```diff
@@ -601,6 +601,11 @@ def __init__(self, document, builder):
                     (',numfigreset=%s' % self.numfig_secnum_depth)
             else:
                 self.elements['sphinxpkgoptions'] += ',nonumfigreset'
+            try:
+                if builder.config.math_numfig:
+                    self.elements['sphinxpkgoptions'] += ',mathnumfig'
+            except AttributeError:
+                pass
```
jfbu, line 608: I have corrected the bare except but I would prefer more direct way to access the mathbase extension ``math_numfig`` config value

doc/ext/math.rst
```diff
@@ -47,12 +47,16 @@ or use Python raw strings (``r"raw"``).
 .. confval:: math_numfig
 
    If ``True``, displayed math equations are numbered across pages in html and
-   related (epub, ...) output when :confval:`numfig` is enabled.
+   related (epub, ...) output (also with latex) when :confval:`numfig` is enabled.
```
tk0miya: I think these mentions about format is not needed by this change. because other formats don't support math itself.

doc/ext/math.rst
tk0miya: It's okay to me. Anyway, this text should move to outside of the directive.

sphinx/writers/latex.py
tk0miya: Ideally, this code should move to the extension. This is sphinx-core.
And we should provide a way to enhance LaTeX output.

But, as I proposed in #4333, math feature should be merged to sphinx-core in future.
So I'm okay this as is.

tk0miya (approved):

doc/ext/math.rst
```diff
@@ -46,11 +46,10 @@ or use Python raw strings (``r"raw"``).
...
 
-   If ``True``, displayed math equations are numbered across pages in html and
-   related (epub, ...) output when :confval:`numfig` is enabled.
-   :confval:`numfig_secnum_depth` is respected.  The :rst:role:`eq` role must
-   be used to reference equation numbers, not the :rst:role:`numref` role.
-   Default is ``True``.
+   If ``True``, displayed math equations are numbered across pages when
+   :confval:`numfig` is enabled.  The :confval:`numfig_secnum_depth` setting
+   is respected.  The :rst:role:`eq`, not :rst:role:`numref`, role
+   must be used to reference equation numbers.  Default is ``True``.
```
jfbu, line 52: I removed references to format and also did not include explanations about what latex does if the config setting is not used

jfbu (commented):
thanks @tk0miya for review, will merge after testing completes

PR #4335, opened 2017-12-21
Closes #4334: sphinx-apidoc: References to non-existing files in TOC

### LINKED ISSUE #4334: sphinx-apidoc generates references to non-existing files in TOC for folders containing only an empty __init__.py
### Problem
When using sphinx-apidoc to generate RST files, there are references to non-existing files generated in the TOC of a package when encountering a directory with only an empty __init__.py.

```
$ sphinx-apidoc ./sphinx-toc-test/ -o ./source
```

#### Error logs / results
```
Creating file ./source/mypackage.rst.
Creating file ./source/mypackage.something.rst.
Creating file ./source/modules.rst.
```

#### Expected results
In ./source/mypackage.rst:
```
mypackage package
=================

Subpackages
-----------

.. toctree::

    mypackage.resource
    mypackage.something
...
```
*mypackage.resource* is reference, but the file *mypackage.resource.rst* does not exists.
The same inconsistencies exists with the --implicit-namespaces flag.

The solution should be to only reference files that get created. I've been working on a patch that I want to submit, including test cases.

A zip file with an example file structure is provided.
[sphinx-toc-test.zip](https://github.com/sphinx-doc/sphinx/files/1580096/sphinx-toc-test.zip)

### Environment info
- OS: OSX 10.10.5
- Python version: 3.6
- Sphinx version: 1.6.5

zufallsgenerator:
Subject:  Don't generate references to non-existing files in in TOC of python packages or modules

### Feature or Bugfix
- Bugfix

### Purpose
- Don't generate references to non-existing files in TOC of package modules when encountering a directory only containing an empty init.py. This is decided in the method shall_skip in apidoc.py, so changing the behaviour there to include directories with empty init.py would still make this fix valid.

### Detail
- Bug fix for #4334 

### Relates
- #4334

### DIFF
--- a/sphinx/apidoc.py
+++ b/sphinx/apidoc.py
@@ -116,7 +116,12 @@ def create_package_file(root, master_package, subroot, py_files, opts, subs, is_
         text += '\n'
 
     # build a list of directories that are szvpackages (contain an INITPY file)
-    subs = [sub for sub in subs if path.isfile(path.join(root, sub, INITPY))]
+    # and also checks the INITPY file is not empty, or there are other python
+    # source files in that folder.
+    # (depending on settings - but shall_skip() takes care of that)
+    subs = [sub for sub in subs if not
+            shall_skip(path.join(root, sub, INITPY), opts)]
+
     # if there are some package directories, add a TOC for theses subpackages
     if subs:
         text += format_heading(2, 'Subpackages')
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,7 @@ Features added
 Bugs fixed
 ----------
 
+* #4334: sphinx-apidoc: Don't generate references to non-existing files in TOC
 * #4206: latex: reST label between paragraphs loses paragraph break
 * #4231: html: Apply fixFirefoxAnchorBug only under Firefox
 * #4221: napoleon depends on autodoc, but users need to load it manually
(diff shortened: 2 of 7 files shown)


sphinx/apidoc.py
```diff
@@ -116,7 +116,12 @@ def create_package_file(root, master_package, subroot, py_files, opts, subs, is_
         text += '\n'
 
     # build a list of directories that are szvpackages (contain an INITPY file)
-    subs = [sub for sub in subs if path.isfile(path.join(root, sub, INITPY))]
+    # and also checks the INITPY file is not empty, or there are other python
```
arcivanov, line 119: an empty INITPY is still a valid INITPY making the directory a package as opposed to namespace

zufallsgenerator: True. But before this patch there would be a reference to that package, although the package .rst file would never be created. It seems to have been the behaviour of apidoc.py since many years to exclude those packages.

See:
https://github.com/sphinx-doc/sphinx/blob/87cebf1a070a2c8956ef2939f43d53bf826600e9/sphinx/apidoc.py#L159-L169

That's why I'm reluctant to change an existing deliberate design, but tried to fix the inconsistency. If we change the behaviour in apidoc.py to not skip empty __init__.py files in shall_skip it'd still be consistent. @birkenfeld ?

arcivanov: Sounds good.

tk0miya (approved):
I also feel the skip for empty `__init__.py` is strange. But it is the default behavior of apidoc. So let's keep this as is. :-p

tk0miya:
Thank you for contribution!

PR #4337, opened 2017-12-22
The great toxification (part 6)

stephenfin:
Subject: Integrate [codecov.io](https://codecov.io).

Let's start running and using coverage stats in our CI. This will allow us to figure out where the gaps are in testing. From the looks of things, our coverage is already pretty high 🎆 🍾 

### Feature or Bugfix
- Feature

### Purpose
To paraphrase the Zen of Python (import this):

> tox is one honking great idea -- let's use more of it!

Start modernizing the test infrastructure that's using make and other hand-crafted tools by replacing them with tox-based derivatives.

**This is part six.**

### Detail
This is a multi-step process, with the end goal of completely removing the Makefile and running everything through `tox`. The first fours steps of these can be done in parallel, but combined they block the final two.

1. Make `tox` a little more friendly by enabling extras like coverage, test profiling etc. by default, and closing gaps with the current Makefile targets such as PyLint and building docs in particular formats
2. Replace/remove custom tooling found in `utils` in favor of modern alternatives like `flake8`
3. Centralize as much configuration as possible inside `setup.cfg` and `tox.ini`, to avoid polluting the top level directory.
4. Clean up requirements (`test-reqs.txt`, `setup.py`) to ensure virtualenv and non-virtualenv builds are identical

The remaining two steps are blocked by the above:

5. Switch CIs to use `tox` instead of the Makefile, and update `CONTRIBUTING.rst` to refer to tox. Deprecate the Makefile  (**this change**)
6. (Bonus) Integrate [codecov.io](https://codecov.io) to make use of the coverage stats we now have

In this part, we make the following changes:

- gitignore: Ignore 'htmlcov' directory

  This is generated if you run 'coverage html'. Hide it from Git.

- tox: Report coverage to user

  Run 'coverage report' after a successful coverage run. There's enough coverage-related stuff here at this point to warrant its own section.

- travis: Enable codecov coverage

  This necessitates adding some basic coverage-py configuration [1] and making sure the pytest-cov plugin uses said configuration [2]. Badges are included.

  Note that we do not run the 'coverage' tox target, which is reserved for users.

  [1] https://github.com/codecov/example-python
  [2] https://bitbucket.org/ned/coveragepy/issues/512/

### Relates
- N/A

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,17 @@ matrix:
     - python: 'pypy'
       env: TOXENV=pypy
     - python: '2.7'
-      env: TOXENV=du13
+      env:
+        - TOXENV=du13
+        - PYTEST_ADDOPTS = --cov sphinx --cov-append --cov-config setup.cfg
     - python: '3.4'
       env: TOXENV=py34
     - python: '3.5'
       env: TOXENV=py35
     - python: '3.6'
-      env: TOXENV=py36
+      env:
+        - TOXENV=py36
+        - PYTEST_ADDOPTS = --cov sphinx --cov-append --cov-config setup.cfg
     - python: 'nightly'
       env: TOXENV=py37
     - python: '3.6'
@@ -35,7 +39,10 @@ addons:
       - imagemagick
 
 install:
-  - pip install -U tox
+  - pip install -U tox codecov
 
 script:
   - tox -- -v
+
+after_success:
+  - codecov
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ TAGS
 .tox
 .venv
 .coverage
+htmlcov
 .DS_Store
 sphinx/pycode/Grammar*pickle
 distribute-*
--- a/README.rst
+++ b/README.rst
@@ -22,6 +22,10 @@
    :target: https://circleci.com/gh/sphinx-doc/sphinx
    :alt: Build Status (CircleCI)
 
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -14,9 +14,13 @@ matrix:
...
     - python: '2.7'
-      env: TOXENV=du13
+      env:
+        - TOXENV=du13
+        - PYTEST_ADDOPTS = --cov sphinx --cov-config setup.cfg
     - python: '3.4'
-      env: TOXENV=py34
+      env:
+        - TOXENV=py34
+        - PYTEST_ADDOPTS = --cov sphinx --cov-config setup.cfg
```
tk0miya: Why do you enable coverages on py34? I think py36 is much better because it's latest release.

tk0miya (changes requested):
It seems broken. could you check it please?

.travis.yml
stephenfin: Done

tk0miya (approved):

tk0miya:
LGTM!

tk0miya:
Thank you always :-)

stephenfin:
@tk0miya As I'm not a maintainer, you need to do the necessary steps to enable coverage on [codecov.io](https://codecov.io/gh). Can you do this, please? You can ping me here or on IRC (stephenfin on freenode) if you have any issues.

tk0miya:
@stephenfin I'd already enabled codecov.io. But it seems not working well. Do you know what happened?
https://codecov.io/gh/sphinx-doc/sphinx

PR #4338, opened 2017-12-22
Test cleanup

stephenfin:
Subject: Make better use of `pytest` features

### Feature or Bugfix

- Refactor

### Purpose

- Get rid of `tests/run.py` in favour of `python -m pytest`

### Detail

- circleci: Install packages ourselves

  Start installing Python dependencies and Sphinx itself as part of the CircleCI job rather than expecting it to be done in the Docker container. This ensures we will use the version of the packages for a given commit.

  This is the other half of sphinx-doc/docker-ci#1.

- tests: Stop explicitly checking for modules

  This will have already been handled by setuptools. If the user isn't using this then they've dug their own grave.

- tests: Use `pytest_report_header`

  This is the recommended way to print extra headers [1].

  [1] https://docs.pytest.org/en/latest/example/simple.html#adding-info-to-test-report-header

- tests: Use `pytest_sessionstart`

  This is the recommended way to do pre-session configuration in pytest if not using session fixtures [1].

  With this, we're able to remove the custom `test/run.py` script in its entirety and run `pytest` like everyone else does.

  [1] https://stackoverflow.com/a/12600154/613428

- Resolve warning filter issues

  There's an issue with `pkgutils` and namespace packages. This has been reported against setuptools [1], but until this is resolved, we simply need to live with it. Ensure said warnings are filtered from `tox` too and remove an unnecessary one.

  [1] https://github.com/pypa/setuptools/issues/1111

### Relates
- N/A


tk0miya (approved):
LGTM :-)
but it seems broken in AppVeyor and Circle CI.
Could you check them please?

tk0miya:
@jfbu could you check this doesn't effect to your work?

jfbu (changes requested):
Stubbornly attached to `make test` ;-), after setting up a Python 3.6 virtual env, and installing Sphinx in it as editable, I needed (unexpectedly) to myself pip install additional dependencies (mock and html5lib) and perhaps another dependency. After that I had an issue which I debugged to originate from a non erased temporary directory ``$TMPDIR/root/_build/latex`` (``TMPDIR`` is set in my environment). Indeed when running ``make test TEST=test_build_latex.py::test_build_latex_doc`` the 6 TeX runs may inherit auxiliary files from previous testing, and this causes failure because the ``SphinxTests.toc`` file contain ``\chapter`` from a ``'manual'`` run but is then used for ``'howto'`` and on first ``\section`` encountered in this toc file, the LaTeX run raises a build error (very illuminatingly named ``perhaps a missing \item`` ;-) ). 

[the LaTeX explanation is a bit short but for TeX hackers what happens is that in article (i.e. ``'howto'``) class `\contentsline{chapter}{foo}{..}` does nothing special a macro is missing and treated as `\relax` but it does print foo, and then when the first `\contentsline{section}{bar}{...}` is encountered TeX is in horizontal mode and an error is raised due to `\l@section` code as defined in LaTeX2e kernel -- giving the non inspiring "perhaps a missing item"]

jfbu:
Additional minor remark, I enjoy the current console output of ``make test``. With this PR, unit tests are first printed as a dot (if success), I find it nicer with the actual name of the unit test displayed on the spot.

jfbu:
The temporary folder issue affects many other targets than latex.  I get

```
============== 3 failed, 1184 passed, 9 skipped in 107.38 seconds ==============
```

after a `rm -fr $TMPDIR/*`, but then on a second `make test`

```
========= 23 failed, 1153 passed, 9 skipped, 11 error in 84.77 seconds =========
```

<strike>
As an aside I am surprised by this

```
/private/var/folders/np/4_9pppfs5....[some path]...warnings/autodoc_fodder.py:docstring of autodoc_fodder.MarkupError:2: WARNING: Explicit markup ends without a blank line; unexpected unindent.
```

but other warnings never have this `/private` at start of this path but use the environment `TMPDIR` as below:

```
/var/folders/np/4_9pppfs5...[path].../warnings/index.rst:19: WARNING: Encoding 'utf-8-sig' used for reading included file '/var/folders/np/4_9pppfs54xdv9m79m7rwx4w0000gn/T/warnings/wrongenc.inc' seems to be wrong, try giving an :encoding: option
```

Any reason testing would write to `/private/$TMPDIR`?

_edit_: on master, I do see a `/private/$TMPDIR/pytest-of-myusername` repertory being created by a `make test`, as well as a file `sphinx-err-tnwye4lc.log` (and possibly three `tmp...` repertories I guess from usage  of ``mktemp``). But on this PR, I get `/private/$TMPDIR/` populated by root tests it seems.
</strike>

sigh... at my locale `/var` is a symlink to `/private/var` (and `/tmp` to `/private/tmp`) 😖

stephenfin:
> after setting up a Python 3.6 virtual env, and installing Sphinx in it as editable, I needed (unexpectedly) to myself pip install additional dependencies (mock and html5lib) and perhaps another dependency.

Test test dependencies aren't installed by default - you need to install them like so:

    pip install -e .[test,websupport]

The reason it's done this way is because `html5lib` and `mock` are only used in testing. Installing them as part of the default requirements is unnecessary. Note that `tox` will do this by default for you 😄 

> After that I had an issue which I debugged to originate from a non erased temporary directory $TMPDIR/root/_build/latex (TMPDIR is set in my environment)

This confuses me slightly. If I'm reading the correctly, it seems running `make test` previously deleted this directory but no longer does? I can't find anything in `run.py` which would do this, hence my confusion. Perhaps you were running `make clean`?

> Additional minor remark, I enjoy the current console output of make test. With this PR, unit tests are first printed as a dot (if success), I find it nicer with the actual name of the unit test displayed on the spot.

Let me see if I can root out the difference here.

stephenfin:
Figured the two issues above out. This should be working now.

jfbu (approved):
This appears now to allow `make test` as formerly, but the console output prefixes each unit test with `tests/`. Also `make test TEST=...` requires the `tests/` prefix being added by user.

It does use `tests/build/root/_build` as formerly, it seems. Some ` tmpf01ciund` type repertories accumulate in `$TMPDIR` but that might have been the case already, due to period of year, I did not investigate if this is case.

Thus LGTM with nits!

sphinx/testing/fixtures.py
```diff
@@ -228,4 +228,6 @@ def tempdir(tmpdir):
     temporary directory that wrapped with `path` class.
     this fixture is for compat with old test implementation.
     """
+    # TODO(stephenfin): Deprecate this in favour of pytest's native tmpdir
+    # extension: https://docs.pytest.org/en/latest/tmpdir.html
```
tk0miya: Hmm? Please tell me why. Sphinx's `tempdir` is a bit useful to me.

tk0miya (changes requested):

tox.ini
```diff
@@ -1,6 +1,7 @@
 [tox]
 minversion = 2.0
 envlist = docs,flake8,mypy,coverage,py{27,34,35,36,py},du{11,12,13,14}
+skipsdist = True
```
tk0miya, line 4: This looks very good!

sphinx/testing/fixtures.py
stephenfin: Because of this line

> this fixture is for compat with old test implementation.

I found it rather difficult to use, in that you need to configure an environment variable to use it and if you don't, it defaults to using '/tmp' instead of a subdirectory in that folder like I thought it would. I will address the TODO in a future change, I assure you 😄

tox.ini
stephenfin: Aye, it speeds up tox initialization massively and it was totally unnecessary, as noted in the commit. Win-win!

sphinx/testing/fixtures.py
tk0miya: Hmm? This is a simple wrapper for `tmpdir` fixture of pytest's. It returns path object instead string.

With using this, we can create a text file like following:
```
def test(tempdir):
    (tempdir / "filename.txt").write_text("hello sphinx!")
```

I believe this is almost same as `tmpdir` fixture. What do you want to go?

stephenfin: Maybe I've misunderstood something here so. What I want to remove is the `_initialize_test_directory` function in `tests/conftest.py`. IMO, by using the `tempdir` fixture, we should automatically get a subdirectory in `TMPDIR` that's used for the entire test. This doesn't seem to be the case. Maybe I've just misunderstood something though. I've dropped this TODO and can look at this in more detail later.

tk0miya (approved):

tk0miya:
Thanks!

PR #3819, opened 2017-05-28
Fix #3807 by using old Makefile in LaTeX build repertory for windows

jfbu:
use old Makefile for latex build repertory; the new one from 1.6.1 is not compatible with Windows syntax, even with Latexmk installed, at least for Japanese documents.

refs: #3807, #3082

Issue #3807, opened 2017-05-26
Cannot generate PDFs with Sphinx version 1.6.1 on Windows. Perl Missing.


NormanDunbar:
Subject: Cannot generate PDFs with Sphinx version 1.6.1 on Windows. Perl Missing.

### Problem
- After upgrading to Sphinx 1.6.1 from 1.5.x, I am no longer able to generate pdf documents with `make latexpdf` which was working before. The problem is that `latexmk` is now in use and requires `perl` to be installed. I don't have `perl` installed and I am unable to install additional software as this is a work computer.

I have looked around, Google, and here on the issues list, and I'm unable to find anything that works. The command `make latexpdf PDFLATEX=pdflatex` as found [here](https://github.com/sphinx-doc/sphinx/issues/2513#issuecomment-216229066), doesn't work - it gives exactly the same errors about `perl` being missing. (See below)

#### Procedure to reproduce the problem
```
make latexpdf
```

#### Error logs / results
```
Running Sphinx v1.6.1
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [latex]: all documents
updating environment: 0 added, 0 changed, 0 removed
looking for now-outdated files... none found
processing UpdatingSuperBASICManual.tex...index Part1 p1_intro p1_software Part2 p2_download Part3 p3_rst Part4 P4_intro

resolving references...
writing... done
copying images... [100%] images\LogIssue.png
copying TeX support files...
done
build succeeded.
latexmk -pdf -dvi- -ps-  'UpdatingSuperBASICManual.tex'
latexmk: The script engine could not be found.
latexmk: Data: scriptEngine="perl.exe"
Makefile:33: recipe for target 'UpdatingSuperBASICManual.pdf' failed
mingw32-make: *** [UpdatingSuperBASICManual.pdf] Error 1

Build finished; the PDF files are in build/latex.
```
#### Expected results
I would expect that when I request a pdf to be built, it would be built. As it did in the previous version, or, in the event that a component of the build has changed dramatiacally, as in this case, that the install/upgrade notes would make the additional dependencies known, and/or an option to use a different utility be supplied to revert to the previous working operation.

Windows isn't the best of operating systems for having decent utilities installed by default! ;-)

### Reproducible project / your project
- Not required? Any project will do, just don't have `perl` installed. 

### Environment info
- OS: Windows 7. 64 bit.
- Python version: 3.5.1
- Sphinx version: 1.6.1
- MikTex: 2.9

[...]

PR #3082, opened 2016-10-23
Use latexmk for pdf (and dvi) targets

jfbu:
This will need some review because I don't know if latexmk is available on all platforms where Sphinx is available. I know latexmk comes with TeXLive installations and I think the TeXLive has its own perl so that latexmk runs (also on windows).

rel #3064, #2513

```
modified:   sphinx/texinputs/Makefile_t
new file:   sphinx/texinputs/latexmkrc
```

PR #4343, opened 2017-12-24
Add 'latexpdf' builder

stephenfin:
Subject: Add `latexpdf` builder

### Feature or Bugfix
- Feature
- Bugfix

### Purpose
- Add a `latexpdf` builder as a replacement for the equivalent target in make mode

### Detail

- make_mode: Convert to use 'pdflatex' builder
    
  This simplifies the file yet further.
    
- builders: Add LaTeXPDFBuilder
    
  This is a subclass of LaTeXBuilder that simply invokes the Makefile generated by same to build a PDF upon successful LaTeX generation. A future change may modify this to call the underlying commands natively instead of via `make`.


tk0miya:
LGTM, but we have to test this on Windows before merging.
(But I don't have it)

sphinx/builders/latex.py
```diff
@@ -256,6 +258,23 @@ def copy_image_files(self):
...
+    """
+    Utility builder that builds LaTeX output followed by a PDF.
+    """
+    name = 'latexpdf'
+
+    def finish(self):
+        # type: () -> None
+        super(LaTeXPDFBuilder, self).finish()
+        makecmd = os.environ.get('MAKE', 'make')
+        # TODO(stephenfin): Call things natively instead of using the Makefile
```
jfbu: the latex Makefile contains set-up for latexmk, if you move the logic here you will have to take into account setting of latex_engine.

edit: and `make all-pdf`  can indeed produce multiple pdf files as there can be multiple tex files, for example from 

```
latex_documents = [
    (master_doc, 'FOO.tex', 'FOO Documentation',
     'JFB', 'manual'),
    (master_doc, 'FOOhowto.tex', 'FOO Documentation',
     'JFB', 'howto'),
]
```

and, the Japanese documents need to run extractbb on image files.

There is quite some logic built-in in the latex Makefile, and the reason we don't have a nice one for Windows is not only due to GNU/Make possibly lacking, it is also in syntax of system calls which are set-up in the latexmkrc and latexmkjarc files.

Hence, I am not sure what "natively" means, because it is always shell dependent in a way, in the end.

sphinx/make_mode.py
```diff
@@ -194,10 +194,10 @@ def build_latex(self):
 
     def build_latexpdf(self):
         # type: () -> int
-        if self.run_generic_build('latex') > 0:
+        if self.run_generic_build('latexpdf') > 0:
             return 1
-        with cd(self.builddir_join('latex')):
-            return subprocess.call([self.makecmd, 'all-pdf'])
+        print("Build finished; the PDF and LaTeX files are in %s." %
+              self.builddir_join('latex'))
```
jfbu: the py36 test on Travis complains about "missing return statement"

jfbu (commented):
LGTM with nits, I can't check on Windows.

sphinx/builders/latex.py
stephenfin: Sorry, I meant using Python and `subprocess`, rather than the `make` wrapper. I'd like to investigate this further but it's difficult to do, as you say, and might not even be possible. As such, I've dropped the TODO and replaced it with a note explaining why we use the `make` command (basically what you just said)

sphinx/make_mode.py
stephenfin: The need for explicit returns is one thing I really don't like about `mypy`. I've added this now though 😄 

EDIT: Wait - I was mixing this up with something else. This was actually wrong. My bad!

sphinx/builders/latex.py
stephenfin: > There is quite some logic built-in in the latex Makefile, and the reason we don't have a nice one for Windows is not only due to GNU/Make possibly lacking, it is also in syntax of system calls which are set-up in the latexmkrc and latexmkjarc files.

@jfbu Could you provide a little more info on this, just so I have the context? It seems that we don't use `latexmk` on Windows even though it's available. Is this because of those configuration files you're referring to above?

stephenfin:
> LGTM, but we have to test this on Windows before merging.
> (But I don't have it)

I don't think this will work on Windows. I was going to add a check to ensure we're on Linux/macOS, but decided not to because I wasn't 100% sure. Perhaps I should add it anyway until we've assessed it?

jfbu:
Sorry, I must revise my assessment:

- wIth this PR, `make latex` becomes duplicate because it uses its `build/latex` repertory, whereas `make latexpdf` uses `build/latexpdf`. Hence this creates duplicate builds and wastes disk space.

- as soon as one starts working with more complex documents which produce multiple `.tex` files, it is important to be able to build individual PDF and not the whole documentation suite when one has made a change in source. Thus, one will do `make latex`, and then `make -C _build/latex FOO.pdf` for example. Additionally, perhaps one will be content with `make latex` and delay the PDF build to later. But the  `latexpdf` as builder confuses the situation because it does not have the granularity to build only one PDF and moreover does the whole work in another build repertory (repeat of previous item).

Thus, I would ask for `latexpdf` to use same build repertory as `latex` per default.

There _is_ currently  a bug with Sphinx's `make latexpdf`: it rebuilds the latex files even if nothing has changed in the sources and a previous `make latex` completed. This is waste of time. One can observe this by building the Sphinx own docs.

The `make` approach to PDF builds has various utilities for advanced LaTeX users (which might well constitute a very small subset of Sphinx users ;-), but still ...), for example `make -C _build/latex clean` is sometimes necessary to get rid of LaTeX auxiliary files. This includes changing the document language or switching from <strike>``howto`` to ``manual``</strike> ``manual`` to ``howto``, or also sometimes after extensive changes in sources; but then I do top-level ``make clean`` per habit, although it removes also other build targets previous works.

Now that I think about it, I find it confusing that `sphinx-build` "make-mode" allows `clean` as target. Imho, `sphinx-build` should only have its proper builder names as targets, and employ an option to say not to build but to clean, and there would be `--cleanall`  for example. Indeed for reasons mentioned above sometimes one focuses on one target and for time efficiency it would be nice to remove previous build of only one target and not lose all build repertories.

Generally speaking, `sphinx-build` belongs to Python world and should govern Python activities. Then the GNU/Linux make approach is ideal for governing at a higher level other activities such as producing the PDF which currently is completely non-Python thing (after the latex files production has been done.)

jfbu (changes requested):
I would like latexpdf and latex builders to share same target repertory and to latexpdf not to recreate latex sources if latex builder was exercised right before.

sphinx/builders/latex.py
jfbu: @stephenfin sorry for delay. 

- First, a tool such as  ``Latexmk`` is needed: formerly Sphinx did 5 runs of ``pdflatex`` which in most cases is too much and in some cases possibly not enough (although circumstances are rare and possibly hard to recreate from a Sphinx project). At read the docs, last time I checked they do only 2 runs of ``pdflatex`` and this is not enough rtfd/readthedocs.org#2857; we needed to apply a trick at 6422f7b8c for our own docs, it is probable the problem is widespread across PDF builds at RTD. 

- Then ``Latexmk`` is a Perl script, which needs to be configured via a "latexmkrc" file. There we set up its Perl variables in particular for shell commands. Formerly (i.e. prior to using ``Latexmk``) all was in Makefile which miraculously worked on Windows. But at https://github.com/sphinx-doc/sphinx/blob/master/sphinx/texinputs/latexmkjarc#L3 in particular we need to workaround a bug of ``mendex`` executable which contrarily to ``makeindex`` raises error status if the ``foo.idx`` file generated from ``foo.tex`` is empty. This workaround was not needed when the rule was executed from Makefile (the rule had a ``-`` to say to ignore error). Now the Linux syntax of this is definitely not Windows-like, and for lack of expertise and time we had to postpone production of Windows-compatible syntax.

- Finally, many Windows users do not have a Perl available, and although [Latexmk](http://personal.psu.edu/jcc8/software/latexmk/) provides information how to install it, it looks as if activating its use by default for Sphinx builds will trigger many users into reporting issues (as #3807 which led to #3819).

For background on how Sphinx adopted Latexmk see #3082, but some tweaks were applied after the final version was merged.

tk0miya:
>I don't think this will work on Windows. I was going to add a check to ensure we're on Linux/macOS, but decided not to because I wasn't 100% sure. Perhaps I should add it anyway until we've assessed it?

Now we also support Windows platform and it works well. So such change should be tested in Windows.
Unfortunately, active developers don't have windows testbed. So we should call a help for users.
(I know @shimizukawa has it. But he is very busy at last months.)

@shimizukawa can you check this in this winter break if possible?

tk0miya:
>There is currently a bug with Sphinx's make latexpdf: it rebuilds the latex files even if nothing has changed in the sources and a previous make latex completed. This is waste of time. One can observe this by building the Sphinx own docs.

@jfbu  Could you file this as a new issue please? I will take it later. (maybe after releasing 1.7...)

stephenfin:
This needs another approach. Abandoning this until I figure that out.

Issue #4344, opened 2017-12-24
Docs build is currently broken


stephenfin:
Subject: More than one target for cross-reference `add_object_type`

### Problem
Building docs with current `master` and the `-W` flag results in an error.

#### Procedure to reproduce the problem
```
$ sphinx-build doc/ build/ -b html -W
```

#### Error logs / results
```
Running Sphinx v1.7+/a9efb2517
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 70 source files that are out of date
updating environment: 70 added, 0 changed, 0 removed
reading sources... [100%] websupport                                                                                                                                                                               
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] websupport                                                                                                                                                                                

Warning, treated as error:
/home/sfinucan/Development/sphinx/sphinx/doc/markup/inline.rst:65:more than one target found for cross-reference u'add_object_type': sphinx.domains.Domain.add_object_type, sphinx.application.Sphinx.add_object_type
```

#### Expected results
Docs should build

jfbu:
I had seen the _warning_ using ``make -C doc html`` but build completes then :smile: ... 

The problem possibly originates with 838f7fc29 (#4305), can you check @tk0miya ?

stephenfin:
This is resolved in #4346 and should not happen again thanks to #4349

PR #4346, opened 2017-12-27
docs: make explicitly :meth: reference (refs: #4344)

tk0miya:
refs: #4344


stephenfin (approved):
Looks good to me. We should start building docs in the gate to prevent this happening again.

tk0miya:
Thank you for reviewing :-)

PR #4347, opened 2017-12-27
Diet tests on appveyor

tk0miya:
I dropped a test pattern for docutils-0.13.1 and py27 on Windows in this commit.

Because we already runs tests for several versions of docutils on Travis CI.
So I think it is unnecessary.

@shimizukawa Is there any special reason for it?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,19 +6,12 @@ environment:
 
   matrix:
     - PYTHON: 27
-      DOCUTILS: 0.13.1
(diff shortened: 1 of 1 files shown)


stephenfin (approved):
Agreed. This should test basic functionality on Windows and no more. We may want to move to more integration-style testing (e.g. building PDFs) but that's a future change.

shimizukawa:
+1 to diet.

tk0miya:
Thanks,!

PR #4349, opened 2017-12-28
travis: Build own docs as a test

tk0miya:
refs: #4344 and #4346 

I know this increases our test times, but this is important for the health of our docs.


stephenfin (approved):
It's failing CircleCI but that is unrelated. Looks good to me.

PR #4352, opened 2017-12-28
Fix missing metainfo for sphinx.io

jakobandersen:
Subject: ``sphinx.io`` extension disables parallel reading/writing

The PR simply adds missing metainfo, asserting that parallel reading/writing is ok.
Note: I don't know what extension is doing and haven't actually checked if the change is safe. However, I guess a builtin extension shouldn't disable parallel builds.

(Would it be possible to have a generic parallel build test that fails if not all builtin extensions are parallel read/write safe?)


tk0miya (approved):

tk0miya:
Thanks!

>(Would it be possible to have a generic parallel build test that fails if not all builtin extensions are parallel read/write safe?)

I'll try it later

PR #4320, opened 2017-12-18
Simplified 'sphinx-build'

stephenfin:
Subject: Simplify calling of `sphinx-build`

### Feature or Bugfix
Feature

### Purpose
Make `sphinx-build` easier to call by moving configuration of the build directory to the configuration file and auto-detecting the source directory. This will allow most users to call `sphinx-build` without any arguments.

### Detail
- `sphinx-build`: Configure `outdir` from `conf.py`

  This is part one of two of the "make 'sphinx-build' easier to call" effort. In this part, we allow users to define the build directory in the 'conf.py' file. This means a user no longer needs to pass the option via the command line, though they can continue to do so if they wish.

  We use the behavior of the setuptools extension, whereby the results are output to a subfolder in the outdir named after the code of the builder. For example, if using the 'html' builder with 'output_dir = "test"', files would be written to 'test/html'.

- `sphinx-build`: Guess `sourcedir` if none provided

  This is part two of two of the "make 'sphinx-build' easier to call" effort. In this part, we attempt to guess where the source directory is if the user does not provide one. This is possible because many users use one of three directory names - doc, docs, or Documentation - for their documentation. We check these, along with a possible 'source' subdirectory, which is based on Sphinx's own defaults, for a 'conf.py' file. Is one is found, we assume this is the documentation source. This means a user no longer needs to pass any argument to 'sphinx-build' is they do not wish to.

  This is, of course, best-effort, and users can continue to provide these options manually if they so wish.

You can validate this right now by pulling down the pull request and running `sphinx-build` without any arguments. The required change has been made to `conf.py`.

PR #4354, opened 2017-12-28
Add 'Builder.epilog' attribute

stephenfin:
Subject: Add `epilog` attribute to `sphinx.builder.Builder`

### Feature or Bugfix
- Feature

### Purpose
The `Makefile`, `make.bat`, and `sphinx-build -M` (make mode) tools all output very similar messages on successful build. Reduce this duplication (and the need for the latter) by moving this into the builders themselves.

### Detail
- builders: Add `Builder.epilog` option

  This allows builders to emit a final epilog message containing information such as where resulting files can be found. This is only emitted if the build was successful.

  This allows us to remove this content from the `make_mode` tool and the legacy `Makefile` and `make.bat` templates. There's room for more dramatic simplification of the former, but this will come later.

- `make_mode`: Remove unnecessary `make_*` functions

  These are handled by the default case.

- `Makefile`: Remove unnecessary targets

  Most of these are not necessary now that we're not printing different messages for various builders.

- `make.bat`: Remove unnecessary targets

  As with the Makefile previously, these are not necessary now that we're not printing anything different for various builders.

- `Makefile`: Make `SOURCEDIR` configurable

  It's unlikely that anyone will need to do this but at least give them the opportunity.

- `make.bat`: Make `SOURCEDIR` configurable

  It's unlikely that anyone will need to do this but at least give them the opportunity.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -338,6 +338,13 @@ class Sphinx(object):
                                  (status, self._warncount)))
             else:
                 logger.info(bold(__('build %s.') % status))
+
+            if self.statuscode == 0 and self.builder.epilog:
+                logger.info('')
+                logger.info(self.builder.epilog % {
+                    'outdir': path.relpath(self.outdir),
+                    'project': self.config.project
+                })
         except Exception as err:
             # delete the saved env to force a fresh build next time
             envfile = path.join(self.doctreedir, ENV_PICKLE_FILENAME)
--- a/sphinx/builders/texinfo.py
+++ b/sphinx/builders/texinfo.py
@@ -9,6 +9,7 @@
     :license: BSD, see LICENSE for details.
 """
 
+import os
 from os import path
 
 from docutils import nodes
@@ -97,6 +98,12 @@ class TexinfoBuilder(Builder):
     """
     name = 'texinfo'
     format = 'texinfo'
+    epilog = 'The Texinfo files are in %(outdir)s.'
+    if os.name == 'posix':
+        epilog += ("\nRun 'make' in that directory to run these through "
+                   "makeinfo\n"
+                   "(use 'make info' here to do that automatically).")
+
     supported_image_types = ['image/png', 'image/jpeg',
                              'image/gif']
     default_translator_class = TexinfoTranslator
--- a/sphinx/make_mode.py
+++ b/sphinx/make_mode.py
@@ -101,95 +101,60 @@ class Make(object):
         # type: () -> int
         if self.run_generic_build('html') > 0:
             return 1
-        print()
-        print('Build finished. The HTML pages are in %s.' % self.builddir_join('html'))
         return 0
 
     def build_dirhtml(self):
         # type: () -> int
         if self.run_generic_build('dirhtml') > 0:
             return 1
-        print()
-        print('Build finished. The HTML pages are in %s.' %
-              self.builddir_join('dirhtml'))
         return 0
 
     def build_singlehtml(self):
         # type: () -> int
         if self.run_generic_build('singlehtml') > 0:
             return 1
-        print()
-        print('Build finished. The HTML page is in %s.' %
-              self.builddir_join('singlehtml'))
         return 0
 
     def build_pickle(self):
         # type: () -> int
         if self.run_generic_build('pickle') > 0:
             return 1
-        print()
-        print('Build finished; now you can process the pickle files.')
         return 0
 
     def build_json(self):
         # type: () -> int
         if self.run_generic_build('json') > 0:
             return 1
-        print()
-        print('Build finished; now you can process the JSON files.')
         return 0
 
     def build_htmlhelp(self):
         # type: () -> int
         if self.run_generic_build('htmlhelp') > 0:
             return 1
-        print()
-        print('Build finished; now you can run HTML Help Workshop with the '
-              '.hhp project file in %s.' % self.builddir_join('htmlhelp'))
         return 0
 
     def build_qthelp(self):
         # type: () -> int
         if self.run_generic_build('qthelp') > 0:
             return 1
-        print()
-        print('Build finished; now you can run "qcollectiongenerator" with the '
-              '.qhcp project file in %s, like this:' % self.builddir_join('qthelp'))
-        print('$ qcollectiongenerator %s.qhcp' % self.builddir_join('qthelp', proj_name))
-        print('To view the help file:')
-        print('$ assistant -collectionFile %s.qhc' %
-              self.builddir_join('qthelp', proj_name))
         return 0
 
     def build_devhelp(self):
         # type: () -> int
         if self.run_generic_build('devhelp') > 0:
             return 1
-        print()
-        print("Build finished.")
-        print("To view the help file:")
-        print("$ mkdir -p $HOME/.local/share/devhelp/" + proj_name)
-        print("$ ln -s %s $HOME/.local/share/devhelp/%s" %
-              (self.builddir_join('devhelp'), proj_name))
-        print("$ devhelp")
         return 0
 
     def build_epub(self):
         # type: () -> int
         if self.run_generic_build('epub') > 0:
             return 1
-        print()
-        print('Build finished. The ePub file is in %s.' % self.builddir_join('epub'))
         return 0
 
     def build_latex(self):
         # type: () -> int
         if self.run_generic_build('latex') > 0:
             return 1
-        print("Build finished; the LaTeX files are in %s." % self.builddir_join('latex'))
-        if os.name == 'posix':
-            print("Run `make' in that directory to run these through (pdf)latex")
-            print("(use `make latexpdf' here to do that automatically).")
         return 0
 
     def build_latexpdf(self):
@@ -210,19 +175,12 @@ class Make(object):
         # type: () -> int
         if self.run_generic_build('text') > 0:
             return 1
-        print()
-        print('Build finished. The text files are in %s.' % self.builddir_join('text'))
         return 0
 
     def build_texinfo(self):
         # type: () -> int
         if self.run_generic_build('texinfo') > 0:
             return 1
-        print("Build finished; the Texinfo files are in %s." %
-              self.builddir_join('texinfo'))
-        if os.name == 'posix':
-            print("Run `make' in that directory to run these through makeinfo")
-            print("(use `make info' here to do that automatically).")
         return 0
 
     def build_info(self):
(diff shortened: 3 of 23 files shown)


sphinx/application.py
```diff
@@ -338,6 +338,17 @@ def build(self, force_all=False, filenames=None):
                                  (status, self._warncount)))
             else:
                 logger.info(bold(__('build %s.') % status))
+
+            if self.statuscode == 0 and self.builder.epilog:
+                # TODO(stephenfin): Should this use logging or print?
```
stephenfin: I would like your thoughts on both of these items, @tk0miya 😄

sphinx/builders/texinfo.py
```diff
@@ -97,6 +98,12 @@ class TexinfoBuilder(Builder):
     """
     name = 'texinfo'
     format = 'texinfo'
+    epilog = 'The Texinfo files are in %(outdir)s.'
+    if os.name == 'posix':
+        epilog += ("Run 'make' in that directory to run these through "
```
jfbu: a space (or rather a ``\n``) is missing at start of string

stephenfin: Done

sphinx/application.py
tk0miya: Please use logging module to output messages, warnings and errors.

sphinx/application.py
```diff
@@ -338,6 +338,17 @@ def build(self, force_all=False, filenames=None):
...
+                logger.info('')
+                # TODO(stephenfin): Should this be printf-based or
+                # .format-based?
```
tk0miya: Either is fine. Personally, I preferred printf-based one.

sphinx/application.py
```diff
@@ -338,6 +338,17 @@ def build(self, force_all=False, filenames=None):
                                  (status, self._warncount)))
             else:
                 logger.info(bold(__('build %s.') % status))
+
+            if self.statuscode == 0 and self.builder.epilog:
```
tk0miya, line 342: What do you think about moving this to the end of `Builder.build()`? I think emitting epilog message is responsibility of builder itself.

In addition, I'm worried about `epilog` is better way or not. I think a method of builder is also good like `Builder.show_finish_message()`.

sphinx/application.py
stephenfin: OK, I'll stick with printf. None of the disadvantages I can see with `printf` would affect us here.

sphinx/application.py
stephenfin: > What do you think about moving this to the end of `Builder.build()`? I think emitting epilog message is responsibility of builder itself.

I can do this. The reason I kept it separate was I wasn't sure if `Builder.build()` was ever overridden by subclasses. If it was, we could end up emitting the message before any subclasses were called. For example:

    class MySpecialBuilder(Builder):
        ...
        def build(self, docnames, summary=None, method='update'):
            super(MySpecialBuilder, self).build(docnames, summary, method)
            # do something else here

It's your call though. Let me know what you'd prefer.

> In addition, I'm worried about epilog is better way or not. I think a method of builder is also good like `Builder.show_finish_message()`.

I prototyped with both a function and a property but both seemed overkill. The logic is always the same: if the build succeeds then emit the message. Using `printf` formatting will also ensure we get consistent builder messages. Is there any reason _not_ to do things this way?

stephenfin: > It's your call though. Let me know what you'd prefer.

Actually, there was another reason. I wanted the summary message to come *after* the `build succeeded` message. For example:

    ...
    build succeeded

    You can now process the pickle files in build/pickle.

I wasn't able to do this if I called this as part of `Builder.finish()`. If we move this here, we should also move the generation of the `build succeeded` message. I don't know if we want to do this.

tk0miya: > If it was, we could end up emitting the message before any subclasses were called.
...
>Actually, there was another reason. I wanted the summary message to come after the build succeeded message. 

Reasonable. My idea was wrong. It should be shown on `Builder.build()`.
Please forget this comment.

> Is there any reason not to do things this way?

My worry is how to customize the epilog message on subclasses.
If they want to show more complicated message, an attribute is not extensible.
For example, it is hard to use `out_suffix` to the message. On the other hand, a method is easy to override in subclasses.

I can't determine the interface is important or not. So I'd like to hear your idea.

stephenfin: > > Is there any reason not to do things this way?
> 
> My worry is how to customize the epilog message on subclasses.
> If they want to show more complicated message, an attribute is not extensible.
> For example, it is hard to use `out_suffix` to the message. On the other hand, a method is easy to override in subclasses.
> 
> I can't determine the interface is important or not. So I'd like to hear your idea.

Sorry for the delay. The `epilog` attribute covers all the provided builders. Also, if users _really_ need to override this, then can use properties.

    @property
    def epilog(self):
        ...

Personally, I think this is adequate.

tk0miya: fair enough

sphinx/make_mode.py
```diff
@@ -97,101 +97,6 @@ def build_help(self):
             if not osname or os.name == osname:
                 print('  %s  %s' % (blue(bname.ljust(10)), description))
 
-    def build_html(self):
-        # type: () -> int
-        if self.run_generic_build('html') > 0:
-            return 1
-        print()
-        print('Build finished. The HTML pages are in %s.' % self.builddir_join('html'))
```
tk0miya, line 105: It seems "Build finished." message has been lost. Is this intended?

sphinx/templates/quickstart/Makefile_t
```diff
@@ -5,14 +5,16 @@
...
 SPHINXBUILD  ?= sphinx-build
 PAPER        ?=
+SOURCEDIR     = {{ rsrcdir }}
 BUILDDIR      = {{ rbuilddir }}
 
 # Internal variables.
 PAPEROPT_a4     = -D latex_elements.papersize=a4
 PAPEROPT_letter = -D latex_elements.papersize=letter
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) {{ rsrcdir }}
+# $(O) is meant as a shortcut for $(SPHINXOPTS)
```
tk0miya, line 14: Do you know where `$(O)` comes from? Is this special variable for GNU Make?

sphinx/templates/quickstart/Makefile_t
```diff
@@ -187,49 +89,9 @@ info:
...
-	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
 
-.PHONY: dummy
-dummy:
-	$(SPHINXBUILD) -b dummy $(ALLSPHINXOPTS) $(BUILDDIR)/dummy
-	@echo
-	@echo "Build finished. Dummy builder generates no files."
+# Catch-all target: route all unknown targets to Sphinx
+.PHONY: Makefile
+%: Makefile
```
tk0miya, line 95: Note: this Makefile allows any argument, so this is like a Makefile for make-mode.
I think it's okay.

sphinx/make_mode.py
stephenfin: Yes. We already emit `build succeeded` or `build finished with problems` [here](https://github.com/sphinx-doc/sphinx/blob/3b01697e11445ddfeda891aa53d410259fab6163/sphinx/application.py#L333-L340). I thought this was duplicate information.

sphinx/templates/quickstart/Makefile_t
stephenfin: No, it's just a shortcut added when make_mode was added. @jfbu noted that he uses this so I added it here too

sphinx/templates/quickstart/Makefile_t
stephenfin: Yes. This will break if a user passes in an unknown builder, but Sphinx should show sufficient information. I can harden this if it would help (i.e. only allow one argument) but I don't know if that's necessary? Your call 😄

sphinx/make_mode.py
tk0miya: okay, I understand

sphinx/templates/quickstart/Makefile_t
tk0miya: Is it shell variable like `O=... make html` ?
Okay, I almost understand. I don't know such shortcut.

sphinx/templates/quickstart/Makefile_t
tk0miya: Could you show me the result of `make foo bar`? Is sphinx-build invoked twice?

Anyway, make-mode is not used by default. So almost of users would not see this file. So either is fine to me.

tk0miya (approved):
LGTM

tk0miya:
@shimizukawa do you have any opinion? I will merge this before 1.7 feature freeze.

sphinx/templates/quickstart/Makefile_t
stephenfin: Exactly. It was added as [part of make mode](https://github.com/sphinx-doc/sphinx/blob/1.6.6/sphinx/templates/quickstart/Makefile.new_t#L17-L18). It's just a usability helper.

sphinx/templates/quickstart/Makefile_t
stephenfin: > Could you show me the result of make foo bar? Is sphinx-build invoked twice?

Sure. First, using valid builder names.

```
$ make html dirhtml                                 
sphinx-build -b "html" -d build/doctrees    source "build/html"
Running Sphinx v1.7+
loading pickled environment... done
updating environment: 0 added, 0 changed, 0 removed
no targets are out of date.
build succeeded.

The HTML pages are in build/html.
sphinx-build -b "dirhtml" -d build/doctrees    source "build/dirhtml"
Running Sphinx v1.7+
loading pickled environment... done
building [dirhtml]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
no targets are out of date.
build succeeded.

The HTML pages are in build/dirhtml.
```

Then using invalid builders.

```
$ make foo html
sphinx-build -b "foo" -d build/doctrees    source "build/foo"
Running Sphinx v1.7+

Sphinx error:
Builder name foo not registered or available through entry point
make: *** [Makefile:96: foo] Error 2
```

> Anyway, make-mode is not used by default. So almost of users would not see this file. So either is fine to me.

I think you mean make-mode *is* used by default. This was changed in Sphinx 1.6, I think? This affects the non-make-mode change. Users won't see it yet, but they will when I deprecate make mode in the future 😄

shimizukawa:
+0. 

I think that changing the old Makefile conflicts with the meaning of the `--no-use-make-mode` option. However, since we should remove the old Makefile feature in the near future, I think that there is no big problem with this PR.

tk0miya:
Okay, let's go forward.

stephenfin:
> However, since we should remove the old Makefile feature in the near future, I think that there is no big problem with this PR.

To be clear, I'm hoping to do the opposite. I want to remove (well, deprecate and later remove) make mode by making `sphinx-build` easier to use. I won't achieve this in 1.7, but I hope to do so in 1.8.

tk0miya:
I have to say no to you. We should keep make-mode for a while to keep compatibility.
There are many generated `Makefile` in many repository. So removing make-mode means breaking these builds.
At same time, it means non-make-mode also should not be removed for a while.

Of course, I also feel make-mode is not good. So I can understand and agree with you to remove it in future version. But is is not during 1.x. (I can't say it is okay in 2.0)

So all we can do now is refactoring it simple or creating new CLI (called `sphinx` and its subcommands). In any case, we have to discuss about new `sphinx-build` command #4320 (sorry for late, but I still remember it. I'm in busy for next release now)

stephenfin:
> I have to say no to you. We should keep make-mode for a while to keep compatibility. There are many generated Makefile in many repository. So removing make-mode means breaking these builds. At same time, it means non-make-mode also should not be removed for a while.
>
> Of course, I also feel make-mode is not good. So I can understand and agree with you to remove it in future version. But is is not during 1.x. (I can't say it is okay in 2.0)

Oh, of course. I'm talking long term. We can't break users. I would expect a major version bump before we can remove it.

> So all we can do now is refactoring it simple or creating new CLI (called `sphinx` and its subcommands). In any case, we have to discuss about new `sphinx-build` command #4320 (sorry for late, but I still remember it. I'm in busy for next release now)

I'm working on this. Expect a pull request by tomorrow :smile:

PR #4355, opened 2017-12-28
templates: Remove 'epub3' targets

stephenfin:
This appears to have been missed in commit 69d1dfaa. One could add an alias for this to restore behavior, but no one has reported the issue since Sphinx meaning this seems unnecessary.


tk0miya:
You're right! Good PR!

PR #4356, opened 2017-12-28
doc: Add missing sphinx-apidocs opt, envvar

### LINKED ISSUE #2250: Document sphinx-apidoc --separate
At least the `--separate` option is missing from [the docs for `sphinx-apidoc`](http://www.sphinx-doc.org/en/stable/invocation.html#invocation-of-sphinx-apidoc).
I didn't check whether there's any other options missing overall.

stephenfin:
Add docs for `--module-first` option and `SPHINX_APIDOC_OPTIONS` environment variable. Per the closest thing we have to official man page guidelines [1]:

    ENVIRONMENT

    lists all environment variables that affect the program or function
    and how they affect it.

[1] https://linux.die.net/man/7/man-pages
Fixes #2250

### DIFF
--- a/doc/man/sphinx-apidoc.rst
+++ b/doc/man/sphinx-apidoc.rst
@@ -91,7 +91,7 @@ Options
 
    Interpret paths recursively according to PEP-0420.
 
-.. option:: -M
+.. option:: -M, --module-first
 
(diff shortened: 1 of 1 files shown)


tk0miya:
Thanks!

Issue #3967, opened 2017-07-29
Smart-quotes should not operate on the doctree


matthew-brett:
As of Sphinx 1.6 (and commit https://github.com/sphinx-doc/sphinx/commit/edc7f30b9cc92f4d721eeef62ffa826f23a09a9f) the smart quoting search / replace operates at the parsing step of the doctree, rather than at the output (Writer) stage.  See the commit.

I don't think this change is correct, although I can see it is convenient from the point of view of using docutils code.  It forces all output writers to deal with fancy unicode characters that were not in the original ReST file, to no good purpose, it seems to me.  For example, the fancy quotes now appear in the output of the `text` builder, when previously the output text files would have only the ASCII characters that were in the original ReST.   This makes the text files more difficult to parse for simple text readers.

For my own application, I am building Markdown and Python script versions of pages.  For the same reason, these Markdown / Python code files have unexpected and unwanted fancy quote unicode in them now, when they did not have them previously.

See [this thread](https://groups.google.com/forum/#!topic/sphinx-dev/v43JedWILR0) for some extra discussion, and some breakage introduced by this change.

Would you consider reverting commit https://github.com/sphinx-doc/sphinx/commit/edc7f30b9cc92f4d721eeef62ffa826f23a09a9f ? Or refactoring so that the transformation is at the writer stage rather than the doctree reading stage?

jfbu:
I am in favour of at least not applying Smart quotes by default, as it affects all writers. It is currently applied by default because this was the former default setting of [html_smarty_pants](http://www.sphinx-doc.org/en/stable/config.html#confval-html_use_smartypants). I have tried to clarifiy the docs (5dbfc43ddada57a3b7bf6ee4cf1e8e07a6a58db2, 6a897af1d9e22c0f933edeee7714114a03ff2b61), but it is still quite deficient.

I agree that using smart quotes by default for all writers was a breaking change, arguably a bug.

mitya57:
In my opinion Unicode is no problem in 2017. But if it really causes issues for you, you can disable smart quotes via `docutils.conf`.

matthew-brett:
[...]

Issue #4142, opened 2017-10-12
Sphinx fails to distinguish open and close quotation marks


cocoatomo:
Subject: On the translation project of Python document, smart quotes, the feature from docutils, fails to determine whether some quotation mark is open or close.

### Problem
- docutils transforms `"Title"` to `”Title”` (`&#8221`, `T`, `i`, `t`, `l`, `e`, `&#8221`), or `」Title」` with smart quotes enabled and language set to Japanese.

#### Procedure to reproduce the problem
```bash
git clone https://github.com/python-doc-ja/cpython-doc-intl.git
cd cpython-doc-intl
git checkout 87caf1ccad1457387bec57579fc742af3f2b703b

git clone https://github.com/python-doc-ja/py36-locale.git ./Doc/locale
cd ./Doc/locale
git checkout 6f42a6b453026fff60e18da1a1900a9db9593cfb

cd ..
sphinx-build -j4 -D language=ja -D locale_dirs=locale -D gettext_compact=0 -A versionswitcher=1 -E -d build/doctrees . ./_build/html
```

Then, open `_build/html/bugs.html` with your browser.

#### Error logs / results
![screen shot 2017-10-12 at 11 30 30](https://user-images.githubusercontent.com/223050/31476337-dd608716-af40-11e7-87b7-28bf3cdaa9b1.png)

#### Expected results
docutils transforms `"Title"` to `「Title」` (also `"Type"` to `「Type」`),
or do not transforms `"Title"` at all.

### Reproducible project / your project
- https://github.com/python-doc-ja/cpython-doc-intl
- https://github.com/python-doc-ja/py36-locale

### Environment info
- OS: macOS 10.13
- Python version: 3.6.2
- Sphinx version: 1.6.4

jfbu:
I can reproduce it with Docutils.

Test file ``index.rst``:

```rst
.. comment
   https://github.com/python-doc-ja/py36-locale/blob/master/ja/LC_MESSAGES/bugs.po#L122

.. comment https://github.com/sphinx-doc/sphinx/issues/4142

Welcome to FOO's documentation!
===============================

Hello \"Title\" World

バグ報告フォームには幾つかのフィールドがあります。\"Title\" フィールドには、問題の説明を *非常に*

```

Process with ``rst2html.py --language "ja" --smart-quotes "on" index.rst test.html``.

html file contains:

```
<p>Hello 「Title」 World</p>
[...]

PR #4112, opened 2017-10-04
Don't override the smart_quotes setting if it was already set

alex:
This is needed to fix: https://github.com/sphinx-contrib/spelling/issues/1

Rebase of #4110 onto the stable branch.

PR #4360, opened 2017-12-29
Do not apply smart quotes if Sphinx says no, even if Docutils says yes

### LINKED ISSUE #4359: A global ``.docutils`` smart_quotes setting overrides html_use_smartypants or extension setting
Consider situation [mentioned](https://github.com/sphinx-doc/sphinx/issues/3967#issuecomment-339408120) by @matthew-brett: user has a ``~/.docutils`` file which activates or deactivates smart_quotes.

Then this will override Sphinx's own setting either from (deprecated)  html_use_smartypants, or from an extension which would initiate a `smart_quotes` setting (cf ee52c7cd4f5e0a300c8fdd897c9529f5c633c7ed). This is bad because it is to be expected that a local setting overrides a global one.

Also, the setting in a `docutils.conf` file will override one set by an extension despite ee52c7cd.

Thus if we want to deactivate by default smart quotes for some builders (man, text, cf #4357), via a simple-minded usage of a `'smart_quotes'` setting as in ``environment/__init__.py`` this will fail is this issue is not fixed first.

One can test this by creating a ``$HOME/.doctutils`` file with

```
[parsers]
	smart_quotes=no
```

and use `html_use_smartypants = True` in `conf.py`. No smart quotes will be applied.

Or set ``smart_quotes=yes`` in Docutils user config file and `html_use_smartypants = False` in `conf.py` of local Sphinx project then smart quotes will be applied.

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.6.5, stable, master

Relates #4112

jfbu:
This closes #4359.

Relates #3967.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,10 @@ Features added
   :ref:`LaTeX 'sphinxsetup' <latexsphinxsetup>` key (refs: #4285)
 * Easier customizability of LaTeX macros involved in rendering of code-blocks
 * Show traceback if conf.py raises an exception (refs: #4369)
+* Add :confval:`smartquotes` to disable smart quotes through ``conf.py``
+  (refs: #3967)
+* Add :confval:`smartquotes_action` and :confval:`smartquotes_excludes`
+  (refs: #4142, #4357)
 
 Bugs fixed
 ----------
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -353,6 +353,63 @@ General configuration
 
    .. versionadded:: 1.3
 
+.. confval:: smartquotes
+
+   If true, the `Docutils Smart Quotes transform`__, originally based on
+   `SmartyPants`__ (limited to English) and currently applying to many
+   languages, will be used to convert quotes and dashes to typographically
+   correct entities.  Default: ``True``.
+
+   __ http://docutils.sourceforge.net/docs/user/smartquotes.html
+   __ https://daringfireball.net/projects/smartypants/
+
+   .. versionadded:: 1.6.6
+      It replaces deprecated :confval:`html_use_smartypants`.
+      It applies by default to all builders except ``man`` and ``text``
+      (see :confval:`smartquotes_excludes`.)
+
+   A `docutils.conf`__ file located in the configuration directory (or a
+   global :file:`~/.docutils` file) is obeyed unconditionally if it
+   *deactivates* smart quotes via the corresponding `Docutils option`__.  But
+   if it *activates* them, then :confval:`smartquotes` does prevail.
+
+   __ http://docutils.sourceforge.net/docs/user/config.html
+   __ http://docutils.sourceforge.net/docs/user/config.html#smart-quotes
+
+.. confval:: smartquotes_action
+
+   This string, for use with Docutils ``0.14`` or later, customizes the Smart
+   Quotes transform.  See the file :file:`smartquotes.py` at the `Docutils
+   repository`__ for details.  The default ``'qDe'`` educates normal **q**\
+   uote characters ``"``, ``'``, em- and en-**D**\ ashes ``---``, ``--``, and
+   **e**\ llipses ``...``.
+
+   .. versionadded:: 1.6.6
+
+   __ https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/
+
+.. confval:: smartquotes_excludes
+
+   This is a ``dict`` whose default is::
+
+     {'languages': ['ja'], 'builders': ['man', 'text']}
+
+   Each entry gives a sufficient condition to ignore the
+   :confval:`smartquotes` setting and deactivate the Smart Quotes transform.
+   Accepted keys are as above ``'builders'`` or ``'languages'``.
+   The values are lists.
+
+   .. note:: Currently, in case of invocation of :program:`make` with multiple
+      targets, the first target name is the only one which is tested against
+      the ``'builders'`` entry and it decides for all.  Also, a ``make text``
+      following ``make html`` needs to be issued in the form ``make text
+      O="-E"`` to force re-parsing of source files, as the cached ones are
+      already transformed.  On the other hand the issue does not arise with
+      direct usage of :program:`sphinx-build` as it caches
+      (in its default usage) the parsed source files in per builder locations.
+
+   .. versionadded:: 1.6.6
+
 .. confval:: tls_verify
 
    If true, Sphinx verifies server certifications.  Default is ``True``.
@@ -784,15 +841,11 @@ that use Sphinx's HTMLWriter class.
 
 .. confval:: html_use_smartypants
 
-   If true, `SmartyPants <https://daringfireball.net/projects/smartypants/>`_
-   will be used to convert quotes and dashes to typographically correct
+   If true, quotes and dashes are converted to typographically correct
    entities.  Default: ``True``.
 
    .. deprecated:: 1.6
-      To disable or customize smart quotes, use the Docutils configuration file
-      (``docutils.conf``) instead to set there its `smart_quotes option`_.
-
-      .. _`smart_quotes option`: http://docutils.sourceforge.net/docs/user/config.html#smart-quotes
+      To disable smart quotes, use rather :confval:`smartquotes`.
 
 .. confval:: html_add_permalinks
 
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -20,8 +20,9 @@ import warnings
 from os import path
 from copy import copy
 from collections import defaultdict
+from contextlib import contextmanager
 
-from six import BytesIO, itervalues, class_types, next
+from six import BytesIO, itervalues, class_types, next, iteritems
 from six.moves import cPickle as pickle
 
 from docutils.io import NullOutput
@@ -46,7 +47,7 @@ from sphinx.util.parallel import ParallelTasks, parallel_available, make_chunks
 from sphinx.util.websupport import is_commentable
 from sphinx.errors import SphinxError, ExtensionError
 from sphinx.locale import __
-from sphinx.transforms import SphinxTransformer
+from sphinx.transforms import SphinxTransformer, SphinxSmartQuotes
 from sphinx.versioning import add_uids, merge_doctrees
 from sphinx.deprecation import RemovedInSphinx17Warning, RemovedInSphinx20Warning
 from sphinx.environment.adapters.indexentries import IndexEntries
@@ -54,7 +55,7 @@ from sphinx.environment.adapters.toctree import TocTree
 
 if False:
     # For type annotation
-    from typing import Any, Callable, Dict, IO, Iterator, List, Pattern, Set, Tuple, Type, Union  # NOQA
+    from typing import Any, Callable, Dict, IO, Iterator, List, Pattern, Set, Tuple, Type, Union, Generator  # NOQA
     from docutils import nodes  # NOQA
     from sphinx.application import Sphinx  # NOQA
     from sphinx.builders import Builder  # NOQA
@@ -89,6 +90,22 @@ versioning_conditions = {
     'text': is_translatable,
     'commentable': is_commentable,
 }  # type: Dict[unicode, Union[bool, Callable]]
+
+
+@contextmanager
+def sphinx_smartquotes_action(env):
+    # type: (BuildEnvironment) -> Generator
+    if not hasattr(SphinxSmartQuotes, 'smartquotes_action'):
+        # less than docutils-0.14
+        yield
+    else:
+        # docutils-0.14 or above
+        try:
+            original = SphinxSmartQuotes.smartquotes_action
+            SphinxSmartQuotes.smartquotes_action = env.config.smartquotes_action
+            yield
+        finally:
+            SphinxSmartQuotes.smartquotes_action = original
 
 
 class NoUri(Exception):
@@ -600,7 +617,8 @@ class BuildEnvironment(object):
             # remove all inventory entries for that file
             app.emit('env-purge-doc', self, docname)
             self.clear_doc(docname)
-            self.read_doc(docname, app)
+            with sphinx_smartquotes_action(self):
+                self.read_doc(docname, app)
 
     def _read_parallel(self, docnames, app, nproc):
         # type: (List[unicode], Sphinx, int) -> None
@@ -612,8 +630,9 @@ class BuildEnvironment(object):
         def read_process(docs):
             # type: (List[unicode]) -> unicode
             self.app = app
-            for docname in docs:
-                self.read_doc(docname, app)
+            with sphinx_smartquotes_action(self):
+                for docname in docs:
+                    self.read_doc(docname, app)
             # allow pickling self to send it back
             return BuildEnvironment.dumps(self)
 
@@ -677,14 +696,25 @@ class BuildEnvironment(object):
         language = self.config.language or 'en'
         self.settings['language_code'] = language
         if 'smart_quotes' not in self.settings:
-            self.settings['smart_quotes'] = True
+            self.settings['smart_quotes'] = self.config.smartquotes
             if self.config.html_use_smartypants is not None:
                 warnings.warn("html_use_smartypants option is deprecated. Smart "
                               "quotes are on by default; if you want to disable "
-                              "or customize them, use the smart_quotes option in "
-                              "docutils.conf.",
+                              "them, use the smartquotes option.",
                               RemovedInSphinx17Warning)
                 self.settings['smart_quotes'] = self.config.html_use_smartypants
+
+            # some conditions exclude smart quotes, overriding smart_quotes
+            for valname, vallist in iteritems(self.config.smartquotes_excludes):
+                if valname == 'builders':
+                    # this will work only for checking first build target
+                    if self.app.builder.name in vallist:
+                        self.settings['smart_quotes'] = False
+                        break
+                elif valname == 'languages':
+                    if self.config.language in vallist:
+                        self.settings['smart_quotes'] = False
+                        break
 
         # confirm selected language supports smart_quotes or not
         for tag in normalize_language_tag(language):
--- a/sphinx/io.py
+++ b/sphinx/io.py
@@ -18,7 +18,7 @@ from sphinx.transforms import (
     ApplySourceWorkaround, ExtraTranslatableNodes, CitationReferences,
     DefaultSubstitutions, MoveModuleTargets, HandleCodeBlocks, SortIds,
     AutoNumbering, AutoIndexUpgrader, FilterSystemMessages,
-    UnreferencedFootnotesDetector
+    UnreferencedFootnotesDetector, SphinxSmartQuotes
 )
 from sphinx.transforms.compact_bullet_list import RefOnlyBulletListTransform
 from sphinx.transforms.i18n import (
@@ -97,6 +97,16 @@ class SphinxStandaloneReader(SphinxBaseReader):
                   HandleCodeBlocks, AutoNumbering, AutoIndexUpgrader, SortIds,
                   RemoveTranslatableInline, PreserveTranslatableMessages, FilterSystemMessages,
                   RefOnlyBulletListTransform, UnreferencedFootnotesDetector]
+
+    def __init__(self, app, parsers={}, *args, **kwargs):
+        SphinxBaseReader.__init__(self, app, parsers, *args, **kwargs)
+        self.smart_quotes = app.env.settings['smart_quotes']
+
+    def get_transforms(self):
+        transforms = SphinxBaseReader.get_transforms(self)
+        if self.smart_quotes:
+            transforms.append(SphinxSmartQuotes)
+        return transforms
 
 
 class SphinxI18nReader(SphinxBaseReader):
--- a/sphinx/parsers.py
+++ b/sphinx/parsers.py
@@ -13,8 +13,6 @@ import docutils.parsers
 import docutils.parsers.rst
 from docutils.transforms.universal import SmartQuotes
 
-from sphinx.transforms import SphinxSmartQuotes
-
 if False:
     # For type annotation
     from typing import Any, Dict, List, Type  # NOQA
@@ -60,10 +58,11 @@ class RSTParser(docutils.parsers.rst.Parser):
 
     def get_transforms(self):
         # type: () -> List[Type[Transform]]
-        """Sphinx's reST parser replaces a transform class for smart-quotes by own's"""
+        """Sphinx's reST parser replaces a transform class for smart-quotes by own's
+
+        refs: sphinx.io.SphinxStandaloneReader"""
         transforms = docutils.parsers.rst.Parser.get_transforms(self)
         transforms.remove(SmartQuotes)
-        transforms.append(SphinxSmartQuotes)
         return transforms
 
 
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -134,6 +134,11 @@ class Config(object):
 
         tls_verify = (True, 'env'),
         tls_cacerts = (None, 'env'),
+        smartquotes = (True, 'env'),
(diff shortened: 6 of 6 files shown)


sphinx/io.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, app, parsers={}, *args, **kwargs):
             parser = parser_class()
             if hasattr(parser, 'set_application'):
                 parser.set_application(app)
+            if hasattr(parser, 'set_smartquotes_allowed'):
+                parser.set_smartquotes_allowed(app.env.settings['smart_quotes'])
```
jfbu: as far as I understand, this will be called from a BuildEnvironment object ``read_doc()`` method, when a ``reader`` is being set up, and at that time the ``app.env.settings['smart_quotes']`` always exists with value either True or False due to earlier code executed by the BuildEnvironment instance.

Now, perhaps the accessor method is superfluous here and we should set the Boolean directly rather than having such an instance method added to RSTParser class.

jfbu:
@tk0miya perhaps we can reconsider now the 79df05b36 which reverted bfd39c12b ? I forgot the reason why bfd39c12b was reverted. With this PR, a Sphinx `smart_quotes` conf.py setting will be obeyed and override whatever Docutils global config file says. It will also override local ``docutils.conf`` so we should modify our docs.

edit1: "override" is not correct; it applies only in case the `smart_quotes` is activated via ``docutils.conf``, then if Sphinx decides to turn it off, it will work. But it `smart_quotes` is de-activated (explicitely) via ``docutils.conf``, then no matter what Sphinx config says, it will remain de-activated. In brief, de-activation whether coming from Docutils or Sphinx side always wins.

edit2: I have ressurected bfd39c12b in modified form in 2nd commit of this PR.

jfbu:
I should clarify that initial commit of this PR does only what the title says: if Sphinx setting disallows smart quotes, then they are not applied and Docutils setting is ignored. But if Sphinx setting does allow smart quotes, then they still can be unactivated if Docutils configuration files ``docutils.conf`` or ``~/.docutils`` says so.

sphinx/parsers.py
```diff
@@ -58,12 +58,21 @@ def set_application(self, app):
 class RSTParser(docutils.parsers.rst.Parser):
     """A reST parser customized for Sphinx."""
 
+    smartquotes_allowed = True
+
+    def set_smartquotes_action(self, action):
```
tk0miya: Let's make `RSTParser` class a subclass of `sphinx.parsers.Parser`. After that we can refer config object by `self.config`.
I think storing status to class variable is not good way. Because it behaves like global variables.

doc/config.rst
```diff
@@ -353,6 +353,39 @@ General configuration
 
    .. versionadded:: 1.3
 
+.. confval:: smart_quotes
```
tk0miya: I agree the basic concept of this PR. But we would need to be able to disable smart_quotes by several condition.
For example, #4142 tells us to disable it for Japanese document. I saw somebody wants to disable it for arbitrary builder in some issues.
So I think this should not be boolean. (but I don't have good alternative idea...)

tk0miya:
>perhaps we can reconsider now the 79df05b which reverted bfd39c1 ?

The reason of the revert is only it is not intended commit. It was not my completed work as I commented above. I think simple boolean setting is not good for this case. So we have to consider the way to configure smart_quotes.

It's okay to introduce a configuration for smart_quotes itself to me.

tk0miya:
I'd forggeten to mention about parsers. I think smart quotes is not only for reST files. It is a setting for representation, so it should effect to other formats.
If user want to do that, it is better if configurable.

sphinx/parsers.py
jfbu: agreed, done (will comment after push)

sphinx/parsers.py
```diff
@@ -55,15 +55,23 @@ def set_application(self, app):
...
 
-class RSTParser(docutils.parsers.rst.Parser):
+class RSTParser(docutils.parsers.rst.Parser, Parser):
     """A reST parser customized for Sphinx."""
 
+    def set_application(self, app):
+        # type: (Sphinx) -> None
+        Parser.set_application(self, app)
+        if hasattr(SphinxSmartQuotes, 'smartquotes_action'):
+            SphinxSmartQuotes.smartquotes_action = self.env.settings['smartquotes_action']
```
jfbu: I have a problem here which is that afaik SphinxSmartQuotes is a class, not an instance of that class, and in get_transforms below it is the class object which is added to list of transforms. Hence I needed to modify the class attribute here. Don't know if this can have undesired consequences.

sphinx/environment/__init__.py
```diff
@@ -692,6 +691,7 @@ def read_doc(self, docname, app=None):
                 break
         else:
             self.settings['smart_quotes'] = False
+        self.settings['smartquotes_action'] = self.config.smartquotes_action
```
jfbu: Docutils has no smartquotes_action user config setting yet; but it introduced the docutils.transforms.universal.SmartQuotes class attribute at 0.14, and it is useful for us, because Sphinx users may want to for example not apply smart quotes to ellipsis conversion (from  ``...``). For Docutils < 0.14, there was at some point some bad config which educated backticks also, and we override that in our own sphinx.util.smartypants (this module is used by Sphinx only for Docutils < 0.14).

jfbu:
> I'd forggeten to mention about parsers. I think smart quotes is not only for reST files. It is a setting for representation, so it should effect to other formats.
If user want to do that, it is better if configurable.

My knowledge of how Sphinx interacts with Docutils remains quite limited. At 69cf328b you created a class ``RSTParser`` apparently especially for facilitating customization at Sphinx level of Docutils Smart Quotes transform. Do you mean here that this code should actually be in Sphinx Parser class, not RSTParser?

jfbu:
>  I think simple boolean setting is not good for this case. So we have to consider the way to configure smart_quotes.

@tk0miya I have added confval setting ``smartquotes_excludes``. It has limitations particularly for builders, as multiple builders can share same pickled environment, so user has to explicitely request re-parsing of source files when doing html then text builds for example; or when doing `make html text` then smart quotes will be applied to text target too, despite setting of ``smartquotes_excludes``. Best I can do for the moment due to shared BuildEnvironment.

I have not addressed the issue about extending PR to other than RSTParser, because my knowledge of Sphinx/Docutils workflow is currently too limited.

doc/config.rst
jfbu: I have addressed that by supplementary `smartquotes_excludes` dict configuration setting at 88a6e2451 (fixed typo in doc at 4b3a09133)

doc/config.rst
```diff
@@ -386,6 +387,22 @@ General configuration
...
+
+     {'language': ['ja'], 'builder': ['man', 'text']}
+
+   It gives sufficient condition to ignore the :confval:`smart_quotes` setting
+   and deactivate the Smart Quotes transform.  The keys must be ``'builder'``
+   or names of configuration settings.  The values are lists.  Please note
+   that in case of invocation of ``make`` with multiple builder names, the
+   first one counts.  Also, when using ``sphinx-build`` in succession for
+   different builders, one needs to use ``-E`` option (e.g. in case of an
+   ``html`` build followed by a ``man`` build, or conversely.)
```
jfbu: ah sorry, I am not used to ``sphinx-build`` but as it creates by default a ``.doctrees`` individual to each target, there is no need for ``-E``. The problem arises when using the "make-mode" only. Will fix docs.

sphinx/environment/__init__.py
```diff
@@ -677,21 +677,37 @@ def read_doc(self, docname, app=None):
         language = self.config.language or 'en'
         self.settings['language_code'] = language
         if 'smart_quotes' not in self.settings:
-            self.settings['smart_quotes'] = True
+            self.settings['smart_quotes'] = self.config.smart_quotes
```
jfbu: now that Sphinx core has a ``smart_quotes`` config setting, possibly the logic for

    if 'smart_quotes' not in self.settings:

(see #4112) is gone? can an extension override a config setting rather than add it to environment as done at  https://github.com/sphinx-contrib/spelling/commit/6209fa8d2c16182b30053f40770dc81b11c77f00 ?

jfbu:
Currently the sphinx-quickstart created Makefile (whether make-mode enabled or not) configures builds to share `doctrees/` across build targets. On the other hand the default behaviour of `sphinx-build` does not do that. I see the speed gain into sharing the cached files, but the smart quotes issue shows that there is a problem when the Docutils parsing should behave differently depending on builder. It is a bit of thorny issue because indeed it will be disconcerting to user if ``make html epub latex man`` re-parses source files (which takes time on big docs) for each target. We can not escape that by implementing "man will always force re-parsing", because then ``make man html`` will inhibit smart quotes in html output. However we could set "man" and "text" builders to force re-parsing *and* store the parsed files in special location. I.e. there would be ``doctrees/`` and ``doctrees-nosmartquotes/``.

Then the builders listed in ``smartquotes_excludes['builder']`` would use ``doctrees-nosmartquotes`` and the others ``doctrees/``. This looks a bit ugly for only one problem and not very scalable in (improbable) case some other Docutils implemented transform than Smart Quotes arises in future with similar issues.

doc/config.rst
```diff
@@ -353,6 +353,57 @@ General configuration
...
+
+   .. versionadded:: 1.6.6
+
+   __ https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/
+
+.. confval:: smartquotes_excludes
+
+   This is a ``dict`` whose default is::
+
+     {'language': ['ja'], 'builder': ['man', 'text']}
```
tk0miya: I think it is better to rename keys plural form.

sphinx/environment/__init__.py
tk0miya: I think it is not a public interface, but it is needed for workaround. So please keep this as is.

Note: It might be refactored in future. Personally, I feel `read_doc` is not a part of responsibility of "BuildEnvironment".

sphinx/environment/__init__.py
```diff
@@ -677,21 +677,37 @@ def read_doc(self, docname, app=None):
...
+                              "them, use the smart_quotes option",
                               RemovedInSphinx17Warning)
                 self.settings['smart_quotes'] = self.config.html_use_smartypants
 
+            # some conditions exclude smart quotes, overriding smart_quotes
+            for valname, vallist in iteritems(self.config.smartquotes_excludes):
+                if valname == 'builder':
+                    # this will work only for checking first build target
+                    if self.app.builder.name in vallist:
+                        self.settings['smart_quotes'] = False
```
tk0miya: Ideally, smart quotes transform should be invoked after the pickling (write phase) to support each builders.
I know it is a bit big refactoring. so +1 this for stable.

sphinx/parsers.py
tk0miya: Right, class attribute is a kind of global variable. So modifying it is not good way. So it would be better to update it temporarily like following:
```
from contextlib import contextmanager

@contextmanager
def sphinx_smartquotes_action(env):
    try:
        original = SphinxSmartQuotes.smartquotes_action
        SphinxSmartQuotes.smart_quotes_action = env.config.smartquotes_action
        yield
     finally:
         SphinxSmartQuotes.smart_quotes_action = original

# in env.doc_read() method:
with sphinx_smartquotes_action(self):
    doctree = read_doc(...)
```

sphinx/environment/__init__.py
tk0miya: Is there any reason for setting this into `self.settings`?

tk0miya:
>Do you mean here that this code should actually be in Sphinx Parser class, not RSTParser?

docutils picks up transform modules by `get_components()` of related components. So I think `SphinxStandaloneReader` is better.

sphinx/environment/__init__.py
jfbu: It was in anticipation that Docutils might offer such an option in future to configure the `smartquotes_action` attribute of its docutils.transforms.universal.SmartQuotes, which we inherit for SphinxSmartQuotes. Also, I did this before RSTParser inherited from Parser hence got access to app. Thinking twice, it might indeed be better not to set this into ``self.settings``, else we will again have problem that a Docutils config file would override the Sphinx user conf.py, if such an option is introduced in future by Docutils.

sphinx/parsers.py
jfbu: Thanks, will do.

sphinx/environment/__init__.py
jfbu: ok, I will leave it. I find "BuildEnvironment" complicated...

doc/config.rst
jfbu: I was using singular for easing up re-use of the key as ``getattr(self.config, key)``, but I agree it is more self-explanatory to use plurals. In theory any config key is usable here to condition usage of Smart Quotes, but in practice I guess I can just drop the final "s". Or we allow only conditioning on language and builder? Also, I will explain better in docs that the logic is simply "OR".

sphinx/environment/__init__.py
tk0miya: If docutils does not implement the option yet, it might be too premature action. -1 for adding key.

sphinx/environment/__init__.py
tk0miya: Absolutely, `BuildEnvironment` has multiple responsibility. So I'd like to split it into some modules in future.

jfbu:
@tk0miya I have implemented your advices. I checked that SmartQuotes then applies also to Markdown files as added following [the docs](http://www.sphinx-doc.org/en/stable/markdown.html).

> If docutils does not implement the option yet, it might be too premature action. -1 for adding key.

I removed it ;-) (I mean, I do not add it to `env.settings` but I keep it as a Sphinx option, as it is useful)

sphinx/environment/__init__.py
jfbu: I have removed it from ``self.settings``, but I do keep it as confval because it is useful for Sphinx users. Is that ok?

doc/config.rst
tk0miya: I can't think of the usecase for any config key (without `language`). So I feel it's okay to restrict the key to `languages` and `builders`.

>Also, I will explain better in docs that the logic is simply "OR".

+1

sphinx/parsers.py
```diff
@@ -55,15 +53,16 @@ def set_application(self, app):
         self.info = app.info
 
 
-class RSTParser(docutils.parsers.rst.Parser):
+class RSTParser(docutils.parsers.rst.Parser, Parser):
```
jfbu: now inheriting from Parser is not needed, because we don't need access to `app`. Should I delete that added inheritance ?

sphinx/io.py
```diff
@@ -98,6 +98,11 @@ class SphinxStandaloneReader(SphinxBaseReader):
                   RemoveTranslatableInline, PreserveTranslatableMessages, FilterSystemMessages,
                   RefOnlyBulletListTransform, UnreferencedFootnotesDetector]
 
+    def __init__(self, app, parsers={}, *args, **kwargs):
+        SphinxBaseReader.__init__(self, app, parsers, *args, **kwargs)
+        if app.env.settings['smart_quotes']:
+            self.transforms.append(SphinxSmartQuotes)
```
jfbu: with this, also ``.md`` source files (for example) are in the scope of Smart Quotes, which was not the case before.

sphinx/environment/__init__.py
```diff
@@ -677,15 +692,37 @@ def read_doc(self, docname, app=None):
...
+                    try:
+                        attr = getattr(self.config, valname[:-1])
+                        if attr in vallist:
+                            self.settings['smart_quotes'] = False
+                            break
+                    except AttributeError:
+                        warnings.warn("'%s' (from '%s' in smartquotes_excludes) "
+                                      "isn't a valid configuration variable. "
+                                      "Did you forget an ending 's'?"
+                                      % (valname[:-1], valname))
```
jfbu: perhaps the whole ``else`` clause could be dropped if we limit to conditioning on language and builder

doc/config.rst
```diff
@@ -353,6 +353,63 @@ General configuration
...
+   settings.  The values are lists.
+
+   .. note:: Currently, in case of invocation of :program:`make` with multiple
+      targets, the first target name is the only one which is tested against
+      the ``'builders'`` entry and it decides for all.  Also, a ``make text``
+      following ``make html`` needs to be issued in the form ``make text
+      O="-E"`` to force re-parsing of source files, as the cached ones are
+      already transformed.  On the other hand the issue does not arise with
+      direct usage of :program:`sphinx-build` as it caches
+      (in its default usage) the parsed source files in per builder locations.
```
jfbu, line 409: should we document like this our defects ;-) or leave it to user to discover ;-) ?

sphinx/environment/__init__.py
```diff
@@ -91,6 +92,18 @@
 }  # type: Dict[unicode, Union[bool, Callable]]
 
 
+@contextmanager
+def sphinx_smartquotes_action(env):
+    try:
+        original = SphinxSmartQuotes.smartquotes_action
+        setattr(SphinxSmartQuotes, 'smartquotes_action',
```
tk0miya: Is there any reason to using `setattr`? An assignment is much simpler to me.

sphinx/io.py
tk0miya: Please override `get_transforms()` and add `SphinxSmartQuotes` conditionally. because `self.transforms` is a class variable. So `SphinxSmartQuotes` will be added on each instantiation.
```
def __init__(...):
    ...
    self.smart_quotes = app.env.settings['smart_quotes']

def get_transforms(self):
    transforms = SphinxBaseReder.get_transforms()
    if self.smart_quotes:
        transforms.append(SphinxSmartQuotes)

    return transforms
```

sphinx/parsers.py
tk0miya: `Parser` super class is no longer needed.

tk0miya (changes requested):
LGTM with nits

sphinx/io.py
jfbu: sorry for my basic mistake. Indeed build time considerably increased due to it. I will fix.

doc/config.rst
tk0miya: This is internally and hard to understand behavior. So it should be documented. Let's go with as is.

sphinx/io.py
tk0miya: :-)

sphinx/environment/__init__.py
tk0miya: As I commented above, I can't imagine the use case for this else block.
I don't opposite it, but I'd like to know an example use case before merging.

sphinx/environment/__init__.py
jfbu: I used an assignment but it did not have expected effect in my testing. Was it because SphinxSmartQuotes is imported separately in io.py?

jfbu: hmm... works perfectly well with an assignment. Don't know what went wrong (but I am doing various things at the same time, bad habit...). Ok, will push fix.

sphinx/parsers.py
jfbu: done

sphinx/environment/__init__.py
```diff
@@ -96,10 +96,9 @@
 def sphinx_smartquotes_action(env):
     try:
```
jfbu: the try is needed because Docutils SmartQuotes does not have the ``smartquotes_action`` attribute at Docutils < 0.14.

sphinx/parsers.py
```diff
@@ -53,7 +53,7 @@ def set_application(self, app):
         self.info = app.info
 
 
-class RSTParser(docutils.parsers.rst.Parser, Parser):
+class RSTParser(docutils.parsers.rst.Parser):
```
jfbu: The Parser inheritance was added in an earlier commit of this PR (#4360), but is now not needed

sphinx/environment/__init__.py
jfbu: sorry I overlooked your comment, I will remove that block as I don't see obvious use case either

sphinx/environment/__init__.py
```diff
@@ -91,6 +92,17 @@
 }  # type: Dict[unicode, Union[bool, Callable]]
 
 
+@contextmanager
+def sphinx_smartquotes_action(env):
+    try:
+        original = SphinxSmartQuotes.smartquotes_action
+        SphinxSmartQuotes.smartquotes_action = env.config.smartquotes_action
+        yield
+        SphinxSmartQuotes.smartquotes_action = original
```
tk0miya: This is not executed when the error has happened at yield block. Usually, it does not effect to the processing because `sphinx-build` crashes on the error. But the behavior might be changed in future. So it should be write back correctly using `final` clause.
How about this implementation?
```
if not hasattr(SphinxSmartQuotes, 'smartquotes_action'):
    # less than docutils-0.14
    yield
else:
    # docutils-0.14 or above
    try:
        original = SphinxSmartQuotes.smartquotes_action
        SphinxSmartQuotes.smartquotes_action = env.config.smartquotes_action
    finally:
        SphinxSmartQuotes.smartquotes_action = original
```

sphinx/environment/__init__.py
```diff
@@ -91,6 +92,17 @@
 }  # type: Dict[unicode, Union[bool, Callable]]
 
 
+@contextmanager
+def sphinx_smartquotes_action(env):
```
tk0miya, line 96: Please add type annotation: `# type: (BuildEnvironment) -> Generator`.

tk0miya (changes requested):

sphinx/environment/__init__.py
tk0miya: +1

sphinx/environment/__init__.py
jfbu: +1, I guess I add a ``yield`` in the ``try`` ;-)

jfbu:
closing and reopening to trigger again testing

tk0miya (approved):
Great work!

CHANGES
```diff
@@ -17,6 +17,10 @@ Features added
 * ``VerbatimHighlightColor`` is a new
   :ref:`LaTeX 'sphinxsetup' <latexsphinxsetup>` key (refs: #4285)
 * Easier customizability of LaTeX macros involved in rendering of code-blocks
+* Add :confval:`smart_quotes` to disable smart quotes through ``conf.py``
```
tk0miya: I noticed `smartquotes_disables` does not mentioned here.

Anyway, these settings for smart quotes are named `smart_quotes` or `smartquotes`. I feel they should have same prefix. What do you think?

jfbu: `smartquotes_excludes` is mentioned at next line ...

I agree about using same prefix `smartquotes` for all. Docutils has a `smart_quotes` or `smart-quotes` option and this why I originally used `smart_quotes`. But Docutils has the `smartquotes_action` class attribute hence I also followed. But I feel `smartquotes` prefix for all conf.py variables is much better indeed.

tk0miya: oh, sorry > `smartquotes_excludes`

tk0miya (approved):
LGTM!

tk0miya:
ping @jfbu 

Do you still have any remaining work for this PR? If none, could you merge this please?
Then I will release 1.6.6 in this weekend.

If you have tasks, I can wait it. So please let me know.

Thanks,

jfbu:
@tk0miya Thanks for very thorough reviewing and expert Python guidance! To make for cleaner commit history and due to the fact the PR was importantly refactored mid-way I will rebase and squash commits and then I will merge. ;-)

(I am preserving for some time original history [there](https://github.com/jfbu/sphinx/tree/4360_before_squash))

jfbu:
I will merge once testing completes :tada:

jfbu:
Merging because Travis fails due to connect timeout errors.

tk0miya:
🎉  Great work!

Issue #3221, opened 2016-12-10
Deprecate epub2 builder


tk0miya:
Since 1.5, we've start to use epub3 builder by default for `make epub` command.
AFAIK, EPUB3 is upper compatible with EPUB2. So there is no reason to keep providing the epub2 builder.
To simplify our package, it would be nice if the builder would be removed in future version.
How about doing it in 1.6?

shimizukawa:
+1 for removing.
I think we should emit DprecationWarning to remove the feature.
I wrote a deprecation policy document. Please read and give me your opinion.
http://www.sphinx-doc.org/en/1.5/devguide.html?highlight=deprecation#deprecation-policy

tk0miya:
Note:
The epub3 builder is a derived class of epub2 builder.
Before deprecation, we have to split the implementation.

PR #4363, opened 2017-12-30
Full and enhanced implementation of the epub2 builder

tantale:
### Feature or Bugfix

- Bugfix

### Purpose

The purpose of this pull request is to do a full and enhanced implementation
of the epub2 builder: implement missing features, fix some mistakes
and add a small and backward-compatible improvement.
The aim is to have a full-featured and professional epub2 builder.

> **notice**
>
> Even if the 'epub2' builder is considered deprecated (see #3221),
> this format is still widely used, notably by professional publishers
> (like [Lulu.com publishing platform](http://www.lulu.com)).
> Also notice that 'epub3' is **not** a superset of 'epub2' because
> HTML5 is **not** a superset of XHTML, even if this two languages are similar.

This PR contains the following improvements:

* add missing implementation for `<navMap>` in `toc.ncx`;
* fix the `EpubBuilder.build_navpoints` method that contains a ``raise`` statement with no arguments;
* remove the ``role`` attribute from the basic theme HTML templates when the DOCTYPE is not HTML5;
* remove a trailing newline after the DOCTYPE declaration in template;
* remove leading newline before ``block`` and ``macro`` in template;
* add the XML declaration before the XHTML DOCTYPE;
* replace the old format "%(uid)s" by the Jinja format "{{ uid }}" in `content.opf`;
* change `content.opf` to use the unique-identifier instead of the uuid;
* improve the epub2 conversion to add the metadata: creation, modification and publication dates.

This PR also add an unit test for 'epub2'.

> **notice**
>
> Some fixes and enhancements could also be applied to the 'epub3' builder,
> but I decided not to do it in order to keep the PR understandable.

### Detail

Missing implementation for `<navMap>` in `toc.ncx`
--------------------------------------------------

Fix in 'epub2' template: add missing implementation for `<navMap>` in `toc.ncx` (copied from 'epub3' template).

`raise` statement with no arguments
-----------------------------------

Fix the `EpubBuilder.build_navpoints()` method that contains a `raise` statement with no arguments:
use `ValueError` instead (to have an explicit message). Notice: if no expressions are present,
`raise` [re-raises](https://docs.python.org/3/tutorial/errors.html#raising-exceptions)
the last exception that was active in the current scope -- which is wrong here.

The change is done as follow:

```python
if node['level'] == level:
    [...]
else:
    raise ValueError('Invalid node level: got %d, expected %d or lower' % (node['level'], level + 1))
```

> **notice**
>
> The same error exists in `Epub3Builder.build_navlist()` method,
> but I decided not to correct it in order to keep the PR understandable.

Remove the `role` attribute
---------------------------

The following templates are modified to remove the `role` attribute when HTML5 is not used:

- layout.html
- searchbox.html
- sourcelink.html

A Jinja condition is added, for instance:

```html
<div class="body" {%- if html5_doctype %} role="main"{%- endif %}>
    {% block body %} {% endblock %}
</div>
```

> **notice**
>
> In order the keep this PR simple, the fix is applied only for the "basic" templates.
> Other templates may need the same correction.

Remove a trailing newline after the DOCTYPE declaration
-------------------------------------------------------

Correct the basic theme "layout.html" to remove a trailing newline after the DOCTYPE
declaration (which can cause validation problem with third party validators).

This newline was added in the Jinja comment:

```
{# XXX necessary? #}
```

Remove leading newline before `block` and `macro`
-------------------------------------------------

Correct the basic theme template to avoid a leading newline before `block` and `macro`,
because XHTML files must starts with a tag (XML declaration or DOCTYPE or root element), not a newline.

Add the XML declaration before the XHTML DOCTYPE
------------------------------------------------

Fix the 'epub2' builder to add the XML declaration before the XHTML DOCTYPE,
because the XML declaration is required for some third party epub2 validators.

Quoting the [W3C documentation](https://www.w3.org/TR/xhtml1/#strict):

> An XML declaration is not required in all XML documents; however XHTML
> document authors are strongly encouraged to use XML declarations in all
> their documents.

Code change:

```python
DOCTYPE = '''<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN"
  "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">'''
```

Replace the old format "%(uid)s" by the Jinja format
----------------------------------------------------

Fix in 'epub2' template: replace the old format "%(uid)s" by the Jinja format "{{ uid }}" in `content.opf`.

Code change:

```xml
<package xmlns="http://www.idpf.org/2007/opf" version="2.0"
      unique-identifier="{{ uid }}">
[...]
</package>
```

Change `content.opf` to use the unique-identifier instead of the uuid
---------------------------------------------------------------------

Fix in 'epub2' template and builder: the `meta@dtb:uid` content
should conform to unique-identifier in `content.opf`: "{{ id }}".

If the `content.opf` file contains:

```xml
<package xmlns="http://www.idpf.org/2007/opf" version="2.0" unique-identifier="BookId">
    <metadata xmlns:opf="http://www.idpf.org/2007/opf" xmlns:dc="http://purl.org/dc/elements/1.1/">
        <dc:identifier id="BookId" opf:scheme="ISBN">9780244627768</dc:identifier>
        [...]
    </metadata>
    [...]
</package>
```

Then, the `toc.ncx` file should contain:

```xml
<ncx xmlns="http://www.daisy.org/z3986/2005/ncx/" version="2005-1">
  <head>
    <meta name="dtb:uid" content="9780244627768"/>
    [...]
  </head>
  [...]
</ncx>
```

> **notice**
>
> EpubCheck v3.0.1 warns about this if the value is different.
> Furthermore, some professional publishers could refuse your epub if this condition is not satisfied.

New metadata: creation, modification and publication dates
----------------------------------------------------------

Improve the epub2 conversion to add the metadata: creation, modification and publication dates.

*Note: this feature is a small improvement.*

This improvement follow the recommendation of the Open Packaging Format about
[dates](http://www.idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.2.7).
Furthermore, these additional attributes are required by some professional publishers.

The single `<dc:date>` is replaced by:

```xml
<dc:date opf:event="creation">{{ date_creation }}</dc:date>
<dc:date opf:event="modification">{{ date_modification }}</dc:date>
<dc:date opf:event="publication">{{ date_publication }}</dc:date>
```

To do that, we need to add 3 new configuration values:

```python
app.add_config_value('epub_date_creation', '', 'html')
app.add_config_value('epub_date_modification', '', 'html')
app.add_config_value('epub_date_publication', '', 'html')
```

The format must follow the ISO 8601 specification for XML document:
[Date and Time Formats](https://www.w3.org/TR/NOTE-datetime>)
(i.e.: "YYYY[-MM[-DD]]"). Each date defaults to the current date.

Resources
---------

- Epub 2 specification in the [International Digital Publishing Forum](http://idpf.org/epub/201).
- Test results can be validated using [EpubCheck](https://github.com/IDPF/epubcheck).
  The epub files were checked with the version 3.0.1 (used by professional publishers).
- Some epub examples are available in the [PressBooks web site](https://pressbooks.com/sample-books/).
  These examples use epub2 format.


tantale:
I notice that the message history is in a mess, there are duplicates. I'll try to rewrite it…

tantale:
OK, the git history is cleaned (I rebased, removed duplicates and force pushed to my branch).

tk0miya:
Thank you for contribution, but unfortunately we've determined to drop supporting epub2 builder.
I understand epub2 is still used, but we don't have enough resource to maintain builders more.
How about forking it as an extension? It would be nice :-)

tantale:
Hi Takeshi, pleased to meet you, and happy new year 🎉.

This PR is mainly a bug fix for the 1.6.6 version; it adds a small improvement (dates). All I want to do is to fix the 1.6.5 version, not to add new features. Your decision to drop the epub2 builder is understandable and I want to respect that. For this reason, I forked the "stable" branch, not the "master".

Moreover, I am preparing a PR to fix/improve the epub3 builder. But currently the epub3 format is not widely supported by professionals. So, it is worthwhile to work on epub2 until epub3 become a real standard. The work done for epub2 could be easily profitable for epub3.

I can offer you to remove the improvement about dates. That way, there is no new feature.

I have thought of developing a epub2 extension, but I need a guidance for that. Do you know if there is a existing project I can contribute to?

Best reagards,
– Laurent LAPORTE.

tk0miya:
>This PR is mainly a bug fix for the 1.6.6 version; it adds a small improvement (dates).

Now we're planning next major release 1.7 in several months. So 1.6 series will be deprecated at same time. I can accept this only for stable branch. But it will be lost soon. I feel it is not good solution.

>Your decision to drop the epub2 builder is understandable and I want to respect that.

There are two main reasons. First one is manpower. We have only a few time for maintenance. So we'd like to make sphinx-core slim.
The another is knowledge. We don't have knowledge for each output formats. So, IMO, we can't updates the builders any more.

> I can offer you to remove the improvement about dates. That way, there is no new feature.

If you'd like to merge this into stable branch, please reduce the new feature.
The branch is only used for bug fix. It should keep compatibility with older version basically.

>I have thought of developing a epub2 extension, but I need a guidance for that. Do you know if there is a existing project I can contribute to?

AFAIK there are no project. So you need to start a new project.

tk0miya:
I must mention you about HTML builder.
Current Sphinx uses HTML4 builder by default (not XHTML). But in nearly future, we will move to HTML5 builder (in 1.8 or 2.0?).
At that time, the contents of EPUB would be moved to HTML5 also, because epub* builder uses HTML builder internally.

stephenfin:
@tantale If you were willing to maintain it, I would be happy to provide the info you need to make this a separate plugin and help you in releasing it.

tantale:
@stephenfin Thank you. I took a look at the documentation, but some advice could be very useful. Anyway, I think this is a pity because implementing a plug-in would required more efforts than this fix. We need to keep in mind that my aim is to have something "professional", not a kind of Proof of Concept.

tk0miya:
Finally, we released 1.7 yesterday, and removed epub2 builder completely.
Thanks,

tantale:
I saw that.
This PR is a bug fix for the 1.6.x branch anyway.
Any chance to accept it?
Thanks.

tk0miya:
Unfortunately, the support of 1.6 series is ended.
So we don't have a plan to update it any more.

stephenfin:
@tantale I'm still working on an extension to add the `epub2` package as a separate builder. Will ping you when I finish it (it's been a busy few weeks for me)

PR #3116, opened 2016-11-03
Breakable inline literals in LaTeX

jfbu:
Addresses #3110. Also, unify handling of inline literals by applying to all types (literalintitle, literalstrong, literalemphasis) use of ``\sphinxcode`` rather than simply ``\texttt`` (which was the old definition of ``\sphinxcode`` aka ``\code``). Also, use ``\protected`` prefix (Sphinx assumes for long time e-TeX via use of  ``\ifdefined``, ``\dimexpr`` ... hence we should also use ``\protected`` for more robust macros).

PR #2627, opened 2016-06-06
LaTeX use straight quotes also in inline code

jfbu:
Was already the case in code-blocks, but not for `'inline'` mark-up.

I tested on Sphinx's own doc for example.

PR #4370, opened 2018-01-02
Easy-up user customization of ``\sphinxcode`` LaTeX macro

jfbu:
Split out fancy LaTeX macros from ``\sphinxcode`` into ``\sphinxupquote``

Since #2627 (1.4.4), `\code`, and then again at #3116 (1.5) `\sphinxcode` which is the new name has become more complicated than the original `\texttt{#1}`. This was to obtain straight quotes in PDF output, and to allow long inline literals to break across lines.

This means though that users who want to customize `\sphinxcode`, for example to not only do `\texttt` but to use some colour, have to copy about 10 lines of complicated LaTeX macros which should be not modified in any way. This commit moves all the code out of `\sphinxcode` into a separate macro `\sphinxupquote`.

The LaTeX writer will output `\sphinxcode{\sphinxupquote{foo}}` in place of former `\sphinxcode{foo}`. Moving the `\texttt` from innermost to outermost level is with no consequence.

A second commit uses easier to understand way of defining the other styling macros, so that LaTeX user examining `sphinx.sty` sees more easily how the macros are supposed to be defined.

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -1580,21 +1580,21 @@
 
 % additional customizable styling
 % FIXME: convert this to package options ?
-\protected\def\sphinxstyleindexentry     {\texttt}
-\protected\def\sphinxstyleindexextra #1{ \emph{(#1)}}
-\protected\def\sphinxstyleindexpageref   {, \pageref}
-\protected\def\sphinxstyletopictitle #1{\textbf{#1}\par\medskip}
+\protected\def\sphinxstyleindexentry   #1{\texttt{#1}}
+\protected\def\sphinxstyleindexextra   #1{ \emph{(#1)}}
+\protected\def\sphinxstyleindexpageref #1{, \pageref{#1}}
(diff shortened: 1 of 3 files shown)


tk0miya (approved):
LGTM!

jfbu:
Thanks for quick review @tk0miya ! I will move the

    \protected\def\sphinxstrong#1{{\textbf{#1}}}

line next to the other macro definitions, (i.e. after the long `\sphinxupquote` macro) and remove from this`\sphinxstrong` a superfluous brace pair which is a left-over from ab5d4107c810e360b644990135ff6ef26bab53aa when `\bf` got replaced by `\textbf`.

Before merge I will force push this squashed change for a cleaner commit history.

I have been tempted to replace `\sphinxcode` when used in `\sphinxbfcode`, `\sphinxstyleliteralemphasis` etc... (in `sphinx.sty` file) by a simple `\texttt` in the spirit of separate customizability, but as the user can always herself redefine these macros if she changes `\sphinxcode`, this not such a pressing issue.

jfbu:
a space gobbling issue in `\sphinxupquote` (if `#1` ends with a space) got fixed at c892fe98f

Issue #4246, opened 2017-11-16
Limit page width in all themes


timhoffm:
### Problem
The following themes do not have a width limit on their pages:

- nature
- classic
- traditional
- pyramid
- bizstyle

This can result in extremely long text lines on large screens, making text difficult to read.

### Proposal

I propose to change the css of the above projects to limit the page width.

In case there are concerns that this changes the themes and breaks backward compatibility, one could at least add an option, so that users could set the width in the config.

timhoffm:
If the basic idea is ok, I could write a pull request.

tk0miya:
+1

tk0miya:
Now #4246 is merged. It will be release as 1.7.
Thank you for your contribution!

AndrewShepherd:
I understand why you did this, but is there a way that we can configure our setup (add to the `conf.py` file) so it reverts to `max-width:none` or `max-width:fill-available`?

timhoffm:
It should be possible to set the body_max_width parameter to strings like initial, none or fill-available.

See also 
https://www.sphinx-doc.org/en/master/usage/theming.html

pradyunsg:
Noting that the PR that fixes this was https://github.com/sphinx-doc/sphinx/pull/4376 (not #4246 -- which is this issue itself -- as noted by @tk0miya).

PR #4376, opened 2018-01-04
#4246: Limit width of text body for all themes

timhoffm:
### Feature: Limit width of text body for all themes

Implements #4246 

The purpose of this PR is to limit the text with of the documentation pages to a reasonable size:
- Do not make too long lines, even if the horizontal screen space is very large.
- Do not wrap to arbitrarily small widths (instead add a scrollbar if necessary).

This PR adds `min-width` and `max-width` CSS properties to the body. 

They are configurable via the theme options `minbodywidth` and `maxbodywidth`.

Note: I've named the options without any separating chars as this seems to be the general style for theme options (e.g. `sidebarlinkcolor`). This makes it a bit hard to read and I wouldn't mind renaming to `min_body_width` (also for a more pythonic feeling of `conf.py` `html_theme_options`). Please let me know what the naming conventions for theme options are.


doc/theming.rst
```diff
@@ -117,6 +117,12 @@ These themes are:
   - **sidebarwidth** (an integer): Width of the sidebar in pixels.  (Do not
     include ``px`` in the value.)  Defaults to 230 pixels.
 
+  - **minbodywidth** (an integer): Minimal width of the document body in
+    pixels.  Defaults to 450 pixels.
+
+  - **minbodywidth** (an integer): Maximal width of the document body in
+    pixels.  Defaults to 800 pixels.
```
tk0miya: Does this value also go well with themes not having sidebars (ex. haiku, scrolls and so on)?

tk0miya:
Note: test has been failed because of timed-out. So I restarted it manually and passed now.

timhoffm:
I've refined the PR:

- Options are now `body_min_width` / `body_max_width`. I realized that there are already some options using underscores. So having underscores does not break conventions, and it's more readable. Also I've changed the pattern to `[element]_[css_property]`.
- The new options as well as the existing `sidebarwidth` do now support integer values (interpreted as pixels) as well as any valid CSS dimension string, e.g. '70em'.
- Works now with all themes (except epub, where IMO the option does not make sense - but I'm no expert with epub).

doc/theming.rst
timhoffm: The refined version does now.

tk0miya:
Thanks!

PR #4377, opened 2018-01-05
Improved sidebar search field style

timhoffm:
### Feature: Improved sidebar search field style

The current sidebar search does not align the search button nicely:

grafik

This is a tuning of the CSS of the side bar search field inspired by: https://www.w3schools.com/howto/howto_css_search_button.asp

See below for the looks in different themes.

*Note:* As opposed to the above template, I have not used the magnifying glass icon for the button, because I did not want to introduce the FontAwesome deprendency. The text button may become an issue when the text is translated to a longer one. Then the button may still wrap below the input field.
If you're ok with the dependency, I could use the icon.

### Examples

bizstyle
grafik

classic
grafik

nature
grafik

pyramid
grafik

sphinxdoc
grafik

### DIFF
--- a/sphinx/themes/basic/searchbox.html
+++ b/sphinx/themes/basic/searchbox.html
@@ -10,12 +10,14 @@
 {%- if pagename != "search" and builder != "singlehtml" %}
 <div id="searchbox" style="display: none" role="search">
   <h3>{{ _('Quick search') }}</h3>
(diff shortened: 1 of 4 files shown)


tk0miya:
LGTM!

tk0miya:

PR #4381, opened 2018-01-06
test: Reduce DeprecationWarning on testing (from docutils)

tk0miya:
It seems many DeprecationWarnings are shown on Travis CI.
```
  /home/travis/build/sphinx-doc/sphinx/.tox/py36/lib/python3.6/site-packages/docutils/io.py:245: DeprecationWarning: 'U' mode is deprecated
    self.source = open(source_path, mode, **kwargs)
```

This try to filter them.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,6 +44,10 @@ incremental = True
 check_untyped_defs = True
 warn_unused_ignores = True
 
+[tool:pytest]
+filterwarnings =
+    ignore::DeprecationWarning:docutils.io
+
 [coverage:run]
 branch = True
 source = sphinx


setup.cfg
```diff
@@ -44,6 +44,10 @@ incremental = True
 check_untyped_defs = True
 warn_unused_ignores = True
 
+[tool:pytest]
+filterwarnings =
+    ignore::DeprecationWarning:docutils.io
```
tk0miya, line 49: At first, I added this filter into  `_filter_warnings()` in `tests/conftest.py`. But is is not work as intended.
I doubt the function is working.

And I also noticed `PYTHONWARNINGS` is not in passenv of tox.ini. And there is definition of `setenv PYTHONWARNINGS` in tox.ini.
So the definition on `.travis.yml` might have only made noise...

svenevs:
Hey @tk0miya I don't know if this does what you think.  I tried adding this to my own project to ignore the `docutils.io` deprecation warning, but it actually turned off _all_ warnings.  So right now I have three warnings: stop using `app.info`, stop using `app.warn`, and the `docutils.io` deprecation warning.  Adding

```ini
filterwarnings = 
    ignore::DeprecationWarning:docutils.io
```

turned off all of these warnings.  I don't really understand how this works, but I had to first turn on the deprecation warnings and then ignore

```ini
filterwarnings =
    default::DeprecationWarning
    default::PendingDeprecationWarning
    ignore::DeprecationWarning:docutils.io
```

The idea came from the last note [in this section of the `pytest` docs](https://docs.pytest.org/en/latest/warnings.html#pytest-mark-filterwarnings)

> `DeprecationWarning` and `PendingDeprecationWarning` are hidden by the standard library by default so you have to explicitly configure them to be displayed in your `pytest.ini`:
>
> ```ini
> [pytest]
> filterwarnings =
>     once::DeprecationWarning
>     once::PendingDeprecationWarning
> ```

`once` seems to work as advertised but I cannot find the documentation on it so I went with `default` instead [shown here](https://docs.python.org/3/library/warnings.html#describing-warning-filters).

Anyway the point is, it may be worth double checking that this didn't silence all deprecation warnings.  If it isn't obvious, I don't really understand any of this stuff, just trying to get around it xD  Hope this is helpful in some way.  This seems like something specific to `pytest`, but I'm not really sure...

tk0miya:
It seems some deprecation warnings are shown in our testings.
https://travis-ci.org/sphinx-doc/sphinx/jobs/428170247#L1880-L1912

We also uses `PYTHONWARNINGS` envvar. It might be related.
https://github.com/sphinx-doc/sphinx/blob/2e9d311ad51d0a1f622b38b301a8d10746b84878/tox.ini#L22-L23

Issue #4256, opened 2017-11-21
The 'conf.py' generated by 'sphinx-quickstart' has too much noise


stephenfin:
The default `conf.py` file generated by Sphinx is far too complicated for its own good, containing a lot of stuff that really isn't necessary. We should come up with a plan to simplify this.

This was noted in the discussion for #4174

stephenfin:
This is the opposite side of #4148.

tk0miya:
I don't think so. I believe we can adopt both together.
I think there are only a few commonly used (changed) settings. So we can reduce it.

IMO, giving better default setting also goes to simplify conf.py.
For example, the default value of `master_doc` is `contents`, but many people might use `index` for it. because quickstart generated it as by default. I think this is very stupid behavior.

timhoffm:
There are two different philosophies for a config file:

1. **Short:** Make it really short and include only the most common options. The advantage is that the user can easily see the active options. On the downside, it's more difficult to add options because you have to look them up in the manual. This heavily relies on the html documentation as an additional resource. I've attached a possible config file in a really stripped down version below.

2. **Complete:** Add a full list of options including descriptions. The config file can stand for itself, you can probably do most stuff without going to the manual. Jupyter notebook does it this way. They have a special formatting to distinguish between commented out (i.e. inactive) options and their preceeding desription. See below.

The current version has the differences of both worlds. It's quite verbose but you still need the docs because its not complete. That said, IMO both directions are viable ways for improvement and I have no clear preference. 

### Appendix

#### Excerpt from jupyter_notebook_config.py
~~~
#------------------------------------------------------------------------------
# NotebookApp(JupyterApp) configuration
#------------------------------------------------------------------------------

## Set the Access-Control-Allow-Credentials: true header
#c.NotebookApp.allow_credentials = False

## Set the Access-Control-Allow-Origin header
#  
#  Use '*' to allow any origin to access your server.
#  
#  Takes precedence over allow_origin_pat.
#c.NotebookApp.allow_origin = ''
~~~

#### Possible config file in a really stripped down version

~~~
# -*- coding: utf-8 -*-
#
# Configuration file for the Sphinx documentation builder.
#
# This file does only contain a selection of the most common options. For a
# full list see the documentation:
# http://www.sphinx-doc.org/en/stable/config

# -- Path setup -----------------------------------------------------------

# -- Project information --------------------------------------------------

project = 'Flaboo'
copyright = '2018, Tim Hoffmann'
[...]

PR #4385, opened 2018-01-06
Updates to conf.py generated by sphinx-quickstart

timhoffm:
As a follow-up to #4256, this PR makes some changes to the generated `conf.py`

In particular:
- Reduce the explanations at the top of the file and link to the documentation instead.
- Put the *Path setup* in a dedicated section.
- Create a section *Project information* and collect the basic information there.

Of course, one could do a lot more, but IMO this is a good first and self-contained step.

Here is how the new `conf.py` will look like.

~~~
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
#
# Flaboo documentation build configuration file, created by
# sphinx-quickstart on Sat Jan  6 19:23:50 2018.
#
# This file does only contain a selection of the most common options. For a
# full list see the documentation:
# http://www.sphinx-doc.org/en/stable/config

# -- Path setup -----------------------------------------------------------

#
# import os
# import sys


# -- Project information --------------------------------------------------

project = 'Flaboo'
copyright = '2018, Tim Hoffmann'
author = 'Tim Hoffmann'

# The short X.Y version
version = ''
# The full version, including alpha/beta/rc tags
release = '1.0'


# -- General configuration ------------------------------------------------

[...]
~~~

Some related questions:
- Is there a use for the shebang? AFAIK, this file will never be executed directly.
- What's the desired line width limit? The section headers stop at 76 chars, but text lines go up to 79 chars. That could be unified.


stephenfin:
Looks good. Thanks for tackling this 👍 

> Is there a use for the shebang? AFAIK, this file will never be executed directly.

Well, [it actually is](https://github.com/sphinx-doc/sphinx/blob/1.6.6/sphinx/config.py#L156-L157) 😉 However, shebangs only apply to the shell and this is never executed there. I think you can drop these.

> What's the desired line width limit? The section headers stop at 76 chars, but text lines go up to 79 chars. That could be unified.

I'd say 79, personally. It's entirely a matter of opinion.

sphinx/templates/quickstart/conf.py_t
```diff
@@ -6,14 +6,11 @@
 # {{ project }} documentation build configuration file, created by
 # sphinx-quickstart on {{ now }}.
```
stephenfin, line 7: I wonder if there would be more value in printing out the Sphinx version instead of the date? I don't see why the latter would be helpful, to be honest.

sphinx/templates/quickstart/conf.py_t
```diff
@@ -6,14 +6,11 @@
...
-# containing dir.
-#
-# Note that not all possible configuration values are present in this
-# autogenerated file.
-#
-# All configuration values have a default; values that are commented out
-# serve to show the default.
+# This file does only contain a selection of the most common options. For a
+# full list see the documentation:
+# http://www.sphinx-doc.org/en/stable/config
```
stephenfin, line 11: I've been meaning to do this for aaaaages. Thanks!

stephenfin (commented):
One nit, but nothing I'd block on. Nice work!

sphinx/templates/quickstart/conf.py_t
tk0miya: IMO, both are not worthy...

tk0miya:
LGTM!

>Is there a use for the shebang? AFAIK, this file will never be executed directly.

I also feel it is not needed. I'll remove it later.

>What's the desired line width limit? The section headers stop at 76 chars, but text lines go up to 79 chars. That could be unified.

I don't know and don't mind.

tk0miya (approved):

sphinx/templates/quickstart/conf.py_t
stephenfin: The advantage of adding the version is so you know for sure what version of Sphinx will work with a `conf.py` file (we sometimes drop support for features and options). However, I'd prefer less noise too so we can drop both.

tk0miya:

sphinx/templates/quickstart/conf.py_t
tk0miya: The generated `conf.py` might not depend on specific version of Sphinx (maybe). So it is not important.
And user can declare required version of Sphinx with `needs_sphinx` config implicitly. So I think it is not needed by default.

timhoffm: I agree, both are not relevant. The less clutter the better. I would have adapted this PR, but since it has been merged already. This change will have to go to another PR. There's more to do anyway.

PR #4386, opened 2018-01-06
Added partly more robust AST-based signature mangling

flying-sheep:
Subject: Improved signature mangling

- Bugfix

### Purpose
It can now handle (and strip) type annotations

It’s more robust for function with defaults that have a `__str__` method that returns valid python code.


tk0miya:
I fixed this problem by #6353. So I'm closing this now.
Thank you for your proposal.

Note: As discussed in #6311, the arguments list is not valid python code. So using ast is not best way. I'll consider the way later.

PR #4387, opened 2018-01-06
Napoleon: custom section headers

justaPCWingo:
Add the parsing of custom section headers in Google-style docstrings to the napoleon extension.

### Feature or Bugfix
- Feature

### Purpose
- I didn't see an obvious way to pass custom section header tags, so I modified napoleon with an extra option (napoleon_custom_sections, which is a list).

### Detail
The new feature is activated by passing a list in the *napoleon_custom_sections* option. Each option can be:
- string: A new section header identifier that will be parsed as a generic section.
- tuple of two string: Add another alias for a custom section, with the first string being the alias to use, 
   and the second argument being the builtin section that the alias maps to.
- test_ext_napoleon_docstring.py was extended with a simple unit test for these features (see 
   GoogleDocstringTest.test_custom_generic_sections).


RobRuana:
@tk0miya FYI, I have a free week coming up to take care of all the Napoleon issues backlog.

justaPCWingo:
Hmm. The errors reported by Travis CI appear to be issues that existed with code I didn't modify. Is this something that I should be worried about?

tk0miya:
Unfortunately, Travis CI has (network related?) trouble now. So some tests are failed even if no changes.
I restarted it manually last night and it seems resolved now.

RobRuana (approved):
LGTM!

Issue #4093, opened 2017-09-28
Sphinx build make mode creates empty directories for unknown targets/builders


markkuriekkinen:
Subject: the new Sphinx build "make mode" creates empty directories under the build directory even when the builder argument refers to a non-existent builder

### Problem
- Sphinx quickstart 1.6 generates a Makefile with the catch-all target that uses the Sphinx "make mode" for building
- `make` can not check if the given argument is sensible since anything hits the catch-all target
- The Sphinx make mode gives an error when the given builder is not found, however, an empty directory with the target name has already been created under the build directory
- The empty directory should not be created at all if the builder does not even exist

#### Procedure to reproduce the problem
```
make fake
```

#### Error logs / results
```
Running Sphinx v1.6.4
making output directory...
loading translations [en]... done

Sphinx error:
Builder name fake not registered or available through entry point
Makefile:27: recipe for target 'fake' failed
make: *** [fake] Error 1
```
Sphinx gives an error that the `fake` builder does not exist, but nonetheless, an empty directory `fake` has been created under the build directory.

#### Expected results
The Sphinx make mode should not create empty directories for unknown builders.

### Environment info
- OS: Ubuntu Linux 16.04
- Python 3.5.2
- Sphinx version: 1.6.4
- GNU Make 4.1

tk0miya:
Fixed by #4389 
Thank you for reporting!

PR #4389, opened 2018-01-07
Fix creating build directory for unknown build target

jbking:
Subject: Fix creating build directory for unknown build target

### Feature or Bugfix
- Bugfix

### Purpose
- Fix building target specified to make command becomes build directory name even unknown target or typo.
  Sphinx ensures their existence at bootstrap. So unexpected build directory was created.

### Relates
- #4093

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -157,10 +157,6 @@ class Sphinx(object):
         # status code for command-line application
         self.statuscode = 0
 
-        if not path.isdir(outdir):
-            logger.info('making output directory...')
-            ensuredir(outdir)
-
         # read config
         self.tags = Tags(tags)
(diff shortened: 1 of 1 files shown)


tk0miya:
Note: The directory creation has been moved after these processing:

* execute `conf.py` (but it could not refers outdir)
* setup_extension() for bundled and user-given extensions
* preload_builder (call setup_extension() for given builder-name)


As a result, `app.outdir` might point a non-existing directory during setup extensions.
This is surely incompatible change, but IMO it does not harm anybody.
So +1 for this!

tk0miya (approved):

shimizukawa (approved):
LGTM!

tk0miya:

PR #4391, opened 2018-01-07
Prefer https & readthedocs.io instead of readthedocs.org for links

jdufresne:
Read the Docs moved hosting to readthedocs.io instead of
readthedocs.org. Fix all links in the project.

For additional details, see:

https://blog.readthedocs.com/securing-subdomains/

> Starting today, Read the Docs will start hosting projects from subdomains on
> the domain readthedocs.io, instead of on readthedocs.org. This change
> addresses some security concerns around site cookies while hosting user
> generated data on the same domain as our dashboard.


tk0miya:
Good point! Thanks :-)

PR #4392, opened 2018-01-08
Builder's interrupt event. Fix #1649

### LINKED ISSUE #1649: Detect when html_extra_path files need to be copied
If you add files into `html_extra_path`, Sphinx doesn't rebuild to copy them. It will help if Sphinx did an extra check that `html_extra_path` files are present in target directory before falling out.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1649
- Originally reported by: anatoly techtonik
- Originally created at: 2014-12-13T14:33:27.513

jbking:
Subject: Builder's finish event can be separate some events.

### Feature or Bugfix

- Feature
### Purpose
- By design, build process can be interrupted or finished.
  Some tasks executed in finish state should also be executed in interrupt state.

### Detail
- Introduce interrupt event hook point at Builder
- Fix #1649 by execute copying tasks also in interrupt case

### Relates
- #1649

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -99,6 +99,7 @@ class Builder(object):
 
         # these get set later
         self.parallel_ok = False
+        self.interrupt_tasks = None  # type: Any
         self.finish_tasks = None  # type: Any
 
     def set_environment(self, env):
@@ -361,6 +362,9 @@ class Builder(object):
         else:
             if method == 'update' and not docnames:
                 logger.info(bold('no targets are out of date.'))
+                self.interrupt_tasks = SerialTasks()
+                self.interrupt()
+                self.interrupt_tasks.join()
                 return
 
         # filter "docnames" (list of outdated files) by the updated
@@ -474,6 +478,14 @@ class Builder(object):
         # type: (unicode, nodes.Node) -> None
         """Handle parts of write_doc that must be called in the main process
         if parallel build is active.
+        """
+        pass
+
+    def interrupt(self):
+        # type: () -> None
+        """Interrupt the building process.
+
+        The default implementation does nothing.
         """
         pass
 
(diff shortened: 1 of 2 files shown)


sphinx/builders/__init__.py
```diff
@@ -361,6 +362,9 @@ def build(self, docnames, summary=None, method='update'):
         else:
             if method == 'update' and not docnames:
                 logger.info(bold('no targets are out of date.'))
+                self.interrupt_tasks = SerialTasks()
+                self.interrupt()
+                self.interrupt_tasks.join()
```
tk0miya, line 367: I think these process is not "interrupted". So I feel the naming is strange.
In addition, `Builder.interrupt()` looks like the builder goes to interrupt the process because of the name. But, in actual, this is a handler on skipping build. So I feel also strange.
(I know another handler on finished is named as `Builder.finish()`.)

So I feel another naming is good to this case. How about adding a parameter to `Builder.finish()` (like `status`)?

Note: in future, IMO, it would be nice if we can rename `Builder.finish()` also.

tk0miya: Anyway, we should not use the word "interrupt" for this case. This is not interruption.

sphinx/builders/html.py
```diff
@@ -581,6 +580,14 @@ def finish(self):
         # dump the search index
         self.handle_finish()
 
+    def interrupt(self):
+        # type: () -> None
+        self.interrupt_tasks.add_task(self.copy_image_files)
+        self.interrupt_tasks.add_task(self.copy_download_files)
+        # not only copying, generating.
+        # self.interrupt_tasks.add_task(self.copy_static_files)
```
tk0miya, line 588: We should call `copy_static_files` also if you want to fix #1649.
I know this resolved the problem of #1649 itself. But this is not consistent behavior. So the another issue will be posted.

tk0miya (changes requested):
I feel good for your proposal. So let's improve this to more stable and good API :-)

jbking:
Thanks for the review. I'm going to improve this change with your idea.

tk0miya:
Now I determined to close this because we can't find the better name for this API. Thank you for your contribution!

PR #4395, opened 2018-01-08
Fix #4081: Warnings and errors colored the same when building

### LINKED ISSUE #4081: Warnings and errors colored the same when building
### Problem

Running `make html` (presumably the same behavior occurs for other targets) results in warnings that are colored in red the same as errors. This is a severe usability problem as it requires significantly more attention to determine if there is actually an issue that needs to be addressed or not.

```
make html
```

#### Expected results

Warnings should use a color that is not red (perhaps yellow instead).

### Environment info
- OS: Mac and Linux
- Python version: any
- Sphinx version: 1.6.1, 1.6.2

tk0miya:
refs: #4081 

With this change, warnings are shown as following:

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -58,7 +58,6 @@ Features added
 * #4332: Let LaTeX obey :confval:`math_numfig` for equation numbering
 * #4093: sphinx-build creates empty directories for unknown targets/builders
 
-
 Features removed
 ----------------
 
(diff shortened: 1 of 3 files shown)


shimizukawa (approved):
LGTM ;)

tk0miya:
Thank you for reviewing, merged.

PR #4396, opened 2018-01-08
Fix #1922: Upper characters problem in French HTML Search

### LINKED ISSUE #1922: Upper characters problem in French stemmer
I've found a bug in the HTML / Javascript search engine with the French language. The bug is actually related to the indexing of the doctree. The HTML / Javascript search engine expects only lower characters in the index. When indexing, the English stemmer removes upper characters and replace them with lower ones (with .lower()). However, the French stemmer (in sphinx/search/fr.py) keeps the upper characters. Consequently, any word with an upper character in wrongly indexed in French.
This can easily be fixed by adding ".lower()" in SearchFrench.stem().

tk0miya:
refs: #1922

### DIFF
--- a/sphinx/search/fr.py
+++ b/sphinx/search/fr.py
@@ -207,4 +207,4 @@ class SearchFrench(SearchLanguage):
         self.stemmer = snowballstemmer.stemmer('french')
 
     def stem(self, word):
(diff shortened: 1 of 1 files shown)


tk0miya:
Confirmed at #1922.

PR #4397, opened 2018-01-08
Fixes bug when evaluating entry to bool

glyg:
As reported here: https://github.com/rtfd/readthedocs.org/issues/3411 sphinx sometimes fails with the error:
```python
  File "/home/docs/checkouts/readthedocs.org/user_builds/drf-yasg/envs/latest/lib/python3.5/site-packages/sphinx/ext/viewcode.py", line 74, in has_tag
    if entry is None or entry[0] != code:
TypeError: 'bool' object is not subscriptable
```
This is not critical as whipping the build or even just running it again fixes it, but the error is confusing...  
I believe switching the two if statement above should prevent this from happening.

Subject: Fix a bug in evaluating entry

### Bugfix
- Bugfix

### Purpose
- Fix a `TypeError` when evaluating `entry` in `view_code.py`. 

### Detail

 When `entry` is `False` in the current version of the code, is first goes through  
```python
if entry is None or entry[0] != code: 
```
which raises a `TypeError` before evaluating `if entry is False`
Switching the `if / elif` statement should fix that without any side effects.

### Relates
- <URL or Ticket>

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,10 @@ Features added
   :ref:`LaTeX 'sphinxsetup' <latexsphinxsetup>` key (refs: #4285)
 * Easier customizability of LaTeX macros involved in rendering of code-blocks
 * Show traceback if conf.py raises an exception (refs: #4369)
+* Add :confval:`smartquotes` to disable smart quotes through ``conf.py``
+  (refs: #3967)
+* Add :confval:`smartquotes_action` and :confval:`smartquotes_excludes`
+  (refs: #4142, #4357)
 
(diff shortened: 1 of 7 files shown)


tk0miya:
Could you rebase this to stable branch? Then I will include this to 1.6.6 release (I'll ship it today)

tk0miya:
The error on Travis CI is misinformation. I restarted these tests manually, so they will be resolved soon.

glyg:
Hi @tk0miya I rebased onto stable, opening a PR against stable now

PR #4399, opened 2018-01-08
Fixes bug when evaluating entry to bool

glyg:
As reported here: https://github.com/rtfd/readthedocs.org/issues/3411 sphinx sometimes fails with the error:
```python
  File "/home/docs/checkouts/readthedocs.org/user_builds/drf-yasg/envs/latest/lib/python3.5/site-packages/sphinx/ext/viewcode.py", line 74, in has_tag
    if entry is None or entry[0] != code:
TypeError: 'bool' object is not subscriptable
```
This is not critical as whipping the build or even just running it again fixes it, but the error is confusing...  
I believe switching the two if statement above should prevent this from happening.

### Feature or Bugfix
- Bugfix

See #4397 for details

### DIFF
--- a/sphinx/ext/viewcode.py
+++ b/sphinx/ext/viewcode.py
@@ -71,12 +71,12 @@ def doctree_read(app, doctree):
             code = analyzer.code.decode(analyzer.encoding)
         else:
             code = analyzer.code
-        if entry is None or entry[0] != code:
+        if entry is False:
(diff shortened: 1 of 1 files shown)


tk0miya:
Merged. Thank you for quick work!

glyg:
Thanks @tk0miya !

PR #4388, opened 2018-01-07
stable to master

tk0miya:
I'll merge this after Travis says okay.

PR #4405, opened 2018-01-09
Fix smartquotes has been ignored

tk0miya:
This fixes mistake on merging (refs: #4388)

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -665,7 +665,7 @@ class BuildEnvironment(object):
         language = self.config.language or 'en'
         self.settings['language_code'] = language
         if 'smart_quotes' not in self.settings:
-            self.settings['smart_quotes'] = True
+            self.settings['smart_quotes'] = self.config.smartquotes
 
             # some conditions exclude smart quotes, overriding smart_quotes
             for valname, vallist in iteritems(self.config.smartquotes_excludes):


sphinx/environment/__init__.py
```diff
@@ -665,7 +665,7 @@ def prepare_settings(self, docname):
         language = self.config.language or 'en'
         self.settings['language_code'] = language
         if 'smart_quotes' not in self.settings:
-            self.settings['smart_quotes'] = True
+            self.settings['smart_quotes'] = self.config.smartquotes
 
             # some conditions exclude smart quotes, overriding smart_quotes
             for valname, vallist in iteritems(self.config.smartquotes_excludes):
```
jfbu, line 671: Currently we leave this conditional to `'smart_quotes' not in self.settings`. This means that smartquotes_excludes is disabled for extensions which set `'smart_quotes'`. As this was for a work-around for people who wanted to disable smart quotes, perhaps we could make this ``for valname, vallist`` block non-conditional now. It only helps to disable smart quotes even more `:-)`.

tk0miya: I just changed this behavior on #4407. Could you review this please instead?
https://github.com/sphinx-doc/sphinx/pull/4407/files#diff-41cfa2a9a00ea16c558726885eb1239aR221

Issue #4371, opened 2018-01-03
Move smart_quotes feature to post_transform


tk0miya:
### Problem
- Not to cache the result of smart_quotes, the processing should be moved to post_transform.

#### Procedure to reproduce the problem
```
$ make html text
or
$ make text html
```
(with #4360)

#### Error logs / results
- Some outputs are applied smart_quotes or not applied.

#### Expected results
- Applied smart_quotes correctly under `smartquotes_disables`.

#### Note:

- This might requires a big refactoring. So it would be implemented 1.7 or later.

CendioOssman:
I'm not sure what the issue is here? The transform doesn't look like it is doing terribly complex things. So what is preventing it from just being changed to a post transform?

We got hit by this by doing both man pages and html output with a shared doctree cache. And suddenly the order of things became very important.

tk0miya:
At present, some transforms are invoked after smart_quotes transform and are collecting contents from the "smart-quoted" doctree. For example, the standard domain collects section titles, titles of anchors, and so on. If we'll move the smart_quotes transform to post transforming phase, these collected contents will not be smart-quoted.

To change the order of processing, we need to create a new mechanism to do smart-quoting the collected contents.

CendioOssman:
Ah, that does complicate things. Most cases would hopefully be handled if smart quotes is run sufficiently late in the process though? E.g. reference titles.

And I don't suppose anyone has considered moving those collections to somewhere later? I know I had some issues with the fact that adding certain nodes in the post transform stage didn't work as the standard domain didn't collect information from those.

tk0miya:
We also have similar problems for i18n feature. And we've not been able to find the better way to do them on at the later step yet.

gmilde:
Alternatively, determine if the setting (rsp. setting default) changed and clear the cache.
(We cannot put every transition last...)

jdillard:
#8974 (duplicate issue) provides a standalone reproducer: build with the text builder, then the XML builder using the same environment.

The XML output incorrectly retains straight quotes because it reuses a doctree cached without SmartQuotes. This confirms the problem occurs when builder-specific transform results are shared across builders.

PR #4407, opened 2018-01-09
Move smartquotes to post_transform

tk0miya:
refs: #4371.

Move SmartQuotes to post_transforms. It enables to evaluate it on each build.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -65,7 +65,7 @@ Features added
 * #947: autodoc now supports ignore-module-all to ignore a module's ``__all__``
 * #4332: Let LaTeX obey :confval:`math_numfig` for equation numbering
 * #4093: sphinx-build creates empty directories for unknown targets/builders
-* :confval:`smartquates` is evaluated on each build
+* :confval:`smartquotes` is evaluated on each build
 
 Features removed
 ----------------
--- a/CHANGES
+++ b/CHANGES
@@ -65,6 +65,7 @@ Features added
 * #947: autodoc now supports ignore-module-all to ignore a module's ``__all__``
 * #4332: Let LaTeX obey :confval:`math_numfig` for equation numbering
 * #4093: sphinx-build creates empty directories for unknown targets/builders
+* :confval:`smartquates` is evaluated on each build
 
 Features removed
 ----------------
--- a/sphinx/transforms/post_transforms/__init__.py
+++ b/sphinx/transforms/post_transforms/__init__.py
@@ -12,14 +12,16 @@
 import warnings
 
 from docutils import nodes
+from docutils.transforms.universal import SmartQuotes
 from docutils.utils import get_source_line
 
 from sphinx import addnodes
 from sphinx.deprecation import RemovedInSphinx20Warning
-from sphinx.environment import NoUri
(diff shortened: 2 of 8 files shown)


CHANGES
```diff
@@ -65,6 +65,7 @@ Features added
 * #947: autodoc now supports ignore-module-all to ignore a module's ``__all__``
 * #4332: Let LaTeX obey :confval:`math_numfig` for equation numbering
 * #4093: sphinx-build creates empty directories for unknown targets/builders
+* :confval:`smartquates` is evaluated on each build
```
jfbu: typo s/quates/quotes/

sphinx/transforms/post_transforms/__init__.py
```diff
@@ -197,11 +200,71 @@ def apply(self):
...
+            return
+
+        SmartQuotes.apply(self)
+
+    def is_available(self):
+        # type: () -> bool
+        builders = self.config.smartquotes_excludes.get('builders', [])
+        languages = self.config.smartquotes_excludes.get('languages', [])
+
+        if self.env.settings['smart_quotes'] is False:
```
jfbu, line 221: +1 for this way

jfbu (commented):
(I just left one minor remark, I have tasks and will review later)

CHANGES
tk0miya: Fixed

sphinx/transforms/post_transforms/__init__.py
```diff
@@ -197,11 +200,71 @@ def apply(self):
...
 
 
+class SphinxSmartQuotes(SmartQuotes, SphinxTransform):
+    """
+    Customized SmartQuotes to avoid transform for some extra node types.
+    """
+    def apply(self):
+        # type: () -> None
+        if self.is_available():
+            self.env.settings['smart_quotes'] = True
```
tk0miya, line 210: This should override `self.document.settings` instead. docutils' SmartQuotes class refers it.

`env.settings` is not referred here. It is prepared on `BuildEnvironment` class and used for create document object (see `sphinx.io:read_doc()`). So modifying the key is too late here.

sphinx/transforms/post_transforms/__init__.py
```diff
@@ -197,11 +200,71 @@ def apply(self):
                     node.replace_self(nodes.comment())
 
 
+class SphinxSmartQuotes(SmartQuotes, SphinxTransform):
```
tk0miya, line 203: Ah, I noticed post_transforms does not effect to toctree and other data. As a result, some testcases are failed now. I need to refactor this more and more.

tk0miya: In addition, the document object generated in `SphinxTransformer` does not have `settings.smart_quotes` attribute. Because it does not use `RSTParser` on instantiation.
As a result, some tests are failed (ex `test_ext_todo.py`).

tk0miya:
I feel the moving is a bit difficult. So I postponed this to 1.8 milestone.

tk0miya:
I withdraw this because much rework is needed.

PR #4410, opened 2018-01-10
Refactor smartquotes

tk0miya:
Before moving SmartQuotes to post transforms, I refactored it as standard Transform module.
This is simply refactoring.


jfbu (approved):
Testing looked ok. I checked also with Docutils 0.13.1 (as we monkey patch then docutils.smartquotes.smartchars). But my review is limited by still narrow bounds of my understanding of Sphinx's inner workings...

sphinx/ext/autodoc/importer.py
```diff
@@ -82,15 +82,9 @@ def __getattr__(self, name):
 
 
 class _MockImporter(object):
-    def __init__(self, names):
+    def __init__(self, modules):
```
tk0miya: oh, this contains my test code. this is not related with smart quotes...

tk0miya:

PR #4411, opened 2018-01-10
latex: Do not display Release label if :confval:`release` is not set

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- Since 1.6, we can omit `release` configuration, but LaTeX writer expects required
- This allows to omit it for LaTeX output

### Relates
- https://groups.google.com/forum/#!topic/sphinx-users/L5PUfwVu8f0


jfbu (approved):
LGTM

After merge I will fix the problem of the comma showing in header if both releasename and release (version number) are left empty.

tk0miya:
Thank you for reviewing :-)

PR #4413, opened 2018-01-11
autodoc: only mock specified modules with their descendants.

### LINKED ISSUE #2557: autodoc mock imports the entire module hierarchy when mock importing a submodule
I'm trying to mock the import of a platform specific submodule `app.plugin.win64`, but I would like to have autodoc generate documentation for `app` and `app.plugin` parent modules.

`conf.py`:

``` python
autodoc_mock_imports = ['app.plugin.amd64']
```

`autodoc.py` appears to recursively walk up the module hierarchy generating mock objects until there is a mock object for each node in the hierarchy:

``` python
def mock_import(modname):
    if '.' in modname:
        pkg, _n, mods = modname.rpartition('.')
        mock_import(pkg)
    mod = _MockModule()
    sys.modules[modname] = mod
    return mod
```

This generates:

``` python
sys.modules['app'] = _MockModule()
sys.modules['app.plugin'] = _MockModule()
sys.modules['app.plugin.win64'] = _MockModule()
```

autodoc automatically generates mock objects for `app` and `app.plugin` modules, even though I didn't specify them in the `autodoc_mock_imports` configuration option.

Is it possible to mock a submodule without autodoc mocking each parent node in the module hierarchy?

rjarry:
Do not mock the ancestors of the specified modules in `autodoc_mock_imports`. Only mock the modules themselves and their descendants (as specified in the docs). This changes behavior compared to the original mock import implementation that was done in version 1.3 where all ancestors were also mocked implicitly (may require warning in release notes?).

I fixed the test configs accordingly.

Fixes #2557

### DIFF
--- a/sphinx/ext/autodoc/importer.py
+++ b/sphinx/ext/autodoc/importer.py
@@ -23,7 +23,7 @@ from sphinx.util.inspect import isenumclass, safe_getattr
 
 if False:
     # For type annotation
-    from typing import Any, Callable, Dict, Generator, List, Optional, Set  # NOQA
+    from typing import Any, Callable, Dict, Generator, List, Optional  # NOQA
 
 logger = logging.getLogger(__name__)
 
@@ -84,13 +84,7 @@ class _MockModule(ModuleType):
 class _MockImporter(object):
     def __init__(self, names):
         # type: (List[str]) -> None
-        self.base_packages = set()  # type: Set[str]
-        for n in names:
-            # Convert module names:
-            #     ['a.b.c', 'd.e']
-            # to a set of base packages:
-            #     set(['a', 'd'])
-            self.base_packages.add(n.split('.')[0])
+        self.names = names
         self.mocked_modules = []  # type: List[str]
         # enable hook by adding itself to meta_path
         sys.meta_path = sys.meta_path + [self]
@@ -106,9 +100,10 @@ class _MockImporter(object):
 
     def find_module(self, name, path=None):
         # type: (str, str) -> Any
-        base_package = name.split('.')[0]
-        if base_package in self.base_packages:
-            return self
+        # check if name is (or is a descendant of) one of our base_packages
+        for n in self.names:
+            if n == name or name.startswith(n + '.'):
+                return self
         return None
 
     def load_module(self, name):
--- a/tests/roots/test-root/conf.py
+++ b/tests/roots/test-root/conf.py
@@ -69,9 +69,10 @@ extlinks = {'issue': ('http://bugs.python.org/issue%s', 'issue '),
 
 autodoc_mock_imports = [
     'missing_module',
-    'missing_package1.missing_module1',
-    'missing_package2.missing_module2',
-    'missing_package3.missing_module3',
+    'missing_package1',
+    'missing_package2',
+    'missing_package3',
+    'sphinx.missing_module4',
 ]
 
 # modify tags from conf.py
--- a/tests/roots/test-root/autodoc_missing_imports.py
+++ b/tests/roots/test-root/autodoc_missing_imports.py
@@ -4,6 +4,8 @@ from missing_module import missing_name
 import missing_package1.missing_module1
 from missing_package2 import missing_module2
 from missing_package3.missing_module3 import missing_name
+import sphinx.missing_module4
+from sphinx.missing_module4 import missing_name2
 
 @missing_name
 def decoratedFunction():
@@ -16,3 +18,5 @@ class TestAutodoc(object):
     def decoratedMethod(self):
         """TestAutodoc::decoratedMethod docstring"""
         return None
+
+sphinx.missing_module4.missing_function(len(missing_name2))


sphinx/ext/autodoc/importer.py
```diff
@@ -84,13 +84,13 @@ def __getattr__(self, name):
 class _MockImporter(object):
     def __init__(self, names):
         # type: (List[str]) -> None
-        self.base_packages = set()  # type: Set[str]
-        for n in names:
+        self.base_packages = set()  # type: Set[Tuple[str, ...]]
```
tk0miya: Is this base package? It is better to me to rename like `target_packages` this because the purpose is changed.

sphinx/ext/autodoc/importer.py
```diff
@@ -106,9 +106,11 @@ def disable(self):
 
     def find_module(self, name, path=None):
         # type: (str, str) -> Any
-        base_package = name.split('.')[0]
-        if base_package in self.base_packages:
-            return self
+        package = tuple(name.split('.'))
+        # check if name is (or is a descendant of) one of our base_packages
+        for pkg in self.base_packages:
+            if package[:len(pkg)] == pkg:
```
tk0miya: I feel comparing `package[:len(pkg)]` is a bit unreadable.
How about this?
```
# self.target_packages contains package names as string (not tuple).
for pkg in self.target_packages:
    if name == pkg or name.startswith(pkg + '.'):
        return self
```

sphinx/ext/autodoc/importer.py
```diff
@@ -127,11 +129,13 @@ def load_module(self, name):
 @contextlib.contextmanager
 def mock(names):
     # type: (List[str]) -> Generator
+    importer = None
     try:
         importer = _MockImporter(names)
         yield
     finally:
-        importer.disable()
+        if importer is not None:
```
tk0miya: Why did you check `importer here`? Did you mean instantiation of `_MockImporter` might be failed?

tests/roots/test-root/conf.py
```diff
@@ -69,9 +69,9 @@
 
 autodoc_mock_imports = [
     'missing_module',
-    'missing_package1.missing_module1',
-    'missing_package2.missing_module2',
-    'missing_package3.missing_module3',
+    'missing_package1',
+    'missing_package2',
+    'missing_package3',
```
tk0miya, line 74: We have to add `existing_package.missing_module` pattern.

tk0miya (changes requested):
As commented at #2557, we need to add a new configuration to switch old and new behaviors to keep compatibility.
And make it new behavior by default :-)

tk0miya:
In addition, we have feature freeze for 1.7 release in this weekend.
Can you work until then? If not, I'd like to modify this.

sphinx/ext/autodoc/importer.py
rjarry: Ok

sphinx/ext/autodoc/importer.py
rjarry: Yes, I put the explanation in the git commit message:

https://github.com/sphinx-doc/sphinx/commit/fe24d5ca608589fcda3d6602ec3f742214c99064

rjarry:
I'm not sure I'll have time to work on it before this weekend. If you want to take over my draft, please do. I don't mind :)

rjarry:
@tk0miya I pushed the changes you suggested. Tell me what you think :)

sphinx/ext/autodoc/importer.py
tk0miya: Oh, sorry. I'd overlooked it. I understand.

tk0miya:
>As commented at #2557, we need to add a new configuration to switch old and new behaviors to keep compatibility.
>And make it new behavior by default :-)

we need to add this for backward compatibility. Can you work it?

To do that, we have to give an option to `_MockImporter`.
```
mocker = _MockImporter(names, mock_ascendants=True)
...
mocker.disable()
```
and give an option from autodoc side:
```
with mock(names, config.autodoc_mock_ascendants):
    ...
```
(I don't have time now, so the naming is very roughly)

rjarry:
I'm not sure adding a new option is a good idea.

In fact we are only fixing bad behavior. If some users are relying on this corner case, they can can change `autodoc_mock_imports` to directly specify base packages instead of sub modules.

tk0miya (approved):

tk0miya:
Absolutely, adding base package will resolve it instead option.
Okay, I'll mention that incompatible change in CHANGES.

rjarry:
Thanks @tk0miya!

tk0miya:
Merged. Thank you for great work!

tk0miya:
Now I added an incompatible entry for this: 0d61b251c5a6d55eb8a0ce6977e72078664e0a06.
Please let me know if wrong (I'm not a native and not good at English :-p)

rjarry:
Don't worry, I'm not a native either :)

I added a comment on the commit https://github.com/sphinx-doc/sphinx/commit/0d61b251c5a6d55eb8a0ce6977e72078664e0a06

Thanks @tk0miya!

PR #4424, opened 2018-01-13
Fix #4079: Add notranslate class to let Google Translate know they are not translatable

### LINKED ISSUE #4079: Better to adapt to Google Translate
Subject: <what happen when you do on which document project>

### Problem

When I use the chrome browser to translate the function to translate this page to Chinese.

http://gluon.mxnet.io/chapter07_distributed-learning/hybridize.html

image

But those examples of code are also translated, resulting in translation results can not read.

If the implementation of the following statement after the translation, the page is like this, so much better.

`jQuery('pre, .math, code').addClass('notranslate')`

image

Whether the Sphinx itself can provide better support for Google translation.

### Environment info
- OS: Mac
- Sphinx version: 1.6.3
- Browser:  Chrome 61.0.3163.91

tk0miya:
refs: #4079


stephenfin (approved):
Looks good to me

stephenfin:
This should be backport to 1.6, assuming we're planning to make a `1.6.7` release?

tk0miya:
@stephenfin Thank you for reviewing. I classified this as a new feature because this changes the behavior of generated HTML. So I would not backport this to 1.6.
(Anyway 1.7 will be released soon. So I don't worry)

@shibukawa Could you advice me something? I worried is there anything i'm missing...

shibukawa:
It seems good.

tk0miya:
Thanks!

tk0miya:

PR #4271, opened 2017-12-02
sphinx-build: Add support for "-j auto"

emfloyd2:
Adjust `-j` argument so it can accept special keyword `auto`, which will set `-j` to the number of CPUs on machine.

Change was inspired by headaches while trying to find the best way to get sphinx-build to use the max number of cores from any machine we build docs on (Ubuntu, Mac, AWS Linux, etc). Rather than write external scripts to set this (sysctl calls, grepping /proc/cpuinfo,etc), it made more sense to build the functionality into sphinx-build itself.

### Feature or Bugfix
- Feature

### Purpose
- Make it easy to scale `sphinx-build` to use the max number of cores on any machine (Without relying on external scripts)

### Detail
- New custom type in sphinx.cmdline for handling -j argument (required to handle both `int` and `str` input)
- Adjusted help string for `-j` flag

PR #4425, opened 2018-01-13
sphinx-build: Add support for "-j auto"

tk0miya:
This is improvements of #4271.


CHANGES
```diff
@@ -75,6 +75,8 @@ Features added
 * #4183: doctest: ``:pyversion:`` option also follows PEP-440 specification
 * #4235: html: Add :confval:`manpages_url` to make manpage roles to hyperlinks
 * #3570: autodoc: Do not display 'typing.' module for type hints
+* #4271: sphinx-build supports an option called ``-j0`` to adjust numbers of
+  processes automatically.
```
stephenfin: This says `-j0`, yet the code says `-jauto`. Which one is correct? Maybe both?

stephenfin (changes requested):

CHANGES
tk0miya: It's my wrong. `-j auto` is still correct. But we need to choose which is better.

Note: I think supporting both is worse. we should choose one. I love python's slogan; [TOOWTDI](https://wiki.python.org/moin/TOOWTDI)

stephenfin: Ah, personally I'd consider supporting both a usability improvement (`-j0` is more concise and can be used without the space, but `-j auto` is clearer), however, I wouldn't block on it. We can extend it in the future if anyone wants it.

PR #4426, opened 2018-01-13
latex: Emit warnings if the document contains aligned figures (refs: #3289)

tk0miya:
refs: #3289 

This is not fix for #3289. But this emits warnings to users to check aligned figures layouted fine.
@jfbu what do you think? is this useful?

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -24,7 +24,7 @@ from docutils.writers.latex2e import Babel
 from sphinx import addnodes
 from sphinx import highlighting
 from sphinx.errors import SphinxError
-from sphinx.locale import admonitionlabels, _
+from sphinx.locale import admonitionlabels, _, __
 from sphinx.transforms import SphinxTransform
 from sphinx.util import split_into, logging
 from sphinx.util.i18n import format_date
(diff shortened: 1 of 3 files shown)


sphinx/writers/latex.py
```diff
@@ -1852,6 +1852,11 @@ def visit_figure(self, node):
             self.body.append('\\begin{wrapfigure}{%s}{%s}\n\\centering' %
                              (node['align'] == 'right' and 'r' or 'l', length or '0pt'))
             self.context.append(ids + '\\end{wrapfigure}\n')
+
+            # emit a warning about wrapfig (refs: #3289)
+            logger.warning(__('A figure having :align: option has detected. '
+                              'LaTeX might not layout it well. Please check generated PDF.'),
```
jfbu, line 1858: Perhaps:
 
    'LaTeX perhaps did not layout it correctly. Please check generated PDF.'

But checking PDFs is time costly...

jfbu (commented):
-0 (or +0 if warning emitted only once)

I think a warning is not very useful if no workaround is proposed: people might need the align (left or right) for html in particular. There could be config "latex_warn_if_aligned_figure" and another one "latex_dontuse_wrapfig". But I have not thought about it hard (unfortunately I have tasks). I can make PR after this one.

tk0miya:
Yes, this fix is not essentially. If you can fix it in future, this is not needed.
Okay, i'm withdrawing this.

PR #4427, opened 2018-01-13
Minor conf.py cleanup

timhoffm:
Follow-up to #4385. As discussed there:

- Make the section delimiters 79 char wide.
- Simplify the text in the header. I now just says:
  ~~~
  ~~~
  instead of 
  ~~~
  # Flaboo documentation build configuration file, created by
  # sphinx-quickstart on Sat Jan  6 19:23:50 2018.
  ~~~
  Motivation: created-by and creation date are not relevant. The project name is also given in `project`.
  All of this information is usually also available via the context of the file. It will be part of a VCS repository. If really needed, one can look up the creation date there.. Also the user will know what project this is for. However, he may not know that `conf.py` is a config file for sphinx. So that's important to mention.

### DIFF
--- a/sphinx/templates/quickstart/conf.py_t
+++ b/sphinx/templates/quickstart/conf.py_t
@@ -1,13 +1,12 @@
 # -*- coding: utf-8 -*-
 #
-# {{ project }} documentation build configuration file, created by
-# sphinx-quickstart on {{ now }}.
+# Configuration file for the Sphinx documentation builder.
 #
 # This file does only contain a selection of the most common options. For a
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/stable/config
 
-# -- Path setup -----------------------------------------------------------
+# -- Path setup --------------------------------------------------------------
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
@@ -27,7 +26,7 @@ sys.path.insert(0, u'{{ module_path }}')
 {% endif -%}
 {% endif %}
 
-# -- Project information --------------------------------------------------
+# -- Project information -----------------------------------------------------
 
 project = u'{{ project_str }}'
 copyright = u'{{ copyright_str }}'
@@ -39,7 +38,7 @@ version = u'{{ version_str }}'
 release = u'{{ release_str }}'
 
 
-# -- General configuration ------------------------------------------------
+# -- General configuration ---------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
 #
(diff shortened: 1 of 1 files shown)


sphinx/templates/quickstart/conf.py_t
```diff
@@ -1,13 +1,12 @@
 # -*- coding: utf-8 -*-
 #
-# {{ project }} documentation build configuration file, created by
-# sphinx-quickstart on {{ now }}.
+# Configuration file for the Sphinx documentation builder.
```
tk0miya, line 3: I think ` for Sphinx` is simple and better. What do you think?

tk0miya (approved):

jfbu:
Leaving in conf.py the release number of Sphinx at time of sphinx-quickstart execution would be useful.

sphinx/templates/quickstart/conf.py_t
timhoffm: I've added the "documentation builder" part to give people unfamiliar with Sphinx a hint to the real purpose of the file (configure the documentation builder). I can also remove it. Either is fine by me.

timhoffm:
@jfbu It was argued in #4385 that the version would not be relevant.

> The generated conf.py might not depend on specific version of Sphinx (maybe). So it is not important.
And user can declare required version of Sphinx with needs_sphinx config implicitly. So I think it is not needed by default.

What's the motivation to include it?

jfbu:
sorry I had not followed in details. Thinking twice I agree it is not important. Maybe I was thinking too much of my library of test projects... it is useful for me to have some idea for what version I first created the thing. Naturally, I only have to discipline myself into adding it manually.

tk0miya:
It is better to use `needs_sphinx`, `requirements.txt`, `setup.py` or other tools (like docker) to define your dependencies. I think the comments does not help users.

Basically, user does not re-create `conf.py` for existing project and modify it manually. So it might not be updated the date even if an user users a new feature of Sphinx. So I think it does not help them.

sphinx/templates/quickstart/conf.py_t
tk0miya: Either is fine to me too. So go as you like it.

I guess the documents are usually placed at `docs` directory or `foobar-docs` repository. So users might infer it's configuration file for documentation tooling (am I optimistic?).

timhoffm: In that case, I slightly perfer the way I've originally written it. I'll leave it as is. By me, the PR is ready to be merged.

tk0miya:

Issue #4223, opened 2017-11-07
sphinx.ext.autodoc breaks reporting of sphinx.ext.doctest


Zac-HD:
Subject: `sphinx.ext.autodoc` breaks reporting of failing examples for `sphinx.ext.doctest`

### Problem
If you are using autodoc and doctest in the same project, doctests in docstrings are reported as existing in the file they are included in, rather than their actual location on disk.  This makes fixing failed doctests much harder than it should be!

To make things more difficult, the reported line number is also wrong.  Instead of being the location in the file, it appears to be the location in the docstring - plus two!  This makes it possible to have any number of completely different tests reported as failing in exactly the same place.

### Procedure to reproduce the problem
I wrote a tiny example that shows this problem: [example.zip](https://github.com/sphinx-doc/sphinx/files/1448833/example.zip) (total of 15 lines across `conf.py`, `index.rst`, and `foo.py`)

`sphinx-build -b doctest . _build` with attached example.  

Note that the example is reported as failing in `File "index.rst", line 3, in default`, when it's actually in `foo.py` at line 2.  (and this isn't an off-by-one error; try adding empty lines before the docstring in `foo.py` or `.. automodule::` directive in `index.rst`)

I've run into this several times before, but HypothesisWorks/hypothesis-python#962 motivated me to finally write up a useful report.

### Expected results
Report the failing doctest in the correct file (`foo.py`), at the correct file-relative line number (2).

It would be even better if the report could include an absolute or relative path, rather than an bare filename, but reporting the *correct* filename would be a good start!

### Environment info
- OS: Windows 7/8/10, Linux (Ubuntu, Centos, etc)
- Python version: 3.6.1 (and 3.5.x, 2.7.x, etc)
- Sphinx version: 1.6.4 (and earlier versions)

Zac-HD:
Further investigation suggests that the problem is in `ext.doctest`, assuming that [`doc2path`](http://www.sphinx-doc.org/en/stable/extdev/envapi.html#sphinx.environment.BuildEnvironment.doc2path) is an appropriate way to describe which document we're in.  This is incorrect when `node.source` points to a different file, eg via autodoc or include directives.

It's not yet obvious where best to make the change, but I'm getting a better sense of how nodes, TestCases, and other data are passed around.  Hopefully with some more work an elegant fix will be revealed.

Zac-HD:
[...]

PR #4243, opened 2017-11-16
Fixes #4223: Report failing doctests in correct file at correct line

Zac-HD:
### Purpose
Giving the wrong location for a failing test is not good.  When humans are running the test we can usually find the true location reasonably quickly and remember it, but bots aren't so great.  Reporting an accurate location will make it far easier and more reliable to automatically fix failing doctests in downstream projects.

### Detail
- When doctests are included in a file via `sphinx.ext.autodoc`, the failure message points to the actual source file instead of the document file.
- Auto-doc'd doctests *also* report their line number relative to start of file, instead of relative to `inspect.cleandoc` of the runtime docstring.
- Compensated for an off-by-one error in line number for all other doctests (hackish, because I couldn't find the origin)
- Added tests so that later changes (including better fixes) don't break doctest reporting again (at least in the same way)
- One commit of unrelated cleanup, where I did some refactoring while trying to trace the problem around.  No change in behaviour, just easier to read.

### Relates
- #4223, HypothesisWorks/hypothesis-python#962

PR #4428, opened 2018-01-14
Regression test for doctest reporting

Zac-HD:
This pull replaces #4243, and closes issue #4223. 

Well, mostly - it turns out the underlying issue has been fixed on the master branch in the last four months, so really all I'm adding is the regression tests I wrote and related code cleanup.  CC  @tk0miya - this should be much faster to review than expected!

I have added the reporting issue to the changelog, even though this pull is no longer what fixed it, and my name to the authors list for the tests (hopefully that's enough).


Zac-HD:
Oh dear - the underlying problem is fixed in my local environment, but *not* in CI.  I think I'll need to reapply the patches, which means checking again to see if I can fix the underlying issue instead of applying patches drawn from #4243.  ETA another day or two - let me know if the schedule is tight and I'll prioritise.

tk0miya:
It seems tests are failed. Could you check this please?

```
E           assert 'File "dir/inner.rst", line 1, in default' in ['File "dir/inner.rst", line 2, in default', 'File "dir/inner.rst", line 2, in default', 'File "index.rst", line 2, in default', 'File "index.rst", line 5, in default']
```
https://travis-ci.org/sphinx-doc/sphinx/jobs/328654000#L1801

Note: I think this is not a new feature, so we can merge this after feature freeze.

Zac-HD:
Yeah, that's the bug - the doctests are actually in four files not two, and not at those line numbers in any case 😭

Next step, I'll try again to reproduce locally...  see my comment above.

tk0miya:
Oh, I'd overlooked your last comment. I can't say X-day at this moment because I don't see status of other issues now. But I think this is not a new feature and it seems not breaking changes.
So please go at your pace.

sphinx/environment/__init__.py
```diff
@@ -358,17 +357,16 @@ def merge_info_from(self, docnames, other, app):
...
         """
         if filename.startswith(self.srcdir):
-            filename = filename[len(self.srcdir) + 1:]
+            filename = os.path.relpath(filename, self.srcdir)
         for suffix in self.config.source_suffix:
-            if fnmatch.fnmatch(filename, '*' + suffix):
-                return filename[:-len(suffix)]
-        # the file does not have docname
+            if filename.endswith(suffix):
+                return filename.rstrip(suffix)
```
tk0miya: I noticed this is not good. `str.rstrip()` is not alternative of `filename[:-len(suffix]`.
```
>>> "rrrr.rst".rstrip(".rst")
''
```

sphinx/environment/__init__.py
```diff
@@ -358,17 +357,16 @@ def merge_info_from(self, docnames, other, app):
         app.emit('env-merge-info', self, docnames, other)
 
     def path2doc(self, filename):
-        # type: (unicode) -> unicode
+        # type: (unicode) -> Optional[unicode]
```
tk0miya, line 352: why is `filename` optional?

Zac-HD: That's the return value, not the `filename` argument - and `Optional[unicode]` means that it may return a unicode string *or* `None` (if no suffix matches).

tk0miya: oh, sorry. Please forget my comment...

Zac-HD:
Alright, I'm fairly happy with this now.  There are four commits:

- Add tests that doctests report their locations correctly, with real filenames and line numbers
- Tangential refactoring (I was in the area...)
- A nice, principled fix - our reported filename is now based on the block source, rather than the output document
- Paper over the problems with line numbers, by setting to `None` for doctests or subtracting to cancel an off-by-one-error, and making the test weaker.

I'm not delighted by the last bit, but it's still a strict improvement on the status quo and I don't have time to investigate deeply again before the deadline for 1.7.  Assuming that this pull passes tests and review (cc @tk0miya), I'd therefore suggest merging it and leaving a "proper fix" for follow-up work.

Zac-HD:
Ping @tk0miya; I think this is ready to merge now 😄

sphinx/ext/doctest.py
```diff
@@ -366,6 +367,39 @@ def write(self, build_docnames, updated_docnames, method='update'):
...
             self.test_doc(docname, doctree)
 
+    def get_filename_for_node(self, node, docname):
+        # type: (nodes.Node, unicode) -> str
+        """Try to get the file which actually contains the doctest, not the
+        filename of the document it's included in."""
+        try:
+            filename = (
+                path.relpath(node.source, self.env.srcdir)
+                .replace(path.sep, '/')
```
tk0miya, line 377: Why do you replace separators by `/`?

sphinx/ext/doctest.py
```diff
@@ -366,6 +367,39 @@ def write(self, build_docnames, updated_docnames, method='update'):
...
+                .rsplit(':docstring of ', maxsplit=1)[0]
+            )
+        except Exception:
+            filename = self.env.doc2path(docname, base=None)
+        if PY2:
+            return filename.encode(fs_encoding)
```
tk0miya, line 383: Do you know why is this needed? I think this is not needed, and this should return unicode path string.

CHANGES
```diff
@@ -243,6 +243,7 @@ Bugs fixed
 * #4198: autosummary emits multiple 'autodoc-process-docstring' event. Thanks
   to Joel Nothman.
 * #4081: Warnings and errors colored the same when building
+* #4223: doctest: failing tests reported in wrong file, at wrong line
```
tk0miya, line 246: Please move this to 1.7.0b3 section.

tests/test_ext_doctest.py
```diff
@@ -55,3 +56,27 @@ def test_is_allowed_version():
...
+        'File "index.rst", line 4, in default',
+    ]
+    for location in expected:
+        assert location in failures
+
+
+if PY2:
+    test_reporting_with_autodoc = pytest.mark.xfail(
+        reason='node.source points to document (not filename) under Python 2'
+    )(test_reporting_with_autodoc)
```
tk0miya, line 82: Please use decorator style instead.

You can give a condition to `xfail` as an argument.
https://docs.pytest.org/en/2.7.3/skipping.html#mark-a-test-function-as-expected-to-fail

tk0miya (changes requested):
LGTM with nits.

And could you rebase this to 1.7-release branch?
Then I will merge this into 1.7.0 soon.

PR #4429, opened 2018-01-14
Remove unneeded coding from LaTeXTranslator.visit_figure()

jfbu:
Simple trimming of unneeded conditionals

PR #4202, opened 2017-10-31
Emit warning if both tabularcolumns and :widths: are given (refs: #4196)

tk0miya:
refs: #4196

PR #4430, opened 2018-01-14
Add :confval:`latex_use_wrapfig`

jfbu:
This builds upon #4426 (and incorporates #4429).

However I am not satisfied with this, mainly because most projects using figures will now receive many warnings when trying to build PDF (from cherry-picked #4426).

I don't know good solution yet. It would be better if after first warning next ones would be less verbose and just tell where in sources is the problem, and user could fix it by adding a new figure option ``:nowrapfig:`` if needed to not use wrapfig, or to use wrapfig ``:nowrapfigwarn:`` with no warning.

Thus, this PR is for discussion.

both

- Feature
- Bugfix

refs: #3289

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1711,14 +1711,35 @@ class LaTeXTranslator(nodes.NodeVisitor):
                 self.body.append('\\capstart')
             self.context.append(labels + '\\end{sphinxfigure-in-table}\\relax\n')
         elif node.get('align', '') in ('left', 'right'):
-            length = None
-            if 'width' in node:
-                length = self.latex_image_length(node['width'])
-            elif 'width' in node[0]:
-                length = self.latex_image_length(node[0]['width'])
-            self.body.append('\\begin{wrapfigure}{%s}{%s}\n\\centering' %
-                             (node['align'] == 'right' and 'r' or 'l', length or '0pt'))
-            self.context.append(labels + '\\end{wrapfigure}\n')
+            if self.builder.config.latex_use_wrapfig:
+                length = None
+                if 'width' in node:
+                    length = self.latex_image_length(node['width'])
+                elif 'width' in node[0]:
+                    length = self.latex_image_length(node[0]['width'])
+                self.body.append('\\begin{wrapfigure}{%s}{%s}\n\\centering' %
+                                 (node['align'] == 'right' and 'r' or 'l',
+                                  length or '0pt'))
+                self.context.append(labels + '\\end{wrapfigure}\n')
+
+                # emit an information about wrapfig usage (refs: #3289)
+                logger.info(__('A figure requiring left or right alignment '
+                               'has been detected. LaTeX uses wrapfig macros '
+                               '(cf. ``latex_use_wrapfig`` config variable), '
+                               'please check layout in generated PDF.'),
+                            location=node, type='latex')
+            else:
+                if self.in_minipage:
+                    self.body.append('\n\\begin{flush%s}' % node.attributes['align'])
+                    self.context.append('\\end{flush%s}\n' % node.attributes['align'])
+                else:
+                    self.body.append('\n\\begin{figure}[%s]\\begin{flush%s}\n' %
+                                     (self.elements['figure_align'],
+                                      node.attributes['align']))
+                    if any(isinstance(child, nodes.caption) for child in node):
+                        self.body.append('\\capstart\n')
+                    self.context.append(labels + '\\end{flush%s}\\end{figure}\n' %
+                                        node.attributes['align'])
         elif self.in_minipage:
             self.body.append('\n\\begin{center}')
             self.context.append('\\end{center}\n')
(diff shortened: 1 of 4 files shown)


doc/config.rst
```diff
@@ -1753,6 +1754,15 @@ These options influence LaTeX output. See further :doc:`latex`.
 
    .. versionadded:: 1.6
 
+.. confval:: latex_use_wrapfig
+
+   The default is ``True``: figures with ``:align:`` option specifying to use
+   left or right alignment will be rendered using ``wrapfig`` macros. But,
+   this is known to have layouting issues in some cases. Set the option to
+   ``False`` to avoid usage of ``wrapfig``.
```
jfbu: clearly it is bad to have only a global setting. The choice should be made per figure, default being to use wrapfig and only in case of detected problems to not use it.

But it is impossible for Sphinx to detect problem, only to detect potential problem: if the figure follows immediately or precedes a list in particular.

sphinx/writers/latex.py
```diff
@@ -1844,37 +1844,43 @@ def visit_figure(self, node):
...
+                    self.body.append('\n\\begin{flush%s}' % node.attributes['align'])
+                    self.context.append('\\end{flush%s}\n' % node.attributes['align'])
+                else:
+                    self.body.append('\n\\begin{figure}[%s]\\begin{flush%s}\n' %
+                                     (self.elements['figure_align'],
+                                      node.attributes['align']))
+                    if any(isinstance(child, nodes.caption) for child in node):
+                        self.body.append('\\capstart\n')
+                    self.context.append(ids + '\\end{flush%s}\\end{figure}\n' %
+                                        node.attributes['align'])
```
jfbu, line 1742: the above code lines are ressurection of what #4429 removed ...

doc/config.rst
tk0miya: I think reST is a kind of lightweight mark up. Mainly it is marked up text. So I'd like not to add layout notations and directive options to reST.
Anyway this global option would help who has layouting problem. so +1 for this.

sphinx/writers/latex.py
```diff
@@ -1844,37 +1844,43 @@ def visit_figure(self, node):
...
+                    length = self.latex_image_length(node['width'])
+                elif 'width' in node[0]:
+                    length = self.latex_image_length(node[0]['width'])
+                self.body.append('\\begin{wrapfigure}{%s}{%s}\n\\centering' %
+                                 (node['align'] == 'right' and 'r' or 'l',
+                                  length or '0pt'))
+                self.context.append(ids + '\\end{wrapfigure}\n')
+
+                # emit a warning about wrapfig (refs: #3289)
+                logger.warning(_('A figure requiring left or right alignment '
```
tk0miya: Please use double underscore method instead. It's for messaging.
The single one is used for a part of document.

tk0miya: Is this better to use info level message like #4202?

tk0miya (commented):
LGTM with nits

tk0miya (approved):

doc/config.rst
jfbu: I propose to add a LaTeX boolean which can be inserted by user to fix only one figure

```rst
.. raw:: latex

   \sphinxusewrapfigfalse

.. figure:: foo.png
   :align: right

.. raw:: latex

   \sphinxusewrapfigtrue
```

what do you think? I know it requires user to realize PDF had layout problem and then do the manual intervention.

jfbu: ah sorry, I need to think more. This method needs sphinx to provide mother environment which expands either to figure or wrapfigure.

sphinx/writers/latex.py
jfbu: I think info level indeed is better. I will do it.

doc/config.rst
tk0miya: I feel using `raw` decreases the readability and writability.
So I choose the directive option rather than it.

Anyway, this discussion is not related with this PR directly. So let's file a new issue.

sphinx/writers/latex.py
```diff
@@ -1844,37 +1844,44 @@ def visit_figure(self, node):
...
+                                 (node['align'] == 'right' and 'r' or 'l',
+                                  length or '0pt'))
+                self.context.append(ids + '\\end{wrapfigure}\n')
+
+                # emit an information about wrapfig usage (refs: #3289)
+                logger.info(__('A figure requiring left or right alignment '
+                               'has been detected. LaTeX uses wrapfig macros '
+                               '(cf. ``latex_use_wrapfig`` config variable), '
+                               'please check layout in generated PDF.'),
+                            location=node, type='latex')
```
jfbu, line 1730: I have replaced warning by less intrusive info. In my testing it does not report source file and line which limits usability. Is there some user option? (time I read our docs ``:)`` )

jfbu (commented):
Replaced warning by info

sphinx/writers/latex.py
tk0miya: At current implementation, info message does not have location. So `location` argument is ignored.

tk0miya:
ping. any updates?

jfbu:
Rebased upon current master.

jfbu:
I closed this while doing some clean-up of my stale branches. I will re-open when I will have had time to re-examine the situation and to finish the PR.

PR #4433, opened 2018-01-14
Fix #4431: "Undefined control sequence \\hlineSomeText"

### LINKED ISSUE #4431: latex "Undefined control sequence \\hlineSomeText" for make in latex output
The reason is in line 1476 of writers/latex.py:

```
            self.body.append('\\hline')
````

A space needs to be added to ``\\hline``.

jfbu:
I changed `\hline` to `\hline<space>` also in the templates, but that is only to ease up mass updates to tests... a specific test is added, but the real fix of #4431 should modify that probably.


tk0miya (approved):
LGTM

jfbu:
Closing because this is only workaround and does not address core issue.

Follow-up at https://github.com/sphinx-doc/sphinx/issues/4431#issuecomment-358588311

PR #4436, opened 2018-01-15
app: Centralize directory validation

stephenfin:
This allows us to avoid duplication of code and ensure validation happens regardless of who's initializing the class.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -28,7 +28,8 @@ from docutils.parsers.rst import directives, roles
 import sphinx
 from sphinx import package_dir, locale
 from sphinx.config import Config
-from sphinx.errors import ConfigError, ExtensionError, VersionRequirementError
+from sphinx.errors import (
+    ApplicationError, ConfigError, ExtensionError, VersionRequirementError)
 from sphinx.deprecation import RemovedInSphinx20Warning
 from sphinx.environment import BuildEnvironment
 from sphinx.events import EventManager
@@ -43,6 +44,7 @@ from sphinx.util.osutil import ENOENT, ensuredir
 from sphinx.util.console import bold  # type: ignore
 from sphinx.util.docutils import is_html5_writer_available, directive_helper
 from sphinx.util.i18n import find_catalog_source_files
+from sphinx.util.osutil import abspath
 
 if False:
     # For type annotation
@@ -123,10 +125,24 @@ class Sphinx(object):
         self.enumerable_nodes = {}              # type: Dict[nodes.Node, Tuple[unicode, Callable]]  # NOQA
         self.html_themes = {}                   # type: Dict[unicode, unicode]
 
-        self.srcdir = srcdir
+        # validate provided directories
+        self.srcdir = abspath(srcdir)
+        self.outdir = abspath(outdir)
+        self.doctreedir = abspath(doctreedir)
         self.confdir = confdir
-        self.outdir = outdir
-        self.doctreedir = doctreedir
+        if self.confdir:  # confdir is optional
+            self.confdir = abspath(self.confdir)
+            if not path.isfile(path.join(self.confdir, 'conf.py')):
+                raise ApplicationError("config directory doesn't contain a "
+                                       "conf.py file (%s)" % confdir)
+
+        if not path.isdir(self.srcdir):
+            raise ApplicationError('Cannot find source directory (%s)' %
+                                   self.srcdir)
+
+        if self.srcdir == self.outdir:
+            raise ApplicationError('Source directory and destination '
+                                   'directory cannot be identical')
 
         self.parallel = parallel
 
--- a/sphinx/setup_command.py
+++ b/sphinx/setup_command.py
@@ -148,16 +148,12 @@ class BuildDoc(Command):
         if self.build_dir is None:
             build = self.get_finalized_command('build')
             self.build_dir = os.path.join(abspath(build.build_base), 'sphinx')  # type: ignore
-            self.mkpath(self.build_dir)  # type: ignore
 
         self.doctree_dir = os.path.join(self.build_dir, 'doctrees')
 
-        self.mkpath(self.doctree_dir)  # type: ignore
         self.builder_target_dirs = [
             (builder, os.path.join(self.build_dir, builder))
             for builder in self.builder]  # type: List[Tuple[str, unicode]]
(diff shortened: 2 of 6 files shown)


sphinx/application.py
```diff
@@ -123,10 +124,24 @@ def __init__(self, srcdir, confdir, outdir, doctreedir, buildername,
         self.enumerable_nodes = {}              # type: Dict[nodes.Node, Tuple[unicode, Callable]]  # NOQA
         self.html_themes = {}                   # type: Dict[unicode, unicode]
 
-        self.srcdir = srcdir
+        # validate provided directories
+        self.srcdir = abspath(srcdir)
+        self.outdir = abspath(outdir)
+        self.doctreedir = abspath(doctreedir or path.join(outdir, '.doctrees'))
```
tk0miya: IMO, it is better that `Sphinx` class simply uses given `doctreedir` argument.

sphinx/application.py
```diff
@@ -123,10 +124,24 @@ def __init__(self, srcdir, confdir, outdir, doctreedir, buildername,
...
-        self.outdir = outdir
-        self.doctreedir = doctreedir
+        if self.confdir:
+            self.confdir = abspath(self.confdir)
+            if not path.isfile(path.join(self.confdir, 'conf.py')):
+                raise ConfigError("config directory doesn't contain a "
+                                  "conf.py file (%s)" % confdir)
+
+        if not path.isdir(self.srcdir):
+            raise ConfigError('Cannot find source directory (%s)' %
```
tk0miya: +1 for checking directory existence and regulate abspath at `Sphinx` class.
But I feel `ConfigError` is not good for this case. It is used for configuration error.
So these validation should raise other error.

sphinx/setup_command.py
```diff
@@ -138,20 +138,14 @@ def finalize_options(self):
...
-        self.config_dir = abspath(self.config_dir)
 
         self.ensure_string_list('builder')
         if self.build_dir is None:
             build = self.get_finalized_command('build')
             self.build_dir = os.path.join(abspath(build.build_base), 'sphinx')  # type: ignore
             self.mkpath(self.build_dir)  # type: ignore
-        self.build_dir = abspath(self.build_dir)
         self.doctree_dir = os.path.join(self.build_dir, 'doctrees')
+        # TODO(stephenfin): Is this necessary?
```
tk0miya: It is not needed. Builder class creates doctreedir if not exist.
https://github.com/sphinx-doc/sphinx/blob/a1f13e30000cdeceef7fc29b2bf3d0f45d5750ca/sphinx/builders/__init__.py#L85

sphinx/util/osutil.py
```diff
@@ -219,7 +221,11 @@ def abspath(pathdir):
     # type: (unicode) -> unicode
     pathdir = path.abspath(pathdir)
     if isinstance(pathdir, bytes):
-        pathdir = pathdir.decode(fs_encoding)
+        try:
+            pathdir = pathdir.decode(fs_encoding)
+        except UnicodeError:
+            raise ConfigError('multibyte filename not supported on this '
```
tk0miya: I must say it is very strange that `abspath()` raises `ConfigError`.

tk0miya (changes requested):
I agree with basic concept of this PR.

sphinx/util/osutil.py
stephenfin: @tk0miya I used this because it seemed like a configuration issue. However, it is not a configuration *file* issue. What would you suggest instead? Remember that we'll raise this exception for any and all issues with provided paths.

sphinx/application.py
stephenfin: OK. I think my reasoning for doing this was that the pattern was duplicated in two places. That's probably not a bad thing though. Perhaps I should raise an exception if these are unset?

sphinx/application.py
stephenfin: This is different to the issue below. IMO, the source directory is part of the configuration. For example, calling `sphinx-build SRCDIR OUTDIR` *configures* Sphinx to read input from `SRCDIR` and write output to `OUTDIR`. However, as noted below, perhaps you think that `ConfigError` is only for configuration file errors? If so, perhaps I should add a new `ConfigError` exception and rename the existing one `ConfigFileError`? The renamed exception can subclass the new one.

sphinx/setup_command.py
stephenfin: I'll fix this in the next version

sphinx/application.py
tk0miya: Yes, I think so. I think new responsibility of `Sphinx` class is regulating paths and checking their existence (validation). So exceptions should be raised if invalid parameter given.

I think reducing code duplication is not important. IMO, it is more important to clear responsibility of each modules. Before, `Sphinx` class and its clients are strongly coupled. `Sphinx` always trusts arguments from clients and does not check them.
With this change, `Sphinx` class becomes defensive. It also means less coupled.

Note: I think choosing `doctreedir` is a responsibility of clients. It's simple.

sphinx/application.py
tk0miya: >However, as noted below, perhaps you think that ConfigError is only for configuration file errors?

Yes. I understand the command line arguments are part of configuration in a broad sense. But , in narrow sense, it is not. In a broad sense, almost of user inputs are "configuration". But it is too ambiguous.

To be clearly, I propose you to add new exception like `ArgumentError`. I prefer to use different name for these errors.

sphinx/util/osutil.py
tk0miya: No, this is not related with case of `Sphinx` class. this `abspath()` is an utility function. So it is used for other paths. So I think using standard exception is better. So I think it is better to raise `UnicodeError` directly.

stephenfin: Done

sphinx/setup_command.py
stephenfin: Done

sphinx/application.py
stephenfin: Done

sphinx/application.py
stephenfin: Done

tk0miya (approved):

tk0miya:
Thank you for update. LGTM!

PR #4445, opened 2018-01-17
[RFC] Add ability to format :doc:

mwoehlke:
Teach `:doc:` to format references, by replacing `{name}` and `{number}` in explicit link text, a la `:numref:`. This makes it possible to include the section number in a `:doc:` reference, which is not otherwise possible, since `:numref:` does not resolve document names.

This is a request for comment, as the PR is obviously incomplete in its current state (needs doc, possibly tests). Particularly, is this a useful feature and a reasonable implementation thereof? If so, what work needs to be done to make it merge-ready? (If it's interesting, I will make the necessary changes and update the PR.)

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -782,11 +782,15 @@ class StandardDomain(Domain):
         if docname not in env.all_docs:
             return None
         else:
+            title = clean_astext(env.titles[docname])
             if node['refexplicit']:
                 # reference with explicit title
                 caption = node.astext()
+                if '{name}' in caption or '{number}' in caption:
+                    number = env.toc_secnumbers[docname].get('')
+                    caption = caption.format(name=title, number=number)
             else:
-                caption = clean_astext(env.titles[docname])
+                caption = title
             innernode = nodes.inline(caption, caption, classes=['doc'])
             return make_refnode(builder, fromdocname, docname, None, innernode)
 


sphinx/domains/std.py
```diff
@@ -782,11 +782,15 @@ def _resolve_doc_xref(self, env, fromdocname, builder, typ, target, node, contno
         if docname not in env.all_docs:
             return None
         else:
+            title = clean_astext(env.titles[docname])
             if node['refexplicit']:
                 # reference with explicit title
                 caption = node.astext()
+                if '{name}' in caption or '{number}' in caption:
+                    number = env.toc_secnumbers[docname].get('')
```
tk0miya, line 790: Is this work fine with LaTeX output?
AFAIK, LaTeX manages section numbers by itself.

sphinx/domains/std.py
```diff
@@ -782,11 +782,15 @@ def _resolve_doc_xref(self, env, fromdocname, builder, typ, target, node, contno
         if docname not in env.all_docs:
             return None
         else:
+            title = clean_astext(env.titles[docname])
             if node['refexplicit']:
                 # reference with explicit title
                 caption = node.astext()
+                if '{name}' in caption or '{number}' in caption:
```
tk0miya, line 789: We have to discuss what keyword is best.
I feel `number` is already used in `:numref:` role. so it is good. But it is ambiguous to me in `:doc:` role.
I don't know which is better...

sphinx/domains/std.py
tk0miya, line 790: I think it is better to notice if the document does not have section numbers.

tk0miya (changes requested):
+0; I don't know this is needed or not (Personally, I prefer to use labels and `:ref:` instead). But I don't opposite this.

sphinx/domains/std.py
mwoehlke: I honestly have no idea; I don't use (or know how to use) LaTeX...

mwoehlke: ...do you mean to add a warning here if `number` is `None`? (Otherwise I don't think I understand your comment.)

sphinx/domains/std.py
mwoehlke: I think there are three options here:

1. Add substitutions to `:doc:`, as in this patch. This gives cause to wonder if `:ref:` should be enhanced similarly, which suggests that `:numref:` could be deprecated.
2. Introduce a new command that is to `:doc:` as `:numref:` is to `:ref:`. But this requires more logic vs. the first approach for (IMHO) dubious benefit.
3. Teach `:numref:` to also understand document names. But this might be the most difficult.

At any rate, I'm open to suggestions. What the substitution keywords are named is mostly unimportant, except that using the same ones as `:numref:` is less for users to learn/remember. (An alternative I would be okay with, if `:numref:` was also taught to handle it for section references, would be to use `{secnumber}`.)

tk0miya: Hmm? Only my concern is about substitution keywords of `:doc:` role.
The target of `:numref:` is a numbered object (ex. figure, table, section and so on). So `{number}` is understandable to me.
On the other hand, the target of `:doc:` is a document, not a section. So I feel `{number}` might not be good for `:doc:` a little.

But I agree using same keywords is easy to learn.

sphinx/domains/std.py
tk0miya: Hmm, I'll take a look if I have time.
LaTeX is important format for us. So we have to support it also.

tk0miya: That's right

mwoehlke: Agreed; it's just hard for me to implement because I have near-zero knowledge of LaTeX.

sphinx/domains/std.py
mwoehlke: Ah, I see. Well... the normal textual resolution of `:doc:` is the document title. And, at least when it appeared in a numbered `toctree` via the `index.rst`, that *does* have a number. It is specifically that number that I want from `{number}`. Thus, FWIW, it makes perfect sense to me, for my use case 🙂.

My original use case — part of it — is that I have a bunch of appendices, e.g. "Appendix I: Foo", "Appendix II: Bar", and I want to write a reference that displays as "Appendix II" rather than "Bar". (The appendix numbers are assigned via a numbered `:toctree:` so I don't need to change them if the order or number of appendices changes, and of course, I want the references to also be automatically in sync.)

(The other part is where I want the *full* page name, e.g. "Appendix I: Foo" rather than "Foo", which I can get with ``:doc:`{number}: {name} <appendix-foo>` ``.)

mtgstuber:
I'm looking for this exact feature.  See https://github.com/orgs/sphinx-doc/discussions/14588
It doesn't appear this was ever merged.  Any chance of something like this being supported?

Issue #654, opened 2015-01-02
Headline inconsistencies when chanf between use parts or normal


shimizukawa:
Hi when i change to using parts in latex

latex_use_parts = True;

Then headlines are treated differently.

when using parts the headline.

```
#!rst
===============
A Headline
===============
```

is seen as the same level as

```
#!rst
Another Headline
================
```

Both headlines is seen as parts, meaning i have like 20+ parts
when i change from using chapters to using parts.

when using chapters only i would get main headlines from

```
#!rst
==================
Main headline
==================
```

And the next level would be this

```
#!rst
Next highest headline level
=================
```

So i have to change a lot of headline underline characters
just because i change between Parts and Chapters. 

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/654
- Originally reported by: Michael Kell Jensen
- Originally created at: 2011-03-12T16:32:50.410

shimizukawa:
_From Georg Brandl on 2011-09-23 08:51:46+00:00_

I cannot reproduce this. Can you make a small example project to demonstrate?

shimizukawa:
_From Georg Brandl on 2014-10-18 07:32:19+00:00_

Removing version: 1.0.7 (automated comment)

PR #4449, opened 2018-01-17
fix ext.apidoc to include "empty" packages that contain modules

dhellmann:
Commit 2d99648e9982325bbd670da11df5f809e3134284 changed the apidoc
extension to ignore packages if the __init__.py file was empty. That
breaks the toctree structure if those packages do contain submodules
and subpackages. This patch adds a check to ensure that empty
__init__.py modules are only skipped if there are no other python
modules in the same directory.

Addresses bug #654

Signed-off-by: Doug Hellmann <doug@doughellmann.com>

### Feature or Bugfix

- Feature
- Bugfix

### Purpose

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>

### DIFF
--- a/sphinx/apidoc.py
+++ b/sphinx/apidoc.py
@@ -16,6 +16,7 @@
 """
 from __future__ import print_function
 
+import glob
 import os
 import sys
 import optparse
@@ -194,7 +195,17 @@ def shall_skip(module, opts):
 
     # skip it if there is nothing (or just \n or \r\n) in the file
     if path.exists(module) and path.getsize(module) <= 2:
-        return True
+        skip = True
+        if os.path.basename(module) == '__init__.py':
+            pattern = path.join(path.dirname(module), '*.py')
+            # We only want to skip packages if they do not contain any
+            # .py files other than __init__.py.
+            other_modules = list(glob.glob(pattern))
+            other_modules.remove(module)
+            skip = not other_modules
+
+        if skip:
+            return True
 
     # skip if it has a "private" name and this is selected
     filename = path.basename(module)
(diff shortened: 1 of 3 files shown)


sphinx/apidoc.py
```diff
@@ -194,7 +195,14 @@ def shall_skip(module, opts):
...
     if path.exists(module) and path.getsize(module) <= 2:
-        return True
+        skip = True
+        if module.endswith('__init__.py'):
+            pattern = path.join(path.dirname(module), '*.py')
+            # We only want to skip packages if they do not contain any
+            # .py files other than __init__.py.
+            other_modules = list(glob.glob(pattern))
+            other_modules.remove(module)
+            skip = not other_modules
```
tk0miya, line 205: `skip` is not used.  I guess `return skip` is needed here.

sphinx/apidoc.py
```diff
@@ -194,7 +195,14 @@ def shall_skip(module, opts):
 
     # skip it if there is nothing (or just \n or \r\n) in the file
     if path.exists(module) and path.getsize(module) <= 2:
-        return True
+        skip = True
+        if module.endswith('__init__.py'):
```
tk0miya: Please use `os.path.basename(module) == '__init__.py'`. `endswith` is not good.

tk0miya (changes requested):
Could you add a test for this if possible?

sphinx/apidoc.py
dhellmann: Done

sphinx/apidoc.py
dhellmann: Done

tk0miya (approved):

tk0miya:
Merged.
Thank you for your help!

dhellmann:
Thanks!

I'm not familiar with the development workflow for sphinx. Do you still want #4450 for the master branch?

tk0miya:
Don't mind. I'll take a look later :-)

PR #3435, opened 2017-02-20
Fix #1574 via improvement to latex tables (allow general contents in merged cells)

jfbu:
Related: #2620, #1574, #3288

Main purpose of this PR is to improve table cells containing paragraphs or other formatted contents like lists. So far, lists were not possible. And paragraphs had a tendancy to spill out of the table.

- [x] make lists and multi-paragraph contents possible

- [x] update tests

- [x] make code-blocks work in multirow cells

- [x] make code-blocks work in multicolumn cells

- [x] make code-blocks work in multirow-multicolumn cells !

- [x] check that things working previouly are not broken. I have not done that extensively, though.

- [x] try to document what works. It appears things acceptable in regular cells are now also acceptable in merged cells of grid table. But the documentation did not say explicitely what worked and what did not work earlier, hence I can not say exactly. My feeling is that whenever something works in a normal table cell, it should now also work in a merged cell of a grid table. They are some know issues, I will mention them in a comment in this thread.

I had another task, but I drop it for the time being for lack of time.

> examine to what extent the ``has_problematic`` is still relevant and whether ``tabulary`` could be used in situations where it currently isn't.

### Feature or Bugfix
both

### Detail
The LaTeX details are a bit involved

- the ``multirow`` package has a bug not allowing empty lines in input. Furthermore, its `*` specifier always renders things horizontally making it interact very badly with `tabulary`. 2016 version has a `=` specifier (does not fix the bug about not allowing empty lines), but not all issues are solved. In the end I decided to write from scratch specific ``\sphinxmultirow``. Thanks to recent reformatting of table code by @tk0miya I could exploit the numbering on Sphinx side of "cells" and transfer it to LaTeX side to ensure uniformity of row heights in a multirow.

- the `tabulary` package has a bug in its definition of `\multicolumn` : like the bug of `multirow` blank lines are forbidden. Thus I added a patch to fix that.

- I don't use anymore `eqparbox` package because the problem is to control heights not widths. I use `varwidth`  package which provides an environment like `minipage` but for which the specified width acts as a maximum value. I encountered a problem which is that `varwidth` is not hyperref aware. Thus it worked sort of and emitted warnings. It seems I have found a way to add suitable macros to fix that.

The main change in behaviour is that theoretically multi-paragraph contents is not going to spill over bottom of table. Further, in a multirow, each row is assigned the same height. This of course change the looks of the table.

[...]

PR #4452, opened 2018-01-18
Fix "Undefined control sequence \\hlineSomeText" (#4431; refs: #3435)

jfbu:
Actually, the cause is some oversight of mine at #3435 which kept a test for cells with three or more paragraphs whereas original did it only for merged cells.

Also our tests with tables (although numerous) did not have a single example of a non-merged cell in first column with three or more paragraphs...

In my testing simply removing the two bad lines fixes it, and I could not find bad effect.


tk0miya (approved):
LGTM. Good work!

jfbu:
Thanks for review @tk0miya, merging.

PR #4458, opened 2018-01-19
Drop unused method; LoggingReporter.set_source()

tk0miya:
(T/O)

### DIFF
--- a/sphinx/util/docutils.py
+++ b/sphinx/util/docutils.py
@@ -183,21 +183,6 @@ class LoggingReporter(Reporter):
         stream = WarningStream()
         Reporter.__init__(self, source, report_level, halt_level,
                           stream, debug, error_handler=error_handler)
-        self.source_and_line = None  # type: SphinxFileInput
-
-    def set_source(self, source):
-        # type: (SphinxFileInput) -> None
-        self.source_and_line = source
-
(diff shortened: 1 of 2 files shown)


sphinx/util/docutils.py
```diff
@@ -183,21 +183,6 @@ def __init__(self, source, report_level, halt_level,
...
                           stream, debug, error_handler=error_handler)
-        self.source_and_line = None  # type: SphinxFileInput
-
-    def set_source(self, source):
-        # type: (SphinxFileInput) -> None
-        self.source_and_line = source
-
-    def system_message(self, *args, **kwargs):
-        # type: (Any, Any) -> Any
-        if kwargs.get('line') and isinstance(self.source_and_line, ViewList):
```
tk0miya, line 194: `self.source_and_line` is an instance of `SphinxFileInput` (see typehint of `set_source()` method).
So this block is not working. And `set_source()` is also meaningless...

This block tries to modify the location of system_messages (info and warnings). But Sphinx-1.7 uses `StringList` hack between `SphinxRSTFileInput` and `RSTParser` (in detail, please read the docstring of SphinxRSTFileInput`). So these are not used finally.

Note: these implementations are added during 1.7 development and they are not included in stable. so they are not released.

PR #4463, opened 2018-01-20
Drop special support for rst2pdf

tk0miya:
For some historical reason, we give a special treatment to rst2pdf.
But it is not maintained for long time, and it does not support python3.
So I think it's time to drop these special treatment. And I also dropped the mentions for rst2pdf on our docs.

This does not mean rst2pdf does not work with Sphinx.


sphinx/registry.py
```diff
@@ -332,8 +332,6 @@ def load_extension(self, app, extname):
 
         if metadata is None:
             metadata = {}
-            if extname == 'rst2pdf.pdfbuilder':
-                metadata['parallel_read_safe'] = True
```
tk0miya, line 336: With this removal, `rst2pdf` will be warned "the rst2pdf extension does not declare if it is safe for parallel reading" if users use `-j` option.

stephenfin (approved):
Yup, totally agree.

shimizukawa (approved):
LGTM!

tk0miya:
Thank you for reviewing! I'll merge this after sending a PR to rst2pdf project tomorrow.

tk0miya:
Just posted a PR: https://github.com/rst2pdf/rst2pdf/pull/640

PR #4465, opened 2018-01-20
Improve config

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add `config-inited` event
- Add `sphinx.config.Any` to represent the config value accepts any type of value

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -213,6 +213,7 @@ class Sphinx(object):
 
         # now that we know all config values, collect them from conf.py
         self.config.init_values()
+        self.emit('config-inited', self.config)
 
         # check extension versions if requested
         verify_required_extensions(self, self.config.needs_extensions)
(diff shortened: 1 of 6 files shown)


shimizukawa (approved):
I love it ;)
I think it is a good change to promote loose coupling for those who want to validate the setting values..

sphinx/application.py
```diff
@@ -225,6 +226,7 @@ def __init__(self, srcdir, confdir, outdir, doctreedir, buildername,
         self.builder = self.create_builder(buildername)
         # check all configuration values for permissible types
         self.config.check_types()
+        self.emit('config-inited', self.config)
```
tk0miya: This event must be emitted just after `config.init_values()`.

With current implementation, builders can touch config values before `config-inited` handlers invoked.

tk0miya:
Merged. thank you for reviewing!

PR #4475, opened 2018-01-21
[RFC] Implement delayed resolution in TOC

### LINKED ISSUE #4448: External references are not resolved in TOC tree sidebar
Let's say I have a document like:

```rest
Title
=====

:doc:`some-page`
----------------

This is a section (but not the page title)
whose name is the same as,
and incidentally also a link to,
another piece of documentation.
I used ``:doc:`` here,
but at least ``:ref:``
|--| very possibly everything that uses ``XRefRole``
and the delayed resolution thereof |--|
shows the same problem.
```

In the *body* of the document, the reference title is resolved as expected. (Let's say for example that the document title is "Describe Some Page Here". Then, in the body, the heading appears as "Describe Some Page Here".) However, in the sidebar TOC, the heading appears as "some-page". That is, the TOC uses the raw text of the reference, rather than the resolved text. Besides looking ugly, this is confusing and almost certainly not intended, as it means that the heading text in the TOC and the document body are different.

On consideration, I suspect the problem is that the TOC is using `astext` on `XRefRole` nodes, which drops the external reference bits before they can be resolved. This may be a tricky fix, as I suspect that, besides tweaking the TOC to allowed delayed text resolution (which in itself may be non-trivial), all of `StandardDomain.resolve_xref` may need to sprout a mode to resolve *text* but not actually emit a reference (since in the TOC case, the reference we want is to the section that *used* the external reference in its title, and not what the external reference actually references).

This should be reproducible with the following `index.rst`::

```rest
Hello, World
============

:doc:`index`
------------
```

(Verified by making a comparable change to the `index.rst` of a sphinx-quickstart project.)

mwoehlke:
Implement a new node that converts its contents to text (via `astext()`), but delays doing so until the post-transform stage. Modify TOC collection to use this. Modify TOC extraction to perform post-transforms (specifically, reference resolution and resolution of the aforementioned new nodes).

This fixes an annoying behavior where the use of an external reference in a section name would use the raw reference text rather than the resolved reference text in the sidebar table of contents, resulting in a mismatch between the section title as it appears in said table of contents and how it appears in the document body.

Fixes #4448.

I'm labelling this as RFC for now because it has a couple of potential issues, in addition to the question of whether this is even a reasonable approach:

1. It wasn't immediately obvious to me how to extract the document name for `_entries_from_toctree`. (OTOH, this looks like it might be dead code?)
2. Currently, I am creating a `pending_xtext` for ALL TOC entries, when probably this should only be done if the entry contains an xref. (This should be an easy fix, which I am happy to address if the patch is otherwise acceptable.)

### DIFF
--- a/sphinx/addnodes.py
+++ b/sphinx/addnodes.py
@@ -263,6 +263,16 @@ class pending_xref(nodes.Inline, nodes.Element):
     """
 
 
+class pending_xtext(nodes.Inline, nodes.Element):
+    """Node for textual cross-references that cannot be resolved without
+    complete information about all documents.
+
+    These nodes are text-only nodes whose text depends on the resolution of
+    a pending_xref.  They are resolved before writing output, in
+    BuildEnvironment.resolve_references.
+    """
+
+
 class number_reference(nodes.reference):
     """Node for number references, similar to pending_xref."""
 
--- a/sphinx/environment/adapters/toctree.py
+++ b/sphinx/environment/adapters/toctree.py
@@ -158,7 +158,7 @@ class TocTree(object):
                         maxdepth = self.env.metadata[ref].get('tocdepth', 0)
                         if ref not in toctree_ancestors or (prune and maxdepth > 0):
                             self._toctree_prune(toc, 2, maxdepth, collapse)
-                        self.process_only_nodes(toc)
+                        self.sanitize_toc('', toc) # FIXME
                         if title and toc.children and len(toc.children) == 1:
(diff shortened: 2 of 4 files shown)


mwoehlke:
It appears that sphinx generating its own doc is somewhere obtaining a TOC that does not get the transforms applied. Any ideas how/where it is doing that?

AA-Turner:
I would like to find another solution to #4448 than introducing new nodes, delayed resolution, etc.

PR #4491, opened 2018-01-24
autodoc: prefer _MockImporter over other importers in sys.meta_path

Emantor:
Subject: Add _MockImporter in front of other importers in sys.meta_path

### Feature or Bugfix
- Bugfix

### Purpose
In case we want to mock modules which are actually available in the current
environment, we need to add the _MockImporter before other importers. We ran
into this problem in our application, where importing the existing modules
caused side effects which were not fixed by autodoc_mock_imports.

### Detail
-/-

### Relates
-/-

This is debatable, i am not sure whether this is the intended behaviour of autodoc_mock_imports. At least from my point of view, the documentation implied that it not only mocked missing modules, but in sys.path available modules as well.

A backport to stable would be needed as well. Thanks for your time and work on sphinx 👍

### DIFF
--- a/sphinx/ext/autodoc/importer.py
+++ b/sphinx/ext/autodoc/importer.py
@@ -87,7 +87,7 @@ class _MockImporter(object):
         self.names = names
         self.mocked_modules = []  # type: List[str]
         # enable hook by adding itself to meta_path
-        sys.meta_path = sys.meta_path + [self]
+        sys.meta_path.insert(0, self)
 
     def disable(self):
         # remove `self` from `sys.meta_path` to disable import hook


tk0miya:
@shimizukawa could you give us comments?

Note: The failure on Travis is not related with this PR. It will be resolved with master HEAD.

shimizukawa:
Actually, I did not specifically consider hook order. I think it's okay to hook before the defaults. If user don't specify mock target, hook will not do anything, and if user explicitly specify target, there is an intention to mock even existing modules.

shimizukawa (approved):

tk0miya (changes requested):
@Emantor could you rebase this to 1.7-release branch? Then I'll merge this soon.
Thanks,

Emantor:
Rebased, should be ready to merge after tests 👍

tk0miya (approved):

tk0miya:
Thank you for quick response.
Merged!

Issue #4492, opened 2018-01-24
New in 1.7: samefile check against smartypants in flake8 plugin raises OSError


oleksandr-pavlyk:
Subject: OSError raised in flake8 plugin during a work-around `samefile` check

While making a [filesame comparison to smartypants.py](https://github.com/sphinx-doc/sphinx/blob/1.7-release/utils/checks.py#L51) in the flake8 plugin an `OSError` is being raised. 

Consequentialy, build of scikit-image fails due to error in `flake8`  run when sphinx 1.7.0b1 is used with error ``OSError: [Errno 2] No such file or directory: './sphinx/util/smartypants.py'``.

The scikit-image PR code runs cleanly with current stable version of sphinx, 1.6.6

See https://github.com/scikit-image/scikit-image/pull/2935 for more details.

Support for smartypants is documented to have been removed in 1.7, as far as I understand, but I can see file `sphinx/util/smartypants.py` in sources though.

tk0miya:
@stephenfin could you check this please?
It seems `utils` package is installed globally. This is not good. As you mentioned in `setup.py`, it is better to use `flake8:local-plugins` instead.

tk0miya:
Fixed by #4498.
Please test HEAD of 1.7-release branch or wait for 1.7b2.
Thank you for reporting

PR #4494, opened 2018-01-25
Check that if ref. file exists prior to calling `samefile`

oleksandr-pavlyk:
Closes https://github.com/sphinx-doc/sphinx/issues/4492

Prior to running `os.path.samefile(fn, 'path/to/file')` check that `'path/to/file/` exists.

This prevents the OSError  from being raised, but perhaps a better solution is in order (use absolute path?)


stephenfin (commented):
This isn't necessary. The plugin should only have run locally. I have resolved this in #4498.

oleksandr-pavlyk:
Thanks, @stephenfin . Closing then.

PR #4002, opened 2017-08-13
#3987 remove alabaster-specific sidebars from sphinx-quickstart template

TimKam:
Fixes #3987 

Subject: Removes alabaster-specific sidebars from sphinx-quickstart template

### Feature or Bugfix
- Bugfix

### Purpose
Removes alabaster-specific sidebars from sphinx-quickstart template.
Otherwise, changing the theme after initiating a project breaks the build.
One has to adjust ``html_sidebars`` in the config file first, to fix this.
This is a bit confusing.
With the adjusted settings, it's possible to build common themes right away. I tested with:
* ``basic``
* ``alabaster``
* ``sphinxdoc``

### Detail
#3987

Issue #3987, opened 2017-08-08
Changing theme from alabaster causes HTML build to fail


tristan957:
Subject: Building HTML fails if theme is not alabaster

### Problem
- Can't use any theme that is not alabaster

#### Procedure to reproduce the problem
```
- Change theme to something other than alabaster
```

#### Error logs / results
```
# Sphinx version: 1.6.3
# Python version: 3.6.1 (CPython)
# Docutils version: 0.14 
# Jinja2 version: 2.9.6
# Last messages:
#   checking consistency...
#   
#   done
#   
#   preparing documents...
#   
#   done
#   
#   writing output... [  5%] about
#   
# Loaded extensions:
#   alabaster (0.7.10) from C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\alabaster\__init__.py
#   sphinx.ext.autodoc (1.6.3) from C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\ext\autodoc.py
#   sphinx.ext.doctest (1.6.3) from C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\ext\doctest.py
#   sphinx.ext.todo (1.6.3) from C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\ext\todo.py
#   sphinx.ext.mathjax (1.6.3) from C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\ext\mathjax.py
#   sphinx.ext.viewcode (1.6.3) from C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\ext\viewcode.py
Traceback (most recent call last):
  File "C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\cmdline.py", line 306, in main
    app.build(opts.force_all, filenames)
  File "C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\application.py", line 339, in build
    self.builder.build_update()
  File "C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\builders\__init__.py", line 331, in build_update
    'out of date' % len(to_build))
  File "C:\Users\591709\Documents\Projects\icp-data-tool\lib\site-packages\sphinx\builders\__init__.py", line 397, in build
    self.write(docnames, list(updated_docnames), method)
[...]

PR #3849, opened 2017-06-03
Add html_sidebars for alabaster by default

segevfiner:
Subject: Add html_sidebars for alabaster by default

### Feature or Bugfix
- Bugfix

### Purpose
It's required for alabaster to render correctly, see:
http://alabaster.readthedocs.io/en/latest/installation.html

### Detail
- Without this the sidebar renders quite weird with the default alabaster theme.

PR #4177, opened 2017-10-22
HTML themes can set up default sidebars through ``theme.conf``

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Give a better default sidebars setting from 3rd party theme (cf. alabaster)

### Detail
- Add `sidebar` setting to `theme.conf`
- It changes the default sidebar setting
- It allows to override it by `html_sidebars`

### Relates
- #4102

PR #4496, opened 2018-01-25
Re-add Alabaster specific html_sidebars conditionally

segevfiner:
Subject: Re-add Alabaster specific html_sidebars conditionally

### Feature or Bugfix
- Bugfix

### Purpose
- Making Alabaster look right in a newly created project.

### Detail
A second attempt of #3849, reverting #4002. Should keep #3987 fixed while also making sure Alabaster looks right in a newly created project.

This was fixed in a different way for 1.7, moving the default sidebars config into the theme.conf in b78f309.

### Relates
- #3849
- #3987
- #4002

@jfbu @tk0miya @joinghel @shimizukawa

### DIFF
--- a/sphinx/templates/quickstart/conf.py_t
+++ b/sphinx/templates/quickstart/conf.py_t
@@ -112,12 +112,16 @@ html_static_path = ['{{ dot }}static']
 #
 # This is required for the alabaster theme
 # refs: http://alabaster.readthedocs.io/en/latest/installation.html#sidebars
-html_sidebars = {
-    '**': [
-        'relations.html',  # needs 'show_related': True theme option to display
-        'searchbox.html',
-    ]
-}
+if html_theme == 'alabaster':
+    html_sidebars = {
+        '**': [
+            'about.html',
+            'navigation.html',
+            'relations.html',  # needs 'show_related': True theme option to display
+            'searchbox.html',
+            'donate.html',
+        ]
+    }
 
 
 # -- Options for HTMLHelp output ------------------------------------------


sphinx/templates/quickstart/conf.py_t
```diff
@@ -112,12 +112,16 @@ html_static_path = ['{{ dot }}static']
...
-    ]
-}
+if html_theme == 'alabaster':
+    html_sidebars = {
+        '**': [
+            'about.html',
+            'navigation.html',
+            'relations.html',  # needs 'show_related': True theme option to display
+            'searchbox.html',
+            'donate.html',
```
jfbu, line 122: not clear to me if donate should be included per default

sphinx/templates/quickstart/conf.py_t
```diff
@@ -112,12 +112,16 @@ html_static_path = ['{{ dot }}static']
...
-        'relations.html',  # needs 'show_related': True theme option to display
-        'searchbox.html',
-    ]
-}
+if html_theme == 'alabaster':
+    html_sidebars = {
+        '**': [
+            'about.html',
+            'navigation.html',
+            'relations.html',  # needs 'show_related': True theme option to display
```
jfbu, line 120: as this needs extra activation step for Alabaster, I think it should be commented out

jfbu (commented):
LGTM with nits

sphinx/templates/quickstart/conf.py_t
segevfiner: This was what it was before being removed in https://github.com/sphinx-doc/sphinx/pull/4002. [donate.html](https://github.com/bitprophet/alabaster/blob/da3b2595b95bc94deacc45c617096ea57ce6790a/alabaster/donate.html) is activated conditionally by Alabaster from additional theme configuration. It won't be shown by default.

This is the same default we will get with [#4177 - html.py](https://github.com/sphinx-doc/sphinx/pull/4177/files#diff-c17608ed63bfe1714c6befb9d4201feeR878), and probably will be the same once `sidebars` is added to Alabaster's `theme.conf`. Of course, that is for 1.7 (master). This PR is for 1.6 (stable).

sphinx/templates/quickstart/conf.py_t
segevfiner: The whole point of the `if` is that it's *not commented out* so that it's active by default for a new project, since the *default* theme is Alabaster. When the user changes the theme, he will likely forget to remove this, which is what #3987 was about. The `if` means it won't apply even if the user doesn't notice this, which means his newly selected theme will work correctly.

jfbu: yes I understand that. But I think for Alabaster user it makes more sense to have ``'relations.html'`` commented out because it needs extra step as explained in comment. Having the whole line commented out would make it even clearer to user.

sphinx/templates/quickstart/conf.py_t
jfbu: ok no strong opinion on this

jfbu (approved):

tk0miya:
-0; 1.7 will be released soon. so 1.6.7 might not be released, or supported only in short period.

tk0miya:
any comments? if none, I'll reject this because of 1.7 release.

segevfiner:
> any comments? if none, I'll reject this because of 1.7 release.

Are you planning not to release a 1.6.7? If so, than of course there is no need for this PR. If you do plan to release a 1.6.7 than it still applies of course like any other bug fix that was applied against 1.6.

sphinx/templates/quickstart/conf.py_t
```diff
@@ -112,12 +112,16 @@ html_static_path = ['{{ dot }}static']
 #
 # This is required for the alabaster theme
 # refs: http://alabaster.readthedocs.io/en/latest/installation.html#sidebars
-html_sidebars = {
-    '**': [
-        'relations.html',  # needs 'show_related': True theme option to display
-        'searchbox.html',
-    ]
-}
+if html_theme == 'alabaster':
```
tk0miya, line 115: What will be happened for non alabaster themes?

tk0miya, line 115: this does not work if user uses `-D html_theme=alabaster` from command line.
It also does not work if user updates `html_theme` variable after this block.

tk0miya:
No, we'll release 1.6.7 in this weekend. And 1.7 will be released in next weekend. That means 1.6.7 is supported during a week only.
But generated `conf.py` will alive for a long time. because nobody usually re-generate it.

I know 1.6 series will still contain this bug even after 1.6.7. But it will be fixed after a week.
So I prefer not to merge this into stable branch now.

sphinx/templates/quickstart/conf.py_t
segevfiner: `html_sidebars` won't be set so it should be left at it's default which should be correct for all the other bundled themes. For any other theme that require a custom sidebar setting, the user will have to update it by itself. That's what the 1.7 fix fixes in general.

segevfiner: Yeah, it's quite a stopgap fix. I'm not sure if I can query the setting from the command line, and there is not much I can do if the user updates it after this if block.

The 1.7 fix avoids this if `html_sidebars` is left unset but if you do set it in `conf.py` than `-D html_theme=...` can also cause breakage. For example, if the `html_sidebars` is set to something special and you pass a theme that requires a different `html_sidebars` setting.

tk0miya:
Thank you for all your work, but I decided not to merge this into 1.6.7 release.

Thanks,

PR #4498, opened 2018-01-26
Don't install flake8 plugin globally

stephenfin:
Resolve issue #4492.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,6 +34,10 @@ max-line-length = 95
 ignore = E116,E241,E251,E741
 exclude = .git,.tox,.venv,tests/*,build/*,doc/_build/*,sphinx/search/*,sphinx/pycode/pgen2/*,doc/ext/example*.py
 
+[flake8:local-plugins]
+extension =
+    X101 = utils.checks:sphinx_has_header
+
 [mypy]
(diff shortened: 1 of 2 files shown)


tk0miya:
Could you check `utils/checks.py` is installed or not? If installed, we have to remove it from packages definition.

stephenfin:
> Could you check utils/checks.py is installed or not? If installed, we have to remove it from packages definition.

It is not, but we don't include the `util` directory in `MANIFEST.in` anyway so I doubt it matters

tk0miya:
It seems utils package is installed.
```
$ virtualenv test
$ ./test/bin/pip install git+https://github.com/sphinx-doc/sphinx
$ ./test/bin/python
Python 3.6.4 (default, Dec 20 2017, 01:09:24)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> import utils.checks
>>> utils.checks
<module 'utils.checks' from '/Users/tkomiya/work/issuesgraph/test/lib/python3.6/site-packages/utils/checks.py'>
```

Anyway, this is not related with this PR.

tk0miya:
Merged! thank you always

Issue #4375, opened 2018-01-04
Sphinx not parallel installable for python2 and python3


gtristan:
There is no stable entry point one can rely on to invoke sphinx-apidoc or sphinx-build specifically for python3, as far as I can see there is only *${prefix}/bin/sphinx-build* and  *${prefix}/bin/sphinx-apidoc*.

It is unclear, on any given system, if the `/usr/bin/sphinx-*` tools will be using python2 or python3, however projects which depend on sphinx must be able to assert that sphinx is installed for their python version, and have an API stable way of invoking sphinx for their python version.

I tried installing python-sphinx and python3-sphinx on debian, it did not complain about a conflict and happily installed both, the resulting `/usr/bin/sphinx-build` is for python2, and I cannot find an entry point it may have added for python3.

stephenfin:
@tk0miya I discussed this with @gtristan on `#sphinx-doc` today and am happy to work on it. Looks like something that we'll need to work with packagers to really solve and it won't be necessary after Sphinx 2.0 (which should be Python 3-only). However, it would be a good enhancement until then.

tk0miya:
@stephenfin thank you for your help.
I don't have knowledge for this. So it would be nice if you help us.

tk0miya:
It seems `/usr/bin/sphnx-*` are hyperlink to `/usr/share/sphinx/scripts/python*/sphinx-*`.
```
root@f45348219a96:/# ls -l /usr/bin/sphinx-build
lrwxrwxrwx 1 root root 44 Jan 21 18:01 /usr/bin/sphinx-build -> ../share/sphinx/scripts/python2/sphinx-build
```
I think you can use Sphinx with python3 if you add `/usr/share/sphinx/scripts/python3` to PATH.

I think this is packaging problem, not Sphinx itself. So I recommend you to report this to Debian project.

gtristan:
if the user has to setup any special environment to specify which API to use, it is not truely parallel installable.

Parallel installability would just require that your entry point be accessible under a different name (something like sphinx3-build would work), for access via python imports this is not an issue because of how people generally install into versioned subdirectories of a python installation, and once you are running python code you have already selected your python version so only the desired API is found, this is not so when using the entry point, as such, the command "sphinx-build" is ambiguous on any given host.

[...]

PR #4499, opened 2018-01-26
setup: Add versioned binaries

stephenfin:
This is similar to what pip does [1].

[1] https://github.com/pypa/pip/blob/9.0.0/setup.py#L80-L84

Fixes: #4375

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,7 @@ import os
 import sys
 from distutils import log
 from distutils.cmd import Command
+import itertools
 
 import sphinx
 
@@ -57,6 +58,16 @@ extras_require = {
         'typed_ast',
     ],
 }
+
+console_scripts = list(itertools.chain.from_iterable([
+    [
+        'sphinx-build%s = sphinx.cmd.build:main' % version,
+        'sphinx-quickstart%s = sphinx.cmd.quickstart:main' % version,
+        'sphinx-apidoc%s = sphinx.ext.apidoc:main' % version,
+        'sphinx-autogen%s = sphinx.ext.autosummary.generate:main' % version,
+    ] for version in ('', '-%s' % sys.version[:1], '-%s' % sys.version[:3])
+]))
+
 
 # Provide a "compile_catalog" command that also creates the translated
 # JavaScript files if Babel is available.
@@ -217,12 +228,7 @@ setup(
     packages=find_packages(exclude=['tests']),
     include_package_data=True,
     entry_points={
-        'console_scripts': [
-            'sphinx-build = sphinx.cmd.build:main',
-            'sphinx-quickstart = sphinx.cmd.quickstart:main',
-            'sphinx-apidoc = sphinx.ext.apidoc:main',
-            'sphinx-autogen = sphinx.ext.autosummary.generate:main',
-        ],
+        'console_scripts': console_scripts,
         'distutils.commands': [
             'build_sphinx = sphinx.setup_command:BuildDoc',
         ],


setup.py
```diff
@@ -58,6 +59,16 @@
     ],
 }
 
+console_scripts = list(itertools.chain.from_iterable([
+    [
+        'sphinx-build%s = sphinx.cmd.build:main' % version,
+        'sphinx-quickstart%s = sphinx.cmd.quickstart:main' % version,
+        'sphinx-apidoc%s = sphinx.ext.apidoc:main' % version,
+        'sphinx-autogen%s = sphinx.ext.autosummary.generate:main' % version,
+    ] for version in ('', '-%s' % sys.version[:1], '-%s' % sys.version[:3])
```
tk0miya, line 68: Why `version[:3]` is needed?

tk0miya:
I worry about wheel package. Is `universal = 1` okay to this change?

setup.py
```diff
@@ -58,6 +59,16 @@
     ],
 }
 
+console_scripts = list(itertools.chain.from_iterable([
+    [
+        'sphinx-build%s = sphinx.cmd.build:main' % version,
+        'sphinx-quickstart%s = sphinx.cmd.quickstart:main' % version,
```
tk0miya, line 65: Is this also needed? quickstart is not related with python version.

tk0miya:
I noticed that these `sphinx-build-N` is not used in `Makefile`. So I think this is not helpful so much.

stephenfin:
> I noticed that these sphinx-build-N is not used in Makefile. So I think this is not helpful so much.

No, this is intentional. We don't _need_ to use the versioned Sphinx binary for 99% of use cases, so we should leave it to the default. If/when users want to override it, they can customize the `Makefile` or set the `SPHINXBUILD` environment variable.

setup.py
stephenfin: This was taken from `pip`. I assume newer versions of Python3 add features that are not backwards compatible, like the addition of the [`async` keyword in Python3.4](https://docs.python.org/3/library/asyncio.html). As a result, a fully versioned binary would be helpful

stephenfin:
> I worry about wheel package. Is universal = 1 okay to this change?

Good spot. Wheels are not valid with this, as seen [here](https://github.com/pypa/pip/blob/9.0.0/pip/wheel.py#L425-L487). I assume this means we need to start building versioned wheels?

setup.py
stephenfin: I just converted everything to be consistent. I can drop this if you prefer?

tk0miya:
I doubt this is helpful. This might be only useful for who installs multiple version of pythons and also installs Sphinx to all of pythons. Of course, they don't use virtualenv.
In addition, python version is only related with autodoc family (apidoc and autosummary). It means the user is a python engineer. Is this really needed for them? I believe `pythonX.Y -m sphinx` is also useful for them.

setup.py
tk0miya: It might be useful if python has (or will have) breaking compatibility between minor version.
If not, this is not needed. So I don't feel this is needed.

stephenfin:
> I doubt this is helpful. This might be only useful for who installs multiple version of pythons and also installs Sphinx to all of pythons. Of course, they don't use virtualenv.
> In addition, python version is only related with autodoc family (apidoc and autosummary). It means the user is a python engineer. Is this really needed for them? I believe pythonX.Y -m sphinx is also useful for them.

OK, can you close this and mark the bug as invalid in that case?

tk0miya:
Okay, I marked this as wontfix and close now.
Thanks,

PR #4505, opened 2018-01-27
Use flake8-import-order

tk0miya:
### Feature or Bugfix
- Coding styles

### Purpose
- To regulate order of imports
    - standard libraries
    - third parties
    - Sphinx modules


setup.cfg
```diff
@@ -31,8 +31,10 @@ directory = sphinx/locale/
 
 [flake8]
 max-line-length = 95
-ignore = E116,E241,E251,E741
+ignore = E116,E241,E251,E741,I101
```
tk0miya, line 34: >I101: The names in your from import are in the wrong order.

I will enable this on next PR.

PR #4506, opened 2018-01-27
Fix #3447: latex: longtable may render "horizontally" and overflow in right margin

### LINKED ISSUE #3447: when assigning longtable class to table for PDF, it may render "horizontally" and overflow in right margin
### Problem

Table overflows in margin when assigned longtable class, paradoxically this happens when table does not contain "problematic" contents like lists...

```
.. table::
   :class: longtable

   +-------------+-------------+
   | column 1    | column 2    |
   +-------------+-------------+
   |             |             |
   |Sometimes you|The default  |
   |are not      |templating   |
   |looking for  |language in  |
   |the commit   |Sphinx is    |
   |that         |Jinja. It’s  |
   |introduced a |Django/Smarty|
   |breakage, but|inspired and |
   |rather for a |easy to      |
   |commit that  |understand.  |
   |caused a     |The most     |
   |change       |important    |
   |between some |concept in   |
   |other "old"  |Jinja is     |
   |state and    |template     |
   |"new" state. |inheritance, |
   |For example, |which means  |
   |you might be |that you can |
   |looking for  |overwrite    |
   |the commit   |only specific|
   |that         |blocks within|
   |introduced a |a template,  |
   |particular   |customizing  |
   |fix. Or you  |it while also|
   |might be     |keeping the  |
   |looking for  |changes at a |
   |the first    |minimum.     |
   |commit in    |             |
   |which the    |             |
   |source-code  |             |
   |filenames    |             |
   |were finally |             |
   |all converted|             |
   |to your      |             |
   |company’s    |             |
   |naming       |             |
   |standard. Or |             |
   |whatever.    |             |
   +-------------+-------------+


.. table::
   :class: longtable

   +-------------+-------------+
   | column 1    | column 2    |
   +-------------+-------------+
   |             |             |
   |1. a         |The default  |
   |             |templating   |
   |2. b         |language in  |
   |             |Sphinx is    |
   |Sometimes you|Jinja. It’s  |
   |are not      |Django/Smarty|
   |looking for  |inspired and |
   |the commit   |easy to      |
   |that         |understand.  |
   |introduced a |The most     |
   |breakage, but|important    |
   |rather for a |concept in   |
   |commit that  |Jinja is     |
   |caused a     |template     |
   |change       |inheritance, |
   |between some |which means  |
   |other "old"  |that you can |
   |state and    |overwrite    |
   |"new" state. |only specific|
   |For example, |blocks within|
   |you might be |a template,  |
   |looking for  |customizing  |
   |the commit   |it while also|
   |that         |keeping the  |
   |introduced a |changes at a |
   |particular   |minimum.     |
   |fix. Or you  |             |
   |might be     |             |
   |looking for  |             |
   |the first    |             |
   |commit in    |             |
   |which the    |             |
   |source-code  |             |
   |filenames    |             |
   |were finally |             |
   |all converted|             |
   |to your      |             |
   |company’s    |             |
   |naming       |             |
   |standard. Or |             |
   |whatever.    |             |
   +-------------+-------------+

```

#### Error logs / results

capture d ecran 2017-02-22 a 16 19 07

#### Expected results

table does not overflow

### Environment info
- Python version: 3.5.2
- Sphinx version: 1.4.x, 1.5.x
- TeXLive 2016 for TeX

tk0miya:
refs: #3447

This is just an idea. But it seems working fine.
@jfbu could you give me a comment for this?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,8 @@ Bugs fixed
 * #4472: DOCUMENTATION_OPTIONS is not defined
 * #4491: autodoc: prefer _MockImporter over other importers in sys.meta_path
 * #4490: autodoc: type annotation is broken with python 3.7.0a4+
+* #3447: latex: longtable may render "horizontally" and overflow in right margin
+
 
 Testing
 --------
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -407,6 +407,13 @@ class Table(object):
             return '{|' + ('T|' * self.colcount) + '}\n'
         elif self.has_oldproblematic:
             return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
+        elif self.is_longtable():
+            # Workaround: Give a default colspec to avoid overflows if ``:widths:`` or
+            # ``tabularcolumns`` are not specified.
+            # see https://github.com/sphinx-doc/sphinx/issues/3447
+            total = self.width
+            colspecs = ['\\X{%d}{%d}' % (1, total) for _ in range(total)]
+            return '{|%s|}\n' % '|'.join(colspecs)
         else:
             return '{|' + ('l|' * self.colcount) + '}\n'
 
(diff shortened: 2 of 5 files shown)


jfbu (commented):
-0 It does fix the original issue, but many projects will now have longtables with wider columns than earlier, even if such width is not needed.

For this reason I seem some other way is preferable. I need to think about it.

edit: see #4507

tk0miya:
Okay. Thank you for comment.
I withdraw this.

I'm glad that this leads to your nicer PR :-)

PR #4507, opened 2018-01-27
[WIP] LaTeX: estimate colwidths via Python for better tables

jfbu:
Relates #3447, #4506 

This is proof of concept.

Some aspects need improvement (see review below), but I also tested that it gives better results in some cases than LaTeX macro package `tabulary`. Indeed, it may open possibility to remove entirely usage of `tabulary` at some point. For example with this

```rst
.. table::

   ========= ======== =======================================================
   Tag       Required Description
   ========= ======== =======================================================
   columns   No       Columns of the content to return. Each content listing
                      has a set of defaults columns that make sense for that
                      particular content type. By default, these columns are
                      returned in addition to what was requested here.
   xquery    No       A  xquery
   filter    No       Return files containing the specified filter string
   ========= ======== =======================================================
```

and modifying latex.py to not use tabulary but tabular with the `\Y` columns as determined by latex writer itself one gets this:

capture d ecran 2018-01-27 a 22 10 53

which is better than the currently produced output using tabulary:

capture d ecran 2018-01-27 a 22 59 44

(such table currently needs to be fixed by using tabularcolumns, but image above shows Sphinx could get it wright itself)

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -324,8 +324,8 @@ class Table:
         self.colcount = 0
         self.colspec = None                     # type: str
         self.colwidths = []                     # type: List[int]
+        self.empirical_colwidths = defaultdict(float)  # type: Dict[int, float]
         self.has_problematic = False
-        self.has_oldproblematic = False
         self.has_verbatim = False
         self.caption = None                     # type: List[str]
         self.stubs = []                         # type: List[int]
@@ -395,15 +395,26 @@ class Table:
             total = sum(self.colwidths)
             colspecs = ['\\X{%d}{%d}' % (width, total) for width in self.colwidths]
             return '{|%s|}\n' % '|'.join(colspecs)
-        elif self.has_problematic:
-            return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
-        elif self.get_table_type() == 'tabulary':
-            # sphinx.sty sets T to be J by default.
-            return '{|' + ('T|' * self.colcount) + '}\n'
-        elif self.has_oldproblematic:
-            return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
+        elif not self.get_table_type() == 'tabulary':
+            total_Y, colcount_l = 0.0, 0
+            for i in range(self.colcount):
+                estimate = self.empirical_colwidths[i]
+                if estimate > 16:
+                    total_Y += estimate
+                else:
+                    colcount_l += 1
+            linewidthratioY = total_Y / (total_Y + 16 * colcount_l)
+            colspecs = []
+            for i in range(self.colcount):
+                estimate = self.empirical_colwidths[i]
+                if estimate > 16:
+                    colspecs.append('\\Y{%.3f}' % (float(estimate) /
+                                                   total_Y * linewidthratioY))
+                else:
+                    colspecs.append('l')
+            return '{|%s|}\n' % '|'.join(colspecs)
         else:
-            return '{|' + ('l|' * self.colcount) + '}\n'
+            return '{|' + ('T|' * self.colcount) + '}\n'
 
     def add_cell(self, height, width):
         # type: (int, int) -> None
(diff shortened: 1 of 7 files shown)


sphinx/writers/latex.py
```diff
@@ -1513,30 +1524,33 @@ def visit_entry(self, node):
                        '\\vbox{\\hbox{\\strut}}\\end{varwidth}%\n') + context
             self.needs_linetrimming = 1
         if len(node.traverse(nodes.paragraph)) >= 2:
-            self.table.has_oldproblematic = True
+            for i in range(cell.col, cell.col + cell.width):
+                self.table.empirical_colwidths[i] = max(self.table.empirical_colwidths[i], 17)
```
jfbu, line 1392: admittedly I have not yet tested this works well with merged cells

sphinx/writers/latex.py
```diff
@@ -1513,30 +1524,33 @@ def visit_entry(self, node):
...
             # Remove empty lines from top of merged cell
             while body and body[0] == "\n":
                 body.pop(0)
-            self.body.extend(body)
 
+        self.body.extend(body)
         self.body.append(self.context.pop())
 
         cell = self.table.cell()
+        bodywidthpercol = float(sum(len(i) for i in body)) / cell.width
```
jfbu, line 1414: I am not happy with this at it counts characters of LaTeX mark-up such as a footnote or `\textbackslash{}` which artificially increase weight of corresponding column

jfbu: What I would like is simply measure the source length for the cell contents, before LaTeX escapes or added mark-up; for lack of enough knowledge of docutils nodes, I used this alternative, which is not satisfactory.

sphinx/writers/latex.py
```diff
@@ -1513,30 +1524,33 @@ def visit_entry(self, node):
                        '\\vbox{\\hbox{\\strut}}\\end{varwidth}%\n') + context
             self.needs_linetrimming = 1
         if len(node.traverse(nodes.paragraph)) >= 2:
-            self.table.has_oldproblematic = True
+            for i in range(cell.col, cell.col + cell.width):
```
jfbu, line 1391: goal here is that multi-paragraph cells must be rendered in p-type columns (``\Y`` or ``\X`` but not ``l``)

sphinx/writers/latex.py
jfbu: OMG, initially I used only integers in empirical_colwidths, so 17 made sense as next one after 16. But know they are floats. So I must fix this (will do later, if idea of this PR is accepted)

sphinx/writers/latex.py
```diff
@@ -400,15 +400,26 @@ def get_colspec(self):
...
-        elif self.get_table_type() == 'tabulary':
-            # sphinx.sty sets T to be J by default.
-            return '{|' + ('T|' * self.colcount) + '}\n'
-        elif self.has_oldproblematic:
-            return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
+        elif not self.get_table_type() == 'tabulary':
+            total_Y, colcount_l = 0.0, 0
+            for i in range(self.colcount):
+                estimate = self.empirical_colwidths[i]
+                if estimate > 16:
```
jfbu, line 402: this threshold at 16 was chosen also because any kind of LaTeX environment will force use of a ``\Y`` type, not an ``l`` column.

The idea is that narrow cells should be rendered in l columns. Cells with more than a few words should use p type columns with a width proportional to contents.

This is what LaTeX macro package tabulary does, but it looks easier on Python side, although we want to measure only the weight of rst source (is thaat ok with substitutions?)

jfbu (commented):
I stressed some defects

jfbu:
I think it would be worthwile to extend this indeed to a complete replacement of LaTeX macro package tabulary

- tabulary has complicated many things because it can't handle literal contents in cells, and any kind of list or quotes 

- it has caused special difficulties with footnotes,

- the code in sphinxmulticell.sty hacks into some tabulary innards, but this can be maintained only by expert LaTeX user.

This PR gives idea how to emulate, and even improve, at Python level some job actually done in favourable cases by tabulary at LaTeX macro stage.

My idea is that at some point indeed all usage of tabulary should be removed from Sphinx. I don't know if 1.7 is still acceptable deadline for that? ping @tk0miya

sphinx/writers/latex.py
```diff
@@ -400,15 +400,26 @@ def get_colspec(self):
...
             colspecs = ['\\X{%d}{%d}' % (width, total) for width in self.colwidths]
             return '{|%s|}\n' % '|'.join(colspecs)
-        elif self.has_problematic:
-            return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
-        elif self.get_table_type() == 'tabulary':
-            # sphinx.sty sets T to be J by default.
-            return '{|' + ('T|' * self.colcount) + '}\n'
-        elif self.has_oldproblematic:
-            return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
+        elif not self.get_table_type() == 'tabulary':
```
tk0miya, line 398: I feel positive condition is a bit simpler. How about `elif self.get_table_type() in ('longtable', 'tabular')` instead?

sphinx/writers/latex.py
tk0miya: +1 basically; but we have to improve this a little.

* add comment for description
* replace 16 with constant (it is magic number!)

sphinx/writers/latex.py
```diff
@@ -400,15 +400,26 @@ def get_colspec(self):
...
+            colspecs = []
+            for i in range(self.colcount):
+                estimate = self.empirical_colwidths[i]
+                if estimate > 16:
+                    colspecs.append('\\Y{%.3f}' % (float(estimate) /
+                                                   total_Y * linewidthratioY))
+                else:
+                    colspecs.append('l')
+            return '{|%s|}\n' % '|'.join(colspecs)
         else:
```
tk0miya, line 416: it might be useful if there is a comment like `# for taburary`

sphinx/writers/latex.py
tk0miya: I think `LaTeXTranslater` should not access `table.empirical_colwidths` directly.
I think encapsulation is good technique for boundary of responsibility.
How about `table.update_colwidth(cell, bodywidthpercol=None)`?

tk0miya (changes requested):
LGTM!

tk0miya:
>My idea is that at some point indeed all usage of tabulary should be removed from Sphinx. I don't know if 1.7 is still acceptable deadline for that? ping @tk0miya

I feel it is a big change. Is my feeling right? If so, the migration should go to 1.8.
If not, I'm okay to add it into 1.7 (of course, it has to enough stable).

tk0miya:
Sorry, I just removed `1.7-release` branch last night. As a result, github closes this automatically. It was not intended. so I reopened this again and changed target branch to `1.7`.

jfbu:
rebased on current master

AA-Turner:
@jfbu this PR has been WIP for ~6 years, is it worth keeping open or can we close it?

A

jfbu:
> @jfbu this PR has been WIP for ~6 years, is it worth keeping open or can we close it?
> 
> A

alas, this project never got completed. Better to close it.

PR #4516, opened 2018-01-29
Fix BuildInfo comparison

tk0miya:
OMG, `__eq__` is not used for `!=` operator...


shimizukawa (approved):
wow...

tk0miya:
This is my mistake. I misread python's docs.

PR #4525, opened 2018-01-30
Fix autosectionlabel does not support parallel build

tk0miya:
Unfortunately, autosectionlabel extension does not returns its metadata...


tk0miya:
Ah, I noticed this should be merged into stable branch. (just after merged...)
I will do it manually.

PR #4544, opened 2018-02-04
Update type annotations

tk0miya:
### Feature or Bugfix
- Refactor

### Purpose
- Add type annotations


stephenfin:
Could I ask how you did this? By hand?

tk0miya:
Yes, by my hand. I'd tried some tools once, but I don't like them. At present, there might be a better tool. But I don't research that...

PR #4561, opened 2018-02-05
Cleanup

stephenfin:
Remove some dead code and mark deprecated functions with `warnings`.


sphinx/addnodes.py
```diff
@@ -200,6 +200,7 @@ class index(nodes.Invisible, nodes.Inline, nodes.TextElement):
     """
 
 
+# TODO(stephenfin): Remove this in Sphinx 2.0
```
stephenfin: I actually tried to add `__init__` here so we could raise a warning, but it seems we have a `Centered` directive which needs to be removed first.

tk0miya: Why removing the directive?

sphinx/util/__init__.py
```diff
@@ -172,6 +174,9 @@ def copy_static_entry(source, targetdir, builder, context={},
 
     Handles all possible cases of files, directories and subdirectories.
     """
+    warnings.warn('sphinx.util.copy_static_entry is deprecated for removal',
+                  RemovedInSphinx20Warning)
```
tk0miya: See our deprecation policy. It should be deprecated in 3.0.
https://github.com/sphinx-doc/sphinx/blob/master/CONTRIBUTING.rst#deprecating-a-feature

I know this is not a feature, but we should also wait a moment on removal.

sphinx/util/osutil.py
```diff
@@ -181,8 +183,10 @@ def make_filename(string):
 
 def ustrftime(format, *args):
     # type: (unicode, Any) -> unicode
-    # [DEPRECATED] strftime for unicode strings
-    # It will be removed at Sphinx-1.5
+    """[DEPRECATED] strftime for unicode strings."""
+    warnings.warn('sphinx.util.osutil.ustrtime is deprecated for removal',
+                  RemovedInSphinx20Warning)
```
tk0miya: This is also.

tk0miya (changes requested):
LGTM with nits.

sphinx/util/__init__.py
stephenfin: But both of these have been marked as deprecated for years now. I'm just making it more obvious with warnings. I don't think there's any reason not to do this.

sphinx/addnodes.py
stephenfin: See the comment - this is deprecated since before 2010

tk0miya: This node is surely marked as deprecated. But `centered` directive is not marked as deprecated.

IMO, the directive is not important for sphinx-core. so +1 for deprecation. But it would be nice if we can split it into other package (for workaround).

sphinx/util/__init__.py
tk0miya: In my experience, nobody reads docs and CHANGES. So it would be better to emit deprecation warnings.
If extensions won't work after upgrade, users will be in trouble. So I'd like to emit deprecation warnings obviously to provide a deprecation period for users and developers.

stephenfin: OK, let me rework these so.

tk0miya: Thank you for update!
Sorry for boring fix. These deprecation with only comments were my mistakes.

tk0miya (approved):

tk0miya:
Merged. Thank you always!

tk0miya:
ah, sorry. I didn't merge yet. flake8 warns this. So I can't merge this.
Could you check it please?

stephenfin:
> ah, sorry. I didn't merge yet. flake8 warns this. So I can't merge this.
> Could you check it please?

Done

tk0miya:
Merged!

Issue #4398, opened 2018-01-08
Proposal: Change release and tag model


tk0miya:
### Problem
- The document of sphinx is published as x.y.z, not x.y
    - ex. http://www.sphinx-doc.org/en/1.6.5/

#### Expected results
- All the documents are published as x.y

### Proposal

- Add a tag `x.y` to latest release of `x.y` series
    - The tag will be moved on each minor release
- Give a `x.y.0` version to first major release package (cf. 1.7.0)
    - So far the first major releases don't have revision number (cf. 1.5, 1.6)
    - It reduces confusion of `x.y` series with `x.y` release
    - It enables to follow semver rule
- Rename release tags to `releases/x.y.z`.
    - It also reduces confusion of series with release

### Not to do

- This does not change branch models
    - After this change, we continue to use both `stable` and `master`
    - New branch model does not related with our document management model. So it is not needed to change.
    - We don't want to maintain the old releases (at least now)

### Related issues

- #4211

tk0miya:
@shimizukawa @stephenfin @jfbu This is my proposal for branch and tag model discussed at #4211.
What do you think? IMO, it is simple and easy to go for us to arrange tag rules.

shimizukawa:
+1.

On our development scale, I think that it is appropriate to operate only the 2 branches rather than having many past branches.

stephenfin:
A (soft 😄) -1.

I completely understand the reason not to have many past branches: we simply do not have the bandwidth. Given that you folks are the maintainers, any change here is going to overwhelmingly impact you. However, tags are supposed to represent a single point in time and I'm not sure about moving them. To me, it sounds like force pushing. I'd like to propose one of the two below options:

- We use `stable/X.Y` branches but we emphasise that only the latest `stable/X.Y` branch is supported. Any pull requests to older branches can be closed by a bot. If someone wants to maintain older branches in the future, we can re-evaluate this.
- We keep things as they are and simply add a "this is not the latest version of Sphinx" message to the top of old pages. This requires an extra click from the user but will solve my biggest issue.

I'd much rather the first option, but it's not really my call.

PS: Thanks for following up on this, @tk0miya 😄

jfbu:
[...]

PR #4565, opened 2018-02-06
Add procedures for beta and final releases

tk0miya:
This describes release procedure for #4398.

### DIFF
--- a/utils/release-checklist
+++ b/utils/release-checklist
@@ -5,7 +5,7 @@ for stable releases
 -------------------
 
 * open https://travis-ci.org/sphinx-doc/sphinx/branches and check **X.Y** branch is green
-* Run `git status` and check nothing changed
+* Run `git fetch; git status` and check nothing changed
 * `python utils/bump_version.py X.Y.Z`
 * Check diff by `git diff`
 * Edit CHANGES if empty section exists
@@ -29,7 +29,7 @@ for first beta releases
 -----------------------
 
 * open https://travis-ci.org/sphinx-doc/sphinx/branches and check **master** branch is green
-* Run `git status` and check nothing changed
+* Run `git fetch; git status` and check nothing changed
 * Run `python setup.py extract_messages`
 * Run `(cd sphinx/locale; tx push -s)`
 * `python utils/bump_version.py X.Y.0b1`
@@ -59,7 +59,7 @@ for other beta releases
 -----------------------
 
 * open https://travis-ci.org/sphinx-doc/sphinx/branches and check **X.Y** branch is green
-* Run `git status` and check nothing changed
+* Run `git fetch; git status` and check nothing changed
 * `python utils/bump_version.py X.Y.0bN`
 * Check diff by `git diff`
 * Edit CHANGES if empty section exists
@@ -83,7 +83,7 @@ for major releases
 ------------------
 
 * open https://travis-ci.org/sphinx-doc/sphinx/branches and check **X.Y** branch is green
-* Run `git status` and check nothing changed
+* Run `git fetch; git status` and check nothing changed
 * Run `(cd sphinx/locale; tx pull -a -f)`
 * Run `python setup.py compile_catalog`
 * Run `git add sphinx`
--- a/utils/release-checklist
+++ b/utils/release-checklist
@@ -1,19 +1,11 @@
 Release checklist
 =================
 
+for stable releases
+-------------------
+
 * open https://travis-ci.org/sphinx-doc/sphinx/branches and check **X.Y** branch is green
 * Run `git status` and check nothing changed
-* Update sphinx/locale/sphinx.pot if first major release (beta1)
-
-  * Run `python setup.py extract_messages`
-  * Run `(cd sphinx/locale; tx push -s)`
-
-* Update sphinx/locale/<lang>/ files if final major release ...
-
-  * Run `(cd sphinx/locale; tx pull -a -f)`
-  * Run `python setup.py compile_catalog`
-  * Run `git commit -am 'Update message catalogs'`
-
 * `python utils/bump_version.py X.Y.Z`
 * Check diff by `git diff`
 * Edit CHANGES if empty section exists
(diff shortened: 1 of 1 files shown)


utils/release-checklist
```diff
@@ -1,37 +1,112 @@
...
+* Add new version/milestone to tracker categories
+* Write announcement and send to sphinx-dev, sphinx-users and python-announce
 
-  * Run `python setup.py extract_messages`
-  * Run `(cd sphinx/locale; tx push -s)`
+for first beta releases
+-----------------------
 
-* Update sphinx/locale/<lang>/ files if final major release ...
+* open https://travis-ci.org/sphinx-doc/sphinx/branches and check **master** branch is green
```
stephenfin, line 31: We could fix this by making `master` a protected branch in GitHub settings. There is documentation [here](https://help.github.com/articles/about-protected-branches/) about that. Could we do that first?

utils/release-checklist
```diff
@@ -1,37 +1,112 @@
...
+* Run `git status` and check nothing changed
```
stephenfin: This isn't valid rST. Also, you should use `git fetch` first

stephenfin (changes requested):

utils/release-checklist
tk0miya: Ah, I know that and also understand its goodness. On the other hand, I usually update CHANGES file and push it master branch directly. So the restriction is a bit boring to me...

utils/release-checklist
tk0miya: Where is invalid?

I fixed about `git fetch`.

stephenfin: Well, not invalid. It's just that the default role has no significance here. That's noted in the docs:

    The default role (`content`) has no special meaning by default. You are free to use it for
    anything you like, e.g. variable names; use the default_role config value to set it to a known
    role.

I think it would be better to format it as a literal with double backticks.

utils/release-checklist
stephenfin: Well then that sounds like an issue we should resolve. It shouldn't be so hard to modify CHANGES. Could you add a TODO to change this policy once we figure out a better tool to help us here?

utils/release-checklist
tk0miya: You're right. fixed.

utils/release-checklist
tk0miya: No, this is release manual. It's not good to note a TODO.
Please file a new issue please.

stephenfin: It is good. If we put it here, all maintainers know that we consider this part of the process to be suboptimal and that we want to improve it. If we put it in the bug tracker, it will be forgotten about. This is not a bug or RFE for Sphinx - it's a process improvement we'd like to see.

There are two TODOs.

- Find a way to ensure CHANGES updated with every pull request. This probably means modifying pull requests or using something other than a single CHANGES file.
- Once that's done, turn on protected branches

That's my thinking anyway. Protected branches make like a _lot_ easier for us, and we really should use them.

tk0miya: > If we put it here, all maintainers know that we consider this part of the process to be suboptimal and that we want to improve it.

Who is "all maintainers"? This file is referred only on release operation. So it will be sawn by only me.

>That's my thinking anyway. Protected branches make like a lot easier for us, and we really should use them.

easier? really? I don't think so. It restricts developer's operation to keep repository safe. I understand it is meaningful, but it does not make our work "easier". Is my understanding wrong?

In this case, it forbids to push any commits to master branch even if it is fixing typos or update CHANGES. Is it really helpful?
If we are in large team, I agree it is very useful and we should use it ASAP. But our team is very small and inactive. Usually only I and jfbu are working.

Please let me know the advantage of protected branch **for this project**.

Note: force-pushing is already disabled. But other protection like "require status check" is not enabled.

stephenfin: Before I address anything, I should point out that I'm sensing a little anger here. If so, I apologise for anything I might have said that has offended. To be clear, I'm hoping only to state my opinion and see if you agree or disagree. If you think I'm being rude, please let me know.

---

In the OpenStack world, we submit _everything_ though our code review tool, Gerrit, whether it's a huge new feature or a one word typo fix, or whether the project has hundreds of contributors or a handful. Does this introduce some overhead? Yes, of course. However, there are great advantages to this:

- Better visibility. Everyone can see the changes going in, which helps them feel more involved
- Better quality. Because everyone can see the changes going in, they have a chance to review those changes. In addition, the CI runs on everything, preventing broken master. This means users can always deploy from Git if they want cutting edge (e.g. because a bugfix isn't included in a release yet).
- Consistency. At the moment, some changes (like this one) are submitted via a PR. Others are pushed directly to master. The determination for which is which is not clear.

In addition, there are multiple maintainers and, while only two are consistently committing at the moment, surely you want this to grow in the future? Having a policy like this will help massively with onboarding these future maintainers.

Speaking from personal experience, I am not (yet!) a maintainer. As such, I have to submit my changes via pull request. However, this is a good thing. This has led to some very important comments from you and others. Likewise, I hope my comments have helped improve your own patches. I have even left comments about typos and CHANGES which you have incorporated. Any overhead of submitting PRs and waiting for CI is offset by the improvements outlined above.

stephenfin: Perhaps if we are able to make `CHANGES` easier to modify, we would solve one of the things that requires you to push directly to master. What do you think of using [`reno`](https://pypi.python.org/pypi/reno), @tk0miya?

tk0miya: I also have to apologize you that I was a bit frustrating in that day. I feel text based communication is very difficult again. It might come from my poor English, or a lack of mutual understanding. Anyway, we have to talk more and more to understand each other. I think you're good contributor for Sphinx. I'd still like to continue to work with you.

At last, I have to say you that I never feel you're rude. I feel you're polite man :-).

---

Thank you for explanation. I understand the advantages of branch protection you said, in general. But I wonder who wants them to this project. I know our development is inconsistent. But we have not so bad visibility and quality.

TBH, I'm afraid of take overhead of my work. Now I'm going to fix bugs, investigate troubles, improve features, refactor implements, answer questions, review PRs, update docs and do any other for Sphinx. In fact, last August and November, this project was almost stopped because I was absent. So I don't like to increase costs of my work more than now.

Anyway, I should not determine such an important rules without agreement from other maintainers (I'm not a leader!). Could you propose the idea as a new issue please (like about reno)? It should not be discussed as a comment tree of this PR. I'll agree other maintainers say okay to your proposal.

tk0miya: In addition, do you think this discussion is a blocker of this PR? I think it is another topic. So I'd like to merge this if possible.

stephenfin: > I also have to apologize you that I was a bit frustrating in that day. I feel text based communication is very difficult again. It might come from my poor English, or a lack of mutual understanding. Anyway, we have to talk more and more to understand each other. I think you're good contributor for Sphinx. I'd still like to continue to work with you.

Good to hear 😄 It is not problem. Just making sure I'm not being too annoying, heh

> In addition, do you think this discussion is a blocker of this PR? I think it is another topic. So I'd like to merge this if possible.

Good point, it is not. Let's discuss this in another issue 😄 The rest of the change looks good to me 👍

tk0miya: Thanks :-)

PR #4570, opened 2018-02-07
Unify expression of terms in the log message

cocoatomo:
Subject: Unify expression of terms in the log message

### Feature or Bugfix
- Bugfix

### Purpose
- Unify expression of terms in the log message

### Detail
- Replace `Source input` with `source_input`

### Relates
- N/A


tk0miya:
Merged!

PR #4579, opened 2018-02-08
Fix #4564: rst_prolog/rst_epilog not working

### LINKED ISSUE #4564: rst_prolog/rst_epilog not working on master
Subject: ``rst_prolog`` and ``rst_epilog`` are not added to files

### Problem
A bit of git bisect shows that 2b6c8b33be0e66160c6f5d14fd78c7a47e3389b0 is the problematic commit. I guess that somehow it interacts badly with #2445.

Create a new project with ``sphinx-quickstart`` and to ``conf.py`` assign some code to ``rst_prolog``/``rst_epilog``, say
```python
rst_prolog = ".. thisShouldGiveAWarning::\n\n"
rst_epilog = rst_prolog
```

#### Error logs / results
With 2b6c8b33be0e66160c6f5d14fd78c7a47e3389b0 the code won't have effect. With any version without the commit (including the 1.7-release branch) it works as expected.

tk0miya:
refs: #4564

### DIFF
--- a/sphinx/registry.py
+++ b/sphinx/registry.py
@@ -281,11 +281,13 @@ class SphinxComponentRegistry(object):
             parser.set_application(app)
         return parser
 
-    def add_source_input(self, filetype, input_class):
-        # type: (unicode, Type[Input]) -> None
-        if filetype in self.source_inputs:
-            raise ExtensionError(__('source_input for %r is already registered') % filetype)
-        self.source_inputs[filetype] = input_class
+    def add_source_input(self, input_class):
+        # type: (Type[Input]) -> None
+        for filetype in input_class.supported:
+            if filetype in self.source_inputs:
+                raise ExtensionError(__('source_input for %r is already registered') %
+                                     filetype)
+            self.source_inputs[filetype] = input_class
 
     def get_source_input(self, filetype):
         # type: (unicode) -> Type[Input]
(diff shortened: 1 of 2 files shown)


sphinx/registry.py
```diff
@@ -281,11 +281,13 @@ def create_source_parser(self, app, filename):
             parser.set_application(app)
         return parser
 
-    def add_source_input(self, filetype, input_class):
-        # type: (unicode, Type[Input]) -> None
-        if filetype in self.source_inputs:
-            raise ExtensionError(__('source_input for %r is already registered') % filetype)
-        self.source_inputs[filetype] = input_class
+    def add_source_input(self, input_class):
```
stephenfin, line 284: Doesn't this need a deprecation period and release note, given that we're changing behaviour on a public interface?

stephenfin (changes requested):

sphinx/registry.py
tk0miya: This is not a public interface because methods of `app.registry` is not documented. Nobody knows about `app.registry`. But I agree this is breaking change for hackers.
It might be better to move this to 1.7.

stephenfin: We should probably rename `app.registry` to `app._registry` to emphasise this.

Personally, I'm fine with just dropping it. I'm just making sure we're not doing anything wrong, heh

tk0miya: Python does not have strict access control, so renaming is not effect. In fact, some developers had accessed underscored members and methods once.
So I feel it is meaningless so much. And I personally prefer not to use underscored methods.

(I know it declares the methods and members are private obviously.)

tk0miya:
I withdraw this for moving this to 1.7-release.

PR #4584, opened 2018-02-08
Report true file location of doctests

Zac-HD:
This pull replaces #4428 (which itself replaced #4243), and closes issue #4223. 

In short, the locations reported for doctests that have been included from another file (eg a docstring) were given as the location of the *output* document, which is of no use at all when trying to find and fix a failing test.  This pull fixes that, and also admits that we don't know the true line number of anything in a docstring, which docutils gives relative to the processed docstring instead of the file.

Future improvements could include Python 2 support (underlying bug), and finding line numbers - neither of which I have time for before 1.7.  

CC @tk0miya - I've fixed most of your comments on the last pull, but left `filename` as `str` instead of `unicode`, because that's what Mypy requires when passing it on to `doctest.DocTest`.  Ready to merge again?


tk0miya:

PR #4585, opened 2018-02-10
EXAMPLES: move Arcade to rtd section, add aria2

vedranmiletic:
Subject: Minor fix and an addition to EXAMPLES

### Feature or Bugfix
- Arcade moved
- aria2 added


tk0miya:
Thanks!

PR #4590, opened 2018-02-10
Doc rework (user guide, part 2)

stephenfin:
Continue building up the user guide by adding a tutorial document. This is based off a cleaned up tutorial document.

You should view this on a commit-by-commit basis to best understand the individual changes.


doc/usage/quickstart.rst
```diff
@@ -145,6 +149,8 @@ an argument to see which targets are available.
    toolchain for you.
 
 
+.. todo:: Move this whole section into a guide on rST or directives
```
stephenfin, line 152: I have all these `todo`\'s resolved in follow-up patches. I've submitted this separately to keep things easy to review

doc/usage/quickstart.rst
```diff
@@ -300,15 +318,14 @@ features of intersphinx.
 More topics to be covered
 -------------------------
 
-- :doc:`Other extensions <extensions>`:
-
```
tk0miya: Did you check the rendered page? I think this empty line is needed.

doc/usage/quickstart.rst
```diff
@@ -320,6 +337,6 @@ More topics to be covered
        another directory, the :term:`configuration directory`.  Refer to the
        :program:`sphinx-build man page <sphinx-build>` for more information.
 
-.. |more| image:: more.png
+.. |more| image:: ../_static/more.png
```
tk0miya: You can use `/_static/more.png` instead. Sphinx considers a absolute path as a relative path from srcdir.

doc/usage/quickstart.rst
tk0miya: Reasonable, but this todo is useful only for us. What about comment it out?

tk0miya (changes requested):
LGTM with nits

doc/usage/quickstart.rst
stephenfin: I don't think it matters. I'll be fixing this literally within days. Can we just keep this as is?

doc/usage/quickstart.rst
stephenfin: Done

doc/usage/quickstart.rst
stephenfin: Done

stephenfin:
This is done now. As noted, the TODOs will be removed very soon. I think it's important that they're left there just in case I disappear of the face of the earth or something, heh

tk0miya:
I said to replace TODOs by comments. Why do you want to show TODO for users?

stephenfin:
> I said to replace TODOs by comments. Why do you want to show TODO for users?

So users will know this is a WIP. They will only be there for a few days. In addition, I don't want to have to rework the ~20 patches that I have in the series after these (I'm submitting these separately).

tk0miya:
I don't like showing TODO block for users. But it says "time flies like an arrow". So let's go with as is.

tk0miya:
Merged. Thanks,

stephenfin:
Yay. Thanks! :+1:

PR #4592, opened 2018-02-11
Fix #4581: apidoc: Excluded modules still included

### LINKED ISSUE #4581: Excluded modules still included by apidoc
Subject: I have a command which excludes some modules and that works with Sphinx 1.6.6 but fails with 1.6.7

### Problem
I think the issue I'm seeing was caused by https://github.com/sphinx-doc/sphinx/pull/4449

```
git clone https://github.com/kiwitcms/Kiwi
cd Kiwi/
pip install -r requirements/devel.txt 
pip install -r requirements/mysql.txt
pip install --update Sphinx ## to make sure we're using 1.6.7
cd docs/
make apidoc
git diff
```

The Makefile specifies:

```
apidoc:
        $(SPHINXAPIDOC) -o source/modules/ -f -e -M -T ../ ../manage.py ../setup.py \
                        ../*/tests.py ../*/tests/*.py ../*/test_*.py \
                        ../settings/test/*.py ../*/migrations/*.py ../tcms_api/setup.py
```

#### Expected results

Excluded modules are not included by apidoc

- https://github.com/kiwitcms/Kiwi

### Environment info
- OS: Linux, RHEL 7.4
- Python version: 3.6
- Sphinx version: 1.6.7

tk0miya:
refs: #4581

### DIFF
--- a/sphinx/ext/apidoc.py
+++ b/sphinx/ext/apidoc.py
@@ -108,8 +108,8 @@ def create_module_file(package, module, opts):
     write_file(makename(package, module), text, opts)
 
 
-def create_package_file(root, master_package, subroot, py_files, opts, subs, is_namespace):
-    # type: (unicode, unicode, unicode, List[unicode], Any, List[unicode], bool) -> None
+def create_package_file(root, master_package, subroot, py_files, opts, subs, is_namespace, excludes=[]):  # NOQA
+    # type: (unicode, unicode, unicode, List[unicode], Any, List[unicode], bool, List[unicode]) -> None  # NOQA
     """Build the text of the file and write the file."""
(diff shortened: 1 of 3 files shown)


sphinx/ext/apidoc.py
```diff
@@ -108,8 +108,8 @@ def create_module_file(package, module, opts):
     write_file(makename(package, module), text, opts)
 
 
-def create_package_file(root, master_package, subroot, py_files, opts, subs, is_namespace):
-    # type: (unicode, unicode, unicode, List[unicode], Any, List[unicode], bool) -> None
+def create_package_file(root, master_package, subroot, py_files, opts, subs, is_namespace, excludes):  # NOQA
```
tk0miya: I think new argument should be optional.

PR #4594, opened 2018-02-11
Change interface of add_source_parser() and add add_source_suffix()

tk0miya:
### Feature or Bugfix
- Refactoring
- Bugfix (in master branch: see #4564)

### Purpose
- Separate registering of suffix and parser

### Detail
- Please read the diff of `doc/extdev/appapi.rst`.

### DIFF
--- a/doc/extdev/parserapi.rst
+++ b/doc/extdev/parserapi.rst
@@ -3,6 +3,35 @@
 Parser API
 ==========
 
+`The docutils documentation describes`__ parsers as follows:
+
+    The Parser analyzes the input document and creates a node tree
+    representation.
+
+__ http://docutils.sourceforge.net/docs/dev/hacking.html#parsing-the-document
+
+In Sphinx, the parser modules works as same as docutils.  The parsers are
+registered to Sphinx by extensions using Application APIs;
+:meth:`Sphinx.add_source_suffix()` and :meth:`Sphinx.add_source_parsers()`.
+
+The *source suffix* is a mapping from file suffix to file type.  For example,
+``.rst`` file is mapped to ``'restructuredtext'`` type.  Sphinx uses the
+file type to looking for parsers from registered list.  On searching,
+Sphinx refers to the ``Parser.supported`` attribute and picks up a parser
+which contains the file type in the attribute.
+
+The users can override the source suffix mappings using
+:confval:`source_suffix` like following::
+
+    # a mapping from file suffix to file types
+    source_suffix = {
+        '.rst': 'restructuredtext',
+        '.md': 'markdown',
+    }
+
+You should indicate file types your parser supports. This will allow users
+to configure their settings appropriately.
+
 .. module:: sphinx.parsers
 
 .. autoclass:: Parser
--- a/sphinx/registry.py
+++ b/sphinx/registry.py
@@ -239,10 +239,23 @@ class SphinxComponentRegistry(object):
         else:
             self.source_suffix[suffix] = filetype
 
-    def add_source_parser(self, suffix, parser):
-        # type: (unicode, Type[Parser]) -> None
-        logger.debug('[app] adding search source_parser: %r, %r', suffix, parser)
-        self.add_source_suffix(suffix, suffix)
+    def add_source_parser(self, *args):
+        # type: (Any) -> None
+        logger.debug('[app] adding search source_parser: %r', args)
+        if len(args) == 1:
+            # new sytle arguments: (source_parser)
+            suffix = None       # type: unicode
+            parser = args[0]    # type: Type[Parser]
+        else:
+            # old style arguments: (suffix, source_parser)
+            warnings.warn('app.add_source_parser() does not support suffix argument. '
+                          'Use app.add_source_suffix() instead.',
+                          RemovedInSphinx30Warning)
+            suffix = args[0]
+            parser = args[1]
+
+        if suffix:
+            self.add_source_suffix(suffix, suffix)
 
         if len(parser.supported) == 0:
             warnings.warn('Old source_parser has been detected. Please fill Parser.supported '
@@ -259,8 +272,9 @@ class SphinxComponentRegistry(object):
 
         # also maps suffix to parser
         #
-        # This allows parsers not having ``supported`` filetypes.
-        self.source_parsers[suffix] = parser
+        # This rescues old styled parsers which does not have ``supported`` filetypes.
+        if suffix:
+            self.source_parsers[suffix] = parser
 
     def get_source_parser(self, filetype):
         # type: (unicode) -> Type[Parser]
@@ -398,13 +412,13 @@ class SphinxComponentRegistry(object):
 def merge_source_suffix(app):
     # type: (Sphinx) -> None
     """Merge source_suffix which specified by user and added by extensions."""
-    for suffix in app.registry.source_suffix:
+    for suffix, filetype in iteritems(app.registry.source_suffix):
         if suffix not in app.config.source_suffix:
-            app.config.source_suffix[suffix] = suffix
+            app.config.source_suffix[suffix] = filetype
         elif app.config.source_suffix[suffix] is None:
             # filetype is not specified (default filetype).
             # So it overrides default filetype by extensions setting.
-            app.config.source_suffix[suffix] = suffix
+            app.config.source_suffix[suffix] = filetype
 
     # copy config.source_suffix to registry
     app.registry.source_suffix = app.config.source_suffix
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,8 @@ Deprecated
   ``sphinx.util.import_object()`` instead.
 * Drop function based directive support.  For now, Sphinx only supports class
   based directives.
+* ``Sphinx.add_source_parser()`` has changed;  the *suffix* argument has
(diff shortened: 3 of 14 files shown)


doc/extdev/parserapi.rst
```diff
@@ -3,6 +3,35 @@
...
+    The Parser analyzes the the input document and creates a node tree
+    representation.
+
+__ http://docutils.sourceforge.net/docs/dev/hacking.html#parsing-the-document
+
+In Sphinx, the parser modules works as same as docutils.  The parsers are
+registered to Sphinx by extensions using Application APIs;
+:meth:`Sphinx.add_source_suffix()` and :meth:`Sphinx.add_source_parsers()`.
+
+The source_suffix is a mapping from file suffix to file type.  For example,
```
stephenfin: s/source_suffix/:confval:`source_suffix`/ ?

doc/extdev/parserapi.rst
```diff
@@ -3,6 +3,35 @@
...
+``.rst`` file is mapped to ``'restrcturedtext'`` type.  Sphinx uses the
```
stephenfin: s/restrcturedtext/restructuredtext/

doc/extdev/parserapi.rst
```diff
@@ -3,6 +3,35 @@
...
+The users can override the source_suffix mappings using
+:confval:`source_suffix` like following::
+
+    # a mapping from file suffix to file types
+    source_suffix = {
+        '.rst': 'restrcturedtext',
+        '.md': 'markdown',
+    }
+
+So please let users know what file types your parser supports.  Then users
```
stephenfin: This paragraph doesn't make sense. I _think_ you mean:

> You will need to configure this setting in order to use a different file types for a given parser.

sphinx/registry.py
```diff
@@ -239,10 +239,23 @@ def add_source_suffix(self, suffix, filetype):
         else:
             self.source_suffix[suffix] = filetype
 
-    def add_source_parser(self, suffix, parser):
-        # type: (unicode, Type[Parser]) -> None
-        logger.debug('[app] adding search source_parser: %r, %r', suffix, parser)
-        self.add_source_suffix(suffix, suffix)
+    def add_source_parser(self, *args):
```
stephenfin, line 242: nit: How about doing this instead?

    empty_parser = object()

    def add_source_parser(self, suffix, parser=empty_parser):
        if parser != empty_parser:
            # emit warning here

It would keep a similarly (clear) function style and allow you to do the same thing. It works because every `object` instance is unique. You could also use `parser=None`, but if the user called this with `None` intentionally, they wouldn't see the issue

stephenfin (changes requested):
Couple of typos plus some helpful suggestions

sphinx/registry.py
tk0miya: Do you mean like this? I feel `parser = suffix` is strange.
```
def add_source_parser(self, suffix, parser=empty_parser):
    if parser != empty_parser:
        # old style arguments: (suffix, source_parser)
        warnings.warn('app.add_source_parser() does not support suffix argument. '
                      'Use app.add_source_suffix() instead.',
                      RemovedInSphinx30Warning)
    else:
        # new sytle arguments: (source_parser)
        suffix = None
        parser = suffix
```

>You could also use parser=None, but if the user called this with None intentionally, they wouldn't see the issue

I don't mind. If developers gives such value, Sphinx will be crashed (without this change).

doc/extdev/parserapi.rst
tk0miya: I meant that the author of Parser should tell a file type supported by the Parser to users.
The users need to know it to configure `source_suffix` variable.

I think the target of this document is a third party developer. So your paragraph is not good for them.

doc/extdev/parserapi.rst
tk0miya: No, this `source_suffix` is a concept. Both users and extensions can modify it using `:confval:source_suffix` and `add_source_suffix()` API. I'd like to describe it here because the authors of parser module should understand how it works.

doc/extdev/parserapi.rst
tk0miya: done

tk0miya:
Thank you for comment!

doc/extdev/parserapi.rst
stephenfin: That's fair. I'd probably remove the underscore in that case and make it italic so it stands out:

    *source suffix*

Otherwise though, this is fine

doc/extdev/parserapi.rst
stephenfin: Oh, OK. I understand now. This could still do with clarification though, given that it's already confused one person (me!). How about:

    You should indicate file types your parser supports. This will allow users to configure their settings appropriately.

Or something like that

sphinx/registry.py
stephenfin: Ohhh, wait a second. I mixed up the order of the parameters. `suffix` is the first parameter - not the second. You can ignore this comment so 😄

doc/extdev/parserapi.rst
tk0miya: Sounds good.

doc/extdev/parserapi.rst
tk0miya: Very good, thanks!

sphinx/registry.py
tk0miya: Okay :-)

tk0miya (commented):
Updated.

doc/extdev/parserapi.rst
```diff
@@ -3,6 +3,35 @@
...
+file type to looking for parsers from registered list.  On searching,
+Sphinx refers ``Parser.supported`` attribute, and picks up a parser which
+contains the file type in the attribute.
+
+The users can override the source suffix mappings using
+:confval:`source_suffix` like following::
+
+    # a mapping from file suffix to file types
+    source_suffix = {
+        '.rst': 'restrcturedtext',
```
stephenfin: Oops. I assume you meant `restructuredtext`?

doc/extdev/parserapi.rst
```diff
@@ -3,6 +3,35 @@
 Parser API
 ==========
 
+`The document of docutils describes`__ about the parsers like following:
+
+    The Parser analyzes the the input document and creates a node tree
```
stephenfin: s/the the/the/

doc/extdev/parserapi.rst
```diff
@@ -3,6 +3,35 @@
 Parser API
 ==========
 
+`The document of docutils describes`__ about the parsers like following:
```
stephenfin: The docutils documentation describes parsers as follows:

doc/extdev/parserapi.rst
```diff
@@ -3,6 +3,35 @@
...
+__ http://docutils.sourceforge.net/docs/dev/hacking.html#parsing-the-document
+
+In Sphinx, the parser modules works as same as docutils.  The parsers are
+registered to Sphinx by extensions using Application APIs;
+:meth:`Sphinx.add_source_suffix()` and :meth:`Sphinx.add_source_parsers()`.
+
+The *source suffix* is a mapping from file suffix to file type.  For example,
+``.rst`` file is mapped to ``'restructuredtext'`` type.  Sphinx uses the
+file type to looking for parsers from registered list.  On searching,
+Sphinx refers ``Parser.supported`` attribute, and picks up a parser which
```
stephenfin: refers to the ``Parser.supported`` attribute and picks

(missing 'to the'. Drop the comma before 'and')

stephenfin (commented):
I missed a few things 😞

doc/extdev/parserapi.rst
tk0miya: done

doc/extdev/parserapi.rst
tk0miya: It seems this is a typo of original text.
https://github.com/docutils-mirror/docutils/blob/master/docs/dev/hacking.txt#L69

doc/extdev/parserapi.rst
tk0miya: done

doc/extdev/parserapi.rst
tk0miya: done

tk0miya (commented):
Fixed again

sphinx/environment/__init__.py
```diff
@@ -481,7 +482,7 @@ def get_outdated_files(self, config_changed):
                     changed.add(docname)
                     continue
                 # finally, check the mtime of dependencies
-                for dep in self.dependencies[docname]:
+                for dep in self.dependencies[None].union(self.dependencies[docname]):
```
tk0miya: oh, I noticed this change is not related with this PR.
I have to remove this.

tk0miya (commented):
I will merge this soon after removal.

tk0miya:
Merged.
@stephenfin I'm merging this now to step forward. Please let me know if you still have any opinion for this.

stephenfin:
Yup, I was on vacation. This looks good to me. Thanks for the follow-ups

tk0miya:
Have a good vacation!

Issue #4439, opened 2018-01-15
Make script_files deprecated to?


Blendify:
Subject: script_files should be made depricated

### Problem
Since `css_files` was made deprecated in 83d3fd2c365f314877086c72eef79df2c43da69d Should `script_files` also be made deprecated?

#### Procedure to reproduce the problem
N/a

#### Error logs / results
N/a

#### Expected results
`script_files` marked as deprecated.

### Reproducible project / your project
N/a

### Environment info
- OS: N/a
- Python version: N/a
- Sphinx version: 1.6.x

tk0miya:
Yes, I think so. But we still modify it inside HTML theme. so we have to refactor it first.
So I marked this as 1.8 milestone.

tk0miya:
Modified by #5021.
Thanks,

Issue #2442, opened 2016-04-12
Setting html_context['css_files'] renders add_stylesheet() inoperative


mgeier:
When I use something like this in my `conf.py`:

``` python
html_static_path = ['_static']
html_context = {
    'css_files': ['_static/my_css_file.css'],
}
```

... and an extension uses [add_stylesheet()](http://www.sphinx-doc.org/en/stable/extdev/appapi.html#sphinx.application.Sphinx.add_stylesheet), latter has no effect.

Is this intentional?

I would expect the `css_files` mentioned in `conf.py` to be _appended_ to the list of CSS files from the theme and from extensions.

tk0miya:
I don't know the behavior is intentional or not.

Now sphinx provides `add_stylesheet()` API to add css files.
But `html_context["css_files"]` is undocumented internal key to manage css files. I think it is not proper way to add CSS files.
So your problem is unexpected, but undefined behavior.

Any reason to use `html_context["css_files"]` instead of `add_stylesheet()`?

mgeier:
I think the one obvious way for a "normal" user to provide custom settings to Sphinx is defining variables (e.g. `html_title` or `latex_elements`) in `conf.py`.

OTOH, the one obvious way for "extension writers" to extend (and "enhance") the user-specified settings is to use the [application API](http://www.sphinx-doc.org/en/stable/extdev/appapi.html) from the extension's `setup()` function, e.g. by using `app.add_stylesheet()`.

I know it's also possible for "normal" users to define a `setup()` function directly in their `conf.py`, getting access to the full application API. This is great, but I think it should only be used for very specialized use cases. It shouldn't be the "one obvious way to do it". It shouldn't become the default for "normal" users. Casual users should never have to worry about a `setup()` function.

_Adding_ style definitions to the existing CSS definitions is IMHO a quite "normal" thing to do for a "normal" user, therefore it should be possible by simply assigning to a variable in `conf.py`.
It seems to me that `html_context["css_files"]` is a reasonable place to do that.
But I would, as I said initially, expect the user settings to be _appended_ to the settings given by Sphinx, the theme and any extensions. This would be the same behavior as when assigning to `latex_elements['preamble']` (which doesn't _replace_ the existing preamble but _appends_ to it).

If for some technical reason `html_context` cannot be modified as I suggested in my original comment (which I doubt very much), a new configuration variable should probably be created.

[...]

PR #4595, opened 2018-02-11
Add html stylesheets

tk0miya:
### Feature or Bugfix
- Feature
- Refactor

### Purpose
- Make HTML output more easier to customize by users. So far, users need to call APIs to add own CSS and JS.
- This enables to add them simply without any codes.

### Relates
- #4439
- #2442

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -59,9 +59,6 @@ Important points to note:
   Note that the current builder tag is not available in ``conf.py``, as it is
   created *after* the builder is initialized.
 
-.. seealso:: Additional configurations, such as adding stylesheets,
-   javascripts, builders, etc. can be made through the :doc:`/extdev/appapi`.
-
 
 General configuration
 ---------------------
@@ -878,6 +875,14 @@ that use Sphinx's HTMLWriter class.
       The dotfiles in the extra directory will be copied to the output directory.
       And it refers :confval:`exclude_patterns` on copying extra files and
       directories, and ignores if path matches to patterns.
+
+.. confval:: html_stylesheets
+
+   .. versionadded:: 1.8
+
+.. confval:: html_javascripts
+
+   .. versionadded:: 1.8
 
 .. confval:: html_last_updated_fmt
 
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -797,6 +797,9 @@ that use Sphinx's HTMLWriter class.
    theme.  If you only want to add or override a few things compared to the
    theme's stylesheet, use CSS ``@import`` to import the theme's stylesheet.
 
+   .. deprecated:: 1.8
+      Use :confval:`html_stylesheets` instead.
+
 .. confval:: html_title
 
    The "title" for HTML documentation generated with Sphinx's own templates.
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -878,9 +878,23 @@ that use Sphinx's HTMLWriter class.
 
 .. confval:: html_stylesheets
 
+   A list of CSS files.  The entry must be a *filename* string or a tuple
+   containing the *filename* string, the *alternate* (of boolean type) and
+   *title* (a string).  The *filename* must be relative to the
+   :confval:`html_static_path`, or a full URI with scheme like
+   ``http://example.org/style.css``.  It defaults to an empty list.
+
+   For more information about *alternate*, refer to the `MDN web docs`__.
+
+   __ https://mdn.io/Web/CSS/Alternative_style_sheets
+
    .. versionadded:: 1.8
 
 .. confval:: html_javascripts
+
+   A list of JavaScript *filename*.  The *filename* must be relative to the
+   :confval:`html_static_path`, or a full URI with scheme.  It defaults to
+   an empty list.
 
    .. versionadded:: 1.8
 
@@ -1543,6 +1557,13 @@ the `Dublin Core metadata <http://dublincore.org/>`_.
    It is a list of tuples containing the file name and the title.  This option
    can be used to add an appendix.  If the title is empty, no entry is added
    to :file:`toc.ncx`.  The default value is ``[]``.
+
+.. confval:: epub_stylesheets
+
+   A list of CSS files.  See :confval:`html_stylesheets` for details.  By
+   default, it defaults to the value of ``html_stylesheets``.
+
+   .. versionadded:: 1.8
 
 .. confval:: epub_exclude_files
 
--- a/sphinx/builders/html.py
+++ b/sphinx/builders/html.py
@@ -90,7 +90,7 @@ def get_stable_hash(obj):
 
 
 class CSSContainer(list):
-    """The container of stylesheets.
+    """The container for stylesheets.
 
     To support the extensions which access the container directly, this wraps
     the entry with Stylesheet class.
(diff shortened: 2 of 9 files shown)


sphinx/builders/html.py
```diff
@@ -153,6 +153,39 @@ def __new__(cls, filename, title, rel):
         return self
 
 
+class JSContainer(list):
+    """The container for javascripts."""
```
stephenfin: s/javascripts/JavaScript scripts/

doc/config.rst
```diff
@@ -878,10 +878,24 @@ that use Sphinx's HTMLWriter class.
...
+
+   For more information about *alternate*, refer to the `MDN web docs`__.
+
+   __ https://mdn.io/Web/CSS/Alternative_style_sheets
+
    .. versionadded:: 1.8
 
 .. confval:: html_javascripts
 
+   A list of JavaScript *filename*.  The *filename* must be relative to the
```
stephenfin: A list of JavaScript files.  The entry must be a *filename* string and must be relative...

?

sphinx/registry.py
```diff
@@ -341,6 +347,16 @@ def add_autodoc_attrgetter(self, typ, attrgetter):
         # type: (Type, Callable[[Any, unicode, Any], Any]) -> None
         self.autodoc_attrgettrs[typ] = attrgetter
 
+    def add_javascript(self, filename):
```
stephenfin, line 350: Why do we store this stuff in `Registry`? I thought this object was for global configuration, but scripts and stylesheets only apply to the HTML builders. Couldn't we keep storing this as an attribute of the HTML builders but simply read from the configuration value instead?

stephenfin (commented):
I may need a little more context

sphinx/registry.py
tk0miya: >Couldn't we keep storing this as an attribute of the HTML builders but simply read from the configuration value instead?

This is not a configuration value. The extensions and themes can add JS and CSS files via API. This is an API for them. `app.add_stylesheet()` is also.

About storing, current implementation stores additional files into "class attributes" of HTML builders like following:
```
            StandaloneHTMLBuilder.script_files.append(filename)
```
This is my point of refactoring. Using class attributes is like a global variable. It is a stateful, and cached during the same OS process. I think it is not good behavior. So I'd like to store them into "instance variable".

FYI: On this stage, builder is not instantiated yet. So we have to store them other place temporarily.


Until 1.4 (or 1.5), almost of components are stored into such global settings. In nearly releases, I continue to refactor not to store these components as instance variable. To do that, `sphinx.registry` was added.
(#4605 is one of them).

I know JS and CSS is not sphinx core component. So I agree registry is not best place to store them. I welcome better idea to do that.

Note:
Personally, `app.add_javascript()` and `app.add_documenter()` APIs are also not so good. Because both HTML builder and autodoc are not core components. They are only one of builder (or extension) for Sphinx. so application object should not provide APIs for individual sub-components.
So these should be changed in future (if there is good idea!)

sphinx/builders/html.py
tk0miya: Is the wording usually used? Anyway either is okay to me.

stephenfin: Yup, this would be more common in my experience.

sphinx/registry.py
stephenfin: > I know JS and CSS is not sphinx core component. So I agree registry is not best place to store them. I welcome better idea to do that.
>
> Note:
> Personally, app.add_javascript() and app.add_documenter() APIs are also not so good. Because both HTML builder and autodoc are not core components. They are only one of builder (or extension) for Sphinx. so application object should not provide APIs for individual sub-components.
> So these should be changed in future (if there is good idea!)

Yes, this is my concern **exactly**. It would seem we need a way for extensions to override configuration values that are not related to the `sphinx.application.Application` object. We could allow them to override various values read in from `conf.py` but that will create all types of chaos if done incorrectly (i.e. a misbehaving extension could completely override user-provided configuration). It would seem we are stuck with this API for now.

stephenfin:
OK, once that one typo is corrected, this will be a-ok to me

sphinx/builders/html.py
tk0miya: done

sphinx/registry.py
tk0miya: No, your understanding is not right. These APIs does not related with users configurations basically. They are used to store "components" for Sphinx only. For example, builders, domains, parsers, and so on. JS and CSS files are one of them.

Certainly `add_javascript()` is related with `html_javascripts` which I added in this PR. This is an exception of rules above. `html_javascripts` is a configuration to register JS files to HTML builders. So it also registers "components".
FYI: If both are used at once, all javascripts are enabled, not overrided.


BTW, what I meant at last comment is that sphinx-core is a kind of framework. And HTML builder is one of its component. So sphinx-core should not treat sub-components (like JS and CSS) of these components directly. IMO, it is a responsibility of each component, not application's.

tk0miya: FYI: The extensions can override the users' configuration on runtime if they want.
```
def on_config_inited(app, config):
    config.some_value = 1
    config.other_value = 2

def setup(app):
    app.connect('config-inited', on_config_inited)
```

stephenfin: > No, your understanding is not right. These APIs does not related with users configurations basically. They are used to store "components" for Sphinx only. For example, builders, domains, parsers, and so on. JS and CSS files are one of them.

Sorry, when I said "them" I meant extension developers - not users.

> BTW, what I meant at last comment is that sphinx-core is a kind of framework. And HTML builder is one of its component. So sphinx-core should not treat sub-components (like JS and CSS) of these components directly. IMO, it is a responsibility of each component, not application's.

Yup, I was actually agreeing with you. It seems I was not clear enough 😄

tk0miya: Okay. Anyway this PR does not change our API at present. We already provide APIs for adding JS and CSS, so this does not change public APIs.

So let's discuss about better APIs in other issue. As I commented above, I always welcome better interfaces.

sphinx/builders/html.py
```diff
@@ -283,10 +320,54 @@ def create_build_info(self):
...
+
+        for filename in self.get_builder_config('javascripts', 'html'):
+            self.add_javascript(filename)
+
+        if self.config.language and self._get_translations_js():
+            self.add_javascript('translations.js')
+
+    def add_javascript(self, filename):
+        # type: (unicode) -> None
+        if '://' not in filename:
```
benjaoming, line 363: What if the filename is `/absolute/path/file.js` ? Isn't there a prettier name to handle this?

sphinx/builders/html.py
```diff
@@ -283,10 +320,54 @@ def create_build_info(self):
...
+            filename = posixpath.join('_static', filename)
```
benjaoming, line 364: I think `_static` is a configurable path?

tk0miya: At present, it is not configurable.

sphinx/builders/html.py
tk0miya: >What if the filename is /absolute/path/file.js ? 

The filename must be a relative path from HTML static paths or URI.
http://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_javascript

>Isn't there a prettier name to handle this?

I don't have good idea.

mgeier:
Here's my comment from https://github.com/sphinx-doc/sphinx/issues/2442#issuecomment-366011772:

I'm not a native speaker, but `html_javascripts` sounds a bit weird to me.
What about `html_javascript_files` or `html_js_files`?

And correspondingly `html_css_files`?

@tk0miya answered (https://github.com/sphinx-doc/sphinx/issues/2442#issuecomment-366214201):

> Note: I don't mind about wording. I'm also not native. But we have to consider that it also accepts URLs of JS (or CSS). I feel `_files` suffix is not good for URLs.

Well they are still *files* on the server, at least conceptually, aren't they?
And in most cases, the URLs will also contain an actual file name. I don't see a problem with this.

I could turn this around and argue: `html_javscripts` might as well be a sequence of strings directly containing JS code ...?

---

Are there any native speakers around who have an opinion on this?

shimizukawa:
+1 for `html_css_files` because it is explained as "A list of CSS files." in description.

tk0miya:
@shimizukawa thank you for comment. Okay, I'll change the name later.

doc/config.rst
```diff
@@ -800,6 +797,9 @@ that use Sphinx's HTMLWriter class.
    theme.  If you only want to add or override a few things compared to the
    theme's stylesheet, use CSS ``@import`` to import the theme's stylesheet.
 
+   .. deprecated:: 1.8
+      Use :confval:`html_stylesheets` instead.
```
tk0miya, line 801: I noticed `html_style` can't replace this. I'll remove this from this PR. Let's discuss in another PR.

tk0miya:
If we use `html_css_files` to config variable, should `app.add_stylesheet()` also be renamed to `app.add_css_file()`?
@mgeier @shimizukawa  What do you think?

mgeier:
Hmmm ... I don't know.
I think it would sound more consistent, but isn't this a backwards compatibility issue?

tk0miya:
Thank you for comment. I'll consider again later.

tk0miya:
Finally, I separated this PR to #4815. In the PR, I renamed the API to `app.add_css_file()`.
So I'm withdrawing this now. I'll work for JS later.

Thank you for reviewing!

PR #4598, opened 2018-02-12
Makefile cleanup

stephenfin:
Remove dead targets from `Makefile`

More information is provided in the commit messages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,6 @@ PYTHON ?= python
 .PHONY: all
 all: clean-pyc clean-backupfiles style-check type-check test
 
-.PHONY: style-check
-style-check:
-	@flake8
-
-.PHONY: type-check
-type-check:
-	mypy sphinx/
-
 .PHONY: clean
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -59,22 +51,22 @@ clean-buildfiles:
...
 	rm -rf .mypy_cache/
 
+.PHONY: style-check
+style-check:
+	@flake8
+
+.PHONY: type-check
+type-check:
+	mypy sphinx/
+
```
tk0miya, line 61: off topic:  I'm using this on my local (especially type-check target). The `:make` command of vim can handle the result easier. So I'd like to keep them as is.
http://vim.wikia.com/wiki/Automatically_open_the_quickfix_window_on_:make

tk0miya (approved):
LGTM

Makefile
stephenfin: Oooh, neat 😄 I generally have a second pane open in `tmux` and just jump between them. Maybe I should investigate `tox` integration for `vim` too

**EDIT:** no intention of removing these quite yet, don't worry

tk0miya:
Merged!

PR #4622, opened 2018-02-15
key changes to 'scheme' in epub3 opf template

turky:
Subject: fix 'epub_scheme' parameter doesn't refrect into content.opf

### Feature or Bugfix
- Bugfix

### Purpose
- epub_scheme paprameter in conf.py is specify identifier scheme. but it doesn't work. This PR fix typo
in opf template in /sphinx/templates/epub3/content.opf_t

Issue #4611, opened 2018-02-13
Run epubcheck as a testing


tk0miya:
### Problem
- Sphinx-1.7.0 generates invalid epub file (see #4608)
- We don't test that generated epub file is correct

#### Expected results
Run epubcheck as a part of testing.

tk0miya:
As a trial, I ran it with following Dockerfile:
```
FROM tk0miya/sphinx-html

RUN rm -rf /var/lib/apt/lists/*
RUN apt update
RUN apt install -y openjdk-8-jre-headless epubcheck
RUN git clone http://github.com/sphinx-doc/sphinx
WORKDIR /docs/sphinx
RUN sphinx-build -b epub doc/ build/epub
RUN java -jar /usr/share/java/epubcheck.jar build/epub/sphinx.epub
```
And I got following result. We have to fix them all before enabling epubcheck.
```
Validating using EPUB version 3.0.1 rules.
ERROR(OPF-049): build/epub/sphinx.epub/content.opf(196,33): Item id 'epub-101' was not found in the manifest.
ERROR(RSC-005): build/epub/sphinx.epub/content.opf(196,33): Error while parsing file 'itemref element idref attribute does not resolve to a manifest item element'.
ERROR(RSC-005): build/epub/sphinx.epub/authors.xhtml(5,64): Error while parsing file 'value of attribute "http-equiv" is invalid; must be a string matching the regular expression "([Dd][Ee][Ff][Aa][Uu][Ll][Tt]\-[Ss][Tt][Yy][Ll][Ee])|([Rr][Ee][Ff][Rr][Ee][Ss][Hh])|([Cc][Oo][Nn][Tt][Ee][Nn][Tt]\-[Tt][Yy][Pp][Ee])"'.
ERROR(RSC-005): build/epub/sphinx.epub/builders.xhtml(5,64): Error while parsing file 'value of attribute "http-equiv" is invalid; must be a string matching the regular expression "([Dd][Ee][Ff][Aa][Uu][Ll][Tt]\-[Ss][Tt][Yy][Ll][Ee])|([Rr][Ee][Ff][Rr][Ee][Ss][Hh])|([Cc][Oo][Nn][Tt][Ee][Nn][Tt]\-[Tt][Yy][Pp][Ee])"'.
[...]

PR #4640, opened 2018-02-18
Show opt:scheme attribute only for EPUB3.1 (refs: #4622)

tk0miya:
refs: #4611 

@turky I investigated about `opt:scheme` again. And I found it is allowed since EPUB3.1, (not EPUB3.0). To generate valid EPUB, I'd like to generate the attribute only for EPUB3.1.
What do you think?

Note: In this PR, I added `epub_version` confval, but it is experimental, and undocumented variable.


tk0miya:
I think this does not harm anybody. So I'm merging this now.

PR #4645, opened 2018-02-18
Also accept classnames in function pointers.

rkapl:
For example, "void (*class::fn)()" would be a valid declaration. Doxygen
(and thus breathe) produces declaration of this kind.


tk0miya:
There are flake8 warning. Could you fix it please?

rkapl:
Flake8 fixed.

jakobandersen:
The declaration looks like C++ due to the ``::``. Also, the C domain has now been rewritten so more complex declarations should now work. If possible, can you test with the 3.x branch and see if there is still a problem? If so, please open an issue.

PR #4653, opened 2018-02-19
Fix error reporting for parameterless ImportErrors

adam-azarchs:
In some cases, a module may raise an `ImportError` without any arguments, e.g.

```python
if not_allowed():
    raise ImportError
```

In this case, the exception has no args.  This is bad practice, but it happens.

Currently, autodoc will crash with
```
IndexError: tuple index out of range
```

Subject: Fix a crash in autodoc.
### Feature or Bugfix
- Bugfix

### Purpose
- Fix a crash on bad code.


tk0miya (changes requested):
+1; could you rebase this into 1.7 branch?
Then I will merge this soon.

adam-azarchs:
Rebased to 1.7 branch.

tk0miya (approved):

tk0miya:
Merged. Thanks!

PR #4656, opened 2018-02-20
Fix #4520: Fix skipping none empty packages

### LINKED ISSUE #4520: Subpackage not in toc
Subject: Subpackage not in toc

### Problem
In [this commit](https://github.com/sphinx-doc/sphinx/commit/023dad6c6697b551a0a6a5f952a895185bce178c) it seems, there was a bug introduce. The bug does is the reverse of what the commit is trying to achieve.
Subpackages in version `Sphinx (sphinx-apidoc) 1.6.6` are only picked up if they DO NOT contain a `__init__.py` file. Which doesn't make them packages but folder.

```shell
mkdir -p parent/child
touch parent/__init__.py
touch parent/child/__init__.py

pip install sphinx==1.6.5
sphinx-apidoc -f --separate -o . .
cat parent.rst | grep Subpackages ; echo $?

pip install sphinx==1.6.6
sphinx-apidoc -f --separate -o . .
cat parent.rst | grep Subpackages ; echo $?
```

### Environment info
- OS: macOS & Ubuntu 16.04
- Python version: 3.6.4
- Sphinx version: 1.6.6

codingjoe:
Subject: Fix skipping none empty packages

### Feature or Bugfix
- Bugfix

### Purpose
Packages that contain nothing but sub-packages
should not be skipped since they are in fact not empty.

### Detail

**Reproduce**:
```bash
pip install django
django-admin startproject example .
mkdir -p example/management/commands
touch example/management/__init__.py
touch example/management/commands/__init__.py
echo '"""Docstring"""' > example/management/commands/my_customer_command.py
```

### Relates
- #4520

### DIFF
--- a/sphinx/ext/apidoc.py
+++ b/sphinx/ext/apidoc.py
@@ -195,7 +195,7 @@ def shall_skip(module, opts, excludes=[]):
         return True
 
     # skip it if there is nothing (or just \n or \r\n) in the file
-    if path.exists(module) and path.getsize(module) <= 2:
+    if path.exists(module):
         if os.path.basename(module) == '__init__.py':
             # We only want to skip packages if they do not contain any
             # .py files other than __init__.py.


codingjoe:
Closed in favor of #4670

Issue #4655, opened 2018-02-20
Incomplete localization strings in Polish and Chinese


tockrock:
Subject: Incomplete localization strings in Polish and Chinese

I can't suggest the improved strings, but would like to mention it here.

- [x] Missing `%r`
https://github.com/sphinx-doc/sphinx/blob/9a82e1098b1827f25773b1fb11dda64d6db136a8/sphinx/locale/pl/LC_MESSAGES/sphinx.po#L1293-L1296

- [x] `%%(cel)e` should be `%%(target)s`?
https://github.com/sphinx-doc/sphinx/blob/9a82e1098b1827f25773b1fb11dda64d6db136a8/sphinx/locale/pl/LC_MESSAGES/sphinx.po#L1298-L1301

- [x] Missing `%r`
https://github.com/sphinx-doc/sphinx/blob/9a82e1098b1827f25773b1fb11dda64d6db136a8/sphinx/locale/zh_CN/LC_MESSAGES/sphinx.po#L1295-L1298
 
### Environment info
- Sphinx version: 1.7

Hope this helps!

tk0miya:
Thank you for reporting. I fixed the second one. But I can't for others...

tk0miya:
I just call a help for sphinx-users list:
https://groups.google.com/forum/#!topic/sphinx-users/yDac7FnK4wc

I hope somebody help us :-)

m-aciek:
In first (Polish) case it should be:
```
msgid "more than one target found for 'any' cross-reference %r: could be %s" 
msgstr "znaleziono więcej niż jeden cel dla cross-referencji „any” %r: może być %s"
```

tockrock:
Sent a pull request with the suggestion from @m-aciek

tk0miya:
@m-aciek @tockrock Thanks!

tk0miya:
I noticed there is one more invalid translation.
https://github.com/sphinx-doc/sphinx/blob/d48e1ef9e5d2fda22b48e25d5f2289b19b919b31/sphinx/locale/zh_CN/LC_MESSAGES/sphinx.po#L265-L268

tk0miya:
Fixed by #4655.
Thank you for reporting!

PR #4663, opened 2018-02-22
Fix incomplete localization strings in Polish (refs: #4655)

tockrock:
refs: #4655


tk0miya:
Thanks!

tk0miya:
Note: cherry-picked into 1.7 branch manually.
And also fixed message on transifex.

PR #4670, opened 2018-02-24
Closes #4520 - apidoc: Subpackage not in toc. 

### LINKED ISSUE #4520: Subpackage not in toc

zufallsgenerator:
Subject: Fix subpackage not in TOC

### Feature or Bugfix
- Bugfix

### Purpose
- Fix error/unwanted behaviour in TOC generation skipping empty subpackages even though they had non-empty subpackages under them. Also The rule of skipping folders with only an empty __init__.py has been removed. The reason for this is that it was never working consistently in the first place and made the code unnecessary hard to reason about. Tests for the TOC generation have been added, as well as tests for the exclude mechanism since they are coupled. One test (test_ext_apidoc.py::test_exclude) has also been modified to reflect the new behaviour.

### Detail
- #4520 

### Relates
- #4520


codingjoe (approved):
Awesome thx! I can confirm that the tests match the reported bug.

sphinx/ext/apidoc.py
```diff
@@ -194,16 +194,15 @@ def shall_skip(module, opts, excludes=[]):
...
-        else:
+    # Are we a package (here defined as __init__.py, not the folder in itself)
+    if os.path.basename(module) == INITPY:
+        # Yes, check if we have any non-excluded modules at all here
+        basemodule = path.dirname(module)
+        for module in glob.glob(path.join(basemodule, '*.py')):
+            if not is_excluded(path.join(basemodule, module), excludes):
+                # There's a non-excluded module here, we won't skip
+                all_skipped = False
+        if all_skipped:
```
tk0miya, line 205: It seems this variable does not initialized. Did you test a case all-skipped?

tk0miya (changes requested):
To apply this change to next stable release (1.7.x), could you rebase this onto 1.7 branch please?

tk0miya:
To merge this into 1.7.2 release, I start to updating this.

sphinx/ext/apidoc.py
zufallsgenerator: Oh, all_skipped should indeed be initialized. However, the case never occurs, unless __init__.py because we're looking for .py files in the same folder as __init__.py - and __init__.py will be there (unless it is matched by the excludes).

Since this commit removes the requirement of skipping modules with an empty __init__.py, I think this entire loop becomes unnecessary - we can simply check if the filename is __init__.py and return false when that's the case. However, maybe shall_skip should be replaced with two different methods - the cases when it's called for __init__.py and some other .py file are very different.


I'll try to take a deeper look at this this weekend.

tk0miya: I got the error with following testcase:
```
@pytest.mark.apidoc(coderoot='test-apidoc-toc/mypackage', excludes=['something/__init__.py'])
def test(apidoc):
    pass
```
```
        # Are we a package (here defined as __init__.py, not the folder in itself)
        if os.path.basename(module) == INITPY:
            # Yes, check if we have any non-excluded modules at all here
            basemodule = path.dirname(module)
            for module in glob.glob(path.join(basemodule, '*.py')):
                if not is_excluded(path.join(basemodule, module), excludes):
                    # There's a non-excluded module here, we won't skip
                    all_skipped = False
>           if all_skipped:
E           UnboundLocalError: local variable 'all_skipped' referenced before assignment
```

tk0miya: @zufallsgenerator I'd like to release 1.7.2 today. So I'd like to fix the UnboundLocalError and merge it (I added the fix for it on #4766).
I think simplifier is a refactoring. So it is okay after the release.
What do you think?

zufallsgenerator:
Well it makes it a no-op most of the time, so as long as the tests pass I
think it'd be fine. I'm on a mobile device now.

tk0miya:
Thank you for comment!

tk0miya (approved):

tk0miya:
Merged manually! (refs: #4766 )

tk0miya:
Note: We still have remaining work for this.

> Since this commit removes the requirement of skipping modules with an empty init.py, I think this entire loop becomes unnecessary - we can simply check if the filename is init.py and return false when that's the case. However, maybe shall_skip should be replaced with two different methods - the cases when it's called for init.py and some other .py file are very different.

tk0miya:
To my understanding, no refactoring is not needed for `shall_skip()`. So I'd like to close this now.
If you feel there is still remaining work, please send a new PR to us.

Thanks,

PR #4671, opened 2018-02-24
Document: showing links & targets of Intersphinx mapping file

### LINKED ISSUE #4244: Document python -msphinx as allowing to dump an intersphinx db
Subject: python -msphinx.ext.intersphinx is a practical way to dump an intersphinx db

### Problem
Sometimes a doc is missing an intersphinx db entry, causing headscratching for other projects that want to link to it.  `python -msphinx.ext.intersphinx url-or-path` is a practical, already existing way to print the contents of an intersphinx db, allowing one to quickly confirm that the entry is missing.  It would benefit from being documented at http://www.sphinx-doc.org/en/stable/ext/intersphinx.html.

n/a

#### Error logs / results
n/a

#### Expected results
n/a

n/a

### Environment info
n/a

TimKam:
Subject: Document: showing links & targets of Intersphinx mapping file

### Feature or Bugfix
- Docs

### Purpose
Closes #4244 

### Relates
#4244

### DIFF
--- a/doc/ext/intersphinx.rst
+++ b/doc/ext/intersphinx.rst
@@ -84,7 +84,7 @@ linking:
    To add links to modules and objects in the Python standard library
    documentation, use::
 
-      intersphinx_mapping = {'python': ('https://docs.python.org/3.4', None)}
+      intersphinx_mapping = {'python': ('https://docs.python.org/3.6', None)}
 
    This will download the corresponding :file:`objects.inv` file from the
    Internet and generate links to the pages under the given URI.  The downloaded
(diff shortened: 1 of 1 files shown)


doc/ext/intersphinx.rst
```diff
@@ -84,7 +87,7 @@ linking:
    To add links to modules and objects in the Python standard library
    documentation, use::
 
-      intersphinx_mapping = {'python': ('https://docs.python.org/3.4', None)}
+      intersphinx_mapping = {'python': ('https://docs.python.org/3.6', None)}
```
tk0miya: How about `https://docs.python.org/3` instead?

tk0miya (commented):
+1!

doc/ext/intersphinx.rst
TimKam: Sure!

tk0miya:
Thanks!

PR #4672, opened 2018-02-24
Fix #4669: sphinx.build_main and sphinx.make_main throw NameError

### LINKED ISSUE #4669: sphinx.build_main and sphinx.make_main deprecated, but throw NameError
Subject: Attempts to directly invoke sphinx.build_main() or sphinx.make_main() result in NameError

### Problem
- I have a wrapper script that directly invokes sphinx.make_main(); with Sphinx 1.7.1 this now results in a NameError exception instead of just the intended deprecation warning.

Attempt to directly call the (now deprecated) sphinx.make_main().

#### Error logs / results
Here's the relevant traceback:
```
  File "./build-deploy.py", line 141, in <module>
    errno = build_html(args.sourcedir, args.outputdir, args.branch, args.nitpicky, args.force)
  File "./build-deploy.py", line 65, in build_html
    errno = _sphinx_build(arglist)
  File "./build-deploy.py", line 44, in _sphinx_build
    return sphinx.make_main(arglist)
  File "/usr/local/lib/python3.6/site-packages/sphinx/__init__.py", line 91, in make_main
    return build.make_main(argv[1:])  # skip first argument to adjust arguments (refs: #4615)
NameError: name 'build' is not defined
```

#### Expected results
I'd expect to get the deprecation warning, but for the invocation to still work, as it would with sphinx.main() where I see:
```
def main(*args, **kwargs):
    from .cmd import build
```

I realize that I'm likely calling into a function that's not part of the Sphinx public API, but I don't see much reason for the warnings.warn() calls to be there in addition to the exception. It might be friendlier if both sphinx.build_main() and sphinx.make_main() also did the import of build module so they continued to work as well as warn.

### Environment info
- OS: MacOS 10.12.6
- Python version: Python 3.6.4
- Sphinx version: 1.7.1

tk0miya:
refs: #4669

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,8 @@ Features added
 Bugs fixed
 ----------
 
+* #4669: sphinx.build_main and sphinx.make_main throw NameError
+
 Testing
 --------
 
--- a/sphinx/__init__.py
+++ b/sphinx/__init__.py
@@ -73,6 +73,7 @@ def main(*args, **kwargs):
 
 def build_main(argv=sys.argv):
     """Sphinx build "main" command-line entry."""
+    from .cmd import build
     warnings.warn(
         '`sphinx.build_main()` has moved to `sphinx.cmd.build.build_main()`.',
         RemovedInSphinx20Warning,
@@ -83,6 +84,7 @@ def build_main(argv=sys.argv):
 
 def make_main(argv=sys.argv):
     """Sphinx build "make mode" entry."""
+    from .cmd import build
     warnings.warn(
         '`sphinx.build_main()` has moved to `sphinx.cmd.build.make_main()`.',
         RemovedInSphinx20Warning,


stephenfin:
Damn it. This should really have regression test to make sure it passes. Could you add these, mocking out the implementation specifics?

tk0miya:
I tested this manually. So I don't think the test is needed.

stephenfin:
> I tested this manually. So I don't think the test is needed.

Right, but it's broken once. A test would ensure this wouldn't break again. Of course you don't have to but it does make things easier.

tk0miya:
This is already marked as deprecated. So next change is removal.
If you feel the testing is important, please send another PR. Either is fine to me. But I would not work for it.

stephenfin:
OK, that's understandable.

tk0miya:
Merged. Thank you for comment.

PR #4675, opened 2018-02-24
doc: Add deprecation list

tk0miya:
So far, deprecated APIs are only listed as CHANGES entry.  But
they are not listed in one place.  In this change, I gathered them
to docs as a list.
As a side effect, it makes CHANGES entry simple.


tk0miya:
Hmm... the list is protruded a bit. Is there good idea?

tk0miya:
I just arranged the text to fold labels.

PR #4684, opened 2018-02-27
Remove incorrectly formatted translation

tuchfarber:
Subject: Removing a broken translation

### Feature or Bugfix

- Feature
- Bugfix

### Purpose
- Running compiletranslations on this .po file causes an error of `number of format specifications in 'msgid' and 'msgstr' does not match` due to the missing `%r` reference in the translation. This removes the translation because I don't know the language to fix it. A fixed translation would be preferred but this is outside of my skillset.

This translation was added in bc068e41435fea05b419133f23b33e7c19f4f39c. Prior to that, it was an empty string, so this reverts to the empty string.


tuchfarber:
Created issue for this PR #4692

tuchfarber:
PyPy test is failing for unrelated reason.

tk0miya:
Merged by manually (refs: #4750 )
Thank you for contribution!

PR #4674, opened 2018-02-24
Improve i18n

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Enable to i18nize console messages (by locale envvars like `LC_*`)
- Add new API to translate messages on extensions
- Integrate `l_()` into `_()`
- Refactor `sphinx.locale` module to do that.

PR #4704, opened 2018-03-02
Use "%r" to stringify exceptions

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
On py2, an exception having i18nized message causes UnicodeError on
stringify.  This uses format string "%r" to stringify them.

This is related with #4674.

### DIFF
--- a/sphinx/directives/code.py
+++ b/sphinx/directives/code.py
@@ -15,6 +15,7 @@ from typing import TYPE_CHECKING
 from docutils import nodes
 from docutils.parsers.rst import Directive, directives
 from docutils.statemachine import ViewList
+from six import text_type
 
 from sphinx import addnodes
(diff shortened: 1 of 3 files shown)


sphinx/directives/code.py
```diff
@@ -159,7 +160,7 @@ def run(self):
             try:
                 literal = container_wrapper(self, literal, caption)
             except ValueError as exc:
-                return [document.reporter.warning(str(exc), line=self.lineno)]
+                return [document.reporter.warning(text_type(exc), line=self.lineno)]
```
tk0miya, line 163: Finally I used `text_type()` to stringify. `"%r"` converts exceptions to `ValueError('blah blah...')`. It's not good for message.

PR #4706, opened 2018-03-02
WIP: Don't sort module and attribute members

ostrokach:
Subject: Maintain definition order when using autosummary and autodoc in Python 3.6.

### Feature or Bugfix

- Feature

### Purpose

In Python 3.6, class definition order is preserved, due to [PEP 520](https://www.python.org/dev/peps/pep-0520/). One of the top reasons for PEP 520 was to enable documentation generators to maintain definition order in the generated documentation. A good example is NamedTuples, where you want do document parameters in the order in which they should be input (and not in alphabetical order, as is currently the case).

### Detail

- This is just a proof of concept. Not sure if this behaviour should be made configurable.

### Relates

- https://www.python.org/dev/peps/pep-0520/
- https://stackoverflow.com/a/4692947/2063031

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -562,12 +562,12 @@ class Documenter(object):
                 else:
                     logger.warning('missing attribute %s in object %s' %
                                    (name, self.fullname))
-            return False, sorted(selected)
+            return False, selected
         elif self.options.inherited_members:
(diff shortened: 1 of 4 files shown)


tk0miya:
Please let me know the status of this PR. Is this still in WIP?

tk0miya:
Now I'm closing this now. And we'll reconsider this again in #5270 and #5271.
Thank you for proposal.

PR #4715, opened 2018-03-07
Fix filename of epub (refs: #4611)

tk0miya:
Fix wrong filename in our testcase (refs: #4611)


stephenfin (approved):
Looks good to me

tk0miya:
Thanks,

PR #4721, opened 2018-03-07
util.requests: Support SPHINX_NO_NETWORK kill-switch

mgorny:
Subject: Make it possible to disable all network access while building documentation

### Feature or Bugfix
- Feature

### Purpose
Support a global kill-switch to disable network use in all Sphinx
processes. When SPHINX_NO_NETWORK is defined to a non-null value
in the environment, all attempts to fetch remote data via Sphinx
util.requests module will fail via exception.
    
The main use case is building documentation in environments where
network access is undesired. A lot of modern packages use Intersphinx
and disabling it per-package became really unfeasible. I have originally
considered patching Sphinx locally for Gentoo but I've reached
the conclusion that there might be other users who would benefit
from an option to disable network access.

### Detail
A detailed reasoning for disabling network access is explained
in the Gentoo developer documentation. The particular article pertains
to use of network in tests but it most of the arguments apply also
to building documentation and other build-related activities:
https://devmanual.gentoo.org/ebuild-writing/functions/src_test/index.html#tests-that-require-network-or-service-access

### Relates
- https://bugs.gentoo.org/634790

### DIFF
--- a/sphinx/util/requests.py
+++ b/sphinx/util/requests.py
@@ -11,6 +11,7 @@
 
 from __future__ import absolute_import
 
+import os
 import warnings
 from contextlib import contextmanager
 from typing import TYPE_CHECKING
@@ -135,6 +136,9 @@ def get(url, **kwargs):
     """Sends a GET request like requests.get().
 
     This sets up User-Agent header and TLS verification automatically."""
+    if os.environ.get('SPHINX_NO_NETWORK', ''):
+        raise RuntimeError('Network use disabled via SPHINX_NO_NETWORK')
+
     kwargs.setdefault('headers', dict(useragent_header))
     config = kwargs.pop('config', None)
     if config:
@@ -149,6 +153,9 @@ def head(url, **kwargs):
     """Sends a HEAD request like requests.head().
 
     This sets up User-Agent header and TLS verification automatically."""
+    if os.environ.get('SPHINX_NO_NETWORK', ''):
+        raise RuntimeError('Network use disabled via SPHINX_NO_NETWORK')
+
     kwargs.setdefault('headers', dict(useragent_header))
     config = kwargs.pop('config', None)
     if config:


tk0miya:
Thank you for posting.
Please let me know who is the target of this feature? I feel this is useful for Gentoo developers only.
If so, I would not like to merge this into sphinx-core.

mgorny:
I should point out that Gentoo is a source-based distribution, so it does not affect *just* Gentoo developers but all of Gentoo *users*. Currently, the Sphinx-based packages frequently attempt to contact Internet without their permission and (as explained in the linked document) this may have serious security implications and/or cause additional Internet fees. Not to mention that in some cases it causes the packages to fail randomly for some of our users.

Sure, we can patch it locally. However, I am aware that there are security conscious people who are not Gentoo users (or who need to use other distributions), so I preferred submitting the patch upstream so that others could benefit from it as well.

tk0miya:
Thank you for clarification. I understood.

-0 for your proposal. I don't think the users want such option. But I can accept it if many people votes to this.

Arfrever:
I vote +1 for this feature.

chutz:
I vote for this feature as well. Distributions other than Gentoo can make use of this, most distros have rules about reproducible builds that forbid packages from downloading things while they are building.

tk0miya:
@chutz thank you for comment. Please let me know some example of distros?

tk0miya:
I have to say this does not prevent all of network accesses. This only stops communications from sphinx-core. It can't stop the communications from 3rd party extensions, user code (`conf.py`), sub processes and so on.

For example, third party extensions and user configuration file (`conf.py`) can use network if they are not using `sphinx.util.requests` module. The module is introduced only for sphinx-core. So extensions and users might not use this.

In another case, `sphinx.ext.graphviz` invokes `dot` command internally as a subprocess. If the graph contains remote images. `dot` command will fetch them to build a graph. 


If you really want to drop all of network access, you have to monkey-patch to `socket` module of python.

tk0miya:
Hi @mitya57 

I'd like to hear your opinion if you have time.
What do you think about such an option? Is it important for Debian/Ubuntu packaging?

mitya57:
Yes, I think this feature will be useful for us (Debian/Ubuntu) as we build packages without network access.

mitya57:
Though thinking more about it, there are already environment variables that helps to achieve the same thing.

```bash
export http_proxy=http://127.0.0.1:9/
export https_proxy=$http_proxy
```

Setting these will also lead to an exception raised, and it should be more universal.

So maybe there is little point in adding another environment variable.

sbraz:
@mitya57 It's a bit hacky, though, isn't it?
I'm in favour of merging this feature, it seems like a proper solution to a problem that several distributions probably encounter.

mitya57:
I just said that we already have some solution in Debian/Ubuntu. If there is another switch then it won’t hurt us, we can set both.

Everything else is up to Sphinx’ maintainers to decide.

AA-Turner:
Would disabling `socket.socket` be a better approach here? It would also disable network access (that used Python APIs) for extension modules, etc.

A

AA-Turner:
Closing this, Docutils itself accesses the network, and anything we add to Sphinx core we can't guarantee will be respected by the extensions. Blacking out network access at the OS level seems the only true way of achieving the end goal here (even socket.socket can be worked around by using an external library). If new facts arise, please open an issue to discus further.

A

Issue #4574, opened 2018-02-07
vertical space before equation in latex


aszekMosek:
It seems to me that equations without a label are written to the latex source as

```
handles problems of the form:
\begin{equation*}
```
and equations with a label as:

```
form of a linear minimization problem is:

\phantomsection\label{\detokenize{linear:equation-eq-linear-standard1}}\begin{equation}\label{equation:linear:eq-linear-standard1}
```

The extra empty line before ``phantomsection`` starts a paragraph and causes an unnecessarily large vertical space before ``equation``, unlike for ``equation*`` where it is OK. 

Removing the empty line fixes the problem.

Also, I'm not sure ``phantomsection`` is necessary for equations. ``hyperref`` seems to produce links to equations without it just as well.

Sphinx 1.6.7

jfbu:
Fixed at e1e46cdf8. Thanks for report!

Issue #4333, opened 2017-12-21
Make directives and roles for math to sphinx-core


tk0miya:
### Problem
- Since docutils-0.8, `math` directive and `math` role became a part of reST spec.
  - http://docutils.sourceforge.net/docs/ref/rst/directives.html#math
  - http://docutils.sourceforge.net/docs/ref/rst/roles.html#math
- In sphinx, users have to enable one of math extension to use them.

#### Expected results
Enable to use the directive and role by default

jfbu:
I think policy of Docutils to insert contents verbatim if output format does not support natively math input is fine. And that it is strange for latex output to have to activate either mathjax or imgmath extensions which are for html like targets. Thus +1 for this feature.

PR #4729, opened 2018-03-11
Fix vertical space before equation in latex (closes: #4574)

jfbu:
Relates #4574

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1959,6 +1959,10 @@ class LaTeXTranslator(nodes.NodeVisitor):
             if id.startswith('index-'):
                 return
 
+            # equations also need no extra blank line nor hypertarget
+            if id.startswith('equation-'):
+                return
+
             # insert blank line, if the target follows a paragraph node
             index = node.parent.index(node)
             if index > 0 and isinstance(node.parent[index - 1], nodes.paragraph):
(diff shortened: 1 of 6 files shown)


tk0miya (approved):
LGTM

sphinx/writers/latex.py
```diff
@@ -1959,6 +1959,10 @@ def add_target(id):
             if id.startswith('index-'):
                 return
 
+            # equations also need no extra blank line nor hypertarget
+            if id.startswith('equation-'):
```
tk0miya: math elements are provided as extension. According to basis, LaTeX writer should not know about `equation-` label.
Could you add a TODO here please?

jfbu:
Thanks for review @tk0miya. I had originally

    math_eqref_format = 'Equation number {number}'

in my test file to double-check that the change did not introduce conflict with that feature. But at last minute I did not include that in the test file of the PR.

(I was surprised that the change needed no update to test for `make test` to pass it seems math equations are not much tested)

sphinx/writers/latex.py
jfbu: OK, I will. Aren't you planning to merge mathbase into Sphinx-core ?

tk0miya: Yes! see #4333

tk0miya:
LGTM!

Issue #3357, opened 2017-01-19
Implement parts (of book)


mixmastamyk:
Hi, I have an website/ebook I'm working on with parts, e.g.:

**Part I: Basics**

1. Ch. One
2. Ch. Two

**Part II: Intermediate**

3. Ch. Three
4. Ch. Four

**Part III: Advanced Stuff**

5. Ch. Five

etc.  But can't seem to figure out how to do it.  The toctree caption helps with the title of the first part.  If I use several toctrees however they restart numbering from 1.  How can I use one toctree to do this, or perhaps several that can pick up numbering where the last left off?  Thank you,

mixmastamyk:
BTW, I need this and would consider giving a $100 tip (perhaps by Amazon gift card) to get it done.  Sorry I can't afford a larger amount.

It might be as simple as implementing an offset parameter for the toctree.

tk0miya:
As a quick hack, I made a simple extension to resolve this:
https://gist.github.com/25bd52887adea5314fe15f1e3afc0949
(it works only with Sphinx-1.5)

This extension assigns numbers to each sections with following rule:

* treat top level sections as "parts"
* treat second level sections as "sections"
* Do not assign numbers to "parts"
* Assign sequential number to "sections" whole of document

At my local, it seems working.


BTW, we should consider carefully how to support ToC like this case.
I feel my extension is too adhoc to merge sphinx-core.

mixmastamyk:
Awesome, does this mean my chapters would become parts?  That might be hard to do because my book is getting long and would be hard to rearrange.

I wonder if it would be easier to add an `:offset:` parameter to the toctree, so that it would start numbering at a given number or continue where it left off.  So then I could add it to the second toctree to get it to start at the right place.

tk0miya:
My extension considers first level sections as parts and second level ones as chapters.
I believe this document structure is useful for LaTeX output.
With `latex_toplevel_sectioning = 'part'`, LaTeX builder also considers parts and sections same as my extension.

At this moment, how do you make chapters in your reST? I'd like to see any example.

BTW, your idea, `:offset:` option is hard to realize in LaTeX output to my knowledge. In addition, we need to consider side effects. for example, duplicated section numbers, skipped numbers and so on.
So I'd like to consider carefully to improve ToC feature in sphinx-core.

mixmastamyk:
[...]

PR #4733, opened 2018-03-13
Implement sequential numbering for multiple toctrees

thfanning:
Subject: Implement sequential numbering for multiple toctrees in the same document.

### Feature or Bugfix
- Feature (see #3357)

### Purpose
Multiple, *numbered* toctrees in the same document normally start with 1. This simple update allows authors to sequentially number multiple toctrees in the same document, or to specify a specific starting number.

### Detail
Multiple toctrees of the form

      Part I
      ======

      .. toctree::
         :numbered:

         foo
         bar

      Part II
      =======

      .. toctree::
         :numbered:
         :continue:

         baz

will result with the heading of `baz` being numbered as 3. Starting a toctree at a specific number is also possible by giving a positive value as a numeric argument to ``continue``.

A test case is provided, and a unit test is included to verify behavior. The Sphinx documentation has also been updated.

I have only been able to test this with the html builder, but am interested in feedback.

### Relates
- #3357

### DIFF
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -21,6 +21,7 @@ if False:
     from typing import Any, Callable, Dict, Iterable, List, Tuple, Type, Union  # NOQA
     from docutils import nodes  # NOQA
     from docutils.parsers.rst.states import Inliner  # NOQA
+    from sphinx import addnodes  # NOQA
     from sphinx.builders import Builder  # NOQA
     from sphinx.environment import BuildEnvironment  # NOQA
     from sphinx.roles import XRefRole  # NOQA
@@ -265,7 +266,7 @@ class Domain:
                                   self.__class__)
 
     def process_doc(self, env, docname, document):
-        # type: (BuildEnvironment, unicode, nodes.Node) -> None
+        # type: (BuildEnvironment, unicode, nodes.document) -> None
         """Process a document after it is read by the environment."""
         pass
 
@@ -283,7 +284,7 @@ class Domain:
 
     def resolve_xref(self, env, fromdocname, builder,
                      typ, target, node, contnode):
-        # type: (BuildEnvironment, unicode, Builder, unicode, unicode, nodes.Node, nodes.Node) -> nodes.Node  # NOQA
+        # type: (BuildEnvironment, unicode, Builder, unicode, unicode, addnodes.pending_xref, nodes.Element) -> nodes.Element  # NOQA
         """Resolve the pending_xref *node* with the given *typ* and *target*.
 
         This method should return a new node, to replace the xref node,
@@ -300,7 +301,7 @@ class Domain:
         pass
 
     def resolve_any_xref(self, env, fromdocname, builder, target, node, contnode):
-        # type: (BuildEnvironment, unicode, Builder, unicode, nodes.Node, nodes.Node) -> List[Tuple[unicode, nodes.Node]]  # NOQA
+        # type: (BuildEnvironment, unicode, Builder, unicode, addnodes.pending_xref, nodes.Element) -> List[Tuple[unicode, nodes.Element]]  # NOQA
         """Resolve the pending_xref *node* with the given *target*.
 
         The reference comes from an "any" or similar role, which means that we
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -295,7 +295,7 @@ class CDomain(Domain):
 
     def resolve_xref(self, env, fromdocname, builder,
                      typ, target, node, contnode):
-        # type: (BuildEnvironment, unicode, Builder, unicode, unicode, nodes.Node, nodes.Node) -> nodes.Node  # NOQA
+        # type: (BuildEnvironment, unicode, Builder, unicode, unicode, addnodes.pending_xref, nodes.Element) -> nodes.Element  # NOQA
         # strip pointer asterisk
         target = target.rstrip(' *')
         # becase TypedField can generate xrefs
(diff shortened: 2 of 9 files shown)


jfbu:
Regarding the `latex` builder, I tested briefly (only by way of experiment, I did not look closely at PR) that this does not impact it. One can already obtain the effect by using `latex_toplevel_sectioning = 'part'`  in `conf.py`, because the default behaviour in LaTeX `book` class is to number "chapters" continuously, even if enclosed in parts.

By the way for better parallel  with html builds, perhaps LaTeX should not number the parts, i.e. use `\part*`, not `\part` (but then `\addcontentsline{toc}{part}{Name of part}` have to be added to latex file).

mixmastamyk:
One minor question, does it interact properly with toctree captions?  I assume it works, however the added docs use headings, while I've used captions in the past for parts of a book.  Thanks again.

thfanning:
@mixmastamyk - yes, captions work properly. The `:continue:` option simply tracks and/or alters the starting number for numbered toctrees (within the same document). No other manipulation is done.

mixmastamyk:
Hi, would someone mind taking a look at this pull request? @tk0miya @jfbu

Would be **really** helpful to be able to write a book with parts in Sphinx. :-D

Thank you,

thfanning:
I've updated the pull request to resolve conflicts with upstream. Hopefully this will facilitate review and acceptance.

mixmastamyk:
@tk0miya @jfbu any comments?

mixmastamyk:
@tk0miya @jfbu any comments?

thfanning:
@tk0miya @jfbu 
I've worked hard to keep this up-to-date with the upstream. This update has only a few new lines of code (mainly in toctree.py). The rest is documentation and tests. Please take a moment to review. Thanks!

tk0miya:
Sorry for very late response.

We have to design whole of ToC. I can't say what way is better to define better ToC (inc. appendix, indices and so on).

At least, ToC settings and mark-ups should effect to all of HTML, EPUB and PDF (LaTeX) documents. So we need to update this before merging if we choose this way.

mixmastamyk:
As the extension merely changes the chapter numbers, I'd be surprised it didn't work with other output formats.  Is that the case?  Has someone tested it and found that not to be working?

thfanning:
@tk0miya I agree with the need to consider the ToC in an integrated sense. In this case, however, it is entirely up to the user if they want to start numbering at a point other than 1. This PR helps them do that.

Thanks!

tk0miya:
@mixmastamyk see https://github.com/tk0miya/sphinx_issue_4733

HTML version:


PDF version:

mixmastamyk:
Thanks, I guess the question is….

- Is the bug in the toctree, where this PR should fix it?  
- Or is there a lack of support in the non-html output builders?

If the second, support could be tracked and worked on in separate issues and timeframes.

tk0miya:
>Is the bug in the toctree, where this PR should fix it?

No, this is not a bug. I think this PR introduces a new feature; "part like" sectioning. But this PR is only available for HTML output. So I think this PR should also add the feature to LaTeX builder.


Note: This is not main reason why I'm not merging this now. As I commented above, we have to re-design ToC model of Sphinx. I can't say `:continue:` option is best way to do that. So we have to discuss what is best form of ToC before implementation.

mixmastamyk:
It does feel like a design flaw that some builders don't pick up chapter numbers, it means their mechanisms must differ somehow, perhaps numbering needs to be moved up into a parent class.

From the user interface/markup perspective, perhaps a `:part: NAME_STRING` parameter that combines the current functionality of `:caption:` and `:continue:`.  (You mentioned removing the sidebar additions from `:caption:` in the other bug, but I'm currently relying on that.)   Perhaps `:part:` would capture the intent better.

If you need someone to give feedback, I'm available.  Where could we go to confer?

thfanning:
> If the second, support could be tracked and worked on in separate issues and timeframes

My understanding is that it is due to a lack of support in non-html builders. Up until now, it's only been by coincidence that they match. All this PR does is modify the `section` tuple during parsing. Clearly the LaTeX builder doesn't even use it. It's even numbering the un-numbered headings outside of the top-level `:numbered:` toctrees. Remove the `:continue:` option from the test case and the LaTeX builder will still get it wrong.

Even though I numbered the "Parts" in the test case (as "Part I" and "Part II"), don't assume those necessarily carry any numeric meaning. For example, the part headings could be "Prologue", "Statics", "Dynamics", "Glossary", "Index", and "Epilogue". Maybe only "Statics" and "Dynamics" should be numbered sequentially (with `:continue:`) and the others aren't numbered at all.

mixmastamyk:
Ok, thanks.  Wouldn't a `:numbered:` parameter determine if they are numbered or not?  Perhaps `section` is a more generic name.

jessetan:
> I can't say :continue: option is best way to do that. So we have to discuss what is best form of ToC before implementation.

It would be nice if this PR was used as input for a larger discussion of how toctree is used, see e.g. #4133 about more generic numbering (and lettering) of sections

thfanning:
> It would be nice if this PR was used as input for a larger discussion of how toctree is used, see e.g. #4133 about more generic numbering (and lettering) of sections

@jessetan, I agree. However I would like to see this PR merged first. Then, we could take the numeric secnumber list (was tuple) that is generated and transform it by a custom function into any complex numbering format that is desired (numeric, alphabetic, roman, etc.). Right now, that "function" is simply `'.'.join(map(str, node['secnumber']))` in the html writer.

choldgraf:
It's unclear to me if this is implemented or not - is it? I see that @thfanning closed the issue in a commit but I'm not sure how we can trigger this behavior if we wish

mixmastamyk:
No, it continues to be ignored.

thfanning:
I did not intend to close this pull request. We needed to update our internal implementation, but my master branch was so far out of sync with upstream that it was a mess. So I had to force my own master to update. My apologies for not realizing the impact. Honestly, I'm a wiz at Subversion but suck with git.

For my current implementation, see thfanning@c2ed5bb315cfdb88b282ea75239fc25d03f1d82c. Unfortunately the current branch includes a tweak to equation numbering as well, but the TOC issue is trivial to implement.

choldgraf:
Maybe @tk0miya could re-open this? I think the issue is an important one because (to my knowledge) this makes sub-sections not have unique numbers (because you'll have a section in two different chapters, each called `1.1`, instead of `1.1` and `2.1`)

tk0miya:
Unfortunately, I also can't reopen.

PR #4740, opened 2018-03-14
Fix #1435: qthelp builder should htmlescape keywords

### LINKED ISSUE #1435: qthelp builder should htmlescape keywords
Building qch documentation fails without this patch, as the keywords contain `'<CONFIG>'` etc:

 http://www.cmake.org/cmake/help/v3.0/variable/CMAKE_EXE_LINKER_FLAGS_CONFIG.html

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1435
- Originally reported by: Anonymous
- Originally created at: 2014-03-31T15:02:05.418

tk0miya:
refs: #1435.

This does not contain last change of the patch (attached at #1435).
I think it is too escaped because the `ref` argument will be escaped inside `keyword_item()` method.

### DIFF
--- a/sphinx/builders/qthelp.py
+++ b/sphinx/builders/qthelp.py
@@ -300,9 +300,9 @@ class QtHelpBuilder(StandaloneHTMLBuilder):
 
         if id:
             item = ' ' * 12 + '<keyword name="%s" id="%s" ref="%s"/>' % (
-                name, id, ref[1])
+                name, id, htmlescape(ref[1]))
         else:
-            item = ' ' * 12 + '<keyword name="%s" ref="%s"/>' % (name, ref[1])
+            item = ' ' * 12 + '<keyword name="%s" ref="%s"/>' % (name, htmlescape(ref[1]))
         item.encode('ascii', 'xmlcharrefreplace')
         return item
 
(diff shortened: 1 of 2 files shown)


sphinx/builders/qthelp.py
```diff
@@ -300,9 +300,9 @@ def keyword_item(self, name, ref):
 
         if id:
             item = ' ' * 12 + '<keyword name="%s" id="%s" ref="%s"/>' % (
-                name, id, ref[1])
+                name, id, htmlescape(ref[1]))
         else:
-            item = ' ' * 12 + '<keyword name="%s" ref="%s"/>' % (name, ref[1])
+            item = ' ' * 12 + '<keyword name="%s" ref="%s"/>' % (name, htmlescape(ref[1]))
```
tk0miya, line 305: We have to move this to templates in future.

sphinx/builders/qthelp.py
```diff
@@ -300,9 +300,9 @@ def keyword_item(self, name, ref):
 
         if id:
             item = ' ' * 12 + '<keyword name="%s" id="%s" ref="%s"/>' % (
-                name, id, ref[1])
+                name, id, htmlescape(ref[1]))
```
tk0miya, line 303: I noticed `name` is not escaped. double-quotes can break XML tag.
I withdraw this PR.

PR #4742, opened 2018-03-14
Refactor enumerable nodes

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- As a first step of #4333, I'd like to move MathDomain to `sphinx/domains`.
- Before moving, I tried to refactor around managing enumerable nodes.
- So far, enumerable nodes are registered to StandardDomain. So our implementation is strongly coupled with it.
- In this PR, I moved the responsibility of managing enum-nodes to all Domain classes.


sphinx/domains/std.py
```diff
@@ -944,6 +946,17 @@ def has_child(node, cls):
...
             return figtype
 
+    def get_figtype(self, node):
+        # type: (nodes.Node) -> unicode
+        """Get figure type of nodes.
+
+        .. deprecated:: 1.8
+        """
+        warnings.warn('StandardDomain.get_figtype() is deprecated. '
+                      'Please use get_figtype() instead.',
```
tk0miya: This message is wrong.

sphinx/domains/__init__.py
```diff
@@ -333,6 +335,12 @@ def get_type_name(self, type, primary=False):
             return type.lname
         return _('%s %s') % (self.label, type.lname)
 
+    def get_enumerated_node_type(self, node):
```
tk0miya: This should be named `enumerable_node`.

sphinx/domains/__init__.py
```diff
@@ -150,6 +150,8 @@ class Domain(object):
     indices = []            # type: List[Type[Index]]
     #: role name -> a warning message if reference is missing
     dangling_warnings = {}  # type: Dict[unicode, unicode]
+    #: node_class -> (enum_node_type, title_getter)
+    enumerable_nodes = {}   # type: Dict[nodes.Node, Tuple[unicode, Callable]]
```
tk0miya, line 154: I'd like to discuss who should manage enumerable nodes. In this PR, I moved it to Domain class.
But I'm not sure it is responsibility of Domain or not, and I can't say this is best.
If you have any idea, please let me know.

shimizukawa: I have no objection to this implementation. It is reasonable to handle the enumeration of nodes belonging to a math domain within that domain.

One concern is, implementations that use this mechanism may not appear because the enumeration node does not come up with the existing "figure", "table", "code" and "math".

shimizukawa (approved):
LGTM

tk0miya:
Thank you for reviewing!

PR #4760, opened 2018-03-20
Add a new keyword argument ``override`` to Application APIs

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add a new keyword argument `override` to Application APIs


sphinx/application.py
```diff
@@ -770,48 +784,57 @@ def override_domain(self, domain):
         of the existing one.
 
         .. versionadded:: 1.0
+        .. deprecated:: 1.8
+           Integrated to :meth:`add_domain`.
```
tk0miya, line 788: I noticed this deprecations is not mentioned on CHANGES. I'll fix it later.

PR #4764, opened 2018-03-20
Fix #4744: Add canonical URL to docs

### LINKED ISSUE #4744: stable/master 1.7.1/1.6.8+
According to https://github.com/sphinx-doc/sphinx/blob/master/CONTRIBUTING.rst#id32 there are two branches: master and major.minor (=stable).

The git repository has indeed only two branches.

The online documentation is available at:
* http://www.sphinx-doc.org/en/master/ - showing in the title 1.8.0+, and
* http://www.sphinx-doc.org/en/stable/ - showing in the title 1.6.8+

In particular, 

www.sphinx-doc.org/en/stable/ext/autodoc.html doesn't mention `autodoc_inherit_docstrings`, which is part of 1.7, while http://www.sphinx-doc.org/en/master/ext/autodoc.html mentions it.

If 1.7.1 is the stable version, please replace all occurrences at /en/stable with the documentation for 1.7, otherwise clarify, that 1.6.8 is the stable version.

tk0miya:
### Feature or Bugfix
- Docs

### Purpose
- To avoid to index old documents, this sets canonical URL to each pages.
- I'd like to copy this into other old documents

### Relates
- #4744

### DIFF
--- a/doc/_themes/sphinx13/layout.html
+++ b/doc/_themes/sphinx13/layout.html
@@ -13,6 +13,11 @@
 {% block sidebar1 %}{{ sidebar() }}{% endblock %}
 {% block sidebar2 %}{% endblock %}
 
+{% block linktags %}
+{{ super() }}
+    <link rel="canonical" href="http://www.sphinx-doc.org/en/master/{{ pagename }}{{ file_suffix }}" />
+{% endblock %}
+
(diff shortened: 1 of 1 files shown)


shimizukawa (approved):
LGTM

PR #3653, opened 2017-04-22
Add a new extension: sphinx.ext.imgconverter

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
This extension converts images in your document to appropriate format for builders.
For example, it allows you to use SVG images with LaTeX builder.
As a result, you don't mind what image format the builder supports.

Internally, this extension uses Imagemagick to convert images.

Imagemagick: https://www.imagemagick.org/script/index.php

### Relates
- #2859

PR #2859, opened 2016-08-16
Automatically convert SVG to PDF images in the latex builder

alyjak:
This adds the ability to automatically convert svg images to pdfs using Inkscape when using the latex builder. This gets around the fact that latexpdf cannot natively handle svg images.

This change should be non invasive -- no new dependencies are required. Only if Inkscape is already installed will the code spawn a subprocess to call it to convert svgs into pdfs. If Inkscape is not available, the existing code should be unaffected besides an extra console print that lets users know that if they install Inkscape they will get this new feature.

Notes:
- I haven't tested this with windows.
- Some users might find the extra print out to be spammy. Let me know what you think and I'd be happy to remove it.
- I've run all the unit tests with and without Inkscape installed -- they all seem to pass in python 3.5, but when run in python 2.7 a couple seem to fail but they look to be unrelated to this change
- The unit tests have different behavior based off of whether Inkscape is installed. If its not, test_environment.py `test_images()` works the same as it did. If it is installed, `test_images()` expects to see `svgimg.svg: svgimg_svg2pdf.pdf` in its image map.
- I didn't see any place in the docs that needed to be updated for this change. Let me know if you think there's any documentation that needs updating.
- I didn't think this was a big enough change to update `CHANGES`, I also wasn't sure which section should be updated anyways.

PR #4772, opened 2018-03-23
Add LaTeX SVG to PDF converter extension

sephalon:
Subject: Add LaTeX SVG to PDF converter extension

### Feature or Bugfix
- Feature

### Purpose
The imgconverter extension converts vector graphics like SVG and PDF to PNG, which is a very portable approach. However, for quality and reusability reasons we prefer to keep all images in SVG format and only perform a conversion if necessary; right now, the LaTeX builder seems to be the only one.

### Open Issues
- I don't like the extension name yet, any better ideas?
- Do we have to consider any other builders that need conversion?
- If yes, we could add configuration options to fine-tune conversion rules.
- Support rsvg-convert (lower footprint) in addition to Inkscape and allow the user to choose
- No documentation yet.

### Relates
- #3653
- #2859

### DIFF
--- a/sphinx/ext/latex_svg2pdf_converter.py
+++ b/sphinx/ext/latex_svg2pdf_converter.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+"""
+    sphinx.ext.latex_svg2pdf_converter
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    SVG to PDF converter extension for LaTeX builder
+
+    :copyright: Copyright 2007-2018 by the Sphinx team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+import subprocess
+
+from sphinx.errors import ExtensionError
+from sphinx.locale import __
+from sphinx.transforms.post_transforms.images import ImageConverter
+from sphinx.util import logging
+from sphinx.util.osutil import ENOENT, EPIPE, EINVAL
+
+if False:
+    # For type annotation
+    from typing import Any, Dict  # NOQA
+    from sphinx.application import Sphinx  # NOQA
+
+
+logger = logging.getLogger(__name__)
+
+
+class InkscapeConverter(ImageConverter):
+    conversion_rules = [
+        ('image/svg+xml', 'application/pdf'),
+    ]
+
+    def is_available(self):
+        # type: () -> bool
+        """Confirms the converter is available or not."""
+        try:
+            args = [self.config.latex_svg2pdf_converter, '--version']
+            logger.debug('Invoking %r ...', args)
+            ret = subprocess.call(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+            if ret == 0:
+                return True
+            else:
+                return False
+        except (OSError, IOError):
+            logger.warning(__('convert command %r cannot be run.'
+                              'check the image_converter setting'),
+                           self.latex_svg2pdf_converter)
+            return False
+
+    def convert(self, _from, _to):
+        # type: (unicode, unicode) -> bool
+        """Converts the image to expected one."""
+        if self.app.builder.name != 'latex':
+            return
+
+        try:
+            args = ([self.config.latex_svg2pdf_converter] +
+                    self.config.latex_svg2pdf_converter_args +
+                    ['--export-pdf=' + _to, _from])
+            logger.debug('Invoking %r ...', args)
+            p = subprocess.Popen(args, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        except OSError as err:
+            if err.errno != ENOENT:  # No such file or directory
+                raise
+            logger.warning(__('convert command %r cannot be run.'
+                              'check the image_converter setting'),
+                           self.config.latex_svg2pdf_converter)
+            return False
+
+        try:
+            stdout, stderr = p.communicate()
+        except (OSError, IOError) as err:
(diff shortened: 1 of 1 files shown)


tk0miya:
-1; it's hard to support them on this project because we don't have enough time. How about third party extension?

sphinx/ext/latex_svg2pdf_converter.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+"""
+    sphinx.ext.latex_svg2pdf_converter
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    SVG to PDF converter extension for LaTeX builder
```
tk0miya, line 6: Why only for LaTeX builder?

sephalon: HTML, ePUB and so on handle SVG fine (if the viewer supports it, but that's another story). AFAIK LaTeX is the only one that needs mandatory conversion to PDF. However we could provide a configuration option to fine-tune conversion rules.

sephalon:
@tk0miya Yeah, I kind of expected that ;-) In that case I'd like to collect some opinions here and publish it as a third-party extension.

sphinx/ext/latex_svg2pdf_converter.py
tk0miya: `ImageConveter` will be called only if the builder does not support the image format.
For example, HTML builder supports `image/svg+xml` format. So this converter will not be called on that case.
https://github.com/sphinx-doc/sphinx/blob/69383f83bd55fe2394e5f43392ff231b5c643b40/sphinx/builders/html.py#L227-L228

On the other hand, LaTeX builder does not support it. But it supports `application/pdf`.
https://github.com/sphinx-doc/sphinx/blob/69383f83bd55fe2394e5f43392ff231b5c643b40/sphinx/builders/latex.py#L57

`ImageConverter` class and its subclass will be called based on `conversion_rules` mapping.
```
+    conversion_rules = [
+        ('image/svg+xml', 'application/pdf'),
+    ]
```
This mapping table consists of list of `(source format, destination format)`.

For this reason, you don't need to restrict this only for LaTeX builder.

tk0miya:
Good to know. I agreed this is very useful. +1!

jfbu:
I have a proposal about Mac OS: as explained at https://bugs.launchpad.net/inkscape/+bug/1449251, when using a binary install of Inkscape on Mac OS X (as one gets from the Download page, currently https://inkscape.org/en/release/0.92.2/mac-os-x/107/dmg/ is latest available, 0.92.3 must be compiled from sources), one must use filename with absolute paths.

For example, this

```
$ /Applications/Inkscape.app/Contents/Resources/script -z --export-pdf=/Users/xxxxxx/Downloads/pic1.pdf testpsspecials.svg

** (inkscape-bin:1493): WARNING **: Can't open file: testpsspecials.svg (doesn't exist)
```

fails and one must use

```console
$ /Applications/Inkscape.app/Contents/Resources/script -z --export-pdf=/Users/xxxxxx/Downloads/pic1.pdf /Users/xxxxxx/Downloads/testpsspecials.svg
```

Thus to better support Inkscape on Mac OS X, you would need also to check for existence of ``/Applications/Inkscape.app/Contents/Resources/script`` and additionally to make sure all used filenames are absolute.

sephalon:
I moved the extension into [sphinxcontrib-svg2pdfconverter](https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter). Let's continue the discussion over there; pull request can be closed.

@jfbu At least under Linux, Sphinx already uses absolute paths by default. I don't have macOS available; could you please test the extension on your system?

sphinx/ext/latex_svg2pdf_converter.py
sephalon: Thanks, you're completely right. For some reason I must have missed that…

jfbu:
@sephalon 

Yes, Sphinx imgconverter uses absolute paths which is good for Inkscape on MacOS. I have made simple test of your https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter and it seems to work fine. I added to PATH 

```
/Applications/Inkscape.app/Contents/Resources/bin
```

because I only have Inkscape installed as a MacOS application. I thought that invoking it as in my [comment](https://github.com/sphinx-doc/sphinx/pull/4772#issuecomment-375984109) with ``Resources/script -z`` would have been the better way (I am no expert user and copied pasted some long time ago from the net) but it seems to behave like ``inkscape``, by this I mean in both cases an XQuartz session is opened, which adds a bit overhead first time.

Not being Inkscape-knowledgeable I don't know if one can run it on the Mac from a script without having XQuartz launched, when Inkscape is installed only in /Applications.

Because all seems well, I add my comments here, not at https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter because I see no immediate issue to raise ;-).

**edit**: I have found one, on ``make clean html`` Inkscape seems to be invoked in some way despite no conversion needed because XQuartz gets opened.

Ah well, console output seems to give no account of your extension activity, perhaps it could be useful to have a verbose mode? (github has no good interface for discussions like this...)

sephalon:
@jfbu 

> Ah well, console output seems to give no account of your extension activity, perhaps it could be useful to have a verbose mode? (github has no good interface for discussions like this...)

You can see debug messages if you invoke Sphinx with `-vv` (by e.g. adding them to `SPHINXOPTS` in the Makefile).

PR #4773, opened 2018-03-23
Refactor config

tk0miya:
### Feature or Bugfix
- Refactor

### Purpose
- Make `Config` class simple container.


sphinx/config.py
```diff
@@ -159,30 +148,27 @@ class Config(object):
...
+        # type: (Any) -> None
+        if len(args) == 4:
+            # old style arguments: (dirname, filename, overrides, tags)
+            warnings.warn('The argument of Config() class has been changed. '
+                          'Use Config.read() to read configuration from conf.py.',
+                          RemovedInSphinx30Warning)
+            dirname, filename, overrides, tags = args
+            config = eval_config_file(dirname, filename, tags)
+        else:
+            # new style arguments: (config={}, overrides={})
```
tk0miya: This allows to create a vanilla `Config` object easily.
I separated constructor which reading `conf.py` to `Config.read()` method.

sphinx/config.py
```diff
@@ -391,9 +370,76 @@ def convert_source_suffix(app, config):
...
+                logger.warning(msg.format(name=confval.name,
+                                          current=type(confval.value),
+                                          default=type(default)))
+
+
 def setup(app):
     # type: (Sphinx) -> Dict[unicode, Any]
     app.connect('config-inited', convert_source_suffix)
+    app.connect('config-inited', correct_copyright_year)
+    app.connect('config-inited', check_confval_types)
```
tk0miya: This refactoring separates config-manipulators from `Config` class.
I believe they are not responsibility of the class.

sphinx/config.py
```diff
@@ -201,6 +194,13 @@ def __init__(self, dirname, filename, overrides, tags):
                     config[k] = copyright_year_re.sub(r'\g<1>%s' % format_date('%Y'),
                                                       config[k])
 
+    @classmethod
+    def read(cls, confdir, filename, overrides=None, tags=None):
```
stephenfin: I have a proof of concept patch ready that allows us to read INI-style configuration files. Could we call this something less generic? Maybe `read_py` or `load_py`?

sphinx/config.py
```diff
@@ -201,6 +194,13 @@ def __init__(self, dirname, filename, overrides, tags):
...
+        # type: (unicode, unicode, Dict, Tags) -> Config
+        """Create a Config object from configuration file."""
+        namespace = eval_config_file(confdir, filename, tags)
+        return Config(namespace, overrides or {})
```
stephenfin: You should probably use `cls` here.

    return cls(namespace, overrides or {})

This also means people can subclass this if necessary.

sphinx/config.py
```diff
@@ -365,6 +365,37 @@ def filter(self, rebuild):
         return (value for value in self if value.rebuild in rebuild)  # type: ignore
 
 
+def eval_config_file(confdir, filename, tags):
+    # type: (unicode, unicode, Tags) -> Dict[unicode, Any]
+    """Evaluate a config file."""
+    if confdir is None:
```
stephenfin: This doesn't belong here. This is a side-effect of how `sphinx-build` is implemented, whereby not passing `confdir` indicates that we're loading configuration from the command line instead (the `-C` option). Given that it's not possible to load configuration _without_ a config file, we can remove this from here and make `confdir` a required parameter.

sphinx/config.py
```diff
@@ -422,9 +413,24 @@ def convert_source_suffix(app, config):
...
+    """correct values of copyright year that are not coherent with
+    the SOURCE_DATE_EPOCH environment variable (if set)
+
+    See https://reproducible-builds.org/specs/source-date-epoch/
+    """
+    if getenv('SOURCE_DATE_EPOCH') is not None:
+        for k in ('copyright', 'epub_copyright'):
+            if k in config:
+                replace = r'\g<1>%s' % format_date('%Y')
+                config[k] = copyright_year_re.sub(replace, config[k])  # type: ignore
```
stephenfin, line 389: Yup, great idea :+1:

sphinx/config.py
stephenfin: I'm not so sure about this one. Unlike `check_copyright_year` and `convert_source_suffix`, which are concerned with specific options, this one concerns the public API for adding options. It seems like this is less about manipulation and more about validation. Surely validation should remain part of the functionality of the class?

stephenfin (changes requested):
Mostly that leaking of `sphinx-build` functionality into `Config` that I'd like to see removed, but there's other stuff here that's debatable too.

sphinx/config.py
tk0miya: Good point. I think `Config.read_from_conf_py()` or `Config.from_conf_py()` is better. What do you think?

ping @shimizukawa

sphinx/config.py
tk0miya: Absolutely

tk0miya: fixed. (will be pushed later)

sphinx/config.py
tk0miya: I added this to make `Config.__init__()` compatible with old class. But, as you said, it is not related with this function. So I moved this if-block to `Config.__init__()`.
Thanks,

sphinx/config.py
tk0miya: In this PR, I changed the `Config` class to simple container. I think this validator is too big for the container. So I'd like to divide it.
As a side effect, we can improve the validator simply. Personally, I'd like to modify more and more. It can't represent container types like "a list of int".

stephenfin: Ok, you know best so I'll trust your judgement on this :smile:

sphinx/config.py
```diff
@@ -192,59 +181,18 @@ def __init__(self, dirname, filename, overrides, tags):
...
-        # correct values of copyright year that are not coherent with
-        # the SOURCE_DATE_EPOCH environment variable (if set)
-        # See https://reproducible-builds.org/specs/source-date-epoch/
-        if getenv('SOURCE_DATE_EPOCH') is not None:
-            for k in ('copyright', 'epub_copyright'):
-                if k in config:
-                    config[k] = copyright_year_re.sub(r'\g<1>%s' % format_date('%Y'),
-                                                      config[k])
+    @classmethod
+    def read(cls, confdir, filename, overrides=None, tags=None):
```
stephenfin: I think we can go one step further and combine `confdir` and `filename` into something like a `path` attribute. There's no reason to keep these separate anymore, IMO.

sphinx/config.py
```diff
@@ -350,19 +298,50 @@ def __contains__(self, name):
...
 
     def filter(self, rebuild):
         # type: (Union[unicode, List[unicode]]) -> Iterator[ConfigValue]
         if isinstance(rebuild, string_types):
             rebuild = [rebuild]
-        return (value for value in self if value.rebuild in rebuild)  # type: ignore
+        return (value for value in self if value.rebuild in rebuild)
+
+
+def eval_config_file(confdir, filename, tags):
```
stephenfin: Same comment here. You could just replace `confdir` and `filename` with `path` or similar. This seems like an improvement to me, anyway.

sphinx/application.py
```diff
@@ -187,8 +185,8 @@ def __init__(self, srcdir, confdir, outdir, doctreedir, buildername,
 
         # read config
         self.tags = Tags(tags)
-        self.config = Config(self.confdir, CONFIG_FILENAME,
-                             confoverrides or {}, self.tags)
+        self.config = Config.read(self.confdir, CONFIG_FILENAME,
```
stephenfin: I think the logic to decide if we want to load from a file or not belongs here. Something like this would work, IMO.

    if self.confdir:
        self.config = Config.read(...)
    else:
        self.config = Config({}, confoverrides or {})

I may have got the API call wrong, but this should be roughly what is needed.

sphinx/config.py
stephenfin: Of the two, I prefer `Config.read_from_conf_py()`. `Config.read_py()` also seems fine though. Alternatively, I could add a `mode` parameter that we set to `ini` or `py` later. Whatever you prefer.

    def read(cls, path, overrides=None, tags=None, mode='py'):

sphinx/config.py
```diff
@@ -159,30 +148,30 @@ class Config(object):
...
-        # type: (unicode, unicode, Dict, Tags) -> None
+    def __init__(self, *args):
+        # type: (Any) -> None
+        if len(args) == 4:
+            # old style arguments: (dirname, filename, overrides, tags)
+            warnings.warn('The argument of Config() class has been changed. '
+                          'Use Config.read() to read configuration from conf.py.',
+                          RemovedInSphinx30Warning)
+            dirname, filename, overrides, tags = args
+            if dirname is None:
```
stephenfin, line 155: I still don't think this should go here. Instead, it should go into `sphinx.application.Application`. I've added a comment to this effect above.

stephenfin (changes requested):
Almost there. I'd like one further change and I'd be happy to +1

sphinx/application.py
tk0miya: Agreed. New code is more clearly what we're going to (reading conf.py or create an empty one).

sphinx/config.py
tk0miya: Unfortunately, this is needed to keep compatibility for who uses `Config` class directly.
(I don't know he/she really exists. But everyone loves making extensions and patches)

sphinx/application.py
tk0miya: Note: With this PR, we can get an empty `Config` object with `Config()`. It is one of goal of this PR. We can always get a bare object simply. I believe it makes our test simple and easy.

sphinx/config.py
tk0miya: Indeed. combined.

sphinx/config.py
tk0miya: done.

sphinx/config.py
tk0miya: :-)

sphinx/config.py
```diff
@@ -192,59 +181,18 @@ def __init__(self, dirname, filename, overrides, tags):
...
-                        name=name, current=type(current),
-                        permitted=str([cls.__name__ for cls in permitted])))
-                else:
-                    logger.warning(CONFIG_TYPE_WARNING.format(
-                        name=name, current=type(current), default=type(default)))
+        warnings.warn('Config.check_types() is deprecated. Use check_confval_types() instead.',
+                      RemovedInSphinx30Warning)
+        check_confval_types(None, self)
 
     def check_unicode(self):
```
tk0miya, line 194: Note: This validator will be unneeded in Sphinx-2.0 because py3 supports naturally.
It might be better to mark "deprecated" on this time to drop it early.

But this should be invoked before i18n-init. So we can't use `config-inited` event for this.
Hence we should call the separated helper function manually:
```
# in application.py
config = Config.read(confdir, ...)
check_unicode(config)
```

Any comment?

sphinx/config.py
```diff
@@ -192,59 +181,18 @@ def __init__(self, dirname, filename, overrides, tags):
...
-        # correct values of copyright year that are not coherent with
-        # the SOURCE_DATE_EPOCH environment variable (if set)
-        # See https://reproducible-builds.org/specs/source-date-epoch/
-        if getenv('SOURCE_DATE_EPOCH') is not None:
-            for k in ('copyright', 'epub_copyright'):
-                if k in config:
-                    config[k] = copyright_year_re.sub(r'\g<1>%s' % format_date('%Y'),
-                                                      config[k])
+    @classmethod
+    def read(cls, filename, overrides=None, tags=None):
```
tk0miya: Note: I'm waiting for a comment from @shimizukawa about naming `Config.read()`.
https://github.com/sphinx-doc/sphinx/pull/4773#discussion_r176955543

sphinx/config.py
stephenfin: Ohhh, wait - this is only for the old style config. Yup, this is fine so.

sphinx/application.py
```diff
@@ -187,10 +185,13 @@ def __init__(self, srcdir, confdir, outdir, doctreedir, buildername,
 
         # read config
         self.tags = Tags(tags)
-        self.config = Config(self.confdir, CONFIG_FILENAME,
-                             confoverrides or {}, self.tags)
-        self.config.check_unicode()
-        # defer checking types until i18n has been initialized
+        if self.confdir is None:
+            self.config = Config()
```
stephenfin: It's possible that the user has provided `confoverrides` but no config file, using a combination of `-C` (no config file) with multiple `-D` (override a general setting) and `-A` (override a `html_context` setting) options. As a result, I think you should pass in `confoverrides` here. I noted this earlier but reproducing here.

```
if self.confdir:
    self.config = Config.read(...)
else:
    self.config = Config({}, confoverrides or {})
```

stephenfin: You'll also want to move `check_unicode` outside the if-else block, if you do this.

stephenfin (changes requested):
Think you missed something

sphinx/application.py
tk0miya: You're right. I have to give `confoverrides` in this case.
Thanks,

tk0miya: done

stephenfin (commented):
Still awaiting feedback on the `read` function, but this is otherwise good to me. Can approve once that's done.

sphinx/config.py
shimizukawa: My preference is:

1. `Config.from_conf_py()`
2. `Config.read_from_conf_py()`

I think `mode='py'` parameter is not good.

shimizukawa (approved):
LGTM!

tk0miya:
@shimizukawa thank you for comment. updated!

tk0miya:
@stephenfin any comments?

stephenfin:
> @stephenfin any comments?

My one comment is that I'm trying to think how this would work with an INI-style config file. If we had that, would the expectation be that we would support/require both (like `setup.py` and `setup.cfg`) or you could have one or the other? If the former, perhaps we actually want to modify the signature of `from_conf_py` to take the path to the config *directory* instead of the config *file*. For example:

    def read(cls, confdir, overrides=None, tags=None):
        conf_py = os.path.join(confdir, 'conf.py')
        conf_ini = os.path.join(confdir, 'sphinx.ini')
        if os.path.exists(conf_py):
            ...

If it's the latter, what we have is clearly OK. I would personally prefer to support both as it would allow you to declare most stuff in the INI file while deferring to the `conf.py` for more dynamic stuff/small extensions etc.

Sorry if I sent you the wrong direction. I only spotted it when I tried rebasing my work on top of this.

tk0miya:
TBH, I've never had any thoughts about config-ini. So I don't have well thought out comment. But I can't imagine the case using both .ini and .py.

stephenfin:
> TBH, I've never had any thoughts about config-ini. So I don't have well thought out comment. But I can't imagine the case using both .ini and .py.

Hmm, so we need to think about that more. However, in that case how about the change above, i.e. calling it `read` again and passing in `confdir` instead (this one makes sense even with the current function name - it's reading a `conf.py` file so why do we need to pass that information)

tk0miya:
I think API should follow the purpose. So we should discuss about config-ini first.
But I feel the feature is not a blocker of this PR. It should be an another topic.
Let's make an issue and start discussion if you want. Not to late to modify the API after discussion.

tk0miya:
I'll merge this in this weekend if no objection.

stephenfin:
> I'll merge this in this weekend if no objection.

Personally, I would like to see the changes I suggested above just so we don't have to rework the API in the future. To be clear, that would mean using the signature:

    def read(cls, confdir, overrides=None, tags=None):

Instead of:

    def from_conf_py(cls, filename, overrides=None, tags=None):

Even if we don't add INI support, this pattern still makes sense (IMO) because we always use the name `conf.py` so passing in that full path is unnecessary. It would totally make life easier if we did add `sphinx.ini` support though (still working on it). However, I certainly won't block on this and if you would prefer to keep what we have then I can't/won't stop you :smile:

tk0miya:
If filename is omitted, who checks the existence of the conf.py? `Config` class or `Sphinx` class?
My concerns are 1) who should know the filename of `conf.py` and 2) How the `Config.read()` behaves if `conf.py` not found.

In this PR:
* 1) `Sphinx` class only knows. `Config` class does not mind the filename.
* 2) It raises an exception. It means loading configuration was failed.

stephenfin:
> If filename is omitted, who checks the existence of the conf.py? Config class or Sphinx class?

Personally, I think it should be `Config` class - after all, it is responsible for loading configuration. If it's unable to do so (because the file does not exist), it should alert you. You'll note we already kind of do this - `sphinx-build` doesn't have an option to point you to the `conf.py` file but rather an option to point you to [the `confdir`](https://github.com/sphinx-doc/sphinx/blob/b5ddc8c6977ac964e3ba4a2eb4647decdc742f4f/sphinx/cmd/build.py#L156-L158).

> My concerns are 1) who should know the filename of conf.py and 2) How the Config.read() behaves if conf.py not found.
> 
> In this PR:
> 
> - Sphinx class only knows. Config class does not mind the filename.
> - It raises an exception. It means loading configuration was failed.

I think 1) `Config` because that's what doing the loading and 2) `Config` should should raise an exception, though `Sphinx` may catch this and raise its own exception. I would only change this if we decide to allow users to define their own filename in the future. However, I don't see why we would do that.

tk0miya:
Thank you for comment. I understood and agreed with you.

>If filename is omitted, who checks the existence of the conf.py? Config class or Sphinx class?

I was confused at this time. Even at present, only `Config` class checks the existence. `Sphinx` does not. It only checks the existence of *confdir*, not `conf.py`.

So, I give +1 for `Config.read(confdir)`.

@shimizukawa any comments?

tk0miya:
Okay, it seems @shimizukawa is busy. Let's go with `Config.read(confdir)`.

tk0miya:
Updated.
@stephenfin If you're okay, please approve this. Then, I'll merge this soon.

stephenfin:
Yup, this looks good to me. If you squashed all the commits when you merge then it would make the commit history easier to read, but I'll let you make that decision :smile:

shimizukawa:
Oops, I've overlooked. Please go ahead :)

tk0miya:
Merged (without squash).
Thank you for reviewing!

Issue #4725, opened 2018-03-08
Use of typing module introduced in python 3.5


analog-cbarber:
It appears that starting with sphinx 1.7.1, the code uses the typing module that was introduced in python 3.5. This presumably means the code will no longer work with python 3.4, which is still documented as being the oldest supported version. Furthermore the use of the `TYPE_CHECKING` attribute means that versions prior to python 3.5.3 will not work (which is how I noticed this).

For instance, see [here](https://github.com/sphinx-doc/sphinx/blob/2994c125132b6beb73c51de4cc224a2b1d55f741/sphinx/cmd/build.py#L19)

It seems like you should either avoid the typing module entirely, conditionalize the imports somehow or update your documented dependencies to state that you require 3.5.3.

tk0miya:
OMG.... I thought the `TYPE_CHECKING` is supported since 3.5. I'll check it ASAP.
Thank you for letting me know!

analog-cbarber:
The doc for `TYPE_CHECKING` does not state that it was not in the original version of the typing module, so there would be no way to know if you didn't explicitly check older versions.

In any case, there is still the issue that the typing module itself was not present in 3.4. Is sphinx dropping 3.4 support?

tk0miya:
Sphinx depends on `typing` backports package when <= 3.4. So py34 is still supported.
https://pypi.python.org/pypi/typing

cocoatomo:
`typing.TYPE_CHECKING` is introduced at this commit https://github.com/python/cpython/commit/91185fe0284a04162e0b3425b53be49bdbfad67d#diff-015978a768b517a38abbc0ecdea87f5aR72 (on 9 Jun 2016),
and the Python version at this point was 3.5.1.
3.5.1 final was released on December 6, 2015 according [PEP 478](https://www.python.org/dev/peps/pep-0478/), so Sphinx should use `typing` backports package when <= 3.5.0. (Maybe that case are rare.)

analog-cbarber:
For what its worth, I observed this problem on a system that says it has python 1.5.1, and I definitely did not install a prerelease. Furthermore, if you look at the online docs for the typing module the `TYPE_CHECKING` variable does not show up until 1.5.3.  You should check actual deployments before making any conclusions about what versions work.

I just created conda environment for python 1.5.1 on Mac OSX and sure enough there is no `TYPE_CHECKING`. It does appear in the 1.5.2.

So you should use the backports package when <= 3.5.1

[...]

PR #4780, opened 2018-03-25
Drop typing.TYPE_CHECKING

tk0miya:
refs: #4725

### DIFF
--- a/sphinx/apidoc.py
+++ b/sphinx/apidoc.py
@@ -10,12 +10,12 @@
 """
 
 import warnings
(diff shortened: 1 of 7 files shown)


stephenfin (approved):
I'd personally have just reverted b6c2ba0d57 as it would make this easier to understand to the casual developer. However, this does the job.

tk0miya:
>I'd personally have just reverted b6c2ba0 as it would make this easier to understand to the casual developer.

Ah, you're right. But I'll go with this.

tk0miya:

PR #4781, opened 2018-03-25
Deprecate env._nitpick_ignore

tk0miya:
### Feature or Bugfix
- Refactor

### Purpose
- According to the comment, `env._nitpick_ignore` was added to be updated. But it never updated from anywhere.
- This drops it, and refers `config.nitpick_ignore` directly.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -32,6 +32,7 @@ Deprecated
   deprecated
 * ``sphinx.locale.l_()`` is deprecated
 * #2157: helper function ``warn()`` for HTML themes is deprecated
+* ``env._nitpick_ignore`` is deprecated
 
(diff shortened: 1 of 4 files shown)


tk0miya:
Note: It was added to `BuildEnvironment.__init__()` at f5ecd62d05, and moved to `BuildEnvironment.update()` at 123572e1db.
I think the comment "to be updated every time" means `env._nitpick_ignore` should be updated on "every build".

It does not mean `env._nitpick_ignore` is updated during build. So we can drop it, and we can read it from config object simply.

stephenfin (approved):
LGTM

tk0miya:

PR #4788, opened 2018-03-27
Fix #4787: Update docstrings for add_directive()

### LINKED ISSUE #4787: Wrong argument name?
In https://github.com/sphinx-doc/sphinx/blob/a014caa4f04dd69df5fabd855080c54259557c4e/sphinx/application.py#L674

There is no argument called ``directiveclass`` should this be ``obj``? Sorry if this is a stupid question - I am new to Sphinx.

All the best,
Morten

tk0miya:
### Feature or Bugfix
- Docs

### Purpose
- see #4787

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -671,7 +671,7 @@ class Sphinx(object):
           function and determine whether the directive has content, arguments
           and options, respectively.  **This style is deprecated.**
 
-        - In the docutils 0.5 style, *directiveclass* is the directive class.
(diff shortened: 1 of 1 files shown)


stephenfin (approved):
Looks good to me. As an aside, are we going to deprecate the older style at some point?

tk0miya:
Thank you for reviewing.

To make `directive_helper` deprecated at same time, I added deprecation warning into the helper. So old style is marked as deprecated fine.

PR #4793, opened 2018-03-28
Let LaTeX with XeTeX/LuaTeX use normal sized typeface in code-blocks

### LINKED ISSUE #4791: latex builds should not use "small" typeface for code-blocks with xelatex/lualatex
For code-blocks in PDF builds, Sphinx uses by default `\fvset{fontsize=\small}`. This is because the legacy default font choice for pdflatex build was Courier which is bigger in apparence than Time which is used as text font. Furthermore, once upon a time, long lines would not wrap in code-blocks.

For xelatex/lualatex engines, the font is by default the one chosen by LaTeX for these engines, i.e. Latin Modern OpenType. Then the `\small` is really too small, I believe. I propose that Sphinx should not use `\small` by default then.

Here is a screenshot from a PDF build of the SciPy documentation v1.0.0. I added use of `xelatex` because of many Unicode characters.

capture d ecran 2018-03-28 a 00 27 53

I feel the code-blocks uses a too small typeface. Here is at normal size.

capture d ecran 2018-03-28 a 00 32 30

jfbu:
closes: #4791

I have tested normal size typeface in codeblocks with xelatex as latex engine for a build of the v1.0.0 SciPy docs, and it looks much better with normal size than small size, because the monospaced font is already of the same design family as the font for main text.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -2007,7 +2007,7 @@ These options influence LaTeX output. See further :doc:`latex`.
         ``'\\fvset{fontsize=auto}'``, because the default fonts are part of
         one unified typeface family (Latin Modern OpenType).
 
-        .. versionadded:: 1.7.3
+        .. versionadded:: 1.8
    * Keys that are set by other options and therefore should not be overridden
      are:
 
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1998,7 +1998,16 @@ These options influence LaTeX output. See further :doc:`latex`.
         differently or append some content after the index. For example
         ``'\\footnotesize\\raggedright\\printindex'`` is advisable when the
         index is full of long entries.
+     ``'fvset'``
+        Customization of ``fancyvrb`` LaTeX package. Defaults to
+        ``'\\fvset{fontsize=\\small}'``, because default font (Courier) used in
+        code-blocks is wider and taller than default text font (Times).
 
+        For ``'xelatex'`` and ``'lualatex'``, defaults to
+        ``'\\fvset{fontsize=auto}'``, because the default fonts are part of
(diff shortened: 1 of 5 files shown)


doc/config.rst
```diff
@@ -1998,7 +1998,16 @@ These options influence LaTeX output. See further :doc:`latex`.
...
+     ``'fvset'``
+        Customization of ``fancyvrb`` LaTeX package. Defaults to
+        ``'\\fvset{fontsize=\\small}'``, because default font (Courier) used in
+        code-blocks is wider and taller than default text font (Times).
 
+        For ``'xelatex'`` and ``'lualatex'``, defaults to
+        ``'\\fvset{fontsize=auto}'``, because the default fonts are part of
+        one unified typeface family (Latin Modern OpenType).
+
+        .. versionadded:: 1.7.3
```
tk0miya: At present, the target of this PR is master branch. Is this right?
If I had to choose a target, I like master branch a bit because this is not a obvious bug.

jfbu: @tk0miya Yes this is right. I chose master branch because it is not obvious bug.

jfbu: It is only default, and user can override it already to fix it without waiting for 1.8. (long time ago it was not customizable)

jfbu: ah sorry I see what you mean, ok I will edit to 1.8, OMG!

jfbu: fixed...

tk0miya: Confirmed. Go ahead :-)

jfbu: Thanks @tk0miya, now merging into master.

PR #4794, opened 2018-03-29
installation.rst: correct Windows Run keyboard shortcut

kenyon:
You type Windows-R to get the Run dialog, not Control-R.

### DIFF
--- a/doc/usage/installation.rst
+++ b/doc/usage/installation.rst
@@ -108,7 +108,7 @@ Windows
 
 Most Windows users do not have Python installed by default, so we begin with
 the installation of Python itself.  If you are unsure, open the *Command
-Prompt* (:kbd:`Control-r` and type :command:`cmd`).  Once the command prompt is
+Prompt* (:kbd:`⊞Win-r` and type :command:`cmd`).  Once the command prompt is
 open, type :command:`python --version` and press Enter.  If Python is
 available, you will see the version of Python printed to the screen.  If you do
 not have Python installed, refer to the `Hitchhikers Guide to Python's`__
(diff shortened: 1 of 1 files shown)


stephenfin (approved):
Oh, that was dumb. Good catch. I'm a little concerned about the use of the unicode characters, but the docs build seems OK with it so why not

tk0miya:
Thanks!

jfbu:
This causes failed PDF build

```
! Package inputenc Error: Unicode char ⊞ (U+229E)
```

I will fix later.

jfbu:
(fixed Unicode issue for PDF builds via pdflatex at 39ce9d9ab)

jfbu:
At https://en.wikipedia.org/wiki/Windows_key, uppercase letters are used in the description ``⊞ Win + R`` for example (and I used a no break space before Win). What is common habit in Windows world? (I am completely ignorant)

stephenfin:
Do we not build this stuff with CI? If not, we should add something to ensure this doesn't happen again.

jfbu:
@stephenfin At my locale `make latexpdf` on a clean state takes about 1minute

```
real	0m57.888s
user	0m57.108s
sys	0m0.528s
```

compared to 12.5s for html

```
The HTML pages are in _build/html.

real	0m12.628s
user	0m12.400s
sys	0m0.207s
```

The overhead in PDF build isn't into the actual pdflatex processing but into the Sphinx build of the LaTeX source.

I have not checked our CI settings, but building PDF docs might have been deactivated because of that. When building first html then latexpdf, the latter takes only about 48s. TIming both together I get about 62s.

PR #4798, opened 2018-03-29
refactor: Patch docutils.frontend.OptionParser only during Sphinx processing

tk0miya:
### Feature or Bugfix
- Refactor

### Purpose
- So far, Sphinx overrides the attributes of `OptionParser`. And they are not restored after builds.
- This tries to patch it only during builds.

### DIFF
--- a/sphinx/cmd/build.py
+++ b/sphinx/cmd/build.py
@@ -293,7 +293,7 @@ def build_main(argv=sys.argv[1:]):  # type: ignore
 
     app = None
     try:
-        with patch_docutils(), docutils_namespace():
+        with patch_docutils(args.sourcedir), docutils_namespace():
             app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
                          args.doctreedir, args.builder, confoverrides, status,
                          warning, args.freshenv, args.warningiserror,
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -19,7 +19,6 @@ from collections import defaultdict
 from copy import copy
 from os import path
 
-from docutils.frontend import OptionParser
 from docutils.utils import Reporter, get_source_line
 from six import BytesIO, itervalues, class_types, next
 from six.moves import cPickle as pickle
@@ -562,9 +561,8 @@ class BuildEnvironment(object):
         """Parse a file and add/update inventory entries for the doctree."""
         self.prepare_settings(docname)
 
+        # Add srccdir/docutils.conf to dependencies list if exists
         docutilsconf = path.join(self.srcdir, 'docutils.conf')
-        # read docutils.conf from source dir, not from current dir
-        OptionParser.standard_config_files[1] = docutilsconf
         if path.isfile(docutilsconf):
             self.note_dependency(docutilsconf)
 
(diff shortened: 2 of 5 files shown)


sphinx/environment/__init__.py
```diff
@@ -562,9 +561,8 @@ def read_doc(self, docname, app=None):
         """Parse a file and add/update inventory entries for the doctree."""
         self.prepare_settings(docname)
 
+        # Add srccdir/docutils.conf to dependencies list if exists
         docutilsconf = path.join(self.srcdir, 'docutils.conf')
-        # read docutils.conf from source dir, not from current dir
-        OptionParser.standard_config_files[1] = docutilsconf
```
tk0miya, line 567: Sphinx only overrides `standard_config_files[1]`.
The original value is here:
```
    standard_config_files = [
        '/etc/docutils.conf',           # system-wide
        './docutils.conf',              # project-specific
        '~/.docutils']                  # user-specific
```

Should we support both system-wide and user-specific docutils.conf?

stephenfin: I think this would be a good idea, but it might be confusing for some users as it will change behaviour. Maybe add a TODO to do this for 2.0, which is a major change and will allow us to do such a thing.

tk0miya: My last comment is not clear. What I want to say is " is this support really needed?".
I think system-wide and user-specific configuration changes output only locally. It means the outputs can't be reproduced on other environments.

In addition, our document does not mention about both of them. It only mentions about project-specific one.

>An optional file docutils.conf can be added to the configuration directory to adjust Docutils configuration if not otherwise overriden or set by Sphinx.
http://www.sphinx-doc.org/en/master/config.html#module-conf

stephenfin: Oh, then my comment still stands but in reverse. I think we _should_ use them because that's what docutils does, but if we remove this feature then we should wait for a major version bump.

shimizukawa: For document builds, I think that the same product should be created independent of the environment. However, if `/etc/docutils.conf` is referenced, different results may be generated for each environment. It is good to refer only to the files included in the Sphinx project.

shimizukawa (approved):
LGTM :)

sphinx/environment/__init__.py
tk0miya: I noticed `OptionParser` also refers `$DOCUTILSCONFIG` envvar. docutils gives priority to it rather than `standard_config_files`. It means `{confdir}/docutils.conf` might be ignored if `$DOCUTILSCONFIG` is set.

I think it is not good for Sphinx. We should find docutils.conf with our own searching rules.

sphinx/cmd/build.py
```diff
@@ -293,7 +293,7 @@ def build_main(argv=sys.argv[1:]):  # type: ignore
 
     app = None
     try:
-        with patch_docutils(), docutils_namespace():
+        with patch_docutils(args.sourcedir), docutils_namespace():
```
tk0miya, line 296: OMG, this argument is wrong. It must be confdir instead.

tk0miya:
I'm withdrawing this because of `$DOCUTILSCONF`.

PR #4799, opened 2018-03-29
Doc rework (user guide, part 4)

stephenfin:
Add markdown and config sections to the user guide. As before, this is mostly done by shuffling around existing content.


doc/usage/configuration.rst
```diff
@@ -1644,7 +1645,8 @@ the `Dublin Core metadata <http://dublincore.org/>`_.
 Options for LaTeX output
 ------------------------
 
-These options influence LaTeX output. See further :doc:`latex`.
+These options influence LaTeX output. Refer to :doc:`/latex` for more
```
jfbu, line 1701: why not move `latex.rst` also to `usage`?

jfbu (commented):
What is motivation for moving config.rst to ``usage/``. The also ``latex.rst`` should be moved there, and also intl, and also ``templating.rst``, it is hard to draw a limit.

-1 for moving ``config``to ``usage/``.

doc/usage/configuration.rst
```diff
@@ -36,8 +34,8 @@ Important points to note:
   ``"sphinx.builders.Builder"`` means the ``Builder`` class in the
   ``sphinx.builders`` module.
 
-* Remember that document names use ``/`` as the path separator and don't contain
-  the file name extension.
+* Remember that document names use ``/`` as the path separator and don't
+  contain the file name extension.
```
jfbu, line 40: original sentence has 80 chars, why is it wrapped? it looks as if in fact the entire file was wrapped at shorted length perhaps 78 chars. edit: not sure about this, perhaps 79, perhaps even 80.

jfbu (commented):
lots of the diff for config.rst is for line wrap. It seems to have been done globally (with some exceptions) and as a result some paragraphs use longer lines. It would be clearer to modify only needed locations with no other changes in a single commit for easier review.

Some lines are still overlong such as the one starting with `` If nonempty, an `OpenSearch ``
or ``Output is processed with HTML5 writer``.

edit: using ``whitespace-mode`` in an Emacs buffer with the variable ``fill-column`` set to 80 and the variable ``whitespace-line-column`` set to ``nil`` allows to identify quickly the few paragraphs in ``config.rst`` with overlong lines. It seems to be quite less than what is implied by this changeset.

doc/usage/configuration.rst
stephenfin: Yup, I plan to move this but this is only part 4. I'm moving stuff one by one so it's possible to review everything. I will move this in part 5, which moves everything related to builders (I have the changes sitting in my local repo)

stephenfin:
The `usage` section is our user guide. I plan to have three guides:

- User
- Developer (core)
- Developer (extensions)

 I want every document to belong in one of those sections. I think `config` is something you care about when using Sphinx so it belongs in that section. Do you think otherwise? If so, where would you put it?

I can separate out the re-wrapping of config to a separate commit, yes. I was using doc8 to validate my changes and those changes were necessary to make the doc pass

doc/usage/configuration.rst
stephenfin: I'm pretty sure it was 81 but I can check. I was wrapping with vim (`v` to select text then `gq` to wrap).

jfbu:
I agree `config` would belong to `User` section in this scheme. But where will `theming` end up? are people writing up themes for html "users" or "developers"? Currently, how would you describe the current docs relative to User/DevCore/DevExt scheme? I think much of it is for User. Then why not keep everything User at top level, and if needed develop the DevCore and DevExt as units. In this perpective the refactoring of docs would start by DevCore and DevExt first and leave User sort of as is.

doc/usage/configuration.rst
```diff
@@ -520,25 +520,26 @@ Project information
...
-      The default is now ``'default'``. It is similar to ``'python3'``;
-      it is mostly a superset of ``'python'``. but it fallbacks to
-      ``'none'`` without warning if failed.  ``'python3'`` and other
-      languages will emit warning if failed.  If you prefer Python 2
-      only highlighting, you can set it back to ``'python'``.
+      The default is now ``'default'``. It is similar to ``'python3'``; it is
+      mostly a superset of ``'python'``. but it fallbacks to ``'none'`` without
+      warning if failed.  ``'python3'`` and other languages will emit warning
+      if failed.  If you prefer Python 2 only highlighting, you can set it back
+      to ``'python'``.
```
jfbu: This paragraph is good example of a paragraph with originally short lines getting rewrapped to longer lines. Is there really necessity to do that? if yes I propose all such changes with no alterations in contents are strictly delimited to one commit with nothing else.

doc/usage/configuration.rst
```diff
@@ -520,25 +520,26 @@ Project information
...
 
 .. confval:: highlight_options
 
    A dictionary of options that modify how the lexer specified by
    :confval:`highlight_language` generates highlighted source code. These are
-   lexer-specific; for the options understood by each, see the
-   `Pygments documentation <http://pygments.org/docs/lexers/>`_.
+   lexer-specific; for the options understood by each, see the `Pygments
+   documentation <http://pygments.org/docs/lexers/>`_.
```
jfbu: is doc8 mandating that the link stand on a line of its own? but reStructured Text allows the linebreak.

doc/usage/configuration.rst
```diff
@@ -520,25 +520,26 @@ Project information
...
 
    .. versionadded:: 1.3
 
 .. confval:: pygments_style
 
    The style name to use for Pygments highlighting of source code.  If not set,
-   either the theme's default style or ``'sphinx'`` is selected for HTML output.
+   either the theme's default style or ``'sphinx'`` is selected for HTML
+   output.
```
jfbu, line 551: original line has exactly 80 characters. Why does it get wrapped ?

doc/usage/configuration.rst
```diff
@@ -2306,10 +2309,10 @@ Options for the linkcheck builder
...
 .. confval:: linkcheck_anchors_ignore
 
-   A list of regular expressions that match URIs that should skip checking
-   the validity of anchors in links. This allows skipping entire sites, where
-   anchors are used to control dynamic pages, or just specific anchors within
-   a page, where javascript is used to add anchors dynamically, or use the
+   A list of regular expressions that match URIs that should skip checking the
+   validity of anchors in links. This allows skipping entire sites, where
+   anchors are used to control dynamic pages, or just specific anchors within a
+   page, where javascript is used to add anchors dynamically, or use the
```
jfbu: here again the original paragraph had shorter lines, and overall had more equilibrium

jfbu (commented):
more rant about paragraph re-flowing ;-) I hope you don't mind, my point is that if we start doing that the whole documentation will get re-wrapped for not very clear reasons (to me)

doc/usage/configuration.rst
jfbu: oops, here actually the original had the link on one line in source...

stephenfin:
> I agree config would belong to User section in this scheme. But where will theming end up? are people writing up themes for html "users" or "developers"?

A theme is a sort of extension. Like any of the extension docs, it would be split into two parts: the developer part which focuses on how to write a theme, and the user part which focuses on how to enable it (e.g. `html_theme` and related options).

> Currently, how would you describe the current docs relative to User/DevCore/DevExt scheme? I think much of it is for User. Then why not keep everything User at top level, and if needed develop the DevCore and DevExt as units. In this perpective the refactoring of docs would start by DevCore and DevExt first and leave User sort of as is.

When I say _user_ I mean someone that is using Sphinx with pre-existing extensions and themes: they may not know anything about Python, HTML, LaTeX etc. Anyone that goes beyond that and starts writing their own extensions becomes a _developer_. A similar gap exists between a user and an administrator for other projects: an administrator is a superset of a user. Also note that docutils uses a [very similar design](http://docutils.sourceforge.net/docs/), with `user/` (User), `dev/` (Developer (core)) and `howto/` (Developer (extensions)) sections.

Does this make sense? If not, maybe you just need to trust me on it and give out to me if I get it wrong? :)

stephenfin:
> more rant about paragraph re-flowing ;-) I hope you don't mind, my point is that if we start doing that the whole documentation will get re-wrapped for not very clear reasons (to me)

Perfectly fine :smile: To be honest, this was a really dumb wrap - I literally went through and used Vim's auto-wrapping feature to wrap everything. This was a lot less tedious than manually figuring out what lines were too long and wrapping those (this is a big undertaking). Let me split things out and see if it improves things.

stephenfin:
> Let me split things out and see if it improves things.

I've done this now and removed some of the unnecessary wrapping. Let me know if it's clearer.

jfbu:
> This was a lot less tedious than manually figuring out what lines were too long and wrapping those (this is a big undertaking)

I had tried out my recipe using whitespace-mode in Emacs, and it had given me impression the whole config.rst could be handled rather fast (in a minute or so, max). I only have to be cautious to obey by the "two spaces end sentence" rule, which in my country isn't so much in use, but there is an Emacs variable (of course). Anyway, thanks for your work. I won't quibble more (but maybe will reproach it to you in ten years from now if we meet someday `;-)`) about docs reflowing ...

stephenfin:
> I had tried out my recipe using whitespace-mode in Emacs, ...

#Vim4life :heart: :wink:

stephenfin:
@tk0miya Is this ready to merge? @jfbu seems happy with it and I'd like to move on to the next few docs :smile:

doc/markdown.rst
```diff
@@ -1,45 +1,43 @@
 .. highlightlang:: python
 
-.. _markdown:
```
tk0miya: Please don't remove anchor. It may have been referred by third party docs (see #4801).

doc/markdown.rst
```diff
@@ -1,45 +1,43 @@
...
+#. Add the Markdown filename extension to the  ``source_suffix`` configuration
+   variable::
 
       source_suffix = ['.rst', '.md']
 
-#. You can further configure recommonmark to allow custom syntax that standard CommonMark doesn't support. Read more in
-   the `recommonmark documentation <https://recommonmark.readthedocs.io/en/latest/auto_structify.html>`__.
+#. You can further configure recommonmark to allow custom syntax that standard
+   CommonMark doesn't support. Read more in the `recommonmark documentation
+   <shttp://recommonmark.readthedocs.io/en/latest/auto_structify.html>`__.
```
tk0miya: `https`?

doc/config.rst
```diff
@@ -1179,25 +1184,25 @@ that use Sphinx's HTMLWriter class.
...
+         required.
       :'sphinx.search.ja.JanomeSplitter':
          Janome binding. To use this splitter,
          `Janome <https://pypi.org/project/Janome/>`_ is required.
 
-      To keep compatibility, ``'mecab'``, ``'janome'`` and ``'default'`` are also
-      acceptable. However it will be deprecated in Sphinx-1.6.
-
+      To keep compatibility, ``'mecab'``, ``'janome'`` and ``'default'`` are
+      also acceptable. However it will be deprecated in Sphinx-1.6.
```
tk0miya: I just made a PR for this deprecation: #4910

doc/usage/configuration.rst
```diff
@@ -2483,5 +2483,5 @@ Options for the C++ domain
 Example of configuration file
 =============================
 
-.. literalinclude:: _static/conf.py.txt
+.. literalinclude:: ../_static/conf.py.txt
```
tk0miya: Please use `/_static/conf.py.txt` instead. A path starting with `/` is considered as relative path from srcdir. It is much simpler.

tk0miya (commented):
almost LGTM

tk0miya:
Note: (for redirection settings)
* markdown -> usage/markdown
* config -> usage/configuration

doc/markdown.rst
stephenfin: Done

doc/markdown.rst
stephenfin: Done

doc/config.rst
stephenfin: Done

doc/usage/configuration.rst
stephenfin: Done

tk0miya (approved):
LGTM!

tk0miya:
Ah, it gets an error.

```
Warning, treated as error:
/home/travis/build/sphinx-doc/sphinx/doc/usage/markdown.rst:42:Duplicate explicit target name: "markdown".
ERROR: InvocationError for command '/home/travis/build/sphinx-doc/sphinx/.tox/docs/bin/python setup.py build_sphinx -v' (exited with code 1)
```

doc/usage/markdown.rst
```diff
@@ -0,0 +1,45 @@
+.. highlight:: python
+
+.. _markdown:
+
+========
+Markdown
+========
+
+`Markdown`_ is a lightweight markup language with a simplistic plain text
```
tk0miya: This hyperlink and label on the top of document are conflicted...

stephenfin: Done

tk0miya:
Merged.
Thank you always!

PR #4804, opened 2018-03-30
Add SphinxDirective class

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add `SphinxDirective` class as a base class of Directive classes.

### DIFF
--- a/sphinx/util/docutils.py
+++ b/sphinx/util/docutils.py
@@ -20,7 +20,7 @@ from distutils.version import LooseVersion
 import docutils
 from docutils import nodes
 from docutils.languages import get_language
-from docutils.parsers.rst import directives, roles, convert_directive_function
+from docutils.parsers.rst import Directive, directives, roles, convert_directive_function
 from docutils.statemachine import StateMachine
 from docutils.utils import Reporter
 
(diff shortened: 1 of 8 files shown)


sphinx/util/docutils.py
```diff
@@ -273,6 +274,26 @@ def switch_source_input(state, content):
...
+        """Reference to the :class:`.BuildEnvironment` object."""
+        return self.state.document.settings.env
+
+    @property
+    def config(self):
+        # type: () -> Config
+        """Reference to the :class:`.Config` object."""
+        return self.env.config
+
+
```
stephenfin, line 296: Would it be worth exposing `app` too?

    self.state.document.settings.env.app

stephenfin (approved):
Mostly happy with this. I've been annoyed by this in the past

sphinx/util/docutils.py
tk0miya: Good point. The reason why I did not add `app` is making Sphinx components less coupled with `app` object. I agree with `app` is very useful because it behaves as a hub. With using it, we can access all of the component.
But I guess what we really want to get is not `app` object. There are very few features belongs to `app` object.
In addition, on testing, providing `app` object is a bit expensive. To create it, we have to initialize all of submodules; config, extensions, env, builders and so on.

But this is just my idea, and not designed well. I know many extensions are strongly coupled with `app`. And it is very important and useful at present. So I can add it if it is reasonable.

shimizukawa (approved):
LGTM! I think it's good to make the code explicit and easy to understand. Thanks!

tk0miya:

Please let me know if you want `app` helper for `SphinxDirective`.

PR #4806, opened 2018-03-31
Mark raw string as such

stephenfin:
Resolves the following warning when building docs:

    sphinx/ext/autosummary/__init__.py:94: DeprecationWarning: invalid escape sequence \.
      periods_re = re.compile('\.(?:\s+)')

Signed-off-by: Stephen Finucane <stephen@that.guru>


tk0miya (changes requested):
Basically LGTM, but I think this should be merged into 1.7 branch.

tk0miya:
I cherry-picked this manually.
Thank you always!

PR #4797, opened 2018-03-29
Fix #4789: imgconverter: confused by converter.exe of Windows

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4789

PR #4808, opened 2018-04-02
Fix error handling on imgconverter

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- it seems my change on #4797 was wrong (thanks @usaturn for confirming)
- This tries to fix it again.

### DIFF
--- a/sphinx/ext/imgconverter.py
+++ b/sphinx/ext/imgconverter.py
@@ -8,6 +8,7 @@
     :copyright: Copyright 2007-2018 by the Sphinx team, see AUTHORS.
     :license: BSD, see LICENSE for details.
 """
+import locale
 import subprocess
 
 from sphinx.errors import ExtensionError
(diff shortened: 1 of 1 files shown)


sphinx/ext/imgconverter.py
```diff
@@ -52,9 +53,10 @@ def is_available(self):
             stdout, stderr = p.stdout.read(), p.stderr.read()
             p.wait()
         if p.returncode != 0:
+            encoding = sys.getdefaultencoding()
             logger.warning(__('convert exited with error:\n'
                               '[stderr]\n%s\n[stdout]\n%s'),
-                             (stderr, stdout))
+                           stderr.decode(encoding), stdout.decode(encoding))
```
tk0miya: I worried about this `decode` is needed or not. So I'm waiting for @usaturn 's feedback.

tk0miya: With current implementation, decode error is raised.
```
Encoding error:
'utf-8' codec can't decode byte 0x96 in position 0: invalid start byte
The full traceback has been saved in C:\Users\gohki\AppData\Local\Temp\sphinx-err-r5gsafxm.log, if you want to report the issue to the developers.
```

Note: `sys.getdefaultencoding()` returns `'utf-8'`. It seems `locale.getpreferredencoding()` is better in this case.

tk0miya:
This PR was confirmed working fine on Windows platform by @usaturn .
Thanks!

PR #4809, opened 2018-04-02
Update package classifiers

tk0miya:
### Feature or Bugfix
- Refactor

### Purpose
- Update package classifiers


setup.py
```diff
@@ -168,6 +168,9 @@ def _run_domain_js(self, domain):
         'Environment :: Web Environment',
         'Intended Audience :: Developers',
         'Intended Audience :: Education',
+        'Intended Audience :: End Users/Desktop',
+        'Intended Audience :: Science/Research',
+        'Intended Audience :: System Administrators',
```
tk0miya, line 173: I think the target of Sphinx is spread. Our goal contains to create ebooks, thesis and operation manuals. So I think these users are part of "Intended Audience".

shimizukawa (approved):

PR #4810, opened 2018-04-02
Proposal: Define priority for transforms

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- docutils already provides [Transform Priority Range Categories](http://docutils.sourceforge.net/docs/ref/transforms.html#transform-priority-range-categories).
- But there are no standards for priority in Sphinx. So I guess developers are confused what priority they should use.
- In this PR, I made a priority table for Sphinx. I believe this will help developers to choose better priority.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -968,7 +968,30 @@ class Sphinx(object):
         Add the standard docutils :class:`Transform` subclass *transform* to
         the list of transforms that are applied after Sphinx parses a reST
         document.
-        """
+
+        .. list-table:: priority range categories for Sphinx transforms
+
+           * - Priority
+             - Main purpose in Sphinx
+           * - 0-99
+             - Fix invalid nodes by docutils. Translate a doctree.
+           * - 100-299
+             - Preparation
+           * - 300-399
+             - early
+           * - 400-699
+             - main
+           * - 700-799
+             - Post processing. Deadline to modify text and referencing.
+           * - 800-899
+             - Collect referencing and referenced nodes. Domain processing.
+           * - 900-999
+             - Finalize and clean up.
+
+        refs: `Transform Priority Range Categories`__
+
+        __ http://docutils.sourceforge.net/docs/ref/transforms.html#transform-priority-range-categories
+        """  # NOQA
         self.registry.add_transform(transform)
 
     def add_post_transform(self, transform):


sphinx/application.py
```diff
@@ -968,6 +968,29 @@ def add_transform(self, transform):
...
+           * - 100-299
+             - Preparation
+           * - 300-399
+             - early
+           * - 400-699
+             - main
+           * - 700-799
+             - Post processing. Deadline to modify text and referencing.
+           * - 800-899
+             - Collect referencing and referenced nodes. Domain processing.
```
tk0miya, line 987: Note: At present, domain processing is invoked after transform processing.
With this proposal, I'll move it to here.

tk0miya: `doctree-event` is also.

shimizukawa (approved):
I think this is useful information for other developers. 
If we have more detail information rather than 'early' and 'main', it would be perfect.

However, LGTM to current information. Thanks!

sphinx/application.py
```diff
@@ -968,7 +968,30 @@ def add_transform(self, transform):
...
+           * - 0-99
+             - Fix invalid nodes by docutils. Translate a doctree.
+           * - 100-299
+             - Preparation
+           * - 300-399
+             - early
+           * - 400-699
+             - main
+           * - 700-799
+             - Post processing. Deadline to modify text and referencing.
```
tk0miya, line 985: I noticed `SphinxSmartQuotes` should be invoked here. At present, it has `default_priority: 850`.

tk0miya:
@shimizukawa 'early' and 'main' are came from docutils' definition. I keep them as is. Do you have any idea?

shimizukawa:
I wrote the above comment because I felt that the explanation was written
more carefully than docutils (exclude early and main).
So, if there is no material to carefully write, I think that there is no
problem as it is.

If there is transformations belonging to early and main, how about writing
it as an example?

2018年4月8日(日) 23:41 Takeshi KOMIYA <notifications@github.com>:

tk0miya:
I checked transforms of docutils. It seems only `SmartQuotes` is related with Sphinx.
```
# docutils/transforms/components.py:
class Filter(Transform):
    default_priority = 780
```
```
# docutils/transforms/misc.py:
class CallBack(Transform):
    default_priority = 990

class Transitions(Transform):
    default_priority = 830
```
```
# docutils/transforms/parts.py:
class SectNum(Transform):
    default_priority = 710

class Contents(Transform):
    default_priority = 720
```
```
# docutils/transforms/peps.py:
class PEPZero(Transform):
    default_priority =760
```
```
# docutils/transforms/references.py:
class DanglingReferences(Transform):
    default_priority = 850

class Decorations(Transform):
    default_priority = 820

class ExposeInternals(Transform):
    default_priority = 840

class Messages(Transform):
    default_priority = 860

class FilterMessages(Transform):
    default_priority = 870

class TestMessages(Transform):
    default_priority = 880

class StripComments(Transform):
    default_priority = 740

class SmartQuotes(Transform):
    default_priority = 850
```
```
# docutils/transforms/writer_aux.py:
class Compound(Transform):
    default_priority = 910

class Admonitions(Transform):
    default_priority = 920
```

tk0miya:
>If there is transformations belonging to early and main, how about writing
>it as an example?

TBH, I don't have good idea and example for these range.

sphinx/application.py
tk0miya: I made #4853 for this.

tk0miya:
Okay, let's go with this.
Please update again if we'll find better description :-)

PR #4815, opened 2018-04-05
Refine adding CSS files

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Allows to add CSS files through config value: `html_css_files` and `epub_css_files`
- Renames `add_stylesheet()` to `add_css_file()`
- Support additional options for `<link>` tag (e.g. `media`, `integrity` and so on)
- Do not store additional CSS files into class variable

### Relates
- This is another version of #4595 
- #2442

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -1001,13 +1001,27 @@ class Sphinx(object):
             StandaloneHTMLBuilder.script_files.append(
                 posixpath.join('_static', filename))
 
-    def add_stylesheet(self, filename, alternate=False, title=None):
-        # type: (unicode, bool, unicode) -> None
+    def add_stylesheet(self, filename, **kwargs):
+        # type: (unicode, **unicode) -> None
         """Register a stylesheet to include in the HTML output.
 
         Add *filename* to the list of CSS files that the default HTML template
-        will include.  Like for :meth:`add_javascript`, the filename must be
-        relative to the HTML static path, or a full URI with scheme.
+        will include.  The filename must be relative to the HTML static path,
+        or a full URI with scheme.  The keyword arguments are also accepted for
+        attributes of ``<link>`` tag.
+
+        Example::
+
+            app.add_stylesheet('custom.css')
+            # => <link rel="stylesheet" href="_static/custom.css" type="text/css" />
+
+            app.add_stylesheet('print.css', media='print')
+            # => <link rel="stylesheet" href="_static/print.css"
+            #          type="text/css" media="print" />
+
+            app.add_stylesheet('fancy.css', rel='alternate stylesheet', title='fancy')
+            # => <link rel="alternate stylesheet" href="_static/fancy.css"
+            #          type="text/css" title="fancy" />
 
         .. versionadded:: 1.0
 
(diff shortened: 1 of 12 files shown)


sphinx/application.py
```diff
@@ -1001,13 +1003,27 @@ def add_javascript(self, filename):
...
         Add *filename* to the list of CSS files that the default HTML template
-        will include.  Like for :meth:`add_javascript`, the filename must be
-        relative to the HTML static path, or a full URI with scheme.
+        will include.  The filename must be relative to the HTML static path,
+        or a full URI with scheme.  The keyword arguments are also accepted for
+        attributes of ``<link>`` tag.
+
+        Example::
+
+            app.add_stylesheet('custom.css')
```
shimizukawa: It must be: `app.add_css_file('custom.css')` I think.

sphinx/application.py
```diff
@@ -1001,13 +1003,27 @@ def add_javascript(self, filename):
...
+            # => <link rel="stylesheet" href="_static/custom.css" type="text/css" />
+
+            app.add_stylesheet('print.css', media='print')
```
shimizukawa: same as above

sphinx/application.py
```diff
@@ -1001,13 +1003,27 @@ def add_javascript(self, filename):
...
+            # => <link rel="stylesheet" href="_static/print.css"
+            #          type="text/css" media="print" />
+
+            app.add_stylesheet('fancy.css', rel='alternate stylesheet', title='fancy')
```
shimizukawa: same as above

sphinx/themes/basic/layout.html
```diff
@@ -97,8 +97,8 @@ <h3>{{ _('Navigation') }}</h3>
     <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
     <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
     {%- for css in css_files %}
-      {%- if css|attr("rel") %}
-    <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
+      {%- if css|attr("filename") %}
+    {{ css_tag(css) }}
```
shimizukawa, line 101: 👍

sphinx/registry.py
```diff
@@ -65,6 +65,9 @@ def __init__(self):
         #: autodoc documenters; a dict of documenter name -> documenter class
         self.documenters = {}           # type: Dict[unicode, Type[Documenter]]
 
+        #: css_files; a list of tuple of filename and attributes
+        self.css_files = []             # type: List[Tuple[unicode, Dict[unicode, unicode]]]
```
shimizukawa, line 69: I think this implementation strongly binds to html related builders.
To be clarity, is the idea that registry has responsibility for setting irrespective of builder?

shimizukawa (commented):
I reviewed early.

sphinx/registry.py
tk0miya: Surely this is used for HTML builders. I know they are bundled, but not core components. So storing them onto registry is a bit strange.
But we already provide `app.add_stylesheet()` as a part of Sphinx API. So I believe this does not change the responsibilities of our modules.

tk0miya:

PR #4237, opened 2017-11-12
viewcode does not import when viewcode_import is False

AWhetter:
### Feature or Bugfix
- Bugfix and Feature

### Purpose
With the `viewcode` extension loaded, and `viewcode_import` set to `False`, `viewcode` will still attempt to load documented modules.

### Detail
- Fixes #4035 
- Added `viewcode_source_dirs` config variable to tell viewcode where to search for module source code (instead of sys.path).

### Relates
- #4035

PR #4816, opened 2018-04-05
Plugins can find source code for viewcode

### LINKED ISSUE #4035: viewcode does not abide by viewcode_import
### Problem
- With the `viewcode` extension loaded, and `viewcode_import` set to `False`, `viewcode` will still attempt to load documented modules. This is because it uses `sphinx.pycode.ModuleAnalyzer.for_module` (https://github.com/sphinx-doc/sphinx/blob/cb0220de1a0d8acd00292b898d2ef03b588179b0/sphinx/ext/viewcode.py#L66), which uses `sphinx.util.get_module_source()`, which attempts to import the module. 

```
sphinx-build source build
```
with the `viewcode` extension loaded, and `viewcode_import` set to `False`

#### Expected results
Ideally, there would be an option to `viewcode` that let's you specify source directories, and it would use these to find source code, falling back to `ModuleAnalyzer` only if `viewcode_import` is True.

You could also attempt to make `sphinx.util.get_module_source()` optionally attempt to get the source without loading the module. This would likely require using something like `pkgutil.get_loader()` but would add a dependency. Plus handing eggs could be difficult because the code exists in an egg, rather than on disk, ready to read. Overall the implementation of this method would be a lot more messy.

I don't have a sample project but a simple test case would be to define a module in the python path with:
```
raise RuntimeError()

class MyClass(object):
    """This is my class."""
```

and to document the class in an rst file. The sphinx config would use `viewcode` and set `viewcode_import` to `False`.

### Environment info
- OS: Linux
- Python version: 2.7.3
- Sphinx version: 1.6.3

AWhetter:
### Feature or Bugfix
- Feature

### Purpose
- Plugins can find and specify source code to the viewcode extension. Doing so would also skip viewcode having to import the module.

### Detail
- Fixes #4035
- Added the `viewcode-find-source` event to let plugins decide how to find source code.

Unfortunately the suggestion in https://github.com/sphinx-doc/sphinx/pull/4237#issuecomment-377916991 is not ideal because we still need the ModuleAnalyzer to parse the source code to get the tags out. Returning the source code from the event means that we would be bypassing the caches on ModuleAnalyzer because they are not used in `ModuleAnalyzer.for_string`.
A benefit to allowing the plugin to specify the tags means that plugins can also choose how to parse the source code. Whether they use `ModuleAnalyzer.for_file`, `ModuleAnalyzer.for_string`, or something totally different.

### Relates
- #4035


tests/test_ext_viewcode.py
```diff
@@ -60,3 +62,21 @@ def test_linkcode(app, status, warning):
...
+    def unload_spam(app, doctree):
+        sys.path.pop(0)
+        for module in list(sys.modules):
+            if module.startswith('spam.') or module == 'spam':
+                del sys.modules[module]
+
+    listeners = app.events.listeners['doctree-read']
+    viewcode_read = listeners.popitem()[1]
+    app.connect('doctree-read', unload_spam)
+    app.connect('doctree-read', viewcode_read)
```
tk0miya: I feel this code is very hacky. I can't understand what this code doing intuitively. How about adding new test project? Then we don't hack `sys.path` and event listeners.

tk0miya (changes requested):
LGTM with nits

tk0miya:
@shimizukawa could you review this please?

tests/test_ext_viewcode.py
```diff
@@ -60,3 +62,21 @@ def test_linkcode(app, status, warning):
...
+
+    test_viewcode(app, status, warning, include_short_names=False)
```
shimizukawa: Please DON'T call another test function.

shimizukawa (changes requested):

AWhetter:
This is ready for another review.

doc/ext/viewcode.rst
```diff
@@ -62,3 +65,17 @@ There is an additional config value:
       Some reader's rendering result are corrupted and
       `epubcheck <https://github.com/IDPF/epubcheck>`_'s score
       becomes worse even if the reader supports.
+
+.. event:: viewcode-find-source (app, modname)
+
+   .. versionadded:: 1.8
+
+   Find the source code for a module.
+   The event returns a tuple of the source code itself,
```
tk0miya: I feel "A event handler for this event should return" is better.

tk0miya (approved):
LGTM with nits :-)

doc/ext/viewcode.rst
AWhetter: Good point. I've made this change.

shimizukawa (approved):
LGTM.
Sorry to keep you waiting.

tk0miya:
Merged.
Thank you for you work!

PR #4827, opened 2018-04-10
Fix #4784: latex_show_urls assigns incorrect footnote numbers

### LINKED ISSUE #4784: URL footnotes don't start at 1 in latexpdf build
### Problem
I'm building a multi-chapter PDF document in Sphinx with footnotes that reference URLs outside of the document. 

- The first footnote in the document does not start at 1, it starts at some large number (32 in my case)
- Footnote numbers are skipped when assigning numbers in subsequent chapters. That is, if chapter 1 includes footnotes 32-40, the first footnote in chapter 2 is not 41, it's some larger value. 

Originally I found this problem when I used InterSphinx and set `latex_show_urls = 'footnote'` in conf.py but later I found I can reproduce it simply by using a footnote that references an rst_prolog substitution that contains a URL. 

More details are here: https://groups.google.com/forum/#!topic/sphinx-users/e93WiwjrQLg 
(This thread also includes a workaround to override footnote numbering that seems to be working for me.)

The attached example project reproduces the problem when built with `make latexpdf`.

I saw slightly different behavior when I built the project on Sphinx 1.6.6 compared to building on 1.7.1. In 1.6.6 I believe it did not skip additional footnote numbers between chapters, it only started at a large number (also the first footnote was around number 43 instead of 32). 

#### Error logs / results

see sample project
 
#### Expected results

Footnotes should start at 1 and be numbered sequentially throughout the document without skipping numbers.

(It would be nice if we could easily configure whether or not to restart footnote numbering each chapter, but that's a separate issue.)

Example project: 
[footnote_debug.zip](https://github.com/sphinx-doc/sphinx/files/1848430/footnote_debug.zip)

### Environment info
- OS: MacOS 10.13.3
- Python version: 2.7
- Sphinx version: 1.6.6 and 1.7.1

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4784

### DIFF
--- a/sphinx/transforms/references.py
+++ b/sphinx/transforms/references.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+"""
+    sphinx.transforms.references
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+    Docutils transforms used by Sphinx.
+
+    :copyright: Copyright 2007-2018 by the Sphinx team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+from docutils import nodes
+from docutils.transforms.references import Substitutions
+
+from sphinx.transforms import SphinxTransform
+
+
+class SubstitutionDefinitionsRemover(SphinxTransform):
+    """Remove ``substitution_definition node from doctrees.
+
+    .. note:: In Sphinx-1.7, this transform is only used in LaTeX builder.
+    """
+
+    # should be invoked after Substitutions process
+    default_priority = Substitutions.default_priority + 1
+
+    def apply(self):
+        # type: () -> None
+        for node in self.document.traverse(nodes.substitution_definition):
+            node.parent.remove(node)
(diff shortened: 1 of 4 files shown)


sphinx/transforms/references.py
```diff
@@ -0,0 +1,30 @@
...
+from docutils import nodes
+from docutils.transforms.references import Substitutions
+
+from sphinx.transforms import SphinxTransform
+
+
+class SubstitutionDefinitionsRemover(SphinxTransform):
+    """Remove ``substitution_definition node from doctrees.
+
+    .. note:: In Sphinx-1.7, this transform is only used in LaTeX builder.
```
tk0miya, line 21: I added this as local transform for LaTeX builder. But `substituion_definition` node is not used in all of builders. So I'll move this to normal transform which is invoked in reading phase, in master branch.

shimizukawa (approved):
LGTM

tk0miya:
@shimizukawa Thank you for reviewing!

PR #2843, opened 2016-08-10
Add :start-at: and :end-at: parameters for literalinclude

rthardin:
This feature is discussed as part of issue #625.

Adds `:start-at:` and `:end-at:` parameters to the LiteralInclude directive. These
behave like `:start-after:` and `:end-before:`, but are inclusive.

Issue #625, opened 2015-01-02
New :start-at:, :end-at: and :block: parameters for literalinclude


shimizukawa:
(edit: added a second version of the patch, but I could not find how to remove the first one. In the new version :braceblock: was renamed to :block:, and now the opening and closing delimiters can be changed using :block-begin: and :block-end:, although they still default to braces)

Hi,

I really like Sphinx. I am attaching a patch that adds three parameters to literalinclude that I have found useful: :start-at: and :end-at: work just like :start-after: and :end-before:, but they include the matching line. :block: allows inclusion of any delimiter-enclosed block of a file (a poor-man's generic version of pyobject, if you will :). The included block starts at the first line that matches the string given as the value to :block:. It then looks for the first opening brace ({), and includes everything until the line that contains the matching closing brace (}). For example, given the file:

```
sub a {
  something
  { something else }
}
sub b {
  some other things
}
```

Specifying

```
  :block: sub a
```

would result in the inclusion of only the "sub a" block.

The opening and closing brace for :block: are { and } by default, but they can be changed by specifying them using :block-begin: and :block-end:. This makes it possible to extract, for example, do .. end blocks from Ruby files.

I hope you find it useful, it would be nice to see them (or similar capabilities) in future versions of Sphinx.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/625
- Originally reported by: Diego Zamboni
- Originally created at: 2011-02-02T08:47:25.569

shimizukawa:
_From Kevin Horn on 2011-02-04 04:41:09+00:00_

I think this is very cool.

Might I also suggest that project-wide config file settings for :block-begin: and :block-end: would be quite useful

shimizukawa:
_From Georg Brandl on 2014-10-18 07:32:19+00:00_

Removing version: 1.0.7 (automated comment)

rthardin:
@shimizukawa, is the fix for this issue alive somewhere? I have a use case where `:start-at:` would be very useful - wanted to check if there's a possibility of picking this back up. I also don't mind taking a stab at it myself if there's not work currently being done on it. Thank you very much!

Edit: Went ahead and opened a PR for this since `code.py` is very extensible, and it was quick to add. The PR only includes `:start-at:` and `:end-at:`, and does not include the `block` behavior described by Diego Zamboni.

shimizukawa:
Now I close this issue because #2843 provides `:start-at:` and `:end-at:` options.
[...]

PR #4832, opened 2018-04-11
Add missing literalinclude options documentation

ndevenish:
Subject: The `start-at` and `end-at` options `literalinclude` options weren't documented

### Feature or Bugfix
- Bugfix

### Purpose
- The `start-at` and `end-at` options were added in v1.5a1, but there was no documentation for these options on the directives (or old code) pages.

### Detail
- Only found that they were already implemented whilst investigating the code to see how hard the feature would be to implement myself; the file I wanted to include had only blank lines before the line I wanted to select.
- These *were* mentioned in the documentation, but only under the 1.5 release notes

### Relates
- Issue #625
- Pull request https://github.com/sphinx-doc/sphinx/pull/2843


tk0miya:
Thanks!

PR #4834, opened 2018-04-12
Ensure set object descriptions are reproducible.

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that sphinx could generate output that is not reproducible.

In particular, the rendering of `set` objects in default arguments
and elsewhere is currently non-determinstic. For example:

    class A_Class(object):
        a_set = {'a', 'b', 'c'}

Might be rendered as any of:

     {'a', 'b', 'c'}
     {'a', 'c', 'b'}
     {'b', 'a', 'c'}
     {'b', 'c', 'a'}
     {'c', 'a', 'b'}
     {'c', 'b', 'a'}

Patch attached that sorts the contents of sets whilst rendering.
This is parallel to the `dict` key sorting.

This was originally filed in Debian as #895553 [1].

 [0] https://reproducible-builds.org/
 [1] https://bugs.debian.org/895553

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -244,7 +244,9 @@ def object_description(object):
         except TypeError:
             pass  # Cannot sort dict keys, fall back to generic repr
         else:
-            items = ("%r: %r" % (key, object[key]) for key in sorted_keys)
+            items = ("%s: %s" %
+                     (object_description(key), object_description(object[key]))
+                     for key in sorted_keys)
             return "{%s}" % ", ".join(items)
     try:
         s = repr(object)
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -248,6 +248,15 @@ def object_description(object):
                      (object_description(key), object_description(object[key]))
                      for key in sorted_keys)
             return "{%s}" % ", ".join(items)
+    if isinstance(object, set):
+        try:
+            sorted_values = sorted(object)
+        except TypeError:
+            pass  # Cannot sort set values, fall back to generic repr
+        else:
+            template = "{%s}" if PY3 else "set([%s])"
+            return template % ", ".join(object_description(x)
+                                        for x in sorted_values)
     try:
         s = repr(object)
     except Exception:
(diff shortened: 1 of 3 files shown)


sphinx/util/inspect.py
```diff
@@ -246,6 +246,10 @@ def object_description(object):
         else:
             items = ("%r: %r" % (key, object[key]) for key in sorted_keys)
```
tk0miya: I think values should be also processed for dicts contains set as a value.

tk0miya (changes requested):

sphinx/util/inspect.py
mitya57: How about calling `object_description` recursively? And below too, replace`repr(x)` with `object_description(x)`.

tk0miya: Yes, `object_description(object[key])` is better to me.

mitya57: Ok. This was my fault when I added the dict processing code, but I hope @lamby will fix it :)

lamby: Pushed, including the call to `object_description` in the dict handling too :)

sphinx/util/inspect.py
```diff
@@ -244,8 +244,15 @@ def object_description(object):
...
-            items = ("%r: %r" % (key, object[key]) for key in sorted_keys)
+            items = ("%s: %s" %
+                     (object_description(key), object_description(object[key]))
+                     for key in sorted_keys)
             return "{%s}" % ", ".join(items)
+    if isinstance(object, set):
+        # Sort set contents
+        template = "{%s}" if PY3 else "set([%s])"
+        return template % ", ".join(object_description(x)
+                                    for x in sorted(object))
```
tk0miya: This will be failed if members are not sortable:
```
>>> sorted({1, 2, 'str'})
TypeError: '<' not supported between instances of 'str' and 'int'
```
```
>>> sorted({1, 2, object()})
TypeError: '<' not supported between instances of 'object' and 'int'
```

mitya57: So a `try…except` is needed, like in the dict case.

lamby: Good shout. Have updated with a testcase.

tk0miya (approved):
LGTM! I'll merge this soon.

lamby:
Thanks all!

Issue #3271, opened 2016-12-21
Allow `Class{T}` as an alternative syntax for `Class<T>` references


firegurafiku:
Looking through [C++ domain documentation][doc] I noticed the following highlighted note:

> **Note on References with Templates Parameters/Arguments**  
> Sphinx’s syntax to give references a custom title can interfere with linking to template classes, if nothing follows the closing angle bracket, i.e. if the link looks like this: ``:cpp:class:`MyClass<int>` ``. This is interpreted as a link to `int` with a title of `MyClass`. In this case, please escape the opening angle bracket with a backslash, like this: ``:cpp:class:`MyClass\<int>` ``.

Unfortunately, C++ is full of templates, so having to escape `<` (or, at least, worry about that) sounds like a bad news.

How about choosing different type of brackets to get rid of this issue completely? For example, curly braces are never used in (sane) function and type signatures, so it would be safe to write the above reference as ``:cpp:class:`MyClass{int}` ``. The similar approach is used by Microsoft in their C# XML comments where reference to `MyClass<T1, T2>` may be written either as `MyClass&lt;,&gt;` or `MyClass{,}`.

Is something like this possible with Sphinx?

[doc]: http://www.sphinx-doc.org/en/stable/domains.html

jakobandersen:
It is indeed annoying to escape the angle brackets but I'm not sure if the use of curly brackets is unambiguous when combined with the template introduction syntax form the Concepts TS.
The original problem is because angle brackets are used to specify a different link text, so another solution could be to introduce "raw" role arguments. E.g., `` :cpp:class:`MyClass\<int>` `` and something like `` :cpp:class:R`MyClass<int>` ``, where you in raw mode can not provide your own link text.
This is however a change in the general Sphinx machinery.

other-mickk:
@jakobandersen Some food for thought: something like

```rst
:cpp:expr:`MyClass<int>`
```

currently works surprisingly well. However as you would expect it does style like an expression, not a `cpp:any` reference.

jakobandersen:
Ah, right, that essentially acts as a "raw text" thing. Regarding styling, it should be easy to add a `:cpp:texpr:`/`:cpp:text-expr:` role for which the only difference is that the CSS classes emitted are different.

other-mickk:
@jakobandersen after sketching a quick PoC I believe I can tackle that, do you want me to open a PR here or on your fork?

[...]

PR #4836, opened 2018-04-13
C++, add refs role

other-mickk:
ping @jakobandersen 

The tagline of this PR is to make the C++ domain parser more available since it's so great. In particular this adds a `cpp:refs` role to insert cross-references with normal text styling---acting as a sibling to `cpp:expr` which already does the same, but with inline code styling. As discussed this mitigates, but arguably does not entirely solve #3271. Breakdown of the contents:

- implement `cpp:refs`
- update docs to help the user choose the right cross-reference role for the job
- add testcase for ensuring that the cross-references generated by e.g. `any` and `cpp:any` are consistent with those generated by `cpp:refs` and `cpp:expr`---since the role objects associated with the latters don't inherit from `XRefRole`, we want to be told when there is a change (partial implementation, more on this below)

see the effects for yourself:

- [demo of the current state of affairs](http://mickk-on-cpp.github.io/sphinx-demo/2018-04/alabaster/master/index.html), for reference
- [demo of the PR](http://mickk-on-cpp.github.io/sphinx-demo/2018-04/alabaster/inline/index.html)
- [generated docs](http://mickk-on-cpp.github.io/sphinx-demo/2018-04/docs/usage/restructuredtext/domains.html#cross-referencing)

This makes the changes non-intrusive and non-breaking, but I dislike the inconsistency between the various kind of references. I have some work to ensure [looks consistency](http://mickk-on-cpp.github.io/sphinx-demo/2018-04/alabaster/code/index.html), but it's not done yet (the generated layout for the `cpp:expr` text ends up wonky). I thought I'd submit this first and ask for consensus before doing more work.

I'm also open to suggestions for the name of the role.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -61,6 +61,8 @@ Features added
   fontsize in code-blocks (refs: #4793)
 * Add :confval:`html_css_files` and :confval:`epub_css_files` for adding CSS
   files from configuration
+* C++, add a ``cpp:refs`` role as a sibling to ``cpp:expr`` for references
+  inside normal text
 
 Bugs fixed
 ----------
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -724,8 +724,8 @@ visibility statement (``public``, ``private`` or ``protected``).
 
          **Valid Expressions**
 
-         - :cpp:expr:`*r`, when :cpp:expr:`r` is dereferenceable.
-         - :cpp:expr:`++r`, with return type :cpp:expr:`It&`, when :cpp:expr:`r` is incrementable.
+         - :cpp:expr:`*r`, when :cpp:any:`r` is dereferenceable.
+         - :cpp:expr:`++r`, with return type :cpp:refs:`It&`, when :cpp:any:`r` is incrementable.
 
    This will render as follows:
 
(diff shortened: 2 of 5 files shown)


jakobandersen:
Thanks for your work, I have committed a simplified version on the master branch (future version 1.8). The role is called ``cpp:texpr`` to emphasize that it is ``cpp:expr`` but with text styling instead. It's then up to the users to decide when to use which.

benjaoming:
@jakobandersen just flagging these two direct commit to master without a PR - you're missing out on CI checks and community feedback :)

other-mickk:
@jakobandersen Thanks!

Here's my follow up on looks consistency:

- [the state of affairs on master](http://mickk-on-cpp.github.io/sphinx-demo/2018-06/alabaster/master/)
- [my proof of concept](http://mickk-on-cpp.github.io/sphinx-demo/2018-06/alabaster/poc/)

A quick breakdown of why I think it would be an improvement:

- references inside `:texpr:` look like overall `:any:` etc. references, and carry the xref class (while the remaining text doesn't)
- `:expr:` has the same grey background as with code, inside references carry the xref class (while the remaining code doesn't)

That is, the 'smart' roles look more similar to their supposed respective counterparts. (I did not try other writers.) On the other hand, here are some reservations:

- ultimately I don't know if it matters to put e.g. the xref class on the right spot; the stock CSS rules don't care about it though other themes or user-specified styles might
- the references inside `:expr:` end up with the colour of external links and I don't know how to improve on that from just the cpp domain side (but this is the same as on master)

As a final note, the code that generated the PoC is a dirty dynamic hack. If I were to make a proper PR, my intention would be to parametrise the whole of `describe_signature` with a new styler parameter so that [the code here](https://github.com/sphinx-doc/sphinx/blob/1c3eadcf61ab2725801218b207640bc7634b0f8f/sphinx/domains/cpp.py#L1312,L1336) will use caller-supplied node functions. This would be a bit of churn but I think it's worth it. Let me know what everybody thinks.

edit: I missed that alabaster is not the default theme these days, as usual I can generate more themes on demand though:

- [sphinx_rtd_theme on master](http://mickk-on-cpp.github.io/sphinx-demo/2018-06/sphinx_rtd_theme/master/)
- [sphinx_rtd_theme with the poc](http://mickk-on-cpp.github.io/sphinx-demo/2018-06/sphinx_rtd_theme/poc/)

which has its own flaws, e.g. plain-styled angle brackets next to code-styled references in `:texpr:`

PR #4838, opened 2018-04-14
Sort .hhp file list

bmwiedemann:
Without this change, the pgadmin3 openSUSE package differed for every build
(happens in a disposable VM) because pgadmin3.hhp contained entries
in indeterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this matters.

https://github.com/bmwiedemann/theunreproduciblepackage/tree/master/readdir also has some details on this topic.

### DIFF
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -247,6 +247,8 @@ class HTMLHelpBuilder(StandaloneHTMLBuilder):
                 outdir += os.sep
             olen = len(outdir)
             for root, dirs, files in os.walk(outdir):
+                dirs.sort()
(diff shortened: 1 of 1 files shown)


tk0miya:
+1; reasonable.

It seems this only changes the output of .hhp file. I think it is stable change. How about rebasing this to 1.7 branch?
Then you can use this on next stable release: 1.7.2.

Note: The error on Travis CI was related with new mypy release. I'm trying to fix it in #4839. So it will be fixed soon.

bmwiedemann:
rebased it to 1.7 branch

tk0miya:
Merged.

PR #4840, opened 2018-04-14
Remove substitution_definition nodes from doctree on reading phase (refs: #4827)

tk0miya:
### Feature or Bugfix
- Refactor

### Purpose
- To make doctree simple, remove all `substitution_defitinion` nodes from doctree on reading phase.
- They are applied to all `substituion_reference` before the removal. So they are not used on latter phase.


sphinx/builders/latex.py
```diff
@@ -217,7 +216,6 @@ def apply_transforms(self, doctree):
         # type: (nodes.document) -> None
         transformer = SphinxTransformer(doctree)
         transformer.set_environment(self.env)
-        transformer.add_transforms([SubstitutionDefinitionsRemover])
         transformer.apply_transforms()
```
tk0miya: I left `SpinxTransformer` as is because I will add another Transform module here later.

PR #4845, opened 2018-04-14
Deprecate highlightlang directive

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- highlightlang directive is not documented, and marked as old in v0.2 (see f2ff0419bdeb5a144fd6f74e000ad985447339d3)
- Now it is called as `highlight` directive. It is no longer needed.
- It has been longly alive. So I'd like to remove it in 4.0.


shimizukawa (approved):
LGTM!

tk0miya:
:-)

Issue #4777, opened 2018-03-24
Support for Python coroutines


ovekaaven:
Sphinx does not currently seem to support Python coroutines (async functions or methods).

I've written a patch that fixes this, although I'm not sure if you'd prefer a different design. (If you do think it looks good, I can make a pull request.)

- the pycode/parser.py change fixes 'bysource' sorting of coroutines
- the domains/python.py (and doc/domains.rst) change adds an 'async' option, which can be used with py:function, py:method, etc.
- the rest adds autodoc support

In the output, "async" will be prepended to the signature prefix (i.e. where thinds like "static" and "classmethod" show up) if a function or method is a coroutine.

```
diff --git a/doc/domains.rst b/doc/domains.rst
index 5bed02cf4..975f1f7ad 100644
--- a/doc/domains.rst
+++ b/doc/domains.rst
@@ -299,6 +299,17 @@ you can use brackets to specify the optional parts:
 It is customary to put the opening bracket before the comma.
 
 
+.. _coroutines:
+
+Coroutines
+~~~~~~~~~~
+
+If a function or method is a coroutine, you can add the ``:async:`` option::
+
+   .. py:function:: compile(source : string, filename, symbol='file') -> ast object
+      :async:
+
+
 .. _info-field-lists:
 
 Info field lists
diff --git a/sphinx/domains/python.py b/sphinx/domains/python.py
index b0900b385..e9b65801c 100644
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -170,6 +170,7 @@ class PyObject(ObjectDescription):
         'noindex': directives.flag,
         'module': directives.unchanged,
         'annotation': directives.unchanged,
+        'async': directives.flag,
     }
 
     doc_field_types = [
@@ -253,6 +254,8 @@ class PyObject(ObjectDescription):
         signode['fullname'] = fullname
 
         sig_prefix = self.get_signature_prefix(sig)
+        if 'async' in self.options:
+            sig_prefix = 'async ' + sig_prefix
         if sig_prefix:
             signode += addnodes.desc_annotation(sig_prefix, sig_prefix)
 
diff --git a/sphinx/ext/autodoc/__init__.py b/sphinx/ext/autodoc/__init__.py
index 89ca1f44c..3ced7ab70 100644
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
[...]

PR #4847, opened 2018-04-14
Support for Python coroutines

ovekaaven:
Subject: Add support for Python coroutines

This feature was proposed in #4777.

I've split the patch I proposed into 3 different changesets, and simplified the autodoc part by putting the async check into the common Documenter.add_directive_header instead of its subclasses, which may or may not be to your liking. I can go back to overriding add_directive_header in the subclasses if you prefer that, but I doubt it makes much practical difference.

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -290,6 +290,16 @@ argument support), you can use brackets to specify the optional parts:
 
 It is customary to put the opening bracket before the comma.
 
+.. _coroutines:
+
+Coroutines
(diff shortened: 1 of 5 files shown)


tk0miya (changes requested):
Sorry for very late reviewing.

LGTM with nits.
Could you add testcase for coroutines please?

tk0miya:
ping. The feature freeze for Sphinx-1.8 comes soon (see #4986).
So please let me know status of this PR.

I really want to merge this into next release.

tk0miya:
Fixed by #6295 instead.

PR #4856, opened 2018-04-16
Fix translation for zh_CN and zh_TW.

lepture:
Subject: Correction translation for zh_CN and zh_TW

### Feature or Bugfix

- Feature
- Bugfix

### Purpose

As a native Chinese speaker, the previous translations are weird.

1. "Permalink to XXX", it is correct in zh_TW, but incorrect in zh_CN
2. "Go", this "Go" is actually "search" in the context. Translating it into "搜" makes more sense.

### DIFF
--- a/sphinx/locale/zh_CN/LC_MESSAGES/sphinx.po
+++ b/sphinx/locale/zh_CN/LC_MESSAGES/sphinx.po
@@ -425,7 +425,7 @@ msgstr "行规范 %r：未能从包含文件 %r 中拉取行"
 
 #: sphinx/directives/other.py:157
 msgid "Section author: "
-msgstr "节作者："
+msgstr "章节作者："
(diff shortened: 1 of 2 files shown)


tk0miya:
We are using transifex for translation platform.
So could you join to our transifex group and update them please?
https://www.transifex.com/sphinx-doc/

lepture:
@tk0miya I've send the request.

tk0miya:
Welcome to our project!

tk0miya:
Note: I merged this manually into 1.7 branch to apply these translations to next stable release.
But it only effects to 1.7.x series. So please also update them on transifex. Then they are also applied in next major release.

lepture:
I’ve submit them in transifex, but they are just suggestions.

Issue #1618, opened 2015-01-03
HTML Search results aren't reader friendly


shimizukawa:
The HTML built-in search is very useful, especially for offline help, but the results content isn't reader friendly.

For example, I get such content:

```
#!python

*************** Report *************** .. important:: For method, 
```

which isn't understandable by the average user.

As searchtools.js use files in _sources, which are a copy of the Rest sources, this is happening.
But the search only need text files I think, not real sources files.

By replacing the content of _sources with the output of sphinx to text, and when setting 

```
#!python

text_sectionchars = '       '
```

, I get a better result:

```
#!python

Report Important: For method,
```

It is a lot better, even if the \* of bold are still visible.

It would be great if this rendering to text is automated when doing the HTML rendering.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1618
- Originally reported by: [C_PYB](https://bitbucket.org/C_PYB)
- Originally created at: 2014-11-06T18:07:35.870

shimizukawa:
_From Andrea Cassioli on 2014-12-17 10:08:54+00:00_

I totally agree, using the rest text as output is misleading and not nice at all!

KacerCZ:
+1 for this feature
I was surprised that search is displaying source files in search result. 
Display of rendered HTML content is something what users expect. For now using workaround with replacement of source files with text files.

TimKam:
The search results would be much better if one:
- removed all markup (i.e. headings, images, bold and italic print) when building the .txt files.
- built the .txt files as part of the normal html build and places them in the `_sources` directory of the build.

Is there any downside in doing this by default?

lehmannro:
I'm pretty sure lots of machinery in Sphinx assumes there's only ever one build happening per runtime, so doing `text` as part of `html` might be both, wasteful in terms of build time and complex to implement.

TimKam:
Thanks for the info. Regarding the build of the txt file, a small customization of the makefile can do the trick.

Still, it would be nice if one could produce more "search-friendly" .txts. What do you think?

Currently, we run some custom script to remove the remaining markup.
I suppose this might be a relatively common problem Sphinx users have.

TimKam:
In case this is a relevant issue for somebody else:

I wrote a - so far very basic - extension that fixes this issue and builds the search result snippets without markup.
GitHub: https://github.com/TimKam/sphinx-pretty-searchresults
[...]

PR #4857, opened 2018-04-17
Improve search summary by filtering reStructuredText

timhoffm:
### Purpose
The search summary contains excerpts of the .rst sources. ReST code elements tend to clutter the summary, making it hard to find useful information within.

This PR tries to improve readability be filtering out some ReST code elements. In particular, this:

- removes link targets
- removes some directives (currently used only for one single-line directives as we don't have a full directive block parser)
- removes section markers

This is just a heuristic filter for removing some text elements that clutter the search result summary text without providing helpful information. The motivation is a low-effort way for improving the search result summary. This is not trying to be a serious reStructuredText parser in any way.

### Example

Here are two screenshots showing the effect of the filter.

before:
grafik

with filter:
grafik

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js_t
+++ b/sphinx/themes/basic/static/searchtools.js_t
@@ -458,6 +458,68 @@ var Search = {
     return results;
   },
 
+    /*
+   * helper function to remove some reStructuredText elements
+   * from the text for better readability of the plain text
+   * source. In particular, this:
+   *
+   * - removes link targets
+   * - removes some directives (use only for single-line
+   *   directives as we don't have a full directive block parser)
+   * - removes section markers
+   *
+   * This is just a heuristic filter for removing some text elements
+   * that clutter the search result summary text without providing
+   * helpful information. The motivation is a low-effort way
+   * for improving the search result summary. This is not trying to
+   * be a serious reStructuredText parser in any way.
+   */
+  filterRestructuredText : function(text) {
+    var reLinkTarget = /^\.\.\ _[\w\.]+:$/;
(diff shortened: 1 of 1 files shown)


timhoffm:
I don't understand codecov. This PR does not touch any python code, yet it claims a change in `sphinx/__init__.py`.

tk0miya:
Sorry for late response.

-1: As a workaround, we already have prettysearchresults extension.
https://pypi.org/project/sphinxprettysearchresults/

As proper way, we have to improve Sphinx-core itself to build both HTML and text at once.

timhoffm:
Closing as duplicate. There are several methods in discussion: #1618

It's really unfortunate that the default gives bad output and one has to use an extension to fix it. Is there anything one can do to promote a sphinx-internal solution?

tk0miya:
Absolutely. We have to improve sphinx-core to provide good behavior by default.

>Is there anything one can do to promote a sphinx-internal solution?

I think `sphinx-build` should support build multiple document at same. It should be nice if we can invoke both HTML and Text builder at once.

timhoffm:
Unfortunately building the complete text docs takes a lot of extra time. Therefore a small javascript filter as proposed above would still be the preferred option for my use case. Is there any way to extend an official sphinx build by injecting javascript, or would I have to maintain my own sphinx branch to achieve that?

tk0miya:
Sorry for late. I agreed with you. As you said, improving build process is hard work. So this workaround is surely useful. Now I'm talking with @TimKam same theme in #1618.

I'm not good at JavaScript. AFAIK, other maintainers are also. So maintainers for the search script is needed for us.

PR #4861, opened 2018-04-18
Add support for list of abbreviations to LaTeX builder

sephalon:
Subject: Add support for list of abbreviations to LaTeX builder

### Feature or Bugfix
- Feature

### Purpose
Adds support for generating a list of abbreviations as well as cross-references in the LaTeX builder. All the heavy-lifting is done by the LaTeX ``glossaries`` package, not internally in Sphinx.

As index creation is already been done by LaTeX packages, I think it is only logical to use existing, well-tested tooling for creating acronyms and not reinvent the wheel in Sphinx. Therefore I'd also like to propose handling glossaries in the same way.

### DIFF
--- a/sphinx/texinputs/latexmkrc_t
+++ b/sphinx/texinputs/latexmkrc_t
@@ -26,7 +26,8 @@ sub xindy {
 {% else -%}
 $makeindex = 'makeindex -s python.ist %O -o %D %S';
 {% endif -%}
-add_cus_dep( "glo", "gls", 0, "makeglo" );
-sub makeglo {
- return system( "makeindex -s gglo.ist -o '$_[0].gls' '$_[0].glo'" );
+add_cus_dep( "glo", "gls", 0, "run_makeglossaries" );
+add_cus_dep( "acn", "acr", 0, "run_makeglossaries" );
+sub run_makeglossaries {
+  system "makeglossaries '$_[0]'";
 }
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -147,6 +147,7 @@ DEFAULT_SETTINGS = {
                         '\\usepackage{hypcap}% it must be loaded after hyperref.\n'
                         '% Set up styles of URL: it should be placed after hyperref.\n'
                         '\\urlstyle{same}'),
+    'glossaries':      '\\usepackage[acronym,toc]{glossaries}',
     'contentsname':    '',
     'preamble':        '',
     'title':           '',
@@ -154,11 +155,13 @@ DEFAULT_SETTINGS = {
     'author':          '',
     'releasename':     '',
     'makeindex':       '\\makeindex',
+    'makeglossaries':  '\\makeglossaries',
     'shorthandoff':    '',
     'maketitle':       '\\sphinxmaketitle',
     'tableofcontents': '\\sphinxtableofcontents',
     'atendofbody':     '',
     'printindex':      '\\printindex',
+    'printacronyms':   '\\printacronyms',
     'transition':      '\n\n\\bigskip\\hrule\\bigskip\n\n',
     'figure_align':    'htbp',
     'tocdepth':        '',
@@ -270,6 +273,7 @@ class Table:
         # type: (nodes.Element) -> None
         self.header = []                        # type: List[str]
         self.body = []                          # type: List[str]
+
         self.align = node.get('align')
         self.colcount = 0
         self.colspec = None                     # type: str
@@ -457,6 +461,7 @@ class LaTeXTranslator(SphinxTranslator):
         # type: (nodes.document, LaTeXBuilder) -> None
         super().__init__(document, builder)
         self.body = []  # type: List[str]
+        self.abbreviations = {}  # type: Dict[unicode, unicode]
 
         # flags
         self.in_title = 0
@@ -634,7 +639,6 @@ class LaTeXTranslator(SphinxTranslator):
         self.footnote_restricted = None     # type: nodes.Element
         self.pending_footnotes = []         # type: List[nodes.footnote_reference]
         self.curfilestack = []              # type: List[str]
-        self.handled_abbrs = set()          # type: Set[str]
 
     def pushbody(self, newbody):
         # type: (List[str]) -> None
@@ -659,7 +663,8 @@ class LaTeXTranslator(SphinxTranslator):
         # type: () -> str
         self.elements.update({
             'body': ''.join(self.body),
-            'indices': self.generate_indices()
+            'indices': self.generate_indices(),
+            'abbreviations': self.generate_abbreviations()
         })
         return self.render('latex.tex_t', self.elements)
 
@@ -745,6 +750,15 @@ class LaTeXTranslator(SphinxTranslator):
                     generate(content, collapsed)
 
         return ''.join(ret)
+
+    def generate_abbreviations(self):
+        # type: (Builder) -> unicode
+        gen_abbrs = ''
+        for abbr, explanation in self.abbreviations.items():
+            gen_abbrs += '\\newacronym{%s}{%s}{%s}\n' % (
+                    abbr, abbr, self.encode(explanation)
+            )
+        return gen_abbrs
 
     def render(self, template_name, variables):
         # type: (str, Dict) -> str
@@ -1975,13 +1989,12 @@ class LaTeXTranslator(SphinxTranslator):
     def visit_abbreviation(self, node):
         # type: (nodes.Element) -> None
         abbr = node.astext()
-        self.body.append(r'\sphinxstyleabbreviation{')
-        # spell out the explanation once
-        if node.hasattr('explanation') and abbr not in self.handled_abbrs:
-            self.context.append('} (%s)' % self.encode(node['explanation']))
-            self.handled_abbrs.add(abbr)
+        if node.hasattr('explanation'):
+            self.body.append(r'\gls{')
+            self.abbreviations[abbr] = node['explanation']
         else:
-            self.context.append('}')
+            self.body.append('{')
+        self.context.append('}')
 
     def depart_abbreviation(self, node):
         # type: (nodes.Element) -> None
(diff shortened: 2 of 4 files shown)


sphinx/writers/latex.py
```diff
@@ -2206,13 +2210,13 @@ def depart_literal_strong(self, node):
...
         abbr = node.astext()
-        self.body.append(r'\sphinxstyleabbreviation{')
-        # spell out the explanation once
-        if node.hasattr('explanation') and abbr not in self.handled_abbrs:
-            self.context.append('} (%s)' % self.encode(node['explanation']))
-            self.handled_abbrs.add(abbr)
+        if node.hasattr('explanation'):
+            self.body.append(r'\gls{')
+            self.abbr.add('\\newacronym{' + abbr + '}{' + abbr + '}{' +
+                    node['explanation'] + '}\n')
```
jfbu: perhaps self.encode(node['explanation'])?

sphinx/writers/latex.py
```diff
@@ -2206,13 +2210,13 @@ def depart_literal_strong(self, node):
...
         else:
-            self.context.append('}')
+            self.body.append(r'{')
```
jfbu: why ``r``?

sphinx/writers/latex.py
jfbu: in case input has duplicated possibly diverging explanations, the original code stored only the first explanation ; here it seems ``\newacronym` will be inserted in LaTeX multiple times in such cases. This does not cause problem?

sphinx/texinputs/latexmkrc_t
```diff
@@ -11,7 +11,13 @@ $pdflatex = 'xelatex ' . $ENV{'LATEXOPTS'} . ' %O %S';
...
- return system( "makeindex -s gglo.ist -o '$_[0].gls' '$_[0].glo'" );
+add_cus_dep( "glo", "gls", 0, "run_makeglossaries" );
+add_cus_dep( "acn", "acr", 0, "run_makeglossaries" );
+sub run_makeglossaries {
+  if ( $silent ) {
+    system "makeglossaries -q '$_[0]'";
+  }
+  else {
+    system "makeglossaries '$_[0]'";
+  };
```
jfbu: I guess you do that because silent option of latexmk does not suffice ? (untested)

sphinx/writers/latex.py
sephalon: Fixed.

sphinx/writers/latex.py
sephalon: Good point, the glossaries packages complains if there are multiple abbreviations with diverging explanations. I deferred the writing of the abbreviation definitions after all abbreviations have been collected, which ensures that the last definition always "wins".

sphinx/texinputs/latexmkrc_t
sephalon: Removed the silent option, seems unnecessary.

sephalon:
I've incorporated comments from @jfbu and added some documentation as well.

doc/config.rst
```diff
@@ -2014,6 +2014,11 @@ These options influence LaTeX output. See further :doc:`latex`.
 
         .. versionadded:: 1.5
            Previously this was done from inside :file:`sphinx.sty`.
+     ``'glossaries'``
+        "glossaries" package inclusion, defaults to
+        ``'\\usepackage[acronym,toc]{}'``.
```
jfbu: missing ``glossaries`` here

jfbu (commented):
I would like to keep this optional. Rationale: `glossaries` package is rather big with many LaTeX dependencies. It will slow down compilation. It is needed to use `xindy` for any language other than English to fully benefit from utf-8 support (admittedly the current situation with `\makeindex` does not support it). The `glossaries` package is big but Sphinx can't provide interface to all its macros. It solves things at LaTeX level which are more easily done at Python level. Its manual runs into something like 250+ pages. Problems may arise on LaTeX side if people have too old TeX installation. I agree current situation is not very good (I added the theoretical support at #3082). For abbreviations (acronyms) alternatives exist, more light-weight on LaTeX side. But of course, this can be seen as a step towards glossaries per se. For reasons such as the above, I need first to think about how making it easier for Sphinx users to switch to this. At any rate, usage of glossaries LaTeX package and accompanying perl script makeglossaries (+xindy which is new dependency), should be kept optional.

Edit: here is some info from the "short" user manual (still 36 pages):

> Don’t use \gls in chapter or section headings as it can have some unpleasant
> side-effects. Instead use \glsentrytext for regular entries and one of
> \glsentryshort, \glsentrylong or \glsentryfull for acronyms.
> Alternatively use glossaries-extra which provides special commands for use in
> section headings, such as \glsfmtshort{label}.

This is quite typical of the complications which can arise with LaTeX. Most packages are not done for automated document production but for single-user authoring, responsible for all the needed mark-up.

doc/config.rst
sephalon: Fixed.

sephalon:
I'll try to put this issue into a broader context; I think there are three topics where we have to decide whether and to what extend we want to rely on LaTeX package support. In the following, I'm describing the current situation, propose a basic solution (no or minimal LaTeX dependencies, does not cover complicated cases like e.g. internationalization) and an extended solution that relies on a sophisticated LaTeX package.

* Glossary
  * **Current situation:** Handled in Python, but in a limited way
  * **Basic solution:**: Likely current situation?
  * **Extended solution:** `glossaries` package
* Acronym (lists)
  * **Current situation:** Acronym expansion supported, lists unsupported
  * **Basic solution:** Add simple list support like for glossaries?
  * **Extended solution:** `glossaries` package
* Index
  * **Current situation:** `makeindex`
  * **Basic solution:** Handle in Python? Keep `makeindex`?
  * **Extended solution:** `xindy`

Correct me if I'm wrong, but I think the extended solutions cannot be provided as extensions as integration into the builder/writer is too deep at this point. So I guess configuration options would be the way to go (defaulting on basic).

What do you think about this?

sephalon:
I'd like to raise this issue again; what do you think about my proposal? In the meanwhile we've got (optional) xindy support for indexes, but for glossaries and acronym lists we still lack an extended solution.

sephalon:
As we really need abbreviation lists in our company, I need to keep maintaining this feature in our own fork. Any chance we can get this feature upstream in one or the other way? I made some proposal quite some time ago; any comments on that?

jfbu:
@sephalon sorry for late. Your proposal is not forgotten. I am currently lacking the needed time. But I am planning to address this at worst in 2-3 months. Hoping for your understanding. Of course any help is welcome.

sephalon:
This has been stale for quite some now; @jfbu, any chance that you have some time to look at my proposal so we can decide on a direction to drive this further?

jfbu:
@sephalon sorry indeed for long time. I don't have much time.

jfbu:
> I'll try to put this issue into a broader context; I think there are three topics where we have to decide whether and to what extend we want to rely on LaTeX package support. In the following, I'm describing the current situation, propose a basic solution (no or minimal LaTeX dependencies, does not cover complicated cases like e.g. internationalization) and an extended solution that relies on a sophisticated LaTeX package.
> 
>     * Glossary
>       
>       * **Current situation:** Handled in Python, but in a limited way
>       * **Basic solution:**: Likely current situation?
>       * **Extended solution:** `glossaries` package
> 

See #11814 for some comments I made which detail how to support LaTeX `glossaries` package on our side.  The current Latexmk custom rule in our `latexmkrc` template is plain wrong even if we attempted to use `makeindex` as it refers to config file `gglo.ist` which is for LaTeX package authors to process their `\changes{}` mark-up in their commented LaTeX code.  If we use the `gloassaries` package we only have to replace it by `makeglossaries` shell command, which is a Perl script.


>     * Acronym (lists)
>       
>       * **Current situation:** Acronym expansion supported, lists unsupported
>       * **Basic solution:** Add simple list support like for glossaries?
>       * **Extended solution:** `glossaries` package
> 

Indeed if we process glossary entries by `glossaries` package for LaTeX builder, we can as well do that for acronyms.


>     * Index
>       
>       * **Current situation:** `makeindex`
>       * **Basic solution:** Handle in Python? Keep `makeindex`?
>       * **Extended solution:** `xindy`
> 

The xindy has been implemented for some years now, see config varaible `latex_use_xindy` which is automatically set to True if `latex_engine` is `'xelatex'` or `'lualatex'`.

> 
> Correct me if I'm wrong, but I think the extended solutions cannot be provided as extensions as integration into the builder/writer is too deep at this point. So I guess configuration options would be the way to go (defaulting on basic).
> 
> What do you think about this?

As with `latex_use_xindy` we could add `latex_glossary_via_latex` or a better name for a boolean config.

jfbu:
@sephalon I think a change like your PR should be accompanied with similar for glossary situation.  It will be simpler probably due to the latexmkrc template to do both acronyms and glossaries via `makeglossaries`.  

In my comments to #11814 I explained that some digging into LaTeX `glossaries` package is needed to see if we can maintain printing of glossaries as in the current Sphinx via-Python way, i.e. where there are located by the use of the `.. glossary::` directive.  This seems to mandate one more config variable to choose between that or a single glossary at the end, or glossaries at the end (from brief look at `glossaries` doc, I see it has a command `\printglossaries` which prints all defined individual glossaries at same location but possibly one can print them individually I don't know.

ping @picnixz

picnixz:
> where there are located by the use of the .. glossary:: directive

Yes. When we parse the glossary directive we could add in which document it was defined. Currently, visiting `glossary` nodes does nothing (i.e., we simply visit its children normally) so writers could remember the glossary locations. For LaTeX, it would then be possible to know that a glossary should be printed "here".

The `glossaries` package allows us you to create "categories" so if you have multiple glossaries in multiple places, if you want to keep them where they were located, you need to create independent glossaries (see §9 of [https://mirror.foobar.to/CTAN/macros/latex/contrib/glossaries/glossaries-user.pdf](https://mirror.foobar.to/CTAN/macros/latex/contrib/glossaries/glossaries-user.pdf)) and add the terms you defined in that document to this independent glossary.

Then, we simply call `\printglossary[type=local-glossary-name]` and it would be printed normally. If we want a global glossary, it's easier since we simply add it to the main glossary (the default one) and call `\printglossary` at the end of the book. For the 'local' glossaries, we could put them at the beginning of each section for instance (though I'm not sure whether we define "sections" in the LaTeX output).

PR #4866, opened 2018-04-20
Wrap graphviz diagrams in <div class="graphviz">

### LINKED ISSUE #4865: CSS in default theme breaks clickable map for inheritance-diagrams
Both the default Sphinx theme and the Read-the-Docs theme define

    .rst-content img {max-width: 100%;}

in their `theme.css`. This also gets applied to the diagrams created by the `sphinx.ext.inheritance_diagram` extension, if `graphviz_output_format = 'png'` in `docs/conf.py` (the default). When the diagram is large, the `max-width` CSS setting will effectively scale it down. In this case the clickable map that is overlaid on top of the png image is *not* scaled, resulting in an unusable diagram. This is illustrated in the following screenshot:

screenshot

Note the position of the mouse pointer, which is hovering over the map-location associated with the "KetPlus" box. That click location, however, is not at the visual location of the "KetPlus" box in the diagram.

The problem does not occur for `graphviz_output_format = 'svg'`. A user might work around the problem for the png format by customizing the templates for their documentation by creating the file `docs/_templates/layout.html` to contain

    {% extends "!layout.html" %}
    {% set css_files = css_files + ["_static/mycss.css"] %}

and then define

    img.inheritance {
      max-width: none;
    }

in `docs/_static/mycss.css`.

However, both with this workaround for the png format, and when using the svg format, another problem comes up: The diagrams now extend into the margin, which looks very ugly with most themes. It would be preferable if the diagrams were unscaled, but in a scrollable box (the `overflow: auto` CSS attribute).

Unfortunately, the code generated by the `sphinx.ext.graphviz` plugin (which `sphinx.ext.inheritance_diagram` relies on) creates an `` tag for png or an `<object>` tag for svg, for neither of which the CSS attribute is effective. Therefore, I would suggest that these tags should be enclosed in a `<div class="graphviz">...</div>` to make the diagrams fully stylable via CSS.
Ideally, the standard themes would be modified to both have `max-width: none;`, as above, and

    div.graphviz {
      overflow: auto;
    }

At least, having the additional `<div>` would allow any user to work around the problem in their own custom CSS file.

goerz:
The graphviz extension is modified to wrap the `` and `<object>`
tags it creates for diagrams in the png and svg formats in an additional
`<div class="graphviz">...</div>` tag. This allows to apply additional
CSS styling to the the diagrams that cannot be applied to `` and
`<object>` directly. Most notably,

    div.graphviz {
      overflow: auto;
    }

would be a useful style that the standard themes might consider.

In any case, it would allow users to add custom CSS code to their
project that gets around the problems with inheritance-diagrams
discussed in https://github.com/sphinx-doc/sphinx/issues/4865

The tests for the `grahviz` and `inheritance_diagram` have been modified
to test for the presence of the additional `<div>` tag.

Closes #4865

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -290,9 +290,11 @@ def render_dot_html(self, node, code, options, prefix='graphviz',
             self.body.append('<div align="%s" class="align-%s">' %
                              (node['align'], node['align']))
         if format == 'svg':
-            svgtag = '''<div class="graphviz"><object data="%s" type="image/svg+xml">
-            <p class="warning">%s</p></object></div>\n''' % (fname, alt)
-            self.body.append(svgtag)
+            self.body.append('<div class="graphviz">')
+            self.body.append('<object data="%s" type="image/svg+xml" %s>\n' %
+                             (fname, imgcss))
+            self.body.append('<p class="warning">%s</p>' % alt)
+            self.body.append('</object></div>\n')
         else:
             with codecs.open(outfn + '.map', 'r', encoding='utf-8') as mapfile:  # type: ignore
                 imgmap = ClickableMapDefinition(outfn + '.map', mapfile.read(), dot=code)
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -290,21 +290,25 @@ def render_dot_html(self, node, code, options, prefix='graphviz',
             self.body.append('<div align="%s" class="align-%s">' %
                              (node['align'], node['align']))
         if format == 'svg':
-            svgtag = '''<object data="%s" type="image/svg+xml">
-            <p class="warning">%s</p></object>\n''' % (fname, alt)
+            svgtag = '''<div class="graphviz"><object data="%s" type="image/svg+xml">
+            <p class="warning">%s</p></object></div>\n''' % (fname, alt)
             self.body.append(svgtag)
         else:
             with codecs.open(outfn + '.map', 'r', encoding='utf-8') as mapfile:  # type: ignore
                 imgmap = ClickableMapDefinition(outfn + '.map', mapfile.read(), dot=code)
                 if imgmap.clickable:
                     # has a map
-                    self.body.append('<img src="%s" alt="%s" usemap="#%s" %s/>\n' %
+                    self.body.append('<div class="graphviz">')
+                    self.body.append('<img src="%s" alt="%s" usemap="#%s" %s/>' %
                                      (fname, alt, imgmap.id, imgcss))
+                    self.body.append('</div>\n')
                     self.body.append(imgmap.generate_clickable_map())
                 else:
                     # nothing in image map
-                    self.body.append('<img src="%s" alt="%s" %s/>\n' %
+                    self.body.append('<div class="graphviz">')
+                    self.body.append('<img src="%s" alt="%s" %s/>' %
                                      (fname, alt, imgcss))
+                    self.body.append('</div>\n')
         if 'align' in node:
             self.body.append('</div>\n')
 
--- a/tests/test_ext_graphviz.py
+++ b/tests/test_ext_graphviz.py
@@ -22,24 +22,26 @@ def test_graphviz_png_html(app, status, warning):
     app.builder.build_all()
 
     content = (app.outdir / 'index.html').text()
-    html = (r'<div class="figure" .*?>\s*<img .*?/>\s*<p class="caption">'
+    html = (r'<div class="figure" .*?>\s*'
+            r'<div class="graphviz"><img .*?/></div>\s*<p class="caption">'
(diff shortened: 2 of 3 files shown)


goerz:
http://qnet.readthedocs.io/en/latest/symbolic_algebra.html illustrates the result of the applied patch, together with the proper custom CSS to solve #4865: it has scrollable diagrams with proper click-maps

goerz:
Travis reports

    ./sphinx/ext/graphviz.py:301:96: E501 line too long (111 > 95 characters)
    ./sphinx/ext/graphviz.py:306:96: E501 line too long (98 > 95 characters)

Would you like me to fix this?

tests/test_ext_graphviz.py
```diff
@@ -52,36 +54,36 @@ def test_graphviz_svg_html(app, status, warning):
...
     html = (r'<div align=\"center\" class=\"align-center\">'
-            r'<object data=\".*\.svg\".*>\n'
+            r'<div class="graphviz"><object data=\".*\.svg\".*>\n'
             r'\s+<p class=\"warning\">digraph foo {\n'
             r'centered\n'
-            r'}</p></object>\n'
+            r'}</p></object></div>\n'
             r'</div>')
-    assert re.search(html, content, re.S)
+    #assert re.search(html, content, re.S)
```
tk0miya: Why did you comment out here?

sphinx/ext/graphviz.py
```diff
@@ -290,21 +290,26 @@ def render_dot_html(self, node, code, options, prefix='graphviz',
...
-            svgtag = '''<object data="%s" type="image/svg+xml">
-            <p class="warning">%s</p></object>\n''' % (fname, alt)
+            svgtag = '''<div class="graphviz"><object data="%s" type="image/svg+xml">
+            <p class="warning">%s</p></object></div>\n''' % (fname, alt)
             self.body.append(svgtag)
         else:
             with codecs.open(outfn + '.map', 'r', encoding='utf-8') as mapfile:  # type: ignore
-                imgmap = ClickableMapDefinition(outfn + '.map', mapfile.read(), dot=code)
+                imgmap = ClickableMapDefinition(
+                    outfn + '.map', mapfile.read(), dot=code)
```
tk0miya: Why did you fold this line?

sphinx/ext/graphviz.py
```diff
@@ -290,21 +290,26 @@ def render_dot_html(self, node, code, options, prefix='graphviz',
...
                 if imgmap.clickable:
                     # has a map
-                    self.body.append('<img src="%s" alt="%s" usemap="#%s" %s/>\n' %
-                                     (fname, alt, imgmap.id, imgcss))
+                    self.body.append(
+                        '<div class="graphviz"><img src="%s" alt="%s" '
+                        'usemap="#%s" %s/></div>\n' %
+                        (fname, alt, imgmap.id, imgcss))
```
tk0miya: I feel additional `self.body.append` is better:
```
+ self.body.append('<div class="graphviz">')
  self.body.append('<img src="%s" alt="%s" usemap="#%s" %s/>' %
                   (fname, alt, imgmap.id, imgcss))
+ self.body.append('</div>\n')
```

What do you think?

tk0miya (changes requested):
LGTM with nits.

tests/test_ext_graphviz.py
goerz: While fixing the tests, I went through the different `assert`s in the tests one by one, and then I forgot to un-comment the last one. Good catch! Fixed.

sphinx/ext/graphviz.py
goerz: The first time I pushed, TravisCI complained about some overlong lines. In the process of fixing these , I also folded the line in question, even through it wasn't over Travis' 95 character limit. But I suppose you're right, and I shouldn't do "PEP8 cleanups" on lines not directly related to my patch. Reverted.

sphinx/ext/graphviz.py
goerz: I agree, your solution looks cleaner. I've changed the code to use additional appends. I'll force-push a new version with all of the changes above in a minute.

goerz:
@tk0miya I just force-pushed a new version with all your requested changes. This should be ready to merge now. Thanks for the review!

tk0miya (approved):
Thank you for quick work.
LGTM!

goerz:
Actually, can you hold off on merging for a minute? I found the tests do not cover the inheritance-diagram with SVG.

Also, for consistency, it would probably best if the SVG `<object>` had the same `class="inheritance"` attribute that the PNG `` has.

tk0miya:
+1 again.
I'll merge this within a week if no objection.

tk0miya:
Merged. Thank you for contribution!

PR #4879, opened 2018-04-22
Let docutils know the location of ``docutils.conf`` for Sphinx

tk0miya:
### Feature or Bugfix
- Refactoring
- Bugfix

### Purpose
- refs: #4798
- Let docutils know correct location of `docutils.conf` through envvar `$DOCUTILSCONF`
- After the change, we don't have to override class variable of `OptionParser` class directly.
- This introduces an breaking change; Sphinx ignores `docutils.conf` on `$HOME` or `/etc`.

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -19,7 +19,6 @@ from collections import defaultdict
 from copy import copy
 from os import path
 
-from docutils.frontend import OptionParser
 from docutils.utils import Reporter, get_source_line
 from six import BytesIO, class_types, next
 from six.moves import cPickle as pickle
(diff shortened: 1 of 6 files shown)


sphinx/environment/__init__.py
```diff
@@ -561,9 +560,8 @@ def read_doc(self, docname, app=None):
         """Parse a file and add/update inventory entries for the doctree."""
         self.prepare_settings(docname)
 
-        docutilsconf = path.join(self.srcdir, 'docutils.conf')
```
tk0miya, line 564: So far, Sphinx searches `docutils.conf` from srcdir. But I think confdif is much appropriate for this case.

shimizukawa (approved):
LGTM!

tk0miya:

PR #4887, opened 2018-04-23
Fix issue #4885

knorth55:
Subject: Fix issue reported in #4885

### Feature or Bugfix

- Bugfix

### Purpose
- I got same error as #4885 and I want to fix it.

### Detail
- `self.env` is not defined
  - `self.env` -> `env`
- `modules[modname]` is tuple
  - `modules[modname]` -> `modules[modname][0]`

### Relates
- https://github.com/sphinx-doc/sphinx/issues/4885


tk0miya:
Absolutely right.
Thank you for your work!

PR #4891, opened 2018-04-23
Fix PyPI link in theming docs

brainwane:
Subject: Fix link to PyPI in theming documentation

### Feature or Bugfix
- Bugfix

### Purpose
- Followup to #4862. I duplicated the functionality of the old link (to Sphinx themes as PyPI packages), which I checked using https://legacy.pypi.org/pypi/?:action=browse&c=599 .


tk0miya:
Good point!
Thank you for your work!

PR #4905, opened 2018-04-25
Add template rendering logic to :download: directive

beamath2:
Subject: Add template rendering logic to :download: directive

### Feature or Bugfix
- Feature

### Purpose
- Add the possibility of passing :download: directive filename strings through a sandboxed Jinja2 renderer.
- Mostly useful to do string interpolation when trying to include a dynamically-named build product as a downloadable item (for example, I use it to include a PDF with a filename that contains the documentation version).

### Detail
- Added a bool config variable : _download_interpolate_
- If _download_interpolate_ is True, then every filename string in :download: directives is treated as a Jinja2 template.
- The template renderer has access to all config variables, including those in conf.py
- Added some useful error messages considering the expected use case.


tk0miya:
Please tell me what config value is useful for download filenames?
I guess you'd like to add it through `add_config_value()` API. Is this right?

beamath2:
Yes, in my project I use `add_config_value()` (I added one for a PDF filename with the project name + documentation revision number), but you should be able to use any config value, even those that are just defined in your `conf.py`

It would be up to users whether they want to use this feature or not, which is why it's off by default.

tk0miya:
-1; It is better to provide such feature as an extension. You can do it with Transform module and `app.add_transform()` API.

beamath2:
Right, that might be better.

Would you consider adding it as a builtin extension? Otherwise I'll close this PR and open a new one in sphinx-contrib once I'm done with the modifications.

tk0miya:
I think this is useful, but it might not be widely used.
So I propose you to make this the 3rd party extension.

tk0miya:
Thank you for proposal!

Issue #4915, opened 2018-04-29
links on search page are broken when using dirhtml builder


davidism:
#4295 moved the documentation options into a separate JS file. This breaks the search page when using the `dirhtml` builder because links no longer point to the parent directory, they point below the `/search/` page.

Previously, the JS was generated on each page, so `pathto('', 1)` generated `../` for the search page. However, now that the file is separate which doesn't extend `layout.html` (since it's not an HTML template), `url_root` is undefined and the code always generates `''`.

https://github.com/sphinx-doc/sphinx/blob/feee462836677b23041a910734be221bcc0731bb/sphinx/themes/basic/static/documentation_options.js_t#L2

Before the change, a link from the search page would look like`../api/`, so the URL ended up being `/api/`. Now, it looks like `api/`, so the URL is  `/search/api/`.

 See pallets/flask#2743 for an example of this issue.

davidism:
I'm not clear why that was seen as a solution to the CSP error with inline JS, since the search page still contains inline JS.

davidism:
My solution for now is to override the `extrahead` block in `layout.html`:

```html+jinja
{% extends 'basic/layout.html' %}
{% block extrahead %}
<script>DOCUMENTATION_ROOT.URL_ROOT = '{{ url_root }}';</script>
{{ super() }}
{% endblock %}
```

ghost:
Hi, thanks for pointing out the issue. You are correct on two fronts: the non-HTML template expands config variables (but not the built-in functions) and it seems to only get expanded once anyway, so it's not a solution. As for the search page containing inline JS, that needs to be fixed and thanks for catching that. Here's a quick fix:

```
-  <script type="text/javascript" src="{{ pathto('_static/js/documentation_options.js', 1) }}"></script>
+  <script type="text/javascript" id="documentation_options" data-url_root="{{ pathto('', 1) }}" src="{{ pathto('_static/js/documentation_options.js', 1) }}"></script>
```

```
-    URL_ROOT: '{{ url_root }}',
+    URL_ROOT: document.getElementById("documentation_options").getAttribute('data-url_root'),
```

Let me know if that works for you before I submit a PR.

tk0miya:
Fixed by #4916.
Thank you for reporting!

PR #4916, opened 2018-04-30
Fix blank URL_ROOT

ghost:
Subject: Fix blank URL_ROOT

### Feature or Bugfix
- Bugfix

### Purpose
pathto doesn't expand in non-HTML templates so the workaround involves
expanding it in the HTML file and passing it as a data attribute to
the script.

### Detail
- Bug: Test it with the dirhtml builder and visiting the search page. Because URL_ROOT is blank, the result links are not the expected URL.

### Relates
- #4915

### DIFF
--- a/sphinx/themes/basic/layout.html
+++ b/sphinx/themes/basic/layout.html
@@ -87,7 +87,7 @@
 {%- endmacro %}
 
 {%- macro script() %}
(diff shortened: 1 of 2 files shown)


tk0miya (changes requested):
+1: But this should be merged into 1.7.x release. Could you rebase this into 1.7 branch please?

tk0miya (approved):

tk0miya:

PR #4922, opened 2018-05-01
Specify version in README.rst readthedocs shield

mondeja:
### Subject
Fix readthedocs shield in README.rst file

### Feature or Bugfix
Readthedocs status is shown as `docs unknown` before this change because any version is specified on image url.


tk0miya:
Thanks!

PR #4927, opened 2018-05-01
Display a warning when invalid values are passed to linenothreshold parameter of Highlight directive

mondeja:
Display a warning in `Highlight` directive if `linenothreshold` parameter value isn't valid and coerce exception cacthing to `ValueError` instead of a general `Exception` object bacuse `ValueError` is the unique possible error raised at that context.

Subject: Display a warning passing invalid values to `Highlight` directive

### Feature or Bugfix
- Feature

### Purpose
- Code refactoring and manage explicitly uncached errors.

### DIFF
--- a/sphinx/directives/code.py
+++ b/sphinx/directives/code.py
@@ -53,8 +53,14 @@ class Highlight(SphinxDirective):
         if 'linenothreshold' in self.options:
             try:
                 linenothreshold = int(self.options['linenothreshold'])
-            except Exception:
-                linenothreshold = 10
+            except ValueError:
+                default_linenothreshold = 10
+                location = self.state_machine.get_source_and_line(self.lineno)
+                logger.warning(__('linenothreshold value must be int typed. '
+                                  'Passed %r. Using default value (10).'
+                                  % self.options['linenothreshold']),
+                              location=location)
+                linenothreshold = default_linenothreshold
         else:
             linenothreshold = sys.maxsize
         return [addnodes.highlightlang(lang=self.arguments[0].strip(),
--- a/sphinx/directives/code.py
+++ b/sphinx/directives/code.py
@@ -57,8 +57,9 @@ class Highlight(SphinxDirective):
                 default_linenothreshold = 10
                 location = self.state_machine.get_source_and_line(self.lineno)
                 logger.warning(__('linenothreshold value must be int typed. '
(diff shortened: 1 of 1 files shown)


sphinx/directives/code.py
```diff
@@ -53,8 +53,15 @@ def run(self):
         if 'linenothreshold' in self.options:
             try:
                 linenothreshold = int(self.options['linenothreshold'])
-            except Exception:
-                linenothreshold = 10
+            except ValueError:
+                default_linenothreshold = 10
+                location = self.state_machine.get_source_and_line(self.lineno)
+                logger.warning(__('linenothreshold value must be int typed. '
+                                  'Passed %r. Using default value (%d).'
```
tk0miya, line 60: `10` is not default value. Please see "else" clause. `sys.maxsize` is default value.
I don't know why `10` is used for fallback.

sphinx/directives/code.py
```diff
@@ -53,8 +53,15 @@ def run(self):
         if 'linenothreshold' in self.options:
             try:
                 linenothreshold = int(self.options['linenothreshold'])
-            except Exception:
-                linenothreshold = 10
+            except ValueError:
+                default_linenothreshold = 10
```
tk0miya, line 57: I prefer to use constants for such case. How about changing this to constant?

sphinx/directives/code.py
```diff
@@ -53,8 +53,15 @@ def run(self):
...
+                location = self.state_machine.get_source_and_line(self.lineno)
+                logger.warning(__('linenothreshold value must be int typed. '
+                                  'Passed %r. Using default value (%d).'
+                                  % (self.options['linenothreshold'],
+                                     default_linenothreshold)),
+                              location=location)
```
tk0miya, line 63: It seems flake8 warns this line. Check it please.
https://travis-ci.org/sphinx-doc/sphinx/jobs/373627298#L491

tk0miya (commented):
LGTM with nits.

sphinx/directives/code.py
tk0miya: This number was appeared since the beginning of `linenothreshold` option (f2ff0419bd).

I think this number is meaningless. I think fallback to `sys.maxint` is better.

tk0miya:
I made an alternative of this PR: #5003.

PR #4939, opened 2018-05-06
fix #4938: i18n doesn't handle node.title correctly

### LINKED ISSUE #4938: i18n doesn't handle node.title correctly; contents, topic, admonition, table and section
Original report is: https://groups.google.com/d/msg/sphinx-users/hdzuCi7AvRY/etw-3qWBBAAJ

### Problem
There are two problems

1. `make gettext` doesn't generate msgid for title of `topic` directive and `contents` directive

ex.
```
.. contents:: Contents Title

.. topic:: Topic Title
```

2. `make html` with i18n catalog doesn't apply translated message if msgid starts with `1. `

ex. `index.rst`
```
1. A Header
===========

.. admonition:: 1. Admonition Title

   Admonition body

.. topic:: 1. Topic Title

   Topic body

.. contents:: 1. Contents Title

.. table:: 1. Table Title

   +---+
   |Foo|
   +---+
```

`conf.py`:
```
locale_dirs = ['./locale']
language = 'ru'
gettext_compact = False
```

commands:
```
make gettext
sphinx-intl update -l ru
# edit ./locale/ru/LC_MESSAGES/index.po
make html
```

#### Error logs / results
no errors

#### Expected results
All text in the `index.rst` are translated with using `index.po` catalog.

[4938.zip](https://github.com/sphinx-doc/sphinx/files/1977426/4938.zip)

### Environment info
- OS: Any (Windows 10 for me)
- Python version: 3.6.1
- Sphinx version: 1.7.4

shimizukawa:
### Feature or Bugfix
- Bugfix for #4938 

### Purpose
1. `make gettext` generates msgid for title of `topic` directive and `contents` directive
2. `make html` with i18n catalog apply translated message even if msgid starts with `1. `

### Detail
see #4938


sphinx/transforms/i18n.py
```diff
@@ -247,13 +247,24 @@ def is_named_target(node):
             if isinstance(node, LITERAL_TYPE_NODES):
                 msgstr = '::\n\n' + indent(msgstr, ' ' * 3)
 
+            # Structural Subelements phase1
+            # There is a possibility that only the title node is created.
+            # see: http://docutils.sourceforge.net/docs/ref/doctree.html#structural-subelements
+            if isinstance(node, nodes.title):
+                # This generates: <section ...><title>msgstr</title></section>
+                msgstr = msgstr + '\n' + '-' * len(msgstr) * 2
```
shimizukawa, line 255: I don't think this code is good. I should refactor this `sphinx.transform.i18n` module later.

sphinx/util/nodes.py
```diff
@@ -47,6 +47,11 @@ def apply_source_workaround(node):
         node.rawsource = node.astext()  # set 'classifier1' (or 'classifier2')
     if isinstance(node, nodes.image) and node.source is None:
         node.source, node.line = node.parent.source, node.parent.line
+    if isinstance(node, nodes.title) and node.source is None:
+        # Uncomment these lines after merging into master(1.8)
+        # logger.debug('[i18n] PATCH: %r to have source: %s',
+        #              get_full_module_name(node), repr_domxml(node))
```
shimizukawa, line 53: this comment lines will uncomment after merging master

sphinx/util/nodes.py
```diff
@@ -132,6 +137,11 @@ def is_translatable(node):
 META_TYPE_NODES = (
     addnodes.meta,
 )
+ALLOWED_TRANSLATED_NODES = (
```
tk0miya: What is difference from `translatable` nodes?

On i18n transform, it is described as following:
```
# ignore block markups
if not isinstance(patch, ALLOWED_TRANSLATED_NODES):
    continue  # skip
```
Are these same meaning?

tk0miya (approved):
LGTM with nits

sphinx/util/nodes.py
shimizukawa: Not same.
When `msgstr` in po files contain block markup, `if not isinstance(patch, ALLOWED_TRANSLATED_NODES):` ignore such translations.

tk0miya: I feel it is confusable. Could you rename this please to more appropriate name.

shimizukawa: I thought about defining it by naming rather than listing node classes in the code of `i18n.py`. So, I named `ALLOWED_TRANSLATED_NODES` to them, but this was not a property of being given a single name.
The nodes.title that added this time is a hacky code to handle in pair with the processing in `i18n.py`.
I'll revert this part and add a comment for each node classes.

shimizukawa:
I'll merge this after test passing

PR #4941, opened 2018-05-06
Rename viewcode_import to viewcode_follow_imported_members (refs: #4035)

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- see discussion in #4035

### DIFF
--- a/doc/ext/viewcode.rst
+++ b/doc/ext/viewcode.rst
@@ -17,12 +17,11 @@ from the source to the description will also be inserted.
 
 .. warning::
 
-   If :confval:`viewcode_import` is True,
-   or if the :event:`viewcode-find-source` event does not find source code
-   for the given module,
-   ``viewcode`` will import the modules being linked to.
-   If any modules have side effects on import, these will be
-   executed by ``viewcode`` when ``sphinx-build`` is run.
+   If :confval:`viewcode_follow_imported_members` is True, or if the
+   :event:`viewcode-find-source` event does not find source code for the given
+   module, ``viewcode`` will import the modules being linked to.  If any modules
+   have side effects on import, these will be executed by ``viewcode`` when
+   ``sphinx-build`` is run.
 
    If you document scripts (as opposed to library modules), make sure their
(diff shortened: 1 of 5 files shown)


doc/ext/viewcode.rst
```diff
@@ -17,12 +17,11 @@ from the source to the description will also be inserted.
...
-   or if the :event:`viewcode-find-source` event does not find source code
-   for the given module,
-   ``viewcode`` will import the modules being linked to.
-   If any modules have side effects on import, these will be
-   executed by ``viewcode`` when ``sphinx-build`` is run.
+   If :confval:`viewcode_follow_imported_members` is True, or if the
+   :event:`viewcode-find-source` event does not find source code for the given
+   module, ``viewcode`` will import the modules being linked to.  If any modules
+   have side effects on import, these will be executed by ``viewcode`` when
+   ``sphinx-build`` is run.
```
tk0miya, line 24: I only modified the settings name here.

But I doubt `If viewcode_follow_imported_members is True` is right. I think viewcode tries to import the target module even if it is False. Is this right?

shimizukawa: You are right.

In bellow case will import `package` and it causes next importing for `package.foo` whether  `viewcode_follow_imported_members` is True or False.

`package/__init__.py`:
```
from .foo import func1
```
... and `index.rst`:
```
.. py:module:: package

.. py:function:: func1(a, b)

   This is func1
```

`viewcode_follow_imported_members` just control providing `[source]` link for imported members.
It's my wrong: https://github.com/sphinx-doc/sphinx/commit/e8b870de0cbaf724986b65de9552262ac253a6d2#diff-9c97d36a5b9915e67dc7c4850a190095R33

So, I think this warning message should be written as a caveat for viewcode extension itself regardless of `viewcode_follow_imported_members`.

tk0miya: Please look at the warning again. It is already written in top section. Not `viewcode_follow_imported_members`.
Anyway, I'll post another PR to fix this.

Thanks,

PR #4945, opened 2018-05-09
fix lang_COUNTRY not fallback correctly for IndexBuilder

zhsj:
### Feature or Bugfix
- Bugfix

### Detail

In IndexBuilder, it will fallback to `<lang>` when the lang
is passed as `<lang_COUNTRY>`, and will fallback to en when
not found.

https://github.com/sphinx-doc/sphinx/blob/169297d0b76bf0b503033dadeb14f9a2b735e422/sphinx/search/__init__.py#L266-L274

So there's no need to check lang before creating the IndexBuilder
instance.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -103,6 +103,7 @@ Bugs fixed
 
 * i18n: message catalogs were reset on each initialization
 * #4850: latex: footnote inside footnote was not rendered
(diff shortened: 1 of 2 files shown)


shimizukawa:
Thanks!

PR #4951, opened 2018-05-10
Make Config and BuildEnvironment pickable

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Add `__getstate__()` and `__setstate__()` helper to make `Config` and `BuildEnvironment` pickable
- It makes them simple and easy to understand

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -62,6 +62,12 @@ Deprecated
 * ``sphinx.writers.latex.LaTeXWriter.restrict_footnote()`` is deprecated
 * ``sphinx.writers.latex.LaTeXWriter.unrestrict_footnote()`` is deprecated
 * ``LaTeXWriter.bibitems`` is deprecated
+* ``BuildEnvironment.load()`` is deprecated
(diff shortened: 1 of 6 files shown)


shimizukawa (approved):
LGTM ;)

Using `__getstate__` and `__setstate__` for pickling is in favor. It's also good to deprecate the `frompickle` class method.
I'm a little worried that the process of pickle/unpickle environment object has been distributed to `application.py` and `builders/__init__.py`, but I think that is not a problem.

tk0miya:
Thank you for reviewing.
Certainly pickling and unpickling methods are placed both in app and in builder. I agree that is asymmetrical and not a good design. Is it better to add methods for storing and restoring?
Then nobody have to know how to serialize the environment object.
I know the pickle library is standard, so I might not worry so much.

shimizukawa:
I might not worry so much, too.

tk0miya:
Okay, let's go with this.
Thank you for comment!

PR #4966, opened 2018-05-13
Add BuildEnvironment.setup() to re-initialize (after unpickle)

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Make initialization of env simple

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -17,6 +17,7 @@ from docutils import nodes
 from six.moves import cPickle as pickle
 
 from sphinx.deprecation import RemovedInSphinx20Warning
+from sphinx.environment import CONFIG_OK, CONFIG_CHANGED_REASON
 from sphinx.environment.adapters.asset import ImageAdapter
 from sphinx.errors import SphinxError
 from sphinx.io import read_doc
@@ -422,11 +423,10 @@ class Builder(object):
         Store all environment docnames in the canonical format (ie using SEP as
         a separator in place of os.path.sep).
         """
-        updated, reason = self.env.update_config(self.config, self.srcdir, self.doctreedir)
-
         logger.info(bold('updating environment: '), nonl=True)
 
         self.env.find_files(self.config, self)
+        updated = (self.env.config_status != CONFIG_OK)
         added, changed, removed = self.env.get_outdated_files(updated)
 
         # allow user intervention as well
@@ -440,6 +440,7 @@ class Builder(object):
             changed.update(self.env.glob_toctrees & self.env.found_docs)
 
         if changed:
+            reason = CONFIG_CHANGED_REASON.get(self.env.config_status, '')
             logger.info('[%s] ', reason, nonl=True)
         logger.info('%s added, %s changed, %s removed',
                     len(added), len(changed), len(removed))
@@ -472,6 +473,9 @@ class Builder(object):
         for retval in self.app.emit('env-updated', self.env):
             if retval is not None:
                 docnames.extend(retval)
+
+        # workaround: marked as okay to call builder.read() twice in same process
+        self.env.config_status = CONFIG_OK
 
         return sorted(docnames)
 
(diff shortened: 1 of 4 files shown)


sphinx/builders/__init__.py
```diff
@@ -422,11 +432,10 @@ def read(self):
         Store all environment docnames in the canonical format (ie using SEP as
         a separator in place of os.path.sep).
         """
-        updated, reason = self.env.update_config(self.config, self.srcdir, self.doctreedir)
```
tk0miya, line 425: So far, `env.config` is replaced by new one here. It means all features refer old config object on incremental build.
I think that is not intended behavior. So I move the replacement to more early stage.

sphinx/builders/__init__.py
```diff
@@ -473,6 +483,9 @@ def read(self):
             if retval is not None:
                 docnames.extend(retval)
 
+        # workaround: marked as okay to call builder.read() twice in same process
+        self.env.config_status = CONFIG_OK
```
tk0miya, line 478: I feel touching status of env object directly here is not good. But I also feel ugly to add a new method...

sphinx/environment/__init__.py
```diff
@@ -193,6 +198,10 @@ def __init__(self, app):
         # attributes of "any" cross references
         self.ref_context = {}       # type: Dict[unicode, Any]
 
+        # set up environment
+        if app:
+            self.setup(app)
```
tk0miya, line 209: I made `app` argument optional, but I don't add DeprecationWarning here. Because I don't have strong opinion for it. Either is fine to me.

sphinx/environment/__init__.py
```diff
@@ -451,44 +515,6 @@ def check_dependents(self, app, already):
             if docname not in already:
                 yield docname
 
-    def update_config(self, config, srcdir, doctreedir):
```
tk0miya, line 454: Note: I didn't add DeprecationWarning here. Because both `update_config` and `_update_settings` are not released yet. They are only in master branch.

sphinx/environment/__init__.py
```diff
@@ -69,6 +69,12 @@
 # NOTE: increase base version by 2 to have distinct numbers for Py2 and 3
 ENV_VERSION = 53 + (sys.version_info[0] - 2)
 
+# config status
+CONFIG_OK = 1
+CONFIG_NEW = 2
+CONFIG_CHANGED = 3
+CONFIG_EXTENSIONS_CHANGED = 4
```
shimizukawa, line 76: Memo: We can use `Enum` for this definition. Because we still support Py2, we need to rely on the `enum34` package to use `Enum`.

sphinx/builders/__init__.py
```diff
@@ -52,6 +55,13 @@
 logger = logging.getLogger(__name__)
 
 
+CONFIG_CHANGED_REASON = {
```
shimizukawa: nits: Why is this REASON defined here instead of `environment/__init__.py`?

sphinx/environment/__init__.py
shimizukawa: OK

shimizukawa (approved):
LGTM with nits

sphinx/environment/__init__.py
tk0miya: Yes. Let's move after dropping py2!

sphinx/builders/__init__.py
tk0miya: These message catalog is only used here. Either is fine to me.

shimizukawa: I like putting it in `environment/__ init__.py`, but I do not have a strong opinion.

tk0miya: Acceptable. done

tk0miya:

sphinx/builders/__init__.py
jayaddison: This seems to somehow influence the bug reported in #12120 - I'm not sure exactly how/why yet, but it means that a test that should refresh its environment settings does not, and reads some stale/inaccurate env config.

PR #4972, opened 2018-05-14
Fix #4969: autodoc: constructor method should not have return annotation

### LINKED ISSUE #4969: autodoc maybe should ignore return annotations on classes
Subject: autodoc adds "→ None" to classes

### Problem
We noticed on Trio (https://github.com/python-trio/trio/issues/529) that some classes that have a return annotation on `__init__` are rendered by autodoc with "→ None".

Example:
http://trio.readthedocs.io/en/latest/reference-core.html#trio.Event
`class trio.Event → None`

This class uses `attrs`, which [adds a `-> None` annotation to the generated `__init__` method](https://github.com/python-attrs/attrs/commit/fc033b41a11c9ef3e5fd7bee7f936427d4bb5f3d#diff-21756223332b04173cec8910fda6655bR1264).

A minimal example could be:

```python
class Foo:
    """Class docstring"""
    def __init__(self) -> None:
        pass
```

And `.. autoclass:: Foo` will generate `class Foo → None`.

I would expect autodoc to ignore the return annotation for the `__init__` method when it is used to build the signature for the class itself. A return annotation of `None` is legal for `__init__` but it doesn't make sense to pick it up to document the class signature.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4969

### DIFF
--- a/tests/test_util_inspect.py
+++ b/tests/test_util_inspect.py
@@ -280,6 +280,10 @@ def test_Signature_annotations():
     sig = inspect.Signature(f11).format_args()
     assert sig == '(x: CustomAnnotation, y: 123) -> None'
 
+    # has_retval=False
+    sig = inspect.Signature(f11, has_retval=False).format_args()
+    assert sig == '(x: CustomAnnotation, y: 123)'
+
 
 def test_safe_getattr_with_default():
     class Foo(object):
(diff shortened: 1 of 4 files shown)


tests/test_util_inspect.py
```diff
@@ -280,6 +280,10 @@ def test_Signature_annotations():
     sig = inspect.Signature(f11).format_args()
     assert sig == '(x: CustomAnnotation, y: 123) -> None'
 
+    # no_retval=True
```
tk0miya: wrong comment...

PR #4982, opened 2018-05-18
Fix #4980: latex: Explicit labels on code blocks are duplicated

### LINKED ISSUE #4980: Explicit labels on code blocks are duplicated in LaTeX output
### Problem

The following snippet causes laTeX to complain about multiply-defined labels:

```rst
.. _test:

.. code:: python

   pass
```

This is low-priority, because using a proper `:name:` annotation works fine:

```
.. code:: python
   :name: test

   pass
```

(But it's still an issue when one wants to place an anchor before a list of multiple code samples.)

#### Error logs / results
```
Latexmk: List of undefined refs and citations:
  Label `index:test' multiply defined
Latexmk: Summary of warnings:
  Latex found 1 multiply defined reference(s)
```

### Environment info
- OS: GNU/Linux
- Python version: 3.5.2
- Sphinx version: 1.7.4

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4980


sphinx/writers/latex.py
```diff
@@ -1305,12 +1313,7 @@ def visit_table(self, node):
...
-        labels = ''  # type: unicode
-        for labelid in self.pop_hyperlink_ids('table'):
-            labels += self.hypertarget(labelid, anchor=False)
-        if node['ids']:
-            labels += self.hypertarget(node['ids'][0], anchor=False)
-
+        labels = self.hypertarget_to(node)
         table_type = self.table.get_table_type()
         table = self.render(table_type + '.tex_t',
                             dict(table=self.table, labels=labels))
```
tk0miya, line 1324: Note: I noticed the labels are enabled only if table has a caption. I will fix it in another PR.

sphinx/writers/latex.py
```diff
@@ -1754,16 +1757,8 @@ def depart_image(self, node):
...
-        for id in self.pop_hyperlink_ids('figure'):
-            ids += self.hypertarget(id, anchor=False)
-        if node['ids']:
-            ids += self.hypertarget(node['ids'][0], anchor=False)
+        targets = self.hypertarget_to(node)
         self.restrict_footnote(node)
-        if (len(node.children) and
-           isinstance(node.children[0], nodes.image) and
-           node.children[0]['ids']):
-            ids += self.hypertarget(node.children[0]['ids'][0], anchor=False)
```
tk0miya, line 1770: Originally, this if-block was introduced to adjust `:name:` option of the `figure` directive. But it is not needed at present because we added patched `figure` directive to hack it (see `sphinx/directives/patches.py`).

Therefore, I removed this block simply.

sphinx/writers/latex.py
```diff
@@ -1908,22 +1908,16 @@ def add_target(id):
                         node.parent.parent.index(node.parent)]
                 else:
                     raise
-            if isinstance(next, nodes.section):
+            domain = self.builder.env.get_domain('std')
+            figtype = domain.get_figtype(next)
+            if figtype and domain.get_numfig_title(next):
```
tk0miya, line 1921: This block is still needed only for captioned `literal_block`. It seems they are not broken. So I will refactor them on master branch later.

sphinx/writers/latex.py
```diff
@@ -1032,9 +1040,9 @@ def visit_title(self, node):
...
                     # just use "subparagraph", it's not numbered anyway
                     self.body.append(r'\%s%s{' % (self.sectionnames[-1], short))
-                self.context.append('}\n')
-
+                self.context.append('}\n' + self.hypertarget_to(node.parent))
                 self.restrict_footnote(node)
+
                 if self.next_section_ids:
                     for id in self.next_section_ids:
                         self.context[-1] += self.hypertarget(id, anchor=False)
```
tk0miya, line 1053: Note: Handling `next_section_ids` is still used for `:doc:` reference target.

sphinx/writers/latex.py
```diff
@@ -727,6 +733,10 @@ def hypertarget(self, id, withdoc=True, anchor=True):
         return (anchor and '\\phantomsection' or '') + \
             '\\label{%s}' % self.idescape(id)
 
+    def hypertarget_to(self, node, anchor=False):
+        # type: (nodes.Node, bool) -> unicode
+        return ''.join(self.hypertarget(node_id, anchor=anchor) for node_id in node['ids'])
+
```
jfbu: What do you think about inserting the `\phantomsection` only once, in case `anchor` is not False. Something like

```
    if node.get('ids'):
        return (anchor and '\\phantomsection' or '') + \
            ''.join(self.hypertarget(node_id, anchor=False) for node_id in node['ids'])
    return ''
```

sphinx/writers/latex.py
```diff
@@ -1754,16 +1757,8 @@ def depart_image(self, node):
 
     def visit_figure(self, node):
         # type: (nodes.Node) -> None
-        ids = ''  # type: unicode
-        for id in self.pop_hyperlink_ids('figure'):
-            ids += self.hypertarget(id, anchor=False)
-        if node['ids']:
-            ids += self.hypertarget(node['ids'][0], anchor=False)
+        targets = self.hypertarget_to(node)
```
jfbu: in depart_table you used `labels`. I would propose to use `labels` here too. It is easier for me because I know this is what the LaTeX code will look like: with  `\label` and possibly a `\phantomsection` before...

sphinx/writers/latex.py
```diff
@@ -1775,7 +1770,7 @@ def visit_figure(self, node):
                 self.body.append('\\begin{sphinxfigure-in-table}\n\\centering\n')
             if any(isinstance(child, nodes.caption) for child in node):
                 self.body.append('\\capstart')
-            self.context.append(ids + '\\end{sphinxfigure-in-table}\\relax\n')
+            self.context.append(targets + '\\end{sphinxfigure-in-table}\\relax\n')
```
jfbu: see my remark about `s/targets/labels/g`

sphinx/writers/latex.py
```diff
@@ -2223,15 +2217,10 @@ def visit_literal_block(self, node):
             self.in_parsed_literal += 1
             self.body.append('\\begin{sphinxalltt}\n')
         else:
-            ids = ''  # type: unicode
-            for id in self.pop_hyperlink_ids('code-block'):
-                ids += self.hypertarget(id, anchor=False)
-            if node['ids']:
-                # suppress with anchor=False \phantomsection insertion
-                ids += self.hypertarget(node['ids'][0], anchor=False)
+            targets = self.hypertarget_to(node)
```
jfbu: `s/targets/labels/g`  ?

jfbu (commented):
I made minor remarks (I do not master enough understanding of globality for deeper review for time being).

sphinx/writers/latex.py
tk0miya: Good idea! I'll improve this later.

sphinx/writers/latex.py
tk0miya: Okay :-)

tk0miya:
updated :-)

jfbu (approved):
LGTM. In future, should displaymath node be made part of HYPERLINK_SUPPORT_NODES?

tk0miya:
>LGTM. In future, should displaymath node be made part of HYPERLINK_SUPPORT_NODES?

Absolutely! I'll do it later (in master branch).

tk0miya:
Merged.

Thank you for reviewing!

sphinx/writers/latex.py
tk0miya: see  #5008.

sphinx/writers/latex.py
tk0miya: see  #5008.

tk0miya:
About displaymath, I just posted an issue: #5010

PR #4987, opened 2018-05-19
Fix #4979: latex: Incorrect escaping of curly braces in index entries

### LINKED ISSUE #4979: Incorrect escaping of curly braces in index entries
### Problem

As explained in https://tex.stackexchange.com/questions/153291/index-unmatched-braces-in-latex, the LaTeX `\index` command requires matched braces: it's not correct to write this: `\index{main \{}`.

Unfortunately, that's the code that Sphinx generates for the following example:

```rst
.. index::
   single: main {
```

```latex
\index{main \{}\ignorespaces 
````

This problem happens in real life, ironically, when trying to index a collection of error messages from a compiler. One of the error messages is something like `Unmatched delimiter: {`, which trips Sphinx.

#### Error logs / results

```
Runaway argument?
{main \{}\ignorespaces 
! Paragraph ended before \@wrindex was complete.
<to be read again> 
                   \par 
l.71 
```

#### Expected results

As mentioned in the StackOverflow post above, the solution would be to escape backslashes more aggressively — using a separate macro definition, for example.

### Environment info
- OS: GNU/Linux
- Python version: 3.5.2
- Sphinx version: 1.7.4

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- see #4979

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -30,6 +30,7 @@ Bugs fixed
 * #4969: autodoc: constructor method should not have return annotation
 * latex: deeply nested enumerated list which is beginning with non-1 causes
   LaTeX engine crashed
+* #4979: latex: Incorrect escaping of curly braces in index entries
 
 Testing
 --------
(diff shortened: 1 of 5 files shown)


jfbu (commented):
LGTM!

jfbu (approved):

tk0miya:
Thank you for quick reviewing :-)

PR #4989, opened 2018-05-19
refactor: Move math_reference node to sphinx.builders.latex.nodes

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- The `math_reference` node is only used in latex builder.  So this makes it builder-specific node.

### DIFF
--- a/sphinx/ext/mathbase.py
+++ b/sphinx/ext/mathbase.py
@@ -14,7 +14,7 @@ import warnings
 from docutils import nodes
 
 from sphinx.addnodes import math, math_block as displaymath
-from sphinx.addnodes import math_reference as eqref  # NOQA  # to keep compatibility
+from sphinx.builders.latex.nodes import math_reference as eqref  # NOQA  # to keep compatibility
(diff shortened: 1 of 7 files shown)


sphinx/ext/mathbase.py
```diff
@@ -14,7 +14,7 @@
 from docutils import nodes
 
 from sphinx.addnodes import math, math_block as displaymath
-from sphinx.addnodes import math_reference as eqref  # NOQA  # to keep compatibility
+from sphinx.builders.latex.nodes import math_reference as eqref  # NOQA  # to keep compatibility
```
shimizukawa, line 17: (off topic) It would be better to emit DeprecationWarning to users who import `eqref` from this module.

shimizukawa (approved):
LGTM!

sphinx/ext/mathbase.py
tk0miya: Okay, let's talk it in another PR.

PR #4992, opened 2018-05-19
Fix #4973: latex: glossary directive adds whitespace to each item

### LINKED ISSUE #4973: phantomsection from glossary directive adds whitespace
### Problem
When the `glossary` directive is used, each term gets a `\phantomsection` with a `\label` so that it can later be referenced. This can add extra whitespace if using `enumitem` with `style=nextline` (and maybe other styles). There is probably good LaTeX terminology for what is happening, but the pictures and suggested fix should explain enough.

In `conf.py`:
```py
latex_elements = {
    'preamble': '\\usepackage{enumitem}\\setlist[description]{style=nextline}',
}
```

reStructured text document snippet:

```rst
Definition list
---------------

this is a label
    this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description

Glossary
--------

.. glossary::

    this is a label
        this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description
```

#### Results
PDF output (notice the difference in indentation between the first line of the glossary text and the rest):
extra space

Generated LaTeX:
```tex
\section{Definition list}
\label{\detokenize{index:definition-list}}\begin{description}
\item[{this is a label}] \leavevmode
this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description

\end{description}


\section{Glossary}
\label{\detokenize{index:glossary}}\begin{description}
\item[{this is a label\index{this is a label|textbf}}] \leavevmode\phantomsection\label{\detokenize{index:term-this-is-a-label}}\relax
this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description

\end{description}
```

#### Expected results
properly aligned

### Possible solution
Add a `\relax` after the label:
```tex
\item[{this is a label\index{this is a label|textbf}}] \leavevmode\phantomsection\label{\detokenize{index:term-this-is-a-label}}\relax
this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description this is a description
```
This is equivalent (I think) to placing the description directly after the label, without a linebreak. Using `%` instead of `\relax` also works.

Putting the phantom section and the label before `\item` also works, but that may effect how the hyperlink target works.

### Environment info
- OS: macOS 10.13.4
- Python version: 2.7.10
- Sphinx version: 1.7.4

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4973

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1555,7 +1555,11 @@ class LaTeXTranslator(nodes.NodeVisitor):
     def visit_term(self, node):
         # type: (nodes.Node) -> None
         self.in_term += 1
-        ctx = ''.join(self.hypertarget(node_id) for node_id in node['ids'])
+        ctx = ''  # type: unicode
+        if node.get('ids'):
+            ctx = '\\phantomsection'
+            for node_id in node['ids']:
+                ctx += self.hypertarget(node_id, anchor=False)
         ctx += '}] \\leavevmode'
         self.body.append('\\item[{')
         self.restrict_footnote(node)
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -725,7 +725,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
         # type: (unicode, bool, bool) -> unicode
         if withdoc:
             id = self.curfilestack[-1] + ':' + id
-        return (anchor and '\\phantomsection' or '') + \
+        return (anchor and r'\phantomsection\relax' or '') + \
             '\\label{%s}' % self.idescape(id)
 
     def hyperlink(self, id):
@@ -1555,9 +1555,8 @@ class LaTeXTranslator(nodes.NodeVisitor):
     def visit_term(self, node):
         # type: (nodes.Node) -> None
         self.in_term += 1
-        ctx = '}] \\leavevmode'  # type: unicode
-        if node.get('ids'):
-            ctx += self.hypertarget(node['ids'][0])
+        ctx = ''.join(self.hypertarget(node_id) for node_id in node['ids'])
+        ctx += '}] \\leavevmode'
         self.body.append('\\item[{')
         self.restrict_footnote(node)
         self.context.append(ctx)
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -725,7 +725,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
         # type: (unicode, bool, bool) -> unicode
         if withdoc:
             id = self.curfilestack[-1] + ':' + id
-        return (anchor and r'\phantomsection\relax' or '') + \
+        return (anchor and '\\phantomsection' or '') + \
             '\\label{%s}' % self.idescape(id)
 
     def hyperlink(self, id):
--- a/CHANGES
+++ b/CHANGES
@@ -36,6 +36,7 @@ Bugs fixed
 * #4979: latex: Incorrect escaping of curly braces in index entries
 * #4956: autodoc: Failed to extract document from a subclass of the class on
   mocked module
+* #4973: latex: glossary directive adds whitespace to each item
 
 Testing
 --------
--- a/tests/roots/test-glossary/conf.py
+++ b/tests/roots/test-glossary/conf.py
@@ -1,0 +1,7 @@
+# -*- coding: utf-8 -*-
+
+master_doc = 'index'
+
+latex_documents = [
+    (master_doc, 'test.tex', 'The basic Sphinx documentation for testing', 'Sphinx', 'report')
+]
--- a/tests/roots/test-glossary/index.rst
+++ b/tests/roots/test-glossary/index.rst
@@ -1,0 +1,22 @@
+test-glossary
+=============
+
+.. glossary::
+   :sorted:
+
+   boson
+      Particle with integer spin.
+
+   *fermion*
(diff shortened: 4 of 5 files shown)


sphinx/writers/latex.py
```diff
@@ -1555,9 +1555,10 @@ def depart_definition_list_item(self, node):
     def visit_term(self, node):
         # type: (nodes.Node) -> None
         self.in_term += 1
-        ctx = '}] \\leavevmode'  # type: unicode
-        if node.get('ids'):
-            ctx += self.hypertarget(node['ids'][0])
+        ctx = ''  # type: unicode
+        for node_id in node['ids']:
+            ctx += self.hypertarget(node_id, anchor=False)
+        ctx += '}] \\leavevmode'  # type: unicode
```
tk0miya: After this change, `\label` goes inside `\item`.  I believe `\phantomsection` is not needed in this case.

sphinx/writers/latex.py
```diff
@@ -1555,9 +1555,10 @@ def depart_definition_list_item(self, node):
     def visit_term(self, node):
         # type: (nodes.Node) -> None
         self.in_term += 1
-        ctx = '}] \\leavevmode'  # type: unicode
-        if node.get('ids'):
-            ctx += self.hypertarget(node['ids'][0])
+        ctx = ''  # type: unicode
+        for node_id in node['ids']:
+            ctx += self.hypertarget(node_id, anchor=False)
+        ctx += '}] \\leavevmode'
```
jfbu: The problem with this is that for description environment there is no counter which is stepped automatically at each item. The LaTeX `\label` can only create an hyperlink target for `\ref` when a counter was stepped.

jfbu: The `\phantomsection` of hyperref is done when no LaTeX counter is intrinsically associated with  a location.

One could do the `\phantomsection` before the `\leavevmode` but I think there could be a risk the link target goes to  previous page if a pagebreak happens precisely at the item (notice to understand that the item label is printed only when TeX is processing the item contents, i.e., **after** it has absorbed the `\item` command, not during it).

Thus I think good way is simply to add the `\relax` as asked by @jessetan

jfbu (changes requested):
Moving the `\label` inside the `\item` works only with enumerated lists. Description list items to not step a counter so `\label` must be helped with `\phantomsection`. I would go for the `\relax` as commented by OP at #4973

sphinx/writers/latex.py
jfbu: Ah sorry, I had not checked fully one point. In the context of #4973, the phantom section target will be below item label not above it, (due to interference of added enumitem package). It works in my testing with the \phantomsection inside the `\item[...]`. Brief testing did not allow me to reproduce a problem with link target of `\ref` going to page before, in case of page break right at item location.

jfbu: Thus I recommend you do same but with a `\phantomsection` right before the `\label{..}`,  all inside `\item[...]`.

tk0miya: Updated. Thank you for your wisdom!

sphinx/writers/latex.py
```diff
@@ -725,7 +725,7 @@ def hypertarget(self, id, withdoc=True, anchor=True):
         # type: (unicode, bool, bool) -> unicode
         if withdoc:
             id = self.curfilestack[-1] + ':' + id
-        return (anchor and '\\phantomsection' or '') + \
+        return (anchor and r'\phantomsection\relax' or '') + \
```
jfbu: ah sorry I had not seen that change. Is it really needed?

sphinx/writers/latex.py
```diff
@@ -1555,9 +1555,12 @@ def depart_definition_list_item(self, node):
...
         # type: (nodes.Node) -> None
         self.in_term += 1
-        ctx = '}] \\leavevmode'  # type: unicode
+        ctx = ''  # type: unicode
         if node.get('ids'):
-            ctx += self.hypertarget(node['ids'][0])
+            ctx = '\\phantomsection'
+            for node_id in node['ids']:
+                ctx += self.hypertarget(node_id, anchor=False)
+        ctx += '}] \\leavevmode'
```
jfbu, line 1563: I pushed this to avoid unneeded \relax in output

jfbu (changes requested):
sorry @tk0miya I pushed directly to your branch but I had not seen a nother modification. I will push a second commit please feel free to revert both to start discussion afresh.

jfbu (approved):
I have now reverted extra `\relax`es.

Nota Bene: while testing I discovered that `\phantomsection` methods to create a target via `\label`  (usable from LaTeX with `\ref`)  fails if there is not at least a section in document. For example the glossary unit test creates such a file. But in real life documents have sections (a least a Glossary in the present context) so I don't think this is a problem.

(by "fail" I mean that the `\ref` will not be able to create a link in PDF output, not that there is failure of build)

jfbu:
@tk0miya I have pushed directly, can you review ? (I think github has no interface for requesting review from PR's author...)

tk0miya:
LGTM :-)
I'd misread your comments. Thank you for your help!

PR #4993, opened 2018-05-19
Fix #4946: py domain: type field could not handle "None" as a type

### LINKED ISSUE #4946: Too strict class/function/... matching in Sphinx >= 1.7.0
Subject: Too strict class/function/... matching in Sphinx >= 1.7.0

### Problem
- I think that matching of class / function / ... is too strict in Sphinx 1.7.0. This was introduced with the fix for #3927.

Consider the following example:
```
def ham(ham_par):
    """Ham.

    :param None ham_par: Ham parameter.
    """
    pass

def spam(spam_par):
    """Spam.

    :param int or None spam_par: Spam parameter.
    """
    pass
```
The method "ham" will build fine, spam will not. From what I understand both functions are documented using perfectly fine syntax (at least PyCharm can deal with it).

#### Error logs / results
```
 WARNING: py:class reference target not found: None
```

#### Expected results
If a strict match (in this case py:class None) is not found there should be some fallback to use other types (in this case py:data None, from what I can see in the intersphinx inventory).

### Environment info
- OS: Windows 10
- Python version: 2.7.x
- Sphinx version: 1.7.4

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4946

### DIFF
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -156,7 +156,15 @@ class PyGroupedField(PyXrefMixin, GroupedField):
 
 
 class PyTypedField(PyXrefMixin, TypedField):
-    pass
+    def make_xref(self, rolename, domain, target,
+                  innernode=nodes.emphasis, contnode=None, env=None):
+        # type: (unicode, unicode, unicode, nodes.Node, nodes.Node, BuildEnvironment) ->  nodes.Node  # NOQA
+        if rolename == 'class' and target == 'None':
+            # None is not a type, so use obj role instead.
+            rolename = 'obj'
+
+        return super(PyTypedField, self).make_xref(rolename, domain, target,
+                                                   innernode, contnode, env)
 
 
 class PyObject(ObjectDescription):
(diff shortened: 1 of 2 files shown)


tk0miya:
@shimizukawa I feel this change is a bit hacky. But this can keep "None" type available.
It is introduced in our document:
http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists

Could you let me know your opinion, please?

sphinx/domains/python.py
```diff
@@ -156,7 +156,15 @@ class PyGroupedField(PyXrefMixin, GroupedField):
 
 
 class PyTypedField(PyXrefMixin, TypedField):
-    pass
+    def make_xref(self, rolename, domain, target,
+                  innernode=nodes.emphasis, contnode=None, env=None):
+        # type: (unicode, unicode, unicode, nodes.Node, nodes.Node, BuildEnvironment) ->  nodes.Node  # NOQA
+        if rolename == 'class' and target == 'None':
+            # None is not a type, so use obj role instead.
```
shimizukawa, line 163: understood.

shimizukawa (approved):
LGTM!

tk0miya:
Thank you for quick reviewing!

PR #5000, opened 2018-05-21
Fix #4983: productionlist directive generates invalid IDs for the tokens

### LINKED ISSUE #4983: productionlist and token do not pass their IDs through nodes.make_id
Take this example from the manual:

```rst
.. productionlist::
   try_stmt: try1_stmt | try2_stmt
   try1_stmt: "try" ":" `suite`
            : ("except" [`expression` ["," `target`]] ":" `suite`)+
            : ["else" ":" `suite`]
            : ["finally" ":" `suite`]
   try2_stmt: "try" ":" `suite`
            : "finally" ":" `suite`
```

The generated HTML looks like this:

```html
…<strong id="grammar-token-try_stmt">try_stmt </strong> ::=  try1_stmt | try2_stmt…
```

Note how the ID contains an `_` character.  The responsible code is here:

https://github.com/sphinx-doc/sphinx/blob/4cdb51be8385baaa81ae815e2e84f1f81c934adf/sphinx/domains/std.py#L435

For comparison, here's how index entries are generated:

https://github.com/sphinx-doc/sphinx/blob/4cdb51be8385baaa81ae815e2e84f1f81c934adf/sphinx/domains/std.py#L256

Is it an oversight?  Should the code call `nodes.make_id`?

To clarify: I don't think it's a big issue, except when production names contain symbols, like this:

```rst
.. productionlist::
   except#name : except `name`
   except#expr : except (`expr`)

:token:`except#name`
```

which generates this `<strong id="grammar-token-except#name">except#name</strong>` and `<a class="reference internal" href="#grammar-token-except#name">…</a>`.  It's also not obvious how to fix this, since a fix might break links pointing to existing documentation.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4983


CHANGES
```diff
@@ -34,6 +34,7 @@ Incompatible changes
 * #1857: latex: :confval:`latex_show_pagerefs` does not add pagerefs for
   citations
 * #4648: latex: Now "rubric" elements are rendered as unnumbered section title
+* #4983: html: The URL for the productionlist has been changed
```
cpitclaudel, line 37: Maybe `anchors for productionlist tokens` rather than `URL for productionlist`?

tk0miya: Good point

PR #5001, opened 2018-05-21
Close #1431: latex: Add alphanumeric enumerated list support

### LINKED ISSUE #1431: Alphanumeric lists not possible in LaTeX and EPUB
I discovered that it is possible to make alphanumeric lists in HTML like this:

```
#!rst

I. Schnittblumen

   a. Rosen
   b. Nelken
   c. Tulpen

A. Topfblumen

   i. Petunien
   ii. Geranien
   iii. Pelargonien
```

Unfortunately the output in LaTeX and EPUB is not what I expected. It is rendered more or less like a normal enumerated list. This can be an issue of EPUB and LaTeX. 

Maybe this should be mentioned in the documentation, if it is not possible have alphanumeric lists in LaTeX and EPUB in the near future. As alphanumeric lists are not mentioned at all in the documentation I marked this a minor issue.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1431
- Originally reported by: Jan Ulrich Hasecke
- Originally created at: 2014-03-25T09:23:36.489

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #1431

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1525,12 +1525,14 @@ class LaTeXTranslator(nodes.NodeVisitor):
                 return get_nested_level(node.parent)
 
         enum = "enum%s" % toRoman(get_nested_level(node)).lower()
+        enumnext = "enum%s" % toRoman(get_nested_level(node) + 1).lower()
         style = ENUMERATE_LIST_STYLE.get(get_enumtype(node))
 
         self.body.append('\\begin{enumerate}\n')
-        self.body.append('\\renewcommand{\\the%s}{%s{%s}}\n' % (enum, style, enum))
-        self.body.append('\\makeatletter\\renewcommand{\\p@%s}{%s\\the%s%s}\\makeatother\n' %
-                         (enum, node['prefix'], enum, node['suffix']))
+        self.body.append('\\def\\the%s{%s{%s}}\n' % (enum, style, enum))
+        self.body.append('\\def\\label%s{\\the%s.}\n' % (enum, enum))
+        self.body.append('\\makeatletter\\def\\p@%s{\\p@%s\\the%s.}\\makeatother\n' %
+                         (enumnext, enum, enum))
         if 'start' in node:
             self.body.append('\\setcounter{%s}{%d}\n' % (enum, node['start'] - 1))
         if self.table:
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -59,6 +59,14 @@ HYPERLINK_SUPPORT_NODES = (
     nodes.table,
     nodes.section,
 )
+ENUMERATE_LIST_STYLE = defaultdict(lambda: r'\arabic',
+                                   {
+                                       'arabic': r'\arabic',
+                                       'loweralpha': r'\alph',
+                                       'upperalpha': r'\Alph',
+                                       'lowerroman': r'\roman',
+                                       'upperroman': r'\Roman',
+                                   })  # type: Dict[unicode, unicode]
 
 DEFAULT_SETTINGS = {
     'latex_engine':    'pdflatex',
@@ -1498,6 +1506,15 @@ class LaTeXTranslator(nodes.NodeVisitor):
 
     def visit_enumerated_list(self, node):
         # type: (nodes.Node) -> None
+        def get_enumtype(node):
+            # type: (nodes.Node) -> unicode
+            enumtype = node.get('enumtype', 'arabic')
+            if 'alpha' in enumtype and 26 < node.get('start', 0) + len(node):
+                # fallback to arabic if alphabet counter overflows
+                enumtype = 'arabic'
+
+            return enumtype
+
         def get_nested_level(node):
             # type: (nodes.Node) -> int
             if node is None:
@@ -1507,10 +1524,15 @@ class LaTeXTranslator(nodes.NodeVisitor):
             else:
                 return get_nested_level(node.parent)
 
+        enum = "enum%s" % toRoman(get_nested_level(node)).lower()
+        style = ENUMERATE_LIST_STYLE.get(get_enumtype(node))
+
         self.body.append('\\begin{enumerate}\n')
+        self.body.append('\\renewcommand{\\the%s}{%s{%s}}\n' % (enum, style, enum))
+        self.body.append('\\makeatletter\\renewcommand{\\p@%s}{%s\\the%s%s}\\makeatother\n' %
+                         (enum, node['prefix'], enum, node['suffix']))
         if 'start' in node:
-            enum_depth = "enum%s" % toRoman(get_nested_level(node)).lower()
-            self.body.append('\\setcounter{%s}{%d}\n' % (enum_depth, node['start'] - 1))
+            self.body.append('\\setcounter{%s}{%d}\n' % (enum, node['start'] - 1))
         if self.table:
             self.table.has_problematic = True
 
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1530,9 +1530,10 @@ class LaTeXTranslator(nodes.NodeVisitor):
 
         self.body.append('\\begin{enumerate}\n')
         self.body.append('\\def\\the%s{%s{%s}}\n' % (enum, style, enum))
-        self.body.append('\\def\\label%s{\\the%s.}\n' % (enum, enum))
-        self.body.append('\\makeatletter\\def\\p@%s{\\p@%s\\the%s.}\\makeatother\n' %
-                         (enumnext, enum, enum))
+        self.body.append('\\def\\label%s{%s\\the%s %s}\n' % 
+                         (enum, node['prefix'], enum, node['suffix']))
+        self.body.append('\\makeatletter\\def\\p@%s{\\p@%s %s\\the%s %s}\\makeatother\n' %
+                         (enumnext, enum, node['prefix'], enum, node['suffix']))
         if 'start' in node:
             self.body.append('\\setcounter{%s}{%d}\n' % (enum, node['start'] - 1))
         if self.table:
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1510,7 +1510,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
 
         self.body.append('\\begin{enumerate}\n')
         self.body.append('\\def\\the%s{%s{%s}}\n' % (enum, style, enum))
-        self.body.append('\\def\\label%s{%s\\the%s %s}\n' % 
+        self.body.append('\\def\\label%s{%s\\the%s %s}\n' %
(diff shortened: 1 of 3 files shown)


sphinx/writers/latex.py
```diff
@@ -1500,9 +1517,13 @@ def get_nested_level(node):
             else:
                 return get_nested_level(node.parent)
 
+        enum_depth = "enum%s" % toRoman(get_nested_level(node)).lower()
+        enum_label = ENUMERATE_LABELS.get(get_enumtype(node))
+
         self.body.append('\\begin{enumerate}\n')
+        self.body.append('\\renewcommand{\\label%s}{%s%s{%s}%s}\n' %
+                         (enum_depth, node['prefix'], enum_label, enum_depth, node['suffix']))
```
jfbu: Some discussion is needed here because this will break the `\label/\ref` system. However out-of-the hat it is possible that Sphinx uses mainly `\hyperref` for hyperlinks, it doesn't rely on LaTeX's `\ref` being made hyperlink by package hyperref (I have volatile memory and should review latex.py first). Here is extract from LaTeX source code:

```
Enumeration is done with four counters: enumi, enumii, enumiii and enumiv,
where enumN controls the numbering of the Nth level enumeration. The label
is generated by the commands \labelenumi . . . \labelenumiv, which should be
defined by the document style. Note that \p@enumN\theenumN defines the output
of a \ref command. A typical definition might be:

\def\theenumii{\alph{enumii}}
\def\p@enumii{\theenumi}
\def\labelenumii{(\theenumii)}

which will print the labels as `(a)', `(b)', . . . and print a \ref as `3a'.
```

Thus it would be advisable to modify rather `\theenumi` to be what one wants (without the dot). If one leaves `\labelenumii` untouched for example then at second depth level `A. ` in source will become `(A)` in output (at least in English language and no additional package). WIth present PR, it would become `A. ` rather. And references from usage of `\ref` elsewhere in document would be wrong if a `\label` is located there. (but again, I am not sure writers/latex.py uses anything of `\label/\ref`).

jfbu: Another remark is that with present PR input such as this

```
a. first

b. second

c. third

   A. subfirst
   B. subsecond
   C. subthird

      #. subsub
      #. subsub
      #. subsub
```

will give in output at third level `1.`, `2.`,  `3.` whereas LaTeX's default (for English language at least) would be at third level to do `i.`, `ii.`, `iii.`.

jfbu (commented):
I believe it is better to modify `\theenumi` rather than `\labelenumi` etc..., and one should also adjust `\p@enumii` for second level (annoying because of `@`) for example

```
\renewcommand{\theenumii}{\Alph{enumii}}
\makeatletter\renewcommand{\p@enumii}{\theenumi.}\makeatother
```
This might not be needed for Sphinx links if it does not use regular `\ref`, but if not `\ref` is broken. In the case above that would create references as say `b.E` if top list is with alpha and nested one with Alpha.

jfbu:
About 

> This might not be needed for Sphinx links if it does not use regular  `\ref`, but if not `\ref` is broken. 

actually, sphinx.writers.latex uses extensively `\label/\hyperref` combination, but never `\label/\ref` (it does use `\autopageref`). And `\hyperref` requires an argument for the <strike>name</strike> used text of the link (`\hyperref[label of target]{text}`) , hence it is not important if `\ref{label of target}` would print a wrong text not in sync with actual text seen in PDF as only macro `\hyperref` is used and not `\ref`.

Nevertheless it would be better to get potentially `\ref` to work fine, because user may want to insert it via raw directive, or it could be used via some add-on LaTeX package.

Side note: the terminology in sphinx.latex.writers is potentially confusing to LaTeX person because LaTeXTranslator provides methods ``hypertarget()`` (uses LaTeX ``\label``)  and ``hyperlink()`` (uses hyperref's ``\hyperref``) but on the other hand hyperref package  provides macros which are similarly named:`\hypertarget`, resp. `\hyperlink` but they are **not** used by Sphinx LaTeX writer!

sphinx/writers/latex.py
jfbu: About my previous comment, output in PDF in `1.`, `2.`, `3.` at 3rd level from `#.` in source is same as in HTML, so arguably this is +1.

tk0miya:
Just updated with `\theenumi` and `\p@enumi`. I don't check `\label` and `\ref` system. But it seems working fine in Sphinx world.
@jfbu Could you check this please?

About terminology, I perfectly agreed with you. The responsibility of the `Translator` module is mapping doctree node to LaTeX representation. So each handlers are named as `visit_[nodename]` or `depart_[nodename]`.
On the other hand, `hypertarget()` and `hyperlink()` are helper function for these handlers. So they are mainly used for representation of LaTeX document. So they should be named as called in LaTeX world. +1 for renaming.

jfbu (changes requested):
Unfortunately the prefix business is no good. Something looking like this

```
\makeatletter\renewcommand{\p@enumii}{\theenumi.}\makeatother
```

is needed (this is only an example).

i.e. the `\p@enum...` macros give the _prefix_ i.e. the numbers or letters used at less deep level. It should not make reference to the current level.

jfbu:
Thus when we modify the rendering of a given level, this may require modifying prefix of deeper level. For example if default of LaTeX is to use <strike>brace</strike> parentheses, but we don't want them now that we enumerate our own way. I will work on this when I get time.

jfbu:
For info

```
~$ latexdef p@enumi

\p@enumi:
macro:->

~$ latexdef p@enumii

\p@enumii:
\long macro:->\theenumi 

~$ latexdef p@enumiii

\p@enumiii:
\long macro:->\theenumi (\theenumii )

~$ latexdef p@enumiv

\p@enumiv:
\long macro:->\p@enumiii \theenumiii 

~$ latexdef p@enumv

\p@enumv:
undefined
```

With usage `'maxlistdepth'` option, Sphinx defines more:

```
p@enumi has meaning macro:->
p@enumii has meaning \long macro:->\theenumi 
p@enumiii has meaning \long macro:->\theenumi (\theenumii )
p@enumiv has meaning \long macro:->\p@enumiii \theenumiii 
p@enumv has meaning macro:->\p@enumiv \theenumiv .
p@enumvi has meaning macro:->\p@enumv \theenumv .
p@enumvii has meaning macro:->\p@enumvi \theenumvi .
(etc...)
```

jfbu (approved):
@tk0miya I have pushed what I think is needed on LaTeX side. I did not know what `node['prefix']` and `node['suffix']` actually referred too, so I did not use them. I checked that `\label/\ref` works as expected in LaTeX world. Can you review this?

sphinx/writers/latex.py
```diff
@@ -1498,6 +1506,15 @@ def depart_bullet_list(self, node):
 
     def visit_enumerated_list(self, node):
         # type: (nodes.Node) -> None
+        def get_enumtype(node):
+            # type: (nodes.Node) -> unicode
+            enumtype = node.get('enumtype', 'arabic')
+            if 'alpha' in enumtype and 26 < node.get('start', 0) + len(node):
+                # fallback to arabic if alphabet counter overflows
+                enumtype = 'arabic'
+
```
jfbu, line 1495: there is package [http://www.ctan.org/pkg/alphalph](alphalph) which may be used for extended alphabetical indexing. `\alphalph{\value{counter}}` uses `a..z`, then `aa, ab, ..., az, ba, ..., zz` then `aaa, aab, ...` etc (sic). Unfortunately we need `\value`, so if we use package we could define macro `\sphinxalphalph` which adds the `\value` wrapper automatically, and then use `\sphinxalphalph`rather than `\alph` same for `\Alph`.

tk0miya:
Sorry for late. I confirmed. It works fine.

>I did not know what node['prefix'] and node['suffix'] actually referred too, so I did not use them.

What do you mean? Both came from by docutils. At present, all of `1.`, `1)` and `(1)` are rendered as `1.` on LaTeX output. It is better to support them.

jfbu:
> >    I did not know what node['prefix'] and node['suffix'] actually referred too, so I did not use them.
> 
> What do you mean? Both came from by docutils. At present, all of 1., 1) and (1) are rendered as 1. on LaTeX output. It is better to support them.

Of course, my bad. I will fix that.

sphinx/writers/latex.py
```diff
@@ -1507,10 +1524,18 @@ def get_nested_level(node):
             else:
                 return get_nested_level(node.parent)
 
+        enum = "enum%s" % toRoman(get_nested_level(node)).lower()
+        enumnext = "enum%s" % toRoman(get_nested_level(node) + 1).lower()
+        style = ENUMERATE_LIST_STYLE.get(get_enumtype(node))
+
         self.body.append('\\begin{enumerate}\n')
+        self.body.append('\\def\\the%s{%s{%s}}\n' % (enum, style, enum))
+        self.body.append('\\def\\label%s{%s\\the%s %s}\n' %
```
jfbu: I use ``\\the%s %s`` and not ``\the%s%s`` in case suffix is a letter. I know this is perhaps not possible. Anyhow, for LaTeX the space after the control sequence is gobbled.

sphinx/writers/latex.py
```diff
@@ -1507,10 +1524,18 @@ def get_nested_level(node):
...
 
+        enum = "enum%s" % toRoman(get_nested_level(node)).lower()
+        enumnext = "enum%s" % toRoman(get_nested_level(node) + 1).lower()
+        style = ENUMERATE_LIST_STYLE.get(get_enumtype(node))
+
         self.body.append('\\begin{enumerate}\n')
+        self.body.append('\\def\\the%s{%s{%s}}\n' % (enum, style, enum))
+        self.body.append('\\def\\label%s{%s\\the%s %s}\n' % 
+                         (enum, node['prefix'], enum, node['suffix']))
+        self.body.append('\\makeatletter\\def\\p@%s{\\p@%s %s\\the%s %s}\\makeatother\n' %
```
jfbu: similar spaces here, which do not make it to PDF, they are gobbled by LaTeX.

jfbu (commented):
sorry for overlooking those `prefix` and `suffix` in previous commit.

sphinx/writers/latex.py
tk0miya: alphalph list is only generated by following markup. docutils doesn't support `aa.` as enum-list notation.
```
a.
b.
...
#.
#.
```

I guess alph-alph might not be used often. So it's okay to fallback arabic one to me.

sphinx/writers/latex.py
tk0miya: +1

tk0miya (commented):
LGTM!

tk0miya:
Merged!
@jfbu Thank you always. I really appreciate your help :-)

jfbu:
@tk0miya thanks, next time I need to be more careful about flake8, sorry about that!

PR #5004, opened 2018-05-22
Fix #4648: latex: Now "rubric" elements are rendered as unnumbered section title

### LINKED ISSUE #4648: "rubric" should create an unnumbered \paragraph in latex mode
I use `.. rubric` to create a heading with no number and no entry in TOC. It works fine in HTML, but in the latex output I get something like:

```
1.1.3 Normal heading

Lorem ipsum...

1.1.3.0.0.1 "Rubric" heading
```

which makes no sense. I wanted it unnumbered and got a full set of numbers.

I think `.. rubric` should just create `\paragraph*{...}` instead of `\paragraph{...}`. (Changing "secnumdepth" is not an option, as I have other paragraphs (and subparagraphs) that should be numbered.)

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4648

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1252,7 +1252,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
         if len(node.children) == 1 and node.children[0].astext() in \
            ('Footnotes', _('Footnotes')):
             raise nodes.SkipNode
-        self.body.append('\\paragraph{')
+        self.body.append('\\subsubsection*{')
         self.context.append('}\n')
         self.in_title = 1
 
--- a/CHANGES
+++ b/CHANGES
@@ -33,6 +33,7 @@ Incompatible changes
   :confval:`viewcode_follow_imported_members` (refs: #4035)
 * #1857: latex: :confval:`latex_show_pagerefs` does not add pagerefs for
   citations
+* #4648: latex: Now "rubric" elements are rendered as unnumbered section title
 
 Deprecated
 ----------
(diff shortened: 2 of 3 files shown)


sphinx/writers/latex.py
```diff
@@ -1252,7 +1252,7 @@ def visit_rubric(self, node):
         if len(node.children) == 1 and node.children[0].astext() in \
            ('Footnotes', _('Footnotes')):
             raise nodes.SkipNode
-        self.body.append('\\paragraph{')
+        self.body.append('\\paragraph*{')
```
tk0miya: In this PR, I did not add any migration pass. I suppose nobody want to keep the behavior.

tk0miya: Note: docutils uses `\subsubsection*` to rubric.
http://repo.or.cz/docutils.git/blob/HEAD:/docutils/docutils/writers/latex2e/__init__.py#l634

jfbu: +1 for `\subsubsection*`.

jfbu (approved):
I think `\subsubsection*` would make better sense than `\ paragraph*`. Of course either one is better than `\paragraph`.

sphinx/writers/latex.py
tk0miya: updated

jfbu (approved):
LGTM

tk0miya:
Thank you for quick reviewing!

PR #5005, opened 2018-05-22
Fix #3606: MathJax should be loaded with async attribute

### LINKED ISSUE #3606: MathJax should be loaded with async attribute
It is possible and seems recommended (see the [MathJax CDN change announcement](https://www.mathjax.org/cdn-shutting-down/)) to use the ``async`` attribute for the loading of MathJaX from CDN resource.

This is not done currently,and as explained https://github.com/sphinx-doc/sphinx/pull/3602#issuecomment-291370524:

> But, unfortunately, we don't have an interface to do that.
> sphinx.ext.mathjax uses app.add_javascript() API to add mathjax.js to each HTML file. But this API does not have any options to add async attribute.
> So we have to improve the API as a first step.
> 

From [w3schools page](https://www.w3schools.com/tags/att_script_async.asp), HTML5 should use simply ``async`` but XHTML should use the ``async="async"`` syntax.

### Environment info
- Sphinx version: 1.5.5

Relates to #3602

tk0miya:
### Feature or Bugfix
- Feature
- Refactoring

### Purpose
- refs: #3606
- As discussed in #4595, I renamed `app.add_javascript()` to `add.add_js_file()`
    -  I give long deprecation period to the API because it has been used longly.

### DIFF
--- a/doc/extdev/appapi.rst
+++ b/doc/extdev/appapi.rst
@@ -71,7 +71,7 @@ package.
 
 .. automethod:: Sphinx.add_post_transform(transform)
 
-.. automethod:: Sphinx.add_javascript(filename, **kwargs)
+.. automethod:: Sphinx.add_js_file(filename, **kwargs)
 
 .. automethod:: Sphinx.add_stylesheet(filename, alternate=None, title=None)
 
--- a/doc/extdev/appapi.rst
+++ b/doc/extdev/appapi.rst
@@ -71,7 +71,7 @@ package.
 
 .. automethod:: Sphinx.add_post_transform(transform)
 
-.. automethod:: Sphinx.add_javascript(filename)
+.. automethod:: Sphinx.add_javascript(filename, **kwargs)
 
 .. automethod:: Sphinx.add_stylesheet(filename, alternate=None, title=None)
 
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -998,25 +998,36 @@ class Sphinx(object):
         """
         self.registry.add_post_transform(transform)
 
-    def add_javascript(self, filename):
-        # type: (unicode) -> None
+    def add_javascript(self, filename, **kwargs):
+        # type: (unicode, **unicode) -> None
         """Register a JavaScript file to include in the HTML output.
 
         Add *filename* to the list of JavaScript files that the default HTML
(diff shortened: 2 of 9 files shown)


sphinx/application.py
```diff
@@ -998,25 +998,45 @@ def add_post_transform(self, transform):
...
+        from sphinx.builders.html import StandaloneHTMLBuilder, JavaScript
         if '://' in filename:
-            StandaloneHTMLBuilder.script_files.append(filename)
+            js = JavaScript(filename, **kwargs)
         else:
-            StandaloneHTMLBuilder.script_files.append(
-                posixpath.join('_static', filename))
+            js = JavaScript(posixpath.join('_static', filename), **kwargs)
+
+        StandaloneHTMLBuilder.script_files.append(js)
```
tk0miya, line 1039: This still uses class variable as a storage. I'll refactor it in another PR.

sphinx/builders/html.py
```diff
@@ -1479,6 +1498,31 @@ def convert_html_css_files(app, config):
     config.html_css_files = html_css_files  # type: ignore
 
 
+def setup_js_tag_helper(app, pagename, templatexname, context, doctree):
+    # type: (Sphinx, unicode, unicode, Dict, nodes.Node) -> None
+    """Set up js_tag() template helper.
+
+    .. note:: This set up function is added to keep compatibility with webhelper.
```
tk0miya, line 1512: Unfortunately, adding this to `HTMLBuilder.handle_page()` because websupport builder overrides it. It means websupport builder and HTML builder are strongly coupled.
We have to fix it later.

doc/extdev/appapi.rst
```diff
@@ -71,7 +71,7 @@ package.
 
 .. automethod:: Sphinx.add_post_transform(transform)
 
-.. automethod:: Sphinx.add_javascript(filename)
+.. automethod:: Sphinx.add_js_file(filename, **kwargs)
 
 .. automethod:: Sphinx.add_stylesheet(filename, alternate=None, title=None)
```
tk0miya: Note: this line was wrong. I'll fix it in another PR.

tk0miya: see #5013

PR #5006, opened 2018-05-23
C++: parse pack expansions in function calls.

other-mickk:
Currently the parser has support for pack expansions in e.g. angle bracket lists:

```rST
.. cpp:var:: template<typename... Args> int angled = foo<args...>
```

This adds support for parsing pack expansions in function calls:

```rST
.. cpp:function:: template<typename... Args> void foo(Args... args)

.. cpp:function:: template<typename... Args> decltype( foo(args...) ) bar(Args... args)

    N.b. not valid C++ (args is not in scope before the declaration of the parameter pack),
    but Sphinx links correctly.
```

ping @jakobandersen


jakobandersen:
Great, thanks! As it's basically a bugfix I have rebased it to the 1.7 branch.

PR #5024, opened 2018-05-31
Fix #5022: latex: crashed with docutils package provided by Debian/Ubuntu

### LINKED ISSUE #5022: Problem with 1.7.5 on Travis CI ("No module named roman")
Running Sphinx v 1.7.5 on Travis I am seeing this error:

> "Could not import extension sphinx.builders.latex (exception: No module named roman)"

Ubuntu 14.04 ships with a very old version of Sphinx (it has problems parsing the C++ declarations in my docs) so my CI script installs a new version of Sphinx via pip.

```
        sudo apt-get -qq update
        sudo apt-get install doxygen python-docutils
        sudo apt-get remove python-requests python-openssl
        sudo pip install requests pyopenssl
        sudo pip install sphinx
```

The removal/reinstall of requests and python-openssl is to workaround the bug described in 
https://stackoverflow.com/questions/32779919/no-module-named-for-requests

We are not actually trying to build a PDF during CI, but it looks like the missing module causes `sphinx-build` to exit with a non-zero error code which causes the whole build to fail.

Something I do not understand about this error is that the Ubuntu `python-roman` package is already installed (it is a dependency of `python-docutils`).

#### Error logs / results
https://travis-ci.org/randombit/botan/jobs/384926972

In the build logs the pip/apt-get operations are run from the subscript `./src/scripts/ci/setup_travis.sh` at line 474

#### Expected results
Version 1.7.4 and earlier work fine. I only investigated this after my build suddenly broke. After I set pip to install ==1.7.4 my build worked again https://travis-ci.org/randombit/botan/builds/384949254

- The branch https://github.com/randombit/botan/tree/fix-docs-build strips out all the non-docs CI builds so should be easy to reproduce the problem.

### Environment info
- OS: Ubuntu 14.04 (Travis CI)
- Python version: 2.7.5
- Sphinx version: 1.7.5

Let me know if there is any additional info I can provide, and thanks for this great tool.

tk0miya:
### Feature or Bugfix
- Bugfix ?

### Purpose
- refs: #5022

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -18,7 +18,6 @@ from collections import defaultdict
 from os import path
 
 from docutils import nodes, writers
-from docutils.utils.roman import toRoman
 from docutils.writers.latex2e import Babel
 from six import itervalues, text_type
 
@@ -31,6 +30,12 @@ from sphinx.util.i18n import format_date
 from sphinx.util.nodes import clean_astext
 from sphinx.util.template import LaTeXRenderer
 from sphinx.util.texescape import tex_escape_map, tex_replace_map
+
+try:
+    from docutils.utils.roman import toRoman
+except ImportError:
+    # In Debain/Ubuntu, roman package is provided as roman, not as docutils.utils.roman
+    from roman import toRoman
 
 if False:
     # For type annotation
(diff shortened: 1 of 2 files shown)


sphinx/writers/latex.py
```diff
@@ -32,6 +31,11 @@
 from sphinx.util.template import LaTeXRenderer
 from sphinx.util.texescape import tex_escape_map, tex_replace_map
 
+try:
+    from docutils.utils.roman import toRoman
+except ImportError:
+    from roman import toRoman
```
tk0miya: I have to add comment why we are doing such importing.

PR #5030, opened 2018-06-01
Fix sphinx.testing uses deprecated pytest API; Node.get_marker(name)

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Since pytest-3.6 `Node.get_marker(name)` was marked as deprecated (see https://docs.pytest.org/en/latest/mark.html#marker-revamp-and-iteration )
- This follows the migration rule


tk0miya:
Updated again.

PR #5040, opened 2018-06-05
Fix: #5037

### LINKED ISSUE #5037: LaTeX \sphinxupquote{} breaks in Russian
Exactly the same issue as #4978 but for Russian with preamble
```
\documentclass[letterpaper,10pt,russian]{sphinxmanual}
```

jfbu:
Subject: fix #5037

- Bugfix

The first commit tries to update the list of languages. Only the " and the = needs special treatments for not causing problems with `\sphinxupquote` and similar contexts. Info picked from LaTeX-babel documentation and docutils.writers.latex2e.

The second commit replaces this by generic TeX code, so that there is no need to maintain the language list in future.

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -200,42 +200,10 @@ class ExtBabel(Babel):
 
     def get_shorthandoff(self):
         # type: () -> unicode
-        shortlang = self.language.split('_')[0]
-        if shortlang in ('br', 'breton',
-                         'bg', 'bulgarian',
-                         'ca', 'catalan',
-                         'cs', 'czech',
-                         'da', 'danish',
-                         'de', 'ngerman',
-                         'de-1901', 'german',
-                         'de-AT', 'naustrian',
-                         'de-AT-1901', 'austrian',
-                         'es', 'spanish',
-                         'et', 'estonian',
-                         'eu', 'basque',
-                         'hsb', 'uppersorbian',
-                         'gl', 'galician',
-                         'is', 'icelandic',
-                         'it', 'italian',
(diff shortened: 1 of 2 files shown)


sphinx/writers/latex.py
```diff
@@ -200,15 +200,10 @@ def __init__(self, language_code, use_polyglossia=False):
...
-                         'italian', 'pt-BR', 'brazil'):
-            return '\\ifnum\\catcode`\\"=\\active\\shorthandoff{"}\\fi'
-        elif shortlang in ('tr', 'turkish'):
-            # memo: if ever Sphinx starts supporting 'Latin', do as for Turkish
-            return '\\ifnum\\catcode`\\=\\string=\\active\\shorthandoff{=}\\fi'
-        return ''
+        return ('\\ifdefined\\shorthandoff\n'
+                '  \\ifnum\\catcode`\\=\\string=\\active\\shorthandoff{=}\\fi\n'
+                '  \\ifnum\\catcode`\\"=\\active\\shorthandoff{"}\\fi\n'
+                '\\fi')
```
tk0miya, line 206: Awesome! In master branch, we can embed this to template string (`BEGIN_DOC`) directly. `shorthandoff` variable is no longer needed.

tk0miya (approved):

sphinx/writers/latex.py
jfbu: Thanks for quick review @tk0miya I will make PR on master branch afterwards. Now merging into 1.7 branch.

jfbu: I made PR at #5045

Issue #4877, opened 2018-04-21
latex: Do not use Babel.get_language().


tk0miya:
### Problem
- The method was marked as obsoleted. And it was kept only for Sphinx.

```
    def get_language(self):
        # Obsolete, kept for backwards compatibility with Sphinx
        return self.language
```

- I checked docutils-0.11 which is oldest one supported by current release. It contains above comment.

#### Expected results
- Do not use the method.

jfbu:
+1

tk0miya:
Fixed by #4940

PR #4000, opened 2017-08-13
latex writer: move over template parts into latex.tex_t

GabrielL:
In order to be able to have custom templates for latex that have content at the beginning of the document, we need to have it in the template, and not in the code.

PR #5045, opened 2018-06-05
Remove usage of internal shorthandoff variable from LaTeX writer

jfbu:
Subject: brings simplicity by removing shorthandoff internal variable from LaTeX writer

Neither bugfix nor feature, but refactoring.

Pros: simplified code

Cons: now, some LaTeX code gets exported to all Sphinx produced LaTeX documents ; in the past it was incorporated only language conf val was set to certain languages. Code was not included for Japanese, or when latex_engine was lualatex xelatex and document used polyglossia, or when engine was pdflatex but certain languages were used.

But the TeX code is crafted to be usable in all circumstances (platex incorporates e-TeX extensions, hence `\ifdefined` can be used).

added in edit: sphinx.sty has had for some time already numerous usages of `\numexpr`, `\dimexpr` and a few `\ifdefined`, so e-TeX has been mandatory for all engines. I have been told on TeXLive list (I don't know for MikTeX but expect it to be same) that platex includes e-TeX since 2011. For pdflatex it is since 2004. For xelatex and lualatex they always incorporated e-TeX.

### Relates
- #5040 

see https://github.com/sphinx-doc/sphinx/pull/5040#discussion_r193122150

**attention** this PR includes no deprecation and simply removes the internal variable. Maybe this is bad...

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -26,7 +26,7 @@ from six import itervalues, text_type
 from sphinx import addnodes
 from sphinx import highlighting
 from sphinx.builders.latex.nodes import captioned_literal_block, footnotetext
-from sphinx.deprecation import RemovedInSphinx30Warning, RemovedInSphinx20Warning
+from sphinx.deprecation import RemovedInSphinx30Warning
 from sphinx.errors import SphinxError
 from sphinx.locale import admonitionlabels, _, __
 from sphinx.util import split_into, logging
@@ -216,7 +216,7 @@ class ExtBabel(Babel):
     def get_shorthandoff(self):
         # type: () -> unicode
         warnings.warn('ExtBabel.get_shorthandoff() is deprecated.',
-                      RemovedInSphinx20Warning)
+                      RemovedInSphinx30Warning)
         return SHORTHANDOFF
 
     def uses_cyrillic(self):
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -44,7 +44,10 @@ logger = logging.getLogger(__name__)
 
 BEGIN_DOC = r'''
 \begin{document}
-%(shorthandoff)s
+\ifdefined\shorthandoff
+  \ifnum\catcode`\=\string=\active\shorthandoff{=}\fi
+  \ifnum\catcode`\"=\active\shorthandoff{"}\fi
+\fi
 %(maketitle)s
 %(tableofcontents)s
 '''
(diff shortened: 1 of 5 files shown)


sphinx/writers/latex.py
```diff
@@ -217,13 +219,6 @@ def __init__(self, language_code, use_polyglossia=False):
         self.supported = True
         super(ExtBabel, self).__init__(language_code or '')
 
-    def get_shorthandoff(self):
```
tk0miya: Please follow our deprecation rule. I know this is minor and internal API. But sometimes extensions uses such an API. So let deprecate this before removing.

* Add deprecation warnings `warnings.warn()`
* Update CHANGES and deprecation list on `doc/extdev/index.rst`

tk0miya (changes requested):
LGTM with nits

jfbu:
@tk0miya updated. I set removal of ``ExtBabel.get_shorthandoff()`` at 2.0, perhaps should be 3.0 ?

I had not noticed during merging of #5040 that there was also your d17f3955a5546a9eb30ee9161946f8bb900441c5 which fixed #4877. This partially reverts it.

Should also ``latex_elements['shorthandoff']`` get deprecated before some future removal? Currently I have left it untouched but added `SHORTHANDOFF` constant.

tk0miya:
>I set removal of ExtBabel.get_shorthandoff() at 2.0, perhaps should be 3.0 ?

In our deprecation policy, it should be removed after a release. So 3.0 would be better.
http://www.sphinx-doc.org/en/master/devguide.html#deprecation-policy

>Should also latex_elements['shorthandoff'] get deprecated before some future removal? Currently I have left it untouched but added SHORTHANDOFF constant.

`shorthandoff` is only used in `BEGIN_DOC` variable. So I think the removal does not effect to templates. In addition, the template variable will be replaced with empty string even if user templates refers it. Hence the removal of the key is okay to me.

jfbu:
> 
> 
> >     I set removal of ExtBabel.get_shorthandoff() at 2.0, perhaps should be 3.0 ?
> 
> 
> In our deprecation policy, it should be removed after a release. So 3.0 would be better.
> http://www.sphinx-doc.org/en/master/devguide.html#deprecation-policy
> 

Done

> >     Should also latex_elements['shorthandoff'] get deprecated before some future removal? Currently I have left it untouched but added SHORTHANDOFF constant.
> 
> 
> shorthandoff is only used in BEGIN_DOC variable. So I think the removal does not effect to templates. In addition, the template variable will be replaced with empty string even if user templates refers it. Hence the removal of the key is okay to me.

I simply removed mention of it from doc/usage/configuration.rst.

Memo #4000 is impacted by this. It could be possible to move SHORTHANDOFF contents entirely to LaTeX template, then we can remove `latex_elements['shorthandoff']`. Currently I keep it to keep former behaviour that for example Japanese does nothing special.

tk0miya (approved):
LGTM. Thank you for update!

jfbu:
@tk0miya thanks,  merging.

PR #5058, opened 2018-06-08
Link to cdnjs was incorrect

jodemaey:
Subject: Link to cdnjs was incorrect

Link to cdnjs was incorrect in math support page


tk0miya:
Thanks a lot!

PR #5012, opened 2018-05-24
Fix #5009: latex: a label for table is vanished if table does not have a caption

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5009

PR #5060, opened 2018-06-08
LaTeX, follow-up to PR #5012, avoid label separated by pagebreak

jfbu:
PR #5012 took care of this by anticipation for longtable, but the problem can also arise
with tabular/tabulary and is much easier to fix.

### Relates
- #5012


tk0miya (approved):
LGTM

jfbu:
@tk0miya thanks for quick reviewing!

PR #5063, opened 2018-06-09
Search: make a difference between fullname and dispname

jakobandersen:
Subject: search should use ``dispname`` instead of ``fullname`` from ``Domain.get_objects()``

### Feature or Bugfix
- Bugfix

### Purpose
Often there is no difference between the full name and the display name of objects, but when there is the search is still based on the full name instead of the display name.

### Relates
Needed for PR #5062 to use/display anonymous names as ``[anonymous]`` in search, instead of ``@someName``.

### DIFF
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -331,13 +331,13 @@ class IndexBuilder(object):
         for domainname, domain in sorted(iteritems(self.env.domains)):
             for fullname, dispname, type, docname, anchor, prio in \
                     sorted(domain.get_objects()):
-                # XXX use dispname?
                 if docname not in fn2index:
                     continue
                 if prio < 0:
                     continue
                 fullname = htmlescape(fullname)
-                prefix, name = rpartition(fullname, '.')
+                dispname = htmlescape(dispname)
+                prefix, name = rpartition(dispname, '.')
                 pdict = rv.setdefault(prefix, {})
                 try:
                     typeindex = otypes[domainname, type]
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -161,7 +161,7 @@ def test_IndexBuilder():
         'docnames': ('docname', 'docname2'),
         'envversion': '1.0',
         'filenames': ['filename', 'filename2'],
-        'objects': {'': {'objname': (0, 0, 1, '#anchor')}},
+        'objects': {'': {'objdispname': (0, 0, 1, '#anchor')}},
         'objnames': {0: ('dummy', 'objtype', 'objtype')},
         'objtypes': {0: 'dummy:objtype'},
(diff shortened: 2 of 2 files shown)


jakobandersen:
Could this break something not covered by the tests? (I'm not familiar with the search module)

tk0miya:
The latest release of mypy breaks out testings...
Could you rebase this onto HEAD of master please?

jakobandersen:
Sure, done, and the tests look good now.

tk0miya (approved):
+0; I don't know purpose of dispname. So I don't opposite to this.

tk0miya:
@shimizukawa do you have any opinion for this?

sphinx/search/__init__.py
```diff
@@ -331,13 +331,13 @@ def get_objects(self, fn2index):
...
                     sorted(domain.get_objects()):
-                # XXX use dispname?
                 if docname not in fn2index:
                     continue
                 if prio < 0:
                     continue
                 fullname = htmlescape(fullname)
-                prefix, name = rpartition(fullname, '.')
+                dispname = htmlescape(dispname)
+                prefix, name = rpartition(dispname, '.')
```
tk0miya, line 340: This might be an incompatible change. So we should announce this in CHANGES.

shimizukawa:
https://github.com/sphinx-doc/sphinx/blob/3d6c7d147d535a85facccb2c13c8ccc3fdfa2d27/sphinx/domains/__init__.py#L311-L322

IMO, for searching/linking purpose,  `dispname` is better choice.

@jakobandersen Please give us an example on how this usage and display change with this PR.

jakobandersen:
Sure. The only domains I could find in Sphinx that yields different strings for "fullname" and "dispname" from ``get_objects()`` are ``std`` and now ``cpp``. I'm not sure how to trigger differences with the ``std`` domain, for ``cpp`` we can use the Sphinx website (with #5100 applied):
On the master version:
- Searching for ``data`` we get (among other results):
  - ``Data::@data (C++ union, in Domains)``
  - ``Data::@data::a (C++ member, in Domains)``
  - ``Data::@data::b (C++ member, in Domains)``
- Searching for ``anon`` yields no results.

With this PR:
- Searching for ``data`` we essentially get the same results, but the three lines are rendered as:
  - ``Data::[anonymous] (C++ union, in Domains)``
  - ``Data::[anonymous]::a (C++ member, in Domains)``
  - ``Data::[anonymous]::b (C++ member, in Domains)``
- Searching for ``anon`` now yields exactly those three lines.

tk0miya:
ping @shimizukawa; any comments?
If no objection, I'd like to merge this to 1.8.
Please let me know your opinion.

tk0miya:
It seems no comment. So I'm merging this.

PR #5072, opened 2018-06-11
Builder.build: save env also with only new documents

blueyed:
Saving/pickling the environment is necessary for intersphinx to have its
cache saved.

Currently a new build would fetch the inventories, but the cache would
not get saved.

This appears to break no existing test, but should probably get a new
one then?

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -372,14 +372,14 @@ class Builder(object):
         else:
             logger.info(__('none found'))
 
-        if updated_docnames:
-            # save the environment
-            from sphinx.application import ENV_PICKLE_FILENAME
-            logger.info(bold(__('pickling environment... ')), nonl=True)
-            with open(path.join(self.doctreedir, ENV_PICKLE_FILENAME), 'wb') as f:
-                pickle.dump(self.env, f, pickle.HIGHEST_PROTOCOL)
-            logger.info(__('done'))
+        # save the environment
+        from sphinx.application import ENV_PICKLE_FILENAME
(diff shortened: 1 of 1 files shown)


sphinx/builders/__init__.py
```diff
@@ -390,6 +383,13 @@ def build(self, docnames, summary=None, method='update'):
                 logger.info(bold(__('no targets are out of date.')))
                 return
 
+        # save the environment
+        from sphinx.application import ENV_PICKLE_FILENAME
+        logger.info(bold(__('pickling environment... ')), nonl=True)
+        with open(path.join(self.doctreedir, ENV_PICKLE_FILENAME), 'wb') as f:
+            pickle.dump(self.env, f, pickle.HIGHEST_PROTOCOL)
+        logger.info(__('done'))
+
```
tk0miya: It would be better to move this block before the if-block to keep the processing order. At present, this change invokes the block after consistency-check phase.

tk0miya (changes requested):
+0; essentially, the inventory of interxsphinx is not a environment data. So it should be moved out of environment object. But this change is a bit useful for other extensions. So I don't object to this.

sphinx/builders/__init__.py
blueyed: Done.

blueyed:
Thanks for the review, addressed your comment.

tk0miya (approved):

tk0miya:
Thank you for update. I'll merge this within a week if no objection.

tk0miya:
Merged.

Issue #5081, opened 2018-06-13
Translation doesn't seem to work


dombre:
I'm failing to translate a simple sphinx project and missing apparently some steps not mentioned in the documentation.

#### Procedure to reproduce the problem

Step 1: Create a project with sphinx-quickstart
```
sphinx-quickstart -q -p "Test project" -a "Test author"
```
Step 2: Follow the guide mentioned here: http://www.sphinx-doc.org/en/master/intl.html#quick-guide
I also tried to set the configuration options (`locale_dirs` and `gettext_compact`) in conf.py but that also made no difference.
```
make gettext
sphinx-intl update -p _build/gettext -l de
```
Step 3: translated the locales\de\LC_MESSAGES\index.po file
I tried it manually with a text editor but also with poedit which will create the mo file automatically.

Step 4: generate the translated project
```
set SPHINXOPTS=-D language='de'
make.bat html
```

#### Expected results
After the steps mentioned above I would expect a translated HTML document at `_build\html`. 

But I only get the source version without any translation. So apparently I'm missing a step which I cannot find in the documentaion.

#### Error logs / results
After Step 4 I get the following output:

```
X:\Temp\test6>make.bat html
Running Sphinx v1.7.5
loading translations ['de']... not available for built-in messages
making output directory...
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: 1 added, 0 changed, 0 removed
reading sources... [100%] index
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index
generating indices... genindex
writing additional pages... search
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded.
```

My thoughts on the output:

- The warning `loading translations ['de']... not available for built-in messages` shouldn't be a problem, I need to include the sphinx.po files (see e.g. https://github.com/sphinx-doc/sphinx/issues/1242)
- The wanring `building [mo]: targets for 0 po files that are out of date` seems more critical to me

Can I increase the loglevel of the build procedure? Or do you already see a step I'm missing? 

### Environment info
- OS: Windows 10
- Python version: 2.7.9
- Sphinx version: 1.7.5 (installed with pip)

dombre:
Found my error. Instead of 

```
> set SPHINXOPTS=-D language='de'
> .\make.bat html
```

[...]

PR #5083, opened 2018-06-13
Fixed internationalization quick guide

dombre:
### Feature or Bugfix
- Bugfix in the documentation

### Purpose
The internationalization quick guide didn't work for me on a Windows 10 machine.

### Detail
There were additional apostrophs in the commands I had to remove. The commands now work in cmd.exe and powershell on my Windows 10 machine.

Can someone else double check, if this is the right fix? Or is it a system issue/difference on my site?

I have also verified the GNU make example, and this works as expected on my linux system.

### Relates
- #5081

### DIFF
--- a/doc/intl.rst
+++ b/doc/intl.rst
@@ -123,14 +123,14 @@ This section describe an easy way to translate with sphinx-intl.
 
    .. code-block:: console
 
-      > set SPHINXOPTS=-D language='de'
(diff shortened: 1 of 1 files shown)


tk0miya:
@shimizukawa could you review this please?

shimizukawa (approved):
Confirmed. Thanks!

PR #5090, opened 2018-06-13
BuildDoc: link verbosity to distutils'

blueyed:
This preserves the current behavior (using 0), but picks up distutils' `-v`
and `-q` options.

### DIFF
--- a/sphinx/setup_command.py
+++ b/sphinx/setup_command.py
@@ -105,7 +105,8 @@ class BuildDoc(Command):
         self.config_dir = None  # type: unicode
         self.link_index = False
         self.copyright = ''
-        self.verbosity = 0
+        # Link verbosity to distutils' (which uses 1 by default).
+        self.verbosity = self.distribution.verbose - 1
         self.traceback = False
 
     def _guess_source_dir(self):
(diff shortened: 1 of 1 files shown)


sphinx/setup_command.py
```diff
@@ -105,7 +105,8 @@ def initialize_options(self):
         self.config_dir = None  # type: unicode
         self.link_index = False
         self.copyright = ''
-        self.verbosity = 0
+        # Link verbosity to distutils' (which uses 1 by default).
+        self.verbosity = self.distribution.verbose - 1
```
tk0miya, line 109: Are `self.distribution` and `self.distribution.verbose` public API? It seems there are no document for them.

tk0miya, line 109: mypy warns this line as following. Could you check this please?:
>sphinx/setup_command.py:109:26: error: "BuildDoc" has no attribute "distribution"
https://travis-ci.org/sphinx-doc/sphinx/jobs/391888857#L493

tk0miya (changes requested):

sphinx/setup_command.py
tk0miya: It seems distlib provides `Command.distribution` and `Distribution.verbose` since py27 (I just checked in GitHub).
https://github.com/python/cpython/blob/master/Lib/distutils/cmd.py#L61
https://github.com/python/cpython/blob/3.5/Lib/distutils/dist.py#L134

Unfortunately they are not documented. But I suppose it's okay.
https://docs.python.org/3/distutils/apiref.html

tk0miya:
Just merged (via #8001). Thank you for your contribution!

PR #5093, opened 2018-06-15
Delete the illegal character

flycrane01:
Subject: <Delete the illegal character>

- Bugfix

### DIFF
--- a/doc/usage/installation.rst
+++ b/doc/usage/installation.rst
@@ -110,7 +110,7 @@ Windows
 
 Most Windows users do not have Python installed by default, so we begin with
 the installation of Python itself.  If you are unsure, open the *Command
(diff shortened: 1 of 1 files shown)


doc/usage/installation.rst
```diff
@@ -110,7 +110,7 @@ Windows
 
 Most Windows users do not have Python installed by default, so we begin with
 the installation of Python itself.  If you are unsure, open the *Command
-Prompt* (:kbd:`⊞Win-r` and type :command:`cmd`).  Once the command prompt is
```
tk0miya, line 113: I seems this character imitates logo of microsoft. So I think this is not illegal. Is there any problem?
http://www.fileformat.info/info/unicode/char/229e/index.htm

AFAIK, it is also used in Wikipedia.
https://en.wikipedia.org/wiki/Windows_key

PR #5094, opened 2018-06-15
Fix #5066: html: "relations" sidebar is not shown by default

### LINKED ISSUE #5066: sphinx-quickstart creates non-satisfying html default set-up with Alabaster
I tested 1.7.5 and current master, both consider ``'alabaster'`` to be default HTML theme, but the `conf.py` is ill-configured for them and creates un-navigable documents.

For example, this:

capture d ecran 2018-06-09 a 15 27 38

with an `index.rst` like this

```rst
.. toctree::
   :maxdepth: 2

   foo1
   foo2
```

There is no link to go back to table of contents or to foo1 page.

#### Expected results

Things should work out-of-the box. It does not make sense to make default ``alabaster`` and not provide correct set-up to use it.

### Environment info
- OS: Mac
- Python version: 3.6.5
- Sphinx version: 1.7.5 and master
- Alabaster theme 0.7.10

Relates:

#3830, #4496, #3849, #3987, #4002 and probably others. Edit: #4177

I know this has been issued many times and I did personally at least two commits to fix it but it always end up overridden in some way. So I would really like this to be fixed once and for all.

Update: later investigation in my comments below reminded me #4177 should have fixed this, but there is a problem remaining when theme does not provide sidebars config in theme.conf.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5066

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,7 @@ Bugs fixed
 * #5019: autodoc: crashed by Form Feed Character
 * #5032: autodoc: loses the first staticmethod parameter for old styled classes
 * #5036: quickstart: Typing Ctrl-U clears the whole of line
+* #5066: html: "relations" sidebar is not shown by default
 
 Testing
 --------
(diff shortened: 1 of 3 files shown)


jfbu (approved):
Works fine in my testing, LGTM!

tk0miya:
Thank you for reviewing!

PR #5102, opened 2018-06-17
doc: Use badge to show latest release

tk0miya:
### Feature or Bugfix
- Docs

### Purpose
- Under current branch model, the unreleased version is always shown on document.
- With this change, latest version is always shown as badge!


shimizukawa (approved):
LGTM!

PR #4101, opened 2017-10-01
Sphinx apidoc extension

stephenfin:
Subject: Make 'sphinx-apidoc' an extension

### Feature or Bugfix
- Feature

### Purpose
- This is an alternative approach to #1135. Rather than modifying the 'build_sphinx' setuptools extension, we allow 'sphinx-apidoc' to be run as an extension.

  This is very similar to what we do with 'sphinx-autogen', which can be run automatically using the `autosummary_generate` configuration option instead.

  Not all configuration options are exposed - I wanted some insight into which of these I should expose before doing so.

  My eventual goal would be to remove both the `sphinx-apidoc` and `sphinx-autogen` binaries at some point, but that's a Sphinx 2.0+ target. I will open an RFE issue to do this.
- There are no specific environment requirements for this change. Tests are included.

### Detail

- Register a 'sphinx.ext.apidoc' extension and related configuration options.
- Add tests and documentation for same

### Relates

- #1135

PR #3668, opened 2017-04-25
Don't ignore first argument passed to sphinx.apidoc.main

adamjstewart:
### Feature or Bugfix

Bugfix? Depends on whether you consider it a bug or a feature 😉 

### Purpose

We use Sphinx to generate our documentation. Our API changes fairly often, so in order to make sure our API docs stay up-to-date, we always re-run `sphinx-apidoc` with each build. Unfortunately, Read the Docs skips our Makefile and directly calls:
```
$ sphinx-build -b html . _build/html
```
To get around this, we run `sphinx-apidoc` from our `conf.py`. Our call looks something like:
```python
import sphinx.apidoc.main as sphinx_apidoc

sphinx_apidoc(['--force', '--no-toc', '--output-dir=.', '../spack'])
```
At first I was really confused why this gave different results than:
```
$ sphinx-apidoc --force --not-toc --output-dir=. ../spack
```
but after reading the source code, I quickly discovered that `main` ignores the first argument it is passed. This makes perfect sense when running `sphinx-apidoc` from the command line, as the first argument will be the name of the script. But it is much less intuitive when being called directly from `main`. This PR fixes that.

### Detail

By default, `main` will still process `sys.argv[1:]`, but if passed a list of arguments, it will now process all of them. I'm not sure if this bug (feature?) is documented already. My only concern is that this change isn't really backwards compatible. If you wanted it to be, I could theoretically check if the first argument is `sphinx-apidoc` and skip it if so.

### Relates

First reported here:
https://github.com/LLNL/spack/pull/3982/files#r113266150

PR #5106, opened 2018-06-18
Fix #5104: apidoc: Interface of ``sphinx.apidoc:main()`` has changed

### LINKED ISSUE #5104: sphinx-build: error: argument -d/--maxdepth: invalid int value
I recently updated to the latest version of Sphinx and am now unable to build the documentation for my project.

### Problem

When I try to build the documentation for my project, I encounter the following error message:
```
$ make
sphinx-build -b html -d _build/doctrees  -E . _build/html
Running Sphinx v1.7.5
usage: sphinx-build [OPTIONS] -o <OUTPUT_PATH> <MODULE_PATH> [EXCLUDE_PATTERN, ...]
sphinx-build: error: argument -d/--maxdepth: invalid int value: '_build/doctrees'

Configuration error:
The configuration file (or one of the modules it imports) called sys.exit()
make: *** [html] Error 2
```

```
$ git clone https://github.com/spack/spack.git
$ cd spack/lib/spack/docs
$ make
```

#### Error logs / results

This was first reported in https://github.com/spack/spack/issues/8491#issuecomment-397923309. This issue appears to be identical to https://github.com/rtfd/readthedocs-build/issues/44. The RtD issue suggests locking the version of Sphinx to 1.7.0. Indeed, this issue has never caused us trouble as we lock our version to 1.7.0 for Travis and RtD, but I would like to be able to build the docs locally.

#### Expected results

The docs should build correctly. According to `sphinx-build --help`, `-d` should accept a path, not an int:
```
  -d PATH           path for the cached environment and doctree files
                    (default: OUTPUTDIR/.doctrees)
```
Perhaps this is coming from `sphinx-apidoc`?
```
  -d MAXDEPTH, --maxdepth MAXDEPTH
                        maximum depth of submodules to show in the TOC
                        (default: 4)
```

https://github.com/spack/spack

### Environment info
- OS: macOS 10.13.5
- Python version: 3.6.4
- Sphinx version: 1.7.5

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5104

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,7 @@ Bugs fixed
 * #5019: autodoc: crashed by Form Feed Character
 * #5032: autodoc: loses the first staticmethod parameter for old styled classes
 * #5036: quickstart: Typing Ctrl-U clears the whole of line
+* #5104: apidoc: Interface of ``sphinx.apidoc:main()`` has changed
 
 Testing
 --------
--- a/sphinx/apidoc.py
+++ b/sphinx/apidoc.py
@@ -9,20 +9,20 @@
     :license: BSD, see LICENSE for details.
 """
 
+import sys
 import warnings
 
 from sphinx.deprecation import RemovedInSphinx20Warning
 from sphinx.ext.apidoc import main as _main
 
 
-def main(*args, **kwargs):
+def main(argv=sys.argv):
     warnings.warn(
         '`sphinx.apidoc.main()` has moved to `sphinx.ext.apidoc.main()`.',
         RemovedInSphinx20Warning,
         stacklevel=2,
     )
-    args = args[1:]  # skip first argument to adjust arguments (refs: #4615)
-    _main(*args, **kwargs)
+    _main(argv[1:])  # skip first argument to adjust arguments (refs: #4615)
 
 
 # So program can be started with "python -m sphinx.apidoc ..."


adamjstewart:
Doesn't this suffer from the same problem that was fixed in #3668? Shouldn't it be `argv=sys.argv[1:]` and `_main(argv)`?

tk0miya:
`sphinx.apidoc:main()` is only provided to keep compatibility with 1.6 or older. So it should not omit the first item of `sys.argv`. On the other hand, `sphinx.ext.apidoc:main()` takes `sys.argv[1:]`.
I think this is correct.

adamjstewart:
@tk0miya Shouldn't both `sphinx.ext.apidoc.main()` and `sphinx.apidoc.main()` behave identically? Currently, both default to `sys.argv[1:]`, the only difference is that `sphinx.apidoc.main()` throws away the first argument it is given, while `sphinx.ext.apidoc.main()` does not. This seems like it is reverting the behavior introduced in #3668.

tk0miya:
Yes, #3668 is incompatible change to past release. So this revert the change. The most important feature of `sphinx.apidoc:main()` is taking full `sys.argv` as Sphinx-1.6 doing.

adamjstewart:
Why revert #3668? It fixed an important bug where the first argument passed to `sphinx.apidoc.main` is completely ignored with no warning. Reintroducing bugs to preserver compatibility with old releases doesn't make sense to me.

tk0miya:
`sphinx.apidoc:main()` is only used to keep compatibility with past release. It should be same as 1.6's one doing. That means it should take whole of `sys.argv`, not `sys.argv[1:]`.

The change of #3668 is merged into `sphinx.ext.apidoc:main()`. It is successor of `sphinx.apidoc` module. I think its interface is same as what you want.

>It fixed an important bug where the first argument passed to sphinx.apidoc.main is completely ignored with no warning. 

I don't think this is not a bug. The main function is not public API. So there are no specification.

adamjstewart:
Ah, I see. I guess I assumed that #3668 had been merged and released before 1.7.

> I don't think this is not a bug. The main function is not public API. So there are no specification.

Fair enough. Is `sphinx.ext.apidoc.main()` a public API?

tk0miya:
>Fair enough. Is sphinx.ext.apidoc.main() a public API?

It's not. The public APIs are listed in document:
http://www.sphinx-doc.org/en/master/extdev/index.html

But it does not mean other functions are fragile. We'll work to keep all of functions stable. So please let me know if something goes wrong.

adamjstewart:
If `sphinx.ext.apidoc.main` isn't public, is there a recommended way to automatically generate API docs? Is it preferable to call `sphinx-apidoc` in a subprocess?

tk0miya:
AFAIK, `apidoc` is useful to create skelton of API document project. So it is not needed to invoke it through APIs. For that purpose, we provides `sphinx.ext.autosummary` module. How about this?
Please let me know your goal.

Anyway, your request is not related with this bug. So I'd like to merge this if this works fine. Could you move to new issue please?

stephenfin:
> If `sphinx.ext.apidoc.main` isn't public, is there a recommended way to automatically generate API docs? Is it preferable to call `sphinx-apidoc` in a subprocess?

@adamjstewart try [sphinx-contrib/apidoc](https://github.com/sphinx-contrib/apidoc). We're using this extensively in OpenStack until I can talk @tk0miya around into accepting #4101 :wink: 

> AFAIK, `apidoc` is useful to create skelton of API document project. So it is not needed to invoke it through APIs. For that purpose, we provides `sphinx.ext.autosummary` module. How about this?
> Please let me know your goal.

@tk0miya I'm afraid it's more complicated than that. There are big differences between `sphinx.ext.autosummary` and `sphinx-apidoc`. I described these in https://github.com/sphinx-doc/sphinx/pull/4101#issuecomment-375359695. Until we resolve those differences, `sphinx-apidoc` remains the only option for OpenStack at least.

stephenfin:
> Why revert #3668? It fixed an important bug where the first argument passed to sphinx.apidoc.main is completely ignored with no warning. Reintroducing bugs to preserver compatibility with old releases doesn't make sense to me.

FWIW, #3668 was necessary due to a change in how argparse and optparse work. That defaults to using `sys.argv[1:]` unless you tell it otherwise, which we needed to do so we could call it programatically for unit tests and the likes. As you've noted though, neither original nor updated version are correct in that there's a disconnect between what we expected in the past (a list of arguments referenced by `argv`) and what we expect today (an option list of args and kwargs, referenced via `arg` and `kwarg` respectively). As a result, this fix seems correct.

stephenfin (approved):
Looks good to me.

jakirkham:
> ...try sphinx-contrib/apidoc.

Thanks for sharing @stephenfin. Glad to hear there are proper tools out there. Have been resorting to hacks like calling `sphinx.ext.apidoc.main` from `conf.py` for some time.

If it helps, @tk0miya, lots of people try to do something similar. ( https://github.com/rtfd/readthedocs.org/issues/1139 ) Am sure they would also appreciate proper support in Sphinx for this use case. :)

jakirkham:
In any event, would still be great to have this PR land.

tk0miya:
@stephenfin Thank you for your help. it was very helpful to me (because I'm very poor at English...)

tk0miya:
Merged. Thank you for your comment. Let's talk about the API or modules in another issue.

I'll take a look #4101 later. Please wait a moment.

PR #5110, opened 2018-06-19
Doc rework (user guide, part 5)

stephenfin:
Add builders and extensions to the user guide. As before, this is mostly done by shuffling around existing content. There are also some TODOs - as always, I will resolve these in separate changes.


tk0miya (approved):
LGTM!

tk0miya:
Merged!

PR #5111, opened 2018-06-20
Run everything but unit tests under Python 3

stephenfin:
The demise of Python 2.7 is not far off and we intend to drop support for it in Sphinx 2.0. As a result, there probably isn't a better time than now to start testing with Python 3 by default. There are a couple of advantages to this:
    
- flake8 is stricter under Python 3 and we don't need to ignore files with Python 3 syntax
- We get to dogfood Sphinx against its own documentation using Python 3
- pylint gains support for Python 3 syntax

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -27,12 +27,14 @@ commands=
     pytest -Wall --durations 25 {posargs}
 
 [testenv:flake8]
+basepython = python3
 description =
     Run style checks.
 commands =
     flake8
 
 [testenv:pylint]
+basepython = python3
 description =
     Run source code analyzer.
 deps =
@@ -42,6 +44,7 @@ commands =
     pylint --rcfile utils/pylintrc sphinx
 
 [testenv:coverage]
+basepython = python3
 description =
     Run code coverage checks.
 setenv =
@@ -51,6 +54,7 @@ commands =
     coverage report
 
 [testenv:mypy]
+basepython = python3
 description =
     Run type checks.
 deps =
@@ -59,6 +63,7 @@ commands=
     mypy sphinx/
 
 [testenv:docs]
+basepython = python3
 description =
     Build documentation.
 commands =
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -27,12 +27,14 @@ commands=
     pytest -Wall --durations 25 {posargs}
 
 [testenv:flake8]
+basepython = python3
```
tk0miya, line 30: Now we control the version of python by `.travis.yml`. I feel this setting is not needed.

tk0miya (commented):
LGTM with nits

tox.ini
stephenfin: We control this for Travis but not for local runs. You'll see that I have two patches in this pull request: the first was necessary for me because I'm using Fedora 28 which defaults to Python 3. I think tests and linters should run the same locally as they do in the CI, right?

jakobandersen:
For consistency I suggest changing the default interpreter in ``Makefile`` and ``doc/Makefile`` to ``python3`` as well.

stephenfin:
> For consistency I suggest changing the default interpreter in `Makefile` and `doc/Makefile` to `python3` as well.

I've done the former. Wasn't sure if the latter was necessary so I've left that for now

tox.ini
tk0miya: Ah, you're right. Go ahead.

BTW, flake8 is defined as `2.7` in `.travis.yml`. Should this be also changed?
https://github.com/sphinx-doc/sphinx/blob/f14d60dcd2772b7e973cc7dee7a6e625b44926c6/.travis.yml#L33-L34

tk0miya (approved):

tox.ini
stephenfin: Yes, I will change this shortly.

tk0miya:
Merged. Thank you always!

PR #5112, opened 2018-06-20
setup.py: Move configuration to setup.cfg

stephenfin:
Recent versions of Setuptools allows most configuration to be kept in
'setup.cfg' instead of 'setup.py' [1]. This allows for some automation
scenarios and results in less duplication than we might otherwise see.

There is one functional change included: the requirement for setuptools
30.3.0 (8 December 2016). This is required to support the 'setup.cfg'
format.

[1] https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,90 @@
 [metadata]
+name = Sphinx
+version = attr: sphinx.__version__
+author = 'Georg Brandl
+author_email = 'georg@python.org
+description = 'Python documentation generator
+long_description = file: README.rst
+license = BSD
 license_file = LICENSE
+url = http://sphinx-doc.org/
+download_url = https://pypi.org/project/Sphinx/
+platforms = any
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Environment :: Console
+    Environment :: Web Environment
+    Intended Audience :: Developers
+    Intended Audience :: Education
+    Intended Audience :: End Users/Desktop
+    Intended Audience :: Science/Research
+    Intended Audience :: System Administrators
+    License :: OSI Approved :: BSD License
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 2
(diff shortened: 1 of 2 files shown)


tk0miya:
Does this require newer setuptools to all developers and contributors? I feel it is too new for some environment. It might not work bare Ubuntu LTS.

stephenfin:
> Does this require newer setuptools to all developers and contributors? I feel it is too new for some environment. It might not work bare Ubuntu LTS.

Yeah, good point. I see the failing CI now. This requires setuptools 30.3.0. Ubuntu 14.04 currently provides [3.3](https://packages.ubuntu.com/trusty/python-setuptools), Ubuntu 16.04 currently provides [20.7.0](https://packages.ubuntu.com/xenial/python-setuptools), and Ubuntu 18.04 currently provides [39.0.1](https://packages.ubuntu.com/bionic/python-setuptools). This means only 18.04 will support this using the distro version of setuptools.

That being said, this only affects users installing the package from PyPI. It is very unlikely that people installing packages from PyPI will not also install new versions of pip and setuptools (indeed, it's probably a necessity for 14.04 and 16.04). In addition, we already use the `python_requires` setting, which is only available from setuptools v24.2.0 (added in commit [98477464e](https://github.com/pypa/setuptools/commit/98477464e)) which rules out the still-supported Ubuntu 14.04. As a result, I'd rather we adjust the CI to install setuptools from PyPI instead of hamstringing ourselves like this but I may very well be missing something :smile:

tk0miya:
>It is very unlikely that people installing packages from PyPI will not also install new versions of pip and setuptools (indeed, it's probably a necessity for 14.04 and 16.04).

I don't think so. Please take a look issues on github. There are many real cases. I suppose this change will cause big trouble...

About `python_requires`, it did not make effect to old setuptools. I think it is not related with this change.

Issue #5114, opened 2018-06-21
sphinx make html error:【WinError 5】


Sunnysky080:
Subject: <cannot use the command make html with an WinError 5>

### Problem
```
Running Sphinx v1.7.5
loading pickled environment... not yet created

Exception occurred:
  File "c:\program files\python36\lib\site-packages\sphinx\util\osutil.py", line 102, in walk
    names = os.listdir(top)
PermissionError: [WinError 5] 拒绝访问。: 'C:\\Users\\hp\\AppData\\Local\\Application Data'
The full traceback has been saved in C:\Users\hp\AppData\Local\Temp\sphinx-err-lxhm478i.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
>
```

#### Procedure to reproduce the problem
```
make html
```

#### Error logs / results
```
# Sphinx version: 1.7.5
# Python version: 3.6.5 (CPython)
# Docutils version: 0.14 
# Jinja2 version: 2.10
# Last messages:

# Loaded extensions:
Traceback (most recent call last):
  File "c:\program files\python36\lib\site-packages\sphinx\application.py", line 282, in _init_env
    self.env = BuildEnvironment.frompickle(filename, self)
  File "c:\program files\python36\lib\site-packages\sphinx\environment\__init__.py", line 128, in frompickle
    with open(filename, 'rb') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\hp\\_build\\doctrees\\environment.pickle'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\program files\python36\lib\site-packages\sphinx\cmdline.py", line 303, in main
    args.warningiserror, args.tags, args.verbosity, args.jobs)
  File "c:\program files\python36\lib\site-packages\sphinx\application.py", line 231, in __init__
    self._init_env(freshenv)
  File "c:\program files\python36\lib\site-packages\sphinx\application.py", line 293, in _init_env
    self._init_env(freshenv=True)
  File "c:\program files\python36\lib\site-packages\sphinx\application.py", line 275, in _init_env
    self.env.find_files(self.config, self.builder)
  File "c:\program files\python36\lib\site-packages\sphinx\environment\__init__.py", line 418, in find_files
    exclude_matchers=matchers):
[...]

PR #5116, opened 2018-06-21
Handle OSError on scanning documents (refs: #5114)

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5114


sphinx/environment/__init__.py
```diff
@@ -416,37 +416,40 @@ def find_files(self, config, builder):
...
+                    catalog_files = find_catalog_files(
+                        docname,
+                        self.srcdir,
+                        self.config.locale_dirs,
+                        self.config.language,
+                        self.config.gettext_compact)
+                    for filename in catalog_files:
+                        self.dependencies[docname].add(filename)
+        except OSError as exc:
+            raise SphinxError(__('Failed to scan documents in %s: %r') % (self.srcdir, exc))
```
tk0miya: @shimizukawa I'd like to your opinion to this patch.

1. Is `SphinxError` appropriate for this case? Should I add a new error class like `BuildError` or `DocumentError`?
2. Which module is better to handle such exceptions? `env`, `builder` or `app`?

shimizukawa: 1. +1 for `BuildError` and/or `DocumentError` instead of direct usage of `SphinxError`. 
2. I think that these primitive exceptions should be caught by the direct use code. Instead, I think it's better to catch SphinxError and its descendant exceptions in one place and deal with it.

shimizukawa (approved):
LGTM with nits

sphinx/environment/__init__.py
tk0miya: done

Issue #5121, opened 2018-06-25
Display all warnings with -W (warningiserror)


blueyed:
Currently using `-W` will stop on the first warning, but that makes it harder to interate over existing warnings, since you only get the first one.

This appears to happen via the `WarningIsErrorFilter` currently, which raises `SphinxWarning`.

I think using `-W` should display all warnings (i.e. keep on processing), but then cause a non-zero exit status only (and e.g. forcing a fresh build the next time, like it appears to happen for exceptions currently in general).

tk0miya:
Thank you for proposal. But that is already posted as #4614.
So I'm closing this because of dup.

PR #5122, opened 2018-06-25
BuildDoc: add and pass through nitpicky option

blueyed:
This allows for using e.g. `tox -e docs -- -n` with Sphinx's own
`build_sphinx` distutils command.

Related: #5121 (since Sphinx uses `build_sphinx.warning-is-error = 1` in `setup.cfg` and therefore it will stop at the first warning)

### DIFF
--- a/sphinx/setup_command.py
+++ b/sphinx/setup_command.py
@@ -87,9 +87,10 @@ class BuildDoc(Command):
         ('link-index', 'i', 'Link index.html to the master doc'),
         ('copyright', None, 'The copyright string'),
         ('pdb', None, 'Start pdb on exception'),
+        ('nitpicky', 'n', 'nit-picky mode, warn about all missing references'),
     ]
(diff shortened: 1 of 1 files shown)


sphinx/setup_command.py
```diff
@@ -174,6 +176,8 @@ def run(self):
             confoverrides['today'] = self.today
         if self.copyright:
             confoverrides['copyright'] = self.copyright
+        if self.nitpicky:
+            confoverrides['nitpicky'] = self.nitpicky
```
tk0miya, line 180: It seems mypy warns for this line.
https://travis-ci.org/sphinx-doc/sphinx/jobs/396257990

Could you add type hints to `# type: Dict[unicode, Any]` to L.168 please?

tk0miya (changes requested):
LGTM with nits

tk0miya:
Merged.
Thank you for contribution!

PR #5127, opened 2018-06-28
quickstart: check default path for validity

blueyed:
`sphinx-quickstart` without arguments will currently overwrite e.g. a
project's `Makefile`, since it only checks for `conf.py` explicitly in
`ask_user`.

This change uses the same check for both a provided path via arguments
and in `ask_user` now.

### DIFF
--- a/sphinx/cmd/quickstart.py
+++ b/sphinx/cmd/quickstart.py
@@ -265,12 +265,7 @@ Selected root path: %s''') % d['path']))
 Enter the root path for documentation.'''))
         d['path'] = do_prompt(__('Root path for the documentation'), '.', is_path)
 
-    while path.isfile(path.join(d['path'], 'conf.py')) or \
-            path.isfile(path.join(d['path'], 'source', 'conf.py')):
-        print()
-        print(bold(__('Error: an existing conf.py has been found in the '
-                      'selected root path.')))
-        print(__('sphinx-quickstart will not overwrite existing Sphinx projects.'))
+    while not check_valid_dir(d):
         print()
         d['path'] = do_prompt(__('Please enter a new root path (or just Enter '
                                  'to exit)'), '', is_path)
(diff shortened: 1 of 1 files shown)


tk0miya:
I agree that overwriting `Makefile` is not good behavior. But I also feel `valid_dir()` does too explicit checking because both `Makefile` and `make.bat` are not required files for Sphinx project. They are optional.

I propose you to write `Makefile` and `make.bat` only if they don't exist. What do you think?

blueyed:
> I propose you to write `Makefile` and `make.bat` only if they don't exist. What do you think?

Might be good already - at least overwriting the Makefile was what got me.

But then it should maybe just only act on empty dirs, which the error with `-q` states incorectly already:

```
% sphinx-quickstart -q -p foo -a author EXAMPLES

Error: specified path is not a directory, or sphinx files already exist.
sphinx-quickstart only generate into a empty directory. Please specify a new root path.
```

tk0miya:
Sorry, my comment above is not important. Please forget it.
Anyway, CI raises errors for this PR. Could you check them please?

blueyed:
I've rebased it and added a fixup: it moves the check in ask_user, because it uses things being asked for only later.

I am not sure if this is the correct way (since there might be missing feedback when the check is done, and there is no simpler check now directly after asking for the root), but at least tests should be good now.
It should be improved further from here.

sphinx/cmd/quickstart.py
```diff
@@ -347,6 +335,13 @@ def ask_user(d):
         d['master'] = do_prompt(__('Please enter a new file name, or rename the '
                                    'existing file and press Enter'), d['master'])
 
+    while not check_valid_dir(d):
```
tk0miya, line 338: I understand all of `sep`, `dot`, `master` and `suffix` are missing just after asking root path. But this place is too far from question.
I think the goal of this PR is not to override `Makefile` and `make.bat` by default. To do that, how about changing `overwrite` option of `generate()` function to False?

blueyed: > But this place is too far from question.

What do you mean?

> I think the goal of this PR is not to override Makefile and make.bat by default. To do that, how about changing overwrite option of generate() function to False?

I don't know - but please feel free to do it like that.  While not perfect it would still be a good enough fix then I think.

You can push updates to my branch/this PR, or just create a new one and close this.

tk0miya: >What do you mean?

What I meant is that the check for root path is going after 6 steps after from its question. I worried users might get confused.

>I don't know - but please feel free to do it like that. While not perfect it would still be a good enough fix then I think.

Okay, I'll reconsider about this later.

tk0miya:
Finally, I meged #5378 instead of this.

Issue #4074, opened 2017-09-19
Links created for items in "Attributes:" section


andy-maier:
Subject: Sphinx with autodoc and napoleon tries to create cross-reference links on the items in an "Attributes:" section that document public instance attributes of a class.

### Problem

We are using Sphinx with autodoc and the napoleon extension for a Python project. We are using "Attributes:" sections in class docstrings to describe the public instance attributes of the class that are not defined as Python properties.

This works fine, except that Sphinx tries to create links on the attribute names. It attempts to locate the target of the link automatically and seems to search all classes for potential targets. Sphinx uses for example Python properties in *other* classes as potential targets.

That behavior is not understandable for me. An item in an Attributes section stands on its own, i.e. should be considered its own target, so I don't understand why a link on that attribute would make any sense?
Also, why are *other* classes searched for a property with that name?

In cases where more than one  other class has a Python property with the name of an item in the Attributes section, the following warning is raised:

```
pywbem/cim_obj.py: docstring of pywbem.CIMInstanceName:: WARNING: more than one target found for cross-reference u'host': pywbem.WBEMListener.host, pywbem.WBEMConnection.host
```
See our issue pywbem/pywbem#761 for more details on this particular case.

#### Procedure to reproduce the problem
```
PYTHONPATH=. sphinx-build -b html -v -d build/doctrees -c . . build/html
```

With the following `test.py` file in that directory:
```
#!/usr/bin/env python

class B1(object):
    """A B1"""

    @property
    def name(self):
        """Name of the B1."""
        return 'dummy'

class B2(object):
    """A B2"""

    @property
    def name(self):
        """Name of the B2."""
        return 'dummy'

class A(object):
    """
    An A.

    Attributes:

      name (unicode): Name of the A.

      foo (int): A foo.
    """

    def __init__(self, name, foo):
        """
        Initialize an A.

        Parameters:

          name (str or unicode): Name of the A.

          foo (int): A foo.
        """
        self.name = to_unicode(name)
        self.foo = foo
```

#### Error logs / results

In the resulting HTML output, the attribute `A.name` is generated as a link to `test.B1.name`.

```
Running Sphinx v1.6.3
making output directory...
/home/maiera/virtualenvs/pywbem27/lib/python2.7/site-packages/sphinx/util/compat.py:40: RemovedInSphinx17Warning: sphinx.util.compat.Directive is deprecated and will be removed in Sphinx 1.7, please use docutils' instead.
[...]

PR #5129, opened 2018-06-29
Qualify attribute names for generated ivar

toslunar:
Subject: Make Napoleon qualify attribute names for generated ivar

### Feature or Bugfix
- Feature

### Purpose
This PR gives a partial fix of #4074.  I hope this works in the most cases (an exception is the same attribute name in different classes with the same name, in python<3.3).

### Detail
- Make Napoleon apply a workaround that my team has manually done.
For example, the `Attributes` items in
```python
class A(object):
    """
    An A.

    Attributes:
        foo (int): Blah blah.
    """
```
will be treated as
```
    Attributes:
        ~module_name.A.foo (int): Blah blah.
```
which is converted to
```
    :ivar ~module_name.A.foo: Blah blah.
    :vartype ~module_name.A.foo: int
```
in order to avoid unintended links.

- This PR does not break "manually fixed" lines.

### Relates
- #4676

### DIFF
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -43,6 +43,18 @@ _enumerated_list_regex = re.compile(
     r'(?(paren)\)|\.)(\s+\S|\s*$)')
 
 
+def _qualify_name(attr_name, klass):
+    if klass and not '.' in attr_name:
+        if attr_name.startswith('~'):
+            attr_name = attr_name[1:]
+        try:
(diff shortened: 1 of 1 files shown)


RobRuana (changes requested):
This looks great! I would like to see at least one unit test that exercises the new code. Add a unit test, and I'll approve this!

RobRuana (approved):

toslunar:
Thank you very much for refactoring and merging this!

Issue #5133, opened 2018-06-30
latex: index headings "Symbols" and "Numbers" not internationalized


jfbu:
Consider following `index.rst`

```
==========
Index Test
==========


.. index:: état

.. index:: Deutschland

.. index:: 666

.. index:: **


```

When building with language set to German (`'de'`)  one gets this

![capture d ecran 2018-06-30 a 15 09 04](https://user-images.githubusercontent.com/2589111/42125569-ec0fe1ba-7c79-11e8-9a9d-9cb3bb2b4b44.png)

When buildind with language set to Italian, one gets this:

![capture d ecran 2018-06-30 a 15 11 24](https://user-images.githubusercontent.com/2589111/42125571-f9c4ee18-7c79-11e8-8291-d056dce6ccdd.png)

Remark: the fact that word `état` starting with utf-8 letter ends up in `Symbols` is related to use of `makeindex`, this is other issue (see #5132 for related issue)

### Problem

The problem is that `Symbols` and `Numbers` headings are not internationalized.

This originates in hard-coded https://github.com/sphinx-doc/sphinx/blob/1140e81df7ee9df1135ef9f35302ffd72d5bd764/sphinx/texinputs/python.ist#L12-L13

### Environment info
- OS: Mac
- Python version: 3.6.5
- Sphinx version: 1.7.x, master
- TeXLive 2018

Issue #5130, opened 2018-06-29
xetex issue: Improper discretionary list.


JulienPalard:
Subject: Building pdf is failing with latex bug.

### Problem

I'm running sphinx-build on the french documentation translation of Python, and pdf build is failing.

#### Procedure to reproduce the problem
```
git clone https://github.com/python/python-docs-fr.git
git checkout 3.7
make MODE=autobuild-dev
```

#### Error logs / results
```
[...]
Writing index file howto-curses.idx

LaTeX Warning: Unused global option(s):
    [a4].

(./howto-curses.aux) (/usr/share/texlive/texmf-dist/tex/latex/base/ts1cmr.fd)
*geometry* driver: auto-detecting
*geometry* detected driver: xetex
(/usr/share/texlive/texmf-dist/tex/latex/hyperref/nameref.sty
(/usr/share/texlive/texmf-dist/tex/generic/oberdiek/gettitlestring.sty))

Package hyperref Warning: Rerun to get /PageLabels entry.

(/usr/share/texlive/texmf-dist/tex/latex/amsfonts/umsa.fd)
(/usr/share/texlive/texmf-dist/tex/latex/amsfonts/umsb.fd) [1] [2]
! Improper discretionary list.
<recently read> }
                 
l.264 ...gin\PYGZus{}y} \PYG{o}{=} \PYG{l+m+mi}{7}
                                                  
?  
! Emergency stop.
<recently read> }
                 
l.264 ...gin\PYGZus{}y} \PYG{o}{=} \PYG{l+m+mi}{7}
                                                  
Output written on howto-curses.pdf (2 pages).
Transcript written on howto-curses.log.
=== TeX engine is 'XeTeX'
Latexmk: Index file 'howto-curses.idx' was written
Latexmk: References changed.
Latexmk: Errors, so I did not complete making targets
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 1
      Refer to 'howto-curses.log' for details
Latexmk: Use the -f option to force complete processing,
 unless error was exceeding maximum runs, or warnings treated as errors.
Makefile:32 : la recette pour la cible « howto-curses.pdf » a échouée
make[3]: *** [howto-curses.pdf] Erreur 12
make[3] : on quitte le répertoire « /home/mdk/clones/cpython/Doc/build/latex »
Makefile:130 : la recette pour la cible « dist » a échouée
make[2]: *** [dist] Erreur 2
[...]

PR #5134, opened 2018-06-30
Add new config value ``latex_use_xindy``

### LINKED ISSUE #5132: lualatex issue: PDF build fails if indexed word starts with Unicode characters

Do ``make latexpdf`` with this index.rst:

```
.. index:: état
```

and  having configured ``latex_engine = 'lualatex'`` in conf.py

#### Error logs / results
```
! String contains an invalid utf-8 sequence.
l.4   \bigletter 
               ?^^@
? 
```

#### Expected results

PDF build succeeds.

It does with `xelatex` but this is luck. In fact, usage of `makeindex` program to create `.ind` file results in a non-valid file for utf-8 encoding. Here is how the file looks like when opened as if latin-1 encoded:

```
\begin{sphinxtheindex}
\def\bigletter#1{{\Large\sffamily#1}\nopagebreak\vspace{1mm}}

  \bigletter Ã
  \item Ã©tat, \hyperpage{1}

\end{sphinxtheindex}
```

Probably `xindy` should be used with `lualatex/xelatex`. But this requires particular set-up.

Relates: https://github.com/sphinx-doc/sphinx/issues/5130#issuecomment-401450137

### Environment info
- OS: Mac
- Python version: 3.6.5
- Sphinx version: 1.7.4

jfbu:
### Work in progress (now finished)

- [x] add basic possibility to use xindy
- [x] workaround xindy's incompatibility with hyperref
- [x] workaround xindy's inability with pdflatex+Cyrillic

(the merged PR got a subsequent fix at 9b31ee8d5, due to oversight regarding pdflatex+Cyrillic)

Unfortunately the ``xindy`` binary (which exists for two decades and is documented succinctly in the LaTeX Companion) has no hyperref compatibility, from man page of ``texindy``:

> In particular, this means that by default the LaTeX package "hyperref" will create raw index files that cannot be processed with xindy. This is not a bug, this is the unfortunate result of an intented incompatibility. It is currently not possible to get both hyperref's index links and use xindy.

This means that Sphinx LaTeX writer must use `hyperindex=false` option of `hyperref` and add its own mark-up in usage of ``\index`` and specific ``.xdy`` xindy style file. I will update when ready. **done**

Note: this applies only for general index corresponding to usage of ``.. index::`` directive or role. The domain indices are generated entirely by Sphinx and do not process via the LaTeX mechanism of `\makeindex` creating an `.idx` file and then `makeindex/xindy` converting the `.idx` file into `.ind` file which is included by `\printindex`.

### Feature or Bugfix

This fixes #5132.

It is also new feature for PDF builds with default engine (`pdflatex`), but is not activated by default.

It would make sense to activate it by default for xelatex/lualatex, **done !**

the problem though is that complete language support is currently only for languages with Latin scripts ; xindy also supports Cyrillic and Greek but I am not sure how to configure it via usage of texindy. Usage of texindy is the simplest way to add the feature as replacement of `makeindex` and former customization (file python.ist, now file python.xdy for xindy) with minimal complications.

### Purpose

Consider this source:

```rest
==========
Index Test
==========


.. index:: état

.. index:: Deutschland

.. index:: 666

.. index:: **

```

Currently the following index is created in PDF:

capture d ecran 2018-06-30 a 23 42 40

With this PR and the option set to True it will be

capture d ecran 2018-06-30 a 23 44 14

(I have not found a way so far to configure separately Symbols and Numbers groups).

The difference is that the indexed word `état` does not end up with symbols.

Besides, without this PR, this source gives

- broken PDF build with `lualatex` (see #5132)

- strange index with `xelatex`: 

    capture d ecran 2018-06-30 a 23 02 26

   where  the word `état` would be at very end of index.

The default words `Symbols and Numbers` are made a translatable string by this PR (the issue #5133 with `makeindex` is not fixed by this PR, which is only about adding `xindy` usage).

### Relates
- #5133, #5132, #5130 (because it would allow usage of `lualatex` to work around `xelatex` problem with French language).

It was not very easy to put up this PR, because the `xindy` [doc](http://xindy.sourceforge.net/) is a bit sparse. I used the man page of `texindy` and examined the files coming with `xindy` in TeX Live 2018. This [tex.sx answer](https://tex.stackexchange.com/a/199841/4686) is also useful (although I did not reproduce some of the problem mentioned there with `-L` option).


mitya57:
Thanks for working on this! Maybe it will be finally possible to have a working index for Russian.

I tested it with our project (adding `hyperindex=false` option for hyperref), but it fails with this error:

> ERROR: CHAR: index 0 should be less than the length of the string

Do you know what can cause this? [Here is][1] our `.idx` file (renamed because GitHub does not allow `.idx` extension). We are using pdflatex engine.

[1]: https://github.com/sphinx-doc/sphinx/files/2163191/instruction.idx.txt

jfbu:
@mitya57 I don't know what could be causing the issue you report, but it may very well be caused by the fact that my PR configures `texindy` well only for Latin scripts, not Cyrillic. I downloaded the `.idx` file and I will examine the question. But I am offline now for 3 weeks. I will have time to peruse the xindy doc which is very incomplete and seemingly outdated for years and its sources if I have them in my TeXLive 2018.

jfbu:
PS: in my testing with Latin languages the hyperref incompatibilty shows in absence of cross-links in case of `see` or problems when one location is marked via `!` as the main one. The build succeeds but there are no hyperlinks. Basic indexed terms get functional hyperlinks. Anyway PDF output gives only backlinks to page, not to line where term was indexed.

This could be improved in future but would be more extensive PR.

jfbu:
@mitya57 I have pushed commits to get xindy+Cyrillic+pdflatex to work. The error

>     ERROR: CHAR: index 0 should be less than the length of the string

arises when xindy "merge-rules" create an empty string. Unfortunately this arises with `.idx` files full of `\IeC` macros when one uses `texindy` script which defines some merge-rules allowing a few TeX macros but then discards all others, among them those `\IeC` related to Cyrillic LaTeX Internal Character Representation. Once I understood that I could find a workaround.

jfbu:
Notice that this PR includes a fix of #5176, which I separated into #5177 for 1.7 branch. This was needed to test this PR with building CPython docs.

tk0miya (approved):

jfbu:
@tk0miya thanks for review! Now merging. 

@mitya57 please try out if you have time on your project with `pdflatex`, I hope it will work. You need to clean LaTeX build repertory first, or do `make latexpdf` twice, because old `.idx` file contains hyperref mark-up ignored by xindy, but this PR maintains the hyperlinks in index by another way.

**edit** actually the reason for cleaning LaTeX build (as asked for in CHANGES) is that on first run after having turned on the `latex_use_xindy` the old `.ind` file contains `\hyperindexformat` usage, but with `xindy`, the hyperref is used with option `hyperindex=false` and that macro is now undefined. This is why one needs `make clean` in LaTeX build repertory (or `make clean latexpdf`) when moving to `latex_use_xindy = True`.

mitya57:
Now the build succeeds, but all index entries go to “Non-alphabetical” group, even those with latin names. Also the sorting is broken: it sorts the `\CYR*` macros by their latin names, not by the letters’ actual positions in the Russian alphabet. E.g. `\CYRC` (Ц) should go after `\CYRH` (Х), not after `CYRB` (Б).

With the [previously posted `instruction.idx`](https://github.com/sphinx-doc/sphinx/files/2163191/instruction.idx.txt), I get [this `instruction.ind`](https://github.com/sphinx-doc/sphinx/files/2197321/instruction.ind.txt).

jfbu:
@mitya57 thanks a lot for feedback, my bad. It was working at some earlier stage but I made some change and did not test.... I will ping you when this is fixed. Thanks again.

jfbu:
@mitya57 I hope to have fixed the issue. Can you please try it out?

mitya57:
@jfbu Much better, thanks! There is a minor issue remaining — the index entries with latin names are still marked as non-alphabetical. Is it possible to do anything with that?

The new version of generated instruction.ind is [here](https://github.com/sphinx-doc/sphinx/files/2197511/instruction.ind.txt).

jfbu:
@mitya57 thanks for testing! The instruction.ind you get is as expected. About Latin names ending in separate group:

- you can rename this heading using `\renewcommand\sphinxnonalphabeticalgroupname{whatever you want}` in preamble, (using ``'preamble'`` key of latex_elements). I chose "Non-alphabetical" where initially I had "Symbols and Numbers".

- It is certainly possible to get Latin names ordered first, or after the Cyrillic letters, as xindy is very flexible (but with very limited documentation). I did not do it yet because I had somehow in my head the view of two mixed alphabets and I was afraid of having to decide how to order the letters, but thinking only now I see the two are clearly distinct, so the idea would be to order all Latin names either before or after the Cyrillic ones an separatedly from symbols. It is a matter to add extra-rules, I did not do it yet, because the behaviour you see (with pdflatex, which xindy did not support out of the box with Cyrillic in UTF-8), is the default one of xindy for say Russian document. I will think further about it.

@tk0miya sorry that I merged PR #5134 which needed subsequent fix at 9b31ee8d5, I had solved the Cyrillic issue first but then extended to many languages and did a mistake then.

jfbu:
@mitya57 for helping me test out Russian can you give a hint how you configure your conf.py, because just using language set to `ru` and `\usepackage[T1,T2A]{fontenc}` is not enough I have issue with words such as ``Базовый учебный план`` due to the accent.

mitya57:
@jfbu there is nothing non-standard in my `conf.py`. In your string the letter `й` is decomposed into `и` (U+0438) and combining breve `◌̆` (U+0306), while it should be in composed form (U+0439). Maybe that will help.

Having all Latin names in the same group is not a problem if I can rename it. Thanks for the hint!

jfbu:
@mitya57 ah! I got it from copying pasting from PDF which explains the decomposed form, I was aware of it but did not know off-hand what was the composed form.

The problem with Latin names currently with Cyrillic language is that they will end up in same group as genuine symbols as you noticed. Also, they don't get individual letter headings. I will try to get more familiar with xindy to see if there is some easy way to fix that. The xindy files such as

```
/usr/local/texlive/2018/texmf-dist/xindy/modules/lang/russian/utf8.xdy
```

do not make any provision for also indexing Latin names, as far as I can tell.

mitya57:
Indices with both cyrillic and latin names are quite rare, so as I said having them in a single group is not a problem (at least for me). What you did so far is already a much better experience than before, thanks again!

jfbu:
@mitya57 sorry for giving wrong hint you need to wrap with `\AtBeginDocument`:

```
    'preamble': r'''
\AtBeginDocument{\renewcommand\sphinxnonalphabeticalgroupname{whatever you want}}
''',
```

<strike>
or use

```
    'preamble': r'''
\addtocaption{russian}{\renewcommand\sphinxnonalphabeticalgroupname{whatever you want}}
''',
```

</strike>

jfbu:
arrgh, sorry, going too fast:

```
    'preamble': r'''
\addto\captionsrussian{\renewcommand\sphinxnonalphabeticalgroupname{whatever you want}}
''',
```

mitya57:
Ok, thanks!

tk0miya:
@jfbu is there any remaining work? Can I start releasing?

jfbu:
@tk0miya no remaining task for me for 1.7.6, go ahead ;)

(this thread is about an 1.8 addition)

tk0miya:
oh, sorry. I was confused...

jfbu:
that's my fault with my ping  in https://github.com/sphinx-doc/sphinx/pull/5134#issuecomment-405231014, sorry to have confused you!

jfbu:
Memo: after merge, this PR was followed by two corrections: 9b31ee8d5 (Cyrillic) and fd3e7a56c (issues with Lisp syntax in xindy's "merge rules").

PR #5140, opened 2018-07-03
Add default Accept header to the linkcheck

honzajavorek:
Adds default `Accept` header to the `linkcheck` as some servers have troubles to respond correctly without it. The added `Accept` header resembles what Firefox sets by default.

### Feature or Bugfix
- Feature

### Purpose

There are situations where requested server replies with a different content (in my particular case HTTP 404) when there is no accept header, possibly because it evaluates the content negotiation to an API request instead of a browser request. This change adds a default Accept header, which equals to what my Firefox sets out of the box to its requests.

I stumbled upon this when checking a link to https://crates.io/crates/dredd-hooks. While

```
curl -i https://crates.io/crates/dredd-hooks
```

returns HTTP 404, following results in an expected HTTP 200 response with HTML body:

```
curl -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8' -i https://crates.io/crates/dredd-hooks
```

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -120,11 +120,12 @@ class CheckExternalLinksBuilder(Builder):
 
     def check_thread(self):
         # type: () -> None
-        kwargs = {}
+        kwargs = {
+            'allow_redirects': True,
+            'headers': {'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'},
+        }
         if self.app.config.linkcheck_timeout:
             kwargs['timeout'] = self.app.config.linkcheck_timeout
-
-        kwargs['allow_redirects'] = True
 
         def check_uri():
             # type: () -> Tuple[unicode, unicode, int]
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -122,7 +122,10 @@ class CheckExternalLinksBuilder(Builder):
         # type: () -> None
         kwargs = {
             'allow_redirects': True,
-            'headers': {'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'},
+            'headers': {
+                'Accept': ('text/html,application/xhtml+xml,'
+                           'application/xml;q=0.9,*/*;q=0.8')
+            },
         }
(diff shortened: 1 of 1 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -120,12 +120,16 @@ def init(self):
 
     def check_thread(self):
         # type: () -> None
-        kwargs = {}
+        kwargs = {
+            'allow_redirects': True,
+            'headers': {
+                'Accept': ('text/html,application/xhtml+xml,'
+                           'application/xml;q=0.9,*/*;q=0.8')
```
tk0miya, line 127: I don't understand why `application/xml` is needed. Can Sphinx recognize XML document?

tk0miya (approved):
LGTM with nits

sphinx/builders/linkcheck.py
honzajavorek: I'd have to test it again without `application/xml` to verify, but in my opinion even without it, it would still solve the issue I had with `crates.io`.

The reason I added the header like this is to mimic a real browser, 1:1. So I just took whatever my own Firefox is sending.

Since the link check verifies whether the external links in the docs are "not dead" and still accessible for humans reading the docs and clicking the links, displaying them in their browsers, I think it is a correct approach. If browsers are generally fine with XML and prefer it over other media types, then I think it should be there.

tk0miya: The linkcheck builder uses `HTMLParser` to parse the response. I'm okay if it can handle any XML data. To send `Accept: application/xml` means the client can accept it as a data format. But I don't know our builder supports it.

tk0miya: Note: Even `application/xml` not present, `creates.io` returns its contents:
```
$ curl -H 'Accept: text/html,application/xhtml+xml;q=0.9,*/*;q=0.8' -i https://crates.io/crates/dredd-hooks
HTTP/1.1 200 OK
...
```

tk0miya:
To step forward, I just made #5226 which does not contain `application/xml` in Accept header.
I'll merge it after tests passed. Please let me know if you have any opinion for #5226.
Thanks,

honzajavorek:
Thank you @tk0miya!

PR #5144, opened 2018-07-03
C++: support for contracts

other-mickk:
Implement support for C++2a contracts, allowing for things of the style:

```rST
.. cpp::function:: double sqrt(double arg)  \
                       [[expects: 0 <= arg]] \
                       [[ensures root: is_approx(root * root, arg)]]

    A user can document program invariants as if they were assertions, too,
    via the expression roles:

    | :cpp:expr:`[[assert: rounding_mode == FE_TOWARDZERO]]`
```

In the rendered output the contract attributes are moved to their separate line, in the spirit of the separate line that is afforted to template parameters. There is a `:contract-line-spec:` directive option to put one pre- or post-condition per line as well.

[Take a look at the proof-of-concept here.](https://mickk-on-cpp.github.io/sphinx-demo/2019-05/default/pr/)

Some reservations / possible concerns:

- I changed the order of description of function qualifiers and attributes to match that of C++ ([here](https://github.com/sphinx-doc/sphinx/pull/5144#pullrequestreview-239165472))
- I would like for the function pre- and post-conditions to be indented or somehow set apart at least through user styles, but none of the output has e.g. classes set (the same could be said of template parameters etc.)
- parsing is not strict, producing warnings rather than rejecting constructs which would not otherwise be valid C++ such as attaching contracts to any function type; this can be noisy when the parser backtracks

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -67,6 +67,8 @@ logger = logging.getLogger(__name__)
     - 'templateParams', if the line has template parameters
       Note: such lines might get a new tag in the future.
     - 'templateIntroduction, if the line is on the form 'conceptName{...}'
+    - 'contracts', if the line starts a contract list
+    - 'contract', if the line lists a contract
     No other desc_signature nodes should exist (so far).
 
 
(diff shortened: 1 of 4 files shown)


other-mickk:
@jakobandersen I have fleshed out this PR and rebased it on the 2.0 branch.

sphinx/domains/cpp.py
```diff
@@ -2504,13 +2573,17 @@ def _add_text(signode, text):
...
+            attr.describe_signature(signode, env, symbol)
+        if mode != 'lastIsName':
+            for contract in self.contracts:
+                signode += nodes.Text(' ')
+                contract.describe_signature(signode, env, symbol)
         if self.final:
             _add_anno(signode, 'final')
         if self.override:
             _add_anno(signode, 'override')
-        for attr in self.attrs:
```
other-mickk: Attributes in a function type used to appear after `final` and `override` qualifiers.

Issue #4033, opened 2017-08-31
graphviz: Node with image="blah.png" doesn't render via sphinx (but ok in dot)


ghost:
Using dot, I can use a png image as a graphviz node, e.g in a .dot file containing
`graph { mynode [image="myimage.png"]}` and `dot -Tpng -O <file>` (or `-Tsvg`)
Resulting png/svg includes the image, sphinx's (latex or html) output doesn't. All other sphinx-generated graphviz stuff renders fine.

I tried adding graphviz_dot_args = ['-v'] to get some clues from the logs but no joy there either (doesn't seem to use the supplied option).

tk0miya:
In my short investigation:

* We need to parse dot files to know the path of images
* We have to rewrite the dot code if the graph depends on images in different directories (only same directory is allowed).

I think that is a bit tough work...

tk0miya:
Fixed by #5156!
Thank you for reporting :-)

PR #5156, opened 2018-07-09
graphviz: run dot in the document directory

rokroskar:
Subject: use the current document's directory as the build directory for the `dot` subprocess. 

### Feature or Bugfix
- Bugfix

### Purpose
- Allow the use of relative file paths for files included as a part of `dot` graphviz code

### Detail
- The current implementation runs `dot` wherever the sphinx command is run. This means that if images are used for nodes and specified as relative paths, they will not be found either during the build or while the built docs are viewed.
- Debugging this behavior was made more difficult by the lack of any logging from the `dot` subprocess. The change in the PR includes sending the contents of `stderr` to the logs. 

### Relates
- #4033

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -238,7 +238,8 @@ def render_dot(self, code, options, format, prefix='graphviz'):
     if format == 'png':
         dot_args.extend(['-Tcmapx', '-o%s.map' % outfn])
     try:
-        p = Popen(dot_args, stdout=PIPE, stdin=PIPE, stderr=PIPE)
+        p = Popen(dot_args, stdout=PIPE, stdin=PIPE, stderr=PIPE, cwd=path.dirname(
+            path.join(self.builder.srcdir, self.builder.current_docname)))
     except OSError as err:
         if err.errno != ENOENT:   # No such file or directory
             raise
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -235,11 +235,14 @@ def render_dot(self, code, options, format, prefix='graphviz'):
     dot_args = [graphviz_dot]
     dot_args.extend(self.builder.config.graphviz_dot_args)
     dot_args.extend(['-T' + format, '-o' + outfn])
+
+    cwd = path.dirname(
+        path.join(self.builder.srcdir, self.builder.env.docname))
+
     if format == 'png':
         dot_args.extend(['-Tcmapx', '-o%s.map' % outfn])
     try:
-        p = Popen(dot_args, stdout=PIPE, stdin=PIPE, stderr=PIPE, cwd=path.dirname(
-            path.join(self.builder.srcdir, self.builder.current_docname)))
+        p = Popen(dot_args, stdout=PIPE, stdin=PIPE, stderr=PIPE, cwd=cwd)
     except OSError as err:
         if err.errno != ENOENT:   # No such file or directory
             raise
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -155,7 +155,10 @@ class Graphviz(SphinxDirective):
                     line=self.lineno)]
         node = graphviz()
         node['code'] = dotcode
-        node['options'] = {}
+        node['options'] = {
+            'docpath': path.splitext(self.state.document.current_source)[0],
+        }
+
         if 'graphviz_dot' in self.options:
             node['options']['graphviz_dot'] = self.options['graphviz_dot']
         if 'alt' in self.options:
@@ -192,7 +195,9 @@ class GraphvizSimple(SphinxDirective):
         node = graphviz()
         node['code'] = '%s %s {\n%s\n}\n' % \
                        (self.name, self.arguments[0], '\n'.join(self.content))
-        node['options'] = {}
+        node['options'] = {
+            'docpath': path.splitext(self.state.document.current_source)[0],
+        }
         if 'graphviz_dot' in self.options:
             node['options']['graphviz_dot'] = self.options['graphviz_dot']
         if 'alt' in self.options:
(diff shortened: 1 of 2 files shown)


sphinx/ext/graphviz.py
```diff
@@ -238,7 +238,8 @@ def render_dot(self, code, options, format, prefix='graphviz'):
     if format == 'png':
         dot_args.extend(['-Tcmapx', '-o%s.map' % outfn])
     try:
-        p = Popen(dot_args, stdout=PIPE, stdin=PIPE, stderr=PIPE)
+        p = Popen(dot_args, stdout=PIPE, stdin=PIPE, stderr=PIPE, cwd=path.dirname(
+            path.join(self.builder.srcdir, self.builder.current_docname)))
```
tk0miya: `builder.current_docname` is only available in HTML builder. So please use `self.builder.env.docname` instead.

tk0miya: I feel constructing path in argument list is a bit ugly. How about make a new variable?

tk0miya (changes requested):
I think this is an incompatible change. So please add an entry to CHANGES file.

rokroskar:
@tk0miya thanks for the suggestions, they've been implemented. 

The tests now seem to be failing because `docname` is not found for any of the builders -- is there something else that needs to be modified in the test setup to make this available?

tk0miya:
OMG. I just noticed why `env.docname` is not initialized on writing phase.
Some builders concatenate multiple documents to single one on writing. For example, Single HTML builder processes only master document. So referring `env.docname` or `builder.current_docname` is not good way for such builders.

I think it would be better to save basedir of document or .dot file to graphviz node on parsing phase (`Graphviz.run()`).

Sorry for my wrong comment.

rokroskar:
Thanks @tk0miya - did you see my last commit? I tried to initialize the `temp_data` before write. Seems to mostly work though not all tests pass yet.

tk0miya:
Yes, of course. I'd seen your commit. But it does not effect to some builders as I commented.
For example, Single HTML builder overrides `write()` method. So the initialization code you added is never called.
https://github.com/sphinx-doc/sphinx/blob/58e3ef8be95b46d4f122cb7b088aab7b2b5da684/sphinx/builders/html.py#L1406-L1408

rokroskar:
Thanks for the clarification -- I implemented what you suggest in 5602051. Still some trouble with the failing tests which I can't reproduce locally.

sphinx/ext/graphviz.py
```diff
@@ -192,7 +195,9 @@ def run(self):
         node = graphviz()
         node['code'] = '%s %s {\n%s\n}\n' % \
                        (self.name, self.arguments[0], '\n'.join(self.content))
-        node['options'] = {}
+        node['options'] = {
+            'docpath': path.splitext(self.state.document.current_source)[0],
```
tk0miya: Sphinx usually uses `docname` for such case.
http://www.sphinx-doc.org/en/master/glossary.html#term-document-name

Could you use it instead of docpath? You can access it by `self.env.docname`.

sphinx/ext/graphviz.py
```diff
@@ -235,10 +240,13 @@ def render_dot(self, code, options, format, prefix='graphviz'):
     dot_args = [graphviz_dot]
     dot_args.extend(self.builder.config.graphviz_dot_args)
     dot_args.extend(['-T' + format, '-o' + outfn])
+
+    cwd = path.dirname(options.get('docpath', self.builder.srcdir))
```
tk0miya: In case of `docname`:
```
docname = options.get('docname', 'index')
cwd = path.dirname(path.join(self.env.srcdir, docname))
```

sphinx/ext/graphviz.py
```diff
@@ -252,6 +260,9 @@ def render_dot(self, code, options, format, prefix='graphviz'):
         # Graphviz may close standard input when an error occurs,
         # resulting in a broken pipe on communicate()
         stdout, stderr = p.communicate(code)
+        # print stderr if it is there
+        if stderr:
+            logger.warning(stderr.decode())
```
tk0miya: This new warning causes testing failed:
```
>       assert html_warnings == ""
E       AssertionError: assert '\x1b[31;01mW...b[39;49;00m\n' == ''
E         - WARNING: Fontconfig warning: ignoring C.UTF-8: not a valid language tag
E         - 
E         - WARNING: Fontconfig warning: ignoring C.UTF-8: not a valid language tag
E         - 
E         - WARNING: Fontconfig warning: ignoring C.UTF-8: not a valid language tag
E         - 
E         - WARNING: Fontconfig warning: ignoring C.UTF-8: not a valid language tag
E         - 
E         - WARNING: Fontconfig warning: ignoring C.UTF-8: not a valid language tag
E         - 

tests/test_ext_inheritance.py:44: AssertionError
```
https://circleci.com/gh/sphinx-doc/sphinx/2883?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Is this additional warning needed? IMO, it is not needed when succeeded.

tk0miya (changes requested):
Almost LGTM

sphinx/ext/graphviz.py
rokroskar: This was added to show errors/warnings from the `dot` run -- otherwise the user is left in the dark if something is wrong with the `dot` build. Perhaps it could be `info` instead of `warning`?

rokroskar:
@tk0miya implemented the change -- the tests seem to be passing now. One of the travis builds stalled, however.

sphinx/ext/graphviz.py
tk0miya: We already show stderr message if exist status is non-zero.
https://github.com/sphinx-doc/sphinx/blob/f697717c0e4870a2f5dd99dba42c785f10595db4/sphinx/ext/graphviz.py#L262-L264

My concerns are following:

* I don't know the message on that situation is important for users.
* The message is shown twice if dot command exists with non-zero.
* The message is shown without any headers. So users might not understand where the message comes from.

>Perhaps it could be info instead of warning?

As you said, this is "errors/wranings from dot". So WARNING level is better to me if we emits this.

tk0miya:
I just restarted Travis CI manually. So the failure will be resolved soon.

sphinx/ext/graphviz.py
rokroskar: Right, I see your point. I suppose for debugging the user could always copy whatever is inside the graphviz directive to a file and manually run dot. My intention was to remove the need for this extra step, but it's a bit of a hassle to make it work nicely within this context. I'll remove it.

tk0miya (approved):

tk0miya:
Thank you for updating. Finally I merged this into master branch!
Thank you for contribution!

rokroskar:
fantastic, thanks for your patience!

PR #5159, opened 2018-07-10
Fix links to sphinx-build man page

abdinoor:
Subject: Fix links to sphinx-build man page

- Bugfix

### Purpose
- Swap``program`` for ``doc`` links to sphinx-build man page.

### Detail
- Old links were broken because the man page is a document not a program.


tk0miya:
Thanks!

PR #5167, opened 2018-07-13
Fix formatting type annotations for tuples with more than two arguments

mitya57:
Currently in Python 3.6, Sphinx formats this function:
```python3
def foo() -> Tuple[int, str, int]:
    pass
```
in the following way (which can be seen in the [Travis log for the first commit in this PR][1]):

    () -> Tuple[[int, str], int]

This is because the tuple gets treated as a function, and the last `int` is treated like a return value.
This affects only `format_annotation_old` function. `format_annotation_new` which is used for Python 3.7 is not affected.

This pull request fixes this bug, and adds a test.

There is a `hasattr(annotation, '__tuple_params__')` clause below in that function. I am not sure for which Python version it is because all versions I tested (2.7, 3.5, 3.6, 3.7) do not have this attribute. Maybe it is for 3.4? Anyway I added an opposite condition (`not hasattr`) to be on the safe side.

[1]: https://travis-ci.com/mitya57/sphinx/jobs/134445530

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -521,6 +521,13 @@ class Signature(object):
 
         if annotation.__module__ == 'builtins':
             return annotation.__qualname__  # type: ignore
+        elif (hasattr(typing, 'TupleMeta') and
+              isinstance(annotation, typing.TupleMeta) and  # type: ignore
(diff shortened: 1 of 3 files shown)


tk0miya (approved):
LGTM!

PR #5174, opened 2018-07-15
Fix #5164: incremental build has broken with external source_parser

### LINKED ISSUE #5164: All sources are re-built, even if nothing has changed (using nbsphinx extension)
This problem was introduced in PR #4966. Before that, everything was working fine.

This only happens when the `nbsphinx` extension is used (but maybe other extensions are affected, too?).

This can be reproduced by building the docs for the `nbsphinx` module.
Installation instructions: https://github.com/spatialaudio/nbsphinx/blob/master/CONTRIBUTING.rst

In short:

```
git clone https://github.com/spatialaudio/nbsphinx.git
cd nbsphinx
python3 -m pip install -e . --user
python3 setup.py build_sphinx
```

All source files (including Jupyter notebooks) are built at this point.

If I re-build immediately after that (without touching any files) ...

```
python3 setup.py build_sphinx
```

... all source files are built again.

#### Expected results

The second time the source files should not be processed again.
Instead, there should be this message:

```
updating environment: 0 added, 0 changed, 0 removed
no targets are out of date.
```

- https://github.com/spatialaudio/nbsphinx

### Environment info
- OS: Linux
- Python version: Python 3.6.6rc1+
- Sphinx version: `master` branch, every commit after PR #4966 (commit 3ffde92c54c6e9d594b2a41a365a90dc0dcbac6c)

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5164

### DIFF
--- a/sphinx/registry.py
+++ b/sphinx/registry.py
@@ -39,6 +39,7 @@ if False:
     from docutils.transforms import Transform  # NOQA
     from sphinx.application import Sphinx  # NOQA
     from sphinx.builders import Builder  # NOQA
+    from sphinx.config import Config  # NOQA
     from sphinx.domains import Domain, Index  # NOQA
     from sphinx.environment import BuildEnvironment  # NOQA
     from sphinx.ext.autodoc import Documenter  # NOQA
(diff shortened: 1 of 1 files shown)


mgeier:
This solves the problem for me, thanks a lot!

Issue #5161, opened 2018-07-11
 mkdir(name, mode) NotADirectoryError: [Errno 20] Not a directory: '..doc/_build/html/_static/fonts/RobotoSlab'


disencd:
Subject:  mkdir(name, mode) NotADirectoryError: [Errno 20] Not a directory: '..doc/_build/html/_static/fonts/RobotoSlab'

### Problem
```
 mkdir(name, mode) NotADirectoryError: [Errno 20] Not a directory: '..doc/_build/html/_static/fonts/RobotoSlab'
```

#### Procedure to reproduce the problem
```
 mkdir(name, mode) NotADirectoryError: [Errno 20] Not a directory: '..doc/_build/html/_static/fonts/RobotoSlab'
```

#### Error logs / results
```
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... WARNING: search index couldn't be loaded, but not all documents will be built: the index will be incomplete.
done
writing output... [100%] library                                                                                                                              
generating indices... genindex py-modindex
writing additional pages... search
copying static files... 
Exception occurred:
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
NotADirectoryError: [Errno 20] Not a directory: '/Users/ddevassy/Documents/logi/Seam/lci/doc/_build/html/_static/fonts/RobotoSlab'
The full traceback has been saved in /var/folders/0n/by0ts5_d6qg7yfry647x5fxh2ngr3b/T/sphinx-err-r2et3w57.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 2
```

- <public link of unexpected result if you have>

#### Expected results
<Describe what to actually do>

### Reproducible project / your project
- <link to your project, or attach zipped small project sample>

### Environment info
- OS: Mac
- Python version: 3.6
- Sphinx version: sphinx-quickstart 1.7.5
- <Extra toolsGoogle Chrome

disencd:
[sphinx-err-r2et3w57.log](https://github.com/sphinx-doc/sphinx/files/2182988/sphinx-err-r2et3w57.log)

Attaching the log

tk0miya:
[...]

PR #5175, opened 2018-07-15
Fix html: crashes if copying static files are failed (refs: #5161)

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5161

### DIFF
--- a/sphinx/builders/html.py
+++ b/sphinx/builders/html.py
@@ -740,85 +740,93 @@ class StandaloneHTMLBuilder(Builder):
                 try:
                     copyfile(path.join(self.srcdir, src),
                              path.join(self.outdir, '_downloads', dest))
-                except Exception as err:
+                except EnvironmentError as err:
                     logger.warning('cannot copy downloadable file %r: %s',
                                    path.join(self.srcdir, src), err)
 
     def copy_static_files(self):
         # type: () -> None
-        # copy static files
-        logger.info(bold('copying static files... '), nonl=True)
-        ensuredir(path.join(self.outdir, '_static'))
-        # first, create pygments style file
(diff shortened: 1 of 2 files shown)


sphinx/builders/html.py
```diff
@@ -746,79 +746,85 @@ def to_relpath(f):
...
+            if self.config.html_favicon:
+                iconbase = path.basename(self.config.html_favicon)
+                icontarget = path.join(self.outdir, '_static', iconbase)
+                if not path.isfile(path.join(self.confdir, self.config.html_favicon)):
+                    logger.warning('favicon file %r does not exist', self.config.html_favicon)
+                elif not path.isfile(icontarget):
+                    copyfile(path.join(self.confdir, self.config.html_favicon),
+                             icontarget)
+            logger.info('done')
+        except Exception as err:
```
shimizukawa: If we do not want to catch unexpected exceptions such as NameError that is caused by programming error, it would be nice to catch `OSError` instead of `Exception`.
https://docs.python.org/3.7/library/exceptions.html#exception-hierarchy

shimizukawa (approved):
LGTM with nits;

Although the range of try is wide, I think that there is no problem. Even if we caught an exception in detail here, there seems to be little merit for the trouble.

sphinx/builders/html.py
tk0miya: In py2, `IOError` is independent exception class. So it is better to catch `EnvironmentError` instead. It is common parent class of `IOError` and `OSError` in py2. And it is an alias of `OSError` in py3.

tk0miya: done

shimizukawa: Sounds good!

tk0miya:

PR #5177, opened 2018-07-15
Fix #5176

### LINKED ISSUE #5176: LaTeX: indexing of terms containing ``@`` character, or ``!``, or ``"`` fail
Consider this source:

```
==========
Index Test
==========

This is problematic :index:`@` even with makeindex. No error but not
indexed...
```

Then `make latexpdf` silently fails to index correctly. The `makeindex` log contains

```
This is makeindex, version 2.15 [TeX Live 2018] (kpathsea + Thai support).
Scanning style file ./python.ist.......done (7 attributes redefined, 0 ignored).
Scanning input file FOO.idx...
!! Input index error (file = FOO.idx, line = 1):
   -- Illegal null field.
.done (1 entries accepted, 1 rejected).
Sorting entries...done (0 comparisons).
Generating output file FOO.ind....done (8 lines written, 0 warnings).
Output written in FOO.ind.
Transcript written in FOO.ilg.
```

Also this fails:

```
==========
Index Test
==========

This is problematic due to exclamation mark

.. index::
   operator: !=

There are eight comparison operations in Python. (:file:`library/stdtypes.rst`)
```

This fails to get indexed correctly in PDF output, the `makeindex` log file contains similar

```
This is makeindex, version 2.15 [TeX Live 2018] (kpathsea + Thai support).
Scanning style file ./python.ist.......done (7 attributes redefined, 0 ignored).
Scanning input file FOO.idx...
!! Input index error (file = FOO.idx, line = 1):
   -- Illegal null field.
```

The above is real life example from CPython Docs...

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.7.x
- TeXLive 2018, makeindex

jfbu:
fix: in LaTeX, characters ``@``, ``!`` and ``"`` needs so-called "quoting". Works both with makeindex and xindy. The ``|`` is anyhow already TeX-escaped.

Documentation is found in chapter on makeindex in LaTeX companion.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -44,6 +44,7 @@ Bugs fixed
   sorting
 * #5139: autodoc: Enum argument missing if it shares value with another
 * #4946: py domain: rtype field could not handle "None" as a type
+* #5176: LaTeX: indexing of terms containing ``@``, ``!``, or ``"`` fails
 
 Testing
 --------
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1955,6 +1955,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
             value = self.encode(value)
             value = value.replace(r'\{', r'{\sphinxleftcurlybrace}')
             value = value.replace(r'\}', r'{\sphinxrightcurlybrace}')
+            value = value.replace('@', '"@')
             return value
 
         if not node.get('inline', True):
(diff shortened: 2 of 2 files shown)


tk0miya (approved):

jfbu:
@tk0miya thanks for quick review! I will merge after testing completes.

Issue #4272, opened 2017-12-03
PDF builds of French projects have issues with XeTeX from TL2016 or 2017


jfbu:
### Problem

- the polyglossia French module has not been modified upstream to adapt to internal changes at xetex 0.99994 or later, in practice this means for people using TeXLive 2016 (it has xetex 0.99996 whereas TL2015 has xetex 0.99992) or derived TeX distributions, or TeXLive 2017.

- for reference on this, see http://tug.org/pipermail/xetex/2017-December/027215.html

- some expected typographical rules of French are broken by this.

#### Procedure to reproduce the problem

file `index.rst`:

```rst
Bienvenue
=========

Un point-virgule ; doit être précédé par une espace fine typographique.

Un point-virgule; doit être précédé par une espace fine typographique.
```

in `conf.py`:

```
language = 'fr'
latex_engine = 'xelatex'
```

Then do `make latexpdf`.

#### Error logs / results

With TeXLive 2016 (`XeTeX 3.14159265-2.6-0.99996 (TeX Live 2016)`) PDF looks like this:

![capture d ecran 2017-12-03 a 14 34 21](https://user-images.githubusercontent.com/2589111/33525906-6b7b71da-d838-11e7-9051-a4dea08619b9.png)

The two lines should be identical; but the top one, whose source had a space before the `;` is wrong. It uses a too big space in output.

#### Expected results

Here is with TeXLive 2013's latex packages and xetex binary (`XeTeX 3.1415926-2.5-0.9999.3-2013061817 (TeX Live 2013)`)

![capture d ecran 2017-12-03 a 14 33 24](https://user-images.githubusercontent.com/2589111/33525904-5971520c-d838-11e7-965f-b2d9fe981f33.png)

As is expected and correct the two lines are identical with a thin space before the semi-colon.

### Environment info
- OS: Mac
- Python version: 3.5.4
- Sphinx version: 1.6.6+ (stable branch)
- TeXLive 2013 for good result, TeXLive 2016 for bad result.

### How to fix ?

Possibly, Sphinx should use babel + french and not polyglossia + french when latex engine is `xelatex` or `lualatex`.

Memo:

- independently of this polyglossia-french bug, the babel-french has more extensive support of French typographical rules, in particular for lists -- this babel-french customization is lacking in some aspects, but it does comply rather closely to official typography whereas polyglossia-french does _nothing at all_ in this area.

[...]

PR #5178, opened 2018-07-15
LaTeX: for French and xelatex, use babel not polyglossia (fix: #4272, #5130)

jfbu:
Unfortunately, xelatex currently has problems with polyglossia+French. Nowadays, LateX/babel+French is more actively maintained than polyglossia+French.

Some aspects were raised as #4272, and more recently at #5130, it showed as failure to build the PDF Docs for CPython in French translation.

This PR would solve #5130, although only with also #5134, for 1.8 release, for correct indices of terms containing French characters like ``é``, will #5130 be completely solved.


tk0miya (approved):

jfbu:
Thanks @tk0miya for quick review. I merge now for 1.7.6.

Issue #3742, opened 2017-05-12
RTD builds of Sphinx own PDF docs do not do enough pdflatex runs


jfbu:
From http://readthedocs.org/projects/sphinx/builds/5413951/ one sees

```
pdflatex -interaction=nonstopmode /home/docs/checkouts/readthedocs.org/user_builds/sphinx/checkouts/1.6-release/doc/_build/latex/sphinx.tex
makeindex -s python.ist sphinx.idx
pdflatex -interaction=nonstopmode /home/docs/checkouts/readthedocs.org/user_builds/sphinx/checkouts/1.6-release/doc/_build/latex/sphinx.tex
```

but there should be 2 `pdflatex` runs after the `makeindex`: else there will (and is) no ``index`` entry at end of table of contents, as one can confirm at https://media.readthedocs.org/pdf/sphinx/1.6-release/sphinx.pdf, and this applies also to the bookmarks of the pdf.

I think the back links from the index are ok from `1 pdflatex + makeindex + 2 pdflatex` but I have not checked thoroughly.

Best for RTD builds would be to use `make latexpdf` from Sphinx, which uses `latexmk` rather than the above recipes.

This is not Sphinx issue per se, but I did not know where better to report. The remark will apply to **any** project having a general index if PDF build does only `1 pdflatex + makeindex + 1 pdflatex`. (in a clean repertory) I am not familiar with how RTD is configured for Sphinx doc.

tk0miya:
I might be wrong, but the source code of readthedocs is hosted at https://github.com/rtfd/readthedocs.org .
So you should report this into the repository.

Thanks,

jfbu:
This appears to be indeed a [general RTD issue](https://github.com/rtfd/readthedocs.org/issues/1557). I will add a comment to the discussion there.

edit I moved my comment to https://github.com/rtfd/readthedocs.org/pull/1583#issuecomment-301231607 which seems better suited.

jfbu:
Temporary workaround at 6422f7b. Opened a new ticket https://github.com/rtfd/readthedocs.org/issues/2857 at RTD.

(fixed flake8 at b18a3f82b476e04a88e641134919b643facb9ea1)

PR #5181, opened 2018-07-16
LaTeX: extend to all projects the #3742 fix for PDF builds at RTD

jfbu:
Refs: https://github.com/rtfd/readthedocs.org/issues/2857

Sphinx own documentation already uses this kind of workaround. This PR extends it to all projects.

This is a bugfix, but not to our project, rather to RTD's one.

### Relates
- #3742
- https://github.com/rtfd/readthedocs.org/issues/2857


jfbu:
@tk0miya arguably this could be rebased to 1.7 for 1.7.6 or 1.7.7, but it is not urgent in my view as the better thing would be for RTD to use our LaTeX Makefile and not do only their 2 pdflatex runs which is insufficient in general. Hence I set the milestone to 1.8.

tk0miya (approved):
LGTM. It's reasonable and helpful!

PR #5192, opened 2018-07-17
LaTeX: for Cyrillic and with xindy, get Latin letter groups too

jfbu:
Refs: #5134

This allows to get Latin names indexed in their own letter groups, after the Cyrillic index.

This is still partial in so far as words starting with letters such as ``é``, which are beyond ascii, will end up in the "non-alphabetical" group. But English acronyms or words like ``eLIBRARY.ru`` will be correctly indexed, in my testing.

It is surely be possible to extend this to iso-latin-1, but then I am not sure about how to order letters, and the xindy set of rules would become more complicated.

**edit** the PR has been updated to cover all of Latin 1, 2, 3 for Latin names in index of Cyrillic document.


jfbu:
Memo: it might be better not to have all the ``:again`` switches. To test this, one only needs to go to LaTeX build repertory, modify the ``sphinx-basiclatin.xdy`` therein to remove all the ``:again``, and then do ``make clean all-pdf``. Doing so should intuitively speed-up a bit processing and possibly better order Latin names. I did not test much, feedback would be appreciated.

**edit** this comment is now irrelevant in view of subsequent commits added to initial PR

tk0miya (approved):

jfbu:
@tk0miya thanks for review. While being offline I developed a more complete solution which I pushed now to that PR. With this, Cyrillic documents can have index with both Cyrillic words and Latin ones, and inclusive of fancy letters like `é` or even `œ`.

jfbu:
@mitya57 with this PR, you can index in your Russian document also Latin names, even containing letters such as `é` or `œ`.... the Latin names are listed at end of index, with their initials A...Z, after the Russian names. (document should contain `\usepackage[T1,T2A]{fontenc}` as T1 is needed for the French letters such as `œ` for example)

mitya57:
I won't be able to test it soon, but thank you very much anyway!

sphinx/texinputs/LICRlatin2utf8.xdy
```diff
@@ -0,0 +1,236 @@
...
+;;   to preamble of LaTeX document.  However this character is not supported
+;;   by the TeX "times" font used by default by Sphinx for pdflatex engine.
+;;
+;; - ẞ needs \DeclareUnicodeCharacter{1E9E}{\SS} (but ß needs no extra set-up).
+;;
+;; - U+02DB (˛) and U+02D9 (˙) are also not supported by inputenc
+;;   out of the box and require
+;;       \DeclareUnicodeCharacter{02DB}{\k{}}
+;;       \DeclareUnicodeCharacter{02D9}{\.{}}
+;;   to be added to preamble.
```
jfbu, line 38: MEMO: dev repo of LaTeX contains a commit https://github.com/latex3/latex2e/commit/ed72e7171789b792a07f040d6e5d07f96623aec8 which adds these 4 missing characters to the list of those supported out of the box by pdflatex + utf8 inputenc

PR #5194, opened 2018-07-19
Closes #4614: sphinx-build: Add :option:`--keep-going` option to show all warnings

### LINKED ISSUE #4614: Sphinx should show all warnings with -W
When using the `-W` flag to turn warnings into errors, the first warning will result in an error and cause the build to stop. It would be nice if all warnings were shown, as without using -W. 

This was first mentioned at https://github.com/sphinx-doc/sphinx/issues/4606

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #4614


asmeurer:
I think it should be the default (with `--no-keep-going`) being the flag, but I don't feel too strongly as long as the option is there.

tk0miya:
I'd like to keep compatibility. So I'd not like to change the behavior of `-N` option.

@shimizukawa any comments?

asmeurer:
I don't really see how it's a compatibility break. The only difference here is that broken builds might take longer.

shimizukawa (approved):
LGTM!

tk0miya:
>The only difference here is that broken builds might take longer.

AFAIK, there are very large documents using Sphinx. It takes a few minutes on building.
I'm afraid this change affects to the author of them.

So I'd like to add a new option for keep-going.

tk0miya:
If agreed, I will merge this to 1.8 release soon.

asmeurer:
Yes, the SymPy docs are one of them (the original motivator of this issue). It takes over 3 minutes to build them from scratch on my laptop.

Of course, the whole reason we use -W is that we never want the docs to have any warnings or errors. So the in the ideal case the docs work as expected and take the full time to build.

tk0miya:
Merged. Thank you for comment!

Issue #5186, opened 2018-07-16
JavaScript tests?


TimKam:
According to the GitHub stats, 26.2% of the Sphinx code is JS. This metric is inaccurate, as it includes code like stemmers that is copied from other projects and requires little maintenance effort. Still, it probably makes sense to add JavaScript tests to avoid regressions and increase maintainability in the long run.

If deemed beneficial, I would gladly work on introducing JavaScript tests and propose proceeding as follows:

1. Reduce the code that is contained in JavaScript "templates" to global configuration variables and move the rest of the code into ordinary `.js` files.
2. Write a handful of example JavaScript tests.
3. Set up JavaScript CI with Travis or Circle CI.
4. Gradually increase test coverage.

tk0miya:
+1: it's very helpful

TimKam:
As the branch is merged, I'll close this issue. I will add more tests and create new issues for each PR to have a 1-1 Issue - PR mapping.

PR #5207, opened 2018-07-22
5186 set up JavaScript test infrastructure

TimKam:
Subject: Set up JavaScript test infrastructure

### Feature or Bugfix
- Maintenance

### Purpose
It would be good to have tests for Sphinx's JavaScript code. The goal of this PR is to set up the necessary infrastructure for this. Later, we can gradually increase test coverage.

### Detail
- Refactor JavaScript code to minimize code in template files
- Configure karma.js and write first test
- Configure travis-CI

### Notes
- It looks strange at first glance, but `package-lock.json` should in fact be committed, see: [npm documentation](https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md#description)

### Relates
- #5186

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4456 @@
+{
+  "name": "sphinx",
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "accepts": {
+      "version": "1.3.5",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
+      "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
+      "dev": true,
+      "requires": {
+        "mime-types": "2.1.19",
+        "negotiator": "0.6.1"
+      }
+    },
+    "addressparser": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/addressparser/-/addressparser-1.0.1.tgz",
+      "integrity": "sha1-R6++GiqSYhkdtoOOT9HTm0CCF0Y=",
+      "dev": true,
+      "optional": true
+    },
+    "after": {
+      "version": "0.8.2",
+      "resolved": "https://registry.npmjs.org/after/-/after-0.8.2.tgz",
+      "integrity": "sha1-/ts5T58OAqqXaOcCvaI7UF+ufh8=",
+      "dev": true
+    },
+    "agent-base": {
+      "version": "4.2.1",
+      "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.1.tgz",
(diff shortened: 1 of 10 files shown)


TimKam:
@tk0miya Can you give feedback? If this PR goes into the right direction, I can continue adding tests.

tk0miya:
It seems there are no test runners for `node_js` in `scripts:` section. Would you like to add them in another PR? If intended, I'm okay to this. Please go ahead.

TimKam:
The default Node.js build script is `npm test`, see [here in the Travis-CI docs](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Default-Build-Script). That's why I think it's not necessary. The test runs without it. Or am I missing something?

tk0miya:
Please read the log for node_js job.
https://travis-ci.org/sphinx-doc/sphinx/jobs/406888385

We've already overrided it by travis.yml. So default setting is not used. I think you need to call manually like this:
```
  - if [ $RUN_TOX = false ]; then npm test; fi
```
I know `$RUN_TOX` is not good naming for such case.

TimKam:
Sorry for missing this. Now it works. If you think it's okay to merge, we could do so and I'll add new PRs with more tests every now and then.

tk0miya (approved):
LGTM! Let's merge this :-)

package-lock.json
```diff
@@ -0,0 +1,4456 @@
...
+      "version": "7.1.2",
+      "resolved": "https://registry.npmjs.org/glob/-/glob-7.1.2.tgz",
+      "integrity": "sha512-MJTUg1kjuLeQCJ+ccE4Vpa6kKVXkPYJ2mOCQyUuKLcLQsdrMCpBPUi8qVE6+YuaJkozeA9NusTAw3hLr8Xe5EQ==",
+      "dev": true,
+      "requires": {
+        "fs.realpath": "1.0.0",
+        "inflight": "1.0.6",
+        "inherits": "2.0.3",
+        "minimatch": "3.0.4",
+
```
tk0miya: I recieved security alert from github.

>We found a potential security vulnerability in a repository for which you have been granted security alert access.
>
>Known moderate severity security vulnerability detected in hoek < 4.2.1defined in package-lock.json.
>package-lock.json update suggested: hoek ~> 4.2.1.

@TimKam Could you check this please? I don't know about this library so much. But it seems too old.

TimKam: There is a problem with dependencies of [karma.js](https://github.com/karma-runner/karma), the test runner, see: https://github.com/karma-runner/karma/issues/2994. It is not possible to update the dependency in the test setup. However, I am confident the vulnerabilities cannot be exploited in our usage scenario, as the libraries are not used in a security-critical context. I suggest it is okay to wait for Karma to update the dependencies and to up our Karma version subsequently.

tk0miya: Thank you for explanation. I understand. And I agree with you.
Okay, let's keep as is :-)

PR #4076, opened 2017-09-20
New conf.py directives autodoc_exclude_members + autodoc_special_members

peterjc:
### Feature or Bugfix
- Feature

### Purpose
- Adds ``conf.py`` setting ``autodoc_exclude_members`` (*update*) and ``autodoc_special_members``

### Relates
- #4057 and #4075 for trying to use ``:exclude-members:`` with apidoc and autodoc.

Issue #4075, opened 2017-09-20
limitations to autodoc_default_flags in conf.py


peterjc:
### Problem

http://www.sphinx-doc.org/en/stable/ext/autodoc.html#confval-autodoc_default_flags says:

> *autodoc_default_flags*
> This value is a list of autodoc directive flags that should be automatically applied to all autodoc directives. The supported flags are 'members', 'undoc-members', 'private-members', 'special-members', 'inherited-members' and 'show-inheritance'.

I want to be able to set values in ``conf.py`` equivalent to adding additional configuration entries under ``:autodoc:`` including some which have arguments, including sorting members according to the source code, e.g.:

```
    :member-order: bysource
    :special-members: __init__
    :exclude-members: __dict__,__weakref__
```

While member order seems to be covered specifically via ``autodoc_member_order = ...`` in ``conf.py``, this seems to be a special case: http://www.sphinx-doc.org/en/stable/ext/autodoc.html#confval-autodoc_member_order

In general it seems I cannot give optional arguments to ``autodoc`` flags via ``conf.py``.

See also #4057, cannot give optional arguments to ``autodoc`` flags via ``sphinx-apidoc`` 

#### Procedure to reproduce the problem

Reading the code, these are the only possible flags which can be given using ``autodoc_default_flags = [...]`` in ``conf.py``, this is missing ``member-order`` and ``exclude-members`` which I want to use.

```python
    # flags that can be given in autodoc_default_flags
    _default_flags = set([
        'members', 'undoc-members', 'inherited-members', 'show-inheritance',
        'private-members', 'special-members',
    ])
```

Source: https://github.com/sphinx-doc/sphinx/blob/1.6.3/sphinx/ext/autodoc.py#L1769

Moreover, the values from the ``conf.py`` file's ``autodoc_default_flags = ...`` entry are treated as a simple list - there is no scope for arguments:

https://github.com/sphinx-doc/sphinx/blob/1.6.3/sphinx/ext/autodoc.py#L1812

#### Expected results

I can achieve the desired results by editing the ``sphinx-apidoc`` generated RST files before running Sphinx, but I wanted to avoid this step via configuration instead.

One suggested syntax, either allow a list/tuple/set as now:

```
autodoc_default_flags = ['members', 'undoc-members', 'special-members', 'show-inheritance']
```

Or, a dict where keys are a string giving the arguments, and I suggest the empty string or ``None`` if no arguments are needed:

```python
autodoc_default_flags = {
[...]

PR #5215, opened 2018-07-26
Allow setting audodoc default options in conf

LewisHaley:
Subject: Set global default values for directives in config

### Feature or Bugfix
- Feature

### Purpose
- In RST files, you can set, for example: `:special-members: __init__, __iter__`. In config, you can set this flag, but only to include *all* special members, by including it in a list `autodoc_default_flags`. This PR make it so you can set the same values in global config as you can individually in RST files.

### Detail
- I wanted to write a test for this but I couldn't get the existing setup in `test_autodoc` to work for me, because `autodoc.directives.process_documenter_options` wants a `options` object which in the unit tests is mocked to a `Struct`, which doesn't work with `process_documenter_options`.
- I have managed to use this code to produce the docs for my own project, and it worked as expected using
```python
autodoc_default_values = {
    'members': None,
    'undoc-members': None,
    'special-members': None,
    'show-inheritance': None,
    'member-order': 'bysource',
    'exclude-members': '__dict__,__weakref__,__module__',
}
```
in conf.py

### Relates
- Issue: https://github.com/sphinx-doc/sphinx/issues/4075 (see final comment)
- Overlaps/replaces https://github.com/sphinx-doc/sphinx/pull/4076

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -15,6 +15,7 @@ import inspect
 import re
 import sys
 import warnings
+from typing import Any
 
 from docutils.statemachine import ViewList
 from six import iteritems, itervalues, text_type, class_types, string_types
@@ -41,6 +42,7 @@ if False:
     from docutils import nodes  # NOQA
     from docutils.utils import Reporter  # NOQA
     from sphinx.application import Sphinx  # NOQA
+    from sphinx.config import Config  # NOQA
     from sphinx.environment import BuildEnvironment  # NOQA
     from sphinx.ext.autodoc.directive import DocumenterBridge  # NOQA
 
@@ -679,8 +681,13 @@ class Documenter(object):
 
         # remove members given by exclude-members
         if self.options.exclude_members:
-            members = [(membername, member) for (membername, member) in members
-                       if membername not in self.options.exclude_members]
+            members = [
+                (membername, member) for (membername, member) in members
+                if (
+                    self.options.exclude_members is ALL or
+                    membername not in self.options.exclude_members
+                )
+            ]
 
         # document non-skipped members
         memberdocumenters = []  # type: List[Tuple[Documenter, bool]]
@@ -1528,6 +1535,34 @@ def autodoc_attrgetter(app, obj, name, *defargs):
     return safe_getattr(obj, name, *defargs)
 
 
+def convert_autodoc_default_flags(app, config):
+    # type: (Sphinx, Config) -> None
+    """This converts the old list-of-flags (strings) to a dict of Nones."""
+    if isinstance(config.autodoc_default_flags, dict):
+        # Already new-style
+        return
+
+    elif not isinstance(config.autodoc_default_flags, list):
+        # Not old-style list
+        logger.error(
+            __("autodoc_default_flags is invalid type %r"),
+            config.autodoc_default_flags.__class__.__name__
+        )
+        return
+
+    autodoc_default_flags = {}  # type: Dict[unicode, Any]
+    for option in config.autodoc_default_flags:
+        if isinstance(option, string_types):
+            autodoc_default_flags[option] = None
+        else:
+            logger.warning(
+                __("Ignoring invalid option in autodoc_default_flags: %r"),
+                option
+            )
+
+    config.autodoc_default_flags = autodoc_default_flags  # type: ignore
+
+
 def setup(app):
     # type: (Sphinx) -> Dict[unicode, Any]
     app.add_autodocumenter(ModuleDocumenter)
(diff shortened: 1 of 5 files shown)


tk0miya:
As discussed in #4076, it would be better to enhance `autodoc_default_flags` to accept both list and dict:
```
autodoc_default_flags = [
    'members', 'undoc-members', ...
]

or

autodoc_default_flags = {
    'members': None,
    'undoc-members': None,
    ...
    'exclude-members': '__dict__,__weakref__,__module__',
}
```

Since 1.8, Sphinx provides `config-inited` event. I think it is good for conversion.
For example, HTML builder installs a converter for `html_js_files`:
https://github.com/sphinx-doc/sphinx/blob/8ddca8c62051d06654b22746a6f2d07a2e0ce750/sphinx/builders/html.py#L1605-L1620

tk0miya:
I just made #5223 for better testing (I'll merge it soon). I hope this helps this case.

LewisHaley:
I didn't like the idea of overloading `autodoc_default_flags` because (to me, at least) "flags" implies on/off only (which is how the option currently behaves - if "members" is present then "members = True", else "members = False"). I think the idea of giving arguments to "flags" is perhaps more confusing than just adding a new option. The old option could always be deprecated eventually.

But I'm happy to follow your suggestion of using the converter function with the original option name.
Thanks for the unittest - I'll try to follow it's example.

tk0miya:
OMG, I just made #5277 as an alternative of this PR. Okay, let's discuss what is the best way to configure the default settings. I'll reconsider this later (It's time to sleep in japan 😄 ).

@shimizukawa Please let me know your opinion.

LewisHaley:
I've made the changes to use `config-inited` and the conversion hook.
I found a latent bug in the autodoc tests when I added some new tests, which I've fixed: https://github.com/sphinx-doc/sphinx/pull/5215/commits/50ad4f184d5341c8ed7588c6cfd5b39b7d8a26f7.

I've added a load of tests for this, partly because the different combinations of options can be quite surprising. I found testing the diff between option on/off was the best for showing what the *full* effect is of each of the options/values.

shimizukawa:
+1 for `autodoc_default_values`. `autodoc_default_options` is better to me because `:members:` is a option of `autodoc` directive.

LewisHaley:
@tk0miya Just checking in about whether you're still considering this PR.

tk0miya:
Okay, let's make it an alias of `autodoc_default_flags` on next step.

sphinx/ext/autodoc/__init__.py
```diff
@@ -41,6 +42,8 @@
     from docutils import nodes  # NOQA
     from docutils.utils import Reporter  # NOQA
     from sphinx.application import Sphinx  # NOQA
+    from sphinx.config import Config  # NOQA
+    from sphinx.config import Config  # NOQA
```
tk0miya: It seems duplicated.

tests/test_autodoc.py
```diff
@@ -1411,22 +1417,248 @@ def test_partialmethod(app):
...
+        'members': None,
+        'undoc-members': None,
+    }
     actual = do_autodoc(app, 'class', 'target.EnumCls')
     assert '   .. py:attribute:: EnumCls.val1' in actual
     assert '   .. py:attribute:: EnumCls.val4' in actual
+
+    # with :special-members:
+    # Note that :members: must be *on* for :special-members: to work.
+    # TODO Is this a bug?
```
tk0miya: AFAIK, it's well-known behavior.

tests/test_autodoc.py
```diff
@@ -1411,22 +1417,248 @@ def test_partialmethod(app):
...
+    assert '      Create a new `CustomIter`.' in actual
+    assert '   .. py:method:: CustomIter.__iter__()' in actual
+    assert '      Iterate squares of each value.' in actual
+    if not IS_PYPY:
+        assert '   .. py:attribute:: CustomIter.__weakref__' in actual
+        assert '      list of weak references to the object (if defined)' in actual
+
+
+@pytest.mark.sphinx('html', testroot='ext-autodoc')
+@pytest.mark.parametrize('target, options, expected_diff', [
```
tk0miya: I feel this parameterized test is hard to understand...
Could you change this to simple testcases?

In addition, is doing diff important for this case? We've already done testing for each options. So I think the main purpose of this case is `autodoc_default_flags` modifies the options of each autodoc results.

tk0miya (changes requested):
LGTM with nits

sphinx/ext/autodoc/__init__.py
LewisHaley: fixed

tests/test_autodoc.py
LewisHaley: removed TODO comment

tests/test_autodoc.py
LewisHaley: Done. I found the diffs useful for myself as I wasn't completely familiar with what each option did but, as you say, there are tests for this already. I've left the tests quite verbose but they don't use `parametrize` or look at the diffs.

LewisHaley:
Re. future work about making an alias for the option so its name makes more sense, could I suggest as well that the accepted types for the dict is more tightly defined. For example, `:undoc-members:` would make more sense as only allowing True/False, while `:exclude-members:` only make sense as a list of strings (if the list is empty, nothing is excluded).

`:members:`, `:special-members:` are more difficult because they need a "say yes to everything" as well as a limiting list. You could use `{'members': ["*"]}` (meaning glob) to include everything? This way, the values in the dictionary are *always* a list of strings.

Just some thoughts I've had. :)

tk0miya (approved):

tk0miya:
Just merged!
Thank you for contribution!

PR #5227, opened 2018-07-29
refactor: Move repository of changesets to domain from env

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Move repository of changesets to domain from env

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -84,9 +84,11 @@ Deprecated
 * ``env._read_parallel()`` is deprecated
 * ``env.write_doctree()`` is deprecated
 * ``env._nitpick_ignore`` is deprecated
+* ``env.versionchanges`` is deprecated
 * ``env.dump()`` is deprecated
 * ``env.dumps()`` is deprecated
 * ``env.topickle()`` is deprecated
+* ``env.note_versionchange()`` is deprecated
 * ``sphinx.writers.latex.Table.caption_footnotetexts`` is deprecated
 * ``sphinx.writers.latex.Table.header_footnotetexts`` is deprecated
 * ``sphinx.writers.latex.LaTeXTranslator.footnotestack`` is deprecated
--- a/CHANGES
+++ b/CHANGES
@@ -118,6 +118,7 @@ Deprecated
 * ``sphinx.ext.mathbase.is_in_section_title()`` is deprecated
 * ``sphinx.ext.mathbase.MathDomain`` is deprecated
 * ``sphinx.ext.mathbase.setup_math()`` is deprecated
+* ``sphinx.directives.other.VersionChanges`` is deprecated
 * ``sphinx.highlighting.PygmentsBridge.unhighlight()`` is deprecated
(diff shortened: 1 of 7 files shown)


sphinx/directives/other.py
```diff
@@ -261,8 +261,8 @@ def run(self):
                                                 classes=['versionmodified']),
                                    translatable=False)
             node.append(para)
-        # XXX should record node.source as well
```
tk0miya: Note: This is already done on `set_source_info()` calling above.

tk0miya:

CHANGES
```diff
@@ -83,9 +83,11 @@ Deprecated
 * ``env._read_parallel()`` is deprecated
 * ``env.write_doctree()`` is deprecated
 * ``env._nitpick_ignore`` is deprecated
+* ``env.versionchanges`` is deprecated
 * ``env.dump()`` is deprecated
 * ``env.dumps()`` is deprecated
 * ``env.topickle()`` is deprecated
+* ``env.note_versionchange()`` is deprecated
```
shimizukawa, line 91: The impact seems to be low

影響は少なそう

sphinx/builders/changes.py
```diff
@@ -60,28 +62,30 @@ def get_outdated_docs(self):
...
             return
         logger.info(bold('writing summary file...'))
-        for type, docname, lineno, module, descname, content in \
-                self.env.versionchanges[version]:
-            if isinstance(descname, tuple):
-                descname = descname[0]
-            ttext = self.typemap[type]
-            context = content.replace('\n', ' ')
-            if descname and docname.startswith('c-api'):
+        for changeset in domain.get_changesets_for(version):
```
shimizukawa, line 73: It became easier to read.

読みやすくなりました

shimizukawa (approved):
This refactoring makes it easier to understand how to hold data.
However, unlike other domains, the changeset domain does not provide a namespace of reST notation. For this reason, I feel it is incompatible to call it a domain. 
Ignoring compatibility, it seems good to change the directive as follows.

- `versionadded` -> `version:added`
- `versionchanged` -> `version:changed`
- `deprecated` -> `version:deprecated`

In reality, such a change is not possible due to its large impact, but I think that it may be good to move the implementation code of these three directives into `sphinx/domains/changeset.py` as well.

tk0miya:
@shimizukawa Thank you for reviewing. I'm happy to agree with you.

I know namespace is one of main feature of Sphinx domains. But it is not a requirement. I think the most important responsibility of Sphinx domain is a collection of indices.
Indeed, the `version*` directives don't belong to namespace. But it works as collector of changesets. And changes builder also works as exporter of changesets. I know there are no namespace, cross reference and index page. But I believe this must be a Sphinx domain because it is a collection of changesets.

+1 for moving the implementation of directives under `sphinx.domains.changesets`. So I'll do it later.

tk0miya:
moved.

shimizukawa (approved):
again, LGTM!

tk0miya:
Thank you for reviewing!

Issue #5238, opened 2018-07-31
LaTeX: #5181 workaround an RTD issue causes unexpected consequences


jfbu:
Unfortunately the #5181 fix for RTD issue about doing only 2 pdflatex runs causes problem for documents lacking a general index (created from usage of ``:index:`` role or directive).

RTD does things like

```
pdflatex -interaction=nonstopmode /home/docs/checkouts/readthedocs.org/user_builds/sphinx/checkouts/master/doc/_build/latex/sphinx.tex
makeindex -s python.ist sphinx.idx
pdflatex -interaction=nonstopmode /home/docs/checkouts/readthedocs.org/user_builds/sphinx/checkouts
```

in place of using the Sphinx Makefile which uses Latexmk and does the automatic number of runs.

#5181 fixes that by anticipating index creation. This patch originated in #3742 which applied only to Sphinx's own documentation. But the overlooked problem at #5181 is that for documents not having indexed terms, that patch will leave the PDF with a dangling "Index" bookmark. Only  a third pdflatex run would get rid of it. This problem does not show when PDF build is done by `make -C  <buildir> all-pdf`, but only with only manual invocation of `pdflatex` as in the above RTD transcript related to Sphinx own documentation.

This means that #5181 will have to get reverted and a better solution found, if any, for 1.8 release.

- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.8+

Relates: 

- https://github.com/rtfd/readthedocs.org/issues/2857
- https://github.com/rtfd/readthedocs.org/issues/4454

jfbu:
Fixed at https://github.com/sphinx-doc/sphinx/pull/5239/commits/f0067f4a040a8b37afa3029c54806da9e32772e6

PR #5239, opened 2018-07-31
Revert "LaTeX: extend to all projects the #3742 fix for PDF builds at RTD"

jfbu:
This reverts commit 9b2aac68ead15067205470ec28c94af909c080c0.

### Relates
- #5238, #5181 
- https://github.com/rtfd/readthedocs.org/issues/2857
- https://github.com/rtfd/readthedocs.org/issues/4454


tk0miya (approved):

jfbu:
Thanks @tk0miya for quick review and sorry for oversight at #5181 that I did not think of projects without general index. Anyway, RTD policy of only two pdflatex runs and a hard-coded makeindex call causes problem also with xindy usage, so I hope they will use our Makefile and Latexmk, even though it basically means 3 pdflatex runs, but that's more or less unavoidable as latex is a multi-pass system.

Now merging.

PR #5241, opened 2018-07-31
LaTeX: make xindy usage opt-in also for xelatex/lualatex

### LINKED ISSUE #5240: LaTeX: ``xindy`` does not detect that index is empty, creates empty Index section in PDF
master branch has incorporated usage of xindy for UTF-8 index in PDF builds. Unfortunately, contrarily to `makindex`, `xindy` does not detect that `.idx` file is empty. With an empty `.idx` file (no usage of ``index`` role or directive in reST sources), `makeindex` produces an empty `.ind` file, but `xindy` produces an `.ind` file including the preamble/postamble the `\begin{sphinxtheindex}...\end{sphinxtheindex}` configured in `sphinx.xdy`. This causes an empty Index section at end of document.

Do `make clean latexpdf O="-D latex_use_xindy=1"` on any project which does not use ``index`` directive or role.

#### Error logs / results

The PDF contains an empty Index section at end of PDF.

#### Expected results

No empty Index section is created.

For pdflatex, this bug is not really an issue because `latex_use_xindy` is an opt-in option.

But for xelatex/lualatex it is on by default.

@tk0miya the easy way to fix this is to let default's value of `latex_use_xindy` always be `False`, even for `xelatex`, and `lualatex`.

If I find no solution at xindy's side, this is what I suggest to do.

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.8+
- TeXLive 2018

Relates:

#5134, #5192, #5212

jfbu:
This is option to fix #5240.

Either this or #5242 should be merged but *not* both.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,10 +4,9 @@ Release 1.8.0 (in development)
 Dependencies
 ------------
 
-* LaTeX: :confval:`latex_use_xindy`, if ``True`` (default for
-  ``xelatex/lualatex``), instructs ``make latexpdf`` to use :program:`xindy`
-  for general index.  Make sure your LaTeX distribution includes it.
-  (refs: #5134)
+* LaTeX: :confval:`latex_use_xindy`, if ``True`` instructs ``make latexpdf``
+  to use :program:`xindy` for general index.  Make sure your LaTeX
+  distribution includes it.  (refs: #5134)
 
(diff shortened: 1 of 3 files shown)


jfbu:
Closing this as #5242 was preferred.

PR #5242, opened 2018-07-31
Let Latexmk check if idx file is empty before xindy call

jfbu:
Fix: #5240

This is alternative to #5241. It works at my locale (MacOS X, bash shell, but I think `-s` is already Bourne shell `sh` syntax). This way, `xindy` remains default for `xelatex/lualatex` contrarily to #5241.

The point is that if FOO.idx file is empty (no `\index` in LaTeX document), then an empty FOO.ind file is created, and `xindy` is not used. This avoids empty Index section at end of PDF.

Either this or #5241 should be merged but *not* both.

### DIFF
--- a/sphinx/texinputs/latexmkrc_t
+++ b/sphinx/texinputs/latexmkrc_t
@@ -11,7 +11,7 @@ $pdflatex = 'xelatex ' . $ENV{'LATEXOPTS'} . ' %O %S';
 $lualatex = 'lualatex ' . $ENV{'LATEXOPTS'} . ' %O %S';
 $xelatex = 'xelatex --no-pdf ' . $ENV{'LATEXOPTS'} . ' %O %S';
 {% if xindy_use -%}
-$makeindex = 'xindy ' . $ENV{'XINDYOPTS'} . ' %O -o %D %S';
+$makeindex = '[ ! -s %S ] && : > %D || xindy ' . $ENV{'XINDYOPTS'} . ' %O -o %D %S';
 {% else -%}
 $makeindex = 'makeindex -s python.ist %O -o %D %S';
 {% endif -%}


sphinx/texinputs/latexmkrc_t
```diff
@@ -11,7 +11,7 @@ $pdflatex = 'xelatex ' . $ENV{'LATEXOPTS'} . ' %O %S';
 $lualatex = 'lualatex ' . $ENV{'LATEXOPTS'} . ' %O %S';
 $xelatex = 'xelatex --no-pdf ' . $ENV{'LATEXOPTS'} . ' %O %S';
 {% if xindy_use -%}
-$makeindex = 'xindy ' . $ENV{'XINDYOPTS'} . ' %O -o %D %S';
+$makeindex = '[ ! -s %S ] && : > %D || xindy ' . $ENV{'XINDYOPTS'} . ' %O -o %D %S';
```
tk0miya, line 14: How about this instead?
```
$makeindex = 'internal xindy ' . $ENV{'XINDYOPTS'} . ' %O -o %D %S';
sub xindy {
  my @args = @_;
  if (-s $args[-1]) {
    # create an empty .ind file if .idx is empty
    open(FH, ">" . $args[-2]);
    close(FH);
  } else {
    return system("xindy", @args);
  }
}
```

Unfortunately, I don't have enough time until this weekend. So this code is not tested (sorry!).
But I think it is better to implement this with perl script instead of shell script to support windows platform (in future).

jfbu: +1 for this idea! currently it does not work in my testing (case of empty file not detected) but I will work a bit on it. Unfortunately I have little time until September but I will give it a try now.

jfbu: @tk0miya I obtain success with 

```
$makeindex = 'internal xindy ' . $ENV{'XINDYOPTS'} . ' %O -o %D %S';
sub xindy {
  my @args = @_;
  if (-z $args[-1]) {
    # create an empty .ind file if .idx is empty
    print "Raw index file is empty, ignoring.\n";
    return system(": > $args[-2]");
  }
  else
  {
    return system("xindy", @args);
  }
}
```

but I needed some system call because Latexmk analyses exit status (there is a division at line 7020 in latexmk version ``25 May 2018. Version 4.56``). And I am no-ob in Perl. Also I changed to `-z` to `-s`. Surely there is better way.

PR #5245, opened 2018-08-01
LaTeX: streamline ``\DeclareUnicodeCharacter`` usage

jfbu:
Simple refactoring of LaTeX macros to reduce duplication in output .tex files and facilitate maintenance if list of extra Unicode characters should get extended.


tk0miya (approved):

jfbu:
Thanks for reviewing @tk0miya !

PR #5250, opened 2018-08-01
Add :rst:role:`math:numref` role to refer equations (Same as :rst:role:`eq`)

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add `math:numref` role as an alias of `eq` role.

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -1325,6 +1325,25 @@ These roles are provided to refer to the described objects:
 .. rst:role:: rst:dir
               rst:role
 
+.. _math-domain:
+
+The Math Domain
+---------------
+
+The math domain (name **math**) provides the following roles::
+
+.. rst:role:: math:numref
+
+   Role for cross-referencing equations defined by :rst:dir:`math` directive
+   via their label.  Example::
+
+      .. math:: e^{i\pi} + 1 = 0
+         :label: euler
+
+      Euler's identity, equation :math:numref:`euler`, was elected one of the
+      most beautiful mathematical formulas.
+
+   .. versionadded:: 1.8
 
 More domains
 ------------
--- a/CHANGES
+++ b/CHANGES
@@ -186,6 +186,7 @@ Features added
 * Apply :confval:`trim_doctest_flags` to all builders (cf. text, manpages)
 * #5140: linkcheck: Add better Accept header to HTTP client
 * #4614: sphinx-build: Add ``--keep-going`` option to show all warnings
+* Add :rst:role:`math:numref` role to refer equations (Same as :rst:role:`eq`)
 
 Bugs fixed
 ----------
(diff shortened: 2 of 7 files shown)


doc/usage/restructuredtext/domains.rst
```diff
@@ -1325,6 +1325,25 @@ These roles are provided to refer to the described objects:
...
               rst:role
 
+.. _math-domain:
+
+The Math Domain
+---------------
+
+The math domain (name **math**) provides the following roles::
+
+.. rst:role:: rst:numref
```
jfbu: ``math:numref`` ?

doc/usage/restructuredtext/domains.rst
```diff
@@ -1325,6 +1325,25 @@ These roles are provided to refer to the described objects:
...
+
+   Role for cross-referencing equations defined by :rst:dir:`math` directive
+   via their label.  Example::
+
+      .. math:: e^{i\pi} + 1 = 0
+         :label: euler
+
+      Euler's identity, equation :eq:`euler`, was elected one of the most
```
jfbu: did you mean to use ``:math:numref:`euler` `` here ?

jfbu (commented):
LGTM with nits

jfbu:
are you planning to deprecate `:eq:` in future?

doc/usage/restructuredtext/domains.rst
tk0miya: oh... thanks. fixed

doc/usage/restructuredtext/domains.rst
tk0miya: right. done

tk0miya:
No, I'd like to keep `:eq:` role for a while (or forever?).
I think there are many document contain the old named role notations. And I believe documents have very longer life than program.

jfbu:
> I think there are many document contain the old named role notations. And I believe documents have very longer life than program.

agreed :) I am annoyed that LaTeX babel-french for example will one day remove legacy "frenchb" in favour of only "french" and rename "\frenchbsetup" to "\frenchsetup". This means old sources from ten years ago may need manual edit to compile... bad. (my example here is for direct use of LaTeX, not via Sphinx...)

**edit** another better example is the brutal suppression by KOMA-script classes of support of `\bf`, `\rm` etc... about one year ago after more than 20 years of support. This is really silly decision, because the macros are still defined in LaTeX standard classes. It is impossible to understand why one day the authors decide to stop supporting them, breaking many documents which use LaTeX packages which refer to it and are perhaps now unmaintained.

jfbu (approved):

tk0miya:
I can't promise all directives and roles will keep supported in 10+ yrs later. But we have to care to keep them as possible.
On the other hand, sustainability is also important for all of maintainers. So we have to discuss how to keep readability and maintainability of our code.

Anyway,  I think keeping `:eq:` role is not hard at all to us :-)

tk0miya:
Merged. Thank you for reviewing.

PR #5261, opened 2018-08-04
Fix #5246: Add :confval:`singlehtml_sidebars` to configure sidebars for singlehtml

### LINKED ISSUE #5246: Configure sidebars for singlehtml
It's common to configure one set of sidebars for the index and another set for all other pages. For example, the index page will show project links in the sidebar and the full table of contents in the body, while other pages will show the local table of contents and navigation.

```python
html_sidebars = {
    'index': [
        'project.html',
        'versions.html',
        'searchbox.html',
    ],
    '**': [
        'localtoc.html',
        'relations.html',
        'versions.html',
        'searchbox.html',
    ]
}
```

When using the singlehtml builder, the page name is always "index". It would be desirable to show the table of contents in the sidebar for this builder, since due to #2589 the table of contents is replaced by the pages and won't show up in the body. But since the page name is "index", it uses the index sidebars configured for the regular html builder.

A solution to this would be to add a `singlehtml_sidebars = []` configuration, and fall back to the `html_sidebars` configuration if it's not present. It would also be good to be able to configure the default in `theme.conf` like how `html_sidebars` works.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #5246

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1333,6 +1333,16 @@ that use Sphinx's HTMLWriter class.
 
    .. versionadded:: 1.6
 
+Options for Single HTML output
+-------------------------------
+
+.. confval:: singlehtml_sidebars
+
+   Custom sidebar templates, must be a dictionary that maps document names to
+   template names.  And it only allows a key named `'index'`.  All other keys
+   are ignored.  For more information, refer to :confval:`html_sidebars`.  By
+   default, it is same as :confval:`html_sidebars`.
+
 
 .. _htmlhelp-options:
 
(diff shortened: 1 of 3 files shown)


doc/usage/configuration.rst
```diff
@@ -1333,6 +1333,15 @@ that use Sphinx's HTMLWriter class.
 
    .. versionadded:: 1.6
 
+Options for Single HTML output
+-------------------------------
+
+.. confval:: singlehtml_sidebars
+
+   Custom sidebar templates, must be a dictionary that maps document names to
```
davidism, line 1341: This should mention that the only valid key is `index` since that's the only page being built.

tk0miya: done

tk0miya:
Merged. @davidism  Thank you for reviewing!

PR #5262, opened 2018-08-04
Use 3.8-dev for testing instead of nightly

tk0miya:
At present, it seems nightly build points 3.7.0a4+.  So use 3.8-dev
for our testing instead.  This change will be reverted after Travis
fixes nightly build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,8 @@ matrix:
       sudo: true
     - python: 'nightly'
       env: TOXENV=py38
+      dist: xenial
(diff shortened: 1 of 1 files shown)


tk0miya:
Unfortunately, Travis CI is in trouble. So I'll try this after recovered.
https://www.traviscistatus.com/incidents/118bjpkgkg00

tk0miya:
I found  `python: nightly` in `dist: xenial` points 3.8+. So this simply updates its environment finally.

PR #5265, opened 2018-08-04
WIP: 5258 fix search highlight svg ff

### LINKED ISSUE #5258: Search highlighting in SVGs does not work in Firefox
Subject: Search highlighting in SVGs does not work in Firefox

### Problem
Adding JS tests yields the first bug results.
I found that search keywords highlighting in SVGs does not work in Firefox.

1. Add the following markup to a Sphinx page:
```
.. raw:: html

    <span id="svg-highlight-test">
        <svg xmlns="http://www.w3.org/2000/svg" height="50" width="500">
            <text x="0" y="15">
              This is the шеллы and Gänsefüßchen test!
            </text>
        </svg>
    </span>
```
2. Run the HTML build.

3. Open the documentation in Firefox.

4. Search for "шеллы".

#### Error logs / results
Firefox throws:
```
NS_ERROR_NOT_IMPLEMENTED: 
```
in line 90 of ``doctools.js``:

```JavaScript
var bbox = span.getBBox();
```
This is a known Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=612118 It works in Chrome. I'll look for a workaround.

#### Expected results
The browser should highlight the search terms in the SVG.

Copy & paste:
```
.. raw:: html

    <span id="svg-highlight-test">
        <svg xmlns="http://www.w3.org/2000/svg" height="50" width="500">
            <text x="0" y="15">
              This is the шеллы and Gänsefüßchen test!
            </text>
        </svg>
    </span>
```

### Environment info
- Browser: Firefox 61

TimKam:
Subject: Fix: search keywords highlighting in SVGs does not work in Firefox.
### Feature or Bugfix
- Bugfix

### Purpose
Use `node.parentElement.getBBox()` instead of `span.getBBox()`:
    
* Same highlighting behavior in Chrome, Opera
* Avoid issue with getBBox() in FF, see: https://bugzilla.mozilla.org/show_bug.cgi?id=612118
* Microsoft browsers to be tested (hence still WIP)

There is another, smaller issue with search highlighting in SVG: the highlighting function does not highlight single works, but encompassing ``text`` blocks. But I think it's not necessary to address it in this PR (rather in a follow-up ticket).

### Detail
- Fixes #5258


tk0miya:
Could you rebase or merge into HEAD of master please.
The tests for `python: nightly` has been broken since yesterday. And I just fixed it in #5262.

I know this is still in WIP. So I'll take a look this after all your work.

TimKam:
After testing on Windows machine, I found out that SVG highlighting did not work before with Internet Explorer, and still does not work now. However, considering that IE is legacy and presumably rarely used by readers of Sphinx-generated documentation (especially of docs that include SVGs), I do not think this is worth fixing.

@tk0miya can I merge?

tk0miya (approved):
+1

PR #5275, opened 2018-08-07
refactor: Move settings for pLaTeX to ADDITIONAL_SETTINGS

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To be latex writer simple

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -177,6 +177,9 @@ ADDITIONAL_SETTINGS = {
     },
     'platex': {
         'latex_engine': 'platex',
+        'babel':        '',
+        'classoptions': ',dvipdfmx',
+        'fncychap':     '',
         'geometry':     '\\usepackage[dvipdfm]{geometry}',
     },
 }  # type: Dict[unicode, Dict[unicode, unicode]]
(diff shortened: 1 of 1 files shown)


sphinx/writers/latex.py
```diff
@@ -570,8 +573,8 @@ def __init__(self, document, builder):
             self.elements['logo'] = '\\sphinxincludegraphics{%s}\\par' % \
                                     path.basename(builder.config.latex_logo)
 
-        if builder.config.language \
-           and 'fncychap' not in builder.config.latex_elements:
+        if (builder.config.language and builder.config.language != 'ja' and
+                'fncychap' not in builder.config.latex_elements):
             # use Sonny style if any language specified
             self.elements['fncychap'] = ('\\usepackage[Sonny]{fncychap}\n'
```
tk0miya, line 579: This means there are 3 types fncychap settings:
* no language: `Bjarne`
* language = 'ja': None
* else: `Sonny`

I understand about 'ja' case. But I don't understand why switching it on other cases. To be more simple our code, it would be better to make `Sonny` as a default setting.

sphinx/writers/latex.py
```diff
@@ -586,13 +589,10 @@ def __init__(self, document, builder):
             logger.warning(__('no Babel option known for language %r'),
                            builder.config.language)
 
-        # simply use babel.get_language() always, as get_language() returns
-        # 'english' even if language is invalid or empty
-        self.elements['classoptions'] += ',' + self.babel.get_language()
```
tk0miya, line 591: I'm moving this to both `babel` and `polyglossia` cases.
After this refactoring, all of `babel`, `polyglossia` and `multilingual` key become empty string for `platex` engine. So this line will not be called for the engine.

PR #5276, opened 2018-08-07
Add LaTeXBuilder.get_doc_context(): Build context object for each document

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Now `LaTeXWriter.__init__()` becomes too large.
- To lessen the responsibility of the module, this starts to move the initialization of context object to `LaTeXBuilder.get_doc_context()`.
- I don't know it is best place. But I think it would be more approriate than writer.
- I think this also helps our tests.

@jfbu @shimizukawa please let me know your opinion.


sphinx/builders/latex/__init__.py
```diff
@@ -220,6 +225,37 @@ def write(self, *ignored):
...
+        # Apply user settings to context
+        context.update(self.config.latex_elements)
+        context['release'] = self.config.release
+
+        # Override settings by latex_documents
+        context['title'] = title
+        context['author'] = author
+        context['use_xindy'] = self.config.latex_use_xindy
+
+        # for compatibilities
```
tk0miya: Now we can translate message on templates directly. So I move them to templates. It means users can't modify the message by `latex_elements`. I think it is not expected feature.

Note: I have to add them to CHANGES entry.

sphinx/templates/latex/latex.tex_t
```diff
@@ -47,11 +47,15 @@
...
+\renewcommand{\releasename}{<%= releasename or _('Release') %>}
+<%- else %>
+\renewcommand{\releasename}{}
+<%- endif %>
 <%= makeindex %>
 <%= body %>
 <%= atendofbody %>
 <%= indices %>
-\renewcommand{\indexname}{<%= indexname %>}
+\renewcommand{\indexname}{<%= _('Index') %>}
```
tk0miya: Should this be defined in document environment?

sphinx/builders/latex/__init__.py
```diff
@@ -220,6 +225,37 @@ def write(self, *ignored):
             docwriter.write(doctree, destination)
             logger.info("done")
 
+    def get_doc_context(self, title, author, docclass):
```
tk0miya: It might be better to mark as unstable because this API is under construction.

tk0miya:
Oh, testing has been failed. I'll take a look tomorrow...

sphinx/builders/latex/__init__.py
```diff
@@ -167,6 +172,31 @@ def init_document_data(self):
                 docname = docname[:-5]
             self.titles.append((docname, entry[2]))
 
+    def init_context(self):
+        # type: () -> None
+        self.context = DEFAULT_SETTINGS.copy()
```
shimizukawa, line 177: The method of preparing this context in the builder is the same as what we do in the `builders/ html.py` code, so I didn't feel uncomfortable with this change

shimizukawa (approved):
LGTM!

sphinx/builders/latex/__init__.py
```diff
@@ -126,11 +128,14 @@ class LaTeXBuilder(Builder):
 
     def init(self):
         # type: () -> None
+        self.context = {}           # type: Dict[unicode, Any]
```
tk0miya, line 131: Note: In this PR, the responsibility of context object is unbalanced.

* Builder: creating context object
* Writer: updating context object and generate .tex file using context

I think it should be refactored this in future. It would be better we should move the responsibility of generating .tex file to builder module (like HTML builder does).

sphinx/templates/latex/latex.tex_t
```diff
@@ -47,11 +47,15 @@
 \release{<%= release %>}
 \author{<%= author %>}
 \newcommand{\sphinxlogo}{<%= logo %>}
-\renewcommand{\releasename}{<%= releasename %>}
+<%- if releasename or release %>
+\renewcommand{\releasename}{<%= releasename or _('Release') %>}
```
tk0miya: I noticed this should be escaped.

tk0miya:
ping @jfbu. any comment for this?

sphinx/templates/latex/latex.tex_t
jfbu: not sure what you mean, which part should be defined there? It can't be moved before `<%= indices %>` because they have their own `\renewcommand\indexname`.  In writers/latex.py:

https://github.com/sphinx-doc/sphinx/blob/c659a562915d9771e3c246a3cade2ea9079ca10d/sphinx/writers/latex.py#L862-L863

(Python Module Index e.g.). Or perhaps you mean `sphinx.writers.generate_indices()` should incorporate this line as last line of `<%= indices =>`? But then if user uses custom template with general index located before domain indices, there is problem.

jfbu (approved):
LGTM

sphinx/templates/latex/latex.tex_t
tk0miya: I thought preamble is better. But it's my wrong. Thank you for explanation.

sphinx/templates/latex/latex.tex_t
tk0miya: done

tk0miya:
Merged. Thank you!

PR #5277, opened 2018-08-07
Fix #4075: autodoc: autodoc_default_flags allows dict style configuration

### LINKED ISSUE #4075: limitations to autodoc_default_flags in conf.py

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #4075, #4076 and #5215

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -191,6 +191,8 @@ Features added
 * #4614: sphinx-build: Add ``--keep-going`` option to show all warnings
 * Add :rst:role:`math:numref` role to refer equations (Same as :rst:role:`eq`)
 * quickstart: epub builder is enabled by default
+* #4075: autodoc: :confval:`autodoc_default_flags` allows dict style
+  configuration
 
 Bugs fixed
 ----------
--- a/doc/usage/extensions/autodoc.rst
+++ b/doc/usage/extensions/autodoc.rst
@@ -353,7 +353,18 @@ There are also new config values that you can set:
 
    the directive will be interpreted as if only ``:members:`` was given.
 
+   It also allows a dict of flags and values like this::
+
+      autodoc_default_flags = {
+          'members': 'var1, var2',  # Give a parameter as a value of dict
+          'undoc-members': None,    # Set "None" if value is not needed
+      }
+
    .. versionadded:: 1.0
+
+   .. versionchanged:: 1.8
+
+      It takes dict style configuration also.
 
 .. confval:: autodoc_docstring_signature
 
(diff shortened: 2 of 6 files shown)


tk0miya:
I withdraw this because #5215 is improved now.

PR #5279, opened 2018-08-09
#5278: re-format links

### LINKED ISSUE #5278: Refactor links on Examples page
As discussed in #5259: I think it would be nice to use properly formatted links instead of raw URLs at http://www.sphinx-doc.org/en/master/examples.html.
If there are no objections, I re-format all links there, i.e.:

```
Alabaster: https://alabaster.readthedocs.io/
```
->

```
`Alabaster`__

__ https://alabaster.readthedocs.io/
```
and so on.

linnekback:
Subject: re-format links in examples

### Feature or Bugfix
- documentation

Closes #5278

### DIFF
--- a/EXAMPLES
+++ b/EXAMPLES
@@ -293,7 +293,6 @@ Documentation using sphinx_rtd_theme
 * `Certbot`__
 * `Chainer`__ (customized)
 * `CherryPy`__
-* `Chainer`__
 * `CodeIgniter`__
 * `Conda`__
 * `Corda`__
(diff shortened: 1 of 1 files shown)


tk0miya:
I found it is a bit boring to use anonymous hyperlink target notation for this list. On editing the list, we have to edit a pair of the hyperlink with keeping the order.
I feel inline hyperlink is better for this case:
```
* `Alabaster <https://alabaster.readthedocs.io/>`_
```

What do you think?

TimKam:
I will keep this mind as a future todo if that's ok :-)

PR #5282, opened 2018-08-10
quickstart: set conf.py pygments_style value to None

schettino72:
### Feature or Bugfix

- Bugfix

### Purpose

Currently quickstart sets a value for `pygments_style`,
this is not desirable because it overrides the theme's default `pygments_style`.
It should be `None` so theme`s value is used by default.

### DIFF
--- a/sphinx/templates/quickstart/conf.py_t
+++ b/sphinx/templates/quickstart/conf.py_t
@@ -78,7 +78,7 @@ language = {{ language | repr }}
 exclude_patterns = [{{ exclude_patterns }}]
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+pygments_style = None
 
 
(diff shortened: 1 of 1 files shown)


sphinx/templates/quickstart/conf.py_t
```diff
@@ -78,7 +78,7 @@ language = {{ language | repr }}
 exclude_patterns = [{{ exclude_patterns }}]
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+pygments_style = None
```
tk0miya, line 81: Note: the original setting was born at 7d9721dd85 (v0.2). So this setting is a part of Sphinx itself even if it's a bug.

tk0miya:
I just understand this is a bug of Sphinx. So I'll merge this into 1.8.0 release.

tk0miya:
Merged. Thank you for contribution!

PR #5280, opened 2018-08-09
change format_args to prefer annotations from typing.get_type_hints

terencehonles:
Subject: change format_args to prefer annotations from typing.get_type_hints

### Feature or Bugfix
- Bugfix

### Purpose

The current code looks for the pattern:

```python
class Node:
    def parent(self) -> 'Node': pass
```

This change allows more complex forward references:

```python
class Node:
    def children(self) -> Set['Node']: pass
```

PR #5286, opened 2018-08-10
update .travis.yml to not break when building against branch 1.7

terencehonles:
Subject: Fixes travis build against #5280

### Feature or Bugfix
- Bugfix

### Relates
- #5280

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,12 +49,12 @@ matrix:
         - sh -e /etc/init.d/xvfb start
 
 install:
-  - if [ $IS_PYTHON = true ]; then pip install -U tox codecov; fi
-  - if [ $IS_PYTHON = false ]; then npm install; fi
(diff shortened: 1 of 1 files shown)


tk0miya:
On 1.7 branch, `IS_PYTHON` flag does not exist. So I think the error for #5280 was a problem of Travis CI.
So I'm closing this now.
Thanks,

terencehonles:
@tk0miya the error was `npm install` didn't find the `package.json` on the 1.7 branch, so I made it check to see if it exists and the npm test will skip if it can't run

tk0miya:
Yes. But there are no `IS_PYTHON` flag in travis.yml on the 1.7 branch. So the main reason of the error is that Travis invoked `npm install` for 1.7 branch.
https://github.com/sphinx-doc/sphinx/blob/1.7/.travis.yml

Issue #5273, opened 2018-08-07
Feature request: Skip doctest conditionally


akaihola:
Subject: Unit tests can be skipped conditionally using the `@unittest.skipIf()` decorator. A similar feature would be useful for doctests when run by Sphinx.

### Example use cases
- different tests for specific versions of dependencies
- skip tests in absence of optional dependencies
- skip tests in absence of a network or VPN
- skip tests in absence of some hardware

The feature was also mentioned on Stack Overflow in the [Sphinx doctest: conditional test skipping](https://stackoverflow.com/questions/51292196/) question.

### Proposed syntax
```
.. doctest::
   :skipif: <condition>

   >>> test()
   <output>

.. testcode::
   :skipif: <condition>

   test()

.. testoutput::
   :skipif: <condition>

   <output>
```

### Global setup
The code in the `doctest_global_setup` and `doctest_global_cleanup` configuration settings should be run before and after the condition. This would enable users to store complex conditions in a global variable and avoid repetition in doctest definitions.

### Expected results
When a `<condition>` evaluates to a true value, the test or the test output assertion should be skipped.

akaihola:
In my patch above, I simply return an empty list from the `run()` method of the doctest directive classes in case the expression in the new `:skipif:` option evaluates to a true value.

I noticed that in the `:pyversion:` directive, a `SKIP` doctest flag is injected instead. However, this method only seems to be usable for the `doctest::` directive but not for `testcode::` and `testoutput::`. For this reason, I adopted the empty list method for all of the directives. As a side effect, conditional skipping can also be used for `testsetup::` and `testcleanup::`.

If the proposed feature and my approach make sense, I'll proceed to writing tests and documentation next.

tk0miya:
+0; I don't know doctest module so much. So I don't have any opinion. But it seems worthy.

ping @shimizukawa

shimizukawa:
+0. I am anxious about whether it will be too many options.
However, it seems good to have an option to exclude doctest under specific conditions from testing.

tk0miya:
Okay, let's go forward.
@akaihola could you send us a PR please? It seems your patch is enough good. I will merge it soon if 
 you'll add testcode for it.

Thanks,

tk0miya:
Note:
We'll call feature freeze for 1.8 in this weekend. It will be released as 1.8 if the patch will be merged until then.

akaihola:
[...]

PR #5307, opened 2018-08-16
Feature: skip doctests conditionally

akaihola:
Subject: A `:skipif:` option for doctest directives

### Feature or Bugfix

- Feature

### Purpose
- Unit tests can be skipped conditionally using the `@unittest.skipIf()` decorator. A similar feature would be useful for doctests when run by Sphinx.

### Detail
Example use cases:
- different tests for specific versions of dependencies
- skip tests in absence of optional dependencies
- skip tests in absence of a network or VPN
- skip tests in absence of some hardware

### Relates
- #5273
- The feature was also mentioned on Stack Overflow in the [Sphinx doctest: conditional test skipping](https://stackoverflow.com/questions/51292196/) question.


tk0miya (approved):
Perfect! LGTM!

tk0miya:
Merged.
Thank you for great contribution!

akaihola:
Thank you @tk0miya, I'm really happy to get this feature in. And the experience of contributing to Sphinx was one of the best I've had with any Open Source project. Great work guys!

timokau:
Is it somehow possible to add the `:skipif:` option to doctest blocks (i.e. without explicit `.. doctest::` directive)?

Example:

```
>>> print(42)
42
```

How can I skip this test conditionally? The  only option I found is adding `.. doctest::`, but that makes it necessary to change indentation as it apparently has to have one indentation level less than the doctest block.

akaihola:
@timokau, hi, I'm not 100% sure, but I would guess that it's not possible to apply `:skipif:` to such doctest blocks.

Note that the code in a `.. doctest::` block can be indented by three or more spaces. So a four-space indent works just as well!

timokau:
Okay, thanks for the reply. I'll just live with the indentation "issue" then. For what its worth, this illustrates the issue I'm talking about:

The indentation issue is:

```
def fn():
"""
This does something.

Examples:

	Here it prints something:

.. doctest::
   :skipif: True

	>>> fn()
	"something"

"""
```

wojdyr:
Thanks for this feature!

I just realized that `:skipif:` is evaluated also when building documentation (`make html`) not just during `make doctest`. So if a certain module is missing on ReadTheDocs then corresponding doctest blocks are not included in documentation. This is easy to workaround, but I wonder if it's intentional?

akaihola:
@wojdyr, that's a really good point. Skipping on `make html` was not intentional. I remember it kind of occurring to me at some point, but apparently I forgot about it. Feel free to create another issue about this, it would very much make sense to fix it.

wojdyr:
@akaihola I've opened #6068 .
Thanks again for `:skipif:`, it solved the problem I had with doctest blocks.

PR #5315, opened 2018-08-18
Add autodoc_default_options

tk0miya:
### Feature or Bugfix
- Feature

### Purpose

- An alternative of #5215.
- This adds `autodoc_default_options` as a successor of `autodoc_default_flags` which only accepts dict-style options.
- This also reverts the change in #5215 that `autodoc_default_flags` accepts dict-style options. This is same as it does in v1.7.
- refs: https://github.com/sphinx-doc/sphinx/pull/5215#issuecomment-412891044

@shimizukawa @LewisHaley What do you think for this? At first, I'd like to add an alias of `autodoc_default_flags`. But I stopped to do that because it will be deprecated. I think it is simple for users to keep it as is (not supporting dict-style options) and to add new configuration instead.

Thanks,

### DIFF
--- a/doc/usage/extensions/autodoc.rst
+++ b/doc/usage/extensions/autodoc.rst
@@ -114,8 +114,17 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
         .. autoclass:: Noodle
            :members: eat, slurp
 
-   * If you want to make the ``members`` option (or other flag options described
-     below) the default, see :confval:`autodoc_default_flags`.
+   * If you want to make the ``members`` option (or other options described
+     below) the default, see :confval:`autodoc_default_options`.
+
+     .. tip::
+
+        You can use a negated form, :samp:`'no-{flag}'`, as an option of
+        autodoc directive, to disable it temporarily.  For example::
+
+           .. automodule:: foo
+              :no-undoc-members:
+
 
    * Members without docstrings will be left out, unless you give the
      ``undoc-members`` flag option::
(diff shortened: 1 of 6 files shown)


doc/usage/extensions/autodoc.rst
```diff
@@ -344,35 +353,34 @@ There are also new config values that you can set:
...
+           'members': 'var1, var2',
+           'member-order': 'bysource',
+           'special-members': '__init__',
+           'undoc-members': None,
+           'exclude-members': '__weakref__'
+       }
 
-   .. versionadded:: 1.0
+   Setting ``None`` is equivalent to giving the option name in the list format
+   (i.e. it means "yes/true/on").
```
shimizukawa, line 377: "`None` means yes/true/on` get confusing to me. However I have no good good alternative.

Finally I've suspended my opinion because this is not a topic of this PR.

shimizukawa (approved):
LGTM!

doc/usage/extensions/autodoc.rst
tk0miya: I agree. `None` is not good to represent flags. It is internal representation...
But I also don't have good idea...

tk0miya:
Now I'm merging this because today is feature freeze of 1.8.
Please let me know if you have comment even if after merged.

Thanks,

doc/usage/extensions/autodoc.rst
```diff
@@ -344,35 +353,34 @@ There are also new config values that you can set:
...
 
-     autodoc_default_flags = {
-        'members': 'var1, var2',
-        'member-order': 'bysource',
-        'special-members': '__init__',
-        'undoc-members': None,
-        'exclude-members': '__weakref__'
-    }
+   The default options for autodoc directives.  They are applied to all autodoc
+   directives automatically.  It must be a dictionally which maps option names
```
LewisHaley, line 365: Sorry I didn't see this earlier but there's a typo: `dictionally` -> `dictionary`

tk0miya: thanks!

Issue #5340, opened 2018-08-24
Incorrect signatures in sphinx docs.


anntzer:
At http://www.sphinx-doc.org/en/master/extdev/logging.html#sphinx.util.logging.SphinxLoggerAdapter, the signatures of the logging functions are given as `error(level, msg, *args, **kwargs)` (and so on), but the functions (other than `log` itself) don't actually take `level` as first argument.
Also, it's a bit curious that half of the kwargs are documented under `warning` and the other half under `debug`.
Finally, note that there is a contradiction between the claims "By default, warning level logs are colored as "darkred". The others are not colored." and "By default, debug level logs are colored as "darkgray", and debug2 level ones are "lightgray". The others are not colored." :)

tk0miya:
Thank you for letting us know. I'll fix it later.

About color option,  first sentence describes about `error`, `critical` and `warning`. Another one describes about `log`, `info`, `verbose` and `debug`. So I think there are no contradiction.

tk0miya:
Note: I found the description of default color is not correct.

tk0miya:
Fixed by #5343.
Thank you for reporting.

PR #5343, opened 2018-08-24
docs: Fix wrong descriptions for logging APIs

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5340


anntzer:
Just move `log` out and group the other six together?

tk0miya:
The options are different for each group. So we can't merge them.
See `type`, `subtype` and `nonl`.

PR #5345, opened 2018-08-25
autodoc: Fix typing.get_type_hints() raises AttributeError for partial objects

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- old `typing.get_type_hints()` raises AttributeError for partial objects
- As a result, our `sphinx.util.inspect:Signature` class emits warning for them forcedly. But nobody stop it.
- This prevents such warnings.

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -360,8 +360,13 @@ class Signature(object):
         try:
             self.annotations = typing.get_type_hints(subject)  # type: ignore
         except Exception as exc:
-            logger.warning('Invalid type annotation found on %r. Ingored: %r', subject, exc)
-            self.annotations = {}
+            if (3, 5, 0) <= sys.version_info < (3, 5, 3) and isinstance(exc, AttributeError):
+                # python 3.5.2 raises ValueError for partial objects.
+                self.annotations = {}
+            else:
+                logger.warning('Invalid type annotation found on %r. Ingored: %r',
+                               subject, exc)
+                self.annotations = {}
 
         if bound_method:
             # client gives a hint that the subject is a bound method


sphinx/util/inspect.py
```diff
@@ -360,8 +360,13 @@ def __init__(self, subject, bound_method=False, has_retval=True):
         try:
             self.annotations = typing.get_type_hints(subject)  # type: ignore
         except Exception as exc:
-            logger.warning('Invalid type annotation found on %r. Ingored: %r', subject, exc)
-            self.annotations = {}
+            if (3, 5, 0) <= sys.version_info < (3, 5, 3) and isinstance(exc, AttributeError):
+                # python 3.5.2 raises ValueError for partial objects.
+                self.annotations = {}
+            else:
+                logger.warning('Invalid type annotation found on %r. Ingored: %r',
```
pganssle, line 367: s/Ingored/Ignored

pganssle:
This warning does not seem to provide a huge amount of information here. I'm not using type annotations at all and it's started causing [errors with the `dateutil` docs build](https://travis-ci.org/dateutil/dateutil/jobs/428228219#L498). I have no idea where to look to figure out what to do there...

pganssle:
I think my issue is something about the fact that `dateutil.rrule.rrulestr` is implemented as a callable object, but I'm not sure why this is all of a sudden throwing warnings when I'm not annotating anything...

tk0miya:
strange. This class is supposed to be used only to classes, functions and methods. But it seems an object was given to the class in your case.
Could you file a new issue please? I'll take a look later.

Thanks,

sphinx/util/inspect.py
tk0miya: Thanks, fixed #5430

PR #5346, opened 2018-08-25
[Proposal] Add a helper class for Node.traverse()

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- In sphinx, there are many small inline functions to pick up nodes conditionally.
- I'd like to add a helper class named `NodeMatcher`.
- It can search nodes simply.

### DIFF
--- a/sphinx/util/nodes.py
+++ b/sphinx/util/nodes.py
@@ -11,6 +11,7 @@
 from __future__ import absolute_import
 
 import re
+from typing import Any
 
 from docutils import nodes
 from six import text_type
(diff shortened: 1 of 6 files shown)


sphinx/util/nodes.py
```diff
@@ -33,6 +34,49 @@
...
+        # type: (nodes.Node, Any) -> None
+        self.classes = classes
+        self.attrs = attrs
+
+    def match(self, node):
+        # type: (nodes.Node) -> bool
+        try:
+            if self.classes and not isinstance(node, self.classes):
+                return False
+            else:
```
cocoatomo: This if-else statement might be in early-return form as following:

```
if self.classes and not isinstance(node, self.classes):
    return False

for key, value in self.attrs.items():
    if key not in node:
...
```

Early-return form reduce depth of indentation and make it easier to read source codes.

sphinx/util/nodes.py
```diff
@@ -33,6 +34,49 @@
 caption_ref_re = explicit_title_re  # b/w compat alias
 
 
+class NodeMatcher(object):
+    """A helper class for Node.traverse().
```
cocoatomo, line 38: Descriptions about `Any` value of some attribute are necessary.

cocoatomo (changes requested):
The topic of early-return is about code style, so it is an approval.
But the topic of documentation is mandatory.

sphinx/util/nodes.py
tk0miya: Thanks, I think so too.

sphinx/util/nodes.py
tk0miya: Absolutely!

shimizukawa (approved):
LGTM!!

tk0miya:
Merged. Thank you for reviewing.

PR #5349, opened 2018-08-25
Fix #5348: download reference to remote file is not displayed

### LINKED ISSUE #5348: download reference to remote file is displayed
### Problem
- download reference to remote file is warned

```
:download:`http://www.sphinx-doc.org/en/master/_static/sphinxheader.png`
```

#### Error logs / results
```
/Users/tkomiya/work/tmp/doc/test/index.rst:9: WARNING: download file not readable: /Users/tkomiya/work/tmp/doc/test/http:/www.sphinx-doc.org/en/master/_static/sphinxheader.png
```
And download link is disabled in generated HTML.

#### Expected results
- No warnings
- Enable hyperlink to remote file

### Environment info
- OS: Mac
- Python version: 3.7.0
- Sphinx version: 1.7.7

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5348

### DIFF
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -564,10 +564,20 @@ class HTMLTranslator(BaseTranslator):
 
     def visit_download_reference(self, node):
         # type: (nodes.Node) -> None
-        if self.builder.download_support and node.hasattr('filename'):
-            self.body.append(
-                '<a class="reference download internal" href="%s" download="">' %
-                posixpath.join(self.builder.dlpath, node['filename']))
+        atts = {'class': 'reference download',
+                'download': ''}
+
+        if not self.builder.download_support:
+            self.context.append('')
+        elif 'refuri' in node:
+            atts['class'] += ' external'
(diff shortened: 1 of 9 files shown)


sphinx/writers/html.py
```diff
@@ -564,17 +564,20 @@ def depart_compact_paragraph(self, node):
...
-            self.body.append(
-                '<a class="reference download internal" href="%s" download="">' %
-                posixpath.join(self.builder.dlpath, node['filename']))
-            self.context.append('</a>')
+            atts['class'] += ' internal'
+            atts['href'] = posixpath.join(self.builder.dlpath, node['filename'])
         else:
-            self.context.append('')
+            atts['class'] += ' external'
+            atts['href'] = node['reftarget']
```
tk0miya: This `else` clause is not correct. This block is also used for non-download-supported builders.

tk0miya: Note: I have to modify html5 writer also.

Issue #4034, opened 2017-08-31
Incorrect links with :download:, duplicate names, and parallel builds


bluekeyes:
### Problem

When building documentation in parallel (with the `-j` flag), using the `:download:` role to link to files with duplicate names can result in incorrect links in the output.

I believe this is because the [`DownloadFileCollector`](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/environment/collectors/asset.py#L137) rewrites link targets on name conflict, but [`FilenameUniqDict#merge_other`](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/__init__.py#L153) (which is used to implement the `env-merge-info` hook) can change the filenames again without updating the link nodes.

#### Procedure to reproduce the problem

The shell script below generates a minimal project with the correct properties, but any project with links to files with the same name from multiple documents should reproduce the problem. Run `sphinx-build` with the `-j` flag:

```
sphinx-build -b html -j4 "." "_build" 
```

#### Expected results

In the example project, the link on each of the 20 pages should lead to a matching file, e.g. "File 1" should lead to a file containing the text `File 1`, "File 2" should lead to a file containing the text `File 2`, etc.

Instead, the linked files are mixed up and the content does not match the origin page.

### Reproducible project

Run the following shell script in an empty directory to generate a test project:

```bash
#!/bin/bash
set -e

cat > conf.py <<EOF
exclude_patterns = ['_build']
master_doc = 'index'
project = u'sphinx-download-bug'
source_suffix = '.rst'
EOF

cat > index.rst <<EOF
Welcome to sphinx-download-bug's documentation!
===============================================

.. toctree::
EOF

for i in {1..20}; do
    mkdir -p files/$i

    echo "File $i" > files/$i/file.txt
    cat > $i.rst <<EOF
File $i
#######

Download :download:\`File $i <files/$i/file.txt>\`
EOF
    
    echo "   $i" >> index.rst
done
```

### Environment info

- OS: Ubuntu 16.04
- Python version: 2.7.12 
- Sphinx version: 1.6.3

tk0miya:
Fixed by #5377.
Thank you for reporting!

PR #5350, opened 2018-08-25
[WIP] Fix #2720, #4034: Incorrect links with :download:, duplicate names, and parallel builds

### LINKED ISSUE #2720: Create unique directories, not unique filenames in _downloads
We have a similar problem as bug #823 with files such as CMakeLists.txt and Makefile. The reader will get confused when we tell them to download a demo file along with a CMakeLists.txt file and then they get CMakeLists5.txt which is not recognized by the cmake program, so they must remember to  rename the file before running the demo. 

We do not require as large a change as #823 (which is why I created a new issue), only that the file base names remain unchanged and unique prefix directories are created. A quick and ugly fix is stick this at the end of conf.py, but hotpatching internal Sphinx APIs is not very maintainable in the long run ...

``` python
# Hack to support avoid renaming download files with same names
# coming from different directories (CMakeLists.txt, Makefile etc)
from sphinx.util import FilenameUniqDict
from os import path
def add_file(self, docname, newfile):
    if newfile in self:
        self[newfile][0].add(docname)
        return self[newfile][1]
    uniquename = path.basename(newfile)
    i = 0
    while uniquename in self._existing:
        i += 1
        uniquename = path.join('subdir%s' % i, path.basename(newfile))
    self[newfile] = (set([docname]), uniquename)
    self._existing.add(uniquename)
    return uniquename
FilenameUniqDict.add_file = add_file

# Hack to support sub-directories within the download directory
from sphinx.builders.html import StandaloneHTMLBuilder, ensuredir, copyfile
from docutils.utils import relative_path
from sphinx.util.console import brown
def copy_download_files(self):
    def to_relpath(f):
        return relative_path(self.srcdir, f)
    # copy downloadable files
    if self.env.dlfiles:
        ensuredir(path.join(self.outdir, '_downloads'))
        for src in self.status_iterator(self.env.dlfiles,  # self.app.status_iterator in master ...
                                        'copying downloadable files... ',
                                        brown, len(self.env.dlfiles),
                                        stringify_func=to_relpath):
            dest = self.env.dlfiles[src][1]
            subdirs, filename = os.path.split(dest)
            if subdirs:
                ensuredir(path.join(self.outdir, '_downloads', subdirs))
            try:
                copyfile(path.join(self.srcdir, src),
                         path.join(self.outdir, '_downloads', dest))
            except Exception as err:
                self.warn('cannot copy downloadable file %r: %s' %
                          (path.join(self.srcdir, src), err))
StandaloneHTMLBuilder.copy_download_files = copy_download_files
```

This works-for-me! Would something similar be appreciated as a patch? I guess more builders would need to be changed?

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #2720, #4034

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,8 @@ Dependencies
 Incompatible changes
 --------------------
 
+* The URL of download files are changed
+
 Deprecated
 ----------
 
@@ -22,6 +24,8 @@ Bugs fixed
 * html: search box overrides to other elements if scrolled
 * i18n: warnings for translation catalogs have wrong line numbers (refs: #5321)
 * #5325: latex: cross references has been broken by multiply labeled objects
+* #2720, #4034: Incorrect links with ``:download:``, duplicate names, and
+  parallel builds
 
 Testing
 --------
--- a/sphinx/builders/html.py
+++ b/sphinx/builders/html.py
@@ -17,6 +17,7 @@ import types
 import warnings
 from hashlib import md5
 from os import path
+from typing import cast
 
 import docutils
 from docutils import nodes
@@ -33,6 +34,7 @@ from sphinx.application import ENV_PICKLE_FILENAME
 from sphinx.builders import Builder
 from sphinx.config import string_classes
 from sphinx.deprecation import RemovedInSphinx20Warning, RemovedInSphinx30Warning
+from sphinx.domains.download import DownloadDomain
 from sphinx.environment.adapters.asset import ImageAdapter
 from sphinx.environment.adapters.indexentries import IndexEntries
 from sphinx.environment.adapters.toctree import TocTree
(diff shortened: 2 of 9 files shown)


sphinx/domains/download.py
```diff
@@ -0,0 +1,86 @@
...
+    # For type annotation
+    from typing import Any, Dict, Iterable, Set  # NOQA
+    from docutils import nodes  # NOQA
+    from sphinx.application import Sphinx  # NOQA
+    from sphinx.environment import BuildEnvironment  # NOQA
+
+logger = logging.getLogger(__name__)
+
+
+class DownloadDomain(Domain):
```
tk0miya, line 32: `DownloadFileDomain` might be better.

sphinx/environment/collectors/asset.py
```diff
@@ -140,7 +140,6 @@ def process_doc(self, app, doctree):
 def setup(app):
     # type: (Sphinx) -> Dict
     app.add_env_collector(ImageCollector)
-    app.add_env_collector(DownloadFileCollector)
```
tk0miya, line 143: Note: I have to deprecate `DownloadFileCollector`.

sphinx/builders/html.py
```diff
@@ -859,15 +861,18 @@ def to_relpath(f):
             # type: (unicode) -> unicode
             return relative_path(self.srcdir, f)
         # copy downloadable files
-        if self.env.dlfiles:
```
tk0miya, line 862: Note: I have to deprecate `env.dlfiles`.

CHANGES
```diff
@@ -7,6 +7,8 @@ Dependencies
 Incompatible changes
 --------------------
 
+* The URL of download files are changed
```
tk0miya, line 10: After this change, all download files are placed at `{outdir}/_downloads/{hash}/{filename}`.
The `hash` is generated from its pathname. So it would be changed the location of srcdir.

Is there better idea?

sphinx/environment/collectors/asset.py
tk0miya: I don't know which is better way to implement this by Domain or EnvironmentCollector.

sphinx/domains/download.py
```diff
@@ -0,0 +1,86 @@
...
+                               location=node, type='download', subtype='not_readable')
+                node['invalid'] = True
+                continue
+
+            downloads.add(filename)
+            node['reftarget'] = self.get_destination_filename(filename)
+
+    def get_download_files(self):
+        # type: () -> Set[unicode]
+        return reduce(lambda x, y: x | y, self.data['objects'].values(), set())
```
shimizukawa, line 69: lambda can be replaced with `set.__or__`
```
reduce(set.__or__, self.data['objects'].values(), set())
```

I found better way.
```
set.union(*d.values())
```

shimizukawa (commented):
Whether this `download` is an independent domain or part of the `std` domain, I don't have a clear answer so far.

The current code is implemented as an independent domain. As the code itself has improved visibility, I think this direction is good.

tk0miya:
I withdrew this because this is breaking change. I'll work this again with another style.

sphinx/domains/download.py
tk0miya: Thanks.
I think `set().union(*d.values())` is much better because your code gets errored for empty list.

Issue #5351, opened 2018-08-26
C++: test_domain_cpp:test_expressions has been failed


tk0miya:
### Problem
- `test_domain_cpp:test_expressions` has been failed on my local
- It is only raised on 1.8 branch and py27.
    - 1.7 branch is fine.
    - 1.8 branch and py3 is also fine
- On Travis CI, it is not in trouble too.

@jakobandersen could you check this please? It seems this was happened only on my local.

#### Procedure to reproduce the problem
```
$ tox -e py27
```

#### Error logs / results
```
======================================================== test session starts =========================================================
platform darwin -- Python 2.7.15, pytest-3.7.2, py-1.5.4, pluggy-0.7.1
Running Sphinx test suite (with Python 2.7.15)...
rootdir: /Users/tkomiya/work/sphinx, inifile: setup.cfg
plugins: cov-2.5.1
collected 1323 items
(snip)
__________________________________________________________ test_expressions __________________________________________________________

    def test_expressions():
        def exprCheck(expr, id):
            ids = 'IE1CIA%s_1aE'
            check('class', 'template<> C<a[%s]>' % expr, {2: ids % expr, 3: ids % id})
        # primary
        exprCheck('nullptr', 'LDnE')
        exprCheck('true', 'L1E')
        exprCheck('false', 'L0E')
        ints = ['5', '0', '075', '0xF', '0XF', '0b1', '0B1']
        unsignedSuffix = ['', 'u', 'U']
        longSuffix = ['', 'l', 'L', 'll', 'LL']
        for i in ints:
            for u in unsignedSuffix:
                for l in longSuffix:
                    expr = i + u + l
                    exprCheck(expr, 'L' + expr + 'E')
                    expr = i + l + u
                    exprCheck(expr, 'L' + expr + 'E')
        for suffix in ['', 'f', 'F', 'l', 'L']:
            expr = '5.0' + suffix
            exprCheck(expr, 'L' + expr + 'E')
        exprCheck('"abc\\"cba"', 'LA8_KcE')  # string
        exprCheck('this', 'fpT')
        # character literals
        for p, t in [('', 'c'), ('u8', 'c'), ('u', 'Ds'), ('U', 'Di'), ('L', 'w')]:
            exprCheck(p + "'a'", t + "97")
            exprCheck(p + "'\\n'", t + "10")
            exprCheck(p + "'\\012'", t + "10")
            exprCheck(p + "'\\0'", t + "0")
[...]

PR #5356, opened 2018-08-26
C++, conditionally disable test on sys.maxunicode

jakobandersen:
### Feature or Bugfix
- Bugfix

### Relates
- #5351


tk0miya (approved):
Thank you for quick work!

PR #5359, opened 2018-08-27
Make generated texinfo files reproducible by sorting the anchors

mitya57:
This fixes the problem with reproducible builds originally reported in Debian: https://bugs.debian.org/907352.

### DIFF
--- a/tests/test_build_texinfo.py
+++ b/tests/test_build_texinfo.py
@@ -50,6 +50,10 @@ def test_texinfo_warnings(app, status, warning):
 def test_texinfo(app, status, warning):
     TexinfoTranslator.ignore_missing_images = True
     app.builder.build_all()
+    result = (app.outdir / 'SphinxTests.texi').text(encoding='utf8')
+    assert ('@anchor{markup doc}@anchor{12}'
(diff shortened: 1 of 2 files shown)


tests/test_build_texinfo.py
```diff
@@ -53,6 +53,12 @@ def test_texinfo(app, status, warning):
     # now, try to run makeinfo over it
     cwd = os.getcwd()
     os.chdir(app.outdir)
+    with open('SphinxTests.texi') as texi_file:
+        # make sure anchors are sorted
+        texi_source = texi_file.read(8000)
```
tk0miya: We usually use `path.text()` to read the content on testing scripts. Then you don't need to open a file.
Could you use `(app.outdir / 'SphinxTests.texi').text(encoding='utf8')` instead?

tk0miya (changes requested):
LGTM with nits

mitya57:
@tk0miya Fixed! The Python 3.5 build on Travis got stuck, can you please retry it?

tk0miya (approved):

tk0miya:
Merged. Thank you always your support!

PR #5363, opened 2018-08-29
Fix #5361: crashed on incremental build if document uses include directive

### LINKED ISSUE #5361: v1.7.8 breaks compiling document previously compiled with older version
### Problem
v1.7.8 fails compiling a document (in our case [the HyperSpy documentation](https://hyperspy.readthedocs.io/en/latest/)) that v1.7.7 compiles happily.

v1.7.7 works fine on the same document.

```
make html
```

#### Error logs / results


```bash
reading sources... [  0%] api/hyperspy                                                                                                                                                                                                                                                    
Exception occurred:
  File "/home/user/anaconda/anaconda/lib/python3.6/site-packages/sphinx/environment/__init__.py", line 313, in clear_doc
    self.included.pop(docname, None)
TypeError: pop() takes no arguments (2 given)
The full traceback has been saved in /tmp/sphinx-err-q3vhqr4r.log, if you want to report the issue to the developers.
```

Full build error in ReadTheDocs: https://readthedocs.org/projects/hyperspy/builds/7701605/

https://github.com/hyperspy/hyperspy/tree/FIX_RTD_apidoc


### Environment info

ReadTheDocs and local Anaconda installation running on Debian.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5361

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -72,7 +72,7 @@ default_settings = {
 # or changed to properly invalidate pickle files.
 #
 # NOTE: increase base version by 2 to have distinct numbers for Py2 and 3
-ENV_VERSION = 52 + (sys.version_info[0] - 2)
+ENV_VERSION = 54 + (sys.version_info[0] - 2)
 
 
 versioning_conditions = {
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,8 @@ Features added
 Bugs fixed
 ----------
 
+* #5361: crashed on incremental build if document uses include directive
+
 Testing
 --------
 


sphinx/environment/__init__.py
```diff
@@ -72,7 +72,7 @@
 # or changed to properly invalidate pickle files.
 #
 # NOTE: increase base version by 2 to have distinct numbers for Py2 and 3
-ENV_VERSION = 52 + (sys.version_info[0] - 2)
+ENV_VERSION = 53 + (sys.version_info[0] - 2)
```
mitya57: According to the comment above it should be increased by 2, so 54 instead of 53.

Otherwise Sphinx 1.7.9 with Python 2 will have the same version as Sphinx 1.7.8 with Python 3.

mitya57 (changes requested):

sphinx/environment/__init__.py
tk0miya: oh, you're right. I'd forgotten it...

PR #5368, opened 2018-09-02
Deprecate arguments of env.doc2path()

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- To make `env.doc2path()` simple, I'd like to deprecate two arguments
    - `suffix`: it is only used to create new path from docname. I think string-concatenation is enough to do that.
    - `base` (basepath): it is only used to create new path from docname. It is same as `os.path.join()`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,7 @@ Deprecated
 ----------
 
 * The ``suffix`` argument of ``env.doc2path()`` is deprecated.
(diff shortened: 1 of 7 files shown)


shimizukawa (approved):
LGTM!

tk0miya:
Thank you for reviewing!

PR #5369, opened 2018-09-02
Add project object

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To reduce the responsibility of BuildEnvironment, add a Project class to manipulate project files.

### DIFF
--- a/doc/extdev/index.rst
+++ b/doc/extdev/index.rst
@@ -143,6 +143,11 @@ The following is a list of deprecated interfaces.
      - 4.0
      - N/A
 
+   * - ``sphinx.util.get_matching_docs()``
+     - 2.0
+     - 4.0
+     - ``sphinx.util.get_matching_files()``
+
    * - ``sphinx.util.osutil.walk()``
      - 2.0
      - 4.0
--- a/doc/extdev/index.rst
+++ b/doc/extdev/index.rst
@@ -85,6 +85,7 @@ APIs used for writing extensions
 
    tutorial
    appapi
+   projectapi
    envapi
    builderapi
    collectorapi
--- a/doc/extdev/projectapi.rst
+++ b/doc/extdev/projectapi.rst
@@ -1,0 +1,9 @@
+.. _project-api:
+
+Project API
+===========
+
+.. currentmodule:: sphinx.project
+
+.. autoclass:: Project
+   :members:
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -390,7 +390,7 @@ class BuildEnvironment:
             exclude_paths = (self.config.exclude_patterns +
                              self.config.templates_path +
                              builder.get_asset_paths())
-            self.project.discovery(exclude_paths)
+            self.project.discover(exclude_paths)
 
             # Current implementation is applying translated messages in the reading
             # phase.Therefore, in order to apply the updated message catalog, it is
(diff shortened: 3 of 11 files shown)


doc/extdev/index.rst
```diff
@@ -217,6 +223,11 @@ The following is a list of deprecated interface.
      - 3.0
      - N/A
 
+   * - ``BuildEnvironment.get_outdated_files()``
```
tk0miya: This PR does not contains this change. This is noise...

doc/extdev/projectapi.rst
```diff
@@ -0,0 +1,9 @@
+.. _project-api:
```
tk0miya, line 1: OMG, this file is orphaned!

sphinx/environment/__init__.py
```diff
@@ -407,25 +396,22 @@ def relfn2path(self, filename, docname=None):
             enc_rel_fn = rel_fn.encode(sys.getfilesystemencoding())
             return rel_fn, path.abspath(path.join(self.srcdir, enc_rel_fn))
 
+    @property
+    def found_docs(self):
+        # type: () -> Set[unicode]
+        """contains all existing docnames."""
```
tk0miya, line 381: Note: `env.found_docs` has used for long time. So I'd not like to deprecate this. It might be a big breaking change!

doc/extdev/projectapi.rst
tk0miya: This comment was my wrong. This is surely referenced from `doc/extdev/index.rst`. np.

sphinx/project.py
```diff
@@ -0,0 +1,98 @@
...
+
+        #: The name of documents belongs to this project.
+        self.docnames = set()  # type: Set[unicode]
+
+    def restore(self, other):
+        # type: (Project) -> None
+        """Take over a result of last build."""
+        self.docnames = other.docnames
+
+    def discovery(self, exclude_paths=[]):
```
shimizukawa: It must be `discover` as a verb.

sphinx/environment/__init__.py
shimizukawa: OK. It would be better if the reason was written in the docstring of this function.

shimizukawa (approved):
LGTM with nits.

sphinx/project.py
tk0miya: done

sphinx/environment/__init__.py
tk0miya: what did you mean "reason"?

shimizukawa: It means your 'Note'

> Note: `env.found_docs` has used for long time. So I'd not like to deprecate this. It might be a big breaking change!

tk0miya: I feel strange. Are you saying to add "this is not deprecated yet"?

shimizukawa: Since you wrote a note for developers on PR, I thought it was worth keeping in the comments of the code. I think that the part of "(Removing this method) might be a big breaking change" is an important note.

tk0miya: This is a public API which has been longly used.
http://www.sphinx-doc.org/en/master/extdev/envapi.html#sphinx.environment.BuildEnvironment.found_docs

So removing this will bring big breaking change as you know.
I don't think it is not important decision. To keep API stable is our responsibility. I believe Not to remove public APIs is not special things for us. Is it really need to add comment? Should I add to all promised APIs to the comment?

tk0miya: My concern is what will be given for the readers of our code or document. Removing public APIs  will always bring breaking change. So I think the comment is meaningless.

tk0miya: I think this discussion is stalled. But it's trivial. So I'm merging this now.
If you still have opinion, please let me know in this thread. I'll respond to your comment even after merged.

Thanks,

PR #5377, opened 2018-09-02
Fix #2720, #4034: Incorrect links with ``:download:``, duplicate names, and parallel builds

### LINKED ISSUE #2720: Create unique directories, not unique filenames in _downloads

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #2720, #4034
- An alternative of #5350

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@ Incompatible changes
 
 * #5282: html theme: refer ``pygments_style`` settings of HTML themes
   preferentially
+* The URL of download files are changed
 
 Deprecated
 ----------
@@ -30,6 +31,8 @@ Bugs fixed
 * #5348: download reference to remote file is not displayed
 * #5282: html theme: ``pygments_style`` of theme was overrided by ``conf.py``
   by default
+* #2720, #4034: Incorrect links with ``:download:``, duplicate names, and
+  parallel builds
 
 Testing
 --------
--- a/sphinx/builders/html.py
+++ b/sphinx/builders/html.py
@@ -864,10 +864,10 @@ class StandaloneHTMLBuilder(Builder):
             for src in status_iterator(self.env.dlfiles, __('copying downloadable files... '),
                                        "brown", len(self.env.dlfiles), self.app.verbosity,
                                        stringify_func=to_relpath):
-                dest = self.env.dlfiles[src][1]
                 try:
-                    copyfile(path.join(self.srcdir, src),
-                             path.join(self.outdir, '_downloads', dest))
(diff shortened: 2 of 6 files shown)


MarkMarsh:
This has broken our external download links - could it be made optional? We had no problem with the original behaviour.

PR #5384, opened 2018-09-03
quickstart: Simplify generated conf.py (for latex)

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #4256
- This removes `latex_documents` and `latex_elements`
- I will also remove settings for other builders if this approved.


sphinx/util/osutil.py
```diff
@@ -175,13 +175,19 @@ def copyfile(source, dest):
...
 no_fn_re = re.compile(r'[^a-zA-Z0-9_-]')
+project_suffix_re = re.compile(' Documentation')
 
 
 def make_filename(string):
     # type: (str) -> unicode
     return no_fn_re.sub('', string) or 'sphinx'
 
 
+def make_filename_from_project(project):
```
tk0miya: I don't know here is appropriate location of this function. But I placed it beside of `make_filename()`.

sphinx/util/osutil.py
```diff
@@ -175,13 +175,19 @@ def copyfile(source, dest):
...
+    # type: (str) -> unicode
+    return make_filename(project_suffix_re.sub('', project))
```
tk0miya: It would be better to make lowercase.

sphinx/templates/quickstart/conf.py_t
```diff
@@ -116,35 +116,6 @@ html_static_path = ['{{ dot }}static']
...
-    #
-    # 'figure_align': 'htbp',
-}
-
-# Grouping the document tree into LaTeX files. List of tuples
-# (source start file, target name, title,
-#  author, documentclass [howto, manual, or own class]).
-latex_documents = [
-    (master_doc, '{{ project_fn }}.tex', u'{{ project_doc_texescaped_str }}',
-     u'{{ author_texescaped_str }}', 'manual'),
```
tk0miya: We still have these template variables in quickstart code.
We have to deprecate them in nearly future.

sphinx/builders/latex/__init__.py
```diff
@@ -405,6 +405,16 @@ def default_latex_use_xindy(config):
     return config.latex_engine in {'xelatex', 'lualatex'}
 
 
+def default_latex_documents(config):
+    # type: (Config) -> List[Tuple[unicode, unicode, unicode, unicode, unicode]]
+    """ Better default latex_use_xindy settings. """
```
jfbu: is this docstring mentioning `latex_use_xindy` an unrelated left-over?

jfbu (commented):
LGTM with nits

- I always found `latex_documents` a bit frightening, so removing it from default conf.py sound ok

- but then whatabout `man` and `texinfo`?

- regarding `latex_elements` I found useful in learning Sphinx to have it in conf.py, at least to see immediately how to set up for `a4paper`, or for `12pt`.

So I have mixed feelings about not having some `latex_elements` in conf.py for new project.

tk0miya:
>but then whatabout man and texinfo?

They will be also removed in next PR.

sphinx/builders/latex/__init__.py
tk0miya: oh, it's my wrong.

sphinx/templates/quickstart/conf.py_t
shimizukawa: I think it would  be better if there was a link to a documentation of latex configuration.

shimizukawa (approved):
LGTM with nits.

sphinx/templates/quickstart/conf.py_t
tk0miya: We've already have a link for configurations.
https://github.com/sphinx-doc/sphinx/blob/d2c3d54bb9a46c5e7a0d14c3345e489d7592de9a/sphinx/templates/quickstart/conf.py_t#L7

Did you mean that we should give a help only for latex configuration especially?

shimizukawa: Yes. Guidance to the document has effects like DeprecationWarning for those who search the keyword "latex" in "conf.py", I think that we can reduce troublesome question mails/issues.

tk0miya: I'll also remove settings for manpages, texinfo and others. If my understanding is right, many comments will be listed in new conf.py. Is your goal like this? Or, would you like to add comment only for LaTeX output?
```
# -- Options for LaTeX output ------------------------------------------------
# see http://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-latex-output

# -- Options for man pages output ------------------------------------------------
# see http://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-manual-page-output

# -- Options for Texinfo output ------------------------------------------------
# see http://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-texinfo-output
```

stephenfin: I'd personally be against this. If we're cleaning this up then there's no reason to go add more noise. The documentation for each of those builders is very easy to find on http://www.sphinx-doc.org/en/master/config. I don't think we need to be even more specific. Just my 2c

tk0miya: ping @shimizukawa . Any comments?

jfbu: The config documentation page is big. Regarding latex and latex_documents in particular I do think the conf.py must have at least one commented out use case. With this, 99% of people will be up an running quick, without it they will go to http://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-latex-output and a large percentage will get scared and uncomfortable about spending 20 minutes reading customization where perhaps they don't know what LaTeX is.

tk0miya: The purpose of `latex_documents` is:

* To split a document to multiple PDF files
* To change a docclass of LaTeX output
* To give a special title and authors information to LaTeX output

I understand changing a docclass is a bit important for some users. But others are less (I think).


If it was important, `man_pages` and `texinfo_documents` are also important. Is this right?

stephenfin: I don't have a vote but I personally agree with everything you just said there, @tk0miya. Like it or not, HTML is still the default builder chosen by most users. These are likely the only config options we really need.

jfbu: To the extent `latex_documents` is deprecated and a suitable default will get set from the project configuration metadata (project, authors, release), I guess indeed it should be removed. 

I would suggest moving http://www.sphinx-doc.org/en/master/usage/configuration.html#project-information to top of configuration documentation.

I personally used Sphinx for some time helped with the sphinx-quickstart created conf.py and only later started having a look at general documentation; which I have actually never managed to read fully yet... thus I think I will keep some nostalgy for richer conf.py, but go ahead.

tk0miya:
I just made a change for texinfo in #5607.
I think latex is major and important component for Sphinx. So it is hard to discuss as a first step. So let's discuss about texinfo first.

tk0miya:
I just updated this PR. This removes both `latex_documents` and `latex_elements`. I think only less users customize them. So we can remove them like `texinfo_documents` (#5607), `htmlhelp_basename` (#5698) and `man_pages` (#5699).
If you feel both (or either one) is important, I will withdraw this PR.

Please let me know your opinion (again).

Thanks,

jfbu (commented):
LGTM with one question: I noticed that on this branch the latex builds (i.e. doing pdflatex etc... on produced LaTeX files) from `tests/test_build_latex.py` are skipped, 

```
tests/test_build_latex.py::test_build_latex_doc[pdflatex-howto] SKIPPED  [  1%]
tests/test_build_latex.py::test_build_latex_doc[pdflatex-manual] SKIPPED [  3%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-howto] SKIPPED  [  4%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-manual] SKIPPED [  6%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-howto] SKIPPED   [  7%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-manual] SKIPPED  [  9%]
```

is this intended?

jfbu (commented):
I am wondering if the conf.py could not contain

```
# -- Options for LaTeX output -------------------------------------------------
# http://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-latex-output
# http://www.sphinx-doc.org/en/master/latex.html
```

or maybe we discussed that already and it is felt to be redundant?

jfbu (commented):
Sorry for my sloppy reviewing, another question, I noticed `sphinx.util.osutil.make_filename_from_project()` will keep only `a-zA-Z0-9_-` characters, doesn't that possibly create empty names from a Japanese project name?

ah sorry in that case `'sphinx'` is used...

tk0miya:
>LGTM with one question: I noticed that on this branch the latex builds (i.e. doing pdflatex etc... on produced LaTeX files) from tests/test_build_latex.py are skipped,

Wow! It's not intended. Must be a bug. I'll take a look tomorrow.

>or maybe we discussed that already and it is felt to be redundant?

It was discussed at this thread: https://github.com/sphinx-doc/sphinx/pull/5384#discussion_r221449246 . But it was not continued...
IMO, adding hyperlinks to all of builders is ugly.

My personal goal is that the generated conf.py contains only very few settings; project information (`project`, `author` and `version`), `language`, `extensions` and so on. In other words, conf.py would be about 10 lines by default (if my idea is adopted). To do that, I'd like to shrink the content of the file.

> Sorry for my sloppy reviewing, another question, I noticed sphinx.util.osutil.make_filename_from_project() will keep only a-zA-Z0-9_- characters, doesn't that possibly create empty names from a Japanese project name?

It is almost same as quickstart does.

tk0miya:
>LGTM with one question: I noticed that on this branch the latex builds (i.e. doing pdflatex etc... on produced LaTeX files) from tests/test_build_latex.py are skipped,

I fixed this. I used wrong filename, but MacOS does not treat it as an error because of filesystem.

sphinx/builders/latex/__init__.py
```diff
@@ -468,6 +468,16 @@ def default_latex_use_xindy(config):
     return config.latex_engine in {'xelatex', 'lualatex'}
 
 
+def default_latex_documents(config):
+    # type: (Config) -> List[Tuple[str, str, str, str, str]]
+    """ Better default latex_documents settings. """
+    return [(config.master_doc,
+             make_filename_from_project(config.project) + '.tex',
+             texescape.escape(config.project),
+             texescape.escape(config.author),
```
jfbu: It could make sense to also use the new `\texescape.escape_abbr` here.

jfbu (approved):
LGTM, see remark about possibly using `texescape.escape_abbr`.

sphinx/builders/latex/__init__.py
jfbu: sorry for `\tex..`, I do too much LaTeX macros `:)`

tk0miya: fixed.

jfbu (approved):
LGTM! Time to merge!

tk0miya:
Merged. Thank you all for reviewing!

jfbu:
> > LGTM with one question: I noticed that on this branch the latex builds (i.e. doing pdflatex etc... on produced LaTeX files) from tests/test_build_latex.py are skipped,
> 
> I fixed this. I used wrong filename, but MacOS does not treat it as an error because of filesystem.

@tk0miya at master (97d99f830) PDF builds are still skipped in testing

```
============================= test session starts ==============================
platform darwin -- Python 3.6.8, pytest-4.1.1, py-1.7.0, pluggy-0.8.1 -- /xxxyyyzzz/sphinxvenv/bin/python3
cachedir: .pytest_cache
libraries: Sphinx-2.0.0+/97d99f830, docutils-0.14
rootdir: /xxxyyywww/sphinx, inifile: setup.cfg
collected 64 items                                                             

tests/test_build_latex.py::test_build_latex_doc[pdflatex-howto] SKIPPED  [  1%]
tests/test_build_latex.py::test_build_latex_doc[pdflatex-manual] SKIPPED [  3%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-howto] SKIPPED  [  4%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-manual] SKIPPED [  6%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-howto] SKIPPED   [  7%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-manual] SKIPPED  [  9%]
tests/test_build_latex.py::test_writer PASSED                            [ 10%]
...

==================== 57 passed, 7 skipped in 13.39 seconds =====================
(sphinxvenv) (base) 
```

PR #5385, opened 2018-09-03
The default setting for master_doc is changed to 'index'

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #4256
- To diet `conf.py`, this changes the default setting of `master_doc`.


shimizukawa (approved):
Wow, it's seems big change, but looks good to me.
I think most documents don't prepare the top page with their own custom template.

stephenfin (approved):
Ooh, yes please!

tk0miya:
Merged!

jfbu:
@tk0miya this change breaks builds of projects such as CPython Doc. They don't have an `index.rst` but a `contents.rst` and we get following error:

```
(sphinxvenv) 
[19:02:08] foo:/Users/xxxxxx/cpython/Doc
Doc$ make html VENVDIR=/Users/xxxxxx/sphinxvenv
mkdir -p build
Building NEWS from Misc/NEWS.d with blurb
PATH=/Users/xxxxxx/sphinxvenv/bin:$PATH sphinx-build -b html -d build/doctrees -D latex_elements.papersize=  . build/html 
Running Sphinx v2.0.0+/f51ef51db6
building [html]: targets for 475 source files that are out of date
updating environment: 475 added, 0 changed, 0 removed
reading sources... [100%] whatsnew/index                                        

Sphinx error:
master file /Users/xxxxxx/cpython/Doc/index.rst not found
make: *** [Makefile:45: build] Error 2
```

See https://github.com/python/cpython/tree/master/Doc

jfbu (commented):
This apparently breaks builds of old projects which did not have an `index.rst` but a `contents.rst`. e.g. https://github.com/python/cpython/tree/master/Doc, I can't build it after this patch.

stephenfin:
@jfbu They just need to manually set `master_doc ='contents.rst'`, right? It's expected that things would break in a major release.

stephenfin:
It should also go without saying but they should be versioning their dependencies. It's why semver exists :smile:

jfbu:
@stephenfin 

> They just need to manually set `master_doc ='contents.rst'`, right? It's expected that things would break in a major release.

I agree with that fix except that it should be `master_doc = 'contents'` but I wonder if the documentation is explicit enough about what to do when upgrading Sphinx to 2.0.

stephenfin:
Ah, correct. Personally, I would rather we used a proper release notes tool for this stuff (`reno` includes a per-release `prelude` section for just this purpose) but I've lost that battle. As things stand, I guess if we don't have `.. versionchanged` notes in the option we could add them. We could probably also add an intro to the change log noting big changes like this. Would that be sufficient?

jfbu:
@stephenfin +1 for some extra sectioning (I don't know which title to use; perhaps "Upgrading") in `CHANGES`. This patch already includes an entry in section `Incompatible changes` but I feel it would not hurt to explicitely state "Add `master_doc = 'contents'` to your conf.py to maintain backwards compatibility in your existing projects". This could be simply added in `Incompatible changes` or be part of a new "Upgrading" section.

tk0miya:
+1 for Upgrading. In addition, it would be nice if we give better error message for the projects which does not set `master_doc`.

Issue #1063, opened 2015-01-03
automodule :undoc-members: does not include undocumented module constants


shimizukawa:
The only way of getting undocumented constants defined in the module into the documentation seems to be to explicitly define:

```
   .. py:data:: SOMECONST
```

It would be less surprising if :undoc-members: option to automodule sucked in undocumented constants as well.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1063
- Originally reported by: Daniel Svensson
- Originally created at: 2012-12-18T22:32:14.449

tk0miya:
Certainly, module global variables not having docstring is not documented even if `:undoc-members:`.

jsmedmar:
Wow, this issue was created in 2012. Are there any plans of making sure undocumented constants make it to the doc when using `:undoc-members:
`?

tk0miya:
While six years, nobody had not tried to fix this. How about you? PR is always welcome :-)

jsmedmar:
haha! guess no one has been lazy enough to complain about having to doc their constants

jsmedmar:
Ok @tk0miya, see #5387.

jsmedmar:
@tk0miya, do you know how I could differentiate from objects defined in the module between those imported? That was the roadblock I hit whilst implementing a solution for this ticket in #5387.

tk0miya:
AFAIK, there are no way to investigate it. All classes and functions have `__module__` attribute indicates where the object defined. But instances does not have it.
I guess our pycode module (`sphinx.pycode`) help it.

I'll take a look after releasing 1.8.0.

Thank you for your help.

rmorshea:
You can work around this by adding your constants to `__all__`. This is tedious though.

tk0miya:
Fixed by #6423 .
Thank you for reporting.

tk0miya:
Reopened now. Because #6423 causes some troubles. I'll try this again in future version.

goekce:
> You can work around this by adding your constants to `__all__`. This is tedious though.

Did not work for me. I am using `sphinx-build` 3.41 with `conf.py`:

```
...
autodoc_default_options = {
    'members': True,
    'undoc-members': True,
}
```
I included some module-level variables in `__all__`, but only the variables which have `#: comment` were picked up :-/

con-f-use:
https://github.com/sphinx-doc/sphinx/pull/6423 doesn't seem to have fixed this or we have a regression.

[...]

PR #5387, opened 2018-09-04
[WIP] include doc-less data objects with undoc-members

jsmedmar:
Make sure undocumented module level constants (or data objects) are included  when `undoc_members=True`, see #1063.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ TAGS
 .tox/
 .tx/
 .venv/
+.vscode/
 .coverage
 htmlcov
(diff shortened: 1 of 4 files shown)


jsmedmar:
Ok, this is wrong. This will include also imports haha... Now I see why this was not implemented before. I may close the PR.

jsmedmar:
Closing since couldn't find a way to differentiate those objects defined in a module between those imported into the module. If someone knows how to do it, it would be greatly appreciated.

PR #5398, opened 2018-09-08
Drop branches for sys.version_info < (3, 4)

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- As a next step of farewell python 2, this drops PY2 and PY3 flags from our code.


sphinx/ext/autodoc/importer.py
```diff
@@ -245,6 +236,9 @@ def get_object_members(subject, objpath, attrgetter, analyzer=None):
     # other members
     for name in dir(subject):
         try:
+            # the members directly defined in the class
+            obj_dict = attrgetter(subject, '__dict__', {})
```
tk0miya: This should be placed out of loop. This was my wrong.

sphinx/util/__init__.py
```diff
@@ -677,8 +677,7 @@ def xmlname_checker():
         [u'\u2C00', u'\u2FEF'], [u'\u3001', u'\uD7FF'], [u'\uF900', u'\uFDCF'],
         [u'\uFDF0', u'\uFFFD']]
 
-    if sys.version_info.major == 3:
-        name_start_chars.append([u'\U00010000', u'\U000EFFFF'])
+    name_start_chars.append([u'\U00010000', u'\U000EFFFF'])
```
tk0miya, line 680: No reason to append them at here. This should be merged to the definition.

PR #5402, opened 2018-09-08
Remove use of Python 2 os.getcwdu()

jdufresne:
os.getcwdu() is Python 2 only. It was removed from Python 3. As Sphinx
is now Python 3 only, can remove the workaround.

```
$ python2 -c 'import os; print(os.getcwdu)'
<built-in function getcwdu>
$ python3 -c 'import os; print(os.getcwdu)'
  File "<string>", line 1, in <module>
AttributeError: module 'os' has no attribute 'getcwdu'
```

### DIFF
--- a/sphinx/util/osutil.py
+++ b/sphinx/util/osutil.py
@@ -25,7 +25,7 @@ from os import path
 
 from six import PY2, PY3, text_type
 
-from sphinx.deprecation import RemovedInSphinx30Warning
+from sphinx.deprecation import RemovedInSphinx30Warning, RemovedInSphinx40Warning
 
 if False:
     # For type annotation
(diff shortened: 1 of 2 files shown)


sphinx/util/osutil.py
```diff
@@ -241,17 +241,10 @@ def abspath(pathdir):
     return pathdir
 
 
-def getcwd():
```
tk0miya: To keep compatibility, please don't remove this. Unfortunately some extensions might has used this.
To do that, please do following procedure:

1. Show deprecation message using `warnings.warn()` and `RemovedInSphinx40Warning` inside this function.
2. Move this to `sphinx.util.compat`.
3. Import it on this module.

tk0miya (changes requested):
LGTM with nits

jdufresne:
> 2. Move this to sphinx.util.compat.
> 3. Import it on this module.

Unfortunately, this creates a circular import. `sphinx.util.compat` has the import `from sphinx.util import import_object` and `sphinx/util/__init__.py` has the import `from sphinx.util.osutil import fs_encoding`. So `sphinx/util/osutil.py` can't also import `sphinx.util.compat` as the top level.

I have restored `sphinx.util.osutil.getcwd()` with a deprecation warning.

tk0miya (approved):

tk0miya:
Thank you for quick work! And sorry for my wrong comment.

PR #5403, opened 2018-09-08
Remove unused argument, content, from guess_mimetype

jdufresne:
Unused since its introduction in be261ed71ec797d32fe88dae1acf73adb2b33c8e.

### DIFF
--- a/sphinx/util/images.py
+++ b/sphinx/util/images.py
@@ -12,12 +12,15 @@ from __future__ import absolute_import
 
 import base64
 import imghdr
+import warnings
 from collections import OrderedDict
 from os import path
 from typing import NamedTuple
 
 import imagesize
 from six import PY3, BytesIO, iteritems
+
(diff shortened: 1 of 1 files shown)


tk0miya:
OMG. It's not my intension. We should use it for remote images...

tk0miya:
Okay, I understand why I did not use the argument. Surely, it is not needed because all images are passed by file, not a string.

As commented other PR, could you add deprecation message instead of removal?
```
if content:
    warning.warn('content argument of guess_mimetype() is deprecated', RemovedInSphinx30Warning)
```

tk0miya (changes requested):
LGTM with nits.
In addition, it would be better to rebase this into 1.8 branch.

Thanks,

jdufresne:
- Changed removal to deprecation with a warning (restored tests).
- Rebased to 1.8

tk0miya (approved):

tk0miya:

PR #5410, opened 2018-09-11
Deprecate unused argument, mode, from HTMLHelpBuilder.open_file()

jdufresne:
Unused since its introduction in d976bb25f4b6ef646c9a3e51aa6e153bcf44eae8.

### DIFF
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -13,6 +13,7 @@ from __future__ import print_function
 
 import codecs
 import os
+import warnings
 from os import path
 
 from docutils import nodes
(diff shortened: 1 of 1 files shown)


sphinx/builders/htmlhelp.py
```diff
@@ -207,6 +209,9 @@ def init(self):
 
     def open_file(self, outdir, basename, mode='w'):
         # type: (unicode, unicode, unicode) -> IO
+        if mode != 'w':
+            warnings.warn('The mode argument of HTMLHelpBuilder.open_file() is deprecated.',
+                          RemovedInSphinx40Warning)
```
tk0miya, line 214: Instead of this, it would be better to give `mode` option explicitly to me. I think "explicit" is important concept for Python programs.
So -1 for this.

jdufresne:
>  So -1 for this.

Makes sense. No problem.

tk0miya:
I just made #5415 as an answer for your PR. I think using bare `open()` function is simple way. No special wrapper is not needed.

PR #5411, opened 2018-09-11
Prefer builtin open() over io.open() and codecs.open()

jdufresne:
In Python3, the functions `io.open()` is an alias of the builtin `open()` and `codecs.open()` is functionally equivalent. To reduce indirection, number of imports, and number of patterns, always prefer the builtin.

https://docs.python.org/3/library/io.html#high-level-module-interface

> io.open()
>
> This is an alias for the builtin open() function.


tk0miya:
Merged this manually. Thanks!

PR #5412, opened 2018-09-11
Remove use of six.iteritems()

jdufresne:
In Python 3, `dict.items()` is always an iterator.


tk0miya:
Now I'm afraid of these dropping py2 refactoring causes conflicts with 1.8 branch. 1.8.x will be released soon. And it will be maintained until the release of 2.0 (maybe 2019 Spring). So merging smoothly is important point for us.

On the other hand, I'd like to drop "six" module from our dependency list. 

@shimizukawa what do you think? Do you have any idea?

tk0miya:
It might be better to add helper function like following:
```
# in sphinx.util.pycompat
def iteritems(itererable):
    return iterable.iteritems()
```

I know main code is still ugly. But this can lessen the changes of conflicts.

tk0miya:
On a side note, I think we can easily merge 1) small changes like #5413 2) removal whole of blocks like #5398. But this and #5411 (`io.open()` and `codecs.open()`) are widely used in main code. So they might be conflicted with the bug fixes in 1.8 branch often.

I know we have many test codes. So they will save us and our code. TBH, I'm simply afraid of maintenance of old branch becomes bothersome and tired.

jdufresne:
Thanks for the feedback @tk0miya. I see the concern.

What do you think about holding off on merging this and other big changes until closer to the 2.0 release? We could wait until 1.8 settles down and stabilizes to get beyond the merging concern. That way, we'd still be able to move the code to modern Python idioms while still supporting the 1.8 branch for a sufficient amount of time. If you think this is a good approach, please feel free to put this (and other PRs) on hold and then ping me when it would be a good time to resume work on them. I don't mind rebasing and continuing the work at that time. WDYT?

tk0miya:
Good idea. I like it :-)

tk0miya:
Finally, I determine to merge this. I think removal of py2 will bring large changes to our codebase in addition to this.
So postponing the merge does not resolve our problem.

I know it will be painful. But I'd like to make our code clean ASAP. So I'm merging this.

PR #5413, opened 2018-09-11
Remove use of six.class_types

jdufresne:
In Python 3, type is the only class type.


tk0miya (approved):
LGTM!

tk0miya:
Merged. Thanks!

PR #5414, opened 2018-09-11
Remove unnecessary object from class definitions

jdufresne:
In Python 3, all classes are new-style classes. The object in the definition is redundant and unnecessary.


tk0miya (approved):
LGTM!

tk0miya:
Merged. Thanks!

PR #5415, opened 2018-09-12
refactor: Repalce HTMLHelpBuilder.open_file() by standard open()

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- An alternative of #5410.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -70,6 +70,7 @@ Deprecated
   ``IndexBuilder.feed()`` method is deprecated.
 * ``sphinx.addnodes.abbreviation``
 * ``sphinx.application.Sphinx._setting_up_extension``
+* ``sphinx.builders.htmlhelp.HTMLHelpBuilder.open_file()``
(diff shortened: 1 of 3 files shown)


jdufresne:
I don't see anything wrong with this as long as you find the duplication of `encoding=self.encoding, errors='xmlcharrefreplace'` acceptable. I see some benefit of centralizing how encodings and errors are handled, although, admittedly small.

tk0miya:
oh, I'd forgotten to comment to this...

Indeed `open_file()` reduces duplication a bit. But, IMO, the standard function `open()` is better to open a file. No special macros are not needed only for opening a file. You do not have to read the code to know what `open_file()` doing.
I think this is part of "zen of python"; explicit, simple and readable (easy to understand).

jdufresne (approved):

tk0miya:
Thank you for reviewing. Merged.

PR #5429, opened 2018-09-14
Fix #5422: lambda object causes PicklingError on storing environment

### LINKED ISSUE #5422: Pickling Error with 1.8
Subject:  Sphinx 1.8 introduced a build error for JupyterLab

### Problem
- It appears that in 1.8 the `url_resolver` config is trying to get pickled.  We had a lambda function, but that broke with (cf https://travis-ci.org/jupyterlab/jupyterlab/jobs/427847089#L672):

```
AttributeError: Can't pickle local object 'setup.<locals>.<lambda>'
```

Changing it to a separate function yields (cf https://travis-ci.org/jupyterlab/jupyterlab/jobs/428072324#L763 and https://github.com/jupyterlab/jupyterlab/pull/5308/files):

```
_pickle.PicklingError: Can't pickle <function url_resolver at 0x2b0913492e18>: attribute lookup url_resolver on __main__ failed
```

```
git clone https://github.com/jupyterlab/jupyterlab
cd jupyterlab/docs
pip install sphinx=1.8 sphinx_rtd_theme recommonmark
make html
```

### Environment info
- OS: MacOS 10.13.6
- Python version: 3.6.3
- Sphinx version: 1.8.0

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5422

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Bugs fixed
 
 * #5418: Incorrect default path for sphinx-build -d/doctrees files
 * #5421: autodoc emits deprecation warning for :confval:`autodoc_default_flags`
+* #5422: lambda object causes PicklingError on storing environment
 
 Testing
 --------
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

PR #5440, opened 2018-09-16
Merge setting for filterwarnings

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- We have two settings for filterwarnings. This merges them to single setting.


tk0miya:
Unfortunately, this does not work well. Withdrawed.

PR #5442, opened 2018-09-16
Fix #5437: autodoc: crashed on modules importing eggs

### LINKED ISSUE #5437: `get_module_source` exception in Sphinx 1.8.0 when using autodoc on modules importing eggs
Hi, I ran into an issue when trying to deploy our documentation on a Win10 CI server, using the recent Sphinx 1.8.0 release.

### Problem

`get_module_source` raises a `ValueError` for some packages installed as `.egg`.
See https://github.com/tkhyn/sphinx180-automodule-bug

On Windows 7 or 10, Python 3.6.5 as the default python interpreter and from a fresh clean virtual environment:

```
git clone git@github.com:tkhyn/sphinx180-automodule-bug.git
cd sphinx180-automodule-bug
python bootstrap.py
bin\buildout
bin\sphinx-apidoc -o doc\_apidoc autodoc_bug
bin\sphinx-build doc doc\_build
```

#### Error logs / results
```
D:\dev\.env\venv\buildout\Scripts\python.exe D:/dev/sandbox/python/sphinx-1.8.0_bug/bin/sphinx-build-script.py doc doc\_build
Running Sphinx v1.8.0
loading pickled environment... done
updating environment: [] 0 added, 1 changed, 0 removed
reading sources... [100%] _apidoc/sphinx180_bug
Exception occurred:
  File "d:\dev\.env\buildout\eggs\sphinx-1.8.0-py3.6.egg\sphinx\util\__init__.py", line 318, in get_module_source
    eggpath, _ = re.split('(?<=\\.egg)/', filename)
ValueError: not enough values to unpack (expected 2, got 1)
The full traceback has been saved in C:\Users\Thomas\AppData\Local\Temp\sphinx-err-o4loxs60.log, if you want to report the issue to the developers.

Process finished with exit code 2
```

#### Expected results
As with Sphinx 1.7.8, no Exception raised and documentation generated normally

See https://github.com/tkhyn/sphinx180-automodule-bug

### Environment info
- OS: Win7, Win10, could not reproduce on Linux
- Python version: 3.6.5
- Sphinx version: 1.8.0

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5437

### DIFF
--- a/sphinx/util/__init__.py
+++ b/sphinx/util/__init__.py
@@ -314,8 +314,9 @@ def get_module_source(modname):
             filename += 'w'
     elif not (lfilename.endswith('.py') or lfilename.endswith('.pyw')):
         raise PycodeError('source is not a .py file: %r' % filename)
-    elif '.egg' in filename:
-        eggpath, _ = re.split('(?<=\\.egg)/', filename)
+    elif ('.egg' + os.path.sep) in filename:
+        pat = '(?<=\\.egg)' + re.escape(os.path.sep)
+        eggpath, _ = re.split(pat, filename, 1)
         if path.isfile(eggpath):
             return 'file', filename
 
--- a/CHANGES
+++ b/CHANGES
@@ -20,6 +20,7 @@ Bugs fixed
 * #5421: autodoc emits deprecation warning for :confval:`autodoc_default_flags`
 * #5422: lambda object causes PicklingError on storing environment
 * #5417: Sphinx fails to build with syntax error in Python 2.7.5
+* #5437: autodoc: crashed on modules importing eggs
 
 Testing
 --------


sphinx/util/__init__.py
```diff
@@ -315,7 +315,7 @@ def get_module_source(modname):
     elif not (lfilename.endswith('.py') or lfilename.endswith('.pyw')):
         raise PycodeError('source is not a .py file: %r' % filename)
     elif '.egg' in filename:
-        eggpath, _ = re.split('(?<=\\.egg)/', filename)
+        eggpath, _ = re.split('(?<=\\.egg)/', filename, 1)
```
tkhyn: This doesn't work (expanding a list of size 1 into 2 variables)!

I think you wanted to write `eggpath = re.split('(?<=\\.egg)/', filename, 1)[0]` (which I have tested and works fine)

tkhyn (changes requested):
Thanks for your promptness at looking into this!

It did not work as such, but the suggested change (see code comments) made it work ok.

Thanks again

sphinx/util/__init__.py
tk0miya: OMG. I also have to rewrite elif-clause like `elif '.egg/' in filename`...

sphinx/util/__init__.py
```diff
@@ -314,8 +314,8 @@ def get_module_source(modname):
             filename += 'w'
     elif not (lfilename.endswith('.py') or lfilename.endswith('.pyw')):
         raise PycodeError('source is not a .py file: %r' % filename)
-    elif '.egg' in filename:
-        eggpath, _ = re.split('(?<=\\.egg)/', filename)
+    elif '.egg/' in filename:
+        eggpath, _ = re.split('(?<=\\.egg)/', filename, 1)
```
tkhyn: Thanks again. Indeed that's more sensible! That works in my case for the cma module, but I think the following improvement would be more cross-platform:

```python
    elif '.egg' + os.path.sep in filename:
        eggpath, _ = re.split('(?<=\\.egg)' + re.escape(os.path.sep), filename, 1)
```

tkhyn (changes requested):

tk0miya:
Thank you for reviewing. Merged.

tkhyn:
> Thank you for reviewing. Merged.

You're welcome, thanks for fixing it so quickly!

PR #5455, opened 2018-09-18
Fix #5453 and move LaTeX ``\pagestyle`` to template

### LINKED ISSUE #5453: PDF builds of 'howto' documents have no page numbers
With the LaTeX docclass set to ``'howto'`` (which, for non-Japanese, uses ``article`` class), the PDF has no page numbers.

But the bookmarks, the table of contents and the indices all refer to page numbers.

I tested Sphinx v1.0, v1.2 (a commit was merged at 1.2, see #995 discussed below), v1.3.1 ...

#### Expected results

In so far as the PDF bookmarks, the table of contents, and the Indices refer to page numbers, page numbers should be visible, not hidden! Hyperlinked page numbers from Index work, so viewing on screen is ok, but printed version on paper is hardly usable.

(the page numbering is ok, the only problem is that the footer containing it is using an "empty" style hence nothing shows)

#### Relates

#995 seems to indicate that page numbers showed, but using roman numbers, not arabic.

However whether before or after https://github.com/sphinx-doc/sphinx/commit/44d1f1e952bf5a6b195987c4329c119b8487e56b (which was merged to 1.2) page numbers do not show at bottom of page in my testing. They show for the index, not for body of document. The merged commit indeed has the effect that the page numbers show ok in TOC and bookmarks and general index, but still pages of document have no page numbers until index.

### Environment info
- OS: Mac
- Python version: 3.6.5 (I used python 2.7 for testing 1.1, 1.2)
- Sphinx version: all versions?
- TeXLive 2018

jfbu:
Bugfix: #5453 

No change for 'manual' class (or Japanese documents), but the `\sphinxtableofcontents` for 'manual' has been modified: the `\pagestyle` commands therein have been moved to LaTeX template.

Overall this changes nothing to 'manual' documents except for those users who have customized `\sphinxtableofcontents` in their own way and perhaps don't want the `\pagestyle` now in LaTeX templates: I expect very few if any project have done that, and in that case the user only needs to use custom LaTeX template to suppress the `\pagestyle` inserted from it.

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -6,7 +6,7 @@
 %
 
 \NeedsTeXFormat{LaTeX2e}[1995/12/01]
-\ProvidesPackage{sphinx}[2018/07/18 v1.8 LaTeX package (Sphinx markup)]
+\ProvidesPackage{sphinx}[2018/09/18 v1.8.1 LaTeX package (Sphinx markup)]
 
 % provides \ltx@ifundefined
 % (many packages load ltxcmds: graphicx does for pdftex and lualatex but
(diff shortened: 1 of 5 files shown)


sphinx/texinputs/sphinx.sty
```diff
@@ -517,17 +533,24 @@
     % define chaptermark with \@chappos when \@chappos is available for Japanese
     \ltx@ifundefined{@chappos}{}
       {\def\chaptermark##1{\markboth{\@chapapp\space\thechapter\space\@chappos\space ##1}{}}}
-  }
+    }
   % Update the plain style so we get the page number & footer line,
   % but not a chapter or section title.  This is to keep the first
   % page of a chapter and the blank page between chapters `clean.'
```
jfbu: I do not understand the "blank page between chapters" bit from this legacy comment. In report class (used by 'manual'), the `\chapter` command does indeed issue `\thispagestyle{plain}` after having done `\cleardoublepage`,  but there is no modification of the header/footer style for the last page of previous chapter, if the latter has no contents. Only the first page of new chapter is re-styled via the `\thispagestyle{plain}` arising from `\chapter`.

Thus I should probably remove this old comment which I believe to be misleading.

jfbu:
Codecov seems to have gone beserk (perhaps due to me force pushing twice shortly after having already made PR).

If there is one more Codecov report soon I will close and reopen to try to stop the flood...

tk0miya (approved):
+1

jfbu:
Thanks for review @tk0miya. I will push a commit to remove the legacy comment about "keeping clean the blank page between chapters" which I find misleading and to add a CHANGES entry, then I will merge.

PR #5465, opened 2018-09-21
Fix #5454: latex: Index has disappeared from PDF for Japanese documents

### LINKED ISSUE #5454: Index has disappeared from PDF for Japanese documents at v1.8.0
### Problem

Japanese projects have lost index in PDF output since Sphinx v1.8.0.

Build PDF for this

```


Section
=======

:index:`aaa`

:index:`目次`
```

with `language` set to `'ja'`

The PDF contains no index.

When built with Sphinx v1.7.9 there is index in PDF.

First "bad" commit appears to be ba97fd4f8d0f6c82ba44f2298093a040bde28c51

### Environment info
- OS: Mac
- Python version: 3.6.5
- Sphinx version: v1.8.0
- TeXLive 2018

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5454

### DIFF
--- a/sphinx/texinputs/latexmkjarc
+++ b/sphinx/texinputs/latexmkjarc
@@ -5,11 +5,9 @@ sub mendex {
   my ($source, $basename, $destination) = @_;
   my $dictfile = $basename . ".dic";
   unlink($destination);
-  if (-f $dictfile) {
-    system("mendex", "-U", "-f", "-d", $dictfile, "-s", "python.ist", $source);
-    if ($? > 0) {
-      print("mendex exited with error code $? (ignored)\n");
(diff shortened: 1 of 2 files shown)


sphinx/texinputs/latexmkjarc
```diff
@@ -5,11 +5,9 @@ sub mendex {
   my ($source, $basename, $destination) = @_;
   my $dictfile = $basename . ".dic";
   unlink($destination);
-  if (-f $dictfile) {
```
tk0miya, line 8: I found nobody makes dictfile on default sequence...

PR #5474, opened 2018-09-23
Adds summary statistics to coverage report

jorgecarleitao:
Subject: Adds a table with a summary of the coverage report:

```
+-----------------------+----------+--------------+
| Module                | Coverage | Undocumented |
+=======================+==========+==============+
| schemaflow.exceptions | 100.00%  | 0            |
+-----------------------+----------+--------------+
| schemaflow.pipe       | 83.33%   | 1            |
+-----------------------+----------+--------------+
| schemaflow.pipeline   | 100.00%  | 0            |
+-----------------------+----------+--------------+
| schemaflow.types      | 70.00%   | 3            |
+-----------------------+----------+--------------+
| TOTAL                 | 82.61%   | 4            |
+-----------------------+----------+--------------+
```

### Feature or Bugfix
- Feature

### Purpose

The current implementation of `ext.coverage` outputs which methods/classes/functions are documented. This PR adds a short summary of this report in terms of `documented objects / total number of objects`, both per module and total.

The purpose of this is to support a currently not mainstream but relevant use-case: a coverage report on the number of objects that are documented.

By having the statistics on the report or on the stdout, a regex expression can capture the coverage percentage (`re.search(r'TOTAL.*?([0-9.]{4,6}\%)', d).group(1)` in Python regex) and use it e.g. as another CI report, in a Badge, etc.

The current version of this PR does not do anything by default to not break backward compatibility. Two options were added to the configuration to allow a table to be outputted to the report and/or to stdout.

There are still steps missing:

1. simplify code
2. unit-test 
3. documentation

however, I have the following questions before proceeding:

1. is this something that you would like to have?
2. should we make the coverage report to contain the statistics by default?
3. should we make the coverage to output the table to stdout by default?

My hypothesis is that the answer is affirmative for all 3 cases: the most common usage of `coverage` is to know how the coverage of docs is and a nice table as output seems a good output for the command. However, I am not sure about backward compatibility.


AA-Turner:
Probably need to bump the env version here so that old pickle files are ignored.

A

sphinx/ext/coverage.py
```diff
@@ -310,6 +388,8 @@ def setup(app: Sphinx) -> dict[str, Any]:
     app.add_config_value('coverage_c_regexes', {}, False)
     app.add_config_value('coverage_ignore_c_items', {}, False)
     app.add_config_value('coverage_write_headline', True, False)
+    app.add_config_value('coverage_statistics_to_report', True, False, (bool,))
+    app.add_config_value('coverage_statistics_to_stdout', False, False, (bool,))
```
AA-Turner:
```suggestion
    app.add_config_value('coverage_statistics_to_stdout', True, False, (bool,))
```

AA-Turner:
Thank you @jorgecarleitao!

A

PR #5475, opened 2018-09-23
Deprecate Python2 compat shim sphinx.util.osutil.walk()

jdufresne:
Code should use `os.walk()` instead, which works with either `str` or `bytes`. All internal uses use `str`.

### DIFF
--- a/sphinx/util/osutil.py
+++ b/sphinx/util/osutil.py
@@ -90,39 +90,12 @@ def ensuredir(path):
             raise
 
 
-# This function is same as os.walk of Python2.7 except a customization
-# that check UnicodeError.
-# The customization obstacle to replace the function with the os.walk.
 def walk(top, topdown=True, followlinks=False):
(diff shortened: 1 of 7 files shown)


sphinx/util/osutil.py
```diff
@@ -98,6 +98,10 @@ def walk(top, topdown=True, followlinks=False):
     """Backport of os.walk from 2.6, where the *followlinks* argument was
     added.
     """
+    warnings.warn('sphinx.util.osutil.walk() is deprecated for removal. '
+                  'Please use os.walk() instead.',
+                  RemovedInSphinx40Warning)
+
     names = os.listdir(top)
 
     dirs, nondirs = [], []
```
stephenfin: Could we just replace all of this with a call to `os.walk`? It should be a separate change but in theory, they do the exact same thing now, right? That means we could delete any tests that might exist for this too.

jdufresne:
Good idea. Done in the latest revision. The function does not have any tests to delete.

tk0miya:
Merged. Thanks!

PR #5477, opened 2018-09-23
Correct installation docs about supported Python versions

jdufresne:
Since commit 9412bd76b7f5fdf0adc231ef8130e45bda130164, older Python versions have been dropped.


tk0miya:
Thanks!

PR #5478, opened 2018-09-23
Remove uses of six.moves that did not cause any type errors

jdufresne:
Removal of the remaining imports may require passing `--python-version 3.5` to the mypy command.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -13,6 +13,7 @@
 from __future__ import print_function
 
 import os
+import pickle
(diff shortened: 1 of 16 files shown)


tk0miya:
Thanks!

>Removal of the remaining imports may require passing --python-version 3.5 to the mypy command.

To do that, we have to annotate to docutils. Because typehints for docutils has been broken ...
Now @cocoatomo and me are working for that.
So please wait for a while.

PR #5485, opened 2018-09-24
Remove unnecessary __ne__ definitions

jdufresne:
In Python 3, `__ne__` defaults to the inverse of `__eq__`.

From https://docs.python.org/3/reference/datamodel.html#object.__ne__

> By default, `__ne__()` delegates to `__eq__()` and inverts the result unless it is `NotImplemented`.


tk0miya:
:-)

PR #5500, opened 2018-09-30
#5460 Make JS refactoring compatible with old templates

### LINKED ISSUE #5460: Search does not work for docs built with 1.8.0
Subject: Docs built with sphinx 1.8.0 have broken search

### Problem
- Search returns no results for all queries. When I build using 1.7.9 everything works fine. 

 - build docs with sphinx 1.8.0
 - try to search the docs using the search textbox
 - no results ever show

#### Expected results
 Search works

### Environment info
- OS: Fedora 28
- Python version: 3.6
- Sphinx version: 1.8.0

TimKam:
Subject: Make JS refactoring compatible with old templates

### Feature or Bugfix
- Bugfix

### Purpose
- Fixes #5460, prevents old templates from breaking search function

### Detail
This is my attempt of fixing this issue.
Feedback on implementation details are appreciated.
* Inject ``documentation_options.js`` script as part of HTML build
* Set attributes of the script accordingly
* Remove script reference from basic template (would be duplication)
* Add fallback for URL_ROOT variable
* Tested with Read the Docs template and Sphinx template

### Relates
- #5460

### DIFF
--- a/sphinx/themes/basic/layout.html
+++ b/sphinx/themes/basic/layout.html
@@ -87,7 +87,6 @@
 {%- endmacro %}
 
 {%- macro script() %}
-    <script type="text/javascript" id="documentation_options" data-url_root="{{ pathto('', 1) }}" src="{{ pathto('_static/documentation_options.js', 1) }}"></script>
     {%- for js in script_files %}
     {{ js_tag(js) }}
     {%- endfor %}
--- a/sphinx/builders/html.py
+++ b/sphinx/builders/html.py
@@ -1564,11 +1564,22 @@ def convert_html_js_files(app, config):
     config.html_js_files = html_js_files  # type: ignore
 
 
+def add_documentation_options_js(app, config):
+    # type: (Sphinx, Config) -> None
+    """Adds path of ``add_documentation_options.js_t`` to the ``html_js_files``
+    variable. This avoids issues with older templates that use a custom
+    layout file."""
+    config.html_js_files.append('documentation_options.js')
+
+
 def setup_js_tag_helper(app, pagename, templatexname, context, doctree):
     # type: (Sphinx, unicode, unicode, Dict, nodes.Node) -> None
-    """Set up js_tag() template helper.
+    """Set up js_tag() template helper and add special handling for
+    ``add_documentation_options``.
 
-    .. note:: This set up function is added to keep compatibility with webhelper.
+    .. note:: This set up function is added to keep compatibility with
+    webhelper, as well as with older templates that use a custom
+    layout file.
     """
     pathto = context.get('pathto')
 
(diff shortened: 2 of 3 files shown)


tk0miya (changes requested):
This would not work with old themes. Because they has not used `js_tag()` function. It was added in 1.8.

To support them, we need to modify #5207. They did not read `documentation_options.js`.

TimKam:
Thanks for reviewing, @tk0miya. 

The idea of the change is the following:
* The ``documentation_options`` script is added in _any case_ to the HTML file, on ``config-inited``.
* If a "new" template is used, the ``setup_js_tag_helper`` function adds the attributes to the ``documentation_options`` script tag that are needed to define the URL root.
* If an "old" template is used, the attributes are not added. However, the URL root is determined _within the old templates_, and caught by the following change in ``documentation_options.js_t``:

```
URL_ROOT: document.getElementById("documentation_options") ?
        document.getElementById("documentation_options").getAttribute('data-url_root') :
        DOCUMENTATION_OPTIONS.URL_ROOT,
```

Or am I missing something?

mitya57:
I just tested this commit with sphinx-rtd-theme 0.4.1 tag, and it seems to work.

Maybe you can change the base branch from master to 1.8 (and rebase accordingly)?

TimKam:
Great, I'll wait for feedback from @tk0miya and rebase then.

TimKam:
@tk0miya Sorry for pinging you again, but could you take a look at my justification above?

tk0miya:
We have four generations for searching mechanism:

1. `DOCUMENTATION_OPTIONS` is defined in layout.html directly (I guess this might be not used)
2. `DOCUMENTATION_OPTIONS` is defined in `documentation_options.js`. But `<script>` tag does not have any ID
3. "2." and `<script>` tag has `documentation_options` ID
4. "3." and support `js_tag()` to generate `<script>` tag. In addition, some codes were moved to `documentation_options.js` from `searchtools.js`.

I think current `documentation_options.js` contains a stemmer, Scorer and word-splitter. But this does not write-back them to `serachtools.js`. So this would not work for non-English document.
(There are some fallback functions. So it would not be crashed.)

I think reverting #5207 would fix the situation.

sphinx/themes/basic/layout.html
```diff
@@ -87,7 +87,6 @@ <h3>{{ _('Navigation') }}</h3>
 {%- endmacro %}
 
 {%- macro script() %}
-    <script type="text/javascript" id="documentation_options" data-url_root="{{ pathto('', 1) }}" src="{{ pathto('_static/documentation_options.js', 1) }}"></script>
```
tk0miya, line 90: As commented above, 3rd generation themes does not use `js_tag()` helper. So this would not be moved.

In addition, to support this tag by `js_tag()`, we have to give a special code for this like you modified. I feel it is too much.

mitya57:
How about moving stemmer, scorer, word-splitter and SEARCH_LANGUAGE_STOP_WORDS to a separate JS file (e.g. `language_data.js`), which will then be in `script_files`? This way it will be included even for custom, old-style templates.

@tk0miya Will you accept a pull request implementing such a change?

tk0miya:
+1 for @mitya57 
From my understanding, they were moved for testing searchtools.js (see #5207).
So to create a new file might be a better solution for this problem.

mitya57:
@tk0miya Submitted as #5590.

tk0miya:
Merged #5590 instead. So I'm closing this now.

PR #5504, opened 2018-10-02
MathJax config: Allow single $ delimiters, only apply to CSS "math" class

mgeier:
I'm not quite sure if that's the right thing to do. This should be understood as a question, but I made this PR so that it's easy to test the suggested behavior.

### Feature or Bugfix

Something inbetween.

### Purpose

* Support single `$` as math delimiters (probably only relevant in `:nowrap:` math directives).
* Reduce MathJax processing to CSS `math` class.

### Detail

This is some example reST code to check this out:

```rst
Dollars in normal text: $a^2$.
A single dollar sign: $.
An escaped dollar sign: \$.

Inline math: :math:`a^2`.

Normal math directive:

.. math::

    a^2

"nowrap" math with ``$``:

.. math::
    :nowrap:

    $a^2$

"nowrap" with normal text:

.. math::
    :nowrap:

    This is a \LaTeX{} equation: $a^2 + b^2 = c^2$

    This is a dollar sign: \$.
```

Note that this PR mainly fixes the `:nowrap:` cases above. This might not be important for most "normal" users. But when auto-generating reST files which have arbitrary math markup (out of my control), this is the only way I found to make it work.

### DIFF
--- a/doc/usage/extensions/math.rst
+++ b/doc/usage/extensions/math.rst
@@ -181,7 +181,8 @@ Sphinx.
            'jax': ['input/TeX', 'output/HTML-CSS'],
        }
 
-   The default is empty (not configured).
+   The default is to enable single ``$`` signs as delimiters (which are disabled
+   in the MathJax default settings).
 
 .. _Using in-line configuration options: https://docs.mathjax.org/en/latest/configuration.html#using-in-line-configuration-options
 
--- a/doc/usage/extensions/math.rst
+++ b/doc/usage/extensions/math.rst
@@ -182,7 +182,7 @@ Sphinx.
        }
 
    The default is to enable single ``$`` signs as delimiters (which are disabled
-   in the MathJax default settings).
+   in the MathJax default settings) and to only operate on the "math" CSS class.
 
 .. _Using in-line configuration options: https://docs.mathjax.org/en/latest/configuration.html#using-in-line-configuration-options
 
--- a/sphinx/ext/mathjax.py
+++ b/sphinx/ext/mathjax.py
@@ -115,7 +115,7 @@ def setup(app):
                              'tex2jax': {
                                  'inlineMath': [['$', '$'], ['\\(', '\\)']],
                                  'processEscapes': True,
-                                 'ignoreClass': '.*',
+                                 'ignoreClass': 'document',
                                  'processClass': 'math',
                              },
                          }, 'html')
--- a/sphinx/ext/mathjax.py
+++ b/sphinx/ext/mathjax.py
@@ -115,6 +115,8 @@ def setup(app):
                              'tex2jax': {
                                  'inlineMath': [['$', '$'], ['\\(', '\\)']],
                                  'processEscapes': True,
+                                 'ignoreClass': '.*',
+                                 'processClass': 'math',
                              },
                          }, 'html')
     app.connect('env-check-consistency', install_mathjax)
--- a/sphinx/ext/mathjax.py
+++ b/sphinx/ext/mathjax.py
@@ -110,7 +110,13 @@ def setup(app):
     app.add_config_value('mathjax_options', {}, 'html')
     app.add_config_value('mathjax_inline', [r'\(', r'\)'], 'html')
     app.add_config_value('mathjax_display', [r'\[', r'\]'], 'html')
-    app.add_config_value('mathjax_config', None, 'html')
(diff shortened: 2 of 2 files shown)


jfbu:
```
.. math::
    :nowrap:

    This is a \LaTeX{} equation: $a^2 + b^2 = c^2$

    This is a dollar sign: \$.
```

This modifies expected behaviour of math directive. It should not be used for arbitrary text. The `:nowrap:` intent is only to give Sphinx user possibility to decide of math display environment of his/her choice. The contents remained constrained to be LaTeX math not text.

At least that's how I have always considered it.

tk0miya:
-0: I also feel strange because `math` directive represents "block" equations. So using `$` is strange. I know it is rendered well in both imgmath extension and LaTeX builder. But it is not promised behavior to me.

tk0miya:
In above comment, I said "block" equations. But "display" equations would be much better to describe it.

jfbu:
The example I quoted from @mgeier uses non-math paragraphs, in LaTeX encoding, inside `math` directive. This feels wrong to me in using the `math` directive. Besides for html+MathJax, MathJax is not LaTeX but a Javascript library partially implementing TeX mark-up particularly for math. So arbitrary LaTeX mark-up can not be allowed if it is to be rendered also by html + MathJax.

mgeier:
I agree that it is "strange" and it even "feels wrong" to write such `math` directives by hand. But in my use case (showing Jupyter notebook code cells with the `nbsphinx` extension) something like this would be necessary to get the same behavior as in the notebook application.

Is it possible to add this MathJax config from an extension without interfering when users have their own `mathjax_config` settings?

But apart from "being strange" and "feeling wrong", is there a real downside to my suggestion?

And what about the second part of this PR, about restricting MathJax to `math` CSS classes?

I think it isn't a huge problem, but it would be more "correct" to enable MathJax only in `math` classes, right?

This would be an admittedly quite unrealistic case where MathJax would be wrongly used in normal text:

```rst
I love $$ Money $$!
```

Interestingly it is still processed by MathJax if the dollars are escaped:

```rst
I love \$\$ Money \$\$!
```

mgeier:
Some background on why I need support for single `$` signs: IPython just merged a PR that uses `$\displaystyle ...$` for math output (https://github.com/ipython/ipython/pull/11357) and I hope SymPy and others will follow suit. The reason for that change is left-aligning math in code cells of Jupyter notebooks when they are converted to LaTeX (via `nbconvert` or `nbsphinx` or other tools).

The reason for this PR here is to display this properly in Sphinx's HTML output when using Jupyter notebooks as source files (using the `nbsphinx` extension).

mgeier:
I've just re-based my commits for your merging convenience.

mgeier:
Just a little update: I've found out that using `'.*'` as `ignoreClass` is a bit restrictive, because it disables MathJax for nested elements (without the `math` class) within an element with the `math` class. I've changed this in c07938ef6fdd6a664dd8852056bd760630519045.

Also, I've included the change directly into `nbsphinx`:

https://github.com/spatialaudio/nbsphinx/blob/68b709e72c22cd194470155d5db0b0dcbee19a6b/src/nbsphinx.py#L1756-L1766

... so I don't really need this PR anymore.

Feel free to close it!

AA-Turner:
@mgeier if possible, please could you add a CHANGES entry to this (& a potentially a test, though that might not be as easy, not a requirement) -- if this is useful to people beyond ``nbsphinx`` I'd like to merge it.

A

AA-Turner:
@mgeier is this PR still needed/relevant? We can try and bring it up to date if so, or close it if not.

mgeier:
As mentioned above (https://github.com/sphinx-doc/sphinx/pull/5504#issuecomment-446290233), I have worked around this in `nbsphinx` (see https://github.com/spatialaudio/nbsphinx/blob/215bf0d9b2108c739a224736d8763dffc91f3479/src/nbsphinx/__init__.py#L1549-L1567), so I don't need this anymore.

There is still a potential problem when people use double dollar signs (as described above: https://github.com/sphinx-doc/sphinx/pull/5504#issuecomment-428171765), but since nobody seems to have reported this, it might not be a thing that happens in practice.

AA-Turner:
OK, closing, and we can revisit the double-dollar-sign issue if it is brought up again. Thanks!

A

PR #5508, opened 2018-10-04
directive-code: do not force linenos value on run

jdknight:
Now that `highlightlang` directive is deprecated \[1\], the `linenothreshold` option is to be used via the `highlight` directive. The `highlight` directive will walk-through literal blocks to apply a `linenos` value if:

1) The literal block has not been explicitly configured with the `linenos` option.
2) If there is enough content (when comparing literal block's line count to `linenothreshold`) that `linenos` should be explicitly enabled or disabled \[2\].

While the `hightlight` directive should be able to explicitly define if a literal block needs to enable line numbers, the logic is always ignored since the code block directive already configures `linenos` when checking for the `linenos` and `lineno-start` options on the node \[3\]. This effectively prevents `linenothreshold` from being used.

To allow `linenothreshold` to be used in literal blocks, this commit disables the explicit configuration `linenos` on a literal block instance when the `CodeBlock` directive is processed.

\[1\]: b35198d8475fe89aaf9a5224a9832fb394e73cca
\[2\]: https://github.com/sphinx-doc/sphinx/blob/v1.8.1/sphinx/transforms/post_transforms/code.py#L95-L97
\[3\]: https://github.com/sphinx-doc/sphinx/blob/v1.8.1/sphinx/directives/code.py#L156-L157

### DIFF
--- a/sphinx/directives/code.py
+++ b/sphinx/directives/code.py
@@ -154,8 +154,8 @@ class CodeBlock(SphinxDirective):
             code = '\n'.join(lines)
 
         literal = nodes.literal_block(code, code)  # type: nodes.Element
-        literal['linenos'] = 'linenos' in self.options or \
-                             'lineno-start' in self.options
+        if 'linenos' in self.options or 'lineno-start' in self.options:
+            literal['linenos'] = True
         literal['classes'] += self.options.get('class', [])
         if self.arguments:
             # highlight language specified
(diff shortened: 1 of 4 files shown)


tk0miya (changes requested):
Sorry for late response

It seems tests failed. Could you check that please?
In addition, it would be nice if there are new testcases to confirm the behavior of `highlight` directive.

tk0miya (approved):

tk0miya:

PR #5509, opened 2018-10-05
Remove future imports

stephenfin:
These are not necessary in Python 3.

### DIFF
--- a/sphinx/__init__.py
+++ b/sphinx/__init__.py
@@ -12,8 +12,6 @@
 # Keep this file executable as-is in Python 3!
 # (Otherwise getting the version out of it from setup.py is impossible.)
 
-from __future__ import absolute_import
-
 import os
 import warnings
 from os import path
--- a/sphinx/builders/devhelp.py
+++ b/sphinx/builders/devhelp.py
@@ -10,7 +10,6 @@
     :copyright: Copyright 2007-2018 by the Sphinx team, see AUTHORS.
     :license: BSD, see LICENSE for details.
 """
-from __future__ import absolute_import
 
 import gzip
 import re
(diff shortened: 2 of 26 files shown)


tk0miya (changes requested):
LGTM, but travis raises an error.
Could you check it please?

stephenfin:
> LGTM, but travis raises an error.
> Could you check it please?

I can't figure out what's happening here. I've simply inserted a TODO to investigate this in the future.

tk0miya:
OMG, I found the reason. We've used mypy under py2 mode because it typeshed contains broken typehints for docutils. As a result, it does not support `file` parameter for `print()` function.

BTW, I feel `f.write()` is better to write data to a file. How about replace `print()` calls by it?

stephenfin:
> OMG, I found the reason. We've used mypy under py2 mode because it typeshed contains broken typehints for docutils. As a result, it does not support `file` parameter for `print()` function.
> 
> BTW, I feel `f.write()` is better to write data to a file. How about replace `print()` calls by it?

Yes, that's one problem and I can fix this. There is another problem in `tests/roots/test-ext-viewcode-find/not_a_package/__init__.py` though.

https://github.com/sphinx-doc/sphinx/blob/e781747f618e25a533da20185374640720e6ee72/tests/roots/test-ext-viewcode-find/not_a_package/__init__.py#L1-L2

Do you have any idea why this is failing?

jdufresne:
Now that mypy is running with Python3, this could be rebased and completed.

stephenfin:
> Now that mypy is running with Python3, this could be rebased and completed.

Done. I still don't know why I need to keep the future import in that one test case, but it's otherwise working as expected.

tk0miya (approved):

tk0miya:
Thank you for update. Let's go ahead. And I'll take a look about last piece later.

tk0miya:
I post #5824 for it.

PR #5512, opened 2018-10-07
latex: Separate message catalogs to sphinxmessage.sty

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To make main latex output simple, I'd like to move message resources to separated file.
- This is first step of separation. I will work for other messages in other work if agreed.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,8 @@ Incompatible changes
 * Drop docutils 0.11 support
 * The default setting for :confval:`master_doc` is changed to ``'index'`` which
   has been longly used as default of sphinx-quickstart.
+* LaTeX: Move message resources to ``sphinxmessage.sty``
+* LaTeX: Stop using ``\captions<lang>`` macro for some labels
 
 Deprecated
 ----------
@@ -22,6 +24,7 @@ Deprecated
 * ``sphinx.ext.doctest.doctest_encode()``
 * ``sphinx.testing.util.remove_unicode_literal()``
 * ``sphinx.util.osutil.walk()``
+* ``translatablestrings`` variable for LaTeX template
 
 For more details, see `deprecation APIs list
 <http://www.sphinx-doc.org/en/master/extdev/index.html#deprecated-apis>`_
--- a/sphinx/builders/latex/__init__.py
+++ b/sphinx/builders/latex/__init__.py
@@ -34,6 +34,7 @@ from sphinx.util.docutils import SphinxFileOutput, new_document
 from sphinx.util.fileutil import copy_asset_file
 from sphinx.util.nodes import inline_all_toctrees
 from sphinx.util.osutil import SEP, make_filename
+from sphinx.util.template import LaTeXRenderer
 from sphinx.writers.latex import (
     ADDITIONAL_SETTINGS, DEFAULT_SETTINGS, LaTeXWriter, LaTeXTranslator
 )
@@ -327,6 +328,7 @@ class LaTeXBuilder(Builder):
     def finish(self):
         # type: () -> None
         self.copy_image_files()
+        self.write_message_catalog()
 
         # copy TeX support files from texinputs
         # configure usage of xindy (impacts Makefile and latexmkrc)
@@ -389,6 +391,11 @@ class LaTeXBuilder(Builder):
                 except Exception as err:
                     logger.warning(__('cannot copy image file %r: %s'),
                                    path.join(self.srcdir, src), err)
+
+    def write_message_catalog(self):
+        # type: () -> None
+        filename = path.join(package_dir, 'templates', 'latex', 'sphinxmessage.sty_t')
+        copy_asset_file(filename, self.outdir, context={}, renderer=LaTeXRenderer())
 
 
 def validate_config_values(app, config):
--- a/sphinx/templates/latex/latex.tex_t
+++ b/sphinx/templates/latex/latex.tex_t
@@ -29,6 +29,7 @@
 <%= fontpkg %>
 <%= fncychap %>
 \usepackage<%= sphinxpkgoptions %>{sphinx}
+\input{sphinxmessage.sty}
 <%= sphinxsetup %>
 <%= fvset %>
 <%= geometry %>
@@ -36,7 +37,6 @@
 <%= hyperref %>
 <%= contentsname %>
 <%= numfig_format %>
-<%= translatablestrings %>
 <%= pageautorefname %>
 <%= tocdepth %>
 <%= secnumdepth %>
--- a/sphinx/templates/latex/latex.tex_t
+++ b/sphinx/templates/latex/latex.tex_t
@@ -29,7 +29,6 @@
 <%= fontpkg %>
 <%= fncychap %>
 \usepackage<%= sphinxpkgoptions %>{sphinx}
-\input{sphinxmessage.sty}
 <%= sphinxsetup %>
 <%= fvset %>
 <%= geometry %>
@@ -37,6 +36,7 @@
 <%= hyperref %>
 <%= contentsname %>
 <%= numfig_format %>
+\input{sphinxmessage.sty}
 <%= pageautorefname %>
 <%= tocdepth %>
 <%= secnumdepth %>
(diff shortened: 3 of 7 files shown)


sphinx/templates/latex/sphinxmessage.sty_t
```diff
@@ -0,0 +1,6 @@
+<%= babel(renewcommand('\literalblockcontinuedname', escape(_('continued from previous page')))) %>
```
tk0miya: It would be much better if we can move `babel(...)` to TeX macro...

sphinx/builders/latex/__init__.py
```diff
@@ -390,6 +404,27 @@ def copy_image_files(self):
                     logger.warning(__('cannot copy image file %r: %s'),
                                    path.join(self.srcdir, src), err)
 
+    def write_message_catalog(self):
+        # type: () -> None
+        def babel(definition):
```
tk0miya: I think `babel()` is not good naming. I think `addtocaptions()` might be better.

sphinx/templates/latex/latex.tex_t
```diff
@@ -29,6 +29,7 @@
 <%= fontpkg %>
 <%= fncychap %>
 \usepackage<%= sphinxpkgoptions %>{sphinx}
+\usepackage{sphinxmessage}
```
jfbu: As it would be cumbersome to add a `\ProvidesPackage` to that file, it is simpler to use `\input{sphinxmessage.sty}` or `\makeatletter\input{sphinxmessage.sty}\makeatother` rather if the file contains `@` letter in macros.

sphinx/templates/latex/sphinxmessage.sty_t
jfbu: not sure I understand what you mean here

jfbu: what about `sphinxcaptions.sty` as filename ? or `sphinxlocalizedstrings.sty`? or perhaps "message" is familiar from gettext etc...

jfbu (approved):
LGTM with nits (mainly, not using `\usepackage{sphinxmessage}` but rather `\input{message.sty}` with perhaps `\makeatletter...\makeatother` either in the file or in the 
preamble.

In reviewing this I realized that currently, Sphinx abuses a bit the `\addto\captions<language>`. A priori this should be used for things LaTeX babel knows as captionnames such as `\contentsname`, `\figurename`, `\tablename` etc... Sphinx has in LaTeX writer https://github.com/sphinx-doc/sphinx/blob/ef97997070c8f75e8adb6022bb9e836ed89c10da/sphinx/writers/latex.py#L751-L758
which uses `\addto\extras<language>` with some `\def` rather than `\renewcommand`. It is used once for `\pageautorefname` which intervenes via https://github.com/sphinx-doc/sphinx/blob/ef97997070c8f75e8adb6022bb9e836ed89c10da/sphinx/writers/latex.py#L730-L732 
i.e. the `\autopageref` macro provided by `hyperref.sty` (automatically loaded by Sphinx).

I don't know if there are any difference between `\captions...` and `\extras...` with babel, except that polyglossia obeys `\addto\captions...` but has no use for `\extras...`. These things are designed to allow multi-lingual documents. One must use the `\addto\captions...` for `\contentsname`, `\figurename`,  etc... else user definitions in preamble will get overwritten by babel package. But for all the custom `\sphinx...` prefixed macros, it did not make much sense as Sphinx does not support multilingual documents. We could as well have used `\def\sphinx...name{...}` directly in latex document preamble. But perhaps in future Sphinx will have some multilingual support, then it makes sense to use the `\addto\extras...` for those macros a priori unknown to babel.

Anyway if we move to a template that the user can overwrite or remove this forces also that sphinx.sty already contains at least some defaults, so `\renewcommand` can be used all the time.Things are not yet all moved to the sphinxmessage.sty file, in part

Will `\figurename` etc...:
https://github.com/sphinx-doc/sphinx/blob/ef97997070c8f75e8adb6022bb9e836ed89c10da/sphinx/writers/latex.py#L762
also move to sphinxmessage.sty

sphinx/builders/latex/__init__.py
jfbu: +1 for `addtocaptions()`. Or `addtomessages()`?

sphinx/templates/latex/sphinxmessage.sty_t
tk0miya: ah, sorry. I think representing this code by LaTeX is better and simple:
```
\addtocaptions{\renewcommand{\literalblockcontinuedname}{<%= _('continued from previous page' | e %>}}
```

jfbu: problem is that it should be for example `\addto\captionsfrench{...}` and if neither babel nor polyglossia is used one is not sure `\captions<languagename>` exists, so then one does directly `\renewcommand{\literalblockcontinuedname}`.

jfbu: as Sphinx does not support multilingual documents, we could forget completely the business about `\addto\captions...`. Directly do `\renewcommand`. The `\addto\captions...` are only mandatory for things like `\figurename`.

tk0miya: That's good news. We've not supported multilingual documents. So it would not be needed at least now.
+1 for removing :-)

jfbu: +1 also for directly using `\renewcommand`'s in the message file for all those `\sphinx...name` macros, and things like `\autopagerefname`. Here is the list of things needing `\addto\captions...` when babel (or I guess polyglossia) is loaded:

```latex
\prefacename\refname\abstractname\bibname\chaptername\appendixname
\contentsname\listfigurename\listtablename\indexname\figurename
\tablename\partname\enclname\ccname\headtoname\pagename\seename
\alsoname\proofname\glossaryname
```

tk0miya:
Updated.

jfbu (approved):
LGTM! much simplified...

CHANGES
```diff
@@ -13,6 +13,8 @@ Incompatible changes
 * Drop docutils 0.11 support
 * The default setting for :confval:`master_doc` is changed to ``'index'`` which
   has been longly used as default of sphinx-quickstart.
+* LaTeX: Move message resources to ``sphinxmessage.sty``
+* LaTeX: Stop to use ``\captions<lang>`` macro for some labels
```
jfbu: sorry for late but `Stop using` is better than `Stop to use`

jfbu:
the line at https://github.com/sphinx-doc/sphinx/pull/5512/files#diff-0aa57c753d7666f812251beecbf9ea4fR40 should be removed from latex template

jfbu (commented):
Sorry for belated remarks:

- the line at https://github.com/sphinx-doc/sphinx/pull/5512/files#diff-0aa57c753d7666f812251beecbf9ea4fR40 should be removed from latex template
-  and `'translatablestrings'` should be removed from `DEFAULT_SETTINGS` in sphinx.writers.latex

tk0miya:
Good catch!

sphinx/templates/latex/latex.tex_t
```diff
@@ -29,14 +29,14 @@
...
 \usepackage<%= sphinxpkgoptions %>{sphinx}
+\input{sphinxmessage.sty}
 <%= sphinxsetup %>
 <%= fvset %>
 <%= geometry %>
 <%= usepackages %>
 <%= hyperref %>
 <%= contentsname %>
 <%= numfig_format %>
-<%= translatablestrings %>
```
jfbu, line 39: perhaps move the `\input{sphinxmessage.sty}` to this location? (no strong optinion)

jfbu (approved):
LGTM

sphinx/templates/latex/latex.tex_t
tk0miya: I also don't have strong opinion. So let's move.

sphinx/templates/latex/sphinxmessage.sty_t
```diff
@@ -0,0 +1,5 @@
+\renewcommand{\literalblockcontinuedname}{<%= _('continued from previous page') | e %>}
```
tk0miya, line 1: I'll put a header in next PR.

tk0miya:
Merged. @jfbu, Thank you for reviewing!

PR #5513, opened 2018-10-07
Fix #3704: (latex) wrong ``\label`` positioning for figures with a legend

### LINKED ISSUE #3704: LaTeX: the ``\label`` in a figure should be immediately after ``\caption``, not after legend.
Subject: a [figure](http://docutils.sourceforge.net/docs/ref/rst/directives.html#figure) may have a legend. This consists of arbitrary body elements after the caption.

Currently the LaTeX writer inserts the ``\label`` right before the end of the ``figure`` environment (or ``wrapfigure``,  ``sphinxfigure-in-table``). But, theoretically, the legend contents could contain for example a table having itself a caption. Then the ``\label`` inserted after it will result to document hyperlinks inserted by Sphinx to point to the table in the legend not to the image at top of figure environment.

Proof of concept:

capture d ecran 2017-05-04 a 10 39 48

edit: the screenshot was taken with #3702 applied. Without it the very last words ``caption code`` are slightly shifted down, because the interline spacing is not correct for the last paragraph.

Source:
```rst

.. figure:: more.png
   :name: afigure

   This is the caption.

   This is some legend which contains a table.

   Here is some random extra paragraph for filling up.

   Again some random paragraph and now the table:

   .. table:: A table in a legend of a figure
      :name: atable

      ===== ===== =====
      A     B     C
      ===== ===== =====
      a     b     c
      ===== ===== =====

   Notice also that the small font size does not apply to the table
   caption, due to some hard-coded ``\normalsize`` in LaTeX's caption code.

Now the link to the figure: :ref:`incorrect <afigure>`. It actually brings us
to the table.
```

- Sphinx version: 1.5.5

jfbu:
Figures may have a "legend" which may contain arbitrary contents, so `\label` should be located immediately after `\caption`, else hyperlink might get wrong target.

see #3704 for example of what was wrong

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Bugs fixed
 
 * #5490: latex: enumerated list causes a crash with recommonmark
 * #5492: sphinx-build fails to build docs w/ Python < 3.5.2
+* #3704: latex: wrong ``\label`` positioning for figures with a legend
 
 Testing
 --------
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1720,7 +1720,6 @@ class LaTeXTranslator(nodes.NodeVisitor):
 
     def visit_figure(self, node):
         # type: (nodes.Node) -> None
-        labels = self.hypertarget_to(node)
         if self.table:
             # TODO: support align option
             if 'width' in node:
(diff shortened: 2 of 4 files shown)


tk0miya (approved):
LGTM!

jfbu:
Thanks @tk0miya  for quick review! merging

Issue #5501, opened 2018-10-01
build should not unconditionally rewrite pickle file


eqvinox:
Starting as of 1.8.0, with the change in commit 20f625b4d39d0e981b4922f629212acc74271665, sphinx-build unconditionally writes the `environment.pickle` file.  This breaks the build system expectation that a
`-b pickle` run can be done first, and then various output formats can be built off that file.

Before the above commit, the timestamp of the pickle file could be compared against an output format's timestamp to determine whether it is out of date; now it's always out of date since it keeps getting rewritten.

Reference build system integration: https://github.com/FRRouting/frr/blob/master/doc/subdir.am#L32 (uses text instead of pickle to support older sphinx versions)

eqvinox:
(sorry, forgot mention @blueyed for commit relation)

tk0miya:
#5525 is merged now.
Thanks,

Note: I opened #5594 for the original issue of 20f625b.

PR #5525, opened 2018-10-11
[1.8] Revert "Builder.build: save env also with only new documents"

eqvinox:
* [x] bugfix
* [x] 1.8 backport of #5524

This reverts commit 20f625b4d39d0e981b4922f629212acc74271665, which
introduced a regression in dependency tracking for build systems
invoking sphinx-build (described in issue #5501.)

The correct solution is to properly track when the pickle file needs to
be written;  unfortunately I don't know enough about sphinx to implement
this.  Also, I'm not sure the intersphinx cache belongs into this at
all.

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -369,14 +369,14 @@ class Builder(object):
         else:
             logger.info(__('none found'))
 
-        # save the environment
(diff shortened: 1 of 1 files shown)


tk0miya:

PR #5540, opened 2018-10-16
Fix ensuredir() in case of pre-existing file

jdemeyer:
Subject: `sphinx.util.osutil.ensuredir()` should give an error when an ordinary file (not directory) exists with the given name.

### Feature or Bugfix
- Bugfix

### Purpose
I'm currently in the process of debugging an issue with SageMath when upgrading from Sphinx 1.7 to Sphinx 1.8. I get a warning `[Errno 20] Not a directory` while copying static files due to #5541 which causes the static "directory" to be an ordinary file. This should be detected earlier by `ensuredir()`.

The current pull request also changes the philosophy of the test: we don't care about the process (why did it fail?) but we care about the result (we want a directory with the given name). By not using `errno`, it is very likely also more portable, which is an additional advantage.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -24,6 +24,7 @@ Bugs fixed
 * #5495: csv-table directive with file option in included file is broken (refs:
   #4821)
 * #5498: autodoc: unable to find type hints for a ``functools.partial``
+* #5480: autodoc: unable to find type hints for unresolvable Forward references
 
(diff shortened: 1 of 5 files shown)


tk0miya (changes requested):
+1
Could you rebase this to 1.8 branch?

PR #5541, opened 2018-10-16
Correctly deal with non-existing static dir in graphviz extension

jdemeyer:
Subject: if the `_static` destination directory does not exist, the file `graphviz.css` is copied to a *file* called `_static` instead of a file in the directory `_static`.

### Feature or Bugfix
- Bugfix


tk0miya (changes requested):
+1
Could you rebase this to 1.8 branch please?

PR #5552, opened 2018-10-19
extlinks fix #1622: support callable URL & prefix

### LINKED ISSUE #1622: sphinx.ext.extlinks: Allow Callback for URL Format String
It would be nice if instead of a formatting string, you could pass a callable that returns the combined URL for extlinks.  This would allow cases where the input `part` isn't quite part of the url.

For example, this Docker reference uses lowercase anchors, but uppercase section titles: https://docs.docker.com/reference/builder/

Here's my current extlinks config:

```
    'Dockerfile': (
        'https://docs.docker.com/reference/builder/#%s',
        ''
    )
```

To handle the difference in captalization, I need to link as follows:

```
:Dockerfile:`RUN <run>`
```

It would be preferable to define the extlink using a callable handling the case conversion:

```
    'Dockerfile': (
        lambda part: 'https://docs.docker.com/reference/builder/#%s' % part.lower(),
        ''
    )
```

I suspect this is one of many possible use cases that would benefit.  And since you could test for a callable vs a string, it should be fully backwards compatible.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1622
- Originally reported by: Kevin Stone
- Originally created at: 2014-11-09T18:38:07.878

abathur:
### Feature or Bugfix
- Feature

### Purpose
- Support more complex extlinks use cases by providing a callable in the place of the URL and/or the prefix. 

### Relates
- #1622

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,6 +31,7 @@ Other contributors, listed alphabetically, are:
 * Kevin Dunn -- MathJax extension
 * Josip Dzolonga -- coverage builder
 * Buck Evan -- dummy builder
+* Travis A. Everett -- extlinks improvements
 * Matthew Fernandez -- todo extension fix
 * Hernan Grecco -- search improvements
 * Horst Gutmann -- internationalization support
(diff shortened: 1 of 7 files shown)


abathur:
Hrm. I think I must be misusing fixtures or roots here in a way that left something dangling in _build unexpectedly. 

I'm going to close for now, extract create a separate smaller root for the added tests, and give this another try later.

PR #5553, opened 2018-10-19
sphinx.ext.extlinks: support callable URL & prefix (resolves #1622)

### LINKED ISSUE #1622: sphinx.ext.extlinks: Allow Callback for URL Format String

abathur:
### Feature or Bugfix
- Feature

### Purpose
- Support more complex extlinks use cases by providing a callable in the place of the URL and/or the prefix.

### Relates
- #1622

### DIFF
--- a/sphinx/ext/extlinks.py
+++ b/sphinx/ext/extlinks.py
@@ -31,29 +31,37 @@ from sphinx.util.nodes import split_explicit_title
 
 if False:
     # For type annotation
-    from typing import Any, Dict, List, Tuple  # NOQA
+    from typing import Any, Callable, Dict, List, Tuple, Union  # NOQA
     from docutils.parsers.rst.states import Inliner  # NOQA
     from sphinx.application import Sphinx  # NOQA
     from sphinx.util.typing import RoleFunction  # NOQA
 
 
 def make_link_role(base_url, prefix):
-    # type: (unicode, unicode) -> RoleFunction
+    # type: (Union[unicode, Callable[[unicode], unicode]], Union[unicode, Callable[[unicode], unicode]]) -> RoleFunction  # NOQA
     def role(typ, rawtext, text, lineno, inliner, options={}, content=[]):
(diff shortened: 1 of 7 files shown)


sphinx/ext/extlinks.py
```diff
@@ -31,29 +31,37 @@
...
         except (TypeError, ValueError):
             inliner.reporter.warning(
                 'unable to expand %s extlink with base URL %r, please make '
                 'sure the base contains \'%%s\' exactly once'
                 % (typ, base_url), line=lineno)
-            full_url = base_url + part
+            if callable(base_url):
+                full_url = base_url(part)
+            else:
+                full_url = base_url + part
```
abathur, line 59: Not sure how I feel about this part. Running the user's callable could _also_ trigger errors caught here.

Rather than try and fiddle with separating the conditions based on the exception message text, I'm just re-calling if the url is callable so that the exception can propagate.

Separating these code paths better could work around this, but I think it'd be at the expense of making it less obvious what the code does. What do you think?

Issue #5561, opened 2018-10-24
make all-pdf fails invocation of xindy


JulienPalard:
On cpython, branch 3.7, if we run a `make all-pdf` in build/latex/ we're getting:
```
$ make all-pdf
latexmk -pdf -dvi- -ps-  'c-api.tex'
Latexmk: This is Latexmk, John Collins, 11 Nov. 2012, version: 4.35.
**** Report bugs etc to John Collins <collins at phys.psu.edu>. ****
Latexmk: applying rule 'makeindex c-api.idx'...
Rule 'makeindex c-api.idx': File changes, etc:
   Non-existent destination files:
      'c-api.ind'
------------
Run number 1 of rule 'makeindex c-api.idx'
------------
------------
Running 'internal xindy -L english -C utf8  -M sphinx.xdy -I xelatex  -o "c-api.ind" "c-api.idx"'
------------
Latexmk: calling xindy( -L english -C utf8 -M sphinx.xdy -I xelatex -o c-api.ind c-api.idx )
Unsupported input markup xelatex.

usage: xindy [-V?h] [-qv] [-d magic] [-o outfile.ind] [-t log] \
            [-L lang] [-C codepage] [-M module] [-I input] \
            [--interactive] [--mem-file xindy.mem] \
	    [idx0 idx1 ...]

GNU-STYLE LONG OPTIONS FOR SHORT OPTIONS:

 -V / --version
 -? / -h / --help
 -q / --quiet
 -v / --verbose
 -d / --debug          (multiple times)
                       (supported: script, keep_tmpfiles, markup, level=n)
 -o / --out-file
 -t / --log-file
 -L / --language
 -C / --codepage
 -M / --module         (multiple times)
 -I / --input-markup   (supported: latex, omega, xindy)

Latexmk: Errors, so I did not complete making targets

```

### Environment info
- OS: Ubuntu 14.04.5 LTS
- Python version: 3.6.2
- Sphinx version: 1.8.1
- xindy from http://us.archive.ubuntu.com/ubuntu/ version 2.4-1.3

JulienPalard:
With xindy 2.5.1.20160104-5 from http://ftp.fr.debian.org/debian buster/main it works.

tk0miya:
I checked recent releases:

* xindy on bionic (18.04) ... suppoted
* xindy on xenial (16.04) ... not supported

@jfbu I think `-I xelatex` option for `xindy` is too new. So I think it should be removed. What do you think?

[...]

PR #5574, opened 2018-10-30
LaTeX: do not use `-I xelatex` option with xindy (Fix: #5561)

jfbu:
It is not supported by old xindy (TeXLive 2013 or earlier), and probably
irrelevant anyhow as the .idx file should not have ^^ab or ^^^^abcd
entries.

### Relates
- #5561, #5134


jfbu:
Currently there is a problem with building cpython 3.7 PDF docs, which is unrelated to this.

(it is caused by trying to index the `;` character, which creates in latex output a misformed `\index` usage, trying to index the empty set... with xindy, this causes a failed build...)

tk0miya (approved):

Issue #4086, opened 2017-09-26
Different numfig_format settings for caption prefix and reference


asfg84:
Since the numfig_format is used as caption prefix and also for referencing figures/tables etc., it is not possible to use different formats. I would like to have a reference in text as _"see Figure 1"_, and the caption of my figure should be 

_"Figure 1: Example image"_

Instead, with `numfig_format={'figure': 'Figure %s'}` I get 

_"Figure 1 Example image"_ (without colon)

The output in Latex pdf is correct, this only applies to the html output.

When I add the colon to the `numfig_format numfig_format={'figure': 'Figure %s:'}`, I also get the colon in the text reference, which I don't want to have. (And I have two colons in the caption of the latex output and no space between Figure and the number anymore).

### Environment info
- Sphinx version: 1.6.3

misev:
+1, I guess this should be trivial to correct?

Also I came across a bug: if you change only the 'figure' format with `numfig_format={'figure': 'Figure %s'}`, and also have Tables for example in the document, you get this:
```
Exception occurred:
  File "/usr/lib/python2.7/dist-packages/sphinx/writers/html.py", line 279, in append_fignumber
    self.body.append(prefix % numsstr + ' ')
TypeError: not all arguments converted during string formatting
The full traceback has been saved in /tmp/sphinx-err-PjF2Qy.log, if you want to report the issue to the developers.
```
Specifying all formats fixed the problem:
```
numfig_format = {'figure': 'Figure %s', 'table': 'Table %s',
                 'code-block': 'Listing %s', 'section': 'Section %s'}
```

tk0miya:
>+1, I guess this should be trivial to correct?

Really? We always welcome your contribution. Please send a PR for us :-)

On the other hand, second one is another bug. Could you file a new issue please?
I'll take a look later.

misev:
> Really? We always welcome your contribution. Please send a PR for us :-)

I don't think it is very efficient for me to go into fixing something small (at least it looks so, can't be sure of course), given that so far I have zero experience contributing to this project :-)

New issue for the other problem: #4828

tk0miya:
Thanks for filing. I'll take a look later.

About another case, I feel it is difficult to fix. So I can't promise you exactly when. Please challenge the problem If you are interested in :-)

tesch1:
I'm interested in it :o)

Before I dive in, here's my proposal, what do you all think?

[...]

Issue #3209, opened 2016-12-09
Smart :numref: for Section or Chapter referencing


rakhimov:
It would be great if `numref` used "Chapter %s" for top-level references, e.g., Chapter 1,
and "Section %s" for sections within chapters. e.g., Section 1.1.
This could be deduced from "%s" part.

Is there a way to provide smarter handler for "numfig_format['section']"?
For example: 'lambda s: "Section %s" %s if len(s.split('.')) > 1 else "Chapter %s" % s'

I tried a simple class with overloaded __mod__ operator but no luck with Pickler complaining.
I would appreciate if this feature is available.

rakhimov:
Also "Appendix %s" if s is alphabetic.

chezou:
+1 for this feature. Currently, I wrote "Section %s" manually as follows:

```
:numref:`Section %s <some_label>`
```

It would be nice to have the function.

paternal:
I submitted a pull request #5575 that implements a way to solve this, but it seems to have been unnoticed…

tk0miya:
Note: we have to implementation sectioning-unit feature at first. It would be related with navigation (menus, prev/next links and so on).

PR #5575, opened 2018-10-30
`numfig_format` can be defined using a `numfig_format()` node method.

paternal:
⚠️ This is a proof-of-concept. It does not include tests or documentation; I will write it if you accept this idea.

Subject: Allow more complex formatting for `numfig_format`.

# Feature description

Caption and references of an enumerable node are defined using:

1. Configuration variable [numfig_format](http://www.sphinx-doc.org/en/stable/usage/configuration.html#confval-numfig_format).
2. If conf variable `numfig_format` is not defined: method `numfig_format()` of the node (this is the feature implemented by this pull request).
3. If method `numfig_format()` is not defined: the empty string.

Being able to define `numfig_format` by a method is much more powerful than a string.

# Writers affected

This feature only works with html and html5 writers.

* LaTeX writer: the way `numfig_format` is managed right now is [kind of a hack](https://github.com/sphinx-doc/sphinx/blob/c97441e2afaa01eebde01716ebbee8b0d40c715d/sphinx/writers/latex.py#L744-L782), and would not work easily with this new feature. So, it is up to the extension writer to handle LaTeX on her own.
* Other writers: I do not use/know them.

# Problems solved by this feature

- In my sphinx extension [sphinxcontrib-proof](https://sphinxcontrib-proof.readthedocs.io), I define definitions, theorems, etc, which share the same counter. I cannot name them differently (*Theorem 1*, *Definition 2*, etc.) using the string `numfig_format`. With this pull request, I achieve this by defining the method `Node.numfig_format()`:

  ~~~python
    class NumberedStatementNode(_StatementNode):
        """Statement with a number."""

        def numfig_format(self, builder, figtype, ref):
            """Give each theorem type its own name."""
            # `self["name"]` is "Theorem" or "Definition" or etc.
            return self["name"] + " %s "
  ~~~

- Issue sphinx-doc/sphinx#3209 can be solved by implementing such a `numfig_format()` method.
- Issue sphinx-doc/sphinx#4086 can be also solved by implementing such a method, for a boolean `ref` argument is provided to this `numfig_format()` (which make it possible to return different names whether the returned string is used  to name a node or to reference it).

# Next

If you agree with this idea, I can write test and documentation.

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -746,8 +746,12 @@ class StandardDomain(Domain):
         try:
             if node['refexplicit']:
                 title = contnode.astext()
+            elif figtype in env.config.numfig_format:
+                title = env.config.numfig_format[figtype]
+            elif hasattr(target_node, "numfig_format"):
+                title = target_node.numfig_format(builder, figtype, ref=True)
             else:
-                title = env.config.numfig_format.get(figtype, '')
+                title = ''
 
             if figname is None and '{name}' in title:
                 logger.warning(__('the link has no caption: %s'), title, location=node)
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -332,13 +332,15 @@ class HTMLTranslator(BaseTranslator):
             if figure_id in self.builder.fignumbers.get(key, {}):
                 self.body.append('<span class="caption-number">')
                 prefix = self.builder.config.numfig_format.get(figtype)
+                if prefix is None and hasattr(node, "numfig_format"):
+                    prefix = node.numfig_format(self.builder, figtype, ref=False)
                 if prefix is None:
                     msg = __('numfig_format is not defined for %s') % figtype
                     logger.warning(msg)
                 else:
                     numbers = self.builder.fignumbers[key][figure_id]
                     self.body.append(prefix % '.'.join(map(str, numbers)) + ' ')
-                    self.body.append('</span>')
+                self.body.append('</span>')
 
         figtype = self.builder.env.domains['std'].get_enumerable_node_type(node)
         if figtype:
--- a/sphinx/writers/html5.py
+++ b/sphinx/writers/html5.py
@@ -300,13 +300,15 @@ class HTML5Translator(BaseTranslator):
             if figure_id in self.builder.fignumbers.get(key, {}):
                 self.body.append('<span class="caption-number">')
                 prefix = self.builder.config.numfig_format.get(figtype)
+                if prefix is None and hasattr(node, "numfig_format"):
+                    prefix = node.numfig_format(self.builder, figtype, ref=False)
                 if prefix is None:
                     msg = __('numfig_format is not defined for %s') % figtype
                     logger.warning(msg)
                 else:
                     numbers = self.builder.fignumbers[key][figure_id]
                     self.body.append(prefix % '.'.join(map(str, numbers)) + ' ')
-                    self.body.append('</span>')
+                self.body.append('</span>')
 
         figtype = self.builder.env.domains['std'].get_enumerable_node_type(node)
         if figtype:


tk0miya:
I'm very sorry for late response.

We already have `app.add_enumerable_node()` to register enumerable nodes to Sphinx. So it is better to enhance the API to me. It's natural.

paternal:
> We already have `app.add_enumerable_node()` to register enumerable nodes to Sphinx. So it is better to enhance the API to me. It's natural.

Either I was not clear enough in my explanations (and you did not understand my problem), or I did not understand your answer.

I am the author of the [proof](https://git.framasoft.org/spalax/sphinxcontrib-proof) extension, which define directives to typeset theorems, definitions, etc. ([example](https://jouets.readthedocs.io/fr/latest/dobble/math/#combinatoire)). I would like the user to be able to define the needed environments on the fly. For instance, in LaTeX, one would write the following lines to define two environments.

~~~latex
\newtheorem{theorem}{Theorem}
\newtheorem{definition}{Definition}
~~~

In the current implementation, there is *one* `NumberedStatementNode` for everything (theorems, definitions, etc.).

If I use the enumerable nodes without hacks, I add a [numfig_format](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-numfig_format) entry `numfig_format['proof'] = 'Foo %s'`, and I have two problems.

1. My environment are all labelled with the same word (e.g. "Foo 1", "Foo 2", "Foo 3", although I want "Theorem 1", "Definition 2", "Theorem 3").
2. They are referenced with the same word (e.g. "Foo 1", "Foo 2", "Foo 3", although I want "Theorem 1", "Definition 2", "Theorem 3").

With some work, I managed to solve 1 (although I consider it a dirty hack), but I could not solve 2.

**To sum up, I would like to customize `numfig_format`.**

I have considered other solutions:

- I can define one Node subclass for Theorem, one Node subclass for Definition, etc. But they would be hard-coded, and I want the user to be able to define them on the fly (as done with LaTeX).
- Using a metaclass, I can define on-the-fly one Node subclass for each Theorem, Definition, etc., but at some point, data is pickled, and pickle does not like metaclasses (see for instance [this stackoverflow answer](https://stackoverflow.com/a/48623956)).

At last, using one single Node class for every Theorem, Definition, etc., they also share the `numfig_format` entry. Hence this pull request, which makes `numfig_format` customizable: each node object can define its `numfig_format` entry on-the-fly, when it is used (and customize it depending of this specific node attributes; see the first item of part `Problems solved by this feature` in [my previous comment](#issue-227100939)).

Sorry for the long post… and thanks for your work.
Louis

tk0miya:
To my understanding, the document tree model (a.k.a. doctree) expects that one node represents only one document structure. There are many kind of nodes to represent its document structure ([ref](http://docutils.sourceforge.net/docs/ref/doctree.html)). Therefore, based on the model, it would be better that theorem and definitions are defined as each other node. Sphinx also obeys the doctree model. So it expects same kind of nodes are enumerated as same.

I understand your extension goes that dynamically. But I believe it is conflicted with the concept of doctree. So, -1 for switching numfig_format by methods or attributes. But it does not mean "never support it". I can agree to add new API to realize this feature if it is reasonable. For example, to enhance Domains might be better for this case. I'll consider how we can do for the issue.

Thanks,

paternal:
Thanks for your answer.

> same kind of nodes are enumerated as same

Last thing I forgot: in math papers, it is customary to *share one single counter* between definitions, theorems, etc. (e.g. "Definition 1; Theorem 2; Definition 3" instead of "Definition 1 ; Theorem 1 ; Definition 2" ; see [a real example](https://www.doc.ic.ac.uk/~imh/papers/can.pdf)), which is one of the reasons I used a single Node class for everything (Definitions, theorems, etc.). Not sure I can achieve this by using one Node class for definitions, one Node class for theorems, etc.

(And happy new year!)

Issue #5576, opened 2018-10-30
latex writer incorrectly handles indexing a sole semi-colon character


jfbu:
Consider the following source:

```
.. index::
   single: clause
   single: suite
   single: ; (semicolon)
```

which can be seen in [Python documentation sources](https://github.com/python/cpython/blob/3.7/Doc/reference/compound_stmts.rst)

When processed by LaTeX writer the semi-colon is mistakendly interpreted as separator and as a result an invalid empty index entry arises.

When the PDF is built with `latex_use_xindy = True` this causes failed build with this error:

```
ERROR: CHAR: index 0 should be less than the length of the string
```

This arose in discussion of #5561 issue.

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.8.2+

jfbu:
fixed at 4df3b067 (refs: #5583)

PR #5577, opened 2018-10-30
LaTeX: use lookbehind regex to not match semicolon at start

jfbu:
Closes: #5576

Here is screen shot of a portion of the index in library.pdf from CPython 3.7 documentation, showing indexing of `;` now works:

capture d ecran 2018-10-30 a 23 27 42

I used a look-behind regex as advocated at this [stackoverflow answer](https://stackoverflow.com/a/15669590/4184837)

### Relates

- #5561

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1884,7 +1884,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
         # type: (nodes.Node) -> None
         self.body.append('\n\\end{flushright}\n')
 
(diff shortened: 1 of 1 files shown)


tk0miya (approved):

jfbu:
@tk0miya thanks a lot for quick review. I will soon make however another PR to fix #5579 and it fixes the #5576 issue as well. But it is less minimal change.

jfbu:
This is closed without merging because it got superseded by #5583.

Issue #5579, opened 2018-10-31
LaTeX: entries in general index are not styled like entries of domain indices


jfbu:
### Problem

Here for example from CPython 3.7 library.pdf (build using PR #5577 to workaround #5576 issue). The Python module index looks like this:

![capture d ecran 2018-10-31 a 16 21 27](https://user-images.githubusercontent.com/2589111/47798632-1590aa00-dd29-11e8-9e29-b9ab9ec7e9c1.png)

In contrast the general index looks like this

![capture d ecran 2018-10-31 a 16 17 57](https://user-images.githubusercontent.com/2589111/47798476-b7fc5d80-dd28-11e8-9541-ee0c112246f1.png)

It doesn't use the same monospaced font, and extras in parentheses are not in italics.

Furthermore, there are problems with underscores for private methods, it would be better with monospace font as for literals.

![capture d ecran 2018-10-31 a 16 18 19](https://user-images.githubusercontent.com/2589111/47798565-ed08b000-dd28-11e8-84db-b3b64d5227f1.png)

The above build of CPython library.pdf uses xelatex and xindy, but it would be the same with pdflatex and makeindex.

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.8.1 + #5577 patch for issue #5576

jfbu:
Fixed at 00a75b27 (refs: #5583)

PR #5583, opened 2018-10-31
LaTeX: add styling to general index, similar to domain indices

### LINKED ISSUE #5576: latex writer incorrectly handles indexing a sole semi-colon character

jfbu:
Closes: #5579. Also closes #5576.

With this here is for example how CPython 3.7 library.pdf renders:

capture d ecran 2018-10-31 a 19 16 38

and 

capture d ecran 2018-10-31 a 19 16 52

Compare to the images posted at #5579.

About the underscore character, it shows fine because CPython uses xelatex for pdf builds and this means another font is used than the Sphinx default `Courier`. Unfortunately the Courier font used by default for monospace font by Sphinx has an underscore glyph which creates continuous lines... one can see that in Sphinx own pdf doc, see top left.

capture d ecran 2018-10-31 a 19 09 22

Sphinx own doc uses `\footnotesize\raggedright` which explains smaller size.

It fixes #5576 in a way different from #5577: the regex argument of `visit_index()` in latex.py is used for something completely different: separate from the index a parenthesized comment ("extra") and style it especially as in Python Module Index.

Another slightly breaking change is in `\sphinxstyleindexextra`, the parentheses stay upright. I feel it looks better. (I will do more detailed review some time later)

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -1623,8 +1623,11 @@
 
 % additional customizable styling
 \def\sphinxstyleindexentry   #1{\texttt{#1}}
-\def\sphinxstyleindexextra   #1{ \emph{(#1)}}
+\def\sphinxstyleindexextra   #1{ (\emph{#1})}
 \def\sphinxstyleindexpageref #1{, \pageref{#1}}
+\def\sphinxstyleindexpagemain#1{\textbf{#1}}
+\protected\def\spxentry#1{#1}% will get \let to \sphinxstyleindexentry in index
+\protected\def\spxextra#1{#1}% will get \let to \sphinxstyleindexextra in index
 \def\sphinxstyleindexlettergroup #1%
     {{\Large\sffamily#1}\nopagebreak\vspace{1mm}}
 \def\sphinxstyleindexlettergroupDefault #1%
--- a/sphinx/texinputs/sphinx.xdy
+++ b/sphinx/texinputs/sphinx.xdy
@@ -3,11 +3,12 @@
 ;; Unfortunately xindy is out-of-the-box hyperref-incompatible.  This
 ;; configuration is a workaround, which requires to pass option
 ;; hyperindex=false to hyperref.
-;; textit and emph not currently used by Sphinx LaTeX writer.
-(define-attributes (("textbf" "textit" "emph" "default")))
+;; textit and emph not currently used, spxpagem replaces former textbf
+(define-attributes (("textbf" "textit" "emph" "spxpagem" "default")))
 (markup-locref :open "\textbf{\hyperpage{" :close "}}" :attr "textbf")
 (markup-locref :open "\textit{\hyperpage{" :close "}}" :attr "textit")
 (markup-locref :open "\emph{\hyperpage{" :close "}}" :attr "emph")
+(markup-locref :open "\spxpagem{\hyperpage{" :close "}}" :attr "spxpagem")
 (markup-locref :open "\hyperpage{" :close "}" :attr "default")
 
 (require "numeric-sort.xdy")
@@ -193,6 +194,9 @@
 (markup-index :open  "\begin{sphinxtheindex}
 \let\lettergroup\sphinxstyleindexlettergroup
 \let\lettergroupDefault\sphinxstyleindexlettergroupDefault
+\let\spxpagem\sphinxstyleindexpagemain
+\let\spxentry\sphinxstyleindexentry
+\let\spxextra\sphinxstyleindexextra
 
 "
 	      :close "
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1884,7 +1884,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
         # type: (nodes.Node) -> None
         self.body.append('\n\\end{flushright}\n')
 
-    def visit_index(self, node, scre=re.compile(r';\s*')):
+    def visit_index(self, node, extrare=re.compile(r'^(.*\S)\s+\(([^()]*)\)\s*$')):
         # type: (nodes.Node, Pattern) -> None
         def escape(value):
             value = self.encode(value)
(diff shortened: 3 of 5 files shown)


sphinx/texinputs/sphinx.sty
```diff
@@ -1623,8 +1623,11 @@
 
 % additional customizable styling
 \def\sphinxstyleindexentry   #1{\texttt{#1}}
-\def\sphinxstyleindexextra   #1{ \emph{(#1)}}
+\def\sphinxstyleindexextra   #1{ (\emph{#1})}
```
jfbu, line 1626: I moved the parentheses outside the scope of `\emph`. imho, it looks better this way.

sphinx/texinputs/sphinx.sty
```diff
@@ -1623,8 +1623,11 @@
...
 \def\sphinxstyleindexpageref #1{, \pageref{#1}}
+\def\sphinxstyleindexpagemain#1{\textbf{#1}}
+\protected\def\spxentry#1{#1}% will get \let to \sphinxstyleindexentry in index
+\protected\def\spxextra#1{#1}% will get \let to \sphinxstyleindexextra in index
```
jfbu, line 1630: The `visit_index()` from LaTeX writer will use `\spxentry` and `\spxextra`, this makes smaller LaTeX file than using everywhere `\sphinxstyleindexentry`, `\sphinxstyleindexextra`. The macros need to be `\protected` because they are used inside `\index{}` macro, and this may occur for example in a `tabulary` environment (same reason why verbatim can't be used inside `tabulary`).

sphinx/texinputs/sphinx.xdy
```diff
@@ -3,11 +3,12 @@
...
 ;; hyperindex=false to hyperref.
-;; textit and emph not currently used by Sphinx LaTeX writer.
-(define-attributes (("textbf" "textit" "emph" "default")))
+;; textit and emph not currently used, spxpagem replaces former textbf
+(define-attributes (("textbf" "textit" "emph" "spxpagem" "default")))
 (markup-locref :open "\textbf{\hyperpage{" :close "}}" :attr "textbf")
 (markup-locref :open "\textit{\hyperpage{" :close "}}" :attr "textit")
 (markup-locref :open "\emph{\hyperpage{" :close "}}" :attr "emph")
+(markup-locref :open "\spxpagem{\hyperpage{" :close "}}" :attr "spxpagem")
 (markup-locref :open "\hyperpage{" :close "}" :attr "default")
```
jfbu, line 12: Initially I had also added a `spxpage` attribute so that one would would have `\spxpage{\hyperpage{37}}` for example, allowing to style the page number. But this would mean all `\index` macros would end in `\index{...|spxpage]` increasing size and overhead of LaTeX file. Then I thought about simply modifying the `"default"` above to do `...\spxpage{\hyperpage{...`, but this method would have been xindy-only. Only the `\index{....|spxpage}` would have also affected `makeindex` way.

In the end I dropped this: it is always possible if wanted to style the index page number at LaTeX macro level by modifying `\hyperpage`.

sphinx/texinputs/sphinx.xdy
```diff
@@ -193,6 +194,9 @@
 (markup-index :open  "\begin{sphinxtheindex}
 \let\lettergroup\sphinxstyleindexlettergroup
 \let\lettergroupDefault\sphinxstyleindexlettergroupDefault
+\let\spxpagem\sphinxstyleindexpagemain
+\let\spxentry\sphinxstyleindexentry
+\let\spxextra\sphinxstyleindexextra
```
jfbu, line 199: Regarding `\sphinxstyleindexentry` and `\sphinxstyleindexextra` they are the macros used by the LaTeX writer for the mark-up associated with domain indices. The method above allows same mark-up also in general index. But it means too that separate customization will mean from user some extra measure. One method would add to preamble some definition of `\stylegeneralindexentry`,  `\stylegeneralindexextra` macros, and insert `\let\sphinxstyleindexentry\stylegeneralindexentry\let\...` before the final index via the `latex_elements` `'printindex'` key.

I also hesitated about moving these `\let` to the definition of the `sphinxtheindex` environment, but the latter is used for all indices, including domain indices, where `\spxentry` and `\spxextra` do not arise.

sphinx/writers/latex.py
```diff
@@ -1884,7 +1884,7 @@ def depart_attribution(self, node):
         # type: (nodes.Node) -> None
         self.body.append('\n\\end{flushright}\n')
 
-    def visit_index(self, node, scre=re.compile(r';\s*')):
+    def visit_index(self, node, extrare=re.compile(r'^(.*\S)\s+\(([^()]*)\)\s*$')):
```
jfbu: This is breaking change for extensions who counted on the third argument of `visit_index` to be the regex allowing to recognize semi-colon. A priori this was not used as public API? so I felt there is no problem using this for something completely different.

The aim of the regex is to identify strings of the type `"text (extra in parentheses)"`. Then the `extra in parenthesis` will be styled especially. This is a new feature which has drawback that sometimes the parentheses were used by Sphinx user with a somewhat different intent. I found an example in Sphinx own docs, [at line 125 of doc/man/sphinx-quickstart.rst](https://github.com/sphinx-doc/sphinx/blob/master/doc/man/sphinx-quickstart.rst):

```
.. option:: --use-make-mode (-m), --no-use-make-mode (-M)
```

In this case this will produce in LaTeX file:

```
\index{--use-make-mode (-m), --no-use-make-mode (-M)@\spxentry{--use-make-mode (-m), --no-use-make-mode}\spxextra{-M}!...}
```

which causes the `-M` to be emphasized. In general any 

```
.. option:: --long-name (-l)
```

will get the `-l` styled by `\spxextra`, where the `--long-name` gets styled by `\spxentry`.

But this syntax is not the one described at https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option, so maybe this is not really breaking change. And it only affects typography.

sphinx/writers/latex.py
```diff
@@ -1895,33 +1895,51 @@ def escape(value):
             value = value.replace('!', '"!')
             return value
 
+        def style(string):
+            match = extrare.match(string)
+            if match:
+                return match.expand(r'\\spxentry{\1}\\spxextra{\2}')
```
jfbu, line 1903: This mark-up is modeled on 

https://github.com/sphinx-doc/sphinx/blob/8b687ac30481e7204677138d8f2f3b49e354963e/sphinx/writers/latex.py#L800-L804

which applies to entries for domain indices. But here the "extra" is recovered from this input syntax:
```
.. index::
   single: foo (bar)
```

sphinx/writers/latex.py
```diff
@@ -1895,33 +1895,51 @@ def escape(value):
...
-                    p = scre.sub('!', escape(string))
-                    self.body.append(r'\index{%s%s}' % (p, m))
+                    try:
+                        p1, p2 = [escape(x) for x in split_into(2, 'single', string)]
+                        P1, P2 = style(p1), style(p2)
+                        self.body.append(r'\index{%s@%s!%s@%s%s}' % (p1, P1, p2, P2, m))
+                    except ValueError:
+                        p = escape(split_into(1, 'single', string)[0])
+                        P = style(p)
+                        self.body.append(r'\index{%s@%s%s}' % (p, P, m))
```
jfbu, line 1927: This also fixes #5576.

sphinx/writers/latex.py
```diff
@@ -1895,33 +1895,51 @@ def escape(value):
...
                     p1, p2, p3 = [escape(x) for x in split_into(3, 'triple', string)]
                     self.body.append(
-                        r'\index{%s!%s %s%s}\index{%s!%s, %s%s}'
-                        r'\index{%s!%s %s%s}' %
-                        (p1, p2, p3, m, p2, p3, p1, m, p3, p1, p2, m))
+                        r'\index{%s@\spxentry{%s}!%s %s@\spxentry{%s %s}%s}'
+                        r'\index{%s@\spxentry{%s}!%s, %s@\spxentry{%s, %s}%s}'
+                        r'\index{%s@\spxentry{%s}!%s %s@\spxentry{%s %s}%s}' %
+                        (p1, p1, p2, p3, p2, p3, m, p2, p2, p3, p1, p3, p1, m,
+                         p3, p3, p1, p2, p1, p2, m))
```
jfbu: Here for triples  I dropped addition of `\spxextra` styling. But I should add it, I guess.

sphinx/writers/latex.py
```diff
@@ -1895,33 +1895,51 @@ def escape(value):
...
                 elif type == 'see':
                     p1, p2 = [escape(x) for x in split_into(2, 'see', string)]
-                    self.body.append(r'\index{%s|see{%s}}' % (p1, p2))
+                    P1 = style(p1)
+                    self.body.append(r'\index{%s@%s|see{%s}}' % (p1, P1, p2))
```
jfbu, line 1946: I did not make any attempt into adding mark-up in the `see{%s}` part as I expected problems, especially to maintain compatibility with both `makeindex` and `xindy`.

sphinx/texinputs/sphinx.sty
jfbu: I decided to re-use the same `\sphinxstyleindexentry` and `\sphinxstyleindexextra` already used by LaTeX writer for domain indices at https://github.com/sphinx-doc/sphinx/blob/8b687ac30481e7204677138d8f2f3b49e354963e/sphinx/writers/latex.py#L784.

An alternative would be to define independent macros `\sphinxstylegeneralindexentry`, etc...

The `\spxentry/\spxextra` are privately used, only `\sphinxstyleindexentry/...` should get redefined by user. The reason for `\spxentry/\spxextra` is for shorter macro names, hence smaller file.

On the xindy side I looked a bit about using `markup-indexentry` which currently (in `sphinx.xdy`) is

```
(markup-indexentry :open "~n  \item "           :depth 0)
```
but did not see a way to add there the  `\spxentry{..}` wrapper. Because in case of sub-items, a closing `}` for example would enclose all sub-items, causing errors. And anyway, the `@\spxentry{...}` works both with `makeindex` and `xindy`. Furthermore I later decided (to match the domain indices) to use `\spxentry{...}\spxextra{...}`,  not `\spxentry{...\spxextra{...}}`, which I had started with.

sphinx/writers/latex.py
jfbu: I will push update to incorporate `\spxextra` styling for `p1`, `p2`, `p3` if they use `foo (bar)` style also in this "triple" case. (done at 0f45cf75b)

sphinx/writers/latex.py
tk0miya: Yes, I agree with this is not a public API.  But this must be a breaking change. So -1 for changing signature of method.

I think marking `scre` keyword as deprecated is better. And I think `extrare` is not needed to substitute. So how about moving it to constant?
```
EXTRA_RE = re.compile('...')

...

def visit_index(self, node, scre=None):
    if scre:
        warnings.warn('scre option for visit_index() is deprecated')

    ...
```

tk0miya (changes requested):
LGTM with nits

tk0miya (approved):
LGTM!

jfbu:
@tk0miya thanks for quick reviewing! I will merge now into 1.8.2 although this is less minimal change than #5577. My estimate indeed is that absence of styling of general index in LaTeX output was a bug.

PR #5590, opened 2018-11-04
Move language-specific data into a new JS file, language_data.js

### LINKED ISSUE #5460: Search does not work for docs built with 1.8.0

mitya57:
Subject: Make search work with old templates

### Feature or Bugfix
Bugfix

### Purpose
Not all templates use `documentation_options.js`. The examples are old versions of sphinx-rtd-theme and many other third-party themes. However currently searchtools.js does not work without some classes, which are defined in `documentation_options.js`. This pull request moves these classes to a separate file `language_data.js` which is always included, even for old templates.

### Relates
Fixes #5460.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -44,6 +44,7 @@ epub_post_files = [('usage/installation.xhtml', 'Installing Sphinx'),
 epub_exclude_files = ['_static/opensearch.xml', '_static/doctools.js',
                       '_static/jquery.js', '_static/searchtools.js',
                       '_static/underscore.js', '_static/basic.css',
+                      '_static/language_data.js',
                       'search.html', '_static/websupport.js']
 epub_fix_images = False
 epub_max_image_width = 0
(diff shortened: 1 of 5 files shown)


TimKam (commented):
I think this looks good and can confirm it fixes the issue with the ReadTheDocs template. Still, I wonder if it makes more sense to include ``documentation_options.js_t`` via ``self.add_js_file()`` and to fix everything in ``documentation_options.js_t``. What do you think, @tk0miya?

mitya57:
@TimKam it is difficult to load it via `self.add_js_file()` because it needs the `data-url_root` attribute. I am not sure I understand fully why that attribute is used/needed, so I implemented a solution that is easier.

TimKam:
Thanks for clarifying. Then I think there is no trivial way to improve this. I'm fine with this PR :-)

tk0miya (approved):

tk0miya:
+1.

I just released 1.8.2 without merging this because of overlooking. So I'm merging this for 1.8.3.

PR #5596, opened 2018-11-07
Simplify XRefRole constructor

cocoatomo:
Subject: Simplify XRefRole constructor

### Feature or Bugfix

- Feature

### Purpose
- Original code setting `nodeclass` and `innernodeclass` is unclear, because default values and custom values assignment statements are separated. So, restructuring and simplifying of constructor is needed.

### Detail
- (Edited) ~Put default values on appropriate place.~
- Setting `nodeclass` and `innernodeclass` property by single statement respectively.
- More concrete type annotation to represent which custom classes are to be passed.

### Relates
N/A

### DIFF
--- a/sphinx/roles.py
+++ b/sphinx/roles.py
@@ -68,19 +68,14 @@ class XRefRole:
     * Subclassing and overwriting `process_link()` and/or `result_nodes()`.
     """
 
-    nodeclass = addnodes.pending_xref  # type: Type[nodes.Node]
-    innernodeclass = nodes.literal
-
     def __init__(self, fix_parens=False, lowercase=False,
                  nodeclass=None, innernodeclass=None, warn_dangling=False):
-        # type: (bool, bool, Type[nodes.Node], Type[nodes.Node], bool) -> None
+        # type: (bool, bool, Type[nodes.reference], Type[nodes.TextElement], bool) -> None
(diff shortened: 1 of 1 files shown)


sphinx/roles.py
```diff
@@ -68,19 +68,14 @@ class XRefRole:
     * Subclassing and overwriting `process_link()` and/or `result_nodes()`.
     """
 
-    nodeclass = addnodes.pending_xref  # type: Type[nodes.Node]
-    innernodeclass = nodes.literal
```
tk0miya: At this moment, subclasses can override the default nodeclass and innernodeclass by overriding the class attribute. But this change breaks it. The technique is not used in Sphinx-core. But I can't say about 3rd party extensions.
I don't know it is promised interface, but this must be a breaking change. So we have to provide migration pass and deprecation warnings.

IMO, it seems this is styling change. So the migration is a bit tough work compared with benefit.

tk0miya (changes requested):

sphinx/roles.py
cocoatomo: > At this moment, subclasses can override the default nodeclass and innernodeclass by overriding the class attribute.

I have missed this point, so I will revert the initialization code.

tk0miya (approved):
LGTM!

Issue #5562, opened 2018-10-24
Can't use intersphinx with Sphinx's documentation settings


ericholscher:
### Problem
- I can't use intersphinx with Sphinx's documentation settings

#### Procedure to reproduce the problem
```
intersphinx_mapping = {
    'sphinx': ('http://sphinx.readthedocs.io/en/latest/', None),
}
:ref:`gettext_uuid <sphinx:gettext_uuid>`
```

#### Error logs / results
```
/Users/eric/projects/readthedocs.org/docs/guides/manage-translations.rst:43: WARNING: undefined label: sphinx:gettext_uuid (if the link has no caption the label must precede a section header)
```

#### Expected results

It should generate a URL to the proper Sphinx docs

### Debugging

I tracked this down to the fact that my Sphinx run doesn't know about the `std:confval` reference type. I have it in the intersphinx data:

```
ipdb> env.intersphinx_inventory.keys()
dict_keys(['py:class', 'py:method', 'py:module', 'py:attribute', 'py:data', 'py:function', 'py:classmethod', 'py:exception', 'py:staticmethod', 'std:option', 'std:label', 'std:setting', 'std:envvar', 'std:templatefilter', 'std:term', 'std:fieldlookup', 'std:templatetag', 'std:django-admin', 'std:doc', 'c:var', 'c:type', 'c:function', 'c:member', 'c:macro', 'std:pdbcommand', 'std:cmdoption', 'std:opcode', 'std:token', 'std:2to3fixer', 'cpp:class', 'cpp:union', 'cpp:member', 'cpp:type', 'cpp:templateParam', 'cpp:function', 'cpp:functionParam', 'cpp:concept', 'js:function', 'js:class', 'rst:role', 'rst:directive', 'std:confval', 'std:event'])
```

But when I create the list of possible targets from my list of domains, it isn't there:

```
ipdb> objtypes
[...]

PR #5598, opened 2018-11-07
Make :any: work with unknown object types in intersphinx.

sir-sigurd:
Subject: Make `:any:` work with unknown object types in intersphinx

### Feature or Bugfix

Feature

### Purpose
As suggested in https://github.com/sphinx-doc/sphinx/issues/5562#issuecomment-434296574, if you want to refer to some custom object type defined in foreign project you had to define it in your project which is inconvenient. It's needed because intersphinx searches foreign inventory only for registered object types. This PR makes `:any:` role to not limit targets by registered object types.

### Relates
- #5562


tk0miya (changes requested):
It seems tests are broken. Could you check it please?

sir-sigurd:
@tk0miya, tests are fixed now.

PR #5611, opened 2018-11-11
WIP: #5605 fix Chinese search index

### LINKED ISSUE #5605: English words cannot be searched in Chinese word segmentation mode
## Problem:
I can't search for English words in Chinese search mode.

image

## Procedure to reproduce the problem

1. When I set `html_search_language = 'zh'` to use Chinese search, .
2. And add the following string `可以查看 FAQ 模块中 test 部分`.
3. search keyword `test`.

## Error results
When searching keyword `test`, there will be no result display.

## Expected results
There will be test string in search results.

## Sugestions
Why do space characters appear at the end of the word segmentation result?

I look up the source code and send the following regular expression, which will identify the space.
[sphinx/search/zh.py](https://github.com/sphinx-doc/sphinx/blob/16f47cded956bfba55380034ff30988d2ab5ee7c/sphinx/search/zh.py#L236)

Now test as follows：
```
In [3]: re.compile(u'(?u)\\w+[\u0000-\u00ff]').findall("可以查看 FAQ 模块中 test 部分")
Out[3]: ['可以查看 ', 'FAQ ', '模块中 ', 'test ']

```
The result clearly contains the space character, here is the source of the problem.
Can we solve this problem by adjusting regular expressions?

## Environment info

- OS: MacOS 10.14.1 with Python 2.7 in a sphinx specific virtualenv
- Python version: Python 2.7.10
- Sphinx version: 1.8.1
- Browser Chrome

TimKam:
Subject: generate search index for Latin words correctly if search language is Chinese

### Feature or Bugfix
- Bugfix

### Purpose
- As described at #5605
- Conf: ``html_search_language = 'zh'``

### Detail
- Removes trailing (and preceding) spaces from Latin search index terms
- WIP, because a Chinese-speaking person needs to confirm the correctness

### Relates
- Fixes #5605

### DIFF
--- a/sphinx/search/zh.py
+++ b/sphinx/search/zh.py
@@ -234,7 +234,7 @@ class SearchChinese(SearchLanguage):
     js_stemmer_code = js_porter_stemmer
     stopwords = english_stopwords
     latin1_letters = re.compile(u'(?:(?![\\s.,])[\x00-\xFF])+')
-    latin1 = []  # type: List[unicode]
+    latin_terms = []  # type: List[unicode]
 
     def init(self, options):
         # type: (Dict) -> None
@@ -251,9 +251,10 @@ class SearchChinese(SearchLanguage):
         if JIEBA:
             chinese = list(jieba.cut_for_search(input))
 
-        self.latin1 =\
+        latin1 = \
             [term.strip() for term in self.latin1_letters.findall(input)]
-        return chinese + self.latin1
+        self.latin_terms.extend(latin1)
+        return chinese + latin1
 
     def word_filter(self, stemmed_word):
         # type: (unicode) -> bool
@@ -266,7 +267,7 @@ class SearchChinese(SearchLanguage):
         # if not stemmed, but would be too short after being stemmed
         # avoids some issues with acronyms
         should_not_be_stemmed = (
-            word in self.latin1 and
+            word in self.latin_terms and
             len(word) >= 3 and
             len(self.stemmer.stem(word.lower())) < 3
         )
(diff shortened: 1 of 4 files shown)


TimKam:
@DaGeger can you test once again? What do you think, @tk0miya?

sphinx/search/zh.py
```diff
@@ -250,13 +251,24 @@ def split(self, input):
...
         # type: (unicode) -> bool
         return len(stemmed_word) > 1
 
     def stem(self, word):
         # type: (unicode) -> unicode
+
+        # Don't stem Latin words that are long enough to be relevant for search
+        # if not stemmed, but would be too short after being stemmed
+        # avoids some issues with acronyms
+        should_not_be_stemmed =\
```
tk0miya: I feel folding with `\` is a bit ugly. How about using parenthesis like this?
```
should_not_be_stemmed = (
            word in self.latin1 and
            len(word) >= 3 and
            len(self.stemmer.stem(word.lower())) < 3
)
```

tk0miya (approved):
LGTM with nits

tk0miya (changes requested):
I noticed this should be merged into 1.8 branch because this is a bug fix.

tk0miya (approved):
LGTM

tk0miya:
Is this still WIP?

TimKam:
Sorry, yes. I wrote some tests locally and found out it needs more work. Will get back to it at the weekend.

tk0miya:
Okay. Please let me know if you finished. I'll review this again :-)

TimKam:
@tk0miya Could you have another look? I needed to add a list that contains all Latin terms and change the regexp that splits Chinese from Latin characters. @DaGeger comments from you regarding the initial problem are also welcome.

tk0miya (approved):
LGTM!

PR #5613, opened 2018-11-11
Simplify tox configuration with 'extras' option

jdufresne:
https://tox.readthedocs.io/en/latest/config.html#conf-extras

> extras(MULTI-LINE-LIST)
>
> New in version 2.4.
>
> A list of “extras” to be installed with the sdist or develop install.
> For example, extras = testing is equivalent to [testing] in a pip
> install command.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-minversion = 2.0
+minversion = 2.4.0
 envlist = docs,flake8,mypy,coverage,py{35,36,37,38},du{12,13,14}
 
 [testenv]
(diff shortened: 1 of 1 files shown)


tk0miya:
Is there any advantages? I think raising up base version of tox is disadvantage for developers. So -1 for this.

jdufresne:
I think the advantage is that this configuration integrates better with tox. Rather than install Sphinx twice as part of setup process (once without extras, once with extras), it'll be installed only once which is a slightly faster test startup. It also affects the tox output.

It cleans up the TODO comment from @stephenfin. Perhaps they have additional opinions.

[tox 2.4.0](https://tox.readthedocs.io/en/latest/changelog.html#v2-4-0-2016-10-12) was released 2016-10-12. This isn't a bleeding edge feature.

jdufresne:
This also allowed removing the custom [`install_command`](https://tox.readthedocs.io/en/latest/config.html?highlight=install_command#conf-install_command) and fallback to tox/pip defaults.

stephenfin:
I, for one, am in favour of this. The default tox installed in the latest Ubuntu stable, Debian and Fedora versions exceed this requirement. I see no reason to hold ourselves back when this is so widespread.

stephenfin (approved):
Let's do it

tk0miya:
@mitya57 Can you tell me your opinion for this change? I'd like to know this does not effect to package maintainers of distro.

mitya57:
In Debian/Ubuntu we do not use tox, so this does not affect us. I also don't think this will affect other distros.

tk0miya:
@mitya57 thank you for comment.

Okay, let's upgrade it.

tk0miya (approved):

PR #5620, opened 2018-11-12
Deprecate and drop internal use of force_decode()

jdufresne:
In the Python 3 only code base, this function is no longer necessary. The type of values is well understood and deliberate. Code should avoid arbitrary mixing of `bytes` & `str`.

By removing `force_decode()` calls from docstring values, can deprecate the now unused 'encoding' arguments to various autodoc methods.

### DIFF
--- a/sphinx/util/__init__.py
+++ b/sphinx/util/__init__.py
@@ -258,7 +258,7 @@ def save_traceback(app):
     last_msgs = ''
     if app is not None:
         last_msgs = '\n'.join(
-            '#   %s' % strip_colors(force_decode(s, 'utf-8')).strip()  # type: ignore
+            '#   %s' % strip_colors(s).strip()
             for s in app.messagelog)
     os.write(fd, (_DEBUG_HEADER %
                   (sphinx.__display_version__,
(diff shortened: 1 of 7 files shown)


sphinx/util/__init__.py
```diff
@@ -458,6 +458,7 @@ def parselinenos(spec, total):
 def force_decode(string, encoding):
     # type: (unicode, unicode) -> unicode
     """Forcibly get a unicode string out of a bytestring."""
+    warnings.warn('force_decode() is deprecated.', RemovedInSphinx40Warning)
```
tk0miya: Please add `stacklevel=2` parameter to let developers know where this function is called from.

tk0miya (changes requested):
LGTM with nits

jdufresne:
Good call. Added `stacklevel=2`. :slightly_smiling_face:

tk0miya (approved):
LGTM!

kloczek:
Any chance to make soon a new release (4.0.3?) with that commit?

astrojuanlu:
This was merged in 2018, and it's present in Sphinx 4.0.2 as well as other versions.

tk0miya:
I noticed the removal of `force_decode()` harms some projects and people. So it would be better to extend the deprecation period of it until 5.0.

PR #5621, opened 2018-11-12
Fix all BytesWarning identified through tests

jdufresne:
Code should not arbitrarily mix `bytes` & `str` values. The values should be understood and deliberate. When Python is run with the `-b` option, this change fixes warnings of the form:

```
  BytesWarning: str() on a bytes instance
```

### DIFF
--- a/tests/test_setup_command.py
+++ b/tests/test_setup_command.py
@@ -54,8 +54,8 @@ def setup_command(request, tempdir, rootdir):
 def test_build_sphinx(setup_command):
     proc = setup_command.proc
     out, err = proc.communicate()
-    print(out)
-    print(err)
+    print(out.decode())
+    print(err.decode())
     assert proc.returncode == 0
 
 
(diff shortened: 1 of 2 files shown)


tests/test_setup_command.py
```diff
@@ -53,27 +53,21 @@ def setup_command(request, tempdir, rootdir):
...
+    proc.communicate()
     assert proc.returncode == 0
 
 
 @pytest.mark.setup_command('-b', 'html,bar')
 def test_build_sphinx_multiple_invalid_builders(setup_command):
     proc = setup_command.proc
-    out, err = proc.communicate()
-    print(out)
-    print(err)
```
tk0miya, line 76: Why removing these print functions? I think they are useful if tests were broken.

jdufresne:
Makes sense. I've restored the `print()` functions but they now decode the bytes first to avoid the `BytesWarning`.

tk0miya (approved):

tk0miya:
Thank you for update!

Issue #1726, opened 2015-02-14
sphinxcontrib.findanything


DrLulz:
Might someone here know how to fix this? Most appreciative. 

`easy_install sphinxcontrib-findanything`

```
extensions = [
    'sphinx.ext.todo',
    'sphinx.ext.mathjax',
    'sphinxcontrib.googlechart',
    'sphinxcontrib.findanything',
]
```

```
Running Sphinx v1.3b2
Extension error:
Could not import extension sphinxcontrib.findanything (exception: cannot import name any)
```

lehmannro:
Yikes!  It seems sphinxcontrib-findanything is not compatible with Sphinx 1.3+.  Until we get this resolved, you could monkeypatch it like this somewhere in `conf.py`:

```
import sphinx.util.pycompat                                                     
sphinx.util.pycompat.any = any
```

@birkenfeld, @shimizukawa:  I think we need to keep `sphinx.util.pycompat`'s API stable for backwards-compatibility.  What do you think?

shimizukawa:
@lehmannro sounds good.

IMO, we should define API to keep or not keep compatibility.
For a moment, we should recover compatibility on `pycompat.py` with deprecation warning.

lehmannro:
PR coming in a few minutes.

birkenfeld:
Hmm, I wouldn't have considered pycompat to be public API. With the switch from 2to3 to single-source the compatibility infrastructure is expected to change in any case.

However, since you already did the work, I see nothing wrong with offering compatibility with deprecation warnings.

PR #5626, opened 2018-11-13
Deprecate sphinx.util.pycompat.u

jdufresne:
It is now simply a constant equal to the empty string. Provides no further utility.

### DIFF
--- a/sphinx/util/pycompat.py
+++ b/sphinx/util/pycompat.py
@@ -27,7 +27,7 @@ NoneType = type(None)
 # Python 2/3 compatibility
 
 # prefix for Unicode strings
-u = ''
+u = ''  # RemovedInSphinx40Warning
 
 
 # sys_encoding: some kind of default system encoding; should be used with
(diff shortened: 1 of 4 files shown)


sphinx/util/pycompat.py
```diff
@@ -27,7 +27,7 @@
 # Python 2/3 compatibility
 
 # prefix for Unicode strings
-u = ''
+u = ''  # RemovedInSphinx40Warning
```
tk0miya, line 30: At once, we'd used DeprecationWrapper class to emit deprecation warnings for properties.
https://github.com/sphinx-doc/sphinx/blob/1.5/sphinx/util/pycompat.py#L150-L162

I feel this is too much. But deprecation warning is important for 3rd party developers.

@shimizukawa do you have any opinion?

shimizukawa: -0 for deprecation warning. I also feel overly worried too much.
Past discussion is here: https://github.com/sphinx-doc/sphinx/issues/1726#issuecomment-74366880

tk0miya: thanks

tk0miya (approved):
Let's go with this.

PR #5630, opened 2018-11-13
LaTeX: compatibility with caption package

### LINKED ISSUE #5520: Incompatibility with caption package, no hooks to style captions
### Problem
In Sphinx < 1.6, the [caption](https://ctan.org/pkg/caption?lang=en) package could be used to style captions, e.g.:
```latex
\usepackage{caption}
\captionsetup{font=it}
```

Since Sphinx 1.6, including the caption package causes the caption to jump to the right.
sphinx 1.6 with caption package

index.rst:
```rst
.. list-table:: My table

    * - foo
      - foo
    * - foo
      - foo
```

conf.py:
```py
latex_elements = {
    'preamble': '\\usepackage{caption}\n\\captionsetup{font=it}'
}
```

#### Expected results
Either:

- Sphinx is compatible with caption package so that it does not indent/move caption and can be used to style captions
- Or alternatively, allow caption styling using a Sphinx-specific LaTeX command that can be renewed (similar to `\sphinxtablecontinued` and `\sphinxstyleemphasis`)

### Environment info
- Sphinx version: 1.6+
- TexLive version: MacTex 2.17

jfbu:
This compatibility is mainly re-instored for convenience of user to
style the fonts used for the caption, and also possibly influence the
horizontal position via "width" or "margin" option of caption package
(attention that caption package obeys the document class tacit "twoside"
option, so if left and right margins are not set-up to be the same,
positioning of caption will depend on parity of the page number).

Regarding vertical skips, for captions placed on top (which is the table
templates default and also the literalblockcappos default), this commit
ensures that the vertical spacing separating the caption last baseline
to the top of framing is governed by \sphinxbelowcaptionspace in all
these three cases: tabular(y), longtable, literal blocks; the "skip"
option of caption package is ignored for them.

The "skip" is obeyed for the spacing between an image and its caption,
which currently in Sphinx is always below the image (no customization of
the figure caption vertical placement is currently available).

If literalblockcappos is "b" (captions follow code-blocks), this commit
removes the caption-package added \abovecaptionskip, so that "skip" is
also ignored for code-blocks in this case. This looks better due to the
extra space already added by the framing of the code-block and achieves
same spacing as without caption package (presumably loaded mainly to
style the fonts used for the caption). However in future maybe caption's
package "skip" should be obeyed for "literalblock" caption type.

For testing with caption package I mainly used this:

```
   'preamble': r'''
\usepackage{caption}
\captionsetup{justification=RaggedRight,
              margin={20pt,20pt},
              labelfont=bf, textfont=it,
              format=hang,
              singlelinecheck=false}
'''
```

I checked in particular behaviour of captions of code-blocks inside table cells whether 

```
    'sphinxsetup': 'literalblockcappos=b',
```

is made use of or if the default of top positioning of captions for code-blokcs is obeyed.

I made very quick check that hyperlinks to captions still work with `caption` package loaded as above.

### Feature or Bugfix

- mainly Bugfix

Fixes #5520

### DIFF
--- a/sphinx/templates/latex/tabular.tex_t
+++ b/sphinx/templates/latex/tabular.tex_t
@@ -12,6 +12,7 @@
 <%- endif %>
 <% if table.caption -%>
 \sphinxcapstartof{table}
+\sphinxthecaptionisattop
 \sphinxcaption{<%= ''.join(table.caption) %>}<%= labels %>
 \sphinxaftertopcaption
 <% elif labels -%>
--- a/sphinx/templates/latex/tabular.tex_t
+++ b/sphinx/templates/latex/tabular.tex_t
@@ -13,7 +13,7 @@
 <% if table.caption -%>
 \sphinxcapstartof{table}
 \sphinxcaption{<%= ''.join(table.caption) %>}<%= labels %>
-\sphinxaftercaption
+\sphinxaftertopcaption
 <% elif labels -%>
 \phantomsection<%= labels %>\nobreak
 <% endif -%>
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -82,11 +82,13 @@
 % User interface to set-up whitespace before and after tables:
 \newcommand*\sphinxtablepre {0pt}%
 \newcommand*\sphinxtablepost{\medskipamount}%
+% Space from caption baseline to top of table or frame of literal-block
 \newcommand*\sphinxbelowcaptionspace{.5\sphinxbaselineskip}%
 % as one can not use \baselineskip from inside longtable (it is zero there)
 % we need \sphinxbaselineskip, which defaults to \baselineskip
 \def\sphinxbaselineskip{\baselineskip}%
 % These commands are inserted by the table templates
+% A. Table with longtable
 \def\sphinxatlongtablestart
    {\par
     \vskip\parskip
@@ -95,35 +97,51 @@
     \LTpre\z@skip\LTpost\z@skip % set to zero longtable's own skips
     \edef\sphinxbaselineskip{\dimexpr\the\dimexpr\baselineskip\relax\relax}%
    }%
-\def\sphinxatlongtableend{\prevdepth\z@\vskip\sphinxtablepost\relax}%
+% Compatibility with caption package
+\def\sphinxthelongtablecaptionisattop{%
+    \spx@ifcaptionpackage{\noalign{\vskip-\belowcaptionskip}}{}%
+}%
+% Achieves exactly \sphinxbelowcaptionspace below longtable caption
 \def\sphinxlongtablecapskipadjust
-   {\dimexpr-\dp\strutbox-\sphinxbaselineskip+\sphinxbelowcaptionspace\relax}%
-% Now for tables not using longtable
-\def\sphinxattablestart
-   {\par
-    \vskip\dimexpr\sphinxtablepre\relax
-   }%
+   {\dimexpr-\dp\strutbox
+            -\spx@ifcaptionpackage{\abovecaptionskip}{\sphinxbaselineskip}%
+            +\sphinxbelowcaptionspace\relax}%
+\def\sphinxatlongtableend{\prevdepth\z@\vskip\sphinxtablepost\relax}%
+% B. Table with tabular or tabulary
+\def\sphinxattablestart{\par\vskip\dimexpr\sphinxtablepre\relax}%
 \let\sphinxattableend\sphinxatlongtableend
+% This is used by tabular and tabulary templates
 \newcommand*\sphinxcapstartof[1]{%
    \vskip\parskip
    \vbox{}% force baselineskip for good positioning by capstart of hyperanchor
+   % hyperref puts the anchor 6pt above this baseline; in case of caption
+   % this baseline will be \ht\strutbox above first baseline of caption
    \def\@captype{#1}%
    \capstart
-% move back vertically to compensate space inserted by next paragraph
+% move back vertically, as tabular (or its caption) will compensate
    \vskip-\baselineskip\vskip-\parskip
 }%
-% longtable wraps captions to a maximal width of \LTcapwidth
-% so we do the same for tabular(y) tables (see their templates)
-\def\sphinxtablecapwidth{\LTcapwidth}% (default is 4in: max width of caption)
-% TODO: add alignment options? but user can employ caption package instead.
+\def\sphinxthecaptionisattop{% locate it after \sphinxcapstartof
+    \spx@ifcaptionpackage
+      {\caption@setposition{t}%
+       \vskip\baselineskip\vskip\parskip
+       \vskip-\belowcaptionskip\vskip-\dp\strutbox
+      }%
+      {}%
+}%
+\def\sphinxthecaptionisatbottom{% (not finalized)
+    \spx@ifcaptionpackage{\caption@setposition{b}}{}%
+}%
+% The aim of \sphinxcaption is to apply to tabular(y) the maximal width
+% of caption as done by longtable
+\def\sphinxtablecapwidth{\LTcapwidth}%
 \newcommand\sphinxcaption{\@dblarg\spx@caption}%
 \long\def\spx@caption[#1]#2{%
    \noindent\hb@xt@\linewidth{\hss
       \vtop{\@tempdima\dimexpr\sphinxtablecapwidth\relax
 % don't exceed linewidth for the caption width
             \ifdim\@tempdima>\linewidth\hsize\linewidth\else\hsize\@tempdima\fi
-% longtable ignores \abovecaptionskip/\belowcaptionskip, so add hooks here
-% to uniformize control of caption distance to tables
+% longtable ignores \abovecaptionskip/\belowcaptionskip, so do the same here
             \abovecaptionskip\sphinxabovecaptionskip
             \belowcaptionskip\sphinxbelowcaptionskip
             \caption[{#1}]%
@@ -131,14 +149,19 @@
            }\hss}%
    \par\prevdepth\dp\strutbox
 }%
-\def\spx@abovecaptionskip{\abovecaptionskip}% for literal blocks captions
-\newcommand*\sphinxabovecaptionskip{\z@skip}
-\newcommand*\sphinxbelowcaptionskip{\z@skip}
-%
+\def\sphinxabovecaptionskip{\z@skip}% Do not use! Flagged for removal
+\def\sphinxbelowcaptionskip{\z@skip}% Do not use! Flagged for removal
+% This wrapper of \abovecaptionskip is to allow modified setting
+% in case of a literal block inside a table cell.
+% TODO: unify spacing above captions of tables with the one for code-blocks,
+%       this means \abovecaptionskip must be without any effect
+\def\spx@abovecaptionskip{\abovecaptionskip}%
+% Achieve \sphinxbelowcaptionspace below a caption located above a tabular
+% or a tabulary
 \newcommand\sphinxaftertopcaption
-{% this default definition serves with a caption *above* a table, to make sure
- % its last baseline is \sphinxbelowcaptionspace above table top
- \nobreak
+{%
+   \spx@ifcaptionpackage
+     {\par\prevdepth\dp\strutbox\nobreak\vskip-\abovecaptionskip}{\nobreak}%
    \vskip\dimexpr\sphinxbelowcaptionspace\relax
    \vskip-\baselineskip\vskip-\parskip
 }%
@@ -697,13 +720,6 @@
 % ensure same width of caption to all kinds of tables (tabular(y), longtable),
 % because caption package has its own width (or margin) option
        \def\sphinxcaption{\caption}%
-% vertical spacing from last baseline of a top caption to top rule of
-% table will still be \sphinxbelowcaptionspace for tabular/tabulary,
-% (i.e. \sphinxaftertopcaption needs no modification) but we now need
-% this definition for the corrective skip inserted by longtable template:
-       \def\sphinxlongtablecapskipadjust
-          {\dimexpr-\dp\strutbox-\abovecaptionskip+
-                   \sphinxbelowcaptionspace\relax}% this is the vspace we want
       }%
       {\let\spx@ifcaptionpackage\@secondoftwo}%
 }
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -95,6 +95,7 @@
 %      caption is exactly \parskip+\baselineskip+ height of a strut.
 %   c) the caption text will wrap at width \LTcapwidth (4in)
 % - make sure this works also if "caption" package is loaded by user
+%   (with its width or margin option taking place of \LTcapwidth role)
 % TODO: obtain same for caption of literal block: a) & c) DONE, b) TO BE DONE
 %
 % To modify space below such top caption, adjust \sphinxbelowcaptionspace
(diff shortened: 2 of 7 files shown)


sphinx/templates/latex/tabular.tex_t
```diff
@@ -13,7 +13,7 @@
 <% if table.caption -%>
 \sphinxcapstartof{table}
 \sphinxcaption{<%= ''.join(table.caption) %>}<%= labels %>
-\sphinxaftercaption
+\sphinxaftertopcaption
```
jfbu, line 17: This macro was never publicly documented; I changed its name for clarification that it should not be used if caption is moved to after the table body.

sphinx/texinputs/sphinx.sty
```diff
@@ -114,26 +112,30 @@
 % move back vertically to compensate space inserted by next paragraph
    \vskip-\baselineskip\vskip-\parskip
 }%
-% use \LTcapwidth (default is 4in) to wrap caption (if line width is bigger)
-\newcommand\sphinxcaption[2][\LTcapwidth]{%
+% longtable wraps captions to a maximal width of \LTcapwidth
+% so we do the same for tabular(y) tables (see their templates)
+\def\sphinxtablecapwidth{\LTcapwidth}% (default is 4in: max width of caption)
+% TODO: add alignment options? but user can employ caption package instead.
+\newcommand\sphinxcaption{\@dblarg\spx@caption}%
```
jfbu, line 154: `\sphinxcaption` is used in the tabular and tabulary templates, its role is to imitate the longtable caption which has a maximal width (4in per default). The `\sphinxcaption` allowed possibility of optional argument to use something else than `\LTcapwidth`, but this was bad because usually optional argument of `\caption` serves to specify a "short" title for making an entry in the corresponding "\listof<tables, figures>". Notice that `\sphinxcaption` so far is not known to LaTeX writers (figures) but only used via the table templates or via the sphinx.sty code (sphinxVerbatim environment, `\sphinxSetupCaptionForVerbatim`) for literal blocks. So it is still non-public internal tool and this change will not modify any existing project.

sphinx/texinputs/sphinx.sty
```diff
@@ -114,26 +112,30 @@
...
+\long\def\spx@caption[#1]#2{%
    \noindent\hb@xt@\linewidth{\hss
-      \vtop{\@tempdima\dimexpr#1\relax
+      \vtop{\@tempdima\dimexpr\sphinxtablecapwidth\relax
```
jfbu, line 157: This new undocumented macro only serves to implement functioning of  `\sphinxcaption` for maximal width of captions of tabular/tabulary to be same as longtable, because `\sphinxtablecapwidth` by default will simply be the longtable variable `\LTcapwidth`.

sphinx/texinputs/sphinx.sty
```diff
@@ -114,26 +112,30 @@
...
             \abovecaptionskip\sphinxabovecaptionskip
             \belowcaptionskip\sphinxbelowcaptionskip
-            \caption[{#2}]%
+            \caption[{#1}]%
                {\strut\ignorespaces#2\ifhmode\unskip\@finalstrut\strutbox\fi}%
            }\hss}%
    \par\prevdepth\dp\strutbox
 }%
-\def\spx@abovecaptionskip{\abovecaptionskip}
+\def\spx@abovecaptionskip{\abovecaptionskip}% for literal blocks captions
```
jfbu: The title of a literal block caption is set in a minipage (see `\sphinxVerbatim@TitleBox`) <strike>and the insertion of `\abovecaptionskip` from standard expansion of LaTeX `\caption` is without effect, </strike> (wrong comment) so the `sphinxVerbatim` environment does the skip itself on top of a caption which is positioned above a code-block (following standard but somewhat faulty LaTeX behaviour which always uses `\abovecaptionskip` even if caption is on top, although of course it was intended primarily to be used when caption is at bottom; see the `caption` package documentation for more on this incoherency of LaTeX2e). This `\spx@abovecaptionskip` is given meaning `\sphinxverbatimsmallskipamount` inside `sphinxVerbatimintable` to reduce this spacing without modifying `\abovecaptionskip` itself (I don't remember why I preferred `sphinxVerbatimintable` not to give locally new value to `\abovecaptionskip`).

**see** https://github.com/sphinx-doc/sphinx/pull/5630#discussion_r233349050 for corrections to the above

sphinx/texinputs/sphinx.sty
```diff
@@ -114,26 +112,30 @@
...
 \newcommand*\sphinxabovecaptionskip{\z@skip}
 \newcommand*\sphinxbelowcaptionskip{\z@skip}
-
-\newcommand\sphinxaftercaption
+%
+\newcommand\sphinxaftertopcaption
```
jfbu, line 179: As explained above I have modified name of this non-documented LaTeX variable for clarity of future maintenance.

sphinx/texinputs/sphinx.sty
```diff
@@ -683,8 +685,29 @@
...
+% vertical spacing from last baseline of a top caption to top rule of
+% table will still be \sphinxbelowcaptionspace for tabular/tabulary,
+% (i.e. \sphinxaftertopcaption needs no modification) but we now need
+% this definition for the corrective skip inserted by longtable template:
+       \def\sphinxlongtablecapskipadjust
+          {\dimexpr-\dp\strutbox-\abovecaptionskip+
+                   \sphinxbelowcaptionspace\relax}% this is the vspace we want
+      }%
+      {\let\spx@ifcaptionpackage\@secondoftwo}%
+}
```
jfbu, line 743: The LaTeX code comments explain what this does. This is compatibility layer with caption package: if detected, do not use `\sphinxcaption` own code at all and slightly adjust corrective measure for longtable because `caption` package patches longtable code and makes it use an other vertical spacing.

Some other modifications of behaviour in presence of `caption` package are implemented in the `sphinxVerbatim` environement and use the `\spx@ifcaptionpackage` conditional defined here.

sphinx/texinputs/sphinx.sty
```diff
@@ -1049,17 +1072,28 @@
        \vskip\spx@abovecaptionskip
        \def\sphinxVerbatim@Before
            {\sphinxVerbatim@Title\nointerlineskip
-            \kern\dimexpr-\dp\strutbox+\sphinxbelowcaptionspace\relax}%
+            \kern\dimexpr-\dp\strutbox+\sphinxbelowcaptionspace
+                 % if no frame (code-blocks inside table cells), remove
+                 % the "verbatimsep" whitespace from the top (better visually)
+                 \ifspx@opt@verbatimwithframe\else-\sphinxverbatimsep\fi
```
jfbu, line 1112: This removal of some vertical space in case the literal-block is in a table cell (so Sphinx uses no framing but still uses `\sphinxverbatimsep` whitespace on the 4 sides) is unrelated to handling caption package, but it improves the PDF, and it was occasion to do so.

sphinx/texinputs/sphinx.sty
```diff
@@ -1049,17 +1072,28 @@
...
+                 % caption package adds \abovecaptionskip vspace, remove it
+                 \spx@ifcaptionpackage{-\abovecaptionskip}{}\relax}%
```
jfbu, line 1114: If caption package is active and caption is on top, it adds `\abovecaptionskip` *below* the caption, contrarily to standard LaTeX. We remove it for consistency and then only `\sphinxbelowcaptionspace` governs for all four: tabular, tabulary, longtable, and literalblock the spacing to frame of object.

sphinx/texinputs/sphinx.sty
```diff
@@ -1049,17 +1072,28 @@
...
+           \sphinxVerbatim@Title}%
      \fi
      \def\@captype{literalblock}%
      \capstart
      % \sphinxVerbatimTitle must reset color
      \setbox\sphinxVerbatim@TitleBox
             \hbox{\begin{minipage}{\linewidth}%
+     % caption package may detect wrongly if top or bottom, so we help it
+                    \spx@ifcaptionpackage
+                      {\caption@setposition{\spx@opt@literalblockcappos}}{}%
```
jfbu, line 1130: inside the minipage, the caption package fails to detect automatically if caption is before or after. As we know exactly which is the case, we inform caption package here. Its version of `\caption` will thus put `\abovecaptionskip` (default 10pt) *between* the caption and the object, and `\belowcaptionskip` (default 0pt) on the other side.

tests/roots/test-latex-table/expects/table_having_caption.tex
```diff
@@ -4,7 +4,7 @@
 \centering
 \sphinxcapstartof{table}
 \sphinxcaption{caption for table}\label{\detokenize{tabular:id1}}
-\sphinxaftercaption
+\sphinxaftertopcaption
```
jfbu, line 8: Again, I don't think this change of name is breaking. There was no reason to use custom template with `\sphinxaftercaption` elsewhere: if user has custom table template with caption at bottom, there was no reason to use `\sphinxaftercaption` at all.

sphinx/texinputs/sphinx.sty
jfbu: ah sorry

> and the insertion of \abovecaptionskip from standard expansion of LaTeX `\caption` is without effect,

this is plain wrong comment. The `\abovecaptionskip` would work even at start of minipage, but I had forgotten when writing up that comment that by default Sphinx uses not `\caption` but its own `\sphinxcaption` inside the `sphinxVerbatim@TitleBox`, and this sets locally `\abovecaptionskip` to `\sphinxabovecaptionskip` which by default is `0pt`. So the real reason for `\spx@abovecaptionskip` is because `\abovecaptionskip` is set locally to `0pt` inside the TitleBox, when `\sphinxcaption` is made use of. As this is also the case for `sphinxVerbatimintable` it would have remained without effect to set `\abovecaptionskip` to another value there locally, because this is overruled by `\sphinxcaption` (in absence of `caption` package of course, now).

sphinx/texinputs/sphinx.sty
```diff
@@ -1049,17 +1072,28 @@
        \vskip\spx@abovecaptionskip
```
jfbu, line 1106: If caption package is loaded it will add a `\vskip\belowcaptionskip` **above** a *top* caption, as is the case here. By default `\belowcaptionskip` is `0pt` so this changes nothing. If non-zero, there will be extra vertical space above caption. And for a *bottom*  caption, this extra space will be after it. The same will apply to table captions (because Sphinx then does not use its own `\sphinxcaption` but the standard `\caption` as patched by `caption` package). So the behaviour of captions for tables, figures, and literal-blocks remains coherent even then regarding spacing before a top caption or after a bottom caption, so I decided not to add any "corrective" measure here.

sphinx/texinputs/sphinx.sty
jfbu: When one uses `caption` package it succeeds in achieving same spacing for positioning of a top caption for both `table` environment (with a tabular) or `longtable`. But Sphinx does not use `table` environment for a `tabular`, I have observed that in such non-table environment, the `caption` package does not succeed in positioning vertically the caption exactly like it does for longtable. Thus, the necessity remains to keep the `\sphinxaftertopcaption` macro (especially because `parskip` package is loaded) and its way of achieving `\sphinxbelowcaptionspace` vertical spacing. So we also need to achieve this for literal blocks, which is done here.

The situation is thus completely different between the status of `\abovecaptionskip` (used with `caption` package between caption and object) and the one of `\belowcaptionskip` (used on external side if `caption` package used). In the former case Sphinx needs special corrective measures for achieving identical result with tabular(y), longtable and literal blocks, in the latter case, it does not have to worry.

tk0miya (approved):
I don't know caption package so much. But I trust you :-)

jfbu:
Thanks @tk0miya! I got verbose in my comments/review as usual, let's say that this could prove useful in some years if everybody at Sphinx has forgotten why the LaTeX code got like that. I too wasn't too familiar with `caption` package, and in the end I was relieved to confirm little was needed for making it Sphinx compatible (again; but this PR does the additional steps to unify literal-blocks and table looks, which `caption` package by itself of course does not as it does not know about `literal blocks` and the special way Sphinx uses to place their captions) (the latter is complicated due to usage of `framed.sty`: the caption has to be inside, because `framed` always encourages pagebreaks before and after its framing).

I'll wait a bit in case @jessetan has any comment before merging this. It will still be possible to add in future some `\sphinxstylecaptiontext` macros for some amount of customizability with no need of extra package. But in my testing `caption` package works reasonably well, so we can always tell people to use it.

The principle point to improve in Sphinx LaTeX currently regarding captions is placement of figure captions which is not under user control. Tables and literal-blocks are under user control (for tables via the templates).

jessetan:
This solves my reported incompatibility with the captions package, thanks!
Only difference I see is that for a regular table (not longtable) the distance between the caption and the paragraph preceding it is decreased if the caption package is enabled. See screenshots below (put side by side to see more clearly).

Sphinx 1.8.3:
sphinx

Sphinx 1.8.3 with `'preamble': '\\usepackage{caption}'`:
caption

jfbu:
@jessetan thanks, yes you are right. This is due to the fact that `caption` package typesets the caption below an invisible rule, which disrupts TeX baseline placement algorithm.  `longtable` did that already, and Sphinx took measure to ensure same spacing for normal table and longtable. But as `caption` modifies the captions of normal table to act more like captions of longtable, we now have a mismatch.

Investigating this, it reminded me that I had not unified spacing to captions between tables and literal-blocks. They never coincided I think, and I sort of remember working hard during various literal-block evolutions to maintain exact legacy spacing of literal blocks. Anyway, this may be occasion to unify literal blocks with tables. Here is minimal example to play with.

```
Section
=======


Text before |bluearrow|\ |redarrow|

.. list-table:: |showbaseline|

    * - foo
      - foo
    * - foo
      - normal table

Text before |bluearrow|

.. rst-class:: longtable

.. list-table:: |showbaseline|

    * - foo
      - foo
    * - foo
      - longtable

Text before |bluearrow|\ |greenarrow|

.. code-block:: python
   :caption: |showbaseline|

   def foo(x):
       return x

.. |bluearrow| raw:: latex
             
               \textcolor{blue}{\rule{\dimexpr.5\linewidth-1cm-10pt\relax}{1pt}\kern-1pt
               \smash{\rule[-26.4pt]{1pt}{26.4pt}}}

.. |redarrow| raw:: latex
             
               \textcolor{red}{\rule{10pt}{1pt}\kern-1pt
               \smash{\rule[-18.4pt]{1pt}{18.4pt}}}

.. |greenarrow| raw:: latex
             
               \textcolor{green}{\rule{16pt}{1pt}\kern-1pt
               \smash{\rule[-24.4pt]{1pt}{24.4pt}}}

.. |showbaseline| raw:: latex
             
                  \rule{8pt}{1pt}\kern2pt\relax
```

This image was obtained with `caption` package:

capture d ecran 2018-11-17 a 12 14 55

The same source without caption would give the same result with the sole difference that the first table caption would match the blue line, not the red line. (as you reported)

The green line indicates exact distance to caption of code-block. It is the same whether or not `caption` package is loaded. It differs slightly from blue line, but is governed by completely different parameters. I hardcoded in the source above the distances, but I know how they are build up.

I think it could be worthwile to unify Listings with tables (assuming captions are on top), at the same time that the `caption` package issue is fixed.

By the way  `caption` package mistakendly think that the caption is at bottom of table; if it had thought it was on top, the distance would have been different (shorter). Anyway, it is possible to inform `caption` package of the caption position and this will be part of fix. The fix will be something like, if `caption` package is detected, "do for normal tables what we already did for longtables".

jfbu:
@jessetan Updated. I decided to only fix #5520 and not unify with captions of literal blocks, this will be done on another occasion.

*edit*: unfortunately it is still not perfect, there is a slight discrepancy when the caption occupies multiple lines, between table and longtable, with caption package loaded. Looks as if I must dig more into caption package code... If I get time I will do, else I will merge this for 1.8.3. No problem I could see with captions on only one line.

jfbu:
I understood the remaining problem with multi-line captions not behaving exactly same for tabular or longtable if caption package is loaded by user.

I will merge this after testing completes; the vertical spacing above literal blocks as discussed [above](https://github.com/sphinx-doc/sphinx/pull/5630#issuecomment-439610250) is not modified, this PR is for #5520 only.

Issue #5538, opened 2018-10-15
Inheritance diagram extension class names should be resolved with __qualname__, not __name__


KM4YRI:
### Problem

The inheritance diagram extension displays nested class names incorrectly, in the final diagram image.  Nested classes are only displayed as the final class name, instead of the fully qualified path.

Per [PEP 3155](https://www.python.org/dev/peps/pep-3155/), the `__qualname__` attribute is a superior way to resolve a string name of nested classes.

#### Procedure to reproduce the problem

1. Create a new python package called `foo_package`, add `__init__.py`, `doc` folder, etc..
2. Create a new sphinx project for `foo_package` under `/doc/`
2. In `conf.py`, set it up to point to the new package, and add `sphinx.ext.inheritance_diagram` to the `extensions` list.
3. Add the code below into `foo_package` and `index.rst`:

`__init__.py` within `foo_package`:
```
class Foo:
    class Bar:
        pass
    class Baz(Bar):
        pass

```

`index.rst` for `foo_package`:
```
.. inheritance-diagram::
    foo_package.Foo.Bar
    foo_package.Foo.Baz
```

#### Error logs / results

No error / exception / traceback, however the final diagram is incorrect, as the classes will be rendered with the names:

* `foo_package.Bar` instead of `foo_package.Foo.Bar`
* `foo_package.Baz` instead of `foo_package.Foo.Baz`

#### Expected results

Classes in the diagram should be rendered with the fully-qualified name, e.g. nested classes should be joined with '.', e.g. `foo_package.Foo.Baz` instead of `foo_package.Baz`.

### Environment info
- OS: Windows 7
- Python version: 3.5.1
- Sphinx version: 1.8.1 (with a fix applied that prevented the released version from running with 3.5.1)

### Proposed solution

Update the method `class_name()` within `inheritance_diagram.py` to use `__qualname__` instead of `__name__`.

Proposed change to `class_name()` below:

https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/inheritance_diagram.py#L219

```
    def class_name(self, cls, parts=0, aliases=None):
        # type: (Any, int, Optional[Dict[unicode, unicode]]) -> unicode
        """Given a class object, return a fully-qualified name.

        This works for things I've tested in matplotlib so far, but may not be
        completely general.
        """
        module = cls.__module__
        if module in ('__builtin__', 'builtins'):
            fullname = cls.__name__
        else:
[...]

PR #5637, opened 2018-11-15
Fix incorrect handling of nested class names

bwohlberg:
Subject: Fix incorrect handling of nested class names in sphinx.ext.inheritance_diagram and when the autodoc show-inheritance option is enabled

### Feature or Bugfix
- Bugfix

### Purpose
- Correct incorrect handling of nested class names

### Detail
- Names of nested classes are currently incorrect when inheritance details are included

### Relates
- See issue #5538


tk0miya (changes requested):
Thank you for your work. Could you add a testcase for this please?

bwohlberg:
I assumed this would be covered by existing tests since the changes are so minor. Presumably you'd like the new test to confirm that the correct name is now being used for nested classes?

bwohlberg:
@tk0miya: Tests added as requested.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1124,8 +1124,8 @@ def add_directive_header(self, sig):
             self.add_line('', sourcename)
             if hasattr(self.object, '__bases__') and len(self.object.__bases__):
                 bases = [b.__module__ in ('__builtin__', 'builtins') and
-                         ':class:`%s`' % b.__name__ or
-                         ':class:`%s.%s`' % (b.__module__, b.__name__)
+                         u':class:`%s`' % b.__name__ or
+                         u':class:`%s.%s`' % (b.__module__, b.__qualname__)
```
tk0miya, line 1128: u-prefix is no longer needed because we drop all py2 support now.

tests/roots/test-inheritance/dummy/test_nested.py
```diff
@@ -0,0 +1,14 @@
+r"""
```
tk0miya, line 1: Why is this raw prefixed?

tk0miya (approved):
Sorry for very very late response. LGTM with nits. But they are very minor. so I'll work for it after merging.

tk0miya:

PR #5639, opened 2018-11-16
Fix two nits with the TextBuilder class

stephenfin:
- 'logger' was unused and can be removed
- 'TextTranslator.sectionchars' was overridden in '__init__' and doesn't
  need to be defined


tk0miya (approved):
LGTM!

PR #5640, opened 2018-11-16
Simplify ensuredir() with Python3 stdlib features

jdufresne:
- Simplify `ensuredir()` to equivalent `os.makedir(name, exist_ok=True)`
- Do not check if a directory exists before calling `ensuredir()` (`ensuredir()` already handles it)
- Add `exist_ok` argument to `path.makedirs()` to follow same pattern
- Drop unnecessary `.exists()` check immediately before `.isdir()`
- Add tests for `ensuredir()`

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -9,12 +9,15 @@
     :license: BSD, see LICENSE for details.
 """
 
+import os
+import tempfile
+
 import pytest
 from mock import patch
 
 from sphinx.testing.util import strip_escseq
 from sphinx.util import (
-    display_chunk, encode_uri, parselinenos, status_iterator, xmlname_checker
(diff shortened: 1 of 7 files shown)


tests/test_util.py
```diff
@@ -34,6 +37,20 @@ def test_encode_uri():
     assert expected, encode_uri(uri)
 
 
+def test_ensuredir():
+    with tempfile.TemporaryDirectory() as tmp_path:
+        # Does not raise an exception.
```
tk0miya: Could you add comment about purpose of the case like "does not raise an exception for existing directory" please?

tk0miya (changes requested):
LGTM with nits

jdufresne:
Thanks! I have added the suggested comment. Let me know if you any other suggestions.

tk0miya (approved):
Thank you for quick work!

PR #5641, opened 2018-11-16
Replace mock dependency with stdlib unitest.mock

jdufresne:
Mock was added to the stdlib in Python 3.3. Can drop the dependency.

https://docs.python.org/3/library/unittest.mock.html


jdufresne:
Ah, ok. This can't be done until Python 3.5 support is dropped for the `Mock.assert_called()` feature.

PR #5642, opened 2018-11-16
Drop explicit dependency: typed_ast

jdufresne:
typed-ast has been a required dependency of mypy since v0.470. Can rely on it being installed as a transient dependency and not list it explicitly. For the change in mypy, see:

https://github.com/python/mypy/commit/a207754a7e426359812c7ae490ea13c979762b51


tk0miya (approved):

tk0miya:
Thanks!

Issue #5251, opened 2018-08-01
[RFC] LaTeX: provide suitable default font configuration for XeLaTeX/LuaLaTeX


jfbu:
Although LaTeX from recent years with xelatex or lualatex uses by default Latin Modern, issue #5247  shows Sphinx should make a better default configuration of font package. Currently it makes no font choice at all.

I don't know what could be good choice. Here are two proposals:

**Attention:** on many systems XeLaTeX requires `\setmainfont` with filenames when the font are in the texlive repertories. I used font names below, and it worked at my locale because I have there symlinks to places XeLaTeX look at for fonts.

#### Computer Modern Unicode

```diff
diff --git a/sphinx/writers/latex.py b/sphinx/writers/latex.py
index bf2215a51..52685f2af 100644
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -159,7 +159,9 @@ ADDITIONAL_SETTINGS = {
         'polyglossia':  '\\usepackage{polyglossia}',
         'babel':        '',
         'fontenc':      '\\usepackage{fontspec}',
-        'fontpkg':      '',
+        'fontpkg':     ('\\setmainfont{CMU Serif}\n'
+                        '\\setsansfont{CMU Sans Serif}\n'
+                        '\\setmonofont{CMU Typewriter Text}'),
         'utf8extra':   ('\\catcode`^^^^00a0\\active\\protected\\def^^^^00a0'
                         '{\\leavevmode\\nobreak\\ }'),
         'fvset':        '\\fvset{fontsize=auto}',
@@ -169,7 +171,9 @@ ADDITIONAL_SETTINGS = {
         'polyglossia':  '\\usepackage{polyglossia}',
         'babel':        '',
         'fontenc':      '\\usepackage{fontspec}',
-        'fontpkg':      '',
+        'fontpkg':     ('\\setmainfont{CMU Serif}\n'
+                        '\\setsansfont{CMU Sans Serif}\n'
+                        '\\setmonofont{CMU Typewriter Text}'),
         'utf8extra':   ('\\catcode`^^^^00a0\\active\\protected\\def^^^^00a0'
                         '{\\leavevmode\\nobreak\\ }'),
         'fvset':        '\\fvset{fontsize=auto}',
```

Pros: comes with any reasonable TeX installation. Provides coverage of Latin, Greek and Cyrillic scripts. It is mentioned as canonical choice in babel-russian documentation.

[...]

Issue #5248, opened 2018-08-01
LaTeX: Greek letters in section titles disappear from PDF bookmarks


jfbu:
Consider any project with a Greek letter in text, for example α. Sphinx will escape it to `\(\alpha\)` in the LaTeX output.

If this is in a section title, hyperref will then remove it with some warning during PDF build. The result  is that in the PDF bookmark the Greek letter has disappeared.

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.7.x
- TeXLive 2018

Relates #1682

Issue #5247, opened 2018-08-01
LaTeX: PDF does not build with default font config and xelatex/lualatex for Russian language


jfbu:
When one initiates a new project with ``language='ru'`` and ``latex_engine='xelatex'`` the PDF build aborts with an error

```
! Package polyglossia Error: The current roman font does not contain the Cyrill
ic script!
(polyglossia)                Please define \cyrillicfont with \newfontfamily.
```

The problem is that although the Latin Modern font used by default does contain the cyrl script, it is loaded by LaTeX in a way which does not make it available. One must load it explicitely after `\usepackage{fontspec}` with something like this

```
\setmainfont{Latin Modern Roman}
\setsansfont{Latin Modern Sans}
\setmonofont{Latin Modern Mono}
```

then there is no error. 

Thus, this issue is fixed once some decent default setting of ``'fontpkg'`` key is provided in sphinx/writers/latex.py

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.8+
- TL 2018

jfbu:
Hmm, I said

> with something like this

   ```
     \setmainfont{Latin Modern Roman}
     \setsansfont{Latin Modern Sans}
     \setmonofont{Latin Modern Mono}
   ```
>   then there is no error.

But in fact the cyrillic glyph are missing in PDF. This is in LaTeX log:

```
Missing character: There is no П in font Latin Modern Roman 10 Regular/OT:scrip
t=cyrl;language=DFLT;mapping=tex-text;!
Missing character: There is no л in font Latin Modern Roman 10 Regular/OT:scrip
t=cyrl;language=DFLT;mapping=tex-text;!
Missing character: There is no а in font Latin Modern Roman 10 Regular/OT:scrip
t=cyrl;language=DFLT;mapping=tex-text;!
Missing character: There is no н in font Latin Modern Roman 10 Regular/OT:scrip
t=cyrl;language=DFLT;mapping=tex-text;!
```

from `index.rst`:

```
План
```

I don't quite understand as Latin Modern does offer the cyrillic script. No such problem arises with

```
\setmainfont{CMU Serif}
\setsansfont{CMU Sans Serif}
\setmonofont{CMU Typewriter Text}
```

I need to understand why this does not work with Latin Modern...

(see https://github.com/sphinx-doc/sphinx/issues/5251#issuecomment-409742312)

Issue #1682, opened 2015-01-16
Latex writer should not translate Greek unicode, but use textgreek package


callegar:
The current behavior in the latex writer is to translate all Greek letters in latex math. For instance, a Greek Δ becomes `\(\Delta\)`.

This creates issues when Greek text appears in a section title, because hyperref has trouble in creating bookmarks including math.  Furthermore, it breaks formatting because it make it impossible to have the Greek test follow the current text style with respect to things like italics, bold, etc.  Most important, it is completely unnecessary, since pdftext supports unicode just fine given that you are using a unicode `\textinput`. To this aim it is only necessary to load the `\textgreek` package with a suitable font... e.g. `\usepackage[artemisia]{textgreek}` and to pass the `unicode` option to `hyperref`.

jfbu:
Thanks for report and suggestion, and sorry for long delay before actually doing something... I think this has now been fixed in #5645 for future Sphinx 2.0. See also http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-latex_elements, the discussion of ``'fontpkg'``, ``'fontenc'``, and ``'textgreek'`` key.

Closing now.

PR #5645, opened 2018-11-16
LaTeX: support for Greek and Cyrillic

jfbu:
0. do not escape Unicode Greek letters via LaTeX math mark-up: pass them
   through un-modified to LaTeX document,

1. if "fontenc" receives extra option LGR, then pdflatex will support
   Unicode Greek letters (not in math), and with extra option T2A it
   will support (most) Unicode Cyrillic letters.

2. for pdflatex with LGR, this will use "textalpha" LaTeX package and
   "substitutefont" package to set up some automatic font substitution
   to work around the unavailability of Greek with "times"
   package (which is default font package chosen by Sphinx for
   pdflatex), same with T2A and "substitutefont" for Cyrillic.

3. for xelatex/lualatex, set up Computer Modern Unicode as default font,
   as it supports Cyrillic and Greek scripts,

4. for platex, don't do anything special as the engine already has
   its default font supporting Cyrillic and Greek (even in math mode!)

**edit**

The final version differs from the description above:

- for xelatex/lualatex use FreeFont (because on Ubuntu xenial, CMU requires the big package `texlive-fonts-extra`, and FreeFont has its own separate `fonts-freefont-otf`)

- also X2 encoding is handled for pdflatex, and the fonts are the base ones in LGR or X2 or T2A encoding, providedon Ubuntu xenial from `cm-super` package (and `texlive-lang-greek`,  `texlive-lang-cyrillic`). These extra dependencies are optional only, to support the new feature of support of Greek and Cyrillic in European language document.

TODO: indexing with xindy has merge-rules for Greek, but for Cyrillic no, because they are loaded only for Cyrillic documents, so indexing in non-Cyrillic document of Cyrillic words will work but all such words will end-up in "non-alphabetical" group.

Closes: #5251
Fixes:  #5248
Fixes:  #5247
Closes: #1682


CHANGES
```diff
@@ -4,7 +4,41 @@ Release 2.0.0 (in development)
...
+     * - cm-lgc
+       - texlive-fonts-extra
+       - Cyrillic letters (in non-Cyrillic documents)
+
+  These extra package are not required by default. The first two are needed if
+  the :confval:`latex_elements`.\ ``'fontenc'`` key has been modify to declare
+  the use of the ``LGR`` (Greek) and/or ``T2A`` (Cyrillic) font encoding. Even
+  then, the last four are font packages arising in the default value for
+  :confval:`latex_elements`.\ ``'fontpkg'``, and may be replaced by other font
+  packages providing ``LGR`` and/or ``T2A`` support.
```
jfbu: These explanations may move later on to builders/index.rst from CHANGES, after 2.0 release

CHANGES
```diff
@@ -15,6 +49,11 @@ Incompatible changes
   has been longly used as default of sphinx-quickstart.
 * LaTeX: Move message resources to ``sphinxmessage.sty``
 * LaTeX: Stop using ``\captions<lang>`` macro for some labels
+* LaTeX: Greek letters in text are not escaped to math mode mark-up, and they
+  will use the text font not the math font. If (and only if) the document
+  contains such Greek Unicode letters *and* the :confval:`latex_engine` is
+  ``'pdflatex'`` then the :confval:`latex_elements`.\ ``'fontenc'`` key
+  **must** be used to declare usage of the ``LGR`` font encoding.
```
jfbu: Initially I added LGR and usage of packages `textalpha` and `substitutefont` to the default pdflatex setting so that no document would need any change for transition to Sphinx 2.0. Unfortunately the `textalpha` is on Ubuntu xenial available via `texlive-lang-greek` which wasn't previous dependency.

So after some hesitation I decided on the contrary to not change anything to the default `'fontenc'` but this means that if a document had a Greek Unicode in text, which previously was escaped to math mark-up, now, what will happens is that project author must explicitely use `latex_elements` `'fontenc'`  key to add usage of ``LGR`` encoding. And the TeX installation might need additional LaTeX packages (listed above).

CHANGES
```diff
@@ -58,10 +97,18 @@ __ https://github.com/sphinx-contrib/sphinx-pretty-searchresults
 * #4018: htmlhelp: Add :confval:`htmlhelp_file_suffix` and
   :confval:`htmlhelp_link_suffix`
 * #5559: text: Support complex tables (colspan and rowspan)
+* LaTeX: support rendering (not in math, yet) of Greek and Cyrillic Unicode
```
jfbu, line 75: it is possible via `alphabeta` package to support Greek Unicode literals in math directive with `pdflatex`, but the problem is that an upright Unicode alpha will be mapped to `\alpha` and rendered by italic math font, contrarilty to MathJax rendering, which will keep separate the upright and the italic. Anyway there is no issue yet about this on your tracker, and besides, only xelatex/lualatex are reasonable for supporting Unicode input in math, it is not worthwile to go to extreme lengths to try to support it with pdflatex.

I was surprised to discover that platex (Japanese) has all set-up and individual Greek and Cyrillic letters work fine both in text mode and in math mode...

CHANGES
```diff
@@ -58,10 +97,18 @@ __ https://github.com/sphinx-contrib/sphinx-pretty-searchresults
...
+  letters in non-Cyrillic document even with ``'pdflatex'`` as
+  :confval:`latex_engine`
 
 Bugs fixed
 ----------
 
+* #5247: LaTeX: PDF does not build with default font config for Russian
+  language and ``'xelatex'`` or ``'lualatex'`` as :confval:`latex_engine`
+  (refs: #5251)
+* #5248: LaTeX: Greek letters in section titles disappear from PDF bookmarks
```
jfbu, line 85: There still remains #5249, but see previous comment, user should choose xelatex and add usage of `unicode-math` package for this.

doc/usage/configuration.rst
```diff
@@ -2043,17 +2043,48 @@ information.
...
+             engine) to do:
+
+             .. code-block:: latex
+
+                \substitutefont{LGR}{\rmdefault}{artemisia}
+                \substitutefont{LGR}{\sfdefault}{neohellenic}
+                \substitutefont{LGR}{\ttdefault}{cmtt}
+                \substitutefont{T2A}{\rmdefault}{fcm}
+                \substitutefont{T2A}{\sfdefault}{fcs}
+                \substitutefont{T2A}{\ttdefault}{fct}
```
jfbu: the T2A fonts are from the cm-lgc LaTeX font package

doc/usage/configuration.rst
```diff
@@ -2043,17 +2043,48 @@ information.
...
+
+             For this however, the ``'fontenc'`` key must be used to tell
+             LaTeX to load the ``LGR`` (Greek) or ``T2A`` (partial Cyrillic)
+             font encoding. If ``'fontenc'`` is not modified the above lines
+             are not executed.
```
jfbu: as explained in previous comment, in earlier version of this PR, I had to the contrary automatically added LGR and T2A to `fontenc` loading. But to support this fully with the above default set-up this meant requiring Greek and Cyrillic fonts for all Sphinx projects with ``'pdflatex'`` engine, even those not having usage for Greek and Cyrillic Unicode letters which is almost the totality of existing projects: Cyrillic was not supported (except of course for Cyrillic documents), and Greek was via escaping to math mark-up but I think few people must have used this.

doc/usage/configuration.rst
```diff
@@ -2043,17 +2043,48 @@ information.
...
+             In a custom ``'fontpkg'`` setting, do not use ``\substitutefont``
+             with a font encoding not also declared via ``'fontenc'``.
+
+           - For ``'xelatex'`` and ``'lualatex'``, the default is
+             ``'\\setmainfont{CMU Serif}'`` (and similar for sans
+             serif and monospace) . This OpenType font family supports
+             both Cyrillic and Greek scripts (contrarily to the
+             default font configured by LaTeX for ``xelatex/lualatex``
+             if ``'fontpkg'`` is left to empty string, as was the case
+             prior to 2.0).
```
jfbu: This is to close #5251 and fix #5247

doc/usage/configuration.rst
```diff
@@ -2043,17 +2043,48 @@ information.
...
+
+           - ``'platex'`` (Japanese documents) engine supports individual
+             Cyrillic and Greek letters with no need of extra user set-up.
```
jfbu, line 2087: I did not know that, but this is what my brief testing showed to me...

doc/usage/configuration.rst
```diff
@@ -2130,13 +2161,69 @@ information.
...
+
+     ``'textgreek'``
+        The default (``'pdflatex'`` only) is
+        ``'\\usepackage{textalpha}'``, but only if ``'fontenc'`` was
+        modified by user to include ``LGR`` option.  If not, the key
+        value will be forced to be empty string.
+
+        This is needed for ``pdfLaTeX`` to support Unicode input of Greek
+        letters such as φύσις. Expert users may want to load the ``textalpha``
+        package with its option ``normalize-symbols``.
```
jfbu, line 2215: the only reason for documenting this key is to allow usage with this extra option

sphinx/texinputs/sphinx.sty
```diff
@@ -361,6 +363,28 @@
...
+    \expandafter\DeclareTextSymbolDefault\expandafter
+               {\csname cyr\@tempa\endcsname}{T2A}%
+    \expandafter\uppercase\expandafter{\expandafter
+                 \def\expandafter\@tempa\expandafter{\@tempa}}%
+    \expandafter\DeclareTextSymbolDefault\expandafter
+      {\csname CYR\@tempa\endcsname}{T2A}%
+  }%
+  \DeclareTextSymbolDefault{\CYRpalochka}{T2A}%
+\fi
+
```
jfbu: this fancy LaTeX code is what is needed for LaTeX to automatically use T2A font encoding when encountering a Cyrillic letter.

sphinx/writers/latex.py
```diff
@@ -64,6 +64,23 @@
...
+  \substitutefont{LGR}{\rmdefault}{artemisia}   % gfsartemisia
+  \substitutefont{LGR}{\sfdefault}{neohellenic} % gfsneohellenic
+  \substitutefont{LGR}{\ttdefault}{cmtt}        % cbfonts
+\fi
+\expandafter\ifx\csname T@T2A\endcsname\relax
+\else
+% T2A was declared as font encoding
+  \substitutefont{T2A}{\rmdefault}{fcm}
+  \substitutefont{T2A}{\sfdefault}{fcs}
+  \substitutefont{T2A}{\ttdefault}{fct}
```
jfbu: this would be bad if T2A is main font encoding; but this can only happen if user is building a project in a Cyrillic language, and then certainly ``'fontpkg'`` key is used, because "times" package has no Cyrillic support with pdflatex. Then this default is overwrittent by user setting hence these font substitution are not executed. This is why I did not add a test to check if T2A is the main font encoding (`\LastDeclaredEncoding` could have been used in a latex conditional)

sphinx/writers/latex.py
```diff
@@ -86,7 +103,9 @@
     'multilingual':    '',
     'babel':           '\\usepackage{babel}',
     'polyglossia':     '',
-    'fontpkg':         '\\usepackage{times}',
+    'fontpkg':         PDFLATEX_DEFAULT_FONT_PKG,
```
jfbu: because it is cumbersome to add facny latex code here in this dictionary

sphinx/writers/latex.py
```diff
@@ -86,7 +103,9 @@
...
+    'substitutefont':  '',
+    'textgreek':       '\\usepackage{textalpha}',
```
jfbu, line 140: perhaps `'textalpha'` is better name for this key...

sphinx/writers/latex.py
```diff
@@ -121,29 +140,28 @@
...
+                        '    \\def\\sphinxDUC#1{\\DeclareUnicodeCharacter{"#1}}\n'
+                        '  \\else\n'
+                        '    \\let\\sphinxDUC\\DeclareUnicodeCharacter\n'
+                        '  \\fi\n'
+                        '  \\sphinxDUC{00A0}{\\nobreakspace}\n'
+                        '  \\sphinxDUC{2500}{\\sphinxunichar{2500}}\n'
+                        '  \\sphinxDUC{2502}{\\sphinxunichar{2502}}\n'
+                        '  \\sphinxDUC{2514}{\\sphinxunichar{2514}}\n'
+                        '  \\sphinxDUC{251C}{\\sphinxunichar{251C}}\n'
+                        '  \\sphinxDUC{2572}{\\textbackslash}\n'
```
jfbu, line 185: this is only refactoring, for better readability of what the code does

sphinx/writers/latex.py
```diff
@@ -121,29 +140,28 @@
...
     },
     'xelatex': {
         'latex_engine': 'xelatex',
         'polyglossia':  '\\usepackage{polyglossia}',
         'babel':        '',
         'fontenc':      '\\usepackage{fontspec}',
-        'fontpkg':      '',
+        'fontpkg':     ('\\setmainfont{CMU Serif}\n'
+                        '\\setsansfont{CMU Sans Serif}\n'
+                        '\\setmonofont{CMU Typewriter Text}'),
```
jfbu: These fonts are visually close to previous default with `xelatex`:  previous default was LaTeX choice i.e. Latin Modern OpenType, which is not far from original Knuth Computer Modern. The CMU Serif is the Unicode version of Computer Modern, it exists in serif, sans serif, monospace and supports both Greek and Cyrillic.

Alternative is Libertinus, which is more Times-like (derived from Linux Libertine) but this is more recent font, currently still under development recently.

sphinx/writers/latex.py
```diff
@@ -552,6 +575,15 @@ def __init__(self, document, builder):
                            builder.config.language)
 
         # set up multilingual module...
+        if self.elements['latex_engine'] == 'pdflatex':
+            if ('T2A' in self.elements['fontenc'] and
+                not self.babel.uses_cyrillic()):
+                self.elements['substitutefont'] = '\\usepackage{substitutefont}'
+                self.elements['sphinxpkgoptions'] += ',cyrnocyr'
```
jfbu: I prefered to move indeed to `sphinx.sty` the fancy latex, rather than have it in template

sphinx/writers/latex.py
```diff
@@ -552,6 +575,15 @@ def __init__(self, document, builder):
...
+            if 'LGR' in self.elements['fontenc']:
+                self.elements['substitutefont'] = '\\usepackage{substitutefont}'
```
jfbu, line 617: Sphinx does not support Greek as main language I think, so it makes sense not to worry about LGR being *last*  font encoding in ``'fontenc'`` key.

sphinx/writers/latex.py
```diff
@@ -552,6 +575,15 @@ def __init__(self, document, builder):
...
+            else:
+                self.elements['textgreek'] = ''
```
jfbu, line 619: This way no dependency is added to existing Sphinx projects only needing T1 encoding (i.e. all (non-Cyrillic) projects which did not have Unicode Greek letters in their text, previously rendered via TeX math font)

jfbu:
@tk0miya sorry for flake8, but for some time now I can't execute the tests at my locale because I use Anaconda Python distribution and it has ceased supporting my system version of Mac OS X, and flake8 simply refuses to work.

As per circleci, the build succeeds with lualatex (showing CMU Serif is available) and fails with xelatex which can not find the font, I will try to modify it to call the font by filename but this is more cumbersome. <strike>I wonder if testing is done on Mac OS ?</strike>

I had forgotten xetex can't find TeXLive fonts by font names, it needs filenames, because at my locale I have symlinks to work around that problem.

CHANGES
```diff
@@ -4,7 +4,44 @@ Release 2.0.0 (in development)
...
+       - Greek letters (in text, not math)
+     * - gfsneohellenic
+       - fonts-gfs-neohellenic (texlive-fonts-extra)
+       - Greek letters (in text, not math)
+     * - cbfonts
+       - texlive-lang-greek
+       - Greek letters (in text, not math)
+     * - cm-lgc
+       - texlive-fonts-extra
+       - Cyrillic letters (in non-Cyrillic documents)
```
jfbu: They are alternative which may be better: `texlive-fonts-extra` appears to be rather big [Ubuntu/Debian package](https://packages.ubuntu.com/xenial/texlive-fonts-extra), which installs many many TeX fonts. Thus, [`texlive-lang-cyrillic`](https://packages.ubuntu.com/xenial/texlive-lang-cyrillic) is much smaller and  provides some Computer Modern fonts in "cyrillic" core LaTeX bundle which is not `cm-lgc`. They are available in X2 encoding which covers even more Cyrillic glyphs. I must check if in PDF they are rendered as scalable fonts. (I expect so)

It would be logical to tell  people: for support of occasional Greek, you will need `texlive-lang-greek` and for support of occasional Cyrillic `texlive-lang-cyrillic` but not require them the big `texlive-fonts-extra`.

For xelatex/lualatex, this is different, I have no qualms about requiring full up-to-date installation if necessary. But for vast majority of projects using pdflatex it makes sense to try not to add big dependencies.

doc/conf.py
```diff
@@ -56,10 +56,17 @@
                     'Georg Brandl', 'manual', 1)]
 latex_logo = '_static/sphinx.png'
 latex_elements = {
+    'fontenc': r'\usepackage[LGR,T2A,T1]{fontenc}',
```
jfbu: I will probably switch to X2 in place of T2A to cover even more of Cyrillic Unicode block.

doc/usage/builders/index.rst
```diff
@@ -164,8 +164,11 @@ The builder's "name" must be given to the **-b** command-line option of
    * ``texlive-latex-recommended``
    * ``texlive-fonts-recommended``
    * ``texlive-latex-extra``
+   * ``texlive-fonts-extra``, ``texlive-lang-greek`` (if needed to
```
jfbu: It would be much better to say here `texlive-fonts-cyrillic` in place of `texlive-fonts-extra`

doc/usage/configuration.rst
jfbu: I think I should replace the last three lines with `\substitutefont{X2}{\rmdefault}{cmr}` etc.. because `x2cmr.fd` exists in [`texlive-lang-cyrillic`](https://packages.ubuntu.com/xenial/texlive-lang-cyrillic), and covers more Cyrillic than T2A encoded font files.

doc/usage/configuration.rst
jfbu: TODO: `\setmainfont{CMU Serif}` syntax works only with lualatex, for xelatex one needs in fact filename, as I did in latex.py. Update doc.

sphinx/texinputs/sphinx.sty
jfbu: I will probably replace it by more extensive code covering X2 which is superset of T2A for Cyrillic. And [tex.sx answer](https://tex.stackexchange.com/a/460325/4686) did already the work, in a more agreeable syntax using Unicode letters. I think I should externalize this to a sphinxcyrillic.sty file probably, to keep sphinx.sty simpler. Besides with a separate sphinxcyrillic.sty file, its loading can be done from template rather than pass an option to sphinx.

sphinx/texinputs/sphinx.xdy
```diff
@@ -147,40 +147,61 @@
...
+(merge-rule "\IeC {\textPi }" "Π" :string)
+(merge-rule "\IeC {\textRho }" "Ρ" :string)
+(merge-rule "\IeC {\textSigma }" "Σ" :string)
+(merge-rule "\IeC {\textTau }" "Τ" :string)
+(merge-rule "\IeC {\textUpsilon }" "Υ" :string)
+(merge-rule "\IeC {\textPhi }" "Φ" :string)
+(merge-rule "\IeC {\textChi }" "Χ" :string)
+(merge-rule "\IeC {\textPsi }" "Ψ" :string)
+(merge-rule "\IeC {\textOmega }" "Ω" :string)
+(merge-rule "\IeC {\textohm }" "Ω" :string)
```
jfbu, line 204: all these merge rules do nothing bad with xelatex/lualatex where Unicode letters remain identical to themselves when written to file; they serve only with pdflatex because the `.idx` file will have those macros rather than the Unicode letters. So we map the macros back to the Unicode letters.

sphinx/writers/latex.py
jfbu: I will change that presumably to X2 with `cmr`, so that only `texlive-lang-cyrillic` is needed not the much bigger `texlive-fonts-extra`.

sphinx/writers/latex.py
```diff
@@ -121,29 +140,37 @@
...
+                        '  ItalicFont = cmunti.otf,\n'
+                        '  BoldItalicFont = cmunbi.otf]\n'
+                        '\\setsansfont{cmunss.otf}[\n'
+                        '  BoldFont = cmunsx.otf,\n'
+                        '  ItalicFont = cmunsi.otf,\n'
+                        '  BoldItalicFont = cmunso.otf]\n'
+                        '\\setmonofont{cmuntt.otf}[\n'
+                        '  BoldFont = cmuntb.otf,\n'
+                        '  ItalicFont = cmunit.otf,\n'
+                        '  BoldItalicFont = cmuntx.otf]'),
```
jfbu: Well what a pain that is. On my MacOS, the problem did not arise because I used symlinks from font repertory `~/Library/Fonts` to the TeXLive location, so XeTeX knows how to find fonts per font name, not filename, so in initial commit I had done for xelatex as for lualatex.

jfbu: Unfortunately this seems to require [`texlive-fonts-extra`](https://packages.ubuntu.com/search?searchon=contents&keywords=cmunrm.otf&mode=exactfilename&suite=xenial&arch=any). But one may expect from Sphinx xetex/lualatex users complete up-to-date TeX installations, so perhaps this is acceptable. Libertinus font is more recent yet.

sphinx/writers/latex.py
jfbu: As said above I think now the fancy latex should go into some `sphinxcyrillic.sty`.

jfbu (commented):
While working on this PR, I did not pay attention to the dependencies on say Ubuntu Xenial. Indeed I mainly work on my own Mac OS X with a (quasi-full) TeXLive installation.

I feel I need to think a bit more about minimizing dependencies resulting from choice of fonts. For most Sphinx users wishing to use some Greek or Cyrillic exceptionally in some non-Cyrillic document, important is that the Unicode letter shows in PDF, not that the font is well-matched with the font for Latin text.

CHANGES
jfbu: The X2 encoded Computer Modern seem to require not only `texlive-lang-cyrillic` but also [`cm-super-minimal`](https://packages.ubuntu.com/xenial/cm-super-minimal) (because PDF needs subsetting in particular font file `sfrm1000.pfb`, if I look at log of `pdflatex` at my locale). Investigating if `cm-super-minimal` is enough or `cm-super` is needed.

tk0miya (approved):

CHANGES
jfbu: >  Investigating if cm-super-minimal is enough or cm-super is needed.

Yes, cm-super-minimal is enough, but `\usepackage[10pt]{type1ec}` must be used before `\usepackage[...]{fontenc}`. From the README.debian of the `cm-super-minimal` package, 

>  If you DON'T have cm-super (the full package) installed you have to use
>       \usepackage[10pt]{type1ec}
>  which uses only those cm-super fonts available in cm-super-minimal and
>  scale those fonts for other design sizes.

and I confirmed by looing at contents of `type1ec.sty` file to see what it does.

jfbu:
I am not sure the testing includes building out document to PDF: if it does, the `cm-super` package is required now.

In my last commit I have

- tried to remove the clutter and be less verbose in explanations,

- decided for FreeFont as default choice of fonts for the Unicode engines,

- added support to pdflatex for X2-encoded Cyrillic which covers a lot of Cyrillic glyphs

- reduced extra requirements on Ubuntu xenial to a minimum: `texlive-fonts-extra` never required, only individual packages such as `fonts-freefont-otf` or `cm-super(-minimal)`. And these requirements are optional, only to support previously non existing support of Greek/Cyrillic in English or European languages documents.

jfbu:
Testing completed. I will push one last commit to avoid using same option prefix for sphinx.sty and (new) sphinxcyrillic.sty and then I will merge.

jfbu:
@tk0miya merged! thanks for reviewing, I since externalized to new sphinxcyrillic.sty, fixed up font choice for xelatex/lualatex, clarified the Ubuntu package dependencies and improved the documentation.

jfbu:
@tk0miya It seems our LaTeX testing does not include building our PDF documentation `sphinx.pdf` (I agree this takes time,  less now thanks to your work on speeding up the gathering of references phase). MEMO: to build `sphinx.pdf` this PR adds the following requirements `texlive-lang-greek`, `texlive-lang-cyrillic`, and `cm-super` (because Greek and Cyrillic is used for demonstration in document). <strike>`fonts-latex-extra`</strike> `texlive-fonts-extra` is *not* required.

tk0miya:
At present, we only do testing to build testroot document to PDF, not our document.

On the other hand, PDF is always built in readthedocs.org on every commit. I know it's not testing, but might be enough.

jfbu:
ah yes, and RTD build succeeded. As per the requirements for our continuous testing I see they are changed e.g. at circleci via a docker file `sphinxdoc/docker-ci`? I was worried I could not modify it. What is needed for LaTeX PDF builds if one tests the new features of this PR is `texlive-lang-greek`, `texlive-lang-cyrillic`, `cm-super`. For xelatex/lualatex, `fonts-freefont-otf`. The (big) Ubuntu package `texlive-fonts-extra` is not needed. 

BTW i need to do s/fonts-latex-extra/texlive-fonts-extra/g everywhere... (I got mixed up with Ubuntu package names, and I am not on Ubuntu at my locale...)

*edit*: actually I mixed up only in my comments, not in this PR

tk0miya:
Yes, Circle CI uses `sphinxdoc/docker-ci` image (https://hub.docker.com/r/sphinxdoc/docker-ci/).
And it comes from https://github.com/sphinx-doc/docker-ci .
Circle CI has always fetch our image on each testing. So it would be nice if we can make it slim :-)

jfbu:
@tk0miya *afaict*, `texlive-fonts-extra` (in https://github.com/sphinx-doc/docker-ci, which is a big dependency) could be replaced by
`fonts-freefont-otf` (to support xelatex and lualatex builds on master branch), and `fonts-lmodern` (to support them on 1.8 branch). (I think no platex builds are done)

*edit* I see `lmodern` (which includes `fonts-lmodern`) is already listed as dependency (the dependencies are in alphabetical order but for some reason the name of this one does not start with `texlive-` so I missed it). After 2.0 release it will become superfluous I think and `fonts-freefont-otf` will be enough.

If we add some Greek or Cyrillic letters to the testroot (still with language English), then it would be needed to add `texlive-lang-greek`, `texlive-lang-cyrillic` and `cm-super` for the PDFLaTeX build of testroot to succeed. But this is a bit redundant because RTD build of our own documentation tests this now.

tk0miya:
We should test for Greek and Cyrillic at least once. And I think we've already done it on rtd. So it's okay to keep as is.

PR #5653, opened 2018-11-21
Replace six.string_types with native str

jdufresne:


sphinx/ext/napoleon/docstring.py
```diff
@@ -132,8 +130,8 @@ def __init__(self, docstring, config=None, app=None, what='', name='',
         self._name = name
         self._obj = obj
         self._opt = options
-        if isinstance(docstring, string_types):
-            docstring = docstring.splitlines()
+        if isinstance(docstring, str):
+            docstring = docstring.splitlines()  # type: ignore
```
tk0miya: How about like this?
```
if isinstance(docstring, str):
    self._lines = docstring.splitlines()
else:
    self._lines = docstring
```

I guess this change removes the `type: ignore` comment.

sphinx/testing/fixtures.py
```diff
@@ -103,7 +103,7 @@ def test_params(request):
     result.update(kwargs)
 
     if (result['shared_result'] and
-            not isinstance(result['shared_result'], string_types)):
+            not isinstance(result['shared_result'], str)):
```
tk0miya: I think this change allows not to fold the condition of if-clause.

sphinx/util/inspect.py
```diff
@@ -381,7 +381,7 @@ def format_args(self):
                               param.KEYWORD_ONLY):
                 arg.write(param.name)
                 if param.annotation is not param.empty:
-                    if isinstance(param.annotation, string_types) and \
+                    if isinstance(param.annotation, str) and \
                             param.name in self.annotations:
```
tk0miya: This is also unfold-able.

tk0miya (changes requested):
LGTM with nits

sphinx/ext/napoleon/docstring.py
jdufresne: I made this change, but there is still a mypy error on the `self._lines = docstring` line:

```
sphinx/ext/napoleon/docstring.py: note: In member "__init__" of class "GoogleDocstring":
sphinx/ext/napoleon/docstring.py:136:13: error: Incompatible types in assignment (expression has type "Union[unicode, List[unicode]]", variable has type "List[str]")
```

I think the `unicode`/`str` issues can be resolved once mypy can be passed the `--python-version` argument. 

Without that, resolving the above error would require checking a different type on the `isinstance` line (such as `six.text_type`). What do you prefer? Ignoring the type check or keeping the `six` import?

jdufresne: `._lines` is never actually accessed as an instance variable. So I've converted it to a local variable.

jdufresne:
I've made the suggested edits. All additional feedback welcome. :slightly_smiling_face:

tk0miya (approved):
LGTM!

PR #5656, opened 2018-11-22
Deprecate evaluating Python 2 configuration files

jdufresne:
Converting Python 2 configuration files to Python 3 is now deprecated. This is the only internal use of
`sphinx.util.pycompat.convert_with_2to3()`, so that is now also deprecated.

### DIFF
--- a/sphinx/util/pycompat.py
+++ b/sphinx/util/pycompat.py
@@ -16,9 +16,15 @@ from textwrap import indent  # type: ignore # NOQA
 
 from six import text_type
 
+from sphinx.locale import __
+from sphinx.util import logging
+
 if False:
     # For type annotation
     from typing import Any, Callable, Generator  # NOQA
+
+
+logger = logging.getLogger(__name__)
 
 
 NoneType = type(None)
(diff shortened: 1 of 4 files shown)


sphinx/util/pycompat.py
```diff
@@ -79,11 +83,12 @@ def execfile_(filepath, _globals, open=open):
...
-            # 3.x syntax using 2to3
-            source = convert_with_2to3(filepath)
-            code = compile(source, filepath_enc, 'exec')
-        else:
-            raise
+        # maybe the file uses 2.x syntax; try to refactor to
+        # 3.x syntax using 2to3
+        source = convert_with_2to3(filepath)
+        code = compile(source, filepath_enc, 'exec')
+        warnings.warn(
```
tk0miya: Could you emit a warning for users instead using `logger.warning()` please?
It must be show a warning for users first before deprecating API.
We have to keep compatibility during 2.x (or 3.x).
So this helper function should not be deprecated during 2.x.

tk0miya (changes requested):

sphinx/util/pycompat.py
jdufresne: Sure thing. That makes sense. For user facing warnings, should the feature be removed in 3.x or 4.x?

tk0miya: Personally, removal in 3.x is a bit early. Because Sphinx-3.0 would be released at one year after 2.0 release. So I'd like to vote to removal in 4.0.

jdufresne:
I have converted the warning to `logger.warning()`. Let me know what you think. All additional feedback is welcome.

tk0miya (approved):
LGTM!

tk0miya:
Merged. Thank you always :-)

PR #5657, opened 2018-11-22
Make it possible to construct Sphinx twice with warningiserror=True

mitya57:
Subject: Make it possible to construct Sphinx application twice with `warningiserror=True`

### Feature or Bugfix
Bugfix

### Purpose
Currently you can only construct the Sphinx application object once, if the `warningiserror=True` argument is used. When you try to construct it again, you get error about directives already registered. However it may be useful to construct it multiple times for testing.

For example, many projects are using [sphinx-testing](https://github.com/sphinx/sphinx-testing) framework. And their tests fail with the current Sphinx, see e.g. mcmtroffaes/sphinxcontrib-bibtex#149.

This pull request makes it possible to construct the application twice if only the standard extensions are available. If `conf.py` enables more extensions it may fail because of similar constructs in other extensions.

An alternative approach would be for register functions to raise warnings only when the directive/role object being registered differs from the already registered one. This works fine for VersionChange or AutodocDirective classes, but does not work for MathReferenceRole because it is constructed dynamically and differs from time to time.

### Relates
Fixes sphinx-doc/sphinx-testing#9.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -1152,7 +1152,7 @@ class Sphinx(object):
         logger.debug('[app] adding autodocumenter: %r', cls)
         from sphinx.ext.autodoc.directive import AutodocDirective
         self.registry.add_documenter(cls.objtype, cls)
-        self.add_directive('auto' + cls.objtype, AutodocDirective)
+        self.add_directive('auto' + cls.objtype, AutodocDirective, override=True)
 
     def add_autodoc_attrgetter(self, typ, getter):
         # type: (Type, Callable[[Any, unicode, Any], Any]) -> None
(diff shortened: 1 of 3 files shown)


tk0miya:
I propose you to use `sphinx.util.docutils:docutils_namespace()` instead in sphinx-testing.
It behaves like a namespace of docutils extensions and it clean up directives and roles after the test:
```
with docutils_namespace():
    TestApp(...)

# at here, all directives and roles are uninstalled
```

mitya57:
@tk0miya I agree that my fix is not elegant.

However it is not *my* code, but there are several projects that are broken with new Sphinx, and because of this I cannot update it in Debian.

Apart from mentioned sphinxcontrib-bibtex, there is also a [similar issue](https://launchpadlibrarian.net/397287493/buildlog_ubuntu-disco-amd64.sphinxcontrib-programoutput_0.11-3_BUILDING.txt.gz) in sphinxcontrib-programoutout. I did not file a bug for it because I was hoping to get it fixed in sphinx-testing.

Maybe there is a way to use `docutils_namespace` in sphinx-testing?

mitya57:
@tk0miya I created new pull request: sphinx-doc/sphinx-testing#10. Closing this now.

PR #5676, opened 2018-11-28
Fix annotaions for extensions

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Update annotations for better world!

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -33,9 +33,14 @@ from sphinx.util.osutil import ensuredir, ENOENT, EPIPE, EINVAL
 if False:
     # For type annotation
     from docutils.parsers.rst import Directive  # NOQA
-    from typing import Any, Dict, List, Tuple  # NOQA
+    from typing import Any, Dict, List, Tuple, Union  # NOQA
     from sphinx.application import Sphinx  # NOQA
     from sphinx.util.typing import unicode  # NOQA
+    from sphinx.writers.html import HTMLTranslator  # NOQA
+    from sphinx.writers.latex import LaTeXTranslator  # NOQA
+    from sphinx.writers.manpage import ManualPageTranslator  # NOQA
+    from sphinx.writers.texinfo import TexinfoTranslator  # NOQA
+    from sphinx.writers.text import TextTranslator  # NOQA
 
 logger = logging.getLogger(__name__)
 
@@ -93,7 +98,7 @@ class graphviz(nodes.General, nodes.Inline, nodes.Element):
 
 
 def figure_wrapper(directive, node, caption):
-    # type: (Directive, nodes.Node, unicode) -> nodes.figure
+    # type: (Directive, graphviz, unicode) -> nodes.figure
     figure_node = nodes.figure('', node)
     if 'align' in node:
         figure_node['align'] = node.attributes.pop('align')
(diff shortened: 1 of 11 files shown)


sphinx/ext/graphviz.py
```diff
@@ -227,7 +231,7 @@ def render_dot(self, code, options, format, prefix='graphviz'):
         return relfn, outfn
 
     if (hasattr(self.builder, '_graphviz_warned_dot') and
-       self.builder._graphviz_warned_dot.get(graphviz_dot)):
+       self.builder._graphviz_warned_dot.get(graphviz_dot)):  # type: ignore  # NOQA
```
tk0miya, line 234: I think `_graphviz_warned_dot` is ugly hack. It should be refactored in future.

sphinx/ext/graphviz.py
```diff
@@ -213,7 +217,7 @@ def run(self):
 
 
 def render_dot(self, code, options, format, prefix='graphviz'):
-    # type: (nodes.NodeVisitor, unicode, Dict, unicode, unicode) -> Tuple[unicode, unicode]
+    # type: (Union[HTMLTranslator, LaTeXTranslator, TexinfoTranslator], unicode, Dict, unicode, unicode) -> Tuple[unicode, unicode]  # NOQA
```
tk0miya, line 220: I feel this Union types is not good way. Any abstract class is needed.

sphinx/ext/intersphinx.py
```diff
@@ -287,7 +287,7 @@ def load_mappings(app):
 
 
 def missing_reference(app, env, node, contnode):
-    # type: (Sphinx, BuildEnvironment, nodes.Node, nodes.Node) -> None
+    # type: (Sphinx, BuildEnvironment, nodes.Element, nodes.TextElement) -> None
```
tk0miya, line 290: I guess `node` might be `nodes.reference` class. I'll take a look in another PR.

sphinx/ext/todo.py
```diff
@@ -61,10 +65,11 @@ def run(self):
...
             self.options['class'] = ['admonition-todo']
 
-        (todo,) = super(Todo, self).run()
-        if isinstance(todo, nodes.system_message):
-            return [todo]
+        (node,) = super(Todo, self).run()
+        if isinstance(node, nodes.system_message):
+            return [node]
 
+        todo = cast(todo_node, node)
```
tk0miya: I feel this "cast" is a bit ugly...

sphinx/ext/todo.py
```diff
@@ -212,17 +219,17 @@ def merge_info(app, env, docnames, other):
...
-    # type: (nodes.NodeVisitor, todo_node) -> None
+    # type: (HTMLTranslator, todo_node) -> None
     self.depart_admonition(node)
 
 
 def latex_visit_todo_node(self, node):
-    # type: (nodes.NodeVisitor, todo_node) -> None
-    title = node.pop(0).astext().translate(tex_escape_map)
+    # type: (LaTeXTranslator, todo_node) -> None
+    title_node = cast(nodes.title, node.pop(0))
```
tk0miya: This is also ugly. This goes pop out and casting at same time. I worried it would confuse somebody.

tk0miya: reverted.

sphinx/ext/todo.py
tk0miya: refactored.

PR #5683, opened 2018-11-29
Add SphinxTranslator as an abstract class

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To help annotation, and to make translators simple, I propose to Introduce `SphinxTranslator` class

### DIFF
--- a/sphinx/util/docutils.py
+++ b/sphinx/util/docutils.py
@@ -39,6 +39,7 @@ if False:
     from types import ModuleType  # NOQA
     from typing import Any, Callable, Generator, List, Set, Tuple, Type  # NOQA
     from docutils.statemachine import State, ViewList  # NOQA
+    from sphinx.builders import Builder  # NOQA
     from sphinx.config import Config  # NOQA
     from sphinx.environment import BuildEnvironment  # NOQA
     from sphinx.io import SphinxFileInput  # NOQA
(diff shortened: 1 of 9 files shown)


tk0miya:
@cocoatomo what do you think about this?

sphinx/writers/latex.py
```diff
@@ -505,8 +507,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
 
     def __init__(self, document, builder):
         # type: (nodes.document, LaTeXBuilder) -> None
-        super(LaTeXTranslator, self).__init__(document)
-        self.builder = builder
+        super(LaTeXTranslator, self).__init__(builder, document)
```
tk0miya, line 514: Unfortunately, the order of arguments are different between Translators.
I think it's time to integrate to one way.

sphinx/util/docutils.py
```diff
@@ -344,6 +345,32 @@ def config(self):
         return self.env.config
 
 
+class SphinxTranslator(nodes.NodeVisitor):
```
tk0miya, line 387: Is there any appropriate place for this class? `util` package is not best to me, but I don't understand where is better.

sphinx/writers/latex.py
tk0miya: It must be a breaking change. So I'll post this as another PR.

jfbu (commented):
Since this PR, in LaTeX output for document with no language settings one sees

    \addto\captionsenglish{\renewcommand{\contentsname}{None}}

with "None" in place of expected "Contents". Ping @tk0miya 

(or rather than Contents, there simply should *not* be any such line in LaTeX document)

PR #5693, opened 2018-11-30
LaTeX (doc): explain how to allow Unicode input in math

### LINKED ISSUE #5249: LaTeX: Unicode Greek letters in math directive break PDF build
Consider any project using Unicode α for example in math mode. The `make html` works fine with MathJax rendering. But the `make latexpdf` fails.

File `index.rst`:

```
.. math::

   E=mα^2
```

Do `make latexpdf`.

#### Error logs / results
```
! Package inputenc Error: Unicode character α (U+3B1)

See the inputenc package documentation for explanation.
 ...                                              
                                                  
l.83 \begin{split}E=mα^2\end{split}
```

#### Expected results

Perhaps it would be nice to be able to use Unicode Greek both in text and in math.

In text this is done by Sphinx escaping it to `\(\alpha\)` etc... macros. In math mode this is more complicated.

If user employes xelatex/lualatex, then loading unicode-math LaTeX package should solve this issue. But it is very large package.

A discussion should be done whether a solution is to be provided with pdflatex too.

### Environment info
- OS: Mac
- Python version: 3.6.4
- Sphinx version: 1.7.x
- TeXLive 2018

jfbu:
Closes: #5249

This is only update to documentation.

There is no solution for pdflatex, and it appears unneeded to add per default ``unicode-math`` usage for ``xelatex/lualatex`` (especially as some option may be needed to pass to the package as explained in the added documentation from this PR).

Thus I propose this as "fix" of #5249.

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1798,16 +1798,27 @@ information.
 
    ``'pdflatex'``\ 's support for Unicode characters is limited.
 
-   If your project uses Unicode characters, setting the engine to
-   ``'xelatex'`` or ``'lualatex'`` and making sure to use an OpenType font
-   with wide-enough glyph coverage (since Sphinx 2.0, the default font is the
-   FreeFont family) is often easier than trying to make
-   ``'pdflatex'`` work with the extra Unicode characters.
-
    .. note::
 
-      2.0 adds to ``'pdflatex'`` automatic support in Latin language document
-      of occasional Cyrillic or Greek letters or words.
+      2.0 adds to ``'pdflatex'`` support in Latin language document of
+      occasional Cyrillic or Greek letters or words.  This is not automatic,
+      see the discussion of the :confval:`latex_elements` ``'fontenc'`` key.
+
+   If your project uses Unicode characters, setting the engine to
+   ``'xelatex'`` or ``'lualatex'`` and making sure to use an OpenType font
+   with wide-enough glyph coverage is often easier than trying to make
+   ``'pdflatex'`` work with the extra Unicode characters.  Since Sphinx 2.0
+   the default is the GNU FreeFont which covers well Latin, Cyrillic and Greek.
+
+   Contrarily to :ref:`MathJaX math rendering in HTML output <math-support>`,
+   LaTeX requires some extra configuration to support Unicode literals in
+   :rst:dir:`math`: the only comprehensive solution (as far as we know) is to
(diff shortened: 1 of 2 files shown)


doc/usage/configuration.rst
```diff
@@ -1804,10 +1804,28 @@ information.
...
+        ``'lualatex'`` *and* to add ``r'\usepackage{unicode-math}'`` via the
+        :confval:`latex_elements` ``'preamble'`` key. By default ``α``
+        (U+03B1) input will be rendered as :math:`\alpha`. To keep it as
+        it was input, use ``r'\usepackage[math-style=literal]{unicode-math}'``.
+      - for ``'pdflatex'``, one way to obtain support for Unicode Greek (but
+        no further Unicode symbols) in math mark-up is to activate the support
+        for Greek in text (see discussion of the :confval:`latex_elements`
+        ``'fontenc'`` key) *and* to add ``r'\usepackage{alphabeta}'`` to
+        ``'preamble'``. An ``α`` (U+03B1) input will render as :math:`\alpha`.
+
```
jfbu: I did not comment on Japanese case: α is ok but for example 𝚹 is not in my brief testing.

jfbu (commented):
I did not test much HTML+MathJaX but it renders well α and 𝚹 and ∫ for example. Only way I know in LaTeX for same level of support is to use xelatex/lualatex with unicode-math. On the other hand it is debatable to add unicode-math to default xelatex/lualatex conf, as it adds overhead and slows down compilation significantly. And its usage may need some user customization such as `math-style=literal` which we can not guess.

CHANGES
```diff
@@ -87,6 +87,7 @@ Bugs fixed
   language and ``'xelatex'`` or ``'lualatex'`` as :confval:`latex_engine`
   (refs: #5251)
 * #5248: LaTeX: Greek letters in section titles disappear from PDF bookmarks
+* LaTeX: document how to allow use of Unicode literals in math (refs: #5249)
```
tk0miya: I think this isn't a bug fix. So I feel a bit strange for this entry.

tk0miya (approved):

CHANGES
jfbu: of course, don't know why I added this entry. I will remove that. And I will try to make the added explanations a bit less verbose. (as matter of fact I find our LaTeX doc occupies already much space...)

jfbu:
merging, thanks for reviewing! (I refactored a bit after that, I hope all for the best :) )

PR #5706, opened 2018-12-02
Fix annotations for _TranslationProxy

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To make annotations better


tk0miya:
OMG, this annotations require py37 mode for mypy.

Issue #5038, opened 2018-06-04
New warnings with multiple `Sphinx` instances


larsoner:
### Problem

Sphinx 1.8 dev emits a warning when instantiating multiple Sphinx instances, which makes previously-passing warning-checking code (sphinx-gallery) fail.

### Procedure to reproduce the problem

```
import os
from sphinx.application import Sphinx
import tempfile
from io import StringIO


with tempfile.TemporaryDirectory() as dir_:
    srcdir = os.path.join(dir_, 'src')
    os.mkdir(srcdir)
    with open(os.path.join(srcdir, "conf.py"), "w") as conffile:
        conffile.write('')

    app = Sphinx(srcdir, srcdir, os.path.join(srcdir, "_build"),
                 os.path.join(srcdir, "_build", "toctree"),
                 "html", warning=StringIO())
    build_warn = app._warning.getvalue()
    assert build_warn == '', build_warn

    app2 = Sphinx(srcdir, srcdir, os.path.join(srcdir, "_build"),
                  os.path.join(srcdir, "_build", "toctree"),
                  "html", warning=StringIO())
    build_warn2 = app2._warning.getvalue()
    assert build_warn2 == '', build_warn2
```

### Error logs / results

```
Running Sphinx v1.8.0+
making output directory...
Running Sphinx v1.8.0+
Traceback (most recent call last):
  File "minimal.py", line 23, in <module>
    assert build_warn2 == '', build_warn2
AssertionError: WARNING: while setting up extension sphinx.domains.math: role 'eq' is already registered, it will be overridden
```

### Expected results
On 1.7.x this code passes (no warning).

### Reproducible project / your project

https://github.com/sphinx-gallery/sphinx-gallery/pull/385

### Environment info

Linux / Travis, Python 3.

If it's not obvious to maintainers what the cause/solution is, I can try bisecting.

tk0miya:
Sphinx itself and its extensions register directives and roles to docutils core. That is global change.
As a result, they are conflicted when invoking Sphinx twice or more.

Please use `sphinx.util.docutils:docutils_namespace()` to create a namespace for docutils on each invoking:
```
from sphinx.util.docutils import docutils_namespace

with docutils_namespace():
    app = Sphinx(...)
```

tk0miya:
Note:
`sphinx.util.docutils.patch_docutils()` is also used to patch docutils temporarily. So I'd like to propose you to use it also. (added since 1.6)

[...]

PR #5718, opened 2018-12-04
refactor: Use app.add_node() to register Sphinx own nodes

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- At present, custom nodes of Sphinx is registered to docutils on importing.
- IMO, it's like pollution of global namespace.
- After this change, we can enable it only during running using `docutils_namespace()`.

### DIFF
--- a/sphinx/addnodes.py
+++ b/sphinx/addnodes.py
@@ -17,7 +17,8 @@ from sphinx.deprecation import RemovedInSphinx30Warning
 
 if False:
     # For type annotation
-    from typing import List, Sequence  # NOQA
+    from typing import Any, Dict, List, Sequence  # NOQA
+    from sphinx.application import Sphinx  # NOQA
     from sphinx.util.typing import unicode  # NOQA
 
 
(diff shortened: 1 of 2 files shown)


sphinx/addnodes.py
```diff
@@ -351,7 +352,42 @@ class manpage(nodes.Inline, nodes.FixedTextElement):
...
+    app.add_node(start_of_file)
+    app.add_node(highlightlang)
+    app.add_node(tabular_col_spec)
+    app.add_node(meta)
+    app.add_node(pending_xref)
+    app.add_node(number_reference)
+    app.add_node(download_reference)
+    app.add_node(literal_emphasis)
+    app.add_node(literal_strong)
+    app.add_node(abbreviation, override=True)
```
tk0miya, line 392: I found this `abbreviation` node is also on docutils.

tk0miya: Wow, it was added at Jan 2003!
https://github.com/docutils-mirror/docutils/commit/1ff0361ecf0641265310b31d69df6a6e4d950188

We should consider remove it from our implementation.
Okay, let's do it in another issue.

larsoner:
Just a FYI in case anyone else hit a cryptic error with Sphinx recently -- over in `sphinx-gallery` for the last day or so we have been getting this error when building with latest Sphinx `master`:
```
self = <docutils.writers._html_base.SimpleListChecker object at 0x7ff474b27748>
node = <compact_paragraph: <reference...>>
    def unknown_visit(self, node):
        """
        Called when entering unknown `Node` types.
    
        Raise an exception unless overridden.
        """
        if  (self.document.settings.strict_visitor
             or node.__class__.__name__ not in self.optional):
            raise NotImplementedError(
                '%s visiting unknown node type: %s'
>               % (self.__class__, node.__class__.__name__))
E           NotImplementedError: <class 'docutils.writers._html_base.SimpleListChecker'> visiting unknown node type: compact_paragraph
```
See the related [Travis log](https://travis-ci.org/sphinx-gallery/sphinx-gallery/jobs/463698115#L1174) for full traceback.

Rolling back to de37ad7 made this problem go away. It looks like this has to do with our use of `docutils_namespace` context manager as suggested in #5038 [here](https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/tests/test_gen_gallery.py#L72-L76). I had to change it to also do the `build` in the same context, rather than separately, to fix things.

tk0miya:
@larsoner Thank you for comment, and I'm sorry for breaking change... I strongly recommend to use `docutils_namespace()` for testing. It prevents global namespace pollution by Sphinx. Thanks,

PR #5737, opened 2018-12-09
refactor: special rubric for footnotes

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To make our code simple


tk0miya:
Note:
I don't know the reason why we handle a rubric having "Footnotes" text specially. The behavior is only supported by LaTeX, manpage and texinfo writers. They were implemented at c5580ebd91, 121b864f31 and d26bc39ec7.

This does not harm the feature. It only refactors the code.

PR #5742, opened 2018-12-09
Remove _TranslationProxy methods that duplicate UserString

jdufresne:
The warning in the comment:

> replace function from UserString; it instantiates a self.__class__ for
> the encoding result

Doesn't apply to these methods as they return non text types. Mostly
bool. By removing these methods, there is less to maintain, annotate,
etc.


tk0miya (approved):
LGTM!

PR #5717, opened 2018-12-04
 refactor ``mock()`` to based on PEP-451 implementation

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- importing framework is now updated on PEP-451
- In addition, `Loader.find_module()` API was deprecated on py3.4

PR #5743, opened 2018-12-09
Define _MockImporter as a MetaPathFinder

jdufresne:
The instance is added to sys.meta_path. Per the documentation, this should be a `MetaPathFinder`:

https://docs.python.org/3/library/sys.html#sys.meta_path

Correct the `_MockImporter.find_module()` type signature per typeshed.

https://github.com/python/typeshed/blob/0b49ce75b478fdf283dda5dd1368759ac342dfe2/stdlib/3/importlib/abc.pyi#L56-L57

### DIFF
--- a/sphinx/ext/autodoc/importer.py
+++ b/sphinx/ext/autodoc/importer.py
@@ -98,7 +98,7 @@ class _MockModule(ModuleType):
         return o
 
 
-class _MockImporter:
+class _MockImporter(MetaPathFinder):
     def __init__(self, names):
         # type: (List[str]) -> None
         self.names = names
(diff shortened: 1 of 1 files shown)


tk0miya:
Thank you for your work. But I did same fix in #5717. And `_MockImporter` was replaced with `MockFinder` and `MockLoader` now.

jdufresne:
Thanks for the link. I agree with adding `MockFinder`.

However, that commit didn't also fix `_MockImporter`. Without this fix, `mypy --python-version 3.5` issues the error:

```
sphinx/ext/autodoc/importer.py: note: In member "__init__" of class "_MockImporter":
sphinx/ext/autodoc/importer.py:106:9: error: Argument 2 to "insert" of "list" has incompatible type "_MockImporter"; expected "MetaPathFinder"
```

tk0miya (approved):
+1

tk0miya:
Thank you for comment. +1 for suppressing warnings!

PR #5745, opened 2018-12-09
Deprecate sphinx.util.typing.unicode and remove all usage

jdufresne:
The definition is unnecessary.


tk0miya:
Sorry, I'm using this now for migration to mypy-py3 mode. So please keep this as is for a while.

Note: this is added only in master branch. So deprecation is not needed because it is not released yet.

jdufresne:
Instead of using the alias, wouldn't it make more sense to change the uses of `unicode` to `str`? What is the rationale? Or is that considered "step 2" in the process?

tk0miya:
Yes, I'll replace all `unicode` types after migration to mypy-py3 mode. I hope this alias makes the "diff" of the migration pull request small and review-able.

tk0miya:
>Yes, I'll replace all unicode types after migration to mypy-py3 mode.

Now I'm working for mypy-py3 mode in #5760. And I'll merge it after a while (if no objection). After that, I'll replace all of `unicode` type with `str` type.

tk0miya:
I replaced all "unicode" type by "str" in #5773.
Thanks,

Issue #703, opened 2015-01-02
MultiByte filename handling


shimizukawa:
Current version of Sphinx handle filename as Unicode, but few implementation lines breaks that cause UnicodeEncodeError / UnicodeDecodeError exception.

(forked repository's changeset: https://bitbucket.org/shimizukawa/sphinx/changeset/40c9407c28e0 )

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/703
- Originally reported by: Takayuki Shimizukawa
- Originally created at: 2011-05-24T17:23:14.868

shimizukawa:
_From Takayuki Shimizukawa on 2011-06-30 00:43:15+00:00_

fix again: new patch for multibyte filename handling.

shimizukawa:
_From Takayuki Shimizukawa on 2011-06-30 02:50:03+00:00_

fix again again..: please check only sphinx-1.0.7-multibyte-filename3.diff.

shimizukawa:
_From Georg Brandl on 2011-09-23 08:18:32+00:00_

Hmm, why doesn't docutils like unicode as its source? Can you investigate?

In any case, to get rid of all (?) of these ugly Unicode issues with path names, best use Sphinx 1.1 on Python 3.

shimizukawa:
_From Takayuki Shimizukawa on 2012-11-05 12:16:38+00:00_

Issue #1016 was marked as a duplicate of this issue.

shimizukawa:
_From Takayuki Shimizukawa on 2012-11-05 12:18:51+00:00_

I updated implementation that is here: [PR61](https://bitbucket.org/birkenfeld/sphinx/pull-request/61/support-multibyte-filename-handling-issue/diff).

shimizukawa:
_From Toshio Kuratomi on 2013-04-12 21:48:54+00:00_

Just tried to build sphinx-1.2b1 where PR61 was merged.  This patch causes breakage when the filenames are undecodable (or unencodable) in the current locale.

For a simple method to detect this on a Linux system:

$ cd Sphinx-1.2b1/tests
$ LC_ALL=C ./run.py -x

Running Sphinx test suite...

# ....................E

## ERROR: test_build.test_multibyte_path

Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(_self.arg)
[...]

PR #5747, opened 2018-12-09
Remove unused `except UnicodeDecodeError` block

jdufresne:
Both `self.srcdir` and `rel_fn` are defined as type unicode. Calling
`os.path.join()` on two unicode values will always return a unicode value
without decoding it. Therefore, the except block is never executed.

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -10,7 +10,6 @@
 
 import os
 import pickle
-import sys
 import warnings
 from collections import defaultdict
 from copy import copy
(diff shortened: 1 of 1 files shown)


tk0miya:
@shimizukawa Could you review this please? I still don't understand the mechanism of `UnicodeDecodeError` at here well.

jdufresne:
Rebased to resolve merge conflicts.

shimizukawa:
This is old code forgotten to remove on Sphinx-1.2.

(History;
These lines were introduced at a commit https://github.com/sphinx-doc/sphinx/commit/ea5b0088fadd79584846242769ac8b7d2d264690 to support non-ASCII characters in a path of `literalinclude` that has been shipped with Sphinx-0.6.
From Sphinx 1.2, the entire code was changed to support non-ASCII file paths: https://github.com/sphinx-doc/sphinx/commit/1bb4923da1aa1878af44c4608d63025e4fae9d3b and https://github.com/sphinx-doc/sphinx/commit/58be2f19b3aeeabf4d9a3d10be84824508c0fda2 for #703
)

shimizukawa (approved):
LGTM!

shimizukawa:
Merged. Thanks for your contribution!

tk0miya:
Thanks!

PR #5757, opened 2018-12-11
Simplify open() call by removing default mode

jdufresne:
The open() function opens files in read-only text mode by default. Drop the mode argument to be slightly simpler and more idiomatic.

https://docs.python.org/3/library/functions.html#open

> The default mode is 'r' (open for reading text, synonym of 'rt').


tk0miya (commented):
Reasonable!

Issue #5755, opened 2018-12-11
C++: duplicate error with template function redefinition


pitrou:
I was redirected here from the Breathe issue tracker, specifically https://github.com/michaeljones/breathe/issues/406

I have two different implementations of a C++ template method that are selected using the [SFINAE pattern](https://en.cppreference.com/w/cpp/language/sfinae) (implementations omitted below):
```cpp

class BooleanBuilder : public ArrayBuilder {
 public:
  /// \brief Append a sequence of elements in one shot, with a specified nullmap
  /// \param[in] values_begin InputIterator to the beginning of the values
  /// \param[in] values_end InputIterator pointing to the end of the values
  /// \param[in] valid_begin InputIterator with elements indication valid(1)
  ///  or null(0) values
  /// \return Status
  template <typename ValuesIter, typename ValidIter>
  typename std::enable_if<!std::is_pointer<ValidIter>::value, Status>::type AppendValues(
      ValuesIter values_begin, ValuesIter values_end, ValidIter valid_begin) {
     ...
  }

  /// \brief Append a sequence of elements in one shot, with a specified nullmap
  /// \param[in] values_begin InputIterator to the beginning of the values
  /// \param[in] values_end InputIterator pointing to the end of the values
  /// \param[in] valid_begin uint8_t* indication valid(1) or null(0) values.
  ///  nullptr indicates all values are valid.
  /// \return Status
  template <typename ValuesIter, typename ValidIter>
  typename std::enable_if<std::is_pointer<ValidIter>::value, Status>::type AppendValues(
      ValuesIter values_begin, ValuesIter values_end, ValidIter valid_begin) {
     ...
  }
};
```

When trying to integrate Doxygen documentation for that class's members, I get the following error:
```
Warning, treated as error:
/home/antoine/arrow/docs/source/cpp/api/builder.rst:31:Duplicate declaration.
Makefile:71: recipe for target 'html' failed
make: *** [html] Error 2
```

Sphinx version is 1.8.2, Python version is 3.7.

jakobandersen:
Thanks. I wonder how complex a comparison the compiler is supposed to do. As example:
```c++
#include <type_traits>

template<bool B>
typename std::enable_if<B>::type f() {}

template<bool B>
typename std::enable_if<!B>::type f() {}

template<bool B> // not a duplicate until instantiation
typename std::enable_if<!!B>::type f() {}

template<bool A, bool B>
typename std::enable_if<A && B>::type g() {}

[...]

PR #5760, opened 2018-12-11
Migrate to mypy py3 mode

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Before moving to "type annotation" syntax in py3, we have to move to mypy-py3 mode.
- To do that, we have to use `docutils-stubs` typing package to check our typing.
- This try to migrate to mypy-py3 mode.
- Thanks to @cocoatomo. He helps me to make `docutils-stubs` package and this PR.

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -2232,7 +2232,7 @@ class ASTNestedName(ASTBase):
             #       so it can remove it in inner declarations.
             dest = signode
             if mode == 'lastIsName':
-                dest = addnodes.desc_addname()
+                dest = addnodes.desc_addname()  # type: ignore
             for i in range(len(names)):
                 nne = names[i]
                 template = self.templates[i]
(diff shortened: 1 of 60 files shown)


sphinx/writers/html5.py
```diff
@@ -420,7 +420,7 @@ def depart_caption(self, node):
 
     def visit_doctest_block(self, node):
         # type: (nodes.doctest_block) -> None
-        self.visit_literal_block(node)
+        self.visit_literal_block(node)  # type: ignore
```
tk0miya: I feel currrent annotations for visitor methods are too strict. I understand it is better to use concrete types to represent what type of node visitor methods accept. But it is too sensitive for authors of derived implementations.
Personally, using `nodes.Element` would better for writers. It is ambiguous, but easy to use.

@cocoatomo what do you think?

sphinx/domains/cpp.py
```diff
@@ -6395,6 +6395,10 @@ def add_target_and_index(self, ast, sig, signode):
             signode['first'] = (not self.names)  # hmm, what is this about?
             self.state.document.note_explicit_target(signode)
 
+    def get_index_text(self, name):
+        # type: (unicode) -> unicode
+        raise NotImplementedError()
```
tk0miya: Note: This PR contains a commit of #5755. So this should be merged after it.

tk0miya: merged and rebased.

sphinx/writers/html5.py
tk0miya: I talked with him directly, and we decided to use `nodes.Element` instead of concrete node subclasses because strictness harms developers.
refs: https://github.com/tk0miya/docutils-stubs/pull/30

I'll update annotations of Sphinx side later.

tk0miya:
We did it!

PR #5761, opened 2018-12-12
Fix annotations

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To make annotations better

### DIFF
--- a/sphinx/writers/manpage.py
+++ b/sphinx/writers/manpage.py
@@ -75,7 +75,7 @@ class NestedInlineTransform:
                     if matcher(subnode):
                         node.parent.insert(pos + 1, subnode)
                     else:
-                        newnode = node.__class__('', subnode, **node.attributes)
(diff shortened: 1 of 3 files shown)


sphinx/writers/manpage.py
```diff
@@ -75,7 +75,7 @@ def apply(self, **kwargs):
                     if matcher(subnode):
                         node.parent.insert(pos + 1, subnode)
                     else:
-                        newnode = node.__class__('', subnode, **node.attributes)
+                        newnode = node.__class__('', '', subnode, **node.attributes)
```
tk0miya, line 78: To be exact, this is not annotation fix.
In this else-clause, `subnode` is a Text node. That is implicit, and not understandable from code.

This change explicitly give `subnode` as 3rd argument of new node. It is more explicit and type-safe.

PR #5764, opened 2018-12-12
explicitly recommend to inherit CSS via @import

evgeni:
Subject: explicitly recommend to inherit CSS via @import

### Feature or Bugfix
- Feature

(well, actually Documentation ;))
### Purpose
while `@import` is mentioned in the `stylesheet` section, it is not clear that `inherit` won't add the base themes stylesheet by default


tk0miya (approved):
LGTM!

tk0miya:
Thanks!

PR #5767, opened 2018-12-12
Add super method; CPPObject.get_index_text()

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- `CPPObject.add_target_and_index()` calls `self.get_index_text()`. But it is not defined at `CPPObject` class. Only subclasses have it.
- This PR adds it as a super method.


tk0miya:
@jakobandersen Can I add this super method?

jakobandersen:
Sure, go ahead.

tk0miya:
Thanks!

PR #5769, opened 2018-12-14
Use nodes.Element for visitor/departure methods

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
To follow docutils-stubs rule, we use nodes.Element for
visitor/departure methods.

### DIFF
--- a/sphinx/writers/manpage.py
+++ b/sphinx/writers/manpage.py
@@ -128,125 +128,125 @@ class ManualPageTranslator(SphinxTranslator, BaseTranslator):
         return tmpl % self._docinfo
 
     def visit_start_of_file(self, node):
-        # type: (addnodes.start_of_file) -> None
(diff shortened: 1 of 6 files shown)


tk0miya:
see https://github.com/sphinx-doc/sphinx/pull/5760#discussion_r241811009

sphinx/writers/manpage.py
```diff
@@ -477,46 +477,46 @@ def visit_title(self, node):
...
         self.visit_centered(node)
 
     def depart_math_block(self, node):
-        # type: (nodes.math_block) -> None
+        # type: (nodes.Element) -> None
         self.depart_centered(node)
 
     def unknown_visit(self, node):
-        # type: (nodes.Node) -> None
+        # type: (nodes.Element) -> None
```
tk0miya: This should not be changed.

PR #5773, opened 2018-12-15
Replace all "unicode" type by "str"

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Now annotations for Sphinx is migrated to py3 (#5760)
- So "unicode" type is no longer needed. Let's move to `str` type!
- refs: https://github.com/sphinx-doc/sphinx/pull/5745#issuecomment-446674870 cc: @jdufresne 

Note:
I'd like to migrate py3 syntax annotation like `def foo(a: int) -> int` in another step (if I have time...)


sphinx/application.py
```diff
@@ -113,7 +113,7 @@
     # Strictly, alabaster theme is not a builtin extension,
     # but it is loaded automatically to use it as default theme.
     'alabaster',
-)  # type: Tuple[unicode, ...]
+)  # type: Tuple[str, ...]
```
tk0miya: This annotation would be able to removed.

Issue #5772, opened 2018-12-14
LaTeX: should the Bjarne style of fncychap be used for English also if passed as language option?


jfbu:
Currently  there is https://github.com/sphinx-doc/sphinx/blob/ca2ce49830f2e574aa0e9426670d08d44078187b/sphinx/writers/latex.py#L595-L601

but this has the effect that when `sphinx-build` is executed on our documentation at RTD to produce the PDF, it will use "Sonny" `fncychap` style because `language` is explicitely set. If the documentation is build locally (for testing) it uses the "Bjarne" style from this default: https://github.com/sphinx-doc/sphinx/blob/ca2ce49830f2e574aa0e9426670d08d44078187b/sphinx/writers/latex.py#L149

The question is whether it would make sense to apply the Sonny style only for non-English, when a language is explicitely selected. It is a bit weird that explicitely specifying English as language turns off some style elements of PDF which were tested with English... and apply if no language is specified.

tk0miya:
Our document says:

>Inclusion of the “fncychap” package (which makes fancy chapter titles), default '\\usepackage[Bjarne]{fncychap}' for English documentation (this option is slightly customized by Sphinx), '\\usepackage[Sonny]{fncychap}' for internationalized docs (because the “Bjarne” style uses numbers spelled out in English). 
http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-latex_elements

So +1 for applying Bjarne for English document even if language is explicitly given.

PR #5776, opened 2018-12-15
LaTeX: use Bjarne option of fncychap also if language is set to 'en'

jfbu:
Closes: #5772

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -2094,6 +2094,13 @@ information.
         "fncychap" styles you can try are "Lenny", "Glenn", "Conny", "Rejne" and
         "Bjornstrup".  You can also set this to ``''`` to disable fncychap.
 
+        The default is ``''`` for Japanese documents.
+
+        .. versionchanged:: 2.0
+           Previously, explicitly setting the language to ``'en'`` (in the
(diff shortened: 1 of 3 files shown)


doc/usage/configuration.rst
```diff
@@ -2094,6 +2094,13 @@ information.
         "fncychap" styles you can try are "Lenny", "Glenn", "Conny", "Rejne" and
         "Bjornstrup".  You can also set this to ``''`` to disable fncychap.
 
+        The default is ``''`` for Japanese documents.
+
+        .. versionchanged:: 2.0
+           Previously, explicitly setting the language to ``'en'`` (in the
+           :file:`conf.py` or via a command line option as on RTD) led to
+           usage of ``Sonny`` as for other languages.  Now, ``Bjarne`` is used.
```
tk0miya: I believe this change obeys to our document. In other words, it is a bug simply. So no changes are needed for our docs, I think.

Of course, it's okay to mention about Japanese :-)

tk0miya (approved):
LGTM with nits

doc/usage/configuration.rst
jfbu: agreed, thanks. I pushed a commit to remove that paragraph and I will squash when merging for a cleaner history.

PR #5783, opened 2018-12-15
Replace EnvironmentError and IOError by OSError

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
Since python 3.3, EnvironmentError and IOError were merged into OSError.
refs: https://docs.python.org/3/library/exceptions.html#OSError

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -26,7 +26,7 @@ from sphinx.util.docutils import SphinxDirective
 from sphinx.util.fileutil import copy_asset
 from sphinx.util.i18n import search_image_for_language
 from sphinx.util.nodes import set_source_info
-from sphinx.util.osutil import ensuredir, EPIPE, EINVAL
+from sphinx.util.osutil import ensuredir
 
 if False:
     # For type annotation
@@ -144,7 +144,7 @@ class Graphviz(SphinxDirective):
             try:
                 with open(filename, encoding='utf-8') as fp:
                     dotcode = fp.read()
-            except (IOError, OSError):
+            except OSError:
                 return [document.reporter.warning(
                     __('External Graphviz file %r not found or reading '
                        'it failed') % filename, line=self.lineno)]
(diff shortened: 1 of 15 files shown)


sphinx/ext/graphviz.py
```diff
@@ -261,7 +261,7 @@ def render_dot(self, code, options, format, prefix='graphviz'):
         # Graphviz may close standard input when an error occurs,
         # resulting in a broken pipe on communicate()
         stdout, stderr = p.communicate(code.encode('utf-8'))
-    except (OSError, IOError) as err:
+    except OSError as err:
```
jdufresne: Given the `errno` check below, could this catch `BrokenPipeError` instead?

https://docs.python.org/3/library/exceptions.html#BrokenPipeError

tk0miya: It seems better. but it only responds to `EPIPE`. So we still have to catch `OSError`, right?

jdufresne: Yeah, that is right. So `OSError` is still necessary to handle the `EINVAL` case, so there isn't much point to changing it. Do you know what triggers an `EINVAL`? I tried to find more details, but didn't find anything specific.

tk0miya: It comes from 9350b3b46b. The commit message says:

>This happen on Python2.7 and Graphviz version 2.12 (Mon Dec  4 22:04:37 UTC 2006)
on Windows when a not recognized renderer type is asked.

I don't know latest python and windows causes the `EINVAL` error. We have to confirm that before removing...

tk0miya: The `EINVAL` has not been thrown at present. That was fixed at https://bugs.python.org/issue19612
https://github.com/python/cpython/blob/master/Lib/subprocess.py#L880

@gjo Thank you for advice!

tk0miya: fixed.

jdufresne (approved):

shimizukawa (approved):
LGTM!

tk0miya:

PR #5796, opened 2018-12-15
Use Python 3 super() argument-less syntax

jdufresne:
The form is less verbose and more idiomatic for Python 3 only code.

https://docs.python.org/3/library/functions.html#super


tk0miya (approved):
cool!

PR #5797, opened 2018-12-15
Replace ENOENT errno checks with Python 3 FileNotFoundError

jdufresne:
The error is more specific and self documenting.

This removes the last use of `sphinx.util.osutil.ENOENT`, so it is now deprecated for removal. `sphinx.util.osutil.EEXIST` was already unused so that is deprecated as well.


tk0miya (approved):
Resonable!

PR #5798, opened 2018-12-15
Remove unnecessary check for type bytes

jdufresne:
On Python 3, the `__file__` attribute is always a string. See docs:

<https://docs.python.org/3/reference/import.html#__file__>

> `__file__` is optional. If set, this attribute’s value must be a string.


tk0miya (approved):
:-)

PR #5800, opened 2018-12-15
TODOs inside a :param: have a uri, but not a refid.

daleevans:
Subject: This avoids a KeyError when generating a link to a TODO that crashes sphinx, and gives a warning so the offending TODO can be tracked down.

### Feature or Bugfix
- Bugfix

### Purpose
- avoid a specific crash when building todolists

### Detail
- a todo inside a :param: will have a relative uri, but not a refid.  instead of crashing, just skip adding the target to the link, and emit a warning that tells the user where the todo is so they can move it or ignore it.  besides that, a missing refid in other cases will be easier to debug with this warning message rather than just a stack trace and a crash.

### DIFF
--- a/sphinx/ext/todo.py
+++ b/sphinx/ext/todo.py
@@ -174,7 +174,15 @@ def process_todo_nodes(app, doctree, fromdocname):
             try:
                 newnode['refuri'] = app.builder.get_relative_uri(
                     fromdocname, todo_info['docname'])
-                newnode['refuri'] += '#' + todo_info['target']['refid']
+                try:
+                    newnode['refuri'] += '#' + todo_info['target']['refid']
+                except KeyError:
+                    # presently, there's no refid for a todo that appears
+                    # inside a param, possibly other places.  avoid crashing
+                    # and leave a warning so we can track it down.
+                    logger.warning(__("TODO target missing refid: %s"),
+                                   todo_info['source'],
+                                   location=node)
             except NoUri:
                 # ignore if no URI can be determined, e.g. for LaTeX output
                 pass


tk0miya:
Could you share some example to reproduce the error please?
I think this does not resolve the problem itself. So I'd like to check it.

daleevans:
> 
> 
> Could you share some example to reproduce the error please?
> I think this does not resolve the problem itself. So I'd like to check it.

Sure, if you clone this:

https://github.com/daleevans/sphinx-toc-bug-demo

and run `make html` in the sphinx directory you should get this:

```
Running Sphinx v1.8.2
making output directory...
building [html]: targets for 3 source files that are out of date
updating environment: 3 added, 0 changed, 0 removed
reading sources... [100%] modules
/home/xxx/bug/sphinx-toc-bug-demo/module.py:docstring of module.Test.a:4: WARNING: Duplicate ID: "module-module".
pickling environment... done
checking consistency... /home/xxx/bug/sphinx-toc-bug-demo/sphinx/source/modules.rst: WARNING: document isn't included in any toctree
done
preparing documents... done
writing output... [ 33%] index
Exception occurred:
  File "/home/xxx/venv/lib/python3.6/site-packages/docutils/nodes.py", line 567, in __getitem__
    return self.attributes[key]
KeyError: 'refid'
The full traceback has been saved in /tmp/sphinx-err-b96xdwu5.log, if you want to report the issue to the developers.
Makefile:19: recipe for target 'html' failed
make: *** [html] Error 2

```

tk0miya:
Thank you for example. I found the root cause of the error. docutils does not assign `refid` attributes to the target nodes under TextElement.
I just pushed #5848 to follow such case. I believe it fixes the error perfectly.
Could you check this please?

daleevans:
> 
> 
> Thank you for example. I found the root cause of the error. docutils does not assign `refid` attributes to the target nodes under TextElement.
> I just pushed #5848 to follow such case. I believe it fixes the error perfectly.
> Could you check this please?

Yes, this works perfectly.  Thanks!

tk0miya:
Thank you for confirm. I just merged #5848 instead.
Thanks,

PR #5801, opened 2018-12-15
Remove unnecessary bytes type check in ModuleAnalyzer.for_string()

jdufresne:
All uses always pass a str, never a bytes. Per the type signature, only str types are allowed.


tk0miya (approved):
LGTM

PR #5802, opened 2018-12-15
Remove unnecessary bytes type check from path.write_text()

jdufresne:
All code passes type str to the method. Per the type signature, only str is allowed.


tk0miya (approved):
Exactly

PR #5803, opened 2018-12-16
Avoid respecifying default encoding for .encode()/.decode() calls

jdufresne:
In Python 3, both `.encode()` and `.decode()` default the encoding to 'utf-8'. See the docs:

https://docs.python.org/3/library/stdtypes.html#str.encode
https://docs.python.org/3/library/stdtypes.html#bytes.decode

Simplify and shorten the code by using the default instead of respecifying it.


tk0miya (approved):
Wow! I don't know this technique. Great! :-)

PR #5804, opened 2018-12-16
Remove unnecessary bytes/str type check in FileAvoidWrite.write()

jdufresne:
All calls to `FileAvoidWrite.write()` always pass a text string. Additionally, the type signature only allows type str.

### DIFF
--- a/sphinx/util/osutil.py
+++ b/sphinx/util/osutil.py
@@ -18,10 +18,8 @@ import shutil
 import sys
 import time
 import warnings
-from io import BytesIO, StringIO
+from io import StringIO
 from os import path
-
-from six import text_type
 
 from sphinx.deprecation import RemovedInSphinx30Warning, RemovedInSphinx40Warning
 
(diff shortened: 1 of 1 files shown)


sphinx/util/osutil.py
```diff
@@ -238,17 +236,13 @@ class FileAvoidWrite:
     def __init__(self, path):
         # type: (str) -> None
         self._path = path
-        self._io = None  # type: Union[StringIO, BytesIO]
+        self._io = None  # type: StringIO
 
     def write(self, data):
```
tk0miya, line 241: To make sure `data` is an instance of str, please add `assert` statement here.

tk0miya (commented):
LGTM with nits

sphinx/util/osutil.py
jdufresne: The assert would be redundant with Python 3 StringIO's own type checking. Observe:

```py
>>> import io
>>> buf = io.StringIO()
>>> buf.write(b'')
TypeError: string argument expected, got 'bytes'
```

WDYT?

tk0miya: You're right. No assertion is needed.

tk0miya (approved):

PR #5805, opened 2018-12-16
Remove unnecessary bytes/str type check from doctree_read()

jdufresne:
ModuleAnalyzer.code is always type str. Decoding is unnecessary.

### DIFF
--- a/sphinx/ext/viewcode.py
+++ b/sphinx/ext/viewcode.py
@@ -13,7 +13,6 @@ import traceback
 import warnings
 
 from docutils import nodes
-from six import text_type
 
 import sphinx
 from sphinx import addnodes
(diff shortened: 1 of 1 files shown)


sphinx/ext/viewcode.py
```diff
@@ -75,10 +74,7 @@ def has_tag(modname, fullname, docname, refname):
                 env._viewcode_modules[modname] = False  # type: ignore
                 return
 
-            if not isinstance(analyzer.code, text_type):
-                code = analyzer.code.decode(analyzer.encoding)
-            else:
-                code = analyzer.code
+            code = analyzer.code
```
tk0miya: How about moving this line after `find_tags()` call. I feel it is a bit better.

tk0miya (commented):
LGTM with nits

sphinx/ext/viewcode.py
jdufresne: Done :smile:

tk0miya (approved):
:-)

PR #5806, opened 2018-12-16
Replace use of six.text_type with str

jdufresne:
This removes the last use of the six package allowing Sphinx to remove it as a dependency.

### DIFF
--- a/sphinx/builders/latex/__init__.py
+++ b/sphinx/builders/latex/__init__.py
@@ -12,7 +12,6 @@ import os
 from os import path
 
 from docutils.frontend import OptionParser
-from six import text_type
 
 from sphinx import package_dir, addnodes, highlighting
 from sphinx.builders import Builder
@@ -24,7 +23,7 @@ from sphinx.builders.latex.transforms import (
 from sphinx.config import ENUM
 from sphinx.environment import NoUri
 from sphinx.environment.adapters.asset import ImageAdapter
-from sphinx.errors import SphinxError, ConfigError
+from sphinx.errors import SphinxError
 from sphinx.locale import _, __
 from sphinx.transforms import SphinxTransformer
 from sphinx.util import texescape, logging, status_iterator
@@ -400,23 +399,6 @@ class LaTeXBuilder(Builder):
 
 def validate_config_values(app, config):
     # type: (Sphinx, Config) -> None
-    for document in config.latex_documents:
-        try:
-            text_type(document[2])
-        except UnicodeDecodeError:
-            raise ConfigError(
-                __('Invalid latex_documents.title found (might contain non-ASCII chars. '
-                   'Please use u"..." notation instead): %r') % (document,)
-            )
-
-        try:
-            text_type(document[3])
-        except UnicodeDecodeError:
-            raise ConfigError(
-                __('Invalid latex_documents.author found (might contain non-ASCII chars. '
-                   'Please use u"..." notation instead): %r') % (document,)
-            )
-
     for key in list(config.latex_elements):
         if key not in DEFAULT_SETTINGS:
             msg = __("Unknown configure key: latex_elements[%r]. ignored.")
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -16,8 +16,6 @@ from collections import OrderedDict
 from os import path, getenv
 from typing import Any, NamedTuple, Union
 
-from six import text_type
-
 from sphinx.deprecation import RemovedInSphinx30Warning, RemovedInSphinx40Warning
 from sphinx.errors import ConfigError, ExtensionError
 from sphinx.locale import _, __
@@ -41,7 +39,7 @@ copyright_year_re = re.compile(r'^((\d{4}-)?)(\d{4})(?=[ ,])')
 
 ConfigValue = NamedTuple('ConfigValue', [('name', str),
                                          ('value', Any),
-                                         ('rebuild', Union[bool, text_type])])
+                                         ('rebuild', Union[bool, str])])
 
 
 def is_serializable(obj):
@@ -78,7 +76,7 @@ class ENUM:
 
 
 # RemovedInSphinx40Warning
-string_classes = [text_type]  # type: List
+string_classes = [str]  # type: List
 
 
 class Config:
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -12,12 +12,10 @@
 
 import inspect
 import re
-import sys
 import warnings
 from typing import Any
 
 from docutils.statemachine import StringList
-from six import text_type
 
 import sphinx
 from sphinx.deprecation import RemovedInSphinx30Warning, RemovedInSphinx40Warning
@@ -461,13 +459,7 @@ class Documenter:
     def get_sourcename(self):
         # type: () -> str
         if self.analyzer:
-            # prevent encoding errors when the file name is non-ASCII
-            if not isinstance(self.analyzer.srcname, text_type):
-                filename = text_type(self.analyzer.srcname,
-                                     sys.getfilesystemencoding(), 'replace')
-            else:
-                filename = self.analyzer.srcname
-            return '%s:docstring of %s' % (filename, self.fullname)
+            return '%s:docstring of %s' % (self.analyzer.srcname, self.fullname)
         return 'docstring of %s' % self.fullname
 
     def add_content(self, more_content, no_docstring=False):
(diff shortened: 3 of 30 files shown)


sphinx/builders/latex/__init__.py
```diff
@@ -403,15 +402,15 @@ def validate_config_values(app, config):
     # type: (Sphinx, Config) -> None
     for document in config.latex_documents:
         try:
-            text_type(document[2])
+            str(document[2])
```
tk0miya: This and next call try to check the member of `latex_documents` contains unicode chars or not.
But this validation has been finished their role because py3 syntax does not generate bytes object in standard use.

I think removing both block is better. What do you think?

sphinx/config.py
```diff
@@ -78,7 +76,7 @@ def match(self, value):
             return value in self.candidates
 
 
-string_classes = [text_type]  # type: List
+string_classes = [str]  # type: List
```
tk0miya, line 79: Originally, `string_classes` has been used to switch string classes with python interpreters:
https://github.com/sphinx-doc/sphinx/blob/c899b27b4a659f95a20d80bf467f1e6d21ffb883/sphinx/config.py#L85-L87

After our refactoring, it does not have any role. So it would be better to remove this now.

sphinx/directives/code.py
```diff
@@ -170,7 +169,7 @@ def run(self):
             try:
                 literal = container_wrapper(self, literal, caption)
             except ValueError as exc:
-                return [document.reporter.warning(text_type(exc), line=self.lineno)]
+                return [document.reporter.warning(str(exc), line=self.lineno)]
```
tk0miya: I found `Reporter` takes an exception object as its argument:
https://github.com/docutils-mirror/docutils/blob/e88c5fb08d5cdfa8b4ac1020dd6f7177778d5990/docutils/utils/__init__.py#L152-L160

So this conversion is not needed.

sphinx/directives/code.py
```diff
@@ -465,7 +464,7 @@ def run(self):
 
             return [retnode]
         except Exception as exc:
-            return [document.reporter.warning(text_type(exc), line=self.lineno)]
+            return [document.reporter.warning(str(exc), line=self.lineno)]
```
tk0miya: Same as above.

sphinx/ext/autodoc/__init__.py
```diff
@@ -463,9 +462,9 @@ def get_sourcename(self):
         # type: () -> str
         if self.analyzer:
             # prevent encoding errors when the file name is non-ASCII
-            if not isinstance(self.analyzer.srcname, text_type):
-                filename = text_type(self.analyzer.srcname,
-                                     sys.getfilesystemencoding(), 'replace')
+            if not isinstance(self.analyzer.srcname, str):
+                filename = str(self.analyzer.srcname,
+                               sys.getfilesystemencoding(), 'replace')
```
tk0miya: @shimizukawa Is this code still needed in py3? I can't imagine the case `srcname` is not `str`.

sphinx/ext/autosummary/__init__.py
```diff
@@ -161,7 +160,7 @@ def autosummary_table_visit_html(self, node):
             par = cast(nodes.paragraph, col1_entry[0])
             for j, subnode in enumerate(list(par)):
                 if isinstance(subnode, nodes.Text):
-                    new_text = text_type(subnode.astext())
+                    new_text = str(subnode.astext())
```
tk0miya: `Node.astext()` always returns `str` instance. So this conversion is not needed. Let's remove it.

sphinx/ext/graphviz.py
```diff
@@ -284,7 +283,7 @@ def render_dot_html(self, node, code, options, prefix='graphviz',
                                    "'svg', but is %r") % format)
         fname, outfn = render_dot(self, code, options, format, prefix)
     except GraphvizError as exc:
-        logger.warning(__('dot code %r: %s'), code, text_type(exc))
+        logger.warning(__('dot code %r: %s'), code, str(exc))
```
tk0miya: I think formatting with `%s` equals conversion with `str`. So this is duplicated. Is my understanding right?

sphinx/util/__init__.py
```diff
@@ -72,7 +71,7 @@ def path_stabilize(filepath):
     # type: (str) -> str
     "normalize path separater and unicode string"
     newpath = filepath.replace(os.path.sep, SEP)
-    if isinstance(newpath, text_type):
+    if isinstance(newpath, str):
```
tk0miya: I think this condition is always passed. Should this be removed?

sphinx/util/inventory.py
```diff
@@ -22,7 +20,7 @@
     from sphinx.builders import Builder  # NOQA
     from sphinx.environment import BuildEnvironment  # NOQA
 
-    Inventory = Dict[text_type, Dict[text_type, Tuple[text_type, text_type, text_type, text_type]]]  # NOQA
+    Inventory = Dict[str, Dict[str, Tuple[str, str, str, str]]]  # NOQA
```
tk0miya: It would be better to move this definition to `sphinx.util.typing` module. This is commonly used in two or more modules.

tk0miya (commented):
LGTM with nits

sphinx/ext/graphviz.py
jdufresne: Yup! I agree. Done in #5814.

sphinx/builders/latex/__init__.py
jdufresne: Done in #5813

sphinx/directives/code.py
jdufresne: Done in #5813

sphinx/ext/autodoc/__init__.py
jdufresne: According to the `ModuleAnalyzer.__init__` type definition, `srcname` must be a `str`. Shall I remove this branch then?

https://github.com/sphinx-doc/sphinx/blob/30ec4b6bba3cb256b700ca1d7438ddd01b3e948c/sphinx/pycode/__init__.py#L82-L85

sphinx/util/inventory.py
jdufresne: Done in #5817

sphinx/config.py
jdufresne: Done in #5818.

jdufresne:
I'll wait for some of the linked PRs to progress or merge and then I'll rebase this when ready. Thanks!

tk0miya:
Could you update this PR please?

jdufresne:
Rebased. I believe I have addressed all the comments, but if I missed something please let me know. Ready for additional feedback.

sphinx/search/__init__.py
```diff
@@ -353,7 +351,7 @@ def get_objects(self, fn2index):
                     if otype:
                         # use unicode() to fire translation proxies
```
tk0miya: This comment should be updated also.

sphinx/testing/path.py
```diff
@@ -222,4 +220,4 @@ def listdir(self):
 
     def __repr__(self):
         # type: () -> str
-        return '%s(%s)' % (self.__class__.__name__, text_type.__repr__(self))
+        return '%s(%s)' % (self.__class__.__name__, str.__repr__(self))
```
tk0miya: Which would be better `str.__repr__()` or `super().__repr__()`?

sphinx/transforms/post_transforms/images.py
```diff
@@ -119,7 +118,7 @@ def handle(self, node):
                 self.app.env.images.add_file(self.env.docname, path)
         except Exception as exc:
             logger.warning(__('Could not fetch remote image: %s [%s]') %
-                           (node['uri'], text_type(exc)))
+                           (node['uri'], str(exc)))
```
tk0miya: This conversion is not needed here.

tests/test_domain_py.py
```diff
@@ -30,22 +29,22 @@ def parse(sig):
...
-    assert text_type(rv) == "source : string, filename, symbol='file'"
+    assert str(rv) == "source : string, filename, symbol='file'"
 
     rv = parse('func(a=[], [b=None])')
-    assert text_type(rv) == 'a=[], [b=None]'
+    assert str(rv) == 'a=[], [b=None]'
 
     rv = parse('func(a=[][, b=None])')
-    assert text_type(rv) == 'a=[], [b=None]'
+    assert str(rv) == 'a=[], [b=None]'
```
tk0miya: I found `parse()` returns `str` value. So this conversion is not needed.

tk0miya (changes requested):
LGTM with nits

tk0miya:
@jakobandersen Could you review the changes of cpp domain?
I'd like to know the conversions are still needed or not. If the variables are string, `text_type()` should be removed simply.

Thanks,

jdufresne:
I have addressed the latest round of feedback. Thanks.

tk0miya (approved):

tk0miya:
It seems there are no special feedback. So I'm merging this now.
Thanks,

PR #5814, opened 2018-12-16
Remove redundant coerce to str in string formatting

jdufresne:
When an object is passed to a string format placeholder `'%s'`, Python will implicitly call `str()` on the object. This applies to `print()` and logging as well.


tk0miya (approved):
LGTM!

PR #5817, opened 2018-12-16
Define a common Inventory type for reuse across modules

jdufresne:


sphinx/util/typing.py
```diff
@@ -28,3 +28,5 @@
 
 # title getter functions for enumerable nodes (see sphinx.domains.std)
 TitleGetter = Callable[[nodes.Node], text_type]
+
+Inventory = Dict[str, Dict[str, Tuple[str, str, str, str]]]
```
jdufresne, line 32: Let me know if you'd like a comment above this line like the other types. I'm not sure what that text should be.

If the order of these definitions are important, let me know I can place this on whatever line you think is best.

tk0miya: Please note `# inventory data on memory`.

tk0miya: I'll do it after merge.

tk0miya (approved):

tk0miya:
Merged. Thank you always for your hard work!

PR #5818, opened 2018-12-16
Deprecate sphinx.config.string_classes; remove all internal uses

jdufresne:
With only a single text type across supported Python versions, the `string_classes` is no longer necessary.

Internally, all uses were converted to the value `[str]`. For `.add_config_value()` uses that also supply a default string, the type is inferred.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -13,7 +13,7 @@ import mock
 import pytest
 
 import sphinx
-from sphinx.config import Config, ENUM, string_classes, check_confval_types
+from sphinx.config import Config, ENUM, check_confval_types
 from sphinx.errors import ExtensionError, ConfigError, VersionRequirementError
 from sphinx.testing.path import path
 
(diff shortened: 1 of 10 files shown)


tests/test_config.py
```diff
@@ -228,12 +228,8 @@ class C(A):
     ('value8', B(), None, C(), False),                          # sibling type
     ('value9', None, None, 'foo', False),                       # no default or no annotations
     ('value10', None, None, 123, False),                        # no default or no annotations
-    ('value11', None, [str], 'bar', False),                     # str vs unicode
-    ('value12', 'string', None, 'bar', False),                  # str vs unicode
-    ('value13', None, string_classes, 'bar', False),            # string_classes
-    ('value14', None, string_classes, 'bar', False),            # string_classes
-    ('value15', 'unicode', None, 'bar', False),                 # str vs unicode
-    ('value16', 'unicode', None, 'bar', False),                 # str vs unicode
```
jdufresne, line 236: Some of these were removed as they are now redundant or duplicates.

tk0miya (approved):

PR #5819, opened 2018-12-16
Remove unnecessary encoding cookie from Python source files

jdufresne:
In Python 3, the default encoding of source files is utf-8. The encoding cookie is now unnecessary and redundant so remove it. For more details, see the docs:

https://docs.python.org/3/howto/unicode.html#the-string-type

> The default encoding for Python source code is UTF-8, so you can > simply include a Unicode character in a string literal ...

Includes a fix for the flake8 header checks to stop expecting an encoding cookie.


tk0miya (commented):
:-)

PR #5827, opened 2018-12-17
Refactor sphinx.io

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Now Sphinx uses custom docutils components which deviate the responsibility of them. Input is one of them.
- This moves event emitter to StandaloneReader component, and makes our Input simple.

### DIFF
--- a/doc/extdev/index.rst
+++ b/doc/extdev/index.rst
@@ -268,6 +268,11 @@ The following is a list of deprecated interfaces.
      - 3.0
      - N/A
 
+   * - ``sphinx.io.SphinxBaseFileInput``
+     - 2.0
+     - 3.0
+     - N/A
+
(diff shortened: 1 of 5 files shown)


doc/extdev/index.rst
```diff
@@ -268,11 +268,26 @@ The following is a list of deprecated interfaces.
...
+     - 2.0
+     - 3.0
+     - N/A
+
    * - ``sphinx.io.SphinxRSTFileInput``
      - 2.0
      - 3.0
      - N/A
 
+   * - ``sphinx.registry.SphinxComponentRegistry.add_source_input()``
```
tk0miya, line 286: I'd like to deprecate these classes and API in 3.0 because they are internal and young (added in 1.7).

PR #5828, opened 2018-12-17
Reorder the arguments for translators

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- The order for constructor of Translator is not standadized.
- This tries to standardize the order of them.

### DIFF
--- a/sphinx/util/docutils.py
+++ b/sphinx/util/docutils.py
@@ -392,8 +392,8 @@ class SphinxTranslator(nodes.NodeVisitor):
               This class is strongly coupled with Sphinx.
     """
 
-    def __init__(self, builder, document):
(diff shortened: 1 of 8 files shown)


sphinx/util/docutils.py
```diff
@@ -392,8 +392,8 @@ class SphinxTranslator(nodes.NodeVisitor):
               This class is strongly coupled with Sphinx.
     """
 
-    def __init__(self, builder, document):
-        # type: (Builder, nodes.document) -> None
+    def __init__(self, document, builder):
```
tk0miya, line 395: The base class of SphinxTranslator; NodeVisitor takes a document node on its instantiate. So it would be better to take a document node first.
For this reason, I determined to swap these arguments.

PR #5831, opened 2018-12-17
Deprecate quickstart.term_decode() and remove internal uses

jdufresne:
Per the Python 3 docs, `input()` always returns a string:

https://docs.python.org/3/library/functions.html#input

> The function then reads a line from input, converts it to a string (stripping a trailing newline), and returns that.

The stubs from typeshed say the same:

https://github.com/python/typeshed/blob/5c69373890dfaf4f07f0638766fb0a4903352892/stdlib/3/builtins.pyi#L835

Here is the implementation from CPython with also shows a call to `PyUnicode_Decode` on the result:

https://github.com/python/cpython/blob/3.7/Python/bltinmodule.c#L1960-L2143

As the value is always a string, there is nothing to decode. Therefore the call to `term_decode()` unnecessary and can safely be removed.

With this in mind, must adjust quickstart tests to be more representative.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -55,6 +55,8 @@ Deprecated
   ``IndexBuilder.feed()`` method is deprecated.
 * ``sphinx.addnodes.abbreviation``
 * ``sphinx.application.Sphinx._setting_up_extension``
+* ``sphinx.cmd.quickstart.term_decode()``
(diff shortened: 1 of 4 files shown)


tk0miya (approved):
LGTM!

PR #5832, opened 2018-12-17
Add bindep integration

stephenfin:
### Feature or Bugfix
- Feature

### Purpose
The bindep (BINary DEPendency automation) tool is a tool for checking
the presence of binary packages needed to use an application / library.
Use it to specify the dependencies needed to successfully pass the LaTex
tests. Only Fedora targets are currently provided.

### Relates
This is partial fix for #5823. A more complete fix would be to skip the tests if packages were missing.


bindep.txt
```diff
@@ -0,0 +1,15 @@
...
+texlive-wrapfig [platform:rpm]
+texlive-upquote [platform:rpm]
+texlive-capt-of [platform:rpm]
+texlive-needspace [platform:rpm]
+texlive-polyglossia [platform:rpm]
+texlive-luatex85 [platform:rpm]
+texlive-anyfontsize [platform:rpm]
+texlive-ctablestack [platform:rpm]
+texlive-gnu-freefont [platform:rpm]
+latexmk [platform:rpm]
```
tk0miya, line 15: Is python needed here?

tk0miya (approved):
+0; AFAIK no maintainer uses RedHat base OS. So contribution is needed to keep maintained this.

bindep.txt
stephenfin: It would be unnecessary. `bindep` relies on Python, so without that we can't use bindep :smile:

stephenfin:
> +0; AFAIK no maintainer uses RedHat base OS. So contribution is needed to keep maintained this.

If there is a list of Ubuntu/Debian packages that are required, it's possible to add them to the same file. As far as I can tell, `bindep` is available on Ubuntu.

jfbu:
I am not familiar with Fedora packaging of TeXLive. What happens if there is a local TeXLive installation done directly without the package manager? Will `bindep` then not wrongly complain that some TeX dependency is missing? Personally if I had a Fedora box, I would not rely on the Fedora provided LaTeX packaging but manage my own directly to have the latest one (and keep ones from earlier years).

stephenfin:
> I am not familiar with Fedora packaging of TeXLive. What happens if there is a local TeXLive installation done directly without the package manager? Will `bindep` then not wrongly complain that some TeX dependency is missing? Personally if I had a Fedora box, I would not rely on the Fedora provided LaTeX packaging but manage my own directly to have the latest one (and keep ones from earlier years).

It would, yes. I think if that were the case though, you're already beyond the need for `bindep`?

jfbu:
> It would, yes. I think if that were the case though, you're already beyond the need for `bindep`?

indeed, but only with respect to TeX dependencies... my  remark is rhetorical as I see here that your PR is indeed devoted fully to TeX dependencies.

We very rarely change the TeX dependencies but there will be need for some RedHat based distro-user for maintaining requirement file if that happens...

jfbu (approved):

bindep.txt
tk0miya: I see

tk0miya: I noticed graphviz and imagemagick are missing.
I think this Dockerfile is helpful for listing up dependencies.
https://github.com/sphinx-doc/docker-ci/blob/master/Dockerfile

tk0miya:
>If there is a list of Ubuntu/Debian packages that are required, it's possible to add them to the same file. As far as I can tell, bindep is available on Ubuntu.

My concern is how to maintain this rpm list. I understand this would be useful for new developers and package maintainers. But I worried it would not be maintained.

bindep.txt
stephenfin: Good point. I'll add those.

tk0miya:
any comments?

IMO, +0 if anyone will maintain the distro.

stephenfin:
@tk0miya For now, I can maintain the RPM list. Going forward, I'd like to do this via CI and will investigate this. If it turns out to be unmaintained, it's very easy to delete it again.

tk0miya:
Okay, let's merge this!

PR #5835, opened 2018-12-18
Use mathjax CDN "latest" URL by default

### LINKED ISSUE #5725: Use latest mathjax version by default
The current `sphinx.ext.mathjax` plugin uses mathjax version 2.7.1, see
https://github.com/sphinx-doc/sphinx/blob/c2fcb82221d4b6b2b57bc8cd01954eda239ad7e8/sphinx/ext/mathjax.py#L108

This version is not the most current one, and is known to have problems with accents in Safari (https://github.com/mathjax/MathJax/issues/1737) and on iOS (https://github.com/mathjax/MathJax/issues/2091).

I would advocate switching to `https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js` as the default `mathjax_path` (as recommended by http://docs.mathjax.org/en/latest/configuration.html). Anyone who needs to pin a specific version can still set that version in their `conf.py`, but for most people, using "latest" will probably cause the least problems.

goerz:
Closes #5725

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,7 @@ Features added
 Bugs fixed
 ----------
 
+* #5725: mathjax: Use CDN URL for "latest" version by default
 * #5460: html search does not work with some 3rd party themes
 * #5520: LaTeX, caption package incompatibility since Sphinx 1.6
 * #5614: autodoc: incremental build is broken when builtin modules are imported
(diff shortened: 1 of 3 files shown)


tk0miya (approved):

tk0miya:
Thanks!

PR #5839, opened 2018-12-19
Deprecate unused Exception attributes

jdufresne:
The attributes were used only for the string representation, but that is also the default behavior of the `Exception` class. Observe:

```
>>> str(Exception('foo'))
'foo'
>>> print(Exception('foo'))
foo
```


tk0miya (approved):
LGTM

PR #5842, opened 2018-12-21
Move immediate append operation to list literal

jdufresne:
Was previously separate for Python 2 support, but this is no longer
necessary. Can simply include the item in the literal.

https://github.com/sphinx-doc/sphinx/blob/v1.8.2/sphinx/util/__init__.py#L680-L681


tk0miya (changes requested):
Could you remove a comment above `name_start_chars` which mentions about py33 please?

jdufresne:
Done. :slightly_smiling_face:

tk0miya (approved):
Thank you for quick work!

PR #5852, opened 2018-12-22
LaTeX: move some hard-coded re-definitions to user hook for TOC

jfbu:
This is done to avoid hard-coded definitions of `\l@section` and `\l@subsection` currently done in ``sphinxmanual.cls`` and which potentially may complicate using custom document class. The definitions are moved into a new macro ``\sphinxtableofcontentshook``,  which only gets executed via ``\sphinxtableofcontents`` (it is executed in a scope limiting TeX group).

Defect is that if some user added some custom `\l@section` definition in preamble, this is now overwritten back to sphinx default, such custom definitions must be refactored as user re-definition of ``\sphinxtableofcontentshook``.

For the ``'howto'`` class, no such redefinitions were done by Sphinx, the new hook macro is also added but with empty contents to keep same output as current.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -45,6 +45,11 @@ Incompatible changes
   block.  As a result, they are highlighted as python3 by default.
 * The order of argument for ``HTMLTranslator``, ``HTML5Translator`` and
   ``ManualPageTranslator`` are changed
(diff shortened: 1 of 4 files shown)


tk0miya (approved):
LGTM!

jfbu:
thanks for review, merged!

Issue #5846, opened 2018-12-21
chm compiler can't deal with hex escaping


ghost:
### Problem
If hex escaping (e.g. `&#x27;`) appears in .hhc/.hhk files, chm compiler will simply ignore it.

Sadly, Python 3's [`html.escape(s, quote=True)`](https://github.com/python/cpython/blob/4a9ee26750aa8cb37b5072b2bb4dd328819febb4/Lib/html/__init__.py#L24) function generates a hex escaping `&#x27;` for single quote `'`.

This leading to a compiling issue for Python chm document, please see: https://bugs.python.org/issue35482

### Suggestion
`htmlhelp` builder uses `html.escape()` twice [\[1\]](https://github.com/sphinx-doc/sphinx/blob/96c5605374135652cf7ec0ad517b532b9261ef19/sphinx/builders/htmlhelp.py#L281)  [\[2\]](https://github.com/sphinx-doc/sphinx/blob/96c5605374135652cf7ec0ad517b532b9261ef19/sphinx/builders/htmlhelp.py#L314) for .hhc/.hhk files, we can make a wrapper function for `html.escape()` that convert hex escaping to decimal escaping, e.g. `&#x27;` -> `&#39;`

### Environment info
- OS: Win
- Python version: any
- Sphinx version: 1.8.2 / 2.0
- Microsoft HTML Help Compiler: 4.74.8702

PR #5853, opened 2018-12-23
[1.8] htmlhelp: convert hex escaping to decimal escaping in .hhc/.hhk files

ghost:
### Feature or Bugfix

- Bugfix

### Purpose
.hhc/.hhk files don't recognize hex escaping, we need convert hex escaping to decimal escaping.
For example: `&#x27;` -> `&#39;`

### Detail
If hex escaping (e.g. `&#x27;`) appears in .hhc/.hhk files, chm compiler will simply ignore it.

For example: use `&#x45;` to represent upper letter `E` in .hhc/.hhk files, this letter doesn't apper in chm content/index. After using its decimal form `&#69;` it appears as expected.

Sadly, Python 3's `html.escape(s, quote=True)` function generates a hex escaping `&#x27;` for single quote `'`. This leading to a compiling issue for Python chm document, please see: https://bugs.python.org/issue35482

### Relates
Issue: https://github.com/sphinx-doc/sphinx/issues/5846
Environment: Microsoft HTML Help Compiler: 4.74.8702

### DIFF
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -171,7 +171,7 @@ chm_locales = {
 
 
 def chm_htmlescape(*args, **kwargs):
-    # type: (unicode, Any) -> unicode
+    # type: (*unicode, **bool) -> unicode
     """
     chm_htmlescape() is a wrapper of htmlescape().
     .hhc/.hhk files don't recognize hex escaping, we need convert
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -180,7 +180,7 @@ def chm_htmlescape(*args, **kwargs):
     quote `'`, this wrapper fixes this.
     """
     def convert(matchobj):
-        # type: (re.Match) -> unicode
+        # type: (Match[str]) -> unicode
         codepoint = int(matchobj.group(1), 16)
         return '&#%d;' % codepoint
     return re.sub(r'&#[xX]([0-9a-fA-F]+);',
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -28,7 +28,7 @@ from sphinx.util.pycompat import htmlescape
 
 if False:
     # For type annotation
-    from typing import Any, Dict, IO, List, Tuple  # NOQA
+    from typing import Any, Dict, IO, List, Tuple, Match  # NOQA
     from sphinx.application import Sphinx  # NOQA
 
 
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -13,6 +13,7 @@ from __future__ import print_function
 
 import codecs
 import os
+import re
 from os import path
 
 from docutils import nodes
@@ -168,6 +169,18 @@ chm_locales = {
     'zh_TW': (0x404, 'cp950'),
 }
 
+# chm_htmlescape() is a wrapper of htmlescape()
+# .hhc/.hhk files don't recognize hex escaping, we need convert
+# hex escaping to decimal escaping. for example: `&#x27;` -> `&#39;`
+# htmlescape() may generates a hex escaping `&#x27;` for single
+# quote `'`, this wrapper fixes this.
+def chm_htmlescape(*args, **kwargs):
+    def convert(matchobj):
+        codepoint = int(matchobj.group(1), 16)
+        return '&#%d;' % codepoint
+    return re.sub(r'&#[xX]([0-9a-fA-F]+);',
+                  convert,
+                  htmlescape(*args, **kwargs))
 
 class HTMLHelpBuilder(StandaloneHTMLBuilder):
     """
@@ -278,7 +291,7 @@ class HTMLHelpBuilder(StandaloneHTMLBuilder):
                         write_toc(subnode, ullevel)
                 elif isinstance(node, nodes.reference):
                     link = node['refuri']
-                    title = htmlescape(node.astext()).replace('"', '&quot;')
+                    title = chm_htmlescape(node.astext()).replace('"', '&quot;')
                     f.write(object_sitemap % (title, link))
                 elif isinstance(node, nodes.bullet_list):
                     if ullevel != 0:
@@ -311,7 +324,7 @@ class HTMLHelpBuilder(StandaloneHTMLBuilder):
                     item = '    <param name="%s" value="%s">\n' % \
                         (name, value)
                     f.write(item)
-                title = htmlescape(title)
+                title = chm_htmlescape(title)
                 f.write('<LI> <OBJECT type="text/sitemap">\n')
                 write_param('Keyword', title)
                 if len(refs) == 0:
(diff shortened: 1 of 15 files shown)


sphinx/builders/htmlhelp.py
```diff
@@ -168,6 +169,18 @@
     'zh_TW': (0x404, 'cp950'),
 }
```
tk0miya, line 171: Could you add 2 blank lines before a function and classes?
flake8 warns here and before `HTMLHelpBuilder` class.
https://travis-ci.org/sphinx-doc/sphinx/jobs/471503849#L520-L521

sphinx/builders/htmlhelp.py
```diff
@@ -168,6 +169,18 @@
...
 
+# chm_htmlescape() is a wrapper of htmlescape()
+# .hhc/.hhk files don't recognize hex escaping, we need convert
+# hex escaping to decimal escaping. for example: `&#x27;` -> `&#39;`
+# htmlescape() may generates a hex escaping `&#x27;` for single
+# quote `'`, this wrapper fixes this.
```
tk0miya: Please move this to a docstring of `chm_htmlescape()`.

sphinx/builders/htmlhelp.py
```diff
@@ -168,6 +169,18 @@
...
+def chm_htmlescape(*args, **kwargs):
+    def convert(matchobj):
```
tk0miya, line 182: Please add annotations to these functions.

tk0miya (changes requested):
Looks good. It would be nice if there is testcases. Could you try it please?

animalize:
Now it's prepared for review.
Sorry for my unskilled.

ghost:
Now it's green, sorry for my unskilled.

tests/roots/test-root/markup.txt
```diff
@@ -379,6 +379,10 @@ Index markup
    !Main, !Other
    !single: entry; pair
 
+.. index:: triple-quoted string, Unicode Consortium, raw string
+   single: """; string literal
+   single: '''; https://github.com/sphinx-doc/sphinx/issues/5846
```
tk0miya: Oh, sorry. Please add new test example to `tests/roots/test-[new-test-directory]`. It is commonly used testcase. So I'd like to keep as is.

In addition, you need to add testcase in `tests/test_build_htmlhelp.py`.

ghost: 1.8 branch doesn't have `tests/test_build_htmlhelp.py`.
Even in 2.0 branch, [`tests/test_build_htmlhelp.py`](https://github.com/sphinx-doc/sphinx/blob/master/tests/test_build_htmlhelp.py) is amlost empty.

I don't know how to create a `htmlhelp` testcase...

I replace installed `sphinx/builders/htmlhelp.py ` with this version, then I can build a good `python372rc1.chm`, can this be seem as a test?

tk0miya: Yes, it is manual testing. But we need continuous testing via test code.

Is it hard to check generated `.chm` file? This PR changes the escaping algorithm. It should be confirmed in the case.

ghost: Testcase added.

tk0miya (approved):
LGTM!

tk0miya:

sphinx/builders/htmlhelp.py
```diff
@@ -169,6 +170,24 @@
...
+    .hhc/.hhk files don't recognize hex escaping, we need convert
+    hex escaping to decimal escaping. for example: `&#x27;` -> `&#39;`
+    htmlescape() may generates a hex escaping `&#x27;` for single
+    quote `'`, this wrapper fixes this.
+    """
+    def convert(matchobj):
+        # type: (Match[unicode]) -> unicode
+        codepoint = int(matchobj.group(1), 16)
+        return '&#%d;' % codepoint
+    return re.sub(r'&#[xX]([0-9a-fA-F]+);',
```
tk0miya, line 186: I noticed `html.escape()` only generates `&#x27;`. Other characters (including unicode) are not changed.
Is this regexp correct? Is it better to use `str.replace()` instead like this?
```
def chm_htmlescape(s):
    s = htmlescape(s)
    s = s.replace('&#x27;', '&#27;')  # re-escape as decimal
    return s
```

ghost: Yes, that works, but I want to sovle this problem completely.

If a document use `&#x45;` to represent upper letter `E`, `&#x45;` will appear in .hhc/.hhk files, so `E` doesn't display in chm's content/index. After using its decimal form `&#69;` it displays as expected. Though this kind of document is very rare.

ghost: Maybe you are right, `&#x45;` will be escaped to `&amp;#x45;` by the `Writer`, so we don't need worry about it...

tk0miya: I found this is current behavior:
```
def test_chm_htmlescape():
    assert chm_htmlescape('Hello world') == 'Hello world'
    assert chm_htmlescape(u'Unicode 文字') == u'Unicode 文字'

    if PY2:
        assert chm_htmlescape('<Hello> "world"') == '&lt;Hello&gt; "world"'
        assert chm_htmlescape('<Hello> "world"', True) == '&lt;Hello&gt; &quot;world&quot;'
        assert chm_htmlescape('<Hello> "world"', False) == '&lt;Hello&gt; "world"'
    else:
        assert chm_htmlescape('<Hello> "world"') == '&lt;Hello&gt; &quot;world&quot;'
        assert chm_htmlescape('<Hello> "world"', True) == '&lt;Hello&gt; &quot;world&quot;'
        assert chm_htmlescape('<Hello> "world"', False) == '&lt;Hello&gt; "world"'

    if PY2:
        assert chm_htmlescape("Hello 'world'") == "Hello 'world'"
        assert chm_htmlescape("Hello 'world'", True) == "Hello 'world'"
        assert chm_htmlescape("Hello 'world'", False) == "Hello 'world'"
    else:
        assert chm_htmlescape("Hello 'world'") == "Hello &#39;world&#39;"
        assert chm_htmlescape("Hello 'world'", True) == "Hello &#39;world&#39;"
        assert chm_htmlescape("Hello 'world'", False) == "Hello 'world'"
```

It seems single quote does not escaped on py2. Is this intended? If none, I'll fix it also from now on.

ghost: `htmlescape()` is defined in file `sphinx.util.pycompat.py`:
```python
# htmlescape()
if PY3:
    from html import escape as htmlescape
else:
    from cgi import escape as htmlescape  # NOQA
```

In py3, we use `html.escape()` from stdlib, it convert signle quote if `quote=True`.
https://github.com/python/cpython/blob/4a9ee26750aa8cb37b5072b2bb4dd328819febb4/Lib/html/__init__.py#L24
This function escapes single quote.

In py2, we use `cgi.escape()` from stdlib:
https://github.com/python/cpython/blob/bc6f74a520112d25ef40324e3de4e8187ff2835d/Lib/cgi.py#L1063-L1072
This function doesn't escape single quote.

I think it's not a matter if we always use `"` to wrap attributes in .html templates.

tk0miya: Okay, I'll keep this as is.

PR #5854, opened 2018-12-23
[2.0] htmlhelp: convert hex escaping to decimal escaping in .hhc/.hhk files

ghost:
### Feature or Bugfix

- Bugfix

### Purpose
.hhc/.hhk files don't recognize hex escaping, we need convert hex escaping to decimal escaping.
For example: `&#x27;` -> `&#39;`

### Detail
If hex escaping (e.g. `&#x27;`) appears in .hhc/.hhk files, chm compiler will simply ignore it.

For example: use `&#x45;` to represent upper letter `E` in .hhc/.hhk files, this letter doesn't apper in chm content/index. After using its decimal form `&#69;` it appears as expected.

Sadly, Python 3's `html.escape(s, quote=True)` function generates a hex escaping `&#x27;` for single quote `'`. This leading to a compiling issue for Python chm document, please see: https://bugs.python.org/issue35482

### Relates
Issue: https://github.com/sphinx-doc/sphinx/issues/5846
Environment: Microsoft HTML Help Compiler: 4.74.8702

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -144,6 +144,7 @@ Bugs fixed
   if passed as language option?
 * #5179: LaTeX: (lualatex only) escaping of ``>`` by ``\textgreater{}`` is not
   enough as ``\textgreater{}\textgreater{}`` applies TeX-ligature
+* #5846: htmlhelp: convert hex escaping to decimal escaping in .hhc/.hhk files
 
 Testing
 --------
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -11,6 +11,7 @@
 
 import html
 import os
+import re
 from os import path
 
 from docutils import nodes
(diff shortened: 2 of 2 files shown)


tk0miya:
Thank you for PR. You don't need to post bugfixes to both 1.8 and 2.0. I'll merge #5853 into 1.8 branch. And it will be merged into msater branch on merging 1.8 into master.
Thanks,

ghost:
I'm afraid there is a code conflict.
2.0 branch is using `html.escape()` from Python 3 stdlib, 1.8 branch is using `htmlescape()` comes from  `from sphinx.util.pycompat import htmlescape`

tk0miya:
I know that. But we usually merge 1.8 branch into master. So #5853 will be merged automatically in our development flow. As a result, this is not needed.

PR #5857, opened 2018-12-23
Remove instructions to install Python 2 on Windows

jdufresne:
As only Python 3 is supported, should not advise users to install Python 2.


tk0miya (approved):
:-)

PR #5858, opened 2018-12-23
Remove unnecessary enum existence guard

jdufresne:
The enum module is always available since Python 3.4.


tk0miya (approved):

tk0miya:
Merged. Thank you always

Issue #5788, opened 2018-12-15
Add htmlhelp_ascii_output option to htmlhelp builder


ghost:
Subject: non-ASCII characters can not display in chm

### Problem
chm is an old tech from 20 years ago, maybe for backward compatibility reason, `hh.exe` using system's local encoding to render chm content by default.

Then some non-ASCII characters display as mojibakes, for example if a page's charset is cp1252, and it has a string `classical “sequential” programming`, (`“` \u201C and `”` \u201D are legal in cp1252).
Now we open this chm in a Simplified Chinese Windows, then `hh.exe` using gbk as default encoding, `“s` (0x93 0x73 byte sequence) is decoded as `搒`, `” ` (0x94 0x20 byte sequence) is decoded as `�`, the entire string looks like `classical 搒equential� programming`.

Python's chm doc suffered this problem about one year, see https://bugs.python.org/issue32174 , you can see some screenshots in this issue.

It seems impossible to specify the encoding of .chm to ASCII [1], the available encodings of .chm are limited to a list [2].

[1] https://github.com/sphinx-doc/sphinx/blob/96c5605374135652cf7ec0ad517b532b9261ef19/sphinx/builders/htmlhelp.py#L203-L205
[2] https://github.com/sphinx-doc/sphinx/blob/96c5605374135652cf7ec0ad517b532b9261ef19/sphinx/builders/htmlhelp.py#L136-L169

### Suggestion
We can add a `htmlhelp_ascii_output` option to htmlhelp builder, when this is set to `True`, the output html using ASCII encoding, it escapes every non-ASCII characters to ASCII form, then we can get the right visual effect everywhere.

I wrote a patch to demonstrate this idea, see https://github.com/sphinx-doc/sphinx/pull/5808/commits/2fc00f4ec8945422e2558ecef9843c6b6e1da298
Do you like this idea?

This is one screenshot:
![Python chm](https://raw.githubusercontent.com/animalize/pics/master/issues/chm.png)

animalize:
not prepared yet, close it.

ghost:
> What "Binary" means? Is there any description for these options?

There is another hh.exe problem. "Index search in CHM help crashes the viewer", after turning "Binary Index" to "Yes", hh.exe no longer crash.

Please read: https://bugs.python.org/issue35228
So I add this option.

> If my understanding is right, this is a hack to choose encoding via configuration. If true, I feel htmlhelp_encoding is much better for this case.

[...]

PR #5863, opened 2018-12-24
Add htmlhelp_ascii_output option to htmlhelp builder

ghost:
Subject: Add `htmlhelp_ascii_output option` to htmlhelp builder

### Feature or Bugfix
- Feature

### Purpose
`htmlhelp_ascii_output` option escapes non-ASCII characters after `<head>` tag in .html files to ASCII.
For example: `ß` -> `&szlig;`

This will fix .chm's visual effect on some MBCS Windows systems.

### Detail
please read https://github.com/sphinx-doc/sphinx/issues/5788

### Relates
https://github.com/sphinx-doc/sphinx/issues/5788

### DIFF
--- a/sphinx/builders/htmlhelp.py
+++ b/sphinx/builders/htmlhelp.py
@@ -19,6 +19,12 @@ from os import path
 from docutils import nodes
 from six import PY3
 
+# codepoint2name dict
+if PY3:
+    from html.entities import codepoint2name
+else:
+    from htmlentitydefs import codepoint2name
+
 from sphinx import addnodes
 from sphinx.builders.html import StandaloneHTMLBuilder
 from sphinx.environment.adapters.indexentries import IndexEntries
@@ -237,7 +243,7 @@ class HTMLHelpBuilder(StandaloneHTMLBuilder):
         # type: (unicode) -> unicode
         """If ``htmlhelp_ascii_output`` is True, non-ASCII characters
            after <head> tag will be escaped to ASCII form.
-           e.g. ``ß`` -> ``&#223;``.
+           e.g. ``ß`` -> ``&szlig;``.
         """
         # only process when htmlhelp_ascii_output is True
         if not self.config.htmlhelp_ascii_output:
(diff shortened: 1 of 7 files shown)


sphinx/builders/htmlhelp.py
```diff
@@ -232,6 +239,36 @@ def update_page_context(self, pagename, templatename, ctx, event_arg):
...
+                codepoint = ord(matchobj.group(0))
+                name = codepoint2name.get(codepoint)
+                if name is None:
+                    return '&#%d;' % codepoint
+                else:
+                    return '&%s;' % name
+            return re.sub(r'[^\x00-\x7F]', escape, string)
+
+        # escape non-ASCII characters after head tag
+        m = re.search(r'^.*?</head>', htm, re.DOTALL|re.IGNORECASE)
```
tk0miya: I suppose `<head>` tag can be processed by htmlentities. It does not cause moji-bakes to CP1252 encodings.

ghost: >I suppose <head> tag can be processed by htmlentities.

not so, it seems `hh.exe` only uses `cp1252` encoding to process meta data (inside `<head>` tag) of .html..

tk0miya:
I feel there are no reason to add a new option `htmlhelp_ascii_output`.
We can replace non-ASCII characters always when encoding is cp1251.
What do you think this idea?
https://github.com/tk0miya/sphinx/commit/570f789773c1506a562990117d20655c972da1f4

ghost:
>We can replace non-ASCII characters always when encoding is cp1251.

Is it `cp1252`? `cp1251` is for Cyrillic languages (Russian, Bulgarian...).

If escape the entire .html file, it's the same as **Method B**, squares will display in search result.
We need keep `<head>` tag using the specified encoding, e.g. `cp1252`.
Anyway, you may leave this issue alone, the priority is not very high.

sphinx/builders/htmlhelp.py
tk0miya: OMG... hh.exe is too tricky...

tk0miya:
>Is it cp1252? cp1251 is for Cyrillic languages (Russian, Bulgarian...).

oh, sorry. it's typo.

sphinx/builders/htmlhelp.py
ghost: chm is from 20 years ago, a pre-Unicode age, it seems Microsoft doesn't develop it since long ago.
But it's a successful product, people still using it.
It would be nice to have a modern replacement for chm.

Issue #3707, opened 2017-05-05
LaTeX replacement of `✓` by macro `\checkmark` can not be boldened


jfbu:
At https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/texescape.py#L39, there is 

```
 ('✓', r'\(\checkmark\)'),
```

There are some issues:

1. ``\checkmark`` which is provided by `amssymb` can work in math mode as well as in text mode to the same result. Thus the `\(..\)` are superfluous and would raise an error if the thing was encountered while being already in math mode

2. There is no way to bolden the character: it does not react to `\textbf`,  `\boldmath`, etc.. So it does not provide candidate for the heavy Unicode variant ✔.

3. The replacement itself is unfriendly to lualatex, xelatex users, as OpenType fonts will have the character (often). And even with pdflatex, there are better ways, for example `\usepackage{pifont}` and `\ding{51}`. Then one can use `\ding{52}`  for ✔ (via `\newunicodechar{✔}{\ding{52}}`.

### Environment info
- Sphinx version: 1.5.5

tk0miya:
I don't have strong opinion for this.

Note:

1. `texescape.py` is not applied to math nodes. so it does not cause an error. Of course, removing is okay to me.
2. No idea.
3. I'm okay to switch a rule of escaping by `latex_engine` :-)

PR #5867, opened 2018-12-27
LaTeX: escape Unicode HEAVY CHECK MARK into latex macros

jfbu:
Closes: #3707

The `\pmb` macro (aka "poor man bold") is provided with `amsmath` package so this adds no new dependency, contrarily to for example using `pifont` package.

For the record, the https://tex.stackexchange.com/a/383798 solution based on `pdfrender` package, does not work with `platex+dvipdfmx` (Japanese).

Arguably, xelatex and lualatex would be better with no TeX escaping of the ✔ character, but this may be delayed to some future refactoring of LaTeX escaping depending on engine.

The defect of this proposal is that ``**✓**`` input will still give a non-bold checkmark in PDF (`\checkmark` macro does not react to `\textbf`), the input must use the ✔ Unicode literal.


tk0miya (approved):
+1

Issue #5871, opened 2019-01-02
`TexinfoTranslator` won't allow section name of `.`


brandonwillard:
Subject: `TexinfoTranslator` won't allow section name of `.` 

### Problem
- `TexinfoTranslator.escape_id` strips `.` from possible node names.

#### Procedure to reproduce the problem
```bash
> cd /tmp
> yes n | sphinx-quickstart -p test -a test --dot _ -v 0 -r 0 -l rst --master index --suffix .rst --makefile ./
> echo "\n.\n-\n\nThis is a subsection with a ``.`` as the title." >> index.rst
> make info
```

#### Error logs / results
```make
Running Sphinx v1.8.3
loading translations [rst]... not available for built-in messages
making output directory...
building [mo]: targets for 0 po files that are out of date
building [texinfo]: all documents
updating environment: 1 added, 0 changed, 0 removed
reading sources... [100%] index                                                                          
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
processing test.texi...index 
resolving references...
writing... 
Exception occurred:
  File "/home/bwillard/apps/anaconda3/lib/python3.6/site-packages/sphinx/writers/texinfo.py", line 309, in collect_node_menus
    assert 'node_name' in node and node['node_name']
AssertionError
The full traceback has been saved in /tmp/user/1000/sphinx-err-hqoqrlb7.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [Makefile:19: info] Error 1
```

#### Expected results
Should produce a `test.info` file containing the following:
```info
@menu
* .:: 

@end menu

@node .,,,Indices and tables
@anchor{index id1}@anchor{3}
@chapter .


This is a subsection with a . as the title.
```
### Reproducible project / your project
- https://github.com/hylang/hy/issues/1717

### Environment info
- OS: Ubuntu 18.10
- Python version: Python 3.6.7 :: Anaconda, Inc.
- Sphinx version: 2.0.0+/58f4c658f
- texi2any (GNU texinfo) 6.5

tk0miya:
Fixed by #6007.
Thanks,

PR #5872, opened 2019-01-02
Allow periods in texinfo node names

brandonwillard:
Subject: Allow periods in texinfo node names

### Feature or Bugfix
- Bugfix

### Purpose
- Allow `.` in texinfo node names

### Relates
- #5871

### DIFF
--- a/sphinx/writers/texinfo.py
+++ b/sphinx/writers/texinfo.py
@@ -387,7 +387,7 @@ class TexinfoTranslator(SphinxTranslator):
     def escape_id(self, s):
         # type: (str) -> str
         """Return an escaped string suitable for node names and anchors."""
-        bad_chars = ',:.()'
+        bad_chars = ',:()'
         for bc in bad_chars:
             s = s.replace(bc, ' ')
         s = ' '.join(s.split()).strip()


tk0miya:
Thank you for pull request.
What name is generated for the "." section with this patch? Are hyperlinks for the section works fine?
In addition, I'd like to know what is happened for other bad characters.
Thanks,

brandonwillard:
With this change, the section name is just `.` and links appear to work as expected for the resulting `.info` file.

The same error as https://github.com/sphinx-doc/sphinx/issues/5871 is produced for `,` and&mdash;assumedly&mdash;the other `bad_char` characters in `TexinfoTranslator.escape_id`.

tk0miya:
Indeed, this PR allow to build .texi file on Sphinx. But it seems menus and navigations are broken after compile by makeinfo command.

This is the requirements of node name.

>Unfortunately, you cannot reliably use periods, commas, or colons within a node name; these can confuse the Info reader. Also, a node name may not start with a left parenthesis preceding a right parenthesis, as in (not)allowed, since this syntax is used to specify an external manual. (Perhaps these limitations will be removed some day.)
https://www.gnu.org/software/texinfo/manual/texinfo/html_node/Node-Line-Requirements.html#Node-Line-Requirements

I think this change is not good for texinfo format...

brandonwillard:
It says you can't _reliably_ use it and that it will only emit a warning&mdash;not entirely break compilation like it currently does before attempting texinfo compilation.  

Also, the remaining part of that quote is as follows:

>...
>makeinfo warns about such problematic usage in node names, menu items, and cross-references. If you don’t want to see the warnings, you can set the customization variable INFO_SPECIAL_CHARS_WARNING to ‘0’ (see Other Customization Variables).
>
>Also, if you insist on using these characters in node names (accepting the resulting substandard Info output), in order not to confuse the Texinfo processors you must still escape those characters, by using either special insertions (see Inserting a Comma) or @asis (see @asis). For example:
>
>@node foo@asis{::}bar
>As an example of avoiding the special characters, the following is a section title in this manual:
>
>@section @code{@@unnumbered}, @code{@@appendix}: ...
>But the corresponding node name lacks the commas and the subtitle:
>
>@node @unnumbered @appendix

In other words, it looks like we can take their recommendation and simply change this PR to make use of `@asis` and let users produce substandard Info output without Sphinx-induced crashes.  If you're fine with that, I'll make the change.

brandonwillard:
What about changing these problem-characters to some other valid set of node name characters: e.g. `.` to `Period` or `<Period>`?  That way, links would work (although the output would most likely be unexpected).

tk0miya:
Sorry for late response. I just merged this with a little change (see #6007).
It will be released soon as 1.8.4.
Thank you for your help :-)

brandonwillard:
Gladly; thanks for making it happen!

PR #3584, opened 2017-03-24
apidoc: Fix bad escaping of input in conf.py_t

eric-wieser:
Otherwise paths like `somefolder\test` would be treated as containing a tab
character

PR #5881, opened 2019-01-03
refactor: Use repr() to make escaped string in template for conf.py

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Use repr() to make escaped string in template for conf.py
- alternative of #3584 
  - We can use simple `repr()` because we'd dropped py2 support!


jdufresne (approved):

shimizukawa:
My question is: as a Python specification, is this certainly true? `exec(repr(obj)) == obj`
In this PR case, maybe OK. I will investigate the python reference to solve that concern.

shimizukawa:
https://docs.python.org/3.6/library/functions.html#repr
> For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to eval(),

`For may types`.

https://docs.python.org/3.6/reference/datamodel.html#object.__repr__
> Called by the repr() built-in function to compute the “official” string representation of an object. If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment).

`that could be used to recreate an object`

In my investigation, there is no fact that it is not good to pass a `str` to `repr`.
(On py2, unexpected result case is exist: https://github.com/sphinx-doc/sphinx/pull/3584#issuecomment-295736278 )

shimizukawa (approved):
LGTM!

tk0miya:
Thank you for reviewing carefully :-)

PR #5890, opened 2019-01-04
Fix #5889: LaTeX: user numfig_format is stripped of spaces and may cause build failure

### LINKED ISSUE #5889: LaTeX: user ``numfig_format`` is stripped of spaces and may cause build failure

`index.rst`:

```
====
test
====

Section
=======

.. figure:: foo.png

   is a figure

```

`conf.py`:

```
# -*- coding: utf-8 -*-

master_doc = 'index'
html_theme = 'classic'
numfig = True
numfig_format = {
    'figure': 'F %s F',
    'table':  'T %s T',
    'code-block': 'L %s L',
    'section': 'S %s S',
}
```

Do `make latexpdf`.

#### Error logs / results
```
! Undefined control sequence.
\fnum@figure ->\figurename \thefigureF 
                                       
l.85 \caption{is a figure}
                          \label{\detokenize{index:id1}}\end{figure}
? 
```

#### Expected results

PDF via LaTeX backends works: the stripped space before `F` causes `\thefigureF` non-existent macro to be used.

Remark1: by the way the default definition of `\fnum@figure` is `\figurename\nobreakspace\thefigure`. This explains why the default `Fig. %s` setting gives nevertheless a space in output.

Remark2: the HTML backend preserves the spaces.

### Environment info
- OS: Mac OS
- Python version: 3.6.6
- Sphinx version: 1.8 branch at 7ff7dfc4 (1.8.4+)
- TeXLive 2018

Related: #5516, which migrates code that this issue, once fixed, will modify. Thus merging to master will need some care.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5889

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -796,7 +796,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
             ret.append(self.babel_renewcommand('\\figurename', definition))
             if figure[1]:
                 ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@figure{\\figurename\\thefigure%s}\n' %
+                ret.append('\\def\\fnum@figure{\\figurename\\thefigure\\relax{}%s}\n' %
                            text_type(figure[1]).strip().translate(tex_escape_map))
                 ret.append('\\makeatother\n')
 
@@ -809,7 +809,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
             ret.append(self.babel_renewcommand('\\tablename', definition))
             if table[1]:
                 ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@table{\\tablename\\thetable%s}\n' %
+                ret.append('\\def\\fnum@table{\\tablename\\thetable\\relax{}%s}\n' %
                            text_type(table[1]).strip().translate(tex_escape_map))
                 ret.append('\\makeatother\n')
 
--- a/CHANGES
+++ b/CHANGES
@@ -15,8 +15,11 @@ Features added
 
 Bugs fixed
 ----------
+
 * #5605: with the documentation language set to Chinese, English words could not
   be searched.
+* #5889: LaTeX: user ``numfig_format`` is stripped of spaces and may cause build
+  failure
 
 Testing
 --------
--- a/tests/test_build_latex.py
+++ b/tests/test_build_latex.py
@@ -289,9 +289,9 @@ def test_numref_with_prefix2(app, status, warning):
     print(status.getvalue())
     print(warning.getvalue())
     assert '\\addto\\captionsenglish{\\renewcommand{\\figurename}{Figure:}}' in result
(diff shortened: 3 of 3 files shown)


sphinx/writers/latex.py
```diff
@@ -796,7 +796,7 @@ def generate_numfig_format(self, builder):
             ret.append(self.babel_renewcommand('\\figurename', definition))
             if figure[1]:
                 ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@figure{\\figurename\\thefigure%s}\n' %
+                ret.append('\\def\\fnum@figure{\\figurename\\thefigure\\relax{}%s}\n' %
```
jfbu, line 799: It is enough to use ``\\thefigure %s``.

sphinx/writers/latex.py
```diff
@@ -809,7 +809,7 @@ def generate_numfig_format(self, builder):
             ret.append(self.babel_renewcommand('\\tablename', definition))
             if table[1]:
                 ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@table{\\tablename\\thetable%s}\n' %
+                ret.append('\\def\\fnum@table{\\tablename\\thetable\\relax{}%s}\n' %
```
jfbu, line 812: same remark, ``\thetable %s`` fixes it too and is simpler. Now if `%s` was a space this would not give a space in output. But currently `%s` is stripped (which I find is disputable) so it does not matter. Probably best is ``\thetable{}`` (no need for ``\relax``, only ``{}``) because if `%s` is a space it will be obeyed. I think LaTeX should behave like HTML and in my brief testing HTML does not strip spaces.

sphinx/writers/latex.py
```diff
@@ -796,7 +796,7 @@ def generate_numfig_format(self, builder):
             ret.append(self.babel_renewcommand('\\figurename', definition))
             if figure[1]:
```
jfbu, line 797: the difficulty is that we don't renew `\fnum@figure` it will be `\figurename\nobreakspace\thefigure`. This might be the reason why `\figurename` was fetched from `numfig_format` via action of `strip()`. I think preferable to not `strip()` and always redefine `\fnum@figure` without the `\nobreakspace`.

jfbu (commented):
LGTM with nits. But I am worried about usage of `strip()`. I would prefer not to use `strip()` and leave spaces to user responsability when setting up `numfig_format`. This way LaTeX will behave as HTML. Currently ``MyFigName %s`` has the space stripped then re-installed via default `\fnum@figure` which inserts `\nobreakspace` giving a space in PDF. And usage of something after ``%s`` was never done by users else #5889 would have been seen. So I feel it is safe to not strip spaces but always redefine `\fnum@figure`, `\fnum@table`, ...

jfbu:
I edited the PR to be for merge into 1.8 not master, and then I closed and reopened in the hope to trigger CI tests, but this did not trigger it. An alternative PR is #5891.

PR #5891, opened 2019-01-04
LaTeX: obey spaces from `numfig_format` (figure and table only)

jfbu:
Only for figure and table, currently. Code-bloks have some FIXME code to
still be handled.

PDF output should remain unchanged so this is backwards compatible. The
`\nobreakspace` from LaTeX default `\fnum@figure` and replaced by a normal
space (if `numfig_format` left to its `Fig. %s` default).

If figname is an abbreviation (as in default), the dot is followed by a `\@` to
avoid double space. At the location where the `\@` is inserted the `@`
has catcode non-letter so a space after it is not swallowed and remains significant.

MEMO: when this is will be moved to sphinxmessages.sty on master branch
it will be needed to use `\@{}` or to modify locally catcode of `@` to avoid
space swallowing, because then `@` will have catcode letter.

Closes: #5889

- Bugfix

### Relates
- #5890 (see comments in the review I did of it)

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -796,7 +796,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
             ret.append(self.babel_renewcommand('\\figurename', definition))
             if figure[1]:
                 ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@figure{\\figurename\\thefigure%s}\n' %
+                ret.append('\\def\\fnum@figure{\\figurename\\thefigure{}%s}\n' %
                            text_type(figure[1]).strip().translate(tex_escape_map))
                 ret.append('\\makeatother\n')
 
@@ -809,7 +809,7 @@ class LaTeXTranslator(nodes.NodeVisitor):
             ret.append(self.babel_renewcommand('\\tablename', definition))
             if table[1]:
                 ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@table{\\tablename\\thetable%s}\n' %
+                ret.append('\\def\\fnum@table{\\tablename\\thetable{}%s}\n' %
                            text_type(table[1]).strip().translate(tex_escape_map))
                 ret.append('\\makeatother\n')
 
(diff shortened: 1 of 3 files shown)


jfbu (commented):
I left the branches with `if len(figure) == 1` as is, as anyhow they are destined to be removed, as `%s` is mandatory (but our doc should make it clearer)

sphinx/writers/latex.py
```diff
@@ -792,26 +792,24 @@ def generate_numfig_format(self, builder):
...
-            definition = text_type(figure[0]).strip().translate(tex_escape_map)
+            definition = escape_abbr(text_type(figure[0]).translate(tex_escape_map))
             ret.append(self.babel_renewcommand('\\figurename', definition))
-            if figure[1]:
-                ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@figure{\\figurename\\thefigure%s}\n' %
-                           text_type(figure[1]).strip().translate(tex_escape_map))
-                ret.append('\\makeatother\n')
+            ret.append('\\makeatletter\n')
+            ret.append('\\def\\fnum@figure{\\figurename \\thefigure{}%s}\n' %
```
jfbu: actually there is no reason why I inserted a space after `\\figurename` here, it is not significant and swallowed by TeX, independently of actual contents of `\\figurename` macro. I will remove that.

sphinx/writers/latex.py
```diff
@@ -792,26 +792,24 @@ def generate_numfig_format(self, builder):
...
-            definition = text_type(table[0]).strip().translate(tex_escape_map)
+            definition = escape_abbr(text_type(table[0]).translate(tex_escape_map))
             ret.append(self.babel_renewcommand('\\tablename', definition))
-            if table[1]:
-                ret.append('\\makeatletter\n')
-                ret.append('\\def\\fnum@table{\\tablename\\thetable%s}\n' %
-                           text_type(table[1]).strip().translate(tex_escape_map))
-                ret.append('\\makeatother\n')
+            ret.append('\\makeatletter\n')
+            ret.append('\\def\\fnum@table{\\tablename \\thetable{}%s}\n' %
```
jfbu: Here too, there is no reason for the added space after `\\tablename` I will remove.

jfbu (commented):
will push a commit to remove unneeded spaces I used after LaTeX macro name

tk0miya (commented):
LGTM!

jfbu:
Thanks for reviewing, merging...

PR #5892, opened 2019-01-05
Merge into master 5891

jfbu:
This is merge of 1.8 at e44f56a8 into master at 833e5ef78

I needed to add some changes to extra files, can you please check this.


tests/test_build_latex.py
```diff
@@ -276,7 +276,7 @@ def test_numref_with_prefix1(app, status, warning):
     result = (app.outdir / 'sphinxmessages.sty').text(encoding='utf8')
     print(result)
     assert r'\addto\captionsenglish{\renewcommand{\figurename}{Figure:}}' in result
-    assert r'\addto\captionsenglish{\renewcommand{\tablename}{Tab_}}' in result
+    assert r'\addto\captionsenglish{\renewcommand{\tablename}{Tab\_}}' in result
```
jfbu, line 279: notice that absence of TeX escaping was an unsignaled bug

sphinx/writers/latex.py
```diff
@@ -2571,26 +2571,24 @@ def generate_numfig_format(self, builder):
             ret.append('\\def\\fnum@figure{%s}\n' %
                        str(figure[0]).strip().translate(tex_escape_map))
         else:
-            definition = str(figure[0]).strip().translate(tex_escape_map)
+            definition = escape_abbr(str(figure[0]).translate(tex_escape_map))
```
jfbu, line 2574: the `escape_abbr` here is the one from latex.py, as it is for backwards compatibility it is left unmodified (it does not need `\@{}` because it is used with `@` of catcode "other" so `\@` does not swallow space after it)

sphinx/util/texescape.py
```diff
@@ -81,6 +84,12 @@ def escape(s):
     return s.translate(tex_escape_map)
 
 
+def escape_abbr(text):
+    # type: (str) -> str
+    """Adjust spacing after abbreviations. Works with @ letter or other."""
+    return re.sub(r'\.(?=\s|$)', r'.\@{}', text)
```
jfbu, line 90: as this will be used in `sphinxmessages.sty` where `@` is of catcode "letter", it is needed to use `\@{}` to avoid space swallowing. The `\@{}` would work also fine if `@` is of catcode "other".

sphinx/writers/latex.py
jfbu: all this code originates from 1.8 branch after #5891 merge

jfbu:
if ok I will do this merge and push to repo, so that there is no extra merge in history of this dummy PR

tk0miya (approved):
LGTM!

PR #5516, opened 2018-10-07
Move labels related with numfig_format to sphinxmessage.sty

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Move labels for `numfig_format` to `sphinxmessage.sty`
- ~Still WIP. This contains #5515.~

PR #5894, opened 2019-01-05
LaTeX: fix ``addtocaptions`` for sphinxmessages.sty_t if 'babel' empty

jfbu:
For example if config setting `language` is 'ja' one will see in `sphinxmessages.sty` in latex build repertory:

```
{\renewcommand{\figurename}{図}}

{\renewcommand{\tablename}{表}}
```

but the extra braces annihilate effect of `\renewcommand`.  This was introduced in #5516, sorry for not having paid attention to this.

I use `\@iden` (same as `\@firstofone`) which is LaTeX macro removing braces. We can use `@` letter because we are now in a package file (extension `.sty` and `\ProvidesPackage` macro).

Notice: it is easier to merge 1.8 first (cf #5892) then to merge master into this PR to fix conflicts which will arise (in the two Japanese tests modified here).

### DIFF
--- a/tests/test_build_latex.py
+++ b/tests/test_build_latex.py
@@ -357,9 +357,9 @@ def test_numref_with_language_ja(app, status, warning):
     # sphinxmessages.sty
     result = (app.outdir / 'sphinxmessages.sty').text(encoding='utf8')
     print(result)
-    assert '\n{\\renewcommand{\\figurename}{図 }}' in result
-    assert '\n{\\renewcommand{\\tablename}{表 }}' in result
-    assert '\n{\\renewcommand{\\literalblockname}{リスト}}' in result
+    assert '\\@iden{\\renewcommand{\\figurename}{図 }}' in result
+    assert '\\@iden{\\renewcommand{\\tablename}{表 }}' in result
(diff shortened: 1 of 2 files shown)


tests/test_build_latex.py
```diff
@@ -357,9 +357,9 @@ def test_numref_with_language_ja(app, status, warning):
     # sphinxmessages.sty
     result = (app.outdir / 'sphinxmessages.sty').text(encoding='utf8')
     print(result)
-    assert '\n{\\renewcommand{\\figurename}{図}}' in result
-    assert '\n{\\renewcommand{\\tablename}{表}}' in result
-    assert '\n{\\renewcommand{\\literalblockname}{リスト}}' in result
+    assert '\n\\@iden{\\renewcommand{\\figurename}{図}}' in result
+    assert '\n\\@iden{\\renewcommand{\\tablename}{表}}' in result
+    assert '\n\\@iden{\\renewcommand{\\literalblockname}{リスト}}' in result
```
tk0miya: I used `\n` to test nothing exists before the braces. So it is no longer needed with this change.

tk0miya (commented):
LGTM with nits!

jfbu:
Thanks for reviewing and of #5892 as well. Will merge after testing completes.

PR #5895, opened 2019-01-05
Fix warning message when a node class is already registered

mitya57:
`self._setting_up_extension` is a list, we should only take its last item. This is what `add_directive()` and `add_role()` already do.

Otherwise the message looks like this:

    while setting up extension ['?', 'myextension']: ...


tk0miya (approved):
+1

tk0miya:
Merged. Thank you always, best wishes for 2019 :-)

mitya57:
Best wishes to you too!

PR #5898, opened 2019-01-06
Remove redundant mypy config 'incremental'

jdufresne:
The mypy config 'incremental` has defaulted as true since version 0.590. Can drop the local override. For details, see:

https://github.com/python/mypy/commit/6b13652a466ccb102987a2ab1fa93d4b52fd8e3f


tk0miya (approved):
LGTM!

PR #5899, opened 2019-01-06
Remove outdated fallback to 'import Image'

jdufresne:
The fallback 'import Image' was added in commit 99621d7a01638872a5af727aaa8714b269c0f769 for the PIL package. However, the PIL package is no longer maintained and does not support Python 3.
Development has moved to the Pillow package which always installs to the PIL namespace.

Pillow repo: https://github.com/python-pillow/Pillow
Old PIL website: http://www.pythonware.com/products/pil/

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1713,9 +1713,9 @@ the `Dublin Core metadata <http://dublincore.org/>`_.
 
    This flag determines if sphinx should try to fix image formats that are not
    supported by some epub readers.  At the moment palette images with a small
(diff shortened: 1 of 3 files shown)


tk0miya:
In addition to this, there are some comments and messages mentions about PIL.
We need to update them at same moment. Please check sphinx/builders/_epub_base.py.
Thanks,

jdufresne:
Thanks! I've updated all comments referencing PIL. Please let me know if you'd like to see any other edits.

tk0miya (approved):
LGTM!

PR #5900, opened 2019-01-06
Add TODO comment to help future developer to refactor guess_mimetype

jdufresne:
All internal uses always pass a filename to `guess_mimetype()` so the default of the empty string has no practical use.

### DIFF
--- a/sphinx/util/images.py
+++ b/sphinx/util/images.py
@@ -81,6 +81,8 @@ def guess_mimetype(filename='', content=None, default=None):
     if ext in mime_suffixes:
         return mime_suffixes[ext]
     elif content:
+        # TODO: When the content argument is removed, make filename a required
+        # argument.
         warnings.warn('The content argument of guess_mimetype() is deprecated.',
                       RemovedInSphinx30Warning, stacklevel=2)
         return guess_mimetype_for_stream(BytesIO(content), default=default)


sphinx/util/images.py
```diff
@@ -75,8 +75,13 @@ def guess_mimetype_for_stream(stream, default=None):
         return default
 
 
-def guess_mimetype(filename='', content=None, default=None):
+def guess_mimetype(filename=None, content=None, default=None):
     # type: (str, bytes, str) -> str
+    if filename is None:
+        warnings.warn('The first positional argument of guess_mimetype() will '
+                      'be required in Sphinx 4.0.',
+                      RemovedInSphinx40Warning, stacklevel=2)
```
tk0miya: I guess this warning is mainly shown for developers who calls this with `content` parameter. And we've already emitted a warning for them below. So I wonder is this really helpful.

jdufresne: Makes sense. Do you think no additional steps are required to make the first positional argument required once `content` is removed then? 

Do you think it would be useful to add a helper comment to remind the future developer to make the argument required?

tk0miya: I think no additional steps is needed. We'll modify it without any declaration. It only affect to `guess_mimetype(default='foo')` :-p

How about adding TODO comment to the warning for `content`? It will help us on removal.

jdufresne: Done in the latest revision. Thanks!

tk0miya (approved):
Thank you for your work!

PR #5902, opened 2019-01-07
Stop to use $SPHINX_TEST_TEMPDIR envvar

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
At once, we added $SPHINX_TEST_TEMPDIR to keep intermediate files of
testing to investigate the failures.  At present, we've used pytest
as a test runner.  And it keeps temporary directories of last 3
testings.  So this starts to use it instead of $SPHINX_TEST_TEMPDIR.
https://docs.pytest.org/en/latest/tmpdir.html#the-default-base-temporary-directory

### DIFF
--- a/sphinx/testing/fixtures.py
+++ b/sphinx/testing/fixtures.py
@@ -14,7 +14,6 @@ import sys
 from collections import namedtuple
 from io import StringIO
 from subprocess import PIPE
-from tempfile import gettempdir
 
 import pytest
 
@@ -221,11 +220,15 @@ def if_graphviz_found(app):
 
 
 @pytest.fixture(scope='session')
-def sphinx_test_tempdir():
+def sphinx_test_tempdir(tmpdir_factory):
     """
     temporary directory that wrapped with `path` class.
     """
-    return util.path(os.environ.get('SPHINX_TEST_TEMPDIR', gettempdir())).abspath()
+    tmpdir = os.environ.get('SPHINX_TEST_TEMPDIR')  # RemovedInSphinx40Warning
+    if tmpdir is None:
+        tmpdir = tmpdir_factory.getbasetemp()
+
+    return util.path(tmpdir).abspath()
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


sphinx/testing/fixtures.py
```diff
@@ -223,11 +222,15 @@ def if_graphviz_found(app):
 
 
 @pytest.fixture(scope='session')
-def sphinx_test_tempdir():
+def sphinx_test_tempdir(tmpdir_factory):
     """
     temporary directory that wrapped with `path` class.
     """
-    return util.path(os.environ.get('SPHINX_TEST_TEMPDIR', gettempdir())).abspath()
+    tmpdir = os.environ.get('SPHINX_TEST_TEMPDIR')
```
shimizukawa: I think a deprecation declaration (or comment) is needed if this `SPHINX_TEST_TEMPDIR` is kept for backward compatibility,

shimizukawa (approved):
LGTM with nits ;)

please rebase this PR because #5941 is merged into master.

jfbu:
@tk0miya In the make way of running tests, I don't get any console output of where the temporary build dir is located 

```
============================= test session starts ==============================
platform darwin -- Python 3.6.8, pytest-4.1.1, py-1.7.0, pluggy-0.8.1 -- /xxx/yyy/www/sphinxvenv/bin/python3
cachedir: .pytest_cache
libraries: Sphinx-2.0.0+/7a4741acc, docutils-0.14
rootdir: /xxx/zzz/sphinx, inifile: setup.cfg
collected 64 items                                                             

tests/test_build_latex.py::test_build_latex_doc[pdflatex-howto] SKIPPED  [  1%]
...

==================== 57 passed, 7 skipped in 13.45 seconds =====================

```

This makes it inconvenient. Using `tox` is orders of magnitude slower than having a virtualenv set-up. I don't use `tox` for that reason. I tested it and was not happy with it.

jfbu:
Of course I can export SPHINX_TEST_TEMPDIR environment variable for now, and also I have now located where pytest stores the last 3 build repertories, I will only need to bookmark that. Sorry for noise. I wonder though if there could be a way to print to console the tmpdir value, even when running test via the make test interface.

PR #5915, opened 2019-01-09
Psychopy

jona-sassenhagen:
Adding the chapter for programming experiments


jona-sassenhagen:
Ugh that's an ugly commit history ...

jona-sassenhagen:
@aylinsgl do you know how we can check how it renders? Is it in the Circle builds?

jfbu:
It seems you are targeting the wrong project, this is the Sphinx project itself here `:)`

jona-sassenhagen:
Argh sorry :D

jona-sassenhagen:
@jfbu would it be possible to delete this PR? I've been inaccurately classified as having contributed to sphinx development because of it.

jfbu:
@jona-sassenhagen I think only an Owner can do that, if at all possible. @tk0miya what do you think?

tk0miya:
I don't know the way to delete this. This post says you need to mail to github staff:
https://stackoverflow.com/questions/18318097/delete-a-closed-pull-request-from-github

jona-sassenhagen:
Hm, it does not fall in that category (TOS violations/sensitive data). I guess then it will forever look like I contributed 20k lines of js and css on my GitHub.

Thanks for the help! (Sphinx is a great project!)

PR #5918, opened 2019-01-10
LaTeX: no need for \sphinxVerbatimEnvironment (refactoring)

jfbu:
It is not needed as fancyvrb's `\VerbatimEnvironment` can be used in
nested way and the outer one prevails (sorry for adding the complication
initially when providing sphinxVerbatimintable environment).

That internal macro is not public API, and it is hard to deprecate it
from LaTeX as the Sphinx custom Verbatim environments redefined it
on-the-fly.

I searched and found no usage on internet of this, so there does not
seem any need to deprecate this non-public internal macro before
removal.


tk0miya (approved):
+1

jfbu:
thanks for review, merging

PR #5924, opened 2019-01-11
githubpages: support custom domains

arxanas:
Subject: Add support for custom domains in the `githubpages` extension.

### Feature or Bugfix

- Feature

### Purpose

Github Pages allows you to link a custom domain to your Github Pages site by adding a `CNAME` file at the top-level of your `docs` directory. The `githubpages` extension already inserts a `.nojekyll` file in the `docs` directory, so it's a good place to add support for this `CNAME` file as well.

This is mostly a copy of my commit for my own repo at https://github.com/pytch-lang/pytch/commit/75a4b04433eea9010ba060d0108246b88160ae51 (see https://pytch-lang.org for an example of it working in production).

Let me know what needs to be done on the documentation front.


sphinx/ext/githubpages.py
```diff
@@ -23,11 +23,20 @@
 def create_nojekyll(app, env):
     # type: (Sphinx, BuildEnvironment) -> None
     if app.builder.format == 'html':
-        path = os.path.join(app.builder.outdir, '.nojekyll')
-        open(path, 'wt').close()
```
tk0miya, line 26: I feel no reason to rewrite the name of variable.

sphinx/ext/githubpages.py
```diff
@@ -23,11 +23,20 @@
...
+        nojekyll_path = os.path.join(app.builder.outdir, '.nojekyll')
+        open(nojekyll_path, 'wt').close()
+
+        cname = app.config['githubpages_cname']
```
tk0miya: Please access the config value like an attribute: `app.config.githubpages_cname`.

sphinx/ext/githubpages.py
```diff
@@ -23,11 +23,20 @@
...
+        if cname:
+            cname_path = os.path.join(app.builder.outdir, 'CNAME')
+            with open(cname_path, 'wt') as cname_f:
```
tk0miya: How about using `f`?

tk0miya (changes requested):
Putting `CNAME` file is good idea! I like it :-)

tk0miya:
In addition, this is a new feature. So please move this to master branch.

arxanas:
@tk0miya: thanks for the prompt review!

Changes:

* Moved to master branch.
* Removed intermediate path variables.
* Accessed config value using `.` syntax.
* Renamed `cname_f` to `f`.

sphinx/ext/githubpages.py
```diff
@@ -22,11 +22,18 @@
...
+            with open(os.path.join(app.builder.outdir, 'CNAME'), 'wt') as f:
+                # NOTE: don't write a trailing newline. The `CNAME` file that's
+                # auto-generated by the Github UI doesn't have one.
+                f.write(cname)
 
 
 def setup(app):
     # type: (Sphinx) -> Dict[str, Any]
     app.connect('env-updated', create_nojekyll)
+    app.add_config_value("githubpages_cname", "", "html")
```
tk0miya: Thank you for update. But, I'd overlooked we already have a configuration `html_baseurl` which indicates the URL of document. It would be better to use it instead of `githubpages_cname` to me. What do you think?

arxanas: It seems like `html_baseurl` expects a string like `https://sphinx-doc.org`, but the CNAME wants a string like `sphinx-doc.org`. How can we convert the first form into the second form across Python versions? Ordinarily I would use `urllib`, but it looks like the API has changed significantly between Python 2 and 3; and doing some string manipulation like splitting on `/` seems suspect to me.

arxanas: Actually, I think we can't do this. We may want to set `html_baseurl` to something like `https://sphinx-doc.github.io/sphinx-doc/sphinx/`. In that case, we don't actually own the domain `github.io`, and we don't want to generate the `CNAME` file. It's also possible there's some use of `html_baseurl` that I haven't thought of yet, but it would be hard to change the behavior of this extension in the future. Therefore I think it's best to opt in explicitly using a setting specifically for the extension.

tk0miya: Sorry for late response.

>How can we convert the first form into the second form across Python versions?

Since 2.0, we'll only support py3.5+. So you can use `urllib.parse` for such purpose.

>In that case, we don't actually own the domain `github.io`, and we don't want to generate the `CNAME` file.

We can test it with `hostname.endswith('.github.io')` simply.

arxanas:
Changes:

* Renamed function to `create_nojekyll_and_cname`.
* Used `urllib.parse` to parse the hostname component of `html_baseurl`. Don't generate the `CNAME` file if it appears to be a `github.io` domain.
* Updated tests.

tk0miya (approved):

tk0miya:
Thank you for updating. LGTM!

PR #5930, opened 2019-01-12
Use subprocess.run() instead of Popen()

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
Since python3.5, subprocess.run() has been introduced. It works a
wrapper of Popen, and it looks much simple and better. This uses it
instead of Popen to make our code simple.

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -11,9 +11,10 @@
 
 import posixpath
 import re
+import subprocess
 from hashlib import sha1
 from os import path
-from subprocess import Popen, PIPE
+from subprocess import CalledProcessError, PIPE
 
 from docutils import nodes
 from docutils.parsers.rst import directives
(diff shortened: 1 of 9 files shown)


sphinx/ext/graphviz.py
```diff
@@ -243,31 +244,24 @@ def render_dot(self, code, options, format, prefix='graphviz'):
...
         dot_args.extend(['-Tcmapx', '-o%s.map' % outfn])
+
     try:
-        p = Popen(dot_args, stdout=PIPE, stdin=PIPE, stderr=PIPE, cwd=cwd)
-    except FileNotFoundError:
+        ret = subprocess.run(dot_args, input=code.encode(), stdout=PIPE, stderr=PIPE,
+                             cwd=cwd, check=True)
+        if not path.isfile(outfn):
+            raise GraphvizError(__('dot did not produce an output file:\n[stderr]\n%s\n'
+                                   '[stdout]\n%s') % (ret.stderr, ret.stdout))
```
tk0miya: `ret.stderr` and `ret.stdout` is not decoded. Using `%r` is better for this case.

Note: It is hard to determine the correct encodings. So it is better to show them as bytes.

tk0miya:
There are some use of Popen in our test code. They also should be fixed.

tests/test_build_latex.py
```diff
@@ -43,43 +44,32 @@
...
         return True
+    except (OSError, CalledProcessError):
+        return False  # command not found or exit with non-zero
 
 
 # compile latex document with app.config.latex_engine
 def compile_latex_document(app, filename='python.tex'):
     # now, try to run latex over it
-    with cd(app.outdir):
         try:
```
shimizukawa: MUST: please dedent this block

shimizukawa (approved):
LGTM! Excellent!! 

with a nit ;)

PR #5934, opened 2019-01-12
Add 'cpp:struct' directive

Symaxion:
Subject: Add `cpp:struct` directive

### Feature or Bugfix

- [x] Feature
- [ ] Bugfix

### Purpose
This pull requests adds a new directive to the C++ domain, `cpp:struct`, to complement the existing `cpp:class` directive. The two main differences are that objects declared with `cpp:struct` are labeled as *struct* instead of *class* (both in the generated documentation itself as in the index), and that structs have public access by default. Other than that (as in C++ itself), there is no difference between classes and structs.

Having a separate directive is useful for code bases that use the `class` and `struct` keywords for different purposes (e.g., structs for data containers without logic) and want to make that distinction clear in documentation.

### Detail
Code is largely based on that of `cpp:class`, with a few tweaks. I've tried to minimise code duplication as much as possible. 

I updated the code for the C++ domain, as well as the unit tests and documentation. Let me know if there's anything I've missed.

### Relates
- N/A

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -153,6 +153,7 @@ Features added
 * #5841: apidoc: Add --extensions option to sphinx-apidoc
 * #4981: C++, added an alias directive for inserting lists of declarations,
   that references existing declarations (e.g., for making a synopsis).
+* C++: Add ``cpp:struct`` to complement ``cpp:class``
 
 Bugs fixed
 ----------
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -549,7 +549,7 @@ visibility statement (``public``, ``private`` or ``protected``).
 
 .. rst:directive:: .. cpp:class:: class specifier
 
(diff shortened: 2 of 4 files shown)


jakobandersen:
Thanks a lot @Symaxion! Some of the internal naming around "object types" is quite misleading, as there is both the type of directive used for a declaration and then what the basic C++ declaration type it is. Based on you work I have created an alternative PR (#5973) which is less intrusive, i.e., not adding a new declaration type to be handled in the core.
The PR also removes the old behaviour of "simplifying" the visibility of base classes. The output now mirrors what the user wrote.
Is the new PR ok for your use cases?

Symaxion:
Hi @jakobandersen, thanks for the quick response! I tried your PR on a couple of test documents which I used to test my own patch, and everything looks OK to me. Thanks for merging in this feature!

This ticket can be closed for me.

jakobandersen:
Great, I'll close this PR and merge the other.

PR #5905, opened 2019-01-07
Update deprecation policy by semantic versioning rules

tk0miya:
### Feature or Bugfix
- docs

### Purpose
- We'll adopt semantic versioning since 2.0 release.
- This updates our deprecation policy for post 2.0 release.

PR #5938, opened 2019-01-12
Update deprecation policy by semantic versioning rules

shimizukawa:
This is a alternative proposal of #5905.

### Feature or Bugfix
- docs

### Purpose
- We'll adopt semantic versioning since 2.0 release.
- This updates our deprecation policy for post 2.0 release.

--------

#5905 following operational rules.

* Perform stable release in PATCH version
* Perform master release in MINOR version

Based on the discussion with @tk0miya, the following operational rules seemed to be good so I propose.

* PATCH release: urgent bug fixes (currently shipped with stable release)
* MINOR release: normal bug fixes (currently shipped with stable release) and addition of new features that preserving backwards-compatibility (currently shipped with master release)
* MAJOR release: incompatible changes (currently shipped with master release)

So, I updated the document on this assumption.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -91,16 +91,19 @@ These are the basic steps needed to start developing on Sphinx.
 
 #. Checkout the appropriate branch.
 
-   Sphinx uses Semantic Versioning 2.0.0 (refs: https://semver.org/ ).
+   Sphinx adopts Semantic Versioning 2.0.0 (refs: https://semver.org/ ).
 
    For changes that preserves backwards-compatibility of API and features,
-   it should be included in the next MINOR release, use the ``X.Y`` branch. ::
+   they should be included in the next MINOR release, use the ``X.Y`` branch.
+   ::
 
        git checkout X.Y
 
    For incompatible or other substantial changes that should wait until the
-   next MAJOR release, use the ``master`` branch  (see `Branch Model`_ for
-   detail).
+   next MAJOR release, use the ``master`` branch.
+
+   For urgent release, a new PATCH branch must be branched from the newest
+   release tag (see `Branch Model`_ for detail).
 
 #. Setup a virtual environment.
 
@@ -319,7 +322,8 @@ Versioning 2.0.0 (refs: https://semver.org/ ).
 
 ``master``
     Development for MAJOR version.
-    All incompatible behavior and public API changes are allowed.
+    All changes including incompatible behaviors and public API updates are
+    allowed.
 
 ``X.Y``
     Where ``X.Y`` is the ``MAJOR.MINOR`` release.  Used to maintain current
@@ -336,7 +340,9 @@ Versioning 2.0.0 (refs: https://semver.org/ ).
     version is used for urgent bug fix.
 
     ``MAJOR.MINOR.PATCH`` branch will be branched from the ``v`` prefixed
-    release tag (ex. v2.3.0) when a urgent release is needed.
+    release tag (ex. make 2.3.1 that branched from v2.3.0) when a urgent
+    release is needed. When new PATCH version is released, the branch will be
+    deleted and replaced by an equivalent tag (ex. v2.3.1).
 
 
 Deprecating a feature
@@ -369,8 +375,8 @@ silence any warnings generated when running the tests.
 Deprecation policy
 ------------------
 
-A MAJOR and MINOR releases may deprecate certain features from previous
-releases. If a feature is deprecated in MINOR release A.x, it will continue to
+MAJOR and MINOR releases may deprecate certain features from previous
+releases. If a feature is deprecated in a release A.x, it will continue to
 work in all A.x.x versions (for all versions of x). It will continue to work
 in all B.x.x versions but raise deprecation warnings. Deprecated features
 will be removed at the C.0.0. It means the deprecated feature will work during
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -61,8 +61,8 @@ of the core developers before it is merged into the main repository.
 #. If you feel uncomfortable or uncertain about an issue or your changes, feel
    free to email the *sphinx-dev* mailing list.
 #. Fork `the repository`_ on GitHub to start making your changes to the
-   ``master`` branch for next major version, or ``X.Y`` branch for next
-   minor version (see `Branch Model`_).
+   ``master`` branch for next MAJOR version, or ``X.Y`` branch for next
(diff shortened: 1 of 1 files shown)


shimizukawa:
Memo:

As a premise of discussion, semver has the following rules as a general rule.

* PATCH: Application of bug fixes with backward compatibility
* MINOR: Added functions that maintain backward compatibility. Bug fixes may be included
* MAJOR: backward incompatible change

## About the PATCH version

Observing several OSS projects, it seems that only releasing the PATCH version for a few hours ~ few days just after raising the MAJOR or MINOR.

Even with the release of Sphinx so far, there was a re-release immediately after release [1.7.9](https://www.sphinx-doc.org/en/master/changes.html#release-1-7-9-released-sep-05-2018).
I think that we will raise the PATCH version in such a case on semver.

Therefore, the PATCH version should be released as soon as there is a urgent problem, and it seems that we should not make a PATCH release unless there is a urgent problem.

## About the MINOR version

So far, the stable version (z part of 1.8.z) does not introduce new features. In the semver, MINOR can include new features that preserve backwards-compatibility. For this reason, if we release MINOR every other month, we may be able to release new features at that frequency.

## About the MAJOR version

So far, we introduced new features in the master version (y part of 1.y.0) and made incompatible changes. In the semver, we will make incompatible changes and drop deprecations in the MAJOR release.

jfbu (approved):
Thanks for detailed context. I think this is indeed better alternative than #5905. +1

CONTRIBUTING.rst
```diff
@@ -91,13 +91,15 @@ These are the basic steps needed to start developing on Sphinx.
 
 #. Checkout the appropriate branch.
 
-   For changes that should be included in the next minor release (namely bug
-   fixes), use the ``X.Y`` branch. ::
+   Sphinx uses Semantic Versioning 2.0.0 (refs: https://semver.org/ ).
```
tk0miya: adopts or follows?

CONTRIBUTING.rst
```diff
@@ -91,13 +91,15 @@ These are the basic steps needed to start developing on Sphinx.
...
+
+   For changes that preserves backwards-compatibility of API and features,
+   it should be included in the next MINOR release, use the ``X.Y`` branch. ::
```
tk0miya: *they* should be...?

CONTRIBUTING.rst
```diff
@@ -91,13 +91,15 @@ These are the basic steps needed to start developing on Sphinx.
 
 #. Checkout the appropriate branch.
 
-   For changes that should be included in the next minor release (namely bug
-   fixes), use the ``X.Y`` branch. ::
+   Sphinx uses Semantic Versioning 2.0.0 (refs: https://semver.org/ ).
+
```
tk0miya, line 95: It seems there are no description about urgent or critical bugs. Is it unnecessary?

CONTRIBUTING.rst
```diff
@@ -312,18 +314,29 @@ Debugging Tips
...
 
-Sphinx project uses following branches for developing.
+Sphinx project uses following branches for developing that conforms to Semantic
+Versioning 2.0.0 (refs: https://semver.org/ ).
 
 ``master``
-    Used for main development.  All improvement and refactoring, bug fixes
-    are allowed.
+    Development for MAJOR version.
+    All incompatible behavior and public API changes are allowed.
```
tk0miya: How about "All changes including incompatible behaviors and public API updates are allowed."?
I think we can also add bug fixes and any minor changes on master branch.

CONTRIBUTING.rst
```diff
@@ -312,18 +314,29 @@ Debugging Tips
...
+    new MAJOR version is released, the old ``MAJOR.MINOR`` branch will be
+    deleted and replaced by an equivalent tag.
+
+``X.Y.Z``
+    Where ``X.Y.Z`` is the ``MAJOR.MINOR.PATCH`` release.  Only
+    backwards-compatible bug fixes are allowed. In Sphinx project, PATCH
+    version is used for urgent bug fix.
+
+    ``MAJOR.MINOR.PATCH`` branch will be branched from the ``v`` prefixed
+    release tag (ex. v2.3.0) when a urgent release is needed.
```
tk0miya: We should mention about replacing the branch by `vX.Y.Z` tag after the release. I believe this branch lives only short period (a few days).

CONTRIBUTING.rst
```diff
@@ -356,23 +369,22 @@ silence any warnings generated when running the tests.
 Deprecation policy
 ------------------
 
-A feature release may deprecate certain features from previous releases. If a
-feature is deprecated in feature release 1.A, it will continue to work in all
-1.A.x versions (for all versions of x) but raise warnings. Deprecated features
-will be removed in the first 1.B release, or 1.B.1 for features deprecated in
-the last 1.A.x feature release to ensure deprecations are done over at least 2
-feature releases.
+A MAJOR and MINOR releases may deprecate certain features from previous
```
tk0miya: An article "A" is not needed here.

CONTRIBUTING.rst
```diff
@@ -356,23 +369,22 @@ silence any warnings generated when running the tests.
...
+releases. If a feature is deprecated in MINOR release A.x, it will continue to
```
tk0miya: This sentence mentions only about MINOR release. But there are no mentions about MAJOR release. How about using "in a release" instead?

tk0miya (changes requested):
LGTM with nits

CONTRIBUTING.rst
shimizukawa: `adopts` +1

CONTRIBUTING.rst
shimizukawa: `they` +1

CONTRIBUTING.rst
shimizukawa: Added
https://github.com/sphinx-doc/sphinx/pull/5938/files#diff-ea1f65a8eaac0cec7ae151dfdbbbc587R105

CONTRIBUTING.rst
shimizukawa: done: https://github.com/sphinx-doc/sphinx/pull/5938/files#diff-ea1f65a8eaac0cec7ae151dfdbbbc587R344

shimizukawa:
@tk0miya updated. re-check plz

tk0miya (approved):
Thank you for update!

shimizukawa:
Thanks for your reviewing!

tk0miya:
🎉

Issue #5936, opened 2019-01-12
PDF build broken by inclusion of image taller than page height in an admonition


jfbu:
### Problem

PDF build is broken if inside an admonition (warning, caution, attention, danger, error) there is an included image file which a requested height close to or larger than the total text area in a page.

#### Procedure to reproduce the problem

```
Test
====

.. attention::

   .. image:: rimg.*
      :height: 1000px
```

With this image `rimg.png`: 

![rimg](https://user-images.githubusercontent.com/2589111/51074062-6b16fa80-167a-11e9-99ab-f5a3371ed3ba.png)

(or any image at least about twice taller than wide, so that the Sphinx cap on maximal width to `\linewidth` does not modify requested height, bigger than text height)

#### Error logs / results

PDF build crashes after spitting out close to 5000 empty pages.

```
! TeX capacity exceeded, sorry [input stack size=5000].
\@ifundefined #1->\ifcsname #1
                              \endcsname \@ifundefin@d@i \else \@ifundefin@d...
l.89 \end{sphinxadmonition}
                           
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on FOO.log.
=== TeX engine is 'pdfTeX'
Latexmk: Index file 'FOO.idx' was written
Latexmk: Errors, so I did not complete making targets
```

#### Expected results

The image is either resized or split to not create errors.

### Comment

The problem comes from `framed.sty` : an image can not be split in LaTeX (one can lie to LaTeX about its actual size however, but then it will collide with following paragraph) and an over-sized image is first moved to next page, then `framed.sty` realizes there is not enough room, the image is again moved to next page, etc... creating infinite loop until some TeX memory is exhausted.

It is not possible to split the image from inside LaTeX, but it is possible to include it multiple times and clip each inclusion to successive horizontal slices. In recent pdftex/xetex this will not increase PDF filesize. However, this is definitely complicated method. See [this tex.sx question](https://tex.stackexchange.com/questions/469741/broken-build-if-a-too-big-picture-is-included-in-a-framed-environment-how-to-in) I posted on a TeX/LaTeX forum.

More reasonable is to modify `\sphinxincludegraphics` which is the Sphinx provided wrapper of LaTeX `\includegraphics` to force a maximal image height compatible with standing alone on one page (with the framing). Already, `\sphinxincludegraphics` forces the width to not exceed `\linewidth`, so this keeps the same spirit.

### Environment info
- OS: Mac
[...]

PR #5939, opened 2019-01-12
LaTeX: \sphinxsafeincludegraphics for sphinxShadowBox/sphinxheavybox

jfbu:
Fixes: #5936

Relates: https://github.com/spatialaudio/nbsphinx/pull/277 (the nbsphinx PR is independent and can still be used at same time as this PR).

I tested this with the `rimg.png` file of #5936 and

```rest
.. attention::

   .. image:: rimg.*
      :height: 1000px

.. topic:: test

   .. image:: rimg.*
      :height: 1000px

.. attention::

   NESTED !

   .. attention::

      NESTED DEEPER !

      .. attention::

         .. image:: rimg.*
            :height: 1000px
```

I also tested with a4paper/letterpaper, 10pt/12pt, and this in conf.py `latex_elements`,

```
'sphinxsetup': 'shadowsep=12pt, shadowsize=24pt, shadowrule=3pt, attentionborder=10pt',
```

When an image is too big it is resized to a little less than the ~~pageheight~~ text area height, the image is then placed on next page, with the framing.

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -6,7 +6,7 @@
 %
 
 \NeedsTeXFormat{LaTeX2e}[1995/12/01]
-\ProvidesPackage{sphinx}[2018/12/16 v1.8.3 LaTeX package (Sphinx markup)]
+\ProvidesPackage{sphinx}[2019/01/12 v1.8.4 LaTeX package (Sphinx markup)]
 
 % provides \ltx@ifundefined
 % (many packages load ltxcmds: graphicx does for pdftex and lualatex but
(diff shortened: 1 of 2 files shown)


sphinx/texinputs/sphinx.sty
```diff
@@ -716,10 +716,38 @@
       \setbox\spx@image@box\box\voidb@x % clear memory
       \includegraphics[#1,width=\linewidth]{#2}%
     \else
+      \setbox\spx@image@box\box\voidb@x % clear memory
```
jfbu, line 719: I added this because it was obviously lacking from that branch; if we were handling  a regular LaTeX package, that modification would be breaking change because some other LaTeX developers could have used the after-effect that `\spx@image@box` was not cleared.

I don't think this will break anyone's Sphinx document though because it is highly improbable the after-effect of a non-emptied scratch box register in a certain branch was used by some LaTeX developer out there... but in all rigor this "fix" could  await 2.0. 

It is indeed completely unrelated to the other changes in this PR.

sphinx/texinputs/sphinx.sty
```diff
@@ -1339,6 +1367,12 @@
 % could nest if LaTeX writer authorized it
 \newenvironment{sphinxShadowBox}
   {\def\FrameCommand {\spx@ShadowFBox }%
+   \advance\spx@image@maxheight
+   -\dimexpr2\sphinxshadowrule
+           +2\sphinxshadowsep
+           +\sphinxshadowsize
+           +\baselineskip\relax
```
jfbu, line 1374: This recipe is a bit experimental, testing showed it worked. I don't know what is the exact dimension to subtract (I determined it to 4 fractional digits precision, but for example it differs if letterpaper or a4paper for some mysterious reasons; the `framed.sty` dimensions are very complicated).

sphinx/texinputs/sphinx.sty
```diff
@@ -716,10 +716,38 @@
...
   \fi
 }
+% \sphinxsafeincludegraphics resizes images larger than the line width,
+% or taller than about the text height (whether or not height/width options
+% were used). This is requested to avoid a crash with \MakeFramed as used by
+% sphinxShadowBox (topic/contents) and sphinxheavybox (admonitions), and also
+% by sphinxVerbatim (but a priori no image inclusion there).
+\newdimen\spx@image@maxheight
+% default maximal setting will get reduced by sphinxShadowBox/sphinxheavybox
+\AtBeginDocument{\spx@image@maxheight\textheight}
```
jfbu, line 731: In this PR only sphinxShadowBox and sphinxheavybox will update value of `\spx@image@maxheight`. Also sphinxVerbatim uses framing: but image inclusion is a priori not possible there, so I did not add the corresponding code to it to update `\spx@image@maxheight`.

jfbu: Actually, an image could possibly be part of the caption of a code-block, and the caption is rendered using framing code, but I figured it was not needed to enter into unneeded complications as inclusion of a big image in such a caption title is really improbable.

tk0miya (approved):
LGTM!

jfbu:
Thanks for reviewing, merging now

PR #5951, opened 2019-01-14
Drop feature: websupport

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- It's time to drop the websupport from core.
- Now it works fine as a extension; sphinxcontrib-websupport


sphinx/themes/basic/static/websupport.js
```diff
@@ -1,808 +0,0 @@
-/*
```
tk0miya, line 1: I found we still have an asset file in core repository.
We have to move this to sphinxcontrib-websupport package ASAP.

tk0miya: done

Issue #5954, opened 2019-01-14
LaTeX: ``:scale:`` image option may break PDF build if image in an admonition


jfbu:
From

```
Test
====

.. attention::

   .. image:: rimg.*
      :height: 500px
      :scale: 200%
```

and graphics file `rimg.png` as here:
![rimg](https://user-images.githubusercontent.com/2589111/51116849-87e33780-180c-11e9-87d8-a80a59475bea.png)
or any image we get again recently fixed issue #5936.

The cause is simply at https://github.com/sphinx-doc/sphinx/blob/0081482eb7b65cb286afd6ed1c466990545226be/sphinx/writers/latex.py#L1534-L1539

as the `\scalebox` will create a box too tall to fit in the framing causing the formerly reported bug.

The fix must consist in never using `\scalebox` but rather modify height or width option as passed to LaTeX to already incorporate the scale setting, following the [Docutils specs](http://docutils.sourceforge.net/docs/ref/rst/directives.html#image)

> height : length
>     The desired height of the image. Used to reserve space or scale the image vertically. When the "scale" option is also specified, they are combined. For example, a height of 200px and a scale of 50 is equivalent to a height of 100px with no scale.

PR #5955, opened 2019-01-14
LaTeX: remove usage of \scalebox by incorporating scale in height/width

jfbu:
Closes: #5954

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -424,21 +424,32 @@ def escape_abbr(text):
     return re.sub(r'\.(?=\s|$)', r'.\@', text)
 
 
-def rstdim_to_latexdim(width_str):
-    # type: (unicode) -> unicode
+def rstdim_to_latexdim(width_str, scale = 100):
+    # type: (unicode, int) -> unicode
     """Convert `width_str` with rst length to LaTeX length."""
     match = re.match(r'^(\d*\.?\d*)\s*(\S*)$', width_str)
     if not match:
         raise ValueError
     res = width_str
     amount, unit = match.groups()[:2]
-    float(amount)  # validate amount is float
-    if unit in ('', "px"):
-        res = "%s\\sphinxpxdimen" % amount
-    elif unit == 'pt':
-        res = '%sbp' % amount  # convert to 'bp'
-    elif unit == "%":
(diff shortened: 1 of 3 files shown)


jfbu:
Testing [failed](https://travis-ci.org/sphinx-doc/sphinx/jobs/479456435) for reasons non related to this PR:

```
E       AssertionError: assert not 'links.txt:3: [broken] https://www.w3.org/TR/2006/REC-xml-names-20060816/#defaulting: argument must be a string without NUL characters\n'
```

in execution of tests/test_build_linkcheck.py::test_anchors_ignored. Possibly transient error?

sphinx/writers/latex.py
```diff
@@ -424,21 +424,32 @@ def escape_abbr(text):
...
-    elif unit == "%":
-        res = "%.3f\\linewidth" % (float(amount) / 100.0)
+    if scale == 100:
+        float(amount)  # validate amount is float
+        if unit in ('', "px"):
+            res = "%s\\sphinxpxdimen" % amount
+        elif unit == 'pt':
+            res = '%sbp' % amount  # convert to 'bp'
+        elif unit == "%":
+            res = "%.3f\\linewidth" % (float(amount) / 100.0)
```
jfbu, line 442: In future `%.3f` when merging to 2.0 should be replaced by `%.5f` because TeX has about 5 decimal digits precision in such computations.

For time being I am keeping exact same code as we are on stable branch.

sphinx/writers/latex.py
```diff
@@ -424,21 +424,32 @@ def escape_abbr(text):
...
+    else:
+        amount_float = float(amount) * scale / 100.0
```
jfbu, line 444: mypy complained when I used amount = float(amount) etc..., and then in all next, because it is not able to recognize the new type assigned to variable; in my mind this is deficiency of mypy, although I am aware using same identifier with two distinct types might not be best taste.

tk0miya (approved):
LGTM!

Note: `tests/test_build_linkcheck.py` has failed often. So we have to update the case if possible. At this time, I manually restarted the job on Travis CI.

jfbu:
Thanks for reviewing and relaunching tests! Merging.

PR #5961, opened 2019-01-17
Don’t collapse python fields

flying-sheep:
PyTypedField and PyGroupedField are both GroupedFields with the default `list_type = nodes.bullet_list`

This means that their list items can’t stand alone, neither in a doctree nor in HTML, LaTeX, …


flying-sheep:
Whoops, my bad: it doesn’t actually add list items if `can_collapse=True`

PR #5964, opened 2019-01-17
Closes: #5960

### LINKED ISSUE #5960: Different latexpdf output on Linux and Windows
**Describe the bug**
When running pdflatex on a .tex file that includes the sphinx package, the vertical spacing after section headers is different between Linux and Windows.
This breaks the whole layout of my document, because two lines of a new section regularly spill to the previous page.

**To Reproduce**
Use the latex builder to create a sample tex file for a project. Replace the generated tex file with this MWE:
```
\def\sphinxdocclass{report}
\documentclass{sphinxmanual}
\usepackage{sphinx}
\usepackage{hyperref}
\usepackage{blindtext}
\begin{document}
\section{Overview}
\blindtext
\end{document}
```

**Expected behavior**
The output should look the same on Windows and Linux, but it does not.
Note that I reuse the generated latex folder on both, Windows and Linux, and do not let sphinx regenerate the folder / styles again (thus there should be no differences in the .sty files etc.).

When I remove `\usepackage{sphinx}` from the MWE .tex file, the output from both systems is identical. 

**Your project**
[demo.zip](https://github.com/sphinx-doc/sphinx/files/2765188/demo.zip)

**Screenshots**
Linux:
sphinx-lin
Windows:
sphinx-win

**Environment info**
- OS: Linux, Windows
- Python version: 3.6.8 and 3.6.5
- Sphinx version: 1.7.8 and 1.8.3
- Extra tools: TeX Live 2018

jfbu:
Mumble

### Relates

- https://github.com/FrankMittelbach/fmitex/issues/6

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -30,6 +30,8 @@ Bugs fixed
   in an admonition
 * #5231: "make html" does not read and build "po" files in "locale" dir
 * #5954: ``:scale:`` image option may break PDF build if image in an admonition
+* #5960: LaTeX: modified PDF layout since September 2018 TeXLive update of
+  :file:`parskip.sty`
 
 Testing
 --------
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -220,7 +220,13 @@
 % For floating figures in the text. Better to load after float.
 \RequirePackage{wrapfig}
 % Separate paragraphs by space by default.
-\RequirePackage{parskip}
+\IfFileExists{parskip-2001-04-09.sty}% since September 2018 TeXLive update
+% new parskip.sty, but let it rollback to old one.
+% hopefully TeX installation not broken and LaTeX kernel not too old
+   {\RequirePackage{parskip}[=v1]}
+% standard one from 1989. Admittedly \section of article/book gives possibly
+% anomalous spacing, but we can't require September 2018 release for some time.
+   {\RequirePackage{parskip}}
 % For parsed-literal blocks.
 \RequirePackage{alltt}
 % Display "real" single quotes in literal blocks.


tk0miya (approved):

nioncode:
Actually I now prefer the layout of the new parskip package, because the vertical spacing after section headings really is too much with the old one.
Are there any strong objections to bundle the new parskip with Sphinx and copying it to the latex build directory? That's what I now have in my conf.py.

The downside is that the new Sphinx release would then have a different pdf output than earlier versions.

jfbu:
@nioncode Thanks for suggestion, but no. The new parskip is used to modify core LaTeX macros, as well as core `titlesec` macros and an amsmath macro, modifying their behaviour in presence of core TeX skip parameter `\parskip` being non zero. Is this correct methodology? What if user directly does some `\setlength{\parskip}{...}` without loading parskip package? Then the (now considered) "misbehaviour" of original macros shows. Is `parskip` a core LaTeX package, is it part of the core or base LaTeX distribution? Shouldn't bugs be corrected upstream? Definitely, a novel package was needed, but not modifying behaviour of old package which had remained exactly the same since 1989.

jfbu:

nioncode:
I don't really know what's right or wrong in LaTeX terms, but to me it seems like the new parskip really has the right behavior now, because previously spacing after section headers were really big (which can be seen by having many (sub) sections that have only 1-2 lines; the spacing then is just weird).

My understanding of `\parskip` was that one should never modify this manually anyway, because it breaks in weird ways and that's what the parskip package is actually for.

There should at least be a way to NOT use the old revision in Sphinx. I think the only way after your merge would be to have a copy of parskip.sty in the build folder that ignores the old revision parameter. Is this the suggested usage from your point of view?

jfbu:
@nioncode The new `parskip` is not usable on old systems due to this

```
\NeedsTeXFormat{LaTeX2e}[2018-04-01]
```

line at its start. But assuming you deploy your project only on TeXLive 2018 type distributions (I don't recall offhand if initial TeXLive 2018 had LaTeX 2018-04-01, will check later), as you observed initial release did not have new parskip so you can add it your project by putting new file `parskip.sty` in your source repertory and add this to `conf.py`:

```
latex_elements = {
    'passoptionstopackages': r'\usepackage{parskip}',
}
latex_additional_files = ['parskip.sty']
```

This will override the `sphinx.sty` issued `\usepackage{parskip}[=v1]` on those (updated after September) TeXLive 2018 releases, which **already** have the new parskip. The abuse of `'passoptionstopackages'` key is to load it early in preamble, before actual loading of `sphinx.sty` itself.

nioncode:
Awesome, thanks for being so responsive and for fixing the original issue so quickly!

jfbu:
If you want to use the new parskip on old systems, it should work by applying my previous comment but with  a custom `parskip.sty` modified as indicated in this diff:

```diff
diff --git a/parskip.sty b/parskip.sty
index 5e3b3db..c0b9f33 100644
--- a/parskip.sty
+++ b/parskip.sty
@@ -36,11 +36,14 @@
 %%   (C) Copyright 1989 H.Partl, TU Wien
 %%   (C) Copyright 2001 Robin Fairbairns
 %%   (C) Copyright 2018-2019 Frank Mittelbach
-\NeedsTeXFormat{LaTeX2e}[2018-04-01]
+%%
+%% HACKED FOR PERSONAL USE
+\NeedsTeXFormat{LaTeX2e}%%[2018-04-01]
 
-\DeclareRelease       {v1}{2001-04-09}{parskip-2001-04-09.sty}
-\DeclareCurrentRelease{v2}{2018-08-24}
-\ProvidesPackage{parskip}[2019-01-16 v2.0c non-zero parskip adjustments]
+%%\DeclareRelease       {v1}{2001-04-09}{parskip-2001-04-09.sty}
+%%\DeclareCurrentRelease{v2}{2018-08-24}
+%%\ProvidesPackage{parskip}[2019-01-16 v2.0c non-zero parskip adjustments]
+\ProvidesPackage{parskip}[2019/01/16 v2.0c non-zero parskip adjustments]
 \RequirePackage{kvoptions}
 \SetupKeyvalOptions{family=parskip,prefix=parskip@}
 \DeclareStringOption[0pt]{indent}[\parindent]
```

(untested).

I neither recommend nor endorse this and maybe legally you will need to remove some more of the comments in the file.

Edit: this indeed works on old systems. And it works on new (2018/04/01) LaTeX too, whether or not they have the new parskip installed.

PR #5976, opened 2019-01-21
Fix #5959: classifier delimiter is not shown

### LINKED ISSUE #5959: BUG: numpydoc + autosumarry broken
#4611 broke rendering for me of a `numpydoc` + `autosummary` `html` build where the input is like:
```
...
    Parameters
    ----------
    data : array, shape (n_chan,)
        The data values to plot.
...
```
Which `numpydoc` turns into something like:
```
...
:Parameters:

    **data** : array, shape (n_chan,)
        The data values to plot.
...
```
On the commit in `master` before this (b04bfcfab7b4a47dca1a5c9dda88e810ffc81bcd) I get the correct output with `:` between items:

screenshot from 2019-01-15 15-03-42

On `master` after this PR (on commit 7459d8771af1afaf6c2342b689fb70a56ba00efa) the colon is gone (`<span>` missing in output):

screenshot from 2019-01-15 14-57-04

I'm not sure how to fix this because I'm out of my depth, though :(

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5959

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -537,6 +537,12 @@ dl.glossary dt {
     font-style: oblique;
 }
 
+.classifier:before {
+    font-style: normal;
+    margin: 0.5em;
+    content: ":";
+}
+
 abbr, acronym {
     border-bottom: dotted 1px;
     cursor: help;


sphinx/themes/basic/static/basic.css_t
```diff
@@ -537,6 +537,12 @@ dl.glossary dt {
     font-style: oblique;
 }
 
+.classifier:before {
+    font-style: normal;
+    margin: 0.5em;
+    content: ":";
+}
```
tk0miya, line 544: In HTML4CSS1 writer, `<span class="classifier-delimiter">:</span>` is used for the delimiter. On the other hand, In HTML5-polyglot writer, it is represented by CSS hack.
Since 2.0, we adopt HTM5-polyglot style.

jschueller (approved):
Does fix our problem! Thanks!

tk0miya:
Thank you for confirming! just merged.

larsoner:
Indeed, thanks for the quick fix!

PR #5979, opened 2019-01-22
#4649 

### LINKED ISSUE #4649: Broken Search
### Problem
- Broken Search

build docs with: $make html works as a charm. Pointing docs via webserver is working fine, but search doesn't works:
- cannot find any numbers: index.rst contain "1234567", which is displayed on browser,but if I search it no results displayed. This happen also with IP Address, and others numbers;
- Lots of words with dots doesn't appear: if I search www.example.com, search doesn't find it.

#### Error logs / results
No errors displayed

#### Expected results
Search results.

Private project.

### Environment info
- OS: Linux Ubuntu 14.04.5
- Python version: 2.7.6
- Sphinx version: 1.7.0
- same behaviour in multiple browsers.

TimKam:
Subject: Fix: search always returns nothing when one term is shorter than 3

### Feature or Bugfix
- Bugfix

### Purpose
The search function considers every page a "miss"/no match if the search contains a term with length <= 2; 
This is problem when users search for filenames, identifiers, et cetera.
This PR changes the behavior to be more lenient in such cases: if an additional search term is of length <= 2, it is ignored (no longer a necessary condition for a match).

### Relates
- Closes #4649

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -428,12 +428,12 @@ var Search = {
       var valid = true;
 
       // check if all requirements are matched
-      var filteredTerms = // as search terms with length < 3 are discarded: ignore
+      var filteredTermCount = // as search terms with length < 3 are discarded: ignore
         searchterms.filter(function(term){return term.length > 2}).length
-      var allRequirementsMatched = fileMap[file].length != searchterms.length &&
-        fileMap[file].length != filteredTerms
-      if (allRequirementsMatched)
-          continue;
+      if (
+        fileMap[file].length != searchterms.length &&
+        fileMap[file].length != filteredTermCount
+      ) continue;
 
       // ensure that none of the excluded terms is in the search result
       for (i = 0; i < excluded.length; i++) {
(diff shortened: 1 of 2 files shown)


sphinx/themes/basic/static/searchtools.js
```diff
@@ -428,7 +428,11 @@ var Search = {
       var valid = true;
 
       // check if all requirements are matched
-      if (fileMap[file].length != searchterms.length)
+      var filteredTerms = // as search terms with length < 3 are discarded: ignore
```
tk0miya: I think this value is a count of terms. How about `filteredTermCount`?

sphinx/themes/basic/static/searchtools.js
```diff
@@ -428,7 +428,11 @@ var Search = {
...
+        searchterms.filter(function(term){return term.length > 2}).length
+      var allRequirementsMatched = fileMap[file].length != searchterms.length &&
```
tk0miya: IMO, this variable is not needed. I prefer to implement this as simple if-condition. What do you think?

tk0miya (commented):
LGTM with nits

sphinx/themes/basic/static/searchtools.js
TimKam: I agree. It's more concise without this assignment. I will work in the feedback and merge when the CI is green.

PR #5984, opened 2019-01-26
Stop to use cElementTree (deprecated since py33)

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- refs: https://docs.python.org/3/library/xml.etree.elementtree.html


tk0miya:
It seems more refactoring is needed. So withdrawing.

Issue #5956, opened 2019-01-14
Let LaTeX be more rational in handling image dimensions


jfbu:
Currently, here is situation for image directive with options width, height, scale:

- if neither width nor height, then `\sphinxincludegraphics` will make sure that the graphics width does not exceed "`\linewidth`", (after obeying ``:scale:`` option since Sphinx 1.5),

- if either width or height is explicitly set, then no rescaling happens. https://github.com/sphinx-doc/sphinx/blob/0081482eb7b65cb286afd6ed1c466990545226be/sphinx/texinputs/sphinx.sty#L704-L707

Recently an exception to this has been added (#5939): inside an admonition a too tall image can break the PDF build, so new macro `\sphinxsafeincludegraphics` is used which makes sure the height is never too tall. PR #5955 will improve this by making sure `\scalebox` is never used as it could ruin the cap on height, by increasing again the size.

I suggest that the "`\linewidth`" limitation should always be applied whether or not the user has added ``:width:`` and/or ``:height:`` option. Indeed these options mostly target usually HTML builds, and it does not make sense that they should be absolutely obeyed in LaTeX even if the page size (which can be modified) is too small.

And the height also should be limited to "`\textheight`", even outside of admonitions.

This is easy to do in LaTeX, because when one specifies both ``width=``, ``height=``, then with ``keepaspectratio`` keyword as last option, the `\includegraphics` macro will ensure that the image original aspect ratio is preserved and that width and height are maximal while obeying the specifications.

Basically, this all means (once #5955 is merged thus removing some complications with LaTeX handling of `scale=` option in presence of `width=` or `height=`) we should make `\sphinxincludegraphics` an alias for `\sphinxsafeincludegraphics` (the `\spx@image@maxheight` is by default set to `\textheight`): what this does is

- compute the graphics dimensions from user options,

- if width turns out bigger than "`\linewidth`", or  height bigger than "`\textheight`" (or suitably smaller value if picture inside a frame), then downscale picture (keeping aspect ratio) so that it fits.

The only thing to decide I see is this: when user specifies both ``:width:`` and ``:height:`` and those are not too big, the picture will obey exactly the user demands, hence possibly get distorted ; if one or both are too bigs, the code will give a picture keep same aspect ratio as native dimensions, fitting in both width and height. Thus

[...]

PR #5992, opened 2019-01-28
LaTeX: safer image inclusion

jfbu:
Closes: #5956

However currently I can't test at my locale because all pdflatex compilations are skipped (cf https://github.com/sphinx-doc/sphinx/pull/5384#issuecomment-458234474), and besides the tex files are at some tmp locations which I don't know about (cf https://github.com/sphinx-doc/sphinx/pull/5902#issuecomment-458236847).

This PR implements resizing of images for PDF inclusions to respect the available horizontal and vertical space. If options width and height are both used, the aspect ratio is according to them.

Formerly resizing applied only in absence of width and height (except in some context where too tall images causes PDF build failure).

About testing, maybe it is not needed to do the pdflatex runs ? (I did the tests at my locale, with this PR rebased onto 1.8 branch, using "SphinxTests" as filename as is mandatory there for the pdflatex runs).


tests/roots/test-latex-includegraphics/conf.py
```diff
@@ -0,0 +1,47 @@
...
+    \fi
+          }%
+    \ifdim\dimexpr\ratiocheck-1pt\relax > 0.01pt
+      \ASPECTRATIOERROR
+    \fi
+}
+\def\sphinxincludegraphics#1#{\tempincludegraphics#1}
+\makeatother
+''',
+}
```
jfbu, line 49: This looks complicated, but check (if possible) in build repertory the file ending in "-dimensions.txt" and it will become clearer.

tests/roots/test-latex-includegraphics/index.rst
```diff
@@ -0,0 +1,30 @@
...
+Tests with only width or height
+-------------------------------
+
+.. topic:: Oversized images
+
+   .. image:: img.png
+      :width: 1000
+
+.. image:: img.png
+   :height: 1000
```
jfbu, line 36: Due to the code in `conf.py` this test file must only test oversized images, so that `\spx@image@requiredheight` and `\spx@image@requiredwidth` as used in patched from conf.py  include graphics do actually refer to current image and not a previous one.

tests/test_build_latex.py
```diff
@@ -1383,3 +1383,19 @@ def test_default_latex_documents():
     expected = [('index', 'stasi.tex', 'STASI™ Documentation',
                  r"Wolfgang Schäuble \& G'Beckstein.\@{}", 'manual')]
     assert default_latex_documents(config) == expected
+
+
+@skip_if_requested
+@skip_if_stylefiles_notfound
+@pytest.mark.sphinx('latex', testroot='latex-includegraphics',
+    confoverrides={'latex_documents': [
+                       ('index', 'python.tex', 'Test PDF build',
```
jfbu: on 1.8 branch I would use here SphinxTests

tests/test_build_latex.py
```diff
@@ -1383,3 +1383,19 @@ def test_default_latex_documents():
...
+                        'Georg Brandl', 'howto')
+                   ]})
+def test_includegraphics_oversized(app, status, warning):
+    app.builder.build_all()
+    print(status.getvalue())
+    print(warning.getvalue())
+    compile_latex_document(app)
+    result = (app.outdir / 'pdflatex/python-dimensions.txt').text(encoding='utf8')
```
jfbu: I would use SphinxTests-dimensions.txt on 1.8 branch. Oh I see now that I had initially here 

```
compile_latex_document(app, 'python.tex')
```

but I dropped the second argument when rebasing for testing on 1.8 and forgot to put it back when moving again to master.

Anyway this should work, but the pdflatex build is skipped as all others.

tests/test_build_latex.py
```diff
@@ -1383,3 +1383,19 @@ def test_default_latex_documents():
...
+    print(result)
```
jfbu: I hesitated about analyzing the output file with dimension informations from Python, but finally did it all from internal LaTeX code (see the conf.py)

jfbu (commented):
The testing forgot the case that the image is too wide and after adjusting the width to linewidth then the height is still too much. Indeed the used image is about as wide as tall, one would need an elungated image rather.

tests/roots/test-latex-includegraphics/conf.py
```diff
@@ -0,0 +1,47 @@
+master_doc = 'index'
+
+exclude_patterns = ['_build']
+
+# only to be used on image inclusions with both width and height
```
jfbu: sorry that comment was for earlier version, I then refactored to work with width and/or height. But the index.rst must only contain image inclusions leading to either oversized width or oversized height.

jfbu:
I force-pushed to fix some mistakes I had done, sorry about that. The extended test of PDF build showed my mistakes and now passes.

jfbu:
The test, when executed with pdflatex produces this file which I will comment to explain

- ```
  original options = width=1000\sphinxpxdimen ,height=200\sphinxpxdimen 
  width = 469.75502pt, linewidth = 469.75502pt
  height = 93.92514pt, maxheight = 650.43001pt
  obtained H/W = 0.19995pt
  asked for H/W = 0.2pt
  ratio of ratios = 0.99977pt
  ```

  The requested width is too much, it got reduced to current linewdith, exactly. The height was rescaled in a way preserving asked for aspect ratio of H/W = 1/5 (up to intrinsic errors due to TeX smallest dimension unit being 1/65536 pt and the fact that the code does not check the original options to find expected aspect ratio but the box dimensions from first pass with `\includegraphics` which is only precise up to 3 or four decimal digits in its internal computations). (the "pt" in `0.2pt` and `0.99977pt` above is artefact of how the ratio is computed by TeX code).


- ```
  original options = width=200\sphinxpxdimen ,height=1000\sphinxpxdimen 
  width = 124.6978pt, linewidth = 458.95503pt
  height = 623.63002pt, maxheight = 623.63002pt
  obtained H/W = 5.00113pt
  asked for H/W = 5.0pt
  ratio of ratios = 1.00023pt
  ```

  Here it is requested height which is too much. Image downscaled to fit on a page, preserved `H/W` = 5 ratio.

- ```
  original options = width=1000\sphinxpxdimen ,height=3000\sphinxpxdimen 
  width = 207.73604pt, linewidth = 458.95503pt
  height = 623.63002pt, maxheight = 623.63002pt
  obtained H/W = 3.00203pt
  asked for H/W = 3.0pt
  ratio of ratios = 1.00067pt
  ```

  Here both width and height are too much, and when width is diminished to `\linewidth` then height is still too big. So it is height which ends up being rescaled to maxheight, and aspect ratio of `3` is preserved as requested by width and height options in `image` directive.

- ```
  original options = width=1000\sphinxpxdimen 
  width = 108.94032pt, linewidth = 458.95503pt
  height = 623.63002pt, maxheight = 623.63002pt
  obtained H/W = 5.72452pt
  ```

  Here width is too big, but image has native ratio H/W which is big, so when width is rescaled to linewidth, it is still not enough, so it is height which is reset to maxheight, and width is adjusted to preserve native image dimensions ratio.

- ```
  original options = height=1000\sphinxpxdimen 
  width = 469.75502pt, linewidth = 469.75502pt
  height = 82.06029pt, maxheight = 650.43001pt
  obtained H/W = 0.17468pt
  ```

  Here height is too big, but image is very wide, so when reducing it to maxheight, the width is still too big, so width is rescaled to linewidth and ratio H/W is the native one from image (up to errors mentioned above).

The `\sphinxincludegraphics` could be improved to obtain the asked-for ratio from parsing the options, but when only width or height is specified, it can get the native ratio only from trusting the box dimensions computed by first pass `\includegraphics` which as said above introduces about `0.1%` error at worst, which is acceptable I think. So I did not bother to add extra code for case with both explicit width and height.

tests/roots/test-latex-includegraphics/conf.py
jfbu: I have added detailed explanations of output on test file

tests/test_build_latex.py
tk0miya: Is this option still needed on master branch?

tests/test_build_latex.py
tk0miya: I think this debug print would not display any information even if failed...

tk0miya (approved):
I don't review TeX macros in detail. But it works fine on my local :-)
LGTM with nits!

tests/test_build_latex.py
jfbu: No, it works fine without it.

tests/test_build_latex.py
jfbu: You are right (of course :) ).

jfbu:
Thanks for reviewing! updated...

jfbu:
Thanks for reviewing I am merging now...

PR #5996, opened 2019-02-01
Show tempdir header on testing (refs: #5902)

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- refs: https://github.com/sphinx-doc/sphinx/pull/5902#issuecomment-458236847
- This shows "base tempdir" as a header of pytest.
```
======================================================== test session starts =========================================================
platform darwin -- Python 3.7.1, pytest-4.1.1, py-1.7.0, pluggy-0.8.1
cachedir: .tox/py37/.pytest_cache
libraries: Sphinx-2.0.0+/583e19427, docutils-0.14
base tempdir: /var/folders/rj/80_5mv452nsfnhkvkyqjwj040000gp/T/pytest-of-tkomiya/pytest-209
rootdir: /Users/tkomiya/work/sphinx, inifile: setup.cfg
plugins: cov-2.6.1
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,8 +27,12 @@ def rootdir():
 
 
 def pytest_report_header(config):
-    return ("libraries: Sphinx-%s, docutils-%s" %
-            (sphinx.__display_version__, docutils.__version__))
+    header = ("libraries: Sphinx-%s, docutils-%s" %
+              (sphinx.__display_version__, docutils.__version__))
+    if hasattr(config, '_tmp_path_factory'):
+        header += "\nbase tempdir: %s" % config._tmp_path_factory.getbasetemp()
+
+    return header
 
 
 def pytest_assertrepr_compare(op, left, right):


jfbu (commented):
LGTM with nits (and thanks): but with

```
$ export SPHINX_TEST_TEMPDIR=tests/build/
```

the build output is according to the set variable, but the reported repertory is not this one but another one which remains empty and would have been used with this variable unset.

This being said, and relative to my https://github.com/sphinx-doc/sphinx/pull/5902#issuecomment-458236847 I am *much embarrassed* to acknowledge I had completely forgotten about `TMPDIR` env variable which is set in my mac os x environment, and it is used by pytest, with `$TMPDIR/pytest-of-username/pytest-<number>` at my locale, so my feature request was based on ignorabimus state. I think your PR provides useful addition, for dummies like myself, but may be unneeded by the real pros out there.

jfbu:
@tk0miya by the way when I do

```
$ export SPHINX_TEST_TEMPDIR=tests/build/
```

it appears I must delete explicitly the tests/build/ repertory after each 

```
```

else the tests are not done again. 

More precisely, perhaps pytest has clever md5 hashes and recreates output files only if modified (I did notice that files like `latexmkrc` from templates or `sphinxhighlight.sty` appear to be recreated each time but not `python.tex` files.) Can I trust files really are updated if they need to  be?

jfbu (approved):
@tk0miya Don't worry about my previous comment about

```
$ export SPHINX_TEST_TEMPDIR=tests/build/
```

I have gotten some problems from files not being deleted in ``tests/build/`` from previous test, as they can cause LaTeX build failures.

As anyhow this usage of ``SPHINX_TEST_TEMPDIR`` env variable is deprecated I think I should stop forcing some explicit location for the build repertory  and simply get used to examine build test files at the location reported by the modified header from this PR. It is not important that this location is wrong if ``SPHINX_TEST_TEMPDIR`` is set, because using it is not only deprecated, but even broken.

tk0miya:
Okay, I understand your problem. It's not intended behavior.
Indeed, I marked `SPHINX_TEST_TEMPDIR` "deprecated" at #5902. But it does not mean destroy it instantly.
I'd like to revert it partially to remove previous testing results if `SPHINX_TEST_TEMPDIR` set in another PR.

Thanks,

Issue #6004, opened 2019-02-02
PDF title page has extra space after author (or authors) name (names)


jfbu:
For example for `'manual'`:

![capture d ecran 2019-02-02 a 17 48 36](https://user-images.githubusercontent.com/2589111/52166856-ec5e2c00-2712-11e9-9294-98e482ff3fc5.png)

I feel

![capture d ecran 2019-02-02 a 17 49 15](https://user-images.githubusercontent.com/2589111/52166858-f122e000-2712-11e9-8355-5d2e99232406.png)

would be nicer.

I don't know if this qualifies as bug. It shows more prominently in `'howto'` case, because there is less vertical whitespace.

The cause is in this code from `sphinxmanual.cls`:

```latex
        \begin{tabular}[t]{c}
          \@author
        \end{tabular}
```

which is copied-pasted from standard `\maketitle` macro, but the original LaTeX article/book class does that in a `center` environment, and Sphinx does it in `flushright`. The `tabular` always adds (intercolumn, `\tabcolsep`) space at left and right, and this what is seen at right.

PR #6005, opened 2019-02-02
LaTeX: remove extra space after author on title page in PDF

jfbu:
Closes: #6004

Comment: ``\begin{tabular}[@{}c@{}]`` does not work in case of multiple
authors separated by \and in latex_documents, because latex has this
definition of \and :

```
macro:->\end {tabular}\hskip 1em \@plus .17fil\begin {tabular}[t]{c}
```

which means we would need to modify it to read

```
macro:->\end {tabular}\hskip 1em \@plus .17fil\begin {tabular}[t]{@{}c@{}}
```

(or only one ``@{}`` at end) modifying space between two authors.

If requalified as bugfix, I will rebase on 1.8 and update CHANGES accordingly.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,8 @@ Features added
 Bugs fixed
 ----------
 
+* LaTeX: Remove extraneous space after author names on PDF title page (refs: #6004)
+
 Testing
 --------
 
--- a/sphinx/texinputs/sphinxhowto.cls
+++ b/sphinx/texinputs/sphinxhowto.cls
@@ -46,7 +46,7 @@
     {\Large
       \begin{tabular}[t]{c}
         \@author
-      \end{tabular}}\par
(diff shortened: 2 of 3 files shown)


CHANGES
```diff
@@ -190,6 +190,7 @@ Bugs fixed
 * QtHelp: OS dependent path separator is used in .qhp file
 * HTML search: search always returns nothing when multiple search terms are
   used and one term is shorter than three characters
+* Remove extraneous space after author names on PDF title page (refs: #6004)
```
tk0miya: Could you add "LaTeX" prefix here?

tk0miya (approved):
LGTM with nits.
Either branch is okay to me :-)

CHANGES
jfbu: Done, and rebased on 1.8. Thanks for reviewing :)

jfbu:

PR #6015, opened 2019-02-03
Close #1851: Allow to omit an argument for code-block directive

### LINKED ISSUE #1851: code-block default language
Hi,
it would be useful and smart to let the language in code-block to be the global one by default. For instance if I write

```
.. highlight:: c

.. code-block::

   int a;
```

I would expect the code block highlighting to be in c. Does it make sense?

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #1851

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -167,6 +167,8 @@ Features added
 * #1341 the HTML search considers words that contain a search term of length
   three or longer a match.
 * #4611: epub: Show warning for duplicated ToC entries
+* #1851: Allow to omit an argument for :rst:dir:`code-block` directive.  If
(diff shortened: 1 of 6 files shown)


shimizukawa (approved):
LGTM!

It seems good to be able to use code-block with the same usability as `::`.

jfbu:
+1 I think this is good-idea. I find `code-block` gives better sources than `::` which is too discreet. But specifying langage each time was cumbersome.

tk0miya:
Thank you for comment :-)
Merged.

PR #6018, opened 2019-02-04
#6016 add placeholder for search summary

### LINKED ISSUE #6016: Make HTML search result keep their position when search finishes
I noticed this on the search page of the online Python documentation (https://docs.python.org/3/search.html), which uses Sphinx. I created on issue about it on the Python issue tracker and was asked to open an issue here since it's better to fix this on the Sphinx side. I didn't notice the same issue on Sphinx's own search page though (http://www.sphinx-doc.org/en/master/search.html): it seems to display all search results together. I don't know what causes that difference.

The search function starts displaying results while the search continues in the background, which is very nice.

Only problem is: when the search finishes, a line with the text "Search finished, found x page(s) matching the search query." appears which pushes all the search results down a bit. When the result you were looking for was already displayed, you suddenly have to aim the mouse cursor at a new position, or it can happen that you accidentally open the wrong link because of the results not staying in their place.

Is it possible to allocate space for the "Search finished, ..." line from the beginning, so that search results stay in the same place the whole time?

TimKam:
Subject: Add placeholder for search summary

### Feature or Bugfix
- Feature (enhancement)

### Purpose
- Adds placeholder for search summary to prevent search result links from changing position when search terminates

### Details
The change is kinda low-tech, but more stable than adjustments via CSS that would always rely on the style of the template a project is using.

### Relates
- Closes #6016


tk0miya (approved):
LGTM!

PR #6025, opened 2019-02-05
Introduce SphinxRole class as a base class of roles

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- So far, we use functions to implement roles. But its argument is difficult to understand for beginners. And it is hard to access `env` and `config` objects originated Sphinx.
- This introduces `SphinxRole` class as a base class of roles. It works like "Directive" class for directive implementations. Developers can create custom roles by subclassing it and override `run()` method.
- As a first step, I replaced `:abbr:` role by the base class.
- If approved, I'll continue to replace other role implemantions.


shimizukawa (commented):
No objection.

PR #6028, opened 2019-02-06
Ensure the graphviz filenames are reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org) effort, we noticed that sphinx could generate output that is not reproducible.

In particular, the graphviz extension module would construct filenames based on, inter alia, the contents of the `options` dictionary.

As this contained the absolute build path of the source file embedded in the `docname` variable this meant that builds of documentation were not independent of where on a filesystem they were built from

Example filenames might be:

```
  -  html/_images/graphviz-9e71e0f9ba91d0842b51211b676ec4adb7e7afb8.png
  +  html/_images/graphviz-6241bbfd7ac6bd4e2ad9af451ab0dfb8719988d2.png
```

We fix this by limiting how much of the `docname` variable ends up in the final constructed filename; I assume there is a good reason for including the `options` dictionary in the first place, otherwise we could simply omit it.

This was originally filed in @Debian as [#921513](https://bugs.debian.org/921513).

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -216,7 +216,9 @@ def render_dot(self, code, options, format, prefix='graphviz'):
     # type: (nodes.NodeVisitor, unicode, Dict, unicode, unicode) -> Tuple[unicode, unicode]
     """Render graphviz code into a PNG or PDF output file."""
     graphviz_dot = options.get('graphviz_dot', self.builder.config.graphviz_dot)
-    hashkey = (code + str(options) + str(graphviz_dot) +
+    options_for_hash = options.copy()
+    options_for_hash = path.basename(options_for_hash.pop('docname', ''))
+    hashkey = (code + str(options_for_hash) + str(graphviz_dot) +
                str(self.builder.config.graphviz_dot_args)).encode('utf-8')
 
     fname = '%s-%s.%s' % (prefix, sha1(hashkey).hexdigest(), format)


tk0miya:
Hmm, it seems the docname was wrong.
We have to replace `current_source` by `self.env.docname` like 00f476040e30371044437860ef07cb023b94489a did.

tk0miya:
Note: this is a bugfix. so please rebase this to 1.8 branch. then I will merge this to 1.8.5 release.
Thanks,

lamby:
Rebased against `1.8` branch.

mitya57:
@lamby Please also edit this pull request and change target branch from master to 1.8.

lamby:
Ah, thanks. No wonder it was still saying "conflict"...

mitya57:
> We fix this by limiting how much of the `docname` variable ends up in the final constructed filename; I assume there is a good reason for including the `options` dictionary in the first place, otherwise we could simply omit it.

I want to note that your patch actually omits all options except the filename. Though in my opinion it is fine, as the resulting image depends only on source file name and graphviz arguments, which are present in the hash anyway.

lamby:
> the resulting image depends only on source file name

Nod. Note that it cannot include include the *absolute* path to the source file name, otherwise it would be unreproducible as before. ;-)

tk0miya:
As I commented above, the value of docname must be wrong. So this is not only "not reproducible", but also a bug.
So we need to fix the value itself. `self.env.docname` is more appropriate for this case.
https://github.com/sphinx-doc/sphinx/blob/12c614911ac1ea94e88ec781fd36281ffd026f0a/sphinx/ext/graphviz.py#L199-L201

lamby:
> As I commented above, the value of docname must be wrong.

Don't follow that, sorry...

tk0miya:
Merged #6111 instead.
Thank you for reporting :-)

PR #6034, opened 2019-02-07
Close #5993: Use envvar ``$SSL_CERT_FILE`` to the default for tls_cacerts

### LINKED ISSUE #5993: Enable sphinx-build to use environment variables for certificates (Requests/urllib2)
It appears that although Sphinx uses `Requests` a user may still need to set the certificate directory in their conf.py manually via a `tls_cacerts = '/path/to/your/certs'` as it doesn't appear to honor the `REQUESTS_CA_BUNDLE` or the other SSL environment variables commonly used to pass trusted certs. This has made using `sphinx-build linkcheck` very tricky in an enterprise environment with a proxy, especially with some of the resources also require a login to view, and hardcoding the login credentials anywhere is a bad idea.

It would be good to document whether it supports a directory of certificates in the openssl style ie `/usr/local/etc/openssl/certs` or if it requires a bundle of all the certs in a single `cacerts.pem` file.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #5993 
- I choose `SSL_CERT_FILE`. It seems commonly used in other softwares.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -176,7 +176,8 @@ Features added
 * #6016: HTML search: A placeholder for the search summary prevents search
   result links from changing their position when the search terminates.  This
   makes navigating search results easier.
-  
+* #5993: Use envvar ``$SSL_CERT_FILE`` to the default for :confval:`tls_cacerts`
+
 Bugs fixed
 ----------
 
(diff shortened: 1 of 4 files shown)


tk0miya:
I'm withdrawing this because Sphinx supports `REQUESTS_CA_BUNDLE`. No new envvar is needed.

Issue #248, opened 2015-01-02
autosummary: option for all modules of a package


shimizukawa:
Like in this example:

``` rst

.. autosummary::
   :toctree: generated/

   MaskedArray.base
   MaskedArray.ctypes
   MaskedArray.dtype
   MaskedArray.flags
```

one currently has to explicitly state the modules that should be documented by autosummary.

Please add option to automatically create stub files for all files in a package:

coule be like: 

``` rst

.. autosummary::
   :toctree: generated
   :nosignatures:
   :fullpackage: mypackage_toplevel_dir
```

This should then automatically include all modules in teh package "mypackage_toplevel_dir".

P.S.: I hope that this is not a misunderstanding or misconfiguration on my side. maybe this is already implemented?

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/248
- Originally reported by: [timmie](https://bitbucket.org/timmie)
- Originally created at: 2009-08-29T17:55:22.407

shimizukawa:
_From [jasongrout](https://bitbucket.org/jasongrout) on 2014-08-23 05:03:39+00:00_

FYI, the astropy automodsumm has functionality something like this (see https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/sphinx/ext/automodsumm.py).  It looks like sphinxydoc might be something good to look at too from https://github.com/sphinx-doc/sphinx/issues/317 (see https://github.com/Ciantic/sphinkydoc)

maikelmeyers:
Is this still not implemented? I would like to see this feature.

ben-spiller:
+1 

Would be best as a general capability for autosummary to include/exclude members (whether members of a module or a class/exception), similar to how automodule/autoclass work. 

e.g.
```
.. autosummary:: MyModuleOrClass
  :members:
  :private-members:
  :inherited-members:
  :exclude-members: not_this_one
```

We already have logic for getting the members of a module. Hopefully we can find some way to also reuse the autodoc code for include/excludes for autosummary.

PR #6040, opened 2019-02-08
Feature autosummary packages

woutdenolf:
Subject: recursively add modules and sub-packages to a package with autosummary

### Feature or Bugfix

- Feature

### Purpose
- Add two new items to the module template: "modules" and "packages".
- This allows to automatically add modules and sub-packages for modules that are packages.
- It works recursively.

For example a [project](https://github.com/woutdenolf/sphinx_example):
```script
package/
|-- __init__.py
|-- foo.py
|
|-- subpackage/
|   |-- __init__.py
|   |-- foo.py
|
|-- doc/
|   |-- source
|       |--index.rst
|       |--conf.py
```
with this index.rst:
```script
API Reference
=============

.. rubric:: Packages

.. autosummary::
   :toctree: apidocs

    package

```
will generate
```script
doc/source/apidocs
|-- package.rst
|
|-- modules/
|   |-- package.foo.rst
|
|-- packages/
|   |-- package.subpackage.rst
|   |-- modules/
|       |-- package.subpackage.foo.rst
```
which renders as

### Detail
- Recursion is disabled by default (old behavior)
- Parameter to control recursion depth
- Exclude/include pattern is not supported but may be useful

### Relates
- issue #248


doc/usage/extensions/autosummary.rst
```diff
@@ -236,6 +236,19 @@ The following variables available in the templates:
    List containing names of "public" attributes in the class.  Only available
    for classes.
 
+.. data:: modules
+
+   List containing names of "public" modules in the package.  Only available for
+   modules that are packages.
+
+   .. versionadded:: 2.0.0
```
tk0miya: We usually use `2.0` instead for versionadded.

sphinx/ext/autosummary/generate.py
```diff
@@ -188,6 +189,29 @@ def get_members(obj, typ, include_public=[], imported=True):
                           if x in include_public or not x.startswith('_')]
                 return public, items
 
+            def get_package_members(out_dict, include_public=[], imported=True):
```
tk0miya: Why did you not use return value? I feel `out_dict` is strange interface.

In addition, is `include_public` needed? It seems not given.

tk0miya: Please add type annotations here.

sphinx/ext/autosummary/generate.py
```diff
@@ -188,6 +189,29 @@ def get_members(obj, typ, include_public=[], imported=True):
...
+                out_dict['packages'], out_dict['all_packages'] = [], []
+                for _, modname, ispkg in pkgutil.iter_modules(obj.__path__):
+                    fullname = name + '.' + modname
+                    try:
+                        import_by_name(fullname)
+                    except ImportError as e:
+                        warn('[autosummary] failed to import %s: %s' % (fullname, e))
+                        continue
+                    if ispkg:
+                        kpublic, kall = 'packages', 'all_packages'
```
tk0miya: What does "k" prefix mean?

sphinx/ext/autosummary/generate.py
```diff
@@ -188,6 +189,29 @@ def get_members(obj, typ, include_public=[], imported=True):
...
                 return public, items
 
+            def get_package_members(out_dict, include_public=[], imported=True):
+                # https://docs.python.org/3/reference/import.html#packages
+                # ... any module that contains a __path__ attribute is
+                # considered a package
+                if not hasattr(obj, '__path__'):
+                    return
+                out_dict['modules'], out_dict['all_modules'] = [], []
+                out_dict['packages'], out_dict['all_packages'] = [], []
```
tk0miya: When are `all_modules` and `all_packages` used?

tk0miya (changes requested):
Great work! LGTM with nits.

Note: we have to check whole of document for autosummary again.

tk0miya:
I noticed autosummary does not detect new submodules after a build once. Because it skip to process if .rst files are already generated. To create right document, we need to clear generated .rst files before building.

sphinx/ext/autosummary/generate.py
woutdenolf: Type annotation added. `get_members` had an unused `include_public` so I added it to `get_package_members`  as well. If we remove it we should do it for both.

sphinx/ext/autosummary/generate.py
woutdenolf: Refactored

sphinx/ext/autosummary/generate.py
woutdenolf: Analog to `all_functions`, `all_classes`, `all_methods`, etc... I suppose they are only used when you specify them in your template.

woutdenolf:
Proposal to deal with new and removed submodules and subpackages: in `generate_autosummary_docs` existing .rst files are overwritten when they correspond to a package.

Because this is not always what you want (you may have a user created .rst for a package already), I also propose to add the config parameter which enables/disables package recursion. Instead of a boolean I used an integer so you can additionally control the recursion depth:

- `autosummary_recursion_limit == 0`: disable recursion (default, old behavior)
- `autosummary_recursion_limit < 0`: unlimited recursion
- `autosummary_recursion_limit > 0` limited recursion depth

Caveat: generated .rst files are not deleted when rebuilding with a decreased recursion limit, which leads to warnings about rst files not being included in the toctree.

Alternatively one could somehow mark the .rst files generated by autosummary to distinguish them from user generated files, so you would know whether they can be deleted (in case of decreased recursion limit) and overwritten (in case of package recursion).

doc/usage/extensions/autosummary.rst
woutdenolf: Done

sphinx/ext/autosummary/generate.py
tk0miya: No, `include_public` for `get_members() has been used. So we can't remove it.
https://github.com/sphinx-doc/sphinx/blob/4628264551a2c2ef16213abdd6965c7baf65c536/sphinx/ext/autosummary/generate.py#L205-L206

sphinx/ext/autosummary/generate.py
tk0miya: Reasonable.

doc/usage/extensions/autosummary.rst
```diff
@@ -25,14 +25,19 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
    These files by default contain only the corresponding
    :mod:`sphinx.ext.autodoc` directive, but can be customized with templates.
 
+3. Optionally, the new :confval:`autosummary_recursion_limit` config value can be
```
tk0miya: The term "new" is not needed here.

doc/usage/extensions/autosummary.rst
```diff
@@ -25,14 +25,19 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
...
+   used to generate "stub" files recursively for packages. Warning: this will overwrite
+   existing files for packages.
```
tk0miya: `autosummary_generate` option always overwrite files. not only files for packages. Is this warning needed?

doc/usage/extensions/autosummary.rst
```diff
@@ -54,6 +59,27 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
 
        .. currentmodule:: sphinx.ext.autosummary
 
+   When :confval:`autosummary_recursion_limit != 0`, the :rst:dir:`autosummary`
```
tk0miya: The `confval` role takes the variable name as a content. And `!= 0` is not a variable name. So this marking up is not good. How about ```When :confval:`autosummary_recursion_limit` is not 0, ...``` instead?

doc/usage/extensions/autosummary.rst
```diff
@@ -54,6 +59,27 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
...
+   directive recursively generates stub ``.rst`` files for packages.
+
+   For example, ::
+
+       .. currentmodule:: sphinx
```
tk0miya: `sphinx.ext` ?

doc/usage/extensions/autosummary.rst
tk0miya: I fixed "new" in the list item above: https://github.com/sphinx-doc/sphinx/pull/6104

doc/usage/extensions/autosummary.rst
```diff
@@ -136,13 +162,25 @@ also use this new config value:
...
    Boolean indicating whether to scan all found documents for autosummary
-   directives, and to generate stub pages for each.
+   directives, and to generate stub pages for each. It is disabled by default.
 
    Can also be a list of documents for which stub pages should be generated.
 
    The new files will be placed in the directories specified in the
    ``:toctree:`` options of the directives.
 
+Packages can be explored recursively when generating stub pages.
```
tk0miya: I think this is a part of description for `autosummary_recursion_limit`. So this should be placed under the entry.

sphinx/__main__.py
```diff
@@ -12,4 +12,5 @@
 
 from sphinx.cmd.build import main
 
-sys.exit(main(sys.argv[1:]))
+if __name__ == "__main__":
+    sys.exit(main(sys.argv[1:]))
```
tk0miya: Is this change related with this PR?

sphinx/ext/autosummary/__init__.py
```diff
@@ -731,7 +734,7 @@ def setup(app):
     app.connect('doctree-read', process_autosummary_toc)
     app.connect('builder-inited', process_generate_options)
     app.add_config_value('autosummary_generate', [], True, [bool])
+    app.add_config_value('autosummary_recursion_limit', 0, True, [int])
     app.add_config_value('autosummary_mock_imports',
                          lambda config: config.autodoc_mock_imports, 'env')
-
```
tk0miya: why removing this line?

sphinx/ext/autosummary/generate.py
```diff
@@ -90,12 +91,19 @@ def _underline(title, line='='):
...
+def generate_autosummary_docs(sources,                 # type: List[str]
+                              output_dir=None,         # type: str
+                              suffix='.rst',           # type: str
+                              warn=_simple_warn,       # type: Callable
+                              info=_simple_info,       # type: Callable
+                              base_path=None,          # type: str
+                              builder=None,            # type: Builder
+                              template_dir=None,       # type: str
+                              imported_members=False,  # type: bool
+                              recursion_limit=0,       # type: int
```
tk0miya: Could you add this to the tail of argument list to keep compatibility? I'm afraid of changing the order of parameters.

tests/roots/test-ext-autosummary-package/conf.py
```diff
@@ -0,0 +1,10 @@
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath('.'))
+
+extensions = ['sphinx.ext.autosummary']
+autosummary_generate = True
+
+# The suffix of source filenames.
+source_suffix = '.rst'
```
tk0miya: Since 2.0, the default setting of `source_suffix` becomes `.rst`. So this setting is not needed.

sphinx/ext/autosummary/generate.py
```diff
@@ -151,11 +159,32 @@ def generate_autosummary_docs(sources, output_dir=None, suffix='.rst',
             warn('[autosummary] failed to import %r: %s' % (name, e))
             continue
 
+        # https://docs.python.org/3/reference/import.html#packages
+        # ... any module that contains a __path__ attribute is
+        # considered a package ...
+        ispackage = hasattr(obj, '__path__')
+        if ispackage:
+            depth = len(name.split('.')) - 1
+            recurse_package = depth < recursion_limit or recursion_limit < 0
```
tk0miya: It seems `recursion_limit` is compared with the depth of package from root. I think it is not called "recursion".

doc/usage/extensions/autosummary.rst
```diff
@@ -54,6 +59,27 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
 
        .. currentmodule:: sphinx.ext.autosummary
 
+   When :confval:`autosummary_recursion_limit != 0`, the :rst:dir:`autosummary`
+   directive recursively generates stub ``.rst`` files for packages.
+
+   For example, ::
```
tk0miya: Is this example related with above sentence? It seems `autosummary_recursion_limit` is not set on this document.

tk0miya (commented):
I reviewed partially (I'll review again later if i have time)

sphinx/ext/autosummary/generate.py
woutdenolf: Yes I should specify in the docs that this is recursion depth from root. At first I was thinking of limiting the `generate_autosummary_docs` call depth but I think this gives confusing results, depending on the situation. Let me know if you have an issue with the concept or just the name of the parameter. Not sure what name would make more sense `autosummary_depth_limit`, `autosummary_package_depth_limit`, ...

doc/usage/extensions/autosummary.rst
woutdenolf: done

doc/usage/extensions/autosummary.rst
woutdenolf: removed the warning

doc/usage/extensions/autosummary.rst
woutdenolf: paragraph removed altogether

doc/usage/extensions/autosummary.rst
woutdenolf: paragraph removed altogether

doc/usage/extensions/autosummary.rst
woutdenolf: removed

sphinx/ext/autosummary/__init__.py
woutdenolf: sorry, put it back

sphinx/ext/autosummary/generate.py
woutdenolf: done

tests/roots/test-ext-autosummary-package/conf.py
woutdenolf: done

sphinx/ext/autosummary/generate.py
woutdenolf: Called it `autosummary_depth_limit`. Docs mention it is the depth starting from root.

doc/usage/extensions/autosummary.rst
woutdenolf: paragraph removed altogether

shimizukawa:
My opinion is:

* +1 for recursive.
* -1 for limitation for recursive. I don't 
* -1 for separated templates of module and package. I think users don't care whether namespace is module or package.

woutdenolf:
Thanks for the feedback!

> * -1 for limitation for recursive.

I agree limiting recursion depth may not be that useful. I would personally use it to eliminate the test suites from my API docs, but that's just because all of my test suites are in subpackages at the same depth, while no other subpackage exists at this depth. Exclusion patterns may be a better way to do this though. Maybe for another PR if this one gets accepted.

> * -1 for separated templates of module and package. I think users don't care whether namespace is module or package.

I find module/package separation cleaner, but that's just personal taste.

If we can get a second opinion on these two things (@tk0miya?) I will make changes accordingly.

tk0miya:
>-1 for limitation for recursive.

I don't have opinion for this. But I only feel depth limitation *from specified level* might be useful (ex. limitation 2 levels from `sphinx.environment`), not from root. It would be better if exclusion is added instead.

>-1 for separated templates of module and package. I think users don't care whether namespace is module or package.

I don't understand which is better. From author side, modules and packages are different. But it would not be important for readers. IMO, the difference is not important for users of the library. For example, I misunderstood `os` is a package. But the difference does not harm me ever.
Is there any advantage to separate them?

woutdenolf:
I removed the recursion limit (`autosummary_recursive` is now a boolean) and removed the separation between modules and packages (all end up under the `modules` rubric).

sphinx/ext/autosummary/generate.py
```diff
@@ -168,11 +176,26 @@ def generate_autosummary_docs(sources, output_dir=None, suffix='.rst',
...
+        # Skip it if it exists and not a package.
         if os.path.isfile(fn):
-            continue
+            if ispackage and recursive:
+                # Overwrite the file because modules/subpackages
+                # could have been added/removed from the package.
+                # Warning: this file could have been created by the user
+                # in which case it is lost.
+                warn('[autosummary] overwriting docs of package %s' % (repr(name)))
+                os.remove(fn)
```
tk0miya: I will add an option for overwriting stub files in #6422.
So this special handling is not needed.

sphinx/ext/autosummary/generate.py
```diff
@@ -201,6 +224,21 @@ def get_members(obj, types, include_public=[], imported=True):
                           if x in include_public or not x.startswith('_')]
                 return public, items
 
+            def get_modules(obj, include_public=[]):
```
tk0miya: It seems `include_public` is never given. Why is this needed?

sphinx/ext/autosummary/generate.py
```diff
@@ -201,6 +224,21 @@ def get_members(obj, types, include_public=[], imported=True):
...
+                # type: (Any, List[str]) -> Tuple[List[str], List[str]]
+                items = []  # type: List[str]
+                for _, modname, ispkg in pkgutil.iter_modules(obj.__path__):
+                    fullname = name + '.' + modname
+                    try:
+                        import_by_name(fullname)
```
tk0miya: Why do you check the module is importable here? No check is needed here to me.

sphinx/ext/autosummary/templates/autosummary/module.rst
```diff
@@ -34,3 +34,15 @@
    {%- endfor %}
    {% endif %}
    {% endblock %}
+
+{% block modules %}
+{% if modules %}
+.. rubric:: modules
+
+.. autosummary::
+   :toctree: modules
```
tk0miya, line 43: Is output directory fixed?

tests/test_ext_autosummary.py
```diff
@@ -222,6 +224,115 @@ def test_autosummary_generate(app, status, warning):
...
+        root = root / dir
+        for name in files:
+            generated_paths.append(root / name)
+        for name in subdirs:
+            generated_paths.append(root / name)
+    assert set(generated_paths) == set(expected_paths)
+
+
+@pytest.mark.sphinx('dummy', testroot='ext-autosummary-recursive')
+def test_autosummary_recursive(make_app, app_params):
```
tk0miya: I feel this testcase is too complicated... How about making it simple?

I think main concern of this testcase is recursive generation. So it is fine to just check existence of stub files.
And it would be great if check root stub file contains `autosummary` definition for recursive call.
```
@pytest.mark.sphinx('dummy', testroot='ext-autosummary-recursive', confoverrides={'autosummary_recursive': True})
def test_autosummary_recursive_enabled(...):
    app.build()
    assert (app.srcdir / 'generated' / 'package.rst').exists()
    assert (app.srcdir / 'generated' / 'package.module.rst').exists()
    ...

    content = (app.srcdir / 'generated' / 'package.rst').text()
    assert '.. autosummary::\n  :toctree: module\n...' in content


@pytest.mark.sphinx('dummy', testroot='ext-autosummary-recursive', confoverrides={'autosummary_recursive': True})
def test_autosummary_recursive_disabled(...):
    app.build()
    assert (app.srcdir / 'generated' / 'package.rst').exists()
    assert not (app.srcdir / 'generated' / 'package.module.rst').exists()
    ...

```

tests/roots/test-ext-autosummary-recursive/package/package/package/package/module_importfail.py
```diff
@@ -0,0 +1,4 @@
+import sys
```
tk0miya: Why is such deeper module needed? I think it is enough to test sub-package.

doc/usage/extensions/autosummary.rst
```diff
@@ -136,13 +141,20 @@ also use these config values:
...
    Boolean indicating whether to scan all found documents for autosummary
-   directives, and to generate stub pages for each.
+   directives, and to generate stub pages for each. It is disabled by default.
 
    Can also be a list of documents for which stub pages should be generated.
 
    The new files will be placed in the directories specified in the
    ``:toctree:`` options of the directives.
 
+.. confval:: autosummary_recursive
```
tk0miya, line 158: It would be also good to add `:recursive:` option to `autosummary` directive. It allows to switch recursive-ness for each autosummary calls. What do you think?

sphinx/ext/autosummary/generate.py
```diff
@@ -126,12 +127,19 @@ def render(self, template_name, context):
...
+def generate_autosummary_docs(sources,                 # type: List[str]
+                              output_dir=None,         # type: str
+                              suffix='.rst',           # type: str
+                              warn=_simple_warn,       # type: Callable
+                              info=_simple_info,       # type: Callable
+                              base_path=None,          # type: str
+                              builder=None,            # type: Builder
+                              template_dir=None,       # type: str
+                              imported_members=False,  # type: bool
+                              recursive=False,         # type: bool
```
tk0miya: Reordering arguments is always breaking change. Could you add a new option to the tail please? I'm afraid that 3rd party application might give parameter as positional.

tk0miya (changes requested):

doc/usage/extensions/autosummary.rst
shimizukawa: I don't know what to do without discussing based on the use case.

* What kind of cases do you want to control recursion with common settings?
* What kind of cases do you want to control recursion separately for the autosummary directive?

sphinx/ext/autosummary/generate.py
woutdenolf: Same goes for the already existing `get_members`, I just stayed consistent.

sphinx/ext/autosummary/generate.py
woutdenolf: Removed it

sphinx/ext/autosummary/templates/autosummary/module.rst
woutdenolf: What else do you suggest? Some options one can think of:
 - specify nothing `:toctree:`: all files are generated in the same directory
 - use package name `:toctree: {{ name }}`: might be confusing (dir tree does not reflect package/subpackage structure)
 - inherit parent toctree `:toctree: {{ ??? }}`: Needs another context variable added by  `generate_autosummary_content`. Not sure how to retrieve the parent toctree name.

tests/test_ext_autosummary.py
woutdenolf: Done

tests/roots/test-ext-autosummary-recursive/package/package/package/package/module_importfail.py
woutdenolf: Done

sphinx/ext/autosummary/generate.py
woutdenolf: Done

sphinx/ext/autosummary/generate.py
woutdenolf: Merged with master.

doc/usage/extensions/autosummary.rst
woutdenolf: @tk0miya Would this require adding something like `sphinx/ext/autodoc/directive.py` to the autosummary extension? I'm not familiar with this aspect but I could try to figure it out.

@shimizukawa This would allow the user to exclude specific packages from being explored when `autosummary_recursive == True`. To work it requires `autosummary_overwrite == False` and the user adding for example `package.excludepackage.rst`. The `excludepackage` package itself would be in the docs though, so I'm not sure what the use case is for this.

woutdenolf:
Lets celebrate the one year anniversary of the PR ;-). Anything I can do? Or should I give up on this?

doc/usage/extensions/autosummary.rst
tk0miya: As a trial, I implemented :recursive: option. Please see https://github.com/tk0miya/sphinx/tree/6040_autosummary_recursive.

sphinx/ext/autosummary/templates/autosummary/module.rst
tk0miya: At I commented first, I thought inheriting parent toctree is better. But, now, I feel it seems no reasons to make a sub directory. Why `generated/modules/modules/modules/...` directories are needed? IMO, same directory is better to me.

tests/test_ext_autosummary.py
```diff
@@ -254,6 +255,64 @@ def test_autosummary_generate_overwrite2(app_params, make_app):
...
+    assert (generated / 'package.package.package.rst').exists()
+    content = (generated / 'package.package.package.rst').text()
+    assert 'package.package.package.module' in content
+    assert 'package.package.package.package' not in content
+
+    # Last sub-package has no sub-packages
+    generated /= toctree
+    assert (generated / 'package.package.package.module.rst').exists()
+    assert not (generated / 'package.package.package.package.rst').exists()
+    if toctree:
```
tk0miya, line 292: Why this if condition needed? It seems always true.

tests/test_ext_autosummary.py
```diff
@@ -254,6 +255,64 @@ def test_autosummary_generate_overwrite2(app_params, make_app):
...
+    assert 'package.package.module' in content
+    assert 'package.package.package' in content
+
+    # Recursively generate modules of sub-package
+    generated /= toctree
+    assert (generated / 'package.package.module.rst').exists()
+    assert (generated / 'package.package.package.rst').exists()
+    content = (generated / 'package.package.package.rst').text()
+    assert 'package.package.package.module' in content
+    assert 'package.package.package.package' not in content
```
tk0miya, line 286: These deeper packages might be needed for checking the behavior of recursive_limit. But they're no longer needed now.

sphinx/ext/autosummary/generate.py
```diff
@@ -151,6 +153,16 @@ def get_members(obj: Any, types: Set[str], include_public: List[str] = [],
                   if x in include_public or not x.startswith('_')]
         return public, items
 
+    def get_modules(obj: Any, include_public: List[str] = []) -> Tuple[List[str], List[str]]:
```
tk0miya, line 156: `include_public` is not used. It should be removed. Same option in `get_members()` is only used for `__init__()` method. No reason to add such option to here.

tk0miya (changes requested):
Sorry for very very very late response. Let's step forward again for 3.0 release. I believe this is one of key feature of the release!

tk0miya:
Now I'm merging this (see #7527). Thank you for your work!

PR #6048, opened 2019-02-09
Doc rework (development guide, part 1)

stephenfin:
This is combines two things:

- A rewrite of the existing tutorials to present a consistent style
- The addition of a new tutorial, based on a guide by @ofosos, which covers
  domains and cross-references


doc/development/tutorials/helloworld.rst
```diff
@@ -1,39 +1,54 @@
-Developing a "Hello world" directive
+Developing a "Hello world" extension
 ====================================
 
-The objective of this tutorial is to create a very basic extension that adds a new
-directive that outputs a paragraph containing `hello world`.
+The objective of this tutorial is to create a very basic extension that adds a
+new directive. This directive will output a paragraph containing `hello world`.
```
tk0miya: I noticed single backtick is not good for this case. How about replacing by double-quotes?

doc/development/tutorials/helloworld.rst
```diff
@@ -1,39 +1,54 @@
...
 
-Your extension file could be in any folder of your project. In our case,
-let's do the following:
+Overview
+--------
 
-#. Create an :file:`_ext` folder in :file:`source`.
+We want the extension to add the following to Sphinx:
+
+* A ``helloworld`` directive, that will simply output the text `hello world`.
```
tk0miya: This is also.

doc/development/tutorials/helloworld.rst
```diff
@@ -56,30 +72,33 @@ Open :file:`helloworld.py` and paste the following code in it:
...
+   class HelloWorld(Directive):
 
-.. seealso::
+       def run(self):
+           paragraph_node = nodes.paragraph(text='Hello World!')
+           return [paragraph_node]
 
-   :doc:`todo`
+This class extends the docutils_' ``Directive`` class. All extensions that
+create directives should extend this class.
```
tk0miya, line 76: I wonder we should introduce `SphinxDirective` class here instead of docutils'.
It provides helper methods for Sphinx framework. For example, user directives can access the config object via `self.config` attribute. I believe it is very useful for developers. 

On the other hand, the directive becomes only for Sphinx. It can't use with other docutils declarative tools.

doc/development/tutorials/helloworld.rst
```diff
@@ -89,74 +108,93 @@ The ``nodes.paragraph`` class creates a new paragraph node. A paragraph
...
-.. note::
+.. code-block:: python
+
+   def setup(app):
+       app.add_directive("helloworld", HelloWorld)
 
-   The first argument is the name of the directive itself as used in an rST file.
+The simplest thing you can do it call the :meth:`~Sphinx.add_directive`
+method, which is what we've done here. For this particular call, the first
+argument is the name of the directive itself as used in an rST file. In this
```
tk0miya: We usually use "reST" for abbreviation of reStructuredText.

doc/development/tutorials/todo.rst
```diff
@@ -217,12 +327,17 @@ The node structure that the directive returns looks like this::
         | ...                |
         +--------------------+
 
+.. rubric:: The event handlers
 
-The Event Handlers
-------------------
+Event handlers are one of Sphinx's most powerful features, providing a way to do
+hook into any part of the documentation process. There are many hooks available,
+as detailed in :doc:`/extdev/appapi`, and we're going to use a subset of them
```
tk0miya: I think using `:ref:` is better to here. `/extdev/appapi` is a page for Application APIs, not only for event handlers.

doc/development/tutorials/todo.rst
```diff
@@ -283,17 +398,149 @@ to be done::
...
+<build-phases>` and allows custom resolving to be done. The handler we have
+written for this event is a bit more involved.  If the ``todo_include_todos``
+config value (which we'll describe shortly) is false, all ``todo`` and
+``todolist`` nodes are removed from the documents. If not, ``todo`` nodes just
+stay where and how they are.  ``todolist`` nodes are replaced by a list of todo
+entries, complete with backlinks to the location where they come from.  The list
+items are composed of the nodes from the ``todo`` entry and docutils nodes
+created on the fly: a paragraph for each entry, containing text that gives the
+location, and a link (reference node containing an italic node) with the
+backreference.  The reference URI is built by ``app.builder.get_relative_uri``
```
tk0miya: We already have document for `get_relative_uri()`. So it's better to refer it here.

doc/development/tutorials/todo.rst
```diff
@@ -283,17 +398,149 @@ to be done::
 
            node.replace_self(content)
 
-It is a bit more involved.  If our new "todo_include_todos" config value is
-false, all todo and todolist nodes are removed from the documents.
+The :event:`doctree-resolved` event is emitted at the end of :ref:`phase 3
```
tk0miya, line 208: I feel it is better to use name of phases here instead. It is related with the event name.

doc/development/tutorials/todo.rst
```diff
@@ -283,17 +398,149 @@ to be done::
...
+  ``False`` (this also tells Sphinx that it is a boolean value).
+
+  If the third argument was ``'html'``, HTML documents would be full rebuild if the
+  config value changed its value.  This is needed for config values that
+  influence reading (build :ref:`phase 1 <build-phases>`).
+
+* :meth:`~Sphinx.add_node` adds a new *node class* to the build system.  It also
+  can specify visitor functions for each supported output format.  These visitor
+  functions are needed when the new nodes stay until :ref:`phase 4 <build-phases>`
+  -- since the ``todolist`` node is always replaced in :ref:`phase 3 <build-phases>`,
```
tk0miya: Here is also. Let's replace phases by their names.

doc/development/tutorials/todo.rst
```diff
@@ -283,17 +398,149 @@ to be done::
...
+For example:
+
+.. code-block:: python
+
+   import os
+   import sys
+
+   sys.path.append(os.path.abspath("./_ext"))
+
+   extensions = ['helloworld']
```
tk0miya: `todo`?

doc/development/tutorials/helloworld.rst
```diff
@@ -115,10 +99,10 @@ the ``text`` parameter.
...
 Sphinx.
 
-.. code-block:: python
-
-   def setup(app):
-       app.add_directive("helloworld", HelloWorld)
+.. literalinclude:: examples/helloworld.py
+   :language: python
+   :linenos:
+   :lines: 12-
```
tk0miya, line 106: How about using `:pyobject: setup` option instead? The line numbers are very fragile. So the option makes our document robust.

tk0miya: ah, it seems other inclusions contain multiple python objects. So `:pyobject:` would not help us in this case...

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+        signode += addnodes.desc_type(text='Recipe')
+        return sig
+
+    def add_target_and_index(self, name_cls, sig, signode):
+        signode['ids'].append('recipe' + '-' + sig)
+        if 'noindex' not in self.options:
+            name = '{}.{}.{}'.format('rcp', type(self).__name__, sig)
+            imap = self.env.domaindata['rcp']['obj2ingredient']
+            imap[name] = list(self.options.get('contains').split(' '))
+            objs = self.env.domaindata['rcp']['objects']
```
tk0miya: I think accessing internal data of domain directly is not good way. How about using domain methods like following:
```
recipes = self.env.get_domain('rcp')
recipes.add_recipe(...)
```

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+                         'Recipe',
+                         self.env.docname,
+                         'recipe' + '-' + sig,
+                         0))
+
+
+class IngredientIndex(Index):
+    """A custom directive that creates an ingredient matrix."""
+
+    name = 'ing'
```
tk0miya: Why are you using abbreviation here?

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+    localname = 'Ingredient Index'
+    shortname = 'Ingredient'
+
+    def __init__(self, *args, **kwargs):
+        super(IngredientIndex, self).__init__(*args, **kwargs)
+
+    def generate(self, docnames=None):
+        """Return entries for the index given by *name*.
```
tk0miya: Is this docstring needed? It would be better to move this to the document.

Note: `RecipeIndex.generate()` also has same docstring. It is duplicated and hard to read the example code.

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+        objs = {name: (dispname, typ, docname, anchor)
+                for name, dispname, typ, docname, anchor, prio
+                in self.domain.get_objects()}
+
+        imap = {}
+        ingr = self.domain.data['obj2ingredient']
+        for name, ingr in ingr.items():
+            for ig in ingr:
+                imap.setdefault(ig,[])
+                imap[ig].append(name)
```
tk0miya: It would be better to add API to domain class.

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+                docname, '', typ
+            ))
+        re = [(k, v) for k, v in sorted(content.items())]
+
+        return (re, True)
+
+
+class RecipeDomain(Domain):
+
+    name = 'rcp'
```
tk0miya: Why abbreviation too?

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+
+    StandardDomain.initial_data['labels']['recipeindex'] = (
+        'rcp-rcp', '', 'Recipe Index')
+    StandardDomain.initial_data['labels']['ingredientindex'] = (
+        'rcp-ing', '', 'Ingredient Index')
+
+    StandardDomain.initial_data['anonlabels']['recipeindex'] = (
+        'rcp-rcp', '')
+    StandardDomain.initial_data['anonlabels']['ingredientindex'] = (
+        'rcp-ing', '')
```
tk0miya: I feel touching `StandardDomain` is very hacky solution. So it must not be introduced as an example.

doc/development/tutorials/helloworld.rst
```diff
@@ -89,74 +108,93 @@ The ``nodes.paragraph`` class creates a new paragraph node. A paragraph
...
 
 This function is a requirement. We use it to plug our new directive into
 Sphinx.
-The simplest thing you can do it call the ``app.add_directive`` method.
 
-.. note::
+.. code-block:: python
+
+   def setup(app):
+       app.add_directive("helloworld", HelloWorld)
```
tk0miya: I'd forgotten to mention the return value of `setup()`.
It would be better to metadata of the extension.
https://www.sphinx-doc.org/en/master/extdev/index.html#extension-metadata

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+   .. _opensource.com: https://opensource.com/article/18/11/building-custom-workflows-sphinx
+
+
+Overview
+--------
+
+We want the extension to add the following to Sphinx:
+
+* A ``recipe`` :term:`directive`, containing some content describing the recipe
+  steps, along with a ``:contains:`` argument highlighting the main ingredients
```
tk0miya: `:contains:` is a directive option, not an argument. Arguments of directive are parameters following `::` mark:
```
.. directive-name::  this is arguments for directive
   :directive-option1: parameter for the option
   :directive-option2: parameter

   contents of directives.
   blah blah blah...
```

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+  of the recipe.
+
+* A ``reref`` :term:`role`, which provides a cross-reference to the recipe
+  itself.
```
tk0miya, line 26: I feel `ref` is better for such pupose. Users can use it via `:rcp:ref:`. (This is why I prefer to use non abbreviated domain name. Then we can use `:recipe:ref:` instead.)

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+#. Create an :file:`_ext` folder in :file:`source`
+#. Create a new Python file in the :file:`_ext` folder called :file:`recipe.py`
+
+Here is an example of the folder structure you might obtain:
+
+.. code-block:: text
+
+      └── source
+          ├── _ext
+          │   └── todo.py
```
tk0miya: `recipe.py`?

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+
+* A new directive called ``recipe``
+
+* New indexes to allow us to reference ingredient and recipes
+
+* A new domain called ``rcp``, which will contain the ``recipe`` directive and
+  ``reref`` role
+
+
+Prerequisites
```
tk0miya, line 41: I wonder this section is really needed for each example. It is too much duplicated. I understand verbosity helps newbies. But hyperlinks can also help them.

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+.. literalinclude:: examples/recipe.py
+   :language: python
+   :linenos:
+
+Let's look at each piece of this extension step-by-step to explain what's going
+on.
+
+.. rubric:: The directive class
+
+The first thing to examine is the ``RecipeNode`` directive:
```
tk0miya: `RecipeDirective` class?

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+for describing things like classes, functions, or, in our case, recipes. More
+specifically, ``handle_signature`` implements parsing the signature of the
+directive and passes on the object's name and type to its superclass, while
+``add_taget_and_index`` adds a target (to link to) and an entry to the index
+for this node.
+
+We also see that this directive defines ``has_content``, ``required_arguments``
+and ``option_spec``. Unlike the ``TodoDirective`` directive added in the
+:doc:`previous tutorial <todo>`, this directive takes a single argument, the
+recipe name, and an optional argument, ``contains``, in addition to the nested
```
tk0miya: `contains` is an option, not "argument".

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+.. todo:: Add brief overview of indices
+
+.. literalinclude:: examples/recipe.py
+   :language: python
+   :linenos:
+   :lines: 44-172
+
+Both ``IngredientIndex`` and ``RecipeIndex`` are derived from :class:`Index`.
+They implement custom logic to generate a tuple of values that define the
+index. Note that ``RecipeIndex`` is a degenerate index that has only one entry.
```
tk0miya: What "degenerate index" means?

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+:class:`sphinx.domains.Domain.resolve_xref` method. This method takes two
+arguments, ``typ`` and ``target``, which refer to the cross-reference type and
+its target name. We'll use ``target`` to resolve our destination from our
+domain's ``objects`` because we currently have only one type of node.
+
+Moving on, we can see that we've defined two items in ``intitial_data``:
+``objects`` and ``obj2ingredient``. These contain a list of all objects defined
+(i.e. all recipes) and a hash that maps a canonical ingredient name to the list
+of objects. The way we name objects is common across our extension and is
+defined in the ``get_full_qualified_name`` method. For each object created, the
```
tk0miya: Oh, `Domain.get_full_qualified_name()` is undocumented... we have to add document for it.

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+here, via the ``directives``, ``roles`` and ``indices`` attributes, rather than
+via calls later on in ``setup``. We can also note that we aren't actually
+defining a custom role and are instead reusing the
+:class:`sphinx.roles.XRefRole` role and defining the
+:class:`sphinx.domains.Domain.resolve_xref` method. This method takes two
+arguments, ``typ`` and ``target``, which refer to the cross-reference type and
+its target name. We'll use ``target`` to resolve our destination from our
+domain's ``objects`` because we currently have only one type of node.
+
+Moving on, we can see that we've defined two items in ``intitial_data``:
```
tk0miya: Note: `initial_data` will be copied to the `env.domaindata[domain_name]` as an initial data of the domain. And domain instance can access it via `self.data`.

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,239 @@
...
+        'recipe': RecipeDirective,
+    }
+
+    indices = {
+        RecipeIndex,
+        IngredientIndex
+    }
+
+    initial_data = {
+        'objects': [],  # object list
```
tk0miya: Nobody restricts the name of keys. So it would be easy to understand if we rename this to `recipes` key.

doc/development/tutorials/recipe.rst
```diff
@@ -0,0 +1,224 @@
...
+``objects`` and ``obj2ingredient``. These contain a list of all objects defined
+(i.e. all recipes) and a hash that maps a canonical ingredient name to the list
+of objects. The way we name objects is common across our extension and is
+defined in the ``get_full_qualified_name`` method. For each object created, the
+canonical name is ``rcp.<typename>.<objectname>``, where ``<typename>`` is the
```
tk0miya: I feel `<typename>` is not needed in this case. What about `recipe.<recipe-name>` instead? (`<recipe-name>` is same as object name).

The full qualified name is used to look up the objects from indices in intersphinx extension. So it is okay if enough unique. The intersphinx extension searches the object from the return value of `get_objects()`. So The form of the qualified should be used in the method.

tk0miya (changes requested):

doc/development/tutorials/helloworld.rst
stephenfin: Can do.

doc/development/tutorials/helloworld.rst
stephenfin: I don't understand the difference, to be honest. In fact, I thought we had removed that in favour of the docutils `Directive` class, but maybe I'm thinking of something else. Do you see any reason to do this here or in a later tutorial?

doc/development/tutorials/todo.rst
stephenfin: Can do

doc/development/tutorials/helloworld.rst
stephenfin: Yeah. I don't think this will we be too much hassle, in any case. If it would help, I could put a README in the folder saying that people should update the line numbers in the docs if they change the file.

doc/development/tutorials/examples/recipe.py
stephenfin: I just copy-pasted this from the blog (see the commit messages). This is probably a question for @ofosos

doc/development/tutorials/examples/recipe.py
stephenfin: Yes, I'll move it

doc/development/tutorials/examples/recipe.py
stephenfin: Could you give me an example? I'm not sure what you mean. Do you mean adding a `add_recipe` method?

doc/development/tutorials/examples/recipe.py
stephenfin: `rcp` is short for `recipe`, I guess. What would you suggest?

doc/development/tutorials/examples/recipe.py
stephenfin: OK. How should we do this?

doc/development/tutorials/helloworld.rst
stephenfin: Good point. I'll do this for the other ones too

doc/development/tutorials/recipe.rst
stephenfin: Ah, I was using the argparse terminology of "positional arguments" and "optional arguments". My mistake

doc/development/tutorials/recipe.rst
stephenfin: Good call. I'll change this.

doc/development/tutorials/recipe.rst
stephenfin: Whoops

doc/development/tutorials/recipe.rst
stephenfin: Yes, I wanted to avoid telling people to go back and forth between documents. Let's see if I can make it shorter though.

doc/development/tutorials/recipe.rst
stephenfin: Whoops

doc/development/tutorials/recipe.rst
stephenfin: I'm not sure. This was copy-pasted. @ofosos?

doc/development/tutorials/recipe.rst
stephenfin: +1

doc/development/tutorials/examples/recipe.py
stephenfin: +1

doc/development/tutorials/helloworld.rst
stephenfin: Added `README.rst`

doc/development/tutorials/examples/recipe.py
stephenfin: Used `recipe` as noted elsewhere.

doc/development/tutorials/examples/recipe.py
stephenfin: I guessed you meant `add_recipe` and did this

doc/development/tutorials/examples/recipe.py
stephenfin: Changed to `ingredient`

doc/development/tutorials/recipe.rst
ofosos: It's just an index with a single 'category', meaning there are not multiple index categories that then have entries.

ofosos: I'm not sure, probably you want the wording changed to say 'simple' or such. In retrospect that wording was not brilliant :)

doc/development/tutorials/helloworld.rst
tk0miya: Hmm? removed? Now we're using `SphinxDirective` class to make custom directives.
https://github.com/sphinx-doc/sphinx/blob/6bdac80b5fb75641f7cd6f4b8c47bf6241360e2c/sphinx/directives/__init__.py#L49

I believe `SphinxDirective` is useful to develop Sphinx extension. So it would be better to introduce it in this tutorial.

doc/development/tutorials/helloworld.rst
tk0miya: It seems not added. But I don't think the file will help us...

doc/development/tutorials/examples/recipe.py
tk0miya: Unfortunately, no alternative APis.

doc/development/tutorials/recipe.rst
tk0miya: Simply, I'm not good at English :-) It's okay if you feel natural.

doc/development/tutorials/recipe.rst
tk0miya: My comment above is wrong. It is already documented (but it says nothing...)
https://www.sphinx-doc.org/en/master/extdev/domainapi.html

doc/development/tutorials/examples/recipe.py
stephenfin: @ofosos What was the reason for doing this? I can remove it and things work as expected. Is it necessary?

stephenfin:
@tk0miya Think I've addressed all but one comment. The only thing let is using the `SphinxDirective` instead of `Directive`. Could you elaborate what this does so I know why I should use it? I'll include your explanation in the docs.

Also, I'm thinking it would be good to actually test these examples somehow. Have you suggestions for how to do this, @tk0miya?

doc/development/tutorials/helloworld.rst
stephenfin: As noted below, I'm still not sure why to use this. Could you elaborate on why it's useful. I can add a note saying we're extending from the docutils directive class but one should extend from the Sphinx one if they need access to `self.config` or the likes?

doc/development/tutorials/examples/recipe.py
ofosos: I think the reason for this was to have the indices show up in the TOC.

doc/development/tutorials/helloworld.rst
tk0miya: With docutils' Directive class, you need to write a complex code to access Sphinx objects (ex. `app`, `env`, `config`, `builder`).
For example, in case of a directive would like to access `myext_config` confval, you need to write following code:
```
class MyDirective(docutils.parsers.rst.Directive):
    def run(self):
        if self.state.document.settings.env.config.myext_config:
            ...
```

It is hard to write and understand for all developers.

With `SphinxDirective` class, you can access it simply with `self.config`:
```
class MyDirective(sphinx.util.docutils.SphinxDirective):
    def run(self):
        if self.config.myext_config:
            ...
```

If custom directives does not access to Sphinx objects, our `SphinxDirective` is not needed. So helloworld and todo extensions would not need to inherit it. But developers would need to know the way to access to Sphinx objects in real case. So I think here is good point to let them know about `SphinxDirective`.

tk0miya:
> Also, I'm thinking it would be good to actually test these examples somehow. Have you suggestions for how to do this, @tk0miya?

We've provided `sphinx.testing` to test Sphinx itself and extensions. So it would be nice if we'll test our examples. I suppose that would also be a kind of tutorial which developers would like to know.

doc/development/tutorials/examples/recipe.py
```diff
@@ -0,0 +1,171 @@
...
+
+class IngredientIndex(Index):
+    """A custom index that creates an ingredient matrix."""
+
+    name = 'ingredient'
+    localname = 'Ingredient Index'
+    shortname = 'Ingredient'
+
+    def __init__(self, *args, **kwargs):
+        super(IngredientIndex, self).__init__(*args, **kwargs)
```
tk0miya: It seems this method does not anything. How about removing?

tk0miya (approved):

doc/development/tutorials/examples/recipe.py
tk0miya: Okay, I understood. But there are no way to add cross-reference target from extensions. If you've still needed it, please file a new issue. I'll consider to add new API.

doc/development/tutorials/examples/recipe.py
stephenfin: Good call. Done.

stephenfin:
> > Also, I'm thinking it would be good to actually test these examples somehow. Have you suggestions for how to do this, @tk0miya?
> 
> We've provided `sphinx.testing` to test Sphinx itself and extensions. So it would be nice if we'll test our examples. I suppose that would also be a kind of tutorial which developers would like to know.

OK, let's do this in a follow-up.

doc/development/tutorials/helloworld.rst
stephenfin: I've added this to the todo case. Let me know if you think this is still unclear.

doc/development/tutorials/examples/recipe.py
stephenfin: In the interim, I've simply removed this code. We can think about this long term.

stephenfin:
@tk0miya anything else needed here?

tk0miya:
@stephenfin It seems flake8 emits warnings to example codes. Could you check them please?
https://travis-ci.org/sphinx-doc/sphinx/jobs/494899313#L232-L238

stephenfin:
> @stephenfin It seems flake8 emits warnings to example codes. Could you check them please?
> [travis-ci.org/sphinx-doc/sphinx/jobs/494899313#L232-L238](https://travis-ci.org/sphinx-doc/sphinx/jobs/494899313#L232-L238)

That's what I get for posting from my phone. I should have spotted that. Fixed now.

tk0miya:
Just merged!

PR #6062, opened 2019-02-11
docs: Use "Courier New" font to code blocks

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- In MacOS, monospace font is not used for code-blocks
- This adds "Courier New" to the candidates of the font settings.
- cssfontstack.com says it is used in 95% for MacOS users: https://www.cssfontstack.com/


tk0miya:
# Before:


# After:


(with MacOS Mojave + Chrome)

Issue #3620, opened 2017-04-12
Content Security Policy


rudolphfroger:
Subject: adjust themes to allow a sufficient strict Content-Security-Policy

### Problem
- Settings a sufficiently strict Content-Security-Policy raises many blocked actions

Also see: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
#### Procedure to reproduce the problem

Serving the generated HTML using a Content Security Policy header like below gives many exceptions:
```
Content-Security-Policy: "media-src 'self'; img-src 'self' data:; connect-src 'self'; child-src 'self'; style-src 'self'; default-src 'none'; font-src 'self'; script-src 'self'"
```
#### Expected results

I expected to be able to set a Content-Security-Policy header to protect the site against XSS and injection attacks.

### Environment info
- OS: any
- Python version: n.a.
- Sphinx version:any

shimizukawa:
@rudolphfroger I think you mentioned about HTTP header, however sphinx *IS NOT* a server application. Sphinx just generate html files. So I think there is nothing todo from sphinx side.

rudolphfroger:
I understand that Sphinx won't be able to set an HTTP header. At the moment the inline Javascript in the Sphinx themes prevent anyone from setting a reasonable Content Security Policy header without breaking Sphinx functionality. Moving all inline Javascript and CSS into the separate JS and CSS files would enable people to serve Sphinx in a better secured setup.

shimizukawa:
Thanks for your explanation. Indeed, current templates are not good.

ghost:
> I think you mentioned about HTTP header, however sphinx IS NOT a server application. Sphinx just generate html files. So I think there is nothing todo from sphinx side.

CSP can either be served via a header (i.e, via Apache) or put into a meta tag (i.e, in a Sphinx theme). The point of CSP is that it can be configured to disallow inline scripts/styles to prevent XSS/CSRF attacks so it *is* a Sphinx problem. 

Currently the RTD theme sets DOCUMENTATION_OPTIONS expected by the basic theme (doctools, searchtools, etc.) by doing this:

```
    <script type="text/javascript">
        var DOCUMENTATION_OPTIONS = {
             URL_ROOT:'{{ url_root }}',
             VERSION:'{{ release|e }}',
             COLLAPSE_INDEX:false,
             FILE_SUFFIX:'{{ '' if no_search_suffix else file_suffix }}',
[...]

Issue #322, opened 2015-01-02
Improve search responsiveness


shimizukawa:
When user is redirected to search page from other page and searchindex is not cached and is big (800k or more), the search may appear broken to user. This is described in http://bugs.python.org/issue7731

The patch below make search page load index only after form controls are animated and user is receiving feedback of what is going on. The proper version would, of course, display message "Caching search index 20k/800k 2.5&#37; complete..." and used received JSON directly instead of reevaluating it, but this requires more time to develop that I had to research this problem.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/322
- Originally reported by: anatoly techtonik
- Originally created at: 2010-01-18T12:39:01.654

shimizukawa:
_From anatoly techtonik on 2010-04-18 08:20:07+00:00_

Seems like I've lost the patch. Reuploading one attached to Python bugtracker.

shimizukawa:
_From Anonymous on 2010-05-25 04:59:33+00:00_

=== //// ====== === Your title here... === ===

shimizukawa:
_From Georg Brandl on 2010-05-30 13:50:22+00:00_

OK, I committed a similar patch as <<cset f3e6c80204950d14cc64cbe59b249970dfd9b34e>>. Thanks!

shimizukawa:
_From anatoly techtonik on 2010-06-03 08:21:34+00:00_

Do you really think that search functionality should be a part of a theme? I would say it is an essential feature to be present in Sphinx independently of themes.

Is it for 1.0 or for 0.6.x ?

shimizukawa:
_From Georg Brandl on 2010-06-03 08:34:06+00:00_

Don't worry about themes. The search functionality is part of the "basic" theme, which every other theme inherits.  The commits are made in the 1.0 branch.

shimizukawa:
_From anatoly techtonik on 2010-06-03 08:44:11+00:00_

Seems good to me then. =)

shimizukawa:
_From Georg Brandl on 2013-03-31 13:57:36+00:00_

Removing version: 0.6.3 (automated comment)

PR #6091, opened 2019-02-21
#3620: Defer searchindex.js rather than loading it via ajax

ghost:
Subject: Defer searchindex.js rather than loading it via ajax

### Feature or Bugfix
- Bugfix

### Purpose
- Content security policy is insecure when allowing inline scripts. So we need to remove them. However, `searchindex.js` is loaded via AJAX for reasons outlined in [#322](https://github.com/sphinx-doc/sphinx/issues/322). However, the `defer` attribute equally accomplishes this task. Further, the script should be cached just fine even when using this method. Also, I'm not entirely sure what the purpose of `searchindexloader` is, but I was able to test this locally in Chromium and Firefox without issues.

### Relates
- #3620

### DIFF
--- a/sphinx/themes/basic/search.html
+++ b/sphinx/themes/basic/search.html
@@ -14,12 +14,7 @@
     <script type="text/javascript" src="{{ pathto('_static/searchtools.js', 1) }}"></script>
 {%- endblock %}
 {% block extrahead %}
-  <script type="text/javascript">
-    jQuery(function() { Search.loadIndex("{{ pathto('searchindex.js', 1) }}"); });
-  </script>
-  {# this is used when loading the search index using $.ajax fails,
-     such as on Chrome for documents on localhost #}
-  <script type="text/javascript" id="searchindexloader"></script>
+  <script type="text/javascript" src="{{ pathto('searchindex.js', 1) }}" defer></script>
   {{ super() }}
 {% endblock %}
 {% block body %}
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -75,16 +75,6 @@ var Search = {
       }
   },
 
-  loadIndex : function(url) {
-    $.ajax({type: "GET", url: url, data: null,
-            dataType: "script", cache: true,
-            complete: function(jqxhr, textstatus) {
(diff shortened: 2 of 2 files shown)


jumarko:
I quickly looked at it and it seems there is at least [one other problematic script in search.html](https://github.com/sphinx-doc/sphinx/pull/6091/files#diff-453969dcf47b1378674f807c54aca249R23).

ghost:
@jumarko Thanks. However, that shouldn't be a problem since it gets rendered into a `<noscript>` block in the end.

tk0miya (changes requested):
LGTM. But it would be better to merge this to `2.0` branch. Could you change the target branch of this please?
Note: It seems Opera does not support `defer` keyword. So we have to check Opera works fine with this.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

ghost:
According to [caniuse](https://caniuse.com/script-defer) Opera supports it. Maybe the MDN page is outdated? Someone with an Opera browser can test this [script](https://raw.githubusercontent.com/phiggins42/has.js/master/detect/script.js#script-defer).

tk0miya (approved):

tk0miya:
Thank you for quick work. LGTM!

>According to [caniuse](https://caniuse.com/script-defer) Opera supports it.

Sounds good :-)

tk0miya:
@TimKam do you have any comment for this?

TimKam:
I don't see any issue. According to the [entry at caniuse.com](https://caniuse.com/#feat=script-defer) ``defer`` works fine in Opera, but there might be issues with Opera Mini. Let's test this before we merge.

TimKam (commented):
I don't see any issues, but think we make sure there are no browser compatibility issues. According to the [entry at caniuse.com](https://caniuse.com/#feat=script-defer) `defer` works fine in Opera, but there might be issues with Opera Mini. Let's test this before we merge.

TimKam:
@remyabel @tk0miya I just tested in the latest versions of Opera and Opera mini and it looks good :-)

tk0miya:
@TimKam thank you for confirming!

@remyabel Thank you for your contribution :-)

PR #6104, opened 2019-02-26
docs: Remove a term "new" from config values

tk0miya:
### Feature or Bugfix
- docs

### Purpose
- I feel a term "new" is not appropriate in these case.
- I guess this term was used like "additional".


doc/usage/extensions/autosummary.rst
```diff
@@ -131,7 +131,7 @@ Generating stub pages automatically
 -----------------------------------
 
 If you do not want to create stub pages with :program:`sphinx-autogen`, you can
-also use this new config value:
+also use this config value:
```
tk0miya: `autosummary` provides multiple config values. so this should be plural form.

PR #6109, opened 2019-02-27
refactor: test_autodoc

tk0miya:
### Feature or Bugfix
- Refactoring
- Merge test_autodoc.py and test_ext_autodoc.py


tk0miya:
>* Merge test_autodoc.py and test_ext_autodoc.py

We need to refactor `test_autodoc.py` more and more...
So I'm withdrawing this change partially.

PR #6129, opened 2019-03-03
Refactor util.nodes

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Rename find_source_node() to get_node_source() (to be more appropriate name)
- Add copy_source_info() (to copy source_info simple)
- Add a helper method ``SphinxDirective.set_source_info()``

### DIFF
--- a/sphinx/util/nodes.py
+++ b/sphinx/util/nodes.py
@@ -476,6 +476,12 @@ def set_role_source_info(inliner, lineno, node):
     node.source, node.line = inliner.reporter.get_source_and_line(lineno)  # type: ignore
 
 
+def copy_source_info(src, dst):
+    # type: (nodes.Element, nodes.Element) -> None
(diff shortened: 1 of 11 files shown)


sphinx/util/nodes.py
```diff
@@ -458,6 +467,8 @@ def make_refnode(builder, fromdocname, todocname, targetid, child, title=None):
 
 def set_source_info(directive, node):
     # type: (Any, nodes.Node) -> None
+    warnings.warn('set_source_info() is deprecated.',
+                  RemovedInSphinx40Warning)
```
tk0miya: IMO, `set_source_info()` function will be replaced by `SphinxDirective.set_source_info()`. So I'd like to deprecate this in the future because it is no longer needed.

tk0miya: Finally, I reverted this.

PR #6132, opened 2019-03-03
Closes #2155: Support ``code`` directive

### LINKED ISSUE #2155: Use reStructuredText's `code` directive
Since [docutils 0.9](http://docutils.sourceforge.net/RELEASE-NOTES.html#release-0-9-2012-05-02), reStructuredText supports [`code` directive](http://docutils.sourceforge.net/docs/ref/rst/directives.html#code), but Sphinx, unfortunately, can't handle this. By design, `code` directive supports syntax highlighting, but Sphinx ignores this due to `visit_literal_block` reimplementation in [custom `HTMLTranslator`](https://github.com/sphinx-doc/sphinx/blob/0dd6ce3063ddb2f8b7f084e923ecc7f4ab4e9423/sphinx/writers/html.py#L307-L335).

So basically:
- Sphinx handles `code` directive as literal block.
- There's no syntax highlighting in produced block (it's supposed to be generated by docutils, and it does, but as children nodes) due to HTMLTranslator limitations.

Well, I see two possible ways to solve this issue:
1. Use Sphinx's `CodeBlock` implementation for `code` directive:
   
   ``` python
   directives.register_directive('code', CodeBlock)
   ```
   
   Pros:
   - easy to implement
   - features/options similar `code-block` / `sourcecode`
   
   Cons:
   - doesn't compatible with docutils implementation
2. Use docutils' (native) implementation for `code` directive.
   
   Pros:
   - backward compatible, since standard (native) implementation is used
   
   Cons:
   - requires changes in many places, since we need:
     - distinguish Sphinx's code blocks from docutils' one on HTMLTranslator level (could be solved by injecting some marker from Sphinx's implementation)
     - use default `visit_literal_block` implementation for non-Sphinx code blocks
     - pass `short` Pygments CSS notation to docutils (`long` is default)
   - user will lack of Sphinx-specific `code-block` options

Folks, could you please share your opinion on this?

@birkenfeld It'd be nice to see your input.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #2155 
- It would be better to integrate `code` directive with `code-block` directive. But we need to have deeper discussion about it. So this only resolve that `code` directive has been not working in Sphinx. Let's discuss about integration in another topic.
- Note: `sourcecode` directive still has been incompatible with docutils'. In Sphinx, it is an alias of `code-block` directive (since its birth, f2ff0419bd). On the other hand, in docutils, it is an alias of `code` directive.


tk0miya:
I did not update our document. I suspect two "code block" directives will confuse users. This change is only needed for compatibility with standard reStructuredText, I believe.

tk0miya:
I'm merging this now. Please let me know if you have comment.

PR #6133, opened 2019-03-03
Clean up import for annotations

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- New flake8 command supports py2 style type annotations. So `NOQA` comment is not needed for them.
- This removes the comment all, and also removes unused import

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -40,7 +40,7 @@ except ImportError:
 
 if False:
     # For type annotation
-    from typing import Any, Callable, Dict, Iterable, List, Sequence, Set, Tuple, Type, Union  # NOQA
+    from typing import Any, Dict, Iterable, List, Sequence, Set, Tuple, Type, Union  # NOQA
     from sphinx.application import Sphinx  # NOQA
     from sphinx.config import Config  # NOQA
     from sphinx.environment import BuildEnvironment  # NOQA
--- a/sphinx/builders/_epub_base.py
+++ b/sphinx/builders/_epub_base.py
@@ -36,7 +36,6 @@ except ImportError:
 if False:
     # For type annotation
     from typing import Any, Dict, List, Set, Tuple  # NOQA
-    from sphinx.application import Sphinx  # NOQA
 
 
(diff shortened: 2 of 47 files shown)


stephenfin:
Have you a link to the flake8 change?

stephenfin (changes requested):
Do we need the `if False:` still? In fact, why was that ever needed? Also, we should set a minimum version for `flake8` in `tox.ini`

tk0miya:
It was added pyflakes-2.1.0. And flake8-3.7 depends on it.

>Mark names used in PEP 484 # type: ... comments as used
>https://github.com/PyCQA/pyflakes/blob/master/NEWS.rst

>Pyflakes has been updated to >= 2.1.0, < 2.2.0 (See also GitLab!283, GitLab!285)
>https://github.com/PyCQA/flake8/blob/master/docs/source/release-notes/3.7.0.rst

tk0miya:
>Do we need the if False: still? In fact, why was that ever needed? Also, we should set a minimum version for flake8 in tox.ini

It is needed for avoiding circular import. It is same as `typing.TYPE_CHECKING` (Added since py3.5.2).
https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

Some of the import would be able to move to top level. But we need to be careful to do that. So I'd like to do that in another PR.

stephenfin:
> It was added pyflakes-2.1.0. And flake8-3.7 depends on it.

OK, could you update this so and then I'm happy :+1: 

https://github.com/sphinx-doc/sphinx/blob/b58d0e88ce9cb477a87bf112adde8075535c8690/setup.py#L46

tk0miya:
I'd not like to upgrade dependency. So I stopped to remove `NOQA` comment. This PR only removes unused imports.

tk0miya:
I prefer to migrate our annotations to py3 style. I believe it does not requires new flake8.

stephenfin (approved):

stephenfin:
OK, then I'm happy with this as-is

tk0miya:
Thank you for reviewing!

Issue #6134, opened 2019-03-04
UnboundLocalError when building changes from module docs


bz2:
Building changes fails if a `::module` directive is included in an rst source file.

**To Reproduce**
For instance, currently trying to build the cpython master changes fails:
```
$ git clone git@github.com:python/cpython.git
$ cd cpython/Doc
$ make venv
$ make changes
```

**Error log**
```
# Python version: 3.6.7 (CPython)
# Docutils version: 0.14 
# Jinja2 version: 2.10
# Last messages:
#   pickling environment...
#   
#   done
#   
#   checking consistency...
#   
#   done
#   
#   writing summary file...
#   
# Loaded extensions:
#   sphinx.ext.mathjax (1.8.4) from /home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/sphinx/ext/mathjax.py
#   alabaster (0.7.12) from /home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/alabaster/__init__.py
#   sphinx.ext.coverage (1.8.4) from /home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/sphinx/ext/coverage.py
#   sphinx.ext.doctest (1.8.4) from /home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/sphinx/ext/doctest.py
#   pyspecific (1.0) from /home/martin/branches/cpython/Doc/tools/extensions/pyspecific.py
#   c_annotations (1.0) from /home/martin/branches/cpython/Doc/tools/extensions/c_annotations.py
#   escape4chm (1.0) from /home/martin/branches/cpython/Doc/tools/extensions/escape4chm.py
Traceback (most recent call last):
  File "/home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/sphinx/cmd/build.py", line 304, in build_main
    app.build(args.force_all, filenames)
  File "/home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/sphinx/application.py", line 341, in build
    self.builder.build_update()
  File "/home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 342, in build_update
    self.build(['__all__'], to_build)
  File "/home/martin/branches/cpython/Doc/venv/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 412, in build
    self.write(docnames, list(updated_docnames), method)
[...]

PR #6135, opened 2019-03-04
Fix UnboundLocalError when building changes (1.8)

bz2:
Split testing of changes builder to its own test file and root. Improve coverage a little, and add case that fails if a module directive is included in the rst source.

Correctly handle changesets with a module declared.

~~Ignore coverage annotations files in git.~~

Fixes https://github.com/sphinx-doc/sphinx/issues/6134 for the 1.8 branch.

Same change basically applies for master, can propose that separately or whatever the process is for landing on multiple branches.

### DIFF
--- a/tests/test_build_changes.py
+++ b/tests/test_build_changes.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+"""
+    test_build_changes
+    ~~~~~~~~~~~~~~~~~~
+
+    Test the ChangesBuilder class.
+
+    :copyright: Copyright 2007-2019 by the Sphinx team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+import pytest
+
+
+@pytest.mark.sphinx('changes', testroot='changes')
+def test_build(app):
+    app.build()
+
+    # TODO: Use better checking of html content
+    htmltext = (app.outdir / 'changes.html').text()
+    assert 'New in version 0.6: Some funny stuff.' in htmltext
+    assert 'Changed in version 0.6: Even more funny stuff.' in htmltext
+    assert 'Deprecated since version 0.6: Boring stuff.' in htmltext
+
+    path_html = (
+        '<b>Path</b>: <i>deprecated:</i> Deprecated since version 0.6:'
+        ' So, that was a bad idea it turns out.')
+    assert path_html in htmltext
+
+    malloc_html = (
+        '<b>Test_Malloc</b>: <i>changed:</i> Changed in version 0.6:'
+        ' Can now be replaced with a different allocator.</a>')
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 *.egg
 *.so
 *.swp
+*,cover
```
tk0miya: I feel this is not needed.

tests/test_build_changes.py
```diff
@@ -0,0 +1,43 @@
...
+    assert path_html in htmltext
+
+    malloc_html = (
+        '<b>Test_Malloc</b>: <i>changed:</i> Changed in version 0.6:'
+        ' Can now be replaced with a different allocator.</a>')
+    assert malloc_html in htmltext
+
+
+@pytest.mark.sphinx(
+    'changes', testroot='changes', srcdir='changes-none',
```
tk0miya, line 37: I think `srcdir` is not needed in this case. is this needed?

tk0miya (changes requested):
LGTM with nits

.gitignore
bz2: I ran:
```sh
tox -e coverage
coverage annotate
```
Which gave me a pile of these mixed in with my changes, but I guess maybe it's more useful to be able to `git clean -f` without `-x`.

tests/test_build_changes.py
bz2: Without this the test as written fails as the build artefacts persist from the last test. This was annoying during testing, but maybe just checking status and not caring about the build output is fine?

.gitignore
```diff
@@ -2,7 +2,6 @@
 *.egg
 *.so
 *.swp
-*,cover
```
tk0miya: Oh, sorry. I don't know about "coverage annotate".

tests/test_build_changes.py
```diff
@@ -34,10 +34,10 @@ def test_build(app):
 
 
 @pytest.mark.sphinx(
-    'changes', testroot='changes', srcdir='changes-none',
```
tk0miya: Thank you for explanation. Indeed, this is needed.

bz2:
@tk0miya I think this can land in current state? Would be good to improve the testing later but that can be a separate change.

tk0miya:
@bz2 I think the `srcdir` parameter for `test_no_changes` is necessary. So I'd like to revert it. Then I'll merge this ASAP.

bz2:
@tk0miya Amended.

Would you like me to also put up a branch for master, or is your process merging forwards from the 1.8 branch anyway?

tk0miya (approved):

tk0miya:
Thank you for updating! I'm merging this into 1.8 branch. And it will be merged into master branch later.

Issue #3743, opened 2017-05-12
Consider switching py:function and py:method to supporting different function types using options instead of directives?


njsmith:
One source of frustration I've run into when using Sphinx to document Python projects is this: out-of-the-box, Sphinx's `py` domain has directives for `function`, `method`, `staticmethod`, `classmethod`, `decorator`, and `decoratormethod`. But this leaves a bunch of common Python types that it doesn't support: async functions/methods, abstract functions/methods, context managers, async context managers, generators, async generators, and combinations of the above.

This is particularly annoying for new projects using async/await, since this is a critical part of function signatures in the language now, and Sphinx has no standard way to handle it. But what makes it extra tricky is that the async/not-async distinction is orthogonal to other distinctions like regular method/staticmethod/classmethod. Recently I found myself trying to document an abstract async classmethod, which drove home the limitations of the current system :-).

My solution was to write [sphinxcontrib-trio](https://sphinxcontrib-trio.readthedocs.io), which is an extension that – instead of trying to add a dozen+ new directives for every possible method type – replaces the standard `py:function` and `py:method` with new versions that take *options* describing the different attributes of the function/method. So that abstract async classmethod is just:
```rst
.. method::
   :abstractmethod:
   :async:
   :classmethod:
```

Doing this via an extension is convenient because it makes the functionality available immediately. But as far as I can tell, this approach is a 100% backwards compatible extension that makes the `py` domain much more usable for modern Python projects. So I'm wondering: is there any interest in building this functionality into Sphinx itself?

Rapptz:
There is e.g. [`sphinxcontrib-asyncio`](http://pythonhosted.org/sphinxcontrib-asyncio/) which does something similar. One problem I've encountered when using it is that it doesn't work very well with Sphinx's autodoc module but I suspect with proper support in the module it'd be a non-issue.

njsmith:
Re: sphinxcontrib-asyncio: [I'm aware](https://sphinxcontrib-trio.readthedocs.io/en/latest/#acknowledgements) :-). (And I don't know what problems you had with sphinxcontrib-asyncio, but as far as I know sphinxcontrib-trio integrates nicely with autodoc; if you run into any problems then I'd certainly like to know.)

tk0miya:
That sounds good. Indeed, the decorations of functions, methods and so on is orthogonal. I feel your approach is better than current way.
I don't know what way is best solution, but +1 for improving. We need many helps for improving and maintaining.

[...]

PR #6138, opened 2019-03-06
Detect Abstract methods and turn staticmethod, classmethod into options

### LINKED ISSUE #744: Mark abstract methods/properties as such
It would be useful if autodoc produced special markup for abstract methods and such, as defined by the abc module. Possibly, the "raw" directives in the Python domain should then also support marking something as abstract.

The "obvious" way to mark this up would be to prepend "abstract" in the same way as "class" is prepended to class definitions. This, I recall, is effectively what the sphinx extension for zope.interface does with interfaces. An argument against it might be that "abstract" is neither a keyword nor the way something is marked abstract in Python.

Another idea might be to make it look decorator-like and prepend a line like `[abstract](https://bitbucket.org/abstract)` or somesuch.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/744
- Originally reported by: Dag Odenhall
- Originally created at: 2011-08-04T01:37:19.823

lukasjuhrich:
Subject: Fix #744

### Feature or Bugfix
- Feature

### Purpose
Implement detection of abstract methods and similar.

### Detail
- Abstract {methods,staticmethods,classmethods,properties} are detected by autodoc.
- `PyClassmember`s have an additional `ReST`-Option `:abstract:` which renders as prefixing `abstract` where e.g. the `classmethod` prefix would appear.
- Tests have been added for the inspection function and the desired `autodoc` outcome.

- The actual output has been tested on a [proof-of-concept project of mine](https://github.com/lukasjuhrich/sphinx-abc-repro) as follows:
  1. `mkvirtualenv sphinx -p $(which python3.7) -a .`
  1. `pip install -e $your-sphinx-dev-path`
  1. `make clean html`
  1. Observe the following output:

example _(`web.archive.org`ed for your convenience after the 14day timeout)_

### Relates
- #744


lukasjuhrich:
This draft PR still misses documentation, which I'm trying to add according to the contributor's guide.
Please let me know if this pull request still misses something else, as I've never contributed to sphinx before.

lukasjuhrich:
I'm uncertain to which extent one should add documentation in `autodoc.rst`.

lukasjuhrich:
Can I pay for the missing code coverage with 0.01% of my soul? ;-)

sphinx/ext/autodoc/__init__.py
```diff
@@ -418,9 +419,12 @@ def format_signature(self):
         else:
             return ''
 
-    def add_directive_header(self, sig):
-        # type: (str) -> None
+    def add_directive_header(self, sig, additional_options=None):
```
tk0miya: How about renaming `additional_options` to `options` simply?

sphinx/ext/autodoc/__init__.py
```diff
@@ -418,9 +419,12 @@ def format_signature(self):
...
+        # type: (str, list) -> None
```
tk0miya: `List[str]`?

sphinx/ext/autodoc/__init__.py
tk0miya: In addition, how about giving options via keyword argument?
I think `add_directive_header(sig, abstract=True)` is much better. Passing options via both argument and `self.option` is a bit hard to understand to me.

sphinx/ext/autodoc/__init__.py
```diff
@@ -673,22 +685,26 @@ def keyfunc(entry):
         for documenter, isattr in memberdocumenters:
             documenter.generate(
                 all_members=True, real_modname=self.real_modname,
-                check_module=members_check_module and not isattr)
+                check_module=members_check_module and not isattr,
+                additional_options=['abstract'])
```
tk0miya: Why giving `abstract` option always?

sphinx/ext/autodoc/__init__.py
```diff
@@ -1369,16 +1390,20 @@ def import_object(self):
         else:
             # if it's not a data descriptor
             self._datadescriptor = False
+
+        if isabstract(self.object):
+            self.options['abstract'] = True
```
tk0miya, line 1407: Why not in `MethodDocumenter`?

tk0miya (changes requested):
Sorry for late reviewing. I just post some comments. Could you check them please?

tk0miya:
>Can I pay for the missing code coverage with 0.01% of my soul? ;-)

Personally, I don't mind about coverage. So you also don't need to take care it :-)

tk0miya:
Note: [sphinxcontrib-trio](https://sphinxcontrib-trio.readthedocs.io/en/latest/) has been used to `:abstractmethod:` option for such purpose. I feel either is okay.

sphinx/ext/autodoc/__init__.py
lukasjuhrich: Sorry, I didn't pay much mind about the best place, so I put it in the most generic class. Judging by the type hierarchy, I think the `ClassLevelDocumenter` would be the right level (`MethodDocumenter` doesn't sound like it covers properties as well)

lukasjuhrich: Let me look over this a bit, I misjudged some things and need to get used to the codebase again.
I'll get back to you shortly.

lukasjuhrich: This was the wrong place, I now did this in the `MethodDocumenter` and `AttributeDocumenter`, respectively. :-)

sphinx/ext/autodoc/__init__.py
lukasjuhrich: This was wrong or at least misleading. `additional_options` should have better been named `revealed_options`, and this only instructs the `Documenter` to include the `abstract` flag in the output *if it has been set in `self.options`*.

When you look at the current diff, this is being done in the `_revealed_header_options` property of `Documenter`, which is overridden in the `ClassLevelDocumenter`.

sphinx/ext/autodoc/__init__.py
lukasjuhrich: Rendered obsolete due recent changes in the diff.

sphinx/ext/autodoc/__init__.py
lukasjuhrich: As mentioned below, that argument should have better been named `revealed_options`.
I think I've come up with a way better implementation of what I've wanted to achieve. Could you check the new diff and see if it is understandable to you?

The main advantage of only telling `Documenter.add_directive_header` which parts of the `options` dict to be included in the ReST ouput is that extending becomes easier. For instance, to add an `:async:` flag, you don't need to introduce a new argument to `add_directive_header`, you can instead just set the `async` key in the options dict and add it to the `_revealed_header_options` whitelist. Changing a signature for that would be more effort for something that doesn't really change functionality.

lukasjuhrich:
@tk0miya Thanks for your review and sorry for not reacting earlier.

I've found a significantly easier method to include options in the directive header, it would be great if you could check out the new diff and give feedback on whether it is more readable than my first attempt.
Also, I've cleaned the commit history up a bit by rebasing, so the log be easier to digest.

sphinx/ext/autodoc/__init__.py
lukasjuhrich: Well, it *is* in the `MethodDocumenter`, but here we are in the `AttributeDocumenter`. This is responsible for documenting abstract `@property`s: If you comment this out, the test case fails precisely at the point where it's trying to verify the `:abstract:` flag on `AbstractBase.value`, which is defined as an abstract property using `@abstractmethod` in the usual way.

sphinx/ext/autodoc/__init__.py
lukasjuhrich: The relevant change is in https://github.com/sphinx-doc/sphinx/pull/6138/commits/9c58354de8c76ef1b3587a6caaf1b7e248e87354.

lukasjuhrich:
Left to do:

- [x] Add alias for `:abstractmethod:`
- [x] Add tests that `:abstract:` and `:abstractmethod:` both give the correct prefix when built
- [x] Add option aliases for `classmethod::`, `staticmethod::` (#3743)
- [ ] Fix the index text
- [x] Add tests for the index text

lukasjuhrich:
@tk0miya while rebasing I realized that the tests have been adapted and I can verify things a lot easier. I will get back to you once I extended & rewrote the tests.

lukasjuhrich:
@tk0miya Is the inclusion of `abstract` in the index text necessary in your opinion? While it was easy to add `abstract` to the prefix, the index is way harder to build due to its l10n – checking for `'abstract' in self.options` is doubling the number of cases we have to consider.

If we decide to include this, a similar case would probably be made for `async`, which would further double the number of cases.

lukasjuhrich:
Since this was thematically fitting, I included a fix for #3743 by implementing `staticmethod::` and `classmethod::` as aliases for options.  The modified behaviour sums up into:

- `:abstractmethod:` is an alias for `:abstract:`
- `:staticmethod:` sets the flag `:static:` and resets the objtype to `:method:`
- `:classmethod:` sets the flag `:classmethod:` and resets the objtype to `:method:`

Sorry for the unexpected increased review effort.

tk0miya:
Finally, I merged #6365 instead. Thank you for your work!

Issue #6149, opened 2019-03-08
"Use of \@icentercr doesn't match its definition." error building latexpdf


dpage:
I'm trying to build a PDF version of the docs for a project I work on using the latexpdf target - the doc source is at https://git.postgresql.org/gitweb/?p=pgadmin4.git;a=tree;f=docs/en_US for anyone that may be interested. On both macOS and Fedora I get the following error when trying to build the latexpdf target.

See discussion and workaround provided by Jean-François at https://groups.google.com/forum/#!topic/sphinx-users/ZlQ95_QikXo.

Full output of build:

```
(pgadmin4) dpage@hal:~/git/pgadmin4$ sphinx-build --version
sphinx-build 1.8.4
(pgadmin4) dpage@hal:~/git/pgadmin4$ python --version
Python 3.7.2
(pgadmin4) dpage@hal:~/git/pgadmin4$ make docs-pdf
LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 /Applications/Xcode.app/Contents/Developer/usr/bin/make -C docs/en_US -f Makefile.sphinx latexpdf
Generating code-snippet.rst for some of the important classes...
sphinx-build -W -b latex -d _build/doctrees   . _build/latex
Running Sphinx v1.8.4
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [latex]: all documents
updating environment: [] 0 added, 3 changed, 0 removed
reading sources... [100%] release_notes_4_4                                                                                                                                                                                                     
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
[...]

PR #6150, opened 2019-03-08
Fix 'Use of \@icentercr doesn't match its definition.' LaTeX build crash

jfbu:
Closes: #6149

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -6,7 +6,7 @@
 %
 
 \NeedsTeXFormat{LaTeX2e}[1995/12/01]
-\ProvidesPackage{sphinx}[2019/01/12 v1.8.4 LaTeX package (Sphinx markup)]
+\ProvidesPackage{sphinx}[2019/03/08 v1.8.5 LaTeX package (Sphinx markup)]
 
 % provides \ltx@ifundefined
 % (many packages load ltxcmds: graphicx does for pdftex and lualatex but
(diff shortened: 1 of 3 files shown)


sphinx/texinputs/sphinx.sty
```diff
@@ -1823,4 +1823,30 @@
...
+      % incomprehensible low level LaTeX errors will not be raised
+      % This behaves ok in section titles : the \index, \glossary, \label
+      % will swallow their arguments as usual and not end up in toc file.
+      \let\index\relax
+      \let\glossary\relax
+      \let\label\relax
+      \afterassignment\myendgroup
+      \edef}%
+}%
+% cf source2.pdf File O: ltfinal.dtx, \MakeUppercase definition
```
jfbu, line 1844: this should be ``source2e.pdf``. I will fix later if merging.

tk0miya:
Is it better to handle this in LaTeX world? We can also do it in python world with moving the index node to after title node.

tk0miya (approved):

jfbu:
> Is it better to handle this in LaTeX world? We can also do it in python world with moving the index node to after title node.

I will not say it is better in LaTeX world, it was faster for me to work this out :). And in the LaTeX world one can consider various options:

- do it the way of this PR. In my opinion LaTeX `\MakeUppercase` is buggy and this PR does a hack to fix it. It would not have given much longer code to simply copy over `\MakeUppercase` full original code and modify it "in-place". In this approach, the indexed termed end up uppercased. This looks logical because they are indexed the way they are printed.

- rewrite LaTeX `\MakeUppercase` via another LaTeX method which would try to locate all usages of `\index` and extract them from original, then it can be decided to uppercase or not the indexed term. This is longer LaTeX code patch, has complications with braces etc... but this is only way if we want indexed term not to end up uppercased in index: I know of no way to make some part of argument immune to effect of the TeX primitive `\uppercase` as is used by LaTeX `\MakeUppercase` ; only way I see is to parse the argument to remove from it all we don't want uppercased.

- much simpler on LaTeX side is simply to configure fncychap+Bjarne to not uppercase chapter titles ; this is very simple (`\UCTfalse`) but is a breaking change.

For these reasons I chose first option, but I am aware this may give maintenance problem in future as one needs to be able to read the LaTeX source code lingua perfectly. Also, as it is a hack, it may interfere with some package which would change `\MakeUppercase`.

Altogether I do think it would be better to move index node to after title. This is only strictly needed for this issue at top level sectioning with `'manual'` docclass, but it may be safer to do it systematically. Of course then, the indexed term will not be uppercased, but it will appear uppercased in Chapter titles with Bjarne style. Perhaps this is actually better, as probably author wants the index to represent the word as it is given in rst source?

If you have time to cook-up the Python approach that would be fine with me; we can still merge this temporarily and revert it at later time, as it should be transparent to users.

tk0miya:
@jfbu Thank you for comment. As we do for `\label`, it would be better to move `index` to after the sectioning macros. I just posted #6152 to do that. Could you check it please?

jfbu:
I am withdrawing this as it is better done on Python side in #6152 :)

PR #6157, opened 2019-03-09
refactor: Remove lists of transforms from SphinxStandalonReader

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Now transform classes and Readers are strongly coupled via constants.
- This reduces it.


sphinx/io.py
```diff
@@ -136,12 +123,12 @@ class SphinxI18nReader(SphinxBaseReader):
...
-                  DefaultSubstitutions, MoveModuleTargets, HandleCodeBlocks,
-                  AutoNumbering, SortIds, RemoveTranslatableInline,
-                  FilterSystemMessages, RefOnlyBulletListTransform,
-                  UnreferencedFootnotesDetector, SphinxSmartQuotes, ManpageLink,
-                  SubstitutionDefinitionsRemover]
+    def get_transforms(self):
+        # type: () -> List[Type[Transform]]
+        unused = [PreserveTranslatableMessages, Locale, RemoveTranslatableInline,
+                  AutoIndexUpgrader, FigureAligner, SphinxDomains, DoctreeReadEvent,
+                  UIDTransform]
```
tk0miya, line 132: i18n reader still has list of transforms. It would be better if we can remove this also. Any idea?

sphinx/io.py
```diff
@@ -136,12 +123,12 @@ class SphinxI18nReader(SphinxBaseReader):
...
+        return self.app.registry.get_transforms() - unused
```
tk0miya: Incompatible change: This does not contains transforms which super class returns.

sphinx/io.py
shimizukawa: +1 for this implementation.
i18n builder requires different setup and processing from other builders in various code. Because of this, I think it's difficult to focus on transforms and find a good implementation.

shimizukawa (approved):
LGTM!

PR #6164, opened 2019-03-11
Fix docs of inheritance-diagram "parts" option

### LINKED ISSUE #4872: Documentation of inheritance_diagram extension does not reflect actual behavior
The [documentation of `sphinx.ext.inheritance_diagram`](http://www.sphinx-doc.org/en/master/ext/inheritance.html#directive-inheritance-diagram) states

    This directive supports an option called ``parts`` that, if given, must be an
    integer, advising the directive to remove that many parts of module names
    from the displayed names.  (For example, if all your class names start with
    ``lib.``, you can give ``:parts: 1`` to remove that prefix from the displayed
    node names.)

However, this is not the actual behavior of the code! Due to https://github.com/sphinx-doc/sphinx/blob/96c5605374135652cf7ec0ad517b532b9261ef19/sphinx/ext/inheritance_diagram.py#L234
`parts` specifies the number of parts that are kept (right-to-left), not the number of parts that are dropped (left-to-right). In principle, the documented behavior would be achieved with negative values for `parts`, except that this is prevented by
https://github.com/sphinx-doc/sphinx/blob/96c5605374135652cf7ec0ad517b532b9261ef19/sphinx/ext/inheritance_diagram.py#L333-L338

One would have to change line 334 to `'parts': int,`.

So, I could imagine four possible solutions, which I'll outline below (in order of least to most effort), so people can vote on them:

goerz:
The documentation now correctly describes the behavior of the ``parts`` option in an inheritance-diagram directive: it gives the number of parts that are *kept*, not dropped. The option now also accepts negative values, which drops parts from the left (which is the what the documentation incorrectly claimed the option would do for positive values)

As a form of testing of the new functionality, the documentation for the inheritance_diagram extension now includes a section "Examples" that demonstrate the different possibilities. This would fail to build without the patch.

Closes #4872 ("Solution 3")

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ matrix:
       services: xvfb
 
 install:
+  - "sudo apt-get install graphviz"
   - if [ $IS_PYTHON = true ]; then pip install -U tox codecov; fi
   - if [ $IS_PYTHON = false ]; then npm install; fi
 
--- a/CHANGES
+++ b/CHANGES
@@ -69,6 +69,7 @@ Bugs fixed
 * AssertionError is raised when custom ``citation_reference`` node having
   classes attribute refers missing citation (refs: #6147)
 * #2155: Support ``code`` directive
+* #4872: ext.inheritance_diagram: correctly describe behavior of ``parts`` option in docs, allow negative values.
 
 Testing
 --------
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -7,7 +7,7 @@ import sphinx
 
 extensions = ['sphinx.ext.autodoc', 'sphinx.ext.doctest', 'sphinx.ext.todo',
               'sphinx.ext.autosummary', 'sphinx.ext.extlinks',
-              'sphinx.ext.viewcode']
+              'sphinx.ext.viewcode', 'sphinx.ext.inheritance_diagram']
 
 master_doc = 'contents'
 templates_path = ['_templates']
--- a/doc/usage/extensions/inheritance.rst
+++ b/doc/usage/extensions/inheritance.rst
@@ -25,12 +25,18 @@ It adds this directive:
    graph.
 
    This directive supports an option called ``parts`` that, if given, must be an
-   integer, advising the directive to remove that many parts of module names
-   from the displayed names.  (For example, if all your class names start with
-   ``lib.``, you can give ``:parts: 1`` to remove that prefix from the displayed
-   node names.)
+   integer, advising the directive to keep that many dot-separated parts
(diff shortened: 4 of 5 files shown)


goerz:
I think there's a small problem with `dot` not being available on the Travis CI server. This prevents the "Examples" in the docs (which also serve as a test for this feature working properly) from being generated.

goerz:
I added a `sudo apt-get install graphviz` into the Travis CI configuration in a second commit, which fixes the problem. This should now be ready to merge.

.travis.yml
```diff
@@ -37,6 +37,7 @@ matrix:
       services: xvfb
 
 install:
+  - "sudo apt-get install graphviz"
```
tk0miya, line 40: Is this needed for testing? We're using Travis CI only for testing, and using readthedocs for building document.

tk0miya: If needed, please use APT addon of Travis CI because `sudo` feature is disabled.
https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-with-the-apt-addon

tk0miya (changes requested):
LGTM with nits.
And could you rebase this to 2.0 branch? Then I'll release this as Sphinx-2.0.

Thanks,

.travis.yml
goerz: It is needed, because Travis also runs the tox environment for generating the docs. This is why the first commit https://github.com/sphinx-doc/sphinx/pull/6164/commits/1f136215c18fc75eb9402ec7da083f9022703d71 failed the CI, see https://travis-ci.org/sphinx-doc/sphinx/builds/504521299?utm_source=github_status&utm_medium=notification

Also, Travis now *always* provides sudo (`sudo: false` has no effect, see https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration). Hence, installing graphviz in the second commit does work, which is why the CI is now passing: https://travis-ci.org/sphinx-doc/sphinx/builds/504845281?utm_source=github_status&utm_medium=notification. In light of this, do you still want me to use the APT add-on instead?

You mention readthedocs. I don't see a `readthedocs.yml` file to configure the RTD environment. I don't know if the RTD server has graphviz installed, but if not, building the documentation will fail without some change to the configuration.

tk0miya: Oh sorry, I don't know the change. Okay, let's keep as is. (we need to remove `sudo: false` instead).

About readthedocs, I'll take a look after merging this. And I'll change settings if needed.

tk0miya:
Note: I cherry-picked this into 2.0 branch manually.

tk0miya:

Issue #6174, opened 2019-03-12
AttributeError: 'Sphinx' object has no attribute 'add_description_unit' (pytest pytest-4.3.1.dev110+g51bf7c3a, sphinx 2.0.0b1)


hroncok:
**Describe the bug**
pytest documentation cannot be built.

See https://github.com/pytest-dev/pytest/issues/4912

**To Reproduce**

```
[tmp]$ git clone https://github.com/pytest-dev/pytest.git
...
[tmp]$ cd pytest
[tmp]$ git rev-parse HEAD
51bf7c3aef8f018f44e7c36378be894aa13e32b5
[pytest (master)]$ python3.7 -m venv __venv__
[pytest (master)]$ . __venv__/bin/activate
(__venv__) [pytest (master)]$ pip install --pre sphinx
...
Successfully installed Jinja2-2.10 MarkupSafe-1.1.1 Pygments-2.3.1 alabaster-0.7.12 babel-2.6.0 certifi-2019.3.9 chardet-3.0.4 docutils-0.14 idna-2.8 imagesize-1.1.0 packaging-19.0 pyparsing-2.3.1 pytz-2018.9 requests-2.21.0 six-1.12.0 snowballstemmer-1.2.1 sphinx-2.0.0b1 sphinxcontrib-applehelp-1.0.1 sphinxcontrib-devhelp-1.0.1 sphinxcontrib-htmlhelp-1.0.1 sphinxcontrib-jsmath-1.0.1 sphinxcontrib-qthelp-1.0.2 sphinxcontrib-serializinghtml-1.1.1 urllib3-1.24.1
(__venv__) [pytest (master)]$ cd doc/en
(__venv__) [en (master)]$ pip install -r requirements.txt
...
Successfully installed pygments-pytest-1.2.0 sphinx-removed-in-0.1.3 sphinxcontrib-trio-1.0.2
(__venv__) [en (master)]$ pip install ../..
...
Successfully installed atomicwrites-1.3.0 attrs-19.1.0 more-itertools-6.0.0 pluggy-0.9.0 py-1.8.0 pytest-4.3.1.dev110+g51bf7c3a
(__venv__) [en (master)]$ sphinx-build -W -b html . _build
Running Sphinx v2.0.0b1
making output directory... done

Exception occurred:
  File "/home/churchyard/tmp/pytest/doc/en/conf.py", line 338, in setup
    app.add_description_unit(
[...]

PR #6176, opened 2019-03-13
docs: drop mention to deprecated method (refs: #6174)

tk0miya:
### Feature or Bugfix
- docs

### Purpose
- refs: #6174


hroncok (approved):

tk0miya:
@hroncok Thank you for reviewing!

PR #6185, opened 2019-03-17
Python-3-only clean ups discovered by pyupgrade

jdufresne:
https://github.com/asottile/pyupgrade

> A tool to automatically upgrade syntax for newer versions of the language.

- Drop u str prefix
- Drop base object inheritance
- Drop args to super()
- Use set literals
- Use dict comprehension
- Use set comprehension


tk0miya:
Thanks you always!

tk0miya:
Note: cherry-picked this into 2.0 branch manually.

Issue #4187, opened 2017-10-24
EN DASH disappears from PDF bookmarks in Japanese documents


jfbu:
Consider following index.rst

```rst
Welcome to FOO's documentation!
===============================

もしもし–––さようなら
---------------------

何もない

```

and ``language = 'ja'``. Then ``make latexpdf`` includes following console output:

```
dvipdfmx:warning: No character mapping available.
 CMap name: EUC-UCS2
 input str: <85>
```

and after enquiry it comes from the PDF bookmarks. Indeed the EN DASH are ok in PDF, but absent from bookmark. Picture:

![capture d ecran 2017-10-24 a 17 40 52](https://user-images.githubusercontent.com/2589111/31953186-d5ebe650-b8e2-11e7-9015-17e068763555.png)

Sphinx escapes the U+2013 (EN DASH) to ``\textendash{}`` in LaTeX, but this is not the cause of the problem. It happens also when using the approach at https://github.com/jfbu/sphinx/tree/fixplatexendash and inserting literal ``–`` in the tex file.

### Environment info
- OS: Mac
- Python version: 3.5.4
- Sphinx version: 1.6.5
- <Extra tools e.g.: Browser, tex or something else>

jfbu:
The issue may have to do with this

```
  \RequirePackage{atbegshi}
  \ifx\ucs\undefined
    \ifnum 42146=\euc"A4A2
      \AtBeginShipoutFirst{\special{pdf:tounicode EUC-UCS2}}
    \else
      \AtBeginShipoutFirst{\special{pdf:tounicode 90ms-RKSJ-UCS2}}
    \fi
  \else
    \AtBeginShipoutFirst{\special{pdf:tounicode UTF8-UCS2}}
  \fi
```

Is this a problem Japanese LaTeX users know how to solve?

Using  ``\PassOptionsToPackage{pdfencoding=unicode}{hyperref}`` fixes the EN DASH problem in bookmarks but destroys all Japanese characters there...

tk0miya:
To resolve this, we need to migrate to uplatex. With platex, we use EUC-JP as internal encoding of PDF bookmarks. But the encoding does not contain EN-DASH. As a workaround, we can replace them by other characters (ex. hyphen in ASCII) before outputting .tex code. But to keep using pLaTeX is not important. So it would be better to migrate.

PR #6189, opened 2019-03-19
LaTeX: stop using extractbb for image inclusion in Japanese documents

jfbu:
Since TeXLive2015, the dvipdfmx binary does not need extra .xbb files
for images (which were produced using extractbb).

I propose to wait for Sphinx 3.0 because although a TeXLive2015-based TeX distribution is listed as requirement since Sphinx 2.0, it is not needed to really force users to upgrade only to help us trim a few lines on our side which are neither an added feature nor a bugfix but simply some steps which were mandatory in the past and are now not needed.

Or at least wait until we shift from platex to uplatex for Japanese projects (refs: https://github.com/sphinx-doc/sphinx/issues/4187#issuecomment-471262204)

edit: I forgot to add [this link](https://tug.org/pipermail/tex-live/2017-January/039545.html) for source of information about `extractbb` unneeded since TL2015.

### DIFF
--- a/sphinx/texinputs/Makefile_t
+++ b/sphinx/texinputs/Makefile_t
@@ -10,7 +10,6 @@ ALLDVI = $(addsuffix .dvi,$(ALLDOCS))
 ALLXDV =
 {% endif -%}
 ALLPS  = $(addsuffix .ps,$(ALLDOCS))
-ALLIMGS = $(wildcard *.png *.gif *.jpg *.jpeg)
 
 # Prefix for archive names
 ARCHIVEPREFIX =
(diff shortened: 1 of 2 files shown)


wizardwhite2009:
This is all Greek to me, but I personally have all of Adobes products.

tk0miya (approved):
+1, but current `master` branch is going forward v2.1.
At present, no appropriate branch for v3.0 there.

tk0miya:
Now master branch is ready for 3.0!

jfbu:
Thanks for reviewing! merging now that master is moving towards 3.0!

PR #6201, opened 2019-03-23
Fix #6165: autodoc: ``tab_width`` setting of docutils has been ignored

### LINKED ISSUE #6165: `sphinx.ext.autodoc` should respect `docutils` `tab_width` configuration
Currently the `sphinx.ext.autodoc` directives (e.g. `.. automodule::`) expand tabs to 8 spaces irrespective of the `tab_width` configuration of `docutils`.

This could be fixed by changing the `expandtabs()` call in `sphinx.util.docstrings.prepare_docstring` to pass the correct `tab_width` value.

(See #642 for more context).

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6165


sphinx/ext/autodoc/directive.py
```diff
@@ -49,19 +49,25 @@ def __getitem__(self, key):
...
+        self.state = state
         self.filename_set = set()  # type: Set[str]
         self.result = StringList()
 
     def warn(self, msg):
         # type: (str) -> None
         logger.warning(msg, location=(self.env.docname, self.lineno))
 
+    @property
+    def settings(self):
```
tk0miya: I feel this is not docutils' way. Directive class does not have settings object directly.

sphinx/ext/autodoc/directive.py
```diff
@@ -49,19 +49,25 @@ def __getitem__(self, key):
 class DocumenterBridge:
     """A parameters container for Documenters."""
 
-    def __init__(self, env, reporter, options, lineno):
-        # type: (BuildEnvironment, Reporter, Options, int) -> None
+    def __init__(self, env, reporter, options, lineno, state):
```
tk0miya: `DocumenterBridge` is a private, not public API class. But this must be a breaking change. So it would be better to make `state` argument optional.

PR #6203, opened 2019-03-24
Add CitationDomain for citation processing

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To reduce responsibility of StandardDomain and make it simple, this migrates the responsibility to new domain; CitationDomain


sphinx/domains/citation.py
```diff
@@ -0,0 +1,162 @@
...
+                               type='ref', subtype='citation', location=(docname, lineno))
+
+    def resolve_xref(self, env, fromdocname, builder, typ, target, node, contnode):
+        # type: (BuildEnvironment, str, Builder, str, str, addnodes.pending_xref, nodes.Element) -> nodes.Element  # NOQA
+        docname, labelid, lineno = self.citations.get(target, ('', '', 0))
+        if not docname:
+            if 'ids' in node:
+                # remove ids attribute that annotated at
+                # transforms.CitationReference.apply.
+                del node['ids'][:]
```
tk0miya: Before merging, I have to check this is still needed.

sphinx/domains/std.py
```diff
@@ -705,7 +653,10 @@ def resolve_xref(self, env, fromdocname, builder, typ, target, node, contnode):
         elif typ == 'option':
             resolver = self._resolve_option_xref
         elif typ == 'citation':
-            resolver = self._resolve_citation_xref
+            warnings.warn('pending_xref(domain=std, type=citation) is deprecated: %r' % node,
+                          RemovedInSphinx40Warning)
```
tk0miya, line 680: I'm worry about this is breaking change or not.

sphinx/domains/citation.py
tk0miya: Now we've been wrap citation labels with `inline` node. As a result, `Element.replace_self()` has no longer warned. So this code is not needed now.

sphinx/domains/std.py
```diff
@@ -659,14 +615,6 @@ def add_program_option(self, program, name, docname, labelid):
         # type: (str, str, str, str) -> None
         self.data['progoptions'][program, name] = (docname, labelid)
 
-    def check_consistency(self):
```
tk0miya, line 685: The superclass; `Domain` class still provides empty `check_consistency()` method. So this is not breaking change.

sphinx/domains/std.py
```diff
@@ -584,31 +565,6 @@ def merge_domaindata(self, docnames, otherdata):
                 self.data['anonlabels'][key] = data
 
     def process_doc(self, env, docname, document):
-        # type: (BuildEnvironment, str, nodes.document) -> None
-        self.note_citations(env, docname, document)
-        self.note_citation_refs(env, docname, document)
-        self.note_labels(env, docname, document)
```
tk0miya, line 614: I noticed these removal is breaking change.

sphinx/domains/std.py
shimizukawa: https://semver.org/ mentioned as:

1. MAJOR version when you make incompatible API changes,
2. MINOR version when you add functionality in a backwards-compatible manner, and
3. PATCH version when you make backwards-compatible bug fixes.

and I think 'deprecation warning' is not a breaking change so it is OK to be included in MINOR version.

shimizukawa (approved):
LGTM!

tk0miya:
Thank you for reviewing!

PR #6209, opened 2019-03-26
Support coverage_ignore_py in the coverage extension

Gallaecio:
Subject: Extend the existing coverage extension with an option that allows ignoring Python API members regardless of their type, and based on their full import path.

### Feature or Bugfix
- Feature

### Purpose
Currently, the coverage extension only allows ignoring Python API members based on their base name and type. This means that you cannot ignore a specific attribute name on a specific class, it’s either all or none. It also mean that you cannot use a single expression to ignore Python API members of different types. You cannot ignore specific class attributes either.

This new setting allows doing all of it with a simple list of regular expressions that are checked against the full import path of any Python API member.

### Relates
- https://github.com/scrapy/scrapy/pull/3567

### DIFF
--- a/doc/usage/extensions/coverage.rst
+++ b/doc/usage/extensions/coverage.rst
@@ -22,6 +22,16 @@ should check:
 
 .. confval:: coverage_ignore_classes
 
+.. confval:: coverage_ignore_pyobjects
+
+   List of `Python regular expressions`_.
+
+   If any of these regular expressions matches any part of the full import path
+   of a Python object, that Python object is excluded from the documentation
+   coverage report.
+
+   .. versionadded:: 2.1
+
 .. confval:: coverage_c_path
 
 .. confval:: coverage_c_regexes
@@ -40,3 +50,5 @@ should check:
    ``False`` by default.
 
    .. versionadded:: 1.1
+
+.. _Python regular expressions: https://docs.python.org/library/re
--- a/sphinx/ext/coverage.py
+++ b/sphinx/ext/coverage.py
@@ -79,6 +79,8 @@ class CoverageBuilder(Builder):
                                                 self.config.coverage_ignore_classes)
         self.fun_ignorexps = compile_regex_list('coverage_ignore_functions',
                                                 self.config.coverage_ignore_functions)
+        self.py_ignorexps = compile_regex_list('coverage_ignore_pyobjects',
+                                               self.config.coverage_ignore_pyobjects)
 
     def get_outdated_docs(self):
         # type: () -> str
@@ -130,6 +132,12 @@ class CoverageBuilder(Builder):
                     op.write(' * %-50s [%9s]\n' % (name, typ))
                 op.write('\n')
 
+    def ignore_pyobj(self, full_name):
+        for exp in self.py_ignorexps:
+            if exp.search(full_name):
+                return True
+        return False
+
     def build_py_coverage(self):
         # type: () -> None
         objects = self.env.domaindata['py']['objects']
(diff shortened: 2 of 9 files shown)


Gallaecio:
I looked into the existing tests, to extend them, but they did not seem to cover the existing type-specific options, and the comments seem to suggest that testing this extensions is rather difficult. Please, let me know if I should work on covering the new option with tests nonetheless.

Gallaecio:
Could you let me know if I should change something other than fixing conflicts with the `CHANGES` file?

Gallaecio:
@tk0miya Are you the one [I should be bugging](http://www.sphinx-doc.org/en/master/devguide.html#contributing-to-sphinx)? :)

doc/usage/extensions/coverage.rst
```diff
@@ -22,6 +22,14 @@ should check:
 
 .. confval:: coverage_ignore_classes
 
+.. confval:: coverage_ignore_py
```
tk0miya: `ignore_py` is a bit ambiguous to me. What about `coverage_ignore_pyobject` instead?

doc/usage/extensions/coverage.rst
```diff
@@ -22,6 +22,14 @@ should check:
...
+
+   List of `Python regular expressions`_.
+
+   If any of these regular expressions matches any part of the full import path
+   of a Python API member, that Python API member is excluded from the
```
tk0miya: "Python API member" is common wording?

doc/usage/extensions/coverage.rst
```diff
@@ -22,6 +22,14 @@ should check:
...
+   documentation coverage report.
+
```
tk0miya, line 34: Please add `versionadded:: 2.1` here.

tk0miya (changes requested):
Sorry for late response. LGTM for this idea!

Could you add a testcase to `tests/test_ext_coverage.py` please? New features should be tested.
Thanks,

doc/usage/extensions/coverage.rst
Gallaecio: Switched to `coverage_ignore_pyobjects` (plural to keep it coherent with the other 3 Python options).

doc/usage/extensions/coverage.rst
Gallaecio: Switched to ‘Python objects’. I had no idea what the common wording was :)

tests/roots/test-ext-coverage/python.txt
```diff
@@ -0,0 +1,2 @@
+Undocumented Python objects
+===========================
```
tk0miya: No results does not mean exclusion working fine. Please add an example not filtered.

tk0miya: Is there reason to store expected result to file?
I prefer to define it inside test script. How about that?

sphinx/ext/coverage.py
```diff
@@ -130,6 +132,12 @@ def write_c_coverage(self):
                     op.write(' * %-50s [%9s]\n' % (name, typ))
                 op.write('\n')
 
+    def ignore_py(self, full_name):
```
tk0miya: I think this should be also renamed to `ignore_pyobj`.

tk0miya (changes requested):
LGTM with nits

tk0miya:
In addition, please rebase this to 2.0 branch. master branch is used for 3.0 release now.

tests/roots/test-ext-coverage/python.txt
Gallaecio: Done.

Both the ignored and the non-ignored modules have now the same content, which includes an undocumented class and undocumented methods within a documented class.

Gallaecio:
Rebased.

tk0miya (approved):
LGTM!

tk0miya:

Issue #5877, opened 2019-01-02
autosummary: Add option to document imported members


flying-sheep:
`generate_autosummary_docs` has an option `imported_members` which is by default `False`.

`process_generate_options` calls it without that being set to `True`. I’d like to be able to override this.

tk0miya:
Fixed by #6212 .
Thank you for reporting!

Issue #4372, opened 2018-01-03
Getting apidoc --imported-members behaviour with autosummary


hameerabbasi:
Subject: Getting apidoc `--imported-members` behaviour with autosummary

### Problem
- I'm trying to find out how to get `sphinx-apidoc`'s `--imported-members` functionality with `autosummary_generate = True`. I can't seem to find anything for this in the issue tracker. It also ignores `__all__` in `__init__.py` for the module.

#### Procedure to reproduce the problem
```
rm -rf build/ docs/_build/ docs/generated/ && python setup.py build_sphinx
```

#### Error logs / results
```
running build_sphinx
creating /Users/hameerabbasi/PycharmProjects/sparse/build
creating /Users/hameerabbasi/PycharmProjects/sparse/build/sphinx
creating /Users/hameerabbasi/PycharmProjects/sparse/build/sphinx/doctrees
creating /Users/hameerabbasi/PycharmProjects/sparse/build/sphinx/html
Running Sphinx v1.6.5
loading pickled environment... not yet created
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/numpy/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/scipy/reference/objects.inv...
[autosummary] generating autosummary for: index.rst, modules.rst
[autosummary] generating autosummary for: /Users/hameerabbasi/PycharmProjects/sparse/docs/_autosummary/sparse.rst
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 2 source files that are out of date
updating environment: 3 added, 0 changed, 0 removed
reading sources... [100%] modules                                               
/Users/hameerabbasi/PycharmProjects/sparse/docs/modules.rst:7: WARNING: Error in "autosummary" directive:
unknown option: "imported_members".

.. autosummary::
   :toctree: _autosummary
   :imported_members: True

   sparse
looking for now-outdated files... none found
pickling environment... done
checking consistency... /Users/hameerabbasi/PycharmProjects/sparse/docs/_autosummary/sparse.rst: WARNING: document isn't included in any toctree
done
preparing documents... done
writing output... [100%] modules                                                
generating indices... genindex py-modindex
writing additional pages... search
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded, 2 warnings.
```
- <public link of unexpected result if you have>

#### Expected results
[...]

PR #4247, opened 2017-11-17
BUG: autosummary to list all class members

jorisvandenbossche:
Fixes the availability of the list of attributes in the autosummary template.

### Feature or Bugfix
- Bugfix

### Purpose

This PR intends to be a more general fix for https://github.com/sphinx-doc/sphinx/commit/b3e0e29f4cd3a3bec12a07336ebbda430e25e27e. That commit fixed issue https://github.com/sphinx-doc/sphinx/issues/3900 for the methods being no longer available in the autosummary template. This generalizes that so also the Attributes are again available in the autosummary template.

The fact that those methods/attributes are no longer available is/was a regression in the 1.6.x line.

### Detail
The rationale of the fix is that the "imported_members=False" (which is the default) is not relevant for a class, but only for a module. Therefore I removed this in the `elif doc.objtype == 'class':` branch.

### Relates
- https://github.com/sphinx-doc/sphinx/issues/2336, https://github.com/sphinx-doc/sphinx/pull/3235
- https://github.com/sphinx-doc/sphinx/issues/3900
- https://github.com/sphinx-doc/sphinx/issues/4058

PR #6214, opened 2019-03-27
Allow autosummary to document modules exposed through __all__

tsanikgr:
Allow `autosummary` to document modules exposed through `__all__`

### Feature or Bugfix
- Feature

### Purpose
Autosummary ignores the `__all__` attribute of modules, and does not provide a way of automatically generating documentation them.

Consider the example of the included test:

`__init__.py`:
```python
__all__ = [
    'bar'
]
```

`bar/__init__.py`:
```python
# empty file
```

`bar/foo.py`:
```python
class Baz:
   pass
```

This PR allows auto-documenting the module `bar`, by using the new jinja template variable `public_modules` (which contains everything in `__all__`), provided that the user provides an appropriate `_template`. See the included test for detail

### Relates
- #4247
- #4372
- #5877

### DIFF
--- a/sphinx/ext/autosummary/generate.py
+++ b/sphinx/ext/autosummary/generate.py
@@ -18,6 +18,7 @@
 """
 
 import argparse
+import importlib
 import locale
 import os
 import pydoc
@@ -186,6 +187,19 @@ def generate_autosummary_docs(sources, output_dir=None, suffix='.rst',
                           if x in include_public or not x.startswith('_')]
                 return public, items
 
+            def get_public_modules(obj, typ):
+                # type: (Any, str) -> List[str]
+                items = []  # type: List[str]
+                for item in getattr(obj, '__all__', []):
+                    try:
+                        importlib.import_module(name + '.' + item)
+                    except ImportError:
+                        continue
+                    finally:
+                        sys.modules.pop(name + '.' + item)
+                    items.append(name + '.' + item)
+                return items
+
             ns = {}  # type: Dict[str, Any]
 
             if doc.objtype == 'module':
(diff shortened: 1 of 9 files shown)


tk0miya:
Note: This is related with #6040

sphinx/ext/autosummary/generate.py
```diff
@@ -186,6 +187,20 @@ def get_members(obj, typ, include_public=[], imported=True):
                           if x in include_public or not x.startswith('_')]
                 return public, items
 
+            def get_public_modules(obj, typ):
```
tk0miya, line 190: It seems `typ` is not used. Why is this needed?

sphinx/ext/autosummary/generate.py
```diff
@@ -186,6 +187,20 @@ def get_members(obj, typ, include_public=[], imported=True):
...
+                # type: (Any, str) -> List[str]
+                items = []  # type: List[str]
+                for item in getattr(obj, '__all__', []):
+                    try:
+                        importlib.import_module(name + '.' + item)
+                    except ImportError:
+                        continue
+                    finally:
+                        if item in sys.modules:
+                            sys.modules.pop(name + '.' + item)
```
tk0miya, line 200: Why do you unload this? I'm afraid this unloading affects to other programs...

sphinx/ext/autosummary/generate.py
```diff
@@ -196,6 +211,7 @@ def get_members(obj, typ, include_public=[], imported=True):
                     get_members(obj, 'class', imported=imported_members)
                 ns['exceptions'], ns['all_exceptions'] = \
                     get_members(obj, 'exception', imported=imported_members)
+                ns['public_modules'] = get_public_modules(obj, 'module')
```
tk0miya, line 214: I feel `public_members`, `public_functions`, etc. are also needed. Why is `public_modules` only added?

tk0miya:
Now we adopted #6040 to build a document for the package instead. So I'm closing this now.

PR #6215, opened 2019-03-27
Clarify MathJax extension usage

rkdarst:
### Feature or Bugfix
- Bugfix (docs)

### Purpose

- We recently hit an issue where we expected the MathJax extension to   work with the default MathJax markup.  Obviously that shouldn't bethe case (since then other formats couldn't use the math), but it can't hurt to include a hint.

### Detail

- Link from the math HTML extension page to the relevant roles/directives.
- Also make it explicit that MathJax has a default, and that it's an external resource that will be dynamically linked.
- My only consideration is that possibly it doesn't deserve an explicit mention in the MathJax section, only in the page prologue.  But assuming MathJax is so common and has it's other similar syntax, I mentioned it there too.
- The mention of the CDN was to make people think at least a little bit before including stuff from all over.

Note: we tested this the wrong way, but not the right way.  Also let me know if this is the wrong way to do things.

I'm submitting this to the 2.0, please let me know if it should target another branch.

### DIFF
--- a/doc/usage/extensions/math.rst
+++ b/doc/usage/extensions/math.rst
@@ -16,7 +16,7 @@ Math support for HTML outputs in Sphinx
 
 Since mathematical notation isn't natively supported by HTML in any way, Sphinx
 gives a math support to HTML document with several extensions.  These use the
-Sphinx :rst:dir:`directives <math>` and :rst:role:`roles <math>`.
+reStructuredText math :rst:dir:`directive <math>` and :rst:role:`role <math>`.
 
 :mod:`sphinx.ext.imgmath` -- Render math as images
 --------------------------------------------------
@@ -129,12 +129,16 @@ built:
 
 This extension puts math as-is into the HTML files.  The JavaScript package
 MathJax_ is then loaded and transforms the LaTeX markup to readable math live in
-the browser.  Reminder: you should use the math :rst:dir:`directives <math>` and
-:rst:role:`roles <math>`, not the native MathJax ``$$``, ``\(``, etc.
+the browser.
 
 Because MathJax (and the necessary fonts) is very large, it is not included in
-Sphinx and your build will include resources from a external site (by
-default cdnjs.cloudflare.com).
+Sphinx but is set to automatically include it from a third-party site.
+
+.. attention::
+
+   You should use the math :rst:dir:`directive <math>` and
+   :rst:role:`role <math>`, not the native MathJax ``$$``, ``\(``, etc.
+
 
 .. confval:: mathjax_path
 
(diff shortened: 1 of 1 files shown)


doc/usage/extensions/math.rst
```diff
@@ -15,7 +15,8 @@ Math support for HTML outputs in Sphinx
    So mathbase extension is no longer needed.
 
 Since mathematical notation isn't natively supported by HTML in any way, Sphinx
-gives a math support to HTML document with several extensions.
+gives a math support to HTML document with several extensions.  These use the
+Sphinx :rst:dir:`directives <math>` and :rst:role:`roles <math>`.
```
tk0miya: `math` directive and role are not Sphinx's original. They come from reStructuredText. So "Sphinx" word is not needed here.
http://docutils.sourceforge.net/docs/ref/rst/directives.html#math
http://docutils.sourceforge.net/docs/ref/rst/roles.html

tk0miya: And I think both `directive` and `role` should be singular form.

doc/usage/extensions/math.rst
```diff
@@ -128,10 +129,12 @@ built:
 
 This extension puts math as-is into the HTML files.  The JavaScript package
 MathJax_ is then loaded and transforms the LaTeX markup to readable math live in
-the browser.
+the browser.  Reminder: you should use the math :rst:dir:`directives <math>` and
```
tk0miya: Could you use `.. attension::` directive here?

doc/usage/extensions/math.rst
```diff
@@ -128,10 +129,12 @@ built:
...
+:rst:role:`roles <math>`, not the native MathJax ``$$``, ``\(``, etc.
 
 Because MathJax (and the necessary fonts) is very large, it is not included in
-Sphinx.
+Sphinx and your build will include resources from a external site (by
+default cdnjs.cloudflare.com).
```
tk0miya: It would be better to merge this into description of `mathjax_path` setting. What do you think?

tk0miya (commented):
LGTM with nits

doc/usage/extensions/math.rst
rkdarst: Very true, I remotely started thinking of this a little bit too late.  Fixed.

rkdarst: Good point, fixed (and also added "math" so that it reads "math directive and role")

doc/usage/extensions/math.rst
rkdarst: You are right, it belongs below and is basically already said there.  But then I thought the real reason I change it was that this sentence makes it seem like it won't work by default.  So, I changed this part to just say "it's not included but will work automatically by using a third party site", which I think is useful to have there.  And moved the explicit note warning of third-party resources to the lower part.  It's sort of said twice, but I think it's OK.  What do you think?

doc/usage/extensions/math.rst
rkdarst: Done, good idea.

rkdarst:
Thanks for the review, all good points.  Pushed with updates - let me know if still you'd like more.

tk0miya (approved):
Thank you for quick update. LGTM!

tk0miya:

rkdarst:
Most importantly, thanks for Sphinx, too.

PR #6223, opened 2019-03-29
Drop features and APIs deprecated in 1.8

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Removed features marked as RemovedInSphinx30.


sphinx/application.py
```diff
@@ -25,17 +25,15 @@
 from sphinx import package_dir, locale
 from sphinx.config import Config
 from sphinx.config import CONFIG_FILENAME  # NOQA # for compatibility (RemovedInSphinx30)
```
tk0miya, line 26: I noticed this comment. But there are no deprecation warnings for this variable. So I'd like to postpone this to 4.0 (after emitting warning).

tk0miya: I determined to emit warning this on 2.0.1 (refs: #6224)

Issue #6221, opened 2019-03-29
WARNING: unknown option for :option:


buzz:
**Description**

- I receive several `WARNING: unknown option: -v` when building docs.
- Not all <code>:option:\`-v\`</code> are working.

**To Reproduce**

```
$ git clone https://github.com/buzz/sphinx-option-testcase.git
$ cd sphinx-option-testcase
$ python -m venv venv
$ . venv/bin/activate
$ pip install -r requirements.txt
$ make html
Running Sphinx v2.0.0
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 4 source files that are out of date
updating environment: 4 added, 0 changed, 0 removed
reading sources... [100%] otherb                                                                                                                                              
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] otherb                                                                                                                                               
/tmp/sphinx-option-testcase/cli.rst:4: WARNING: unknown option: -v
/tmp/sphinx-option-testcase/othera.rst:4: WARNING: unknown option: -v
/tmp/sphinx-option-testcase/otherb.rst:4: WARNING: unknown option: -v
generating indices... genindex
writing additional pages... search
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded, 3 warnings.

The HTML pages are in _build/html.
$ # open _build/html/index.html in browser
$ # only the 2nd reference to option -v on cli.rst worked
```

**Expected behavior**

- All four references to option `-v` should work regardless if they occur *before* or *after* the `.. option:: -v` or in another document.
- No warning should be given.

**Your project**
Minimal test case: https://github.com/buzz/sphinx-option-testcase

**Environment info**
- OS: `Linux 4.20.8-arch1-1-ARCH #1 SMP PREEMPT Wed Feb 13 00:06:58 UTC 2019 x86_64 GNU/Linux`
- Python version: 3.7.2
- Sphinx version: 2.0.0 (same for 1.8.5)
- Sphinx extensions:  none

tk0miya:
The following example describes `-v` option for `cli` command. So it means `cli -v`.
```
.. program:: cli

.. option:: -v

	verbose
```

So you need to refer the `-v` option by ``` :option:`cli -v` ``` or like following:
```
.. program:: cli

:option:`-v`
```

refs: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-program

[...]

PR #6228, opened 2019-03-30
Clarify option directive usage

buzz:
Subject: Add example to ``option`` directive used within ``program``.

### Feature or Bugfix
- Docs update

### Purpose
- It's not quite clear how to reference an option correctly. This PR makes it more explicit in the docs.

### Detail
- Add a sentence to the `option` directive description with an example how to use it within the `program` directive.

### Relates
- #6221

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -1225,6 +1225,8 @@ There is a set of directives allowing documenting command-line programs:
    The directive will create cross-reference targets for the given options,
    referenceable by :rst:role:`option` (in the example case, you'd use something
    like ``:option:`dest_dir```, ``:option:`-m```, or ``:option:`--module```).
+   When used within the :rst:dir:`program` directive, however, it is mandatory
+   to also specify the program *name* (e.g. ``:option:`python -m```).
 
    ``cmdoption`` directive is a deprecated alias for the ``option`` directive.
 


stephenfin (approved):
Yup, this is my understanding of things too

doc/usage/restructuredtext/domains.rst
```diff
@@ -1225,6 +1225,8 @@ There is a set of directives allowing documenting command-line programs:
    The directive will create cross-reference targets for the given options,
    referenceable by :rst:role:`option` (in the example case, you'd use something
    like ``:option:`dest_dir```, ``:option:`-m```, or ``:option:`--module```).
+   When used within the :rst:dir:`program` directive, however, it is mandatory
```
tk0miya, line 1228: As commented https://github.com/sphinx-doc/sphinx/issues/6221#issuecomment-478344019, the program name is not mandatory. It's optional.

tk0miya (changes requested):

doc/usage/restructuredtext/domains.rst
buzz: If you leave out the program name, Sphinx fails as I showed in #6221.

In order for Sphinx not to give warnings and be able to resolve references it seems, one *must* specify the program name (hence mandatory).

(I see, you pointed out a [special case](https://github.com/sphinx-doc/sphinx/issues/6221#issuecomment-478344019), where the reference in fact does resolve without program name when it's used *after* the program directive. Note, that this only works within the same file. Also, I would deem this behavior counter-intuitive and undesirable. Also I fail to find it documented anywhere.)

In case I misunderstood, could you please provide an example how it is optional when used with the program directive?

tk0miya: What I meant is a "special case" you mentioned. And I agree with the behavior is not documented. So it would be better to describe the behavior of option role including how program directive works.

Issue #1368, opened 2015-01-03
Makefile options not overridable from command line


shimizukawa:
doc/Makefile states:

```
#!make

# You can set these variables from the command line.
```

However to make the variables overridable from the command line,
the assignments must be changed:

```
#!patch

--- Makefile.orig       Fri Jan 31 18:02:10 2014
+++ Makefile    Fri Jan 31 18:02:33 2014
@@ -2,11 +2,11 @@
 #

 # You can set these variables from the command line.
-SPHINXOPTS   =
-SPHINXBUILD  = python ../sphinx-build.py
-SPHINXPROJ   = sphinx
-SOURCEDIR    = .
-BUILDDIR     = _build
+SPHINXOPTS  ?=
+SPHINXBUILD ?= python ../sphinx-build.py
+SPHINXPROJ  ?= sphinx
+SOURCEDIR   ?= .
+BUILDDIR    ?= _build

 # Has to be explicit, otherwise we don't get "make" without targets right.
 help:

```

This is also an issue in every project that adopts this Makefile.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1368
- Originally reported by: [minusf](https://bitbucket.org/minusf)
- Originally created at: 2014-01-31T12:06:57.338

shimizukawa:
_From [minusf](https://bitbucket.org/minusf) on 2014-01-31 11:24:04+00:00_

i just realised this should be fixed in `sphinx.quickstart.MAKEFILE`

while we are here, is the new `doc/Makefile` an indication of what will happen to `sphinx.quickstart.MAKEFILE`? I am asking because the new `Makefile` works only with GNU make (BSD make does not recognise `%` as a catchall target :( so it would be nice to keep it like it is...

tk0miya:
Fixed by #6232.
Thank you, @svenevs!

jfbu:
I would like to reopen discussion on this. I feel it is dangerous as [commented here](https://github.com/sphinx-doc/sphinx/pull/6277#pullrequestreview-226355286). It is a recent addition, so perhaps can still be partially reverted with not too much impact on Sphinx users. There are many ways it could go wrong in relation to  `make clean` but not only. Basically I feel it is bad to obey environment variables with such a generic name as BUILDDIR.

svenevs:
[...]

PR #6232, opened 2019-03-31
Enable CLI override of Makefile variables.

svenevs:
Subject: Enable CLI override of Makefile variables.  Generate POSIX compliant Makefile.

### Feature or Bugfix

- Bugfix

### Purpose

The `?=` setting in `Makefile` allows users to specify e.g., `SPHINXOPTS='...' make <builder>`.  Expanded comment to give example of how to do that in template.

The other bug being fixed here is that the generated `Makefile` is missing the final newline.  I figured I'd just throw the fix in here since it's pretty minor.  I can create a separate issue and PR if desired, but that feels like overkill to me :)

### Detail

- Make sure `Makefile` generates `?=` variable setting so that they actually can be overriden on the command line.
- Make sure rendered `Makefile` has a final newline.

### Relates

- #1368

P.S. I think I got the right branches here, let me know if I should cherry pick this to another branch.

### DIFF
--- a/sphinx/templates/quickstart/Makefile.new_t
+++ b/sphinx/templates/quickstart/Makefile.new_t
@@ -1,11 +1,14 @@
 # Minimal makefile for Sphinx documentation
 #
 
-# You can set these variables from the command line.
-SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
-SOURCEDIR     = {{ rsrcdir }}
-BUILDDIR      = {{ rbuilddir }}
+# You can set these variables from the command line.  For example:
+#     SPHINXOPTS='-E -W -n' make html
+# will run the html builder in a clean environment (-E), treating warnings
+# as errors (-W), in nitpicky mode (-n).
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     ?= {{ rsrcdir }}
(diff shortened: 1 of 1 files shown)


tk0miya (approved):
+1

tk0miya:

svenevs:
Sure thing!  GitHub didn't seem to actually close the linked issue (normally that's automated when doing `Closes` or `Fixes`).

jfbu (commented):
I think we should revert this. The only thing to fix was to say that only SPHINXOPTS can be set on command line or exported from environment. It has always worked for me for example ``make html O="-D language=fr"``. Allowing the other variables to be set from environment is too dangerous. This patch is merged in 2.1.0 which is not yet released. I propose to revert it. More minor: I don't see the necessity of a blank line at end of the file.

**edit**: the blank line is needed at end of the file, as it is a template. Without it the produced Makefiles lack a final EOL as explained by @svenevs. Thanks!

**edit**: I did another error in my review which is to imply that SPHINXOPTS could be set from the environment; that was an error. (I though being assigned empty value in Makefile it was especially handled). Actually it was true as commented in Makefile that the variables could be overriden from command line. There is no need for  ``?=`` for that. The ``?=`` allows override from exported environment variables. Thus, the extra explanations in the PR are useful, but the  ``?`` is not needed for command line variables.

PR #6233, opened 2019-03-31
refactor quickstart

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Replace triple-quotes string by concatenated string to improve readability
- This causes the change of message catalogs. So this should be merged in next major release.


stephenfin (approved):
Yup, definitely a fan of this kind of cleanup.

tk0miya:
@stephenfin Thank you for reviewing!

PR #6234, opened 2019-03-31
Update type annotations for python domain

tk0miya:
### Feature or Bugfix
- Refactoring


sphinx/directives/__init__.py
```diff
@@ -166,7 +166,7 @@ def run(self):
         node['objtype'] = node['desctype'] = self.objtype
         node['noindex'] = noindex = ('noindex' in self.options)
 
-        self.names = []  # type: List[str]
+        self.names = []  # type: List[Any]
```
tk0miya, line 169: `self.names` stores the result of `handle_signature`. So this should be `Any` value.

Issue #6255, opened 2019-04-05
Create hrefs for :bcp:


ioggstream:
## I wish

Support for :BCP: like :RFC: and :PEP:

## It could be done

reusing RFC/PEP code in [roles.py](sphinx/roles.py)

tk0miya:
This is first time I hear about BCP. So I can't determine it is commonly used or not.
If it is commonly used and worthy, it should be added into reStructuredText standard. How about proposing this to docutils project?

IMO, adding to new directives and roles into Sphinx causes division of reST syntax more.

ioggstream:
BCP are IETF Best Current Practices https://en.wikipedia.org/wiki/Best_current_practice
which are served from the same RFC website.

A couple of BCP are https://tools.ietf.org/html/bcp47 (language tags) and https://tools.ietf.org/html/bcp42

@tk0miya if there's an easy way to contribute it to `docutils` (eg. their github seem stale, and the sourceforge page seems very old http://docutils.sourceforge.net/ ) please let me know: I have some patches to propose them too.

tk0miya:
They're still working on sourceforge. The page you mentioned is active :-)
To contribute them, please post a patch from here: https://sourceforge.net/p/docutils/patches/.

tk0miya:
I'm closing this now.
Thank you for your proposal. I wish this proposal will be merged into docutils!

ioggstream:
Ok, for future reference, see https://sourceforge.net/p/docutils/feature-requests/61/

PR #6256, opened 2019-04-05
Fix: #6255. Support IETF :BCP: anchors, like :RFC:.

ioggstream:
Subject: support IETF :BCP: (best current practices)

### Feature or Bugfix
- Feature

### Purpose
- Support IETF :BCP: anchors, like :RFC:.

### Detail
Like RFC, IETF releases Best Current Practices (BCP) that are usually referenced  in technical documents.

This code reuses :RFC: code in roles.py to process BCP.

### Relates
- #6255

### DIFF
--- a/sphinx/roles.py
+++ b/sphinx/roles.py
@@ -338,6 +338,43 @@ class RFC(ReferenceRole):
             return base_url + self.inliner.rfc_url % int(ret[0])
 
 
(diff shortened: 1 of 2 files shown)


tk0miya:
As I commented in #6255, it would be better to propose this to docutils. So I'm closing this now.
Thank you for proposal.

PR #6257, opened 2019-04-06
Fix #6244: html: Search function is broken with 3rd party themes

### LINKED ISSUE #6244: Search function broken in 2.0.0
**Describe the bug**
When searching in the built HTML file, no results are diplayed and only "Searching..." is showing.

The browser console shows:
```
jquery.js:2 jQuery.Deferred exception: Search.loadIndex is not a function TypeError: Search.loadIndex is not a function
    at HTMLDocument.<anonymous> (http://localhost:63342/documentation/_build/search.html?q=redis&check_keywords=yes&area=default:250:32)
    at j (http://localhost:63342/documentation/_build/_static/jquery.js:2:29999)
    at k (http://localhost:63342/documentation/_build/_static/jquery.js:2:30313) undefined
```

Looks like a regression from #467

**To Reproduce**
```
sphinx-build -b html -W -n -E -a -j auto ./docs ./_build
```

**Expected behavior**
Search is working

**Your project**
Even an empty index.rst does it.

**Screenshots**

**Environment info**
- OS: Win
- Python version: 3.6.6
- Sphinx version: 2.0.0
- Sphinx extensions:  /
- Extra tools: Latest chrome or firefox

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6244

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -29,6 +29,7 @@ Bugs fixed
   references
 * #6245: circular import error on importing SerializingHTMLBuilder
 * #6243: LaTeX: 'releasename' setting for latex_elements is ignored
+* #6244: html: Search function is broken with 3rd party themes
 
 Testing
 --------
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -75,6 +75,16 @@ var Search = {
       }
   },
 
+  loadIndex : function(url) {
+    $.ajax({type: "GET", url: url, data: null,
+            dataType: "script", cache: true,
+            complete: function(jqxhr, textstatus) {
+              if (textstatus != "success") {
+                document.getElementById("searchindexloader").src = url;
+              }
+            }});
+  },
+
   setIndex : function(index) {
     var q;
     this._index = index;


bnmnetp:
Search is still broken for me in 2.1.2.   I do have a custom theme and have been trying to figure out what is happening.

Javascript throws an error on line 66 of searchtools.js because docContent is undefined.

The loadIndex function is not called for me anywhere.  Nor do I find any element in the generated source that would match '[role=main]'

Any hints on how to fix my templates or where to look from here would be much appreciated!

Thanks,

vsalvino:
I am also getting the same error with `docContent` undefined. Tried using various stock themes instead of my custom theme, same issue persists throughout.

One item of note, it seems to be "more broken" when building with `dirhtml` mode versus `html` mode. The error occurs in both but it seems to only occur about half of the time in `html` mode. This leads me to believe there is some kind of race condition at play, such as a file or request not being finished before the search script starts working.

vsalvino:
Update: I figured out the issue, it appears to be a bug in `searchtools.js` that was incorrectly querying the URLs. And another semi-related bug which might related to 3rd party themes (the `docContent` is undefined) if your HTML output does not contain a `role="main"` element. See PR #7192 .

PR #6260, opened 2019-04-06
Make EventManager portable

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose

So far, we need to bypass application object for modules to emit
an event.  This make EventManager portable and easy to pass event
emitter.  This brings modules less coupled with application object.

### DIFF
--- a/sphinx/events.py
+++ b/sphinx/events.py
@@ -10,8 +10,10 @@
     :license: BSD, see LICENSE for details.
 """
 
+import warnings
 from collections import OrderedDict, defaultdict
 
+from sphinx.deprecation import RemovedInSphinx40Warning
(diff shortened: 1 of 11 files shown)


sphinx/events.py
```diff
@@ -42,20 +46,25 @@
 
 
 class EventManager:
-    def __init__(self):
-        # type: () -> None
+    """Event manager for Sphinx."""
+
+    def __init__(self, app):
+        # type: (Sphinx) -> None
+        self.app = app
```
tk0miya, line 58: This breaks compatibility. So it would be better to make this optional.

tk0miya:
I merged this into 2.0 branch manually.

PR #6270, opened 2019-04-08
ENH: Explain warnings-as-errors

larsoner:
Subject: Make clear why build fails on CIs

### Feature or Bugfix
- Feature

### Purpose
- In our automated doc builds we run `-W --keep-going`. Contributors make doc errors that we rightfully treat as warnings. They are confused why the CircleCI build log says there were errors. This makes it clearer that "build finished with problems, 1 warning." is actually problematic, and that they should look to see what the warning is, since it's treated as an error.

### Detail
Changes:
```
build finished with problems, 1 warning.
```
To
```
build finished with problems, 1 warning (with warnings treated as errors).
```

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -348,9 +348,16 @@ class Sphinx:
             status = (self.statuscode == 0 and
                       __('succeeded') or __('finished with problems'))
             if self._warncount:
-                logger.info(bold(__('build %s, %s warning.',
-                                    'build %s, %s warnings.', self._warncount) %
-                                 (status, self._warncount)))
+                if self.warningiserror:
+                    msg = __('build %s, %s warning (with warnings treated as errors).',
+                             'build %s, %s warnings (with warnings treated as errors).',
+                             self._warncount)
+                else:
+                    msg = __('build %s, %s warning.',
+                             'build %s, %s warnings.',
+                             self._warncount)
+
+                logger.info(bold(msg % (status, self._warncount)))
             else:
                 logger.info(bold(__('build %s.') % status))
 


sphinx/application.py
```diff
@@ -343,9 +343,14 @@ def build(self, force_all=False, filenames=None):
...
                       __('succeeded') or __('finished with problems'))
             if self._warncount:
-                logger.info(bold(__('build %s, %s warning.',
-                                    'build %s, %s warnings.', self._warncount) %
-                                 (status, self._warncount)))
+                if self.warningiserror:
+                    explanation = ' (with warnings treated as errors)'
+                else:
+                    explanation = ''
+                logger.info(bold(__('build %s, %s warning%s.',
```
stephenfin: I _think_ you need to do the concatenation of strings before you log, otherwise the `explanation` string won't be translated. If I'm correct, you should probably do:

    msg = 'build %s, %s warning%s.'
    if self.warningiserror:
        msg += ' (with warnings treated as errors)'

    logger.info(bold(__(msg, msg, self.warn_count) % (status, self._warncount)))

But I'll wait for someone else more familiar with the translation framework to speak.

tk0miya: our translation tool searches `__("...")` and extracts them as a translation message. So both `"build %s,..."` and `" (with warnings ..."` should be surrounded by `__(...)` function.

BTW, the concatenation is hard to translate because translators could not understand the context for each message (they don't read source code). So I prefer this:
```
                if self.warningiserror:
                    msg = __('build %s, %s warning (with warnings treated as errors).',
                             'build %s, %s warnings (with warnings treated as errors).',
                             self._warncount)
                else:
                    msg = __('build %s, %s warning.',
                             'build %s, %s warnings.',
                             self._warncount)

                logger.info(bold(msg % (status, self._warncount)))
```

tk0miya (changes requested):
Could you rebase this to 2.0 branch? Then this will be released as 2.1

larsoner:
Changed the base to 2.0, cherry-picked my commit, modified the message, amended the commit, and push-forced. Should be good now.

tk0miya (approved):
LGTM!

tk0miya:

PR #6277, opened 2019-04-11
Fix #6271: make clean is catastrophically broken if building into '.'

### LINKED ISSUE #6271: Make clean is catastrophically broken if building into '.'
**Describe the bug**

If the build directory is . the default makefile destroys the entire document tree and configuration when you make clean

**To Reproduce**
Set the build directory to .
make clean

**Expected behavior**
Not to trash the source document

**Environment info**
- Linux (Fedora)
- Python 3.6.8
- Sphinx version: 1.7.5

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6271

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -68,6 +68,7 @@ Features added
 * #6232: Enable CLI override of Makefile variables
 * #6212 autosummary: Add :confval:`autosummary_imported_members` to display
   imported members on autosummary
+* #6271: ``make clean`` is catastrophically broken if building into '.'
 
 Bugs fixed
 ----------
--- a/sphinx/cmd/make_mode.py
+++ b/sphinx/cmd/make_mode.py
@@ -72,10 +72,18 @@ class Make:
 
     def build_clean(self):
         # type: () -> int
+        srcdir = path.abspath(self.srcdir)
+        builddir = path.abspath(self.builddir)
         if not path.exists(self.builddir):
             return 0
         elif not path.isdir(self.builddir):
             print("Error: %r is not a directory!" % self.builddir)
+            return 1
+        elif srcdir == builddir:
+            print("Error: %r is same as source directory!" % self.builddir)
+            return 1
+        elif path.commonpath([srcdir, builddir]) == builddir:
+            print("Error: %r directory contains source directory!" % self.builddir)
             return 1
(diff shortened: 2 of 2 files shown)


jfbu (commented):
As far as I know the only way to get the ourput directory to be `.` is to use BUILDDIR, am I right? this is not proposed by sphinx-quickstart. But then this PR will not intercept the case that BUILDDIR was set by user to some other strange value such as $HOME. IMHO, an error should be raised for any Sphinx build action if the BUILDDIR contains the source dir, or contains the configuration file conf.py.

jfbu (commented):
Probably it would be too restrictive to limit the OUTPUTDIR argument of sphinx-build, because many projects may use it their way, but regarding the BUILDDIR in Makefile the make mode should arguably refuse any value not being a sub repertory of the project root, which must be distinct and must not include source repertory.  Else indeed potential for disaster exists if BUILDDIR env variable is set in Makefile and source project gets distributed to others with instructions to use the Makefile.

jfbu (changes requested):
Hence the make mode should check for inclusion of BUILDDIR in root dir and for SOURCEDIR not being a sub-repertory of BUILDDIR.

The restriction on BUILDDIR is presumably a breaking change, but currently nothing prevents someone to set `BUILDDIR = ..` in Makefile or worse and the   `make clean` is really dangerous then.

tk0miya:
@jfbu Thank you for reviewing. I updated this PR to check ancestor directories also.

jfbu (approved):
LGTM. While reviewing this I noticed that for some time now the Makefile created by quickstart defines the BUILDDIR using ``?=`` syntax, which means it can be set in the environment. BUILDDIR could very well be set in environment for another software. The current PR will not avoid `make clean` to remove such a foreign BUILDDIR. For better security it could be argued that BUILDDIR should be named SPHINXBUILDDIR. This would not avoid malicious manœuvers but reduce risks of accidents. Notice also that if you work on two Sphinx projects and have somehow set BUILDDIR in the environment, the `make clean` car remove the build directory of the other project even it includes its source directory.

**edit** the `?=`  is only currently merged into master see #6232 and discussion at #1368.

tk0miya:
Thank you for reviewing!

PR #6288, opened 2019-04-13
Add support for bound methods posing as functions in the module

### LINKED ISSUE #6287: Unable to document bound instance methods exported as module functions
We're building an API documentation site for Horovod project (http://horovod.ai) and faced the problem with documenting bound methods that we export as module functions (duck typing yay!)

Simple example:
```python
class _Boo(object):
    def m(self):
        """Everybody needs a method that returns 42, right?"""
        return 42


_b = _Boo()
n = _b.m
```

Real example: [declaration](https://github.com/horovod/horovod/blob/68f525c64cdd5a31536f5485448960fbb00d19fc/horovod/common/__init__.py#L58), [usage](https://github.com/horovod/horovod/blob/6643d6aa4abd583e5d254e2658e413a309f47a54/horovod/tensorflow/mpi_ops.py#L64).

I found that it does not show up in the docs because it does not match either of these conditions:

`FunctionDocumenter`:
```python
@classmethod
def can_document_member(cls, member, membername, isattr, parent):
    # type: (Any, unicode, bool, Any) -> bool
    return isfunction(member) or isbuiltin(member)
```

`MethodDocumenter`:
```python
@classmethod
def can_document_member(cls, member, membername, isattr, parent):
    # type: (Any, unicode, bool, Any) -> bool
    return inspect.isroutine(member) and \
        not isinstance(parent, ModuleDocumenter)
```

I'd like those doc strings to show up in the docs.

I tried modifying both `MethodDocumenter` and `FunctionDocumenter` conditions to allow for `inspect.isroutine(member) and isinstance(parent, ModuleDocumenter)` and found that `FunctionDocumenter` gives more satisfying results.

What is the best way to document those functions?  Would you take a PR where `FunctionDocumenter` is modified as follows:

```python
@classmethod
def can_document_member(cls, member, membername, isattr, parent):
    # type: (Any, unicode, bool, Any) -> bool
    return isfunction(member) or isbuiltin(member) or (inspect.isroutine(member) and  isinstance(parent, ModuleDocumenter))
```

... or maybe there's another way to achieve this?

alsrgv:
Subject: add support for bound methods posing as functions in the module

Fixes #6287

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -993,7 +993,8 @@ class FunctionDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # typ
     @classmethod
     def can_document_member(cls, member, membername, isattr, parent):
         # type: (Any, str, bool, Any) -> bool
-        return inspect.isfunction(member) or inspect.isbuiltin(member)
+        return (inspect.isfunction(member) or inspect.isbuiltin(member) or
+                (inspect.isroutine(member) and isinstance(parent, ModuleDocumenter)))
 
     def format_args(self):
         # type: () -> str
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -993,6 +993,7 @@ class FunctionDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # typ
     @classmethod
     def can_document_member(cls, member, membername, isattr, parent):
         # type: (Any, str, bool, Any) -> bool
+        # supports functions, builtins and bound methods exported at the module level
         return (inspect.isfunction(member) or inspect.isbuiltin(member) or
                 (inspect.isroutine(member) and isinstance(parent, ModuleDocumenter)))
 
--- a/tests/test_autodoc.py
+++ b/tests/test_autodoc.py
@@ -259,6 +259,11 @@ def test_format_signature():
     assert formatsig('method', 'H.foo', H.foo2, None, None) == '(*c)'
     assert formatsig('method', 'H.foo', H.foo3, None, None) == r"(d='\\n')"
 
+    # test bound methods interpreted as functions
+    assert formatsig('function', 'foo', H().foo1, None, None) == '(b, *c)'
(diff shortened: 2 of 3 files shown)


alsrgv:
@tk0miya, please suggest additional tests, if any.

sphinx/ext/autodoc/__init__.py
```diff
@@ -991,7 +991,8 @@ class FunctionDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # typ
     @classmethod
     def can_document_member(cls, member, membername, isattr, parent):
         # type: (Any, str, bool, Any) -> bool
-        return isfunction(member) or isbuiltin(member)
+        return (isfunction(member) or isbuiltin(member) or
+                (inspect.isroutine(member) and isinstance(parent, ModuleDocumenter)))
```
tk0miya, line 998: Could you add comment for this code? I've often lost sight of purpose of the code because autodoc is too complicated.

tests/test_autodoc.py
```diff
@@ -279,6 +284,7 @@ def foo3(self, d='\n'):
         '(b, c=42, *d, **e)'
 
 
+
```
tk0miya: This empty line is not needed.

tk0miya (changes requested):
At first, could you rebase this to 2.0 branch? Then, it will be released as Sphinx-2.1.0.

Next, could you add a testcase which generates reST marked up text like `test_partialfunction()` and so on?

1. Add a test subject to `tests/roots/test-ext-autodoc/target/functions.py`
2. Add a testcase to `tests/test_autodoc.py`. There are many examples for generating internal reST code.

sphinx/ext/autodoc/__init__.py
alsrgv: Is the added comment sufficient, or did you have something more in mind?

alsrgv:
@tk0miya, thanks for the feedback.  I've updated the PR, please let me know if you have more comments.

sphinx/ext/autodoc/__init__.py
tk0miya: LGTM :-)

tk0miya (approved):
Thank you for update! LGTM.

tk0miya:

PR #4092, opened 2017-09-27
Make it easier to override variables in Makefiles generated by quickstart

mitya57:
This way one can use e.g. `SPHINXBUILD=/usr/bin/sphinx-build make html` to override the default sphinx-build path.

References (but not fixes) #3965.

PR #6303, opened 2019-04-14
Partially revert #6232, synchronize new Makefile with old (refs: #4092)

jfbu:
### Relates
- #6232, #4092, discussion at #1368 

At #4092 it was decided to use ``?=`` for old Makefile and ``SPHINXOPTS`` and ``SPHINXBUILD``. So these variables can be overriden not only from command line but from environment. I have some doubts about whether this was good idea (as command line override of Makefile variables is already available), but Sphinx shipped with it for some time now.

The present patch keeps the ``?=`` of #6232 only for these two variables, and of course it keeps the needed blank line at end of the template (else generated Makefiles have no final EOL).

### DIFF
--- a/sphinx/templates/quickstart/Makefile.new_t
+++ b/sphinx/templates/quickstart/Makefile.new_t
@@ -1,14 +1,12 @@
 # Minimal makefile for Sphinx documentation
 #
 
-# You can set these variables from the command line.  For example:
-#     SPHINXOPTS='-E -W -n' make html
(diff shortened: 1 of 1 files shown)


svenevs (approved):
Awesome thanks for talking this out and fixing it :slightly_smiling_face:

tk0miya (approved):

jfbu:
Thanks for reviewing!

PR #6304, opened 2019-04-14
Fix the string in quickstart for 'path' argument of parser

jfbu:
I noticed that doing ``sphinx-quickstart --help`` produced

```
positional arguments:
  PROJECT_DIR           répertoire de sortie
```

in French translation, but checking the sphinx.po file I saw that problem was in original string which looked as `output path`. This appears to be wrong. This variable refers to the project root.


jfbu:
Closing because I forgot to rename my local branch from 2.0.2...

PR #6305, opened 2019-04-15
Fix the string in quickstart for 'path' argument of parser

jfbu:
I noticed that doing ``sphinx-quickstart --help`` produced

```
positional arguments:
  PROJECT_DIR           répertoire de sortie
```

in French translation, but checking the sphinx.po file I saw that problem was in original string which looked as `output path`. This appears to be wrong. This variable refers to the project root.

This is same as #6304 but with the correct branch in my fork of main repo...


tk0miya (approved):
LGTM.

jfbu:

PR #6306, opened 2019-04-15
Add a label to search input for accessability purposes

thedrow:
Subject: Improve accessibility of the quick search box

### Feature or Bugfix

- Feature

### Purpose
Accessibility is important. Those with screen readers should be made aware that this is the quick search input.

### Detail
I just added an id to the "Quick Search" header and pointed the ARIA attribute to use that as a description.

### Relates
None

### DIFF
--- a/sphinx/themes/basic/searchbox.html
+++ b/sphinx/themes/basic/searchbox.html
@@ -9,10 +9,10 @@
 #}
 {%- if pagename != "search" and builder != "singlehtml" %}
 <div id="searchbox" style="display: none" role="search">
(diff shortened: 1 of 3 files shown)


tk0miya (changes requested):
+1; Could you update testcases also? And it would be nice if this PR rebases to 2.0 branch.

thedrow:
So this shouldn't be merged to master?

tk0miya:
I merged this into 2.0 branch. Then this will be released as 2.1 soon.
Thank you for contribution!

PR #6308, opened 2019-04-15
imgmath: Use template file for LaTeX

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Before modifying imgmath, this starts to use template file to generate LaTeX source.
- It makes imgmath module simple.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -44,6 +44,9 @@ Deprecated
 * ``sphinx.ext.autodoc.importer.MockLoader``
 * ``sphinx.ext.autodoc.importer.mock()``
 * ``sphinx.ext.autosummary.autolink_role()``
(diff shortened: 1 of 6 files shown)


jfbu (approved):
Sorry I posted #6310 before seeing this one. I think it is good idea, because it helps propose in futuer solution to Sphinx users  without having to modify core Sphinx. Only thing is perhaps we should discuss structure of header: why the utf8x? I will need to check later history to see the reason. +1 at any rate.

tk0miya:
Thank you for comment. Merging.

>Only thing is perhaps we should discuss structure of header: why the utf8x? I will need to check later history to see the reason.

Okay, let's update them. I guess there are no special reasons.

Issue #5301, opened 2018-08-14
SVG output is not as nice as MathJax


certik:
The default SVG output looks like this:

![screenshot_2018-08-14 5 1 maxwell s equations theoretical physics reference 0 5 documentation](https://user-images.githubusercontent.com/20568/44106624-67423742-9fb2-11e8-9c35-a61d3653e638.png)

As you can see, the math font is much thinner than the surrounding ReST text, so it looks incompatible. Also the equations are quite thin and look gray instead of black.

MathJax, on the other hand, looks like this:

![screenshot_2018-08-14 5 1 maxwell s equations theoretical physics reference 0 5 documentation 1](https://user-images.githubusercontent.com/20568/44106705-9a56cc38-9fb2-11e8-85a6-1042bf9389dc.png)

The fonts are very much compatible with the surrounding ReST text, and also the equations are a bit fuller, and so they look good and the color is black.

In general, when mixing ReST and math in one sentence, the MathJax looks natural, while SVG does not.

How can this be fixed? It seems the solution is to do both of these:

* Increase the size of equations/fonts in LaTeX
* Somehow make the fonts "fuller" / "thicker".

asmeurer:
SVG is just using Computer Modern, so if you want it to look like the surrounding text, you should use the same font. MathJax also uses a different font from the main text (your main text font isn't even serif), but I guess it happens to look closer. 

This should all be modifiable by changing the preamble.

certik:
@asmeurer thanks. I need to figure out what font MathJax uses and then try to use the same font in LaTeX. I'll update the issue here if I figure it out.

asmeurer:
I think it uses STIX. Apparently you can change it https://docs.mathjax.org/en/latest/font-support.html

tk0miya:
Is there anything to do from Sphinx side? If nothing, I'd like to close this issue.

certik:
@tk0miya what Sphinx should do is to make the SVG equations by default look like the MathJax ones.

jfbu:
Maybe https://github.com/jennigorham/mathjax-fonts-in-latex can help you. Follow installation instructions (for your locale or on the build system you use) and adjust accordingly [imgmath_latex_preamble](https://www.sphinx-doc.org/en/master/usage/extensions/math.html?highlight=svg#confval-imgmath_latex_preamble). (untested)

[...]

PR #6310, opened 2019-04-16
Make imgmath_use_preview work also for svg output

jfbu:
Currently, inline mathematics in svg format is not well aligned with surrounding text.

This PR improves the situation.

Relates: #5301

### DIFF
--- a/doc/usage/extensions/math.rst
+++ b/doc/usage/extensions/math.rst
@@ -95,19 +95,17 @@ built:
 
 .. confval:: imgmath_use_preview
 
-   ``dvipng`` has the ability to determine the "depth" of the rendered text: for
-   example, when typesetting a fraction inline, the baseline of surrounding text
-   should not be flush with the bottom of the image, rather the image should
-   extend a bit below the baseline.  This is what TeX calls "depth".  When this
-   is enabled, the images put into the HTML document will get a
-   ``vertical-align`` style that correctly aligns the baselines.
+   ``dvipng`` and ``dvisvgm`` have the ability to collect from LaTeX the
+   "depth" of the rendered text: an inline image should use this "depth" in a
+   ``vertical-align`` style to be correctly aligned with surrounding text.
 
-   Unfortunately, this only works when the `preview-latex package`_ is
-   installed. Therefore, the default for this option is ``False``.
+   This mechanism requires the `LaTeX preview package`_ (available as
+   ``preview-latex-style`` on Ubuntu xenial).  Therefore, the default for this
+   option is ``False`` but it is strongly recommended to set it to ``True``.
 
-   .. versionchanged:: 2.1.0
+   .. versionchanged:: 2.1
 
-      This option can also be used with ``imgmath_image_format`` set to ``'svg'``.
+      This option can be used with the ``'svg'`` :confval:`imgmath_image_format`.
 
 .. confval:: imgmath_add_tooltips
 
@@ -221,4 +219,4 @@ package jsMath_.  It provides this config value:
 .. _dvisvgm: http://dvisvgm.bplaced.net/
 .. _MathJax: https://www.mathjax.org/
 .. _jsMath: http://www.math.union.edu/~dpvc/jsmath/
-.. _preview-latex package: https://www.gnu.org/software/auctex/preview-latex.html
+.. _LaTeX preview package: https://www.gnu.org/software/auctex/preview-latex.html
--- a/doc/usage/extensions/math.rst
+++ b/doc/usage/extensions/math.rst
@@ -105,8 +105,9 @@ built:
    Unfortunately, this only works when the `preview-latex package`_ is
    installed. Therefore, the default for this option is ``False``.
 
-   Currently this option is only used when ``imgmath_image_format`` is
-   ``'png'``.
+   .. versionchanged:: 2.1.0
+
+      This option can also be used with ``imgmath_image_format`` set to ``'svg'``.
 
 .. confval:: imgmath_add_tooltips
 
--- a/doc/usage/extensions/math.rst
+++ b/doc/usage/extensions/math.rst
@@ -30,74 +30,21 @@ This extension renders math via LaTeX and dvipng_ or dvisvgm_ into PNG or SVG
 images. This of course means that the computer where the docs are built must
 have both programs available.
 
-There are various config values you can set to influence how the images are
-built:
+There are various configuration values you can set to influence how the images
+are built:
 
 .. confval:: imgmath_image_format
 
-   The output image format. The default is ``'png'``.  It should be either
-   ``'png'`` or ``'svg'``.
-
-.. confval:: imgmath_latex
-
-   The command name with which to invoke LaTeX.  The default is ``'latex'``; you
-   may need to set this to a full path if ``latex`` is not in the executable
-   search path.
-
-   Since this setting is not portable from system to system, it is normally not
-   useful to set it in ``conf.py``; rather, giving it on the
-   :program:`sphinx-build` command line via the :option:`-D <sphinx-build -D>`
-   option should be preferable, like this::
-
-      sphinx-build -b html -D imgmath_latex=C:\tex\latex.exe . _build/html
-
-   This value should only contain the path to the latex executable, not further
-   arguments; use :confval:`imgmath_latex_args` for that purpose.
-
-.. confval:: imgmath_dvipng
-
-   The command name with which to invoke ``dvipng``.  The default is
-   ``'dvipng'``; you may need to set this to a full path if ``dvipng`` is not in
-   the executable search path. This option is only used when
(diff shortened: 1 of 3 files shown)


sphinx/ext/imgmath.py
```diff
@@ -76,7 +76,7 @@ class InvokeError(SphinxError):
 '''
 
 DOC_BODY_PREVIEW = r'''
-\usepackage[active]{preview}
+\usepackage[active,tightpage]{preview}
```
jfbu: This is needed for dvisvgm to gather "depth" information. I hope it does not affect dvipng usage (with its -Ttight option).

sphinx/ext/imgmath.py
```diff
@@ -85,7 +85,7 @@ class InvokeError(SphinxError):
 '''
 
 depth_re = re.compile(br'\[\d+ depth=(-?\d+)\]')
-
+depthsvg_re = re.compile(br'.*, depth=(.*)pt')
```
jfbu, line 93: Here is some example output from dvisvgm when the dvi was produced with tightpage option of preview.

```
$ dvisvgm --no-fonts testsvg
pre-processing DVI file (format version 2)
processing page 1
  computing extents based on data set by preview package (version 11.91)
  width=103.354867pt, height=9.225002pt, depth=3.075001pt
  graphic size: 103.354867pt x 12.300003pt (36.325081mm x 4.322957mm)
  output written to testsvg.svg
1 of 1 page converted in 0.253939 seconds
```

Perhaps the regex can be made better?

sphinx/ext/imgmath.py
```diff
@@ -197,8 +197,17 @@ def convert_dvi_to_svg(dvipath, builder):
...
-    convert_dvi_to_image(command, name)
-    return filename, None
+    stdout, stderr = convert_dvi_to_image(command, name)
+
+    depth = None
+    if builder.config.imgmath_use_preview:
+        for line in stderr.splitlines():  # not stdout !
+            matched = depthsvg_re.match(line)
+            if matched:
+                depth = round(float(matched.group(1)) * 100 / 72.27)  # assume 100ppi
```
jfbu, line 245: as shown above the depth may have fractional digit. The unit is the TeX point (72.27pt=1in). We could convert it to HTML "pt" unit (72pt=1in) but due to shared code with "png" format, we have to return an integral number of pixels. Thus I round it to an integer here and assume 100ppi.

Clearly any advice welcome on how to better handle this. Probably I have to change more of imgmath to handle a non integer depth. Should we pass it over to html output using pt unit or if we use px how to best convert? The `dvipng` returns an integral number of `px` but I don't know how it does the conversion internally from TeX dimensions to `px` unit.

sphinx/ext/imgmath.py
```diff
@@ -197,8 +197,17 @@ def convert_dvi_to_svg(dvipath, builder):
     command.extend(builder.config.imgmath_dvisvgm_args)
     command.append(dvipath)
 
-    convert_dvi_to_image(command, name)
-    return filename, None
+    stdout, stderr = convert_dvi_to_image(command, name)
+
+    depth = None
+    if builder.config.imgmath_use_preview:
+        for line in stderr.splitlines():  # not stdout !
```
jfbu, line 242: The dvisvgm output (as shown above) goes to stderr... whereas the dvipng used stdout...

jfbu (commented):
This gives somewhat satisfactory result in my testing of inline math with depth (for example anything using ``(``, ``)`` or letters ``p, q,...``). But it is frustrating to lose the precision of the ``dvisvgm`` depth indication.

Consider this a bit of WIP. Suggestions welcome on how to improve this.

jfbu:
I think the usage of preview package to get the needed depth info should not be optional (whether for PNG or SVG). The preview package is old package, we can assume it to be available in Sphinx user LaTeX installation (I will check on Ubuntu xenial). And inline math without this depth correction basically looks very bad (be it with PNG or SVG).

jfbu (commented):
I forgot to comment on lines 238-240. For png it seems Sphinx hardcodes the depth inside the png file (for reuse). Probably this must be extended to SVG case. As I did not see immediately how to handle this (and wondering if the whole thing might become obsolete once we let dvipng or dvisvgm handle many equations at once instead of one per one) I put together this PR in this WIP state, sorry.

jfbu:
On xenial I got [this](https://packages.ubuntu.com/search?searchon=contents&keywords=preview.sty&mode=exactfilename&suite=xenial&arch=any):

```

/usr/share/texmf/tex/latex/preview/preview.sty | preview-latex-style
-- | --
/usr/share/xemacs21/xemacs-packages/etc/auctex/latex/preview.sty | xemacs21-basesupport

```

I propose we add ``preview-latex-style`` as dependency at some later Sphinx release so that ``imgmath_use_preview``  defaults to True. Display mathematics does not need it but inline math snippets is basically broken without it for PNG and SVG.

sphinx/ext/imgmath.py
jfbu: In fact adding this option did change the dvipng console output and thus broke imgmath for it because the regex did not find matches anymore. Fixed in second commit.

jfbu: We probably don't need to use ``-T tight`` option of dvipng anymore.

jfbu:
Unfortunately the ``tightpage``  option of preview package breaks PNG rendering of display math: equations get shifted to the right. This might be caused by whitespace in LaTeX source which we did not see because of ``dvipng -Ttight`` but under the ``tightpage`` option of preview package, ``dvipng`` acts differently.

jfbu:
I will update this PR. The dvipng and dvisvgm must be separated more one reason being that there is bad interaction of ``tightpage`` option of ``preview`` package with dvipng in case of display math: the display gets shifted to the right whether or not we use the ``-Ttight`` option of ``dvipng``.  Thus, we must separate more the code branches for PNG and SVG cases.

**edit** I have done that now.

sphinx/ext/imgmath.py
jfbu: The ``tightpage`` option does not play well with ``dvipng``. In my latest commit I am thus using it *only*  for ``dvisvgm``. As per my previous comment, the ``-T tight`` option of ``dvipng`` remains required.

jfbu:
> I forgot to comment on lines 238-240. For png it seems Sphinx hardcodes the depth inside the png file (for reuse). Probably this must be extended to SVG case.

Done now.

jfbu:
Squashed and rebased on master tip after merge of #6308

sphinx/ext/imgmath.py
```diff
@@ -90,15 +90,35 @@ class InvokeError(SphinxError):
...
-def generate_latex_macro(math, config, confdir=''):
-    # type: (str, Config, str) -> str
+def read_svg_depth(filename):
+    # type: (str) -> int
+    """Read the depth from comment at last line of SVG file
+    """
+    with open(filename, 'r') as f:
+        for line in f:
+            pass
+        return int(line[11:-4])
```
tk0miya: I feel this is a bit fragile. Could you replace this by regexp?

sphinx/ext/imgmath.py
tk0miya: +1: enough to me.

tk0miya (approved):
Sorry for late. LGTM with nits.

doc/usage/extensions/math.rst
```diff
@@ -105,8 +105,9 @@ built:
    Unfortunately, this only works when the `preview-latex package`_ is
    installed. Therefore, the default for this option is ``False``.
 
-   Currently this option is only used when ``imgmath_image_format`` is
-   ``'png'``.
+   .. versionchanged:: 2.1.0
```
tk0miya: I've written `versionchnaged:: 2.1` (not 2.1.0) in another PR. I don't know what is best.

jfbu: I will use `2.1` too. I still need to adjust a bit to our new branch/release model.

sphinx/ext/imgmath.py
jfbu: I think you mean to use a regex for the parsing of very last line (cf `write_svg_depth()`), and we can pass over silently other lines?

sphinx/ext/imgmath.py
jfbu: I think in future more separate handling of `dvipng` and `dvisvgm` could be considered. Because here I coerce to an `int` mainly due to shared code between the two.

Also, `dvisvgm` can generate its own hashing for naming svg files.

jfbu (commented):
I will update, thanks for reviewing.

tk0miya (approved):
Great work! LGTM :-)

jfbu:
Thanks for reviewing :-) merging...

PR #6313, opened 2019-04-22
Improve CI runtime.

thedrow:
We now run with two pytest workers in CI in order to cut the build time
in half.

In addition, I added a small fix to avoid unnecessary bash-fu.

I've also added a package which arranges the output nicely on Travis-CI. It doesn't affect anywhere else.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -33,7 +33,7 @@ test_script:
       if (-not $test_ignore) { $test_ignore = '' }
       $tests = $env:TEST
       if (-not $tests) { $tests = '' }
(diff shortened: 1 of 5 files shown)


thedrow:
The coverage decrease can be handled using coveragerc.
I'll do it if you're going to accept this PR.

tk0miya:
It seems testing time on Travis CI has not been changed. I don't know why. But this does not effect our tests.

thedrow:
I'll check. But we have manged to improve the time on windows.

tk0miya:
Sadly, we moved to GitHub Actions. So this improvement is no longer needed. Closing.
Thank you for your work.

Issue #4701, opened 2018-03-02
Misplaced ``<div>`` in viewcode html output since Sphinx 1.7.0


jfbu:
Image shows it:

![capture d ecran 2018-03-02 a 15 13 27](https://user-images.githubusercontent.com/2589111/36902950-81851e38-1e2c-11e8-91fb-30f2bf1a4d92.png)

there is an extra line above the final one of the procedure, and the html source (in ``_build/html/_modules/foo.html``) looks like this

```html
<div class="viewcode-block" id="pgcd"><a class="viewcode-back" href="../index.html#foo.pgcd">[docs]</a><span class="k">def</span> <span class="nf">pgcd</span><span class="p">(</span><span class="n">a</span><span class="p">,</span> <span class="n">b</span><span class="p">):</span>
    <span class="sd">&quot;&quot;&quot;Calcule le PGCD par l&#39;algorithme d&#39;Euclide.</span>
<span class="sd">    &quot;&quot;&quot;</span>
    <span class="n">a</span><span class="p">,</span> <span class="n">b</span> <span class="o">=</span> <span class="nb">abs</span><span class="p">(</span><span class="n">a</span><span class="p">),</span> <span class="nb">abs</span><span class="p">(</span><span class="n">b</span><span class="p">)</span>
    <span class="k">while</span> <span class="n">b</span> <span class="o">&gt;</span> <span class="mi">0</span><span class="p">:</span>
        <span class="n">a</span><span class="p">,</span> <span class="n">b</span> <span class="o">=</span> <span class="n">b</span><span class="p">,</span> <span class="n">a</span> <span class="o">%</span> <span class="n">b</span></div>
    <span class="k">return</span> <span class="n">a</span>
```

the closing ``</div>`` is mis-placed (not at end of last line but at end of previous one).

This does not happen with Sphinx 1.6.6 but happens with Sphinx 1.7.0 and 1.7.1

I am attaching 
[misplaceddiv.zip](https://github.com/sphinx-doc/sphinx/files/1775396/misplaceddiv.zip)
which one needs to compile with ``make html``.

### Environment info
- OS: Mac
- Running Sphinx v1.8.0+
- Sphinx version: 1.8.0+
- Python version: 3.6.4 (CPython)
- Docutils version: 0.14 
- Jinja2 version: 2.10

Also tested with 1.7.0 and 1.7.1. No problem with 1.6.6

jfbu:
[...]

PR #6319, opened 2019-04-23
 Add line numbers in ``sphinx.ext.viewdoc``

benkrikler:
Subject: Add an option to the viewcode extension to include line-numbering in the output

### Feature or Bugfix
- Feature
- (and included in here, I believe is potentially a Bugfix; this wasn't the original intention, so I could split it out)

### Purpose
I felt that it could help a project I was working on if I the source code as highlighted by the `viewcode` extension could include line numbers.  When I looked into the viewcode source, it seemed relatively straightforward to add this, and so as a way to learn a bit about the internals of sphinx I took a go at it.
At that point, it seemed like it could be something of use / interest and worth contributing back to the main repository.  I'm sorry that this hasn't followed what I believe is the normal approach to adding a feature (I read the Contributing guidelines, and whilst it didn't look like a hard rule, I think I should have opened an issue first)!

### Detail
Along the way though, I noticed that, at least in my browser (Firefox 66.0.2) the extra `div` inserted around the `[docs]` link in the highlighted sourcecode output was adding an extra line intermittently (this was with tag v2.0.1 of Sphinx).  The div only seems to have served to add a css class, so I have moved that css class to the `<a>` element instead and removed the encasing `div`.  I appreciate that might be a bit drastic (for example, styles might be selecting on the `div`), but it did fix the issue.  It seems #4701 and #4709 concerned this problem as well, but as far as I could tell the fix as implemented has only partially worked.

### Relates
- #4701
- #4709

### DIFF
--- a/sphinx/ext/viewcode.py
+++ b/sphinx/ext/viewcode.py
@@ -253,7 +253,7 @@ def collect_pages(app: Sphinx) -> Generator[tuple[str, dict[str, Any], str], Non
             lexer = env.config.highlight_language
         else:
             lexer = 'python'
-        linenos = 'inline' * env.config.viewcode_linenumbers
+        linenos = 'inline' * env.config.viewcode_line_numbers
         highlighted = highlighter.highlight_block(code, lexer, linenos=linenos)
         lines = highlighted.splitlines()
         # split off wrap markup from the first line of the actual code
@@ -267,7 +267,7 @@ def collect_pages(app: Sphinx) -> Generator[tuple[str, dict[str, Any], str], Non
             backlink = urito(pagename, docname) + '#' + refname + '.' + name
             lines[start] = (
                 '<a class="viewcode-block viewcode-back" ' +
-                'id="%s" href="%s">%s</a>' % (name, backlink, _('[docs]')) +
+                f'id="{name}" href="{backlink}">{_("[docs]")}</a>\n' +
                 lines[start])
 
         # try to find parents (for submodules)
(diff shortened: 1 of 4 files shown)


sphinx/ext/viewcode.py
```diff
@@ -166,25 +183,30 @@ def collect_pages(app):
...
         # now that we have code lines (starting at index 1), insert anchors for
-        # the collected tags (HACK: this only works if the tag boundaries are
-        # properly nested!)
-        maxindex = len(lines) - 1
+        # the collected tags
         for name, docname in used.items():
             type, start, end = tags[name]
             backlink = urito(pagename, docname) + '#' + refname + '.' + name
             lines[start] = (
-                '<div class="viewcode-block" id="%s"><a class="viewcode-back" '
```
tk0miya, line 271: This `<div>` tag is used for highlighting the source code when user came from description:
スクリーンショット 2019-04-26 1 15 02

I think this removal disables the highlights. This is not a bug fix.

sphinx/ext/viewcode.py
```diff
@@ -141,6 +141,23 @@ def missing_reference(app, env, node, contnode):
     return None
 
 
+def split_code_lines(lines):
```
tk0miya: I feel this code is a bit tricky. To be our code robust, we need to make a custom Formatter for pygments. I'll take a look in this weekend.

tk0miya (changes requested):

AA-Turner:
Thanks @benkrikler!

A

PR #6322, opened 2019-04-24
Deprecate EpubBuilder.esc() in favor of stdlib html.escape()

jdufresne:
Available since Python 3.2.

https://docs.python.org/3/library/html.html#html.escape

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,7 @@ Deprecated
 ----------
 
 * ``sphinx.builders.latex.LaTeXBuilder.apply_transforms()``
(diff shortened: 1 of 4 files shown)


tk0miya (changes requested):
+1: Reasonable.

Could you rebase this to 2.0 branch? It would be helpful if you'll add an entry to CHANGES and doc/extdev/deprecation.rst  :-)

jdufresne:
Done in the latest revision.

tk0miya (approved):
LGTM!

tk0miya:
Merged. Thank you always!

PR #6326, opened 2019-04-28
Refactor sphinx.highlighting

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Before enhance `sphinx.highlighting` module, this makes it simple
- This separates language determining logic to post transform.

### DIFF
--- a/sphinx/highlighting.py
+++ b/sphinx/highlighting.py
@@ -32,6 +32,7 @@ if False:
     # For type annotation
     from typing import Any, Dict  # NOQA
     from pygments.formatter import Formatter  # NOQA
+    from pygments.style import Style  # NOQA
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


sphinx/transforms/post_transforms/code.py
```diff
@@ -96,21 +103,61 @@ def visit_literal_block(self, node):
             node['linenos'] = (lines >= setting.lineno_threshold - 1)
 
 
+class HighlightLanguageDetector(SphinxTransform):
```
tk0miya: To determine language on post transform phase is good idea. But this implementation conceals what language is specified by user. On highlighting, highlighter does not emit warnings for `default` language even if error. We should consider such case.

tk0miya: I'll try this in another PR.

PR #6334, opened 2019-05-03
Removing broken links

SuryaGrandhi:
Using sphinx with Read the Docs, the Getting Started link has been updated.

Subject: Update links to Read the Docs.

### Feature or Bugfix
- Bugfix

### Purpose
- While developers use this for the first time they again need to google rtd. Now that we have updated it would be clearer.

### Detail
- The Getting Started link is broken.

### Relates
- Observation

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -51,11 +51,11 @@ Using Sphinx with...
 --------------------
 
 Read the Docs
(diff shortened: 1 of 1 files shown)


tk0miya:
Thanks!

Issue #6336, opened 2019-05-03
Not all dependencies are installed from an editable pip install of sphinx in a virtualenv


jfbu:
Creating a fresh virtualenv and then executing

` ` ` 
pip install -e .
` ` ` 

from a git clone of sphinx-doc/sphinx dev repo, some dependencies are not installed for a fully functional install:

- sphinxcontrib-websupport,

- graphviz

Both are needed when one builds the doc from sphinx/doc/.

Absence of graphviz causes a failed `make latexpdf` due to graphics inclusion of non-existent "None" file. This comes from documentation of [`sphinx.ext.inheritance-diagram`](http://www.sphinx-doc.org/en/master/usage/extensions/inheritance.html)

Also

- pytest

- html5lib

are not automatically installed. But there are needed for running test with make. Maybe this is anyhow deprecated from using `tox`?

jfbu:
(I do understand graphviz is not a Python dependency .... :) )

tk0miya:
About latter dependencies, you can install them with `pip install -e '.[test]'` command. They're defined as "extra" dependencies tagged by "test". So you need to give `[test]` parameter to pip to install them all.
https://github.com/sphinx-doc/sphinx/blob/b0d070c8572c8316bf72889d6b04b95e6ce905bf/setup.py#L36-L50

On the other hand, former one is not defined yet. So it would be better to define them on setup.py soon. But, I don't have good idea to describe dependencies for non python packages like graphviz. bindep.txt might help us.

jfbu:
> About latter dependencies, you can install them with `pip install -e '.[test]'` command.

Works for me thanks.

tk0miya:
I just added extra dependency named "docs" in #6337.
Closing.

PR #6337, opened 2019-05-04
Add extra requirements for docs (refs: #6336)

tk0miya:
### Feature or Bugfix
- Packaging

### Purpose
- refs: #6336

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,9 @@ extras_require = {
     ':sys_platform=="win32"': [
         'colorama>=0.3.5',
     ],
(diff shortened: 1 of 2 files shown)


jfbu (approved):
LGTM, I had to guess (I am a slow-learner) that one needed ` pip install -e '.[docs]'` to get sphinxcontrib-websupport installed...

tk0miya:
Yes, websupport module was unbundled since 2.0 release. So we need to install it as extra dependency. This defines the extra dependency explicitly.

PR #6339, opened 2019-05-04
imgmath: Replace utf8x by utf8 in templates for latex snippets

jfbu:
Follow-up to #6308 (and #6310).

### DIFF
--- a/sphinx/templates/imgmath/preview.tex_t
+++ b/sphinx/templates/imgmath/preview.tex_t
@@ -1,5 +1,5 @@
 \documentclass[12pt]{article}
-\usepackage[utf8x]{inputenc}
+\usepackage[utf8]{inputenc}
 \usepackage{amsmath}
 \usepackage{amsthm}
 \usepackage{amssymb}
--- a/sphinx/templates/imgmath/template.tex_t
+++ b/sphinx/templates/imgmath/template.tex_t
@@ -1,5 +1,5 @@
 \documentclass[12pt]{article}
-\usepackage[utf8x]{inputenc}
+\usepackage[utf8]{inputenc}
 \usepackage{amsmath}
 \usepackage{amsthm}
 \usepackage{amssymb}


sphinx/templates/imgmath/preview.tex_t
```diff
@@ -1,5 +1,5 @@
 \documentclass[12pt]{article}
-\usepackage[utf8x]{inputenc}
+\usepackage[utf8]{inputenc}
 \usepackage{amsmath}
 \usepackage{amsthm}
 \usepackage{amssymb}
```
jfbu, line 5: There follows `\usepackage{amsfonts}` line, but `amssymb` loads `amsfonts` so next line could be removed.

sphinx/templates/imgmath/template.tex_t
```diff
@@ -1,5 +1,5 @@
 \documentclass[12pt]{article}
-\usepackage[utf8x]{inputenc}
+\usepackage[utf8]{inputenc}
 \usepackage{amsmath}
 \usepackage{amsthm}
 \usepackage{amssymb}
```
jfbu, line 5: There follows `\usepackage{amsfonts}` line, but `amssymb` loads `amsfonts` so next line could be removed.

tk0miya (approved):
LGTM. I wonder 2.0 branch is better for this.

sphinx/templates/imgmath/preview.tex_t
```diff
@@ -1,5 +1,5 @@
 \documentclass[12pt]{article}
-\usepackage[utf8x]{inputenc}
+\usepackage[utf8]{inputenc}
```
jfbu, line 2: The `utf8x` seems legacy of initial imgmath extension. I am not sure what it is needed for, as here it is only matter of math mode. AFAICT only input of the type `\text{...}` (or `\textrm{...}`)  could potentially need it but then non-ascii characters will cause problems from absence of any fontencoding specification in the preamble. Here text will use OT1 LaTeX fontencoding (with Computer Modern) and non-ascii characters needing utf8x will not display correctly. For (silly) example if one try `\text{π}` inside math, one will get

```
Undefined control sequence.
\u-default-960 #1->\textpi 
                           
l.12 $\text{π}
               $
```

because no suitable fontencoding is loaded which would define  the `\textpi` macro. One would need an extra `\usepackage[LGR]{fontenc}` because `\textpi` is provided by `LGR` latex font encoding. SImilarly for any other non-ascii thing. And furthermore LaTeX is very dumb and `\usepackage[LGR,T1]{fontenc}` would make it fail again, because it requires from user to manually use `\fontencoding{..}\selectfont`... (one would need things like `alphabeta/substitutefont` package as done by Sphinx for latex target). (sorry for long explanations)

Also nowadays utf8x is old, and only utf8 is actively maintained (directly by LaTeX team; since 2018/04/01 LaTeX release it is even done by default automatically.)

What do you think @xuhdev? was there any reason for using `utf8x` initially?

jfbu (commented):
ping @xuhdev

jfbu:
> LGTM. I wonder 2.0 branch is better for this.

I initially started on 2.0 branch but shifted to master which is where #6310 got merged. But I thin you are right. Should I rebase to 2.0?

tk0miya:
ah, you're right. I'm confusing the branches.

jfbu:
This could indeed be merge to 2.0 because #6308 got merged there. The #6310 is independent. What is your opinion?

sphinx/templates/imgmath/preview.tex_t
jfbu: Actually my explanations on why utf8x is strange apply to utf8 itself also : for rendering any non-ascii character LaTeX (except xelatex, lualatex) needs suitable specification of font encoding; as here there is no such specification in this template, the `\usepackage[utf8]{inputenc}` has the same issues I described, AFAICT, of `\usepackage[utf8x]{inputenc}`, but at least it avoids extra rather big dependency (`ucs` package).

jfbu: (sorry for never ending review). A `\usepackage[utf8]{inputenc}` is not useless because it allows input of European letters such as `\text{à}` in math mode. Indeed LaTeX will know how to convert this into using some accent as in ``\`a`` primitive TeX mark-up. But as explained prior more exotic non-ascii letters will not go through in general.

xuhdev: I'm not sure why it is there. It was actually put in the predecessor of imgmath, i.e., pngmath: c41f516ddaa3b083e5aa5a5c982b2a7cc4819eb7

@birkenfeld is probably the right person to ask.

tk0miya:
>This could indeed be merge to 2.0 because #6308 got merged there. The #6310 is independent. What is your opinion?

According to our branch model, backward compatibility is required for all changes in 2.0 branch. So I think it would better to merge this to 2.0 branch if enough stable. But, if this change depends on other breaking changes, it should be merged into master branch.

I suppose this is not breaking change. So 2.0 branch is appropriate for this.

tk0miya:
Oh, I'd forgotten to mention about #6310. It seems not a breaking change. So we can cherry-pick it into 2.0 branch :-)

jfbu:
Thanks for branch model summary, I have cherry-picked it and modified target branch to 2.0. Will merge after CI completes.

sphinx/templates/imgmath/preview.tex_t
jfbu: Thanks @xuhdev. I don't see what `utf8x` brought even back then, perhaps `utf8` from LaTeX2e team had some limitations, but as I explained input encoding in pdfLaTeX can only work if matched to suitable font encodings. Perhaps @birkenfeld will comment later on.

PR #6344, opened 2019-05-06
Fix #6311: autosummary: autosummary table gets confused by complex type hints

### LINKED ISSUE #6311: autosummary table gets confused by long / complex type hints
**Describe the bug**

While creating the table for `.. autosummary::`, the function

https://github.com/sphinx-doc/sphinx/blob/15bc5a32bb0b78b432803e55fffa10c801182c75/sphinx/ext/autosummary/__init__.py#L430

will not produce the expected results, some type hints remain, albeit broken up.  Here is the long function signature in question:

```py
from pathlib import Path
# ...
def filter_file(path: Union[Path, str], pattern: str,
                repl: Union[Callable[[Match], str], str], count: int = 0,
                flags: int = 0, backup_extension: str = ".orig",
                line_based: bool = False, demand_different: bool = True,
                encoding: Optional[str] = None) -> Path:
```

In the `.. autosummary::` table, the following is produced:

```
filter_file(path, str], pattern, repl, str], …)
```

The rendered docs can be found [online here](https://ci-exec.readthedocs.io/en/latest/api/patch.html#module-ci_exec.patch), but the root of the problem can be seen by the `path` and `repl` parameters (first and third parameter).

Stepping through, what actually happens is that everything just gets split on `', '`:

https://github.com/sphinx-doc/sphinx/blob/15bc5a32bb0b78b432803e55fffa10c801182c75/sphinx/ext/autosummary/__init__.py#L448-L454

So when we have something like `Union[Path, str]` or the (rather ugly) more complicated `Union[Callable[[Match], str], str]`, splitting on `', '` will give us something like this (splitting each argument on line to make clear):

```py
(Pdb) p args
[
    'path: Union[pathlib.Path',
    'str]',
    'pattern: str',
    'repl: Union[Callable[[Match[AnyStr]]',
    'str]',
    # ... remaining arguments ...
]
```

This is actually a pretty hard problem to solve with regular expressions, if it's even possible to do so (unless I'm missing something, it's kind of late xD).  I think we would have to start counting how many `[` we have seen and match them with `]` (after having seen a `:`).

So I was playing around with alternatives, found a nice SO answer that has some relevant stuff, here's a rough prototype:

```py
#!/usr/bin/env python3

import ast
import re
class HintDestroyer(ast.NodeTransformer):
    # See: https://stackoverflow.com/a/42734810/3814202
    # we only need to visit function since we only create function
    def visit_FunctionDef(self, node):
        node.returns = None
        if node.args.args:
            for arg in node.args.args:
                # just for a glimpse ...
                print("Removing annotation for {}: {}".format(arg.arg, arg.annotation.__dict__))
                arg.annotation = None
        return node

# sample input that breaks
sig = "(path: Union[pathlib.Path, str], pattern: str, repl: Union[Callable[[Match[AnyStr]], str], str], count: int = 0, flags: int = 0, backup_extension: str = '.orig', line_based: bool = False, demand_different: bool = True, encoding: Optional[str] = None) -> pathlib.Path"

# create a valid function signature to parse with ast module.
function = "def foo{sig}: pass\n".format(sig=sig)

# parse and transform
parsed_function = ast.parse(function)
transformed = HintDestroyer().visit(parsed_function)
# we only created a single "function", there should be a single module with a
# function as its body.
transformed_function = transformed.body[0]

# is this what we want?
args = ", ".join([item.arg for item in transformed_function.args.args])
print(args)
# output:
# path, pattern, repl, count, flags, backup_extension, line_based, demand_different, encoding
```

Right now I _believe_ what `args` and `opts` in the current autosummary code are doing is trying to shove `opts` at the end _if there is space_ (since there's a maximum length here).  I'm not sure why this happens, since all of the keyword arguments have to be last anyway right?

Basically we can use `ast` and just limited join on the transformed thing if we don't care about necessarily preserving existing behavior.  But I'm not sure about the implications of all this ... so thought I would open up the issue for discussion of other possible solutions or maybe downsides to use `ast` (it definitely doesn't come for free, but it's also a simple parse).

**To Reproduce**

```
$ git clone https://github.com/svenevs/ci_exec.git
$ cd ci_exec
$ tox -e docs
$ open .tox/docs/tmp/html/index.html

# ... or ...

$ pip install -r docs/requirements.txt
$ cd docs
$ make html 
$ open _build/html/index.html
```

**Expected behavior**
No stray `str]` end up as a "parameter" in the autosummary table.

**Your project**
https://github.com/svenevs/ci_exec

**Screenshots**

over_parse

**Environment info**
- OS: linux
- Python version: 3.7.2
- Sphinx version: 2.0.1
- Sphinx extensions:  sphinx.ext.autosummary
- Extra tools: n/a

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6311

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -95,6 +95,7 @@ Bugs fixed
 * commented term in glossary directive is wrongly recognized
 * #6299: rst domain: rst:directive directive generates waste space
 * #6331: man: invalid output when doctest follows rubric
+* #6311: autosummary: autosummary table gets confused by complex type hints
 
 Testing
 --------
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -441,6 +441,10 @@ def mangle_signature(sig, max_chars=30):
     s = re.sub(r"\\'", "", s)
     s = re.sub(r"'[^']*'", "", s)
 
+    # Strip arguments of typehints (ex. Tuple[int, str] -> Tuple)
+    while re.search(r'\[[^],][^\[\]]*\]', s):
+        s = re.sub(r'\[[^],][^\[\]]*\]', '', s)
+
     # Parse the signature to arguments + options
     args = []  # type: List[str]
     opts = []  # type: List[str]
--- a/tests/test_ext_autosummary.py
+++ b/tests/test_ext_autosummary.py
@@ -49,6 +49,8 @@ def test_mangle_signature():
     (a=1, b=2, c=3) :: ([a, b, c])
     (a=1, b=<SomeClass: a, b, c>, c=3) :: ([a, b, c])
     (a: int, b: int) -> str :: (a, b)
+    (a: Tuple[int, str], b: int) :: (a, b)
+    (a: Union[Tuple[int, str], str], b: int) :: (a, b)
     """
 
     TEST = [[y.strip() for y in x.split("::")] for x in TEST.split("\n")


svenevs:
This is close, but does not work on nested `[[]]`.

Before (latest sphinx): 

before

After (this branch):

after

Test case to add: something with `Callable` parametrized maybe?  Annotation in question is `repl: Union[Callable[[Match], str], str]`.

tk0miya:
OMG. It's tough work... I'll try it tomorrow.

svenevs:
Haha seriously. I tried a couple of times but your regex-fu is way better than mine. So if this doesn't end up working out, a more expensive but easier to reason about approach could be considered. Something like counting number of `[` vs `]` seen so far, continue until the counts match (regardless of how many `,` have been seen, because that's the underlying dilemma).

I'll see if I can get a dumb version working for the as a backup tonight.

tk0miya:
I think #6353 is better than this. So I'm withdrawing this.

PR #6353, opened 2019-05-12
Fix #6311: autosummary: autosummary table gets confused by complex type hints

### LINKED ISSUE #6311: autosummary table gets confused by long / complex type hints

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6311

### DIFF
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -329,7 +329,12 @@ class Autosummary(SphinxDirective):
 
             # -- Grab the signature
 
-            sig = documenter.format_signature()
+            try:
+                sig = documenter.format_signature(show_annotation=False)
+            except TypeError:
+                # the documenter does not support ``show_annotation`` option
+                sig = documenter.format_signature()
+
             if not sig:
                 sig = ''
             else:
@@ -445,7 +450,7 @@ def mangle_signature(sig, max_chars=30):
     args = []  # type: List[str]
     opts = []  # type: List[str]
 
-    opt_re = re.compile(r"^(.*, |)([a-zA-Z0-9_*]+)=")
+    opt_re = re.compile(r"^(.*, |)([a-zA-Z0-9_*]+)\s*=\s*")
     while s:
         m = opt_re.search(s)
         if not m:
--- a/CHANGES
+++ b/CHANGES
@@ -95,6 +95,7 @@ Bugs fixed
 * commented term in glossary directive is wrongly recognized
 * #6299: rst domain: rst:directive directive generates waste space
 * #6331: man: invalid output when doctest follows rubric
+* #6311: autosummary: autosummary table gets confused by complex type hints
 
 Testing
 --------
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -365,8 +365,8 @@ class Documenter:
             return False
         return True
 
-    def format_args(self):
-        # type: () -> str
+    def format_args(self, **kwargs):
+        # type: (Any) -> str
         """Format the argument signature of *self.object*.
 
         Should return None if the object does not have a signature.
@@ -385,8 +385,8 @@ class Documenter:
         # directives of course)
         return '.'.join(self.objpath) or self.modname
 
-    def format_signature(self):
-        # type: () -> str
+    def format_signature(self, **kwargs):
+        # type: (Any) -> str
         """Format the signature (arguments and return annotation) of the object.
 
         Let the user process it via the ``autodoc-process-signature`` event.
@@ -397,7 +397,13 @@ class Documenter:
         else:
             # try to introspect the signature
             try:
-                args = self.format_args()
+                try:
+                    args = self.format_args(**kwargs)
+                except TypeError:
+                    warnings.warn("Documenter.format_args() takes keyword arguments "
+                                  "since Sphinx-2.1",
+                                  RemovedInSphinx40Warning)
+                    args = self.format_args()
             except Exception as err:
                 logger.warning(__('error while formatting arguments for %s: %s') %
                                (self.fullname, err), type='autodoc')
@@ -953,15 +959,15 @@ class DocstringSignatureMixin:
             return lines
         return super().get_doc(None, ignore)  # type: ignore
 
-    def format_signature(self):
-        # type: () -> str
+    def format_signature(self, **kwargs):
+        # type: (Any) -> str
         if self.args is None and self.env.config.autodoc_docstring_signature:  # type: ignore
             # only act if a signature is not explicitly given already, and if
             # the feature is enabled
             result = self._find_signature()
             if result is not None:
                 self.args, self.retann = result
-        return super().format_signature()  # type: ignore
+        return super().format_signature(**kwargs)  # type: ignore
 
 
 class DocstringStripSignatureMixin(DocstringSignatureMixin):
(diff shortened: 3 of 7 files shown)


sphinx/ext/autosummary/__init__.py
```diff
@@ -329,7 +329,12 @@ def get_items(self, names):
 
             # -- Grab the signature
 
-            sig = documenter.format_signature()
+            try:
+                sig = documenter.format_signature(show_annotation=False)
```
tk0miya, line 333: But this handles `TypeError` to support custom documenters. So I believe this does not break compatibility.

svenevs:
Just wanted to drop a quick note saying

1. Thank you!!!  This looks like a much better resolution.
2. I can test it either late tonight or tomorrow (mother's day here in the states :))

svenevs:
Just tested it out, works perfectly!!!  Very cool :slightly_smiling_face:

tk0miya:
Thank you for verifying!

svenevs:
DISREGARD!  I got confused.

~There seem to have been bad merge?~

- ~[merge 2.0 into 6311_autosummary_confused_by_complex_typehints2](https://github.com/sphinx-doc/sphinx/commit/59d0f076ec2bd50ad0e59ce81e1fef1e4c1a9cd1)~
- ~[merge branch back to 2.0](https://github.com/sphinx-doc/sphinx/commit/774c59821c3c71c9aaac63f51d957264d30311ab)~

~Right now if I `pip install git+https://github.com/sphinx-doc/sphinx.git@2.0` things don't work again, but they worked when I tested before 2.0 was merged to 6311_... branch (now deleted).  Steps to replicate:~

```console
sven:~/Desktop> python -m venv venv
sven:~/Desktop> source venv/bin/activate
(venv) sven:~/Desktop> git clone https://github.com/svenevs/ci_exec.git
Cloning into 'ci_exec'...
remote: Enumerating objects: 126, done.
remote: Counting objects: 100% (126/126), done.
remote: Compressing objects: 100% (80/80), done.
remote: Total 253 (delta 76), reused 80 (delta 44), pack-reused 127
Receiving objects: 100% (253/253), 102.80 KiB | 1.09 MiB/s, done.
Resolving deltas: 100% (134/134), done.
(venv) sven:~/Desktop> cd ci_exec/docs/
(venv) sven:~/Desktop/ci_exec/docs (master)> pip install git+https://github.com/sphinx-doc/sphinx.git@2.0
Collecting git+https://github.com/sphinx-doc/sphinx.git@2.0
  Cloning https://github.com/sphinx-doc/sphinx.git (to revision 2.0) to /tmp/pip-req-build-i21gi6uc
   ...
(venv) sven:~/Desktop/ci_exec/docs (master)> pip install -r requirements.txt
...
(venv) sven:~/Desktop/ci_exec/docs (master)> make html
...
(venv) sven:~/Desktop/ci_exec/docs (master)> open build/html/api/patch.html
```

~Autosummary table at the top of that page is not doing what it did when I tested this :/  You can see similar problems on the autosummary at the top of other modules there too.~

merge_regression

~I will re-review the diffs again, I combed through it for about an hour but don't see what went wrong.  Will need fresh eyes...~

svenevs:
OMG I'm so sorry @tk0miya as soon as I started tracing through `mangle_signature` I realized that the `[, count, ...]` is intended behavior, these are the optional parameters as denoted by the `[...]`.

Thanks again for fixing this, sorry for getting confused about this behavior!

tk0miya:
:-)

PR #6364, opened 2019-05-14
Fix latex figure in admonition

wallento:
Subject: Fix latex figure in admonition

### Feature or Bugfix
- Bugfix

### Purpose
- Fix usage of figures in admonitions for the LaTeX writer.

### Detail
- Using figures in an admonition produces a LaTeX error ("Not in outer par mode."). This is because it generates a float in a float. This can be trivially fixed by overwriting the alignment to H, which is also what is most probably intended.


tests/roots/test-latex-figure-in-admonition/conf.py
```diff
@@ -0,0 +1,3 @@
+master_doc = 'index'
```
tk0miya: By default, `master_doc` equals `'index'`. So this is not needed here.

sphinx/writers/latex.py
```diff
@@ -1680,20 +1684,24 @@ def depart_legend(self, node):
     def visit_admonition(self, node):
         # type: (nodes.Element) -> None
         self.body.append('\n\\begin{sphinxadmonition}{note}')
+        self.in_box += 1
```
tk0miya: Is `in_box` int value? It is defined as boolean in `__init__()`.

tests/test_build_latex.py
```diff
@@ -1389,6 +1389,12 @@ def test_latex_labels(app, status, warning):
     assert result.count(r'\label{\detokenize{index:section1}}') == 1
 
 
+@pytest.mark.sphinx('latex', testroot='latex-figure-in-admonition')
+def test_latex_figure_in_admonition(app, status, warning):
+    app.builder.build_all()
+    compile_latex_document(app, 'python.tex')
```
tk0miya: Could you check how `.tex` code is generated here? Compiling is too expensive for testing.

tk0miya (changes requested):

tests/roots/test-latex-figure-in-admonition/conf.py
wallento: Of course, removed

tests/test_build_latex.py
wallento: Yes, I changed it accordingly. The compilation shows the original error. I have changed it now to be a regression. Thanks!

wallento:
Thanks, I resolved the comments, amended and force pushed.

sphinx/writers/latex.py
```diff
@@ -1640,7 +1644,7 @@ def visit_figure(self, node):
             self.context.append('\\end{center}\n')
         else:
             self.body.append('\n\\begin{figure}[%s]\n\\centering\n' %
-                             self.elements['figure_align'])
+                             align)
```
tk0miya: Could you concatenate this line with above one? No reason to fold a line.

wallento: Yes, amended and force-pushed

tk0miya (approved):

tk0miya:
LGTM. I'm okay to merge this if @jfbu approve this.

CHANGES
```diff
@@ -18,6 +18,7 @@ Bugs fixed
 
 * #6286: C++, allow 8 and 9 in hexadecimal integer literals.
 * #6305: Fix the string in quickstart for 'path' argument of parser
+* LaTeX: Figures in adminitions produced errors (float in float)
```
jfbu: typo: s/adminitions/admonitions

Also "float in float" is not technically the exact LaTeX problem so it should be suppressed.

sphinx/writers/latex.py
```diff
@@ -473,6 +473,7 @@ def __init__(self, document, builder):
         self.in_term = 0
         self.needs_linetrimming = 0
         self.in_minipage = 0
+        self.in_box = 0
```
jfbu: Although indeed the sphinx.sty LaTeX macros use "sphinxlightbox" and "sphinxheavybox" terminology, in the former case (note, hint, important, tip), the contents are not encapsulated in a box, and there was no crash in those cases. However it does make sense to prohibit floating figures also for them. Perhaps `no_latex_floats` (or `no_floating_figures` as Sphinx uses LaTeX floats only for figures) would be better than `in_box`.

sphinx/writers/latex.py
```diff
@@ -1680,20 +1683,24 @@ def depart_legend(self, node):
     def visit_admonition(self, node):
         # type: (nodes.Element) -> None
         self.body.append('\n\\begin{sphinxadmonition}{note}')
+        self.in_box += 1
```
jfbu: as explained above, it does make sense to prohibit floating figures even if user has special custom LaTeX macros fro admonitions. But ``in_box`` is a bit confusing because it is not true that default admonition types use TeX boxes : only `warning`, `caution`, ... do.

jfbu (approved):
LGTM with nits! (sorry for late review, thanks for catching this problem)

- I made some comments on ``in_box`` which is slightly confusing, because not all admonitions by default do use a TeX encapsulation box, only the "heavy" types; besides the user can renew the environments to do something else. But in all cases it seems a priori OK to prohibit floating figures, even if not causing build crash. Thus I proposed something like ``no_latex_floats`` as better name.

- there is a missing EOL at end of the conf.py of the test file.

OK to merge, naming issue can be modified later.

wallento:
Thanks @jfbu. I have amended the fixes and force-pushed.

jfbu (approved):
LGTM!

jfbu:
Thanks for your contribution, sorry for late reviewing. Merging.

Issue #6371, opened 2019-05-15
Allow suppressing warnings for autosectionlabel


gibfahn:
Raising an issue as per the suggestion in https://github.com/sphinx-doc/sphinx/pull/2346#issuecomment-261267441 , adding a `type` and `subtype` to this warning would be helpful:

https://github.com/sphinx-doc/sphinx/blob/ba86f4ea7895ff59368a602d43bb85fc4d453546/sphinx/ext/autosectionlabel.py#L56-L58

The reason I need this is that I would like to autogenerate section labels to resolve https://github.com/rtfd/recommonmark/issues/155.

This works well but some files have duplicate headings, which I can resolve by manually specifying a section. So to run with `-W` I need to be able to ignore the warnings.

tk0miya:
Closing now. Thanks,

PR #2346, opened 2016-02-22
Add :confval:`suppress_warnings` to suppress arbitrary warning message

tk0miya:
This enables sphinx to suppress arbitrary warning by `suppress_warnings` configuration.

With this, all of warnings can be classified with `type` and `subtype`. For example, warnings for options are classified as `type=ref` and `subtype=option`.
And, users are enable to suppress these warnings with following settings:

```
supress_warnings = ["ref.option"]
```

This might help the situations of #2229.
At same time, I believe this is better implementation of #1969

PR #6373, opened 2019-05-15
autosectionlabel: add type for suppress_warnings

gibfahn:
This sets the subtype to the filename, which means that you can suppress warnings for individual paths.

For example for the following warning:

```
path/to/file.rst:5547: WARNING: duplicate label path/to/file:label, other instance in /path/to/file.rst
```

You would add this to `config.py` to suppress all warnings:

```python
extensions = [
    'sphinx.ext.autosectionlabel',
]

autosectionlabel_prefix_document = True

suppress_warnings = [
    'autosectionlabel.*',
]
```

To suppress warnings for the individual file path you would do:

```python
suppress_warnings = [
    'autosectionlabel.path/to/file',
]
```

Fixes: https://github.com/sphinx-doc/sphinx/issues/6371

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -99,6 +99,7 @@ Features added
 * #6306: html: Add a label to search form for accessability purposes
 * #6358: The ``rawsource`` property of ``production`` nodes now contains the
   full production rule
+* #6373: Allow suppression of autosectionlabel warnings
 
 Bugs fixed
 ----------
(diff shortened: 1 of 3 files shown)


gibfahn:
@tk0miya I mostly improvised this based on your work in https://github.com/sphinx-doc/sphinx/pull/2346, if you think this should be done a different way please let me know!

~Also maybe I should change the target branch to [`2.0`](https://github.com/sphinx-doc/sphinx/tree/2.0)?~ Done

I should probably add a test somewhere too, not sure where the best place for that would be. I did test this manually on my project and it does seem to work.

tk0miya (approved):
LGTM!

tk0miya:

PR #6386, opened 2019-05-21
Enable doclinter by default on CI

tk0miya:
### Feature or Bugfix
- Refactoring


utils/doclinter.py
```diff
@@ -14,21 +14,39 @@
 from typing import List
 
 
-MAX_LINE_LENGTH = 100
+MAX_LINE_LENGTH = 90
```
tk0miya, line 17: It might be better to restrict by 80 columns. But much updates is needed to do that. So I set this to 90 for now.

Issue #4318, opened 2017-12-18
The toctree ``:numbered:`` option has no effect in PDF


jfbu:
#### Procedure to reproduce the problem
File index.rst

```
====================
Test Numbered Levels
====================

.. toctree::
   :numbered:

   page
```

File page.rst

```
==============
Top sectioning
==============

level 2 sectioning
==================

level 3 sectioning
------------------

level 4 sectioning
~~~~~~~~~~~~~~~~~~

level 5 sectioning
^^^^^^^^^^^^^^^^^^

Hello
```

#### Error logs / results

HTML output is numbered all the way, PDF via ``make latexpdf`` only down to level 3.

#### Expected results

PDF also obeys the ``:numbered:`` option, inclusive of its optional parameter.

### Environment info
- OS:  Mac
- Python version: 3.6.3
- Sphinx version: 1.6.3 and master

tk0miya:
I have a plan (idea?) to restructure ToC structure of Sphinx in future. Personally, I and friends wrote a Japanese book for Sphinx beginners in this year. But current Sphinx can't represent prefaces and appendixes.
And I know some users uses Sphinx to write books and thesis. So it would be nice to enhance ToC tree structure for the purpose.

I don't know when we can modify it (1.8 or 2.x?). But I feel the modification is related with this. So I notice you my idea :-)

jfbu:
Thanks for sharing your idea ;-) 

Recently I needed to produce a pdf with appendices. I solved this via some bit of LaTeX hacking in particular some raw LaTeX.

In case people come across this, here is how it looks (I leave some extras for PDF bookmarks). But this is really a hack I needed to get the job done quickly, perhaps there is easier way.

```
latex_docclass = {'howto': 'article', 'manual': 'book'} # and we use 'manual' -> book class now

latex_elements = {
    # some other stuff, and

    'fncychap': '\\usepackage[Bjornstrup]{fncychap}',

    'preamble': r'''
\usepackage{bookmark}% for customization of PDF bookmarks
\makeatletter
\renewcommand{\sphinxtableofcontents}{% custom to control \pagestyle usage and page numbering
  \pagenumbering{roman}%
  \pagestyle{plain}%
  \begingroup
    \parskip \z@skip
    \pdfbookmark[1]{Table des matières}{MAINTOC}%
    \tableofcontents
  \endgroup
  \ifdefined\fancyhf\pagestyle{normal}\fi
}
[...]

PR #6401, opened 2019-05-27
Fix #4390: html: Consistent and semantic CSS for signatures

### LINKED ISSUE #4390: Consistent and semantic CSS for signatures
## Subject: Consistent and semantic CSS for signatures

I propse to introduce a consistent set of semantic CSS classes for easier styling of the signature.

### Current state

Example signature:
<dt id="matplotlib.axes.Axes.vlines">
<code class="descclassname">Axes.</code>
<code class="descname">vlines</code>
<span class="sig-paren">(</span>
<em>x</em>, <em>y</em>, <em>linestyles='solid'</em>
<span class="sig-paren">)</span>

The corresponding HTML looks like this:

~~~
<dt id="matplotlib.axes.Axes.vlines">
<code class="descclassname">Axes.</code>
<code class="descname">vlines</code>
<span class="sig-paren">(</span>
<em>x</em>, <em>y</em>, <em>linestyles='solid'</em>
<span class="sig-paren">)</span>
~~~

### Proposed change

Use this set of semantic CSS classes:
- sig
  - sig-prename
  - sig-name
  - sig-paren
  - sig-param

These could simply be added to maintain compatibility with existing CSS files:

~~~
<dt id="matplotlib.axes.Axes.vlines" class="sig">
<code class="sig-prename descclassname">Axes.</code>
<code class="sig-name descname">vlines</code>
<span class="sig-paren">(</span>
<em class="sig-param">x</em>, <em class="sig-param">y</em>, <em class="sig-param">linestyles='solid'</em>
<span class="sig-paren">)</span>
~~~

In the long run `descclassname` and `descname` could be deprecated. Also one could debate if one should change the tags `<code>` and `<em>` to `<span>`, but I have no strong opinion on this.

This scheme can be extend when more complex signatures arise, as proposed in #4289. For example one could add classes `sig-paramname` and `sig-paramvalue`.

### LINKED ISSUE #5868: python: Improve formatting of signatures with type hints
Python functions with type hint annotations often have very long signatures. The arguments in these signatures often have internal spaces, so naive line-wrapping produces poor results with lines broken in the middle of an argument instead of between arguments. 

For example, see [tornado.httpclient.HTTPRequest](https://www.tornadoweb.org/en/latest/httpclient.html#tornado.httpclient.HTTPRequest):

image

This is kind of a monster no matter what you do with it, but it would be more legible if it were formatted with one parameter per line [as in the source code](https://github.com/tornadoweb/tornado/blob/9ff87e3ad62e8278f4f595509ba2226cede35417/tornado/httpclient.py#L358-L393) (which is auto-formatted with `black`, FWIW). Or for less-intrusive changes, consider converting whitespace within an argument to non-breaking spaces (although I occasionally have [single arguments whose type annotations span multiple lines](https://github.com/tornadoweb/tornado/blob/9ff87e3ad62e8278f4f595509ba2226cede35417/tornado/httpserver.py#L152-L170)), or at least putting a `<span>` with a designated class around each argument so CSS could perhaps be used for styling the arguments.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #4318 

@timhoffm Could you check this please? I don't know this resolves #5868 case.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -101,6 +101,7 @@ Features added
 * rst domain: Add :rst:dir:`directive:option` directive to describe the option
   for directive
 * #6306: html: Add a label to search form for accessability purposes
+* #4390: html: Consistent and semantic CSS for signatures
 * #6358: The ``rawsource`` property of ``production`` nodes now contains the
   full production rule
 * #6373: autosectionlabel: Allow suppression of warnings
--- a/sphinx/writers/html5.py
+++ b/sphinx/writers/html5.py
@@ -117,7 +117,7 @@ class HTML5Translator(SphinxTranslator, BaseTranslator):
 
     def visit_desc_addname(self, node):
         # type: (nodes.Element) -> None
-        self.body.append(self.starttag(node, 'code', '', CLASS='descclassname'))
+        self.body.append(self.starttag(node, 'code', '', CLASS='sig-prename descclassname'))
 
     def depart_desc_addname(self, node):
         # type: (nodes.Element) -> None
@@ -141,7 +141,7 @@ class HTML5Translator(SphinxTranslator, BaseTranslator):
 
     def visit_desc_name(self, node):
         # type: (nodes.Element) -> None
-        self.body.append(self.starttag(node, 'code', '', CLASS='descname'))
+        self.body.append(self.starttag(node, 'code', '', CLASS='sig-name descname'))
 
     def depart_desc_name(self, node):
         # type: (nodes.Element) -> None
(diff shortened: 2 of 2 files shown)


timhoffm:
This goes definitively in the right direction (c.f. #4390).

For #5868 you could achieve at least different colors of the commas and the parameters. However more fine-grained control would be desireable: Separate spans for
- parameter name
- default value
- type hint

AFAICS `param_name: type = default` is currently a single element. This would require splitting the parameter further in the parser.

tk0miya:
I agree that more fine-grained control is better. But goal of #5868 is folding long arguments by each argument, I think. So the separation is not filed to issues yet. Could you file an issue please?

PR #6408, opened 2019-05-29
Fix the ranking of search results

jia-kai:
### Feature or Bugfix
- Bugfix

### Purpose

The ranking of search results in `searchtools.js` depends on calculated matching score. The matching is checked by `fullname.toLowerCase().indexOf(object)`, however the scoring for matched object names is based on original `fullname` rather than lower-cased `fullname`. For example, this would casuse the result `some.package.FooBar` to be ranked very low in a search query of `FooBar`.

This PR fixes such behavior by computing match scores based on lower-cased `fullname`.


tk0miya:

PR #6419, opened 2019-05-31
ENH: Explain rebuilds

larsoner:
Subject: Explain why rebuilds are being triggered

### Feature or Bugfix
- Feature

### Purpose
Give reasons why a full build is being done.

### Detail
On long sphinx builds, full rebuilds are painful. With this PR, the rebuild is explained in logging:
```
Running Sphinx v3.0.0+/2c826795b
loading pickled environment... Config changed (item numpydoc_xref_aliases)... done
[autosummary] generating autosummary for:
...
updating environment: (build configuration changed) 663 added, 0 changed, 0 removed
```
The "Config changed" and parenthetical for `updating environment` are new.

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -95,14 +95,15 @@ class BuildEnvironment:
     # --------- ENVIRONMENT INITIALIZATION -------------------------------------
 
     def __init__(self, app: "Sphinx" = None):
-        self.app = None             # type: Sphinx
-        self.doctreedir = None      # type: str
-        self.srcdir = None          # type: str
-        self.config = None          # type: Config
-        self.config_status = None   # type: int
-        self.events = None          # type: EventManager
-        self.project = None         # type: Project
-        self.version = None         # type: Dict[str, str]
+        self.app = None                  # type: Sphinx
+        self.doctreedir = None           # type: str
+        self.srcdir = None               # type: str
+        self.config = None               # type: Config
+        self.config_status = None        # type: int
+        self.config_status_extra = None  # type: str
+        self.events = None               # type: EventManager
+        self.project = None              # type: Project
+        self.version = None              # type: Dict[str, str]
 
         # the method of doctree versioning; see set_versioning_method
         self.versioning_condition = None  # type: Union[bool, Callable]
@@ -232,16 +233,25 @@ class BuildEnvironment:
     def _update_config(self, config: Config) -> None:
         """Update configurations by new one."""
         self.config_status = CONFIG_OK
+        self.config_status_extra = ''
         if self.config is None:
             self.config_status = CONFIG_NEW
         elif self.config.extensions != config.extensions:
             self.config_status = CONFIG_EXTENSIONS_CHANGED
+            extensions = sorted(
+                set(self.config.extensions) ^ set(config.extensions))
+            if len(extensions) == 1:
+                extension = extensions[0]
+            else:
+                extension = '%d' % (len(extensions),)
+            self.config_status_extra = ' (%r)' % (extension,)
         else:
             # check if a config value was changed that affects how
             # doctrees are read
             for item in config.filter('env'):
                 if self.config[item.name] != item.value:
                     self.config_status = CONFIG_CHANGED
+                    self.config_status_extra = ' (%r)' % (item.name,)
                     break
 
         self.config = config
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -388,9 +388,11 @@ class Builder:
             # ... but not those that already were removed
             changed.update(self.env.glob_toctrees & self.env.found_docs)
 
-        if changed:
-            reason = CONFIG_CHANGED_REASON.get(self.env.config_status, '')
+        if updated:  # explain the change iff build config status was not ok
+            reason = (CONFIG_CHANGED_REASON.get(self.env.config_status, '') +
+                      (self.env.config_status_extra or ''))
             logger.info('[%s] ', reason, nonl=True)
+
         logger.info(__('%s added, %s changed, %s removed'),
                     len(added), len(changed), len(removed))
 
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -7,6 +7,8 @@
     :copyright: Copyright 2007-2019 by the Sphinx team, see AUTHORS.
     :license: BSD, see LICENSE for details.
 """
+import os
+import shutil
 import pytest
 
 from sphinx.builders.html import StandaloneHTMLBuilder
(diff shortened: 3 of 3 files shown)


tk0miya:
Sorry for late reviewing.

Once, reading phase were consisted of both builder and environment. And it was hard to understand how building process works.

After some refactoring, builder only goes reading phase now. The environment object behaves like a database. According to the design, environment does not show any console messages (there are some exceptions, I know). So I'd like to move these implementations to `Builder.read()`.

I think outputting messages across components is hard to understand. Also hard to control. Especially, this PR outputs partial some messages. It makes hard to call this method when you like. For example, creating another instance of environment will output message. Is this expected behavior?

larsoner:
> According to the design, environment does not show any console messages (there are some exceptions, I know). So I'd like to move these implementations to Builder.read().

I can look to see how to do this. But it will probably require creating a new env property, is this okay? If so, any preferred name for it?

> For example, creating another instance of environment will output message. Is this expected behavior?

This part was intentional, yes. But I thought it was better to always explain why there is a rebuild (current behavior) than to only sometimes explain why there is a rebuild (i.e., if I got rid of the new environment message).

tk0miya:
>I can look to see how to do this. But it will probably require creating a new env property, is this okay? If so, any preferred name for it?

`config_diff` or `old_config`? I don't have good idea for it because I still don't understand why is the changed item needed.

>This part was intentional, yes. But I thought it was better to always explain why there is a rebuild (current behavior) than to only sometimes explain why there is a rebuild (i.e., if I got rid of the new environment message).

I think creating an instance of environment does not related with rebuild always. In future, it might be helpful for parallel build. This message adds restriction for Sphinx to create only one environment during build.
(We don't have any plan to do that at this moment. But I'd not like to add new restriction)

larsoner:
Okay I've refactored the code so that the printing is done in `Builder.read` and for:

- repeated/unchanged config, says nothing new (just says which are actually changed):
   ```
   updating environment: 0 added, 1 changed, 0 removed
   ```
- new config: says its a new config
   ```
   updating environment: [new config] 671 added, 0 changed, 0 removed
   ```
- changed extensions:
   ```
   updating environment: [extensions changed ('gen_commands')] 671 added, 0 changed, 0 removed
   ```
- changed config['env']:
   ```
   updating environment: [config changed ('numpydoc_xref_aliases')] 671 added, 0 changed, 0 removed
   ```

sphinx/environment/__init__.py
```diff
@@ -240,12 +242,20 @@ def _update_config(self, config):
...
                 if self.config[item.name] != item.value:
                     self.config_status = CONFIG_CHANGED
+                    self.config_status_extra = __(' (%r)') % (item.name,)
                     break
 
             # this value is not covered by the above loop because it is handled
             # specially by the config class
             if self.config.extensions != config.extensions:
                 self.config_status = CONFIG_EXTENSIONS_CHANGED
+                diff = sorted(
```
tk0miya: mypy warns for this variable.
```
sphinx/environment/__init__.py: note: In member "_update_config" of class "BuildEnvironment":
sphinx/environment/__init__.py:255:21: error: Incompatible types in assignment (expression has type "str", variable has type "List[str]")
sphinx/environment/__init__.py:257:21: error: Incompatible types in assignment (expression has type "str", variable has type "List[str]")
```

How about renaming this list to `diffs` or `extensions`?

sphinx/environment/__init__.py
```diff
@@ -240,12 +242,20 @@ def _update_config(self, config):
             for item in config.filter('env'):
                 if self.config[item.name] != item.value:
                     self.config_status = CONFIG_CHANGED
+                    self.config_status_extra = __(' (%r)') % (item.name,)
```
tk0miya: `__(...)` function is used for translation. Is this need to translate?

sphinx/environment/__init__.py
```diff
@@ -240,12 +242,20 @@ def _update_config(self, config):
...
             # specially by the config class
             if self.config.extensions != config.extensions:
                 self.config_status = CONFIG_EXTENSIONS_CHANGED
+                diff = sorted(
+                    set(self.config.extensions) ^ set(config.extensions))
+                if len(diff) == 1:
+                    diff = diff[0]
+                else:
+                    diff = '%d' % (len(diff),)
+                self.config_status_extra = __(' (%r)' % (diff,))
```
tk0miya: If you need to translate this, please wrap only format string by `__(...)` function? Our tooling extracts i18n strings from this declaration. So it would be better to surround only format string by `__(...)` function simply.

tk0miya (commented):
LGTM with nits.

I'm thinking about we should add a testcase for this or not. What do you think?

larsoner:
Comments addressed and tests added

tests/test_build_changes.py
```diff
@@ -32,6 +46,45 @@ def test_build(app):
...
+    # change a file
+    fname = os.path.join(app.srcdir, confoverrides['master_doc'] + '.rst')
+    with open(fname, 'a') as fid:
+        fid.write('\nfoo')
+    app2 = make_app(*args, confoverrides=confoverrides, **kwargs)
+    assert app2.env.config_status == CONFIG_OK
+    app2.build()
+    assert "0 added, 1 changed, 0 removed" in app2._status.getvalue()
+
+    # extensions changed (this check does not trigger properly...?)
```
larsoner: @tk0miya if you look I think these changes should trigger a CONFIG_EXTENSIONS_CHANGED but they trigger a CONFIG_NEW. You can see what I tried below. Instead of this `conf.py` writing approach, I also tried:
```
monkeypatch.setattr(sphinx.ext.todo, 'setup', lambda app: {'version': '0.00'})
```
Which seemed like it should work (detect changed version of extension) but it did not.

I'm not sure if this is a problem with the test fixtures, Sphinx, or my usage of them. But in any case the code is almost fully tested now, just the EXTENSIONS_CHANGED case is missing, which might be good enough for this PR.

tests/test_build_changes.py
```diff
@@ -9,12 +9,18 @@
...
+import os
 import pytest
+import sphinx
+from sphinx.environment import (CONFIG_CHANGED, CONFIG_NEW, CONFIG_OK,
+                                CONFIG_EXTENSIONS_CHANGED)
 
 
 @pytest.mark.sphinx('changes', testroot='changes')
-def test_build(app):
+def test_build(app, make_app, app_params):
```
tk0miya: I'm not sure why did you add testcase for this to testcase of changes builder; which generates CHANGELOG. Is this related with changes builder? If not so, please move this to new testcase to `test_build.py`

tests/test_build_changes.py
tk0miya: Good point. It seems there is a bug. I'll try to fix it.

tests/test_build_changes.py
```diff
@@ -32,6 +38,45 @@ def test_build(app):
...
+
+    # no changes
+    confoverrides = dict(master_doc=app0.env.config['master_doc'])
+    app1 = make_app(*args, confoverrides=confoverrides, **kwargs)
+    assert app0.env.config_status == CONFIG_OK
+    app1.build()
+    assert "0 added, 0 changed, 0 removed" in app1._status.getvalue()
+
+    # change a file
+    fname = os.path.join(app.srcdir, confoverrides['master_doc'] + '.rst')
```
tk0miya: Please use `index.rst` directly. I think concreteness increases readability of testcase.

tests/test_build_changes.py
tk0miya: I just posted https://github.com/sphinx-doc/sphinx/pull/6532. It will raise `CONFIG_EXTENSIONS_CHANGED` status.

larsoner:
Rebased to latest 2.0 branch, changed my PR to be against that branch instead of `master`, and moved my tests to just add to the new ones from #6532. I assumed this is what you wanted based on the request to rebase against 2.0, let me know if that's wrong and I can `cherry-pick` the commit and make the PR against `master` again.

larsoner:
There was a merge conflict so I rebased and force-pushed

larsoner:
Travis failure looks like an unrelated `apt-get` problem

tk0miya:
Sorry for late response. Merging!

PR #6427, opened 2019-06-02
Fix #5502: linkcheck: Consider HTTP 503 response as not an error

### LINKED ISSUE #5502: allow linkcheck to tolerate transient HTTP 503 errors
I have a link in my docs to https://www.python.org/dev/peps/pep-0020/, which just happened to have been giving 503 errors when my daily Travis build ran this morning, causing my build to fail. I know I can configure `linkcheck_ignore` so that broken links matching specified URL patterns are ignored, but that wouldn't help here, since I want to be able to only warn (not fail) on *any* link giving a 503 error, but still have any link fail on e.g. 404 responses.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #5502

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -159,6 +159,9 @@ class CheckExternalLinksBuilder(Builder):
                 if err.response.status_code == 401:
                     # We'll take "Unauthorized" as working.
                     return 'working', ' - unauthorized', 0
+                elif err.response.status_code == 503:
+                    # We'll take "Service Unavailable" as ignored.
+                    return 'ignored', str(err), 0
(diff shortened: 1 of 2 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -159,6 +159,9 @@ def check_uri():
                 if err.response.status_code == 401:
                     # We'll take "Unauthorized" as working.
                     return 'working', ' - unauthorized', 0
+                elif err.response.status_code == 503:
+                    # We'll take "Service Unavailable" as working.
```
tk0miya: This comment was wrong.

PR #6439, opened 2019-06-03
Make generated download links reproducible

mitya57:
Without this change, the hexdigest generated in `DownloadFiles.add_file()` depends on `source_dir`, which causes problems for [reproducible builds] effort. Just using the path fragment relative to `source_dir` should be sufficient.

In the second commit I added a test to make sure download files with the same name but in different directories are still handled correctly.

[reproducible builds]: https://reproducible-builds.org/

### DIFF
--- a/tests/test_build_html.py
+++ b/tests/test_build_html.py
@@ -453,6 +453,8 @@ def test_html_download_role(app, status, warning):
     app.build()
     digest = md5(b'dummy.dat').hexdigest()
     assert (app.outdir / '_downloads' / digest / 'dummy.dat').exists()
+    digest_another = md5(b'another/dummy.dat').hexdigest()
+    assert (app.outdir / '_downloads' / digest_another / 'dummy.dat').exists()
 
     content = (app.outdir / 'index.html').text()
     assert (('<li><p><a class="reference download internal" download="" '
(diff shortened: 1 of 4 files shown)


tests/test_build_html.py
```diff
@@ -451,15 +451,24 @@ def test_html_download(app):
 @pytest.mark.sphinx('html', testroot='roles-download')
 def test_html_download_role(app, status, warning):
     app.build()
-    digest = md5((app.srcdir / 'dummy.dat').encode()).hexdigest()
+    digest = md5(b'dummy.dat').hexdigest()
     assert (app.outdir / '_downloads' / digest / 'dummy.dat').exists()
+    assert not (app.outdir / '_downloads' / digest / 'dummy.dat').text()
```
tk0miya: What this line confirms? I feel this is meaningless.

tests/test_build_html.py
```diff
@@ -451,15 +451,24 @@ def test_html_download(app):
...
+    digest_another = md5(b'another/dummy.dat').hexdigest()
+    assert (app.outdir / '_downloads' / digest_another / 'dummy.dat').exists()
+    assert (app.outdir / '_downloads' / digest_another / 'dummy.dat').text()
```
tk0miya: This is also.

tk0miya (changes requested):
Reasonable. LGTM with nits.
Please rebase this to 2.0 branch. And it would be nice if you add an entry to CHANGES file.

Thanks,

tests/test_build_html.py
mitya57: One of the `dummy.dat` files is empty, another one (which I added) has some text.

These lines make sure the proper files are picked — empty in the first case, non-empty in the second one.

mitya57: I have added an explanatory comment now.

tk0miya: Yes, I understand that. But why do we need to check the contents of `dummy.dat`?
Is it important to check the file is empty?

If you'd like to confirm copying goes well, this is not good testing. I think 1) filling the some content to the file, and 2) comparing the source and destination files is better.

mitya57: I wanted to make sure that in the right file (of two) is copied, not that the copying goes well. But OK, I will remove these checks now.

tk0miya (approved):
LGTM. Thank you for update.

Issue #6440, opened 2019-06-03
sphinx-apidoc output missing blank lines between modules


bmerry:
sphinx-apidoc produces invalid RST: it fails to output a blank line after each `automodule` block and before the next section header. This causes warnings when building the docs.

This seems to be a regression since 1.8.1. I haven't bisected it.

**To Reproduce**
Steps to reproduce the behavior: in an empty directory,
```sh
$ mkdir dummy doc
$ touch dummy/__init__.py dummy/mod1.py dummy/mod2.py
$ sphinx-apidoc -f -o doc dummy
```

Now `doc/dummy.rst` contains the following:
```
dummy package
=============

Submodules
----------

dummy.mod1 module
-----------------

.. automodule:: dummy.mod1
   :members:
   :undoc-members:
   :show-inheritance:
dummy.mod2 module
-----------------

.. automodule:: dummy.mod2
   :members:
   :undoc-members:
   :show-inheritance:


Module contents
---------------

.. automodule:: dummy
   :members:
   :undoc-members:
   :show-inheritance:
```

**Expected behavior**
There should be a blank line immediately above the header for `dummy.mod2.`

**Environment info**
- OS: Ubuntu 18.04
- Python version: 3.6.6
- Sphinx version: 2.1.0
- Sphinx extensions: sphinx.ext.autodoc

dangunter:
I have also seen this issue. Since we check that the docs build without warnings, it is breaking our builds and we had to pin Sphinx to an older version.

stephenfin:
I've tested this and it's not an issue with 2.0.0 but is introduced with 2.1.0.

PR #6445, opened 2019-06-04
apidoc: Insert newlines between submodules

stephenfin:
Fix a regression introduced in 2.1.0 where submodules didn't have a
newline between them resulting in the following warning:

    Explicit markup ends without a blank line; unexpected unindent.

Signed-off-by: Stephen Finucane <stephen@that.guru>
Closes: #6440

### DIFF
--- a/sphinx/templates/apidoc/package.rst
+++ b/sphinx/templates/apidoc/package.rst
@@ -40,8 +40,8 @@ Submodules
 {{- [submodule, "module"] | join(" ") | e | heading(2) }}
 {% endif %}
 {{ automodule(submodule, automodule_options) }}
-{%- endfor %}
-{% endif %}
+{% endfor %}
+{%- endif %}
 {% endif %}
 
 {%- if not modulefirst and not is_namespace %}
--- a/tests/test_ext_apidoc.py
+++ b/tests/test_ext_apidoc.py
@@ -467,7 +467,8 @@ def test_package_file(tempdir):
     outdir = path(tempdir)
     (outdir / 'testpkg').makedirs()
     (outdir / 'testpkg' / '__init__.py').write_text('')
-    (outdir / 'testpkg' / 'example.py').write_text('')
+    (outdir / 'testpkg' / 'hello.py').write_text('')
+    (outdir / 'testpkg' / 'world.py').write_text('')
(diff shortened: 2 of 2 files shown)


tk0miya (changes requested):
It seems failed. Could you check the code please?
In addition, I'll release 2.1.1 package for some reasons. So it would be nice if this goes into 2.1.1 branch.

stephenfin:
> It seems failed. Could you check the code please?
> In addition, I'll release 2.1.1 package for some reasons. So it would be nice if this goes into 2.1.1 branch.

My mistake. Jinja2 gets complicated real fast. Should be fixed now.

tk0miya:
It seems template is fixed. But this still goes forward to master branch. Could you rebase this please?
In addition, please add an entry of CHANGES.
Thanks,

stephenfin:
Which branch? I don't see a 2.1 branch?

tk0miya:
Please merge into 2.1.1 branch.

stephenfin:
> Please merge into 2.1.1 branch.

Okay, done. Unrelated but can we rename this to the `2.1` branch (i.e. the branch that all 2.1.x releases will be made from)? The current name is very confusing. I thought `2.1.1` was a tag. We already have a `2.0` branch so `2.1` makes sense. Note that it also means we can't browse documentation at https://www.sphinx-doc.org/en/2.1/ even though we can browse https://www.sphinx-doc.org/en/2.0/

tk0miya (approved):
LGTM. Go ahead!

tk0miya:
>Note that it also means we can't browse documentation at https://www.sphinx-doc.org/en/2.1/ even though we can browse https://www.sphinx-doc.org/en/2.0/

IMO, document for each minor release is not needed. I agree with `.../2.0` path is confusable. Personally, it is enough to provide the only one document for each major release. So renaming it to `.../2.x` is better.

>Unrelated but can we rename this to the 2.1 branch (i.e. the branch that all 2.1.x releases will be made from)?
..

I agree that `2.0` is confusable for 2.x series. So +0 for renaming. But I think `2.x` is better. Anyway, we need to discuss more. So please file an another issue to propose the change.
ping @shimizukawa; Can you tell us your idea for us in the coming issue.

Issue #967, opened 2015-01-03
SVG inheritance diagram is not hyperlinked (clickable)


shimizukawa:
PNG inheritance diagrams can be clicked on, while the SVG variant is not clickable. 

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/967
- Originally reported by: [eudoxos](https://bitbucket.org/eudoxos)
- Originally created at: 2012-07-14T08:48:02.161

tk0miya:
Fixed at e859543.

PR #6454, opened 2019-06-06
add svg_link_prefix option to inheritance_diagram

eteq:
Subject: Add an option to set the prefix for links in SVG inheritance diagrams

### Feature or Bugfix
- Feature

### Purpose
- Provide a way for extensions or projects that embed the inheritance diagram SVG in pages to have the links work.

### Detail
As discussed in astropy/astropy#4935 and #2484, the inheritance diagram machinery currently yields broken links as we implemented it in sphinx-automodapi and astropy.  You can see an example here: https://docs.astropy.org/en/v3.1.2/coordinates/index.html#class-inheritance-diagram .  The gist of the problem (as discussed in #2484) is that we embed the SVG in the page. This means the path to the actual class pages don't have the `'..'` that apparently gets created when embedding as a separate file. (FWIW, the links *originally* worked when we first did this, but then the changes in #967 ended up breaking it.  And here we are years later with me finally having realized the underlying cause!).

So this PR introduces a new config option in inheritance_diagram that sets the prefix.  The default is still `../`, so it shouldn't break other code, but it will then be possible to supress the `..` in cases like ours where it's breaking the links.  I *think* it may partly address #3176 by giving that use case the flexibility to pick their own prefix as well (although I'm not 100% sure about that one).

### Relates
- #2484
- #3176
- astropy/astropy#4935

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -35,6 +35,7 @@ Deprecated
 
 Features added
 --------------
+* #6454: An ``svg_link_prefix`` option was added to the inheritance diagram extension
 
 Bugs fixed
 ----------
--- a/sphinx/ext/inheritance_diagram.py
+++ b/sphinx/ext/inheritance_diagram.py
@@ -403,25 +403,26 @@ def html_visit_inheritance_diagram(self, node):
     Output the graph for HTML.  This will insert a PNG with clickable
     image map.
     """
+    env = self.builder.env
     graph = node['graph']
 
     graph_hash = get_graph_hash(node)
     name = 'inheritance%s' % graph_hash
 
     # Create a mapping from fully-qualified class names to URLs.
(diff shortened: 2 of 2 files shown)


eteq:
Also, I wasn't entirely sure if this was supposed to be against the 2.1.1 or the master branch - I'm not clear on whether adding new configuration items is considered backwards-compatibility breaking or not, so using `master` seemed like the best choice.  I can re-submit it against the 2.1.1 branch if I mis-interpreted that, though.

tk0miya:
What case are users need to change this? If my understanding correct, this is only used for automodapi extension. If so, I propose you to modify `node['refuri']` for SVG images on automodapi side.
Thanks,

tk0miya:
I'm closing this because main usecase of this PR is not clear.
Please feel free to reopen if you'd like to merge this again.
Thanks,

Issue #6462, opened 2019-06-08
C++ domain: Trips over complex constant with unknown escape sequence


t-b:
Using latest breathe and sphinx and applying the following patch in breathe

```diff
$git diff .
diff --git a/examples/doxygen/example.cpp b/examples/doxygen/example.cpp
index 15aee3f..c9ac551 100644
--- a/examples/doxygen/example.cpp
+++ b/examples/doxygen/example.cpp
@@ -17,3 +17,7 @@ void Test5::example() {}
  * This is an example of how to use the Test5 class.
  * More details about this example.
  */
+
+#include <string>
+
+std::string RES_FORMAT_STR = "\\[1\\K(%d, %d, %d)\\{\"%%s\", FloatWithMinSigDigits(%s[%d], numMinSignDigits = 2)}\\]1\K(0, 0, 0)"; ///< some comment
```

I get

```
/home/firma/devel/breathe/documentation/source/doxygen.rst:64: WARNING: Invalid definition: Expected end of definition. [error at 50]
  std::string RES_FORMAT_STR = "\[1\K(%d, %d, %d)\{"%%s", FloatWithMinSigDigits(%s[%d], numMinSignDigits = 2)}\]1K(0, 0, 0)"
```

but the code is valid C++

```
$git diff .
diff --git a/examples/doxygen/example.cpp b/examples/doxygen/example.cpp
index 15aee3f..c9ac551 100644
--- a/examples/doxygen/example.cpp
+++ b/examples/doxygen/example.cpp
@@ -17,3 +17,7 @@ void Test5::example() {}
  * This is an example of how to use the Test5 class.
  * More details about this example.
  */
+
+#include <string>
+
+std::string RES_FORMAT_STR = "\\[1\\K(%d, %d, %d)\\{\"%%s\", FloatWithMinSigDigits(%s[%d], numMinSignDigits = 2)}\\]1\K(0, 0, 0)"; ///< some comment
```

Looking at the result

![grafik](https://user-images.githubusercontent.com/1185677/59150702-f88f2880-8a27-11e9-910e-7f63a0aefe23.png)

it seems that sphinx is removing some backslashes and then get's tripped.

The output from sphinx 1.8.3 can be viewed [here](https://alleninstitute.github.io/MIES/file/_m_i_e_s___oscilloscope_8ipf.html?highlight=res_format#_CPPv314RES_FORMAT_STR).

**Environment info**
- OS: Linux
- Python version: 3.6.7
[...]

PR #6473, opened 2019-06-11
Keep all backslashes in domain directives.

jakobandersen:
Subject: modify the central line handler for domain directives to not touch double backslashes.

### Feature or Bugfix
- Bugfix

### Purpose
It looks like a long time ago an extra regex replacement was inserted in function that handles escaped line breaks in domain directives. It replaces two consecutive backslashes with a single backlash. This means that users/tools must escape all backslashes in declarations once more. Breathe doesn't do this, does autodoc?
This PR changes the default behaviour so not backslash stripping is being done by default.
If a domain is specifically designed to rely on the stripping, it can set ``strip_backslashes = True`` in the subclasses of ``ObjectDescription``.
Users can selectively reenable the stripping in specific domains with a new configuration value ``signature_backslash_strip_domain_override`` (a list of domain names). The default is ``None``, while the empty list ``[]`` reverts *all* domains to the old behaviour.

### Relates
- #6462 (see also this for an example for reproduction)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -42,6 +42,12 @@ Incompatible changes
   The existing ones are now more strict, resulting in new warnings.
 * The attribute ``sphinx_cpp_tagname`` in the ``desc_signature_line`` node
   has been renamed to ``sphinx_line_type``.
+* #6462: double backslashes in domain directives are no longer replaced by
+  single backslashes as default. Each class derived from ObjectDescription
+  may reenable the stripping for it self. A new configuration value
+  :confval:`signature_backslash_strip_domain_override` can be used by users
+  to reenable it per domain as well. Setting it to ``[]`` will reinstate the
+  old behaviour with backslash stripping in every domain.
 
 Deprecated
 ----------
--- a/CHANGES
+++ b/CHANGES
@@ -43,11 +43,8 @@ Incompatible changes
 * The attribute ``sphinx_cpp_tagname`` in the ``desc_signature_line`` node
   has been renamed to ``sphinx_line_type``.
 * #6462: double backslashes in domain directives are no longer replaced by
-  single backslashes as default. Each class derived from ObjectDescription
-  may reenable the stripping for it self. A new configuration value
-  :confval:`signature_backslash_strip_domain_override` can be used by users
-  to reenable it per domain as well. Setting it to ``[]`` will reinstate the
-  old behaviour with backslash stripping in every domain.
+  single backslashes as default. A new configuration value
+  :confval:`strip_signature_backslash` can be used by users to reenable it.
 
 Deprecated
 ----------
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -634,14 +634,13 @@ General configuration
    .. versionchanged:: 1.1
       Now also removes ``<BLANKLINE>``.
 
-.. confval:: signature_backslash_strip_domain_override
+.. confval:: strip_signature_backslash
 
-    A list of domain names for which to forcibly reinstate backslash stripping.
-    The value ``None`` means "no domains" while ``[]`` means every domain
-    (i.e., the behaviour before version 3.0).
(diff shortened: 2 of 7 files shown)


tk0miya:
This was added at 10 year ago (bd32a22946). I don't know how useful this stripping. If this was introduced only for helping editors' highlighting as commented, I think this is worse approach because this brings inconsistency to our notation. But this code has stripped backslashes in signatures during 10 years. That makes many signatures escaped. As a result, this change is a breaking change (or destructive change) even if this is right approach.

Therefore, we need to switch this behavior carefully. So it would be better to merge this into major release. In addition, it would be better to provide migration path like some warnings, configurations and so on.

@shimizukawa do you have any opinion?

t-b:
@tk0miya @jakobandersen 
(I'm the reporter of the bug)

Sphinx does not get tripped in 1.8.3 so this looks like a regression to me. Or did the C++ domain get better and this error is only now exposed?

jakobandersen:
@t-b, I'm quite sure that it is not due to changes in backslash handling. Rather I think it may be because the C++ domain has started doing more detailed parsing of string literals.

jakobandersen:
@tk0miya, good points. I have change the PR to include various ways to get the old behaviour (see the updated PR description and the code changes). How does it look to you?

t-b:
@jakobandersen Works as advertised. Thanks for the PR!

t-b:
@jakobandersen @tk0miya Can this be part of the 3.0 release?

jakobandersen:
Ping @tk0miya, if you find time before 3.0 freeze.

tk0miya:
Oh, sorry for not responding longer.

TBH, I hesitate to add such a complex implementation to the core. Because I suppose nobody (or only very few people) uses this behavior (because nobody knows it!). So I'd like to add a small migration code during 1 or 2 major versions. And dropping codes by historical reason finally. For such purpose, this is too much to support old behavior.

If my understanding correct, this feature only affects users. And no object-descriptions depends on it because it is a preprocessor before they work. So I suppose no options like `strip_backslashes` are not needed. How about add simply `signature_backslash_strip = True | False` to keep compatibility? It only enables/disables the stripping backslashes (defaults to False). It's much simpler than now.

jakobandersen:
@tk0miya, if not even autodoc relies on this I completely agree. I'll implement your suggestion soon and update the PR.

tk0miya:
Basically autodoc does not depend on the behavior. But I found an exceptional case:

Input:
```
def foo(name='\\'):
    pass
```
Output:
```
.. py:function:: foo(name='\\\\')
   :module: example
```

It uses the stripping to suppress escaped backslash. I'll take a look from now on.

tk0miya:
I posted #7334 as an example for this. But it seems better to drop stripping backslashes also for autodoc. What do you think?

tk0miya:
Sorry, please forget my comment above.

autodoc expects the stripping in ObjectDescription and do escaping in 2 methods. So this should be disabled by `signature_backslash_strip` setting.
https://github.com/sphinx-doc/sphinx/blob/8cac7fdd580cf574c3963310c5bf7ce46d06ac98/sphinx/ext/autodoc/__init__.py#L1046-L1048
https://github.com/sphinx-doc/sphinx/blob/8cac7fdd580cf574c3963310c5bf7ce46d06ac98/sphinx/ext/autodoc/__init__.py#L1440-L1442

jakobandersen:
The code has now been simplified and the two cases in autodoc @tk0miya found are now conditioned on the stripping being active.

There is a test case in ``test_autodoc.py`` that now fails. @tk0miya, do you know how easily to split this particular one into two cases: one with stripping and one without? Otherwise I can simplify it to only test the new behaviour.

tk0miya:
+1 for testing with new behavior only.

tk0miya:
LGTM, merging now!

Issue #6470, opened 2019-06-10
sphinx.cmd.build Syntax Error


kdschlosser:
**Traceback**
File "sphinx-2.1.0-py2.7.egg\sphinx\cmd\build.py", line 39
    file=stderr)
        ^
SyntaxError: invalid syntax

**To Reproduce**
```python
from sphinx.cmd.build import build_main
```

**Expected behavior**
No Syntax Error

**Your project**
python-openzwave
what I am working on has not yet been pushed to the repository

**Environment info**
- OS: Windows 7 x64 SP1
- Python version: 2.7.15
- Sphinx version: 2.1.0
- Sphinx extensions:  didn't get that far into the build process
- Extra tools: N/A

**Additional context**
What I am currently working on is making an extension for distutils "build_docs". I made it to use the exact same command line arguments as sphinx. This would make it easier to port if someone wanted to do the same as what I am doing. It also builds the conf.py file. I did have it working with an older version of Sphinx. I do not recall the exact version. The information I can provide is that build_main was directly accessible from the main sphinx module. example below shows how the older version worked. 

``` python

import sphinx

sphinx.build_main()
```
I do not know if __future__.print_function has support in it for setting the file. This may be the only thing that needs to be added.

kdschlosser:
OK so I have just looked at the Pypi page for Sphinx. It states on the page python >= 3.5

But yet Pypi returns the URL for a wheel for 2.1.0 on python 2.x. easy_install is what setuptools uses to install requirements. It has a different approach to locating a compatible version of a module then pip does (pretty obvious at this point). pip errors when trying to install 2.1.0. easy_install does not.

Here is the URL for the wheel

https://files.pythonhosted.org/packages/e1/81/1dd79c5cd4320f2e904d7fbed7990d07c3b4e7a21efd2071be88dd5806f1/Sphinx-2.1.0-py2.py3-none-any.whl#sha256=2c5becc0fd6706dc0aeb4703f9f1f8a1d1eecacf02e9ac5943cbae48b11e5e42

1.8.5 is the last version made that is python 2.x compatible. My initial issue is solved. 

There is still a bug so I am not going to close this issue.

aodag:
i thinks kdschlosser installed sphinx with easy_install, because that installed in egg directory.
easy_install can't install wheel format distribution, and easy_install may not support python_requires.

I tryed install sphinx on python 2.7 using pip19.1.1 and i got sphinx 1.8.5.
[...]

PR #6480, opened 2019-06-13
Remove "universal" configuration from setup.cfg

jdufresne:
Sphinx is Python-3-only and therefore is not a universal wheel. It
should still distribute a 'py3' wheel, just not a universal one.

From https://wheel.readthedocs.io/en/stable/user_guide.html

> If your project contains no C extensions and is expected to work on
> both Python 2 and 3, you will want to tell wheel to produce universal
> wheels by adding this to your setup.cfg file: …

As Sphinx doesn't work on Python 2, it should not be universal.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,9 +5,6 @@ license_file = LICENSE
 tag_build = .dev
 tag_date = true
 
-[bdist_wheel]
-universal = 1
-
 [aliases]
 release = egg_info -Db ''
 upload = upload --sign --identity=36580288


tk0miya (changes requested):
Reasonable.
Could you rebase this to 2.0 branch?

tk0miya:
Note: this is related with #6470.

jdufresne:
>  Could you rebase this to 2.0 branch?

No problem. Done. :slightly_smiling_face:

tk0miya (approved):

tk0miya:
Thanks!

PR #6481, opened 2019-06-13
Remove empty code block in sphinx.util.requests

jdufresne:
Empty since 19a187cb5443808c284dd9e8f85ad13cba98182b.

If requests is installed with the security extras, they will be loaded
automatically by requests. The is no need to proactively load it. The
block simply did a "pass" in all cases.


tk0miya:
Thank you always!

Issue #6478, opened 2019-06-12
ObjectDescription doc field type map incorrectly set


ViktorHaag:
**Describe the bug**
Refactoring the setting of an ObjectDescription class's type map out of the DocFieldTransformer and into the ObjectDescription class itself wasn't done correctly, I think. Typed fields built into an object description that's part of a domain aren't always parsed and rendered correctly.

**To Reproduce**
Start with a fresh Sphinx 2.1.1 install, and install `sphinxcontrib.httpdomain`. Then, create a brand new sphinx project, and make two changes:

Change the project's `conf.py` so that:
``` python
extensions = [ 'sphinxcontrib.httpdomain' ]
```

Change the project's `index.rst` so that you first add a simple HTTP Domain documentation block to the bottom:
```
.. http:get:: /users/(id)

   Retrieve a user record by ID.

   :param string id: User identifier.
   :query email: Include email address.
   :qtype email: boolean
```

When you now `make html` the project, you will see that the output for the docs is as you expect; you see the syntax of the typed `query` and `qtype` field pair produce a field entry that reads:
```
Query Parameters:  * email (boolean) -- Include email address.
```
Now add a documentation block from a different domain just above the http:get block you already have:
```
.. py:function:: foo(id) -> string

   Create a foo.

   :param string id: Identifier.


.. http:get:: /users/(id)

   Retrieve a user record by ID.

   :param string id: User identifier.
   :query email: Include email address.
   :qtype email: boolean
```
When you `make clean html` now you will notice that Sphinx has stopped properly processing the field syntax for the fields under the http:get entry; the output looks like this now:
```
Query email: Include email address.
Qtype email: boolean
```

Note that if you add the python documentation block _after_ the HTTP get block, and you use a typed field that isn't already defined in the HTTP domain's list of field types, then it's the Python documentation that gets effected:
```
.. http:get:: /users/(id)

   Retrieve a user record by ID.

   :param string id: User identifier.
   :query email: Include email address.
   :qtype email: boolean

.. py:function:: foo(id) -> string

   Create a foo.

   :param string id: Identifier.
   :variable bar: Bar variable.
   :vartype bar: string
```

With this arrangement, the output is like this:
```
GET /users/(id)
  Retrieve a user record by ID.

  Parameters:  * id (string) – User identifier.
  Query Parameters:  * email (boolean) – Include email address.


foo(id) → string
  Create a foo.

  Parameters:  * id (string) – Identifier.
[...]

PR #6482, opened 2019-06-13
Explicitly treat ObjectDescription._doc_field_type_map as an instance variable

ViktorHaag:
Subject: Aims to address [issue #6478](https://github.com/sphinx-doc/sphinx/issues/6478)

### Feature or Bugfix

- Bugfix

### Purpose
Explicitly treat `ObjectDescription._doc_field_type_map` population as an instance variable so that we can be sure that Domains with objects that define their own field types will render properly on the output.

### Detail

- Left `ObjectDescription._doc_field_type_map` declaration alone and in place in
  case there are other factors at play that require it declared on the class.

- Ensured that `ObjectDescription.get_field_type_map()` made sure to populate the instance scoped version of `_doc_field_type_map` and not the class scoped version of the variable. 

### Relates
- #6478

### DIFF
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -71,6 +71,7 @@ class ObjectDescription(SphinxDirective):
     _doc_field_type_map = {}  # type: Dict[str, Tuple[Field, bool]]
 
     def _process_type_map(self, typelist):
+        # type: () -> Dict[str, Tuple[Field, bool]]
         typemap = {}
         for field in typelist:
             for name in field.names:
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -70,25 +70,17 @@ class ObjectDescription(SphinxDirective):
     # Warning: this might be removed in future version. Don't touch this from extensions.
     _doc_field_type_map = {}  # type: Dict[str, Tuple[Field, bool]]
 
-    def _process_type_map(self, typelist):
-        # type: (List[Field]) -> Dict[str, Tuple[Field, bool]]
-        typemap = {}
-        for field in typelist:
-            for name in field.names:
-                typemap[name] = (field, False)
-            if field.is_typed:
-                typed_field = cast(TypedField, field)
-                for name in typed_field.typenames:
-                    typemap[name] = (field, True)
-        return typemap
-
-    def __init__(self, *args, **kwargs):
-        # type: (*Any, **Any) -> None
-        SphinxDirective.__init__(self, *args, **kwargs)
-        self._doc_field_type_map = self._process_type_map(self.doc_field_types)
-
     def get_field_type_map(self):
         # type: () -> Dict[str, Tuple[Field, bool]]
+        if self._doc_field_type_map == {}:
+            self._doc_field_type_map = {}
+            for field in self.doc_field_types:
+                for name in field.names:
+                    self._doc_field_type_map[name] = (field, False)
+                if field.is_typed:
+                    typed_field = cast(TypedField, field)
+                    for name in typed_field.typenames:
+                        self._doc_field_type_map[name] = (field, True)
         return self._doc_field_type_map
 
     def get_signatures(self):
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -70,18 +70,23 @@ class ObjectDescription(SphinxDirective):
     # Warning: this might be removed in future version. Don't touch this from extensions.
     _doc_field_type_map = {}  # type: Dict[str, Tuple[Field, bool]]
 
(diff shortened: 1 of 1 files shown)


sphinx/directives/__init__.py
```diff
@@ -70,18 +70,25 @@ class ObjectDescription(SphinxDirective):
...
+
+    def __init__(self, *args, **kwargs):
+        # type: (*Any, **Any) -> None
+        SphinxDirective.__init__(self, *args, **kwargs)
+        self._doc_field_type_map = self._process_type_map(self.doc_field_types)
+
     def get_field_type_map(self):
         # type: () -> Dict[str, Tuple[Field, bool]]
-        if self._doc_field_type_map == {}:
-            for field in self.doc_field_types:
```
tk0miya: Could you override `self._doc_field_type_map` here please? For some reasons, I'd not like to add a new method.

tk0miya (changes requested):
LGTM with nits.

sphinx/directives/__init__.py
ViktorHaag: Hey @tk0miya -- I'm unsure exactly what you'd like here, but are you thinking something like this?
``` python
_doc_field_type_map = {}

def __init__ (self, *args, **kwargs):
  SphinxDirective.__init__(self, *args, **kwargs)
  self._doc_field_type_map = self.get_field_type_map(self):

def get_field_type_map(self):
  if self._doc_field_type_map == {}:
    // if unset, set self._doc_field_type_map, as the old code used to
  return self._doc_field_type_map
```

ViktorHaag: Hey @tk0miya -- I'm sorry, I'm unsure what you mean here... do you mean to move the code I'd put into `_process_type_map` into the constructor? Or do you mean that you'd prefer for `get_field_type_map` to always rebuild the type map?

ViktorHaag: Hey @tk0miya -- I'm sorry, I'm unsure what you mean here... do you mean to move the code I'd put into `_process_type_map` into the constructor? Or do you mean that you'd prefer for `get_field_type_map` to always rebuild the type map?

ViktorHaag: For example, this suggested change seems lighter and more in line with the existing code, but adds an optional keyword argument to `get_field_type_map`:
``` python
    # Warning: this might be removed in future version. Don't touch this from extensions.
    _doc_field_type_map = {}  # type: Dict[str, Tuple[Field, bool]]

    def __init__(self, *args, **kwargs):
        # type: (*Any, **Any) -> None
        SphinxDirective.__init__(self, *args, **kwargs)
        self._doc_field_type_map = self.get_field_type_map(force_rebuild=True)

    def get_field_type_map(self, force_rebuild=False):
        # type: (bool) -> Dict[str, Tuple[Field, bool]]
        if force_rebuild or self._doc_field_type_map == {}:
            for field in self.doc_field_types:
                for name in field.names:
                    self._doc_field_type_map[name] = (field, False)
                if field.is_typed:
                    typed_field = cast(TypedField, field)
                    for name in typed_field.typenames:
                        self._doc_field_type_map[name] = (field, True)
        return self._doc_field_type_map

```

tk0miya: sorry for poor comment. I meant this code.
```
    def get_field_type_map(self) -> Dict[str, Tuple[Field, bool]]:
        if self._doc_field_type_map == {}:
            self._doc_field_type_map = {}    # OVERRIDE the class variable by instance's
            for field in self.doc_field_types:
                for name in field.names:
                    self._doc_field_type_map[name] = (field, False)

                if field.is_typed:
                    typed_field = cast(TypedField, field)
                    for name in typed_field.typenames:
                        self._doc_field_type_map[name] = (field, True)

        return self._doc_field_type_map
```

ViktorHaag: Oh! I see -- yes, that is much simpler. I have pushed a new commit to change as you suggest.

sphinx/directives/__init__.py
```diff
@@ -73,15 +73,14 @@ class ObjectDescription(SphinxDirective):
     def get_field_type_map(self):
         # type: () -> Dict[str, Tuple[Field, bool]]
         if self._doc_field_type_map == {}:
+            self._doc_field_type_map = {}
             for field in self.doc_field_types:
                 for name in field.names:
                     self._doc_field_type_map[name] = (field, False)
-
```
tk0miya: This PR removes unrelated empty lines. Could you fix this? Then I'll merge this soon.

ViktorHaag: OK

tk0miya (approved):
Thank you for quick work!

tk0miya:

PR #6491, opened 2019-06-15
Rewrite documents for code-block directives by directive:option

tk0miya:
### Feature or Bugfix
- Refactoring
- Documents

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -491,72 +491,98 @@ __ http://pygments.org/docs/lexers/
    :rst:dir:`highlight` directive will be used.  If not set,
    :confval:`highlight_language` will be used.
 
-   **Additional options**
-
-   Pygments can generate line numbers for code blocks. To enable this for, use
-   the ``linenos`` flag option. ::
-
-      .. code-block:: ruby
-         :linenos:
-
-         Some more Ruby code.
-
-   The first line number can be selected with the ``lineno-start`` option.  If
-   present, ``linenos`` flag is automatically activated::
-
-      .. code-block:: ruby
-         :lineno-start: 10
-
-         Some more Ruby code, with line numbering starting at 10.
-
-   Additionally, an ``emphasize-lines`` option can be given to have Pygments
-   emphasize particular lines::
-
-    .. code-block:: python
-       :emphasize-lines: 3,5
-
-       def some_function():
-           interesting = False
-           print 'This line is highlighted.'
-           print 'This one is not...'
-           print '...but this one is.'
-
-   A ``caption`` option can be given to show that name before the code block.
-   A ``name`` option can be provided implicit target name that can be
-   referenced by using :rst:role:`ref`.  For example::
-
-     .. code-block:: python
-        :caption: this.py
-        :name: this-py
-
-        print 'Explicit is better than implicit.'
-
-   A ``dedent`` option can be given to strip indentation characters from the
-   code block. For example::
-
-      .. code-block:: ruby
-         :dedent: 4
-
-             some ruby code
-
-   A ``force`` option can ignore minor errors on highlighting.
-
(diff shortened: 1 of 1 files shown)


doc/usage/restructuredtext/directives.rst
```diff
@@ -483,72 +491,91 @@ __ http://pygments.org/docs/lexers/
...
+              print 'This line is highlighted.'
+              print 'This one is not...'
+              print '...but this one is.'
 
-             some ruby code
+      .. versionadded:: 1.1
+      .. versionchanged:: 1.6.6
+         LaTeX supports the ``emphasize-lines`` option.
 
-   A ``force`` option can ignore minor errors on highlighting.
```
tk0miya, line 534: `:force:` option was vanished on merging...

doc/usage/restructuredtext/directives.rst
```diff
@@ -563,42 +590,74 @@ __ http://pygments.org/docs/lexers/
...
+         .. literalinclude:: example.py
+            :pyobject: Timer.start
 
-   This would only include the code lines belonging to the ``start()`` method
-   in the ``Timer`` class within the file.
+      This would only include the code lines belonging to the ``start()`` method
+      in the ``Timer`` class within the file.
 
    Alternately, you can specify exactly which lines to include by giving a
    ``lines`` option::
```
tk0miya: I'll work another options later.

tk0miya: I'll try in another PR.

doc/usage/restructuredtext/directives.rst
```diff
@@ -483,72 +491,98 @@ __ http://pygments.org/docs/lexers/
...
+      .. versionadded:: 1.3
 
-     .. code-block:: python
-        :caption: this.py
-        :name: this-py
+   .. rst:directive:option:: emphasized-lines: line numbers
+      :type: comma separated numbers
 
-        print 'Explicit is better than implicit.'
+      Empahsize particular lines of the code block::
```
jfbu, line 525: maybe we can use a code-block with reST lexer and emphasize lines number 6 and 8 ...
~~~rest
   Additionally, an ``emphasize-lines`` option can be given to have Pygments
   emphasize particular lines (we are showing output effect on top of input...):

   .. code-block:: rest
      :emphasize-lines: 6,8

      .. code-block:: python
         :emphasize-lines: 3,5

         def some_function():
             interesting = False
             print 'This line is highlighted.'
             print 'This one is not...'
             print '...but this one is.'
~~~

jfbu (commented):
LGTM +1! I did not know about `:linenothreshold:`, I should read our docs more...

and I did not know about `:force:` either...

jfbu:
ah this was already merged... +1 anyway!

doc/usage/restructuredtext/directives.rst
tk0miya: With this change, it will be shown as:
スクリーンショット 2019-12-10 1 25 51

I feel emphasize input source code is strange a bit. But it might help users to imagine its behavior.

tk0miya:
Yes, we have a lot of options for directives. But it is difficult to understand because they are described as narrative document. I think new `rst:directive:option` directive helps to read its definition.
With this change, each option is rendered as a definition-block:
Before: https://www.sphinx-doc.org/en/1.8/usage/restructuredtext/directives.html#directive-code-block
After: https://www.sphinx-doc.org/en/2.0/usage/restructuredtext/directives.html#directive-code-block

I'll keep working remaining options in another PR :-)

doc/usage/restructuredtext/directives.rst
jfbu: I agree it looks a bit strange (with the highlighting going to the left of `def`). -0 to add this change...

jfbu:
> I'll keep working remaining options in another PR :-)

+1 for that if you have time :-)

Issue #6527, opened 2019-06-28
last_updated variable from sphinx build wrongly assumes UTC timezone


ViktorHaag:
**Describe the bug**
Because of an interaction between the way that the Sphinx html builder generates a last_updated datetime, and the assumptions made by the underlying `babel` library on formatting the time, the "last updated time" used in Sphinx templates is assumed to be a UTC time value, when that's not the case.

**To Reproduce**
* Use a generic Sphinx document with a template that includes "last_updated" variable in the output.
* In conf.py set `html_last_updated_fmt = "%Y-%m-%d (%H:%M %Z)" `; this will prompt the last updated date-time format used to print out the timezone info as part of the formatted date-time string.
* Build the docs: the timezone in the output will get reported as UTC, but the actual time will use whatever date-time/timezone your local system clock is set to.

**Expected behavior**
Either the build code should properly determine what the local platform's timezone is in order to capture that information in the `datetime` passed on to the babel formatter, or the builder code should always create a `datetime` that captures the system clock time in UTC so that babel correctly reports the timezone when it claims the timezone is UTC.

**Environment info**
- OS: MacOS 10.14.5
- Python version: 3.7.3
- Sphinx version: 2.1.2
- Sphinx extensions:  HTML builder used
- Extra tools: n/a

**Additional context**
* Starting about line 482 in `/sphinx/builders/html` is this call to the i18n.format_date function (note that this only passes in a format string, and a locale/language, not a date-time value itself):
```python
...
        # format the "last updated on" string, only once is enough since it
        # typically doesn't include the time of day
        lufmt = self.config.html_last_updated_fmt
        if lufmt is not None:
            self.last_updated = format_date(lufmt or _('%b %d, %Y'),
                                            language=self.config.language)
        else:
            self.last_updated = None
...
```

* Starting about line 280 in `/sphinx/util/i18n.py` is the definition for the `format_date` function:
```python
...
def format_date(format, date=None, language=None):
    # type: (str, datetime, str) -> str
    if date is None:
        # If time is not specified, try to use $SOURCE_DATE_EPOCH variable
        # See https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal
        source_date_epoch = os.getenv('SOURCE_DATE_EPOCH')
        if source_date_epoch is not None:
            date = datetime.utcfromtimestamp(float(source_date_epoch))
        else:
            date = datetime.now()
...
```

[...]

PR #6528, opened 2019-06-28
Force last update datetime value to always be UTC

ViktorHaag:
Subject: We should use UTC datetime values for build times, always

### Feature or Bugfix
- Bugfix

### Purpose
This fix forces `last_updated` build-time values to always be expressed as UTC times, so that if someone builds docs that use a full date-time stamp, with timezone, in their output template, the formatted time gets properly represented as (and labelled as) a UTC time value.

### Details
- If SOURCE_DATE_EPOCH is set in the environment, then the build date will use
  UTC as the timezone.

- If it's not set in the environment, we should do the same thing and use UTC
  as the timezone for the build date.

- `datetime.now()` produces a naive system-time datetime object; `datetime.utcnow()`
  produces the aware, UTC-oriented datetime object.

### Relates
Please see https://github.com/sphinx-doc/sphinx/issues/6527

### DIFF
--- a/sphinx/util/i18n.py
+++ b/sphinx/util/i18n.py
@@ -286,7 +286,7 @@ def format_date(format, date=None, language=None):
         if source_date_epoch is not None:
             date = datetime.utcfromtimestamp(float(source_date_epoch))
         else:
-            date = datetime.now()
+            date = datetime.utcnow()
 
     result = []
     tokens = date_format_re.split(format)


tk0miya:
Thank you for your work!

tk0miya:
I found another way to get a datetime object having correct timezone (local-system-timzone).
https://stackoverflow.com/questions/2720319/python-figure-out-local-timezone/39079819#39079819

```
>>> from datetime import datetime
>>> from datetime import timezone
>>> datetime.now()
datetime.datetime(2019, 6, 29, 10, 30, 11, 992366)
>>> datetime.now(timezone.utc).astimezone()
datetime.datetime(2019, 6, 29, 10, 30, 22, 40327, tzinfo=datetime.timezone(datetime.timedelta(seconds=32400), 'JST'))
```

tk0miya:
I just posted #6529 to adjust the timestamp to localtime (with keeping timezone info). Could you check this please?

ViktorHaag:
> I just posted #6529 to adjust the timestamp to localtime (with keeping timezone info). Could you check this please?

Thank you! I apologize for the delay in responding to this new PR -- I had several days of holidays.

PR #6529, opened 2019-06-29
format_date(): use "aware" local datetime by default (refs: #6528)

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- This gives system timezone information to default value of ``format_date()``.  And it does not modifies the timestamp to UTC.
- refs: #6528

### DIFF
--- a/sphinx/util/i18n.py
+++ b/sphinx/util/i18n.py
@@ -12,7 +12,7 @@ import os
 import re
 import warnings
 from collections import namedtuple
-from datetime import datetime
+from datetime import datetime, timezone
 from os import path
 from typing import Callable, Generator, List, Set, Tuple
 
@@ -274,7 +274,7 @@ def format_date(format: str, date: datetime = None, language: str = None) -> str
         if source_date_epoch is not None:
             date = datetime.utcfromtimestamp(float(source_date_epoch))
         else:
-            date = datetime.utcnow()
+            date = datetime.now(timezone.utc).astimezone()
 
     result = []
     tokens = date_format_re.split(format)


sphinx/util/i18n.py
```diff
@@ -274,7 +274,7 @@ def format_date(format: str, date: datetime = None, language: str = None) -> str
         if source_date_epoch is not None:
             date = datetime.utcfromtimestamp(float(source_date_epoch))
         else:
-            date = datetime.utcnow()
+            date = datetime.now(timezone.utc).astimezone()
```
ViktorHaag, line 277: Hmmm... I'm not sure this does what we expect it to do.

* If we want `date` to be set to a datetime in UTC, then I think we want this: `date = datetime.now().astimezone(timezone.utc)`, or more simply `date = datetime.now(timezone.utc)` or `datetime.utcnow()`.

* If we want `date` to always be a non-naive datetime that uses the system timezone, then I think we want this: `date = datetime.now().astimezone()`.

I think in either case, this will format as expected when passed to the `babel` library because the `date` datetime that babel now gets will not be naive: it will either be explicitly in UTC, or it will explicitly be in whatever the system timezone is. Here's a quick demonstration from the iPython repl:

```
Python 3.7.3 (default, Mar 27 2019, 09:23:15) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.6.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from datetime import datetime, timezone                                 

In [2]: from babel.dates import format_datetime                                 

In [3]: format_datetime(datetime.utcnow())                                      
Out[3]: 'Jul. 2, 2019, 9:22:48 p.m.'

In [4]: format_datetime(datetime.now(timezone.utc))                             
Out[4]: 'Jul. 2, 2019, 9:23:00 p.m.'

In [5]: format_datetime(datetime.now().astimezone())                            
Out[5]: 'Jul. 2, 2019, 5:23:09 p.m.'
```

ViktorHaag: I didn't know about `datetime.now().astimezone()` ... that seems like a great improvement!

tk0miya:
Merged. Thank you for comment!
I confirmed sphinx refers environment varaible `$TZ` to determine timezone :-)

PR #6533, opened 2019-06-29
LaTeX: refactor visit_enumerated_list() to use \sphinxsetlistlabels

### LINKED ISSUE #6511: Autonumbered list can not be configured to be alphanumeric in LaTeX since Sphinx 1.8.0
**Describe the bug**

Since #5001 was merged to fix #1431, Sphinx supports alphanumeric list items (a,b,c instead of 1,2,3) when this is explicitly present in the RST source.

If automatic numbering is used (with `#.`), lists are automatically numbered with 1,2,3.

Before 1.8.0 it was possible to change the automatic numbering using `enumitem` package. This is no longer possible because each `enumerate` environment defines its own `\theenumi`, `\theenumii` etc.

**To Reproduce**

A piece of RST:

```rst
This is a list with explicit numbers and letters

1.  Explicit number

    a. Explicit letter
    b. Explicit letter

2.  Explicit number

This is a list with implicit numbers and letters

#.  Implicit number

    #. Implicit number or letter?
    #. Implicit number or letter?

#.  Implicit number
```

With the following in the LaTeX preamble:

```latex
\usepackage{enumitem}
\setlist[enumerate,1]{label=\arabic*.}
\setlist[enumerate,2]{label=\alph*.}
```

Produces the following on Sphinx 1.8.0+:

Screen Shot 2019-06-20 at 09 34 15

**Expected behavior**

On Sphinx 1.7.9 and lower, the result respects the numbering type set by `enumitems`:

Screen Shot 2019-06-20 at 09 58 07

Because `\theenumi` etc is set by each `enumerate`, this is hard to work around by hooking into an environment in the preamble.

It would be great if this could be styled using a package like `enumitem`, but a configuration option in Sphinx is also possible.

jfbu:
Fixes: #6511

This is only refactoring. But as a result it provides way to fix #6511. As it is not yet decided if #6511 is a "bug" or not, I did not include any addition to CHANGES file.

I hesitated to ask merge to 2.1.3 as it is not clear if bugfix. But this is refactoring which is backwards compatible.

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -413,6 +413,18 @@
 \newcommand\sphinxsetup[1]{\setkeys{sphinx}{#1}}
 
 
+%% ALPHANUMERIC LIST ITEMS
+\newcommand\sphinxsetlistlabels[5]
+{% #1 = style, #2 = enum, #3 = enumnext, #4 = prefix, #5 = suffix
+ % #2 and #3 are counters used by enumerate environement e.g. enumi, enumii.
+ % #1 is a macro such as \arabic or \alph
+ % prefix and suffix are strings (by default empty and a dot).
+ \@namedef{the#2}{#1{#2}}%
+ \@namedef{label#2}{#4\@nameuse{the#2}#5}%
+ \@namedef{p@#3}{\@nameuse{p@#2}#4\@nameuse{the#2}#5}%
+}%
+
+
 %% MAXLISTDEPTH
 %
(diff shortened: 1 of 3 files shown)


tk0miya (commented):
LGTM!

tk0miya:
Note: 2.1.x is a branch for urgent releases. So normal bugs is better to merge 2.0 branch.

jfbu:
Thanks for reviewing. Rebased for merge to 2.0 branch, I will merge after testing completes.

PR #6543, opened 2019-06-30
Proposal: Add IndexDomain for database of general indices

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- So far, we've stored general index entries to env.indexentries.
- This moves them all to new `IndexDomain`.
- adapter and collector for indexentries will be also replaced (in later step).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,8 @@ Incompatible changes
 Deprecated
 ----------
 
+* ``sphinx.environment.BuildEnvironment.indexentries``
+* ``sphinx.environment.collectors.indexentries.IndexEntriesCollector``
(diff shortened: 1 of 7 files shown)


shimizukawa:
> A domain is a collection of markup (reStructuredText directives and roles) to describe and link to objects belonging together, e.g. elements of a programming language. 
http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

I thought "domain" is a set of directives and roles for each language, but "language" is just one use case.
Now I accept `IndexDomain`.

shimizukawa (approved):
LGTM

tk0miya:

PR #6560, opened 2019-07-07
Close #6558: glossary: emit a warning for duplicated glossary entry

### LINKED ISSUE #6558: Sphinx does not warn when glossary term appears twice
**Describe the bug**
Sphinx does not warn when glossary term appears twice

**To Reproduce**
Build the following document.
```
.. glossary::

   foo
       description

.. glossary::

   foo
       description
```

Then it succeeded without any errors. And ```:term:`foo` ``` refers last "foo" entry.

**Expected behavior**
Got warned for duplicated term.

**Your project**
None

**Screenshots**
None

**Environment info**
- OS: Mac
- Python version: 3.7.3
- Sphinx version: 2.2.0+
- Sphinx extensions:  None
- Extra tools: None

**Additional context**
None

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #6558

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -22,7 +22,7 @@ from docutils.statemachine import StringList
 
 from sphinx import addnodes
 from sphinx.addnodes import desc_signature, pending_xref
-from sphinx.deprecation import RemovedInSphinx40Warning
+from sphinx.deprecation import RemovedInSphinx40Warning, RemovedInSphinx50Warning
 from sphinx.directives import ObjectDescription
 from sphinx.domains import Domain, ObjType
 from sphinx.locale import _, __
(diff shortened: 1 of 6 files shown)


sphinx/domains/std.py
```diff
@@ -538,6 +540,23 @@ def __init__(self, env: "BuildEnvironment") -> None:
     def objects(self) -> Dict[Tuple[str, str], Tuple[str, str]]:
         return self.data.setdefault('objects', {})  # (objtype, name) -> docname, labelid
 
+    def note_object(self, objtype: str, name: str, labelid: str, location: Any = None
+                    ) -> None:
```
tk0miya, line 545: I added a new helper method to register an object to std domain. This does not provides `docname` argument like `add_object()`. I think it was wrong design. Extensions can register objects only during processing the document. So no `docname` argument is needed.

PR #6585, opened 2019-07-15
Use Python 3 in Quick Start example

stevepiercy:
Subject: Use Python 3 in Quick Start example

### Feature or Bugfix
- Feature

### Purpose
- Python 2 has been EOLed, and we should use Python 3 going forward.

### DIFF
--- a/sphinx/templates/quickstart/conf.py_t
+++ b/sphinx/templates/quickstart/conf.py_t
@@ -105,7 +105,7 @@ html_static_path = ['{{ dot }}static']
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/': None}
+intersphinx_mapping = {'https://docs.python.org/3/': None}
 {%- endif %}
 {%- if 'sphinx.ext.todo' in extensions %}
 
--- a/CHANGES
+++ b/CHANGES
@@ -221,8 +221,6 @@ Features added
 
 * #6180: Support ``--keep-going`` with BuildDoc setup command
 * ``math`` directive now supports ``:class:`` option
-* #6310: imgmath: let :confval:`imgmath_use_preview` work also with the SVG
-  format for images rendering inline math
 * todo: ``todo`` directive now supports ``:name:`` option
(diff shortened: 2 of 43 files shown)


sphinx/templates/quickstart/conf.py_t
```diff
@@ -105,7 +105,7 @@ html_static_path = ['{{ dot }}static']
 # -- Options for intersphinx extension ---------------------------------------
 
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/': None}
+intersphinx_mapping = {'https://docs.python.org/3/': None}
```
tk0miya, line 108: Why is this needed? If my understanding correct, `https://docs.python.org/` returns appreciate URL of python document. It is not only for 2.x.
```
$ curl -I https://docs.python.org/
HTTP/2 302
server: nginx
content-type: text/html
location: https://docs.python.org/3/
x-clacks-overhead: GNU Terry Pratchett
strict-transport-security: max-age=315360000; includeSubDomains; preload
via: 1.1 varnish
accept-ranges: bytes
date: Thu, 01 Aug 2019 15:47:56 GMT
via: 1.1 varnish
age: 112150
x-served-by: cache-jfk8140-JFK, cache-tyo19950-TYO
x-cache: HIT, HIT
x-cache-hits: 1, 3
x-timer: S1564674477.715171,VS0,VE0
content-length: 154
```

stevepiercy: It saves the network overhead of a second client request from the redirect: `HTTP/2 302`.

😆 : `x-clacks-overhead: GNU Terry Pratchett` oh, Python peeps are the best!

tk0miya: Hmm... The title and purpose of this PR are not related with change, right? If true, I can understand what are you doing. Indeed, this reduces overhead of communication.

tk0miya (changes requested):
Could you rebase this into 2.0 branch? Then I'll merge this soon.

stevepiercy:
Bugger. Closing in favor of a proper rebase in #6626. Also found a second link that was out of date and now uses https.

PR #6594, opened 2019-07-21
Pin to docutils<0.15

jab:
Every dependency should be pinned to an upper bound that represents
a compatible version. The previous docutils pin had no upper bound, and
just broke downstream users now that a new, incompatible docutils version
has come out (0.15), which users pull transitively via direct dependency on
Sphinx. Pin to docutils<0.15 until Sphinx is known to work with docutils 0.15
(at which point the pin should be bumped to "<0.16", keeping the pin within
a compatible range).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ install_requires = [
     'sphinxcontrib-qthelp',
     'Jinja2>=2.3',
     'Pygments>=2.0',
-    'docutils>=0.12',
+    'docutils>=0.12,<0.15',
     'snowballstemmer>=1.1',
     'babel>=1.3,!=2.0',
     'alabaster>=0.7,<0.8',


jab:
Via its unbounded dependency on docutils, Sphinx broke my project's CI for Python 2.7 by allowing the incompatible docutils 0.15 to satisfy its docutils dependency. Adding a direct dependency on docutils 0.14 worked around this successfully.

Workaround:
- https://github.com/jab/bidict/commit/210fe8f9844e16318fce65205de3eda3d56839be

Failing build before the workaround:
- https://travis-ci.org/jab/bidict/builds/561743755

Failing Python 2.7 jobs before the workaround:
- https://travis-ci.org/jab/bidict/jobs/561743757
- https://travis-ci.org/jab/bidict/jobs/561743759
- https://travis-ci.org/jab/bidict/jobs/561743760

Successful build after the workaround:
- https://travis-ci.org/jab/bidict/builds/561810706

jab:
See also https://semver.org/#spec-item-4
> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

From http://docutils.sourceforge.net/ it's not clear whether docutils attempts to follow semver. All the more reason to narrow the pin to known-good versions.

tk0miya:
The accident of docutils-0.15 was release trouble, not incompatible change. So there are no no special reason to pin version, IMO.

This trouble was caused by operation error. So similar trouble might be caused when bugfix release will be out. So pinning by exact version is needed if we determined to pin the version. But I think it is too much.

jab:
Please correct me if I'm wrong but I don't think this is a matter of interpretation. Leaving docutils' dependency declarations unpinned within compatible version ranges means that docutils users will break when any of your dependencies publishes an incompatible new version. This is not just a theoretical risk; this actually broke my project, which is why I opened this.

tk0miya:
If you're saying about compatibility of dependencies, all packages should be pinned too. There are no reason to pin the version of docutils only. requests and other libraries should be also pinned. Is this right?

If your concern is mainly about trouble of docutils-0.15, pinning does not help us. That release was caused by an unfortunate mistake, not comptibility problem. If you avoid the trouble, it would be better to pin all dependencies via exact match, not range. Because everybody sometimes goes mistakes. It can also happen even if bugfix release (like docutils-0.14.1). To do that, `pip freeze` and `requirements.txt` is better for this case.

Personally, I worry about pinning brings conflicts and stagnation. Sometimes, pinning causes conflicts on users' environment. Some library depends on `docutils < 0.15` and another one depends on `docutils >= 0.15`. As a result, they will conflict. I think this is very harmful. Another one, stagnation comes from lack of manpower. Who unpin it? We always need someones help. It would be better to less needless works...

mitya57:
I absolutely agree with @tk0miya here. I am a downstream (Debian) packager, and pinned dependencies cause us only troubles. If such dependencies are used in Sphinx, I would have to patch them out.

tk0miya:
I'm closing this now.
Thank you for proposal.

Issue #6613, opened 2019-07-30
Isolate section number in span tags for HTML output


yves-chevallier:
# Goal

Currently, the section numbers are part of the title: 

```html
<h2>
    9.1 Calling conventions
    <a class="headerlink" href="#calling-conventions">¶</a>
</h2>
```

It would be useful to have a custom CSS style for the numbers  and in this case, isolating the numbers woule be useful.

```html
<h2>
    <span class="numsect">9.1</span>Calling conventions
    <a class="headerlink" href="#calling-conventions">¶</a>
</h2>
```

## Implementation 

It seems the implementation would be quite straightforward, by modifying the `add_secnumber` method in the `html5` writer. 

```python
    def add_secnumber(self, node):
        # type: (nodes.Element) -> None
        if node.get('secnumber'):
            text = '.'.join(map(str, node['secnumber'])) + self.secnumber_suffix            
        elif isinstance(node.parent, nodes.section):
            if self.builder.name == 'singlehtml':
                docname = self.docnames[-1]
                anchorname = "%s/#%s" % (docname, node.parent['ids'][0])
                if anchorname not in self.builder.secnumbers:
                    anchorname = "%s/" % docname  # try first heading which has no anchor
            else:
                anchorname = '#' + node.parent['ids'][0]
                if anchorname not in self.builder.secnumbers:
                    anchorname = ''  # try first heading which has no anchor
            if self.builder.secnumbers.get(anchorname):
                numbers = self.builder.secnumbers[anchorname]
                text = '.'.join(map(str, numbers)) + self.secnumber_suffix

        self.body.append('<span class="secnum">%s</span>' % text)
```

tk0miya:
+1: Reasonable. I'll apply the patch if you make a PR :-)

PR #6622, opened 2019-08-01
Add get_secnumber to simplify add_secnumber

yves-chevallier:
Subject: isolate complexity of `add_secnumber` to fulfil #6613

### Purpose

Single responsability of method is not respected on `add_fignumber` which does:

- Retrieve the section number
- Insert it into the document

### Relates
- #6613

### DIFF
--- a/sphinx/writers/html5.py
+++ b/sphinx/writers/html5.py
@@ -277,11 +277,10 @@ class HTML5Translator(SphinxTranslator, BaseTranslator):
         # type: (nodes.Element) -> None
         self.depart_admonition(node)
 
-    def add_secnumber(self, node):
+    def get_secnumber(self, node):
         # type: (nodes.Element) -> None
         if node.get('secnumber'):
-            self.body.append('.'.join(map(str, node['secnumber'])) +
-                             self.secnumber_suffix)
+            return node['secnumber']
         elif isinstance(node.parent, nodes.section):
             if self.builder.name == 'singlehtml':
                 docname = self.docnames[-1]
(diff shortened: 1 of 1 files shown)


sphinx/writers/html5.py
```diff
@@ -277,11 +277,10 @@ def depart_seealso(self, node):
         # type: (nodes.Element) -> None
         self.depart_admonition(node)
 
-    def add_secnumber(self, node):
+    def get_secnumber(self, node):
         # type: (nodes.Element) -> None
```
tk0miya, line 281: This PR changes the return value of this method. Please update this also.

sphinx/writers/html5.py
```diff
@@ -293,9 +292,13 @@ def add_secnumber(self, node):
                 if anchorname not in self.builder.secnumbers:
                     anchorname = ''  # try first heading which has no anchor
             if self.builder.secnumbers.get(anchorname):
-                numbers = self.builder.secnumbers[anchorname]
-                self.body.append('.'.join(map(str, numbers)) +
-                                 self.secnumber_suffix)
+                return self.builder.secnumbers[anchorname]
+        return None
+
+    def add_secnumber(self, node):
```
tk0miya, line 298: Could you add type annotation here please?

tk0miya (changes requested):
LGTM with nits.
In addition, could you rebase this into 2.0 branch? Then it will be released as stable release (v2.x).

Thanks,

sphinx/writers/html5.py
yves-chevallier: What do you mean?

sphinx/writers/html5.py
yves-chevallier: `add_secnumber` is a new method, and `get_secnumber` was not returning anything. I do not understand your point :(

jfbu: @yves-chevallier I would say that tk0miya refers to the type annotation on line 281 which is now wrong because the method does not return None

tk0miya: It looks like `get_secnumber()` returns `node['secnumber']`, `self.builder.secnumbers[anchorname]` or `None`. So this should be annotated as `-> Tuple[int, ...]`.

sphinx/writers/html5.py
tk0miya: Pleasee append type annotation line like `# type: (nodes.Element) -> None` after function signature.

yves-chevallier:
Since rebased on branch 2.0, I made a new PR with the requested changes: #6632

Issue #6262, opened 2019-04-06
The display of the field list has changed


usaturn:
**Describe the bug**
This is a comparison of 1.8 and 2.0.
Please display like 1.8.

**To Reproduce**
```
Welcome to test's documentation!
================================

:spam: Spiced Ham
:ham: pork
:eggs: Chicken egg
```

```
make html
```

`html_theme = 'alabaster'` sphinx 1.8.5
![2019-04-06_15h09_25](https://user-images.githubusercontent.com/1503215/55665781-ed7d1800-587f-11e9-92ae-c29ef368a806.png)

`html_theme = 'alabaster'` sphinx 2.0
![2019-04-06_15h10_32](https://user-images.githubusercontent.com/1503215/55665776-e6eea080-587f-11e9-84f1-89f93fc20202.png)

`html_theme = 'bizstyle'` 1.8.5
![2019-04-06_15h11_32](https://user-images.githubusercontent.com/1503215/55665786-038ad880-5880-11e9-97a3-dd2329c06bb1.png)

`html_theme = 'bizstyle'` 2.0
![2019-04-06_15h11_01](https://user-images.githubusercontent.com/1503215/55665791-12718b00-5880-11e9-8ee1-48b82ff22846.png)

**Your project**
[tmp.zip](https://github.com/sphinx-doc/sphinx/files/3050222/tmp.zip)

**Environment info**
- OS: Windows10
- Python version: 3.6.4
- Sphinx extensions: none

tk0miya:
Fixed by #6266.
Thank you for reporting!

PR #6625, opened 2019-08-02
Fix field lists after switch to HTML5 writer

### LINKED ISSUE #6604: Field list rendering style with HTML5 writer
**Describe the bug**
#6262 mentioned a problem in the Bizstyle theme, but it seems to apply to all themes as 12e65061f57e78fdfecead0d36b51107ddccd495 introduced the extra CSS in the basic theme. I think the fix #6266, or a variation, should be applied in the basic theme instead, though my CSS knowledge is rather limited.

**To Reproduce**
Use the archive in #6266 with different themes (e.g., ``classic``) and Sphinx 1.8 and master. The tabular rendering of fields is fine with 1.8, but broken with master (except in the Bizstyle theme).

jakobandersen:
### Feature or Bugfix

- Bugfix

Subject: fix HTML rendering of field lists after original switch to the HTML5 writer

### Purpose
See #6604. There was a general problem with alignment between dt dd elements, i.e., field type and field content. The classic theme additionally was missing some colour.

Themes affected/fixed due to changes in basic:
- bizstyle (only technically, it already was fixed in #6262)
- haiku
- classic
- nature
- pyramid
- sphinxdoc

See [field_lists.zip](https://github.com/sphinx-doc/sphinx/files/3461293/field_lists.zip) for screen shots of for version 1.8, 2.13, and with this PR.

Note that both haiku and pyramid still has some alignment problems, though the seem to be present in 1.8 as well.

### Relates
Fixes #6604
Perhaps fixes sympy/sympy#16585

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -520,14 +520,15 @@ dl.citation > dd:after {
 }
 
 dl.field-list {
-    display: flex;
-    flex-wrap: wrap;
+    display: grid;
+    grid-template-columns: fit-content(30%) auto;
 }
 
 dl.field-list > dt {
-    flex-basis: 20%;
     font-weight: bold;
     word-break: break-word;
+    padding-left: 0.5em;
+    padding-right: 5px;
 }
 
 dl.field-list > dt:after {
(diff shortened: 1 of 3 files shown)


tk0miya (approved):
LGTM!

Issue #6618, opened 2019-07-31
LaTeX: Avoid section names at the end of a page


sebhub:
I have a large document with lots of small sections. Currently, the likelihood that a section name is the last content on a page is quite high. This looks ugly. One way to prevent this is the following patch:
```patch
diff --git a/sphinx/texinputs/sphinx.sty b/sphinx/texinputs/sphinx.sty
index 2eebbb914..9198d739b 100644
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -40,7 +40,7 @@
 % for \text macro and \iffirstchoice@ conditional even if amsmath not loaded
 \RequirePackage{amstext}
 \RequirePackage{textcomp}% "warn" option issued from template
-\RequirePackage{titlesec}
+\RequirePackage[nobottomtitles*]{titlesec}
 \@ifpackagelater{titlesec}{2016/03/15}%
  {\@ifpackagelater{titlesec}{2016/03/21}%
   {}%
```

jfbu:
I am interested in some concrete source exhibiting the problem in latexpdf output. I tried such source

```rest
===============================
Welcome to FOO's documentation!
===============================

===
Foo
===

.. raw:: latex

   \pdfsetrandomseed 12345678

Section
=======

A

.. raw:: latex

   \vspace{\pdfuniformdeviate 10 mm}


B

```

with the section repeated 100 times to create some random spacing. But the circa 20 pages output has not a single isolated section title at a bottom of a page.

However maybe your use case has some lists or images after section titles?

jfbu:
For a stronger test use `\pdfuniformdeviate 7 cm` rather. (for all 100 sections).

The point is that LaTeX has a mechanism for not allowing easily a section title to be left alone at bottom of page if followed by regular text. I am thus interested in concrete example (a screen shot at least).

sebhub:
Yes, I was a bit surprised to see this in a LaTeX output. I created the attached example with `sphinx-quickstart`. It seems to be related to the `.. topic::` directive.
[index.txt](https://github.com/sphinx-doc/sphinx/files/3453322/index.txt)
[test.pdf](https://github.com/sphinx-doc/sphinx/files/3453327/test.pdf)

jfbu:
Thanks, indeed the topic directive uses in latex rendering a framed box which encourages a pagebreak before it even if the element is a section title.

Your workaround can be implemented in conf.py of your project:

```
[...]

PR #6633, opened 2019-08-05
LaTeX: Avoid section titles at the end of a page

sebhub:
The topic directive uses in LaTeX rendering a framed box which
encourages a pagebreak before it even if the element is a section title.
This can be avoided with the nobottomtitles option of the titlesec
package.

Fixes: #6618

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -40,7 +40,7 @@
 % for \text macro and \iffirstchoice@ conditional even if amsmath not loaded
 \RequirePackage{amstext}
 \RequirePackage{textcomp}% "warn" option issued from template
-\RequirePackage{titlesec}
+\RequirePackage[nobottomtitles*]{titlesec}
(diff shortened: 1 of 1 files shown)


jfbu (changes requested):
Thanks. Could you please rebase this to 2.0 branch (after doing it locally, force push, then edit the PR to change the branch to merge to), so that it can be merged to become part of 2.3.0 release ? (It is some kind of a bug fix indeed but not an urgent one, and I don't feel it adds strong backwards incompatibilities although it may slightly modify PDFs, overall it looks like it should rather improve output; and I checked the option has long been available (since 1999) in the `titlesec` package).

**edit** I have now rebased your branch on tip of 2.0. Hope you don't mind.

jfbu (approved):

jfbu:
Thanks for contribution!

sebhub:
Thanks, for rebasing it for me. I am still recovering from holidays.

PR #6635, opened 2019-08-06
ENH: Upgrade jquery

larsoner:
I thought I found a bug in my docs due to `jQuery` but it turned out just to be bad CSS.

But in debugging I tried upgrading `jQuery` to 3.4.1. If this is useful feel free to merge. It probably contains bugfixes of some sort. If it isn't worth upgrading jQuery until someone finds a need to do so, feel free to close.


tk0miya:
Same as #6609. Thanks,

PR #6648, opened 2019-08-16
Specify the target Python version in the mypy configuration

jdufresne:
Currently targets Python 3.5.

This caught a bug. The `subprocess.run()` function only started taking the 'encoding' keyword argument starting with Python 3.6.

https://docs.python.org/3/library/subprocess.html#subprocess.run


tk0miya:
Thank you always!

PR #6653, opened 2019-08-17
Switch uses of __import__ to importlib.get_module()

jdufresne:
The Python docs for `__import__` recommend using `importlib.get_module()`.

https://docs.python.org/3/library/functions.html#__import__

> Note: This is an advanced function that is not needed in everyday Python programming, unlike `importlib.import_module()`.

As `importlib.get_module()` uses the Python module cache and returns the module, this also allows simplifying many module cache checks of use of `sys.modules`.

`importlib.get_module()` has been available since Python 3.3.

### DIFF
--- a/sphinx/deprecation.py
+++ b/sphinx/deprecation.py
@@ -31,10 +31,7 @@ RemovedInNextVersionWarning = RemovedInSphinx30Warning
 
 def deprecated_alias(modname, objects, warning):
     # type: (str, Dict, Type[Warning]) -> None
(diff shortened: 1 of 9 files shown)


tk0miya (changes requested):
LGTM. Could you rebase this into 2.0 branch?

jdufresne:
Sure thing. Rebased.

tk0miya (approved):
Thank you for quick work!

PR #6666, opened 2019-08-21
Add support for booktabs-style tables to LaTeX builder

sephalon:
Subject: Add support for booktabs-style tables to LaTeX builder

### Feature or Bugfix
- Feature

### Purpose
Render tables without vertical rules and horizontal rules of varying
thickness (with additional space above and below) using the booktabs
package.

Feature is not polished yet (see below), but should be enough to keep the discussion started.

### TODO
- Support column spans in headers (via writing `\cmidrule` commands)
- Allow selective enabling/disabling of booktabs-style tables via a custom option (like `:latex-booktabs:`)
- longtable test fails, need to have a look into that


jfbu:
+1, reasonable. The loading of LaTeX package `booktabs`  itself should propbably be made conditional on the Python configuration `latex_booktabs` you are adding. I can't help immediately but will have some time in the coming weeks.

sphinx/texinputs/sphinx.sty
```diff
@@ -65,6 +65,7 @@
    \fi
   }%
  }{}
+\RequirePackage{booktabs}
```
jessetan: This is not needed if `latex_booktabs` is `False`

tk0miya:
Does anyone know other (well-used) table style? If anything exists, `latex_table_style` might be better slightly for the name of configuration to me.

sphinx/writers/latex.py
```diff
@@ -1123,6 +1127,12 @@ def visit_table(self, node):
                 logger.info(__('both tabularcolumns and :widths: option are given. '
                                ':widths: is ignored.'), location=node)
         self.next_table_colspec = None
+        if self.builder.config.latex_booktabs:
+            self.table.booktabs = True
+            self.table.colsep = ''
+        else:
+            self.table.booktabs = False
+            self.table.colsep = '|'
```
tk0miya: Could you initialize these members inside `Table.__init__()` please? I think defining new members outside of the class is not good.

sphinx/templates/latex/longtable.tex_t
```diff
@@ -10,25 +10,51 @@
 <%- if table.caption -%>
 \sphinxthelongtablecaptionisattop
 \caption{<%= ''.join(table.caption) %>\strut}<%= labels %>\\*[\sphinxlongtablecapskipadjust]
+<%- if table.booktabs -%>
```
tk0miya: I'm thinking about new templates set is needed for booktabs because these if-block makes templates complex. I worry that it will be hard to support yet another table style.

tk0miya (changes requested):
+1 for your idea! But I feel some refactoring is needed for this. And it seems testing are failed. Could you check it please? I'll take a look later if I have time.

In addition, new testcase is needed for this option. Please add them later after discussion.

lcnittl:
Not very sure if this is somehow useful, but it seems that [docutils has a booktabs option implemented](https://docutils.sourceforge.io/docs/user/config.html#id58)

sephalon:
Sorry, I got sidetracked by other projects… I will rebase this and have a look at your comments.

@tk0miya, what is the status of #7417? It seems to go into a similar direction, but has been closed recently without being merged. Is there a specific reason for that?

sphinx/templates/latex/longtable.tex_t
jfbu: A possibility would be to use in the template `\sphinxtoprule`, `\sphinxmidrule`, `\sphinxbottomrule` with `\def\sphinxtoprule{\hline}` etc... and other definitions with booktabs triggering option.

sphinx/writers/latex.py
```diff
@@ -330,16 +330,20 @@ def get_colspec(self):
...
-            return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
+            return '{%s*{%d}{\\X{1}{%d}%s}}\n' % (self.colsep, self.colcount,
+                    self.colcount, self.colsep)
         elif self.get_table_type() == 'tabulary':
             # sphinx.sty sets T to be J by default.
-            return '{|' + ('T|' * self.colcount) + '}\n'
+            return '{' + self.colsep + (('T' + self.colsep) * self.colcount) + '}\n'
         elif self.has_oldproblematic:
-            return '{|*{%d}{\\X{1}{%d}|}}\n' % (self.colcount, self.colcount)
+            return '{%s*{%d}{\\X{1}{%d}%s}}\n' % (self.colsep, self.colcount,
```
jfbu: The definitions of the `\X` and `\Y` column types in sphinxlatextables.sty need adjustment probably as they use `\arrayrulewidth` (except if booktabs set this to zero, I must check; but about this I think you must allow not using vertical rules without tying user to using booktabs package, so the better terminology throughout the LaTeX writer would be not to refer to booktabs but to something such as "nocolsep").  Some thinking required, depositing this comment here.

jfbu:
I am sorry about very long delay.  But the world of LaTeX tables is a terrible jungle mess. I am worried about setting a precedent if we provide a LaTeX package specific adaptation, what if it turns incompatible with using some other package which adds colours to rows etc... There are some other things in the code base which are tied to using default columns lines and need adjustments.  Sphinx has a sophisticated architecture with no equivalent in the world of LaTeX table packages to handle grid tables with merged cells.  Ideally, the safest would not to worry about external LaTeX packages and provide all necessary customization hooks internally.  But this probably means maintenance becomes impossible because requires to much knowledge of LaTeX and TeX macros and primitives. 

Anyway I will try to focus my available Sphinx time on this and possibly push to this branch. (hopefully, no promises made..)

jfbu:
> Not very sure if this is somehow useful, but it seems that [docutils has a booktabs option implemented](https://docutils.sourceforge.io/docs/user/config.html#id58)

Indeed. It is actually a value we can pass as one of a comma separated class values to the `table_style` option of `[latex writers]`. This raises question whether we should support a similar `latex_table_style`.  Notice that I tested passing `borderless` and it removes all borders in PDF output using Docutils `rst2latex.py` but their [documentation](https://docutils.sourceforge.io/docs/user/config.html#table-style) says
```
borderless

    No borders around the table.
```
ah well, html output also removes everything.  Anyway I wanted to check if the Docutils class values were additive or mutually exclusive, but here combining `booktabs` and `borderless` is not very illustrative as `borderless` removes all borders anyhow.  Anyway, let's worry about this later.

jfbu:
I am in the process of rebasing this on 5.x branch, hopefully for 5.2.0 (as it is a feature addition I think it is suitable for 5.2.0 if we maintain exact backwards compatibility).

jfbu:
well, mypy chokes on the `colspec` in the LaTeX writer Python code probably in relation to [this comment](https://github.com/sphinx-doc/sphinx/pull/6666#discussion_r336741610) and I fill fix this later (perhaps tomorrow).

jfbu:
@tk0miya I rebase don 5.x but I hesitate if I should keep this PR for merge to master.  Indeed I expect to be replacing `\hline` mark-up in tex output by `\sphinxtoprule`, `\sphinxmidrule`,  `\sphinxbottomrule` defaulting to `\hline` without the booktabs activation.  This means changing the output mark-up in tex file.  I have also changed output mark-up recently at 5.1.0 for implementing rounded boxes for code blocks but I am not so sure I can do this on 5.x liberally.

jfbu:
I don't know what causes the linkchek test failures I have seen recently often: https://github.com/sephalon/sphinx/runs/7706483793?check_suite_focus=true#step:8:2088.  This time at Python 3.10 but on a nother recent occasion it was Python 3.7.  And after merge no failure was reported.

jfbu:
I have reimplemented the whole thing.  Apart from removing `|` from the column specifications and related collaterals, the table mark-up is the same whether the `latex_use_booktabs` option is True or False: the `\sphinxtoprule` etc... will expand to the `booktabs` `\toprule` only if the option was True.

jfbu:
Ready for review.  

TBH I think this option will start making sense only when "zebra-style" colouring for tables will also be available (which however may be tricky for complex tables with merged cells, but we can perhaps simply warn).
(relates: #6671).

jfbu:
I am closing this as it has been transfered to #10759 for further development.

PR #6671, opened 2019-08-26
Add support for zebra-striped tables to LaTeX builder

sephalon:
Subject: Add support for zebra-striped tables to LaTeX builder

### Feature or Bugfix
- Feature

### Purpose
Render tables with alternating background colors for even and odd rows
(so called "zebra striping").

Most useful with #6666.

### TODO

- Allow selective enabling/disabling of zebra-striped tables via a custom option (like :latex-zebra-stripes:)

### DIFF
--- a/doc/latex.rst
+++ b/doc/latex.rst
@@ -646,6 +646,18 @@ macros may be significant.
 
     .. versionadded:: 1.6.6
 
+``RowEvenColor``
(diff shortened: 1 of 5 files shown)


jfbu:
Closing because this was cherry-picked to #10759 for further development.

PR #6679, opened 2019-08-29
agogo theme: ensure hidden toctree is show on Table of Contents

carandraug:
The hidden option for the toctree directive is meant to hide from the
document at the location of the directive.  This for when the links
are to be inserted in the HTML sidebar.  However, since version 1.2,
the toctree() template function requires the includehidden=True
option.

---

Using the agogo theme, if a document has the toctree set to hidden (because one wants to have it displayed on the sidebar instead of inserted in the document), then the Table of Contents on the side bar will be empty. This fixes it by setting `includehidden=True` when calling `toctree()` template function. This is required since sphinx 1.2.

I am not sure if the same should be done to `sphinx/themes/basic/globaltoc.html`

### DIFF
--- a/sphinx/themes/agogo/layout.html
+++ b/sphinx/themes/agogo/layout.html
@@ -44,7 +44,7 @@
         <div class="sidebar">
           {%- block sidebartoc %}
           <h3>{{ _('Table of Contents') }}</h3>
-          {{ toctree() }}
+          {{ toctree(includehidden=True) }}
           {%- endblock %}
           {%- block sidebarsearch %}
           <div role="search">


carandraug:
Hi. This PR was closed but seems that the issue is not fixed yet. Did I miss something?

AA-Turner:
I meant to label this PR rather than close it -- thanks for catching. I've gone through the others I triaged & seen no other similar errors.

A

JulienPalard:
CI failing due to https://github.com/psf/requests/issues/6140

PR #6681, opened 2019-09-01
Fix #6668: LaTeX: Longtable before header has incorrect distance

### LINKED ISSUE #6668: Longtable before header has incorrect distance
**Describe the bug**
A longtable before a header produces smaller vertical whitespace than other content before a header

**To Reproduce**

`index.rst`:
```rst
************
Some chapter
************

Header distance
===============

.. list-table:: Long table before text
    :class: longtable

    * - a
      - b
      - c
    * - 1
      - 2
      - 3

This is some content that is not a longtable. bla bla bla bla bla bla

Normal header distance
----------------------

.. list-table:: Long table before not long table
    :class: longtable

    * - a
      - b
      - c
    * - 1
      - 2
      - 3

.. list-table:: Table not longtable

    * - a
      - b
      - c
    * - 1
      - 2
      - 3

Normal header distance again
----------------------------

.. list-table:: Long table before long table
    :class: longtable

    * - a
      - b
      - c
    * - 1
      - 2
      - 3

.. list-table:: Long table before header
    :class: longtable

    * - a
      - b
      - c
    * - 1
      - 2
      - 3

SMALLER HEADER DISTANCE AFTER LONGTABLE
---------------------------------------

foo foo foo foo
```

**Result**

The distance longtable-header == distance longtable-text and the distance text-header == distance table-header.

Screen Shot 2019-08-22 at 15 23 37

**Expected behavior**

The distance longtable-header == distance text-header == distance table-header.

**Additional info**

I've tried to work around this by messing with `\sphinxatlongtableend` and `\sphinxtablepost` but did not succeed since this only happens before a header and not in all situations

Possibly related to #3504

**Environment info**
- Python version: 3.7.3
- Sphinx version: 2.2.0

jfbu:
Bugfix

### Relates
-  latex3/latex2e#173

- #6668

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -6,7 +6,7 @@
 %
 
 \NeedsTeXFormat{LaTeX2e}[1995/12/01]
-\ProvidesPackage{sphinx}[2019/06/04 v2.1.1 LaTeX package (Sphinx markup)]
+\ProvidesPackage{sphinx}[2019/09/01 v2.3.0 LaTeX package (Sphinx markup)]
 
 % provides \ltx@ifundefined
 % (many packages load ltxcmds: graphicx does for pdftex and lualatex but
@@ -119,7 +119,8 @@
    {\dimexpr-\dp\strutbox
             -\spx@ifcaptionpackage{\abovecaptionskip}{\sphinxbaselineskip}%
             +\sphinxbelowcaptionspace\relax}%
-\def\sphinxatlongtableend{\prevdepth\z@\vskip\sphinxtablepost\relax}%
+\def\sphinxatlongtableend{\@nobreakfalse % latex3/latex2e#173
+    \prevdepth\z@\vskip\sphinxtablepost\relax}%
 % B. Table with tabular or tabulary
 \def\sphinxattablestart{\par\vskip\dimexpr\sphinxtablepre\relax}%
 \let\sphinxattableend\sphinxatlongtableend
--- a/CHANGES
+++ b/CHANGES
@@ -19,6 +19,11 @@ Features added
 Bugs fixed
 ----------
 
+* #6668: LaTeX: Longtable before header has incorrect distance
+  (refs: `latex3/latex2e#173`_)
+
+.. _latex3/latex2e#173: https://github.com/latex3/latex2e/issues/173
+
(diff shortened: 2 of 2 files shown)


sphinx/texinputs/sphinx.sty
```diff
@@ -119,7 +119,8 @@
    {\dimexpr-\dp\strutbox
             -\spx@ifcaptionpackage{\abovecaptionskip}{\sphinxbaselineskip}%
             +\sphinxbelowcaptionspace\relax}%
-\def\sphinxatlongtableend{\prevdepth\z@\vskip\sphinxtablepost\relax}%
+\def\sphinxatlongtableend{\@nobreakfalse % latex3/latex2e#173
```
jfbu, line 122: Sphinx LaTeX uses `parskip` package which sets the paragraph indentation to zero. For the record, let me point out that this patch induces following change if paragraph indentation `\parindent` is not zero:

After a section title, LaTeX suppresses the indentation of the first paragraph following it,
Without this patch, if a longtable immediately follows a section title, the LaTeX mechanism applies to suppress indentation of first text paragraph after longtable,
With this patch, in the `title then longtable then text` situation, the text will get indented, rather.

I don't think we should worry about it, as anyhow Sphinx LaTeX defaults to setting `\parindent` to `0pt` (btw, over the last few years some effort went into making various LaTeX code of Sphinx compatible with a non-zero `\parindent`, which it was not formerly). And there are various easy means for advanced Sphinx LaTeX users to revert this, either e.g. by redefining `\sphinxatlongtableend` or by using a modified longtable template file in their project.

Besides, this made me realize advanced Sphinx LaTeX users working with positive `\parindent` currently face an inconsistency between cases `title -> tabular -> text` and `title -> longtable -> text` as in former case indentation applies but in latter case indentation does not apply.

With this patch it applies in both cases and the inconsistency is resolved.

jfbu: `\sphinxattableend` is defined in `sphinx.sty` as an identical copy of `\sphinxatlongtableend` hence inherits the `\@nobreakfalse`. This does not matter because the code for `tabular` or `tabulary` anyhow triggers the LaTeX paragraph builder hence does it already. Thus there was no need to give `\sphinxattableend` a separate definition.

Issue #6674, opened 2019-08-26
sphinx-apidoc documentation has a bad comma after EXCLUDE_PATTERN


Jacob-Stevens-Haas:
TL; DR: remove the comma after \`_EXCLUDE\_PATTERN_\` in the sphinx-apidoc man page and help text.

**Describe the bug**
__sphinx-apidoc__ sees a comma at the end of exclude pattern as part of fnmatch style string.  Recommend removing the comma from the syntax on the man page and --help response, or changing the program to ignore a trailing comma in exclude patterns

**To Reproduce**
```
> mkdir mwe
> cd mwe
> mkdir docs
> type nul > ignore1.py # aka touch ignore1.py
> type nul > ignore2.py
> type nul > keep.py
> cd docs
> sphinx-quickstart #Choose separate build & source directories
> sphinx-apidoc -o source ../ ../ignore1.py, ../ignore2.py
Creating file source\ignore1.rst.
Creating file source\keep.rst.
Creating file source\modules.rst.
```

**Expected behavior**
Expect either the documentation to lose the comma, or sphinx-apidoc to treat the trailing comma in `../ignore1.py,` as a separator between arguments and not part of the match string.  Currently, the correct command is
```
> sphinx-apidoc -f -o source ../ ../ignore1.py ../ignore2.py
Creating file source\keep.rst.
Creating file source\modules.rst.
```

**Environment info**
- OS: Windows 10
- Python version: 3.7.3
- Sphinx version: 2.1.2
- Sphinx extensions:  none.  Bare bones from `sphinx-quickstart`
- Extra tools: none

**Additional context**
I hope this doesn't come across as cavil, but it caused a lot of confusion for my colleague and me. The [man page for sphinx-apidoc][spa] writes the syntax of the command as:

__sphinx-apidoc__ [_OPTIONS_] -o <_OUTPUT\_PATH_> <_MODULE\_PATH_> [_EXCLUDE\_PATTERN_, …]

That comma at the end of \`_EXCLUDE\_PATTERN_,\` confused me.  It isn't necessary; sphinx-apidoc sees the comma as part of the fnmatch style string, meaning that only my last exclude pattern was matching the files I wanted it to.  For comparison, the [man page for sphinx-build][spb] does not use the spurious comma:

__sphinx-build__ [_options_] <_sourcedir_> <_outputdir_> [_filenames_ …]

[...]

PR #6682, opened 2019-09-02
Get rid of comma in sphinx-apidoc documentation

Jacob-Stevens-Haas:
Subject: Get rid of comma in sphinx-apidoc documentation

### Feature or Bugfix
- Bugfix

### Purpose
- Nixing the comma in documentation for sphinx-apidoc (`EXCLUDE PATTERN, ...`) because it's not supposed to be there.  Experienced coders will omit the comma automatically, but beginners will struggle with it.  It's not doctrinaire and it's also against the style of sphinx-build documentation, which does not include a comma to separate arguments.  Basically, if I struggled for an hour with this, I'm sure others have too.

### Detail
-  See the **additional context** section of the cited issue below for references to style guides and stackoverflow questions on the topic.

### Relates
- #6674

### DIFF
--- a/doc/man/sphinx-apidoc.rst
+++ b/doc/man/sphinx-apidoc.rst
@@ -5,7 +5,7 @@ Synopsis
 --------
 
 **sphinx-apidoc** [*OPTIONS*] -o <*OUTPUT_PATH*> <*MODULE_PATH*>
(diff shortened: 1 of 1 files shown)


tk0miya:

Issue #6710, opened 2019-09-30
LaTeX not well configured for Greek language as main language


jfbu:
**Describe the bug**

When using `language = 'el'` and using latexpdf, the PDF will be ok only if entirely in Greek. Any Latin letter will become converted into a Greek letter. Besides there are problems with Unicode letters even as common as © or ™, due to a [current bug](https://github.com/latex3/babel/issues/30) in LaTeX package `babel` for non-English support.

Thus, Sphinx should use `xelatex` per default.

However there is currently an [issue](https://github.com/reutenauer/polyglossia/issues/306) in LaTeX package `polyglossia` which requires to incorporate a workaround.

And our documentation should explicitly state that the advices for supporting small Greek snippets do not apply when language is `'el'`.

**Environment info**
- OS: [e.g. Unix/Linux/Mac/Win/other with version]
- Python version: [e.g. 3.7.1]
- Sphinx version: 2.0+
- Sphinx extensions:  [e.g. sphinx.ext.autodoc, recommonmark]
- Extra tools: TeXLive2019

**Additional context**

There was [discussion](https://groups.google.com/d/msgid/sphinx-users/qmrb5l%247jjq%241%40blaine.gmane.org) about this at the sphinx-users mailing list

PR #6711, opened 2019-09-30
Add latexpdf support for Greek as main language (via xelatex use)

jfbu:
Closes: #6710

### Feature or Bugfix
- Bugfix

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1841,7 +1841,17 @@ These options influence LaTeX output.
    ``'xelatex'`` or ``'lualatex'`` and making sure to use an OpenType font
    with wide-enough glyph coverage is often easier than trying to make
    ``'pdflatex'`` work with the extra Unicode characters.  Since Sphinx 2.0
-   the default is the GNU FreeFont which covers well Latin, Cyrillic and Greek.
+   the default is the GNU FreeFont which covers well Latin, Cyrillic and
+   Greek.
+
+   .. versionchanged:: 2.1.0
+
+      Use ``xelatex`` (and LaTeX package ``xeCJK``) by default for Chinese
+      documents.
+
+   .. versionchanged:: 2.2.1
+
+      Use ``xelatex`` by default for Greek documents.
 
    Contrarily to :ref:`MathJaX math rendering in HTML output <math-support>`,
    LaTeX requires some extra configuration to support Unicode literals in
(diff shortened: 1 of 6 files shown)


jfbu:
I have squashed the commits. Python 3.6 mypy testing fails for reasons unrelated to this PR.

doc/usage/configuration.rst
```diff
@@ -1841,7 +1841,17 @@ These options influence LaTeX output.
...
-   the default is the GNU FreeFont which covers well Latin, Cyrillic and Greek.
+   the default is the GNU FreeFont which covers well Latin, Cyrillic and
+   Greek.
+
+   .. versionchanged:: 2.1.0
+
+      Use ``xelatex`` (and LaTeX package ``xeCJK``) by default for Chinese
+      documents.
+
+   .. versionchanged:: 2.2.1
```
tk0miya, line 1852: Now, we are using x.y.Z branch for urgent bugfix release. I think 2.0 branch is appropriate for this fix.

sphinx/builders/latex/__init__.py
```diff
@@ -418,6 +418,8 @@ def default_latex_engine(config: Config) -> str:
         return 'platex'
     elif (config.language or '').startswith('zh'):
         return 'xelatex'
+    elif (config.language or '') == 'el':
```
tk0miya: `or ''` is not needed for this case. It is useful to call methods of string object even if `config.language` is None. But this line simply compares with string.

tk0miya (changes requested):
LGTM with nits.

doc/usage/configuration.rst
tk0miya: Ah, sorry. I just read #6710 and understand this is a bug. Okay, go with 2.2.1.

tk0miya:
>I have squashed the commits. Python 3.6 mypy testing fails for reasons unrelated to this PR.

I fixed the error in latest 2.2.1 branch today. Please merge it into this. Then error will go away.

sphinx/builders/latex/__init__.py
jfbu: blindly copying pasting is fine, but thinking is better, thanks for the explanation :) will fix

jfbu:
Thanks for reviewing @tk0miya now merging into 2.2.1

sulaeman512 (approved):
To android ku

PR #6727, opened 2019-10-10
Close #267: html: Eliminate prompt characters of doctest block from c…

### LINKED ISSUE #267: Eliminate prompt characters in examples from copyable text
It would be nice if the prompt characters in examples were easily copyable and pasteable somehow. For example, if this showed up on the document:

```
>>> for i in xrange(32)
...     print i
```

And you selected it, only these bits were selected.

```
for i in xrange(32)
    print i
```

This would make it easier to actually paste the examples into an interpreter.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/267
- Originally reported by: Michael Hoffman
- Originally created at: 2009-10-06T17:05:39.855

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #267

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,7 @@ Features added
 --------------
 
 * #6707: C++, support bit-fields.
+* #267: html: Eliminate prompt characters of doctest block from copyable text
 
 Bugs fixed
 ----------
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -672,6 +672,10 @@ div.code-block-caption + div > div.highlight > pre {
     margin-top: 0;
 }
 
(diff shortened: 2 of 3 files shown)


ankostis:
That should be a goof fix till the javascript solution allows more flexibility to enable/disable copiable content.

tk0miya:
Merged. This will be released as 2.3.0 later.

Note: I'm afraid that JS helper (like python-doc does) effects to HTML family builders (EPUB, HTMLHelp, AppleHelp and so on.). So I'd not like to add it for now.

become-nice:
May i ask a question ? how to config it ? I can't find any information in docs.

tk0miya:
No configurations. This change is always applied.

PR #6741, opened 2019-10-16
Fix search bug with empty html_file_suffix using html builder

rdb:
searchtools.js erroneously assumes that when FILE_SUFFIX is empty, the dirhtml builder is used.  This results in it erroneously appending a '/' to target links if the regular html builder is used with an empty `html_file_suffix`.

This is fixed in this PR by using null instead of the empty string in the case that the dirhtml builder is used.  Though, perhaps it would be cleaner to leave FILE_SUFFIX alone and instead rely on a separate BUILDER property?

### DIFF
--- a/sphinx/builders/dirhtml.py
+++ b/sphinx/builders/dirhtml.py
@@ -9,7 +9,7 @@
 """
 
 from os import path
-from typing import Any, Dict, Set
+from typing import Any, Dict
 
 from sphinx.application import Sphinx
 from sphinx.builders.html import StandaloneHTMLBuilder
@@ -44,10 +44,6 @@ class DirectoryHTMLBuilder(StandaloneHTMLBuilder):
                                     'index' + self.out_suffix)
 
         return outfilename
-
-    def prepare_writing(self, docnames: Set[str]) -> None:
-        super().prepare_writing(docnames)
-        self.globalcontext['no_search_suffix'] = True
 
 
 # for compatibility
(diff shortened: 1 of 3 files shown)


tk0miya:
It is not good design to use one variable for two different purpose to me. I feel separating it into two variables. How about adding a new variable `DIRHTML: true` into `documentation_options.js` instead?

rdb:
I agree.  I have changed the PR to instead add a BUILDER variable, which requires no additions to the context and may be more flexible for future use (since it can also distinguish singlehtml).

I've changed the base for the PR to the 2.0 branch.

tk0miya (approved):
LGTM

tk0miya:

blueyed:
@rdb 
This appears to have caused a regression with RTD builds, when using the dirhtml builder there.
(I've tested this commit (https://github.com/sphinx-doc/sphinx/commit/254624c8d) and the one before there, noticed it when upgrading)

I've tested https://github.com/sphinx-doc/sphinx/pull/7278 then also (https://github.com/sphinx-doc/sphinx/commit/fdd8955d1246b2013a8401f94ebc04418725d3dd), but it still appends .html to search results there.

The workaround would be to also setup redirections, but that should not be needed.

After looking closer at it this is caused by them using "-b readthedocsdirhtml", and that not being matched with this patch..!

blueyed:
So it seems like using "DIRHTML" explicitly would be better.

=> https://github.com/sphinx-doc/sphinx/pull/7283

PR #6748, opened 2019-10-20
Fix #6742: end-before option of literalinclude skips first line

### LINKED ISSUE #6742: :end-before improvement for literalinclude
With `literalinclude`, `:end-before:` would be more useful if started matching *after* the first line of the block so that it doesn't generate an empty block.

Illustration:

My input file has the usual ini file syntax, and I want to quote one section in the page. (Similar examples would include quoting one class or struct from a source file containing several class/struct definitions, or generally, quoting from one delimiter line up to the next similar delimiter line). The input file:

```
# file someconfig.ini
[Config Foo]
one = 111
two = 5543
three = 23

[Config Bar]
four = 444
etc = 345
```

And the document source:

```
.. literalinclude:: someconfig.ini
   :start-at: Config Foo
   :end-before: Config
```

Currently it produces an empty block because `:end-before:` matches immediately, but it would be much better if it stopped at the `[Config Bar]` line and included the whole `Foo` section. The current behavior is not that useful (who wants to purposefully add an empty block in the output?)

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #6742 
- This is incompatible change. But I think this change does not harm users because nobody wants to get empty code-block.

### DIFF
--- a/sphinx/directives/code.py
+++ b/sphinx/directives/code.py
@@ -348,7 +348,7 @@ class LiteralIncludeReader:
                         return lines[:lineno + 1]
                     else:
                         if lineno == 0:
-                            return []
+                            pass  # end-before ignores first line
                         else:
                             return lines[:lineno]
             if inclusive is True:
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,9 @@ Dependencies
 Incompatible changes
 --------------------
 
+* #6742: ``end-before`` option of :rst:dir:`literalinclude` directive does not
+  match the first line of the code block.
+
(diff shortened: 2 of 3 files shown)


sphinx/directives/code.py
```diff
@@ -343,12 +343,13 @@ def end_filter(self, lines: List[str], location: Tuple[str, int] = None) -> List
 
         if end:
             for lineno, line in enumerate(lines):
+                print(lineno, line)
```
tk0miya: oops

Issue #6733, opened 2019-10-13
Additional vertical spacing in sphinxVerbatim environment (latex output)


skirpichev:
In the [diofant docs](https://github.com/diofant/diofant/tree/master/docs) I use xelatex engine and following preamble:
```latex
\setmainfont{DejaVu Serif}
\setsansfont{DejaVu Sans}
\setmonofont{DejaVu Sans Mono}
```

HTML output from the sphinx looks fine (see e.g. [this](https://diofant.readthedocs.io/en/v0.10.0/tutorial/intro.html), an example of ``solve()`` usage) after slight css adaptation
```css
.rst-content div[class^="highlight"] pre {
    line-height: normal;
    font-family: 'DejaVu Sans Mono', 'DejaVuSansMonoBook', 'DejaVuSansMonoBookFallback', monospace;
}
```

But in the PDF output there are additional vertical spacing (like one, removed by line-height directive in the HTML output).  See screenshot from the project issue https://github.com/diofant/diofant/issues/754.

tk0miya:
CSS does not effect to LaTeX output. You need to customize LaTeX output by LaTeX way (I don't know it). So I don't think this is a bug.

skirpichev:
> CSS does not effect to LaTeX output.

I know)

> You need to customize LaTeX output by LaTeX way (I don't know it). So I don't think this is a bug.

Hmm, apparently, verbatim output is broken.

BTW, it seems, that ``ℯ`` unicode symbol (SCRIPT SMALL E) in the code block was replaced by ``e`` in the latex output.  Is this not a bug?

tk0miya:
About line spacing, the default settings of LaTeX is not zero. I don't think it's not a bug. Please modify it via your custom macro. You can modify line spacing with modifying `\baselinestretch` or `\baselineskip`.
For example, following setting modifies line spacing of whole of documents:
```
latex_elements = {
    'preamble': r'\renewcommand{\baselinestretch}{0}'
}
```

>Hmm, apparently, verbatim output is broken.

Please let me know in detail. How did you write input text in reStructuredText? It is very helpful for us to fill our issue template. Thanks,

skirpichev:
> You can modify line spacing with modifying \baselinestretch

Thank you!  That way it looks much better now.

Yet, I see a problem with the mentioned above case when "SCRIPT SMALL E" was replaced with "e" in the latex output.  See screenshot (with adapted baselinestretch):
[...]

PR #6750, opened 2019-10-20
Fix #6738: Do not replace unicode characters by LaTeX macros on unicode supported LaTeX engine

### LINKED ISSUE #6738: LaTeX output: sub/superscripts and other Unicode glyphs get "TeX-escaped" even if engine is xelatex
Arguably, [TeX-escaping](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/texescape.py) when using Unicode aware xelatex engine should be essentially limited to the few characters with special TeX meaning such as `$`, `{`, `\` and a few others.

For the initial report where the too extensive TeX escaping causes problems see #6733. 

However some Unicode code-points such as `ℯ` seem to have limited font support (particularly monospace, it is lacking both from DejaVu Sans Mono and GNU FreeMono), and should perhaps be retained at least for the near future in the escape list. Indeed such missing glyphs do not raise errors but only LaTeX warnings and they may escape attention.

tk0miya:
### Feature or Bugfix
- Feature?
- Bugfix?

### Purpose
- refs: #6738

### DIFF
--- a/sphinx/util/texescape.py
+++ b/sphinx/util/texescape.py
@@ -9,7 +9,7 @@
 """
 
 import re
-from typing import Dict
+from typing import Callable, Dict
 
 tex_replacements = [
     # map TeX special chars
@@ -46,6 +46,14 @@ tex_replacements = [
     ('|', r'\textbar{}'),
     ('ℯ', r'e'),
     ('ⅈ', r'i'),
+    # Greek alphabet not escaped: pdflatex handles it via textalpha and inputenc
+    # OHM SIGN U+2126 is handled by LaTeX textcomp package
+]
+
+# A map Unicode characters to LaTeX representation
+# (for LaTeX engines which don't support unicode)
+unicode_tex_replacements = [
+    # superscript
     ('⁰', r'\(\sp{\text{0}}\)'),
     ('¹', r'\(\sp{\text{1}}\)'),
     ('²', r'\(\sp{\text{2}}\)'),
@@ -56,6 +64,7 @@ tex_replacements = [
     ('⁷', r'\(\sp{\text{7}}\)'),
     ('⁸', r'\(\sp{\text{8}}\)'),
     ('⁹', r'\(\sp{\text{9}}\)'),
+    # subscript
     ('₀', r'\(\sb{\text{0}}\)'),
     ('₁', r'\(\sb{\text{1}}\)'),
     ('₂', r'\(\sb{\text{2}}\)'),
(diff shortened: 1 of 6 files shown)


sphinx/util/texescape.py
```diff
@@ -46,6 +46,14 @@
     ('|', r'\textbar{}'),
     ('ℯ', r'e'),
     ('ⅈ', r'i'),
+    # Greek alphabet not escaped: pdflatex handles it via textalpha and inputenc
+    # OHM SIGN U+2126 is handled by LaTeX textcomp package
+]
+
+# A map Unicode characters to LaTeX representation
+# (for LaTeX engines which don't support unicode)
+unicode_tex_replacements = [
```
tk0miya, line 55: I don't know what characters should be kept as-is.

jfbu: I tested  the characters (one on each line below):

```
    ¶
    §
    €
    ∞
    ±
    →
    ‣
    ✓
    ✔
    ﻿
    ⎽
    –
    |
    ℯ
    ⅈ
```

with xelatex and GNU FreeFont. 

- Serif contains them all
- Sans and Mono are lacking some.

Here are the relevant warnings: (you need to see that with a font which has the glyphs...)

```
Missing character: There is no ✓ in font [FreeSans.otf]/OT:script=latn;language
=DFLT;mapping=tex-text;!
Missing character: There is no ✔ in font [FreeSans.otf]/OT:script=latn;language
=DFLT;mapping=tex-text;!
Missing character: There is no ﻿ in font [FreeSans.otf]/OT:script=latn;language
=DFLT;mapping=tex-text;!
Missing character: There is no ⎽ in font [FreeSans.otf]/OT:script=latn;language
=DFLT;mapping=tex-text;!
Missing character: There is no ⅈ in font [FreeSans.otf]/OT:script=latn;language
=DFLT;mapping=tex-text;!
 
Missing character: There is no ✓ in font [FreeMono.otf]/OT:script=latn;language
=DFLT;!
Missing character: There is no ✔ in font [FreeMono.otf]/OT:script=latn;language
=DFLT;!
Missing character: There is no ℯ in font [FreeMono.otf]/OT:script=latn;language
=DFLT;!
Missing character: There is no ⅈ in font [FreeMono.otf]/OT:script=latn;language
=DFLT;!
```

Sans is used in titles and Mono in code-blocks.

I thus think a daring but still cautious approach currently would be maintain the TeX escaping even with Unicode engines for the following characters: ``✓ , ✔ , ⎽ ,  ℯ , ⅈ`` and the ZWNBSP  (U+FEFF, also used as BYTE ORDER MARK).

But for ``¶,  §,  €,  ∞,  ±,  →,  ‣,  –,  |`` I don't see obvious reason why they should be TeX-escaped with Unicode engines.

Note 1: contrarily to web browsers, LaTeX has no easily usable concept of "fall-back" font when a glyph is missing from a font. It is up to user to add some `\DeclareUnicodeCharacter` or `\newunicodechar` to tell LaTeX that for this or that glyph, another font must be used. And making this fontfamily known to fontspec is also under user responsabillity.

Note 2: one may hesitate about escaping `ℯ` as it is availble in FreeSerif, but it is lacking from FreeMono and also from DejaVu Sans Mono. (as seen in #6733 original report). Probably same for `ⅈ`.

I am tempted to not escape them, but hope that user community will share advice on how to use `\newunicodechar` (I have not looked myself in detail and in particular how to use a fallback font for them only when needed, i.e. not in normal text which uses FreeSerif).

jfbu (approved):
LGTM with nits!

jfbu:
Merging as specific Unicode characters not needing replacements can be handled later. Thanks!

jfbu:
@tk0miya I have a problem with trying to move in particular the ∞ to `unicode_tex_replacements`. If I do that the ``test_markup.py::test_inline`` fails because ∞ is not TeX-escaped contrarily to expectation:

```
=================================== FAILURES ===================================
_ test_inline[verify-::\n\n @\u0393\\\u221e${}-None-\\begin{sphinxVerbatim}[commandchars=\\\\\\{\\}]\n@\u0393\\PYGZbs{}\\(\\infty\\)\\PYGZdl{}\\PYGZob{}\\PYGZcb{}\n\\end{sphinxVerbatim}] _
```

Other question, should

https://github.com/sphinx-doc/sphinx/blob/c2d64a9b0a7ef1fbd56c620d292e067ded462200/sphinx/ext/todo.py#L300-L308

use rather the `texescape.get_escape_func()` ?

tk0miya:
Thank you for merging. I'd missed your comment.

About `∞` is a bug. It seems highlighting does not refer new unicode table. I'll take a look.
And about todo extension, it is also a bug.

Issue #1493, opened 2015-01-03
Trying to follow "Tutorial: Writing a simple extension" leads to PicklingError


shimizukawa:
Steps to reproduce:

Attempt to follow

http://svn.python.org/projects/external/Sphinx-1.0.5/doc/ext/tutorial.rst

to implement todo or todolist (changing names to avoid any possible clashes), placing all code at the end of conf.py.  In your index.rst place your directive.  Run `make clean; make html`.

You will successfully read your codebase in preparation for documentation, but then fall foul of a PicklingError:  Can't pickle <class 'my_todo'>: attribute lookup **builtin**.my_todo failed.

The documentation is silent on this, and needs a lot more examples and definition of expected behaviour so that someone attempting to apply it can tell whether or not they are doing the correct thing.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1493
- Originally reported by: Anonymous
- Originally created at: 2014-06-18T11:57:21.728

shimizukawa:
_From Takayuki Shimizukawa on 2014-06-19 04:45:21+00:00_

A newer version of the documentation is here: http://sphinx-doc.org/extdev/index.html#dev-extensions

The reason of the exception `PicklingError` is that you've implemented extension code into conf.py. Generally, conf.py is not proper place to develop a extension because instances of class that is implemented in conf.py can't pickling (similar issue: #1424). You should implement extensions code in another .py file like todo.py as the documentation mentioned.

webknjaz:
FTR: this quirk is now documented in the tutorial: https://github.com/sphinx-doc/sphinx/pull/6754.

Issue #1424, opened 2015-01-03
PicklingError on environment when config option value is a callable


shimizukawa:
https://travis-ci.org/pypa/pip/jobs/20852384

I am trying to add extension to Sphinx that does [filtering on HTML](https://bitbucket.org/techtonik/wikify/) based on callable regex rules specified in `conf.py`. However, my build fails with PicklingError when I specify extension options.

```
...
looking for now-outdated files... none found
pickling environment...
Exception occurred:
  File "C:\Python33\lib\site-packages\sphinx-1.3a0dev_20140315-py3.3.egg\sphinx\environment.py", line 146, in topickle
    pickle.dump(self, picklefile, pickle.HIGHEST_PROTOCOL)
_pickle.PicklingError: Can't pickle <class 'function'>: attribute lookup builtins.function failed
The full traceback has been saved in c:\users\user\appdata\local\temp\sphinx-err-z8bc2u.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://bitbucket.org/birkenfeld/sphinx/issues/>. Thanks!
```

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1424
- Originally reported by: anatoly techtonik
- Originally created at: 2014-03-16T04:16:09.388

shimizukawa:
_From anatoly techtonik on 2014-03-16 03:17:51+00:00_

See intended usage scenario here - https://github.com/pypa/pip/pull/1647

shimizukawa:
_From anatoly techtonik on 2014-03-20 20:31:51+00:00_

The problem was caused by function in closure returned by function call in `conf.py`. I solved the issue by transforming the closure into object.

shimizukawa:
_From Takayuki Shimizukawa on 2014-03-20 22:41:59+00:00_

Yes, it's limitation of the sphinx. Because conf.py is loaded by a `exec` from `sphinx.config` module, objects/functions in the conf.py are not able to pickle.

shimizukawa:
_From anatoly techtonik on 2014-03-21 01:28:36+00:00_

That's not completely true. After converting closure to object, the error is gone.

webknjaz:
@techtonik do you recall what you meant by converting the closure to object?

techtonik:
[...]

PR #6754, opened 2019-10-22
Warn about dangers of inheriting nodes in conf.py

### LINKED ISSUE #6751: [docs][ext] Emphasise that one shouldn't put their node child classes in `conf.py`
**Describe the bug**

Docs don't mention that I absolutely must put the extension into a separate module file. As a result (knowing that things are supposed to work the same way) I was trying to have a custom directive declared in `conf.py`. This resulted in me wasting a lot of time and postponing completing the extension several times because when I just imported `nodes` "it worked" and it crashed when I inherited them.

Turns out, that there are at least these two issues reported: https://github.com/sphinx-doc/sphinx/issues/1493 / https://github.com/sphinx-doc/sphinx/issues/1424#issuecomment-68588465.
But they are hard to find for some reason and retrieving the solution from them is not trivial when you don't understand the cause of the problem.

**To Reproduce**

Create an extension in `conf.py` instead of a separate module as the tutorial suggests (but not insists).

**Expected behavior**

No pickling error. Or at least: **clear explanation and warning in docs about this problem**.
Also, I'm pretty sure that Sphinx can be changed to provide a more comprehensive background in the error message.

**Your project**

This was in the sibling `conf.py` originally: [scm_tag_titles_ext]

**Screenshots**

N/A

**Environment info**
- OS: Gentoo Linux
- Python version: 3.7.4
- Sphinx version: 2.2.0
- Sphinx extensions: 
  * sphinx.ext.autodoc
  * sphinx.ext.extlinks
  * sphinx.ext.intersphinx
  * [jaraco.packaging.sphinx](https://github.com/jaraco/jaraco.packaging/blob/master/jaraco/packaging/sphinx.py)
  * [sphinx_tabs.tabs](https://github.com/djungelorm/sphinx-tabs)
  * [scm_tag_titles_ext] (the extension I was developing)
 [e.g. sphinx.ext.autodoc, recommonmark]
- Extra tools: [tox](https://github.com/cherrypy/cheroot/blob/76cbe69ecfd2a3a5d8793a9b97c9c50eb95310e0/tox.ini#L33-L57) (just as a build runner)

**Additional context**

- https://github.com/sphinx-doc/sphinx/issues/1493
- https://github.com/sphinx-doc/sphinx/issues/1424#issuecomment-68588465

[scm_tag_titles_ext]: https://github.com/cherrypy/cheroot/blob/76cbe69ecfd2a3a5d8793a9b97c9c50eb95310e0/docs/scm_tag_titles_ext.py

webknjaz:
Closes #6751

Subject: Tell the readers of the TODO extension
         tutorial about the dangers of inheriting
         nodes in conf.py

### Feature or Bugfix

- Docs
- Bugfix

### Purpose
- Help users not to fall into an unobvious trap of
  putting their inherited `docutils` nodes into
  `conf.py` that causes a hard-to-debug
  `PickleError` popping up in the error log output

### Detail
- https://github.com/sphinx-doc/sphinx/issues/6751
- https://github.com/sphinx-doc/sphinx/issues/1493
- https://github.com/sphinx-doc/sphinx/issues/1424

### Relates

N/A

### DIFF
--- a/doc/development/tutorials/todo.rst
+++ b/doc/development/tutorials/todo.rst
@@ -107,6 +107,20 @@ is just a "general" node.
    <http://docutils.sourceforge.net/docs/ref/doctree.html>`__ and :ref:`Sphinx
    <nodes>`.
 
+.. DANGER::
+
+   It is important to know that while you can extend Sphinx without
+   leaving your ``conf.py``, if you declare an inherited node right
+   there, you'll hit an unobvious :py:class:`PickleError`. So if
+   something goes wrong, please make sure that you put inherited nodes
+   into a separate Python module.
+
+   For more details, see:
+
+   - https://github.com/sphinx-doc/sphinx/issues/6751
+   - https://github.com/sphinx-doc/sphinx/issues/1493
+   - https://github.com/sphinx-doc/sphinx/issues/1424
+
 .. rubric:: The directive classes
 
 A directive class is a class deriving usually from


webknjaz:
@tk0miya do you have anything to add here?

doc/development/tutorials/todo.rst
```diff
@@ -107,6 +107,20 @@ is just a "general" node.
    <http://docutils.sourceforge.net/docs/ref/doctree.html>`__ and :ref:`Sphinx
    <nodes>`.
 
+.. DANGER::
```
tk0miya, line 110: "danger" is a bit surprising to me. This tutorial let new developers know to create a new python file for extensions. What about using "attention" instead?

webknjaz: I'm not sure. To me, it feels more like a danger because it took me nearly half a year, on and off, to figure out how to complete what I wanted.
I understand that if the reader is following the tutorial blindly, they won't hit this issue, but it may still be very confusing and painful to debug to ones who don't do this for various reasons and has some initial knowledge resulting in a misleading perception of the problem.

stephenfin: Yeah, not a fan of danger either. Important or even just info would be more than appropriate

tk0miya:
Finally I merged this with replacing "danger" with "attention".
Thank you for your work!

webknjaz:
@tk0miya thanks!

PR #6758, opened 2019-10-22
Avoid collisions of path name

tkoyama010:
Hi sphinx project. I made PR which use sha512 to avoid using md5 and sha1. Thank you for reading.

Subject: Avoid using insecure hash functions

### Feature
- Feature

### Purpose
- Use of insecure MD2, MD4, MD5, or SHA1 hash function. Use SHA2 or SHA3 instead.

### Detail
- None

### Relates
- None

### DIFF
--- a/sphinx/builders/html.py
+++ b/sphinx/builders/html.py
@@ -13,7 +13,7 @@ import posixpath
 import re
 import sys
 import warnings
-from hashlib import md5
+from hashlib import sha512
 from os import path
 from typing import Any, Dict, IO, Iterable, Iterator, List, Set, Tuple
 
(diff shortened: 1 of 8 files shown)


tk0miya:
What did you mean "insecure"? I think these hash functions are mainly used to generate filename of temporary file. I agree conflicts of hash value is troublesome. But it does not insecure.

Note: SHA-512 generates 128 characters. It means these filename become too long (ex. `be688838ca8686e5c90689bf2ab585cef1137c999b48c70b92f67a5c34dc15697b5d11c982ed6d71be1e1e7f7b4e0733884aa97c3f7a339a8ed03577cf74be09.png`). I feel this is too long for URL...

tkoyama010:
@tk0miya Thank you for your comment. I used the expression that it is not safe in the sense that the paths collide. Can you give me some estimate information about image path collisions in Sphinx? I will withdraw this PR if there is no problem at present.

> I feel this is too long for URL...

I didn't notice that point. I agree that the path is too long.

tk0miya:
I don't have any estimate information. But I've never seen such reports for now.

tkoyama010:
Thanks. I close this PR.

PR #6765, opened 2019-10-24
Fixes #6759: validation of html static paths and extra paths no longe…

### LINKED ISSUE #6759: validate_html_static_path raises exception when static path and outdir are in different drives
**Describe the bug**
Hey sphinx team. Made a github account just for this, pretty excited to be potentially contributing to a project I use all the time 

I think the bug is caused by the implementation for #1464

Bug is pretty much outlined by the title. In the code for sphinx.builders.html, the validate_html_static_path title will raise an error if the static path is in a different drive than the outdir. I assume this will also happen in validate_html_extra_path. Specifically, the outdir is using a UNC directory (like \\network\some\path), but this happens with any two drives.

This occurs because these functions make a call to os.path.commonpath, which raises an error if the arguments are in different drives. Previously, sphinx had no problem with writing the output to a different drive.

**To Reproduce**
Easy enough to reproduce, no need for any real tinkering, since the error occurs before any real run happens.

```
$ git clone https://github.com/.../some_project
$ cd some_project
$ cd docs
$ make html BUILDDIR="D:\any\other\drive"
$ # we don't get past the init stage of the application
```

**Expected behavior**
In 2.1.2, neither of these functions exist, so the documents are successfully written in a different
drive than the source files. This is important for my project in particular, but seems like it's probably a common enough use case.

**Your project**
Can't submit the particular project. Hope this is okay, it's a project for work. I don't believe there's anything specific to the project that affects the bug, I reproduced it easily at home.

**Environment info**
- OS: Windows 10
- Python version: 3.7.4
- Sphinx version: 2.2.0
- Sphinx extensions: sphinx.ext.autodoc, sphinx.ext.doctest, sphinx.ext.mathjax, sphinx.ext.napoleon, sphxarg.ext
- Extra tools: sphinx_rtd_theme

**Additional context**
I've solved it temporarily by simply putting the calls to commonpath in try-except blocks. I think this is fine unless there is a specific reason this cross-drive writing shouldn't be allowed.

bizzfitch:
…r throws an error if the paths are in different directories

Subject: Reimplement behavior from 2.1.2

### Feature or Bugfix
- Bugfix

### Purpose
See my issue created. Hopefully I'm doing this right, this is my first pull request, kinda nervous
### Detail
In essence, the sys.path.commonpath function throws a Value Error if the paths are in different drives. Sphinx has previously had no issue writing to a different drive than the source, but this validation step throws an error. Now, if the drives are different, the error is caught. Since the purpose of the function was to check if the source and static directories was the same, it would evaluate to True if commonpath didn't raise an error.

### Relates
- https://github.com/sphinx-doc/sphinx/issues/6759

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,6 +25,7 @@ Other contributors, listed alphabetically, are:
 * Henrique Bastos -- SVG support for graphviz extension
 * Daniel Bültmann -- todo extension
 * Marco Buttu -- doctest extension (pyversion option)
+* Nathan Damon -- bugfix in validation of static paths in html builders
 * Etienne Desautels -- apidoc module
 * Michael Droettboom -- inheritance_diagram extension
 * Charles Duffy -- original graphviz extension
--- a/CHANGES
+++ b/CHANGES
@@ -90,6 +90,7 @@ Bugs fixed
 
 * #6641: LaTeX: Undefined control sequence ``\sphinxmaketitle``
 * #6710: LaTeX not well configured for Greek language as main language
+* #6759: validation of html static paths and extra paths no longer throws an error if the paths are in different directories
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -90,7 +90,8 @@ Bugs fixed
 
 * #6641: LaTeX: Undefined control sequence ``\sphinxmaketitle``
 * #6710: LaTeX not well configured for Greek language as main language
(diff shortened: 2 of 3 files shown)


tk0miya:
Now I'm preparing to merge this into 2.2.1 branch on #6774.

tk0miya:
I finished to merge this into 2.2.1 branch now.

Issue #3545, opened 2017-03-11
Add templating support to sphinx-apidoc


goerz:
The `sphinx-apidoc` utility has some shortcomings, including

* For modules with a lot of members, there is no overview of the members. The resulting page is very hard to navigate.
* If modules expose imported objects through `__all__`, `apidoc` creates stubs for both the exporting module and the module where the object was originally defined. This leads to duplicate references. It is generally hard to control which module members will show up in the documentation
* Apidoc produces potentially unwanted headings "Submodules", "Subpackages" and "Module contents". There is no option besides post-processing to disable these.

These and other concerns could be addressed by adding templating to the `sphinx-apidoc` utility. I've started to explore this option by forking the `apidoc` script to a tentative [better-apidoc](https://github.com/goerz/better-apidoc). It's a fairly small patch (I made sure to have the diff to the original [apidoc.py](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/apidoc.py) as small as possible) that adds a`-t/--templates` option to the script. If this option is present, Jinja-templates `module.rst` and `package.rst` are used to render the output. See the [`better-apidoc` README](https://github.com/goerz/better-apidoc#templating) for which variables are exposed to the template and example templates. Currently, this is designed (and may further evolve) to meet my needs to document the [QNET](https://github.com/mabuchilab/QNET) project.

I would welcome any comments and suggestions on this, especially regarding to which template variables would be useful. If there is a consensus that having template support would be useful, I would be happy to eventually prepare a pull-request to merge `better-apidoc` back into Sphinx (after some period of gestation).

I'm aware that [sphinx-autosummary](http://www.sphinx-doc.org/en/stable/ext/autosummary.html) already has support for templating, so there is some overlap here. However, as far as I can tell, autosummary will generate individual files for all module members. It doesn't seem to cover the use case of having one file per module or package combining both useful summaries and the full documentation of each member.

hanshoi:
For me this seems like a feature well worth implementing. Keep up the good job.

tk0miya:
+1; it would be nice if somebody sends us PR!

goerz:
Great, I'll work this into a pull request as soon as I have some time

EricG-Personal:
I would love to see this implemented as well.

goerz:
I will get back to this at some point, eventually..

goerz:
Well, basic templating was added in v2.2, and I've decided to finish the job with #6768.

[...]

PR #6768, opened 2019-10-24
Waiting for review: Add get_members function to apidoc templates (better_apidoc backport)

goerz:
This make a `get_members` function available to the `module.rst_t` and `package.rst_t` templates used by `apidoc`. This function allows to obtain the members of the current module or package in various forms and filtered by "type" (exception, class, function, data).

Building upon the recent addition of templating to `apidoc` in v2.2, this is a backport of https://github.com/goerz/better-apidoc

The original motivation for `better-apidoc` was that I find `apidoc`'s default template to be extremely hard to navigate for modules with a lot of members, see #3545. I strongly prefer to have a summary/table of contents at the beginning of each module's documentation that lists all the members of the module, and specifies which members are available through `__all__`

Examples of projects that use such a format (currently using `better-apidoc`, but possible with the features added in this PR):

* https://krotov.readthedocs.io/en/stable/API/krotov.html
* https://qnet.readthedocs.io/en/latest/API/qnet.html

The `get_members` function provided in the PR provides an extremely powerful mechanism for custom templates. The `better-apidoc` project has had a long gestation period, and has undergone several iterations over the last few years. Based on my experience of using it in a number of projects like the ones linked above, I am fully confident that the functionality provided in this PR can be considered stable. I will be available to maintain this part of `apidoc` going forward.

Within a template, `get_members` can be used e.g. as follows:

    {%- set members = get_members() %}
    {%- set all = get_members(in_list='__all__', include_imported=True) %}
    {%- set all_refs = get_members(in_list='__all__', include_imported=True, out_format='refs', known_refs='__known_refs__') -%}
    {%- set exceptions = get_members(typ='exception') -%}
    {%- set classes = get_members(typ='class') -%}
    {%- set functions = get_members(typ='function') -%}
    {%- set data = get_members(typ='data', in_list='__all__') -%}

The lists ``{{ members }}``, ``{{ all }}`` etc. are then available in the remainder of the template.

There is extensive further documentation and examples for the templating features in `apidoc`, both of those added in `v2.2` (which were largely undocumented) and for the `get_members` function added in this PR. This documentation is in the `apidoc` man page. The PR includes tests with full coverage of the new features.


goerz:
This is a backwards-compatible feature addition on `v2.2`. Thus, it should go in `v2.3`, and I've assumed this in the documentation (https://github.com/sphinx-doc/sphinx/pull/6768/commits/cd07183f3eba26bd3a7fdb4387b3be4778597c2b#diff-4b14242400560fc0f03fe3972cc0d25bR163)

However, it appears that currently no `2.3` branch exists on Github. Thus, I've made this PR against `master`. If I should rebase and force-push based on another branch, please let me know.

doc/man/sphinx-apidoc.rst
```diff
@@ -126,29 +126,253 @@ These options are used when :option:`--full` is specified:
...
+It uses the :data:`automodule_options` and :data:`show_headings` variables of
+``package.rst_t``, and additionally:
+
+.. data:: qualname
+
+The fully qualified name of the module.
+
+.. data:: basename
+
+An alias for :data:`qualname`.
```
goerz, line 251: This documentation matches what is currently implemented. Most likely, it's not what was intended, however. I would assume that `qualname` should be the fully qualified name of the module (with dots), whereas `basename` should only be the part after the last dot. I'll leave fixing this for another PR, though, after this is merged in.

doc/man/sphinx-apidoc.rst
```diff
@@ -126,29 +126,253 @@ These options are used when :option:`--full` is specified:
...
+The default ``module.rst_t`` is
+
+.. literalinclude:: ../../sphinx/templates/apidoc/module.rst_t
+   :language: jinja
+
+
+Accessing module members in templates
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.3
```
goerz, line 262: This assumes the PR will be merged and released in v2.3 (the next feature release). Otherwise, this will have to be adjusted.

goerz:
I still have to fix some mypy type errors before this is ready to be merged

goerz:
This is now ready for review/merging @tk0miya

doc/man/sphinx-apidoc.rst
goerz: The alternative would be for me to change the documentation of `basename` to "The part of the :data:`qualname` after the last dot". This would knowingly create my definition of a "bug" (a mismatch between documentation and behavior). If your preferred definition of "bug" is "mismatch between obvious intent and behavior", on the other hand, we can leave this for now.

goerz:
@tk0miya Is there any chance you could review and merge this in the near future? I'd really love to have this in the next release, so that I can start some new projects that don't depend on my `better-apidoc` hack.

goerz:
So it looks like 2.3 and 2.3.1 were released without this PR. Is there any interest in merging this? I'll have to modify the patch to fix the `..versionadded` to whatever release this will target.

TejasAvinashShetty:
eagerly waiting for this merge.

tk0miya:
OMG, I must apologize you not to review this. I'll do it within a week. Please wait a moment. Thanks,

doc/man/sphinx-apidoc.rst
tk0miya: I guess "qualname" came from PEP-3155's `__qualname__`. It is a relative path from its module. It was introduced mainly for nested classes. 
https://www.python.org/dev/peps/pep-3155/

tk0miya (commented):
I'm very sorry to late response. I'd missed your ping so long

I feel this is too complex. It seems the responsibility of this function is listing attributes, filtering, formatting and so on.  I can understand some people want to "hack" apidoc. And this would be helpful to them because this is very programatic. But it is hard to use and maintain to me.

I don't say this is rejected. I agreed current apidoc is poor. So improvement is still needed.

goerz:
I don't see how this can move forward then, unless you can think of any way to make it less complex while maintaining the functionality. So probably the only option for me is to maintain this as a permanent hard fork of apidoc (probably under a better name than the original `better-apidoc` that I used while drafting this).

gbroques:
@tk0miya Can you suggest how to make this less complex while maintaining the functionality?

Templates for apidoc is a very helpful feature.

EricG-Personal:
I would hate to see the idea dropped of adding the functionality of better-apidoc.

gbroques:
@goerz @tk0miya 

Any thoughts on reaching a middle-ground, or compromise here? 

### Proposal 1 (Ultra Simple)
How about simplifying this to _just_ include a `members` variable in the `module.rst_t` and `package.rst_t` templates that's the simple names of all qualifying members?

---
### Proposal 2 (Less Simple)
Alternatively, 

What about keeping a `get_members` function and removing some of the *less-critcial* options like `out_format`, `include_imported`, `in_list`, and / or `include_private`?

Formatting could be handled by users if they have access to `members` in the templates, and these changes could respect other options for including imported, private, or `__all__` members such as `imported-members`, `private-members`, and `ignore-module-all`.

---
### Proposal 3
**?**

doc/man/sphinx-apidoc.rst
```diff
@@ -126,29 +126,260 @@ These options are used when :option:`--full` is specified:
...
+      the given name. This is used only in conjunction with
+      ``out_format=refs``, to override automatically detected reference
+      location, or to provide references for object that cannot be located
+      automatically. The most common example for this is data members that are
+      exported by a package but are defined in a sub-module.
+
+   :return:
+      a list of strings, depending on `out_format`.
+
+      * If ``out_format='names'`` (default), the simple names of all qualifying
```
gbroques, line 309: If we remove the `out_format` option from this entirely, then that simplifies the changes greatly.

> I feel this is too complex. It seems the responsibility of this function is listing attributes, filtering, **formatting** and so on

That addresses removing the formatting concern.

doc/man/sphinx-apidoc.rst
```diff
@@ -126,29 +126,260 @@ These options are used when :option:`--full` is specified:
...
+
+.. versionadded:: 2.3
+   Advanced templating with ``get_members``
+
+Both the ``package.rst_t`` and ``module.rst_t`` templates have access to a
+:func:`get_members` function that allows to extract the members of the package or
+module being rendered:
+
+
+.. py:function:: get_members(fullname, typ=None, include_imported=False, out_format='names', in_list=None, include_private=*private*, known_refs=None)
```
gbroques, line 270: @goerz  Is changing `get_members` from a function to a `members` variable that's the simple names of all qualifying members too much a radical simplification?

@tkoyama010 Do you think you would approve and merge such changes?

AA-Turner:
Hi Michael (@goerz) -- is this PR no longer required? I missed the reason for closing, sorry.

Thanks,
Adam

goerz:
It’s not going to be merged, and I have switched from Python to Julia for all my work, so why keep it open?

Issue #6000, opened 2019-02-02
have backslash also be an inline literal word wrap break character in LaTeX


kenyon:
**Is your feature request related to a problem? Please describe.**
When long Windows filename paths (directories separated by backslash (`\`) characters) are used with the `:file:` role, they are not word wrapped, they just keep going off the side of the page.

**Describe the solution you'd like**
The backslash character should also be a word wrap break character.

**Describe alternatives you've considered**
Tried to use `\renewcommand*\sphinxbreaksafteractivelist {\do\.\do\,\do\;\do\?\do\!\do\/\do\\}` in my preamble to add backslash, but get this error when building:

    ! You can't use `macro parameter character #' in horizontal mode.

**Additional context**
#3116, #2943, #3110

jfbu:
Thanks for reporting. When used as is ``\`` in a ```` ``\C\aaa`` ````, or escaped as ``\\`` in a ``:file:`\\C\\aaa` ``, the backslash ends us as `\textbackslash{}` in LaTeX code.

So the trick would be to hack into  `\sphinxupquote` macro to modify behaviour of `\textbackslash{}` within it to allow line breaks. (I will provide a way later, unfortunately does not have it seems a clear hook entry point).

On the other hand inside a code-block, the backslash is escaped by Pygmentize library into `\PYGbs` macro. But your question made me realize that Sphinx does nothing special regarding linebreaks for this: it is omitted from both `\sphinxbreaksbeforelist` and `\sphinxbreaksafterlist` macros. Perhaps because it was not known what was best: to break before it or break after it at end of line, or perhaps simply an oversight. I will open an issue on this.

But your issue is another one, it regards `\textbackslash{}` behaviour.

jfbu:
As a temporary workaround, use either this

```
latex_elements = {
    'preamble': r'''
\let\origbs\textbackslash
\newcommand\allowbreaksbeforebackslashinliterals{%
  \def\textbackslash{\discretionary{}{\origbs}{\origbs}}% breaks before \
}%
\makeatletter
\g@addto@macro\sphinx@literal@nolig@list{\allowbreaksbeforebackslashinliterals}
\makeatother
''',
}
```

or that

```
latex_elements = {
    'preamble': r'''
\let\origbs\textbackslash
\newcommand\allowbreaksafterbackslashinliterals {%
  \def\textbackslash{\discretionary{\origbs}{}{\origbs}}% breaks after \
}%
\makeatletter
[...]

Issue #6001, opened 2019-02-02
LaTeX does not wrap long code lines at backslash character


jfbu:
There is no line-wrapping at ``\`` in code-blocks.

```rest
::

   \\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\
```
produces this via `make latexpdf`

![capture d ecran 2019-02-02 a 15 14 46](https://user-images.githubusercontent.com/2589111/52165168-71d6e180-26fd-11e9-835b-597682d3e505.png)

**Additional context**

- #6000

Not sure if qualifies at bug, so currently labelling it as feature request.

tk0miya:
I think this must be "a bug", but minor. So it is better to mark this as "prio:low" and "bug".

jfbu:
Fixed at https://github.com/sphinx-doc/sphinx/commit/a7f53a7957968e9f7daa3678ea695213af3e296e

PR #6799, opened 2019-11-06
LaTeX: Allow linebreaks at \ (in literals, code-blocks, parsed-literal)

jfbu:
Default configuration is to allow the linebreak after the backslash.
It is possible at user side via 'preamble' key and some LaTeX code
to allow the linebreak before the backslash, but each of inline-literal,
code-block, and parsed-literal has its more or less cumbersome way for
that.

Closes: #6000, #6001

This commit handles only the \ character, which had been left aside
by accident so far. Inline-literals could break at more characters
but this is not yet done in this commit. In this context, in future
it could be useful to modify sphinx.util.texescape to use mark-up
such as \sphinxtextbackslash rather than naked \textbackslash and
similarly for some other characters. This could then be used to
allow linebreaks even in regular text paragraphs.

Both
- Feature
- Bugfix

### Relates
- #6000, #6001

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -28,6 +28,8 @@ Features added
 * #6762: latex: Allow to load additonal LaTeX packages via ``extrapackages`` key
   of :confval:`latex_elements`
 * #1331: Add new config variable: :confval:`user_agent`
+* #6000: LaTeX: have backslash also be an inline literal word wrap break
+  character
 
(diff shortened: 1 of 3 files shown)


tk0miya (approved):
LGTM

jfbu:
Thanks for reviewing!

PR #6801, opened 2019-11-07
Close #2546: apidoc: .so file support

### LINKED ISSUE #2546: sphinx-apidoc does not find ".so" files
`recurse_tree()` only looks for files with extension `.py` and `.pyx`. Compiled python modules with the extension `.so` work perfectly fine with sphinx automodule but are not found by by `recurse_tree()`. Simply adding `.so` to the search path would be problematic because it would not properly handle situation where there is a directory and a `.so` file with the same name. 

I note that refactoring this function to use `pkgutil.walk_packages()` would be the most correct solution, as it searches the directory using the same code python actually uses to load modules. However, it would exclude the possibility of having `followlinks=false` even though I am not sure why someone would want this option to be false.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #2546

### DIFF
--- a/sphinx/ext/apidoc.py
+++ b/sphinx/ext/apidoc.py
@@ -21,6 +21,7 @@ import os
 import sys
 import warnings
 from fnmatch import fnmatch
+from importlib.machinery import EXTENSION_SUFFIXES
 from os import path
 from typing import Any, List, Tuple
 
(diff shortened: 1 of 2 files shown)


sphinx/ext/apidoc.py
```diff
@@ -45,7 +46,7 @@
     ]
 
 INITPY = '__init__.py'
-PY_SUFFIXES = {'.py', '.pyx'}
+PY_SUFFIXES = ('.py', '.pyx', sysconfig.get_config_var('EXT_SUFFIX'))
```
tk0miya: It seems `importlib.machinery.EXTENSION_SUFFIXES` is much better than this.
https://docs.python.org/3/library/importlib.html#importlib.machinery.EXTENSION_SUFFIXES

PR #6812, opened 2019-11-11
Give a warning when extensions are explicitly not parallel safe

astrofrog:
### Feature or Bugfix
I would personally treat this as a bug fix since the previous behavior was confusing.

### Purpose
Currently, if building a set of docs with a number of extensions in parallel, if one of the extensions explicitly declares itself to be unsafe, the build falls back to a serial build silently, with no way of knowing what extension caused this. Rather than have to dig into the source code for each extension, it's much simpler to just warn the user if the build is falling back to a serial build and what extension caused this. I think it makes sense to include this since if the user is opting in to a parallel build, anything that prevents that from happening should be explicitly stated.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -1190,28 +1190,33 @@ class Sphinx:
 
         ``typ`` is a type of processing; ``'read'`` or ``'write'``.
         """
+
         if typ == 'read':
             attrname = 'parallel_read_safe'
-            message = __("the %s extension does not declare if it is safe "
-                         "for parallel reading, assuming it isn't - please "
-                         "ask the extension author to check and make it "
-                         "explicit")
+            message_none = __("the %s extension does not declare if it is safe "
+                              "for parallel reading, assuming it isn't - please "
+                              "ask the extension author to check and make it "
+                              "explicit")
+            message_false = "the %s extension is not safe for parallel reading"
         elif typ == 'write':
             attrname = 'parallel_write_safe'
-            message = __("the %s extension does not declare if it is safe "
-                         "for parallel writing, assuming it isn't - please "
-                         "ask the extension author to check and make it "
-                         "explicit")
+            message_none = __("the %s extension does not declare if it is safe "
+                              "for parallel writing, assuming it isn't - please "
+                              "ask the extension author to check and make it "
+                              "explicit")
+            message_false = "the %s extension is not safe for parallel writing"
         else:
             raise ValueError('parallel type %s is not supported' % typ)
 
         for ext in self.extensions.values():
             allowed = getattr(ext, attrname, None)
             if allowed is None:
-                logger.warning(message, ext.name)
+                logger.warning(message_none, ext.name)
                 logger.warning(__('doing serial %s'), typ)
                 return False
             elif not allowed:
+                logger.warning(message_false, ext.name)
+                logger.warning(__('doing serial %s'), typ)
                 return False
 
         return True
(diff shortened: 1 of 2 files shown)


sphinx/application.py
```diff
@@ -1126,28 +1126,33 @@ def is_parallel_allowed(self, typ):
 
         ``typ`` is a type of processing; ``'read'`` or ``'write'``.
         """
+
```
tk0miya: Could you remove this empty line please?

sphinx/application.py
```diff
@@ -1126,28 +1126,33 @@ def is_parallel_allowed(self, typ):
...
             attrname = 'parallel_read_safe'
-            message = __("the %s extension does not declare if it is safe "
-                         "for parallel reading, assuming it isn't - please "
-                         "ask the extension author to check and make it "
-                         "explicit")
+            message_none = __("the %s extension does not declare if it is safe "
+                              "for parallel reading, assuming it isn't - please "
+                              "ask the extension author to check and make it "
+                              "explicit")
+            message_false = "the %s extension is not safe for parallel reading"
```
tk0miya: Please wrap with `_()` function like above. It is needed for i18n.

sphinx/application.py
```diff
@@ -1126,28 +1126,33 @@ def is_parallel_allowed(self, typ):
...
+                              "explicit")
+            message_false = "the %s extension is not safe for parallel writing"
         else:
             raise ValueError('parallel type %s is not supported' % typ)
 
         for ext in self.extensions.values():
             allowed = getattr(ext, attrname, None)
             if allowed is None:
-                logger.warning(message, ext.name)
+                logger.warning(message_none, ext.name)
```
tk0miya: I feel `message_none` and `message_false` not describe its purpose. I think `message_not_declared` and `message_not_allowed` are better. What do you think?

tests/test_application.py
```diff
@@ -98,6 +98,8 @@ def test_add_is_parallel_allowed(app, status, warning):
 
     app.setup_extension('read_serial')
     assert app.is_parallel_allowed('read') is False
+    assert ("the read_serial extension is not safe for parallel reading") in warning.getvalue()
```
tk0miya: Coudl you remove parenthesis please? They are used only for folding long text to multiple lines.

tk0miya (commented):
LGTM with nits.

In addition, could you rebase this into 2.0 branch please? Then I'll release this as 2.3.0.

sphinx/application.py
astrofrog: Done

sphinx/application.py
astrofrog: Done

sphinx/application.py
astrofrog: Fixed - I settled on ``message_not_declared`` and ``message_not_safe`` - is that ok?

tests/test_application.py
astrofrog: Done

astrofrog:
@tk0miya - thanks for the review! I've addressed all your comments and changed this PR to be against 2.0.

sphinx/application.py
tk0miya: Good point

tk0miya (approved):
Great! I'll merge this after CI passed.

tk0miya:
Merged.

astrofrog:
Thanks!

PR #6814, opened 2019-11-12
Shorten message on search page

timhoffm:
Subject: Shorten message on search page

### Purpose
The text was a bit wordy. It's obvious that the search page can be used to search. Also, we don't need to tell people how to use a search field.

The only relevant content of the paragraph is that matches contain all words. That can be described in one concise sentence, which is more likely to be read than a whole paragraph.

Note: This will need translations updates as a follow-up. Should I update the .po files within this PR? It seems these file are not up-to-date anyway and there would be other changes included.

### DIFF
--- a/sphinx/themes/basic/search.html
+++ b/sphinx/themes/basic/search.html
@@ -27,10 +27,8 @@
   </p>
   </div>
   <p>
-    {% trans %}From here you can search these documents. Enter your search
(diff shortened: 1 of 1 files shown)


tk0miya (approved):
+0

tk0miya:

PR #6825, opened 2019-11-15
Fetch `intersphinx` inventories concurrently

### LINKED ISSUE #6818: Intersphinx could fetch remote inventories concurrently
Hypothesis has optional integrations with many other libraries, and [our `intersphinx` config](https://github.com/HypothesisWorks/hypothesis/blob/583db593c5bf2d4d2985aa9d31decf785ad364e8/hypothesis-python/docs/conf.py#L66-L73) currently links to six other sets of documentation.  This works beautifully, and we all really appreciate Sphinx!  ❤️

However, when the cache is cold or doing a clean build, fetching all of these inventories can take considerably longer than the rest of the build.  I propose to use a `concurrent.futures` thread pool to run [the inner loop in `load_mappings()`](https://github.com/sphinx-doc/sphinx/blob/e8925e78ca3be64a94c061e71a1da7d161640fa1/sphinx/ext/intersphinx.py#L196-L216) when there is more than one inventory that needs to be fetched.

This could give slightly better or slightly worse performance for all-local inventories, but will be *much* faster for remote inventories.  I am happy to open a PR with this change.

Zac-HD:
This patch modifies the Intersphinx extension to fetch each inventory - an operation that often means downloading it from the web - in parallel using a `concurrent.futures.ThreadPoolExecutor()`.

Closes #6818, and makes a clean build of Hypothesis' docs build roughly 2x faster :smile:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,7 +36,7 @@ Other contributors, listed alphabetically, are:
 * Hernan Grecco -- search improvements
 * Horst Gutmann -- internationalization support
 * Martin Hans -- autodoc improvements
-* Zac Hatfield-Dodds -- doctest reporting improvements
+* Zac Hatfield-Dodds -- doctest reporting improvements, intersphinx performance
 * Doug Hellmann -- graphviz improvements
 * Tim Hoffmann -- theme improvements
 * Antti Kaihola -- doctest extension (skipif option)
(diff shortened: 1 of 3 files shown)


tk0miya (approved):
Good work!

tk0miya:

PR #6827, opened 2019-11-15
Allow regex patterns for start/end patterns

### LINKED ISSUE #6826: Allow regexes for start/end specifiers on literalinclude
When specifying start and end search patterns for `literalinclude`, it’s often impossible to fully specify one or both of the patterns. Common issues I run into are the end brackets for C structure declarations, when other structures or unions are embedded in the main `struct`.

I’d like to be able to use regex patterns with the four start/end options.

I have a patch implementing this.

skitt:
This adds a feature to handle regex patterns in start and end options for `literalinclude`. Adding a `regex` flag will cause all `start-after`, `start-at`, `end-before`, and/or `end-at` options to be handled as regex patterns instead of fixed strings.

Closes: #6826

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -51,6 +51,7 @@ Features added
 * #6707: C++, support bit-fields.
 * #267: html: Eliminate prompt characters of doctest block from copyable text
 * #6729: html theme: agogo theme now supports ``rightsidebar`` option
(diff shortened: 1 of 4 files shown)


skitt:
Would it be appropriate to backport this to 2.2.x? (Assuming it’s appropriate for master...)

PR #6833, opened 2019-11-16
Fix #6738: Use get_encode_func in todo extension

### LINKED ISSUE #6738: LaTeX output: sub/superscripts and other Unicode glyphs get "TeX-escaped" even if engine is xelatex

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- see https://github.com/sphinx-doc/sphinx/pull/6750#issuecomment-554515205

### DIFF
--- a/sphinx/ext/todo.py
+++ b/sphinx/ext/todo.py
@@ -30,7 +30,7 @@ from sphinx.locale import _, __
 from sphinx.util import logging
 from sphinx.util.docutils import SphinxDirective
 from sphinx.util.nodes import make_refnode
-from sphinx.util.texescape import tex_escape_map
+from sphinx.util.texescape import get_escape_func
 from sphinx.writers.html import HTMLTranslator
 from sphinx.writers.latex import LaTeXTranslator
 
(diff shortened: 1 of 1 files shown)


jfbu (approved):
LGTM, how to inform third-party extensions about this? Fortunately they will work as before but simply miss the better handling of Unicode for LaTeX output, afaict

tk0miya:
Good point. We need to let them know to use new utility function in CHANGES file.

PR #6834, opened 2019-11-16
Reconsider public APIs of sphinx.util.texescape

tk0miya:
### Feature or Bugfix
- Refactoring
  - Deprecate `sphinx.util.texescape.tex_escape_map`
  - Deprecate `sphinx.util.texescape.tex_hl_escape_map_new`
  - Make `escape()` and `hlescape()` public APIs of `sphinx.util.texescape` instead of `get_escape_func()` and `get_hlescape_func()`.

### DIFF
--- a/sphinx/util/texescape.py
+++ b/sphinx/util/texescape.py
@@ -9,7 +9,7 @@
 """
 
 import re
-from typing import Callable, Dict
+from typing import Dict
 
 from sphinx.deprecation import RemovedInSphinx40Warning, deprecated_alias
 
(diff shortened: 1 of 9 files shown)


jfbu (commented):
I think

jfbu (commented):
sorry for keyboard mistake sending empty comment

the sphinx.highlightting still has usage of sphinx.util.texescape.tex_hl_escape_map_new

tk0miya:
I found an usage of `sphinx.util.texescape:escape()` directly. I'll reconsider the public API of texescape module again.

sphinx/util/texescape.py
```diff
@@ -78,47 +81,38 @@
...
-tex_hl_escape_map_new = {}  # type: Dict[int, str]
-tex_hl_escape_map_new_without_unicode = {}  # type: Dict[int, str]
+tex_replace_map = {}  # type: Dict[int, str]
 
+_tex_escape_map = {}  # type: Dict[int, str]
+_tex_escape_map_without_unicode = {}  # type: Dict[int, str]
+_tex_hlescape_map = {}  # type: Dict[int, str]
+_tex_hlescape_map_without_unicode = {}  # type: Dict[int, str]
 
-def get_escape_func(latex_engine: str) -> Callable[[str], str]:
```
tk0miya, line 88: I determined to drop `get_escape_func()` and add a feature to switch escaping map to `escape()` function itself. I think enough simple and allows to keep compatibility for modules already using `escape()` function.

jfbu: +1 I will review a bit later

jfbu (commented):
LGTM with nits! I merge 2.0 (at 948ac3a1f07, which thus incorporates #6835) into this for testing, and a `€` in a literal block becomes `@texteuro[]` in the contents of `sphinxVerbatim` environment. I wonder why but I have not tried to understand yet. Inline literals are ok. Possibly some interaction with Pygments.

```
::
   €
```

with of course ``latex_engine = 'xelatex'``.

jfbu (approved):
LGTM! the problem with € in code block is already there at current 2.0 at 948ac3a1f07e9 (which incorporates #6835), hence it needs separate resolution already.

jfbu:
The  € problem is a consequence of 0103ff2a2b612994b077d9d1e37692ba896d86dc, see https://github.com/sphinx-doc/sphinx/blob/948ac3a1f07e954e51938678334a984ca1216145/sphinx/writers/latex.py#L2167-L2168

tk0miya:
Thank you for reviewing. merged.

PR #6835, opened 2019-11-16
LaTeX: escape fewer Unicode chars if engine is lualatex or xelatex

jfbu:
### Relates
- #6738, #6750, #6832 

See https://github.com/sphinx-doc/sphinx/pull/6750#discussion_r337611209

Note: I don't know why `|` gets TeX-escaped. I have left it for backwards compatibility but perhaps it can be removed (for all engines; I tested briefly no problem with Japanese if not TeX-escaped).

edit: the `|` needs some kind of escaping if it ends up in `\index{...}` LaTeX macro. See https://github.com/sphinx-doc/sphinx/pull/6835#discussion_r347154883

### DIFF
--- a/sphinx/util/texescape.py
+++ b/sphinx/util/texescape.py
@@ -20,29 +20,28 @@ tex_replacements = [
     ('_', r'\_'),
     ('{', r'\{'),
     ('}', r'\}'),
-    ('[', r'{[}'),
-    (']', r'{]}'),
-    ('`', r'{}`'),
     ('\\', r'\textbackslash{}'),
     ('~', r'\textasciitilde{}'),
+    ('^', r'\textasciicircum{}'),
+    # map chars to avoid mis-interpretation in LaTeX
+    ('[', r'{[}'),
+    (']', r'{]}'),
+    # map chars to avoid TeX ligatures
+    # 1. ' - and , not here for some legacy reason
+    # 2. no effect with lualatex (done otherwise: #5790)
+    ('`', r'{}`'),
     ('<', r'\textless{}'),
     ('>', r'\textgreater{}'),
-    ('^', r'\textasciicircum{}'),
+    # map char for some unknown reason.  TODO: remove this?
+    ('|', r'\textbar{}'),
     # map special Unicode characters to TeX commands
-    ('¶', r'\P{}'),
-    ('§', r'\S{}'),
-    ('€', r'\texteuro{}'),
-    ('±', r'\(\pm\)'),
-    ('→', r'\(\rightarrow\)'),
-    ('‣', r'\(\rightarrow\)'),
     ('✓', r'\(\checkmark\)'),
     ('✔', r'\(\pmb{\checkmark}\)'),
     # used to separate -- in options
     ('﻿', r'{}'),
     # map some special Unicode characters to similar ASCII ones
+    # (even for Unicode LaTeX as may not be supported by OpenType font)
     ('⎽', r'\_'),
-    ('–', r'\textendash{}'),
-    ('|', r'\textbar{}'),
     ('ℯ', r'e'),
     ('ⅈ', r'i'),
     # Greek alphabet not escaped: pdflatex handles it via textalpha and inputenc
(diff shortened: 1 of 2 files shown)


sphinx/util/texescape.py
```diff
@@ -20,29 +20,28 @@
...
+    # map chars to avoid TeX ligatures
+    # 1. ' - and , not here for some legacy reason
+    # 2. no effect with lualatex (done otherwise: #5790)
     ('`', r'{}`'),
-    ('\\', r'\textbackslash{}'),
-    ('~', r'\textasciitilde{}'),
     ('<', r'\textless{}'),
     ('>', r'\textgreater{}'),
-    ('^', r'\textasciicircum{}'),
+    # map char for some unknown reason.  TODO: remove this?
```
tk0miya, line 35: This was added in 9a2bf336f7. Unfortunately there are no mentions about reasons.

tk0miya (approved):
LGTM!

sphinx/util/texescape.py
tk0miya: +0 for removal.

jfbu: I think `|` is used as LaTeX shorthand in some languages, that could be the reason. Escaping avoids mis-interpretation. I will investigate later, but for time being safer to keep as is!

jfbu:
Thanks as always!

sphinx/util/texescape.py
jfbu: I  have found only one reason so far: the `|` plays special role in LaTeX `\index` macro hence must be escaped (and if that is only reason maybe that could be done by ``visit_index()`` escaping procedure in LaTeX writer rather than globally in document).

Currently ``.. index:: a|b`` produces ``\index{a\textbar{}b@\spxentry{a\textbar{}b}}`` which looks a bit weird but does print readable index entry and raises no error. The normal LaTeX escaping should be ``\index{a"|b@\spxentry{a"|b}}`` but it is broken by hyperref package, even ``\index{a"|b}`` is broken. I need to investigate that (has nothing to do with Sphinx LaTeX). I know that makeindex program has an *encap* parameter whose default is `|`, if we remove that then possibly the `|` can be used non-escaped (but attention it is used with this meaning in `visit_index()` of LaTeX writer) but I need to check xindy program, and also check how this would interact with hyperref, and besides it may break extensions of Sphinx...

I checked `|` is in fact not used as shorthand, as far as I can tell.


Appendix: info on shorthands,  I know it is now off topic but as I gathered it I prefer to keep it as memo :)

- Languages with no shorthands Croatian, English (any variety), Indonesian, Hebrew, Interlingua, Irish, Lower Sorbian, Malaysian, North Sami, Romanian, Scottish, Welsh

- Languages with only " as defined shorthand character Albanian, Bulgarian, Danish, Dutch, Finnish, German (old and new orthography, also Austrian), Icelandic, Italian, Norwegian, Polish, Portuguese (also Brazilian), Russian, Serbian (with Latin script), Slovene, Swedish, Ukrainian, Upper Sorbian

- Basque " ' ~
- Breton : ; ? !
- Catalan " ' `
- Czech " -
- Esperanto ^
- Estonian " ~
- French (all varieties) : ; ? !
- Galician " . ' ~ < >
- Greek ~
- Hungarian `
- Kurmanji ^
- Latin " ^ =
- Slovak " ^ ' -
- Spanish " . < > '
- Turkish : ! =

- In addition, the babel core declares ~ as a one-char shorthand which is let, like the standard ~, to a non breaking space.

(all the above copied straight from latex babel package documentation)

Thus, the `|` is not a priori used as shorthand. We have special measures to always deactivate ``"`` and ``=``. And some shorthands above need special option. For example with Spanish the `'` is a shorthand only with an extra "activeacute" option for latex babel package (if activated, it transforms `'a` into `á` for example). Some other shorthands such as those of French are less problematic (from Sphinx point of view) because they don't combine with other characters (they add extra space only).

jfbu: upstream ticket https://github.com/latex3/hyperref/issues/109

PR #6840, opened 2019-11-17
Fix link to pygments lexer in docs

sciencewhiz:
### Feature or Bugfix

- Bugfix

### Purpose
Pygments website changed hosts and the lexer link with trailing slash gives a 404. Removing trailing slash works.


tk0miya:
Thanks!

PR #6841, opened 2019-11-17
Fix #4186: LaTeX: Support upLaTeX as a new latex_engine (experimental)

### LINKED ISSUE #4186: Feature Request: support ``upLaTeX`` for PDF builds of Japanese projects
Many projects will use Unicode characters. Sphinx uses ``pLaTeX`` for Japanese documents but ``pLaTeX`` is not Unicode aware. 

It appears [upLaTeX](http://www.t-lab.opal.ne.jp/tex/uptex_en.html) adds Unicode support and can help handling problems with stray Unicode characters: in particular using `\usepackage[utf8]{inputenc}` and `babel` is possible  with `upLaTeX`. From the little I understand it is easier with ``upLaTeX`` to have mixed languages documents, for example with parts in English and others in Japanese.

Currently one can induce Sphinx into using ``upLaTeX`` via this ``latexmkjarc`` file in the repertory of the project

```
$latex = 'uplatex ' . $ENV{'LATEXOPTS'} . ' %O %S';
$dvipdf = 'dvipdfmx %O -o %D %S';
$makeindex = 'rm -f %D; upmendex -f -d %B.dic -s python.ist %S || echo "upmendex exited with error code $? (ignoring)" && : >> %D';
add_cus_dep( "glo", "gls", 0, "makeglo" );
sub makeglo {
 return system( "upmendex -s gglo.ist -o '$_[0].gls' '$_[0].glo'" );
}
```

and an ``latex_additional_files = ['latexmkjarc']`` line in ``conf.py``. I am not sure the options to  ``uplatex`` and ``upmendex`` above are optimal. I just took the current file which uses ``platex`` and ``mendex`` and dropped options not existing for  ``uplatex/upmendex``.

But we also need extra ``uplatex`` option in document class in case of ``'manual'`` type (which uses ``jsbook.cls``). 

This option does not work for ``'howto'`` documents and one must use ``ujreport.cls`` as replacement for ``jreport.cls``. Currently this setting appears to require some monkey-patch by user of sphinx.builders.latex's ``default_latex_docclass``. (see https://github.com/sphinx-doc/sphinx/issues/4136#issuecomment-338802296)

Perhaps ``hyperref`` options for the encoding of PDF strings (particularly, bookmarks) is perhaps not good with ``upLaTeX``. Currently ``sphinx.sty`` contains this logic:

```
% to make pdf with correct encoded bookmarks in Japanese
% this should precede the hyperref package
\ifx\kanjiskip\undefined
% for non-Japanese: make sure bookmarks are ok also with lualatex
  \PassOptionsToPackage{pdfencoding=unicode}{hyperref}
\else
  \RequirePackage{atbegshi}
  \ifx\ucs\undefined
    \ifnum 42146=\euc"A4A2
      \AtBeginShipoutFirst{\special{pdf:tounicode EUC-UCS2}}
    \else
      \AtBeginShipoutFirst{\special{pdf:tounicode 90ms-RKSJ-UCS2}}
    \fi
  \else
    \AtBeginShipoutFirst{\special{pdf:tounicode UTF8-UCS2}}
  \fi
\fi
```

I don't know if this is the good thing to do for `upLaTeX`. (I don't even know if it is so good for `pLaTeX`,  for example I noticed there are problems with EN DASH).

Adding ``upLaTeX`` support could be good in particular for localization to Japanese of documents when localization is only partial. The CPython docs is an example.

### Relates

the discussion at #4136

### Environment info

- Sphinx version: 1.6.5

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #4186 
- I'm not familiar with LaTeX and Japanese LaTeX. So I don't know this is appropriate settings. But it seems working fine to me.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -31,6 +31,7 @@ Features added
 * #1331: Add new config variable: :confval:`user_agent`
 * #6000: LaTeX: have backslash also be an inline literal word wrap break
   character
+* #4186: LaTeX: Support upLaTeX as a new :confval:`latex_engine` (experimental)
 * #6812: Improve a warning message when extensions are not parallel safe
 * #6818: Improve Intersphinx performance for multiple remote inventories.
 * #2546: apidoc: .so file support
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1836,6 +1836,7 @@ These options influence LaTeX output.
    * ``'xelatex'`` -- XeLaTeX
    * ``'lualatex'`` -- LuaLaTeX
    * ``'platex'`` -- pLaTeX (default if :confval:`language` is ``'ja'``)
+   * ``'uplatex'`` -- upLaTeX (experimental)
 
    ``'pdflatex'``\ 's support for Unicode characters is limited.
 
@@ -1860,6 +1861,10 @@ These options influence LaTeX output.
    .. versionchanged:: 2.2.1
 
       Use ``xelatex`` by default for Greek documents.
+
+   .. versionchanged:: 2.3
+
+      Add ``uplatex`` support.
 
    Contrarily to :ref:`MathJaX math rendering in HTML output <math-support>`,
    LaTeX requires some extra configuration to support Unicode literals in
--- a/sphinx/builders/latex/__init__.py
+++ b/sphinx/builders/latex/__init__.py
@@ -427,8 +427,12 @@ def default_latex_engine(config: Config) -> str:
 def default_latex_docclass(config: Config) -> Dict[str, str]:
     """ Better default latex_docclass settings for specific languages. """
     if config.language == 'ja':
-        return {'manual': 'jsbook',
-                'howto': 'jreport'}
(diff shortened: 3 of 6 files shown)


jfbu (commented):
Unfortunately I know nothing about uplatex and Japanese language... what about using [upmendex](https://ctan.org/pkg/upmendex) as indexing tool?

but +1 anyway as best way to test is to at least add this new possibility for `latex_engine`...

tk0miya:
Unfortunately, ubuntu LTS does not contain upmendex (since TeXLive 2016). So it is better to switch to it in another time.
https://packages.ubuntu.com/search?suite=xenial&arch=any&mode=exactfilename&searchon=contents&keywords=upmendex

jfbu (approved):
LGTM! thanks for Ubuntu xenial link. Indeed I had not checked.

tk0miya:
Now I'm merging this. I still don't know this fix is enough or not. But we can improve this even after shipping :-)

PR #6842, opened 2019-11-17
Replace `a and b or c` by the more legible `b if a or c`.

anntzer:
Subject: Replace `a and b or c` by the more legible `b if a or c`.

### Feature or Bugfix
Style.

### Purpose
Just a small legibility improvement (IMO).

### Detail
N/A

### Relates
N/A

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -355,8 +355,8 @@ class Sphinx:
             if self._warncount and self.keep_going:
                 self.statuscode = 1
 
-            status = (self.statuscode == 0 and
-                      __('succeeded') or __('finished with problems'))
+            status = (__('succeeded') if self.statuscode == 0
+                      else __('finished with problems'))
             if self._warncount:
                 if self.warningiserror:
                     msg = __('build %s, %s warning (with warnings treated as errors).',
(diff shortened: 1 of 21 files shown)


tk0miya (changes requested):
It seems some tests are failed. Could you check them please?

tk0miya:
IMO, -0 for this. I don't like python's ternary operator because it is not easily understandable to me. `value_a if cond else value_b` is not legible to me. I'm confused always why condition-block is placed at middle in the sentence. I know this is my personal opinion. It's bikeshed discussion. But...

anntzer:
Oops, should be fixed now.
Style-wise I can't help if you don't like `true if cond else false`; I think it's much more common nowadays (and thus more readable for most devs) than `conf and true else false` but ultimately it's your choice, so feel free to close if you don't want it.

stephenfin (approved):
I'm personally in favour of this, for the opposite of the reason mentioned above (I find this more readable). Willing to be overruled though

tk0miya (approved):

tk0miya:
Okay, let's merge this. My opinion is not important. I know new style is well used.
BTW, could you rebase this into 2.0 branch please? It is much better to develop Sphinx to us.

anntzer:
rebased

tk0miya:
Thank you for rebasing. Merged!

PR #6844, opened 2019-11-17
6824 fix search js bug

### LINKED ISSUE #6824: The addition of partial search support in the HTML builds has a logic error
**Describe the bug**
In certain edge cases, the 'Search module' in searchtools.js will exit the 'performTermsSearch' method early (in the 'check if all requirements are matched' section), resulting in no results being returned, when there should be results returned.

**To Reproduce**
There may be other ways that this error manifest, but I'll give the simplest steps that I'm aware of.
1. quick-start a project
2. add a directory ('foo') and a page ('bar.rst') within foo directory
2. add foo to the index.rst, and change the max depth to 1
3. in bar.rst, add a title ('test title')
4. add some paragraph text containing the term 'mysql' and no other terms that include 'sql'
5. add a section title containing the term 'sql' and no other terms that include 'sql'
6. finally do a search for sql. It should return no results if the bug manifest.

**Expected behavior**
The 'performTermsSearch' method was created before 'support for partial matching' was added earlier this year. The partial matching allows for a condition where a partial match can exist in the 'titleterms' _and_ a full match can exist in the 'terms'. Whereas the performTermsSearch method is coded expecting for this condition of matches in both titleterms and terms to never be able to occur.

**Additional context**
I think the fix is within these two code blocks. 

```
        // set score for the word in each file to Scorer.term
        for (j = 0; j < _files.length; j++) {
          file = _files[j];
          if (!(file in scoreMap))
            scoreMap[file] = {}
--------->else if (word in scoreMap[file])
----------->scoreMap[file][word] += o.score;
--------->else
----------->scoreMap[file][word] = o.score;
        }
      });

      // create the mapping
      for (j = 0; j < files.length; j++) {
        file = files[j];
        if (file in fileMap) {
--------->if (!(fileMap[file].includes(word)))
              fileMap[file].push(word);
        }
        else
          fileMap[file] = [word];
      }
    }


```

TimKam:
Subject: Fix bug in JS search

### Feature or Bugfix
- Bugfix

### Purpose
Fix: HTML search: If a search term is partially matched in the title and fully matched in a text paragraph on the same page, the search does not include this match.

### Detail
When I added partial matching, I did not account for this case, in which matches are added to the results overview of a file the twice. I extended a condition to prevent this.

### Relates
Fixes #6824

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -70,6 +70,9 @@ Bugs fixed
 * #6813: An orphan warning is emitted for included document on Windows.  Thanks
   to @drillan
 * #6850: Fix smartypants module calls re.sub() with wrong options
+* #6824: HTML search: If a search term is partially matched in the title and
+  fully matched in a text paragraph on the same page, the search does not
+  include this match.
 
 Testing
 --------
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -424,7 +424,7 @@ var Search = {
         for (j = 0; j < _files.length; j++) {
           file = _files[j];
           if (!(file in scoreMap))
-            scoreMap[file] = {}
+            scoreMap[file] = {};
           scoreMap[file][word] = o.score;
         }
       });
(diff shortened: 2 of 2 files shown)


tk0miya (approved):
LGTM. But bug fixes should be merged into 2.0 branch. Please rebase this before merging.

PR #6845, opened 2019-11-18
LaTeX: stop escaping | to LaTeX macro, except if in an index entry

jfbu:
### Feature or Bugfix
- Feature ?
- Bugfix ?

### Relates

As discussed at #6835 the sole reason found so far why `|` is submitted to TeX escaping is that it has rôle in the syntax of the LaTeX `\index` macro. 

This PR thus limits escaping to this sole context via the `visit_index()`  of LaTeX writer. Due to an upstream [latex bug](https://github.com/latex3/hyperref/issues/109), the escaping can not be to the `"|` format as described in the LaTeX companion book, we use a workaround, somewhat like for handling indexing with `{` and `}` characters.

Testing can only be done via a PDF build, I tested that this input works fine:

```
AA

.. index:: a|b

.. index:: a|b{c}

.. index:: a|{b}|c

BB
```

and for outside of index entries, there does not seem to be any reason to escape the `|`.

NOTE: +0 for this proposal, because escaping sometimes has advantages on LaTeX side  it is very easy to redefine a macro and much more complicated with  a bare character such as `|` (but if we wnated that we should use `\sphinxtextbar` not `\textbar` because redefining the latter directly could cause problems).

On the other hand using bare `|` makes smaller files and marginally faster builds...

Besides I should have proposed for master branch, but I needed some work only in 2.0 branch and did not merge it yet.

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -1881,10 +1881,9 @@
     \let\sphinxtermref  \@firstofone
 }}
 
-% For curly braces and vertical bar inside \index macro
+% For curly braces inside \index macro
 \def\sphinxleftcurlybrace{\{}
 \def\sphinxrightcurlybrace{\}}
-\def\sphinxverticalbar{|}
 
 % Declare Unicode characters used by linux tree command to pdflatex utf8/utf8x
 \def\spx@bd#1#2{%
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -1881,9 +1881,10 @@
     \let\sphinxtermref  \@firstofone
 }}
 
-% For curly braces inside \index macro
+% For curly braces and vertical bar inside \index macro
 \def\sphinxleftcurlybrace{\{}
 \def\sphinxrightcurlybrace{\}}
+\def\sphinxverticalbar{|}
 
 % Declare Unicode characters used by linux tree command to pdflatex utf8/utf8x
 \def\spx@bd#1#2{%
--- a/sphinx/texinputs/sphinx.xdy
+++ b/sphinx/texinputs/sphinx.xdy
@@ -109,7 +109,6 @@
 
 (merge-rule "\sphinxleftcurlybrace{}" "{" :string)
 (merge-rule "\sphinxrightcurlybrace{}" "}" :string)
-(merge-rule "\sphinxverticalbar{}" "|" :string)
 (merge-rule "\_" "_" :string)
 (merge-rule "{[}" "[" :string)
 (merge-rule "{]}" "]" :string)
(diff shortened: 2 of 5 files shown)


sphinx/texinputs/sphinx.sty
```diff
@@ -1881,9 +1881,10 @@
     \let\sphinxtermref  \@firstofone
 }}
 
-% For curly braces inside \index macro
-\def\sphinxleftcurlybrace{\{}
-\def\sphinxrightcurlybrace{\}}
+% For curly braces and vertical bar inside \index macro
+\def\sphinxleftcurlybrace#1{\{}
+\def\sphinxrightcurlybrace#1{\}}
+\def\sphinxverticalbar#1{|}
```
jfbu: The extra `#1` serve to swallow the brace pairs `{}` added after the macro names (mandatory to avoid letters following glued to macro name) <s>thus producing better `.idx` files with no extraneous `{}`'s.</s> sorry this statement was wrong because I was confused at that time. These macros end up as is in the `.idx` file, except if `\index` macro is used inside some argument of a LaTeX macro. Thus the sorting done by makeindex/xindy will be with extra `{}` anyway. Having or not the `#1` is thus almost of no importance, final result is identical.

tk0miya (approved):

jfbu:
Thanks for reviewing! I have finally opted for maximal backwards compatibility to use `\textbar{}` escaping not `\sphinxverticalbar{}`, this way there is no modification of the ordering in the indices for Sphinx projects. And the other reason is that when `index` directive or rôle  is used in a table cell and the table is rendered by `tabulary` in LaTeX output, the`.idx` file is modified and with the non-robust, non-protected `\sphinxverticalbar` of previous commit, this actually gave faulty syntax (I should have `\protected` the `\sphinxverticalbar` or use some other escaping such as `{"|}`).  I have indeed discovered on this occasion that using index directive or rôle in table cell for the cases where the LaTeX output uses `tabulary` for the table can lead, generally speaking, to duplicated index entries, but this will be handled elsewhere. I will merge when testing completes. Sorry for LaTeX complications (there are always LaTeX complications!)

jfbu:
For the record, the `|` needs to be escaped in absence of `\usepackage[T1]{fontenc}` because the legacy TeX uses then OT1-encoded Knuth fonts and the `|` slot (ascii 124) gives a long horizontal dash (except for typewriter font). And the `\textbar` macro picks up the glyph in another TeX font file automatically. Thus old books probably say to escape it. But I think `\usepackage[T1]{fontenc}` was in Sphinx LaTeX writer from the beginnings and then the `\textbar` is superfluous, because T1 encoded fonts do have a `|` at slot location 124. I tested other font encodings such as X2 (Slavic languages) and the `|` also works out of the box for them. I expect only the "old TeX encoding" OT1 had this problem.

PR #6851, opened 2019-11-20
Fix the URL to https://pygments.org/docs/lexers.html

sylvestre:
https://www.sphinx-doc.org/en/1.8/usage/restructuredtext/directives.html#showing-code-examples
is pointing to https://pygments.org/docs/lexers/ (404)


tk0miya:
Thanks!

PR #6854, opened 2019-11-22
DOC: add Greek to list of supported languages

jfbu:
A new volunteer (see https://www.transifex.com/sphinx-doc/public/) has completed translation of Sphinx into Greek!

Support for Greek as main language for LaTeX is there since 2.2.1. I don't know if any special difficulty with other builders than HTML and LaTeX which would explain why Greek was not yet listed (it already got some translation work back in 2015).


tk0miya:
Thanks!

PR #6856, opened 2019-11-22
Fix #6855: Non-RST translated text should be parsed by the appropriate parser

### LINKED ISSUE #6855: Non-RST source text is parsed as RST after translation
**Describe the bug**

a12399dbe429f552ab012099491c577de668a81d introduced a regression: If the source text is in Markdown, it gets parsed as RST after translation.

In Sphinx 1.6.7 (and earlier), in `publish_msgstr`, `reader.read()` (i.e. `SphinxBaseReader.read()`) would select the parser based on the `source_path` suffix. After that commit, the parser is instead set by:

```python
parser = app.registry.create_source_parser(app, '')
```

This means that it's always RST that parses the translated text. Obviously, if the translated text is in Markdown syntax, this will not result in a correct translation.

I'll submit a PR that determines the filetype based on the `source_path`. Ideally, this can be applied to 2.2.x.

**To Reproduce**

I have open PRs on CommonMark.py and recommonmark that also need to be merged to be able to translate Markdown in Sphinx, so it will be hard to create a 'small' example.

**Expected behavior**

I expect the translated text to be parsed as Markdown (not RST) if the source text is Markdown.

**Your project**

https://github.com/open-contracting-extensions/public-private-partnerships

**Screenshots**

N/A

**Environment info**

- OS: Mac
- Python version: 3.6.1
- Sphinx version: 2.2.1
- Sphinx extensions:  recommonmark
- Extra tools: N/A

**Additional context**

This was originally fixed in aefab514e2bbce775a00c1aaae057044a02033fb and then broken in a12399dbe429f552ab012099491c577de668a81d

I have open PRs on CommonMark.py and recommonmark that also need to be merged to be able to translate Markdown in Sphinx: 

* https://github.com/readthedocs/commonmark.py/pull/225
* https://github.com/readthedocs/recommonmark/pull/187

jpmckinney:
Closes #6855

### Feature or Bugfix

- Bugfix

### DIFF
--- a/sphinx/errors.py
+++ b/sphinx/errors.py
@@ -126,3 +126,8 @@ class PycodeError(Exception):
 class NoUri(Exception):
     """Raised by builder.get_relative_uri() if there is no URI available."""
     pass
+
+
+class FiletypeNotFoundError(Exception):
+    "Raised by get_filetype() if a filename matches no source suffix."
+    pass
--- a/sphinx/io.py
+++ b/sphinx/io.py
@@ -26,7 +26,7 @@ from sphinx.transforms.i18n import (
     PreserveTranslatableMessages, Locale, RemoveTranslatableInline,
 )
 from sphinx.transforms.references import SphinxDomains
-from sphinx.util import logging
+from sphinx.util import logging, get_filetype
 from sphinx.util import UnicodeDecodeErrorHandler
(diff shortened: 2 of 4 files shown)


tk0miya:
I merged this manually into 2.0 (see: #6927).

PR #6872, opened 2019-11-30
builders/linkcheck: include "experimental" HTTP 308 as "permanently"

birkenfeld:
Also remove redundant "default" case.


tk0miya (commented):
LGTM, boss. Could you rebase this to 2.0 branch? Then I'll ship this in next stable release (maybe in next weekend).
Thanks,

tk0miya:
Note: it seems tests were failed. But it is not related with this patch, and it was already fixed at HEAD of master.

birkenfeld:
Rebased! Nice to be committing again, even if it's only one line :)

tk0miya:
Merged. Thanks,
We're always welcome you :-)

PR #6873, opened 2019-11-30
Close #5923: autodoc: allow not to document inherited members of specific super class

### LINKED ISSUE #5923: `special-members` with `inherited-members` shows all default special methods
When using `special-members` together with `inherited-members`, the default versions of special methods which were not specialized are shown, which means the doc for even the simplest class will list all special methods such as `__len__`, etc. (with a docstring such as "Return len(self).").

It would be nice if autodoc could detect that a special method was not defined and skip documenting it.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #5923

### DIFF
--- a/doc/usage/extensions/autodoc.rst
+++ b/doc/usage/extensions/autodoc.rst
@@ -157,7 +157,7 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
 
    * For classes and exceptions, members inherited from base classes will be
      left out when documenting all members, unless you give the
-     ``inherited-members`` flag option, in addition to ``members``::
+     ``inherited-members`` option, in addition to ``members``::
 
         .. autoclass:: Noodle
            :members:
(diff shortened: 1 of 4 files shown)


doc/usage/extensions/autodoc.rst
```diff
@@ -166,11 +166,19 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
      This can be combined with ``undoc-members`` to document *all* available
      members of the class or module.
 
+     It can take an anchestor class not to document inherited members from it.
+     By default, members of ``object`` class are not documented.  To show them
+     all, give ``None`` to the option.
```
shimizukawa, line 171: I suggest appending 2 example sentences to avoid misunderstanding by users.

-------

For example; If your class `Foo` is derived from `list` class and you don't want to document `list.__len__`, you should specify a option `:inherited-members: list` to avoid special members of `list` class.
Another example; If your class `Foo` has `__str__` special method and autodoc directive has both `inherited-members` and `special-members`, `__str__` will be documented as in the past, but other special method that are not implemented in your class `Foo`.

shimizukawa (approved):
LGTM with a suggestion.

doc/usage/extensions/autodoc.rst
tk0miya: added.

tk0miya:
Merged.
@shimizukawa Thank you for reviewing! I'm encouraged by your review :-)

PR #6877, opened 2019-11-30
LaTeX: avoid ragged line endings in case of multi-line author display

### LINKED ISSUE #6876: LaTeX: mark-up using`\\and` to separate authors in `latex_documents` may lead to ragged edge
When using `latex_documents` to properly use the official LaTeX separator `\and` of multiple authors (for `\author` command), if they are enough authors to let the output occupy multiple lines, only the last line of multi-line display will be correctly aligned to the right margin.

(not using `\and` leads to overflow see #6875, and one can also try `\\` which will stack things one above the other, but each will be centered, whereas the whole thing is globally flushed to the right of the page, and the end result looks quite bad)

Example in conf.py:
```
master_doc = 'index'
author = 'John Hunter, Darren Dale, Eric Firing, Michael Droettboom and the matplotlib development team'
# other stuff....
# latex config
latex_documents = [
    (master_doc, 'foo.tex', 'Project',
     author.replace(', ', '\\and ').replace(' and ', '\\and and '),
     'manual'),
]
```
which means that LaTeX file will contain this command
```latex
\author{John Hunter\and Darren Dale\and Eric Firing\and Michael Droettboom\and and the matplotlib development team}
```
It does correspond to the officlal LaTeX recommandations (cf this snapshot from the official LaTeX book)
Capture d’écran 2019-11-30 à 22 39 41

But one obtains the following ragged looks:
Capture d’écran 2019-11-30 à 22 33 01

Only the last line is well justified (due to #6005; but this also caused it to not abutt at same position as other lines).

The root problem is that the whole LaTeX mechanism for using `\author` data was initially thought out  (3 decades ago?) for *centering* on the page the author names (article and book LaTeX document classes do that). The code in `sphinxmanual.cls` uses `flushright` but it did not modify what `\and` does, and it took over "as is" the original article or book class code (which uses `tabular` in order to be able via `\\` to stack each author name on top of its address or institution, in a centered way).

For example it appears impossible without contorsions to have each author name on its line, aligned at the right margin. One can not do tricks inside `\author{..}` also because the contents must be genuine enough to not offset hyperref dealings.

The fix at #6005 was only for the case where all authors fit on a single line. But if they are many authors, or their names are long, the LaTeX code will create a multi-line rendering and only the last one will have correct alignement with respect to the document title.

A possible fix is to redefine `\and` command with gory LaTeX. Original definition is `\end {tabular}\hskip 1em \@plus .17fil\begin {tabular}[t]{c}` and if redefined according to
```
\DeclareRobustCommand\and{\end{tabular}\nobreak\kern-\tabcolsep
    \allowbreak\hskip\dimexpr1em+\tabcolsep plus .17fil\begin{tabular}[t]{c}}
```
this would fix the issue.
Capture d’écran 2019-11-30 à 22 58 06

It is very far from providing a well though out template like structure for title page, but Sphinx code is simply very close to original LaTeX code from the 1980's which was very far from being easily  customizable.

jfbu:
Closes #6876

- Bugfix

### Relates
- #6005

The comment at #6005 about `\begin{tabular}[c@{}]` applies here too, as using this in `\sphinxmaketitle` would need to be matched anyhow also with a change to  `\and` macro. This patch only redefines `\and`.

The output for projects with author names fitting on one signe line is exactly unchanged.

Sphinx users can obtain some special effect by own definition of `\and`, code comment indicates how to get each author name on its own line (however if mark-up uses `\\` to add some address underneath of author name, the result will look bad due to `c` usage in `tabular` giving centered columns; one can replace by `r` in `\and` redefinition, but there still remains one `c` in `\sphinxmaketitle` macro...)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -80,6 +80,7 @@ Bugs fixed
   include this match.
 * #6848: config.py shouldn't pop extensions from overrides
 * #6867: text: extra spaces are inserted to hyphenated words on folding lines
+* #6876: LaTeX: multi-line display of authors on title page has ragged edges
 
 Testing
 --------
--- a/sphinx/texinputs/sphinxhowto.cls
+++ b/sphinx/texinputs/sphinxhowto.cls
@@ -3,7 +3,7 @@
 %
 
 \NeedsTeXFormat{LaTeX2e}[1995/12/01]
-\ProvidesClass{sphinxhowto}[2018/12/23 v2.0 Document class (Sphinx howto)]
+\ProvidesClass{sphinxhowto}[2019/12/01 v2.3.0 Document class (Sphinx howto)]
 
 % 'oneside' option overriding the 'twoside' default
 \newif\if@oneside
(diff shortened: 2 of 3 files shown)


tk0miya (approved):
LGTM

jfbu:
Thanks for quick reviewing! I know my LaTeX explanations at #6876 and here are long and verbose... but it helps me document details for future maintenance either for me as I quickly forget or other maintainers :). Merging now.

PR #6883, opened 2019-12-01
Fix #6821: i18n: some translation messages has been broken

### LINKED ISSUE #6821: Translatable strings using `%r` and `%s` cause problems if translation reverses order
On master branch (at circa ``14-11-2019 16:42:29 CET`` for transifex pull)

```
cd sphinx/locale
tx pull -f -a
cd ../..
python setup.py compile_catalog
```

output (trimmed)

```
running compile_catalog
...
error: sphinx/locale/ja/LC_MESSAGES/sphinx.po:1280: positional format placeholders are unbalanced
error: sphinx/locale/ja/LC_MESSAGES/sphinx.po:2415: incompatible format for placeholder 1: 'r' and 's' are not compatible
...
error: sphinx/locale/zh_CN/LC_MESSAGES/sphinx.po:403: incompatible format for placeholder 1: 'r' and 's' are not compatible
compiling catalog sphinx/locale/zh_CN/LC_MESSAGES/sphinx.po to sphinx/locale/zh_CN/LC_MESSAGES/sphinx.mo
...
Compiling failed.
```

Apart from the `％` problem at `sphinx/locale/ja/LC_MESSAGES/sphinx.po:1280` (there were others earlier so I think transifex has been updated recently) the problem is with strings such as 

```
#: sphinx/ext/coverage.py:46
#, python-format
msgid "invalid regex %r in %s"
msgstr "%s 内に無効な正規表現 %r があります"
```

where translation permutes order. For Chinese:

```
#: sphinx/registry.py:464
#, python-format
msgid ""
"the extension %r was already merged with Sphinx since version %s; this "
"extension is ignored."
msgstr "自版本 %s 开始，扩展 %r 已合并至 Sphinx；该扩展被忽略。"
```

When more translation work will be done at transifex more and more such occurrences will arise for languages where order is permuted with respect to English.

<s>It seems usage of `.fmt` should solve this.</s>

edit: I of course meant "usage of `.format()` would solve this"

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6821


jfbu (approved):
LGTM ;-) I checked French and even Chinese :stuck_out_tongue_winking_eye:

jfbu (commented):
I fixed a typo upstream at transifex for French, can you pull it?

Side note: I noticed some strings do not get translated because the catalog files are (afaik) for 2.0 branch and this is 2.2.2 branch. (this is the case for  ``__('generating indices')`` in sphinx.builders.html). This looks unavoidable for 2.2.2.

(regarding French I prefer the full import as it has many language fixes compared to only cherry-picking the sole Sphinx breaking strings, even though some translation work will remain invisible before 2.3.0 release, thus I am quite happy with `tx pull -a -f`)

tk0miya:
updated. I will merge this after CI passed.

PR #6885, opened 2019-12-04
Ensure getdoc always trims whitespace

eric-wieser:
This change makes it so the whitespace is always trimmed, for consistency.

### Feature or Bugfix

- Bugfix

### Purpose
Make it so inherited and non-inherited docstrings with `autoclass_content = both` render the same way.

### Detail
Before this patch:
* `getdoc(obj, allow_inherited=True)` trims whitespace
* `getdoc(obj, allow_inherited=False)` does not trim whitespace

After:
* Both calls trim whitespace

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -609,7 +609,10 @@ def getdoc(obj: Any, attrgetter: Callable = safe_getattr,
     doc = attrgetter(obj, '__doc__', None)
     if ispartial(obj) and doc == obj.__class__.__doc__:
         return getdoc(obj.func)
-    elif doc is None and allow_inherited:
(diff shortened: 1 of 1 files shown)


eric-wieser:
Looks like I was overly optimistic. I'll check this out locally at some point and start debugging.

The alternative option here, instead of always cleaning docstring, would be to never clean them (and vendor or just use `inspect._finddoc`).

tk0miya:
+1: Reasonable.

tk0miya:
Is there anything to do from ourside? I'd like to see a real case or a test case for this PR.

tk0miya:
I'm closing this because not updated. Please post a new PR if improved.
Thanks,

Issue #3181, opened 2016-11-25
pLaTeX crashes with a section contains endash


tk0miya:
In Japanese docs, pLaTeX crashes with a section contains emdash
```
EN DASH:  –
-------------
```
```
$ make clean latexpdfja
...
LaTeX Font Warning: Font shape `JT1/mc/bx/it' undefined
(Font)              using `JT1/mc/bx/n' instead on input line 143.


LaTeX Font Warning: Font shape `JY1/mc/bx/it' undefined
(Font)              using `JY1/mc/bx/n' instead on input line 143.

(/usr/local/texlive/2016/texmf-dist/tex/latex/amsfonts/umsa.fd)
(/usr/local/texlive/2016/texmf-dist/tex/latex/amsfonts/umsb.fd) [1] [2]
(./test.toc
! Text line contains an invalid character.
l.1 ...sline {chapter}{\numberline {1}EN DASH: ^^@
                                                  ^^80^^93}{1}{chapter.1}
```

tk0miya:
Fixed at 69820e1.
I replaced the EM DASH character by `\textemdash{}` macro.

jfbu:
I know nothing about Japanese and platex. But I am wondering if some set-up related to platex is missing. Your example is with EN DASH (U+2013). If I insert an EM DASH (U+2014) I notice that platex appears to convert it to HORIZONTAL BAR (U+2015) in aux and toc file (but might be after-effect if platex is not at all unicode aware). I tried with `jreport` class, it then seems ok in pdf output. But there is problem with EN DASH, why is that (I tried ``platex -etex`` but it doesn't change anything). 

I tried``uplatex`` with ``utarticle`` class. The EN DASH and EM DASH end up as is in aux files. The EN DASH seems ok in pdf but the EM DASH ends up rotated it seems with respect to text.

tk0miya:
Thank you for your help :-)

The unicode support of pLaTeX is not complete. pLaTeX uses EUC-JP or Shift-JIS as a internal encoding. On my local, EUC-JP is used to internal encoding. But, unfortunately, EN DASH is not contained in the character set of EUC-JP (a.k.a JIS X 0208; JIS means Japan Industrial Standard).
I guess this is why the EN DASH is broken on aux file.

Note: EM DASH is contained in JIS X 0208. so it is not broken.

[...]

PR #6888, opened 2019-12-04
LaTeX: inhibit so-called TeX Ligatures with xelatex and lualatex

### LINKED ISSUE #6886: PDF build via xelatex transforms straight double quotes into curly ones (smartquotes being turned off)
Consider following source
```rest
Test
====

"versions" 'version'
```
And

```
latex_engine = 'xelatex'
smartquotes = False
```

in conf.py. On ``make latexpdf`` we find in PDF output curly quotes ``”versions” ’version’``

Note that the latex file itself does correctly contain ``"versions" 'version'``
**Environment info**
- OS: Mac
- Python version: 3.7.5
- Sphinx version: 2.x
- Sphinx extensions:  none
- Extra tools: TeXLive 2019

This is reminiscent to  #5179 (which was lualatex specific) and the fix as #5790 works.

In contrast,  straight double quotes remains with pdflatex straight in PDF output as long as `\usepackage[T1]{fontenc}` is used, which is Sphinx default. Thus `xelatex/lualatex` behaves differently on same latex input.

But ``'version'`` in input will give ``’version’`` in PDF for all engines (not tested Japanese).  I am surprised that this has not been raised as issue so far, I need to investigate a bit, keeping it separate. With the default `smartquotes = True` the output will have the expected typographical curly quotes``‘version’`` (and ``“versions”``) expected for English language.

jfbu:
Closes: #6886

It is difficult to understand why fontspec _by default_ turns straight
double quotes " into ” i.e. right curly ones, calling this "TeX
ligature", but that's life and fortunately fontspec also provides a way
to turn this off. With this workaround also a straight ' in LaTeX file,
(i.e. a straight ' in source and smartquotes = False was used) will not
become a curly quote and this is an advantage compared to the pdflatex
situation.

We suppress TeX Ligatures also for \ttfamily (used in code-blocks) but
that appears to be already the case with fontspec although user manual
does not seem explicit (and no list of what exactly the TeX Ligatures
are is to be found in fontspec user documentation, one has to go into
its source code). (the fact that Pygmentize escapes the " to \char`\"
TeX mark-up would not by itself prevent the transformation into a curly
double quote, it is because fontspec does not apply this transform to
monospace typeface that it did not show).

### Feature or Bugfix
- Bugfix

### Relates

Closes #6886. Relates #5790 in so far as the fix used at #5790 for a lualatex behavior is taken over and applied also to xelatex.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -81,6 +81,8 @@ Bugs fixed
   include this match.
 * #6848: config.py shouldn't pop extensions from overrides
 * #6867: text: extra spaces are inserted to hyphenated words on folding lines
+* #6886: LaTeX: xelatex converts straight double quotes into right curly ones
+  (shows when :confval:`smartquotes` is ``False``)
 * #6876: LaTeX: multi-line display of authors on title page has ragged edges
 
 Testing
--- a/doc/latex.rst
+++ b/doc/latex.rst
@@ -334,12 +334,19 @@ into the generated ``.tex`` files.  Its ``'sphinxsetup'`` key is described
      .. versionchanged:: 2.0
         ``'lualatex'`` executes
         ``\defaultfontfeatures[\rmfamily,\sffamily]{}`` to disable TeX
-        ligatures.
+        ligatures transforming `<<` and `>>` as escaping working with
+        ``pdflatex/xelatex`` failed with ``lualatex``.
      .. versionchanged:: 2.0
         Detection of ``LGR``, ``T2A``, ``X2`` to trigger support of
         occasional Greek or Cyrillic (``'pdflatex'`` only, as this support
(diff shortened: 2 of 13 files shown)


jfbu (commented):
There is a problem that https://github.com/sphinx-doc/sphinx/blob/master/sphinx/templates/latex/longtable.tex_t contains an instance of `--`. This will now render as two hyphens rather than an en dash with xelatex/lualatex. In fact it should be coded directly as Unicode en dash `–` (U+2013), or perhaps `\textendash{}` (see #3181) (I tested with platex a Unicode endash and had no problem). I will push a commit.

tk0miya (approved):
LGTM!

jfbu:
Thanks for reviewing!

Issue #6890, opened 2019-12-05
LaTeX: even with smartquotes off, PDF output transforms straight quotes and consecutive hyphens into curly quotes and dashes


jfbu:
With this input

```
'test' "test" test--test---test...
```

and `smartquotes = False` in conf.py, on doing `make latexpdf`, the PDF output will contain

```
’test’ "test" test–test—test...
```

i.e. the straight quote is converted to a right curly quote, the double quote is intact, the hyphens give en-dash and em-dash, the triple dots is ok, it does not give ellipsis.

Using `xelatex`  the PDF is slightly different

```
’test’ ”test” test–test—test...
```

i.e. also the double quote is curly (right curly double quote). This is already reported #6886.

**Screenshots**

using pdflatex

![Capture du 2019-12-05 14-54-56](https://user-images.githubusercontent.com/2589111/70241399-42ef2900-176f-11ea-94ed-a70eb07de1ed.png)

using xelatex

![Capture du 2019-12-05 14-54-03](https://user-images.githubusercontent.com/2589111/70241404-471b4680-176f-11ea-80eb-5b495c5e50de.png)

using platex (Japanese)

![Capture du 2019-12-05 15-10-37](https://user-images.githubusercontent.com/2589111/70242640-7468f400-1771-11ea-9e7f-27ee3b6b09d1.png)

**Environment info**
- OS: [e.g. Unix/Linux/Mac/Win/other with version]
- Python version: 3.7.5
- Sphinx version: tested with 1.8.5 and 2.x, applies to all versions I believe (prior to 1.6.6 there was no smartquotes toggle)
- Sphinx extensions: N/A
- Extra tools: TL2019. Regarding xelatex perhaps issue did not show with old fontspec, it is not documented in user manual when applying "TeX ligatures" became default, and it may depend also on the LaTeX release date as some versions of Sphinx prior to 2.x left LaTeX to decide default fonts to use for xelatex.

**Additional context**

This problem has always been there I believe. For example with the [sphinx.util.texescape as at Sphinx 1.5](https://github.com/sphinx-doc/sphinx/blob/v1.5/sphinx/util/texescape.py), the problem applied. It has evolved in backwards compatible way since. For standard text Sphinx applies some tex escaping to prevent TeX ligatures e.g. transform of `>>`, but it never did any specific TeX escaping of the straight quote  `'` or the hyphen  `- `. 

[...]

PR #6891, opened 2019-12-05
LaTeX: avoid quotes and TeX ligature replacements in PDF output

jfbu:
Refs: #6890

<s>
But I had to leave out the hyphen because replacing it would break URLs
containing it, e.g. the ones generated by the pep role.
</s>

*Finally I also escape the hyphen to avoid en-dash and em-dash ligatures.*

<s>
By the way I wonder if URLs should get some specific restricted TeX escaping, in a way similar to the highlighting variant of tex escaping.
</s>

*Not clear what I meant.*

<s>
About tex escaping of highlighted blocks, I had to be careful about the latter, because it applies not only to the body of code-block but also to its start
~~~
\begin{Verbatim}[commandchars=\\\{\}]
~~~
coming back from Pygments but that part may have options, separated by commas, and the comma ended up escaped. It would be simpler not to worry about this.
</s>

edit: *Anyway, I opted later on to not escape the comma. Thus the problem is not urgent, but still, LaTeX escaping of options to `\usepackage{Verbatim}[options,...]` coming back from Pygments looks like not a correct thing to do, in principle. It does not harm currently but nevertheless I did separate `ascii_tex_replacements` which will not apply there.*

Tex escaping of highlighted blocks makes sense only for Unicode replacements because anyway all special chars are either Pygments-escaped or are allowed to be there as is in Verbatim. But currently some Unicode replacements are still in the same map as escaping of TeX specials, so situation is not optimal.

### Feature or Bugfix
- Bugfix


sphinx/util/texescape.py
```diff
@@ -49,6 +43,22 @@
...
 ]
 
+# A map to avoid TeX ligatures or character replacements in PDF output
+# xelatex/lualatex/uplatex are handled differently (#5790, #6888)
+ascii_tex_replacements = [
+    # Note: the " renders curly in OT1 encoding but straight in T1, T2A, LY1...
+    #       escaping it to \textquotedbl would break documents using OT1
+    #       Sphinx does \shorthandoff{"} to avoid problems with some languages
+    # Can't replace - by {}- because it breaks URLs
+    # ('-', '{}-'),  # -- and --- are TeX ligatures
```
jfbu: Some solution is needed to escape - to {}- only outside of URLs

sphinx/util/texescape.py
```diff
@@ -49,6 +43,22 @@
     # OHM SIGN U+2126 is handled by LaTeX textcomp package
 ]
 
+# A map to avoid TeX ligatures or character replacements in PDF output
+# xelatex/lualatex/uplatex are handled differently (#5790, #6888)
+ascii_tex_replacements = [
```
jfbu, line 48: I would prefer tex_ascii_replacements but the other one was already called unicode_tex_replacements ...

sphinx/util/texescape.py
```diff
@@ -130,6 +140,10 @@ def init() -> None:
         _tex_escape_map_without_unicode[ord(a)] = b
         tex_replace_map[ord(a)] = '_'
 
+    for a, b in ascii_tex_replacements:
```
jfbu, line 150: the idea is that this will not be done for the _tex_hlescape_map and _tex_hlescape_map_without_unicode. One reason is that highlighter tex escaping includes the option of Verbatim which are comma separated and replacing `,` by `{},` breaks sphinxVerbatim

sphinx/util/texescape.py
```diff
@@ -49,6 +43,22 @@
...
 
+# A map to avoid TeX ligatures or character replacements in PDF output
+# xelatex/lualatex/uplatex are handled differently (#5790, #6888)
+ascii_tex_replacements = [
+    # Note: the " renders curly in OT1 encoding but straight in T1, T2A, LY1...
+    #       escaping it to \textquotedbl would break documents using OT1
+    #       Sphinx does \shorthandoff{"} to avoid problems with some languages
+    # Can't replace - by {}- because it breaks URLs
+    # ('-', '{}-'),  # -- and --- are TeX ligatures
+    (',', '{},'),  # ,, is a TeX ligature in T1 encoding
```
jfbu: maybe this is pushing too far doing that? , ,  in non  literal text is probably a typo anyhow. And this change forced to adapt a few tests.

tests/test_build_latex.py
```diff
@@ -1416,7 +1416,7 @@ def test_default_latex_documents():
     config.init_values()
     config.add('latex_engine', None, True, None)
     expected = [('index', 'stasi.tex', 'STASI™ Documentation',
-                 r"Wolfgang Schäuble \& G'Beckstein.\@{}", 'manual')]
+                 r"Wolfgang Schäuble \& G\textquotesingle{}Beckstein.\@{}", 'manual')]
```
jfbu, line 1419: the handling of author by tex escaping seems to be oblivious to smartquotes setting of course as it is not parsed by Docutils I believe. This thus is breaking change because people using  a straight quote in author names saw a curly one in PDF output, but now it will be a straight one (independently of smartquotes setting)

jfbu (commented):
I have not tested if this is ok with uplatex.

sphinx/util/texescape.py
```diff
@@ -49,6 +43,22 @@
     # OHM SIGN U+2126 is handled by LaTeX textcomp package
 ]
 
+# A map to avoid TeX ligatures or character replacements in PDF output
+# xelatex/lualatex/uplatex are handled differently (#5790, #6888)
```
jfbu, line 47: by the way, shouldn't sphinx.util.texescape.escape() treats uplatex like xelatex and lualatex (untested). In fact the comment assumed that but it is not yet the case.

sphinx/util/texescape.py
jfbu: I have not tested yet but escaping rather to `\sphinxhyphen{}` with a suitable definition of `\sphinxhyphen` and a re-definition of it inside `\sphinxhref`, `\sphinxurl`, `\sphinxnolinkurl` should do it wih no need of other change in Sphinx in particular no need to change current way by latex writer to handle hyperlinks.

sphinx/util/texescape.py
```diff
@@ -49,6 +43,22 @@
...
+ascii_tex_replacements = [
+    # Note: the " renders curly in OT1 encoding but straight in T1, T2A, LY1...
+    #       escaping it to \textquotedbl would break documents using OT1
+    #       Sphinx does \shorthandoff{"} to avoid problems with some languages
+    # Can't replace - by {}- because it breaks URLs
+    # ('-', '{}-'),  # -- and --- are TeX ligatures
+    (',', '{},'),  # ,, is a TeX ligature in T1 encoding
+    # the next two require textcomp package
+    ("'", r'\textquotesingle{}'),  # else ' renders curly, and '' is a ligature
+    ('`', r'\textasciigrave{}'),   # else \` and \`\` render curly
```
jfbu, line 58: It will be possible to allow line-breaks in parsed literals at the left and right ticks by updating `\sphinxbreaksattexescapedchars` in sphinx.sty but this would be easier if escaping here is `\sphinxquotesingle` and `\sphinxasciigrave`. 

Actually all escaping should be to `\sphinx...` prefixed macros for flexibility in obtaining special effects simply by redefining them.

sphinx/util/texescape.py
jfbu: Because commas are very frequent in text, escaping to  `\sphinxcomma{}` is not very enticing. It will augment size of tex file. Overall I am actually tempted to do nothing with the comma, because avoiding the `,,` T1-ligature does not seem important. Visually the ligature looks like `,,` except it is a single character hence copying pasting from PDF will not give back `,,`. +1 for actually not doing anything with comma.

sphinx/texinputs/sphinx.sty
```diff
@@ -1879,8 +1887,18 @@
     \let\sphinxemail    \@firstofone
     \let\sphinxcrossref \@firstofone
     \let\sphinxtermref  \@firstofone
+    \def\sphinxhyphen#1{-}% the #1 is there to gobble the {}
```
jfbu: Well actually it should be without the `#1` gobbling the `{}` else hyperref seems to encode `--` into a Unicode en-dash in the PDF bookmarks, whereas a section title with `-{}-{}` does give two hyphens also in bookmarks of PDF. Will update after some additional testing.

sphinx/util/texescape.py
jfbu: (this comment is now partially obsolete because the escaping of the comma has been removed)

sphinx/writers/latex.py
```diff
@@ -487,7 +487,7 @@ def __init__(self, document: nodes.document, builder: "LaTeXBuilder") -> None:
         self.first_document = 1
         self.this_is_the_title = 1
         self.literal_whitespace = 0
-        self.no_contractions = 0
+        self.no_contractions = 0  # not used
```
jfbu: I was wondering is this needed some deprecation and how actually achieve that?

sphinx/util/texescape.py
jfbu: well actually I did move the needed changes to latex.py, not in sphinx.sty, for handling of hyperlinks

sphinx/util/texescape.py
jfbu: Done, the comma is not escaped anymore.

sphinx/writers/latex.py
tk0miya: Please replace this member by property.
This is an example.
https://github.com/sphinx-doc/sphinx/blob/24908e0fa9868ea278db25161d6b51038f1db3ca/sphinx/writers/latex.py#L2300-L2304

tests/test_build_latex.py
tk0miya: +1; Let's do smartquotes in another PR.

tk0miya (approved):
LGTM

sphinx/writers/latex.py
jfbu: Done

tk0miya (approved):

jfbu:
Thanks @tk0miya for reviewing and advice. It seems [Travis testing fails](https://travis-ci.org/sphinx-doc/sphinx/jobs/625120201?utm_medium=notification&utm_source=github_status) with some internal error for Python 3.8 also for #6908. To not interfere with your release work I prefer to leave merging to you :smile:

tk0miya:
Ya, I also saw the error. It seems the error is raised inside pytest-cov. Now I'm working for it.
I believe it would not be related with our product. But I must leave from internet a few hours. So please wait a moment.

tk0miya:
Hmm... coverage package is released just today. I guess this might be related with this problem.
https://pypi.org/project/coverage/#history

tk0miya:
I posted https://github.com/sphinx-doc/sphinx/pull/6924 to investigate the error. But it's time over. I'm away from keyboard for a while. See you later :-)

jfbu:
Thanks, I promise I won't break anything while you are offline :wink:

tk0miya:
@jfbu Could you rebase or merge HEAD of 2.0 branch please?

jfbu:
ok please wait a minute or two

jfbu:
Rebased and squashed commits.

"1 or minute or 2" actually meant 12 minutes in good French :smile:

tk0miya:
OMG, Travis CI is just under maintenance now... But we've already know this PR has been already passed. So I'll merge this without Travis CI.
https://travis-ci.org/sphinx-doc/sphinx

jfbu:
Ready for merge... thanks for reviewing and good luck with release work, now leaving ;-)

tk0miya:
Merged. Thank you always!

Issue #6900, opened 2019-12-08
Not possible to set  boolean `latex_use_xindy` via `-D` option to sphinx-build


jfbu:
**Describe the bug**

Impossible to override [`latex_use_xindy`](http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-latex_use_xindy) from command line invocation of `sphinx-build`.

**To Reproduce**
With any project

```
sphinx-build -b latex . _build/latex -D latex_use_xindy=0
```

produces

```
WARNING: The config value `latex_use_xindy' has type `str', defaults to `bool'.
```
And the latex output directory will actually be configured for PDF build to use `xindy`.

Notice that for example
```
sphinx-build -b latex . _build/latex -D latex_use_latex_multicolumn=0
```
raises no warning and behaves as expected. Same for
```
sphinx-build -b latex . _build/latex -D smartquotes=0
```
which works fine.

**Expected behavior**
People reading documentation are rewarded with Sphinx behaving as promised :smile:

**Your project**

Any project.

**Environment info**
- OS: Mac
- Python version: 3.7.5
- Sphinx version: tested with 2.0.1 and current 2.0 branch

**Additional context**

- #6898

tk0miya:
Fixed by #6908!

Issue #6898, opened 2019-12-07
Latex build should remove existing output files when it starts


jfehrle:
If you get certain severe errors from `sphinx-build`, the output files (including the pdf file) are not updated if they're already present in the build directory, whereas if the output files didn't already exist (and you get the same error) you do get useful output.  This is confusing.

In my case, I'm running on Windows/WSL, which AFAIK always fails when it tries to run `xindy` because of WSL limitations (i.e., it doesn't run a Linux virtual machine).  Even with this error, if the output files aren't already present in the directory, I get a useful pdf, albeit with some of the indexing missing.

```
Latexmk: calling xindy( -L english -C utf8 -M sphinx.xdy -o CoqRefMan.ind CoqRefMan.idx )
Cannot map memory to address 0x800965be3000 .
```

I'm told that running the build on Linux with the output files already present and without the error will indeed update the output files as expected.

**To Reproduce**
I believe this will work on Linux:
1. run sphinx on your favorite project that builds a pdf
2. rename the `xindy` binary so it won't be found
3. run sphinx again, I expect you will get the same failure behavior

**Your project**
I expect you don't want to download all of Coq's dependencies and build all of Coq.
See https://github.com/coq/coq/pull/11232 and https://github.com/coq/coq/issues/11218.

**Environment info**
- Windows 10 with WSL (Ubuntu)
- Python version: 3.6.8
- Sphinx version: 2.2.2

jfbu:
not addressing the main issue, but can you try at your locale with <s>``-D latex_use_xindy=False``</s> (edited) ``-D latex_use_xindy=0`` (I don't know if quoting is needed on Windows). As you are using xelatex, it is possible that the build will succeed despite the fact `makeindex` creates `idx` file which is not valid as UTF-8 encoded file. At least that is what I said at https://github.com/sphinx-doc/sphinx/issues/5132 when I reported using lualatex was not even possible (when some indexed word starts with a multi-byte letter; the problem is that `makeindex` picks up the first byte thinking it is a full letter to make a "big letter")

**edit**: sorry I made an error in the above the boolean value [should be set](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#id2) via ``-D latex_use_xindy=0``.  But currently it does not work.   There appears to be possibly  a bug but I need to examine more. (#6900)

jfehrle:
@jfbu Not sure what you have in mind.  There is no `-D` option for make or latexmk:

```
[...]

PR #6902, opened 2019-12-08
Allow setting boolean latex_use_xindy from command line

jfbu:
Fixes: #6900

Is there any other boolean configuration whose default is a callable?

Relates indirectly #6898 in so far as #6900 prevents indicated advice to solve #6898 use case.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* #6900: Not possible to set :confval:`latex_use_xindy` via ``-D`` option to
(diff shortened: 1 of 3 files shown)


jfbu (commented):
Arguably this is more of a workaround that a complete fix which should happen in sphinx.config.Config() class

jfbu:
I have added alternative PR #6905. Please look at it also. It identifies better the root problem.

jfbu:
I am closing this without merge because I think #6905 better puts forward the core underlying issue.

PR #6905, opened 2019-12-09
Handle latex_use_xindy bool confval with callable default specially

jfbu:
Fixes: #6900

- Bugfix

This is alternative to #6902. It identifies better the root of the problem but still feels more like a temporary workaround than a fully satisfying fix.

For this reason I did not add a test yet.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* #6900: Not possible to set :confval:`latex_use_xindy` via ``-D`` option to
+  :program:`sphinx-build`
+
 Testing
 --------
 
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -222,7 +222,7 @@ class Config:
                                  (name, name + '.key=value'))
             elif isinstance(defvalue, list):
                 return value.split(',')
-            elif isinstance(defvalue, int):
+            elif isinstance(defvalue, int) or name == 'latex_use_xindy':
                 try:
                     return int(value)
                 except ValueError:


sphinx/config.py
```diff
@@ -222,7 +222,7 @@ def convert_overrides(self, name, value):
                                  (name, name + '.key=value'))
             elif isinstance(defvalue, list):
                 return value.split(',')
-            elif isinstance(defvalue, int):
+            elif isinstance(defvalue, int) or name == 'latex_use_xindy':
```
jfbu, line 225: This solves the problem but only for this special case.

jfbu (commented):
The problem thus is that configuration value whose default is set according to a callable will never undergo during confoverrides an attempt to convert its value to int type and if it comes from command line it thus remains of str type which ultimately will raise an error if the callable return type is not str but bool or something else.

sphinx/config.py
tk0miya: -1: hard coding is not extensible. I suppose there are more boolean settings in another extension. How about the way like #6908 instead?

CHANGES
```diff
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* #6900: Not possible to set :confval:`latex_use_xindy` via ``-D`` option to
+  :program:`sphinx-build`
```
tk0miya, line 20: 2.2.3 should be used severe bug like crashes. So I prefer to use 2.0 branch instead in such case.

jfbu:
Closing this without merge as #6908 starts a better proposal. I only consider indeed this as a workaround to display where problem lies. Thanks for reviewing!

PR #6906, opened 2019-12-09
Update parser.py

kpnr:
encoding damages non-ASCII characters

- Bugfix

### Purpose
- support for sources coded in cp1251 and such codepages

### Detail
- remove errornous conversion to/from bytes/unicode

### Relates
- none

### DIFF
--- a/sphinx/pycode/parser.py
+++ b/sphinx/pycode/parser.py
@@ -479,7 +479,7 @@ class Parser:
 
     def parse_comments(self) -> None:
         """Parse the code and pick up comments."""
-        tree = ast.parse(self.code.encode())
+        tree = ast.parse(self.code)
         picker = VariableCommentPicker(self.code.splitlines(True), self.encoding)
         picker.visit(tree)
         self.comments = picker.comments


tk0miya (changes requested):
LGTM with nits.

* Could you cherry-pick this fix into 2.0 branch? Then I'll ship in next stable release
* Could you add a testcase please?

kpnr:
>     * Could you cherry-pick this fix into 2.0 branch? Then I'll ship in next stable release
I'll try :-)

>     * Could you add a testcase please?

I`ll try. But I not familar with CI/CD :-(

tk0miya:
Please add a small example using multibytes characters to `tests/test_pycode_parser.py`. I suppose original code raises an error for the case, but your fix resolves it.
https://github.com/sphinx-doc/sphinx/blob/2.0/tests/test_pycode_parser.py

kpnr:
> 
> 
> Please add a small example using multibytes characters to `tests/test_pycode_parser.py`. I suppose original code raises an error for the case, but your fix resolves it.
> https://github.com/sphinx-doc/sphinx/blob/2.0/tests/test_pycode_parser.py

It can rise if UTF16/UTF32 Python3 str type unable to convert to UTF-8 bytes. I think all UTF16/32 strings can be converted to UTF8 and vice versa, so no exception should occur.
I'll try to make test case of docstring corruption. But I have almost no free time :-((

sphinx/pycode/parser.py
```diff
@@ -479,7 +479,7 @@ def parse(self) -> None:
 
     def parse_comments(self) -> None:
         """Parse the code and pick up comments."""
-        tree = ast.parse(self.code.encode())
+        tree = ast.parse(self.code)
```
tk0miya, line 482: Note:
The reference of `ast.parse()` says:
>Parse the source into an AST node. Equivalent to compile(source, filename, mode, ast.PyCF_ONLY_AST).

And `compile()`'s says:
>source can either be a normal string, a byte string, or an AST object.

This change calls `ast.parse()` correctly. So +1 for merging without tests.

tk0miya (approved):

tk0miya:
Finally I determined to merge this without testings.

sphinx/pycode/parser.py
kpnr: Thank you for patience to such badly-skilled contributor as me :-)

PR #6907, opened 2019-12-09
Non-ASCII & non utf-8 charset corruption fix

kpnr:
Prefer native loader against fileIO

### Feature or Bugfix
- Bugfix

### Purpose
- prefer native Python loader over file-io

### Detail
- support for coding directive in Python sources

### Relates
- none

### DIFF
--- a/sphinx/pycode/__init__.py
+++ b/sphinx/pycode/__init__.py
@@ -67,7 +67,7 @@ class ModuleAnalyzer:
             if source is not None:
                 obj = cls.for_string(source, modname, filename if filename is not None else '<string>')
             elif filename is not None:
-                obj = cls.for_file(source, modname)
+                obj = cls.for_file(filename, modname)
         except PycodeError as err:
             cls.cache['module', modname] = err
             raise
--- a/sphinx/pycode/__init__.py
+++ b/sphinx/pycode/__init__.py
@@ -11,8 +11,9 @@
 import re
 from io import StringIO
 from os import path
-from typing import Any, Dict, IO, List, Tuple
+from typing import Any, Dict, IO, List, Tuple, Optional
 from zipfile import ZipFile
+from importlib import import_module
 
 from sphinx.errors import PycodeError
 from sphinx.pycode.parser import Parser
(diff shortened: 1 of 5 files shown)


sphinx/util/__init__.py
```diff
@@ -245,18 +245,17 @@ def get_module_source(modname: str) -> Tuple[str, str]:
         raise PycodeError('error importing %r' % modname, err)
     filename = getattr(mod, '__file__', None)
     loader = getattr(mod, '__loader__', None)
+    try:  # prefer Native loader, as it respects #coding directive
```
tk0miya: It is better to check loader is not None and loader has `get_source` method here.

sphinx/util/__init__.py
```diff
@@ -245,18 +245,17 @@ def get_module_source(modname: str) -> Tuple[str, str]:
...
+        filename = loader.get_source(modname)
+        if filename:
+            return 'string', filename
+    except Exception as err:
```
tk0miya: document says `loader.get_source()` raises ImportError only. So this should handle it.
https://docs.python.org/3/library/importlib.html#importlib.abc.InspectLoader.get_source

tk0miya (changes requested):
LGTM, but you need to update testcase.

In addtion, 2.0 branch is better for bugfix. Could you rebase this please?

sphinx/util/__init__.py
kpnr: Ok. I`ll fix it in a few days

sphinx/util/__init__.py
kpnr: Ok. I`ll update it

sphinx/pycode/__init__.py
```diff
@@ -63,11 +63,11 @@ def for_module(cls, modname: str) -> "ModuleAnalyzer":
             return entry
 
         try:
-            type, source = get_module_source(modname)
-            if type == 'string':
-                obj = cls.for_string(source, modname)
-            else:
-                obj = cls.for_file(source, modname)
+            filename, source = get_module_source(modname)
```
tk0miya: Why are you changing interface of `get_module_source()`? I think this is incompatible change. Unfortunately, there are many hackers to touch sphinx module directly. So it would be better not to change our interface. Or we have to obey deprecation process to change it. But it seems no special reason to change it in this time.

tests/test_pycode.py
```diff
@@ -31,12 +31,19 @@ def test_ModuleAnalyzer_for_file():
...
     assert analyzer.srcname in (SPHINX_MODULE_PATH,
                                 os.path.abspath(SPHINX_MODULE_PATH))
-    assert analyzer.encoding == 'utf-8'
+    # source should be loaded via native loader, so don`t know file enconding
+    assert analyzer.encoding == None
+    path = rootdir / 'test-pycode'
+    sys.path.insert(0, path)
+    analyzer = ModuleAnalyzer.for_module('cp_1251_coded')
+    docs = analyzer.find_attr_docs()
+    sys.path.pop(0)
```
tk0miya: It would be better to use finally block for unloading.

sphinx/pycode/__init__.py
kpnr: I change get_module_source() return type to be able to return filename and source of module together. Filename stored to Sphinx internals and used in Warnings test to show user exact warning location, source used in Analyzer to get docstrings.
Original interface can return only filename or only source, not both.
With a full search in Sphinx sources I found just two calls for get_module_source() - in pycode and in test_util and update both, so I hope change get_module_source() interface doesn't break any code.
If you see better way how to implement this - just say. I'll implement it.

tests/test_pycode.py
kpnr: Totally agree. Fixed.

sphinx/pycode/__init__.py
tk0miya: In our experience, 3rd party extension sometimes uses Sphinx core functions even undocumented. So it would be better to keep compatibility. But, as you said, it should return both values if possible.

I think it would be better to copy new `get_module_source()` function to `sphinx.pycode`. As you commented, it is used only from pycode. So it should be moved inside pycode module. And original `get_module_source()` will be marked as deprecated to remove it in future. How about this?

tk0miya:
Oh, I've forgotten to mention you to rebase this to 2.0 branch. master branch is only used for 3.0 release. 2.0 branch is better for compatible changes.

sphinx/pycode/__init__.py
kpnr: May be make it "protected" (underscored) static function in ModuleAnalizer class to mark fact that this function used only by this class? And what to do with Tests? Move from test_util to test_pycode file?

kpnr:
> 
> 
> Oh, I've forgotten to mention you to rebase this to 2.0 branch. master branch is only used for 3.0 release. 2.0 branch is better for compatible changes.

You write it in first comment. I rebase my branch to "2.0" (base at commint 1044ac4ac141ee063d76fafede3b78a8b386fe9f ). Should I change "target" of pull request from "sphinx/master" to "sphinx/2.0" branch? You mean this?

sphinx/pycode/__init__.py
tk0miya: In my experience of OSS development, third party developers usually touches the inside of framework to implement their extensions. I think there are no reason to support such impolite extensions. But dropping old API will cause their users trouble. So I'd like to keep compatibility of APIs as possible.

On such sight, either public or protected is okay to me. I consider both are almost public APIs for hackers...
(If I have to choose one, IMO, I prefer to use public interface. I feel protection model is too much for python. It's like gentlemen's agreements. So it does not protect impolite hackers...)

>And what to do with Tests? Move from test_util to test_pycode file?

Yes, please copy and modify testcode to test_pycode.

tk0miya:
>Should I change "target" of pull request from "sphinx/master" to "sphinx/2.0" branch? You mean this?

Yes, I changed the target just before.

sphinx/pycode/__init__.py
kpnr: Done :-) Can you take a look to `sphinxcontrib-htmlhelp` pull request? It seems not-only-cp1251 problem - python `cpXXXX` codepages are not recognised by hh.exe See also https://www.iana.org/assignments/character-sets/character-sets.xhtml ,please.

tk0miya: Take a look later (maybe in new year)

tk0miya (approved):

tk0miya:
Merged. Thank you for working!

PR #6908, opened 2019-12-09
Fix #6900: sphinx-build: Allow to pass boolean value via ``-D`` option

### LINKED ISSUE #6900: Not possible to set  boolean `latex_use_xindy` via `-D` option to sphinx-build

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #6900

### DIFF
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -217,6 +217,12 @@ class Config:
             defvalue = self.values[name][0]
             if self.values[name][2] == Any:
                 return value
+            elif type(defvalue) is bool or self.values[name][2] == [bool]:
+                if value == '0':
+                    # given falsy string from command line option
+                    return False
+                else:
+                    return bool(value)
             elif isinstance(defvalue, dict):
                 raise ValueError(__('cannot override dictionary config setting %r, '
                                     'ignoring (use %r to set individual elements)') %
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -119,6 +119,20 @@ def test_overrides():
     assert config.value8 == ['abc', 'def', 'ghi']
 
 
+def test_overrides_boolean():
+    config = Config({}, {'value1': '1',
+                         'value2': '0',
+                         'value3': '0'})
+    config.add('value1', None, 'env', [bool])
+    config.add('value2', None, 'env', [bool])
+    config.add('value3', True, 'env', ())
+    config.init_values()
+
+    assert config.value1 is True
+    assert config.value2 is False
+    assert config.value3 is False
+
+
 @mock.patch("sphinx.config.logger")
 def test_errors_warnings(logger, tempdir):
     # test the error for syntax errors in the config file
--- a/CHANGES
+++ b/CHANGES
@@ -87,6 +87,8 @@ Bugs fixed
 * #6876: LaTeX: multi-line display of authors on title page has ragged edges
 * #6887: Sphinx crashes with docutils-0.16b0
 * #6920: sphinx-build: A console message is wrongly highlighted
+* #6900: sphinx-build: ``-D`` option does not considers ``0`` and ``1`` as a
+  boolean value
 
 Testing
 --------
--- a/sphinx/builders/latex/__init__.py
+++ b/sphinx/builders/latex/__init__.py
@@ -464,7 +464,7 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value('latex_logo', None, None, [str])
     app.add_config_value('latex_appendices', [], None)
     app.add_config_value('latex_use_latex_multicolumn', False, None)
-    app.add_config_value('latex_use_xindy', default_latex_use_xindy, None)
+    app.add_config_value('latex_use_xindy', default_latex_use_xindy, None, [bool])
     app.add_config_value('latex_toplevel_sectioning', None, None,
                          ENUM(None, 'part', 'chapter', 'section'))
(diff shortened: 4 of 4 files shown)


sphinx/config.py
```diff
@@ -217,6 +217,12 @@ def convert_overrides(self, name, value):
             defvalue = self.values[name][0]
             if self.values[name][2] == Any:
                 return value
+            elif self.values[name][2] == [bool]:
+                if value in ('0', 'false', 'False'):
```
jfbu: I am a bit uneasy about this. Why not then allow `'no'` and `'off'` ? And our docs of ``-D`` would need update because it suggests  only `0` and `1` from command line (becoming when argsv is parsed `'0'` then `'1'`). Why would not `smartquotes` or `latex_use_latex_multicolumn` benefit from such extra leniency? On the other hand I have myself make the mistake to recommend using `=False` from command line... thus I can appreciate more freedom in input form but I would prefer only strict Python notation `True` and `False` in addition to `0` and `1` then. Or is there some reason you allowed `false`?

jfbu (commented):
Looks to me better than my #6905 (which I will close) of course but I feel some nits are there...

tk0miya:
I have to add test for Config.

jfbu:
Can you rebase this to 2.0 branch?

tk0miya:
oh, I have wrong target branch. changed.

sphinx/config.py
tk0miya: Okay, let's start with small patterns. I'll remove `false` from the list.

>Why would not smartquotes or latex_use_latex_multicolumn benefit from such extra leniency?

This patch only modifies `latex_use_xindy`. Of course we can modify other configurations later.

tk0miya: I found a way to support for config_value which having boolean as a default value, without any change. With this change,  `smartquotes` and `latex_use_latex_multicolumn` will accepts `-D` overrides automatically.

tests/test_config.py
```diff
@@ -119,6 +119,26 @@ def test_overrides():
...
+def test_overrides_boolean():
+    config = Config({}, {'value1': 'True',
+                         'value2': 'False',
+                         'value3': '1',
+                         'value4': '0',
+                         'value5': 'False'})
+    config.add('value1', None, 'env', [bool])
+    config.add('value2', None, 'env', [bool])
+    config.add('value3', None, 'env', [bool])
+    config.add('value4', None, 'env', [bool])
```
tk0miya: We need to add `[bool]` to 4th parameter; `types` as a hint for type of the variable. It is required when it is hard to guess the type of default value.

sphinx/config.py
```diff
@@ -217,6 +217,12 @@ def convert_overrides(self, name, value):
             defvalue = self.values[name][0]
             if self.values[name][2] == Any:
                 return value
+            elif type(self.values[name][0]) is bool or self.values[name][2] == [bool]:
```
jfbu: perhaps use `defvalue` here in place of `self.values[name][0]`

sphinx/config.py
```diff
@@ -217,6 +217,12 @@ def convert_overrides(self, name, value):
...
+                if value in ('0', 'False'):
```
jfbu: shouldn't our document mention  for bool config values one can now use `=True` or `=False` on command line?

jfbu (approved):
LGTM with nits!

sphinx/config.py
tk0miya: I found our document has already defined this behavior.
>For boolean values, use 0 or 1 as the value.
>https://www.sphinx-doc.org/en/master/man/sphinx-build.html#id2

I think handling `0` is a bugfix. And handling `False` is adding a new feature. I don't understand we should support `False` value also. So I drop it from this PR and post another PR if needed. Do you have any opinion for `False` value?

sphinx/config.py
tk0miya: fixed. thanks!

sphinx/config.py
jfbu: I don't have a strong opinion for `False` value. All I know is that I spontaneously advised user to set option via `=False` then only later recalled one should use `=0`. I think for time being you made right choice to not add this new feature.

jfbu (approved):
LGTM!

sphinx/config.py
tk0miya: Okay, it would be better to follow "Zen of python". So let's go without `False` value. And let's discuss again when somebody requests.

>There should be one-- and preferably only one --obvious way to do it.

PR #6916, opened 2019-12-12
WIP: Add inline syntax highlighting using docutils :code: role

### LINKED ISSUE #5157: Inline code is not syntax highlighted
### Problem

It is not possible to apply syntax highlighting to inline literals or code.

The docutils documentation provides a way to do this using the `role` directive to define a new role, based on the `code` role, but the resulting HTML contains classes like `.keyword` instead of classes like `.highlight .k` which are used in Pygments CSS.

#### Expected results

Sphinx to provide a `code` role that applies proper classes for Pygments to style.

There is [an extension to add this](https://github.com/sphinx-contrib/inlinesyntaxhighlight), but it is from 2013 (so maybe out of date with recent Sphinx), and it does not contain code to generate highlighting in LaTeX.

```rst
.. role:: js(code)
    :language: js

This is inline code :js:`function a() { return 1; }`
```

### Environment info

- Sphinx version: 1.7.5

jessetan:
Subject: support inline syntax highlighting in HTML and LaTeX without extensions

### Feature or Bugfix
- Feature

### Purpose
This PR adds the ability to do syntax highlighting of inline text using the docutils :code: role.

Changes:

- Change default of docutils setting that controls the type of Pygments classes that are added (long names or short). Anyone adding custom CSS 
- Add a CSS class to enable highlighting
- Changes LaTeX output to use `\PYG` for inline code

Things this PR does not do (yet):

- Passing inline code to PygmentsBridge for adding classes. Classes are added by docutils instead. This can cause a small difference in what classes are added, but I could not find a nice way to do this.
- A toggle switch to enable/disable this. It is now always on for :code:
- Verification with a wide array of input
- Documentation (will need to be added if approach is OK)
- Automated tests (will need to be added if approach is OK)

### Relates
- Fixes #5157

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -46,6 +46,7 @@ logger = logging.getLogger(__name__)
 default_settings = {
     'embed_stylesheet': False,
     'cloak_email_addresses': True,
+    'syntax_highlight': 'short',
     'pep_base_url': 'https://www.python.org/dev/peps/',
     'pep_references': None,
     'rfc_base_url': 'https://tools.ietf.org/html/',
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -464,7 +464,6 @@ class LaTeXTranslator(SphinxTranslator):
         self.in_footnote = 0
         self.in_caption = 0
         self.in_term = 0
-        self.in_code = 0
         self.needs_linetrimming = 0
         self.in_minipage = 0
         self.no_latex_floats = 0
@@ -2045,8 +2044,6 @@ class LaTeXTranslator(SphinxTranslator):
     def visit_literal(self, node):
         # type: (nodes.Element) -> None
         self.no_contractions += 1
-        if('code' in node.get('classes', [])):
-            self.in_code = 1
         if self.in_title:
             self.body.append(r'\sphinxstyleliteralintitle{\sphinxupquote{')
         else:
@@ -2055,7 +2052,6 @@ class LaTeXTranslator(SphinxTranslator):
     def depart_literal(self, node):
         # type: (nodes.Element) -> None
         self.no_contractions -= 1
-        self.in_code = 0
         self.body.append('}}')
 
     def visit_footnote_reference(self, node):
@@ -2298,7 +2294,7 @@ class LaTeXTranslator(SphinxTranslator):
         elif classes in [['accelerator']]:
             self.body.append(r'\sphinxaccelerator{')
             self.context.append('}')
-        elif classes and self.in_code:
+        elif classes and node.get('code', False):
             self.body.append(r'\PYG{%s}{' % ','.join(classes))
             self.context.append('}')
         elif classes and not self.in_title:
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -464,6 +464,7 @@ class LaTeXTranslator(SphinxTranslator):
         self.in_footnote = 0
         self.in_caption = 0
         self.in_term = 0
+        self.in_code = 0
         self.needs_linetrimming = 0
         self.in_minipage = 0
         self.no_latex_floats = 0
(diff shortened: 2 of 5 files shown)


sphinx/environment/__init__.py
```diff
@@ -46,6 +46,7 @@
 default_settings = {
     'embed_stylesheet': False,
     'cloak_email_addresses': True,
+    'syntax_highlight': 'short',
```
tk0miya, line 49: Note: this option lets pygments making short option. It is same as Sphinx uses inside.
```
-<p>This is inline code <code class="code js docutils literal notranslate highlight"><span class="keyword declaration"><span class="pre">function</span></span> <span class="name other"><span class="pre">a</span></span><span class="punctuation"><span class="pre">()</span></span> <span class="punctuation"><span class="pre">{</span></span> <span class="keyword"><span class="pre">return</span></span> <span class="literal number integer"><span class="pre">1</span></span><span class="punctuation"><span class="pre">;</span></span> <span class="punctuation"><span class="pre">}</span></span></code></p>
+<p>This is inline code <code class="code js docutils literal notranslate highlight"><span class="kd"><span class="pre">function</span></span> <span class="nx"><span class="pre">a</span></span><span class="p"><span class="pre">()</span></span> <span class="p"><span class="pre">{</span></span> <span class="k"><span class="pre">return</span></span> <span class="mi"><span class="pre">1</span></span><span class="p"><span class="pre">;</span></span> <span class="p"><span class="pre">}</span></span></code></p>
```

sphinx/writers/latex.py
```diff
@@ -2294,6 +2298,9 @@ def visit_inline(self, node):
         elif classes in [['accelerator']]:
             self.body.append(r'\sphinxaccelerator{')
             self.context.append('}')
+        elif classes and self.in_code:
```
tk0miya: I think it is better to use Transform module to propagation the inline nodes under literal node instead of remember its status.
```
class InlineCodeTransform(SphinxTransform):
    default_priority = 200

    def apply(self):
        for node in self.document.traverse(nodes.literal):
            for subnode in node:
                if isinstance(subnode, nodes.inline):
                    # mark inline nodes just under literal node as "code" to handle them in writer!
                    subnode['code'] = True
```

This makes writer simple. We have some transform modules for LaTeX writer in `sphinx.builders.latex.transforms`.

tk0miya (commented):
I love this implementation. Let's go with this! I'm waiting for update :-)

tk0miya:
>Passing inline code to PygmentsBridge for adding classes. Classes are added by docutils instead. This can cause a small difference in what classes are added, but I could not find a nice way to do this.

I'll also check this. It would be nice if both are same.

>A toggle switch to enable/disable this. It is now always on for :code:

AFAIK, there are no way to disable it for code-blocks. So I feel it is not needed.

jessetan:
> > A toggle switch to enable/disable this. It is now always on for :code:
> 
> AFAIK, there are no way to disable it for code-blocks. So I feel it is not needed.

You are correct that there is no option to disable it per `:code:` role. I meant an entry in `conf.py` to disable it document-wide.

tk0miya:
What I meant is that we also don't have such option for code-blocks document-wide. I consider it is not requested.

sphinx/writers/latex.py
jessetan: I've update the PR with this code.

jessetan (commented):
The [inline syntax highlighting extension](https://github.com/sphinx-contrib/inlinesyntaxhighlight/blob/master/sphinxcontrib/inlinesyntaxhighlight.py) actually does use the Pygments bridge, and then applies a regex to remove the unnecessary `<div>` that Pygments wraps it in.
Is that any better than using the lexed code from docutils?

tk0miya:
Oh, I don't know that extension. It seems better to me in HTML case. It seems HtmlFormatter has "nowrap" option not to generate div and pre tag. It would be nice for this case.
https://pygments.org/docs/formatters/#HtmlFormatter

But, on the other hand, there are no such option for LaTeXFormatter... So it is hard to use this implementation to sphinx-core.

tk0miya:
I posted a PR for adding `nowrap` option to LatexFormatter to pygments. It will help us if merged.
https://github.com/pygments/pygments/pull/1343

jessetan:
Thanks @tk0miya, I will change the implementation to use PygmentsBridge

PR #6917, opened 2019-12-13
Fix documentation typo on the directives page

omad:
### Feature or Bugfix
- Docs Bugfix

### Purpose
Fix that the `emphasize-lines` option is mistyped as `emphasized-lines` on the documentation page.


jfbu:
Thanks for catching this!

PR #6925, opened 2019-12-15
Remove redundant type="text/javascript" from <script> elements

jdufresne:
In HTML5, `<script>` elements default to MIME type text/javascript. The HTML5 living standard and MDN recommend against including the attribute.

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type

> The HTML5 specification urges authors to omit the attribute rather than provide a redundant MIME type.

From https://html.spec.whatwg.org/#the-script-element

> Authors should omit the type attribute instead of redundantly setting it.

### DIFF
--- a/sphinx/builders/html.py
+++ b/sphinx/builders/html.py
@@ -146,7 +146,6 @@ class JavaScript(str):
         self = str.__new__(cls, filename)  # type: ignore
         self.filename = filename
         self.attributes = attributes
-        self.attributes.setdefault('type', 'text/javascript')
 
         return self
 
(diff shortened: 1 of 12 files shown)


tk0miya (commented):
LGTM. Could you rebase this into 2.0 branch? Then I'll merge this soon.

jdufresne:
Rebased :slightly_smiling_face:

sphinx/builders/html.py
```diff
@@ -146,8 +146,6 @@ def __new__(cls, filename: str, **attributes: str) -> "JavaScript":
         self = str.__new__(cls, filename)  # type: ignore
         self.filename = filename
         self.attributes = attributes
-        self.attributes.setdefault('type', 'text/javascript')
-
```
tk0miya, line 150: trivial: This empty line is not related with this change. Please keep it as is.

tk0miya (changes requested):
LGTM with nits.

jdufresne:
Restored blank line.

tk0miya (approved):
Thank you for quick fix!

tk0miya:
Merged.

Issue #6352, opened 2019-05-11
Autosection Label Extension does not work


Daltz333:
I have the [following project and branch](https://github.com/Daltz333/frc-docs/tree/feature/autosection-label)

In this project I have the 'sphinx.ext.autosectionlabel' extension installed, see [this](https://github.com/Daltz333/frc-docs/blob/feature/autosection-label/source/conf.py).

I have a RST file called [ip-networking](https://github.com/Daltz333/frc-docs/blob/feature/autosection-label/source/docs/networking/ip-networking.rst) and [networking-basics](https://github.com/Daltz333/frc-docs/blob/feature/autosection-label/source/docs/networking/networking-basics.rst). Whenever I try to reference ip-networking in networking-basics, I am given the following warning with no link generated. 

```
C:\Users\Dalton Smith\Documents\frc-docs\source\docs\networking\networking-basics.rst:64: WARNING: undefined label: ip-networking (if the link has no caption the label must precede a section header)
```

I am referencing it using the following syntax
```
See the :ref:`IP Networking Article <ip-networking>` for more information.
```

I don't know if this is improper project configuration, or a bug. Thanks :)

I'll be happy to provide any additional information.

**Development Environment:** 
**OS:** Windows 10 1903
**Sphinx Version:** 2.0.1
**Python Version:** 3.7.3

tk0miya:
Please use sectioning title for reference (ex. ```:ref:`IP Networking` ```). It seems `ip-networking` and `networking-basics` are not title of your document.

If you want to create a link to the top of "document", please use `:doc:` role instead.
https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-doc

Daltz333:
The docs stated that it prefixes the filename to the front. So would it not be `ip-networking:sectiontitle`? @tk0miya

tk0miya:
By default, the prefix is not needed. It is only needed when you enables `autosectionlabel_prefix_document` option in your `conf.py`.

Daltz333:
The option is indeed enabled. That is why I am confused on the syntax.

tk0miya:
On my local, ```:ref:`ip-networking:IP Networking` ``` works fine with this settings:
```
autosectionlabel_prefix_document = True
```

Could you check this please?

Daltz333:
Ohh I see. You use the document file-name only when referencing a section specifically, otherwise its the title of the document. That was quite confusing and annoying to find out. Thanks.

[...]

PR #6933, opened 2019-12-20
Clarify docs on using autosectionlabel prefixes with directories

poikilotherm:
Subject: Clarify docs on using autosectionlabel prefixes with directories

### Bugfix
- Bugfix (documentation)

### Purpose
- People are confused on the usage of `sphinx.ext.autosectionlabel` with files below root directories. The docs should provide a simple example to make this more obvious to use. See also #6352 or https://stackoverflow.com/questions/51161726

### Detail
- See #6352. This is a very simple doc fix.
- I did not add myself to AUTHORS or CHANGES, as this is a very minor doc bug fix.

### Relates
- #6352

### DIFF
--- a/doc/usage/extensions/autosectionlabel.rst
+++ b/doc/usage/extensions/autosectionlabel.rst
@@ -39,6 +39,13 @@ Configuration
    avoiding ambiguity when the same section heading appears in different
    documents.
 
+   When ``index.rst`` doesn't reside at your docs root directory, the prefix
+   has to contain any directories relative to it.
+
+   Example: if ``index.rst`` lives at ``example/hello/world/index.rst``, use
+   ``:ref:`example/hello/world/index:Introduction```. (Carefull: no leading
+   ``/`` as necessary with ``:doc:``!)
+
 .. confval:: autosectionlabel_maxdepth
 
    If set, autosectionlabel chooses the sections for labeling by its depth. For


doc/usage/extensions/autosectionlabel.rst
```diff
@@ -39,6 +39,13 @@ Configuration
    avoiding ambiguity when the same section heading appears in different
    documents.
 
+   When ``index.rst`` doesn't reside at your docs root directory, the prefix
+   has to contain any directories relative to it.
```
tk0miya, line 43: "document name" is already defined in glossary. It's common word for sphinx users, I believe. So what we should do is add `:term:` role to above description.
https://www.sphinx-doc.org/en/master/glossary.html#term-document-name

doc/usage/extensions/autosectionlabel.rst
```diff
@@ -39,6 +39,13 @@ Configuration
...
+
+   Example: if ``index.rst`` lives at ``example/hello/world/index.rst``, use
+   ``:ref:`example/hello/world/index:Introduction```. (Carefull: no leading
+   ``/`` as necessary with ``:doc:``!)
```
tk0miya, line 47: `:doc:` role can accept docname in relative and absolute form. It is a special case of "document name". I think it is not good example...

Issue #6656, opened 2019-08-18
Quickstart no longer asks for autodoc


nralbers:
**Describe the bug**
According to the quickstart documentation, quickstart will ask if you want to run autodoc. However, the latest version no longer does this.

**To Reproduce**
```
sphinx-quickstart

Welcome to the Sphinx 2.1.2 quickstart utility.

Please enter values for the following settings (just press Enter to
accept a default value, if one is given in brackets).

Selected root path: .

You have two options for placing the build directory for Sphinx output.
Either, you use a directory "_build" within the root path, or you separate
"source" and "build" directories within the root path.
> Separate source and build directories (y/n) [n]: y

The project name will occur in several places in the built documentation.
> Project name: Structlog Extensions
> Author name(s): Niels Albers
> Project release []: 0.0.3

If the documents are to be written in a language other than English,
you can select a language here by its language code. Sphinx will then
translate text that it generates into that language.

For a list of supported codes, see
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language.
> Project language [en]: 

Creating file ./source/conf.py.


```

**Expected behavior**
Either: ask for autodoc and configure, as stated in the documentation.
Or: Update the documentation to conform to the new behaviour, and document
how to configure autodoc in the quick start documentation

**Environment info**
- OS: Linux
- 3.6.0
- Sphinx version: 2.1.2
- Sphinx extensions:  sphinx.ext.autodoc

tk0miya:
Thank you for reporting.
Since 2.0, quickstart does not ask for extensions (see #4148). So our docs are wrong for this case.

tapaswenipathak:
Hi folks: I will PRing.

nerdenator:
I am looking through Google to find a tutorial to learn how to use Sphinx to document a project. Pretty much every single one of them recommends using sphinx-quickstart, and does so using the old set of questions. To me, this is a major, breaking change that is preventing me from using sphinx. I don't suppose there's any way to add in the old questions using a command line flag or some other optional route?

JacksonWuxs:
I am wondering what is the change between V2.0+ and V1.0+. If the updates are not significant, should I use V1.0+ to fix this problem?

domdog:
I ran into the same problem.

Would like a command line flag to have access to the old behaviour,  so commonly mentioned across the web, or a clear document that describes:

    -what is completely setup as a default now.  There is nothing to do with these and they are working in the present setup.
[...]

Issue #5618, opened 2018-11-11
Provide `sphinx` command and integrate sphinx commands


shimizukawa:
### Proposal

I propose to integrate multiple commands into ONE `sphinx` command and provide existing commands as subcommands.

<details>
<summary>In Japanese</summary>
Subject: sphinx コマンドを提供し、sphinxの各コマンドを統合する

私は、複数あるコマンドを1つのsphinxコマンドに統合し、既存のコマンドをサブコマンドとして提供することを提案します。
</details>

#### Background

We have several sphinx related commands as:

- make (by Makefile, make.bat)
- sphinx-quickstart
- sphinx-build
- sphinx-apidoc
- sphinx-autogen

`Makefile` and` make.bat` are provided to use the make command as a thin wrapper for `sphinx-build`. The make command is very flexible and easy to understand for people familiar with make. However, it was unfortunate for some unfamiliar persons to hinder proper reporting when something goes wrong (#1693, #2521), or not to work in some shell environments (#3145). We are also receiving proposals to make `Makefile`/`make.bat` highly functional which wants to make such an unkind part more kind, more convenient (e3ae24f8, #3741). At the same time, there is also a proposal to discard the make command and replace it with a make command that can be written in Python (#3196).

It is the purpose of this proposal to finish these situations and make it a simple world.
For this reason, I propose the following changes to the following.

1. take over existing command features by introducing `sphinx` command.
2. Optionalization of `Makefile` /` make.bat` generating on quickstart.
3. Stepwise deprecate and remove the existing `sphinx- *` commands
4. Enhancement of `sphinx` command (Provide subcommand plugin feature etc)

<details>
<summary>In Japanese</summary>
`Makefile` と `make.bat` は `sphinx-build` の薄いラッパーとしてmakeコマンドを使えるように提供されています。makeコマンドはmakeを使い慣れている人にとってはとても柔軟で分かりやすいものです。しかし、不慣れな人にとってはなにか問題が発生した場合に適切にレポートする妨げとなったり (#1693, #2521)、一部のシェル環境では動作しないといった不親切な面もありました(#3145)。こういった不親切な部分をより親切にしたい、より便利にしたいというMakefile/make.batを高機能化する提案もいただいています(e3ae24f8, #3741)。同時に、makeコマンドを捨ててPythonで書けるmake系コマンドに置き換える提案もあります (#3196)。

こういった状況を終わらせ、シンプルな世界にするのがこの提案の目的です。
[...]

PR #6938, opened 2019-12-21
Add unified 'sphinx' command (WIP!)

stephenfin:
### Feature or Bugfix
- Feature

### Purpose
Rather than having multiple separate executables - `sphinx-build`,
`sphinx-quickstart`, etc. - add a single `sphinx` command with multiple
subcommands.

### Detail
TODO.

### Relates
- N/A

### DIFF
--- a/sphinx/cmd/base.py
+++ b/sphinx/cmd/base.py
@@ -1,0 +1,103 @@
+"""
+    sphinx.cmd.base
+    ~~~~~~~~~~~~~~~
+
+    Base command class to be used elsewhere.
+
+    :copyright: Copyright 2007-2019 by the Sphinx team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+import abc
+import argparse
+import inspect
+import sys
+from typing import Dict, List, Type
+
+import pkg_resources
+
+from sphinx import __display_version__
+from sphinx.util.console import color_terminal, nocolor
+
+
+class SphinxApplication:
+
+    namespace = 'sphinx.commands'
+    epilog = 'For more information, visit <http://sphinx-doc.org/>.'
+
+    def __init__(self) -> None:
+        self.commands = {}  # type: Dict[str, Type[SphinxCommand]]
+        self._load_commands()
+
+    def _load_commands(self) -> None:
+        for entry_point in pkg_resources.iter_entry_points(self.namespace):
+            self.commands[entry_point.name] = entry_point.load()
+
+    def get_parser(self) -> argparse.ArgumentParser:
+        parser = argparse.ArgumentParser(
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            epilog=self.epilog)
+
+        parser.add_argument('--version', action='version', dest='show_version',
+                            version='%%(prog)s %s' % __display_version__)
+
+        subparsers = parser.add_subparsers(title='subcommands')
+
+        for name, command in self.commands.items():
+            obj = command()
+
+            description = None
+            summary = None
+            if command.__doc__:
+                description = inspect.cleandoc(command.__doc__)
+                summary = description.partition('\n\n')[0]
+
+            command_parser = subparsers.add_parser(
+                name=name,
+                help=summary,
+                description=description,
+                epilog=self.epilog,
+                formatter_class=argparse.RawDescriptionHelpFormatter
+            )
+            obj.add_parser(command_parser)
+            command_parser.set_defaults(run=obj.run)
+
+        return parser
+
+    def run(self, argv: List[str]) -> int:
+        if not color_terminal():
+            nocolor()
+
+        parser = self.get_parser()
+
+        try:
+            args = parser.parse_args(argv)
+        except SystemExit as err:
+            return err.code
+
+        return args.run(args)
+
+
+class SphinxCommand(abc.ABC):
+
+    @abc.abstractmethod
+    def add_parser(self, parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
+        """Generate the subparser for the command."""
+        pass
+
+    @abc.abstractmethod
+    def run(self, args: argparse.Namespace) -> int:
+        """Main entry point for the command.
+
+        Run the command including any subcommands.
+        """
+        pass
+
+
+def main(argv=None):
+    argv = argv or sys.argv[1:]
+    return SphinxApplication().run(argv)
+
+
+if __name__ == '__main__':
+    main()
--- a/sphinx/cmd/build.py
+++ b/sphinx/cmd/build.py
@@ -94,6 +94,13 @@ def jobs_argument(value: str) -> int:
             return jobs
 
 
+def exists_path(path: str) -> str:
+    """Validate that file path is a valid one."""
+    if not os.path.exists(path):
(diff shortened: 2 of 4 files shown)


stephenfin:
@tk0miya I've only ported two of the four commands currently available in tree as part of this PR. Before I work on the remainder, I have a few questions I'd like your input on.

1. Does this idea make sense and are you happy with the approach?
2. I'm not planning to migrate makemode because I don't like it and don't think it's necessary as noted at https://github.com/sphinx-doc/sphinx/pull/4354#issuecomment-357428922. Do you agree?

stephenfin:
As noted at https://github.com/sphinx-doc/sphinx/pull/6939#issuecomment-568220441, another thing we could do here in change the behavior of the outputdir and maybe the sourcedir options for `sphinx build`. We could transition from:

    sphinx-build sourcedir outputdir file_a file_b ...

to

    sphinx build sourcedir file_a file_b ...

This would mean we'd need to be able to define `outdir` somewhere else. I'm suggesting `conf.py` (and later `sphinx.ini`, when I finally finish that work). We could provide a vanity option for people that really didn't want to do this, of course:

    sphinx build -o outdir sourcedir file_a file_b

tk0miya:
>Does this idea make sense and are you happy with the approach?

Yes, of course! I just read a part of this patch yet. There might be trivial comments and requests later. But  it looks good to me. I'll continue to read this later (maybe my new years task!).

>I'm not planning to migrate makemode because I don't like it and don't think it's necessary as noted at #4354 (comment). Do you agree?

+1. new sphinx command should not follow historical interfaces. We can forget both make and non-make mode!

sphinx/cmd/base.py
```diff
@@ -0,0 +1,103 @@
...
+                description=description,
+                epilog=self.epilog,
+                formatter_class=argparse.RawDescriptionHelpFormatter
+            )
+            obj.add_parser(command_parser)
+            command_parser.set_defaults(run=obj.run)
+
+        return parser
+
+    def run(self, argv: List[str]) -> int:
```
tk0miya, line 67: Please call initialization of locales at beginning of the application.
```
    sphinx.locale.setlocale(locale.LC_ALL, '')
    sphinx.locale.init_console(os.path.join(package_dir, 'locale'), 'sphinx')
```

stephenfin:
Good to hear. I am holding off on this until I determine whether I can configure the output directory as an optional flag (so `sphinx build -o outdir srcdir` instead of `sphinx build srcdir outdir`), as discussed at https://github.com/sphinx-doc/sphinx/pull/6939#issuecomment-568271054. Other people have suggested this idea, such as https://github.com/sphinx-doc/sphinx/issues/5618#issuecomment-500596337 though I disagree with `-c conf.py` being almost required. I'd prefer to default `srcdir` to the current directory.

sphinx/cmd/build.py
```diff
@@ -94,190 +97,216 @@ def jobs_argument(value: str) -> int:
...
-        try:
-            key, val = val.split('=')
-        except ValueError:
-            parser.error(__('-A option argument must be in the form name=value'))
-        try:
-            val = int(val)
-        except ValueError:
-            pass
-        confoverrides['html_context.%s' % key] = val
+    return BuildCommand._run(args)
```
tk0miya, line 303: We need to discuss `sphinx build` (new stye) is same as `sphinx-build` (old style). I still don't have an answer.

sphinx/cmd/quickstart.py
```diff
@@ -464,87 +465,146 @@ def valid_dir(d: Dict) -> bool:
...
-        description=description)
+class InitCommand(SphinxCommand):
+    """Generate required files for a Sphinx project.
 
-    parser.add_argument('-q', '--quiet', action='store_true', dest='quiet',
-                        default=None,
-                        help=__('quiet mode'))
-    parser.add_argument('--version', action='version', dest='show_version',
-                        version='%%(prog)s %s' % __display_version__)
+    'sphinx init' is an interactive tool that asks some questions about your
```
tk0miya, line 471: In some tools, `init` subcommand goes project initialization silently. So I wonder we should separate quickstart feature into `sphinx init` and `sphinx quickstart` (?) (or `sphinx init --interactive`).

sphinx/cmd/build.py
stephenfin: The only behavior change I've suggested so far is to go from:

    sphinx-build sourcedir builddir [file, ...]

to

    sphinx build sourcedir [file, ...] [-o builddir]

But perhaps there are other options that we think are badly named and should be renamed now? I hadn't thought of that so good idea :+1:

sphinx/cmd/quickstart.py
stephenfin: Are you thinking of `git init`? I think it's okay for `sphinx init` to be interactive, so long as you can specify these options from the command line and avoid the interactive prompts. We do need some way to retrieve required information like `project` and `author`.

sphinx/cmd/base.py
stephenfin: Ack

sphinx/cmd/build.py
tk0miya: It's just an idea, `sphinx build html` is so so good to me.

sphinx/cmd/quickstart.py
tk0miya: Yes, I was thinking of `git init` and `bundle init`. But I found `npm init` asks questions interactively.  So please forget my comment.

>We do need some way to retrieve required information like project and author.

Actually, `project` and `author` are not required configuration. So we can suppress all of questions.

stephenfin: > Actually, `project` and `author` are not required configuration. So we can suppress all of questions.

Oh, I didn't know that. Good to know. However, is it really useful to ignore these? What advantage would `sphinx init` provide over `touch conf.py` if we didn't ask for these? I think it makes sense to ask for these at a minimum, even if people can skip past them.

tk0miya: It seems some users would like to do interactions more verbosely (see #6656). I still don't have an answer for it. But, as one of idea, it might be better to provide both silent and verbose way.

Note: silent means only "less". It does not equal to no question.

stephenfin: Sorry for the delay. How about we make interactive *not the default* by adding an `--interactive` flag? With this flag, we would request *all* configuration options in an interactive manner from the user, but users that don't want to do this can simply pass the relevant args they care about, e.g. `--project`? That would satisfy the request of #6656 while not annoying people like you that just wants to start a really quick project? The other alternative is a combination of `--quick` (default) and `--full` arguments, with the latter enabling configuration of extensions etc. What do you think?

tk0miya: I prefer an option like `--full`. It is too much to choose non interactive mode by default to me. But it is annoying to me to list all configurations. I think it is difficult to choose to appropriate options or extensions for standard users. So it would be better to ask minimum configurations by default.

AA-Turner:
Closing this, we intend to have a unified `sphinx` command through the `_cli` sub-package, which is slowly progressing. Thanks to Stephen for originally proposing a design for this feature!

A

PR #6944, opened 2019-12-22
Modify release process

stephenfin:
The current release process is rather confusing, with no less than three
branches in active use at any given time. Given we already have a well
documented deprecation policy that allows us to transition users from
one API to another over the course of a couple of releases, there
doesn't appear to be any reason to plan the next MAJOR release far ahead
of time. Instead, we can copy what Django do and simply release these on
a roughly annual basis.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -317,13 +317,13 @@ Debugging Tips
 Branch Model
 ------------
 
-Sphinx project uses following branches for developing that conforms to Semantic
-Versioning 2.0.0 (refs: https://semver.org/ ).
+Sphinx project uses following branches for developing that conforms to
+`Semantic Versioning 2.0.0 <https://semver.org/>`__.
 
 ``master``
-    Development for MAJOR version.
+    Development for next version.
     All changes including incompatible behaviors and public API updates are
-    allowed.
+    allowed, assuming the :ref:`deprecation-policy` is followed.
 
 ``X.Y``
     Where ``X.Y`` is the ``MAJOR.MINOR`` release.  Used to maintain current
@@ -334,15 +334,18 @@ Versioning 2.0.0 (refs: https://semver.org/ ).
     new MAJOR version is released, the old ``MAJOR.MINOR`` branch will be
     deleted and replaced by an equivalent tag.
 
-``X.Y.Z``
-    Where ``X.Y.Z`` is the ``MAJOR.MINOR.PATCH`` release.  Only
-    backwards-compatible bug fixes are allowed. In Sphinx project, PATCH
-    version is used for urgent bug fix.
+A new MAJOR release can be expected every year, while MINOR releases are made
+roughly every three months. For example:
 
-    ``MAJOR.MINOR.PATCH`` branch will be branched from the ``v`` prefixed
-    release tag (ex. make 2.3.1 that branched from v2.3.0) when a urgent
-    release is needed. When new PATCH version is released, the branch will be
-    deleted and replaced by an equivalent tag (ex. v2.3.1).
+.. csv-table:: Release Schedule
+   :header: "Release Series", "Release Date"
+   :align: center
+
+   "2.0", "March 2019"
+   "2.1", "June 2019"
+   "2.2", "August 2019"
+   "2.3", "December 2019"
+   "3.0", "March 2020"
 
 
 Deprecating a feature


stephenfin:
This is a request for comments. Personally, having not had the chance to work on Sphinx for a few months, I've found the current release process quite jarring. Having to consider two upcoming releases at once seems like unnecessary work. It would seem more sane to just worry about removing deprecated APIs on a yearly basis and leave them until then.

jfbu:
I think what causes problems to people is that our non-API-breaking branch is called `2.0` (I guess for reasons of our  www.sphinx-doc.org using ReadTheDocs). As a result many PRs are done on master branch without people realizing this is for distant release.

Apart from that I do not quite agree the branches are complex. Much less so than in projects handling backports of fixes. We have three branches:

- the one for urgent bug fixes C

- the one for say quaterly releases B

- the one for long term horizon (once a year I think) A

but C always end up merged into B and B always gets merged into A. Simply we can merge in A features which breaks expected behaviour and are not allowed in B. But all we add to B ends up in A. When A is released, B will start from that anew. When B is released, C starts from it anew.

stephenfin:
> I think what causes problems to people is that our non-API-breaking branch is called `2.0` (I guess for reasons of our [www.sphinx-doc.org](http://www.sphinx-doc.org) using ReadTheDocs). As a result many PRs are done on master branch without people realizing this is for distant release.
> 
> Apart from that I do not quite agree the branches are complex. Much less so than in projects handling backports of fixes. We have three branches:
> 
> * the one for urgent bug fixes C
> 
> * the one for say quaterly releases B
> 
> * the one for long term horizon (once a year I think) A
> 
> 
> but C always end up merged into B and B always gets merged into A. Simply we can merge in A features which breaks expected behaviour and are not allowed in B. But all we add to B ends up in A. When A is released, B will start from that anew. When B is released, C starts from it anew.

That's a better synopsis than I had. Nice work :smile: So my question is this: do we need the A branch? Instead of the A branch, could we use the B branch for all releases and just designate every Nth release as a new MAJOR release? This is what projects like Django do and it helps ensure people keep the deprecation policy in mind.

jfbu:
My synopsis forgot to mention only "A" has stable name i.e. it is "master" branch. "B" has stable name only in-between two releases of "A". Currently it is called "2.0" but prepares for "2.4.0" release. It will be deleted and recreated as "3.0" when Sphinx 3.0 is released (from master branch). And "C" name changes at each hotfix release. Currently it is at "2.3.2".

The A branch is thus MAJOR release "2.0", "3.0". I think we do need it because it is allowed to add breaking changes. The B branch (currently "2.0") is not allowed to add breaking changes compared to last MAJOR release.

I think there is a constraint in naming the branches which is related to how RTD presents our documentation, and validity of links to it. In the past our current "2.0" was called "stable". When Sphinx 2.0.0 was released, all 1.x maintenance stopped at 1.8.5. I am not sure if the plan at "3.0" release is similarly not maintain "2.x" series beyond the last hotfix release. INaming the branches respectively "development", "releases", "hotfix", would confuse people expecting "development" to be merged into "releases" whereas it is exactly the opposite.  Thus A branch (aka "master") could be renamed rather better than "developement" into "deprecating" or "breaking" or perhaps "longterm" or "distanthorizon", or "futures" say rather "future" to avoid financial disaster :smile:. We could thus have "future", "stable", "hotfix".

jfbu:
I personally always found confusing that RTD defaults to master branch which in fact is "not yet released". It would make much better sense if user by default would be directed to documentation of last release (currently 2.0 branch, preparing 2.4.0 release).

This appears important now that sphinx-quickstart produces a lean conf.py... and people have to get documentation otherwise than from comments in the conf.py

stephenfin:
> My synopsis forgot to mention only "A" has stable name i.e. it is "master" branch. "B" has stable name only in-between two releases of "A". Currently it is called "2.0" but prepares for "2.4.0" release. It will be deleted and recreated as "3.0" when Sphinx 3.0 is released (from master branch). And "C" name changes at each hotfix release. Currently it is at "2.3.2".

Understood. I'm suggesting we merge the "2.0" and "master" branch. I think the only releases we should care about at the moment are 2.4 (the next release) and 2.3.2 (the next bugfix).

> The A branch is thus MAJOR release "2.0", "3.0". I think we do need it because it is allowed to add breaking changes. The B branch (currently "2.0") is not allowed to add breaking changes compared to last MAJOR release.

and here, I'm suggesting you should never allow breaking changes without a deprecation cycle. At a fixed cadence (every 4th release, or once a year), we would decide enough deprecations have built up that we're going to remove the deprecated features and release a new MAJOR version.

> I think there is a constraint in naming the branches which is related to how RTD presents our documentation, and validity of links to it. In the past our current "2.0" was called "stable". When Sphinx 2.0.0 was released, all 1.x maintenance stopped at 1.8.5. I am not sure if the plan at "3.0" release is similarly not maintain "2.x" series beyond the last hotfix release. INaming the branches respectively "development", "releases", "hotfix", would confuse people expecting "development" to be merged into "releases" whereas it is exactly the opposite. Thus A branch (aka "master") could be renamed rather better than "developement" into "deprecating" or "breaking" or perhaps "longterm" or "distanthorizon", or "futures" say rather "future" to avoid financial disaster :smile:. We could thus have "future", "stable", "hotfix".

Orrr, we could do the above and merge the "master" and "2.0" branches. We could then add a flag to the top of "https://sphinx.readthedocs.io/en/latest/" to say this is the development version and you should go to "https://sphinx.readthedocs.io/en/2.3/" like the Django docs do. If this doesn't exist yet, I'd be happy to work on it.

stephenfin:
To summarize, in case I'm still not clear, I want to merge `2.0` back into `master` and stop having two future release branches because I don't see the point of `master` as it currently exists

jfbu:
> To summarize, in case I'm still not clear, I want to merge `2.0` back into `master` and stop having two future release branches because I don't see the point of `master` as it currently exists

Do you mean you would like current 2.0 being called "master"? But we do need a branch for PRs introducing breaking changes. We want to be able to review, accept, and merge them into _some_ branch. And in our development model, when we are ready (once a year?) to issue a new MAJOR release, it will be that branch (which regularly gets updated with the developments in the stable one).

stephenfin:
> > To summarize, in case I'm still not clear, I want to merge `2.0` back into `master` and stop having two future release branches because I don't see the point of `master` as it currently exists
> 
> Do you mean you would like current 2.0 being called "master"?

Yes, exactly.

> But we do need a branch for PRs introducing breaking changes. We want to be able to review, accept, and merge them into _some_ branch.

That's my point. We shouldn't allow a breaking change without a deprecation period, right? As such, I think we reject an PRs that don't respect that deprecation policy. Basically, if you want to remove this feature that was deprecated in 2.x, you have to wait until the last 3.x release has been released before doing so (since our deprecation policy says we wait until {N+2}.0 to drop things deprecated in N.x)

> And in our development model, when we are ready (once a year?) to issue a new MAJOR release, it will be that branch (which regularly gets updated with the developments in the stable one).

As above, I'm suggesting we don't need a branch for this. Instead, we just know that these things aren't allowed until we've released the last MINOR release in a MAJOR series. This seems far clearer to me, anyway :smile:

jfbu:
I think there is a difference to be made between breaking change (meaning output may look different) and API deprecation. Our current master branch is careful to follow a careful deprecation policy. For example `html_smartypants` (or however it was called) was deprecated but still worked. At the same time Smart Quotes applied also to LaTeX builder which they didn't before. Thus we both have a here a breaking change (output is not the same) and a deprecation (a config variable is still there but marked as being removed in future) and all of this was done in then master branch.

stephenfin:
> I think there is a difference to be made between breaking change (meaning output may look different) and API deprecation. Our current master branch is careful to follow a careful deprecation policy. For example `html_smartypants` (or however it was called) was deprecated but still worked. At the same time Smart Quotes applied also to LaTeX builder which they didn't before. Thus we both have a here a breaking change (output is not the same) and a deprecation (a config variable is still there but marked as being removed in future) and all of this was done in then master branch.

I think that's okay though. The breaking changes I think most users would care about are about are for removed features ("I used to do this thing and no longer can because X"). I don't think output looking a bit different is an issue unless it breaks workflows but that case would suggest a bug, no?

Maybe I should try another tack. Are there other notable projects that operate under this model? I'm not aware of any but that doesn't mean they don't exist, of course

jfbu:
@stephenfin I think our CHANGES file gives a better idea of deprecation policy in particular the 3.0 section for future release contains:

```
Incompatible changes
--------------------

* Drop features and APIs deprecated in 1.8.x
```
 and a bit below in 2.4.0 section we see

```
Release 2.4.0 (in development)
==============================

Dependencies
------------

Incompatible changes
--------------------

Deprecated
----------

* ``sphinx.environment.BuildEnvironment.indexentries``
* ``sphinx.environment.collectors.indexentries.IndexEntriesCollector``
* ``sphinx.io.FiletypeNotFoundError``
* ``sphinx.io.get_filetype()``

```

meaning that these bits of API get deprecated at a MINOR release but their actual removal is responsability of MAJOR release, currently that will be 4.0 release, and there is a summary at [extdev/deprecated.rst](https://github.com/sphinx-doc/sphinx/blob/master/doc/extdev/deprecated.rst)

stephenfin:
I understand, and I'm saying we don't need to worry about 4.0 yet. Let's designate every Nth release a MAJOR release instead of a MINOR release, and accept deprecation removal patches only during **those** release cycles. So `master` will be used for `2.4.0` development now and then `3.0.0` once that's released. Does that make sense?

tk0miya:
+1 for moving 2.0 branch to master (called as B above)
-1 for dropping a branch for future release (called as A)

I understand our branch model is not easily understandable. Especially we goes to add new features to stable branch if not breaking anything. But many contributors does not know our branch model well. So they have usually post PRs to master branch. As a result, I've mentioned for authors of PR to rebase 2.0 branch many time. It's sad situation. So I agree to rename it to master. It's better to understand than now. I also feel good for renaming to `2.x`.

On the other hand, there are some breaking changes except dropping APIs. We've added some breaking changes into stable branch if minor. But major ones should be released as major release. For example, changing default settings or behavior is major breaking change. I know it is (very) less to commit into master branch now. But it is surely needed.

CONTRIBUTING.rst
```diff
@@ -334,15 +334,18 @@ Versioning 2.0.0 (refs: https://semver.org/ ).
     new MAJOR version is released, the old ``MAJOR.MINOR`` branch will be
     deleted and replaced by an equivalent tag.
 
-``X.Y.Z``
-    Where ``X.Y.Z`` is the ``MAJOR.MINOR.PATCH`` release.  Only
-    backwards-compatible bug fixes are allowed. In Sphinx project, PATCH
-    version is used for urgent bug fix.
+A new MAJOR release can be expected every year, while MINOR releases are made
+roughly every three months. For example:
```
tk0miya, line 338: FYI: We've try to release in every other months. So 2.4.0 will be released in February. And 3.0 will be April.

stephenfin:
> +1 for moving 2.0 branch to master (called as B above)
> -1 for dropping a branch for future release (called as A)
> 
> I understand our branch model is not easily understandable. Especially we goes to add new features to stable branch if not breaking anything. But many contributors does not know our branch model well. So they have usually post PRs to master branch. As a result, I've mentioned for authors of PR to rebase 2.0 branch many time. It's sad situation. So I agree to rename it to master. It's better to understand than now. I also feel good for renaming to `2.x`.

Excellent.

> On the other hand, there are some breaking changes except dropping APIs. We've added some breaking changes into stable branch if minor. But major ones should be released as major release. For example, changing default settings or behavior is major breaking change. I know it is (very) less to commit into master branch now. But it is surely needed.

My question is this: could we say "thank you for the change which we will merge at [future date]"? We won't be releasing from master until the next time there is a release, is there any difference to letting this unreleased code sit on a branch for N months vs. leaving the pull request unmerged for N months? I think these kind of changes are so rare that it would be okay to keep a list of them somewhere and then just merge all of them at once before we release a major version.

CONTRIBUTING.rst
stephenfin: You're saying the release cycle is every two months and not every three months? I can fix this if so.

tk0miya: Yes, please.

tk0miya:
I hesitate to keep PRs unmerged for N months. After time passes, merging becomes hard. Who resolves the conflicts? It's terrible to me. In addition, it makes resolved issues and PRs opening. It's stressful to me.

I don't still understand why would you like to reduce branch. I think main concern will go away with using master branch for stable release.

stephenfin:
That's a good point, actually. I still don't think it's necessary but I can happily ignore it if I want :smile: I'm away this week but I'll respin the week after

tk0miya:
I noticed github supports to specify default branch of PRs. That's good for this purpose. So I vote to set up `2.0` as a default branch instead of changing branch models.

tk0miya:
>I noticed github supports to specify default branch of PRs. That's good for this purpose. So I vote to set up 2.0 as a default branch instead of changing branch models.

FYI: Now I changed the default branch of Sphinx.

tk0miya:
I posted a minimal change of branch model on as #7121.

tk0miya:
Now we choose #7121 instead of this. Closing.

PR #6945, opened 2019-12-22
Expand use of path2doc, combine 'compile_xxx' and 'build_xxx' functions and move file validation into builder

stephenfin:
### Feature or Bugfix
- Refactoring

### Purpose
We provide a `path2doc` helper to convert filenames to their equivalent "docnames". However, we do this translation manually in a few places. Stop doing that.

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -267,30 +267,31 @@ class Builder:
 
     def build_specific(self, filenames: List[str]) -> None:
         """Only rebuild as much as needed for changes in the *filenames*."""
-        # bring the filenames to the canonical format, that is,
-        # relative to the source directory and without source_suffix.
-        dirlen = len(self.srcdir) + 1
-        to_write = []
-        suffixes: Tuple[str] = tuple(self.config.source_suffix)  # type: ignore
+        docnames: List[str] = []
+
         for filename in filenames:
             filename = path.normpath(path.abspath(filename))
+
             if not filename.startswith(self.srcdir):
                 logger.warning(__('file %r given on command line is not under the '
                                   'source directory, ignoring'), filename)
                 continue
+
             if not path.isfile(filename):
                 logger.warning(__('file %r given on command line does not exist, '
                                   'ignoring'), filename)
                 continue
-            filename = filename[dirlen:]
-            for suffix in suffixes:
-                if filename.endswith(suffix):
-                    filename = filename[:-len(suffix)]
-                    break
-            filename = filename.replace(path.sep, SEP)
-            to_write.append(filename)
-        self.build(to_write, method='specific',
-                   summary=__('%d source files given on command line') % len(to_write))
+
+            docname = self.env.path2doc(filename)
+            if not docname:
+                logger.warning(__('file %r given on command line is not a valid '
+                                  'document, ignoring'), filename)
+                continue
+
+            docnames.append(docname)
+
+        self.build(docnames, method='specific',
+                   summary=__('%d source files given on command line') % len(docnames))
 
     def build_update(self) -> None:
         """Only rebuild what was changed or added since last build."""
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -252,6 +252,7 @@ class Builder:
         message = __('targets for %d po files that are specified') % len(catalogs)
         self.compile_catalogs(catalogs, message)
 
+    # TODO(stephenfin): This would make more sense as 'compile_outdated_catalogs'
     def compile_update_catalogs(self) -> None:
         repo = CatalogRepository(self.srcdir, self.config.locale_dirs,
                                  self.config.language, self.config.source_encoding)
@@ -263,6 +264,8 @@ class Builder:
 
     def build_all(self) -> None:
         """Build all source files."""
+        self.compile_all_catalogs()
+
         self.build(None, summary=__('all source files'), method='all')
 
     def build_specific(self, filenames: List[str]) -> None:
@@ -290,11 +293,15 @@ class Builder:
 
             docnames.append(docname)
 
+        self.compile_specific_catalogs(filenames)
+
         self.build(docnames, method='specific',
                    summary=__('%d source files given on command line') % len(docnames))
 
     def build_update(self) -> None:
         """Only rebuild what was changed or added since last build."""
+        self.compile_update_catalogs()
+
         to_build = self.get_outdated_docs()
         if isinstance(to_build, str):
             self.build(['__all__'], to_build)
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -275,14 +275,14 @@ class Builder:
         for filename in filenames:
             filename = path.normpath(path.abspath(filename))
 
+            if not path.isfile(filename):
+                logger.warning(__('file %r given on command line does not exist, '),
+                               filename)
+                continue
+
             if not filename.startswith(self.srcdir):
                 logger.warning(__('file %r given on command line is not under the '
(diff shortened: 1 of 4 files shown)


sphinx/builders/__init__.py
```diff
@@ -257,38 +259,50 @@ def compile_update_catalogs(self) -> None:
...
-                logger.warning(__('file %r given on command line is not under the '
-                                  'source directory, ignoring'), filename)
+                logger.debug(__('file %r given on command line is not under the '
+                                'source directory, ignoring'), filename)
                 continue
-            if not path.isfile(filename):
-                logger.warning(__('file %r given on command line does not exist, '
-                                  'ignoring'), filename)
+
+            docname = self.env.path2doc(filename)
```
tk0miya, line 288: `path2doc()` is provided by project module now. And it is not related with `env` module now.
So I think following steps are better to me:

1. add `builder.project` attribute on instantiation
2. using `self.project.path2doc()` instead here

sphinx/builders/__init__.py
```diff
@@ -257,38 +259,50 @@ def compile_update_catalogs(self) -> None:
...
-            filename = filename.replace(path.sep, SEP)
-            to_write.append(filename)
-        self.build(to_write, method='specific',
-                   summary=__('%d source files given on command line') % len(to_write))
+
+            docnames.append(docname)
+
+        if missing_files:
+            raise ApplicationError('files were not found in %r: %r' % (
+                self.srcdir, missing_files))
```
tk0miya: Why stopping application here? I think this is not a critical error. So it is better to emit warnings.

sphinx/builders/__init__.py
```diff
@@ -257,38 +259,50 @@ def compile_update_catalogs(self) -> None:
...
+                logger.debug(__('file %r given on command line is not under the '
+                                'source directory, ignoring'), filename)
                 continue
-            if not path.isfile(filename):
-                logger.warning(__('file %r given on command line does not exist, '
-                                  'ignoring'), filename)
+
+            docname = self.env.path2doc(filename)
+            if not docname:
+                logger.debug(__('file %r given on command line is not a valid '
```
tk0miya: Does this debug messsage helps developers? ApplicationError already describes the error (if we'll throw it)

sphinx/builders/__init__.py
```diff
@@ -246,6 +247,7 @@ def to_domain(fpath: str) -> str:
         message = __('targets for %d po files that are specified') % len(catalogs)
         self.compile_catalogs(catalogs, message)
 
+    # TODO(stephenfin): This would make more sense as 'compile_outdated_catalogs'
```
tk0miya, line 255: Do you mean renaming? If so, +1 (in another PR).

sphinx/builders/__init__.py
```diff
@@ -257,38 +259,50 @@ def compile_update_catalogs(self) -> None:
 
     def build_all(self) -> None:
         """Build all source files."""
+        self.compile_all_catalogs()
```
tk0miya, line 267: Why do you move these methods? I suppose these change is not related with "expand use of path2doc".

tk0miya: BTW, I think separate message catalog building process to individual class is better. It would works better if we'll do multiple builds at once running.

tk0miya (changes requested):
Please choose correct title or separate PR. This contains two different changes.
+1 for usage of path2doc with nits.
But another one needs more review and design.

sphinx/builders/__init__.py
tk0miya: As a trial, I made a example of above comment. With improving this, Sphinx might be able to build multiple output at once in future.
https://github.com/sphinx-doc/sphinx/compare/2.0...tk0miya:multiple_builders?expand=1

sphinx/builders/__init__.py
stephenfin: > I think this is not a critical error.

Are you sure? The user has asked us to build for specific documents, but those documents do not exist. I think this means there is a user error and we should fail. I consider this similar to `make foo bar baz` where `foo` and `baz` are valid targets but `bar` is not.

sphinx/builders/__init__.py
stephenfin: Probably not. I was retaining previous behaviour, though things are different now that we have ApplicationError. I can remove this or the ApplicationError. It's your call.

sphinx/builders/__init__.py
stephenfin: Can I do that in a follow-up, since that involves more work? `env.path2doc` does exist and isn't deprecated so I think it's okay to use it.

stephenfin:
> Please choose correct title or separate PR. This contains two different changes.
> +1 for usage of path2doc with nits.
> But another one needs more review and design.

Done. As usual, I'm too used to Gerrit and have submitted multiple separate commits (which you can see [here](https://github.com/sphinx-doc/sphinx/pull/6945/commits)) as one PR. I can submit these as separate PRs if I _really_ have to :smile:

sphinx/builders/__init__.py
stephenfin: > Why do you move these methods? I suppose these change is not related with "expand use of path2doc".

As noted [here](https://github.com/sphinx-doc/sphinx/pull/6945/commits/3eb6589637dab6fccbc4adc80bf3487700315353)

>  builder: Combine 'compile_xxx' and 'build_xxx' functions
> 
> These are only called from the one location and we already use 'build'
> to do both building and writing. Given that we intend to simplify the
> handling of the 'filenames' argument, this makes our lives there much
> easier.

It looks like there is good reason not to do this though, so I can revert if you want?

sphinx/builders/__init__.py
stephenfin: Yup

sphinx/builders/__init__.py
tk0miya: +0: I don't think it's not hard. But either is okay to me.

sphinx/builders/__init__.py
tk0miya: I feel using warning is better in this case. +1 for reverting this to warnings.

sphinx/builders/__init__.py
tk0miya: I feel we need to design whole of builders and its interface carefully. I still don't know my rough code is good idea or not. So we should discuss this in another PR. +1 for reverting.

sphinx/builders/__init__.py
tk0miya: So far, Sphinx raises an error only when it can't continue the processing. It keep processing when minor error like `.. image:: not_found.png`. And I consider this case is not critical because Sphinx can continue builds.
BTW, user can promote the warnings to error with `-W` option.

AA-Turner:
Rebased.

A

AA-Turner:
I changed the failure on non-existing documents to warnings so I can merge, personally I sympathise with @stephenfin's logic that these should be hard failures but I think better if that change was a distinct PR (with CHANGES entry) as opposed to this internal refactoring PR.

A

AA-Turner:
Thanks @stephenfin (3 years late!)

A

stephenfin:
> Thanks @stephenfin (3 years late!)

Thank **you** :smile:

stephenfin:
> I changed the failure on non-existing documents to warnings so I can merge, personally I sympathise with @stephenfin's logic that these should be hard failures but I think better if that change was a distinct PR (with CHANGES entry) as opposed to this internal refactoring PR.

...and yes, this makes sense. I'll try to get a separate PR up (if you or someone else doesn't beat me to it)

PR #6946, opened 2019-12-23
Doc rework (user guide, part 7)

stephenfin:
This should be the final change required for the user guide. In this series,
we split the builders guide into multiple docs, which allows us to move the
HTML and LaTeX-specific information into their own guides. The templating
and theming docs are kept separate since these are more developer than end-user
focused.


jfbu:
I will comment purely on LaTeX: this makes its documentation location hard to understand. Why should the `latex_elements` documentation of all its keys now be found under Builders>Built-in builders>LaTeX builders>class sphinx.builders.latex.LaTeXBuilder>Configuration ? problem is that most of it originally was in Configuration>Options for LaTeX output but it grew too big hence got moved to a special section LaTeX customization. Now this section is moved to description of LaTeX builder. But in fact most of it has nothing to do with Sphinx LaTeX build, as it governs the actual PDF output which is produced from pdflatex/xelatex/lualatex/platex/uplatex.This is the case of ``'papersize'``, ``'pointsize'`` etc to start with them, basically all of `The latex_elements configuration setting`.

edit: I mean by that that these options most of the time change nothing to actual process of LaTeX build from Sphinx, except that the config is handed over via mark-up in the the tex file to the next stage which is actual processing (via `make all-pdf` in the latex build repertory) by the LaTeX toolchain leading to actual PDF. Hence it appears counter-intuitive to place the documentation of all relevant options at the place chosen in this PR. Logical place would be in `Options for LaTeX output`. At least it would be more visible (for those poor souls like me who actually do build Sphinx doc regularly to PDF to check how it looks like...)

tk0miya:
Could you separate PR please? It is hard to read the PR contains both modifying and moving. I know we can read each individual change via commit log. But it does not helps merging. I think large PR prevents to merge whole of it. (Personally, I don't have time to read whole of this PR now... It also prevents to review.)

I think discussions might be needed to small changes (ex. bindep). But it might be needed for large change (like moving). So mixing them is hard to review. So small PR helps to merge it. It also helps reviews!

stephenfin:
I can, but I do have this broken up at the moment with separate commits already. Is it possible to review these commits separately? If not, I'll separate each commit into a pull request.

Unrelated: I find the GitHub pull request workflow quite poor 😞 Gerrit is much easier. Alas, it's all we have to work with.

stephenfin:
> I will comment purely on LaTeX: this makes its documentation location hard to understand. Why should the `latex_elements` documentation of all its keys now be found under Builders>Built-in builders>LaTeX builders>class sphinx.builders.latex.LaTeXBuilder>Configuration ? problem is that most of it originally was in Configuration>Options for LaTeX output but it grew too big hence got moved to a special section LaTeX customization. Now this section is moved to description of LaTeX builder. But in fact most of it has nothing to do with Sphinx LaTeX build, as it governs the actual PDF output which is produced from pdflatex/xelatex/lualatex/platex/uplatex.This is the case of `'papersize'`, `'pointsize'` etc to start with them, basically all of `The latex_elements configuration setting`.
>
> edit: I mean by that that these options most of the time change nothing to actual process of LaTeX build from Sphinx, except that the config is handed over via mark-up in the the tex file to the next stage which is actual processing (via `make all-pdf` in the latex build repertory) by the LaTeX toolchain leading to actual PDF.

But the output `.tex` files are modified, yes? As such, it is the `latex` builder that's affected. You could make the exact same argument for the `html` builder. To paraphrase, the config is handed over via markup in the html files to the next stage which is actual processing by the web browser leading to the actual website. Right?

> Hence it appears counter-intuitive to place the documentation of all relevant options at the place chosen in this PR. Logical place would be in `Options for LaTeX output`. At least it would be more visible (for those poor souls like me who actually do build Sphinx doc regularly to PDF to check how it looks like...)

So you're suggesting I move this documentation back into the config guide? I'm okay with this. I could simply link to that.

stephenfin:
@jfbu On an unrelated note, I've been thinking it might make sense to deprecate the `latex_elements` and `sphinxsetup` options in favour of multiple separate options like `latex_element_papersize`, `latex_element_pointsize`, etc. I think this would be easier to document and configure and would better map to how the other builders work, since none of those seem to use dict-style options. Is there a reason that wasn't done from day one?

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
-        - In order to support occasional Cyrillic (физика частиц)
-          or Greek letters (Σωματιδιακή φυσική) in
-          a document whose language is English or a  Latin European
-          one, the default set-up is enhanced (only for ``'pdflatex'``
-          engine) to do:
+Keys that you may want to override include:
+
+``'papersize'``
+   Paper size option of the document class (``'a4paper'`` or
+   ``'letterpaper'``)
```
tk0miya: a4paper and letterpaper are one of examples. So it would be better to add "ex." to represent that.

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
+
+   :Default: ``'letterpaper'``
+
+``'pointsize'``
+   Point size option of the document class (``'10pt'``, ``'11pt'`` or
+   ``'12pt'``)
```
tk0miya: This is also one of examples.

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
-        Support for individual Greek and Cyrillic letters:
-
-        - In order to support occasional Cyrillic (физика частиц)
-          or Greek letters (Σωματιδιακή φυσική) in
-          a document whose language is English or a  Latin European
-          one, the default set-up is enhanced (only for ``'pdflatex'``
-          engine) to do:
+Keys that you may want to override include:
+
+``'papersize'``
```
tk0miya: I think it would be better to add a new directive like `.. latex-element`. Then these are also displayed in index page: https://www.sphinx-doc.org/en/master/genindex.html

It would be useful to me. What do you think?

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
+   For ``'xelatex'`` and ``'lualatex'``, the default is to use the FreeFont
+   family: this OpenType font family supports both Cyrillic and Greek scripts
+   and is available as separate Ubuntu Xenial package ``fonts-freefont-otf``.
+   It is not necessary to install the much larger ``texlive-fonts-extra``
+   package.
+
+   ``'platex'`` (Japanese documents) engine supports individual Cyrillic and
+   Greek letters with no need of extra user set-up.
+
+   :Default: ``'\\usepackage{times}'`` (or ``''`` when using a Cyrillic script)
```
tk0miya: This is not correct. `fontpkg` is changed by `latex_engine` and `language`.

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
+
+   .. versionchanged:: 2.0
+      Added support for individual Greek and Cyrillic letters:
+
+``'fncychap'``
+   Inclusion of the "fncychap" package (which makes fancy chapter titles),
+   default ``'\\usepackage[Bjarne]{fncychap}'`` for English documentation
+   (this option is slightly customized by Sphinx),
+   ``'\\usepackage[Sonny]{fncychap}'`` for internationalized docs (because
+   the "Bjarne" style uses numbers spelled out in English).  Other
```
tk0miya: It seems duplicated with `:Default:` field. How about removing them from description?

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
+   "Bjornstrup".  You can also set this to ``''`` to disable fncychap.
+
+   :Default: ``'\\usepackage[Bjarne]{fncychap}'`` for English documents,
+       ``'\\usepackage[Sonny]{fncychap}'`` for internationalized documents, and
+       ``''`` for Japanese documents.
+
+``'preamble'``
+   Additional preamble content.  One may move all needed macros into some file
+   :file:`mystyle.tex.txt` of the project source repertory, and get LaTeX to
+   import it at run time::
```
tk0miya: I think second sentence is a tip to make preamble simple. So it would be better to separate from main description of preamble field. cc: @jfbu

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
+   .. code-block:: python
+
+      latex_additional_files = ["mystyle.sty"]
+
+   :Default: ``''``
+
+``'figure_align'``
+   Latex figure float alignment. Whenever an image doesn't fit into the current
+   page, it will be 'floated' into the next page but may be preceded by any
+   other text.  If you don't like this behavior, use 'H' which will disable
```
tk0miya: 'H' should be a literal.

doc/latex.rst
```diff
@@ -65,433 +49,487 @@ avoid interpretation as escape sequences, or use raw strings (as is done here).
...
 
-        - The key setting is silently *ignored* if ``\usepackage{enumitem}``
-          is executed inside the document preamble. Use then rather the
-          dedicated commands of this LaTeX package.
+   With ``'platex'`` (Japanese),  ``'xelatex'`` or ``'lualatex'``, this
+   key is ignored.
 
-     .. versionadded:: 1.5
+   :Default: ``'\\usepackage{textalpha}'`` or ``''`` if ``fontenc`` does not
+       include the ``LGR`` option.
```
tk0miya: Addition to this, the empty setting is used for non-pdflatex engines. see `ADDITIONAL_SETTINGS`.

doc/latex.rst
```diff
@@ -506,46 +544,40 @@ It defaults to empty.  If non-empty, it will be passed as argument to the
...
-   at the start of the chapter and::
+   latex_elements = {
+       'passoptionstopackages': r'\PassOptionsToPackage{svgnames}{xcolor}',
+   }
 
-     .. raw:: latex
+It is possible to insert further uses of the ``\sphinxsetup`` LaTeX macro
+directly into the body of the document, via the help of the :rst:dir:`raw`
+directive. This chapter is styled in the PDF output using the following at the
+start of the chaper::
```
tk0miya: chap *t* er?

doc/latex.rst
```diff
@@ -506,46 +544,40 @@ It defaults to empty.  If non-empty, it will be passed as argument to the
...
-              hintBorderColor={named}{LightCoral},
-              attentionborder=3pt,
-              attentionBorderColor={named}{Crimson},
-              attentionBgColor={named}{FloralWhite},
-              noteborder=2pt,
-              noteBorderColor={named}{Olive},
-              cautionborder=3pt,
-              cautionBorderColor={named}{Cyan},
-              cautionBgColor={named}{LightCyan}}
+The colors used in the above are provided by the ``svgnames`` option of the
```
tk0miya: "above" is not appropriate after changed.

doc/latex.rst
```diff
@@ -506,46 +544,40 @@ It defaults to empty.  If non-empty, it will be passed as argument to the
    \usepackage{sphinx}
    \sphinxsetup{key1=value1, key2=value2,...}
 
-.. versionadded:: 1.5
-
-.. hint::
```
tk0miya, line 511: Why did you remove `hint` directive? The "raw" use case is a tip for `\sphinxsetup` macro in LaTeX level. But this section describes about the `latex_elements['sphinxsetup']` key first. It is confusable.

doc/latex.rst
```diff
@@ -579,55 +612,68 @@ macros may be significant.
...
+
+    :Default: ``r``
 
     .. versionadded:: 1.7
 
 ``parsedliteralwraps``
-    default ``true``. Tells whether long lines in :dudir:`parsed-literal`\ 's
-    contents should wrap.
+    Boolean to specify if long lines in :dudir:`parsed-literal`\ 's contents
+    should wrap.
```
tk0miya: contents are wrapped? (like `verbatimwrapslines`)

doc/latex.rst
```diff
@@ -579,55 +612,68 @@ macros may be significant.
...
+    :Default: ``true``
 
     .. versionadded:: 1.5.2
        set this option value to ``false`` to recover former behaviour.
 
 ``inlineliteralwraps``
-    default ``true``. Allows linebreaks inside inline literals: but extra
-    potential break-points (additionally to those allowed by LaTeX at spaces
-    or for hyphenation) are currently inserted only after the characters
+    Boolean to specify is linebreaks are allowed inside inline literals: but
```
tk0miya: Boolean to specify *if* ?

doc/latex.rst
```diff
@@ -652,31 +699,41 @@ macros may be significant.
...
 ``VerbatimColor``
-    default ``{rgb}{1,1,1}``. The background colour for
-    :rst:dir:`code-block`\ s. The default is white.
+    The background colour for :rst:dir:`code-block`\ s.
+
+    :Default: ``{rgb}{1,1,1}`` (white)
 
 ``VerbatimBorderColor``
-    default ``{rgb}{0,0,0}``. The frame color, defaults to black.
+    The frame color.
```
tk0miya: colo *u* r?

doc/latex.rst
```diff
@@ -652,31 +699,41 @@ macros may be significant.
...
+
+    :Default: ``{rgb}{0,0,0}`` (black)
 
 ``VerbatimHighlightColor``
-    default ``{rgb}{0.878,1,1}``. The color for highlighted lines.
+    The color for highlighted lines.
```
tk0miya: colo *u* r?

tk0miya (commented):
LGTM with nits (for this commit).

tk0miya:
I just reviewed first three commits. I don't review other commits. But it seems more discussion is needed. So how about separate this to two PRs?

stephenfin:
Okay, I will split this into two PRs, soon as I find the time :sweat_smile:

Issue #6672, opened 2019-08-26
Remove chapters for custom howto LaTeX classes


sephalon:
**Is your feature request related to a problem? Please describe.**
Since the `article` LaTeX class, which the Sphinx `howto` class is based upon, does not support chapters they are removed according to the following control structure in `sphinx/writers/latex.py`:

        # we assume LaTeX class provides \chapter command except in case
        # of non-Japanese 'howto' case
        self.sectionnames = LATEXSECTIONNAMES[:]
        if self.settings.docclass == 'howto':
            docclass = self.config.latex_docclass.get('howto', 'article')
            if docclass[0] == 'j':  # Japanese class...
                pass
            else:
                self.sectionnames.remove('chapter')
        else:
            docclass = self.config.latex_docclass.get('manual', 'report')
        self.elements['docclass'] = docclass

We use custom LaTeX classes that are based on (i.e. `\LoadClass` them) the default Sphinx `howto` and `manual` classes. However, it is not possible to trigger the removal of chapters with a custom class name. The option `latex_toplevel_sectioning` only works on a global level and does not allow us to mix `howto`/`manual`-style classes in the `latex_documents` list anymore.

**Describe the solution you'd like**
An easy solution would be to replace the line

        if self.settings.docclass == 'howto':

by

        if 'howto' in self.settings.docclass:

enabling the removal of chapters for any class containing the keyword `howto`.

**Describe alternatives you've considered**
Have not come up with any other solution so far.

jfbu:
Do I understand correctly that you use something like `'sephalonhowto'` as docclass, and that you have a LaTeX custom class file `sephalonhowto.cls` (added to [`latex_additionalfiles`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-latex_additional_files)) which will execute `\LoadClass{sphinxhowto}`? (this is not something expected...)

+1 for your proposal, reasonable.

Alternative I see is to have in your project source a custom file `sphinxhowto.cls` which would rather than use `\LoadClass` physically incorporate the original official one. Then you would use `'howto'` as docclass.

sephalon:
> Do I understand correctly that you use something like 'sephalonhowto' as docclass, and that you have a LaTeX custom class file sephalonhowto.cls (added to latex_additionalfiles) which will execute \LoadClass{sphinxhowto}? (this is not something expected...)

[...]

PR #6969, opened 2019-12-28
LaTeX Theming support

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add LaTeX Theming feature (experimental)
- In this PR, LaTeX Theme is a collection of settings:
  - document class (ex. article, report, book)
  - wrapper document class (ex. sphinxhowto, sphinxmanual)
  - top level sectioning unit (ex. part, chapter and section)
- As a built-in themes, `'manual'` and `'howto'` has been provided as in the past.
- Users can choose LaTeX themes via `latex_theme` option or 5th element of `latex_documents`.
- Users can define own custom themes via `theme.conf`

  ```
  [theme]
  docclass = article
  wrapperclass = sphinxmanual
  toplevel_sectioning = chapter
  ```

This PR is a draft. More discussion is needed to determine what is LaTeX themes to be merged.

## remaining tasks

- [x] document
- [x] tests


sphinx/builders/latex/theming.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class Theme:
+    """A set of LaTeX configurations."""
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.docclass = name
+        self.wrapperclass = name
+        self.toplevel_sectioning = 'chapter'
```
tk0miya, line 31: I think there are some candidates for theme's configurations:
* papersize
* pointsize
* fncychap
* classoptions (ex. openany, twocolumn)

And users can overrided these theme's settings via `latex_elements` if they want.

sphinx/builders/latex/theming.py
```diff
@@ -0,0 +1,118 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Theme:
+    """A set of LaTeX configurations."""
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self.docclass = name
+        self.wrapperclass = name
```
tk0miya, line 30: I wonder wrapperclass is a part of LaTeX theme. Actually it is used in sphinxhowto.cls and sphinxmanual.cls. But other themes may not follow the implementation, I think. It would be better to separate styles from sphinxhowto.cls and sphinxmanual.cls into indivisual .sty file.

sphinx/builders/latex/__init__.py
```diff
@@ -228,6 +230,7 @@ def write(self, *ignored) -> None:
 
         for entry in self.document_data:
             docname, targetname, title, author, docclass = entry[:5]
+            theme = self.themes.get(docclass)
```
tk0miya: Now we're using themes only for generating .tex files. Addition to this, it should contain additional files (ex. .sty file). So I must write a handler for coping theme-related files.

kryskool:
Hi

Why don't use the entrypoint **sphinx.latex_themes** like [sphinx.html_themes](https://github.com/sphinx-doc/sphinx/blob/af62fa61e6cbd88d0798963211e73e5ba0d55e6d/sphinx/theming.py#L207) ?

Regards

tk0miya:
@kryskool This PR is still in draft. I'd like to discuss what is LaTeX theming first. Do you satisfy this structure?

tk0miya:
@jfbu Could you review this please? I think this is first version of LaTeX themes. It is still in experimental. But it is a seed of theming feature like HTML.

sphinx/builders/latex/theming.py
jfbu: Of course wrapperclass is an heritage of Sphinx own way: it uses a base class such as report or article. Then some document layout provided via Sphinx (title page for example) is configured in a way depending taking into account base class. But in the context of a user document class, it does not make much sense to have same mechanism: either the user will base a LaTeX class on book or report or article or whatever. So the user wishing to theme LaTeX should only have to provide own class and matched template. Possibly the gist of sphinxmanual/sphinxhowto should be moved as you say into .sty files, say sphinxreportstyle.sty and sphinxarticlestyle.sty (thus, we drop manual/howto in naming) and sphinxjsreportstyle.sty etc... A user is the free to  import or not one of those in a custom class.

sphinx/writers/latex.py
```diff
@@ -306,21 +322,32 @@ def __init__(self, document: nodes.document, builder: "LaTeXBuilder") -> None:
...
+            # of non-Japanese 'howto' case
+            if document.get('docclass') == 'howto':
+                docclass = self.config.latex_docclass.get('howto', 'article')
+                if docclass[0] == 'j':  # Japanese class...
+                    pass
+                else:
+                    self.sectionnames.remove('chapter')
+            else:
+                docclass = self.config.latex_docclass.get('manual', 'report')
+            self.elements['docclass'] = docclass
```
jfbu, line 350: relates to #6672

doc/usage/configuration.rst
```diff
@@ -2087,6 +2082,33 @@ These options influence LaTeX output.
...
+.. confval:: latex_theme
+
+   The "theme" that the LaTeX output should use.  It is a collection of settings
+   for LaTeX output (ex. document class, top level sectioning unit and so on).
+
+   As a built-in LaTeX themes, ``manual`` and ``howto`` are bundled.
+
+   ``manual``
+     A LaTeX theme for writing a manual.  It based on ``report`` document class
+     (``jsbook`` is also used for Japanese documents).
```
jfbu: typo: It is based.

It imports the ``report`` document class (Japanese documents use ``jsbook``).

doc/usage/configuration.rst
```diff
@@ -2087,6 +2082,33 @@ These options influence LaTeX output.
...
+
+   ``howto``
+     A LaTeX theme for writing an article.  It based on ``article`` document class
+     (``jreport`` is also used for Japanese documents).  :confval:`latex_appendices`
```
jfbu: It imports the ``article`` document class (Japanese documents use ``jreport`` rather).

doc/usage/configuration.rst
```diff
@@ -2087,6 +2082,33 @@ These options influence LaTeX output.
...
+     is available only for this theme.
```
jfbu, line 2099: is that true? I did not know.

jfbu (approved):
Overall I feel this is good idea. However it may raise too much hope on user side. LaTeX predates HTML, XML, etc..., Python, Jinja,... Despite a widespread misunderstanding, it is not really a mark-up language (it does not construct a tree of the document structure and never has any global view of the document structure). Many things such as refined table layout are very hardly amenable to automatized procedures. Typical example is wrapfig package. Basically each use of it requires manual check of PDF and adjustments. And it conflicts with the "list environments" which are a basic part of the LaTeX document classes. Things such as memoir class are very problematic to us because of potential conflicts with many packages we use explicitly in sphinx.sty; we already have quite a few configurations via latex_elements and sphinxsetup. Some configuration for example of the way notices/admonitions is doable for LaTeX knowledgeable users. But achieving a level where we can do this simply by an interface accessible to people who do not know what a TeX macro is, would require a tremendous amount of work. A first step is doable which is to split sphinx.sty into separate smaller pieces. Users could then replace by their own. For example we could try to move sphinxVerbatim into a level where user can decide to use tcolorbox for rendering.

Of course pointsize etc... are simple things, document class options. For them it is easy and calling that Theming is possible but only a name. Going much beyond that will be confronted with the fact many Sphinx users are not experts in LaTeX. We can not at Sphinx do the job that the LaTeX community has not achieved in 40 years of development. It is impossible to extract a framework which will leave people free to plug-in their own extension of memoir class, or if they prefer to use the Koma-Script classes.

Actually the whole concept of LaTeX document class has always looked fishy to me. There are tons of such classes developed for University Thesis requirement for example, and very often they create more problems than they solve things. The "document class" concept of LaTeX is very disappointing because LaTeX is disappointing in lacking a real philosophy of what the document layout is. It pretends to be a structured mark-up language, but in many aspects it is not consistent (for example table of contents rendering conflict with LaTeX concept of list layout). There are many locations where LaTeX code is but a slim layer on TeX macros and the amount of flexibility is very little. Another example is titlepage layout. I am thus not very optimistic about how much we can achieve apart from perhaps move some configuration into a more Pythonic way.

doc/usage/configuration.rst
tk0miya: Yes, it is expanded only if howto. I don't know why.
https://github.com/sphinx-doc/sphinx/blob/39a191294e615af6465e347b8e31ae2f76d51653/sphinx/builders/latex/__init__.py#L244-L246

doc/usage/configuration.rst
tk0miya: thanks. updated.

doc/usage/configuration.rst
tk0miya: done

sphinx/writers/latex.py
tk0miya: Yes, it's start point of this PR :-)

tk0miya (commented):
Thank you for reviewing. I understand this LaTeX theming is not perfect and it still requires knowledge of both Sphinx and LaTeX deeply. But it allows challengers to customize LaTeX output. So far, LaTeX writer and output are strongly coupled. So it is difficult to customize by developers. I think #6672 is one of the cases. I still don't know what is the goal of LaTeX theming. But helping LaTeX experts is one of them, I believe. To help them, it is also needed to separate sphinx.sty to small pieces.

Anyway, I'm going to merge this into 3.0 and keep improving.

Thank you always!

PR #6970, opened 2019-12-29
index directive and role emit deprecation warnings for python index-types

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- python specific index-types are marked as deprecated in this commit: https://github.com/sphinx-doc/sphinx/commit/3e9182550a (10years ago!)
- This emits deprecation warnings to remove them in future.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,8 @@ Features added
   old stub file
 * #5923: autodoc: ``:inherited-members:`` option takes a name of anchestor class
   not to document inherited members of the class and uppers
+* :rst:dir:`index` and :rst:role:`index` emit deprecation warnings for python
+  specific index-types
 
 Bugs fixed
 ----------
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -852,6 +852,8 @@ mainly contained in information units, such as the language reference.
       creates the entries ``module; hashlib`` and ``hashlib; module``.  (These
       are Python-specific and therefore deprecated.)
 
+      .. deprecated:: 1.0
+
    You can mark up "main" index entries by prefixing them with an exclamation
    mark.  The references to "main" entries are emphasized in the generated
    index.  For example, if two pages contain ::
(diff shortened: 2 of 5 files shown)


shimizukawa:
LGTM for the diff.
My concern is that the official Python documentation uses a lot of `module` and `object`.

* [object](https://github.com/python/cpython/search?q=%22index+object%22)
* [module](https://github.com/python/cpython/search?q=%22index+module%22)

Python official documentation seems to take time to adapt to new writing style like `pair: module; blabla`.

shimizukawa (commented):
LGTM for the diff.
My concern is that the official Python documentation uses a lot of `module` and `object`.

* [object](https://github.com/python/cpython/search?q=%22index+object%22)
* [module](https://github.com/python/cpython/search?q=%22index+module%22)

Python official documentation seems to take time to adapt to new writing style like `pair: module; blabla`.

tk0miya:
@JulienPalard @cocoatomo We'd like to mark some python-specific index-types deprecated obviously (accurately it has already been marked as deprecated in our document). It does not mean they will be removed right away. It will be removed in Sphinx-5.0 (maybe 2 years later). It seems python-doc contains some index entries (see @shimizukawa 's comment). Do you have any opinions for this idea?

JulienPalard:
@tk0miya Thanks for the heads up!

Would a proper fix looks like this https://github.com/python/cpython/pull/18004?

If so, feel free to copy-past my `sed`s to the page documenting this deprecation and how to fix it.

JulienPalard:
@tk0miya have you looked at https://github.com/sphinx-doc/sphinx/pull/6970#issuecomment-574368589?

tk0miya:
Sorry for the late response. I found problems on i18n with this deprecation. The python index-types are translated on index pages automatically via `language` configuration.

For example, `.. index: builtin: len` is translated to the pair of "組み込み関数" and "len" in Japanese document. On the other hand, `.. index:: pair: builtin: len` is not translated. So it is displayed as pair of "builtin" and "len" even in Japanese docs.

We need to resolve this before the deprecation. So I withdraw this proposal for a while.

Sorry for bothering you.

AA-Turner:
Resolved by #11412.

PR #6979, opened 2020-01-01
Close #6830: autodoc: consider a member private if docstring contains :private:

### LINKED ISSUE #6830: A ".. private:: True/False" directive to override whether autodoc/autosummary treat a member as private (regardless of underscores)
Two problems:
- some of the items I'm documenting have internal/private APIs that unfortunately do not start with an underscore (but can't be easily renamed). With sphinx autodoc/autosummary configued to skip private members, I want to indicate that those should be to be skipped as well, but since they have docstrings that are useful for internal purposes I don't want to actually delete the docstring. 
- although I mostly want to skip items starting with _ or __ occasionally I have APIs that are effectively public but do start with underscores (e.g. sometimes a single underscore is used in Python to indicate a "protected" API for use by subclasses). I'd like to be able to selectively enable documentation of these methods on a case-by-case basis from my docstrings. 

A directive a bit like "deprecated" that can be used in a docstring to override whether that member is considered private or not, by autodoc and autosummary.

e.g. 
```
def _protectedMethod(self):
   """ .. private:: False

   This is a great method which is in my public API. 
   """

def semanticallyPrivateMethod(self):
   """ .. private:: True

   This is a great internal method, but shouldn't show up in API docs. 
   """
```

Important detail: needs to work with data/attribute docstrings (i.e. implementation shouldn't rely on __doc__). 

- Could be done via setting undoc-members on parent class/module, but that's really combersome since no locality-of-reference, plus I might want to do a separate sphinx run with private members visible for internal purposes. 

- As a workaround I've been able to mostly get it working for autodoc using autodoc-skip-member (though that doesn't work for autosummary, yet) like this; but it'd be better as a core capability, especially as I have no way to get the docstring Sphinx has extracted for any data/attributes. 
```
def autodoc_skip_member(app, what, name, obj, skip, options):
	if obj.__doc__ and '.. private:: True' in obj.__doc__: return True
```

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #6830 
- autodoc considers a member private if its docstring contains `:private:` field in its info-field-list

  For example:

  ```
  def func(name):
    """blah blah blah

    :param name: description of name
    :type name: str
    :private:
    """
  ```

### DIFF
--- a/doc/extdev/appapi.rst
+++ b/doc/extdev/appapi.rst
@@ -216,6 +216,14 @@ connect handlers to the events.  Example:
 
    .. versionadded:: 0.5
 
+.. event:: object-description-transform (app, domain, objtype, contentnode)
+
+   Emitted when an object description directive has run.  The *domain* and
+   *objtype* arguments are strings indicating object description of the object.
+   And *contentnode* is a content for the object.  It can be modified in-place.
+
+   .. versionadded:: 3.0
+
 .. event:: doctree-read (app, doctree)
 
    Emitted when a doctree has been parsed and read by the environment, and is
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -354,6 +354,9 @@ Info field lists
 ~~~~~~~~~~~~~~~~
 
 .. versionadded:: 0.4
+.. versionchanged:: 3.0
+
+   meta fields are added.
 
 Inside Python object description directives, reST field lists with these fields
 are recognized and formatted nicely:
@@ -367,6 +370,10 @@ are recognized and formatted nicely:
 * ``vartype``: Type of a variable.  Creates a link if possible.
 * ``returns``, ``return``: Description of the return value.
 * ``rtype``: Return type.  Creates a link if possible.
+* ``meta``: Add metadata to description of the python object.  The metadata will
+  not be shown on output document.  For example, ``:meta private:`` indicates
+  the python object is private member.  It is used in
+  :py:mod:`sphinx.ext.autodoc` for filtering members.
 
 .. note::
 
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -29,7 +29,7 @@ from sphinx.pycode import ModuleAnalyzer, PycodeError
 from sphinx.util import inspect
 from sphinx.util import logging
 from sphinx.util import rpartition
-from sphinx.util.docstrings import prepare_docstring
+from sphinx.util.docstrings import extract_metadata, prepare_docstring
 from sphinx.util.inspect import (
     Signature, getdoc, object_description, safe_getattr, safe_getmembers
 )
@@ -560,6 +560,13 @@ class Documenter:
                     doc = None
             has_doc = bool(doc)
 
+            metadata = extract_metadata(doc)
+            if 'private' in metadata:
+                # consider a member private if docstring has "private" metadata
+                isprivate = True
+            else:
+                isprivate = membername.startswith('_')
+
             keep = False
             if want_all and membername.startswith('__') and \
                     membername.endswith('__') and len(membername) > 4:
@@ -575,14 +582,14 @@ class Documenter:
                     if membername in self.options.special_members:
                         keep = has_doc or self.options.undoc_members
             elif (namespace, membername) in attr_docs:
-                if want_all and membername.startswith('_'):
+                if want_all and isprivate:
                     # ignore members whose name starts with _ by default
                     keep = self.options.private_members
                 else:
                     # keep documented attributes
                     keep = True
                 isattr = True
-            elif want_all and membername.startswith('_'):
+            elif want_all and isprivate:
                 # ignore members whose name starts with _ by default
                 keep = self.options.private_members and \
                     (has_doc or self.options.undoc_members)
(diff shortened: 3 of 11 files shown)


doc/extdev/appapi.rst
```diff
@@ -218,6 +218,14 @@ connect handlers to the events.  Example:
 
    .. versionadded:: 0.5
 
+.. event:: object-description-transform (app, domain, objtype, contentnode)
```
tk0miya, line 219: I'm looking for better name for this event. Does anybody have a idea?

doc/extdev/appapi.rst
```diff
@@ -218,6 +218,14 @@ connect handlers to the events.  Example:
...
+
+   Emitted when a object description directive has run.  The *domain* and
```
tk0miya: s/a/an/.

doc/usage/restructuredtext/domains.rst
```diff
@@ -367,6 +370,7 @@ are recognized and formatted nicely:
 * ``vartype``: Type of a variable.  Creates a link if possible.
 * ``returns``, ``return``: Description of the return value.
 * ``rtype``: Return type.  Creates a link if possible.
+* ``private``: Indicates the Python object is private member.
```
tk0miya: We have to mention that private marker will not be shown to output.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1524,6 +1533,21 @@ def merge_autodoc_default_flags(app: Sphinx, config: Config) -> None:
             )
 
 
+def filter_private_field(app: Sphinx, domain: str, objtype: str, content: Element) -> None:
```
tk0miya: This must be moved into python domain. Not a part of autodoc.

doc/usage/restructuredtext/domains.rst
shimizukawa: In my case, I divide the API modules into public and private, and only document the public APIs. Because of this, I would not use private markers to specify exclusions for documentation.
Also, I feel uncomfortable with the specification that it is not documented because it is private. `:undoc:` flag looks good for flags that are not documented.
<details><summary>In Japanese</summary>
私の場合、モジュールを公開用と非公開用に分けて、公開用だけをドキュメント化の対象とします。このため、私はドキュメント化の対象外を指定するためにprivateマーカーを使わないでしょう。
また私は、privateだからドキュメント化しないという仕様に違和感を感じます。ドキュメント化しないフラグなら `:undoc:` という名称がよさそうに思います。
</detail>

tk0miya: Thank you for comment. -0 for `:undoc:`. I feel a bit strange to write an operation to info-field-list. It would be better to write its attribute or behavior. In addition, it is a bit surreal `:undoc:` marker is placed in docstring. It's already documented :-p  It might be a bit better to use `:hidden:` instead.

BTW, according to our policy, private members are not shown by default. They are only shown when `:undoc-members:` option is given. Therefore `:private:` is following that policy.

tk0miya:
Side note: I thought adding `:private:` on the top of docstring at first. But I noticed it conflicts with `autodoc_docstring_signature` feature. So I chose info-field-list to specify its attribute finally.

doc/usage/restructuredtext/domains.rst
shimizukawa: I agreed that `:undoc:` is strange because it's a control command for sphinx behavior. However I think `:hidden:` is also unclear. I feel "what is hidden? hidden from what?"
So, I think name starts with `_` is best way to make it private (or separate private/public module).

tk0miya: Agreed. withdrawing "hidden". Let's think another keyword.

>So, I think name starts with _ is best way to make it private (or separate private/public module).

Personally, I prefer not to use "_" prefix to represent private member. It's mere gentlemen's agreement, not an access control. So I'd like to provide ways to make members private. I know Zen of Python says "There should be one.", but...

tk0miya: I and @shimizukawa discussed about this in face to face and agreed with "private" is enough good.

Addition to this, we also discussed the way to declaration. @shimizukawa worried about it is hard to understand `:private:` keyword is displayed to users or not. In this PR, I thought it is not shown to output document. But it is not obviously for writers of docstring. Now, other keywords are shown on document. So it is not easy to understand the behavior. Finally, we determined to use `:meta private:` instead to describe "private" is a metadata for the python object. We believe `:meta:` field will make info-field-list extendable in future like `:meta name: value`.

doc/extdev/appapi.rst
tk0miya: Note: @shimizukawa says LGTM for this.

doc/extdev/appapi.rst
tk0miya: done

doc/usage/restructuredtext/domains.rst
tk0miya: Note: I noticed new key to info-field-list brings incompatible changes. So it would be better to merge this into master branch.

tk0miya: Reimplemented with review comments!

sphinx/ext/autodoc/__init__.py
tk0miya: moved.

shimizukawa (approved):
LGTM!

tk0miya:
Thank you for reviewing. merged!

PR #6982, opened 2020-01-01
Close #2755: autodoc: Support type_comment style annotation

### LINKED ISSUE #2755: Support for Python 2/3 "straddling" type hint comments
Many libraries will likely maintain Python 2 compatibility for some time yet. It would be fantastic if Sphinx apidoc supported [the Python 2.7-friendly "straddling" syntax for type hints that was added to PEP 484](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code) (i.e. `# type:` comments), for rendering in the same way as Python 3 type hints currently do. Additionally, support for `.pyi` files could also help for similar use cases.

PyCharm supports the `# type:` comments now, so demand might increase from people maintaining Python 2-compatible code.

I'm not at all familiar with Sphinx internals but I could try to work on this with a bit of guidance, perhaps.

Note that [GvR directed me](https://twitter.com/gvanrossum/status/750365065778384896) to [dropbox/typed_ast](https://github.com/dropbox/typed_ast) for getting at the information, urging not to try to use mypy as a library due to instability of its API. typed_ast supports parsing Python 2.7 code but unfortunately itself only runs in Python 3.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #2755 
- Note: python3.8+ or typed_ast is required to use this feature


sphinx/util/inspect.py
```diff
@@ -315,6 +317,76 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
...
+def parse_argtypes(s: str) -> Generator[str, None, None]:
+    """Parse argument part of type_comment."""
+    start = 0
+    parens = 0
+    for i, char in enumerate(s):
+        if char == '[':
+            parens += 1
+        elif char == ']':
+            parens -= 1
+        elif char == ',' and parens == 0:
```
eric-wieser: This doesn't correctly handle type comments like `# type: (Literal['['], int) -> int`

sphinx/util/inspect.py
```diff
@@ -315,6 +317,76 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
...
+        else:
+            module = parse(source)
+            subject = module.body[0]  # type: ignore
+        type_comment = subject.type_comment
+
+        if type_comment is None:  # no type_comment
+            return {}
+        else:
+            type_hints = {}  # type: Dict[str, Any]
+            argtypes, rtype = type_comment_re.match(type_comment).groups()
```
eric-wieser: Instead of using string parsing, can you use `ast.parse(type_comment, mode='func_type')`?

sphinx/util/inspect.py
tk0miya: You're right. I should already think about `Literal["\""]`...

sphinx/util/inspect.py
```diff
@@ -372,6 +444,14 @@ def __init__(self, subject: Callable, bound_method: bool = False,
             # we try to build annotations from argspec.
             self.annotations = {}
 
+        # merge type_comment-based type hints
```
tk0miya: Note: I just added a code to here. But it would be better to add a hook point outside of `Signature` class. It would be also needed a handler for stub files (.pyi). It would be better to enhance online object itself via `obj.__annotations__` interface.

tk0miya: see #4824

sphinx/util/inspect.py
tk0miya: Thank you for the info. I'll take a look it tomorrow!

tk0miya: Fully rewritten with `mode=func_type` now!

sphinx/util/inspect.py
tk0miya: Done.

sphinx/ext/autodoc/__init__.py
```diff
@@ -983,8 +983,10 @@ def format_args(self, **kwargs: Any) -> str:
                     not inspect.isbuiltin(self.object) and
                     not inspect.isclass(self.object) and
                     hasattr(self.object, '__call__')):
+                self.env.app.emit('autodoc-before-format-args', self.object.__call__, False)
```
tk0miya, line 986: I don't think this is good way to do this. But we have to let handlers know the target object is bound method or not...

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,109 @@
...
+        if isinstance(typ, ast.Name):
+            return typ.id
+        elif isinstance(typ, ast.Subscript):
+            return "%s[%s]" % (getvalue(typ.value), getvalue(typ.slice))
+        elif isinstance(typ, ast.Index):
+            return getvalue(typ.value)
+        elif isinstance(typ, ast.Tuple):
+            return ', '.join(str(getvalue(e)) for e in typ.elts)
+        elif isinstance(typ, (ast.Constant, ast.NameConstant)):  # type: ignore
+            return typ.value
```
eric-wieser:
```suggestion
            return repr(typ.value)
```

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,109 @@
...
+    """Get type hints from type_comment style annotation."""
+    def getvalue(typ: ast.AST) -> Any:
+        if isinstance(typ, ast.Name):
+            return typ.id
+        elif isinstance(typ, ast.Subscript):
+            return "%s[%s]" % (getvalue(typ.value), getvalue(typ.slice))
+        elif isinstance(typ, ast.Index):
+            return getvalue(typ.value)
+        elif isinstance(typ, ast.Tuple):
+            return ', '.join(str(getvalue(e)) for e in typ.elts)
```
eric-wieser, line 58:
```suggestion
            return ', '.join(getvalue(e) for e in typ.elts)
```

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,109 @@
...
+    yield from (a.arg for a in func.args.args)
+    if func.args.vararg:
+        yield func.args.vararg.arg
+    if func.args.kwarg:
+        yield func.args.kwarg.arg
+
+
+def get_type_hints_from_type_comment(obj: Any, bound_method: bool) -> Dict[str, str]:
+    """Get type hints from type_comment style annotation."""
+    def getvalue(typ: ast.AST) -> Any:
```
eric-wieser: Simpler to make this return string, I think
```suggestion
    def getvalue(typ: ast.AST) -> str:
```

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,109 @@
...
+def iter_args(func: Union[ast.FunctionDef, ast.AsyncFunctionDef]
+              ) -> Generator[str, None, None]:
+    """Get an iterator for arguments names from FunctionDef node."""
+    if hasattr(func.args, "posonlyargs"):  # py38 or above
+        yield from (a.arg for a in func.args.posonlyargs)  # type: ignore
+    yield from (a.arg for a in func.args.args)
+    if func.args.vararg:
+        yield func.args.vararg.arg
+    if func.args.kwarg:
+        yield func.args.kwarg.arg
```
eric-wieser: This function becomes a fair bit simpler if you drop the `.arg` and move it to line 89 - both comprehensions vanish.

tests/roots/test-ext-autodoc/target/typehints.py
```diff
@@ -2,9 +2,25 @@ def incr(a: int, b: int = 1) -> int:
...
     def incr(self, a: int, b: int = 1) -> int:
         return a + b
+
+    def decr(self, a, b = 1):
+        # type: (int, int) -> int
+        return a - b
+
+
+def complex_func(arg1, arg2, arg3=None, *args, **kwargs):
+    # type: (str, List[int], Tuple[int, Union[str, Unknown]], str, str) -> None
```
eric-wieser: According to [this post](https://stackoverflow.com/a/38066950/102441)
```suggestion
    # type: (str, List[int], Tuple[int, Union[str, Unknown]], *str, **str) -> None
```

Can you add a test for `Literal` too?

tk0miya:
@eric-wieser Thank you for quick and wise comment! Updated.

sphinx/ext/autodoc/typehints.py
eric-wieser: Why did you mark this resolved? There's no point calling `str` when `getvalue` always returns `str` already.

tk0miya:
OMG. `Literal` causes test failures on py37 and older environment...

tk0miya:
oh, I closed this PR wrongly. I'll post another one again

eric-wieser:
Error message [seems truncated](https://ci.appveyor.com/project/sphinxdoc/sphinx/builds/29856148/job/oukgg4ce27lrmhh1#L1887):
```
WARNING: error while formatting arguments for target.typehints.complex_func:
```
Is there a way to get more info?

eric-wieser:
> I'll post another one again

Why not just reopen the PR?

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,109 @@
...
+        elif isinstance(typ, ast.Subscript):
+            return "%s[%s]" % (getvalue(typ.value), getvalue(typ.slice))
+        elif isinstance(typ, ast.Index):
+            return getvalue(typ.value)
+        elif isinstance(typ, ast.Tuple):
+            return ', '.join(str(getvalue(e)) for e in typ.elts)
+        elif isinstance(typ, (ast.Constant, ast.NameConstant)):  # type: ignore
+            return repr(typ.value)
+        else:
+            raise NotImplementedError
```
eric-wieser, line 62:
```suggestion
            raise NotImplementedError('Unable to render {} instances'.format(type(type).__name__))
```

tk0miya:
Reposted as #6984

PR #6984, opened 2020-01-02
Close #2755: autodoc: Support type_comment style annotation

### LINKED ISSUE #2755: Support for Python 2/3 "straddling" type hint comments

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #2755 
- Note: python3.8+ or typed_ast is required to use this feature


sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,109 @@
...
+        elif isinstance(typ, ast.Subscript):
+            return "%s[%s]" % (getvalue(typ.value), getvalue(typ.slice))
+        elif isinstance(typ, ast.Index):
+            return getvalue(typ.value)
+        elif isinstance(typ, ast.Tuple):
+            return ', '.join(getvalue(e) for e in typ.elts)
+        elif isinstance(typ, (ast.Constant, ast.NameConstant)):  # type: ignore
+            return repr(typ.value)
+        else:
+            raise NotImplementedError
```
eric-wieser: https://github.com/sphinx-doc/sphinx/pull/6982#discussion_r362491043

tk0miya: done

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,112 @@
...
+
+def iter_args(func: "Union[ast.FunctionDef, ast.AsyncFunctionDef]"
+              ) -> Generator["ast.arg", None, None]:
+    """Get an iterator for arguments names from FunctionDef node."""
+    if hasattr(func.args, "posonlyargs"):  # py38 or above
+        yield from func.args.posonlyargs  # type: ignore
+    yield from func.args.args
+    if func.args.vararg:
+        yield func.args.vararg
+    if func.args.kwarg:
```
tk0miya: I have to handle `func.args.kwonlyargs`.

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,112 @@
...
+        yield func.args.kwarg
+
+
+def get_type_hints_from_type_comment(obj: Any, bound_method: bool) -> Dict[str, str]:
+    """Get type hints from type_comment style annotation."""
+    def getvalue(typ: "ast.AST") -> str:
```
tk0miya: I have to support `List` node in case of `Callable[[int, int], int]`

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,112 @@
...
+            module = ast_parse('if True:\n' + source)
+            subject = module.body[0].body[0]  # type: ignore
+        else:
+            module = ast_parse(source)
+            subject = module.body[0]  # type: ignore
+
+        if subject.type_comment is None:  # no type_comment
+            return {}
+        else:
+            func = ast.parse(subject.type_comment, mode='func_type')  # type: Any
```
tk0miya: I have to use `ast_parse` instead.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1555,10 +1562,12 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value('autodoc_typehints', "signature", True, ENUM("signature", "none"))
     app.add_config_value('autodoc_warningiserror', True, True)
     app.add_config_value('autodoc_inherit_docstrings', True, True)
+    app.add_event('autodoc-before-format-args')
```
tk0miya: This should be documented in CHANGES and doc.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1555,10 +1562,12 @@ def setup(app: Sphinx) -> Dict[str, Any]:
...
     app.add_event('autodoc-process-docstring')
     app.add_event('autodoc-process-signature')
     app.add_event('autodoc-skip-member')
 
     app.connect('config-inited', merge_autodoc_default_flags)
+    app.setup_extension('sphinx.ext.autodoc.typehints')
```
tk0miya: I feel this implementation is too young. So it would be better to be experimental and not to load it automatically.

sphinx/ext/autodoc/typehints.py
```diff
@@ -0,0 +1,112 @@
...
+    if hasattr(func.args, "posonlyargs"):  # py38 or above
+        yield from func.args.posonlyargs  # type: ignore
+    yield from func.args.args
+    if func.args.vararg:
+        yield func.args.vararg
+    if func.args.kwarg:
+        yield func.args.kwarg
+
+
+def get_type_hints_from_type_comment(obj: Any, bound_method: bool) -> Dict[str, str]:
```
tk0miya: Note: I just implement more rich version of this function on #6990

sphinx/ext/autodoc/typehints.py
tk0miya: Finally I used `Signature.parameters` instead. Fixed.

sphinx/ext/autodoc/typehints.py
tk0miya: I cherry-pickeed `ast.unparse()` from #6990. Fixed.

sphinx/ext/autodoc/typehints.py
tk0miya: done

sphinx/ext/autodoc/__init__.py
tk0miya: done

sphinx/ext/autodoc/typehints.py
tk0miya: cherry-picked.

sphinx/ext/autodoc/__init__.py
tk0miya: Withdrawing this comment.

sphinx/pycode/ast.py
```diff
@@ -0,0 +1,79 @@
...
+        # typed_ast is used to parse type_comments if installed.
+        return ast.parse(code, mode=mode)
+
+
+def unparse(node: ast.AST) -> str:
+    """Unparse an AST to string."""
+    if node is None:
+        return None
+    elif isinstance(node, ast.Attribute):
+        return "%s.%s" % (unparse(node.value), node.attr)
```
eric-wieser, line 42: ~~I suspect this will fail for something like `(a + b).c`, which will be emitted as `a + b.c`. Maybe worth just adding a TODO for that, since it's a corner case that's hard to fix.~~

Nevermind, you're safe because you don't parse any binary operators.

tk0miya: Yes, this function does not support BinOp. So it will not come here. I don't know binary operators are allowed in type_comment. But I guess nobody uses it in actual code. (also in signature; #6990 )

CHANGES
```diff
@@ -35,6 +35,10 @@ Features added
   images (imagesize-1.2.0 or above is required)
 * #6994: imgconverter: Support illustrator file (.ai) to .png conversion
 * autodoc: Support Positional-Only Argument separator (PEP-570 compliant)
+* #2755: autodoc: Support type_comment style (ex. ``# type: (str) -> str``)
+  annotation (python3.8+ or `typed_ast <https://github.com/python/typed_ast>`_
+  is required)
+* #2755: autodoc: Add new event: :event:`autodoc-before-format-args`
```
tk0miya: Oh, this is old name.

sphinx/ext/autodoc/type_comment.py
```diff
@@ -0,0 +1,69 @@
...
+from sphinx.application import Sphinx
+from sphinx.pycode.ast import parse as ast_parse
+from sphinx.pycode.ast import unparse as ast_unparse
+from sphinx.util import inspect
+from sphinx.util import logging
+
+logger = logging.getLogger(__name__)
+
+
+def get_type_comment(obj: Any) -> ast.FunctionDef:
```
tk0miya, line 26: It would be better to add docstring here.

tk0miya (commented):
It's time to merge. I'll merge this after minor fixes.

tk0miya:
Merged. Thank you for reviewing! @shimizukawa @eric-wieser

eric-wieser:
What's the purpose of `autodoc-before-process-signature` here? Did you envisage any use case other than parsing type comments?

It seems like this code would become  simpler if you just built the type comment knowledge into `sphinx.util.inspect.signature`.

eric-wieser:
Edit: Ah, I see it's to make it opt in. Maybe the extension should be called `sphinx.ext.typecomments`, and be hooked into at the `inspect.signature` level?

tk0miya:
It was added to update the signature of function before processing. The use case of the event are 1) type_comments and 2) .pyi files (refs: #4824). I don't know there are any more use case for the event. But I thought it is simple to separate them from `inspect.signature()`. What do you think?

In addition, the `sphinx.ext.autodoc.type_comment` was young during 2.4.x. So I marked it as experimental, and separate it to an independent extension from autodoc not to load by default. But, now it becomes a formal feature since 3.0. So it is a part of autodoc and loaded by default.

eric-wieser:
> But I thought it is simple to separate them from `inspect.signature()`. What do you think?

What I've found is that you seem to end up having to implement the internals of `inspect.signature` repeatedly, because every time it visits a function, you need to call the hook before looking at it.

See #7384 for an example of this.

Good point about the `.pyi` files, I hadn't thought about that. Although perhaps simpler would be a third party module that consumes a `.pyi` file and applies all the annotations in it, rather than only applying them when requested.

tk0miya:
Indeed. The autodoc + type_comment + Sphinx's inspect module would be an enhanced version of `inspect.signature()`. It generates documents for functions (signatures) from living objects and source code. I considered the responsibility of `inspect.signature()` is to obtain a signature from a living object. And to parse type_comment from source code is not. I don't know this is a best design.
Personally, I don't like to create a complex function having multiple responsibilities. But I'm okay to change the design if your proposal is good :-)

Anyway, the event has been released as stable. So we need to keep it working for a while if we'll mark it deprecated.

About the `.pyi` files, I still don't take a look in detail. So I can't discuss it well. As you said, it might be better to hook them on loading a target module.

PR #6989, opened 2020-01-03
updating docstring for add_js_file for body element

### LINKED ISSUE #6973: Document ability to add <script> text to header without using a file
I'm working on a sphinx extension that uses a little Javascript library, and was looking for ways to use variables in the sphinx configuration *within* the Javascript code. I couldn't find a way to do this, then I noticed that the MathJax library calls the `add_js_file` method like so:

`.add_js_file(None, type="text/x-mathjax-config", body=body)`

I gave this a shot using something like this:

`.add_js_file(None, body="var myvariable = 'foo';)`

and it worked! (it was placed in the head of my built docs). 

I couldn't find this functionality documented anywhere. Is this a reasonable thing to do? And if so, would it be helpful if I made a change to the docstring of `add_js_file`? (or elsewhere in the docs)

choldgraf:
Documents how to add to the body of `script` elements with `add_js_file`.

closes https://github.com/sphinx-doc/sphinx/issues/6973

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -861,8 +861,10 @@ class Sphinx:
 
         Add *filename* to the list of JavaScript files that the default HTML
         template will include.  The filename must be relative to the HTML
-        static path , or a full URI with scheme.  The keyword arguments are
-        also accepted for attributes of ``<script>`` tag.
+        static path , or a full URI with scheme.  If the keyword argument
+        ``body`` is given, its value will be added between the
+        ``<script>`` tags. Extra keyword arguments are included as
+        attributes of the ``<script>`` tag.
 
         Example::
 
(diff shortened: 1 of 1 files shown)


sphinx/application.py
```diff
@@ -871,6 +873,9 @@ def add_js_file(self, filename: str, **kwargs: str) -> None:
 
             app.add_js_file('example.js', async="async")
             # => <script src="_static/example.js" async="async"></script>
+
```
tk0miya: Oh, flake8 warns for this line. Could you remove spaces from here please?

tk0miya (commented):
LGTM with nits.

sphinx/application.py
choldgraf: done!

tk0miya:
Merged! Thank you for your contribution!

Issue #6417, opened 2019-05-31
More fine-grained highlighting of parameters in function signatures


timhoffm:
Parameters in function signatures are currently parsed to a single html element such as
`<em class="sig-param">myparam=4</em>`

It would be nice to split this down further into classed spans, so that we could highlight different parts of the parameters separately. For example:
~~~
<em class="sig-param">
<span class="sig-param-name">myparam</span>=<span class="sig-param-value">4</span>
</em>
~~~
(line breaks only for better readability. Actual code wound not have linebreaks as they would result in whitespace in the rendered HTML.)

Things to consider:
- This requires further splitting at the parser level.
- Maybe even add a class for `=`?
- Type annotations could be considered as well, as a separate span.

This is a follow-up on #4390/#6401 and #5868.

tk0miya:
Will do.

Thom1729:
Instead of `em`, `var` seems semantically appropriate.

tk0miya:
I just made a draft PR #6990 to realize this feature. Please check it if you have time :-)

tk0miya:
Now I merged #6990. It allows you to decorate function signatures via CSS more flexibly. Could you try it please? I don't think it still does not cover the complex case like #5868. So we need to improve the output more.
Thanks,

timhoffm:
@tk0miya Thanks! This works for me.

Issue #4289, opened 2017-12-11
Flexible styling of declarations


jakobandersen:
Currently there are very few options for styling declarations, essentially only a few ``desc_*`` nodes in ``sphinx/addnodes.py``. For some languages (e.g., C++) declarations may contain many keywords and built-in types that should be coloured to help readability. It would be cool if themes could define custom styles.

I'm not sure what the best way to go about this is, but how about the following.
From Pygments (``pygments.tokens``) I found the following types of tokens:
'Comment', 'Error', 'Escape', 'Generic', 'Keyword', 'Literal', 'Name', 'Number', 'Operator', 'Other', 'Punctuation', 'STANDARD_TYPES', 'String', 'Text', 'Token', 'Whitespace'
To ``addnodes.py`` we could add additional ``desc_*`` nodes for those we don't have yet, and that makes sense. Each should be constructible with a domain name. In terms of CSS each of them should emit two classes; a generic one and a domain-specific one (if a domain is specified).

- Can this be done in a format-agnostic way with Docutils? (via the ``classes`` attribute?)
- Can we make the nodes inherit from a common class so writers can opt-in to support this?
- As this overlaps a lot with Pygments, is there anything we can reuse? E.g., the style maps?
- Is there a better way to implement this?
- Are people against this feature?

jakobandersen:
In https://github.com/jakobandersen/sphinx/tree/4289_decl_styling there is a proof-of-concept implementation, based on the ideas outlined above (The current changes simply colours the C++ keyword ``template`` blue in the HTML output):
- There is a single new node type for text elements in declarations (to avoid a lot of boiler-plate code in each writer).
- The node type is not supposed to be constructed directly, but by functions for each syntactic construct.
- The node always adds 2 or 3 elements to the ``classes`` attribute of it self (are the class names appropriate?):
   - ``d``, a generic class for all description elements.
   - A class specific for the syntactic construct (e.g., ``k`` for keywords).
   - If a domain is specified, also ``<e>-<domain>`` where ``<e>`` is the is the element above, and ``<domain>`` is the given domain (e.g., ``k-cpp`` for a keyword in the C++ domain).
- Only the HTML formats are actually affected by these changes, though I expect a set of macros for the Latex format could be introduced, controlled by checking the ``classes`` attribute in the Latex writer.
 Maybe @jfbu already have some thoughts of how to do this best?
[...]

PR #6990, opened 2020-01-04
py domain: Allow to make a style for arguments of functions and methods

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #6417 
- This tries to parse function signatures with AST parser
- If succeeded, it wraps argument name, annotation and default value with inline node to decorate by CSS or LaTeX macros.
- ~This adds new dependency: `cached-property`. So I'll merge this into master branch.~

### Remaining tasks

- [x] tests
- [x] ~~documentation~~
- [x] design
  - ~~additional nodes is needed?~~ => No.
  - style name is appropriate?

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -19,6 +19,8 @@ Incompatible changes
   when ``:inherited-members:`` and ``:special-members:`` are given.
 * #6830: py domain: ``meta`` fields in info-field-list becomes reserved.  They
   are not displayed on output document now
+* #6417: py domain: doctree of desc_parameterlist has been changed.  The
+  argument names, annotations and default values are wrapped with inline node
 * The structure of ``sphinx.events.EventManager.listeners`` has changed
 * Due to the scoping changes for :rst:dir:`productionlist` some uses of
   :rst:role:`token` must be modified to include the scope which was previously
@@ -67,6 +69,7 @@ Features added
 * #6830: py domain: Add new event: :event:`object-description-transform`
 * #6895: py domain: Do not emit nitpicky warnings for built-in types
 * py domain: Support lambda functions in function signature
+* #6417: py domain: Allow to make a style for arguments of functions and methods
 * Support priority of event handlers. For more detail, see
   :py:meth:`.Sphinx.connect()`
 * #3077: Implement the scoping for :rst:dir:`productionlist` as indicated
--- a/sphinx/addnodes.py
+++ b/sphinx/addnodes.py
@@ -12,7 +12,7 @@ import warnings
 from typing import Any, Dict, List, Sequence
 
 from docutils import nodes
-from docutils.nodes import Node
+from docutils.nodes import Element, Node
 
 from sphinx.deprecation import RemovedInSphinx40Warning
 
@@ -174,6 +174,31 @@ class desc_content(nodes.General, nodes.Element):
     """
 
 
+class desc_sig_element(nodes.inline):
+    """Common parent class of nodes for inline text of a signature."""
+    classes = []  # type: List[str]
+
+    def __init__(self, rawsource: str = '', text: str = '',
+                 *children: Element, **attributes: Any) -> None:
+        super().__init__(rawsource, text, *children, **attributes)
+        self['classes'].extend(self.classes)
+
+
+class desc_sig_name(desc_sig_element):
+    """Node for a name in a signature."""
+    classes = ["n"]
+
+
+class desc_sig_operator(desc_sig_element):
+    """Node for an operator in a signature."""
+    classes = ["o"]
+
+
+class desc_sig_punctuation(desc_sig_element):
+    """Node for a punctuation in a signature."""
+    classes = ["p"]
+
+
 # new admonition-like constructs
 
 class versionmodified(nodes.Admonition, nodes.TextElement):
@@ -332,6 +357,9 @@ def setup(app: "Sphinx") -> Dict[str, Any]:
     app.add_node(desc_optional)
     app.add_node(desc_annotation)
     app.add_node(desc_content)
+    app.add_node(desc_sig_name)
+    app.add_node(desc_sig_operator)
+    app.add_node(desc_sig_punctuation)
     app.add_node(versionmodified)
     app.add_node(seealso)
     app.add_node(productionlist)
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -75,35 +75,42 @@ def _parse_arglist(arglist: str) -> addnodes.desc_parameterlist:
     for param in sig.parameters.values():
         if param.kind != param.POSITIONAL_ONLY and last_kind == param.POSITIONAL_ONLY:
             # PEP-570: Separator for Positional Only Parameter: /
-            params += addnodes.desc_parameter('', nodes.Text('/'))
+            params += addnodes.desc_parameter('', '', addnodes.desc_sig_operator('', '/'))
         if param.kind == param.KEYWORD_ONLY and last_kind in (param.POSITIONAL_OR_KEYWORD,
                                                               param.POSITIONAL_ONLY,
                                                               None):
             # PEP-3102: Separator for Keyword Only Parameter: *
-            params += addnodes.desc_parameter('', nodes.Text('*'))
+            params += addnodes.desc_parameter('', '', addnodes.desc_sig_operator('', '*'))
 
         node = addnodes.desc_parameter()
         if param.kind == param.VAR_POSITIONAL:
-            node += nodes.Text('*' + param.name)
+            node += addnodes.desc_sig_operator('', '*')
+            node += addnodes.desc_sig_name('', param.name)
         elif param.kind == param.VAR_KEYWORD:
-            node += nodes.Text('**' + param.name)
+            node += addnodes.desc_sig_operator('', '**')
+            node += addnodes.desc_sig_name('', param.name)
         else:
-            node += nodes.Text(param.name)
+            node += addnodes.desc_sig_name('', param.name)
 
         if param.annotation is not param.empty:
-            node += nodes.Text(': ' + param.annotation)
+            node += addnodes.desc_sig_punctuation('', ':')
+            node += nodes.Text(' ')
+            node += addnodes.desc_sig_name('', param.annotation)
         if param.default is not param.empty:
             if param.annotation is not param.empty:
-                node += nodes.Text(' = ' + str(param.default))
+                node += nodes.Text(' ')
+                node += addnodes.desc_sig_operator('', '=')
+                node += nodes.Text(' ')
             else:
-                node += nodes.Text('=' + str(param.default))
+                node += addnodes.desc_sig_operator('', '=')
+            node += nodes.inline('', param.default, classes=['default_value'])
 
         params += node
         last_kind = param.kind
 
     if last_kind == Parameter.POSITIONAL_ONLY:
         # PEP-570: Separator for Positional Only Parameter: /
-        params += addnodes.desc_parameter('', nodes.Text('/'))
+        params += addnodes.desc_parameter('', '', addnodes.desc_sig_operator('', '/'))
 
     return params
 
--- a/sphinx/transforms/post_transforms/__init__.py
+++ b/sphinx/transforms/post_transforms/__init__.py
@@ -8,7 +8,7 @@
     :license: BSD, see LICENSE for details.
 """
 
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Tuple, Type
 from typing import cast
 
 from docutils import nodes
@@ -22,6 +22,7 @@ from sphinx.errors import NoUri
 from sphinx.locale import __
 from sphinx.transforms import SphinxTransform
 from sphinx.util import logging
+from sphinx.util.docutils import SphinxTranslator
 from sphinx.util.nodes import process_only_nodes
 
 
@@ -186,9 +187,41 @@ class OnlyNodeTransform(SphinxPostTransform):
         process_only_nodes(self.document, self.app.builder.tags)
 
 
+class SigElementFallbackTransform(SphinxPostTransform):
+    """Fallback desc_sig_element nodes to inline if translator does not supported them."""
+    default_priority = 200
+
+    SIG_ELEMENTS = [addnodes.desc_sig_name,
+                    addnodes.desc_sig_operator,
+                    addnodes.desc_sig_punctuation]
+
+    def run(self, **kwargs: Any) -> None:
+        def has_visitor(translator: Type[nodes.NodeVisitor], node: Type[Element]) -> bool:
+            return hasattr(translator, "visit_%s" % node.__name__)
+
+        translator = self.app.builder.get_translator_class()
+        if isinstance(translator, SphinxTranslator):
+            # subclass of SphinxTranslator supports desc_sig_element nodes automatically.
+            return
+
+        if all(has_visitor(translator, node) for node in self.SIG_ELEMENTS):
+            # the translator supports all desc_sig_element nodes
+            return
+        else:
+            self.fallback()
+
+    def fallback(self):
+        for node in self.document.traverse(addnodes.desc_sig_element):
+            newnode = nodes.inline()
+            newnode.update_all_atts(node)
+            newnode.extend(node)
+            node.replace_self(newnode)
+
+
 def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_post_transform(ReferencesResolver)
     app.add_post_transform(OnlyNodeTransform)
+    app.add_post_transform(SigElementFallbackTransform)
 
     return {
         'version': 'builtin',
--- a/tests/test_build_html.py
+++ b/tests/test_build_html.py
@@ -177,7 +177,8 @@ def test_html4_output(app, status, warning):
     ],
     'autodoc.html': [
         (".//dl[@class='py class']/dt[@id='autodoc-target-class']", ''),
-        (".//dl[@class='py function']/dt[@id='autodoc-target-function']/em", r'\*\*kwds'),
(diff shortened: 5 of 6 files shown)


CHANGES
```diff
@@ -4,6 +4,7 @@ Release 3.0.0 (in development)
 Dependencies
 ------------
 
+* Add ``cached-property`` package (for py37 or older)
```
tk0miya: https://github.com/pydanny/cached-property is a very small library to implement "cached property" for py3.7 or older.

At first, I thought to implement this as `sphinx.util:cached_property` once. But reinventing wheels is boring. In addition, `cached_property` is "battery included" since python 3.8. So I think no reason to implement it again on Sphinx.
https://docs.python.org/3/library/functools.html#functools.cached_property

sphinx/domains/python.py
```diff
@@ -62,6 +63,23 @@
...
+def _parse_arglist(arglist: str) -> addnodes.desc_parameterlist:
+    """Parse a list of arguments using python parser"""
+    params = addnodes.desc_parameterlist(arglist)
+    sig = Signature('(%s)' % arglist)
+    for param in sig.parameters.values():
+        params += addnodes.desc_parameter()
+        params[-1] += nodes.inline('', param.name, classes=['argument'])
+        if param.annotation is not param.empty:
+            params[-1] += nodes.Text(': ')
+            params[-1] += nodes.inline('', param.annotation, classes=['annotation'])
```
tk0miya: In future, we can add cross reference feature to here. I believe type hints is one of good usage for cross reference.

sphinx/domains/python.py
```diff
@@ -62,6 +63,23 @@
 }
 
 
+def _parse_arglist(arglist: str) -> addnodes.desc_parameterlist:
+    """Parse a list of arguments using python parser"""
+    params = addnodes.desc_parameterlist(arglist)
+    sig = Signature('(%s)' % arglist)
+    for param in sig.parameters.values():
+        params += addnodes.desc_parameter()
+        params[-1] += nodes.inline('', param.name, classes=['argument'])
```
tk0miya: As commented at https://github.com/sphinx-doc/sphinx/issues/6417#issuecomment-499100847, `<var>` is better to represent variable names in HTML5. So it would be better to use new node to represent it. But it causes incompatibility to writers...

sphinx/domains/python.py
```diff
@@ -62,6 +63,23 @@
...
+        if param.annotation is not param.empty:
+            params[-1] += nodes.Text(': ')
```
tk0miya: Is some style needed for such symbols?

sphinx/domains/python.py
```diff
@@ -284,7 +302,11 @@ def handle_signature(self, sig: str, signode: desc_signature) -> Tuple[str, str]
 
         signode += addnodes.desc_name(name, name)
         if arglist:
-            _pseudo_parse_arglist(signode, arglist)
+            try:
+                signode += _parse_arglist(arglist)
+            except RuntimeError:
+                # failed to parse arglist by python's parser
+                _pseudo_parse_arglist(signode, arglist)
```
tk0miya: Historically, we also accepts extended syntax to represent optional argument: `func(foo[, bar])`. So old parser is still needed.

sphinx/pycode/function.py
```diff
@@ -0,0 +1,124 @@
...
+    if node is None:
+        return None
+    elif isinstance(node, ast.Attribute):
+        return '%s.%s' % (stringify(node.value), node.attr)
+    elif isinstance(node, ast.Bytes):
+        return repr(node.s)
+    elif isinstance(node, ast.Call):
+        args = ([stringify(e) for e in node.args] +
+                ['%s=%s' % (k.arg, stringify(k.value)) for k in node.keywords])
+        return '%s(%s)' % (stringify(node.func), ', '.join(args))
```
tk0miya: I worried this implementation working fine. Call node can contain many kinds of expr...

sphinx/pycode/function.py
```diff
@@ -0,0 +1,124 @@
...
+        items = (k + ': ' + v for k, v in zip(keys, values))
+        return '{' + ', '.join(items) + '}'
+    elif isinstance(node, ast.Ellipsis):
+        return '...'
+    elif isinstance(node, ast.Constant):
+        return repr(node.value)
+    elif isinstance(node, ast.Index):
+        return stringify(node.value)
+    elif isinstance(node, ast.Lambda):
+        return '<function <lambda>>'  # TODO
```
tk0miya: Supporting lambda is heavy task to me. But it would be important for some users, I think.

sphinx/pycode/function.py
```diff
@@ -0,0 +1,124 @@
...
+        return "{" + ', '.join(stringify(e) for e in node.elts) + "}"
+    elif isinstance(node, ast.Str):
+        return repr(node.s)
+    elif isinstance(node, ast.Subscript):
+        return "%s[%s]" % (stringify(node.value), stringify(node.slice))
+    elif isinstance(node, ast.Tuple):
+        return ', '.join(stringify(e) for e in node.elts)
+    else:
+        # TODO: f-string
+        raise NotImplementedError('Unable to render %s instance' % type(node).__name__)
```
tk0miya: At present, this function raises `NotImplementedError`. So it caused crashes. So we need to handle this.

sphinx/pycode/function.py
```diff
@@ -0,0 +1,124 @@
...
+from typing import cast
+from typing import Mapping
+
+if sys.version_info > (3, 8):
+    from functools import cached_property
+else:
+    from cached_property import cached_property
+
+
+def stringify(node: ast.AST) -> str:
```
eric-wieser: `unparse` might be better - a future version of ast will have a builtin `unparse` member

sphinx/pycode/function.py
```diff
@@ -0,0 +1,124 @@
...
+
+    This class provides ``inspect.Signature`` like interface for string based call signatures::
+
+        sig = Signature('(a, b, *args, **kwargs)')
+        sig.parameters  # => returns a dict for Parameters
+    """
+
+    def __init__(self, signature: str) -> None:
+        module = ast.parse('def func' + signature + ': pass')
+        self.definition = cast(ast.FunctionDef, module.body[0])
```
eric-wieser: Alternatively, to avoid needing any of the `stringify` stuff:
```python
d = {}
exec("""
from __future__ import annotations  # 3.7 onwards
def func{sig}: pass
""".format(sig=signature), globals=d)
self.definition = inspect.signature(d['func'])
# annotations will be strings already
```

sphinx/pycode/function.py
```diff
@@ -0,0 +1,124 @@
...
+    elif isinstance(node, ast.Subscript):
+        return "%s[%s]" % (stringify(node.value), stringify(node.slice))
+    elif isinstance(node, ast.Tuple):
+        return ', '.join(stringify(e) for e in node.elts)
+    else:
+        # TODO: f-string
+        raise NotImplementedError('Unable to render %s instance' % type(node).__name__)
+
+
+class Signature:
```
eric-wieser: I don't think you need a new class here. Why not
```
def signature_from_str(s):
    ...
    return inspect.Signature(parameters=..., return_annotation=...)
```

sphinx/pycode/function.py
tk0miya: Unfortunately we're still supporting py35 and py36.

sphinx/pycode/function.py
tk0miya: OMG. I mistook `inspect.signature` (function) and inspect.Signature` (class). As you said, we can use `inspect.Signature` class directly.

sphinx/pycode/function.py
tk0miya: Good to hear. It's good naming!
https://mail.python.org/archives/list/python-dev@python.org/thread/JAQDBMC23HW2PQ27HQNJ7G244T423IDD/

CHANGES
tk0miya: After refactoring, no reason to use the library. So there are also no reason to merge this into master branch.

tk0miya: Ah, it might be needed if we'll add a new docutils node...

sphinx/pycode/function.py
tk0miya: done.

sphinx/pycode/function.py
tk0miya: done.

sphinx/util/inspect.py
```diff
@@ -313,6 +315,98 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
...
+    elif isinstance(node, ast.Subscript):
+        return "%s[%s]" % (ast_unparse(node.value), ast_unparse(node.slice))
+    elif isinstance(node, ast.Tuple):
+        return ', '.join(ast_unparse(e) for e in node.elts)
+    else:
+        # TODO: f-string
+        raise NotImplementedError('Unable to render %s instance' % type(node).__name__)
+
+
+def signature_from_str(signature: str) -> inspect.Signature:
```
eric-wieser: It's possible there's a trick you can play here using `__text_signature__`, will see if I can construct it when I'm back at a shell

Thom1729:
Would it make sense to wrap each argument in an element? Something like:

```py
def join(*items: List, delimiter: str = ''):
```

```html
<span class="argument">
    <span class="argument-sigil">*</span><span class="argument-name">items</span>
    : <span class="annotation">List</span>
</span>,
<span class="argument">
    <span class="argument-name">delimiter</span>
    : <span class="annotation">str</span>
    = <span class="default-value">''</span>
</span>
```

This way, you could apply CSS to the entire argument, e.g. to prefer line breaks between arguments rather than within them.

tk0miya:
Now HTML writer use `<em>` to wrap each argument. +1 for changing it to `<span>` tag :-)

jakobandersen:
Unfortunately I have not had time to take a closer look at this, so the following may be redundant.
The need for styling declarations is there in almost every domain so it would be nice if this PR does not exclude a general solution, or reserves generic CSS class names for the Python domain. For example, a generic CSS class ``argument-name`` should be used to style across all domains and ``py-argument-name`` should be used to apply styling to Python declarations only.
As explored and discussed a bit in #4289, one approach for the styling categories is to reuse (and refine) what is already in Pygments. Note that it should be possible to set the theme for declarations independently of the Pygments theme used for code snippets.

sphinx/domains/python.py
tk0miya: comment added.

sphinx/pycode/function.py
tk0miya: catched.

sphinx/util/inspect.py
eric-wieser: Ah, I'm thinking of:
```python
>>> inspect._signature_fromstr(inspect.Signature, lambda: None, "(a, b, c)")
<Signature (a, b, c)>
```
but unfortunately
```python
>>> inspect._signature_fromstr(inspect.Signature, lambda: None, "(a, b, c: int)")
ValueError: Annotations are not currently supported
```
It would be great to fix this upstream in CPython

sphinx/domains/python.py
```diff
@@ -62,6 +63,34 @@
...
+    params = addnodes.desc_parameterlist(arglist)
+    sig = signature_from_str('(%s)' % arglist)
+    for param in sig.parameters.values():
+        node = addnodes.desc_parameter()
+        if param.kind == param.VAR_POSITIONAL:
+            node += nodes.inline('', '*' + param.name, classes=['argument'])
+        elif param.kind == param.VAR_KEYWORD:
+            node += nodes.inline('', '**' + param.name, classes=['argument'])
+        else:
+            node += nodes.inline('', param.name, classes=['argument'])
```
eric-wieser: What about the `*` and `/` from signatures like the following? Does something insert those elsewhere?
```python
def f1(a, b, *, c=2): pass
def f2(a, b, /, **kwargs): pass
def f3(a, b, /, *, d=2): pass
def f4(a, b, /, c, *, d=2): pass
def f5(a, b, /): pass
```

eric-wieser: The logic needs to be something like:
```python
last_param_kind = None
for param in params:
    if last_param_kind == POSITIONAL_ONLY and param.kind != POSITIONAL_ONLY :
        params += nodes.Text('/')
    if last_param_kind != VAR_POSITIONAL and param.kind == KEYWORD_ONLY:
        params += nodes.Text('*')
    last_param_kind = param.kind

    node = addnodes.desc_parameter()
    # ... as before

# trailing /, like `def foo(a, b, /): pass`
if last_param_kind == POSITIONAL_ONLY:
    params += nodes.Text('/')
```

sphinx/util/inspect.py
tk0miya: As its docstring says, `_signature_fromstr()` is private. We have to discuss about exposure before fixing.
Anyway, it is not a scope of this PR. marked as resolved.

tk0miya: I noticed it might be better to rename function name from "..._from_str" to "..._fromstr". But I don't mind either.

eric-wieser: > We have to discuss about exposure before fixing.

I think this part of the patch would be valuable upstream whether sphinx uses it or not

shimizukawa:
I reviewed the doctree structure. I think it is enough to use the `desc_parameter` node. I don't think we need to add nodes for argument names or type names. So, LGTM.

eric-wieser:
It would be neat if sphinx could parse a type annotation like `MyCollection[MyType]` into links to both `MyCollection` and `MyType`, but I think that should be left for a separate patch.

sphinx/pycode/function.py
tk0miya: I guess only very few people write lambda functions in signature of `py:function` directive.

tk0miya:
>The need for styling declarations is there in almost every domain so it would be nice if this PR does not exclude a general solution, or reserves generic CSS class names for the Python domain.

It is better to surround these element by domain-classed element. It helps to choose individual element by CSS selector like `dl.py.function > .argument-name`. I'll work for it in another PR.

jakobandersen:
> It is better to surround these element by domain-classed element. It helps to choose individual element by CSS selector like `dl.py.function > .argument-name`. I'll work for it in another PR.

Good point, and nice solution. That probably also minimizes the size of the output.

sphinx/domains/python.py
tk0miya: done

sphinx/domains/python.py
```diff
@@ -62,6 +64,51 @@
...
 
+def _parse_arglist(arglist: str) -> addnodes.desc_parameterlist:
+    """Parse a list of arguments using python parser"""
+    params = addnodes.desc_parameterlist(arglist)
+    sig = signature_from_str('(%s)' % arglist)
+    last_kind = None
+    for param in sig.parameters.values():
+        if param.kind != param.POSITIONAL_ONLY and last_kind == param.POSITIONAL_ONLY:
+            # PEP-570: Separator for Positional Only Parameter: /
+            params += nodes.inline('', '/', classes=['operator'])
```
tk0miya: To go forward this, we need to discuss more about how to name CSS classes (refs: #4289)...

tk0miya:
I think some discussion points are still remained.
* CSS rules (refs: #4289)
* DOM structure

To concentrate these points, I'm going to make another PR for signature parser based on this PR. I think it makes this PR simple.

tk0miya:
Just updated. ready for review now!

sphinx/addnodes.py
```diff
@@ -174,6 +174,31 @@ class desc_content(nodes.General, nodes.Element):
     """
 
 
+class desc_sig_element(nodes.inline):
```
tk0miya, line 177: I added desc_sig_element and children to realize pygments based inline text decorator (see #4289).
But I also added `SigElementFallbackTransform` to support old translators. It helps to render these nodes as simple inline node in old translators.
It means this does not break compatibility! I confirmed working fine with docxbuilder on my local.

sphinx/domains/python.py
```diff
@@ -72,35 +72,42 @@ def _parse_arglist(arglist: str) -> addnodes.desc_parameterlist:
     for param in sig.parameters.values():
         if param.kind != param.POSITIONAL_ONLY and last_kind == param.POSITIONAL_ONLY:
             # PEP-570: Separator for Positional Only Parameter: /
-            params += nodes.Text('/')
+            params += addnodes.desc_sig_operator('', '/')
```
tk0miya: I found another bug around separators. see #7212

sphinx/domains/python.py
```diff
@@ -72,35 +72,42 @@ def _parse_arglist(arglist: str) -> addnodes.desc_parameterlist:
...
         if param.default is not param.empty:
             if param.annotation is not param.empty:
-                node += nodes.Text(' = ' + str(param.default))
+                node += nodes.Text(' ')
+                node += addnodes.desc_sig_operator('', '=')
+                node += nodes.Text(' ')
             else:
-                node += nodes.Text('=' + str(param.default))
+                node += addnodes.desc_sig_operator('', '=')
+            node += nodes.inline('', param.default, classes=['default_value'])
```
tk0miya, line 106: I can't find proper toke name from pygments. So I used inline node for default value. Is there good idea?
https://github.com/pygments/pygments/blob/master/pygments/token.py

tk0miya:
Is there any comments? If none, I'll merge this soon. The feature freeze period for 3.x will come soon.

sphinx/domains/python.py
shimizukawa: I have no idea...

shimizukawa (approved):
LGTM!

sphinx/addnodes.py
shimizukawa: Excellent!

tk0miya:

PR #6993, opened 2020-01-05
Refactor sphinx.util.inspect:Signature

tk0miya:
### Feature or Bugfix
- Refactoring


tk0miya:
I thought it would be better to deprecate whole of `Signature` class. So I refactor this again as #7002 . So I'm withdrawing this.

PR #6995, opened 2020-01-07
Update documentation of download role for addition of hash in subdir

rgommers:
This was changed in gh-5377, but that forgot the (now slightly misleading) documentation.

Discovered in https://github.com/numpy/numpy.org/pull/14, where we were relying on download links being stable.


tk0miya:
Thanks!

PR #7001, opened 2020-01-08
Stop testing with nightly python

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
At present, latest typted_ast does not support python-3.9a1 or later.
As a result, nightly python in Travis CI gets errored in nearly running.
This stops to use nightly python for testing temporarily.

refs: https://github.com/python/typed_ast/issues/129


gvanrossum:
By upgrading to typed-ast 1.4.1 this should work again.

PR #7002, opened 2020-01-08
refactor: Add sphinx.util.inspect.signature()

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
As a successor of sphinx.util.inspect.Singnature, this adds
signature() function behaves like `inspect.signature()`.  It is
very similar to way of python's inspect module.

In addition, this also adds stringify_annotation() helper to
sphinx.util.inspect module.  With these two functions, we can move
to python's Signature object to represent function signatures
perfectly.  It's natural design for python developers than ever.

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -315,6 +315,112 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
     return getattr(builtins, safe_getattr(cls, '__name__', '')) is cls
 
 
+def signature(subject: Callable, bound_method: bool = False) -> inspect.Signature:
+    """Return a Signature object for the given *subject*.
+
+    :param bound_method: Specify *subject* is a bound method or not
+    """
+    # check subject is not a built-in class (ex. int, str)
+    if (isinstance(subject, type) and
+            is_builtin_class_method(subject, "__new__") and
+            is_builtin_class_method(subject, "__init__")):
+        raise TypeError("can't compute signature for built-in type {}".format(subject))
+
+    try:
+        signature = inspect.signature(subject)
+        parameters = list(signature.parameters.values())
+        return_annotation = signature.return_annotation
+    except IndexError:
+        # Until python 3.6.4, cpython has been crashed on inspection for
+        # partialmethods not having any arguments.
+        # https://bugs.python.org/issue33009
+        if hasattr(subject, '_partialmethod'):
+            parameters = []
+            return_annotation = inspect.Parameter.empty
+        else:
+            raise
+
+    try:
+        # Update unresolved annotations using ``get_type_hints()``.
+        annotations = typing.get_type_hints(subject)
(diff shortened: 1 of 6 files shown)


sphinx/util/inspect.py
```diff
@@ -315,6 +315,114 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
...
+            pass
+        else:
+            if len(parameters) > 0:
+                parameters.pop(0)
+
+    return inspect.Signature(parameters, return_annotation=return_annotation)
+
+
+def stringify_signature(sig: inspect.Signature, show_annotation: bool = True,
+                        show_return_annotation: bool = True) -> str:
```
tk0miya, line 369: Note: I consider `has_retval` behavior is a responsibility of stringify, not create a signature object. So I moved this to here.

sphinx/util/inspect.py
```diff
@@ -315,6 +315,114 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
...
+    for param in sig.parameters.values():
+        # skip first argument if subject is bound method
+        arg = StringIO()
+
+        # insert '*' between POSITIONAL args and KEYWORD_ONLY args::
+        #     func(a, b, *, c, d):
+        if param.kind == param.KEYWORD_ONLY and last_kind in (param.POSITIONAL_OR_KEYWORD,
+                                                              param.POSITIONAL_ONLY,
+                                                              None):
+            args.append('*')
```
tk0miya, line 382: I thought we should support a separator for positional only parameter. I'll do it in another PR.

sphinx/util/inspect.py
```diff
@@ -315,6 +315,114 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
...
+
+        if param.kind in (param.POSITIONAL_ONLY,
+                          param.POSITIONAL_OR_KEYWORD,
+                          param.KEYWORD_ONLY):
+            arg.write(param.name)
+            if show_annotation and param.annotation is not param.empty:
+                arg.write(': ')
+                arg.write(stringify_annotation(param.annotation))
+            if param.default is not param.empty:
+                if param.annotation is param.empty or show_annotation is False:
```
tk0miya: It would be better to invert this condition. Because other conditions are inverted.

sphinx/util/inspect.py
```diff
@@ -315,6 +315,114 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
...
+def stringify_signature(sig: inspect.Signature, show_annotation: bool = True,
+                        show_return_annotation: bool = True) -> str:
+    """Stringify a Signature object.
+
+    :param show_annotation: Show annotation in result
+    """
+    args = []
+    last_kind = None
+    for param in sig.parameters.values():
+        # skip first argument if subject is bound method
```
tk0miya: This comment should be removed.

PR #4197, opened 2017-10-30
Closes #3160: Use <kbd> tag for :kbd: role in html writers

dlech:
Subject: Modify html writers to use `<kbd>` tags for `:kbd:` role.

### Feature or Bugfix

- Feature
- Bugfix
- Both?

### Purpose
- HTML has a `<kbd>` tag to represent keyboard keys, so we should use it.

### Relates
- #3160

Issue #3160, opened 2016-11-18
Why doesn't Sphinx translate :kbd: roles to HTML's <kbd> tags?


Paebbels:
I want to mark keyboard keys and inputs as key "icons". Sphinx provides a `:kbd:` role to markup single characters or words, but the resulting HTML contains `<code class="kbd ......">` tags.

So why doesn't Sphinx translate :kbd: roles to HTML's \<kbd> tags?
Or is this a Read-The-Docs issue? - I'm using their theme

My question on StackOverflow: http://stackoverflow.com/questions/40686837/why-doesnt-sphinx-translate-kbd-roles-to-htmls-kbd-tags

--------------------------------
I found that some CSS can transform `<kbd>`-tags into nice looking keys.

    kbd
    {
      -moz-border-radius:3px;
      -moz-box-shadow:0 1px 0 rgba(0,0,0,0.2),0 0 0 2px #fff inset;
      -webkit-border-radius:3px;
      -webkit-box-shadow:0 1px 0 rgba(0,0,0,0.2),0 0 0 2px #fff inset;
      background-color:#f7f7f7;
      border:1px solid #ccc;
      border-radius:3px;
      box-shadow:0 1px 0 rgba(0,0,0,0.2),0 0 0 2px #fff inset;
      color:#333;
      display:inline-block;
      font-family:Arial,Helvetica,sans-serif;
      font-size:11px;
      line-height:1.4;
      margin:0 .1em;
      padding:.1em .6em;
      text-shadow:0 1px 0 #fff;
    }

Source: http://meta.superuser.com/questions/4788/css-for-the-new-kbd-style

tk0miya:
Maybe no reason. Now sphinx represents a `:kbd:` role as a literal. And HTML writers simply convert it to `<span>` element.

+1 for `<kdb>` tag.

Paebbels:
Does the HTML writer belong to Sphinx or are we in another scope like docutils?

tk0miya:
HTML writer belongs to both docutils and Sphinx.
docutils provides basic feature of HTML writer. and Sphinx customize it.

Anyway, this behavior came from `:kbd:` role. It is provided by Sphinx.
So this is a issue of Sphinx.

dlech:
Related: https://sourceforge.net/p/docutils/mailman/message/35668725/

dlech:
I added this to my `conf.py` as a workaround:

```python

# -- Hackery --

[...]

PR #7005, opened 2020-01-09
Add LaTeX styling hook for :kbd: role

jessetan:
Subject: Add a renewable LaTeX command for keyboard input defined with `:kbd:` role

### Feature or Bugfix
- Feature

### Purpose
The `:kbd:` role can be used to denote input from a keybpard. In HTML, this results in a `<kbd>` element that can be styled with CSS.
In LaTeX, this was treated the same as any other literal. With this PR, it uses a new `\sphinxkeypress` command (defaults to same as literal) which may be renewed to change the appearance

### Relates
- #3160
- #4197

Note: this is based of master, I'm not sure if that was correct.

### DIFF
--- a/doc/latex.rst
+++ b/doc/latex.rst
@@ -817,6 +817,8 @@ Macros
      multiple paragraphs in header cells of tables.
   .. versionadded:: 1.6.3
      ``\sphinxstylecodecontinued`` and ``\sphinxstylecodecontinues``.
+  .. versionadded:: 2.4.0
+     ``\sphinxkeyboard``
 - ``\sphinxtableofcontents``: it is a
   wrapper (defined differently in :file:`sphinxhowto.cls` and in
   :file:`sphinxmanual.cls`) of standard ``\tableofcontents``.  The macro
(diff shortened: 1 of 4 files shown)


doc/latex.rst
```diff
@@ -783,6 +783,8 @@ Macros
      multiple paragraphs in header cells of tables.
   .. versionadded:: 1.6.3
      ``\sphinxstylecodecontinued`` and ``\sphinxstylecodecontinues``.
+  .. versionadded:: 3.0.0???
```
tk0miya: I think this PR does not change the behavior of Sphinx. So this is not a breaking change. So I prefer to add this as 2.4 release. How about rebasing this into 2.0 branch?

doc/latex.rst
```diff
@@ -783,6 +783,8 @@ Macros
...
+     ``\sphinxkeypress``
```
tk0miya: It seems `kbd` does not have a meaning for key "press". `\sphinxkeyboard` or `\sphinxkbd` might be better?
https://html.spec.whatwg.org/#the-kbd-element

jessetan: Changed to `\sphinxkeyboard`

doc/latex.rst
jessetan: OK, I rebased to 2.0 (I think)

tk0miya (approved):
Thank you for quick fix. LGTM!

tk0miya:

PR #7009, opened 2020-01-10
SphinxTranslator calls visitor/departure method for super node class

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- This calls visitor/departure methods for super node class if visitor/departure method for the node class not found.
- It will protect a translator from new node.
- Notice: This only effects to subclasses of `SphinxTranslator`. So it is not almighty.

### DIFF
--- a/sphinx/util/docutils.py
+++ b/sphinx/util/docutils.py
@@ -452,7 +452,10 @@ class ReferenceRole(SphinxRole):
 class SphinxTranslator(nodes.NodeVisitor):
     """A base class for Sphinx translators.
 
-    This class provides helper methods for Sphinx translators.
+    This class adds a support for visitor/departure method for super node class
+    if visitor/departure method for node class is not found.
+
+    It also provides helper methods for Sphinx translators.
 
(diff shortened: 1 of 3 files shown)


sphinx/util/docutils.py
```diff
@@ -464,6 +467,42 @@ def __init__(self, document: nodes.document, builder: "Builder") -> None:
...
         self.settings = document.settings
 
+    def dispatch_visit(self, node):
+        """
+        Dispatch node to appropriate visitor method.
+        The priority of visitor method is:
+
+        1. ``self.visit_{node_class}()``
+        2. ``self.visit_{supre_node_class}()``
+        3. ``self.unknown_visit()``
```
shimizukawa, line 477: If user add a `visit_node` method, user won't reach `unknown_visit`.
However I don't think there is a real problem.

shimizukawa (approved):
LGTM!

tk0miya:
Thank you for reviewing. merged

PR #7010, opened 2020-01-10
[WIP] Add new visitor method for pair of parent node and current node to SphinxTranslator

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Some of visitor methods become complex because its behavior is determined by parent node.
- For example, `nodes.title` is used for sections, admonitions, table and so on.
- This allows to make a visitor method for pair of parent and child node. (ex. `visit_section_title()`).
- It makes visitor methods simple.

### DIFF
--- a/sphinx/util/docutils.py
+++ b/sphinx/util/docutils.py
@@ -452,7 +452,10 @@ class ReferenceRole(SphinxRole):
 class SphinxTranslator(nodes.NodeVisitor):
     """A base class for Sphinx translators.
 
-    This class provides helper methods for Sphinx translators.
+    This class adds a support for visitor/departure method for combination of
+    parent node and current node (ex. ``visit_section_title()``).
+
+    It also provides helper methods for Sphinx translators.
 
     .. note:: The subclasses of this class might not work with docutils.
               This class is strongly coupled with Sphinx.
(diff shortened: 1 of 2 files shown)


sphinx/util/docutils.py
```diff
@@ -464,6 +467,44 @@ def __init__(self, document: nodes.document, builder: "Builder") -> None:
...
+        Dispatch node to appropriate visitor method.
+        The priority of visitor method is:
+
+        1. ``self.visit_{parent_node_class}_{node_class}()``
+        2. ``self.visit_{node_class}()``
+        3. ``self.unknown_visit()``
+        """
+        parent_node_name = node.parent.__class__.__name__
+        node_name = node.__class__.__name__
+        method = getattr(self, 'visit_%s_%s' % (parent_node_name, node_name), None)
```
shimizukawa, line 481: It would be better if `visit_%s__%s` (separate with double underscore) to avoid collision with node name like `foo_bar`.

shimizukawa (approved):
My feeling: Is this really okay?
I'm not convinced that it can be used safely, but I didn't come up with any obvious problems.

tk0miya:
Thank you for comment. I agree this is too hacky solution. So I'm withdrawing this temporarily.

PR #7039, opened 2020-01-20
C++, test role target checks and fix two cases

jakobandersen:
#7038, but on the correct branch

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,10 @@ Features added
 Bugs fixed
 ----------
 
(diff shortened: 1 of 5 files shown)


tk0miya:
Is an urgent release needed? If so, I'll do within a day.

jakobandersen:
It is purely a bug fix (with tests), hence the merge into 2.3.2. Though, to my knowledge I am the only one who has encountered the crash, so I don't think it is urgent at this time.

tk0miya:
Okay, I'll release this if somebody post a report for this :-)

PR #7042, opened 2020-01-21
Added a way to specify emphasize-lines as text fragments

viktorasm:
Subject: Less fragile way to specify emphasized line numbers when using literal includes.

### Feature or Bugfix
- Feature

### Purpose
- For users who want to be able to highlight lines based on contents, not on line numbers

### Detail
- See changes in `directives.rst`.

### Relates
- (no issue raised for this feature)


viktorasm:
Will re-raise as 2.0 PR

PR #7043, opened 2020-01-21
Added a way to specify emphasize-lines as text fragments 

viktorasm:
Subject: Less fragile way to specify emphasized line numbers when using literal includes.

### Feature or Bugfix
- Feature

### Purpose
- For users who want to be able to highlight lines based on contents, not on line numbers

### Detail
- See changes in `directives.rst`.

### Relates
- (no issue raised for this feature)


viktorasm:
@tk0miya shameless bump - can we consider this for 3.4.0 (or scrap the PR, either works for me)?

tk0miya:
I'm very sorry to not respond to you for so long. I feel the `literalinclude` directive is already too complicated. So it would be better not to add new options more. But I understand you need to add such options to the directive. To do that, I'd like to add a hook to enhance it. Does it help you?

viktorasm:
what do you mean by "hook"?
let me know how I can help.

PR #7053, opened 2020-01-23
Production list scoping and line continuation

### LINKED ISSUE #1027: support backslash line continuation in productionlist entries
This issue is to add support for the backslash line continuation character to productionlist entries.

Currently, function signatures support the use of backslashes to continue lines, as documented here:

http://sphinx.pocoo.org/domains.html?highlight=backslash#basic-markup

However, productionlist entries do not support this.

This feature would be useful if you would like a productionlist line following a colon (":") to render on a single line, but without the source line itself being too long (e.g. by exceeding 79 characters).

See the comment below for a specific example.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1027
- Originally reported by: Chris Jerdonek
- Originally created at: 2012-10-26T01:59:03.893

### LINKED ISSUE #3077: productionlist doesn't appear to do anything with "name" argument
The productionlist definition (`.. productionlist:: [name]`) in doc/markup/para.rst cites an optional name argument which `serves to distinguish different sets of production lists that belong to different grammars`

To me, at least, this statement suggests that the [name] argument creates namespaces so that multiple productionlists can define the same tokens as long as they are in a different productionlist namespace (and, of course, that you could successfully link to them). In domains/std.py, the ProductionList's class run method loops over argument lines but does nothing with the identified optional first argument:

```
for rule in self.arguments[0].split('\n'):
    if i == 0 and ':' not in rule:
        # production group
        continue
    ...
```

Either the documentation needs to clarify that this distinction is purely decorative, or the argument should have some substantive effect. If the argument is meaningfully implemented, the `:token:` role should probably also be updated with some way to specify a namespace along with the token.

jakobandersen:
Subject: implement scoping and line continuation for production lists 

### Feature or Bugfix
- Feature
- (Bugfix: the production list documentation mentions scoping, but it was not implemented)

### Purpose
Enable using backslash to escape a line break, making it the source rst easier to read for long production rules.

One may have production lists for unrelated grammars that happens to share token names, resulting in wrong cross-references and warnings about duplicate tokens.
With this PR the code follows the documentation: if the first line of the production list does not contain a colon it will be taken as the token group that the productions belong too, e.g.:
```rst
.. productionlist:: python
   statement: something somethingElse
```
When using the `:token:` role for referencing a token in a production list with a token group, then the group must be specified, e.g., ``:token:`python:statement` ``. As default the whole name is shown, but prefixing with a tilde will show just the token name, e.g., ``:token:`~python:statement` ``. The documentation has been updated to (try to) explain this.

### Details
Some token roles out there may break and give a warning about not being able to find the target. Making a fallback would not be a good idea, as it would silently change target if new unscoped production lists are introduced. Additionally it is ambiguous which scoped token to select.

External links should still work as both a new id and the old id is attached to the production list nodes.

### Relates
- Fixes #3077, scoping
- Fixes #1027, line continuation


doc/usage/restructuredtext/directives.rst
```diff
@@ -1139,24 +1139,32 @@ derived forms), but provides enough to allow context-free grammars to be
...
 
    This directive is used to enclose a group of productions.  Each production
    is given on a single line and consists of a name, separated by a colon from
    the following definition.  If the definition spans multiple lines, each
    continuation line must begin with a colon placed at the same column as in
    the first line.
 
-   The argument to :rst:dir:`productionlist` serves to distinguish different
-   sets of production lists that belong to different grammars.
+   The ``tokenGroup`` argument to :rst:dir:`productionlist` serves to
```
tk0miya: In our doc, normal emphasis `*tokenGroup*` is usually used to decorate argument name.

doc/usage/restructuredtext/directives.rst
```diff
@@ -1139,24 +1139,32 @@ derived forms), but provides enough to allow context-free grammars to be
...
+   define rules in the same scope.
 
    Blank lines are not allowed within ``productionlist`` directive arguments.
 
    The definition can contain token names which are marked as interpreted text
-   (e.g. ``sum ::= `integer` "+" `integer```) -- this generates
+   (e.g. "``sum ::= `integer` "+" `integer```") -- this generates
    cross-references to the productions of these tokens.  Outside of the
-   production list, you can reference to token productions using
+   production list, you can reference token productions using
```
tk0miya: you can **refer**?

sphinx/domains/std.py
```diff
@@ -406,15 +406,17 @@ def run(self) -> List[Node]:
         return messages + [node]
 
 
-def token_xrefs(text: str) -> List[Node]:
+def token_xrefs(text: str, productionGroup: str) -> List[Node]:
```
tk0miya: Could you give a default value `''` to produnctionGroup for compatibility?

sphinx/domains/std.py
```diff
@@ -437,11 +439,18 @@ class ProductionList(SphinxDirective):
     def run(self) -> List[Node]:
         domain = cast(StandardDomain, self.env.get_domain('std'))
         node = addnodes.productionlist()  # type: Element
+        # The backslash handling is from ObjectDescription.get_signatures
+        nl_escape_re = re.compile(r'\\\n')
+        strip_backslash_re = re.compile(r'\\(.)')
+        lines = nl_escape_re.sub('', self.arguments[0]).split('\n')
+        # remove backslashes to support (dummy) escapes; helps Vim highlighting
```
tk0miya: Is this still useful for vim? I don't know this is still needed.

tests/roots/test-productionlist/index.rst
```diff
@@ -0,0 +1,27 @@
+.. toctree::
+
+	P1
+	P2
+	Bare
+	Dup1
+	Dup2
+	firstLineRule
+	LineContinuation
```
tk0miya: Please not to use TABs in reST

sphinx/domains/std.py
```diff
@@ -451,17 +460,38 @@ def run(self) -> List[Node]:
...
             if subnode['tokenname']:
-                idname = nodes.make_id('grammar-token-%s' % subnode['tokenname'])
+                idname = 'grammar-token2-%s_%s' \
+                         % (nodes.make_id(productionGroup), nodes.make_id(name))
                 if idname not in self.state.document.ids:
                     subnode['ids'].append(idname)
+
+                idnameOld = nodes.make_id('grammar-token-' + name)
+                if idnameOld not in self.state.document.ids:
+                    subnode['ids'].append(idnameOld)
```
tk0miya, line 470: Is this old node_id still needed?

sphinx/domains/std.py
```diff
@@ -451,17 +460,38 @@ def run(self) -> List[Node]:
             subnode = addnodes.production(rule)
             subnode['tokenname'] = name.strip()
             if subnode['tokenname']:
-                idname = nodes.make_id('grammar-token-%s' % subnode['tokenname'])
+                idname = 'grammar-token2-%s_%s' \
```
tk0miya: I hesitate to add "2" suffix for new styled node_id. I feel it is a bit ugly.

CHANGES
```diff
@@ -19,6 +19,9 @@ Incompatible changes
 * #6830: py domain: ``meta`` fields in info-field-list becomes reserved.  They
   are not displayed on output document now
 * The structure of ``sphinx.events.EventManager.listeners`` has changed
+* Due to the scoping changes for :rst:dir:`productionlist` some uses of
+  :rst:role:`token` must be modified to include the scope which was previously
+  ignored.
```
tk0miya, line 24: I feel this is too big incompatible change. How about adding warnings?

On my local, this works fine.
```
diff --git a/sphinx/domains/std.py b/sphinx/domains/std.py
index 725b879a9..6f61e877a 100644
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -725,6 +725,8 @@ class StandardDomain(Domain):
                           RemovedInSphinx40Warning)
             domain = env.get_domain('citation')
             return domain.resolve_xref(env, fromdocname, builder, typ, target, node, contnode)
+        elif typ == 'token':
+            resolver = self._resolve_token_xref
         else:
             resolver = self._resolve_obj_xref

@@ -859,6 +861,27 @@ class StandardDomain(Domain):
         return make_refnode(builder, fromdocname, docname,
                             labelid, contnode)

+    def _resolve_token_xref(self, env: "BuildEnvironment", fromdocname: str,
+                            builder: "Builder", typ: str, target: str,
+                            node: pending_xref, contnode: Element) -> Element:
+        docname, labelid = '', ''
+        if ('token', target) in self.objects:
+            docname, labelid = self.objects['token', target]
+        elif ':' not in target:
+            # old styled reference
+            logger.warning('old styled token ref found. Please add a tokenGroup as a prefix',
+                           location=node)
+
+            target = ':' + target
+            for objtype, name in self.objects:
+                if objtype == 'token' and name.endswith(target):
+                    docname, labelid = self.objects[objtype, name]
+                    break
+
+        if not docname:
+            return None
+        return make_refnode(builder, fromdocname, docname, labelid, contnode)
+
     def _resolve_obj_xref(self, env: "BuildEnvironment", fromdocname: str,
                           builder: "Builder", typ: str, target: str,
                           node: pending_xref, contnode: Element) -> Element:
```

sphinx/domains/std.py
```diff
@@ -406,15 +406,17 @@ def run(self) -> List[Node]:
         return messages + [node]
 
 
-def token_xrefs(text: str) -> List[Node]:
+def token_xrefs(text: str, productionGroup: str) -> List[Node]:
+    if len(productionGroup) != 0:
+        productionGroup += ':'
```
tk0miya, line 411: Which is better `tokenGroup` or `productionGroup`?

tk0miya (changes requested):
Basically LGTM.

doc/usage/restructuredtext/directives.rst
jakobandersen: Fixed.

doc/usage/restructuredtext/directives.rst
jakobandersen: Fixed, was a stray edit.

sphinx/domains/std.py
jakobandersen: Good point, fixed.

sphinx/domains/std.py
jakobandersen: I don't know, but I am all for simplifying it if possible. The code was copied from ``ObjectDescription``.

tests/roots/test-productionlist/index.rst
jakobandersen: Fixed.

sphinx/domains/std.py
jakobandersen: That depends. If someone has copied a link to a rule, then it would be broken without this code. However, there are no explicit permalinks in production lists, link there is on the various domain directives. So I guess that technically there is no promise that such links will continue to work :-).

sphinx/domains/std.py
jakobandersen: It is indeed, and is actually not needed as the ``_`` is enough that IDs will not clash. I am added a comment in the code to explain the details.

CHANGES
jakobandersen: In nitpicky-mode there will already be a warning about a reference not being resolved.
Note that the *productionGroup* argument is not becoming required with this PR so ``:token:`something` `` is still perfectly valid with, e.g.:
```rst
... productionlist::
   something: a b c
```
That means that as long as authors did not provide a *productionGroup* then there should be no extra warnings for those ``:token:`` roles.

sphinx/domains/std.py
jakobandersen: Good point, changed to be *productionGroup* in the documentation.

sphinx/domains/std.py
tk0miya: It's not good to add not understood code. This code is strange to me. How about removal? It's not too late to add it after if needed.

sphinx/domains/std.py
tk0miya: Understood. Indeed, an URL of old `:token:` role causes an error. +0 for keeping this. My large concern is that codes for compatibility makes our code complex. So it would be better to remove this in future. I've sometime confused by "historical" code...

CHANGES
tk0miya: Thank you for explanation. Okay, go with this.

tk0miya (approved):

sphinx/domains/std.py
jakobandersen: Indeed, not need to introduce it when it wasn't there before. It has been removed.

sphinx/domains/std.py
jakobandersen: Indeed, some cleanup would be nice. Though I'm not sure what the best approach is so I'll let the old ID be there for now.

tk0miya:
LGTM. Thank you always!

PR #7057, opened 2020-01-23
Close #7051: autodoc: Support instance variables without defaults (PEP-526)

### LINKED ISSUE #7051: Support instance variables without defaults
[PEP 526](https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations) allows to define type annotations for class/instance variables. Sphinx supports vars with default values, but not without.

Consider the following example from PEP 526, enhanced with a docstring:

```python
# pep526.py

from typing import ClassVar, Dict

class BasicStarship:
    """
    An example from :pep:`526`.

    :ivar str captain: instance variable with default
    :ivar int damage: instance variable without default
    :cvar stats: class variable
    :vartype stats: dict[str, int]
    """
    captain: str = 'Picard'               # instance variable with default
    damage: int                           # instance variable without default
    stats: ClassVar[Dict[str, int]] = {}  # class variable
```
Generating docs from

```rst
``pep526`` Example
==================

.. automodule:: pep526
   :members:
   :undoc-members:
   :show-inheritance:
```

yields a warning in the terminal:

```
pep526.py:docstring of pep526.BasicStarship:: WARNING: py:obj reference target not found: damage
```

and renders the docs like this:

image

The request is thus to treat an instance var without default value same way as a regular instance var, if it's possible at all. I'm fine with the rendered result, but we rely heavily on type hints and I have to put each warning to `nitpick_ignore` as our CI pipeline builds the docs with `-Wn`. Another great disadvantage is that the generated inventory doesn't contain the attributes. In the above example, the inventory would only have

```
py:attribute
	pep526.BasicStarship.captain             api/pep526.html#pep526.BasicStarship.captain
	pep526.BasicStarship.stats               api/pep526.html#pep526.BasicStarship.stats
```
so no crossref'ing to `damage` attribute is possible.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #7051


tests/test_autodoc.py
```diff
@@ -1388,6 +1388,46 @@ def test_partialmethod_undoc_members(app):
...
+        '   ',
+        '',
+        '.. py:data:: attr1',
+        '   :module: target.typed_vars',
+        "   :annotation: = ''",
+        '',
+        '   attr1',
+        '   ',
+        '',
+        '.. py:attribute:: attr2',
```
tk0miya: This should be `py:data`.

sphinx/ext/autodoc/__init__.py
```diff
@@ -24,17 +24,15 @@
...
+from sphinx.ext.autodoc.importer import import_object, get_module_members, get_object_members
 from sphinx.ext.autodoc.mock import mock
 from sphinx.locale import _, __
 from sphinx.pycode import ModuleAnalyzer, PycodeError
 from sphinx.util import inspect
 from sphinx.util import logging
 from sphinx.util import rpartition
 from sphinx.util.docstrings import prepare_docstring
-from sphinx.util.inspect import (
-    getdoc, object_description, safe_getattr, safe_getmembers, stringify_signature
```
tk0miya, line 36: `safe_getmembers` is no longer needed.

PR #7062, opened 2020-01-25
Close #7032: html_scaled_image_link is disabled for individual image

### LINKED ISSUE #7032: html_scaled_image_link behavior is too restrictive for extension developers
I'm  developing sphinx-extensions and my code is using ``nodes.image`` from docutils to add images to the documentation which shall be used as buttons, status or other indicators.

But Sphinx is surrounding my images with ``nodes.reference``, so that they become clickable links, which open the image in the browser.

This is not the desired behavior, as I bind the images to javascrip (html only) to use them as buttons for specific actions.

**Problem**
https://github.com/sphinx-doc/sphinx/blob/cad1f86dd895836e36e6a57407a5d82c990dfaa8/sphinx/builders/html.py#L808

The handling of ``html_scaled_image_link`` does only look on the global config, but not on the image node itself, if maybe some attributes are set to avoid adding of references.

So I can only switch off the behavior globally, which for sure is not an option.

**Solution / Proposal**
I propose the usage of a ``target`` attribute of the image node.
If this is set, the target-value is used as refuri.
If it is set to None but part of node.attributes-dict, no reference should be added.
If it is not set (which is currently always the case), normal handling as of now.

As far as I have understood the docutils code, the handlers of images-nodes do not care about a ``target`` attribute.

**Workaround**
Luckily I found on, but it is very ugly.

First off all, you need to add the image-node to a reference-node.
In this case, the ``html_scaled_image_link`` handling is aborted, as it stops if parent of an image-node is already a reference-node.

But now you have a not wanted reference in your html-code, which forces the browser to jump/scroll the page when clicked on it.
So I must use some javascript to avoid this :(

Here the code for the important parts:
```
# Wrap image-node inside a reference node
...
ref_node = nodes.reference('', '', refuri='#', classes=['no_link'])  # Set a specific class for later js handling
ref_node.append(image_node)
return ref_node
```

A loaded js-file needs to have this in it:
```
// JQuery is used here
$(document).ready(function() {
    $('a.no_link').click(function (e) {
        e.preventDefault();
    });
});
```

**So, what do you think?**.
I feel not really happy with the proposed solution.
Maybe someone else has a better idea, how to "deactivate" this behavior for sphinx extension developers.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #7032


CHANGES
```diff
@@ -51,6 +51,8 @@ Features added
   in the documentation.
 * #1027: Support backslash line continuation in :rst:dir:`productionlist`.
 * #7108: config: Allow to show an error message from conf.py via ``ConfigError``
+* #7032: html: :confval:`html_scaled_image_link` is disabled for images having
```
shimizukawa: nits
```suggestion
* #7032: html: :confval:`html_scaled_image_link` will be disabled for images having
```

shimizukawa (commented):
Q. Is this feature valid for figure directives?

tk0miya:
> Q. Is this feature valid for figure directives?

Yes, it also works for figures.

doc/usage/configuration.rst
```diff
@@ -1357,8 +1357,21 @@ that use Sphinx's HTMLWriter class.
...
+
+   .. code-block:: rst
+
+      .. image:: sphinx.png
+         :scale: 50%
+         :class: no-scaled-link
+
    .. versionadded:: 1.3
 
+   .. versionchanged:: 2.4
```
mgeier, line 1371: Shouldn't this be 3.0?

tk0miya: Right. Will be fixed in #7647

PR #7063, opened 2020-01-25
C++, fix some cross referencing issues

### LINKED ISSUE #5078: C++, references to parameters in compound directives
ping @jakobandersen 

By compound directives I mean an object used with multiple arguments:

```rST
.. cpp:concept:: template<typename ParamA> A
                 template<typename ParamA> B
                 template<typename ParamB> X
                 template<typename ParamB> Y

    References to parameters: :any:`ParamA`, :any:`ParamB`.

    References to parameters: :cpp:any:`ParamA`, :cpp:any:`ParamB`.

.. cpp:function:: void a(int param_a)
                  void b(int param_a)
                  void x(int param_b)
                  void y(int param_b)

    .. NYI on base :any: role?
    .. Reference to parameters: :any:`param_a`, :any:`param_b`.

    References to parameters: :cpp:any:`param_a`, :cpp:any:`param_b`.
```

While it's convenient for attaching a description to multiple related items at once, it does not play nice with references to parameters: they will only resolve to the last introduced item (i.e. `Y` and `y` respectively).

To be clear I don't think it's a problem that e.g. `ParamB` references resolve to the parameter of `Y` and not that of `X`. It's a problem that `ParamA` and `param_a` don't resolve to anything.

As a footnote, the base `any` role warns about not finding a target for `ParamA` while `cpp:any` is silent, which I think is intended and desirable? `any` crashes for `param_a` though.

Separating the items into multiple directives does not really act as a workaround: references to `ParamA` won't link if the description is attached to `X`/`Y` and vice versa for references to `ParamB` attached to `A`/`B`.

jakobandersen:
### Feature or Bugfix
- Bugfix

### Detail
When a cross-referencing role is used it must know which scope to start lookup in. This scope specification did not take function overloads into account meaning that it references to template and function parameters could reported as not found, if the wrong starting point happened to be selected.

When more than one declaration is written in a directive it should be possible to reference any template and function paramters, even though the starting point for lookup technically is a just one of the declarations. Now references will select the first match starting from the last declaration.

See the two new test files for specific examples.

### Relates
- Fixes #5078

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -61,7 +61,7 @@ default_settings = {
 
 # This is increased every time an environment attribute is added
 # or changed to properly invalidate pickle files.
-ENV_VERSION = 57
+ENV_VERSION = 56
 
 # config status
 CONFIG_OK = 1
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -61,7 +61,7 @@ default_settings = {
 
 # This is increased every time an environment attribute is added
 # or changed to properly invalidate pickle files.
-ENV_VERSION = 56
+ENV_VERSION = 57
 
 # config status
 CONFIG_OK = 1
--- a/CHANGES
+++ b/CHANGES
@@ -43,7 +43,10 @@ Features added
 Bugs fixed
 ----------
 
-* C++, fix cross reference lookup in certain cases involving function overloads.
+* C++, fix cross reference lookup in certain cases involving
+  function overloads.
+* #5078: C++, fix cross reference lookup when a directive contains multiple
+  declarations.
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -43,6 +43,8 @@ Features added
 Bugs fixed
 ----------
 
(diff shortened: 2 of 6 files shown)


jakobandersen:
@tk0miya, can you confirm that the change to the environment version is correct? It is due to a change in attached data on the nodes created from C++ roles.

sphinx/environment/__init__.py
```diff
@@ -61,7 +61,7 @@
 
 # This is increased every time an environment attribute is added
 # or changed to properly invalidate pickle files.
-ENV_VERSION = 56
+ENV_VERSION = 57
```
tk0miya: Instead of this, please update `env_version` field which `sphinx.domains.cpp:setup()` returns. It describes the version of domain data structure now.

tk0miya (commented):
I don't read the changes in `sphinx.domains.cpp`. But you can always update env_version when you changed internal data structure of CppDomain (a.k.a. `self.data`).

sphinx/environment/__init__.py
jakobandersen: Thanks!

PR #7064, opened 2020-01-25
Close #7025: html search: full text search can be disabled via metadata

### LINKED ISSUE #7025: Make it possible to ignore certain files when searching
We use sphinx for both code documentation and user documentation. As the code documentation is used way less then the user documentation, especially when searching, we would like to hide useless results for many of our users, i.e. the code documentation. 

I would like a new configuration option that would allow me to ignore certain files when creating the search index.

I could subclass the search implementation and implement this manually, but I think a configuration option might also be useful.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #7025

### DIFF
--- a/doc/usage/restructuredtext/field-lists.rst
+++ b/doc/usage/restructuredtext/field-lists.rst
@@ -51,3 +51,12 @@ At the moment, these metadata fields are recognized:
        :orphan:
 
    .. versionadded:: 1.0
+
+``nosearch``
+   If set, full text search for this file is disabled. ::
+
+       :nosearch:
+
+   .. note:: object search is still available even if `nosearch` option is set.
+
+   .. versionadded:: 2.4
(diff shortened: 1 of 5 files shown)


doc/usage/restructuredtext/field-lists.rst
```diff
@@ -45,3 +45,12 @@ At the moment, these metadata fields are recognized:
        :orphan:
 
    .. versionadded:: 1.0
+
+``nosearch``
```
tk0miya, line 55: I have another candidate: `:search: no`. But I chose `nosearch` because it is similar to existing metadata: `nocomments`.

shimizukawa: +1 for `nosearch`

shimizukawa (approved):
LGTM!

PR #6320, opened 2019-04-23
Support property (autodoc, py-domain)

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add `:property:` option to `py:method` directive
- autodoc supports property detection

PR #7068, opened 2020-01-27
Document properties as attributes

### LINKED ISSUE #6785: PropertyDocumenter causes inconsistent handling of Python descriptors

### LINKED ISSUE #7183: :attr: links to intersphinx-reachable properties are still broken
**Describe the bug**
#7061 contained a workaround to unbreak the bug where `:attr:` wasn’t able to refer to properties.

unfortunately the workaround doesn’t address intersphinx links, as intersphinx relies on the `missing-reference` hook to resolve references. The `resolve_xref` function containing the workaround always returns `None` for intersphinx links, and intersphinx’ `missing-reference` hook doesn’t contain a workaround.

#7068 is the correct fix for this problem that makes further workarounds unnecessary.

**To Reproduce**

```bash
echo >conf.py <<EOF
extensions = ["sphinx.ext.intersphinx"]
intersphinx_mapping = dict(pandas=("https://pandas.pydata.org/pandas-docs/stable/", None))
EOF
echo ':attr:`pandas.DataFrame.iloc`' >index.rst
sphinx build . html
```

**Expected behavior**
- `:attr:` links to properties work with intersphinx as well (fixed by #7068, could also be fixed by another workaround)
- `objects.inv` contains entries for properties under `py:attribute` (fixed by #7068)

**Environment info**
- OS: All
- Python version: 3.8
- Sphinx version: 2.4.1
- Sphinx extensions: sphinx.ext.intersphinx

flying-sheep:
Subject: Properties are more like attributes than they’re like methods.

### Feature or Bugfix
- Bugfix

### Purpose
Fixes regression introduced by #6320 which resulted in properties being filed in as `std:method` instead of `std:attribute` and therefore people’s `:py:attr:` links to properties breaking.

### Detail
A method is a descriptor that returns a bound method on `__get__` (a bound method is always callable, with the first argument fixed to the instance).

A property is a descriptor that directly returns a value that’s usually not callable on `__get__`. 

Therefore it makes no sense to view properties as “methods”. Besides, as said, #6320 broke links in many many projects.

You already list them as attributes in `generate_autosummary_content` too.

### Relates
- Fixes #6785
- Fixes #7183


tests/test_domain_py.py
```diff
@@ -398,21 +394,12 @@ def test_pymethod_options(app):
     assert 'Class.meth4' in domain.objects
     assert domain.objects['Class.meth4'] == ('index', 'method')
 
-    # :property:
-    assert_node(doctree[1][1][8], addnodes.index,
-                entries=[('single', 'meth5() (Class property)', 'Class.meth5', '', None)])
```
flying-sheep: This demonstrates what I mean: Why are there parentheses after `meth5`? `instance.meth5` wouldn’t be callable! Of course you could’ve added an exception to not render the parentheses, but there’s a perfectly good `PyAttribute` that doesn’t render those parentheses.

sphinx/domains/python.py
```diff
@@ -552,15 +552,11 @@ class PyMethod(PyObject):
...
-        'property': directives.flag,
         'staticmethod': directives.flag,
     })
 
     def needs_arglist(self) -> bool:
-        if 'property' in self.options:
-            return False
-        else:
-            return True
+        return True
```
flying-sheep: Another exception that I was able to remove!

flying-sheep:
Can we please pull this soon and make a release? Things are already breaking down everywhere, e.g. our links to pandas are breaking because their objects.inv now has properties in the wrong section…

sphinx/ext/autodoc/__init__.py
```diff
@@ -1449,7 +1449,7 @@ def add_directive_header(self, sig: str) -> None:
         super().add_directive_header(sig)
         sourcename = self.get_sourcename()
         if inspect.isabstractmethod(self.object):
-            self.add_line('   :abstractmethod:', sourcename)
+            self.add_line('   :abstract:', sourcename)
```
flying-sheep: Since properties are not methods, having the following doesn’t make sense:

```restructuredtext
.. attribute:: thing
   :abstractmethod:
```

The correct way to create an abstract property is of course still

```py
@property
@abstractmethod
def thing(self):
   ...
```

but I see it this way:

1. `def thing(self):` without decorators is a method
2. adding `@abstractmethod` makes a method into an abstract method
3. adding `@property` makes a method into a property (and therefore an abstract method into an abstract property)

Of course I’d also have called `.. method::`’s option `:abstract:` instead of `:abstractmethod:`, but it’s too late for that.

flying-sheep:
Hi @tk0miya, my links are degrading more and more as other projects start compiling their docs with the buggy release. Could you please merge this soon and make a release?

sphinx/domains/python.py
```diff
@@ -552,15 +552,11 @@ class PyMethod(PyObject):
         'abstractmethod': directives.flag,
         'async': directives.flag,
         'classmethod': directives.flag,
-        'property': directives.flag,
```
tk0miya, line 625: This is a breaking change. Deprecation period is needed even if remove this.

sphinx/ext/autodoc/__init__.py
tk0miya: -1: It's inconsistent. If we choose the way to options describes by its character, we should re-design all of options like `:staticmethod:`, `:abstractmethod:`, `:classmethod:` and so on. This is one of reason why we chose to define properties by "how to defined". I believe this PR is a changing of policy. We have to design whole of designs.

BTW, we use `@abstractmethod` to define abstract objects in python. And we use `:abstractmethod:` for methods. On the other hand, we should use `:abstract:` only for properties. It is difficult to learn how to mark up.

tk0miya: Can we use only `abstract` option without `property`?

tk0miya:
@flying-sheep I think this changes policy of python domain. So we have to design how to mark up whole of python object. So I can't merge this soon. I believe #7061 is a workaround to avoid the broken references.

sphinx/ext/autodoc/__init__.py
flying-sheep: > Can we use only `abstract` option without `property`?

I think so:

```py
class AbstractFruit(ABC):
    color: Color

class Orange(AbstractFruit):
    color = Color.orange

class Apple(AbstractFruit):
	def __init__(self, ripe: bool):
        self.ripe = ripe
    @property
    def color(self):
        return Color.red if self.ripe else Color.green
```

- `AbstractFruit` has an abstract attribute
- `Orange` overrides it with a concrete attribute
- `Apple` overrides it with a concrete property

sphinx/domains/python.py
flying-sheep: How to deal with it? Leave it in and raise a warning if it’s used?

If so, which warning class?

flying-sheep:
OK, no problem, #7061 circumvents the problem well enough for the time being.

I’m happy to work on this PR in the meantime until it fullfills your requirements :smiley:

sphinx/ext/autodoc/__init__.py
flying-sheep: > If we choose the way to options describes by its character, we should re-design all of options like :staticmethod:, :abstractmethod:, :classmethod: and so on.

`:abstract:` could just become an alias for `:abstractmethod:` (make both work for methods). Or we keep `:abstractmethod:` and call the version for `.. attribute::` `:abstractattribute:` or similar.

tk0miya: I'll call `AbstractFruit.color` as "an instance attribute of the abstract class" if I need to mention to it. But I understand it is an abstract attribute. So it will make a sense.

>:abstract: could just become an alias for :abstractmethod: (make both work for methods). Or we keep :abstractmethod: and call the version for .. attribute:: :abstractattribute: or similar.

If we choose former one, `:class:` and `:static:` are also needed, right?  But we already use `:class:` for styling. So they're conflicted...

sphinx/domains/python.py
tk0miya: Yes, a warning will be needed for users. Please emit it by `logger.warning()`.

flying-sheep: done

sphinx/ext/autodoc/__init__.py
flying-sheep: We could just allow `:abstract:` as well as `:abstractmethod:` for methods. That way nobody is confused that one or the other doesn’t work.

tk0miya: I prefer "Zen of Python". So there should be one obvious way to do it. Of course, we need to provide an old options to keep compatibility. But the ideal option should be one. So we should choose a new option from these candidates. -1 for adding both.

If we choose `:abstract:` for attributes, other options will be also changed to describe its behavior. So methods and attributes will be marked up as following:
```
.. py:method:: hello()
   :abstract:
   :async:
   :static:
   :classmethod:

   Note: ``:class:`` is not able to use to describe classmethod
   because we have already have ``:class:`` option for commonly used options in reST directives.

.. py:attribute:: name
   :abstract:
   :async:
   :property:
   :type:
   :value:
```

If we choose `:abstractattribute:`, they are marked up as following:
```
.. py:method:: hello()
   :abstractmethod:
   :async:
   :staticmethod:
   :classmethod:

.. py:attribute:: name
   :abstractattribute:
   :async:
   :property:
   :type:
   :value:
```
Is my understanding correct?

I feel former one is not consistent because of `:classmethod:`. On the other hand, latter one is a bit non natural to me because `:abstractattribute:` is not a keyword of python.

tk0miya: Just idea, it might be better to add a new directive `py:property`. It helps to separate options from `py:attribute`:
```
.. py:attribute:: name
   :type:
   :value:

.. py:property:: name
   :abstractmethod:
   :async:
   :type:
```

shimizukawa: IMO, as a document writer, it's easy to use the terms in Python itself. Otherwise, if sphinx require names like `abstract`,` static`, `class`, etc., document writers will need to remember the names for writing documentation in Sphinx separately from the Python terminology. For this reason, `:abstractmethod:` is better than the common option `:abstract:`.

By the way, your `AbstractFruit` is instanciable...
```
>>> class AbstractFruit(ABC):
...     color: Color
...
>>> af = AbstractFruit()
>>> af
<__main__.AbstractFruit object at 0x0000014C8D051588>
```
If we want to create such class, it is:
```
>>> class AbstractFruit(ABC):
...     @property
...     @abstractmethod
...     def color(self) -> Color:
...         pass
...
>>> af = AbstractFruit()
TypeError: Can't instantiate abstract class AbstractFruit with abstract methods color
```
I think this `color` attribute is easy to understand in the documentation as follows:
```
.. py:method:: color
  :abstractmethod:
  :property:

  describing string

  :return: color value
  :rtype: Color
```

To be honest, it would be nice to be able to write short as:
```
.. py:method:: color
  :abstractmethod:
  :property:
  :type: Color
```

flying-sheep:
Let’s continue the discussion here. I agree about `:abstractproperty:` being weird, so I think we have three options, right? **A)** simplicity **B)** familiarity **C)** Just **B** but with `py:property`

I’m not sure what we should go with, I think the first two of them have balanced pros and cons:

**A)** IMHO the following would be the best and simplest option *if sphinx was designed that way from the start*. It just pains me to see `:XXXmethod:` thrice and then `:async:`. So inconsistent! We already have the directives `py:staticmethod` and `py:classmethod`, so if we change a lot anyway, we can deprecate the corresponding options in favor of those two directives. Since instance methods, class methods, and static methods are mutually exclusive, this would make things much simpler!

But it would confuse the hell out of people accustomed to the old way of doing it! On the other hand just deprecating the old way for years to come is possible.

```restructuredtext
.. py:method:: hello()
   :abstract:
   :async:
   :return:
   :rtype:
.. py:classmethod:: hello()
   ...
.. py:staticmethod:: hello()
   ...

.. py:attribute:: name
   :abstract:
   :property:
   :type:
   :value:
```

**B)** If we decide to stay with `:abstractmethod:` and stay with `py:attribute`, we have a good package with little change: Everything but `:abstractmethod:` reflects how things are used and what all options that are introduced by a decorator match the decorator’s name. I think we gain little by going with a dedicated `py:property` if we also stay with `:abstractmethod:`. Properties and attributes are too similar in practice and share all other options.

```restructuredtext
.. py:method:: hello()
   :abstractmethod:
   :staticmethod:
   :classmethod:
   :async:
   :return:
   :rtype:

.. py:attribute:: name
   :abstractmethod:
   :property:
   :type:
   :value:
```

sphinx/ext/autodoc/__init__.py
flying-sheep: Let’s continue discussing below, OK? @shimizukawa I understand what you mean about mirroring the code. Since `.. py:method:: […] :property:` doesn’t make sense though, I assume you’d vote for the following? I filed that as option **B** below.

```restructuredtext
.. py:attribute:: color
  :abstractmethod:
  :property:
```

tk0miya:
At this moment, I vote to plan C.

## About A)

>It just pains me to see :XXXmethod: thrice and then :async:. So inconsistent!

I don't think so. We usually see XXXmethod decorators and asyc keyword in python code. I believe writer of python docs must be a python programmer. So it is not inconsistent with python.

To me, a characteristic of plan A is to use simple keywords to describe their behavior: abstract, async and etc. But they are different with keywords in python, Sphinx originals. IMO, this plan is not easy to understand. I prefer to learn less keywords. This is why I hesitate to choose plan A.

## About B)

>Everything but :abstractmethod: reflects how things are used and what all options that are introduced by a decorator match the decorator’s name.

I think abstractmethod is also same as python's decorator. AFAIK, we usually use [abc.abstractmethod](https://docs.python.org/3/library/abc.html#abc.abstractmethod) decorator to make an abstract method, right?

>Properties and attributes are too similar in practice and share all other options.

I think they can take different options. A property can take `abstract` (or `abstractmethod`), `async` and `type` options. On the other hand, an attribute can take only `type` and `value` options (It would be able to take `:instance:` option to describe it is an instance attribute in future). This is why I proposed `py:property`.

| | property | attribute | notes |
| --- | --- | --- | --- |
| async | Y | N | Accurately, this is not allowed in official syntax. We can do it with [async-property](https://pypi.org/project/async-property/) |
| abstract | Y | N | |
| type | Y | Y | |
| value | N | Y | |
| instance? | N | Y | Not implemented yet. But there are some requests to describe class attributes and instance attributes. |

flying-sheep:
One counter argument is that properties *must* end up as `py:attribute` in `objects.inv`:

Say people have links like ``:attr:`mod_a.TestClass.test` ``. `mod_a`’s author decides to convert it from attribute to property. This shouldn’t have any impact on `objects.inv` and therefore the fact that people are referring to it via `:attr:` links.

And if the correct way to refer to attributes and properties is via `:attr:`, I think having `py:property` and `py:attribute` is a bit confusing: People will try to refer to them via `:prop:` or so which shouldn’t and won’t ever exist.

tk0miya:
>One counter argument is that properties must end up as py:attribute in objects.inv:

Why? If you worry about referencing role, type in object.inv is not related with it. As we did in 2.4 release, we can refer `py:attribute` entry via `:py:attr:` role (with some hacks). I don't think its structure is not a blocker for moving `py:property`.

I think providing new role `:py:prop:` as a formal way to refer properties. And `:py:attr:` is also allowed to refer properties as in the past. It may seem inconsistent to users, but I think it is useful.

(Note: please post a new comment if you rewrite it. Updates are not emailed.)

flying-sheep:
As said: Moving an API from property to attribute or the other way is a transparent act, that consumers of the API won’t notice. If Sphinx makes links break or behave in any way different beween attributes and properties, then Sphinx breaks this guarantee. Therefore `:prop:` shouldn’t exist.

sphinx/ext/autodoc/__init__.py
shimizukawa: > Let’s continue discussing below, OK? 

sure :)

tk0miya:
Sorry, I'm confused. Adding `:prop:` role does not break links of `:attr:` role. I think there are no breaking changes for users. Users can continue to use `:attr:` role to refer both attributes and properties. What problem do you think?

flying-sheep:
OK, so you basically just want to introduce the `prop` role as an individually stylable link to attributes and properties – in case someone wants to style links to properties differently than links to attributes?

flying-sheep:
Okay, done!

sphinx/domains/python.py
```diff
@@ -571,15 +571,12 @@ class PyMethod(PyObject):
...
+        'property': directives.flag,  # deprecated
         'staticmethod': directives.flag,
     })
 
     def needs_arglist(self) -> bool:
-        if 'property' in self.options:
-            return False
-        else:
-            return True
+        return True
```
tk0miya: Please don't remove codes for properties from PyMethod. This breaks built document. It should only emit a warning during deprecation period.

sphinx/domains/python.py
```diff
@@ -679,7 +679,29 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
             else:
                 return name
 
-        return _('%s (%s attribute)') % (attrname, clsname)
+        return self.fmt % (attrname, clsname)
+
+
+class PyProperty(PyAttribute):
```
tk0miya: Please inherit `PyObject` instead. PyProperty is not a kind of PyAttribute. For example, it does not support `value` option.

sphinx/domains/python.py
```diff
@@ -679,7 +679,29 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
...
+
+    option_spec = PyAttribute.option_spec.copy()
+    option_spec.update({
+        'abstractmethod': directives.flag,
+    })
+    fmt = _('%s (%s property)')
+
+    def run(self) -> List[Node]:
+        # properties and attributes need to be transparently switchable:
+        # making one into the other doesn’t break APIs so it shouldn’t break doc links.
```
tk0miya: What API you meant? -1 for replacing.

tk0miya (commented):
Thank you for update. But it seems there are misunderstanding between us... I still don't understand what you meant API is.

sphinx/domains/python.py
flying-sheep: That’s the main point of the whole PR, and I explained it several times so far.

1. Developers write a class with an attribute: `Cls.test`
2. Consumers start relying on `Cls().test` in their code and create intersphinx links using ``:attr:`mod.Cls.test` ``.
3. The developers then change `test` to a property. The consumer code won’t need to change, it’s completely transparent: `Cls().test` will evaluate to a value with the documented type no matter if it’s a property or an attribute.
   
   **The same should apply to Sphinx**. The intersphinx link created using `:attr:` should just continue to work and look the same. The consumer shouldn’t have to know or care if the thing they link to is an attribute or a property.

Therefore the `objects.inv` category and link role for attribute and class should be the same. Since the `py:attribute` category and `attr` role already exist, I chose to use those (instead of e.g. creating a new `py:attrprop` category or so for both).

sphinx/domains/python.py
flying-sheep: Done!

sphinx/domains/python.py
flying-sheep: Okay, makes sense!

flying-sheep:
The remaining mypy failures are not mine. No idea why they are there.

sphinx/domains/python.py
tk0miya: As I commented in this PR several times so far, property is not a kind of attribute. So -1 to give `py:attribute` for properties. I think it is better to give a new category `py:property` to properties and make `:py:attr:` role referencible for it like `:py:meth:` does now.

I agreed with allowing to refer properties via `:py:attr:` role. But I did not say they are same as attribute. If so, no reason to add `py:property` directive.

flying-sheep: I’m OK with that, thank you for the clarification. done!

tk0miya:
Thank you for update. I think this update is large and complicated to implement. How about separate a PR to some parts; py domain, autodoc, napoloen and so on? I believe we reached new design of python domain (directives and roles). So it is not difficult to implement step by step.

flying-sheep:
The tests just failed because there were upstream changes, which I now fixed. Now all tests pass, so why split it up?

sphinx/domains/python.py
```diff
@@ -614,23 +615,37 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
             return ''
 
 
-class PyMethod(PyObject):
+class PyClassMemberBase(PyObject, ABC):
```
tk0miya, line 618: Why did you inherit ABC?

sphinx/domains/python.py
```diff
@@ -614,23 +615,37 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
...
+    def _split_member_name(self, mod: str, name: str) -> Tuple[Optional[str], str]:
```
tk0miya: I feel return value of this method is odd. It returns a pair of class name and member name or a pair of `None` and index-text name. It is too complicated to me. Latter pair would not be a result of "split member name"...

sphinx/domains/python.py
```diff
@@ -651,21 +670,12 @@ def get_signature_prefix(self, sig: str) -> str:
...
-            else:
-                return '%s()' % name
+        clsname, methname = self._split_member_name(modname, name_cls[0])
+        if clsname is None:
+            return methname
 
         if 'classmethod' in self.options:
             return _('%s() (%s class method)') % (methname, clsname)
-        elif 'property' in self.options:
-            return _('%s() (%s property)') % (methname, clsname)
```
tk0miya, line 668: Please keep as is. No reason to remove this.

sphinx/domains/python.py
```diff
@@ -985,6 +1018,7 @@ class PythonDomain(Domain):
         'classmethod':     PyClassMethod,
         'staticmethod':    PyStaticMethod,
         'attribute':       PyAttribute,
+        'property':        PyProperty,
```
tk0miya, line 1030: Note: I'll add `:prop:` role later.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1508,7 +1508,6 @@ def add_directive_header(self, sig: str) -> None:
         sourcename = self.get_sourcename()
         if inspect.isabstractmethod(self.object):
             self.add_line('   :abstractmethod:', sourcename)
-        self.add_line('   :property:', sourcename)
```
tk0miya, line 1511: It would be better to generate `:type:` option if possible.

sphinx/ext/napoleon/docstring.py
```diff
@@ -579,7 +579,14 @@ def _parse_attributes_section(self, section: str) -> List[str]:
                 if _type:
                     lines.append(':vartype %s: %s' % (_name, _type))
             else:
-                lines.extend(['.. attribute:: ' + _name, ''])
+                prop = getattr(self._obj, _name, None)
+                if isinstance(prop, property) and (
```
tk0miya: Please use `sphinx.util.inspect:isproperty()` instead. In this moment, it simply calls `isinstance()` as same as here. But it is easy to support complicated object (ex. decorated-partial properties) in future.

sphinx/ext/napoleon/docstring.py
```diff
@@ -579,7 +579,14 @@ def _parse_attributes_section(self, section: str) -> List[str]:
...
+                    prop.__doc__ is None or
+                    not prop.__doc__.startswith('Alias for field number ')
```
tk0miya, line 587: Where this string came from?

sphinx/ext/napoleon/docstring.py
```diff
@@ -998,7 +1005,9 @@ def push_item(name: str, rest: List[str]) -> None:
             'constant': 'const',
             'const': 'const',
             'attribute': 'attr',
-            'attr': 'attr'
+            'attr': 'attr',
+            'property': 'attr',
+            'prop': 'attr',
```
tk0miya, line 1012: Note: I'll replace this by prop role later.

tests/roots/test-root/objects.txt
```diff
@@ -56,6 +56,8 @@ Referring to :func:`nothing <>`.
 
    .. attribute:: attr
 
+   .. property:: prop
```
tk0miya, line 59: Could you move this to other testcase please? test-root is a too fat testcase and commonly used in multiple testcases. So we'd not like to add new data for it not to become slow our testing.

I guess you're using this data on `test_object_inventory` and `test_html_inventory_entries`. If so, they are not needed. Because object inventory and inventory database is almost same as `domain.objects` in `test_pyproperty`.

tests/test_domain_py.py
```diff
@@ -573,8 +560,7 @@ def test_pyattribute(app):
     assert_node(doctree, (addnodes.index,
                           [desc, ([desc_signature, ([desc_annotation, "class "],
                                                     [desc_name, "Class"])],
-                                  [desc_content, (addnodes.index,
-                                                  desc)])]))
+                                  [desc_content, (addnodes.index, desc)])]))
```
tk0miya: I guess this is unexpected change, right?

tk0miya (changes requested):
Oh, sorry. I'm okay if you're not confused.
It seems additional code is needed for referring properties by `:attr:` role.
see #7061 and #7187.

sphinx/domains/python.py
```diff
@@ -614,23 +615,37 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
...
+            elif self.needs_arglist():
+                return None, '%s()' % name
+            else:
+                return None, name
+
+
+class PyMethod(PyClassMemberBase):
     """Description of a method."""
 
     option_spec = PyObject.option_spec.copy()
```
tk0miya: I don't know why mypy warns this line. But it seems `PyObject.option_spec` is needed to annotated like `ObjectDescription.option_spec`:
https://github.com/sphinx-doc/sphinx/blob/372813a137cd40a13a3be55cd64cbb48c08c41c2/sphinx/directives/__init__.py#L63-L65

Could you try to add it please?

sphinx/domains/python.py
flying-sheep: Please don’t, properties and attributes should be linked with the same role, as the fact if something is a property or attribute is an implementation detail.

sphinx/ext/autodoc/__init__.py
flying-sheep: I don’t understand, what do you mean?

sphinx/ext/napoleon/docstring.py
flying-sheep: Namedtuple. In some Python versions its fields are attributes, in others properties.

This code unifies that.

sphinx/ext/napoleon/docstring.py
flying-sheep: as said above: please don’t.

tests/test_domain_py.py
flying-sheep: True, will change back

sphinx/domains/python.py
flying-sheep: OK, will do!

sphinx/domains/python.py
flying-sheep: This class just exists to abstract away some common behavior and not to be directly instantiated. IDK if mypy enforces ABC to be here, but it’s correct.

sphinx/domains/python.py
flying-sheep: The idea is that it either splits the name for further processing or returns a finished name. How to do that best? Return a `Union[str, Tuple[str, str]]`?

sphinx/ext/napoleon/docstring.py
flying-sheep: OK, will do!

tests/roots/test-root/objects.txt
flying-sheep: I think this is a premature optimization, but sure, why not.

sphinx/domains/python.py
flying-sheep: OK, makes sense!

sphinx/domains/python.py
tk0miya: Now we are going to add a new objtype "property" to describe properties. It means attributes and properties are not same thing. As discussed in many time, I consider they are similar, but different. So I feel it's natural to add a new role to refer "property" objects. The describing directive and referencing role is a pair.

At same time, as you said, attributes and properties are very similar. So `:py:attr:` role is also able to refer "property" objects. I agree it's natural. But it does not prevent to add `:py:prop:` role. User can use either `:py:attr:` or `:py:prop:` role to refer properties. It is not harmful to add new role for `:py:attr:` role users.

I believe this change does not affect for who uses `py:attribute::` directive and `:py:attr:` to describe and refer to properties. Even after autodoc generates `py:property` directive, users does not need to change their document. `:py:attr:` role will be able to refer properties until now, and from now on. I don't know why you object to add a new role. If you're saying attributes and properties are same, we have to discuss again from how we describe properties.

sphinx/domains/python.py
tk0miya: Understand. This is my first case to use ABC. Is it prevent to instantiate? Or a marker for developers not to instantiate directly?

sphinx/domains/python.py
tk0miya: I prefer to simple and stupid function. It means one function has one responsiblity. So it is good to me that this method only does splitting the name and subclasses process a finished name. I know some codes are duplicated in subclasses. But I prefer a simple code.

sphinx/ext/autodoc/__init__.py
tk0miya: It is better to generate `:type:` option if target property has an annotation for return value:
```
@property
def prop(self) -> int:
    """blah blah blah"""
```
This will converted into 
```
.. py:property:: prop
   :type: int

   blah blah blah
```

sphinx/ext/napoleon/docstring.py
tk0miya: Understand. I found an article for it.
https://chrisdown.name/2015/09/20/removing-namedtuple-docstrings-from-sphinx.html

+0 for supporting. Could you add a comment like `# namedtuple` please?

tests/roots/test-root/objects.txt
tk0miya: I guess test-root is built 20+ times in our testing. I don't know it is a premature optimization. I think separating is better, but it's okay to keep as is if you don't like it.

tk0miya:
Unfortunately, we'll call feature freeze this weekend. To integrate this to the next major release, I made a small PR #7298 to add `py:property` directive and role at first. It's almost as same as you're doing for py domain in this PR (I don't do refactoring in it because we can do it even after feature freeze). I'll post another PR for autodoc, napoleon, and intersphinx in the next step. Please review it if you have a time.
Sorry for not merging your code. But I believe it is based on your effort.

sphinx/domains/python.py
flying-sheep: they’re not the same thing to implement, which is why a separate directive makes sense.

they’re the same thing to access which is why only a single way to refer to them makes sense.

sphinx/domains/python.py
flying-sheep: It prevents instantiation, yes.

sphinx/domains/python.py
tk0miya: Okay, please implement as you like. I'll stop to maintenance python domain. It's yours. Bye.

sphinx/domains/python.py
flying-sheep: okay

sphinx/domains/python.py
flying-sheep: I don’t understand. Didn’t you just ask me to discuss this? You said:

> I don't know why you object to add a new role. If you're saying attributes and properties are same, we have to discuss again from how we describe properties.

My comment above was the answer to that:

- I think they **are not** the same on the implementer’s side (the one writing `@property\ndef foo(self):`/`.. property:: foo` **or** `self.foo = ...`/`.. attribute:: foo`)
- and I think they **are** the same on the user’s side (the one writing `instance.foo`/``:attr:`foo` ``)

sphinx/ext/autodoc/__init__.py
flying-sheep: OK, sure, thank you!

flying-sheep: Hmm, actually properties can have a different return type and “settable” type, e.g.:

```py
@property
def path(self) -> Path:
    return self._path

@path.setter
def path(self, value: typing.PathLike):
    self._path = Path(value)
```

I don’t know how to best represent this. For now just the return type might be enough.

sphinx/ext/napoleon/docstring.py
flying-sheep: sure!

tests/roots/test-root/objects.txt
flying-sheep: It already has 214 lines. I don’t think this one makes a difference.

sphinx/domains/python.py
tk0miya: I must apologize to you. I had been frustrated that night. Please forget my comment.

>and I think they are the same on the user’s side (the one writing 

Our opinions still don't match. I think they are different things for users.

* Some properties do not support writing (in other words: assigning or setting)
* Some properties do not support deleting
* Some properties have a side effect (ex. network access, database access, caching)

I understand properties behave like attributes transparently. So But I consider they are similar, but different things. That's why I added `py:property` directive (and why I allow being able to refer a property via `:py:attr:` role). If they are the same, I never add it to the python domain.

I believe there is no reason to add only `py:property` directive. It and `:py:prop:` role must be a pair. If you object to the role, you should object to the directive also.

flying-sheep: All good :smile: we all have those days. As a fellow maintainer I sure know how it can be. I hope things got better for you!

About the topic:

- I maintain that from an user perspective, they’re the same. `.. property::` is the library developer perspective and `:attr:` is the user perspective.
- It’s all semantics/naming. What if properties were called “property attributes”? E.g. `staticmethod`s really are functions, but they use `:meth:` and not e.g. `:sfunc:`.
- I don’t think `.. property::` needs to exist (`.. attribute::\n:property:` would be enough), but it’s OK if it exists, and doesn’t require that `:prop:` exists.

<details><summary>Here in long form:</summary>

What you describe is all behavior that the library developer wanted to attach to attribute getting/setting. Users just see an attribute that they can get and/or set via what’s called “*attribute* access syntax”. E.g. it could well be that a developer just switched from attribute to property to prevent setting an attribute (because users did it too often and it broke the code).

I don’t think that it’s more consistent to have roles for both compared to the rest of sphinx. There’s a lot of things that can be accessed through e.g. `:meth:` that’s not a instance method. E.g. `.. staticmethod::`s are functions in reality (as they don’t get bound to `self`), they just happen to be called “methods”. Just pretend “properties” were called “property attributes” or “custom attributes” and my thinking will make sense.

I’m not against `.. property::`, but just because I don’t think it matters if it’s `.. property::` or `.. attribute::\n:property:`. The latter would work for me, too, since  I don’t really share your counter argument, i.e. that `:value:` and `:property:` would be mutually exclusive. many directives have mutually exclusive options that one could incorrectly set together. E.g. `.. method::` has `:classmethod:` and `:staticmethod:` that don’t make sense to be both set at the same time.

</details>

tk0miya: >I’m not against .. property::, but just because I don’t think it matters if it’s .. property:: or .. attribute::\n:property:. 

Please let me know why the feature describing a property is needed.  I feel no reason to describe "this attribute is a property" if attributes and properties are the same for users. For library developers?

tk0miya: TBH, I'm very tired to discuss properties more. How about removing all supports for properties from python domain? That is nothing different from Sphinx-2.2. I can't understand what is correct.

flying-sheep: The only benefit of `.. property::` right now (compared to just `.. attribute::`) is that it supports `:abstractmethod:`. In the future we could also add support for different getter and setter types:

```py
@property
def p(self) -> Path:
    return self._p
@p.setter
def p(self, value: PathLike):
    self._p = Path(p)
```

<center>↓</center>

```restructuredtext
.. property:: p
   :type: ~pathlib.Path
   :setter-type: ~typing.PathLike
```

But anything is preferable to documenting properties using `.. method::`.

tk0miya: OMG... I've perfectly forgotten about `:abstractmethod:`. To describe such options, I also prefer `py:property` directive. But a discussion about `:py:prop:` role is not concluded yet...

flying-sheep:
So what’s left to do for me here?

tk0miya:
I don't have an idea of how to move this discussion forward. It's completely stopped.

flying-sheep:
Haha me neither. Let’s try this: My motivation is that when a developer changes attribute to a property or the other way round:

- autosummary should generate the correct directive, with `.. py:property::` displaying different information than `.. py:attribute::`
- the consumer/user shouldn’t have to change anything, and the links should look and work the same

Therefore I propose that

- properties and attributes should be under the same header when autosummary documents a class:

  > ### Attributes and properties
  > <table>
  > <tr><td><strong><code>some_prop</code></strong></td><td>Property documentation</td></tr>
  > <tr><td><strong><code>some_attr</code></strong></td><td>Attribute documentation</td></tr>
  > </table>

- The different link roles should have a purpose or be removed. If people can refer to the same thing interchangably via `:prop:`, `:meth:`, `:attr:` and maybe more, I don’t understand why we have them. Why not deprecate them all and just use `:obj:`? Am I missing something?

- Links to properties and attributes should render the same and be customizable together using the CSS class `py-attr`:

  - ``:obj:`Test.meth` `` → “**`Test.meth()`**” (with one of its CSS classes being `py-func`)
  - ``:obj:`Test.prop_or_attr` `` → “**`Test.prop_or_attr`**” (with both properties and attributes having the CSS class `py-attr`)

tk0miya:
>properties and attributes should be under the same header when autosummary documents a class:

I prefer your idea. It's nice for the default settings of autosummary. But it does not mean they are always grouped. It would be better that users can separate attributes and properties to different lists (using customized templates or other methods).

>The different link roles should have a purpose or be removed. If people can refer to the same thing interchangably via :prop:, :meth:, :attr: and maybe more, I don’t understand why we have them. Why not deprecate them all and just use :obj:? Am I missing something?

It's a difficult question. Surely, `:py:obj:` will match all kinds of objects; a class, a method, an attribute, a property and anything else. In my understanding, Sphinx provides cross-reference roles for each object type to refer to a python object explicitly. The idea might be based on Zen of python; it says "Explicit is better than implicit".  If not so, it is better to use `:any:` role for all references.

And I think directives and roles should be a pair (except `:py:obj:`). For example, the object described by `py:attribute::` directive can be referred by `:py:attr:` role. So I need to add `:py:prop:` role.

Note: As discussed above, I agree to refer properties via `:py:attr:` role.

>Links to properties and attributes should render the same and be customizable together using the CSS class py-attr:

-1. I don't think so because they are different things.

I think this is a root cause of this discussion. I consider they are different. And you consider they are same. I think both are correct. It's only different perspective. But it is difficult to get agreed.

flying-sheep:
> It's nice for the default settings of autosummary. […] It would be better that users can separate attributes and properties to different lists

great, let’s do it that way then!

> I think directives and roles should be a pair (except :py:obj:).

If we have a link role for most directives, it is only consistent to have one for every directive. I agree. My issue with having both is that if they’d do the same, they’d be purely cosmetic, see the next point:

> In my understanding, Sphinx provides cross-reference roles for each object type to refer to a python object explicitly. The idea might be based on Zen of python; it says "Explicit is better than implicit".

I think that’s a great principle, but I think it refers to things with actual semantics attached to it, i.e. different behavior. If two or more things do exactly the same, then another principle applies: “There should be one – and preferably only one – obvious way to do it”

> I consider they are different. And you consider they are same.

I don’t. They’re clearly different in implementation. I only consider them to be the same from a user/consumer perspective, i.e. for someone accessing an instance attribute/property (`test_instace.thing`) or linking to it (``:any:`Test.thing` ``).

Can we agree to give property links the `py-attr` *and* `py-prop` CSS classes? Then existing themes still work and people can still style them differently if they want.

If you really want both roles, there’s only one question left: should a `:attr:` link create a element with the `py-prop` class if the object is a property? It would make sense that the link reflects what it really links to if you really want the roles to be cosmetic.

tk0miya:
>>I consider they are different. And you consider they are same.
>
>I don’t. They’re clearly different in implementation. I only consider them to be the same from a user/consumer perspective, i.e. for someone accessing an instance attribute/property (test_instace.thing) or linking to it (:any:`Test.thing` ).

Sorry. It's my lack of words. But nobody object to they're same in implementation if they knows python. So this is a concern about documentation (in Sphinx). I believe they are different (but similar). And you're thinking they should be the same, right?

I think how to describe (author side) and how to represent (user/consumer side) is strongly coupled. So, based on user/consumer perspective you said, attributes and properties should be described as same in my idea. It means we should use `py:attribute::` directive to describe a propery. No `py:property` directive is needed.

>I think that’s a great principle, but I think it refers to things with actual semantics attached to it, i.e. different behavior. If two or more things do exactly the same, then another principle applies: “There should be one – and preferably only one – obvious way to do it”

The another principle does not match this case because they are different.

>Can we agree to give property links the py-attr and py-prop CSS classes? Then existing themes still work and people can still style them differently if they want.

Agreed. I think they are similar. So no reason to object that.

> If you really want both roles, there’s only one question left: should a :attr: link create a element with the py-prop class if the object is a property? It would make sense that the link reflects what it really links to if you really want the roles to be cosmetic.

Hmm... it's a difficult question. +0 for the idea if we can keep consistent. Now we can refer a class entry via `:py:attr:` role. But the reference is rendered as an element only with the py-attr. We need to refactor such (a bit broken) mechanism to realize the feature.

flying-sheep:
> So this is a concern about documentation (in Sphinx). I believe they are different (but similar). And you're thinking they should be the same, right?

Documentation *referring* to them should be the same because using them from code is the same and the implementation can be transparently switched at any moment (i.e. without breaking anyone’s code).

E.g.:

1. I use library `testlib` v1.0.0, which implements and documents a `class Test`.
2. I use the attribute `foo = testlib.Test().thing` in my code.
3. I link to the attribute via ``:attr:`Test.thing` `` in my code’s documentation
4. The author of the `Test` class switches the implementation to a property and releases `testlib` v1.0.1 (i.e. no breaking changes, whatever the property does won’t affect my code)
5. I expect my code to work exactly the same and I want my documentation to look and build exactly the same.

I don’t understand why in my use case my link should suddenly gain a `py-prop` class, or the “more correct” way to refer to `Test.thing` is now through `:prop:`, as nothing at all should change on my side.

tk0miya:
>I don’t understand why in my use case my link should suddenly gain a py-prop class, or the “more correct” way to refer to Test.thing is now through :prop:, as nothing at should change on my side.

What these expectations came from? About CSS rule was just proposed at above comment. Nothing determined. And who determines "more correct" way? I think both `:py:prop:` and `:py:attr:` are correct. There are no need to change their document that refers properties via `:py:attr:` role.

flying-sheep:
If there’s no correct way, there’s no reason to have two roles for the same job, no?

tk0miya:
My answer is always same. Because they are different.

shimizukawa:
It's been a month since I've commented. I've read the discussions over the past month and my opinion has not changed.
As a document author, it's more intuitive to use the terminology of Python itself. For this reason, it is not uncomfortable to represent a property as a kind of method. (This is an answer to https://github.com/sphinx-doc/sphinx/pull/7068#discussion_r382915885).

```
... py:method:: color
  :abstractmethod:
  :property:

  Describing string

  :return: color value
  :rtype: Color
```

While documentation should be written for the reader, I believe it is important to be able to directly describe the representation of the implementation in order to accurately convey the information that the reader expects. In that sense, it is not appropriate for a reader to see the property as `py:attribute` or its derivative. It is difficult for readers to understand that dynamic processing is performed when getting and assigning property values. If the implementation expression is `py:method:: :property:`, there is no difference between the document writer and the reader. In other words, I think we should treat ATTRIBUTE and PROPERTY as distinctly different things for the reader.

The problem of the above sample code is that it shows `()` because it treats color as a method, and this needs to be fixed.

I think merging the implementation of #7298 would "directly describe the representation of the implementation and allow the reader to get the information without misunderstanding". It also solves the problem of adding `()`, and you can use the `:abstractmethod:` option. The above sample code can now be described as follows
```
... py:property:: color
  :abstractmethod:
  :type: Color

  color value blah blah blah blah ...
```

My opinion summary: 

* We should treat ATTRIBUTE and PROPERTY as distinctly different things for the reader.
* let's merge #7298.

flying-sheep:
OK, I’ll review it.

PR #7072, opened 2020-01-30
py domain: Support type annotations for variables

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
This adds `:type:` and `:value:` options to both `py:data` and `py:attribute` directives.
It allows to describe its annotation in detail.

~~Note: This will make `:annotation:` option deprecated in future.~~

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -195,6 +195,18 @@ The following directives are provided for module and class contents:
    as "defined constants."  Class and object attributes are not documented
    using this environment.
 
+   .. rubric:: options
+
+   .. rst:directive:option:: type: type of the variable
+      :type: text
+
+      .. versionadded:: 2.4
+
+   .. rst:directive:option:: value: initial value of the variable
+      :type: text
+
+      .. versionadded:: 2.4
+
 .. rst:directive:: .. py:exception:: name
 
    Describes an exception class.  The signature can, but need not include
@@ -228,6 +240,18 @@ The following directives are provided for module and class contents:
    Describes an object data attribute.  The description should include
    information about the type of the data to be expected and whether it may be
    changed directly.
+
+   .. rubric:: options
+
+   .. rst:directive:option:: type: type of the attribute
+      :type: text
+
+      .. versionadded:: 2.4
+
+   .. rst:directive:option:: value: initial value of the attribute
+      :type: text
+
+      .. versionadded:: 2.4
 
 .. rst:directive:: .. py:method:: name(parameters)
 
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -446,6 +446,25 @@ class PyFunction(PyObject):
 class PyVariable(PyObject):
     """Description of a variable."""
 
+    option_spec = PyObject.option_spec.copy()
+    option_spec.update({
+        'type': directives.unchanged,
+        'value': directives.unchanged,
+    })
+
+    def handle_signature(self, sig: str, signode: desc_signature) -> Tuple[str, str]:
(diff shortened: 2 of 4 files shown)


sphinx/domains/python.py
```diff
@@ -446,6 +446,25 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
...
+        'type': directives.unchanged,
+        'value': directives.unchanged,
+    })
+
+    def handle_signature(self, sig: str, signode: desc_signature) -> Tuple[str, str]:
+        fullname, prefix = super().handle_signature(sig, signode)
+
+        typ = self.options.get('type')
+        if typ:
+            signode += addnodes.desc_annotation(typ, ': ' + typ)
```
tk0miya, line 460: I'd like to add a new node for this purpose. But it will bring incompatibility. So I keep it as is.
In another usage, we can split "type" value itself and markers like this:
```
signode += addnodes.desc_annotation(typ, typ, prefix=': ')
```
I think it is much better. But it is also incompatible change...

sphinx/domains/python.py
```diff
@@ -446,6 +446,25 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
 class PyVariable(PyObject):
     """Description of a variable."""
 
+    option_spec = PyObject.option_spec.copy()
+    option_spec.update({
+        'type': directives.unchanged,
+        'value': directives.unchanged,
+    })
```
tk0miya, line 453: Note: I did not make `:annotation:` option deprecated in this moment. Is it still needed? If not, I'll deprecate it soon.

tk0miya:
oh... I wrongly pushed this branch to sphinx-doc/sphinx temporarily. It causes CIs errored. Please ignore...

sphinx/domains/python.py
shimizukawa: I also think changing node structures and adding new nodes affect many 3rd party builders.
So IMO keeping compatibility is better.

If we would provide fall-back mechanism for such situation, we can changing/adding nodes.

sphinx/domains/python.py
shimizukawa: AFAIK, `:annotation:` option is for `autodata` and `autoattribute` directives:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

Ah, I found it.. https://github.com/sphinx-doc/sphinx/blame/edd5fcf8df066f9bd1afb98e2fab04db16ca6f54/sphinx/domains/python.py#L201
but it is [not documented](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html) I think.. 

IMO, it would be better if `annotation` will be a alias of `value` option.

shimizukawa (approved):
Basically, LGTM!

doc/usage/restructuredtext/domains.rst
```diff
@@ -195,6 +195,14 @@ The following directives are provided for module and class contents:
    as "defined constants."  Class and object attributes are not documented
    using this environment.
 
+   .. rubric:: options
+
+   .. rst:directive:option:: type: type of the variable
+      :type: text
+
+   .. rst:directive:option:: value: initial value of the variable
+      :type: text
```
tk0miya, line 206: We should add `versionadded` to here

sphinx/domains/python.py
tk0miya: Okay, let's keep as is.

sphinx/domains/python.py
tk0miya: Okay, let's keep as is.

Note: `:annotation:` expects the value starts with '=' mark. But `:value:` option does not. So it is not compatible.

Issue #1671, opened 2015-01-10
index feature and cross reference targets


mwiebusch78:
The documentation states that the .. index:: and :index: roles generate cross reference targets. The names of these targets seem to be something non-descriptive like index-0 etc. Is there a way to set the name of the cross reference targets, so that one can use them in :ref: roles?

The use case is the following: Let's say I define some abbreviation or technical term somewhere in my documentation. The abbreviation should appear in the index. But to make things simpler for the reader I also want to put a link to the definition in other places where I use the abbreviation. It seems at the moment the only way to do this is to set a custom label and an index entry. Would be nice if one could do it in one step.

birkenfeld:
Currently the anchors are only for use by the index. But your use case is valid. I'm just struggling a bit to find an acceptable syntax. It's also not clear what role should be used for referencing -- :ref: should be reserved for labels I think.

mwiebusch78:
But :ref: is used for all sorts of targets (sections, subsections, figures, etc.). Why not locations in the text where abbreviations are defined? My current workaround is to add a label as well as an .. index:: role.

```
.. index:: MyAbbreviation

.. _myabbr:

Definition of MyAbbreviation
```

A simple fix would be to change the .. index syntax so that you can put the label on the same line

```
.. index:: MyAbbreviation <myabbr>
```

or something similar. Of course the same should be possible with the inline :index:`...` role.

geographika:
Perhaps a new directive such as `:indexref:` could be used? There is a [:numref:](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-figures-by-figure-number) role that serves a similar purpose. 

I'm currently using hard-coded links for the MapServer docs e.g. https://mapserver.org/mapfile/class.html#index-1 which isn't ideal (as part of [Python API documentation](https://github.com/mapserver/mapserver/pull/5870)). The API docs make extensive use of an index with all keywords indexed such as [this example](https://github.com/mapserver/docs/blob/branch-7-4/en/mapfile/class.txt)

```
.. index::
   pair: CLASS; BACKGROUNDCOLOR
```

It would be great to be able to reference this location using syntax such as:

```
:indexref:`My index reference <CLASS; BACKGROUNDCOLOR>`
```

tk0miya:
We can put same index to multiple places. So `:indexref:` will be confused and not working well in such case. Of course, we can generate two hyperlinks for `:indexref:`. But I don't understand it is useful.

[...]

PR #7087, opened 2020-02-01
Add name option for the index directive

geographika:
Subject: To allow index hyperlinks to be constructed using a new `:name:` option. 
Feedback on approach welcome, and open to attempting to implement a different approach. 

### Feature or Bugfix
- Feature

### Purpose
The pull request allows the hrefs generated by the [Index Directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#index-generating-markup) to be given a name to be used when creating index links rather than simply an integer id that can change if text on a page is reordered. 

For example:

`.. index:: Test1` will generate `test.html#index-0`

Whereas supplying a new name option such as below:

```
.. index:: single: Test1
   :name: mylink
```

Will generate: `test.html#index-mylink`

If the same `name` option is used multiple times for index entries on the same page a warning similar to below is given:

`index.rst:17: WARNING: Duplicate ID: "index-mylink".`

### Detail
An alternative to simply modifying the href would be to construct a proper `:ref:` node. 
This could be achieved using `from sphinx.util.nodes import make_refnode`?

### Relates
-  #1671

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -885,7 +885,7 @@ mainly contained in information units, such as the language reference.
         .. index:: Python
            :name: py-index
 
-   .. versionadded:: 2.5
+   .. versionadded:: 3.0
 
 .. rst:role:: index
 
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -874,6 +874,19 @@ mainly contained in information units, such as the language reference.
    .. versionchanged:: 1.1
       Added ``see`` and ``seealso`` types, as well as marking main entries.
 
+   .. rubric:: options
+
+   .. rst:directive:option:: name: a label for hyperlink
+      :type: text
+
+      Define implicit target name that can be referenced by using
+      :rst:role:`ref`.  For example::
+
+        .. index:: Python
+           :name: py-index
+
+   .. versionadded:: 2.5
+
 .. rst:role:: index
 
    While the :rst:dir:`index` directive is a block-level markup and links to the
--- a/sphinx/domains/index.py
+++ b/sphinx/domains/index.py
@@ -77,8 +77,8 @@ class IndexDirective(SphinxDirective):
         arguments = self.arguments[0].split('\n')
 
         if 'name' in self.options:
-            targetid = self.options['name']
-            targetnode = nodes.target('', '', names=[targetid])
+            targetname  = self.options['name']
+            targetnode = nodes.target('', '', names=[targetname])
         else:
             targetid = 'index-%s' % self.env.new_serialno('index')
             targetnode = nodes.target('', '', ids=[targetid])
@@ -89,7 +89,7 @@ class IndexDirective(SphinxDirective):
         indexnode['inline'] = False
         self.set_source_info(indexnode)
         for entry in arguments:
-            indexnode['entries'].extend(process_index_entry(entry, targetid))
+            indexnode['entries'].extend(process_index_entry(entry, targetnode['ids'][0]))
         return [indexnode, targetnode]
 
 
--- a/sphinx/domains/index.py
+++ b/sphinx/domains/index.py
@@ -12,6 +12,7 @@ from typing import Any, Dict, Iterable, List, Tuple
 
 from docutils import nodes
 from docutils.nodes import Node, system_message
+from docutils.parsers.rst import directives
 
 from sphinx import addnodes
 from sphinx.domains import Domain
@@ -68,11 +69,16 @@ class IndexDirective(SphinxDirective):
     required_arguments = 1
     optional_arguments = 0
     final_argument_whitespace = True
-    option_spec = {}  # type: Dict
+    option_spec = {
+        'name': directives.unchanged,
+    }  # type: Dict
 
     def run(self) -> List[Node]:
         arguments = self.arguments[0].split('\n')
-        targetid = 'index-%s' % self.env.new_serialno('index')
+        if 'name' in self.options:
+            targetid = 'index-%s' % self.options['name']
+        else:
+            targetid = 'index-%s' % self.env.new_serialno('index')
         targetnode = nodes.target('', '', ids=[targetid])
         self.state.document.note_explicit_target(targetnode)
         indexnode = addnodes.index()
--- a/sphinx/domains/index.py
+++ b/sphinx/domains/index.py
@@ -71,7 +71,7 @@ class IndexDirective(SphinxDirective):
     final_argument_whitespace = True
     option_spec = {
         'name': directives.unchanged,
-    }  # type: Dict
+    }
 
     def run(self) -> List[Node]:
         arguments = self.arguments[0].split('\n')
--- a/sphinx/domains/index.py
+++ b/sphinx/domains/index.py
@@ -77,7 +77,7 @@ class IndexDirective(SphinxDirective):
         arguments = self.arguments[0].split('\n')
 
         if 'name' in self.options:
-            targetname  = self.options['name']
+            targetname = self.options['name']
             targetnode = nodes.target('', '', names=[targetname])
         else:
             targetid = 'index-%s' % self.env.new_serialno('index')
--- a/sphinx/domains/index.py
+++ b/sphinx/domains/index.py
@@ -75,11 +75,13 @@ class IndexDirective(SphinxDirective):
 
     def run(self) -> List[Node]:
         arguments = self.arguments[0].split('\n')
-        if 'name' in self.options:
-            targetid = 'index-%s' % self.options['name']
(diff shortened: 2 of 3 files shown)


tk0miya:
Thank you for proposal. I think its concept is good to me. To realize the feature, we need to discuss how to implement it.

`:name:` option is commonly used in directives of reST. And it generates "hyperlink target" and give a name to it as a reference name. So it should be added into `targetnode['names']` instead. Please don't use it to generate node_id. It is internal ID. So users and developers should not know the value of node_id.
https://docutils.sourceforge.io/docs/ref/rst/directives.html#common-options

geographika:
@tk0miya - thanks for the feedback. Using the `names` list for the target works well. 
The updated pull request allows the following syntax for an index entry:

```
.. index::
    single: execution; context
    triple: module; search; path
    :name: my-index-link

My section to index
```

And then this index entry can be referred to using a standard ref for example:

```
:ref:`My link <my-index-link>`
```

I looked briefly at trying to automatically add a title to the ref based on the index entry, but this probably introduces more complexity for little benefit. It means any use of a `ref ` without a specific title will return a warning:

```
:ref:`my-index-link`
```

Returns:

`WARNING: undefined label: my-index-link (if the link has no caption the label must precede a section header)`

doc/usage/restructuredtext/directives.rst
```diff
@@ -874,6 +874,19 @@ mainly contained in information units, such as the language reference.
...
+   .. rst:directive:option:: name: a label for hyperlink
+      :type: text
+
+      Define implicit target name that can be referenced by using
+      :rst:role:`ref`.  For example::
+
+        .. index:: Python
+           :name: py-index
+
+   .. versionadded:: 2.5
```
tk0miya: Our next target is 3.0. So this should be changed.

sphinx/domains/index.py
```diff
@@ -68,12 +69,19 @@ class IndexDirective(SphinxDirective):
     required_arguments = 1
     optional_arguments = 0
     final_argument_whitespace = True
-    option_spec = {}  # type: Dict
+    option_spec = {
+        'name': directives.unchanged,
+    }  # type: Dict
```
tk0miya: I guess mypy does not warn this line even if no type hints. Could you try that please?

sphinx/domains/index.py
```diff
@@ -68,12 +69,19 @@ class IndexDirective(SphinxDirective):
...
 
     def run(self) -> List[Node]:
         arguments = self.arguments[0].split('\n')
         targetid = 'index-%s' % self.env.new_serialno('index')
         targetnode = nodes.target('', '', ids=[targetid])
+
+        if 'name' in self.options:
+            targetname = self.options['name']
+            targetnode['names'].append(targetname)
```
tk0miya: @shimizukawa Do you have any opinion for this?

We can also give a node id based on given name like this:
```
if 'name' in self.options:
    targetname = self.options['name']
    targetnode = nodes.target('', '', names=[targetname])
else:
    targetid = 'index-%s' % self.env.new_serialno('index')
    targetnode = nodes.target('', '', ids=[targetid])
```

pros:
* We can use natural node_id generated from name. Therefore HTML anchor would be also natural.

cons:
* ID generation rule will be changed. As a result, some of URLs for indices will be changed.

tk0miya (commented):
LGTM

tk0miya:
```
.. index::
    single: execution; context
    triple: module; search; path
    :name: my-index-link

My section to index
```

I feel this notation is a bit odd. But it is hard to change the syntax of index directive. So +0 for this.

>I looked briefly at trying to automatically add a title to the ref based on the index entry, but this probably introduces more complexity for little benefit.

I understand your idea. But -0 for it. I think an "index" does not have a title.

tk0miya:
Oh, I noticed this does not have testcases. I'll consider how to do it later.

sphinx/domains/index.py
shimizukawa: For cons, `index-<serialno>` name will be changed by many reasons (new index entry is inserted before existing index entries)  so `index-<serialno>` is not a PERMANENT.

So, +1 for `:name:`  option to get PERMANENT target.

doc/usage/restructuredtext/directives.rst
geographika: Updated

sphinx/domains/index.py
geographika: I removed the type hint here

geographika:
> 
> 
> Oh, I noticed this does not have testcases. I'll consider how to do it later.

@tk0miya - thank you for the review. I've added in a test that should ensure the syntax doesn't break anything, and the two labels are created. I'm happy to change to a different approach if this testing isn't appropriate.

sphinx/domains/index.py
tk0miya: Okay, go with above my code.

tests/test_environment_indexentries.py
```diff
@@ -126,3 +126,21 @@ def test_create_index_by_key(app):
     assert index[0] == ('D', [('docutils', [[('main', '#term-docutils')], [], None])])
     assert index[1] == ('P', [('Python', [[('main', '#term-python')], [], None])])
     assert index[2] == ('ス', [('スフィンクス', [[('main', '#term-0')], [], 'ス'])])
+
+
+@pytest.mark.sphinx('dummy', freshenv=True)
+def test_create_index_with_name(app):
+    text = (".. index:: single: docutils\n   :name: ref1\n"
+            ".. index:: see: Python; interpreter\n   :name: ref2\n")
```
tk0miya: Could you separate these text by each `\n` code? It make this text a bit readable.

tests/test_environment_indexentries.py
```diff
@@ -126,3 +126,21 @@ def test_create_index_by_key(app):
...
+    restructuredtext.parse(app, text)
+    index = IndexEntries(app.env).create_index(app.builder)
+
+    # check index is created correctly
+    assert len(index) == 2
+    assert index[0] == ('D', [('docutils', [[('', '#index-0')], [], None])])
+    assert index[1] == ('P', [('Python', [[], [('see interpreter', [])], None])])
+
+    # check the reference labels are created correctly
+    labels = app.env.domaindata['std']['anonlabels']
```
tk0miya: Please use this code instead. It makes test code robust because it does not touch inside data structure directly.
```
std = app.env.get_domain('std')
assert std.anonlabels['ref1'] == (...)
```
```

File: tests/test_environment_indexentries.py [hunk 6]
```diff
@@ -126,3 +126,21 @@ def test_create_index_by_key(app):
     assert index[0] == ('D', [('docutils', [[('main', '#term-docutils')], [], None])])
     assert index[1] == ('P', [('Python', [[('main', '#term-python')], [], None])])
     assert index[2] == ('ス', [('スフィンクス', [[('main', '#term-0')], [], 'ス'])])
+
+
+@pytest.mark.sphinx('dummy', freshenv=True)
+def test_create_index_with_name(app):
```
tk0miya: Could you move this above `test_create_index_by_key()`? It would be better to place testcases for index directive nearly.

sphinx/domains/index.py
tk0miya: @geographika Could you update this please?

tk0miya (commented):
Thank you for update. Just a little more :-)

tk0miya:
I think testcodes are fine with nits.

sphinx/domains/index.py
geographika: Done - although I used targetid rather than targetname as targetid is required later and I would get a `UnboundLocalError` at `indexnode['entries'].extend(process_index_entry(entry, targetid))`

tests/test_environment_indexentries.py
geographika: Done

tests/test_environment_indexentries.py
geographika: Done

tests/test_environment_indexentries.py
geographika: Done

geographika:
@tk0miya - thanks for the feedback. I've implemented all the suggestions, and updated the test tp check for both a named ref and an autogenerated index entry.

sphinx/domains/index.py
tk0miya: Please use `targetnode['ids'][0]' instead. `document.note_implicit_target()` generates node_id automatically if not assigned yet. So we can use it after doing it.

tk0miya (approved):
Thank you for update! LGTM!!

tk0miya:
Merged 🎉 . Thank you for your effort :-)

geographika:
@tk0miya  - many thanks for taking the time to review and walk me through the process

PR #7089, opened 2020-02-02
Disable arrow keys when a modifier key is pressed

### LINKED ISSUE #7088: Arrow-key navigation ignores modifier keys (Ctrl, Alt etc.)
I just recently discovered the great feature hidden behind the `navigation_with_keys` theme option (of the `basic` theme, which is available in most themes out there).

Sadly, this is disabled by default (see #2064 for the reasoning), so I guess not many people know about it.

**Describe the bug**

Anyway, when enabled, the left and right arrow keys can be used to navigate to the previous and next page, respectively.

However, when holding the `Alt` or `Ctrl` or `Shift` key, I would expect the arrow keys to stop doing their thing.

The reason why I noticed this is a problem is because in the browser I'm using (Firefox), pressing `Alt + left` and `Alt + right` goes back and forward (respectively) in the browser history.

When `navigation_with_keys` is enabled, I now get the strange combined effect of navigating to the previous page in the browser history and from there, to the previous Sphinx page (or the other way round?).

**To Reproduce**

Use this in your `conf.py`:

```
html_theme_options = {
    'navigation_with_keys': True,
}
```

... and press `Alt + right` on one of the generated HTML pages.

**Expected behavior**

Nothing should happen when I press `Alt + right` (unless I have moved back in the browser history before).

mgeier:
Closes #7088.

I had to change `keyup()` to `keydown()` because `keyup()` doesn't include information about modifier keys (nor does `keypress()`).
See https://api.jquery.com/keypress/:

> modifier and non-printing keys such as `Shift`, `Esc`, and `delete` trigger `keydown` events but not `keypress` events.

### Feature or Bugfix

- Bugfix


tk0miya:
@TimKam Could you review this please?

@mgeier 2.0 branch is more appropriate for bug fixes. Could you rebase this please?

mgeier:
> 2.0 branch is more appropriate for bug fixes. Could you rebase this please?

Done.

TimKam:
I will review, and see the point. However, it looks like the branch is not yet properly re-based, @mgeier.

mgeier:
> However, it looks like the branch is not yet properly re-based

Thanks! I did the re-base with Git, but I didn't change the base branch on Github.

Now it should be right?

TimKam (approved):
LGTM. As we are on it, I documented the ``navigation_with_keys`` option.

choldgraf:
Just wanted to give an enthusiastic +1 from me on this - it'll let us strip out some bespoke code that we had included in one of our themes, many thanks @mgeier for discovering and fixing, and thanks @timkam for documenting :-)

PR #7092, opened 2020-02-02
Fix #7059: latex: LaTeX compilation fails into infinite loop

### LINKED ISSUE #7059: Latex build with 2 figures / 1 list hangs forever (100% CPU load)
**Describe the bug**
The .rst file below deterministically causes pdflatex to hang forever with 100 percent CPU load. I'm no LaTeX expert, so I'm unsure how to debug this further.

**To Reproduce**
```
.. Image creation: `convert -size 800x10 canvas:black black.png`
.. figure:: black.png
    :align: left

a

a
 - a


.. figure:: black.png
    :align: left
    :width: 80%
```

Build:
```
convert -size 800x10 canvas:black black.png
sphinx-build -b latex . _build/
sh -c "cd _build; make"
```
**Environment info**
I'm using a docker image with the following Dockerfile:
```
 FROM alpine:latest                                                                                                                                                                                             
 RUN apk add -U texlive python3 make texmf-dist-latexextra
```

Versions:
```
# pdflatex --version
pdfTeX 3.14159265-2.6-1.40.20 (TeX Live 2019/Alpine Linux)
kpathsea version 6.3.1
Copyright 2019 Han The Thanh (pdfTeX) et al.
There is NO warranty.  Redistribution of this software is
covered by the terms of both the pdfTeX copyright and
the Lesser GNU General Public License.
For more information about these matters, see the file
named COPYING and the pdfTeX source.
Primary author of pdfTeX: Han The Thanh (pdfTeX) et al.
Compiled with libpng 1.6.37; using libpng 1.6.37
Compiled with zlib 1.2.11; using zlib 1.2.11
Compiled with poppler version 0.82.0

# sphinx-build --version
sphinx-build 2.3.1
```

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7059

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -69,6 +69,7 @@ Bugs fixed
 * #6925: html: Remove redundant type="text/javascript" from <script> elements
 * #6906, #6907: autodoc: failed to read the source codes encoeded in cp1251
 * #6961: latex: warning for babel shown twice
+* #7059: latex: LaTeX compilation falls into infinite loop (wrapfig issue)
 * #6559: Wrong node-ids are generated in glossary directive
 * #6986: apidoc: misdetects module name for .so file inside module
 * #6899: apidoc: private members are not shown even if ``--private`` given
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1363,6 +1363,8 @@ class LaTeXTranslator(SphinxTranslator):
                 length = self.latex_image_length(node['width'])
             elif isinstance(node[0], nodes.image) and 'width' in node[0]:
                 length = self.latex_image_length(node[0]['width'])
+            self.body.append('\n\n')    # Insert a blank line to prevent infinite loop
+                                        # https://github.com/sphinx-doc/sphinx/issues/7059
             self.body.append('\\begin{wrapfigure}{%s}{%s}\n\\centering' %
                              ('r' if node['align'] == 'right' else 'l', length or '0pt'))
             self.context.append('\\end{wrapfigure}\n')


CHANGES
```diff
@@ -69,6 +69,7 @@ Bugs fixed
 * #6925: html: Remove redundant type="text/javascript" from <script> elements
 * #6906, #6907: autodoc: failed to read the source codes encoeded in cp1251
 * #6961: latex: warning for babel shown twice
+* #7059: latex: LaTeX compilation fails into infinite loop
```
jfbu: perhaps "falls into infinite loop" and then add "(wrapfig issue)"

jfbu (approved):
LGTM!

Further info for future. There is a warning by wrapfig in the context of #7059 example:
```
Package wrapfig Warning: Collision between wrapping environments
```
thus I believe adding this blank line as in the present PR is the right fix: it both removes the `wrapfig` warning and it avoids the reported issue.

About the reported issue: without this fix the #7059 example induces a context where  `\sphinxincludegraphics` expands at a time where `\linewidth` has (unexpectedly, but as result from `wrapfig`) a *negative* setting. A not yet completely clarified sequence of steps then induces core `graphicx` macro `\Gscale@div` to try to do a division by zero, and it does not realize (although it has a guard against some such division by zero) that it is repeatedly multiplying by 2 the (zero) denominator in the hope to make it bigger than 8192. This explains the infinite loop where pdflatex gets forever trapped.

Shouldn't this go into hotfix branch? (I was on leave for some time hence must re-familiarize myself with our branch model).

CHANGES
tk0miya: thanks!

tk0miya:
>Shouldn't this go into hotfix branch? (I was on leave for some time hence must re-familiarize myself with our branch model).

Yes, I should have to merge into 2.3.2 normally. But we have a plan to release 2.4 in this weekend. But I don't have time to work Sphinx in a few days. So I think merging this as 2.4 is also okay.

tk0miya:

PR #7097, opened 2020-02-04
Sanitize image filenames

s-weigand:
Subject: Sanitize filenames of images, which can break the file path

### Feature or Bugfix

- Bugfix

### Purpose
- When creating latex documentation, links to images are downloaded and if needed converted (i.e. for `SVG` this can be done by `sphinxcontrib.rsvgconverter` from [`sphinxcontrib-svg2pdfconverter`](https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter)).
If the filename contains [characters, which aren't allowed by windows](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file)  (i.e. `:` in the [repology badge](https://repology.org/badge/vertical-allrepos/python:sphinx.svg)) this leads to broken paths.
In the case of `:` on NTFS, it somehow works, since [`:` is used for data streams](https://docs.microsoft.com/de-de/archive/blogs/askcore/alternate-data-streams-in-ntfs), but the converted image is hidden and thus the latex source can't be bundled, since the image would be missing.

### Detail
- Replaces the characters `:`, `;`, `<`, `>`, `|`, `*`, `"` and ` ` in filenames with `_`. 
- The characters `/`, `\` and `?`, are not replaced since, `\` is not allowed in URLs and  `/` / `?` are taken care of in `sphinx.transforms.post_transforms.images.ImageDownloader`.


tk0miya (approved):
LGTM!

tk0miya:
This will be released as 2.4.0 in this weekend :-)

PR #7103, opened 2020-02-07
ENH: linkcheck: also write all links to output.json

westurner:
Subject: Write linkcheck output to JSON lines in `_build/linkcheck/output.json`

This is off 1.8.5. If it would be considered, I'd be glad to rebase it onto master.

(edit)

This is now off master.

### Feature or Bugfix

- Feature

### Purpose
- Write linkcheck output to JSON lines in `_build/linkcheck/output.json`
  (in order to eventually create a tool to prompt to fix links)

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -8,6 +8,7 @@
     :license: BSD, see LICENSE for details.
 """
 
+import json
 import queue
 import re
 import socket
@@ -90,6 +91,8 @@ class CheckExternalLinksBuilder(Builder):
         socket.setdefaulttimeout(5.0)
         # create output file
         open(path.join(self.outdir, 'output.txt'), 'w').close()
+        # create JSON output file
+        open(path.join(self.outdir, 'output.json'), 'w').close()
 
         # create queues and worker threads
         self.wqueue = queue.Queue()  # type: queue.Queue
@@ -225,9 +228,16 @@ class CheckExternalLinksBuilder(Builder):
 
     def process_result(self, result: Tuple[str, str, int, str, str, int]) -> None:
         uri, docname, lineno, status, info, code = result
+
+        filename = self.env.doc2path(docname, None)
+        linkstat = dict(filename=filename, lineno=lineno,
+                        status=status, code=code, uri=uri,
+                        info=info)
         if status == 'unchecked':
+            self.write_linkstat(linkstat)
             return
         if status == 'working' and info == 'old':
+            self.write_linkstat(linkstat)
             return
         if lineno:
             logger.info('(line %4d) ', lineno, nonl=True)
@@ -236,18 +246,22 @@ class CheckExternalLinksBuilder(Builder):
                 logger.info(darkgray('-ignored- ') + uri + ': ' + info)
             else:
                 logger.info(darkgray('-ignored- ') + uri)
+            self.write_linkstat(linkstat)
         elif status == 'local':
             logger.info(darkgray('-local-   ') + uri)
-            self.write_entry('local', docname, lineno, uri)
+            self.write_entry('local', docname, filename, lineno, uri)
+            self.write_linkstat(linkstat)
         elif status == 'working':
             logger.info(darkgreen('ok        ') + uri + info)
+            self.write_linkstat(linkstat)
         elif status == 'broken':
-            self.write_entry('broken', docname, lineno, uri + ': ' + info)
             if self.app.quiet or self.app.warningiserror:
                 logger.warning(__('broken link: %s (%s)'), uri, info,
-                               location=(self.env.doc2path(docname), lineno))
+                               location=(filename, lineno))
             else:
                 logger.info(red('broken    ') + uri + red(' - ' + info))
+            self.write_entry('broken', docname, filename, lineno, uri + ': ' + info)
+            self.write_linkstat(linkstat)
         elif status == 'redirected':
             try:
                 text, color = {
@@ -259,9 +273,11 @@ class CheckExternalLinksBuilder(Builder):
                 }[code]
             except KeyError:
                 text, color = ('with unknown code', purple)
-            self.write_entry('redirected ' + text, docname, lineno,
-                             uri + ' to ' + info)
+            linkstat['text'] = text
             logger.info(color('redirect  ') + uri + color(' - ' + text + ' to ' + info))
+            self.write_entry('redirected ' + text, docname, filename,
+                             lineno, uri + ' to ' + info)
+            self.write_linkstat(linkstat)
 
     def get_target_uri(self, docname: str, typ: str = None) -> str:
         return ''
@@ -301,10 +317,15 @@ class CheckExternalLinksBuilder(Builder):
         if self.broken:
             self.app.statuscode = 1
 
-    def write_entry(self, what: str, docname: str, line: int, uri: str) -> None:
-        with open(path.join(self.outdir, 'output.txt'), 'a', encoding='utf-8') as output:
-            output.write("%s:%s: [%s] %s\n" % (self.env.doc2path(docname, None),
-                                               line, what, uri))
+    def write_entry(self, what: str, docname: str, filename: str, line: int,
+                    uri: str) -> None:
+        with open(path.join(self.outdir, 'output.txt'), 'a') as output:
+            output.write("%s:%s: [%s] %s\n" % (filename, line, what, uri))
+
+    def write_linkstat(self, data: dict) -> None:
+        with open(path.join(self.outdir, 'output.json'), 'a') as output:
+            output.write(json.dumps(data))
+            output.write('\n')
 
     def finish(self) -> None:
         for worker in self.workers:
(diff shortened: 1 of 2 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -230,14 +234,21 @@ def check():
             if uri is None:
                 break
             status, info, code = check()
-            self.rqueue.put((uri, docname, lineno, status, info, code))
+            self.rqueue.put(
+                OrderedDict((
+                    ('status', status), ('code', code), ('uri', uri),
+                    ('info', info), ('docname', docname), ('lineno', lineno))
+                ))
```
tk0miya: No reason to change the data structure of `self.rqueue`. Could you pass these variables to `write_json_entry()` instead of `result` please?

sphinx/builders/linkcheck.py
```diff
@@ -312,6 +329,13 @@ def write_entry(self, what, docname, line, uri):
             output.write("%s:%s: [%s] %s\n" % (self.env.doc2path(docname, None),
                                                line, what, uri))
 
+    def write_json_entry(self, data):
```
tk0miya: It would be better to rename "json entry" to another one. I think "json entry" does not describe what the data is.

sphinx/builders/linkcheck.py
```diff
@@ -312,6 +329,13 @@ def write_entry(self, what, docname, line, uri):
...
+        # type: (dict) -> None
+        data['docpath'] = self.env.doc2path(data['docname'], None)
+        with codecs.open(path.join(self.outdir, 'output.json'), 'a', 'utf-8') as output:  # type: ignore  # NOQA
```
tk0miya: Since 3.x, `codecs.open()` is no longer needed. Please use `open()` simply. In addition, last argument `'utf-8'` is also no longer needed.

tk0miya (changes requested):
I can accept your idea. But it seems conflicted with latest code. Could you rebase this into HEAD?

sphinx/builders/linkcheck.py
westurner: Done. https://docs.python.org/2/howto/pyporting.html#text-versus-binary-data

sphinx/builders/linkcheck.py
westurner: Changed to ``write_jsonline(linkstat)``

sphinx/builders/linkcheck.py
westurner: I've replaced the one dict with the original immutable tuple and an additional dict of refs.

westurner:
Seems to build fine here with ``tox -e py37 -- -v --pdb ./tests/test_build_linkcheck.py``.
Do I need to do something to get the CI to run again?

westurner:
The output order is nondeterministic presumably due to variance in network latency.
I could just drop the row numbers in tests/test_build_linkcheck.py and revert to testing for string match?

sphinx/builders/linkcheck.py
```diff
@@ -225,9 +228,15 @@ def check() -> Tuple[str, str, int]:
 
     def process_result(self, result: Tuple[str, str, int, str, str, int]) -> None:
         uri, docname, lineno, status, info, code = result
+
+        linkstat = dict(docname=docname, lineno=lineno,
+                        status=status, code=code, uri=uri, info=info)
+        linkstat['docpath'] = docpath = self.env.doc2path(docname, None)
```
tk0miya: In sphinx, we usually call this as mere "filename" simply. It seems there are first usage of "docpath". How about renaming?

tk0miya: I think defining `docpath` (or filename) before creating linkstat will make this code simple.
```
docpath = self.env.doc2path(...)
linkstat = dict(..., docpath=docpath)
```

sphinx/builders/linkcheck.py
```diff
@@ -301,10 +316,15 @@ def write_doc(self, docname: str, doctree: Node) -> None:
...
-    def write_entry(self, what: str, docname: str, line: int, uri: str) -> None:
-        with open(path.join(self.outdir, 'output.txt'), 'a', encoding='utf-8') as output:
-            output.write("%s:%s: [%s] %s\n" % (self.env.doc2path(docname, None),
-                                               line, what, uri))
+    def write_entry(self, what: str, docname: str, docpath: str, line: int,
+                    uri: str) -> None:
+        with open(path.join(self.outdir, 'output.txt'), 'a') as output:
+            output.write("%s:%s: [%s] %s\n" % (docpath, line, what, uri))
+
+    def write_jsonline(self, data: dict) -> None:
```
tk0miya: I think jsonline is a name of data structure. We should name this like `write_linkstat()` to describe what this method doing. It also means we should give a good naming to this data.

sphinx/builders/linkcheck.py
```diff
@@ -301,10 +316,15 @@ def write_doc(self, docname: str, doctree: Node) -> None:
         if self.broken:
             self.app.statuscode = 1
 
-    def write_entry(self, what: str, docname: str, line: int, uri: str) -> None:
-        with open(path.join(self.outdir, 'output.txt'), 'a', encoding='utf-8') as output:
-            output.write("%s:%s: [%s] %s\n" % (self.env.doc2path(docname, None),
-                                               line, what, uri))
+    def write_entry(self, what: str, docname: str, docpath: str, line: int,
+                    uri: str) -> None:
```
tk0miya, line 321: Why changing format of output.txt? Is it needed for this change?

westurner: Because it is only necessary to calculate docpath/filename once; in process_result.

sphinx/builders/linkcheck.py
westurner: Since it also describes doc, also using 'doc' as a prefix for the variable is logical. I will rename to 'filename' inspecific.

westurner: Perhaps 'docfilename' would be better here?!

westurner: Given that the method is named 'doc2path', docpath seems a fitting name for the variable.

westurner: I've changed docpath to 'filename' inspecific.

westurner: done.

sphinx/builders/linkcheck.py
westurner: Changed to `write_linkstat()`

sphinx/builders/linkcheck.py
westurner: So that it's not necessary to call `doc2path()` twice.

sphinx/builders/linkcheck.py
westurner: I have renamed to `filename`

sphinx/builders/linkcheck.py
westurner: Changed to `write_linkstat(linkstat)`

westurner: (Though the other method is named `write_line`, so `write_jsonline` is logical and consistent)

westurner:
... The python: nightly job is returning this, which is probably not an error with this PR?
```python
$ if [ $IS_PYTHON = true ]; then pip install -U tox codecov; fi
  File "/home/travis/virtualenv/python3.9-dev/bin/pip", line 6, in <module>
    from pip._internal.cli.main import main
ModuleNotFoundError: No module named 'pip'
```

https://travis-ci.org/sphinx-doc/sphinx/jobs/648641307

sphinx/builders/linkcheck.py
tk0miya: Okay, I understand. If so, please replace `docname` parameter by new `filename` parameter. It seems `docname` is no longer used. So it is better.

tk0miya (commented):
LGTM with nits.

tk0miya:
In addition, please merge master branch to this branch to fix the CI error.

sphinx/builders/linkcheck.py
westurner: I've removed docname from the object in favor of just `filename`

westurner: But not from the call to `write_entry` because people could be / could've been parsing that file?

westurner:
I'm tempted to start an http.server as a `@pytest.fixture` in order to eliminate the multiple unnecessary HTTP requests to google on every test run due to the links to google in links.txt. If that's okay, would either of these or another solution be preferable?

- pytest-xprocess: https://github.com/pytest-dev/pytest-xprocess/issues/19#issue-563490465
- pytest-services: https://github.com/pytest-dev/pytest-services

westurner:
This is otherwise rebased onto latest master with:

```bash
git checkout master
git pull upstream master
git checkout linkcheck_json_output
git rebase -i master
```

And squashed

tk0miya (approved):

tk0miya:

westurner:
Thanks!

PR #7117, opened 2020-02-08
Deprecate desc_signature['first']

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Deprecate `desc_signature['first']` attribute
- It never become `True`. So no longer used.

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -203,7 +203,6 @@ class CObject(ObjectDescription):
         if targetname not in self.state.document.ids:
             signode['names'].append(targetname)
             signode['ids'].append(targetname)
-            signode['first'] = (not self.names)
             self.state.document.note_explicit_target(signode)
 
             domain = cast(CDomain, self.env.get_domain('c'))
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -115,10 +115,6 @@ class HTMLTranslator(SphinxTranslator, BaseTranslator):
     def visit_desc_signature(self, node: Element) -> None:
         # the id is set automatically
         self.body.append(self.starttag(node, 'dt'))
-        # anchor for per-desc interactive data
-        if node.parent['objtype'] != 'describe' \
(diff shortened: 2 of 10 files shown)


sphinx/domains/c.py
```diff
@@ -203,7 +203,6 @@ def add_target_and_index(self, name: str, sig: str, signode: desc_signature) ->
         if targetname not in self.state.document.ids:
             signode['names'].append(targetname)
             signode['ids'].append(targetname)
-            signode['first'] = (not self.names)
```
tk0miya, line 206: Before calling `add_target_and_index()`, `self.names` are always filled. So this assigns `False` always.

sphinx/writers/html.py
```diff
@@ -115,10 +115,6 @@ def depart_desc(self, node: Element) -> None:
     def visit_desc_signature(self, node: Element) -> None:
         # the id is set automatically
         self.body.append(self.starttag(node, 'dt'))
-        # anchor for per-desc interactive data
-        if node.parent['objtype'] != 'describe' \
-           and node['ids'] and node['first']:
-            self.body.append('<!--[%s]-->' % node['ids'][0])
```
tk0miya, line 121: It seems the `first` flag has been used to output this comment tag. I don't know why this is needed. Anyway, this is no longer working. So we can drop this feature now.

shimizukawa (approved):
LGTM!

--

I found first appearance of 'first':
https://github.com/sphinx-doc/sphinx/commits/d60ca8d49d1654968886612883a2affa9dccfaa8/sphinx/directives.py

'first' is functional in the commit like:
https://github.com/sphinx-doc/sphinx/blob/d60ca8d49d1654968886612883a2affa9dccfaa8/sphinx/directives.py#L279-L304

However, no one seems to be in trouble now, so I think it's okay to delete it.

tk0miya:
Thank you for investigation. You're archenologist!

tk0miya:
Note: Merged into 3.x branch manually.

PR #7125, opened 2020-02-10
Clarify documentation on use of doctest extension

nealmcb:
Subject: Clarify documentation on use of doctest extension

### Feature or Bugfix
- Feature

### Purpose
When I first read this section, I was mightily confused due to lack of context and the use of the passive voice.
I think this phrasing clarifies the context and indicates more clearly what the documentation author does
vs what the doctest extension and builder do.
But I'm a newbie here, so please ensure that I got this right....

### DIFF
--- a/doc/usage/extensions/doctest.rst
+++ b/doc/usage/extensions/doctest.rst
@@ -11,11 +11,15 @@
            pair: testing; snippets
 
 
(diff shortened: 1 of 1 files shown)


tk0miya (approved):
LGTM!

tk0miya:

PR #7129, opened 2020-02-10
BLD: tox.ini: flake8: add whitelist_externals for latest tox instead of adding dep, pass posargs through

westurner:
BLD: tox.ini: flake8: add whitelist_externals for latest tox instead of adding dep, pass posargs through

Subject: Update tox.ini for latest tox version and pass posargs through

### Feature or Bugfix
- Feature: ENH
- Bugfix: BUG
- Build: BLD

### Purpose
- BLD: tox.ini: flake8: add whitelist_externals for latest tox instead of adding dep, pass posargs through

### Detail
Passing posargs through makes it possible to check specific files:
```
tox -e flake8 -- ./tests/test_build.py
```

Specifying `whitelist_externals` eliminates an error message without requiring (re)install of flake8 for every run as adding to `deps` would result in.
https://tox.readthedocs.io/en/latest/config.html#conf-whitelist_externals

### Relates
- #7103

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,10 +28,12 @@ commands=
 basepython = python3
 description =
     Run style checks.
+whitelist_externals =
+    flake8
 extras =
     lint
 commands =
-    flake8
+    flake8 {posargs}
 
 [testenv:pylint]
 basepython = python3


westurner:
... The python: nightly job is returning this, which is probably not an error with this PR?
```python
$ if [ $IS_PYTHON = true ]; then pip install -U tox codecov; fi
  File "/home/travis/virtualenv/python3.9-dev/bin/pip", line 6, in <module>
    from pip._internal.cli.main import main
ModuleNotFoundError: No module named 'pip'
```

https://travis-ci.org/sphinx-doc/sphinx/jobs/648641307

tk0miya:
>... The python: nightly job is returning this, which is probably not an error with this PR?

Yes, it comes from nightly image itself. For more details, see #7132.

tk0miya:

westurner:
Thanks for the review!

PR #7142, opened 2020-02-13
Add option for auxiliary pygments styles

### LINKED ISSUE #7101: Media query dependent pygments style
Currently it is not easily possible to add an alternative pygments style based on a media query. This prevents simple implementations of a dark themes based on the `prefers-color-scheme` media query. The only way to do this currently is to hard code another css file with fitting overwrites which in not very user friendly and annoying if you want to change it.

I would like an option inside the html section of the `theme.conf` file similar to the html_css_files. This option would then create additional pygments stylesheets, add them to the static files with a fitting postfix and link them as an html `<link>` element with a media attribute.

Manually linking the pygments stylesheets would require one to manually generate them which would not be fitting for a theme as it does not happen automatically in the build process.

**Additional context**
This would most likely require only small changes to [create_pygments_style_file](https://github.com/sphinx-doc/sphinx/blob/cad1f86dd895836e36e6a57407a5d82c990dfaa8/sphinx/builders/html.py#L716) and [init_highlighter](https://github.com/sphinx-doc/sphinx/blob/cad1f86dd895836e36e6a57407a5d82c990dfaa8/sphinx/builders/html.py#L264)

We stumbled upon this problem whilst implement a dark theme here https://github.com/python/python-docs-theme/pull/44

septatrix:
First draft of a patch that would fix #7101


septatrix:
For some additional Infos and aspects which need consider see this comment in the issue https://github.com/sphinx-doc/sphinx/issues/7101#issuecomment-585676137

septatrix:
A possible change would be to also enable users to pass additional/custom classes to the option which would then result in a different selector inside the pygments file. This way it would also be easy for people to implement different styles depending not only on media queries

septatrix:
May I ask why this was closed?

tk0miya:
OMG... sorry, it's not intended. This is a side effect of change of branch model. Reopened now.

septatrix:
I now changed the implementation in a way to be consistent with pygments_style. One can now define a `pygments_dark_style` in a themes ini file and overwrite it using `html_pygments_dark_style` in a projects `conf.py`.

Additionally I decoupled the auxiliary themes option as I think only a few projects will want to implement multiple dark/light themes. So far I can only think of rust doing this but they don't use sphinx. Additional pygments styles are generated for these options but not automatically included in the html. As this is a very niche option I think most people using it will benefit from the freedom this gives them. Likewise it is possible to pass custom css_selectors as using multiple themes will need different selectors anyhow (as long as you don't want to use prefers-high-contrast e.g. media queries but then you could still leave the default selector and include them with a corresponding media query). Configuration wise these styles can be added in a themes file in an `[auxiliary_styles]` section using the style as the key and the css_selector as the value. They can be overwritten by setting `html_pygments_aux_styles` in the `conf.py` file to a corresponding dictionary. Honestly I am not really satisfied with this approach but can't think of another good way of making this option accessible to users as well as theme developers beside forcing theme developers to set the option using:
```python3
    app.config.values['html_aux_pygments_styles'] = (
        {'monokai': '(prefers-color-scheme: dark)'}, 'html', [])
```
in the themes `__init__.py` file.

However I think that the dark option for pygments styles is ready to be merged.

sphinx/builders/html/__init__.py
```diff
@@ -240,10 +241,10 @@ def _get_translations_js(self) -> str:
...
                                 'LC_MESSAGES', 'sphinx.js')
                       for dir in self.config.locale_dirs] + \
-                     [path.join(package_dir, 'locale', self.config.language,
-                                'LC_MESSAGES', 'sphinx.js'),
-                      path.join(sys.prefix, 'share/sphinx/locale',
-                                self.config.language, 'sphinx.js')]
+            [path.join(package_dir, 'locale', self.config.language,
+                       'LC_MESSAGES', 'sphinx.js'),
+             path.join(sys.prefix, 'share/sphinx/locale',
+                       self.config.language, 'sphinx.js')]
```
tk0miya: Why unindented?

sphinx/builders/html/__init__.py
```diff
@@ -263,14 +264,35 @@ def init_templates(self) -> None:
 
     def init_highlighter(self) -> None:
         # determine Pygments style and create the highlighter
+        style = 'sphinx'
         if self.config.pygments_style is not None:
             style = self.config.pygments_style
         elif self.theme:
             style = self.theme.get_config('theme', 'pygments_style', 'none')
-        else:
-            style = 'sphinx'
```
tk0miya, line 271: It seems this change is not related to the goal. Could you revert this please?

sphinx/builders/html/__init__.py
```diff
@@ -483,7 +505,8 @@ def prepare_writing(self, docnames: Set[str]) -> None:
             'parents': [],
             'logo': logo,
             'favicon': favicon,
-            'html5_doctype': html5_ready and not self.config.html4_writer
+            'html5_doctype': html5_ready and not self.config.html4_writer,
+            'pygments_dark_style': self.dark_highlighter is not None,
```
tk0miya: How about calling `add_css_file()` API instead?
```
def init_highlighter(self):
    ...
    if self.dark_highlighter:
        self.add_css_file('pygments_dark.css', media="(prefers-color-scheme: dark)", id="pygments_dark_css")
```

Then you don't need to modify layout.html.

tk0miya: I noticed there are no link tags for auxiliary styles.

sphinx/builders/html/__init__.py
septatrix: Probably auto formatter, I will revert this

sphinx/builders/html/__init__.py
septatrix: Yes, just wanted to make it consistent with my own

septatrix: Yes sure

sphinx/builders/html/__init__.py
septatrix: > How about calling `add_css_file()` API instead?

Used it in an earlier version though I think manually placing it in the layout file grants more control over the order of insertion if someone want to move pygments style files in their theme.

> I noticed there are no link tags for auxiliary styles.

You are right. I will refer to my previous comment:

> Additional pygments styles are generated for these options but not automatically included in the html. As this is a very niche option I think most people using it will benefit from the freedom this gives them.

Manual inclusion of these files would allow theme developers to e.g. postpone fetching of the stylesheets until the user actively decides to use these themes.

tk0miya: If it is a very niche, -1 for adding. TBH, there are no enough resources for maintenance. So I'd like to add only important features to the core.

It is tough work to deprecate once published configurations, template variables, APIs and so on. So we should design carefully before adding them. Is the order of insertion really needed?

septatrix: Okay, so should I call `add_css_files()` for the `pygmnents_dark.css` file, the auxiliary styles or all of them? Or do you mean dropping complete support for auxiliary styles and only add support for a single additional dark pygments stylesheet?

septatrix: ?

septatrix: I removed the link from the template and switched to using `add_css_file()` now

tk0miya: Sorry for the late comment. I meant dropping complete support for auxiliary styles.

sphinx/builders/html/__init__.py
```diff
@@ -718,6 +739,20 @@ def create_pygments_style_file(self) -> None:
         with open(path.join(self.outdir, '_static', 'pygments.css'), 'w') as f:
             f.write(self.highlighter.get_stylesheet())
 
+        if self.dark_highlighter:
+            with open(path.join(self.outdir, '_static', 'pygments_dark.css'), 'w') as f:
+                f.write(self.dark_highlighter.get_stylesheet())
+            self.add_css_file('pygments_dark.css',
+                              media='(prefers-color-scheme: dark)',
+                              id='pygments_dark_css')
```
tk0miya: Is this working expected? I suppose this does not work fine because this method is called after writing HTML files. It would be better to call this in `init_highlighter()` instead.

sphinx/builders/html/__init__.py
```diff
@@ -271,6 +272,26 @@ def init_highlighter(self) -> None:
             style = 'sphinx'
         self.highlighter = PygmentsBridge('html', style)
 
+        if self.config.html_pygments_dark_style is not None:
```
tk0miya: Note: We use `html_theme_options` to override the options of HTML themes. So I thought `html_theme_options['pygments_dark_style']` would be better than `html_pygments_dark_style`. But we also use `pygments_style` to override the default style of pygments. It's a common option for all output formats, not only for HTML. I can't say which is better for this case. Either is fine to me.

tk0miya: I noticed the document for this option is needed.

tk0miya: Tests are also needed.

sphinx/builders/html/__init__.py
septatrix: As far as I have tested this works as intended but adding them earlier may make it easier to overwrite these styles later so I will move this as you suggested

sphinx/builders/html/__init__.py
septatrix: I will leave it at `html_pygments_dark_style` for now though I am open for feedback if someone has some advantages/disadvantages

septatrix: Yes I will add them

septatrix:
Do you think we could still merge this for the 3.0 release?

tk0miya:
I'll call feature freeze and make a beta package in 24hours (maybe). I'm okay if updates will be finished until then.

But I'm not sure `html_pygments_dark_style` is the best way to configure dark style for users (or configuration is really needed for them). So it might not be merged into 3.0 release. (Sorry, please wait a moment more)

septatrix:
Okay, how about splitting this into two parts and merging the theme
developer facing option (as far as I can tell there isn't much more to
discuss) and taking our time to see how this gets used and how we should
implement user facing overwrites (or even if as it possible using
html_theme_options)?

tk0miya:
Sounds good to me. I don't opposite to add a new option for the HTML theme. So I can merge it ASAP if separated.

tk0miya:
Note: My concern about the configuration for users is 1) appropriate way to configure it, 2) it is really needed or not. I guess nobody wants it at present (because nobody knows it :-p).

septatrix:
If I remove the `html_pygments_dark_style` option I would need to move the `pygments_dark_theme` to the themes `[options]` section to be able to overwrite it for tests (as none of our built-in themes use this option and it otherwise is not overwriteable by `html_theme_options`. Is this correct?

tk0miya:
I don't have a strong opinion about the section. +0 for moving.

sphinx/builders/html/__init__.py
septatrix: Removed the option totally for now

septatrix: Documentation and tests added

septatrix:
I added tests for dark styles in the theming selection where I check if the files are generated and the link is injected into the page. Additionally I test if no formatter is used if no option is given and test the correct initialization of the highlighter if an option is given.

sphinx/highlighting.py
```diff
@@ -157,9 +157,9 @@ def highlight_block(self, source: str, lang: str, opts: Dict = None,
             # MEMO: this is done to escape Unicode chars with non-Unicode engines
             return texescape.hlescape(hlsource, self.latex_engine)
 
-    def get_stylesheet(self) -> str:
+    def get_stylesheet(self, *, css_selector: Union[str, Iterable[str]] = '.highlight') -> str:
         formatter = self.get_formatter()
         if self.dest == 'html':
-            return formatter.get_style_defs('.highlight')
+            return formatter.get_style_defs(css_selector)
```
tk0miya: Is this change still needed?

tk0miya (commented):
LGTM with nits.

sphinx/highlighting.py
septatrix: No

tk0miya (approved):

tk0miya:

PR #7147, opened 2020-02-13
Fix #7146: support (...) in type hint comments 

### LINKED ISSUE #7146: autodoc.type_comments throw IndexError
**Describe the bug**
Since 2.4.0, we get a failure on our test suites:
```
Warning, treated as error:
1116error while formatting arguments for scapy.contrib.http2.HPackHdrTable.parse_txt_hdrs: list index out of range
```
I mocked sphinx to display the stack trace:
```
Exception occurred:
2914  File "/home/travis/build/secdev/scapy/.tox/docs/lib/python3.8/site-packages/Sphinx-2.4.2.dev20200213-py3.8.egg/sphinx/ext/autodoc/type_comment.py", line 62, in update_annotations_using_type_comments
2915    annotation = ast_unparse(function.argtypes[i])  # type: ignore
```

Here's what the function looks like:
https://github.com/secdev/scapy/blob/0dd974e46b3678f02e422dacafc507f32cacde89/scapy/contrib/http2.py#L2636

The issue happens whenever there is a mismatch between the amount of arguments and the amount of type hints comments. This happens when using the `(...)`

This triggers it:
```python

def pony(a,
         b,  # type: str
         c=None  # type: Optional[str]
         ):
    # type: (...) -> str
    return b + (c or "")
```

This is an issue when using functions within class. You don't put a type hint for `self` (https://mypy.readthedocs.io/en/stable/cheat_sheet.html#classes)

<details>

<summary>Full travis logs</summary>

https://travis-ci.com/secdev/scapy/jobs/286619573
Backup:
https://pastebin.com/raw/vxhT1cGq

</details>

**To Reproduce**

```
$ git clone https://github.com/gpotter2/issues-sphinx.git
$ cd issues-sphinx
$ make html
```

**Your project**
https://github.com/secdev/scapy

**Environment info**
- OS: linux
- Python version: 3.8.1
- Sphinx version: 2.4.1

Thanks for the time.

gpotter2:
Subject: Add support for `(...)` in type hints comments

- Bugfix: fix [REDACTED]

This is one way of doing it. This will set `Any` to variables that were forgotten when using the `(...)` format, rather than crash. This is not what python does by default (it just leaves no annotation), but it's the easiest to integrate in this piece of code without having to refactor the whole file.


sphinx/ext/autodoc/type_comment.py
```diff
@@ -41,7 +41,18 @@ def get_type_comment(obj: Any) -> ast.FunctionDef:
             subject = cast(ast.FunctionDef, module.body[0])  # type: ignore
 
         if getattr(subject, "type_comment", None):
-            return ast_parse(subject.type_comment, mode='func_type')  # type: ignore
+            comment = subject.type_comment  # type: ignore
+            if comment.startswith("(...)"):
+                # resolve parameters individually.
+                args_comment = [
+                    (arg.type_comment or "Any")  # type: ignore
+                    for arg in subject.args.args
```
tk0miya: There some kind of arguments (ex. posonlyargs, vararg, etc.). So they would be supported.

tk0miya (changes requested):
Thank you for proposal. Could you add testcases for them please?

tk0miya:
In addition, I'd like to ship this as bug fixed release. So could you rebase this into 2.4.2 branch? Then it will be released as soon.

tk0miya:
For example, I added a example case to `tests/roots/test-ext-autodoc/target/typehints.py`. And I confirmed autodoc crashed with `tox -e py37 tests/test_ext_autodoc_configs.py -- -vv`
```
diff --git a/tests/roots/test-ext-autodoc/target/typehints.py b/tests/roots/test-ext-autodoc/target/typehints.py
index 842530c13..e0ce5d129 100644
--- a/tests/roots/test-ext-autodoc/target/typehints.py
+++ b/tests/roots/test-ext-autodoc/target/typehints.py
@@ -22,3 +22,11 @@ class Math:
 def complex_func(arg1, arg2, arg3=None, *args, **kwargs):
     # type: (str, List[int], Tuple[int, Union[str, Unknown]], *str, **str) -> None
     pass
+
+
+def type_comment_for_each_args(arg1,  # type: int
+                               arg2,  # type: str
+                               *args,  # type: Any
+                               kwarg=None):  # type: int
+    # type: (...) -> int
+    pass
```

gpotter2:
I've made an alternative PR with a fix i prefer: https://github.com/sphinx-doc/sphinx/pull/7152
please have a look and tell me which one you prefer.
The second one is less likely to break

PR #7152, opened 2020-02-14
Fix #7146: support (...) in type hint comments (V2)

### LINKED ISSUE #7146: autodoc.type_comments throw IndexError

gpotter2:
An alternative fix to #7146, with tests.

I prefer this one: there is no compromises like the ones in #7147 (we don't re build then re parse the arguments, and we never fail to bind the correct attribute to the correct name).

However this changes the output of `get_type_comment`. I haven't found other uses of the function.
It could also return a tuple instead of the dictionary. I'm open to changes


sphinx/ext/autodoc/type_comment.py
```diff
@@ -41,7 +42,34 @@ def get_type_comment(obj: Any) -> ast.FunctionDef:
...
+                node = subject
+                results = {
+                    'returns': node.returns,
+                    'explicit': True,
+                    'args': {
+                        arg.arg: arg.type_comment
+                        for arg in itertools.chain(
+                            getattr(node.args, "posonlyargs", []) or [],
+                            getattr(node.args, "args", []) or [],
+                            getattr(node.args, "vararg", []) or [],
```
tk0miya: vararg and kwarg are not list. So this code would not work. Please test with `(*args, **kwargs)`.

sphinx/ext/autodoc/type_comment.py
```diff
@@ -41,7 +42,34 @@ def get_type_comment(obj: Any) -> ast.FunctionDef:
...
+            else:
+                node = subject
+                results = {
+                    'returns': node.returns,
+                    'explicit': True,
+                    'args': {
+                        arg.arg: arg.type_comment
+                        for arg in itertools.chain(
+                            getattr(node.args, "posonlyargs", []) or [],
+                            getattr(node.args, "args", []) or [],
```
tk0miya: Except posonlyargs, other attributes (ex. args, varargs) always exist. So you don't need to call `getattr()`. It makes code difficult to read (a bit).

Note: posonlyargs exists only python3.8 and later.

sphinx/ext/autodoc/type_comment.py
```diff
@@ -41,7 +42,34 @@ def get_type_comment(obj: Any) -> ast.FunctionDef:
...
+                    ast.FunctionDef,
+                    ast_parse(comment, mode='func_type')
+                )
+                results = {
+                    'returns': node.returns,
+                    'explicit': False,
+                    'args': node.argtypes  # type: ignore
+                }
+            else:
+                node = subject
```
tk0miya: Why did you rename `subject` here?

sphinx/ext/autodoc/type_comment.py
```diff
@@ -41,7 +42,34 @@ def get_type_comment(obj: Any) -> ast.FunctionDef:
             subject = cast(ast.FunctionDef, module.body[0])  # type: ignore
 
         if getattr(subject, "type_comment", None):
-            return ast_parse(subject.type_comment, mode='func_type')  # type: ignore
+            comment = subject.type_comment  # type: ignore
```
tk0miya: Why did you make a new variable `comment` here? It is used only twice (in first if-block).

tests/test_ext_autodoc_configs.py
```diff
@@ -498,6 +506,10 @@ def test_autodoc_typehints_signature(app):
         '',
         '.. py:function:: incr(a: int, b: int = 1) -> int',
         '   :module: target.typehints',
+        '',
+        '',
+        '.. py:function:: missing_attr(c: None, a: str, b: Optional[str] = None) -> None',
```
tk0miya: I feel strange for non-annotated arguments are typed as `None`. I think they should not be annotated.

tk0miya: Hmm? It seems the type_comment of `missing_attr()` is `# type: (...) -> str`. The returns annotation should be `str`. Is this testcase correct?

tk0miya:
As a trial, I just made a patch to improve this: 98d2446. Please merge it if you like it.

gpotter2:
That's a much cleaner patch. Great job, thanks a lot

tk0miya:
Thank you for update. LGTM!! I'll release this as 2.4.2 within a few days :-)

gpotter2:
Great thanks !

PR #7155, opened 2020-02-14
py domain: Use AST parser to convert signature to doctree

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- As a preparation of #6990, Use AST parser to convert signature to doctree
- It allows to decorate each tokens in latter step.
- Originally, this is a part of #6990.

### DIFF
--- a/tests/test_domain_py.py
+++ b/tests/test_domain_py.py
@@ -8,6 +8,7 @@
     :license: BSD, see LICENSE for details.
 """
 
+import sys
 from unittest.mock import Mock
 
 import pytest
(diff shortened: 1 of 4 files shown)


tests/test_domain_py.py
```diff
@@ -241,7 +242,63 @@ def test_pyfunction_signature(app):
...
+                                                        "=None")])])
+
+    # case: separator in the middle
+    text = ".. py:function:: hello(a, /, b, *, c)"
+    doctree = restructuredtext.parse(app, text)
+    assert_node(doctree[1][0][1],
+                [desc_parameterlist, ([desc_parameter, "a"],
+                                      "/",
+                                      [desc_parameter, "b"],
+                                      "*",
```
tk0miya, line 292: So far, all children of `desc_parameter` are only `desc_parameter`. But this PR makes `Text` node as children of it. I don't think this is a breaking change, but...

tk0miya: Note: The `Text` nodes will be replaced by `inline` node in #6990.

shimizukawa: I think it's OK.

tk0miya: Thank you for comment!

PR #7167, opened 2020-02-16
Close #3106: Register a hyperlink target for index page automatically

### LINKED ISSUE #3106: Link to custom module index?
I am using the Domain API and generate a custom index using:
http://www.sphinx-doc.org/en/1.4.8/extdev/domainapi.html#sphinx.domains.Index.generate

1)
:ref:`modindex` seems hardcoded to "py-modindex.html" whether or not this page exists or not..

2)
A custom Domain (name='msk') and Index (name='mskindex') does not seem to create a label. I have not inspected the rather huge doctree, but simply tried :ref:`mskindex`, :ref:`msk-mskindex` and :ref:`msk-modindex` among others..

Am I missing something or am suppose to add the target my self? In http://www.sphinx-doc.org/en/1.4.8/extdev/domainapi.html#sphinx.domains.Index.generate I tried::

    tnode = nodes.target('', '', ids=['msk-mskindex'])
    self.state.document.note_explicit_target(tnode)
    content.insert(0, tnode)

but I could not find a reference to state from from the Index object.

As work-around I am using an explicit html link given that the location is known..

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Register a hyperlink target for index page automatically
- It allows users to refer index page via `:ref:` role.
- refs: #3106

### DIFF
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -11,6 +11,7 @@
 
 import copy
 from typing import Any, Callable, Dict, Iterable, List, NamedTuple, Tuple, Union
+from typing import cast
 
 from docutils import nodes
 from docutils.nodes import Element, Node, system_message
@@ -70,6 +71,9 @@ class Index:
     a domain, subclass Index, overriding the three name attributes:
 
     * `name` is an identifier used for generating file names.
+      It is also used for a hyperlink target for the index. Therefore, users can
+      refer the index page using ``ref`` role and a string which is combined
+      domain name and ``name`` attribute (ex. ``:ref:`py-modindex```).
     * `localname` is the section title for the index.
     * `shortname` is a short name for the index, for use in the relation bar in
       HTML output.  Can be empty to disable entries in the relation bar.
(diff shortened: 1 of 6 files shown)


sphinx/domains/std.py
```diff
@@ -544,12 +544,10 @@ class StandardDomain(Domain):
         'objects': {},          # (type, name) -> docname, labelid
         'labels': {             # labelname -> docname, labelid, sectionname
             'genindex': ('genindex', '', _('Index')),
-            'modindex': ('py-modindex', '', _('Module Index')),
```
tk0miya: With this change, reference title will be changed because `sphinx.domains.python:ModuleIndex.name` is "Python Module Index". But I don't have opinion for that.

sphinx/domains/__init__.py
```diff
@@ -219,6 +226,17 @@ def __init__(self, env: "BuildEnvironment") -> None:
...
+    def setup(self) -> None:
+        """Set up domain object."""
+        from sphinx.domains.std import StandardDomain
+
+        # Add special hyperlink target for index pages (ex. modindex)
+        std = cast(StandardDomain, self.env.get_domain('std'))
+        for index in self.indices:
+            if index.name and index.localname:
+                docname = "%s-%s" % (self.name, index.name)
+                std.note_hyperlink_target(index.name, docname, '', index.localname)
```
tk0miya: I wonder it would be better to prepend domain name to name of reference target (ex. ```:ref:`py-modindex` ``` or ```:ref:`py:modindex` ```).

shimizukawa: +1 for "prepend domain name" to avoid conflict.

shimizukawa (commented):
+1 for "prepend domain name" to avoid conflict.

sphinx/domains/__init__.py
tk0miya: Thanks, updated.

PR #7179, opened 2020-02-18
Fix broken handling of whitespace in GenericObject.

silene:


tk0miya:
Could you let me know what trouble is happened in your document. I agree this is a mistake. But I'd like to know the troubles in real world.

silene:
This is explained in the commit message (not sure why Github decided to hide it). Here it is:

    Fix broken handling of whitespace in GenericObject.
    
    This actually makes the code matches its comment, i.e, whitespaces are now
    normalized the same way XRefRole does.
    
    Without the fix, when creating a directive/role via add_object_type, the
    following testcase fails with "WARNING: 'baz' reference target not found:
    foo bar".
    
        :baz:`foo bar`
        .. baz:: foo bar
    
    The current workaround is to explicitly mention the broken target, e.g.,
    
        :baz:`foo bar <foobar>`
        .. baz:: foo bar

tk0miya:
Thank you for explanation. Absolutely, this is a bug! Confirmed with this:
```
.. envvar:: HOME
.. envvar:: HOME HOME

:envvar:`HOME`
:envvar:`HOMEHOME`
:envvar:`HOME HOME`
```

I only worry about this PR changes the behavior of `GenericObject` incompatibly. I don't know how many people will be affected by this. But I think there are no way to avoid the breaking change.
@shimizukawa What do you think?

tk0miya (approved):

silene:
I agree that compatibility is the main issue. Since you seem to have a major revision on the way (Sphinx 3), it seems like a good time to fix the bug without worrying too much about compatibility.

Also, I am not sure that many people will be impacted. For example, your `envvar` testcase is a good reproducer, but no sane person would ever put a whitespace inside an environment variable. My guess is that most people use application-specific domains rather than `std` when it comes to whitespace-sensitive references (which I ended up doing too), and thus the bug in `GenericObject` just goes ignored.

tk0miya:
Agreed. I also the incompatible change harms nobody. So I'll merge this if other committers don't object to this. I'll merge this just before feature freeze of 3.0. Please wait a moment.

tk0miya:
It seems no objections to this. So I'm merging this now. Thank you for your contribution!

PR #7182, opened 2020-02-19
fix a missing letter in a message in french translation catalog

deronnax:
Subject: fix a typo in french translation

- bugfix (kindka)


JulienPalard:
@deronnax Sphinx is only using Transifex: https://www.transifex.com/sphinx-doc/ could you please join their group on tx and apply your patch there?

PR #7187, opened 2020-02-20
Fix #7183: intersphinx: ``:attr:`` reference to property is broken

### LINKED ISSUE #7183: :attr: links to intersphinx-reachable properties are still broken

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7183

### DIFF
--- a/sphinx/ext/intersphinx.py
+++ b/sphinx/ext/intersphinx.py
@@ -282,6 +282,9 @@ def missing_reference(app: Sphinx, env: BuildEnvironment, node: Element, contnod
     if 'std:cmdoption' in objtypes:
         # until Sphinx-1.6, cmdoptions are stored as std:option
         objtypes.append('std:option')
+    if 'py:attribute' in objtypes:
+        # Since Sphinx-2.1, properties are stored as py:method
+        objtypes.append('py:method')
     to_try = [(inventories.main_inventory, target)]
     if domain:
         full_qualified_name = env.get_domain(domain).get_full_qualified_name(node)
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Bugs fixed
 
 * #7184: autodoc: ``*args`` and ``**kwarg`` in type comments are not handled
   properly
+* #7183: intersphinx: ``:attr:`` reference to property is broken
 
 Testing
 --------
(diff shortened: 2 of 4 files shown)


sphinx/ext/intersphinx.py
```diff
@@ -282,6 +282,9 @@ def missing_reference(app: Sphinx, env: BuildEnvironment, node: Element, contnod
     if 'std:cmdoption' in objtypes:
         # until Sphinx-1.6, cmdoptions are stored as std:option
         objtypes.append('std:option')
+    if 'py:attribute' in objtypes:
+        # Since Sphinx-2.4, properties are stored as py:method
```
flying-sheep: The bug exists since sphinx 2.1, right?

Can you somewhere add documentation about all the hacks that went into this so we can remove them all at once in the future?

1. `PyMethod` has some branching that exists only for properties
2. `PyDomain.resolve_xref` now contains a hack
3. `intersphinx.missing_reference` now contains a hack
4. maybe I forgot something?

Also this probably needs a test, right?

flying-sheep (changes requested):

sphinx/ext/intersphinx.py
tk0miya: >The bug exists since sphinx 2.1, right?

half correct. properties are stored since 2.1. But `py:attribute` refers it since 2.4. Anyway, my comment is wrong.

>Can you somewhere add documentation about all the hacks that went into this so we can remove them all at once in the future?

The goal of this PR should be a bug fix. It would be better to do in another PR. In addition, nothing are determined yet. So marking as removal is too much.

>Also this probably needs a test, right?

Okay, I'll try it.

PR #7191, opened 2020-02-21
Feature: Update event autodoc-skip-member to pass what object the member is from

terencehonles:
In some cases the name and object is not sufficient information to know if a member should be skipped or not. This PR adds the object `obj` is a member of (`member_of`) in order to give more context to the handler.

In order to be a minimally breaking change the event system is updated to allow events to be registered with an "upgrader"/"validator" function which can return a wrapped fallback. This will work properly when a user "connects" to the event with a function that expects an explicit number of arguments. If the function instead takes `*args` and then unpacks or uses subscripts in the handler this will be a breaking change.

The reason for this change is you may want to redact `obj` only if it is a member of a subclass of some type `T` and as is you either have to blindly redact `obj` if the name matches some name you may want to look for, but you have no ability to actually make sure it's on the right type. By having `member_of` you could check `issubclass(member_of, T)` to make sure you are redacting the right members for the right object(s).

### DIFF
--- a/doc/usage/extensions/autodoc.rst
+++ b/doc/usage/extensions/autodoc.rst
@@ -574,9 +574,13 @@ Skipping members
 autodoc allows the user to define a custom method for determining whether a
 member should be included in the documentation by using the following event:
 
-.. event:: autodoc-skip-member (app, what, name, obj, skip, options)
+.. event:: autodoc-skip-member (app, what, name, obj, member_of, skip, options)
 
    .. versionadded:: 0.5
+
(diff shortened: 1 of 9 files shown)


AA-Turner:
This PR is out-of-date following the autodoc refactor (#13751). If revisiting this, I'd probably prefer to add a new event and deprecate the current/old one, rather than to add more complexity to the events system.

I'll close for now, but happy to revisit a rebased/fresh PR.

A

Issue #6387, opened 2019-05-21
Search function broken in 2.0.1


tk0miya:
**Describe the bug**
HTML search is still broken in 2.0.1 (refs: #6244)

**To Reproduce**
Build HTML with `haiku` theme.

**Expected behavior**
search succeeded.

**Your project**
N/A

**Screenshots**
N/A

**Environment info**
- OS: Mac
- Python version: 3.7.2
- Sphinx version: 2.0.1
- Sphinx extensions: N/A

PR #7192, opened 2020-02-21
Fix search in dirhtml output

vsalvino:
Subject: The built-in JavaScript search exhibits a rather major bug in that it seems to completely break when the output is generated in `dirhtml` mode, OR if a custom theme is missing a `role="main"` HTML element. This change fixes the search to work correctly in both cases.

### Feature or Bugfix
- Bugfix

### Detail
When built in `dirhtml`, URLs such as `docs/test.html` now become `docs/test/`. However, in many cases the search was performing incorrect ajax queries such as `docs/test/index/` and `docs/test/.html`. My change looks for the correct URL when making ajax calls in the search.

These invalid URLs of course trigger 404s which in turn exposes another bug when trying to process the response HTML and look for a `[role='main']` element. This could also have been causing issues with custom themes that do not implement an element with `role=main`. My change also handles that case: if that element does not exist, an empty string instead is returned.

### Relates
There are a couple related issues, which are marked as "closed", but the issue still persists on sphinx 2.4.x. See #6244 , #6257 , #6387

---

Please provide feedback on if this should be merged into a different branch, or if my Pull Request is missing anything.

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -63,6 +63,9 @@ var Search = {
       htmlElement.innerHTML = htmlString;
       $(htmlElement).find('.headerlink').remove();
       docContent = $(htmlElement).find('[role=main]')[0];
+      if(docContent === undefined) {
+        return "";
+      }
       return docContent.textContent || docContent.innerText;
   },
 
@@ -245,6 +248,7 @@ var Search = {
       if (results.length) {
         var item = results.pop();
         var listItem = $('<li style="display:none"></li>');
+        var requestUrl = "";
         if (DOCUMENTATION_OPTIONS.BUILDER === 'dirhtml') {
           // dirhtml builder
           var dirname = item[0] + '/';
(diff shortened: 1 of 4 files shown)


sphinx/themes/basic/static/searchtools.js
```diff
@@ -63,6 +63,9 @@ var Search = {
       htmlElement.innerHTML = htmlString;
       $(htmlElement).find('.headerlink').remove();
       docContent = $(htmlElement).find('[role=main]')[0];
+      if(docContent === undefined) {
+        return "";
```
tk0miya, line 67: It might be helpful if there are debug message in JS console.

tk0miya (approved):
LGTM! Thanks!

tk0miya:
I confirmed this works fine on my local.

tk0miya:

sphinx/themes/basic/static/searchtools.js
tk0miya: I posted it as an another PR: https://github.com/sphinx-doc/sphinx/pull/7194

rdb:
Note that a side effect of this change isn't just that the AJAX requests become absolute, but also that the links themselves become absolute links, including for non-dirhtml builders.  Was this intended?

vsalvino:
It wasn't necessarily indented that links become absolute. However, the links were being incorrectly generated and were producing all 404s (in dirhtml mode). So the intention was to generate the correct URLs.

Is this change causing breakage?

sciunto:
I think so. https://github.com/scikit-image/scikit-image/issues/4616

Issue #6903, opened 2019-12-08
some domain directives generate wrong target ID


tk0miya:
**Describe the bug**
Some domain directives generate wrong target ID. For example, both `py:module` and `js:module` generate target IDs with wrong operation. So they might be conflicted.

**To Reproduce**

This input are converted to following:
```
blah blah blah

.. js:module:: foo

blah blah blah

.. py:module:: foo

blah blah blah
```
```
<p>blah blah blah</p>
<span class="target" id="module-foo"></span><p>blah blah blah</p>
<span class="target" id="module-foo"></span><p>blah blah blah</p>
```

**Expected behavior**
All domain directives generate valid target IDs and they are not conflicted.

**Your project**
N/A

**Screenshots**
N/A

**Environment info**
- OS: Mac
- Python version: 3.8.0
- Sphinx version: 2.2.2
- Sphinx extensions: N/A

PR #7210, opened 2020-02-23
js domain: Generate node_id in the right way

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6903 
- So far, JS domain generates node_ids (a.k.a. target id) from their object name (ex. `func` from `.. js:function:: func()`).
- JS domain checks conflicts of node_ids between JS objects. But it does not check for other domains. Therefore, they are often conflicted.
- In addition, the generated node_ids do not obey the rule of docutils specification.
- On the other hand, this makes an incompatible changes for domain internal data.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -23,6 +23,10 @@ Incompatible changes
 * Due to the scoping changes for :rst:dir:`productionlist` some uses of
   :rst:role:`token` must be modified to include the scope which was previously
   ignored.
+* #6903: js domain: Internal data structure has changed. Both objects and
+  modules have node_id for cross reference
+* #7210: js domain: Non intended behavior is removed such as ``parseInt_`` links
+  to ``.. js:function:: parseInt``
 
 Deprecated
 ----------
--- a/sphinx/domains/javascript.py
+++ b/sphinx/domains/javascript.py
@@ -341,8 +341,8 @@ class JavaScriptDomain(Domain):
     def note_object(self, fullname: str, objtype: str, location: Any = None) -> None:
         if fullname in self.objects:
             docname = self.objects[fullname][0]
-            logger.warning(__('duplicate object description of %s, other instance in %s'),
-                           fullname, docname, location=location)
+            logger.warning(__('duplicate %s description of %s, other %s in %s'),
+                           objtype, fullname, objtype, docname, location=location)
         self.objects[fullname] = (self.env.docname, objtype)
 
     @property
--- a/sphinx/domains/javascript.py
+++ b/sphinx/domains/javascript.py
@@ -28,10 +28,19 @@ from sphinx.roles import XRefRole
 from sphinx.util import logging
 from sphinx.util.docfields import Field, GroupedField, TypedField
 from sphinx.util.docutils import SphinxDirective
-from sphinx.util.nodes import make_refnode
+from sphinx.util.nodes import make_id, make_refnode
 
 
 logger = logging.getLogger(__name__)
+
+
+def make_old_jsmod_id(modname: str) -> str:
+    """Generate old styled node_id for JS modules.
+
+    .. note:: Old Styled node_id was used until Sphinx-3.0.
+              This will be removed in Sphinx-5.0.
+    """
+    return 'module-' + modname
 
 
 class JSObject(ObjectDescription):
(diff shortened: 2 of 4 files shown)


sphinx/domains/javascript.py
```diff
@@ -106,20 +106,19 @@ def add_target_and_index(self, name_obj: Tuple[str, str], sig: str,
                              signode: desc_signature) -> None:
         mod_name = self.env.ref_context.get('js:module')
         fullname = (mod_name + '.' if mod_name else '') + name_obj[0]
-        if fullname not in self.state.document.ids:
-            signode['names'].append(fullname)
```
tk0miya, line 110: This PR also removes hypertarget names for JS objects. I don't know this is official feature or not. But it works well at this moment.
```
.. js:function:: func

   blah blah blah

reference with func_.
```

But I don't consider this is not intended behavior. Users should use `:js:*:` role instead to refer them.
So it is disabled in this PR.

sphinx/domains/javascript.py
```diff
@@ -106,20 +106,19 @@ def add_target_and_index(self, name_obj: Tuple[str, str], sig: str,
...
-            signode['ids'].append(fullname.replace('$', '_S_'))
```
tk0miya, line 111: It might be better to generate an old styled anchor as an alias for a while. It will helps to old hyperlinks. But that also means we will not remove old styled ones in future... Is there any opinion?

tests/test_domain_js.py
```diff
@@ -120,21 +120,28 @@ def find_obj(mod_name, prefix, obj_name, obj_type, searchmode=0):
 
     assert (find_obj(None, None, 'NONEXISTANT', 'class') == (None, None))
     assert (find_obj(None, None, 'NestedParentA', 'class') ==
-            ('NestedParentA', ('roles', 'class')))
+            ('NestedParentA', ('roles', 'nestedparenta', 'class')))
     assert (find_obj(None, None, 'NestedParentA.NestedChildA', 'class') ==
-            ('NestedParentA.NestedChildA', ('roles', 'class')))
+            ('NestedParentA.NestedChildA',
+             ('roles', 'nestedparenta-nestedchilda', 'class')))
```
tk0miya, line 126: This is good example for new_id. It is regulated to small case and invalid characters are converted to hyphens.

sphinx/domains/javascript.py
tk0miya: I noticed we have to mention about this on CHANGES.

CHANGES
```diff
@@ -8,6 +8,8 @@ Dependencies
   Japanese documents as ``.xbb`` files are unneeded by :program:`dvipdfmx`
   since TeXLive2015 (refs: #6189)
 * babel-2.0 or above is available (Unpinned)
+* #6903: js domain: Internal data structure has changed. Both objects and
+  modules have node_id for cross reference
```
tk0miya: This should be added to Incompatible changes section.

sphinx/domains/javascript.py
tk0miya: updated.

sphinx/domains/javascript.py
```diff
@@ -316,47 +313,48 @@ class JavaScriptDomain(Domain):
...
 
     @property
-    def objects(self) -> Dict[str, Tuple[str, str]]:
-        return self.data.setdefault('objects', {})  # fullname -> docname, objtype
+    def objects(self) -> Dict[str, Tuple[str, str, str]]:
+        return self.data.setdefault('objects', {})  # fullname -> docname, node_id, objtype
 
-    def note_object(self, fullname: str, objtype: str, location: Any = None) -> None:
+    def note_object(self, fullname: str, objtype: str, node_id: str,
+                    location: Any = None) -> None:
```
tk0miya, line 356: This method takes an additional argument `node_id`. It would be a breaking change...

sphinx/domains/javascript.py
tk0miya: I think no reason to not to support old styled anchors.

tk0miya: done

sphinx/domains/javascript.py
tk0miya: I thought other APIs in JS domain are also changed. And I don't have good idea to keep compatibility. So I'm going with this implementation.

sphinx/domains/javascript.py
shimizukawa: >this is not intended behavior

Agreed

CHANGES
```diff
@@ -23,6 +23,11 @@ Incompatible changes
 * Due to the scoping changes for :rst:dir:`productionlist` some uses of
   :rst:role:`token` must be modified to include the scope which was previously
   ignored.
+* #6903: js domain: Internal data structure has changed. Both objects and
+  modules have node_id for cross reference
+* #7210: js domain: Reference JavaScript entries via hyperlink target reference
+  notation (ex. ``parseInt_``).  Only cross reference notation are available
+  (ex. ``:js:func:`parseInt```).
```
shimizukawa: I couldn't figure out what would happen from this changelog.
How about: 'Non intended behavior is removed such as ``parseInt_` links to `.. js:function:: parseInt``

shimizukawa (approved):
LGTM with 1 proposal.

CHANGES
tk0miya: 👍

tk0miya:
Thank you for reviewing!

PR #7213, opened 2020-02-24
Add a procedure to update docker image

tk0miya:
### Feature or Bugfix
- Refactoring

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ distribute-*
 env/
 build/
 dist/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -23,6 +23,7 @@ distribute-*
 env/
 build/
 dist/
+docker/
```
tk0miya, line 26: It might be better to use git-submodule for this purpose. But I'm not good at it...

PR #7221, opened 2020-02-27
Latex: replace polyglossia with babel for zh

oxnz:
Subject:  replace polyglossia with babel for zh (Chinese environment)

### Feature or Bugfix

- Bugfix

### Purpose

as title

### Detail

polyglossia does not compatible with xeCJK.
it would override all Chinese font with Serif family.

### Relates
- <URL or Ticket>

### DIFF
--- a/sphinx/builders/latex/constants.py
+++ b/sphinx/builders/latex/constants.py
@@ -184,6 +184,8 @@ ADDITIONAL_SETTINGS = {
         'babel':        '\\usepackage{babel}',
     },
     ('xelatex', 'zh'): {
+        'polyglossia':  '',
+        'babel':        '\\usepackage{babel}',
         'fontenc':      '\\usepackage{xeCJK}',
     },
     ('xelatex', 'el'): {


jfbu (commented):
Do you know for how long babel supports Chinese? Unfortunately I can't say from its documentation.

It seems indeed polyglossia does not support Chinese indeed out of the box. Do you mean our current zh set-up is completely broken?

What about [ctex](https://ctan.org/pkg/ctex) ([home page](http://www.ctex.org/HomePage)). Should we use specific classes (as in done for Japanese)?

oxnz:
babel support Chinese.(zh) and is compatible with xeCJK.
I don't known the specific percentage. but xeCJK is totally fine in supporting Chinese typesetting and is very popular.

`polyglossia` messes font settings at least. so whatever you set, you only got one typeface for Chinese related characters.
It seems Chinese users prefer babel to polyglossia.
while polyglossia doesn't support Chinese at all.

> Do you mean our current zh set-up is completely broken?

it's a point of view.
current status is that you can't use it out of box. must do some tweaking.
there's general ones and specific ones.
general ones like:
1. line break

```
\XeTeXlinebreakskip = 0pt plus 1pt minus 0.1pt
 \XeTeXlinebreaklocale "zh" 
```

2. line spread (almost no one would use linespread{1.0} for Chinese, bad readability)

3. indentation (Chinese paragraph most always has 2em indent)

```
\usepackage{indentfirst}
\setlength{\parindent}{2em}
\setlength{\parskip}{0pt}
```

but these are some kind of refinements. (good to have, ok if not)

> What about [ctex](https://ctan.org/pkg/ctex) ([home page](http://www.ctex.org/HomePage)). Should we use specific classes (as in done for Japanese)?

`ctex` is another option, and has its own user group.  its development is relative slow when compared to xeCJK. and its forum has been shutdown since (2018) forever.
So I would suggest we keep xeCJK, because it's more general and more mature.

jfbu (approved):
@oxnz Thanks for your help and advice. +1 for switching to babel by default for zh.

(I see from [babel repo](https://github.com/latex3/babel) and my archive of TL installs that apparently babel documentation mentions Chinese explicitly only starting late 2017 but perhaps there was Chinese support via extra packages already prior to that date although I know babel support for xetex and luatex only started really around that date and that it has been actively maintained since. Anyway even now polyglossia seems to not support Chinese, so we can not break more user set-up by choosing babel)

tk0miya:
+0

tk0miya:
Merged.

tk0miya:

Issue #7189, opened 2020-02-21
autodoc + classmethod coroutines do not play well together


ewjoachim:
**Describe the bug**
When trying to use autodoc to generate documentation for an async classmethod, the method is shown as a classmethod, not mentionning that it is a coroutine

**To Reproduce**
Python code:
```py
class Test:
    @classmethod
    async def a(cls):
        """yay"""
        pass
```
Documentation:
```rst
.. autoclass:: procrastinate.aiopg_connector.Test
    :members:
```

Generated rst:
```rst
.. py:class:: Test
   :module: procrastinate.aiopg_connector

   .. py:method:: Test.a()
      :module: procrastinate.aiopg_connector
      :classmethod:
   
      yay
      
```

Generated doc:
![Screenshot from 2020-02-21 15-11-02](https://user-images.githubusercontent.com/1457576/75041221-6eef9000-54bc-11ea-9b4f-febff513d151.png)

**Expected behavior**
Method should appear as an `async classmethod`, not just a `classmethod`

**Your project**
https://github.com/peopledoc/procrastinate

**Screenshots**
see above

**Environment info**
- OS: Linux
- Python version: 3.7.3
- Sphinx version: 2.4.2
- Sphinx extensions: 
```python
extensions = [
    "sphinx.ext.napoleon",
    "sphinx.ext.autodoc",
    "sphinx_autodoc_typehints",
    "sphinxcontrib.programoutput"
]
```
- Extra tools: not applicable

**Additional context**
I've tried [sphinxcontrib-asyncio](https://pythonhosted.org/sphinxcontrib-asyncio/) and they have the exact same problem but reverse: they indicate it's async, but not a classmethod.

tk0miya:
Thank you for reporting. Good point!

Issue #6573, opened 2019-07-10
Callables implementing descriptor protocol are documented as attributes


peterbell10:
**Describe the bug**

A callable that implements the descriptor protocol will be documented by autosummary as an attribute rather than a function. e.g.

```python
class _MyFunction:
    def __call__(self):
        """Example function-like object"""
        print("Hello, world")

    def __get__(self, instance, owner):
        from functools import partial
        return partial(self.__call__, instance)
```

**To Reproduce**
From the root of the included project,
```
$ make -C docs html
```
`docs/source/generated/autobugdemo.func.rst` now contains the autosummary for the callable function, which is using `autoattribute`.

**Expected behavior**

I expect `autofunction` to be generated since it is a callable object.

**Your project**
[sphinx-autofunc-bug.zip](https://github.com/sphinx-doc/sphinx/files/3379203/sphinx-autofunc-bug.zip)

**Environment info**
- OS: Ubuntu Linux
- Python version: 3.7.1
- Sphinx version: 2.1.2
- Sphinx extensions:  sphinx.ext.autodoc, sphinx.ext.autosummary

**Additional context**
I am creating a function decorator that wraps the function in a callable object. This object must implement the descriptor protocol in order to correctly act like a [method][1]. 

Sphinx does have checks for built in Function-like types (e.g. `WrapperDescriptorType`), but as far as I can tell there's no equivalent for user types. I think the relevant special casing is in `isattributedescriptor`

https://github.com/sphinx-doc/sphinx/blob/5d89c30dab810117421b6262362b1b45432c9f8e/sphinx/util/inspect.py#L171-L195

[1]: https://stackoverflow.com/a/30105234/2556706

PR #7222, opened 2020-02-28
Fix: autodoc: `__wrapped__` functions are not documented correctly

terencehonles:
### Bugfix

Related: #7189 (not caused by, and just fixed with #7189)
Maybe related: #6573 

Functions that are decorated with `@lru_cache` or other `functools` decorators may not even be detected as a function. This results in the documentation not having the `()` or even trying to render the function signature.

This change updates the `sphinx.util.inspect` code to unwrap `__wrapped__` functions before determining if they can be documented.

`@lru_cache` and its associated test is an example of a decorated function that is incorrectly identified as an attribute rather than a module level function and when rendering the signature (upon changing `isattributedescriptor`) the decorated function is still incorrectly identified as a C function.

This change also renames the newly introduced `unwrap` as `unwrap_all` because it is different than the prexisting Python supplied `inspect.unwrap`.

See `update_wrapper` "Changed in version 3.4" for more background: https://docs.python.org/3/library/functools.html#functools.update_wrapper

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -17,7 +17,7 @@ import typing
 import warnings
 from functools import partial, partialmethod
 from inspect import (  # NOQA
-    isclass, ismethod, ismethoddescriptor, isroutine
+    isclass, ismethod, ismethoddescriptor, unwrap
 )
 from io import StringIO
 from typing import Any, Callable, Mapping, List, Tuple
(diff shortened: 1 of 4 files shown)


sphinx/util/inspect.py
```diff
@@ -111,11 +111,16 @@ def getargspec(func):
                                kwonlyargs, kwdefaults, annotations)
 
 
-def unwrap(obj: Any) -> Any:
-    """Get an original object from wrapped object."""
+def unwrap_all(obj: Any) -> Any:
```
tk0miya, line 114: -1 for renaming. I think python's and sphinx's are almost same.

tk0miya (commented):
LGTM with nits.

sphinx/util/inspect.py
terencehonles: @tk0miya sorry for the late reply, hope you are doing well with everything that is going on in the world...

I split this out because one of the tests would break if I used this form of `unwrap` because it would take off the classmethod/staticmethod descriptors. `unwrap` from the stdlib will only take away `__wrapped__` functions which are going to only be defined if you had wrapped with `@functools.wraps` or one of the other functions that was designed to emulate a wrapped function this way. In this PR you will find both uses of `inspect.unwrap` and `inspect.unwrap_all`.

To save you some looking for the `unwrap` which is needed from the stdlib:
- https://github.com/sphinx-doc/sphinx/pull/7222/files#diff-48ca4d6544b092f35d729127500bff5fR979
- https://github.com/sphinx-doc/sphinx/pull/7222/files#diff-48ca4d6544b092f35d729127500bff5fR1341
- https://github.com/sphinx-doc/sphinx/pull/7222/files#diff-84bfd2ad46b7c8e97e5a06abb886c065R207

tk0miya: Okay, I understand.

PR #7224, opened 2020-02-29
Fix #7223: Sphinx builds has been slower since 2.4.0

### LINKED ISSUE #7223: Upgrading my project to version 2.4.0+ multiplies x10 output time
I would like to upgrade my project to the latest Sphinx version but I have encountered myself that if I upgrade to version 2.4.0+, when the project is in phase "writing output" it takes >10min to finish, when any other version (eg 2.3.1 and lower until 1.3.1) takes around 100s.

I have discovered that the bottleneck is in the sidebar template: 
```
<div id="toctree">
{{ toctree(maxdepth=6|toint, collapse=False,includehidden=True|tobool) }}
</div>
```
If I dont add a sidebar then everything is smooth as always, but I do need the sidebar with the entire tree for navigation purporses.

Touching here and there the installed package I have discovered that removing in file `environment/adapters/toctree.py` line 218 `entries.extend(children)` stops this slow performance, but of course the toctree creation disappears (in version 2.4.0).

My project is quite big, it has over 400 source files so I understand building the toctree is time consuming, but honestly upgrading and multiplying by 10 the output time...stops me from upgrading to the latest version. 

My source files didnt changed from one version to another and I am using  Python 3.7.6. 

I cannot share with you the source files because the project is private but in case it helps you here goes a link to the current documentation site https://doc.synapticon.com/ (which uses Python 2.7.15 and Sphinx 1.3.1, hence the urge to upgrade) and the conf file of the current under progress upgraded project (to new version):

```
import sys, os, re
import sphinx_bootstrap_theme

project = 'sncn_documentation'
copyright = u''
author = 'Begoña Álvarez - Boxfish Studio SL'
sys.path.append("./ext")
extensions = [
    'sphinx.ext.ifconfig',
    'sphinx.ext.mathjax',
    'div',
    'tabselection',
]
exclude_patterns = [
    '_build', '_static', 'ext', '_templates', 'Thumbs.db', 
    '.DS_Store', '.support*','.sources*','.doxygen*', 
    'README.rst','**/README.rst','**/pseudocode.rst',
    '**/CHANGELOG.rst', '*.#*'
    ]
html_theme = 'bootstrap'
html_title = "Synapticon Documentation"

html_sidebars = {'**/**': ['toctree.html']}
html_show_sourcelink = False
```

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7223 
- debug message should not be evaluated when not debug mode.

### DIFF
--- a/sphinx/util/docutils.py
+++ b/sphinx/util/docutils.py
@@ -479,8 +479,8 @@ class SphinxTranslator(nodes.NodeVisitor):
         for node_class in node.__class__.__mro__:
             method = getattr(self, 'visit_%s' % (node_class.__name__), None)
             if method:
-                logger.debug('SphinxTranslator.dispatch_visit calling %s for %s' %
-                             (method.__name__, node))
+                logger.debug('SphinxTranslator.dispatch_visit calling %s for %s',
+                             method.__name__, node)
                 return method(node)
         else:
             super().dispatch_visit(node)
@@ -497,8 +497,8 @@ class SphinxTranslator(nodes.NodeVisitor):
         for node_class in node.__class__.__mro__:
             method = getattr(self, 'depart_%s' % (node_class.__name__), None)
             if method:
-                logger.debug('SphinxTranslator.dispatch_departure calling %s for %s' %
-                             (method.__name__, node))
+                logger.debug('SphinxTranslator.dispatch_departure calling %s for %s',
+                             method.__name__, node)
                 return method(node)
         else:
             super().dispatch_departure(node)
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,7 @@ Bugs fixed
 ----------
 
 * #7197: LaTeX: platex cause error to build image directive with target url
+* #7223: Sphinx builds has been slower since 2.4.0
 
 Testing
 --------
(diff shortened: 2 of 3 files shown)


jfbu (commented):
There is also https://github.com/sphinx-doc/sphinx/blob/942dbc6be6db6fed580b16e3295d1b408b29cf94/sphinx/registry.py#L348-L350 but probably it is irrelevant/benign ?

tk0miya:
Good point. It might not be related with this problem. But I fixed it too.

sphinx/util/docutils.py
```diff
@@ -479,8 +479,8 @@ def dispatch_visit(self, node):
         for node_class in node.__class__.__mro__:
             method = getattr(self, 'visit_%s' % (node_class.__name__), None)
             if method:
-                logger.debug('SphinxTranslator.dispatch_visit calling %s for %s' %
-                             (method.__name__, node))
```
tk0miya, line 483: Before this change, this line goes to format message before calling `logger.debug()` even not in debug mode. And it sometime causes large node-to-string conversion always. Because `%` operator is stronger than method calling.

After the change, the conversion will go only in debug mode. So we should give parts of messages via parameters of `logger.debug()`.

jfbu: This is a subtle find...

PR #7230, opened 2020-02-29
std domain: Generate node_id for objects in the right way

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6903, #7210


sphinx/domains/std.py
```diff
@@ -214,16 +214,25 @@ def handle_signature(self, sig: str, signode: desc_signature) -> str:
...
-            signode['names'].append(targetname)
+                prefixes.append(currprogram)
+            if not optname.startswith(('-', '/')):
+                prefixes.append('arg')
+            prefix = '-'.join(prefixes)
+            node_id = make_id(self.env, self.state.document, prefix, optname)
+            signode['ids'].append(node_id)
+
+            # Assign old styled node_id not to break old hyperlinks (if possible)
+            # Note: Will be removed in Sphinx-5.0 (RemovedInSphinx50Warning)
```
tk0miya: Unfortunately, this does not helps old hyperlinks because old anchors are generated via "names".

PR #7236, opened 2020-03-01
py domain: Generate node_id for objects and modules in the right way

tk0miya:
### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
- refs: #6903, #7210


doc/usage/extensions/autosummary.rst
```diff
@@ -273,6 +273,7 @@ Additionally, the following filters are available
    replaces the builtin Jinja `escape filter`_ that does html-escaping.
 
 .. function:: underline(s, line='=')
+   :noindex:
```
tk0miya, line 276: I think this would be an another bug. I'll take a look carefully tomorrow.

PR #7249, opened 2020-03-04
Use "docs" extra on CircleCI

blueyed:
"websupport" is not in `extras_require`.

It might not be necessary at all - but I expect it to result in more coverage likely.
If not necessary/wanted it should be removed instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,5 +7,5 @@ jobs:
     steps:
       - checkout
       - run: /python3.6/bin/pip install -U pip setuptools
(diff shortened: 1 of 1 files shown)


blueyed:
Doesn't seem to make a difference:

Latest 3.x build:

> 1379 passed, 10 skipped, 6 warnings

This:

> 1379 passed, 10 skipped, 6 warnings

So it looks like it can be removed?

tk0miya:
Indeed. It might not be necessary. Let's remove it.

blueyed:
Amended.

tk0miya:
Thanks!

PR #7250, opened 2020-03-05
Use a string with warning in `handle_signature`

blueyed:
Previously it would pass in the exception object itself, which then
might crash filters that are using `record.msg.startswith` etc.

The warning was triggered for me with `sig = foo: int = -1`, where it
appears to not handle negative numbers (have not investigated, Python 3.8.1).

The code was added recently (c4d7f4d6c), so this is targeting 3.x.


tk0miya:

Issue #7243, opened 2020-03-03
py domain: All `:py:*` roles can refer python objects even if objtype not matched.


tk0miya:
**Describe the bug**
py domain: All `:py:*` roles can refer python objects even if objtype not matched.

**To Reproduce**

Both `:py:meth:` and `:py:class:` are converted into hyperlinks to `.. py:function:: foo`.

```
.. py:function:: foo

:py:meth:`foo`
:py:class:`foo`
```

**Expected behavior**
They fails to create hyperlinks for python object when objtype is different.

**Your project**
N/A

**Screenshots**
N/A

**Environment info**
- OS: Mac
- Python version: 3.8.2
- Sphinx version: 2.4.4
- Sphinx extensions:  N/A

tk0miya:
I found same problem in C domain:
```
.. c:function:: hello()

:c:member:`hello`
:c:macro:`hello`
:c:func:`hello`
```

jakobandersen:
First: @tk0miya, thanks for all the work and maintenance! It's really nice.

I'll look more closely at the C domain soon(TM), but just a few semi-relevant comments for now.

This is definitely an issue that should be resolve in each domain, though I'm not entirely sure it should be solved in the same style everywhere, nor that it shouldn't be. Different languages/domains may have different semantics that begs for different resolutions.
In general, for the programming language domains, I believe that xref roles should resolve almost* independent of object types. Once resolved, then if there is a mismatch of role type and object type, a warning should be issued.
E.g., for the example:
```rst
.. py:function:: foo

:py:meth:`foo`, great, role type matches the object it resolves to
:py:class:`foo`, warning, target ``foo`` is a function but the xref role is a class
```
That is, the second xref becomes the exact same link** as the first. This is the scheme I implemented in the C++ domain.
From a brief view, PR #7256 implements the alternative where the second link triggers a warning about ``foo`` not found, which I personally find misleading.***

`*` Different roles may parse their argument with differently. E.g., the ``cpp:func`` role allows for specifying a specific function overload. This parsing option is not enabled in the other roles.
`**` Different roles may transform the link text in different ways. Notably, the function roles in multiple (all?) domains has the ``fix_parens`` option enabled, which ensures that the link text ends with ``()``, e.g., ``:py:meth:`foo` `` becomes a link with text ``foo()``.
`***` When time allows I can make the necessary fixes in the C domain, though it may be some more weeks before I get to it.

tk0miya:
[...]

PR #7256, opened 2020-03-05
All :c:*: roles should check objtype on resolving refs (refs: #7243)

tk0miya:
### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
All :c:*: roles can refer all kind of C objects even if objtype
not matched.  To search objects correctly, this change the form of
key for objects to a pair of objtype and objname.
refs: #7243 

### Remaining tasks

- [x] tests

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -271,57 +271,64 @@ class CDomain(Domain):
     }  # type: Dict[str, Dict[str, Tuple[str, Any]]]
 
     @property
-    def objects(self) -> Dict[str, Tuple[str, str]]:
-        return self.data.setdefault('objects', {})  # fullname -> docname, objtype
+    def objects(self) -> Dict[Tuple[str, str], str]:
+        return self.data.setdefault('objects', {})  # objtype, fullname -> docname
 
     def note_object(self, name: str, objtype: str, location: Any = None) -> None:
-        if name in self.objects:
-            docname = self.objects[name][0]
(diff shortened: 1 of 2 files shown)


sphinx/domains/c.py
```diff
@@ -271,65 +271,72 @@ class CDomain(Domain):
     }  # type: Dict[str, Dict[str, Tuple[str, Any]]]
 
     @property
-    def objects(self) -> Dict[str, Tuple[str, str]]:
-        return self.data.setdefault('objects', {})  # fullname -> docname, objtype
+    def objects(self) -> Dict[Tuple[str, str], str]:
+        return self.data.setdefault('objects', {})  # objtype, fullname -> docname
```
tk0miya, line 275: I'll change the structure of `self.objects` later to add node_id into the value in next PR. It is also needed to avoid conflicts with other domains (refs: #6903).

tk0miya:
I worry about cpp domain also have same issue. But it is too complicated to me.

jakobandersen:
As hinted over in #7243, I'm not sure this is a good resolution for C. Instead of things silently linking but with misleading rst code, then now there are no links but with warnings about targets not being found. As one really shouldn't have two entities with the same name (C does not support overloading) I believe warnings on role/objtype mismatch is a better resolution. If you are ok with this, I'll make a PR with a solution along those lines.

tk0miya:
Your idea seems better than mine. Withdrawing with pleasure :-)

PR #7263, opened 2020-03-06
[2.4.x] Specify the target Python version in the mypy configuration

blueyed:
Subject: backport of https://github.com/sphinx-doc/sphinx/pull/6648 to fix mypy on 2.x

But this also appears to fix a real bug after all:

> This caught a bug. The subprocess.run() function only started taking the 'encoding' keyword argument starting with Python 3.6.

(cherry picked from commit acb2eadc8e692be4c8212cf1d3697e10e298a384)

Conflicts (trivial):
	setup.cfg

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,7 @@ paths =
     .
 
 [mypy]
+python_version = 3.5
 disallow_incomplete_defs = True
 show_column_numbers = True
 show_error_context = True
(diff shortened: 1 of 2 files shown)


tk0miya:
Thanks!

tk0miya:
Note: On 4.0 branch, we can upgrade to python 3.6. So I just post #7264 to revert this.

PR #7264, opened 2020-03-07
Drop py35 support

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- 4.0 will be released in 2021 Spring. At that time, Ubuntu 16.04 will become EOL. Therefore, we can update dependencies for Ubuntu 18.04 based one.
- For 4.0 development, we can use syntax and features since python 3.6; variable annotations, integrated pathlib support and so on.
- As a first step, this drops python3.5 and docutils-0.14 supports.

### DIFF
--- a/sphinx/util/nodes.py
+++ b/sphinx/util/nodes.py
@@ -11,7 +11,7 @@
 import re
 import warnings
 from typing import Any, Callable, Iterable, List, Set, Tuple, Type
-from typing import cast
+from typing import TYPE_CHECKING, cast
 
 from docutils import nodes
 from docutils.nodes import Element, Node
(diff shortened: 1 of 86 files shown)


sphinx/util/nodes.py
```diff
@@ -537,15 +535,15 @@ def process_only_nodes(document: Node, tags: "Tags") -> None:
                 node.replace_self(nodes.comment())
 
 
-# monkey-patch Element.copy to copy the rawsource and line
-# for docutils-0.14 or older versions.
-
 def _new_copy(self: Element) -> Element:
+    """monkey-patch Element.copy to copy the rawsource and line
+    for docutils-0.14 or older versions."""
```
tk0miya: Unfortunately, this patch is still needed. I found another error on copying system_message node...
refs: https://sourceforge.net/p/docutils/patches/165/

tk0miya:
Note: Python 3.5 will become EOL in 2020-09-13. And this change will be released after the EOL.
https://devguide.python.org/#status-of-python-branches

PR #7267, opened 2020-03-07
Use absolute path as location for logging records

blueyed:
autodoc passes in an absolute path here already (ending in `.rst`),
which then resulted in `/path/to/doc.rst.rst` (with incorrect extension)
being logged in warnings.

Hopefully some test(s) fail that can be adjusted.
Not sure about the target branch also - should it go into 3.x instead?

### DIFF
--- a/sphinx/util/logging.py
+++ b/sphinx/util/logging.py
@@ -10,6 +10,7 @@
 
 import logging
 import logging.handlers
+import os
 from collections import defaultdict
 from contextlib import contextmanager
 from typing import Any, Dict, Generator, IO, List, Tuple, Union
@@ -449,10 +450,13 @@ class SphinxLogRecordTranslator(logging.Filter):
         location = getattr(record, 'location', None)
         if isinstance(location, tuple):
             docname, lineno = location
-            if docname and lineno:
-                record.location = '%s:%s' % (self.app.env.doc2path(docname), lineno)
-            elif docname:
-                record.location = '%s' % self.app.env.doc2path(docname)
+            if docname:
+                if os.path.isabs(docname):
+                    record.location = docname
+                else:
+                    record.location = self.app.env.doc2path(docname)
+                if lineno:
+                    record.location += ":{}".format(lineno)
             else:
                 record.location = None
         elif isinstance(location, nodes.Node):


sphinx/util/logging.py
```diff
@@ -449,10 +450,13 @@ def filter(self, record: SphinxWarningLogRecord) -> bool:  # type: ignore
...
-                record.location = '%s:%s' % (self.app.env.doc2path(docname), lineno)
-            elif docname:
-                record.location = '%s' % self.app.env.doc2path(docname)
+            if docname:
+                if os.path.isabs(docname):
+                    record.location = docname
+                else:
+                    record.location = self.app.env.doc2path(docname)
+                if lineno:
+                    record.location += ":{}".format(lineno)
```
blueyed, line 459: Initially I've thought about fixing/handling this in doc2path - but I've figured that calling it with a path already is wrong, but have not checked other usages of it.

tk0miya:
Could you share an example for wrong paths? I think autodoc should be fixed instead.

blueyed:
> Could you share an example for wrong paths? I think autodoc should be fixed instead.

Sure, sorry for missing it earlier: https://github.com/sphinx-doc/sphinx/blob/60fbf4105ceeee588ac6f9e700ba529bfaaa5e5b/sphinx/ext/autodoc/directive.py#L124-L140

tk0miya:
Thank you for reporting. I thought #7273 is better to fix the problem instead.

sphinx/util/logging.py
blueyed: Since https://github.com/sphinx-doc/sphinx/pull/7273 will be taken instead I've wondered if it would make sense to assert / and or warn that the location is always a docname / no full path here then?

tk0miya: -1 for assert. Surely it would help Sphinx developers, but it brings trouble to users.
+0 for warning. It might help Sphinx developers (if they find warnings) and it does not harm users.

tk0miya:
Oh, sorry. GH automatically closes this by merging #7273 . Reopened.

sphinx/util/logging.py
blueyed: Yeah, the "assert" should certainly not be used in production, but e.g. in tests.

Therefore a warning seems feasible: it could have a type that breaks tests, but otherwise gets only displayed / can be filtered.

tk0miya: Some user uses sphinx via `python -m sphinx`. If my understanding correct, we should notice them to give `-O` option to invoke Sphinx, right?
It would be wonderful if it works only in testing.

blueyed: I think using `assert` is not really good, but will evaluate if a special warning class can be used, that would be not shown by default (but also would not hurt really I guess?), and cause an error during tests.

Will close this PR, and keep that in mind/inbox to follow up eventually.

blueyed:
> Oh, sorry. GH automatically closes this by merging #7273 . Reopened.

Can be closed.

PR #7269, opened 2020-03-07
c domain: Generate node_id for objects in the right way

tk0miya:
### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
- refs: #6903

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -27,7 +27,7 @@ from sphinx.locale import _, __
 from sphinx.roles import XRefRole
 from sphinx.util import logging
 from sphinx.util.docfields import Field, TypedField
-from sphinx.util.nodes import make_refnode
+from sphinx.util.nodes import make_id, make_refnode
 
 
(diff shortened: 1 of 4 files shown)


sphinx/domains/c.py
```diff
@@ -197,21 +197,23 @@ def get_index_text(self, name: str) -> str:
             return ''
 
     def add_target_and_index(self, name: str, sig: str, signode: desc_signature) -> None:
-        # for C API items we add a prefix since names are usually not qualified
-        # by a module name and so easily clash with e.g. section titles
-        targetname = 'c.' + name
-        if targetname not in self.state.document.ids:
-            signode['names'].append(targetname)
```
tk0miya, line 204: I think mention in CHANGES is not needed for C domain. Because it might be not official and hidden feature.
```
.. c:function:: func

ref with `c.func`_
```

tk0miya:
Note: This will conflict with #7256. I'll update either this or #7256 on merging.

PR #7277, opened 2020-03-07
Fix #7276: cpp: objects generate hypertarget names unexpectedly

### LINKED ISSUE #7276: cpp: objects generate hypertarget names unexpectedly
**Describe the bug**
The C++ objects generate hyperlink names internally. As a result, it conflicts with normal hyperlink target.

**To Reproduce**
Input:
```
.. _say_hello:

.. cpp:function:: say_hello()
```
warnings:
```
/Users/tkomiya/work/tmp/doc/index.rst:14: WARNING: Duplicate explicit target name: "say_hello".
```

As a side effect, C++ objects can be refer via hyperlink reference syntax:
```
.. cpp:function:: say_hello()

Both :cpp:func:`say_hello` and say_hello_ can refer above definition
```

**Expected behavior**
I suppose this is not unexpected behavior. So it should not define hyperlink names internally.

**Your project**
N/A

**Screenshots**
N/A

**Environment info**
- OS: Mac
- Python version: 3.8.2
- Sphinx version: 2.4.4
- Sphinx extensions:  No
- Extra tools: No

tk0miya:
### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
- refs: #7276 
- I think this would not be an expected behavior. So I'm going to remove `signode['names']` assignment. Is my understand correct?
- I also thought to change node_id generation mechanism to use `make_id()`. But cpp domain is too complicated to me. And it seems it already has its own ID generation engine. So it does not cause a problem like #6903. So I'd like to keep as is now.

@jakobandersen Could you review this and give an advice to me.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,8 @@ Incompatible changes
 * #6903: Internal data structure of Python, reST and standard domains have
   changed.  The node_id is added to the index of objects and modules.  Now they
   contains a pair of docname and node_id for cross reference.
+* #7276: C++ domain: Non intended behavior is removed such as ``say_hello_``
+  links to ``.. cpp:function:: say_hello()``
 * #7210: js domain: Non intended behavior is removed such as ``parseInt_`` links
   to ``.. js:function:: parseInt``
 * #7229: rst domain: Non intended behavior is removed such as ``numref_`` links
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -6512,10 +6512,6 @@ class CPPObject(ObjectDescription):
             names = self.env.domaindata['cpp']['names']
             if name not in names:
                 names[name] = ast.symbol.docname
-                signode['names'].append(name)
-            else:
-                # print("[CPP] non-unique name:", name)
-                pass
             # always add the newest id
             assert newestId
             signode['ids'].append(newestId)


CHANGES
```diff
@@ -26,6 +26,8 @@ Incompatible changes
 * #6903: Internal data structure of Python, reST and standard domains have
   changed.  The node_id is added to the index of objects and modules.  Now they
   contains a pair of docname and node_id for cross reference.
+* #7276: c domain: Non intended behavior is removed such as ``say_hello_`` links
```
jakobandersen: Typo, I guess: "c" -> "C++"

sphinx/domains/cpp.py
```diff
@@ -6512,10 +6512,6 @@ def add_target_and_index(self, ast: ASTDeclaration, sig: str, signode: desc_sign
             names = self.env.domaindata['cpp']['names']
             if name not in names:
                 names[name] = ast.symbol.docname
-                signode['names'].append(name)
```
jakobandersen, line 6515: Hmm, I can reproduce the warning this is supposed to fix, but I don't actually know where it comes from. Do you know where it is generated from? (perhaps inside docutils?) In short: besides avoiding the warning I'm actually not sure what the specific effect of removing this line is.

jakobandersen:
> * I also thought to change node_id generation mechanism to use `make_id()`. But cpp domain is too complicated to me. And it seems it already has its own ID generation engine. So it does not cause a problem like #6903. So I'd like to keep as is now.

Indeed. IDs are generated from the declaration such that clashes can never happen (assuming no bugs). As features are added there has been a need to change IDs for existing objects. There are thus multiple IDs for the same entity (currently ``_max_id = 4``), each with their own prefix to avoid clashes (see the ``_id_prefix`` variable). This makes old links still work, though some day the oldest ones should probably be removed.

sphinx/domains/cpp.py
tk0miya: Yes, the warning came from inside docutils.

>In short: besides avoiding the warning I'm actually not sure what the specific effect of removing this line is.

In docutils world, `node['names']`  is used to create a hyperlink target that is able to refer via its name. On the other hand, Sphinx provides some roles (ex. `:cpp:func:`) to realize the cross-reference feature because docutils does not support a reference between multiple documents. Therefore, it is not needed to generate names for the hyperlink target.

CHANGES
tk0miya: oops... fixed.

sphinx/domains/cpp.py
jakobandersen: Thanks, looks good to merge.

tk0miya:
Thank you for reviewing! merged.

PR #7278, opened 2020-03-07
Fix use of html_file_suffix instead of html_link_suffix in search results

rdb:
The links in the search results have `html_file_suffix` appended instead of `html_link_suffix`.  This fixes that.

To reproduce the bug, build any documentation with `html_file_suffix` set to ".html" and `html_link_suffix` set to "".  The regular links between pages will be extensionless, but the search results will have .html at the end.

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -473,6 +473,7 @@ class StandaloneHTMLBuilder(Builder):
             'show_source': self.config.html_show_sourcelink,
             'sourcelink_suffix': self.config.html_sourcelink_suffix,
             'file_suffix': self.out_suffix,
+            'link_suffix': self.link_suffix,
             'script_files': self.script_files,
             'language': self.config.language,
             'css_files': self.css_files,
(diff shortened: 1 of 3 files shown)


blueyed:
I'm not familiar with Sphinx' tests yet, but think it would be great to have a test for this - in case you are more familiar with it already.. :)

tk0miya (changes requested):
LGTM. But it would be better to merge this into 3.x branch. So please resolve conflicts with 3.x branch. Then I'll merge this.

Note: Unfortunately, there are no good testcases for search JS. You can add a case to `tests/js/doctools.js`, but it seems many works are needed to do that. So I'm okay with no testing.

tk0miya:
BTW, I don't still understand how useful `html_link_suffix` is. Could you let me know what purpose you've used if possible.

rdb:
OK, I will do that.

In my set-up, I want to upload .html files to my server, but I want the user to see clean URLs without the .html extension at the end.  So I configured my web server to automatically add on the .html extension when asking for a URL without extension, and set html_link_suffix to the empty string.  This generally works pretty well:
https://docs.panda3d.org/1.10/python/index

tk0miya:
Thank you for update and explanation. I understand.

tk0miya (approved):

tk0miya:

Issue #7279, opened 2020-03-08
ci: improve codecov integration


blueyed:
I'd like to disable codecov's comments (which are large and interrupt the conversion on PRs).

IIRC codecov statuses have been disabled (because of flakiness?), but comments where never a topic.

I therefore think it would be good to 1. disable codecov's comments, and 2. enable the status for the patch.

This would make PRs less "noisy" with regard to comments, and the "patch" check from Codecov would indicate how much the patch/PR is covered.

tk0miya:
@stephenfin Could you check this please?

Personally, I don't refer status of codecov. So +1 for disabling.

stephenfin:
+0. I thought they would be helpful but if not, kill em

blueyed:
I think the comments are getting in the way a lot, but value the status checks in general (admins can override merging (if checks are required to pass in the first place), e.g. when something is not covered).

I've also enabled the status check in https://github.com/sphinx-doc/sphinx/pull/7286, but can remove it again.

tk0miya:
It seems nobody objects to your proposal. So let's merge it. It seems status checks for patches are not bother to us.

PR #7286, opened 2020-03-09
Improve codecov integration

blueyed:
Ref: https://github.com/sphinx-doc/sphinx/issues/7279 - let's discuss/decide it there.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,4 @@
+comment: false
 coverage:
   status:
     project:
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,6 +4,3 @@ coverage:
     project:
       default:
         enabled: no
-    patch:
-      default:
-        enabled: no


tk0miya:
Hmm... after merging this, pytest-cov raise an exception. @blueyed do you know what happened?
```
coverage.misc.CoverageException: Can't combine arc data with line data
```

blueyed:
@tk0miya 
> Hmm... after merging this, pytest-cov raise an exception.

> ```
> coverage.misc.CoverageException: Can't combine arc data with line data
> ```

Should not be related to codecov, but rather the pytest unpinning (if not "flaky").

This typically happens when coverage runs with and without branch=1, and this then is being tried to be combined.  It can happen e.g. when a subprocess does not use the config with branch=1.

Is it happening constantly now?

btw: I'm also in the IRC channel, if you are using that.

tk0miya:
No, it seems unrelated. because unpinning was merged into 3.x branch. But the error was happened on 2.4.x branch.
The same error was also raised in 3.x branch. But this change is not merged into 3.x branch yet. So I guess these errors happened unrelated with these PRs.
https://travis-ci.org/sphinx-doc/sphinx/jobs/660199606

tk0miya:
Sorry for posting a comment in the middle. I've not used IRC.

tk0miya:
>Is it happening constantly now?

I restarted the failed CI manually twice, but all tries were failed with the same error. I'm going to stop sending coverage data temporarily to recover our CI.

tk0miya:
For now, CI is recovered with stopping coverage. It would be better if we can investigate why the error has happened.
Thanks,

PR #7290, opened 2020-03-10
Ignore bdb.BdbQuit when handling exceptions

blueyed:
`bdb.BdbQuit` is used when quitting the debugger.
It should not a) cause the debugger to be started (with `-P` / `--pdb`),
and b) not a "crash" to be logged.

This helps when using `pdb.set_trace()` manually, and quitting it with
`q`.

It gets used in `build_main`, and `BuildDoc.run` (distutils command).

### DIFF
--- a/sphinx/cmd/build.py
+++ b/sphinx/cmd/build.py
@@ -29,6 +29,11 @@ from sphinx.util.docutils import docutils_namespace, patch_docutils
 
 
 def handle_exception(app: Sphinx, args: Any, exception: BaseException, stderr: IO = sys.stderr) -> None:  # NOQA
+    import bdb
+
+    if isinstance(exception, bdb.BdbQuit):
+        return
+
     if args.pdb:
         import pdb
         print(red(__('Exception occurred while building, starting debugger:')),


sphinx/cmd/build.py
```diff
@@ -29,6 +29,11 @@
 
 
 def handle_exception(app: Sphinx, args: Any, exception: BaseException, stderr: IO = sys.stderr) -> None:  # NOQA
+    import bdb
```
tk0miya, line 32: Why not importing this at the top of file?

blueyed: Thought this would be more in line with the lazy `import pdb` below.

I think `handle_exception` is worth importing lazily, since it is not expected to be used in the common case.

Happy to move it though, of course.

tk0miya: IMO, it would be better to move all imports to the top of the script if no special reason (avoiding circular imports, becomes slow and so on). If lazy loading is needed for some special reason, adding comments will help us in the future.

tk0miya: I'll fix this after merging.

PR #7291, opened 2020-03-10
Handle singledispatch functions with rewritten signatures.

anntzer:
Subject: Handle singledispatch functions with rewritten signatures.

### Feature or Bugfix
- Bugfix

### Purpose
If a singledispatch function has its `__signature__` rewritten, autodoc
fails to annotate that signature because `func.__annotations__` is not
consulted.  Instead, directly assign to `__signature__` instead.

(See example with `@assign_signature` in tests.)

### Relates
- https://github.com/sphinx-doc/sphinx/pull/7234 (thanks for the original implementation!)

---

This fails CI because mypy emits the spurious
```
sphinx/ext/autodoc/__init__.py: note: In member "annotate_to_first_argument" of class "SingledispatchFunctionDocumenter":
275sphinx/ext/autodoc/__init__.py:1113:13: error: "Callable[..., Any]" has no attribute "__signature__"
```
which is clearly wrong, but I'm not sure how to placate it.

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -13,6 +13,7 @@
 import importlib
 import re
 import warnings
+from inspect import Parameter
 from types import ModuleType
 from typing import Any, Callable, Dict, Iterator, List, Sequence, Set, Tuple, Type, Union
 from unittest.mock import patch
@@ -1108,9 +1109,10 @@ class SingledispatchFunctionDocumenter(FunctionDocumenter):
         if len(sig.parameters) == 0:
             return
 
-        name = list(sig.parameters)[0]
-        if name not in func.__annotations__:
-            func.__annotations__[name] = typ
+        params = list(sig.parameters.values())
+        if params[0].annotation is Parameter.empty:
+            params[0] = params[0].replace(annotation=typ)
+            func.__signature__ = sig.replace(parameters=params)  # type: ignore
 
 
 class DecoratorDocumenter(FunctionDocumenter):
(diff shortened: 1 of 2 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -1107,9 +1108,10 @@ def annotate_to_first_argument(self, func: Callable, typ: Type) -> None:
         if len(sig.parameters) == 0:
             return
 
-        name = list(sig.parameters)[0]
-        if name not in func.__annotations__:
-            func.__annotations__[name] = typ
+        param_list = list(sig.parameters.values())
```
tk0miya: How about `params` or `parameters` for the variable? I prefer not to use its structures for the varname.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1107,9 +1108,10 @@ def annotate_to_first_argument(self, func: Callable, typ: Type) -> None:
...
+        if param_list[0].annotation is Parameter.empty:
+            func.__signature__ = sig.replace(parameters=[
+                param_list[0].replace(annotation=typ), *param_list[1:]])
```
tk0miya: How about this?
```
params[0] = params[0].replace(annotation=typ)
func.__signature__ = sig.replace(parameters=params)
```

tk0miya (changes requested):
Thank you for PR. Looks good!

But we need to merge this into 3.x branch. Could you rebase this into 3.x branch?
In addition, it would be better to fix `SingledispatchMethodDocumenter` also.

sphinx/ext/autodoc/__init__.py
anntzer: done

sphinx/ext/autodoc/__init__.py
anntzer: done

anntzer:
thanks for the review, rebased to 3.x

tk0miya (approved):
LGTM! Thank you for quick updates!

PR #7298, opened 2020-03-11
py domain: Add py:property directive to describe a property (refs: #7068)

tk0miya:
### Feature or Bugfix
- Feature
- Refactoring

### Purpose
- refs: #7068

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -316,6 +316,22 @@ The following directives are provided for module and class contents:
 
       .. versionadded:: 4.0
 
+.. rst:directive:: .. py:property:: name
+
+   Describes an object property.
+
+   .. versionadded:: 4.0
+
+   .. rubric:: options
+
+   .. rst:directive:option:: abstractmethod
+      :type: no value
+
+      Indicate the property is abstract.
+
+   .. rst:directive:option:: type: type of the property
+      :type: text
+
 .. rst:directive:: .. py:method:: name(parameters)
 
    Describes an object method.  The parameters should not include the ``self``
@@ -367,6 +383,10 @@ The following directives are provided for module and class contents:
       Indicate the method is a property.
 
       .. versionadded:: 2.1
+
+      .. deprecated:: 4.0
+
+         Use :rst:dir:`py:property` instead.
 
    .. rst:directive:option:: staticmethod
       :type: no value
@@ -583,6 +603,8 @@ a matching identifier is found:
 .. rst:role:: py:attr
 
    Reference a data attribute of an object.
+
+   .. note:: The role is also able to refer to property.
 
 .. rst:role:: py:exc
 
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -825,6 +825,46 @@ class PyAttribute(PyObject):
         return _('%s (%s attribute)') % (attrname, clsname)
 
 
+class PyProperty(PyObject):
+    """Description of an attribute."""
+
+    option_spec = PyObject.option_spec.copy()
+    option_spec.update({
+        'abstractmethod': directives.flag,
+        'type': directives.unchanged,
+    })
+
+    def handle_signature(self, sig: str, signode: desc_signature) -> Tuple[str, str]:
+        fullname, prefix = super().handle_signature(sig, signode)
+
+        typ = self.options.get('type')
+        if typ:
+            signode += addnodes.desc_annotation(typ, ': ' + typ)
+
+        return fullname, prefix
+
+    def get_signature_prefix(self, sig: str) -> str:
+        prefix = ['property']
+        if 'abstractmethod' in self.options:
+            prefix.insert(0, 'abstract')
+
+        return ' '.join(prefix) + ' '
+
+    def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
+        name, cls = name_cls
+        try:
+            clsname, attrname = name.rsplit('.', 1)
+            if modname and self.env.config.add_module_names:
+                clsname = '.'.join([modname, clsname])
+        except ValueError:
+            if modname:
+                return _('%s (in module %s)') % (name, modname)
+            else:
+                return name
+
+        return _('%s (%s property)') % (attrname, clsname)
+
+
 class PyDecoratorMixin:
     """
     Mixin for decorator directives.
@@ -1056,6 +1096,7 @@ class PythonDomain(Domain):
         'classmethod':  ObjType(_('class method'),  'meth', 'obj'),
         'staticmethod': ObjType(_('static method'), 'meth', 'obj'),
         'attribute':    ObjType(_('attribute'),     'attr', 'obj'),
+        'property':     ObjType(_('property'),      'attr', '_prop', 'obj'),
         'module':       ObjType(_('module'),        'mod', 'obj'),
     }  # type: Dict[str, ObjType]
 
@@ -1068,6 +1109,7 @@ class PythonDomain(Domain):
         'classmethod':     PyClassMethod,
         'staticmethod':    PyStaticMethod,
         'attribute':       PyAttribute,
+        'property':        PyProperty,
         'module':          PyModule,
         'currentmodule':   PyCurrentModule,
         'decorator':       PyDecoratorFunction,
@@ -1205,8 +1247,17 @@ class PythonDomain(Domain):
                                 type, searchmode)
 
         if not matches and type == 'attr':
-            # fallback to meth (for property)
+            # fallback to meth (for property; Sphinx-2.4.x)
+            # this ensures that `:attr:` role continues to refer to the old property entry
+            # that defined by ``method`` directive in old reST files.
             matches = self.find_obj(env, modname, clsname, target, 'meth', searchmode)
+        if not matches and type == 'meth':
+            # fallback to attr (for property)
+            # this ensures that `:meth:` in the old reST files can refer to the property
+            # entry that defined by ``property`` directive.
+            #
+            # Note: _prop is a secret role only for internal look-up.
+            matches = self.find_obj(env, modname, clsname, target, '_prop', searchmode)
 
         if not matches:
             return None
--- a/tests/test_domain_py.py
+++ b/tests/test_domain_py.py
@@ -201,6 +201,10 @@ def test_resolve_xref_for_properties(app, status, warning):
             ' title="module_a.submodule.ModTopLevel.prop">'
             '<code class="xref py py-meth docutils literal notranslate"><span class="pre">'
             'prop</span> <span class="pre">method</span></code></a>' in content)
+    assert ('Link to <a class="reference internal" href="#module_a.submodule.ModTopLevel.prop"'
+            ' title="module_a.submodule.ModTopLevel.prop">'
+            '<code class="xref py py-attr docutils literal notranslate"><span class="pre">'
+            'prop</span> <span class="pre">attribute</span></code></a>' in content)
 
 
 @pytest.mark.sphinx('dummy', testroot='domain-py')
@@ -796,6 +800,29 @@ def test_pyattribute(app):
     assert_node(doctree[1][1][1][0][1][3], pending_xref, **{"py:class": "Class"})
     assert 'Class.attr' in domain.objects
     assert domain.objects['Class.attr'] == ('index', 'Class.attr', 'attribute', False)
+
+
+def test_pyproperty(app):
+    text = (".. py:class:: Class\n"
+            "\n"
+            "   .. py:property:: prop\n"
+            "      :abstractmethod:\n"
+            "      :type: str\n")
+    domain = app.env.get_domain('py')
+    doctree = restructuredtext.parse(app, text)
+    assert_node(doctree, (addnodes.index,
+                          [desc, ([desc_signature, ([desc_annotation, "class "],
+                                                    [desc_name, "Class"])],
+                                  [desc_content, (addnodes.index,
+                                                  desc)])]))
+    assert_node(doctree[1][1][0], addnodes.index,
+                entries=[('single', 'prop (Class property)', 'Class.prop', '', None)])
+    assert_node(doctree[1][1][1], ([desc_signature, ([desc_annotation, "abstract property "],
+                                                     [desc_name, "prop"],
+                                                     [desc_annotation, ": str"])],
+                                   [desc_content, ()]))
+    assert 'Class.prop' in domain.objects
+    assert domain.objects['Class.prop'] == ('index', 'Class.prop', 'property', False)
 
 
 def test_pydecorator_signature(app):
--- a/CHANGES
+++ b/CHANGES
@@ -66,6 +66,7 @@ Features added
   the location where the object is defined
 * #7199: py domain: Add :confval:`python_use_unqualified_type_names` to suppress
   the module name of the python reference if it can be resolved (experimental)
+* #7068: py domain: Add :rst:dir:`py:property` directive to describe a property
 * #7784: i18n: The alt text for image is translated by default (without
   :confval:`gettext_additional_targets` setting)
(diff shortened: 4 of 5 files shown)


sphinx/domains/python.py
```diff
@@ -674,6 +674,8 @@ def get_signature_prefix(self, sig: str) -> str:
         if 'classmethod' in self.options:
             prefix.append('classmethod')
         if 'property' in self.options:
+            logger.warning(__('Use the directive `py:property` instead of '
+                              'the directive `py:method` for properties.'))
```
tk0miya: Hmm... it seems it is better to separate deprecation to another PR to avoid breaking tests.

doc/usage/restructuredtext/domains.rst
```diff
@@ -253,6 +253,22 @@ The following directives are provided for module and class contents:
...
+   Describes an object property.
+
+   .. versionadded:: 3.0
+
+   .. rubric:: options
+
+   .. rst:directive:option:: abstractmethod
+      :type: no value
+
+      Indicate the property is an abstract method.
```
flying-sheep:
```suggestion
      Indicate the property is abstract.
```

Properties aren’t methods, they just happen to use `@abstractmethod`

doc/usage/restructuredtext/domains.rst
```diff
@@ -253,6 +253,22 @@ The following directives are provided for module and class contents:
...
+
+   .. rst:directive:option:: type: type of the property
+      :type: text
```
flying-sheep, line 333: Maybe we should also add the type the setter accepts? E.g. I have a property that can accept any `typing.PathLike` but always coerces it to `pathlib.Path`, so you can do:

```py
settings.path = './foo'
assert settings.path == Path('./foo')

File: doc/usage/restructuredtext/domains.rst [hunk 4]
```diff
@@ -506,6 +526,10 @@ a matching identifier is found:
 
    Reference a data attribute of an object.
 
+.. rst:role:: py:prop
+
+   Reference a property of an object.
```
flying-sheep:
```suggestion
   Reference a property or attribute of an object.
```

Since both can do both, this should also be the new description of `:attr:`

sphinx/domains/python.py
```diff
@@ -1159,8 +1202,14 @@ def resolve_xref(self, env: BuildEnvironment, fromdocname: str, builder: Builder
...
         if not matches and type == 'attr':
-            # fallback to meth (for property)
+            # fallback to prop (for property)
+            matches = self.find_obj(env, modname, clsname, target, 'prop', searchmode)
+        if not matches and type == 'attr':
+            # fallback to meth (for property; Sphinx-2.4.x)
             matches = self.find_obj(env, modname, clsname, target, 'meth', searchmode)
+        if not matches and type == 'meth':
+            # fallback to prop (for property)
+            matches = self.find_obj(env, modname, clsname, target, 'prop', searchmode)
```
flying-sheep: You also need to add the following, right?

Since people can transparently change attributes to properties, too?

```py
if not matches and type == 'prop':
    matches = self.find_obj(env, modname, clsname, target, 'attr', searchmode)
```

And maybe more informative comments would be:

1. for attr/prop:
   
   ```py
   # properties and attributes are transparently interchangeable,
   # so their roles need to be able to refer to the respective other
   ```

2. for `meth`:

   ```py
   # fallback to meth (in Sphinx 2.4.x, properties were stored as py:method)
   ```

tests/test_domain_py.py
```diff
@@ -178,6 +180,10 @@ def test_resolve_xref_for_properties(app, status, warning):
             ' title="module_a.submodule.ModTopLevel.prop">'
             '<code class="xref py py-meth docutils literal notranslate"><span class="pre">'
             'prop</span> <span class="pre">method</span></code></a>' in content)
+    assert ('Link to <a class="reference internal" href="#module-a-submodule-modtoplevel-prop"'
+            ' title="module_a.submodule.ModTopLevel.prop">'
+            '<code class="xref py py-prop docutils literal notranslate"><span class="pre">'
```
flying-sheep:
```suggestion
            '<code class="xref py py-attr py-prop docutils literal notranslate"><span class="pre">'
```

Backwards compatibility for themes without sacrificing the ability to style both differently.

flying-sheep (commented):
A few changes to address the concerns raised in #7068

doc/usage/restructuredtext/domains.rst
eric-wieser: Not sure how you'd want to write or display that. Perhaps the folks over at mypy can recommend a spelling?

flying-sheep: In code it’s obvious to express this: A return type annotation for the getter and a parameter type annotation on the setter.

In rST it could be `:get-type:` or `:type:`, and if a setter exists, also `:set-type:`.

I don’t know a concise way to display it (like “**Return Type**” for functions), except for spelling out “Any $Type can be assigned to it”

eric-wieser: Perhaps if `typing.Descriptor` could be added, as was suggested in the comment thread 
 at https://github.com/python/mypy/issues/244#issuecomment-620113658, then this could just be written `:type: Descriptor[GetType, SetType]`.

tk0miya: I think this should be an another topic. The goal of this PR is replacing property-description to `py:property` directive.

sphinx/domains/python.py
tk0miya: Never discussed that `py:prop` role can refer to attributes. And I don't think so.

flying-sheep: It should, as people can always also transparently change properties to attributes. It’s going to happen rarely, but it will.

eric-wieser: I haven't followed the full discussion here, but I think I side with @flying-sheep here - in my opinion either:

* Only `:attr:` should exist
* `:attr:` / `:prop:` should be completely interchangeable

I'd argue that while these's some value to describing `.. attribute` and `.. property` differently 
 to draw attention to dynamic behavior, there's no value in referencing them differently, as to a consumer there's no difference.

flying-sheep: That’s been my argument all along. @tk0miya thinks the consistency of having a one role per directive is important, while I tended towards going more in the “Only `:attr:` should exist” direction, because of the Python zen: “there should be one – and preferably only one – obvious way to do it”. But @tk0miya and me had a loooong discussion that we don’t want to repeat so I think the best compromise is “`:attr:` / `:prop:` should be completely interchangeable”

sphinx/domains/python.py
```diff
@@ -887,6 +887,46 @@ def get_index_text(self, modname: str, name_cls: Tuple[str, str]) -> str:
...
+            clsname, attrname = name.rsplit('.', 1)
+            if modname and self.env.config.add_module_names:
+                clsname = '.'.join([modname, clsname])
+        except ValueError:
+            if modname:
+                return _('%s (in module %s)') % (name, modname)
+            else:
+                return name
+
+        return _('%s (%s property)') % (attrname, clsname)
```
eric-wieser, line 865: Better to keep as much as possible out of the `try`:
```suggestion
        except ValueError:
            if modname:
                return _('%s (in module %s)') % (name, modname)
            else:
                return name

        if modname and self.env.config.add_module_names:
            clsname = '.'.join([modname, clsname])

        return _('%s (%s property)') % (attrname, clsname)
```

sphinx/domains/python.py
```diff
@@ -1016,6 +1056,23 @@ def process_link(self, env: BuildEnvironment, refnode: Element,
         return title, target
 
 
+class PyPropXRefRole(PyXRefRole):
+    """A special xref role class for :py:prop: role.
+
+    It adds a "py-attr" style to the role for backward compatibility.
+
+    Note: RemovedInSphinx50Warning
```
eric-wieser: @flying-sheep: This isn't about backwards-compatibility of _sphinx_, right? It's about backwards compatibility of libraries changing between `attr` and `prop` - so I don't think it makes sense to ever deprecate it.

sphinx/domains/python.py
tk0miya: @eric-wieser Who is the "producer" (provider?) and "consumer"?

I think no reason to describe them differently if they are transparent. Is it worthy to describe them differently? If so, they're different things, I think.

eric-wieser: To be clear, the "provider" is the docs that use `.. py:property`, such as a framework that provides the property. A "consumer" is a docstring in a docstream package that wants to reference the property.

> I think no reason to describe them differently if they are transparent. Is it worthy to describe them differently? If so, they're different things, I think.

Let's use `:abstract:` as an example - it's useful for readers to know that a method is abstract when reading its docstring, but when linking `` :abstract-meth:`some_meth` `` wouldn't be useful.

So by that reasoning I'd argue that `:property:` and `:abstract:` should be treated the same way:
```
.. method:: my_abstract_meth
    :abstract:

    This is the doc for :meth:`my_abstract_meth`, which I do not mention
    the abstract-ness of when I link.

.. attribute:: my_prop
    :property:

    This is the doc for :attr:`my_prop`, which I do not mention
    the property-ness of when I link.
```
I realize I'm at risk of reviving @flying-sheep's discussion here though.

tk0miya: Do you mean to use `py:attribute` directive to describe property (as an option)? If so, I can understand your idea. But I choose to add a new type `py:property` because it is unnatural to describe "abstract" or "async" properties via `py:attribute` directive. Additionally, client library sometimes needs to know the "member" of object is an attribute or a property (ex. assigning value, inheritance, etc).

Should we discuss "what is a property"?

sphinx/domains/python.py
flying-sheep: Yes, it’s about backwards compat of themes and custom css, and it doesn’t make sense to remove it. Most themes will probably forever style both kinds of links the same way.

In fact, I can’t think of a reason not to style the links the same way.

sphinx/domains/python.py
eric-wieser: I think this comment still applies?

tk0miya:
Finally, I determined to drop the `py:prop` role because I felt tired to discuss more.

sphinx/domains/python.py
eric-wieser: > Do you mean to use py:attribute directive to describe property (as an option)

Yes, I guess I was suggesting that

> because it is unnatural to describe "abstract" or "async" properties

The former is `@abstractproperty`, and the latter has a PyPI package; https://pypi.org/project/async-property/.

So I think actually that's an argument _for_ `:property:` just being metadata used in `..attribute`

sphinx/domains/python.py
```diff
@@ -1046,6 +1086,7 @@ class PythonDomain(Domain):
         'classmethod':  ObjType(_('class method'),  'meth', 'obj'),
         'staticmethod': ObjType(_('static method'), 'meth', 'obj'),
         'attribute':    ObjType(_('attribute'),     'attr', 'obj'),
+        'property':     ObjType(_('property'),      'prop', 'attr', 'obj'),
```
eric-wieser: I thought you said you had removed the `prop` role - but isn't that what this line is defining? Or do I misunderstand?

tk0miya: What do you mean? This is a definition of object-type, not a role.

eric-wieser: The documentation of `ObjType` says the arguments are:

> *roles*: all the roles that can refer to an object of this type

tk0miya: Ah, okay. understand.

sphinx/domains/python.py
```diff
@@ -1195,8 +1237,11 @@ def resolve_xref(self, env: BuildEnvironment, fromdocname: str, builder: Builder
                                 type, searchmode)
 
         if not matches and type == 'attr':
-            # fallback to meth (for property)
+            # fallback to meth (for property; Sphinx-2.4.x)
             matches = self.find_obj(env, modname, clsname, target, 'meth', searchmode)
+        if not matches and type == 'meth':
+            # fallback to attr (for property)
+            matches = self.find_obj(env, modname, clsname, target, 'attr', searchmode)
```
tk0miya: As a fallback for `meth` role, python domain searches the property using `attr` role. So it might hit an incorrect one.

eric-wieser: So as I understand it; the first `if` is for:

* Users who still use `..method` to define their properties
* Users loading an intersphinx file generated by an older sphinx that write `meth` to `objects.inv`.

While the latter `if` is for:

* Users who still use `:meth:` to refer to properties

Does it make sense to add a check that the object being referred to is actually a property?

tk0miya: I think a "prop" role is a polite way to do that. But it seems all of you don't like to introduce it. Please don't ask me.

eric-wieser: Ah, I finally understand what you're getting at. There's no way for sphinx to know if `` :meth:`maybe_prop` `` is a property  or not, as it can't see the old-style `:property:` marker; that information doesn't make it into `objects.inv`.

What I think I'm suggesting is that a warning should be emitted. I'll make another comment with a better set of lines selected...

eric-wieser: Actually, this is what I'd suggest:
```suggestion
            # this ensures that `:meth:` in old rst files can refer to `..property` in new rst files.
            matches = self.find_obj(env, modname, clsname, target, 'attr', searchmode)
            # don't allow `:meth:` ever to refer to regular attributes.
            matches = [match for match in matches if match[1] == 'property']
```

sphinx/domains/python.py
```diff
@@ -1195,8 +1237,11 @@ def resolve_xref(self, env: BuildEnvironment, fromdocname: str, builder: Builder
                                 type, searchmode)
 
         if not matches and type == 'attr':
-            # fallback to meth (for property)
+            # fallback to meth (for property; Sphinx-2.4.x)
```
eric-wieser, line 1250:
```suggestion
            # fallback to meth (for property; Sphinx-2.4.x)
            # this ensures that `:attr:` continues to refer to the deprecated `.. method::` `:property:` in
            # old rst files.
```

sphinx/domains/python.py
tk0miya: Why do you reinvent the prop role? Is it better to implement it for internal use?

eric-wieser: Adding a prop role wouldn't help, because this `if` block is dealing with old rST files which are incorrectly using a `:meth:` role. If you add a new `:prop:` role, these old files will still incorrectly use a `:meth:` role.

Conversely, new rST files have nothing to gain from a `:prop:` role, as the existing `:attr:` role is all they need.

The messiness of the code I suggest is simply due to trying to be backwards compatible. In sphinx 5, you could remove both of these if statements if you wanted.

tk0miya: I'd like to say use a prop role as a fallback of the `:meth:` role in the old reST files. It behaves the same as what your suggested code doing.
```
            matches = self.find_obj(env, modname, clsname, target, 'attr', searchmode)
            matches = [match for match in matches if match[1] == 'property']
```
and
```
            matches = self.find_obj(env, modname, clsname, target, 'prop', searchmode)
```
will be equivalent.

eric-wieser: Am I right in thinking that with that change, users will be able to type `` :prop:`...` `` in their rST files though, which is what we were trying to avoid? Or does "role" mean two different things here, one internal to sphinx and one exposed through rST

If the latter, how about `_prop_backcompat` to make it clear that this is for internal use?

tk0miya: Yes, I meant the latter one. I feel `_prop_backcompat` is too much. Adding a comment is enough helping us.

eric-wieser: How about just `_prop` then with a comment?

tk0miya: Okay, go with this.

tk0miya:
Merged. Thank you for your comments!

PR #7303, opened 2020-03-12
Revert "test: Disable codecov"

blueyed:
Reverts 1dba84096d888a15ae171c56572f12ca8d1cbd0d.

Will investigate (from) here.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,6 +59,7 @@ markers =
     apidoc
     setup_command
     test_params
+testpaths = tests
 
 [coverage:run]
 branch = True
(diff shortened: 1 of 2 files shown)


blueyed:
Seems to work (again).

While the config/setup can be improved, I'd suggest to re-enable it for now already.

setup.cfg
```diff
@@ -59,6 +59,7 @@ markers =
     apidoc
     setup_command
     test_params
+testpaths = tests
```
tk0miya, line 62: refs: https://docs.pytest.org/en/latest/reference.html#confval-testpaths

tk0miya (approved):
LGTM. Thank you for your work!

PR #7309, opened 2020-03-14
Propagate maxdepth option through package documents

garciparedes:
Subject: Propagate maxdepth option through package documents

### Feature or Bugfix

- Feature

### Purpose
Currently, the `maxdepth` parameter does only affect to the table of contents located at `tocfile`. I think that it could be more usefull if it applies to all recursively created table of contents (on subpackages or submodules).

### Detail
The applied change is so simple. It just injects the `maxdepth` parameter into the package's template and sets to the `:maxdepth:` option of the corresponding `toctree`.

### Relates
- Corresponding `sphinx-apidoc`'s documentation https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html#cmdoption-sphinx-apidoc-d

PR #7314, opened 2020-03-15
Propagate maxdepth option through package documents

garciparedes:
Subject: Propagate maxdepth option through package documents

### Feature or Bugfix

- Feature

### Purpose
Currently, the `maxdepth` parameter does only affect to the table of contents located at `tocfile`. I think that it could be more usefull if it applies to all recursively created table of contents (on subpackages or submodules).

### Detail
The applied change is so simple. It just injects the `maxdepth` parameter into the package's template and sets to the `:maxdepth:` option of the corresponding `toctree`.

### Relates
- Corresponding `sphinx-apidoc`'s documentation https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html#cmdoption-sphinx-apidoc-d
- Propagate maxdepth option through package documents #7309

### DIFF
--- a/sphinx/ext/apidoc.py
+++ b/sphinx/ext/apidoc.py
@@ -175,6 +175,7 @@ def create_package_file(root: str, master_package: str, subroot: str, py_files: 
         'separatemodules': opts.separatemodules,
         'automodule_options': options,
         'show_headings': not opts.noheadings,
+        'maxdepth': opts.maxdepth,
     }
     text = ReSTRenderer([user_template_dir, template_dir]).render('package.rst_t', context)
     write_file(pkgname, text, opts)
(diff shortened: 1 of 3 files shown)


garciparedes:
Hi @tk0miya !

I've a question about the test modifications I applied: 
Is it okay to adapt it by only adding the `:maxdepth: 4` (the default value) to each `toctree`, or it would  be preferable to modify with which value the tests have been launched (2 for example) and then check if   it has been applied to results? 

From one side, the answer would be negative because the default value is set on argparse parameter (it would work properly), but from the other side, the answer would be positive because is an `apidoc` functionality not checked right now. 

I think the first option is reasonable, but I wanted to tell you to know your opinion 🙂.

tk0miya:
Either is okay to me. IMO, it is better to add an additional case to confirm the output is changed by `--maxdepth` option explicitly. But we know it is already confirmed by the existing test case. So I'm merging this as is.

tk0miya:

PR #7315, opened 2020-03-15
Close #7293: html search: Allow to override JavaScript splitter

### LINKED ISSUE #7293: Overriding Python splitter results in mismatch with JavaScript splitter
**Describe the bug**
Sphinx has two ways to split words for searching. There is a Python splitter that parses words to include in the search index and a JavaScript splitter that parses user input to determine what words to look for in the search index. Users can override the Python splitter by subclassing `SearchLanguage`, but they cannot change the JavaScript splitter, which could result in terms in the search index that cannot be searched from an HTML page.

A specific case where this is a problem is if you override the splitter to include dashes in words and you have words (such as part numbers) that you want to index like "A-1234".

**To Reproduce**
```
# Assumes Ubuntu 16.04

# Install Sphinx
sudo apt install python3-pip
sudo pip3 install sphinx

# Download the test file attached to this issue and unzip it
unzip splitter_test.zip

# Build the docs
cd splitter_test
make html

# Verify that "a-1234" is in the search index. The below command should print some
# JavaScript code that includes the string "a-1234".
grep "a-1234" _build/html/searchindex.js

# View the docs
firefox _build/html/index.html &

# Now search for the string "a-1234"
```

**Expected behavior**
The test page comes up in the search results.

**Actual behavior**
The search results are empty.

**Environment info**
- OS: Ubuntu 16.04
- Python version: 3.5.2
- Sphinx version: 2.4.4

**Additional context**
Test project: [splitter_test.zip](https://github.com/sphinx-doc/sphinx/files/4314861/splitter_test.zip)

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #7293

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -86,6 +86,8 @@ Features added
 * #7103: linkcheck: writes all links to ``output.json``
 * #7025: html search: full text search can be disabled for individual document
   using ``:nosearch:`` file-wide metadata
+* #7293: html search: Allow to override JavaScript splitter via
+  ``SearchLanguage.js_splitter_code``
 
 Bugs fixed
 ----------
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -43,6 +43,14 @@ class SearchLanguage:
        This is a set of stop words of the target language.  Default `stopwords`
        is empty.  This word is used for building index and embedded in JS.
 
+    .. attribute:: js_splitter_code
+
+       Return splitter funcion of JavaScript version.  The function should be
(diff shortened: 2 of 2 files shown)


tst-lsavoie:
Thanks for the quick turn-around. I'll test this as soon as possible, hopefully later today.

tst-lsavoie:
I have verified that this fix works for my use case. Thanks for the help.

What version of Sphinx will this fix be in?

tk0miya:
Thank you for confirmation. I'll merge this after testing again.
This will be released as v3.0.0 maybe within a month.

PR #7319, opened 2020-03-15
C domain rewrite

### LINKED ISSUE #1539: Structure members break, when accompanied with brackets
Example:

```
#!rst

.. member:: CK_UTF8CHAR model[16]
```

Gives me both the type and the name bolded and hyperlink not working.

Practically we are forced to use this cases like that:

```
#!rst

.. member:: model
```

 Because otherwise it would be incorrect (use it without array and with the same type).

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1539
- Originally reported by: Vsevolod Glumov
- Originally created at: 2014-08-19T12:22:51.205

### LINKED ISSUE #2377: Using regular expressions for c function prototypes
The regular expressions [here](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/domains/c.py#L30), are not correct for C programming language. They won't parse some POSIX-compatible functions, like the following:

`// valid function prototype`
`void (*signal(int sig, void (*func)(int)))(int);`

In general, using regular expressions to parse programming languages is not good idea. For this purpose using something like parsec will be more appropriate.

Thanks, Paul.

jakobandersen:
Subject: essentially a rewrite of the C domain to fix issues and enable further development

### Feature or Bugfix
- Feature
- Bugfix
- Refactoring

### Purpose
The original implementation of the C domain relied on regex matching to parse signatures. While it works for many cases, there is a limit. This rewrite is in the same style as the C++ domain, and some details are shared between the two domains. It is naturally a large breaking change, but I believe it is necessary for furture improvement of the domain. In order to make some simple kind of judgement of the changes I have updated the cpython documentation as best as possible: https://github.com/jakobandersen/cpython/tree/doc_fixes

- xref lookup now respects the scope that it was declared in.
- Anonymous entities can now be declared with appropriate lookup rules.
- The ``c:type`` directive seemed to have been a catch-all for many signatures which made it difficult to parse and generate appropriate output. It has been changed to be much stricter and only allow either a name or a typedef-like declaration.
- There are new directives for structs, unions, enums, and enumerators. Lookup rules for enumerators are as expected from the language.
- New xref roles has been added in conjunction with the directives. (The issue from #7243 is not yet fixed yet)
- An ``c:expr`` role has been added to typeset a type of an expression in text. In particular while one before could write ``... a :c:type:`int` and then ...`` or ``... a :c:type:`char*` and then ...`` this will now trigger an error as the ``c:type`` strictly is for cross-referencing. Instead one should use the ``c:expr`` role.
- This PR incidentally reverts PR #7269 (sorry), except the removal of the ``signode['names'].append(targetname)`` thing. However, the ID generation is now guaranteed to produce different IDs for different declarations, and they are produced from parsed input (i.e., no arbitrary code injection).

### Detail
- Fixes #1539
- Fixes #2377

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -1215,8 +1215,14 @@ class ASTDeclaration(ASTBase):
                            env: "BuildEnvironment", options: Dict) -> None:
         verify_description_mode(mode)
         assert self.symbol
-
-        mainDeclNode = signode
+        # The caller of the domain added a desc_signature node.
+        # Always enable multiline:
+        signode['is_multiline'] = True
+        # Put each line in a desc_signature_line node.
+        mainDeclNode = addnodes.desc_signature_line()
+        mainDeclNode.sphinx_c_tagname = 'declarator'
+        mainDeclNode['add_permalink'] = not self.symbol.isRedeclaration
+        signode += mainDeclNode
 
         if self.objectType == 'member':
             pass
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -19,7 +19,7 @@ from docutils.nodes import Element, Node, TextElement, system_message
 from docutils.parsers.rst.states import Inliner
 
 from sphinx import addnodes
-from sphinx.addnodes import pending_xref, desc_signature
+from sphinx.addnodes import pending_xref
 from sphinx.application import Sphinx
 from sphinx.builders import Builder
 from sphinx.directives import ObjectDescription
@@ -29,7 +29,7 @@ from sphinx.locale import _, __
 from sphinx.roles import XRefRole
 from sphinx.util import logging
 from sphinx.util.cfamily import (
-    NoOldIdError, ASTBase, verify_description_mode, StringifyTransform,
+    NoOldIdError, ASTBaseBase, verify_description_mode, StringifyTransform,
     BaseParser, DefinitionError, UnsupportedMultiCharacterCharLiteral,
     identifier_re, anon_identifier_re, integer_literal_re, octal_literal_re,
     hex_literal_re, binary_literal_re, float_literal_re,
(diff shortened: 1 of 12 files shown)


sphinx/domains/c.py
```diff
@@ -73,158 +2859,162 @@ class CObject(ObjectDescription):
...
-    # These C types aren't described anywhere, so don't try to create
-    # a cross-reference to them
-    stopwords = {
-        'const', 'void', 'char', 'wchar_t', 'int', 'short',
-        'long', 'float', 'double', 'unsigned', 'signed', 'FILE',
-        'clock_t', 'time_t', 'ptrdiff_t', 'size_t', 'ssize_t',
-        'struct', '_Bool',
-    }
+    def warn(self, msg: Union[str, Exception]) -> None:
+        self.state_machine.reporter.warning(msg, line=self.lineno)
```
tk0miya, line 2955: Now we've used the `logger` library. It's simple and portable. So `warn()` methods are no longer needed. Please consider the replacement of them in the future.

sphinx/domains/c.py
```diff
@@ -26,36 +28,2820 @@
...
+        return self.data
+
+    def describe_signature(self, signode: desc_signature, mode: str,
+                           env: "BuildEnvironment", symbol: "Symbol") -> None:
+        txt = str(self)
+        signode.append(nodes.Text(txt, txt))
+
+
+class ASTParenExpr(ASTBase):
+    def __init__(self, expr):
```
tk0miya: Please annotate variables when you have time. It would be helpful to me for understanding C & C++ domain. Of course, it's not in a hurry. Please consider after merging.

tk0miya (approved):
Unfortuantely, I don't have time to understand the new implementation of the C domain. But it would go well because you're author :-) +1 for merging.

sphinx/domains/c.py
jakobandersen: Ah, I didn't notice that change. I'll get it fixed in the near'ish future.

sphinx/domains/c.py
jakobandersen: Good point. It certainly also helps me keeping things straight, so I'll fix it soon.

jakobandersen: Almost all should now be annotated.

tk0miya: Thanks!

PR #7323, opened 2020-03-16
Fix sphinx-autogen commandline in examples

mitya57:
Subject: Fix errors in an example in documentation

### Feature or Bugfix
- Bugfix

### Purpose
- The command name is `sphinx-autogen`, not `sphinx-autodoc`.
- The examples uses `docs` directory, not `doc`.


tk0miya:
Thank you always!

PR #7355, opened 2020-03-22
Fix #7331: autodoc: a cython-function are not recognized as a function

### LINKED ISSUE #7331: autodoc does not recognize Cython signature
Signatures generated by Cython with `binding=True` are not recognized by autodoc, although help can see it just right.

### To Reproduce
First create `cyspam.pyx` and `pyspam.py` with the same content as follows:
```python
def foo(bar: int) -> bool: """Docstring."""
```

Then write a minimal `setup.py` for it:
```python
from distutils.core import setup
from Cython.Build import cythonize

setup(name='autospam', py_modules=['pyspam'],
      ext_modules=cythonize('cyspam.pyx',
                            compiler_directives={'binding': True}))
```

After `pip install .` one should be able to get help from the interpreter, e.g.
```python
>>> import cyspam, pyspam
>>> help(cyspam.foo)
...
foo(bar: 'int') -> 'bool'
    Docstring.
>>> help(pyspam.foo)
...
foo(bar: int) -> bool
    Docstring.
```

Note that with `from __future__ import annotations` (PEP 563), the signature of `pyspam.foo` would be the same as the other.

In Sphinx ReST source, add this:
```rst
cyspam
------

.. automodule:: cyspam
   :members:

pyspam
------

.. automodule:: pyspam
   :members:
```

Then the generated HTML would be something like

> ### cyspam
> cyspam.foo
>
>     Docstring.
>
> ### pyspam
> pyspam.foo(bar: int) → bool
>
>     Docstring.

### Expected behavior
Signature generated for `cyspam.foo` should have been `foo(bar: 'int') -> 'bool'`.

### Sample project
[autospam.zip](https://github.com/sphinx-doc/sphinx/files/4346953/autospam.zip)

### Environment info
- OS: Debian GNU/Linux testing
- Python version: 3.7.6
- Sphinx version: 2.4.3
- Extra tools: Cython

### Additional context
This is originally filed against cython/cython#3438, however, Cython is not aware of the mechanism autodoc uses to fetch the signature to apply an appropriate patch.

tk0miya:
### Feature or Bugfix
- Bug fix

### Purpose
- refs: #7331


sphinx/ext/autodoc/__init__.py
```diff
@@ -1011,9 +1011,6 @@ def format_args(self, **kwargs: Any) -> str:
         if self.env.config.autodoc_typehints in ('none', 'description'):
             kwargs.setdefault('show_annotation', False)
 
-        if inspect.isbuiltin(self.object) or inspect.ismethoddescriptor(self.object):
-            # cannot introspect arguments of a C function or method
-            return None
```
tk0miya, line 1016: It should be better for using the same way as method does.

tests/roots/test-ext-autodoc/target/cython.pyx
```diff
@@ -0,0 +1,12 @@
+# cython: binding=True
+
+def foo(*args, **kwargs):
+    """Docstring."""
```
McSinyx, line 4: Coincidentally, this is the only one passed correctly for functions.  All pure Python functions defined in Cython has the same signature, e.g. https://mcsinyx.github.io/palace/html/reference.html#palace.sample_size

The definition of those are here: https://github.com/McSinyx/palace/blob/master/src/palace.pyx#L175

tk0miya: Thank you for the info. I found my implementation does not work well.

PR #7361, opened 2020-03-22
Fix typo in the french translation

ewjoachim:
Subject: Fix typo in the french translation

### Feature or Bugfix
- Bugfix

### Purpose
- I'm removing an extraneous space in the french translation.


tk0miya:
Thank you for your contribution. But we have translated messages in Transfix. So PR is not good way to fix it. I'll apply this change on transifex later.

ewjoachim:
No problem. The [contributing guide](https://github.com/sphinx-doc/sphinx/blob/3.x/CONTRIBUTING.rst#locale-updates) mentions transifex, but

1. Does not mention that translations must come from there
2. Does not say what the protocol is for fixing an erroneous translation (or help the translation effort)

Issue #7138, opened 2020-02-11
sphinx.ext.autodoc.typehints throws ValueError on pyqtSignal/Signal (PyQt5/PySide2)


samh:
**Describe the bug**
Code using `pyqtSignal` from PyQt5 causes Sphinx to throw `ValueError: no signature found for builtin <unbound signal activated>`. For PySide2 (`Signal`), it throws `ValueError: callable <PySide2.QtCore.Signal object at 0x7fa76b0c4270> is not supported by signature`.

**To Reproduce**
```
$ git clone https://github.com/samh/sphinx-typehints-bug
$ cd sphinx-typehints-bug
$ pip install -r requirements.txt
$ make html
```

**Expected behavior**
Expected no error.

**Your project**
Sample project: https://github.com/samh/sphinx-typehints-bug

In my real project I am using PyQt5, but a similar error appears to happen with PySide2, and that was easier to set up for a sample.

**Screenshots**

Output using PySide2 5.14.1:

```
(.venv) [samh@devdmsa sphinx-pyqt5-bug]$ make html
Running Sphinx v2.4.1
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: [extensions changed ('sphinx.ext.autodoc.typehints')] 1 added, 0 changed, 0 removed
reading sources... [100%] index                                                                                                                       
Exception occurred:
  File "/opt/python36/lib/python3.6/inspect.py", line 2381, in _signature_from_callable
    raise ValueError('callable {!r} is not supported by signature'.format(obj))
ValueError: callable <PySide2.QtCore.Signal object at 0x7fa76b0c4270> is not supported by signature
The full traceback has been saved in /tmp/sphinx-err-q6sdtg7i.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [html] Error 2
```

Output using PyQt 5.4.2:

```
cd docs && make html
<removed>
make[1]: Entering directory `<removed>'
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v2.4.1
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 1 changed, 0 removed
[...]

Issue #3610, opened 2017-04-06
Support for @overload in Python


orenbenkiki:
Sphinx ignores the type signature of methods which use @overload in Python.
Granted, support for @overload in normal classes is very new (as of mypy version 0.501, March 2017).
At any rate:

### Problem

The automatically-generated documentation lists just the final concrete signature, which has no types, instead of listing the typed signatures.

#### Procedure to reproduce the problem
```
class Foo:

    @overload
    def bar(x: int) -> int:
        pass

    @overload
    def bar(x: str) -> str:
        pass

    def bar(x):
        """Documentation..."""
        return x
```

#### Results

The documentation generated by `sphinx.ext.autodoc` shows a single signature `bar(x)`.

#### Expected results

The documentation generated by `sphinx.ext.autodoc` should include two signatures, `bar(x: int) -> int` and `bar(x: str) -> str`. Ideally, the function documentation should look like:

- The type-less signature (this would serve as the anchor for links to the function or method)
- A sub-list of the typed signatures
- The content of the docstring of the type-less (real) body

eric-wieser:
As a proof of concept, here's how to extract the overloads by patching `typing.overload`:

```python
import inspect
import typing

SPHINX_IS_RUNNING = True

if SPHINX_IS_RUNNING:
    _overload_data = {}

    old_overload = typing.overload

    def new_overload(f):
        frame = inspect.currentframe()
        key = frame.f_back, f.__name__
        if key not in _overload_data:
            _overload_data[key] = []
        _overload_data[key].append(f)
        return old_overload(f)
    typing.overload = new_overload


    def get_overloads():
        return {
            frame.f_locals[name]: overloads
            for (frame, name), overloads in _overload_data.items()
        }

class F:
    @typing.overload
    def foo(x: str):
        pass

    @typing.overload
    def foo(x: int):
        pass

    def foo(x):
        """ overloaded"""
        pass


if SPHINX_IS_RUNNING:
    for func, overloads in get_overloads().items():
        prefix = '.. function:: '
        for i, o in enumerate(overloads):
            print((prefix if i == 0 else " "*prefix) + func.__qualname__ + str(inspect.signature(o)))
        print("    " + func.__doc__)
```

Actually supporting rST markup for the overloaded method would require gh-2106

eric-wieser:
[...]

PR #7384, opened 2020-03-26
autodoc: Respect type annotations in `__new__`, `metacls.__call__`, and builtin types

eric-wieser:
### Feature or Bugfix

- Bugfix

### Purpose
None of the following work correctly before this patch
* Signatures defined by `__new__`
* Signatures defined by metaclasses
* Signatures defined by builtin base classes

Patch includes a test showing that these now all work.

It's possible other tests will have been affected, will update if needed.

Ping @tk0miya, since you recently looked at annotation stuff.


eric-wieser:
Failures are all of the form
```
E         -  '.. py:class:: DocstringSig()',
E         ?                             --
E         +  '.. py:class:: DocstringSig',
```

I think this change is correct - with it, these two classes are now documented identically:
```
class WithInit:
    def __init__(self):
        pass
class WithoutInit:
    pass
```
Whereas previously the former would have `()` but the latter would not.

The two classes give exactly the same result from `inspect.signature`, so I would consider this a bug.
I've added a second commit to fix this.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1175,22 +1175,73 @@ def format_args(self, **kwargs: Any) -> str:
...
+            call = get_user_defined(type(self.object), '__call__')
+
+            if call is not None:
+                # types which have confusing metaclass signatures it would be best
+                # not to show
+                call_blacklist = [
+                    'enum.EnumMeta.__call__',
+                ]
+                if "{0.__module__}.{0.__qualname__}".format(call) in call_blacklist:
+                    call = None
```
eric-wieser: This is a little unfortunate, because for an enum
```python
class MyEnum(enum.Enum):
    VAL = 1
```
the signature is actually:
```python
>>> inspect.signature(MyEnum)
<Signature (value, names=None, *, module=None, qualname=None, type=None, start=1)>
```
This signature isn't actually a lie, it's just something no sane person would ever call:
```python
>>> MyEnum("MyTotallyNewEnum", names="why did i do this")
<enum MyTotallyNewEnum>
```

sphinx/ext/autodoc/__init__.py
```diff
@@ -1175,22 +1175,73 @@ def format_args(self, **kwargs: Any) -> str:
...
-        # classes without __init__ method, default __init__ or
-        # __init__ written in C?
-        if initmeth is None or \
-                inspect.is_builtin_class_method(self.object, '__init__') or \
-                not(inspect.ismethod(initmeth) or inspect.isfunction(initmeth)):
-            return None
-        try:
-            self.env.app.emit('autodoc-before-process-signature', initmeth, True)
-            sig = inspect.signature(initmeth, bound_method=True)
+        def get_user_defined(obj, attr):
```
tk0miya: Could you add annotations please? And it would be better to add docstring if possible.

tk0miya: nits: `get_user_defined_function()` or `get_user_defined_function_or_method()` is better to me. What do you think?

sphinx/ext/autodoc/__init__.py
```diff
@@ -1175,22 +1175,73 @@ def format_args(self, **kwargs: Any) -> str:
...
+            if inspect.is_builtin_class_method(obj, attr):
+                return None
+            attr = self.get_attr(obj, attr, None)
+            if not (inspect.ismethod(attr) or inspect.isfunction(attr)):
```
tk0miya, line 1262: FWIW, some kind of cython functions are not treated as "function" (refs: #7331). I'm not good at cython, but it might be needed to check `inspect.isroutine()`.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1175,22 +1175,73 @@ def format_args(self, **kwargs: Any) -> str:
...
+        # in its metaclass
+        if sig is None:
+            call = get_user_defined(type(self.object), '__call__')
+
+            if call is not None:
+                # types which have confusing metaclass signatures it would be best
+                # not to show
+                call_blacklist = [
+                    'enum.EnumMeta.__call__',
+                ]
```
tk0miya: I think such constants should be placed at the module level and be capitalized.

sphinx/ext/autodoc/__init__.py
tk0miya: Reasonable.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1175,22 +1175,73 @@ def format_args(self, **kwargs: Any) -> str:
...
+                if "{0.__module__}.{0.__qualname__}".format(call) in call_blacklist:
+                    call = None
+
+            if call is not None:
+                self.env.app.emit('autodoc-before-process-signature', call, True)
+                try:
+                    sig = inspect.signature(call, bound_method=True)
+                except TypeError:
+                    print("Faild, oh no")
```
tk0miya: This must be updated. Should this be replaced with a warning? Or be suppressed?

sphinx/ext/autodoc/__init__.py
```diff
@@ -1175,22 +1175,73 @@ def format_args(self, **kwargs: Any) -> str:
...
+            init = get_user_defined(self.object, '__init__')
+            if init is not None:
+                self.env.app.emit('autodoc-before-process-signature', init, True)
+                try:
+                    sig = inspect.signature(init, bound_method=True)
+                except TypeError:
+                    print("Faild, oh no")
+
+        # if none of the attributes are user-defined, we can skip
+        # `autodoc-before-process-signature`, and just let inspect handle it
```
tk0miya: Why skipped here?

tests/test_ext_autodoc_configs.py
```diff
@@ -508,8 +521,9 @@ def test_autodoc_typehints_signature(app):
         '',
         '.. py:function:: missing_attr(c, a: str, b: Optional[str] = None) -> str',
         '   :module: target.typehints',
-        ''
+        '',
     ]
+    assert list(actual) == expected
```
tk0miya: +0 for changing style. But all other cases should be updated also. Anyway, that should be another PR.

tests/test_autodoc.py
```diff
@@ -510,14 +545,14 @@ def test_autodoc_members(app):
     # default (no-members)
     actual = do_autodoc(app, 'class', 'target.inheritance.Base')
     assert list(filter(lambda l: '::' in l, actual)) == [
-        '.. py:class:: Base',
+        '.. py:class:: Base()',
```
tk0miya: Note: `py:class` directive skips empty parenthesis. So this change harms nobody.

tk0miya:
Should we also modify `get_doc()` to obtain docstring in order of this sequence?

sphinx/ext/autodoc/__init__.py
eric-wieser: This is the same logic as what was there before, so I'd like to declare this out of scope.

sphinx/ext/autodoc/__init__.py
eric-wieser: Whoops, this was a debug statement!

sphinx/ext/autodoc/__init__.py
eric-wieser: Because if the type is native, there's no source code to process, and the assumption is `autodoc-before-process-signature` has nothing to do.

tests/test_ext_autodoc_configs.py
eric-wieser: This was leftover from a previous change which needed the separate variable. Will revert.

tests/test_autodoc.py
eric-wieser: Excellent news.

eric-wieser:
> Should we also modify get_doc() to obtain docstring in order of this sequence?

We could, but I don't think that matters so much, and would prefer not to do it in this PR.

sphinx/ext/autodoc/__init__.py
eric-wieser: Be suppressed, because that's what this code did before.

sphinx/ext/autodoc/__init__.py
eric-wieser: Updated to use `get_user_defined_function_or_method`, no opinion from me on the name.
Annotation and docstring added.

sphinx/ext/autodoc/__init__.py
eric-wieser: Updated the comment a little.

eric-wieser:
Comments addressed I think

tests/test_util_inspect.py
```diff
@@ -29,12 +29,14 @@ def test_signature():
...
 
-    # builitin classes
-    with pytest.raises(TypeError):
-        inspect.signature(int)
-
-    with pytest.raises(TypeError):
-        inspect.signature(str)
+    # builtins are supported on a case-by-case basis, depending on whether
+    # they define __text_signature__
+    if getattr(list, '__text_signature__', None):
```
tk0miya, line 34: I think this if-condition will be removed in the future, right?
It seems python3.7+ supports inspection for the list object. How about change this to `sys.version_info > (3, 7)`?

sphinx/ext/autodoc/__init__.py
tk0miya: I think this event is public one. So it should be emitted even if no source code.

tk0miya (changes requested):
Sorry for late. Let's go forward this.

tests/test_util_inspect.py
eric-wieser: Could do, although I wonder if running python with `-OO` affects whether whether these attributes are populated.

At any rate, the presence of documentation on builtins sort of feels like an implementation detail. Switching to a version check is likely to break pypy or something.

sphinx/ext/autodoc/__init__.py
eric-wieser: The only way to ensure it is emitted would be to either:
* Emit it for the object itself on this line, rather than attempting to find the right `__new__` / `__init__` etc
* Duplicating even more of the logic inside `inspect.signature`

sphinx/ext/autodoc/__init__.py
```diff
@@ -1175,22 +1183,70 @@ def format_args(self, **kwargs: Any) -> str:
...
+                try:
+                    sig = inspect.signature(init, bound_method=True)
+                except TypeError:
+                    pass
+
+        # if none of the attributes are user-defined, then it's unlikely that
+        # `autodoc-before-process-signature` will be able to do anything.
+        # Let's just let inspect handle it to ensure we at least get a signature.
+        if sig is None:
+            try:
```
eric-wieser, line 1297:
```suggestion
            # we don't know the exact method that inspect.signature will read the signature from,
            # so just pass our hook the object itself.
            self.env.app.emit('autodoc-before-process-signature', obj, False)
            try:
```

eric-wieser: @tk0miya: is this an acceptable compromise?

eric-wieser: @tk0miya, ping

tk0miya: I don't understand what the comment says. And `inspect.signature()` does not refer the docstring, AFAIK. Anyway, we always pass the object itself even if we don't know the exact location.

eric-wieser: You're right, that wording was bad. Updated.

tk0miya: Thanks, Good to me!

tk0miya (approved):
LGTM

eric-wieser:
Needs a rebase to pass tests it seems

eric-wieser:
Rebased, hopefully CI will give me some hints this time.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1198,22 +1206,72 @@ def format_args(self, **kwargs: Any) -> str:
...
+
+        # None of the attributes are user-defined, so fall back to let inspect
+        # handle it.
+        if sig is None:
+            # We don't know the exact method that inspect.signature will read
+            # the signature from, so just pass the object itself to our hook.
+            self.env.app.emit('autodoc-before-process-signature', self.object, False)
+            try:
+                sig = inspect.signature(self.object, bound_method=False)
+            except TypeError:
```
tk0miya: It seems `inspect.signature()` raises `ValueError` also. Please handle it too.
```
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 395, in format_signature
    args = self.format_args(**kwargs)
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 1266, in format_args
    sig = inspect.signature(self.object, bound_method=False)
  File "/Users/tkomiya/work/sphinx/sphinx/util/inspect.py", line 406, in signature
    signature = inspect.signature(subject)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 3093, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2842, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2372, in _signature_from_callable
    raise ValueError(
ValueError: no signature found for builtin type <class 'sphinx.errors.SphinxError'>
```

tk0miya: A testcase for exception class is also needed:
```
$ python
Python 3.8.2 (default, Mar  2 2020, 00:44:41)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
>>> import inspect
>>> inspect.signature(Exception)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 3093, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2842, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2372, in _signature_from_callable
    raise ValueError(
ValueError: no signature found for builtin type <class 'Exception'>
```

eric-wieser: Test added.

eric-wieser:
CI looks like it will pass now.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1188,22 +1196,72 @@ def format_args(self, **kwargs: Any) -> str:
...
+            if inspect.is_builtin_class_method(obj, attr):
+                return None
+            attr = self.get_attr(obj, attr, None)
+            if not (inspect.ismethod(attr) or inspect.isfunction(attr)):
+                return None
+            return attr
+
+        sig = None
+
+        # this sequence is copied from inspect._signature_from_callable
```
tk0miya: I noticed this comment now. Do we need to care for its license?

sphinx/ext/autodoc/__init__.py
```diff
@@ -1188,22 +1196,72 @@ def format_args(self, **kwargs: Any) -> str:
...
+
+            if call is not None:
+                if "{0.__module__}.{0.__qualname__}".format(call) in _METACLASS_CALL_BLACKLIST:
+                    call = None
+
+            if call is not None:
+                self.env.app.emit('autodoc-before-process-signature', call, True)
+                try:
+                    sig = inspect.signature(call, bound_method=True)
+                except ValueError:
```
tk0miya: I think we need to handle both `ValueError` and `TypeError`.

tk0miya (commented):
LGTM with nits.

sphinx/ext/autodoc/__init__.py
eric-wieser: Can you come up with an example that raises `TypeError`?

sphinx/ext/autodoc/__init__.py
eric-wieser: The sequence of attributes to look at is copied, but not the code itself. I don't think the license is relevant.

sphinx/ext/autodoc/__init__.py
tk0miya: I got an error with this class.
```
class Foo:
    __signature__ = 1
```
```
WARNING: error while formatting arguments for example.Foo:
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 395, in format_signature
    args = self.format_args(**kwargs)
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 1255, in format_args
    sig = inspect.signature(self.object, bound_method=False)
  File "/Users/tkomiya/work/sphinx/sphinx/util/inspect.py", line 406, in signature
    signature = inspect.signature(subject)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 3093, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2842, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2252, in _signature_from_callable
    raise TypeError(
TypeError: unexpected object 1 in __signature__ attribute

During handling of the above exception, another exception occurred:

  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 398, in format_signature
    args = self.format_args()
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 1255, in format_args
    sig = inspect.signature(self.object, bound_method=False)
  File "/Users/tkomiya/work/sphinx/sphinx/util/inspect.py", line 406, in signature
    signature = inspect.signature(subject)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 3093, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2842, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2252, in _signature_from_callable
    raise TypeError(
TypeError: unexpected object 1 in __signature__ attribute
```

I know this is crazy example. But I received such report sometimes. So it would be better to handle `ValueErorr` also.

sphinx/ext/autodoc/__init__.py
tk0miya: I understand. np.

sphinx/ext/autodoc/__init__.py
tk0miya: I found related problem in autofunction: https://github.com/sphinx-doc/sphinx/issues/7629

tk0miya: I found a real bug report: https://github.com/sphinx-doc/sphinx/issues/7138

eric-wieser: That real bug report is `ValueError`, which is already handled.

I think I should try to make that traceback half as long though, pushing an update now.

tk0miya: Oh, sorry. The last one is my wrong.

sphinx/ext/autodoc/__init__.py
```diff
@@ -391,11 +391,15 @@ def format_signature(self, **kwargs: Any) -> str:
...
-                    args = self.format_args()
+                if kwargs:
+                    try:
+                        args = self.format_args(**kwargs)
+                    except TypeError:
+                        # avoid chaining tracebacks, by putting nothing here
+                        pass
+
+                # retry without arguments for old documenters
+                args = self.format_args()
```
tk0miya: It seems `format_args()` is called twice.

eric-wieser: It was called twice before too - this way, we at least:
* Do not call it twice with identical arguments
* Do not chain the exception of the first call with the exception of the second

Does anything actually pass non-empty `kwargs`?

tk0miya: >It was called twice before too

It's an unexpected result. The `TypeError` handler is added only for old Documener classes which does not accept keyword arguments. It implicitly expects `format_args()` method does not raise `TypeError`.

In this case, I think no reason to raise `TypeError` on `format_args()`. It should be handled inside it.

>Does anything actually pass non-empty kwargs?

Yes. The autosummary extension gives keyword argument.
https://github.com/sphinx-doc/sphinx/blob/62711bc081744cecba2af799ad8dbf678f6a97a4/sphinx/ext/autosummary/__init__.py#L345



Note: It's after 2 AM in Japan. So this is my last comment today. Good night!

eric-wieser: I take it back, you're correct. Updated in my latest push.

eric-wieser:
@tk0miya: Think you'll have time to look at this again before 3.1.0?

tk0miya:
Of course, Yes. But my request is not changed. Please catch a `TypeError` in `format_args()`. Then I'll merge this soon.

eric-wieser:
> Please catch a TypeError in format_args(). Then I'll merge this soon.

Would you mind adding a comment on the relevant lines? I've moved things around a bit now.

sphinx/ext/autodoc/__init__.py
tk0miya: >>Please catch a TypeError in format_args(). Then I'll merge this soon.
>
>Would you mind adding a comment on the relevant lines? I've moved things around a bit now.

Sorry. As commented in this thread, `inspect.signature()` raises `TypeError`. And I think `format_args()` should not pass through it. So please catch it like `FunctionDocumenter.format_args()`.

eric-wieser: Thanks for the reminder, updated.

tk0miya: Thank you for update. I'll review this tomorrow (or in this weekend) with cleared eyes.

tk0miya (approved):
LGTM! Thank you for your patience and cooperation!

eric-wieser:
I assume now #7613 and your todo comment from #3610 will be marginally easier.

tk0miya:
Yes, I can improve #7716 now. I'm waiting for merging this :-)

PR #7387, opened 2020-03-27
Add ObjectDescription.transform_content()

jakobandersen:
I unfortunately forgot about this change before the closure for beta, but it would be really nice to get it merged for 3.0 if possible.

### Feature or Bugfix
- Feature

### Purpose
The ``before_content()`` and ``after_content()`` methods are used to set scoping information in the domains. However, there is no way (at least not an easy one) to extend existing object descriptions for injecting extra content. Currently this gives a lot of problems with wrong scoping in Breathe. This change will make it very easy to inject additional content as if it was generated by the rst parser.

### Detail
Another solution would have been to add ``self`` as argument when the ``object-description-transform`` event is emitted, so all data is available there. Though, that would be breaking change, and I think this has better usability.

### DIFF
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -126,6 +126,15 @@ class ObjectDescription(SphinxDirective):
         """
         pass
 
+    def transform_content(self, contentnode: addnodes.desc_content) -> None:
+        """
+        Called after creating the content through nested parsing,
+        but before the ``object-description-transform`` event is emitted,
+        and before the info-fields are transformed.
+        Can be used to manipulate the content.
+        """
+        pass
+
     def after_content(self) -> None:
         """
         Called after parsing content. Used to reset information about the
@@ -198,6 +207,7 @@ class ObjectDescription(SphinxDirective):
             self.env.temp_data['object'] = self.names[0]
         self.before_content()
         self.state.nested_parse(self.content, self.content_offset, contentnode)
+        self.transform_content(contentnode)
         self.env.app.emit('object-description-transform',
                           self.domain, self.objtype, contentnode)
         DocFieldTransformer(self).transform_all(contentnode)
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,8 @@ Deprecated
 Features added
 --------------
 
+* Added ``ObjectDescription.transform_content()``.
+
 Bugs fixed
 ----------
 


tk0miya:
I don't know that modifying the content is really a part of the responsibility of object-description directives.  Please let me know how you'd like to use the method for breathe's case.

With intuition, they are different components. So I prefer to the "another solution". But I still don't understand why you need to access `self` object.

Note: I don't think this will not break a release. So +0 for adding it. But we need to discuss the design of interface (or components) more. So I can't promise you this will be added into 3.0.0.

jakobandersen:
The only reason I know of is for making Breathe work better, so from that perspective the "another solution" is more elegant. The reason for passing ``self`` as well is to have a handle to which directive was run. If you are ok with that change, I'll update the PR accordingly.

Breathe dynamically creates object descriptions from Doxygen XML and runs them to creates signatures. It then appends content from the XML as well, but this content needs to be converted as if it was a nested rst parse, but is currently not (resulting all kinds of problems).

tk0miya:
>The reason for passing self as well is to have a handle to which directive was run.

I'd like to know why you need to know "which directive was run". I think you can already obtain it via `domain` and `objtype` parameters of `object-description-transform` event. I still don't understand why the directive object is needed for doctree-transformation.

jakobandersen:
Indeed, the event handler gets the directive domain and name, but not the actual *instance* that emitted the event. The goal is to smuggle per-instance associated data to the event handler, so additional content can be generated from it, in-between ``before_content()`` and ``after_content()``.

tk0miya:
I still don't understand why you need to get the instance of directive. Is there any important data? If so, I think the transform module is strongly coupled with the directive. So it would be better to add a new method.

jakobandersen:
> I still don't understand why you need to get the instance of directive. Is there any important data?

Not in Sphinx, but extensions like Breathe needs to get a per-instance handle to generate additional content. It instantiates object descriptions on the fly and needs thus needs to add content dynamically as well.
One way is to subclass the object descriptions and add extra attributes, and then implemented the propose ``after_nested_parse`` method. Modifying the event, and then passing the object description instance would give a similar capability (though being a breaking change).

tk0miya:
I thought the breathe's use case is strongly coupled with the directive. So I feel no reason to separate a transform component from the directive. Is my understanding correct? If so, it's reasonable to add `after_nested_parse()` instead of handling the event.

On the other hand, I have an idea to pass an instance of the directive to the `object-description-transform event to refactor `DocFieldTransformer` (not determined yet. Just an idea). So it might be better to add two ways.

sphinx/directives/__init__.py
```diff
@@ -198,6 +205,7 @@ def run(self) -> List[Node]:
             self.env.temp_data['object'] = self.names[0]
         self.before_content()
         self.state.nested_parse(self.content, self.content_offset, contentnode)
+        self.after_nested_parse(contentnode)
```
tk0miya: The method name should represent what the method doing. It only describes when this method called, but it does not describe the purpose of the method. How about `transform_content()` instead?

jakobandersen:
> I thought the breathe's use case is strongly coupled with the directive. So I feel no reason to separate a transform component from the directive. Is my understanding correct? If so, it's reasonable to add `after_nested_parse()` instead of handling the event.

It is indeed tied with the directive, so let's stick to adding a method then.

> On the other hand, I have an idea to pass an instance of the directive to the `object-description-transform event to refactor `DocFieldTransformer` (not determined yet. Just an idea). So it might be better to add two ways.

How about the following: 
1. Rename ``after_nested_parse`` to ``create_content``.
2. Make the default implementation of ``create_content`` be essentially everything between ``before_content`` and ``after_content``.
   ```python
   self.state.nested_parse(self.content, self.content_offset, contentnode)
   self.env.app.emit('object-description-transform',
                             self.domain, self.objtype, contentnode)
           DocFieldTransformer(self).transform_all(contentnode)
   ```
3. And then let ``run()`` have:
   ```python
   self.before_content()
   self.create_content()
   self.env.temp_data['object'] = None
   self.after_content()
   ```

This should be non-breaking, and subclasses can decide when they want to call ``super().create_content(contentnode)`` if they override the method.

jakobandersen:
(and ``self.env.temp_data['object'] = None`` could be moved into ``after_content`` as well.)

sphinx/directives/__init__.py
jakobandersen: (See my PR comment)

tk0miya:
I'm okay if you're saying both `create_content()` and `transform_content()`. It makes `run()` method simple. That's a good idea.
But if you meant that only adding former one, -1 for the change. It makes that subclasses have to call a super method explicitly. I guess developers who made a subclass want to hook a content-transformation handler, not a content-creation handler. So `create_content()` is not good to enhance to me.

jakobandersen:
I have renamed the method to ``transform_content``, but left out ``create_content`` for now as it would be a single line of code.
I suggest merging this for now to allow new developments in Breathe (and others), and then we can make a breaking refactoring in master to make everything cleaner.

tk0miya (approved):
LGTM!

jakobandersen:
Great, thanks!

PR #7411, opened 2020-04-04
LaTeX: Add `latex_theme_options` to override theme options

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add `latex_theme_options` to override theme options by user settings.
- There are still no overridable settings now. This is a preparation to add a new theme option.

### DIFF
--- a/sphinx/builders/latex/theming.py
+++ b/sphinx/builders/latex/theming.py
@@ -25,6 +25,7 @@ class Theme:
     """A set of LaTeX configurations."""
 
     LATEX_ELEMENTS_KEYS = ['papersize', 'pointsize']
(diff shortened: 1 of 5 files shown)


sphinx/builders/latex/theming.py
```diff
@@ -30,37 +34,34 @@ def __init__(self, name: str) -> None:
         self.wrapperclass = name
         self.toplevel_sectioning = 'chapter'
 
+    def update(self, options: Dict) -> None:
```
tk0miya: I noticed this conflicts with #7413 . So I'll update this PR later.

PR #7417, opened 2020-04-05
[WIP] Close #6740: LaTeX: Support tables without column rules

### LINKED ISSUE #6740: Removing LaTeX column borders for automatic colspec
I can specify `.. tabularcolumns:: JJ` before a two-column table to make both columns justified and autosizing, without a vertical line between the columns.

It would be nice if there was a way to do this for all tables, so that I do not need to write ``tabularcolumns`` manually every time.
I've looked into the templates used to generate the LaTeX code for tables and I don't think I can provide my own template to do this.
The template contains `\begin{tabulary}{\linewidth}[t]<%= table.get_colspec() -%>` and `table.get_colspec()` already [adds the pipe `|` characters](https://github.com/sphinx-doc/sphinx/blob/75477633b69b11858ae27c27f3af6e3ff2db928f/sphinx/writers/latex.py#L320) which define the vertical lines.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #6740, #6666 
- This contains #7411


CHANGES
```diff
@@ -13,6 +13,12 @@ Deprecated
 Features added
 --------------
 
+* LaTeX: Make the ``toplevel_sectioning`` setting optional in LaTeX theme
+* LaTeX: Allow to override papersize and pointsize from LaTeX themes
+* LaTeX: Add :confval:`latex_theme_options` to override theme options
+* #6740: LaTeX: Support tables without column rules via a theme option:
+  ``show_table_column_rules``
```
tk0miya, line 38: I'm still thinking about what is the best wording for a vertical line of the table. column rule? vertical rule (vrule)? vertical line (vline)? column separator (colsep)?

sphinx/builders/latex/theming.py
```diff
@@ -24,61 +24,89 @@
...
+    LATEX_ELEMENTS_KEYS = ['papersize', 'pointsize']
+    UPDATABLE_KEYS = ['papersize', 'pointsize', 'show_table_column_rules']
+
     def __init__(self, name: str) -> None:
         self.name = name
         self.docclass = name
         self.wrapperclass = name
+        self.papersize = 'letterpaper'
+        self.pointsize = '10pt'
+        self.show_table_column_rules = True
```
tk0miya, line 36: Now I use a boolean to show/hide column rules. But it might be better to choose a character for column separator instead.
```
table_column_rules = |
```
```
table_column_rules = ||
```
```
table_column_rules = :
# see https://tex.stackexchange.com/questions/122301/how-to-add-vertical-and-horizontal-dashed-lines-in-array-and-tabular-enviroments
```

tk0miya: In addition, we can use `:stub-columns:` option to make the table having vertical headings. So special column rule; header separator would be needed for such case...

sphinx/writers/latex.py
```diff
@@ -872,7 +877,10 @@ def visit_table(self, node: Element) -> None:
                 '%s:%s: deeply nested tables are not implemented.' %
                 (self.curfilestack[-1], node.line or ''))
 
-        self.tables.append(Table(node))
+        table_options = {
+            'column_rules': self.theme.show_table_column_rules,
```
tk0miya, line 881: Note: multicolumn should refer to this option also.

tk0miya: done

CHANGES
tk0miya: Note: The reference of booktab package calls it as "rule":

>In British typesetting, a ‘line’ is always called a ‘rule’. Perhaps confusingly (for
historic reasons in fact), the ‘thickness’ of rule is often referred to as is its ‘width’
(whereas just about everyone else would call this ‘depth’ or ‘height’, if they were
thinking of a horizontal rule). A ‘thick black line’ is called a ‘heavy rule’. I have
used this terminology in most of the new commands below. If nothing else it
avoids confusion with \hline.
https://www.ctan.org/pkg/booktabs

PR #7426, opened 2020-04-06
Escape some links in HTML templates

mgeier:
### Feature or Bugfix

- Bugfix

### Purpose

If a user chooses a strange `master_doc`, e.g.:

    master_doc = 'my"index'

... the links to the main page will be broken.

Same for `logo`, `favicon` and style sheets.

If a user chooses strange characters for sub-directories of the source directory, "previous" and "next" links might be broken, too.

### DIFF
--- a/sphinx/themes/agogo/layout.html
+++ b/sphinx/themes/agogo/layout.html
@@ -14,13 +14,13 @@
     <div class="header-wrapper" role="banner">
       <div class="header">
         {%- if logo %}
-          <p class="logo"><a href="{{ pathto(master_doc) }}">
+          <p class="logo"><a href="{{ pathto(master_doc)|e }}">
             <img class="logo" src="{{ pathto('_static/' + logo, 1) }}" alt="Logo"/>
           </a></p>
         {%- endif %}
         {%- block headertitle %}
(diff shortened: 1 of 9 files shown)


tk0miya (changes requested):
+0; I suppose such characters might cause other errors. So I'd not like to support them like this approach. But I don't know what kind of characters effect to our build. Therefore this is reasonable until we determine the prohibited characters for docnames.

BTW, we have a bug fix release 3.0.1 in nearly days. So it's nice if you rebase this into the 3.0.x branch.
Thanks,

mgeier:
Thanks for the review!

> I suppose such characters might cause other errors.

I tried it and there were no other errors.

I don't think any of this has any real consequences. I just noticed it while creating a theme and playing around with escaping and I thought I'd make a PR, just for completeness' sake.

> rebase this into the 3.0.x branch

Done.

tk0miya:
>I tried it and there were no other errors.

Sphinx supports many kind of formats (including 3rd party extensions). I worried these strange characters cause trouble in one of them.

tk0miya (approved):

tk0miya:
Thank you for rebasing. Merged.

PR #7443, opened 2020-04-08
Add globaltoc_collapse and globaltoc_includehidden options

mgeier:
I thought since the `toctree()` function has these options, it would make sense to have those as theme options as well, wouldn't it?

If this is considered for merging, I can add docs.

### DIFF
--- a/doc/usage/theming.rst
+++ b/doc/usage/theming.rst
@@ -155,6 +155,21 @@ These themes are:
     previous/next page using the keyboard's left and right arrows.  Defaults to
     ``False``.
 
+  .. versionadded:: 3.1
+
+    The following options:
+
+  - **globaltoc_collapse** (true or false): Only expand subsections
+    of the current document in ``globaltoc.html``
+    (see :confval:`html_sidebars`).
+    Defaults to ``True``.
+
+  - **globaltoc_includehidden** (true or false): Show even those
+    subsections in ``globaltoc.html`` (see :confval:`html_sidebars`)
+    which have been included with the ``:hidden:`` flag of the
+    :rst:dir:`toctree` directive.
+    Defaults to ``False``.
+
 **alabaster**
   `Alabaster theme`_ is a modified "Kr" Sphinx theme from @kennethreitz
   (especially as used in his Requests project), which was itself originally
--- a/doc/usage/theming.rst
+++ b/doc/usage/theming.rst
@@ -155,20 +155,20 @@ These themes are:
     previous/next page using the keyboard's left and right arrows.  Defaults to
     ``False``.
 
-  .. versionadded:: 3.1
-
-    The following options:
-
   - **globaltoc_collapse** (true or false): Only expand subsections
     of the current document in ``globaltoc.html``
     (see :confval:`html_sidebars`).
     Defaults to ``True``.
+
+    .. versionadded:: 3.1
 
   - **globaltoc_includehidden** (true or false): Show even those
     subsections in ``globaltoc.html`` (see :confval:`html_sidebars`)
     which have been included with the ``:hidden:`` flag of the
     :rst:dir:`toctree` directive.
     Defaults to ``False``.
+
+    .. versionadded:: 3.1
 
 **alabaster**
   `Alabaster theme`_ is a modified "Kr" Sphinx theme from @kennethreitz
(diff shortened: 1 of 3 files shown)


tk0miya:
+1 for adding `globaltoc_collapse`.
-0 for adding `globaltoc_includehidden`. I still don't understand why `includehidden` option is needed.

mgeier:
> I still don't understand why `includehidden` option is needed.

I have never used this myself, but some people seem to like to use `:hidden:` on their `toctree` to not show the subsection titles (the TOC) within the main text, but some (or most?) of those people would still like to see those subsections in the table of contents in the sidebar.

Some popular themes have this option (with different names), and they even change the default value to switch it on (which I didn't in this PR):

* https://github.com/bitprophet/alabaster/blob/3b68afcfe55a80508254b22904294100a160e6a7/alabaster/theme.conf#L37
* https://github.com/readthedocs/sphinx_rtd_theme/blob/35034930943c38e6f839ad89a4e3bcba811010d5/sphinx_rtd_theme/theme.conf#L12

And here is where I got the name for the option, which is IMHO much better:

* https://github.com/bashtage/sphinx-material/blob/4768f0cfca58968ef834d14e99866d989a452852/sphinx_material/sphinx_material/theme.conf#L49-L50

And here are some more:

* https://bitbucket.org/vinay.sajip/sphinx_sizzle_theme/src/9e79c78daa1635f67a6f8849a3e6aca5e8fd6c11/sphinx_sizzle_theme/theme.conf#lines-10
* https://github.com/guzzle/guzzle_sphinx_theme/blob/eefd45b79383b1b4aab1607444e41366fd1348a6/guzzle_sphinx_theme/guzzle_sphinx_theme/theme.conf#L40
* https://github.com/ryan-roemer/sphinx-bootstrap-theme/blob/a6dfc6f9054f6b4cf3eb1acadf715a679ed53a7b/sphinx_bootstrap_theme/bootstrap/theme.conf#L36-L43

tk0miya:
>This will still notify Sphinx of the document hierarchy, but not insert links into the document at the location of the directive – this makes sense if you intend to insert these links yourself, in a different style, or in the HTML sidebar.
This will still notify Sphinx of the document hierarchy, but not insert links into the document at the location of the directive – this makes sense if you intend to insert these links yourself, in a different style, or in the HTML sidebar.
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree

Hmm... our document has already mentioned about usage in the sidebar.  Okay, let's go as is.

mgeier:
I've re-based this PR and added a commit with documentation: c17f00c7a26975d09f1c3d3617b36e99aa0a2aeb

While doing that, I saw that there are also the parameters `maxdepth` and `titles_only`.

Shall I add configuration options for those, too?

For comparison: the RTD theme has both options, Alabaster has none. The other themes mentioned above have the "depth" option but not "titles_only".

doc/usage/theming.rst
```diff
@@ -155,6 +155,21 @@ These themes are:
     previous/next page using the keyboard's left and right arrows.  Defaults to
     ``False``.
 
+  .. versionadded:: 3.1
```
tk0miya: This breaks the bullet list. It causes the document hard to read.  Could you move this under each list item?

tk0miya (changes requested):
>Shall I add configuration options for those, too?

I feel nobody requests them at present. So it's not too late to consider to add them after requested.

doc/usage/theming.rst
mgeier: Done: 3baa6e09e2c488a29c2d6d18172aa47422dc4f1c

mgeier:
> > Shall I add configuration options for those, too?
> 
> I feel nobody requests them at present.

I don't think it's a surprise that nobody requests this.

"Normal" users will not request this because the most popular themes already have this option (I'm talking about the "TOC depth" option).

And most theme authors would rather create their own `globaltoc.html` than trying to add the option right here in Sphinx itself.

But if we *would* add this option, future theme authors would probably not have to create their own `globaltoc.html` anymore.

> So it's not too late to consider to add them after requested.

Indeed it's not too late.

I personally don't need this option, so for my own usage I don't care.

If somebody requests that for my theme in the future, I will come back to this.

tk0miya (approved):

tk0miya:
Merged. Thank you always!

Issue #7421, opened 2020-04-06
Regular expressions removed from sphinx/domains/c.py (e.g. c_funcptr_sig_re) break downstreams


dvzrv:
**Describe the bug**

`c_funcptr_sig_re` amongst other regular expressions have been removed from sphinx/domains/c.py in commit 0f49e30c51. These are used in the Linux kernel html documentation scripts.

When trying to compile the HTML documentation for Linux using sphinx 3.0.0 this leads to:

```
  SPHINX  htmldocs --> file:///build/linux-rt/src/linux-5.6.2/Documentation/output
  PARSE   include/uapi/linux/dvb/audio.h
  PARSE   include/uapi/linux/dvb/ca.h
  PARSE   include/uapi/linux/dvb/dmx.h
  PARSE   include/uapi/linux/dvb/frontend.h
  PARSE   include/uapi/linux/dvb/net.h
  PARSE   include/uapi/linux/dvb/video.h
  PARSE   include/uapi/linux/videodev2.h
  PARSE   include/uapi/linux/media.h
  PARSE   include/uapi/linux/cec.h
  PARSE   include/uapi/linux/lirc.h
Running Sphinx v3.0.0

Extension error:
Could not import extension cdomain (exception: cannot import name 'c_funcptr_sig_re' from 'sphinx.domains.c' (/usr/lib/python3.8/site-packages/sphinx/domains/c.py))
make[1]: *** [Documentation/Makefile:81: htmldocs] Error 2
```

This change was introduced without any notification in the changelog of this release and breaks downstream builds of the Linux Kernel in distributions (when building the HTML documentation).

**To Reproduce**

```
wget https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.6.2.tar.xz
tar xvJf linux-5.6.2.tar.xz
cd linux-5.6.2/
make htmldocs
```

**Expected behavior**
The Linux HTML documentation is generated.

**Your project**
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/sphinx/cdomain.py#n40

**Screenshots**

**Environment info**
- OS: Arch Linux
- Python version: 3.8.2
- Sphinx version: 3.0.0
- Sphinx extensions:  sphinx_rtd_theme
- Extra tools:

**Additional context**

- https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/Makefile#n81
[...]

PR #7468, opened 2020-04-12
C, add scoping directives

jakobandersen:
Subject: add scoping directives to the C domain in the same style as the C++ domain.

### Feature or Bugfix

- Feature

### Relates
Will give more options for writing documentation. E.g., see #7421.

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -706,6 +706,72 @@ Inline Expressions and Types
    .. versionadded:: 3.0
 
 
+Namespacing
+~~~~~~~~~~~
+
+.. versionadded:: 3.1
+
+The C language it self does not support namespacing, but it can sometimes be
+useful to emulate it in documentation, e.g., to show alternate declarations.
+The feature may also be used to document members of structs/unions/enums
+separate from their parent declaration.
+
+The current scope can be changed using three namespace directives.  They manage
+a stack declarations where ``c:namespace`` resets the stack and changes a given
+scope.
+
+The ``c:namespace-push`` directive changes the scope to a given inner scope
+of the current one.
+
+The ``c:namespace-pop`` directive undoes the most recent
+``c:namespace-push`` directive.
+
+.. rst:directive:: .. c:namespace:: scope specification
+
+   Changes the current scope for the subsequent objects to the given scope, and
+   resets the namespace directive stack. Note that nested scopes can be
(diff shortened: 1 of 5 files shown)


doc/usage/restructuredtext/domains.rst
```diff
@@ -706,6 +706,72 @@ Inline Expressions and Types
...
+
+The current scope can be changed using three namespace directives.  They manage
+a stack declarations where ``c:namespace`` resets the stack and changes a given
+scope.
+
+The ``c:namespace-push`` directive changes the scope to a given inner scope
+of the current one.
+
+The ``c:namespace-pop`` directive undoes the most recent
+``c:namespace-push`` directive.
```
tk0miya, line 727: I think documentation is not a program. So I prefer not to add the push and pop mechanism into document. Is it natural to use the mechanism in the C world? If not, I think the declarative way; only using `c:namespace` directive is better.

jakobandersen: One reason is for consistency with the C++ domain, and while the languages are of course different, I see it more as a documentation tool. The situation I have used it in is to have a namespace declaration in the top of a document or in the beginning of an example. Then locally, e.g., for a struct, push to get into that scope and document members and pop afterwards. For example:
```rst
Example 1
---------

.. c:namespace:: @example_1

The datastructure can be document as

.. c:struct:: MyStruct

with the following members

.. c:namespace-push:: MyStruct

.. c:var:: int a

.. c:namespace-pop::

```

Basically, it offers convenience for the style of writing where nested declarations are written (semi-)independently of the parent declaration.

tk0miya: IMO, using full namespace declaration is explicit than the push/pop approach.
```
Example 1
---------

.. c:namespace:: @example_1

...

.. c:namespace:: @example_1.MyStruct

...

.. c:namespace:: @example_1
```

Further speaking, it would be better if `c:struct` directive automatically modifies the namespace between processing.

Anyway, I did not know that the approach has been already added to the cpp domain. So either is okay to me.

jakobandersen: The ``c:struct`` directive is indeed tracking the scope, but that only applies for nested declarations. This is aimed at declarations inside the scope, but textually written somewhere else. While the full namespace declaration is indeed more explicit, it becomes very impractical when script-generated rst is involved or when the distance between the declarations is large.

Issue #4058, opened 2017-09-14
Autosummary extension has many deficits for proper using


maikelmeyers:
First to mention is that the bug found in #3900 is not really fixed, because the same applies to using class attributes `{{ attributes }}` in the templates. Autosummary does not find attributes anymore. Also the fix that was done for this in b3e0e29 is maybe not a good solution because `{{ methods }}` now returns also all inherited methods of a class.

Autosummary with it's templates is a really good solution for auto-documenting a library package without the need for any hand-written ResT file except the master-doc. But at the moment the available variables in the templates do not have enough granularity to achieve this. Beside of the available variables:

- members
- functions
- classes
- exceptions
- methods
- attributes

I miss the following variables for having the ability to fully automize the documentation:

- modules
- packages (sub packages)
- module-level variables
- class attributes
- instance attributes
- inherited class attributes
- inherited instance attributes
- methods
- inherited methods
- static methods
- inherited static methods

With this variables it would be possible to walk recursively thru a package with 3 templates (packages.rst, modules.rst and class.rst).  Especially the modules and packages variables would be important for this). And it would give us the full freedom for the layout of a package- module, and class- page. In example I could write my own headers in between the different groups of members inside the summary or in the details section because this variables mirror the granularity which we have with the autodoc directives.

### Environment info

OS: Windows 10
Python version: 2.7
Sphinx version: 1.6.4

jorisvandenbossche:
FYI, I opened a PR to properly fix #3900 (so to make sure also attributes are available again in the autosummary template): https://github.com/sphinx-doc/sphinx/pull/4247

tk0miya:
Fixed by #4247.
Thank you for reporting

jorisvandenbossche:
I am not the original reporter of the issue, but to me it seems that his enhancement request was more broad than what #4247 fixed (expose more variables to autosummary).

tk0miya:
oh, I'd misunderstood. Thanks,

TheMagSec:
I'm working on a recursive auto summary to cover packages, sub-modules and sub-packages.

PR #7469, opened 2020-04-12
autosummary tables include module attributes

dfremont:
Subject: Adds module attributes to autosummary tables

### Feature or Bugfix

- Feature

### Purpose

As pointed out in #4058, autosummary currently does not expose module attributes (called "module-level variables" there). Even if they are documented using doc-comments, such attributes do not get mentioned in the generated summary tables. For example, if you use autosummary on this module:

```
#: A module attribute which should be mentioned in the summary
my_attr = 3.14

def func():
    """A function which is actually mentioned in the summary"""
    pass
```

then the generated page documents both `my_attr` and `func`, but only lists `func` in the summary tables. I work on a library which uses module attributes as part of the public API, and would like to make them prominent by listing them in the summary tables for the module; this small PR makes that possible.

### Detail

The PR makes three changes:

- Defines the `attributes` and `all_attributes` template variables (already used for classes) when autosummarizing modules. The `all_attributes` variable lists all module attributes with doc-comments; the `attributes` variable lists the subset of these whose name does not start with an underscore.
- Updates the autosummary documentation to reflect the above (I don't document `all_attributes` since the `all_` versions of all such template variables are undocumented and I'm not sure whether they're supposed to be).
- Adds a new section to the autosummary module template called "Module Attributes", which lists the elements of `attributes` in the same way as currently done for `functions`, etc.

 Some notes:

- I'm not familiar with Sphinx, so it's possible the way I'm searching for the documented module attributes is wasteful (in particular I create a `ModuleAnalyzer` since I couldn't figure out how to access an existing one). Please feel free to completely change my implementation: I would like Sphinx to support this feature, but don't care how it's done.
- If you would rather not have this PR affect existing code, you can remove the change to the module template. Then people will still have the option to make a summary table of module attributes, even if one isn't included by default.
- I'm afraid I don't understand Sphinx enough to write an automated test for this PR - sorry! It does work correctly with my code base, at least.


dfremont:
Oh dear, it looks like a test is failing due to a docstring being processed twice: presumably this is related to the issue I mentioned about creating a new `ModuleAnalyzer`. If someone could give me a hint about how to properly orchestrate this I can try to fix it.

tk0miya:
>Oh dear, it looks like a test is failing due to a docstring being processed twice

Finally, I found the reason of this behavior. I just post a fix #7580 now. I'll self-review and merge it tomorrow. Please wait a moment.

sphinx/ext/autosummary/generate.py
```diff
@@ -197,6 +197,21 @@ def get_members(obj: Any, types: Set[str], include_public: List[str] = [],
             get_members(obj, {'class'}, imported=imported_members)
         ns['exceptions'], ns['all_exceptions'] = \
             get_members(obj, {'exception'}, imported=imported_members)
+
+        # Find module attributes with docstrings
+        attrs, public = [], []
+        from sphinx.pycode import ModuleAnalyzer, PycodeError
```
tk0miya: Could you move this line to the top of this module, please? Basically we go importing there if no reason.

sphinx/ext/autosummary/templates/autosummary/module.rst
```diff
@@ -2,6 +2,17 @@
 
 .. automodule:: {{ fullname }}
 
+   {% block attributes %}
+   {% if attributes %}
+   .. rubric:: Module Attributes
+
+   .. autosummary::
+   {% for item in attributes %}
+      {{item}}
```
tk0miya: Please add spaces between a template variable.

sphinx/ext/autosummary/generate.py
```diff
@@ -197,6 +197,21 @@ def get_members(obj: Any, types: Set[str], include_public: List[str] = [],
...
+        try:
+            analyzer = ModuleAnalyzer.for_module(name)
+            attr_docs = analyzer.find_attr_docs()
+            for _, attr_name in attr_docs:
+                if attr_name in ns['members']:
+                    attrs.append(attr_name)
+                    if not attr_name.startswith('_'):
+                        public.append(attr_name)
+        except PycodeError as err:
+            pass
```
tk0miya: Please make a small internal function like `get_members()`. I think it makes this function a bit readable.

sphinx/ext/autosummary/generate.py
```diff
@@ -197,6 +197,21 @@ def get_members(obj: Any, types: Set[str], include_public: List[str] = [],
...
+        ns['attributes'], ns['all_attributes'] = attrs, public
```
tk0miya: Please add these members to `ns['members']` too. It should contain all of the members including attributes.

doc/usage/extensions/autosummary.rst
```diff
@@ -260,8 +260,8 @@ The following variables available in the templates:
 
 .. data:: attributes
 
-   List containing names of "public" attributes in the class.  Only available
-   for classes.
+   List containing names of "public" attributes in the class/module.  Only
+   available for classes and modules.
```
tk0miya: Please add here the changes in this PR. For example:
```
.. versionchanged:: 3.1

   Attributes of modules are supported.
```

tk0miya (changes requested):
LGTM with nits.

To add a test case, please do the following steps:

1. add a module variable to `tests/roots/test-ext-autosummary/autosummary_dummy_module.py`
2. update a testcase `test_autosummary_generate()` in `tests/test_ext_autosummary.py` to confirm that the stub file will be generated expectedly


Thanks,

tk0miya:
Sorry, I closed this PR by my mistake. So reopened again.

tk0miya:
> Finally, I found the reason of this behavior. I just post a fix #7580 now. I'll self-review and merge it tomorrow. Please wait a moment.

I merged the fix to 3.x branch. Please merge it to your branch.

sphinx/ext/autosummary/generate.py
```diff
@@ -197,6 +197,21 @@ def get_members(obj: Any, types: Set[str], include_public: List[str] = [],
...
             get_members(obj, {'exception'}, imported=imported_members)
+
+        # Find module attributes with docstrings
+        attrs, public = [], []
+        from sphinx.pycode import ModuleAnalyzer, PycodeError
+        try:
+            analyzer = ModuleAnalyzer.for_module(name)
+            attr_docs = analyzer.find_attr_docs()
+            for _, attr_name in attr_docs:
+                if attr_name in ns['members']:
```
tk0miya: This is not a good way to filter the documented attributes. The key of `attr_docs` is pairs of namespace and name. The namespace is a parent name of the variable. The namespace of global attributes is an empty string. This code should ensure empty. If none, it would pick up under class object.

sphinx/ext/autosummary/generate.py
tk0miya: This is my wrong. withdrawing.

dfremont:
Thanks for the detailed comments and help writing the test! I think I've adjusted everything you pointed out above (and fixed a bug where I got the lists of public and all attributes backwards). Let me know if I should change anything else.

doc/usage/extensions/autosummary.rst
```diff
@@ -285,8 +285,12 @@ The following variables available in the templates:
...
+
+    .. versionchanged:: 3.1
+
+    Attributes of modules are supported.
```
tk0miya, line 293: This content should be indented inside the `versionchanged` directive. But it's minor bug. So I'll work this after merging.

tk0miya (approved):
Thank you for update. LGTM!

PR #7471, opened 2020-04-13
C, parse attributes

jakobandersen:
### Feature or Bugfix
- Feature
- Bugfix

The 3.0 version of the C domain got more strict, resulting in warnings for declarations that are not technically C, but are rather common, so it would be nice to release this PR soon.

### Purpose
Parse attributes in declarations and allow users to define their own identifiers to be parsed as attributes (e.g., when the actual attribute is hidden behind a macro). For example:
```rst
.. c:function:: LIGHTGBM_C_EXPORT int LGBM_BoosterFree(int handle)
```
Here the ``LIGHTGBM_C_EXPORT`` identifier is a macro that expands to a symbol visibility attribute.

### Relates
- michaeljones/breathe#500

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,9 @@ Deprecated
 Features added
 --------------
 
+* C, parse attributes and add :confval:`c_id_attributes`
+  and :confval:`c_paren_attributes` to support user-defined attributes.
+
 Bugs fixed
 ----------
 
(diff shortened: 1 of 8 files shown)


tk0miya:
I'm not good at the implementation of C/C++ domain. So I can't review this sooner.
But I can say okay to merged this into 3.0.x branch if you worried about merging "enhancement" changes into the branch.

jakobandersen:
> But I can say okay to merged this into 3.0.x branch if you worried about merging "enhancement" changes into the branch.

Thanks, I'll merge it as soon as the CI goes through.

PR #7476, opened 2020-04-13
Relbar breadcrumb should contain current page

timhoffm:
The relbar at the top contains the sequence of parents, which looks like [breadcrumb navigation](https://en.wikipedia.org/wiki/Breadcrumb_navigation):
grafik

However, breadcrumbs also contain the current page itself (see linked wikipedia article above). This makes sense as it's a self-contained display of the current location. Having the current page missing always confuses me.

This PR adds the current page as well:

grafik

This is a change of the layout. If you have concerns about backward-compatibility, we could make it configurable via an option. (But IMHO it's generally the right thing to do, so not sure if we can just regard this as bug fix).


tk0miya (changes requested):
LGTM. But it would be better to merge this into the 3.x branch. Could you rebase this please?

timhoffm:
Thanks! Rebased.

tk0miya (approved):

tk0miya:

PR #7481, opened 2020-04-14
basic theme: Add right margin to footnote/citation labels

mgeier:
The problem is not visible with single-digit footnote labels.

But with longer labels (e.g. when using multiple back-links, or with citations), there is no space between the label and the description text.

### Feature or Bugfix

- Bugfix

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -506,6 +506,7 @@ li > p:last-child {
 dl.footnote > dt,
 dl.citation > dt {
     float: left;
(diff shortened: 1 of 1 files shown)


tk0miya:
Could you paste screenshots for before and after this change?

mgeier:
Sure, I should have done that in the first place:

```rest
Some text [99]_.

.. [99] A Footnote.
```

Using `html_theme = 'basic'`, but of course this also affects other themes.

Before this PR:

image

After this PR:

image

tk0miya (approved):
Thank you for explanation. LGTM!

PR #7483, opened 2020-04-14
Fix #7477: imgconverter: Invoke "magick convert" command on Windows

### LINKED ISSUE #7477: Imgconverter fails to find ImageMagick on Windows
ImageMagick fails on Windows due to backslash acting as a Python delimiter.

## Issue
ImageMagick is installed on my Windows 10 (Build 2004) machine, but it fails with ```Invalid Drive Specification```. Opening a powershell and running ```convert mysvg.svg mysvg.png``` works as intended, so this is a Sphinx specific issue.

## Workaround
Specify ImageMagick path using
```image_converter = "D:\\System Programs\\ImageMagick-7.0.10-Q16\\convert.exe"```

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7477

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,8 @@ Dependencies
 Incompatible changes
 --------------------
 
+* #7477: imgconverter: Invoke "magick convert" command by default on Windows
+
 Deprecated
 ----------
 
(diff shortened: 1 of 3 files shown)


Daltz333 (commented):
Please note that ``os.name`` only works on Python3.6+

tk0miya:
Really? python-doc does not mention that. But, I found Sphinx uses `sys.platform` in other modules. So I'll use it instead.
https://docs.python.org/3/library/os.html#os.name

PR #7484, opened 2020-04-14
basic theme: Avoid clashes between sidebar and other blocks

mgeier:
When following a `sidebar`, these blocks should not be squished next to the sidebar, but they should start *after* the end of the sidebar in their full width:

* admonitions
* topics
* code blocks
* further sidebars

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -320,13 +320,14 @@ div.sidebar {
     background-color: #ffe;
     width: 40%;
     float: right;
+    clear: right
 }
 
 p.sidebar-title {
     font-weight: bold;
 }
 
-div.admonition, div.topic, div.sidebar, pre {
+div.admonition, div.topic, pre {
     clear: both;
 }
 
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -326,6 +326,10 @@ p.sidebar-title {
     font-weight: bold;
 }
 
+div.admonition, div.topic, div.sidebar, pre {
+    clear: both;
+}
+
 /* -- topics ---------------------------------------------------------------- */
 
 div.topic {
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -320,7 +320,7 @@ div.sidebar {
     background-color: #ffe;
     width: 40%;
     float: right;
-    clear: right
+    clear: right;
 }
 
 p.sidebar-title {


mgeier:
I think it might be better to clear only `"right"` in the case of a `sidebar`.
Feel free to squash ae9ae3e749bbc633365256915802615f0f108147.

tk0miya:
Could you paste screenshots for before and after this change?
I can't imagine what is happened now.

mgeier:
Sure.

```reST
.. sidebar:: Sidebar 1

   Text.

.. sidebar:: Sidebar 2

   Text.

.. topic:: Topic

   Text.
```

Before:

image

After:

image

The second one is better, isn't it?

sphinx/themes/basic/static/basic.css_t
```diff
@@ -320,12 +320,17 @@ div.sidebar {
     background-color: #ffe;
     width: 40%;
     float: right;
+    clear: right
```
tk0miya: Could you add a semicolon here?

tk0miya (changes requested):
LGTM with nits.

tk0miya:
Thank you for explanation. I've forgotten how sidebars are rendered. +1 for bugfix.

sphinx/themes/basic/static/basic.css_t
mgeier: Oh, sorry, I somehow missed that ... I've added it in 57fe9ddc5aac53f51ba3eb25a661f3bdd49cb832.

tk0miya (approved):

tk0miya:
Thanks!

return42:
> The second one is better, isn't it?

depends on the use case .. TLDR; its just another broken solution. 

Since you will normally not really often have one after the other sidebar I prefer as it was .. Did you ever seen a contents and a sidebar directive on the top of a document .. [example](https://raw.githubusercontent.com/return42/searx/searx-next/docs/admin/installation-nginx.rst) .. "second" looks horrible while first is overlapping, which looks much nicer to me.


before  #7484
===========
Bildschirmfoto vom 2020-06-18 16-22-13


after  #7484
==========

Bildschirmfoto vom 2020-06-18 16-23-35

return42:
FWIW: here is my hot-fix (not widely tested):

````CSS
div.sidebar {
    clear: none;
}

div.admonition, div.topic, pre {
    clear: none;
}
````

mgeier:
I'm quite sure that the changes for `div.sidebar` and for code blocks are good for all themes.
@return42 Would you agree?

I admit that I might have been over-eager in adding `div.admonition` and `div.topic` into the mix.
It would probably be better to remove those from the `basic` theme and add them individually to those themes where it is necessary.

@return42 Would you like to make a PR for this (I think it would be good to use the `3.1.x` branch)?

mgeier:
I'm suggesting to partially revert this PR in #7878.

PR #7485, opened 2020-04-14
C++, add support for parameterized noexcept specifier in function dec…

jbab:
…larations
Subject: C++, Add support for parameterized noexcept specifier in function declarations

### Feature or Bugfix
- Feature

### Purpose

A function declaration such as

    cpp:function:: vector(vector&&) noexcept(std::is_nothrow_move_constructible_v<T>)

lead to a parse failure with the message _Parameterised "noexcept" not yet implemented._

This PR implements the missing feature,

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -10,7 +10,7 @@
 
 import re
 from typing import (
-    Any, Callable, Dict, Generator, Iterator, List, Tuple, Type, TypeVar, Union
+    Any, Callable, Dict, Generator, Iterator, List, Tuple, Type, TypeVar, Union, Optional
 )
 
 from docutils import nodes
@@ -1266,7 +1266,7 @@ class ASTNoexceptExpr(ASTExpression):
         self.expr = expr
 
     def _stringify(self, transform: StringifyTransform) -> str:
-        return "noexcept(" + transform(self.expr) + ")"
+        return 'noexcept(' + transform(self.expr) + ')'
 
     def get_id(self, version: int) -> str:
         return 'nx' + self.expr.get_id(version)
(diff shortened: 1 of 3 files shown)


jakobandersen:
Thanks a lot for your effort. However, I do not believe that ``ASTNoexceptExpr`` is the right element to adapt here as it is for the other kind of noexcept. Instead I suggest to add as new ``ASTNoexceptSpec`` (located nearby the AST class for parameters and qualifiers), which models ``noexcept`` or ``noexcept(<constant-expression>)``.

sphinx/domains/cpp.py
```diff
@@ -5495,11 +5514,14 @@ def _parse_parameters_and_qualifiers(self, paramMode: str) -> ASTParametersQuali
         initializer = None
         self.skip_ws()
         if self.skip_string('noexcept'):
-            exceptionSpec = 'noexcept'
-            self.skip_ws()
-            if self.skip_string('('):
-                self.fail('Parameterised "noexcept" not yet implemented.')
-
+            if self.skip_string_and_ws('('):
+                expr = self._parse_expression()
```
jakobandersen: Please change this to use ``_parse_constant_expression()``.

tox.ini
```diff
@@ -17,7 +17,7 @@ description =
     py{35,36,37,38,39}: Run unit tests against {envname}.
     du{12,13,14}: Run unit tests with the given version of docutils.
 deps =
-    git+https://github.com/html5lib/html5lib-python  # refs: https://github.com/html5lib/html5lib-python/issues/419
+    git+https://github.com/html5lib/html5lib-python# refs: https://github.com/html5lib/html5lib-python/issues/419
```
jakobandersen: Looks like a stray change.

jakobandersen:
Thanks for the quick work. I made a two comments for some specific code, but otherwise it looks good. As a last request, it would be great if you could squash the commits to a single one in the end to avoid touching too many lines that were anyway changed back.

sphinx/domains/cpp.py
jbab: Was modeled according to `def _parse_unary_expression` (line 5031). Should it be changed there as well?

tox.ini
jbab: No it was not. In my environment (Ubuntu WSL on Windows 10) tox interpreted the whole string up to the `#` as URL. This change (or deleting the comment completely) was the only way I could run tox.
I admit I did not research further to find alternative solutions or the real root cause.

jakobandersen: Ah, ok. Perhaps @tk0miya knows more about the tox stuff?

sphinx/domains/cpp.py
jakobandersen: No, only here. I try to stay close to the grammar in the standard, which here says constant-expression, which excludes comma expressions and assignment expressions.

tox.ini
tk0miya: I can't understand what happened in @jbab's environment. But, I feel this change is strange. I wrote it as a comment, not a part of URL. -1 for this change.

jakobandersen: Right, it's a bit odd. @jbab, if you change this part back and submit a separate issue/PR to discuss this, then we can get the C++ part merged. Thanks.

jbab: ok, done.

jakobandersen:
Thanks!

PR #7487, opened 2020-04-15
Support singledispatch autodoc directives

### LINKED ISSUE #7447: Support singledispatch functions in autofunction
At the moment, my docs for https://aioca.readthedocs.io/en/latest/api.html are generated by interspersing text with ``autoclass`` and ``autofunction`` directives. This works well, apart from the fact that singledispatch overloaded signatures get lost when doing this compared to just doing ``automodule``. For now I have a workaround of multiple ``automodule`` directives restricted to single members, which makes sphinx moan about duplicate object references.

I would like the output of ``autofunction`` on a singledispatch function to be the same as the output of ``SingledispatchFunctionDocumenter`` with an ``automodule`` directive.

**Additional context**
I have tried merging the``SingledispatchFunctionDocumenter`` with the ``FunctionDocumenter`` and it seems to give the expected results, but it makes the ``FunctionDocumenter`` much bigger. I'm happy to submit a PR for this if that is helpful.

thomascobb:
### Feature or Bugfix
- Feature

### Purpose
Merge SingleDispatch*Documenters into their parent classes so that
autofunction and automethod work with singledispatch functions

### Detail
- Before, although automodule on singledispatch functions produced all signatures, autofunction on a singledispatch function only produced the first signature
- Now, autofunction and automethod work as expected
- Use of `mock.patch` has been replaced by an override of `add_line`

### Relates
- Builds on the work done in #7234
- Fixes #7447


sphinx/ext/autodoc/__init__.py
```diff
@@ -1062,56 +1062,34 @@ def document_members(self, all_members: bool = False) -> None:
...
-        sourcename = self.get_sourcename()
-
-        # intercept generated directive headers
-        # TODO: It is very hacky to use mock to intercept header generation
-        with patch.object(self, 'add_line') as add_line:
-            super().add_directive_header(sig)
-
-        # output first line of header
-        self.add_line(*add_line.call_args_list[0][0])
+    def add_line(self, line: str, source: str, *lineno: int) -> None:
```
tk0miya: -1 for such hack. I agree with my `patch.object()` hack is very ugly. But this is also. I can't accept that the "add_line" method generates content internally.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1062,56 +1062,34 @@ def document_members(self, all_members: bool = False) -> None:
...
+            # of the other implementations
+            self._singledispatch_flag = True
+
         super().add_directive_header(sig)
 
         if inspect.iscoroutinefunction(self.object):
             self.add_line('   :async:', sourcename)
 
-
-class SingledispatchFunctionDocumenter(FunctionDocumenter):
```
tk0miya, line 1071: Could you keep "singledispatch"ed classes as is? That is needed to keep compatibilities.

tk0miya (changes requested):

sphinx/ext/autodoc/__init__.py
thomascobb: Please could you expand on this comment. What compatibilities do I need to keep, is it just that SingledispatchFunctionDocumenter must exist, or that I can't change the code in it?

tk0miya: Yes, SingledispatchFunctionDocumenter must keep existing.

From the experience so far, there is some possibility that a 3rd party extension accesses such old classes. So we keep them during two major versions and let developers know the deprecation of the components.
https://github.com/sphinx-doc/sphinx/blob/3.x/CONTRIBUTING.rst#id20

thomascobb: The registry contains only one documenter for each function type. If we keep that, then autofunction will always point at the FunctionDocumenter. To make it support singledispatch functions, FunctionDocumenter must know how to document them, which means that SingledispatchFunctionDocumenter is effectively an alias for it. I have put it back, but moved the functionality into the baseclass so both can now document singledispatch functions

sphinx/ext/autodoc/__init__.py
thomascobb: Ok, how about this, I added a `format_extra_signature` to the `Documenter` and overrode that instead, is that any better?

thomascobb:
I tried a different approach. I have only modified SingledispatchFunctionDocumenter, so the test I added for the SingledispatchMethodDocumenter is currently failing. If you are happy with the approach then I will fix SingledispatchMethodDocumenter too.

sphinx/ext/autodoc/__init__.py
tk0miya: Sorry for confusing you. Only I need is keeping the classes as is. So they are not needed to be activated (registered). In other words, my main concern is the script that imports `SingledispatchFunctionDocumenter` should not be crashed.

sphinx/ext/autodoc/__init__.py
```diff
@@ -413,6 +413,10 @@ def format_signature(self, **kwargs: Any) -> str:
         else:
             return ''
 
+    def format_extra_signatures(self) -> List[str]:
```
tk0miya: TBH, I hesitate to add a new method to Documenter class because it is too fat class... It is very complex and hard to maintain for me. This is why I used mock module to hack `add_directive_header()`. -0 for adding a new method.

sphinx/ext/autodoc/__init__.py
thomascobb: So is:
```python
class SingledispatchFunctionDocumenter(FunctionDocumenter):
    pass
```
enough to maintain backwards compatibility? Even if it is never actually instantiated by sphinx?

sphinx/ext/autodoc/__init__.py
thomascobb: Ok, I'll go back to your original solution with mock...

thomascobb:
I could also strip down the Singledispatch*Documenter to just this:
```python
class SingledispatchFunctionDocumenter(FunctionDocumenter):
    pass
```
and not call add_autodocumenter() on them if that is helpful

sphinx/ext/autodoc/__init__.py
tk0miya: Yes. It is what I needed. After merging this, I'll add deprecation warning to them. And they will be removed in future.

tk0miya: Could you remove `    app.add_autodocumenter(SingledispatchFunctionDocumenter)` entries from setup function?

tk0miya (approved):
LGTM! Thank you for your patience :-)

PR #7489, opened 2020-04-15
C++, fix spacing issue in east-const notation

jbab:
Subject: Fix a spacing issue which occurs when using east-const notation

### Feature or Bugfix

- Bugfix

### Purpose
Documenting a function in this way

    cpp:function:: int foo(Type const& param)

renders output like this

int **foo**(**Type**_const_ &_param_)

That is, a space between **Type** and the following _const_ is missing.

This PR fixes this bug.

### Detail

The bugfix is a modification of the `ASTDeclSpecs.describe_signature` method, which did not properly insert a space between `trailingTypeSpec` and `rightSpecs`.

Note that the  `ASTDeclSpecs._stringify` method does not have the same problem, it behaves correctly. Therefore the problem is not observable in the unit tests in `test_domain_cpp.py`.

The problem was observed and the fix tested with HTML output.

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -2062,13 +2062,16 @@ class ASTDeclSpecs(ASTBase):
         if self.trailingTypeSpec:
             if addSpace:
                 signode += nodes.Text(' ')
+            numChildren = len(signode)
             self.trailingTypeSpec.describe_signature(signode, mode, env,
                                                      symbol=symbol)
-            numChildren = len(signode)
-            self.rightSpecs.describe_signature(signode)
-            if len(signode) != numChildren:
-                signode += nodes.Text(' ')
+            addSpace = len(signode) != numChildren
 
+            if len(str(self.rightSpecs)) > 0:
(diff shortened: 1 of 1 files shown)


jakobandersen:
Thanks for the PR! If you by chance have time, could you check if the same issue exist in the C domain? If so, feel free to fix it in this PR as well.

jbab:
> 
> 
> Thanks for the PR! If you by chance have time, could you check if the same issue exist in the C domain? If so, feel free to fix it in this PR as well.

Looking at the code of `ASTDeclSpecs.describe_signature` in `c.py` I am certain that the same problem does not exist there. The `rightSpecs` are added to an extra list, `modifiers`, instead of directly to`signode`. If `len(modifiers) > 0` a space is properly inserted and then the modifiers added to `signode`.

jakobandersen:
Great, thanks.

jakobandersen:
I rebased (and squashed) the PR onto the 3.0.x branch to get it out in 3.0.2 (PR #7496). Thanks again.

PR #7494, opened 2020-04-16
Deduplicate some code in `pycode.ast`

eric-wieser:
Some refactoring before an attempt at using `ast` to do more things, after discussion in #7385.

Hopefully this should be non-controversial.


tk0miya:
LGTM!

PR #7497, opened 2020-04-17
Rewrite unparse to use ast.NodeVisitor

eric-wieser:
This should make it possible to reuse the same visitor to generate RST code.

Follows on from #7494 

This is just a refactoring, motivated by:
* Being able to subclass this visitor with methods that return `List[docutils.Node]` instead of `str` (patch in the works)
* Being able to store state on `self` so as to keep track of the current precedence level (#7498)

### DIFF
--- a/sphinx/pycode/ast.py
+++ b/sphinx/pycode/ast.py
@@ -91,9 +91,6 @@ class _UnparseVisitor(ast.NodeVisitor):
         op = " %s " % self.visit(node.op)
         return op.join(self.visit(e) for e in node.values)
 
-    def visit_Bytes(self, node: ast.Bytes) -> str:
-        return repr(node.s)
-
     def visit_Call(self, node: ast.Call) -> str:
         args = ([self.visit(e) for e in node.args] +
                 ["%s=%s" % (k.arg, self.visit(k.value)) for k in node.keywords])
@@ -104,9 +101,6 @@ class _UnparseVisitor(ast.NodeVisitor):
         values = (self.visit(v) for v in node.values)
         items = (k + ": " + v for k, v in zip(keys, values))
         return "{" + ", ".join(items) + "}"
-
-    def visit_Ellipsis(self, node: ast.Ellipsis) -> str:
-        return "..."
 
     def visit_Index(self, node: ast.Index) -> str:
         return self.visit(node.value)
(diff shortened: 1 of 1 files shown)


sphinx/pycode/ast.py
```diff
@@ -64,113 +64,137 @@ def unparse(node: ast.AST) -> str:
...
-        else:
-            name += "=%s" % unparse(default)
-    return name
+    def visit_Lambda(self, node):
+        return "lambda %s: ..." % self.visit(node.args)
 
+    def visit_List(self, node):
+        return "[" + ", ".join(self.visit(e) for e in node.elts) + "]"
 
-def unparse_arguments(node: ast.arguments) -> str:
```
eric-wieser, line 139: Diff here is mostly just an extra level of indent, I'd recommend checking the "ignore whitespace" box to review.

eric-wieser:
Had a bit of a fight with mypy, but CI now passes.

sphinx/pycode/ast.py
```diff
@@ -64,113 +64,142 @@ def unparse(node: ast.AST) -> str:
...
-            return ", ".join(unparse(e) for e in node.elts)
-        else:
-            return "()"
-    elif sys.version_info > (3, 6) and isinstance(node, ast.Constant):
-        # this branch should be placed at last
-        return repr(node.value)
-    else:
-        raise NotImplementedError('Unable to parse %s object' % type(node).__name__)
 
+    if sys.version_info < (3, 8):
```
tk0miya, line 117: IMO, it would be better to move such online-patch codes to the bottom of the class definition. In addition, it would be better to sort methods alphabetically.

tk0miya (changes requested):
LGTM with nits.

sphinx/pycode/ast.py
eric-wieser: I was trying to keep them in the same order as before. Happy to alphabetize though.

tk0miya:
Remind: I will merge this after 1) conflicts resolved and 2) methods are sorted.

Not in a hurry. Please update if you have enough time.

sphinx/pycode/ast.py
tk0miya: done in #7625

eric-wieser:
Thanks for taking over.

I'll see if I can rebase my follow-up work.

PR #5049, opened 2018-06-06
html: Output ``canonical_url`` metadata if :confval:`html_baseurl` set

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- This is an alternative of #4193
- I added new confval `html_baseurl` to generate canonical_url metadata.
- I'd like to use it to another full-URL required metadatas (ex. OGP tags)

PR #7499, opened 2020-04-17
DOC: Use html_baseurl

mgeier:
The `canonical` URL has been added in #4764.

But afterwards, in #5049, the config value `html_baseurl` has been introduced to simplify that.

This PR shouldn't really change anything in the resulting HTML pages, except that I used `https` instead of `http`.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -27,6 +27,7 @@ html_static_path = ['_static']
 html_sidebars = {'index': ['indexsidebar.html', 'searchbox.html']}
 html_additional_pages = {'index': 'index.html'}
 html_use_opensearch = 'http://sphinx-doc.org'
+html_baseurl = 'https://www.sphinx-doc.org/en/master/'
 
 htmlhelp_basename = 'Sphinxdoc'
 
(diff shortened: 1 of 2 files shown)


doc/conf.py
```diff
@@ -27,6 +27,7 @@
 html_sidebars = {'index': ['indexsidebar.html', 'searchbox.html']}
 html_additional_pages = {'index': 'index.html'}
 html_use_opensearch = 'http://sphinx-doc.org'
```
mgeier, line 29: I have no clue about how OpenSearch works, but this is probably the wrong URL?

This leads to a https://www.sphinx-doc.org/en/master/_static/opensearch.xml with:

```
<Url type="text/html" method="get"
       template="http://sphinx-doc.org/search.html?q={searchTerms}"/>
```

This URL doesn't lead to the search page.

This should probably be changed as well?

tk0miya: Maybe yes. Please use `https://www.sphinx-doc.org/en/master` instead.

tk0miya (approved):

doc/conf.py
mgeier: I would prefer not to make this change in this PR, because it's a separate problem.

Also, as I said, I have no clue about OpenSearch, so I'd rather not touch this.

tk0miya: Okay, I'll do that after merging.

tk0miya:
Merged. Thanks!

PR #7502, opened 2020-04-18
doc: Fix html_use_opensearch of sphinx-doc.org

tk0miya:
### Feature or Bugfix
- Bugfix


doc/conf.py
```diff
@@ -26,7 +26,7 @@
 html_static_path = ['_static']
 html_sidebars = {'index': ['indexsidebar.html', 'searchbox.html']}
 html_additional_pages = {'index': 'index.html'}
-html_use_opensearch = 'http://sphinx-doc.org'
+html_use_opensearch = 'https://www.sphinx-doc.org/en/master/'
```
mgeier: I'm not sure, but shouldn't the last `/` be removed?

tk0miya: You're right. TBH, this is my first usage of this configuration!

Issue #7491, opened 2020-04-16
C++domain: Idiomatic spacing of ptr-operator and ref-qualifier


jbab:
## Problem

When declaring objects (e,g, types, members, functions) in the cpp domain, sphinx accepts
arbitrary spacing between types, cv-qualifiers and ref-qualifiers. It then produces output
in a consistent way (which is a plus) but which cannot be influenced by the user (which is a problem).

For example

    cpp:function:: int& foo(const Type & value, Resource * resource)

leads to output

> int &**foo**(*const* **Type** &*value*, **Resource** \**resource*)

The problem here is that the spacing chosen by sphinx is not the _idiomatic way_ used by the C++ community (I am aware that this is a bold statement, but https://en.cppreference.com and https://github.com/cplusplus/draft are good support for this.)

An idiomatic spacing for the example above should look somewhat like

> int& **foo**(*const* **Type**& *value*, **Resource**\* *resource*)

This is important, since IMO documentation should look like the code we write and it should follow conventions which are common in the specific language.

Supporting this kind of spacing could lead more C++ developers to consider using sphinx for documentation.

## Possible solutions

1. All spacing added by the user in the sphinx declarations are transfered as is to the output.
2. A new option for the C++ domain, say `cpp_idiomatic_spacing` is added which produces the desired output. The default leaves the output as is.
3. Sphinx only generates the "idiomatic" output and nothing else.

### Option 1

This probably requires huge changes to the output generating functions in `cpp.py`.
It would make the output depend entirely on the user input. This can be seen as an advantage (full control for the user), but also as a disadvantage (potential for inconsistent output).

### Option 2

Feasible, but would lead to lots of branching and duplication in the code and the tests.
The question is whether someone would still want the default behavior once the option is available and if the effort maintaining two output variants is worth while.

### Option 3

Implementable with relatively low effort.

Since the current choice of spacing by sphinx is somewhat arbitrary (or taken over from Python or C), a different spacing should be just as good. I am not sure, however, if this would still be considered a breaking change.

## Suggested solution

I would appreciate feedback to the above suggestions.
Personally I favor option 3 and would be willing to contribute a PR.

jakobandersen:
> The problem here is that the spacing chosen by sphinx is not the _idiomatic way_ used by the C++ community (I am aware that this is a bold statement, but https://en.cppreference.com and https://github.com/cplusplus/draft are good support for this.)

[...]

PR #7507, opened 2020-04-19
Address #7491 - Configuration of pointer/ref placement in C++ decls

vector-of-bool:
Subject: Allow configuration of the placement of the reference and pointer tokens in C++ declarations in the rendered output.

### Feature or Bugfix

- Feature - 2 additional configuration options, no changes for existing projects 

### Purpose
- This change adds options `cpp_ptr_placement` and `cpp_ref_placement`, which determine where the rendered output will attach the reference `&` or pointer `*` declaration token relative to the main type and the declared name. "right" (the current behavior and default) will attach towards the right on the declared name, whereas "left" will attach towrads the left, closer to the pointed-to/referred type.

### Relates
- #7491

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -8,9 +8,11 @@
     :license: BSD, see LICENSE for details.
 """
 
+import enum
 import re
 from typing import (
-    Any, Callable, Dict, Generator, Iterator, List, Tuple, Type, TypeVar, Union
+    Any, Callable, Dict, Generator, Iterator, List, Tuple, Type, TypeVar, Union,
+    Optional,
 )
 
 from docutils import nodes
@@ -541,6 +543,18 @@ _id_explicit_cast = {
     'const_cast': 'cc',
     'reinterpret_cast': 'rc'
 }
+
+
+class _Placement(enum.Enum):
+    Left = 'left'
+    Right = 'right'
+
+
+def _check_placement_arg(val: str) -> _Placement:
+    if val not in ('left', 'right'):
+        raise RuntimeError('Invalid C++ placement parmeter value '
+                           '"{}", should be either "left" or "right"'.format(val))
+    return _Placement(val)
 
 
 class _DuplicateSymbolError(Exception):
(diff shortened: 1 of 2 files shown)


jbab:
Thanks so much, I really appreciate this.

From a quick glance and a smoke test on my personal, very rudimentary C++-docs-with-sphinx attempts, results look good, except for one thing: I would expect the ampersand in a function ref-qualifier to be attached to the const (or volatile) in the left mode: `int f() const&`, not `int f() const &`.

As for the names of the options, I like them, they are neutral and convey their meaning. Also specifying the option value as an enum is a plus; in this way future extension are possible, e.g. a value "middle".

Having said that I am still a little concerned about the many output variations this results in and the impact this has on testing and maintenance. The crucial question is what is really needed. There must be a reason that another widely-used tool, [clang-format](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), fares well with only one option (and three values): *PointerAlignment*.

vector-of-bool:
I thought about reducing it to a single option for both, and maybe that's a good idea. I don't know of any codebases that mix-and-match the alignment of their `*` and `&` tokens. There's no compelling reason they can't be consolidated to a single option.

I didn't consider the case with ref-qualified functions. I'll look into that one.

sphinx/domains/cpp.py
```diff
@@ -543,6 +545,18 @@
...
+class _Placement(enum.Enum):
+    Left = 'left'
+    Right = 'right'
+
+
+def _check_placement_arg(val: str) -> _Placement:
+    if val not in ('left', 'right'):
+        raise RuntimeError('Invalid C++ placement parmeter value '
+                           '"{}", should be either "left" or "right"'.format(val))
+    return _Placement(val)
```
eric-wieser, line 557: Marginally faster as
```suggestion
    try:
        return _Placement(val)
    except ValueError:
        raise RuntimeError('Invalid C++ placement parmeter value '
                           '"{}", should be either "left" or "right"'.format(val)) from None
```

PR #7512, opened 2020-04-19
Check for __display_version__ in package_dir

mgeier:
Since #5352, the `__display_version__` (and therefore the context variable `sphinx_version`) has been wrong in Sphinx development builds for all Sphinx projects except the Sphinx docs themselves.

The problem is that `git` has been called without specifying a working directory, so it was running in the directory from wherever Sphinx was started.

Therefore, the Git hash was not representing the Sphinx version, but the version of the currently built docs!

### Feature or Bugfix

- Bugfix

### Relates
- #5352


tk0miya (approved):
OMG... you're right!

PR #7517, opened 2020-04-19
C++ domain: Add support for semicolon in declarations

jbab:
Subject: Add support for semicolon in declarations

### Feature or Bugfix

- Feature

### Purpose

Some people prefer the signatures of commented entities to look like real code, which in C++ means including the semicolon to end a declaration.

This PR enables to add a semicolon to documentation directives in the cpp domain and to obtain appropriate output.

*Examples:*

    cpp:function:: void f();

    cpp:class:: MyClass;

### Detail

The change is simple and backwards compatible. It's entirely the user's decision if he/she adds a semicolon, and only then will it be rendered.

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -109,7 +109,7 @@ T = TypeVar('T')
         simple-declaration ->
             attribute-specifier-seq[opt] decl-specifier-seq[opt]
                 init-declarator-list[opt] ;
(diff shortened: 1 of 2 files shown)


jakobandersen:
Thanks, good idea! I found a way to avoid involving ``ASTType`` too much, so #7545 replaces this PR.

Issue #7516, opened 2020-04-19
Sphinx inspects objects, can lead to needless build errors


a-recknagel:
**Describe the bug**
If I have an object in one of my modules that raises some kind of error in its `__getattr__`, it will prevent sphinx (with autodocs) from building docs for said module.

**To Reproduce**
Put something like this in a file and try to generate docs for it.
```python
class Foo:
    def __getattr__(self, item):
        raise RuntimeError('fufufu!')

foo = Foo()
```

See also https://github.com/a-recknagel/Flask-Shortcut/pull/7/checks?check_run_id=600063576 for an actual error, the workflow and the run logs should give instructions on how reproduce it in detail.

**Environment info**
- OS: on both windows and ubuntu
- Python version: 3.6.x
- Sphinx version: 3.0.2 (3.0.1 is still fine)
- Sphinx extensions:  sphinx.ext.autodoc

**Additional context**
This bug was introduced with 3.0.2 in this commit: https://github.com/sphinx-doc/sphinx/commit/e9b64373cd3728dc23e8e6b753a6a8d55e74af73#diff-48ca4d6544b092f35d729127500bff5fR581. I think it should be ok to just surround it with a bare `try..except`, since there is no info whatsoever which kind of errors a user might raise. In my case, it was flask's `request` proxy, which probably a few more people will run into. So here is part of the stacktrace to help them google the issue:

 ```
$ sphinx-build docs/ public/ -b html
Running Sphinx v3.0.2
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 2 source files that are out of date
updating environment: [new config] 2 added, 0 changed, 0 removed
reading sources... [ 50%] autodoc/flask_shortcut
Exception occurred:
  File "...\venv\lib\site-packages\flask\globals.py", line 38, in _lookup_req_object
    raise RuntimeError(_request_ctx_err_msg)
RuntimeError: Working outside of request context.

This typically means that you attempted to use functionality that needed
an active HTTP request.  Consult the documentation on testing for
information about how to avoid this problem.
The full traceback has been saved in ...\AppData\Local\Temp\sphinx-err-rzyl88fm.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```

jace:
[...]

PR #7518, opened 2020-04-20
fix for #7516

a-recknagel:
### Bugfix

Subject: Enable secure object inspection in autodoc'd modules.

### Purpose
Inspecting objects in user modules can be risky if they raise errors in their `__getattr__` implementations. One such case is the `flask.request` proxy which will complain about not being in a valid context if is accessed in any way and throws a `RuntimeError`.

### Detail
I just wrapped the access in a catchall, since users can raise whatever Exception they want. Maybe `BaseException`/a bare `except` would be even better, but it seemed risky to include `SystemExit`.

### Relates
#7516

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -578,9 +578,13 @@ class Documenter:
                 isprivate = membername.startswith('_')
 
             keep = False
-            if getattr(member, '__sphinx_mock__', False):
+            try:
+                is_mock = getattr(member, '__sphinx_mock__', False)
+            except Exception:
+                is_mock = False
+            if is_mock:
                 # mocked module or object
-                keep = False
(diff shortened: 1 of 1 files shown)


a-recknagel:
@tk0miya since you are familiar with this section of the code, can you review this fix to check if it is fine?

tk0miya (changes requested):
Thank you for your work. But we have similar implementation named `safe_getattr()`. So it would be better to use it instead of `getattr()`. Sorry, I forgot to use it...
Could you update this, please?

tk0miya:
In addition, could you rebase this into 3.0.x branch? Then I'll merge this and release as 3.0.3.

a-recknagel:
Closed in favor of https://github.com/sphinx-doc/sphinx/pull/7520

PR #7519, opened 2020-04-20
Change http:// to https://www in HTML footer

mgeier:
It seems to me that's the canonical Sphinx URL, isn't it?

I guess there are several more instance of old Sphinx URLs in the codebase, this was just the one I noticed.


tk0miya:
👍

PR #7541, opened 2020-04-23
basic theme: Add a "clearer" at the end of the "body".

mgeier:
You can see the problem when e.g. using `html_theme = 'classic'` and adding a `sidebar` at the very end of the page. It will reach into the footer.
Same for floating figures.

I'm not sure if this is the best solution, but it's simple and unobtrusive, so IMHO it's at least a step into the right direction.

### DIFF
--- a/sphinx/themes/basic/layout.html
+++ b/sphinx/themes/basic/layout.html
@@ -181,6 +181,7 @@
       {%- endif %}
           <div class="body" role="main">
             {% block body %} {% endblock %}
+            <div class="clearer"></div>
           </div>
       {%- if render_sidebar %}
         </div>


mscuthbert:
I'm afraid that this change broke the docs to my project, music21, which uses basic.css and a sidebar -- will switch to another style.  There should not be the large gap at the bottom of these paragraphs of text that waits for the long TOC to finish.

Screen Shot 2020-09-02 at 05 59 12

example: https://web.mit.edu/music21/doc/usersGuide/usersGuide_28_lyricSearcher.html

mgeier:
@mscuthbert I guess your theme can be fixed as shown in #7689. See also #7970.

PR #7542, opened 2020-04-23
basic theme: Make admonition/topic/sidebar scrollable

mgeier:
Most of the time this isn't a problem because content isn't that wide and un-breakable.
But one important use case are overly wide tables.

For example, using `html_theme = 'classic'` and this source text:

```rst
.. note::

    = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =
    x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x x
    = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = =

```

Before:

image

After:

image

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -321,6 +321,7 @@ div.sidebar {
     width: 40%;
     float: right;
     clear: right;
+    overflow-x: auto;
 }
 
 p.sidebar-title {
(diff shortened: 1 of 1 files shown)


mgeier:
I'm suggesting to partially revert this PR in #7878.

The `overflow` behavior suggested in this PR is good, but it clashes with the "float" behavior, and derived themes might find the latter more important. We can't have both, though.

PR #7543, opened 2020-04-23
basic theme: Add top and bottom margins to tables

mgeier:
All other things have margins, shouldn't tables have them as well?


tk0miya:
Thanks!

PR #7556, opened 2020-04-25
C, parse all types of array declarators

jakobandersen:
### Feature or Bugfix
- Feature
- Bugfix

### Purpose
In C the array declarators can contain ``static``, cvr-qualifiers, and VLA specification ``*``. These were not in the v3 rewrite (as they are C-specific).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,8 @@ Deprecated
 Features added
 --------------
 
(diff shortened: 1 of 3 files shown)


tk0miya:
@jakobandersen Could you merge this into 3.x branch also, please? It seems conflicted. It is helpful to me.

PR #7562, opened 2020-04-26
always use separate fields for param and type in ext.autodoc.typehints

afflux:
this fixes issues where annotations such as `x: typing.Tuple[int, int]`
will be transformed to `:field typing.Tuple[int, int] x:`, which renders
as `*int] x* (**typing.Tuple[int,**) -- `

### Feature or Bugfix

- Bugfix

### Detail
Minimum example:
```
mkdir -p sphinx_signature_type_hints
cd sphinx_signature_type_hints

cat <<'EOF' >type_hint_test.py
import typing
def test(x: typing.Tuple[int, int] = ()) -> int:
    return 1
EOF

mkdir -p docs

cat <<'EOF' >docs/conf.py
extensions = ["sphinx.ext.autodoc", "sphinx.ext.autodoc.typehints"]
autodoc_typehints = 'description'
EOF

cat <<'EOF' >docs/index.rst
.. automodule:: type_hint_test
    :members:
    :undoc-members:
EOF

mkdir -p html
python3.8 -m sphinx -vvv -W -b html --keep-going docs html
```

Unfortunately, I didn't find a testsuite for `sphinx.ext.autodoc.typehints`, so I can't promise that this doesn't have any weird side effects.

### DIFF
--- a/sphinx/ext/autodoc/typehints.py
+++ b/sphinx/ext/autodoc/typehints.py
@@ -104,24 +104,16 @@ def modify_field_list(node: nodes.field_list, annotations: Dict[str, str]) -> No
             continue
 
         arg = arguments.get(name, {})
-        field = nodes.field()
-        if arg.get('param') and arg.get('type'):
-            # both param and type are already filled manually
-            continue
-        elif arg.get('param'):
-            # only param: fill type field
+        if not arg.get('type'):
+            field = nodes.field()
             field += nodes.field_name('', 'type ' + name)
             field += nodes.field_body('', nodes.paragraph('', annotation))
-        elif arg.get('type'):
-            # only type: It's odd...
+            node += field
+        if not arg.get('param'):
+            field = nodes.field()
             field += nodes.field_name('', 'param ' + name)
             field += nodes.field_body('', nodes.paragraph('', ''))
-        else:
-            # both param and type are not found
-            field += nodes.field_name('', 'param ' + annotation + ' ' + name)
-            field += nodes.field_body('', nodes.paragraph('', ''))
-
-        node += field
+            node += field
 
     if 'return' in annotations and 'return' not in arguments:
         field = nodes.field()
--- a/tests/roots/test-ext-autodoc/index.rst
+++ b/tests/roots/test-ext-autodoc/index.rst
@@ -9,3 +9,5 @@
    :members:
 
 .. autofunction:: target.typehints.incr
+
+.. autofunction:: target.typehints.tuple_args
--- a/tests/roots/test-ext-autodoc/target/typehints.py
+++ b/tests/roots/test-ext-autodoc/target/typehints.py
@@ -1,3 +1,6 @@
+from typing import Tuple, Union
+
+
 def incr(a: int, b: int = 1) -> int:
     return a + b
 
@@ -28,6 +31,10 @@ class Math:
               ):
         # type: (...) -> None
         return
+
+
+def tuple_args(x: Tuple[int, Union[int, str]]) -> Tuple[int, int]:
+    pass
 
 
 def complex_func(arg1, arg2, arg3=None, *args, **kwargs):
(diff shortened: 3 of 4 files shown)


eric-wieser:
Tests probably should be added to this one:

https://github.com/sphinx-doc/sphinx/blob/bee44bedb100d325e4e9dad681521819bd774f5f/tests/test_ext_autodoc_configs.py#L564-L578

which pulls signatures from this file

https://github.com/sphinx-doc/sphinx/blob/bee44bedb100d325e4e9dad681521819bd774f5f/tests/roots/test-ext-autodoc/target/typehints.py#L14-L15

It might make sense to rewrite the test in the form of `test_autodoc_typehints_none`, which checks the generated RST rather than the final output.

tk0miya (approved):
LGTM!
I'll merge this after updates of testcase. Please follow @eric-wieser 's advice :-)

afflux:
> It might make sense to rewrite the test in the form of `test_autodoc_typehints_none`, which checks the generated RST rather than the final output.

Somehow, when I set up the test like this, I don't get any typehints in the output at all. Why is that?

```python
@pytest.mark.sphinx('text', testroot='ext-autodoc',
                    confoverrides={'autodoc_typehints': "description"})
def test_autodoc_typehints_description(app):
    options = {"members": None,
               "undoc-members": True}
    actual = do_autodoc(app, 'module', 'target.typehints', options)
    assert list(actual) == ...
```
yields:

```
[...]
.. py:function:: decr(a, b=1)
   :module: target.typehints
.. py:function:: incr(a, b=1)
   :module: target.typehints
[...]
```

So the built RST looks identical to autodoc_typehints=None.

eric-wieser:
That should be `'html'` not `'text'`, I assume?

afflux:
No difference with `'html'`.

Typehints are added on `object-description-transform`, which is never reached because apparently `do_autodoc` only does some basic generation. I'm clearly not very familiar with the output generation stages and I think it might be a good idea for you to fix the tests yourself to save both me and you some headaches.

eric-wieser:
Huh, that's a pain, but explains why that test is written differently.

I don't know how to proceed here, @tk0miya will have to advise what to do next.

tk0miya:
Yes, `do_autodoc()` is not good for this case as you said. Please use `app.build()` instead. I think it would be better to eric's idea.

1) Add a new function to https://github.com/sphinx-doc/sphinx/blob/3.x/tests/roots/test-ext-autodoc/target/typehints.py
   We already have a function having a bit complex type. So new one is not needed in this case.

   https://github.com/sphinx-doc/sphinx/blob/017af4ce2ef1fbc4387100b850e202b17f4c7450/tests/roots/test-ext-autodoc/target/typehints.py#L33-L35

2) Add an autofunction entry to https://github.com/sphinx-doc/sphinx/blob/3.x/tests/roots/test-ext-autodoc/index.rst
    `autofunction:: target.typehints.incr` is a good example for this. The function is also declared at `target/typehints.py`.

3) Update the case `test_autodoc_typehints_description()`. It tries to build `tests/roots/test-ext-autodoc` and check the generated `index.txt`.

afflux:
`test_ext_autodoc.test_autodoc()` fails with the suggested change. Please let me know if you want me to:

a. add the warning to the expected warnings in `test_ext_autodoc.test_autodoc()`
b. get rid of the warning by replacing the `Unknown` type in the test case
c. get rid of the warning by creating a new function without the `Unknown` type and use that for the `test_autodoc_typehints_description()` test case.

tk0miya:
Oh, sorry. Let's choose the c. It's better not to share an example between testcases. The last comment is my wrong.

tk0miya (approved):
LGTM!

tk0miya:

tk0miya:
Note: manually merged into 3.x branch

PR #7084, opened 2020-02-01
autodoc: Support type annotations for variables

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Show type annotations for variables

PR #7564, opened 2020-04-27
autodoc: Fix a logic error that causes annotations not to be shown for descriptors

eric-wieser:
This `if not self._datadescriptor` was originally here before type annotations were supported, to prevent showing `descr4 = <_Descriptor ...>` in the docs.

When type annotations were added in #7084, the support for appending `:type:` was put in the wrong place.
It should have been outside this if, not within it.

---

This diff is easiest to read with the "Hide whitespace changes" box checked.

Should I target 3.0.x with this, since I'd consider it a bug?


tk0miya:
Thanks!

eric-wieser:
@tk0miya, does it make sense to backport this to `3.0.x`? If so, I've done so at #7565, using the same commit.

tk0miya:
The patch branch (`3.0.x`) is used only for urgent release. So it is better not to add normal bug fixes. So -0 for to do that. I know this is not harmful and enough stable. but...

eric-wieser:
It's your project, so I'll let you make the call on the release process that works for you. Feel free to close that PR :)

tk0miya:
I am one of the maintainers, not mine :-)
I consider this is not important bug fix. so I determined not to merge this into 3.0.x. Thank you for your work!

PR #7568, opened 2020-04-27
Fix #7567: autodoc: parametrized types are shown twice for generic types

### LINKED ISSUE #7567: sphinx.util.typing.stringify duplicates type arguments
**Describe the bug**
`Foo[int]` is shown as `Foo[int][int]`

**To Reproduce**
```
In [6]: from typing import Generic, TypeVar

In [7]: T = TypeVar('T')

In [8]: class Foo(Generic[T]): pass

In [11]: sphinx.util.typing.stringify(Foo[int])
Out[11]: '__main__.Foo[int][int]'
```

**Expected behavior**
The output should be `__main__.Foo[int]`

**Environment info**
- OS: Windows 10
- Python version: 3.8.0
- Sphinx version: 3.0.2

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7567

### DIFF
--- a/sphinx/util/typing.py
+++ b/sphinx/util/typing.py
@@ -10,7 +10,7 @@
 
 import sys
 import typing
-from typing import Any, Callable, Dict, List, Tuple, TypeVar, Union
+from typing import Any, Callable, Dict, Generic, List, Tuple, TypeVar, Union
 
 from docutils import nodes
 from docutils.parsers.rst.states import Inliner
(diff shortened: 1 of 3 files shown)


eric-wieser:
Duplicates gh-7569, but your fix looks different to mine

sphinx/util/typing.py
```diff
@@ -93,6 +93,9 @@ def _stringify_py37(annotation: Any) -> str:
             return stringify(annotation.__args__[0])
         elif annotation._special:
             return qualname
+        elif issubclass(annotation.__origin__, Generic):  # type: ignore
+            # subclasses of Generic already contains the types for elements in its name
```
eric-wieser, line 97: I think this fails my testcase.
```
assert stringify(MyList[Tuple[int, int]]) == "test_util_typing.MyList[Tuple[int, int]]"
```

tk0miya: Thanks,

PR #7569, opened 2020-04-27
Do not emit type arguments twice

eric-wieser:
Fixes gh-7567

### Feature or Bugfix

- Bugfix


eric-wieser:
Ping @tk0miya, CI passes now - sorry for not letting you know I was working on the same thing as you before you spent the time!

tk0miya:
Don't mind. I feel your implementation is much better than mine :-)

tk0miya:
oh, I did not noticed the target branch is 3.0.x.

eric-wieser:
Whoops, sorry about that.

tk0miya:
don't mind! I think this is not a critical, but stable.

eric-wieser:
I assume this will end up in 3.1.0 as well?

PR #7571, opened 2020-04-27
Fix #7559: autodoc: misdetects a sync function is async

### LINKED ISSUE #7559: Sphix indicates that a sync function is async
This one is maybe my fault, but I'm not sure what I should be doing.

**Describe the bug**
I'm generating async methods from sync methods this way:
```python
    @functools.wraps(wrapped)
    def wrapper(*args, **kwargs):
        awaitable = wrapped(*args, **kwargs)
        return sync_await(awaitable=awaitable)
```
Because I use ``functools.wraps`` and after https://github.com/sphinx-doc/sphinx/pull/7222, Sphinx sees that my sync function wraps an async function and unwraps it in ``iscoroutinefunction``. the wrapped function is async, so the function is async.

When documenting, both the original and wrapped functions are seen as async.

**To Reproduce**
```
$ git clone https://github.com/peopledoc/procrastinate
$ cd procrastinate
$ pip install tox
$ tox -e docs
$ # open docs/_build/html/reference.html#procrastinate.tasks.Task.defer, it's noted as async
```

**Expected behavior**
Not really sure. For a staticmethod, for example, it makes sense to consider it async even if it's the wrapped method that is async.

**Your project**
https://github.com/peopledoc/procrastinate
https://github.com/peopledoc/procrastinate/blob/9dd6c296ddf59bb524e20cd6bd09703ff00a8f66/procrastinate/utils.py

**Screenshots**
Capture d’écran 2020-04-26 à 14 58 22

**Environment info**
- OS: All
- Python version: 3.7.7
- Sphinx version: Sphinx==3.0.2
- Sphinx extensions:
```
extensions = [
    "sphinx.ext.napoleon",
    "sphinx.ext.autodoc",
    "sphinx_autodoc_typehints",
    "sphinxcontrib.programoutput",
    "changelog",
]
```

**Additional context**
I'm not sure is this is because I shouldn't be using wraps if I expect the wrapper and the wrapped function to be seen as distinct, or if I'm doing something unexpected. I know I'm probably doing something weird in the first place with automatic synchronous functions generation, so if you think this is out of scope, I understand.

If there's a workaround, though, I'm interested.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7559

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -125,13 +125,15 @@ def unwrap(obj: Any) -> Any:
         return obj
 
 
-def unwrap_all(obj: Any) -> Any:
+def unwrap_all(obj: Any, *, stop: Callable = None) -> Any:
     """
     Get an original object from wrapped object (unwrapping partials, wrapped
     functions, and other decorators).
     """
     while True:
-        if ispartial(obj):
+        if stop and stop(obj):
+            return obj
+        elif ispartial(obj):
             obj = obj.func
         elif inspect.isroutine(obj) and hasattr(obj, '__wrapped__'):
             obj = obj.__wrapped__
@@ -287,7 +289,8 @@ def isroutine(obj: Any) -> bool:
 
 def iscoroutinefunction(obj: Any) -> bool:
     """Check if the object is coroutine-function."""
-    obj = unwrap_all(obj)
+    # unwrap staticmethod, classmethod and partial (except wrappers)
+    obj = unwrap_all(obj, stop=lambda o: hasattr(o, '__wrapped__'))
     if hasattr(obj, '__code__') and inspect.iscoroutinefunction(obj):
         # check obj.__code__ because iscoroutinefunction() crashes for custom method-like
         # objects (see https://github.com/sphinx-doc/sphinx/issues/6605)
--- a/CHANGES
+++ b/CHANGES
@@ -74,6 +74,7 @@ Bugs fixed
 * #6588: autodoc: Decorated inherited method has no documentation
 * #7469: autodoc: The change of autodoc-process-docstring for variables is
   cached unexpectedly
+* #7559: autodoc: misdetects a sync function is async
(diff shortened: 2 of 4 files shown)


eric-wieser:
What's the expected behavior of the following?

```python
def wrap_it(f):
   @wraps(f)
   def wrapper(*args, **kwargs):
       return f(*args, **kwargs)
   return wrapper

# async or not?
async_wrapped = wrap_it(async_func)
```

tk0miya:
For now, `inspect.iscoroutinefunction()` considers it is not a coroutine-function. So I followed the behavior. I know the original meaning of coroutine-function is a function that returns coroutine object. But it is a bit difficult to inspect it without the "async" keyword.

```
def my_coroutine_func():
    """This is coroutine function too"""
    async def get_coroutine_obj():
        pass

    return get_coroutine_obj()
```

eric-wieser:
Seems fair, thanks for explaining.

sphinx/util/inspect.py
```diff
@@ -125,13 +125,15 @@ def unwrap(obj: Any) -> Any:
...
+def unwrap_all(obj: Any, *, stop: Callable = None) -> Any:
     """
     Get an original object from wrapped object (unwrapping partials, wrapped
     functions, and other decorators).
     """
     while True:
-        if ispartial(obj):
+        if stop and stop(obj):
+            return obj
+        elif ispartial(obj):
```
eric-wieser, line 136: Can you add a test of a partial async function?

tk0miya: We already have it 
https://github.com/sphinx-doc/sphinx/blob/017af4ce2ef1fbc4387100b850e202b17f4c7450/tests/test_util_inspect.py#L473

tk0miya:
Merged. @eric-wieser Thank you for reviewing!

ewjoachim:
(Ticket OP here, thanks @eric-wieser too !)

PR #7572, opened 2020-04-27
Align equation numbers at the right margin

### LINKED ISSUE #2785: Bad alignment of equation links
This is a new issue for something I brought up in https://github.com/sphinx-doc/sphinx/pull/2643#issuecomment-231580477:

> [...] the link marker moves the equation number a bit to the left and therefore the the numbers don't align with the right text margin anymore, which looks a bit awkward (when not hovering over the equation).

Here's a screenshot for illustration (the mouse pointer is not visible in the screenshot, but I was hovering over the second equation and the ¶ sign is visible in grey):

equation-labels

Would it be possible to move the link marker further to the right so it protrudes into the right margin?

Another option would be to put the link marker to the left of the number, which would restore the nice right margin but probably would look a bit awkward?

When making the screenshot, I found another problem: The equations and their numbers seem to be aligned at the top, which looks a bit strange.
Ideally, they would be aligned at the baseline, but I guess this is not possible with MathJax/PNG/CSS, right?
I think it would still be better to vertically center them.

mgeier:
Closes #2785.

Example (with `html_theme = 'basic'`):

```rst
----

.. math:: x
    :label: x

----
```

Before:

image

After:

image

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -779,8 +779,7 @@ span.eqno {
 }
 
 span.eqno a.headerlink {
-    position: relative;
-    left: 0px;
+    position: absolute;
     z-index: 1;
 }
 


tk0miya:
Thanks!

PR #7601, opened 2020-05-03
autodoc: Show full traceback and exception type in warnings

eric-wieser:

### Feature or Bugfix

- Feature

### Purpose
This should help debug #7384

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -397,8 +397,8 @@ class Documenter:
                     # retry without arguments for old documenters
                     args = self.format_args()
             except Exception as err:
-                logger.warning(__('error while formatting arguments for %s: %s') %
-                               (self.fullname, err), type='autodoc')
+                logger.warning(__('error while formatting arguments for %s:') %
+                               self.fullname, type='autodoc', exc_info=True)
                 args = None
 
         retann = self.retann
@@ -751,7 +751,7 @@ class Documenter:
             # be cached anyway)
             self.analyzer.find_attr_docs()
         except PycodeError as err:
-            logger.debug('[autodoc] module analyzer failed: %s', err)
+            logger.debug('[autodoc] module analyzer failed:', excinfo=True)
             # no source file -- e.g. for builtin and C modules
             self.analyzer = None
             # at least add the module.__file__ as a dependency
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -751,7 +751,7 @@ class Documenter:
             # be cached anyway)
             self.analyzer.find_attr_docs()
         except PycodeError as err:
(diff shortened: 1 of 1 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -751,7 +751,7 @@ def generate(self, more_content: Any = None, real_modname: str = None,
             # be cached anyway)
             self.analyzer.find_attr_docs()
         except PycodeError as err:
-            logger.debug('[autodoc] module analyzer failed: %s', err)
+            logger.debug('[autodoc] module analyzer failed:', excinfo=True)
```
eric-wieser:
```suggestion
            logger.debug('[autodoc] module analyzer failed:', exc_info=True)
```

tk0miya:
+1

sphinx/ext/autodoc/__init__.py
```diff
@@ -397,8 +397,8 @@ def format_signature(self, **kwargs: Any) -> str:
                     # retry without arguments for old documenters
                     args = self.format_args()
             except Exception as err:
```
eric-wieser:
```suggestion
            except Exception:
```

sphinx/ext/autodoc/__init__.py
```diff
@@ -751,7 +751,7 @@ def generate(self, more_content: Any = None, real_modname: str = None,
             # be cached anyway)
             self.analyzer.find_attr_docs()
         except PycodeError as err:
```
eric-wieser:
```suggestion
        except PycodeError:
```

sphinx/ext/autodoc/__init__.py
```diff
@@ -396,9 +396,9 @@ def format_signature(self, **kwargs: Any) -> str:
                 except TypeError:
                     # retry without arguments for old documenters
                     args = self.format_args()
-            except Exception as err:
-                logger.warning(__('error while formatting arguments for %s: %s') %
-                               (self.fullname, err), type='autodoc')
+            except Exception:
+                logger.warning(__('error while formatting arguments for %s:') %
+                               self.fullname, type='autodoc', exc_info=True)
```
tk0miya, line 401: Note: I thought showing full stacktrace is a bit noisy for users at first. On the other hand, it is very important for developers including extension's authors. In addition, `format_args()` is one of the handlers that is unlikely to throw an exception. So +1 for this change.

tk0miya:
Just merged.

sphinx/ext/autodoc/__init__.py
eric-wieser: It might be reasonable to respect a verbose flag here, as some future work.

tk0miya: Absolutely!

tk0miya: I made a patch for this purpose #7608

PR #7603, opened 2020-05-03
Preserve exception info in raised SphinxWarning objects

eric-wieser:
This perhaps isn't enough to do anything useful yet, as `sphinx/command/build.py` discards this information


tk0miya:
I'm not familiar with the exception framework. So I don't any comment for this.

PR #7608, opened 2020-05-04
Show an exception info for warnings when -v option given (refs: #7601)

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: https://github.com/sphinx-doc/sphinx/pull/7601#discussion_r419113848

### DIFF
--- a/sphinx/util/logging.py
+++ b/sphinx/util/logging.py
@@ -10,6 +10,7 @@
 
 import logging
 import logging.handlers
+import sys
 from collections import defaultdict
 from contextlib import contextmanager
 from typing import Any, Dict, Generator, IO, List, Tuple, Union
@@ -131,6 +132,11 @@ class SphinxLoggerAdapter(logging.LoggerAdapter):
         self.log(VERBOSE, msg, *args, **kwargs)
 
     def process(self, msg: str, kwargs: Dict) -> Tuple[str, Dict]:  # type: ignore
+        if 'exc_info' not in kwargs:
+            exc_info = sys.exc_info()
+            if exc_info != (None, None, None):
+                kwargs['exc_info'] = True
+
         extra = kwargs.setdefault('extra', {})
         for keyword in self.KEYWORDS:
             if keyword in kwargs:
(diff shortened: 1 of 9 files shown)


eric-wieser:
I'm not sure if this approach makes sense. I'd argue you always want to record the exception info, and the flag should just affect the log formatter.

eric-wieser:
Or perhaps argued a different way, I think it should still be opt-in at the log call site, because a lot of exceptions are definitely useless.

So maybe the approach you have here, but have the `Filter` set `exc_info` to `None` instead of having it populate it?

tk0miya:
I thought 1) users would not like to read stacktraces. 2) Only Sphinx and extension developers are needed to read it. So it is enough to show it if requested.

eric-wieser:
I suppose my point is that "showing" stack traces is the job of the `Formatter`, not the job of the logger. If users want to connect up an additional log handler in `conf.py`, they should be given the full information in their `LogRecord`, and be allowed to decide what to do with it themselves.

tk0miya:
Okay, I understand your opinion. I don' have special opinion for the logger's design. So +1 for your idea.

tk0miya:
Updated.

sphinx/util/logging.py
```diff
@@ -131,6 +132,11 @@ def verbose(self, msg: str, *args: Any, **kwargs: Any) -> None:
         self.log(VERBOSE, msg, *args, **kwargs)
 
     def process(self, msg: str, kwargs: Dict) -> Tuple[str, Dict]:  # type: ignore
+        if 'exc_info' not in kwargs:
+            exc_info = sys.exc_info()
+            if exc_info != (None, None, None):
+                kwargs['exc_info'] = True
```
eric-wieser, line 138: I'd still probably argue this is up to the caller to decide, and therefore this change isn't necessary - there are cases when the exception is just not interesting, such as:
```python
try:
    val = some_dict[key]
except KeyError:
    log.info("Unable to find data for {}".format(key))
```
Seeing the exception here is just noise.

sphinx/util/logging.py
```diff
@@ -533,6 +539,19 @@ def format(self, record: logging.LogRecord) -> str:
...
+    def __init__(self, app: "Sphinx") -> None:
+        self.verbosity = app.verbosity
+        super().__init__()
+
+    def formatException(self, ei: Any) -> str:
+        """Show exception info only when verbose mode."""
+        if self.verbosity > 0:
+            return super().formatException(ei)
+        else:
+            return ''
```
eric-wieser, line 552: This looks good, thanks!

sphinx/util/logging.py
tk0miya: I'd like to give exc_info flag to all loggers. But it seems not needed. Okay, withdrawing.

eric-wieser: I think the rest of this patch is useful, I'd encourage you to withdraw only these four lines.

tk0miya: I feel too much to add it only for one exc_info call. Do you have a plan to add it more?

PR #7614, opened 2020-05-04
md5 OpenSSL FIPS mode fix #7611

### LINKED ISSUE #7611: md5 fails when OpenSSL FIPS is enabled
Attemting to build documentation on a Redhat Linux with SSL FIPS enabled, results in the following error:

```
File .../sphinx/builders/html/__init__.py", line 80, in get_stable_hash return md5(str(obj).encode()).hexdigest()
ValueError: error:060800A3:digital envelope routines:EVP_DigestInit_ex:disabled for fips
```

**Environment info**
- OS: Redhat Linux 7.8 - FIPS enabled
- Python version: 3.6.8
- Sphinx version: v3.0.3

The md5 algorithm is considered insecure and is not allowed by FIPS. A workaround is to supply the argument 'usedforsecurity=False' to the md5 call. Another solution would be changing the hash algorithm.
The 'usedforsecurity' is not supported on all platforms, so would have to be wrapped in a try/except.

There seems to be no workaround for this issue other than a systemwide disabling of FIPS, which I cannot do, so changing the caching code in sphinx seems like the only solution.

lhupfeldt:
This is a fix for  #7611

### DIFF
--- a/sphinx/util/__init__.py
+++ b/sphinx/util/__init__.py
@@ -10,6 +10,7 @@
 
 import fnmatch
 import functools
+import hashlib
 import os
 import posixpath
 import re
@@ -21,7 +22,6 @@ import warnings
 from codecs import BOM_UTF8
 from collections import deque
 from datetime import datetime
-from hashlib import md5
 from importlib import import_module
 from os import path
 from time import mktime, strptime
@@ -170,7 +170,7 @@ class FilenameUniqDict(dict):
         self._existing = state
 
 
-def fips_safe_md5(data=b'', **kwargs):
+def md5(data=b'', **kwargs):
     """Wrapper around hashlib.md5
 
     Attempt call with 'usedforsecurity=False' if we get a ValueError, which happens when
@@ -181,9 +181,9 @@ def fips_safe_md5(data=b'', **kwargs):
     """
 
     try:
-        return md5(data, **kwargs)  # type: ignore
+        return hashlib.md5(data, **kwargs)  # type: ignore
     except ValueError:
-        return md5(data, **kwargs, usedforsecurity=False)  # type: ignore
+        return hashlib.md5(data, **kwargs, usedforsecurity=False)  # type: ignore
 
 
 class DownloadFiles(dict):
@@ -195,7 +195,7 @@ class DownloadFiles(dict):
 
     def add_file(self, docname: str, filename: str) -> str:
         if filename not in self:
-            digest = fips_safe_md5(filename.encode()).hexdigest()
+            digest = md5(filename.encode()).hexdigest()
             dest = '%s/%s' % (digest, os.path.basename(filename))
             self[filename] = (set(), dest)
 
--- a/sphinx/util/__init__.py
+++ b/sphinx/util/__init__.py
@@ -173,16 +173,17 @@ class FilenameUniqDict(dict):
 def fips_safe_md5(data=b'', **kwargs):
     """Wrapper around hashlib.md5
 
-    Attempt call with 'usedforsecurity=False' if we get a ValueError, which happens when OpenSSL FIPS mode is enabled:
+    Attempt call with 'usedforsecurity=False' if we get a ValueError, which happens when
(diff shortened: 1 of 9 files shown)


sphinx/util/__init__.py
```diff
@@ -170,6 +170,22 @@ def __setstate__(self, state: Set[str]) -> None:
         self._existing = state
 
 
+def fips_safe_md5(data=b'', **kwargs):
```
tk0miya: It would be better to rename this to `md5()` simply.

tk0miya (changes requested):
LGTM with nits.

tk0miya:
BTW, we've used `hashlib.sha1` also. Is it needed to give `usedforsecurity=False` option to them too?

lhupfeldt:
I did not see any errors from sha1 when running the test suite on the FIPS enabled server. Thi is strange as sha1 is not allowed by FIPS either I think. It may depend on a policy on the server? I have not checked whether sha1 has the 'usedforsecurity' option.

tk0miya:
I guess you did not use features that uses SHA-1. Because Sphinx uses it only at 3 modules. 
```
$ grep -r hashlib sphinx | grep sha1
sphinx/ext/graphviz.py:from hashlib import sha1
sphinx/ext/imgmath.py:from hashlib import sha1
sphinx/transforms/post_transforms/images.py:from hashlib import sha1
```

Could you check that `sha1()` accepts the option? If you don't have time, I'll make an environment that enables FIPS.

lhupfeldt:
I ran the sphinx test suite, does that not test it?
I did have one error from image convert, because imagemagick is not installed on the server, maybe that is why I did not see any errors from sha1, it seems the use is image related. If that is the case, then I will not be able to test, imagemagick is not available for the version of redhat installed on the server.

Is it possible to handle possible sha1 issues in another patch, so that this can go in quickly?

tk0miya:
Yes, some external tools are needed to enable these modules.

I guess you can check it via python console. Could you check this please?
```
$ python
Python 3.8.2 (default, Mar  2 2020, 00:44:41)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
>>> from hashlib import sha1
>>> sha1(b'abcdefg').hexdigest()
'2fb5e13419fc89246865e7a324f476ec624e8740'
```
I think FIPS also disallows to use SHA-1, it raises an error like MD5.

lhupfeldt:
I checked the hashlib source, it seems usedforsecurity is supported generically for all algorithms.
I also ran sha1 as you suggested. I do not get an error like I did for md5 - this does not mean that it should not be called with usedforsecurity=False, I think it may still fail depending version of FIPS (or server policy?).

lhupfeldt:
Some info here possibly: https://stackoverflow.com/questions/49320993/how-to-enable-fips-mode-for-libcrypto-and-libssl-packaged-with-python

lhupfeldt:
I did the same change for sha1. I checked from console on the FIPS enabled server that it supports the flag. I ran the test suite on my laptop where all tests pass (ImageMagick is installed), so I think there should be no negative impact from this, but it may help in cases where FIPS forbids sha1.

tk0miya:
Great work! Thank you for your contribution :-)

lhupfeldt:
You're welcome.
Since the usedforsecurity flag seems to become standard from Python 3.9, you should consider reversing the order of the calls at that time.

lhupfeldt:
@tk0miya Can I ask when 3.0.4 is expected to be released? We are unable to build documentation at the moment.

tk0miya:
I don't have a plan for the release. But I understand your situation. Okay, let's ship it in this weekend.

lhupfeldt:
Thank you. Much appreciated.

Best regards / Venlig Hilsen
Lars Hupfeldt Nielsen
Hupfeldt IT

Phone: +45 2060 3546


Den tor. 21. maj 2020 kl. 18.02 skrev Takeshi KOMIYA <
notifications@github.com>:

PR #7616, opened 2020-05-05
doc: Use local inventory for intersphinx_mapping

JanZerebecki:
to avoid fetching it from https://docs.python.org/3/objects.inv during
build time.

### Feature or Bugfix
- Refactoring

### Purpose
- This is useful to make builds reproducible, as remote request prevent
that.


tk0miya:
Who will update `python3.inv`?

JanZerebecki:
Existing references will continue to work even if the inventory gets outdated. When someone adds a reference that is not yet in the copy, they will need to update it.

If any target link in the inventory stops working then the already built docs will also be broken. I hope the Python project doesn't do that. If that ever happens, then whoever fixes the URLs needs to update it or better instead fixes the original URL.

tk0miya:
You are saying the maintainers and contributors should pay the cost to keep the consistency of the links, right? Personally, I don't feel build is slow. There are no advantages of this PR for me. So -1 for this.

Note: I can understand to set `intersphinx_timeout` to our document. The python-doc is not be changed so often. So It's reasonable and awesome.

tk0miya:
There are no positive reactions from other maintainers. Closing.
Thank you for proposal.

PR #7617, opened 2020-05-05
doc: fix link to configuring warnings

JanZerebecki:
### Feature or Bugfix
- Bugfix

### Purpose
- This also fixes an undefined label error by removing the `:ref:` that
isn't supposed to be there.


tk0miya:
It seems the hyperlink is fine. Why did you post such a change?
https://www.sphinx-doc.org/en/master/devguide.html#deprecation-warnings

JanZerebecki:
My mistake: I removed the python intersphinx_mapping to make a non networked build work and misunderstood the syntax here.

PR #7624, opened 2020-05-06
refactor: autodoc: class processors on autofunction is no longer needed

tk0miya:
### Feature or Bugfix
- Refactoring


eric-wieser:
This change is fine because that code was unreachable, but I think you _do_ need the class processors - otherwise you can't pick up `type` comments on `__new__` and `__init__` within `autodoc-before-process-signature`. Perhaps the `autodoc-before-process-signature` hook should look at `__new__` and `__init__` itself, rather than autodoc handling it.

tk0miya:
I posted #7639 to follow the problem of type comments. I believe it's not a blocker of this PR. So merging this now.

PR #7625, opened 2020-05-06
Rewrite unparse to use ast.NodeVisitor

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- This is a refactored version of #7497.
- I sorted the methods of _UnparseVisitor


eric-wieser:
Thanks for picking this up!

PR #7627, opened 2020-05-07
Remove the hacks associated with singledispatch callables

eric-wieser:
This should hopefully be the first step towards gh-3610

### Feature or Bugfix

- Refactoring


tests/test_ext_autodoc.py
```diff
@@ -1563,8 +1563,8 @@ def test_singledispatch(app):
         '',
         '',
         '.. py:function:: func(arg, kwarg=None)',
-        '   func(arg: int, kwarg=None)',
-        '   func(arg: str, kwarg=None)',
+        '                 func(arg: int, kwarg=None)',
+        '                 func(arg: str, kwarg=None)',
         '   :module: target.singledispatch',
```
eric-wieser, line 1568: I don't know enough about rST to know if this whitespace is fine / a good idea. Happy to revert this bit.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1473,12 +1468,8 @@ def format_args(self, **kwargs: Any) -> str:
             args = args.replace('\\', '\\\\')
         return args
 
-    def add_directive_header(self, sig: str) -> None:
-        meth = self.parent.__dict__.get(self.objpath[-1])
-        if inspect.is_singledispatch_method(meth):
```
eric-wieser, line 1478: I don't fully understand what the difference was between:

* `meth`
* `self.object`
* `obj`

Based on the tests still passing, it seems that `meth` and `self.object` were the same?

tests/test_ext_autodoc.py
tk0miya: I did not know reST support such indentation. But it works fine on my local. I prefer this. +1!

sphinx/ext/autodoc/__init__.py
```diff
@@ -1075,43 +1086,27 @@ def format_args(self, **kwargs: Any) -> str:
     def document_members(self, all_members: bool = False) -> None:
         pass
 
-    def add_directive_header(self, sig: str) -> None:
+    def add_directive_header(self, sigs: List[str]) -> None:
```
tk0miya: A documenter class is a public class for extensions. So this is a breaking change. I don't know there are really extensions that override this method. But we need a migration period for such change.

tk0miya (changes requested):

sphinx/ext/autodoc/__init__.py
eric-wieser: Perhaps I'll change the caller to:
```
try:
    self.add_directive_header(sigs=sigs)
except TypeError as e:
    self.add_directive_header(sig=sigs[0])
    emit_deprecation_warning()
```

tk0miya: +1

sphinx/ext/autodoc/__init__.py
```diff
@@ -780,11 +795,15 @@ def generate(self, more_content: Any = None, real_modname: str = None,
...
-        # format the object's signature, if any
-        sig = self.format_signature()
+        # format the object's signatures, if any
+        sigs = self.format_signatures()
 
         # generate the directive header and options, if applicable
-        self.add_directive_header(sig)
+        if len(sigs) == 1:
+            # backwards compatibility for old code
+            self.add_directive_header(sigs)
```
tk0miya: `sigs[0]`?

sphinx/ext/autodoc/__init__.py
```diff
@@ -414,14 +413,30 @@ def format_signature(self, **kwargs: Any) -> str:
         else:
             return ''
 
+    def format_signatures(self, **kwargs):
```
tk0miya: Please add typehints here

sphinx/ext/autodoc/__init__.py
```diff
@@ -840,8 +859,8 @@ def parse_name(self) -> bool:
                            type='autodoc')
         return ret
 
-    def add_directive_header(self, sig: str) -> None:
-        Documenter.add_directive_header(self, sig)
+    def add_directive_header_multi(self, sigs: List[str]) -> None:
+        Documenter.add_directive_header(self, sigs)
```
tk0miya: Note: It would be better to use `super()` instead after this PR.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1075,43 +1094,27 @@ def format_args(self, **kwargs: Any) -> str:
     def document_members(self, all_members: bool = False) -> None:
         pass
 
-    def add_directive_header(self, sig: str) -> None:
```
tk0miya, line 1078: I wonder we should keep this as is because some extension inherits this documenter and call this method.

FYI: zopeext is one of the case that inherits `ClassDocumenter`. But it does not call the `add_directive_header()` directly. So it's not a case I said. It lets us know there are extensions that inherit our documenter classes.
https://github.com/sphinx-contrib/zopeext/blob/master/sphinxcontrib/zopeext/autointerface.py

eric-wieser: I think this is now resolved in my latest revision. Let me know if you see further compatibility issues.

sphinx/ext/autodoc/__init__.py
eric-wieser: Done in this PR.

sphinx/ext/autodoc/__init__.py
eric-wieser: Fixed

sphinx/ext/autodoc/__init__.py
```diff
@@ -780,11 +800,15 @@ def generate(self, more_content: Any = None, real_modname: str = None,
         # reST and no starting newline is present
         self.add_line('', sourcename)
 
-        # format the object's signature, if any
-        sig = self.format_signature()
+        # format the object's signatures, if any
+        sigs = self.format_signatures()
```
tk0miya, line 804: This change means `format_signature()` will be never called from `generate()` method, right? If so, subclasses of FunctionDocumenter or MethodDocumenter should implement `format_signatures()` (plural)  instead of `format_signature()` (singular). I think that is a breaking change.

tk0miya:
Now I'm thinking of the final form of Documenter class. I doubt `add_directive_header_multi` is really a proper API to generate directive headers. IMO, it's great if `add_directive_header` accepts both `str` and `List[str]`. What do you think?

Personally, I'd like to use `inspect.Signature` object to represent signatures instead of `str`. But big bang changes are needed to do that...

eric-wieser:
> IMO, it's great if `add_directive_header` accepts both `str` and `List[str]`. What do you think?

`add_directive_header_multi` is ugly, but it was chosen as a deliberately different name to minimize the chance of causing breakage.

Your suggestion strikes me as more likely to break downstream projects who implement `add_directive_header` and expect to be passed a string. Perhaps that's no worse than what this PR currently does, which is to not attempt to call those projects at all if multiple strings are needed.

It's also awkward because it means any implementer of `add_directive_header` that touches `sigs` has to start with:
```
if isinstance(sig, str):
    sig = [sig]
```

I'll let you make the call.

tk0miya:
>Your suggestion strikes me as more likely to break downstream projects who implement add_directive_header and expect to be passed a string.

You're right. It's my wrong.

eric-wieser:
So just to be clear, our choice is between:

1. `add_directive_header_multi` or some better name. For objects with multiple signatures, like `singledispatch` or in future `@overload`, downstream extensions will not be used at all unless they override the new method instead.
2. `add_directive_header`. For objects with multiple signatures, like `singledispatch` or in future `@overload`, downstream extensions will either:
    * keep working if they did not look at the `sig` member, and just used `super().add_directive_header(sig)` __*__
    * crash, most likely with `TypeError`
    * produce nonsense, because `sig[0]` is now a str not just a single letter

I'm fine with either option. I suspect most users fall into the category marked `*`.

tk0miya:
As a trial, I made another patch to remove singledispatch hacks.
https://github.com/sphinx-doc/sphinx/compare/3.x...tk0miya:refactor_singledispatch?expand=1

It does not add new methods. But it is still tricky a bit. What do you think?

eric-wieser:
Passing a multiline string seems pretty reasonable to me, and is definitely safer than my approach.

Do you want me to update this PR with a similar approach, or do you want to abandon mine and just use your branch?

tk0miya:
I made a PR #7677 from my branch just now. So I'm closing this now.
Thank you for your work. This discussion was much worthy to me. I feel it is very difficult and tough work to keep backward-compatibility of autodoc. So we need to make a brand-new autodoc (autodoc-ng?) to step forward.

Issue #7629, opened 2020-05-07
autofunction raises an error if an integer given


tk0miya:
**Describe the bug**
autofunction raises an error if an integer given

**To Reproduce**

```
.. autofunction:: errno.EINTR
```
```
WARNING: error while formatting arguments for errno.EINTR:
Traceback (most recent call last):
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 1049, in format_args
    sig = inspect.signature(unwrapped)
  File "/Users/tkomiya/work/sphinx/sphinx/util/inspect.py", line 406, in signature
    signature = inspect.signature(subject)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 3093, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2842, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 2216, in _signature_from_callable
    raise TypeError('{!r} is not a callable object'.format(obj))
TypeError: 4 is not a callable object

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 395, in format_signature
    args = self.format_args(**kwargs)
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 1052, in format_args
    if (inspect.is_builtin_class_method(unwrapped, '__new__') and
  File "/Users/tkomiya/work/sphinx/sphinx/util/inspect.py", line 392, in is_builtin_class_method
    classes = [c for c in inspect.getmro(obj) if attr_name in c.__dict__]
  File "/Users/tkomiya/.pyenv/versions/3.8.2/lib/python3.8/inspect.py", line 489, in getmro
    return cls.__mro__
AttributeError: 'int' object has no attribute '__mro__'
```

**Expected behavior**
Shows simple and readable error. (or suppressed?)

**Your project**
No

**Screenshots**
No

**Environment info**
- OS: Mac
- Python version: 3.8.2
- Sphinx version: 3.0.1
- Sphinx extensions: sphinx.ext.autodoc
- Extra tools: No

**Additional context**
[...]

PR #7632, opened 2020-05-08
Do not crash if `is_builtin_class_method` is passed a non-type

eric-wieser:
This partially addresses #7629 (the first of the three points in https://github.com/sphinx-doc/sphinx/issues/7629#issuecomment-625372633)

### Feature or Bugfix

- Bugfix

### Purpose
Avoids chained tracebacks for a user error like `.. autofunction:: errno.EINTR`

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -389,12 +389,23 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
     Why this function needed? CPython implements int.__init__ by Descriptor
     but PyPy implements it by pure Python code.
     """
-    classes = [c for c in inspect.getmro(obj) if attr_name in c.__dict__]
-    cls = classes[0] if classes else object
-
-    if not hasattr(builtins, safe_getattr(cls, '__name__', '')):
+    try:
+        mro = inspect.getmro(obj)
+    except AttributeError:
+        # no __mro__, assume the object has no methods as we know them
         return False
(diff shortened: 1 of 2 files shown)


eric-wieser:
CI failure is unrelated, looks like python nightly changed the behavior of type signatures.

sphinx/util/inspect.py
```diff
@@ -389,12 +389,23 @@ def is_builtin_class_method(obj: Any, attr_name: str) -> bool:
     Why this function needed? CPython implements int.__init__ by Descriptor
     but PyPy implements it by pure Python code.
     """
-    classes = [c for c in inspect.getmro(obj) if attr_name in c.__dict__]
-    cls = classes[0] if classes else object
+    try:
+        mro = inspect.getmro(obj)
```
tk0miya, line 393: off topic: We have many codes that touches `__mro__` directly. Is it better to replace them with `inspect.getmro()`?

tk0miya (approved):
LGTM. I'll merge this after fixing the mainline.

sphinx/util/inspect.py
eric-wieser: No idea. The implementation of `getmro` on my machine is just `.__mro__`.

tk0miya: Okay. I'll keep them as is.

tk0miya:
Merged. Thanks!

PR #7637, opened 2020-05-09
Suppress arguments if all system defined TypeVars on py39

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- It seems the implementation of the typing module has changed since py3.9.0a6+.
- It causes the following error.

  ```
          sig = inspect.signature(f18)
  >       assert stringify_signature(sig) == '(self, arg1: Union[int, Tuple] = 10) -> List[Dict]'
  E       AssertionError: assert '(self, arg1:...Dict[KT, VT]]' == '(self, arg1:...-> List[Dict]'
  E         - (self, arg1: Union[int, Tuple] = 10) -> List[Dict]
  E         + (self, arg1: Union[int, Tuple] = 10) -> List[Dict[KT, VT]]
  E         ?                                                  +++++++ +
  ```
  https://travis-ci.org/github/sphinx-doc/sphinx/jobs/684582074#L1841-L1846
- This fixes them with checking arguments are TypeVars on typing module.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -86,6 +86,7 @@ Bugs fixed
 * #7562: autodoc: a typehint contains spaces is wrongly rendered under
   autodoc_typehints='description' mode
 * #7551: autodoc: failed to import nested class
+* #7637: autodoc: system defined TypeVars are shown in Python 3.9
 * #7551: autosummary: a nested class is indexed as non-nested class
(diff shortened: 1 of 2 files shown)


tk0miya:
I need to merge this into 3.0.x branch...

PR #7653, opened 2020-05-10
Fix multiple directory creation on quickstart script with nested relpath.

mondeja:
Avoid multiple directory creation with quickstart script called with nested relative path.

### Feature or Bugfix
- Bugfix

### Purpose
Call the `sphinx/cmd/quickstart.py` script using next command:
```
python sphinx/cmd/quickstart.py ./foo/bar/../baz --quiet --project hello --author bye
```
You will see that the directories `foo/bar` and `foo/baz` are created, the new project is created inside `foo/baz` and the directory `foo/bar` is empty.

To fix this I've ensured that the project directory path is resolved with [`pathlib.Path.resolve`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) before create it, so the empty directory is not created.

### DIFF
--- a/sphinx/cmd/quickstart.py
+++ b/sphinx/cmd/quickstart.py
@@ -358,7 +358,7 @@ def generate(d: Dict, overwrite: bool = True, silent: bool = False, templatedir:
     d.setdefault('extensions', [])
     d['copyright'] = time.strftime('%Y') + ', ' + d['author']
 
-    d["path"] = pathlib.Path(d['path']).resolve()
+    d["path"] = str(pathlib.Path(d['path']).resolve())
     ensuredir(d['path'])
 
     srcdir = path.join(d['path'], 'source') if d['sep'] else d['path']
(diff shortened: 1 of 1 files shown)


sphinx/cmd/quickstart.py
```diff
@@ -357,6 +358,7 @@ def generate(d: Dict, overwrite: bool = True, silent: bool = False, templatedir:
     d.setdefault('extensions', [])
     d['copyright'] = time.strftime('%Y') + ', ' + d['author']
 
+    d["path"] = str(pathlib.Path(d['path']).resolve())
```
tk0miya: Now sphinx does not use pathlib at all. So it would be better to use `path.abspath()` instead. What do you think?

tk0miya (commented):
LGTM with nits.

sphinx/cmd/quickstart.py
mondeja: Oh yes, a much better solution!

tk0miya:

PR #7655, opened 2020-05-11
Make sphinx.util.typing.stringify render optional unions better

### LINKED ISSUE #7654: Optional[Union[foo, bar]] presented as Union[foo, bar, None]
`sphinx.util.typing.stringify`, when given a type annotation of e.g. `Optional[int]`, will correctly render exactly that. Unfortunately when the annotation is instead e.g. `Optional[Union[int, str]]`, the result is `Union[int, str, None]` which isn't quite good enough.

chipaca:
Without this change, stringify(Optional[Union[int, str]]) returns
'Union[int, str, None]' rather than the expected 'Optional[...]'.

This change fixes that.

fixes: #7654

### DIFF
--- a/sphinx/util/typing.py
+++ b/sphinx/util/typing.py
@@ -91,11 +91,15 @@ def _stringify_py37(annotation: Any) -> str:
 
     if getattr(annotation, '__args__', None):
         if qualname == 'Union':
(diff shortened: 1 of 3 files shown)


tk0miya:
Good point!

tk0miya:

PR #7540, opened 2020-04-23
basic theme: set "clear: both" on "highlight*" divs

mgeier:
This is a refinement of my PR #7484.

For, example, when using `html_theme = 'classic'` and this source text:

```rst
.. sidebar:: Sidebar

    Sidebar content.

::

    code
```

Before:

image

After:

image

PR #7482, opened 2020-04-14
basic theme: CSS spacing for code blocks with captions and line numbers

mgeier:
I'm not sure whether I got all the details right, but I think it's definitely an improvement.

Example document:

```rest
Normal code block:

.. code-block:: python

    print('Hello, wooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooorld!')

With line numbers:

.. code-block:: python
    :linenos:

    print('Hello, wooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooorld!')

With caption:

.. code-block:: python
    :caption: Caption

    print('Hello, wooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooorld!')

With caption and line numbers:

.. code-block:: python
    :caption: Caption
    :linenos:

    print('Hello, wooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooorld!')
```

With `html_theme = 'basic'`, this previously looked like:

image

After this PR, it looks like:

image

PR #7657, opened 2020-05-12
basic theme: CSS margin overhaul

### LINKED ISSUE #7544: classic theme: inconsistent padding in admonitions vs. topic/sidebar
Admonitions have a padding of `7px` on all 4 sides (defined in `classic.css`), while `topic` and `sidebar` have a padding of `7px` only on the left, right and top, not on the bottom (defined in `basic.css`).

Is this intentional?

If yes, we should probably remove the bottom padding from `classic.css`, if no, we should probably add the padding in `basic.css`?

mgeier:
Fixes #7544.

Some details might still be missing, but I think this is at least a step in the right direction.

As mentioned in https://github.com/sphinx-doc/sphinx/issues/7544#issuecomment-624096608:

> I think the "right" solution would be to remove the `margin-bottom` from the `:last-child` and add a consistent `padding: 7px` on all sides.
> The problem is that this would be quite a big change in multiple themes, and people might consider it a "breaking change".

It turned out that I had only to change the `basic` theme, but it's still quite a big change, and some might consider it "breaking".

Any suggestions for improvements?

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -316,7 +316,7 @@ img.align-default, .figure.align-default {
 div.sidebar {
     margin: 0 0 0.5em 1em;
     border: 1px solid #ddb;
-    padding: 7px 7px 0 7px;
+    padding: 7px;
     background-color: #ffe;
     width: 40%;
     float: right;
@@ -336,7 +336,7 @@ div.admonition, div.topic, pre, div[class|="highlight"] {
 
 div.topic {
     border: 1px solid #ccc;
-    padding: 7px 7px 0 7px;
+    padding: 7px;
     margin: 10px 0 10px 0;
     overflow-x: auto;
 }
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -482,6 +482,10 @@ table.field-list td, table.field-list th {
 
 /* -- hlist styles ---------------------------------------------------------- */
 
+table.hlist {
+    margin: 1em 0;
+}
+
 table.hlist td {
     vertical-align: top;
 }
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -426,13 +426,13 @@ table.citation td {
     border-bottom: none;
 }
 
-th > p:first-child,
-td > p:first-child {
+th > :first-child,
+td > :first-child {
     margin-top: 0px;
 }
 
-th > p:last-child,
-td > p:last-child {
+th > :last-child,
+td > :last-child {
     margin-bottom: 0px;
 }
 
@@ -505,11 +505,11 @@ ol.upperroman {
     list-style: upper-roman;
 }
 
-li > p:first-child {
+li > :first-child {
     margin-top: 0px;
 }
 
-li > p:last-child {
+li > :last-child {
     margin-bottom: 0px;
 }
 
@@ -557,7 +557,7 @@ dl {
     margin-bottom: 15px;
 }
 
-dd > p:first-child {
+dd > :first-child {
     margin-top: 0px;
 }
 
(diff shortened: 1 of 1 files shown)


tk0miya:
I'm not an expert in HTML/CSS. So I don't have a comment on this. So let's see what effect to 3rd party themes after merging.

return42:
> You should create a PR with your suggested improvements.

Are you joking? Can you image how much work I have to repair all my projects using up to date sphinx-doc (3.1) .. Did you ever compiled the sphinx-doc project itself? Did you ever had a look on long contents directives? Do you ever compiled one of the big searx projects (like python docs) ..

Or did you just fumble with theses small docutils examples?

You should test elementary patches and not release code and test it by the users, awaiting that they send you patches repairing all you fumbled broken.

If you are not knowing what to do just git-revert the patches.

mgeier:
> > You should create a PR with your suggested improvements.
> 
> Are you joking?

Not at all!
Please make a PR (or multiple)!

> Can you image how much work I have to repair all my projects using up to date sphinx-doc (3.1) .. 

For the short term, it should be sufficient to pin `sphinx<3.1` in your dependencies. So yes, very little work.

Then we can try to fix the issues together, which will need some work done by someone.

Once they are fixed, you can un-pin your dependencies and everything will be honky-dory.

> Did you ever compiled the sphinx-doc project itself?

Yes, I did.

Ironically, the Sphinx docs don't use that many Sphinx features.

> Did you ever had a look on long contents directives?

Yes, I did.

> Do you ever compiled one of the big searx projects (like python docs) ..

No, I've never heard about it.

> Or did you just fumble with theses small docutils examples?

Of course not.

I just used the docutils example in https://github.com/sphinx-doc/sphinx/issues/7838#issuecomment-644190210 to illustrate the visual inconsistencies.

A longer list wouldn't add any useful information.

> You should test elementary patches and not release code and test it by the users, awaiting that they send you patches repairing all you fumbled broken.

Sure, I tested it quite thoroughly, but I can never guarantee that nothing will break.

Also, it is quite hard to get the attention of potential reviewers for every single PR for Sphinx. But it is really easy to get user's attention after things are released (as your comments have shown).

> If you are not knowing what to do just git-revert the patches.

I would say that I know to some degree what I'm doing, but I'm certainly not an expert. I'm very much open to suggestions for improvements!

I think the changes #7657 are a net improvement, so I wouldn't completely revert it.
But there are for sure some things that can still be improved by further PRs.

----

While I'm at it, please let me also respond to your comment https://github.com/sphinx-doc/sphinx/issues/7838#issuecomment-646009605, in order not to disturb the constructive discussions over there:

> @mgeier please stop changing basic CSS !!!

I don't have concrete plans to change further CSS in the `basic` theme, but I can't guarantee that I won't find further problems that I would like to fix.
But for now, I've already done all the changes I wanted to be done:

* #7481
* #7482
* #7484
* #7540
* #7542
* #7543
* #7572
* #7657 (this one)
* #7717

> i'm pretty annoyed because i have been trying to find out how to fix this broken layout in customizing for hours ... and i can do that in all my projects .. thanks for all the work .. 

As mentioned above, ad-hoc work-arounds are probably not the best use of your time.

> how do you get the idea to change the basic CSS so massively without having to test it broadly.

Good question!

I just saw that there are problems with the `basic` CSS and tried to fix them.
I was fully aware that people might complain, but I was also aware that whatever they complain about will also be fixable. And in the end, we all will have a better Sphinx.

----

... and https://github.com/sphinx-doc/sphinx/issues/7838#issuecomment-646078457:

> @mgeier wrotes ..
> 
> > I would like to argue that the new spacing is more consistent,
> 
> Not with the toc directive which is rendered much more _compact_.

If everything is compact, it's still consistent.
I'm not arguing that compact or non-compact is better, as long as it is consistent.

And I think the discussions in #7838 look very promising. I think it is realistic that we will get compact lists (for you) that are still consistent (for me).

mgeier:
@return42 Please check out #7852, which should fix the list spacing issues.

PR #7646, opened 2020-05-09
Fix #7629: autodoc: autofunction emits an unfriendly warning

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7629 
- This also fixes the same problem of automethod.
- I did not resolve a problem for location of the warnings. I'll work on it as another issue.

PR #7658, opened 2020-05-13
Fix: handle errors on event handlers (refs: #7629)

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: https://github.com/sphinx-doc/sphinx/pull/7646#discussion_r424512792

### DIFF
--- a/sphinx/events.py
+++ b/sphinx/events.py
@@ -16,7 +16,7 @@ from operator import attrgetter
 from typing import Any, Callable, Dict, List, NamedTuple
 
 from sphinx.deprecation import RemovedInSphinx40Warning
-from sphinx.errors import ExtensionError
+from sphinx.errors import ExtensionError, SphinxError
 from sphinx.locale import __
 from sphinx.util import logging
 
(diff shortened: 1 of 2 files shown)


sphinx/events.py
```diff
@@ -100,11 +100,16 @@ def emit(self, name: str, *args: Any) -> List:
...
+            try:
+                if self.app is None:
+                    # for compatibility; RemovedInSphinx40Warning
+                    results.append(listener.handler(*args))
+                else:
+                    results.append(listener.handler(self.app, *args))
+            except SphinxError:
+                raise
+            except Exception as exc:
+                raise ExtensionError(exc)
```
eric-wieser:
```suggestion
                raise ExtensionError('Handler {!r} for event {!r} threw an exception'.format(listener, event)) from exc
```
That way the chance is slightly higher that the user can recover the original stack trace.

tk0miya: Thanks!

eric-wieser (approved):
Good idea

tk0miya:

sphinx/events.py
eric-wieser: Any reason for using `%` formatting instead of `.format`? Just so I can pick the right one for future contributions.

tk0miya: Personally, I prefer it to `.format`. I think both are right.

PR #7663, opened 2020-05-14
Fix astext() for two Sphinx nodes

jakobandersen:
### Feature or Bugfix
- Feature
- Bugfix

### Purpose
Make the Docutils ``astext()`` method return a more reasonable representation of multi-line declarations and of parameter lists.


jakobandersen:
This is nice for making better testing of the generated docutils. The ``astext()`` method is used other places in Sphinx, but are there any documented guarantees that this PR breaks?

tk0miya (approved):
+1: good change. AFAIK, `astext()` is less used in the main code. So this change does not effect to output.

jakobandersen:
Great, thanks!

Issue #7660, opened 2020-05-14
Approch to select tree of inheritance on inheritance_diagram


vallsv:
Hi,

I was playing around inheritance_diagram with the project i am working on (https://github.com/silx-kit/silx/pull/3053). Diagram is very useful so thanks for this project. I would like to know what you think about few points:

1) If there is a link between `A` and `C` between a private class `_B` (and the private flag in not set), the extension do not create a link between `A` and `C`. I think it would be good to draw `A -> C`. If you prefer it could be displayed with a dotted arrow, or with an intermediate anonymous class `A -> [] -> C`. What do you think?  

2) When we specify a top classes, i expect to have only the inheritances (top-down) from that classes. But this extension also generate the whole inheritance of classes inheriting from the top class. I think it is noise, it is not the place to display this relationship.

Example
```
class A: pass
class B (A, Otherthing, Otherthing2): pass
```
With top class = `A`, the extension will also show `Otherthing`, `Otherthing2`. 

3) BTW if i specify a top-class, classes inherited by this class with also be displayed. I don't think it is what you want.

4) A last small thing: It is based on syntactic class name. 2 classes with the same name will be merged together. Is it using python classes internally? Or is it based on text parsing?

To summarize I care about  2 use case for a documentation.

- In a specific class, you want to see how it is implemented, then you want to see the tree of classes inherited (bottom-up)
- In a package, you want to see all the implementations from a class (top-down). (use of `top-classes`?)

But it looks like this extension mix both, and i feel like the result will always be difficult to read, cause you do not ail the same thing. Is there a way to do one or the other of this approaches? Do you think it could be implemented?

vallsv:
Looking at the test code, there is obvious inconsistencies like in `test_diagram_w_1_top_class` (A->B is not shown, while A and B are shown). And in `test_diagram_module_w_2_top_classes` there is comment explaining it is not supposed to work this way.

So i think the initial intention, at least for `top-classes` was what i have proposed. Which is nice :-) Cause it is not a big deal to fix.

I am working on a PR.

Issue #5518, opened 2018-10-09
Inheritance diagram can't import classes within a class


KM4YRI:
Apologies if this is a duplicated issue, I searched but nothing turned up.

### Problem

The inheritance diagram plugin fails to import classes within a class within a package/module.

The root-cause of the problem is the function `try_import()` within `inheritance_diagram.py`.  The current implementation tries to import the full object name, and if that fails, it attempts to strip off the class/final module name, and import what was before that.  This works for cases like `foo_package.foo_subpackage.FooClass`, but not `foo_package.FooClass.BarClass`.

#### Procedure to reproduce the problem

1. Create a new python package called `foo_package`, add `__init__.py`, `doc` folder, etc..
2. Create a new sphinx project for `foo_package` under `/doc/`
2. In `conf.py`, set it up to point to the new package, and add `sphinx.ext.inheritance_diagram` to the `extensions` list.
3. Add the code below into `foo_package` and `index.rst`:

`__init__.py` within `foo_package`:
```
class Foo:
    class Bar:
        pass
    class Baz(Bar):
        pass

```

`index.rst` for `foo_package`:
```
.. inheritance-diagram::
    foo_package.Foo.Bar
    foo_package.Foo.Baz
```

#### Error logs / results
```
index.rst:16: WARNING: Could not import class or module 'foo_package.Foo.Bar' specified for inheritance diagram
```

The function `try_import()` is splitting the object name into `foo_package.Foo` and trying to import that.  Which doesn't work because it's a class.

#### Expected results

The inheritance diagram plugin should be able to recognize foo_package.Foo.Bar as a class 'Bar' inside class `Foo`, inside module `foo_package`.  

### Environment info
- OS: Windows 7
- Python version: 3.5.1
- Sphinx version: 1.8.1 (with a fix applied that prevented the released version from running with 3.5.1)

### Proposed solution

Update the function `try_import()` within `inheritance_diagram.py` to correctly split the full object name at the correct point so that it contains the correct module name(s) and class/classes.  

This could be done a few different ways, but I don't think a regex like the one used on line 63-65 is enough to work by itself.  The one that comes to my head is to iteratively pop off attribute names off of the string until it is importable, and then return the result of `getattr(sys.modules.get(modname), attrname, None)`.

KM4YRI:
Oh, almost forgot, I'm willing to write a fix and submit a pull request if it's verified to be a legitimate bug.

[...]

Issue #5523, opened 2018-10-11
Inheritance diagram doesn't work with packages


themightyoarfish:
### Problem
The inability of Sphinx to work with (sub)packages comes up occasionally (see e.g. https://stackoverflow.com/questions/51881692/how-can-i-use-sphinx-with-subpackages-without-duplicating-everything), but I have met this issue again when using the `inheritance-diagrams` extension.

I have this package structure
```
package/
├── __init__.py
└── subpackage
    ├── __init__.py
    └── module.py
```
The files are given here:

**package/__init__.py**
```
from . import subpackage
```
**package.subpackage/__init__.py**
```
from .module import Class
__all__ = ['Class']
```
**package.subpackage/module.py**
```
class Class(object):
    pass
```

So in summary, I'm using subpackages so I can write `package.subpackage.Class` instead of `package.subpackage.module.Class`.

My sphinx configuration looks like this:
**Makefile**
```
SPHINXOPTS    = -W -n
SPHINXBUILD   = sphinx-build
SPHINXPROJ    = test
SOURCEDIR     = source
BUILDDIR      = build

# Put it first so that "make" without argument is like "make help".
help:
        @$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)

.PHONY: help Makefile

%: Makefile
        @$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```

**conf.py**
```
import os
import sys
sys.path.insert(0, os.path.abspath('.'))
sys.path.insert(0, os.path.abspath('../../'))
import sphinx_bootstrap_theme

project = 'test'
copyright = ''
author = ''

version = ''
release = '0'

extensions = [
    'sphinx.ext.autodoc',
    'sphinx.ext.intersphinx',
    'sphinx.ext.mathjax',
    'sphinx.ext.githubpages',
    'sphinx.ext.napoleon',
    'sphinxarg.ext',
    'sphinx.ext.viewcode',
    'sphinx.ext.inheritance_diagram',
]

source_suffix = '.rst'
master_doc = 'index'
language = None
exclude_patterns = []

pygments_style = 'sphinx'
highlight_language = 'python3'

html_theme = 'bootstrap'
html_theme_path = sphinx_bootstrap_theme.get_html_theme_path()

html_theme_options = {
[...]

PR #7665, opened 2020-05-15
UML inheritance diagram

### LINKED ISSUE #2969: inheritance_diagram: allow inverse arrow direction
Currently arrows go from parent class to base class; this is usually the other way around and there should be an option to allow this.

vallsv:
Hi. Here is a first attempt to allow to display inheritance diagram with UML-like notation.

### Feature
- Feature
- Refactoring

### Purpose
The aim was to allow to use UML-like diagram, without leaning graphvis language.

### Detail
- A `:style:` was added
   - It supports `default` for the current look
   - And `uml` which generates edges in the reversed direction and with the right arrow symbol (as specified in the issue) 
- Unittests was refactored in order to:
   - Split a big test (as it was called in a comment in the source code)
   - If you have an example with a module-scope fixture to build stuff, it could be much cleaner (i tried, i failed)
   - Allow to introspect the dot syntax

### Example
Picked from the unittests
```
class Foo(object): pass
class Bar(Foo): pass
class Baz(Bar): pass
```
On side note, this unittest uses a special directive to alias the `Foo` class, but nothing related to this PR.

#### With `:style: uml`
Picked from the unittest.
inheritance-f707c1c85be7bf38531ec0591e8842b3856a1d4f

#### Without `:style: uml` (the default)

inheritance-c4c48c632b4eeb835502bead236b4432c35d88e1

### Relates
- Closes #2969

### DIFF
--- a/doc/usage/extensions/inheritance.rst
+++ b/doc/usage/extensions/inheritance.rst
@@ -101,6 +101,10 @@ It adds this directive:
    It supports two different values: ``default``, which generates a diagram
    with normal arrows; and ``uml`` which generates UML-like arrows.
 
+   The direction of the diagram can be specified with the directive
+   ``direction``. It can be one of the values: ``left-right`` (the default),
+   or ``top-bottom``.
+
 
 Examples
 --------
@@ -134,6 +138,14 @@ part (``sphinx``) from all names::
 .. inheritance-diagram:: sphinx.ext.inheritance_diagram.InheritanceDiagram
    :top-classes: sphinx.util.docutils.SphinxDirective
    :parts: -1
+
+Using a diagram with vertical orientation::
+
+   .. inheritance-diagram:: sphinx.ext.inheritance_diagram.InheritanceDiagram
+      :direction: top-bottom
+
+.. inheritance-diagram:: sphinx.ext.inheritance_diagram.InheritanceDiagram
+   :direction: top-bottom
 
 Using UML-like diagram::
 
--- a/doc/usage/extensions/inheritance.rst
+++ b/doc/usage/extensions/inheritance.rst
@@ -97,6 +97,10 @@ It adds this directive:
    .. versionchanged:: 1.7
       Added ``top-classes`` option to limit the scope of inheritance graphs.
 
+   One of the predefined style can be selected with the directive ``style``.
+   It supports two different values: ``default``, which generates a diagram
+   with normal arrows; and ``uml`` which generates UML-like arrows.
+
 
 Examples
 --------
@@ -131,6 +135,13 @@ part (``sphinx``) from all names::
    :top-classes: sphinx.util.docutils.SphinxDirective
    :parts: -1
 
+Using UML-like diagram::
+
+   .. inheritance-diagram:: sphinx.ext.inheritance_diagram.InheritanceDiagram
+      :style: uml
+
+.. inheritance-diagram:: sphinx.ext.inheritance_diagram.InheritanceDiagram
+   :style: uml
 
 
 Configuration
--- a/doc/usage/extensions/inheritance.rst
+++ b/doc/usage/extensions/inheritance.rst
@@ -16,11 +16,11 @@ It adds this directive:
 .. rst:directive:: inheritance-diagram
 
    This directive has one or more arguments, each giving a module or class
-   name.  Class names can be unqualified; in that case they are taken to exist
+   name. Class names can be unqualified; in that case they are taken to exist
    in the currently described module (see :rst:dir:`py:module`).
 
    For each given class, and each class in each given module, the base classes
-   are determined.  Then, from all classes and their base classes, a graph is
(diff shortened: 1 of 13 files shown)


vallsv:
I am working on something bigger than that. There is a huge clean up to do. How can i contribute? Anybody care if i extend this PR?

eric-wieser:
Can you update the PR description to show what that inheritance graph looks like _without_ this PR too, so we can see before / after?

vallsv:
Yes sure, but that's an optional directive, so if you do not specify `:style: uml` to your diagram, the result is the one without this PR.

eric-wieser:
Sure, I'm just asking for a comparison of the two styles in the description.

vallsv:
I have updated the description as requested.

As reference, here is the current state on my side: https://github.com/vallsv/sphinx/tree/improve-inheritance-diagram (other things than UML), with that i can do what i want, so i do not plan another changes.

doc/usage/extensions/inheritance.rst
```diff
@@ -97,6 +97,10 @@ It adds this directive:
    .. versionchanged:: 1.7
       Added ``top-classes`` option to limit the scope of inheritance graphs.
 
+   One of the predefined style can be selected with the directive ``style``.
+   It supports two different values: ``default``, which generates a diagram
+   with normal arrows; and ``uml`` which generates UML-like arrows.
+
```
tk0miya, line 91: Please add `versionchanged` tag here? I think this would be added into 3.3.0 release.

sphinx/ext/inheritance_diagram.py
```diff
@@ -137,14 +137,15 @@ class InheritanceGraph:
     graphviz dot graph from them.
     """
     def __init__(self, class_names: List[str], currmodule: str, show_builtins: bool = False,
-                 private_bases: bool = False, parts: int = 0, aliases: Dict[str, str] = None,
-                 top_classes: List[Any] = []) -> None:
+                 private_bases: bool = False, parts: int = 0, style: str = "default",
```
tk0miya: Please append a new parameter to the tail of the argument list to keep compatibility.

sphinx/ext/inheritance_diagram.py
```diff
@@ -334,6 +345,7 @@ class InheritanceDiagram(SphinxDirective):
     final_argument_whitespace = True
     option_spec = {
         'parts': int,
+        'style': str,
```
tk0miya: How about using `docutils.parsers.rst.directive:choice` instead? Then you don't need to validate the value manually.

tk0miya (changes requested):
LGTM with nits.

sphinx/ext/inheritance_diagram.py
vallsv: How does it work?

doc/usage/extensions/inheritance.rst
vallsv: I have updated it with `versionadded` trying to follow the previous use.

vallsv:
I have updated the PR, but it contains a little more than what there was. It mostly fixes behaviors on the gtaph generation and not only add `style`.

If you prefer i can remove that extra stuffs for now.

sphinx/ext/inheritance_diagram.py
vallsv: Done

vallsv:
mypy CI is making me crazy, is there an easy way to test it on my side?

sphinx/ext/inheritance_diagram.py
vallsv: I found an example in `graphviz.py` ext

vallsv:
I have reworked the commits. This can be reviewed.

sphinx/ext/inheritance_diagram.py
```diff
@@ -136,19 +140,60 @@ class InheritanceGraph:
     from all the way to the root "object", and then is able to generate a
     graphviz dot graph from them.
     """
+
+    class Node:
+        """Hold a class and it's relations with the other classes"""
+        def __init__(self, cls):
+            self.cls = cls
+            self.based = set([])  # type: Set[Any]
+            self.derived = set([])  # type: Set[Any]
```
eric-wieser: nit:
```suggestion
            self.based = set()  # type: Set[Any]
            self.derived = set()  # type: Set[Any]
```

vallsv: Thanks. No idea why i use to write it this way...

vallsv:
I think this fixes #5518 and #5523

vallsv:
I have pushed the last things i had on my side when i was working on. So it also closes #7660.

vallsv:
@tk0miya it means you have merged the stuff on your own? Or you think it is not mergeable?

eric-wieser:
@vallsv: The 3.x branch has been deleted, which results in all PRs to it being closed automatically.

PR #7669, opened 2020-05-16
Fix #7668: autodoc: wrong retann value is passed to autodoc-proccess-signature

### LINKED ISSUE #7668: wrong retann value is passed to a handler of autodoc-proccess-signature
**Describe the bug**
wrong retann value (a.k.a return_annotation) is passed to a handler of autodoc-proccess-signature.

**To Reproduce**
```
# example.py
def foo(x: int, y: str, z: float) -> None:
    return x + y + z
```
```
# index.rst
.. autofunction:: example.foo
```
```
# conf.py
def on_autodoc_process_signature(app, objtype, fullname, object, options, args, retann):
    print('objtype:', repr(objtype))
    print('fullname:', repr(fullname))
    print('object:', repr(object))
    print('options:', repr(options))
    print('args:', repr(args))
    print('retann:', repr(retann))


def setup(app):
    app.connect('autodoc-process-signature', on_autodoc_process_signature)
```
```
# output
objtype: 'function'
fullname: 'example.foo'
object: <function foo at 0x110c60280>
options: {}
args: '(x: int, y: str, z: float) -> None'
retann: None
```

**Expected behavior**
retann value is passed correctly as an argument (instead of the part of args).

**Your project**
No

**Screenshots**
No

**Environment info**
- OS: Mac
- Python version: 3.8.2
- Sphinx version: 3.1.0dev
- Extra tools: No

**Additional context**
No

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7668


tests/test_ext_autodoc.py
```diff
@@ -161,7 +161,6 @@ def g(a='\n'):
         pass
     assert formatsig('function', 'f', f, None, None) == '(a, b, c=1, **d)'
     assert formatsig('function', 'f', f, 'a, b, c, d', None) == '(a, b, c, d)'
-    assert formatsig('function', 'f', f, None, 'None') == '(a, b, c=1, **d) -> None'
```
tk0miya, line 165: There are no way to pass retann value directly without args in actual usecase. So I removed this.

sphinx/ext/autodoc/__init__.py
```diff
@@ -389,21 +389,27 @@ def format_signature(self, **kwargs: Any) -> str:
...
             try:
+                retann = None
                 try:
                     args = self.format_args(**kwargs)
                 except TypeError:
                     # retry without arguments for old documenters
                     args = self.format_args()
+
+                if args:
+                    matched = re.match(r'^(\(.*\))\s+->\s+(.*)$', args)
```
eric-wieser, line 415: Won't this fail for a signature like `def evil(text=") -> bad") -> None`?

tk0miya: Yes, it would be confused by string. In your example, it works fine because this regexp matches greedily:
```python
>>> sig = '(text=") -> bad") -> None'
>>> re.match(r'^(\(.*\))\s+->\s+(.*)$', sig)
<re.Match object; span=(0, 25), match='(text=") -> bad") -> None'>
>>> _.groups()
('(text=") -> bad")', 'None')
```

But return annotation containing closing-parens and arrow can cheat it.
```python
>>> sig = '(text) -> ") -> None"'
>>> re.match(r'^(\(.*\))\s+->\s+(.*)$', sig)
<re.Match object; span=(0, 21), match='(text) -> ") -> None"'>
>>> _.groups()
('(text) -> ")', 'None"')
```

But I think this will not cause a real problem.

Issue #5270, opened 2018-08-05
autodoc: Keep preserved order of class attributes


tk0miya:
### Problem
- Since python 3.6, class definition order is preserved (see [PEP-520](https://www.python.org/dev/peps/pep-0520/)).
- autodoc should use it by default (instead of alphabet order)
- refs: #4706

Issue #3673, opened 2017-04-26
`autodoc_member_order = 'bysource'` does not work for modules containing __all__


carmenbianca:
Subject: `autodoc_member_order = 'bysource'` does not work for modules containing `__all__`

### Problem
When setting `autodoc_member_order` to `'bysource'` in conf.py, it does not work entirely as expected on modules that contain `__all__`. The members are listed in the order dictated by `__all__`, which is expected. But the members of the members are listed alphabetically.

#### Procedure to reproduce the problem
```
- I have a module `foo` in `foo/__init__.py`
- `__init__.py` itself contains no code other than import
  statements from private modules
  (e.g., `from ._bar import spam, eggs`)
- `__init__.py` has a line that says `__all__ = ['spam', 'eggs']`
- `spam`, in this scenario, is a class that contains two
  methods in this order: `b` and `a`.
- Apply Sphinx magic.
- Module foo in Sphinx is listed in this order:
  - spam
    - a
    - b
  - eggs

```

#### Error logs / results
```
N/A
```
- https://i.imgur.com/jzz6VF6.png

#### Expected results
b and a should be swapped.

### Reproducible project / your project
- https://github.com/carmenbianca/foo-sphinx-bug
  - just `make html` in the docs directory.

### Environment info
- OS: Debian Stretch
- Python version: 3.5.3
- Sphinx version: 1.5.3

the-holger:
I have _almost_ the same issue:
- I'm working on a module
- `__init__.py` only contains Docstring and imports
- `__init__.py` **does not** have an `__all__` assignment
- I have the line `autodoc_member_order = 'bysource'` at the end of my conf.py
- Class functions are ordered alphabetically...

### Environment info

OS: macOS 10.12.6
Python version: 3.6.1
Sphinx version: 1.6.2

MarvBob:
I'm working on a package, there is nothing in `__init__.py`.
When I add the line `autodoc_member_order = 'bysource'` in my conf.py, in the documentation some modules are still ordered alphabetically. I can't find why some of them are well ordered by source and some of them aren't. It seems that the problem occurs more when a module contains class declaration.

### Environment info
- OS: Windows 10
- Python version: 2.7.13
- Sphinx version: 1.6.3

peterjc:
[...]

PR #7676, opened 2020-05-16
Respect member-order="bysource" even when no analyzer is present

eric-wieser:
This takes advantage of the fact that in Python 3.6 and up, the __dict__ attribute of classes and modules is insertion-ordered, which is usually a good approximation of "source order".

Notably, this means that `..automodule:: some_c_extension_module` will document members in the order they were added to the module dict, which is likely what the user wants.

Motivated by the `member-order` being ignored at https://docs.cocotb.org/en/latest/library_reference.html#the-cocotb-simulator-module

### Feature or Bugfix

- Feature

Help would be appreciated testing this (once I have it passing existing tests) - I don't know how to force an analyzer to be absent other than via a C extension module, and I don't really want to have sphinx create one of those just for testing!

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -521,12 +521,12 @@ class Documenter:
                 else:
                     logger.warning(__('missing attribute %s in object %s') %
                                    (name, self.fullname), type='autodoc')
-            return False, sorted(selected)
+            return False, selected
         elif self.options.inherited_members:
-            return False, sorted((m.name, m.value) for m in members.values())
+            return False, [(m.name, m.value) for m in members.values()]
         else:
-            return False, sorted((m.name, m.value) for m in members.values()
-                                 if m.directly_defined)
+            return False, [(m.name, m.value) for m in members.values()
+                           if m.directly_defined]
 
     def filter_members(self, members: List[Tuple[str, Any]], want_all: bool
                        ) -> List[Tuple[str, Any, bool]]:
@@ -699,17 +699,26 @@ class Documenter:
         member_order = self.options.member_order or \
             self.env.config.autodoc_member_order
         if member_order == 'groupwise':
-            # sort by group; relies on stable sort to keep items in the
-            # same group sorted alphabetically
-            memberdocumenters.sort(key=lambda e: e[0].member_order)
-        elif member_order == 'bysource' and self.analyzer:
-            # sort by source order, by virtue of the module analyzer
-            tagorder = self.analyzer.tagorder
+            # sort by group; alphabetically within groups
+            memberdocumenters.sort(key=lambda e: (e[0].member_order, e[0].name))
+        elif member_order == 'bysource':
+            if self.analyzer:
+                # sort by source order, by virtue of the module analyzer
+                tagorder = self.analyzer.tagorder
 
-            def keyfunc(entry: Tuple[Documenter, bool]) -> int:
-                fullname = entry[0].name.split('::')[1]
-                return tagorder.get(fullname, len(tagorder))
-            memberdocumenters.sort(key=keyfunc)
+                def keyfunc(entry: Tuple[Documenter, bool]) -> int:
+                    fullname = entry[0].name.split('::')[1]
+                    return tagorder.get(fullname, len(tagorder))
+                memberdocumenters.sort(key=keyfunc)
+            else:
+                # Assume that member discovery order matches source order.
+                # This is a reasonable assumption in Python 3.6 and up, where
+                # module.__dict__ is insertion-ordered.
+                pass
+        elif member_order == 'alphabetical':
+            memberdocumenters.sort(key=lambda e: e[0].name)
+        else:
+            raise ValueError("Illegal member order {}".format(member_order))
 
         for documenter, isattr in memberdocumenters:
             documenter.generate(


sphinx/ext/autodoc/__init__.py
```diff
@@ -693,17 +693,26 @@ def document_members(self, all_members: bool = False) -> None:
...
+                def keyfunc(entry: Tuple[Documenter, bool]) -> int:
+                    fullname = entry[0].name.split('::')[1]
+                    return tagorder.get(fullname, len(tagorder))
+                memberdocumenters.sort(key=keyfunc)
+            else:
+                # Assume that member discovery order matches source order.
+                # This is a reasonable assumption in Python 3.6 and up, where
+                # module.__dict__ is insertion-ordered.
+                pass
+        elif member_order == 'alphabetic':
```
tk0miya: I noticed our doc says the correct value is `'alphabetical'`. And I feel it is better wording.
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_member_order

sphinx/ext/autodoc/__init__.py
```diff
@@ -693,17 +693,26 @@ def document_members(self, all_members: bool = False) -> None:
...
+            memberdocumenters.sort(key=lambda e: e[0].name)
+        else:
+            raise ValueError("Illegal member order {}".format(member_order))
```
tk0miya, line 721: I'll post another PR to warn such wrong options. So you don't need to handle it here.

sphinx/ext/autodoc/__init__.py
```diff
@@ -693,17 +693,26 @@ def document_members(self, all_members: bool = False) -> None:
...
+
+                def keyfunc(entry: Tuple[Documenter, bool]) -> int:
+                    fullname = entry[0].name.split('::')[1]
+                    return tagorder.get(fullname, len(tagorder))
+                memberdocumenters.sort(key=keyfunc)
+            else:
+                # Assume that member discovery order matches source order.
+                # This is a reasonable assumption in Python 3.6 and up, where
+                # module.__dict__ is insertion-ordered.
+                pass
```
tk0miya, line 717: We need to consider the order of the members not included in `__dict__`.

* enum members (they are not in `__dict__`)
* `__slots__`
* `__annotations__` (type annotated variables, but not initialized yet)
* instance variables
* members defined at ancestor classes
* `module.__all__` (which is strong `__dict__` or `__all__`?)

Note: Now we don't use `__dict__` for discovery. So we have to order them manually if we adopt `__dict__`.

tk0miya (commented):
Thank you for proposal. This is my headache task for next release. Good for discussing.

sphinx/ext/autodoc/__init__.py
eric-wieser: That may be, but all the tests use this spelling

sphinx/ext/autodoc/__init__.py
eric-wieser: Bear in mind this is just a backup option for when no analyzer is available. You're right that this is not a perfect ordering, but sorting alphabetically when "by source" is requested is far worse than a rough approximation to by source.

---

> enum members (they are not in __dict__)

I disagree:
```
>>> C = enum.Enum("Color", "red green blue")
>>> C.__dict__\
mappingproxy({'_generate_next_value_': <function enum.Enum._generate_next_value_(name, start, count, last_values)>,
              '__doc__': 'An enumeration.',
              '__module__': '__main__',
              '_member_names_': ['red', 'green', 'blue'],
              '_member_map_': {'red': <Color.red: 1>,
               'green': <Color.green: 2>,
               'blue': <Color.blue: 3>},
              '_member_type_': object,
              '_value2member_map_': {1: <Color.red: 1>,
               2: <Color.green: 2>,
               3: <Color.blue: 3>},
              'red': <Color.red: 1>,
              'green': <Color.green: 2>,
              'blue': <Color.blue: 3>,
              '__new__': <function enum.Enum.__new__(cls, value)>})
```

> `__slots__`

Also not true:
```
>>> class Foo:
...     __slots__ = ('a', 'b')

>>> Foo.__dict__
mappingproxy({'__module__': '__main__',
              '__slots__': ('a', 'b'),
              'a': <member 'a' of 'Foo' objects>,
              'b': <member 'b' of 'Foo' objects>,
              '__doc__': None})
```

> `__annotations__` (type annotated variables, but not initialized yet)

True, but likely irrelevant for C extension modules

> instance variables

True

> members defined at ancestor classes

True

> `module.__all__`

Likely irrelevant for C extension modules

sphinx/ext/autodoc/__init__.py
eric-wieser: Will that PR convert wrong options into right ones?

sphinx/ext/autodoc/__init__.py
tk0miya: I posted https://github.com/sphinx-doc/sphinx/pull/7679 to migrate the old name to new one with a warning. I suppose it helps to migrate to new options.

sphinx/ext/autodoc/__init__.py
tk0miya: No, it only emits a warning.

eric-wieser:
Can you reopen this?

tk0miya:
sorry. it's mistake.

sphinx/ext/autodoc/__init__.py
eric-wieser: So what should I do if `member_order == "nonsense"` here?

tk0miya: Now it is treated as "alphabetical" (default). I think no reason to change it from now. But I'm also okay if we change the behavior to "not sorted".

If we choose to raise an error, it is better to do that in more earlier stage (ex. config-inited).

sphinx/ext/autodoc/__init__.py
tk0miya: >enum members (they are not in dict)

Sorry, this is my wrong. I mistook `dir()` for `__dict__`. The result of `dir()` for enum classes does not contain methods. But it is not related to sorting. Please forget.
```
Python 3.8.3 (default, May 15 2020, 11:21:44)
[Clang 11.0.3 (clang-1103.0.32.59)] on darwin
>>> import enum
>>> class MyEnum(enum.Enum):
...     foo = 1
...     bar = 2
...     def hello(self): pass
...
>>> dir(MyEnum)
['__class__', '__doc__', '__members__', '__module__', 'bar', 'foo']
```

> __slots__

This is also my wrong.

tk0miya: Note: We have some issues for sorting members: https://github.com/sphinx-doc/sphinx/issues/5270, https://github.com/sphinx-doc/sphinx/issues/3673, https://github.com/sphinx-doc/sphinx/pull/7633, https://github.com/sphinx-doc/sphinx/pull/7177 and so on.
So I'm thinking about the whole of sorting system for autodoc.

But they are not a blocker of this improvement. So no reason to merge this. I'll review this again later (maybe tonight).

sphinx/ext/autodoc/__init__.py
eric-wieser:
```suggestion
        elif member_order == 'alphabetical':
```

sphinx/ext/autodoc/__init__.py
eric-wieser: Thanks for those links, I agree that this is not an end-all solution, but I think removing the initial alphabetic sort is probably a good first step.

I've rebased this PR after your `alphabetical` spelling fix.

tk0miya:
Merged. Thank you always!

PR #7677, opened 2020-05-16
refactor: autodoc: Remove magic mock from singledispatch processing

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- refs: #7627

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -16,7 +16,6 @@ import warnings
 from inspect import Parameter
 from types import ModuleType
 from typing import Any, Callable, Dict, Iterator, List, Sequence, Set, Tuple, Type, Union
-from unittest.mock import patch
 
 from docutils.statemachine import StringList
 
@@ -421,8 +420,15 @@ class Documenter:
         directive = getattr(self, 'directivetype', self.objtype)
         name = self.format_name()
         sourcename = self.get_sourcename()
-        self.add_line('.. %s:%s:: %s%s' % (domain, directive, name, sig),
-                      sourcename)
+
+        # one signature per line, indented by column
+        prefix = '.. %s:%s:: ' % (domain, directive)
+        for i, sig_line in enumerate(sig.split("\n")):
+            self.add_line('%s%s%s' % (prefix, name, sig_line),
+                          sourcename)
+            if i == 0:
+                prefix = " " * len(prefix)
+
         if self.options.noindex:
             self.add_line('   :noindex:', sourcename)
         if self.objpath:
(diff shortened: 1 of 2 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -1075,41 +1076,27 @@ def document_members(self, all_members: bool = False) -> None:
...
+                    self.annotate_to_first_argument(func, typ)
 
-                documenter = FunctionDocumenter(self.directive, '')
-                documenter.object = func
-                self.add_line('   %s%s' % (self.format_name(),
-                                           documenter.format_signature()),
-                              sourcename)
+                    documenter = FunctionDocumenter(self.directive, '')
+                    documenter.object = func
+                    sig += "\n   %s%s" % (self.format_name(), documenter.format_signature())
```
eric-wieser: Might be clearer, and certainly faster, to build the list like I did, and do `'\n'.join` at the end

sphinx/ext/autodoc/__init__.py
```diff
@@ -269,12 +269,13 @@ def documenters(self) -> Dict[str, "Type[Documenter]"]:
         """Returns registered Documenter classes"""
         return self.env.app.registry.documenters
 
-    def add_line(self, line: str, source: str, *lineno: int) -> None:
+    def add_line(self, text: str, source: str, *lineno: int) -> None:
```
eric-wieser: I don't think this change makes sense, given the function name. Perhaps add an extra `add_text` function that loops over each line and calls add_line?

sphinx/ext/autodoc/__init__.py
tk0miya: I suppose the number of singledispatched sub-functions is a few or very few. So both ways are almost same performance. But I agree using a list is more clear. +1

sphinx/ext/autodoc/__init__.py
tk0miya: I have no opinion. So +0 for adding `add_text()`.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1075,41 +1080,28 @@ def document_members(self, all_members: bool = False) -> None:
...
-                                           documenter.format_signature()),
-                              sourcename)
+                    documenter = FunctionDocumenter(self.directive, '')
+                    documenter.object = func
+                    sigs.append("%s%s" % (self.format_name(), documenter.format_signature()))
 
-        # output remains of directive header
-        for call in add_line.call_args_list[1:]:
-            self.add_line(*call[0])
+        return "\n   ".join(sigs)
```
eric-wieser: Adding the indent here feels a bit out of place - perhaps this would be better as I had it in #7627?

sphinx/ext/autodoc/__init__.py
```diff
@@ -421,8 +425,9 @@ def add_directive_header(self, sig: str) -> None:
         directive = getattr(self, 'directivetype', self.objtype)
         name = self.format_name()
         sourcename = self.get_sourcename()
-        self.add_line('.. %s:%s:: %s%s' % (domain, directive, name, sig),
-                      sourcename)
+
+        # Note: use add_text() method to support multi-lined signatures
+        self.add_text('.. %s:%s:: %s%s' % (domain, directive, name, sig), sourcename)
```
eric-wieser: Continuing from my comment above, I think it would be clearer if `add_directive_header` dealt with the necessary indentation for multiline signatures. `format_signature()` can then omit all indentation.
```suggestion
        # one signature per line, indented by column
        prefix = '.. %s:%s:: ' % (domain, directive)
        for i, sig_line in enumerate(sig.split("\n")):
            self.add_line('%s%s%s' % (prefix, name, sig_line),
                          sourcename)
            if i == 0:
                prefix = " " * len(prefix)
```

This also means you can remove `add_text`

tk0miya: Good point!

eric-wieser (approved):

tk0miya:
Thank you always!

PR #7683, opened 2020-05-17
Re-allow NoUri from missing-reference handlers

jakobandersen:
### Feature or Bugfix
- Feature
- Bugfix (for change only in the 3.x branch)

### Purpose
In b9793e5519a3e974886f8316fa1e74339075a1d2 (on the 3.x branch) some nice error handling is introduced for when event handlers raise exceptions. However, previously is was possible to have an extension handling ``missing-reference`` raise ``NoUri`` to silence warnings for specific references. This exception no longer reaches https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/transforms/post_transforms/__init__.py#L100.
I'm not sure this usage was documented before, but I believe something similar is needed. That is, a handler for ``missing-reference`` must be able to produce 3 results:
- Resolving the reference: return a node.
- Not able to resolve the reference: return ``None``.
- Able to resolve, but has no target: ? (previously raise ``NoUri``)

Use #7684 to explicitly allow ``NoUri`` from ``missing-reference`` handlers.

### DIFF
--- a/doc/extdev/appapi.rst
+++ b/doc/extdev/appapi.rst
@@ -230,11 +230,15 @@ connect handlers to the events.  Example:
 
 .. event:: missing-reference (app, env, node, contnode)
 
-   Emitted when a cross-reference to a Python module or object cannot be
-   resolved.  If the event handler can resolve the reference, it should return a
+   Emitted when a cross-reference to an object cannot be resolved.
+   If the event handler can resolve the reference, it should return a
    new docutils node to be inserted in the document tree in place of the node
    *node*.  Usually this node is a :class:`reference` node containing *contnode*
    as a child.
+   If the handler can not resolve the cross-reference,
+   it can either return ``None`` to let other handlers try,
+   or raise :class:`NoUri` to prevent other handlers in trying and suppress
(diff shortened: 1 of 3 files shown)


tk0miya:
+0: But I object that `NoUri` inherits `SphinxError`. How about adding a parameter to `emit()` to indicate exceptions that the event allows?
```
app.event.emit('resolve-reference', env, node, contnode, allowed_exceptions=[NoUri])
```

jakobandersen:
@tk0miya, sure. As long as there is the third option for suppressing warnings on missing references, I'm mostly indifferent to the mechanism :-). Though, your suggestions seems like a more elegant solution so please go ahead.
(Note that the event in question is ``missing-reference``)

tk0miya:
I post https://github.com/sphinx-doc/sphinx/pull/7684 to realize the option. If you're okay, let's go with this.

jakobandersen:
Great, I have updated the PR to use the new functionality. The documentation update for ``missing-reference`` is also there.

tk0miya (approved):
LGTM!

Issue #6861, opened 2019-11-25
Warning about 'optional' in Numpy-style docstrings with sphinx.ext.napoleon


astrofrog:
**Describe the bug**

When using ``sphinx.ext.napoleon`` with Numpy-format docstrings and ``nitpicky`` mode, a warning is emitted about ``optional``.

**To Reproduce**

Create an empty directory and add the following files:

* ``conf.py``:

```python
import sys
sys.path.insert(0, '.')
extensions = ['sphinx.ext.autodoc',
              'sphinx.ext.napoleon',
              'sphinx.ext.intersphinx']
napoleon_google_docstring = False
intersphinx_mapping = {'python': ('https://docs.python.org/3.7', None)}
nitpicky = True
```

* ``mycode.py``:

```python
def my_func(arg1, arg2):
    """
    Parameters
    ----------
    arg1 : str
        Description of `arg1`
    arg2 : int, optional
        Description of `arg2`, defaults to 0
    """
    pass
```

* ``index.rst``:

```rst
 .. autofunction:: mycode.my_func
```

Then run:

```
sphinx-build -b html -d _build/doctrees   . _build/html
```

The output includes a warning about the ``optional`` string which should not be interpreted as a type:

```
Running Sphinx v2.2.1
making output directory... done
loading intersphinx inventory from https://docs.python.org/3.7/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
reading sources... [100%] index                                                                                               
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index                                                                                                
/home/tom/tmp/debug/mycode.py:docstring of mycode.my_func:: WARNING: py:class reference target not found: optional
generating indices...  genindexdone
writing additional pages...  searchdone
copying static files... ... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 1 warning.

The HTML pages are in _build/html.
```

**Expected behavior**

No warning should be emitted since adding ``, optional`` is valid by the numpy doc spec.

**Environment info**
- OS: Ubuntu 19.04
- Python version: 3.7.3
- Sphinx version: 2.2.1
- Sphinx extensions:  sphinx.ext.napoleon, sphinx.ext.autodoc, sphinx.ext.intersphinx

astrofrog:
[...]

Issue #4961, opened 2018-05-12
more than one target found for cross reference should prefer current file's method over other libraries methods


dvorapa:
Subject: more than one target found for cross reference should prefer current file's method over other libraries methods

### Problem
- When building HTML of our project, Sphinx outputs cca 200 warnings like this. But half of them should be never thrown, because Sphinx is not sure, whether to use current file's one or imported library's one. For example, see:
```
/src/pywikibot/interwiki_graph.py:docstring of pywikibot.interwiki_graph.GraphDrawer.__init__:: WARNING: more than one target found for cross-reference 'Subject': scripts.interwiki.Subject, pywikibot.interwiki_graph.Subject
```
The expected one is clear: The one from current file interwiki_graph should be preferred over the imported, but Sphinx still throwns a warning.

#### Procedure to reproduce the problem
Build HTML of Pywikibot (last build by us: https://integration.wikimedia.org/ci/job/pywikibot-core-tox-publish/424/console) or try to simulate the situation

#### Error logs / results
```
/src/pywikibot/interwiki_graph.py:docstring of pywikibot.interwiki_graph.GraphDrawer.__init__:: WARNING: more than one target found for cross-reference 'Subject': scripts.interwiki.Subject, pywikibot.interwiki_graph.Subject
```
- https://integration.wikimedia.org/ci/job/pywikibot-core-tox-publish/424/console

#### Expected results
It should prefer the current file over the imported one

### Reproducible project / your project
- https://www.mediawiki.org/wiki/Manual:Pywikibot

### Environment info
- OS: Debian Jessie
- Python version: 3.4
- Sphinx version: 1.7.4
- Napoleon, Epytext

tk0miya:
Where can I see the content of`interwiki_graph`? I'd like to see your source and try to reproduce this.
I can't understand how to do it. Please let me know the way to reproduce (or investigate).

>Reproducible project / your project
>https://www.mediawiki.org/wiki/Manual:Pywikibot

dvorapa:
>Reproducible project / your project
>https://www.mediawiki.org/wiki/Manual:Pywikibot

As you could find out in that link, the source code can be found for example here: https://phabricator.wikimedia.org/diffusion/PWBC/. Also you could look to https://pypi.org/project/pywikibot/.

[...]

Issue #272, opened 2015-01-02
Consider inheritence when resolving `:meth:` references


shimizukawa:
In the below, it would be great for B.bar to resolve the reference to A.foo:

```
#!python
class A(object):
    def foo(self):
        """foo doc"""

class B(A):
    def bar(self):
        """Calls :meth:`foo`."""
        self.foo()
```

If 'B' explicitly adds 'foo' or uses :inherited-members:, it will resolve to B.foo, but pulling down the inherited member doc text into B's isn't always desired.  Many times I want to reference a single, canonical definition of 'foo', the one on A.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/272
- Originally reported by: jason kirtland
- Originally created at: 2009-10-26T23:19:38.447

shimizukawa:
_From Georg Brandl on 2009-11-08 15:33:23+00:00_

While this requires a bit of work (and won't scale to non-autodoc docs), I think it should be possible to implement.

Setting milestone to 1.1 for now.

shimizukawa:
_From Georg Brandl on 2011-01-06 20:03:32+00:00_

Not yet.

PR #7690, opened 2020-05-17
preprocessing numpy types

keewis:
Subject: `napoleon` should translate the type fields

### Feature or Bugfix
- Feature
- Bugfix

could be both?

### Purpose
As reported in #6861, `napoleon` simply puts everything after the colon of a `numpy`-style parameter description (e.g. the `int` in `val : int`) into a `:type:` field. This works most of the time, but `numpydoc` also uses certain keywords (`optional`, `default`), value sets (e.g. `{"val1", "val2"}`) and literals (as in `default: "val2"`), among others. Right now, using those will raise nit-picky warnings and possibly also result in incorrect parameter documentation.

This PR adds a new role (named `noref`, but could also be `literal` or `text`; I'm not confident in my ability to come up with good names) that is then used to make the type field ignore the keywords / literals above.

The type preprocessing also allows adding a mapping of terms to other links (see the `translations` dict), so terms like `dict-like` or `array-like` / `array_like` can be mapped to `mapping` using a configuration option. It isn't restricted to terms, though, any kind of role should be possible.

I don't really know if this is the best approach to the problem described in #6861 (in that issue, the suggestion to create a new section for optional parameters came up) but I'm submitting this to start making progress.

### Relates
- #6861

### DIFF
--- a/doc/usage/extensions/napoleon.rst
+++ b/doc/usage/extensions/napoleon.rst
@@ -278,7 +278,7 @@ sure that "sphinx.ext.napoleon" is enabled in `conf.py`::
 .. _Google style:
    https://google.github.io/styleguide/pyguide.html
 .. _NumPy style:
-   https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt
+   https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard
 
 .. confval:: napoleon_google_docstring
 
--- a/doc/usage/extensions/napoleon.rst
+++ b/doc/usage/extensions/napoleon.rst
@@ -274,6 +274,7 @@ sure that "sphinx.ext.napoleon" is enabled in `conf.py`::
     napoleon_use_ivar = False
     napoleon_use_param = True
     napoleon_use_rtype = True
+    napoleon_type_aliases = None
 
 .. _Google style:
    https://google.github.io/styleguide/pyguide.html
@@ -435,7 +436,7 @@ sure that "sphinx.ext.napoleon" is enabled in `conf.py`::
        :param arg1: Description of `arg1`
        :type arg1: str
        :param arg2: Description of `arg2`, defaults to 0
-       :type arg2: int, optional
+       :type arg2: :class:`int`, *optional*
 
    **If False**::
 
@@ -480,3 +481,31 @@ sure that "sphinx.ext.napoleon" is enabled in `conf.py`::
    **If False**::
 
        :returns: *bool* -- True if successful, False otherwise
+
+.. confval:: napoleon_type_aliases
+
+   A mapping to translate type names to other names or references. Works
+   only when ``napoleon_use_param = True``. *Defaults to None.*
+
+   With::
+
+       napoleon_type_aliases = {
+           "CustomType": "mypackage.CustomType",
+           "dict-like": ":term:`dict-like <mapping>`",
+       }
+
+   This `NumPy style`_ snippet::
+
+       Parameters
+       ----------
+       arg1 : CustomType
+           Description of `arg1`
+       arg2 : dict-like
+           Description of `arg2`
+
+   becomes::
+
+       :param arg1: Description of `arg1`
+       :type arg1: mypackage.CustomType
+       :param arg2: Description of `arg2`
+       :type arg2: :term:`dict-like <mapping>`
--- a/sphinx/ext/napoleon/__init__.py
+++ b/sphinx/ext/napoleon/__init__.py
@@ -41,6 +41,7 @@ class Config:
         napoleon_use_param = True
         napoleon_use_rtype = True
         napoleon_use_keyword = True
+        napoleon_numpy_type_aliases = None
         napoleon_custom_sections = None
 
     .. _Google style:
@@ -236,6 +237,9 @@ class Config:
 
             :returns: *bool* -- True if successful, False otherwise
 
+    napoleon_numpy_type_aliases : :obj:`dict` (Defaults to None)
+        Add a mapping of strings to string, translating types.
+
     napoleon_custom_sections : :obj:`list` (Defaults to None)
         Add a list of custom sections to include, expanding the list of parsed sections.
 
@@ -263,6 +267,7 @@ class Config:
         'napoleon_use_param': (True, 'env'),
         'napoleon_use_rtype': (True, 'env'),
         'napoleon_use_keyword': (True, 'env'),
+        'napoleon_numpy_type_aliases': (None, 'env'),
         'napoleon_custom_sections': (None, 'env')
     }
 
--- a/sphinx/ext/napoleon/__init__.py
+++ b/sphinx/ext/napoleon/__init__.py
@@ -238,7 +238,8 @@ class Config:
             :returns: *bool* -- True if successful, False otherwise
 
     napoleon_type_aliases : :obj:`dict` (Defaults to None)
-        Add a mapping of strings to string, translating types in numpy style docstrings.
+        Add a mapping of strings to string, translating types in numpy
+        style docstrings. Only works when ``napoleon_use_param = True``.
 
     napoleon_custom_sections : :obj:`list` (Defaults to None)
         Add a list of custom sections to include, expanding the list of parsed sections.
--- a/sphinx/ext/napoleon/__init__.py
+++ b/sphinx/ext/napoleon/__init__.py
@@ -41,7 +41,7 @@ class Config:
         napoleon_use_param = True
         napoleon_use_rtype = True
         napoleon_use_keyword = True
-        napoleon_numpy_type_aliases = None
+        napoleon_type_aliases = None
         napoleon_custom_sections = None
 
     .. _Google style:
@@ -237,8 +237,8 @@ class Config:
 
             :returns: *bool* -- True if successful, False otherwise
 
-    napoleon_numpy_type_aliases : :obj:`dict` (Defaults to None)
-        Add a mapping of strings to string, translating types.
+    napoleon_type_aliases : :obj:`dict` (Defaults to None)
+        Add a mapping of strings to string, translating types in numpy style docstrings.
 
     napoleon_custom_sections : :obj:`list` (Defaults to None)
         Add a list of custom sections to include, expanding the list of parsed sections.
@@ -267,7 +267,7 @@ class Config:
         'napoleon_use_param': (True, 'env'),
         'napoleon_use_rtype': (True, 'env'),
         'napoleon_use_keyword': (True, 'env'),
-        'napoleon_numpy_type_aliases': (None, 'env'),
+        'napoleon_type_aliases': (None, 'env'),
         'napoleon_custom_sections': (None, 'env')
     }
 
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -15,15 +15,15 @@ import re
 from functools import partial
 from typing import Any, Callable, Dict, List, Tuple, Union
 
+from docutils import nodes
+from docutils.nodes import Node, system_message
+from docutils.parsers.rst import roles
+
 from sphinx.application import Sphinx
 from sphinx.config import Config as SphinxConfig
 from sphinx.ext.napoleon.iterators import modify_iter
 from sphinx.locale import _
-
 from sphinx.util.docutils import SphinxRole
-from docutils import nodes
-from docutils.nodes import Node, system_message
-from docutils.parsers.rst import roles
 
 
 class LiteralText(SphinxRole):
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -791,8 +791,8 @@ class GoogleDocstring:
         return lines
 
 
-def _recombine_sets(tokens):
-    tokens = collections.deque(tokens)
+def _recombine_set_tokens(tokens: List[str]) -> List[str]:
+    token_queue = collections.deque(tokens)
     keywords = ("optional", "default")
 
     def takewhile_set(tokens):
@@ -841,10 +841,10 @@ def _recombine_sets(tokens):
             else:
                 yield token
 
-    return list(combine_set(tokens))
+    return list(combine_set(token_queue))
 
 
-def _tokenize_type_spec(spec):
+def _tokenize_type_spec(spec: str) -> List[str]:
     tokens = list(
         item
         for item in _token_regex.split(spec)
@@ -853,7 +853,7 @@ def _tokenize_type_spec(spec):
     return tokens
 
 
-def _token_type(token):
+def _token_type(token: str, location: str = None) -> str:
     if token.startswith(" ") or token.endswith(" "):
         type_ = "delimiter"
     elif (
@@ -867,28 +867,28 @@ def _token_type(token):
         logger.warning(
             __("invalid value set (missing closing brace): %s"),
             token,
-            location=None,
+            location=location,
         )
         type_ = "literal"
     elif token.endswith("}"):
         logger.warning(
             __("invalid value set (missing opening brace): %s"),
             token,
-            location=None,
+            location=location,
         )
         type_ = "literal"
     elif token.startswith("'") or token.startswith('"'):
         logger.warning(
             __("malformed string literal (missing closing quote): %s"),
             token,
-            location=None,
+            location=location,
         )
         type_ = "literal"
     elif token.endswith("'") or token.endswith('"'):
         logger.warning(
             __("malformed string literal (missing opening quote): %s"),
             token,
-            location=None,
+            location=location,
         )
         type_ = "literal"
     elif token in ("optional", "default"):
@@ -901,7 +901,7 @@ def _token_type(token):
     return type_
 
 
-def _convert_numpy_type_spec(_type, translations={}):
+def _convert_numpy_type_spec(_type: str, location: str = None, translations: dict = {}) -> str:
     def convert_obj(obj, translations, default_translation):
         # use :class: (the default) only if obj is not a standard singleton (None, True, False)
         if obj in ("None", "True", "False") and default_translation == ":class:`%s`":
@@ -910,9 +910,9 @@ def _convert_numpy_type_spec(_type, translations={}):
         return translations.get(obj, default_translation % obj)
 
     tokens = _tokenize_type_spec(_type)
-    combined_tokens = _recombine_sets(tokens)
+    combined_tokens = _recombine_set_tokens(tokens)
     types = [
-        (token, _token_type(token))
+        (token, _token_type(token, location))
         for token in combined_tokens
     ]
 
@@ -1035,6 +1035,16 @@ class NumpyDocstring(GoogleDocstring):
         self._directive_sections = ['.. index::']
         super().__init__(docstring, config, app, what, name, obj, options)
 
+    def _get_location(self) -> str:
+        filepath = inspect.getfile(self._obj) if self._obj is not None else ""
+        name = self._name
+        line = ""
+
+        if filepath is None and name is None:
+            return None
+
+        return ":".join([filepath, "docstring of %s" % name, line])
+
     def _consume_field(self, parse_type: bool = True, prefer_type: bool = False
                        ) -> Tuple[str, str, List[str]]:
         line = next(self._line_iter)
@@ -1046,6 +1056,7 @@ class NumpyDocstring(GoogleDocstring):
         _name = self._escape_args_and_kwargs(_name)
         _type = _convert_numpy_type_spec(
             _type,
+            location=self._get_location(),
             translations=self._config.napoleon_type_aliases or {},
         )
 
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -904,10 +904,10 @@ def _token_type(token):
 def _convert_numpy_type_spec(_type, translations={}):
     def convert_obj(obj, translations, default_translation):
         # use :class: (the default) only if obj is not a standard singleton (None, True, False)
-        if obj in ("None", "True", "False") and default_translation == ":class:`{}`":
-            default_translation = ":obj:`{}`"
+        if obj in ("None", "True", "False") and default_translation == ":class:`%s`":
+            default_translation = ":obj:`%s`"
 
-        return translations.get(obj, default_translation.format(obj))
+        return translations.get(obj, default_translation % obj)
 
     tokens = _tokenize_type_spec(_type)
     combined_tokens = _recombine_sets(tokens)
@@ -918,15 +918,15 @@ def _convert_numpy_type_spec(_type, translations={}):
 
     # don't use the object role if it's not necessary
     default_translation = (
-        ":class:`{}`"
+        ":class:`%s`"
         if not all(type_ == "obj" for _, type_ in types)
-        else "{}"
+        else "%s"
     )
 
     converters = {
-        "literal": lambda x: "``{x}``".format(x=x),
+        "literal": lambda x: "``%s``" % x,
         "obj": lambda x: convert_obj(x, translations, default_translation),
-        "control": lambda x: "*{x}*".format(x=x),
+        "control": lambda x: "*%s*" % x,
         "delimiter": lambda x: x,
         "reference": lambda x: x,
     }
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -858,10 +858,10 @@ def _token_type(token):
     if token.startswith(" ") or token.endswith(" "):
         type_ = "delimiter"
     elif (
-            token.isnumeric()
-            or (token.startswith("{") and token.endswith("}"))
-            or (token.startswith('"') and token.endswith('"'))
-            or (token.startswith("'") and token.endswith("'"))
+            token.isnumeric() or
+            (token.startswith("{") and token.endswith("}")) or
+            (token.startswith('"') and token.endswith('"')) or
+            (token.startswith("'") and token.endswith("'"))
     ):
         type_ = "literal"
     elif token.startswith("{"):
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -845,10 +845,17 @@ def _recombine_set_tokens(tokens: List[str]) -> List[str]:
 
 
 def _tokenize_type_spec(spec: str) -> List[str]:
+    def postprocess(item):
+        if item.startswith("default"):
+            return [item[:7], item[7:]]
+        else:
+            return [item]
+
     tokens = list(
         item
-        for item in _token_regex.split(spec)
-        if item is not None and item.strip()
+        for raw_token in _token_regex.split(spec)
+        for item in postprocess(raw_token)
+        if item
     )
     return tokens
 
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -845,14 +845,17 @@ def _convert_numpy_type_spec(_type, translations={}):
     def token_type(token):
         if token.startswith(" ") or token.endswith(" "):
             type_ = "delimiter"
-        elif token.startswith("{") and token.endswith("}"):
-            type_ = "value_set"
+        elif (
+                token.isnumeric()
+                or (token.startswith("{") and token.endswith("}"))
+                or (token.startswith('"') and token.endswith('"'))
+                or (token.startswith("'") and token.endswith("'"))
+                ):
+            type_ = "literal"
         elif token in ("optional", "default"):
             type_ = "control"
         elif _xref_regex.match(token):
             type_ = "reference"
-        elif token.isnumeric() or (token.startswith('"') and token.endswith('"')):
-            type_ = "literal"
         else:
             type_ = "obj"
 
@@ -875,7 +878,6 @@ def _convert_numpy_type_spec(_type, translations={}):
     )
 
     converters = {
-        "value_set": lambda x: f"``{x}``",
         "literal": lambda x: f"``{x}``",
         "obj": lambda x: convert_obj(x, translations, default_translation),
         "control": lambda x: f"*{x}*",
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -10,6 +10,7 @@
     :license: BSD, see LICENSE for details.
 """
 
+import collections
 import inspect
 import re
 from functools import partial
@@ -790,16 +791,33 @@ class GoogleDocstring:
         return lines
 
 
-def _recombine_set_tokens(tokens):
-    def takewhile_set(iterable):
-        yield "{"
+def _recombine_sets(tokens):
+    tokens = collections.deque(tokens)
+    keywords = ("optional", "default")
 
-        open_braces = 1
+    def takewhile_set(tokens):
+        open_braces = 0
+        previous_token = None
+        print("combining set:", tokens)
         while True:
             try:
-                token = next(iterable)
-            except StopIteration:
+                token = tokens.popleft()
+            except IndexError:
                 break
+
+            if token == ", ":
+                previous_token = token
+                continue
+
+            if token in keywords:
+                tokens.appendleft(token)
+                if previous_token is not None:
+                    tokens.appendleft(previous_token)
+                break
+
+            if previous_token is not None:
+                yield previous_token
+                previous_token = None
 
             if token == "{":
                 open_braces += 1
@@ -812,26 +830,28 @@ def _recombine_set_tokens(tokens):
                 break
 
     def combine_set(tokens):
-        iterable = iter(tokens)
         while True:
             try:
-                token = next(iterable)
-            except StopIteration:
+                token = tokens.popleft()
+            except IndexError:
                 break
 
-            yield "".join(takewhile_set(iterable)) if token == "{" else token
+            if token == "{":
+                tokens.appendleft("{")
+                yield "".join(takewhile_set(tokens))
+            else:
+                yield token
 
     return list(combine_set(tokens))
 
 
 def _tokenize_type_spec(spec):
-    tokens = tuple(
+    tokens = list(
         item
         for item in _token_regex.split(spec)
         if item is not None and item.strip()
     )
-    return _recombine_set_tokens(tokens)
-
+    return tokens
 
 
 def _token_type(token):
@@ -842,7 +862,7 @@ def _token_type(token):
             or (token.startswith("{") and token.endswith("}"))
             or (token.startswith('"') and token.endswith('"'))
             or (token.startswith("'") and token.endswith("'"))
-            ):
+    ):
         type_ = "literal"
     elif token.startswith("{"):
         logger.warning(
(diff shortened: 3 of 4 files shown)


sphinx/ext/napoleon/docstring.py
```diff
@@ -872,6 +884,83 @@ def __init__(self, docstring: Union[str, List[str]], config: SphinxConfig = None
...
+            (token, token_type(token))
+            for token in tokens
+        ]
+
+        # TODO: make this configurable
+        translations = {
+            "sequence": ":term:`sequence`",
+            "mapping": ":term:`mapping`",
+            "dict-like": ":term:`dict-like <mapping>`",
+        }
```
tk0miya: Why using terms for these entries?

sphinx/ext/napoleon/docstring.py
```diff
@@ -872,6 +884,83 @@ def __init__(self, docstring: Union[str, List[str]], config: SphinxConfig = None
         self._directive_sections = ['.. index::']
         super().__init__(docstring, config, app, what, name, obj, options)
 
+    def _convert_type_spec(self, _type):
```
tk0miya: I think more test cases are needed for this function. So it would be better to make this as an independent function.

sphinx/ext/napoleon/docstring.py
```diff
@@ -872,6 +884,83 @@ def __init__(self, docstring: Union[str, List[str]], config: SphinxConfig = None
...
+        def recombine_set(tokens):
+            def combine_set(tokens):
+                in_set = False
+                set_items = []
+
+                for token in tokens:
```
tk0miya: It seems this does not support a set that is not closed (ex. `{1, 2`). It is not good to suppress it silently.

sphinx/ext/napoleon/docstring.py
```diff
@@ -872,6 +884,83 @@ def __init__(self, docstring: Union[str, List[str]], config: SphinxConfig = None
...
+                    if token.startswith("{"):
```
tk0miya: Does value_set support a literal of set or dict (ex. `{1, "2", {3}, {"key": 4}}`)?

sphinx/ext/napoleon/docstring.py
```diff
@@ -872,6 +884,83 @@ def __init__(self, docstring: Union[str, List[str]], config: SphinxConfig = None
...
+        def token_type(token):
+            if token.startswith(" ") or token.endswith(" "):
+                type_ = "delimiter"
+            elif token.startswith("{") and token.endswith("}"):
+                type_ = "value_set"
+            elif token in ("optional", "default"):
+                type_ = "control"
+            elif re.match(":[^:]+:`[^`]+`", token):
+                type_ = "reference"
+            elif token.isnumeric() or (token.startswith('"') and token.endswith('"')):
```
tk0miya: How about single-quoted strings?

sphinx/ext/napoleon/docstring.py
```diff
@@ -872,6 +884,83 @@ def __init__(self, docstring: Union[str, List[str]], config: SphinxConfig = None
...
+            "sequence": ":term:`sequence`",
+            "mapping": ":term:`mapping`",
+            "dict-like": ":term:`dict-like <mapping>`",
+        }
+
+        converters = {
+            "value_set": lambda x: f":noref:`{x}`",
+            "literal": lambda x: f":noref:`{x}`",
+            "obj": lambda x: convert_obj(x, translations),
+            "control": lambda x: f":noref:`{x}`",
```
tk0miya: I still don't understand why `:noref:` role is needed. What is different from `lambda x: x`?

tk0miya (changes requested):
Your idea is great! LGTM!

sphinx/ext/napoleon/docstring.py
keewis: yes, I never considered that. Should I raise a warning if the closing curly brace is missing? This would probably require a parser for these expressions (but see also the comment below).

sphinx/ext/napoleon/docstring.py
keewis: not sure, I didn't think of this so it probably won't. Is that valid `numpydoc`?

I suspect that if we want to support it, we somehow need to count opening and closing braces (or write/use a parser (json?))

sphinx/ext/napoleon/docstring.py
keewis: will do

sphinx/ext/napoleon/docstring.py
keewis: these are examples of terms (i.e. not types that could be referenced with `obj`) I have often seen, e.g. in the `pandas` or `xarray` documentation (which is what I'm mostly working with), but I wouldn't put those in the extension.

Instead, I'd like to expose a `napoleon_type_aliases` option that allows translating objects in the type spec to user-specified references (the mapping with the terms is a placeholder that will be deleted when the option is introduced).

sphinx/ext/napoleon/docstring.py
keewis: without this PR,
```
param1 : str, optional
    description
```
gets translated to
```
:param param1: description
:type: str, optional
```
and sphinx will try to reference both `str` and `optional` (using `:obj:`?). To make sure `optional` (or any literal / modifier, etc.) does not get referenced, we could instead use
```
:param param1: description
:type: :obj:`str`, optional
```
but we can't do that if there's no real type (which is valid):
```
param1 : optional
    description
```
In that case, the only way to still make sure `optional` does not get treated as a type I could find is by explicitly marking it with a no-op role, hence the `:noref:`:
```
:param param1: description
:type: :noref:`optional`
```

sphinx/ext/napoleon/docstring.py
keewis: sure, I'll move it out of the class (it doesn't need `self` anyways) and add more tests

sphinx/ext/napoleon/docstring.py
keewis: looking at this a bit more, the style reference is not really clear about this, it only says that this:
```
{'C', 'F', 'A'}
```
is valid. The `numpydoc` extension (which I'd assume is the reference implementation) actually doesn't enforce anything, it simply uses whatever is after the first colon as a type.

So since by parsing the types we are already enforcing a certain format, I guess we can decide if we want to enforce flattened sets.

Another question is how we treat continuation lines: at the moment, only the line of the parameter is passed, but the type for
```
param : {"value1", "value2",
         "value3", "value3"}, optional
```
is `{"value1", "value2",` right now (`numpydoc` doesn't say anything about continuation lines and in numpy/numpydoc#87 it was recommended to use a backslash to continue the line: should we use that, too?)

Edit: it will now accept anything between the first pair of braces, raising a `ValueError` if the number of opening braces don't match the number of closing braces. As mentioned above, we could also try to use the parser from the `json` library, but I'm not sure if that's worth it

sphinx/ext/napoleon/docstring.py
```diff
@@ -773,6 +785,108 @@ def _strip_empty(self, lines: List[str]) -> List[str]:
...
+    def takewhile_set(iterable):
+        yield "{"
+
+        open_braces = 1
+        while True:
+            try:
+                token = next(iterable)
+            except StopIteration:
+                if open_braces != 0:
+                    raise ValueError("invalid value set: {}".format("".join(tokens)))
```
keewis: this should probably become a warning

keewis: @tk0miya, how would I emit that warning?

tk0miya: Please use `logger.warning()` to emit a warning. `__()` is also helpful to make a warning message translatable.

sphinx/ext/napoleon/docstring.py
tk0miya: I still don't understand they should be regarded as terms commonly in many documents. So I can accept to provide `napoleon_type_aliases` to change this behavior. And I prefer not to convert them as `:term:` references.

sphinx/ext/napoleon/docstring.py
tk0miya: I think the error case should be warned. To suppress errors is not kind for users. I think all humans sometimes get mistakes. So we have to support such irregular cases.

sphinx/ext/napoleon/docstring.py
tk0miya: Sorry for the lack of my words. I don't know what is the correct format of numpydoc. But users usually get mistakes. So we have to handle and let users know them.

sphinx/ext/napoleon/docstring.py
tk0miya: Please wait a moment. I consider the way not to use the role. I think it's not good to add `:noref:` role globally because it is not semantic...

Note: We can stop the behavior, text emphasis can be used (ex. `:type: *optional*`). The python domain check the type is single text or not. So the marked up text (like `:noref:` and emphasis) is not converted to the reference.

sphinx/ext/napoleon/docstring.py
keewis: I agree, this should not be the default. So I'll try to create the option and remove the `:term:` references.

sphinx/ext/napoleon/docstring.py
keewis: agreed, I'll do that

sphinx/ext/napoleon/docstring.py
keewis: as I said the new role was the only way I could find, but I didn't try text emphasis (thanks for suggesting that).

Since text emphasis changes the way the text is displayed, we need to consider how to display the different token types: Should control words (`optional`, `default`) be displayed as *optional* or **optional**? What about literals (strings, numbers, booleans, value sets)?

I'd argue we should try to use different styles for both to make the difference visible. To do that, I'd like to propose using double backticks for literals (`True`, `False` and `None` can be referenced):
```
``"xyz"``
``{"abc", "def"}``
``1.4``
``10``
:obj:`True`
:obj:`None`
```
and emphasis for control tokens (`*optional*` and `*default*`). I don't feel strongly about this, though.

Also, the role I chose as default for type tokens (`:obj:`) is displayed differently from the default role of the `:type:` field. I think we should try to be consistent here (either always use `:obj:` or always use the default role of the `:type:` field).

sphinx/ext/napoleon/docstring.py
keewis: done, but the documentation of the new option could be better

sphinx/ext/napoleon/docstring.py
keewis: my current approach for splitting the tokens can easily be broken by putting braces in the strings. I'll look into this a bit more

tests/test_ext_napoleon_docstring.py
```diff
@@ -1950,6 +1951,160 @@ def test_list_in_parameter_description(self):
...
+            actual = _recombine_set_tokens(input_tokens)
+            self.assertEqual(expected, actual)
+
+    def test_recombine_set_tokens_invalid(self):
+        invalid_tokens = (
+            ["{", "1", ", ", "2"],
+        )
+
+        for input_tokens in invalid_tokens:
+            with self.assertWarnsRegex(UserWarning, "invalid value set:"):
```
keewis: @tk0miya: how do I test that invalid value sets emit a warning? Using `assertWarnsRegex` obviously doesn't work since the warning is emitted by the logger and not the `warnings` module

Also, is it okay to use `pytest` here? I think it's already used in some of the other tests, but not in `napoleon` tests.

sphinx/ext/napoleon/docstring.py
```diff
@@ -773,6 +783,106 @@ def _strip_empty(self, lines: List[str]) -> List[str]:
...
+        while True:
+            try:
+                token = next(iterable)
+            except StopIteration:
+                if open_braces != 0:
+                    location = ("", "")
+                    logger.warning(
+                        __("invalid value set: %r"),
+                        "".join(tokens),
+                        location=location,
```
keewis: this warning is not as useful as it could be since there's no information on the location of the invalid line. Is there a way to get that from the attributes of `NumpyDocstring`?

keewis: also, my test project suggests that this code is run twice and thus two identical warnings are raised. Is that intentional or a bug, or is this a problem with my configuration?

keewis:
gentle ping, @tk0miya.

As a summary, this is what I think still needs to be answered / done (see also above):
- what is the default role used by the `:type:` field?
- how do I add location information to the warnings?
- how do I modify the test so that I can verify that a malformed value set emits the warning (something like `self.assertWarnsRegex` or `pytest.warns` that works with the logger)?
- the warning tends to be emitted multiple times when using `autosummary` / `autodoc`. Is that something I can fix?
- regarding continuation lines: I adopted the recommendation in `numpydoc` (it seems that we'd need to modify the `rst` standard on definition blocks if we wanted to do something else). Any objections?
- when extending a library using inheritance, the docstrings of inherited methods often contain references to other objects within the same namespace (without specifying the qualname). Is there a existing way to specify packages that should be checked if it can't be found in the current namespace?
- I just noticed `default` doesn't have a standardized format. `pandas` seems to mostly use `default <default>` but with the current implementation that is really difficult to support (but not impossible). Should I try to get this to work?

keewis:
gentle ping, @tk0miya

sphinx/ext/napoleon/docstring.py
```diff
@@ -773,6 +784,148 @@ def _strip_empty(self, lines: List[str]) -> List[str]:
         return lines
 
 
+def _recombine_sets(tokens):
```
tk0miya: Could you add type annotations to every new function, please?

sphinx/ext/napoleon/docstring.py
```diff
@@ -773,6 +784,148 @@ def _strip_empty(self, lines: List[str]) -> List[str]:
...
+    tokens = collections.deque(tokens)
+    keywords = ("optional", "default")
+
+    def takewhile_set(tokens):
+        open_braces = 0
+        previous_token = None
+        print("combining set:", tokens)
```
tk0miya: I guess this debug-print is no longer needed, right?

sphinx/ext/napoleon/docstring.py
```diff
@@ -29,11 +32,19 @@
...
+_xref_regex = re.compile(
+    r'(?::(?:[a-zA-Z0-9]+[\-_+:.])*[a-zA-Z0-9]+:`.+?`)'
+)
 _bullet_list_regex = re.compile(r'^(\*|\+|\-)(\s+\S|\s*$)')
 _enumerated_list_regex = re.compile(
     r'^(?P<paren>\()?'
     r'(\d+|#|[ivxlcdm]+|[IVXLCDM]+|[a-zA-Z])'
     r'(?(paren)\)|\.)(\s+\S|\s*$)')
+_token_regex = re.compile(
+    r"(\sor\s|\sof\s|:\s|,\s|[{]|[}]"
```
tk0miya, line 48: Must A space be required just after `,` and `:`? Is it better to use `:\s*|,\s*` instead?

sphinx/ext/napoleon/docstring.py
```diff
@@ -773,6 +784,148 @@ def _strip_empty(self, lines: List[str]) -> List[str]:
...
+        open_braces = 0
+        previous_token = None
+        print("combining set:", tokens)
+        while True:
+            try:
+                token = tokens.popleft()
+            except IndexError:
+                break
+
+            if token == ", ":
```
tk0miya, line 807: Note: It might be better to use `token.strip() == ","` if we accept a comma without trailing space.

sphinx/ext/napoleon/docstring.py
```diff
@@ -773,6 +784,148 @@ def _strip_empty(self, lines: List[str]) -> List[str]:
...
+            (token.startswith("{") and token.endswith("}")) or
+            (token.startswith('"') and token.endswith('"')) or
+            (token.startswith("'") and token.endswith("'"))
+    ):
+        type_ = "literal"
+    elif token.startswith("{"):
+        logger.warning(
+            __("invalid value set (missing closing brace): %s"),
+            token,
+            location=None,
```
tk0miya: The `location` option is not needed if None.

sphinx/ext/napoleon/docstring.py
```diff
@@ -773,6 +784,148 @@ def _strip_empty(self, lines: List[str]) -> List[str]:
...
+
+    # don't use the object role if it's not necessary
+    default_translation = (
+        ":obj:`{}`"
+        if not all(type_ == "obj" for _, type_ in types)
+        else "{}"
+    )
+
+    converters = {
+        "literal": lambda x: f"``{x}``",
```
tk0miya: Could you use `%` operation to format a string? f-string is only available since py36. But we still support py35.

tests/test_ext_napoleon_docstring.py
```diff
@@ -1971,6 +1977,228 @@ def test_list_in_parameter_description(self):
...
+        tokens = (
+            "{1, 2",
+            "}",
+            "'abc",
+            "def'",
+            '"ghi',
+            'jkl"',
+        )
+        for token in tokens:
+            # TODO: check for the warning
```
tk0miya: I suppose you can get a StringIO object via its argument named `warning`.
Could you try to change the signature of this method to `test_token_type_invalid(self, warning)`?
Then you can check the warning messages like this:
```
for token in tokens:
    _token_type(token)
    self.assertIn(message, warning.getvalue())
    warning.truncate(0)   # reset warnings
```

tests/test_ext_napoleon_docstring.py
```diff
@@ -1971,6 +1977,228 @@ def test_list_in_parameter_description(self):
...
+            for type_ in types
+            for modifier in modifiers
+        )
+        tokens = tuple(
+            tokens_ + ([", "] + modifier_tokens_ if modifier_tokens_ else [])
+            for tokens_ in type_tokens
+            for modifier_tokens_ in modifier_tokens
+        )
+
+        for type_spec, expected in zip(type_specs, tokens):
```
tk0miya: I feel it is too complicated to generate test patterns via code. I took about 10 minutes to understand this test code. And I can't say the test code does not contain a bug. -1 for maintenance.

tk0miya (commented):
Sorry for very very late response. I'm back from the maintenance of 3.1.x bug fixes now.

>what is the default role used by the :type: field?

By default, `:class:` role is used. As an exception, `:obj:` is used for `None` value.

>how do I add location information to the warnings?

Unfortunately there no good way. The `autodoc-process-docstring` that napoleon uses does not send a location of the docstring. It might be able to use `self._obj` to obtain the name of the docstring owner.

>how do I modify the test so that I can verify that a malformed value set emits the warning (something like self.assertWarnsRegex or pytest.warns that works with the logger)?

Please see my comment below.

>the warning tends to be emitted multiple times when using autosummary / autodoc. Is that something I can fix?

I don't have an answer. I need to investigate what happened in this case. Can I reproduce that in this test case? If true, I'll take a look.

>regarding continuation lines: I adopted the recommendation in numpydoc (it seems that we'd need to modify the rst standard on definition blocks if we wanted to do something else). Any objections?

Of course, yes. no problem. The `NumpyDocstring` class has followed to numpydoc format.

>when extending a library using inheritance, the docstrings of inherited methods often contain references to other objects within the same namespace (without specifying the qualname). Is there a existing way to specify packages that should be checked if it can't be found in the current namespace?

This is another problem. In my memory is correct, a similar issue was already filed (I can't find it now...).

>I just noticed default doesn't have a standardized format. pandas seems to mostly use default <default> but with the current implementation that is really difficult to support (but not impossible). Should I try to get this to work?

I don't know about pandas at all. So I can't determine that.

In addition, the goal of napoleon (I supposed) is not supporting all of the formats for docstrings. They are only for google's and numpy's docstring. So no reason to support it.

sphinx/ext/napoleon/docstring.py
keewis: that would mean we'd allow `int,float` and `default:None` instead of `int, float` and `default: None` which I think we should not support (but I could do that if you want to warn about the missing space – right now a nitpicky warning would be raised instead because there's no class name containing `,` or `:`).

sphinx/ext/napoleon/docstring.py
keewis: sure

sphinx/ext/napoleon/docstring.py
keewis: yes, I forgot that one

sphinx/ext/napoleon/docstring.py
keewis: sure, but see my comment about the trailing space

sphinx/ext/napoleon/docstring.py
keewis: this was a placeholder to remind me to add that later. I'll change that to actually provide a location using your tip with `self._obj` (and `self._name`).

sphinx/ext/napoleon/docstring.py
keewis: is using `.format` acceptable, too?

Edit: never mind, I fixed that.

tests/test_ext_napoleon_docstring.py
keewis: unfortunately, that doesn't work: the warnings are generated using a *logger*, but `pytest`'s `warning` fixture only works with real warnings (e.g. by using `warnings.warn`). I'll check if mocking the logger would work, too.

Edit: turns out I just didn't understand how to use it; I had to make sure to use a `pytest` style test class, not a `unittest.TestCase` class. Also, I see now that that fixture is defined in `sphinx.testing.fixtures`.

tests/test_ext_napoleon_docstring.py
keewis: agreed. How do you feel about using `pytest.mark.parametrize` instead? I'd probably need a non `unittest.TestCase` test in order to be able to use that.

Edit: I needed the non-`unittest` test class to use `warning`, so that's less of an issue now

sphinx/ext/napoleon/docstring.py
keewis: sure.

keewis (commented):
> By default, :class: role is used. As an exception, :obj: is used for None value.

That still doesn't get a consistent link style: the docstring of the test project's `test.multiply` results in
```
:param a: first factor
:type a: :term:`array-like <array_like>`
:param b: second factor
:param \*variables: parameters without use
:type \*variables: str
:param mod: optionally compute the modulo of the product
:type mod: :term:`array-like <array_like>`, *optional*
:param files: save each value of the results into a new file
:type files: :class:`list` of :class:`str` or :class:`list` of :class:`os.PathLike` or :term:`dict-like <mapping>`
:param save_format: file format
:type save_format: ``{"ma{icious", "options", "that need a line break"}``, *default*: ``"options"``
:param test: optional parameter that does absolutely nothing
:type test: *optional*
:param \*\*variable_kwargs: kwargs form of variables
```
but the [rendered version](https://keewis-sphinx-test-project.readthedocs.io/en/latest/generated/test.multiply.html#test.multiply) ([logs](https://readthedocs.org/projects/keewis-sphinx-test-project/builds/)) displays certain objects as preformatted link. Is there any way to make that a plain link, just like the single `str` for `*variables`? Or should I just leave this as-is?

> I don't have an answer. I need to investigate what happened in this case. Can I reproduce that in this test case? If true, I'll take a look.

Unfortunately not by the test case, but with the project linked above. I guess that's because I'm using both `autosummary` and `autodoc`?

Edit: I can't reproduce it anymore. Not sure what happened, but I guess it's fixed?

> Unfortunately there no good way. The autodoc-process-docstring that napoleon uses does not send a location of the docstring. It might be able to use self._obj to obtain the name of the docstring owner.

Thanks, that works; now only the line number is missing. Since all other warnings about docstrings also seem to be missing the line number I guess that's not so bad.

> This is another problem. In my memory is correct, a similar issue was already filed (I can't find it now...).

I'll try to find that issue.

Edit: I can't find it either. However, I can work around that issue using the new `napoleon_type_aliases` setting, so this shouldn't be a blocker.

> I don't know about pandas at all. So I can't determine that.
>
> In addition, the goal of napoleon (I supposed) is not supporting all of the formats for docstrings. They are only for google's and numpy's docstring. So no reason to support it.

`pandas` uses `numpydoc`, but extends it slightly to allow documenting default values as part of the type. I'm *also* proposing to extend `numpydoc`, with a slightly different syntax (`default: None` instead of `pandas`' `default None`). Is it okay to include this for now (and ask on the `numpydoc` issue tracker) or should I wait for a response before including that?

sphinx/ext/napoleon/docstring.py
tk0miya: Is there any reason not to support them? I'd like to know the reason if I'll get a question in the future :-)

tests/test_ext_napoleon_docstring.py
tk0miya: I can't answer to move to pytest because the maintainer of napoleon is now absent. Personally, +1 for the move.

Anyway, I think the most important thing is not generating the expected value. Is it really needed to confirm all combination of types and modifiers? How about choosing some examples from the combinations?

keewis: not sure, I can try reducing that to a few manually generated examples

sphinx/ext/napoleon/docstring.py
keewis: no specific reason, it's just that these will generally reduce the readability of the type spec

tk0miya: Okay. Let's go as is. It would not too late to consider even after we get a report.

tests/test_ext_napoleon_docstring.py
keewis: done

keewis:
with this I think there's only a few issues left:
* formatting the created links consistently (but do tell me if you think that's not an issue)
* the namespacing for referenced objects (there might be another issue about that)
* the extension of `numpydoc` (`default: <value>` is not part of `numpydoc`, but it also won't reject it)
* am I merging into the correct branch? If `napoleon_use_param` is true there's a lot more warnings, not sure if that makes it a breaking change?

tk0miya:
>formatting the created links consistently (but do tell me if you think that's not an issue)

Sorry, my explanation above was not correct. It uses `:class:` and `:obj:` like role, but it uses emphasis node instead of literal. To be precise, they are not a role. The docfield mechanism creates the structure of nodes directly without roles.

And I can't find the way to build type descriptions having the same style.

>the namespacing for referenced objects (there might be another issue about that)

I found issues related to this: https://github.com/sphinx-doc/sphinx/issues/5977, https://github.com/sphinx-doc/sphinx/issues/4961 and https://github.com/sphinx-doc/sphinx/issues/272.

>the extension of numpydoc (default: <value> is not part of numpydoc, but it also won't reject it)
>am I merging into the correct branch? If napoleon_use_param is true there's a lot more warnings, not sure if that makes it a breaking change?

At present, the napoleon extension officially supports only google style and numpy style docstrings. So it should not add a support non-official notation. I'll never accept to support non-numpydoc syntax as numpydoc. The only I can agree is making napoleon extendable from 3rd party's extension. 

> am I merging into the correct branch? If napoleon_use_param is true there's a lot more warnings, not sure if that makes it a breaking change?

Oh, sorry. 3.x branch is better to merge.

About `napoleon_use_param`, we **have to** support both enabled and disabled. It's not good to bring new warnings to existing users after adding a new feature. I don't think to drop the flag for now. So I think it is a bug, not a breaking change.

keewis:
> And I can't find the way to build type descriptions having the same style.

Hmm... well, the links work so while the current state is not ideal, this can be fixed in a new PR.

> At present, the napoleon extension officially supports only google style and numpy style docstrings. So it should not add a support non-official notation. I'll never accept to support non-numpydoc syntax as numpydoc. The only I can agree is making napoleon extendable from 3rd party's extension.

So that means if I can get this to become official `numpydoc` we're fine?

Edit: see numpy/numpydoc#284

> 3.x branch is better to merge.

I'll change that

Edit: done

> About `napoleon_use_param`, we have to support both enabled and disabled. It's not good to bring new warnings to existing users after adding a new feature. I don't think to drop the flag for now. So I think it is a bug, not a breaking change.

I changed the code so the type spec only gets parsed with `napoleon_use_param = True`, which I think is consistent with the current behavior?

About the warnings: those are the ones that are generated for wrong type specs (invalid numpydoc), e.g. if value sets are not closed, end quotes are missing or there's a missing space after a comma. Those warnings have been nitpicky before (because linking to a value set failed) so if you tell me how I can change these to `nitpicky`, I will do that.

keewis:
It seems `numpydoc` (the implementation, not the guide) supports both formats (`default None` and `default: None`) so I guess we probably should support both, too?

keewis:
I added support for the other style, too. With that, I guess the only thing left is to decide what to do with the warnings?

sphinx/ext/napoleon/__init__.py
```diff
@@ -236,6 +237,9 @@ def __unicode__(self):
 
             :returns: *bool* -- True if successful, False otherwise
 
+    napoleon_type_aliases : :obj:`dict` (Defaults to None)
+        Add a mapping of strings to string, translating types in numpy style docstrings.
```
tk0miya: We have to mention this configuration only works when `napoleon_use_param` is True.

tk0miya: In addition, please add a document for this into `doc/usage/extensions/napoleon.rst`.

sphinx/ext/napoleon/docstring.py
```diff
@@ -879,6 +1047,16 @@ def __init__(self, docstring: Union[str, List[str]], config: SphinxConfig = None
...
 
+    def _get_location(self) -> str:
+        filepath = inspect.getfile(self._obj) if self._obj is not None else ""
+        name = self._name
+        line = ""
+
+        if filepath is None and name is None:
+            return None
+
+        return ":".join([filepath, "docstring of %s" % name, line])
```
tk0miya: It seems `line` is always empty. Why do you add it here?

sphinx/ext/napoleon/docstring.py
```diff
@@ -888,6 +1066,12 @@ def _consume_field(self, parse_type: bool = True, prefer_type: bool = False
             _name, _type = line, ''
         _name, _type = _name.strip(), _type.strip()
         _name = self._escape_args_and_kwargs(_name)
+        if self._config.napoleon_use_param:
+            _type = _convert_numpy_type_spec(
+                _type,
+                location=self._get_location(),
+                translations=self._config.napoleon_type_aliases or {},
+            )
```
tk0miya, line 1075: Do you have a plan to support numpy types on `napoleon_use_param` after this PR?

tk0miya:
>I changed the code so the type spec only gets parsed with napoleon_use_param = True, which I think is consistent with the current behavior?

Thank you for your update. I can accept your patch now (with nits). On the other hand, it would be wonderful if we'll support numpy types even for `napoleon_use_param = False`. Either way, it would be better to separate it into another PR to merge this early.

>About the warnings: those are the ones that are generated for wrong type specs (invalid numpydoc), e.g. if value sets are not closed, end quotes are missing or there's a missing space after a comma. Those warnings have been nitpicky before (because linking to a value set failed) so if you tell me how I can change these to nitpicky, I will do that.

I think it's acceptable warnings if the mark-ups are really broken. So it is not needed to change them nitpicky.

>It seems numpydoc (the implementation, not the guide) supports both formats (default None and default: None) so I guess we probably should support both, too?

I'm not familiar with numpydoc. So I'll merged this as is. From the point of view of a maintainer, it would be better to describe (by comment, docstring, or something) why it is implemented if it is not a part of "official" spec.

sphinx/ext/napoleon/docstring.py
keewis: because I was hoping to add that information somehow, but couldn't find out how to do that. Also, the format of the location of other warnings about docstrings (the nitpicky ones?) seems to be `"{path}:docstring of {func}:{line}"`, so to copy that, the empty string is necessary.

sphinx/ext/napoleon/docstring.py
keewis: this would not be terribly difficult, but we'd need to make sure to reference all types; right now, a parameter definition like
```python
param1 : str
    description
```
would not be changed. ~However, that could be a bug since now the type aliases wouldn't work. If so, supporting `napoleon_use_param = False` could be as simple as removing the condition.~ The type aliases work so I guess supporting this when `napoleon_use_param = False` should be a different PR. Note that this will break a few tests that have "invalid" type specs (something like `Arbitrary type` or `MyClass instance`).

keewis (commented):
> why it is implemented if it is not a part of "official" spec

I think that's because they didn't update the docstring guide in quite a while. I'll try submitting a PR changing the docstring guide (the official spec).

> it would be wonderful if we'll support numpy types even for `napoleon_use_param = False`.

Sure, but I guess that would change the current behavior of not linking at all.

sphinx/ext/napoleon/__init__.py
keewis: sure

sphinx/ext/napoleon/docstring.py
tk0miya: We usually suppress a line number on the location when hard to obtain it. For example, our logging module suppresses it automatically:
https://github.com/sphinx-doc/sphinx/blob/5b096c42ff3a907fa7bcce3fa620209da16d40bd/sphinx/util/logging.py#L487-L490

Please remove it from this list.

keewis:
done. I also changed the link to the official spec to its new location (the old one is just a in-text redirect).

sphinx/ext/napoleon/docstring.py
tk0miya: The document of `napoleon_use_param` only says how the options are represented (displayed), and does not say the types are not hyperlinked.
https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#confval-napoleon_use_param

So I think it's an improvement of `napleon_use_param = False` case. It's okay to change test cases.

tk0miya:
>I think that's because they didn't update the docstring guide in quite a while. I'll try submitting a PR changing the docstring guide (the official spec).

I think it should be documented to upstream's doc if they have special meaning for the colon after "default" keyword. So it sounds nice :-)

keewis:
> if they have special meaning for the colon after "default" keyword

the colon doesn't have a special meaning, it's completely optional. `numpydoc` accepts both formats (without mentioning any of them in the docstring guide), so I adapted this PR (see 9b425606e7be92c048e1b08b41562c296fadc56c) to do the same.

doc/usage/extensions/napoleon.rst
```diff
@@ -480,3 +481,31 @@ sure that "sphinx.ext.napoleon" is enabled in `conf.py`::
...
+           Description of `arg1`
+       arg2 : dict-like
+           Description of `arg2`
+
+   becomes::
+
+       :param arg1: Description of `arg1`
+       :type arg1: mypackage.CustomType
+       :param arg2: Description of `arg2`
+       :type arg2: :term:`dict-like <mapping>`
```
tk0miya, line 511: `.. versionadded:: 3.2` tag is needed at the end of description.

sphinx/ext/napoleon/docstring.py
keewis: well, if `napoleon_use_param = False`, the type spec is wrapped in `*{}*`, so I guess there's more to do to support that. Let's leave that to a new PR?

tk0miya (approved):

doc/usage/extensions/napoleon.rst
tk0miya: I'll update this after merging.

tk0miya:
Just Merged. Thank you for your great work!

embray:
Thanks for this--much more sophisticated and complete than my hackish workaround.

PR #7695, opened 2020-05-20
Add viewport meta tag for responsive layout

obulat:
Subject: add meta viewport tag to start creating responsive layout

### Feature or Bugfix

- Feature

### Purpose
- Viewport meta tag is necessary to make layout responsive. Without it, the layout is rendered for desktop size and then scaled down, making text small and difficult to read. It also makes a site fail Google Lighthouse audits.


tk0miya:
Thanks!

PR #7699, opened 2020-05-20
BUG: Allow extensions to pass Sphinx errors during init

larsoner:
In `sphinx-gallery` #7683 causes some problems for us because we used to check that things like `ValueError` were raised during builder init, but now it's always `ExtensionError` with the message "Handler <function generate_gallery_rst at 0x7f88327f69d0> for event 'builder-inited' threw an exception", and the useful error message higher up.

One solution is that we can switch to emitting `ExtensionError` and `ConfigError` as appropriate (I have tested this and it works). Then Sphinx needs the change here, and things work as expected.

Let me know if there is some other solution. It's possible that there are other events that get emitted that might need similar changes.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -33,7 +33,7 @@ from sphinx.deprecation import RemovedInSphinx40Warning
 from sphinx.domains import Domain, Index
 from sphinx.environment import BuildEnvironment
 from sphinx.environment.collectors import EnvironmentCollector
-from sphinx.errors import ApplicationError, ConfigError, VersionRequirementError
+from sphinx.errors import ApplicationError, ConfigError, VersionRequirementError, SphinxError
 from sphinx.events import EventManager
 from sphinx.extension import Extension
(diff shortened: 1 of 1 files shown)


larsoner:
... another option would be to add a new `config` value `allowed_exceptions` where you specify which exceptions to pass through, and then each `self.emit` or `self.events.emit` would use `allowed_exceptions=self.allowed_exceptions` or so. But again I'm not sure if this makes much sense.

eric-wieser:
Is https://github.com/sphinx-doc/sphinx/pull/7658 related in any way here?

tk0miya:
Sorry for bothering you. I don't know your extension raises a `ValueError` on the handler.

BTW, I think current implementation accept `SphinxError` without `allowed_exceptions`. So it would go well with replacing `ValueError` with `SphinxError` family now.

larsoner:
Indeed it does! Sorry for the noise

PR #7713, opened 2020-05-23
Fix #3673: autodoc: bysource order does not work for a module having __all__

### LINKED ISSUE #3673: `autodoc_member_order = 'bysource'` does not work for modules containing __all__

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #3673

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -857,6 +857,7 @@ class ModuleDocumenter(Documenter):
     def __init__(self, *args: Any) -> None:
         super().__init__(*args)
         merge_special_members_option(self.options)
+        self.__all__ = None
 
     @classmethod
     def can_document_member(cls, member: Any, membername: str, isattr: bool, parent: Any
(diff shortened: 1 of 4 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -714,38 +714,43 @@ def document_members(self, all_members: bool = False) -> None:
...
+        for documenter, isattr in memberdocumenters:
+            documenter.generate(
+                all_members=True, real_modname=self.real_modname,
+                check_module=members_check_module and not isattr)
+
+        # reset current objects
+        self.env.temp_data['autodoc:module'] = None
+        self.env.temp_data['autodoc:class'] = None
+
+    def sort_members(self, documenters: List[Tuple["Documenter", bool]],
```
tk0miya, line 730: I hesitated to add a new method to the Documenter class. But it is needed to override the sorting method from subclasses...

PR #7716, opened 2020-05-24
Close #3610: autodoc: Support overloaded functions

### LINKED ISSUE #3610: Support for @overload in Python

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #3610 

### Remaining tasks

- [x] Support overloaded `__new__()`, `__init__()` and `__call__()` methods.


sphinx/ext/autodoc/__init__.py
```diff
@@ -1543,6 +1547,14 @@ def format_signature(self, **kwargs: Any) -> str:
                     documenter.object = func
                     documenter.objpath = [None]
                     sigs.append(documenter.format_signature())
+        if self.analyzer and '.'.join(self.objpath) in self.analyzer.overloads:
+            for overload in self.analyzer.overloads.get('.'.join(self.objpath)):
```
eric-wieser: No need to do the lookup twice:
```suggestion
        if self.analyzer:
            for overload in self.analyzer.overloads.get('.'.join(self.objpath), []):
```

tests/test_ext_autodoc.py
```diff
@@ -1679,6 +1679,41 @@ def test_final(app):
...
+        '.. py:class:: Math',
+        '   :module: target.overload',
+        '',
+        '   docstring',
+        '',
+        '',
+        '   .. py:method:: Math.sum(x, y)',
+        '                  Math.sum(x: int, y: int) -> int',
+        '                  Math.sum(x: float, y: float) -> float',
+        '                  Math.sum(x: str, y: str) -> str',
```
eric-wieser: I'd argue the desired output is
```suggestion
        '   .. py:method:: Math.sum(x: int, y: int) -> int',
        '                  Math.sum(x: float, y: float) -> float',
        '                  Math.sum(x: str, y: str) -> str',
```

tk0miya: We can give type annotations to the main function.
```
@overload
def sum(x: int, y: int) -> int:
   ...

@overload
def sum(x: str, y: str) -> str:
   ...

def sum(x: Any, y: Any) -> Any:
    return x + y
```

I think showing `(x: Any, y: Any) -> Any` is good. What do you think?

eric-wieser: The example for [`typing.overload`](https://docs.python.org/3/library/typing.html#typing.overload) does not annotate the main function. Nor does [the PEP](https://www.python.org/dev/peps/pep-0484/#function-method-overloading).

eric-wieser: Ah, [mypy does](https://mypy.readthedocs.io/en/stable/more_types.html), and says that

> It may or may not have type hints. If it does, mypy
> will check the body of the implementation against the
> type hints.
>
> Mypy will also check and make sure the signature is
> consistent with the provided variants.

This sounds like an implementation helper to me, and not something that is actually interesting to a user reading the docs.

I'd therefore lean towards either:

1. Not showing the signature, as I suggest above
2. Providing an option (default off) to show the main signature

Probably easier to do 1, and wait to see if anyone asks for 2 :)

tk0miya: Acceptable. I'll change to 1.

sphinx/util/inspect.py
```diff
@@ -557,6 +557,45 @@ def signature_from_str(signature: str) -> inspect.Signature:
...
+        for arg in node.args.posonlyargs:  # type: ignore
+            param = Parameter(arg.arg, Parameter.POSITIONAL_ONLY,
+                              annotation=ast_unparse(arg.annotation))
+            params.append(param)
+
+    for i, arg in enumerate(node.args.args):
+        if len(node.args.args) - i <= len(node.args.defaults):
+            default = ast_unparse(node.args.defaults[-len(node.args.args) + i])
+        else:
+            default = Parameter.empty
```
eric-wieser: Possibly simpler as:
```suggestion
    defaults = [Parameter.empty] * (len(node.args.args) - len(node.args.defaults)) + node.args.defaults
    for arg, default in enumerate(node.args.args, defaults):
```

sphinx/util/inspect.py
```diff
@@ -557,6 +557,45 @@ def signature_from_str(signature: str) -> inspect.Signature:
...
 
 
+def signature_from_ast(node: ast.FunctionDef) -> inspect.Signature:
+    """Return a Signature object for the given AST *node*."""
+    params = []
+    if hasattr(node.args, "posonlyargs"):  # for py38+
+        for arg in node.args.posonlyargs:  # type: ignore
+            param = Parameter(arg.arg, Parameter.POSITIONAL_ONLY,
+                              annotation=ast_unparse(arg.annotation))
+            params.append(param)
```
eric-wieser: Positional only arguments can have defaults too

tk0miya: OMG, this comes from another bug...
https://github.com/sphinx-doc/sphinx/issues/7756

sphinx/util/inspect.py
```diff
@@ -557,6 +557,45 @@ def signature_from_str(signature: str) -> inspect.Signature:
     return inspect.Signature(params, return_annotation=return_annotation)
 
 
+def signature_from_ast(node: ast.FunctionDef) -> inspect.Signature:
```
tk0miya: I noticed there are duplicated codes in `signature_from_str()` and this. So it should be merged.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1710,6 +1726,14 @@ def format_signature(self, **kwargs: Any) -> str:
...
                     documenter.objpath = [None]
                     sigs.append(documenter.format_signature())
+        if overloaded:
+            for overload in self.analyzer.overloads.get('.'.join(self.objpath)):
+                if not inspect.isstaticmethod(self.object, cls=self.parent,
+                                              name=self.object_name):
+                    parameters = list(overload.parameters.values())
+                    overload = overload.replace(parameters=parameters[1:])
+                sig = stringify_signature(overload, **kwargs)
+                sigs.append(sig)
```
eric-wieser, line 1766: Could eliminate the `if` if you wanted to.
```suggestion
        for overload in self.analyzer.overloads.get('.'.join(self.objpath), []):
            if not inspect.isstaticmethod(self.object, cls=self.parent,
                                          name=self.object_name):
                parameters = list(overload.parameters.values())
                overload = overload.replace(parameters=parameters[1:])
            sig = stringify_signature(overload, **kwargs)
            sigs.append(sig)
```

eric-wieser (approved):
I don't have the background to review the analyzer, but this approach generally looks very reasonable, and might be worth merging even if `__init__` and `__new__` are not yet supported.

sphinx/ext/autodoc/__init__.py
tk0miya: No, it would be crashed with C modules. Sometimes `self.analyzer` becomes None. The `overloaded` is used to skip the check.

tk0miya:
Now I added overload-support for class constructors. But I can't find the way to do that on metaclasses (`__call__`).

@eric-wieser Do you know a way to obtain the metaclass from concrete class? Either way, it is better to do it on other PR.

eric-wieser:
>  Do you know a way to obtain the metaclass from concrete class?

`type(ConcreteClass)`

sphinx/ext/autodoc/__init__.py
```diff
@@ -1341,6 +1356,7 @@ def get_user_defined_function_or_method(obj: Any, attr: str) -> Any:
         # the signature from, so just pass the object itself to our hook.
         self.env.app.emit('autodoc-before-process-signature', self.object, False)
         try:
+            self._signature_method = None
             return inspect.signature(self.object, bound_method=False)
```
eric-wieser: It's a bit ugly how we can't just return it here. Perhaps we should allow `format_args` to return multiple lines as well, and then we would have access to the return value?

Or as a middle ground, change this to
```suggestion
            return None, inspect.signature(self.object, bound_method=False)
```
and change the line below as
```diff
-sig = self._get_signature()
+sig_method, sig = self._get_signature()
+self._signature_method = sig_method  # TODO: get this to `format_signature` in a tidier way
```

tests/roots/test-ext-autodoc/target/overload.py
```diff
@@ -39,3 +39,33 @@ def sum(self, x: str, y: str) -> str:
...
+    @overload
+    def __init__(cls, x: int, y: int) -> None:
+        ...
+
+    @overload
+    def __init__(cls, x: str, y: str) -> None:
+        ...
+
+    def __init__(cls, x, y):
+        pass
```
eric-wieser, line 71:
```suggestion
        pass

class _BazMeta(type):
    @overload
    def __call__(cls, x: int, y: int) -> None:

    @overload
    def __call__(cls, x: str, y: str) -> None:

    def __call__(cls, x, y) -> None:
        pass


class Baz(metaclass=_BazMeta):
    pass
```

tk0miya:
Thank you for the info. I implemented `__call__` support.

tests/roots/test-ext-autodoc/target/overload.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class Meta(type):
```
eric-wieser, line 74: My suggestion deliberately had this as `_Meta` so that the documentation for this type itself isn't shown. I think you either need to add a `_` here, or update the test output to expect `Meta` too, unless I'm missing something.

eric-wieser: Ah, classes with no docstring are not shown, that solves it.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1370,7 +1386,36 @@ def format_signature(self, **kwargs: Any) -> str:
...
+        sig = super().format_signature(**kwargs)
+        sigs = []
+        overloaded = False
+        fullqualname = None
+        if self._signature_method and self.analyzer:
+            if self._signature_method == '__call__':
+                metaclass = type(self.object)
+                fullqualname = '.'.join([metaclass.__qualname__, '__call__'])
+            else:
+                fullqualname = '.'.join(self.objpath + [self._signature_method])
```
eric-wieser: Would it be better to store `signature_qualname` instead of `signature_method`? Then you wouldn't have this metaclass logic in two places.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1313,6 +1325,7 @@ def get_user_defined_function_or_method(obj: Any, attr: str) -> Any:
         if call is not None:
             self.env.app.emit('autodoc-before-process-signature', call, True)
             try:
+                self._signature_method = '__call__'
```
eric-wieser: As suggested in https://github.com/sphinx-doc/sphinx/pull/7716#discussion_r433282055:
```suggestion
                self._signature_qualname = type(self.object).__qualname__ + '.__call__'
```

tests/roots/test-ext-autodoc/target/overload.py
tk0miya: Yes, this is not documented.

sphinx/ext/autodoc/__init__.py
tk0miya: Sounds good to change retval of `_get_signature()`. I'm okay if we don't change interface of public method (`format_signature()` and `format_args()`). But it causes `_get_signature()` is called twice in both `format_signature()` and `format_args()`.
```
def format_signature(self, **kwargs):
    sig_method, sig = self._get_signature()
    if sig_method and self.analyzer:
        ...
    else:
        overloaded = False
        sig = super().format_signature()

   ...
```

tk0miya: Note: it would be much better to return a tuple of the target class (self.object or metaclass), method name and signature string.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1370,7 +1384,29 @@ def format_signature(self, **kwargs: Any) -> str:
         if self.doc_as_attr:
             return ''
 
-        return super().format_signature(**kwargs)
+        overloaded = False
+        qualname = None
+        cls, method_name, _ = self._get_signature()
+        # TODO: recreate analyzer for the module of class (To be clear, owner of the method)
```
tk0miya, line 1394: `self.analyzer` only loads overloads mapping for the module of current class. But the `__init__`, `__new__` or `__call__` method might be implemented at other module. So we need to recreate an analyzer for the module where the method actually implemented.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1341,20 +1354,21 @@ def get_user_defined_function_or_method(obj: Any, attr: str) -> Any:
...
-        return None
+        return None, None, None
 
     def format_args(self, **kwargs: Any) -> str:
         if self.env.config.autodoc_typehints in ('none', 'description'):
             kwargs.setdefault('show_annotation', False)
 
         try:
-            sig = self._get_signature()
+            _, _, sig = self._get_signature()
```
eric-wieser: What I was proposing earlier was keeping the hack of attaching attributes to self, but moving it out of `_get_signature` and into here.

The problem with calling `_get_signature` twice is that the user's hooks get called twice.

tk0miya: oops. I have to change it again.

tk0miya:
Merged. @eric-wieser Thank you for your help!

PR #7717, opened 2020-05-24
basic CSS: avoid unnecessary padding to the left of line numbers

mgeier:
This is a follow-up to #7482.

I think it's very good to have some space between line numbers and the code block, but I don't think there needs to be additional empty space to the *left* of the line numbers.


tk0miya:
LGTM!

PR #7718, opened 2020-05-24
CSS: Respect Pygments background color

mgeier:
Most built-in themes set the background color of code blocks in some way, overriding the background color provided by Pygments. This means that Pygments styles with bright text colors will not be readable on the (typically bright) hard-coded background.

This PR removes the background color from all built-in themes except:

* `classic`, where there are custom theme options for code text and background colors [see ~#7720~ #7721]
* `scrolls`, which has a custom image as background


tk0miya:
👍

PR #7720, opened 2020-05-24
classic theme: remove codetextcolor/codebgcolor theme options

mgeier:
Since the config option `pygments_style` is supposed to take care of the code text and background colors, it doesn't make much sense to have those options a theme options.

However, removing them would be a breaking change for everyone who customizes those settings in their own `html_theme_options`.

Therefore, you probably don't want to merge this PR.

### DIFF
--- a/doc/usage/theming.rst
+++ b/doc/usage/theming.rst
@@ -219,9 +219,6 @@ These themes are:
   - **headbgcolor** (CSS color): Background color for headings.
   - **headtextcolor** (CSS color): Text color for headings.
   - **headlinkcolor** (CSS color): Link color for headings.
-  - **codebgcolor** (CSS color): Background color for code blocks.
-  - **codetextcolor** (CSS color): Default text color for code blocks, if not
-    set differently by the highlighting style.
 
   - **bodyfont** (CSS font-family): Font for normal text.
   - **headfont** (CSS font-family): Font for headings.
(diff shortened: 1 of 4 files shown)


doc/usage/theming.rst
```diff
@@ -219,9 +219,6 @@ These themes are:
   - **headbgcolor** (CSS color): Background color for headings.
   - **headtextcolor** (CSS color): Text color for headings.
   - **headlinkcolor** (CSS color): Link color for headings.
-  - **codebgcolor** (CSS color): Background color for code blocks.
-  - **codetextcolor** (CSS color): Default text color for code blocks, if not
-    set differently by the highlighting style.
```
mgeier, line 224: I don't think this sentence is correct, it seems like the highlighting style is ignored.

mgeier:
OK, I've found a much better way that obsoletes this PR: #7721.

doc/usage/theming.rst
mgeier: It turns out that this sentence is indeed correct, but there is one case where it is misleading:

When using the `'none'` language, Pygments doesn't set a text color (but it *does* set a background color).

Therefore, when using a dark background (e.g. `pygments_style = 'monokai'`) and using a dark color for `codetextcolor`, the text on `'none'`-language code blocks will be not readable.

Anyway, if the default value for `codetextcolor` is changed to `'unset'` (as suggested in #7721), this will not be a problem anymore.

PR #7689, opened 2020-05-17
bizstyle&sphinxdoc themes: Fix float behavior

mgeier:
### Feature or Bugfix
- Bugfix

### Purpose

- fixes #7687
- fixes #7688

I just copied the CSS from the `classic` theme:

https://github.com/sphinx-doc/sphinx/blob/efc1c5ef081fe0003624816a463dc7b68e78deb0/sphinx/themes/classic/static/classic.css_t#L34-L37

### sphinxdoc: before

image

### sphinxdoc: after

image

### bizstyle: before

image

image

### bizstyle: after

image
image

PR #7733, opened 2020-05-26
Make sphinx13 a re-usable theme

mgeier:
I find it strange that the Sphinx docs don't use one of Sphinx's own themes.

So I tried the convert the ad-hoc `sphinx13` theme (which was living in `doc/_themes`) into a re-usable stand-alone theme (now living in `sphinx/themes` with all the other themes).

I have marked this PR as "draft" because there is no documentation yet.

If this PR is considered for merging, I can add documentation (and any other things that are missing).


sphinx/themes/sphinx13/layout.html
```diff
@@ -1,8 +1,8 @@
 {#
-    sphinxdoc/layout.html
+    sphinx13/layout.html
```
tk0miya, line 2: I feel strange for the name "sphinx13". It should be renamed if we make it public. But we've already used "sphinxdoc"...

tk0miya (commented):
+0 for your proposal.

sphinx/themes/sphinx13/layout.html
mgeier: I don't care about the name.
I was just trying to make a minimal change and fix obvious errors.
If you like to change the name, feel free to do so.
If you like me to change the name, please tell me the name you want to have.

mgeier:
I'm not planning on ever using this theme, so I don't really care whether it is available for common use.

But I don't see any reason why *not* to make it re-usable.
Is there any reason?

I think it may shine a negative light on Sphinx, because it may be perceived as if the Sphinx team thinks that none of the built-in themes are good enough for them.

The main reason, however, why I made this PR, is for ease of maintenance. I was looking at all built-in themes and made #7689 and #7718 to fix some problems. Since the internal `sphinx13` theme is not part of the available themes, I didn't make those fixes there. If it were an "official" theme, it would have been easy to apply those fixes, too.

tk0miya:
I did not object to remake it re-usable. Only my concern is its naming. I'm okay for not renaming it if we update it as a private theme. But we need to rename it to an appropriate one if we make it public. But I have no idea because "sphinxdoc" has been already used. It might be better to rename "sphinxdoc" to "oldsphinxdoc" and rename "sphinx13" to "sphinxdoc"...

mgeier:
I don't think switching the theme associated with the name `sphinxdoc` is a good idea.
For anyone who is using this theme, a Sphinx update would suddenly and unexpectedly switch their theme. Wouldn't this be bad?

It looks like the name `sphinx13` was introduced in 0939d6fcdb99838ed0860f87dcae5b2fd6d1f524 by @birkenfeld. Maybe he has an idea for a new name?

tk0miya:
Indeed. I think "sphinxdoc" is a good new name for "sphinx13". But it has already been in use. So I don't have idea... sphinxdoc2? sphinxdoc-ng?

I think making something public is like a contract. It is difficult to change it. So I'm afraid of use meaningless name for new theme.

mgeier:
What about `sphinxdoc3000`?

shimizukawa:
-1 to making `sphinx13` a part of builtin themes. I've used the same idea before, separating the theme from the site and making it reusable. That increased my maintenance costs.

There is an advantage to not making it public. It means that a small modification or destructive change won't affect it anywhere. This keeps maintenance costs low for maintainers.

I'm sure there would be benefits to be gained by incorporating it into standard themes. However, if we make it a part of standard themes, we'll also have to adapt to new requests, which will increase maintainers costs.

My suggestion is that someone should cut out the `sphinx13` as a 3rd party package and maintain it separately from the Sphinx itself.

Issue #7000, opened 2020-01-08
Support for files outside Sphinx project (README.md in parent dir)


NumesSanguis:
**Is your feature request related to a problem? Please describe.**
Many Python repositories have a structure similar to:
```
project/
  docs/
    conf.py
    index.rst
  src/
  README.md
```
This README is perfect for writing the project's introduction and some minimal usage examples. GitHub also renders these README's on the landing page (not when in the docs folder). This README is often in the format of Markdown, because in multi-contributer projects, this is the most familiar format.

Good documentation includes an introduction, which the README is perfect for, but currently you cannot include this README in the toctree, because it is outside the `docs/` folder.

**Describe the solution you'd like**
One of the following:

1. Allow toctree parsing of source files outside the Sphinx project (which allows the extension [`recommonmark`](https://recommonmark.readthedocs.io/en/latest/) to parse the Markdown formatted README.md)
2. Allow passing on of a `.md` file to an extension before trying to parse the file as reST with the `.. include:: ../README.md` statement.

**Describe alternatives you've considered**
1. Use [`.. include:: ../README.md` in a `readme_link.rst`](https://stackoverflow.com/a/17217041) file (also mentioned in issue: https://github.com/sphinx-doc/sphinx/issues/2840).
    **Problem**: Parses the `.md` file as reST
2. Add [paths to `sys.path` in `conf.py` like: `sys.path.insert(0, os.path.abspath('..'))`](https://stackoverflow.com/a/45953127)
  **Problem**: Somehow this doesn't work for me
3. Use a symlink to the `README.md`
  **Problem**: [Is not cross-platform (does not work on Windows)](https://github.com/sphinx-doc/sphinx/issues/701#issuecomment-280000495)
4. Copy the file on `make` time with a function.
  **Problem:** Far from ideal, needlessly making files.
5. Use [M2R](https://github.com/miyakogi/m2r) with `.. mdinclude:: ../README.md`; **Only option working for me**.
  **Problem**: Using `extensions = ['m2r', 'recommonmark']` (or reversed list) gives the error: [`source_suffix '.md' is already registered` in the `m2r` library](https://github.com/miyakogi/m2r/issues/51).

**Additional context**
- Similar issue can be found here: https://github.com/sphinx-doc/sphinx/issues/2840 (but the difference is that this solution doesn't request Markdown parsing from Sphinx's side with `.. include::`, only to give extensions an interface to do so).
[...]

PR #2354, opened 2016-02-25
Allow explicit markup in toctree directives

nwf:
This adds support for :ref:`` text in toctree listings in particular, but is
reasonably generic.  My website, for example, now contains something like

```
.. toctree::
   _ :ref:`foo <bar>`
   baz
```

which renders as one might expect.

Issue #701, opened 2015-01-02
relative links in toctree


shimizukawa:
I'd like to use relative URLs in toctrees, but can't figure out how.
It seems right now I have the choice between either internal links (where the target is another document), or an (absolute) URL. The latter is identified by means of a sphinx.util.url_re.

Is there a way to have a toctree entry use a relative link, such that the generated html contains 'href=".../some/path.html"' ?

(If it isn't yet supported, perhaps a new pseudo-schema would work, such as 'relative:.../some/path.html', from which the 'relative:' would then be stripped off during html generation ?)

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/701
- Originally reported by: Anonymous
- Originally created at: 2011-05-22T03:17:53.521

shimizukawa:
_From Anonymous on 2012-04-18 14:10:29+00:00_

I would also like to use relative document locations that are not within the directory tree housing conf.py.  For instance, see my stack overflow question:

http://stackoverflow.com/questions/10199233/can-sphinx-link-to-documents-that-are-not-located-in-directories-below-the-root

shimizukawa:
_From [rdburns](https://bitbucket.org/rdburns) on 2012-04-18 14:12:13+00:00_

That previous comment was me, forgot to login.

shimizukawa:
_From Mac Ryan on 2012-08-02 10:26:08+00:00_

I second this request (just in case the number of people requesting this feature influences the priority queue...). :)

shimizukawa:
_From Dmitriy Kruglov on 2012-09-13 09:35:01+00:00_

One more vote to have the feature.

shimizukawa:
_From Marcus Lindblom Sonestedt on 2012-10-17 10:50:22+00:00_

We'd like this too! :)

shimizukawa:
_From Gringo Suave on 2012-12-14 03:35:35+00:00_

Bitbucket demands I keep my readme in the root, sphinx demands I keep it in ./docs ... what to do?  I'd use a filesystem link but project needs to be able to work under Windows and hg doesn't support links there.  
I'll have to keep two copies in the repo unless anyone has a better idea.   :(

shimizukawa:
_From Thibault Kruse on 2013-01-30 12:25:54+00:00_

+1

shimizukawa:
_From Georg Brandl on 2013-01-31 07:37:34+00:00_

[...]

PR #7739, opened 2020-05-28
Add an include directive that allows one filetype to be included in another

eric-wieser:
### Feature or Bugfix

- Feature

### Purpose
- Primarily to allow markdown files to be included within rst files (like `m2r`'s `.. mdinclude`
- Works for all other input parsers too

Upstreaming from https://github.com/pygae/galgebra/pull/413, where this was very useful

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -708,6 +708,21 @@ __ http://pygments.org/docs/lexers
    .. versionchanged:: 2.1
       Added the ``force`` option.
 
+
+.. rst:directive:: .. docinclude:: filename
+
+   Includes another document in this one. Unlike :directive:`include`, this
+   allows inclusion of documents that use different parsers to their parent
+   document, such as including a markdown file in an rst file.
+
+   .. rst:directive:option:: start-line
+                             end-line
+                             start-after
+                             end-after
+                             encoding
+
+      Options treated the same way as the :rst:dir:`include` directive.
+
 .. _glossary-directive:
 
 Glossary
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -711,7 +711,7 @@ __ http://pygments.org/docs/lexers
 
 .. rst:directive:: .. docinclude:: filename
 
-   Includes another document in this one. Unlike :directive:`include`, this
+   Includes another document in this one. Unlike :rst:dir:`include`, this
    allows inclusion of documents that use different parsers to their parent
    document, such as including a markdown file in an rst file.
 
--- a/sphinx/directives/other.py
+++ b/sphinx/directives/other.py
@@ -393,7 +393,6 @@ class IncludeDocument(SphinxDirective):
             path = os.path.join(self.standard_include_path, path[1:-1])
         path = os.path.normpath(os.path.join(source_dir, path))
         path = relative_path(None, path)
-        path = nodes.reprunicode(path)
         encoding = self.options.get(
             'encoding', self.state.document.settings.input_encoding)
         e_handler = self.state.document.settings.input_encoding_error_handler
--- a/sphinx/directives/other.py
+++ b/sphinx/directives/other.py
@@ -6,22 +6,25 @@
     :license: BSD, see LICENSE for details.
 """
 
+import os
 import re
 from typing import Any, Dict, List
 from typing import cast
 
-from docutils import nodes
+from docutils import nodes, io
 from docutils.nodes import Element, Node
 from docutils.parsers.rst import directives
 from docutils.parsers.rst.directives.admonitions import BaseAdmonition
 from docutils.parsers.rst.directives.misc import Class
 from docutils.parsers.rst.directives.misc import Include as BaseInclude
+from docutils.utils import new_document, relative_path
+from docutils.utils.error_reporting import SafeString, ErrorString
 
 from sphinx import addnodes
 from sphinx.deprecation import RemovedInSphinx40Warning, deprecated_alias
 from sphinx.domains.changeset import VersionChange  # NOQA  # for compatibility
 from sphinx.locale import _
-from sphinx.util import url_re, docname_join
+from sphinx.util import url_re, docname_join, get_filetype
 from sphinx.util.docutils import SphinxDirective
 from sphinx.util.matching import Matcher, patfilter
 from sphinx.util.nodes import explicit_title_re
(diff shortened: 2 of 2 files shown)


sphinx/directives/other.py
```diff
@@ -383,6 +472,7 @@ def setup(app: "Sphinx") -> Dict[str, Any]:
     directives.register_directive('hlist', HList)
     directives.register_directive('only', Only)
     directives.register_directive('include', Include)
+    directives.register_directive('includedoc', IncludeDocument)
```
eric-wieser, line 474: This name is bad, would appreciate feedback on a better name

sphinx/directives/other.py
```diff
@@ -361,6 +364,93 @@ def run(self) -> List[Node]:
...
+            # skip content in rawtext after *and incl.* a matching text
+            before_index = rawtext.find(before_text)
+            if before_index < 0:
+                raise self.severe('Problem with "end-before" option of "%s" '
+                                  'directive:\nText not found.' % self.name)
+            rawtext = rawtext[:before_index]
+    
+        # copied code ends
+        app = self.env.app
+        filetype = get_filetype(app.config.source_suffix, path)
```
eric-wieser, line 444: Probably makes sense to provide an option to overload this, for files with weird extensions like `.inc` or `.md.inc` etc, something like
```python
        filetype = self.options.get('filetype', None)
        if filetype is None:
            filetype = get_filetype(app.config.source_suffix, path))
```

tk0miya:
-1: I don't know why such feature is needed. We already have the toctree feature to build multiple documents into the one project. In addition, merging other document is much complicated for both developers and users (path, warning, and so on).

doc/usage/restructuredtext/directives.rst
```diff
@@ -708,6 +708,21 @@ __ http://pygments.org/docs/lexers
...
+.. rst:directive:: .. docinclude:: filename
+
+   Includes another document in this one. Unlike :directive:`include`, this
+   allows inclusion of documents that use different parsers to their parent
+   document, such as including a markdown file in an rst file.
+
+   .. rst:directive:option:: start-line
+                             end-line
+                             start-after
+                             end-after
```
tk0miya, line 721: Are these filtering options needed? In my experience of `literalinclude` directive, many filter options are requested. So I hesitate to add filtering options.

sphinx/directives/other.py
```diff
@@ -361,6 +364,92 @@ def run(self) -> List[Node]:
...
+            self.lineno - self.state_machine.input_offset - 1)
+        source_dir = os.path.dirname(os.path.abspath(source))
+        path = directives.path(self.arguments[0])
+        if path.startswith('<') and path.endswith('>'):
+            path = os.path.join(self.standard_include_path, path[1:-1])
+        path = os.path.normpath(os.path.join(source_dir, path))
+        path = relative_path(None, path)
+        path = nodes.reprunicode(path)
+        encoding = self.options.get(
+            'encoding', self.state.document.settings.input_encoding)
```
tk0miya, line 397: `config.source_encoding` is better for the default value.

sphinx/directives/other.py
```diff
@@ -361,6 +364,92 @@ def run(self) -> List[Node]:
...
+        e_handler = self.state.document.settings.input_encoding_error_handler
+        try:
+            self.state.document.settings.record_dependencies.add(path)
+            include_file = io.FileInput(source_path=path,
+                                        encoding=encoding,
+                                        error_handler=e_handler)
+        except UnicodeEncodeError:
+            raise self.severe(u'Problems with "%s" directive path:\n'
```
tk0miya, line 405: u prefix is no longer needed, right?

sphinx/directives/other.py
tk0miya: +0: But nobody does not know about filetype. So we have to add it to our document.

sphinx/directives/other.py
```diff
@@ -361,6 +364,92 @@ def run(self) -> List[Node]:
...
+                                  'directive:\nText not found.' % self.name)
+            rawtext = rawtext[:before_index]
+
+        # copied code ends
+        app = self.env.app
+        filetype = get_filetype(app.config.source_suffix, path)
+        parser = app.registry.create_source_parser(app, filetype)
+
+        sub_document = new_document(path, self.state.document.settings)
+        parser.parse(rawtext, sub_document)
```
tk0miya, line 448: Why do you generate new document here? It prevents to share local reference context (ex. footnotes, hyperref, etc.).

sphinx/directives/other.py
```diff
@@ -361,6 +364,92 @@ def run(self) -> List[Node]:
...
+            raise self.warning('"%s" directive disabled.' % self.name)
+        source = self.state_machine.input_lines.source(
+            self.lineno - self.state_machine.input_offset - 1)
+        source_dir = os.path.dirname(os.path.abspath(source))
+        path = directives.path(self.arguments[0])
+        if path.startswith('<') and path.endswith('>'):
+            path = os.path.join(self.standard_include_path, path[1:-1])
+        path = os.path.normpath(os.path.join(source_dir, path))
+        path = relative_path(None, path)
+        path = nodes.reprunicode(path)
```
tk0miya: `reprunicode` is an alias of `str`. It is meaningless on py3.

doc/usage/restructuredtext/directives.rst
eric-wieser: Yes, see https://github.com/pygae/galgebra/pull/413 for a useful application. It allows me to split a `README.md` into pieces, and include each one in a different rst document.

These filters are part of the built-in `.. include`, so it seems reasonable to keep them anyway.

sphinx/directives/other.py
eric-wieser: This code is copied from the built-in `..include`

sphinx/directives/other.py
eric-wieser: Indeed - it's here because this code is copied directly from docutils

sphinx/directives/other.py
eric-wieser: Is there another way to invoke the parser?

I figured it didn't matter, because if the included document is in a different file it's very unlikely to reuse footnotes.

sphinx/directives/other.py
eric-wieser: Again, copied verbatim from docutils

eric-wieser: Removed anyway because mypy does not like it.

sphinx/directives/other.py
tk0miya: >I figured it didn't matter, because if the included document is in a different file it's very unlikely to reuse footnotes.

The document object is large database for the doctree. So all references would be broken. Is it okay to conflict node_IDs between parent and included document? I guess `:ref:` role for included document may not work also.

>Is there another way to invoke the parser?

How about `parser.parse(rawtext, self.state.document)` instead?

doc/usage/restructuredtext/directives.rst
```diff
@@ -708,6 +708,21 @@ __ http://pygments.org/docs/lexers
    .. versionchanged:: 2.1
       Added the ``force`` option.
 
+
+.. rst:directive:: .. docinclude:: filename
+
+   Includes another document in this one. Unlike :directive:`include`, this
```
eric-wieser:
```suggestion
   Includes another document in this one. Unlike :rst:dir:`include`, this
```

sphinx/directives/other.py
eric-wieser:
```suggestion
```

sphinx/directives/other.py
eric-wieser: Does docutils allow a document to be parsed into multiple times? How does it know where the current insertion point is?

sphinx/directives/other.py
rpanderson: Alternative directives that better distinguish it from `include` might be: `parse`, `includeother`, `includeparsed`, etc.

RabidCicada:
Holy cow @eric-wieser .  Thankyou for your work on this.  

I currently have a code base that is heavily locally documented in each subdirectory with README.md's about each component authored by developers that are much more comfortable with Markdown.  

I have crafted my system to parse rst, md, plantuml, and auto harvest C++ API docs using doxygen.  After following #701 and looking for a good solution...I've settled on replicating my entire folder structure under `docs` and using softlinks so that relative linking works both natively on gitlab/github with the Markdown documents but also when built by sphinx.

While this does not solve that problem I appreciate your work in contributing something that would be very useful.

RabidCicada:
This might partially solve relative linking instead of relying on softlinks.  So then it would work on Windows.  Any thoughts about if this could take the place of softlinks.  I think your `.. includeDoc::` in README.rst's with recommonmark could replace my softlink approach.

```
docs/
    src/
        folder1/
            README.md <softlink>
        folder2/
            README.md <softlink>
        folder3/
            folder4/
                README.md <softlink>
        README.md <softlink>
        index.rst
        conf.py
    build/
    Makefile
folder1/
    README.md
folder2/
    README.md
folder3/
    folder4/
        README.md
README.md
```

RabidCicada:
I also have looked at and followed a lot of the disussion here: #7000.  similar issues.  

I like your solution here as one that might actually be mergable since while @tk0miya has suggested he supports the idea all attempts so far have been met with only negative criticism of the solution and resistance to changing deep parts of sphinx ( #2354 and #7913).  Not that the criticism isn't warranted or that the guardianship of the code base isn't a good thing.

@tk0miya is there a chance that you can:

-  suggest a positive path forward for implementing a solution that will be acceptable to you for this MR?

Issue #7408, opened 2020-04-03
Clean autosummary generated files with make clean in make mode


vnmabus:
As far as I know, `make clean` in make mode only removes the `_build` directory.

I would want that files generated by other sphinx extensions were removed too. For some extensions this can be done placing the generated files under the `_build` directory. However, autosummary only accepts as directories for generated files relative paths (starting at the directory of the document being processed) or absolute paths starting at the root of the filesystem (which in my opinion is totally useless). There is no easy way to generate the files under `_build`.

I would want that either there were a (platform independent) way to specify additional cleaning in make clean, or that absolute paths in autosummary started at the directory of the conf.py file. Both alternatives are, of course, non-exclusive.

tk0miya:
Absolutely!

tk0miya:
I noticed this is a duplicated with #1999. So I'm closing this.

Issue #1999, opened 2015-08-13
cleaning all generated files


pinsardfr:
Hi

I am using Sphinx v1.3.1 with autosummary extension (just awesome by the way).

I am wondering if the fact that all generated directorties are not removed by the "make clean" is a feature and why ...

Here a real case
```
 $ make html
 ...
$ find . -name "_generated_"
./_build/doctrees/pypago/generated
./_build/html/_sources/pypago/generated
./_build/html/pypago/generated
./source/pypago/generated

$ make clean
Removing everything under '_build'...

$ find . -name "_generated_"
./source/pypago/generated
```

Should'nt this ./source/pypago/generated be also deleted or produced 
somewhere in _build directory (independant of a builder ?) instead ?

Regards

leotrs:
I have the same question.

luccareinehr:
There's an added difficulty to this: the directory where autosummary stores the generated files depends on the arguments passed to the directive when called.
For example,
`
.. autosummary::
  :toctree: generated

  module_name
`
will place the files in the `generated/` folder.

luccareinehr:
Also, it wouldn't be right to just delete all files with the same command as `make clean`. Since the user can edit every generated .rst, he may add text to these docs that would be lost unintentionally.

mgeier:
This might be related: #10641

MKuranowski:
I had just been sent on a goose chase trying to squash out "documents not referenced from toctree" errors due to leftover files from autosummary. I had been running `make clean && make html` thinking that I'm getting a build from scratch, but turns out that wasn't the case.

It's a really stupid interface that clean isn't a proper clean, it's a remove-some-random-arbitrary files "clean".

`clean` should undo **all** files run be previous `make` calls to ensure a new, clean, build.

PR #7742, opened 2020-05-28
Autosummary absolute paths are now relative to documentation dir.

vnmabus:
Make autosummary absolute paths (for the toctree directive) relative to the documentation folder.

### Feature or Bugfix
- Feature

### Purpose
The ultimate purpose of this feature is to allow the placement of the generated files in a separate directory. This is useful, for example, to clean the generated files with make clean (by placing them in the build directory).

### Relates
- #7408 #1999

### DIFF
--- a/sphinx/ext/autosummary/generate.py
+++ b/sphinx/ext/autosummary/generate.py
@@ -296,7 +296,7 @@ def generate_autosummary_docs(sources: List[str], output_dir: str = None,
     template = AutosummaryRenderer(app)
 
     # read
-    items = find_autosummary_in_files(sources)
+    items = find_autosummary_in_files(sources, app=app)
 
     # keep track of new files
     new_files = []
@@ -351,7 +351,7 @@ def generate_autosummary_docs(sources: List[str], output_dir: str = None,
 
 # -- Finding documented entries in files ---------------------------------------
 
-def find_autosummary_in_files(filenames: List[str]) -> List[AutosummaryEntry]:
+def find_autosummary_in_files(filenames: List[str], app: Any) -> List[AutosummaryEntry]:
     """Find out what items are documented in source/*.rst.
 
     See `find_autosummary_in_lines`.
@@ -360,12 +360,12 @@ def find_autosummary_in_files(filenames: List[str]) -> List[AutosummaryEntry]:
     for filename in filenames:
         with open(filename, encoding='utf-8', errors='ignore') as f:
             lines = f.read().splitlines()
-            documented.extend(find_autosummary_in_lines(lines, filename=filename))
+            documented.extend(find_autosummary_in_lines(lines, app=app, filename=filename))
     return documented
 
 
-def find_autosummary_in_docstring(name: str, module: str = None, filename: str = None
-                                  ) -> List[AutosummaryEntry]:
+def find_autosummary_in_docstring(name: str, app: Any, module: str = None, 
+                                  filename: str = None) -> List[AutosummaryEntry]:
     """Find out what items are documented in the given object's docstring.
 
     See `find_autosummary_in_lines`.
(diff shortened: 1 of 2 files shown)


vnmabus:
The test fails because `DummyApplication` does not have a build environment. Should I create a `DummyBuildEnvironment`, use a real `BuildEnvironment`, which requires extra attributes in `DummyApplication`, or check if `env` is not `None` and fallback to the old behaviour?

sphinx/ext/autosummary/generate.py
```diff
@@ -430,8 +430,7 @@ def find_autosummary_in_lines(lines: List[str], module: str = None, filename: st
             if m:
                 toctree = m.group(1)
                 if filename:
-                    toctree = os.path.join(os.path.dirname(filename),
-                                           toctree)
+                    _, toctree = app.env.relfn2path(toctree, filename)
```
tk0miya, line 433: Please let me know what are you doing in this line. The 2nd parameter for `relfn2path()` is a docname, not a filename. I also don't understand why do you want to change this to an absolute path. What happens in this change?

tk0miya:
>Should I create a DummyBuildEnvironment, use a real BuildEnvironment, which requires extra attributes in DummyApplication, or check if env is not None and fallback to the old behaviour?

In the apidoc command, there are no real env object. So you also need to follow how it behaves to use it.

sphinx/ext/autosummary/generate.py
vnmabus: Well, if I understood the code, the `dirname` of the second argument is used as the starting point of relative paths. I tried to preserve the behaviour of relative paths, while making absolute paths relative to the doc folder.

Sorry, because this is the first time that I am trying to contribute to this project and I do not know the terminology. What is the difference between docname and filename in this context?

vnmabus:
> 
> 
> > Should I create a DummyBuildEnvironment, use a real BuildEnvironment, which requires extra attributes in DummyApplication, or check if env is not None and fallback to the old behaviour?
> 
> In the apidoc command, there are no real env object. So you also need to follow how it behaves to use it.

Can you point me where to look? When I was looking for the callers of `generate_autosummary_docs` I only saw two places: ony for testing, in which a `DummyApplication` is passed, and the real one where a real `Sphinx` object is passed (and I assumed that the `env` attribute is valid, is that wrong?).

tk0miya:
I guess the first place you saw is the main function of the apidoc command.
https://github.com/sphinx-doc/sphinx/blob/34f35793302571811dc0a38b4b648dff28bc31f6/sphinx/ext/autosummary/generate.py#L617-L633

The apidoc command is an independent command (not a part of sphinx-build command). So there are no proper `app` and `env`. So `DummyApplication` is used instead.

sphinx/ext/autosummary/generate.py
tk0miya: A docname is an abbreviation of "document name". And it is a logical name for the document, like a relative path string from source directory without file-extension.
https://www.sphinx-doc.org/en/master/glossary.html#glossary

I know the method is almost same as you'd like to do. But it is not appropriate method for this purpose. The `generate_autosummary_docs()` is used in both `sphinx-build` and apidoc. And apidoc does not know source directory and "document".

BTW, I still don't understand what will be changed by this line. In my understanding correct, the `filename` is an absolute path under the source directory (in sphinx-build command). So I guess this does not change the behavior, right?

vnmabus: It will change the `toctree` variable if it is an absolute path prepending the doc folder.

AA-Turner:
Using relative paths in autosummary is conceptually fine, but this PR has a few implementation problems (see comments), and is out of date. Should anyone still be interested in relative paths in autosummary, please open a new PR.

A

PR #7746, opened 2020-05-29
Fix bug in mathjax: newest 2.x version not being loaded by latest.js

rsokl:
### Feature or Bugfix

- Bugfix

### Purpose
There is a bug in `latest.js` in versions 2.7.5 and below; when the current version on the CDN is 3.0 or higher, `latest.js` will not use the highest version of 2.x, but instead will use the version from which `latest.js` has been taken. As such, the sphinx-mathjax extension is stuck loading MathJax 2.7.5 from the CDN, whereas `latest.js` _should_ be loading the newest 2.x version.

See the warning section on this page:  http://docs.mathjax.org/en/latest/upgrading/v2.html

This PR fixes that by pointing to 2.7.7, the latest 2.X version available on the CDN.


tk0miya:
Thanks!

PR #7750, opened 2020-05-30
[doc] Add project internals guide

stephenfin:
This is essentially a refactoring of the huge CONTRIBUTING.rst document. This document is split into multiple separate documents based on the Django documentation. Some minor modifications are made to the documents in the process.


doc/internals/contributing.rst
```diff
@@ -0,0 +1,309 @@
+======================
```
tk0miya, line 1: Could you separate the commits into moving (renaming) and modification? AFAIK, It keeps the change history of the file. Additionally, it helps to read changes from individual commit logs for me.

I know it does not help file separation. But it would be better to keep history even only for "contributing to sphinx" part.

doc/internals/index.rst
```diff
@@ -0,0 +1,17 @@
...
+
+.. toctree::
+   :maxdepth: 2
+
+   contributing
+   code-of-conduct
+   release-process
+   organization
+   authors
+   changes
```
tk0miya: I don't think changes is not a part of "internal". Why do you move this here?

tk0miya (changes requested):

doc/internals/contributing.rst
stephenfin: I've tried to do this. Hopefully it's okay now?

doc/internals/index.rst
stephenfin: Fair point. Moved back to root.

stephenfin:
Rebased onto `3.x`

doc/internals/contributing.rst
tk0miya: No, it's not good. It seems you did moving CONTRIBUTING.rst to doc/internals/contributing.rst and adding new CONTRIBUTING.rst in the same commit. It seems git considers it large deletion of CONTRIBUTING.rst and creation of doc/internals/contributing.rst.
https://github.com/sphinx-doc/sphinx/pull/7750/commits/7e8ae1127577ca60d6ce98b9c958b6ddb554a44e

As a result, the history of doc/internals/contributing.rst has been cleared.
https://github.com/stephenfin/sphinx/blame/doc/internals-section/doc/internals/contributing.rst


Please separate renaming and other actions.
For example, this branch commits 3 times; 1) moving 2) adding 3) modifying. It allows to keep history.
https://github.com/sphinx-doc/sphinx/compare/3.x...tk0miya:moving_test
https://github.com/tk0miya/sphinx/blame/moving_test/doc/internals/contributing.rst

stephenfin: Okay, I think I have this done now.

tk0miya (approved):

tk0miya:
Thank you for updates. I'll merge this after a moment.

tk0miya:
Update redirect settings on RTD. merging now.

PR #7753, opened 2020-05-30
Extend bindep integration to dpkg-based distros (Ubuntu, Debian, ...)

stephenfin:
This provides a way to bootstrap a system for development but currently only Fedora is supported. We should aim to integrate this into the CI someway but that's a job for another day.


tk0miya:
Is it better to merge this into 3.x branch? I think this is available for current release.

stephenfin:
> Is it better to merge this into 3.x branch? I think this is available for current release.

Done

mgeier:
I have no experience with `bindep`, but wouldn't it be better to use `dpkg` instead of `ubuntu`?
This would (I guess) also include Debian and other Debian-based distributions?

stephenfin:
> I have no experience with `bindep`, but wouldn't it be better to use `dpkg` instead of `ubuntu`?
> This would (I guess) also include Debian and other Debian-based distributions?

Quite possibly. Unfortunately I don't have access to a Debian environment so I can't test this myself. If someone's willing to state that these are what's necessary to build PDFs on Debian then I'm happy to trust them and make the change. Without that though, I think this is better than nothing and I can update later once I'm back somewhere with decent enough internet to pull down a Debian image/container and test.

mgeier:
I just tried it on Debian: `ubuntu` does not work (i.e., the entries are ignored and no packages are reported), `dpkg` does work as expected (i.e., the packages are reported if not already installed).

I did *not* try it on Ubuntu.

I stumbled upon an unrelated problem (https://storyboard.openstack.org/#!/story/2001205), which I could avoid by setting `LANG=C`.

stephenfin:
Okay, if that works then I can switch. I'll work on getting that bindep issue fixed next week.

stephenfin:
> Okay, if that works then I can switch. I'll work on getting that bindep issue fixed next week.

PR against bindep is [here](https://review.opendev.org/#/c/732992/). That shouldn't block this though, IMO.

tk0miya:
Okay, merging now.

PR #7765, opened 2020-06-01
Add node_id_compat to fallback normalization mode

utzig:
This adds a new config option to allow using the old normalization code, which prevented uppercase, '_' and '.' symbols. Setting it to `True` avoids breaking links to anchor heavy sites.

This PR is missing unit testing (and possibly other things!), but the purpose is to check if such feature is acceptable for inclusion and next possible steps.


tk0miya:
Unfortunately, this does not switch node_id generator to old one. Because old Sphinx components did not normalize node_ids at all. So `nodes.make_id()` was not called.
I added fallback code to keep node_ids. Could you let me know your problem in detail?

jakobandersen:
I dislike having a fallback mode, as it can never go away. We can attach multiple IDs to each entity, with the shown permalink using the newest ID. This means that all the old IDs will work, but new references will use the new ID. It's not ideal (output bloat), but at least there is a chance to selectively break copied links based on how old they are.

utzig:
@tk0miya I am updating the Zephyr project documentation to use latest Sphinx/Breathe, so for example check this page: https://docs.zephyrproject.org/latest/reference/kconfig/index-arch.html. This page has a plethora of config options and, using the first one as example (`CONFIG_64BIT`), it currently can be linked using `#cmdoption-arg-config-64bit`, but with the change in nodes the new link would look: `#cmdoption-arg-CONFIG_64BIT` (If I remember correctly). This will happen for pretty much every link and my idea is avoid breaking all existing references.

jakobandersen:
I just checked where ``make_id`` is used, and it looks like every case has a "generate old ID as well" code, except two: ``make_glossary_term`` and ``Cmdoption``.
They should get fixed instead as this PR would break new links in the rest of the std domain as well as the rst, python, js domains :-).

utzig:
> I just checked where `make_id` is used, and it looks like every case has a "generate old ID as well" code, except two: `make_glossary_term` and `Cmdoption`.
> They should get fixed instead as this PR would break new links in the rest of the std domain as well as the rst, python, js domains :-).

Thanks for the pointers, I believe it works as long as the old compat id is appended before the new one, because the html5 builder always uses ids[0]. Also most (all?) `make_old_id` methods have comments mentioning they will be removed in Sphinx > 5. I tried a different solution by writing a `SphinxPostTransform` extension that fixes the links and it works OK apart from the `genindex-*` files. I believe those are generated in a different manner where nodes are not processed in the same way, any ideas how this could be fixed?

PR #7770, opened 2020-06-02
Let option directive support args in the form of foo[=bar]

lukel97:
Subject: Let `.. option` directive support args in the form of foo[=bar]

### Feature or Bugfix

- 50% bugfix 50% feature?

### Purpose
In the [Cabal](https://github.com/haskell/cabal) documentation, there are a lot of flags that can work like `--foo` or `--foo[=N]`. These are all documented with the `.. option` directive but because of the way they were parsed, we couldn't refer to the options as ":option:`-j`", since they were parsed as `--foo[` being the option name and `=N]` being an argument.

### Detail
- This tweaks the regex so that `[` is treated in the same manner as `=` when it comes to splitting up the name and the arguments

### Relates
https://github.com/haskell/cabal/pull/6754

### DIFF
--- a/tests/roots/test-root/objects.txt
+++ b/tests/roots/test-root/objects.txt
@@ -180,7 +180,9 @@ Others
 
 .. option:: arg
 
-Link to :option:`perl +p`, :option:`--ObjC++`, :option:`--plugin.option`, :option:`create-auth-token` and :option:`arg`
+.. option:: -j[=N]
+
+Link to :option:`perl +p`, :option:`--ObjC++`, :option:`--plugin.option`, :option:`create-auth-token`, :option:`arg` and :option:`-j`
 
 .. program:: hg
 
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -43,7 +43,7 @@ logger = logging.getLogger(__name__)
 
 
 # RE for option descriptions
-option_desc_re = re.compile(r'((?:/|--|-|\+)?[^\s=]+)(=?\s*.*)')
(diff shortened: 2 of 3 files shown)


tests/roots/test-root/objects.txt
```diff
@@ -180,7 +180,9 @@ Others
 
 .. option:: arg
 
-Link to :option:`perl +p`, :option:`--ObjC++`, :option:`--plugin.option`, :option:`create-auth-token` and :option:`arg`
+.. option:: -j=[N]
```
tk0miya: Is this typo?

tests/roots/test-root/objects.txt
```diff
@@ -180,7 +180,9 @@ Others
...
+
+Link to :option:`perl +p`, :option:`--ObjC++`, :option:`--plugin.option`, :option:`create-auth-token`, :option:`arg` and :option:`-j`
```
tk0miya, line 185: Adding an example does not test anything. Please add an expected result HTML tag to here.
https://github.com/sphinx-doc/sphinx/blob/cee921fe416068021996ec24ad19afb54cd6bbec/tests/test_build_html.py#L321-L333

tk0miya (changes requested):
Acceptable. But you need to update a testcase.

tk0miya:
I thought this disallows an option that contains `[` char (ex. `--option[]`). But I've never seen such option. So +0 for this change.

tk0miya (approved):
Thank you for quick work. LGTM!

PR #7773, opened 2020-06-02
[doc] Split up builder documentation

stephenfin:
Currently, the builder document is one giant file. This is awkward to navigate and doesn't allow us to do things like show related configuration, group similar builders, etc. Resolve this by splitting the documentation into multiple documents. Five groups are indentified: HTML, LaTeX, Serializing, XML and "others".


doc/usage/builders/html.rst
```diff
@@ -1,17 +1,194 @@
...
+chapter :ref:`html-options` for details.
+
+.. module:: sphinx.builders.html
+.. class:: StandaloneHTMLBuilder
+
+   .. autoattribute:: name
+
+   .. autoattribute:: format
+
+   .. autoattribute:: supported_image_types
```
tk0miya, line 28: I wonder this block is necessary for users. I feel the descriptions you wrote are enough good.
The name is only an important field. But we don't need to use autodoc to document it.
But we already have described it in the command reference of sphinx-build:
https://www.sphinx-doc.org/en/3.x/man/sphinx-build.html?highlight=sphinx-build#cmdoption-sphinx-build-b

stephenfin: I think the `supported_image_types` field is useful, given that not all output formats support all image types. As for `name`, while we could document this manually, it seems simpler to automatically document this?

tk0miya: I understand about `supported_image_types`. How about docstrings of builder classes? I think it is duplicated with the descriptions you wrote.

About `name`, it will never (or very very rarely) be changed in the future. So automation does not help us.

PR #7781, opened 2020-06-03
Fix reported dir when throwing ApplicationError

brandonhoughton:
if path.exists(self.outdir) and not path.isdir(self.outdir), error reported self.srcdir which should be self.outdir

### Feature or Bugfix

- Bugfix

### Purpose
- fix error message regarding output directory

### Detail
-


tk0miya:
Thanks!

PR #7800, opened 2020-06-08
[RFC] Revisit/improve generating of ids (`_make_id`)

blueyed:
This started by looking into keeping leading underscores (since
`_foo.bar` is different than `foo.bar`).

From the commit message:

- use "backslashreplace" to not ignore non-ascii characters in general
  ("fünf" will get another id than "fnf")
- keep dots and underscores at the beginning also
- use a separate hyphen for replacement of each non-id (ascii) character
  ("foo::bar" becomes "foo--bar" instead of "foo-bar") - runs of
  whitespace are still replaced by a single hypehn

Ref: https://github.com/sphinx-doc/sphinx/issues/7301 (/cc @mgeier @jakobandersen)

### DIFF
--- a/sphinx/util/nodes.py
+++ b/sphinx/util/nodes.py
@@ -447,16 +447,15 @@ def _make_id(string: str) -> str:
     id = string.translate(_non_id_translate_digraphs)
     id = id.translate(_non_id_translate)
     # get rid of non-ascii characters.
-    # 'ascii' lowercase to prevent problems with turkish locale.
-    id = unicodedata.normalize('NFKD', id).encode('ascii', 'ignore').decode('ascii')
-    # shrink runs of whitespace and replace by hyphen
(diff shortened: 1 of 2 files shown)


jakobandersen:
I still believe that domains should individually generate IDs as it knows best the rules of the language, so the ``_make_id`` at best is a helper function in this process. (Maybe there should be a validation helper function to warn when a domain generates a potentially problematic ID)
So I don't have a particular opinion on this PR yet, except: in essence the function can basically not be changed, only a modified copy can be made. The old IDs must still be generated and attached to the same docutils nodes as external links may rely on them. The newest ID will be shown as the permalink.

tk0miya:
-1 for changing `make_id()` directly. It brings breaking changes to ID system. It means hyperlinks to the Sphinx documents will be broken. So we need to change carefully. And I think it's tough works. So -0 for changing more.

It seems no comments and updates since jakobandersen's comment. So I'm closing this.
Thank you for your proposal.

PR #7822, opened 2020-06-12
Fix #7715: Issue \@addtoreset after loading hyperref

### LINKED ISSUE #7715: numfig_secnum_depth > 1 leads to wrong figure links in the pdf
When using numfig = True and numfig_secnum_depth greater than 2, the references (links) to the figures in the pdf are wrong. 

The referencing for numfig_secnum_depth = (2) is the following:
A figure is the first figure in its subsection, e.g. 3.7.1. When clicking on the link, it jumps to the first figure in chapter 3 which is first in its subsection, e.g. 3.1.1
If a figure is the second in its subsection, e.g. 3.7.2 it jumps to the first figure in chapter 3 which is second in its subsection, e.g. 3.3.2
If a figure is the third in its subsection, e.g. 3.7.3 it jumps to the first figure in chapter 3 which is third in its subsection, e.g. 3.5.3
etc... 

Found in sphinx version 1.8.5 (I could not find any hint in the Changelog that this issue was fixed in later versions)

Jellby:
In LaTeX, the `\@addtoreset` command should be executed after loading hyperref, or the hyperlinks may be wrong. This just puts them in `\AtBeginDocument` to ensure this is the case. Fix #7715

### DIFF
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -881,52 +881,62 @@
        \g@addto@macro\spx@preBthefigure{\fi}}%
 \fi
 \ifnum\spx@opt@numfigreset>1
+  \AtBeginDocument{%
     \@addtoreset{figure}{section}%
     \@addtoreset{table}{section}%
     \@addtoreset{literalblock}{section}%
     \ifspx@opt@mathnumfig
       \@addtoreset{equation}{section}%
-    \fi
+    \fi%
+  }%
     \g@addto@macro\spx@preAthefigure{\ifnum\c@section>\z@\arabic{section}.}%
     \g@addto@macro\spx@preBthefigure{\fi}%
 \fi
 \ifnum\spx@opt@numfigreset>2
+  \AtBeginDocument{%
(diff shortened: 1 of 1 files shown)


tk0miya (changes requested):
Could you rebase this into 3.x branch? Then it will be released as the next minor release.

sphinx/texinputs/sphinx.sty
```diff
@@ -881,52 +881,62 @@
...
 \fi
 \ifnum\spx@opt@numfigreset>1
+  \AtBeginDocument{%
     \@addtoreset{figure}{section}%
     \@addtoreset{table}{section}%
     \@addtoreset{literalblock}{section}%
     \ifspx@opt@mathnumfig
       \@addtoreset{equation}{section}%
-    \fi
+    \fi%
```
jfbu, line 890: the `%` is not needed here, but does not harm either :)

jfbu (approved):
Thanks for fixing this oversight.

tk0miya (approved):

tk0miya:
Thank you for quick update. LGTM!

PR #7832, opened 2020-06-13
Fix exception causes all over the codebase

cool-RR:
This is the continuation of #7820 .

I've used `exc` in all places where there wasn't already a name for the caught exception.


tk0miya:
Looks good. But flake8 warns for some places. Please check it :-)

cool-RR:
@tk0miya We're good now, right?

tk0miya:

PR #7833, opened 2020-06-14
Run lint tools on GH Actions

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- GHA allows us to invoke 20 jobs at the same time. It improves the speed of our CI process.
  https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits
- As a first step, this PR moves linting to GHA from Travis CI. I'll work for other CI processes in the next step.

### DIFF
--- a/.github/workflows/docslint.yml
+++ b/.github/workflows/docslint.yml
@@ -1,0 +1,16 @@
+name: Lint the format of document
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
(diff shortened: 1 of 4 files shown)


tk0miya:
It seems the error on the flake8 job cancels other linting jobs immediately. But CI jobs on windows are not canceled. It means one of the jobs in "matrix build" cancels other jobs in the same build. I feel that the results of all linting tools are helpful for developers, no reason to stop them if some tool failed. So I'd like to separate the jobs to independent workflows.

tk0miya:
I found `jobs.<job_id>.strategy.fail-fast` after merging this...
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

PR #7846, opened 2020-06-17
Use well-formed XML syntax.

abramobagnara:
Subject: fix XML syntax

### Feature or Bugfix
- Feature

### Purpose
- Permit use of XML tools on HTML output


tk0miya:

Issue #7838, opened 2020-06-15
List items have extra vertical space since Sphinx 3.1.x


cjolowicz:
**Describe the bug**
List items in HTML output are displayed with extra vertical space, as if separated by blank lines. Compare the rendering of the feature list in this project, or the screenshots further below:

- https://cookiecutter-hypermodern-python.readthedocs.io/en/2020.5.15/ (Sphinx 3.0.3)
- https://cookiecutter-hypermodern-python.readthedocs.io/en/2020.6.15/ (Sphinx 3.1.0)

Tested on a recent Chrome for Mac, see below for environment info.

It appears this is due to changes in `basic.css` between Sphinx 3.0.4 and 3.1.0.  See the diff of `basic.css` at the bottom of this issue description. The stylesheet rules reset the bottom margin of `<p>` when nested in `<li>`. Apparently this logic was broken in Sphinx 3.1.x for at least one browser.

**To Reproduce**
```
$ git clone https://github.com/cjolowicz/cookiecutter-hypermodern-python
$ git checkout 2020.6.15
$ cd cookiecutter-hypermodern-python
$ pip install -r docs/requirements.txt
$ cd docs
$ sphinx-build . _build
$ open _build/index.html
```

**Expected behavior**
List items are displayed in a compact way, without vertical spacing amounting to blank line.

**Your project**
https://github.com/cjolowicz/cookiecutter-hypermodern-python

**Screenshots**

With Sphinx 3.0.3:

<img width="444" alt="list-sphinx-3 0 3" src="https://user-images.githubusercontent.com/653941/84621501-fd0fdc00-aeda-11ea-982a-986e40d50bb9.png">

With Sphinx 3.1.0:

<img width="448" alt="list-sphinx-3 1 0" src="https://user-images.githubusercontent.com/653941/84621506-026d2680-aedb-11ea-9b22-f18608042091.png">

**Environment info**
- OS: macOS Mojave 10.14.6 (18G4032)
- Python version: 3.8.2
- Sphinx version: 3.1.0, 3.1.1
- Sphinx extensions:  -
- Extra tools: Chrome 83.0.4103.97 (Official Build) (64-bit)

**Additional context**

Here is the diff of basic.css from Sphinx 3.0.4 to 3.1.0 which causes the breakage for me. IOW if I revert these changes, the list is displayed correctly.

```diff
diff --git a/_static/basic.css b/_static/basic.css
index 3a37f63..56f5efc 100644
[...]

PR #7852, opened 2020-06-18
basic CSS: Take "simple" lists into account, resuscitating the "html_compact_lists" config option

mgeier:
See #7838, where the CSS classes `compact` and `simple` have been mentioned.

I'm not sure if that's the best solution, but it should definitely be a basis for further discussion.

This is how it looks with the docutils demo https://github.com/docutils/docutils/blob/master/docutils/docs/user/rst/demo.txt, using the `sphinxdoc` theme:

**Before:**

image

**Between (Sphinx 3.1):**

image

**After:**

image

----

There is still something strange happening in the last case: ~The list containing "Third level." and "Item 2." doesn't get the `simple` class, but the one containing it does.~
~I think those two cases should be reversed, right?~
~I guess that's a bug?~

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -537,6 +537,14 @@ ul ul > li:last-child > :last-child {
     margin-bottom: revert;
 }
 
+ol.simple > li > p,
+ul.simple > li > p,
+ol.compact > li > p,
+ul.compact > li > p {
+    margin-top: 0;
+    margin-bottom: 0;
+}
+
 dl.footnote > dt,
 dl.citation > dt {
     float: left;
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -538,9 +538,7 @@ ul ul > li:last-child > :last-child {
 }
 
 ol.simple > li > p,
-ul.simple > li > p,
-ol.compact > li > p,
-ul.compact > li > p {
+ul.simple > li > p {
     margin-top: 0;
     margin-bottom: 0;
 }
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -513,33 +513,30 @@ ol.upperroman {
     list-style: upper-roman;
 }
 
-ol > li:first-child > :first-child,
-ul > li:first-child > :first-child {
+:not(li) > ol > li:first-child > :first-child,
+:not(li) > ul > li:first-child > :first-child {
     margin-top: 0px;
 }
 
-ol ol > li:first-child > :first-child,
-ol ul > li:first-child > :first-child,
-ul ol > li:first-child > :first-child,
-ul ul > li:first-child > :first-child {
-    margin-top: revert;
-}
-
-ol > li:last-child > :last-child,
-ul > li:last-child > :last-child {
+:not(li) > ol > li:last-child > :last-child,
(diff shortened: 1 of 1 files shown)


mgeier:
For comparison, the rendering of the `docutils` demo from https://docutils.sourceforge.io/docs/user/rst/demo.html#bullet-lists, ~which isn't quite consistent, either~:

image

jakobandersen:
- I don't think the ``compact`` class should be used in the CSS due to redundancy. If it is present, the ``simple`` class should get set anyway.
- The docutils page you link to seem to be generated with the HTML 4 writer (no paragraphs in list items).
- The inconsistency seems to be on purpose in docutils: https://repo.or.cz/docutils.git/blob/HEAD:/docutils/docutils/writers/_html_base.py#l527, only set ``simple`` if the current list is simple and a the nearest parent isn't.
  However, this is only happening in the bullet list, not for enumerated lists. I suggest we overwrite this particular function ``visit_bullet_list`` and make it unconditionally set ``simple`` on each list that is simple.

mgeier:
Thanks @jakobandersen, I've removed the `compact` class in 0051992ba514186473363fadd5ceaa496c276a60.

The other things you mention sound good, but they should be done in a separate issue/PR, right?

jakobandersen:
I'm not sure. In a sense it belongs together with this, and probably also the other general CSS rules for lists discussed in #7838. It's not immediately clear to me which rules are really needed in the end, and the exact reasoning behind them all.

mgeier:
I thought that "compact" lists can contain "non-compact" sub-lists. I was wrong.

I read again the definition at https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_compact_lists:

> If true, a list all whose items consist of a single paragraph and/or a sub-list all whose items etc… (recursive definition) will not use the `<p>` element for any of its items.

This implies, from a CSS point of view, that the `simple` class also applies to nested lists.
There is no need for nested lists to have the `simple` class if a parent list has it (but of course it doesn't hurt if they do have the `simple` class, it's just redundant).

Also I previously wasn't completely aware of the "**and**/or" part of the definition.

With this new (for me) knowledge, I made some adaptations to the CSS in this PR: 82768d8cffe46646db46a4b8baf73435888e2ffc.

With these changes, no changes to the Python code are necessary and I think it finally looks exactly as intended:

image

Note the correct space *after* compact lists which are nested within non-compact lists.

Also note the correct space *between* the separate lists in the last list item.

The space between "1. Arabic numerals." and "a. lower alpha)" might seem out of place given the further sub-lists, but I think it's actually correct. The main list (numbered 1. and 2.) is "non-compact" (because item 2 contains more than "a single paragraph and/or a [single] sub-list") while the nested lists (a., i., A., I.) are "compact".

BTW, the `html_compact_lists = False` option also works correctly:

image

return42:
@mgeier wrote https://github.com/sphinx-doc/sphinx/pull/7657#issuecomment-647040482

> @return42 Please check out #7852, which should fix the list spacing issues.

I am very sorry, I have my own timeline in another Open-Source project .. thanks for your investigation! For the future you better might pick some tests from wider docs .. e.g. the themes  from poco often used .. or use the one from us  :)

- preview: https://asciimoo.github.io/searx/index.html
- [run make docs](https://asciimoo.github.io/searx/dev/contribution_guide.html#documentation)
- BTW we have a really nice [reST primer](https://asciimoo.github.io/searx/dev/reST.html) :)

tk0miya (approved):
Sorry for late. Looks good to me. So merging this now.

PR #7855, opened 2020-06-20
Slightly modify PR template

jakobandersen:
### Type
- Github

### Purpose
Adjust the PR template regarding the type of change. Most PRs should be a single type, but sometimes more than one category apply. Use checkboxes to better signal how this section should be used.


jakobandersen:
Nevermind, I didn't realize that this is interpreted as a task list so every PR would look incomplete.

PR #7864, opened 2020-06-22
Fix #7863: html: HTMLWriter strips the first two entries of meta

### LINKED ISSUE #7863: HTMLWriter strips the first two entries of `meta`
**Describe the bug**
When creating a custom translator which does *not* ultimately inherit from `docutils.writers._html_base.HTMLTranslator`, getting metadata out through HTMLWriter requires "padding" the `meta` fields with two ingnored entries.

**To Reproduce**
* create a translator which only inherits from NodeVisitor and implements everything itself
* define a `meta` field with any number of entries
* set the custom translator as the application's
* generate HTML

**Expected behavior**
A tag for each entry into the corresponding field

**Observed**
Nothing for the first two entries, because [HTMLWriter skips the first two entries when generating clean_meta](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/writers/html.py#L76).

I expect because docutils fills those with a "charset" entry and a "generator" (setting itself as generator). Sadly [the original sphinx commit](https://github.com/sphinx-doc/sphinx/commit/4b9119f9905e608e74335214d9e15514d0adf2f7) does not explain why it strips the content during `translate` instead of the translator removing or fixing the entries it wants gone or replaced.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7863

### DIFF
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -16,6 +16,7 @@ import warnings
 from typing import Any, Iterable, Tuple
 from typing import cast
 
+import docutils
 from docutils import nodes
 from docutils.nodes import Element, Node, Text
 from docutils.writers.html4css1 import Writer, HTMLTranslator as BaseTranslator
@@ -76,7 +77,23 @@ class HTMLWriter(Writer):
                      'footer', 'html_prolog', 'html_head', 'html_title',
                      'html_subtitle', 'html_body', ):
             setattr(self, attr, getattr(visitor, attr, None))
-        self.clean_meta = ''.join(self.visitor.meta[2:])
+        self.clean_meta = self.get_clean_meta()
+
+    def get_clean_meta(self) -> str:
+        visitor = cast(HTMLTranslator, self.visitor)
+        meta = visitor.meta[:]
+
+        # Remove <meta charset> tag from docutils
+        content_type = visitor.content_type % visitor.settings.output_encoding
+        if len(meta) and meta[0] == content_type:
+            meta.pop(0)
+
+        # Remove <meta name=generator> tag from docutils
+        generator = visitor.generator % docutils.__version__
+        if len(meta) and meta[0] == generator:
+            meta.pop(0)
+
+        return ''.join(meta)
 
 
 class HTMLTranslator(SphinxTranslator, BaseTranslator):
--- a/CHANGES
+++ b/CHANGES
@@ -20,6 +20,7 @@ Bugs fixed
 
 * #7839: autosummary: cannot handle umlauts in function names
 * #7715: LaTeX: ``numfig_secnum_depth > 1`` leads to wrong figure links
+* #7863: html: HTMLWriter strips the first two entries of meta
 * #7846: html theme: XML-invalid files were generated
 
 Testing


sphinx/writers/html.py
```diff
@@ -76,7 +77,23 @@ def translate(self) -> None:
...
             setattr(self, attr, getattr(visitor, attr, None))
-        self.clean_meta = ''.join(self.visitor.meta[2:])
+        self.clean_meta = self.get_clean_meta()
+
+    def get_clean_meta(self) -> str:
+        visitor = cast(HTMLTranslator, self.visitor)
+        meta = visitor.meta[:]
+
+        # Remove <meta charset> tag from docutils
+        content_type = visitor.content_type % visitor.settings.output_encoding
```
xmo-odoo, line 87: Given the original report (our use case) is a translator which specifically does not inherit from docutils' HTMLTranslator, I'd say there's no guarantee these attributes will be present.

I can see a few possibilities / alternatives here:

1. check if the visitor is an instance of docutils' HTMLTranslator and only process the items in that case, however if third-parties have implemented a translator which doesn't extend docutils' but sets these it's going to fail
2. check if the *attributes* are present and only strip the attributes in that case (this could handle translators which replicate more of the original behaviour than ours do), however in a case similar to (1) third parties might have set the metas to remove without the corresponding docutils' attributes
3. specifically parse and process the `charset` and `generator` metas, either to strip them entirely or to fix them using sphinx semantics instead (replace `charset="unicode"` by the value of `html_output_encoding`, and mark Sphinx as the [generator](https://html.spec.whatwg.org/multipage/semantics.html#meta-generator) -- at lest if it finds docutils there
4. move the meta-cleaning operation to an (optional) feature of the translator e.g. `self.clean_meta = ''.join(getattr(visitor, 'get_clean_meta', visitor.meta))`, this way the translator has an option to provide clean metas (third parties might see unexpected "padding" metas reappear in the output but that seems like a somewhat lower risk)

I'd favor 3 then 4 personally, the first two seem kinda brittle. Maybe @birkenfeld remembers why they straight stripped the metas created by docutils in the first place though? That might be useful insight.

tk0miya: I noticed Sphinx's HTML builder is already strongly coupled with docutils' HTMLTranslator. It uses `astext()`, `meta` and etc. attributes.
https://github.com/sphinx-doc/sphinx/blob/3336dcc2aea0cc93d54ba661d4da65ea46f088b0/sphinx/writers/html.py#L72-L79

xmo-odoo: > I noticed Sphinx's HTML builder is already strongly coupled with docutils' HTMLTranslator. It uses `astext()`, `meta` and etc. attributes.
> 
> https://github.com/sphinx-doc/sphinx/blob/3336dcc2aea0cc93d54ba661d4da65ea46f088b0/sphinx/writers/html.py#L72-L79

To a limited extent, it does have a few attributes and methods it expects but these are relatively easy to implement without extending docutils' HTMLTranslator. I'm not saying they're super decoupled, but it's really quite easy to get a working translator from just the nodevisitor, the serialisation of the meta list is really the only bit which had me stumped.

tk0miya:
I withdraw this because I can't find the reason to support a translator similar to HTMLTranslator but not inherited. I'll reconsider that again if other developers request to support their custom translators not inheriting HTMLTranslator. Until then, please fill in two dummy items to the `meta` attribute if you create a translator that imitates a subclass of HTMLTranslator.

xmo-odoo:
> I withdraw this because I can't find the reason to support a translator similar to HTMLTranslator but not inherited. I'll reconsider that again if other developers request to support their custom translators not inheriting HTMLTranslator. Until then, please fill in two dummy items to the `meta` attribute if you create a translator that imitates a subclass of HTMLTranslator.

Sure. For what that's worth, our reason was that we needed to override a lot of it to properly integrate all the bootstrap classes to bootstrap-theme our doc (this was started a while back), so it seemed much easier and safer to start from a clean slate and be able to opt into every behaviour specifically (and only implementing the features we were actually using, such that we filled in the translator "on demand" and only filled in what we actually used as we started using them).

PR #7870, opened 2020-06-25
Propagate original extension errors

### LINKED ISSUE #7868: Extension throwing an exception provides no helpful debug info
**Describe the bug**
When an extension throws an exception, sphinx errors and halts, but provides no information on what the exception was or how to debug it further. It would be helpful if more information on the exception thrown could be shown, or the path to a log that can be further used to debug was shown.

**To Reproduce**
```
$ make html -d
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
Reading makefiles...
Reading makefile `Makefile'...
Updating makefiles....
 Considering target file `Makefile'.
  File `Makefile' does not exist.
  Finished prerequisites of target file `Makefile'.
 Must remake target `Makefile'.
 Successfully remade target file `Makefile'.
Updating goal targets....
Considering target file `html'.
 File `html' does not exist.
 Looking for an implicit rule for `html'.
 Trying pattern rule with stem `html'.
 Trying rule prerequisite `Makefile'.
 Found an implicit rule for `html'.
  Pruning file `Makefile'.
 Finished prerequisites of target file `html'.
Must remake target `html'.
Putting child 0x7f9accc05790 (html) PID 51850 on the chain.
Live child 0x7f9accc05790 (html) PID 51850 
Running Sphinx v3.1.1
loading intersphinx inventory from https://docs.python.org/objects.inv...
intersphinx inventory has moved: https://docs.python.org/objects.inv -> https://docs.python.org/3/objects.inv

Extension error:
Handler <function process_automodsumm_generation at 0x7fcacc1d1160> for event 'builder-inited' threw an exception
Reaping losing child 0x7f9accc05790 PID 51850 
make: *** [html] Error 2
Removing child 0x7f9accc05790 PID 51850 from chain.
```

**Expected behavior**
More information on the exception thrown by the extension to be provided, so it can be debugged.

**Your project**
https://github.com/dstansby/soloswapy/

**Environment info**
- OS: [e.g. Unix/Linux/Mac/Win/other with version]: macos 10.15.5
- Python version: [e.g. 3.7.1] 3.8.3
- Sphinx version: [e.g. 1.8.2] 3.1.1
- Extra tools: [e.g. Browser, tex or something else]

dstansby:
Subject: Previously the `exc` error did not get propagated through to the constructor of `ExtensionError`. This fixes that so that the text of the original error gets printed.

### Feature or Bugfix

- Bugfix

### Purpose

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- Fixes https://github.com/sphinx-doc/sphinx/issues/7868

### DIFF
--- a/sphinx/events.py
+++ b/sphinx/events.py
@@ -115,7 +115,7 @@ class EventManager:
                 raise
             except Exception as exc:
                 raise ExtensionError(__("Handler %r for event %r threw an exception") %
-                                     (listener.handler, name)) from exc
+                                     (listener.handler, name), exc) from exc
         return results
 
     def emit_firstresult(self, name: str, *args: Any,


sphinx/events.py
```diff
@@ -115,7 +115,7 @@ def emit(self, name: str, *args: Any,
                 raise
             except Exception as exc:
                 raise ExtensionError(__("Handler %r for event %r threw an exception") %
-                                     (listener.handler, name)) from exc
+                                     (listener.handler, name), exc)
```
tk0miya: I think it is better to show the exception in the message like "threw an exception: %r". What do you think?

Addition to this, `from exc` is still worthy. So adding `exc` as a new argument is better to me.

dstansby: `ExtensionError` has a custom `__str__` which prints `exc`, so I thought it was best to leave it to `__str__` to handle including `exc`:
https://github.com/sphinx-doc/sphinx/blob/03e85df6175258bd6d6ec24aaeed8fc74a335651/sphinx/errors.py#L57-L67

I'm not sure what you mean by the second bit - should I put back `from exc`?

tk0miya: Thank you for your comment. Indeed, it is not needed.

>I'm not sure what you mean by the second bit - should I put back from exc?

Yes. I think both giving `exc` as an argument and passing it via `from` clause is better.

dstansby: 👍 I've put back the `from exc`

tk0miya:
Merged. Thanks!

PR #7878, opened 2020-06-27
basic CSS: allow more "floating"

mgeier:
I wasn't aware of an IMHO very un-intuitive interaction between `overflow` and `float`: If an element has an `overflow` setting, preceding elements cannot "float" into it.

That's why I'm suggesting to remove the `overflow` setting from `admonition` and `topic`.

This partially reverts my previous PR #7542. I'm only keeping the `overflow` setting on `sidebar`, where "floating into it" is not desired anyway.

For the other elements, it looks like we cannot have perfect "overflow" and perfect "float" behavior at the same time.
Derived themes can still add their `overflow` settings if they want.

This also partially reverts my previous PR #7484. It allows "right floats" to float into `admonition` and `topic` elements (as was possible before #7484). It still doesn't allow floating into code blocks, because this would interact badly with their `overflow` behavior (which is more important).

### Feature or Bugfix

- Bugfix

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -32,6 +32,7 @@ Bugs fixed
   an attribute and module that are same name
 * #7806: viewcode: Failed to resolve viewcode references on 3rd party builders
 * #7838: html theme: List items have extra vertical space
+* #7878: html theme: Undesired interaction between "overflow" and "float"
 
 Testing
 --------
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -328,8 +328,8 @@ p.sidebar-title {
     font-weight: bold;
 }
 
-div.admonition, div.topic, pre, div[class|="highlight"] {
-    clear: both;
+div.admonition, div.topic, blockquote {
+    clear: left;
 }
 
 /* -- topics ---------------------------------------------------------------- */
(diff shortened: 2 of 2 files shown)


tk0miya (changes requested):
@mgeier Could you add a CHANGES entry please? I can't describe what is changed by this.

Note: I still don't understand what is happened now. So I'm just merging this...

mgeier:
I've re-based this PR and added `CHANGES` in c40cb79461da8b7b1322fb8ed69c0c16f0191e9f.

There is no separate issue for this, one example for the problem has been reported in https://github.com/sphinx-doc/sphinx/pull/7484#issuecomment-646058972

The problem is that it's not possible to have `overflow` (i.e. scrollbars) on admonitions etc. (as I suggested in #7542) and at the same time allowing preceding elements (like `sidebar`) to "flow" into the admonition.

This PR partially reverts #7542 and #7484 to allow `sidebar` to "flow" into admonitions etc.

This is  more in line with previous behavior (before #7542 and #7484). If a derived theme wants to enable `overflow` behavior (and thereby disable "float" behavior), it can still do that, but I think for the `basic` theme we shouldn't do that.

This PR also adds a so-called "clearfix" (in 3098aba3431eb8c40ea052d6b8fc90328802c389) to avoid elements (like images) to "flow" out of the bottom of admonitions etc.

tk0miya:
Thank you for the detailed explanation! Merged.

PR #7879, opened 2020-06-27
Close #7849: html: Add html_codeblock_linenos

### LINKED ISSUE #7849: Code blocks with :linenos: are not accessible to blind people
**Describe the bug**
When the screen reader reads the code block, it starts to say "1, 2, 3, 4, 5, 6, ..." and, after all numbers, it reads the code.

**To Reproduce**
Just create a code block with linenos (and more than one line of code), and try to read it with a screen reader.

**Expected behavior**
The screen reader will read "1 first line of code, 2 second line of code, ...) or, better, permit to browse the code like a table (with a real table or aria properties).

**Environment info**
- OS: Linux
- Python version: All
- Sphinx version: 3.1.1
- Sphinx extensions: none
- Extra tools: Firefox or Chromium and Orca

**Additional context**
The problem here is the structure of the code block. It's generated in a table with a single line. One cell contains all numbers and one all lines of code. It's impossible to the screen reader to associate a number to a line of code, and the user can't get advantage of line numbers. Fixing this can fix issue #1007.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #7849 
- This adds `html_codeblock_linenos` option to control the style of line numbers for code-blocks.

### Reamining tasks:

- [x] determine the name of the configuration
- [x] add testcases
- [x] documents

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -26,7 +26,7 @@ from docutils.utils import relative_path
 from sphinx import package_dir, __display_version__
 from sphinx.application import Sphinx
 from sphinx.builders import Builder
-from sphinx.config import Config
+from sphinx.config import Config, ENUM
 from sphinx.deprecation import RemovedInSphinx40Warning
 from sphinx.domains import Domain, Index, IndexEntry
 from sphinx.environment.adapters.asset import ImageAdapter
@@ -1226,6 +1226,8 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value('html_search_scorer', '', None)
     app.add_config_value('html_scaled_image_link', True, 'html')
     app.add_config_value('html_baseurl', '', 'html')
+    app.add_config_value('html_codeblock_linenos_style', 'table', 'html',
+                         ENUM('table', 'inline'))
     app.add_config_value('html_math_renderer', None, 'env')
     app.add_config_value('html4_writer', False, 'html')
 
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,8 @@ Deprecated
 Features added
 --------------
 
+* #7849: html: Add :confval:`html_codeblock_linenos_style` to change the style
+  of line numbers for code-blocks
 * #7853: C and C++, support parameterized GNU style attributes.
 * #7888: napoleon: Add aliases Warn and Raise.
 * C, added :rst:dir:`c:alias` directive for inserting copies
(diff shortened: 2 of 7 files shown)


mgeier:
To solve the accessibility problem described in #7849, wouldn't it make more sense to change the default to `'inline'`?

In CSS, the line numbers should have `user-select: none;` to avoid polluting the clipboard with all those line numbers.
It's probably enough to add `span.lineno` to the already existing rule:

https://github.com/sphinx-doc/sphinx/blob/03e85df6175258bd6d6ec24aaeed8fc74a335651/sphinx/themes/basic/static/basic.css_t#L746-L749

tk0miya:
>To solve the accessibility problem described in #7849, wouldn't it make more sense to change the default to 'inline'?

I'm afraid that this causes minor breaking changes (as you mentioned this to sphinx-copybutton). So I thought it is better to add a new configuration during 3.x, and change its default to 'inline' on 4.0 release (if needed).

sphinx/builders/html/__init__.py
```diff
@@ -1226,6 +1226,7 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value('html_search_scorer', '', None)
     app.add_config_value('html_scaled_image_link', True, 'html')
     app.add_config_value('html_baseurl', '', 'html')
+    app.add_config_value('html_codeblock_linenos', 'table', 'html', ENUM('table', 'inline'))
```
tk0miya: Does anyone have a good name for this configuration?

Candidates:
* html_codeblock_linenos_style
* html_codeblock_line_numbers
* html_codeblock_line_numbers_style
* html_pygments_codeblock_linenos_style
* html_pygments_settings['codeblock_linenos']

shimizukawa: I certainly couldn't imagine what `html_codeblock_linenos` would do.

I prefer `html_codeblock_linenos_style` as it is shorter and easier to understand.

Or perhaps `html_pygments_linenos`.
This would be easier to convey the image of passing the pygments' linenos option directly to the `linenos` option.

shimizukawa:
I talked with @tk0miya at the Sphinx(jp) hack-a-thon event, and we discussed that there is no need to leave table output in the future.
The problems I've encountered so far with the `table` output are that the line numbers are slightly out of sync with the code, and that it's a lot of work to design in css.
I thought there was a problem with `inline` mode that the line numbers are copied when copying the code, but if we can avoid it with css, it doesn't seem to be a problem.

So, my opinions are:

* add `html_codeblock_linenos` (or another name) for 3.x as as experimental.
* `html_codeblock_linenos=inline` by default from 4.0 and deprecate it.
* remove `html_codeblock_linenos` option after 2 major version (maybe 6.0)

tk0miya:
Updated.

PR #7905, opened 2020-07-02
[WIP?] C, add compatibility flag for parsing some pre-v3 input

### LINKED ISSUE #1836: Intersphinx support in toctrees
Currently, toctree links can be either internal documents or arbitrary external URLs.

For Sphinx projects which serve as hub of sorts, or to cross-TOC between related projects, it'd be nice if it were possible for toctree to use intersphinx when references are not found locally (and are not URLs)

### LINKED ISSUE #7899: C domain changes of Sphinx 3 prevent to write doc compatible with Sphinx 2 and Sphinx 3
The Python documentation can be found in the Doc/ directory of https://github.com/python/cpython/ It is quite big: 262k lines of .rst files (and 1300 of extensions written in Python: see Doc/tools/).

It is no longer possible to build the Python documentation with Sphinx 3, many syntax which were allowed previously ("tolerated"?) are now treated as errors: see https://bugs.python.org/issue40204 for details.

According to the Sphinx changelog, it is a deliberate that the C domain is more strict. Extract:

> The C domain has been rewritten, with additional directives and roles. The existing ones are now more strict, resulting in new warnings.

The problem is that I fail to find a way to write a documentation which can be built and rendered properly with Sphinx 2 **and** Sphinx 3. Dropping Sphinx 2 support is not an option for the Python project. Too many Linux distributions currently build the Python documentation with Sphinx 2, and it seems like upgrading to Sphinx 3 will break at least Python, but likely other projects documentation.

It seems like we have the following options for the Python documentation:

* Stay at Sphinx forever
* Backport the ":c:struct:" and ":c:expr:" markups somehow in Sphinx 2 using our CPython Sphinx extensions
* Update Sphinx 3 to tolerate our "incorrect" documentation using :c:type: and :c:type:`PyObject\*` (invalid C syntax), maybe as an opt-in option
* Update Sphinx 2 to support Sphinx 3 new syntax: something like Python 2.7 which backported a few Python 3 features like b"bytes string" syntax

Example of issue: `.. c:struct: MyStruct` is required by Sphinx 3 to document a structure, since ``.. c:type: MyStruct` is now treated as an error. But with Sphinx 2, `.. c:struct: MyStruct` section is simply ignored and the whole section miss from the HTML output.

cc @birkenfeld

jakobandersen:
The C domain got much stricter in its parsing in version 3 and added additional directives and roles to provide better markup. This PR adds the configuration variable ``c_allow_pre_v3``. When set to ``True`` the type directive and role will try to convert some previously recognized input to the v3 equivalent.

### Feature or Bugfix
- Feature

### Detail
- Fixes #7899

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -22,6 +22,7 @@ from sphinx import addnodes
 from sphinx.addnodes import pending_xref
 from sphinx.application import Sphinx
 from sphinx.builders import Builder
+from sphinx.deprecation import RemovedInSphinx50Warning
 from sphinx.directives import ObjectDescription
 from sphinx.domains import Domain, ObjType
 from sphinx.environment import BuildEnvironment
@@ -2937,6 +2938,23 @@ class DefinitionParser(BaseParser):
             init = ASTInitializer(initVal)
         return ASTEnumerator(name, init)
 
+    def parse_pre_v3_type_definition(self) -> ASTDeclaration:
+        self.skip_ws()
+        declaration = None  # type: Any
+        if self.skip_word('struct'):
+            typ = 'struct'
+            declaration = self._parse_struct()
+        elif self.skip_word('union'):
+            typ = 'union'
+            declaration = self._parse_union()
+        elif self.skip_word('enum'):
+            typ = 'enum'
+            declaration = self._parse_enum()
+        else:
+            self.fail("Could not parse pre-v3 type directive."
+                      " Must start with 'struct', 'union', or 'enum'.")
+        return ASTDeclaration(typ, typ, declaration, False)
+
     def parse_declaration(self, objectType: str, directiveType: str) -> ASTDeclaration:
         if objectType not in ('function', 'member',
                               'macro', 'struct', 'union', 'enum', 'enumerator', 'type'):
@@ -3114,6 +3132,9 @@ class CObject(ObjectDescription):
     def parse_definition(self, parser: DefinitionParser) -> ASTDeclaration:
         return parser.parse_declaration(self.object_type, self.objtype)
 
+    def parse_pre_v3_type_definition(self, parser: DefinitionParser) -> ASTDeclaration:
+        return parser.parse_pre_v3_type_definition()
+
     def describe_signature(self, signode: TextElement, ast: Any, options: Dict) -> None:
         ast.describe_signature(signode, 'lastIsName', self.env, options)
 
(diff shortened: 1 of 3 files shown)


jakobandersen:
@vstinner, I believe this should take care of most cases that can not be changed to work in both v2 and v3. Let me know if there are more cases.

vstinner:
In short, the PR works as expected, thanks! I just had to remove -W option from sphinx-build command used by Python.

---

I built the Python documentation with this PR.

`make check SPHINXOPTS="-q -W -j4"` succeed.

`make suspicious  SPHINXOPTS="-q -W -j4"` fails with:

```
Doc/extending/extending.rst.rst:160:RemovedInSphinx40Warning: Pre-v3 C type role ':c:type:`PyErr_*`' converted to ':c:expr:`PyErr_*`'.
The original parsing error was:
Invalid C declaration: Expected end of definition. [error at 6]
  PyErr_*
  ------^
```

That's because sphinx-build is run with `-W` to treat warnings as errors.

Without `-W`, `make suspicious SPHINXOPTS="-q -j4" SPHINXERRORHANDLING=""` (check for suspicious syntax) and `make html SPHINXOPTS="-q -j4" SPHINXERRORHANDLING=""` (render as HTML) commands succeed (exit code 0), but log tons of warnings (which is acceptable for now).

sphinx/domains/c.py
```diff
@@ -3128,8 +3149,26 @@ def handle_signature(self, sig: str, signode: TextElement) -> ASTDeclaration:
...
+                try:
+                    ast = self.parse_pre_v3_type_definition(parser)
+                    parser.assert_end()
+                except DefinitionError:
+                    raise eOrig
+                self.object_type = ast.objectType  # type: ignore
+                msg = "{}: Pre-v3 C type directive '.. c:type:: {}' converted to " \
+                      "'.. c:{}:: {}'." \
+                      "\nThe original parsing error was:\n{}"
+                msg = msg.format(RemovedInNextVersionWarning.__name__,
```
vstinner: Is it really needed to remove this code path? Would it be possible to wait before deprecating it, to give time to projects to first upgrade to Sphinx 3? With this PR, it's not possible to upgrade to Sphinx 3 and treat warnings as errors, since c_allow_pre_v3=True emits tons of warnings.

jakobandersen:
There should now be a ``c_warn_on_allowed_pre_v3`` option you can set to ``False`` to disable these warnings. The behaviour is still deprecated though.
Would v4 be late enough for removal, or should we bump it to v5?

sphinx/domains/c.py
jakobandersen: With the added ``c_warn_on_allowed_pre_v3`` option you can now disable the warning.

sphinx/domains/c.py
```diff
@@ -3720,6 +3722,7 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_post_transform(AliasTransform)
 
     app.add_config_value("c_allow_pre_v3", False, 'env')
+    app.add_config_value("c_warn_on_allowed_pre_v3", True, 'env')
```
vstinner, line 3723: Thanks :-)

vstinner:
FYI the issue to track the Sphinx 3 compatibility in Python is: https://bugs.python.org/issue40204

vstinner:
> There should now be a c_warn_on_allowed_pre_v3 option you can set to False to disable these warnings. The behaviour is still deprecated though. Would v4 be late enough for removal, or should we bump it to v5?

FYI in Python, we try to wait at least 2 years between a feature is deprecated and its removal. I have no idea when v4 and v5 are planned. All I know is that so far, I'm only aware of Fedora Rawhide which ugraded to Sphinx 3, and there are multiple packages broken by this upgrade:
https://bugzilla.redhat.com/showdependencytree.cgi?id=SPHINX3&hide_resolved=1

jakobandersen:
If I remember correctly the major versions are released once per year, so v4 would be in spring 2021.

vstinner:
What's the next step for this PEP?

jakobandersen:
Basically it can be merged (after I add docs and changelog entry), but how does it work? Are there more cases where we should try to do something?
I tried compiling the Python docs, but there were other warnings that prevented me from quickly seeing if every case I expected were handled.

vstinner:
> but how does it work?

As I wrote, this PR makes possible to build the Python documentation. For the warnings, I don't have the bandwidth to go through the huge pile of warnings. Some are real issues in the doc, some are deprecation warnings that nobody tried to address, etc.

jakobandersen:
Great, and yes, there were indeed a lot of warnings. Some still related to the C domain changes, but should be fixed on the Python side, and can be in a v2 compatible manner.
I'll get this merged when as soon as I get the docs updated.

jakobandersen:
This should be in the upcomming v3.2, and is scheduled to be removed in v5.0.

PR #7907, opened 2020-07-03
doc: Clarify language specific figures format in docs

tmetzl:
Subject: Clarify language specific figure format in docs

### Feature or Bugfix
- Bugfix (Clarification)

### Purpose
- make it clearer how to name language specific figures

### Detail
- format is given as figure_language_filename, should be figure.language.filename
- add example of how to name a figure


doc/usage/configuration.rst
```diff
@@ -660,10 +660,11 @@ documentation on :ref:`intl` for details.
    generated by Sphinx will be in that language.  Also, Sphinx will try to
    substitute individual paragraphs from your documents with the translation
    sets obtained from :confval:`locale_dirs`.  Sphinx will search
-   language-specific figures named by `figure_language_filename` and substitute
-   them for original figures.  In the LaTeX builder, a suitable language will
-   be selected as an option for the *Babel* package.  Default is ``None``,
-   which means that no translation will be done.
+   language-specific figures named by `figure.language.filename` 
+   (e.g. the German version of `myfigure.png` will be `myfigure.de.png`)
```
tk0miya: Looks better. But ``` `` ``` (double back-ticks) are better for a literal string in reStructuredText. I think the original text `figure_language_filename` is also not good. Could you use it instead of a single one?

tk0miya (changes requested):

doc/usage/configuration.rst
tmetzl: Changed it for the filename example. Should I also change it for the figure.language.filename?

tk0miya: I should be fixed also. But I'll do it on another PR.

tk0miya (approved):

tk0miya:

doc/usage/configuration.rst
```diff
@@ -660,10 +660,11 @@ documentation on :ref:`intl` for details.
    generated by Sphinx will be in that language.  Also, Sphinx will try to
    substitute individual paragraphs from your documents with the translation
    sets obtained from :confval:`locale_dirs`.  Sphinx will search
-   language-specific figures named by `figure_language_filename` and substitute
-   them for original figures.  In the LaTeX builder, a suitable language will
-   be selected as an option for the *Babel* package.  Default is ``None``,
-   which means that no translation will be done.
+   language-specific figures named by `figure.language.filename`
```
tk0miya, line 663: Hmm? Why did you change `figure_language_filename` to `figure.language.filename`? This is a configuration variable. So this should not be changed.

doc/usage/configuration.rst
```diff
@@ -660,10 +660,11 @@ documentation on :ref:`intl` for details.
    generated by Sphinx will be in that language.  Also, Sphinx will try to
    substitute individual paragraphs from your documents with the translation
    sets obtained from :confval:`locale_dirs`.  Sphinx will search
-   language-specific figures named by `figure_language_filename` and substitute
-   them for original figures.  In the LaTeX builder, a suitable language will
-   be selected as an option for the *Babel* package.  Default is ``None``,
-   which means that no translation will be done.
+   language-specific figures named by `figure.language.filename` 
+   (e.g. the German version of ``myfigure.png`` will be ``myfigure.de.png``)
```
tk0miya, line 664: Note: this filename is generated by the default setting of `figure_language_filename`. So we should add a phrase like "by default setting".

doc/usage/configuration.rst
tmetzl: Oh I mistook this for an example. I did not know it was a config variable.

tk0miya: Okay. I fixed it in another PR.

Issue #7650, opened 2020-05-10
autodoc shows undecorated signature for decorated functions


tk0miya:
**Describe the bug**
autodoc shows undecorated signature for decorated functions

**To Reproduce**

```
def wrapper(fn):
    @functools.wraps(fn)
    def func():
        return fn(1, 2, 3)

    return func


@wrapper
def foo(x, y, z):
    return x + y + z
```
```
.. automodule:: example
   :members:
   :undoc-members:
```

This is converted to `foo(x, y, z)`.

**Expected behavior**
The function `foo` should be shown as `foo()`. That is what users see.

**Your project**
No

**Screenshots**
No

**Environment info**
- OS: Mac
- Python version: 3.8.2
- Sphinx version: 3.1.0dev
- Sphinx extensions: sphinx.ext.autodoc
- Extra tools: No

**Additional context**
No

ppwwyyxx:
I would argue that this is actually not a bug, and the old behavior is usually better than the new one, especially when `functools.wraps` is used.

In fact, `functools.wraps` literally copies attributes such as `__doc__` to the new function, which means the decorated function is meant to be called with the same arguments as the original function. However, a decorator is usually implemented with `*args, **kwargs`, like this:

```
def wrapper(fn):
    @functools.wraps(fn)
    def func(*args, **kwargs):
        do something with arguments
        call fn with arguments
        do something with results
        return results
    return func

@wrapper
def foo(x, y, z):
    return x + y + z
```
To see that this is the common pattern, one can search for `functools.wraps` in any large python code base to verify, e.g.: https://github.com/tensorflow/tensorflow/search?q=%40functools.wraps&unscoped_q=%40functools.wraps https://github.com/django/django/search?q=functools.wraps&unscoped_q=functools.wraps

With the current "bugfix", all decorated function will be rendered with `*args, **kwargs`, which is not useful and worse than the old behavior.


I think the solution can be the following:
1. no choice is perfect - it eventually depends on what user does. So some way to let user config the option might be useful, for example, let users to set `func.__sphinx_use_new_signature__ = True`.
2. If (1) is too hard, I think it's a better decision to keep the old behavior when using `functools.wraps`, because this follows the semantics of `functools.wraps` and the word `wrapping`. This can be done by checking `hasattr(func, __wrapped__)`.
3. For other decorators, I'm not sure which choice is better. But in general I guess it's less likely that the decorator has useful signatures than the decorated function, so I still prefer the original behavior.


[...]

PR #7916, opened 2020-07-04
autodoc: the signature of base function will be shown for decorated functions

tk0miya:
### Feature or Bugfix
- Feature
- Bugfix

### Purpose
- refs: https://github.com/sphinx-doc/sphinx/issues/7650#issuecomment-653700393
- This reverts #7651 (partially)

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1157,10 +1157,7 @@ class FunctionDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # typ
 
         try:
             self.env.app.emit('autodoc-before-process-signature', self.object, False)
-            if inspect.is_singledispatch_function(self.object):
-                sig = inspect.signature(self.object, follow_wrapped=True)
-            else:
-                sig = inspect.signature(self.object)
+            sig = inspect.signature(self.object, follow_wrapped=True)
             args = stringify_signature(sig, **kwargs)
         except TypeError as exc:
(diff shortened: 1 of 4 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -1157,10 +1157,7 @@ def format_args(self, **kwargs: Any) -> str:
 
         try:
             self.env.app.emit('autodoc-before-process-signature', self.object, False)
-            if inspect.is_singledispatch_function(self.object):
-                sig = inspect.signature(self.object, follow_wrapped=True)
-            else:
-                sig = inspect.signature(self.object)
+            sig = inspect.signature(self.object, follow_wrapped=True)
```
harupy, line 1160: @tk0miya Should we specify `follow_wrapeed=True` in `inspect.signature` in **`ClassDocumenter`** as well otherwise a wrapper function's signature (e.g. `(*args, **kwargs)`) would be used when ``__init__`` is decorated?

https://github.com/sphinx-doc/sphinx/blob/38b868cc0d0583d9a58496cd121f0bc345bf9eaa/sphinx/ext/autodoc/__init__.py#L1401

harupy: The same goes for the line below:

https://github.com/sphinx-doc/sphinx/blob/38b868cc0d0583d9a58496cd121f0bc345bf9eaa/sphinx/ext/autodoc/type_comment.py#L122

PR #7922, opened 2020-07-06
Doc: Add missing 'env-merge-info' to the todo tutorial.

### LINKED ISSUE #7898: TODO example is missing env-merge-info
Looks like the [tutorials/todo](https://www.sphinx-doc.org/en/master/development/tutorials/todo.html) example is missing the `env-merge-info` step to make it work for parallel builds.

In cpython we use a similar pattern, and it was not working in parallel builds, I added it like this: https://github.com/python/cpython/pull/21237

Am I right? If so I can add it to the example.

JulienPalard:
Closes #7898.

### DIFF
--- a/doc/development/tutorials/examples/todo.py
+++ b/doc/development/tutorials/examples/todo.py
@@ -61,6 +61,13 @@ def purge_todos(app, env, docname):
                           if todo['docname'] != docname]
 
 
+def merge_todos(app, env, docnames, other):
+    if not hasattr(env, 'todo_all_todos'):
+        env.todo_all_todos = []
+    if hasattr(other, 'todo_all_todos'):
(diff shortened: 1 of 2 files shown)


JulienPalard:
Hi @tk0miya, I tried myself at documenting the missing `env-merge-info`, I tested it locally by generating a bunch of `.rst` files to trigger a parallel build and it works as expected. (Also checked that it in fact don't work when I unlink the event). Would appreciate a review.

Hope my wording is correct, don't hesitate to propose enhancement, I'm french.

Have a good day!

tk0miya (approved):
LGTM!!

tk0miya:
Thank you always :-)

Issue #7921, opened 2020-07-06
Building the Sphinx documentation using 'make latexpdf' fails


brechtm:
Trying to build a PDF of the Sphinx 3.1.2 manual using `make latexpdf` in the docs directory fails with

    ! LaTeX Error: Unknown graphics extension: .svg.

Should `sphinx.ext.imgconverter` be added to the list of extensions in _conf.py_?

tk0miya:
Yes, LaTeX does not support SVG images. Please add some imageconverter extensions (ex. `sphinx.ext.imgconverter`, `sphinxcontrib-svg2pdfconverter`, and so on).

brechtm:
Is it ok for this to be broken in a clean checkout? Shouldn't `sphinx.ext.imgconverter` be added to the list of extensions?

tk0miya:
Oh, sorry. it's my bad. I considered you were saying about your own docs. Okay, I understand. I thought adding `sphinx.ext.imgconverter` requires extra third party tools on building PDF. I don't know it is really worthy.

brechtm:
I would vote to include `sphinx.ext.imgconverter` in the list of extensions. It is included with Sphinx and will should affect the HTML build.

I assume `sphinx.ext.imgconverter` will abort with a clear message to the user when ImageMagick could not be found? Building the Sphinx PDF docs now just crashes latex, which makes it look like there's a bug.

brechtm:
I'm wondering whether the Sphinx PDF docs are generated as part of CI. If so, how is this handled?

mgeier:
We are talking about a single static SVG, right?
What about manually converting it to PDF, storing it in the repo and using `translation.*` as image source?

brechtm:
> What about manually converting it to PDF, storing it in the repo and using `translation.*` as image source?

That crossed my mind as well... and then I forgot. :-)
This is probably the best approach!

tk0miya:
Reasonable. +1 for adding .PDF (or .PNG) file.

tk0miya:
Done in #7924. Closing.

PR #7924, opened 2020-07-07
doc: Add translation.png to build latexpdf successfully

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7921 
- I choose PNG format instead of PDF. It is easier to generate via PlantUML (without setting environment).


tk0miya:
Oops... This contains a broken (empty) PDF file... I'll remove it on another PR.

PR #7926, opened 2020-07-07
Add docs.mesa3d.org to EXAMPLES

kusma:
We started using Sphinx for docs.mesa3d.org, thanks for a great product!


tk0miya:
Thanks!

PR #7937, opened 2020-07-10
Fix #7935: autodoc: A default value inspect._empty conseals signatures

### LINKED ISSUE #7935: Default value of inspect.Parameter.empty breaks autofunction signature
**Describe the bug**
When you have a function with (at least) 2 optional parameters and the default value for the 2nd parameter is `inspect.Parameter.empty`, something goes wrong and the documentation generated for this function will not display
any parameters at all in the signature.

**To Reproduce**
Create a python function like this:
```python
import inspect

def func(optional_arg1='whatever',
         optional_arg2=inspect.Parameter.empty,
         ):
    """lorem ipsum"""
```
Document it with the `autofunction` directive:
```rst
.. autofunction:: func
```
Run `make html` and observe the missing parameters in the output:
```
package_name.func()¶
    lorem ipsum
```

**Expected behavior**
The function's parameters should show up in the documentation. I understand that it's difficult to deal with `inspect.Parameter.empty` as a default value, but completely omitting the function's parameters without any kind
of error or warning is not acceptable.

**Environment info**
- OS: linux
- Python version: 3.8
- Sphinx version: 3.1.2

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7935 
- A function signature is not shown when the function has a parameter
having ``inspect._empty`` as its default value because Signature class
validates function signatures on instantiation.

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -484,7 +484,13 @@ def signature(subject: Callable, bound_method: bool = False, follow_wrapped: boo
             if len(parameters) > 0:
                 parameters.pop(0)
 
-    return inspect.Signature(parameters, return_annotation=return_annotation)
+    # To allow to create signature object correctly for pure python functions,
+    # pass an internal parameter __validate_parameters__=False to Signature
+    #
+    # For example, this helps a function having a default value `inspect._empty`.
+    # refs: https://github.com/sphinx-doc/sphinx/issues/7935
+    return inspect.Signature(parameters, return_annotation=return_annotation,  # type: ignore
+                             __validate_parameters__=False)
 
 
 def stringify_signature(sig: inspect.Signature, show_annotation: bool = True,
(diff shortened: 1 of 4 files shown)


sphinx/util/inspect.py
```diff
@@ -484,7 +484,17 @@ def signature(subject: Callable, bound_method: bool = False, follow_wrapped: boo
...
+        # To allow to create signature object correctly for pure python functions,
+        # pass an internal parameter __validate_parameters__=False to Signature
+        #
+        # For example, this helps a function having a default value `inspect._empty`.
+        # refs: https://github.com/sphinx-doc/sphinx/issues/7935
+        return inspect.Signature(parameters,  # type: ignore
+                                 return_annotation=return_annotation,
+                                 __validate_parameters__=False)
+    except Exception:
+        return inspect.Signature(parameters, return_annotation=return_annotation)
```
tk0miya: It would be better to add a comment here why this block is needed.

tk0miya: I added this block for a fallback when target python does not support `__validate_parameters__` option. But it would be enough even after new python dropped the option. So I drop this except-clause now.

PR #7940, opened 2020-07-10
Fix double trailing newline apidoc

### LINKED ISSUE #7890: apidoc generates an extra newline at the end of the rst file if a module has submodules
**Describe the bug**
This is not really a bug but rather an annoyance if you are using pre-commit hooks that format end of file, as it would often make commits fail that wouldn't otherwise.

**To Reproduce**
Create a package with at least one submodule. and use apidoc on it. Something like this
```
├── docs
│   ├── conf.py
│   ├── index.rst
│   ├── make.bat
│   ├── Makefile
│   ├── package
│   │   ├── modules.rst
│   │   └── package.rst
│   ├── _static
│   └── _templates
└── package
    ├── __init__.py
    └── test.py
```
```
$ sphinx-apidoc -f -q -o package ../package -M 
```
**Expected behavior**
`package.rst` ends with a single newline instead of two.

---

From what I could quickly gather from the codebase this can be easily resolved by adding:
```python
if text.endswith('\n\n'):
  text = text[:-1]
```
to `def write_file(name: str, text: str, opts: Any) -> None:` in `apidoc.py`

AleksaC:
Subject: Fix apidoc generating two newlines at the end of the package rst file if it contains submodules and `--module-first` is used.

### Feature or Bugfix
- Bugfix

### Purpose
- Fix #7890

### DIFF
--- a/sphinx/templates/apidoc/package.rst_t
+++ b/sphinx/templates/apidoc/package.rst_t
@@ -35,14 +35,15 @@ Submodules
 ----------
 {% if separatemodules %}
 {{ toctree(submodules) }}
-{%- else %}
+{% endif %}
+{%- if not separatemodules %}
 {%- for submodule in submodules %}
 {% if show_headings %}
 {{- [submodule, "module"] | join(" ") | e | heading(2) }}
 {% endif %}
 {{ automodule(submodule, automodule_options) }}
 {% endfor %}
-{%- endif %}
+{%- endif -%}
 {% endif %}
 
 {%- if not modulefirst and not is_namespace %}
--- a/sphinx/templates/apidoc/package.rst_t
+++ b/sphinx/templates/apidoc/package.rst_t
@@ -35,16 +35,15 @@ Submodules
 ----------
 {% if separatemodules %}
 {{ toctree(submodules) }}
-{% endif %}
-{%- if not separatemodules %}
+{% else %}
 {%- for submodule in submodules %}
 {% if show_headings %}
 {{- [submodule, "module"] | join(" ") | e | heading(2) }}
 {% endif %}
 {{ automodule(submodule, automodule_options) }}
 {% endfor %}
-{%- endif -%}
-{% endif %}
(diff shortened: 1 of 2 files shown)


sphinx/templates/apidoc/package.rst_t
```diff
@@ -35,14 +35,15 @@ Submodules
...
+{% endif %}
+{%- if not separatemodules %}
 {%- for submodule in submodules %}
 {% if show_headings %}
 {{- [submodule, "module"] | join(" ") | e | heading(2) }}
 {% endif %}
 {{ automodule(submodule, automodule_options) }}
 {% endfor %}
-{%- endif %}
+{%- endif -%}
```
tk0miya: I think using stripping mark at both start and end causes template maintainable. I guess the change above (changing "else" to independent "if" statement) came from this change, right? To keep templates clean, it is better to use one of them consistently.

How about the change like this?
```
@@ -43,9 +46,10 @@ Submodules
 {{ automodule(submodule, automodule_options) }}
 {% endfor %}
 {%- endif %}
-{% endif %}
+{%- endif %}

 {%- if not modulefirst and not is_namespace %}
+
 Module contents
 ---------------
```

Notes:
* Inserts a blank line before section headers to add a space only if the block needed.
* Add stripping-mark to the start of "endif" to strip trailing return codes.

I guess some additional change would be needed for this change. But it will keep our templates clean (if my understanding correct). Could you check this please?

AleksaC:
Thanks for the feedback. I tried what you suggested, but `test_pep_0420_enabled_separate` fails as the file lacks newline at the end and there is no clean way to add it. Another problem occurs with modules without submodules where 2 newlines are added before Module contents. 

If using `-` both at the start and the end of the template is a problem it can be moved to the beginning of the next `{% endif %}`, but even that won't be necessary as I found a way to fix this even without changing else to a separate if statement.

tk0miya:
Thank you for update. LGTM!

PR #7947, opened 2020-07-12
Fix #1362: autodoc: Support private class attributes

### LINKED ISSUE #1362: autodoc private-members doesn't work for class attributes
Documenter.filter_members should take a look at attr_docs in case of private_members.

special-members seems to have the same problem.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1362
- Originally reported by: Anonymous
- Originally created at: 2014-01-27T07:09:45.822

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #1362
- So far, autodoc treats a "private" class attribute as a mere attribute.
But its name is mangled by python interpreter. This make it unmangled
name to be documented expectedly.

### DIFF
--- a/sphinx/ext/autodoc/importer.py
+++ b/sphinx/ext/autodoc/importer.py
@@ -11,7 +11,7 @@
 import importlib
 import traceback
 import warnings
-from typing import Any, Callable, Dict, List, Mapping, NamedTuple, Tuple
+from typing import Any, Callable, Dict, List, Mapping, NamedTuple, Optional, Tuple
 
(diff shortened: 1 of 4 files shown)


sphinx/ext/autodoc/importer.py
```diff
@@ -20,6 +21,38 @@
 
 logger = logging.getLogger(__name__)
 
+mangled_name_re = re.compile(r'^_\w+__\w*[^_]_?$')
```
tk0miya: This regexp is not used.

PR #7661, opened 2020-05-14
Fix autosummary directive wrong processing for invalid modules.

mondeja:
Make the extension `autosummary` not process invalid modules that raises an `ImportError`.

### Feature or Bugfix
- Bugfix

PR #7948, opened 2020-07-13
add item even if it could not be imported

Chilipp:
partially reverts 1dcfc44ace8f3da33f4ee62448aadee67edd80cc from #7661

Subject: Add an item to the autosummary table, even if it cannot be imported

### Feature or Bugfix

- Bugfix

### Purpose
Not sure if this was removed by accident in #7661, but the following line has been removed in 1dcfc44ace8f3da33f4ee62448aadee67edd80cc

https://github.com/sphinx-doc/sphinx/blob/1771bbb9277c3b629ac6e708c4c3f355e034c6cd/sphinx/ext/autosummary/__init__.py#L306

I am readding it here and use the `display_name` instead of the `name` (makes more sense, doesn't it?). The reason, why I am interested in is, that otherwise instance attributes are skipped entirely with the `autosummary` directive.

### Relates
- #7661

### DIFF
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -303,6 +303,7 @@ class Autosummary(SphinxDirective):
             except ImportError:
                 logger.warning(__('autosummary: failed to import %s'), name,
                                location=self.get_source_info())
+                items.append((display_name, '', '', name))
                 continue
 
             self.bridge.result = StringList()  # initialize for each documenter


tk0miya:
It's a main purpose of #7661, so this is expected. But I understand your case. So it would be better to improve autosummary in other way. I consider how to check the existence of instance attributes (without importing). So please wait for a whole.

Chilipp:
I see. Thanks @tk0miya for this feedback. Shall I close this PR then and open an issue instead?

tk0miya:
Yes, please.

Chilipp:
closing this in favor of #8011

PR #7955, opened 2020-07-13
Improve warning message from deprecated_alias

### LINKED ISSUE #7952: Improve deprecated_alias warnings
Consider this warning generated by [`deprecated_alias`](https://github.com/sphinx-doc/sphinx/blob/293fa1e5fc7ee663cb2b2c7e446623773fa9735a/sphinx/directives/__init__.py#L281):
```
>>> from sphinx.directives import SeeAlso
__main__:1: RemovedInSphinx40Warning: sphinx.directives.SeeAlso is deprecated. Check CHANGES for Sphinx API modifications.
```
That makes it sound like `SeeAlso` is being removed in Sphinx 4.0 when in reality it seems to only be that importing from `sphinx.directives` is deprecated. That really isn't clear from the warning.

The warning should specify it's just the alias that's depreacted and tell you what the canonical version is. e.g.

> The alias "sphinx.directives.SeeAlso" is deprecated, use "sphinx.directives.other.SeeAlso" instead. Check CHANGES for Sphinx API modifications.

peterbell10:
Subject: Improve warning message from deprecated_alias

### Feature or Bugfix
- Enhancement

### Purpose
Produce more descriptive warning messages for deprecated aliases. Being clear that only the alias is deprecated, and directing the user to the preferred name.

### Relates
Closes #7952

### DIFF
--- a/sphinx/deprecation.py
+++ b/sphinx/deprecation.py
@@ -11,7 +11,7 @@
 import sys
 import warnings
 from importlib import import_module
-from typing import Any, Dict, Tuple
+from typing import Any, Dict
 
 if False:
     # For type annotation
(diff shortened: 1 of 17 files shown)


sphinx/deprecation.py
```diff
@@ -44,10 +44,13 @@ def __init__(self, module: Any, modname: str, objects: Dict, warning: "Type[Warn
...
         if name in self._objects:
-            warnings.warn("%s.%s is deprecated. Check CHANGES for Sphinx "
-                          "API modifications." % (self._modname, name),
-                          self._warning, stacklevel=3)
-            return self._objects[name]
+            obj = self._objects[name]
+            warnings.warn(
+                "The alias '{}.{}' is deprecated, use '{}.{}' instead. Check CHANGES for "
+                "Sphinx API modifications.".format(
+                    self._modname, name, obj.__module__, obj.__name__),
```
tk0miya: We use `deprecated_alias` not only for class and functions. Is this working fine for constants and other objects?

peterbell10: Okay, I've reworked it so the canonical name is passed to `deprecated_alias`. Going through all the calls also revealed instances where this was being used as a full deprecation so I've separated that use case into `deprecated_attribute`.

sphinx/deprecation.py
```diff
@@ -29,23 +29,54 @@ class RemovedInSphinx50Warning(PendingDeprecationWarning):
...
+    sys.modules[modname] = _ModuleWrapperDeprecatedAlias(  # type: ignore
+        module, modname, objects, warning)
 
 
-class _ModuleWrapper:
-    def __init__(self, module: Any, modname: str, objects: Dict, warning: "Type[Warning]"
-                 ) -> None:
+class _ModuleWrapperDeprecatedAlias:
+    def __init__(self, module: Any, modname: str,
+                 aliases: Dict[str, Tuple[str, object]],
```
tk0miya: Could you add an assertion that confirms the value of `aliases` is a tuple to this constructor? I'm afraid to pass a class object as its value as before.

tk0miya (changes requested):
LGTM with nits

peterbell10:
Looks like `sphinxcontrib` is using `deprecated_alias` as well. Does that mean it needs to be backwards compatible, or can `sphinxcontrib` be updated to support the new calling convention?

tk0miya:
OMG. They should be updated too. And it would be better to be backward compatible because they are used in mixed.

tk0miya:
BTW, I found some testcases are still failed. Could you check the result of CI?

tk0miya (approved):

tk0miya:
Good work! Thank you for your contribution :-)

PR #7956, opened 2020-07-14
Close #3828: :glob: option for toctree directive uses natural sorting

### LINKED ISSUE #3828: toctree globbing uses lexicographical instead of natural sorting
When using a toctree with the `glob` option, the filenames in the list are sorted using [Python's `sorted()`](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/directives/other.py#L78), which does lexicographical sorting.
An example result would be:

- file10.rst
- file2.rst
- file8.rst

This is because in lexicographical sorting 1 < 2, hence `item10` will come before `item2`.
The sorting differs from what operating systems typically do, which would be:

- file2.rst
- file8.rst
- file10.rst

This order makes more sense for humans.

The easiest way to implement this would be to add a new option to the toctree directive (e.g. `:natsort:`) that replaces `sorted()` with `natsorted()` from the [Natsorted module](https://pypi.python.org/pypi/natsort). Just changing the behavior without an option is possible but will generate a different order for some document.
I can make a PR for this, but only if there are no objections that:

1. An additional pip requirement is added
2. This only applies to globbed filenames, not actual document titles (see #1430).

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #3828


tests/test_toctree.py
```diff
@@ -20,8 +20,9 @@ def test_relations(app, status, warning):
     assert app.builder.relations['bar/index'] == ['index', 'foo', 'bar/bar_1']
     assert app.builder.relations['bar/bar_1'] == ['bar/index', 'bar/index', 'bar/bar_2']
     assert app.builder.relations['bar/bar_2'] == ['bar/index', 'bar/bar_1', 'bar/bar_3']
-    assert app.builder.relations['bar/bar_3'] == ['bar/index', 'bar/bar_2', 'bar/bar_4/index']
-    assert app.builder.relations['bar/bar_4/index'] == ['bar/index', 'bar/bar_3', 'baz']
+    assert app.builder.relations['bar/bar_3'] == ['bar/index', 'bar/bar_2', 'bar/bar_10']
+    assert app.builder.relations['bar/bar_10'] == ['bar/index', 'bar/bar_3', 'bar/bar_4/index']
+    assert app.builder.relations['bar/bar_4/index'] == ['bar/index', 'bar/bar_10', 'baz']
```
tk0miya, line 25: We need to discuss whether this is the intended behavior for Sphinx. It might be needed to develop natsort for Sphinx (for toctree).

shimizukawa: Well, .. I think it is not good.
For example, I think natural sorting is used for globing chapter files of a book::
```
chapter1.rst
chapter2.rst
chapter3.rst
chapter4/index.rst
chapter5.rst
```
In this case, putting `chapter4/index` page before `chapter5` is a better result to me.

shimizukawa (commented):
Thanks. I've commented.

tests/test_toctree.py
tk0miya: Agreed. I'll work to make a custom sort module for Sphinx. BTW, do you have any opinion to introduce natural sort (without new option)?

shimizukawa: > option

I don't think it's necessary.
I believe that when document authors use numbers for each files, they will want to naturally sort it.

AA-Turner:
Closing this PR, per the comments we need to split on `/`, and then natural sort the resultant tuples.

A

PR #7957, opened 2020-07-14
Fix #4888: i18n: Failed to add an explicit title to ``:ref:`` role

### LINKED ISSUE #4888: Problems with references when using different referencing types in source file and po file
When using a reference like this in the source file:
`:ref:'english title<sect-title>'`,
and in the po file, translate the same reference as this:
`:ref:'sect-title'`,
the reference is not resolved in the pdf and the html, but "sect-title" is put in the text.

When doing it the other way round, using a reference like this in the source file:
`:ref:'sect-title'`,
and in the po file, translate the same reference as this:
`:ref:'translated title<sect-title>'`,
the reference is resolved correctly, but the "translated title" does not appear. Instead, only the title behind sect-title is written in the text.

(I had to use the simple dash ' instead of the actual ` in the code examples above to get it displayed correctly.)
 
### Environment info
- Sphinx version: 1.7.3

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #4888
- To allow to give (or not to give) an explicit title to the
pending_xref nodes on translation, this skips to override the
attribute on merge original and translated documents.

### DIFF
--- a/sphinx/transforms/i18n.py
+++ b/sphinx/transforms/i18n.py
@@ -37,6 +37,13 @@ if False:
 
 logger = logging.getLogger(__name__)
 
+# The attributes not copied to the translated node
+#
+# * refexplict: For allow to give (or not to give) an explicit title
+#               to the pending_xref on translation
+EXCLUDED_PENDING_XREF_ATTRIBUTES = ('refexplicit',)
+
+
 N = TypeVar('N', bound=nodes.Node)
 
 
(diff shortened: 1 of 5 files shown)


sphinx/transforms/i18n.py
```diff
@@ -426,11 +433,8 @@ def get_ref_key(node: addnodes.pending_xref) -> Tuple[str, str, str]:
                 # Copy attributes to keep original node behavior. Especially
                 # copying 'reftarget', 'py:module', 'py:class' are needed.
                 for k, v in xref_reftarget_map.get(key, {}).items():
-                    # Note: This implementation overwrite all attributes.
-                    # if some attributes `k` should not be overwritten,
-                    # you should provide exclude list as:
-                    # `if k not in EXCLUDE_LIST: new[k] = v`
-                    new[k] = v
+                    if k not in EXCLUDED_PENDING_XREF_ATTRIBUTES:
+                        new[k] = v
```
tk0miya, line 437: I have to add a testcase to confirm this behavior. I'll do it later.

PR #7960, opened 2020-07-14
C, don't deepcopy as deep on handling enumerators

### LINKED ISSUE #559: Performance issues in C domain after scoped rendering changes
While working on upgrading Sphinx/breathe versions for building Zephyr's docs (https://docs.zephyrproject.org/latest/) I found out some real slow-downs when running builds on C domain with newer versions. Those docs are currently built with `Sphinx==2.4.4` and `breathe==4.14.2`, and I am trying to switch to `Sphinx==3.0.4` (or newer) and `breathe==4.19.2`.

Currently there is no `breathe_domain_by_extension`, which means it uses the CPP domain. Here are the numbers:

1) `sphinx==2.4.4` / `breathe==4.14.2`
make html  753.11s user 10.48s system 99% cpu 12:47.00 total

2) `sphinx==3.0.4` / `breathe==4.19.2`
make html  809.96s user 10.10s system 99% cpu 13:45.28 total

Which is not too bad, but if I add a `breathe_domain_by_extension = {"h": "c", "c": "c"}` config I now get this numbers:

1) `sphinx==2.4.4` / `breathe==4.14.2`
make html  682.54s user 11.16s system 99% cpu 11:39.24 total

2) `sphinx==3.0.4` / `breathe==4.19.2`
make html  1389.20s user 7.88s system 99% cpu 23:20.49 total

Going from 682s to 1389s seems a bit excessive. Running with cProfile I see the slowdowns result from calls to `CObject._add_enumerator_to_parent()` -> `ASTBaseBase.clone()` -> `deepcopy()`. Not sure this is helpful or correct, and I wonder why `deepcopy()` calls would slow everything down. The specific commit where the issue starts to happen is this one: https://github.com/michaeljones/breathe/pull/512/commits/a2a1085e24022aececd2b93b46ade684f59d6ec0

jakobandersen:
### Feature or Bugfix
- Bugfix

### Purpose
When an enumerator is added it is automatically copied into the scope of the parent as well. The symbol, along with the complete symbol table and all other declarations of the domain, was not supposed to copied.

### Relates
- Fixes michaeljones/breathe#559

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* C, don't deepcopy the entire symbol table and make a mess every time an
+  enumerator is handled.
+
 Testing
 --------
 
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -1330,6 +1330,10 @@ class ASTDeclaration(ASTBaseBase):
         # set by CObject._add_enumerator_to_parent
         self.enumeratorScopedSymbol = None  # type: Symbol
 
+    def clone(self) -> "ASTDeclaration":
+        return ASTDeclaration(self.objectType, self.directiveType,
+                              self.declaration.clone(), self.semicolon)
+
     @property
     def name(self) -> ASTNestedName:
         return self.declaration.name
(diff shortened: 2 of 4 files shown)


utzig (approved):
LGTM

jakobandersen:
(@tk0miya, I don't think this needs to trigger a new release by it self, as it is "just" a performance regression. So if we reach v3.2 it can be folded into that.)

tk0miya:
I understand. For now, we don't have a critical bug for 3.1.x. So there is no plan to release a bug fix release soon. So this would be released as a part of 3.2.0 (maybe in early August).

PR #7969, opened 2020-07-17
Fix #7968: i18n: process math_block as a literal node

### LINKED ISSUE #7968: translated `.. math::` interprets escapes twice
When rendering a translated `.. math::`, escapes aren't properly ignored when parsing the translation.

This string, properly escaped:
```latex
c_1 \oplus c_2 (\text{définition})
```
gets rendered to 

```latex  
\begin{equation*}
\begin{split}begin{aligned} c_1 oplus c_2 (text{définition}) end{aligned}\end{split}
\end{equation*}
```

my understanding is that `sphinx.transforms.i18n.Locale.apply` calls `publish_msgstr`, which interprets `\` as `\x00` 

**To Reproduce**
```
$ git clone https://github.com/multun/crypto101-sphinx-bugreport
$ cd crypto101-sphinx-bugreport
$ python3 -m sphinx -b latex -E -a -D language='fr'  src _build_fr
$ grep oplus _build_fr/latex/python.tex
```

**Expected behavior**

The fact the text comes from a math section is taken into account, and the text is unchanged during post-gettext translation parsing.

**Your project**
We're translating [Crypto101](https://github.com/crypto101/book), an open source book about cryptography!

**Environment info**
- OS: Linux
- Python version: tested on 3.8.2, 3.7.5
- Sphinx version: tested on master (c2df631), 1.8.5, 2.2.2
- Sphinx extensions:  none
- Extra tools: none

multun:
Subject: enable translating math blocks. Before this PR, `\` weren't preserved.

### Feature or Bugfix
- Bugfix

### Purpose
- Without this patch, you can't translate math blocks

### Detail
- #7968

### DIFF
--- a/sphinx/util/nodes.py
+++ b/sphinx/util/nodes.py
@@ -242,6 +242,7 @@ def is_translatable(node: Node) -> bool:
 LITERAL_TYPE_NODES = (
     nodes.literal_block,
     nodes.doctest_block,
+    nodes.math_block,
     nodes.raw,
 )
 IMAGE_TYPE_NODES = (


tkoyama010:
+0 I can't imagine a case where we have to translate a formula. If you don't mind, could you give me an example?

tk0miya:
Thank you for your contribution. Merged!

PR #7974, opened 2020-07-18
Close #2645: html: Support the Open Graph Protocol (OGP)

### LINKED ISSUE #2645: Meta Directive & Open Graph
I'm currently using the `meta` directive to generate appropriate meta descriptions on my Sphinx pages.

```
.. meta::
   :description: This is a meta description.
```

I'd also like to create Open Graph descriptions for each page, using the same description value in the `meta` directive.

```
<meta name="og:description" content="This is a meta description." />
```

Unfortunately, I'm a bit new to Sphinx and I don't know how to obtain this value from the RST source and inject it into my template. Does anyone know of a way to pull this off?

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #2645 
- This adds a OGP feature to HTML builders.

  * Enable or disable whole of OGP feature via :confval:`html_ogp`
  * Set the default OGP metadata of the document via :confval:
    `html_ogp_meta`
  * Generate OGP metadata automatically from its contents
  * Specify document-specific OGP metadata via docinfo

-  This is based on @shimizukawa's implementation:
    https://github.com/sphinx-contrib/ogp

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -966,6 +966,32 @@ that use Sphinx's HTMLWriter class.
       The image file will be copied to the ``_static`` directory of the output
       HTML, but only if the file does not already exist there.
 
+.. confval:: html_ogp
+
+   If ``True``, HTML builder generates metadata for `The Open Graph protocol`__
+   (OGP) from the content and configurations.  Default is ``True``.
+
+   By setting :confval:`html_baseurl`, the ``og:url`` can be added
+   automatically.
+
+   .. versionadded:: 3.2
+   .. __: https://ogp.me/
+
+.. confval:: html_ogp_meta
+
+   A dictionary for the Open Graph Protocol (OGP).  It is used to the default
+   value of the metadata.  It can be overridden by the :ref:`metadata` in
+   each document.
+
+   Example::
+
+       html_ogp_meta = {
+           "twitter:card": "summary",
+           "twitter:site": "@sphinxdoc"
+       }
+
+   .. versionadded:: 3.2
+
 .. confval:: html_css_files
 
    A list of CSS files.  The entry must be a *filename* string or a tuple
--- a/sphinx/builders/html/ogp.py
+++ b/sphinx/builders/html/ogp.py
@@ -1,0 +1,164 @@
+"""
+    sphinx.builders.html.ogp
+    ~~~~~~~~~~~~~~~~~~~~~~~~
+
+    OGP tag plug-in for HTML builders.
+
+    :copyright: Copyright 2007-2020 by the Sphinx team, see AUTHORS.
+    :license: BSD, see LICENSE for details.
+"""
+
+import posixpath
+from typing import Any, Dict, Generator, List
+
+from docutils import nodes
+from docutils.nodes import Element
+
+from sphinx.application import Sphinx
+from sphinx.util.osutil import canon_path
+
+
+class OGDescriptionCollector(nodes.SparseNodeVisitor):
+    """A description collector module for OGP metadata."""
+
+    def __init__(self, document: nodes.document) -> None:
+        super().__init__(document)
+        self.descriptions = []  # type: List[str]
+        self.sections = 0       # type: int
+
+    def visit_compact_paragraph(self, node: Element) -> None:
+        if node.get('toctree'):
+            # ignore toctree
+            raise nodes.SkipChildren
+
+    def visit_section(self, node: Element) -> None:
+        self.sections += 1
+        if self.sections >= 3:
+            # stop processing when the 3rd section found
+            raise nodes.StopTraversal
+
+    def visit_paragraph(self, node: Element) -> None:
+        # collect paragraph as a part of descriptions
+        self.descriptions.append(node.astext())
+
+
+class OGImageCollector(nodes.SparseNodeVisitor):
+    """An image collector module for OGP metadata.
+
+    .. note:: This will be removed on dropping docutils-0.15 support.
+              Since 0.16, this class will be replaced by
+              ``Element.traverse()``.
+    """
+    def __init__(self, document: nodes.document) -> None:
+        super().__init__(document)
+        self.image = None  # type: nodes.image
+
+    def visit_image(self, node: nodes.image) -> None:
+        self.image = node
+        raise nodes.StopTraversal
+
+
+def get_ogpmeta_from_context(context: Dict) -> Dict:
+    """Pick OGP metadata up from the metadata of the document."""
+    htmlmeta = context.get('meta') or {}
+    return {k: v for k, v in htmlmeta.items() if k.startswith('og:')}
+
+
+def apply_html_ogp_meta(app: Sphinx, ogpmeta: Dict) -> None:
+    """Apply :confval:`html_ogp_meta` to document specific OGP metadata."""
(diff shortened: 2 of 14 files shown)


Daltz333:
A similar implementation is https://github.com/wpilibsuite/sphinxext-opengraph

One thing that would be nice is automatically configuring things like og:locale based on the current set locale.

tk0miya:
I thought using `language` configuration to the `og:locale`. But it does not contain "territory" info. ogp.me says its format is "language_TERRITORY".

>og:locale - The locale these tags are marked up in. Of the format language_TERRITORY. Default is en_US.
>https://ogp.me/

In addition, I think the system locale is not related to the document's locale. So I think it would be better to give it via `html_ogp_meta` option.

doc/usage/configuration.rst
```diff
@@ -966,6 +966,29 @@ that use Sphinx's HTMLWriter class.
...
+
+   Example::
+
+       html_ogp_meta = {
+           "twitter:card": "summary",
+           "twitter:site": "@sphinxdoc"
+       }
+
+   .. versionadded:: 3.2
+
```
shimizukawa, line 994: I think it would be better if document has a mention for 'html_baseurl' to generate `og:url` meta like:
```
By setting `html_baseurl`, the `og:url` meta tag can be added to the HTML.
```

sphinx/builders/html/ogp.py
```diff
@@ -0,0 +1,159 @@
...
+        # OGP feature is disabled
+        return
+
+    ogpmeta = get_ogpmeta_from_context(context)
+
+    if 'og:image' in ogpmeta:
+        # convert document specific og:image to a relative path from current document
+        ogimage = ogpmeta['og:image']
+        relpath, abspath = app.env.relfn2path(ogimage, pagename)
+        ogpmeta['og:image'] = canon_path(relpath)
```
shimizukawa: `og:image` should have absolute URL, so if `html_baseurl` is provided, we should use it for `og:image` url.

refs : https://stackoverflow.com/a/9858694/764538

sphinx/builders/html/ogp.py
```diff
@@ -0,0 +1,159 @@
...
+
+    apply_html_ogp_meta(app, pagename, ogpmeta)
+    extract_ogpmeta_from_content(app, pagename, context, ogpmeta, doctree)
+
+    context['metatags'] += '\n'.join(_generate_ogp_tags(context, ogpmeta))
```
shimizukawa: MAY: it would be better if...

* metatags starts with "\n" because the `metatags` will be added to the end of the previous line. see: https://github.com/sphinx-doc/sphinx/blob/aeed7117955cd5f981a7382b5cb70d1039d1d94f/sphinx/themes/basic/layout.html#L123-L124
* metatags have indentation with 4 spaces.

---------------

for example, this is current output
```
<html lang="ja">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" /><meta property="twitter:card" content="summary" />
<meta property="twitter:site" content="@sphinxdoc" />
<meta property="og:type" content="website" />
...
```

this is better output
```
<html lang="ja">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <meta property="twitter:card" content="summary" />
    <meta property="twitter:site" content="@sphinxdoc" />
    <meta property="og:type" content="website" />
...
```

sphinx/builders/html/ogp.py
```diff
@@ -0,0 +1,159 @@
...
+
+        ogpmeta.setdefault(key, value)
+
+
+def extract_ogpmeta_from_content(app: Sphinx, docname: str, context: Dict,
+                                 ogpmeta: Dict, doctree: nodes.document) -> None:
+    """Extract OGP metadata from contents."""
+    # fill system-default OGP metadata
+    ogpmeta.setdefault('og:type', 'website')
+    ogpmeta.setdefault('og:title', context['title'])
```
shimizukawa: `context['title']` contains html tag if it belongs `:numbered:` toctree.
Code for `context['title']` is this: https://github.com/sphinx-doc/sphinx/blob/5850d6b8f9569171d73f8bc5abbc9d646f5d4a77/sphinx/builders/html/__init__.py#L552

Rendered title_node is this:
```
(Pdb) self.render_partial(title_node)['title']
'<span class="section-number">1. </span>Installation'
```
and sphinx generate incorrect html tags:
```
<meta property="og:title" content="<span class="section-number">1. </span>Installation" />
```

(my original code also have this problem.. :(

shimizukawa (commented):
Thanks! I've commented.

sphinx/builders/html/ogp.py
tk0miya: Absolutely. Do it later.

doc/usage/configuration.rst
tk0miya: done (in the description for `html_ogp`).

sphinx/builders/html/ogp.py
tk0miya: Okay, I'll take the title from `env.titles`.

sphinx/builders/html/ogp.py
tk0miya: done

sphinx/builders/html/ogp.py
tk0miya: done

Daltz333:
I think there should be a distinction between things that use OpenGraph, and custom implementation such as Discourse, Discord, Twitter, and Facebook all have their own configurations and the such. Additionally, I think having a catch all for all extra metadata  via html_ogp_meta is redundant, as there should be a feature to embed metadata anyways. Options should be exposured on a per OpenGraph option basis. Additionally, OpenGraph supports multiple images in image metadata.

tk0miya:
>I think there should be a distinction between things that use OpenGraph, and custom implementation such as Discourse, Discord, Twitter, and Facebook all have their own configurations and the such.

I'm not good at the differences. Could you tell me your idea in details?

>Additionally, I think having a catch all for all extra metadata via html_ogp_meta is redundant, as there should be a feature to embed metadata anyways.

Sorry, I don't know what are you saying. 

>Options should be exposured on a per OpenGraph option basis.

It seems there are many kinds of metadata defined in https://ogp.me/. Do you intend to make options for all of them? What is an advantage to `html_ogp_meta`?

>Additionally, OpenGraph supports multiple images in image metadata.

Indeed. But I don't have the way to support it.

ItayZiv:
I'm a contributor on that repo, and for us anything ogp is everything defined on https://ogp.me/ , anything else is custom metadata and we let users pass it in or use some other way to insert that.

I am currently working on a significant rewrite with a few of my idea on how to greatly enhance the extension, this includes locale but also stuff like a directive that can be used to override settings per page since most content in Sphinx is considered an article, this would be configuring the author and release date for example.

Daltz333:
Additionally we plan to have integrated support for locales and readthedocs, which currently isn't in this revision. https://github.com/wpilibsuite/sphinxext-opengraph/issues

It may be better to just point users toward this extension instead of integrating it directly.

tk0miya:
It seems sphinxext-opengraph is much better than my PR. So I withdraw this PR now.

PR #7978, opened 2020-07-18
Fix #7973: imgconverter: Check availability of imagemagick many times

### LINKED ISSUE #7973: ImgConverter runs is_available in HTML builder
Since ImgConverter is only used in the LaTeX builder, this check should not happen in HTML.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #7973 
- To prevent checking the availability of image converters times and times
again, this stores the result to the class variable.  It is not a good
design to have a state globally.  So this should be refactored in the
future.

### DIFF
--- a/sphinx/transforms/post_transforms/images.py
+++ b/sphinx/transforms/post_transforms/images.py
@@ -11,7 +11,7 @@
 import os
 import re
 from math import ceil
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from docutils import nodes
 
@@ -175,6 +175,13 @@ class ImageConverter(BaseImageConverter):
     """
     default_priority = 200
 
+    #: The converter is available or not.  Will be filled at the first call of
+    #: the build.  The result is shared in the same process.
+    #:
+    #: .. todo:: This should be refactored not to store the state without class
+    #:           variable.
+    available = None  # type: Optional[bool]
+
     #: A conversion rules the image converter supports.
     #: It is represented as a list of pair of source image format (mimetype) and
     #: destination one::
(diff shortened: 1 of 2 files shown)


Daltz333:
Probably should rename this PR/commit to "Check availability of ImageMagick only once"

Daltz333:
Oh sorry, it's the name of the bug. My bad

Daltz333 (approved):

tk0miya:
>only once

It would be wonderful if true. But this fix is not so. Even this change, the check will be called some times on parallel build. Because the result of the check is not shared between processes.

sphinx/transforms/post_transforms/images.py
```diff
@@ -187,16 +194,14 @@ class ImageConverter(BaseImageConverter):
...
-                                # Will be checked at first conversion
         super().__init__(*args, **kwargs)
 
     def match(self, node: nodes.image) -> bool:
         if not self.app.builder.supported_image_types:
             return False
         elif self.available is None:
-            self.available = self.is_available()
+            # store the value to the class variable to share it during the build
+            self.__class__.available = self.is_available()
```
shimizukawa, line 204: How about using [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) ?

```
>>> import functools
>>> import random
>>> class F:
...   @functools.lru_cache
...   def rand(self):
...     return random.randint(1,10)
...
>>> f = F()
>>> f.rand()
9
>>> f.rand()
9
>>> f2 = F()
>>> f2.rand()
9
>>> f3 = F()
>>> f3.rand()
9
```

tk0miya: Good point. If we choose `lru_cache()`, it is recommended to all subclasses to wrap `is_available()` method by `lru_cache()`. But it's natural design to me. So +0 for your proposal.

Daltz333:
Any update on this? @tk0miya

sphinx/transforms/post_transforms/images.py
tk0miya: I tested `functools.lru_cache()`. But it does not share the cache between instances. I guess different first argument `self` is passed on each instance.
```
$ python
Python 3.8.6 (default, Sep 27 2020, 22:28:24)
[Clang 11.0.3 (clang-1103.0.32.59)] on darwin
>>> import functools
>>> import random
>>> class F:
...   @functools.lru_cache
...   def rand(self):
...     return random.randint(1,10)
...
>>> f = F()
>>> f.rand()
8
>>> f.rand()
8
>>> f2 = F()
>>> f2.rand()
7
>>> f3 = F()
>>> f3.rand()
7
>>> f4 = F()
>>> f4.rand()
5
```

PR #7981, opened 2020-07-19
[WIP] Fix #7498: understand operator precedence

### LINKED ISSUE #7498: sphinx.pycode.ast.unparse does not understand operator precedence
As an example:
```python
>>> import sphinx.pycode.ast as ast
>>> ast.unparse(ast.parse("(1 + 2) * 3").body[0].value)
'1 + 2 * 3'
>>> ast.unparse(ast.parse("(1, 2) + (3, 4)").body[0].value)
'1, 2 + 3, 4'
```

Easiest way to fix this will be to copy some of the approach from the python 3.9 `ast.unparse`, which needs #7497 in order to keep track of the current precedence level.

tkoyama010:
Subject: Fix #7498: understand operator precedence

### Feature or Bugfix

- Bugfix

### Purpose
- Fix #7498: understand operator precedence

### Detail
- None

### Relates
- None

### DIFF
--- a/sphinx/pycode/ast.py
+++ b/sphinx/pycode/ast.py
@@ -69,137 +69,835 @@ def unparse(node: Optional[ast.AST]) -> Optional[str]:
 
 # a greatly cut-down version of `ast._Unparser`
 class _UnparseVisitor(ast.NodeVisitor):
+    """Methods in this class recursively traverse an AST and
+    output source code for the abstract syntax; original formatting
+    is disregarded."""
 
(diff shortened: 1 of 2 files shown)


tkoyama010:
As said in #7498 we can use source of python.
https://github.com/python/cpython/blob/c53b310e5926266ce267c44a168165cacd786d6e/Lib/ast.py#L1486-L1488

tk0miya:
I merged #8265 instead of this. Thank you for your work!

PR #7986, opened 2020-07-19
basic CSS: fix "highlight-" selector

mgeier:
In 0dc6e3ac458dd6c10ed2c7bd03ca95bd694d7ec6 I've added

```css
div[class^="highlight-"] {
    margin: 1em 0;
}
```

and in 03be040ca8f5aa6c52b19cf3bf3ab10e284159f6 I've added

```css
div[class|="highlight"] {
    clear: both;
}
```

In the first case I wanted to select `highlight-*` but not `highlight`.
In the second case I didn't care whether `highlight` was selected or not.

The problem is that both selectors only work if they are at the beginning of the `class` attribute.
If they are not (which happens e.g. in a `compound` directive), the selectors don't work as expected.

In this PR, I'm suggesting to change both cases to `div[class*="highlight-"]` which should work in both cases as expected.

This also selects classes which have `highlight-` somewhere in the middle or at the end, but I hope such class names are not used in practice.

### Feature or Bugfix

- Bugfix

### Relates

#7540
#7657

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -86,6 +86,7 @@ Bugs fixed
 * #7619: Duplicated node IDs are generated if node has multiple IDs
 * #2050: Symbols sections are appeared twice in the index page
 * #8017: Fix circular import in sphinx.addnodes
+* #7986: CSS: make "highlight" selector more robust
 
 Testing
 --------
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -693,7 +693,7 @@ pre {
     overflow-y: hidden;  /* fixes display issues on Chrome browsers */
 }
 
-pre, div[class|="highlight"] {
+pre, div[class*="highlight-"] {
     clear: both;
 }
 
@@ -704,7 +704,7 @@ span.pre {
     hyphens: none;
 }
 
-div[class^="highlight-"] {
+div[class*="highlight-"] {
     margin: 1em 0;
 }
 


tk0miya (changes requested):
I can't imagine what happens with this change. Could you upload a screenshot if you have? In addition, please add a CHANGES entry.

tk0miya:
It seems you changed the target branch. Is this a critical bug? Only critical bug fixes are allowed to 3.1.x branch.

mgeier:
> Is this a critical bug?

No. I've re-based this PR onto the `3.x` branch and I've changed the base branch here on Github.

> Could you upload a screenshot if you have?

Sure. Here's an example for the `margin` problem:

```rst
.. image:: https://docs.python.org/3/_static/py.png

.. compound::

    .. code-block:: python

        print('hello')

    .. code-block:: python

        print('world')

.. image:: https://docs.python.org/3/_static/py.png
```

**before** (using the ` sphinxdoc` theme):

image

Obviously, the `margin` is missing.

**after**:

image

The second problem (with `clear`) isn't really visible, but it can be experienced with this example:

```rst
.. sidebar:: A Sidebar

    Some text in the sidebar.

.. compound::

    .. code-block:: python

        print('hello')

    Some text in the compound paragraph.
```

When looking at this `<div>` in the browser's developer tools:

```html
<div class="compound-first highlight-python notranslate">
...
</div>
```

... we can see that it covers the sidebar.

With this PR, it does not cover the sidebar.

None of this is visible with the HTML themes I tried, but it might be visible with a different one?

> In addition, please add a CHANGES entry.

Done: 416438366d0df4cef446e302a2b8a7fd9f1ae5be

tk0miya (approved):

tk0miya:
Thank you for your explanation. Merging now.

PR #7990, opened 2020-07-22
adds ability for users to provide a custom pygments Style object

jmaggio14:
Subject: adds ability for users to provide a custom pygments Style object

### Feature or Bugfix
- Feature

### Purpose
Allows users to pass in a pygments Style or a string for the conf.py _pygments_style_ parameter 

This would save them the trouble of adding the Style to their module, making a separate plugin, or monkey patching in a change to pygments directly (as is the in the link below). This was an annoying problem for our project when we wanted to implement custom syntax highlighting, but couldn't figure out the `fully-qualified name of a custom Pygments style class` (link to docs below)

### Relates
docs for  _pygments_style_ parameter: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-pygments_style

monkey patching example: https://stackoverflow.com/questions/48615629/how-to-include-pygments-styles-in-a-sphinx-project)

### DIFF
--- a/sphinx/highlighting.py
+++ b/sphinx/highlighting.py
@@ -10,7 +10,7 @@
 
 from functools import partial
 from importlib import import_module
-from typing import Any, Dict
+from typing import Any, Dict, Union
 
 from pygments import highlight
 from pygments.filters import ErrorToken
@@ -60,7 +60,7 @@ class PygmentsBridge:
     html_formatter = HtmlFormatter
     latex_formatter = LatexFormatter
 
-    def __init__(self, dest: str = 'html', stylename: str = 'sphinx',
+    def __init__(self, dest: str = 'html', stylename: Union[str,Style] = 'sphinx',
                  latex_engine: str = None) -> None:
         self.dest = dest
         self.latex_engine = latex_engine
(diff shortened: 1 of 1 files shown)


jmaggio14:
fixes linter issue

tk0miya (changes requested):
+1 for the idea. Could you update document of `pygments_style` and add a testcase for this?

tk0miya:
Now I change the milestone of this PR to 3.3.0 release (almost 2 months). Please update this until then.

sphinx/highlighting.py
```diff
@@ -73,14 +73,18 @@ def __init__(self, dest: str = 'html', stylename: str = 'sphinx',
             self.formatter = self.latex_formatter
             self.formatter_args['commandprefix'] = 'PYG'
 
-    def get_style(self, stylename: str) -> Style:
+    def get_style(self, stylename: Union[str, Style]) -> Style:
         if stylename is None or stylename == 'sphinx':
             return SphinxStyle
         elif stylename == 'none':
             return NoneStyle
         elif '.' in stylename:
```
tk0miya, line 81: I guess a style object get a failure with this if-block. Is this PR well tested at local?
As I already commented, a testcase is needed to merge this.

tk0miya:
I noticed `pygments_style` configuration supports a fully-qualified python module. It means it's not hard to add a custom style now. So I change my opinion not to merge this into Sphinx.

>Changed in version 0.3: If the value is a fully-qualified name of a custom Pygments style class, this is then used as custom style.

PR #7994, opened 2020-07-22
Add option to use old node-ids in option directive

utzig:
This adds a new option, "compat-id" to option directives, to revert to using the old normalization code, which prevented uppercase, '_' and '.' symbols. It avoids breaking html links to anchors.

This updates #7765, because that implementation was wrong and got stalled. The new proposal allows one to describe options like this:

```
.. option:: SOME_OPTION_HERE
   :compat-id:
```

Which will fallback to the using the old id naming code.

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -223,6 +223,11 @@ class Cmdoption(ObjectDescription):
             node_id = make_id(self.env, self.state.document, prefix, optname)
             signode['ids'].append(node_id)
 
+            old_node_id = self.make_old_id(prefix, optname)
+            if old_node_id not in self.state.document.ids and \
+               old_node_id not in signode['ids']:
+                signode['ids'].append(old_node_id)
+
         self.state.document.note_explicit_target(signode)
 
         domain = cast(StandardDomain, self.env.get_domain('std'))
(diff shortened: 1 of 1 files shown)


sphinx/domains/std.py
```diff
@@ -223,12 +228,22 @@ def add_target_and_index(self, firstname: str, sig: str, signode: desc_signature
             node_id = make_id(self.env, self.state.document, prefix, optname)
             signode['ids'].append(node_id)
 
+            if 'compat-id' in self.options:
```
tk0miya: Please add an old-styled ID to the signode always. Any new option is not needed to do that.

sphinx/domains/std.py
```diff
@@ -223,12 +228,22 @@ def add_target_and_index(self, firstname: str, sig: str, signode: desc_signature
...
+                   old_node_id not in signode['ids']:
+                    signode['ids'].append(old_node_id)
+
         self.state.document.note_explicit_target(signode)
 
         domain = cast(StandardDomain, self.env.get_domain('std'))
         for optname in signode.get('allnames', []):
-            domain.add_program_option(currprogram, optname,
-                                      self.env.docname, signode['ids'][0])
+            if 'compat-id' in self.options:
```
tk0miya: I think it's enough to append an old-styled ID to the signode to keep old hyperlinks available. You don't need to register an old-styled ID to the domain.

sphinx/domains/std.py
```diff
@@ -237,7 +252,20 @@ def add_target_and_index(self, firstname: str, sig: str, signode: desc_signature
             descr = _('command line option')
         for option in sig.split(', '):
             entry = '; '.join([descr, option])
-            self.indexnode['entries'].append(('pair', entry, signode['ids'][0], '', None))
+            if 'compat-id' in self.options:
```
tk0miya: This is also.

tk0miya (changes requested):

sphinx/domains/std.py
utzig: Done! I totally missed that for every `signode` added the html writer generates a span with the proper `id`, so removed the setting.

sphinx/domains/std.py
utzig: Done.

tk0miya (approved):

tk0miya:
Merged!

Issue #6863, opened 2019-11-25
Document literalinclude before-start end-after same string/Doxygen compatibility


nilsdeppe:
First of all, thanks for creating and maintaining Sphinx! It's really awesome :heart: 

**Is your feature request related to a problem? Please describe.**
I want to have the same comment line in the code twice around the block of code I'd like included. For example,
```cpp
# [[awesome_code_block]]
print("I'm doing something")
# [[awesome_code_block]]
```
Sphinx's `literalinclude` directive supports this!
```rst
.. literalinclude:: file.py
   :start-after: [[awesome_code_block]]
   :end-before: [[awesome_code_block]]
```
Which renders the code exactly as I want :heart: 
However, this support isn't made clear in the documentation:
```
Another way to control which part of the file is included is
to use the start-after and end-before options (or only one of them).
If start-after is given as a string option, only lines that follow the first
line containing that string are included. If end-before is given as a
string option, only lines that precede the first lines containing that
string are included.
```
and I'm not sure if there is a test for this feature.

My motivation for having this feature is that Doxygen (a C++ documentation generator) uses snippets in the way I've shown above and we're looking at moving from Doxygen to Sphinx. Snippet support that is compatible with both Sphinx and Doxygen was something we wanted to maintain.

**Describe the solution you'd like**
I think documenting the possibility of having the begin and end lines be comments in the source code and have the same text would resolve the issue. I'm not sure if there is a test for this behavior in Sphinx (or how testing is done within Sphinx), but that would be a (separate) but additional nice thing to have. I also think that having an example of this feature in the documentation would be nice. The relevant paragraph in the docs is:
```
Another way to control which part of the file is included is to use the
start-after and end-before options (or only one of them). If start-after
is given as a string option, only lines that follow the first line containing
that string are included. If end-before is given as a string option, only
lines that precede the first lines containing that string are included. The
start-at and end-at options behave in a similar way, but the lines containing
the matched string are included.
```
Maybe clarifying that
```
If both start-after and end-before are given, then the block ends at the first
occurrence of the end-before string that appears in the file after the start-before
string.
```

ass-spec:
If both start-after and end-before are given, then the block ends at the first
occurrence of the end-before string that appears in the file after the start-before
string.

tk0miya:
Yes, Sphinx supports such usecase. Could you send us a pull request please? It would be helpful to me!

shibukawa:
[...]

PR #7997, opened 2020-07-23
fix 6863: add description literalinclude option (start/end have same string)

shibukawa:
Subject: Add description to document about literalinclude directive

### Feature or Bugfix
- document

### Purpose
There are no information about the behavior when literalinclude's start-at/after, end-at/before options have same string. This PR adds it.

### Relates
https://github.com/sphinx-doc/sphinx/issues/6863

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -656,9 +656,43 @@ __ http://pygments.org/docs/lexers
    string are included. The ``start-at`` and ``end-at`` options behave in a
    similar way, but the lines containing the matched string are included.
 
-   With lines selected using ``start-after`` it is still possible to use
-   ``lines``, the first allowed line having by convention the line number
-   ``1``.
+   ``start-after``/``start-at`` and ``end-before``/``end-at`` can have same string.
+   ``start-after``/``start-at`` filter lines before the line that contains
+   option string (``start-at`` will keep the line). Then ``end-before``/``end-at``
+   filter lines after the line that contains option string (``end-at`` will keep
+   the line and ``end-before`` skip the first line).
+
+   .. note::
+
+      If you want to select only ``[second-section]`` of ini file like the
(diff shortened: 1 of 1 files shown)


doc/usage/restructuredtext/directives.rst
```diff
@@ -656,6 +656,10 @@ __ http://pygments.org/docs/lexers
    string are included. The ``start-at`` and ``end-at`` options behave in a
    similar way, but the lines containing the matched string are included.
 
+   If ``start-after``/``start-at`` and ``end-before``/``end-at`` have same strings,
+   ``end-before``/``end-at`` will match with the second line that contains the
```
tk0miya: This is not incorrect. Regardless of the existence of `start-after` or `start-at`, `end-before` option always skip the first line. On the other hand, `end-at` tries to match the first line always even if `start-*` option given.

I think it would be better to add a tip for the combination of `start-after` and `end-before` instead.

tk0miya (changes requested):

doc/usage/restructuredtext/directives.rst
```diff
@@ -656,9 +656,42 @@ __ http://pygments.org/docs/lexers
...
+   ``start-after``/``start-at`` and ``end-before``/``end-at`` can have same string.
+   ``start-after``/``start-at`` filter lines before the line that contains
+   option string (``start-at`` will keep the line). Then ``end-before``/``end-at``
+   filter lines after the line that contains option string (``end-at`` will keep
+   the line and ``end-before`` skip the first line).
+
+   .. note::
+
+      If you want to select only ``[second-section]`` of ini file like the following,
+      you can use ``:start-at: [second-section]`` and ``:end-before: [third-section]``:
```
tk0miya: Could you fold the text by 80chars?

tk0miya (approved):

tk0miya:
Thanks!

PR #8010, opened 2020-07-27
adding a link to the docutils.nodes documentation to make it easier f…

maltfield:
…or sphinx extension developers to find the list of all useable nodes

Subject: Update docs to link to docutils.nodes

### Feature or Bugfix

- Documentation

### Purpose
- Help devs writing sphinx extensions find the list of useable sphinx nodes

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>

### DIFF
--- a/doc/extdev/nodes.rst
+++ b/doc/extdev/nodes.rst
@@ -3,6 +3,8 @@
 Doctree node classes added by Sphinx
 ====================================
 
+.. note:: This page does not list the complete set of nodes available in sphinx. Please also see the `nodes provided by docutils.nodes <http://epydoc.sourceforge.net/docutils/private/docutils.nodes-module.html>`_
+
 .. module:: sphinx.addnodes
 
 Nodes for domain-specific object descriptions


tk0miya:
How about adding docutils' official document instead?
https://docutils.sourceforge.io/docs/ref/doctree.html

maltfield:
Does that list all the nodes? I did a ctrl+f on that page, and I only found one instance of "nodes," and it said that the reader should instead reference the implentation of the `docutils.nodes` module (but it also provides no link to said documentation)

> This document does not discuss implementation details. Those can be found in internal documentation (docstrings) for the docutils.nodes module, where the document tree data structure is implemented in a class library.

Assuming that "nodes" is the same thing as "elements," then the document you provided is not as complete as the one I provided. For example, the `target` section just says "To be completed". In fact, a ton of their elements just say "to be completed"

This PR is to help guide the user trying to write a sphinx extension to know what nodes are available for their use with `nodes.X()`. I created it after I landed on this page looking to add what I later learned to be a target with `nodes.target( '', '' )`, but the sphix docs didn't really help me with this. The link I'm adding with this PR's commits does.

doc/extdev/nodes.rst
```diff
@@ -3,6 +3,8 @@
 Doctree node classes added by Sphinx
 ====================================
 
+.. note:: This page does not list the complete set of nodes available in sphinx. Please also see the `nodes provided by docutils.nodes <http://epydoc.sourceforge.net/docutils/private/docutils.nodes-module.html>`_
```
tk0miya, line 6: It seems this page has been no longer maintained. This is a document for docutils-0.3.0. It's better to use https://docutils.sourceforge.io/docs/ref/doctree.html instead. This is well maintained upstream's page.

tk0miya (changes requested):

doc/extdev/nodes.rst
tk0miya: Oops. I already commented as same...

tk0miya:
>Does that list all the nodes?

Maybe, yes. On the other hand, the page you pasted does not contains `math` and `math_block` nodes. Because it only describes nodes of docutils-0.3 (17th years ago!).

>(but it also provides no link to said documentation)

The page you used is not an official page. It was created by epydoc project.

PR #8025, opened 2020-08-01
Close #6698: doctest: Add :trim-flags: and :no-trim-flags: options

### LINKED ISSUE #6698: doctest: Need a way to locally disable the trimming of doctest flags

In cpython, we have the [doctest documentation](https://docs.python.org/3/library/doctest.html), in this we're having for example:
```rst
This test also passes, and also requires a directive to do
so::

   >>> print(list(range(20))) # doctest: +ELLIPSIS
   [0, 1, ..., 18, 19]
```
So in this specific case, we need the "# doctest: +ELLIPSIS" to still be visible in the generated documents, because it's part of what's we're documenting.

I'd like to be able to locally disable the stripping by something like:
```rst
.. doctest::
   :trim_flags: False

   >>> print(list(range(20))) # doctest: +NORMALIZE_WHITESPACE
   [0,   1,  2,  3,  4,  5,  6,  7,  8,  9,
   10,  11, 12, 13, 14, 15, 16, 17, 18, 19]
```

In cpython we were messing with `self.highlighter.trim_doctest_flags`, which is now deprecated. `trim_doctest_flags` still exists but as a global option, so it's harder to trick it to act on a single code block with not too much ugly tricks.

**Additional context**
https://bugs.python.org/issue36675

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #6698 
- To control trimming doctest flags manually, this adds new options
:trim-flags: and :no-trim-flags: to doctest directives.  It helps
to describes doctest module itself in python doc (see #6698).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -42,6 +42,8 @@ Features added
 * #7768: i18n: :confval:`figure_language_filename` supports ``docpath`` token
 * #5208: linkcheck: Support checks for local links
 * #5090: setuptools: Link verbosity to distutils' -v and -q option
+* #6698: doctest: Add ``:trim-doctest-flags:`` and ``:no-trim-doctest-flags:``
+  options to doctest, testcode and testoutput directives
 * #7052: add ``:noindexentry:`` to the Python, C, C++, and Javascript domains.
   Update the documentation to better reflect the relationship between this option
   and the ``:noindex:`` option.
--- a/doc/usage/extensions/doctest.rst
+++ b/doc/usage/extensions/doctest.rst
@@ -67,7 +67,7 @@ a comma-separated list of group names.
    default set of flags is specified by the :confval:`doctest_default_flags`
    configuration variable.
 
-   This directive supports three options:
+   This directive supports five options:
 
    * ``hide``, a flag option, hides the doctest block in other builders.  By
      default it is shown as a highlighted doctest block.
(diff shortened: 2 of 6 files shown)


CHANGES
```diff
@@ -34,6 +34,8 @@ Features added
 * #7840: i18n: Optimize the dependencies check on bootstrap
 * #5208: linkcheck: Support checks for local links
 * #5090: setuptools: Link verbosity to distutils' -v and -q option
+* #6698: doctest: Add ``:trim-flags:`` and ``:no-trim-flags:`` options to
+  doctest, testcode and testoutput directives
```
shimizukawa: I found it hard to understand "flag to trim flags".
Also, I think the only time this flag is used is rarely such as explaining doctest itself.

So, I suggest names as: `:trim-doctest-flags:` and `:no-trim-doctest-flags:`

shimizukawa (approved):
LGTM with 1 naming suggestion.

CHANGES
tk0miya: Reasonable.

JulienPalard:
Hi @tk0miya, can this  be backported to Sphinx 2? It would greatly help distribs having Sphinx 2 and wanting to ship Python 3.10, which relies on `:no-trim-flags:`.

tk0miya:
@JulienPalard Sorry, the only what we can do is supporting the latest release. No time and power to support old versions...

JulienPalard:
I understand. Do you have a great idea how we could use :no-trim-flags: while keeping compat with old Sphinx?

tk0miya:
Only we can do is patching the doctest directive on building the old Sphinx.

PR #8042, opened 2020-08-02
Fix #8041: autodoc: An ivar on super class is not shown unexpectedly

### LINKED ISSUE #8041: An annotated instance variable on super class is not documented when derived class has other annotated instance variables
**Describe the bug**
An annotated instance variable on super class is not documented when derived class has other annotated instance variables

**To Reproduce**
```
# example.py
class Foo:
    var1: int


class Bar(Foo):
    var2: int
```
```
# index.rst
.. automodule:: example
   :members:
   :undoc-members:
   :inherited-members:
```

`Bar.var1` is not shown even if `:inherited-members:` option given.

**Expected behavior**
Both `var1` and `var2` are shown.

**Your project**
No

**Screenshots**
No

**Environment info**
- OS: Mac
- Python version: 3.8.2
- Sphinx version: HEAD of 3.x

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #8041 
- An annotated instance variable on super class is not documented when
derived class has also other annotated instance variables because
`obj.__annotations__` is overrided by derived class's type hints.
- To get annotations of the target class correctly, this scans MRO to
get all of annotated instance variables.

### DIFF
--- a/tests/test_ext_autodoc.py
+++ b/tests/test_ext_autodoc.py
@@ -1580,12 +1580,7 @@ def test_autodoc_typed_instance_variables(app):
         '   :module: target.typed_vars',
         '',
         '',
-        '   .. py:attribute:: Derived.attr2',
-        '      :module: target.typed_vars',
-        '      :type: int',
-        '',
-        '',
-        '   .. py:attribute:: Derived.descr4',
+        '   .. py:attribute:: Derived.attr7',
         '      :module: target.typed_vars',
         '      :type: int',
         '',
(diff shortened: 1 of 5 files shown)


tests/test_ext_autodoc.py
```diff
@@ -1615,6 +1610,47 @@ def test_autodoc_typed_instance_variables(app):
...
+        '',
+        '',
+        '   .. py:attribute:: Derived.attr2',
+        '      :module: target.typed_vars',
+        '      :type: int',
+        '',
+        '',
+        '   .. py:attribute:: Derived.attr3',
+        '      :module: target.typed_vars',
+        '      :value: 0',
```
tk0miya, line 1639: Note: type_comment in super class is still not supported.

PR #8048, opened 2020-08-03
Don't copy graphviz.css when building man pages

cjmayo:
_static/graphviz.css is being created alongside the man pages.

-----

### Feature or Bugfix

- Bugfix

### Purpose
- Don't need the CSS file with man pages. Being copied even the source file using Graphviz was not one of those being used for the man pages
- sphinx.ext.graphviz extension

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -385,7 +385,7 @@ def man_visit_graphviz(self: ManualPageTranslator, node: graphviz) -> None:
 
 
 def on_build_finished(app: Sphinx, exc: Exception) -> None:
-    if exc is None:
+    if exc is None and app.builder.format == 'html':
         src = path.join(sphinx.package_dir, 'templates', 'graphviz', 'graphviz.css')
         dst = path.join(app.outdir, '_static')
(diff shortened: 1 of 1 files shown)


sphinx/ext/graphviz.py
```diff
@@ -385,7 +385,7 @@ def man_visit_graphviz(self: ManualPageTranslator, node: graphviz) -> None:
 
 
 def on_build_finished(app: Sphinx, exc: Exception) -> None:
-    if exc is None:
+    if exc is None and hasattr(app.builder, 'add_css_file'):
```
tk0miya: Could you check `app.builder.format == 'html'` instead here? I think it is a more natural way.

tk0miya (changes requested):
Thank you for reporting. LGTM with nits.

sphinx/ext/graphviz.py
cjmayo: Updated as suggested, and tested - works for me.

cjmayo:
Rebased on 3.x after change.

tk0miya (approved):

tk0miya:

PR #8049, opened 2020-08-04
enable type preprocessing without use_param

keewis:
As discussed in #7690, this enables type preprocessing with `napoleon_use_param = False`. I'll open a separate PR for using the type processor for other fields, e.g. `Raises`, `Returns` or `Yields`.

### DIFF
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -930,16 +930,9 @@ def _convert_numpy_type_spec(_type: str, location: str = None, translations: dic
         for token in combined_tokens
     ]
 
-    # don't use the object role if it's not necessary
-    default_translation = (
-        ":class:`%s`"
-        if not all(type_ == "obj" for _, type_ in types)
-        else "%s"
-    )
-
     converters = {
         "literal": lambda x: "``%s``" % x,
-        "obj": lambda x: convert_obj(x, translations, default_translation),
+        "obj": lambda x: convert_obj(x, translations, ":class:`%s`"),
         "control": lambda x: "*%s*" % x,
         "delimiter": lambda x: x,
         "reference": lambda x: x,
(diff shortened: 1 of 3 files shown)


sphinx/ext/napoleon/docstring.py
```diff
@@ -1067,12 +1060,11 @@ def _consume_field(self, parse_type: bool = True, prefer_type: bool = False
...
-            _type = _convert_numpy_type_spec(
-                _type,
-                location=self._get_location(),
-                translations=self._config.napoleon_type_aliases or {},
-            )
+        _type = _convert_numpy_type_spec(
+            _type,
+            location=self._get_location(),
+            translations=self._config.napoleon_type_aliases or {},
+        )
```
keewis, line 1067: my (naive) understanding of `use_param` before actually working on this was that `napoleon_use_param` was a cryptic way of saying "I want to link parameter types" (and something similar for `napoleon_use_rtype`). That doesn't seem to be true, but maybe we should allow opting out of the preprocessor? We could decide based on `napoleon_type_aliases` (don't preprocess if `napoleon_type_aliases is None`), or we can create yet another setting (`napoleon_xref_types` or something like that).

tk0miya: I don't think the additional configuration is not needed. I don't think `napoleon_use_param = False` means "not to convert types to hyperlinks".

Do you mean that it is better to add a new configuration to control the hyperlink-conversion regardless of `napoleon_use_param`?  Or it controls the conversion only if `napoleon_use_param = False`?  If you meant latter one, I feel it is not needed.

tk0miya (approved):
LGTM!

sphinx/ext/napoleon/docstring.py
keewis: the former, actually, but since `napoleon_use_param = True` will always link (are we even able to disable that?) it becomes the latter. So if you think this is fine as is, let's merge?

tk0miya: Are you saying there is who wants not to create hyperlinks? I don't still understand your concern. I don't object to add a new option. But I need to understand why it is needed. Additionally, it would be better to be orthogonal to `napoleon_use_param` configuration if possible because `napoleon_use_param` does not mean "not to create hyperlinks".

keewis: > it would be better to be orthogonal to napoleon_use_param configuration

that's what I was thinking, too, but I wouldn't know how

> Are you saying there is who wants not to create hyperlinks?

exactly, but maybe it's better to only think about that once there are bug reports / feature requests?

tk0miya: >exactly, but maybe it's better to only think about that once there are bug reports / feature requests?

Yes. I think so.

tk0miya:
Merged. Thank you always!

PR #8050, opened 2020-08-04
Preprocess other sections

keewis:
Follow-up to #7690, depends on #8049. This enables the type processor for other sections (`Returns`, `Yields` and `Raises`, possibly more) and also adds tests.

Turns out that the reason why I thought the previous PR didn't work was that I added the call to the preprocessor in the wrong place.

Related to #8004, but for `numpydoc` instead of `google`.


keewis:
I just noticed this doesn't work for `Raises` because you can't have a role in `:raises:`. Do you have any suggestions on how to deal with that, @tk0miya? To be clear, the exceptions don't have to be preprocessed, only translated to a different namespace. E.g. if we have
```rst
Raises
------
MergeError
    failed to merge
OtherError
    description
```
and
```python
napoleon_type_aliases = {
    ...
    "MergeError": "~pandas.errors.MergeError",
    "OtherError": ":py:exc:`~mypackage.OtherError",
    ...
}
```
it should become
```rst
:raises ~pandas.errors.MergeError: failed to merge
:raises ~mypackage.OtherError: description
```

keewis:
not sure if this is worth it, but this change makes sure roles are always stripped, even when they're around invalid names. That means we don't have to make the type preprocessor more complicated than it already is, with the disadvantage that it adds roles that will be stripped directly afterwards.

Edit: tests will work after merging #8049

keewis:
so now that #8049 is merged, the tests pass, too. Is there a reason why you scheduled this for 3.3.0?

tk0miya:
I don't have time for reviewing until release now. I have other topics for 3.2 release. Is this more important than others?

keewis:
not really, but it extends the type preprocessor to other fields (`Returns`, `Yields`, `Raises`), so it would fit pretty well with the other PRs related to that. Also, the diff is pretty small because the reason why it didn't work before was that I added the call to the preprocessor in the wrong place.

If you're too busy I guess it's fine to push this to 3.3.0

tk0miya:
Sorry for waiting. But I need to read the napoleon extension before reviewing because I'm a newbie of the module. So I need time for reviewing even for a minor fix.

keewis:
no worries. Also, thanks for the quick reviews on the other PRs.

keewis:
gentle ping, @tk0miya. Are there any updates on this?

keewis:
gentle ping, @tk0miya

tk0miya (approved):
LGTM!

PR #8051, opened 2020-08-04
use the obj role for all See Also items

keewis:
### Feature or Bugfix
- Bugfix

### Purpose
Before this change, `napoleon` would look for `See Also` items using the same role as the documented object. This breaks once one tries to link to other kinds of objects, e.g. from a function to a method.

With the proposed changes, the code parsing `See Also` sections will try to find the correct role for a referenced object from the intersphinx inventory.

### Relates
- #7690, #8050

### DIFF
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -1131,7 +1131,11 @@ class NumpyDocstring(GoogleDocstring):
         func_name1, func_name2, :meth:`func_name`, func_name3
 
         """
-        inventory = getattr(self._app.builder.env, "intersphinx_inventory", {})
+        try:
+            inventory = self._app.builder.env.intersphinx_inventory
+        except AttributeError:
+            inventory = {}
+
         items = []
 
         def parse_item_name(text: str) -> Tuple[str, str]:
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -1132,7 +1132,7 @@ class NumpyDocstring(GoogleDocstring):
 
         """
         try:
-            inventory = self._app.builder.env.intersphinx_inventory
+            inventory = self._app.builder.env.intersphinx_inventory  # type: ignore
         except AttributeError:
             inventory = {}
 
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -1131,6 +1131,7 @@ class NumpyDocstring(GoogleDocstring):
         func_name1, func_name2, :meth:`func_name`, func_name3
 
         """
+        inventory = getattr(self._app.builder.env, "intersphinx_inventory", {})
         items = []
 
         def parse_item_name(text: str) -> Tuple[str, str]:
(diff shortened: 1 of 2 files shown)


sphinx/ext/napoleon/docstring.py
```diff
@@ -1131,6 +1131,11 @@ def _parse_numpydoc_see_also_section(self, content: List[str]) -> List[str]:
         func_name1, func_name2, :meth:`func_name`, func_name3
 
         """
+        try:
+            inventory = self._app.builder.env.intersphinx_inventory
+        except AttributeError:
+            inventory = {}
```
keewis: is there any way to silence `mypy`? I'm aware that `BuildEnvironment` may not have a attribute named `intersphinx_inventory`, that's why I'm catching `AttributeError` (it's also easier to make the tests work that way)...

keewis: I simply added a `# type: ignore` comment on that line, but I'm not sure if that was the best way to fix this?

keewis:
the changes I just pushed will try to apply `napoleon_type_aliases` to the `See Also` section. This would allow me to manually rewrite unprefixed methods like `Index.any` to `pandas.Index.any`, which might be a solution for the namespacing issues I described in #7690. Not sure if that's the best option to fix that issue, though?

tk0miya:
>Before this change, napoleon would look for See Also items using the same role as the documented object. This breaks once one tries to link to other kinds of objects, e.g. from a function to a method.

I agree this is an incorrect approach. +1 for changing this behavior. I don't know why the original code chooses a role from `self._what`.

But it's too much to access the internal data structure of intersphinx directly. I think it is better to use `:py:obj:` always instead. Then all trouble will be resolved, right?

keewis:
that does work and is indeed simpler. However, `func` and `meth` display the link slightly different (they append `()`), probably not an issue, though?

tk0miya:
Surely. But I doubt that all items in "see also" field for methods (or functions) always refer to methods (or functions). So it would be better to use `:obj:` reference instead. Users can use `:meth:` and `:func:` roles explicitly for each item if they need to refer it as callable.

My main concern is such change is an incompatible change or not. @shimizukawa do you have an opinion?

sphinx/ext/napoleon/docstring.py
```diff
@@ -1151,6 +1151,41 @@ def push_item(name: str, rest: List[str]) -> None:
             items.append((name, list(rest), role))
             del rest[:]
 
+        def search_inventory(inventory, name, hint=None):
```
tk0miya, line 1154: This function is no longer needed now.

sphinx/ext/napoleon/docstring.py
```diff
@@ -1181,37 +1216,20 @@ def push_item(name: str, rest: List[str]) -> None:
...
         lines = []  # type: List[str]
         last_had_desc = True
         for func, desc, role in items:
             if role:
                 link = ':%s:`%s`' % (role, func)
-            elif func_role:
-                link = ':%s:`%s`' % (func_role, func)
             else:
-                link = "`%s`_" % func
+                link = ':%s:`%s`' % (func_role, func)
```
tk0miya, line 1232: It seems `func_role` would not be changed. How about embedding it into the format-string?

tk0miya (commented):
It seems @shimizukawa has been busy. Okay, let's merge this.

tk0miya:
I'll update this code after merging.

tk0miya:

PR #8053, opened 2020-08-04
FIX: Fix tuple parsing

### LINKED ISSUE #8052: BUG: tuple default argument representation
**Describe the bug**
Documenting a function with `def func(x=(1, 2))` shows up as `def func(x=1, 2)` when processed by Sphinx.

According to a `git bisect` this regression was introduced by c4d7f4d6c8f02dde6fb68ff4013caee2baee0abc

**To Reproduce**
From https://github.com/numpy/numpydoc/pull/291, steps to reproduce the behavior:
```
$ git clone https://github.com/larsoner/numpydoc
$ cd numpydoc
$ git checkout -b tuple origin/tuple
$ cd numpydoc/tests/tinybuild
$ make
<click on my_function in the built doc that pops up>
```

**Screenshots**

On commit 822625d14c0cad6607553079b31930ff9e2de717 (correct):

Screenshot from 2020-08-04 11-56-23

On commit c4d7f4d6c8f02dde6fb68ff4013caee2baee0abc (incorrect):

Screenshot from 2020-08-04 12-02-19

**Environment info**
- OS: Linunx
- Python version: 3.8
- Sphinx version: above
- Sphinx extensions: numpydoc, autodoc, autosummary

larsoner:
Closes #8052

### DIFF
--- a/sphinx/pycode/ast.py
+++ b/sphinx/pycode/ast.py
@@ -173,7 +173,7 @@ class _UnparseVisitor(ast.NodeVisitor):
 
     def visit_Tuple(self, node: ast.Tuple) -> str:
         if node.elts:
-            return ", ".join(self.visit(e) for e in node.elts)
+            return "(" + ", ".join(self.visit(e) for e in node.elts) + ")"
         else:
             return "()"
 
--- a/tests/test_domain_py.py
+++ b/tests/test_domain_py.py
@@ -313,7 +313,7 @@ def test_pyfunction_signature(app):
 
 def test_pyfunction_signature_full(app):
     text = (".. py:function:: hello(a: str, b = 1, *args: str, "
-            "c: bool = True, **kwargs: str) -> str")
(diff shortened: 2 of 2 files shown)


larsoner:
I'm happy to change the tests to adjust to the proposed changes. However, I'm not sure if it's the best approach. Seeing the failures, I'm not convinced this is the right fix. There are failures like:
```
assert 'Tuple[(str, int, ...)]' == 'Tuple[str, int, ...]'
```
so the `Tuple` does not always require the enclosing `()` even though it definitely needs it in the parameter list...

tk0miya:
Yes. We should add parenthesis only if needed. To do that, it would be better to copy an implementation from `ast.unparse()` (added since py39!).
refs: https://github.com/sphinx-doc/sphinx/issues/7498

larsoner:
Okay I'll close this since it's not the right approach

hsfzxjy:
Hi, everyone. So at this time, is there any way to correctly display a tuple in the signature?

tk0miya:
No.

PR #8060, opened 2020-08-06
Do not log 'Unparseable C++' warning if reference is of type 'any'

chrisjsewell:
Subject: Do not log 'Unparseable C++' warning if reference is of type 'any'

### Feature or Bugfix

- Bugfix

### Purpose

If `` :any:`index.md` `` is used, you now (as of sphinx 3) get the warning: "WARNING  Unparseable C++ cross-reference: 'index.md'"

It is highly unlikely that an `any` cross-reference is intended to be C++ related, and obviously breaks any builds using `sphinx-build -W`

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -7345,9 +7345,8 @@ class CPPDomain(Domain):
                 # strange, that we don't get the error now, use the original
                 return target, e
             t, ex = findWarning(e)
-            if typ != 'any':
-                logger.warning('Unparseable C++ cross-reference: %r\n%s', t, ex,
-                               location=node)
+            logger.warning('Unparseable C++ cross-reference: %r\n%s', t, ex,
+                           location=node)
             return None, None
         parentKey = node.get("cpp:parent_key", None)  # type: LookupKey
         rootSymbol = self.data['root_symbol']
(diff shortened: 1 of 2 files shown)


jakobandersen:
People may use Sphinx in all kinds of ways, so we should be careful with using likelihood as an argument here. Though I agree with your aim, but I can't reproduce the problem. https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/domains/cpp.py#L7491 should take care of exactly this issue.
However, note that there is also the `cpp:any` role, which _does_ produce warnings. If you have ``.. default-domain:: cpp`` it looks like it picks up the `cpp:any` role when one just writes `:any:`.
@chrisjsewell, can you confirm this is the case in your code?

chrisjsewell:
Ah thanks for the pointer @jakobandersen 
I see the reason now is that I'm using [pytest's caplog](https://docs.pytest.org/en/stable/logging.html#caplog-fixture) in some tests, but in `suppress_logging` you don't turn off [propagation](https://docs.python.org/3/library/logging.html#logging.Logger.propagate), so the warning is still passed to handlers on the root logger.
I have now changed this PR to temporarily turn off logging propagation in `suppress_logging`, do you see any problem with that?

jakobandersen:
Sounds reasoable. Though my knowledge of the logging module is rather limited, so I'll let @tk0miya handle it.

chrisjsewell:
Thanks, no worries if not, I was just trying to find a way to run some tests that didn't require the *full* sphinx infrastructure. But I've worked out a better way to it now 😄 
(here if you are interested: https://github.com/executablebooks/MyST-Parser/blob/40ce80972a0c340f7b888d6b92040045cd77c498/tests/test_renderers/test_myst_refs.py#L24)

Issue #8062, opened 2020-08-07
Migrate `questions` on new issues to Stackoverflow ?


yves-chevallier:
Using Google Groups ten years ago was a great idea, but I think it would be nice to redirect the users having questions about sphinx to post their questions on https://stackoverflow.com/questions/tagged/python-sphinx

Would it be a good idea ?

![image](https://user-images.githubusercontent.com/52489316/89644568-7d8a0200-d8b8-11ea-83bb-8288a80b984d.png)

tk0miya:
+0: SO is a well-known Q&A platform. So it's a nice proposal. But it would be better to post this to sphinx-users forum first. It should not be determined without community.

PR #8063, opened 2020-08-07
Invite users to ask their questions on Stackoverflow 

yves-chevallier:
Subject: Invite users to ask any question about sphinx directly on Stackoverflow

### Purpose
- The mailing list is great, but not really adapted for question/answers
- Inviting users to go to SO for asking questions about Sphinx could be a good idea
- Related to #8062

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -38,6 +38,8 @@ Bugs fixed
 ----------
 
 * #8074: napoleon: Crashes during processing C-ext module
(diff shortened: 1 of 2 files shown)


tk0miya:
@yves-chevallier why did you close this? I'd like to merge this change now.

yves-chevallier:
That's a mistake then it was closed because I did a forced-pushed somewhere... See #8130

Issue #8069, opened 2020-08-07
search experience for words word.len < 3 is inconsistent


munhitsu:
**Describe the bug**
We have document titled "1st go live". It can be found with "1st", with "live", with "1st live", but search result is empty for "go live" and "1st go live".

It's due to an inconsistency between javascript stemmer and python one.
Javascript accepts short words like "go" and looks for them in the index file. Python however explicitly removes rejects anything shorter than 3 characters.
Well, python implementation is also caveating for unicode and cardinal numbers, but that's not important.

**To Reproduce**
```
create document with "1st go live" in text
search for "1st go live" 
```

**Expected behavior**
find the page either by matching "1st" and "live" or by matching all 3 words "1st", "go", and "live".

**Your project**
Proposed fix https://github.com/uktrade/sphinx/tree/stemmer-len3

**Screenshots**
N/A

**Environment info**
- OS: Linux & Mac 10.15.6
- Python version: 3.8.5
- Sphinx version: 3.1.2 & master
- Sphinx extensions: n/a
- Extra tools: n/a

**Additional context**
N/A

PR #8070, opened 2020-08-07
Stemmer len3

munhitsu:
Subject: fix for inconsistent stemmer issue #8069

### Purpose
We have document titled "1st go live". It can be found with "1st", with "live", with "1st live", but search result is empty for "go live" and "1st go live".

It's due to an inconsistency between javascript stemmer and python one.
Javascript accepts short words like "go" and looks for them in the index file. Python however explicitly removes rejects anything shorter than 3 characters.

### Detail
N/A

### Relates
- #8069

### DIFF
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -112,7 +112,7 @@ var Stemmer = function() {
             len(word) == 0 or not (
                 ((len(word) < 3) and (12353 < ord(word[0]) < 12436)) or
                 (ord(word[0]) < 256 and (
-                    len(word) < 3 or word in self.stopwords
+                    word in self.stopwords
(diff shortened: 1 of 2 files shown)


tk0miya:
Note: The size of the index file

Default (word < 3):
```
-rw-r--r--  1 tkomiya  staff  191445  8  9 02:28 build/sphinx/html/searchindex.js
```

word < 2:
```
-rw-r--r--  1 tkomiya  staff  194238  8  9 02:30 build/sphinx/html/searchindex.js
```

word < 1
```
-rw-r--r--  1 tkomiya  staff  196574  8  9 02:33 build/sphinx/html/searchindex.js
```

PR #8073, opened 2020-08-08
Add override to add_directive

yves-chevallier:
Subject: Documentation says to override LiteralIncludeDirective, one should use the override keyword, but the example doesn't reflect that.

Also, the `add_directive` was at the root level which wasn't correct.

### Feature or Bugfix
- Bugfix

### Detail
- Just fix the docstring

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -628,7 +628,9 @@ class Sphinx:
                def run(self):
                    ...
 
-           add_directive('literalinclude', LiteralIncludeDirective)
+           def setup(app):
+               app.add_directive('literalinclude', LiteralIncludeDirective, override=True)
+               ...
 
         .. versionchanged:: 0.6
            Docutils 0.5-style directive classes are now supported.


sphinx/application.py
```diff
@@ -628,7 +628,9 @@ class LiteralIncludeDirective(Directive):
                def run(self):
                    ...
 
-           add_directive('literalinclude', LiteralIncludeDirective)
+           def setup(app):
+               app.add_directive('literalinclude', LiteralIncludeDirective, override=True)
```
tk0miya, line 632: I think this is basic usage of `add_directive()`. It would be better not to use `override` option. So I think renaming directive is better. What do you think?

yves-chevallier: Renaming a directive is always better because it is less intrusive. However sometime you have no choices to override something existing, for instance to wrap around another element. In this case the override is necessary. 

In the case of this PR, I was just confused by the documentation, I didn't know I had to add the `override=True` to override the given example with `literalinclude`. Then I get a warning and I was more confused.  Eventually I realized the documentation wasn't clear and I needed the override flag in this very example.

tk0miya: For such a purpose, we have to add a description for `override` flag. I noticed it has not been described. I think 1) adding a description, and 2) renaming the example not to need overriding are better.

tk0miya: I posted 2 PRs to fix both problems: https://github.com/sphinx-doc/sphinx/pull/8279, https://github.com/sphinx-doc/sphinx/pull/8281

tk0miya:
After a discussion, I made successor PRs of this: #8279, #8281
So I'm closing this now. Thank you for your work :-)

PR #8080, opened 2020-08-09
[Feature] Check if LaTeX package already added

yves-chevallier:
When writing an extension, the `app.add_latex_package` does not check if the same package was already added by another extension. This should be checked at some point a warning should occur. 

We could later add different options what would not emit any warning such as:

- `override` to override the already included package options
- `merge` to merge the options from the already included package


tk0miya:
LGTM!

PR #8082, opened 2020-08-09
Customize jinja comment-tag

yves-chevallier:
In LaTeX templates you could have something like this:

    \par\noindent\textbf{Exercice \thechapter.\theexercise} {#1}

where `{#` acts as a jinja comment. In this case you will get an error: 

    jinja2.exceptions.TemplateSyntaxError: Missing end of comment tag

The solution is to also override the comment tags for LaTeX.

### DIFF
--- a/sphinx/util/template.py
+++ b/sphinx/util/template.py
@@ -84,6 +84,8 @@ class LaTeXRenderer(SphinxRenderer):
         self.env.variable_end_string = '%>'
         self.env.block_start_string = '<%'
         self.env.block_end_string = '%>'
+        self.env.comment_start_string = '<#'
+        self.env.comment_end_string = '<#'
 
 
 class ReSTRenderer(SphinxRenderer):


tk0miya:
Thanks!

sphinx/util/template.py
```diff
@@ -84,6 +84,8 @@ def __init__(self, template_path: str = None, latex_engine: str = None) -> None:
         self.env.variable_end_string = '%>'
         self.env.block_start_string = '<%'
         self.env.block_end_string = '%>'
+        self.env.comment_start_string = '<#'
+        self.env.comment_end_string = '<#'
```
kai687, line 88: Is this not wrong? Should `self.env.comment_end_string` not be `#>`?

tk0miya: Oops... @yves-chevallier Please make a new PR :-)

PR #8089, opened 2020-08-09
Use alt text as short caption

yves-chevallier:
In LaTeX figures captions can be written as: 

    \caption[Short description]{Long description}

Having long descriptions for figures will compromise the `\listoffigures`. It is therefore useful to have short description for figures, but Sphinx does not support this out of the box. 

One solution is to use the `alt` text used in HTML `text` for the LaTeX short description. 

There are too many `if` in my patch. Feel free to propose something better.

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1330,7 +1330,7 @@ class LaTeXTranslator(SphinxTranslator):
     def depart_figure(self, node: Element) -> None:
         self.body.append(self.context.pop())
 
-    def visit_caption(self, node: Element) -> None:
+    def visit_caption(self, node):
         self.in_caption += 1
         if isinstance(node.parent, captioned_literal_block):
             self.body.append('\\sphinxSetupCaptionForVerbatim{')
@@ -1339,7 +1339,14 @@ class LaTeXTranslator(SphinxTranslator):
         elif self.table and node.parent.tagname == 'figure':
             self.body.append('\\sphinxfigcaption{')
         else:
-            self.body.append('\\caption{')
+            # Use alt text as short caption for the \listoffigures
+            short = ''
+            if isinstance(node.parent, nodes.figure):
+                if isinstance(node.parent.children[0], nodes.image):
+                    alt = node.parent.children[0].get('alt')
+                    if alt:
+                        short = '[' + texescape.escape(alt) + ']'
+            self.body.append('\\caption%s{' % short)
 
     def depart_caption(self, node: Element) -> None:
         self.body.append('}')
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1338,8 +1338,8 @@ class LaTeXTranslator(SphinxTranslator):
             self.body.append(r'\captionof{figure}{')
         elif self.table and node.parent.tagname == 'figure':
             self.body.append(r'\sphinxfigcaption{')
-        elif (isinstance(node.parent, node.figure) and
-              isinstance(node.parent.children[0], node.image) and
+        elif (isinstance(node.parent, nodes.figure) and
+              isinstance(node.parent.children[0], nodes.image) and
               'alt' in node.parent.children[0]):
             alt = node.parent.children[0]['alt']
             self.body.append(r'\caption[%s]{' % texescape.escape(alt))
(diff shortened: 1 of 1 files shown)


sphinx/writers/latex.py
```diff
@@ -1330,7 +1330,7 @@ def visit_figure(self, node: Element) -> None:
     def depart_figure(self, node: Element) -> None:
         self.body.append(self.context.pop())
 
-    def visit_caption(self, node: Element) -> None:
+    def visit_caption(self, node):
```
tk0miya: Why did you remove annotations?

sphinx/writers/latex.py
```diff
@@ -1339,7 +1339,14 @@ def visit_caption(self, node: Element) -> None:
...
             self.body.append('\\sphinxfigcaption{')
         else:
-            self.body.append('\\caption{')
+            # Use alt text as short caption for the \listoffigures
+            short = ''
+            if isinstance(node.parent, nodes.figure):
+                if isinstance(node.parent.children[0], nodes.image):
+                    alt = node.parent.children[0].get('alt')
+                    if alt:
+                        short = '[' + texescape.escape(alt) + ']'
```
tk0miya: How about this? I think this condition should be flatten as others above.
```suggestion
    elif (isinstance(node.parent, node.figure) and isinstance(node.parent.children[0], node.image) and 'alt' in node.parent.children[0]):
        alt = node.parent.children[0]['alt']
        self.body.append(r'\caption[%s]{' % texescape.escape(alt))
    else:
```

tk0miya (changes requested):
+1 for your idea.

sphinx/writers/latex.py
yves-chevallier: Bad bad, I as testing it at first...

tk0miya:
Thank you for update. LGTM.
I'll merge this after you'll add a testcase for this :-)

mgeier:
I don't think it's a good idea to ab-use the `alt` text as "short description" for the list of figures.

The `alt` text has a quite specific intention, mostly for accessibility, see e.g. https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt.

I think if a "short description" is needed, a separate field should be added.

yves-chevallier:
@mgeier I partially agree with you. `alt` is not a short description, but as you said a text that is visible for accessibility (when the figure is missing or when the user has some trouble to see the image. 

That said an image that shows the Addams Family with a caption such as:

> The Addams family, from left to right Gomez, Morticia, Wednesday and the Thing on her shoulder.

A good `alt` would be simply 

> The Addams Family

and it would also perfectly fit the purpose the short description...

However, where you are right is if the sphinx user is working with blind people and write the following `alt` for this HTML output, it won't work as short description...

> This is a picture (aspect ratio 3:2) describing the Addams Family in front of their castle at night with some zombies on the backend...

Your suggestion of adding a `short-description` option is good, but everything is currently done in docutils.

yves-chevallier:
@mgeier I thought more about what you said and I think you're right. If somebody truly use `alt` text for its primarily purpose in HTML, he doesn't want his LaTeX code to be alt'ered with this patch. 

So yes, adding a `short-caption` attribute may be the right option. Then comes the next question : how to do it properly? `figure` is a Docutils node and adding this kind of attribute seems difficult. Any clues would be appreciated.

tk0miya:
>So yes, adding a short-caption attribute may be the right option. Then comes the next question : how to do it properly? figure is a Docutils node and adding this kind of attribute seems difficult. Any clues would be appreciated.

The proper way is adding a new option to the figure directive. So it would be better to propose this to docutils project. But I don't know that will be accepted. I think reST is a lightweight mark-up language. So it is not a goal to represent all kind of documentations, I believe.

yves-chevallier:
How difficult is it to enhance the `figure` directive on sphinx side only?

tk0miya:
No so difficult. But I object to do that for two reasons.

* It makes compatibility problem with docutils. It causes confuses not good to add a directive that is similar to common directive, but different. 
* As I commented above, reST's goal is not representing all kind of document. It's a lightweight mark-up language.

yves-chevallier:
So I agree with you... What do we do with this PR, should I close it or do we go with the `alt` as *short-caption*?

tk0miya:
As mgeier said, `:alt:` is not good for short description. So we should close this, sadly.

Issue #8093, opened 2020-08-10
Highlight warn not the current node location 


yves-chevallier:
In `writer/html5.py/visit_literal_block` we find this:

        highlighted = self.highlighter.highlight_block(
            node.rawsource, lang, opts=opts, linenos=linenos,
            location=(self.builder.current_docname, node.line), **highlight_args
        )

The good news is that it works in most cases. However if a node is moved during the build process, you'll get an error with the wrong document name as we use the `node.line`, but the current docname.

Wouldn't be better to use `node.source` and `node.line`?

tk0miya:
I think `current_docname` is intentionally used here. Could you check `node.source` is working fine on dirhtml or singlehtml builders?

yves-chevallier:
In either case (`html`, `dirhtml`, `singlehtml`, `latex`), `node.source` equals the absolute path of the file : 

    '/home/me/test-sandbox/other.rst'

The value of `self.builder.current_docname` is less verbose, it is 

    'other'

But the output is in either case the abs path and it work with all builders.

tk0miya:
Ah, my last comment was wrong. Absolutely you're right. We have to use `location=node` instead of `current_docname`. It uses `node.source` internally.

PR #8102, opened 2020-08-11
Use current node location in case of error (related to #8093)

yves-chevallier:
In `writer/html5.py/visit_literal_block` we find this:

        highlighted = self.highlighter.highlight_block(
            node.rawsource, lang, opts=opts, linenos=linenos,
            location=(self.builder.current_docname, node.line), **highlight_args
        )

The good news is that it works in most cases. However if a node is moved during the build process, you'll get an error with the wrong document name as we use the `node.line`, but the current docname.

Wouldn't be better to use `node.source` and `node.line`?

### DIFF
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -450,7 +450,7 @@ class HTMLTranslator(SphinxTranslator, BaseTranslator):
 
         highlighted = self.highlighter.highlight_block(
             node.rawsource, lang, opts=opts, linenos=linenos,
-            location=(self.builder.current_docname, node.line), **highlight_args
+            location=node, **highlight_args
         )
         starttag = self.starttag(node, 'div', suffix='',
                                  CLASS='highlight-%s notranslate' % lang)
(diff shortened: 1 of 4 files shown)


sphinx/writers/html.py
```diff
@@ -450,7 +450,7 @@ def visit_literal_block(self, node: Element) -> None:
 
         highlighted = self.highlighter.highlight_block(
             node.rawsource, lang, opts=opts, linenos=linenos,
-            location=(self.builder.current_docname, node.line), **highlight_args
+            location=(node.source, node.line), **highlight_args
```
tk0miya: As I commented at https://github.com/sphinx-doc/sphinx/issues/8093#issuecomment-672088541, I think using `location=node` is better. It tries to get a source and line in the best effort (ex. it tries to trace parent node if `node.source` and `node.line` is not set). No reason to use `node.source` and `node.line` directly.

tk0miya (changes requested):

tk0miya:
It seems some errors is caused by this change. Could you check them please?

sphinx/writers/html.py
yves-chevallier: I'll give a try

tk0miya (approved):

tk0miya:
Thanks!

PR #8104, opened 2020-08-12
Make linkcheck more robust on HEAD failure

dopplershift:
Subject: Make linkcheck more robust on HEAD failure

### Feature or Bugfix

- Bugfix (sorta)

### Purpose
- Fix linkcheck errors connecting to this [website](https://www.ncdc.noaa.gov/data-access/model-data/model-datasets/north-american-regional-reanalysis-narr). It completely drops the connection when trying to do a HEAD request (due to the default set of headers from requests). The simplest solution was to catch `ConnectionError` as well as `HTTPError`.

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -20,7 +20,7 @@ from urllib.parse import unquote, urlparse
 
 from docutils import nodes
 from docutils.nodes import Node
-from requests.exceptions import HTTPError
+from requests.exceptions import HTTPError, ConnectionError
 
 from sphinx.application import Sphinx
 from sphinx.builders import Builder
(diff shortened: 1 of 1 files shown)


dopplershift:
So I'm not sure if catching `ConnectionError` is a great idea, but it did solve my problem.

It also turns out that using `linkcheck_request_headers` can also solve my problem, so no worries if you don't want this change.

tk0miya:
+0: I don't think this does not harm anybody. I'm okay to merge this if you add a comment to describe why ConnectionError is handled.

dopplershift:
I think I'm going to retract this, since my problem is solved with the existing config hooks. I personally like that in general the `ConnectionError` indicates a broken link/server and the linkchecker doesn't proceed to try a full get request. The server I was having issues with is just broken for its behavior.

PR #8118, opened 2020-08-13
Make man writer default month format the same as documented

cjmayo:
Subject:

Current default is the abbreviated month name, change to full month name as today_fmt in documentation

### Feature or Bugfix
- Bugfix

### Detail
Documented as today_fmt in doc/usage/configuration.rst.

This also matches the format to be used for the last modification date
in the BSD mdoc guide:
https://manpages.bsd.lv/mdoc.html

### DIFF
--- a/sphinx/writers/manpage.py
+++ b/sphinx/writers/manpage.py
@@ -109,7 +109,7 @@ class ManualPageTranslator(SphinxTranslator, BaseTranslator):
         if self.config.today:
             self._docinfo['date'] = self.config.today
         else:
-            self._docinfo['date'] = format_date(self.config.today_fmt or _('%b %d, %Y'),
+            self._docinfo['date'] = format_date(self.config.today_fmt or _('%B %d, %Y'),
                                                 language=self.config.language)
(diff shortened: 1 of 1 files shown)


tk0miya:
It seems our document is wrong. Sphinx has not used `'%B %d, %Y'` at all. So we have to fix our document instead...
```
$ grep -r '%B %d, %Y' sphinx
$ grep -r '%b %d, %Y' sphinx | grep -v sphinx/locale
sphinx/writers/texinfo.py:                                format_date(self.config.today_fmt or _('%b %d, %Y'),
sphinx/writers/manpage.py:            self._docinfo['date'] = format_date(self.config.today_fmt or _('%b %d, %Y'),
sphinx/builders/latex/__init__.py:            self.context['date'] = format_date(self.config.today_fmt or _('%b %d, %Y'),
sphinx/builders/html/__init__.py:            self.last_updated = format_date(lufmt or _('%b %d, %Y'),
sphinx/transforms/__init__.py:                    text = format_date(self.config.today_fmt or _('%b %d, %Y'),
```

Thank you for letting us know.

PR #8121, opened 2020-08-14
Add permalink to rubric

yves-chevallier:
According to docutils a rubric is: 

> rubric n. 1. a title, heading, or the like, in a manuscript, book, statute, etc., written or printed in red or otherwise distinguished from the rest of the text. ...

I think it deserves to have a permalink like other titles or headings. A rubric with a headlink can be added as follow: 

```python
class TestRubric(SphinxDirective):
    def run(self):
        node = nodes.section('',
            nodes.rubric('Rubric Title', 'Rubric Title'),
            nodes.Text('Lorem ipsum.'),
            ids=['rubric-permalink-name']
        )
        return [node]

def setup(app):
    app.add_directive('test_rubric', TestRubric)
```

### DIFF
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -430,6 +430,11 @@ class HTMLTranslator(SphinxTranslator, BaseTranslator):
         super().depart_title(node)
 
     # overwritten
+    def depart_rubric(self, node: Element) -> None:
+        self.body.append('</p>\n')
+        self.add_permalink_ref(node.parent, _('Permalink to this headline'))
+
+    # overwritten
     def visit_literal_block(self, node: Element) -> None:
         if node.rawsource != node.astext():
             # most probably a parsed-literal block -- don't highlight
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -256,6 +256,7 @@ h6:hover > a.headerlink,
 dt:hover > a.headerlink,
 caption:hover > a.headerlink,
 p.caption:hover > a.headerlink,
+p.rubric:hover > a.headerlink,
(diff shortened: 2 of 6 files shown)


sphinx/writers/html.py
```diff
@@ -429,6 +429,11 @@ def depart_title(self, node: Element) -> None:
 
         super().depart_title(node)
 
+    # overwritten
+    def depart_rubric(self, node: Element) -> None:
+        self.body.append('</p>\n')
+        self.add_permalink_ref(node.parent, _('Permalink to this headline'))
```
tk0miya, line 435: What is a parent node of the rubric node? I think this is a wrong implementation.

tk0miya (changes requested):

tk0miya:
Could you add a testcase, please?

sphinx/writers/html.py
yves-chevallier: Usually a parent node for `rubric` is a section, or anything with an `ids`.

tk0miya: That's strange. The section and rubric node is not related at all. I guess the anchor generated `add_permalink_ref` will link to the section instead of rubric itself. It's strange.

tk0miya: Ah, I understand why you specified the section node here. It seems your extension creates a rubric node just under the section node. But it is prohibited. The section node should start with a title node.

>section elements begin with a title, and may contain body elements as well as transition, topic, and sidebar elements.
>https://docutils.sourceforge.io/docs/ref/doctree.html#section

tk0miya:
I think this PR was posted from a misunderstanding of the doctree structure. So I'm closing this.
Thank you for your contribution :-)

PR #8122, opened 2020-08-14
Glob names for output assets directories

yves-chevallier:
Currently in the builders, the values of `_images` and `_sources` hard kind of hard-coded. Only the `_images` is initialized on the `__init__` which makes it cumbersome to modify this value in a new builder: 

```python
class MyNewBuilder(StandaloneHTMLBuilder):
    name = 'mybuilder'
    def __init__(*args, **kwargs):
        self.imagedir = 'assets/images' # override
```

With the change made, it is easier:

```python
class MyNewBuilder(StandaloneHTMLBuilder):
    name = 'mybuilder'
    imagedir = 'assets/images'
```

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -84,6 +84,10 @@ class Builder:
     #: The builder supports data URIs or not.
     supported_data_uri_images = False
 
+    # basename of output assets directories
+    imagedir = ""
+    sourcesdir = ""
+
     def __init__(self, app: "Sphinx") -> None:
         self.srcdir = app.srcdir
         self.confdir = app.confdir
@@ -103,8 +107,7 @@ class Builder:
 
         # images that need to be copied over (source -> dest)
         self.images = {}  # type: Dict[str, str]
-        # basename of images directory
-        self.imagedir = ""
+
         # relative path to image directory from current docname (used at writing docs)
         self.imgpath = ""
 
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -85,7 +85,7 @@ class Builder:
     supported_data_uri_images = False
 
     # basename of output assets directories
-    imagedir = ""
+    assets_images_directory = ""
     sourcesdir = ""
 
     def __init__(self, app: "Sphinx") -> None:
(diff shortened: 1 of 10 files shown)


sphinx/builders/__init__.py
```diff
@@ -84,6 +84,10 @@ class Builder:
     #: The builder supports data URIs or not.
     supported_data_uri_images = False
 
+    # basename of output assets directories
+    imagedir = ""
+    sourcesdir = ""
```
tk0miya, line 89: Is `sourcedir` a common attribute of builders? I don't think so.

sphinx/builders/html/__init__.py
```diff
@@ -1032,7 +1036,7 @@ def hasdoc(name: str) -> bool:
             logger.warning(__("error writing file %s: %s"), outfilename, err)
         if self.copysource and ctx.get('sourcename'):
             # copy the source file for the "show source" link
-            source_name = path.join(self.outdir, '_sources',
+            source_name = path.join(self.outdir, self.sourcedir,
```
tk0miya: Please give a more descriptive name to this. I suppose this name will confuse developers (including me!). This name is very similar to `self.srcdir`. It would be better not to make a variable if you can't give a better name.

tk0miya (changes requested):

sphinx/builders/__init__.py
yves-chevallier: Only used in html though...

sphinx/builders/html/__init__.py
yves-chevallier: I agree the name is confusing. I'll try to find another one. 

About having or not a variable for a single use, if you wanna be SOLID, yes you definitely need to be *open for extension and closed for modification*.

tk0miya: I found some usage of `_sources` string in `themes/agogo/layout.html` and `themes/basic/sourcelink.html`. Please consider them also.

tk0miya: I don't object to change it to a variable only if we can give a good name for it. I think extendability is based on good names. So we have to consider the better name to make it extendable.

sphinx/builders/__init__.py
tk0miya: If so, no reason to add it to the common Builder class.

yves-chevallier:
@tk0miya Not really proud of myself. This is a lot of changes for something quite small. I still think users can choose their folder structure, and at least the name for the `_sources` directory. Ultimately I would like to regroup every folders in a `assets` section: 

- templates
- static directory
- css
- extensions (we can have a default folder for the extensions with no need to add the extension folder to the path)
- output/images
- output/sources
- ...

PR #8126, opened 2020-08-14
Move tests on Travis to GitHub Actions

harupy:

### Feature or Bugfix

- Enhancement

### Purpose

Move tests on Travis to GitHub Actions

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>


harupy:
For some reason, a few tests fail in python 3.5:
https://github.com/sphinx-doc/sphinx/pull/8126/checks?check_run_id=985939238#step:7:992

.github/workflows/main.yml
```diff
@@ -1,21 +1,85 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v2
+        if: matrix.python != '3.10-dev'
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Set up Python ${{ matrix.python }} (deadsnakes)
+        uses: deadsnakes/action@v1.0.0
+        if: matrix.python == '3.10-dev'
```
tk0miya: Could you use `endsWith` function like README of deadsnakes/action does? I'll add `3.9-dev` later.

harupy:
The error diff indicates ` func(arg: str, kwarg=None)` and `func(arg: int, kwarg=None)` are swapped?

https://github.com/sphinx-doc/sphinx/pull/8126/checks?check_run_id=987281907#step:8:1598

```diff
E         Full diff:
E           [
E            '',
E            '.. py:module:: target.singledispatch',
E            '',
E            '',
E            '.. py:function:: func(arg, kwarg=None)',
E         +  '                 func(arg: str, kwarg=None)',
E            '                 func(arg: int, kwarg=None)',
E         -  '                 func(arg: str, kwarg=None)',
E            '   :module: target.singledispatch',
E            '',
E            '   A function for general use.',
E            '',
E           ]
```

.github/workflows/main.yml
```diff
@@ -1,21 +1,85 @@
...
+        env:
+          TOXENV: ${{ matrix.tox_env }}
+        run: tox -- -vv
+
+      - name: codecov
+        if: matrix.use_coverage
+        env:
+          TOXENV: ${{ matrix.tox_env }}
+          PYTEST_ADDOPTS: "--cov ./ --cov-append --cov-config setup.cfg"
+        run: codecov -e $TOXENV
```
tk0miya: Is it better to pass `matrix.tox_env` directly instead of envvars? It can lessen a line of YAML.

.github/workflows/main.yml
```diff
@@ -1,21 +1,85 @@
...
+      - name: Install graphviz
+        run: sudo apt-get install graphviz
+
+      - name: Install dependencies
+        run: pip install -U tox codecov
+
+      - name: Run tox
+        env:
+          TOXENV: ${{ matrix.tox_env }}
+        run: tox -- -vv
```
tk0miya: We can pass the TOXENV via `-e` option instead of evnvars. Is it better?

.github/workflows/main.yml
```diff
@@ -1,21 +1,85 @@
...
+            tox_env: "du16"
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v2
+        if: matrix.python != '3.10-dev'
+        with:
+          python-version: ${{ matrix.python }}
+
```
tk0miya: We have to choose the style of YAML; sparse or dense. The old YAML (for windows) contains a few blank lines. But this PR adds many blank lines to make YAML sparse (maybe for readability). To keep the file consistent, it is not good to mix styles on the single file. Please let me know your opinion. Personally, I prefer a dense style a bit.

tk0miya (changes requested):
I think this change makes CI on Travis unneeded. So could you remove `.travis.yml` please?

.github/workflows/main.yml
harupy: makes sense

tk0miya:
I'll investigate the error of py35 later. It is not related to this change.

.github/workflows/main.yml
harupy: I don't have any preference here. Whatever works for you is fine for me. I'll change the style to dense.

.github/workflows/main.yml
harupy: That's better. I'll fix.

.github/workflows/main.yml
```diff
@@ -1,21 +1,71 @@
...
-    - uses: actions/checkout@v2
-    - name: Set up Python
-      uses: actions/setup-python@v1
-      with:
-        architecture: ${{ matrix.architecture }}
-    - name: Install dependencies
-      run:  pip install -U tox
-    - name: Run Tox
-      run:  tox -e py -- -vv
+      - uses: actions/checkout@v2
```
harupy, line 63: @tk0miya Which style would you prefer for an array?

```yml
# 1. no indent
steps:
- ...

# 2. indent
steps:
  - ...
```

tk0miya: No opinion. Either is good to me.

harupy: ok I'll use the current style.

harupy:
@tk0miya 
> I think this change makes CI on Travis unneeded. So could you remove .travis.yml please?

`.travis.yml` contains a test using `node` and `xvfg`. Can I move this in another PR because I'm still trying to figure out how to setup `xvfg` in GitHub Actions?

tk0miya:
Now I merged this manually into our repository. It must be based on your effort. Thank you for your work!

PR #8137, opened 2020-08-19
Closes #8123: Fix plus-handling (+) in search terms for basic html th…

### LINKED ISSUE #8123: Searching for the term `C++` is broken
**Describe the bug**

Instead of searching for `C++` sphinx searches for `C  ` (two spaces after the C)

I've done some investigation on the matter:

https://github.com/sphinx-doc/sphinx/blob/af62fa61e6cbd88d0798963211e73e5ba0d55e6d/sphinx/themes/basic/static/doctools.js#L54 

These two lines
```
    var key = jQuery.urldecode(tmp[0]);
    var value = jQuery.urldecode(tmp[1]);
```

do too much decoding. Instead I suppose that https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent should be used.

**To Reproduce**

Go to https://www.sphinx-doc.org/en/master/search.html?q=C%2B%2B

On the sphinx search page In your browser's Javascript Console, type
```
jQuery.urldecode('C%2B%2B') // e.g. a user wants to search for the language `C++`, leading for the URI to be search.html?q=C%2B%2B
```
-> `C  `
so sphinx will actually search for `C  ` instead of `C++`.

**Expected behavior**

I would like sphinx to search for `C++`

hofmandl1:
Subject: Fix searching for terms containing +

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* #8123: html-search: fix searching for terms containing + (Requires a
+  custom search language that does not split on +)
+
 Testing
 --------
 
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -18,6 +18,7 @@ module.exports = function(config) {
       'sphinx/themes/basic/static/underscore.js',
       'sphinx/themes/basic/static/jquery.js',
       'sphinx/themes/basic/static/doctools.js',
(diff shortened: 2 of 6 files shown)


tk0miya:
Sorry for not responding long. Thank you for your contribution!

hofmandl1:
:D

PR #8140, opened 2020-08-20
📚 DOCS: Improve build process overview

chrisjsewell:

### Feature or Bugfix

- Bugfix

### Purpose

Should be self-explanatory.

This is what I understood from `Builder.build`, `BuildEnvironment.check_dependents` and `Builder.write`

### DIFF
--- a/doc/extdev/appapi.rst
+++ b/doc/extdev/appapi.rst
@@ -179,8 +179,10 @@ type for that event::
    11. event.env-get-updated(app, env)
    11. event.env-check-consistency(app, env)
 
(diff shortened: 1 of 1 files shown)


chrisjsewell:
Also, could you clarify the difference between `env-updated` and `env-get-updated`?
The latter is not documented, but it seems they essentially do the same thing?

tk0miya:
>Also, could you clarify the difference between env-updated and env-get-updated?
>The latter is not documented, but it seems they essentially do the same thing?

I can't remember why I added the latter one. It seems very similar and almost same thing...

tk0miya:
Thank you for update!

PR #8142, opened 2020-08-21
autodoc: fix constructor signatures for classes derived from typing.Generic

jcarrano:
Subject: take the constructor signature from __init__ instead of __new__ for classes derived from typing.Generic

### Feature or Bugfix

- Bugfix

### Purpose

When documenting classes derived from `typing.Generic` (essentially all classes in the typing module) the constructor signature would show an unhelpful `(*args, **kwds).`

`typing.Generic.__new__` was being picked up by sphinx.ext.autodoc. With this patch it is skipped and constructor signatures for generic classes are taken from `__init__`.

### Detail

I copied the code that was used to skip callable metaclasses.

### DIFF
--- a/tests/test_ext_autodoc.py
+++ b/tests/test_ext_autodoc.py
@@ -290,6 +290,22 @@ def test_format_signature(app):
         '(b, c=42, *d, **e)'
 
 
+@pytest.mark.skipif(sys.version_info < (3, 5), reason='typing is available since python3.5.')
+@pytest.mark.xfail
+@pytest.mark.sphinx('html', testroot='ext-autodoc')
+def test_autodoc_process_signature_typing_generic(app):
+    actual = do_autodoc(app, 'class', 'target.generic_class.A', {})
+
+    assert list(actual) == [
+        '',
+        '.. py:class:: A(a, b=None)',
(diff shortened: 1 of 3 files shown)


tests/test_ext_autodoc.py
```diff
@@ -290,6 +290,35 @@ def foo3(self, d='\n'):
         '(b, c=42, *d, **e)'
 
 
+@pytest.mark.skipif(sys.version_info < (3, 5), reason='typing is available since python3.5.')
+def test_autodoc_process_signature_typing_generic(app):
```
tk0miya, line 295: Could you rewrite testcase using `do_autodoc()`?

* Add a target class to `tests/roots/test-ext-autodoc/targets/`
* Call `do_autodoc()`. There are many examples in this file.

tk0miya (changes requested):
LGTM with nits.

tests/test_ext_autodoc.py
jcarrano: Done. I added a new file because I did not feel it fit in any of the existing test files.

tk0miya (approved):

tk0miya:
Merged. Thank you for your support!

PR #8144, opened 2020-08-21
Fix #8143: AttributeError if autodoc_default_options contains False

### LINKED ISSUE #8143: AttributeError: 'bool' object has no attribute 'split' when arg is False in sphinx.ext.autodoc:members_option() in graphql-core
**Describe the bug**
When I attempt to build the documentation of [graphql-python/graphql-core](https://github.com/graphql-python/graphql-core/blob/master/tox.ini) (commit 9d9b465, which is the most recent one in master today) I get the following error ith Sphinx **3.2.1**:

```
Exception occurred:
  File ".../graphql-core/.tox/docs/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 98, in members_option
    return [x.strip() for x in arg.split(',') if x.strip()]
AttributeError: 'bool' object has no attribute 'split'
```

**To Reproduce**
```
$ git clone https://github.com/graphql-python/graphql-core
$ cd graphql-core
$ git switch -d 9d9b465
$ sed -i 's/sphinx>=2.4,<3/sphinx>=3.2/' tox.ini
$ tox -e docs
...
Running Sphinx v3.2.1
building [html]: targets for 23 source files that are out of date
updating environment: [new config] 23 added, 0 changed, 0 removed
reading sources... [  4%] diffs
reading sources... [  8%] index
reading sources... [ 13%] intro
reading sources... [ 17%] modules/error

Exception occurred:
  File ".../graphql-core/.tox/docs/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 97, in members_option
    return [x.strip() for x in arg.split(',') if x.strip()]
AttributeError: 'bool' object has no attribute 'split'
The full traceback has been saved in /tmp/sphinx-err-wxxmynxh.log, if you want to report the issue to the developers.

```

**Expected behavior**
This used to work before (with Sphinx 3.1) and now it no longer works. I don't know where is the `False` value coming from.
I expect the documentation to be generated.

**Your project**
https://github.com/graphql-python/graphql-core

**Environment info**
- OS: Linux (Fedora)
- Python version: 3.8.5, 3.9.0rc1
- Sphinx version: 3.2.1

**Additional context**
We've updated Sphinx in Fedora from 3.1.2 to 3.2.1 and this is the only package that fails to build with the new version.

Full traceback attached: [log.txt](https://github.com/sphinx-doc/sphinx/files/5109491/log.txt)

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #8143 
- autodoc crahses when autodoc_default_options contains False value
unexpectedly.  After this change, it also accepts False.

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -94,7 +94,10 @@ def members_option(arg: Any) -> Union[object, List[str]]:
     """Used to convert the :members: option to auto directives."""
     if arg is None or arg is True:
         return ALL
-    return [x.strip() for x in arg.split(',') if x.strip()]
+    elif arg is False:
+        return None
+    else:
+        return [x.strip() for x in arg.split(',') if x.strip()]
 
 
 def members_set_option(arg: Any) -> Union[object, Set[str]]:
@@ -172,7 +175,7 @@ def merge_members_option(options: Dict) -> None:
 
     members = options.setdefault('members', [])
     for key in {'private-members', 'special-members'}:
-        if key in options and options[key] is not ALL:
+        if key in options and options[key] not in (ALL, None):
             for member in options[key]:
                 if member not in members:
                     members.append(member)
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,8 @@ Bugs fixed
 ----------
 
 * #8085: i18n: Add support for having single text domain
+* #8143: autodoc: AttributeError is raised when False value is passed to
+  autodoc_default_options
 * #8093: The highlight warning has wrong location in some builders (LaTeX,
   singlehtml and so on)
 


sphinx/ext/autodoc/__init__.py
```diff
@@ -94,7 +94,10 @@ def members_option(arg: Any) -> Union[object, List[str]]:
     """Used to convert the :members: option to auto directives."""
     if arg is None or arg is True:
         return ALL
-    return [x.strip() for x in arg.split(',') if x.strip()]
+    elif arg is False:
+        return None
```
hroncok, line 98: I would expect `EMPTY` here. WDYT?

tk0miya: No, I don't think so. `None` means the option is not set. And it skips all private-members automatically (as auto* directive does by default). At present, all `:members:`, `:inherited-members:` and `:private-members:` don't become `EMPTY`. I guess it will break autodoc unexpectedly.

hroncok: Thanks for the explanation.

PR #8147, opened 2020-08-22
Fixes #8146: When identifying bases, only use classes from builtins

### LINKED ISSUE #8146: [inheritance_diagram] Use class objects only from builtins while evaluating bases
In `sphinx.ext.inheritance_diagram` extension, we iterate over all bases and verify if a class object belongs to one of the Python built-in classes. As of now, we simply check for its presence in all `builtins` objects. This may be simple but we are ignoring the fact that besides classes, `builtins` also contains functions and other built-in Python objects. 

The above is a direct cause of the following bug:

**Prerequisite code:**
```
import builtins
import inspect
class classproperty(object):
    def __init__(self, f):
        self.f = f
    def __get__(self, obj, owner):
        return self.f(owner)

class NamedAction(object):
    @classproperty
    def name(self):
        raise NotImplementedError()
```

**As of now:**
```
py_builtins = vars(builtins).values()    # As used in sphinx.ext.inheritance_diagram
cls = NamedAction                        # defined above
cls in py_builtins
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-6-2e6b4af0d81c> in <module>
----> 1 cls in py_builtins

/usr/lib/python3.8/importlib/_bootstrap.py in __eq__(self, other)

<ipython-input-3-a8bb3f065bd0> in __get__(self, obj, owner)
      3         self.f = f
      4     def __get__(self, obj, owner):
----> 5         return self.f(owner)
      6 

<ipython-input-4-b0467820c849> in name(self)
      2     @classproperty
      3     def name(self):
----> 4         raise NotImplementedError()
      5 

NotImplementedError: 
```

**After the proposed fix:**
```
py_builtins = [obj for obj in vars(builtins).values()
               if inspect.isclass(obj)]    # proposed fix
cls = NamedAction                          # defined above
cls in py_builtins
False
```

The solution here is pretty simple. Instead of relying on the entire `builtins`, let's separate out the classes via `inspect.isclass`

N/A

**Additional context**
N/A

return007:
Subject: When identifying bases, only use class objects from `builtins`

### Feature or Bugfix
- Feature
- Minor refactoring

### Purpose
In inheritance_diagram extension, while iterating over bases,
we verify if the base class is one of the Python built-in
class or not.
As of now, we simply check for its presence in all `builtins`
objects. Please note, `builtins` not only has built-in classes,
but also functions (like `open`) and other built-in objects.

To avoid any sort of future problem, it seems better to only
use classes (and of course exception classes).

### Relates
- Issue #8146

### DIFF
--- a/sphinx/ext/inheritance_diagram.py
+++ b/sphinx/ext/inheritance_diagram.py
@@ -66,6 +66,10 @@ module_sig_re = re.compile(r'''^(?:([\w.]*)\.)?  # module names
                            ''', re.VERBOSE)
 
 
+py_builtins = [obj for obj in vars(builtins).values()
+               if inspect.isclass(obj)]
+
+
 def try_import(objname: str) -> Any:
     """Import a object or module using *name* and *currentmodule*.
     *name* should be a relative name from *currentmodule* or
@@ -178,7 +182,6 @@ class InheritanceGraph:
         traverse to. Multiple names can be specified separated by comma.
         """
         all_classes = {}
-        py_builtins = vars(builtins).values()
 
         def recurse(cls: Any) -> None:
             if not show_builtins and cls in py_builtins:


tk0miya:

PR #8149, opened 2020-08-23
Fix comment end string in LaTeXRenderer

kai687:
Fix bug introduced by commit 5f82825e27 in `self.env.comment_end_string`

### Feature or Bugfix

- Bugfix

### Detail
- Changed the comment end string to `#>`


tk0miya:
Thanks!

PR #8153, opened 2020-08-25
Fix typos

domdfcoding:
Subject: Fix two typos in the code. 

### Feature or Bugfix
- Bugfix

### Detail
- One type is in the warning emitted by Sphinx.add_lexer() when passing an instance rather than a class. There was a missing 'of'; without it the sentence doesn't make sense. 
- The other is a comment that should say `# already loaded` but actually says `# alread loaded`.

There doesn't appear to be a test for the warning, so I don't think any changes will be required elsewhere.


tk0miya:
Thanks!

PR #8158, opened 2020-08-28
Add option "include-subclasses" to inheritance diagram 

doerwalter:
This patch adds the following new feature to sphinx: `inheritance-diagram` gains a new option `:include-subclasses:`.

When this option is given, not only the classes listed as options to `inheritance-diagram` are included in the diagram, but all sub classes (and their sub classes etc.) as well.

This makes it possible to generate a complete inheritance tree by just listing the base class and adding `:include-subclasses:`.

This is different from specifying the module that contains this base class, as this module might include other classes that should not be part of the diagram.


doerwalter:
Strange, why does it include all the other commits? I'll try to redo the pull request from scratch.

PR #8160, opened 2020-08-28
domains: c: fix intersphinx anonymous nested names

utzig:
When anonymous enums (and possibly other anonymous elements) exists in a project, the current intersphinx inventory looks something like:

```
$ python3 -m sphinx.ext.intersphinx objects.inv | rg SOME_SYMBOL
[anonymous].SOME_SYMBOL some-long-unique-url
```

When linking external projects through intersphinx one has to write:

```
:c:enumerator:`SOME_SYMBOL <[anonymous].SOME_SYMBOL>`
```

After this commit, the generated intersphinx inventory does remove the "[anonymous]." nested elements, so a symbol can be simply written:

```
:c:enumerator:`SOME_SYMBOL`
```

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -3753,6 +3753,7 @@ class CDomain(Domain):
 
     def get_objects(self) -> Iterator[Tuple[str, str, str, str, str, int]]:
         for refname, (docname, node_id, objtype) in list(self.objects.items()):
+            refname = refname.replace('[anonymous].', '')
             yield (refname, refname, objtype, docname, node_id, 1)
 
 


utzig:
@jakobandersen Any idea when you could take a look?

jakobandersen:
Unfortunately I don't have much time at the moment, but some immediate comments:
- Can you provide a pair of test projects that reproduces the problem to make it easy to check things?
- I find it odd that ``[anonymous]`` is in the inventory. Perhaps it's the inventory generation that should be fixed (in addition).

utzig:
> * I find it odd that `[anonymous]` is in the inventory. Perhaps it's the inventory generation that should be fixed (in addition).

Yeah, I'll try that!

utzig:
> > * I find it odd that `[anonymous]` is in the inventory. Perhaps it's the inventory generation that should be fixed (in addition).
> 
> Yeah, I'll try that!

I dropped that commit which touched intersphinx, and updated the c-domain to bypass anonymous elements. The commit message was updated as well.

utzig:
> > > * I find it odd that `[anonymous]` is in the inventory. Perhaps it's the inventory generation that should be fixed (in addition).
> > 
> > 
> > Yeah, I'll try that!
> 
> I dropped that commit which touched intersphinx, and updated the c-domain to bypass anonymous elements. The commit message was updated as well.

Well, clearly there's more work to fix it properly...

utzig:
> > > > * I find it odd that `[anonymous]` is in the inventory. Perhaps it's the inventory generation that should be fixed (in addition).
> > > 
> > > 
> > > Yeah, I'll try that!
> > 
> > 
> > I dropped that commit which touched intersphinx, and updated the c-domain to bypass anonymous elements. The commit message was updated as well.
> 
> Well, clearly there's more work to fix it properly...

Giving it another try, this works OK for my projects, and it does only touch the intersphinx inventory generation now.

jakobandersen:
Thanks for your work @utzig! When reviewing it I realised that I didn't port ``get_objects()`` from the C++ domain. In  https://github.com/jakobandersen/sphinx/tree/c_get_objects I tried it now, and it essentially gives the same fix as your code, but doing away with the extra ``objects`` dictionary. I only found time to run the ordinary test suite, can you try it out and see if it works for your case?

utzig:
> Thanks for your work @utzig! When reviewing it I realised that I didn't port `get_objects()` from the C++ domain. In https://github.com/jakobandersen/sphinx/tree/c_get_objects I tried it now, and it essentially gives the same fix as your code, but doing away with the extra `objects` dictionary. I only found time to run the ordinary test suite, can you try it out and see if it works for your case?

Sure, I will do it tomorrow morning and let you know. Thanks!

utzig:
> Thanks for your work @utzig! When reviewing it I realised that I didn't port `get_objects()` from the C++ domain. In https://github.com/jakobandersen/sphinx/tree/c_get_objects I tried it now, and it essentially gives the same fix as your code, but doing away with the extra `objects` dictionary. I only found time to run the ordinary test suite, can you try it out and see if it works for your case?

Tested now, works 100%. When do you think a 3.2.2 of Sphinx will come out?

jakobandersen:
@tk0miya, I'm trying to figure out how to set up more in-depth testing of intersphinx for the C and C++ domain, e.g., for the issue raised in this PR. Is it possible to have a test with potentially two ``app`` objects? I would like to construct and inventory from raw rst input, and then use that in a full build containing missing references to be resolved by intersphinx.

utzig:
> @tk0miya, I'm trying to figure out how to set up more in-depth testing of intersphinx for the C and C++ domain, e.g., for the issue raised in this PR. Is it possible to have a test with potentially two `app` objects? I would like to construct and inventory from raw rst input, and then use that in a full build containing missing references to be resolved by intersphinx.

@jakobandersen Could you open another PR with your fix branch, so I can close this one?

jakobandersen:
Thanks for the reminder, done.

tk0miya:
>Is it possible to have a test with potentially two app objects?

Sorry for late response. Of course yes. You can use `make_app` helper to make an instance of the app object.
https://github.com/sphinx-doc/sphinx/blob/054dc5d5df73ac21f1af105aa84ba16797ebba1e/tests/test_environment.py#L24-L28

It helps to create two projects to ensure intersphinx works well. But I don't know such testcase is really needed. I think it is enough to confirm the object is output expectedly as you did in #8242.

PR #8183, opened 2020-09-08
Don't remove substitution_reference nodes (fix #7953)

### LINKED ISSUE #7953: substitution_definition nodes are missing from the doctree
Since PR #4827 all substitution_definition nodes are removed from the docutils doctree. Reasoning behind this was that this node was not being used by any builders included with Sphinx. However, third-party builders might make use of them. rinohtype, for example parses them to be able to reference the definitions in the a document template.

I think it would be better for the document tree that Sphinx passes to builders to match the standard docutils document tree as much as possible. A single builder (LaTeX in this case) having an issue with the document tree shouldn't be a reason to change the doctree for all builders.

I propose making the ´SubstitutionsDefinitionsRemover´ transform local to the LaTeX builder, as was originally intended. I can provide a PR if you wish.

Please also consider third-party builders when making changes like this (e.g. doctree) in the future. Thanks!

brechtm:
SubstitutionDefinitionsRemover is now a SphinxPostTransform, only
applied in the Sphinx builder, as was originally the case (see #4827).

### DIFF
--- a/tests/test_directive_patch.py
+++ b/tests/test_directive_patch.py
@@ -21,7 +21,8 @@ def test_code_directive(app):
             '   print("hello world")\n')
 
     doctree = restructuredtext.parse(app, text)
-    assert_node(doctree, [nodes.document, nodes.literal_block, 'print("hello world")'])
+    assert_node(doctree, ([nodes.literal_block, 'print("hello world")'],
+                          nodes.substitution_definition))
     assert_node(doctree[0], language="default", highlight_args={})
 
     # with language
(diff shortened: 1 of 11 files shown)


tests/test_directive_patch.py
```diff
@@ -21,7 +21,8 @@ def test_code_directive(app):
             '   print("hello world")\n')
 
     doctree = restructuredtext.parse(app, text)
-    assert_node(doctree, [nodes.document, nodes.literal_block, 'print("hello world")'])
+    assert_node(doctree, ([nodes.literal_block, 'print("hello world")'],
+                          nodes.substitution_definition))
```
tk0miya, line 25: Finally, I found the reason why the substituion_definition node appears suddenly here. `tests/roots/test-root/conf.py` contains a substitution definition in its `rst_epilog`. We need to remove it to be these testcases healthy. I'll work on it.

tk0miya: I just posted https://github.com/sphinx-doc/sphinx/pull/8295 to fix it. So please update this PR. Then I'll review this again.

tk0miya:
I merged this partially (without unnecessary changes of testings) at https://github.com/sphinx-doc/sphinx/pull/8354.

brechtm:
Thanks, @tk0miya. I didn't find the time yet to get back to this, so you beat me to it. Cheers!

PR #8188, opened 2020-09-09
domains: c: add missing intersphinx object types

utzig:
For intersphinx to be able to resolve references to C-Domain objects, it needs to have them declared in the obj_types dict. This is currently missing enum, enumerator, struct and union so those can't be referenced by an external project. This commit fixes the issue by adding them.

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -3607,6 +3607,10 @@ class CDomain(Domain):
         'macro': ObjType(_('macro'), 'macro'),
         'type': ObjType(_('type'), 'type'),
         'var': ObjType(_('variable'), 'data'),
+        'enum': ObjType(_('enum'), 'enum'),
+        'enumerator': ObjType(_('enumerator'), 'enumerator'),
+        'struct': ObjType(_('struct'), 'struct'),
+        'union': ObjType(_('union'), 'union'),
     }
 
     directives = {


utzig:
@jakobandersen This is one commit broken away from #8160. I am adding a project attached, with the current build already there in the html folder, but you can run them yourself if you prefer so. If you look at the obj_types html folder, everything is OK, but ext_proj which uses intersphinx to resolve the symbols some are broken (and fixed by this commit). I also found out `:c:member:` does not work and `:c:macro` can't have the `()` for it to work, but those are not my focus right now...
[c_intersphinx_samples.tar.gz](https://github.com/sphinx-doc/sphinx/files/5195187/c_intersphinx_samples.tar.gz)

utzig:
> @jakobandersen This is one commit broken away from #8160. I am adding a project attached, with the current build already there in the html folder, but you can run them yourself if you prefer so. If you look at the obj_types html folder, everything is OK, but ext_proj which uses intersphinx to resolve the symbols some are broken (and fixed by this commit). I also found out `:c:member:` does not work and `:c:macro` can't have the `()` for it to work, but those are not my focus right now...
> [c_intersphinx_samples.tar.gz](https://github.com/sphinx-doc/sphinx/files/5195187/c_intersphinx_samples.tar.gz)

Ah, I also added `struct` and `union` because they were broken, although my initial commit was only concerned with `enum` and `enumerator`!

jakobandersen:
Thanks!

Issue #8141, opened 2020-08-20
Add option to c:alias to alias structure contents


Shamino0:
Thank you very much for the `c:alias` directive.  It allows me to copy bits and pieces from my Doxygen-generated contents (produced using Breathe) to other locations in my document without breaking the links to the definitions.

For functions and macros, the alias is just what I want - I get the symbol, the return value and the parameters.

When displaying an alias of a struct, however, all I see is the name of the struct.  This isn't terribly useful - it's no better than a ``:c:struct:`foo_t` `` statement.

Would it be possible to add an option to `c:alias` so the list of the fields could be displayed with the structure name itself?

For example:
~~~~
.. c:struct:: Data

   .. c:var:: int a

   .. c:var:: double b

Assorted text here

.. c:alias:: Data
   :members:

More random text

.. c:alias:: Data
~~~~

Rendering as:

> struct **Data**
> &nbsp;&nbsp; int **a**
> &nbsp;&nbsp; double **b**
>
> Assorted text here
>
> struct **Data**
> &nbsp;&nbsp; int **a**
> &nbsp;&nbsp; double **b**
>
> More random text
>
> struct **Data**

jakobandersen:
Good idea!

PR #8207, opened 2020-09-13
C, recursive alias declarations

jakobandersen:
### Feature or Bugfix

- Feature

### Purpose
Add a ``maxdepth`` option to the ``c:alias`` directive to insert nested declarations. The default is 1, the old behaviour (just the mentioned declaration) and 0 means infinite depth.

### Relates
- Suggested in #8141

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,8 @@ Features added
 
 * #8100: html: Show a better error message for failures on copying
   html_static_files
+* #8141: C: added a ``maxdepth`` option to :rst:dir:`c:alias` to insert
+  nested declarations.
 
 Bugs fixed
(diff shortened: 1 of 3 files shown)


jakobandersen:
@Shamino0, if possible, can you try out the PR to see if it works for your use case?

In the long run I imagine that it would be beneficial to have filters (only type/variables/..., hide specific members, show only specific members, etc.), but it would be nice to see specific use cases first.

Shamino0:
It works for me.  Setting maxdepth to 2 (or 0) generates what I wanted to see.  Thanks much.

jakobandersen:
Thanks!

Issue #8209, opened 2020-09-14
Add autodata option to remove value but display type info


marlonjames:
**Is your feature request related to a problem? Please describe.**
The cocotb project is using the `autodata` directive to document a global module variable, whose default value is `None` (and assigned in an initialization function). We would prefer to remove the `= None` from the documentation but retain the type information that comes from a type hint comment in the source code.
Using the **:annotation:** option with a blank string removes the `None` value but also removes the type information. 

**Describe the solution you'd like**
Have an option for removing the value but keeping the type information. Something like the **:novalue:** option that was renamed to the current **:annotation:**.

**Describe alternatives you've considered**
Using the **:annotation:** option can replicate the type information manually: `:annotation: : TypeHint`  but this is a manual process and dual maintenance when the information is in the source file.

**Additional context**
PR where this was discussed: https://github.com/cocotb/cocotb/pull/2069

PR #8222, opened 2020-09-18
autodoc: add :novalue: option

marlonjames:
Subject: Add **:novalue:** option to display type hint but remove value

### Feature or Bugfix
- Feature

### Purpose
When documenting using the `.. autodata::` directive, the **:annotation:** option only allows displaying nothing or replacing the annotation manually.
This adds the **:novalue:** option which will display the type hint but not the value. 

### Detail
When the **:annotation:** option is used, the current behavior is preserved, and **:novalue:** option does nothing.
When **:novalue:** is used by itself, the value will not be displayed but any type hint will.

### Relates
#8209

### DIFF
--- a/doc/usage/extensions/autodoc.rst
+++ b/doc/usage/extensions/autodoc.rst
@@ -326,6 +326,15 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
    By default, without ``annotation`` option, Sphinx tries to obtain the value of
    the variable and print it after the name.
 
+   The ``novalue`` option can be used instead of a blank ``annotation`` to show the
+   type hint but not the value::
+
+      .. autodata:: CD_DRIVE
+         :novalue:
+
+   If both the ``annotation`` and ``novalue`` options are used, ``novalue`` has no
+   effect.
+
    For module data members and class attributes, documentation can either be put
    into a comment with special formatting (using a ``#:`` to start the comment
    instead of just ``#``), or in a docstring *after* the definition.  Comments
@@ -365,6 +374,9 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
       option.
    .. versionchanged:: 2.0
       :rst:dir:`autodecorator` added.
+   .. versionchanged:: 3.3
+      :rst:dir:`autodata` and :rst:dir:`autoattribute` now have a ``novalue``
+      option.
 
    .. note::
 
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1598,6 +1598,7 @@ class DataDocumenter(ModuleLevelDocumenter):
     priority = -10
     option_spec = dict(ModuleLevelDocumenter.option_spec)
     option_spec["annotation"] = annotation_option
+    option_spec["novalue"] = bool_option
 
     @classmethod
     def can_document_member(cls, member: Any, membername: str, isattr: bool, parent: Any
@@ -1633,7 +1634,7 @@ class DataDocumenter(ModuleLevelDocumenter):
                                   sourcename)
 
             try:
-                if self.object is UNINITIALIZED_ATTR:
+                if self.object is UNINITIALIZED_ATTR or self.options.novalue:
                     pass
                 else:
                     objrepr = object_description(self.object)
@@ -1917,6 +1918,7 @@ class AttributeDocumenter(DocstringStripSignatureMixin, ClassLevelDocumenter):  
     member_order = 60
     option_spec = dict(ModuleLevelDocumenter.option_spec)
     option_spec["annotation"] = annotation_option
+    option_spec["novalue"] = bool_option
 
     # must be higher than the MethodDocumenter, else it will recognize
     # some non-data descriptors as methods
--- a/tests/roots/test-ext-autodoc/index.rst
+++ b/tests/roots/test-ext-autodoc/index.rst
@@ -11,3 +11,16 @@
 .. autofunction:: target.typehints.incr
 
 .. autofunction:: target.typehints.tuple_args
+
+.. autodata:: target.typed_vars_py35.attr1
+
+.. autodata:: target.typed_vars_py35.attr2
+   :annotation:
+   :novalue:
+
+.. autodata:: target.typed_vars_py35.attr3
+   :annotation: = some integer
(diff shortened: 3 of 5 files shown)


doc/usage/extensions/autodoc.rst
```diff
@@ -325,7 +325,16 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
 
    By default, without ``annotation`` option, Sphinx tries to obtain the value of
    the variable and print it after the name.
+   
+   The ``novalue`` option can be used instead of a blank ``annotation`` to show the
```
cmarqu, line 329: I guess ``:novalue:`` should be mentioned in a ``.. versionchanged::`` at the end of this section.

ktbarrett:
To clarify, this exists to support Python 3.5 and 2.7 compliant codebases where uninitialized variables cannot exist, but they still want to use type hints with Sphinx autodoc. Support for this case is currently missing.

doc/usage/extensions/autodoc.rst
```diff
@@ -326,6 +326,15 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
    By default, without ``annotation`` option, Sphinx tries to obtain the value of
    the variable and print it after the name.
 
+   The ``novalue`` option can be used instead of a blank ``annotation`` to show the
```
tk0miya, line 329: In autodoc, we usually use `no-` prefix to disable some feature (ex. `no-members`). So it is better to rename this to `no-value` (if we adopted this proposal)

tk0miya (commented):
Now both python 3.5 and 2.7 reached EOL. So I wonder this is still needed.

doc/usage/extensions/autodoc.rst
ktbarrett:
```suggestion
   The ``no-value`` option can be used instead of a blank ``annotation`` to show the
```

doc/usage/extensions/autodoc.rst
```diff
@@ -326,6 +326,15 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
...
+   type hint but not the value::
+
+      .. autodata:: CD_DRIVE
+         :novalue:
```
ktbarrett, line 333:
```suggestion
         :no-value:
```

doc/usage/extensions/autodoc.rst
```diff
@@ -326,6 +326,15 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
...
+
+   If both the ``annotation`` and ``novalue`` options are used, ``novalue`` has no
```
ktbarrett, line 335:
```suggestion
   If both the ``annotation`` and ``no-value`` options are used, ``no-value`` has no
```

doc/usage/extensions/autodoc.rst
```diff
@@ -365,6 +374,9 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
       option.
    .. versionchanged:: 2.0
       :rst:dir:`autodecorator` added.
+   .. versionchanged:: 3.3
+      :rst:dir:`autodata` and :rst:dir:`autoattribute` now have a ``novalue``
```
ktbarrett, line 378:
```suggestion
      :rst:dir:`autodata` and :rst:dir:`autoattribute` now have a ``no-value``
```

sphinx/ext/autodoc/__init__.py
```diff
@@ -1598,6 +1598,7 @@ class DataDocumenter(ModuleLevelDocumenter):
     priority = -10
     option_spec = dict(ModuleLevelDocumenter.option_spec)
     option_spec["annotation"] = annotation_option
+    option_spec["novalue"] = bool_option
```
ktbarrett, line 1601:
```suggestion
    option_spec["no-value"] = bool_option
```

doc/usage/extensions/autodoc.rst
ktbarrett: Would that imply this PR would require implementing the non-negated version of `:value:` as well?

tk0miya: If somebody requests it. But there are no such requests now, AFAIK.

doc/usage/extensions/autodoc.rst
```diff
@@ -365,6 +374,9 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
       option.
    .. versionchanged:: 2.0
       :rst:dir:`autodecorator` added.
+   .. versionchanged:: 3.3
```
tk0miya, line 377: I determined to merge this into 3.4. Please update this.

tests/roots/test-ext-autodoc/index.rst
```diff
@@ -11,3 +11,16 @@
 .. autofunction:: target.typehints.incr
 
 .. autofunction:: target.typehints.tuple_args
+
+.. autodata:: target.typed_vars_py35.attr1
```
tk0miya, line 15: Basically, we don't need to update this file because we have `do_autodoc()` helper.

tests/roots/test-ext-autodoc/target/typed_vars_py35.py
```diff
@@ -0,0 +1,11 @@
+#: attr1
```
tk0miya, line 1: `py35` suffix is no longer needed because Sphinx supports py35 or above.

I think `no-value` option is not related to "typed variables" at all. So no reason to add new examples. How about using `target.integer` instead?

tests/test_ext_autodoc.py
```diff
@@ -2024,6 +2025,151 @@ def test_pyclass_for_ClassLevelDocumenter(app):
     ]
 
 
+@pytest.mark.skipif(sys.version_info < (3, 6), reason='py36+ is available since python3.6.')
+@pytest.mark.sphinx('html', testroot='ext-autodoc')
+def test_pydata_for_DataDeclarationDocumenter(app):
```
tk0miya, line 2030: Is there any reason to test with all of data-declaration, py36-style typed attributes and py35-style typed attributes?

tk0miya:
Now I posted #8424 as a refined version of this PR. I'd like to merge it instead of this in a few days. Please let me know your opinion :-)

PR #8229, opened 2020-09-21
Fix #8219: autodoc: A signature of constructor of generic class is incorrect

### LINKED ISSUE #8219: autodoc doesn't handle TypeVars or document user-defined generic classes accurately
**Describe the bug**

1. TypeVar aliases aren't resolved  
When using automodule, if a TypeVar is used, the TypeVar isn't being recognized. With nitpick on, this warning is generated:
`helloworld.py:docstring of helloworld.helloworld.MyGenericType.foo:: WARNING: py:class reference target not found: MyTypeVar`  
I can add it to the nitpick ignore list, but then all of the parameters that reference it aren't recognized and linked.

2. `typing.Generic` type parameters for user-defined generic classes aren't documented  

3. Constructor for user-defined generic classes are always `(*args, **kwargs)` instead of what's actually declared.

**To Reproduce**

Simple doc file `index.rst`:
```
Hello World
===========

code docs
=========

.. automodule:: helloworld.helloworld
    :members:
    :undoc-members:
    :show-inheritance:
```

Here's an example module `helloworld.py`:

```python
from typing import Generic, TypeVar

MyTypeVar = TypeVar('MyTypeVar')

class MyGenericType(Generic[MyTypeVar]):
    def __init__(self, param: MyTypeVar) -> None:
        pass

    def foo(self, param: MyTypeVar) -> None:
        pass
```

**Expected behavior**
Ideally:
1. I would expect `TypeVar` definitions to be recognized as types
2. Generic classes should document the `TypeVar`s passed to `typing.Generic`
3. Constructor for generic classes should reflect the parameters declared in the constructor.

**Work-around for Item 1**

This is less than ideal, but I can get the TypeVar to resolve by manually documenting it as an attribute like this:
```
.. py:attribute:: MyTypeVar
   :type: TypeVar
   :value: TypeVar('MyTypeVar')

This is the type used in MyGenericClass

.. automodule:: helloworld.helloworld
    :members:
    :undoc-members:
    :show-inheritance:
```

**Your project**
I've included a simple example above

**Screenshots**
Here's the generated documentation:
image

Here's the generated output with the manual attribute work-around:
image

**Environment info**
- OS: Ubuntu 18.04, Ubuntu 20.04
- Python version: 3.6.9, 3.8.2
- Sphinx version: 3.1.2, 3.2.1
- Extra tools: Viewing output on Chrome

**Additional context**

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #8219 
- Ignore the signature of typing.Generic.__new__() to get correct a
signature of the constructor from a subclass.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -25,6 +25,7 @@ Bugs fixed
 * #8143: autodoc: AttributeError is raised when False value is passed to
   autodoc_default_options
 * #8103: autodoc: functools.cached_property is not considered as a property
+* #8219: autodoc: A signature of constructor of generic class is incorrect
 * #8192: napoleon: description is disappeared when it contains inline literals
 * #8169: LaTeX: pxjahyper loaded even when latex_engine is not platex
 * #8093: The highlight warning has wrong location in some builders (LaTeX,
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1312,11 +1312,15 @@ class DecoratorDocumenter(FunctionDocumenter):
             return None
 
 
-# Types which have confusing metaclass signatures it would be best not to show.
-# These are listed by name, rather than storing the objects themselves, to avoid
-# needing to import the modules.
+# Types which have confusing metaclass signatures and superclass signatures it
+# would be best not to show.  These are listed by name, rather than storing the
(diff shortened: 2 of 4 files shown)


tk0miya:
dup with #8142.

PR #8231, opened 2020-09-21
Fix #8219: autodoc: Parameters for generic base class are not shown

### LINKED ISSUE #8219: autodoc doesn't handle TypeVars or document user-defined generic classes accurately

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #8219

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -37,7 +37,7 @@ from sphinx.util.docstrings import extract_metadata, prepare_docstring
 from sphinx.util.inspect import (
     evaluate_signature, getdoc, object_description, safe_getattr, stringify_signature
 )
-from sphinx.util.typing import stringify as stringify_typehint
+from sphinx.util.typing import restify, stringify as stringify_typehint
 
 if False:
     # For type annotation
@@ -1574,13 +1574,16 @@ class ClassDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # type: 
         if not self.doc_as_attr and self.options.show_inheritance:
             sourcename = self.get_sourcename()
             self.add_line('', sourcename)
-            if hasattr(self.object, '__bases__') and len(self.object.__bases__):
-                bases = [':class:`%s`' % b.__name__
-                         if b.__module__ in ('__builtin__', 'builtins')
-                         else ':class:`%s.%s`' % (b.__module__, b.__qualname__)
-                         for b in self.object.__bases__]
-                self.add_line('   ' + _('Bases: %s') % ', '.join(bases),
-                              sourcename)
+
+            if hasattr(self.object, '__orig_bases__') and len(self.object.__orig_bases__):
+                # A subclass of generic types
+                # refs: PEP-560 <https://www.python.org/dev/peps/pep-0560/>
+                bases = [restify(cls) for cls in self.object.__orig_bases__]
+                self.add_line('   ' + _('Bases: %s') % ', '.join(bases), sourcename)
+            elif hasattr(self.object, '__bases__') and len(self.object.__bases__):
+                # A normal class
+                bases = [restify(cls) for cls in self.object.__bases__]
+                self.add_line('   ' + _('Bases: %s') % ', '.join(bases), sourcename)
 
     def get_doc(self, encoding: str = None, ignore: int = None) -> List[List[str]]:
         if encoding is not None:
--- a/CHANGES
+++ b/CHANGES
@@ -54,6 +54,9 @@ Features added
 Bugs fixed
 ----------
 
+* #8219: autodoc: Parameters for generic class are not shown when super class is
+  a generic class and show-inheritance option is given (in Python 3.7 or above)
+
 Testing
 --------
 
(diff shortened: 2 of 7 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -1485,13 +1494,32 @@ def add_directive_header(self, sig: str) -> None:
...
             self.add_line('', sourcename)
-            if hasattr(self.object, '__bases__') and len(self.object.__bases__):
-                bases = [':class:`%s`' % b.__name__
-                         if b.__module__ in ('__builtin__', 'builtins')
-                         else ':class:`%s.%s`' % (b.__module__, b.__qualname__)
-                         for b in self.object.__bases__]
-                self.add_line('   ' + _('Bases: %s') % ', '.join(bases),
-                              sourcename)
+
+            def restify(cls: "Type") -> str:
```
tk0miya: This helper fails when a generic base class takes a `Union` object as a parameter.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1485,13 +1494,32 @@ def add_directive_header(self, sig: str) -> None:
...
+                    if cls.__args__:
+                        text += r"\ [%s]" % ", ".join(restify(a) for a in cls.__args__)
+                    return text
+                elif hasattr(cls, '__qualname__'):
+                    return ':class:`%s.%s`' % (cls.__module__, cls.__qualname__)
+                else:
+                    # not a class (ex. TypeVar)
+                    return ':obj:`%s.%s`' % (cls.__module__, cls.__name__)
+
+            if hasattr(self.object, '__orig_bases__') and len(self.object.__orig_bases__):
```
tk0miya, line 1578: It would be better to mention PEP-560 here as a comment.
https://www.python.org/dev/peps/pep-0560/

sphinx/ext/autodoc/__init__.py
```diff
@@ -1485,13 +1494,32 @@ def add_directive_header(self, sig: str) -> None:
...
+                bases = [restify(cls) for cls in self.object.__orig_bases__]
+                self.add_line('   ' + _('Bases: %s') % ', '.join(bases), sourcename)
+            elif hasattr(self.object, '__bases__') and len(self.object.__bases__):
```
tk0miya, line 1583: Is this still needed?

sphinx/ext/autodoc/__init__.py
```diff
@@ -1388,6 +1392,11 @@ def get_user_defined_function_or_method(obj: Any, attr: str) -> Any:
 
         # Now we check if the 'obj' class has a '__new__' method
         new = get_user_defined_function_or_method(self.object, '__new__')
+
+        if new is not None:
+            if "{0.__module__}.{0.__qualname__}".format(new) in _SUPERCLASS_NEW_BLACKLIST:
+                new = None
```
tk0miya: Oops. This is duplicated with #8142.

sphinx/ext/autodoc/__init__.py
tk0miya: Yes, non-generic-subclasses don't have `__orig_bases__` even after PEP-560. So this code is still needed.

PR #8234, opened 2020-09-22
FIX Unwrap user-defined functions or methods to get correct signature

alfaro96:
Subject: This PR fixes the signature of user-defined functions or methods that are wrapped by a decorator.  

### Feature or Bugfix

- Bugfix

### Purpose
- Unwrap the user-defined functions or methods to obtain the correct signature instead of the decorator signature.

### Relates
- scikit-learn/scikit-learn#18434


tk0miya:
Since 3.4.0, we determined to unwrap all of functions and methods on getting signature (refs: #8362). So this change is no longer needed. Thank you for your proposal.

PR #8239, opened 2020-09-23
Fix references to production names.

silene:
### Feature or Bugfix
- Bugfix

### Purpose

When the user tries to align production rules, Sphinx gets confused by spaces around leading tokens. In the following, references to `mygroup:bad1` and `mygroup:bad2` are missing, while the one to `mygroup:correct` is present.

    .. productionlist:: mygroup
       correct: `correct` | `bad1` | `bad2`
       bad1   : `correct` | `bad1` | `bad2`
          bad2: `correct` | `bad1` | `bad2`

Since it is useful to be able to align colons in a production list, this commit systematically strips leading and trailing whitespaces from the production name.

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -500,7 +500,8 @@ class ProductionList(SphinxDirective):
             except ValueError:
                 break
             subnode = addnodes.production(rule)
-            subnode['tokenname'] = name.strip()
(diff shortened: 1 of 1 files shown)


tk0miya:
Fair enough. Thank you for your contribution!

PR #8242, opened 2020-09-25
C, fix names in get_objects()

jakobandersen:
### Feature or Bugfix

- Bugfix

### Purpose
Intersphinx uses ``get_objects()`` on a domain to enumerate all recorded objects. This PR fixes the logical name returned in each entry such that intersphinx can do proper lookups for anon entities. Previously the display name was returned.

### Detail
Fixes/replaces sphinx-doc/sphinx#8160

No tests are added yet (could perhaps wait for another PR). See also https://github.com/sphinx-doc/sphinx/pull/8160#issuecomment-692959069

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Bugs fixed
 
 * #8188: C, add missing items to internal object types dictionary,
   e.g., preventing intersphinx from resolving them.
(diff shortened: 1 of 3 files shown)


utzig (approved):
LGTM

PR #8245, opened 2020-09-27
linkcheck: take source directory into account for local files

mgeier:
In  #7985, the `linkcheck` builder has been extended to also check local links.

However, the local file URIs are checked relative to the global source directory.
If a relative path is used in a source file in a sub-directory, the file is not found and a linkcheck error is reported.

This PR uses the actual source directory of the current source file.

### Feature or Bugfix

- Bugfix

### Relates
- #7985

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -210,7 +210,7 @@ class CheckExternalLinksBuilder(Builder):
                 else:
                     return 'redirected', new_url, 0
 
-        def check() -> Tuple[str, str, int]:
+        def check(srcdir: str) -> Tuple[str, str, int]:
             # check for various conditions without bothering the network
             if len(uri) == 0 or uri.startswith(('#', 'mailto:')):
                 return 'unchecked', '', 0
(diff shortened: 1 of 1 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -256,7 +256,8 @@ def check() -> Tuple[str, str, int]:
             uri, docname, lineno = self.wqueue.get()
             if uri is None:
                 break
-            status, info, code = check()
+            srcdir = path.dirname(self.env.doc2path(docname))
+            status, info, code = check(srcdir)
```
tk0miya: It would be better to pass docname instead of srcdir.

tk0miya (changes requested):

sphinx/builders/linkcheck.py
mgeier: Done: 6b3d4458796d3ed2f536c09b60c2573a276db417

tk0miya (approved):

tk0miya:
Thank you for update. Merged!

PR #8246, opened 2020-09-28
domain: c: fix macro usage inside struct

utzig:
This fixes issues seen when a macro is used inside a struct. A simple example:

```
/** @brief some struct */
typedef struct {
    /** Some macro */
    SOME_MACRO(a, b);
} t;
```

If `SOME_MACRO` was declared outside of the file, Doxygen generates a function element. There current c-domain routines do not handle untyped functions, so this commit adds an extra check when a typed function parsing generates a DefinitionError to also try to parse it as an untyped function.

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -2897,8 +2897,25 @@ class DefinitionParser(BaseParser):
                         decl = self._parse_declarator(named=True, paramMode=outer,
                                                       typed=typed)
         elif outer == 'function':
-            declSpecs = self._parse_decl_specs(outer=outer)
-            decl = self._parse_declarator(named=True, paramMode=outer)
+            start_pos = self.pos
+            prev_errors = []
+            try:
+                declSpecs = self._parse_decl_specs(outer=outer)
+                decl = self._parse_declarator(named=True, paramMode=outer)
+            except DefinitionError as ex_typed:
+                desc = "If the function has a return type"
+                prev_errors.append((ex_typed, desc))
+                self.pos = start_pos
+                try:
+                    declSpecs = self._parse_decl_specs(outer=outer, typed=False)
+                    decl = self._parse_declarator(named=True, paramMode=outer,
+                                                  typed=False)
+                except DefinitionError as ex_untyped:
+                    self.pos = startPos
+                    desc = "If the function has no return type"
+                    prev_errors.append((ex_untyped, desc))
+                    header = "Error when parsing function declaration."
+                    raise self._make_multi_error(prev_errors, header) from ex_untyped
         else:
             paramMode = 'type'
             if outer == 'member':  # i.e., member


utzig:
@jakobandersen Code similar to this already exists in the cpp domain parser; to minimize the changes, I just worked the other way around, first try typed parsing then untyped. I build two huge documentation sets, https://docs.zephyrproject.org/latest/ and http://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/index.html, and haven't found any side-effects.

utzig:
The error seen before this fix:

```
/home/utzig/src/Nordic/ncs/tmp/macro_issue/index.rst:6: WARNING: Error in declarator
If declarator-id with parameters (e.g., 'void f(int arg)'):
  Invalid C declaration: Expected identifier in nested name. [error at 11]
    SOME_MACRO (a, b)
    -----------^
If parenthesis in noptr-declarator (e.g., 'void (*f(int arg))(double)'):
  Error in declarator or parameters
  Invalid C declaration: Expecting "(" in parameters. [error at 13]
    SOME_MACRO (a, b)
    -------------^
```

jakobandersen:
Hmm, I'm not sure the C domain should handle this. I guess the macro is disguising a member declaration?
The C++ domain does not support untyped functions as such, but certain member function just don't have a return type (e.g., constructors).

utzig:
> Hmm, I'm not sure the C domain should handle this. I guess the macro is disguising a member declaration?

Well, I guess it depends on what that really is, in our implementation they are not functions at all, just some funny way of declaring an array or something (and there is a good reason for the use of a macro do provide abstraction). But Doxygen still thinks it's a function at least until it is resolved.

> The C++ domain does not support untyped functions as such, but certain member function just don't have a return type (e.g., constructors).

In this case it might be a bug in C++ domain, but not a bad bug :-)

jakobandersen:
> > Hmm, I'm not sure the C domain should handle this. I guess the macro is disguising a member declaration?
> 
> Well, I guess it depends on what that really is, in our implementation they are not functions at all, just some funny way of declaring an array or something (and there is a good reason for the use of a macro do provide abstraction). But Doxygen still thinks it's a function at least until it is resolved.

Right, I think Doxygen should be fixed in this case. If it is parsed by Sphinx as a function, then it will for example say "C function" in the index and search results for this entry. And if the macro is used multiple times, then you will get duplicate declaration warnings. So, for now I'll close this. At best Doxygen can expand it, but otherwise perhaps Breathe can make an adaptation if there is enough in the XML.

> > The C++ domain does not support untyped functions as such, but certain member function just don't have a return type (e.g., constructors).
> 
> In this case it might be a bug in C++ domain, but not a bad bug :-)

Well, the effect is the same :-), you can give it "untyped functions" without parsing errors, but I can't guarantee that there won't be silent other problems because the domain assumes this form is only used for constructors.

PR #8247, opened 2020-09-28
Allow production lists to refer to tokens from other production groups.

silene:
### Feature or Bugfix
- Feature

### Purpose

Outside production lists, syntax `` `foo:bar` `` already makes it possible to refer to the production `bar` of group `foo`. This commit offers the same feature inside production lists. Similarly to the reference syntax, prefixing with a tilde prevents the group from being displayed.

This commit also makes it possible to use `` `:bar` `` to refer to production `bar` from a production list without a group name. This is especially useful when one has a main (unnamed) grammar and one or several named extensions that need to refer to it.

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -1206,20 +1206,29 @@ the definition of the symbol.  There is this directive:
    the following definition.  If the definition spans multiple lines, each
    continuation line must begin with a colon placed at the same column as in
    the first line.
+   Blank lines are not allowed within ``productionlist`` directive arguments.
+
+   The definition can contain token names which are marked as interpreted text
+   (e.g., "``sum ::= `integer` "+" `integer```") -- this generates
+   cross-references to the productions of these tokens.  Outside of the
+   production list, you can reference to token productions using
+   :rst:role:`token`.
 
    The *productionGroup* argument to :rst:dir:`productionlist` serves to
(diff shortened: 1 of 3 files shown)


tk0miya (changes requested):
Looks good. Could you add testcases for this new syntax?

silene:
I am a bit confused with the testing framework. I suppose I can add some testcases to `tests/roots/test-productionlist/`. But are there some reference files somewhere else that are used to assert that the output is correct? Or is it just a matter of getting the files to compile without errors/warnings?

tk0miya:
This is a testcase that corresponds with `tests/roots/test-productionlist/`. It builds the document and inspect the result. 
https://github.com/sphinx-doc/sphinx/blob/3.x/tests/test_domain_std.py#L352-L405

It's okay to modify them or to add new pair of project under `tests/roots/` and testcase in `test_domain_std.py`.

silene:
Thanks for the help. It seemed a bit too frail and tedious to parse the HTML output. (Contrarily to the other tests for `productionlist`, it is the productions themselves that had to be parsed, and they have no structure.) So, I instead decided to check the doctree, as many other tests do.

tk0miya (approved):

tk0miya:
Thank you for your work. LGTM!

tk0miya:
Note: I was wrongly labeled this as "4.0.0". But it's unexpected.

Issue #7600, opened 2020-05-03
Feature request: Adding html_touch_icon setting.


AtsushiSakai:
**Is your feature request related to a problem? Please describe.**
When I create a document with Sphinx, I want to refer it from iOS or Android. But when I bookmark the doc link, it is not beautiful because it does not have an icon for it. 

**Describe the solution you'd like**
Sphinx already has "html_fabicon" setting. Using the similar way, it would be great to add a "html_touch_icon" or something like that for setting apple touch icon tag in setup.py.

tk0miya:
+1. It would be better to support other icons because there are many kinds of icons.
https://gist.github.com/hsquareweb/2210110

bienchen:
Hello,
is somebody already working on this? I could use the same feature in my project, plus a bunch of other "fancy favicons". If nobody has his already covered, I could prepare something.

Cheers,

B

tk0miya:
I guess nobody is working on this yet. I tried to consider an option for this once. But I could not find a good name for this because there are many kinds of icon tags... So any ideas and PRs are welcome :-)
https://gist.github.com/hsquareweb/2210110

bienchen:
I have now a version that sets the touch icon for the latest iOS version. For me that’s enough. Just need to write the documentation, then I can place a merge request. I still want to do the Windows tile and Android icon. 

Sent by my Toaster.

tk0miya:
Finally, I thought most users do not want to configure the icon in detail. So it would be better to give an image file and convert it to the appropriate size and generate tags.
```
html_icon = 'path/to/icon.png'  # the icon will be converted to the appropriate size of touch icons, ms-application squares and so on.
```

It would be better to provide a way to configure them in detail. But it's not too late if somebody will request it.

tcmetzger:
@tk0miya To continue our discussion from https://github.com/sphinx-doc/sphinx/pull/8249:

[...]

PR #8248, opened 2020-09-28
Feature fancy favicons

bienchen:
Subject: Allow Sphinx to handle Apple touch icons

### Feature or Bugfix

- Feature

### Purpose
Introduces a new HTML builder variable `html_apple_touch_icon` that gets the path to an image file and causes this file being copied to `_static/apple-touch-icon.png`. With that, if an iPhone creates a link to the HTML page on the home screen, this image will be used as icon.

Missing: Translation of the message from the validator. I tried to get this in, but with the commands given on https://www.sphinx-doc.org/en/master/internals/contributing.html#translations, more files than I expect get changed plus I see changes I don't understand.

To be considered: Apple suggests icons in multiple dimensions for different devices. There are 4-5 sizes mentioned at https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon/. Those could all be introduced and i would suggest to name suffix the builder variables with `_<size>`. If this is favourable, this pull request can be skipped and I add the various sizes to create a new pull request. However, I would also like to have some more "favicon variables" for the new style 32x32/ 16x16 icons, MS Windows Tiles and Safari pinned tab icons... not sure if that is desirable for Sphinx and how to proceed with that.

### Relates
- #7600

### DIFF
--- a/doc/_static/conf.py.txt
+++ b/doc/_static/conf.py.txt
@@ -148,6 +148,12 @@ html_theme = 'alabaster'
 #
 # html_favicon = None
 
+# The name of an image file (relative to this directory) to use as touch icon
+# for a home screen link in iOS. This file should be a PNG file (.png) being
+# 180x180 pixels large without transparency.
+#
+# html_apple_touch_icon = None
+
 # Add any paths that contain custom static files (such as style sheets) here,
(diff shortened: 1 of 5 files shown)


bienchen:
flake8 finds a formatting error. Going to fix it and re-submit.

PR #8249, opened 2020-09-28
Feature fancy favicons

bienchen:
Subject: Allow Sphinx to handle Apple touch icons

### Feature or Bugfix

- Feature

### Purpose
Introduces a new HTML builder variable `html_apple_touch_icon` that gets the path to an image file and causes this file being copied to `_static/apple-touch-icon.png`. With that, if an iPhone creates a link to the HTML page on the home screen, this image will be used as icon.

Missing: Translation of the message from the validator. I tried to get this in, but with the commands given on https://www.sphinx-doc.org/en/master/internals/contributing.html#translations, more files than I expect get changed plus I see changes I don't understand.

To be considered: Apple suggests icons in multiple dimensions for different devices. There are 4-5 sizes mentioned at https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon/. Those could all be introduced and i would suggest to name suffix the builder variables with `_<size>`. If this is favourable, this pull request can be skipped and I add the various sizes to create a new pull request. However, I would also like to have some more "favicon variables" for the new style 32x32/ 16x16 icons, MS Windows Tiles and Safari pinned tab icons... not sure if that is desirable for Sphinx and how to proceed with that.

### Relates
- #7600

### DIFF
--- a/doc/_static/conf.py.txt
+++ b/doc/_static/conf.py.txt
@@ -148,6 +148,12 @@ html_theme = 'alabaster'
 #
 # html_favicon = None
 
+# The name of an image file (relative to this directory) to use as touch icon
+# for a home screen link in iOS. This file should be a PNG file (.png) being
+# 180x180 pixels large without transparency.
+#
+# html_apple_touch_icon = None
+
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
--- a/doc/templating.rst
+++ b/doc/templating.rst
@@ -276,6 +276,10 @@ in the future.
 
    The path to the HTML favicon in the static path, or ``''``.
 
+.. data:: apple_touch_icon
+
+   The path to the iOS home screen icon in the static path, or ``''``.
+
 .. data:: file_suffix
 
    The value of the builder's :attr:`~.SerializingHTMLBuilder.out_suffix`
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -977,6 +977,15 @@ that use Sphinx's HTMLWriter class.
       The image file will be copied to the ``_static`` directory of the output
       HTML, but only if the file does not already exist there.
 
+.. confval:: html_apple_touch_icon
+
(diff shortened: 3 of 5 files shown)


tk0miya:
Thank you for your proposal. LGTM! But it is better to consider various sized icons and other icons. It will be annoying if we introduce config variables for each platforms like `msapplication_tile_icon`, `safari_mask_icon` and so on. So it would be better to consider a better configuration way for this.

But some kinds of icon tags use `<link>`, but some use `<meta>`. So we have to specify the tag-name to the configuration like this:
```
html_icons = [
  {'tag': 'link', 'rel': 'apple-touch-icon', 'href': '/images/touch-icon-57.png', 'sizes': '57x57'}
  ...
]
```
It's terrible...

bienchen:
Hm, OK, I get what you are going for. I can do something like that instead of my last pull request. Will take some time since my holidays are almost over. I'll first check if that covers all the cases of "extended favicons" I know. What I don't know is if the "href" attribute will work that way. The html_favicon is copied to static automatically without declaring the "href" bit. The user also doesn't have control of where exactly it is copied. So either we skip "href" (and copy everything to static) or the "html_icons" setting needs to trigger a mechanism assuring that the directory structure of the static dir corresponds to the "href" settings. Some older browser still need those files in the server root, I can check if that would be possible with an individual "href" setting of the html_icons...

tcmetzger:
That would be great to have! I worked on something similar for https://github.com/pydata/pydata-sphinx-theme, together with @jorisvandenbossche and @choldgraf: https://github.com/pydata/pydata-sphinx-theme/pull/359

We currently support the following configuration values:
```
   html_theme_options = {
      "favicons": [
         {
            "rel": "icon",
            "sizes": "16x16",
            "href": "https://secure.example.com/favicon/favicon-16x16.png",
         },
      ]
   }
```
Is this PR still actively developed? If not, I'd propose opening a new PR based on the work we did for pydata-sphinx-theme. This might also be relevant for https://github.com/sphinx-doc/sphinx/issues/7600.

tk0miya:
It seems this PR is stalled now. My large concern for this topic is its interface. So it would be better to discuss before implementation. I don't still find a good way to support many icon formats.

PR #8257, opened 2020-10-01
add parallel to SphinxTestApp

rscohn2:
Subject: allow parallel with SphinxTestApp

### Feature or Bugfix
- Feature

### Purpose
- Make it easier to test extensions with parallel build. I am trying to get all the extensions I use to support parallel read. Some are using SphinxTestApp, so supporting parallel makes it easier to add tests.

### DIFF
--- a/tests/test_build_html.py
+++ b/tests/test_build_html.py
@@ -421,11 +421,10 @@ def test_html5_output(app, cached_etree_parse, fname, expect):
     check_xpath(cached_etree_parse(app.outdir / fname), fname, *expect)
 
 
-@pytest.mark.skipif(docutils.__version_info__ < (0, 13),
-                    reason='docutils-0.13 or above is required')
 @pytest.mark.sphinx('html', parallel=2)
 def test_html_parallel(app):
     app.build()
+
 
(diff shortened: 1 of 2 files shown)


tests/test_build_html.py
```diff
@@ -421,6 +421,12 @@ def test_html5_output(app, cached_etree_parse, fname, expect):
     check_xpath(cached_etree_parse(app.outdir / fname), fname, *expect)
 
 
+@pytest.mark.skipif(docutils.__version_info__ < (0, 13),
+                    reason='docutils-0.13 or above is required')
```
tk0miya: Why are you marking this as docutils-0.13 or above only?

rscohn2: I copied the example below. I don't have any understanding of why it might be needed. Should I remove it?

tk0miya: Yes, please. The above and below test cases surely depends on docutils-0.13+. But this is not.

rscohn2: Removed unnecessary skipif

tk0miya:
Thanks!

PR #8260, opened 2020-10-01
Makefiles: Include clean in help message

Blendify:
Subject: The make clean help command was missing from all make files.

### DIFF
--- a/sphinx/templates/quickstart/Makefile_t
+++ b/sphinx/templates/quickstart/Makefile_t
@@ -95,3 +95,4 @@ gettext:
 .PHONY: Makefile
 %: Makefile
 	$(SPHINXBUILD) -b "$@" $(ALLSPHINXOPTS) "$(BUILDDIR)/$@"
+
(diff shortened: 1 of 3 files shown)


sphinx/templates/quickstart/Makefile_t
```diff
@@ -94,4 +95,3 @@ gettext:
 .PHONY: Makefile
 %: Makefile
 	$(SPHINXBUILD) -b "$@" $(ALLSPHINXOPTS) "$(BUILDDIR)/$@"
-
```
tk0miya: Why removing this?

tk0miya (commented):
LGTM with nits.

sphinx/templates/quickstart/Makefile_t
Blendify: I believe my editor was configured to strip more that 1 trailing white line at the end of the file. The file had two blank lines at the bottom. If this change is too noisy I can revert it.

tk0miya: It seems our test case expects a carriage return before EOF. Please revert this line or update a test case.

tk0miya:
Thank you for the update. Merging now.

PR #8262, opened 2020-10-02
Expand autodoc's `_METACLASS_CALL_BLACKLIST` for Hypothesis

Zac-HD:
Subject: supporting Hypothesis metaclass in autodoc

### Feature or Bugfix
- Feature (a very small one)

### Purpose

Hypothesis provides an abstract base class `ExampleDatabase`, whose metaclass defines a custom `__call__` method.  Unfortunately, at present this takes priority over the actual `__init__` signature of concrete subclasses ([see the current docs here](https://hypothesis.readthedocs.io/en/latest/database.html#hypothesis.database.DirectoryBasedExampleDatabase) and compare to the signature behind `[source]` link).

While I have investigated various ways to fix this for both our own docs and those of downstream providers of concrete database classes, the cleanest by far appears to be adding it to this list upstream.

### Relates
- Investigated as part of https://github.com/HypothesisWorks/hypothesis/pull/2622
- Similar in spirit to https://github.com/PyCQA/astroid/pull/820

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1322,6 +1322,7 @@ class DecoratorDocumenter(FunctionDocumenter):
 # needing to import the modules.
 _METACLASS_CALL_BLACKLIST = [
     'enum.EnumMeta.__call__',
+    'hypothesis.database._EDMeta.__call__',
 ]
 
 


tk0miya:
Thank you for your proposal. But I can't merge this PR for two reasons. 1) I'm afraid if we have to manage many metaclasses in the world via `_METACLASS_CALL_BLACKLIST`. I'm okay to list python standard classes up into the list. But it is too much to list 3rd party's up. It would be better to provide an API to enhance it if we have to do that. 2) It would be better to use docstring defined in the subclass preferentially. So what we should do at first is changing the priority of the docstring searching mechanism.

Zac-HD:
> Thank you for your proposal. But I can't merge this PR for two reasons. 1) I'm afraid if we have to manage many metaclasses in the world via `_METACLASS_CALL_BLACKLIST`. I'm okay to list python standard classes up into the list. But it is too much to list 3rd party's up. 

That makes sense to me :slightly_smiling_face:

I think there is probably not enough interest in this problem for a plugin mechanism to be worth maintaining - I would probably be the only user, and I can carefully append to this list from outside instead.

> 2) It would be better to use docstring defined in the subclass preferentially. So what we should do at first is changing the priority of the docstring searching mechanism.

The docstrings are working as expected - my "problem" is that class signatures are being drawn from the `__call__` method on the class rather than `__init__`.  Which is perfectly logical, and I think our fault for doing something very strange (which seemed like a good idea several years ago...).

Thank you for your prompt reply, and for all your work on Sphinx.  It makes a huge difference to us :heart:

tk0miya:
>The docstrings are working as expected 

Oops. That's my bad. I'd like to mention about signature, not docstring.

RuiLoureiro:
I'm having the same "problem". Did you find any workaround that doesn't involve changing `_METACLASS_CALL_BLACKLIST`? @Zac-HD

PR #8297, opened 2020-10-06
Local autodoc options override or extend autodoc_default_options.

pbudzyns:
Subject: local :exclude-members: extends global option instead of being replaced by it

### Feature or Bugfix
- Feature

### Purpose
- Currently it is impossible to exclude members of a single class when global exclude-members is defined. After change local exclude-members extends the list provided by global options.


sphinx/ext/autodoc/directive.py
```diff
@@ -84,7 +84,12 @@ def process_documenter_options(documenter: "Type[Documenter]", config: Config, o
         else:
             negated = options.pop('no-' + name, True) is None
             if name in config.autodoc_default_options and not negated:
-                options[name] = config.autodoc_default_options[name]
+                if name == "exclude-members":
```
tk0miya: I think all options taking a list of member names should be extended at runtime, not only `exclude-members`. For example,`private-members`, `undoc-members`, `special-members` and so on.
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_default_options

sphinx/ext/autodoc/directive.py
```diff
@@ -84,7 +84,12 @@ def process_documenter_options(documenter: "Type[Documenter]", config: Config, o
...
+                    if config.autodoc_default_options[name]:
+                        options[name] = config.autodoc_default_options[name] \
+                            + options.get(name, '')
```
tk0miya: It seems this simply concatenate to the `autodoc_default_options`. Is it working fine? I guess you need to use `','.join(...)` instead. I think adding a testcase will help this.

tk0miya (changes requested):
I'm still debating it would be better to add a way "not" to extend the global option too. How about a hint notation "+"?
```
.. automodule::
   :exclude-members: +foo,bar,baz
```

pbudzyns:
> I'm still debating it would be better to add a way "not" to extend the global option too. How about a hint notation "+"?
> 
> ```
> .. automodule::
>    :exclude-members: +foo,bar,baz
> ```

That is interesting suggestion, by "not" extending you mean to simply override it? Hint notation with "+" looks good to me. What about to make it "+/-" ?

tk0miya:
>That is interesting suggestion, by "not" extending you mean to simply override it? Hint notation with "+" looks good to me. What about to make it "+/-" ?

Yes. My thought is

* a plus sign like `+foo` means an enhancement; `foo` +  `autodoc_default_options`. It works as you proposed at first.
* Not prefixed; `foo` means an override; only `foo` is used (and `autodoc_default_options` is ignored).

About a minus sign, it's not late to write a code if somebody will request it.

sphinx/ext/autodoc/directive.py
pbudzyns: Added possibility to extend options `members`,`private-members`, `special-members` and `exclude-members`, other options have a priority over `autodoc_default_options`

shimizukawa:
+1 to @tk0miya 's idea.
I feel that introducing the `ｰ` sign would make it harder to understand what the current option is.

sphinx/ext/autodoc/directive.py
```diff
@@ -84,7 +87,19 @@ def process_documenter_options(documenter: "Type[Documenter]", config: Config, o
         else:
             negated = options.pop('no-' + name, True) is None
             if name in config.autodoc_default_options and not negated:
-                options[name] = config.autodoc_default_options[name]
+                if name in options:
+                    # take value from options if present or extend it
+                    # with autodoc_default_options if necessary
+                    if name in AUTODOC_EXTENDABLE_OPTIONS:
```
tk0miya, line 99: Please check `autodoc_default_options[name]` is string here. It can take boolean value or None also. When it equals to True or None, it matches all of members. So enhancement is meaningless.

sphinx/ext/autodoc/directive.py
```diff
@@ -84,7 +87,19 @@ def process_documenter_options(documenter: "Type[Documenter]", config: Config, o
...
+                        if options[name] is not None and options[name].startswith('+'):
+                            options[name] = ','.join([config.autodoc_default_options[name],
+                                                      options[name][1:]])
+                else:
+                    options[name] = config.autodoc_default_options[name]
+
+            elif isinstance(options.get(name), str) and options[name].startswith('+'):
```
tk0miya: You don't need to check the `options.get(name)` via isinstance because it's always a string if given. It's enough to check `options.get(name) is not None` only.

tk0miya (changes requested):
LGTM with nits for the code. Addition to this, could you update document for autodoc (doc/usage/extensions/autodoc.rst)?

tk0miya:
@shimizukawa Thank you for comment.

sphinx/ext/autodoc/directive.py
pbudzyns: That makes sense, so there were mistakes in already existing tests I think:
https://github.com/sphinx-doc/sphinx/blob/3.x/tests/test_ext_autodoc.py#L940
https://github.com/sphinx-doc/sphinx/blob/3.x/tests/test_ext_autodoc.py#L868
etc. 

I replaced these `True`'s with `None`.

tk0miya (approved):

tk0miya:
Merged. Thank you for your great contribution!

PR #8299, opened 2020-10-06
upgrade to the new pocoo style guide

graingert:
Subject: upgrade to the new pocoo style guide
### Feature or Bugfix
- Refactoring

### Purpose
- recently the pocoo styleguide moved to a machine readable format based on pre-commit
- this PR upgrades to this new styleguide

### Detail
- https://github.com/pallets/flask/blob/master/.pre-commit-config.yaml

### Relates
- https://groups.google.com/g/sphinx-dev/c/1k58cKn6Vpc


graingert:
this is a *very* big change, and drops support for python 2.7 and 3.5 and so might need more consideration of which parts of the pocoo styleguide is desired.

shimizukawa:
IMO...

* Sphinx is still supporting Python 3.5. So dropping Python-3.5 is not good.
* Sphinx has been away from the pocoo development team in 2012. Therefore, I don't think it is necessary to follow the new machine-readable pocoo (rather than flask) coding rules.
* However, I have no objection to the adoption of the BLACK.

tk0miya:
-0: Personally, I just can't come to like the black utility. I can accept if all other maintainers will accept it. On the other hand, other tooling is good to me (As shimizukawa told, dropping py35 is not good to us).

jakobandersen:
-1: From a brief check of the files I usually touch I find that a lot of the reformatting makes the code less readable and sometimes confusing.

Note also that merging these changes basically will guarantee unnecessary and annoying merge conflicts with other branches, e.g., PRs, due to how much code it touches.

tk0miya:
It seems Sphinx maintainers team does not prefer the new pocoo style. So I'm closing this.
Thank you for your proposal!

PR #8303, opened 2020-10-07
TST: Add test

larsoner:
It looks like the fix (plus some other stuff) from #8053 made it into #8265, but not the test. Is it worth adding?


tk0miya:
Thanks!

PR #8304, opened 2020-10-08
Register public markers in sphinx.testing.fixtures

hoefling:
Signed-off-by: oleg.hoefling <oleg.hoefling@gmail.com>

Subject: Register public markers in sphinx.testing.fixtures

### Feature or Bugfix
- Feature

### Purpose
When writing tests for a Sphinx extension, the usage of `sphinx` or `test_params` markers raises a `PytestUnknownMarkWarning`:

```
  test_spam.py:0: PytestUnknownMarkWarning: Unknown pytest.mark.sphinx - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.sphinx("html", testroot="spam")
```
The workaround is to register the markers in `pytest.ini`:

```ini
[pytest]
markers =
  sphinx
  test_params
```

To avoid this and offer the markers automatically, this PR moves registration of `sphinx` and `test_params` markers to `sphinx.testing.fixtures` plugin. Only two markers are exposed by the public plugin, so the remaining markers (`apidoc` and `setup_command`) are left in `setup.cfg`.

### DIFF
--- a/sphinx/testing/fixtures.py
+++ b/sphinx/testing/fixtures.py
@@ -22,7 +22,7 @@ from sphinx.testing import util
 from sphinx.testing.util import SphinxTestApp, SphinxTestAppWrapperForSkipBuilding
 
 
-PUBLIC_MARKERS = [
+DEFAULT_MARKERS = [
     (
         'sphinx(builder, testroot=None, freshenv=False, confoverrides=None, tags=None,'
         ' docutilsconf=None, parallel=0): arguments to initialize the sphinx test application.'
@@ -33,7 +33,7 @@ PUBLIC_MARKERS = [
 
 def pytest_configure(config):
     # register custom markers
-    for marker in PUBLIC_MARKERS:
+    for marker in DEFAULT_MARKERS:
         config.addinivalue_line('markers', marker)
 
 
--- a/sphinx/testing/fixtures.py
+++ b/sphinx/testing/fixtures.py
@@ -22,7 +22,7 @@ from sphinx.testing import util
 from sphinx.testing.util import SphinxTestApp, SphinxTestAppWrapperForSkipBuilding
 
 
-markers = [
+PUBLIC_MARKERS = [
     (
         'sphinx(builder, testroot=None, freshenv=False, confoverrides=None, tags=None,'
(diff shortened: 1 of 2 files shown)


hoefling:
@tk0miya can you aid me in resolving the mysterious test failures? The changes in this PR are minimal and shouldn't affect the tests, also I can't reproduce them locally with a 3.10 alpha built from source.

tk0miya:
Note: I found this technique on pytest document: https://docs.pytest.org/en/stable/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs

sphinx/testing/fixtures.py
```diff
@@ -21,6 +21,21 @@
 from sphinx.testing.util import SphinxTestApp, SphinxTestAppWrapperForSkipBuilding
 
 
+markers = [
```
tk0miya: How about renaming this like `DEFAULT_MARKERS` or something similar? I feel it would be better to use a CONSTANT name and good adjective for this.

tk0miya (changes requested):
Sorry for late response. Looks good. But it would be better to release this earlier. So could you rebase this into 3.x branch? Then it will be released in nearly months (3.3 will be shipped within a week, maybe).

About the CI errors, they have been fixed now. Sorry for confusing. That will be also fixed on rebasing.

sphinx/testing/fixtures.py
hoefling: @tk0miya maybe `PUBLIC_MARKERS`?

hoefling: @tk0miya however, maybe `PUBLIC_MARKERS` is unsuitable since it suggests some "private" markers also exist somewhere hidden.

tk0miya: `DEFAULT_INSTALLED_MARKERS` or `DEFAULT_ENABLED_MARKERS`?

hoefling: Changed to `DEFAULT_ENABLED_MARKERS`

tk0miya (approved):

tk0miya:
Thank you for your work! Merged :-)

PR #8307, opened 2020-10-11
Linkcheck: Use Thread daemon argument

francoisfreitag:
Instead of using a separate call.

Argument introduced in Python 3.3. https://docs.python.org/3.5/library/threading.html#threading.Thread

### Feature or Bugfix
- Refactoring


tk0miya (approved):
Reasonable!

Note: https://docs.python.org/3/library/threading.html#threading.Thread

Issue #8321, opened 2020-10-19
Linkcheck is failing on phone numbers


Daltz333:
```
`+1 (866) 511-6285 <tel:+1-866-511-6285>`__
```

Linking to numbers like an external link is causing linkcheck to fail as of 3.2.1

PR #8323, opened 2020-10-20
Linkcheck: Fix broken url not reporting error

TheTripleV:
Subject: Linkcheck: Fix broken url not reporting error

### Feature or Bugfix
- Bugfix

### Purpose
Some links are printed as broken but do not error out the build.
This issue appeared when we include `tel:` links in our build.

### Relates
- #8321

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -226,6 +226,7 @@ class CheckExternalLinksBuilder(Builder):
                             if rex.match(uri):
                                 return 'ignored', '', 0
                         else:
+                            self.broken[uri] = ''
                             return 'broken', '', 0
             elif uri in self.good:
                 return 'working', 'old', 0


tk0miya:
I think `tel:` hyperlink should not be treated as an error in linkcheck. But it seems your PR treated them as errors. Is this expected?

tk0miya:
Note: I just made a PR https://github.com/sphinx-doc/sphinx/pull/8332 to make `tel:` links ignored.

TheTripleV:
This PR only addresses the issue where unknown schemas are reported as broken by linkcheck but do not fail the sphinx build.

Here is a CI run showing the issue:

1. linkcheck does not recognize the `tel:` schema and prints out `(broken)`
https://github.com/wpilibsuite/frc-docs/runs/1276228243#step:6:601

2. The build still succeeds (even though it shouldn't)
https://github.com/wpilibsuite/frc-docs/runs/1276228243#step:6:752

tk0miya:
Okay, I understand. It would be better to merge this to handle an error by unknown schema. Additionally, it would be better to suppress the error of `tel` schema.

tk0miya:

Issue #8055, opened 2020-08-04
A potential display bug with the LaTeX generation step in Sphinx


schoi32:
**Describe the bug**
Sphinx-generated PDF file with an Index section having long lines overlapping each other 

**To Reproduce**
```
<Paste your command-line here which cause the problem>

$ conda create --name qmcpy python=3.6
$ conda activate qmcpy
$ git clone https://github.com/QMCSoftware/QMCSoftware.git
$ cd QMCSoftware
$ git checkout develop
$ pip install -r requirements/dev.txt
$ pip install -e ./
$ make qrng
$ pip install -r requirements/dev_docs.txt
$ make doc_pdf
```

**Expected behavior**
Remove overlap or make one-column Index pages in the last few pages in the PDF (URL below)
	

**Your project**
https://readthedocs.org/projects/qmcpy/

**Screenshots**
Please see the last few pages of the PDF at https://qmcpy.readthedocs.io/_/downloads/en/develop/pdf

**Environment info**
- OS:   Linux/Mac
- Python version: [e.g. 3.6]
- Sphinx version: <2
- Sphinx extensions:   sphinx.ext.autodoc, recommonmark
- Extra tools: gcc, pdflatex

**Additional context**

- https://qmcpy.readthedocs.io/_/downloads/en/develop/pdf/

tk0miya:
Screenshots:
<img width="817" alt="スクリーンショット 2020-08-05 20 31 27" src="https://user-images.githubusercontent.com/748828/89407986-aac09e00-d75a-11ea-85a7-0049292c05e5.png">

rscohn2:
You may be able to avoid this problem by making the index text smaller. See printindex in the docs: https://www.sphinx-doc.org/en/master/latex.html for directions. For especially long text, you may want to make your index single column. See printindex in this unmerged PR #8328

schoi32:
Many thanks, Robert.

I have worked in the fix to make the Index one column: https://github.com/QMCSoftware/QMCSoftware/commit/061624a9f72100a62185f0563382838f9db54aa9

This is the associated PDF:
https://qmcpy.readthedocs.io/_/downloads/en/develop/pdf/ <https://qmcpy.readthedocs.io/_/downloads/en/develop/pdf/>

Sou-Cheng Choi
Research Associate Professor
Department of Applied Mathematics
Illinois Institute of Technology, USA

rscohn2:
Thanks for sharing. That is a better solution. I will use it.

[...]

PR #8328, opened 2020-10-23
Show how to do single column index in latex

rscohn2:
Subject: Show how to do single column index in latex

### Feature or Bugfix
- Feature

### DIFF
--- a/doc/latex.rst
+++ b/doc/latex.rst
@@ -495,7 +495,9 @@ Keys that don't need to be overridden unless in special cases are:
    "printindex" call, the last thing in the file. Override if you want to
    generate the index differently or append some content after the index. For
    example ``'\\footnotesize\\raggedright\\printindex'`` is advisable when the
-   index is full of long entries.
+   index is full of long entries and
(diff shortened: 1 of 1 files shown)


doc/latex.rst
```diff
@@ -495,7 +495,9 @@ Keys that don't need to be overridden unless in special cases are:
    "printindex" call, the last thing in the file. Override if you want to
    generate the index differently or append some content after the index. For
    example ``'\\footnotesize\\raggedright\\printindex'`` is advisable when the
-   index is full of long entries.
+   index is full of long entries and
+   ``'\\let\\oldtwocolumn\\twocolumn\\renewcommand{\\twocolumn}[1][]{#1}\\printindex\\renewcommand{\\twocolumn}[1][]{\\oldtwocolumn}'``
```
tk0miya, line 499: This would be able to work. But it is a workaround, not a good solution. So I hesitate to introduce such a technique.
@jfbu Could you tell me your opinion?

Note: https://stackoverflow.com/questions/2389689/single-column-long-index-pages-in-latex

rscohn2:
Closing this PR without merging, there is a better solution here: #8055

Issue #8254, opened 2020-09-29
html output: Line numbers misalign with code lines (since sphinx 3.1.0, theme independent)


jonascj:
**Describe the bug**
In Sphinx 3.1.0 and later versions the line numbers of code blocks misalign with the code lines they number, as shown in the screenshots below. Version 3.0.4 aligns them properly like older versions 2.x and 1.x. Tested for both the alabaster and classic theme.

**To Reproduce**
```
pip install sphinx==3.1.0

sphinx-quickstart

cat >> index.rst << EOL

.. code-block:: python
    :linenos:

    import math
    for x in range(0,10):
        print(math.sqrt(x))
EOL

make html

firefox _build/html/index.html

```

**Expected behavior**
The line numbers and code lines should align as in the second screenshot (from sphinx v. 3.0.4).

**Your project**
A MWE is attached (``make html`` with sphinx v. 3.1.0 to see the problem)

[lineno-misalign-bug.zip](https://github.com/sphinx-doc/sphinx/files/5298739/lineno-misalign-bug.zip)

**Screenshots**
![scrot_2020-09-29_133401_746x561](https://user-images.githubusercontent.com/3301843/94553906-83b5b280-0259-11eb-90b7-3ea48ba7632f.png)

![scrot_2020-09-29_132448_753x597](https://user-images.githubusercontent.com/3301843/94553908-83b5b280-0259-11eb-9f0e-e3ac30c662cf.png)

**Environment info**
- OS: Linux
- Python version: 3.7.4
- Sphinx version: 3.0.4, 3.1.0, 3.2.1
- Extra tools: a browser to view html output

jonascj:
It basically appears to be 7px of top and bottom padding which is missing from the ``<pre>`` tag which holds the line numbers, or another vertical centering mechanism (present in 3.0.4 but not in 3.1.0).

The below screenshots hint that the ``<pre>``-element containing the line numbers is not equal in height to the ``<pre>`` containing the code. The height of the two ``<td>``-elements (containing line numbers and code respectively) have equal heights, so the line number's ``<pre>``-element is somehow vertical centered within ``<td class="linenos">``-element. That vertical centering mechanism/rule is present in 3.0.4 but missing in 3.1.0 as far as I can see...


![scrot_2020-09-29_225630_1598x878](https://user-images.githubusercontent.com/3301843/94615763-973c3a00-02a8-11eb-888d-b2c52235355c.png)

[...]

PR #8333, opened 2020-10-24
Override Pygments CSS with theme CSS

mgeier:
Before this PR, the Pygments colors for line numbers have overridden the colors selected by Sphinx.

This problem has only become visible with the upgrade to Pygments 2.7.0.

This PR fixes the color problem visible in #8254, but not the alignment problem.

The alignment problem has been fixed by https://github.com/pygments/pygments/pull/1583, which will become available soon in Pygments 2.7.2 [UPDATE: this has already been released in the meantime!].

### DIFF
--- a/sphinx/themes/basic/layout.html
+++ b/sphinx/themes/basic/layout.html
@@ -95,8 +95,8 @@
 {%- endmacro %}
 
 {%- macro css() %}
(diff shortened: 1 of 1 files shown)


tk0miya:
Thanks!

PR #8338, opened 2020-10-24
autodoc: check whether __slots__ is a str

### LINKED ISSUE #8294: autodoc: single-string __slots__ is not handled correctly
When documenting a Python module like this (using the `sphinx.ext.autodoc` extension):

```rst
.. automodule:: mymodule
   :members:
   :undoc-members:
```

... and `mymodule.py` has a class with a `__slots__` attribute that consists of a single string like this:

```python
class MyClassWithSlots:

    __slots__ = 'abc'
```
... the generated class documentation contains each of the characters in addition to the whole string:

image

Please note that [the Python docs](https://docs.python.org/3/reference/datamodel.html#object.__slots__) explicitly allow a single string (and not only iterables):

> This class variable can be assigned a string, iterable, or sequence of strings with variable names used by instances.

When using a tuple or a list of string, everything is fine:

**Expected behavior**

image

**Environment info**
- Sphinx version: any version, including `3.x` and `master` branch

mgeier:
Closes #8294.

### DIFF
--- a/sphinx/ext/autodoc/importer.py
+++ b/sphinx/ext/autodoc/importer.py
@@ -206,7 +206,10 @@ def get_object_members(subject: Any, objpath: List[str], attrgetter: Callable,
     if isclass(subject) and getattr(subject, '__slots__', None) is not None:
         from sphinx.ext.autodoc import SLOTSATTR
 
-        for name in subject.__slots__:
+        slots = subject.__slots__
+        if isinstance(slots, str):
+            slots = [slots]
+        for name in slots:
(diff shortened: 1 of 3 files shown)


tk0miya (changes requested):
LGTM. I'll merge this soon if you'll add a testcase for this.

mgeier:
Done: 82c83ed478618c9b97f1cbf7a24214ee433fa891

tk0miya:
Thank you for your quick work. Merged!

PR #8364, opened 2020-11-03
domains: c: fix use of unitialized variable

utzig:
`_fill_empty` was not updating the state with the value of line, so in recursive calls it would fail the assert checking for its own initialization.

/cc @jakobandersen


jakobandersen:
Thanks, merged via a rebase (5488bda082d3f712d27189dc88ccc46551220b22) to be part of 3.3.1.

(@tk0miya, as this fixes a crash it may be good to get 3.3.1 released as soon as it's otherwise ready)

tk0miya:
Okay, I'll ship it within a few days.

PR #8329, opened 2020-10-24
Fix mypy violations (with mypy-0.790)

tk0miya:
### Feature or Bugfix
- Refactoring

PR #8376, opened 2020-11-05
Test with py310

tk0miya:
### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
- Re-enable testing with python 3.10
- refs: https://github.com/sphinx-doc/sphinx/pull/8329/files#r511331489

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,11 +4,12 @@ on: [push, pull_request]
 
 jobs:
   ubuntu:
-    runs-on: ubuntu-16.04
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         name: [py35, py36, py37, py38, py39]
+        os: [ubuntu-16.04]
         include:
         - name: py35
           python: 3.5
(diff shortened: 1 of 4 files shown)


tests/test_ext_autodoc_configs.py
```diff
@@ -485,64 +485,124 @@ def test_autodoc_typehints_signature(app):
...
-        '   :module: target.typehints',
-        '',
-    ]
+    if sys.version_info < (3, 7):
+        assert list(actual) == [
+            '',
+            '.. py:module:: target.typehints',
+            '',
+            '',
+            '.. py:class:: Math(s: str, o: object = None)',
```
tk0miya: Sadly, py35 and py36 does not support `Optional[object]`. I don't know why we use `object` type annotation for this case. So it would be better to change the example itself.
```
$ python3.5
Python 3.5.10 (default, Oct 24 2020, 15:43:16)
[GCC 4.2.1 Compatible Apple LLVM 11.0.3 (clang-1103.0.32.59)] on darwin
>>> from typing import Optional
>>> Optional[int]
typing.Union[int, NoneType]
>>> Optional[object]
<class 'object'>
```

Issue #3008, opened 2016-10-04
(Low Prio)linkcheck to website with self-signed certificate


hhsprings:
#2988 will be fiexed except for website with self-signed certificate.

Though this case is not good, but I think it is not `broken`. My proposal is to categorize this case as `not secure`, etc.

tk0miya:
Memo: https://quimby.gnus.org/notes/BABYL

PR #8404, opened 2020-11-10
linkchecker: Remove dead code is_ssl_error()

francoisfreitag:
### Feature or Bugfix
- Refactoring

#### Verify invalid SSL handling

Test invalid SSL is reported as broken.

`linkcheck` logic suggests that SSL errors were originally expected to
be ignored.

Blaming the corresponding lines point to issue #3008: linkcheck to
website with self-signed certificates. That issue was fixed by commit
4c7bec6460ae6154705bfd023e363e458f16d1b6, which ignored SSL errors.
Probably because back then, users could not specify a CA bundle.

A broken SSL certificate is a real issue, it should not be ignored.
Users wishing to ignore issues for a specific link can use the
`linkcheck_ignore` option.

The current behavior is to report the site as broken, keep it.

#### Remove `is_ssl_error()`

That function is dead code, it always returns False. The exception
checking stopped working because the Requests library wraps SSL errors
in a `requests.exceptions.SSLError` and no longer throws an
`urllib3.exceptions.SSLError`. The first argument to that exception is
an `urllib3.exceptions.MaxRetryError`.

### Relates
- #3008

### DIFF
--- a/sphinx/util/requests.py
+++ b/sphinx/util/requests.py
@@ -18,6 +18,7 @@ import requests
 
 import sphinx
 from sphinx.config import Config
+from sphinx.deprecation import RemovedInSphinx50Warning
 
 try:
     from requests.packages.urllib3.exceptions import SSLError
@@ -43,6 +44,10 @@ useragent_header = [('User-Agent',
 
 def is_ssl_error(exc: Exception) -> bool:
     """Check an exception is SSLError."""
+    warnings.warn(
+        "is_ssl_error() is outdated and likely returns incorrect results "
+        "for modern versions of Requests.",
+        RemovedInSphinx50Warning)
     if isinstance(exc, SSLError):
         return True
     else:
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,7 @@ Incompatible changes
 Deprecated
 ----------
 
+* The ``is_ssl_error()`` function of ``sphinx.util.requests``
 * The ``follow_wrapped`` argument of ``sphinx.util.inspect.signature()``
 
 Features added
--- a/doc/extdev/deprecated.rst
+++ b/doc/extdev/deprecated.rst
@@ -27,6 +27,11 @@ The following is a list of deprecated interfaces.
      - Alternatives
 
 
+   * - ``sphinx.util.requests.is_ssl_error()``
+     - 3.4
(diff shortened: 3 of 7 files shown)


sphinx/util/requests.py
```diff
@@ -41,18 +35,6 @@
                      'Mozilla/5.0 (X11; Linux x86_64; rv:25.0) Gecko/20100101 Firefox/25.0')]
 
 
-def is_ssl_error(exc: Exception) -> bool:
```
tk0miya, line 44: To avoid the error of 3rd party extensions, we usually obey the deprecation policy: https://www.sphinx-doc.org/en/master/internals/release-process.html?highlight=deprecation#deprecation-policy.
So please don't remove this function, and add a deprecation warning to the top of the function instead.
46f0b3063d is an example of deprecation.

I know nobody accesses this function. So I agree this is a bit boring work. But Sphinx has many extensions. And some of them touches the inside of Sphinx to hack our software. So we need to change the interface of class and functions carefully.

tk0miya (changes requested):
LGTM with nits. Additionally, this conflicts with the HEAD. Could you check it please?

sphinx/util/requests.py
francoisfreitag: Thanks for the explanation and pointers, it was definitely helpful! I’ll keep the deprecation policy in mind for future patches.

I restored `is_ssl_error()` and added a `RemovedInSphinx50Warning`. I believe the warning message make clear that the entire function should be removed?

francoisfreitag:
Rebased on `3.x`.

sphinx/util/requests.py
tk0miya: Looks good. Additionally, please add an entry to CHANGES and docs/extdev/deprecated.rst. These lists are also useful for developers.

francoisfreitag: Added an entry to both documents. Thanks for the review!

francoisfreitag:
For consistency, I renamed the test root `test-linkcheck-https` to `test-linkcheck-localserver-https`.

tk0miya (approved):
LGTM!

tk0miya:
Thank you for your work :-)

PR #8407, opened 2020-11-11
linkcheck: set allow_redirects in requests.head() call

francoisfreitag:
### Feature or Bugfix
- Refactoring

### Purpose
Define the option closer to its use.
Following redirects is the default for other methods. https://requests.readthedocs.io/en/latest/api/#requests.request


francoisfreitag:
Rebased on `3.x`.

PR #8408, opened 2020-11-11
Sort imports with isort

francoisfreitag:
### Feature or Bugfix
- Refactoring

### Purpose

Keep imports alphabetically sorted and their order homogeneous across
Python source files.

The isort project has more feature and is more active than the
flake8-import-order plugin.

Most issues caught were simply import ordering from the same module.
Where imports were purposefully placed out of order, tag with
isort:skip.

### Detail
- https://timothycrosley.github.io/isort/


sphinx/application.py
```diff
@@ -18,7 +18,7 @@
 from collections import deque
 from io import StringIO
 from os import path
-from typing import Any, Callable, Dict, IO, List, Optional, Tuple, Union
+from typing import IO, Any, Callable, Dict, List, Optional, Tuple, Union
```
tk0miya, line 21: Do you know why isort puts `IO` before `Any`? I feel odd...

tk0miya: Okay, I understand. It seems isort considers `IO` is a constant.

tk0miya (approved):

PR #8411, opened 2020-11-11
Run tests with Python Development mode

francoisfreitag:
### Feature or Bugfix
- Refactoring

### Purpose

Helps catching more issues with the test suite.

### Detail

- https://docs.python.org/3/library/devmode.html#devmode


tk0miya (approved):
LGTM!

PR #8414, opened 2020-11-11
linkcheck: Remove requests mocks from tests

francoisfreitag:
### Feature or Bugfix
- Refactoring

### Purpose
Makes the test more realistic by issuing an HTTP request.
Reduces coupling between test and the code under test.


tk0miya:
LGTM :-)

PR #8418, opened 2020-11-13
domain: c: fix warning for intersphinx targets

utzig:
Referencing a symbol using <project:symbol>, which is a valid syntax that can be resolved by intersphinx, results in the C-domain parser printing a warning "Unparseable C cross-reference". The resolution still works correctly because it is postponed to intersphinx.

This commit adds a small check for intersphinx target format validity before printing the warning.

/cc @jakobandersen

### DIFF
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -3754,6 +3754,15 @@ class CDomain(Domain):
                     ourObjects[fullname] = (fn, id_, objtype)
                 # no need to warn on duplicates, the symbol merge already does that
 
+    def _has_intersphinx_prefix(self, env: BuildEnvironment, target: str) -> bool:
+        if 'intersphinx_mapping' not in env.config:
+            return False
+        parts = target.split(':')
+        if len(parts) != 2 or parts[1].startswith(':') or \
+                parts[0] not in env.config.intersphinx_mapping:
+            return False
+        return True
+
     def _resolve_xref_inner(self, env: BuildEnvironment, fromdocname: str, builder: Builder,
                             typ: str, target: str, node: pending_xref,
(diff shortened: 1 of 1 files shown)


jakobandersen:
Hmm, do you have a small example project where this helps?

utzig:
> Hmm, do you have a small example project where this helps?

I just built this little toy project which shows the issue. It has 3 projects, two of them, proj1 and proj2, export a function named "fn". In the "main" project I try to link to "fn". It will always link correctly to one of the "fn" functions, but there is no control over which. If I use "proj1:fn" or "proj2:fn" than it will link to the one I want, but will throw a "Unparseable C cross-reference: bla bla bla" warning on the way.

[inter_warn.tar.gz](https://github.com/sphinx-doc/sphinx/files/5542195/inter_warn.tar.gz)

utzig:
@jakobandersen Have you had a chance to look into this?

jakobandersen:
I find this way of making it work a bit iffy: when you write any kind of role in your input you are requesting a specific handler to be invoked, e.g., for ``:c:func:`` you request lookup in the documented C entities (with intersphinx as backup). But here you are explicitly requesting that the ordinary handler does not do any work.
I therefore suggest that intersphinx should provide a role, say ``inter`` which you can use for this purpose.
For example it could be made possible to write
```rst
:inter:`proj1,c:func,fn`
```
where the three comma-separated components are
1. the project specifier, as intersphinx understands it,
2. the original role, needed by intersphinx for matching,
3. the query symbol as if one would have given it to the original role from (2).

tk0miya:
FYI: Now I just posted a draft version of the `:intersphinx:` role in https://github.com/sphinx-doc/sphinx/pull/9062.

PR #8425, opened 2020-11-14
linkcheck: Prevent REQUESTS_CA_BUNDLE leak in tests

francoisfreitag:
### Feature or Bugfix
- Bugfix

### Purpose
Test `test_connect_to_selfsigned_nonexistent_cert_file` serves two
purposes:
- verify the behavior when the CA bundle file path is incorrect
- flag a leak of REQUESTS_CA_BUNDLE

Assumes that test runs after
`test_connect_to_selfsigned_with_requests_env_var`. Not great, but
better than no testing.

### DIFF
--- a/tests/test_build_linkcheck.py
+++ b/tests/test_build_linkcheck.py
@@ -10,7 +10,6 @@
 
 import http.server
 import json
-import os
 import re
 import textwrap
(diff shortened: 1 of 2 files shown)


tk0miya (approved):
Good!

Issue #8415, opened 2020-11-11
NewType and TypeVar don't resolve correctly


anselor:
**Describe the bug**
Autodoc generates warnings and doesn't resolve NewType or TypeVar correctly.

**To Reproduce**
```
$ git clone https://github.com/anselor/sphinx-issues-examples/
$ cd sphinx-issues-examples/typevar-namespace
$ python -m sphinx -b html docs/ doc-out/ -nvWT
```

**Expected behavior**
Documentation generated with TypeVar and NewType names that link to the documented declaration

**Your project**
https://github.com/anselor/sphinx-issues-examples/tree/main/typevar-namespace

**Screenshots**

**Environment info**
- OS: Ubuntu 18.04, 20.04
- Python version: 3.6.9, 3.8.2
- Sphinx version: 3.2.1
- Sphinx extensions:  sphinx.ext.autodoc
- Extra tools: 

**Additional context**

- [e.g. URL or Ticket]

anselor:
@tk0miya FYI - this is the new issue I created as you suggested.

tk0miya:
@anselor I can't reproduce any warnings from your example.

This is a Dockerfile to reproduce the warning:
```
FROM python:3.8-slim

RUN apt update; apt install -y build-essential curl git make unzip vim
RUN git clone https://github.com/anselor/sphinx-issues-examples/
RUN pip install Sphinx
WORKDIR /sphinx-issues-examples/typevar-namespace/src
RUN pip install -e .
WORKDIR /sphinx-issues-examples/typevar-namespace/docs
RUN sphinx-build . _build/html
```

And its result:
```
Running Sphinx v3.3.1
making output directory... done
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
reading sources... [100%] index

looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index

generating indices... genindex py-modindex done
writing additional pages... search done
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in _build/html.
```

anselor:
I was using 3.2.1 - just updated to sphinx 3.3.1 to make sure. The steps in your docker container do not match the same steps/options that I documented in the ticket. 

[...]

Issue #7119, opened 2020-02-09
Allow unqualified type hint names in generated docs


alkasm:
It would be great if the current type hinting work would support removing qualification from type names. For some modules, the types may be deeply nested which will more easily break the page layout.

### Definitions (by example)

* Qualified name: `pkg.module.ClassName`
* Unqualified name: `ClassName`

### Example

```python
from typing import Tuple
from numpy.polynomial import polynomial

def multiply_polynomial(
    p: polynomial.Polynomial, value: int
) -> Tuple[polynomial.Polynomial, polynomial.Polynomial]:
    return p, polynomial.Polynomial(*(value * p.coef))
```

In this code, the type is `polynomial.Polynomial` but the fully qualified name is `numpy.polynomial.polynomial.Polynomial` which is what will show up in the docs.

With the following in `conf.py`

```python
extensions = [
    "sphinx.ext.autodoc",
    "sphinx.ext.autodoc.typehints",
]
autodoc_typehints = "description"
```

We'd see some generated docs similar to

```
mypkg.mymodule.multiply_polynomial(p, value)
    Parameters:     * p (numpy.polynomial.polynomial.Polynomial) –
                    * value (int) –
    Return type:    Tuple[numpy.polynomial.polynomial.Polynomial, numpy.polynomial.polynomial.Polynomial]
```

Instead, with unqualified names, we'd have a much simpler

```
mypkg.mymodule.multiply_polynomial(p, value)
    Parameters:     * p (Polynomial) –
                    * value (int) –
    Return type:    Tuple[Polynomial, Polynomial]
```

### Rationale

Full qualification inherently gives more context, so it is a sane default. However, intersphinx linkage to the correct types gives *even better* context, since you can just follow directly to the actual type definition. With both intersphinx linkage and unqualified type names, you can have the best of both worlds: readability and quick disambiguation. 

### Similar prior work

The external extension https://github.com/agronholm/sphinx-autodoc-typehints collapses the name to the unqualified name by default, unless otherwise opted for, for an example of how this might work.

### Variant

[...]

PR #8426, opened 2020-11-15
autodoc: an imported TypeVar is not resolved (refs: #8415)

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- So far, a TypeVar is rendered without module name. As a result, it
could not be resolved if it is imported from other modules.  This
prepends its module name and make it resolvable.
- This is available only in ~Python 3.6 or above~ Python 3.7 or above.
- As a side effect, all of TypeVars are displayed with module name. It
should be fixed in latter step (refs: #7119)
- refs: #8415

### DIFF
--- a/tests/test_util_inspect.py
+++ b/tests/test_util_inspect.py
@@ -142,7 +142,13 @@ def test_signature_annotations():
 
     # TypeVars and generic types with TypeVars
     sig = inspect.signature(f2)
-    assert stringify_signature(sig) == '(x: List[T], y: List[T_co], z: T) -> List[T_contra]'
+    if sys.version_info < (3, 7):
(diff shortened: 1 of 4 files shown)


tests/test_util_inspect.py
```diff
@@ -142,7 +142,11 @@ def test_signature_annotations():
 
     # TypeVars and generic types with TypeVars
     sig = inspect.signature(f2)
-    assert stringify_signature(sig) == '(x: List[T], y: List[T_co], z: T) -> List[T_contra]'
+    if sys.version_info < (3, 6):
```
tk0miya: Oops, this restriction is also applicable for python 3.6.

PR #8431, opened 2020-11-15
tests: Always cleanup when exiting modify_env context manager

francoisfreitag:
### Feature or Bugfix
- Refactoring

Previously, an exception would leave the environment changed.


francoisfreitag:
The first version of the cleanup was incomplete.

tk0miya (approved):
Oops, I also overlooked it...

PR #8432, opened 2020-11-15
Fix #7119: Show type hint names unqualified when resolving succeeded

### LINKED ISSUE #7119: Allow unqualified type hint names in generated docs

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- Add pending_xref_condition node

    To choose appropriate content for pending_xref node on resolving,
    this introduces a new custom node `pending_xref_condition`.  It only
    has a condition for the filtering and contents of the reference.

-    Add a new config variable: python_smart_reference.  If enabled, it goes
    to suppress the module name of the python reference if it can be
     resolved.


doc/usage/configuration.rst
```diff
@@ -2615,6 +2615,17 @@ Options for the C++ domain
 
    .. versionadded:: 1.5
 
+Options for the Python domain
+-----------------------------
+
+.. confval:: python_smart_reference
```
tk0miya: I need a better name for this.

sphinx/addnodes.py
```diff
@@ -342,6 +342,51 @@ class pending_xref(nodes.Inline, nodes.Element):
     """
 
 
+class pending_xref_condition(nodes.Inline, nodes.TextElement):
```
tk0miya, line 341: @shimizukawa @jakobandersen What do you think about this new node?

sphinx/addnodes.py
```diff
@@ -342,6 +342,51 @@ class pending_xref(nodes.Inline, nodes.Element):
...
+
+    .. note:: This node is only allowed to be placed under pending_xref node.
+              It is not allows to place it under other nodes.  In addition,
+              pending_xref node must contain only pending_xref_condition
+              nodes if it contains one or more pending_xref_condition nodes.
+
+    The pending_xref_condition node should have **condition** attribute.
+    Domains can be store their individual conditions into the attribute to
+    filter contents on resolving phase.  As a reserved condition name,
+    ``condition="*"`` is used for the fallback of resolution failure.
```
tk0miya, line 380: Now I'm thinking about how intersphinx treats this. I have two idea: 1) adding another reserved condition `"resolved"` as a common hint for extensions, 2) add a filtering method to the domain class.

sphinx/domains/python.py
```diff
@@ -91,7 +91,14 @@ def type_to_xref(text: str, env: BuildEnvironment = None) -> addnodes.pending_xr
     else:
         kwargs = {}
 
-    return pending_xref('', nodes.Text(text),
+    if env.config.python_smart_reference:
+        shortname = text.split('.')[-1]
```
tk0miya, line 99: This is not good way for a nested class. But no way to get `__qualname__` because python domain should not import the python module dynamically.

sphinx/addnodes.py
```diff
@@ -342,6 +342,51 @@ class pending_xref(nodes.Inline, nodes.Element):
     """
 
 
+class pending_xref_condition(nodes.Inline, nodes.TextElement):
+    """Node for cross-references that are used to choose appropriate
+    content of the reference by conditions on the resolving phase.
+
+    When the :py:class:`pending_xref` node contains one or more
+    **pending_xref_conditions** nodes, the cross-reference resolver
```
shimizukawa:
```suggestion
    **pending_xref_condition** nodes, the cross-reference resolver
```

sphinx/addnodes.py
```diff
@@ -342,6 +342,51 @@ class pending_xref(nodes.Inline, nodes.Element):
...
+    should choose the content of the reference using defined conditions
+    in ``condition`` attribute of each pending_xref_conditions node::
```
shimizukawa:
```suggestion
    in ``condition`` attribute of each pending_xref_condition nodes::
```

doc/usage/configuration.rst
shimizukawa: How about a `python_use_unqualified_type_names` ?

sphinx/addnodes.py
shimizukawa: I discussed this with @tk0miya at the Sphinx Hack-a-thon event.

* We can use any name for the condition, but there are reserved and recommended names.
* The 'resolved' is the recommended condition. The reason for the recommendation is that it will be selected if it matches in intersphinx.
* '*' is a reserved condition. It will be selected if no reference is found in the domain or intersphinx.

So, for now, we should choose (1) for this feature.

shimizukawa (commented):
reviewed.

doc/usage/configuration.rst
tk0miya: Good idea!

sphinx/addnodes.py
tk0miya: updated.

sphinx/addnodes.py
tk0miya: @jakobandersen Please let me know your opinion on this new node. I don't know this is worthy for C/C++ domains. But I  believe this will not conflict with them.

sphinx/domains/python.py
```diff
@@ -91,7 +91,14 @@ def type_to_xref(text: str, env: BuildEnvironment = None) -> addnodes.pending_xr
     else:
         kwargs = {}
 
-    return pending_xref('', nodes.Text(text),
+    if env.config.python_use_unqualified_type_names:
```
tk0miya, line 95: It would be better to use this option on `PyField.make_xref()`. I'll do it the next PR.

tk0miya:
It's time to merge!

PR #8444, opened 2020-11-18
Fix #8443: autodata directive does not support PEP-526 ivars

### LINKED ISSUE #8443: autodata could not create document for instance variables with type annotation
**Describe the bug**
autodata could not create document for instance variables with type annotation.

**To Reproduce**

```
# example.py
var: int
```
```
# index.rst
.. autodata:: example.var
```

It causes AttributeError
```
WARNING: autodoc: failed to import data 'var' from module 'example'; the following exception was raised:
  File "/Users/tkomiya/work/sphinx/sphinx/util/inspect.py", line 343, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'example' has no attribute 'var'

The above exception was the direct cause of the following exception:

  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/importer.py", line 106, in import_object
    obj = attrgetter(obj, mangled_name)
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 316, in get_attr
    return autodoc_attrgetter(self.env.app, obj, name, *defargs)
  File "/Users/tkomiya/work/sphinx/sphinx/ext/autodoc/__init__.py", line 2269, in autodoc_attrgetter
    return safe_getattr(obj, name, *defargs)
  File "/Users/tkomiya/work/sphinx/sphinx/util/inspect.py", line 359, in safe_getattr
    raise AttributeError(name) from exc
AttributeError: var
```

**Expected behavior**
Document should be generated successfully like when I used `automodule`.

**Your project**
N/A

**Screenshots**
N/A

**Environment info**
- OS: Mac
- Python version: 3.9.0
- Sphinx version: HEAD of 3.x
- Extra tools: Nothing

**Additional context**
Nothing

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #8443

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -2045,6 +2045,22 @@ class AttributeDocumenter(DocstringStripSignatureMixin, ClassLevelDocumenter):  
 
     def isinstanceattribute(self) -> bool:
         """Check the subject is an instance attribute."""
+        # uninitialized instance variable (PEP-526)
+        with mock(self.config.autodoc_mock_imports):
+            try:
+                ret = import_object(self.modname, self.objpath[:-1], 'class',
+                                    attrgetter=self.get_attr,
+                                    warningiserror=self.config.autodoc_warningiserror)
+                self.parent = ret[3]
+                annotations = get_type_hints(self.parent, None,
+                                             self.config.autodoc_type_aliases)
+                if self.objpath[-1] in annotations:
+                    self.object = UNINITIALIZED_ATTR
+                    return True
+            except ImportError:
+                pass
+
+        # An instance variable defined inside __init__().
         try:
             analyzer = ModuleAnalyzer.for_module(self.modname)
             attr_docs = analyzer.find_attr_docs()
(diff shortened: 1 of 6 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -1697,6 +1697,29 @@ def can_document_member(cls, member: Any, membername: str, isattr: bool, parent:
                             ) -> bool:
         return isinstance(parent, ModuleDocumenter) and isattr
 
+    def import_object(self, raiseerror: bool = False) -> bool:
+        """Never import anything."""
```
tk0miya: This comment is not correct...

sphinx/ext/autodoc/__init__.py
```diff
@@ -1733,6 +1756,14 @@ def get_real_modname(self) -> str:
         return self.get_attr(self.parent or self.object, '__module__', None) \
             or self.modname
 
+    def add_content(self, more_content: Any, no_docstring: bool = False) -> None:
+        """Never try to get a docstring from the object."""
```
tk0miya: This is also.

PR #8445, opened 2020-11-18
Fix #8419: html search: Do not load language_data.js in non-search pages

### LINKED ISSUE #8419: language_data.js is loaded from all pages.
**Describe the bug**

`language_data.js` is used by `searchtools.js`.

But `language_data.js` is loaded in all pages.
https://github.com/sphinx-doc/sphinx/blob/68cc0f7e94f360a2c62ebcb761f8096e04ebf07f/sphinx/builders/html/__init__.py#L304

Can we move it to `search.html`?

https://github.com/sphinx-doc/sphinx/blob/af62fa61e6cbd88d0798963211e73e5ba0d55e6d/sphinx/themes/basic/search.html#L14

**To Reproduce**

See this page and show source.
https://docs.python.org/3/library/stdtypes.html

**Expected behavior**

If language_data.js is used only by searchtool.js, don't load language_data.js unless searchtool.js is loaded.

**Your project**
python/cpython project, Docs/

**Environment info**
- Sphinx version: 2.4.4.

**Additional context**

- https://github.com/python/python-docs-theme/issues/58

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #8419 
- The ``language_data.js`` is only used on search page.  But it is always
loaded meaninglessly.  This fixes not to load it on all non-search pages.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -42,6 +42,7 @@ Bugs fixed
   is decorated
 * #8434: autodoc: :confval:`autodoc_type_aliases` does not effect to variables
   and attributes
+* #8419: html search: Do not load ``language_data.js`` in non-search pages
 
 Testing
 --------
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -301,7 +301,6 @@ class StandaloneHTMLBuilder(Builder):
         self.add_js_file('jquery.js')
         self.add_js_file('underscore.js')
         self.add_js_file('doctools.js')
-        self.add_js_file('language_data.js')
 
         for filename, attrs in self.app.registry.js_files:
             self.add_js_file(filename, **attrs)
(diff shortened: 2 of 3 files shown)


zzzeek:
hey there.     Wondering what a more long term solution would be for maintainers of custom themes to not constantly get hit with these after the fact.     You can see here three custom themes have already been hit.

Looking in Sphinx changelog I can see this issue is buried in a long list of other issues under "bugs fixed" with a terse message "html search: Do not load ``language_data.js`` in non-search pages":

https://github.com/sphinx-doc/sphinx/blob/3.x/CHANGES#L183

there appears to be a more prominent section called "Incompatible changes" with just one API change in it.   Is it possible that fully breaking changes like these appear under "incompatible changes" and perhaps include some more detail what they mean ?     thanks for reading.

PR #8449, opened 2020-11-18
Closes #5603: autodoc will add :canonical: to documented objects

### LINKED ISSUE #5603: Help people using Intersphinx against objects imported from a package __init__.py

### Problem

```python
import marshmallow

    pass
```

```
```

### Work around

```python
intersphinx_aliases = {
}


        try:
            try:
            except KeyError:
                continue
        except KeyError:
            continue


def setup(app):
```

### Varia

cblegare:
Subject: make autodoc apply the ``:canonical: `` option to relevant documented objects

### Feature or Bugfix

- Feature

### Purpose

- Autodoc will add the ``:canonical:`` option with the aliased full real name of the documented object.

### Detail

- The ``:canonical:`` option is only added when the real module name is different from the use module name
- The option is not added to imported objects, unless they are specified in ``__all__``, accordingly with the current behavior of members configurations.

### Relates
- #5603
- supersedes #8448 as it was not based on the right branch


cblegare:
The failing test should be straightforward to fix (later today).  I'll rebase the PR then.

CHANGES
```diff
@@ -33,7 +33,8 @@ Features added
 --------------
 
 * #4826: py domain: Add ``:canonical:`` option to python directives to describe
-  the location where the object is defined
+  the location where the object is defined.  This option is automatically added
+  by :mod:`sphinx.ext.autodoc` when relevant.
```
tk0miya: This is an entry for the python domain, not for autodoc. Please add another entry. I think it would be better to describe how the autodoc extension is improved (not how implemented).

doc/usage/extensions/autodoc.rst
```diff
@@ -299,6 +299,15 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
 
         .. autoclass:: module.name::Noodle
 
+   * For objects defined in multiple modules, autodoc will generate both
+     specific directives and use the ``:canonical:`` option accordingly.
+     This applies when importing a class from a submodule within a package's
+     __init__.py and adding it to the ``__all__`` attribute.
```
tk0miya: I think adding to `__all__` is one of the use cases. It's not a requirement. I worry about this explanation confuses users.

doc/usage/extensions/autodoc.rst
```diff
@@ -299,6 +299,15 @@ inserting them into the page source under a suitable :rst:dir:`py:module`,
 
         .. autoclass:: module.name::Noodle
 
+   * For objects defined in multiple modules, autodoc will generate both
+     specific directives and use the ``:canonical:`` option accordingly.
```
tk0miya: It seems this describes the internal of autodoc itself. Does this explanation help users?

sphinx/ext/autodoc/__init__.py
```diff
@@ -530,6 +530,11 @@ def add_directive_header(self, sig: str) -> None:
             # Be explicit about the module, this is necessary since .. class::
             # etc. don't support a prepended module name
             self.add_line('   :module: %s' % self.modname, sourcename)
+        if self.real_modname != self.modname:
+            # Where the object is actually defined when there is a difference
+            # between the imported modname and the real modname.
+            self.add_line('   :canonical: %s.%s' % (self.real_modname, self.object_name),
```
tk0miya: This does not support `import ... as ...` style imports. We need to obtain the real name of the subject

doc/usage/extensions/autodoc.rst
cblegare: I'll try to reword this.  The idea was to give a example (which cover what seamed to be the most common use case).  This feature effectively adds one entry in the generated documentation, which is quite a change.

sphinx/ext/autodoc/__init__.py
cblegare: Thanks for this.  This line was already different in the upcoming commit to behave properly for non top-level objects (like classes within a class).

I'll be really interested in your feedback when this is pushed.

cblegare:
@tk0miya Since I feel that by sphinx 4.0 the canonical name might be useful, I've added it at the `Documenter` properties directly.

Also, the `canonical` is only added when required.  Do you think we should provide a config key to force autodoc to always add it to the directives headers?

I removed all references to the `__all__` attribute since this PR actualy do not change anything about how Sphinx behaves against it.

Also, I still need to fix a few tests, so I'm keeping this PR as a draft for now.

doc/usage/restructuredtext/domains.rst
```diff
@@ -126,6 +126,8 @@ In short:
   refer to ``Queue.Queue.get`` but only display ``get`` as the link text.
 
 
+.. _python-domain:
```
tk0miya: nits: This anchor is no longer needed now.

tests/test_ext_autodoc_autofunction.py
```diff
@@ -79,10 +79,15 @@ def test_method(app):
...
+    actual = list(actual)
+    assert actual[:3] == [
         '',
         '.. py:function:: umask(mask, /)',
         '   :module: os',
+    ]
+    assert actual[3].startswith("   :canonical: ")
+    assert actual[3].endswith(".umask")
+    assert actual[-3:] == [
         '',
```
tk0miya: How about this? (of course, importing `os.umask` is needed)
```suggestion
    assert list(actual) == [
        '',
        '.. py:function:: umask(mask, /)',
        '   :module: os',
        '   :canonical: %s.umask' % os.umask.__module__,
        '',
```

sphinx/ext/autodoc/__init__.py
```diff
@@ -400,7 +400,7 @@ def parse_name(self) -> bool:
         self.args = args
         self.retann = retann
         self.fullname = (self.modname or '') + \
-                        ('.' + '.'.join(self.objpath) if self.objpath else '')
+                        ('::' + '.'.join(self.objpath) if self.objpath else '')
```
tk0miya: Is this change needed? I can't imagine the side effect of this change. I'm afraid that this change causes errors in 3rd party extensions.

sphinx/ext/autodoc/__init__.py
```diff
@@ -530,6 +530,9 @@ def add_directive_header(self, sig: str) -> None:
             # Be explicit about the module, this is necessary since .. class::
             # etc. don't support a prepended module name
             self.add_line('   :module: %s' % self.modname, sourcename)
+        if self.options.canonical or self.fullname != self.canonical_fullname:
```
tk0miya: `:canonical:` option is not defined in auto* directives. So we can't use `self.options.canonical` here.

sphinx/ext/autodoc/__init__.py
```diff
@@ -899,6 +902,15 @@ def generate(self, more_content: Any = None, real_modname: str = None,
             except PycodeError:
                 pass
 
+        try:
+            subject = inspect.unpartial(self.object)
+            self.canonical_objpath = subject.__qualname__.split(".")
```
tk0miya: It seems nobody uses `canonical_objpath`. Is this needed?

sphinx/ext/autodoc/__init__.py
```diff
@@ -899,6 +902,15 @@ def generate(self, more_content: Any = None, real_modname: str = None,
...
+            self.canonical_fullname = '%s::%s' % (self.real_modname, subject.__qualname__)
```
tk0miya: Could you add an initialization of this member to `__init__()` and fill it in `parse_name()`? It would be better to keep this method simple as possible.

tk0miya: Could you use `safe_getattr` instead of accessing `__qualname__` attribute directly? Users sometimes pass a strange object to autodoc. So we have to use it to access attributes.

tk0miya:
Why did you add `tests/roots/__init__.py`? The `tests/roots` is not a python package. So it's strange to me.

tests/test_ext_autodoc_autofunction.py
cblegare: Clean and flexible, I like it!

sphinx/ext/autodoc/__init__.py
cblegare: Removed it.  There is indeed no formal notion of "fully qualified object name" in python.

sphinx/ext/autodoc/__init__.py
cblegare: Artifact of a local exploration, sorry about that

sphinx/ext/autodoc/__init__.py
cblegare: I used a property.  Using a property feels safer than hoping for users to have `parse_name` used first.  Actually, I'm wondering why there is so little use of properties for lazy loaded attributes

cblegare:
> Why did you add `tests/roots/__init__.py`? The `tests/roots` is not a python package. So it's strange to me.

This new file got through without me noticing. Sorry.

---

I still need to find a way around partials (`functools.partial`, `functools.partialmethod`).

Their `__qualname__` is a local function within functools (not useful and unexpected from the user.

Pointing the canonical name to the wrapped function feels unhelpful

I'll be looking for a way to detect partials and give up canonical name resolution in this case.

---

This PR has proven trickier than I expected, but still fun!

sphinx/ext/autodoc/__init__.py
```diff
@@ -343,6 +344,47 @@ def __init__(self, directive: "DocumenterBridge", name: str, indent: str = '') -
...
+        """
+        if self.real_modname is None or self.objpath is None:
+            # We require the real module name.
+            return None
+
+        subject = self.object
+        if isinstance(subject, ModuleType):
+            # modules do not have object paths
+            canonical_objpath = []
+        elif isinstance(subject, (partial, partialmethod)):
```
cblegare: This part here is not doing the trick I was hoping for it to do.

MuellerSeb:
Is this still ongoing? Would love to see this feature!

tk0miya:
I posted #9026 as an alternative of this implementation. I think it's too much to output `:canonical:` option for all of descriptions. So #9026 implements `:canonical:` option generator only to the ClassDocumenter and FunctionDocumenter.

tk0miya:
Now I merged #9026 instead of this. I really appreciate @cblegare's first implementation. It helps me to implement #9026.
Thanks!

PR #8464, opened 2020-11-21
Deprecate Documenter.get_object_members()

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- At present, the implementation of Documenter.get_object_members() is
only for class objects.  In fact, no subclasses use it (all of them are
overriding the method).
- This deprecates Documenter.get_object_members() and copies it to
ClassDocumenter.get_object_members().

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -23,7 +23,8 @@ from docutils.statemachine import StringList
 import sphinx
 from sphinx.application import Sphinx
 from sphinx.config import ENUM, Config
-from sphinx.deprecation import RemovedInSphinx40Warning, RemovedInSphinx50Warning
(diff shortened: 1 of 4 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -620,6 +621,8 @@ def get_object_members(self, want_all: bool) -> Tuple[bool, ObjectMembers]:
         If *want_all* is True, return all members.  Else, only return those
         members given by *self.options.members* (which may also be none).
         """
+        warnings.warn('The implementation of Documenter.get_object_members() will be '
+                      'removed from Sphinx-6.0.', RemovedInSphinx60Warning)
```
tk0miya, line 625: The `Documenter.get_object_members()` has been used since its beginning. So it would be better to make the deprecation period long. So I'll keep this while two years (4.x and 5.x).

PR #8465, opened 2020-11-21
Fix #8460: autodoc: Support custom types defined by typing.NewType

### LINKED ISSUE #8460: Report underlying type information for NewType and TypeVar
Autodoc for NewType and TypeVar report the name of the type and the developer doc-comments but doesn't report the type metadata. 

For NewType I'd expect it to report the underlying type.
For example:
MyInt = NewType('MyInt', int) should report in the generated docs that the underlying type is int.

ForTypeVar there are multiple attributes. If there's bound type information, that should be reported. Additionally, it should report whether the type is invariant, covariant, or contravariant. As a bonus, cross-linking between invariant, covariant, or contravariant TypeVar declarations for the same underlying type.

The alternative would be  for the developer to manually document the type information they've already declared - which seems redundant since this should be easy to resolve.

**Additional context**
- See also discussion in #8415

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
-   A custom type defined by typing.NewType was rendered as a function
    because the generated type is a function having special attributes.
    This renders it as a variable.

-   Note: The module name where the NewType object defined is lost on
    generating it.  So it is hard to make cross-reference for these custom
    types.


tests/test_ext_autodoc.py
```diff
@@ -1758,6 +1758,13 @@ def test_autodoc_TypeVar(app):
         '   T5',
         '',
         "   alias of TypeVar('T5', contravariant=True)",
+        '',
+        '.. py:data:: T6',
+        '   :module: target.typevar',
+        '',
+        '   T6',
+        '',
+        '   alias of int',
```
tk0miya: It would be better if it is hyperlinked.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1691,7 +1691,12 @@ def can_document_member(cls, member: Any, membername: str, isattr: bool, parent:
 
     def import_object(self, raiseerror: bool = False) -> bool:
         try:
-            return super().import_object(raiseerror=True)
+            ret = super().import_object(raiseerror=True)
+            if inspect.isNewType(self.object):
+                self.options = Options(self.options)
+                self.options['annotation'] = SUPPRESS
```
tk0miya: It is a bit hacky to modify options...

sphinx/ext/autodoc/__init__.py
```diff
@@ -1691,7 +1691,12 @@ def can_document_member(cls, member: Any, membername: str, isattr: bool, parent:
 
     def import_object(self, raiseerror: bool = False) -> bool:
         try:
-            return super().import_object(raiseerror=True)
+            ret = super().import_object(raiseerror=True)
+            if inspect.isNewType(self.object):
```
tk0miya: Note: I have to install a handler to AttributeDocumenter also.

sphinx/ext/autosummary/generate.py
```diff
@@ -89,11 +89,12 @@ def setup_documenters(app: Any) -> None:
                                     DecoratorDocumenter, ExceptionDocumenter,
                                     FunctionDocumenter, GenericAliasDocumenter,
                                     InstanceAttributeDocumenter, MethodDocumenter,
-                                    ModuleDocumenter, PropertyDocumenter,
-                                    SingledispatchFunctionDocumenter, SlotsAttributeDocumenter)
+                                    ModuleDocumenter, NewTypeDataDocumenter,
```
tk0miya: `NewTypeAttributeDocumenter` should be also added.

Issue #7388, opened 2020-03-27
Add support for rate limit detection / retries @ linkcheck


webknjaz:
**Is your feature request related to a problem? Please describe.**

When there's a lot of references to GitHub issues and Pull Requests, linkcheck goes over them quite quickly and at some point, sometimes, it hits their rate limit.

**Describe the solution you'd like**

linkcheck needs some sort of a setting for rate-limiting (leaky bucket?) and retries (exponential backoff?). These settings could be per-domain similar to https://github.com/sphinx-doc/sphinx/issues/7247#issuecomment-594666822 for flexibility.

Also, maybe there could be another setting for a callable supplied by users. This'd allow inspecting the response and returning `True`/`False` from such a callback specifying if the failure is temporary and the request should be retried later (similar to `giveup` callback in https://pypi.org/p/backoff/ library — maybe it could even be integrated?).

**Describe alternatives you've considered**

N/A

**Additional context**

Basically, this fails in CI:
```console
[31;01mWarning, treated as error:[39;49;00m
/home/zuul/src/github.com/cherrypy/cheroot/docs/history.rst.rst:241:broken link: https://github.com/cherrypy/cheroot/pull/80 (429 Client Error: too many requests for url: https://github.com/cherrypy/cheroot/pull/80)
ERROR: InvocationError for command /home/zuul/src/github.com/cherrypy/cheroot/.tox/linkcheck-docs/bin/python -m sphinx -j auto -a -n -W -b linkcheck --color -d /home/zuul/src/github.com/cherrypy/cheroot/.tox/docs_doctree . /home/zuul/src/github.com/cherrypy/cheroot/.tox/docs_out (exited with code 2)
```
_(https://zuul.vexxhost.dev/t/cherrypy/build/e02b806392eb41c7a4030ec4bae735c3/console)_

tk0miya:
+1; Reasonable. I'm not familiar with rate-limit detection. But it would be helpful.
Also +1 for exponential back-off (or retries).

blink1073:
I implemented similar logic here for reference: https://github.com/minrk/pytest-check-links/pull/7

mgeier:
See also #6629.

francoisfreitag:
Should be solved by https://github.com/sphinx-doc/sphinx/pull/8467.

tk0miya:
@francoisfreitag Could I close this?

francoisfreitag:
[...]

Issue #8391, opened 2020-11-08
Enable async capabilities for HTTP requests


francoisfreitag:
**Is your feature request related to a problem? Please describe.**
From #6629, the `linkcheck` command is currently using threads to concurrently check the status of links in the documentation. Threads are not the most efficient way to concurrently check links: once all threads are busy waiting, the work queue stops being consumed.

**Describe the solution you'd like**

Using an event loop allows the URL verifier to yield control to another coroutine until it gets a response. That means a single thread is able to send multiple requests concurrently and process the response as they arrive. It also facilitates handling rate-limiting, because a coroutine can be scheduled to run in the future.

The first step toward using asynchronous concurrency in `linkchecker` is to replace the [`requests`](https://requests.readthedocs.io/en/latest/) library uses with an async-compatible HTTP library. The [`aiohttp`](https://docs.aiohttp.org/en/stable/) library has an API pretty similar to that of `requests` and is well-established and under active development, it seems like a good choice.

**Describe alternatives you've considered**

Tried handling rate limits with a PriorityQueue as described in https://github.com/sphinx-doc/sphinx/issues/6629#issuecomment-723440073.

**TODO**

- [ ] Increase test coverage of the existing code
- [ ] Use a wrapper that makes async code synchronous for sync use cases (get event loop, queue the request, run event loop until complete).
- [ ] Adapt existing code that expects a [`requests.Response`](https://requests.readthedocs.io/en/latest/api/#requests.Response) to use an [`aiohttp.ClientResponse`](https://docs.aiohttp.org/en/stable/client_reference.html#response-object). Both look pretty similar.
- [ ] Make a compatibility wrapper for arguments where a requests object was expected and `aiohttp` expects a different input. Consider `REQUESTS_CA_BUNDLE`, `tls_cacerts`, `auth_info` for the `linkcheck_auth` setting.

francoisfreitag:
I’m hitting an issue with this work: HTTP authentication.

[`linkcheck_auth`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_auth) documents its `auth_info` parameter as

> anything that is understood by the requests library (see [requests Authentication](https://www.sphinx-doc.org/en/master/usage/requests-auth) for details).
 
[...]

PR #8467, opened 2020-11-21
Fix #6629: linkcheck: Handle rate-limiting

### LINKED ISSUE #6629: Add exponential backoff to linkcheck
We link to Github PR/issues for each entry in our changelog, which is included in our documentation. That means that we fire off a hundred or so checks to various pages on Github. Whenever it's not running smoothly, we end up with timeouts, which slows down our development until Github is more reliable again.

Allow the linkcheck builder to retry requests a few times (maybe 3, or a configurable amount) with exponential backoff. https://www.peterbe.com/plog/best-practice-with-retries-with-requests gives a quick and easy implementation of how to do this with `requests.get`, which is what linkcheck currently uses to get pages.

Another thing that might help would be to collect all the links to check in one step, then sort them such that requests to the same domain happen more sequentially than requests to different domains. I.e., you would do one request to each unique domain first before you try a request to a domain you've already tried before.

francoisfreitag:
### Feature or Bugfix
- Feature

### Purpose
Follow the Retry-After header if present, otherwise use an exponential
back-off.

### Relates
Issue #6629
Issue #7388

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -39,6 +39,8 @@ Features added
 * #6914: Add a new event :event:`warn-missing-reference` to custom warning
   messages when failed to resolve a cross-reference
 * #6914: Emit a detailed warning when failed to resolve a ``:ref:`` reference
+* #6629: linkcheck: The builder now handles rate limits. See
+  :confval:`linkcheck_retry_on_rate_limit` for details.
 
 Bugs fixed
 ----------
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -2525,6 +2525,23 @@ Options for the linkcheck builder
 
    .. versionadded:: 2.3
 
+.. confval:: linkcheck_rate_limit_timeout
+
+   The ``linkcheck`` builder may issue a large number of requests to the same
+   site over a short period of time. This setting controls the builder behavior
+   when servers indicate that requests are rate-limited.
+
+   If a server indicates when to retry (using the `Retry-After`_ header),
+   ``linkcheck`` always follows the server indication.
+
+   Otherwise, ``linkcheck`` waits for a minute before to retry and keeps
+   doubling the wait time between attempts until it succeeds or exceeds the
+   ``linkcheck_rate_limit_timeout``. By default, the timeout is 5 minutes.
+
+   .. _Retry-After: https://tools.ietf.org/html/rfc7231#section-7.1.3
+
+   .. versionadded:: 3.4
+
 
 Options for the XML builder
 ---------------------------
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -13,13 +13,17 @@ import queue
 import re
 import socket
 import threading
+import time
+from datetime import datetime, timezone
+from email.utils import parsedate_to_datetime
 from html.parser import HTMLParser
 from os import path
-from typing import Any, Dict, List, Set, Tuple
+from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple
 from urllib.parse import unquote, urlparse
 
 from docutils import nodes
 from docutils.nodes import Node
+from requests import Response
 from requests.exceptions import HTTPError, TooManyRedirects
 
 from sphinx.application import Sphinx
@@ -33,10 +37,14 @@ logger = logging.getLogger(__name__)
 
 uri_re = re.compile('([a-z]+:)?//')  # matches to foo:// and // (a protocol relative URL)
 
+RateLimit = NamedTuple('RateLimit', (('delay', float), ('next_check', float)))
 
 DEFAULT_REQUEST_HEADERS = {
     'Accept': 'text/html,application/xhtml+xml;q=0.9,*/*;q=0.8',
 }
+CHECK_IMMEDIATELY = 0
+QUEUE_POLL_SECS = 1
+DEFAULT_DELAY = 60.0
 
 
 class AnchorCheckParser(HTMLParser):
@@ -98,7 +106,8 @@ class CheckExternalLinksBuilder(Builder):
         open(path.join(self.outdir, 'output.json'), 'w').close()
 
         # create queues and worker threads
-        self.wqueue = queue.Queue()  # type: queue.Queue
+        self.rate_limits = {}  # type: Dict[str, RateLimit]
+        self.wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue
         self.rqueue = queue.Queue()  # type: queue.Queue
         self.workers = []  # type: List[threading.Thread]
         for i in range(self.app.config.linkcheck_workers):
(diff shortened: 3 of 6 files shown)


francoisfreitag:
The Windows checks are failing on tests `test_too_many_requests_user_retry_float` ad `test_too_many_requests_retry_exponential_backoff`. I could not think of a way to meaningfully test the retrying other than actually sleep and measure the wait time. I am happy to try your suggestions.

I can increase the margin on Windows to accommodate for the seemingly longer time needed to start a thread, or skip the test on Windows. I would prefer not to remove these tests.

---

I ended up implementing a version with the Requests library, because of https://github.com/sphinx-doc/sphinx/issues/8391#issuecomment-727266266. I prefer to avoid forcing the use of async when not absolutely mandatory. That way, the project can move on without async, until we decide the time is ripe.

I am happy to pursue the async way, I am convinced it will yield better performance and possibly simplify the implementation, but at the expense of:
- authentication scheme support, because the AIOHTTP project does not support digest auth, OAuth, NTLM, etc.
- some backward incompatibilities described in https://github.com/sphinx-doc/sphinx/issues/8391#issuecomment-731566666

doc/usage/configuration.rst
```diff
@@ -2525,6 +2525,64 @@ Options for the linkcheck builder
...
+   The retry policy function is called immediately after a server returns an
+   HTTP response with status code `429
+   <https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429>`_ (Too Many
+   Requests).
+
+   Example:
+
+   .. code-block:: python
+
+      linkcheck_retry_on_rate_limit = {
```
tk0miya: Unfortunately, a confval should not have lambda as its value. It breaks the incremental build feature of sphinx-build because it's not serializable. I think it's enough to provide options to control these behavior. Additionally, it would be better to provide an event for those who'd like to enhance it via code, but I still don't understand that there is who wants to do that.

sphinx/builders/linkcheck.py
```diff
@@ -33,10 +37,14 @@
 
 uri_re = re.compile('([a-z]+:)?//')  # matches to foo:// and // (a protocol relative URL)
 
+RateLimit = namedtuple('RateLimit', ('delay', 'next_check'))
```
tk0miya: Could you use `typing.NamedTuple` instead? It allows annotating types of members.

Example:
https://github.com/sphinx-doc/sphinx/blob/ad804647b5c1bbbe3bb4bfd23e23bbf0ee20b2dc/sphinx/config.py#L41-L43

sphinx/builders/linkcheck.py
```diff
@@ -172,23 +181,38 @@ def check_uri() -> Tuple[str, str, int]:
...
                                                 auth=auth_info, **kwargs)
                         response.raise_for_status()
             except HTTPError as err:
                 if err.response.status_code == 401:
                     # We'll take "Unauthorized" as working.
                     return 'working', ' - unauthorized', 0
+                elif err.response.status_code == 429:
+                    next_check = self.limit_rate(err.response)
+                    if next_check is not None:
+                        self.wqueue.put((next_check, (uri, docname, lineno)), False)
```
tk0miya: Why do you use nested tuple here? I feel flat tuple is more simple.

sphinx/builders/linkcheck.py
```diff
@@ -247,11 +271,89 @@ def check(docname: str) -> Tuple[str, str, int]:
...
+                continue
             status, info, code = check(docname)
-            self.rqueue.put((uri, docname, lineno, status, info, code))
+            if status == 'rate-limited':
+                logger.info(darkgray('-rate limited-   ') + uri + darkgray(' | sleeping...'))
+            else:
+                self.rqueue.put((uri, docname, lineno, status, info, code))
+            self.wqueue.task_done()
+
+    def limit_rate(self, response):
```
tk0miya: Please add type-hints to here.

tests/test_build_linkcheck.py
```diff
@@ -382,3 +392,198 @@ def test_connect_to_selfsigned_nonexistent_cert_file(app):
...
+        timestamp_str = ts_re.match(line).group("ts")
+        if sys.version_info >= (3, 7):
+            timestamp = datetime.fromisoformat(timestamp_str)
+        else:
+            timestamp = datetime.strptime(timestamp_str, "%Y-%m-%dT%H:%M:%S.%f")
+        timestamps.append(timestamp)
+    # Duration which allocated to the thread to do its work.
+    margin = 1.0
+    assert 1.0 <= (timestamps[1] - start).total_seconds() < 1.0 + margin
+    assert 2.0 <= (timestamps[2] - timestamps[1]).total_seconds() < 2.0 + margin
```
tk0miya: How about patching `time.sleep()`? It does not waste real CI time and help us to measure how long the linkcheck module sleeps. I don't know it is available even under threading. But it's fragile to do sleeping in the real world.

doc/usage/configuration.rst
francoisfreitag: I simplified the implementation to only accept a float value: the maximum timeout. Perhaps that’s enough.

sphinx/builders/linkcheck.py
francoisfreitag: It better separates the priority number from the data, but a flat tuple works as well. I used a flat tuple in the latest revision.

tests/test_build_linkcheck.py
francoisfreitag: I agree that sleeping is both fragile and wasting test time. It would like to avoid it, but do not see a way to do it. Mocking `time.time()` leads to very tight coupling with the implementation.

Patching `time.sleep()` cause consumer threads to busy-loop (instead of sleep) until the `next_check` timestamp (the priority of the link). It wastes CPU cycles for no benefits.
We could peek at the priority queue, see the `next_check`s queued, but that’s sensitive to the timing of the test and the consumer threads still need to wait until the `next_check`s.

Another option is to unit test the `limit_rate` function, verify its return value (the next check) for various inputs.
It does not verify the entire system behavior, thus offers less guarantees. It is also more coupled with the implementation. OTOH, it avoids sleeping, some mocking and parsing of the test webserver logs, and allows more extensive testing of `limit_rate`. I pushed a separate commit with what I have in mind.

francoisfreitag:
Thanks for the review!

I believe I addressed all feedback and the patch is ready for another look.
The test suite is now passing with the changes suggested in https://github.com/sphinx-doc/sphinx/pull/8467#discussion_r528319354.

CHANGES
```diff
@@ -33,6 +33,8 @@ Features added
 * #6914: Add a new event :event:`warn-missing-reference` to custom warning
   messages when failed to resolve a cross-reference
 * #6914: Emit a detailed warning when failed to resolve a ``:ref:`` reference
+* #6629: The linkcheck builder now handles rate limits. See
+  :confval:`linkcheck_retry_on_rate_limit` for details.
```
tk0miya: Could you add a category to the entry?
```suggestion
* #6629: linkcheck: The linkcheck builder now handles rate limits. See
  :confval:`linkcheck_retry_on_rate_limit` for details.
```

doc/usage/configuration.rst
```diff
@@ -2525,6 +2525,23 @@ Options for the linkcheck builder
...
+   when servers indicate that requests are rate-limited.
+
+   If a server indicates when to retry (using the `Retry-After`_ header),
+   ``linkcheck`` always follows the server indication.
+
+   Otherwise, ``linkcheck`` waits for a minute before to retry and keeps
+   doubling the wait time between attempts until it succeeds or exceeds the
+   ``linkcheck_rate_limit_timeout``. By default, the timeout is 5 minutes.
+
+   .. _Retry-After: https://tools.ietf.org/html/rfc2616#section-14.37
```
tk0miya: RFC 2616 was obsoleted. RFC 7231 is better now.
https://tools.ietf.org/html/rfc7231#section-7.1.3

tk0miya (approved):
LGTM with nits.

francoisfreitag:
Rebased, applied suggestions (removed duplicate mention of linkcheck in the CHANGES) and squashed the commits.

tk0miya:
Merged! Thank you for your great work!

PR #8470, opened 2020-11-21
osutil FileAvoidWrite support non-ascii content

panhaoyu:
Subject: make FileAvoidWrite support non-ascii content

branch: 3.x

### Feature or Bugfix
- Bugfix

### Purpose

I created a custom `package.rst_t` and add some non-ascii words, but the encoding was wrong.

With my fix, it works well.

image

image

### DIFF
--- a/sphinx/util/osutil.py
+++ b/sphinx/util/osutil.py
@@ -222,14 +222,14 @@ class FileAvoidWrite:
         self._io.close()
 
         try:
-            with open(self._path) as old_f:
+            with open(self._path, encoding=fs_encoding) as old_f:
                 old_content = old_f.read()
                 if old_content == buf:
                     return
         except OSError:
             pass
 
-        with open(self._path, 'w') as f:
+        with open(self._path, 'w', encoding=fs_encoding) as f:
             f.write(buf)
 
     def __enter__(self) -> "FileAvoidWrite":


tk0miya:
What encoding did you use for your template? Sphinx expects utf-8, not fs_encoding.

sphinx/util/osutil.py
```diff
@@ -222,14 +222,14 @@ def close(self) -> None:
         self._io.close()
 
         try:
-            with open(self._path) as old_f:
+            with open(self._path, encoding=fs_encoding) as old_f:
```
tk0miya, line 225: This is wrong way.  The `source_encoding` configuration should be used instead.

panhaoyu:
image

I think this image will help.

Simply `with open` will not use `utf-8` on my computer.

panhaoyu:
I'm new to sphinx and i don't know what is fs_encoding, but it works well and it is `utf-8`. I think `with open('f', 'w',encoding='utf-8')` is ok and even better.

tk0miya:
`fs_encoding` is defined at the module you modified.
https://github.com/sphinx-doc/sphinx/blob/3ea1ec84cc610f7a9f4f6b354e264565254923ff/sphinx/util/osutil.py#L160

It means "the character encoding of the file system". So it would become non "utf-8" values on different systems. It should be used only for the filenames, not for its content.

Sphinx expects to use `utf-8` encoding to the template files. Please convert the encoding of your templates.

PR #8471, opened 2020-11-22
Fix #8460: autodata directive does not display the detail of TypeVars

### LINKED ISSUE #8460: Report underlying type information for NewType and TypeVar

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- The details are only shown when automodule directive is used.  This
allows to document them via other auto* directives.
- refs: #8460


tests/test_ext_autodoc.py
```diff
@@ -1750,27 +1758,31 @@ def test_autodoc_TypeVar(app):
         '',
         "   alias of TypeVar('T1')",
         '',
+        '',
         '.. py:data:: T3',
         '   :module: target.typevar',
         '',
         '   T3',
         '',
         "   alias of TypeVar('T3', int, str)",
```
tk0miya, line 1767: I'll make these arguments hyperlinked in the next step.

sphinx/ext/autodoc/__init__.py
```diff
@@ -2049,7 +2071,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
 
 
-class AttributeDocumenter(DocstringStripSignatureMixin, ClassLevelDocumenter, NewTypeMixin):  # type: ignore  # NOQA
+class AttributeDocumenter(DocstringStripSignatureMixin, ClassLevelDocumenter, NewTypeMixin, TypeVarMixin):  # type: ignore  # NOQA
```
tk0miya: The order should be changed to correct MRO.

tests/test_ext_autodoc.py
jp-larose: I don't see anything for the `bound`  parameter for TypeVar.  It should be accessible via the `__bound__` attribute.  It's important information to display.

tk0miya: Could you file a new issue? Then we'll consider improving it.

PR #8472, opened 2020-11-22
Fix #8446: html: consecutive spaces are displayed as single space

### LINKED ISSUE #8446: autodoc strips spaces from values
If I define a module level data as:
```
INDENT: str = '    ' 
"""
: Defines the single level indentation that will be used in the generated stubs /  wrapper. Default is four spaces. 
Can be overridden if necessary.
"""
```

and in my .rst file
```
.. autodata:: INDENT
```

It is rendered as as a single space
image

Expected result is four spaces as defined in .py file.

Same result if defined with` .. data::` directive.

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- In HTML, consecutive spaces are considered as single space by HTML
browsers.  To represent them as is, we have to escape them on rendering
them into HTML.  This starts to escape the whole of desc_signature node.
- refs: #8446


sphinx/writers/html.py
```diff
@@ -129,8 +129,10 @@ def depart_desc(self, node: Element) -> None:
     def visit_desc_signature(self, node: Element) -> None:
         # the id is set automatically
         self.body.append(self.starttag(node, 'dt'))
+        self.optional_param_level += 1
```
tk0miya: @jakobandersen Could you tell me your opinion on this change, please? Now I'm going to escape the whole of  `desc_signature` node. As a result, consecutive spaces are rendered as-is in HTML browsers. Additionally, non-spaces becomes not foldable. Is there any side-effect to C and C++ domains?

PR #8483, opened 2020-11-23
Don't require typed_ast in [test] with Python 3.8+

hroncok:
### Feature or Bugfix

- Feature
- Bugfix
- Refactoring

Neither?

### Purpose

To avoid a dependency.


tk0miya:
Thanks!

PR #8484, opened 2020-11-24
Close #8326: Rename master_doc to root_doc

### LINKED ISSUE #8326: conf.py parameter for default document name uses exclusionary language

In a number of projects I am involved with which use sphinx, avoiding exclusionary language, and especially terms that are related to racial injustices, is a priority. In `conf.py`, to override which document is the root, it uses the term `master_doc`.

An alias to the term `master_doc` in `conf.py`. For example, if this was `root_doc`, then using

```
root_doc = "README.rst"
```

would have the same effect as

```
master_doc = "README.rst"
```

If both are defined, this is either an error or `root_doc` takes precedence. I have no strong feelings about the behavior in this corner case.

* We can have alternative names for `root_doc` (for example, `content_doc`)
* For some use cases, taking advantage of the default `index.rst` works well, so I don't need to explicitly set it.
* I considered playing games with global so offensive terms are not explicit in my code. e.g. `globals()[rot13(...)] = "README.rst`
* I can copy files to the default filename.

tk0miya:
### Feature or Bugfix
- Feature
- Refactoring

### Purpose
- refs: #8326 
- To describe the purpose more accurately, the `master_doc` is now renamed
to `root_doc`.
-  The old name is still available.  But it is recommeneded
    to use new one from now on.


doc/usage/configuration.rst
```diff
@@ -175,13 +175,15 @@ General configuration
       Now Sphinx provides an API :meth:`.Sphinx.add_source_parser` to register
       a source parser.  Please use it instead.
 
-.. confval:: master_doc
+.. confval:: root_doc
```
tk0miya, line 191: To keep the hyperlink to the `master_doc`, it would be better not to update it by new name.

sphinx/cmd/quickstart.py
```diff
@@ -358,7 +359,7 @@ def write_file(fpath: str, content: str, newline: str = None) -> None:
     write_file(path.join(srcdir, 'conf.py'), template.render_string(conf_text, d))
 
     masterfile = path.join(srcdir, d['master'] + d['suffix'])
-    write_file(masterfile, template.render('quickstart/master_doc.rst_t', d))
+    write_file(masterfile, template.render('quickstart/root_doc.rst_t', d))
```
tk0miya: This must be a breaking change...

sphinx/templates/htmlhelp/project.hhc
```diff
@@ -18,7 +18,7 @@
     </OBJECT>
```
tk0miya, line 18: Is this still needed?

doc/usage/restructuredtext/directives.rst
```diff
@@ -197,9 +197,9 @@ tables of contents.  The ``toctree`` directive is the central element.
    <metadata>` to let a document be built, but notify Sphinx that it is not
    reachable via a toctree.
 
-   The "master document" (selected by :confval:`master_doc`) is the "root" of
-   the TOC tree hierarchy.  It can be used as the documentation's main page, or
-   as a "full table of contents" if you don't give a ``maxdepth`` option.
+   The "root document" (selected by :confval:`root`) is the "root" of the TOC
```
tk0miya: typo

PR #8494, opened 2020-11-26
Use restify() to reference alias attributes in autodoc

### LINKED ISSUE #8493: References to builtins not working in autodoc class aliases
**Describe the bug**
When aliasing an attribute to another class, sphinx adds the line "alias of ..." to the attribute's documentation.
If the other class is a builtin, this adds e.g.:
```
alias of :class:`builtins.dict`
```
However, the intersphinx mapping to the official Python documentation only works without the "builtins"-part in the module path.

When using the nit-picky mode of sphinx, this results in the following warning:
```
WARNING: py:class reference target not found: builtins.int
```
**To Reproduce**
There's a [test case](https://github.com/sphinx-doc/sphinx/blob/e6e23aa45b9191655fb7790b0b97826ce663119a/tests/test_ext_autodoc.py#L809) which shows this behaviour.

**Expected behavior**
I expect the autodoc extension to generate the correct links suited for intersphinx mappings:
```
alias of :class:`dict`
```

**Environment info**
- OS: Arch Linux
- Python version: 3.7.9 & 3.8.6
- Sphinx version: 3.3.1
- Extra tools: None

**Additional context**
I think it makes sense to use the utilitiy function [restify](https://github.com/sphinx-doc/sphinx/blob/e6e23aa45b9191655fb7790b0b97826ce663119a/sphinx/util/typing.py#L89) for this. I will provide a patch soon.

timobrembeck:
### Feature or Bugfix

- Bugfix
- Refactoring

### Purpose
- Alias references to builtin classes did not work correctly

### Detail
- Make use of the utility function [restify](https://github.com/sphinx-doc/sphinx/blob/e6e23aa45b9191655fb7790b0b97826ce663119a/sphinx/util/typing.py#L89) to determine the correct module path for the `:class:` reference
- Modify expected outcome of test case

Fix #8493

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -68,6 +68,7 @@ Bugs fixed
 * #8131: linkcheck: Use GET when HEAD requests cause Too Many Redirects, to
   accommodate infinite redirect loops on HEAD
 * #8437: Makefile: ``make clean`` with empty BUILDDIR is dangerous
+* #8493: autodoc: references to builtins not working in class aliases
 
 Testing
 --------
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1654,16 +1654,8 @@ class ClassDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # type: 
     def add_content(self, more_content: Optional[StringList], no_docstring: bool = False
                     ) -> None:
         if self.doc_as_attr:
-            classname = safe_getattr(self.object, '__qualname__', None)
-            if not classname:
-                classname = safe_getattr(self.object, '__name__', None)
-            if classname:
-                module = safe_getattr(self.object, '__module__', None)
-                parentmodule = safe_getattr(self.parent, '__module__', None)
-                if module and module != parentmodule:
-                    classname = str(module) + '.' + str(classname)
-                content = StringList([_('alias of :class:`%s`') % classname], source='')
-                super().add_content(content, no_docstring=True)
+            content = StringList([_('alias of %s') % restify(self.object)], source='')
+            super().add_content(content, no_docstring=True)
         else:
             super().add_content(more_content)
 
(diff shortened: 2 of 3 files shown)


timobrembeck:
Can anybody point me in the right direction what to do about the failing tests?

- Did I misinterpret the purpose of the `restify()`-function and should I rather stick to the original implementation + a check on builtins? If so, why is it used [here](https://github.com/sphinx-doc/sphinx/blob/e6e23aa45b9191655fb7790b0b97826ce663119a/sphinx/ext/autodoc/__init__.py#L1583) and [here](https://github.com/sphinx-doc/sphinx/blob/e6e23aa45b9191655fb7790b0b97826ce663119a/sphinx/ext/autodoc/__init__.py#L1730) without problems?
- Is there a bug in the `_restify_py36()` part of `restify()`-function?
- Can the expected docstrings of the failing tests be adapted to the new output of `restify()`?

Output without restify():
```
alias of :class:`typing.List`
```
Output with restify():
```
alias of :class:`List`\\ [:class:`int`]
```
Thanks a lot in advance!

tk0miya:
Thank you for your work! Looks good to me.

>Did I misinterpret the purpose of the restify()-function and should I rather stick to the original implementation + a check on builtins? If so, why is it used here and here without problems?

No special reason to not using `restify()` in there. It was added recently and not applied into other places. So this change is good.

>Is there a bug in the _restify_py36() part of restify()-function?
>Can the expected docstrings of the failing tests be adapted to the new output of restify()?

The CI errors are a side effect of this change. Because generic type aliases are processed by ClassDocumenter in py35 and py36. So test cases should be updated to the new output of `restify()`. It is surely `List[int]`.
https://github.com/sphinx-doc/sphinx/blob/00cd12e7e28a4e69680ee7930db274d79149b238/tests/roots/test-ext-autodoc/target/genericalias.py#L4

tk0miya:
Note: I'll change `GenericAliasDocumenter` also (after this change merged).

timobrembeck:
> The CI errors are a side effect of this change. Because generic type aliases are processed by ClassDocumenter in py35 and py36. So test cases should be updated to the new output of `restify()`. It is surely `List[int]`.

Thanks for your explanation! I updated the failing test case now.

tk0miya (approved):
LGTM!

PR #8495, opened 2020-11-26
fix for #8491

### LINKED ISSUE #8491: `default_role="Any"` + missing reference cause a crash on 3.x branch
**Describe the bug**
If the conf.py file contains the line `default_role="any"`, missing references in docstrings cause a python crash:
```
Extension error:
Handler <function warn_missing_reference at 0x############> for event 'warn-missing-reference' threw an exception (exception: 'NoneType' object has no attribute 'name')
```

**To Reproduce**
```
$ git clone https://github.com/QuentinSoubeyran/pheres
$ cd pheres
$ git checkout dev
IF you want a venv. The project requires python 3.9
$ python3.9 -m venv venv
$ source venv/bin/activate
END if
$ pip install -e .
$ cd sphinx
$ make html
# See crash
```
The missing reference is in `pheres/src/pheres/_jsonable.py`, at line 81, in the docstring there is a ``UndefinedReference`` which is undefined as it name implies.

**Expected behavior**
The warning works correctly instead of crashing sphinx, or the error explains that `default_role="any"` is incorrect if that is the problem.

**Your project**
See [pheres](https://github.com/QuentinSoubeyran/pheres/tree/dev)

**Environment info**
- OS: Ubuntu
- Python version: 3.9
- Sphinx version: commit 5a0e1231ad3fce9405ea52d90b21153e4ea86a27
- Sphinx extensions:  sphinx.ext.autodoc, sphinx.ext.napoleon, sphinx.ext.intersphinx, sphinx_rtd_theme

**Additional context**
The error is likely due to https://github.com/sphinx-doc/sphinx/blob/5a0e1231ad3fce9405ea52d90b21153e4ea86a27/sphinx/domains/std.py#L1110
not checking is the `domain` is not `None`

QuentinSoubeyran:
Subject: fix #8491 

### Feature or Bugfix
- Bugfix

### Purpose
- Subject: fix #8491 by checking that the `domain` argument is defined before accessing its attributes

### Detail
- #8491 is fixed by the proposed check. I am not familiar enough with sphinx to know if `domain` should never be `None`, but at least this prevent a crash.

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -1106,7 +1106,7 @@ class StandardDomain(Domain):
 
 
 def warn_missing_reference(app: "Sphinx", domain: Domain, node: pending_xref) -> bool:
-    if domain.name != 'std' or node['reftype'] != 'ref':
+    if (domain and domain.name != 'std') or node['reftype'] != 'ref':
         return None
     else:
         target = node['reftarget']


tk0miya (approved):
Thanks!

Note: this bug was added at #8075 and not released yet.

PR #8499, opened 2020-11-27
Replace modify_env() with pytest monkeypatch.setenv()

jdufresne:
Subject: Replace modify_env() with pytest monkeypatch.setenv()

### Feature or Bugfix

- Refactoring

### Purpose

The same test utility function is built into pytest. Can avoid the duplication.

### Relates

https://docs.pytest.org/en/latest/monkeypatch.html#monkeypatching-environment-variables


tk0miya (approved):
Oh, I did not know this. Good changing!

PR #8510, opened 2020-11-30
Allow user to use url to reference html logo & favicon

### LINKED ISSUE #2018: ROI : html logo link to user specify URL
It would be nice to be able to specify a URL the LOGO in HTML should point to. This has been posted already on the G group:

https://groups.google.com/forum/#!searchin/sphinx-users/html$20logo/sphinx-users/O4fHXX4b3wo/fkRpLmKFS1wJ

and a workaround proposed. But I guess it would be a natural option to provide.

Mardelor:
Subject: Allow user to use url to reference html logo & favicon

### Feature or Bugfix

- Feature

### Purpose
- To give user the possibility to use url to reference html logo and favicon. If a scheme is specified in the configuration file, the resource isn't copied as a a static file but simply referenced by its url.

### Detail
- html logo & favicon reference attributes are now determined when config values are processed. In html layouts, the directory `_static/` does no longer prefix the html logo & favicon reference. Instead, when checking config values, the builder add the `_static/` prefix is added if the scheme is empty. Otherwise, the config value is taken as the reference. 
- It could impact users who use custom html layouts

### Relates
- Closes #2018

This is my first contribution, so don't hesitate to tell me if I'm doing it wrong

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -962,7 +962,7 @@ class StandaloneHTMLBuilder(Builder):
         if parse_result.scheme:
             ref = uri
         elif parse_result.path:
-            ref = path.basename(parse_result.path)
+            ref = path.join('_static', path.basename(parse_result.path))
         return ref
 
     # --------- these are overwritten by the serialization builder
@@ -1188,14 +1188,14 @@ def validate_html_static_path(app: Sphinx, config: Config) -> None:
 
 def validate_html_logo(app: Sphinx, config: Config) -> None:
     """Check html_logo setting."""
-    if config.html_logo and not path.isfile(path.join(app.confdir, config.html_logo)):
+    if config.html_logo and not path.isfile(path.join(app.confdir, config.html_logo)) and not urlparse(config.html_logo).scheme:
         logger.warning(__('logo file %r does not exist'), config.html_logo)
         config.html_logo = None  # type: ignore
 
 
 def validate_html_favicon(app: Sphinx, config: Config) -> None:
     """Check html_favicon setting."""
-    if config.html_favicon and not path.isfile(path.join(app.confdir, config.html_favicon)):
+    if config.html_favicon and not path.isfile(path.join(app.confdir, config.html_favicon)) and not urlparse(config.html_favicon).scheme:
         logger.warning(__('favicon file %r does not exist'), config.html_favicon)
         config.html_favicon = None  # type: ignore
 
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -1188,14 +1188,16 @@ def validate_html_static_path(app: Sphinx, config: Config) -> None:
 
 def validate_html_logo(app: Sphinx, config: Config) -> None:
     """Check html_logo setting."""
-    if config.html_logo and not path.isfile(path.join(app.confdir, config.html_logo)) and not urlparse(config.html_logo).scheme:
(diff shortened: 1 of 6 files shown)


sphinx/builders/html/__init__.py
```diff
@@ -773,12 +773,12 @@ def onerror(filename: str, error: Exception) -> None:
                        excluded, context=context, renderer=self.templates, onerror=onerror)
 
     def copy_html_logo(self) -> None:
-        if self.config.html_logo:
+        if self.config.html_logo and not urlparse(self.config.html_logo).scheme:
```
tk0miya, line 776: It would be better to add a helper function like `isurl(s)` for readability. I think `sphinx.util` is a good place to add it.

sphinx/builders/html/__init__.py
```diff
@@ -1176,14 +1188,16 @@ def validate_html_static_path(app: Sphinx, config: Config) -> None:
 
 def validate_html_logo(app: Sphinx, config: Config) -> None:
     """Check html_logo setting."""
-    if config.html_logo and not path.isfile(path.join(app.confdir, config.html_logo)):
+    if config.html_logo and not path.isfile(path.join(app.confdir, config.html_logo)) \
+            and not urlparse(config.html_logo).scheme:
```
tk0miya, line 1192: Could you use parenthesis instead of backslash to line-continuation?
```suggestion
    if (config.html_logo and not path.isfile(path.join(app.confdir, config.html_logo)) and
            not urlparse(config.html_logo).scheme):
```

sphinx/builders/html/__init__.py
```diff
@@ -1176,14 +1188,16 @@ def validate_html_static_path(app: Sphinx, config: Config) -> None:
...
         logger.warning(__('logo file %r does not exist'), config.html_logo)
         config.html_logo = None  # type: ignore
 
 
 def validate_html_favicon(app: Sphinx, config: Config) -> None:
     """Check html_favicon setting."""
-    if config.html_favicon and not path.isfile(path.join(app.confdir, config.html_favicon)):
+    if config.html_favicon and not path.isfile(path.join(app.confdir, config.html_favicon)) \
+            and not urlparse(config.html_favicon).scheme:
```
tk0miya, line 1200: Here is also.
```suggestion
    if (config.html_favicon and not path.isfile(path.join(app.confdir, config.html_favicon)) and
            not urlparse(config.html_favicon).scheme):
```

sphinx/themes/agogo/layout.html
```diff
@@ -15,7 +15,7 @@
       <div class="header">
         {%- if logo %}
           <p class="logo"><a href="{{ pathto(master_doc)|e }}">
-            <img class="logo" src="{{ pathto('_static/' + logo, 1)|e }}" alt="Logo"/>
+            <img class="logo" src="{{ pathto(logo, 1)|e }}" alt="Logo"/>
```
tk0miya, line 18: I feel this is a breaking change to 3rd party themes. So it would be better to add another template variable like `logo_url` and keep the original `logo` variable as is.

tk0miya (changes requested):
Sorry for late reviewing. +1 for your idea. Could you refine this PR a bit more?

tk0miya:
Now I just refined this and post it as #8737. I'll merge it instead of this while after if no objections.

Mardelor:
Thanks for the review and the edition, and sorry I couldn't change my PR.
No objection to merge #8737 instead of it.

tk0miya:
Thanks!

tk0miya:
Now I merged #8737. It's based on your this PR. Thank you for your work!

PR #8511, opened 2020-11-30
Support testing from read-only filesystems

jbms:
Currently, tests fail if run from a read-only filesystem because the temporary directory trees copied from the source tree are marked read-only by shutil.copytree.

This change fixes that by ensuring directories created `sphinx.testing.path.copytree` are always writable.

### DIFF
--- a/sphinx/testing/path.py
+++ b/sphinx/testing/path.py
@@ -98,6 +98,15 @@ class path(str):
             pointed to by the symbolic links are copied.
         """
         shutil.copytree(self, destination, symlinks=symlinks)
+        # If source tree is marked read-only (e.g. because it is on a read-only
+        # filesystem), `shutil.copytree` will mark the destination as read-only
+        # as well.  To avoid failures when adding additional files/directories
+        # to the destination tree, ensure destination directories are not marked
+        # read-only.
+        for root, dirs, files in os.walk(destination):
+            os.chmod(root, 0o755)
+            for name in files:
+                os.chmod(os.path.join(root, name), 0o644)
 
     def movetree(self, destination: str) -> None:
         """


jbms:
Note: it seems like this should be merged into the master branch as well, assuming it is deemed acceptable for this branch.  Does that happen automatically or should I create a separate pull request?

sphinx/testing/path.py
```diff
@@ -98,6 +98,16 @@ def copytree(self, destination: str, symlinks: bool = False) -> None:
             pointed to by the symbolic links are copied.
         """
         shutil.copytree(self, destination, symlinks=symlinks)
+        # If source tree is marked read-only (e.g. because it is on a read-only
+        # filesystem), `shutil.copytree` will mark the destination as read-only
+        # as well.  To avoid failures when adding additional files/directories
+        # to the destination tree, ensure destination directories are not marked
+        # read-only.
+        if hasattr(os, 'chmod'):
```
tk0miya: Why do you check `hasattr` here? Is there any environment not having `os.chmod`?

tk0miya (commented):
LGTM with nits.

Note: The 3.x branch will be merged into the master branch (sometimes manually). So you don't need to post another PR for the branch :-)

tk0miya:
Now I'm thinking the adjusting permissions is a responsibility of `Path.copytree()`. Is it good way to change permissions automatically on copying (even if non testing files)? How about doing that inside`app_params` fixture instead? What do you think?

sphinx/testing/path.py
jbms: I assumed Windows did not have it, but checking now I see that `os.chmod` is defined on Windows as well, so that check can be removed.

jbms:
I eliminated the hasattr check.  I assumed the sphinx.testing.path module (where I made the change) was only used for tests.  The reason I put the change there rather than in the app_params fixture (which is actually where I had it originally) is that there are a few tests that call `copytree` directly as well.

tk0miya:
I'm still thinking about changing permissions on copying files by default. I guess it is rare to run tests in read-only FS. So it would be better to enable this feature intently. How about adding an environment variable to enable this?

sphinx/testing/path.py
```diff
@@ -98,6 +98,15 @@ def copytree(self, destination: str, symlinks: bool = False) -> None:
...
         shutil.copytree(self, destination, symlinks=symlinks)
+        # If source tree is marked read-only (e.g. because it is on a read-only
+        # filesystem), `shutil.copytree` will mark the destination as read-only
+        # as well.  To avoid failures when adding additional files/directories
+        # to the destination tree, ensure destination directories are not marked
+        # read-only.
+        for root, dirs, files in os.walk(destination):
+            os.chmod(root, 0o755)
+            for name in files:
+                os.chmod(os.path.join(root, name), 0o644)
```
tk0miya, line 109: It is much better to obey the settings of umask.

tk0miya (changes requested):

PR #8512, opened 2020-11-30
Fix nonascii_srcdir to work when non-ascii directories aren't supported

jbms:
Previously, the nonascii_srcdir in tests/test_build.py contained some
logic to attempt to handle the case where non-ascii directories aren't
supported, but it did not actually work because it fails to copy the
test sources to the temporary directory.

With this commit, the following test invocation, which previously
failed, now pases:

LANG=en_US.iso8859-1 python -m pytest test_build.py

### DIFF
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -37,6 +37,8 @@ def nonascii_srcdir(request, rootdir, sphinx_test_tempdir):
             (rootdir / 'test-root').copytree(srcdir)
     except UnicodeEncodeError:
         srcdir = basedir / 'all'
+        if not srcdir.exists():
+            (rootdir / 'test-root').copytree(srcdir)
     else:
         # add a doc with a non-ASCII file name to the source dir
         (srcdir / (test_name + '.txt')).write_text(dedent("""


tests/test_build.py
```diff
@@ -37,6 +37,8 @@ def nonascii_srcdir(request, rootdir, sphinx_test_tempdir):
             (rootdir / 'test-root').copytree(srcdir)
     except UnicodeEncodeError:
         srcdir = basedir / 'all'
+        if not srcdir.exists():
+            (rootdir / 'test-root').copytree(srcdir)
```
tk0miya, line 41: It would be better to call `pytest.skip` and skip testing if failed to prepare non-ASCII srcdir. What do you think?
https://docs.pytest.org/en/latest/skipping.html

jbms: This fixture is used by test_build_all, which seems like a fairly comprehensive test, and the purpose of that fixture just seemed to be to additionally include a non-ascii path.  From the existing code in nonascii_srcdir it seemed like the intent was for the test to proceed even with a non-ascii path.

After creating this pull request I found that there are other tests that fail with ascii locale, though, and ultimately decided it wasn't worth trying to run the tests with non-utf-8 locale.

jbms: For some context, I was attempting to run the sphinx tests in Bazel, which runs tests with LANG unset, and that causes Python3.6 to assume the filesystem encoding is ascii.  On python3.7 and later, LANG unset leads to a utf-8 fileystem encoding.  Originally I attempted to fix the tests to work with ascii fileystem encoding, but then decided it was easier to just ensure the tests ran with a utf-8 locale.

tk0miya: Okay, I withdraw my last comment. It's wrong.

BTW, what should I do for this PR? Is this PR no longer needed, right? Or, do you have a plan to rewrite this to ensure the utf-8 locale?

jbms: I don't need this anymore myself, but you might want to use it merge it nonetheless, since it's pretty minimal and fixes one test in non-utf8 case even though there are still other broken ones.

For my own purposes I fixed the tests by ensuring they run with utf-8 locale, using a separate wrapper script that also copies the tests in order to avoid the read-only issue, but I wasn't planning to submit that as a pull request.  Perhaps you could just document that the tests should be run from utf-8 locale.

tk0miya: Since python 3.7, as you said, python uses utf-8 encoding for filesystem enconding by default according to PEP-540.
https://www.python.org/dev/peps/pep-0540/

So this behavior only happens with python3.6 if my understanding is correct. And we'll support it for a while (maybe, until we'll switch our base platform to Ubuntu 20.04). So I'm merging this now.

tk0miya:
Thank you for your contribution. Good work!

PR #8523, opened 2020-12-07
fix #8522: sphinx could lead to `__bool__` method call, what can lead to side effect or direct failure in case it generate an error bug 

### LINKED ISSUE #8522: sphinx/ext/autodoc/__init__.py:719 could lead to `__bool__` method call, what can lead to side effect or direct failure in case it generate an error
**Describe the bug**
The code is the following:

        keep = False
        if safe_getattr(member, '__sphinx_mock__', None) is not None:
            # mocked module or object
            pass

But if ``safe_getattr`` returns some object with redefined ``__bool__``, e.g.

     def __bool__(self):
           raise TypeError("Do not pass this object to if and while clause, use all or any function")

     def __bool__(self):
           write_to_db()
           send_email()
           request_http()
           log()
           some_other_side_effect

This code will be executed and will lead to sphinx failure or unexpected side effects.

**To Reproduce**
Unfortunately I can't reproduce it as a toy example, but it lead to failure on my work project.

**Expected behavior**
Do not call user code.

artsiomkaltovich:
Subject: fix #8522.

### Feature or Bugfix
- Bugfix

### Purpose
Prevent user code execution, see examples in #8522.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -97,8 +97,8 @@ Features added
 
 Bugs fixed
 ----------
-* #8522: ``__bool__`` method isn't called now by autodoc module, so it won't lead to
-side effect or direct failure in case it generate an error.
+* #8522: ``__bool__`` method isn't called now by autodoc module, so it won't
+    lead to side effect or direct failure in case it generate an error.
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -98,7 +98,7 @@ Features added
 Bugs fixed
 ----------
 * #8522: ``__bool__`` method isn't called now by autodoc module, so it won't
-    lead to side effect or direct failure in case it generate an error.
+    lead to side effect or direct failure in case it generates an error.
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,5 @@
 Release 3.4.0 (in development)
 ==============================
-
 Dependencies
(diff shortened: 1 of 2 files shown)


CHANGES
```diff
@@ -1,6 +1,5 @@
 Release 3.4.0 (in development)
 ==============================
-
```
tk0miya: Please don't remove this blank line.

CHANGES
```diff
@@ -98,6 +97,8 @@ Features added
 
 Bugs fixed
 ----------
+* #8522: ``__bool__`` method isn't called now by autodoc module, so it won't
+    lead to side effect or direct failure in case it generates an error.
```
tk0miya: Potentially, autodoc may cause a side effect. So this changelog entry is not correct. In my understanding, this change reduces `__bool__` method calls on generating documents. It does not relate to "side effect or direct failure".

artsiomkaltovich: Hello @tk0miya.

Thank you for review. I've just ensured there isn't any else call like this, but you are right other side effects can occure, I will fix changelog.

artsiomkaltovich:
Hello @tk0miya.

I've fixed changes and resolve conflicts, it is ok now?

Regards,
Artyom.

tk0miya:
LGTM. Merging now.

PR #8528, opened 2020-12-12
Drop outdated workaround for distutils' Command._ensure_stringlike()

jdufresne:
Subject: Drop outdated workaround for distutils' Command._ensure_stringlike()

### Feature or Bugfix
- Refactoring

### Purpose

The issue applied to Python 2 only. The current override is identical to
the CPython implementation:

https://github.com/python/cpython/blob/4b8cdfcb22fbeaab9d954cb693a7fb3362a382b6/Lib/distutils/cmd.py#L207-L215

### DIFF
--- a/sphinx/setup_command.py
+++ b/sphinx/setup_command.py
@@ -14,7 +14,7 @@
 import os
 import sys
 from distutils.cmd import Command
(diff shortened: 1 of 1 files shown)


tk0miya:
Thank you for your wisdom!

PR #8535, opened 2020-12-13
Describe the intersphinx mapping a bit more

nedbat:
Subject: Clarify the intersphinx mapping, and provide another example.

### Documentation

I resorted to adding print statements to the Sphinx source in order to understand how the mapping worked.  I have a set of books that reference each other, and it was a pain to push them to readthedocs just to check if all the references were resolving properly.

This explains the mapping just a bit more and gives a new example based on my experience.


tk0miya (approved):
Good work!

PR #8554, opened 2020-12-19
change from apt to pip for debian and ubuntu

Torstein-Eide:
sphinx in apt is old
````bash
apt search python3-sphinx
Sorting... Done
Full Text Search... Done
 
python3-sphinx/focal 1.8.5-7ubuntu3 all
  documentation generator for Python projects (implemented in Python 3)

````

### DIFF
--- a/doc/usage/installation.rst
+++ b/doc/usage/installation.rst
@@ -31,7 +31,8 @@ Install either ``python3-sphinx`` using :command:`apt-get`:
 
 ::
 
-   $ apt-get install python3-sphinx
+   $ apt-get install python3-pip
+   $ pip3 install sphinx
 
 If it not already present, this will install Python for you.
 

### CODE REVIEW

File: doc/usage/installation.rst [hunk 1]
```diff
@@ -31,7 +31,8 @@ Install either ``python3-sphinx`` using :command:`apt-get`:
 
 ::
 
-   $ apt-get install python3-sphinx
+   $ apt-get install python3-pip
```
tk0miya, line 34: The above text also mentions the `python3-sphinx` package. It should be updated too.

Additionally, it would be better to mention not to mix Sphinx package from PyPI and distro's. Usually, people get confused about the difference between them.

tk0miya (changes requested):
LGTM with nits.

@mitya57 Do you have a good way to do this instead of pip? It would be nice if there is a workaround on Debian/Ubuntu.

mitya57:
> @mitya57 Do you have a good way to do this instead of pip?

What exactly do you mean?

Ubuntu Hirsute (development version) has the latest Sphinx, 3.3.1.

Focal has an older version because it was released in April, and it's receiving only bug fixes, not new features.

If you want an official backport of new version for Ubuntu Focal, please file a bug against sphinx package on Launchpad.

tk0miya:
I'd like to know the recommended way to upgrade Sphinx to the newer one in Ubuntu LTS. I think installing it via pip causes trouble. I saw many troubles caused by multiple Sphinx or python installation... So I hope there might be a better way to do that.

Torstein-Eide:
How can we have the normal apt installation in the documentation, if the package that is available, is outdated and not supported?

mitya57:
> I'd like to know the recommended way to upgrade Sphinx to the newer one in Ubuntu LTS. I think installing it via pip causes trouble. I saw many troubles caused by multiple Sphinx or python installation...

I think the only safe way to install newer Sphinx is using a virtualenv and pip inside it. Running pip as root can break system packages. Running it with `--user` is a bit better, but still can break something. Virtualenv is the safest option.

Ubuntu has [virtualenvwrapper][1] packaged which makes it really easy to create and manage virtualenvs.

[1]: https://virtualenvwrapper.readthedocs.io

tk0miya:
I worried about virtualenv makes users confused. It is a unique feature of python. So it is hard to understand for users who don't use python usually.

For now, my answer is upgrading their Ubuntu to the latest one. I'd not like to recommend users to install it via pip or virtualenv because it will increase the questions to forum or GitHub, I guess.

tk0miya:
It would be better to introduce the distro's package first. So I determined not to merge this. Thank you for contribution.

PR #8596, opened 2020-12-27
Close #8022: autodoc: Allow to hide the value of the variables via metadata

### LINKED ISSUE #8022: [autodoc] Allow `:annotation:` to be used in docstring
Currently if I want to add an attribute (or module constant) and document it with-in a code (let's assume autoclass/automodule is in usage), I can't stop Sphinx from showing the value of it in that same docstring. Instead, I have to manually do this in automodule directive:
```diff
 .. automodule:: module_name
     :members:
+    :exclude-members: CONSTANT_NAME
+
+    .. autodata:: CONSTANT_NAME
+        :annotation:
```

I would want to be able to just add `:annotation:` directly in the docstring.

Other than naming this role differently in docstring, I don't think there's anything else to do *if* I want to control this from docstring.

**Additional context**
N/A

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
-   autodata and autoattribute directives does not show right-hand value of
    the variable if its docstring contains ``:meta hide-value:`` in
    info-field-list.
- refs: #8022

### DIFF
--- a/tests/test_ext_autodoc.py
+++ b/tests/test_ext_autodoc.py
@@ -2235,3 +2235,49 @@ def test_name_mangling(app):
         '      name of Foo',
         '',
     ]
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6), reason='python 3.6+ is required.')
+@pytest.mark.sphinx('html', testroot='ext-autodoc')
+def test_hide_value(app):
+    options = {'members': True}
+    actual = do_autodoc(app, 'module', 'target.hide_value', options)
+    assert list(actual) == [
+        '',
+        '.. py:module:: target.hide_value',
+        '',
+        '',
+        '.. py:class:: Foo()',
+        '   :module: target.hide_value',
+        '',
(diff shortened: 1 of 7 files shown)


tests/test_ext_autodoc.py
```diff
@@ -2235,3 +2235,49 @@ def test_name_mangling(app):
...
+        '',
+        '      docstring',
+        '',
+        '      :meta hide-value:',
+        '',
+        '',
+        '   .. py:attribute:: Foo.SENTINEL2',
+        '      :module: target.hide_value',
+        '',
+        '      :meta hide-value:',
```
tk0miya, line 2267: I think this entry should not be shown because there is no document without metadata. It should be only shown if `:undoc-members:` option given.
But, that is out of topic of this PR.

tk0miya: refs: https://github.com/sphinx-doc/sphinx/issues/8597

PR #8628, opened 2020-12-31
bugfix: pickle environment after consistency check

mcmtroffaes:
This patch proposes to move the environment pickle until after the consistency check.

### Purpose

Whilst refactoring the sphinxcontrib-bibtex extension, I've found the need to write domain data to the environment during the consistency check phase. The extension needs to know the list of all cited items from all documents before it can produce the bibliography labels, to avoid clashes between labels and/or to number them appropriately and in the correct order. These labels need to be known at the post-transform phase of every document that has a citation or a bibliography, so the most sensible location to produce the label data is during the consistency check.

This normally poses no problem, however I've found that there is one special case where this data is not restored into the environment after a rebuild, namely, say when you do an html build, and a latex build right after. In that case, there will be no updated documents, so the consistency check phase will never kick in. Similarly, when redoing an html build after removing the outdir but leaving the pickled doctrees folder intact, this also gets triggered.

### Detail

- Move environment pickle until after consistency check.
- Add regression test.

### Relates

- This patch fixes the issue reported at https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/226


mcmtroffaes:
Looks like the test suite is failing on one test, apparently due to the newly added regression test using the same testroot ('basic') as some other tests. Oddly, it's not failing locally here for me (using windows). 😕 Any suggestions?

mcmtroffaes:
Not sure if this is the most elegant solution, but it looks like unlinking the file prior to the build phase fixed that particular regression test.

mcmtroffaes:
I've found this issue also gets triggered when rebuilding after removing the output folder (app.outdir) but leaving the pickled doctrees intact. Test updated accordingly.

tk0miya:
-0: I feel strange modifying env in the "check" phase. So I also feel strange about this change.
+1: I understand it is useful to hook after all of the reading processes. So it would be better to add a new event before pickling.
-0: Changing the pickling timing in a minor release. But adding a new event is okay to me.

mcmtroffaes:
Thank you so much for the quick response.

I agree, it feels out of place to change the environment in a phase called "check consistency". For the bibtex extension, as far as I can tell, unfortunately there is no other place where all data from the reading phase can be assembled for the post-transform phase to generate the correct nodes. I'm open to suggestions as to how to do this differently in the current event system. Originally, I tried to generate labels in the post-transform phase when the bibliography is generated. However, post-transforms are applied on a per-document basis, so if a document contains a citation to a bibliography in another document that has not yet been generated in post-transform, its label is not known yet. This work of figuring out the bibliography labels can only be done on a global basis across all documents, after they have been read, and not on a per-document basis as in post-transforms.

Adding a new event after reading everything, but before pickling the environment, would of course work. However, from a technical design point of view, it feels unnecessary to have two events where just one event can do. Perhaps another solution would be to rename "check-consistency" to something that does not suggest the environment should not be changed (possibly with an alias to the old "check-consistency" event for backward compatibility if so desired)?

Anyway these are my thoughts. I don't feel strongly for this to go in for 3.x. In the meantime, I can document the limitation for the bibtex extension that one should build each target from a fresh environment.  I'm happy to retarget this to the master branch and to help implement whatever solution you would prefer to see. As said, if you can see a way for this to work without changing the env in the check consistency phase, I'd be keen to find out.

mcmtroffaes:
Actually, after stepping through the debugger again just, perhaps env-updated is good point for this? I got the impression from the documentation this was just to signal which documents have been updated, but I guess it is also ok to modify the env in there?

mcmtroffaes:
I've done some tests and, indeed, env-updated works perfectly for the purpose of modifying env after reading everything, so closing this. Sorry for the noise. Happy new year!

tk0miya:
Agreed. Surely `env-updated` is a good point for your case :-)

PR #8639, opened 2021-01-01
Close #8634: html: Allow to change the order of JS/CSS

### LINKED ISSUE #8634: Provide a guarantee of the order in which CSS is included in html output

CSS uses [cascading order](https://www.w3docs.com/snippets/css/how-to-override-css-styles.html#1-cascading-order-3) where the latest stylesheet included takes precedence over earlier ones. Therefore in order for the extension authors and documentation authors to be able to control the appearance of the output, in particular if they want to override the style, a guarantee of the ordering seems necessary. However right now sphinx does not seem to provide any promise of the order in which CSS files are inserted. I am aware of at least two occasions when an extension adding a stylesheet after the user-provided custom CSS has resulted in an issue: https://github.com/jupyter/jupyter-sphinx/issues/149 and https://github.com/executablebooks/sphinx-tabs/issues/15 .

Add an explicit documentation and a guarantee of the order in which the CSS is included in the resulting pages.

In the current implementation it could be as simple as promising that the CSS is included in the following order:

- Theme
- Extensions in the order in which they are listed
- User-provided custom css

I think this should also be taken into consideration when implementing per-page html assets (see #6241, #8631)

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- `Sphinx.add_js_file()` and `Sphinx.add_css_file()` take `priority`
argument to change the order of JS/CSS files.
- refs: #8634, #8631

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -916,21 +916,21 @@ class Sphinx:
         """
         self.registry.add_post_transform(transform)
 
-    def add_javascript(self, filename: str, **kwargs: str) -> None:
+    def add_javascript(self, filename: str, **kwargs: Any) -> None:
         """An alias of :meth:`add_js_file`."""
         warnings.warn('The app.add_javascript() is deprecated. '
                       'Please use app.add_js_file() instead.',
                       RemovedInSphinx40Warning, stacklevel=2)
         self.add_js_file(filename, **kwargs)
 
-    def add_js_file(self, filename: str, **kwargs: str) -> None:
+    def add_js_file(self, filename: str, priority: int = 500, **kwargs: Any) -> None:
         """Register a JavaScript file to include in the HTML output.
 
         Add *filename* to the list of JavaScript files that the default HTML
-        template will include.  The filename must be relative to the HTML
-        static path , or a full URI with scheme.  If the keyword argument
-        ``body`` is given, its value will be added between the
-        ``<script>`` tags. Extra keyword arguments are included as
+        template will include in order of *priority* (ascending).  The filename
+        must be relative to the HTML static path , or a full URI with scheme.
+        If the keyword argument ``body`` is given, its value will be added
+        between the ``<script>`` tags. Extra keyword arguments are included as
         attributes of the ``<script>`` tag.
 
         Example::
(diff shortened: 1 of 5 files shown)


sphinx/application.py
```diff
@@ -975,6 +990,16 @@ def add_css_file(self, filename: str, **kwargs: str) -> None:
             # => <link rel="alternate stylesheet" href="_static/fancy.css"
             #          type="text/css" title="fancy" />
 
+        .. list-table:: priority range for CSS files
+           :widths: 20,80
+
+           * - Priority
+             - Main purpose in Sphinx
```
tk0miya, line 997: Note: At present, Sphinx's CSS files are embeded into theme files directly. So it is hard to control its order. I'll improve it in another PR.

choldgraf:
I think this is great 👍

akhmerov:
I also think this is wonderful: especially when combined with #8634, it is a drastic quality of life improvement—thanks a lot!

I also like that without extensions and users doing anything this impementation guarantees that extension assets always come before the user assets, unlike in the previous implementation.

I have the following remarks for consideration:

- This implementation preserves the implicit existing feature that the order of calls to `add_js_file` and `add_css_file` with equal priority is the same as the order of asset inclusion. This is guaranteed due to `list.sort` being [stable](https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts). The same holds for the user-provided assets (`html_js_files` / `html_css_files`). Should this be made into an explicit guarantee and documented?
- The feature seems to be easier to use for extension authors than for end users, since there's no interface for specifying priority in `html_js_files` and `html_css_files`. This seems to be quite alright: the advanced users may turn their `conf.py` into an extension.

tk0miya:
Oops. I overlooked @akhmerov 's last comment.

>Should this be made into an explicit guarantee and documented?

Indeed. It would be better if documented.

>The feature seems to be easier to use for extension authors than for end users, since there's no interface for specifying priority in html_js_files and html_css_files.

I did not think about the user configurations. But it is also allowed to give priority via attribute with this change:
```
html_css_files = [('custom.css', {'priority': 999}]
```

tk0miya:
I posted https://github.com/sphinx-doc/sphinx/pull/8641 for additional documentation. It would be helpful if you'll review it also (because I'm not good at English...)

sphinx/application.py
tk0miya: Note: I posted https://github.com/sphinx-doc/sphinx/pull/8643.

PR #8640, opened 2021-01-01
DOC: `modules` are only available when `recursive` is on

xuhdev:
See
https://github.com/sphinx-doc/sphinx/blob/7ecf6b88aa5ddaed552527d2ef60f1bd35e98ddc/sphinx/ext/autosummary/generate.py#L314-L315


tk0miya:
Thanks!

PR #8641, opened 2021-01-02
doc: Add explanation for the priority attribute of html_*_files

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- doc: Add explanation for the priority attribute of html_*_files
- doc: Add explanation about stable sort of html_*_files
- refs: #8639


akhmerov (approved):
Thanks, all looks good!

tk0miya:
Thank you for reviewing!

PR #8642, opened 2021-01-02
Fix incorrect Greek translation of dates

alex-tee:
There is no `de` in Greek. `02 Ιανουαρίου, 2021` would be OK for today's date


tk0miya:
Thank you for letting us know. I applied this change to transifex.com manually. Closing.

PR #8648, opened 2021-01-03
refactor: some domain base classes become subclasses of abc.ABC

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- `ObjectDescription`, `Index` and `Domain` classes become subclasses of
`abc.ABC` to indicate methods that must be overrided in the concrete
classes.

### DIFF
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -10,6 +10,7 @@
 """
 
 import copy
+from abc import ABC, abstractmethod
 from typing import (TYPE_CHECKING, Any, Callable, Dict, Iterable, List, NamedTuple, Tuple,
                     Type, Union, cast)
 
@@ -64,7 +65,7 @@ class IndexEntry(NamedTuple):
     descr: str
 
 
-class Index:
+class Index(ABC):
     """
     An Index is the description for a domain-specific index.  To add an index to
     a domain, subclass Index, overriding the three name attributes:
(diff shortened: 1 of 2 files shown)


sphinx/directives/__init__.py
```diff
@@ -45,7 +46,7 @@ def optional_int(argument: str) -> int:
         return value
 
 
-class ObjectDescription(SphinxDirective):
+class ObjectDescription(SphinxDirective, ABC):
```
tk0miya: The `describe` directive uses this class directly. Is it intended?

tk0miya: Finally, I determined not to change `ObjectDescription` to abstractclass.

sphinx/domains/__init__.py
```diff
@@ -287,10 +289,12 @@ def run(self) -> List[Node]:
 
     # methods that should be overwritten
 
+    @abstractmethod
     def clear_doc(self, docname: str) -> None:
         """Remove traces of a document in the domain-specific inventories."""
         pass
 
+    @abstractmethod
     def merge_domaindata(self, docnames: List[str], otherdata: Dict) -> None:
```
tk0miya: Note: This changes 3rd party domains must support parallel builds.

tk0miya: This is a new constraint. So far, extension developers have been able to choose not to support parallel builds. But this enforce them to implement it. It must be a breaking change. I'll revert this.

PR #8663, opened 2021-01-08
Basic theme: cast globaltoc_maxdepth to int

JeanKossaifi:
Subject: In the basic html theme, when specifying the max-depth in `theme.conf` that isn't automatically cast to int, causing a hard to locate error.

### Feature or Bugfix
- Bugfix

### Purpose
Cast maxdepth to int in the basic template.

### Detail
The globaltoc uses ``theme_globaltoc_maxdepth`` without casting it to `int`, causing errors when this is set in `theme.conf`

### DIFF
--- a/sphinx/themes/basic/globaltoc.html
+++ b/sphinx/themes/basic/globaltoc.html
@@ -8,4 +8,4 @@
     :license: BSD, see LICENSE for details.
 #}
 <h3><a href="{{ pathto(master_doc)|e }}">{{ _('Table of Contents') }}</a></h3>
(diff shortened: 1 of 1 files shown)


tk0miya:
Thank you for your contribution. Indeed, this must be a bug. I think it is better to convert the value inside `toctree()` function instead of theme. What do you think?

BTW, could you rebase this into 3.x branch? Then I'll release this soon in next minor release (maybe in mid-Feb.)

JeanKossaifi:
That's a good point, this will avoid the same issue arising with other themes. I made the change directly to 3.x in #8665.

PR #8665, opened 2021-01-09
Cast maxdepth to int in toctree

JeanKossaifi:
Subject: When specifying the toc's max-depth in theme.conf that isn't automatically cast to int, causing a hard to locate error.

### Feature or Bugfix
- Bugfix

### Relates
-  Replaces #8663


tk0miya:
Thank you for quick update! merged :-)

PR #8668, opened 2021-01-10
C, fix namespace lookup for expr role

jakobandersen:
### Feature or Bugfix
- Bugfix

### Purpose
Fix typo in C domain which messes up lookup of symbols in ``expr`` roles.


jakobandersen:
(@tk0miya, unless other people need this fix I don't think it needs to trigger a release by it self)

PR #8690, opened 2021-01-16
Fix #4550: The align attribute of figure nodes becomes None by default

### LINKED ISSUE #4550: Tables should be centered by default.
In general having all tables pushed to the left is an undesirable behaviour. 
Tables should be centred by default and the user should be able to align them the the left or right if desired.

See #3942

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To keep compatibility with the standard doctree model of docutils,
this stops to use 'default' value as a default value of the align
attribute for figure and table nodes.
- refs: #4550

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -106,7 +106,7 @@ class Table:
     def __init__(self, node: Element) -> None:
         self.header = []                        # type: List[str]
         self.body = []                          # type: List[str]
-        self.align = node.get('align')
+        self.align = node.get('align', 'default')
         self.colcount = 0
         self.colspec = None                     # type: str
         self.colwidths = []                     # type: List[int]
@@ -1240,7 +1240,6 @@ class LaTeXTranslator(SphinxTranslator):
                 (1, 'middle'): ('\\raisebox{-0.5\\height}{', '}'),
                 (1, 'bottom'): ('\\raisebox{-\\height}{', '}'),
                 (0, 'center'): ('{\\hspace*{\\fill}', '\\hspace*{\\fill}}'),
-                (0, 'default'): ('{\\hspace*{\\fill}', '\\hspace*{\\fill}}'),
                 # These 2 don't exactly do the right thing.  The image should
                 # be floated alongside the paragraph.  See
                 # https://www.w3.org/TR/html4/struct/objects.html#adef-align-IMG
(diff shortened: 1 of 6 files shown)


tk0miya:
@brechtm Could you check this please?

sphinx/writers/latex.py
```diff
@@ -1240,7 +1240,6 @@ def visit_image(self, node: Element) -> None:
                 (1, 'middle'): ('\\raisebox{-0.5\\height}{', '}'),
                 (1, 'bottom'): ('\\raisebox{-\\height}{', '}'),
                 (0, 'center'): ('{\\hspace*{\\fill}', '\\hspace*{\\fill}}'),
-                (0, 'default'): ('{\\hspace*{\\fill}', '\\hspace*{\\fill}}'),
```
brechtm, line 1243: I think this might be changing the current behavior. Need to set `attrs.setdefault('align', 'default')` and keep this line?

brechtm:
> @brechtm Could you check this please?

Looks good to me, except for the one remark above.

sphinx/writers/latex.py
tk0miya: I thought nobody sets `'default'` to the `align` attribute of the image nodes (Our transform component only sets it to the figure and table nodes!). It means this code would be never used.

brechtm: I see. So this line wasn't ever used, even before this patch?

brechtm:
One more thing that occurred to me. The code changes the doctree (adding the _align_ attribute) in-place. I'm assuming this these changes aren't visible to other builders or saved (pickled) to the doctrees dir?

sphinx/writers/latex.py
tk0miya: Yes, I think so.

tk0miya:
>One more thing that occurred to me. The code changes the doctree (adding the align attribute) in-place. I'm assuming this these changes aren't visible to other builders or saved (pickled) to the doctrees dir?

Yes. I modified the attribute only inside the HTML writers (HTML4 and HTML5). So the change is not stored into the doctrees. And it does not affect other builders (except HTML-inherited-builders).

tk0miya:
Merged.

@brechtm Thank you for reviewing!

PR #3872, opened 2017-06-17
Add latex key to configure literal blocks caption position in PDF output

jfbu:
With ``literalblockcappos=b`` added to ``'sphinxsetup'``, the captions of literal blocks are typeset after the literal block in PDF output. Default is ``t`` which says to typeset them above.

Also added ``verbatimcontinuedalign`` and ``verbatimcontinuesalign`` with possible values ``l, c, r`` for alignment of continuation hints with respect to literal block contents (relates #3792). This gives for example:

capture d ecran 2017-06-17 a 10 15 04

The ``verbatimhintsturnover`` default is now ``true`` so that at 1.7 continuation hints (internationalized) will be the default.

### Details

* In case the caption is printed after the literal block, nevertheless the internal links will be set up to point to start of literal block.

* The caption width is now wrapped to the same maximal width as for table captions: the `\LTcapwidth` parameter whose default is `4in`.

* I have considered horizontal alignment of the caption itself. This can be done indeed for literal blocks, but I dropped it because it is not easy to do the similar thing for tables:

  - for longtable, one can patch macros for alignment with respect to the table sides, but not easily alignment with respect to the line width,
  - for tabular, tabulary, one can extend caption macro to get alignment with respect to the line width, but not with respect to the table sides (as we dropped threeparttable which could be used for that but has many induced problems) except if the table width is known in advance.

  Hence I dropped horizontal alignment for caption of literal blocks.

* In future, if a global config setting is added for caption placement, this PR does it already for literal blocks, so this will facilitate implementation. For tables, it is already not too complicated thanks to the templates. For figures, some refactoring is needed first.

* The `\sphinxcaption` which is used for tables (except longtable) and with this PR for literal blocks too has been made more flexible: earlier it hard-coded reset of `\abovecaptionskip/\belowcaptionskip` to `0pt` for uniform treatement of tables, now the two are reset to values given by some user modifiable macros.

* If caption is on top the distance from baseline of caption text to frame of literal block is now same value as for table captions (`\sphinxbelowcaptionspace` whose default is `6pt = 0.5\baselineskip = \parskip`). This is slightly increase from previous setting, in order to get uniformity with table caption placement.

* More generally, the vertical spacing for literal blocks is now customizable via renewcommand of latex macros (see next).

### Memo on vertical spacing

Here are the default values for vertical spacing (in English documents):

- tables: 
   
[...]

PR #3686, opened 2017-05-01
Drop usage of `threeparttable` (LaTeX)

jfbu:
Subject: LaTeX tables with captions are currently rendered either with `longtable` or `threeparttable` (with tabular or tabulary). The `threeparttable` is thought out to allow table notes, but Sphinx never made usage of this. Besides it produces a caption of the same width of the table itself, but there is no reliable way in LaTeX2e to wrap text around a box of unknown dimensions in a fully automatized way, thus currently, the table is always (like longtable) set in its own paragraph. The user can use `:align:` docutils option to decide if located left, center or right.

Problem is that threeparttable is loaded for almost no benefits. And captions wrapped to the table width look weird compared to captions of longtable which are wrapped at a fixed maximal width of `4in` (it can be customized by length `\LTcapwidth`).

This PR drops usage of `threeparttable`, achieves exact same vertical spacing as for longtable and wraps the table caption also to `\LTcapwidth` space. (by default `4in`).

This fixes issue #3532, with code-blocks or figures in tables having their own captions. It was very difficult to fix this without dropping threeparttable, also because via the table templates the user can position the caption before or after the table, but threeparttable does not expect that `\caption` will be used for anything else than the table caption and redefines it at the start of environment.

I don't know how to best "deprecate" `threeparttable`: there can not be transition, either we use it or we don't. The user can add its usage in the table templates and then add the loading of the package via the `'preamble'` key. A priori the only difference in output for existing documents is that the table caption is not wrapped to the table width but to `\LTcapwidth`.

Not using threeparttable means all tables are now handled the same, whether by longtable or not for the caption width.

### Feature or Bugfix
It fixes bug #3532. Also, using threeparttable appears to me to have been a Sphinx bug from the start, because its most salient features such as table notes were never made used of, and tables were always rendered in their own paragraphs, so wrapping the caption to the table width did not make much sense.

Thus for me it is bug fix. But it is a change for people if they somehow hacked at the LaTeX level the threeparttable to achieve some effect. And it changes the wrapping of table captions when not using longtable. For a long time it was very difficult for Sphinx user to customize tables, table templates make it much easier now but exist for a short time. Thus people using templates in their way can still do it, perhaps only by adding loading of `threeparttable` and adding their own macros to fix issues such as #3491 as was done in #3504.

Notice also that this PR fixes some complicated issue with `\belowcaptionskip` and `\abovecaptionskip` due to `threeparttable` definitions which did not take into account that multiple `\caption` commands could be used in the environment.

[...]

Issue #161, opened 2015-01-02
table captions placement


shimizukawa:
I'd like to define if table titles (captions) are placed before or after the table.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/161
- Originally reported by: Anonymous
- Originally created at: 2009-05-12T21:06:16.385

shimizukawa:
_From Anonymous on 2009-07-03 04:36:06+00:00_

Caption's placement with figures and tables is simply determined by the \caption{...} being given before or after the includegraphics command resp. tabulary environment.

shimizukawa:
_From Georg Brandl on 2010-08-25 12:19:51+00:00_

Removing milestone: 0.6 (automated comment)

jfbu:
This is solved by merge of #3686 (which is based on the table templates from #3389, 2a37b0e). The user can position the caption after table using the templates. The ``\sphinxaftercaption`` should not be used then, and the ``\sphinxcapstartof{table}`` should stay where it is, if one wants internal hyperlink to go to top of table.

What is lacking is a config setting to do this automatically, but already user can use custom templates modifying the defaults.

gmilde:
Docutils implements this with a CSS rule in the "mandatory" style file "minimal.css":
~~~css
table.captionbelow { caption-side: bottom; }
~~~
The "special" class value can be specified globally in the [table-style](https://docutils.sourceforge.io/docs/user/config.html#table-style) setting in docutils.conf or per table with the :class: directive option, e.g.:
~~~rst
.. table:: the caption
   :class: captionbelow

   = = =
   M N O
   = = =
~~~
It would be good, if the Sphinx standard CSS themes supported this, too.
As a workaround, you can add the rule in a custom style sheet.

jdillard:
This issue seems solved so I am going to close it. I also think there should probably be a single issue to track caption placement as there is a lot of overlap https://github.com/sphinx-doc/sphinx/issues/1723

Nice CSS, I learned something new.

PR #8701, opened 2021-01-18
Add a config item to place figure captions on top in LaTeX

### LINKED ISSUE #1723: Feature request: configure placement of caption
By default captions are placed below figures, but above tables and code-blocks. I'd like the option to change this. In particular, I use code-blocks for sample input or output, and the caption to describe them, rather than as a filename, and I'd like these captions below the blocks.

jessetan:
For code/literal blocks and tables, the default caption placement is above the content. For figures, this is below the figure.
There are ways that captions in LaTeX output can be moved for code blocks (through a config item) and tables (by changing templates), but not for figures.

This PR exposes an option to move the figure caption to the top. It does not contain any tests (yet). Comments on the approach are appreciated. A [previous comment suggested this would be better to add to docutils](https://github.com/sphinx-doc/sphinx/issues/1723#issuecomment-239059458), which currently places the caption node after the image node in the doctree. I find it hard to make changes to docutils, so I've opted to do this in Sphinx, by moving the caption node above the image node in the doctree.

### Feature or Bugfix
- Feature, implements #1723

### Relates
- #161, #3686 
- #3872

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -2196,6 +2196,12 @@ These options influence LaTeX output.
 
    .. versionadded:: 3.0
 
+.. confval:: latex_figure_captions_top
+
+   If ``True``, captions for figures are placed above the figure. Defaults to ``False``.
+   
+   .. versionadded:: 3.5
+
 
 .. _text-options:
 
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -2198,8 +2198,9 @@ These options influence LaTeX output.
 
 .. confval:: latex_figure_captions_top
 
-   If ``True``, captions for figures are placed above the figure. Defaults to ``False``.
-   
+   If ``True``, captions for figures are placed above the figure.
+   Defaults to ``False``.
+
    .. versionadded:: 3.5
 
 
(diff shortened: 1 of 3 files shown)


tk0miya:
>I find it hard to make changes to docutils

Please let me know why do you think it's hard? AFAIK, nobody sends a request to docutils.

tk0miya:
Oops. I wrongly posted my comment in the middle.

My opinion for this PR is ...

* +0 for adding a new option to control it.
* -0 for adding it to Sphinx. It would be better to add it to docutils (discussion needed with them).
* -1 for adding the option only for LaTeX. It should be applied to other formats.

jessetan:
My reasons for adding it in this way is because there is also an option to control literal block caption placement for LaTeX only. However you are right, that trying to land it in docutils first is better.
Despite the hurdles of SVN and SourceForge, I've submitted a patch to docutils: https://sourceforge.net/p/docutils/patches/179/

tk0miya:
Great work! I'm in favor of your patch :-)

tk0miya:
I'm closing this with hoping a patch for docutils' will be merged. Let's think this again if rejected!

Issue #4303, opened 2017-12-15
linkcheck performance: downloading page multiple times when checking anchors


rleigh-codelibre:
### Problem
- If my sphinx documentation contains multiple links with anchors to a web page with multiple anchors, it will download the page multiple times, once per anchor to check
- This scales very badly.  If I have many hundreds or thousands of anchors (e.g. for automatically generated documentation), it might download several megabytes × the number of links.  This can end up being multiple gigabytes

#### Procedure to reproduce the problem

- create a document with links to anchors on the same web page
- run the link checker; it will fetch the page multiple times

#### Expected results

- I would suggest that the link checker could cache the anchors on webpages, so that it only downloads each page once, and only checks each link once.  It could build a dictionary of pages to check, and store the anchors as a list or dict within it?  Since we know up front which of our links have anchors, we can skip storing them when we know it's unnecessary.
- There may be other better ways of doing this; I'm not familiar with the internals of the link checker.

### Reproducible project / your project
- https://github.com/openmicroscopy/bioformats/tree/develop/docs/sphinx
- contains lots of links to https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html

### Environment info
- OS: Any
- Python version: Any
- Sphinx version: Any

AA-Turner:
cc @jayaddison

jayaddison:
I'm taking a break from development here for a little while, but had begun investigation and considered an approximate design for how to resolve this:

* We have a [`ConnectionMeasurement` helper](https://github.com/sphinx-doc/sphinx/blob/04bd0df100809de350be89b64bb85c3524867132/tests/test_builders/test_build_linkcheck.py#L104-L107) in our `linkcheck` tests that inspects the number of HTTP connections made when unit tests run.  Following test-driven-development, we could lower the target value for that, and the fixing this bug should help to confirm the results (not the only test coverage required, but a useful confirmation).
* The [`contains_anchor`](https://github.com/sphinx-doc/sphinx/blob/04bd0df100809de350be89b64bb85c3524867132/sphinx/builders/linkcheck.py#L585-L598) function within the `linkcheck` builder is fairly key to fixing this.  Ideally it should be adapted, changed or inverted somehow so that we can check for multiple anchors within a document while only parsing the document once.  Parsing should stop as soon as all anchors are found, the same way that it currently stops when it checks for a single anchor and finds it.
[...]

PR #8702, opened 2021-01-18
linkcheck: Do not check the availability of the same URL repeatedly

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- So far, linkcheck scans all of references and images from documents, and
checks them all simply.
- This collects the URL via post-transforms, and removes duplicated URLs
before checking availability.
- refs: #4303
  * Note: This does not change the behavior for the different anchors to the same page.


francoisfreitag:
Great idea :+1: 

> So far, linkcheck scans all of references and images from documents, and checks them all simply.

I believe the `linkcheck` builder remembers previously checked URLs and verifies if an URL was already checked before to re-check it. However, the current feature is sensitive to race conditions, an URL could be checked multiple times by several threads, until one of them finishes and caches the result.

The suggested approach fixes that race condition by removing the duplicate links before threads start checking them :tada:. I believe we can remove the old “cache”: https://github.com/tk0miya/sphinx/pull/6.

---
### Follow-up ideas

There are preconditions to check a link: https://github.com/sphinx-doc/sphinx/blob/3ed7590ed411bd93b26098faab4f23619cdb2267/sphinx/builders/linkcheck.py#L238-L244

Instead of queuing these links, perhaps they should be skipped by the hyperlink collector?

Similarly, the `self.to_ignore` built from the `linkcheck_ignore` config option, though these need to be logged.

Moving these checks outside of the threads would help keeping the `check_thread` function (and threads) more focused on checking URIs, which would improve the current architecture IMO.

I can make these changes in separate follow-up PRs if you agree?

tests/roots/test-linkcheck-localserver-two-links/index.rst
```diff
@@ -1,6 +1,6 @@
-.. image:: http://localhost:7777/
+.. image:: http://localhost:7777/image.png
    :target: http://localhost:7777/
 
 `weblate.org`_
 
-.. _weblate.org: http://localhost:7777/
+.. _weblate.org: http://localhost:7777/weblate
```
francoisfreitag: With all links being different, all items in the priority queue are now comparable (the uri is the second element of Hyperlink, right after the `next_check` and this change guarantees uris are unique in the queue). We can revert a1b8b1febb6ce0b7423bb78df30b0fc424c12325 :grin:

tk0miya:
>There are preconditions to check a link:

Good to me. But the skipped items should be output to the log. So we need to consider how to do that.

tests/roots/test-linkcheck-localserver-two-links/index.rst
tk0miya: I understand what this test does. Okay, let's remove (or deprecate) them all.

CHANGES
```diff
@@ -57,6 +58,7 @@ Bugs fixed
   + or ^) are used as keystrokes
 * #8629: html: A type warning for html_use_opensearch is shown twice
 * #8665: html theme: Could not override globaltoc_maxdepth in theme.conf
+* #4304: linkcheck: Do not check the availability of the same URL repeatedly
```
francoisfreitag: This entry is misleading. It implies the availability was being checked repeatedly, when in most cases it wasn’t.

Perhaps:

Fix race condition that could lead to checking the availability of the same URL twice

francoisfreitag (approved):
Otherwise LGTM.

CHANGES
tk0miya: Indeed!

tk0miya:
@francoisfreitag Thank you for reviewing!

PR #8710, opened 2021-01-20
minor typing fix

chrisjsewell:
Subject: `domain` is allowed to be `None` in `warn_missing_reference` method

### Feature or Bugfix

- Bugfix


tk0miya:
Thanks!

PR #8715, opened 2021-01-20
Fix some mark-up issues in latex.rst with "Default:" lines

jfbu:
There were indentation problems causing wrong output.


tk0miya:
LGTM

Issue #8711, opened 2021-01-20
LaTeX: backticks in code-blocks trigger latexpdf build warning (and font change) with late TeXLive 2019


jfbu:
**Describe the bug**

For users with LaTeX2e 2020-02-02 or later (i.e. final TL2019 but not early one) latexpdf build console output might get cluttered with 

```
Package textcomp Warning: Symbol \textasciigrave not provided by
(textcomp)                font family pcr in TS1 encoding.
(textcomp)                Default family used instead on input line 98.
```

for each backtick rendered as a literal. The glyph in pdf output is slightly different. Possibly looks better, actually.

**To Reproduce**

Insert this in index.rst and issue `make latexpdf`.

```rst
::

   The :py:func:`enumerate` function can be used for ...

```

**Expected behavior**
No LaTeX warnings.

**Environment info**
- OS: [e.g. Unix/Linux/Mac/Win/other with version]
- Python version: [e.g. 3.7.1]
- Sphinx version: 3.4.0, 4.0.0+/3ed7590ed
- LaTeX: TeXLive 2019 after LaTeX 2e 2020-02-02 update or TeXLive 2020.

**Additional context**

Upstream latex ticket: https://github.com/latex3/latex2e/issues/478

jfbu:
The problem is not only for code-blocks, this rst source:

```rst
We have a LaTeX problem with \`backticks'
```

triggers

```
Package textcomp Warning: Symbol \textasciigrave not provided by
(textcomp)                font family ptm in TS1 encoding.
(textcomp)                Default family used instead on input line 97.
```

if using default Sphinx "times" font for pdf output of main text.

jfbu:
The subsituted glyph looks actually better (at least for mono font). The problem is with the numerous warnings and loading of an extra tex font. Probably the reasonable way out (as LaTeX2e is not going to change back) is to switch to the TeX Gyre fonts. (might improve support for european languages, only Latin ones) This means also that we need to augment our dependencies for Linux distros and tell people to look out for "tex-gyre" or similar named package to install on their systems (Ubuntu xenial has [it](https://packages.ubuntu.com/xenial/tex-gyre).)

I am attaching 
[testasciigrave.pdf](https://github.com/sphinx-doc/sphinx/files/5842304/testasciigrave.pdf) with its latex source 
[...]

PR #8716, opened 2021-01-20
LaTeX: update default font configuration

jfbu:
This replaces times package with tgtermes and tgheros (clones of Times and
Helvetica with better LaTeX support) and the monospace font from txfonts
package (txtt). This font is better matched with Times-like fonts than
Courier clones.

The changes applies to pdflatex/platex/uplatex.

Fixes: #8711


CHANGES
```diff
@@ -6,6 +6,7 @@ Dependencies
 
 * Drop python 3.5 support
 * Drop docutils 0.12 and 0.13 support
+* LaTeX: add ``tex-gyre`` font dependency
```
jfbu, line 9: I guess this should be added to sphinx-doc/docker-ci?

bindep.txt
```diff
@@ -12,10 +12,12 @@ texlive-luatex85 [platform:rpm]
 texlive-anyfontsize [platform:rpm]
 texlive-ctablestack [platform:rpm]
 texlive-gnu-freefont [platform:rpm]
+texlive-tex-gyre [platform:rpm]
```
jfbu, line 15: I could not check what this contains; we don't need the otf versions of the fonts. On Ubuntu, there are indeed in a separate package ``fonts-texgyre``. The ``tex-gyre`` contains the TeX only fonts and LaTeX support files.

bindep.txt
```diff
@@ -12,10 +12,12 @@ texlive-luatex85 [platform:rpm]
...
 latexmk [platform:rpm]
 
 texlive-latex-recommended [platform:dpkg]
 texlive-fonts-recommended [platform:dpkg]
+tex-gyre [platform:dpkg]
```
jfbu, line 20: the docker-ci uses ``texlive-fonts-extra`` but ``tex-gyre`` is really separate.

doc/conf.py
```diff
@@ -58,20 +58,8 @@
...
-\usepackage[sc]{mathpazo}
-\usepackage[scaled]{helvet}
-\usepackage{courier}
-\substitutefont{LGR}{\rmdefault}{cmr}
-\substitutefont{LGR}{\sfdefault}{cmss}
-\substitutefont{LGR}{\ttdefault}{cmtt}
-\substitutefont{X2}{\rmdefault}{cmr}
-\substitutefont{X2}{\sfdefault}{cmss}
-\substitutefont{X2}{\ttdefault}{cmtt}
-''',
```
jfbu, line 71: we should not build Sphinx own docs with a configuration too much separated from our default one, so I streamlined this

doc/latex.rst
```diff
@@ -133,57 +133,33 @@ Keys that you may want to override include:
...
-   For ``'xelatex'`` and ``'lualatex'``, the default is to use the FreeFont
-   family: this OpenType font family supports both Cyrillic and Greek scripts
-   and is available as separate Ubuntu Xenial package ``fonts-freefont-otf``.
-   It is not necessary to install the much larger ``texlive-fonts-extra``
-   package.
-
-   ``'platex'`` (Japanese documents) engine supports individual Cyrillic and
-   Greek letters with no need of extra user set-up.
-
-   Default: ``'\\usepackage{times}'`` (or ``''`` when using a Cyrillic script)
```
jfbu, line 179: I have reorganized the documentation of ``fontpkg``, ``fontenc``, ``textgreek`` and made it globally shorter

doc/latex.rst
```diff
@@ -133,57 +133,33 @@ Keys that you may want to override include:
...
+   Font package inclusion. The default is::
+
+      r"""\usepackage{tgtermes}
+      \usepackage{tgheros}
+      \renewcommand\ttdefault{txtt}
+      """
```
jfbu, line 141: after some hesitation I have also made this default for platex/uplatex but I can't test easily. So far we had indeed same ``\usepackage{times}``. I figured if that work, the new config should too.

There is a slightly better choice than the third line which would be ``\usepackage[zerostyle=d]{newtxtt}`` which is a wrapper of direct usage of ``txtt`` (monospace font from txfonts package, included in ``texlive-fonts-recommended``), which allows to choose the shape of the digit 0 and also to scale the font if needed, but unfortunately on Ubuntu this requires the big ``texlive-fonts-extra`` (close to half a gigabyte).

doc/latex.rst
```diff
@@ -361,35 +328,40 @@ Keys that don't need to be overridden unless in special cases are:
...
-   letters such as φύσις.  Expert users may want to load the ``textalpha``
-   package with its option ``normalize-symbols``.
+``'fontsubstitution'``
+   Ignored if ``'fontenc'`` was not configured to use ``LGR`` or ``X2`` (or
+   ``T2A``).  In case ``'fontpkg'`` key is configured for usage with some
+   TeX fonts known to be available in the ``LGR`` or ``X2`` encodings, set
+   this one to be the empty string.  Else leave to its default.
 
-   .. hint::
+   Ignored with :confval:`latex_engine` other than ``'pdflatex'``.
```
jfbu, line 346: this key contains a dozen lines of LaTeX macros which where previously inserted as part of the default ``'fontpkg'`` key making it difficult for user to customize the key while keeping the feature. Besides this made its way to latex preamble even if not needed, because the macros are conditional. With the refactoring only if user has added LGR of X2 configuration for Greek resp. Cyrillic will these macros end up in latex file.

doc/latex.rst
```diff
@@ -500,19 +472,25 @@ Keys that don't need to be overridden unless in special cases are:
...
-      Due to new default font choice for ``'xelatex'`` and ``'lualatex'``
-      (FreeFont), Sphinx does ``\\fvset{fontsize=\\small}`` also with these
-      engines (and not ``\\fvset{fontsize=auto}``).
+      For ``'xelatex'`` and ``'lualatex'`` defaults to
+      ``'\\fvset{fontsize=\\small}'`` as this
+      is adapted to the relative widths of the FreeFont family.
+
+   .. versionchanged:: 4.0.0
+      Changed default for ``'pdflatex'``. Previously it was using 
+      ``'\\fvset{fontsize=\\small}'``.
```
jfbu, line 493: also ``auto`` for platex/uplatex, as they too will use txtt (the monospace font from txfonts package)

sphinx/builders/latex/__init__.py
```diff
@@ -224,6 +224,8 @@ def init_multilingual(self) -> None:
                 self.context['substitutefont'] = '\\usepackage{substitutefont}'
             else:
                 self.context['textgreek'] = ''
+            if self.context['substitutefont'] == '':
+                self.context['fontsubstitution'] = ''
```
jfbu, line 228: fontsubstitution contains code using `\substitutefont` command from `substitutefont` package; if the latter is not loaded there is no need to insert the fontsubstitution code although it would be ok because it is condiional code, but not inserting something not useful is better

sphinx/builders/latex/constants.py
```diff
@@ -11,7 +11,12 @@
 from typing import Any, Dict
 
 PDFLATEX_DEFAULT_FONTPKG = r'''
-\usepackage{times}
+\usepackage{tgtermes}
+\usepackage{tgheros}
+\renewcommand{\ttdefault}{txtt}
```
jfbu, line 16: as commented above, tgtermes and tgheros render almost 100% like what package times gave for the body font and the sans font. I hesitated about using tgcursor for the monospace font, but really Courier is very bad match to Times-like text font. So bad we had to use ``\small`` LaTeX command in verbatim because it is too wide. The ``txtt`` font is a very nice monospace font, only a bit wider than times-like clones, and even at normal size it it is still a bit narrower than the Courier at small size. So we gain a bit in code-blocks and the other advantage is that if a code-block is an section rendered via a smaller size, either from raw latex mark-up or from being in a footnote, then the size will match the one of the text, due to ``\fvset{fontsize=auto}``

sphinx/builders/latex/constants.py
```diff
@@ -142,6 +148,8 @@
         'fontenc':     ('\\usepackage{fontspec}\n'
                         '\\defaultfontfeatures[\\rmfamily,\\sffamily,\\ttfamily]{}'),
         'fontpkg':      XELATEX_DEFAULT_FONTPKG,
+        'fvset':        '\\fvset{fontsize=\\small}',
```
jfbu, line 151: GNU FreeFont need us to use the ``\small`` because they behave a bit like Courier vs Times

sphinx/builders/latex/constants.py
```diff
@@ -161,7 +171,8 @@
         'latex_engine': 'platex',
         'babel':        '',
         'classoptions': ',dvipdfmx',
-        'fontpkg':      '\\usepackage{times}',
+        'fontpkg':      PDFLATEX_DEFAULT_FONTPKG,
+        'fontsubstitution': '',
```
jfbu, line 175: I am hoping that this is ok for platex: it does not need font substitution to allow individual Greek letters (if I recall correctly from testing 2 years ago)

sphinx/builders/latex/constants.py
```diff
@@ -170,7 +181,8 @@
         'latex_engine': 'uplatex',
         'babel':        '',
         'classoptions': ',dvipdfmx',
-        'fontpkg':      '\\usepackage{times}',
+        'fontpkg':      PDFLATEX_DEFAULT_FONTPKG,
+        'fontsubstitution': '',
```
jfbu, line 185: no idea if this is right thing to do with uplatex

bindep.txt
tk0miya: I found the files on the package for CentOS8:
https://centos.pkgs.org/8/centos-appstream-x86_64/texlive-tex-gyre-20180414-19.el8.noarch.rpm.html

CHANGES
tk0miya: Yes, please.

tk0miya (approved):
LGTM!

CHANGES
jfbu: Done PR at https://github.com/sphinx-doc/docker-ci/pull/5

bindep.txt
jfbu: Thanks, this contains indeed the needed files (such as tgtermes.sty, t1qtm.fd which are LaTeX support files, and the actual font files in TeX format .tfm or for inclusion in pdf .pfb extension)

tk0miya:
Oops. It seems conflicted. Could you update this and merge into 3.x?

jfbu:
Sure, but do you mean "merge into master"?

tk0miya:
Oops, you're right.

jfbu:
Thanks for reviewing! will merge into master after testing completes.

PR #8723, opened 2021-01-21
Deprecate linkcheck builder {broken,good,redirected}

francoisfreitag:
### Feature or Bugfix
- Refactoring

### Purpose
These attributes were used to cache checked links and avoid issuing
another web request to the same URI.

Since 82ef497a8c88f0f6e50d84520e7276bfbf65025d, links are pre-processed
to ensure uniqueness. This caching the results of checked links is no
longer useful.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,9 @@ Incompatible changes
 Deprecated
 ----------
 
+* ``sphinx.builders.linkcheck.CheckExternalLinksBuilder.broken``
+* ``sphinx.builders.linkcheck.CheckExternalLinksBuilder.good``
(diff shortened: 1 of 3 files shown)


tk0miya:
Thank you for being patient. I feel such refactoring is very boring and tired. But it sometimes saves our users from breaks of 3rd party extensions and hacks.

tk0miya (approved):

francoisfreitag:
It’s actually really good to care about users. A smooth upgrade procedure sure encourages users to run on recent versions of the software. I don’t have these constraints at my day job, but they totally make sense for such a big open-source project.
Thanks for keeping an eye out!

PR #8732, opened 2021-01-23
refactor: html theme: Insert documentation_options.js via script_files

tk0miya:
### Feature or Bugfix
- Refactoring


davidism:
This broke the pallets-sphinx-themes used by Flask, etc. because we were working around previous dirhtml behavior by rendering `DOCUMENTATION_OPTIONS.URL_ROOT = '{{ url_root }}'` in our layout template, and this removed the `url_root` variable so we were setting an empty string.

Sphinx seems to have fixed search links for the dirhtml builder, which is what we were working around, so we can just remove that line now, not a problem.

PR #8736, opened 2021-01-23
Fix #8735

### LINKED ISSUE #8735: LaTeX: wrong internal links in pdf to captioned code-blocks when numfig is not True
**Describe the bug**

Captioned code-blocks in distinct sections which are numbered the same confuse the internal links inside pdf output.

**To Reproduce**

Use  `sphinx-quickstart` to create fresh project then use following `index.rst` file
```rst

Section One
-----------

This link: :ref:`last` fails to jump to last code-block in pdf output
(if ``numfig`` not set).

.. code-block:: none
   :caption: first

   a

.. code-block:: none
   :caption: second
   :name: II

   b


.. code-block:: none
   :caption: third

   c

Section Two
-----------

.. code-block:: none
   :caption: last one
   :name: last

   d

.. See also :numref:`II`.

The pdf link :ref:`last` brings us back to first code-block in first section.


```

Execute `make latexpdf` and open pdf output in some viewer. The internal hyperlinks are wrong. Also the latex log file contains following warnings:

```
Chapter 2.
pdfTeX warning (ext4): destination with the same identifier (name{literalblock.
1}) has been already used, duplicate ignored
```

**Expected behavior**

Internal links in pdf work.

**Environment info**
- Python version: 3.6.8 (CPython)
- Sphinx version: tested with v2.0.0 as well as current 3.x
- Extra tools: TeXLive 2020

**Additional context**
With ``numfig = True`` in  `conf.py`, the issue does not show.

jfbu:
It is hard to provide a unit test, because the problem only raises a
LaTeX warning and shows as broken internal links in pdf output.

As the fix moves some LaTeX code to the end of preamble, user or
extension code may have to be moved there too.  However, a large part of
it already was delayed.  I checked that sphinxmessages.tex does not
introduce preamble code which would be broken from missing macros due
to this change.

### Feature or Bugfix

- Bugfix

Closes: #8735

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -81,6 +81,8 @@ Bugs fixed
 * #8683: :confval:`html_last_updated_fmt` generates wrong time zone for %Z
 * #1112: ``download`` role creates duplicated copies when relative path is
   specified
+* #8735: LaTeX: wrong internal links in pdf to captioned code-blocks when
+  :confval:`numfig` is not True
 
 Testing
 --------
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -6,7 +6,7 @@
 %
 
 \NeedsTeXFormat{LaTeX2e}[1995/12/01]
-\ProvidesPackage{sphinx}[2019/09/02 v2.3.0 LaTeX package (Sphinx markup)]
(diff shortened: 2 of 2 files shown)


tk0miya:
Great work!

PR #8746, opened 2021-01-23
Improve Sphinx own pdf docs

jfbu:
Prior to this the depth of the bookmarks is too low, one does not even
get to see for example what are the built-in extensions offered by
Sphinx.

Similarly the LaTeX created table of contents has not enough depth.
Dozens of contiguous pages from our documentation get only a single
link, it is very hard for newcomer to get some feeling of the scope of
Sphinx.  With a more detailed table of contents (be it inside the PDF or
via the collapsable bookmark panel of PDF viewer) learning Sphinx is
easier.

### Feature or Bugfix

- Better own docs

### DIFF
--- a/doc/changes.rst
+++ b/doc/changes.rst
@@ -8,4 +8,9 @@
 Changelog
 =========
 
+.. raw:: latex
+
+   \hypersetup{bookmarksdepth=1}% pdf bookmarks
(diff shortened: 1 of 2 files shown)


tk0miya (approved):
+1

PR #8753, opened 2021-01-24
Linkcheck: Derive number of links from the post-transform result

francoisfreitag:
### Feature or Bugfix
- Refactoring

### Purpose
The number of links to check is the number of links in self.hyperlinks,
populated by the post-transform.


tk0miya (approved):
LGTM!

Issue #8253, opened 2020-09-29
Latex: Figures with no size defined get overscaled


Blendify:
**Describe the bug**
When generating a latex/pdf output figures get scaled up too much if the they have no size defined. Instead the image should only be scaled to its natural size (the size in px from the image file) or scaled to the page width if the image file size is too big.

**To Reproduce**
Rst file
```
.. figure:: /some/image.png

   Caption.

.. figure:: /some/image.png
  :width: 200px

   Caption.
```

image.png is 200px wide.

**Expected behavior**
The results should be the same for the two images

**Your project**
https://svn.blender.org/svnroot/bf-manual/trunk/blender_docs/

**Screenshots**
![image](https://user-images.githubusercontent.com/15183467/94505468-25e18600-01d9-11eb-9f96-9c95c07f035d.png)

**Environment info**
- OS: Ubuntu
- Python version: 3.8
- Sphinx version: 3.2.1
- Sphinx extensions: none
- Extra tools: ``texlive-xetex`` ``latexmk``

jfbu:
Sorry for late, will look into it.

jfbu:
This is at any rate inherited from LaTeX behaviour. I downloaded [your sample image](https://svn.blender.org/svnroot/bf-manual/trunk/blender_docs/manual/images/render_shader-nodes_input_volume-info_node.png) and tried it out in direct latex document this way

```latex
\documentclass{article}
\usepackage{graphicx}
\begin{document}

\noindent\includegraphics{render_shader-nodes_input_volume-info_node.png}

\end{document}
```

this renders exactly at same size as using ``figure`` directive as in your index.rst. The latex log file contains

```
<render_shader-nodes_input_volume-info_node.png, id=1, 223.83624pt x 199.74625p
t>
File: render_shader-nodes_input_volume-info_node.png Graphic file (type png)
<use render_shader-nodes_input_volume-info_node.png>
Package pdftex.def Info: render_shader-nodes_input_volume-info_node.png  used o
n input line 5.
(pdftex.def)             Requested size: 223.8357pt x 199.74576pt.

```

If I open this image file in Preview.app on my mac it tells me the dimensions are 223 x 199 pixels, with dpi of 72px/inch.

[...]

PR #8760, opened 2021-01-24
LaTeX: sync pdftex engine default imageresolution with pxunit

jfbu:
Closes: #8253

The 'pxunit' key from latex_elements instructs how to handle image
dimensions specified in px units.

But pdftex has \pdfimageresolution which is used when an image file does
not provide readable or legit values for the x and/or y resolution.

This commit syncs them: from 'pxunit' the default image resolution in
pixels per inch (an integer) is computed.

This way an image will behave the same if:

- it is loaded with no explicit size set, _and_ no readable image
  resolution data is readable from the file (or that data matches the
  'pxunit' setting)

- or a size is set in figure directive using px units and equal to the
  natural pixel size of the image,

This also with 'lualatex' but is ignored by with 'xelatex' and
'uplatex'.

### Relates
- #8253 

It is hard to provide unit test, because one would have to examine logs of a latex build, and the data inthere might change according to latex upgrades.

This will change the size of images loaded with no explicit dimension set, if the graphics file has no x-resolution or no such resolution understood by pdftex. For this reason I make this PR for merge into master.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -68,6 +68,8 @@ Bugs fixed
   ``:unknown:doc:``)
 * #8711: LaTeX: backticks in code-blocks trigger latexpdf build warning (and font
   change) with late TeXLive 2019
+* #8253: LaTeX: Figures with no size defined get overscaled (compared to images
+  with size explicitly set in pixels) (fixed for ``'pdflatex'/'lualatex'`` only)
 
 Testing
 --------
--- a/sphinx/templates/latex/latex.tex_t
+++ b/sphinx/templates/latex/latex.tex_t
@@ -13,6 +13,9 @@
 \ifdefined\pdfpxdimen
    \let\sphinxpxdimen\pdfpxdimen\else\newdimen\sphinxpxdimen
 \fi \sphinxpxdimen=<%= pxunit %>\relax
+\ifdefined\pdfimageresolution
+    \pdfimageresolution= \numexpr \dimexpr1in\relax/\sphinxpxdimen\relax
+\fi
 <% if use_xindy -%>
 %% turn off hyperref patch of \index as sphinx.xdy xindy module takes care of
 %% suitable \hyperpage mark-up, working around hyperref-xindy incompatibility


jfbu:
For testing this change, one can use this in index.rst:
```rst
.. figure:: image.png

   Caption

.. figure:: image.png
   :width: 223px

   Caption
```
where ``223px`` stands for the natural width of ``image.png``. I used [this one](https://svn.blender.org/svnroot/bf-manual/trunk/blender_docs/manual/images/render_shader-nodes_input_volume-info_node.png) 
render_shader-nodes_input_volume-info_node
If testing with another image.png, replace `223px` by its natural width.

After `make latexpdf` the two images should be same size in pdf, whatever the value of `latex_elements['pxunit']`.

jfbu:
Notice that this is breaking change for png images not providing the x-resolution/y-resolution data (which is in [pHYs data chunk](http://libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.pHYs) although the latter may serve only to specify a deviation of pixel shape from square one, and only optionally uses the meter absolute unit)

When included without any size specification and assuming `'pxunit'` is not changed, they will be rendered at 96ppi compared to previously 72ppi, i.e. they are rendered smaller. This applies for example to `more.png` file in Sphinx own docs ([getting started](https://www.sphinx-doc.org/en/master/usage/quickstart.html) section). In this specific case, this looks like a bugfix, the image size fits better with the text baseline spread in the pdf output.

Using ImageMagick convert utility on unixen, one can set the image resolution in a png file this way:
```
convert -units PixelsPerInch image.png -density 192 image192dpi.png 
```
allowing to check that when a png file does have the pixels per inch (or rather pixels per meter in the [pHYs data chunk](http://libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.pHYs)), using the figure directive with no size spec will have pdftex insert the image in accordance to the pixel resolution set in the file, independently of the setting of `\pdfimageresolution` from Sphinx (according to ``'pxunit'``).

jfbu:
To recap issue #8253 is caused by  a number of factors:
- the png file has no xresolution data understood by pdftex; pdftex then uses a default that there is 72 pixels to an inch
- when one specifies explicitly a width in px, Sphinx uses a default of 96 pixels to an inch
- the actual sample image provided had a pixel width of 223 not 200 (but this is an unrelated error)

It is possible, at least for pdflatex and lualatex, to set the default it uses for images whose data chunks contain no resolution indication.  This pull request does this in a manner that makes 1px used in a width or height specification match with the default dpi used for images with no resolution data.

This means that from then on an image file will behave the same, if it has no resolution data, whether included with no size specification or included with a target width expressed in px and matching its natural x-size in px. This will work independently of the `latex_elements['pxunit']` setting but the image width in pdf will depend on this setting (because this setting also will help tell pdftex what to use as unspecified pixel per inch resolution).

Nothing changes for
- images loaded with specific target size, whether using or not px unit, whether or not their file data contains a pixel per inch (or per meter) spec,
- images loaded with no specific target size, but having among their data a specified resolution which is understood by pdftex; the inclusion will then match that specified resolution and the resulting width is independent on how Sphinx is configured to handle px units; an image of natural width 300px and resolution of 100ppi ends up using 3in of horizontal space in pdf, if it is loaded with no specified target width, and this is independent of 'pxunit' setting. It behave that way prior to that PR and will still behave this way.

I don't know how to achieve this with xelatex and uplatex

tk0miya (approved):

PR #8767, opened 2021-01-26
Fix #8214: duplicate entries in latex index if term also in glossary

### LINKED ISSUE #8214: The :index: role and the glossary generate duplicate entries in the LaTeX index
**Bug**
A document contains both : 
- a glossary with an entry named "term"
- some text, in which I index an occurrence of the word "term" through the role `` :index:`term` ``

The output PDF index contains a duplicate entry for the word "term":
- one entry with a page number of the glossary entry
- one entry with a page number of the occurrence in the text of the document
Below is a snapshot :
image

**Expected behavior**
The index should contain only one entry for the word "term", with 2 page numbers pointing to the occurrences of that word.
Expected output index :
image

**Possible cause**
Looking at the generated `.idx` file:
```
\indexentry{term@\spxentry{term}|hyperpage}{1}
\indexentry{term@\spxentry {term}|hyperindexformat{\spxpagem}}{3}
```
An extra whitespace between the command `\spxentry` and the argument `{term}` is generated from the glossary entry. 
Removing that extra whitespace solves the problem.

**Environment info**
- OS: Mac 10.13.6
- Python version: 3.8.5
- Sphinx version: 3.2.1
- Sphinx extensions:  none
- Extra tools: none

jfbu:
This makes .idx files contain \spxentry{..} with no space, whether or
not the \index latex command is encountered in main text or e.g. in a
label of a description list (like happens for terms from a glossary).

Xindy does not care about spaces, but Makeindex doesn't merge entries
whose typesetting is to be done via \spxentry{..} or \spxentry {..}.

An alternative work-around would be for Sphinx LaTeX writer to use some
wrapper, say, \sphinxindexwrap, only to fetch the argument and force TeX
tokenization and then hand over it to \index command.

But we choose to let \spxentry expand to its own name, with no trailing
space (it gets its final definition only from the python.ist file in
case of Makeindex).

In case both the :index: role and the glossary are on same page, Xindy
and Makeindex behave differently: the former gives only once the page
number, the latter issues them twice (but the indexed term only once),
and there is warning in the .ilg file. But this is unrelated and we
can't do here anything about it.

### Feature or Bugfix
- Bugfix

Closes: #8214

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -98,6 +98,8 @@ Bugs fixed
 * #8683: :confval:`html_last_updated_fmt` generates wrong time zone for %Z
 * #1112: ``download`` role creates duplicated copies when relative path is
   specified
+* #8214: LaTeX: The :rst:role:`index` role and the glossary generate duplicate
+  entries in the LaTeX index (if both used for same term)
 * #8735: LaTeX: wrong internal links in pdf to captioned code-blocks when
   :confval:`numfig` is not True
 * #8442: LaTeX: some indexed terms are ignored when using xelatex engine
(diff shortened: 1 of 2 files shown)


tk0miya (approved):
Confirmed. LGTM!

PR #8770, opened 2021-01-27
linkcheck: Raise on unknown status in process_result()

francoisfreitag:
### Feature or Bugfix
- Feature
- Refactoring

### Purpose
Helps catching programming errors.

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -428,6 +428,8 @@ class CheckExternalLinksBuilder(DummyBuilder):
             self.write_entry('redirected ' + text, docname, filename,
                              lineno, uri + ' to ' + info)
             self.write_linkstat(linkstat)
+        else:
+            raise ValueError("Unknown status %s." % status)
 
     def write_entry(self, what: str, docname: str, filename: str, line: int,
                     uri: str) -> None:


CHANGES
```diff
@@ -10,6 +10,9 @@ Incompatible changes
 * Update Underscore.js to 1.12.0
 * #6550: html: The config variable ``html_add_permalinks`` is replaced by
   :confval:`html_permalinks` and :confval:`html_permalinks_icon`
+* linkcheck: Submitting an unknown status to
+  :py:meth:`~sphinx.builders.linkcheck.CheckExternalLinksBuilder.process_result`
+  raises a :exc:`ValueError` instead of silently ignoring the unknown status.
```
francoisfreitag: I’m not sure where this log should be located. Extensions could theoretically handle their custom status before to call the parent implementation, but this may cause a breakage.

In some way, this is a feature, it helps detect programming mistakes.

tk0miya: I suppose this patch adds a sentinel condition that is never reached, right? If true, it does not affect users and developers. then no CHANGES entry is needed to me.

francoisfreitag: Thanks for the quick answer! That’s true, the condition should never be reached.

tk0miya (approved):

tk0miya:
LGTM!

PR #8772, opened 2021-01-27
Revert alteration at c9480f994 merge of 3.x of doc/conf.py

jfbu:
Compare doc/conf.py after merge at c9480f994 to what it was at 2ee033838.

It loses the modification from #8716 (merged at 38c614347) and thus
reverts doc/conf.py to former font config using mathpazo.

- Bugfix


jfbu:
doc/conf.py was altered by other merged PRs into 3.x, and merging this into master seems to have silently overwritten the doc/conf.py modification from PR #8716 which was merged earlier directly into master.

I will merge after testing completes.

PR #8774, opened 2021-01-27
linkcheck: Skip links to ignore in post transform

francoisfreitag:
### Feature or Bugfix
- Refactoring

### Purpose
Avoid passing the links to the checker threads when it should not be
checked.

The log files are now written by the link checker and the post transform,
share the file object.

The check results structure are now used by the link checker and the post
transform, formalize its structure in a named tuple.

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -44,6 +44,12 @@ Hyperlink = NamedTuple('Hyperlink', (('next_check', float),
                                      ('uri', Optional[str]),
                                      ('docname', Optional[str]),
                                      ('lineno', Optional[int])))
+CheckResult = NamedTuple('CheckResult', (('uri', str),
+                                         ('docname', str),
+                                         ('lineno', int),
+                                         ('status', str),
+                                         ('message', str),
+                                         ('code', int)))
 RateLimit = NamedTuple('RateLimit', (('delay', float), ('next_check', float)))
 
 DEFAULT_REQUEST_HEADERS = {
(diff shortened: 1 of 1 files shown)


francoisfreitag:
In order for tests to pass, the commit from #8773 must be included in this PR.
I’ll rebase on 3.x when #8773 lands.

sphinx/builders/linkcheck.py
```diff
@@ -469,25 +472,31 @@ class HyperlinkCollector(SphinxPostTransform):
     def run(self, **kwargs: Any) -> None:
         builder = cast(CheckExternalLinksBuilder, self.app.builder)
         hyperlinks = builder.hyperlinks
+        to_ignore = [re.compile(x) for x in self.app.config.linkcheck_ignore]
+
+        def add_hyperlink(uri: str, lineno: int) -> None:
+            docname = self.env.docname
+            if any(ignore_re.match(uri) for ignore_re in to_ignore):
+                builder.process_result(CheckResult(uri, docname, lineno, 'ignored', '', 0))
```
tk0miya, line 480: I feel `HyperlinkCollector` and the builder are strongly coupled. It would be better to separate collecting and processing. How about keeping this transform as a collector of hyperlinks?

We can do the same thing on `CheckExternalLinksBuilder.final()`.
```
def final(self):
    ...

    for hyperlink in self.hyperlinks.values():
        if any(ignore_re.match(hyperlink.uri) for ignore_re in to_ignore):
            self.process_result(hyperlink)
        else:
            self.wqueue.put(hyperlink, False)
```

francoisfreitag:
Superseded by #8792.

PR #8779, opened 2021-01-28
Fix #8072: Directive hlist not implemented in LaTeX

### LINKED ISSUE #8072: Directive hlist not implemented in LaTeX
**Describe the bug**
According to the documentation [hlist](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-hlist) must contain a bullet list. It will transform it into a more compact list by either distributing more than one item horizontally, or reducing spacing between items, depending on the builder.

It works `so-so` in html, but it is not implemented in LaTeX:

```python
    def visit_hlist(self, node: Element) -> None:
        # for now, we don't support a more compact list format
        # don't add individual itemize environments, but one for all columns
```

**Improvement?**

I think it could be implemented using a `multicol` environment with this preamble:

```latex
\usepackage{multicol}
\newcommand{\fixspacing}{\vspace{0pt plus 1filll}\mbox{}}
```

I wrote a small example. However I did not find how to inject some headers in the LaTeX document. Modifying `app.builder.env.config` doesn't seem to work.

```python
import re
from sphinx.writers.latex import LaTeXTranslator

def visit_hlist(self, node):
    self.compact_list += 1
    columns = len(node.children)
    self.body.append('\\begin{multicols*}{%d}' % columns)
    self.body.append('\\begin{itemize}\\setlength{\\itemsep}{0pt}'
                        '\\setlength{\\parskip}{0pt}\n')
    if self.table:
        self.table.has_problematic = True


def depart_hlist(self, node):
    self.compact_list -= 1
    self.body.append('\\end{itemize}\n')
    self.body.append('\\fixspacing\n\\end{multicols*}\n')


def inject_packages(app):
    """TODO: Not Working..."""
    config = app.builder.env.config
    to_add = '%% For hlist directive\n'
    if not re.findall(r'\\usepackage(\[.*?\])?\{multicol\}',
                      config.latex_elements['preamble']):
        to_add += '\n\\usepackage{multicol}\n'
    to_add += r'\newcommand{\fixspacing}{\vspace{0pt plus 1filll}\mbox{}}'
    to_add += '\n'
    config.latex_elements['preamble'] += to_add


def setup(app):
    LaTeXTranslator.visit_hlist = visit_hlist
    LaTeXTranslator.depart_hlist = depart_hlist

    app.connect('builder-inited', inject_packages)

    return {
        'version': '0.1',
        'parallel_read_safe': True,
        'parallel_write_safe': True,
    }
```

jfbu:
Adds ``multicol`` LaTeX package requirement, but it is a required
part of any latex distribution.

(hence I make the PR on 3.x branch)

I think this is bugfix as #8072 was tagged as "bug" :)

I used following test file 
[index.rst.txt](https://github.com/sphinx-doc/sphinx/files/5889808/index.rst.txt)
to check it was possible to use `hlist` inside a table, or inside another list (I checked only a `hlist`, but this is only more stringent).

In the test output one sees some phenomenon related to the fact that LaTeX does not hyphenate the first word of a paragraph. This shows here with the word "horizontally" in narrow columns. The non-hyphenated word gets then shifted down...

But this has nothing to do with this PR and I will make another one to address the issue. The bullets in the output may look random but they do match the input which I constructed by random copy paste and adding and suppressing item separators.

Capture d’écran 2021-01-28 à 21 24 38

### Relates
- #8072

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -394,6 +394,8 @@ units as well as normal text.
 
    .. versionadded:: 0.6
 
+   .. versionchanged:: 3.5.0
+      support by the latex builder.
 
 .. _code-examples:
 
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -394,8 +394,6 @@ units as well as normal text.
 
    .. versionadded:: 0.6
 
-   .. versionchanged:: 3.5.0
-      support by the latex builder.
 
 .. _code-examples:
 
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,9 @@ Release 3.5.0 (in development)
 Dependencies
 ------------
 
+* LaTeX: ``multicol`` (it is anyhow a required part of the official latex2e
+  base distribution)
+
 Incompatible changes
 --------------------
 
@@ -103,6 +106,7 @@ Bugs fixed
   specified
 * #7576: LaTeX with French babel and memoir crash: "Illegal parameter number
   in definition of ``\FNH@prefntext``"
+* #8072: LaTeX: Directive :rst:dir:`hlist` not implemented in LaTeX
 * #8214: LaTeX: The :rst:role:`index` role and the glossary generate duplicate
   entries in the LaTeX index (if both used for same term)
 * #8735: LaTeX: wrong internal links in pdf to captioned code-blocks when
--- a/sphinx/directives/other.py
+++ b/sphinx/directives/other.py
@@ -276,6 +276,7 @@ class HList(SphinxDirective):
         npercol, nmore = divmod(len(fulllist), ncolumns)
         index = 0
         newnode = addnodes.hlist()
+        newnode['ncolumns'] = str(ncolumns)
         for column in range(ncolumns):
             endindex = index + ((npercol + 1) if column < nmore else npercol)
             bullet_list = nodes.bullet_list()
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -239,6 +239,8 @@
 \ltx@ifundefined{@removefromreset}
     {\RequirePackage{remreset}}
     {}% avoid warning
+% To support hlist directive
+\RequirePackage{multicol}
 % to make pdf with correct encoded bookmarks in Japanese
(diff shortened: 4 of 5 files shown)


doc/usage/restructuredtext/directives.rst
```diff
@@ -394,6 +394,8 @@ units as well as normal text.
 
    .. versionadded:: 0.6
 
+   .. versionchanged:: 3.5.0
+      support by the latex builder.
```
jfbu: Should this be 3.5 not 3.5.0? And should I capitalize: "Support", not "support"?

jfbu:
LaTeX (except lualatex engine) does not hyphenate first word of a paragraph (see #8780). So with this PR, the input

```
.. hlist::
   :columns: 6

   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
   * horizontally
```

currently renders as (a4paper):
Capture d’écran 2021-01-29 à 13 56 35

With #8781 merged it will render rather as:
Capture d’écran 2021-01-29 à 13 57 26

However, if in this example `horizontally` is replaced by a word that TeX does not know how to hyphenate such as `aaaaaaaaaaaa`, then, even with #8781, the downward shift seen in first picture above will occur. I have no fix for that at this time.

doc/usage/restructuredtext/directives.rst
tk0miya: Both are okay, but I usually use 3.5 and "Support" (no reason and rules :-p)

tk0miya (approved):
LGTM!

doc/usage/restructuredtext/directives.rst
tk0miya: Ah, I did not noticed this is a changes of `hlist` directive. I don't think this changelog is needed because of this change does not modify the feature of the directive itself.

jfbu:
Thanks for review @tk0miya as always. I tested that the used LaTeX  `multicols` environment (from `multicol` package) can be used with no problem in nested lists. One can even nest an `hlist` inside another `hlist`. It is ok also inside table cells (triggering the `table.has_problematic` like other lists). Thus I feel this is not breaking change but new feature, hence now merging!

Issue #3042, opened 2016-10-13
LaTeX renders a table as collapsed by default


tk0miya:
It seems LaTeX renders a table as collapsed.

Input:

```
========= ======== =======================================================
Tag       Required Description
========= ======== =======================================================
columns   No       Columns of the content to return. Each content listing
                   has a set of defaults columns that make sense for that
                   particular content type. By default, these columns are
                   returned in addition to what was requested here.
xquery    No       A  xquery
filter    No       Return files containing the specified filter string
========= ======== =======================================================
```

Result: (with Sphinx-1.4.8, MacTeX 2016, PDFLaTeX)
![2016-10-13 11 51 06](https://cloud.githubusercontent.com/assets/748828/19335081/7071512c-913b-11e6-9b65-8044f07d65ec.png)

Originally, this is reported at sphinx-users list:
https://groups.google.com/forum/#!topic/sphinx-users/7-T1y1Y5JJM

jfbu:
According to tabulary manual, user can issue in LaTeX document

```
\setlength{\tymin}{1.5cm}
```

for example to avoid narrow columns being squeezd this way. This can be inserted globally via `'preamble'` key or locally using 

``` rst
.. raw:: latex

   \begingroup\setlength{\tymin}{1.5cm}%
   \relax

<table spec>

.. raw:: latex

   \endgroup
```

See https://github.com/sphinx-doc/sphinx/issues/3043#issuecomment-253419012 for more convenient way, via specification of tabular columns.

tk0miya:
Thanks,
It seems there are no better setting to adjust size of columns automatically. (at least, while we use tabulary)

jfbu:
the matter with tables in LaTeX is complicated. I will not have time to review the situation in time for Sphinx-1.5.

jfbu:
Unfortunately, this will not be fixed from #8781, because `tabulary` already by default hyphenates first word in table cell. (but #8781
would improve with `tabular` and `longtable`)

Indeed (not surprisingly?) `xquery` has no hyphenation point for TeX and (surprisingly?) `columns` does not have either, but `column` has:

```
$ etex
This is pdfTeX, Version 3.14159265-2.6-1.40.21 (TeX Live 2020) (preloaded format=etex)
 restricted \write18 enabled.
**\relax
entering extended mode

*\showhyphens{xquery}

Underfull \hbox (badness 10000) detected at line 0
[] \tenrm xquery

*\showhyphens{columns}

[...]

PR #8781, opened 2021-01-28
Fix #8780: long words in narrow columns may not be hyphenated

### LINKED ISSUE #8780: LaTeX: long words in narrow columns may not be hyphenated
**Describe the bug**

<strike>For many years, LaTeX had a problem that the first word in a table cell could not be hyphenated. Recent LaTeX relases have fixed that but it remains a general fact that TeX does not (without extra mark-up) hyphenate the first word of a paragraph.</strike> 

Edited:
I don't know where from I got the idea that recent LaTeX hyphenates the first word of a table cell without help, this is wrong.

Generally speaking TeX does not hyphenate the first word of a paragraph. This may be a problem in narrow table cells. 

Package tabulary does it automatically (this shows in the picture below for the first table).  But:

- This is not the case for tabular and longtable (as one can test).
- Even tabulary can have a problem if the cell content is a list, as one sees in the second table below.
- They are are other circumstances than table cells than can lead to narrow widths.

**To Reproduce**

Use following
```

Test non hyphenated first word
------------------------------

.. csv-table:: hyphenation works with *recent* LaTeX for first word in table cell

   horizontally, horizontally, horizontally, horizontally, horizontally, horizontally, horizontally, horizontally  

.. tabularcolumns:: |p{1cm}|p{2cm}|

.. table:: but there is still problem if the cell contains a list

   ==== ====
    A    B
   ==== ====
    1    - horizontally
         - horizontally
         - horizontally
         - horizontally
         - horizontally
         - horizontally
         - horizontally
         - horizontally  
   ==== ====

```

Here is the output with recent LaTeX (<strike>with LaTeX from a few years ago also table 1 will have a problem</strike> edited, that was wrong comment; the table 1 is fine only because rendered by tabulary):

Capture d’écran 2021-01-28 à 21 48 36

In the table 2, the non hyphenated word also causes a vertical shift downward.

**Expected behavior**
First word of paragraph hyphenates if in narrow context such as a table. (other contexts can give this)

**Environment info**
- OS: [e.g. Unix/Linux/Mac/Win/other with version]
- Python version: [e.g. 3.7.1]
- Sphinx version: 3.x
- Extra tools: TeXLive 2020

**Additional context**

With the support of hlist directive by PR #8779, this problem is more likely to show (because it will not benefit from the LaTeX special fix for tables). In fact I became aware of it from testing #8779. I had already encountered the problem in Sphinx own docs, which I fixed by manual insertion (see the raw latex inside file https://github.com/sphinx-doc/sphinx/blob/57126d9ee2400ecd584565a262c909e4adc156f0/doc/extdev/deprecated.rst)

jfbu:
This is done by adding '\hskip0pt\relax\n' whenever a paragraph starts.

### Relates
- #8780, #8779 

*edited* I am removing a screenshot which contained an inaccurate assertion. See updated screenshot in next comment https://github.com/sphinx-doc/sphinx/pull/8781#issuecomment-769717265

As explained (in an edit) at #8780 the hyphenation in the first table worked because it used `tabulary`. The second table also uses `tabulary` but with the cell containing a list, it fails without extra intervention.

Tabular and longtable do not hyphenate without intervention the first word of a paragraph in a table cell.

This phenomenon is not related to tables, but to narrow width. So it will also occur with #8779 adding support for `hlist`'s in LaTeX, which a bit like table have columns of sometimes narrow width.

This being said, I am hesitant about this PR. It seems costly to add the `\hskip0pt\relax` at start of each paragraph (I didn't use `\hspace{0pt}` for reasons of needing to get rid of influence of space token in source after it, possibly, in case of an item list for example this can cause extra whitespace). I had to change many test files to reflect the change.

Also, for this PR to be safe it must be certain that the `\hskip0pt\relax` is really followed always with words: if for example it is first in a list item and then comes a list, that nested list will suffer a vertical shift.

Besides, perhaps we should not merge this at all but only mention the work-around that I had used to fix a problem with our pdf docs. The trick is simply to define a substitution this way:

```
.. |LaTeXHyphenate| raw:: latex

                    \hspace{0pt}
```

and use it manually where needed, like this

```
|LaTeXHyphenate|\ longwordatstartofaparagraphinnarrowwidthcontext
```

I have no strong opinion. And I hesitated to ask for merg in 3.x or master branch.


sphinx/writers/latex.py
```diff
@@ -1161,7 +1161,9 @@ def visit_paragraph(self, node: Element) -> None:
             # (first one is label node)
             pass
         else:
-            self.body.append('\n')
+            # the \hskip0pt\relax is to allow hyphenation of first word of
+            # a paragraph in narrow contexts such as in a table cell
+            self.body.append('\\hskip0pt\\relax\n')
```
jfbu: It is crucial that this is never inserted before "vertical" material, in particular before a list. Else it will create extra vertical space. I hope that visit_paragraph() will always correspond to starting an actual paragraph in TeX sense of the word. But with substitutions I am not sure.

Adding this rather for example at start of each list item like I first considered (due to relevance to `hlist` of #8779) can not work because the item might itself be a list and adding this will, as I mentioned, then create extra whitespace due to entering horizontal mode with no reason, only to leave it immediately for a "vertical" structure.

Took me some time to change all test files, but if there is a doubt here, better to not merge this and find another way.

tests/roots/test-latex-table/expects/complex_spanning_cell.tex
```diff
@@ -1,10 +1,10 @@
-\label{\detokenize{complex:complex-spanning-cell}}
+\label{\detokenize{complex:complex-spanning-cell}}\hskip0pt\relax
 table having …
 \begin{itemize}
-\item {} 
+\item {} \hskip0pt\relax
```
jfbu: I thought about modifiying `visit_list_item` but as the item might be a nested sublist, the insertion must come from actual paragraph. Notice here that the blanks in the sequence ` {} \hskip0pt\relax` introduces no extra space at all in pdf output.

sphinx/writers/latex.py
jfbu: Sorry, this has a bug of suppressing in output a blank line which is needed by TeX to recognize paragraphs. I am correcting this stupidity in next commit, which will abstract the modification into `\sphinxAtStartPar`.

tests/roots/test-latex-table/expects/complex_spanning_cell.tex
jfbu: Now `\hskip0pt\relax` is modified into `\sphinxAtStartPar`, and here it would be on next line so the `\item {}<space>` line remains exactly as in current Sphinx. So this will not modify things for those who parse output file for this specific line for whatever purpose.

jfbu:
The `\sphinxAtStartPar` line is added as first line of almost all text paragraphs. Not in footnotes, but there is already there `\sphinxAtStartFootnote` which indirectly also allows TeX to hyphenate first word.

To review this, the thing to check is whether this insertion of `\sphinxAtStartPar` will really only happen before things that start a paragraph in TeX sense. Indeed it triggers TeX entering into "LR mode" (LaTeX vocabulary), or "horizontal mode" (TeXbook) and if immediately after that there is for example a list environment, this will cause TeX to finish the horizontal mode, hence output in pdf a blank line, resulting in extra vertical space. Also, if a blank line in tex file follows the `\sphinxAtStartPar` this will cause a blank line in output.

jfbu:
Here is updated test file 
[index.rst.txt](https://github.com/sphinx-doc/sphinx/files/5893148/index.rst.txt)

With current Sphinx it produces this

Capture d’écran 2021-01-29 à 11 14 05

With this PR it will give this

Capture d’écran 2021-01-29 à 11 16 06

and
Capture d’écran 2021-01-29 à 11 16 15

Here, the `hor-` is still too long to fit, so it gets shifted down vertically from being first list item, but this is extreme situation and we can't do much about it. If the column had been a bit wider, it would have fitted:

Capture d’écran 2021-01-29 à 11 20 41

note: some captions in the screenshots say that lists in table cells trigger usage of `tabular` or `longtable` and not `tabulary`. However if one uses `tabularcolumns` directive one can again force usage of tabulary even with a list in a cell.

jfbu:
Relates to #3042 but does not fix it.

sphinx/writers/latex.py
```diff
@@ -1161,7 +1161,9 @@ def visit_paragraph(self, node: Element) -> None:
             # (first one is label node)
             pass
         else:
-            self.body.append('\n')
+            # the \sphinxAtStartPar is to allow hyphenation of first word of
+            # a paragraph in narrow contexts such as in a table cell
+            self.body.append('\n\\sphinxAtStartPar\n')
```
tk0miya, line 1166: The children of paragraph node are regularly text nodes or inline nodes. It is defined at the doctree specification.
https://docutils.sourceforge.io/docs/ref/doctree.html

Almost of the inline nodes are decorated text. But, some of them are not; for example, math, image, problematic and raw. Additionally, some extensions might generate broken doctree under the paragraph node.

doc/latex.rst
```diff
@@ -1024,6 +1024,14 @@ Environments
 Miscellany
 ~~~~~~~~~~
 
+- Every text paragraph in document body starts with `\sphinxAtStartPar`.
+  Currently, this is used to insert a zero width horizontal skip which
+  is a trick to allow TeX hyphenation of the first word of a paragraph
+  in a narrow context (like a table cell). For ``'lualatex'`` which
+  does not need the trick, the `\sphinxAtStartPar` does nothing.
+
+  .. versionadded:: 3.5.0
```
tk0miya, line 1033: I don't know this change will break PDF generation. But it would be better to add big change into the next major release. It's planned for this April.

tk0miya (approved):
Very cool!! +1 for merging. If your worry came from the vertical material of LaTeX, you don't need to care it because the children of paragraph node is all inline nodes (basically).

doc/latex.rst
jfbu: To the best of my knowledge the change can not break PDF generation. As far as I can tell, at worst it could add an extra blank line if the first child is implemented via a "vertical" environment: typically a list, but this seems excluded: in my understanding a list can not be first child of a paragraph node. An image is not a problem if it is not embedded into "figure" (for example I image inserted from a substitution). I tested that math like `equation` or `align` do not seem to be affected. At any rate there is no way I can see this can break PDF build whatever nodes follow. The only thing it can cause is this extra blank line. But I can not construct an example from rst source creating this.

sphinx/writers/latex.py
jfbu: A math node is fine (via the `math` role), an image (via a substitution) also is no problem. As per ``raw latex``, also inserted by a substitution then if the raw latex starts some list environment it will be shifted one line down, but it seems then a user error to do this, the raw directive should have been used directly to insert the LaTeX "display" material in place of "inline" material. I can not think of a natural example causing a problem, after all docutils "paragraph" concept pretty well matches with latex "paragraph" concept I hope.

jfbu:
Thank you @tk0miya always for reviewing and advice! What this code does is simply to add in LaTeX output at the start of (most) paragraphs a macro `\sphinxAtStartPar` which does almost nothing (it inserts horizontal glue of zero width). This is a trick to allow hyphenation of first word, especially in tables (tables rendered by tabulary do exactly this same trick already from the latex code of package tabulary, but if user has employed `tabularcolumns`  with some `p{}` specifier, this would be lost). With lualatex the added macro does nothing because lualatex engine does not have this limitation of original (La)TeX.

I think this can not be breaking change, thus now merging into 3.x!

sphinx/writers/latex.py
tk0miya: Agreed. From my perspective, docutils' paragraph model and LaTeX's horizontal mode are similar. (But I did not mention that yesterday because I'm not familiar with LaTeX's mode :-p)

doc/latex.rst
tk0miya: >To the best of my knowledge the change can not break PDF generation. 

Okay. I understand and agreed this does not break PDF generation. Let us see what will be reported.

>(for example I image inserted from a substitution)

Fortunately, reST's substitution only allows inline elements. Nobody can't insert non "inline" elements under the paragraph node without hacks (by raw node or extensions).

>A substitution definition block contains an embedded inline-compatible directive (without the leading ".. "), such as "image" or "replace".
>https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#substitution-definitions

tk0miya:
🎉

PR #8786, opened 2021-01-30
Merge 3.x to master

tk0miya:
### Feature or Bugfix
- Maintenance


sphinx/texinputs/sphinxlatexoptionhandling.sty
```diff
@@ -128,7 +128,12 @@
 \DisableKeyvalOption{sphinx}{numfigreset}
 \DisableKeyvalOption{sphinx}{nonumfigreset}
 \DisableKeyvalOption{sphinx}{mathnumfig}
+% To allow hyphenation of first word in narrow contexts; no option,
+% customization to be done via 'preamble' key
+\newcommand*\sphinxAtStartPar{\nobreak\hskip\z@skip}
+% No need for the \hspace{0pt} trick (\hskip\z@skip) with luatex
+\ifdefined\directlua\let\sphinxAtStartPar\@empty\fi
```
tk0miya, line 135: @jfbu This hunk is copied manually. Could you check the recent updates are surely merged into master (because we need to separate the changes on 3.x on merging into the master)?

sphinx/texinputs/sphinxlatexrequirepackages.sty
```diff
@@ -208,6 +208,8 @@
 \ltx@ifundefined{@removefromreset}
     {\RequirePackage{remreset}}
     {}% avoid warning
+% To support hlist directive
+\RequirePackage{multicol}
```
tk0miya, line 212: This is also.

sphinx/texinputs/sphinxlatexoptionhandling.sty
jfbu: This is LGTM it corresponds to where I located it in 3.x sphinx.sty. We provide an option for `\sphinxAtStartFootnote` and this is why the latter is here. But I thinks such an option to pass LaTeX code via the intermediary of `\sphinxsetup` is not very useful, and unused by people so I did not add a similar one for `\sphinxAtStartPar`. So I located it near options. Before 4.0 release we can perhaps move this hunk to `sphinxlatexstyletext.sty`.

BTW, in case anything goes wrong at 3.5 release and for whatever reason I can't be joined, telling user to simply add ``\def\sphinxAtStartPar{}`` in preamble should be enough as urgent fix.

sphinx/texinputs/sphinxlatexrequirepackages.sty
jfbu: LGTM

tests/roots/test-latex-table/expects/complex_spanning_cell.tex
```diff
@@ -16,39 +19,51 @@
 \hline
 \sphinxmultirow{3}{1}{%
 \begin{varwidth}[t]{\sphinxcolwidth{1}{5}}
+
+\sphinxAtStartPar
```
jfbu, line 23: I wonder why my patch to ``visit_paragraph`` resulted in this context in extra blank line (we can see many instance in this file). Perhaps has to do with the ``needs_linetrimming`` conditional and the injection of ``\n`` at end of ``varwidth``. This is at any rate not a worry as it changes nothing to PDF output. I will make double-check nevertheless.

jfbu (commented):
LGTM!

jfbu:
Le 30 janv. 2021 à 09:09, Jean-François B. <notifications@github.com> a écrit :

>  I did not add a similar one for \sphinxAtStartPar. So I located it near options. 

I meant "Despite this, I located it near options". Sorry for bad wording. I have some unexplained internet connection speed problem and I did not read me twice as the interface was frozen 90% of the time

tk0miya:
Thank you for checking! Merged.

PR #8787, opened 2021-01-30
Let latex writer line trimming from depart_entry() work as before #8781

jfbu:
This is cosmetic as the blank line starting varwidth environment used
for merged table cells in latex output changed nothing to PDF.

Nevertheless I extended a unit test to have a multi-paragraph merged
cell using varwidth. What is important is that \sphinxAtStartPar line
itself is never followed by blank line.

### Relates
- #8781 
- #8786 (https://github.com/sphinx-doc/sphinx/pull/8786#discussion_r567214785)

I had updated tests at #8781 in a semi-automatic matter. The line trimming done in certain circumstances of merged table cells was modified by #8781 but the extra blank line in produced tex file is not important. Just to be sure this fixes it.


sphinx/writers/latex.py
```diff
@@ -1163,7 +1163,8 @@ def visit_paragraph(self, node: Element) -> None:
         else:
             # the \sphinxAtStartPar is to allow hyphenation of first word of
             # a paragraph in narrow contexts such as in a table cell
-            self.body.append('\n\\sphinxAtStartPar\n')
+            # added as two items (cf. line trimming in depart_entry())
+            self.body.extend(['\n', '\\sphinxAtStartPar\n'])
```
jfbu, line 1167: are we sure paragraph first child node will never insert an extra ``\n`` after this? In case of doubt:
- we could replace the `\n` after `\\sphinxAtStartPar` by  a space. Then however `\sphinxAtStartPar<space>` will start most paragraphs in latex output rather than standing on a line of its own. This makes absolutely no difference to LaTeX but is less pleasant for simple difference of output with Sphinx before 3.5. 
- or, the first child node should be trimmed for leading `'\n'` (and if nothing is left of this node, then the next one...)

jfbu (commented):
As I have not studied Docutils closely enough I am not sure if first child of paragraph node can bring its own `\n`. Testing showed no example from our test suite which creates such a situation. If something like this happens there will be blank line after `\sphinxAtStartPar`  in latex file and this will then cause an extra blank line in pdf output.

JelisaCooper (approved):

sphinx/writers/latex.py
tk0miya: As I commented in another PR, all kinds of children for paragraph nodes are inline elements (basically). So no child will insert a blank line before itself. At a glance, it seems visitors in LaTeX writer that inserts `\n` before their macros are block level (vertical mode) elements.

tk0miya (approved):

jfbu:
Thanks @tk0miya for sharing knowledge. All sounds good then, and I will merge!

PR #8769, opened 2021-01-27
Refactor latex into multiple files

jfbu:
### Feature or Bugfix
- Refactoring

### Purpose
Make replacement of big chunks of Sphinx LaTeX code in one go possible, especially for customizing macros, via addition of a file from `latex_additional_ files` rather than bloated conf.py.

There does not exist in the LaTeX world anything even remotely like CSS styling in HTML.

But splitting big file into smaller units makes replacement of smaller unit possible. Actually things are really intertwined with LaTeX builder (influencing preamble contents) and LaTeX writer which outputs the mark-up, and the actual LaTeX code fragments. But for some things, for example customization of headings (now in `sphinxlatexstyleheadings.sty`) it is easier to replace the whole file with the needed changes than to add many things to ``latex_elements``.

For time being I simply split the sphinx.sty along predefined sections which had been structured on some previous refactoring. Arguably one should now move the `\RequirePackage{titlesec}` and associated bugfix from `sphinxlatexrequirepacakages.sty` to `sphinxlatexstyleheadings.sty`, only to give an example, I am not sure there are many which are as clearcut.

Also take this opportunity to adopt some uniform naming convention for LaTeX support files.

(merging 3.x will need I guess some manual intervention in `\ProvidesPackage` of `sphinxpackagefootnote` which is new name of `footnotehyper-sphinx`)

### Relates
- #8045

PR #8790, opened 2021-01-30
Refactor LaTeX style files

jfbu:
This is a (continuation and) re-work of #8769 (e6bf9145dc)

I have reintegrated option handling and most package loading into the
original file sphinx.sty and reorganized completely the filenames of
secondary style files.

sphinx.sty had become too big and first #8769 now this more definitive
refactoring is necessary to clarify structure, dependencies, and ease up
future maintenance.

Unfortunately this means a lot of moving around hunks of latex code with
some alterations.  I tried to carefully check everything is defined in
right order (as LaTeX being a macro expansion language, often one can
manipulate things before them being defined, nevertheless I checked
things are done in order).

Only simple thing is to review is that I added missing EOLs at last
lines of the extracted files...

----

Sorry that it is very difficult to review this due to all moving around of hunks of latex code. This is a one-time change for a better future :-)


tk0miya (approved):
+1; it's too large to review this. So let's take a look at how it works after merging.

jfbu:
Thanks, indeed it is very difficult to review. I am paying much attention that no latex code hunk got lost in refactoring.

I think I will push a small additional commit to again have some `\RequirePackage` only inside the "components" where they are strictly needed. In particular `\RequirePackage{framed}`: the only reason I had put it back in `sphinx.sty` is that it is needed in three "components" (i.e. for literal blocks, for topics boxes, for admonitions), but it is ok that each has its own `\RequirePackage{framed}`. The idea is that the Sphinx latex builder itself does not produce any mark-up directly requiring this package. It is only required indirectly for the environments used for literal blocks, topic boxes, heavy type notices.

I may recreate a `sphinxlatexpackages.sty` for the few packages that don't fit elsewhere but for a reason or another are required by the sphinx mark-up. If an extension modifies the mark-up it should be made easy for the extension to also avoid loading now unneeded packages.

Thus, I will take final decisions, push such a small additional commit and merge after testing completes.

jfbu:
I tried to document which latex writer mark-up is supported by which component of `sphinx.sty`, after the split of `sphinx.sty` into smaller files. This will hopefully help future maintenance. (the process also helped me uncover bug #8796). Now merging!

jfbu:
This caused accidental revert of #8767, fixed at a5939ea58

PR #8792, opened 2021-01-31
refactor: linkcheck: Skip queuing ignored URIs

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To make checker-thread simpler, this checkes the URI is ignored before
queueing.
- This is an alternative of #8774

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -129,6 +129,9 @@ class CheckExternalLinksBuilder(DummyBuilder):
             thread.start()
             self.workers.append(thread)
 
+    def is_ignored_uri(self, uri: str) -> bool:
+        return any(pat.match(uri) for pat in self.to_ignore)
+
     @property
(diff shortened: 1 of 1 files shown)


francoisfreitag (dismissed):
Thanks! I was seeing the post transform as the place to perform the triage, but since the builder outputs the result, best keep the triage logic in the builder.

---
Nit in commit message, checkes -> checks.

francoisfreitag:
How does this change relate to #8806? If the checker is extracted, I believe the ignore list should be part of the checker and not the builder?

tk0miya:
Thank you for reviewing.

>If the checker is extracted, I believe the ignore list should be part of the checker and not the builder?

Yes, it will be moved into the checker after merging this.

tk0miya:
Just updated > commit message.

francoisfreitag (approved):

tk0miya:
Thank you for reviewing!

PR #8795, opened 2021-01-31
refactor: linkcheck: Do not cache the checking result

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Since #8702, the linkcheck builder processes one URL only once.
Therefore no reason to cache its result now.  This removes the whole of
cache mechanism (additional removals for #8723).
- This completely deprecates `broken`, `good` and `redirected`, but allows
to access empty ones via properties not to break 3rd party extensions.


tk0miya:
I found a way to keep compatibility. So withdrawing this.

PR #8797, opened 2021-01-31
Latex sphinxsafeincludegraphics corebug

jfbu:
This bug was introduced at 2.0 (commit 669f9c3a) (sorry)

This core bug seems not to cause harm, so it is no reason to hurry 3.4.4, it can be delayed to 3.5


tk0miya (commented):
LGTM. I prefer to merge non-crtical bugfixes into 3.x branch. But it's okay because 3.5 will be released soon.

tk0miya:
Merged.

PR #8799, opened 2021-01-31
refactor: AttributeError handling for sphinx.util.inspect.get* is not needed

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Internally, `getall()`, `getannotations()`, `getmro()` and `getslots()` in sphinx.util.inspect  use `safe_getattr()` with the `default` keyword.  Therefore it never raises AttributeError even if the subject raises an error on accessing `__mro__` attribute.
- This fixes the wrong their usage.

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1020,10 +1020,6 @@ class ModuleDocumenter(Documenter):
         try:
             if not self.options.ignore_module_all:
                 self.__all__ = inspect.getall(self.object)
-        except AttributeError as exc:
-            # __all__ raises an error.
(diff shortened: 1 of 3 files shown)


tk0miya:
I confirmed with `sphinx.util.inspect.getannotations()` suppresses errors on accessing its attribute with following code:
```
>>> class Foo:
...     @property
...     def __annotations__(self):
...         raise
...
>>>
>>> foo = Foo()
>>> foo.__annotations__
  File "<stdin>", line 4, in __annotations__
RuntimeError: No active exception to reraise
>>> from sphinx.util import inspect
>>> inspect.getannotations(foo)
{}
```

PR #8806, opened 2021-01-31
refactor: linkcheck: Separate worker feature from builder class

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Now, the linkcheck builder becomes bigger and complicated.
- To reduce the complexity of the class, this tries to separate the worker feature from the builder class as a first step.
- I'll work to separate the thread-conductor feature from the builder on the next step.
- Finally, the builder will be changed to the representer feature (logging and output).

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -274,7 +274,7 @@ class CheckExternalLinksBuilder(DummyBuilder):
 
 
 class HyperlinkAvailabilityCheckWorker(Thread):
-    """A worker class for checing the availability of hyperlinks."""
+    """A worker class for checking the availability of hyperlinks."""
 
     def __init__(self, builder: CheckExternalLinksBuilder) -> None:
         self.app = builder.app
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -12,13 +12,13 @@ import json
 import queue
 import re
 import socket
-import threading
 import time
 import warnings
 from datetime import datetime, timezone
 from email.utils import parsedate_to_datetime
 from html.parser import HTMLParser
 from os import path
+from threading import Thread
 from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, cast
 from urllib.parse import unquote, urlparse
 
@@ -129,9 +129,9 @@ class CheckExternalLinksBuilder(DummyBuilder):
         self.rate_limits = {}  # type: Dict[str, RateLimit]
         self.wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue
         self.rqueue = queue.Queue()  # type: queue.Queue
-        self.workers = []  # type: List[threading.Thread]
+        self.workers = []  # type: List[Thread]
         for i in range(self.config.linkcheck_workers):
-            thread = threading.Thread(target=self.check_thread, daemon=True)
+            thread = HyperlinkAvailabilityCheckWorker(self)
             thread.start()
             self.workers.append(thread)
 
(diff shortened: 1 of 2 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -154,9 +152,126 @@ def redirected(self) -> Dict[str, Tuple[str, int]]:
...
+
+    def limit_rate(self, response: Response) -> Optional[float]:
+        warnings.warn(
+            "%s.%s is deprecated." % (self.__class__.__name__, "redirected"),
+            RemovedInSphinx50Warning,
+            stacklevel=2,
+        )
+        return HyperlinkAvailabilityCheckWorker(self).limit_rate(response)
+
+    def process_result(self, result: Tuple[str, str, int, str, str, int]) -> None:
```
tk0miya, line 184: Note: The methods below are not changed at all. I only moved (copied) `check_thread()` and `limit_rate()` to `HyperlinkAvailabilityCheckWorker` class.

sphinx/builders/linkcheck.py
```diff
@@ -154,9 +152,126 @@ def redirected(self) -> Dict[str, Tuple[str, int]]:
...
+    def __init__(self, builder: CheckExternalLinksBuilder) -> None:
+        self.app = builder.app
+        self.anchors_ignore = builder.anchors_ignore
+        self.auth = builder.auth
+        self.config = builder.config
+        self.env = builder.env
+        self.rate_limits = builder.rate_limits
+        self.rqueue = builder.rqueue
+        self.to_ignore = builder.to_ignore
+        self.wqueue = builder.wqueue
```
tk0miya, line 288: I'll refactor these attributes if agreed.

sphinx/builders/linkcheck.py
```diff
@@ -136,7 +134,7 @@ def good(self) -> Set[str]:
             RemovedInSphinx50Warning,
             stacklevel=2,
         )
-        return self._good
+        return set()
```
francoisfreitag: <s>This is not compatible with extensions, as `good` no longer collects the good links. Similarly, `broken` and `redirects` won’t be populated.

Should the change target the master branch?</s>

francoisfreitag (commented):
:+1: I like the idea of splitting the responsibilities, the builder currently does many things.

> I'll work to separate the thread-conductor feature from the builder on the next step.

Sounds like a good plan to me, especially for state shared between workers (rate limits) but not tied to the builder.

sphinx/builders/linkcheck.py
francoisfreitag: I missed commit refactor: “linkcheck: Do not cache the checking result”, which explains this change in behavior is expected.

---

> This completely deprecates `broken`, `good` and `redirected`, but allows to access empty ones via properties not to break 3rd party extensions.

I wonder if that’s really better than removing the properties? The empty properties are likely to give incorrect results to 3rd parties, which may be worse than a plain `AttributeError`. What do you think?

tk0miya: I know returing empty values will cause the 3rd party extensions not to work. But it's better than a crash. In other words, my priority is 1) 100% compatible 2) not working 3) crashed.

tk0miya: I found a way to keep compatibility. So I reverted this change.

sphinx/builders/linkcheck.py
```diff
@@ -157,6 +157,126 @@ def redirected(self) -> Dict[str, Tuple[str, int]]:
         return self._redirected
 
     def check_thread(self) -> None:
+        warnings.warn(
+            "%s.%s is deprecated." % (self.__class__.__name__, "redirected"),
```
francoisfreitag:
```suggestion
            "%s.%s is deprecated." % (self.__class__.__name__, "check_thread"),
```

sphinx/builders/linkcheck.py
```diff
@@ -157,6 +157,126 @@ def redirected(self) -> Dict[str, Tuple[str, int]]:
...
+            RemovedInSphinx50Warning,
+            stacklevel=2,
+        )
+        # do nothing.
+
+    def limit_rate(self, response: Response) -> Optional[float]:
+        warnings.warn(
+            "%s.%s is deprecated." % (self.__class__.__name__, "redirected"),
```
francoisfreitag:
```suggestion
            "%s.%s is deprecated." % (self.__class__.__name__, "limit_rate"),
```

sphinx/builders/linkcheck.py
```diff
@@ -160,6 +160,133 @@ def redirected(self) -> Dict[str, Tuple[str, int]]:
...
+        self.json_outfile.write('\n')
+
+    def finish(self) -> None:
+        logger.info('')
+
+        with open(path.join(self.outdir, 'output.txt'), 'w') as self.txt_outfile,\
+             open(path.join(self.outdir, 'output.json'), 'w') as self.json_outfile:
+            total_links = 0
+            for hyperlink in self.hyperlinks.values():
+                if self.is_ignored_uri(hyperlink.uri):
```
tk0miya, line 254: Note: I'll separate this from the builder in the next step.

tk0miya:
Now ready to merge this. @francoisfreitag Do you have any comment?

sphinx/builders/linkcheck.py
francoisfreitag: I think these attributes should be refactored? At least `app` is not used.

sphinx/builders/linkcheck.py
```diff
@@ -166,6 +166,134 @@ def redirected(self) -> Dict[str, Tuple[str, int]]:
...
+            self.app.statuscode = 1
+
+        self.wqueue.join()
+        # Shutdown threads.
+        for worker in self.workers:
+            self.wqueue.put((CHECK_IMMEDIATELY, None, None, None), False)
+
+
+class HyperlinkAvailabilityCheckWorker(Thread):
+    """A worker class for checing the availability of hyperlinks."""
```
francoisfreitag:
```suggestion
    """A worker class for checking the availability of hyperlinks."""
```

sphinx/builders/linkcheck.py
tk0miya: Sorry for the lack of words. I'll do it on the another PR after merging this. Because it's hard to read the diff until merging this.

francoisfreitag (approved):
Otherwise LGTM.

tk0miya:
@francoisfreitag Thank you for reviewing. I'll merge this soon after CI passed.

tk0miya:
Merged!

Issue #8807, opened 2021-01-31
LaTeX: todo directive inserts `\label` without a proper anchor point (from `\phantomsection` or a counter stepping)


jfbu:
**Describe the bug**
Here is an example from `make latex` on our own docs, one sees in produced `sphinx.tex`:

```
\sphinxAtStartPar
\raisebox{-0.5\height}{\sphinxincludegraphics{{more}.png}} See {\hyperref[\detokenize{usage/configuration::doc}]{\sphinxcrossref{\DUrole{doc}{Configuration}}}} for documentation of all available
config values.

\begin{sphinxadmonition}{note}{\label{\detokenize{usage/quickstart:id5}}Todo:}
\sphinxAtStartPar
Move this entire doc to a different section
\end{sphinxadmonition}

```

However this environment, by default, does not insert an anchor point.

Now, normally this goes completely silent. If the document were to use `usage/quickstart:id5` destination name this would link to the section title which also has label `usage/quickstart:basic-configuration`.

However I hit against this by working on implemeting the feature #6276.

As a work-around I will associate a counter to the `sphinxadmonition` environment so that it can be increased at each such occurrence and thus generate a suitable anchor point via hyperref services.

This will be a fix to this.

**To Reproduce**

Use ``todo`` directive and check latex file from `make latex`.
**Environment info**
- OS: [e.g. Unix/Linux/Mac/Win/other with version]
- Python version: [e.g. 3.7.1]
- Sphinx version: 3.4
- Sphinx extensions:  [e.g. sphinx.ext.autodoc, recommonmark]
- Extra tools: [e.g. Browser, tex or something else]

jfbu:
This was closed by #12508.

Issue #8798, opened 2021-01-31
3.5.0 release plan


tk0miya:
Hi maintainers,

About 1.5months has passed since the last stable release. Based on our release cycle, I'll ship a new package 3.5.0 within 2 weeks. The X-day will be held on 13th or 14th Feb.
Are you ready? If you have any remaining tasks, please let me know.

Note: issues marked as 3.5.0:
https://github.com/sphinx-doc/sphinx/milestone/94

jfbu:
#8838 must be fixed before shipping 3.5 :-(, else we have to revert #8781

Fortunately there was quick fix at #8840 :tada:, I will merge after testing completes...

jfbu:
LaTeX: I have removed 3.5.0 milestone from tasks for which there is no time for me until next week-end. I have no tasks left, there is only #8832 and #8851 left and they are ready for 3.5.0 milestone. And #8854 :)

edit: all clear for me. No tasks left for 3.5.0

edit: except #8877...which better fixes #8838

edit: merged! I am all done for 3.5.0...

tk0miya:
Time has come. Let's start releasing.

tk0miya:
done! 🎉 
Thank you for your effort! Our next target is v4.0 :-)

PR #8808, opened 2021-01-31
LaTeX: add support for named destinations (without prefix) in PDFs

### LINKED ISSUE #6276: Write \hypertarget for labels to allow external linking to a destination within a PDF
PDFs have the concept of "named targets" that can be used by an external application to jump directly to a part of the document. This is very useful to link to a specific part of a long document.

Let's say a document doc.pdf has a section with a named target "my-target", then PDF viewers can jump directly to that section by opening the url `http://example.com/doc.pdf#nameddest=my-target`. Some PDF viewers also support the (non-standard) short form `http://example.com/doc.pdf#my-target`.
This works in Firefox (with pdf.js), Chrome, Sumatra, Evince, Adobe Acrobat Reader and probably others.

It is possible to create a named destination in LaTeX by using the `\hypertarget` command from the `hyperref` package ([example on Stack Overflow](https://tex.stackexchange.com/questions/213860/how-to-generate-a-named-destination-in-pdf#answer-213862))

This could be supported by Sphinx if it would generate a `\hypertarget` around a `\label`.

Example input:
```rst
.. _this-is-label-two:

Heading 2
=========
```

Current output:
```latex
\chapter{Heading 2}
\label{\detokenize{index:heading-2}}\label{\detokenize{index:this-is-label-two}}
```

Proposed output:
```latex
\chapter{Heading 2}
\label{\detokenize{index:heading-2}}\hypertarget{this-is-label-two}{\label{\detokenize{index:this-is-label-two}}}
```

A PDF viewer can now jump directly to this section by opening http://example.com/doc.pdf#nameddest=this-is-label-two.

Note that I've removed the `index:` prefix in the hypertarget label. An external document can know the label to link to, but it is unlikely to know the original rst doc name (`index` in this case).
With this feature it would even be possible to have Intersphinx support linking between PDF files, rather than always pointing to HTML.

P.S. On my TexLive 2016, this works fine with pdflatex, but not with XeLaTeX. This is easily fixed by adding `\special{dvipdfmx:config C 0x0010}` to the preamble. See bug reports for [Xetex](https://sourceforge.net/p/xetex/bugs/132/) and [MikTex](https://sourceforge.net/p/miktex/bugs/2529/).

jfbu:
Allow to link externally to some permanent PDF ids, like this

    http://example.com/doc.pdf#nameddest=my-target

where my-target is chosen label in source (i.e. the document name is not
added as prefix). This allows external linking in a way stable against
renaming of sections or moving documentation from one document to the
other in the same project.

(The "nameddest=" part is even optional in some viewer)

Closes: #6276

<strike>Closes issue #8807</strike>

This adds a feature <strike>(and closes bug #8807 but using  a workaround).</strike>

As per why I don't use `\hypertarget` as proposed in original ticket #6276, it has to do with the fact that it would have to be located before the labeled title or caption.

There is a functionality in hyperref to name destinations. But its option `destlabel` makes it so that it is the first `\label` which decides of the name. In my testing, a user added label in rst source ends up rather as last `\label`. So the approach taken here is to append to last `\label` a dedicated macro. Notice that this avoids any LaTeX hack into `\label`. There is `\label@hook` which hyperref uses under option `destlabel` to insert `\HyperDestRename`. However this is no way to use because we want only the last `\label` to prevail not the first. For simplicity though, the code rather than extending the pre-existing `\HyperDestNameFilter` overwrites it completely with our own version, so if there exists any package influencing naming of destination anchors, and using `\HyperDestNameFilter` as per hyperref doc, their effect will be suppressed. (but a priori if any such thing exists it is for hand-written latex sources, so it is not of Sphinx concern). In particular, our code is stable against passing option `destlabel` to hyperref, which should not change anything. I know that the latex code is hard to understand because one has to read `hyperref` documentation and the relevant part of its source code. We are using its hook `\HyperDestNameFilter` as per its documentation, setting it up as an expandable macro which will convert the internally generated hyperref label into the user provided name.  We imitate `hyperref` way of doing that with option `destlabel` but we have to change things and can not use this option ourselves.

I did not see without more changes to latex.py how to add the mark-up extension *only*  in case of user added labels in source, so I had to allow this extended mark-up for each usages of `hypertarget_to()`. This means that the LaTeX code has to eliminate all duplicates, which will a priori be numerous if the project contains multiple source files, from automated things, many files can have a section "options" for example. Everything is eliminated silently, it is up to user now however to make sure user-added labels are unique across all documentation to avoid pdftex warnings about distinct destinations receiving the same name.

Edit: it may be that some of the above is slightly obsolete because I may have found a way (possibly not natural one) to restrict, so it seems, to only handle hypertargets from explicitly set labels in project source.

### DIFF
--- a/tests/roots/test-latex-table/expects/complex_spanning_cell.tex
+++ b/tests/roots/test-latex-table/expects/complex_spanning_cell.tex
@@ -1,4 +1,4 @@
-\label{\detokenize{complex:complex-spanning-cell}}
+\label{\detokenize{complex:complex-spanning-cell}}\sphinxnameddest{\detokenize{complex-spanning-cell}}
 \sphinxAtStartPar
 table having …
 \begin{itemize}
--- a/tests/roots/test-latex-table/expects/longtable_having_widths.tex
+++ b/tests/roots/test-latex-table/expects/longtable_having_widths.tex
@@ -1,7 +1,7 @@
-\label{\detokenize{longtable:longtable-having-widths-option}}
+\label{\detokenize{longtable:longtable-having-widths-option}}\sphinxnameddest{\detokenize{longtable-having-widths-option}}
 
 \begin{savenotes}\sphinxatlongtablestart\begin{longtable}[c]{|\X{30}{100}|\X{70}{100}|}
-\hline\noalign{\phantomsection\label{\detokenize{longtable:namedlongtable}}\label{\detokenize{longtable:mylongtable}}}%
+\hline\noalign{\phantomsection\label{\detokenize{longtable:namedlongtable}}\label{\detokenize{longtable:mylongtable}}\sphinxnameddest{\detokenize{namedlongtable}}\sphinxnameddest{\detokenize{mylongtable}}}%
 \sphinxstyletheadfamily 
 \sphinxAtStartPar
 header1
--- a/tests/test_build_latex.py
+++ b/tests/test_build_latex.py
@@ -760,6 +760,7 @@ def test_reference_in_caption_and_codeblock_in_footnote(app, status, warning):
     assert '\\subsubsection*{The rubric title with a reference to {[}AuthorYear{]}}' in result
     assert ('\\chapter{The section with a reference to \\sphinxfootnotemark[5]}\n'
             '\\label{\\detokenize{index:the-section-with-a-reference-to}}'
+            '\\sphinxnameddest{\\detokenize{the-section-with-a-reference-to}}'
             '%\n\\begin{footnotetext}[5]\\sphinxAtStartFootnote\n'
             'Footnote in section\n%\n\\end{footnotetext}') in result
     assert ('\\caption{This is the figure caption with a footnote to '
@@ -1447,9 +1448,12 @@ def test_latex_labels(app, status, warning):
             r'\label{\detokenize{index:id1}}'
             r'\label{\detokenize{index:figure2}}'
             r'\label{\detokenize{index:figure1}}'
+            r'\sphinxnameddest{\detokenize{figure2}}'
+            r'\sphinxnameddest{\detokenize{figure1}}'
             r'\end{figure}' in result)
     assert (r'\caption{labeled figure}'
-            '\\label{\\detokenize{index:figure3}}\n'
+            '\\label{\\detokenize{index:figure3}}'
+            '\\sphinxnameddest{\\detokenize{figure3}}\n'
             '\\begin{sphinxlegend}\n\\sphinxAtStartPar\n'
             'with a legend\n\\end{sphinxlegend}\n'
             r'\end{figure}' in result)
@@ -1457,29 +1461,41 @@ def test_latex_labels(app, status, warning):
     # code-blocks
     assert (r'\def\sphinxLiteralBlockLabel{'
             r'\label{\detokenize{index:codeblock2}}'
-            r'\label{\detokenize{index:codeblock1}}}' in result)
+            r'\label{\detokenize{index:codeblock1}}'
+            r'\sphinxnameddest{\detokenize{codeblock2}}'
+            r'\sphinxnameddest{\detokenize{codeblock1}}' in result)
     assert (r'\def\sphinxLiteralBlockLabel{'
-            r'\label{\detokenize{index:codeblock3}}}' in result)
+            r'\label{\detokenize{index:codeblock3}}'
+            r'\sphinxnameddest{\detokenize{codeblock3}}' in result)
 
     # tables
     assert (r'\sphinxcaption{table caption}'
             r'\label{\detokenize{index:id2}}'
             r'\label{\detokenize{index:table2}}'
-            r'\label{\detokenize{index:table1}}' in result)
+            r'\label{\detokenize{index:table1}}'
+            r'\sphinxnameddest{\detokenize{table2}}'
+            r'\sphinxnameddest{\detokenize{table1}}' in result)
     assert (r'\sphinxcaption{table caption}'
-            r'\label{\detokenize{index:table3}}' in result)
+            r'\label{\detokenize{index:table3}}'
+            r'\sphinxnameddest{\detokenize{table3}}' in result)
 
     # sections
     assert ('\\chapter{subsection}\n'
             r'\label{\detokenize{index:subsection}}'
             r'\label{\detokenize{index:section2}}'
-            r'\label{\detokenize{index:section1}}' in result)
+            r'\label{\detokenize{index:section1}}'
+            r'\sphinxnameddest{\detokenize{subsection}}'
+            r'\sphinxnameddest{\detokenize{section2}}'
+            r'\sphinxnameddest{\detokenize{section1}}' in result)
     assert ('\\section{subsubsection}\n'
             r'\label{\detokenize{index:subsubsection}}'
-            r'\label{\detokenize{index:section3}}' in result)
+            r'\label{\detokenize{index:section3}}'
+            r'\sphinxnameddest{\detokenize{subsubsection}}'
+            r'\sphinxnameddest{\detokenize{section3}}' in result)
     assert ('\\subsection{otherdoc}\n'
             r'\label{\detokenize{otherdoc:otherdoc}}'
-            r'\label{\detokenize{otherdoc::doc}}' in result)
+            r'\label{\detokenize{otherdoc::doc}}'
+            r'\sphinxnameddest{\detokenize{otherdoc}}' in result)
 
     # Embedded standalone hyperlink reference (refs: #5948)
     assert result.count(r'\label{\detokenize{index:section1}}') == 1
@@ -1522,6 +1538,7 @@ def test_index_on_title(app, status, warning):
     result = (app.outdir / 'python.tex').read_text()
     assert ('\\chapter{Test for index in top level title}\n'
             '\\label{\\detokenize{contents:test-for-index-in-top-level-title}}'
+            '\\sphinxnameddest{\\detokenize{test-for-index-in-top-level-title}}'
             '\\index{index@\\spxentry{index}}\n'
             in result)
 
--- a/CHANGES
+++ b/CHANGES
@@ -117,8 +117,6 @@ Bugs fixed
 * #8780: LaTeX: long words in narrow columns may not be hyphenated
 * #8788: LaTeX: ``\titleformat`` last argument in sphinx.sty should be
   bracketed, not braced (and is anyhow not needed)
-* #8807: LaTeX: todo directive inserts ``\label`` without a proper anchor point
-  (from ``\phantomsection`` or a counter stepping)
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -117,6 +117,8 @@ Bugs fixed
 * #8780: LaTeX: long words in narrow columns may not be hyphenated
 * #8788: LaTeX: ``\titleformat`` last argument in sphinx.sty should be
   bracketed, not braced (and is anyhow not needed)
+* #8807: LaTeX: todo directive inserts ``\label`` without a proper anchor point
+  (from ``\phantomsection`` or a counter stepping)
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -62,6 +62,7 @@ Features added
 * #2030: :rst:dir:`code-block` and :rst:dir:`literalinclude` supports automatic
   dedent via no-argument ``:dedent:`` option
 * C++, also hyperlink operator overloads in expressions and alias declarations.
+* #6276: LaTeX: Support external hyperlinking to a named destination within a PDF
 * #8247: Allow production lists to refer to tokens from other production groups
 
 Bugs fixed
@@ -115,7 +116,7 @@ Bugs fixed
   (or pdflatex and :confval:`latex_use_xindy` set to True) with memoir class
 * #8780: LaTeX: long words in narrow columns may not be hyphenated
 * #8788: LaTeX: ``\titleformat`` last argument in sphinx.sty should be
-  bracketed, not braced (and is anyhow not needed) 
+  bracketed, not braced (and is anyhow not needed)
 
 Testing
 --------
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -1686,8 +1686,11 @@
                      \csname\@backslashchar color@#2\endcsname }
 
 % the main dispatch for all types of notices
+% work around a problem with todo extension triggering usage of \label
+% without a proper anchor set-up via ref-stepping a counter (or \phantomsection)
 \newcounter{sphinxadmonition}
 \newenvironment{sphinxadmonition}[2]{% #1=type, #2=heading
+  \refstepcounter{sphinxadmonition}% in case extension triggers \label usage
   % can't use #1 directly in definition of end part
   \def\spx@noticetype {#1}%
   % set parameters of heavybox/lightbox
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -1686,11 +1686,8 @@
                      \csname\@backslashchar color@#2\endcsname }
 
 % the main dispatch for all types of notices
-% work around a problem with todo extension triggering usage of \label
-% without a proper anchor set-up via ref-stepping a counter (or \phantomsection)
 \newcounter{sphinxadmonition}
 \newenvironment{sphinxadmonition}[2]{% #1=type, #2=heading
-  \refstepcounter{sphinxadmonition}% in case extension triggers \label usage
   % can't use #1 directly in definition of end part
   \def\spx@noticetype {#1}%
   % set parameters of heavybox/lightbox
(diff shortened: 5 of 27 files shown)


jfbu:
<strike>Ah sorry I will need some help here on the latex python code side. I had not tested enough and</strike> 

Edited after replacing original commit. Now this source

```rst
=================================
=================================

.. _one:


chapter
-------


.. _two:

section
~~~~~~~


.. _three:


section
~~~~~~~


.. _four:


chapter
-------


.. _five:

section
~~~~~~~

.. _six:

section
~~~~~~~
```

generates as expected this latex

```latex
\\chapter{chapter}
\label{\detokenize{index:chapter}}\label{\detokenize{index:one}}\sphinxnameddest{\detokenize{one}}

\section{section}
\label{\detokenize{index:section}}\label{\detokenize{index:two}}\sphinxnameddest{\detokenize{two}}

\section{section}
\label{\detokenize{index:three}}\label{\detokenize{index:id1}}\sphinxnameddest{\detokenize{three}}

\chapter{chapter}
\label{\detokenize{index:four}}\label{\detokenize{index:id2}}\sphinxnameddest{\detokenize{four}}

\section{section}
\label{\detokenize{index:five}}\label{\detokenize{index:id3}}\sphinxnameddest{\detokenize{five}}

\section{section}
\label{\detokenize{index:six}}\label{\detokenize{index:id4}}\sphinxnameddest{\detokenize{six}}

```

<strike>
and unfortunately we see that `\sphinxnameddest` picks us wrong id: only  ``one`` and ``two`` are fine, but then it picks up `id1`etc.... All I need to know to solve this is how to attach the `\sphinxnameddest` only to the name provided by the explicit labelling in source, not the implicit one.

t seems the duplicate title names are the cause to generate the `id1` etc... and these end up at the end of list of labels.

So this is WIP, help wanted...
</strike>

Edit: I forced-pushed a commit to fix my initial version.

However, I feel this may not be very robust what I do in `sphinx.writers.latex` regarding `hypertarget_to`. There must be some more genuine manner but I don't know enough about explicit versus implicit names in docutils + sphinx.

Anyway, it seems to work. And if for example I use

```rst
.. _five:

section
~~~~~~~

.. _five:

section
~~~~~~~
```

A warning about a Duplicate target name (originating from Docutils) is emitted as expected, and then no latex mark-up is generated, which is fine also, that means the PDF build does not raise any alarm. The "five" target will simply not get defined.

tests/test_build_latex.py
```diff
@@ -1447,36 +1447,44 @@ def test_latex_labels(app, status, warning):
             r'\label{\detokenize{index:id1}}'
             r'\label{\detokenize{index:figure2}}'
             r'\label{\detokenize{index:figure1}}'
+            r'\sphinxnameddest{\detokenize{figure1}}'
```
jfbu, line 1452: rst source has two explicit target names 'figure1' and 'figure2'. The code in `hypertarget_to()` will insert only 1 in LaTeX output. Here it is `figure1` which is first in source but I don't know if always the case. I am a currently a bit confused about hyperref explanations about target names, and right now I see why that the latex code is compatible only to define  one name for a given destination and I inherited that from hyperref similar code for "destlabel" option. Although I understand two same names can not point to two distinct targets, I am a bit confused about why a common target could not have two distinct names.

jfbu (commented):
It seems I am bit uneasy both on PDF and on Python sides. There is only LaTeX side I can claim to understand!

jfbu:
Le 1 févr. 2021 à 02:13, Jean-François B. <notifications@github.com> a écrit :

> 
> @jfbu commented on this pull request.
> 
> It seems I am bit uneasy both on PDF and on Python sides. There is only LaTeX side I can claim to understand!
> 
> 

I know on LaTeX side how to proceed to allow multiple names for same destination in PDF.  Hence, I can emit multiple `\sphinxnameddest`  in `hypertarget_to()`. There is no hook in hyperref to achieve that, but one can do it by overwriting slightly one internal. Works very well in my tests.

So please wait a bit before reviewing this, as I can't push now also due to some connection problems to internet.

Besides names do not make it through my latex.py modification if they received some escaping such as transforming an underscore into a dash so also this will need update.

tk0miya:
The ID generation rule is only available in a single document. It means the same rule might be assigned when the targets are in the different documents (ex. `index:id1` and `otherdoc:id1`). Therefore, we have to use docname for the prefix of the labels.

jfbu:
> The ID generation rule is only available in a single document. It means the same rule might be assigned when the targets are in the different documents (ex. index:id1 and otherdoc:id1). Therefore, we have to use docname for the prefix of the labels.

Yes, I see this. First apologies for multiple forced push. I completely changed the latex side of things.

- my intention is not modify at labels
- we add mark-up to latex which in particular for explcit targets will let them known without document prefix
- this will happen also for section titles; of course duplicates will occur, but this is silently ignored in the latex code, deliberately (I have to remove a commented out part about emitting a latex warning)
- for user added target labels, if some are identical (without docname prefix) again silently only the first one will set the destination


(it is responsability of user to set project wide unique ids, not needing prefix; their usage with prefix is NOT suppressed, but a usage without prefix is added)

The whole point of this PR is to provide destinations named in a way which is independent of the document numbering or even of the section titles. Collisions are not a problem when document name prefix is removed, they are ignored. And user set explicit targets will provide permanent URIs to PDF locations, i.e. when a section is renamed or moved, the permanent URI set by the rst label will still be valid, this is the goal.

There is an error in the current version (archived at https://github.com/jfbu/sphinx/pull/new/latex_nameddest_0201_14h30). 

- I am not sure but for those id names with no escaping (like replacing underscore by dash) all names are also in node['ids'] so I should loop simply over node['names'] directly for `\sphinxnameddest` insertion. I had earlier `hypertarget_nameddest()`,  `hypertarget()`, and `hypertarget_to()` and I should again do like this,
- a decision is taken if the destination name should be escaped like an id is (easier for me is only escaping is via `\detokenize` on LaTeX side)
-  the test `node_id in node['names']` is bad because ids are escaped (for underscore in particular). It seems I should loop  directly node['names']

Question: are the escaped sub-strings of node['names'] (where node is the node argument of `hypertarget_to()`) always a sub-set of the node['ids']? appears to be so. The ids are used for labels, I want to generate extra no-prefix named destinations only for "names".

The latex difficulty is solved, and that part seems to work fine, although it is a hack: how to have multiple names for same destination? In fact I am thinking of making my solution starting point of a LaTeX package, because I see here a real limitation of hyperref. Of course one can manually insert `\hypertarget` but I find this cumbersome and it is then complicated to control that no page break happens between the hypertarget and the section title. This is why hyperref hacks into the latex `\refstepcounter` which is executed in the middle of the section commands. Unfortunately the code of hyperref is only for setting one anchor point at that location.

I will mark this as WIP until it is done. Also I currently have serious internet access problems (for example I will copy this message before trying to publish it in case computer gets frozen in middle of a javascript).

 Let's consider WIP. The version current () has a problem which I understand. In node['names'

tk0miya:
>Question: are the escaped sub-strings of node['names'] (where node is the node argument of hypertarget_to()) always a sub-set of the node['ids']? 

Basically, yes. But it can be broken easily because docutils generates count-up IDs like `idX` when the given name is already registered.

For example, this is a hacky example for disproof:
```
.. envvar:: hello

envvar-hello
============
```
This is converted to the following pseudoxml:
```
<document source="/Users/tkomiya/work/tmp/doc/index.rst">
    <index entries="['single',\ 'environment\ variable;\ hello',\ 'envvar-hello',\ '',\ None]">
    <desc classes="std" desctype="envvar" domain="std" noindex="False" objtype="envvar">
        <desc_signature ids="envvar-hello">
            <desc_name xml:space="preserve">
                hello
        <desc_content>
    <section ids="id1" names="envvar-hello">
        <title>
            envvar-hello
```

Note:

* `envvar` directive registers a node_id `envvar-{argname}` (but it does not have node "name").
* The reST parser goes to register the "envvar-hello" section 
  * It registers "envvar-hello" name.
  * But "envar-hello" node_id has been already registered.
  * So the new node_id `id1` is generated by ID generator as a node_id for "envvar-hello" section.

jfbu:
> > Question: are the escaped sub-strings of node['names'] (where node is the node argument of hypertarget_to()) always a sub-set of the node['ids']?
> 
> Basically, yes. But it can be broken easily because docutils generates count-up IDs like `idX` when the given name is already registered.
> 
> For example, this is a hacky example for disproof:

Thanks this is very instructive. But whether or not names correspond to ids finally is not important to me. It was a mistake when I was testing if a name is also an id (I knew not much about docutils.nodes).

This PR does not change anything to the handling of ids and the output of `\label`. Rather I add _extra_ mark-up with `\sphinxnameddest` and this uses the `'names'` because I figured this is where are found the named things, in particular the user added targets. For those only I suppress the prefixes. Thus the pdf has the old targets fully qualified (mentioned in my answer in #6276) and also new ones. Collisions are handled silently: I mean collisons from for example two sections named "Options" in two documents. Collisons from user explicit named targets are filtered out and signaled by Docutils+ Sphinx during build anyhow.

I am attaching a sample project, and the produced pdf. If you visit the pdf in firefox, you can test it works with ``#name`` syntax. (the ``nameddest=`` is optional it seems at least for my version of firefox).


I am really wondering if the `make_id()` escaping from `docutils.nodes` which I am using is a good thing.

I find it frustrating that everything is lowercased etc... Maybe I should not do this escaping and keep the (according to docutils code comments) "user" form of the target name. But I don't know the specs for PDF named destinations.

PS1: I am seriously thinking about extracting the latex code into some official package, because I feel it solves a problem with hyperref functionality. The problem solved is to add named destinations, possibly multiple ones to same spot. There might be other way at level of hyperref pdftex driver but I did not want to output explicit code to PDF file, so I simply hacked my way into executing multiple times the hyperref code, but with possibly many names for same spot in document.

PS2: I don't think trying to use `\hypertarget` will work: if after the section title, it will be a target underneath, and if before, I see no way to prevent a page break. 

[6276_hypertarget.zip](https://github.com/sphinx-doc/sphinx/files/5906031/6276_hypertarget.zip)

[foo.pdf](https://github.com/sphinx-doc/sphinx/files/5906032/foo.pdf)

```rst
.. attention::

   my internet problems obviously were caused from my ISP not happy about me using 1.1.1.1 as DNS! (or it has problems)
```

tests/test_build_latex.py
jfbu: This comment is obsoleted from update code

tests/roots/test-latex-table/expects/complex_spanning_cell.tex
```diff
@@ -1,4 +1,4 @@
-\label{\detokenize{complex:complex-spanning-cell}}
+\label{\detokenize{complex:complex-spanning-cell}}\sphinxnameddest{\detokenize{complex-spanning-cell}}
```
jfbu, line 1: For titles, this PR thus adds some mark-up but does not modify the pre-existing labels.

Naturally as the added mark-up drops the prefix, this is likely to create duplicates. But the LaTeX code handles this silently. This is not a cause for trouble. However, if elsewhere in some other document of the project, the user has set

```
.. _intro:

Introduction
----------------
```

and in some other file

```
INTRO
---------
```

two identical destination names will exist, and to which location in PDF in the end `foo.pdf#nameddest=intro` will link to will depend on which one is first.

tests/roots/test-latex-table/expects/longtable_having_widths.tex
```diff
@@ -1,7 +1,7 @@
-\label{\detokenize{longtable:longtable-having-widths-option}}
+\label{\detokenize{longtable:longtable-having-widths-option}}\sphinxnameddest{\detokenize{longtable-having-widths-option}}
 
 \begin{savenotes}\sphinxatlongtablestart\begin{longtable}[c]{|\X{30}{100}|\X{70}{100}|}
-\hline\noalign{\phantomsection\label{\detokenize{longtable:namedlongtable}}\label{\detokenize{longtable:mylongtable}}}%
+\hline\noalign{\phantomsection\label{\detokenize{longtable:namedlongtable}}\label{\detokenize{longtable:mylongtable}}\sphinxnameddest{\detokenize{namedlongtable}}\sphinxnameddest{\detokenize{mylongtable}}}%
```
jfbu, line 4: Here we see non-prefixed mark-up associated with explicitly set target in source.

Already, Docutils/Sphinx emits warning if explicitly set targets in two distinct sources have same non-prefixed names.

jfbu (commented):
It would be much better if I had found a way to generate the extra mark-up in latex output _only_ for explicitly set targets. Indeed:

- already DocUtils/Sphinx warns if duplicate (non-doc-prefixed) explicit target names are found across whole project.
- there is however no warning if an explicitly set target name is same as a *non-prefixed* implicit target (from a section title for example); in such case the generated LaTeX code will be silent and the destination will be either the spot from explicit target or the implicit one, depending on order
- the generated LaTeX code could emit warning in case as item above; but then it would also generate warnings when non-prefixed implicit target names coincide, which is annoying


(I can not "request changes" it seems on my own project, which is a pity, I feel that github code of conduct does not appropriately recognize my belonging to the community of multliple personalities people, obviously there is much to improve here in our way of handling active minorities and agenda)

jfbu:
I will again move this to Draft status. Indeed I think we can not merge anyway because although it does achieve the goal, it has as main problem that if an explicitly set target name coincides with  a stripped-of-prefix implicit target name, or a stripped-of-prefix name from usage of `:name:` option (for example for a table), then the actual destination to which the `#nameddest=targetname` will link to in PDF depends on order in sources.

The LaTeX side of this PR is done. The Python side needs to be modified to insert `\sphinxnameddest{...}` mark-up **only** for explicitly set targets.

tk0miya:
This weekend, we'll ship a new package; v3.5.0. So how about merging this into the master branch for the 4.0 release?

tk0miya (approved):
Thank you for your explanation and refactoring. It seems good to me at a glance. I'll take a look to review this again later.

jfbu:
Le 5 févr. 2021 à 16:02, Takeshi KOMIYA <notifications@github.com> a écrit :

> 
> This weekend, we'll ship a new package; v3.5.0. So how about merging this into the master branch for the 4.0 release?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 

Yes, agreed. I will rebase on master, and this is still WIP.

other thing: Today, I was going to make PR for a fix to footnotes in captions not being hyperlinked and was hesitating between 3.5 and master branches, what is your optinion?

But I got a delay as I am currently stuck with investigating a breakage of Sphinx under lualatex with latest update to TeXLive 2020. The cmap package now breaks under lualatex whenever a tabular is entered (which triggers OT1-encoded TeX font loading it seems). I wonder if we need cmap package at all under lualatex, actually.

tk0miya:
>This weekend, we'll ship a new package; v3.5.0. So how about merging this into the master branch for the 4.0 release?

Oops. It's my bad. The new package will be released in the "next" weekend (see https://github.com/sphinx-doc/sphinx/issues/8798).
Sorry for confusing you.

tk0miya:
>other thing: Today, I was going to make PR for a fix to footnotes in captions not being hyperlinked and was hesitating between 3.5 and master branches, what is your opinion?

IMO, it's better to provide a bug fix earlier. No reason to hesitate to merge it to 3.x.

My comment about moving the branch is my bad. I misunderstand the X-Day of 3.5.0. I was just afraid to merge a new feature immediately before the release date. But we have another week until the release date. So it's okay to add a new features into 3.x branch, including this. Sorry for confusing you.

jfbu:
Le 5 févr. 2021 à 16:13, Takeshi KOMIYA <notifications@github.com> a écrit :

> 
> other thing: Today, I was going to make PR for a fix to footnotes in captions not being hyperlinked and was hesitating between 3.5 and master branches, what is your opinion?
> 
> IMO, it's better to provide a bug fix earlier. No reason to hesitate to merge it to 3.x.
> 
> My comment about moving the branch is my bad. I misunderstand the X-Day of 3.5.0. I was just afraid to merge a new feature immediately before the release date. But we have another week until the release date. So it's okay to add a new features into 3.x branch, including this. Sorry for confusing you.
> 
> 

no worries, this is minor confusion ;-) in fact I was not confused at all, because my head is currently worried by this luatex85+cmap breakage, after I updated my TeXLive today... so testing fails at my locale with lualatex. But I will now go back to finish and push my branch for solving hyperlinking footnotes in caption titles. Thus I will make PR on 3.x

About this one here, I have mixed feeling, if only I could get my hand on DocUtils ``nametypes`` attribute of the document objects I would know if the target is explicit or implicit and add the ``\sphinxnameddest`` mark-up only if explicit.

jfbu:
Sorry for long discussion in this thread, with many iterations and forced push (I try to be less verbose, but since I started again be active on Sphinx these last two weeks I must admit that I have not improved a lot during absence).

I am removing this from 3.5.0 milestone and keeping draft status because I am not satisfied:
- (main) if a user-set label collides with  a non-prefixed title or name, there is ambiguity of destination
- (less important) the available destination names should be confirmed to user at end of latexpdf run (as they go via `make_id()` they are transformed compared to user input in source); this can be added by latex code.

PR #8812, opened 2021-02-01
Latex url as footnotes in signatures

### LINKED ISSUE #8750: LaTeX: URLs as footnotes fail to show in PDF if originating from inside function type signatures
**Describe the bug**

LaTeX has configuration option `latex_show_urls = 'footnote'` which, additionally to having regular external hyperlinks in PDF also displays the URLs as text strings in footnotes. (This is with the idea of printed docs).

In the case of auto-generated documentation links, e.g. to Python documentation, which comes from function signatures, the footnote marker shows, but there is no footnote.

**To Reproduce**

Use attached project 
[missingfootnotes.zip](https://github.com/sphinx-doc/sphinx/files/5861867/missingfootnotes.zip)

or directly this:

conf.py:

```
extensions = ['sphinx.ext.autodoc',
              'sphinx.ext.intersphinx',
]

intersphinx_mapping = {
}

latex_show_urls = 'footnote'
```

index.rst

```rst

TESTING FOR FAILED PDF RENDERING OF LINKS AS FOOTNOTES IN SIGNATURES
--------------------------------------------------------------------

.. module:: sphinx.ext.autodoc
   :synopsis: Include documentation from docstrings.

.. autofunction:: between
```

Execute `make latexpdf`. In produced PDF the Python ``str`` and ``bool`` are correct external hyperlinks to Python docs. Footnote markers appear next to them, but there are no footnotes at bottom of page.

Also the console output of latex build contains lines such as:

```
pdfTeX warning (dest): name{Hfootnote.3} has been referenced but does not exist
, replaced by a fixed one
```

Note: this applies to Sphinx own docs, with between 100 and 200 footnotes at least going missing.

**Expected behavior**

Either show the footnote, or do not include any footnote marker.

**Environment info**
- OS: Mac
- Python version: 3.6.8 (I do have 3.8 and 3.9...)
- Sphinx version: tested with 3.x
- Sphinx extensions:  sphinx.ext.autodoc, sphinx.ext.intersphinx, but problem is in sphinx.sty
- Extra tools: TeXLive 2020

**Additional context**

The problem is more general, it would probably show with any external link arising in the label of a description list item, which becomes in LaTeX code some

```
\item[.... the footnote is here in label...] foo bar
```

For signatures, Sphinx uses a  ``fulllineitems`` environment in ``sphinx.sty``. This problem can be fixed by adding ``\begin{savenotes}``  at start and ``\end{savenotes}`` at end. But this will catch all footnotes from all items, not only from their labels, and display them at bottom of last page in case the description list spans multiple pages in pdf.

I tried adding the ``\begin{savenotes}`` and ``\end{savenotes}`` only at each `\item` but this causes LaTeX build failure and I expect it is hard to make it work.

As per the Sphinx own PDF docs, it runs at 3.x to about 500 pages and I wonder if people actually need paper copy. Setting ``latex_show_urls = 'no'`` will fix it for Sphinx own docs.

jfbu:
Closes: #8750 

### Feature or Bugfix

- Bugfix

Only solution I found so far is to wrap entire `fulllineitems` in `savenotes`. The footnotes will appear at bottom of page where the object description ends.

I take this opportunity to sync with upstream footnotehyper 1.1c and also a patch from 1.1 which was not incorporated so far.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -118,6 +118,8 @@ Bugs fixed
   :confval:`numfig` is not True
 * #8442: LaTeX: some indexed terms are ignored when using xelatex engine
   (or pdflatex and :confval:`latex_use_xindy` set to True) with memoir class
+* #8750: LaTeX: URLs as footnotes fail to show in PDF if originating from
+  inside function type signatures
 * #8780: LaTeX: long words in narrow columns may not be hyphenated
 * #8788: LaTeX: ``\titleformat`` last argument in sphinx.sty should be
   bracketed, not braced (and is anyhow not needed) 
--- a/sphinx/texinputs/footnotehyper-sphinx.sty
+++ b/sphinx/texinputs/footnotehyper-sphinx.sty
@@ -1,9 +1,9 @@
 \NeedsTeXFormat{LaTeX2e}
 \ProvidesPackage{footnotehyper-sphinx}%
- [2021/01/26 v1.1b hyperref aware footnote.sty for sphinx (JFB)]
+ [2021/01/29 v1.1c hyperref aware footnote.sty for sphinx (JFB)]
 %%
 %% Package: footnotehyper-sphinx
-%% Version: based on footnotehyper.sty 2021/01/26 v1.1b
+%% Version: based on footnotehyper.sty 2021/01/29 v1.1c
 %% as available at https://www.ctan.org/pkg/footnotehyper
 %% License: the one applying to Sphinx
 %%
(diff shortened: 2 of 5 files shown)


jfbu:
This fixes indeed the hundreds of 

```
pdfTeX warning (dest): name{Hfootnote.3} has been referenced but does not exist
, replaced by a fixed one
```

when building sphinx own pdf docs. Now the footnote with URLs from type hints in signatures (for the documentation of our APIs in particular) are indeed at bottom of page and the footnote marker is hyperlinked to the bottom of page correctly. But when viewing the PDF on screen (rather than printed), it is not much excitement to see all those https://docs.python.org/3/library/stdtypes.html#str links whereas the ``str`` from function signature is already itself an hyperlink...

tk0miya:
Indeed, the combination of intersphinx and `latex_show_urls = 'footnote'` is not good. Anyway, it's another topic.

tk0miya (approved):

Issue #8796, opened 2021-01-31
LaTeX: potentially critical low level TeX coding mistake has gone unnoticed so far


jfbu:
**Describe the bug**
There is a low-level bug in sphinx.sty:

https://github.com/sphinx-doc/sphinx/blob/804e46e27bf36185c8e2f7ce4c6f0e9d904a5001/sphinx/texinputs/sphinx.sty#L746-L750

The `\spx@image@box` it supposed to stand for a box register, but it is defined as a dimension! (`\newdimen`).

It does not crash the PDF build, because internally when one uses such low-level TeX primitives, the control sequence `\spx@image@box` is in fact a number. So when one uses it via `\setbox\spx@image@box...` one will overwrite the contents of the box register with number the one indicated by `\spx@image@box`. This was a number assigned by `\newdimen` it should correspond to an available dimension register. It is here used for a box register.

Clearly this could lead to very strange results, very hard to debug, and perhaps has, because a box register gets used which perhaps is also used for other things. As nobody complained PDF output contained garbage there are two explanations:

- nobody checks PDF output (possibly not far from truth :))
- as many more `\dimen` registers are allocated than `\box` registers the used `\box` had such a high number no problem came out of it.

The bug was introduced at commit 669f9c3a13a (#5992) with Sphinx 2.0

PR #8814, opened 2021-02-02
Master with merged 3.x

jfbu:
Master at a5939ea588bd8 with 3.x at 1cf915f7ddc6, merged in. I resolved latex conflicts. There was a conflict in sphinx/domains/python.py::_parse_annotation which I resolved by keeping 3.x as is. @tk0miya can you check ok?


jfbu:
oops, I broke tests! well at least I am doing this in private branch ;-) I wll try to understand what I did wrong in merge.

jfbu:
even my latex merge was no good,lost #8796 due to all the file renamings. But I have now good recipe to catch such silent losses during merges, in future

as per the failing tests on Py38 from test_domain_py.py I detected them too late as I had lost my Py38 virtualenv, else I would have hesitated to push. Anyway, now at least this shows how to merge the latex...

@tk0miya to solve latex conflicts when merging you only have to use the files from this branch in sphinx/texinputs:
- sphinxpackagefootnote.sty
- sphinxlatexgraphics.sty
- sphinx.sty
The first and third were shown as conflicted but the second one no and #8796 got lost.

jfbu:
It seems I fixed the merge of 3.x regarding domains/python.py, please check.

By the way, at my locale wih Python 3.8.7, one test of autodoc warning message fails due to warning having being translated into French.  I guess because I am too lazy to use tox and do it manually with make test?

```
======================================= short test summary info ========================================
FAILED tests/test_ext_autodoc.py::test_parse_name - AssertionError: assert 'signature arguments' in '...
============= 1 failed, 1663 passed, 6 skipped, 1 xfailed, 1 warning in 234.28s (0:03:54) ==============
```

The tested warning is in French:

```
# warning: 
WARNING: arguments de signature ou annotation de return donnés pour l’automodule test
```

and the test fails because it tries to find ``signature arguments`` in there.

jfbu:
I needed to add a ``type: ignore`` to work around``ast.Constant`` type annotation in sphinx/domains/python.py::_parse_annotation

sphinx/domains/python.py
```diff
@@ -101,17 +101,19 @@ def _parse_annotation(annotation: str, env: BuildEnvironment = None) -> List[Nod
     def unparse(node: ast.AST) -> List[Node]:
         if isinstance(node, ast.Attribute):
             return [nodes.Text("%s.%s" % (unparse(node.value)[0], node.attr))]
-        elif isinstance(node, ast.Constant):  # type: ignore
```
tk0miya, line 104: This should be kept.

tk0miya: I'll fix this after merging.

jfbu:
Le 3 févr. 2021 à 13:24, Takeshi KOMIYA <notifications@github.com> a écrit :

> 
> @tk0miya commented on this pull request.
> 
> In sphinx/domains/python.py:
> 
> > @@ -101,17 +101,19 @@ def _parse_annotation(annotation: str, env: BuildEnvironment = None) -> List[Nod
>      def unparse(node: ast.AST) -> List[Node]:
>          if isinstance(node, ast.Attribute):
>              return [nodes.Text("%s.%s" % (unparse(node.value)[0], node.attr))]
> -        elif isinstance(node, ast.Constant):  # type: ignore
> This should be kept.
> 
> 
oh I see what I did wrong, I should have kept both hunks in merge but it was indicated conflicted; in the end I added to the ``else:`` clause

tk0miya:
No worries. It's a minor mistake.

PR #8827, opened 2021-02-04
Fix broken "Extend" link in site header

thibaudcolas:
This fixes the "Extend" link in the site header pointing to a 404 (https://www.sphinx-doc.org/en/master/development.html). Spotted while navigating on https://www.sphinx-doc.org/en/master/.

### Feature or Bugfix

Bugfix

### Detail

I didn’t actually check whether the new link is correct, just based it on the documentation’s folder structure. There might also be a need to set up a redirect, if the page path changed unintentionally?

Edit: looking at the [RTD preview build](https://sphinx--8827.org.readthedocs.build/en/8827/), new link seems to be working as expected.

### Relates

I searched for "broken link" and "Extend" but couldn’t find anything related to this.


tk0miya:
Thanks!

PR #8830, opened 2021-02-05
Linkcheck: Avoid duplicate download of page to check anchors

francoisfreitag:
### Feature or Bugfix
- Feature

### Purpose
Avoid unnecessary round-trips to web servers, speeding up the linkcheck
process when multiple anchors need to be checked on the same page.

### Detail
Anchors (named fragments in the spec) indicate a secondary resource
relative to the primary resource. The page can be fetched once, and all
anchors checked on the retrieved page.

The distinction between anchors and the web page location has been made
more explicit in the work queue. A single page now has a set of anchors
to check.
Previously, the anchor was considered part of the link, such that
http://a.example/#id-1 and http://a.example/#id-2 would be considered
two different links. After this change, the link becomes
http://a.example/ and has two anchors: {'id-1', 'id-2'}.

Errors occurring when the web page is downloaded no longer show the
anchor, the anchor existence is conditioned by the existence of the page
at URI.

### Relates

- #4303

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -20,7 +20,7 @@ from html.parser import HTMLParser
 from os import path
 from threading import Thread
 from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, cast
-from urllib.parse import unquote, urlparse
+from urllib.parse import unquote, urlparse, urlunparse
 
 from docutils import nodes
 from docutils.nodes import Element
@@ -42,6 +42,7 @@ uri_re = re.compile('([a-z]+:)?//')  # matches to foo:// and // (a protocol rela
 
 Hyperlink = NamedTuple('Hyperlink', (('next_check', float),
                                      ('uri', Optional[str]),
+                                     ('anchors', Set[str]),
                                      ('docname', Optional[str]),
                                      ('lineno', Optional[int])))
 CheckResult = NamedTuple('CheckResult', (('uri', str),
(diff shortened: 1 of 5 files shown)


francoisfreitag:
Happy to rebase on #8806 and its successors. I opened this PR before the refactoring is done so that the refactoring plans for a uri preprocessing step (to filter ignored URIs and group anchors), after the link collection by the `HyperlinkCollector`.

sphinx/builders/linkcheck.py
```diff
@@ -441,21 +438,39 @@ def write_linkstat(self, data: dict) -> None:
         self.json_outfile.write(json.dumps(data))
         self.json_outfile.write('\n')
 
+    def queue_hyperlinks(self) -> int:
```
tk0miya: I think it would be better to collect anchors in HyperlinkCollector instead.

sphinx/builders/linkcheck.py
```diff
@@ -42,6 +42,7 @@
 
 Hyperlink = NamedTuple('Hyperlink', (('next_check', float),
                                      ('uri', Optional[str]),
+                                     ('anchors', Set[str]),
```
tk0miya, line 45: Unfortunately, this may not work well because the location info (pair of docname and lineno) become multiple for multiple anchors.

tk0miya (commented):
Good point. This is one of my goal why I introduced HyperlinkCollector.

sphinx/builders/linkcheck.py
francoisfreitag: Thanks for the review!

The `linkcheck_ignore` should be checked before the anchors are split (see https://github.com/sphinx-doc/sphinx/pull/8830/commits/a7f31f79ac95bc570463ec129bd4d662ec6892f8).
Performing the ignore checks in the `HyperlinkCollector` brings use closer to https://github.com/sphinx-doc/sphinx/pull/8774, which has the downside of needing to report results from the collector. Perhaps we could use a separate container for ignored URI, so that the builder handles the logging iterates on that container and logs the results?

tk0miya: >The linkcheck_ignore should be checked before the anchors are split (see a7f31f7).

Ah, okay. I understand. I did not notice that.

sphinx/builders/linkcheck.py
tk0miya: To support multiple pairs of `(docname, lineno)`, it would be better to pass a tuple `(uri_without_anchor, list of hyperlinks)` to wqueue.

Issue #2616, opened 2016-06-03
Improve footnotes on LaTeX


tk0miya:
There are some themes to improve footnotes on LaTeX
- numbering rule of footnotes; manually numbered footnotes and auto numbered
- scope of footnote numbers (per chapter, per section, per page and so on)
- position of footnotes defined inside admonitions
- switching footnote counter (cf. a, b, c...) in adomonitions (discussed in #2530)
- making working hyperlinked (clickable) footnotes from code captions, from inside tables, etc.

jfbu:
also the problem of allowing code-blocks in footnotes.

edit: this was fixed at b8e0e93 (Sphinx 1.5)

jfbu:
also the issue that sometimes there is a space token before the footnote in the LaTeX file, sometimes (in particular for automatically generated footnotes from using `latex_show_urls = 'footnote'`) there is none.

edit: this was fixed at 828c96ec (#3057) (Sphinx 1.5)

PR #8832, opened 2021-02-05
Latex hyperlinked caption footnotes on 3.x

jfbu:
Fix fifth item of #2616

I used this test file:

```rest
section with a footnote [#]_ in title
-------------------------------------

Footnote in term [#]_
    Description ...

    `Term in deflist <http://sphinx-doc.org/>`_
        Description2


.. figure:: image.png

   Figure caption with a footnote [#]_.


.. list-table:: Table caption [#]_
    :header-rows: 1

    * - name [#]_
      - description [#]_
    * - a
      - b


.. list-table:: LongTable caption [#]_
    :header-rows: 1
    :class: longtable

    * - name [#]_
      - description [#]_
    * - a
      - b [#]_



.. rubric:: Footnotes

.. [#] footnote in section title

.. [#] Footnote in term

.. [#] footnote in figure caption

.. [#] footnote in table caption

.. [#] footnote in table header

.. [#] second footnote in table header

.. [#] footnote in longtable caption

.. [#] footnote in longtable header

.. [#] second footnote in longtable header

.. [#] footnote in longtable cell

```

### Feature or Bugfix

- Bugfix

### DIFF
--- a/sphinx/texinputs/footnotehyper-sphinx.sty
+++ b/sphinx/texinputs/footnotehyper-sphinx.sty
@@ -325,8 +325,17 @@
 % some extras for Sphinx :
 % \sphinxfootnotemark: usable in section titles and silently removed from TOCs.
 \def\sphinxfootnotemark [#1]%
-   {\ifx\thepage\relax\else\protect\spx@opt@BeforeFootnote
-                             \protect\footnotemark[#1]\fi}%
+   {\ifx\thepage\relax\else\sphinxfootref{#1}\fi}%
+% \sphinxfootref: almost same as \footref from footmisc package as
+% re-defined by hyperref:
+% - \spx@opt@BeforeFootnote is from BeforeFootnote sphinxsetup option
+% - 'fn:' prefix is output by LaTeX writer in \label of footnotetext
+\protected\def\sphinxfootref#1{\spx@opt@BeforeFootnote
+    \begingroup
+      \unrestored@protected@xdef\@thefnmark{\ref{fn:#1}}% #1 = number
+    \endgroup
+    \H@@footnotemark
+}%
 \AtBeginDocument{%
    % let hyperref less complain
    \pdfstringdefDisableCommands{\def\sphinxfootnotemark [#1]{}}%
--- a/sphinx/texinputs/footnotehyper-sphinx.sty
+++ b/sphinx/texinputs/footnotehyper-sphinx.sty
@@ -176,7 +176,7 @@
     \@footnotemark
 % Sphinx addition: store internal hyperref anchor reference to allow
 % multiple clickable reference to same numbered footnote
-    \expandafter\xdef\csname FNH!\number\value{sphinxscope}!#1\endcsname{\Hy@footnote@currentHref}%
+    \expandafter\xdef\csname FNH!\thesphinxscope!#1\endcsname{\Hy@footnote@currentHref}%
     \def\FNH@endfntext@fntext{\@footnotetext}%
     \FNH@startfntext
 }%
@@ -345,18 +345,18 @@
 %     this <number> as argument and taking into account current <scope>.
 \protected\def\sphinxfootref#1{% #1 always explicit number in Sphinx usage
   \spx@opt@BeforeFootnote
-  \ifcsname FNH!\number\value{sphinxscope}!#1\endcsname
+  \ifcsname FNH!\thesphinxscope!#1\endcsname
      \gdef\@thefnmark{#1}% \@thefnmark always redefined globally by latex
      \let\spx@saved@makefnmark\@makefnmark
      \edef\@makefnmark{% we pre-expand the sphinxscope value, to be safe
-       \noexpand\hyper@linkstart{link}{\@nameuse{FNH!\number\value{sphinxscope}!#1}}%
+       \noexpand\hyper@linkstart{link}{\@nameuse{FNH!\thesphinxscope!#1}}%
        \noexpand\spx@saved@makefnmark
        \noexpand\hyper@linkend
      }%
      \H@@footnotemark
      \let\@makefnmark\spx@saved@makefnmark
   \else
-     \xdef\@thefnmark{\noexpand\ref{\number\value{sphinxscope}.#1}}% #1 = number
+     \xdef\@thefnmark{\noexpand\ref{\thesphinxscope.#1}}% #1 = number
      \H@@footnotemark
   \fi
 }%
--- a/sphinx/texinputs/footnotehyper-sphinx.sty
+++ b/sphinx/texinputs/footnotehyper-sphinx.sty
@@ -174,6 +174,9 @@
      \unrestored@protected@xdef\@thefnmark{\thempfn}%
     \endgroup
     \@footnotemark
+% Sphinx addition: store internal hyperref anchor reference to allow
+% multiple clickable reference to same numbered footnote
+    \expandafter\xdef\csname FNH@!#1\endcsname{\Hy@footnote@currentHref}%
     \def\FNH@endfntext@fntext{\@footnotetext}%
     \FNH@startfntext
 }%
@@ -326,15 +329,37 @@
 % \sphinxfootnotemark: usable in section titles and silently removed from TOCs.
 \def\sphinxfootnotemark [#1]%
    {\ifx\thepage\relax\else\sphinxfootref{#1}\fi}%
-% \sphinxfootref: almost same as \footref from footmisc package as
-% re-defined by hyperref:
+% \sphinxfootref: inspired by \footref from footmisc package as
+% re-defined by hyperref, but with an extra FNH@! branch:
 % - \spx@opt@BeforeFootnote is from BeforeFootnote sphinxsetup option
-% - 'fn:' prefix is output by LaTeX writer in \label of footnotetext
-\protected\def\sphinxfootref#1{\spx@opt@BeforeFootnote
-    \begingroup
-      \unrestored@protected@xdef\@thefnmark{\ref{fn:#1}}% #1 = number
-    \endgroup
-    \H@@footnotemark
+% - \ref{fn:#1} branch:
+%     the latex.py writer inserts at start of footnotetext contents
+%     \phantomsection\label{fn:<num>}, so we can use \ref which
+%     hyperref converts into an hyperlink. This is basically the
+%     idea of the \footref macro from package footmisc.
+% - FNH@! branch:
+%     explicitly numbered footnotes in latex.py output always use first
+%     the footnote environment, then \sphinxfootnotemark. We patched
+%     \FNH@footnoteenv@i to define the \FNH@!<number> macro as storage
+%     for the internal hyperref target id for the hyperref anchor. So
+%     we can then create an hyperlink from any \sphinxfootnotemark with
+%     this <number> as argument.
+\protected\def\sphinxfootref#1{%
+  \spx@opt@BeforeFootnote
+  \ifcsname FNH@!#1\endcsname
+     \def\@thefnmark{#1}%
+     \let\spx@saved@makefnmark\@makefnmark
+     \def\@makefnmark{%
+       \hyper@linkstart{link}{\@nameuse{FNH@!#1}}%
+       \spx@saved@makefnmark
+       \hyper@linkend
+     }%
+     \H@@footnotemark
+     \let\@makefnmark\spx@saved@makefnmark
+  \else
+     \def\@thefnmark{\ref{fn:#1}}% #1 = number
+     \H@@footnotemark
+  \fi
 }%
 \AtBeginDocument{%
    % let hyperref less complain
--- a/sphinx/texinputs/footnotehyper-sphinx.sty
+++ b/sphinx/texinputs/footnotehyper-sphinx.sty
@@ -174,9 +174,6 @@
      \unrestored@protected@xdef\@thefnmark{\thempfn}%
     \endgroup
     \@footnotemark
-% Sphinx addition: store internal hyperref anchor reference to allow
-% multiple clickable reference to same numbered footnote
-    \expandafter\xdef\csname FNH!\thesphinxscope!#1\endcsname{\Hy@footnote@currentHref}%
     \def\FNH@endfntext@fntext{\@footnotetext}%
     \FNH@startfntext
 }%
@@ -331,34 +328,54 @@
    {\ifx\thepage\relax\else\sphinxfootref{#1}\fi}%
 % \sphinxfootref:
 % - \spx@opt@BeforeFootnote is from BeforeFootnote sphinxsetup option
-% - \ref branch:
-%     the latex.py writer inserts at start of footnotetext contents
-%     \phantomsection\label{<scope>.<num>}, so we can use \ref which
-%     hyperref converts into an hyperlink. <scope> is the number of
-%     the scope and <num> the number of the footnote, local to scope.
-% - FNH! branch:
-%     explicitly numbered footnotes in latex.py output always use first
-%     the footnote environment, then \sphinxfootnotemark. We patched
-%     \FNH@footnoteenv@i to define the \FNH!<scope>!<number> macro as storage
-%     for the internal hyperref target id for the hyperref anchor. So
-%     we can then create an hyperlink from any \sphinxfootnotemark with
-%     this <number> as argument and taking into account current <scope>.
+% - \ref:
+%     the latex.py writer inserts a \phantomsection\label{<scope>.<num>}
+%     whenever
+%     - the footnote was explicitly numbered in sources,
+%     - or it was in restrained context and is rendered using footnotetext
+%
+%     These are the two types of footnotes that \sphinxfootnotemark must
+%     handle. But for explicitly numbered footnotes the same number
+%     can be found in document. So a secondary part in  <scope> is updated
+%     at each novel such footnote to know what is the target from then on
+%     for \sphinxfootnotemark and already encountered [1], or [2],...
+%
+% LaTeX package varioref is not supported by hyperref (from its doc: "There
+% are too many problems with varioref. Nobody has time to sort them out.
+% Therefore this package is now unsupported.") So we will simply use our own
+% macros to access the page number of footnote text and decide whether to print
+% it. \pagename is internationalized by latex-babel.
+\def\spx@thefnmark#1#2{%
+  % #1=label for reference, #2=page where footnote was printed
+  \ifx\spx@tempa\spx@tempb
+     % same page
+     #1%
+  \else
+    \sphinxthefootnotemark{#1}{#2}%
+  \fi
+}%
+\def\sphinxfootref@get #1#2#3#4#5\relax{%
+    \def\sphinxfootref@label{#1}%
+    \def\sphinxfootref@page {#2}%
+    \def\sphinxfootref@Href {#4}%
+}%
 \protected\def\sphinxfootref#1{% #1 always explicit number in Sphinx usage
   \spx@opt@BeforeFootnote
-  \ifcsname FNH!\thesphinxscope!#1\endcsname
-     \gdef\@thefnmark{#1}% \@thefnmark always redefined globally by latex
-     \let\spx@saved@makefnmark\@makefnmark
-     \edef\@makefnmark{% we pre-expand the sphinxscope value, to be safe
-       \noexpand\hyper@linkstart{link}{\@nameuse{FNH!\thesphinxscope!#1}}%
-       \noexpand\spx@saved@makefnmark
-       \noexpand\hyper@linkend
+  \ltx@ifundefined{r@\thesphinxscope.#1}%
+    {\gdef\@thefnmark{?}\H@@footnotemark}%
+    {\expandafter\expandafter\expandafter\sphinxfootref@get
+                 \csname r@\thesphinxscope.#1\endcsname\relax
+     \edef\spx@tempa{\thepage}\edef\spx@tempb{\sphinxfootref@page}%
+     \protected@xdef\@thefnmark{\spx@thefnmark{\sphinxfootref@label}{\sphinxfootref@page}}%
+     \let\spx@@makefnmark\@makefnmark
+     \def\@makefnmark{%
+       \hyper@linkstart{link}{\sphinxfootref@Href}%
+       \spx@@makefnmark
+       \hyper@linkend
      }%
      \H@@footnotemark
-     \let\@makefnmark\spx@saved@makefnmark
-  \else
-     \xdef\@thefnmark{\noexpand\ref{\thesphinxscope.#1}}% #1 = number
-     \H@@footnotemark
-  \fi
+     \let\@makefnmark\spx@@makefnmark
+    }%
 }%
 \AtBeginDocument{%
    % let hyperref less complain
--- a/sphinx/texinputs/footnotehyper-sphinx.sty
+++ b/sphinx/texinputs/footnotehyper-sphinx.sty
@@ -176,7 +176,7 @@
     \@footnotemark
 % Sphinx addition: store internal hyperref anchor reference to allow
 % multiple clickable reference to same numbered footnote
-    \expandafter\xdef\csname FNH@!#1\endcsname{\Hy@footnote@currentHref}%
+    \expandafter\xdef\csname FNH!\number\value{sphinxscope}!#1\endcsname{\Hy@footnote@currentHref}%
     \def\FNH@endfntext@fntext{\@footnotetext}%
     \FNH@startfntext
 }%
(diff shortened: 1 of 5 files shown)


sphinx/writers/latex.py
```diff
@@ -1738,7 +1738,8 @@ def depart_footnotemark(self, node: Element) -> None:
     def visit_footnotetext(self, node: Element) -> None:
         label = cast(nodes.label, node[0])
         self.body.append('%%\n\\begin{footnotetext}[%s]'
-                         '\\sphinxAtStartFootnote\n' % label.astext())
+                         '\\phantomsection\\label{fn:%s}%%\n'
```
jfbu: as ``%s`` is a number, I think ``fn:%s`` car never duplicate another label. However in case of doubt, we could simply add a space before ``fn``. And adjust the same in the latex code. Because never will Sphinx otherwise generate space prefixed labels, and a space prefixed label is not a problem for LaTeX.

jfbu:
This creates a problem with explicitly numbered footnotes which are referred to more than once.

```rest

.. rubric:: Footnotes

.. [1] A numerated footnote

First reference [1]_ 

Second reference [1]_
```
gives in latex

```latex
First reference %
\begin{footnote}[1]\sphinxAtStartFootnote
A numerated footnote
%
\end{footnote}

\sphinxAtStartPar
Second reference \sphinxfootnotemark[1]
```

and in PDF output the second reference becomes ``??`` because it has no known target. It tries to reach ``fn:1`` but it has not been defined.

jfbu:
I have fixed the issue with explicitly numbered footnotes which are referred to multiple times in the source. In the past only the first one was a clickable link, other references were not. The first commit of this PR made the situation worse in PDF, as a ``??`` appeared, but now, this is ok, and the number of the footnote is clickable to the correct location.

edit: one can see example in out own docs, in section on ``html_sidebars``, the sentence starting with "The keys can contain glob-style patterns" there is then a footnote, which with this PR is correctly hyperlinked. But at 3.x it is not hyperlinked.

Perhaps the approach here will be backported upstream into LaTeX footnotehyper package in future.

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -174,6 +174,9 @@
      \unrestored@protected@xdef\@thefnmark{\thempfn}%
     \endgroup
     \@footnotemark
+% Sphinx addition: store internal hyperref anchor reference to allow
+% multiple clickable reference to same numbered footnote
+    \expandafter\xdef\csname FNH@!#1\endcsname{\Hy@footnote@currentHref}%
```
jfbu: This kind of hack may in future become a feature of upstream footnotehyper LaTeX package

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -325,8 +328,39 @@
...
+%     \phantomsection\label{fn:<num>}, so we can use \ref which
+%     hyperref converts into an hyperlink. This is basically the
+%     idea of the \footref macro from package footmisc.
+% - FNH@! branch:
+%     explicitly numbered footnotes in latex.py output always use first
+%     the footnote environment, then \sphinxfootnotemark. We patched
+%     \FNH@footnoteenv@i to define the \FNH@!<number> macro as storage
+%     for the internal hyperref target id for the hyperref anchor. So
+%     we can then create an hyperlink from any \sphinxfootnotemark with
+%     this <number> as argument.
```
jfbu: Thus there are different things:
- when footnotetext mark-up is used, hyperref created no target anchor, and we use `\phantomsection` mark-up to add one with a label `\label{fn:number}`. I think this will not coliide with other labels,
- when footnote is used we don't change anything to latex.py output but add a LaTeX hack inside internal macro of footnotehyper-sphinx; this allows then for manually numbered footnotes which are referred to multiple times to be all hyperlinks. This pre-supposes that the first one in latex file is rendered by footnote environment, and all later ones are using `\sphinxfootnotemark` mark-up.

jfbu (commented):
@tk0miya is this true that for manually numbered footnotes, first one is with footnote environment and footnotemark nodes are only used for later usages? if yes, then this PR is fine. It is important that no footnotemark node is used before the location where actual footnote environment is used. For footnotes going via footnotetext path, this is not important because it uses `\label/\ref` approach.

jfbu:
Le 5 févr. 2021 à 20:47, Jean-François B. <notifications@github.com> a écrit :

> @tk0miya is this true that for manually numbered footnotes, first one is with footnote environment and footnotemark nodes are only used for later usages? 

I think it is true from reading the docstring of sphinx.builders.latex.transforms.LaTeXFootnoteTransform(). So we are good here I hope.

sphinx/writers/latex.py
tk0miya: I noticed docutils supports symbol-footnotes feature. It assigns non-numbers (ex. asterisk, daggar, etc.) to the footnotes. And it is not supported in our current implementation.
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#auto-symbol-footnotes

Additionally, we have an idea for the scope of footnote numbers (per chapter, per section, per page and so on). The duplicated footnote number will be shown if we'll support it in the future.

Therefore, it might be better to assign node_id to the footnote definitions and use it on footnote references.

tk0miya:
>is this true that for manually numbered footnotes, first one is with footnote environment and footnotemark nodes are only used for later usages

Yes. The latter references must be converted to the footnotemark node. It's one of the goal of LaTeXFootnoteTransform.
https://github.com/sphinx-doc/sphinx/blob/163c7bbdc12411818de1ce0204ff29cd911bcaf2/sphinx/builders/latex/transforms.py#L241-L271

sphinx/writers/latex.py
jfbu: In standard LaTeX (this is not modified by hyperref or footnotehyper) whenever ``[..]`` syntax is used with ``\footnote`` (or ``footnote`` environment with footnotehyper), ``\footnotemark``, ``\footnotetext`` (or rather ``footnotetext`` for us with footnotehyper), it is mandatorily a number, as it will be assigned to a TeX counter locally, either the footnote counter or the mpfootnote counter. This is done because LaTeX then uses one of the two macros `\thefootnote` or `\thempfootnote` which convert the counter into some printed form which can be customized (using for example roman numerals i, ii, iii, ...).

LaTeX has also a way to use symbols, and for this `\thefootnote` is defined to use `\@fnsymbol` which is a converter from numbers to symbols. The canonical sequence has only 9 symbols, starting at number 1 and gives  ∗ † ‡ § ¶ ‖ ∗∗ †† ‡‡. Probably there is some package to modify this hardcoded choice, or I can do the coding. But whatever happens, we can not use ``[...]`` syntax with something else (edit: I mean something non-numeric; in particular we can not put a symbol there, only a number representing the symbol)

Writing this, I have an idea, that we could use ``[...]`` with negative numbers for example ``[-1]``. Then we only have to modify ``\thefootnote``  to test if input is negative and in that case use a predefined symbol list. 

The other question you raise is the scope of footnote numbering.

- When using `\begin{footnote}[Number]`, anyhow hyperref uses an internal counter `Hfootnote` to label the anchor so there is not much problem: we keep memory of this anchor label in a macro indexed by Number (to allow `\sphinxfootnotemark[Number]` later to refer same footnote), so when scope of Sphinx footnote numbering is reset, the macros will start getting redefined by new footnotes with same number. Problem is that `\sphinxfootnotemark` first checks if the storage macro is defined, so we have to make them undefined at start of scope else there will be confusion with `\sphinxfootnotemark/footnotetext` usage. I will think about it. (I thougt about it by end of next item...)
- When using `\sphinxfootnotemark[N]` and `\begin{footnotetext}[N]` we can solve the scope problem by letting the `fn:` prefix be dynamical with the scope. It could be internally `\spx@fnprefix` and redefined at each scope reset via `\sphinxsetfnprefix{<some new unique prefix>}` from writers/latex.py output. In fact we could even simply have mark-up `\sphinxstepscope`, we would have a counter, at each scope change, it is incremented, the we use as prefix the value of the counter followed by some separator like `:`. If nothing else in Sphinx can generate `Number1:Number2` label then we are fine.

The idea of the scope counter can also be used to address issue in my first item, in place of making undefined all macros holding the hyperref internal anchor labels (because my code in `\sphinxfootref` tests is defined or not) at each scope change, we shall simply integrate the scope number into the storage macro names themselves so no duplicates can ever arise. Ok, this will work.

So I will push a modification to make this PR "scope"-ready. (memo:  `\sphinxstepscope` will have to be inserted before `\section` in future, in case a footnote is inside title).

As per making it "symbol"-ready, as I said, anyhow the LaTeX syntax requires `[..]` to contain always a number, so we will not put the symbol in there, but we can put a negative number and then have some LaTeX internals test sign of number and if negative use a symbol associated to it according to some dispatch macro (LaTeX's default one is `\@fnsymbol` and has only 9 predefined symbols but we can code ourselves the DocUtils specification; notice however that if symbols are Unicode we will have to let them be known to pdflatex, find suitable font holding them, and even with lualatex/xelatex, we have to be sure our font holds them, so there will be complications). Anyway symbols will never make it into labels per se, only the negative numbers which will serve to index them.

sphinx/texinputs/footnotehyper-sphinx.sty
jfbu: Here the #1 is anyhow required by the LaTeX core syntax to be something TeX recognizes as number. But here we use it into `\csname`, so it can not be a TeX count register. If we wanted maximum flexibility (as will probably be done upstream at footnotehyper once they get jealous of our work here), we would use `\the\numexpr#1\relax` here. But anyhow Sphinx mark-up uses exclusively explicit digits this is not needed.

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -174,6 +174,9 @@
      \unrestored@protected@xdef\@thefnmark{\thempfn}%
     \endgroup
     \@footnotemark
+% Sphinx addition: store internal hyperref anchor reference to allow
+% multiple clickable reference to same numbered footnote
+    \expandafter\xdef\csname FNH!\number\value{sphinxscope}!#1\endcsname{\Hy@footnote@currentHref}%
```
jfbu: in Sphinx mark-up ``#1`` is always an explicit number. It has to be something LaTeX recognizes as a number at any rate. Here in the `\csname`, it must be either explicit or expand to digits, it could not be a count register. Upstream at footnotehyper probably for maximal flexibility the will wrap it in `\the\numexpr#1\relax`.

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -325,8 +328,38 @@
...
+     \let\spx@saved@makefnmark\@makefnmark
+     \edef\@makefnmark{% we pre-expand the sphinxscope value, to be safe
+       \noexpand\hyper@linkstart{link}{\@nameuse{FNH!\number\value{sphinxscope}!#1}}%
+       \noexpand\spx@saved@makefnmark
+       \noexpand\hyper@linkend
+     }%
+     \H@@footnotemark
+     \let\@makefnmark\spx@saved@makefnmark
+  \else
+     \xdef\@thefnmark{\noexpand\ref{\number\value{sphinxscope}.#1}}% #1 = number
```
jfbu: We thus use labels of the type digits.digits (the `\label` will be from writers/latex.py)

I guess nothing else in Sphinx generates such labels. We could make even safer by adding a space upfront.

sphinx/texinputs/sphinx.sty
```diff
@@ -844,6 +844,8 @@
 
 %% FOOTNOTES
 %
+% Support scopes for footnote numbering
+\newcounter{sphinxscope}
```
jfbu, line 851: I will have to check this does not get lost when 3.x is merged later into master..

sphinx/writers/latex.py
```diff
@@ -1738,7 +1738,8 @@ def depart_footnotemark(self, node: Element) -> None:
     def visit_footnotetext(self, node: Element) -> None:
         label = cast(nodes.label, node[0])
         self.body.append('%%\n\\begin{footnotetext}[%s]'
-                         '\\sphinxAtStartFootnote\n' % label.astext())
+                         '\\phantomsection\\label{\\number\\value{sphinxscope}.%s}%%\n'
```
jfbu: This is where labels or the type digits.digits are generated.

Notice that `\label` macro forces expansion of its argument immediately so the sphinxscope value wil be the correct one, the one in force at that location.

Currently `sphinxscope` counter is not changed, it will always give 0. In future, if footnote numbering is reset, simply add some `\stepcounter{sphinxscope}` in mark-up at start of new scope (as footnotes can be in section titles, maybe before a `\section` then).

To support symbols as footnotes, we should use negative numbering and a LaTeX auxiliary which converts -1, -2, ... into the symbols. It is anyhow mandatory that ``[..]`` argument of footnotetext be explicit digits (or some expression TeX recognizes as a number, for example `\numexpr1+1\relax`).

jfbu (commented):
Update to allow in future scope reset of footnote numbering.

I tested this with

```rest
section with a footnote [#]_ in title
-------------------------------------

First reference [1]_ 


Footnote in term [#]_
    Description ...

    `Term in deflist <http://sphinx-doc.org/>`_
        Description2

Second reference [1]_

.. figure:: image.png

   Figure caption with a footnote [#]_.


.. list-table:: Table caption [#]_
    :header-rows: 1

    * - name [#]_
      - description [#]_
    * - a
      - b

Third reference [1]_

.. list-table:: LongTable caption [#]_
    :header-rows: 1
    :class: longtable

    * - name [#]_
      - description [#]_
    * - a
      - b [#]_



.. rubric:: Footnotes

.. [1] A numerated footnote

.. [#] footnote in section title

.. [#] Footnote in term

.. [#] footnote in figure caption

.. [#] footnote in table caption

.. [#] footnote in table header

.. [#] second footnote in table header

.. [#] footnote in longtable caption

.. [#] footnote in longtable header

.. [#] second footnote in longtable header

.. [#] footnote in longtable cell
```

sphinx/writers/latex.py
jfbu: > Additionally, we have an idea for the scope of footnote numbers (per chapter, per section, per page and so on). The duplicated footnote number will be shown if we'll support it in the future.

My idea here is fine for per chapter, per section, and to implement it we only need now to add `\stepcounter{sphinxscope}` at each scope change which will reset our generated footnote numbers which make it hard-coded into the latex file.

Per page requires a complete refactoring of our handling of footnote numbering, because only during pdf build will the page numbers be discovered. Currently this looks like difficult task. I think some improved upstream footnotehyper could help which will work around difficulties of LaTeX core since 30years with using multiple times  `\footnotemark` in a table for example and `footnotetext` after table (without even mentioning the added hyperlink complications). This is one of the reasons why in the past we opted for numbering footonotes via `[..]` output latex mark-up.

For per page, we could keep the same output actually (as the LaTeX code can choose to basically ignore it) but we need then some footnotehyper modifications. I don't quite see all problems in advance here, which will arise, and it will require some effort, it is not obvious task.

jfbu:
Moving this to draft status, as I am thinking about some improvements:
- create helper macros only if really needed, i.e. only for explicitly numbered footnotes in source, and not for all footnotes done with footnote environment
- the mark for multiply referred-to footnotes should incorporate the page number; although the link works, for print on paper if footnote body is on other page we should incorporate the info I guess; we could use package varioref `\vref` in place of `\ref`. varioref automatically translates "on page xxx" into used language, at least with babel, perhaps polyglossia (also it says "on previous page", "on next page"), but perhaps it is better to use our own translatable strings.
- I may be very close to a working solution for foonotes per page, and I would like to think it through so that this one prepares for it.

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -325,8 +325,58 @@
...
+%     whenever
+%     - the footnote was explicitly numbered in sources,
+%     - or it was in restrained context and is rendered using footnotetext
+%
+%     These are the two types of footnotes that \sphinxfootnotemark must
+%     handle. But for explicitly numbered footnotes the same number
+%     can be found in document. So a secondary part in  <scope> is updated
+%     at each novel such footnote to know what is the target from then on
+%     for \sphinxfootnotemark and already encountered [1], or [2],...
+%
```
jfbu, line 342: In this latest commit I use `\label` approach for all cases. But as I don't hack into internal footnotehyper macro, some way has to be found to enrich the label to avoid duplicates from identically numbered explicit footnotes. And the correct reference must be reconstituted from the sole indication of the number, and of `\thesphinxscope`, so the latter is enriched by a secondary part which changes at each new explicit footnote.

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -325,8 +325,58 @@
...
+% LaTeX package varioref is not supported by hyperref (from its doc: "There
+% are too many problems with varioref. Nobody has time to sort them out.
+% Therefore this package is now unsupported.") So we will simply use our own
+% macros to access the page number of footnote text and decide whether to print
+% it. \pagename is internationalized by latex-babel.
```
jfbu, line 347: I mention \pagename here but the code using it was moved to sphinx.sty

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -325,8 +325,58 @@
...
+    {\gdef\@thefnmark{?}\H@@footnotemark}%
+    {\expandafter\expandafter\expandafter\sphinxfootref@get
+                 \csname r@\thesphinxscope.#1\endcsname\relax
+     \edef\spx@tempa{\thepage}\edef\spx@tempb{\sphinxfootref@page}%
+     \protected@xdef\@thefnmark{\spx@thefnmark{\sphinxfootref@label}{\sphinxfootref@page}}%
+     \let\spx@@makefnmark\@makefnmark
+     \def\@makefnmark{%
+       \hyper@linkstart{link}{\sphinxfootref@Href}%
+       \spx@@makefnmark
+       \hyper@linkend
```
jfbu, line 374: this method gives a wider clickable area than using some `\hyperref[label]{text}` inside `\@thefnmark`.

sphinx/texinputs/footnotehyper-sphinx.sty
```diff
@@ -325,8 +325,58 @@
...
+% it. \pagename is internationalized by latex-babel.
+\def\spx@thefnmark#1#2{%
+  % #1=label for reference, #2=page where footnote was printed
+  \ifx\spx@tempa\spx@tempb
+     % same page
+     #1%
+  \else
+    \sphinxthefootnotemark{#1}{#2}%
+  \fi
+}%
```
jfbu, line 356: If the current page number is the same as the page number where the footnote body was printed, simply use the label, else use a customizable macro which by default will print something like "Page 79, 1", where 1 is the number of the footnote.

sphinx/texinputs/sphinx.sty
```diff
@@ -844,6 +844,28 @@
...
+\renewcommand{\thesphinxscope}{\number\value{sphinxscope}.\number\value{sphinxexplicit}}
+\newcommand\sphinxthefootnotemark[2]{%
+   % this is used to make reference to an explicitly numbered footnote not on same page
+   % #1=label of footnote text, #2=page number where footnote text was printed
+   \ifdefined\pagename
+      \pagename\space#2, % <- space
+    \else
+      p. #2, % <- space
+   \fi #1% no space
+}
```
jfbu, line 871: I did not how to proceed to make this internationalized; the babel latex has the \pagename, perhaps polyglossia too. There does not seem to be in babel a macro for abbreviated pagename.

sphinx/writers/latex.py
```diff
@@ -841,10 +841,15 @@ def depart_rubric(self, node: Element) -> None:
...
         label = cast(nodes.label, node[0])
+        if 'auto' not in node.attributes:
+            self.body.append('\\sphinxstepexplicit ')
         if self.in_parsed_literal:
             self.body.append('\\begin{footnote}[%s]' % label.astext())
         else:
             self.body.append('%%\n\\begin{footnote}[%s]' % label.astext())
+        if 'auto' not in node.attributes:
+            self.body.append('\\phantomsection'
+                             '\\label{\\thesphinxscope.%s}%%\n' % label.astext())
```
jfbu, line 858: In this function it seems on next line the addition of `\sphinxAtStartFootnote` did not really pay attention to whether in parsed_literal or not. Apparently in parsed_literal we want to avoid adding a `\n`. Maybe we should check that.

sphinx/writers/latex.py
```diff
@@ -841,10 +841,15 @@ def depart_rubric(self, node: Element) -> None:
     def visit_footnote(self, node: Element) -> None:
         self.in_footnote += 1
         label = cast(nodes.label, node[0])
+        if 'auto' not in node.attributes:
+            self.body.append('\\sphinxstepexplicit ')
```
jfbu, line 850: This will increment counter `sphinxexplicit`. In turn this will change `\thesphinxscope`. So never duplicate labels are created.

jfbu (commented):
In our own document we have a few example of explictly numbered footnote, and only one is referred to twice. Here is how it comes out now:
Capture d’écran 2021-02-06 à 20 25 03
This is on page 92 and the actual footnote text is on page 79 indeed. Notice that indication of page is only useful for print on paper. Because anyway the hyperlink is correct.

In a latter location, on page 225, there is another explictly numbered footnote. It comes out this way
Capture d’écran 2021-02-06 à 20 26 02
Because the text is on same page. The link is correct.

jfbu:
I have noticed that footnotes in header of a longtable will always indicate the page reference if the longtable has a page break. As longtable is always wrapped into  `savenotes`,  the footnotes are always delivered at the end. So on first page, it is correct to indicate page number. I think this is not an important blemish. Here is a screenshot (with language set to German):

start of long table
Capture d’écran 2021-02-06 à 22 40 36

next page (page 3)
Capture d’écran 2021-02-06 à 22 41 08
Footnotes are at bottom of this page.

sphinx/writers/latex.py
jfbu: > I noticed docutils supports symbol-footnotes feature. It assigns non-numbers (ex. asterisk, daggar, etc.) to the footnotes. And it is not supported in our current implementation.
> https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#auto-symbol-footnotes

I looked at it, and it is feasible in latex with same default symbols (the last four are available via math mode ``$\spadesuit$`` etc...). Once we support symbol-footnotes at core Sphinx, we can implement them in latex via using negative numbers for ``footnote, footnotetext, \sphinxfootnotemark``: ``[-1]`` to ``[-10]`` for the ten symbols, ``[-11]``, ``[-12]`` ... will mean to double the symbol, ``[-21]``, ..., ``[-30]``  will triple the symbol. It is all feasible. (we can only use ``[..]`` syntax with numbers, it is a core latex requirement).

sphinx/writers/latex.py
tk0miya: How about using the pair of docname and node_id instead? All footnotes already have their original node_id. So we can make a unique id from the combination with docname (ex. `fn:{docname}:{node_id}`).

I understand this counter also helps to generate a unique id. But it makes LaTeX output a bit complicated to me (this feeling may come from my experience of python and docutils).

sphinx/writers/latex.py
```diff
@@ -841,10 +841,15 @@ def depart_rubric(self, node: Element) -> None:
...
         if self.in_parsed_literal:
             self.body.append('\\begin{footnote}[%s]' % label.astext())
         else:
             self.body.append('%%\n\\begin{footnote}[%s]' % label.astext())
+        if 'auto' not in node.attributes:
```
tk0miya: docutils provides a shortcut to check the existence of an attribute: `in` and `not in`. So you don't need to use "attributes" attribute.
```suggestion
        if 'auto' not in node:
```

sphinx/writers/latex.py
jfbu: The LaTeX ouput with ``\thesphinxscope`` was in preparation for in future easying up resetting numbering of footnotes per chaper, section or page. However I will review the rationale for it and follow your suggestion. Right now I feel there was some necessity on LaTeX side, but I need to remember why.

tk0miya: Oops... I noticed the `footnote` nodes surely have their node_ids. But footnotetext and footnotemark do not. So another work is needed to adopt my idea. It would be better to adopt yours.

tk0miya (approved):

sphinx/writers/latex.py
jfbu: @tk0miya The problem is with our ``visit_footnotemark()``. It uses minimal LaTeX markup ``\sphinxfootnotemark[27]`` for example. Then LaTeX must know from this ``27`` what was the label used for original footnote. The way it is done is via maintaining information in the form of two numeric counters: `sphinxscope` (but was introduced only fror future usage), and `sphinxexplicit`. At any given time in latex document `\thesphinxscope` expands to `<num1>.<num2>` where `num1` is "scope" (always 0 currently) and `num2` is the number last modified when an explicit ``footnote`` was encountered. So the three digits `num1` (scope), `num2` (count of explicit footnotes), `N` (from `\sphinxfootnotemark[N]`)  allow to build a label which will correspond to a unique anchor point. I agree it is involved. Originally I was using only `N` but I had overlooked the problem that `N` can be the same when an explictly numbered footnote is referred to multiple times. So I added the `sphinxexplicit` and `\sphinxstepexplicit`  at each explicit footnote to increment it. In the meantime I had also add `sphinxscope` but only as abstraction for future usage.

I can remove `scope` part and only ever refer `explicit` part.

jfbu:
Thanks @tk0miya for reviewing this and help. I don't make it easy with all my numerous comments. They are to help in future if I forget something but already now I feel this is too much. I will merge this as it allows to make, I think, all footnotes now clickable to the correct target.

sphinx/writers/latex.py
tk0miya: Originally, my idea was using the pair of docname and node_id via `\label` system. So footnotemark node will be rendered as `\sphinxfootnotemark[27]{fn:docname:idXX}`, for example (like `\ref`). As a result, we don't need to use counters on LaTeX output. But, as I commented above, both footnotetext and footnotemark nodes don't have their node_id (or referencing node_id).

tk0miya:
I don't think the counters are not wrong. So it's not important to replace them with the node_id mechanism. So let's merge this into 3.x and ship the feature to users. It's not too late to refactor this again after releasing.

jfbu:
> I don't think the counters are not wrong. So it's not important to replace them with the node_id mechanism. So let's merge this into 3.x and ship the feature to users. It's not too late to refactor this again after releasing.

Also footnotehyper package might evolve and make things easier for us by handling labels internally. For example it should include a way to make multiply referred-to footnotes easily.

Thanks for time spent on reviewing, now merging!

tk0miya:
@jfbu Could you merge 3.x branch into master branch when you have time? Some merge operation will be needed for the merging. So it would be helpful to me if you'll do it.

jfbu:
@tk0miya Sure I will do it today as I have time now.

PR #8835, opened 2021-02-05
refactor: linkcheck: Separate thread manager feature from builder class

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To reduce the complexity of the linkcheck builder, this separates
the thread manager feature from the builder class as
HyperlinkAvailabilityChecker.
- refs: #8806

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -29,7 +29,9 @@ from requests.exceptions import HTTPError, TooManyRedirects
 
 from sphinx.application import Sphinx
 from sphinx.builders.dummy import DummyBuilder
+from sphinx.config import Config
 from sphinx.deprecation import RemovedInSphinx50Warning
+from sphinx.environment import BuildEnvironment
 from sphinx.locale import __
 from sphinx.transforms.post_transforms import SphinxPostTransform
 from sphinx.util import encode_uri, logging, requests
@@ -193,7 +195,9 @@ class CheckExternalLinksBuilder(DummyBuilder):
             RemovedInSphinx50Warning,
             stacklevel=2,
         )
-        return HyperlinkAvailabilityCheckWorker(self, None, None, {}).limit_rate(response)
+        worker = HyperlinkAvailabilityCheckWorker(self.env, self.config,
+                                                  None, None, {})
+        return worker.limit_rate(response)
 
     def rqueue(self, response: Response) -> queue.Queue:
         warnings.warn(
@@ -283,7 +287,7 @@ class CheckExternalLinksBuilder(DummyBuilder):
         self.json_outfile.write('\n')
 
     def finish(self) -> None:
-        checker = HyperlinkAvailabilityChecker(self)
+        checker = HyperlinkAvailabilityChecker(self.env, self.config, self)
         logger.info('')
 
         with open(path.join(self.outdir, 'output.txt'), 'w') as self.txt_outfile,\
@@ -296,9 +300,14 @@ class CheckExternalLinksBuilder(DummyBuilder):
 
 
 class HyperlinkAvailabilityChecker:
-    def __init__(self, builder: CheckExternalLinksBuilder) -> None:
+    def __init__(self, env: BuildEnvironment, config: Config,
+                 builder: CheckExternalLinksBuilder = None) -> None:
+        # Warning: builder argument will be removed in the sphinx-5.0.
+        # Don't use it from extensions.
+        # tag: RemovedInSphinx50Warning
         self.builder = builder
-        self.config = builder.config
+        self.config = config
+        self.env = env
         self.rate_limits = {}  # type: Dict[str, RateLimit]
         self.workers = []  # type: List[Thread]
 
@@ -313,8 +322,9 @@ class HyperlinkAvailabilityChecker:
 
     def invoke_threads(self) -> None:
         for i in range(self.config.linkcheck_workers):
-            thread = HyperlinkAvailabilityCheckWorker(self.builder, self.rqueue, self.wqueue,
-                                                      self.rate_limits)
+            thread = HyperlinkAvailabilityCheckWorker(self.env, self.config,
+                                                      self.rqueue, self.wqueue,
+                                                      self.rate_limits, self.builder)
             thread.start()
             self.workers.append(thread)
 
@@ -349,10 +359,14 @@ class HyperlinkAvailabilityChecker:
 class HyperlinkAvailabilityCheckWorker(Thread):
     """A worker class for checking the availability of hyperlinks."""
 
-    def __init__(self, builder: CheckExternalLinksBuilder, rqueue: queue.Queue,
-                 wqueue: queue.Queue, rate_limits: Dict[str, RateLimit]) -> None:
-        self.config = builder.config
-        self.env = builder.env
+    def __init__(self, env: BuildEnvironment, config: Config, rqueue: queue.Queue,
+                 wqueue: queue.Queue, rate_limits: Dict[str, RateLimit],
+                 builder: CheckExternalLinksBuilder = None) -> None:
+        # Warning: builder argument will be removed in the sphinx-5.0.
+        # Don't use it from extensions.
+        # tag: RemovedInSphinx50Warning
+        self.config = config
+        self.env = env
         self.rate_limits = rate_limits
         self.rqueue = rqueue
         self.wqueue = wqueue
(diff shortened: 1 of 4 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -125,16 +125,6 @@ def init(self) -> None:
         # set a timeout for non-responding servers
         socket.setdefaulttimeout(5.0)
 
-        # create queues and worker threads
-        self.rate_limits = {}  # type: Dict[str, RateLimit]
-        self.wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue
-        self.rqueue = queue.Queue()  # type: queue.Queue
-        self.workers = []  # type: List[Thread]
```
tk0miya, line 127: `rate_limits` was added at 6b90a63f08 (v3.4.0). So I think we don't need to pay attention for the compatibility.
On the other hand, `rqueue` and `wqueue` (at c3c879f2c6, since v1.3) and `workers` (at 51852c0e87, since v1.1) were added at too many years ago. So we need to pay attention for the compabitily.

But I can't imagine the 3rd party extensions touch these attributes...

sphinx/builders/linkcheck.py
```diff
@@ -245,47 +235,73 @@ def write_linkstat(self, data: dict) -> None:
...
     """A worker class for checking the availability of hyperlinks."""
 
-    def __init__(self, builder: CheckExternalLinksBuilder) -> None:
+    def __init__(self, builder: CheckExternalLinksBuilder, rqueue: Queue, wqueue: Queue,
+                 rate_limits: Dict[str, RateLimit]) -> None:
         self.app = builder.app
         self.anchors_ignore = builder.anchors_ignore
         self.auth = builder.auth
         self.config = builder.config
         self.env = builder.env
```
tk0miya: After some refactoring, the `builder` argument will be no longer needed. So I'll pass `config` and `env` object directly to this constructor at the final step.

sphinx/builders/linkcheck.py
tk0miya: In my thoughts, nobody touches these attributes. So it's enough to add empty properties not to be crashed on updating. I know they will become not working if they really existed. But I can't imagine the extensions who install their custom workers or enqueue/dequeue to the checker queue.

sphinx/builders/linkcheck.py
tk0miya: Done.

sphinx/builders/linkcheck.py
francoisfreitag: Can you explain the reason why empty properties are better than a crash?

IMHO, a crash is explicit about the incompatibility of an extension, where an empty property allows results to be silently incorrect.

sphinx/builders/linkcheck.py
```diff
@@ -202,7 +191,33 @@ def limit_rate(self, response: Response) -> Optional[float]:
...
+
+    def rqueue(self, response: Response) -> queue.Queue:
+        warnings.warn(
+            "%s.%s is deprecated." % (self.__class__.__name__, "rqueue"),
+            RemovedInSphinx50Warning,
+            stacklevel=2,
+        )
+        return queue.Queue()
+
+    def worker(self, response: Response) -> List[Thread]:
```
francoisfreitag:
```suggestion
    def workers(self, response: Response) -> List[Thread]:
```

sphinx/builders/linkcheck.py
```diff
@@ -268,53 +283,99 @@ def write_linkstat(self, data: dict) -> None:
...
         if self._broken:
             self.app.statuscode = 1
 
+
+class HyperlinkAvailabilityChecker:
+    def __init__(self, env: BuildEnvironment, config: Config,
+                 builder: CheckExternalLinksBuilder = None) -> None:
+        # Warning: builder argument will be removed in the sphinx-5.0.
+        # Don't use it from extensions.
+        # tag: RemovedInSphinx50Warning
```
francoisfreitag, line 307: This could raise a `RemovedInSphinx50Warning` when builder is not None. (same remark for `HyperlinkAvailabilityCheckerWorker`)

sphinx/builders/linkcheck.py
```diff
@@ -268,53 +283,99 @@ def write_linkstat(self, data: dict) -> None:
...
-        self._redirected = builder._redirected
+        if builder:
+            # if given, fill the result of checks as cache
+            self._good = builder._good
+            self._broken = builder._broken
+            self._redirected = builder._redirected
+        else:
+            self._good = set()
+            self._broken = {}
+            self._redirected = {}
```
francoisfreitag: Perhaps add a comment to remove these attributes in version 5. Though it repeats the deprecated.rst document, it’s much more obvious to contributors reading the source.

francoisfreitag (commented):
Thanks a lot, this is looking great :heart:!
It nicely split responsibilities in the link checker, and clarifies the interface with the builder. That will certainly help future extension.

sphinx/builders/linkcheck.py
francoisfreitag: I realize there’ll be a `PendingDeprecationWarning`, which is ignored by default warning filters, so only users of the [Python development mode](https://docs.python.org/3/library/devmode.html#devmode) or who otherwise enabled these warnings see them.

tk0miya: >Can you explain the reason why empty properties are better than a crash?
>IMHO, a crash is explicit about the incompatibility of an extension, where an empty property allows results to be silently incorrect.

What should users do when crashed? Extensions will not be updated as soon. I think "not working" is much better than crashes for them.

>I realize there’ll be a PendingDeprecationWarning, which is ignored by default warning filters, so only users of the Python development mode or who otherwise enabled these warnings see them.

Yes. `RemovedInSphinx50Warning` is a subclass of `PendingDeprecationWarning now. But it will be changed to a subclass of `DeprecationWarning` since 4.x series. Then it will be shown for users by default. As a result, they noticed their extensions (or hacks) uses deprecated APIs and internal functions.

sphinx/builders/linkcheck.py
tk0miya: OMG. Thanks

sphinx/builders/linkcheck.py
tk0miya: At present, our `CheckExternalLinksBuilder` passes the builder argument always. It means linkcheck builder will raise a `RemovedInSphinx50Warning`.  I think it's very noisy for users.

sphinx/builders/linkcheck.py
```diff
@@ -268,53 +283,99 @@ def write_linkstat(self, data: dict) -> None:
...
+        self.rate_limits = {}  # type: Dict[str, RateLimit]
+        self.rqueue = queue.Queue()  # type: queue.Queue
+        self.wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue
+        self.workers = []  # type: List[Thread]
+
+        self.to_ignore = [re.compile(x) for x in self.config.linkcheck_ignore]
+
+    def invoke_threads(self) -> None:
+        for i in range(self.config.linkcheck_workers):
+            thread = HyperlinkAvailabilityCheckWorker(self.builder.env, self.config,
```
francoisfreitag:
```suggestion
            thread = HyperlinkAvailabilityCheckWorker(self.env, self.config,
```

sphinx/builders/linkcheck.py
```diff
@@ -268,53 +283,99 @@ def write_linkstat(self, data: dict) -> None:
...
+            thread.start()
+            self.workers.append(thread)
+
+    def shutdown_threads(self) -> None:
         self.wqueue.join()
-        # Shutdown threads.
         for worker in self.workers:
             self.wqueue.put((CHECK_IMMEDIATELY, None, None, None), False)
 
+    def check(self, hyperlinks: Dict[str, Hyperlink]) -> Generator[Tuple[str, str, int, str, str, int], None, None]:  # NOQA
```
francoisfreitag:
```suggestion
    def check(self, hyperlinks: Dict[str, Hyperlink]) -> Generator[CheckResult, None, None]:
```

francoisfreitag (commented):
Looking good, a couple minor things.

sphinx/builders/linkcheck.py
francoisfreitag: Make sense, thanks for explaining.

sphinx/builders/linkcheck.py
tk0miya: Finally, I determined to keep `builder.rqueue` and `builder.wqueue` as is (not removed, but deprecated) to merge this and step forward.

sphinx/builders/linkcheck.py
tk0miya: Done

tk0miya:

PR #8836, opened 2021-02-06
refactor: LaTeX: separate self.body.append() call to single line output

tk0miya:
### Feature or Bugfix
- Refactoring


sphinx/writers/latex.py
```diff
@@ -737,13 +736,15 @@ def visit_desc_signature(self, node: Element) -> None:
         if not node.get('is_multiline'):
             self._visit_signature_line(node)
         else:
-            self.body.append('%\n\\pysigstartmultiline\n')
+            self.body.append('%\n')
+            self.body.append('\\pysigstartmultiline\n')
```
tk0miya, line 743: I'll replace '\n` by constant on the next step. It allows us to use raw string for LaTeX macros.

jfbu (approved):

PR #8839, opened 2021-02-06
refactor: linkcheck: Remove next_check from Hyperlink object

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- To separate hyperlink itself and checking (intermediate) state, this
removes `next_check` attribute from Hyperlink object and add a new
named-tuple `CheckRequest`.  It was rejected idea on implementing
rate-limiting first (see #8467).  But it's better way to separate
large builder module into small components and make whole of linkcheck
builder simple.
- After this change, `Hyperlink` object represents a hyperlink on the
document.  It has a URI and location info (docname and lineno).
- refs: https://github.com/sphinx-doc/sphinx/pull/8467#discussion_r528217651
- Finally, I have to admit the first implementation of @francoisfreitag is good.

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -19,7 +19,8 @@ from email.utils import parsedate_to_datetime
 from html.parser import HTMLParser
 from os import path
 from threading import Thread
-from typing import Any, Dict, Generator, List, NamedTuple, Optional, Pattern, Set, Tuple, cast
+from typing import (Any, Dict, Generator, List, NamedTuple, Optional, Pattern, Set, Tuple,
+                    Union, cast)
 from urllib.parse import unquote, urlparse
 
 from docutils import nodes
@@ -42,10 +43,11 @@ logger = logging.getLogger(__name__)
 
 uri_re = re.compile('([a-z]+:)?//')  # matches to foo:// and // (a protocol relative URL)
 
-Hyperlink = NamedTuple('Hyperlink', (('next_check', float),
-                                     ('uri', Optional[str]),
-                                     ('docname', Optional[str]),
+Hyperlink = NamedTuple('Hyperlink', (('uri', str),
+                                     ('docname', str),
                                      ('lineno', Optional[int])))
+CheckRequest = NamedTuple('CheckRequest', (('next_check', float),
+                                           ('hyperlink', Optional[Hyperlink])))
 CheckResult = NamedTuple('CheckResult', (('uri', str),
                                          ('docname', str),
                                          ('lineno', int),
@@ -53,6 +55,9 @@ CheckResult = NamedTuple('CheckResult', (('uri', str),
                                          ('message', str),
                                          ('code', int)))
 RateLimit = NamedTuple('RateLimit', (('delay', float), ('next_check', float)))
+
+# Tuple is old styled CheckRequest
+CheckRequestType = Union[CheckRequest, Tuple[float, str, str, int]]
 
 DEFAULT_REQUEST_HEADERS = {
     'Accept': 'text/html,application/xhtml+xml;q=0.9,*/*;q=0.8',
@@ -123,7 +128,7 @@ class CheckExternalLinksBuilder(DummyBuilder):
         socket.setdefaulttimeout(5.0)
 
         # create queues and worker threads
-        self._wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue
+        self._wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue[CheckRequestType]
         self._rqueue = queue.Queue()  # type: queue.Queue
 
     @property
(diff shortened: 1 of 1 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -127,7 +128,7 @@ def init(self) -> None:
 
         # create queues and worker threads
         self.rate_limits = {}  # type: Dict[str, RateLimit]
-        self.wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue
+        self.wqueue = queue.PriorityQueue()  # type: queue.PriorityQueue[CheckRequest]
```
tk0miya: This changes the type of `wqueue`. But I have a plan to make this attribute a dummy (refs: https://github.com/sphinx-doc/sphinx/pull/8835/files#r571341453). So this does not cause a crash.

sphinx/builders/linkcheck.py
```diff
@@ -40,10 +40,11 @@
 
 uri_re = re.compile('([a-z]+:)?//')  # matches to foo:// and // (a protocol relative URL)
 
-Hyperlink = NamedTuple('Hyperlink', (('next_check', float),
-                                     ('uri', Optional[str]),
+Hyperlink = NamedTuple('Hyperlink', (('uri', Optional[str]),
```
tk0miya: Now I'm thinking to enhance this object more. It would be better to add `baseurl` and `anchor`. I hope they will make the latter step (ex. https://github.com/sphinx-doc/sphinx/pull/8830).

sphinx/builders/linkcheck.py
```diff
@@ -40,10 +40,11 @@
...
                                      ('docname', Optional[str]),
                                      ('lineno', Optional[int])))
+CheckRequest = NamedTuple('CheckRequest', (('hyperlink', Optional[Hyperlink]),
+                                           ('next_check', float)))
```
francoisfreitag: Since the check requests are used in the `PriorityQueue`, the comparison should first be on `next_check`, then `hyperlink`. It ensures they get the highest priority hyperlink (lower next_check) first.
```suggestion
CheckRequest = NamedTuple('CheckRequest', (('next_check', float))),
                                           ('hyperlink', Optional[Hyperlink]))
```

sphinx/builders/linkcheck.py
francoisfreitag: IMO, we should keep the `uri`. Keeping the entire URI guarantees we describe the URI in the same way it appeared in the user document (e.g. when logging results).

Ideally, the same checker thread should process all URI with the same `base_url`, so that the page is downloaded once, and all anchors are checked on it. Perhaps we could feed the `PriorityQueue` with tuples `(next_check, hyperlink_group)`, where `hyperlink_group` is a `List[Hyperlink]`? That way, the checker thread can identify there are several anchors to check, extract the list of anchors to check and parse the page only once.

tk0miya: Absolutely! I perfectly agree with you. Originally, I thought new Hyperlink object is a named-tuple like this:
```
Hyperlink = NamedTuple('Hyperlink', (('fulluri', Optional[str]),
                                     ('base_uri', Optional[str]),
                                     ('anchor', Optional[str]),
                                     ('docname', Optional[str]),
                                     ('lineno', Optional[int])))
```

And `CheckRequest` will be replaced with this:
```
CheckRequest = NamedTuple('CheckRequest', (('next_check', float),
                                           ('base_uri', Optional[str]),
                                           ('hyperlinks', Optional[List[Hyperlink]])))
```

sphinx/builders/linkcheck.py
tk0miya: I see.

sphinx/builders/linkcheck.py
francoisfreitag: Great! :+1:
Since `base_uri` and `anchor` are derived from URI, perhaps we can get away without storing the `base_uri` and `anchor`. We’ll see what’s convenient when we implement it.

sphinx/builders/linkcheck.py
```diff
@@ -40,10 +40,11 @@
 
 uri_re = re.compile('([a-z]+:)?//')  # matches to foo:// and // (a protocol relative URL)
 
-Hyperlink = NamedTuple('Hyperlink', (('next_check', float),
-                                     ('uri', Optional[str]),
+Hyperlink = NamedTuple('Hyperlink', (('uri', Optional[str]),
                                      ('docname', Optional[str]),
```
francoisfreitag: With the entire Hyperlink being optional, the threads can be shut down without passing `(None, None, None)` for `uri`, `docname` and `lineno`. That means `uri` and `docname` are no longer optional, right?

`HyperlinkAvailabilityCheckWorker.check()` body first checks `len(uri)`, so a uri of `None` would have broken linkcheck. Similarly, `is_ignored_uri` checks for a pattern in `uri` and would break if `uri` is None.

francoisfreitag (commented):
Looking great!

sphinx/builders/linkcheck.py
tk0miya: Absolutely!

sphinx/builders/linkcheck.py
tk0miya: I determined to keep compatibility with old versions. So old tuple is still accepted.

tk0miya:

PR #8841, opened 2021-02-06
Overloaded function signatures do not require a separating backslash

AWhetter:
### Feature or Bugfix
- Feature

### Purpose
autodoc supports reading multiple docstrings from the start of the docstring. It requires signatures to be separated by a backslash, but it is conventional to not include a backslash as can be seen in the CPython codebase (https://github.com/python/cpython/blob/c0f22fb8b3006936757cebb959cee94e285bc503/Objects/rangeobject.c#L156).
This change allows autodoc to find multiple signatures at the start of the docstring by reading lines from the docstring until it finds the first line that does not look like a signature.

### Detail
N/A

### Relates
- A similar change is being considered in pybind to support this convention: https://github.com/pybind/pybind11/issues/2619

### DIFF
--- a/doc/usage/extensions/autodoc.rst
+++ b/doc/usage/extensions/autodoc.rst
@@ -529,14 +529,18 @@ There are also config values that you can set:
    looks like a signature, use the line as the signature and remove it from the
    docstring content.
 
-   If the signature line ends with backslash, autodoc considers the function has
-   multiple signatures and look at the next line of the docstring.  It is useful
-   for overloaded function.
+   autodoc will continue to look for multiple signature lines,
+   stopping at the first line that does not look like a signature.
+   This is useful for declaring overloaded function signatures.
 
    .. versionadded:: 1.1
    .. versionchanged:: 3.1
 
       Support overloaded signatures
+
+   .. versionchanged:: 4.0
+
+      Overloaded signatures do not need to be separated by a backslash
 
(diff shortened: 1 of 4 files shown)


AWhetter:
Is there anything that I can do to help move this pull request forwards? Should I make an associated issue for it?

brenthuisman (approved):
lgtm!

doc/usage/extensions/autodoc.rst
```diff
@@ -506,15 +506,19 @@ There are also config values that you can set:
...
+   autodoc will continue to look for multiple signature lines,
+   stopping at the first line that does not look like a signature.
+   This is useful for declaring overloaded function signatures.
 
    .. versionadded:: 1.1
    .. versionchanged:: 3.1
 
       Support overloaded signatures
 
+   .. versionchanged:: 3.5
```
tk0miya: This will be merged and released as 4.0. Could you update this please?

tk0miya (changes requested):
Sorry for the late response. At first, I worried about the side effects of this PR. But it can be ignored (very little). So let's go forward. Could you update the document please?

AWhetter:
@tk0miya Thanks for your review. I have made the requested changes. I've also changed the branch to merge into as master, but let me know if that's incorrect.

tk0miya (approved):

tk0miya:

Issue #8852, opened 2021-02-08
Allow non-RST parsers to substitute the Locale transform


jpmckinney:
**Describe the bug**

https://github.com/sphinx-doc/sphinx/commit/79650f5827e146afe3ff50485f8a4bd6755f895d (#4938) added code that adds a line of hyphens under (for example) an admonition's title's text.

However, a line of hyphens is an RST-specific syntax for a heading. It does not indicate a heading in all parsers.

In Markdown, for example, a line of hyphens is a second-level heading https://spec.commonmark.org/0.29/#setext-headings. This can cause the heading level to jump from 0 to 2.

**To Reproduce**

I can create a test scenario, but the bug should be clear from reading the code, with the above context.

**Expected behavior**

For Markdown, we can just change the hyphen to an equals sign, which Markdown will interpret as a first-level heading.

For other parsers, I suppose there would need to be hooks for those parsers to convert the string to a heading.

tk0miya:
I'm interested what is

jpmckinney:
@tk0miya I think your comment got cut off?

Also, the general case isn’t fixed - #8853 just fixed one scenario.

tk0miya:
Oops, Sorry. My understanding is current Locale transform is developed only for reST translation. So it does not work well with other parsers. I merged #8853 to Sphinx because it does not affect translating of the reST document. But it does not mean that Locale transform will support the markdown translation. If your goal is translating the whole of markdown document, we need to develop the mechanism to switch a translation-transform for each parser.

jpmckinney:
Thanks, yes, that seems like the appropriate solution.

jpmckinney:
I'm thinking the Locale transform is mostly generic. There are just a few places where it constructs RST strings. I'm wondering if those lines can be moved to methods (e.g. `title`, `literal_block`, etc.). That way, another parser can subclass the Locale transform and override those methods.

n-peugnet:
If I understand correctly, all these issues come from the fact that in `publish_msgstr()` we parse the source message once again but without the context of the rest of the source file:

https://github.com/sphinx-doc/sphinx/blob/80d5396f9d434d7c71d6f7ac5fa727a972b69584/sphinx/transforms/i18n.py#L73-L78

[...]

PR #8853, opened 2021-02-08
i18n: Locale transform: Change heading syntax to work for both RST and Markdown

jpmckinney:
Bugfix related to #8852


tk0miya:
Thanks!

PR #8854, opened 2021-02-08
LaTeX: optionally apply a second forceful wrapping of long code lines

### LINKED ISSUE #8849: [LaTex] code-block printed out of margin
@jfbu 
this is not handled correctly currently: long hex strings
Screenshot from 2021-02-07 12-37-17

code:
````
DryGASCON128k56:

.. code-block:: shell

   $ python3 -m drysponge.drygascon128_aead e 000102030405060708090A0B0C0D0E0F101112131415161718191A1B1C1D1E1F202122232425262728292A2B2C2D2E2F3031323334353637 000102030405060708090A0B0C0D0E0F "" ""
   28830FE67DE9772201D254ABE4C9788D

````
link to rst file: [examples_cli.rst](https://github.com/sebastien-riou/DryGASCON/blob/3950d559f11e565745e6a0a4b536e4725db2d138/Implementations/drygasconv1_python3/docs/source/examples_cli.rst)

_Originally posted by @sebastien-riou in https://github.com/sphinx-doc/sphinx/issues/8686#issuecomment-774660642_

### PR DESCRIPTION
**Pull request by jfbu**
Closes #8849

To try it out:

```
latex_elements = {
    'sphinxsetup': "verbatimforcewraps",
}
```

This is a bit experimental as I need to confirm that Pygments LaTeXFormatter output, as we use it, always comply with the pre-conceptions which are summarized in the latex code comments.

For this input:

```
DryGASCON128k56:

.. code-block:: shell

   $ python3 -m drysponge.drygascon128_aead e 000102030405060708090A0B0C0D0E0F101112131415161718191A1B1C1D1E1F202122232425262728292A2B2C2D2E2F3031323334353637 000102030405060708090A0B0C0D0E0F "" ""
   28830FE67DE9772201D254ABE4C9788D

.. code-block:: shell

   $ python3 -m drysponge.drygascon128_aead e "000102030405060708090A0B0C0D0E0F101112131415161718191A1B1C1D1E1F202122232425262728292A2B2C2D2E2F3031323334353637" 000102030405060708090A0B0C0D0E0F "" ""
   28830FE67DE9772201D254ABE4C9788D
```

the following output is obtained in PDF:

Capture d’écran 2021-02-08 à 23 15 03

Non highlighted strings like in first example can break at each character. Highlighted strings only every four characters. That here it breaks without one or two characters going in margin or one or two spaces at end is luck. Here is with removing one character before on same line, we see now it breaks one character short of ending line.

Capture d’écran 2021-02-08 à 23 20 53

Actually I could make even highlighted strings have breakpoints at each character. But it makes me pain to let the computer suffer with lots of extra work.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -131,6 +131,7 @@ Bugs fixed
 * #8780: LaTeX: long words in narrow columns may not be hyphenated
 * #8788: LaTeX: ``\titleformat`` last argument in sphinx.sty should be
   bracketed, not braced (and is anyhow not needed) 
+* #8849: LaTex: code-block printed out of margin
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -131,7 +131,9 @@ Bugs fixed
 * #8780: LaTeX: long words in narrow columns may not be hyphenated
 * #8788: LaTeX: ``\titleformat`` last argument in sphinx.sty should be
   bracketed, not braced (and is anyhow not needed) 
-* #8849: LaTex: code-block printed out of margin
+* #8849: LaTex: code-block printed out of margin (see the opt-in LaTeX syntax
+  boolean :ref:`verbatimforcewraps <latexsphinxsetupforcewraps>` for use via
+  the :ref:`'sphinxsetup' <latexsphinxsetup>` key of ``latex_elements``)
 
 Testing
 --------
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -1185,27 +1185,46 @@
           \doublehyphendemerits\z@\finalhyphendemerits\z@
           \strut #1\strut}%
 }%
-% We implement an alternative to wrapping long code lines. This
-% alternative works only if the contents are in the expected
-% Pygments mark-up: i.e. some character escapes such as \PYGZdl{}
-% and the highlighting \PYG macro with always 2 arguments, and no
-% other macros. This means:
-% - the command prefix *must* be PYG
-% - the texcomments Pygments option *must* be set to False (it could
-% work by luck if True)
-% For non-highlighted tokens a break point is installed at each of them.
-% For highlighted tokens (i.e. in 2nd argument of \PYG) every four such
-% characters. \PYGZdl{} etc will count for 2, although corresponding to
-% only one. The result is that no line should be more than 3 characters
-% overfull.
-% First a measurement step is done of what would our standard wrapping
-% approach give. This is a bit tricky, cf TeX by Topic for the basic
-% dissecting technique, because TeX unfortunately when building a
-% vertical box does not store in an accessible way what was the maximal
-% line-width: the width of the box will be the set \hsize.
-% Anyway, if the max width exceed the linewidth by at least 4 character
-% widths, then we apply the "force wrapping" of previous paragraph,
-% else we apply our "standard wrapping".
+%
+% The normal line wrapping allows breaks at spaces and ascii non
+% letters, non digits. The \raggedright above means there will be
+% an overfilled line only if some non-breakable "word" was
+% encountered, which is longer than a line (it is moved always to
+% be on its own on a new line).
+% 
+% The "forced" line wrapping will parse the tokens to add potential
+% breakpoints at each character. As some strings are highlighted,
+% we have to apply the highlighting character per character, which
+% requires to manipulate the output of the Pygments LaTeXFormatter.
+%
+% Doing this at latex level is complicated. The contents should
+% be as expected: i.e. some active characters from
+% \sphinxbreaksviaactive, some Pygments character escapes such as
+% \PYGZdl{}, and the highlighting \PYG macro with always 2
+% arguments. No other macros should be there, except perhaps
+% zero-parameter macros. In particular:
+% - the texcomments Pygments option must be set to False
+%
+% With pdflatex, Unicode input gives multi-bytes characters
+% where the first byte is active. We support the "utf8" macros
+% only. "utf8x" is not supported.
+%
+% The highlighting macro \PYG will be applied character per
+% character. Highlighting via a colored background gives thus a
+% chain of small colored boxes which may cause some artefact in
+% some pdf viewers. Can't do anything here if we do want the line
+% break to be possible.
+%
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by jfbu** [comment]
This breaks with Unicode in the code-blocks.... I got a failure from our own sphinx.pdf due to the examples with directory trees, with a test where the "force wrap" method is to be used always. But then it worked with using ``latex_engine = 'lualatex'``, i.e. all code-blocks of our own docs were forced to use the experimental code and no error was raised. (even with the feature activated, not a single code-block in our pdf docs by itself triggers the experimental code, because all are already wrapping fine if needed).

File: sphinx/texinputs/sphinx.sty [hunk 1]
```diff
@@ -1171,13 +1172,102 @@
...
+    \next
+}%
+\def\spx@forcewrapPYG@ii#1{#1\futurelet\spx@nexttoken\spx@forcewrapPYG@i}%
+% Replace \PYG by itself applied to short strings of 4 characters at a time
+% and insert breakpoints in-between
+\def\spx@forcewrapPYG@PYG\PYG#1#2{%
+    \def\spx@PYGspec{{#1}}%
+    \spx@PYG#2\@empty\@empty\@empty\@empty\relax
+}%
+\def\spx@PYG#1#2#3#4{%
```
jfbu: The idea here is that we split the argument of the highlighting macro (which could for example do `\colorbox`) in small bits of 4 tokens, so that highlighting and breakpoints can both cohabit.

But in case of Unicode code points and latex_engine is "pdflatex", this will break because each of #1, #2, #3, #4 is an octet so we are possibly cutting  apart a multi-bytes Unicode character. It would be lots of complications to make it work.

However if Unicode code-points are frequent in source it is better to use ``'lualatex'``,  ``'xelatex'``,  and ``'uplatex'``. I did tests with ``'lualatex'`` and it works fine (although much slower than pdflatex).

sebastien-riou:
@jfbu 
Thanks, that seems to fix it indeed. I cannot really try that on my side (by lack of skills to modify my sphinx installation). looking forward to see it in a release.

jfbu:
I have added Unicode support for the ``'pdflatex'`` engine as well as a way to customize when forceful algorithm is triggered. In particular it is possible to have all codelines be systematically hard-wrapped at the available width, if desired.

Now, the output looks like this (I removed letters/digits to change length, so examples are only typographical): 

Capture d’écran 2021-02-09 à 11 31 00

In the above I added some non-ascii letters, to check with ``'pdflatex'`` engine. I used
```
latex_elements = {
    'fontenc': "\\usepackage[X2,LGR,T1]{fontenc}",
    'sphinxsetup': "verbatimforcewraps",
}
```
to allow Greek and Cyrillic. As the font changes, the character width also, and as a result, we see a letter sticking a bit in margin, but this is exceptional. If we typeset the same using Unicode engine, all is exactly aligned:

Capture d’écran 2021-02-09 à 11 31 43

and I used for this

```
latex_engine = 'xelatex'

latex_elements = {
#    'fontenc': "\\usepackage[X2,LGR,T1]{fontenc}",
    'sphinxsetup': "verbatimforcewraps",
}
```

Even with ``'pdflatex'`` all will be aligned with non-ascii letters not from Greek or Cyrillic alphabets which trigger a font substitution.

To hard-wrap all code-blocks at the maximal width:
```
latex_elements = {
    'sphinxsetup': "verbatimforcewraps, verbatimmaxunderfull=0",
}
```

sphinx/texinputs/sphinx.sty
```diff
@@ -1171,13 +1174,183 @@
...
+% break to be possible.
+%
+% First a measurement step is done of what would the standard line
+% wrapping give (i.e line breaks only at spaces and non-letter,
+% non-digit ascii characters), cf TeX by Topic for the basic
+% dissecting technique: TeX unfortunately when building a vertical
+% box does not store in an accessible way what was the maximal
+% line-width during paragraph building.
+%
+% If the max width exceed the linewidth by at least 4 character
```
jfbu: The ``3`` in ``4=3+1`` is in truth the value of verbatimmaxoverfull key to 'sphinxsetup'

jfbu:
The Changes only indicates in this PR that #8849  is fixed but makes no mention so far of the new sphinxsetup keys  ``verbatimforcewraps``, ``verbatimmaxoverfull``, ``verbatimmaxunderfull``. Perhaps a line should be added to "Features added" to says something like "possibility to hard-wrap code lines at the linewidth (see ``verbatimforcewraps``)". Else people reading only change log on new release might think #8849 is fixed without user setting, but for security I have set the default of ``verbatimforcewraps`` to inactive. (it will probably break people who add extra decoration to the code-blocks such as raw LaTeX mark-up inside the code-blocks, for example in the comments, using Pygments ``texcomments``)

jfbu:
> @jfbu
> Thanks, that seems to fix it indeed. I cannot really try that on my side (by lack of skills to modify my sphinx installation). looking forward to see it in a release.

@sebastien-riou 
To test you can start from some project and go the latex build directory. Overwrite the ``sphinx.sty`` there with the ``sphinx.sty``from this PR. You also need to open the ``.tex`` main document and replace the line ``\usepackage{sphinx}`` by the line ``\usepackage[verbatimforcewraps]{sphinx}``. Finally, from the command line in this directory emit ``make all-pdf`` or directly ``pdflatex <name of main tex file>``.

Due to potential frailty the ``verbatimforcewraps`` will not be default at 3.5.0 release. Once this release is done you need to add to your conf.py

```
latex_elements['sphinxsetup'] = "verbatimforcewraps"
```

CHANGES
```diff
@@ -131,6 +131,7 @@ Bugs fixed
 * #8780: LaTeX: long words in narrow columns may not be hyphenated
 * #8788: LaTeX: ``\titleformat`` last argument in sphinx.sty should be
   bracketed, not braced (and is anyhow not needed) 
+* #8849: LaTex: code-block printed out of margin
```
tk0miya: As you commented by yourself, it would be better to mention new configurations here.

tk0miya (approved):
LGTM with nits.

CHANGES
jfbu: Thanks for reviewing, done. I will merge after testing completes. The behaviour is "opt-in" so no change to users if not using the feature.

PR #8856, opened 2021-02-09
Add missing French translation

redelmann:
Subject: Add missing French translation for the sentence "Searching for multiple words only shows matches that contain all words."

### Feature or Bugfix
- Bugfix

### Purpose
The purpose is to add a missing French translation for "Searching for multiple words only shows matches that contain all words."
Currently, an English sentence appears on the search results page.


tk0miya:
Thank you for reporting. It has already been translated in transifex.com. So it will be updated in the next major release. Please wait for a while.
Thanks,

PR #8858, opened 2021-02-09
Refactor latex writer

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- replace \n\n by a constant BLANKLINE
- replace \n by a constant CR

### DIFF
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -59,6 +59,7 @@ ENUMERATE_LIST_STYLE = defaultdict(lambda: r'\arabic',
                                        'upperroman': r'\Roman',
                                    })
 
+CR = '\n'
 EXTRA_RE = re.compile(r'^(.*\S)\s+\(([^()]*)\)\s*$')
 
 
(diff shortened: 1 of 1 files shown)


sphinx/writers/latex.py
```diff
@@ -373,8 +375,8 @@ def __init__(self, document: nodes.document, builder: "LaTeXBuilder",
         if (self.config.language not in {None, 'en', 'ja'} and
                 'fncychap' not in self.config.latex_elements):
             # use Sonny style if any language specified (except English)
-            self.elements['fncychap'] = ('\\usepackage[Sonny]{fncychap}\n'
-                                         '\\ChNameVar{\\Large\\normalfont\\sffamily}\n'
+            self.elements['fncychap'] = ('\\usepackage[Sonny]{fncychap}' + CR +
+                                         '\\ChNameVar{\\Large\\normalfont\\sffamily}' + CR +
                                          '\\ChTitleVar{\\Large\\normalfont\\sffamily}')
```
tk0miya, line 380: This change allows us to use raw strings (`r'...'`) in the LaTeX writer. I believe it makes writing TeX macros simple.

jfbu: +1 for idea! If idea goes into tests/test_build_latex.py we will have initial work to unify, sometimes raw strings are used, but mostly note. Raw strings make it easier to check what happens in case of test failures and to adapt tests if things change.

jfbu (commented):
LGTM with a question. I noticed on a number of occasion this has separated a ``self.body.append(CR)``. Now there is ``depart_entry()`` for table cells with its ``if self.needs_linetrimming`` branch. No potential change from the now isolated `CR` appends on this?

tk0miya:
>No potential change from the now isolated CR appends on this?

For the CR code prepending the line, I separated it to the another `self.body.append()` call.
For the CR code on the end of the line (following the line), I replaced it with the concatenation inside `self.body.append()` call.

Following this rule, this change would not break the output (I wish).

sphinx/writers/latex.py
```diff
@@ -565,7 +566,8 @@ def visit_document(self, node: Element) -> None:
             self.first_document = 0
         elif self.first_document == 0:
             # ... and all others are the appendices
-            self.body.append('\n\\appendix\n')
+            self.body.append(CR)
+            self.body.append('\\appendix' + CR)
```
tk0miya: Side note: I thought adding `println()` (print line) helper as a alias of `self.body.append(s + CR)`. But I feel it's too much. So I withdrew the idea and use simple concatenation.

jfbu:
> > No potential change from the now isolated CR appends on this?
> 
> For the CR code prepending the line, I separated it to the another `self.body.append()` call.
> For the CR code on the end of the line (following the line), I replaced it with the concatenation inside `self.body.append()` call.
> 
> Following this rule, this change would not break the output (I wish).

hmm yes I see, the line trimming works by `pop()` so it will not be affected as the separated CR appending is masked by next one

jfbu:
> > > No potential change from the now isolated CR appends on this?
> > 
> > 
> > For the CR code prepending the line, I separated it to the another `self.body.append()` call.
> > For the CR code on the end of the line (following the line), I replaced it with the concatenation inside `self.body.append()` call.
> > Following this rule, this change would not break the output (I wish).
> 
> hmm yes I see, the line trimming works by `pop()` so it will not be affected as the separated CR appending is masked by next one

no wait, I confused myself. It examines the oldest append to temporarily redirected body. So this was my question, now that isolated CR are visible where there were not will this change anything? (I don't recall clearly why the line trimming in table cell)

tk0miya:
OMG. I've misread what `needs_linetrimming` doing. I thought it removes the single CR code *at the tail* of body. But, actually, it removes the single CR code *at the top* of body. Absolutely, this change would hit the rule.

Okay. I'll keep CR codes pretending the line as is (not separating). Thank you for your reviewing!

PR #8861, opened 2021-02-09
refactor: LaTeX: Assign node_ids to footnotemark and footnotetext

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- On converting footnote nodes for the LaTeX writer, node_ids for footnote
nodes are almost cleared.  This copies the original node_id to the
footnotemark node (as refid) and footnotetext node (as node_id).  It's
useful to improve footnote features.
- refs: #8832

### DIFF
--- a/sphinx/builders/latex/transforms.py
+++ b/sphinx/builders/latex/transforms.py
@@ -192,7 +192,7 @@ class LaTeXFootnoteTransform(SphinxPostTransform):
                   headings having footnotes
                   <footnote_reference>
                       1
-              <footnote ids="1">
+              <footnote ids="id1">
                   <label>
                       1
                   <paragraph>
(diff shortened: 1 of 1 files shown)


sphinx/builders/latex/transforms.py
```diff
@@ -203,11 +203,9 @@ class LaTeXFootnoteTransform(SphinxPostTransform):
           <section>
               <title>
                   headings having footnotes
-                  <footnotemark>
+                  <footnotemark refid="id1">
                       1
-              <footnotetext>
-                  footnote body
```
tk0miya, line 209: This node is duplicated.

tk0miya:
This allows us to get the node_id of `footnote` and `footnotetext` nodes via `node['ids']`, and the referencing node_id of `footnotemark` via `node['refid']`. I hope this will help modifying LaTeX writer in the future.

jfbu (approved):
LGTM!

If this had been already available, perhaps I would have organized #8832 otherwise (cf https://github.com/sphinx-doc/sphinx/pull/8832#discussion_r573019934). We can refactor latex later.

I noticed footnotetext does not get "backrefs" attribute, contrarily to what is already existing in footnote. It would be nice to have footnotetext get this also, then we can implement backref links in all cases.

tk0miya:
Thank you for comment. Okay, I'll take a look it soon.

PR #8865, opened 2021-02-10
[bugfix] SubstitutionDefinitionsRemove: only run for LaTeX builder

brechtm:
My PR #8183 includes an oversight that makes it have no effect. The post-transform was still applied when using builders other than the LaTeX builder. This fixes it.

I noticed that also _IndexInSectionTitleTransform_ is being applied for all builders. Perhaps that should be limited to the LaTeX builder as well?

### DIFF
--- a/sphinx/builders/latex/transforms.py
+++ b/sphinx/builders/latex/transforms.py
@@ -44,7 +44,7 @@ class SubstitutionDefinitionsRemover(SphinxPostTransform):
     default_priority = Substitutions.default_priority + 1
     builders = ('latex',)
 
-    def apply(self, **kwargs: Any) -> None:
+    def run(self, **kwargs: Any) -> None:
         for node in self.document.traverse(nodes.substitution_definition):
             node.parent.remove(node)
 


tk0miya:
Thanks!

>I noticed that also IndexInSectionTitleTransform is being applied for all builders. Perhaps that should be limited to the LaTeX builder as well?

I'll fix it on another PR.

jessegv:
@brechtm @tk0miya  I was having a "search" issue.  I narrowed it down to this PR.  If I revert this commit, then my issue is fixed, even on master.  The issue I was having was if I do a search for a keyword, the search results will show up in what looks like all my files, when in fact it should only be in 1 file.  Perhaps this SubstitutionDefinitionsRemove function "should" be run even for html builder.  Do you confirm same behavior?

brechtm:
@jessegv Thanks for reporting. When you say you search for a keyword, is that keyword a word you used in a substitution? Is this substitution definition included in the rst_prolog/rst_epilog? A minimal Sphinx project (along with the search term) showing this issue would be very helpful. You can open a new issue for this.

PR #8867, opened 2021-02-10
Update JavaScript stemmer code to the latest version of Snowball (v2.1.0)

mitya57:
The stemmer code that is currently used in Sphinx was based on an old version of [snowball](https://github.com/snowballstem/snowball) (it was added in 059a74c51e10f5a6 in 2015). Furthermore, it is not pure JS code, but rather “JSX” (not the widely known React JSX, but https://github.com/jsx/JSX, which is also dead since 2015).

Since then, there were [two snowball releases](https://github.com/snowballstem/snowball/releases): v2.0.0 in October 2019 and v2.1.0 recently, in January 2021. In snowballstem/snowball@ddbe1fa9d5da7db0, JSX was replaced with native JS that is much easier to maintain and understand.

I should also note that our Python code uses [snowballstemmer](https://pypi.org/project/snowballstemmer/) module which is also v2.1.0 now, as that module is based on the same snowball code base (so this pull request brings JS up to date with Python).

What I did in this pull request:

* I updated the non-minified JS files to the latest version from snowball. To generate them, I cloned snowball repository and ran `make dist_libstemmer_js`, then copied them from generated archive to Sphinx repository.

* The minified files were embedded in Python code. This makes the code ugly and difficult to maintain. For consistency with non-minified files, I created a separate directory (`minified-js`) and put the files there.

    The minified files were created from non-minified ones using these commands:
    ```bash
    npm install uglifyjs
    for f in $(cd sphinx/search/non-minified-js && ls); do ./node_modules/.bin/uglifyjs sphinx/search/non-minified-js/$f -m -o sphinx/search/minified-js/$f; done
    ```

    The `-m` option of `uglifyjs` enables name mangling, e.g. using short names for variables that are only used internally.

* To make the page load faster, the JS code is still embedded into `language_data.js`, not copied as its own file.

* Note: now there is `base-stemmer.js` which defines the base class for all stemmers. That file is copied/embedded alongside the language-specific code.

With this pull request, it will be also easier to add new languages. In addition to what Sphinx already has, Snowball now supports Arabic, Armenian, Basque, Catalan, Greek, Hindi, Indonesian, Irish, Lithuanian, Nepali, Serbian, Tamil and Yiddish languages. I did not add new languages here, just updated the existing ones.


sphinx/search/__init__.py
```diff
@@ -439,15 +439,31 @@ def context_for_searchtool(self) -> Dict[str, Any]:
...
         return {
-            'search_language_stemming_code': self.lang.js_stemmer_code,
+            'search_language_stemming_code': self.get_js_stemmer_code(),
             'search_language_stop_words': jsdump.dumps(sorted(self.lang.stopwords)),
             'search_scorer_tool': self.js_scorer_code,
             'search_word_splitter_code': js_splitter_code,
         }
 
-    def get_js_stemmer_rawcode(self) -> str:
+    def get_js_stemmer_rawcode(self) -> List[str]:
```
tk0miya: We need to keep the compatibility of the interface. How about adding a new method `get_js_stemmer_rawcodes()` (plural form) instead? Or, how about checking the type of `get_js_stemmer_rawcode()` in HTML builder?

tk0miya (changes requested):
Thank you as always. LGTM with nits.

sphinx/search/__init__.py
mitya57: I did not realize this is actually public API. Now I reverted this method to its previous code, and I am just copying `base-stemmer.js` in HTML builder, if it exists. Is this fine?

tk0miya: In the past, we received some claimes for changing non-public APIs. So I'd like to change the interface explicitly as possible.

I think HTML builder should not know about the filename of stemmers. So it's better to change to your first proposal.

* Add a new method `get_js_stemmer_rawcodes()`
* Make an old method `get_js_stemmer_rawcode()` empty (returns None)
* HTML builder calls both old `get_js_stemmer_rawcode()` and new `get_js_stemmer_rawcodes()` to copy stemmer files.

This does not breaks compatibility. And we can deprecate the old one in the next step.

mitya57: OK, done as you suggested.

tk0miya (approved):

tk0miya:
Thank you for update. Merged.

PR #8868, opened 2021-02-10
searchtools: Don't use slideDown to show search results.

### LINKED ISSUE #8562: Search crashes my browser (on docs.sqlalchemy.org)
When I use the search functionality on docs.sqlalchemy.org it typically starts displaying a huge number of results which in only a couple of seconds results in using GBs of RAM, which then results in my browser crashing when the system runs out of memory.

The documentation is created with Sphinx.

I guess an obvious solution would be to implement pagination, rather than loading unlimited amount of results in the page.

Firefox on Linux.

docs.sqlalchemy.org may be a special case here, as the documentation is so big it's pretty common to get a huge number of search results.

See also: https://github.com/sqlalchemy/sqlalchemy/issues/5790

emilio:
Subject: Don't use slideDown to show search results.

### Feature or Bugfix

- Bugfix

### Purpose

- Fix #8562.

### Detail

Fixes a performance issue with massive lists like #8562.

jQuery.slideDown is pretty bad perf-wise since it thrashes layout
multiple times on all browsers. This is specially bad on slow machines.

The 5ms animation is unnoticeable (it last less than an animation
frame!) and just burns CPU cycles unnecessarily in all browsers.

On Firefox this is specially bad because it hits a performance cliff
with list items / CSS counters
(https://bugzilla.mozilla.org/show_bug.cgi?id=1683910#c26 for all the
gory details) where it causes tons of counter recalc.

An alternative fix for the Firefox cliff would be to set overflow:
hidden on the list item, but I think removing the animation is probably
more sensible, given that as I said it's just burning CPU.

Fixes #8562

### Relates
- #8562

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -248,7 +248,7 @@ var Search = {
       // results left, load the summary and display it
       if (results.length) {
         var item = results.pop();
-        var listItem = $('<li style="display:none"></li>');
+        var listItem = $('<li></li>');
         var requestUrl = "";
         var linkUrl = "";
         if (DOCUMENTATION_OPTIONS.BUILDER === 'dirhtml') {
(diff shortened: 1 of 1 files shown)


tk0miya:
Thanks!

PR #8871, opened 2021-02-11
Update 3rd Party Pip Libraries 

Blendify:
Subject: Update libs to recent versions.

### Feature or Bugfix
- Feature
- Bugfix
- Refactoring

### Purpose
This patch can be split into smaller patches if needed. But mostly I am trying to see which tests might fail.

### Detail
- Improves syntax highlighting for many languages.
- Faster snowballstemmer

### Relates

It would be nice to target this at sphinx 4.0.0


setup.py
```diff
@@ -21,14 +21,14 @@
     'sphinxcontrib-htmlhelp',
     'sphinxcontrib-serializinghtml',
     'sphinxcontrib-qthelp',
-    'Jinja2>=2.3',
-    'Pygments>=2.0',
+    'Jinja2>=2.11, <2.12',
```
tk0miya: Ubuntu-20.04: v2.10
https://packages.ubuntu.com/focal/python3-jinja2

setup.py
```diff
@@ -21,14 +21,14 @@
...
+    'Pygments>=2.7,<2.8',
```
tk0miya: Ubuntu-20.04: v2.3.1
https://packages.ubuntu.com/focal/python3-pygments

setup.py
```diff
@@ -21,14 +21,14 @@
...
     'docutils>=0.14',
-    'snowballstemmer>=1.1',
-    'babel>=1.3',
+    'snowballstemmer>=2.1',
```
tk0miya: Ubuntu-20.04: v2.0.0
https://packages.ubuntu.com/focal/python3-snowballstemmer

setup.py
```diff
@@ -21,14 +21,14 @@
...
+    'babel>=2.9',
```
tk0miya: Ubuntu-20.04: v2.6.0
https://packages.ubuntu.com/focal/python3-snowballstemmer

setup.py
```diff
@@ -21,14 +21,14 @@
...
     'alabaster>=0.7,<0.8',
     'imagesize',
-    'requests>=2.5.0',
+    'requests>=2.25',
```
tk0miya: Ubuntu-20.04: v2.22.0
https://packages.ubuntu.com/focal/python3-requests

tk0miya (commented):
The new requirements conflict with Ubuntu-20.04. Is it intended? I feel this is very strict. It would be better to support active Ubuntu LTS.

Blendify:
The latest versions are not needed, however, some of the current versions are really old. I am fine targeting the versions in Ubuntu-20.04

tk0miya:
According to our policy, we should support Ubuntu-18.04 since 4.0 because 16.04 will be deprecated this April (In other words, we still supporting 16.04 at this moment!)

Blendify:
Those seem like very strict requirements, I don't see a reason to stick with libs from a 4 year old LTS platform. Especially for pip packages when the idea is that they can be easily installed and are cross platform.

tk0miya:
In the past, I experienced that users claim Sphinx does not run on their platform many times. I'd not like to waste time answering them. I think supporting Ubuntu LTS reduces them.

I'd like to know what is the advantage if we drop these old platforms. Do old versions prevent to improve?

Blendify:
Pygments is the only upgrade that is pressing me, newer versions add more languages
https://github.com/readthedocs/sphinx_rtd_theme/issues/1045

Snowball stemmer was updated in #8867 but the dependency was not updated.

These are the two I would like to press to be updated. The others can be addressed at a later date.

tk0miya:
I'd like to know "why dropping old versions", not "why installing newer ones". In my thought, our dependency list does not force you to use old versions. You can install the latest versions of these libraries as expected without changing the list. Please let me know why do you want to drop old versions.

Blendify:
Most users I would say are not aware of the dependency, I think for users it would make sense if installing the latest version of sphinx didn't install 5-year-old libraries. For me, I didn't even know they were this old until I started looking them up on pypi.org.

I think you should revisit the support policy, and instead support lib versions within the current "Hardware and maintenance updates" period. I may be a bit biased, I tend to live on the bleeding edge, however, sticking to 16.04's extended lifespan seems a bit far-fetched.

Also, this patch is intended for sphinx4, Ubuntu 21.04 will only have bundled packages for sphinx3.4.3. https://packages.ubuntu.com/hirsute/sphinx-doc

tk0miya:
>Most users I would say are not aware of the dependency, I think for users it would make sense if installing the latest version of sphinx didn't install 5-year-old libraries. For me, I didn't even know they were this old until I started looking them up on pypi.org.

I can't imagine who installs the newest Sphinx and old dependencies. Do you suppose many users have pinned the versions and upgrade only Sphinx?

>I think you should revisit the support policy, and instead support lib versions within the current "Hardware and maintenance updates" period. 

As I commented above, I suppose the change will increase the support costs. Who paid it? I'd not like to do it. So -1 for the change.

>however, sticking to 16.04's extended lifespan seems a bit far-fetched.

It seems you're misunderstanding. The target version of Sphinx-4.0 is Ubuntu-18.04 because the standard support will be ended this April. (it's the reason why we'll drop python 3.6 support)

Blendify:
After looking this over a bit I think you are right, closing this patch.

Issue #4249, opened 2017-11-20
PDF output: Pygments error highlighting increases line spacing in code blocks


jfbu:
#### Procedure to reproduce the problem

Apply following patch to silence Pygments exception on lexer error token:

```diff
diff --git a/sphinx/highlighting.py b/sphinx/highlighting.py
index eb309d82a..acaca89ad 100644
--- a/sphinx/highlighting.py
+++ b/sphinx/highlighting.py
@@ -44,8 +44,8 @@ lexers = dict(
     rest = RstLexer(stripnl=False),
     c = CLexer(stripnl=False),
 )  # type: Dict[unicode, Lexer]
-for _lexer in lexers.values():
-    _lexer.add_filter('raiseonerror')
+# for _lexer in lexers.values():
+#     _lexer.add_filter('raiseonerror')
 
 
 escape_hl_chars = {ord(u'\\'): u'\\PYGZbs{}',
@@ -136,7 +136,8 @@ class PygmentsBridge(object):
                                    location=location)
                     lexer = lexers['none']
                 else:
-                    lexer.add_filter('raiseonerror')
+                    pass
+                    # lexer.add_filter('raiseonerror')
 
         # trim doctest options if wanted
         if isinstance(lexer, PythonConsoleLexer) and self.trim_doctest_flags:
```

Then do make latexpdf with index.rst being

```rst
Welcome to FOO's documentation!
===============================

.. highlight:: coq

.. raw:: latex

   \begin{multicols}2

::

   Definition logeq_both_false {X Y : UU} : ¬X -> ¬Y -> (X <-> Y).
   Definition logeq_both_false {X Y : UU} : ¬X -> ¬Y -> (X <-> Y).
   Definition logeq_both_false {X Y : UU} : ¬X -> ¬Y -> (X <-> Y).
   Definition logeq_both_false {X Y : UU} : ¬X -> ¬Y -> (X <-> Y).
   Definition logeq_both_false {X Y : UU} : ¬X -> ¬Y -> (X <-> Y).
   Definition logeq_both_false {X Y : UU} : ¬X -> ¬Y -> (X <-> Y).


.. raw:: latex

   \makeatletter
   \def\PYG@tok@err {\def\PYG@bc##1{\strut ##1}}
   \makeatother

::

   Definition logeq_both_false {X Y : UU} : ¬X -> ¬Y -> (X <-> Y).
[...]

PR #8879, opened 2021-02-13
LaTeX: better workaround for a Pygments LaTeXFormatter issue (with Pygments up to 2.7.4)

### LINKED ISSUE #8874: LaTeX: the fix to two minor Pygments LaTeXFormatter output issues ignore Pygments style
**Describe the bug**

https://github.com/sphinx-doc/sphinx/blob/8b786382e45f5a6a3a5300ab3d157ea92323c4bd/sphinx/texinputs/sphinx.sty#L273-L284

goes some redefinition of latex macros from Pygments output. In the first case this is to keep baseline constant even in case of error tokens, in the second case to avoid a change to `\fboxsep` to propagate (I do not recall if this one is cosmetic only, but the first one is definitely needed).

But using for example ``pygments_style = 'autumn'`` in conf.py one sees that original pygments LaTeXFormatter output  is

```latex
\def\PYG@tok@cs{\let\PYG@it=\textit\def\PYG@tc##1{\textcolor[rgb]{0.00,0.00,0.67}{##1}}}
...
\def\PYG@tok@err{\def\PYG@tc##1{\textcolor[rgb]{1.00,0.00,0.00}{##1}}\def\PYG@bc##1{\setlength{\fboxsep}{0pt}\colorbox[rgb]{1.00,0.67,0.67}{\strut ##1}}}
...
```
<sub> (I have edited it to avoid irrelevant ``\expandafter\def\csname...`` low-level stuff which is a slight bug of Pygments because unneeded; the `@` character must have been allowed in latex macro names here, from the fact  that the mark-up uses for example `\PYG@it` with so scruples, so all the ``\expandafter\def...`` from Pygments output have no rationale; **edit:** it has rationale as some digits may creep in, but can be shortened to `\@namedef`  )</sub>

So, the hard-coded fixes which are added after loading of ``sphinxhighlight.sty`` overwrite style specs such as colors, and here the usage of italics for Comment.Special Pygments token type.

By the way, a PR should be done at Pygments to work usptream around the bug(s) and thus make those redefinitions unnecessary.

Blame: 6cb517dc2d857c3f3a1484ec0fc621ae43cfa453 (my bad, sorry, #4250)

**Related**

- #4249

**Environment info**
- Python version: 3.8.7
- Pygments version: 2.7.4
- Sphinx version: 3.4.3
- Extra tools: [e.g. Browser, tex or something else]

jfbu:
Fix #8874 

### Feature or Bugfix

- Bugfix

### Relates
- #4249

The upstream Pygments issue is fixed in their master branch for next release (current is 2.7.4).

So this tests for Pygments version and applies the fix only if needed.

Previously the Sphinx fix corrected the line spacing issue (which is infrequent because one has to use a Pygments style which utilizes "border"  for some tokens and a lexer which produces these token types; this is case for pygments style 'manni' in combination with the diff lexer, see https://github.com/pygments/pygments/pull/1708), but in a way overwriting the style specification. This PR does it better in a way obeying the style, and it moves the fix to better location away from sphinx.sty.


sphinx/highlighting.py
```diff
@@ -50,6 +52,20 @@
 _LATEX_ADD_STYLES = r'''
 \renewcommand\PYGZsq{\textquotesingle}
 '''
+# fix extra space between lines when Pygments highlighting uses \fcolorbox
+# add a {..} to limit \fboxsep scope, and force \fcolorbox use correct value
+# cf pygments #1708 which makes this unneeded for Pygments > 2.7.4
+_LATEX_ADD_STYLES_FIXPYG = r'''
+\makeatletter
+% fix for Pygments <= 2.7.4
+\let\spx@original@fcolorbox\fcolorbox
```
jfbu, line 61: This means `\fcolorbox` (from color/xcolor package) must exist when sphinxhighlight.sty is loaded by sphinx.sty. But this is the case as color/xcolor is loaded very early by sphinx.sty, before handling options. So we are safe here. (this is true also on 3.x even though sphinxhighlight.sty is loaded earlier, but I feel now latex changes must only go to master). One could add some `\AtBeginDocument` to be extra safe if some package modifies `\fcolorbox` but I feel this is unnecessary. Besides in future, this patch will not be used when people have updated their Pygments version.

sphinx/highlighting.py
```diff
@@ -50,6 +52,20 @@
...
+\def\spx@fixpyg@fcolorbox{\fboxsep-\fboxrule\spx@original@fcolorbox}
```
jfbu, line 62: we don't have to insert here a `\string -\fboxrule` in place of `-\fboxrule` as was done at Pygments 2.8.1 to work-around a minted (latex package using Pygments) compatibility issue

PR #8893, opened 2021-02-15
Fix #8885: html: AttributeError for CSS/JS files on html_context

### LINKED ISSUE #8885: `AttributeError: 'str' object has no attribute 'priority'` in Sphinx 3.5.0
Dask documentation builds started failing when picking up the latest Sphinx (3.5.0); pinning to `<3.5.0` seems to work around the issue. See https://github.com/dask/dask/issues/7216 / https://github.com/dask/dask/pull/7217:

> ```
> Exception occurred:
>   File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/sphinx/builders/html/__init__.py", line 1039, in <lambda>
>     ctx['css_files'].sort(key=lambda js: js.priority)
> AttributeError: 'str' object has no attribute 'priority'
> The full traceback has been saved in /tmp/sphinx-err-z6ie2z11.log, if you want to report the issue to the developers.
> Please also report this if it was a user error, so that a better error message can be provided next time.
> A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
> ```

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- Since 3.5.0, priority has been added to control CSS/JS files.  But it's
not working if projects installs custom CSS/JS files via `html_context`
instead of `html_css_files` and `html_js_files`.  This avoids the crash
to keep it "not broken".
- #8885

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -1036,7 +1036,7 @@ class StandaloneHTMLBuilder(Builder):
 
         # sort JS/CSS before rendering HTML
         ctx['script_files'].sort(key=lambda js: js.priority)
-        ctx['css_files'].sort(key=lambda js: js.priority)
+        ctx['css_files'].sort(key=lambda css: css.priority)
 
         try:
             output = self.templates.render(templatename, ctx)
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -1035,8 +1035,20 @@ class StandaloneHTMLBuilder(Builder):
             templatename = newtmpl
 
         # sort JS/CSS before rendering HTML
-        ctx['script_files'].sort(key=lambda js: js.priority)
-        ctx['css_files'].sort(key=lambda css: css.priority)
+        try:
+            # Convert script_files to list to support non-list script_files (refs: #8889)
+            ctx['script_files'] = sorted(list(ctx['script_files']), key=lambda js: js.priority)
+        except AttributeError:
+            # Skip sorting if users modifies script_files directly (maybe via `html_context`).
+            # refs: #8885
+            #
+            # Note: priority sorting feature will not work in this case.
+            pass
+
+        try:
+            ctx['css_files'] = sorted(list(ctx['css_files']), key=lambda css: css.priority)
+        except AttributeError:
+            pass
 
         try:
             output = self.templates.render(templatename, ctx)
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,8 @@ Bugs fixed
 ----------
 
 * #8884: html: minified js stemmers not included in the distributed package
+* #8885: html: AttributeError is raised if CSS/JS files are installed via
+  :confval:`html_context`
 * #8880: viewcode: ExtensionError is raised on incremental build after
   unparsable python module found
 


sphinx/builders/html/__init__.py
```diff
@@ -1035,8 +1035,21 @@ def hasdoc(name: str) -> bool:
...
+        try:
+            ctx['script_files'].sort(key=lambda js: js.priority)
+        except AttributeError:
+            logger.warning(__('Failed to sort script_files. Please use html_js_files '
+                              'configuration to set up JavaScript files.'),
+                           once=True)
+            # Note: warning will be removed at Sphinx-5.0  # RemovedInSphinx50Warning
+
+        try:
+            ctx['css_files'].sort(key=lambda js: js.priority)
```
webknjaz: Nitpick:
```suggestion
            ctx['css_files'].sort(key=lambda css: css.priority)
```

sphinx/builders/html/__init__.py
```diff
@@ -1035,8 +1035,21 @@ def hasdoc(name: str) -> bool:
             templatename = newtmpl
 
         # sort JS/CSS before rendering HTML
-        ctx['script_files'].sort(key=lambda js: js.priority)
-        ctx['css_files'].sort(key=lambda js: js.priority)
+        try:
+            ctx['script_files'].sort(key=lambda js: js.priority)
```
webknjaz: Unrelated to this PR but you may consider just using something like
```suggestion
            get_priority = operator.attrgetter('priority')
            ctx['script_files'].sort(key=get_priority)
```
in the future.

sphinx/builders/html/__init__.py
```diff
@@ -1035,8 +1035,21 @@ def hasdoc(name: str) -> bool:
             templatename = newtmpl
 
         # sort JS/CSS before rendering HTML
```
webknjaz, line 1037: Have you considered doing something like
```suggestion
        # sort JS/CSS before rendering HTML
        ctx['script_files'] = list(ctx['script_files'])
        ctx['css_files'] = list(ctx['css_files'])

```
here?

sphinx/builders/html/__init__.py
tk0miya: +1: I'll do it later.

sphinx/builders/html/__init__.py
tk0miya: Acceptable.

sphinx/builders/html/__init__.py
```diff
@@ -1035,8 +1035,21 @@ def hasdoc(name: str) -> bool:
...
-        ctx['script_files'].sort(key=lambda js: js.priority)
-        ctx['css_files'].sort(key=lambda js: js.priority)
+        try:
+            ctx['script_files'] = sorted(list(ctx['script_files']), key=lambda js: js.priority)
+        except AttributeError:
```
webknjaz, line 1041: I don't think that with this change an `AttributeError` will ever happen... Well, it may happen on `js.priority` but that's a different problem.

tk0miya: Yes. That is #8885.

sphinx/builders/html/__init__.py
```diff
@@ -1035,8 +1035,20 @@ def hasdoc(name: str) -> bool:
...
-        ctx['css_files'].sort(key=lambda js: js.priority)
+        try:
+            # Convert script_files to list to support non-list script_files (refs: #8889)
+            ctx['script_files'] = sorted(list(ctx['script_files']), key=lambda js: js.priority)
+        except AttributeError:
+            # Skip sorting if users modifies script_files directly (maybe via `html_context`).
+            # refs: #8885
+            #
+            # Note: priority sorting feature will not work in this case.
+            pass
```
tk0miya, line 1046: Finally, I determined to keep the list as is and not emitting a warning. It means priority sort will not work. But no extensions do not use it yet.
And I'll deprecate `html_context['*_files]'` in another step to migrate to `html_*_files`.

sphinx/builders/html/__init__.py
```diff
@@ -1035,8 +1035,20 @@ def hasdoc(name: str) -> bool:
             templatename = newtmpl
 
         # sort JS/CSS before rendering HTML
-        ctx['script_files'].sort(key=lambda js: js.priority)
-        ctx['css_files'].sort(key=lambda js: js.priority)
+        try:
```
webknjaz, line 1038: FYI Python 3 has this nice pattern for such cases:
```suggestion
        with contextlib.suppress(AttributeError):
```

PR #8898, opened 2021-02-15
sphinx.ext.extlinks: Allow ``%s`` in link caption string

ghost:
Subject: sphinx.ext.extlinks: Allow ``%s`` in link caption string

### Feature or Bugfix
- Feature

### Purpose
- Allow ``%s`` in link caption string of the sphinx.ext.extlinks extensions, which allows arbitrary placement of the role content in the link caption.
- LT;DR: We replace `prefix + rolecontent` with `prefix % rolecontent`.

### Detail
Currently sphinx.ext.extlinks ("extlinks") allows placing the role contents (often a number)
only at the end of the link caption, like so "issue 123".  However, sometimes it is needed
to place the number (or text in general) at a different location, e.g. "1. quarter", "xyz house".

This patch changes the behaviour of the second part in the extlinks config. The caption
string now allows `%s` which gets replaced with the role content (the same substitution
which happens for base URL).  Until now this string was called *prefix* this is changed to
*caption*.  (In fact most of the changes in this diff are renaming *prefix* to *caption*.) If
you feel *title* is a more appropriate name, go ahead and complain.

Example:

`conf.py`:
```
extlinks = {'quarter': ('https://example.org/quaters/%s', '%s. quarter')}
```

`index.rst`
```
These are the results of the :quarter:`2`.
```

Output
```
These are the results of the 2. quarter.
```

### Note on backwards compatibility:
It is now required that the second entry in the tuple is either None or a string which contains
exactly one `%s`.  If this is not the case, then a warning is printed and we fall back to the
old behaviour, i.e. concatenating the caption label with the role content.  This is the same
mechanism as for the base URL.  Thus, module a warning this is backwards compatible.
I've filed this merge request against 3.5.x, if you feel this is too much breakage, please
reassign to merge with 3.x.

### DIFF
--- a/doc/usage/extensions/extlinks.rst
+++ b/doc/usage/extensions/extlinks.rst
@@ -23,29 +23,28 @@ The extension adds a config value:
 .. confval:: extlinks
 
    This config value must be a dictionary of external sites, mapping unique
-   short alias names to a base URL and a *prefix*.  For example, to create an
+   short alias names to a base URL and a *caption*.  For example, to create an
    alias for the above mentioned issues, you would add ::
 
       extlinks = {'issue': ('https://github.com/sphinx-doc/sphinx/issues/%s',
-                            'issue ')}
+                            'issue %s')}
 
    Now, you can use the alias name as a new role, e.g. ``:issue:`123```.  This
    then inserts a link to https://github.com/sphinx-doc/sphinx/issues/123.
    As you can see, the target given in the role is substituted in the base URL
    in the place of ``%s``.
 
-   The link *caption* depends on the second item in the tuple, the *prefix*:
+   The link caption depends on the second item in the tuple, the *caption*:
 
-   - If the prefix is ``None``, the link caption is the full URL.
-   - If the prefix is the empty string, the link caption is the partial URL
-     given in the role content (``123`` in this case.)
-   - If the prefix is a non-empty string, the link caption is the partial URL,
-     prepended by the prefix -- in the above example, the link caption would be
+   - If *caption* is ``None``, the link caption is the full URL.
+   - If *caption* is a string, then it must contain ``%s`` exactly once.  In
+     this case the link caption is *caption* with the partial URL substituted
+     for ``%s`` -- in the above example, the link caption would be
      ``issue 123``.
 
    You can also use the usual "explicit title" syntax supported by other roles
    that generate links, i.e. ``:issue:`this issue <123>```.  In this case, the
-   *prefix* is not relevant.
+   *caption* is not relevant.
 
 .. note::
 
--- a/doc/usage/extensions/extlinks.rst
+++ b/doc/usage/extensions/extlinks.rst
@@ -23,7 +23,7 @@ The extension adds a config value:
 .. confval:: extlinks
 
    This config value must be a dictionary of external sites, mapping unique
-   short alias names to a base URL and a *caption*.  For example, to create an
+   short alias names to a *base URL* and a *caption*.  For example, to create an
    alias for the above mentioned issues, you would add ::
 
       extlinks = {'issue': ('https://github.com/sphinx-doc/sphinx/issues/%s',
@@ -31,7 +31,7 @@ The extension adds a config value:
 
    Now, you can use the alias name as a new role, e.g. ``:issue:`123```.  This
    then inserts a link to https://github.com/sphinx-doc/sphinx/issues/123.
-   As you can see, the target given in the role is substituted in the base URL
+   As you can see, the target given in the role is substituted in the *base URL*
    in the place of ``%s``.
 
    The link caption depends on the second item in the tuple, the *caption*:
@@ -41,6 +41,11 @@ The extension adds a config value:
      this case the link caption is *caption* with the partial URL substituted
      for ``%s`` -- in the above example, the link caption would be
      ``issue 123``.
+
+   To produce a literal ``%`` in either *base URL* or *caption*, use ``%%``::
+
+        extlinks = {'wikipedia': ('https://en.wikipedia.org/w/index.php?search=%s&title=Special%%3ASearch&fulltext=Search&ns0=1'
+                                'Search %s on Wikipedia')}
 
    You can also use the usual "explicit title" syntax supported by other roles
    that generate links, i.e. ``:issue:`this issue <123>```.  In this case, the
--- a/doc/usage/extensions/extlinks.rst
+++ b/doc/usage/extensions/extlinks.rst
@@ -44,8 +44,8 @@ The extension adds a config value:
 
    To produce a literal ``%`` in either *base URL* or *caption*, use ``%%``::
 
-        extlinks = {'wikipedia': ('https://en.wikipedia.org/w/index.php?search=%s&title=Special%%3ASearch&fulltext=Search&ns0=1'
-                                'Search %s on Wikipedia')}
+      extlinks = {'KnR': ('https://example.org/K%%26R/page/%s',
+                            '[K&R; page %s]')}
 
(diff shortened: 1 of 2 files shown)


ghost:
Is there any interest in picking this up?

doc/usage/extensions/extlinks.rst
```diff
@@ -23,29 +23,28 @@ The extension adds a config value:
...
-   The link *caption* depends on the second item in the tuple, the *prefix*:
+   The link caption depends on the second item in the tuple, the *caption*:
 
-   - If the prefix is ``None``, the link caption is the full URL.
-   - If the prefix is the empty string, the link caption is the partial URL
-     given in the role content (``123`` in this case.)
-   - If the prefix is a non-empty string, the link caption is the partial URL,
-     prepended by the prefix -- in the above example, the link caption would be
+   - If *caption* is ``None``, the link caption is the full URL.
+   - If *caption* is a string, then it must contain ``%s`` exactly once.  In
```
tk0miya, line 40: The requirement of "%s" must be a breaking change. Do you intend to all of extlinks users should update their configuration after the upgrade? I think it would be better to behaves as prefix if the caption does not contain any "%s".

sphinx/ext/extlinks.py
```diff
@@ -50,18 +50,25 @@ def role(typ: str, rawtext: str, text: str, lineno: int,
...
         if not has_explicit_title:
-            if prefix is None:
+            if caption is None:
                 title = full_url
             else:
-                title = prefix + part
+                try:
+                    title = caption % part
+                except (TypeError, ValueError):
+                    inliner.reporter.warning(
```
tk0miya: Could you use `sphinx.util.logging` module instead?

tk0miya (changes requested):
Sorry for the late response. Good idea. But it would be better to keep compatibility. Could you reconsider the behavior?

sphinx/ext/extlinks.py
ghost: Sure, I'll then also update the other ``inliner.reporter.warning`` usages in this extension.

I looked at `sphinx.util.logging` but it seems that in order to produce a location i need a filename and a line number.
Sadly, the ``role()`` function only gets passed the line number but no filename.  How do I gain access to this information?

Sidenote: Even better would be getting the filename and line number of the config file where the corresponding `extlinks` config-entry is "declared" (last assigned(?)) during the setup-phase but that seems to be impossible as the config is a python script...

ghost:
Hi @tk0miya

Thanks for coming back to me.

@tk0miya in the review:

> The requirement of "%s" must be a breaking change. Do you intend to all of extlinks users should update their configuration after the upgrade? I think it would be better to behaves as prefix if the caption does not contain any "%s".

For an old config the behaviour of Sphinx is the same as previous (i.e. use it as a prefix) except additionally a warning will be printed.  In particular, Sphinx generates the same documentation output as previously, so there will be no breakage and no one is forced to upgrade their config to continue building their project.  Instead, we gently tip them off to the newer syntax using the warning, so they might update if they fell like it.

This is the same change as was done when upgrading the URL-string to allow a `%s` (see commit 6b13a28c1152b014a4e7d0ac27f8f5b960be3d75).  Maybe have a look at not only the diff but at the resulting file, where you will see that my change (wrt ``caption``) exactly mimics the behaviour wrt ``full_url``.  Wouldn't it be strange to let these two parameters warn differently wrt to substitution?

I admit the documentation is strongly worded with a *"must have"* and *"exactly one"*, if you like, I rephrase that to the more handwavy phrasing used for the URL part (which I didn't touch but has exactly the behaviour now proposed for ``caption``), namely *"the target given in the role is substituted in the caption in the place of ``%s``"*.  However, I do not like that style for documentation as it leaves the use in the dark what happens in other cases or whether these other cases are considered "valid".

If you still feel that we should (silently) fall back to the old behaviour instead of printing a warning.  Then I'll redo that part and the documentation.  Maybe we can add a *"Falling back to old prefix behaviour"* to the warning text, to make it clear what happened.

doc/usage/extensions/extlinks.rst
ghost: Please see [reply](https://github.com/sphinx-doc/sphinx/pull/8898#issuecomment-815526332) below.

sphinx/ext/extlinks.py
tk0miya: You need to pass a tuple of the current docname and lineno to the logging module. And you can get it via `inliner.document.settings.env.docname` call.

```
docname = inliner.document.settings.env.docname
logger.warning(__('...'), location=(docname, lineno))  # __(...) function is a helper for i18n
```

tk0miya:
Thank you for your explanation. Finally, I understand what you're doing. And my last comment was bad. Indeed, the code keeps compatibility.

Now I'm debating about the deprecation step of the configuration. I hesitate the new warning causes CI errors in many projects (because they use `-W` option usually). So it might be better to support the new format in v4.0 and deprecate (emit a warning) the old one in v5. It would reduce the trouble a bit more. What do you think?

sphinx/ext/extlinks.py
tk0miya: It would be better to check the configuration on the generating role function. Emitting warnings on every extlink-role call is a bit noisy.

```
def make_link_role(base_url: str, caption: str) -> RoleFunction:
  # check the caption is a prefix (old format).
  if '%s' not in caption:
    # convert to the new one
    caption += '%s'

    logger.warning(...)  # emit a warning (if needed, please see another comment)

  def role(...):
     ...
```

In this case, no location info is not needed for the warning. It's an error of configuration, not source text's.

ghost:
@tk0miya wrote:

> Now I'm debating about the deprecation step of the configuration. I hesitate the new warning causes CI errors in many projects (because they use -W option usually). So it might be better to support the new format in v4.0 and deprecate (emit a warning) the old one in v5. It would reduce the trouble a bit more. What do you think?

Reading through the projects Deprecation Policy [0] (I assume this feature hits Sphinx 4.0 or 4.1).
According to the policy:

> * Sphinx 4.x will contain a backwards-compatible replica of the function which will raise a `RemovedInSphinx60Warning`. This is a subclass of `PendingDeprecationWarning`, i.e. it will not get displayed by default.
>
> * Sphinx 5.x will still contain the backwards-compatible replica, but `RemovedInSphinx60Warning` will be a subclass of `DeprecationWarning` then, and gets displayed by default.
>
> * Sphinx 6.0 will remove the feature outright.

To me this looks like a clean and consistent way of getting rid of old cruft while giving users enough time to adapt (Well done!). The most affected ones are arguably developers and CI, which already know the drill.

I updated the code accordingly and took inspiration from `sphinx/ext/autodoc/__init__.py` regarding the deprecation warning mechanism. Once we agreed on the implementation, I'll squash the fixup commits and order them appropriatly.

[0] https://www.sphinx-doc.org/en/master/internals/release-process.html

sphinx/ext/extlinks.py
ghost: Thanks!

tk0miya:
Thank you for the updates. LGTM!

tk0miya (approved):

tk0miya:
Note: To let users know the deprecation, I'll use `logger.warning` instead of `RemovedInSphinxXXWarning` since v5.x. It will cause a crash if users use `-W` option on their CI. I believe it's helpful to notify the change to them surely.

tk0miya:
I wrongly merged this into the 3.5.x branch... To fix my mistake up, I reverted and cherry-picked these commits on our repository directly (without force push).

tk0miya:

ghost:
Thanks to *you* for merging and reviewing!

PR #8903, opened 2021-02-17
Add option "scale_latex" for "graphviz" extension

slvr9:
### Feature or Bugfix

- Feature

### Purpose
This change allows you to set the scale of the [sphinx.ext.graphviz](https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html#module-sphinx.ext.graphviz) graphs in the resulting `*.pdf`.

### Detail
N/A

### Relates
N/A

### DIFF
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -356,11 +356,12 @@ def render_dot_latex(self: LaTeXTranslator, node: graphviz, code: str,
                 post = r'\hspace*{\fill}}'
         self.body.append('\n%s' % pre)
 
-    sphinxincludegraphics_opt=''
-    if 'scale_latex' in node:
-        sphinxincludegraphics_opt=sphinxincludegraphics_opt+r' '+r'scale='+node['scale_latex']
+    sphinxincludegraphics_opt = ''
 
-    self.body.append(r'\sphinxincludegraphics['+sphinxincludegraphics_opt+']{%s}' % fname)
+    if 'scale_latex' in node:
+        sphinxincludegraphics_opt = sphinxincludegraphics_opt + r' ' + r'scale=' + node['scale_latex']
+
+    self.body.append(r'\sphinxincludegraphics[' + sphinxincludegraphics_opt + ']{%s}' % fname)
 
     if not is_inline:
         self.body.append('%s\n' % post)
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -356,12 +356,12 @@ def render_dot_latex(self: LaTeXTranslator, node: graphviz, code: str,
                 post = r'\hspace*{\fill}}'
         self.body.append('\n%s' % pre)
 
-    sphinxincludegraphics_opt = ''
+    opt = ''
 
     if 'scale_latex' in node:
-        sphinxincludegraphics_opt = sphinxincludegraphics_opt + r' ' + r'scale=' + node['scale_latex']
+        opt = opt + r' ' + r'scale=' + node['scale_latex']
 
-    self.body.append(r'\sphinxincludegraphics[' + sphinxincludegraphics_opt + ']{%s}' % fname)
+    self.body.append(r'\sphinxincludegraphics[' + opt + ']{%s}' % fname)
 
     if not is_inline:
         self.body.append('%s\n' % post)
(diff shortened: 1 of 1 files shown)


sphinx/ext/graphviz.py
```diff
@@ -121,6 +121,7 @@ class Graphviz(SphinxDirective):
         'graphviz_dot': directives.unchanged,  # an old alias of `layout` option
         'name': directives.unchanged,
         'class': directives.class_option,
+        'scale_latex': directives.unchanged,
```
tk0miya, line 124: This means we'll add `scale_{format}` options (ex. `scale_html`, `scale_epub`) in the future? I feel it's inconvenient. Why do you limit the option only for LaTeX?

slvr9: 1. I think these settings should be done separately.
By generating HTML, I get an acceptable image. In a PDF file, the diagram looks excessively large.
It seems to me that this is due to the different font sizes specified in "*.DOT", "html_theme" and "latex_theme".
Therefore, customizing the scale is required individually.

2. I was able to find an easy way to scale for LaTeX only.  :sweat:

slvr9: > This means we'll add scale_{format} options (ex. scale_html, scale_epub) in the future? I feel it's inconvenient. 

It is probably better to do a generalized `:scale: auto_by_font_size` options, but it is too complicated for me.

jfbu: I tried as a hack something such as 
```diff
diff --git a/sphinx/ext/graphviz.py b/sphinx/ext/graphviz.py
index f10285086..b4ef39cc5 100644
--- a/sphinx/ext/graphviz.py
+++ b/sphinx/ext/graphviz.py
@@ -249,6 +249,9 @@ def render_dot(self: SphinxTranslator, code: str, options: Dict, format: str,
     if format == 'png':
         dot_args.extend(['-Tcmapx', '-o%s.map' % outfn])
 
+    if format == 'pdf':
+        dot_args.extend(['-Nfontsize=10'])
+
     try:
         ret = subprocess.run(dot_args, input=code.encode(), stdout=PIPE, stderr=PIPE,
                              cwd=cwd, check=True)
```
and it works to reduce the used fontsize. One should also use ``-E`` and perhaps ``-G`` but I am not sure.

slvr9: Perhaps this is the solution. 
However, this will only work if the "DOT" file does not specify the font size. 
If I have file a "1.dot" with an explicit `fontsize=10` and another "2.dot" with `fontsize=20`, then your option probably won't work.

tk0miya: Sphinx is designed one source / multiple output. It means the output formats are not only HTML and LaTeX/PDF. We have to care about other formats; EPUB, texinfo, XML, and other 3rd party builders. I think `scale_{format}` option is not extensible. So -1 for adding the options. We need to find the better way to realize this feature.

slvr9: I understand your point of view.

> We need to find the better way to realize this feature.

Like a stupid idea: GLOBAL for sphinx-doc.
1. Add an option `opt` for ALL directives with detailed settings for generating. 
    The `builder` will select the necessary settings based on the records of the form 
    `:opt:{builder?format}: optName01:optValue01; .. optName0N:optValue0N`.

    ``` *.rst
    .. graphviz:: fileName.dot
       :caption: diagram Name
       :opt:latex: scale:0.6  
       :opt:html: scale:0.9; size:strong  
       :opt:epub: ...

    .. toctree::
       :maxdepth: 2
       :opt:latex: maxdepth:10 
    ```

2. or Only for "options"
    ``` *.rst
    .. graphviz:: fileName.dot
       :caption: diagram Name
       :scale: 0.6  
       :scale:html: 0.9

    .. toctree::
       :maxdepth: 2
       :maxdepth:latex: 10 
    ```

If I understand correctly, the directives [.. only::](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=directive%20only#directive-only) and [.. ifconfig::](https://www.sphinx-doc.org/en/master/usage/extensions/ifconfig.html?highlight=ifconfig) have similar behavior.

tk0miya: Thank you for considering. I prefer the 2nd idea. +1 if implemented.

>If I understand correctly, the directives .. only:: and .. ifconfig:: have similar behavior.

Indeed, they are similar, but different approaches. They control the conditional branches like a program. But the 2nd way is not programmable. Users can choose only the name of the builders.

Issue #8881, opened 2021-02-14
The depth of bookmarks panel in PDF is not enough for navigation (even in absence of ``:maxdepth:`` option with ``toctree`` directive)


jfbu:
**Describe the bug**

Consider for example the Sphinx own docs, which uses a toctree with maxdepth set to 2. In PDF output, if I revert the 
commit 3a3a479f1e739d688 (#8746, merged into 3.5.0), the following shows:
![Capture d’écran 2021-02-14 à 11 08 17](https://user-images.githubusercontent.com/2589111/107873847-d1a01f80-6eb5-11eb-97d4-9af19dcb5fa2.png)
i.e.
- from the opened bookmarks panel one does not see and can not navigate to all extensions built-in into Sphinx,
- from the PDF contents one can not either, and there are about 50 pages with no way to navigate except do a linear continuous reading, whether on screen or on print.

Compare to the html which looks like this

![Capture d’écran 2021-02-14 à 11 09 13](https://user-images.githubusercontent.com/2589111/107873911-2d6aa880-6eb6-11eb-82fb-136e95ee552f.png)

once one has gone to the Extensions page.

The underlying problem here is that the latexpdf route builds a single document.

Another sub-issue is that our docs says in the documentation of ``toctree`` directive:

> A numeric maxdepth option may be given to indicate the depth of the tree; by default, all levels are included.

However, if one does not use the ``maxdepth`` option, then, with ``'manual'`` type only 2 levels are kept, not the maximum number of levels.

File ``file.rst``

```rest
=====
 aaa
=====

bbb
=====

ccc
****

ddd
~~~~

eee
^^^^

fff
----

ggg
```

file ``index.rst``:

```
Welcome to FOO's documentation!
===============================

.. toctree::
   :caption: Contents:

   file
```

Only two levels in TOC and bookmarks in output.

![Capture d’écran 2021-02-14 à 11 47 28](https://user-images.githubusercontent.com/2589111/107874513-71f84300-6eba-11eb-8e38-4ea24dac8282.png)

But the html output uses full depth:

![Capture d’écran 2021-02-14 à 11 46 28](https://user-images.githubusercontent.com/2589111/107874501-61e06380-6eba-11eb-8ee7-a383c8df24d4.png)

**Options to consider**

The options to fix this:

1. use some latex package to add automatically local table of contents in each sectioning level at maxdepth,
2. set the document TOC depth to a high value always
[...]

PR #8907, opened 2021-02-18
LaTeX: let bookmarks use high default depth

jfbu:
Add bookmarksdepth key to ``latex_elements['sphinxsetup']`` (this is less work than adding  full-fledged configuration variable...)

In my experience, PDF viewers always show bookmarks as collapsed by default (if not passed special configuration from inside pdf document).

This is only change to the bookmarks depth so changes nothing to document printed on paper.

I hesitated about making PR on 3.x branch but I found no "3.6.0" entry in CHANGES only "3.5.2". So on master is fine.

### Relates
- #8881


doc/changes.rst
```diff
@@ -10,7 +10,6 @@ Changelog
 
 .. raw:: latex
 
-   \hypersetup{bookmarksdepth=1}% pdf bookmarks
    \addtocontents{toc}{\protect\setcounter{tocdepth}{1}}%
```
jfbu, line 13: I drop the hypersetup, finally in bookmarks it is ok to access finest level for ChangeLog

doc/conf.py
```diff
@@ -61,11 +61,10 @@
     'fontenc': r'\usepackage[LGR,X2,T1]{fontenc}',
     'passoptionstopackages': r'''
 \PassOptionsToPackage{svgnames}{xcolor}
-\PassOptionsToPackage{bookmarksdepth=3}{hyperref}% depth of pdf bookmarks
 ''',
     'preamble': r'''
 \DeclareUnicodeCharacter{229E}{\ensuremath{\boxplus}}
-\setcounter{tocdepth}{3}%    depth of what is kept from toc file
+\setcounter{tocdepth}{3}%    depth of what main TOC shows (3=subsubsection)
```
jfbu, line 67: not related to this PR, but I took opportunity to improve comment

sphinx/texinputs/sphinx.sty
```diff
@@ -65,6 +65,8 @@
 \DeclareStringOption[-1]{numfigreset}
 \DeclareBoolOption[false]{nonumfigreset}
 \DeclareBoolOption[false]{mathnumfig}
+\define@key{sphinx}{bookmarksdepth}{\AtBeginDocument{\hypersetup{bookmarksdepth=#1}}}
+\AtBeginDocument{\define@key{sphinx}{bookmarksdepth}{\hypersetup{bookmarksdepth=#1}}}
```
jfbu, line 69: we can not use ``\hypersetup`` before hyperref is loaded and we can not use ``\PassOptionsToPackage`` after hyperref is loaded. The initial definition could do some `\@ifpackageloaded{hyperref}` but it looked complicated for nothing. Manual `\hypersetup{bookmarksdepth=..}` in preamble will thus be overridden by a usage of bookmarksdepth sphinxsetup key, but user should use the sphinxsetup key.

jfbu: > we can not use `\hypersetup` before hyperref is loaded and we can not use `\PassOptionsToPackage` after hyperref is loaded. 

(background here is that the key allows usage of `\sphinxsetup`  macro, so my comment refers to using directly `\sphinxsetup` macro in preamble, rather than indirectly via conf.py's ``latex_elements['sphinxsetup']``)

tk0miya (approved):

CHANGES
```diff
@@ -75,6 +75,7 @@ Bugs fixed
   change) with late TeXLive 2019
 * #8253: LaTeX: Figures with no size defined get overscaled (compared to images
   with size explicitly set in pixels) (fixed for ``'pdflatex'/'lualatex'`` only)
+* #8881: LaTeX: The depth of bookmarks panel in PDF is not enough for navigation
```
tk0miya, line 98: I think this is an improvement. But either is okay.

jfbu:
Thanks for review, now merging

PR #8923, opened 2021-02-23
Fix #8917: autodoc: Raises a warning if function has wrong __globals__ value

### LINKED ISSUE #8917: Sphinx throws errors when using function in __getattr__ override
**Describe the bug**
When using anonymous functions in getattr sphinx will raise an error:

Handler <function record_typehints at 0x7f3626d18790> for event 'autodoc-process-signature' threw an exception (exception: 'function' object has no attribute 'get')

**To Reproduce**
```
def fun():
    pass

class Tmp:

    def __call__(self, name, content):
        print(name, content)

    def __getattr__(self, item):
        return fun

a = Tmp()
"""asdfasdf"""
```

```
cd docs && \
make clean && \
sphinx-apidoc --force -e -o ./source ../<package> && \
make html
```
**Expected behavior**
No error.

**Environment info**
- Python version: [3.8.5]

Sphinx==3.4.3
sphinx-rtd-theme==0.5.1
sphinxcontrib-applehelp==1.0.2
sphinxcontrib-devhelp==1.0.2
sphinxcontrib-htmlhelp==1.0.3
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==1.0.3
sphinxcontrib-serializinghtml==1.1.4

extensions = ['sphinx.ext.todo', 'sphinx.ext.viewcode', 'sphinx.ext.autodoc', 'sphinx.ext.napoleon']

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- `sphinx.util.inspect:signature()` crashes with AttributeError when
subject has wrong `__globals__` value.  This ignores the error on
building.
- refs: #8917

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -69,6 +69,7 @@ Features added
 Bugs fixed
 ----------
 
+* #8917: autodoc: Raises a warning if function has wrong __globals__ value
 * #8380: html search: Paragraphs in search results are not identified as ``<p>``
 * #8915: html theme: The translation of sphinx_rtd_theme does not work
 * #8342: Emit a warning if a unknown domain is given for directive or role (ex.
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -166,6 +166,15 @@ def getannotations(obj: Any) -> Mapping[str, Any]:
         return {}
 
 
+def getglobals(obj: Any) -> Mapping[str, Any]:
+    """Get __globals__ from given *obj* safely."""
+    __globals__ = safe_getattr(obj, '__globals__', None)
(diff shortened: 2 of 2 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -498,7 +498,7 @@ def format_signature(self, **kwargs: Any) -> str:
                     if matched:
                         args = matched.group(1)
                         retann = matched.group(2)
-            except Exception as exc:
+            except RuntimeError as exc:
```
tk0miya: Oops... this change is not intended. this should be reverted.

PR #8929, opened 2021-02-24
Intersphinx delegation to domains

jakobandersen:
### Feature or Bugfix
- Feature
- Bugfix
- Refactoring

### Purpose
Currently, Intersphinx is performing lookups in inventories purely by string comparison. Unfortunately this does not work in general. For C and especially C++ more complex processing needs to be done.

As an overall goal for Intersphinx: provide cross-referencing to external projects _as if_ those projects were merely a part of the current project. In order to do this I believe Intersphinx must delegate most processing of inventories and lookup to the domains. This means both a new inventory format is needed and changes to lookup. This PR tries to address the latter. See the bottom for notes on the inventory format.

### Detail
- When loading inventories, group entries by domain and object type, and package data only related to Intersphinx in a new class ``sphinx.util.inventory.InventoryItemSet``. This class is private for Intersphinx.
- Intersphinx maintains domain storage of arbitrary type, initialized by a clone of a new domain variable ``initial_intersphinx_inventory``. This mirrors the ordinary domain storage through ``initial_data``.
- Data is then given to the domains through a new method ``intersphinx_add_entries_v2()``, where the domain stores in the given domain storage. The ``v2`` in the name refers to the current version of ``objects.inv`` (the v1 is converted to v2).
- When Intersphinx gets the ``missing-reference`` event it will use two new domain methods ``intersphinx_resolve_xref``. The previously filled domain storage is given here as well.
- The domain must perform lookup and will end up with a (possibly empty) ``sphinx.util.inventory.InventoryItemSet`` which must be returned (the domains doesn't know that it is this class, just some arbitrary data that it got from Intersphinx earlier). Intersphinx will then perform disambiguation between duplicates across the inventories, and create the resulting nodes.

The old lookup code in ``intersphinx.py`` is now the default implementation in the ``Domain`` base class for backwards compatibility. The domain-specific things for ``std`` and ``py`` have been spliced out (though in a slightly haxy manner).

As a kind of proof-of-concept I have implemented custom handling in the C domain, where it reuses the ordinary symbol tables classes and lookup procedures. A test for lookup in relation to anonymous types illustrates that more lookups will work with this implementation.

### Relates
- #7819 will completely break intersphinx lookups without this PR.

### Next steps
We should make a new inventory format where domains can store symbols in whichever format they like: https://github.com/orgs/sphinx-doc/discussions/12204

### DIFF
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -469,9 +469,6 @@ class Domain:
         """
         # for std and py to overwrite for their backwards compatibility
         # we adjust the object types for backwards compatibility
-        if self.name == 'std' and 'cmdoption' in objtypes:
-            # until Sphinx-1.6, cmdoptions are stored as std:option
-            objtypes.append('option')
         if self.name == 'py' and 'attribute' in objtypes:
             # Since Sphinx-2.1, properties are stored as py:method
             objtypes.append('method')
@@ -487,29 +484,9 @@ class Domain:
         """
         for objtype in objtypes:
             if objtype not in store:
-                # Continue if there's nothing of this kind in the inventory
                 continue
-
             if target in store[objtype]:
-                # Case-sensitive match, use it
                 return store[objtype][target]
-            elif self.name == 'std' and objtype in ('label', 'term'):
-                # Some types require case-insensitive matches:
-                # * 'term': https://github.com/sphinx-doc/sphinx/issues/9291
-                # * 'label': https://github.com/sphinx-doc/sphinx/issues/12008
-                target_lower = target.lower()
-                insensitive_matches = list(filter(lambda k: k.lower() == target_lower,
-                                                  store[objtype].keys()))
-                if insensitive_matches:
-                    return store[objtype][insensitive_matches[0]]
-                else:
-                    # No case-insensitive match either, continue to the next candidate
-                    continue
-            else:
-                # Could reach here if we're not a term but have a case insensitive match.
-                # This is a fix for terms specifically, but potentially should apply to
-                # other types.
-                continue
         return None
 
     def intersphinx_resolve_xref(self, env: BuildEnvironment,
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -467,11 +467,7 @@ class Domain:
         After this method returns, the types in *disabled_object_types* are removed
         from *objtypes*. This final list is given to the lookup method.
         """
-        # for std and py to overwrite for their backwards compatibility
-        # we adjust the object types for backwards compatibility
-        if self.name == 'py' and 'attribute' in objtypes:
-            # Since Sphinx-2.1, properties are stored as py:method
-            objtypes.append('method')
+        pass
 
     def _intersphinx_resolve_xref_lookup(self, store: dict[str, dict[str, Any]],
                                          target: str, objtypes: list[str]
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -479,6 +479,12 @@ class Domain:
     def _intersphinx_resolve_xref_lookup(self, store: dict[str, dict[str, Any]],
                                          target: str, objtypes: list[str]
                                          ) -> Any | None:
+        """Default implementation for the actual lookup.
+
+        This method is an internal implementation detail, and may be overridden
+        by the bundled domains, e.g., std, for customizing the lookup, while
+        reusing the rest of the default implementation.
+        """
         for objtype in objtypes:
             if objtype not in store:
                 # Continue if there's nothing of this kind in the inventory
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -489,7 +489,7 @@ class Domain:
                 continue
         return None
 
-    def intersphinx_resolve_xref(self, env: "BuildEnvironment",
+    def intersphinx_resolve_xref(self, env: BuildEnvironment,
                                  store: Any,
                                  typ: str, target: str,
                                  disabled_object_types: list[str],
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -427,7 +427,16 @@ class Domain:
         that should not be inspected by the domain, but merely returned as a result
         of reference resolution.
 
-        The domain must store the given data in the given *store*,
+        For example, for Python the data could look like the following.
+
+        .. code-block:: python
+
+            data = {
+                'class': {'pkg.mod.Foo': SomeIntersphinxData(...)},
+                'method': {'pkg.mod.Foo.bar': SomeIntersphinxData(...)},
+            }
+
+        The domain must store the given inner intersphinx data in the given *store*,
         in whichever way makes sense for later reference resolution.
         This *store* was initially a copy of :attr:`initial_intersphinx_inventory`.
 
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -449,7 +449,24 @@ class Domain:
         assert len(store) == 0  # the method is called at most once
         store.update(data)  # update so the object is changed in-place
 
-    def _intersphinx_adjust_object_types(self, objtypes: list[str]) -> None:
+    def _intersphinx_adjust_object_types(self, env: BuildEnvironment,
+                                         store: Any,
+                                         typ: str, target: str,
+                                         disabled_object_types: list[str],
+                                         node: pending_xref, contnode: Element,
+                                         objtypes: list[str]) -> None:
+        """For implementing backwards compatibility.
+
+        This method is an internal implementation detail used in the std and python domains,
+        for implementing backwards compatibility.
+
+        The given *objtypes* is the list of object types computed based on the *typ*,
+        which will be used for lookup.
+        By overriding this method this list can be manipulated, e.g., adding types
+        that were removed in earlier Sphinx versions.
+        After this method returns, the types in *disabled_object_types* are removed
+        from *objtypes*. This final list is given to the lookup method.
+        """
         # for std and py to overwrite for their backwards compatibility
         # we adjust the object types for backwards compatibility
         if self.name == 'std' and 'cmdoption' in objtypes:
@@ -525,7 +542,8 @@ class Domain:
                 return None
             objtypes = for_role
 
-        self._intersphinx_adjust_object_types(objtypes)
+        self._intersphinx_adjust_object_types(
+            env, store, typ, target, disabled_object_types, node, contnode, objtypes)
         objtypes = [o for o in objtypes if o not in disabled_object_types]
 
         typed_store = cast(dict[str, dict[str, Any]], store)
--- a/sphinx/domains/__init__.py
+++ b/sphinx/domains/__init__.py
@@ -201,6 +201,11 @@ class Domain:
     data: dict
     #: data version, bump this when the format of `self.data` changes
     data_version = 0
+    #: Value for an empty inventory of objects for this domain.
+    #: It must be copy.deepcopy-able.
+    #: If this value is overridden, then the various intersphinx methods in the domain should
+    #: probably also be overridden.
+    initial_intersphinx_inventory: dict = {}
 
     def __init__(self, env: BuildEnvironment) -> None:
         self.env: BuildEnvironment = env
@@ -404,3 +409,122 @@ class Domain:
     def get_full_qualified_name(self, node: Element) -> str | None:
         """Return full qualified name for given node."""
         pass
+
+    def intersphinx_add_entries_v2(self, store: Any,
+                                   data: dict[str, dict[str, Any]]) -> None:
+        """Store the given *data* for later intersphinx reference resolution.
+
+        This method is called at most once with all data loaded from inventories in
+        v1 and v2 format.
+
+        The *data* is a dictionary indexed by **object type**, i.e, a key from
+        :attr:`object_types`. The value is a dictionary indexed by **object name**,
+        i.e., the **name** part of each tuple returned by :meth:`get_objects`.
+        The value of those inner dictionaries are objects with intersphinx data,
+        that should not be inspected by the domain, but merely returned as a result
+        of reference resolution.
+
+        The domain must store the given data in the given *store*,
+        in whichever way makes sense for later reference resolution.
+        This *store* was initially a copy of :attr:`initial_intersphinx_inventory`.
+
+        Later in :meth:`intersphinx_resolve_xref` during intersphinx reference resolution,
+        the domain is again given the *store* object to perform the resolution.
+
+        .. versionadded:: 8.0
+        """
+        store = cast(dict[str, dict[str, Any]], store)
+        assert len(store) == 0  # the method is called at most once
+        store.update(data)  # update so the object is changed in-place
+
+    def _intersphinx_adjust_object_types(self, objtypes: list[str]) -> None:
+        # for std and py to overwrite for their backwards compatibility
+        # we adjust the object types for backwards compatibility
+        if self.name == 'std' and 'cmdoption' in objtypes:
+            # until Sphinx-1.6, cmdoptions are stored as std:option
+            objtypes.append('option')
+        if self.name == 'py' and 'attribute' in objtypes:
+            # Since Sphinx-2.1, properties are stored as py:method
+            objtypes.append('method')
+
+    def _intersphinx_resolve_xref_lookup(self, store: dict[str, dict[str, Any]],
+                                         target: str, objtypes: list[str]
+                                         ) -> Any | None:
+        for objtype in objtypes:
+            if objtype not in store:
+                # Continue if there's nothing of this kind in the inventory
+                continue
+
+            if target in store[objtype]:
+                # Case-sensitive match, use it
+                return store[objtype][target]
+            elif self.name == 'std' and objtype in ('label', 'term'):
+                # Some types require case-insensitive matches:
+                # * 'term': https://github.com/sphinx-doc/sphinx/issues/9291
+                # * 'label': https://github.com/sphinx-doc/sphinx/issues/12008
+                target_lower = target.lower()
+                insensitive_matches = list(filter(lambda k: k.lower() == target_lower,
+                                                  store[objtype].keys()))
+                if insensitive_matches:
+                    return store[objtype][insensitive_matches[0]]
+                else:
+                    # No case-insensitive match either, continue to the next candidate
+                    continue
+            else:
+                # Could reach here if we're not a term but have a case insensitive match.
+                # This is a fix for terms specifically, but potentially should apply to
+                # other types.
+                continue
+        return None
+
+    def intersphinx_resolve_xref(self, env: "BuildEnvironment",
+                                 store: Any,
+                                 typ: str, target: str,
+                                 disabled_object_types: list[str],
+                                 node: pending_xref, contnode: Element
+                                 ) -> Any | None:
+        """Resolve the pending_xref *node* with the given *target* via intersphinx.
+
+        This method should perform lookup of the pending cross-reference
+        in the given *store*, but otherwise behave very similarly to :meth:`resolve_xref`.
+
+        The *typ* may be ``any`` if the cross-references comes from an any-role.
+
+        The *store* was created through a previous call to :meth:`intersphinx_add_entries_v2`.
+
+        The *disabled_object_types* is a list of object types that the reference may not
+        resolve to, per user request through :confval:`intersphinx_disabled_reftypes`.
+        These disabled object types are just the names of the types, without a domain prefix.
+
+        If a candidate is found in the store, the associated object must be returned.
+
+        If no candidates can be found, *None* can be returned; and subsequent event
+        handlers will be given a chance to resolve the reference.
+        The method can also raise :exc:`sphinx.environment.NoUri` to suppress
+        any subsequent resolution of this reference.
+
+        .. versionadded:: 8.0
+        """
+        if typ == 'any':
+            objtypes = list(self.object_types)
+        else:
+            for_role = self.objtypes_for_role(typ)
+            if not for_role:
+                return None
+            objtypes = for_role
+
+        self._intersphinx_adjust_object_types(objtypes)
+        objtypes = [o for o in objtypes if o not in disabled_object_types]
+
+        typed_store = cast(dict[str, dict[str, Any]], store)
+        # we try the target either as is, or with full qualification based on the scope of node
+        res = self._intersphinx_resolve_xref_lookup(typed_store, target, objtypes)
+        if res is not None:
+            return res
+        # try with qualification of the current scope instead
+        full_qualified_name = self.get_full_qualified_name(node)
+        if full_qualified_name:
+            return self._intersphinx_resolve_xref_lookup(
+                typed_store, full_qualified_name, objtypes)
+        else:
+            return None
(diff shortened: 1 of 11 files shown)


jakobandersen:
Rebased to resolve conflicts. @tk0miya, do you by chance have time to review this before we get too close to the v4 release?

tk0miya:
Oh, sorry. I overlooked your PR at all.

I completely agree with you. It's difficult to resolve cross-references only on the intersphinx module and common inventory data structure. So the resolution process should be passed to the domains.

In my thought, it is better to export the whole of the env object (or the whole of the domain object) as an inventory database and import it as a real env (or domain) object on intersphinx. I think it's similar to this PR, but no new methods are needed. What do you think?
```
# image code
inventory_v3 = fetch_remote_inventory(url)
env = BuildEnvironment.from_inventory(inventory_v3)
cpp_domain = env.get_domain('cpp')
cpp_domain.resolve_xref(pending_xref)
```

jakobandersen:
That would be delightfully simply, though I think we would still need domain-specific methods for the loading. The inventory may be generated by an old version, so ``BuildEnvironment.from_inventory()`` would need to delegate to the domains so they can do the migration of the data into the current environment. And we would of course need to support the current inventory formats as well. 

As for lookup: the current intersphinx lookup does some extra tricks, like splitting of an inventory name and perform targeted lookup, or when multiple inventories have the same entry, then disambiguate it in a deterministic way.
I know the C and C++ domains, and I assume the rest, does not directly support this kind of lookup. So in this way each inventory must be kept separate and intersphinx must in turn perform lookup. In a sense this is an extension of what is already going on: try projects in some predefined order, with the current project being the first.
Though, I'm a bit unsure if this is a good idea, as best candidate lookup will not be possible. In C++ a target ``T<int>`` may be resolved to a class template specialization ``template<> class T<int>``, but if no such specialization has been defined, it will resolve to the primary template ``template<typename T> class T`` (if it exists). If the inventories are not merged, with built-in disambiguation, this will not work in general as the first inventory might have the primary template, while a latter the specialization. However, this is probably a quite niche use case.

jakobandersen:
The PR is now updated on top of 5.x where the intersphinx role has been added. I believe it is ready for another round of review.

AA-Turner:
@jakobandersen The implementation looks good, I have a few comments:

- I think it makes more sense to use a dict as the backing structure in `InventoryItemSet`
- I don't see the point of the `_v2` suffixes -- they are confusing and when we move to a better format for `objects.inv`, we would disambiguate by string-based versus structured, I imagine.
- I think a few of the new functions (e.g. `process_disabled_reftypes`) and properties could do with a docstring to explain what they're for
- Some style / typing points (e.g. using `snake_case`) and improving static typing
- In general I think we should strongly avoid `assert ` statements, we should either raise an error or remove the assert.

I've implemented most of my suggested changes in https://github.com/AA-Turner/sphinx/commits/intersphinx_delegation

I would like to merge this sooner rather than later if possible -- please let me know if you're happy with my patches on your branch. Adding docstrings/etc can wait until after this is merged if need be.

A

AA-Turner:
I want to release 5.2.0, so I'm moving this to 5.3.0.

@jakobandersen please let me know if you've time to review, otherwise I can move forward to allow progress on the other changes.

A

jakobandersen:
@AA-Turner, please go ahead with a release without this PR. I won't have time to review the next couple of weeks (either).

AA-Turner:
I've resolved the merge conflicts here.

A

AA-Turner:
I've also updated my branch with the suggested changes above (https://github.com/AA-Turner/sphinx/commits/intersphinx_delegation)

A

jakobandersen:
The PR has now been resurrected and should be ready for review and merge.
 As we are now discussing a new intersphinx format I have reintroduced the ``v2`` suffixes, as there the new format will require a ``v3`` version of those entities.

picnixz:
Since I'm busy until Sunday, I will not review that before next week. It's also better for me to have a fresh head and fresh view on that. I believe @chrisjsewell would also want to have a look at it.

sphinx/domains/__init__.py
```diff
@@ -201,6 +201,11 @@ class Domain:
     data: dict
     #: data version, bump this when the format of `self.data` changes
     data_version = 0
+    #: Value for an empty inventory of objects for this domain.
+    #: It must be copy.deepcopy-able.
+    #: If this value is overridden, then the various intersphinx methods in the domain should
+    #: probably also be overridden.
+    initial_intersphinx_inventory: dict = {}
```
picnixz: Could we get a better type annotation?

sphinx/domains/__init__.py
```diff
@@ -404,3 +409,95 @@ def get_enumerable_node_type(self, node: Node) -> str | None:
...
+    def intersphinx_add_entries_v2(self, store: Any,
+                                   data: dict[str, dict[str, Any]]) -> None:
+        """Store the given *data* for later intersphinx reference resolution.
+
+        This method is called at most once with all data loaded from inventories in
+        v1 and v2 format.
+
+        The *data* is a dictionary indexed by **object type**, i.e, a key from
+        :attr:`object_types`. The value is a dictionary indexed by **object name**,
+        i.e., the **name** part of each tuple returned by :meth:`get_objects`.
```
picnixz, line 425: I'd suggest adding an example (the structure would be clearer IMO), e.g.: 

```python
data = {
	'class': {'pkg.mod.Foo': SomeIntersphinxData(...)},
	'method': {'pkg.mod.Foo.bar': SomeIntersphinxData(...)},
}
```

sphinx/domains/__init__.py
```diff
@@ -404,3 +409,95 @@ def get_enumerable_node_type(self, node: Node) -> str | None:
...
+        Later in :meth:`intersphinx_resolve_xref` during intersphinx reference resolution,
+        the domain is again given the *store* object to perform the resolution.
+
+        .. versionadded:: 8.0
+        """
+        store = cast(dict[str, dict[str, Any]], store)
+        assert len(store) == 0  # the method is called at most once
+        store.update(data)  # update so the object is changed in-place
+
+    def _intersphinx_adjust_object_types(self, objtypes: list[str]) -> None:
```
picnixz: Could you add some comment (mostly for us?)

sphinx/domains/__init__.py
```diff
@@ -404,3 +409,95 @@ def get_enumerable_node_type(self, node: Node) -> str | None:
...
+                                         target: str, objtypes: list[str]
+                                         ) -> Any | None:
+        for objtype in objtypes:
+            if objtype not in store:
+                continue
+            if target in store[objtype]:
+                return store[objtype][target]
+        return None
+
+    def intersphinx_resolve_xref(self, env: "BuildEnvironment",
```
picnixz:
```suggestion
    def intersphinx_resolve_xref(self, env: BuildEnvironment,
```

sphinx/domains/__init__.py
```diff
@@ -404,3 +409,95 @@ def get_enumerable_node_type(self, node: Node) -> str | None:
...
+        """
+        if typ == 'any':
+            objtypes = list(self.object_types)
+        else:
+            for_role = self.objtypes_for_role(typ)
+            if not for_role:
+                return None
+            objtypes = for_role
+
+        self._intersphinx_adjust_object_types(objtypes)
```
picnixz: Now I see how it is used. Would it make sense to also pass the 'typ' ? because it could be helpful for domains to know which role they were trying to resolve. In addition, passing the target would make them suitable for providing overrides in some situations where a target cannot be correctly resolved.

Or... you can just pass *everything* that was given to this method?

sphinx/domains/python/__init__.py
```diff
@@ -824,6 +824,12 @@ def _make_module_refnode(self, builder: Builder, fromdocname: str, name: str,
         return make_refnode(builder, fromdocname, module.docname, module.node_id,
                             contnode, title)
 
+    def _intersphinx_adjust_object_types(self, objtypes: list[str]) -> None:
+        # we adjust the object types for backwards compatibility
+        if 'attribute' in objtypes:
```
picnixz:
```suggestion
        if 'attribute' in objtypes and 'method' not in objtypes:
```

I wouldn't want to renact #11337 :smile:

sphinx/domains/std/__init__.py
```diff
@@ -996,6 +996,35 @@ def resolve_any_xref(self, env: BuildEnvironment, fromdocname: str,
...
+                continue
+
+            if target in store[objtype]:
+                # Case-sensitive match, use it
+                return store[objtype][target]
+            elif objtype in ('label', 'term'):
+                # Some types require case-insensitive matches:
+                # * 'term': https://github.com/sphinx-doc/sphinx/issues/9291
+                # * 'label': https://github.com/sphinx-doc/sphinx/issues/12008
+                target_lower = target.lower()
```
picnixz, line 1025: Just wondering, but target names are only ASCII right?

sphinx/domains/std/__init__.py
```diff
@@ -996,6 +996,35 @@ def resolve_any_xref(self, env: BuildEnvironment, fromdocname: str,
...
+    def _intersphinx_adjust_object_types(self, objtypes: list[str]) -> None:
+        # we adjust the object types for backwards compatibility
+        if 'cmdoption' in objtypes:
+            # until Sphinx-1.6, cmdoptions are stored as std:option
+            objtypes.append('option')
+
+    def _intersphinx_resolve_xref_lookup(self, store: dict[str, dict[str, Any]],
+                                         target: str, objtypes: list[str]
+                                         ) -> Any | None:
+        # Semi-haxy overwriting of the private method in Domain, as we only need to do
```
picnixz: You could say that the method is actually protected :)

sphinx/domains/std/__init__.py
```diff
@@ -996,6 +996,35 @@ def resolve_any_xref(self, env: BuildEnvironment, fromdocname: str,
...
+                return store[objtype][target]
+            elif objtype in ('label', 'term'):
+                # Some types require case-insensitive matches:
+                # * 'term': https://github.com/sphinx-doc/sphinx/issues/9291
+                # * 'label': https://github.com/sphinx-doc/sphinx/issues/12008
+                target_lower = target.lower()
+                insensitive_matches = list(filter(lambda k: k.lower() == target_lower,
+                                                  store[objtype].keys()))
+                if insensitive_matches:
+                    return store[objtype][insensitive_matches[0]]
```
picnixz:
```suggestion
				for object_name, object_data in store[objtype].items():
					if object_name.lower() == target_lower:
						return object_data
```

sphinx/domains/c/_ast.py
picnixz, line 1: I know that you are the one maintaining the C/C++ domains but... is there a reference for all those private implementations? While I can understand most of it, I don't remember everything from one review to another so I'm wondering whether you would be willing to add some internal doc for the C/C++ domain.

sphinx/domains/c/__init__.py
```diff
@@ -723,20 +729,32 @@ def _resolve_xref_inner(self, env: BuildEnvironment, fromdocname: str, builder:
...
+                    msg = f"Target: {target}\nParentKey: {parentKey}\n{root.dump(1)}\n"
+                    raise AssertionError(msg)
         else:
-            parentSymbol = rootSymbol
+            parentSymbol = root
         s = parentSymbol.find_declaration(name, typ,
                                           matchSelf=True, recurseInAnon=True)
         if s is None or s.declaration is None:
             return None, None
+            # TODO: conditionally warn about xrefs with incorrect tagging?
```
picnixz, line 747: Maybe you could add a debug print instead of a logger's warning? it would be less intruisive at least (emitted from the parentSymbol).

sphinx/domains/c/__init__.py
```diff
@@ -779,6 +797,51 @@ def get_objects(self) -> Iterator[tuple[str, str, str, str, str, int]]:
             newestId = symbol.declaration.get_newest_id()
             yield (name, dispname, objectType, docname, newestId, 1)
 
+    def intersphinx_add_entries_v2(self, store: dict,
```
picnixz:
```suggestion
    def intersphinx_add_entries_v2(self, store: dict[str, Any],
```

I now understand why you had an Any for `store` because of Liskov's principle.

sphinx/domains/c/__init__.py
```diff
@@ -779,6 +797,51 @@ def get_objects(self) -> Iterator[tuple[str, str, str, str, str, int]]:
...
+                                   data: dict[str, dict[str, Any]]) -> None:
+        root = store['root_symbol']  # type: Symbol
+        for object_type, per_type_data in data.items():
+            for object_name, item_set in per_type_data.items():
+                parser = DefinitionParser(
+                    object_name, location=('intersphinx', 0), config=self.env.config)
+                try:
+                    ast = parser._parse_nested_name()
+                except DefinitionError as e:
+                    logger.warning("Error in C entry in intersphinx inventory:\n" + str(e))
```
picnixz:
```suggestion
                    logger.warning("Error in C entry in intersphinx inventory:\n%s", e)
```

sphinx/domains/c/__init__.py
```diff
@@ -779,6 +797,51 @@ def get_objects(self) -> Iterator[tuple[str, str, str, str, str, int]]:
...
+                    continue
+                decl = ASTDeclaration(object_type, 'intersphinx',
+                                      ASTIntersphinx_v2(ast, item_set))
+                root.add_declaration(decl, docname="$FakeIntersphinxDoc", line=0)
```
picnixz:
```suggestion
                root.add_declaration(decl, docname="_$FakeIntersphinxDoc", line=0)
```

sphinx/domains/c/__init__.py
```diff
@@ -779,6 +797,51 @@ def get_objects(self) -> Iterator[tuple[str, str, str, str, str, int]]:
...
+
+    def _intersphinx_resolve_xref_inner(self, env: "BuildEnvironment", store: dict,
```
picnixz:
```suggestion
    def _intersphinx_resolve_xref_inner(self, env: "BuildEnvironment", store: dict[str, Any],
```

sphinx/domains/c/__init__.py
```diff
@@ -779,6 +797,51 @@ def get_objects(self) -> Iterator[tuple[str, str, str, str, str, int]]:
...
+        s, name = self._resolve_xref_in_tree(env, store['root_symbol'],
+                                             True, typ, target, node)
+        if s is None:
+            return None
+        assert s.declaration is not None
+        decl = cast(ASTIntersphinx_v2, s.declaration.declaration)
+        return decl.data
+
+    def intersphinx_resolve_xref(self, env: "BuildEnvironment",
+                                 store: Any,
```
picnixz:
```suggestion
                                 store: dict[str, Any],
```

sphinx/domains/c/__init__.py
```diff
@@ -666,6 +667,10 @@ class CDomain(Domain):
         'objects': {},  # fullname -> docname, node_id, objtype
     }
 
+    initial_intersphinx_inventory = {
```
picnixz:
```suggestion
    initial_intersphinx_inventory: dict[str, Any] = {
```

Do you think we should only store the symbol (even for domains subclassing the C-domain?) or do you think we should allow a larger type (in which the subclasses would need to fix probably everything?)

sphinx/ext/intersphinx.py
```diff
@@ -29,50 +30,90 @@
...
+    from sphinx.util.typing import ExtensionMetadata, Inventory, RoleFunction
 
     InventoryCacheEntry = tuple[Union[str, None], int, Inventory]
 
 logger = logging.getLogger(__name__)
 
 
-class InventoryAdapter:
-    """Inventory adapter for environment"""
+def process_disabled_reftypes(env: BuildEnvironment) -> None:
```
picnixz:
```suggestion
def _process_disabled_reftypes(env: BuildEnvironment) -> None:
```

sphinx/ext/intersphinx.py
```diff
@@ -29,50 +30,90 @@
...
+        elif ':' in d:
+            domain, typ = d.split(':', 1)
+            if typ == '*':
+                env.intersphinx_all_domain_disabled.add(domain)  # type: ignore[attr-defined]
+            else:
+                env.intersphinx_disabled_per_domain.setdefault(  # type: ignore[attr-defined]
+                    domain, []).append(typ)
+
+
+class EnvAdapter:
```
picnixz: Could we not change the public name? nor the interface that people might be using in their extensions? 

For 8.x, we would add a deprecation warning for InventoryAdapter and use the class you are suggesting instead. But for compat, I would like not to change any "public" names.

sphinx/ext/intersphinx.py
```diff
@@ -271,237 +330,253 @@ def load_mappings(app: Sphinx) -> None:
...
-        unnamed_vals = []
-        for name, _expiry, invdata in intersphinx_cache.values():
-            if name:
-                named_vals.append((name, invdata))
-            else:
-                unnamed_vals.append((name, invdata))
-        for name, invdata in sorted(named_vals) + unnamed_vals:
-            if name:
-                inventories.named_inventory[name] = invdata
+        # old stuff, still used in the tests
```
picnixz, line 378: Since your PR would land in 8.x at earliest, we won't have the old intersphinx format. So we wouldn't need to support the old configurations where you had unnamed inventories.

See #12083.

sphinx/ext/intersphinx.py
```diff
@@ -271,237 +330,253 @@ def load_mappings(app: Sphinx) -> None:
...
+                domain_entries = entries.setdefault(domain_name, {})
+                per_type = domain_entries.setdefault(object_type, {})
+                for object_name, object_data in inv_objects.items():
+                    item_set = per_type.setdefault(object_name, InventoryItemSet())
+                    item_set.append(inv_name, object_data)
+
+        # and then give the data to each domain
+        for domain_name, domain_entries in entries.items():
+            if _debug:
+                global _debug_indent
```
picnixz, line 409: Honestly, I don't want a global variable like that. I'd prefer having a logger.debug() call.

sphinx/ext/intersphinx.py
picnixz, line 1: I won't comment more on this one since you could perhaps refactor it assuming you don't have the old intersphinx mapping's format where you had unnamed inventories.

sphinx/util/inventory.py
```diff
@@ -187,3 +195,94 @@ def escape(string: str) -> str:
...
+
+    The data stored is a list of tuples:
+
+    - Element one is a unique reference given in the intersphinx_mapping
+      configuration variable.
+    - Element two is data about an inventory item in the form of an
+      InventoryItem tuple.
+    """
+
+    def __init__(self, __items: dict[str | None, list[InventoryItem]] | None = None) -> None:
```
picnixz, line 215:
```suggestion
    def __init__(self, items: dict[str | None, list[InventoryItem]] | None = None, /) -> None:
```

sphinx/util/inventory.py
```diff
@@ -187,3 +195,94 @@ def escape(string: str) -> str:
...
+            return InventoryItemSet()
+
+    def make_reference_node(
+        self,
+        domain_name: str,
+        node: pending_xref,
+        contnode: TextElement,
+    ) -> nodes.reference:
+        # TODO: document and test
+        if len(self._items) == 0:
```
picnixz, line 247:
```suggestion
        if not self._items:
```

sphinx/util/inventory.py
```diff
@@ -187,3 +195,94 @@ def escape(string: str) -> str:
...
+            msg = "No inventory items!"
+            raise ValueError(msg)
+
+        legacy_mapping_items = self._items.get(None, [])
+        if len(legacy_mapping_items) == 0:
```
picnixz, line 252:
```suggestion
        if not legacy_mapping_items:
```

sphinx/util/inventory.py
```diff
@@ -187,3 +195,94 @@ def escape(string: str) -> str:
...
+            inv_name = min(filter(None, self._items))
+            proj, version, uri, dispname = self._items[inv_name][0]
+        elif len(legacy_mapping_items) == 1:
+            # Deprecated path for handling pre-Sphinx 1.0 intersphinx_mapping
+            # xref RemovedInSphinx70Warning
```
picnixz, line 257:
```suggestion
            # xref RemovedInSphinx80Warning
```

sphinx/domains/python/__init__.py
jakobandersen: Indeed :-). I missed this when resurrecting the PR. Fixed.

sphinx/domains/__init__.py
jakobandersen: I have added another sentence to the description to clarify that it is up to the concrete domain to design the dictionary, and that intersphinx won't look into it.

sphinx/domains/c/__init__.py
jakobandersen: I have changed it to the concrete type ``dict[str, Symbol]`` now. Indeed, if someone wants to subclass the domain and change this dictionary, they probably need to change everything else related to the delegation.

sphinx/domains/__init__.py
picnixz: Ah, actually I wasn't sure if mypy wouldn't complain if you don't use something like `dict[Any, Any]`. However. you could perhaps make it a ClassVar?

jakobandersen: You mean as in declaring the type to be ``ClassVar[dict]``? Non of the other attributes have this. Perhaps they should?

sphinx/domains/__init__.py
jakobandersen: Good idea. Done.

sphinx/domains/__init__.py
picnixz: Err... probably, unless they could be instance-specific. But I think domains should be considered as singletons (even if multiple instances could exist, there should only be one instance of a specific domain at any time (assuming no parallelism)) so it would capture the semantic a bit better IMO.

I think we didn't type them because this syntax did not exist yet / was not well-supported but if we add new things, I'd suggest that those new things are "clean". And in a separate PR, we cleanup the old things (rather than having a new thing in an old-style).

sphinx/domains/__init__.py
jakobandersen: I have added a docstring and passed everything to it. It really should be kept as an implementation detail that only the domains bundled with Sphinx uses. Third-party domains should override the top-level intersphinx resolution method and do their own thing.

sphinx/domains/std/__init__.py
jakobandersen: Good question, I guess they in practice are most of the time, but I don't know. This code is merely moved from the intersphinx extension.

sphinx/domains/std/__init__.py
jakobandersen: Indeed, I have added some internal documentation in the base class to make it clear that it should be changed carefully.

sphinx/domains/c/_ast.py
jakobandersen: Definitely, it should be easier for other to help maintain them. With limited bandwidth for working on Sphinx it would be great to hear which parts are in most need of documentation. My guess is the lookup code is the worst, and there are some changes coming of the lookup in the C domain, so I can try to add some more documentation with those.

sphinx/domains/c/__init__.py
jakobandersen: Good idea, I'll try to add that when I fix https://github.com/sphinx-doc/sphinx/issues/7819, and the lookup anyway gets a makeover.

sphinx/ext/intersphinx.py
jakobandersen: Yes, I have reinstated the old class and added a deprecation warning.

PR #8940, opened 2021-02-27
Close #8195: Add mathjax_version to choose the version of MathJax

### LINKED ISSUE #8195: The configuration format for MathJax v3 is changed
**Describe the bug**

**To Reproduce**

```
MathJax.Hub.Config({

});
```

```
<script>
window.MathJax = {
  tex: {
    macros: {
      vector: ["\\vec{#1}", 1]
    },
    inlineMath: [["\\(", "\\)"]],
    processRefs: false,
    processEnvironments: false,
    autoload: {
      color: [],
      colorV2: ['color']
    },
  },
  options: {
  },
  loader: {
    load: ['[tex]/noerrors']
  }
};
</script>
```

**Expected behavior**

**Environment info**

**Additional context**

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #8195

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -64,6 +64,8 @@ Features added
 * #2018: html: :confval:`html_favicon` and :confval:`html_logo` now accept URL
   for the image
 * #8070: html search: Support searching for 2characters word
+* #8195: mathjax: Add :confval:`mathjax_version` to choose the version of
+  MathJax
 * #7830: Add debug logs for change detection of sources and templates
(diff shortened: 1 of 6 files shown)


mgeier:
This will silently break (when switching from Sphinx 3 to 4) if the previously used `mathjax_config` is incompatible with MathJax3, right?

I think it would be better if there were at least a warning in this case.

tk0miya:
Withdrawing now.

PR #8970, opened 2021-03-06
Sphinx is available on Chocolatey

naveen521kk:
Subject: Add Installation instructions for Windows using Chocolatey

### Purpose
People can install sphinx using chocolatey.
```ps
choco install sphinx
````
so modified the docs specifying that.


tk0miya:
Thanks!

PR #8971, opened 2021-03-06
Fix mathjax v3 config

### LINKED ISSUE #8195: The configuration format for MathJax v3 is changed

bryanwweber:
Subject: Fix #8195

### Feature or Bugfix

- Bugfix

### Purpose
- As discussed in #8195, the configuration format for MathJax v3 no longer uses the `Hub.Config()` function. This change sets a mapping on the `window.MathJax` object, which works for v2 and v3.
- Add a warning to configuration that appears to be for v2 which won't work for v3.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,8 @@ Incompatible changes
 * #7425: MathJax: The MathJax was changed from 2 to 3. Users using a custom
   MathJax configuration may have to set the old MathJax path or update their
   configuration for version 3. See :mod:`sphinx.ext.mathjax`.
+* #8971: Update MathJax configuration script to support MathJax v2 and v3. Issue
+  warnings when v2 configuration is used with the default MathJax URL.
 * #7784: i18n: The msgid for alt text of image is changed
 * #5560: napoleon: :confval:`napoleon_use_param` also affect "other parameters"
   section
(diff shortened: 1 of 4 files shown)


mgeier:
> This change sets a mapping on the `window.MathJax` object, which works for v2 and v3.

Cool, I didn't know that.

Just out of curiosity: How did you find out about this?
Is this documented somewhere?
Or did you just try it on a hunch?

This nearly works for my use case, but I can't do it without raising the warning, see https://github.com/sphinx-doc/sphinx/issues/8195#issuecomment-792332913.

bryanwweber:
> Just out of curiosity: How did you find out about this?
> Is this documented somewhere?

It is documented, although you have to be reading the documentation with the intent of finding a similarity between 2.7 and 3.x to see it :joy: The 2.7 page is here: http://docs.mathjax.org/en/v2.7-latest/configuration.html#using-plain-javascript and the 3.1 page is here: https://docs.mathjax.org/en/latest/web/configuration.html#using-a-local-file-for-configuration

I was reading those pages and noticed that the examples used `window.MathJax`. I'm not sure why they have 2 different configuration methods in 2.7, probably for backward compatibility. The `window.MathJax` method was only made available in 2.3.

bryanwweber:
@mgeier I pushed an option to suppress the warning called `mathjax_no_v2_warning`. Let me know if you have a better idea for a name :joy:

mgeier:
As mentioned in https://github.com/sphinx-doc/sphinx/issues/8195#issuecomment-792983843, you should define a warning type and subtype instead of using a custom configuration option.

And as also mentioned there, this still doesn't solve my problem.

bryanwweber:
I'm closing this in favor of #9094.

PR #8983, opened 2021-03-11
Fix #7383: autodoc: Support typehints for properties

### LINKED ISSUE #7383: Property annotations do not show up in autoclass

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- py:property directive now outputs py:property directive to describe
 its type annotation.
- refs: #7383


jbms:
Is this going to be merged into 3.x?  I'm not sure I understand sphinx version 4 --- there don't seem to be any packages released for it.

tk0miya:
We don't have plan to release 3.x package more. The next stable release is 4.0.0. Please wait a moment.

Issue #7241, opened 2020-03-02
No wrapping for cpp:enumerator


senyai:
**Describe the bug**

`make latexpdf` renders to:
![image](https://user-images.githubusercontent.com/76137/75689324-0e2d3800-5cb2-11ea-9adb-1efd6d09b5bf.png)

**To Reproduce**
Render this rst:
```
.. cpp:enumerator:: Test::aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa = 41
.. cpp:enumerator:: Test::bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb = 42
.. cpp:enumerator:: Test::cccccccccccccccccccccccccccccccccc = 43
.. cpp:enumerator:: Test::longval = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa | bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb | cccccccccccccccccccccccccccccccccc
```

**Expected behavior**
Word wrapping must occur

**Additional context**

- Somehow I thought it was an issue with breathe [michaeljones/breathe/issues/476]

senyai:
I figured [sphinx/texinputs/sphinx.sty](https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/texinputs/sphinx.sty#L1689) has this trick:
`1689:\newcommand{\pysigline}[1]{\item[{#1}]}` which puts input text into \item label. I think, that this is semantically incorrect and is the root of my problem.

jakobandersen:
@jfbu, is there a way we can get reasonable line wrapping in this case? Except for links and formatting I would typeset the offending line as something like
```latex
Test::longval = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa |
\phantom{Test::longval = }bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb |
\phantom{Test::longval = }cccccccccccccccccccccccccccccccccc
```
(This is the alignment I referred to in https://github.com/sphinx-doc/sphinx/issues/1514#issuecomment-584799919. How does one do the equivalent in HTML + CSS?)
To fix all these issues with line formatting we should probably come up with a set of new doc node types (if docutils doesn't already have it) that the domains can generate to indicate to the builders what the envisioned line break/indentation/wrapper should be. Each builder then implement those as sophisticated as their target allows, or default to the current formatting behaviour.

[...]

Issue #8995, opened 2021-03-13
LaTeX: legacy ``\pysiglinewithargsret`` does not compute correctly available horizontal space and should use a ragged right style


jfbu:
**Describe the bug**

Here is a screenshot from PDF rendering of Sphinx docs:

![Capture d’écran 2021-03-13 à 15 12 22](https://user-images.githubusercontent.com/2589111/111032887-c259b680-840e-11eb-9b91-2a83b77c3076.png)

This is taken from the section documenting the APIs to the autodoc extension.

We observe:

- some stretch of whitespace in signature arguments,
- hyphenation of ``Tuple`` but there seems to be space available

This is explained from the latex code which has a wrong computation of avaiable horizontal space; and the parbox should use ``\raggedright``. This is will reduce hyphenation of keywords also.

**Environment info**
- OS: [e.g. Unix/Linux/Mac/Win/other with version]
- Python version: [e.g. 3.7.1]
- Sphinx version: applies to all versions at least since 1.5 say
- Sphinx extensions:  [e.g. sphinx.ext.autodoc, recommonmark]
- Extra tools: TeXLive 2021

**Additional context**

- Relates indirectly to #7241 and #8980. To fix the later one should use `\parbox` as indicated in https://github.com/sphinx-doc/sphinx/issues/7241#issuecomment-595751032, and this has the correct computation of available width which is wrong in ``\pysiglinewithargsret`` currently.

With some fixes the above screenshot would look like this
![Capture d’écran 2021-03-13 à 15 24 06](https://user-images.githubusercontent.com/2589111/111033184-3f396000-8410-11eb-80da-edde49b89345.png)

It does not go all the way to right end (which would match ``docu-``) as it is know that ``\raggedright`` makes hyphenation very unlikely, but this is a good thing, not to hyphenate too easily in signature, I think.

jfbu:
Fixed at 2934aac441392f70d9baad994ea866f6bdbd8352

Issue #8980, opened 2021-03-10
missing line break in pysigline


ndarmage:
When producing the documentation for tex by ``make latexpdf`` with Sphinx v.3.5.2, I've noticed long lines extending out of the right page margin in the output pdf file with dictionaries used as class (or module) attributes. Minimal example:

```
mod_attr = {'key1': [i for i in range(10)], 'key2': 2, 'key3': 'ok'}
class example_class:
    "my class"
    cls_attr = {'key1': [i for i in range(100)], 'key2': 2, 'key3': 'key3'}
    def __init__(self, a):
        self.a = a
```

The problem is caused by the following line in the file sphinx.sty:
\newcommand{\pysigline}[1]{\item[{#1}]}
The optional argument of the item, that is the label, is typeset in horizontal mode, so that it is not possible to have line breaks. A possible solution is to use a parbox like:
\newcommand{\pysigline}[1]{\item[{\parbox[b]{\textwidth}{#1}}]}
which fixed my problem.

jfbu:
Thanks for report. This is related to #7241. Maybe https://github.com/sphinx-doc/sphinx/issues/7241#issuecomment-595751032 will help you with better workaround for time being.

jfbu:
Fixed at 2934aac441392f70d9baad994ea866f6bdbd8352

ndarmage:
Hello, this issue is marked as close.
I retrieved sphinx v.3.5.3 under Py 3.9.2.
Still the file .../sphinx/texinputs/sphinx.sty contains the old line :
```python
\newcommand{\pysigline}[1]{\item[{#1}]}
```
which reproduces the bad behavior (line extending out of page margin) that started the same issue.

Thanks for your kind attention.
Best regards,

jfbu:
Le 12 avr. 2021 à 21:34, ndarmage ***@***.***> a écrit :

> 
> Hello, this issue is marked as close.
> I retrieved sphinx v.3.5.3 under Py 3.9.2.
> Still the file .../sphinx/texinputs/sphinx.sty contains the old line :
> 
> \newcommand{\pysigline}[1]{\item[{#1}]}
> which reproduces the bad behavior (line extending out of page margin) that started the same issue.
> 
> Thanks for your kind attention.
> Best regards,
> 
> 
[...]

Issue #1514, opened 2015-01-03
Break long function declarations into one parameter per line


shimizukawa:
C function declarations that have many (or long) parameters resulting in a line break are hard to read. Would it be possible to optionally break them with one parameter per line (like doxygen).

I.e. instead of

```
size_t process_function(const void *source, void *destination, size_t length, void *data, int (*callback)(size_t, size_t, size_t, void *),
void *cbparam)
```

I would like something like

```
size_t process_function(const void *source,
                        void *destination,
                        size_t length,
                        void *data,
                        int (*callback)(size_t, size_t, size_t, void *),
                        void *cbparam)
```

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1514
- Originally reported by: Jørgen Ibsen
- Originally created at: 2014-07-20T21:54:00.516

JavierJF:
I would like to participate in the implementation of this feature @shimizukawa. Do you have any plans or known starting point for it?

tk0miya:
This is not an issue that he posted. This was migrated from bitbucket. The original author is Jørgen Ibsen.
>Originally reported by: Jørgen Ibsen

BTW, this is related with writer modules. So I guess `sphinx/writers/html.py` (`html5.py`) is related with this.

JavierJF:
@tk0miya yes, I knew it was a imported issue, but I saw that was labeled by him several times. Thanks for the info about the related module, I have looked into it and I think it's possible to sketch something. But probably the best will be trying to expose a setting for configuring it, like a number of parameters in which start doing the line-breaking.

jakobandersen:
It should be doable only by modifying individual domains. A domain is given a node to render a signature into. That node is a [addnodes.desc_signature](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/addnodes.py#L88). Instead of directly adding child nodes to it, you can set ``is_multiline = True``, and only add [addnodes.desc_signature_line](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/addnodes.py#L99) nodes as children. In those you can then add fragments of a signature. The C++ domain [does this](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/domains/cpp.py#L3205) for templates.

JavierJF:
@jakobandersen Nice info thanks, I'm going to start with that. Reading first how that was integrated in the C++ domain.

hzhwcmhf:
[...]

PR #8997, opened 2021-03-13
LaTeX: improve behavious of `\pysigline` and `\pysiglinewithargsret` in case of multi-line signature

jfbu:
Bugfix

Closes: #7241, #8980, #8995

Regarding #7241, the fix is minimal it produces output like this 
Capture d’écran 2021-03-13 à 17 14 20

The mechanism for signature works in indented context like this (which uses "quoted" indentation so it is on both sides)

Capture d’écran 2021-03-13 à 17 15 16

An example from our docs:

with this PR:

Capture d’écran 2021-03-13 à 17 16 49

without this PR:
Capture d’écran 2021-03-13 à 17 22 22

Notice also vertical spacing between signature and description is improved.

Relates indirectly: #1514

No direct incidence as `\pysiglinewithargsret` already used a `\parbox` we are only adding `\raggedright` and a final `strut`. But mentioning #1514 in this context due to recent https://github.com/sphinx-doc/sphinx/issues/1514#issuecomment-798458599, and also because the above screenshots show that indeed achieving one argument per line would be much better. But this PR is not involved in this directly.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -92,6 +92,7 @@ Bugs fixed
 * #8915: html theme: The translation of sphinx_rtd_theme does not work
 * #8342: Emit a warning if a unknown domain is given for directive or role (ex.
   ``:unknown:doc:``)
+* #7241: LaTeX: No wrapping for ``cpp:enumerator``
 * #8711: LaTeX: backticks in code-blocks trigger latexpdf build warning (and font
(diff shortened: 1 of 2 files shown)


tk0miya (approved):
Cool!

jfbu:
Thanks as always! merging

PR #9000, opened 2021-03-14
Refactor some internal LaTeX code

jfbu:
Codeline wrapping in literal blocks (possibly hard-wrap in presence of
unbreakable long strings) requires a measurement step. Spring 2021 LaTeX
adds overhead to each and every paragraph to insert hooks at
begin and end of paragraphs. For technical reasons, our measurement
(which uses a paragraph in a scratch box) could be broken by the "after
para" hook. Work around that. Perhaps actual Spring 2021 LaTeX will
have some interface to not execute hook, but this should still work.

### Feature or Bugfix

- Refactoring

### DIFF
--- a/sphinx/texinputs/sphinxlatexliterals.sty
+++ b/sphinx/texinputs/sphinxlatexliterals.sty
@@ -318,6 +318,26 @@
 % box does not store in an accessible way what was the maximal
 % line-width during paragraph building.
 %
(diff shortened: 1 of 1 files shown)


tk0miya (approved):
+1 :-)

jfbu:
Thanks for review, merging!

PR #9013, opened 2021-03-16
Add Danish translation of message on search page 

dybber:
Tiny change, hopefully no need to explain this in detail. 

Currently this strings seems to be the only string that is not translated which appears in the user interface, when using Danish translation of Sphinx.


tk0miya:
I copied yout translation to the transifex.com (we've used to manage translation messages on it).

dybber:
Okay, thanks, and sorry for not noticing that this was the wrong channel to submit translations.

PR #7145, opened 2020-02-13
Close #7144: Add CSS class indicating its domain for each desc node

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #7144

PR #9023, opened 2021-03-20
Flexible declaration styling

jakobandersen:
### Feature or Bugfix
- Feature
- Refactoring

### Purpose
Attempt to resolve a foundational chunk of #4289, building on #7145, #6990, #4390.

### Examples of difference in HTML rendering

Current: https://jakobandersen.github.io/ex_sphinx_decl_styling/master/
With this: https://jakobandersen.github.io/ex_sphinx_decl_styling/with_9023/

For seeing the differences for the Python domain (change to monospace font) see e.g., ``extdev/appapi.html``.
For the difference in styling with colouring in the C++ domain see e.g., ``usage/restructuredtext/domains.html#_CPPv46MyList``.

#### Additional example of C++ differences
Current:
master
With this PR:
with_9023

### Details

- Add more nodes derived from ``addnodes.desc_sig_element``,
  and document them as being the leaf-nodes of signatures.
  That is, they only contain text, no other nodes.
- Document the nodes ``desc``, ``desc_signature``, ``desc_signature_line``, and ``desc_content``
  as the nodes that form the top levels of object descriptions.
- Document the remaining ``desc_``nodes as being high-level structuring nodes for signatures.
- Document which classes are always in the ``desc_`` nodes.
- Revamp the C++ domain to use the new nodes, and never ``nodes.Text`` nodes.
- Restructure the CSS in the ``basic`` theme to group by this node classification.
- Move classes on nodes to ``addnodes`` from the HTML and HTML5 writers.
- Introduce a new top-level ``desc_inline`` node for signature fragments in text.
  This is not for xrefs, but for roles like ``cpp:expr`` which contain arbitrary expressions or types.
- Add new classes on the two top-level signature-containing nodes ``desc_signature`` and ``desc_inline``.
  Both get ``sig`` the domain name, and in addition ``sig-object`` and ``sig-inline`` respectively.
  While ``desc`` contains the domain, this also has arbitrary descriptions as decendants,
  so you can have a ``py:function`` with a description that contains a ``cpp:expr``.
  If you want to style, e.g., keywords differently in the two languages, you can now select
  ``.sig.py .k`` and ``.sig.cpp .k`` without them overlapping.
- HTML: Change ``desc_name`` and ``desc_addname`` from using a ``code`` tag to a ``span``,
  and instead make all ``.sig`` use monospace as font-family consistently.
- Add styling to C++ signatures based on the IntellliJ theme.

### Next steps, perhaps for future PRs

- ~~Only the basic theme has been changed. The rest should be checked.~~ (done)
- @jfbu, the Latex implementation of ``desc_inline`` is quick copy paste. The compiled PDF looks fine to me, but please check, e.g., with something like ``  :cpp:expr:`a + b`  ``
- There are some more ``desc_`` nodes perhaps should be looked at:
  ``desc_type``, ``desc_returns``, ``desc_parameterlist``, ``desc_parameter``, ``desc_optional``, and ``desc_annotation``.
  I suggest doing this in a separate PR.
- As discussed in #4289 it may be nice to also have flexible styling in Latex. As the current classes are way fewer than in Pygments I don't think a direct link to their styles is relevant. I suggest doing this in a separate PR if those using the Latex output would like it, but it would be nice to get feedback already now if there is a problem with the design suggested here (ping @jfbu ?).
- ~~Convert the C domain to use the new nodes.~~ (done)
- Convert other domains to use the new nodes?


jakobandersen:
I think it is now in a potentially mergeable state, ready to review.

sphinx/addnodes.py
```diff
@@ -115,25 +115,53 @@ def extract_original_messages(self) -> List[str]:
         return messages
 
 
-# domain-specific object descriptions (class, function etc.)
+#############################################################
+# Domain-specific object descriptions (class, function etc.)
+#############################################################
+
+class _desc_classes_injector(nodes.TextElement):
```
tk0miya: Is any reason to limit this to the child of `TextElement`? How about inheriting `Element` instead?

sphinx/addnodes.py
```diff
@@ -115,25 +115,53 @@ def extract_original_messages(self) -> List[str]:
...
+    """Helper base class for injecting a fixes list of classes.
+
+    Use as the first base class.
+    """
+
+    classes = []  # type: List[str]
```
tk0miya: Since v4.0, we can use variable annotations syntax because we drop the py35 support. How about using it?
```suggestion
    classes: List[str] = []
```

sphinx/addnodes.py
```diff
@@ -336,6 +452,7 @@ class pending_xref(nodes.Inline, nodes.Element):
     These nodes are resolved before writing output, in
     BuildEnvironment.resolve_references.
     """
+    child_text_separator = ''
```
tk0miya, line 455: Is this change needed?

sphinx/transforms/post_transforms/__init__.py
```diff
@@ -222,24 +218,35 @@ def has_visitor(translator: Type[nodes.NodeVisitor], node: Type[Element]) -> boo
...
+
+    def fallback(self, nodeType: Any) -> None:
+        for node in self.document.traverse(nodeType):
             newnode = nodes.inline()
             newnode.update_all_atts(node)
             newnode.extend(node)
             node.replace_self(newnode)
 
 
+class DescSigAddDomainAsClass(SphinxPostTransform):
```
tk0miya: How about this name?
```suggestion
class PropagateDescDomain(SphinxPostTransform):
```

"propagate" is a term already used in docutils: `docutils.transforms.references.PropagateTargets`

sphinx/writers/manpage.py
```diff
@@ -139,6 +146,21 @@ def visit_desc_signature_line(self, node: Element) -> None:
     def depart_desc_signature_line(self, node: Element) -> None:
         self.body.append(' ')
 
+    def visit_desc_content(self, node: Element) -> None:
+        self.visit_definition(node)
+
+    def depart_desc_content(self, node: Element) -> None:
+        self.depart_definition(node)
+
```
tk0miya, line 154: Is `*_desc_inline` needed here?

sphinx/addnodes.py
```diff
@@ -144,18 +172,63 @@ def child_text_separator(self):
...
+
+    For example, in the declaration of a Python class ``MyModule.MyClass``,
+    the main name is ``MyClass``.
+
+    This node always has the class ``sig-name``.
+    """
+    classes = ['sig-name', 'descname']  # 'descname' is for backwards compatibility
+
+
+class desc_addname(_desc_classes_injector, nodes.Part, nodes.Inline, nodes.FixedTextElement):
```
tk0miya, line 222: Adding a new node brings an incompatible change for 3rd party writers. It would be better to provide a fallback to them like `SigElementFallbackTransform` to keep them working (for a while).

tk0miya:
LGTM with nits!

sphinx/addnodes.py
jakobandersen: Yes, ``nodes.Element`` has ``child_text_separator = '\n\n' ``.

sphinx/writers/manpage.py
jakobandersen: That depends, I have added ``addnodes.desc_inline`` in ``SigElementFallbackTransform``, so technically it isn't needed. But if the goal is to minimize ``SigElementFallbackTransform``, then sure, I can add visit/depart for it as well.

sphinx/addnodes.py
jakobandersen: This is fortunately not a new node, I just moved them a bit around so the order better fits the purpose of them.

sphinx/transforms/post_transforms/__init__.py
jakobandersen: Good idea.

sphinx/addnodes.py
jakobandersen: Indeed, good idea.

sphinx/addnodes.py
jakobandersen: Not really, will change to ``Element``.

jakobandersen:
> LGTM with nits!

Thanks! It should be ready for a second round now.

As a general point: what is your opinion on how to use ``SigElementFallbackTransform``? If we should minimize the use of it, I suggest we add warnings whenever it triggers so builders can become aware of missing implementations.

sphinx/addnodes.py
tk0miya: It seems this comment has been drifted. Originally, I commented to the `desc_inline` node.

sphinx/addnodes.py
tk0miya: I understand this affects to `astext()`. +1 for the change.

sphinx/writers/manpage.py
tk0miya: IMO, adding handlers makes our code explicit. So I prefer to add it here. The fallback transform is a bit magical and hard to understand its behavior on debugging.

jakobandersen: Indeed, I will add ``desc_inline`` handlers for the all the writers.

sphinx/addnodes.py
jakobandersen: Ah, makes sense. That addition to ``SigElementFallbackTransform`` should already be there. I change its logic such that:
- if there is no ``desc_inline`` handler, then transform them.
- (as before) if not all ``desc_sig_element`` have handlers, then transform them all.

tk0miya:
>As a general point: what is your opinion on how to use SigElementFallbackTransform? If we should minimize the use of it, I suggest we add warnings whenever it triggers so builders can become aware of missing implementations.

Ideally, all writers should support all of the new nodes. But it increases the maintenance cost of the 3rd party builders. So I'm afraid that the warnings kill their deprecation because of the lack of maintenance. So -0 for emitting warnings.

jakobandersen:
> So I'm afraid that the warnings kill their deprecation because of the lack of maintenance.

Good point, as noted in the review comment, I'll ensure that at least the writers bundled with Sphinx only needs the fallback for the ``desc_sig_element`` nodes. After that I'll merge into 4.0.0, ok?

tk0miya:
Yes. I'll review this again tomorrow (or the next day). After that, let's merge this into 4.0.x branch.

sphinx/addnodes.py
tk0miya: Reasonable. +1 for the change.

tk0miya:
LGTM again. +1 for merging if fallback-transform supports the `desc_inline` node.

jakobandersen:
Thanks for the reviews! I'll rebase to 4.0.x and merge.

Issue #5589, opened 2018-11-03
autosummary: determine "canonical" name of the class in template


MarSoft:
Subject: it is currently impossible to determine canonical class name from the template

### Problem
- When classes to be included in the summary are referenced by non-canonical paths, `:show-inheritance:` parameter for `autoclass` block produces code referencing the canonical name of the class, but `{{ objname }}` value is the non-canonical one.
- As a result, the link to parent class in `Bases:` is broken and is not clickable.
- If there would be an option to determine canonical name (`myclass.__module__ + '.' + myclass.__name__`) then I could include that as a TOC entry or something like that.

#### Procedure to reproduce the problem
I'll illustrate it with the following minimum example:
```
# mypkg/__init__.py
from .parent import Parent
from .child import Child
```
```
# mypkg/parent.py
class Parent:
    pass
```
```
# mypkg/child.py
from .parent import Parent
class Child(Parent):
    pass
```
```
# docs/_templates/autosummary/class.rst
{{ fullname | escape | underline }}

.. autoclass:: {{objname}}
   :show-inheritance:
   :members:
```
```
# docs/index.rst
.. autosummary::
   :toctree: _autosummary

   mypkg.Parent
   mypkg.Child
```

#### Error logs / results
- Build the docs and observe that class referenced as `mypkg.Child` is displayed as so, but its base class is listed as `mypkg.parent.Parent` and it not clickable.

#### Expected results
It should be possible to determine canonical class name, or class' actual module name.
For now, template variable `module` is derieved from the textual name given as the input. It is not currently possible to access `myobject.__module__`.

### Environment info
- OS: Arch Linux
- Python version: 3.6, 3.7
- Sphinx version: Sphinx==1.8.0

MuellerSeb:
I ran into the same problem!

tk0miya:
Indeed, this is broken. +1 to give canonical name (alias name?).

MuellerSeb:
Anything new here?
One solution could be to provide an `alias_overrides` dictionary in `conf.py` where you can redirect links like it is done in scanpydoc:
https://icb-scanpydoc.readthedocs-hosted.com/en/latest/scanpydoc.elegant_typehints.html

flying-sheep:
An actual fix is in the works, but still not ready: #8449

@cblegare are you going to finish that or are there some blockers?

tk0miya:
[...]

Issue #4944, opened 2018-05-08
python bases in packages


Eldinnie:
### Problem
When documented our library we would like the base classes to be displayed as you can import them (and they're documented aswell)

let's assume the following

```python
# telegram/base.py

class TelegramObject(object):
  pass
```

```python
#telegram/chat.py

from telegram import TelegramObject
class Chat(TelegramObject):
  pass
```

```python
#telegram/__init__.py

from .base import TelegramObject
from .chat import Chat
__all__ = ['TelegramObject', 'Chat']
```

```yaml
index.rst

.. toctree::

  telegram.telegramobject
  telegram.chat
```

```yaml
telegram.telegramobject.rst

telegram.TelegramObject
=======================

.. autoclass:: telegram.TelegramObject
    :members:
    :show-inheritance:
```

```yaml
telegram.chat.rst

telegram.Chat
=============

.. autoclass:: telegram.Chat
    :members:
    :show-inheritance:
```
#### Procedure to reproduce the problem
run sphinx-build -b html

#### Error logs / results
Chat shows it's base as:
`telegram.base.TelegramObject`
http://python-telegram-bot.readthedocs.io/en/stable/telegram.chat.html

#### Expected results
I would like Chat to show it's base as `telegram.TelegramObject`
That way it would link through and we can properly show inheritence
There's also a issue #3014 which would also be fine so we can fix it ourselves

### Reproducible project / your project
https://github.com/python-telegram-bot/python-telegram-bot/tree/master/docs

### Environment info
- OS: Linux/Win/whatever runs on rtd
- Python version: 2.7, 3.6 or 3.7
- Sphinx version: 1.6.5, 1.7.2

MuellerSeb:
https://github.com/sphinx-doc/sphinx/issues/5589 is similar.

tk0miya:
Now I merged #9026 that generates :canonical: option for the imported classes. It allows to refer to the python objects by its canonical name. It will resolve this problem.
Thanks,

tk0miya:
Ah, sorry. #9026 does not resolve this issue. It allows to make a hyperlink as `telegram.base.TelegramObject`, not `telegram.TelegramObject`. So this is not resolved yet. So reopening now.

Issue #4826, opened 2018-04-09
Put qualname in the objects.inv


flying-sheep:
Due to type annotations, more and more tools retrieve qualified object names from live objects and expect them to serve as object ID.

However, code like the following fails when the object is reexported, when its canonical location doesn’t fit its qualified name:

```py
from scipy.sparse import spmatrix

qualname = f'{spmatrix.__module__}.{spmatrix.__qualname__}'  # 'scipy.sparse.base.spmatrix'


from sphinx.ext.intersphinx import inspect_main

sys.stdout = StringIO()
inspect_main(['https://docs.scipy.org/doc/scipy/reference/objects.inv'])

assert qualname in sys.stdout.getvalue()  # nope, only 'scipy.sparse.spmatrix'
```

a fix could either go into the [inventory dumping](https://github.com/sphinx-doc/sphinx/blob/b2bd9f75a66bc8979d198453393c565bfd6db6b9/sphinx/util/inventory.py#L177-L189) or the [python domain](https://github.com/sphinx-doc/sphinx/blob/9efaf18852fb076893e5a25aa624c0cba0316fa8/sphinx/domains/python.py) (if adding objects goes through a domain API and isn’t just doing `env.domains['py'].objects.append(…)`)

tk0miya:
Please let me know your goal. The inventory file of intersphinx is generated for cross reference.
Did you mean that you want to reference `spmatrix` by qualname (ex. `:mod:scipy.sparse.base.spmatrix`)?

flying-sheep:
Yes, since there is no automatic way of knowing what the “canonical” place for an object is.

flying-sheep:
@tk0miya do you need more information? I’d do a PR, but I don’t know enough about the inner workings here and how to start best.

tk0miya:
Sorry for late response.

I feel a bit strange for your proposal. I think the qualified names are not **undocumented**. That means it is private members. I don't think it is good way to expose qualified names automatically.

flying-sheep:
You’re right, but I didn’t propose to just add the qualname to the objects.inv. I do want a solution for the case where I know the qualname but not the canonical documentation location.

Each documented object has 1–∞ aliases, one of which is the the canonical documented location (in objects.inv) and one of which is the canonical qualified name (`thing.__module__ + '.' + thing.__(qual)name__`). Those to names might be (but aren’t always) the same.

[...]

PR #9026, opened 2021-03-21
Close #5603: autodoc: Allow to refer to a python object using canonical name

### LINKED ISSUE #5603: Help people using Intersphinx against objects imported from a package __init__.py

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- This generates `:canonical:` option for `:py:class:` directive if the
target class is imported from other module.  It allows users to refer it
using both the new name (imported name) and the original name (canonical
name).
- It helps a library that implements some class in private module (like
`_io.StringIO`), and publish it as public module (like `io.StringIO`).
- refs: #4826, #4944, #4065, #5603, #5589, #8449

### Remaining Tasks
- [x] ~~Support imported functions~~

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1590,6 +1590,20 @@ class ClassDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # type: 
 
         return []
 
+    def get_canonical_fullname(self) -> Optional[str]:
+        __modname__ = safe_getattr(self.object, '__module__', self.modname)
+        __qualname__ = safe_getattr(self.object, '__qualname__', None)
+        if __qualname__ is None:
+            __qualname__ = safe_getattr(self.object, '__name__', None)
+        if __qualname__ and '<locals>' in __qualname__:
+            # No valid qualname found if the object is defined as locals
+            __qualname__ = None
+
+        if __modname__ and __qualname__:
+            return '.'.join([__modname__, __qualname__])
+        else:
+            return None
+
     def add_directive_header(self, sig: str) -> None:
         sourcename = self.get_sourcename()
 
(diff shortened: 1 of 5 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -1305,13 +1305,28 @@ def format_args(self, **kwargs: Any) -> str:
     def document_members(self, all_members: bool = False) -> None:
         pass
 
+    def get_canonical_fullname(self) -> Optional[str]:
+        __modname__ = safe_getattr(self.object, '__module__', self.modname)
+        __qualname__ = safe_getattr(self.object, '__qualname__', None)
+        if __qualname__ is None:
+            __qualname__ = safe_getattr(self.object, '__name__', None)
+
+        if __modname__ and __qualname__:
```
tk0miya: We have to check the qualname does not contain `<local>`.

sphinx/ext/autodoc/__init__.py
```diff
@@ -1305,13 +1305,31 @@ def format_args(self, **kwargs: Any) -> str:
...
     def add_directive_header(self, sig: str) -> None:
         sourcename = self.get_sourcename()
         super().add_directive_header(sig)
 
         if inspect.iscoroutinefunction(self.object):
             self.add_line('   :async:', sourcename)
 
+        canonical_fullname = self.get_canonical_fullname()
+        if canonical_fullname and self.fullname != canonical_fullname:
+            self.add_line('   :canonical: %s' % canonical_fullname, sourcename)
```
tk0miya: If my understanding is correct, the canonical option is mainly needed for class references because autodoc generates hyperlinks for "real" parent classes and "real" type annotations from its implementation. On the other hand, references to "real" functions will not be generated by autodoc. So I thought canonical-option support for FunctionDocumenter would not be needed.

shimizukawa: +1. I think the function should be out of scope.

shimizukawa (approved):

sphinx/ext/autodoc/__init__.py
tk0miya: Thank you for your wisdom comment!

tk0miya:
Thank you for reviewing @shimizukawa ! Now merging.

0xMashiro:
Long-awaited enhancement ~ 👍

PR #9033, opened 2021-03-25
Fix some local parameter names

stratakis:
Uncovered by a coverity scanner. Those seems to be typos or copy paste errors.


tk0miya:
Thanks!

PR #9036, opened 2021-03-26
HTML Templates: Add blocks to search page

Blendify:
This allows themes to customize the default search page.

Subject: Add blocks to search page

### Feature or Bugfix
- Feature

### Purpose
- Some themes may want to for example style the search box differently, instead of writing the whole search page they can easily override a single block.

### Detail
- None

### Relates
- None


tk0miya:
Thank you for your proposal. Merged!

PR #9037, opened 2021-03-27
Fix error in Pybind11 docstrings

loriab:
Subject: Fix error in Pybind11 docstrings

### Feature or Bugfix

- Bugfix

### Purpose
- Would you consider this fix for an uninvestigated bug? A year ago I noticed that certain docstrings generated by https://github.com/pybind/pybind11 return an object and thus cause Sphinx to error (on `splitlines()`). Since the edited function is already declaring `s` as `str` in the signature, `str(s)` didn't seem too outrageous. Feel free to close if it is.

### DIFF
--- a/sphinx/util/docstrings.py
+++ b/sphinx/util/docstrings.py
@@ -57,10 +57,10 @@ def prepare_docstring(s: str, ignore: int = None, tabsize: int = 8) -> List[str]
     if ignore is None:
         ignore = 1
     else:
-        warnings.warn("The 'ignore' argument to parepare_docstring() is deprecated.",
+        warnings.warn("The 'ignore' argument to prepare_docstring() is deprecated.",
                       RemovedInSphinx50Warning, stacklevel=2)
 
-    lines = s.expandtabs(tabsize).splitlines()
+    lines = str(s).expandtabs(tabsize).splitlines()
     # Find minimum indentation of any non-blank lines after ignored lines.
     margin = sys.maxsize
     for line in lines[ignore:]:


tk0miya:
If my remember is correct, the `s` argument comes from `__doc__` attribute of the target object. Are you saying the objects generated by pybind11 have invalid (non-string) docstring, right? Surely, the patch is very small and no side effects. So it's easy to merge it. But I need to know the root cause of the error. I worry about the invalid docstrings causes another trouble.

loriab:
Thanks for your comments, @tk0miya. I've probed further and submitted a possible fix over at pb11. I don't know why they're not returning strings in the first place, so a fix here may be less disruptive than a fix there. We can wait and see what the pb11 devs say. I'm including a traceback snippet here in case it helps future searchers.

```
  File "/home/psilocaluser/toolchainconda/envs/ghadocs3/lib/python3.8/site-packages/sphinx/util/docstrings.py", line 62, in prepare_docstring
    lines = s.expandtabs(tabsize).splitlines()
AttributeError: 'pybind11_static_property' object has no attribute 'expandtabs'
```

tk0miya:
I consider this is a bug of pybind11. So I prefer to fix this by the pybind11 side.

loriab:
Thanks for closing this, @tk0miya. For anyone also hitting this issue, it turns out to be a bad reaction between pybind11 enums and the automodapi Sphinx extension, explained here https://github.com/pybind/pybind11/pull/2925#issuecomment-839424278 . No solution yet but an easy patch.

PR #9057, opened 2021-04-04
refactor: Use PEP-526 based variable annotation

tk0miya:
### Feature or Bugfix
- Refactoring

### DIFF
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -60,7 +60,7 @@ import sys
 import warnings
 from os import path
 from types import ModuleType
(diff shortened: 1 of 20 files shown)


sphinx/ext/autosummary/__init__.py
```diff
@@ -579,7 +579,7 @@ def get_import_prefixes_from_env(env: BuildEnvironment) -> List[str]:
     Obtain current Python import prefixes (for `import_by_name`)
     from ``document.env``
     """
-    prefixes = [None]  # type: List[str]
+    prefixes: List[str] = [None]
```
tk0miya:
```suggestion
    prefixes: List[Optional[str]] = [None]
```

Issue #9065, opened 2021-04-07
contents directive style breaks with docutils 0.17


ericvanbuhler:
After upgrading `docutils` to 0.17, the local tables of contents on my documentation website changed from begin displayed as bulleted lists with indentation:

![image](https://user-images.githubusercontent.com/10039915/113921704-2b78e380-979b-11eb-8292-b810e51034f4.png)

to a flat list:

![image](https://user-images.githubusercontent.com/10039915/113921571-07b59d80-979b-11eb-9310-75d6e61acbc8.png)

The key difference is the way the headers are defined. with `docutils` 0.16 they are defined like this:
```
<div class="section" id="header-1">
```

and with `docutils` 0.17 they are defined like this:
```
<section id="header-1">
```

**To Reproduce**

Here is the `requirements.txt` file for my project:
```
sphinx==3.3.1
recommonmark
sphinx_markdown_tables
```

Here is a sample of what the .rst file looks like:
```
Title
+++++

Content goes here...

.. contents:: :local:

Header 1
--------

Content goes here...

Subheader 1
-----------

Content goes here...

Header 2
--------

Content goes here...

Subheader 2
-----------

Content goes here...
```

I built with the following command:
```
$ make html SPHINXOPTS="-W --keep-going"
```

**Expected behavior**
I expected the local table of contents to be formatted as a bulleted list with indentation.

**Environment info**
- OS: Debian Linux in Docker
- Python version: 3.6.9
- Sphinx version: 3.3.1
- Sphinx extensions:  'sphinx.ext.autodoc', 'recommonmark', 'sphinx_markdown_tables', 'sphinx.ext.napoleon', 'sphinx.ext.autosectionlabel'

**Workaround**
I was able to work around the issue by locking the `docutils` version to 0.16 in my requirements.txt file:
```
docutils==0.16
sphinx==3.3.1
recommonmark
sphinx_markdown_tables
```

tk0miya:
What theme do you use? It seems the default theme of Sphinx; alabaster works fine.

mgeier:
This sounds similar to https://github.com/readthedocs/sphinx_rtd_theme/issues/1115.

ericvanbuhler:
[...]

Issue #9061, opened 2021-04-05
Newest docutils (v0.17) breaks rendering


kevinzakka:
It seems the newest docutils has broken html rendering in a few places. I'm attaching screenshots showing the output of `make html` with 0.16 and comparing it to 0.17.

**Screenshots**
<img src="https://user-images.githubusercontent.com/10518920/113529642-cfcf1000-9578-11eb-8160-3701d5cc8a59.png" height="200">
<img src="https://user-images.githubusercontent.com/10518920/113529649-d198d380-9578-11eb-8749-94c846bce3fa.png" height="200">

**Your project**
https://github.com/kevinzakka/torchkit

**Environment info**
- OS: [e.g. mac OS]
- Python version: [e.g. 3.8/3.9]
- Sphinx version: [e.g. 3.5.2]
- Sphinx extensions:  [e.g. sphinx.ext.autodoc]

kevinzakka:
Oops, I missed https://github.com/sphinx-doc/sphinx/pull/9059 .. Closing this for now!

mgeier:
This should fix the issue: #9053.

rkdarst:
list items `li` in unordered lists `ul` is also broken (bullet point ::marker are missing).  docutils!=0.17 also fixes it for me.  I haven't dug much deeper yet.

mgeier:
@rkdarst I cannot reproduce the problem. Can you please create a new issue for this, ideally with example reST code and a screenshot? Please also specify which HTML theme you are using.

rkdarst:
You are right, I could only reproduce it in sphinx_rtd_theme, not the default Sphinx theme.  I filed the issue here: https://github.com/readthedocs/sphinx_rtd_theme/issues/1115  .  Reproducing my issue can be found in that issue.

I haven't dug deeper to see if there could be anything else to do in Sphinx itself, but browsing sphinx_rtd_theme's docs I don't see anything obvious.  I guess there is possibility that each theme needs to update some things, but I know nothing about that so far.

Issue #9051, opened 2021-04-03
toctree captions are missing the `caption` class on the main `p` element


ocharles:
**Describe the bug**

Please see https://github.com/readthedocs/sphinx_rtd_theme/issues/1111, though I think I'm encountering a Sphinx bug, rather than a bug in that particular theme.

**To Reproduce**
```
<Paste your command-line here which cause the problem>

$ git clone git://github.com/circuithub/rel8
$ cd docs
$ pip install -r requirements.txt
$ make html
$ # open _build/html/index
```

**Expected behavior**

Each `toctree` caption would be a `<p class="caption">` element, which `sphinx-rtd-theme` will style specially.

**Screenshots**
Screenshots are on https://github.com/readthedocs/sphinx_rtd_theme/issues/1111

**Environment info**
- OS: Linux NixOS
- Python version: 3.8.8
- Sphinx version: 3.5.3

tk0miya:
Good timing! I posted https://github.com/sphinx-doc/sphinx/pull/9053 to fix the trouble. Please try it if you have time :-)

Blendify:
We have some reports on the rtd theme that unordered list are also broken.

tk0miya:
@Blendify 

>We have some reports on the rtd theme that unordered list are also broken.

Could you explain this in detail?

Blendify:
After looking into it, this seems not related to sphinx but our css. The issue comes from the change from `<div class="section">` to `<section>` or css selector for lists is `.section ol` we will need to update our css to fix this. The built sphinx theme may also need to be updated. I see at least an issue here: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/static/basic.css_t#L18

tk0miya:
@Blendify Do you have any example to reproduce the HTML output? I'd like to fix CSS, but I need to reproduce it before fix it.

tk0miya:
I merged #9053 to the 3.5.x branch now. So the main topic of this issue is resolved in the next bugfix release.
I'll wait for a comment of Blendify for a little and will close this.

Blendify:
Just to clarify, the issue with unordered lists is specific to the rtd-sphinx-theme and is an issue on our end.

> Do you have any example to reproduce the HTML output? I'd like to fix CSS, but I need to reproduce it before fix it.

With docutils 0.17 rst headings get <section> before 0.17 they get <section>.

Just use a simple page like:

```
*****
Title
*****

Section 1
=======

[...]

PR #9071, opened 2021-04-08
Pin Sphinx 3.x to docutils <0.17

ericholscher:
This will address the issues that the latest docutils release caused.
I think this is a good practice in general to make sure we have a defined range of docutils versions,
given that they might change in the future.
Having this defined will mean that 3.x versions of Sphinx will always work,
even when docutils has advanced with additional backwards incompatible features.

We've had more than 10 support requests about this in the past week, so I think this is hitting a lot of users throughout the Sphinx ecosystem and it's very hard to debug if you don't understand what is happening. 

Refs #9065 #9063 #9061 #9051

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ install_requires = [
     'sphinxcontrib-qthelp',
     'Jinja2>=2.3',
     'Pygments>=2.0',
-    'docutils>=0.12',
+    'docutils>=0.12,<0.17',
     'snowballstemmer>=1.1',
     'babel>=1.3',
     'alabaster>=0.7,<0.8',


tk0miya:
Could you rebase this into 3.5.x branch? Then I'll merge this soon.

Note: 3.5.4 will be the last release of 3.x series. So I will merge this to the package. But I will not merge this into 4.x series because it should go forward. No reason to restrict the version.

ericholscher:
@tk0miya Done 👍

ericholscher:
I do think that specifically defining a upper bounds of docutils support in Sphinx is a good ongoing policy, but that's a larger discussion. It certainly has downsides, but it could be useful to have Sphinx increase the docutils dependency after it's been tested against the currently released version. Currently I don't believe docutils is testing Sphinx in it's releases, so it seems we'll hit this issue again whenever docutils makes a useful but breaking change.

tk0miya:
Thanks!

PR #9073, opened 2021-04-09
Depart method for node is optional

hason:
Subject: Fix type of the depart method for node

### Feature or Bugfix

- refactoring

### Purpose
- fix mypy error


tk0miya (commented):
LGTM!

tk0miya (approved):

Issue #9078, opened 2021-04-10
autodoc: Async staticmethods and classmethods are considered as not async


tk0miya:
**Describe the bug**
autodoc: Async staticmethods and classmethods are considered as not async.
Since 3.10.0a7, the instances of staticmethod and classmethod have `__wrapped__` attribute. It confuses `sphinx.util.inspect:iscoroutinefunction()`.

refs: 
* https://docs.python.org/3.10/whatsnew/3.10.html#other-language-changes
* https://bugs.python.org/issue43682

**To Reproduce**
see https://github.com/sphinx-doc/sphinx/pull/9077/checks?check_run_id=2312247372

PR #9079, opened 2021-04-10
Fix autodoc: Async staticmethods/ classmethods are considered as not async

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- Since 3.10.0a7, the instances of staticmethod and classmethod have
`__wrapped__` attribute. It confuses sphinx.util.inspect:iscoroutinefunction().

- refs:
  * https://github.com/sphinx-doc/sphinx/issues/9078
  * https://docs.python.org/3.10/whatsnew/3.10.html#other-language-changes
  * https://bugs.python.org/issue43682

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,8 @@ Features added
 Bugs fixed
 ----------
 
+* #9078: autodoc: Async staticmethods and classmethods are considered as non
+  async coroutine-functions
 * #8870: The style of toctree captions has been changed with docutils-0.17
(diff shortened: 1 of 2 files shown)


CHANGES
```diff
@@ -18,6 +18,8 @@ Features added
 Bugs fixed
 ----------
 
+* #9078: autodoc: Async staticmethods and classmethods are considered as non
+  async coroutine-functions
```
tk0miya: It would be better to mention the version of python here.

PR #9094, opened 2021-04-13
Add `mathjax3_config` config option

### LINKED ISSUE #8195: The configuration format for MathJax v3 is changed

mgeier:
This is a draft for an alternative to #8940 and #8971, and it might eventually close #8195 (or not!).

The limitation (for me) is that this still doesn't solve my problem mentioned in https://github.com/sphinx-doc/sphinx/issues/8195#issuecomment-792332913.

Let me describe the problem in more detail:

In `nbsphinx` I would like to provide default values for both `mathjax2_config` and `mathjax3_config`, in order for users to be free to choose whichever version they want with `mathjax_path` and always get my defaults.

If I provide defaults for both `mathjax2_config` and `mathjax3_config`, the current logic in this PR will not raise a warning.
This may sound OK, but if a user previously (and correctly) used `mathjax_config`, their custom config will stop working and there will be no warning!

I could instead only provide a default in `nbsphinx` for `mathjax3_config` (and leave `mathjax2_config`/`mathjax_config` untouched), but this wouldn't help in the previous case where a user uses `mathjax_config`, because the warning would still not be shown.

Any ideas how to make this work?

### Feature or Bugfix

Something in-between.

### Purpose

Provide a way to specify MathJax v3 settings.

### Details

If both `mathjax2_config` and `mathjax3_config` are given, both configurations will be written to the HTML output.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -47,6 +47,8 @@ Features added
 * #9103: LaTeX: imgconverter: conversion runs even if not needed
 * #8127: py domain: Ellipsis in info-field-list causes nit-picky warning
 * #9023: More CSS classes on domain descriptions, see :ref:`nodes` for details.
+* #8195: mathjax: Rename :confval:`mathjax_config` to
+  :confval:`mathjax2_config` and add :confval:`mathjax3_config`
 
 Bugs fixed
 ----------
--- a/doc/usage/extensions/math.rst
+++ b/doc/usage/extensions/math.rst
@@ -144,7 +144,8 @@ are built:
    Version 4.0 changes the version of MathJax used to version 3. You may need to
    override ``mathjax_path`` to
    ``https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js?config=TeX-AMS-MML_HTMLorMML``
-   or update your configuration options for version 3.
+   or update your configuration options for version 3
+   (see :confval:`mathjax3_config`).
 
 .. versionadded:: 1.1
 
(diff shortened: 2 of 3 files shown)


tk0miya:
I also don't have a magical bullet for this problem. But I think this would help for Sphinx users who have their own `mathjax_config` for v2. So I'd like to merge this to Sphinx-4.0. What do you think?

mgeier:
It's not perfect, but I think we can use this!

Thanks to the new explicit config names `mathjax2_config` and `mathjax3_config`, I was able to make this work with `nbsphinx`, see https://github.com/spatialaudio/nbsphinx/pull/558.
In this, I'm only providing default values for one of the MathJax versions and there is only one very little edge case where my default values get lost, but I can live with that.

mgeier:
@tk0miya Should I add a `type` and `subtype` for the warning? If yes, which one?

If you decide to merge this, I can add some documentation (either here or in a separate PR).

tk0miya:
To suppress the warning is meaningless to me. If my understanding correct, the math configuration would not work when the warning is shown. Users should update their configuration instead of suppressing. So it's not needed.

>If you decide to merge this, I can add some documentation (either here or in a separate PR).

I'll merge this after a while (maybe 3-4 hours later). If you have time until then, please add a document and CHANGES entry. It's helpful to me.

mgeier:
OK, I'm on it ...

mgeier:
OK, done, I've added some docs: https://sphinx--9094.org.readthedocs.build/en/9094/usage/extensions/math.html#confval-mathjax3_config

I didn't know a good example value for `mathjax3_config`. I you know one, please let me know or feel free to add it yourself.

tk0miya:
Thank you for your update! Merging now.

tk0miya:
Oops. I wrongly merged this into the 4.x branch. I'll cherry-pick this into the 4.0.x branch manually.

mgeier:
Thanks @tk0miya!

choldgraf:
This looks great, thanks very much @mgeier - many will benefit from this!

PR #9096, opened 2021-04-13
Fix progress logging for parallel builds

peterbell10:
### Feature or Bugfix
- Bugfix

### Purpose
Currently, the progress bar increases when each parallel task is submitted for processing. However, that doesn't necessarily mean any progress has actually been made yet. The result is the progress bar very quickly jumps to 100% and stalls at

> waiting for workers ...

This updates the progress bar as individual tasks are joined, for long-running builds this creates a meaningfully sense of progress. This both creates a nicer user-experience for users, and also prevents CI jobs from timing out from lack of activity.

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -23,7 +23,8 @@ from sphinx.errors import SphinxError
 from sphinx.events import EventManager
 from sphinx.io import read_doc
 from sphinx.locale import __
-from sphinx.util import import_object, logging, progress_message, rst, status_iterator
+from sphinx.util import (display_chunk, import_object, log_status_message, logging,
+                         progress_message, rst, status_iterator)
 from sphinx.util.build_phase import BuildPhase
 from sphinx.util.console import bold  # type: ignore
 from sphinx.util.docutils import sphinx_domains
@@ -450,20 +451,29 @@ class Builder:
             # allow pickling self to send it back
             return pickle.dumps(self.env, pickle.HIGHEST_PROTOCOL)
 
+        chunks = make_chunks(docnames, nproc)
+
+        chunks_complete = 0
+        summary = bold(__('reading sources... '))
+
         def merge(docs: List[str], otherenv: bytes) -> None:
+            # Update progress message for each chunk completed
+            nonlocal chunks_complete
+            chunks_complete += 1
+            log_status_message(display_chunk(docs), summary, 'purple', chunks_complete,
+                               len(chunks), self.app.verbosity)
+
+            # Merge build environment from sub-process
             env = pickle.loads(otherenv)
             self.env.merge_info_from(docs, env, self.app)
 
         tasks = ParallelTasks(nproc)
-        chunks = make_chunks(docnames, nproc)
-
-        for chunk in status_iterator(chunks, __('reading sources... '), "purple",
-                                     len(chunks), self.app.verbosity):
+        for chunk in chunks:
             tasks.add_task(read_process, chunk, merge)
 
         # make sure all threads have finished
-        logger.info(bold(__('waiting for workers...')))
         tasks.join()
+        logger.info('')
 
     def read_doc(self, docname: str) -> None:
         """Parse a file and add/update inventory entries for the doctree."""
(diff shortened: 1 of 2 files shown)


sphinx/util/__init__.py
```diff
@@ -660,25 +660,30 @@ def old_status_iterator(iterable: Iterable, summary: str, color: str = "darkgree
...
 
+def log_status_message(item: str, summary: str, color: str, progress: int,
+                       length: int, verbosity: int = 0) -> None:
+    s = '{}[{:3d}%] {}'.format(summary, (100 * progress) // length,
+                               colorize(color, item))
+    if verbosity:
+        s += '\n'
+    else:
+        s = term_width_line(s)
+    logger.info(s, nonl=True)
```
tk0miya: I think the original code is a bit odd because output "\n" with `nonl=True` option. So it's better to output without the option in verbose mode:
```suggestion
    if verbosity:
        logger.info(s)
    else:
        logger.info(term_width_line(s), nonl=True)
```

sphinx/builders/__init__.py
```diff
@@ -450,20 +451,29 @@ def read_process(docs: List[str]) -> bytes:
...
+        chunks = make_chunks(docnames, nproc)
+
+        chunks_complete = 0
+        summary = bold(__('reading sources... '))
+
         def merge(docs: List[str], otherenv: bytes) -> None:
+            # Update progress message for each chunk completed
+            nonlocal chunks_complete
+            chunks_complete += 1
+            log_status_message(display_chunk(docs), summary, 'purple', chunks_complete,
```
tk0miya, line 463: How about adding a new option to control the timing of messages (ex. before, after, both) to `status_iterator()` instead? I think the concept of `status_iterator()` is separating progress messages from the main code. But this refactoring makes them coupled. It would be nice if we can keep them less coupled.

peterbell10: Your suggestion doesn't relate to the issue because the for loop doesn't wait until the tasks are completed. The entire loop can finish before even a single task has.

peterbell10: @tk0miya in case I wasn't clear, the `status_iterator` here only measures the progress of _submitting_ tasks into the queue with  `add_task`. `make_chunks` can create more than `nproc` chunks, and `add_task` doesn't block when there are more than `nproc` tasks submitted. It just adds them to a queue ready to be run. So, there is no connection whatsoever between the `status_iterator` progress and actual progress of building the documentation. i.e. the progress messages are pretty much useless.

For writing, there is _some_ serialised work executed inside of the for loop which is measured by the `status_iterator`. However, it's still true that none of the parallelised work is represented in the progress messages. So there can still be huge delays between 100% progress and actually finishing the build step.

rgommers: > separating progress messages from the main code

The logic in the PR looks good to me. Progress messages cannot be decoupled completely from doing the actual work - there has to be some signal that says "chunk is done, please send out a new progress message".

@tk0miya is there a chance this can be merged in the near future? This will be quite helpful for SciPy. We have a very heavy doc build (~20 minutes) that sometimes times out on CI because no output is received.

If there's some extra validation or other activity that we can do to help move this forward, please let us know!

peterbell10:
@tk0miya is it possibly this PR is just getting bumped each release because it's mislabelled as a "refactor". This isn't a refactor and does meaningfully fix the bug of progress messages not actually tracking the number of build jobs completed.

danwos:
Would be great to have this.
We need to use the parallel build for all projects, as they are quite huge, and build time is > 30min with 4 cores.

In some scenarios, there is no update in the console for 20min, which let some developers think that the build has stopped and they stop the process :(

danwos:
@tk0miya Anything I can do here to finalize it?

tk0miya:
Sorry for the very very very late response. I posted #10454 as a candidate of the yet-another implementation for this problem. It does not touch the implementation of `status_iterator()`. I think it keeps main code simple. Could you let me know your opinion. Thanks,

tk0miya:
Finally, I determined to merge #10454 instead of this. I believe it is based on this PR. So I must say thank you for all :-)

rgommers:
Thanks a lot @tk0miya, great to have a fix for this!

PR #9097, opened 2021-04-13
Remove unnecesary sleep when submitting parallel tasks

peterbell10:
### Feature or Bugfix
- Refactoring

### Purpose
`ParallelTasks.add_task` calls `_join_one` to poll for finished workers and possibly start the just-submitted task. However, it will also trigger the `time.sleep` call if no sub-processes are completed which unnecessarily slows down the launching of workers. Instead, I raise the `time.sleep` call into `join` so that `add_task` never starts a sleep.

The sleep is only 20 ms per task so this doesn't make a huge difference.

### DIFF
--- a/sphinx/util/parallel.py
+++ b/sphinx/util/parallel.py
@@ -100,12 +100,13 @@ class ParallelTasks:
                                        args=(psend, task_func, arg))
         self._procs[tid] = proc
         self._precvsWaiting[tid] = precv
-        self._join_one()
+        _ = self._join_one()
 
     def join(self) -> None:
         try:
             while self._pworking:
-                self._join_one()
+                if not self._join_one():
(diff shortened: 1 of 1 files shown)


sphinx/util/parallel.py
```diff
@@ -100,12 +100,13 @@ def add_task(self, task_func: Callable, arg: Any = None, result_func: Callable =
                                        args=(psend, task_func, arg))
         self._procs[tid] = proc
         self._precvsWaiting[tid] = precv
-        self._join_one()
+        _ = self._join_one()
```
tk0miya: This assignment is not needed, right?

tk0miya (commented):
LGTM with nits.

tk0miya:
Thank you for the quick update! Merging.

PR #9101, opened 2021-04-15
Fix `user-selection` in Safari and Internet Explorer

V0lantis:
Subject: On safari and Internet explorer, the code's selection wasn't working properly.
Before (Safari) : 
Capture d’écran 2021-04-15 à 12 13 50

After (Still on Safari)
Capture d’écran 2021-04-15 à 12 13 57

This is due to https://stackoverflow.com/a/34372191

### Relates
- #9098

PR #9120, opened 2021-04-20
Feature: 

V0lantis:
Remove `div.doctest` in `basic.css_t` so that the selecting feature in snippets code
are not just available for doctests snippets. Basically, I want the following feature : 

Capture d’écran 2021-04-20 à 11 50 36

which is available for every doctest, to be also available for basic snippet code. 
For example, in the actual `[Sphinx doc](https://www.sphinx-doc.org/en/master/usage/installation.html#installation-from-source)`, when we select a code's snippet, it also selects the `$`sign, which is annoying for multi line snippets : 
Capture d’écran 2021-04-20 à 11 42 39

With the modification, we can now do : 

Capture d’écran 2021-04-20 à 11 43 54

This PR is following #9101. I ask for discussion here, because I don't know what was intended `span.gp` at the first place. From my research, it has been introduced by @tk0miya, in #6727

### DIFF
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -819,7 +819,7 @@ div.code-block-caption code {
 
 table.highlighttable td.linenos,
 span.linenos,
-div.doctest > div.highlight span.gp {  /* gp: Generic.Prompt */
+div.highlight span.gp {  /* gp: Generic.Prompt */
   user-select: none;
   -webkit-user-select: text; /* Safari fallback only */
   -webkit-user-select: none; /* Chrome/Safari */


tk0miya:
> I ask for discussion here, because I don't know what was intended span.gp at the first place. 

As the comment in the CSS file says, `span.gp` means "Generic.Prompt". They're considered as prompt by pygments; a highlighting library.

Note: What I commented in https://github.com/sphinx-doc/sphinx/pull/9101#discussion_r615217904 is here:

>Indeed, it will be helpful for almost of users. But I can't say ALL of the people want to do that. So this would not be changed, I think.

I'm still debating this will not introduce troubles. I'd like to hear comments from @shimizukawa .

shimizukawa:
I think the prompt will get in the way if it is copied when we want to run the procedure as it is in the sample. The only situations I can think of where we would want to copy text with prompts are: 1) to tell people in chat, etc., or 2) to transcribe the same procedure into a document (including doctest).
In this sense, there are very few situations where we would want to copy and use either the `$` symbol or the `>>>` symbol.

(As a side note, the official Python documentation includes a feature to show/hide the `>>>` symbol, but I couldn't find a same feature to switch the `$` symbol.)

tk0miya:
@shimizukawa Thank you for your comment.

Okay, let's merge this and wait for user's reaction.

tk0miya:
Just merged. Thank you for your contribution!

PR #9112, opened 2021-04-18
Recommend MyST-Parser for Markdown usage

astrojuanlu:
Subject: Recommend MyST-Parser for Markdown usage, instad of recommonmark.

### Feature or Bugfix

Not sure if I'd consider this a feature, a bugfix or a refactoring, but if I had to choose I guess it would be a bugfix.

### Purpose
The recommonmark maintainers have proposed to [sunset recommonmark in favor of MyST-Parser](https://github.com/readthedocs/recommonmark/issues/221) since the latter enjoys more community support, is more flexible, and has a brighter future. A few projects have already migrated since the announcement was made, and so far nobody has found unexpected problems. Both projects are configured as Sphinx extensions and work in a very similar way.

_Disclaimer: I started working recently at Read the Docs as Developer Advocate, and my mission is to help projects improve their documentation. In particular, Sphinx is a core piece of what we do ❤️ and I would like to make more contributions in the future!_

### Relates
https://github.com/readthedocs/recommonmark/issues/221

PR #9124, opened 2021-04-21
Recommend MyST-Parser for Markdown usage (backported)

astrojuanlu:
This is exactly the same as #9112, backported for the upcoming 4.0.0 version. Given that the recommonmark deprecation is coming soon, it would be nice to display this in the 4.0 version already. Sorry for the unwanted duplication.


tk0miya:
This has already been merged into the master now. So it's published at sphinx-doc.org: https://www.sphinx-doc.org/en/master/usage/markdown.html. Is this change needed?

astrojuanlu:
Oh, you're right! No, I don't think it's necessary. Closing

PR #4022, opened 2017-08-23
#1618 make search results reader friendly

TimKam:
Subject: Make search results reader friendly
Fixes #1618 

### Feature or Bugfix
- Bugfix?

### Purpose
See my comment on the related issue.
There's one fairly simply way to improve the readability of the search results without adding additional build steps or output files. Currently, the search displays results snippets by requesting the corresponding source files from the server/local file system and extracting the text from them. It's possible to adjust this functionality so that it **requests the HTML files instead of the source files**.
### Detail
- Then, it's fairly simple to extract the text from the HTML during client/browser runtime. Of course, this increases load sizes and computing time a bit, but I don't think the change in performance is significant.
- This would make the [pretty search results](https://pypi.python.org/pypi/sphinxprettysearchresults) extension obsolete.
- Can this break anything? I don't think anyone **expects** the search results to display the source snippets. We could of course add a configuration option and only activate the changes if the option is set to ``True``.

### Relates
- #1618

PR #9129, opened 2021-04-22
HTML: Fix search unnecessarily requiring source files

Blendify:
Subject: Fixes search summeries not showing when `html_copy_source = False`

### Purpose
This seems to have been a mistake with #4022 the ajax call functions correctly without the source files being included in the build (they are never used).

I have tested this out on several themes and now everything works correctly with `html_copy_source = False`

### Relates

#4022


Blendify:
I am not sure how to rebase this for 4.0.x but it would be nice if this could go into 4.0.x

tk0miya:
LGTM. Could you rebase this into the 4.x branch?

tk0miya:
Merged manually.

>I am not sure how to rebase this for 4.0.x but it would be nice if this could go into 4.0.x

You can do it with `git rebase` command. In this case, you can do it with `git rebase HEAD~1 --onto 4.x` (This rebases the last commit (`~1` means one commit) to the `4.x` branch.

tk0miya:
Thank you for your contribution as always :-)

PR #9131, opened 2021-04-24
Ignore nitpick warnings with regular expressions using `nitpick_ignore_regex`

ruro:
Subject: Allow regex patterns in nitpick_ignore.

Edit: this PR originally proposed adding glob-like pattern matching, but after some discussion, it was changed to use regex patterns instead.

### Feature or Bugfix
- Feature

### Purpose

Allow the user to specify regex patterns in the `nitpick_ignore`. Currently, the user is forced to enumerate all the `nitpicks` that should be ignored, even if they follow a predictable pattern. So after some time, the `nitpick_ignore` array might end up looking something like this:

```python
nitpick_ignore = [
    ('py:mod', 'foo'),
    ('py:class', 'foo.Something'),
    ('py:mod', 'foo.bar'),
    ('py:class', 'foo.bar.Baz'),
    ('py:func', 'foo.bar.boo.'),
    # ... etc ...
]
```

With this PR, the user can instead define `nitpick_ignore_regex = [(r'py:.*', r'foo\..*')]`.

### Compatibility

<s>I reuse the `sphinx.util.matching` mechanism to match glob patterns. To be completely honest, I am not 100% sure if it is okay to match the `type` and `target` of the nitpick warnings like paths (i.e. are there any cases where `type` or `target` may contain the symbols `/?[]^*` and **not** be a path). Based on the use cases of `nitpick_ignore`, that I've personally seen, I don't expect this to be an issue, but this is **technically** a breaking change if somebody is currently matching something like `[woops]!`.</s>

Edit: no longer applicable, see further discussion in the comments.

### Tests

I didn't find any tests for the current behaviour of `nitpick_ignore`, so I've implemented the tests both for the current behaviour and the new regex feature. I did this by finding existing tests, which already use `nitpicky=True` and `assert` that there **should** be some errors. I copied these tests, added the `nitpick_ignore_regex` option and `assert`ed that the errors are gone in that case.

### DIFF
--- a/sphinx/transforms/post_transforms/__init__.py
+++ b/sphinx/transforms/post_transforms/__init__.py
@@ -8,6 +8,7 @@
     :license: BSD, see LICENSE for details.
 """
 
+import re
 from typing import Any, Dict, List, Optional, Tuple, Type, cast
 
 from docutils import nodes
@@ -171,13 +172,26 @@ class ReferencesResolver(SphinxPostTransform):
         warn = node.get('refwarn')
         if self.config.nitpicky:
             warn = True
+            dtype = '%s:%s' % (domain.name, typ) if domain else typ
             if self.config.nitpick_ignore:
-                dtype = '%s:%s' % (domain.name, typ) if domain else typ
                 if (dtype, target) in self.config.nitpick_ignore:
                     warn = False
                 # for "std" types also try without domain name
                 if (not domain or domain.name == 'std') and \
                    (typ, target) in self.config.nitpick_ignore:
+                    warn = False
+            if self.config.nitpick_ignore_regex:
+                def matches_ignore(entry_type: str, entry_target: str) -> bool:
+                    for ignore_type, ignore_target in self.config.nitpick_ignore_regex:
+                        if re.fullmatch(ignore_type, entry_type) and \
+                           re.fullmatch(ignore_target, entry_target):
+                            return True
+                    return False
+                if matches_ignore(dtype, target):
+                    warn = False
+                # for "std" types also try without domain name
+                if (not domain or domain.name == 'std') and \
+                   matches_ignore(typ, target):
                     warn = False
         if not warn:
             return
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -419,6 +419,20 @@ General configuration
 
    .. versionadded:: 1.1
 
+.. confval:: nitpick_ignore_regex
+
+   An extended version of :confval:`nitpick_ignore`, which instead interprets
+   the ``type`` and ``target`` strings as regular expressions. Note, that the
+   regular expression must match the whole string (as if the ``^`` and ``$``
+   markers were inserted).
+
+   For example, ``(r'py:.*', r'foo.*bar\.B.*')`` will ignore nitpicky warnings
+   for all python entities that start with ``'foo'`` and have ``'bar.B'`` in
+   them, such as ``('py:const', 'foo_package.bar.BAZ_VALUE')`` or
+   ``('py:class', 'food.bar.Barman')``.
+
+   .. versionadded:: 4.1
+
 .. confval:: numfig
 
    If true, figures, tables and code-blocks are automatically numbered if they
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -131,6 +131,7 @@ class Config:
         'manpages_url': (None, 'env', []),
         'nitpicky': (False, None, []),
         'nitpick_ignore': ([], None, []),
+        'nitpick_ignore_regex': ([], None, []),
         'numfig': (False, 'env', []),
         'numfig_secnum_depth': (1, 'env', []),
         'numfig_format': ({}, 'env', []),  # will be initialized in init_numfig_format()
--- a/tests/roots/test-nitpicky-warnings/conf.py
+++ b/tests/roots/test-nitpicky-warnings/conf.py
@@ -1,0 +1,1 @@
+nitpicky = True
--- a/tests/roots/test-nitpicky-warnings/index.rst
+++ b/tests/roots/test-nitpicky-warnings/index.rst
@@ -1,0 +1,7 @@
+test-nitpicky-warnings
+======================
+
+:py:const:`prefix.anything.postfix`
+:py:class:`prefix.anything`
+:py:class:`anything.postfix`
+:js:class:`prefix.anything.postfix`
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -311,3 +311,34 @@ def test_check_enum_for_list_failed(logger):
     config.init_values()
     check_confval_types(None, config)
     assert logger.warning.called
+
+
+nitpick_warnings = [
+    "WARNING: py:const reference target not found: prefix.anything.postfix",
+    "WARNING: py:class reference target not found: prefix.anything",
+    "WARNING: py:class reference target not found: anything.postfix",
+    "WARNING: js:class reference target not found: prefix.anything.postfix",
+]
+
+
+@pytest.mark.sphinx(testroot='nitpicky-warnings')
+def test_nitpick_base(app, status, warning):
+    app.builder.build_all()
+
+    warning = warning.getvalue().strip().split('\n')
+    assert len(warning) == len(nitpick_warnings)
+    for actual, expected in zip(warning, nitpick_warnings):
+        assert expected in actual
+
+
+@pytest.mark.sphinx(testroot='nitpicky-warnings', confoverrides={
+    'nitpick_ignore': [
+        ('py:const', 'prefix.anything.postfix'),
+        ('py:class', 'prefix.anything'),
+        ('py:class', 'anything.postfix'),
+        ('js:class', 'prefix.anything.postfix'),
+    ],
+})
+def test_nitpick_ignore(app, status, warning):
+    app.builder.build_all()
+    assert not len(warning.getvalue().strip())
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -342,3 +342,46 @@ def test_nitpick_base(app, status, warning):
 def test_nitpick_ignore(app, status, warning):
     app.builder.build_all()
     assert not len(warning.getvalue().strip())
+
+
+@pytest.mark.sphinx(testroot='nitpicky-warnings', confoverrides={
+    'nitpick_ignore_regex': [
+        (r'py:.*', r'.*postfix'),
+        (r'.*:class', r'prefix.*'),
+    ]
+})
+def test_nitpick_ignore_regex1(app, status, warning):
+    app.builder.build_all()
+    assert not len(warning.getvalue().strip())
+
+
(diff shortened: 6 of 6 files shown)


tk0miya:
>I reuse the sphinx.util.matching mechanism to match glob patterns. To be completely honest, I am not 100% sure if it is okay to match the type and target of the nitpick warnings like paths (i.e. are there any cases where type or target may contain the symbols /?[]^* and not be a path).

Good point. `sphinx.util.matching` is an extended fnmatch module. So it's main purpose is to match filename and pattern. So it's not good to compare `nitpick_ignore` and warning types. Some languages would contains many kinds of symbols as you mentioned (like a pointer (`int*`) in C/C++).

So I think using regexp is better for this case. What do you think?

tests/test_domain_c.py
```diff
@@ -591,6 +591,33 @@ def test_build_domain_c_anon_dup_decl(app, status, warning):
     assert "WARNING: c:identifier reference target not found: @b" in ws[1]
 
 
+@pytest.mark.sphinx(testroot='domain-c', confoverrides={
+    'nitpicky': True,
+    'nitpick_ignore': [
```
tk0miya: Thank you for adding testcases. LGTM. But it would be better if they will move to `tests/test_config.py` because `nitpicky_warning` is a common feature, not only for C/C++. Could you move them, please?

ruro: Do you want me to just move the tests to a different file or to also create a separate `testroot` instead of reusing `domain-c` and `domain-cpp`? I haven't looked into how the whole test root setup works yet.

ruro:
> Some languages would contains many kinds of symbols as you mentioned (like a pointer (`int*`) in C/C++).

Nice catch, I totally forgot about pointer/array types in C-like languages.

> So I think using regexp is better for this case. What do you think?

Hmm, I've thought about this a bit more, and it seems that even if we use regex for this (which we probably should), it would still break some existing uses of `nitpick_ignore`. For example, `int*` will no longer match the literal `int*`, but instead will match `in`, `int`, `intt`, `inttt` etc. Which is probably too much of a breaking change.

I see 3 possible solutions for this:

1) Use a separate configuration variable
   ```python
   nitpick_ignore = [
       ('old behaviour', 'int*'),
       ...
   ]
   nitpick_ignore_regex = [
       (r'py:.*', r'foo\..*'),
       ...
   ]
   ```
2) Use `nitpick_ignore` like originally proposed, but allow the user to toggle this behaviour
   ```python
   nitpick_ignore = [
       (r'new behaviour', r'int\*'),
       (r'py:.*', r'foo\..*'),
       ...
   ]
   nitpick_ignore_regex = True
   ```
3) Use `nitpick_ignore`, instead of regex strings, make the user `re.compile` the regular expressions on their own and check `isinstance(v, re.Pattern)` in the implementation
   ```python
   nitpick_ignore = [
       ('old behaviour', 'int*'),
       (re.compile(r'py:.*'), re.compile(r'foo\..*')),
       ...
   ]
   ```

ruro:
@tk0miya ping? Do you have a preference for any of the 3 options?

jakobandersen:
For my self I have an ad-hoc extension where I define different kind of patterns to ignore. Getting rid of it and use build-in ignore functionality would be nice, and I think this PR might be enough.
I think I like option (1) best, it allows for straight-forward incremental adoption.
(2) might be annoying if you need a single pattern, but have a lot of exact strings that need regex escaping.
(3) looks a bit clunky, but is in another sense more elegant than (1), so I'm not against it as such. It still allows for incremental adoption.

One thing which is not explicitly clear, but should be in the documentation in the end: is the regex matched directly, or with begin/end anchors? E.g., for a pattern ``a.*b``, would it match ``sabt``? From your example it looks like each regex will be used as if ``^`` and ``$`` are prepended and appended. E.g., the pattern ``a.*b`` means the target string must match ``^a.*b$``.

tk0miya:
+1 for the 1st idea.

* 1) Simple.
* 2) Difficult to switch normal one to regexp one (escaping needed).
* 3) python knowledge is needed to configure it. It's not matched to the static configuration file: https://github.com/sphinx-doc/sphinx/issues/9040

Additionally, @shimizukawa also votes for the 1st idea. I talked with him at the Hack-a-thon event today.

>One thing which is not explicitly clear, but should be in the documentation in the end: is the regex matched directly, or with begin/end anchors? E.g., for a pattern a.*b, would it match sabt? From your example it looks like each regex will be used as if ^ and $ are prepended and appended. E.g., the pattern a.*b means the target string must match ^a.*b$.

I think it's better to match to the beginning of the types as `re.match()` does.

jakobandersen:
> I think it's better to match to the beginning of the types as `re.match()` does.

So that would mean that ``ns::`` and ``ns::.*`` are equivalent and both means ``^ns::.*$``, right? That is,  the user-specified regex is conceptually put between ``^`` and ``.*$`` for fully matching the string,
If so, agreed, that sounds good to me.

tk0miya:
Wait, it's not good to compare like `re.match()` does.
```
nitpick_ignore_regex = [(r'py::class', r'foo')]
```
This should not match to `subtype=foobar`. It should exactly match to `subtype=foo`. So it's better to evaluate it as `^foo$` as @jakobandersen said. Good point!

jakobandersen:
> Wait, it's not good to compare like `re.match()` does.
> 
> ```
> nitpick_ignore_regex = [(r'py::class', r'foo')]
> ```
> 
> This should not match to `subtype=foobar`. It should exactly match to `subtype=foo`. So it's better to evaluate it as `^foo$` as @jakobandersen said. Good point!

Having it symmetric with respect to start and end anchors is probably easier to understand for those not used to the Python ``re.match()``, but just for good measure: in this case one could be the entry in the ordinary ``nitpick_ignore`` and prevent matching ``foobar``.
As an argument against putting any anchors in implicitly: ``grep foo`` and  ``sed s/foo/f/`` will match ``foo`` anywhere in the line, so if the user is used to those tools, it might be surprising that the regex pattern is not interpreted as is. By that I suggest taking the regex as is. So ``foo`` _would_ match ``foobar``. The user can then insert anchors for start, end, word boundary, etc. as they wish.
For non-Unix-like command line users, what would be least surprising?
Do we have other configuration variables with regexes? and what is done there?

ruro:
Thank you everyone for the feedback. I've implemented option 1 - `nitpick_ignore_regex`.

Regarding how should the regular expression be matched, if I understood correctly, the 3 options were

- `re.match` - matches "glued" to the beginning of the string
- `re.search` - matches anywhere in the string
- `re.fullmatch` - matches the whole string exactly (beginning to end)

My personal opinion is that the behaviour of `re.match` is indeed not very intuitive and doesn't follow the principle of least astonishment in this case. So for me, the choice was between `re.search` and `re.fullmatch`. In the end, I've implemented `re.fullmatch`, but I am open to further <s>bikeshedding</s> discussion of this topic.

My motivation for choosing `fullmatch` instead of `search` is:

1) A plain string (with no special regex characters) would "match" the same way as in `nitpick_ignore`. This makes it *easier* to switch to `nitpick_ignore_regex` for all nitpicks instead of keeping 2 separate config variables.

2) I expect, that some users might need the "advanced" regex matching only for `type` or only for `target`. This might lead them to write a pattern like `('.*', 'target')` or `('type', '.*')`. In both cases `re.search` would unexpectedly (imo) treat the "plain" strings as `.*target.*` and `.*type.*`.

3) `fullmatch` is stricter (it matches a subset of strings, compared to `search`).
    - If the user expected the regular expression to work like `re.search`, but the implementation uses `re.fullmatch`, the nitpicky warning won't disappear, so it would be immediately obvious to the user, that he did something wrong. 

    - If the user expected the regular expression to work like `re.fullmatch`, but the implementation uses `re.search`, then the nitpicky warning that he tried to silence will disappear, but the user would also have accidentally silenced all the warnings that contain the expression he used as a substring, which is not what the user intended.

Just in case, I've added a description of this behaviour to the documentation and a separate unittest for it.

jakobandersen:
Good point about ``fullmatch`` vs. ``search``, and the added documentation makes it clear.

tests/test_domain_c.py
tk0miya: I'd like to separate the file only. "test_domain_c" is not good for nitpicky_warning. It's okay to reuse testroots for domain-c and domain-cpp from other files.
I think no reason to test nitpicky_warning for each domain.

sphinx/transforms/post_transforms/__init__.py
```diff
@@ -171,14 +172,27 @@ def warn_missing_reference(self, refdoc: str, typ: str, target: str,
...
             if self.config.nitpick_ignore:
-                dtype = '%s:%s' % (domain.name, typ) if domain else typ
                 if (dtype, target) in self.config.nitpick_ignore:
                     warn = False
                 # for "std" types also try without domain name
                 if (not domain or domain.name == 'std') and \
                    (typ, target) in self.config.nitpick_ignore:
                     warn = False
+            if self.config.nitpick_ignore_regex:
+                def matches_ignore(ent_typ: str, ent_tgt: str) -> bool:
```
tk0miya: What the "ent" means? entry? I think names should not be abbreviated. "ign" is also.

tk0miya (changes requested):
LGTM with nits.

As I commented above, I prefer the `re.fullmatch()` approach in this case. +1 for this as updated.

sphinx/transforms/post_transforms/__init__.py
ruro: Sure, I changed it to be the longer versions.

About the tests, I am currently using the `filter_warnings` function to check the results. Do you think I should also copy it to `tests/test_config.py`?

Also by "I think no reason to test nitpicky_warning for each domain." do you mean that testing with both `cpp` and `c` is too much or that 2 domains is enough?

jakobandersen: I'm not sure it is good to reuse those test roots unless the tests are in the files. I fear the nitpick tests will falsely pass later if some refactoring is done in the domains.
How about making a new test root with a single ``index.rst`` file with specific tests instead? Then you won't need the ``filter_warnings`` function (unless you make multiple files as well).

tk0miya: The best way is to add a new testroot. It would be nice if all test cases are separated. But it's okay to share the example if it's tough. So the better way is to reuse the testroot and copy the function to `test_config.py`.

>Also by "I think no reason to test nitpicky_warning for each domain." do you mean that testing with both cpp and c is too much or that 2 domains is enough?

Yes. nitpick_ignore_regex is not related to the specific domain. So no reason to test both C and C++. I think no advantage. Is there any purpose to test with multiple domains?

ruro: I've added a separate test root and added the updated tests to `test_config.py`.

ruro:
@tk0miya have I addressed all of your comments?

tk0miya (approved):
Thank you for your work. LGTM!

tk0miya:
Merged!

PR #9133, opened 2021-04-24
Build docs against Sphinx 4.0b1 & docutils 0.17

ericholscher:


tk0miya:
I think this is needed only for testing. It's not needed to merge this, right?

ericholscher:
It depends. I think it might be a good idea to have the Sphinx docs building on the latest released beta of Sphinx to help test for any regressions prior to release. I actually don't know what version of Sphinx they're currently using, but either the latest pre-release would make sense, or the actual commit being built.

tk0miya:
I think it's enough to test it just before the beta release. But, I'm okay if you'd like to pin it here also.

PR #9157, opened 2021-05-02
Fix #8872: autodoc: stacked singledispatches are wrongly rendered

### LINKED ISSUE #8872: stacked singledispatches use wrong type annotation
**Describe the bug**

When multiple `@functools.singledispatch` decorators are stacked (as documented by the stdlib docs at https://docs.python.org/3/library/functools.html#functools.singledispatch), the type annotations reported by sphinx are incorrect.

**To Reproduce**
```
cat >project.py <<EOF
from functools import singledispatch

@singledispatch
def foo(x): "something"

@foo.register(int)
@foo.register(float)
def _(x): "a number"
EOF
sphinx-apidoc . -o . -F -A me -V 0.0
PYTHONPATH=. make html
```
then open _build/html/project.html: you get
```
project.foo(x)                 [source]
project.foo(x: float)
project.foo(x: float)
something
```

**Expected behavior**
The correct overloads are
```
project.foo(x)
project.foo(x: int)
project.foo(x: float)
```

**Your project**
N/A

**Screenshots**
N/A

**Environment info**
- OS: Arch Linux
- Python version: 3.9.1
- Sphinx version: 3.4.3
- Extra tools: N/A

**Additional context**
N/A

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- When multiple singledispatch decorators are stacked, the first typehints
are copied to the subsequent definitions unexpectedly.
- Now autodoc generates a dummy function not to affect typehints to
subsequent functions.
- refs: #8872

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1328,12 +1328,12 @@ class FunctionDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # typ
                 if typ is object:
                     pass  # default implementation. skipped.
                 else:
-                    self.annotate_to_first_argument(func, typ)
-
-                    documenter = FunctionDocumenter(self.directive, '')
-                    documenter.object = func
-                    documenter.objpath = [None]
-                    sigs.append(documenter.format_signature())
+                    dispatchfunc = self.annotate_to_first_argument(func, typ)
+                    if dispatchfunc:
+                        documenter = FunctionDocumenter(self.directive, '')
+                        documenter.object = dispatchfunc
+                        documenter.objpath = [None]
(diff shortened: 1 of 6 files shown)


sphinx/ext/autodoc/__init__.py
```diff
@@ -1320,10 +1320,10 @@ def format_signature(self, **kwargs: Any) -> str:
                 if typ is object:
                     pass  # default implementation. skipped.
                 else:
-                    self.annotate_to_first_argument(func, typ)
+                    dispatchfunc = self.annotate_to_first_argument(func, typ)
```
tk0miya, line 1331: `dispatchfunc` is Optional. So it should be null-checked.

anntzer:
Thanks!

PR #9159, opened 2021-05-02
Fix #8588: autodoc_type_aliases does not support dotted name

### LINKED ISSUE #8588: autodoc_type_aliases does not support dotted name
**Describe the bug**
autodoc_type_aliases does not support dotted name

**To Reproduce**

```
# helloworld.py
from __future__ import annotations  # important!
import struct

def consume_struct(_: struct.Struct) -> None:
    pass
```
```
# conf.py
autodoc_type_aliases = {
    'struct.Struct': 'struct.Struct',
}
```

The autodoc_type_aliases entry does not work as expected.
(refs: #8315)

**Expected behavior**
Supporting

**Your project**
No

**Screenshots**
No

**Environment info**
- OS: Mac
- Python version: 3.8.2
- Sphinx version: HEAD of 3.x
- Extra tools: No

**Additional context**
No

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #8588

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 * #8107: autodoc: Add ``class-doc-from`` option to :rst:dir:`autoclass`
   directive to control the content of the specific class like
   :confval:`autoclass_content`
+* #8588: autodoc: :confval:`autodoc_type_aliases` now supports dotted name. It
+  allows you to define an alias for a class with module name like
+  ``foo.bar.BazClass``
 * #9129: html search: Show search summaries when html_copy_source = False
 * #9120: html theme: Eliminate prompt characters of code-block from copyable
   text
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -18,8 +18,10 @@ import types
 import typing
 import warnings
 from functools import partial, partialmethod
+from importlib import import_module
 from inspect import Parameter, isclass, ismethod, ismethoddescriptor, ismodule  # NOQA
 from io import StringIO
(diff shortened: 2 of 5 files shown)


CHANGES
```diff
@@ -22,6 +22,8 @@ Features added
 Bugs fixed
 ----------
 
+* #8588: autodoc: autodoc_type_aliases does not support dotted name
```
tk0miya: How about describing this as an improvement? It's easier to understand the change.
```suggestion
* #8588: autodoc: autodoc_type_aliases now supports dotted name. It allows you to define an alias for a class with module name like ``foo.bar.BazClass``.
```

sphinx/util/inspect.py
```diff
@@ -501,6 +503,63 @@ def __repr__(self) -> str:
...
+
+    def __call__(self) -> None:
+        # Dummy method to imitate special typing classes
+        pass
+
+    def __eq__(self, other: Any) -> bool:
+        return self.name == other
+
+
+class TypeAliasNamespace(Dict[str, Any]):
```
tk0miya: It might be better to separate a "namespace" and a "module". They are very similar, but different things.

* Namespace
  * Accessed via `__getitem__` API
  * Not having a "current module"
  * Returns a module or forwardref.
  * Raises KeyError if not found
* Module
  * Accessed via `__getattr__` API
  * Has its "current module" name.
  * Returns a sub-module or forwardref, or real object
  * Never raises KeyError.

Issue #1336, opened 2015-01-03
reST container directive for latex output


shimizukawa:
Hi,

reST container directive does not work for latex output.

A patch to solve the issue is attached.

Regards,
Camille

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1336
- Originally reported by: Camille Gouttebroze
- Originally created at: 2013-12-31T11:36:10.088

shimizukawa:
_From Florian Oppermann on 2014-02-18 13:35:53+00:00_

I added another three lines of code for generating labels for the :name: option. Unfortunately I don't know how to convert the node.get('names') to a proper reference identifier. Therefore I used node.get('ids') which contains the parsed name but also contains identifiers from ordinary labels. This leads to multiple labels when a container gets a label not via :name:. It was the best solution I could find quickly.

```
#!python

ids = node.get('ids', [])
for Id in ids:
  self.body.append( '\n' + self.hypertarget(Id) )
```

tk0miya:
I salvaged the patch file from bitbucket: 
[latexContainer.patch](https://files.zenhub.io/56dbf51d7b8f2ce302239543)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

kenmorse:
What is the status of this issue? Would be great if the container directive worked for latex (PDF) as well as for HTML output.

tk0miya:
I salvaged the patch file again: [latexContainer.patch.gz](https://github.com/sphinx-doc/sphinx/files/4943953/latexContainer.patch.gz)

This expands the container node to `\begin{%s}` and `\end{%}` macro. I'm afraid that this breaks PDF build for users who uses container directives suddenly. To apply, or not to apply. That is a question.

kenmorse:
@tk0miya - First, do no harm?

tk0miya:
If applied, you MUST add a LaTeX macro for container classes if you're using. If not, LaTeX failed to build PDF.

dham:
I think #9166 provides a solution to this without the problem identified by @tk0miya above.

PR #9166, opened 2021-05-03
Support containers in LaTeX output.

dham:
- Feature

### Purpose
Support containers in the LaTeX writer by outputting the start and end of a LaTeX group, wrapping the beginning and end of a LaTeX environment. The class name(s) are passed as an argument to the
environment.

### Detail
Container directives currently result in no output in the LaTeX writer, which is unfortunate because they are a very useful mechanism for adding an essentially arbitrary formatting hook in a document. A previous solution was proposed in #1336 which would have emitted an environment with the name of the directive class. This was not merged because emitting previously undefined environment names will break the `latexpdf` build unless the user also defines the environment, for example in `latex_elements`.

This proposed solution avoids that problem by always emitting a `sphinxcontainer` environment. The class name(s) are passed into the environment as an argument. The environment emitted is wrapped in `\bgroup` and `\egroup` so that it also constitutes a LaTeX group. It is argued that this best mirrors the html behaviour and is therefore least likely to surprise a user.

A definition of the `sphinxcontainer` environment is added in a style file input by `sphinx.sty`, so inserting container directives will not break the `latexpdf` build. Should the user wish to insert LaTeX formatting based on a container directive, then they can redefine `sphinxcontainer` in `latex_elements["preamble"]`. For example, if an `.rst` file contains:
```rst
.. container:: red

   red text

.. container:: blue

   black text

more black text.
```
and `conf.py` contains:
```python3
latex_elements = {
    "preamble": r"""
    \usepackage{etoolbox}
    \renewenvironment{sphinxcontainer}[1]{
        \ifstrequal{#1}{red}{\color{red}}{}
    }{}
    """
}
```
then `red text` will be rendered in red, and all the other text will be displayed in black.

### Relates
- #1336

### Limitations

The [documentation for container](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives) links straight through to the [docutils documentation](https://docutils.sourceforge.io/docs/ref/rst/directives.html#container), so I don't know where I should document this change.

Life is a bit more complicated for the user if they need to customise the end of the environment depending on the container class, or if the container specifies multiple classes. However, both of these are really LaTeX limitations. I suggest that this small change provides all the relevant information through to LaTeX, and that fixing LaTeX's limitations in this matter should not be seen as Sphinx's problem.


tk0miya:
In docutils, the container node will be rendered as `\DUclass{classname}{ content }` macro.
https://repo.or.cz/docutils.git/blob/HEAD:/docutils/docutils/writers/latex2e/__init__.py#l1560
https://repo.or.cz/docutils.git/blob/HEAD:/docutils/docutils/writers/latex2e/docutils.sty#l31

It would be better to follow the concept.

@jfbu Do you have opinion for this?

dham:
> In docutils, the container node will be rendered as `\DUclass{classname}{ content }` macro.
> https://repo.or.cz/docutils.git/blob/HEAD:/docutils/docutils/writers/latex2e/__init__.py#l1560
> https://repo.or.cz/docutils.git/blob/HEAD:/docutils/docutils/writers/latex2e/docutils.sty#l31
> 
> It would be better to follow the concept.

I'm a little confused. I agree that a macro would be better than an environment, but the links you provide seem to define a `DUclass` environment and insert `\begin{DUclass}{classname}` and `\end{DUclass}`.

I am completely happy to refactor this to insert a macro as suggested, but that doesn't actually seem consistent with what docutils are doing. Please let me know what you would prefer.

tk0miya:
Ah, sorry. It's my bad. I'd like to mention the DUclass "environment".

dham:
OK, so I am a little unclear as to what you are asking for. Would you like me to rename the environment as `DUclass`? I can do that, happily, though it would not be very consistent with the other LaTeX names that sphinx uses.

tk0miya:
My large concern is generating an environment for each class, not for the whole of the node. For example, a container node having "red bold" will be converted to `\begin{DUclass}{red}\begin{DUclass}{bold} content \end{DUclass}{bold}\end{DUclass}{red}`. And users can customize the behavior for red and bold individually. It will resolve a limitation you mentioned.

>Life is a bit more complicated for the user if they need to customise the end of the environment depending on the container class, or if the container specifies multiple classes. However, both of these are really LaTeX limitations.

About naming, I don't have opinion for this. So I think either '\DUclass` or `\sphinxclass` (?) is okay to me.

dham:
I've changed the code to one environment per class as requested, and renamed that environment to `sphinxcode`. Note that it's not possible to pass parameters to `\end` so the problem of not knowing which container you are ending persists. There are a few approaches which could be taken:

1.  I currently open and close a group around each environment. This makes declarations like `\color` end in the right place.
2. Declaring a macro containing the class name at the start is possible, but basically requires a stack. There is a stack implementation on ctan, [fifo-stack](https://ctan.org/pkg/fifo-stack?lang=en), but we probably don't want to add that random LaTeX dependency.
3. The [xparse](https://ctan.org/pkg/xparse?lang=en) package provides environments in which the end clause can see the arguments, but again this is an additional LaTeX dependency.
4. You decide not to add any more dependencies and leave it to the user to use one of the above if they need them.

tk0miya:
With docutils' style, the following code will be converted the latter one internally
```
\begin{DUclass}{red}\begin{DUclass}{bold} content \end{DUclass}{bold}\end{DUclass}{red}
```
```
\DUCLASSred
\DUCLASSbold
content
\endDUCLASSbold
\endDUCLASSred
```

So users can change the behavior for each class by defining `\DUCLASS***` and `\endDUCLASS***`.

I think it's important to allow customizing each class. It enables that one extension creates "red" class and its macro, and another extension creates "bold" class and its macro.

Additionally, could you add a testcase for this?


Note: I confirmed the docutils' macro via this code:
```
\documentclass{report}
\usepackage{color}

% copied from https://repo.or.cz/docutils.git/blob/HEAD:/docutils/docutils/writers/latex2e/docutils.sty
\newenvironment{DUclass}[1]%
  {% "#1" does not work in end-part of environment.
   \def\DocutilsClassFunctionName{DUCLASS#1}
     \csname \DocutilsClassFunctionName \endcsname}%
  {\csname end\DocutilsClassFunctionName \endcsname}%

\def\DUCLASSred{\color{red}}
\def\endDUCLASSred{}
\def\DUCLASSbold{\bf}
\def\endDUCLASSbold{}

\begin{document}
  \begin{DUclass}{red}
    \begin{DUclass}{bold}
      hello
    \end{DUclass}
    hello
  \end{DUclass}
  hello
\end{document}
```

tk0miya:
Ah, sorry. My last comment was wrong.

```
\begin{DUclass}{red}\begin{DUclass}{bold} content \end{DUclass}\end{DUclass}
```

No argument is passed to the `\end{DUclass}`.

dham:
OK will do. I am a little unclear on how a test of LaTeX output is supposed to work. Can you point me at an example.

dham:
The requested code changes are in. I've browsed the tests and am still slightly mystified as to how they work, so would appreciate some guidance.

tk0miya:
>I've browsed the tests and am still slightly mystified as to how they work, so would appreciate some guidance.

1. Add a Sphinx project for testing under `tests/roots` directory (like `tests/roots/test-latex-container`).
2. Add a testcase to `tests/test_*.py` (like `tests/test_build_latex.py`).

In this case, please add a test case into `tests/test_build_latex.py` like following:

```
@pytest.mark.sphinx('latex', testroot='latex-container')  # The name of sphinx project that added at step 1.
def test_latex_nested_tables(app, status, warning):
    app.builder.build_all()
    result = (app.outdir / 'test.tex').read_text()
    assert r'blah blah' in result
```

dham:
Thanks for the advice. A test is now included.

dham:
Just emerged from marking hell. I see an administrator has triggered the tests and they have passed. Is there anything more I should do in order to get this merged?

Many thanks,

David

sphinx/texinputs/sphinxlatexcontainers.sty
```diff
@@ -0,0 +1,17 @@
...
+% change this info string if making any custom modification
+\ProvidesFile{sphinxlatexcontainers.sty}[2021/05/03 containers]
+
+% The purpose of this file is to provide a dummy environment sphinxclass
+% which will be inserted for each class in each container directive. The class
+% name will be passed as the argument to the environment. User-defined
+% formatting of directives can be achieved by renewing the definition of this
+% environment, and branching on the value of the argument.
+
+\ifx\sphinxclass\undefined % poor man's "provideenvironment"
```
jfbu: I see this test aims at adding flexibility but what is use case envisioned? Isn't it simpler to drop it? A user or extension can use `\renewenvironment` in preamble or via a package itself requiring sphinx.sty hence this file.

tests/roots/test-latex-container/index.rst
```diff
@@ -0,0 +1,4 @@
+.. container:: classname
+
+    text
```
jfbu: nitpicking: we usually indent by 3 spaces directive contents not 4

sphinx/texinputs/sphinxlatexcontainers.sty
```diff
@@ -0,0 +1,17 @@
...
+    \newenvironment{sphinxclass}[1]{
```
jfbu: although this should be always get executed in vertical mode, simply for the sake of being annoying I would prefer `{%` here.

sphinx/texinputs/sphinxlatexcontainers.sty
```diff
@@ -0,0 +1,17 @@
...
+        \def\sphinxClassFunctionName{sphinxCLASS#1}%
+        \csname \sphinxClassFunctionName \endcsname}%
```
jfbu: I think it is more LaTeXian to actually use `\begin` here. This would also have the advantage to enable the "hooks" which every environment now has with recent latex. For example, some extension is styling  a class in a certain way for latex. A user may then define a hook to the sphinxCLASS``<name>`` environment which will for example trigger boldface font to be used. This is simpler for user than redefining the environment. Admittedly this is a bit far-fetched and does not allow everything as the "begin hook" will get executed right before any code the environment may have defined, so can not override stuff still to be executed.

To do this the code could be like this:
```
\ltx@ifundefined{\sphinxClassFunctionName}
        {}% undefined so do nothing
        {\expandafter\begin\expandafter{\sphinxClassFunctionName}}
```

in place of
``\csname \sphinxClassFunctionName \endcsname``.

The `\expandafter` are not really needed, but they do not hurt and will make the `\begin` be used in the safest possible way.

sphinx/texinputs/sphinxlatexcontainers.sty
jfbu: Also, I wonder if `#1` needs some sanitizing of any sort. Accented letters or other special chars are not usable as LaTeX environment names and could cause breakage. But digits and spaces will be ok.

sphinx/texinputs/sphinxlatexcontainers.sty
```diff
@@ -0,0 +1,17 @@
...
+    {\csname end\sphinxClassFunctionName \endcsname}%
```
jfbu: Here one would use 

```
\ltx@ifundefined{\sphinxClassFunctionName}
        {}% we did nothing so we keep doing nothing
        {\expandafter\end\expandafter{\sphinxClassFunctionName}}
```

In fact thinking about it similar code exists in sphinxlatexadmonitions.sty https://github.com/sphinx-doc/sphinx/blob/f14561da80c263b867b51e27a75c4f9cdacfeb29/sphinx/texinputs/sphinxlatexadmonitions.sty#L136-L146

where an `\edef` is used because of the `sphinx` prefix before `\spx@noticetype`, and `\expandafter` would not work.

sphinx/texinputs/sphinxlatexcontainers.sty
```diff
@@ -0,0 +1,17 @@
+%% CONTAINER DIRECTIVES
+%
+% change this info string if making any custom modification
+\ProvidesFile{sphinxlatexcontainers.sty}[2021/05/03 containers]
+
+% The purpose of this file is to provide a dummy environment sphinxclass
+% which will be inserted for each class in each container directive. The class
+% name will be passed as the argument to the environment. User-defined
+% formatting of directives can be achieved by renewing the definition of this
+% environment, and branching on the value of the argument.
```
jfbu: Isn't it more apt to say that user customization goes via defining suitable ``sphinxCLASSname`` environments? Few if any would have to modify the master `sphinxclass` environment which is defined here and whose main purpose is to be innocent if the environment specified by the directive is actually not defined at LaTeX level.

By the way,
I propose to rename this environment to `sphinxuseclass`, and configure it to use for each specific class the ``sphinxclassname`` lowercased name (i.e. in the definition of `\sphinxClassFunctionName`), not `sphinxCLASSname`.

sphinx/writers/latex.py
```diff
@@ -1975,10 +1975,14 @@ def depart_compound(self, node: Element) -> None:
         pass
 
     def visit_container(self, node: Element) -> None:
-        pass
+        classes = node.get('classes', [])
+        for c in classes:
+            self.body.append('\n\\begin{sphinxclass}{%s}' % c)
```
jfbu: as specified above I would prefer `sphinxuseclass` here

jfbu (changes requested):
LGTM with nits!

Thanks for nice contribution and sorry for delay

I assumed during reviewing this is only block-level mark-up as per [docutils doc](https://docutils.sourceforge.io/docs/ref/rst/directives.html#container) so we don't have to worry about space tokens.

sphinx/texinputs/sphinxlatexcontainers.sty
jfbu: Actually what is important is that both `\begin` and `\end` are used in exactly the same way. In the quoted code from sphinxlatexadmonitions.sty, if `\begin{sphinx\spx@noticetype}` was used then it would have to be `\end{shinx\spx@noticetype}` and it would work too.

In our case we either do not use any `\expandafter` or if we use them, we use them both for `\begin` and `\end`.

jfbu:
In our latex codebase we have this https://github.com/sphinx-doc/sphinx/blob/e0b1e1002b500acc63dfd0806f8095dd6b27037b/sphinx/texinputs/sphinxlatexobjects.sty#L167-L180

which was inherited many years ago from DocUtils I presume. Notice it uses `\detokenize` to avoid issues with non-ascii or special characters. We may at a later stage (after this is merged once updated) consider similar.

But then we would need to provide some documented interface like this

```latex
\newsphinxclassenv{red}{\color{red}}{}
```

which would facilitate Sphinx user define the `sphinxclassred` environment. This looks silly but now the user can use ``grün`` for example in place of `red` (assuming here  no special escape happens regarding ``..container:: grün``).

The definition would simply be

```latex
\newcommand\newsphinxclassenv[3][{\newenvironment{\detokenize{#1}}{#2}{#3}}
```

but then we have to use `\detokenize` in the master `sphinxuseclass` dispatcher as in the `DUrole` above.

This can be added later after this will be merged once updated.

EDIT: and then we can do a `\renewsphinxclassenv` [our `renew` could work even in environment in fact does not exist, I have always considered the LaTeX behaviour of raising an error in such case an interface error] and a `\providesphinxclassenv`, making nice for user if in future the functionality is used by extensions. [to be complete also some conditional `\sphinxifclassenvexists{code if it exists}{code if it does not exist}` but perhaps I am going too far].

The `sphinxclass<detokenized name>` environments defined by `\newsphinxclassenv` et alia are of a simple no-parameter type, as we have no interface for that.

sphinx/texinputs/sphinxlatexcontainers.sty
dham: Duly dropped. I cargo-culted this from somewhere and am not about to attempt to justify my decision to do so.

tests/roots/test-latex-container/index.rst
dham: Done

sphinx/texinputs/sphinxlatexcontainers.sty
dham: Done.

sphinx/texinputs/sphinxlatexcontainers.sty
dham: OK. I believe I have implemented your first suggestion.

sphinx/texinputs/sphinxlatexcontainers.sty
dham: I've renamed as proposed and rewritten the docs.

sphinx/writers/latex.py
dham: Done.

dham:
OK. I think I claim I have dealt with all the requested changes. I haven't touched the handling of special characters as I believe you suggest this should be postponed to future work.

jfbu (approved):
LGTM! Thanks for contribution.

We can add later handling of non-ascii class names.

jfbu:
Waiting for testing to complete, then I will push to this branch a CHANGES entry and merge.

Thanks for great addition!

jfbu:
rather I merge now and will push the CHANGES entry later once I fix may github email privacy settings...

tk0miya:
Looks good to me too!

@jfbu Thank you for reviewing :-)

dham:
This is my first contribution to the main sphinx project. Thank you for making it a very positive experience. I hope to have reason to contribute more in the future.

PR #9170, opened 2021-05-05
Implement simple sphinx.yaml config parsing

chrisjsewell:
Subject: A simple implementation of #9040

### Feature or Bugfix

- Feature

### DIFF
--- a/doc/usage/quickstart.rst
+++ b/doc/usage/quickstart.rst
@@ -251,6 +251,13 @@ config values.
 
 .. todo:: Move this entire doc to a different section
 
+.. versionadded:: 4.0.0
+
+   It is now possible to use a :file:`sphinx.yaml`,
+   in-place of the :file:`conf.py` (the two may not co-exist).
+   This should contain a valid `YAML <https://yaml.org>`__
+   dictionary of the configuration variables.
+
 Autodoc
 -------
 
--- a/doc/usage/quickstart.rst
+++ b/doc/usage/quickstart.rst
@@ -254,7 +254,7 @@ config values.
 .. versionadded:: 4.0.0
 
    It is now possible to use a :file:`sphinx.yaml`,
-   in-place of the :file:`conf.py` (the two may not co-exist).
+   in-place of the :file:`conf.py` (the two must not co-exist).
    This should contain a valid `YAML <https://yaml.org>`__
    dictionary of the configuration variables.
 
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,7 @@ install_requires = [
     'requests>=2.5.0',
     'setuptools',
     'packaging',
+    'pyyaml'
 ]
 
 extras_require = {
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -125,7 +125,7 @@ class Sphinx:
     """The main application class and extensibility interface.
 
     :ivar srcdir: Directory containing source.
-    :ivar confdir: Directory containing ``conf.py``.
+    :ivar confdir: Directory containing ``conf.py`` or ``sphinx.yaml``.
     :ivar doctreedir: Directory for storing pickled doctrees.
     :ivar outdir: Directory for storing build documents.
     """
@@ -151,12 +151,9 @@ class Sphinx:
         self.srcdir = abspath(srcdir)
         self.outdir = abspath(outdir)
         self.doctreedir = abspath(doctreedir)
-        self.confdir = confdir
+        self.confdir = abspath(confdir) if confdir else confdir
         if self.confdir:  # confdir is optional
-            self.confdir = abspath(self.confdir)
-            if not path.isfile(path.join(self.confdir, 'conf.py')):
-                raise ApplicationError(__("config directory doesn't contain a "
-                                          "conf.py file (%s)") % confdir)
+            Config.get_config_file(self.confdir)
 
         if not path.isdir(self.srcdir):
             raise ApplicationError(__('Cannot find source directory (%s)') %
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -152,8 +152,6 @@ class Sphinx:
         self.outdir = abspath(outdir)
         self.doctreedir = abspath(doctreedir)
         self.confdir = abspath(confdir) if confdir else confdir
-        if self.confdir:  # confdir is optional
-            Config.get_config_file(self.confdir)
 
         if not path.isdir(self.srcdir):
             raise ApplicationError(__('Cannot find source directory (%s)') %
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -165,8 +165,8 @@ class Config:
         self.extensions: List[str] = config.get('extensions', [])
 
     @classmethod
-    def get_config_file(cls, confdir: str) -> str:
-        """Check the configuration directory for a configuration file."""
+    def read(cls, confdir: str, overrides: Dict = None, tags: Tags = None) -> "Config":
+        """Create a Config object from a configuration file."""
         config_py = path.join(confdir, CONFIG_FILENAME_PY)
         config_yaml = path.join(confdir, CONFIG_FILENAME_YAML)
         if path.isfile(config_py):
@@ -175,24 +175,14 @@ class Config:
                     __("config directory contains both a %s and %s file (%s)"
                        ) % (CONFIG_FILENAME_PY, CONFIG_FILENAME_YAML, confdir)
                 )
-            return config_py
+            namespace = eval_config_file(config_py, tags)
         elif path.isfile(config_yaml):
-            return config_yaml
+            namespace = read_config_file_yaml(config_yaml, tags)
         else:
             raise ApplicationError(
                 __("config directory doesn't contain a "
                   "%s or %s file (%s)") % (CONFIG_FILENAME_PY, CONFIG_FILENAME_YAML, confdir)
             )
-
-    @classmethod
-    def read(cls, confdir: str, overrides: Dict = None, tags: Tags = None) -> "Config":
-        """Create a Config object from configuration file."""
-        filename = cls.get_config_file(confdir)
-        namespace = {}
-        if filename.endswith(CONFIG_FILENAME_PY):
-            namespace = eval_config_file(filename, tags)
-        elif filename.endswith(CONFIG_FILENAME_YAML):
-            namespace = read_config_file_yaml(filename, tags)
         return cls(namespace, overrides or {})
 
     def convert_overrides(self, name: str, value: Any) -> Any:
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -16,13 +16,14 @@ from os import getenv, path
 from typing import (TYPE_CHECKING, Any, Callable, Dict, Generator, Iterator, List, NamedTuple,
                     Optional, Set, Tuple, Union)
 
-from sphinx.errors import ConfigError, ExtensionError
+from sphinx.errors import ApplicationError, ConfigError, ExtensionError
 from sphinx.locale import _, __
 from sphinx.util import logging
 from sphinx.util.i18n import format_date
 from sphinx.util.osutil import cd, fs_encoding
 from sphinx.util.tags import Tags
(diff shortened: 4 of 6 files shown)


sphinx/config.py
```diff
@@ -161,11 +164,37 @@ def __init__(self, config: Dict[str, Any] = {}, overrides: Dict[str, Any] = {})
                 config['extensions'] = self.overrides.pop('extensions')
         self.extensions: List[str] = config.get('extensions', [])
 
+    @classmethod
+    def get_config_file(cls, confdir: str) -> str:
```
tk0miya: What purpose is the return value of this function? I think the filename (and filetype) of config should be encapsulated to this class. So it would be better to rename this as `Config.exists?(confdir)` instead.

Or it's enough to raise an error on reading instead of checking existence.

sphinx/config.py
```diff
@@ -338,6 +367,20 @@ def eval_config_file(filename: str, tags: Optional[Tags]) -> Dict[str, Any]:
     return namespace
 
 
+def read_config_file_yaml(filename: str, tags: Optional[Tags]) -> Dict[str, Any]:
```
tk0miya: Why this function takes `tags`?

tests/test_config.py
```diff
@@ -74,6 +74,43 @@ def test_core_config(app, status, warning):
     assert cfg['project'] == cfg.project == 'Sphinx Tests'
 
 
+@pytest.mark.sphinx(testroot='config_yaml', confoverrides={
+    'root_doc': 'root',
+    'nonexisting_value': 'True',
+    'latex_elements.maketitle': 'blah blah blah',
+    'modindex_common_prefix': 'path1,path2'})
```
tk0miya: It would be better to separate the core feature of the config object to another cases (overrides, `in` API, and so on). They're not related at all.

tk0miya (commented):
LGTM with nits.

sphinx/config.py
chrisjsewell: so that they can be added to the namespace before it is returned

tests/test_config.py
chrisjsewell: is it enough to just remove the overrides and only keep the initial tests?

```
    # simple values
    assert 'project' in cfg.__dict__
    assert cfg.project == 'Sphinx <Tests>'
    assert cfg.templates_path == ['_templates']
```

sphinx/application.py
```diff
@@ -151,12 +151,9 @@ def __init__(self, srcdir: str, confdir: Optional[str], outdir: str, doctreedir:
...
         self.outdir = abspath(outdir)
         self.doctreedir = abspath(doctreedir)
-        self.confdir = confdir
+        self.confdir = abspath(confdir) if confdir else confdir
         if self.confdir:  # confdir is optional
-            self.confdir = abspath(self.confdir)
-            if not path.isfile(path.join(self.confdir, 'conf.py')):
-                raise ApplicationError(__("config directory doesn't contain a "
-                                          "conf.py file (%s)") % confdir)
+            Config.get_config_file(self.confdir)
```
chrisjsewell: is it necessary to have this check here, or as you say below, is it's enough to just raise an error on reading (@ line 214), instead of checking existence?

sphinx/config.py
chrisjsewell: see above 👆

sphinx/application.py
tk0miya: Good point. It seems nobody touches the config file until L214. So it's enough to raise an exception inside `Config.read()`.

sphinx/config.py
tk0miya: Hmm? How is it used? AFAIK, nobody will be able to access it. I still don't understand why it's needed. In `conf.py`'s case, it allows to change the configurations in `conf.py` dynamically. It means the tags variable is the input of `conf.py`, not a return value of the `read_config_*()` function.

tests/test_config.py
tk0miya: Yes :-)

chrisjsewell: done 👍

sphinx/application.py
chrisjsewell: done 👍

sphinx/config.py
chrisjsewell: ah ok I see (I hadn't actually heard about tags until recently), I've removed it from the function then 👍

sphinx/config.py
```diff
@@ -163,9 +166,23 @@ def __init__(self, config: Dict[str, Any] = {}, overrides: Dict[str, Any] = {})
...
     def read(cls, confdir: str, overrides: Dict = None, tags: Tags = None) -> "Config":
-        """Create a Config object from configuration file."""
-        filename = path.join(confdir, CONFIG_FILENAME)
-        namespace = eval_config_file(filename, tags)
+        """Create a Config object from a configuration file."""
+        config_py = path.join(confdir, CONFIG_FILENAME_PY)
+        config_yaml = path.join(confdir, CONFIG_FILENAME_YAML)
+        if path.isfile(config_py):
+            if path.isfile(config_yaml):
+                raise ApplicationError(
```
tk0miya: I noticed this error is related to the configuration. So it would be much better to use `ConfigError` instead.

sphinx/config.py
```diff
@@ -338,6 +355,20 @@ def eval_config_file(filename: str, tags: Optional[Tags]) -> Dict[str, Any]:
...
+def read_config_file_yaml(filename: str, tags: Optional[Tags]) -> Dict[str, Any]:
+    """Evaluate a YAML config file."""
+    try:
+        with open(filename, encoding='utf-8') as fp:
+            namespace = yaml.safe_load(fp)
+    except Exception as exc:
+        msg = __("There is a syntax error in your configuration file:\n\n%s")
+        raise ConfigError(msg % traceback.format_exc()) from exc
+    if not isinstance(namespace, dict):
+        raise ConfigError(__("Configuration file is not a dictionary"))
```
tk0miya: minor: It would be better to mention the filename `config.yml` in these error messages. It lets users know what file they should fix.

Note: I know this is a trial implementation. So I marked this as minor.

chrisjsewell: done and added tests for parsing failures

sphinx/config.py
chrisjsewell: done

tests/test_config.py
```diff
@@ -149,6 +159,18 @@ def test_errors_warnings(logger, tempdir):
     assert logger.called is False
 
 
+@mock.patch("sphinx.config.logger")
```
tk0miya: Why mocking logger?

chrisjsewell: just a copy/paste, now removed

setup.py
```diff
@@ -32,6 +32,7 @@
     'requests>=2.5.0',
     'setuptools',
     'packaging',
+    'pyyaml'
```
tk0miya, line 35: Note: I need to update a dependency list on merging this.

tk0miya:
LGTM. Thank you for your work!

chrisjsewell:
> LGTM. Thank you for your work!

thanks for the quick review!

chrisjsewell:
BTW, what is necessary to get this merged?

tk0miya:
I think the discussion in #9040 has not been agreed yet. So I can't merge this until then.

sphinx/config.py
```diff
@@ -163,9 +166,23 @@ def __init__(self, config: Dict[str, Any] = {}, overrides: Dict[str, Any] = {})
...
     def read(cls, confdir: str, overrides: Dict = None, tags: Tags = None) -> "Config":
-        """Create a Config object from configuration file."""
-        filename = path.join(confdir, CONFIG_FILENAME)
-        namespace = eval_config_file(filename, tags)
+        """Create a Config object from a configuration file."""
+        config_py = path.join(confdir, CONFIG_FILENAME_PY)
+        config_yaml = path.join(confdir, CONFIG_FILENAME_YAML)
+        if path.isfile(config_py):
+            if path.isfile(config_yaml):
+                raise ConfigError(
```
jakobandersen, line 174: I think the initial version should still allow both files, such that one can change to YAML incrementally, and in the end still have dynamic things in ``conf.py`` and static things in ``sphinx.yaml``.

chrisjsewell: No, I disagree.

1. For a version 1 it is best to keep it simple, strict and not add functionality that may need to be later removed
2. We are not removing conf.py, so there is no need for an incremental transition
3. It somewhat defeats the purpose of a static configuration, if you would also have to read `conf.py` to find all the variables

choldgraf: and importantly, I think that a decision like this would require more discussion, and isn't strictly a blocker on a first-step implementation, so IMO we should defer that to a future PR after this feature exists and people have had time to use it and get a feel for how it could be improved

choldgraf (approved):

sphinx/config.py
jakobandersen: Well, if you are sitting with an existing project and want the benefit of the static configuration in the YAML file, you will with this be forced to convert _all_ in one go. There is not incremental path to check that you didn't make an error. And if you have just a single dynamic variable (say ``version``) there is no way to use this.
Regarding 3, as argued in the issue: all projects have conventions in one form or another. One such convention could be to ban ``conf.py`` if you want to guarantee static configuration. Why should that block others from using Sphinx in a different way?

chrisjsewell: As I and others have already noted, this is a version 1 implementation that implements the minimum desired functionality and (intentionally) does not preclude the later addition of co-file support.
If you greatly desire such functionality, then I would encourage you to open a follow-up PR to achieve this.
For now this allows "basic" users of sphinx to use the sphinx.yaml, and in no way affects your current sphinx workflow

chrisjsewell:
@tk0miya I feel this now has enough tacit approval? https://github.com/sphinx-doc/sphinx/issues/9040#issuecomment-833693871, https://github.com/sphinx-doc/sphinx/issues/9040#issuecomment-833693871, https://github.com/sphinx-doc/sphinx/issues/9040#issuecomment-833910428
(and, as mentioned, allows a route for further non-breaking iteration)

doc/usage/quickstart.rst
```diff
@@ -251,6 +251,13 @@ config values.
 
 .. todo:: Move this entire doc to a different section
 
+.. versionadded:: 4.0.0
```
chrisjsewell, line 254: FYI @ericholscher

doc/usage/quickstart.rst
```diff
@@ -251,6 +251,13 @@ config values.
...
+
+   It is now possible to use a :file:`sphinx.yaml`,
+   in-place of the :file:`conf.py` (the two may not co-exist).
```
humitos: In my _basic_ English knowledge, I understand that both may or may not co-exist. So, in my head, this means that I can use `conf.py` and `sphinx.yaml` files together or only one of them. I'd change this sentence to "the two _must not_ co-exist" to avoid this confusion.

humitos (approved):
I'm happy with this PR and I'm 👍🏼 on merging it.

I think in another PR we could improve the documentation to show an example of the `sphinx.yaml` and also guide users with small dynamic requirements about how to proceed (in #9040 @hukkinj1 mentioned `bump2version` for example). We could also point them to write a small extension in case they have bigger dynamic requirements as well.

It may be good to mention also that no dynamic YAML configs are allowed (on purpose) in the `sphinx.yaml` file. In https://github.com/sphinx-doc/sphinx/issues/9040#issuecomment-817009677 I mentioned how to this and I'd expect users wanting this and probably complaining about this not working 😄

doc/usage/quickstart.rst
choldgraf: +1 to "must not" or "cannot" instead of "may not"

chrisjsewell:
```suggestion
   in-place of the :file:`conf.py` (the two must not co-exist).
```

chrisjsewell: done 👍

PR #9174, opened 2021-05-06
Add `Sphinx.add_html_assets_in_all_pages`

### LINKED ISSUE #9115: Config to add JS/CSS to all generated HTML pages

After https://github.com/sphinx-doc/sphinx/pull/8631 was merged, JS/CSS files for MathJax are only added to those pages that contains math formulas. This is not a problem unless you want to embed the content of one of those pages where the JS/CSS are included from a different page.

This problem is clear on my extension [`sphinx-hoverxref`](https://sphinx-hoverxref.readthedocs.io/) that shows a tooltip with the content of the linked page when hovering over a link. Example: if you hover on "tooltip with Mathjax" in https://sphinx-hoverxref.readthedocs.io/en/sphinx-35/usage.html#tooltip-with-mathjax you won't see the MathJax rendered properly inside the tooltip

Screenshot_2021-04-19_11-02-47

and an error in the console "Not triggering MathJax because it is not defined". This is because MathJax is not included in `usage.html` page but it's included in target page (`mathjax.html`)

_Note that I'm mentioning the problem only with MathJax for now since it's a known extension that it's using this new feature, but this applies to all the extension that start including the JS/CSS only on pages where the extension is present_

I think a config to force the old behavior that includes all the JS/CSS in all the pages could work. I know this has some drawbacks, but as far is optional, I think it's good.

1. `sphinx-hoverxref` uses an endpoint from Read the Docs' API to get the content to embed in the tooltip. We are discussing if it's possible to return `"extras": ["mathjax.js", "mathjax.css"]` together with the content to be included so the JS client knows it has to include these assets before rendering the content. See https://github.com/readthedocs/readthedocs.org/pull/8039#discussion_r604449923
2. Another option could be to detect from the Sphinx extension "what are the assets required in the target page and add them into this page as well". Note that this won't work when using intersphinx.

**Additional context**

Reference: https://github.com/sphinx-doc/sphinx/pull/8631#issuecomment-821162722
Downstream issue: https://github.com/readthedocs/sphinx-hoverxref/issues/119

humitos:
Subject: Add an API to enable including assets in all HTML pages.

### Feature or Bugfix

- Feature

### Purpose

This new method in the `Sphinx` object allows extensions to communicate to Sphinx that it's preferred to include HTML assets in all the pages. However, it's extensions developers' responsibility to follow this config and decide whether or not to include the assets required.

Extensions developers' can check `Sphinx.html_assets_in_all_pages` together with any other logic they may have to decide if the assets will be included in the rendered page or not.

### Relates

- Closes #9115


sphinx/application.py
```diff
@@ -1181,6 +1182,13 @@ def add_env_collector(self, collector: Type[EnvironmentCollector]) -> None:
         logger.debug('[app] adding environment collector: %r', collector)
         collector().enable(self)
 
+    def add_html_assets_in_all_pages(self):
```
tk0miya: I'm thinking what is the best interface for this feature.

* "add" is not good for this case. The flag is not addable.
* "set" is better than "add".
* "html_assets_in_all_pages" flag is a bit many worded. A setting like "html_assets_policy=always" would be simpler than a flag.
* It might be better to add a method to set a setting for general purpose like `set_extension_variable('html', 'html_assets_in_all_pages', True)`

sphinx/application.py
```diff
@@ -1181,6 +1182,13 @@ def add_env_collector(self, collector: Type[EnvironmentCollector]) -> None:
...
+        """Tell extensions to insert HTML assets in all the pages.
+
+        .. versionadded: 4.1
+        """
+        self.html_assets_in_all_pages = True
```
tk0miya: Could you store this into the `sphinx.register.SphinxComponentRegistry` instead? It would be better not to have settings or components into the app object directly.

sphinx/application.py
tk0miya: I'd like to hear the opinion from @shimizukawa . So let's wait for him for a while.

sphinx/application.py
humitos: Thanks for pointing at this to me. I wasn't aware of the `SphinxComponentRegistry` 😅  I'll move this there once we have defined the name for the value/method.

tk0miya: I wish https://github.com/sphinx-doc/sphinx/pull/9215 will help you.

humitos: Thanks! That PR helped me to understand what's required here and where to store this value. I think I know what to do now 😄 

I'm just waiting for a better name suggestion to update the PR.

sphinx/application.py
humitos: > "add" is not good for this case. The flag is not addable.
 
Yeah. I tried to follow the pattern that I saw. However, now that I'm reading it more carefully, `add_` is being used with a different meaning than I gave it to my variable.

> "set" is better than "add".

`set_` may works better, yes 👍🏼 

> "html_assets_in_all_pages" flag is a bit many worded. A setting like "html_assets_policy=always" would be simpler than a flag.

IMO, I would say that it makes sense to use this approach if we already know that we will have other different policies --do we?. Otherwise, I think we will be just complicating the API.

> It might be better to add a method to set a setting for general purpose like `set_extension_variable('html', 'html_assets_in_all_pages', True)`

I like the "general purpose" solution. I'd expect to have more of these extension variables as Sphinx keeps growing. However, we still need to define a good name for the variable, right?

shimizukawa: +1 for `set_`.

`html_assets_policy='always'` is fine to me because:

* Shorter name
* Potential for future enhancement

However I can agree with "I think we will be just complicating the API."

tk0miya: @shimizukawa Thank you for your comment. Okay, let's go with `set_html_assets_policy('always')` (The default is `'per_page'`).

sphinx/ext/mathjax.py
```diff
@@ -79,7 +79,7 @@ def install_mathjax(app: Sphinx, pagename: str, templatename: str, context: Dict
                              'mathjax extension to work')
 
     domain = cast(MathDomain, app.env.get_domain('math'))
-    if domain.has_equations(pagename):
+    if app.registry.html_assets_policy == 'always' or domain.has_equations(pagename):
```
humitos, line 82: @tk0miya I don't like too much how this line is checking for the `always` policy. Do you have a better suggestion for this?

Daltz333:
Any bump on this? This PR is probably the only blocker on my team for migration to Sphinx 4.0

tk0miya:

PR #9176, opened 2021-05-07
[IMP] util: detect incorrect locale_dirs path

mart-e:
In case the config is incorrect (e.g. wrong relative path), it is silently ignored.
This makes it easier for debugging.

### Feature or Bugfix
- Feature/Improvement

### Purpose
It was not clear to me that the `locale_dirs` variable was relative to the source dir and not to the config path. It took me a long time to detect this was the reason my translations were ignored when building.
With this logger, I could have understood the error was in my config file.

### DIFF
--- a/sphinx/util/i18n.py
+++ b/sphinx/util/i18n.py
@@ -91,8 +91,11 @@ class CatalogRepository:
 
         for locale_dir in self._locale_dirs:
             locale_dir = path.join(self.basedir, locale_dir)
-            if path.exists(path.join(locale_dir, self.language, 'LC_MESSAGES')):
+            locale_path = path.join(locale_dir, self.language, 'LC_MESSAGES')
+            if path.exists(locale_path):
                 yield locale_dir
+            else:
+                logger.verbose(__('locale_dir %s does not exists'), locale_path)
 
     @property
     def pofiles(self) -> Generator[Tuple[str, str], None, None]:


sphinx/util/i18n.py
```diff
@@ -91,8 +91,11 @@ def locale_dirs(self) -> Generator[str, None, None]:
 
         for locale_dir in self._locale_dirs:
             locale_dir = path.join(self.basedir, locale_dir)
-            if path.exists(path.join(locale_dir, self.language, 'LC_MESSAGES')):
+            locale_path = path.join(locale_dir, self.language, 'LC_MESSAGES')
+            if path.exists(locale_path):
                 yield locale_dir
+            else:
+                logger.info(__('locale_dir %s does not exists'), locale_path)
```
tk0miya: Could you use `logger.warning()` instead here. I believe it's a wrong entry. So it should be fixed by user's side. So using a warning is better for this case.

tk0miya (changes requested):
LGTM with nits

sphinx/util/i18n.py
mart-e: I did put a warning at first but I was wondering if it wasn't making false negative with the for loop. Is it a valid scenario where you have multiple `locale_dirs` but only some containing the translations? Or not having all languages in all `locale_dirs`?

tk0miya: Exactly. It's not good to use the WARNING level for this purpose. From its perspective, the INFO is also noisy. How about using VERBOSE instead? Then the message will be shown only `-v` option is given. I think it's better.

mart-e: Good enough for me :)

tk0miya (approved):

tk0miya:
Thank you for update. Merging now.

tk0miya:
FYI: I posted https://github.com/sphinx-doc/sphinx/pull/9184 to describe this behavior.

PR #9179, opened 2021-05-08
Fix module attributes not output "all attributes"

Lu-Yi-Hsun:
Subject: Fix autosummary module attributes not output "all attributes"

4.x
 

### Feature or Bugfix

- show all module attributes

 
### Detail

sample.py
```py a.py
import networkx as nx
a = 3#Built-in Types
G = nx.DiGraph()# any other object types
```
in the sample
only show "a" if varable is "Built-in Types"
NOT show "G" if varable is "any other object types"

### DIFF
--- a/sphinx/ext/autosummary/generate.py
+++ b/sphinx/ext/autosummary/generate.py
@@ -327,7 +327,7 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
         ns['exceptions'], ns['all_exceptions'] = \
             get_members(obj, {'exception'}, imported=imported_members)
         ns['attributes'], ns['all_attributes'] = \
-            get_module_attrs(ns['members'])
+            get_module_attrs(dir(obj))
         ispackage = hasattr(obj, '__path__')
         if ispackage and recursive:
             ns['modules'], ns['all_modules'] = get_modules(obj)


sphinx/ext/autosummary/generate.py
```diff
@@ -327,7 +327,7 @@ def get_modules(obj: Any) -> Tuple[List[str], List[str]]:
         ns['exceptions'], ns['all_exceptions'] = \
             get_members(obj, {'exception'}, imported=imported_members)
         ns['attributes'], ns['all_attributes'] = \
-            get_module_attrs(ns['members'])
+            get_module_attrs(dir(obj))
```
tk0miya, line 330: This change puts the imported components to `all_attributes`.
In your example, `nx` will be added into `all_attributes`. I think it's not good behavior.

Lu-Yi-Hsun: @tk0miya  in the example, i want document `module attributes` show `a` and `G`,but now only can show Built-in Types(`a`) now. In `class attributes` can show `a` and `G` but `module attributes` can not show `G`.

## i want show `a` and `G` on `module attributes`
image
## click the`G` i can see the detail
image

tk0miya: I said this implementation is not good because it causes regression error. So you need to change the way to realize your request.

AA-Turner:
Closing out-of-date PR (feel free to update & re-open).

A

PR #9236, opened 2021-05-16
Add tests for dataclasses: autodoc and autosummary

pbudzyns:
Subject: Add tests with dataclasses

This PR is a continuation to https://github.com/sphinx-doc/sphinx/pull/9146 where support for instance attributes in `autosummary` was added. As Phil correctly pointed out this change also fixes generated content for `dataclasses`. This change was hard to notice because there were no unit tests including `dataclasses` present. As `dataclasses` became somehow standard thing for python >= 3.7 I believe it's a good idea to test if content generated for them is correct.  

### Feature or Bugfix
- Testing

### Purpose
- Test documenting dataclasses

### Relates
- https://github.com/sphinx-doc/sphinx/pull/9146

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -32,6 +32,7 @@ Features added
   entry and the definition of ``__init__()`` method
 * #8061, #9218: autodoc: Support variable comment for alias classes
 * #3257: autosummary: Support instance attributes for classes
+  and attributes for dataclasses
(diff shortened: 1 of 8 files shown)


tk0miya:
I feel no additional tests are needed for dataclasses. The mechanism is the same as the instance attributes annotations.

Note: I think the testcases for namedtuple is also not needed by the same reason.

PR #9242, opened 2021-05-16
refactor: Make the app argument for BuildEnvironment required

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- At present, some attributes of env object are considered as optional
because they have been initialized by None on the constructor.  But
they have always been fullfilled actually.
- To be clear the type hints of the env object, this makes the `app`
argument for the BuildEnvironment class required.  It can ensure the
attributes of env object are not optional.

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -10,6 +10,7 @@
 
 import os
 import pickle
+import warnings
 from collections import defaultdict
 from copy import copy
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


sphinx/environment/__init__.py
```diff
@@ -181,6 +183,9 @@ def __init__(self, app: "Sphinx" = None):
         # set up environment
         if app:
             self.setup(app)
+        else:
+            warnings.warn("The 'app' argument for BuildEnvironment() becomes required now.",
+                          RemovedInSphinx60Warning, stacklevel=2)
```
tk0miya, line 188: Note: We need to modify `sphinx.ext.autosummary:FakeDirective` too.

Issue #9291, opened 2021-06-03
:term: referenced via intersphinx is case sensitive


tomoinn:
**Describe the bug**

When a doc website A contains a term in a glossary, that term can be referenced without case sensitivity applied. If, however, that term is referenced from a second doc build B, by adding an intersphinx reference to A, only exact cases match.

**To Reproduce**

Add the following intersphinx reference:

```
'oe': ('https://icebreakerone.github.io/open-energy-technical-docs/main/', None)
```

Then attempt to reference a term within the glossary from a second project, i.e.

```
:term:`data provider` - this fails
:term:`Data Provider` - this correctly references back into the external glossary
```

**Expected behavior**

References to terms in external projects linked through intersphinx resolve in the same way as those defined locally.

**Your project**

https://github.com/icebreakerone/open-energy-technical-docs contains source, https://icebreakerone.github.io/open-energy-technical-docs/main/index.html is the site containing the manifest from which I'm trying to import terms.

**Environment info**

Sphinx build run from linux, python 3.8.6, sphinx 4.0.2

jhuitema:
I've also just run into this. From my (non-exhaustive) testing, it looks like the case-insensitive intersphinx term linking stopped working with Sphinx v3. The term links work correctly in Sphinx 2.4.4 and under.

AWhetter:
A git bisect indicates that this was an intentional change made in https://github.com/sphinx-doc/sphinx/commit/a5dadeb890beca64bb9eabe86e6634c9dd3e1d50 to address #7418.
It's documented that terms are now case sensitive, and this was a breaking change introduced for Sphinx v3.
There was a change made in https://github.com/sphinx-doc/sphinx/commit/61f00bf675c1a2f7955f3d342a560dc8680d87c5 (v3.0.2) to fall back to case insensitive matching and to raise a warning when this happens, but from testing it seems like this doesn't happen for matches against an intersphinx mapping.

tomoinn:
Yup, it looks like the resolver for :term: isn't used when loading references from intersphinx? If anyone has better knowledge of how things are structured they'd probably have a better idea, but as far as I can tell the resolution in intersphinx is happening in https://github.com/sphinx-doc/sphinx/blob/4d90277c49633705148793ab02b28353c8a59b48/sphinx/ext/intersphinx.py#L261 and just doing a straight 'is X in Y' check

tomoinn:
[...]

PR #9299, opened 2021-06-04
Allow case-insensitive match of :term: references through intersphinx

tomoinn:
Subject: Allow case-insensitive match of :term: references through intersphinx

### Feature or Bugfix

- Bugfix

### Purpose
- #9291 describes a problem where term references are resolved in a case insensitive fashion when matching a locally defined glossary, but when that same glossary is accessed through intersphinx this stops working and the term can only be matched with exact case. I'm not particularly familiar with the architecture of sphinx so I might well have put this in the wrong place, but it does fix this issue.

### Detail

Simple code change, detects that the objtype is `std:term`, builds a dict of lower case to original case for all keys in the inventory under inspection and uses that to perform case insensitive matching.

### Relates
- #9291

### DIFF
--- a/sphinx/ext/intersphinx.py
+++ b/sphinx/ext/intersphinx.py
@@ -305,19 +305,28 @@ def missing_reference(app: Sphinx, env: BuildEnvironment, node: pending_xref,
                     to_try.append((inventories.named_inventory[setname], full_qualified_name))
     for inventory, target in to_try:
         for objtype in objtypes:
-            # Special case handling for term to search in a case insensitive fashion
-            if objtype == 'std:term' and objtype in inventory:
-                cased_keys = {k.lower(): k for k in inventory[objtype].keys()}
-                if target.lower() in cased_keys:
-                    corrected_target = cased_keys[target.lower()]
-                    proj, version, uri, dispname = inventory[objtype][corrected_target]
+            if objtype not in inventory:
+                # Continue if there's nothing of this kind in the inventory
+                continue
+            if target in inventory[objtype]:
+                # Case sensitive match, use it
+                proj, version, uri, dispname = inventory[objtype][target]
+            elif objtype == 'std:term':
+                # Check for potential case insensitive matches for terms only
+                target_lower = target.lower()
+                insensitive_matches = list(filter(lambda k: k.lower() == target_lower,
+                                                  inventory[objtype].keys()))
+                if insensitive_matches:
+                    proj, version, uri, dispname = inventory[objtype][insensitive_matches[0]]
                 else:
+                    # No case insensitive match either, continue to the next candidate
                     continue
             else:
-                # Default behaviour pre-patch
-                if objtype not in inventory or target not in inventory[objtype]:
-                    continue
-                proj, version, uri, dispname = inventory[objtype][target]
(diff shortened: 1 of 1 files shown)


sphinx/ext/intersphinx.py
```diff
@@ -305,9 +305,19 @@ def missing_reference(app: Sphinx, env: BuildEnvironment, node: pending_xref,
                     to_try.append((inventories.named_inventory[setname], full_qualified_name))
     for inventory, target in to_try:
         for objtype in objtypes:
-            if objtype not in inventory or target not in inventory[objtype]:
-                continue
-            proj, version, uri, dispname = inventory[objtype][target]
+            # Special case handling for term to search in a case insensitive fashion
+            if objtype == 'std:term' and objtype in inventory:
+                cased_keys = {k.lower(): k for k in inventory[objtype].keys()}
```
AWhetter: We could avoid creating this whole dictionary by checking each key in turn?

```
lower_target = target.lower()
if any(lower_target == inv_key.lower() for inv_key in inventory[objtype]):
    ...
```

We could avoid doing even this if a case sensitive search returns a result?

```python
if objtype == 'std:term' and objtype in inventory:
    if target in inventory[objtype]:
        proj, version, uri, dispname = inventory[objtype][target]
    else:
        lower_target = target.lower()
        if any(lower_target == inv_key.lower() for inv_key in inventory[objtype]):
            ...
       else:
            continue
else:
    ...
```

tomoinn: I think you still have to build the dict, because after determining that there *is* a case insensitive match you have to look up the target, which is inherently a case sensitive process as it's just a dict key? That or something like

```python
            if objtype not in inventory:
                continue
            if target in inventory[objtype]:
                # Case sensitive match
                proj, version, uri, dispname = inventory[objtype][target]
            elif objtype == 'std:term':
                # Check for potential case insensitive match
                target_lower = target.lower()
                insensitive_matches = list(filter(lambda k: k.lower() == target_lower, inventory[objtype].keys()))
                if insensitive_matches:
                    proj, version, uri, dispname = inventory[objtype][insensitive_matches[0]]
                else:
                    continue
            else:
                continue
```

but I'm not sure this is any more efficient. I guess it's building a list rather than a dict.

AWhetter: Oh you're totally right :facepalm: This list approach will save a small amount of memory at least, plus it gives us the opportunity to raise a warning if multiple keys match.
I like the structure of the `if` statements that you're proposing too!

tomoinn: I've pushed a commit with the above (plus comments). I guess the argument for doing this with a dict is if there are a lot of cases and you can move the dict generation out of the per-lookup loop, so have it outside of the `for inventory, target in to_try:` block, but you still have the memory occupancy to deal with. Priorities, as always!

tk0miya:
LGTM! Could you add a testcase for this, please?

tomoinn:
I haven't abandoned this - been busy with the project that needs the fix, will add test cases when I have some time available.

tk0miya:
Now I added a testcase for this and merged. Thank you for your contribution!

GonzaloT23 (commented):
Ok

PR #9301, opened 2021-06-05
fix: Fix typo in add_directive docstring

jpmckinney:
The code example is missing `app.` before `add_directive`.


tk0miya:
Thanks!

PR #9317, opened 2021-06-09
Fix missing break statement for arrow shortcuts.

marxin:
When one presses left key at the first page, `case 37` falls
through to `case 39` and the next page is visited.


tk0miya:

PR #9326, opened 2021-06-13
Revert the removal of sphinx.util:force_decode()

tk0miya:
### Feature or Bugfix
- Feature
- Refactoring

### Purpose
- After the release of 4.0.0, some 3rd party extensions have became not
working with the latest Sphinx because `force_decode()` function was
removed.  It was deprecated since Sphinx-2.0 and warned for the removal
since 3.0.
- This reverts the removal and extends its deprecation period to 5.0.0.  I
hope it helps users of these extensions.
- refs: https://github.com/sphinx-doc/sphinx/pull/5620#issuecomment-860163848
- cc: @stevepiercy

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,9 @@ Deprecated
 Features added
 --------------
 
+* Revert the removal of ``sphinx.util:force_decode()`` to become some 3rd party
+  extensions available again during 5.0
+
 Bugs fixed
 ----------
 
--- a/doc/extdev/deprecated.rst
+++ b/doc/extdev/deprecated.rst
@@ -1015,7 +1015,7 @@ The following is a list of deprecated interfaces.
 
    * - ``sphinx.util.force_decode()``
      - 2.0
-     - 4.0
+     - 5.0
      - N/A
 
    * - ``sphinx.util.get_matching_docs()``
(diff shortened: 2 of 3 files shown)


stevepiercy:
@tk0miya thank you for this reversion.

I want to migrate `sphinx.ext.autodoc.AutoDirective` to `application.Sphinx.add_autodocumenter()`, but an [upgrade path is not clear to me](https://github.com/repoze/repoze.sphinx.autointerface/pull/17#issuecomment-841787898). If anyone has a before/after upgrade commit in their project to share with me, I would greatly appreciate it.

tk0miya:
Merged.

@stevepiercy Could you let me know in detail? AFAIK, `AutoDirective` class had been used to build the autodoc framework.  It's not for extension. So I can't imagine how the extension using it.

stevepiercy:
@tk0miya I believe that `AutoDirective` is used to test the extension `repoze.sphinx.autointerface`, but is not used by the extension itself. [`AutoDirective` appears in two places](https://github.com/repoze/repoze.sphinx.autointerface/search?q=AutoDirective) in the test code. I would like to understand how to rewrite these tests.

tk0miya:
We also have similar code named `do_autodoc()` here:
https://github.com/sphinx-doc/sphinx/blob/4.x/tests/test_ext_autodoc.py#L33-L43

I think you can do testing with this helper.

Issue #9330, opened 2021-06-13
LaTeX: :rst:dir:`versionchanged` with contents being a list will cause error during pdf build


jfbu:
**Describe the bug**

When trying to build Sphinx pdf doc on  4.x

```
! LaTeX Error: Something's wrong--perhaps a missing \item.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.26992 ...ged}{Changed in version 4.0.0: }\item {
                                                  }
```

(line number may not be right because actually got this on a branch adding extra text  to latex.rst)

**Your project**

Sphinx own doc

It is caused by 

```rest
   .. versionchanged:: 4.0.0

      - The font substitution commands added at ``2.0`` have been moved
        to the ``'fontsubstitution'`` key, as their presence here made
        it complicated for user to customize the value of ``'fontpkg'``.
      - The default font setting has changed: it still uses Times and
        Helvetica clones for serif and sans serif, but via better, more
        complete TeX fonts and associated LaTeX packages.  The
        monospace font has been changed to better match the Times clone.

```

i.e. a list after `versionchanged` in  latex.rst 

Not sure if actually allowed to use a display after `versionchanged`?

html output fine

- #9132

jfbu:
Edit: html output is **not** fine:

![Capture d’écran 2021-06-13 à 14 18 19](https://user-images.githubusercontent.com/2589111/121806947-49a70800-cc52-11eb-8b56-de1b4fd7f97d.png)

jfbu:
I don't know if feature or bug of domains.changeset.VersionChange() but adding the bug tag anyhow

tk0miya:
I found the reason of this bug and post a fix as #9333 now.

PR #9331, opened 2021-06-13
Workaround #9132 induced problem in building sphinx.pdf (#9330)

jfbu:
Not a hotfix to user, but to to allow pdf build of own Sphinx doc without error

### Relates
- #9330, #9132


jfbu:
I am closing this because root cause is better solved by #9333

PR #9333, opened 2021-06-13
Fix #9330: versionchanged causes error during pdf build

### LINKED ISSUE #9330: LaTeX: :rst:dir:`versionchanged` with contents being a list will cause error during pdf build

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- The versionchanged directive breaks doctree if its contents start with
non-paragraph element (ex. lists, tables, and so on).  It causes build
error in LaTeX.
- This inserts a paragraph if the first child of the contents is not a
paragraph not to break doctree.
- refs: #9330

### DIFF
--- a/sphinx/domains/changeset.py
+++ b/sphinx/domains/changeset.py
@@ -74,8 +74,10 @@ class VersionChange(SphinxDirective):
         if self.content:
             self.state.nested_parse(self.content, self.content_offset, node)
         classes = ['versionmodified', versionlabel_classes[self.name]]
-        if len(node):
-            if isinstance(node[0], nodes.paragraph) and node[0].rawsource:
+        if len(node) > 0 and isinstance(node[0], nodes.paragraph):
+            # the contents start with a paragraph
+            if node[0].rawsource:
+                # make the first paragraph translatable
                 content = nodes.inline(node[0].rawsource, translatable=True)
                 content.source = node[0].source
                 content.line = node[0].line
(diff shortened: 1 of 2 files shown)


sphinx/domains/changeset.py
```diff
@@ -84,10 +86,16 @@ def run(self) -> List[Node]:
 
             para = cast(nodes.paragraph, node[0])
             para.insert(0, nodes.inline('', '%s: ' % text, classes=classes))
+        elif len(node) > 0:
+            # the content does not starts with a paragraph
```
jfbu: s/starts/start/

And I am hesitating with s/content does/contents do/ but not a native speaker :)

sphinx/domains/changeset.py
```diff
@@ -74,8 +74,10 @@ def run(self) -> List[Node]:
         if self.content:
             self.state.nested_parse(self.content, self.content_offset, node)
         classes = ['versionmodified', versionlabel_classes[self.name]]
-        if len(node):
-            if isinstance(node[0], nodes.paragraph) and node[0].rawsource:
+        if len(node) > 0 and isinstance(node[0], nodes.paragraph):
+            # the content starts with a paragraph
```
jfbu: perhaps "contents start" (plural) but I am not native speaker

jfbu (approved):
LGTM with nits!

tk0miya:
I noticed this should be merged into 4.0.x branch. Rebasing.

sphinx/domains/changeset.py
tk0miya: Thanks! I'll follow your intuition :-)

tk0miya:
Merged. Thank you for your support as always :-)

Issue #9322, opened 2021-06-11
KeyError: 'domain' exception being thrown from PropagateDescDomain.run


bneradt:
**Describe the bug**
Apache Traffic Server docs are built using Sphinx. We noticed that when we upgraded to the 4.x release the build breaks with the following output:

```
# Sphinx version: 4.0.2
# Python version: 3.8.3 (CPython)
# Docutils version: 0.16 release
# Jinja2 version: 3.0.1
# Last messages:
#   writing output... [  2%] admin-guide/configuration/transparent-proxy/wccp-service-config.en
#   writing output... [  2%] admin-guide/configuring-traffic-server.en
#   writing output... [  2%] admin-guide/files/cache.config.en
#   writing output... [  3%] admin-guide/files/hosting.config.en
#   writing output... [  3%] admin-guide/files/index.en
#   writing output... [  3%] admin-guide/files/ip_allow.yaml.en
#   writing output... [  3%] admin-guide/files/logging.yaml.en
#   writing output... [  3%] admin-guide/files/parent.config.en
#   writing output... [  4%] admin-guide/files/plugin.config.en
#   writing output... [  4%] admin-guide/files/records.config.en
# Loaded extensions:
#   sphinx.ext.mathjax (4.0.2) from /root/.local/share/virtualenvs/doc-2zq9zEzg/lib/python3.8/site-packages/sphinx/ext/mathjax.py
#   sphinxcontrib.applehelp (1.0.2) from /root/.local/share/virtualenvs/doc-2zq9zEzg/lib/python3.8/site-packages/sphinxcontrib/applehelp/__init__.py
#   sphinxcontrib.devhelp (1.0.2) from /root/.local/share/virtualenvs/doc-2zq9zEzg/lib/python3.8/site-packages/sphinxcontrib/devhelp/__init__.py
#   sphinxcontrib.htmlhelp (2.0.0) from /root/.local/share/virtualenvs/doc-2zq9zEzg/lib/python3.8/site-packages/sphinxcontrib/htmlhelp/__init__.py
#   sphinxcontrib.serializinghtml (1.1.5) from /root/.local/share/virtualenvs/doc-2zq9zEzg/lib/python3.8/site-packages/sphinxcontrib/serializinghtml/__init__.py
#   sphinxcontrib.qthelp (1.0.3) from /root/.local/share/virtualenvs/doc-2zq9zEzg/lib/python3.8/site-packages/sphinxcontrib/qthelp/__init__.py
[...]

PR #9334, opened 2021-06-13
Fix sphinx-doc#9322: KeyError is raised on PropagateDescDomain transform

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- PropageteDescDomain applies the domain name from the "domain" attribute
of parent node (desc node) to the desc_signature node.  The structure
has longly generated by ObjectDescription.  But it must not be a new
rule.
- This allows to build document that contains non standard doctree.
- refs: #9322

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -17,6 +17,7 @@ Bugs fixed
 ----------
 
 * #9313: LaTeX: complex table with merged cells broken since 4.0
+* #9322: KeyError is raised on PropagateDescDomain transform
 
 Testing
 --------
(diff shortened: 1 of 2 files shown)


bneradt (approved):
I verified this patch address the Traffic Server doc build issue (#9322). I reproduced #9322 locally with `Sphinx==4.0.2` and `Python 3.8.3`. The KeyError for `domain` was raised. I then applied this patch locally and rebuilt the ATS docs. The docs build succeeded.

Thank you for turning a patch around so quickly.

tk0miya:
Thank you for your confirmation :-) Just merged.

PR #9337, opened 2021-06-14
Implement new search shortcuts.

### LINKED ISSUE #691: Allow keyboard shortcut `/` to focus on search
Various web services, notably GMail and GitHub, allow the keyboard shortcut `/` for focusing on the search box. It would be nice if Sphinx templates would do the same.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/691
- Originally reported by: Ram Rachum
- Originally created at: 2011-05-03T13:19:11.852

marxin:
Implement new shortcuts:
    
- "/" for Focus search bar
- "ESC" - clear highlighted text
    
Fixes #691.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,9 @@ Features added
 
 * #9494, #9456: html search: Add a config variable
   :confval:`html_show_search_summary` to enable/disable the search summaries
+* #9337: HTML theme, add option ``enable_search_shortcuts`` that enables :kbd:'/' as
+         a Quick search shortcut and :kbd:`Esc` shortcut that
+         removes search highlighting.
 
 Bugs fixed
 ----------
--- a/doc/usage/theming.rst
+++ b/doc/usage/theming.rst
@@ -158,9 +158,18 @@ These themes are:
     dimension string such as '70em' or '50%'. Use 'none' if you don't
     want a width limit. Defaults may depend on the theme (often 800px).
 
-  - **navigation_with_keys** (true or false): Allow navigating to the
-    previous/next page using the keyboard's left and right arrows.  Defaults to
-    ``False``.
+  - **navigation_with_keys** (true or false): Allow navigating
+    with the following keyboard shortcuts:
+
+    - :kbd:`Left arrow`: previous page
+    - :kbd:`Right arrow`: next page
+
+    Defaults to ``False``.
+
+  - **enable_search_shortcuts** (true or false): Allow jumping to the search box
+    with :kbd:`/` and allow removal of search highlighting with :kbd:`Esc`.
+
+    Defaults to ``True``.
 
   - **globaltoc_collapse** (true or false): Only expand subsections
     of the current document in ``globaltoc.html``
--- a/sphinx/themes/basic/static/doctools.js
+++ b/sphinx/themes/basic/static/doctools.js
@@ -154,9 +154,7 @@ var Documentation = {
     this.fixFirefoxAnchorBug();
     this.highlightSearchWords();
     this.initIndexTable();
-    if (DOCUMENTATION_OPTIONS.NAVIGATION_WITH_KEYS) {
-      this.initOnKeyListeners();
-    }
+    this.initOnKeyListeners();
   },
 
   /**
@@ -269,6 +267,13 @@ var Documentation = {
     window.history.replaceState({}, '', url);
   },
 
+   /**
+   * helper function to focus on search bar
+   */
+  focusSearchBar : function() {
+    $('input[name=q]').first().focus();
+  },
+
   /**
    * make the url absolute
    */
@@ -291,27 +296,54 @@ var Documentation = {
   },
 
   initOnKeyListeners: function() {
+    // only install a listener if it is really needed
+    if (!DOCUMENTATION_OPTIONS.NAVIGATION_WITH_KEYS &&
+        !DOCUMENTATION_OPTIONS.ENABLE_SEARCH_SHORTCUTS)
+        return;
+
     $(document).keydown(function(event) {
       var activeElementType = document.activeElement.tagName;
       // don't navigate when in search box, textarea, dropdown or button
       if (activeElementType !== 'TEXTAREA' && activeElementType !== 'INPUT' && activeElementType !== 'SELECT'
-          && activeElementType !== 'BUTTON' && !event.altKey && !event.ctrlKey && !event.metaKey
-          && !event.shiftKey) {
-        switch (event.keyCode) {
-          case 37: // left
-            var prevHref = $('link[rel="prev"]').prop('href');
-            if (prevHref) {
-              window.location.href = prevHref;
-              return false;
-            }
-            break;
-          case 39: // right
-            var nextHref = $('link[rel="next"]').prop('href');
-            if (nextHref) {
-              window.location.href = nextHref;
-              return false;
-            }
-            break;
+          && activeElementType !== 'BUTTON') {
+        if (event.altKey || event.ctrlKey || event.metaKey)
+          return;
+
+          if (!event.shiftKey) {
+            switch (event.key) {
+              case 'ArrowLeft':
+                if (!DOCUMENTATION_OPTIONS.NAVIGATION_WITH_KEYS)
+                  break;
+                var prevHref = $('link[rel="prev"]').prop('href');
+                if (prevHref) {
+                  window.location.href = prevHref;
+                  return false;
+                }
+                break;
+              case 'ArrowRight':
+                if (!DOCUMENTATION_OPTIONS.NAVIGATION_WITH_KEYS)
+                  break;
+                var nextHref = $('link[rel="next"]').prop('href');
+                if (nextHref) {
+                  window.location.href = nextHref;
+                  return false;
+                }
+                break;
+              case 'Escape':
+                if (!DOCUMENTATION_OPTIONS.ENABLE_SEARCH_SHORTCUTS)
+                  break;
+                Documentation.hideSearchWords();
+                return false;
+          }
+        }
+
+        // some keyboard layouts may need Shift to get /
+        switch (event.key) {
+          case '/':
+            if (!DOCUMENTATION_OPTIONS.ENABLE_SEARCH_SHORTCUTS)
+              break;
+            Documentation.focusSearchBar();
+            return false;
         }
       }
     });
--- a/sphinx/themes/basic/static/documentation_options.js_t
+++ b/sphinx/themes/basic/static/documentation_options.js_t
@@ -9,5 +9,6 @@ var DOCUMENTATION_OPTIONS = {
     HAS_SOURCE: {{ has_source|lower }},
     SOURCELINK_SUFFIX: '{{ sourcelink_suffix }}',
     NAVIGATION_WITH_KEYS: {{ 'true' if theme_navigation_with_keys|tobool else 'false'}},
-    SHOW_SEARCH_SUMMARY: {{ 'true' if show_search_summary else 'false' }}
+    SHOW_SEARCH_SUMMARY: {{ 'true' if show_search_summary else 'false' }},
+    ENABLE_SEARCH_SHORTCUTS: {{ 'true' if enable_search_shortcuts|tobool else 'true'}},
 };
--- a/sphinx/themes/basic/theme.conf
+++ b/sphinx/themes/basic/theme.conf
@@ -10,6 +10,7 @@ sidebarwidth = 230
 body_min_width = 450
 body_max_width = 800
 navigation_with_keys = False
+enable_search_shortcuts = True
 globaltoc_collapse = true
 globaltoc_includehidden = false
 globaltoc_maxdepth =


mgeier:
There are already keyboard shortcuts in the `basic` theme using [accesskey](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey), for example using `n` for *next* and `p` for *previous* page. Some derived HTML themes have (I guess mostly inadvertently) disabled them though.

I think it would be confusing to add another mechanism.

Of course another mechanism has already been added long ago with the left and right arrow keys being mapped by some JavaScript (but this is disabled by default in most themes).

I think if shortcuts with "normal" letters are added, they should be added with the `accesskey` technology.

I guess an exception could be made for the "special" character `/`, but I'm not sure whether that's actually a good idea.

In general, I'm not a fan of running some JavaScript that grabs my keypresses, mostly because I'm already using a browser extension (in my case https://github.com/tridactyl/tridactyl, but there are several similar ones) that watches my keypresses and reacts to `/`. Checking for this key in Sphinx's JavaScript will likely break my current `/` key functionality.

In my own HTML theme I'm providing a `s` shortcut via `accesskey` and I've added a more prominent link to remove the highlighted text (see for example https://insipid-sphinx-theme.readthedocs.io/en/0.2.6/intro.html?highlight=accesskey#features). But I guess the suggested `c` shortcut could also be added with `accesskey`, right?

BTW, the suggested change in this PR will also be disabled in most themes unless the `navigation_with_keys` theme option is `True`:

https://github.com/sphinx-doc/sphinx/blob/355937aef80246750fe4bed2bf855bd533ef18b2/sphinx/themes/basic/static/doctools.js#L157-L159

marxin:
> There are already keyboard shortcuts in the `basic` theme using [accesskey](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey), for example using `n` for _next_ and `p` for _previous_ page. Some derived HTML themes have (I guess mostly inadvertently) disabled them though.

Good to know. It seems like a pretty standard mechanism, however, I see some limitations:
1. In Chrome `p` accesskey does not work for me for some reason
1. In Firefox one needs to use ugly `ALT` + 'SHIFT' + 'key' and it only shows URL preview and one needs to click enter.

My inspiration was pages like Gmail, Trello or GitHub. Where shortcuts are pretty simple and easy to use.

> 
> I think it would be confusing to add another mechanism.
> 
> Of course another mechanism has already been added long ago with the left and right arrow keys being mapped by some JavaScript (but this is disabled by default in most themes).

Yes, one has to enable it explicitly.

> 
> I think if shortcuts with "normal" letters are added, they should be added with the `accesskey` technology.
> 
> I guess an exception could be made for the "special" character `/`, but I'm not sure whether that's actually a good idea.
> 
> In general, I'm not a fan of running some JavaScript that grabs my keypresses, mostly because I'm already using a browser extension (in my case https://github.com/tridactyl/tridactyl, but there are several similar ones) that watches my keypresses and reacts to `/`. Checking for this key in Sphinx's JavaScript will likely break my current `/` key functionality.

That's pretty uncommon use to me. I prefer native shortcuts provided by a web page.

> 
> In my own HTML theme I'm providing a `s` shortcut via `accesskey` and I've added a more prominent link to remove the highlighted text (see for example https://insipid-sphinx-theme.readthedocs.io/en/0.2.6/intro.html?highlight=accesskey#features). But I guess the suggested `c` shortcut could also be added with `accesskey`, right?

I like the other navigation shortcuts: `Up`, `Index` and I could also imagine going to `Home`.

> 
> BTW, the suggested change in this PR will also be disabled in most themes unless the `navigation_with_keys` theme option is `True`:

We can consider coming up with a more common name (shortcuts) and possibly enable it.

mgeier:
> In Chrome `p` accesskey does not work for me for some reason

Hmmm, it works for me on Chromium.
Which theme are you using?

> In Firefox one needs to use ugly `ALT` + 'SHIFT' + 'key'

I very quickly got used to this, but I agree that it's not very straightforward.

What's worse, it's hard to discover, it would be great if something could be done about this.

> and it only shows URL preview and one needs to click enter.

I don't understand.
On Firefox, it also works fine for me, I don't have to hit Enter or click anywhere at any point.

> I prefer native shortcuts provided by a web page.

Yeah, it's definitely a matter of preference.

The question is: is it worth how much it helps users (like you) compared to how much it hurts other users (like me)?

I think adding more shortcuts with `accesskey` wouldn't hurt anyone.

marxin:
I would like to extend the group of people who discuss the functionality.
Adding @TimKam who added the shortcuts and others involved in the original PR: @shimizukawa, @lehmannro, @tk0miya.

TimKam:
Nit-pick: consistent use of quotes -> I'd say single quotes instead of double quotes

What are the remaining open questions?
I don't have a strong opinion on default vs. not default, but if we change the name of the config option, we could also consider an object: because on the long run, too many short cuts that are all activated by default can be confusing, and the best option could be an object that defaults to a good compromise.

marxin:
> Nit-pick: consistent use of quotes -> I'd say single quotes instead of double quotes

Sure, adjusted.

> 
> What are the remaining open questions?

Apparently, @mgeier didn't like the idea. However, my patch only extends the existing useful shortcuts.

> I don't have a strong opinion on default vs. not default, but if we change the name of the config option, we could also consider an object: because on the long run, too many short cuts that are all activated by default can be confusing, and the best option could be an object that defaults to a good compromise.

Is it something that can be addressed in the future once this pull request is merged?
Thanks.

tk0miya:
IMO, `/` is a commonly used shortcut. So +1 for adding. On the other hand, clearing highlight is not common. So -0 for it.

Note: I don't have an opinion to implement them because I'm not familiar with HTML and JS.

tk0miya:
Please merge the HEAD of 4.x branch to your PR to fix the CI error. It will be fixed soon.

sphinx/themes/basic/static/doctools.js
```diff
@@ -294,20 +294,26 @@ var Documentation = {
...
               window.location.href = nextHref;
               return false;
             }
+          case '/':
+          case 's':
+            $('input[name=q]').focus();
+            return false;
+          case 'c':
+            newurl = window.location.href.replace(new RegExp('\\?highlight=[^#]*'), '');
+            window.location.href = newurl;
```
mgeier: There is already existing code for this functionality:

https://github.com/sphinx-doc/sphinx/blob/c66b3ad519ed91e79205ee973bfdba17fc06453f/sphinx/themes/basic/static/doctools.js#L261-L267

What you are suggesting is causing a page reload, which I think isn't desirable.

mgeier:
For illustration purposes, I've created #9539, showing hot `accesskey` can be used to hide search matches. However, I don't think that any shortcut key is necessary for that functionality.

jakobandersen:
Whatever the shortcut mechanism is (didn't know about accessKey) I like the addition of both search an clearing highlight. Though, it would be nice if it would clear the actual highlight, clear the highlight part of the URL, and all while not refreshing the page.
I often search for something in order to find an entity to shared a link to, but the perma-link contains the ``highlight=...``-part which then needs to manually be removed.

I would also suggest to add a ? shortcut which shows an overview of shortcuts, e.g., like Gmail has.

marxin:
> Please merge the HEAD of 4.x branch to your PR to fix the CI error. It will be fixed soon.

I've just done that.

sphinx/themes/basic/static/doctools.js
marxin: You are right, I'm not using the existing JS functionality. Good point!

marxin:
> For illustration purposes, I've created #9539, showing hot `accesskey` can be used to hide search matches. However, I don't think that any shortcut key is necessary for that functionality.

I think it's handy as e.g. `sphinx_rtd_theme` template does not show the `Hide Search Matches` button.

marxin:
> Whatever the shortcut mechanism is (didn't know about accessKey) I like the addition of both search an clearing highlight. Though, it would be nice if it would clear the actual highlight, clear the highlight part of the URL, and all while not refreshing the page.

I've just one that, browser history is updated accordingly.

> I often search for something in order to find an entity to shared a link to, but the perma-link contains the `highlight=...`-part which then needs to manually be removed.

That's exactly my use case why I came up with the pull request.

> 
> I would also suggest to add a ? shortcut which shows an overview of shortcuts, e.g., like Gmail has.

Great idea. Does Sphinx have any Overlay that can be used similarly to what Gmail or Github provides?
One small note, I changed the shortcut `c` to `h` (Hide Search Matches), where I think using uppercase shortcuts is not a good practice.

TimKam:
Could you also update the documentation? See: https://www.sphinx-doc.org/en/master/usage/theming.html

I don't think Sphinx has any overlay feature, and I'd suggest that the `?` feature may be addressed in a future PR.

marxin:
> Could you also update the documentation? See: https://www.sphinx-doc.org/en/master/usage/theming.html

I've just done that.

> I don't think Sphinx has any overlay feature, and I'd suggest that the `?` feature may be addressed in a future PR.

Sure, works for me.

tk0miya (approved):

sphinx/themes/basic/static/doctools.js
```diff
@@ -294,21 +294,28 @@ var Documentation = {
...
               return false;
             }
-            break;
+          case '/':
+          case 's':
+            $('input[name=q]').focus();
+            return false;
+          case 'h':
+            newurl = window.location.href.replace(new RegExp('\\?highlight=[^#]*'), '');
+            history.pushState({}, null, newurl);
```
mgeier: Thanks for the update, this works really well, without re-loading the page!

Would it be possible to move this into the `hideSearchWords()` function?

This way, the URL would also be modified when clicking the "hide search matches" link, which I think would be great!

This could even be a separate (less contentious) PR, because it really doesn't have anything to do with shortcuts.

mgeier:
I think it would be a bad idea to merge this as is.

I personally don't like it when JavaScript grabs my key presses (because I'm using them for something else), but this is only one part of the problem.

Even for people who maybe would like this exact implementation, they would still not get it in most cases, because `navigation_with_keys` is disabled by default.

Using `accesskey` would have none of those problems. And that's what Sphinx traditionally uses.

And of course theme authors are free to support additional JavaScript keyboard shortcuts as they see fit.

----

BTW, having this in the `navigation_with_keys` doesn't make too much sense, because neither focusing the search field nor removing search hits has anything to do with "navigation", right?

----

I *do* like the URL manipulation though, I think it would be great if this could be added to `hideSearchWords()` (see https://github.com/sphinx-doc/sphinx/pull/9337#discussion_r687059362).

sphinx/themes/basic/static/doctools.js
marxin: I welcome the idea and I've pushed a commit that does that.

marxin:
> I think it would be a bad idea to merge this as is.

Heh.

> 
> I personally don't like it when JavaScript grabs my key presses (because I'm using them for something else), but this is only one part of the problem.

I have an opposite idea as `accesskey` seems like an interesting generic approach, but the shortcuts are complicated and I'm not aware of a webpage using them (Gmail, Trello, Github). The elegance of shortcuts is that they are short and quick to use.

> 
> Even for people who maybe would like this exact implementation, they would still not get it in most cases, because `navigation_with_keys` is disabled by default.
> 
> Using `accesskey` would have none of those problems. And that's what Sphinx traditionally uses.

But it has other problems (not being known, more complicated to use).

> 
> And of course theme authors are free to support additional JavaScript keyboard shortcuts as they see fit.
> 
> BTW, having this in the `navigation_with_keys` doesn't make too much sense, because neither focusing the search field nor removing search hits has anything to do with "navigation", right?

Sure, but it does not make sense to rename it or come up with a different one. I can read  `navigation_with_keys` as a synonym for shortcuts or hotkeys.

sphinx/themes/basic/static/doctools.js
mgeier: > This could even be a separate (less contentious) PR, because it really doesn't have anything to do with shortcuts.

I've just created such a PR: #9551.

sphinx/themes/basic/static/doctools.js
```diff
@@ -264,6 +264,8 @@ var Documentation = {
   hideSearchWords : function() {
     $('#searchbox .highlight-link').fadeOut(300);
     $('span.highlighted').removeClass('highlighted');
+    newurl = window.location.href.replace(new RegExp('\\?highlight=[^#]*'), '');
+    history.pushState({}, null, newurl);
```
mgeier: This creates a new history entry, which makes sense, but moving back and forth in the history does *not* restore the highlighting. In #9551, I'm suggesting to use `history.replaceState()` instead.

mgeier:
@marxin 

> I think using uppercase shortcuts is not a good practice.

I agree.

I'm not sure if you are referring to #9539, but in case you do:

Browsers seem to ignore upper/lower case, see https://github.com/whatwg/html/issues/4799.

I've used an uppercase `"H"` because that's what the rest of the Sphinx code seems to use. But I guess a case could be made to change all the `accesskey`s to lowercase at some point.

mgeier:
I've just created  #9552 for an alternative way to focus the search box (using `accesskey`).

sphinx/themes/basic/static/doctools.js
marxin: All right, I do support your change in #9551 and I've just removed the history change in this pull request.

marxin:
> I've just created #9552 for an alternative way to focus the search box (using `accesskey`).

I think it's complementary to this pull request and people can select what's preferred way of shortcuts.

mgeier:
I've just tried the `/` shortcut with this PR and it does *not* work on Firefox (version 88.0.1, no plugins). Pressing `/` opens a "Quick find" search bar at the bottom of the window, provided by Firefox.

On Chromium (version 90.0.4430.212) it doesn't work either, but here just nothing happens when hitting `/`.

The `s` shortcut works fine on both browsers.

For comparison, pressing `/` on https://github.com/ focuses Github's search field, as expected, in both browsers.

I've tested this on Linux, BTW.

marxin:
> I've just tried the `/` shortcut with this PR and it does _not_ work on Firefox (version 88.0.1, no plugins). Pressing `/` opens a "Quick find" search bar at the bottom of the window, provided by Firefox.
> 
> On Chromium (version 90.0.4430.212) it doesn't work either, but here just nothing happens when hitting `/`.
> 
> The `s` shortcut works fine on both browsers.

Doing the same testing on openSUSE Linux, it works for me both in Chromium (92) and Firefox (90). There's a demo project that contains this PR: https://splichal.eu/tmp/html/

marxin:
Can we please merge this pull request?

sphinx/themes/basic/static/doctools.js
```diff
@@ -294,21 +294,26 @@ var Documentation = {
...
-          case 39: // right
+          case 'ArrowRight':
             var nextHref = $('link[rel="next"]').prop('href');
             if (nextHref) {
               window.location.href = nextHref;
               return false;
             }
-            break;
+          case '/':
+          case 's':
```
pradyunsg: Minor request: Can we exclude `s` from this? Just `/` is sufficient IMO.

marxin: Resolved.

marxin:
Updated @pradyunsg comment, can we please merge it now?

pradyunsg:
Thanks for the update! :)

I don't have the merge bit, but I imagine the maintainers will take a look at this in time for 4.3.0 release. :)

sphinx/themes/basic/static/doctools.js
```diff
@@ -294,21 +294,25 @@ var Documentation = {
...
-          case 39: // right
+          case 'ArrowRight':
             var nextHref = $('link[rel="next"]').prop('href');
             if (nextHref) {
               window.location.href = nextHref;
               return false;
             }
-            break;
+          case '/':
+            $('input[name=q]').focus();
```
pradyunsg: How does this behave, if there's multiple matches for this selector?

marxin: There's supposed to be only one such element. If there are more, all of them will be focused, which will cause the last one being focused.

marxin:
Sure, I hope it will be selected for the next release as it already received approval from @tk0miya.

sphinx/themes/basic/static/doctools.js
pradyunsg: > There's supposed to be only one such element.

This is a new restriction being added for downstream themes, and shouldn't be done lightly IMO. :)

marxin: All right, I use now `.first()` for matched results.

pradyunsg:
A couple of suggestions having tried this out:

- Let's change the clear highlights shortcut to the escape key. I've seen a couple of folks do this when they opened a site with highlights, and have done this myself as well.
- Would any one be opposed to dropping the `s` key for search right now? I think it wouldn't be _that_ difficult to add it in later, and keeping things minimal seems like a good idea (since downstream themes/users can only add to what Sphinx provides here and not remove functionality + it's easier to add than remove for Sphinx as well).

FWIW, @marxin I do appreciate you being receptive to feedback. I notice that this was pushed back from the release, which, I'm guessing is because the Windows CI failed. It would be great if you could update this PR to fix the CI failure. If you're no longer interested in driving this, I'd be happy to pick up the baton on this.

marxin:
> * Let's change the clear highlights shortcut to the escape key. I've seen a couple of folks do this when they opened a site with highlights, and have done this myself as well.

Changed that, I like the suggested change.

> * Would any one be opposed to dropping the `s` key for search right now? I think it wouldn't be _that_ difficult to add it in later, and keeping things minimal seems like a good idea (since downstream themes/users can only add to what Sphinx provides here and not remove functionality + it's easier to add than remove for Sphinx as well).

I've already done that in the previous update of the patch (so not only `/` is used for searching).

> 
> FWIW, @marxin I do appreciate you being receptive to feedback. I notice that this was pushed back from the release, which, I'm guessing is because the Windows CI failed. It would be great if you could update this PR to fix the CI failure. If you're no longer interested in driving this, I'd be happy to pick up the baton on this.

All CI passed now.

pradyunsg:
You might want to update the PR description! :)

marxin:
> You might want to update the PR description! :)

Sure. Good point.

mgeier:
I still very much dislike a web page's JavaScript to grab my `/` button press and I would like to be able to disable this for my Sphinx theme.

However, I *do* like using the left/right arrow keys for navigation, so I like having the setting `navigation_with_keys` switched on.

The `/` feature should use a separate setting. As mentioned above (https://github.com/sphinx-doc/sphinx/pull/9337#issuecomment-897044840), the name `navigation_with_keys` wouldn't make sense anyway.

TheTripleV:
I'd prefer `/` as the shortcut key. Mozilla's web docs for accesskeys ([linked](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey) previously) even uses it. Also Google and Youtube.

marxin:
> I'd prefer `/` as the shortcut key. Mozilla's web docs for accesskeys ([linked](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey) previously) even uses it. Also Google and Youtube.

Note this pull request does not add any `accesskeys`, it's done in #9552. Can you please explain how `/` (alone) is handled as part of `accesskeys`?

TheTripleV:
I haven't looked into how accesskeys work. I just saw that accesskeys were offered as an alternative to `/` and noticed that the linked documentation itself went out of its way to implement `/` as a search shortcut.

marxin:
Ah, got it. Which is a proof that `/` or (`f`) are wildly used as a quick shortcut.

mgeier (changes requested):
Just to make sure my doubts (that I've already mentioned above) don't get overlooked ...

I of course cannot and will not object to adding the suggested keyboard shortcuts. Whether they are useful or harmful is a matter of taste.

However, it is objectively wrong to call the option to enable them `navigation_with_keys`, because this has nothing to do with "navigation".

Before this PR is merged (and I don't recommend that it should be!), it should introduce a new option with a more appropriate name.

marxin:
> However, it is objectively wrong to call the option to enable them `navigation_with_keys`, because this has nothing to do with "navigation".

Well, I don't mind adding a new configuration option. But I would like to see an opinion from the project maintainers: @jakobandersen and @tk0miya? The pull request is there for almost 6 months and I would like to see a resolution.

doc/usage/theming.rst
```diff
@@ -159,8 +159,9 @@ These themes are:
     want a width limit. Defaults may depend on the theme (often 800px).
 
   - **navigation_with_keys** (true or false): Allow navigating to the
```
jakobandersen:
```suggestion
  - **navigation_with_keys** (true or false): Allow navigating with the following keyboard shortcuts:
```

doc/usage/theming.rst
```diff
@@ -159,8 +159,9 @@ These themes are:
...
-    previous/next page using the keyboard's left and right arrows.  Defaults to
-    ``False``.
+    previous/next page using the keyboard's left and right arrows.
+    Apart from that, '/' jumps to Quick search and Escape key
+    hides search matches. Defaults to ``False``.
```
jakobandersen:
```suggestion

    - :kbd:`Left arrow`: previous page
    - :kbd:`Right arrow`: next page
    - :kbd:`/`: jump to Quick search
    - :kbd:`Esc`: hide search matches
    
    Defaults to ``False``.
```

jakobandersen (changes requested):

jakobandersen:
I don't mind putting this under ``navigation_with_keys`` for now (it's a bit more abstract, but ``/`` and ``Esc`` are still navigating the documentation site).
I suggest to merge this for the next minor release, or possibly major release if it is considered breaking to add the two extra keys.

In the future perhaps we could make a more complex configuration variable, e.g., a dictionary where the user can specify which shortcuts to enable. The work in https://github.com/sphinx-doc/sphinx/pull/9552 could then be part of this, where the user can specify which style of shortcuts to enable.

marxin:
> In the future perhaps we could make a more complex configuration variable, e.g., a dictionary where the user can specify which shortcuts to enable. The work in #9552 could then be part of this, where the user can specify which style of shortcuts to enable.

Sure, I'm willing to work on that once this pull request is merged.
Thanks for the suggestions.

marxin:
I noticed the 4.4.0 milestone is approaching. May I please ping this pull request? I've been waiting for quite some time :( ...

mgeier:
@marxin Do you really want this to be disabled by default?

Currently, this PR implements just about the worst possible situation (a classical lose-lose situation):

* people who actually like this feature will not get it by default
* people who dislike this feature but have `navigation_with_keys` already enabled (because they like to have arrow-key navigation), will have no way of disabling this new feature

If this new feature were to be added to a new configuration parameter, this new configuration parameter could be enabled by default.

I still don't like this feature, but at least this way I would be able to deactivate it in my theme (but users of my theme would still be able to re-activate it if they happen to like it).

marxin:
> If this new feature were to be added to a new configuration parameter, this new configuration parameter could be enabled by default.

All right, so can you please suggest configuration variable names which I should use?

jakobandersen:
I have tried adding separate options for enabling ``/`` and ``Esc``: ``search_with_keys`` and ``remove_highlight_with_keys``. See https://github.com/jakobandersen/sphinx/tree/new-shortcuts
I also found that the check for ``/`` was too restrictive as the shift key is needed on some keyboard layouts to get it (e.g., Danish layout, ``shift+7``).
@marxin, if you like the changes then feel free to pull them into your branch.
Some TODOs if so:
- @mgeier, would this be ok with you (modulo the lack of access-key support which I think can be added separately?)?
- are the options names ok?
- I made the two new ones default to enabled. Is that ok?
- ``Esc`` doesn't rewrite the URL, I guess the PR needs a rebase to pick that up.
- update CHANGES

marxin:
> I have tried adding separate options for enabling `/` and `Esc`: `search_with_keys` and `remove_highlight_with_keys`. See https://github.com/jakobandersen/sphinx/tree/new-shortcuts

Thank you very much. I included that indeed.

> I also found that the check for `/` was too restrictive as the shift key is needed on some keyboard layouts to get it (e.g., Danish layout, `shift+7`).

Are you really sure, now it does not work for me on the English keyboard :/ Can you please test `/` above `8` at the numeric keyboard? And please check how GitHub works for you?

> @marxin, if you like the changes then feel free to pull them into your branch. Some TODOs if so:
> 
> * @mgeier, would this be ok with you (modulo the lack of access-key support which I think can be added separately?)?
> * are the options names ok?
> * I made the two new ones default to enabled. Is that ok?

I hope so, as @mgeier expressed he can disable it in his theme.

> * `Esc` doesn't rewrite the URL, I guess the PR needs a rebase to pick that up.

That's a change that got to `4.x`, but is missing at `master`: #10102

> * update CHANGES

Added.

jakobandersen:
> > I also found that the check for `/` was too restrictive as the shift key is needed on some keyboard layouts to get it (e.g., Danish layout, `shift+7`).
> 
> Are you really sure, now it does not work for me on the English keyboard :/ Can you please test `/` above `8` at the numeric keyboard? And please check how GitHub works for you?

Doh, there was a stupid logic error. It should work now (e.g. cherry-pick the latest commit in my branch).

CHANGES
```diff
@@ -46,6 +46,10 @@ Features added
 * #9391: texinfo: improve variable in ``samp`` role
 * #9578: texinfo: Add :confval:`texinfo_cross_references` to disable cross
   references for readability with standalone readers
+* #9337: Add :confval:`search_with_keys` that enables :kbd:'/' as
```
jakobandersen:
```suggestion
* #9337: HTML theme, add option ``search_with_keys`` that enables :kbd:'/' as
```

CHANGES
```diff
@@ -46,6 +46,10 @@ Features added
...
+         a Quick search shortcut.
+         Add :confval:`remove_highlight_with_keys` that enables :kbd:`Esc` shortcut
```
jakobandersen:
```suggestion
         Add option ``remove_highlight_with_keys`` that enables :kbd:`Esc` shortcut
```

CHANGES
jakobandersen: They are technically not "configuration variables" as that would be those in ``conf.py``.

CHANGES
jakobandersen: Same as above.

jakobandersen (changes requested):

jakobandersen:
Cherry-picking b3e1b139cc435883e409fffae81d941e5bcadb50 and the minor fixes to CHANGES, then I'm good :-).

marxin:
> Cherry-picking [b3e1b13](https://github.com/sphinx-doc/sphinx/commit/b3e1b139cc435883e409fffae81d941e5bcadb50) and the minor fixes to CHANGES, then I'm good :-).

Thanks for the help!

sphinx/themes/basic/static/doctools.js
```diff
@@ -153,10 +153,7 @@ var Documentation = {
   init : function() {
     this.fixFirefoxAnchorBug();
     this.highlightSearchWords();
-    this.initIndexTable();
```
mgeier, line 156: I'm not sure what this does, but why has it been removed?

sphinx/themes/basic/static/doctools.js
```diff
@@ -291,27 +288,52 @@ var Documentation = {
...
       if (activeElementType !== 'TEXTAREA' && activeElementType !== 'INPUT' && activeElementType !== 'SELECT'
-          && activeElementType !== 'BUTTON' && !event.altKey && !event.ctrlKey && !event.metaKey
-          && !event.shiftKey) {
-        switch (event.keyCode) {
-          case 37: // left
-            var prevHref = $('link[rel="prev"]').prop('href');
-            if (prevHref) {
-              window.location.href = prevHref;
+          && activeElementType !== 'BUTTON') {
+        if (!event.altKey && !event.ctrlKey && !event.metaKey && !event.shiftKey) {
```
mgeier: It would probably be clearer if `altKey`, `ctrlKey` and `metaKey` stayed in the outer `if` and this only contained `shiftKey`, which this is all about, right?

sphinx/themes/basic/static/doctools.js
```diff
@@ -291,27 +288,52 @@ var Documentation = {
...
-              window.location.href = nextHref;
+          }
+        }
+        if (!event.altKey && !event.ctrlKey && !event.metaKey) {
+          // some keyboard layouts need Shift to get /
+          switch (event.key) {
+            case '/':
+              if (!DOCUMENTATION_OPTIONS.SEARCH_WITH_KEYS)
+                break;
+              $('input[name=q]').first().focus();
```
mgeier: I'm not sure if this would work reliably, but if this line were factored out into a function (similar to `hideSearchWords()` above?), HTML themes could probably overwrite this function if they need different behavior?

mgeier:
Thanks @marxin and @jakobandersen, this is now much better, but there are still a few things that probably could be improved.

I'm not sure if it's necessary to have two separate options for `/` and `Esc`. What do you think?

And naming is of course hard.
The name `search_with_keys` is a bit weird, because to search something, you'll have to type the search term, which you typically do with keys, right?
So "search" is always "search with keys", isn't it?

What about using `search_shortcuts` or `enable_search_shortcuts` or something like that?
Those would also allow to combine `/` and `Esc`, which are both shortcuts related to search.

Another (minor) problem with both `search_with_keys` and `remove_highlight_with_keys` is the use of the plural "keys", but it's only a single key in each case (except if `Shift` is needed to get `/`). In the future, more keys might be added, so using the singular "key" also won't be a good idea.

I've mentioned above (https://github.com/sphinx-doc/sphinx/pull/9337#issuecomment-899069664) that this feature is broken for me ... now it works! I guess this is because I'm also using a keyboard where `Shift` has to be pressed to get `/`. Thanks for finding and fixing this @jakobandersen!

Finally, it might be interesting to check if this feature plays well with https://github.com/readthedocs/readthedocs-sphinx-search.

sphinx/themes/basic/static/doctools.js
marxin: Good points, fixed!

sphinx/themes/basic/static/doctools.js
marxin: Reworked.

sphinx/themes/basic/static/doctools.js
marxin: > I'm not sure if it's necessary to have two separate options for `/` and `Esc`. What do you think?

Yes, I agree that one option would be better.

> 
> And naming is of course hard. The name `search_with_keys` is a bit weird, because to search something, you'll have to type the search term, which you typically do with keys, right? So "search" is always "search with keys", isn't it?
> 
> What about using `search_shortcuts` or `enable_search_shortcuts` or something like that? Those would also allow to combine `/` and `Esc`, which are both shortcuts related to search.
> 
> Another (minor) problem with both `search_with_keys` and `remove_highlight_with_keys` is the use of the plural "keys", but it's only a single key in each case (except if `Shift` is needed to get `/`). In the future, more keys might be added, so using the singular "key" also won't be a good idea.

All fixed as I chose `enable_search_shortcuts` name now.

> 
> I've mentioned above ([#9337 (comment)](https://github.com/sphinx-doc/sphinx/pull/9337#issuecomment-899069664)) that this feature is broken for me ... now it works! I guess this is because I'm also using a keyboard where `Shift` has to be pressed to get `/`. Thanks for finding and fixing this @jakobandersen!

Now it should work both on layouts where one needs Shift and one doesn't.

> 
> Finally, it might be interesting to check if this feature plays well with https://github.com/readthedocs/readthedocs-sphinx-search.

Can you please test it for me?

Anyway, thanks for useful comments.

marxin:
> I'm not sure if it's necessary to have two separate options for `/` and `Esc`. What do you think?

Yes, I agree that one option would be better.

> 
> And naming is of course hard. The name `search_with_keys` is a bit weird, because to search something, you'll have to type the search term, which you typically do with keys, right? So "search" is always "search with keys", isn't it?
> 
> What about using `search_shortcuts` or `enable_search_shortcuts` or something like that? Those would also allow to combine `/` and `Esc`, which are both shortcuts related to search.
> 
> Another (minor) problem with both `search_with_keys` and `remove_highlight_with_keys` is the use of the plural "keys", but it's only a single key in each case (except if `Shift` is needed to get `/`). In the future, more keys might be added, so using the singular "key" also won't be a good idea.

All fixed as I chose `enable_search_shortcuts` name now.

> 
> I've mentioned above ([#9337 (comment)](https://github.com/sphinx-doc/sphinx/pull/9337#issuecomment-899069664)) that this feature is broken for me ... now it works! I guess this is because I'm also using a keyboard where `Shift` has to be pressed to get `/`. Thanks for finding and fixing this @jakobandersen!

Now it should work both on layouts where one needs Shift and one doesn't.

> 
> Finally, it might be interesting to check if this feature plays well with https://github.com/readthedocs/readthedocs-sphinx-search.

Can you please test it for me?

Anyway, thanks for useful comments.

sphinx/themes/basic/static/doctools.js
marxin: Done, function `focusSearchBar` was added.

mgeier:
> > Finally, it might be interesting to check if this feature plays well with https://github.com/readthedocs/readthedocs-sphinx-search.
> 
> Can you please test it for me?

Yeah, this is kinda hard to test, because it only works on RTD.

I've added it on a (temporary) branch on this dormant project of mine: https://magic-call.readthedocs.io/en/sphinx-search-with-shortcuts/

It seems to work fine, AFAICT.

The extension seems to not add the `highlight` thing when clicking on a search result, though (which is IMHO a bug on their side).

If you have a highlight link though (e.g. https://magic-call.readthedocs.io/en/sphinx-search-with-shortcuts/?highlight=package#the-python-package-magic-call) and then you hit `/` to search but then decide to cancel the search with `Esc`, the highlighting will also be cleared, which is probably not intentional.

I guess that's also a bug on their side, I've made a PR: https://github.com/readthedocs/readthedocs-sphinx-search/pull/103. I couldn't test this though, because I couldn't manage to minify the thing.

marxin:
@mgeier Thank you for the testing effort!

mgeier:
In the meantime I've found out that readthedocs-sphinx-search is already supposed to grab the `/` keypress, but it doesn't seem to work on keyboards which need `Shift` to get `/`, see https://github.com/readthedocs/readthedocs-sphinx-search/pull/31.

jakobandersen (approved):

jakobandersen:
Now that 4.4 has been released, can you rebase it on top of 4.x? And move the CHANGES entry correspondingly?

marxin:
> Now that 4.4 has been released, can you rebase it on top of 4.x? And move the CHANGES entry correspondingly?

Done and moved CHANGES to the right place. Thanks for the approval.

tk0miya:
Is it time to merge this? I feel the discussion becomes too long. So how about merging this if you're okay.

pradyunsg:
Given the thumbs-up reactions, looks like the answer is a yes. :)

marxin:
Yes, please merge it.

tk0miya:
Okay, merging now. Thank you for your contribution, all!

PR #9340, opened 2021-06-14
Update example to use recommended usage

ocket8888:
Subject: The docs recommend enclosing option-arguments in angle brackets, but the very next example code block shows an option-argument not in angle brackets.

### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
- The docs recommend enclosing option-arguments in angle brackets, but the very next example code block shows an option-argument not in angle brackets.

### Detail
- adds angle brackets to the aforementioned option-argument

### Relates
- https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -1673,7 +1673,7 @@ There is a set of directives allowing documenting command-line programs:
 
       .. program:: svn
 
-      .. option:: -r revision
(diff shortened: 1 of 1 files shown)


jfbu (approved):
Indeed, LGTM!

tk0miya (approved):

tk0miya:
Thanks!

PR #9390, opened 2021-06-30
texinfo: fix emission of @footnote directives.

marxin:
Right now, labels are emitted as part of a @footnote directive.
That results in e.g.

```
Note1: (@footnote{@w{(1)}
Future versions of GCC may zero-extend...
})
```

which is incorrect and should be rather:

```
Note1: (@footnote{Future versions of GCC may zero-extend...})
```


marxin:
Texinfo documentation for `@footnote`: https://www.gnu.org/software/texinfo/manual/texinfo/html_node/Footnote-Commands.html

sphinx/writers/texinfo.py
```diff
@@ -1223,7 +1225,10 @@ def depart_sidebar(self, node: Element) -> None:
         self.depart_topic(node)
 
     def visit_label(self, node: Element) -> None:
-        self.body.append('@w{(')
+        if self.in_footnote:
+            raise nodes.SkipNode
+        else:
+            self.body.append('@w{(')
```
tk0miya, line 1242: Note: This is still needed for citations.

sphinx/writers/texinfo.py
```diff
@@ -1223,7 +1225,10 @@ def depart_sidebar(self, node: Element) -> None:
         self.depart_topic(node)
 
     def visit_label(self, node: Element) -> None:
-        self.body.append('@w{(')
+        if self.in_footnote:
+            raise nodes.SkipNode
```
tk0miya, line 1240: Could you add a comment why labels are skipped in footnotes?

sphinx/writers/texinfo.py
```diff
@@ -761,10 +761,12 @@ def visit_title_reference(self, node: Element) -> None:
     # -- Blocks
 
     def visit_paragraph(self, node: Element) -> None:
-        self.body.append('\n')
+        if not self.in_footnote:
+            self.body.append('\n')
```
tk0miya: Is this removal required? It seems Texinfo does not mind if any whitespace exists.

tests/test_build_texinfo.py
```diff
@@ -112,3 +112,12 @@ def test_texinfo_escape_id(app, status, warning):
...
     assert translator.escape_id('Hello world.') == 'Hello world'
     assert translator.escape_id('.') == '.'
+
+
+@pytest.mark.sphinx('texinfo', testroot='footnotes')
+def test_texinfo_footnote(app, status, warning):
+    app.builder.build_all()
+
+    output = (app.outdir / 'python.texi').read_text()
+    assert '@footnote{@w{' not in output
```
tk0miya: It's enough to check the form of the new output. Checking the removal of the footnote label is meaningless.

tk0miya (changes requested):
LGTM with nits.

sphinx/writers/texinfo.py
marxin: Well, the generated texinfo output is more readable and easier for writing a test-case. Excluding the hunk, one would see:

```
@chapter The section with a reference to @footnote{
Footnote in section
}
```

sphinx/writers/texinfo.py
marxin: All right, so should I add a counter that would increment/decrement once a citation is visited (`in_citation`)?

marxin:
@tk0miya May I please remind this pull request where I asked some questions?

sphinx/writers/texinfo.py
tk0miya: Finally, I determined to remove this code to keep our writer simple. I think the simpleness of the writer is more important than the readability of output.

sphinx/writers/texinfo.py
tk0miya: I'm not sure. Let's think that again if we'll need to support.

tk0miya (approved):

tk0miya:

PR #9391, opened 2021-06-30
texinfo: improve variable in :samp: directives

marxin:
The following snippet:
Show :samp:`Samp with a {variable}.`

Ends in .info as:
Show ‘Samp with a `variable'.’

Which is suboptimal and @var{variable} should be rather used.
That results in ‘Samp with a VARIABLE.’.

### DIFF
--- a/sphinx/writers/texinfo.py
+++ b/sphinx/writers/texinfo.py
@@ -194,6 +194,7 @@ class TexinfoTranslator(SphinxTranslator):
         self.curfilestack: List[str] = []
         self.footnotestack: List[Dict[str, List[Union[collected_footnote, bool]]]] = []  # NOQA
         self.in_footnote = 0
+        self.in_samp = 0
         self.handled_abbrs: Set[str] = set()
         self.colwidths: List[int] = None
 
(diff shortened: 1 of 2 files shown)


tk0miya:
Thank you for your contribution. But this affects the extensions that output nested emphasis nodes. I think all nested nodes should not be rendered as `@var`.

marxin:
Are you sure? I mean, `@var` is a valid node even in `@samp` context as mentioned here:
https://www.gnu.org/software/texinfo/manual/texinfo/html_node/_0040var.html

and they mention the following example usage:
```
To delete file @var{filename},
type @samp{rm @var{filename}}.
```

which should correspond to:
```
... type :samp:`rm {filename}`
```

marxin:
@tk0miya May I please remind you this pull request?

tk0miya:
Sorry for not responding long. This code generates `@var` tag for the emphasis nodes under the literal node. It means this changes the output not only for the `:samp:` role, but also for other extensions that generates literal node and emphasis node. I think it's an unexpected change. So we have to tag to the emphasis node to track that they're surely generated by `:samp:` role.

marxin:
> Sorry for not responding long. This code generates `@var` tag for the emphasis nodes under the literal node. It means this changes the output not only for the `:samp:` role, but also for other extensions that generates literal node and emphasis node. I think it's an unexpected change. So we have to tag to the emphasis node to track that they're surely generated by `:samp:` role.

All right, now I've got it. I updated the patch accordingly.

marxin:
@tk0miya Can we get this to `4.3.0`? Btw. when is the next release supposed to happen?

tk0miya:
Sorry for the late response. Just merged.

PR #9398, opened 2021-07-02
Update docs for theming

yuxiaoy1:
We should add the import sentences for `copy_asset_file` and `path`, or it would take time for users to find them.


tk0miya:
Reasonable. +1!

PR #9400, opened 2021-07-03
Improve body text alignment for better aesthetics

yuxiaoy1:
This PR comes from https://github.com/pallets/pallets-sphinx-themes/pull/49, as it occurred to me that the changes should be made to the root repo.


astrojuanlu:
I am not an expert, but I'm not sure if right-justify should be added without deeper inspection. Without proper hyphenation (which is language-specific and not yet well supported by browsers) and whitespace control (to my knowledge, no traces of this for web browsers yet), it can create accessibility problems. See for example:

https://accessible-digital-documents.com/blog/justified-text/

https://www.futurehosting.com/blog/why-does-justified-text-have-such-a-bad-reputation-on-the-web/

https://maxwellforbes.com/posts/web-justified-text

I also think it was a bad idea to merge it in the pallets theme.

Blendify:
I would be against this patch also, I think most typography experts would agree that justified aligned content hurts readability. I think the best choice would be to set `hyphens: auto` if you are looking for better-looking right text blocks. Setting this property won't affect all words but lets the browser decide to only split long words that its algorithm deems best to be hyphenated.

AA-Turner:
I think we should avoid this for ``basic``, though a sub-theme could use fully-justified text.

A

Issue #9413, opened 2021-07-06
Autodoc generates invalid xml 


nemani:
### Describe the bug

When using autodoc along with python function signatures containing "params" with XML builder the generated xml is not valid. 

We think the issue is due to the following commit: https://github.com/sphinx-doc/sphinx/commit/30c594fbc592a741956c067f5188f9bf6ecd25da#

### How to Reproduce

1. git clone https://github.com/nemani/sphinx-bug-report-example 
1. Install sphinx
1. cd docs
1. make xml
1. xmllint --noout docs/build/xml/index.xml

### Expected behavior

The generated XML should be valid and xml lint should not have any errorss

### Your project

https://github.com/nemani/sphinx-bug-report-example

### OS

mac 

### Python version

3.7.5

### Sphinx version

4.0.3

### Sphinx extensions

sphinx.ext.autodoc

### Extra tools

- 

### Additional context

-

nemani:
This is being a huge blocker for me.
@tk0miya any suggested workarounds for this?

nemani:
I have a suggested fix up as a PR.

ghost:
The moment I thought to have finally found a solution to work around `autodoc` bugs... this appears to be also broken at the baseline. Lol. Please, please fix it... the amount of work I've invested already in trying to generate documentation and trying to work around broken parts is getting into the area where it would've been just easier to do it all from scratch.

PR #9416, opened 2021-07-07
#9413 Declare py namespace on document root to generate valid xml

nemani:
Subject: Declare py namespace in XML writer.

### Feature or Bugfix
- Bugfix

### Purpose
- Autodoc uses `py:class`, `py:module`, etc as attributes for nodes. This causes errors as the generated xml does not have a namespace for `py` declared. This PR fixes this by declaring a `xmlns:py` at the document level.

### Relates
- #9413 

Please recommend what URI to use for the declaration of the `py` xml namespace

### DIFF
--- a/sphinx/builders/xml.py
+++ b/sphinx/builders/xml.py
@@ -71,6 +71,7 @@ class XMLBuilder(Builder):
         # work around multiple string % tuple issues in docutils;
         # replace tuples in attribute values with lists
         doctree = doctree.deepcopy()
+        doctree.document.attributes["xmlns:py"] = "https://github.com/sphinx-doc/sphinx/issues/9413"
         for node in doctree.traverse(nodes.Element):
             for att, value in node.attributes.items():
                 if isinstance(value, tuple):


sphinx/builders/xml.py
```diff
@@ -71,6 +71,7 @@ def write_doc(self, docname: str, doctree: Node) -> None:
         # work around multiple string % tuple issues in docutils;
         # replace tuples in attribute values with lists
         doctree = doctree.deepcopy()
+        doctree.document.attributes["xmlns:py"] = "https://github.com/sphinx-doc/sphinx/issues/9413"
```
tk0miya, line 74: I think it's not good to set the URL of #9413. I'd like to use a better one for this. What kind of URL should we use instead? document? project page (GitHub)?

tk0miya, line 74: The "py" namespace comes from the domains mark-up.
https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

And we have many kinds of domains (ex. C, C++, etc.). And developers can add them via API. So it would be better to add it with such code:
```suggestion
        for domain in self.env.domains.values():
            xmlns = "xmlns:" + domain.name
            doctree.document.attributes[xmlns] = "https://URL/to/somewhere"
```

tk0miya (changes requested):
LGTM with nits.

sphinx/builders/xml.py
nemani: Maybe we can point to the documentation of the domains?

tk0miya: We can point only the URL of the "bundled" domains: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html
But this document does not contains the 3rd party domains. For example, sphinx-http-domain: https://pypi.org/project/sphinx-http-domain/.

nemani: Maybe we can keep it as an attribute in the domain class with default pointing to the "bundled" domains doc?

tk0miya:
Now I merged #10100 instead. Thank you for your contribution!

PR #9422, opened 2021-07-08
Add Icelandic to supported languages

TryggviKalman:
Subject: Add Icelandic to supported languages

### Feature or Bugfix
- Feature


tk0miya:
Thank you for your contribution. But we've used transifex to translate message catalogs. Could you join
https://www.transifex.com/sphinx-doc/sphinx-1/ and update the messages, please?

PR #9355, opened 2021-06-16
New Sphinx tutorial, part II

astrojuanlu:
### Feature or Bugfix
- Feature

### Purpose
This is a continuation of the tutorial we started at #9276. It expands some sections already started in the first part and adds some new ones, as originally proposed in #9165:

4. First steps to document our project using Sphinx
5. Customizing Sphinx configuration
6. Writing narrative documentation with Sphinx

Again, the idea is to keep these pull requests manageable, iterate in small chunks, and in the meanwhile keep the tutorial in a complete state.

Rendered version: https://sphinx--9355.org.readthedocs.build/en/9355/tutorial/index.html

cc @ericholscher 

### Relates
- #9276
- #9165

PR #9424, opened 2021-07-08
New Sphinx tutorial (intersphinx, postponed)

astrojuanlu:
### Feature or Bugfix
- Feature

### Purpose
_(Still draft)_ This is a continuation of the work we did in #9276 and #9355. It expands some sections already started in the first part and adds some new ones, as originally proposed in #9165:

6. Writing narrative documentation with Sphinx (adding a section on intersphinx)
7. Describing code in Sphinx

I plan to add quite some content to (7) (still not pushed), and leave "Autogenerating documentation from code in Sphinx" for the fourth PR. Again, the idea is to keep these pull requests manageable, iterate in small chunks, and in the meanwhile keep the tutorial in a complete state.

Rendered version: https://sphinx--9424.org.readthedocs.build/en/9424/tutorial/index.html

cc @ericholscher

### Relates
- #9165
- #9276
- #9355

### DIFF
--- a/doc/tutorial/index.rst
+++ b/doc/tutorial/index.rst
@@ -267,7 +267,7 @@ follows:
        'sphinx.ext.duration',
    ]
 
-After that, every time you generate your documentation, you will see a short
+After that, every time you build your documentation, you will see a short
 durations report at the end of the console output, like this one:
 
 .. code-block:: console
@@ -437,8 +437,8 @@ And make the sentence you added in ``index.rst`` look like this:
    :ref:`install <installation>` the project.
 
 Notice a trick here: the ``install`` part specifies how the link will look like
-(we want it to be a specific word, so the sentence makes sense), whereas the
-``<installation>`` part refers to the actual label we want to add a
+(in this case, using a specific word so the sentence makes sense), whereas the
+``<installation>`` part refers to the actual label you want to add a
 cross-reference to. If you do not include an explicit title, hence using
 ``:ref:`installation```, the section title will be used (in this case,
 ``Installation``). Both the ``:doc:`` and the ``:ref:`` roles will be rendered
--- a/doc/tutorial/index.rst
+++ b/doc/tutorial/index.rst
@@ -444,6 +444,67 @@ cross-reference to. If you do not include an explicit title, hence using
 ``Installation``). Both the ``:doc:`` and the ``:ref:`` roles will be rendered
 as hyperlinks in the HTML documentation.
 
+Referencing external projects with intersphinx
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Most likely you are depending on some open source libraries for your own
+project, and there is a chance those dependencies have their documentation
+written in Sphinx and available online. One of the most powerful features of
+Sphinx is the ability to generate cross-references to objects from other
+projects available online, using the
+:doc:`intersphinx </usage/extensions/intersphinx>` extension.
+
+To enable it, first add ``sphinx.ext.intersphinx`` to the list of extensions:
+
+.. code-block:: python
+   :caption: docs/source/conf.py
+
+   extensions = [
+       'sphinx.ext.duration',
+       'sphinx.ext.intersphinx',
+   ]
+
+The way to know whether you can reference an external project is to check
+if it has a publicly available ``objects.inv``, which can be parsed calling
+``python -m sphinx.ext.intersphinx``. For example, this is the output for
+Sphinx itself:
+
+.. code-block:: console
+   :emphasize-lines: 4
+
+   (.venv) $ python -msphinx.ext.intersphinx https://www.sphinx-doc.org/en/master/objects.inv  | grep 'std:doc' -A5
+   std:doc
+           changes                                  Changelog                               : changes.html
+           contents                                 Sphinx documentation contents           : contents.html
+           development/builders                     Configuring builders                    : development/builders.html
+           development/index                        Extending Sphinx                        : development/index.html
+           development/overview                     Developing extensions overview          : development/overview.html
+
+You can now introduce a cross-reference to the ``contents`` section of the
+Sphinx documentation as follows, prefixing it by `sphinx:` to better identify
+it is an external one:
+
+.. code-block:: rst
+   :caption: docs/source/index.rst
+
+   .. note::
+
+      This documentation is built on Sphinx,
+      :doc:`find out more in their docs <sphinx:contents>`.
+
+And finally, to actually link our documentation with the Sphinx one, add a
+:confval:`intersphinx_mapping` dictionary to your  as follows:
+
+.. code-block:: python
+   :caption: docs/source/conf.py
+
+   intersphinx_mapping = {
+       'sphinx': ('https://www.sphinx-doc.org/en/master/', None),
+   }
+
+After you build the HTML documentation with ``make html``, you will see a link
+pointing to that specific section of the Sphinx documentation!
+
 Where to go from here
 ---------------------
 
--- a/doc/tutorial/index.rst
+++ b/doc/tutorial/index.rst
@@ -447,10 +447,10 @@ as hyperlinks in the HTML documentation.
 Referencing external projects with intersphinx
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Most likely you are depending on some open source libraries for your own
+You are likely depending on some open source libraries for your own
 project, and there is a chance those dependencies have their documentation
 written in Sphinx and available online. One of the most powerful features of
-Sphinx is the ability to generate cross-references to objects from other
+Sphinx is the ability to generate cross-references to other
 projects available online, using the
 :doc:`intersphinx </usage/extensions/intersphinx>` extension.
 
@@ -482,7 +482,7 @@ Sphinx itself:
 
 You can now introduce a cross-reference to the ``contents`` section of the
 Sphinx documentation as follows, prefixing it by `sphinx:` to better identify
-it is an external one:
+it as an external reference:
 
 .. code-block:: rst
    :caption: docs/source/index.rst
@@ -503,7 +503,7 @@ And finally, to actually link our documentation with the Sphinx one, add a
    }
 
 After you build the HTML documentation with ``make html``, you will see a link
-pointing to that specific section of the Sphinx documentation!
+pointing to that specific page of the Sphinx documentation!
 
 Where to go from here
 ---------------------
--- a/doc/tutorial/index.rst
+++ b/doc/tutorial/index.rst
@@ -449,7 +449,7 @@ Referencing external projects with intersphinx
 
 You are likely depending on some open source libraries for your own
 project, and there is a chance those dependencies have their documentation
-written in Sphinx and available online. One of the most powerful features of
+written in Sphinx and available online. One of the powerful features of
 Sphinx is the ability to generate cross-references to other
 projects available online, using the
 :doc:`intersphinx </usage/extensions/intersphinx>` extension.
--- a/doc/tutorial/index.rst
+++ b/doc/tutorial/index.rst
@@ -480,9 +480,19 @@ Sphinx itself:
            development/index                        Extending Sphinx                        : development/index.html
            development/overview                     Developing extensions overview          : development/overview.html
 
+To link your documentation with the Sphinx one, add a
+:confval:`intersphinx_mapping` dictionary to your ``conf.py`` as follows:
+
+.. code-block:: python
+   :caption: docs/source/conf.py
+
+   intersphinx_mapping = {
+       'sphinx': ('https://www.sphinx-doc.org/en/master/', None),
+   }
+
 You can now introduce a cross-reference to the ``contents`` section of the
-Sphinx documentation as follows, prefixing it by `sphinx:` to better identify
-it as an external reference:
+Sphinx documentation as follows, optionally prefixing it by ``sphinx:``, the
+key you used in the ``intersphinx_mapping`` configuration:
 
 .. code-block:: rst
    :caption: docs/source/index.rst
@@ -491,16 +501,6 @@ it as an external reference:
 
       This documentation is built on Sphinx,
       :doc:`find out more in their docs <sphinx:contents>`.
-
-And finally, to actually link our documentation with the Sphinx one, add a
-:confval:`intersphinx_mapping` dictionary to your  as follows:
-
-.. code-block:: python
-   :caption: docs/source/conf.py
-
-   intersphinx_mapping = {
-       'sphinx': ('https://www.sphinx-doc.org/en/master/', None),
-   }
 
 After you build the HTML documentation with ``make html``, you will see a link
 pointing to that specific page of the Sphinx documentation!
--- a/doc/tutorial/index.rst
+++ b/doc/tutorial/index.rst
@@ -472,7 +472,7 @@ Sphinx itself:
 .. code-block:: console
    :emphasize-lines: 4
 
-   (.venv) $ python -msphinx.ext.intersphinx https://www.sphinx-doc.org/en/master/objects.inv  | grep 'std:doc' -A5
+   (.venv) $ python -msphinx.ext.intersphinx https://www.sphinx-doc.org/en/master/objects.inv  | grep 'std:doc' --after-context 5
    std:doc
            changes                                  Changelog                               : changes.html
            contents                                 Sphinx documentation contents           : contents.html


doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+``<installation>`` part refers to the actual label you want to add a
 cross-reference to. If you do not include an explicit title, hence using
 ``:ref:`installation```, the section title will be used (in this case,
 ``Installation``). Both the ``:doc:`` and the ``:ref:`` roles will be rendered
 as hyperlinks in the HTML documentation.
 
+Referencing external projects with intersphinx
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Most likely you are depending on some open source libraries for your own
```
ericholscher:
```suggestion
You are likely depending on some open source libraries for your own
```

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+project, and there is a chance those dependencies have their documentation
+written in Sphinx and available online. One of the most powerful features of
+Sphinx is the ability to generate cross-references to objects from other
```
ericholscher:
```suggestion
Sphinx is the ability to generate cross-references to other
```

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+
+The way to know whether you can reference an external project is to check
+if it has a publicly available ``objects.inv``, which can be parsed calling
+``python -m sphinx.ext.intersphinx``. For example, this is the output for
+Sphinx itself:
+
+.. code-block:: console
+   :emphasize-lines: 4
+
+   (.venv) $ python -msphinx.ext.intersphinx https://www.sphinx-doc.org/en/master/objects.inv  | grep 'std:doc' -A5
```
ericholscher: What is `-A5`? Generally you should use long-form commands in docs, so its more obvious what it's doing. I assume this is `--append 5` or something?

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+   std:doc
+           changes                                  Changelog                               : changes.html
+           contents                                 Sphinx documentation contents           : contents.html
+           development/builders                     Configuring builders                    : development/builders.html
+           development/index                        Extending Sphinx                        : development/index.html
+           development/overview                     Developing extensions overview          : development/overview.html
+
+You can now introduce a cross-reference to the ``contents`` section of the
+Sphinx documentation as follows, prefixing it by `sphinx:` to better identify
+it is an external one:
```
ericholscher:
```suggestion
it as an external reference:
```

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+
+And finally, to actually link our documentation with the Sphinx one, add a
+:confval:`intersphinx_mapping` dictionary to your  as follows:
+
+.. code-block:: python
+   :caption: docs/source/conf.py
+
+   intersphinx_mapping = {
+       'sphinx': ('https://www.sphinx-doc.org/en/master/', None),
+   }
```
ericholscher: Seems this should go in the conf.py file above. Otherwise the `sphinx:` prefix in the previous step is confusing.

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+
+After you build the HTML documentation with ``make html``, you will see a link
+pointing to that specific section of the Sphinx documentation!
```
ericholscher:
```suggestion
pointing to that specific page of the Sphinx documentation!
```

ericholscher (approved):
Looks like a good section. I wonder if we should reference the RTD guide on this to learn more? Or maybe backport our guide into the Sphinx docs for intersphinx?

doc/tutorial/index.rst
astrojuanlu: Oh you're right, I'll change that. It's `--after-context 5`, therefore showing 5 lines after the match.

doc/tutorial/index.rst
astrojuanlu: Good point, will reorder it

doc/tutorial/index.rst
ericholscher: Not 100% sure on this one.

astrojuanlu:
Thanks @ericholscher for the early review, I addressed your comments. About this:

> I wonder if we should reference the RTD guide on this to learn more? Or maybe backport our guide into the Sphinx docs for intersphinx?

I can add a link for now, although if the Sphinx developers agree, we should probably backport the guide into the Sphinx docs. Not sure what the best place would be though, I will open a separate issue.

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
 ``:ref:`installation```, the section title will be used (in this case,
 ``Installation``). Both the ``:doc:`` and the ``:ref:`` roles will be rendered
 as hyperlinks in the HTML documentation.
 
+Referencing external projects with intersphinx
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You are likely depending on some open source libraries for your own
+project, and there is a chance those dependencies have their documentation
+written in Sphinx and available online. One of the most powerful features of
```
jakobandersen:
```suggestion
written in Sphinx and available online. One of the powerful features of
```
While I agree, I think we should be careful with judging. Different people have different features they are interested in.

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+
+.. code-block:: python
+   :caption: docs/source/conf.py
+
+   intersphinx_mapping = {
+       'sphinx': ('https://www.sphinx-doc.org/en/master/', None),
+   }
+
+You can now introduce a cross-reference to the ``contents`` section of the
+Sphinx documentation as follows, optionally prefixing it by ``sphinx:``, the
```
jakobandersen, line 494: This prefixing is a bit of an icky issue. It doesn't work in general for all roles (and fundamentally can't), so in the end this should illustrate the feature with the role introduced by #9062.

doc/tutorial/index.rst
astrojuanlu: Absolutely, good catch.

doc/tutorial/index.rst
astrojuanlu: I was not aware of that pull request. Do we want to introduce it even if it's not merged yet? Also, out of curiosity, can you expand a bit (or give an example) that doesn't work?

jakobandersen: The tutorial should definitely be in sync with the implementation, so it's more that we should be aware of updating this part when relevant. Though, I guess this part will be merged after 4.1 (going out this weekend I believe), so there is a chance of the PR being ready in the same release as this.
As concrete examples, the roles in the C and C++ domains do not support this prefix. There are more details in https://github.com/sphinx-doc/sphinx/pull/8418#issuecomment-758838589, but the fundamental issue is that when you call a role, e.g., ``:py:class:`` you ask the Python domain extension to do whatever it likes. This is most likely to insert a ``pending_xref`` node with meta-information, which that same extension is being asked to resolve at a later stage. When you suddenly add the prefix, you explicitly don't want that domain to handle it, but hope it understands and fails to resolve the xref, such that intersphinx can get a chance to resolve according to the prefix. Hence, fundamentally the syntax should be with a role that directly calls intersphinx directly, without other extensions involved (until a later stage, see also #8929).

tk0miya:
Are you going to add all subsections of tutorials to `doc/tutorial/index.rst`? Is it better to separate this page to multiple ones?

astrojuanlu:
@tk0miya Sooner or later I wanted to split this into pages, yes. Do you think it's the moment already?

doc/tutorial/index.rst
```diff
@@ -437,13 +437,74 @@ And make the sentence you added in ``index.rst`` look like this:
...
+.. code-block:: python
+   :caption: docs/source/conf.py
+
+   extensions = [
+       'sphinx.ext.duration',
+       'sphinx.ext.intersphinx',
+   ]
+
+The way to know whether you can reference an external project is to check
+if it has a publicly available ``objects.inv``, which can be parsed calling
```
tk0miya, line 468: The goal of the Intersphinx extension is referring to remote objects transparently like referring to local objects. We've usually found local objects via reading source documents, I believe. So I think the way to find the remote objects is also reading source documents, first. Is searching `objects.inv` really better way to do that? I don't think so...

This tutorial does not describe the relationship between the `std:doc` and `:doc:` role. Where they came from? I'm worried this explanation makes beginners confused.

doc/tutorial/index.rst
tk0miya: As I commented above, one of the goal of intersphinx is transparency. So I prefer not to use prefix for cross references. As you explained, it is "optional" feature. So it would be better to explain the extension by non-prefixed example at first, I think.

tk0miya:
> @tk0miya Sooner or later I wanted to split this into pages, yes. Do you think it's the moment already?

Yes, I think so. I guess this is why you'd like to add a `content` directive at last PR. I think no reason to avoid adding a new page.

doc/tutorial/index.rst
astrojuanlu: Fair enough, it looked more clar to me that way, but I'm +1 on explaning the fully consistent and transparent way instead. Will rephrase this part.

doc/tutorial/index.rst
astrojuanlu: > So I think the way to find the remote objects is also reading source documents, first. Is searching objects.inv really better way to do that? I don't think so...

This was only to explain what the file is. Perhaps it is too technical? I find myself using this quite often lately, perhaps because I'm still learning the intricacies of intersphinx. @webknjaz has https://webknjaz.github.io/intersphinx-untangled/ and I think it's quite useful too.

In any case, I can clarify that this is optional and that it works in the same way as your own docs (except for custom roles, like Sphinx own `:confval:`.

astrojuanlu: > This tutorial does not describe the relationship between the std:doc and :doc: role. Where they came from? I'm worried this explanation makes beginners confused.

It doesn't because I need to learn it myself...!

astrojuanlu: I had no luck on the Sphinx docs but I found this [somewhere else](https://alnoki.readthedocs.io/en/latest/dev-guide/procedures/sphinx.html#id12):

Screenshot 2021-07-13 at 17-21-41 6 4 Sphinx — AAAAAA 0 4 0 documentation

Now I understand your point. I will think if I can clarify this earlier, otherwise rephrase this part.

ericholscher: I'd love to see this table in the official docs.

astrojuanlu: I am trying to make sense of domains, and the fact that "Cross-references are generated by many semantic interpreted text roles" ([ref](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax)) but it's not explicitly stated _which ones_. Then I found that `std:doc` has some special handling:

https://github.com/sphinx-doc/sphinx/blob/5559e5af1ff6f5fc2dc70679bdd6dc089cfff388/sphinx/domains/std.py#L1017-L1018

jakobandersen: I would strongly prefer to consider ``objects.inv`` an implementation detail that should not go into the tutorial. The current format is quite limiting and it only works for relatively simple reference types, but it needs to be changed in the future to make more complicated cases work (see e.g., https://github.com/sphinx-doc/sphinx/pull/8929). If I remember correctly the ``get_objects()`` method for domains is basically only used for ``objects.inv`` so it will probably get changed as well.

jakobandersen: > I am trying to make sense of domains, and the fact that "Cross-references are generated by many semantic interpreted text roles" ([ref](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax)) but it's not explicitly stated _which ones_. Then I found that `std:doc` has some special handling:

It inherently can't state which, as each domain can independently offer such roles, and domains can be third-party extensions. The ``std`` domain is a bit special, bit generally, if domain has a directive to declare an entity, then there usually is a corresponding role for referring to it, e.g., ``.. rst:directive::`` has ``:rst:dir:`` and ``.. py:class::`` has ``:py:class:``.
In this case, if I understand it correctly, the ``std`` considers each document to be an entity implicitly declared, and offers ``:doc:`` to refer them. Similarly, I think when you create a label somewhere, it is really an entity declared in the ``std`` domain, which you can refer to with ``:ref:``. @tk0miya, is this last part correct?

astrojuanlu: Well, after spending some time on this, I think it's better that I don't give `objects.inv` files such a central treatment in this tutorial. Yes, they are useful (I use them all the time now) but the big problem is that we don't have a reference or a cheatsheet where the user can look up all possible keys that will be found there (`std:doc`, `std:label`, the deprecated `std:cmdoption`, and so forth) with the corresponding roles that can be used to insert cross-references to them (`std:doc` ▶️ `:doc:`, `std:label` ▶️ `:ref:`, etc). [Tutorials should only contain minimum, necessary, explanations](https://diataxis.fr/tutorials/#offer-only-minimum-necessary-explanation), but the example I crafted here is just too magical and can lead to frustration. In summary, I will rework this section.

jakobandersen: It highlights a good point: that we should have better tooling around intersphinx inventories, to help users and hide the implementation details.

astrojuanlu: I have one last caveat:

- In my dummy project, if I enable intersphinx and write

```
:doc:`contents`
```

the link goes to https://www.sphinx-doc.org/en/master/contents.html, as advertised by `objects.inv`.

- However, from Sphinx itself, if I write

```
:doc:`contents`
```

it goes to... https://docs.python.org/3/contents.html! For it to work, I need to write

```
:doc:`/contents`
```

instead. This was very surprising to me, and I think it makes more difficult to introduce the topic: it's not as easy as "write a reference as you would write it in your docs", the user also has to understand some extra rules about how intersphinx targets are generated. Am I missing something @tk0miya @jakobandersen ?

astrojuanlu: (Also, sorry @jakobandersen - I just saw your replies now, somehow the GitHub UI was not showing them to me? Maybe I needed to refresh the page or something 😕)

jakobandersen: I'm not sure about the leading slash and what is going on there, but you example reminds me of  #8981 which allows to ban the transparent resolution through intersphinx, e.g., as simple typos can suddenly resolve to an external entity instead of an internal one.

ericholscher: I'd really like to advocate for explicit remote references for intersphinx. If you view it as namespace, there's lots of ways to have things resolve to something unexpected (as @astrojuanlu mentioned above). I would really like there to be one, well-defined way to reference a document at a specific intersphinx mapping, without any ambiguity as to what I'm referencing. #8981 is a good step towards that 👍

0xMashiro:
Really useful. I think @FateScript will be interested in it.

doc/tutorial/index.rst
webknjaz: > We've usually found local objects via reading source documents, I believe. So I think the way to find the remote objects is also reading source documents, first. Is searching `objects.inv` really better way to do that? I don't think so...
> 

I've never had to do this by reading the source — it's hard to locate such things when you don't know where they are defined and how. Besides, the source is not always linked or available.
This is why I came up with intersphinx-untangled — it helps me discover things that are actually available. I sometimes even search by URL to see what roles link there.

I think there should be some sort of a reference explorer (I mentioned this somewhere on the tracker earlier).

webknjaz: > It highlights a good point: that we should have better tooling around intersphinx inventories, to help users and hide the implementation details.

Yep: https://github.com/sphinx-doc/sphinx/issues/8211

astrojuanlu:
For the record, we are holding off this part of the tutorial until there is more progress on the new intersphinx features (like #9459, #9062) and so that we can reach an agreement on whether users should be using `objects.inv` files or not.

astrojuanlu:
(I didn't intend to close the PR, but it happened automatically because I renamed the branch, sorry 🤷🏽 it lives here now https://github.com/astrojuanlu/sphinx/tree/new-tutorial-part-intersphinx)

doc/tutorial/index.rst
tk0miya: >In this case, if I understand it correctly, the std considers each document to be an entity implicitly declared, and offers :doc: to refer them. Similarly, I think when you create a label somewhere, it is really an entity declared in the std domain, which you can refer to with :ref:. @tk0miya, is this last part correct?

Yes. Some kinds of objects (including labels and documents) belong to the std domain. It behaves like a system domain. So users do not need to give the domain name on referencing objects like `:std:doc` or `:std:ref:`.
https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/domains/std.py#L562-L570

tk0miya: > This was very surprising to me, and I think it makes more difficult to introduce the topic: it's not as easy as "write a reference as you would write it in your docs", the user also has to understand some extra rules about how intersphinx targets are generated. Am I missing something @tk0miya @ jakobandersen ?

I've never thought referencing remote document via intersphinx. So I did not know such a behavior. Indeed, it would be a bug.

astrojuanlu: > I've never thought referencing remote document via intersphinx. So I did not know such a behavior. Indeed, it would be a bug.

Thanks for chiming in @tk0miya ! I opened a discussion on https://groups.google.com/g/sphinx-dev/c/ZfslSpDGNUs/m/RjfzzFzUCgAJ about it, I think it's somewhat addressed by #9459

PR #9425, opened 2021-07-08
Fix link to NumPy Docstring Standard

astrojuanlu:
The current link still works, but requires the user to click again on the new one.


tk0miya:
Thanks!

PR #9427, opened 2021-07-09
Closes #9418: Fix missing bracket for empty Callable annotations.

### LINKED ISSUE #9418: Type annotation: missing bracket for Callable with no arguments
### Describe the bug

This bug is encountered for the sphinx.ext.autodoc extension.
When type-hinting an attribute as `Callable[[], None]` (a function with no parameters that returns None), the generated annotation is missing a bracket: `Callable[], None]` .
Brackets are correctly displayed if:
* `Callable[..., None]` or `Callable[[int], None]` is used;
* the annotation is a function return type instead of an attribute type.

### How to Reproduce

MWE
---
file `testhints.py`:
```python
from typing import Callable

class Test:
    """A test class"""
    def __init__(self):
        self.var: Callable[[], None] = self.get_var()
        """A test attribute"""

    def get_var(self) -> Callable[[], None]:
        """A test function"""
        def f():
            pass
        return f
```
file `index.rst`:
```rst
.. autoclass:: testhint.Test
   :members:
```

### Expected behavior

The type annotation for `var` should be `Callable[[], None]`, as is the case for `get_var()`'s return type, not `Callable[], None]`.

### Your project

MWE in `How to reproduce`

### Screenshots

image

### OS

Windows 10

### Python version

3.7

### Sphinx version

4.0.1

### Sphinx extensions

sphinx.ext.autodoc

### Extra tools

_No response_

### Additional context

_No response_

thjungers:
Subject: Fix missing bracket for empty Callable annotations. This solves issue #9418 

### Feature or Bugfix
- Bugfix

### Purpose
- When rendering a `Callable` type annotation, the output is missing a bracket if the `Callable` requires no parameters (the first element is empty): `Callable[[], None]` is rendered as `Callable[], None]`.

### Detail
When unparsing a parsed annotation containing an `ast.List` object, the following pseudo-code is run:
``` python
result.append('[')
for elem in node.elts:
    result.append(elem)
    result.append(', ')
result.pop()
result.append(']')
```
The last element (the extra comma) is popped. However, if the `ast.List` object is empty, the for-loop is skipped and the popped element is the opening `[`.
An if-condition was added to check whether `node.elts` is empty before popping the last element of `result`.

### Relates
- Issue #9418

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* #9418: a Callable annotation with no parameters (e.g. Callable[[], None]) will
+  be rendered with a bracket missing (Callable[], None])
+
 Testing
 --------
 
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -129,10 +129,14 @@ def _parse_annotation(annotation: str, env: BuildEnvironment = None) -> List[Nod
             return unparse(node.value)
         elif isinstance(node, ast.List):
             result = [addnodes.desc_sig_punctuation('', '[')]
-            for elem in node.elts:
-                result.extend(unparse(elem))
-                result.append(addnodes.desc_sig_punctuation('', ', '))
-            result.pop()
+            if node.elts:
+                # check if there are elements in node.elts to only pop the
+                # last element of result if the for-loop was run at least
+                # once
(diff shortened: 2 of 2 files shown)


tk0miya:
Merge this manually into the 4.x branch now.

PR #9439, opened 2021-07-12
Fix release of Sphinx by forcing upgrades to dependencies

### LINKED ISSUE #9434: 4.1.0 regression: `Could not import extension sphinx.builders.epub3 (exception: cannot import name 'RemovedInSphinx40Warning' from 'sphinx.deprecation'`
### Describe the bug

So this morning I've got an automatic dependabot dependency pin update PR changing Sphinx from v4.0.3 to v4.1.0. This caused the build to error out with the text as in $sbj. It is notable that the builder that causes the error is `epub3` while the builder invoked via CLI is `html`.

### How to Reproduce

```console
$ git clone https://github.com/ansible/pylibssh
$ cd pylibssh
$ pip install tox
$ tox -e build-docs
```

### Expected behavior

No errors.

### Your project

https://github.com/ansible/pylibssh

### Screenshots

_No response_

### OS

GHA CI / Ubuntu 20.04

### Python version

CPython v3.9.5

### Sphinx version

v4.1.0

### Sphinx extensions

sphinx.ext.autodoc, sphinx.ext.autosectionlabel, sphinx.ext.doctest, sphinx.ext.extlinks, sphinx.ext.intersphinx, sphinx.ext.todo, sphinx.ext.coverage, sphinx.ext.viewcode, sphinxcontrib.towncrier, myst_parser

### Extra tools

_No response_

### Additional context

https://github.com/ansible/pylibssh/pull/238/checks?check_run_id=3044079226#step:11:46

ericholscher:
### Feature or Bugfix

- Bugfix

### Purpose

This fixes #9434, and should probably go into a new 4.1.1 release

### Detail

We should probably have a proper versioning policy for these dependencies. Are they really all required? Should some be optional, and not break Sphinx if they aren't able to be imported? 

It seems that each should probably be updated as a dependency on each Sphinx release, given that they cause breaking changes Sphinx if they get out of date. 

### Relates

#9434

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,12 +15,12 @@ if sys.version_info < (3, 6):
     sys.exit(1)
 
 install_requires = [
-    'sphinxcontrib-applehelp',
-    'sphinxcontrib-devhelp',
-    'sphinxcontrib-jsmath',
-    'sphinxcontrib-htmlhelp',
-    'sphinxcontrib-serializinghtml',
-    'sphinxcontrib-qthelp',
+    'sphinxcontrib-applehelp>=1.0.2',
+    'sphinxcontrib-devhelp>=1.0.2',
+    'sphinxcontrib-jsmath>=1.0.1',
+    'sphinxcontrib-htmlhelp>=2.0.0',
+    'sphinxcontrib-serializinghtml>=1.1.5',
+    'sphinxcontrib-qthelp>=1.0.3',
     'Jinja2>=2.3',
     'Pygments>=2.0',
     'docutils>=0.14,<0.18',
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,24 @@
+Release 4.2.0 (in development)
+==============================
+
+Dependencies
+------------
+
+Incompatible changes
+--------------------
+
+Deprecated
+----------
+
+Features added
+--------------
+
+Bugs fixed
+----------
+
+Testing
+--------
+
 Release 4.1.1 (in development)
 ==============================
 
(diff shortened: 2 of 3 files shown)


webknjaz (approved):

setup.py
```diff
@@ -15,12 +15,12 @@
...
-    'sphinxcontrib-devhelp',
-    'sphinxcontrib-jsmath',
-    'sphinxcontrib-htmlhelp',
-    'sphinxcontrib-serializinghtml',
-    'sphinxcontrib-qthelp',
+    'sphinxcontrib-applehelp>=1.0.2',
+    'sphinxcontrib-devhelp>=1.0.2',
+    'sphinxcontrib-jsmath>=1.0.1',
+    'sphinxcontrib-htmlhelp>=2.0.0',
+    'sphinxcontrib-serializinghtml>=1.1.5',
```
tk0miya, line 22: IMO, it's enough to pin only htmlhelp and serializinghtml.

tk0miya (changes requested):
LGTM with nits. Could you rebase this onto the 4.0.x branch, please?

setup.py
ericholscher: It seems like we should pin them all, so that we don't hit this issue in the future? I imagine there will also be breaking changes over time, and we want users to have the latest version. I don't feel strongly though.

ericholscher:
I rebased on 4.1.x, since 4.0.x doesn't exist:

```
-> git pull --rebase origin 4.0.x
fatal: Couldn't find remote ref 4.0.x
```

tk0miya:
OMG, this still contains commits on 4.x branch...

setup.py
tk0miya: -1: too strict

tk0miya:
I'm fixing this in another PR. Thanks,

Issue #9442, opened 2021-07-13
test_enum_class fails with Python 3.10-beta4


befeleme:
### Describe the bug

The Enum behavior was changed in Python 3.10-beta 4 to be the same as in Python 3.9. The changes to behavior are [postponed](https://mail.python.org/archives/list/python-dev@python.org/message/LSTMFAPSPD3BGZ4D6HQFODXZVB3PLYKF/) to Python 3.11. When changes to Enum were reverted, the test started to fail.
Changing the condition to `if sys.version_info < (3, 11)` seems to solve the problem. 

```
_______________________________ test_enum_class ________________________________
app = <SphinxTestApp buildername='html'>
    @pytest.mark.sphinx('html', testroot='ext-autodoc')
    def test_enum_class(app):
        options = {"members": None}
        actual = do_autodoc(app, 'class', 'target.enums.EnumCls', options)
    
        if sys.version_info < (3, 10):
            sig = '(value)'
        else:
            sig = ('(value, names=None, *, module=None, qualname=None, type=None, start=1, '
                   'boundary=None)')
    
>       assert list(actual) == [
            '',
            '.. py:class:: EnumCls%s' % sig,
            '   :module: target.enums',
            '',
            '   this is enum class',
            '',
            '',
            '   .. py:method:: EnumCls.say_goodbye()',
            '      :module: target.enums',
            '      :classmethod:',
            '',
            '      a classmethod says good-bye to you.',
            '',
            '',
            '   .. py:method:: EnumCls.say_hello()',
            '      :module: target.enums',
            '',
            '      a method says hello to you.',
            '',
            '',
            '   .. py:attribute:: EnumCls.val1',
            '      :module: target.enums',
            '      :value: 12',
            '',
            '      doc for val1',
            '',
            '',
            '   .. py:attribute:: EnumCls.val2',
            '      :module: target.enums',
            '      :value: 23',
            '',
            '      doc for val2',
            '',
            '',
            '   .. py:attribute:: EnumCls.val3',
            '      :module: target.enums',
            '      :value: 34',
            '',
            '      doc for val3',
            '',
        ]
E       AssertionError: assert ['', '.. py:c...ass', '', ...] == ['', '.. py:c...ass', '', ...]
E         At index 1 diff: '.. py:class:: EnumCls(value)' != '.. py:class:: EnumCls(value, names=None, *, module=None, qualname=None, type=None, start=1, boundary=None)'
E         Use -v to get the full diff
[...]

PR #9443, opened 2021-07-13
Fix test_enum_class with Python 3.10-beta4

befeleme:
Subject: Fix test_enum_class with Python 3.10-beta4

### Feature or Bugfix
- Bugfix

### Detail
The fix was tested during rpm package build in Fedora, which then succeeded.
Because the changes were only postponed, I propose to keep the if/else clause, but removing it entirely is also an option.

### Relates
#9442

### DIFF
--- a/tests/test_ext_autodoc.py
+++ b/tests/test_ext_autodoc.py
@@ -1398,16 +1398,9 @@ def test_slots(app):
 def test_enum_class(app):
     options = {"members": None}
     actual = do_autodoc(app, 'class', 'target.enums.EnumCls', options)
-
-    if sys.version_info < (3, 10):
-        sig = '(value)'
-    else:
-        sig = ('(value, names=None, *, module=None, qualname=None, type=None, start=1, '
-               'boundary=None)')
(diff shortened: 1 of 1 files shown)


hroncok:
> Because the changes were only postponed, I propose to keep the if/else clause, but removing it entirely is also an option.

See this relevant discussion in pytest, where we decided to remove it entirely: https://github.com/pytest-dev/pytest/pull/8895

tk0miya:
Thanks! I did not notice the change.
+1 for removing it entirely too.

befeleme:
I've changed the code accordingly and tested that it builds in Fedora successfully.

tk0miya:
Hmm... It seems CI has been broken because 3.10-dev @ deadsnakes is still 3.10.0b2+. I don't know why b4 is not still packaged. Anyway, we need to wait for merging this until upgrading.

tk0miya:
Now 3.10-dev @ deadsnakes is updated. So I just merged this. Thank you for your contribution!

PR #9447, opened 2021-07-14
HTML Templates: Expose full sphinx version tuple

Blendify:

### Feature or Bugfix
- Feature

### Purpose
Previously, there was only a string which is problematic
when you want to programmatically compare versions
for compatibility reasons.

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -509,7 +509,7 @@ class StandaloneHTMLBuilder(Builder):
             'script_files': self.script_files,
             'language': self.config.language,
             'css_files': self.css_files,
-            'sphinx_version': sphinx_version,
+            'sphinx_version': __display_version__,
             'sphinx_version_tuple': sphinx_version,
             'style': self._get_style_filename(),
             'rellinks': rellinks,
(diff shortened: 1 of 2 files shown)


sphinx/builders/html/__init__.py
```diff
@@ -508,7 +509,8 @@ def prepare_writing(self, docnames: Set[str]) -> None:
             'script_files': self.script_files,
             'language': self.config.language,
             'css_files': self.css_files,
-            'sphinx_version': __display_version__,
+            'sphinx_version': sphinx_version,
```
astrojuanlu: Shouldn't the `sphinx_version` still have `__display_version__`? I see the need to have `'sphinx_version_tuple': sphinx_version`, but I don't understand why do we need to change the `sphinx_version` dictionary key.

Blendify: Yes you are right originally I changed the existing variable to be a tuple but I reverted that and added a new variable instead. I must have forgotten to revert this part of the code. I will fix that and update the patch.

tk0miya:

Issue #9428, opened 2021-07-10
Substitution messages in rst_epilog gets output in every POT files instead of only when used


alvinhochun:
### Describe the bug

The gettext build in Sphinx 4.0 outputs excessively unused messages to all the POT templates, as opposed to only when used in the page as in Sphinx 3.

### How to Reproduce

```console
$ # Clone https://invent.kde.org/documentation/docs-krita-org/-/tree/krita/4.3
$ git clone https://invent.kde.org/documentation/docs-krita-org.git -b krita/4.3
$ cd docs-krita-org
$ python3 -m venv .venv_sphinx3
$ . .venv_sphinx3/bin/activate
$ pip install sphinx==3.4.3
$ make gettext
$ # Check `404.pot` to see no messages from rst_epilog
$ grep rst_epilog _build/gettext/404.pot
$ # Check total occurences of rst_epilog
$ grep -r --include '*.pot' rst_epilog _build/gettext/ | wc -l
$ deactivate
$ rm _build/ -rf
$ python3 -m venv .venv_sphinx4
$ . .venv_sphinx4/bin/activate
$ pip install sphinx==4.0.3
$ make gettext
$ # Check `404.pot` to see plenty of messages from rst_epilog
$ grep rst_epilog _build/gettext/404.pot
$ # Check total occurences of rst_epilog
$ grep -r --include '*.pot' rst_epilog _build/gettext/ | wc -l
$ deactivate
```

### Expected behavior

`rst_epilog` messages should not be duplicated in every POT files even when it is unused in the page.

### Your project

https://invent.kde.org/documentation/docs-krita-org/-/tree/krita/4.3

### OS

Linux

### Python version

3.9

### Sphinx version

4.0.3

tk0miya:
Understandable.

At present, `rst_prolog` and `rst_epilog` are merged into each document before the parsing process. So Sphinx document processors (parsers, transforms, builders and so on) do not know the target document is consists of multiple sources. To make them separated, we need to change the structure of our components and its data structures.

alvinhochun:
Thanks for the reply. Matt I ask what changes between 3.4.3 and 4.0.3 regarding the handling of rst_prolog and rst_epilog that caused this difference in behaviour?

alvinhochun:
[...]

PR #9454, opened 2021-07-15
Exclude substitution definitions from gettext build

alvinhochun:
Subject: Exclude substitution definitions from gettext build

### Feature or Bugfix

- Bugfix

### Purpose

### Detail

### Relates
- #9428

### DIFF
--- a/sphinx/builders/gettext.py
+++ b/sphinx/builders/gettext.py
@@ -144,6 +144,17 @@ class I18nBuilder(Builder):
         return
 
     def write_doc(self, docname: str, doctree: nodes.document) -> None:
+        def is_node_part_of_substitution_definition(node: nodes.Node) -> bool:
+            """Check the node and its parents to see if any of them is a substitution
+            definition.
+            """
+            # return False
+            if isinstance(node, nodes.substitution_definition):
+                return True
+            if node.parent:
+                return is_node_part_of_substitution_definition(node.parent)
+            return False
+
         catalog = self.catalogs[docname_to_domain(docname, self.config.gettext_compact)]
 
         for toctree in self.env.tocs[docname].traverse(addnodes.toctree):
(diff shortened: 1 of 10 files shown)


alvinhochun:
Do you think this should target the 4.1.x branch instead?

alvinhochun:
Hi @tk0miya, do you have any comments on this change?

tk0miya:
Please wait for a while. I'll take a look soon (maybe until this weekend).

tk0miya:
I think this change makes all substitutions not translatable. -1 for this change.

alvinhochun:
> 
> 
> I think this change makes all substitutions not translatable. -1 for this change.

Sorry, you are right, this change does not work properly. However, I also found out that even without this change, some of the substitutions are still not translatable.

To make a correct version of this patch, I think we have to change `extract_messages` to check all the `substitution_reference` nodes and exclude the `substitution_definition` nodes that aren't being used (perhaps limited to those from `rst_prolog` and `rst_epilog`. However, one problem with the `gettext` builder is that the document tree that `extract_messages` receives already has the substitutions transformation applied, so it seems that I'll have to fix this too.

I now have a few test cases which should help. I will try to get this fixed correctly, but without making `rst_prolog` and `rst_epilog` their own PO catalog files as that may be too much of a change for now.

alvinhochun:
Hi, I pushed some test cases and a slightly more correct "fix" here (also to https://github.com/alvinhochun/sphinx/pull/1 in order to get CI results without waiting). As you can see over there, one of the tests is failing. That shows a pre-existing issue unrelated to my fix.

The cause of it is that when `I18nBuilder.write_doc` in `gettext.py`, the `doctree` it gets already has all the substitutions applied (i.e. passed through `docutils.transforms.references.Substitutions`), and it also has the image URLs "normalized". However when the translations are to be applied via `sphinx.transforms.i18n.Locale`, the document it uses still has `substitution_reference` intact.

I don't have an idea how to fix that properly, but the original "regression" I was concerned about should be solved here.

alvinhochun:
As far as I am concerned, this pull request should now be good. Please let me know if there are any more issues with it.

alvinhochun:
@tk0miya Any comments on the latest revision?

alvinhochun:
@tk0miya Ping, any comments?

therahedwig:
Does @alvinhochun  still need to change anything about this PR?

Issue #9471, opened 2021-07-19
New autodoc_class_signature config option gives an exception


miohtama:
### Describe the bug

The option causes an exception on build when used as documented

### How to Reproduce

When setting `autodoc_class_signature = "separated"` in conf.py

I get:

```
Exception occurred:
  File "/Users/moo/Library/Caches/pypoetry/virtualenvs/capitalgram--bm4msE3-py3.8/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 1452, in __init__
    self.options.special_members.append('__new__')
AttributeError: 'set' object has no attribute 'append'
```

### Expected behavior

The signature setting works or explains if there is something wrong with the configuration file.

### Your project

https://github.com/miohtama/capitalgram-onchain-dex-quant-data/blob/master/docs/source/conf.py

### OS

Mac

### Python version

3.0

### Sphinx version

4.1.1

### Sphinx extensions

sphinx.ext.autodoc

### Additional context

Probably this bit is faulty - cannot decide if the option is set or list
```python
    def __init__(self, *args: Any) -> None:
        super().__init__(*args)

        if self.config.autodoc_class_signature == 'separated':
            # show __init__() method
            if self.options.special_members is None:
                self.options['special-members'] = {'__new__', '__init__'}
            else:
                self.options.special_members.append('__new__')
                self.options.special_members.append('__init__')

        merge_members_option(self.options)
```

PR #9472, opened 2021-07-19
Always cast the special_members as a list

miohtama:
This is a potential fix for 

https://github.com/sphinx-doc/sphinx/issues/9471

https://github.com/sphinx-doc/sphinx/issues/9436

Autodoc is mixing `set` and `list` types for the same `self.options.special_members` variable.

However because the original code is not typed, it is hard to deduct from the context if the fix is "correct" or "clean". I suggest some of the original authors review the fix.

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -1449,6 +1449,8 @@ class ClassDocumenter(DocstringSignatureMixin, ModuleLevelDocumenter):  # type: 
             if self.options.special_members is None:
                 self.options['special-members'] = {'__new__', '__init__'}
             else:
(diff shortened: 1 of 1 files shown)


ruro:
@tk0miya I can no longer leave comments on #4257 for some reason (not sure, why did GitHub even allow me to open that code review in the first place 😅). Also, I didn't notice that somebody already reported this issue, when I was writing that comment, sorry for the confusion. Not sure, if the `special_members` was supposed to be a `set` or a `list` originally, but this PR seems to fix the issue.

tk0miya:
Oops... list should be used for this case. So set is invalid.

tk0miya:
Merged https://github.com/sphinx-doc/sphinx/pull/9493 instead.

Issue #9475, opened 2021-07-19
Can not build Sphinx own PDF docs since 4.1.0


jfbu:
### Describe the bug

Sphinx own PDF document build fails:

```
LaTeX Warning: Command \TeX invalid in math mode on input line 22151.

! You can't use `\spacefactor' in math mode.
\@->\spacefactor 
                 \@m {}
l.22151 ... that the system has a working \(\LaTeX
                                                  \) installation,
? 
```

https://github.com/sphinx-doc/sphinx/pull/9355/files#r672173343

### How to Reproduce

build Sphinx own docs

### Sphinx version

4.1.0, 4.1.1

astrojuanlu:
I see this in the RTD logs:

```
LaTeX Warning: Command \LaTeX invalid in math mode on input line 22894.


LaTeX Warning: Command \TeX invalid in math mode on input line 22894.

! You can't use `\spacefactor' in math mode.
\@->\spacefactor 
                 \@m {}
l.22894 ... that the system has a working \(\LaTeX
                                                  \) installation,
...
Latexmk: Log file says output to 'sphinx.pdf'
Collected error summary (may duplicate other messages):
  pdflatex: Command for 'pdflatex' gave return code 256
Latexmk: Errors, in force_mode: so I tried finishing targets
```

however, the job didn't fail. I wonder if we should consider this a bug on the RTD side @humitos

astrojuanlu:
URL https://readthedocs.org/projects/sphinx/builds/14214434/

mgeier:
@astrojuanlu 

> however, the job didn't fail. I wonder if we should consider this a bug on the RTD side

See also https://github.com/readthedocs/readthedocs.org/issues/7884.

jfbu:
> however, the job didn't fail. I wonder if we should consider this a bug on the RTD side @humitos

In this specific case there is (strangely) this line in latexmk console output:
```
Latexmk: Summary of warnings from last run of *latex:
  =====Latex reported missing or unavailable character(s).
```
Generally speaking, such a warning is indication the PDF is not 100% correct. But how to fix this will generally be not obvious. 

And in this specific case, this warning is rather misleading. We find a particularly strange message in the latex log:
```
Missing character: There is no è in font cmex10!
```
which is not of much help and is another symptom of `$\LaTeX$`. Besides the PDF output is not *that* broken in this specific case.

But generally speaking a 

```
  =====Latex reported missing or unavailable character(s).
```

[...]

PR #9476, opened 2021-07-19
Remove mark-up in Sphinx own docs breaking building it into PDF output

jfbu:
Close: #9475

Refs: #9355

Don't know if to issue PR on 4.1.x or 4.x branch, this is not hot problem as it regards only our own docs. Chose 4.1.x.


tk0miya (approved):

tk0miya:
👍

PR #9503, opened 2021-07-26
doc: :manpage: is a role, not a directive

Un1Gfn:
Subject: doc: :manpage: is a role, not a directive

### Feature or Bugfix

- Bugfix

### Purpose
- Fix documentation on <https://www.sphinx-doc.org/en/master/usage/configuration.html>

### Detail
- Title self-explanatory

### Relates
- https://www.sphinx-doc.org/en/master/usage/configuration.html

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -386,7 +386,7 @@ General configuration
 
 .. confval:: manpages_url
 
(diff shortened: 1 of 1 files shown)


tk0miya:
Thanks!

PR #9518, opened 2021-07-31
Fix autodoc_docstring_signature support for __init__ and __new__

jbms:
### Feature or Bugfix
- Bugfix

### Purpose
The `MethodDocumenter.get_doc` method added by
51ae283a38ed297df3fd5e0554dcc9acaa103e8c prevents
`DocstringSignatureMixin` from working correctly for `__init__` and
`__new__` methods.  Additionally, the `__new__` docstring was not obtained
correctly.

This commit checks for `self._new_docstrings` being set, and also
corrects the logic for obtaining the `__new__` docstring.

There still remains the issue that when the class signature is
obtained from the signature of `__init__` or `__new__`, only the real
signature is used, due to the use of `sphinx.util.inspect.signature`;
the `autodoc_docstring_signature` option does not have any effect.


tk0miya:
Thanks!

PR #9539, opened 2021-08-10
Proof of Concept: add "H" shortcut for hiding search matches

mgeier:
I don't think this is a good idea, I just want to illustrate how part of #9337 could be implemented using `accesskey`.

Rendered: https://sphinx--9539.org.readthedocs.build/en/9539/usage/configuration.html?highlight=search%20term


AA-Turner:
Closing to reduce the number of open PRs given it was noted as not a good idea -- the proof of concept will remain.

A

PR #9541, opened 2021-08-11
Updated EXAMPLES

vedranmiletic:
- removed or replaced missing docs
- removed docs that switched to using other tools
- added some new docs


tk0miya (changes requested):
Thank you for your work. LGTM!! But it seems CI gets errored. Could you check it, please?

vedranmiletic:
> 
> 
> Thank you for your work. LGTM!! But it seems CI gets errored. Could you check it, please?

Sorry, should have checked it better. Should be OK now.

tk0miya (approved):

tk0miya:

PR #9545, opened 2021-08-14
Update link to OpenSearch.

aaugustin:
Amazon repurposed the opensearch.org domain.

It is now hosting their ElasticSearch fork.


tk0miya:
I found the mention about the new repository here: https://opensearch.org/disambiguation.html
LGTM.

tk0miya:
Thanks!

PR #9551, opened 2021-08-15
"Hide Search Matches" removes "highlight" parameter from URL

mgeier:
When clicking "Hide Search Matches", this will modify the URL in the browser, removing the `?highlight=xyz` part (without reloading the page).

This makes it easier to share the link without highlighting.

This is a slight variation of a part of #9337. No new history entry is created, though.

Try it out: https://sphinx--9551.org.readthedocs.build/en/9551/extdev/appapi.html?highlight=sphinx

### DIFF
--- a/sphinx/themes/basic/static/doctools.js
+++ b/sphinx/themes/basic/static/doctools.js
@@ -264,6 +264,9 @@ var Documentation = {
   hideSearchWords : function() {
     $('#searchbox .highlight-link').fadeOut(300);
     $('span.highlighted').removeClass('highlighted');
+    var url = new URL(window.location);
+    url.searchParams.delete('highlight');
+    window.history.replaceState({}, '', url);
   },
 
   /**


humitos:
Does it make sense to keep the link and change it to "Show search matches" after clicked to recover the search highlight?

agjohnson:
A couple suggestions here, though I also don't see any new elements in the Sphinx build in the description -- should I?

I was going to suggest push state for this, but I suppose that is probably what you're referring to when you mention history events. If push state would be a middle ground to removing the query param from the address bar and also adding highlighting back, this might be worth trying.

There is still the issue that Sphinx's default highlighting makes for poor UX on a lot of search queries though:

image


I'm not sure if we can override this behavior easily, or if that even makes sense -- most likely this is better as an upstream patch at Sphinx though. I know this has been a hurdle to enabling the highlighting on RTD though.

I'd might vote to add a bit more UI here though (again, I'm just guessing). I think "Showing results for: `foo` -- <a>Hide results</a>" would be a great way to communicate both why things are highlighted and how to unhighlight them.

mgeier:
> Does it make sense to keep the link and change it to "Show search matches" after clicked to recover the search highlight?

I wouldn't want this in my theme, because the "Hide search matches" link is placed very prominently and taking quite a lot of space (in order to make it easy to see how to get rid of all the noise). I wouldn't want a "Show search matches" link to waste that same space.

For an example, see https://insipid-sphinx-theme.readthedocs.io/en/0.2.7/intro.html?highlight=search.

In other themes it might not waste that much space, but I'd argue that in most themes the "Hide search matches" link is hard to see anyway because it is too small.

Also, at least in my usage, the highlighting is only active for a short time, while the "Show search matches" link would be active as long as the page is open and most of the time it would just be a distraction.

mgeier:
Are there any objections to this?

If not, can this please be merged?

tk0miya:
+1: This must be useful!

PR #9555, opened 2021-08-17
fix: proper error messages if autosummary fails to import module

### LINKED ISSUE #7989: autosummary doesn’t report correct module on ImportError
**Describe the bug**
Import warnings in autosummary mention the same class twice instead of mentioning what actually caused the problem.

**To Reproduce**

1. create a module `a.py` that imports a nonexistant module `b.py`

   ```py
   import b

   class AClass: pass
   ```

2. document something in module `a` using autosummary:

   ```restructuredtext
   .. autosummary::

      a.AClass
   ```

**Actual behavior**

```
WARNING: [autosummary] failed to import 'a.AClass': no module named a.AClass
```

**Expected behavior**
```
WARNING: [autosummary] failed to import 'a.AClass': no module named b
```

**Environment info**
- OS: Linux
- Python version: 3.8
- Sphinx version: 3.1
- Sphinx extensions: sphinx.ext.autosummary

Pro:
Subject: Show proper error messages if autosummary fails to import module

Fixes #7989

### Feature or Bugfix

- Bugfix

### Purpose

As described in #7989, autosummary currently does not provide helpful error messages.

E.g., let's say I have a module called fancy_bob, which in its init.py uses `import filelock` (or any other external module).
Then in my docs I write:

```rst

.. autosummary::
   :toctree: _autosummary
   :recursive:

   fancy_bob
```

If the filelock module is not installed, autosummary will just fail with:

```
[autosummary] failed to import 'fancy_bob': no module named fancy_bob.
```

This error message is completely misleading, since it actually finds the module `fancy_bob` but then fails to import its dependencies.

This MR changes the error output in a way that you at least get some hints what went wrong, e.g., now you get:

```
[autosummary] failed to import 'fancy_bob': no module named fancy_bob. Possible hints: ["No module named 'filelock'"]
```

Or if there is a syntax error in the module's init.py, it will show:

```
[autosummary] failed to import 'fancy_bob': no module named fancy_bob. Possible hints: ['invalid syntax (__init__.py, line 4)']
```

### Reproduction

@astrojuanlu provided a nice example project to reproduce the issue in https://github.com/astrojuanlu/sphinx-autosummary-tracebacks

Not setting the pythonpath properly results in:
```
[autosummary] failed to import 'test_pkg.A': no module named test_pkg.A. Possible hints: ["No module named 'test_pkg'", 'no module named test_pkg']
```

Setting the pythonpath gives the correct error:
```
[autosummary] failed to import 'test_pkg.A': no module named test_pkg.A. Possible hints: ['no module named test_pkg', "No module named 'attr'"]
```

I.e., `"No module named 'attr'`

### Detail
- #7989

### Relates
- <URL or Ticket>

### DIFF
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -645,7 +645,7 @@ def _import_by_name(name: str, last_errors = None) -> Tuple[Any, Any, str]:
                 return getattr(mod, name_parts[-1]), mod, modname
             except (ImportError, IndexError, AttributeError) as e:
                 if last_errors is not None:
-                    last_errors.append(str(str(e.args[0])))
+                    last_errors.append(str(e))
                 pass
 
         # ... then as MODNAME, MODNAME.OBJ1, MODNAME.OBJ1.OBJ2, ...
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -677,7 +677,7 @@ def _import_by_name(name: str, last_errors = None) -> Tuple[Any, Any, str]:
         raise ImportError(*e.args) from e
 
 
-def import_ivar_by_name(name: str, prefixes: List[str] = [None]) -> Tuple[str, Any, Any, str]:
+def import_ivar_by_name(name: str, prefixes: List[str] = [None], last_errors = None) -> Tuple[str, Any, Any, str]:
     """Import an instance variable that has the given *name*, under one of the
     *prefixes*.  The first name that succeeds is used.
     """
@@ -690,7 +690,9 @@ def import_ivar_by_name(name: str, prefixes: List[str] = [None]) -> Tuple[str, A
         # check for presence in `annotations` to include dataclass attributes
         if (qualname, attr) in analyzer.attr_docs or (qualname, attr) in analyzer.annotations:
             return real_name + "." + attr, INSTANCEATTR, obj, modname
-    except (ImportError, ValueError, PycodeError):
+    except (ImportError, ValueError, PycodeError) as e:
+        if last_errors is not None:
+            last_errors.append(str(e.args[0]))
         pass
 
     raise ImportError
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -658,7 +658,7 @@ def _import_by_name(name: str, last_errors = None) -> Tuple[Any, Any, str]:
                 import_module(modname)
             except ImportError as e:
                 if last_errors is not None:
-                    last_errors.append(str(e.args[0]))
+                    last_errors.append(str(e))
                 continue
 
             if modname in sys.modules:
@@ -692,7 +692,7 @@ def import_ivar_by_name(name: str, prefixes: List[str] = [None], last_errors = N
             return real_name + "." + attr, INSTANCEATTR, obj, modname
     except (ImportError, ValueError, PycodeError) as e:
         if last_errors is not None:
-            last_errors.append(str(e.args[0]))
+            last_errors.append(str(e))
         pass
 
     raise ImportError
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -614,7 +614,7 @@ def get_import_prefixes_from_env(env: BuildEnvironment) -> List[str]:
     return prefixes
 
 
-def import_by_name(name: str, prefixes: List[str] = [None]) -> Tuple[str, Any, Any, str]:
+def import_by_name(name: str, prefixes: List[str] = [None], last_errors = None) -> Tuple[str, Any, Any, str]:
     """Import a Python object that has the given *name*, under one of the
     *prefixes*.  The first name that succeeds is used.
     """
@@ -625,14 +625,14 @@ def import_by_name(name: str, prefixes: List[str] = [None]) -> Tuple[str, Any, A
                 prefixed_name = '.'.join([prefix, name])
             else:
                 prefixed_name = name
-            obj, parent, modname = _import_by_name(prefixed_name)
+            obj, parent, modname = _import_by_name(prefixed_name, last_errors)
             return prefixed_name, obj, parent, modname
         except ImportError:
             tried.append(prefixed_name)
     raise ImportError('no module named %s' % ' or '.join(tried))
 
 
-def _import_by_name(name: str) -> Tuple[Any, Any, str]:
+def _import_by_name(name: str, last_errors = None) -> Tuple[Any, Any, str]:
     """Import a Python object given its full name."""
     try:
         name_parts = name.split('.')
@@ -643,7 +643,9 @@ def _import_by_name(name: str) -> Tuple[Any, Any, str]:
             try:
                 mod = import_module(modname)
                 return getattr(mod, name_parts[-1]), mod, modname
-            except (ImportError, IndexError, AttributeError):
+            except (ImportError, IndexError, AttributeError) as e:
+                if last_errors is not None:
+                    last_errors.append(str(str(e.args[0])))
                 pass
 
         # ... then as MODNAME, MODNAME.OBJ1, MODNAME.OBJ1.OBJ2, ...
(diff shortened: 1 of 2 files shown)


AndiEcker:
+1 thank you very much @Pro

sphinx/ext/autosummary/__init__.py
```diff
@@ -643,7 +643,9 @@ def _import_by_name(name: str) -> Tuple[Any, Any, str]:
             try:
                 mod = import_module(modname)
                 return getattr(mod, name_parts[-1]), mod, modname
-            except (ImportError, IndexError, AttributeError):
+            except (ImportError, IndexError, AttributeError) as e:
+                if last_errors is not None:
+                    last_errors.append(str(str(e.args[0])))
```
flying-sheep:
```suggestion
                    last_errors.append(str(e))
```

Don’t stringify twice. Also you can just rely on the builtin `__str__` method that also accesses `self.args[0]` but actually has a fallback in case `len(self.args) == 0` instead of crashing

sphinx/ext/autosummary/__init__.py
```diff
@@ -654,7 +656,9 @@ def _import_by_name(name: str) -> Tuple[Any, Any, str]:
             modname = '.'.join(name_parts[:j])
             try:
                 import_module(modname)
-            except ImportError:
+            except ImportError as e:
+                if last_errors is not None:
+                    last_errors.append(str(e.args[0]))
```
flying-sheep:
```suggestion
                    last_errors.append(str(e))
```

see above

sphinx/ext/autosummary/__init__.py
```diff
@@ -686,7 +690,9 @@ def import_ivar_by_name(name: str, prefixes: List[str] = [None]) -> Tuple[str, A
         # check for presence in `annotations` to include dataclass attributes
         if (qualname, attr) in analyzer.attr_docs or (qualname, attr) in analyzer.annotations:
             return real_name + "." + attr, INSTANCEATTR, obj, modname
-    except (ImportError, ValueError, PycodeError):
+    except (ImportError, ValueError, PycodeError) as e:
+        if last_errors is not None:
+            last_errors.append(str(e.args[0]))
```
flying-sheep:
```suggestion
            last_errors.append(str(e))
```

Here I’m not so sure: What’s `PycodeError`’s `__str__` method?

flying-sheep (changes requested):
Your stringification is a bit wonky, but generally good idea!

sphinx/ext/autosummary/__init__.py
Pro: Oh right, copy pasta error. Thanks!

sphinx/ext/autosummary/__init__.py
Pro: PycodeError already has a __str__ method, so that should be fine, right? No need to handle it different.

https://github.com/sphinx-doc/sphinx/blob/fe47bf46cbb6a615b01d6c5dc9fdcac58c940f69/sphinx/errors.py#L118

sphinx/ext/autosummary/__init__.py
```diff
@@ -614,7 +614,7 @@ def get_import_prefixes_from_env(env: BuildEnvironment) -> List[str]:
     return prefixes
 
 
-def import_by_name(name: str, prefixes: List[str] = [None]) -> Tuple[str, Any, Any, str]:
+def import_by_name(name: str, prefixes: List[str] = [None], last_errors = None) -> Tuple[str, Any, Any, str]:
```
tk0miya, line 617: I feel this change is strange because `last_errors` should be a part of the return value. I understand using argument gives us compatibility. But...

BTW, it should be annotated.

sphinx/ext/autosummary/__init__.py
```diff
@@ -643,7 +643,9 @@ def _import_by_name(name: str) -> Tuple[Any, Any, str]:
             try:
                 mod = import_module(modname)
                 return getattr(mod, name_parts[-1]), mod, modname
-            except (ImportError, IndexError, AttributeError):
+            except (ImportError, IndexError, AttributeError) as e:
+                if last_errors is not None:
+                    last_errors.append(str(e))
                 pass
```
tk0miya: This line is no longer needed.

sphinx/ext/autosummary/__init__.py
```diff
@@ -654,7 +656,9 @@ def _import_by_name(name: str) -> Tuple[Any, Any, str]:
             modname = '.'.join(name_parts[:j])
             try:
                 import_module(modname)
-            except ImportError:
+            except ImportError as e:
+                if last_errors is not None:
+                    last_errors.append(str(e))
```
tk0miya: I think it would be better to store exception object itself instead of string. It's not too late to convert it to string on creating message.

sphinx/ext/autosummary/__init__.py
```diff
@@ -686,7 +690,9 @@ def import_ivar_by_name(name: str, prefixes: List[str] = [None]) -> Tuple[str, A
         # check for presence in `annotations` to include dataclass attributes
         if (qualname, attr) in analyzer.attr_docs or (qualname, attr) in analyzer.annotations:
             return real_name + "." + attr, INSTANCEATTR, obj, modname
-    except (ImportError, ValueError, PycodeError):
+    except (ImportError, ValueError, PycodeError) as e:
+        if last_errors is not None:
+            last_errors.append(str(e))
         pass
```
tk0miya: This line is also.

sphinx/ext/autosummary/generate.py
```diff
@@ -413,16 +413,20 @@ def generate_autosummary_docs(sources: List[str], output_dir: str = None,
         path = output_dir or os.path.abspath(entry.path)
         ensuredir(path)
 
+        last_errors = list()
```
tk0miya: Basically, we use `[]` instead of `list()`.

sphinx/ext/autosummary/__init__.py
Pro: Also thought the same, but that would probably break backwards compatibility? Or should I change last_errors to be a retun value?

sphinx/ext/autosummary/generate.py
```diff
@@ -413,16 +413,22 @@ def generate_autosummary_docs(sources: List[str], output_dir: str = None,
...
         except ImportError as e:
             try:
-                # try to importl as an instance attribute
-                name, obj, parent, modname = import_ivar_by_name(entry.name)
+                # try to import as an instance attribute
+                name, obj, parent, modname = import_ivar_by_name(entry.name, last_errors=last_errors)
                 qualname = name.replace(modname + ".", "")
             except ImportError:
-                logger.warning(__('[autosummary] failed to import %r: %s') % (entry.name, e))
+                # Convert exceptions to strings and remove duplicates (covert to set, then to list)
```
tk0miya:
```suggestion
                # Convert exceptions to strings and remove duplicates (convert to set, then to list)
```

sphinx/ext/autosummary/generate.py
```diff
@@ -413,16 +413,22 @@ def generate_autosummary_docs(sources: List[str], output_dir: str = None,
...
+                last_errors_str = list([str(e) for e in last_errors])
+                last_errors_str = list(set(last_errors_str))
```
tk0miya: How about this?
```suggestion
                last_errors_str = list(set(str(e) for e in last_errors))
```

sphinx/ext/autosummary/__init__.py
tk0miya: I have two ideas:

1. Add another version of `import_by_name` that also returns list of errors. It will be named as different name.
2. Add a boolean argument like `raise_error` to `import_by_name`. And it raises a custom exception that contains list of errors.

Pro: Hmm, raise_error would not work that well since we need to collect all errors during the various import methods there are.

Having another version of the import_by_name is IMHO also not such an ideal solution, since then we would need to duplicate code, not only for the `import_by_name` but also for the sub-methods such as `_import_by_name` which is called inside of `import_by_name`

Pro:
@tk0miya is there still something missing to merge this MR?

tk0miya:
>Add a boolean argument like raise_error to import_by_name. And it raises a custom exception that contains list of errors.

I just post #10031 to implement my idea.

Pro:
> > Add a boolean argument like raise_error to import_by_name. And it raises a custom exception that contains list of errors.
> 
> I just post #10031 to implement my idea.

Cool! Also nice solution!
I will close this MR in favor of yours.

PR #9565, opened 2021-08-20
Check complete ancestry of text nodes for smartquotes eligibility.

### LINKED ISSUE #9564: Smart quotes get applied to highlighted inline code expressions
### Describe the bug

Given some RST that includes text with a `:code:` role and a language option, sphinx will produce HTML with highlighting for that text's code language. However, if that highlighting introduces more than one sub-level of output nodes, we fail to detect that the text is part of a no-smartquotes zone and mangle text that is intended to be presented literally.

### How to Reproduce

The following RST:
```rst
.. role:: python(code)
   :language: python

By setting :python:`foo = 'bar'`, we have assigned a variable.
```
produces HTML like
```html
By setting
<code class="code python docutils literal notranslate">
  <span class="name"><span class="pre">foo</span></span>
  <span class="operator"><span class="pre">=</span></span>
  <span class="literal string single"><span class="pre">‘bar’</span></span>
</code>, we have assigned a variable.
```
Notice the right/left single quotes added by Sphinx in **‘**bar**’**. Because the text node is two ancestors away from the document's `code` node instead of one, it isn't detected as protected from smartquotes.

### Expected behavior

Straight single and double-quote characters in inline code expressions should be left alone, regardless of whether or not a specific language option is applied.

docutils' built-in smartquotes transformer correctly avoids the highlighted quotation marks.

### Your project

Small Sphinx 4 project sample attached.

### Screenshots

_No response_

### OS

macOS

### Python version

3.9

### Sphinx version

4.1.2

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

JustinTArthur:
Subject: Fix erroneous application of smartquotes to highlighted inline code roles.

### Feature or Bugfix
Bugfix

### Purpose
To prevent the smartquotes transformer from transforming text meant to be presented literally, such as text found with the `:code:` role and a language option set.

### Detail
- sphinx's smartquotes transformer already checks text to see if the text's document node is a direct descendent of an element that shouldn't have typographical adjustments. However, it wasn't looking at the full ancestry, so if the text was a sub-sub-element of a document node meant to hint literal presentation, the text ended up having smartquotes' typographical adjustments applied.
- docutils' own smartquotes transformer does not currently have this problem

### Relates
- Fixes sphinx-doc/sphinx#9564


JustinTArthur:
Unfortunately, the test build was cancelled. I'll need @tk0miya or another maintainer to restart the build for commit 1762b9b ([CircleCI pipeline 4598](https://app.circleci.com/pipelines/github/sphinx-doc/sphinx/4598/workflows/34d05398-a4c7-444b-adc5-c216c2d9bbd9))

tk0miya:
Absolutely! Thank you for your contribution!

Issue #9430, opened 2021-07-11
manpage and html builders:  same rst file, different output


hkuno:
**Is your feature request related to a problem? Please describe.**
Many well-established software library projects that predate RST have a large number of existing *roff-style man pages.
Such projects often have both online documentation on the web and also man pages for use on linux operating systems.
When such packages choose to switch their man pages from nroff to RST, they may want to use a single set of RST files as source to produce both their online documentation (html files) and also their linux man pages (nroff files).

However, the manpage.py builder currently always adds a subtitle with a description of the command that is provided as input to config and the html builders do not do this.

This means that the RST author currently has to produce different RST source files for the manpage and the html  builders, 
because if the RST file contains a description of the command then that description will appear twice in the generated man page, and if the RST file does NOT contain a description of the command, then there will be no description in the generated html file.

For example, @jsquyres and I would like to generate both groff files and also html files from a single set of rst man page files for the Open-MPI project (https://github.com/open-mpi/). Here is a simple example: https://github.com/jsquyres/ompi-sphinx-dist/blob/main/docs/src/ompi-man/man3/MPI_Abort.3.rst

If rst/conf.py contains the following:
```
man_pages=[("ompi/mpi/man/man3/MPI_Abort.3","MPI_Abort", "(from conf.py) Terminates MPI execution environment","",3)]
```

... then the generated man page will look like this:
```
MPI_ABORT(3)                            Open MPI                            MPI_ABORT(3)

NAME
       MPI_Abort - (from conf.py) Terminates MPI execution environment

       MPI_Abort - Terminates MPI execution environment.

SYNTAX
   C Syntax
          #include <mpi.h>
          int MPI_Abort(MPI_Comm comm, int errorcode)
...
```

If we omit the "MPI_Abort - Terminates MPI execution environment." line from the RST file, then the generated html page will lose that command description.

**Describe the solution you'd like**
Could sphinx-build  allow users to direct the manpage builder to not print out the descriptive subtitle by submitting an empty description string? For example, see submitted PR #9594.

**Describe alternatives you've considered**
We initially proposed requesting that sphinx-build include a configuration option to not add the descriptive subtitle to the generated man pages, but we closed that PR #9431 because the solution in PR #9594 seems cleaner.

We considered writing a script to parse our rst files and generate separate sets of rst files for use by the manpage builder versus the html file builder. This would mean we'd have two separate sets of RST files.

[...]

PR #9588, opened 2021-08-29
manpage builder skip printing blank descriptions

hkuno:
Update manpage builder to skip printing the description of a command as a subtitle if the supplied description is blank.
This commit addresses #9430

### Feature or Bugfix
- Bugfix

### Purpose
Many well-established software library projects that predate RST have a large number of existing *roff-style man pages.
Such projects often have both online documentation on the web and also man pages for use on linux operating systems.
When such packages choose to switch their man pages from nroff to RST, they may want to use a single set of RST files as source to produce both their online documentation (html files) and also their linux man pages (nroff files).

However, the manpage.py builder currently always adds a subtitle with a description of the command that is provided as input to config and the html builders do not do this.

This means that the RST author currently has to produce different RST source files for the manpage and the html  builders, 
because if the RST file contains a description of the command then that description will appear twice in the generated man page, and if the RST file does NOT contain a description of the command, then there will be no description in the generated html file.

For example, @jsquyres and I would like to generate both groff files and also html files from a single set of rst man page files for the Open-MPI project (https://github.com/open-mpi/). Here is a simple example: https://github.com/jsquyres/ompi-sphinx-dist/blob/main/docs/src/ompi-man/man3/MPI_Abort.3.rst . Some examples of the original nroff files are here: https://github.com/open-mpi/ompi/tree/master/ompi/mpi/man/man3 

If rst/conf.py contains the following:
```
man_pages=[("ompi/mpi/man/man3/MPI_Abort.3","MPI_Abort", "(from conf.py) Terminates MPI execution environment","",3)]
```

... then the generated man page will look like this:
```
MPI_ABORT(3)                            Open MPI                            MPI_ABORT(3)

NAME
       MPI_Abort - (from conf.py) Terminates MPI execution environment

       MPI_Abort - Terminates MPI execution environment.

SYNTAX
   C Syntax
          #include <mpi.h>
          int MPI_Abort(MPI_Comm comm, int errorcode)
...
```

If we omit the "MPI_Abort - Terminates MPI execution environment." line from the RST file, then the generated html page will lose that command description.

This PR would make it possible for us to use sphinx-build to produce both HTML and manpage files from a single set of RST files.

### Detail
 - #9430

### Relates
- Please see issue #9430


hkuno:
I'm sorry. I made a mistake. I should have branched off of 4.1.x instead of master. I will close this PR and open a new one.

PR #9615, opened 2021-09-06
Document `rst-class`

astrojuanlu:
### Feature or Bugfix

- Feature

### Purpose

@nienn and I were looking for a way to add a custom CSS class to some content in Sphinx. We both arrived to [this Stack Overflow question](https://stackoverflow.com/a/33497002/554319), but otherwise it was impossible to find in the Sphinx docs. I think this directive is too important to be explained in a tiny footnote, so this is my attempt to document it.

I know it's technically incorrect that `rst-class` belongs to docutils... But from the user perspective, that's totally irrelevant, since they can't use it anyway.

### DIFF
--- a/doc/usage/restructuredtext/basics.rst
+++ b/doc/usage/restructuredtext/basics.rst
@@ -370,7 +370,7 @@ Docutils supports the following directives:
   - :dudir:`include` (include reStructuredText from another file) -- in Sphinx,
     when given an absolute include file path, this directive takes it as
     relative to the source directory
-  - :dudir:`class` (assign a class attribute to the next element) [1]_
+  - :rst:dir:`rst-class` (assign a class attribute to the next element) [1]_
 
 * HTML specifics:
 
@@ -605,6 +605,6 @@ There are some problems one commonly runs into while authoring reST documents:
 
 .. rubric:: Footnotes
 
-.. [1] When the default domain contains a :rst:dir:`class` directive, this
-       directive will be shadowed.  Therefore, Sphinx re-exports it as
+.. [1] Since some domains contain a ``class`` directive, it shadows
+       :dudir:`class` coming from docutils.  Therefore, Sphinx re-exports it as
        :rst:dir:`rst-class`.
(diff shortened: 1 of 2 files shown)


doc/usage/restructuredtext/basics.rst
```diff
@@ -370,7 +370,7 @@ Docutils supports the following directives:
   - :dudir:`include` (include reStructuredText from another file) -- in Sphinx,
     when given an absolute include file path, this directive takes it as
     relative to the source directory
-  - :dudir:`class` (assign a class attribute to the next element) [1]_
+  - :rst:dir:`rst-class` (assign a class attribute to the next element) [1]_
```
tk0miya, line 373: This is a list of docutils-directives. Not for sphinx custom directives. So this should not be replaced.

doc/usage/restructuredtext/basics.rst
```diff
@@ -605,6 +605,6 @@ There are some problems one commonly runs into while authoring reST documents:
 
 .. rubric:: Footnotes
 
-.. [1] When the default domain contains a :rst:dir:`class` directive, this
-       directive will be shadowed.  Therefore, Sphinx re-exports it as
+.. [1] Since some domains contain a ``class`` directive, it shadows
```
tk0miya, line 608: "some domains contain a class directive" is not related to the shadowing. So this modification is not good. The default_domain causes it. By default, the default_domain; `py` considers a `class` directive call as a `py:class` directive call. Other domains are not related in this case.

tk0miya (changes requested):

doc/usage/restructuredtext/basics.rst
astrojuanlu: You're right: I was confused because `rst-class` is registered always, along with `toctree` and others:

https://github.com/sphinx-doc/sphinx/blob/3d8fbd992cedf1784b106cf43ecbc452953ebe17/sphinx/directives/other.py#L386

But the current wording says "this directive will be shadowed". I see why the change is not good.

astrojuanlu:
Thanks @tk0miya for the review, I had misunderstood part of the documentation. I'll rather close this pull request and maybe open another in the future if I think of a better way of documenting this.

tk0miya:
IMO, it's good to add an explanation of `rst-class` directive. It must be worthy. So let's reopen this and merge it partially.

PR #9617, opened 2021-09-07
themes: use noscript for search js notification

jdknight:
### Feature or Bugfix
- Feature

### Purpose
The "search.html" template will generate a warning admonition about requiring JavaScript for search and automatically hiding the element when supported. While functional, if a client renders the page slowly, the warning notification may be visible to the user for a moment.

Instead of relying on JavaScript to suppress this warning, use a `noscript` tag to hide the warning for clients who do not support JavaScript. This also has the benefit of one less JavaScript call required by a client.

### Detail
The following example shows the loading of a documentation, where the search warning is temporarily visible:

_(during page load)_
image

_(after page load)_
image

The switch to `noscript` tags prevents this.

### DIFF
--- a/sphinx/themes/basic/search.html
+++ b/sphinx/themes/basic/search.html
@@ -21,13 +21,14 @@
 {% block body %}
   <h1 id="search-documentation">{{ _('Search') }}</h1>
   {% block scriptwarning %}
-  <div id="fallback" class="admonition warning">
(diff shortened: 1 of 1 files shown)


tk0miya:
Merged this into the 4.x branch manually. Thank you for your contribution!

PR #9628, opened 2021-09-12
Close #9623: Allow to suppress warnings on excluded document found in toctree

### LINKED ISSUE #9623: Give the ability to supress 'toctree contains reference to excluded document'
I'm generating multiple overlapping documents from a shared set of source `rst` files like in #9596 and I would like to use warnings-as-errors. I also have only one shared `conf.py` and control which documents to exclude by passing tags to sphinx-build:

```python
# somewhere in conf.py

exclude_patterns = []

if tags.eval_condition('not include_statistics'):
    exclude_patterns.append('kernel/statistics.rst')
```

The problem is that I include `statistics.rst` in toctree and sphinx triggers a warning 'toctree contains reference to excluded document', but this warning doesn't have type and subtype so I can supress it. It is not an error, it was excluded intentionally, I want `statistics.rst` to be in `toctree` because there is another document which uses `include_statistics` tag and I that's why I don't want warning here.

It could be special metadata field `:maybe_excluded:` which you can set for toctree entry to supress warning inplace only:

```rst

.. toctree::

    :maybe_excluded:statistics
    first_doc
    second_doc

```

Passing a type and subtype argument to logger.warning so I can supress it in `conf.py` globally (not so good, because they can be useful).

Now I have to patch `toctree` directive and override it in my extrension to supress this warning (very bad).

tk0miya:
### Feature or Bugfix
- Feature
- Refactoring

### Purpose
- Close #9623: Allow to suppress warnings on excluded document found in toctree
- Fix i18n: messages in sphinx.directives.other are not translated
- refactor: Use logger.warning() instead of reporter.warning()


sphinx/directives/other.py
```diff
@@ -106,9 +107,8 @@ def parse_content(self, toctree: addnodes.toctree) -> List[Node]:
                     toctree['entries'].append((None, docname))
                     toctree['includefiles'].append(docname)
                 if not docnames:
-                    ret.append(self.state.document.reporter.warning(
-                        'toctree glob pattern %r didn\'t match any documents'
-                        % entry, line=self.lineno))
+                    logger.warning(__('toctree glob pattern %r didn\'t match any documents'),
```
astrojuanlu, line 110: Out of curiosity, what are the implications of using `logger.warning` rather than `document.reporter.warning`, if any?

tk0miya: There are some advantages of `sphinx.util.logging`:

* It can be used even if no reporter object.
* It supports additional features like log-type, location, and so on.

In this case, I used it to make this log message suppressible.

PR #9629, opened 2021-09-13
Exclude substitution definitions from gettext build (take 2)

alvinhochun:
### Feature or Bugfix

- Bugfix

### Purpose

### Detail

### Relates
- #9428

Supersedes #9454

cc @therahedwig


alvinhochun:
Hi @tk0miya, will you be willing to review this change if I recreate the pull request again? I have been waiting for this to be merged.

PR #9635, opened 2021-09-15
Do not add a language fallback by the stripping country code

### LINKED ISSUE #9634: When building for zh_TW, zh_CN is incorrectly used as fallback language
### Describe the bug

In `sphinx/locale/__init__.py:26` it uses the *language* part of the locale code as a fallback, so using `zh_TW` means `[ 'zh_TW', 'zh' ]` will be passed to `gettext.translation`.

In CPython, `gettext.translation` uses `gettext.find`, which finds the `.mo` file for each language specified. For `zh_TW` it finds the `.mo` file for `zh_TW` normally. However, when it gets to `zh` it first normalizes the locale code to `zh_CN`, which means it ends up finding the `.mo` file for `zh_CN` and adding that as a fallback.

`gettext.translation` already includes the language-without-country-code fallback. The "fallback" that Sphinx adds is unneeded and incorrect.

### How to Reproduce

```
$ https://invent.kde.org/documentation/docs-krita-org.git -b krita/5.0
$ cd docs-krita-org
$ pip install sphinx
$ sphinx-intl build
$ make html SPHINXOPTS="-D language=zh_TW"
$ # open _build/html/user_manual/getting_started.html and see Simplified Chinese mixed in the text
```

### Expected behavior

Untranslated text shall use the original English text and not "fall back" to Simplified Chinese.

### Your project

https://invent.kde.org/documentation/docs-krita-org

### Screenshots

<details>
<summary>Expected:</summary>

圖片
</details>

<details>
<summary>Incorrect:</summary>

圖片
</details>

### OS

Linux

### Python version

Python 3.9

### Sphinx version

4.2

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

alvinhochun:
### Feature or Bugfix

- Bugfix

### Purpose
- Fix incorrect language fallback which may cause the wrong language to be used as a fallback.

### Detail
- When the specified language contains a country code (e.g. zh_TW), do not strip the country code and add that as a fallback.

### Relates
- Fixes https://github.com/sphinx-doc/sphinx/issues/9634


alvinhochun:
@tk0miya Is there anything blocking this change?

uranusjr:
I think the root cause here is that `zh_TW` and `zh_CN` are not full IETF tags. They should be better formed as `zh-Hant_TW` and `zh-Hans_CN`, and providing `zh-Hant_*` as fallbacks to `zh-Hant_TW` (e.g. `zh-Hant_HK`) is correct, while `zh-Hans_CN` will only be providing fallback to `zh-Hans` languages (e.g. `zh-Hans_SG`) and not interfere with `zh-Hant` languages. IIRC Babel (which Sphinx uses for i18n) supports the full IETF syntax, so perhaps a correct solution is _not_ to change Sphinx’s fallback logic (which is arguably correct), but to change the locale names in your Sphinx i18n setup.

alvinhochun:
> I think the root cause here is that `zh_TW` and `zh_CN` are not full IETF tags. They should be better formed as `zh-Hant_TW` and `zh-Hans_CN`, and providing `zh-Hant_*` as fallbacks to `zh-Hant_TW` (e.g. `zh-Hant_HK`) is correct, while `zh-Hans_CN` will only be providing fallback to `zh-Hans` languages (e.g. `zh-Hans_SG`) and not interfere with `zh-Hant` languages. IIRC Babel (which Sphinx uses for i18n) supports the full IETF syntax, so perhaps a correct solution is _not_ to change Sphinx’s fallback logic (which is arguably correct), but to change the locale names in your Sphinx i18n setup.

Fro what I get, Sphinx only uses Babel to produce MO files from the PO files. The part which translates messages is still using the `gettext` module from the Python standard library, which [uses the `locale` module](https://github.com/python/cpython/blob/cf7eaa4617295747ee5646c4e2b7e7a16d7c64ab/Lib/gettext.py#L216), which [uses glibc locale codes](https://github.com/python/cpython/blob/cf7eaa4617295747ee5646c4e2b7e7a16d7c64ab/Lib/locale.py#L872), not BCP 47 / IETF language tag.

If you actually trace the code ([1](https://github.com/python/cpython/blob/cf7eaa4617295747ee5646c4e2b7e7a16d7c64ab/Lib/gettext.py#L511), [2](https://github.com/python/cpython/blob/cf7eaa4617295747ee5646c4e2b7e7a16d7c64ab/Lib/gettext.py#L483), [3](https://github.com/python/cpython/blob/cf7eaa4617295747ee5646c4e2b7e7a16d7c64ab/Lib/gettext.py#L218)), you will see why Sphinx's fallback logic is incorrect. If you pass `zh` to `gettext.translation`, it will resolve to `zh_CN`, which leads to the incorrect fallback behaviour.

uranusjr:
That’s unfortunate. Maybe we should fix that in Sphinx :p It should be a good thing but I don’t know if the scope is too large; if it is, a better approach is probably to build in some kind of mapping logic to build fallbacks, instead of blindly splitting `_`.

AA-Turner:
@alvinhochun -- should we instead special-case this for ``zh``? For e.g. ``fr``, ``de``, ``en``, the current behaviour is likely desirable as the difference in dialects between countries better than having no translations.

A

uranusjr:
I believe there are other similar situations. Serbian, for example, has two writing systems (Latin and Cyrillic), and using one as fallback to another would be also weird. But I guess there are few enough languages having this problem that special-casing each of them can be technically reasonable.

alvinhochun:
> @alvinhochun -- should we instead special-case this for `zh`? For e.g. `fr`, `de`, `en`, the current behaviour is likely desirable as the difference in dialects between countries better than having no translations.

For those locales you will typically have translations use locale codes without a country code, this way the normal fallback already works, e.g. `pt_BR` -> `pt`, `en_GB` -> `en`. As I wrote in the linked issue:

> `gettext.translation` already includes the language-without-country-code fallback. The "fallback" that Sphinx adds is unneeded and incorrect.

AA-Turner (approved):

Issue #9574, opened 2021-08-23
Canonical links a.k.a redirection


marxin:
Let's consider an option like:

```rst
:option:`-fmy-option`
```

The option documentation lives somewhere here:
`example.com/foo/bar/baz/a-manual-page.html#cmdoption-fmy-option`

My question is if Sphinx can emit redirection for links like `example.com/redirect/cmdoption-fmy-option`?
Having that, one can then rename any component in `foo/bar/baz/a-manual-page.html` and the links is gonna be stable.

astrojuanlu:
@marxin have a look at https://sphinxext-rediraffe.readthedocs.io/, https://documatt.gitlab.io/sphinx-reredirects/, or https://docs.readthedocs.io/en/stable/automatic-redirects.html.

marxin:
Thank you for the tips. However, I'm not sure any of these can help with the `example.com/redirect/cmdoption-fmy-option`.
I would need something that can inspect `searchindex.js` or `sphinx.ext.intersphinx`'s `object.inv`:

```
$ python3 -msphinx.ext.intersphinx example.com
...
std:cmdoption
        -Wbuiltin-macro-redefined                gcc-command-options/options-to-request-or-suppress-warnings.html#cmdoption-Wbuiltin-macro-redefined
        -Wc++-compat                             gcc-command-options/options-to-request-or-suppress-warnings.html#cmdoption-Wc-compat
        -Wc++11-compat                           gcc-command-options/options-to-request-or-suppress-warnings.html#cmdoption-Wc-11-compat
        -Wc++11-extensions                       gcc-command-options/options-to-request-or-suppress-warnings.html#cmdoption-Wc-11-extensions
        -Wc++14-compat                           gcc-command-options/options-to-request-or-suppress-warnings.html#cmdoption-Wc-14-compat

...
```

jakobandersen:
@marxin, can you elaborate a bit on the use case?
If an object is moved to a different page then the roles will be resolved to the new location, so I guess your interest is in when another project, B, wants to link to your project, A, via intersphinx? In that case the links can indeed break until the other project recompiles with an updated ``objects.inv``. Is that the case you mean?
Assuming that, I think two parts are needed:
1. A needs to gain this redirect-by-index page.
2. intersphinx during compilation of B should not use the normal URLs, but always link to the redirect page.

[...]

PR #9649, opened 2021-09-18
Searchindex fix when objects have the same name

jakobandersen:
### Feature or Bugfix
- Bugfix

### Purpose
Store all objects with the same name from different domains in ``searchindex.js``.

### Detail
Objects in ``searchindex.js`` are stored as a dictionary indexed by the object name. However, it is perfectly reasonable to have, e.g., both a Python and C function with the same name. This PR changes the storage to an array.

To test out:
1. Clone https://github.com/jakobandersen/sphinx_redirect_prototype and build to HTML.
2. Search for "f". This should list at least entries for ``fpy``, ``fc``, and ``fcpp``.
3. Search for ``g``. This only lists the Python function, even though there is a C and C++ function of the same name.

Repeat the same steps with this PR should in step 3 yield all three ``g`` objects.

### Relates
Started from https://github.com/sphinx-doc/sphinx/issues/9574#issuecomment-922267316..

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -24,6 +24,8 @@ Bugs fixed
 * #9630: autosummary: Failed to build summary table if :confval:`primary_domain`
   is not 'py'
 * #9670: html: Fix download file with special characters
+* HTML search: when objects have the same name but in different domains,
+  return all of them as result instead of just one.
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,9 @@ Dependencies
 Incompatible changes
 --------------------
 
+* #9649: ``searchindex.js``: the embedded data has changed format to allow
+  objects with the same name in different domains.
+
 Deprecated
 ----------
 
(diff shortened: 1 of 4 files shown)


tk0miya:
LGTM. I'm not sure why the names are unique. I guess there is no reason to keep keywords unique. So +1 for this change.

BTW, we need to discuss about the side effect of this change. After this change, the structure of searchindex.js will be changed. It would be a breaking change if some 3rd party tool directly reads. AFAIK, there are no such tools. So I'm okay to merge this into the 4.x branch. But we need to announce it via CHANGES.

tk0miya (approved):

jakobandersen:
@tk0miya, I have added a note to CHANGES to warn about it. While it is indeed a breaking change I think the chance of someone using ``searchindex.js`` is somewhat low. The code calls a method on a specific object, passing the index as an object, so if someone is using it they anyway need to mock the interface assumed.
Let me know if you are fine with the PR with this CHANGES entry (or feel free to merge it your self).

tk0miya (approved):
LGTM. Merging!

jakobandersen:
Thanks!

Issue #9655, opened 2021-09-20
Suppress mock object warnings from autodoc


sdhiscocks:
**Is your feature request related to a problem? Please describe.**
When building with warnings as errors (`-W` flag), mock objects cause build failure (for example below).
```
WARNING: A mocked object is detected: 'example.mock.object'
```
This occurs with class attributes, or enum members, pointing to mock objects, even if they are documented with `#:`, which produce valid documentation.

**Describe the solution you'd like**
As suggested in #9631, it would be useful to have a `subtype` on the warning (e.g. `mock_objects`) so these warnings can be ignored.

**Describe alternatives you've considered**
Suppressing all autodoc warnings works, but this isn't ideal.

**Additional context**
- #9631

tk0miya:
I think the warnings should not be emitted if the mocked object can produce valid docs. I think it's a bug of autodoc that emits incorrect warnings. So I don't think it's better to suppress them.

Could you share an example to reproduce the warning? I'll try to fix it.

sdhiscocks:
So issue we've seen is with [this Enum we have defined](https://github.com/dstl/Stone-Soup/blob/66055168998f599da52fe10c0893def084cc5157/stonesoup/detector/tensornets.py#L30-L48), where the enum name points to a (mocked) function. The [Enum documentation looks fine](https://stonesoup.readthedocs.io/en/v0.1b6/stonesoup.detector.html#stonesoup.detector.tensornets.Networks) despite the warning.

We have a [recent build on CircleCI](https://app.circleci.com/pipelines/github/dstl/Stone-Soup/1500/workflows/1a11b011-d507-49af-812d-377f209ada40/jobs/8951) which got caught on this warning, but no issues with [documentation it produced](https://8951-98420803-gh.circle-artifacts.com/0/docs/stonesoup.detector.html#stonesoup.detector.tensornets.Networks).

tk0miya:
Thank you for example. I understand what happened. The enum subclass provides `_generate_next_value_()` method that generates value from `tensornets` module (it's mocked!) for `enum.auto()`. I also reproduced.

Note: Dockefile to reproduce the error:
```
FROM python:3.8-slim

RUN apt update; apt install -y build-essential curl git unzip vim
RUN git clone https://github.com/dstl/Stone-Soup
WORKDIR Stone-Soup
RUN pip install -e .[dev]
WORKDIR /Stone-Soup/docs
RUN make HTML
```

PR #9656, opened 2021-09-20
Fix issue with warnings without subtype being incorrectly suppressed

sdhiscocks:
### Feature or Bugfix
Bugfix

### Purpose
This fixes an issue with warnings that have been raised with no subtype being suppressed if a suppress warnings value with a subtype has been set. e.g. all `autodoc` warnings should not be suppressed if `autodoc.import_object` is only set to be suppressed.

### Detail
This issue occurred as logic previously only checked if error `subtype` was `None`, without checking that the `subtarget` was also `None`.

### Relates
Noticed this issue when looking at #9655


tk0miya:

Issue #9636, opened 2021-09-15
``:dedent: 0`` not respected and :dedent: without argument seems broken


ghost:
### Describe the bug

# `:dedent: 0` not respected

The `.. code-block::` (and probably related directives, too) does not respect the ``:dedent: 0`` option.
Instead of doing nothing, it does the same as ``:dedent:`` (without argument).
I believe ``:dedent: 0`` should be a noop to be in line with the expectation of dedenting by 0 characters.

Use case:

```
First part

.. code-block::

   class PythonClass:

Second part

.. code-block::
   :dedent: 0

       def __init__():
           pass

third part
```
Without the :dedent: the code block would (by I think the ReST parser, not the renderer) be read with ONE 7 space indent (instead of ONE 3 space indent and a 4 space indent inside the code block). Thus, messing up the code layout.  The :dedent: 0 would help in such cases as it indicates ONE 3 space indent for the directive.
As a workaround `:class: dummy` can also be used, however, this pollutes the class attribute of the node.

## Proposed solution:

Instead of `if not dedent:` use `if dedent is None:` in sphinx/directives/code.py line 60

# :dedent: (without argument)

Also, I noticed that :dedent: (without any argument) seemingly produces garbage, namely it removes all newlines and only the indent of the first line, producing a single long line.

## Proposed solution

The reason seems to be that the function dedent_lines() (sphinx/directives/code.py line 59) uses
```
        return textwrap.dedent(''.join(lines)).splitlines(True)
```
instead of the line preserving
```
        return textwrap.dedent('\n'.join(lines)).split('\n')
```

# Closing words

If the above analysis is correct and you agree with the proposed solutions, I'm happy to provide a PR if that reduces the workload for you.

### How to Reproduce

MWE is given above in the use case.

### Your project

Sorry that's a private one, the MWE above should do.

### OS

Linux

### Python version

3.9

### Sphinx version

4

tk0miya:
I don't think this is a bug because your goal is not "dedenting" the content by 0. I think you'd like to use the `:dedent: 0` option as a workaround to keep indentation, right? To fix this properly, we need to add a new syntax or parsing rules to reStructuredText.

But, at this moment, supporting 0 value is a good workaround for your case. And it is low cost to implement it and it does not harm anybody. So I can agree to add the change as a "secret" feature.

>If the above analysis is correct and you agree with the proposed solutions, I'm happy to provide a PR if that reduces the workload for you.

[...]

PR #9661, opened 2021-09-21
code-block: Fix handling of :dedent: and add unit tests

ghost:
Subject: code-block: Fix handling of :dedent: and add unit tests

### Feature or Bugfix
Depending on the point of view this is one or a combination of:
- Feature
- Bugfix
- Refactoring

### Purpose
- an argument of ``0`` to dedent is handled as numeric argument and not as "no argument".
- Preserve newlines in the literal block when :dedent: without argument is given.

### Relates
- #9636


doc/usage/restructuredtext/directives.rst
```diff
@@ -598,10 +598,23 @@ __ https://pygments.org/docs/lexers
       are removed via :func:`textwrap.dedent()`.  For example::
 
          .. code-block:: ruby
+            :linenos:
```
tk0miya, line 601: Why did you add this option here?

doc/usage/restructuredtext/directives.rst
```diff
@@ -598,10 +598,23 @@ __ https://pygments.org/docs/lexers
...
             :dedent: 4
 
                 some ruby code
 
+      Using ``0`` as argument does not strip anything and can be used to
```
tk0miya: As I commented in the issue, I can agree on this feature as a "secret" feature. So this should not be mentioned in the document. To add a proper feature, we have to consider another way to keep indentation because it's common problem of reST directives, not only this directive.

tk0miya (changes requested):

doc/usage/restructuredtext/directives.rst
ghost: Because there is no difference in

```rst
.. code-block:: ruby

       some ruby code
```

and

```rst
.. code-block:: ruby
   :dedent: 4

       some ruby code
```

due to how ReST handles whitespace. This confused me the first time I read the docs and played with the directive. Adding another option helps to "fixate" the ReST indent (as we have already discussed), thus adding ``:dedent: 4`` now makes a difference.  Maybe it might be worth pointing out this caveat out somewhere in the documentation (it mostly affects code as whitespace is usually not significant in prose)?

doc/usage/restructuredtext/directives.rst
ghost: Ok, I'll remove this section.

doc/usage/restructuredtext/directives.rst
tk0miya: Hmm? This example contains both `:linenos:` and `:dedent: 4` option. I don't still understand why you added the former one.

>(it mostly affects code as whitespace is usually not significant in prose)

I'm not sure because there are many kinds of directives (including 3rd parties) in the world. So there might be some directives that give a special meaning to indentation. IMO, this is common problem (or tip) for the reST directive, not only for `code-block`.

ghost: > Hmm? This example contains both :linenos: and :dedent: 4 option. I don't still understand why you added the former one.

For the reason explained above:

The output of the following two is the same:

```rst
.. code-block:: ruby
   :dedent: 4

       some ruby code
```

```rst
.. code-block:: ruby

       some ruby code
```

so a person might come to the conclusion that ``:dedent:`` has seemingly no effect.

Whereas

```rst
.. code-block:: ruby
   :linenos:
   :dedent: 4

       some ruby code
```

```rst
.. code-block:: ruby
   :linenos:

       some ruby code
```

produce different outputs.
Thus, demonstrating the purpose of ``:dedent:`` much better in my opinion.

More crucial, from a didactic point of view, the current example mixes the effect of ``:dedent:`` with a not so obvious problem/quirk/behaviour of ReST whitespace parsing making it harder to understand what is really going on.  Additionally, from an algorithm point of view the first two directives (the ones without :linenos:) see different contents: One sees ``"some ruby code"`` whereas the other sees ``"    no ruby code"`` (not the spaces [which are partially butchered by GitHub]).
Do you now understand why I think that this is a better example and added another option?

Remark: The name of the additional option is irrelevant, I chose ``:lineos:`` just because it was the first option under ``.. code-block::``, we can also use any other option: ``:class:``, ``:name:``, ``:caption:``, ...


> IMO, this is common problem (or tip) for the reST directive, not only for code-block.

It is and shouldn't it be therefore mentioned more prominently?
It does not seem to be documented anywhere except buried deep in some specification.
At least I found nothing in:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html
https://www.sphinx-doc.org/en/master/tutorial/index.html (and its subpages)

I'm happy to provide a PR with a proposal.  Since you say its a common problem with all directives, I would place it in https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives ?

tk0miya: Thank you for your explanation. I understood what you'd like to do.

In my thought, this paragraph is a description of the `:dedent:` option. So it should describe how it works. But it seems you're going to describe what is happened if the option not given, right?

Indeed, the examples you shown are converted to the same result. But, the meaning and its internal processing are different at all. So they should be separated.

>It is and shouldn't it be therefore mentioned more prominently?
>It does not seem to be documented anywhere except buried deep in some specification.

IMO, the best place is the spec of reST. Our document does not describe the whole of the syntax. So it's best.
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#directives

Another idea is the "Directive" section of the "reST Primer" document. It describes the indentation of the directives as following. It would be better to make it details.

>The directive content follows after a blank line and is indented relative to the directive start.

ghost: > Thank you for your explanation. I understood what you'd like to do.
>
> In my thought, this paragraph is a description of the :dedent: option. So it should describe how it works. But it seems you're going to describe what is happened if the option not given, right?

No, I'm describing what happens if the option is given. Isn't describing the effect of an option if it is present not part of describing how it works?  In other words describing the effect what happens when the option is added? That's what my example does, you see an effect and thus see how the option works. The current example has no effect when the option ``:dedent:`` is added.

> Indeed, the examples you shown are converted to the same result. But, the meaning and its internal processing are different at all. So they should be separated.

Sorry, don't quiet understand what you are trying to say.

From my point of view the current example (see below) does different internal processing:

```rst
.. code-block:: ruby
   :dedent: 4

       some ruby code
```
```rst
.. code-block:: ruby

       some ruby code
```

Namely, they change the meaning of the spaces before ``some ruby code``. In the first it is two indent steps, in the second its only one. This blurries the effect what the option ``:dedent:`` has on the output, thus concealing how it works.

However, in my proposed example:

```rst
.. code-block:: ruby
   :linenos:
   :dedent: 4

       some ruby code
```
```rst
.. code-block:: ruby
   :linenos:

       some ruby code
```

the interpretation of the indent and thus the content of the directive stays the same.
Thus showing how the ``:dedent:`` option affects the content.

To prevent potential misunderstandings, I'm not comparing examples which differ in the presence of the option ``:lineos:``.


Please see this example from a (new) Sphinx *user's* point of view, not from a Sphinx developer's. The user is concerned about the output they get. If you compare the ``:dedent:`` example with all the other examples for the options of ``.. code-block::``, they all show their effect, i.e. produce different output compared to when they are not present. The ``:dedent:`` example is the only one which does not, because its effect is overshadowed by a different interpretation/parsing behaviour of the indent.


---

I think maybe we don't quite get what the other's point of view is or how they think an example should be. To prevent running in circles and to move forward in this discussion, I suggest

*  either the maintainer (I guess that would be you) makes a decision and gives a thumbs up or down for this commit
*  or we expand the audience and ping some other people, maybe they see things we are currently not.

ghost: > Another idea is the "Directive" section of the "reST Primer" document. It describes the indentation of the directives as following. It would be better to make it details.
>
>>  The directive content follows after a blank line and is indented relative to the directive start.

I've provided a proposal, see #9679.

tk0miya (approved):

Issue #3097, opened 2016-10-25
URL-decodable paths don't seem to work


mgeier:
This is not something I would do in a "normal" Sphinx project, but it came up while using auto-generated files as input to Sphinx. See https://github.com/spatialaudio/nbsphinx/issues/67.

When using a sequence like `%AB` (where `A` and `B` are any hexadecimal digits) in a source file or directory name, the link to the resulting page in the HTML output will not work.

For example, try using a source file named `%3f.rst`. Sphinx will generate a link to `%3f.html`, which the browser will interpret as URL-encoded string and will try to open a file named `?.html` (which doesn't exist).

In order for this to work, Sphinx would have to URL-encode the link to `%253f.html`. The generated file is correctly named `%3f.html`, it's just the link that's broken.

When using LaTeX output, this is not a problem. However, when using an URL-decodable file name for the `image` directive, it breaks in both HTML and LaTeX mode. Here again, the generated image file name is correct, but the image is not displayed. There may be other directives which don't work with such a file name.

Is this a bug or a feature?

For an image URL it would make sense to expect an URL-encoded string, but what about a local image?
And what about a local source file in a `toctree`?

tk0miya:
IMO, this is not an intended behavior. So it's a bug.
But I believe this is not a standard usecase. so I marked this as low priority.

PR #9670, opened 2021-09-24
Fix download file with special characters

jonppe:
Subject: Fix encoding special characters in download filenames

### Purpose
- Allow user to link files with special characters in filename

### Relates
https://github.com/sphinx-doc/sphinx/issues/3097

### DIFF
--- a/tests/roots/test-root/includes.txt
+++ b/tests/roots/test-root/includes.txt
@@ -3,6 +3,7 @@ Testing downloadable files
 
 Download :download:`img.png` here.
 Download :download:`this <subdir/img.png>` there.
+Download :download:`file with special characters <file_with_special_#_chars.xyz>`.
 
 Test file and literal inclusion
 ===============================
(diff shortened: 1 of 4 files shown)


tests/roots/test-root/includes.txt
```diff
@@ -3,6 +3,7 @@ Testing downloadable files
 
 Download :download:`img.png` here.
 Download :download:`this <subdir/img.png>` there.
+Download :download:`file with special characters <file_with_special_#?_chars.xyz>`.
```
tk0miya: Sadly, this filename was invalid on Windows platform. Could you rename it please?

tk0miya (changes requested):
LGTM with nits.

tests/roots/test-root/includes.txt
jonppe: I dropped the question mark. Hopefully works now.

tk0miya (approved):

Issue #9663, opened 2021-09-22
Provide better CSS selectors for API-documentation content


pradyunsg:
**Is your feature request related to a problem? Please describe.**

It is fairly difficult to stylise the content produced by autodoc, due to lack of selectors for certain elements.

- All the themes for Sphinx, that stylise autodoc generated content, need to have weird selectors to ensure that autodoc content is stylised correctly. There is no good way to select an autodoc-generated definition list.

  https://github.com/pradyunsg/furo/blob/677d421c54c79845f84c54ee68394dfb4befeafb/src/furo/assets/styles/content/_api.sass#L6
  https://github.com/readthedocs/sphinx_rtd_theme/blob/ddbe6c69ea1c328fd670cd9fd599a235b4a35c80/src/sass/_theme_rst.sass#L427

- There is no way to select only the parameter name in the generated signature or in the generated parameter list.

- There is no way to select only the parameter type annotations in the generated signature or in the generated parameter list.

- There is no way to consistently select for keywords like `class` while also not selecting the named attributes (both use `.property`).
  - https://tox.readthedocs.io/en/rewrite/plugins_api.html#tox.execute.request.StdinSource.OFF
  - https://tox.readthedocs.io/en/rewrite/plugins_api.html#tox.execute.request.StdinSource.user_only

**Describe the solution you'd like**

Provide selectors for these cases, so that this content can be stylised to look good. :)

**Describe alternatives you've considered**

Not providing these selectors, and maintaining status quo? That's suboptiomal but certainly an option. It is also possible for downstream themes to rewrite whatever output autodoc generates, 

**Additional context**

pdoc3 has IMO very nice signature presentation for API documentation. Presentation like that is not possible with Sphinx, due to the lack of such selectors.

https://pdoc3.github.io/pdoc/doc/pdoc/html_helpers.html

jakobandersen:
Basically I agree. Part of it may be lack of documentation of some of the more recent changes to make CSS styling easier. After #9672 and #9695 has been merged I'll try to get started on the documentation update.
As the content is generated from many different places in Sphinx (each domain, field list handling, autodoc?, etc.) is may be useful at some point if you can elaborate/enumerate what "these cases" are.

pradyunsg:
> "these cases" are.

These cases refers to the bullets listed above, which are the specific bits of content that can't be selected through the CSS to stylize.

pradyunsg:
[...]

Issue #9599, opened 2021-09-01
Add an option to not create links within a `sig-param` block


mattip:
### Describe the bug

Over at Pytorch/vision, we moved from Sphinx 2.x to Sphinx 3.5.4. When rendering python functions with annotations, the rendering looks like this (the raw html is in the detail, broken up by lines).

<dt id="torchvision.datasets.Caltech101">
[...]

PR #9672, opened 2021-09-25
Py, use more desc_sig_* nodes

jakobandersen:
### Feature or Bugfix
- Feature
- Bugfix?
- Refactoring

### Purpose
A continuation of #9023, changing many parts of the Python domain to use more of the ``addnodes.desc_sig_*`` nodes for markup. This especially includes the use of the space node to make sure spaces are not swallowed if themes use alternative CSS (#9599).

### Detail
All of the changes are relatively simple, but it is not complete: the whole ``pendinx_xref`` handling is a bit more complicated than I dare to touch right now.

### TODO and Context
- Any semi-trivial places I missed?
- Probably for another PR: the ``pending_xref`' handling.
  It seem like a name ``A.B.C`` becomes a single xref with the content being a ``nodes.Tex``.
  I think it would be good to change it to something like a 
  1. ``pending_xref`` with target ``A`` and content ``desc_sig_name`` with ``A``
  2. ``.`` as punctuation
  3. ``pending_xref`` with target ``A.B`` and content ``desc_sig_name`` with ``B``
  4. ``.`` as punctuation
  5. ``pending_xref`` with target ``A.B.C`` and content ``desc_sig_name`` with ``C``
- For another PR: write some documentation for how to style domain elements in HTML
- For another PR: work around the ``<span class="pre">`` wrapping in
  https://github.com/sphinx-doc/sphinx/blob/699d03f597130bc39ffa95c6e2cbcf006c6385f6/sphinx/writers/html.py#L709 and https://github.com/sphinx-doc/sphinx/blob/699d03f597130bc39ffa95c6e2cbcf006c6385f6/sphinx/writers/html5.py#L645

### Relates
- Continuation of #9023
- Fixes some parts of #9599 (ping @mattip)
- Perhaps fixes some of #9663? (ping @pradyunsg)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,6 +22,7 @@ Features added
 * #9691: C, added new info-field ``retval``
   for :rst:dir:`c:function` and :rst:dir:`c:macro`.
 * C++, added new info-field ``retval`` for :rst:dir:`cpp:function`.
+* #9672: More CSS classes on Python domain descriptions
 
 Bugs fixed
 ----------
--- a/CHANGES
+++ b/CHANGES
@@ -12,7 +12,9 @@ Incompatible changes
 * #9672: The rendering of Python domain declarations is implemented
   with more docutils nodes to allow better CSS styling.
   It may break existing styling.
-
+* #9672: the signature of
+  :py:meth:`domains.py.PyObject.def get_signature_prefix` has changed to
+  return a list of nodes instead of a plain string.
 
 Deprecated
 ----------
--- a/CHANGES
+++ b/CHANGES
@@ -13,7 +13,7 @@ Incompatible changes
   with more docutils nodes to allow better CSS styling.
   It may break existing styling.
 * #9672: the signature of
-  :py:meth:`domains.py.PyObject.def get_signature_prefix` has changed to
+  :py:meth:`domains.py.PyObject.get_signature_prefix` has changed to
   return a list of nodes instead of a plain string.
 
 Deprecated
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,10 @@ Incompatible changes
 
 * #9649: ``searchindex.js``: the embedded data has changed format to allow
   objects with the same name in different domains.
+* #9672: The rendering of Python domain declarations is implemented
+  with more docutils nodes to allow better CSS styling.
+  It may break existing styling.
+
 
 Deprecated
 ----------
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -850,7 +850,11 @@ class PyAttribute(PyObject):
 
         value = self.options.get('value')
         if value:
-            signode += addnodes.desc_annotation(value, ' = ' + value)
+            signode += addnodes.desc_annotation(value, '',
+                                                addnodes.desc_sig_space(),
+                                                addnodes.desc_sig_punctuation('', '='),
+                                                addnodes.desc_sig_space(),
+                                                nodes.Text(value))
 
         return fullname, prefix
 
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -903,15 +903,18 @@ class PyProperty(PyObject):
 
         return fullname, prefix
 
-    def get_signature_prefix(self, sig: str) -> str:
-        prefix = []
+    def get_signature_prefix(self, sig: str) -> List[nodes.Node]:
+        prefix: List[nodes.Node] = []
         if 'abstractmethod' in self.options:
-            prefix.append('abstract')
+            prefix.append(nodes.Text('abstract'))
+            prefix.append(addnodes.desc_sig_space())
         if 'classmethod' in self.options:
-            prefix.append('class')
+            prefix.append(nodes.Text('class'))
+            prefix.append(addnodes.desc_sig_space())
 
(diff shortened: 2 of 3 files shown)


jakobandersen:
While the use of nodes is undocumented this may still break some themes, so perhaps there should be an "incompatible changes" entry as well?

sphinx/domains/python.py
```diff
@@ -118,11 +118,23 @@ def unparse(node: ast.AST) -> List[Node]:
...
         elif isinstance(node, ast.Constant):  # type: ignore
             if node.value is Ellipsis:
                 return [addnodes.desc_sig_punctuation('', "...")]
+            elif isinstance(node.value, bool):
+                return [addnodes.desc_sig_keyword('', repr(node.value))]
+            elif isinstance(node.value, int):
+                return [addnodes.desc_sig_literal_number('', repr(node.value))]
+            elif isinstance(node.value, str):
+                return [addnodes.desc_sig_literal_string('', repr(node.value))]
+            elif node.value is None:
```
tk0miya: Is this branch needed? It seems nothing changed from the else block.

sphinx/domains/python.py
```diff
@@ -418,11 +434,11 @@ class PyObject(ObjectDescription[Tuple[str, str]]):
 
     allow_nesting = False
 
-    def get_signature_prefix(self, sig: str) -> str:
+    def get_signature_prefix(self, sig: str) -> List[nodes.Node]:
```
tk0miya, line 435: Good change. But it causes incompatible change. So this API change should be merged into the master branch.

tests/test_domain_py.py
```diff
@@ -376,7 +387,7 @@ def test_pyfunction_signature(app):
     assert_node(doctree[1][0][1],
                 [desc_parameterlist, desc_parameter, ([desc_sig_name, "name"],
                                                       [desc_sig_punctuation, ":"],
-                                                      " ",
+                                                      [desc_sig_space],
```
tk0miya: Brackets are not needed here. They're useful to match nested nodes.

tk0miya (commented):
LGTM with nits.

>Any semi-trivial places I missed?

It seems OK. Let's fix them if we'll find them later.

>While the use of nodes is undocumented this may still break some themes, so perhaps there should be an "incompatible changes" entry as well?

I think side effects are nothing or very less. But I'm not sure nobody gives styles to these classes. So it would be better to mention it.

sphinx/domains/python.py
jakobandersen: Technically not, it was mostly for completeness such that the else-case shouldn't trigger. I have removed the branch and updated the comment.

sphinx/domains/python.py
jakobandersen: Is this really part of the public API?

tests/test_domain_py.py
jakobandersen: Fixed.

sphinx/domains/python.py
tk0miya: No, it's a private one, I believe. But, as you know, we've seen many extensions and hacks that depend on private APIs. So this change might cause trouble. So it would be better not to change the interface as possible during the same major version (or let's declare it as an incompatible change).

jakobandersen: Right ([Hyrum's Law](https://www.hyrumslaw.com/) :-(). Even if this method is not changed, a new can't be introduced early either, as the new call in ``run()`` would be incompatible for anyone who subclassed ``PyObject``.
So, as far as I can see:
- Option 1: I can split the PR so some can be merged in 4.x, and the incompatible part in master.
- Option 2: Merge it all now to 4.x with an extra incompatibility note for anyone who messed with these private APIs.
- Option 3: Make a parallel version and hax ``run()`` to detect if the dynamic type of ``self`` is one of the normal Python objects and if not check whether the subclass overwrites the new method, and use that to call the right version. (This seems quite brittle and a lot of work to shield those that messed with private APIs.)

tk0miya: Yes. My first proposal is option 1. But I admit it's a very cowardly approach. So I can agree with option 2 (Note: We have to mention it as incompatibility change in CHANGES). Both are okay to me.

jakobandersen: I have added an incompatibility note (option 2).

jakobandersen: Well, I wanted to add the note. _Now_ it should be pushed.

CHANGES
```diff
@@ -9,6 +9,12 @@ Incompatible changes
 
 * #9649: ``searchindex.js``: the embedded data has changed format to allow
   objects with the same name in different domains.
+* #9672: The rendering of Python domain declarations is implemented
+  with more docutils nodes to allow better CSS styling.
+  It may break existing styling.
+* #9672: the signature of
+  :py:meth:`domains.py.PyObject.def get_signature_prefix` has changed to
```
tk0miya: typo?
```suggestion
  :py:meth:`domains.py.PyObject.get_signature_prefix` has changed to
```

tk0miya (commented):
LGTM with nits. Please merge this after fixing the typo.

tk0miya (approved):

PR #9675, opened 2021-09-25
Proof of Concept: Only create a target for object descriptions

ghost:
Subject: Only create a target for object descriptions

### Feature or Bugfix
- Feature

### Purpose
Adding feature for #9662 (ability to only create a target for a domain description).

Thanks to @jakobandersen for the inspiration.

### Detail
- This is a proof of concept for implementing the replacing in the run() method.
- #9671 is an alternative implementation using PostTransform
- The difference between the two is that targets in the PostTransform are not merged with the following element/node, while with the run() method they are.
- As a consequence the targets for headings are in the section of the heading and not in the previous section.

- Does only cover domain directives deriving from ObjectDescription other directives are (not yet) covered, e.g. ``.. py:module::``.  (However ``.. py:module`` does produce only a target and no content anyway, so is not a good example here.)
- This is still work in progress and needs proper documentation and testing.

### Relates
- #9662
- #9671

### DIFF
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -60,6 +60,7 @@ class ObjectDescription(SphinxDirective, Generic[T]):
     final_argument_whitespace = True
     option_spec: OptionSpec = {
         'noindex': directives.flag,
+        'hidden': directives.flag,
     }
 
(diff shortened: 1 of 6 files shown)


AA-Turner:
Closing explicit PoC PR as there's been no discussion in six months.

A

PR #9685, opened 2021-09-29
Set refdoc to be used by missing-reference, intersphinx

### LINKED ISSUE #7634: Fixing broken references on multi-repository site build
I am trying to fix some issues with broken links when building this site: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/index.html

The build steps are given here: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/doc_build.html

It basically consists on cloning many different open-source repositories, where the documentation is distributed among them. The details are not relevant, and are handled automatically by tooling. Since those projects consist primarily of C (or C++) source code and doxygen documentation, `breathe` is used to build from it, and `intersphinx` is used to resolve references between different projects, which seems to be the source of the issue we're having.

One example of a broken link in a document can be seen here: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/include/bluetooth/gatt_dm.html#_CPPv4N15bt_gatt_dm_attr4uuidE. If trying to read the documentation for `bt_uuid` it goes to a broken link. At the same time links that were not generated from source code , eg: https://developer.nordicsemi.com/nRF_Connect_SDK/doc/latest/nrf/samples/nrf9160/lwm2m_carrier/README.html, going to "From Zephyr" and clicking on "MQTT" goes to the correct site.

I did some digging and found a few things. We currently use relative paths in our `intersphinx_mapping`, and one obvious solution would be to use a URL and everything would work, with a few new issues, like documentation not working locally, versioned documentation requiring extra handling on builds, etc.

Trying to find the reason for why some relative path based links work, and some don't, I found out that `intersphinx` resolves links on its `missing_reference` callback, for relative paths it happens here: https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/ext/intersphinx.py#L312. It uses the `refdoc` value (a document path) to deduce how many directories it has to move up before joining with the path given in `intersphinx_mapping`. This works correctly for documentation generated from .rst files, eg `:ref:'MQTT <zephyr:mqtt_socket_interface>'`, the working example I linked above. For the broken links, which are originated from source code parsed by the cpp domain parser, `refdoc` is not available, which prevents `intersphinx` from correctly resolving the paths.

To fix the issue I made something that feels like a terrible hack, but seems to work correctly here. I  changed the `transformation` code where references are resolved with the following diff:

```diff
--- a/sphinx/transforms/post_transforms/__init__.py
+++ b/sphinx/transforms/post_transforms/__init__.py
@@ -74,7 +74,9 @@ class ReferencesResolver(SphinxPostTransform):
 
             typ = node['reftype']
             target = node['reftarget']
-            refdoc = node.get('refdoc', self.env.docname)
+            if not 'refdoc' in node:
+                node['refdoc'] = self.env.docname
+            refdoc = node['refdoc']
             domain = None
 
             try:
```

Although this seems wrong to do (and modifying a parameter!), adding the `refdoc` there fixes the links.

Since I am not very acquainted with the sphinx projects' source code just yet, it would be nice to hear opinions on how to properly fix the issue.

MatthijsBurgh:
### Feature or Bugfix
Bugfix

### Purpose
Intersphinx checks for `refdoc` attribute in node during the `missing-reference` event. To determine to correct relative reference uri.
https://github.com/sphinx-doc/sphinx/blob/cf38356a088acc0052d6cbb50f04c3773ab690c4/sphinx/ext/intersphinx.py#L330-L332

It is called from `post_transforms`, all other options that are tried/checked before, do get a separate `refdoc` as an argument. Though the `missing-reference` event only gets it if it was already in the node, as it doesn't have a separate `refdoc` argument.
https://github.com/sphinx-doc/sphinx/blob/cf38356a088acc0052d6cbb50f04c3773ab690c4/sphinx/transforms/post_transforms/__init__.py#L98-L100

I have not experienced a situation yet, where it was already set.

By not providing the `missing-reference` event with the back-up 'refdoc' from https://github.com/sphinx-doc/sphinx/blob/cf38356a088acc0052d6cbb50f04c3773ab690c4/sphinx/transforms/post_transforms/__init__.py#L81

Relative reference uri's don't work in intersphinx.

So either we need to set the `refdoc` attribute to node or change the API of `missing-reference`, so that it accepts `refdoc` as a separate argument.  I choose the first option, but I am open to implement the second one if you prefer that one.

P.S.: also the viewcode extension expects the `refdoc` attribute of the `node` to be set. So that would favor the first option, which I implemented.
https://github.com/sphinx-doc/sphinx/blob/cf38356a088acc0052d6cbb50f04c3773ab690c4/sphinx/ext/viewcode.py#L204-L205

### Detail
By not having the `refdoc` in intersphinx, it doesn't know how deep you are in your package, therefore it can't determine relative paths to references. It will use assume your are in the root of your package, which will generate broken links as the depth mismatches.

### Relates
Fixes #7634

### DIFF
--- a/sphinx/transforms/post_transforms/__init__.py
+++ b/sphinx/transforms/post_transforms/__init__.py
@@ -78,7 +78,8 @@ class ReferencesResolver(SphinxPostTransform):
 
             typ = node['reftype']
             target = node['reftarget']
-            refdoc = node.get('refdoc', self.env.docname)
+            node.setdefault('refdoc', self.env.docname)
+            refdoc = node.get('refdoc')
             domain = None
 
             try:


jakobandersen:
I can't immediately judge if this is the right solution, but I think it is very similar to what was suggested in #7634 and thereby also fix that.

MatthijsBurgh:
Yes, I have read the issue. This looks like the same problem, but different solution.

I think we both agree that the `refdoc` attribute should be set. I think this is the correct tile, as the back-up is passed to the other functions that try to resolve the reference, but not the `missing-reference` event.

So we should only decide where in that function we added it as an attribute or pass it as an extra argument.

jakobandersen:
> So we should only decide where in that function we added it as an attribute or pass it as an extra argument.

While the ideal thing perhaps would be to add it as an extra argument for the event, that would be a breaking change. The attribute already exists so I think your suggested solution is the right one.

tk0miya:
As I commented at https://github.com/sphinx-doc/sphinx/issues/7634#issuecomment-626113427, this is a bug of the code that generates the pending_xref node. But, as you said, I understand this code automatically recovers the missing `refdoc` attribute bug. So I can agree with your proposal.

off topic: I don't understand why the `refdoc` is needed. If my understanding is correct, it always equals to the current docname (ex. `env.docname`). So it's no reason to store it to the node...

sphinx/transforms/post_transforms/__init__.py
```diff
@@ -95,6 +95,7 @@ def run(self, **kwargs: Any) -> None:
                     newnode = self.resolve_anyref(refdoc, node, contnode)
                 # no new node found? try the missing-reference event
                 if newnode is None:
+                    node['refdoc'] = refdoc
```
tk0miya: I feel it's better to fill this field before calling `domain.resolve_xref()`. So I prefer to call `node.setdefault('refdoc', self.env.docname)` at line 81. What you think?

MatthijsBurgh: That is also totally fine by me.

MatthijsBurgh:
> As I commented at [#7634 (comment)](https://github.com/sphinx-doc/sphinx/issues/7634#issuecomment-626113427), this is a bug of the code that generates the pending_xref node. But, as you said, I understand this code automatically recovers the missing `refdoc` attribute bug. So I can agree with your proposal.

I semi agree with you. As it should be set on generation of the nodes, then we could fix it there. But what code does generate the nodes?
On the other side, as we assign a backup value here to be used by the other resolve function, then also provide it to the intersphinx event.
 
> off topic: I don't understand why the `refdoc` is needed. If my understanding is correct, it always equals to the current docname (ex. `env.docname`). So it's no reason to store it to the node...

Why it is needed, is because intersphinx uses it to determine the relative path of the node to the root of the package, so that it can resolve relative paths to references correctly.

tk0miya:
>But what code does generate the nodes?

I'm not sure. We need to investigate if needed.

> Why it is needed, is because intersphinx uses it to determine the relative path of the node to the root of the package, so that it can resolve relative paths to references correctly.

AFAIK, it is always the same as `env.docname`. If my understanding is correct, it's okay to use `env.docname` instead of `node['refdoc']` in intersphinx.

MatthijsBurgh:
The use of `refdoc` was chosen in 2013 by @birkenfeld. Maybe he knows why `refdoc` was chosen over using `env.docname`.

Otherwise I suggest we just go for `refdoc` solution for now. As `refdoc` is provided to the other resolve functions. And then we can investigate what code generates the nodes and we can add a good solution there.

birkenfeld:
sadly, no idea anymore.

MatthijsBurgh:
I have discovered two places where a `pending_xref` node is created.
1. `roles.py` https://github.com/sphinx-doc/sphinx/blob/c922189920ec36fe34aab41ee467a854849a57b8/sphinx/roles.py#L128-L133
`self.nodeclass` is set as classattribute https://github.com/sphinx-doc/sphinx/blob/c922189920ec36fe34aab41ee467a854849a57b8/sphinx/roles.py#L68
So here the refdoc is correctly set. Sphinx ended up here when parsing `autoapisummary` (https://github.com/readthedocs/sphinx-autoapi)
2. `util/docfields.py` https://github.com/sphinx-doc/sphinx/blob/c922189920ec36fe34aab41ee467a854849a57b8/sphinx/util/docfields.py#L84-L85
Here the `refdoc` attribute is not set. I have experienced that the nodes created here are the nodes causing trouble.
This code is reached from the `make_xref` function of `PyXrefMixin` in [`domains/python.py`](https://github.com/sphinx-doc/sphinx/blob/c922189920ec36fe34aab41ee467a854849a57b8/sphinx/domains/python.py#L312)

So the second option is the cause of the problem. I think we can set the `refdoc` in the `make_xref` function of the `Field` class in `docfields.py`. Other Field classes inherit from that class, so that provides a solution for all of them.

jakobandersen:
Unfortunately it's not only in those two places:
```
[]/sphinx/sphinx$ grep -Rn pending_xref\(
util/docfields.py:68:        refnode = addnodes.pending_xref('', refdomain=domain, refexplicit=False,
domains/python.py:105:    return pending_xref('', *contnodes,
domains/std.py:466:        refnode = pending_xref(title, reftype='token', refdomain='std',
domains/std.py:804:            contnode = pending_xref('')
domains/citation.py:136:            ref = pending_xref(target, refdomain='citation', reftype='ref',
domains/c.py:140:            pnode = addnodes.pending_xref('', refdomain='c',
domains/cpp.py:613:            pnode = addnodes.pending_xref('', refdomain='cpp',
domains/cpp.py:635:            pnode = addnodes.pending_xref('', refdomain='cpp',
domains/cpp.py:1641:            pnode = addnodes.pending_xref('', refdomain='cpp',
# and other hits for the actual class and the writer methods
```
It seems that the ``refdoc`` attribute is basically not optional anymore. How many of the other attributes are actually optional?
Would it be better to make a breaking change for ``pending_xref.__init__`` so we are forced to provide all the things that are mandatory? Of course that should only be done in master, but then the change in this PR could be used as a partial fix in 4.x.

MatthijsBurgh:
I am just a user of Sphinx. As Sphinx is pretty big, I don't have the complete overview how everything is/should be working. Therefore I can't say which attributes are/should be required and which are/should be optional.

Also it is undocumented what the meaning of all attributes are(So it is unknown which attributes there are). So the best solution would be to gather all the attributes for all nodes. Document their value, purpose and which node requires which attribute. And combine this with making the attributes required in the code as well.

I would prefer merging this PR as a partial fix. I don't mind my solution or your solution (https://github.com/sphinx-doc/sphinx/pull/9685#discussion_r720862400)

tk0miya:
>I would prefer merging this PR as a partial fix. I don't mind my solution or your solution (#9685 (comment))

As I commented on it, I prefer to fill the `refdoc` on the top of the handler. I will merge this if fixed.

Additionally, as I commented as "off topic", the `refdoc` attribute should be deprecated in the near future. But it does not prevent merging this.

MatthijsBurgh:
@tk0miya I have updated the PR to match your suggestion. I also changed the target branch to 4.x, as you added this PR to the 4.3.0 milestone.

MatthijsBurgh:
@tk0miya anything that is blocking this PR to be approved/merged?

tk0miya:
Sorry for not responding long. I'm back now. It seems good to me. I'll merge this after CI passed.

tk0miya:

PR #9688, opened 2021-09-30
Fix bug of Sphinx's .. code:: directive not recognizing :class: option

ghost:
Subject: Fix bug of Sphinx's .. code:: directive not recognizing :class: option

### Feature or Bugfix
- Bugfix

### Purpose

Sphinx's own .. code:: directive (but not docutils') does not recognise
the :class: option but only the :classes: option.  This is probably due
to an oversight that the user-visible option is called ``:class:`` while
the Python attribute is called ``classes`` (to not collide with the
keyword ``class``).

Fix it by checking for the option ``class`` instead of ``classes``.

Remark: Theoretically, someone could depend on ``classes`` being
recognised by Sphinx's ``.. code::`` directive.  However, I highly
doubt that because:

a) ``class`` is the canonical and documented way to add classes
b) ``classes`` was never documented by Sphinx.
c) Sphinx encourages ``.. code-block::`` over ``.. code::``.


astrojuanlu:
Thanks for the PR @latosha-maltba ! Looks like it has conflicts though. Perhaps you branched off `master`, but this is being merged in `4.x` instead?

ghost:
Hi, @astrojuanlu

> Thanks for the PR @latosha-maltba ! Looks like it has conflicts though. Perhaps you branched off master, but this is being merged in 4.x instead?

Not quite, there was just some other pull request merged and that lead to a conflict in CHANGES as is to expect when documenting the changelog in a monolithic file.

tk0miya (changes requested):
Could you call `docutils.parsers.rst.roles.set_classes()` before accessing `self.options['classes']`? It's a well-known technique to implement directives on docutils. This is a bug that the code directive does not call the helper function before accessing the option. You can see examples under the docutils.parsers.rst.directives package.

CHANGES
```diff
@@ -41,6 +41,7 @@ Bugs fixed
 * #9649: HTML search: when objects have the same name but in different domains,
   return all of them as result instead of just one.
 * #9678: linkcheck: file extension was shown twice in warnings
+* Fix bug ``.. code::`` not recognising ``:classes:`` option instead of ``:class:`` option.  The new behaviour is only to accept ``:class:``.
```
tk0miya: This is not correct. The `code` directive does not recognize the `:classes:` option. Simply, it ignores given `:class:` option.

ghost:
Been there, done that.

tk0miya (approved):

tk0miya:
Merged! Thank you for your work :-)

PR #9691, opened 2021-10-01
C, add support for return value list

gmarull:
Subject: C, add support for return value list

### Feature or Bugfix

- Feature

### Purpose

- Improve C domain capabilities

### Detail

- Add support for a new doc field type that allows to display a list of return values.

### Relates

https://github.com/michaeljones/breathe/pull/749


jakobandersen:
Good idea, thanks!

PR #9736, opened 2021-10-14
Fix rendering of nested type annotations with alias defined

JanLuca:
Subject: Fix rendering of nested type annotations with alias defined

### Feature or Bugfix
- Bugfix

### Purpose
- Make autodoc usable with nested types with alias defined

### Detail
The current implementation replaces type annotations with a alias defined by the class TypeAliasForwardRef. This class is later again replaced by the type defined in the alias map.

This commit fixes the problem that this second replacement does not happen if the aliased type is nested in another generic type (e.g. Iterator[aliased_type]). The new function goes recursivly through the nested types and replaces the internal class TypeAliasForwardRef again by the correct annotation.

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -614,6 +614,17 @@ def _should_unwrap(subject: Callable) -> bool:
     return False
 
 
+def _fix_type_alias_forward_ref(annotation: Any) -> Any:
+    if isinstance(annotation, TypeAliasForwardRef):
+        return annotation.name
+
+    if hasattr(annotation, "__args__"):
+        new_args = tuple(_fix_type_alias_forward_ref(arg) for arg in annotation.__args__)
+        return annotation.copy_with(new_args)
+
+    return annotation
+
+
 def signature(subject: Callable, bound_method: bool = False, follow_wrapped: bool = None,
               type_aliases: Dict = {}) -> inspect.Signature:
     """Return a Signature object for the given *subject*.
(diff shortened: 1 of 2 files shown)


asmeurer:
I've been getting the error `unhashable type: 'TypeAliasForwardRef'` with nitpicky=True with certain types of nested type hints. This PR makes them go away.

(side question: is it possible to tell autodoc to just completely not try to make type hints into cross-references?)

sphinx/util/inspect.py
```diff
@@ -614,6 +614,17 @@ def _should_unwrap(subject: Callable) -> bool:
     return False
 
 
+def _fix_type_alias_forward_ref(annotation: Any) -> Any:
```
tk0miya, line 617: How about `expand_type_alias_forward_ref()` instead?

Additonally, it would be better to add a docstring to this.

sphinx/util/inspect.py
```diff
@@ -614,6 +614,17 @@ def _should_unwrap(subject: Callable) -> bool:
...
+    if isinstance(annotation, TypeAliasForwardRef):
+        return annotation.name
+
+    if hasattr(annotation, "__args__"):
+        new_args = tuple(_fix_type_alias_forward_ref(arg) for arg in annotation.__args__)
+        return annotation.copy_with(new_args)
```
tk0miya, line 623: Is this method available on python3.6 and 3.10?

tk0miya (changes requested):
Please add a testcase for this implementation.

Issue #8904, opened 2021-02-17
Broken HTML rendering of index entry that contains a semicolon


pmoura:
**Describe the bug**
I'm generating documentation for a Prolog system implementation using Sphinx. The Prolog language includes conjunction and disjunction control constructs that are represented by, respectively, the compound terms `(';')/2`, `(,)/2`, and `!/0`. The index entries for these control constructs must contain this exact text.

But the following index directives don't generate correct entries due to the presence of the comma, semicolon, and exclamation mark:

```rst
.. index:: (;)/2

.. index:: (',')/2

.. index:: !/0
```

For the first two cases, a solution was mentioned in StackOverflow related question (https://stackoverflow.com/questions/66229133/escaping-special-characters-in-index-directives) :

```rst
.. index:: single: (',')/2

.. index:: single: !/0
```

But this solution doesn't work for the `(';')/2`. Instead, I get (using the HTML render):

```html
      <li>
    (

      <ul>
        <li><a href="my_index.html#index-1">)/2</a>
</li>
      </ul></li>
```

Also tried to use a backslash for escaping the semicolon to no avail. As mentioned in the SO answer, the use of `single:` work if the text that follows consists only of the semicolon.

**To Reproduce**
The bug handling text that include a semicolon as an index entry can be reproduced using either:

```rst
.. index:: (;)/2
```

or:

```rst
.. index:: single: (;)/2
```

**Expected behavior**
A well formatted index entry showing as text `(';')/2`.

**Your project**
The bug is found in a private company project.

**Screenshots**
(none)

**Environment info**
- OS: macOS 11.2
- Python version: 3.6.12
- Sphinx version: v3.4.3
- Sphinx extensions:  (none)
- Extra tools: (any web browser)

**Additional context**
StackOverflow related question and answer with further discussion:
https://stackoverflow.com/questions/66229133/escaping-special-characters-in-index-directives

KaKkouo:
It seems to depend on split_into function

sphinx/util/__init__.py ( Sphinx 4.2.0 )

```python
365
366 def split_into(n: int, type: str, value: str) -> List[str]:
367     """Split an index entry into a given number of parts at semicolons."""
368     parts = [x.strip() for x in value.split(';', n - 1)]
369     if sum(1 for part in parts if part) < n:
370         raise ValueError('invalid %s index entry %r' % (type, value))
371     return parts
372
```

sphinx/environment/adapters/indexentries.py ( Sphinx 4.2.0 )

[...]

PR #9749, opened 2021-10-18
#8904: value.split(';', n - 1) -> value.split('; ', n - 1)

KaKkouo:
Subject: <#8904: fix about '(;)/2'>

### Feature or Bugfix

- Bugfix

### Purpose
- correct the behavior of the split_into function.
- Since this is a fix for parsing strings in index directives/role, there is no environment or extension to depend on in html.
  - As for latex, the code looks fine to me, but I'm not sure.

### Detail
about split_into function

source file:

- sphinx/util/__init__.py

analysis:

- this function is used by two files.
  -  sphinx/environment/adapters/indexentries.py
  - sphinx/writers/latex.py, LaTeXTranslator class, visit_index method.
- When writing multiple terms, they are supposed to be separated by a semicolon, but I understand that the semicolon is followed by a space. With this change, we now include the whitespace after the semicolon to separate them.

### Relates
- #8904 
  - I wrote about the work in #8904.
- [Escaping special characters in index directives](https://stackoverflow.com/questions/66229133/escaping-special-characters-in-index-directives/69598134)

### DIFF
--- a/sphinx/util/__init__.py
+++ b/sphinx/util/__init__.py
@@ -365,7 +365,7 @@ def rpartition(s: str, t: str) -> Tuple[str, str]:
 
 def split_into(n: int, type: str, value: str) -> List[str]:
     """Split an index entry into a given number of parts at semicolons."""
-    parts = [x.strip() for x in value.split(';', n - 1)]
+    parts = [x.strip() for x in value.split('; ', n - 1)]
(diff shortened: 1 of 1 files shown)


KaKkouo:
This has a error on circleci. so, I will try that from git clone.

Issue #9761, opened 2021-10-22
Allow 403: Forbidden as acceptable in linkcheck


robinovitch61:
**Is your feature request related to a problem? Please describe.**
I'd like our sphinx build to check links with the `linkcheck` builder. A number of links are behind oauth and throw 403s when requested. I'd like to make this acceptable and not report as a broken link.

The current linkcheck accepts 401s:
https://github.com/sphinx-doc/sphinx/blob/cf96c93cf17c2659f1947201a684bb8e2b6c6250/sphinx/builders/linkcheck.py#L475-L477

[403 is similar](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403):
> The HTTP 403 Forbidden client error status response code indicates that the server understands the request but refuses to authorize it.
>
> This status is similar to 401, but in this case, re-authenticating will make no difference. The access is permanently forbidden and tied to the application logic, such as insufficient rights to a resource.

**Describe the solution you'd like**
Add 403 as an acceptable, not-broken link response code.

**Describe alternatives you've considered**
None

**Additional context**
I've made a PR for this [here](https://github.com/sphinx-doc/sphinx/pull/9762). Thank you!

AA-Turner:
Closing, though the allowable responses config option in the linked PR would be considered as a feature.

A

PR #9762, opened 2021-10-22
Allow 403 response in linkcheck

robinovitch61:
Subject: Allows 403 response in linkcheck

### Feature or Bugfix
- Feature

### Purpose
Described in https://github.com/sphinx-doc/sphinx/issues/9761

### Relates
https://github.com/sphinx-doc/sphinx/issues/9761

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -472,7 +472,7 @@ class HyperlinkAvailabilityCheckWorker(Thread):
                                                 auth=auth_info, **kwargs)
                         response.raise_for_status()
             except HTTPError as err:
-                if err.response.status_code == 401:
+                if err.response.status_code in (401, 403):
                     # We'll take "Unauthorized" as working.
                     return 'working', ' - unauthorized', 0
                 elif err.response.status_code == 429:


sphinx/builders/linkcheck.py
```diff
@@ -472,7 +472,7 @@ def check_uri() -> Tuple[str, str, int]:
                                                 auth=auth_info, **kwargs)
                         response.raise_for_status()
             except HTTPError as err:
-                if err.response.status_code == 401:
+                if err.response.status_code in (401, 403):
```
tk0miya, line 475: I'm not sure all 403 responses can be considered as working. Is it better to control this behavior by configuration?

francoisfreitag: I agree with @tk0miya, getting an answer with 403 in your use case indicates the resource exists on the server, but a 403 may have other causes.

The status code 401 indicates the resource exists, but is protected. The request would have succeeded with valid credentials. It’s acceptable to consider the link working in that case.

From the [RFC 7231, section 6.5.3 (403 Forbidden)](https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.3):

> ### 6.5.4 403 Forbidden
> The 403 (Forbidden) status code indicates that the server understood the request but refuses to authorize it.  A server that wishes to make public why the request has been forbidden can describe that reason in the response payload (if any).
>
> If authentication credentials were provided in the request, the server considers them insufficient to grant access.  The client SHOULD NOT automatically repeat the request with the same credentials.  The client MAY repeat the request with new or different credentials.  However, a request might be forbidden for reasons unrelated to the credentials.
>
> An origin server that wishes to "hide" the current existence of a forbidden target resource MAY instead respond with a status code of 404 (Not Found).

It may indeed be that some form of authentication will actually help, but linkcheck cannot tell that from the server response. Linkcheck cannot assume all servers will accept to answer provided authentication and consider the link valid.

The use case is fairly niche IMO. Servers protecting resources tend to use the 404 status, to avoid leaking the resource existence. One option is to [`linkcheck_ignore`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_ignore) domains that require OAuth (or any other complex auth system). That is not as helpful as verifying an access-related error code is emitted by the web server though.

If we’re to introduce a new configuration option for this use case, I think it should map regexps to a list of acceptable status code. Something like:
```python
linkcheck_ignore_status_codes = {
    r'^https://my-site\.org/': [401, 403, 503],
    r'.*': [401],
}
```
Making the catchall rule the default would preserve the existing behavior. It would also make it possible to treat 401 as an error, if users want to ensure a page is publicly available.

What do you think?

robinovitch61: I don't fully grasp why assuming 403 isn't broken is different than assuming 401 isn't broken, considering the [MDN docs on 403](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403):
> ...This status is similar to 401, but in this case, re-authenticating will make no difference. The access is permanently forbidden and tied to the application logic, such as insufficient rights to a resource.

But I do really like the idea of `linkcheck_ignore_status_codes` with regex -> codes as described above. I'm not sure I'll be able to get to this change anytime this week, but I'll try to make it a priority soon. If anyone else would like to work on it in the meantime, I wouldn't be upset.

Thanks for the feedback and consideration!

francoisfreitag: The difference is that a 403 can be due to reasons unrelated to the credentials. Linkcheck can’t tell whether the link is working or not.
With 401, the server indicated the issue came from credentials.

AA-Turner:
Closing, though the allowable responses config option would be considered as a feature.

A

PR #9769, opened 2021-10-25
Fix linkcheck_auth link to Requests authentication

francoisfreitag:
The link was directing to
https://www.sphinx-doc.org/en/master/usage/requests-auth>. Prefer using
the intersphinx module to generate the link, it’s more robust than
directly linking to the page.

### Feature or Bugfix
- Bugfix


tk0miya:
LGTM

PR #9772, opened 2021-10-25
Closes #9733: Fix for logging handler flushing warnings in the middle of the docs build

### LINKED ISSUE #9733: Sphinx silently drops warnings when libraries like 'airflow' or 'mlflow' are imported in the middle of the build
### Describe the bug

I'm finding that sphinx does not emit warnings that it should be emitting.

For example, there is a typo here: https://github.com/dagster-io/dagster/blob/master/docs/sphinx/sections/api/apidocs/execution.rst#L56

`ReeconstructablePipeline` should be `ReconstructablePipeline` (there is an extra 'e' character), which should produce a Sphinx warning since the class does not exist.

Confusingly, I'm finding that the expected warnings start appearing when I reduce the number of libraries in my environment, even when the libraries that I remove are not specifically required by the sphinx build. I noticed this when I tried to start using autodoc_mock_imports to reduce our imports - suddenly, a bunch of unrelated warnings started appearing. 

It does not seem that including a specific library causes the problem (I verified this by attempting to bisect the problem by reducing libraries one by one) - once I reduce the number of libraries in my virtual environment enough, the expected warnings begin appearing again during the build. 

See specific reproduction instructions below. Thanks for any help you can provide.

### How to Reproduce

(in a fresh virtualenv with pip 19.2.3 installed)
```
git clone https://github.com/dagster-io/dagster.git --branch sphinxbugreport --depth 1
cd dagster/docs
pip install -r docs-requirements.txt
cd sphinx; make json SPHINXOPTS="-W";
```
(sphinx will build successfully with no warnings, despite having the ReeconstructablePipeline typo mentioned above that should generate a warning and thus an error due to the -W flag)

Now, to make the warning appear, follow these steps to remove a bunch of libraries that are not actually needed by the sphinx build (it does not seem that there's one individual library that specifically causes the problem - I attempted to find one by bisecting, but it seems that it's just a matter of removing enough libraries):
```
$ cd ..
$ cat toremove | xargs pip uninstall -y
$ cd sphinx; make json SPHINXOPTS="-W";
```
The sphinx build will now correctly create a warning for the typo:

```
Warning, treated as error:
autodoc: failed to import class 'ReeconstructablePipeline' from module 'dagster.core.definitions.reconstructable'; the following exception was raised:
  File "/Users/dgibson/.pyenv/versions/3.8.3/envs/sphinxreport/lib/python3.8/site-packages/sphinx/util/inspect.py", line 393, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'dagster.core.definitions.reconstructable' has no attribute 'ReeconstructablePipeline'
```

### Expected behavior

Sphinx will always produce warnings for missing classes

### Your project

https://github.com/dagster-io/dagster/tree/sphinxbugreport/docs/sphinx

### Screenshots

_No response_

### OS

Mac

### Python version

3.8.3

### Sphinx version

3.5.2

### Sphinx extensions

sphinx.ext.autodoc, sphinx.ext.ifconfig, sphinx.ext.intersphinx, sphinx.ext.viewcode, nbsphinx, sphinx.ext.autosectionlabel

### Extra tools

_No response_

### Additional context

_No response_

gibsondan:
Subject: Fix for logging handler flushing warnings in the middle of the docs build

### Feature or Bugfix
- Bugfix

### Purpose
My project was mysteriously dropping warnings (see https://github.com/sphinx-doc/sphinx/issues/9733 for detailed repro) and I realized that it's because it imports libraries like airflow and mlflow that set up loggers automatically when they are imported. This causes this handler to flush even though shouldFlush is set to always return False. A simple workaround is to override flush to be a no-op.

### Relates
https://github.com/sphinx-doc/sphinx/issues/9733

### DIFF
--- a/sphinx/util/logging.py
+++ b/sphinx/util/logging.py
@@ -171,6 +171,11 @@ class MemoryHandler(logging.handlers.BufferingHandler):
     def shouldFlush(self, record: logging.LogRecord) -> bool:
         return False  # never flush
 
+    def flush(self) -> None:
+        # suppress any flushes triggered by importing packages that flush
+        # all handlers at initialization time
+        pass
+
     def flushTo(self, logger: logging.Logger) -> None:
         self.acquire()
         try:


tk0miya:

PR #6879, opened 2019-12-01
Fix #6803: Disable parallel build on macOS and py38+

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #6803 and #6865 
- Since python3.8, the default start method of multiprocessing has changed to "spawn" on macOS for security reasons. The implementation for parallel build are strongly coupled with "fork" method. So this disables the feature temporarily (until reimplemented whole of parallel build feature).
- We can switch the start method to "fork" manually. But it is risky choice. So this does not do that.

PR #9793, opened 2021-10-28
Set multiprocessing start method to `fork`

samdoran:
Subject: Set multiprocessing start method to `fork`

### Feature or Bugfix

- Bugfix

### Purpose
Explicitly set the multiprocessing start method to `fork` since the code depends on this behavior.

### Detail
Since Python 3.8, the default start method was changed to `spawn` on macOS, which is incompatible with how Sphinx handles parallelization. The change was made in macOS *not* due to a new security concern but because Apple decided to prevent *potential* problems with calling something between `fork()` and `exec()` by failing early. This problem is not unique to macOS or even Python, and depends on the implementation.

From looking at the use of `multiprocessing` in Sphinx, I don't believe this will cause issues on macOS.

### Relates
- https://github.com/sphinx-doc/sphinx/pull/6879

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -12,7 +12,6 @@
 
 import os
 import pickle
-import platform
 import sys
 import warnings
 from collections import deque
@@ -194,12 +193,6 @@ class Sphinx:
 
         # say hello to the world
         logger.info(bold(__('Running Sphinx v%s') % sphinx.__display_version__))
-
-        # notice for parallel build on macOS and py38+
-        if sys.version_info > (3, 8) and platform.system() == 'Darwin' and parallel > 1:
-            logger.info(bold(__("For security reasons, parallel mode is disabled on macOS and "
-                                "python3.8 and above. For more details, please read "
-                                "https://github.com/sphinx-doc/sphinx/issues/6803")))
 
         # status code for command-line application
         self.statuscode = 0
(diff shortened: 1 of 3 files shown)


webknjaz (approved):

tk0miya:
Thank you for the proposal. It would be exciting if we can support a parallel build on macOS again!

>From looking at the use of multiprocessing in Sphinx, I don't believe this will cause issues on macOS.

Before merging this, could you let us know why do you think so? I'm not expert of the internal Python and macOS. So I can't determine why Sphinx will not cause issues.

samdoran:
> Before merging this, could you let us know why do you think so? I'm not expert of the internal Python and macOS. So I can't determine why Sphinx will not cause issues.

The problem is that calling into the Objective-C runtime between `fork()` and `exec()` is unsafe. This has always been the case. Apple chose to make calls like this fail early and deterministically rather than (potentially) failing later and in unpredictable ways.

From looking at the system calls using `dtruss` that Sphinx makes as well as testing parallel builds on macOS, Sphinx does not run into this, explicitly or implicitly.

tk0miya:
Thank you for your detailed explanation. I read https://bugs.python.org/issue33725 again and understand the problem (maybe). Indeed, Sphinx does not do anything between fork and exec. Okay, let's see what happens with this.

samdoran:
Thank you!

PR #9800, opened 2021-10-30
Proposal: check if hardcoded URLs can be replaced with extlinks

hoefling:
Subject: This PR adds checks for hardcoded URLs whether they can be replaced with an extlink. This is a feature similar to https://github.com/sphinx-doc/sphinx/pull/9626, but for `extlinks` instead of `intersphinx`.

### Feature or Bugfix
- Feature

### Purpose
If a hardcoded link can be replaced with a crossref, `sphinx` will emit a warning like this:

```
pytest/doc/en/reference/reference.rst:127: WARNING: hardcoded link 'https://github.com/octocat' could be replaced by an extlink (try using ':user:`octocat`' instead)
```

### Example output from running over Sphinx repository
```
/home/oleg.hoefling/projects/private/sphinx/doc/tutorial/narrative-documentation.rst:92: WARNING: hardcoded link 'https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#hyperlink-targets' could be replaced by an extlink (try using ':duref:`hyperlink-targets`' instead)
```

### Relates
- https://github.com/sphinx-doc/sphinx/pull/9626


sphinx/ext/extlinks.py
```diff
@@ -34,13 +35,54 @@
...
 
 
+class ExternalLinksChecker(SphinxPostTransform):
+    """
+    For each external link, check if it can be replaced by an extlink.
+
+    We treat each ``reference`` node without ``internal`` attribute as an external link.
+    """
+
+    default_priority = 900
```
tk0miya: Is there any reason for this priority? It's invoked on a very late step.
https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_transform

sphinx/ext/extlinks.py
```diff
@@ -34,13 +35,54 @@
...
+        lineno = sphinx.util.nodes.get_node_line(refnode)
+        extlinks_config = getattr(self.app.config, 'extlinks', dict())
+
+        for alias, (base_uri, caption) in extlinks_config.items():
+            uri_pattern = re.compile(base_uri.replace('%s', '(?P<value>.+)'))
+            match = uri_pattern.match(uri)
+            if match and match.groupdict().get('value'):
+                # build a replacement suggestion
+                replacement = f":{alias}:`{match.groupdict().get('value')}`"
+                location = (self.env.docname, lineno)
```
tk0miya: You can use a node for the `location` argument of `logger.warning()`. So you don't need to create this variable.

sphinx/ext/extlinks.py
```diff
@@ -34,13 +35,54 @@
...
+        """
+        If the URI in ``refnode`` has a replacement in ``extlinks``,
+        emit a warning with a replacement suggestion.
+        """
+        if 'internal' in refnode or 'refuri' not in refnode:
+            return
+
+        uri = refnode['refuri']
+        lineno = sphinx.util.nodes.get_node_line(refnode)
+        extlinks_config = getattr(self.app.config, 'extlinks', dict())
```
tk0miya: Why do you use `getattr()` here? It will always be available if extlinks extension is loaded.

sphinx/ext/extlinks.py
```diff
@@ -34,13 +35,54 @@
...
+
+        for alias, (base_uri, caption) in extlinks_config.items():
+            uri_pattern = re.compile(base_uri.replace('%s', '(?P<value>.+)'))
+            match = uri_pattern.match(uri)
+            if match and match.groupdict().get('value'):
+                # build a replacement suggestion
+                replacement = f":{alias}:`{match.groupdict().get('value')}`"
+                location = (self.env.docname, lineno)
+                logger.warning(
+                    'hardcoded link %r could be replaced by an extlink (try using %r instead)',
```
tk0miya: Please wrap this message by `sphinx.locale.__()`. It helps message translation (i18n).

CHANGES
```diff
@@ -12,6 +12,9 @@ Deprecated
 
 Features added
 --------------
+ 
+* #9800: extlinks: Emit warning if a hardcoded link is replaceable
+  by an extlink, suggesting a replacement.
```
tk0miya, line 17: I think this improvements are enough stable (not a breaking change). So it would be better to release as 4.x. How about rebase this into the 4.x branch?

tk0miya (changes requested):
LGTM with nits.

hoefling:
@tk0miya thank you, those are very helpful! In fact, I'll apply those to #9626 as well.

sphinx/ext/extlinks.py
tk0miya: IMO, there is no special reason to invoke this feature earlier or later. So it would be better to put it on the normal priority: 500. I'll modify it after merging.

tk0miya (approved):

tk0miya:
Thank you for update. Merging now.

PR #9801, opened 2021-10-30
Support docutils-0.18: Consume generator of Element.traverse()

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- The last fix in i18n module was incorrect because it compares the
"already consumed" generators.  It should compares the lists of nodes.
- refs: #9780 and #9777


tk0miya:
I hope this reduces the errors on the test with docutils-latest. Let's see what tests will be still in failed on next CI.
https://github.com/sphinx-doc/sphinx/actions/workflows/docutils-latest.yml

PR #9815, opened 2021-11-03
base theme: Wrap sidebar components in <div>s

not-my-profile:
Responsive themes inheriting from the base theme might want to display
the sidebar horizontally on narrower screens, which can be easily
achieved with flexbox. This however requires that the individual sidebar
components are wrapped in HTML elements so that flexbox can properly lay
them out.

### DIFF
--- a/sphinx/themes/basic/localtoc.html
+++ b/sphinx/themes/basic/localtoc.html
@@ -8,6 +8,8 @@
     :license: BSD, see LICENSE for details.
 #}
 {%- if display_toc %}
-  <h3><a href="{{ pathto(root_doc)|e }}">{{ _('Table of Contents') }}</a></h3>
-  {{ toc }}
+  <div>
+    <h3><a href="{{ pathto(root_doc)|e }}">{{ _('Table of Contents') }}</a></h3>
+    {{ toc }}
+  </div>
 {%- endif %}
--- a/sphinx/themes/basic/relations.html
+++ b/sphinx/themes/basic/relations.html
@@ -8,12 +8,16 @@
     :license: BSD, see LICENSE for details.
 #}
 {%- if prev %}
-  <h4>{{ _('Previous topic') }}</h4>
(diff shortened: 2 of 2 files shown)


tk0miya:
Looks good to me.

@ericholscher Before merging, I'd like to ask an advice from readthedocs team. Does this bring troubles to sphinx_rtd_theme? Please let me know your opinion.

astrojuanlu:
cc @nienn

ericholscher:
@tk0miya Thanks for the ping. Juanlu already pinged one of our frontend folks, who can hopefully give some feedback soon. @agjohnson would be the other person who would be good to comment on it as well.

nienn:
I don't think we are including any of the modified templates in our theme. Instead we replace the entire **[layout.html](https://github.com/sphinx-doc/sphinx/blob/7a6272f93f48a157027bd3e88f55500f1661676a/sphinx/themes/basic/layout.html)** with **[our custom one](https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html)**.

But I'd rather @agjohnson confirmed this.

_For quick reference, in the basic theme the affected files are included here:_
https://github.com/sphinx-doc/sphinx/blob/7a6272f93f48a157027bd3e88f55500f1661676a/sphinx/themes/basic/layout.html#L68-L71

agjohnson:
sphinx_rtd_theme does technically inherit from `basic`:
https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/theme.conf#L2

However the theme also defines it's own local toc and bread crumbs for page relations, so is not reusing either of these files. If we were to reuse these files, we might have to alter our selectors lsightly, if any changes are required at all.

This change seems safe for sphinx_rtd_theme :+1:  Thanks for the heads up @tk0miya !

tk0miya:
Thank you for your comments and contributions. Now merging.

Issue #9682, opened 2021-09-27
Intersphinx references work without a namespace/identifier but fail if used


umarcor:
### Describe the bug

We are trying to cross-reference several sites built with Sphinx. We found that some references are properly resolved, but others fail. Precisely, if we reference labels in other sites without specifying the namespace/identifier, it does work. However, that is ambiguous, because multiple sites might have the same label. Unfortunately, when using the namespace/identifier, references are not resolved. To our surprise, that is not consistent. It works for some sites, but not for others.

I'm not sure about this being a bug or a misconfiguration of some of the sites.

### How to Reproduce

```
$ git clone https://github.com/edaa-org/edaa-org.github.io
$ cd edaa-org.github.io
$ git clone -b v1.0.0-rc1 https://github.com/buildthedocs/sphinx.theme _theme
$ make html
$ # open _build/html/index and see "Conceptual Model > 3 | Language Model".
```

As seen, refs to `python` or `OSVB` namespaces/identifiers are resolved, but references to pyVHDLModel or pySystemVerilogModel do fail.

### Expected behavior

Refs to pyVHDLModel or pySystemVerilogModel Glossary labels should work as `python:comparisons`  or `OSVB:API:Core`.

### Your project

https://github.com/edaa-org/edaa-org.github.io

### Screenshots

![image](https://user-images.githubusercontent.com/38422348/134972180-58c26aa5-84c1-4d69-ae90-88a5bc46291b.png)

### OS

Linux and MSYS2

### Python version

3.9

### Sphinx version

4.2

### Sphinx extensions

https://github.com/edaa-org/edaa-org.github.io/blob/19f14102835c2fc1224363ee46296058c06a1753/conf.py#L168-L172

### Additional context

Same issue in VHDL/pyVHDLModel#32.

/cc @Paebbels

umarcor:
Upon further testing, this issue seems rather contrived:

- If the intersphinx mapping identifier starts with `py`, it seems that none of the references to that namespace will work. Is it possible that `value[0:2] == 'py'` is compared somewhere in the code without checking for the length of value?

- If the intersphinx mapping identifier is lowercase:
  - The identifier can be used in role `doc` in lowercase only.
  - In role `ref`, it does not matter if the identifier is uppercase, lowercase or camelcase.
- If the intersphinx mapping identifier is camelcase, all roles fail.

astrojuanlu:
[...]

PR #9822, opened 2021-11-06
Intersphinx role (2)

jakobandersen:
Extension of #9062 (not sure if there is a good way simply to reuse that PR somehow)

### Feature or Bugfix
- Feature

### Purpose
Continuation of the work by @tk0miya in #9062 to add a role for performing explicit intersphinx lookups.

### Detail

- the role(-prefix) is called ``external``. I think it is ok to have a bit more verbose than just ``ext`` to make it stand out that this is not an ordinary cross-reference.
- https://github.com/sphinx-doc/sphinx/pull/9062#issuecomment-879861369, when a specific inventory lookup is requested I opted for ``+`` as the delimiter. When I tried the other options the inventory name sort of looked like part of the domain/reftype name instead of a separate item. Semantically a ``+`` also reads nicely as ``:external:python+py:class:`` indeed is requesting a lookup constrained to the ``python`` project and constrained to ``py:class`` objects.
- the role will not fall back to the local project. It will warn if the lookup does not go well:

  - if the inventory name doesn't exist,
  - if the reftype/domain:reftype doesn't exist as a role (taking default-domain and the special ``std`` domain shorthands into acocunt), and
  - if a/the inventory doesn't contain the target.

- it looks like Docutils mandates that role names are made lower-case before the role function is called, which is not good when an inventory name is present. So I haxed the original role name into the role processing. Not sure if there is a better way. Having the inventory name as part of the target text is problematic as intersphinx has no knowledge of the target format, so extracting information from it seems brittle.
- Deprecating the old style prefixing will be done in a subsequent PR, but I think I got the intersphinx docs fully updated to mention this role instead.

### Relates
- #9062
- https://github.com/sphinx-doc/sphinx/pull/8418#issuecomment-758838589
- #9682


sphinx/ext/intersphinx.py
```diff
@@ -466,6 +470,142 @@ def missing_reference(app: Sphinx, env: BuildEnvironment, node: pending_xref,
...
 
 
+class IntersphinxDispatcher(CustomReSTDispatcher):
+    """Custom dispatcher for external role.
+
+    This enables :external:***: roles on parsing reST document.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
```
tk0miya: My bad. This method is unneeded.

doc/usage/extensions/intersphinx.rst
```diff
@@ -162,21 +167,50 @@ linking:
...
+     e.g., ``:external:py:class:`zipfile.ZipFile```, or
+   - ``:external:reftype:`target```,
+     e.g., ``:external:doc:`installation```.
+
+   If you would like to constrain the lookup to a specific external project,
+   then the key of the project, as specified in :confval:`intersphinx_mapping`,
+   is added as well to get the two forms 
 
+   - ``:external:invname+domain:reftype:`target```,
+     e.g., ``:external:python+py:class:`zipfile.ZipFile```, or
```
tk0miya: When I saw this notation for the first time, I felt using `+` sign is cool. But this example makes my opinion opposite.

In my eyes, `:external:python+doc:` is good, but `:external:python+py:class:` is not so good. I guess the `+` sign confuses me. So I found 3 parts `external`, `python+py` and `class` from the latter example.

How about changing this notation?

* `:external+python:py:class:`
  *  Add a invname as a suffix of the "external" part.
  * It's almost similar to the current idea.
* `:external@python:py:class:`
  * Add a invname as a suffix of the "external" part using a `@` sign.
  * `@` sign is usually used to represent a location because it's equivalent to "at".
  * It looks like a bit complicated.

jakobandersen: I agree, the ``+`` looks like it has "higher precedence" than ``:``. I like ``@`` but if I read Docutils correctly (https://github.com/sphinx-doc/sphinx/pull/9062#issuecomment-879861369) then we can't use it. Only ``-._+`` can be used.
I think your first suggestion with ``:external+python:py:class:`` is the best. Just for the visual picture, the full set of variations are then
- ``:external+python:py:class:``, specific inventory, specific domain,
- ``:external:py:class:``, all inventories, specific domain,
- ``:external+python:class:``, specific inventory, default domain,
- ``:external:class:``, all inventories, default domain.

sphinx/ext/intersphinx.py
jakobandersen: Indeed, removed now.

doc/usage/extensions/intersphinx.rst
jakobandersen: The PR is now updated with this new syntax.

astrojuanlu:
Would be cool to get this in for 4.4.0! 🚀

jakobandersen:
> Would be cool to get this in for 4.4.0! rocket

Indeed, though I think @tk0miya is planning to release it this weekend, so not sure if this is enough time.

sphinx/ext/intersphinx.py
```diff
@@ -466,6 +470,145 @@ def missing_reference(app: Sphinx, env: BuildEnvironment, node: pending_xref,
...
 
+class IntersphinxDispatcher(CustomReSTDispatcher):
+    """Custom dispatcher for external role.
+
+    This enables :external:***:/:external+***: roles on parsing reST document.
+    """
+
+    def role(self, role_name: str, language_module: ModuleType, lineno: int, reporter: Reporter
+             ) -> Tuple[RoleFunction, List[system_message]]:
+        if len(role_name) > 9 and role_name.startswith('external') and role_name[8] in ':+':
```
tk0miya: How abou this?
```suggestion
        if len(role_name) > 9 and role_name.startswith(('external:', 'external+')):
```

sphinx/ext/intersphinx.py
```diff
@@ -466,6 +470,145 @@ def missing_reference(app: Sphinx, env: BuildEnvironment, node: pending_xref,
...
+            # :external+inv:role:        or
+            # :external+inv:domain:role:
+            inv, name = name.split(':', 1)
+            return inv, name
+        else:
+            assert typ == ':'
+            # we look in all inventories, i.e.,
+            # :external:role:        or
+            # :external:domain:role:
+            return None, name
```
tk0miya: It's better to use regexp to parse the role name, I think.
```
pattern = re.compile("external(\+\s+)?:.*")
return pattern.match(name).groups()  # => (invname, rolename)
```

tk0miya (approved):
LGTM with nits.

My comments are minor refactoring ideas. So it's okay to fix them after merging. What do you think?

tk0miya:
I'll make a new package tomorrow night in JST (maybe 20hours later from now). So it's not too late to merge this.

sphinx/ext/intersphinx.py
jakobandersen: Good idea, didn't know you could give a tuple to ``startswith``.

sphinx/ext/intersphinx.py
jakobandersen: Hmm, it would make the code a bit more elegant. I'll try to take a closer look before the release, but otherwise feel free to merge and we can refactor it afterwards.

jakobandersen: Done.

tk0miya:
Merging now. Thank you for your great work!

PR #9830, opened 2021-11-08
Process only directive before traversing doctree

### LINKED ISSUE #9819: toctree does not consider "only" directive
### Describe the bug

`toctree` when used in templates does not consider `.. only` directive.

In `index.rst` Use `.. only` to only add some documents for singlehtml builder.

In the template, using `toctree`, and building multi-page html, all the toc tree shows up.

Screen Shot 2021-11-04 at 4 57 51 PM

Since this is a multi-page html build, should the `.. only` cause the `toctree` to only have `two`?

### How to Reproduce

Use sphinx-quickstart to set up a project with any values.

Update source/index.rst with this toc:
```
.. only:: singlehtml_builder

   .. toctree::
      one

.. toctree::
   :maxdepth: 2
   :caption: Contents:

   two
```
Create `source/one.rst` and `source/two.rst`.
Create a custom `_templates/layout.html` that uses `toctree()`:
```
{% extends "!layout.html" %}

{%- block sidebar1 %}
{{ super() }}
{{ toctree() }}
{% endblock %
```

`sphinx-build html`

Open `build/html/index.html`

### Expected behavior

TOC only contains link to "two"

### Your project

https://github.com/ngzhian/sphinx-repro

### Screenshots

_No response_

### OS

Linux

### Python version

3.9.7

### Sphinx version

4.2.0

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

ngzhian:
This changes the output of `toctree` if it is wrapped by a `only`, so it is a breaking change. I think this is a bugfix, so it can maybe go into 4.2 too?

### Feature or Bugfix

- Bugfix (consider this a bug, `toctree` should take into acount `only`.

### Purpose
- This ensures that a `toctree` that is wrapped by a `only` is only traversed when generated the global `toctree` when the tags match.

### Relates
- Fixes #9819.

### DIFF
--- a/sphinx/environment/adapters/toctree.py
+++ b/sphinx/environment/adapters/toctree.py
@@ -324,6 +324,8 @@ class TocTree:
         else:
             kwargs['maxdepth'] = int(kwargs['maxdepth'])
         kwargs['collapse'] = collapse
+        # Handle only directives that wrap an entire toctree.
+        process_only_nodes(doctree, builder.tags)
         for toctreenode in doctree.traverse(addnodes.toctree):
             toctree = self.resolve(docname, builder, toctreenode, prune=True, **kwargs)
             if toctree:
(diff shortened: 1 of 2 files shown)


ngzhian:
Ping, any interest in this patch?

AA-Turner:
Please resolve conflicts and add a CHANGES entry

A

PR #9831, opened 2021-11-09
Allow autosummary to respect __all__

### LINKED ISSUE #2021: autosummary ignores __all__ for modules
I'm having an issue where in `module.rst` template I do the following:

```
{% for item in classes %}
.. autoclass:: {{ item }}
   :members:
{%- endfor %}
```

But some imported object are not being filtered out. I checked the implementation:

https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/generate.py#L167

And this function is not considering the `__all__` attribute for modules to determine the public objects.

The main issue is because I tried the following:

```
.. autosummary::
   :toctree: _reference
{% for item in classes %}
   {{ item }}
{%- endfor %}
```

To try to generate a recursive page for each class. 

The problem is that the page gets generated, but the `autosummary` directive does check if the object is truly public (considering `__all__`, [i think here](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autodoc.py#L1003)) and thus it doesn't output that entry to the `toctree`, causing a build error when running with `-W` because the generated `*.rst` file doesn't belong to any `toctree`.

Regards.

Yoshanuikabundi:
Python's `__all__` attribute defines the public API of a module or package. Autosummary should support this.

### Feature or Bugfix

- Feature

### Purpose

Python modules and packages can supply a list of names in the `__all__` attribute to declare the public interface. If this is present, Autosummary should use it to determine which objects to document.

### Detail
- Adds the `autosummary_ignore___all__` configuration value, `True` by default. This anticipates a future breaking change where this will be set to `False` as discussed in #9455 
- If `autosummary_ignore___all__` is `False`, autosummary will use the `__all__` attribute to discover the module's members
- If `autosummary_ignore___all__` is `False` but a module does not have the `__all__` attribute, the previous behaviour is kept
- Regression tests have been added and updated
- The `-a`/`--respect-__all__` command line switch has been added to autogen to support this behaviour on the CLI
- The configuration value is documented
- Currently, all submodules are documented, regardless of whether they're in `__all__`. This matches the behaviour of `pydoc`, but I think submodules may or may not be a part of the public API and should be controlled by the same mechanism. I haven't had time to track this down yet but should be able to do so tomorrow. Thoughts?

### Relates
- closes #2021

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@ Features added
 
 * #9815: html theme: Wrap sidebar components in div to allow customizing their
   layout via CSS
+* #9831: Autosummary now documents only the members specified in a module's ``__all__`` attribute if :confval:`autosummary_ignore___all__` is set to ``False``. The default behaviour is unchanged.
 
 Bugs fixed
 ----------
--- a/CHANGES
+++ b/CHANGES
@@ -16,7 +16,7 @@ Features added
 * #9815: html theme: Wrap sidebar components in div to allow customizing their
   layout via CSS
 * #9831: Autosummary now documents only the members specified in a module's
-  ``__all__`` attribute if :confval:`autosummary_ignore___all__` is set to
+  ``__all__`` attribute if :confval:`autosummary_ignore_module_all` is set to
   ``False``. The default behaviour is unchanged. Autogen also now supports
   this behavior with the ``--respect-module-all`` switch.
 
--- a/CHANGES
+++ b/CHANGES
@@ -15,7 +15,10 @@ Features added
 
 * #9815: html theme: Wrap sidebar components in div to allow customizing their
   layout via CSS
-* #9831: Autosummary now documents only the members specified in a module's ``__all__`` attribute if :confval:`autosummary_ignore___all__` is set to ``False``. The default behaviour is unchanged.
+* #9831: Autosummary now documents only the members specified in a module's
+  ``__all__`` attribute if :confval:`autosummary_ignore___all__` is set to
+  ``False``. The default behaviour is unchanged. Autogen also now supports
+  this behavior with the ``--respect-module-all`` switch.
 
 Bugs fixed
 ----------
--- a/doc/usage/extensions/autosummary.rst
+++ b/doc/usage/extensions/autosummary.rst
@@ -201,12 +201,12 @@ also use these config values:
 
    .. versionadded:: 2.1
 
-   .. versionchanged:: 4.3
+   .. versionchanged:: 4.4
 
-      If ``autosummary_ignore___all__`` is ``False``, this configuration value
-      is ignored for members listed in ``__all__``.
+      If ``autosummary_ignore_module_all`` is ``False``, this configuration
+      value is ignored for members listed in ``__all__``.
 
-.. confval:: autosummary_ignore___all__
+.. confval:: autosummary_ignore_module_all
 
    If ``False`` and a module has the ``__all__`` attribute set, autosummary
    documents every member listed in ``__all__`` and no others. Default is
@@ -215,10 +215,10 @@ also use these config values:
    Note that if an imported member is listed in ``__all__``, it will be
    documented regardless of the value of ``autosummary_imported_members``. To
    match the behaviour of ``from module import *``, set
-   ``autosummary_ignore___all__`` to False and ``autosummary_imported_members``
-   to True.
+   ``autosummary_ignore_module_all`` to False and
+   ``autosummary_imported_members`` to True.
 
-   .. versionadded:: 4.3
+   .. versionadded:: 4.4
 
 .. confval:: autosummary_filename_map
 
--- a/doc/usage/extensions/autosummary.rst
+++ b/doc/usage/extensions/autosummary.rst
@@ -201,6 +201,25 @@ also use these config values:
 
    .. versionadded:: 2.1
 
+   .. versionchanged:: 4.3
+
+      If ``autosummary_ignore___all__`` is ``False``, this configuration value
+      is ignored for members listed in ``__all__``.
+
+.. confval:: autosummary_ignore___all__
+
+   If ``False`` and a module has the ``__all__`` attribute set, autosummary
+   documents every member listed in ``__all__`` and no others. Default is
+   ``True``
+
+   Note that if an imported member is listed in ``__all__``, it will be
+   documented regardless of the value of ``autosummary_imported_members``. To
+   match the behaviour of ``from module import *``, set
+   ``autosummary_ignore___all__`` to False and ``autosummary_imported_members``
+   to True.
+
+   .. versionadded:: 4.3
+
 .. confval:: autosummary_filename_map
 
    A dict mapping object names to filenames. This is necessary to avoid
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -826,6 +826,6 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value('autosummary_mock_imports',
                          lambda config: config.autodoc_mock_imports, 'env')
     app.add_config_value('autosummary_imported_members', [], False, [bool])
-    app.add_config_value('autosummary_ignore___all__', True, 'env', bool)
+    app.add_config_value('autosummary_ignore_module_all', True, 'env', bool)
 
     return {'version': sphinx.__display_version__, 'parallel_read_safe': True}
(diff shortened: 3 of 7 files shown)


doc/usage/extensions/autosummary.rst
```diff
@@ -201,6 +201,25 @@ also use these config values:
 
    .. versionadded:: 2.1
 
+   .. versionchanged:: 4.3
+
+      If ``autosummary_ignore___all__`` is ``False``, this configuration value
+      is ignored for members listed in ``__all__``.
+
+.. confval:: autosummary_ignore___all__
```
tk0miya: In autodoc extension, this option is called as `:ignore-module-all:`. So it would be better to follow its name.
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-members

doc/usage/extensions/autosummary.rst
```diff
@@ -201,6 +201,25 @@ also use these config values:
...
+
+   If ``False`` and a module has the ``__all__`` attribute set, autosummary
+   documents every member listed in ``__all__`` and no others. Default is
+   ``True``
```
tk0miya, line 213: Please disable this by default at this moment. I agree this must be useful. But I'd not like to change the behavior of autosummary during 4.x to keep compatibility. So it should be changed since 5.0.

doc/usage/extensions/autosummary.rst
```diff
@@ -201,6 +201,25 @@ also use these config values:
...
+
+   Note that if an imported member is listed in ``__all__``, it will be
+   documented regardless of the value of ``autosummary_imported_members``. To
+   match the behaviour of ``from module import *``, set
+   ``autosummary_ignore___all__`` to False and ``autosummary_imported_members``
+   to True.
+
+   .. versionadded:: 4.3
```
tk0miya: Unfortunately, we'll ship 4.3.0 soon. So it's timed out to merge this into it. Please change this to 4.4.

CHANGES
```diff
@@ -52,6 +52,7 @@ Features added
   inventory specification. Specific types of cross-references can be disabled,
   e.g., ``std:doc`` or all cross-references in a specific domain,
   e.g., ``std:*``.
+* #9831: Autosummary now documents only the members specified in a module's ``__all__`` attribute if :confval:`autosummary_ignore___all__` is set to ``False``. The default behaviour is unchanged.
```
tk0miya: Could you fold text by 80 chars?

sphinx/ext/autosummary/generate.py
```diff
@@ -630,6 +645,10 @@ def get_parser() -> argparse.ArgumentParser:
                         dest='imported_members', default=False,
                         help=__('document imported members (default: '
                                 '%(default)s)'))
+    parser.add_argument('-a', '--respect-__all__', action='store_false',
```
tk0miya: +1 for the word "respect". So this will be renamed to `--respect-module-all`.
Please add an entry for this option into `doc/man/sphinx-autogen.rst` and CHANGES.

tests/test_ext_autosummary.py
```diff
@@ -313,6 +340,7 @@ def test_autosummary_generate(app, status, warning):
     assert doctree[3][0][0][2][5].astext() == 'autosummary_dummy_module.qux\n\na module-level attribute'
 
     module = (app.srcdir / 'generated' / 'autosummary_dummy_module.rst').read_text()
+    print(module)
```
tk0miya: :-)

tk0miya (commented):
LGTM basically. Let's go forward!

doc/usage/extensions/autosummary.rst
Yoshanuikabundi: The new behaviour is disabled by default - ie, we ignore `__all__` by default, so `automodule_ignore_module_all` is `True`. If this is too confusing I can flip it so it's `automodule_respect_module_all`, but I thought this way around would make more sense once the default is changed in 5.0

Yoshanuikabundi:
I believe this is now ready to merge. From everything I've read, I don't think submodules are supposed to go into `__all__`, and it looks like that would have to be a more complicated change anyway, so it might be better to leave it to a future PR.

CHANGES
```diff
@@ -57,6 +57,10 @@ Features added
   e.g., ``std:*``.
 * #9623: Allow to suppress "toctree contains reference to excluded document"
   warnings using :confval:`suppress_warnings`
+* #9831: Autosummary now documents only the members specified in a module's
+  ``__all__`` attribute if :confval:`autosummary_ignore___all__` is set to
```
tk0miya: This reference should be also updated. And it would be nice if you'll move this entry to 4.4.

doc/usage/extensions/autosummary.rst
tk0miya: Ah, sorry. It's my wrong.

sphinx/ext/autosummary/__init__.py
```diff
@@ -826,5 +826,6 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value('autosummary_mock_imports',
                          lambda config: config.autodoc_mock_imports, 'env')
     app.add_config_value('autosummary_imported_members', [], False, [bool])
+    app.add_config_value('autosummary_ignore___all__', True, 'env', bool)
```
tk0miya: This should be also renamed.

sphinx/ext/autosummary/generate.py
```diff
@@ -68,6 +68,7 @@ def __init__(self, translator: NullTranslations) -> None:
 
         self.config.add('autosummary_context', {}, True, None)
         self.config.add('autosummary_filename_map', {}, True, None)
+        self.config.add('autosummary_ignore___all__', True, 'env', bool)
```
tk0miya: This file also refers to old names.

sphinx/ext/autosummary/generate.py
```diff
@@ -630,6 +645,10 @@ def get_parser() -> argparse.ArgumentParser:
                         dest='imported_members', default=False,
                         help=__('document imported members (default: '
                                 '%(default)s)'))
+    parser.add_argument('-a', '--respect-module-all', action='store_false',
+                        dest='ignore___all__', default=True,
```
tk0miya: `ignore_all` ?

sphinx/ext/autosummary/__init__.py
Yoshanuikabundi: Whooooops

sphinx/ext/autosummary/generate.py
Yoshanuikabundi: Renamed to `respect_module_all` and flipped the boolean, then added a `not` to line 668

Yoshanuikabundi:
Fixed those, grepped everything for any other option names I missed, and rebased on current 4.x

sphinx/ext/autosummary/generate.py
```diff
@@ -216,11 +217,25 @@ def scan(self, imported_members: bool) -> List[str]:
...
-                # list not-imported members up
+                # list not-imported members
+                members.append(name)
+            elif '__all__' in dir(self.object) and not self.app.config.autosummary_ignore_module_all:
+                # list members that have __all__ set
                 members.append(name)
 
         return members
 
+def members_of(conf: Config, obj: Any) -> Sequence[str]:
```
tk0miya: The main target of this function is `obj`. So it would be better to move it to the first argument of the arglist. Sorry for the noise. But I think the devil's in the details.

tk0miya (changes requested):
LGTM with nits :-)

tk0miya:
It seems flake8 and mypy warns something. Could you check them please?

sphinx/ext/autosummary/generate.py
Yoshanuikabundi: No worries :)

Yoshanuikabundi:
Fixed. What are your thoughts on the submodules thing? 

> Currently, all submodules are documented, regardless of whether they're in __all__. This matches the behaviour of pydoc, but I think submodules may or may not be a part of the public API and should be controlled by the same mechanism. Thoughts?

Having thought about it I think I do want this behaviour for my projects, but I can't find clear docs on how modules should be included in `__all__` - absolute path, relative path, imported and then treated as locals... Would welcome any thoughts you had! And would be happy to leave this as separately configurable behaviour for a future PR.

tk0miya (approved):
LGTM!

tk0miya:
>What are your thoughts on the submodules thing?

I agree it would be better to add a way to control what submodules are documented (or undocumented) on the recursive build. But I'm not sure using `__all__` is better for that purpose. IMO, it does not related to the submodules.

tk0miya:
As I commented above, controlling submodules is another topic. So I'd like to merge this. What do you think?

Yoshanuikabundi:
Cool! I'll open an issue to discuss design of the submodules thing when I get a chance. I agree that this is ready to merge :)

tk0miya:
Okay, merging now!

tk0miya:
Thank you for your great work!

Issue #1863, opened 2015-04-23
`make singlehtml` theme=basic: does not have table of contents


bialix:
There is no table of contents in singlehtml if I select basic theme. I need simple html file with no side bar but with table of contents at the begining of the document.

I've tried to use reST directive `.. contents::` but it provided only 2 links: to the title of document and to "Indices and tables" section at the end, but not to my sections.

sphinx 1.3.1.

bialix:
I've invoked sphinx as 

make singlehtml SPHINXOPTS="-Dhtml_theme=basic"

no theme selected in my conf.py

davidism:
Duplicate of #2589

westurner:
Is there a solution for this?
- https://github.com/sphinx-contrib/confluencebuilder/pull/547/files

westurner:
- https://github.com/sphinx-doc/sphinx/blob/7406d4222f1d105cf9f3c2b2b36d0700658148d1/sphinx/builders/singlehtml.py#L141-L152

DarkArc:
If I'm understanding correctly, this was closed as a duplicate of https://github.com/sphinx-doc/sphinx/issues/2589.

Neither issue seems to be exactly what we're experiencing but it's close, so I wanted to make sure this is captured somewhere.

Effectively, when we do a dirhtml build we get a nice sidebar like:
```
- A
  - A.1
  - A.2
- B
  - B.1
  - B.2
```
(with e.g., the subitems like A.1 and A.2 only visible if you're on page A)

however using the singlehtml build we get a not as nice sidebar like:
```
- A
- B
```

It would be much better if that sidebar was either fully elaborated or elaborated based on where the browser is scrolled in the document (using some JavaScript).

PR #9839, opened 2021-11-11
option to include toctree in inlined builds

jdknight:
Subject: option to include toctrees in inlined builds

### Feature or Bugfix
- Feature

### Purpose
- nodes: support not replacing inlined toctree nodes
- singlehtml: option to include toc tree entries in the main document

### Detail

The `inline_all_toctrees` will replace toctree nodes with the contents of each toctree entry. While replacing the toctree node is preferable for most builders (and themes), some uses cases may wish to leave the TOC tree in. For example, if an HTML theme does not support a sidebar which produces a TOC tree, the resulting documentation will not have any TOC tree in the final output.

This commit introduces a `replace` option to the `inline_all_toctrees` function which allows a builder to explicitly decide if toctree nodes are replaced (the previous use case and the default state) or leave the toctree nodes in (which can later be resolved on).

Documentation which use an HTML theme which do not support a table of contents (e.g. `basic`) or a configuration which has `nosidebar` set, there will be no TOC tree generated in the final document. To support the use case where a user wishes to have TOC tree entries rendered in the document, a `singlehtml_toctree` option can be set to `True` to enable this.

---

An end goal of this pull request is to extend the `inline_all_toctrees` capability to third-party builders which may also want to inline documents, but support leaving the TOC tree entries on a page.

### Relates
-  #1863
- sphinx-contrib/confluencebuilder#447

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1553,6 +1553,13 @@ Options for Single HTML output
    are ignored.  For more information, refer to :confval:`html_sidebars`.  By
    default, it is same as :confval:`html_sidebars`.
 
+.. confval:: singlehtml_toctree
+
+   If true, TOC tree directives are inserted at their current locations,
+   instead of being replaced.  Default is ``False``.  This can be useful if a
+   documentation's theme does not support sidebar TOC trees or is configured
+   with sidebars disabled.
+
 
 .. _htmlhelp-options:
 
--- a/sphinx/builders/singlehtml.py
+++ b/sphinx/builders/singlehtml.py
@@ -76,9 +76,10 @@ class SingleFileHTMLBuilder(StandaloneHTMLBuilder):
     def assemble_doctree(self) -> nodes.document:
         master = self.config.root_doc
         tree = self.env.get_doctree(master)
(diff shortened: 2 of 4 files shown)


jdknight:
Another consideration with respect to changing the `inline_all_toctrees` function, I have been also looking into a way to have a `singlehtml` builder's sections not nest inlined sections based off a parent one. So if a documentation has an index page with a toctree for other documents, instead of only having a single h1 header, inlined documents can also have h1 headers. This is performed with a crude hack:

```
        for node in newnodes:
            toctreenode.parent.parent.parent.append(node)
```

Not saying this should be done (and I can imaging there can be issues beyond the initial toctree entry), but I am wondering if there may be a better argument introduction into `inline_all_toctrees`. For example, maybe providing a call where the invoker can manipulate how populated new nodes are injected in.

```
    def inline_all_toctrees(builder: "Builder", docnameset: Set[str], docname: str,
                        tree: nodes.document, colorfunc: Callable, traversed: List[str],
                        hook: Callable = None) 
            ...

            if hook:
                hook(docname, toctreenode, newnodes)
            else:
                toctreenode.parent.replace(toctreenode, newnodes)
    ...
```

PR #9846, opened 2021-11-13
Exclude substitution definitions from gettext build (take 3)

alvinhochun:
### Feature or Bugfix

- Bugfix

### Purpose

### Detail

### Relates
- #9428

Supersedes #9629

cc @therahedwig


sphinx/builders/gettext.py
```diff
@@ -144,6 +144,17 @@ def compile_catalogs(self, catalogs: Set[CatalogInfo], message: str) -> None:
...
+        def is_node_part_of_substitution_definition(node: nodes.Node) -> bool:
+            """Check the node and its parents to see if any of them is a substitution
+            definition.
+            """
+            # return False
+            if isinstance(node, nodes.substitution_definition):
+                return True
+            if node.parent:
+                return is_node_part_of_substitution_definition(node.parent)
+            return False
```
AA-Turner:
```suggestion
            while node.parent:
                if isinstance(node, nodes.substitution_definition):
                    return True
                node = node.parent
            return False
```

sphinx/builders/gettext.py
```diff
@@ -144,6 +144,17 @@ def compile_catalogs(self, catalogs: Set[CatalogInfo], message: str) -> None:
         return
 
     def write_doc(self, docname: str, doctree: nodes.document) -> None:
+        def is_node_part_of_substitution_definition(node: nodes.Node) -> bool:
```
AA-Turner:
```suggestion
        def is_node_in_substitution_definition(node: nodes.Node) -> bool:
```

sphinx/builders/gettext.py
```diff
@@ -152,6 +163,8 @@ def write_doc(self, docname: str, doctree: nodes.document) -> None:
                 catalog.add(msg, node)
 
         for node, msg in extract_messages(doctree):
+            if is_node_part_of_substitution_definition(node):
```
AA-Turner:
```suggestion
            if is_node_in_substitution_definition(node):
```

sphinx/environment/collectors/asset.py
```diff
@@ -79,8 +79,11 @@ def process_doc(self, app: Sphinx, doctree: nodes.document) -> None:
 
                 # Update `node['uri']` to a relative path from srcdir
                 # from a relative path from current document.
+                original_uri = node['uri']
                 node['uri'], _ = app.env.relfn2path(imguri, docname)
                 candidates['*'] = node['uri']
+                if node['uri'] != original_uri:
+                    node['original_uri'] = original_uri
```
AA-Turner, line 78: @alvinhochun -- is this needed for this PR? If not, could it be in a different PR please?

A

sphinx/builders/gettext.py
```diff
@@ -148,6 +148,17 @@ def compile_catalogs(self, catalogs: set[CatalogInfo], message: str) -> None:
         return
 
     def write_doc(self, docname: str, doctree: nodes.document) -> None:
+        def is_node_in_substitution_definition(node: nodes.Node) -> bool:
+            """Check the node and its parents to see if any of them is a substitution
+            definition.
+            """
+            # return False
```
AA-Turner:
```suggestion
```

tests/test_build_gettext.py
```diff
@@ -202,3 +202,73 @@ def test_build_single_pot(app):
...
+
+    def msgid_getter(msgid):
+        m = _msgid_getter(msgid)
+        if m:
+            return m.groups()[0]
+        return None
+
+    assert (app.outdir / 'prolog_epilog_substitution.pot').isfile()
+    pot = (app.outdir / 'prolog_epilog_substitution.pot').read_text()
+    msgids = [_f for _f in map(msgid_getter, pot.splitlines()) if _f]
```
AA-Turner:
```suggestion
    msgids = list(filter(None, map(msgid_getter, pot.splitlines())))
```

tests/test_build_gettext.py
```diff
@@ -202,3 +202,73 @@ def test_build_single_pot(app):
...
+
+    _msgid_getter = re.compile(r'msgid "(.*)"').search
+
+    def msgid_getter(msgid):
+        m = _msgid_getter(msgid)
+        if m:
+            return m.groups()[0]
+        return None
+
+    assert (app.outdir / 'prolog_epilog_substitution.pot').isfile()
```
AA-Turner:
```suggestion
    assert os.path.isfile(app.outdir / 'prolog_epilog_substitution.pot')
```

sphinx/environment/collectors/asset.py
alvinhochun: Hi, I don't exactly remember the context for this commit, if that change make sense to you would you mind splitting that commit into its own PR for me? Otherwise you can try to revert or remove that commit from this PR.

(FWIW it *might* have been related to image URIs inside substitutions, but I'm just guessing from the test cases. See if the tests pass without this commit.)

alvinhochun: The context might be [this comment](https://github.com/sphinx-doc/sphinx/pull/9454#issuecomment-884172757).

AA-Turner:
Thanks @alvinhochun!

A

PR #9867, opened 2021-11-18
Fix #9864: mathjax: Failed to render equations via MathJax v2

### LINKED ISSUE #9864: Add option to choose mathjax load method

Plotly uses mathjax v2, so to render graphs using this package we have to make sphinx also use mathjax v2 with (see #9563):

```python
mathjax_path = (
    "https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js?config=TeX-AMS-MML_HTMLorMML"
)
```
The problem in Sphinx 4.3 is that now, as discussed in #9450, mathjax is loaded with the “defer” option instead of “async”, and this breaks plotly rendering.

In the mathjax2_config dict, it would be nice to have an option "load_method" in which we would be able to choose "async" or "defer"

**Additional context**
Error that shows up in the console:
image

image

tk0miya:
### Feature or Bugfix
- Feature
- Bugfix

### Purpose
- MathJax library has been loaded via "defer" strategy since v4.3.0.  But
it prevents to work MathJax v2.  This rollbacks the change and use
"async" strategy as default again.
- To support changing "defer" strategy, this introduces a new confval;
mathjax_loading_method to switch the loading method of MathJax.
- refs: #9864

### DIFF
--- a/sphinx/ext/mathjax.py
+++ b/sphinx/ext/mathjax.py
@@ -85,7 +85,6 @@ def install_mathjax(app: Sphinx, pagename: str, templatename: str, context: Dict
         if app.config.mathjax_options:
             options.update(app.config.mathjax_options)
         if 'async' not in options and 'defer' not in options:
-            print(options, app.config.mathjax3_config)
             if app.config.mathjax3_config:
                 # Load MathJax v3 via "defer" method
                 options['defer'] = 'defer'
--- a/sphinx/ext/mathjax.py
+++ b/sphinx/ext/mathjax.py
@@ -81,10 +81,18 @@ def install_mathjax(app: Sphinx, pagename: str, templatename: str, context: Dict
     domain = cast(MathDomain, app.env.get_domain('math'))
     if app.registry.html_assets_policy == 'always' or domain.has_equations(pagename):
         # Enable mathjax only if equations exists
-        options = {'defer': 'defer'}
+        options = {}
         if app.config.mathjax_options:
             options.update(app.config.mathjax_options)
-        app.add_js_file(app.config.mathjax_path, **options)  # type: ignore
+        if 'async' not in options and 'defer' not in options:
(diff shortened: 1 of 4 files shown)


mgeier:
Is it really necessary to add a new config value for this?

What about allowing users to specify

```python
mathjax_options = {'async': 'async'}
```

?

The logic in the implementation would have to change like this: add `'defer': 'defer'` to the options, but only if they don't already contain `'async': 'async'`. That seems straightforward to me, or am I missing something?

mgeier:
It's probably better to check if the `config_options` contain either of the keys `'async'` or `'defer'`, and only add `'defer': 'defer'` if none of those *keys* is available (regardless of their values).

tk0miya:
It's a good idea.

>It's probably better to check if the config_options contain either of the keys 'async' or 'defer', and only add 'defer': 'defer' if none of those keys is available (regardless of their values).

My thought is the default setting should be "async" during v4.x to keep compatibility.

sphinx/ext/mathjax.py
```diff
@@ -81,10 +81,18 @@ def install_mathjax(app: Sphinx, pagename: str, templatename: str, context: Dict
     domain = cast(MathDomain, app.env.get_domain('math'))
     if app.registry.html_assets_policy == 'always' or domain.has_equations(pagename):
         # Enable mathjax only if equations exists
-        options = {'defer': 'defer'}
+        options = {}
         if app.config.mathjax_options:
             options.update(app.config.mathjax_options)
-        app.add_js_file(app.config.mathjax_path, **options)  # type: ignore
+        if 'async' not in options and 'defer' not in options:
+            print(options, app.config.mathjax3_config)
```
mgeier: This is probably not intended to stay in there.

tk0miya: oops...

Issue #9872, opened 2021-11-19
Class namespace collision between autodoc signatures and docutils >= 0.17


ebolyen:
### Describe the bug

In [docutils version 0.17](https://docutils.sourceforge.io/HISTORY.html#release-0-17-2021-04-03), the following change was made:

> `Use HTML text-level tags <small>, <s>, <q>, <dfn>, <var>, <samp>, <kbd>, <i>, <b>, <u>, <mark>, and <bdi> if a unique, matching class value is found in inline and literal elements. Use <ins> and <del> if a unique matching class value is found in inline, literal, or container elements. Use <small> for generated code line numbers.`

This appears to conflict with the signature line generated by autodoc, which adds Pygment classes to certain nodes:

https://github.com/sphinx-doc/sphinx/blob/ecee2335c561618800ce7562db890e68646ad020/sphinx/addnodes.py#L332-L334

**The result is a rather funky looking signature with an unusual DOM:**
![image](https://user-images.githubusercontent.com/3976804/142688671-8cc76ce0-1e6d-4945-a8eb-d79a9252d5f5.png)
![image](https://user-images.githubusercontent.com/3976804/142688740-49bdc3f1-5b21-42d8-ab17-6ef0340d4251.png)

---

**This can be fixed by changing the added class to something like `pygments-s`:**
![image](https://user-images.githubusercontent.com/3976804/142688885-1edfa72d-cf7d-4283-9b48-ead387b54553.png)
![image](https://user-images.githubusercontent.com/3976804/142690348-c1bf559d-6af7-415f-ad77-7a019cef3cff.png)

That said, the above probably won't play nicely with the many existing pygment stylesheets, so I don't really know the best approach.

### How to Reproduce

Use a `typing.Literal['foo']` in an autodoc'ed signature line.

### Your project

https://github.com/qiime2/qiime2/pull/610

### OS

Linux

### Python version

3.8

### Sphinx version

4.3

### Sphinx extensions

sphinx.ext.autodoc

### Extra tools

Firefox/Chrome

ebolyen:
I should mention I also tried returning multiple classes like `["pygments", "s"]`, 
but all that earned me from docutils was a `<s class="pygments">...</s>`

lorenzncode:
I also see this issue building docs for another project:

https://cadquery--916.org.readthedocs.build/en/916/classreference.html#cadquery.Workplane.mirror

[...]

PR #9873, opened 2021-11-19
BUG: prevent automatic inline conversion by class

ebolyen:
### Feature or Bugfix
- Bugfix

### Purpose
To prevent inline nodes from being automatically coerced into specific HTML 5 elements.

This was a new feature added in docutils 0.17 and conflicts with some of the short class names added in https://github.com/sphinx-doc/sphinx/pull/9023, namely "s" which is automatically coerced into a strike-through element.

### Relates
See #9872 for motivation.

---

This is largely a drive-by effort. I don't know if this is the way you all would prefer to solve this, but I was able to get it working, and it seems like a very mild change. No hard feelings if this isn't the way you want to go.

### DIFF
--- a/sphinx/writers/html5.py
+++ b/sphinx/writers/html5.py
@@ -13,7 +13,7 @@ import posixpath
 import re
 import urllib.parse
 import warnings
-from typing import TYPE_CHECKING, Iterable, Tuple, cast
+from typing import TYPE_CHECKING, Iterable, Set, Tuple, cast
 
 from docutils import nodes
 from docutils.nodes import Element, Node, Text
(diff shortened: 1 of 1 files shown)


sphinx/writers/html5.py
```diff
@@ -70,6 +70,10 @@ def __init__(self, document: nodes.document, builder: Builder) -> None:
         self._table_row_indices = [0]
         self._fieldlist_row_indices = [0]
         self.required_params_left = 0
+        # Override docutils.writers.html5_polyglot:HTMLTranslator
+        # otherwise, nodes like <inline classes="s">...</inline> will be
+        # converted to <s>...</s> by `visit_inline`.
+        self.supported_inline_tags = set()
```
tk0miya: In the superclass, this is defined as a class variable. So it would be better to override this as a class variable also.

tk0miya (commented):
Good catch! I've never noticed such a behavior. LGTM with nits.

tk0miya:
I think this is a critical bug. So could you rebase this into the 4.3.x branch? Then I'll release a new package including this fix as 4.3.1 soon.

ebolyen:
@tk0miya thanks for the review! I'll update to a class-var and rebase first thing Monday :)

ebolyen:
@tk0miya ready for re-review, thanks!

ebolyen:
Just kidding...

ebolyen:
Ok, copy-paste error fixed. All set.

tk0miya:

PR #9928, opened 2021-12-02
doc: Fix autodoc extension example

joukewitteveen:
Subject: Minor fixes to the example code of the autodoc extension tutorial

### Feature or Bugfix

- Bugfix (in example code)

---
Documenter class attributes:
- `directivetype` is set to mimic `ClassDocumenter`. Reflect that.

`can_document_member` method fixes:
- `isinstance` would work on the enum members, but that is not what we want here. `issubclass` raises a TypeError when called on objects that are not classes.


tk0miya:
Thanks!

Issue #9921, opened 2021-12-01
Make build process robust against missing nodes


astrojuanlu:
**Is your feature request related to a problem? Please describe.**
Projects often pick extensions that introduce new nodes, and at the same time they try to build output other than HTML, perhaps provided by another extension.

A specific case of this is https://github.com/sphinx-doc/sphinx/issues/9919, which uses Sphinx-diagrams (adding a new `diagram` node) combined with sphinxcontrib-confluencebuilder, which doesn't understand the concept of diagrams. Other recent examples include nbsphinx + pdf build.

In all these cases, the build crashes, even if `-W` hasn't been used.

**Describe the solution you'd like**
It would be nice to to produce an empty box or generic node with a warning in these cases, rather than failing the build. At least most of the documentation could be build.

Extra points if the warning says which extension registered the node, so the issue can be opened in the right repository straight away.

**Describe alternatives you've considered**
Naturally, one alternative is not doing anything. I think the current status is problematic, but I haven't weighed the cost of changing it.

This behavior could be made configurable, but I guess it can be treated like any other Sphinx warning.

jdknight:
To give an extension developer's perspective, raising an `NotImplementedError` exception appears to be the "good practice" approach when first diving into Sphinx. When looking at Sphinx's implementation for reference, it is used throughout the internal implementation and replicating methods like this appear to be the "right" way forward:

```
    def unknown_visit(self, node: Node) -> None:
        raise NotImplementedError('Unknown node: ' + node.__class__.__name__)
```

But if `NotImplementedError` generates an exception, a user thinks of this as a generic Sphinx issue and then reports to this issue tracking (which I'm sure that's how #9919 came to light), I can agree that there is an issue here.

Personally, I do not think extension writers (in most cases) should even need to implement an `unknown_visit` hook -- if the sole purpose is to report that it is an unsupported node. From what I can tell, [docutils already does this](https://repo.or.cz/docutils.git/blob/d169015ee0f412cffd69b33654d8a119d99bc0f3:/docutils/nodes.py#l2048) (even in [v0.14](https://repo.or.cz/docutils.git/blob/53716a13b48128af6045139d3cd2909f61e7ed8e:/docutils/nodes.py#l1897)). Would it be possible to just drop all existing Sphinx-managed translators from defining their own `unknown_visit` call? This could help prevent future extensions created from replicating these calls.

[...]

PR #9943, opened 2021-12-05
drop translator-specific unknown_visit calls

jdknight:
### Feature or Bugfix
- Refactoring

### Purpose

- Removes the need for various `unknown_visit` calls which are always supported by docutils.
- Generates unknown node warning messages for all Sphinx translators over an exception.

### Detail

Removes the need for various translators from raising a `NotImplementedError` exception when missing support for a specific node type. docutils will already raise \[1\]\[2\] a `NotImplementedError` exception for these cases. This help reduce the implementation inside Sphinx as well as prevents the possible undesired replication of unknown-node handling with third-party extensions \[3\].

In most cases, generating a warning message for an unsupported node type can be preferred. Providing an indication that a node is not supported can be easier for a user of Sphinx to understand a limitation of a builder over a generic "not implemented" exception. This commit takes the logging call which is already used by `texinfo` and applies it to the `SphinxTranslator` base class -- which any Sphinx translator implementation can use.

\[1\]: https://repo.or.cz/docutils.git/blob/d169015ee0f412cffd69b33654d8a119d99bc0f3:/docutils/nodes.py#l2048
\[2\]: https://repo.or.cz/docutils.git/blob/53716a13b48128af6045139d3cd2909f61e7ed8e:/docutils/nodes.py#l1897
\[3\]: https://github.com/sphinx-doc/sphinx/issues/9921

---

This provides an initial implementation based off the discussion outlined in https://github.com/sphinx-doc/sphinx/issues/9921. There are some notables to consider being any attempt to merge this in:

1) While `texinfo` generates a warning for unknown nodes, there may be some test cases for builders like `html` where using a warning message may not be desired. If tests are not configured to stop a warning scenario (assumed to be the case), these changes will lose the chance at catching unknown nodes events (if this is common).
2) docutils' unknown visit handling takes into consideration of a `strict_visitor` option and `optional` option. It will most likely be better to adjust the proposed `unknown_visit` method in `SphinxTranslator` to take these options into consideration (i.e. instead of generating a warning, call the docutils `unknown_visit` call). Although if either of these options are used to skip the handling of warning generation, note that it may be good to reinstate the always-warn implementation may into `texinfo`.

### Relates
- https://github.com/sphinx-doc/sphinx/issues/9921


jdknight:
Note: I can correct the linter errors if the overall pull request is desired; or can make the adjustments if other changes are needed.

tk0miya:
+1: Good idea. Let's go forward!

tk0miya:
Merged. Thank you for your contribution.

Issue #9926, opened 2021-12-02
LaTeX: too large vertical whitespace above .. note role


marxin:
### Describe the bug

Consider:

```rst
Code block:

.. code-block::

  gcc a.c
  ./a.out

And then there's a note:

.. note::

   I am a note!

And then there's something else.
```

I think the margin around `note` is quite big. What about making it smaller?

![Screenshot from 2021-12-02 11-19-51](https://user-images.githubusercontent.com/2658545/144403261-b4ae6d6b-0ba1-44b1-8b22-381bfe30bbb5.png)

### How to Reproduce

Build the snippet.

### Your project

Build the snippet

### OS

Linux

### Python version

3.8

### Sphinx version

4.3.0

marxin:
And it's even bigger what `note` follows an `.. option::` directive:

![Screenshot from 2021-12-02 11-21-12](https://user-images.githubusercontent.com/2658545/144403536-ec82cf62-7cfe-4d48-8164-e9605a71c02d.png)

jakobandersen:
I guess this would be theme-dependent. Which one are you using?

jakobandersen:
> I guess this would be theme-dependent. Which one are you using?

Doh, didn't see "Latex" in the title.

marxin:
> Doh, didn't see "Latex" in the title.

Yes, it's Latex :)

marxin:
@jfbu Can you help us with that, please?

jfbu:
About the extra spacing when the note is first thing in an option description I noticed there is some strange code in Sphinx latex writer: 
https://github.com/sphinx-doc/sphinx/blob/edd14783f3cc6222066fd63efbe28c2728617e18/sphinx/writers/latex.py#L759-L763

The `~` which ends up in LaTeX appears to be the culprit. I will have to check what motivated its insertion (this old legacy code is at least "14 years ago" https://github.com/sphinx-doc/sphinx/blame/c0ef6a92656cbada6702d0102b19d58c881763ed/sphinx/writers/latex.py#L816-L817).

This being said the note is rendered in LaTeX via the sphinxlightbox environment which does not really have hooks: https://github.com/sphinx-doc/sphinx/blob/3ae124f8ac038a3ca79aa266d4558fd85c6e6032/sphinx/texinputs/sphinxlatexadmonitions.sty#L34-L57

[...]

PR #9945, opened 2021-12-05
Fix #9944 (LaTeX writer visit_desc_content())

### LINKED ISSUE #9944: LaTeX: extra vertical whitespace for some nested declarations
### Describe the bug

Extra vertical whitespace in pdf output:

Capture d’écran 2021-12-05 à 21 43 47

### How to Reproduce

```
            -msse
            -msse2

   .. note::

      A note in the option description

.. c:function:: PyObject *PyType_GenericAlloc(PyTypeObject *type, Py_ssize_t nitems)

   .. note::

      Extra vertical space also in this case if note is first item

.. cpp:class:: Data

   .. cpp:union:: @data

      .. cpp:var:: int a

      .. cpp:var:: double b

Explicit ref: :cpp:var:Data::@data::a. Short-hand ref: :cpp:var:Data::a.
```

### Additional context

Relates #9926

(I edited this as the output from the new issue template is not very useful, and input was a bit painful too)

jfbu:
Relates #9926 

With this, the following

```
.. option:: -mmmx
            -msse
            -msse2

   .. note::

      A note in the option description

.. c:function:: PyObject *PyType_GenericAlloc(PyTypeObject *type, Py_ssize_t nitems)

   .. note::

      Extra vertical space also in this case if note is first item

.. cpp:class:: Data

   .. cpp:union:: @data

      .. cpp:var:: int a

      .. cpp:var:: double b

Explicit ref: :cpp:var:Data::@data::a. Short-hand ref: :cpp:var:Data::a.
```

produces

Capture d’écran 2021-12-05 à 21 44 56

Comparing to the output shown in issue #9944, it only fixes the vertical space for the option with note and nested declaration case, not for the note at start of function description.

The PR removes from LaTeX markup  an  `~`  which was added 14 years ago at 3761223d85b16. The `~` causes LaTeX to enter into paragraph mode and then this causes vertical space if some "display block" type of entity (such as a note) follows).

It *seems* not to be needed anymore. I could not find an example where the `~` to avoid empty contents would avoid a "formatting bug" and don't know where to look.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Bugs fixed
 
 * #9917: C and C++, parse fundamental types no matter the order of simple type
   specifiers.
+* #9944: LaTeX: extra vertical whitespace for some nested declarations
 
 Testing
 --------
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -757,9 +757,7 @@ class LaTeXTranslator(SphinxTranslator):
         self._depart_signature_line(node)
 
     def visit_desc_content(self, node: Element) -> None:
-        if node.children and not isinstance(node.children[0], nodes.paragraph):
-            # avoid empty desc environment which causes a formatting bug
-            self.body.append('~')
+        pass
 
     def depart_desc_content(self, node: Element) -> None:
         pass


jfbu:
(I had completely forgotten to check the unit tests, but in fact I see from succesful testing that none existed for a case where latex writer would output the `~`  or at least none tested for its presence in output...)

jfbu:
Sorry this got closed from me renaming my local branch via the web interface at my fork. Will open a new PR with the renamed branch.

PR #9946, opened 2021-12-06
Fix #9944 (LaTeX writer visit_desc_content())

### LINKED ISSUE #9944: LaTeX: extra vertical whitespace for some nested declarations

jfbu:
Relates #9926 

With this, the following

```
.. option:: -mmmx
            -msse
            -msse2

   .. note::

      A note in the option description

.. c:function:: PyObject *PyType_GenericAlloc(PyTypeObject *type, Py_ssize_t nitems)

   .. note::

      Extra vertical space also in this case if note is first item

.. cpp:class:: Data

   .. cpp:union:: @data

      .. cpp:var:: int a

      .. cpp:var:: double b

Explicit ref: :cpp:var:Data::@data::a. Short-hand ref: :cpp:var:Data::a.
```

produces

Capture d’écran 2021-12-05 à 21 44 56

Comparing to the output shown in issue #9944, it only fixes the vertical space for the option with note and nested declaration case, not for the note at start of function description.

This PR removes from LaTeX markup  an  `~`  which was added (14 years ago) at 3761223d85b16 to fix some "bad formatting issue in case of empty description".

(*edit after further investigation*) the `~` was needed at least in nested cases like the  `cpp` example above. Without it and at Sphinx `<1.5` the `cpp` example above gives mangled pdf output.

Capture d’écran 2021-12-06 à 09 52 49

But since `1.5`, the LaTeX mark-up adds `\pysigstartmultiline/\pysigstopmultiline` pairs.  The `\pysigstopmultiline` appears to fix the same issue but without the extra spacing. So it seems `~` should not be inserted, at least in such examples.

*note*: I did not need to update tests as apparently no unit test existed to check for the `~`  insertion in latex object descriptions in some cases.

*this PR replaces #9945 which was accidentally closed from renaming the branch at my fork underlying it*

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -36,6 +36,8 @@ Bugs fixed
 * #9878: mathjax: MathJax configuration is placed after loading MathJax itself
 * #9857: Generated RFC links use outdated base url
 * #9909: HTML, prevent line-wrapping in literal text.
+* #9925: LaTeX: prohibit also with ``'xelatex'`` line splitting at dashes of
+  inline and parsed literals
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -38,6 +38,7 @@ Bugs fixed
 * #9909: HTML, prevent line-wrapping in literal text.
 * #9925: LaTeX: prohibit also with ``'xelatex'`` line splitting at dashes of
   inline and parsed literals
+* #9944: LaTeX: extra vertical whitespace for some nested declarations
 
 Testing
 --------
--- a/sphinx/texinputs/sphinxlatexliterals.sty
+++ b/sphinx/texinputs/sphinxlatexliterals.sty
@@ -1,7 +1,7 @@
 %% LITERAL BLOCKS
 %
 % change this info string if making any custom modification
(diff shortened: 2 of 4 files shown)


tk0miya (approved):
LGTM. But it's better to merge this into the 4.x branch instead.

jfbu:
Thanks for review @tk0miya I have rebased on 4.x and will merge after testing completes

PR #9948, opened 2021-12-06
doc: Improve autodoc extension example

joukewitteveen:
Subject: Minor fixes to the example code of the autodoc extension tutorial (follow-up on #9928)

### Feature or Bugfix

- Feature (minor change of example code)

### Purpose

Enumerations can have aliases, which should be documented as well. The current example code can easily be changed to include this improved functionality.

### Relates
- #9928


tk0miya:
Thanks!

PR #9956, opened 2021-12-10
address some code quality issues

danieleades:
a bunch of small, non-breaking, code quality improvements

### DIFF
--- a/sphinx/cmd/quickstart.py
+++ b/sphinx/cmd/quickstart.py
@@ -175,10 +175,7 @@ class QuickstartRenderer(SphinxRenderer):
               It will be removed in the future without deprecation period.
         """
         template = path.join(self.templatedir, path.basename(template_name))
-        if self.templatedir and path.exists(template):
-            return True
-        else:
-            return False
+        return bool(self.templatedir) and path.exists(template)
 
     def render(self, template_name: str, context: Dict) -> str:
         if self._has_custom_template(template_name):
--- a/sphinx/ext/apidoc.py
+++ b/sphinx/ext/apidoc.py
@@ -181,11 +181,7 @@ def is_skipped_package(dirname: str, opts: Any, excludes: List[str] = []) -> boo
         return True
 
     # Check there is some showable module inside package
-    if all(is_excluded(path.join(dirname, f), excludes) for f in files):
-        # all submodules are excluded
-        return True
-    else:
-        return False
+    return all(is_excluded(path.join(dirname, f), excludes) for f in files)
 
 
 def is_skipped_module(filename: str, opts: Any, excludes: List[str]) -> bool:
@@ -193,11 +189,8 @@ def is_skipped_module(filename: str, opts: Any, excludes: List[str]) -> bool:
     if not path.exists(filename):
         # skip if the file doesn't exist
         return True
-    elif path.basename(filename).startswith('_') and not opts.includeprivate:
-        # skip if the module has a "private" name
-        return True
-    else:
-        return False
+    # skip if the module has a "private" name
+    return path.basename(filename).startswith('_') and not opts.includeprivate
 
 
 def walk(rootpath: str, excludes: List[str], opts: Any
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -2362,10 +2362,7 @@ class SlotsMixin(DataDocumenterMixinBase):
         """Check the subject is an attribute in __slots__."""
         try:
             __slots__ = inspect.getslots(self.parent)
-            if __slots__ and self.objpath[-1] in __slots__:
-                return True
-            else:
-                return False
+            return bool(__slots__) and self.objpath[-1] in __slots__
         except (ValueError, TypeError):
             return False
 
@@ -2402,10 +2399,7 @@ class SlotsMixin(DataDocumenterMixinBase):
     def _datadescriptor(self) -> bool:
         warnings.warn('AttributeDocumenter._datadescriptor() is deprecated.',
                       RemovedInSphinx60Warning)
-        if self.object is SLOTSATTR:
-            return True
-        else:
-            return False
+        return self.object is SLOTSATTR
 
 
 class RuntimeInstanceAttributeMixin(DataDocumenterMixinBase):
@@ -2427,10 +2421,7 @@ class RuntimeInstanceAttributeMixin(DataDocumenterMixinBase):
         # An instance variable defined in __init__().
         if self.get_attribute_comment(parent, self.objpath[-1]):  # type: ignore
             return True
-        elif self.is_runtime_instance_attribute_not_commented(parent):
-            return True
-        else:
-            return False
+        return self.is_runtime_instance_attribute_not_commented(parent)
 
     def is_runtime_instance_attribute_not_commented(self, parent: Any) -> bool:
         """Check the subject is an attribute defined in __init__() without comment."""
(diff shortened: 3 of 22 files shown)


sphinx/ext/viewcode.py
```diff
@@ -68,10 +68,7 @@ def is_supported_builder(builder: Builder) -> bool:
         return False
     elif builder.name == 'singlehtml':
         return False
-    elif builder.name.startswith('epub') and not builder.config.viewcode_enable_epub:
-        return False
-    else:
-        return True
+    return not (builder.name.startswith('epub') and not builder.config.viewcode_enable_epub)
```
tk0miya: Indeed, this is equivalent to the current code. But this changes the meaning of the code, I feel. I think it's not good to expand consecutive else-if block to such code.

tk0miya (changes requested):
LGTM with nits.

tk0miya:
It seems mypy emits warnings. Could you check them, please?

sphinx/ext/viewcode.py
danieleades: done

danieleades:
> It seems mypy emits warnings. Could you check them, please?

mypy issues have been addressed, i believe. would you mind retriggering the CI for me @tk0miya?

danieleades:
@tk0miya i *think* this should be good to merge now (I don't think this one is too controversial!)

sphinx/cmd/quickstart.py
```diff
@@ -175,10 +175,7 @@ def _has_custom_template(self, template_name: str) -> bool:
               It will be removed in the future without deprecation period.
         """
         template = path.join(self.templatedir, path.basename(template_name))
-        if self.templatedir and path.exists(template):
-            return True
-        else:
-            return False
+        return bool(self.templatedir) and path.exists(template)
```
tk0miya, line 178: Why do you call `bool()` here?

sphinx/ext/apidoc.py
```diff
@@ -181,23 +181,16 @@ def is_skipped_package(dirname: str, opts: Any, excludes: List[str] = []) -> boo
...
     if not path.exists(filename):
         # skip if the file doesn't exist
         return True
-    elif path.basename(filename).startswith('_') and not opts.includeprivate:
-        # skip if the module has a "private" name
-        return True
-    else:
-        return False
+    # skip if the module has a "private" name
+    return path.basename(filename).startswith('_') and not opts.includeprivate
```
tk0miya, line 193: I feel this refactoring divides this if-blocks to an early return condition (`path.exists()`) and the main condition.  Do you think that the main condition of "is_skipped_module()` is to check the filename is private or not? IMO, it's one of the conditions. So they should not be separated.

sphinx/ext/autodoc/__init__.py
```diff
@@ -2427,10 +2421,7 @@ def is_runtime_instance_attribute(self, parent: Any) -> bool:
         # An instance variable defined in __init__().
         if self.get_attribute_comment(parent, self.objpath[-1]):  # type: ignore
             return True
-        elif self.is_runtime_instance_attribute_not_commented(parent):
-            return True
-        else:
-            return False
+        return self.is_runtime_instance_attribute_not_commented(parent)
```
tk0miya, line 2424: This is also.

sphinx/ext/autodoc/__init__.py
```diff
@@ -2571,10 +2559,7 @@ def can_document_member(cls, member: Any, membername: str, isattr: bool, parent:
             return False
         elif inspect.isattributedescriptor(member):
             return True
-        elif not inspect.isroutine(member) and not isinstance(member, type):
-            return True
-        else:
-            return False
+        return not inspect.isroutine(member) and not isinstance(member, type)
```
tk0miya, line 2562: This is also.

sphinx/ext/autodoc/type_comment.py
```diff
@@ -26,13 +26,10 @@ def not_suppressed(argtypes: List[ast.AST] = []) -> bool:
     """Check given *argtypes* is suppressed type_comment or not."""
     if len(argtypes) == 0:  # no argtypees
         return False
-    elif len(argtypes) == 1 and ast_unparse(argtypes[0]) == "...":  # suppressed
+    return not (len(argtypes) == 1 and ast_unparse(argtypes[0]) == "...")  # suppressed
         # Note: To support multiple versions of python, this uses ``ast_unparse()`` for
         # comparison with Ellipsis.  Since 3.8, ast.Constant has been used to represent
         # Ellipsis node instead of ast.Ellipsis.
```
tk0miya, line 32: Could you move this comment to the above of the return statement?

sphinx/ext/autodoc/type_comment.py
```diff
@@ -26,13 +26,10 @@ def not_suppressed(argtypes: List[ast.AST] = []) -> bool:
     """Check given *argtypes* is suppressed type_comment or not."""
     if len(argtypes) == 0:  # no argtypees
         return False
-    elif len(argtypes) == 1 and ast_unparse(argtypes[0]) == "...":  # suppressed
+    return not (len(argtypes) == 1 and ast_unparse(argtypes[0]) == "...")  # suppressed
```
tk0miya, line 29: The comment on the tail is no longer needed.

sphinx/util/inspect.py
```diff
@@ -341,24 +335,18 @@ def isattributedescriptor(obj: Any) -> bool:
                                     WrapperDescriptorType)):
             # attribute must not be a method descriptor
             return False
-        elif type(unwrapped).__name__ == "instancemethod":
-            # attribute must not be an instancemethod (C-API)
-            return False
-        else:
-            return True
+        # attribute must not be an instancemethod (C-API)
+        return not type(unwrapped).__name__ == "instancemethod"
```
tk0miya, line 339: This is also.

sphinx/util/inspect.py
```diff
@@ -393,28 +381,19 @@ def iswrappedcoroutine(obj: Any) -> bool:
             # staticmethod, classmethod and partial method are not a wrapped coroutine-function
             # Note: Since 3.10, staticmethod and classmethod becomes a kind of wrappers
             return False
-        elif hasattr(obj, '__wrapped__'):
-            return True
-        else:
-            return False
+        return hasattr(obj, '__wrapped__')
```
tk0miya, line 384: This is also.

sphinx/util/inspect.py
```diff
@@ -435,11 +414,8 @@ def isgenericalias(obj: Any) -> bool:
     elif (hasattr(types, 'GenericAlias') and  # only for py39+
           isinstance(obj, types.GenericAlias)):  # type: ignore
         return True
-    elif (hasattr(typing, '_SpecialGenericAlias') and  # for py39+
-            isinstance(obj, typing._SpecialGenericAlias)):  # type: ignore
-        return True
-    else:
-        return False
+    return (hasattr(typing, '_SpecialGenericAlias') and  # for py39+
+            isinstance(obj, typing._SpecialGenericAlias))  # type: ignore
```
tk0miya, line 418: This is also.

tk0miya (changes requested):

sphinx/cmd/quickstart.py
danieleades: mypy returns a false positive here without explicitly coercing to a bool. there might be a better way to address this. will investigate

sphinx/ext/apidoc.py
danieleades: hmm. If you wanted to return from within the if-else block in these sorts of cases, i'd be more inclined to go with a construction like the following:

```python
if (
    # skip if the file doesn't exist
    not path.exists(filename) or
    # skip if the module has a "private" name
    path.basename(filename).startswith('_') and not opts.includeprivate
):
    return True
else:
    return False
```

AA-Turner:
Closed in favour of #10516.

A

PR #9966, opened 2021-12-13
refactor: Remove RemovedInSphinx50Warning (partially)

tk0miya:
### Feature or Bugfix
- Refactoring


sphinx/domains/c.py
```diff
@@ -3256,13 +3255,6 @@ def handle_signature(self, sig: str, signode: TextElement) -> ASTDeclaration:
                 except DefinitionError:
                     raise eOrig
                 self.object_type = ast.objectType  # type: ignore
-                if self.env.config['c_warn_on_allowed_pre_v3']:
```
tk0miya: @jakobandersen It seems `c_warn_on_allowed_pre_v3` will no longer be used in v5.0. Is it okay to remove it?

sphinx/writers/latex.py
```diff
@@ -444,14 +410,6 @@ def popbody(self) -> List[str]:
         self.body = self.bodystack.pop()
         return body
 
-    def format_docclass(self, docclass: str) -> str:
-        """Prepends prefix to sphinx document classes"""
-        warnings.warn('LaTeXWriter.format_docclass() is deprecated.',
-                      RemovedInSphinx50Warning, stacklevel=2)
-        if docclass in self.docclasses:
```
tk0miya, line 451: `self.docclass` is no longer used. So it should be marked as deprecated. I'll do it on another PR.

sphinx/domains/c.py
jakobandersen: Hmm, I think we should keep it for another version.

tk0miya: Are you saying to keep these warnings even after v5.0? If so, I'll revert the removals from this PR.

If not, I'm not sure why it's needed. IMO, no reason to keep it because it will not be referred at all (after this PR merged).
```
$ grep -r c_warn_on_allowed_pre_v3 sphinx
sphinx/domains/c.py:    app.add_config_value("c_warn_on_allowed_pre_v3", True, 'env')
```

jakobandersen: > Are you saying to keep these warnings even after v5.0? If so, I'll revert the removals from this PR.

Yes, please revert it. The version should be bumped to 6 for now. Should I submit a PR, or can you easily make the modification?

> > If not, I'm not sure why it's needed. IMO, no reason to keep it because it will not be referred at all (after this PR merged).

It has been a while since I looked at it, but basically it is a warning for users to detect if they rely on old syntax that will be deprecated at some point. It was introduced because the Linux kernel documentation needs to be build on old distributions where the Sphinx version is very, very old, so it will probably take a couple of years before the old syntax can be removed along with this warning.

tk0miya: Okay, just updated.

jakobandersen: Great, thanks!

PR #9994, opened 2021-12-20
[proposal] the "ja" message of _("see %s") and _("see also %s") in sphinx.po

KaKkouo:
Subject: Adjust the message so that it is natural as Japanese.

### Feature or Bugfix
- Refactoring

### Purpose
Adjust the message so that it is natural as Japanese.

### Detail
_("see %s") of the "ja" sphinx.po

- current status.: "%sを参照"
- proposal: "「%s」を参照"

_("see also %s") of the "ja" sphinx.po

- current status: "%sも参照"
- proposal:  "「%s」も参照"

### Relates
none.

### other
You can use locale_dirs to deal with this, but I suggested it because I thought it was a change that many Japanese would accept.

### DIFF
--- a/sphinx/locale/ja/LC_MESSAGES/sphinx.po
+++ b/sphinx/locale/ja/LC_MESSAGES/sphinx.po
@@ -2371,12 +2371,12 @@ msgstr "ドキュメントはどの toctree にも含まれていません"
 #: sphinx/environment/adapters/indexentries.py:78
 #, python-format
 msgid "see %s"
-msgstr "%sを参照"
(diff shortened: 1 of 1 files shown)


tk0miya:
Thank you for your contribution. But we've used transifex.com to manage translation messages. So could you join our team in transifex.com and update translations, please?
https://www.transifex.com/sphinx-doc/sphinx-1/dashboard/

KaKkouo:
Thanks for the info. I will register as a user first and take my time learning how to use it.

PR #9998, opened 2021-12-20
Close #9993: std domain: Allow to refer an inline target via ref role

### LINKED ISSUE #9993: Sphinx: In-line cross referencing target should not need explicit title
Yes, in-line targets such as  `` _`this one` `` cannot be easily referred using role based `:ref:` cross-referencing, which is needed across RST files of the same document.  It needs an explicit title.  For example, ``:ref:`this one` `` does not insert the target text as the title.  It needs an explicit tltle like this `` :ref:`That one<this one>` ``.  Without explicit link, there is a warning 
```
WARNING: Failed to create a cross reference. A title or caption not found: this one
```

``:ref:`this one` `` must insert the text `this one` and also link it to the target.  This will avoid duplicate typing of the same text, when avoidable.

Duplicating the target text as the title. `` :ref:`this one<this one>` ``

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #9993

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -30,6 +30,8 @@ Features added
   checking in matched documents.
 * #9793: sphinx-build: Allow to use the parallel build feature in macOS on macOS
   and Python3.8+
+* #9993: std domain: Allow to refer an inline target (ex. ``_`target name```)
+  via :rst:role:`ref` role
 * #9391: texinfo: improve variable in ``samp`` role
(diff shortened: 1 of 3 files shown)


stephenfin:
FYI, this introduced a rather significant breaking change. I've opened #10177 to discuss this and posted a revert at #10178.

PR #10001, opened 2021-12-21
Add option_detailed_parse config value.

### LINKED ISSUE #9965: Allow a variable part in ..option directives
### Describe the bug

When I refer to options, I commonly use
```rst
:option:`-foo`:samp:`={style}`
```

where `style` is a variable part as defined here:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html?highlight=samp#role-samp

I would also appreciate that support in `option` directive:

```rst
.. option:: -foo={style}
```

so that the `style` option uses italic font in an output format.
@jakobandersen: What do you think?

### How to Reproduce

Build the snippet.

### Expected behavior

_No response_

### Your project

Build the snippet

### Screenshots

_No response_

### OS

Linux

### Python version

3.8

### Sphinx version

4.3.0

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

marxin:
Support parsing of "variable part" for option directives similarly
to samp roles.

Fixes: #9965.


marxin:
I think the pull request is not correct. The difference I see based on the new config value:

```
Args:  directory
samp_role.parse: [<#text: ' directory'>]
desc_addname: <desc_addname classes="sig-prename descclassname" xml:space="preserve"> directory</desc_addname>
```

@tk0miya Can you please help me?

marxin:
> @tk0miya Can you please help me?

I think I understand it now and hope I have a proper solution that fixes cases where whitespaces were lost in HTML output.

jakobandersen:
> > @tk0miya Can you please help me?
> 
> I think I understand it now and hope I have a proper solution that fixes cases where whitespaces were lost in HTML output.

In the ``allowable_client`` example, what is the HTML around the whitespace?
For future-proofing it might be good to render a space explicitly by a ``addnodes.desc_sig_space()``.

marxin:
> In the `allowable_client` example, what is the HTML around the whitespace?

Looks like this:

```html
<span id="cmdoption-git-commit-allowable-client"></span><span class="sig-name descname"><span class="pre">-allowable_client</span></span><span class="sig-prename descclassname"> </span><em><span class="pre">client_name</span></em><a class="headerlink" href="#cmdoption-git-commit-allowable_client" title="Permalink to this definition">¶</a></dt>
```

> For future-proofing it might be good to render a space explicitly by a `addnodes.desc_sig_space()`.

Do you render this before and after a "variable" part?

sphinx/domains/std.py
```diff
@@ -202,7 +204,16 @@ def handle_signature(self, sig: str, signode: desc_signature) -> str:
             if count:
                 signode += addnodes.desc_addname(', ', ', ')
             signode += addnodes.desc_name(optname, optname)
-            signode += addnodes.desc_addname(args, args)
+            if self.env.config.option_parse_variable_part:
+                for part in samp_role.parse(args):
+                    if isinstance(part, nodes.Text):
+                        # in order to preserve whitespaces, call desc_addname
+                        text = part.astext()
+                        signode += addnodes.desc_addname(text, text)
```
jakobandersen: I guess here one could do something like
```py
if len(text) != 0 and len(text.strip()) == 0:
   signode += addnodes.desc_sig_space()
else:
   signode += addnodes.desc_addname(text, text)
```
but it is more a theoretical exercise at this point I guess.

jakobandersen:
> > In the `allowable_client` example, what is the HTML around the whitespace?
> 
> Looks like this:
> 
> ```
> <span id="cmdoption-git-commit-allowable-client"></span><span class="sig-name descname"><span class="pre">-allowable_client</span></span><span class="sig-prename descclassname"> </span><em><span class="pre">client_name</span></em><a class="headerlink" href="#cmdoption-git-commit-allowable_client" title="Permalink to this definition">¶</a></dt>
> ```
> 
> > For future-proofing it might be good to render a space explicitly by a `addnodes.desc_sig_space()`.
> 
> Do you render this before and after a "variable" part?

I'm not completely sure what you mean, but the HTML that in your quote is ``<span class="sig-prename descclassname"> </span>`` would instead be generated by the ``addnodes.desc_sig_space()``, see inline code comment.

marxin:
> I'm not completely sure what you mean, but the HTML that in your quote is `<span class="sig-prename descclassname"> </span>` would instead be generated by the `addnodes.desc_sig_space()`, see inline code comment.

All right, I've just done that.

sphinx/domains/std.py
```diff
@@ -202,7 +204,19 @@ def handle_signature(self, sig: str, signode: desc_signature) -> str:
...
-            signode += addnodes.desc_addname(args, args)
+            if self.env.config.option_parse_variable_part:
+                for part in samp_role.parse(args):
+                    if isinstance(part, nodes.Text):
+                        text = part.astext()
+                        if not text.strip():
+                            signode += addnodes.desc_sig_space()
+                        else:
+                            # in order to preserve whitespaces, call desc_addname
+                            signode += addnodes.desc_addname(text, text)
```
jakobandersen: Hmm, is ``desc_addname`` appropriate? In the "normal" programming language domains it is used for additional name parts, i.e., qualification of the main declared entity. But here it is not really for typesetting any name part.
Now that we are anyway changing this: to be conservative I suggest using the original ``nodes.Text`` until something a more specific markup is deemed appropriate.

marxin: Hm, then I would need to find a way how to preserve whitespaces for a `nodes.Text` node.
E.g. for:

```
.. option:: -o directory, --object-directory directory, --object-file file
```

parts are: `_directory`, `_directory`, `_file`, where `_` is a space. If I don't do it, I would end up with:
Screenshot from 2022-03-23 10-36-23

marxin:
@tk0miya May I please ping this pull request? Thanks.

doc/usage/configuration.rst
```diff
@@ -700,6 +700,14 @@ General configuration
 
     .. versionadded:: 3.0
 
+.. confval:: option_parse_variable_part
+
+   Default is ``False``.
+   Similarly to ``samp`` role, support parsing of "variable part" of a literal text.
+   For example, in ``..option: -foption={TYPE}``,
```
jakobandersen:
```suggestion
   For example, in ``.. option:: -foption={TYPE}``,
```

doc/usage/configuration.rst
```diff
@@ -700,6 +700,14 @@ General configuration
...
+   the part ``TYPE`` would be emphasized.
+
+   .. versionadded:: 4.5
```
jakobandersen:
```suggestion
   .. versionadded:: 5.0
```

sphinx/domains/std.py
jakobandersen: How about these changes: https://github.com/jakobandersen/sphinx/tree/marxin-option_parse_variable_part

jakobandersen:
I suggest making the conf var name more generic, e.g., ``option_detailed_parse``, and then change the docs to describe that it in general makes Sphinx parse the options more in detail to provide better markup, which includes the parsing of ``{stuff}`` as if by the ``:samp:`` role to signify variables, and that the curly braces can be escaped, etc.

marxin:
Made the suggested change to the config value name. Feel free to suggest wording improvements for the config value.

jakobandersen:
> Made the suggested change to the config value name. Feel free to suggest wording improvements for the config value.

With the name change I suggest pulling in some variation of the changes I made in https://github.com/jakobandersen/sphinx/tree/marxin-option_parse_variable_part.

PR #10005, opened 2021-12-22
Extend documentation for xref of options.

### LINKED ISSUE #10002: Support :option: reference defined in a program scope
@jakobandersen
I think it's impossible doing such a reference for:

```rst
.. program:: make

.. option:: --verbose

Use verbose mode.

.. program:: None

Use :option:`--verbose` mode.
```

I would like to use something like:
```rst
:option:`make.--verbose`
```

if I'm out of the `make` program scope.
Thank you.

marxin:
Explain better xrefs of options that live in context of a program.
Fixes: #10002.

### DIFF
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -1739,6 +1739,8 @@ There is a set of directives allowing documenting command-line programs:
 
    then ``:option:`rm -r``` would refer to the first option, while
    ``:option:`svn -r``` would refer to the second one.
+   Note using ``:option:`-r <rm -r>``` would refer the first option, while
+   display text would be only ``-r``.
 
    If ``None`` is passed to the argument, the directive will reset the
    current program name.


doc/usage/restructuredtext/domains.rst
```diff
@@ -1739,6 +1739,8 @@ There is a set of directives allowing documenting command-line programs:
 
    then ``:option:`rm -r``` would refer to the first option, while
    ``:option:`svn -r``` would refer to the second one.
+   Note using ``:option:`-r <rm -r>``` would refer the first option, while
+   display text would be only ``-r``.
```
tk0miya, line 1743: This notation is not only for the `:option:` role, but for other cross-reference roles.
And it has been already described at the top of this page:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax (the first item of the bullet list)

So -0 for adding this.

marxin:
You are right, both things are documented elsewhere.

Issue #9595, opened 2021-08-31
Make the distutils/setuptools integration an opt-in for projects


pradyunsg:
**Is your feature request related to a problem? Please describe.**

Sphinx currently injects a "magical" `setup.py build_sphinx` command into setuptools-based projects. While this is likely be useful for some user workflows, recently this has been a major source of confusion in the Python ecosystem, due to some folks who are repackaging a substantial number of open source Python packages.

These repackagers have assumed that all Python projects that use setuptools and Sphinx also use this integration -- which is not true. Most users are not aware of this integration and do not use it or support it. By making projects explicitly opt-in to this integration, it would help avoid such confusion and problems due to repackagers trying to build documentation using this integration and that not working -- since it's not a way that the specific project supports for building documentation (eg: if they use any non-setuptools build backend).

Further, this also puts setuptools in a uniquely priviledged position with Sphinx -- something that Python packaging tooling has been actively working to avoid and move away from (see pyproject.toml based builds, introduced in PEP 517 / PEP 518).

**Describe the solution you'd like**

Move the setuptools integration into a separate package on PyPI. We can also add it as an optional dependency of Sphinx (via the extras mechanism), so that users who want this can use `sphinx[setuptools]` in their requirements.txt or `pip install` invocations. 

**Describe alternatives you've considered**

Not doing anything, which... works. 🤷🏽‍♂️

I think this is suboptimal though -- see https://github.com/search?q=build_sphinx+is%3Aissue&type=issues -- there are *many* issues filed on the premise that this integration means that projects should actively support this, which is a generally harmful suggestion to make IMO. These assumptions break down for non-setuptools-based projects as well.

**Additional context**

I guess, it's worth noting that a certain individual has been filed 100s of issues against various Python projects regarding this exact issue -- demanding that the entire ecosystem support this integration as a way to build their documentation (see the search above).

astrojuanlu:
Recent conversation on the topic (that got derailed, anyway) on one of such issues https://github.com/sphinx-doc/sphinx/issues/9517#issuecomment-890477274 . I should have opened a new issue about it, thanks @pradyunsg. +1 from me.

hukkin:
+1 to this.

An alternative low effort solution would be to simply deprecate and eventually remove this feature entirely in favor of a direct `sphinx-build` invocation (possibly via `tox`). My understanding is that `setuptools` is actively trying to remove all `python setup.py <cmd>` invocations, so IMO would make sense to do the same thing here.

[...]

PR #10007, opened 2021-12-22
Replace pkg_resources with importlib.metadata

takluyver:
### Feature or Bugfix

- Feature

### Purpose

As discussed in #9595, `importlib.metadata` is the modern way to find entry points, replacing `pkg_resources`. It's in the standard library from 3.8, but part of the API I've used is new in 3.10, so the [`importlib_metadata` backport](https://pypi.org/project/importlib-metadata/) is used on older versions.

Concretely, `pkg_resources` scans metadata from all packages at import time, which is generally considered a bad thing. `importlib.metadata` only scans metadata when it's called.

### DIFF
--- a/sphinx/registry.py
+++ b/sphinx/registry.py
@@ -21,6 +21,7 @@ from docutils.nodes import Element, Node, TextElement
 from docutils.parsers import Parser
 from docutils.parsers.rst import Directive
 from docutils.transforms import Transform
+
 try:  # Python < 3.10 (backport)
     from importlib_metadata import entry_points
 except ImportError:
--- a/sphinx/registry.py
+++ b/sphinx/registry.py
@@ -21,7 +21,7 @@ from docutils.nodes import Element, Node, TextElement
 from docutils.parsers import Parser
 from docutils.parsers.rst import Directive
 from docutils.transforms import Transform
-try:
+try:  # Python < 3.10 (backport)
     from importlib_metadata import entry_points
 except ImportError:
     from importlib.metadata import entry_points
--- a/sphinx/registry.py
+++ b/sphinx/registry.py
@@ -147,9 +147,9 @@ class SphinxComponentRegistry:
             return
 
         if name not in self.builders:
-            builder_eps = entry_points(group='sphinx.builders')
+            builder_entry_points = entry_points(group='sphinx.builders')
             try:
-                entry_point = builder_eps[name]
+                entry_point = builder_entry_points[name]
             except KeyError as exc:
                 raise SphinxError(__('Builder name %s not registered or available'
                                      ' through entry point') % name) from exc
--- a/sphinx/registry.py
+++ b/sphinx/registry.py
@@ -21,7 +21,10 @@ from docutils.nodes import Element, Node, TextElement
 from docutils.parsers import Parser
 from docutils.parsers.rst import Directive
 from docutils.transforms import Transform
-from pkg_resources import iter_entry_points
+try:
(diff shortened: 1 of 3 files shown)


takluyver:
The mypy failure appears to be unrelated. Mypy 0.930 was just released today, and if I use this on the 4.x branch, it fails. Mypy 0.921 passes on both branches. I can limit the version range used in `setup.py` if you like, but I'd rather not dig in to unrelated type checking issues.

sphinx/registry.py
```diff
@@ -21,7 +21,11 @@
 from docutils.parsers import Parser
 from docutils.parsers.rst import Directive
 from docutils.transforms import Transform
-from pkg_resources import iter_entry_points
+
+try:  # Python < 3.10 (backport)
```
webknjaz, line 25: FYI the try/except idiom is known to confuse mypy. There's a known bug that hasn't been fixed for years AFAIK. The workaround is to use if/else.

webknjaz (approved):

sphinx/registry.py
takluyver: mypy 0.921 seems to be happy with this - as far as I can tell, the mypy CI failures are due to the recent release of mypy 0.930, and entirely orthogonal to this PR.

sphinx/registry.py
```diff
@@ -143,14 +147,14 @@ def preload_builder(self, app: "Sphinx", name: str) -> None:
             return
 
         if name not in self.builders:
-            entry_points = iter_entry_points('sphinx.builders', name)
+            builder_eps = entry_points(group='sphinx.builders')
```
tk0miya: What does `eps` means? If you mean abbr of "EntryPointS", it would be better to use `builder_entry_points` instead. It's enough short and simple.

tk0miya (commented):
LGTM with nits.

sphinx/registry.py
webknjaz: Yeah, looks like you're right.

sphinx/registry.py
webknjaz: I agree with the sentiment — clarity is important.

tk0miya:
Now I just fixed the warnings from mypy on the HEAD of 4.x branch now. Could you merge it to your branch, please? Then CI will go fine.

sphinx/registry.py
takluyver: Sure, done - and the same for the other file I've modified, of course.

takluyver:
Merged 4.x into this branch, and the CI is indeed OK again. :slightly_smiling_face:

tk0miya:
Thank you for your contribution. Merged!

sphinx/registry.py
```diff
@@ -21,7 +21,11 @@
 from docutils.parsers import Parser
 from docutils.parsers.rst import Directive
 from docutils.transforms import Transform
-from pkg_resources import iter_entry_points
+
+try:  # Python < 3.10 (backport)
+    from importlib_metadata import entry_points
+except ImportError:
+    from importlib.metadata import entry_points
```
jakirkham, line 28: Is there a reason the builtin isn't tried before the backport?

takluyver: Yes: the code is using APIs that were changed in 3.10, so for 3.8 and 3.9 we want to use the backport over the standard library module. :slightly_smiling_face:

jakirkham: Got it. Thanks for the context! 😄

PR #10008, opened 2021-12-23
Remove the build_sphinx command

pradyunsg:
Subject: Drop the distutils `build_sphinx` command

### Feature or Bugfix
- Feature?

### Purpose
- Implements option 2 from https://github.com/sphinx-doc/sphinx/issues/9595#issuecomment-1000069963.

### Detail
- Removes `sphinx.setup_command` and all direct references to it.
- Updates the documentation for the setuptools integration to note why it was removed, and what the alternatives are for users.

### Relates

#9595


tox.ini
```diff
@@ -85,7 +85,7 @@ description =
 extras =
     docs
 commands =
-    python setup.py build_sphinx {posargs}
+    sphinx-build docs/ docs/_build/ -W {posargs}
```
webknjaz, line 88:
```suggestion
    sphinx-build docs/ docs/_build/ -W --keep-going {posargs}
```

webknjaz (approved):

tk0miya:
As I commented in #9595, we have to obey to our application policy. So we should emit a warning first for users to let users know the deprecation.

PR #9874, opened 2021-11-20
Drop jQuery and UnderscoreJS

pradyunsg:
Toward #7405 

- [ ] Port doctools.js
    - [ ] (missing piece right now is make the search highlighting work)
	- [ ] Update tests
- [ ] Port searchtools.js
	- [ ] Update tests
- [ ] Documentation callouts
- [ ] Changelog entry

---

Live-ish list of changes:

- Make all JS function utilities "private", by prefixing an underscore.
- When highlighting content, only highlight content within `[role="main"]`,
  since that is where the search is done.
- Only handle toggle buttons for domainindex.js, in `[role="main"]`.
- Drop unused JS functions: `addContextElements`, `getCurrentURL`, `makeURL`.

PR #10028, opened 2021-12-29
Drop JavaScript Frameworks

### LINKED ISSUE #7405: Remove dependency on jQuery
Most things we use jQuery for can now easily be done using vanilla JS even in IE11.

Removing jQuery therefore does not impact browser coverage significantly and instead leads to a smaller website to serve. As this is be a breaking change I would propose to discuss about implementing this for the 4.x release next year.

If the consensus is to go ahead with removing jQuery I would volunteer for this.

AA-Turner:
xref #7405, #9874 (Closes #7405)

This is another attempt to drop jQuery & underscore.js. Tests pass & everything is converted. I did initially miss @pradyunsg's implementation of this (I started having a go at this [back in 2020](https://github.com/AA-Turner/peps/commit/a333cee79790ea94cf2d7fa59885087de4b715fd) and never got around to upstreaming!), so there are differences between the two, Happy to withdraw mine or work on merging so as not to have competing implementations -- whichever is easier for the maintainers (this is why I've opened the PR as draft).

Main differences between mine and his architecturally are I haven't moved to JavaScript classes, although that might be a good idea -- I didn't consider it at all really.

### Feature or Bugfix

- Refactoring

A


tk0miya:
I'm not sure which is better implementation because I'm not an expert of JS. So I'd like to ask opinions from @TimKam and @pradyunsg .

AA-Turner:
No worries! Happy to wait for opinions

A

AA-Turner:
(to note -- I updated the PR today to remove lingering uses of jQuery throughout the default themes -- I hadn't realised these uses existsed at first. There *should* now be no references to jQuery/underscores in the entire codebase)

TimKam:
Thank you, @AA-Turner. Some broad thoughts:
1. Regarding the two PRs: at first glance, this one looks more mature atm. However @pradyunsg would be an excellent candidate for giving feedback on this PR. 
2. Good that the tests pass, but we should not feel too comfortable about it; most of the JavaScript is not properly tested. I would actually advocate for a two-phase roll-out: first, keep the dependencies and see if there are any bugs in the refactoring. Then, remove the dependencies for good. Before this second stage, we can also ask the folks at Read the Docs for advice: after all, this can have a downstream effect on many themes. Do you agree @tk0miya, @AA-Turner ?

AA-Turner:
> I would actually advocate for a two-phase roll-out: first, keep the dependencies and see if there are any bugs in the refactoring.

This seems a good idea. 

I guess to the point that @pradyunsg made (can't find the thread, sorry) that a large number of Sphinx extensions rely on jQuery (& underscore.js to a lesser extent) implicitly, perhaps we should also put a fairly obvious deprecation warning in the change log, with instructions on how themes can add jQuery back in.

I don't believe there are any good ways to either detect or emit deprecation warnings for jQuery &c, so documentation-based deprecation is probably the best shot.

A

tk0miya:
+1 for the the two-phase roll-out. It's reasonable :-)

AA-Turner:
Ok - I've just reverted the removals of jQuery/underscores and their tests etc.

@tk0miya - Given this is a more gradual deprecation, what would you think about targeting `4.x` for this?

Along the lines of:
- `4.x`: Sphinx stops using jQuery & underscores internally, but still provides it to child themes. Pending removal note added.
- `5.x`: jQuery & underscores are removed by default (followup PR to master).

A

tk0miya:
I'm not sure this refactoring brings will not break 3rd party extensions and themes. So it would be better to change it since 5.0 and remove the dependencies since 6.0.

AA-Turner:
No worries -- will continue to target master then. Added a deprecation note and sample migration/mitigation code -- see https://sphinx--10028.org.readthedocs.build/en/10028/changes.html for a rendered preview.

A

CHANGES
```diff
@@ -22,9 +22,28 @@ Incompatible changes
 Deprecated
 ----------
 
+* #10028: jQuery and underscore.js will no longer be automatically injected into
+  themes from Sphinx 6.0. If your theme's JavaScript uses the ``jQuery``,
```
pradyunsg:
```suggestion
  themes from Sphinx 5.0. If your theme's JavaScript uses the ``jQuery``,
```

CHANGES
```diff
@@ -22,9 +22,28 @@ Incompatible changes
...
+  and underscore.js back to your theme, add the following code to your
+  ``layout.html``.
+
+  .. code-block:: html+jinja
+
+     {%- block scripts %}
+         <script src="{{ pathto('_static/jquery.js', resource=True) }}"></script>
+         <script src="{{ pathto('_static/underscore.js', resource=True) }}"></script>
+         {{ super() }}
+     {%- endblock %}
```
pradyunsg: It's worth noting that this workaround will not work, starting with Sphinx 6.0.

karma.conf.js
```diff
@@ -59,7 +57,7 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['Chrome', 'Firefox'],
+    browsers: ["Firefox"],
```
pradyunsg, line 60: Why the browser removal?

sphinx/themes/basic/static/doctools.js
```diff
@@ -2,322 +2,182 @@
...
- * make the code below compatible with browsers without
- * an installed firebug like debugger
-if (!window.console || !console.firebug) {
-  var names = ["log", "debug", "info", "warn", "error", "assert", "dir",
-    "dirxml", "group", "groupEnd", "time", "timeEnd", "count", "trace",
-    "profile", "profileEnd"];
-  window.console = {};
-  for (var i = 0; i < names.length; ++i)
-    window.console[names[i]] = function() {};
+const _ready = callback => {
```
pradyunsg: Suggestion: Format this code with [prettier](https://prettier.io/).

It makes a fairly reasonable stylistic choices, and notably, removes a large chunk of potential friction about code formatting.

sphinx/themes/basic/static/_compat.js
```diff
@@ -0,0 +1,133 @@
+/*
+ * _compat.js
+ * ~~~~~~~~~~
+ *
+ * Compatability shim for jQuery and underscores.js.
+ *
+ * WILL BE REMOVED IN Sphinx 6.0
```
AA-Turner: Added this compatability module as I realised that themes may rely on the jQuery extensions defined in the original `docutils.js`. These are though all replaceable. 

- `jQuery.urldecode` -> `decodeURIComponent` (or the `URLSearchParams` API)
- `jQuery.urlencode` -> `encodeURIComponent` (or the `URLSearchParams` API)
- `jQuery.getQueryParameters` -> the `URLSearchParams` API
- `jQuery.fn.highlightText` -> the new functions in `docutils.js`

pradyunsg (commented):
Hurrah! Thanks for picking this up @AA-Turner, and taking this over the line! 🎉 

I'm finding the code style of the new code somewhat difficult to read, and have suggested formatting it with prettier below. For the most part though, this looks like it's a roughly 1:1 translation. I don't want to get into an extended debate over automated code formatters; but it's unlikely that I'd have the time/energy to review this PR properly unless the code is reformatted with something similar to prettier's default code style.

In #9874, there were broadly three things that I was doing:

- Dropping the vendored and outdated frameworks.
- Modernising the code, to use classes etc.
- Refactoring the code, to clearly separate the "public" API from the non-public bits (using Python's underscore-prefix naming convention for now).

The third bullet meant taking the liberty to rename JS functions which were clearly not meant to be used outside of the module they were defined in. I think the second and third bullets would still be beneficial to do, but can certainly be achieved in a follow up!

sphinx/themes/basic/static/doctools.js
AA-Turner: In [`c3a63dc8`](https://github.com/sphinx-doc/sphinx/pull/10028/commits/c3a63dc8fa10db26f0204d6a1ed5042f86e3f6d1) I ran prettier with some options -- would it be better to run again with all defaults? (for reference I ran `prettier --print-width 120 --no-semi --quote-props as-needed --no-bracket-spacing --arrow-parens avoid --write sphinx/themes/basic/static`)

A

karma.conf.js
AA-Turner: Only firefox is used in https://github.com/sphinx-doc/sphinx/blob/b22134dcd35c9cffd5c8fe5a3f5224fcb3d3741f/package.json#L4

This was attempting to simplify testing -- if better to drop can do.

sphinx/themes/basic/static/doctools.js
pradyunsg: Just the defaults would be nicer for sure -- I'm extremely used to the default prettier style, TBH. 🙈

CHANGES
pradyunsg: And... it's not just themes; but extensions as well.

CHANGES
AA-Turner: Is this as the source files themselves will be removed from the repo? If so good point -- I should add a note that downsteam authors should add the sources to their `static` directories. (If not, not sure I understand you, sorry!)

CHANGES
AA-Turner: From @tk0miya / @TimKam's comment above, I thought a more gradual deprecation was wanted, not actually stopping providing the files until v6. Very happy to bring forwards to v5, but want to make sure there are no surprises!

sphinx/themes/basic/static/doctools.js
AA-Turner: Done in [`9d9fa72`](https://github.com/sphinx-doc/sphinx/pull/10028/commits/9d9fa72c4c1ed233af6d0d5d15691b430f91e534)

CHANGES
pradyunsg: Ah, I thought 5.0 would stop injecting the files and 6.0 would drop the files. Maybe, I misunderstood that! :)

CHANGES
pradyunsg: > Is this as the source files themselves will be removed from the repo?

Exactly!

karma.conf.js
pradyunsg: Ah, neat. Then, this seems fine by me. I think this should get reintroduced with `--browsers Firefox,Chrome` in the future, but... that's a reasonable follow up PR. :)

sphinx/builders/html/__init__.py
```diff
@@ -315,8 +315,9 @@ def init_js_files(self) -> None:
         self.script_files = []
         self.add_js_file('documentation_options.js', id="documentation_options",
                          data_url_root='', priority=200)
-        self.add_js_file('jquery.js', priority=200)
-        self.add_js_file('underscore.js', priority=200)
+        self.add_js_file('jquery.js', priority=200)  # Remove in Sphinx 6.0
+        self.add_js_file('underscore.js', priority=200)  # Remove in Sphinx 6.0
+        self.add_js_file('_compat.js', priority=200)  # jQuery compatability shim.
```
pradyunsg: nit: This is a very generic name -- something like `_sphinx_jquery_compat.js` will be much less likely to collide with a downstream theme/extensions' JS.

sphinx/themes/basic/static/_compat.js
pradyunsg: nit: Mention `RemovedInSphinx60Warning` in this comment, since it is likely what someone would search for when determining what all to remove as part of preparing for Sphinx 6.0.

sphinx/themes/basic/static/doctools.js
```diff
@@ -2,322 +2,215 @@
...
- * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent#Decoding_query_parameters_from_a_URL
- */
-jQuery.urldecode = function(x) {
-  if (!x) {
-    return x
-  }
-  return decodeURIComponent(x.replace(/\+/g, ' '));
+const _ready = (callback) => {
+  if (document.readyState !== "loading") return callback();
+  document.addEventListener("DOMContentLoaded", callback);
```
pradyunsg: nit: more closely follow a Python-esque code style; that's a bit more "verbose" but has clearer blocks.

```suggestion
  if (document.readyState !== "loading") {
    callback();
  } else {
    document.addEventListener("DOMContentLoaded", callback);
  }
```

sphinx/themes/basic/static/_compat.js
```diff
@@ -0,0 +1,133 @@
...
+    return result;
+};
+
+/**
+ * highlight a given string on a jquery object by wrapping it in
+ * span elements with the given class name.
+ */
+jQuery.fn.highlightText = function(text, className) {
+    function highlight(node, addItems) {
+        if (node.nodeType === 3) {
```
pradyunsg:
```suggestion
        if (node.nodeType === Node.TEXT_NODE) {
```

CHANGES
AA-Turner: > extensions as well

updated note to include extensions

sphinx/builders/html/__init__.py
pradyunsg: nit: Mention `RemovedInSphinx60Warning` in a comment around this, since it is likely what someone would search for when determining what all to remove as part of preparing for Sphinx 6.0.

CHANGES
AA-Turner: updated note to say that downstream must copy the sources from this project

sphinx/themes/basic/static/searchtools.js
```diff
@@ -8,522 +8,493 @@
...
-  };
-}
-
-if (!splitQuery) {
-  function splitQuery(query) {
-    return query.split(/\s+/);
+/**
+ * Simple result scoring code.
+ */
+const Scorer = {
```
pradyunsg: There's a behavior change here -- this should stay a `var`, that only gets assigned under the `if (!Scorer) {...}` block.

Otherwise, someone can't override the scorer settings; which is what this was intended for.

sphinx/themes/basic/static/searchtools.js
```diff
@@ -8,522 +8,493 @@
...
+    const top = startWithContext === 0 ? "" : "...";
+    const tail = startWithContext + 240 < text.length ? "..." : "";
+    let summary = document.createElement("div");
+    summary.classList.add("context");
+    summary.innerText = top + text.substr(startWithContext, 240).trim() + tail;
+    highlightWords.forEach((highlightWord) =>
+      _highlightText(summary, highlightWord, "highlighted")
+    );
+    return summary;
+  },
```
pradyunsg: nit: whitespace to space out chunks of such long-ish functions would be nice.

```suggestion
  makeSearchSummary: (htmlText, keywords, highlightWords) => {
    const text = Search.htmlToText(htmlText).toLowerCase();
    if (text === "") return null;

    const actualStartPosition = [...keywords]
      .map((k) => text.indexOf(k.toLowerCase()))
      .filter((i) => i > -1)
      .slice(-1)[0];
    const startWithContext = Math.max(actualStartPosition - 120, 0);

    const top = startWithContext === 0 ? "" : "...";
    const tail = startWithContext + 240 < text.length ? "..." : "";

    let summary = document.createElement("div");
    summary.classList.add("context");
    summary.innerText = top + text.substr(startWithContext, 240).trim() + tail;

    highlightWords.forEach((highlightWord) =>
      _highlightText(summary, highlightWord, "highlighted")
    );

    return summary;
  },
```

sphinx/themes/classic/layout.html
```diff
@@ -15,3 +15,9 @@
     <script src="{{ pathto('_static/sidebar.js', 1) }}"></script>
     {% endif %}
 {%- endblock %}
+
+{%- block sidebarextra %}{% if theme_collapsiblesidebar|tobool %}
+<div id="sidebarbutton" title="{{ _('Collapse sidebar') }}">
+<span>{{ '»' if theme_rightsidebar|tobool else '«' }}</span>
+</div>
+{% endif %}{% endblock %}
```
pradyunsg, line 23: Hmm... this looks unrelated -- likely from separate work to add a right sidebar to the classic theme?

sphinx/themes/classic/static/classic.css_t
```diff
@@ -153,9 +154,39 @@ div.sphinxsidebar input {
 }
 
 {% if theme_collapsiblesidebar|tobool %}
+{% if theme_rightsidebar|tobool %}
```
pradyunsg, line 157: This also looks unrelated -- likely from separate work to add a right sidebar to the classic theme?

sphinx/themes/basic/layout.html
```diff
@@ -81,6 +81,7 @@ <h3>{{ _('Navigation') }}</h3>
             {%- endblock %}
           {%- endif %}
         </div>
+        {%- block sidebarextra %}{%- endblock %}
```
pradyunsg, line 84: This looks unrelated.

pradyunsg (commented):
Thanks for reformatting the code! ^>^

I've desk-reviewed through most of the PR (i.e. read the patch, but not run the code to try things). I do feel like I need to take another look through the changes in `searchtools.js`, since that code does have a bunch more complexity compared to the doctools.

Beyond that, there seems to be some work related to adding a right sidebar that seems to have leaked in.

And, finally, I tend to prefix some comments with `nit:` which basically means that they are nit-picks. Those are suggestions that are trivial / not worth discussion if you disagree with them. Feel free to mark them as resolved without comments, if you disagree with them or don't feel like addressing them. I don't think they're worth extended discussion.

pradyunsg:
Oh, and because I missed this in the top-level review comment -- This looks pretty great overall! Thanks again for picking this up @AA-Turner! ^.^

AA-Turner:
> Beyond that, there seems to be some work related to adding a right sidebar that seems to have leaked in.

If this is the stuff in `classic`, then it is as it used jQuery to do a load of positioning stuff - I may have gone a little overboard in reimplementing it with CSS etc!

pradyunsg:
> If this is the stuff in `classic`, then it is as it used jQuery to do a load of positioning stuff - I may have gone a little overboard in reimplementing it with CSS etc!

Ah, that would explain a lot! TIL that classic allowed you to move the sidebar to the right. :)

sphinx/themes/basic/static/_compat.js
AA-Turner: Not going to do this one purely to keep everything in the compat module as it is today.

sphinx/themes/basic/static/searchtools.js
AA-Turner: Ahh, I wondered why it was -- i was under the impression it was some ancient workaround (most likely for IE) that never got removed. Will revert, and add a note speaking as to why it is defined as such.

sphinx/themes/basic/layout.html
AA-Turner: Part of the work to fix `classic`

AA-Turner:
@pradyunsg please may I ask for a quick look over my changes? / any more thoughts that would block this.

A

pradyunsg:
I can take a look on Tuesday, at the earliest.

TimKam:
Note that I added an issue for the final removal of the dependencies, which will require some coordination: https://github.com/sphinx-doc/sphinx/issues/10070

doc/_themes/sphinx13/layout.html
```diff
@@ -27,28 +27,25 @@
...
-        var sb = $('.sphinxsidebarwrapper');
-        var win = $(window);
-        var sbh = sb.height();
-        var offset = $('.sphinxsidebar').position()['top'];
-        var wintop = win.scrollTop();
-        var winbot = wintop + win.innerHeight();
+      jQuery(window).scroll(function() {
+        var sb = document.getElementsByClassName('sphinxsidebarwrapper')[0]
+        var sbh = sb.offsetHeight
+        var offset = jQuery('.sphinxsidebar').position()['top'];
```
pradyunsg: This still depends on jQuery. Is that intentional?

pradyunsg (commented):
Still haven’t found time to checkout this branch and run things, so… here’s another round of me reviewing this PR from a handheld device with a touch screen.

doc/_themes/sphinx13/layout.html
AA-Turner: Sorry! Just pushed a commit to fix this.

A

AA-Turner:
Are there any more blockers on this (beyond review)? There's now a merge conflict but I want to check we've accepted this method of communicating the change etc before putting much more work into this.

If you'd prefer I resolve the conflict first though, happy to do so.

A

pradyunsg (approved):
This looks good to me overall! If there's something that's slipped through the cracks that both @AA-Turner and I couldn't spot, we can certainly fix that in a follow up!

AA-Turner:
Fixed merge conflicts, sorry it took a while.

@tk0miya or @TimKam would you be able to advise on next steps please? Pradyun gave the all clear above.

A

TimKam:
Thank you. I would like to click myself through a build in a couple of browsers asap, but I don't see/expect any issues. ECMA 2018 is not (fully) supported by some fringe or very dated browsers. I assume we can live with this, because Sphinx documentation is used primarily by developers or tech-literature folks, so the share of Internet Explorer (and similar) users should be negligible. So if @tk0miya is okay with it, I would most likely merge after I have done some manual testing.
@AA-Turner and @pradyunsg could briefly mention to what extent you have manually tested the build result?

AA-Turner:
Sounds good, thanks! 

Tested on latest Firefox and Chrome, but not much beyond that. Testing in Safari (WebKit) would cover all iOS usages, so I could do that. I assume Google use the same browser engine on mobile and desktop, although not certain from memory.

A

pradyunsg:
As far as I can tell, there's nothing from ECMAScript 2018 in this PR. There's a decent amount of ECMAScript 2015 (ES6) tho, and that's pretty well supported across the board: https://caniuse.com/es6. See also https://caniuse.com/?compare=ie+11,edge+94,edge+95,edge+96,edge+97,firefox+76,firefox+86,firefox+96,chrome+77,chrome+87,chrome+97,safari+13.1,safari+14.1,safari+15.1,safari+15.2-15.3,ios_saf+15.2-15.3,android+97,and_chr+97,and_ff+95&compareCats=JS

I've tried out Chrome and Firefox on MacOS. I considered trying out other browsers, but looking at the patch, nothing stood out as problematic. I realised the ReadTheDocs preview of this PR uses the new JS, so... trying out the basics on other browsers is fairly straightforward: https://sphinx--10028.org.readthedocs.build/en/10028/ -- it won't exercise the translations JS on there tho. I haven't tried, but I won't be surprised if stuff still works on IE 11.

I have access to BrowserStack via their OSS plans (and I'm pretty sure Sphinx maintainers can sign up too), so... in theory, we can test whatever browser combinations we want. If there's a specific combination that you'd like to see tested, lemme know! 

Screenshot 2022-01-29 at 07 54 39

pradyunsg:
Oh, one note about the RtD preview -- they're calling Search.init an extra time in their injected JS, which is why there's a redundant "Searching..." on the search page: https://github.com/readthedocs/readthedocs.org/issues/8862

CHANGES
```diff
@@ -41,6 +58,9 @@ Features added
 * #9075: autodoc: The default value of :confval:`autodoc_typehints_format` is
   changed to ``'smart'``.  It will suppress the leading module names of
   typehints (ex. ``io.StringIO`` -> ``StringIO``).
+* #10028: Removed internal usages of JavaScript frameworks (jQuery and
+  underscore.js) and modernised ``doctools.js`` and ``searchtools.js`` to
+  EMCAScript 2018.
```
pradyunsg, line 63:
```suggestion
  EMCAScript 2015 (ES6).
```

pradyunsg:
> could briefly mention to what extent you have manually tested the build result?

- Do a search.
  - Check that the behaviour is same as regular Sphinx (the period handling, etc). You can slow the network + disable the cache via the browser's development tools to be able to test/check some of this.
- Click a search result.
- Hide search matches.
- Use the console to invoke `Documentation.initOnKeyListeners` and then try those out.
- Open a domain index, and click to expand/collapse stuff.

This tests basically all of doctools, and the main parts of searchtools as far as I can tell. I haven't bothered testing out any of the theme-related changes, especially not the CSS changes to sphinx13/classic.

tk0miya:
+1 for merging.

TimKam:
There is one -- I assume final -- issue that we may need to address. Open this link: [https://sphinx--10028.org.readthedocs.build/en/10028/search.html?q=test](https://sphinx--10028.org.readthedocs.build/en/10028/search.html?q=test) and check the console output. We get an error: `Uncaught TypeError: Search.title.text is not a function`. This only happens with RTD, not with 'vanilla' Sphinx. But because RTD hosts a lot of Sphinx docs for many projects and readers, it is important to at least inform them. Presumably, they can replace `Search.title.text` by `Search.title.innerText` ([here](https://github.com/readthedocs/readthedocs.org/blob/db0c0e26337bfaa42f93dbd8237c28cfca224126/readthedocs/core/static-src/core/js/doc-embed/search.js)). Still, considering that they need to support all kinds of Sphinx versions, I am wondering if it isn't more pragmatic to stick to `Search.title.text` on Sphinx's side. @AA-Turner: what do you think?

AA-Turner:
I believe RTD could do `$(Search.title).text(...)` which would work in any version, as it wraps `Search.title` in a new jQuery object instead of assuming it is one.

We could also add a transition function to enable updating functionally, but this is less elegant. Do you know who the best person would be to ask on the RTD side about this?

A

pradyunsg:
Let's move that discussion over to https://github.com/readthedocs/readthedocs.org/issues/8862? I'm pretty sure @humitos et al will be looking there! :)

TimKam:
Okay. I'll merge then. After all, the error does not break the search for RTD.

TimKam:
Thanks all for the great efforts!

PR #10040, opened 2022-01-01
Close #9595: Deprecate setuptools integration

### LINKED ISSUE #9595: Make the distutils/setuptools integration an opt-in for projects

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- refs: #9595


doc/extdev/deprecated.rst
```diff
@@ -22,6 +22,11 @@ The following is a list of deprecated interfaces.
      - (will be) Removed
      - Alternatives
 
+   * - :doc:`Setuptools integration </usage/advanced/setuptools>`
+     - 5.0
+     - 7.0
+     - N/A
```
tk0miya, line 28: I'm not sure what is the successor of `python setup.py build_sphinx` on building packages. [This article](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) says tox is one of them.

astrojuanlu: Yep, I've been using `tox -e docs` for my projects for quite some time

pradyunsg: Honestly, using `sphinx-build` directly is the only alternative. For integrating this into the users' workflows, they can use something like `tox`, `nox`, `just` etc to simplify invoking sphinx-build (similar to how `make html` makes things easier).

pradyunsg: i.e. `tox -e docs` is not an alternative -- you could very well have `python setup.py build_sphinx` in your `tox.ini` file. Recommending using `sphinx-build` directly is reasonable IMO.

tk0miya: Thank you for your comments. It seems my understanding is correct :-)
It would be nice if `sphinx-build` command will support `pyproject.toml` in the future. But it's off-topic.

Issue #1061, opened 2015-01-03
autosummary document imported functions


shimizukawa:
Hello,

I'm looking for a way to exclude imported modules and functions.

The attached examples list **isfile** in functions of my module.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1061
- Originally reported by: Anonymous
- Originally created at: 2012-12-18T16:59:40.487

shimizukawa:
_From Takayuki Shimizukawa on 2013-01-20 02:12:50+00:00_

I can not reproduce autodocing for imported modules and functions.

Please let me know about your 'autodoc' relative directive using in rst file.

shimizukawa:
_From Tiago Coutinho on 2013-10-10 09:15:55+00:00_

Hello,
I am using sphinx 1.1.3 to document my python project. I can confirm the bug.

I was preparing a bug report when I saw this one. I post my original report with the hope that it can help solve the bug:

I am using autosummary with autosummary_generate = True in the conf.py

The autosummary is generating documentation for functions/classes which are imported from another module.
Example:

```
#!python

# --- module1.py ----
def foo():
    """foo documentation"""
    pass

# --- module2.py ---
from module1 import foo
def bar():
    """bar documentation"""
    foo()
```

and the following RST:

```
#!rst

.. doc.rst

Documentation
============

.. autosummary::
   :toctree:

   module2
```

will generate a RST like:

```
#!rst

module2
=======

.. automodule:: module2

   .. rubric:: Functions

   .. autosummary::

      foo
      bar
```

Please don't hesitate to ask for more info if needed.
Thanks,
Tiago

shimizukawa:
_From [samuele_c](https://bitbucket.org/samuele_c) on 2014-03-05 00:02:35+00:00_

I'm having the exact same issue in Debian stable and Debian testing. Calling

```
.. autosummary:: 
   :toctree: _autosummary
```

includes functions imported from other modules in the summary table

shimizukawa:
_From [mzjn](https://bitbucket.org/mzjn) on 2014-08-23 17:44:07+00:00_

A patch has been suggested at Stack Overflow: http://stackoverflow.com/a/25460763

shimizukawa:
_From Takayuki Shimizukawa on 2014-08-29 01:53:58+00:00_

We need a `:imported: True` option for `autosummary` (I think False is better for default).

It's a new feature and a incompatible change.

[...]

PR #3235, opened 2016-12-13
fix #2336 autosummary imported members

lsaffre:
This fixes #2336

PR #1892, opened 2015-05-19
FIX: Only check if members is True

larsoner:
Closes #1891.
Closes #1822.

`check_module` should only be called if `members` is currently activated, otherwise formely working use cases like those in #1891 are broken. This is also consistent with the docs:

> In an automodule directive with the members option set, only module members whose **module** attribute is equal to the module name as given to automodule will be documented. This is to prevent documentation of imported classes or functions. Set the imported-members option if you want to prevent this behavior and document all available members. Note that attributes from imported modules will not be documented, because attribute documentation is discovered by parsing the source file of the current module.

Ready for review/merge from my end.

PR #10060, opened 2022-01-07
Fix #10058: Autosummary excludes stuff from its table if `autodoc_class_signature`='separated'.

### LINKED ISSUE #10058: autosummary - autosummary_imported_members doesn't apply to items in table
### Describe the bug

If I have
`private.py`
```py
class DocumentMe:
  "document me with a docstring"
  pass
```
`public.py`
```py
from private import DocumentMe
def some_fn():
  'document me as well'
__all__ = ['DocumentMe', 'some_fn']
```

`conf.py`
```py
extensions = [
    'sphinx.ext.autosummary',
    'sphinx.ext.autodoc'
]
autosummary_generate = False
autosummary_imported_members = True
autodoc_class_signature = 'separated'
```

`docs.rst`
```rst

.. currentmodule:: public

.. autosummary::

   some_fn
   DocumentMe
```

Then I get a table with `some_fn` but not `DocumentMe`:
|  |  |
|----|----|
|`some_fn()`|document me as well|

I believe this is because the ModuleDocumenter being used never gets its `imported_modules` flag set, regardless of app config. I'll raise a PR with a basic fix now to illustrate.

### How to Reproduce

Set up files as per above

### Expected behavior

A table that looks like
|  |  |
|----|----|
|`some_fn()`|document me as well|
|`DocumentMe`|document me with a docstring|

### Your project

github.com/akdor1154/csql#caching-1

### Screenshots

_No response_

### OS

Linux

### Python version

3.9

### Sphinx version

4.3.2

### Sphinx extensions

autodoc, autosummary

### Extra tools

_No response_

### Additional context

_No response_

akdor1154:
Subject: Fix #10058: Autosummary excludes stuff from its table if `autodoc_class_signature`='separated'.

### Feature or Bugfix
- Bugfix

### Purpose
When using autosummary with `autodoc_class_signature` = 'separated', it won't include any Classes in its summary table, even if you explicitly ask for them.

This seems to be because when this method is set, the ClassDocumenter gets its `options.members` set to `['__new__', '__init__']`; it seems to be always `None`/`[]` in other cases. So I think the `check_module()` check ([ext/autosummary/\_\_init__.py:376](https://github.com/sphinx-doc/sphinx/blob/6402acde81e39cf010925c925733546dd6a2f822/sphinx/ext/autosummary/__init__.py#L375)) has never really been working properly, but it hasn't mattered because `options.members` is normally falsy in most situations.

The included test shows the broken behaviour pretty clearly, the fix commit seems to make sense to me but I'm not familiar with Sphinx source code, maybe there's a better fix.

### DIFF
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -256,8 +256,9 @@ class Autosummary(SphinxDirective):
     }
 
     def run(self) -> List[Node]:
+        opts = Options({'imported-members': self.env.app.config.autosummary_imported_members})
         self.bridge = DocumenterBridge(self.env, self.state.document.reporter,
-                                       Options(), self.lineno, self.state)
+                                       opts, self.lineno, self.state)
 
         names = [x.strip().split()[0] for x in self.content
                  if x.strip() and re.search(r'^[~a-zA-Z_]', x.strip()[0])]
--- a/tests/roots/test-ext-autosummary-imported_members/conf.py
+++ b/tests/roots/test-ext-autosummary-imported_members/conf.py
@@ -6,3 +6,6 @@ sys.path.insert(0, os.path.abspath('.'))
 extensions = ['sphinx.ext.autosummary']
 autosummary_generate = True
 autosummary_imported_members = True
+
+# test for #10058 - shoud be scoped just to test_autosummary_imported_members_table_10058
+autodoc_class_signature = 'separated'
--- a/tests/roots/test-ext-autosummary-imported_members/index.rst
+++ b/tests/roots/test-ext-autosummary-imported_members/index.rst
@@ -5,3 +5,10 @@ test-ext-autosummary-mock_imports
    :toctree: generated
 
    autosummary_dummy_package
+
+.. currentmodule:: autosummary_dummy_package
+
+.. autosummary::
+
+   foo
+   Bar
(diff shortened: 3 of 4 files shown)


tk0miya:
Is this related to `autodoc_class_signature`?

tk0miya:
> Is this related to `autodoc_class_signature`?

Ah, sorry. Now I understand this bug. It is consist of two bugs. One is an imported object is shown even if `autosummary_imported_members = False`. Another is an imported object is not shown when `autodoc_class_signature` is enabled even if `autosummary_imported_members = True`.

I'll start to review this again from now on.

tk0miya:
Note:

* `check_module()` call was added at https://github.com/sphinx-doc/sphinx/commit/21b838454f4bb5b6a246b85cb2f521deb7552505 to fix https://github.com/sphinx-doc/sphinx/issues/1061.
* Addition to this, `options.members` check was added at https://github.com/sphinx-doc/sphinx/pull/1892. But the option will never be filled. So `check_module()` will not be called basically.
* Finally, https://github.com/sphinx-doc/sphinx/pull/3235 fixed the original bug using another approach.
* It means the `check_module()` call is meaningless at present.

tk0miya:
Merged #10085 instead. Thank you for your contribution!

akdor1154:
thanks!

PR #10077, opened 2022-01-10
collapse nested if blocks (SIM102)

danieleades:
collapses some unnecessarily nested 'if' blocks.

this PR is not exhaustive- there are other examples that are not addressed.

in this humble author's opinion, manually formatting complex boolean logic is a crapshoot, and would be better handled with an autoformatting tool.

### DIFF
--- a/sphinx/builders/_epub_base.py
+++ b/sphinx/builders/_epub_base.py
@@ -418,11 +418,10 @@ class EpubBuilder(StandaloneHTMLBuilder):
                     logger.warning(__('cannot copy image file %r: %s'),
                                    path.join(self.srcdir, src), err)
                 continue
-            if self.config.epub_fix_images:
-                if img.mode in ('P',):
-                    # See the Pillow documentation for Image.convert()
-                    # https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert
-                    img = img.convert()
+            if self.config.epub_fix_images and img.mode in ('P',):
+                # See the Pillow documentation for Image.convert()
+                # https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert
+                img = img.convert()
             if self.config.epub_max_image_width > 0:
                 (width, height) = img.size
                 nw = self.config.epub_max_image_width
--- a/sphinx/domains/c.py
+++ b/sphinx/domains/c.py
@@ -1075,9 +1075,11 @@ class ASTDeclaratorPtr(ASTDeclarator):
             if self.restrict or self.volatile:
                 signode += addnodes.desc_sig_space()
             _add_anno(signode, 'const')
-        if self.const or self.volatile or self.restrict or len(self.attrs) > 0:
-            if self.next.require_space_after_declSpecs():
-                signode += addnodes.desc_sig_space()
+        if (
+            (self.const or self.volatile or self.restrict or len(self.attrs) > 0) and
+            self.next.require_space_after_declSpecs()
+        ):
+            signode += addnodes.desc_sig_space()
         self.next.describe_signature(signode, mode, env, symbol)
 
 
@@ -2576,9 +2578,8 @@ class DefinitionParser(BaseParser):
         if self.match(_simple_type_specifiers_re):
             return self.matched_text
         for t in ('bool', 'complex', 'imaginary'):
-            if t in self.config.c_extra_keywords:
-                if self.skip_word(t):
-                    return t
+            if t in self.config.c_extra_keywords and self.skip_word(t):
+                return t
         return None
 
     def _parse_simple_type_specifiers(self) -> ASTTrailingTypeSpecFundamental:
@@ -2764,22 +2765,18 @@ class DefinitionParser(BaseParser):
                 volatile = False
                 restrict = False
                 while True:
-                    if not static:
-                        if self.skip_word_and_ws('static'):
-                            static = True
-                            continue
-                    if not const:
-                        if self.skip_word_and_ws('const'):
-                            const = True
-                            continue
-                    if not volatile:
-                        if self.skip_word_and_ws('volatile'):
-                            volatile = True
-                            continue
-                    if not restrict:
-                        if self.skip_word_and_ws('restrict'):
-                            restrict = True
-                            continue
+                    if not static and self.skip_word_and_ws('static'):
+                        static = True
+                        continue
+                    if not const and self.skip_word_and_ws('const'):
+                        const = True
+                        continue
+                    if not volatile and self.skip_word_and_ws('volatile'):
+                        volatile = True
+                        continue
+                    if not restrict and self.skip_word_and_ws('restrict'):
+                        restrict = True
+                        continue
                     break
                 vla = False if static else self.skip_string_and_ws('*')
                 if vla:
@@ -2801,13 +2798,16 @@ class DefinitionParser(BaseParser):
             else:
                 break
         param = self._parse_parameters(paramMode)
-        if param is None and len(arrayOps) == 0:
+        if (
+            param is None and
+            len(arrayOps) == 0 and
             # perhaps a bit-field
-            if named and paramMode == 'type' and typed:
-                self.skip_ws()
-                if self.skip_string(':'):
-                    size = self._parse_constant_expression()
-                    return ASTDeclaratorNameBitField(declId=declId, size=size)
+            named and paramMode == 'type' and typed
+        ):
+            self.skip_ws()
+            if self.skip_string(':'):
+                size = self._parse_constant_expression()
+                return ASTDeclaratorNameBitField(declId=declId, size=size)
         return ASTDeclaratorNameParam(declId=declId, arrayOps=arrayOps,
                                       param=param)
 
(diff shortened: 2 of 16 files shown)


sphinx/util/stemmer/porter.py
```diff
@@ -258,9 +258,8 @@ def step2(self) -> None:
                 self.r("ive")
             elif self.ends("biliti"):
                 self.r("ble")
-        elif self.b[self.k - 1] == 'g':  # --DEPARTURE--
-            if self.ends("logi"):
-                self.r("log")
+        elif self.b[self.k - 1] == 'g' and self.ends("logi"):  # --DEPARTURE--
```
tk0miya, line 261: I think this should not be merged. The parent all other conditions in the parent if-block checks `self.b[self.k - 1]`. It's like a switch syntax in other languages. So it's better to keep it, IMO.

sphinx/domains/c.py
```diff
@@ -1075,9 +1075,11 @@ def _add_anno(signode: TextElement, text: str) -> None:
...
                 signode += addnodes.desc_sig_space()
             _add_anno(signode, 'const')
-        if self.const or self.volatile or self.restrict or len(self.attrs) > 0:
-            if self.next.require_space_after_declSpecs():
-                signode += addnodes.desc_sig_space()
+        if (
+            (self.const or self.volatile or self.restrict or len(self.attrs) > 0) and
+            self.next.require_space_after_declSpecs()
+        ):
+            signode += addnodes.desc_sig_space()
```
jakobandersen, line 1082: Should not be changed, the original better conveys the semantics and is more readable.

sphinx/domains/c.py
```diff
@@ -2576,9 +2578,8 @@ def _parse_simple_type_specifier(self) -> Optional[str]:
         if self.match(_simple_type_specifiers_re):
             return self.matched_text
         for t in ('bool', 'complex', 'imaginary'):
-            if t in self.config.c_extra_keywords:
-                if self.skip_word(t):
-                    return t
+            if t in self.config.c_extra_keywords and self.skip_word(t):
+                return t
```
jakobandersen, line 2582: Should not be changed, the original better conveys the semantics and is more readable.

sphinx/domains/c.py
```diff
@@ -2801,13 +2798,16 @@ def parser():
...
-                self.skip_ws()
-                if self.skip_string(':'):
-                    size = self._parse_constant_expression()
-                    return ASTDeclaratorNameBitField(declId=declId, size=size)
+            paramMode == 'type' and typed
+        ):
+            self.skip_ws()
+            if self.skip_string(':'):
+                size = self._parse_constant_expression()
+                return ASTDeclaratorNameBitField(declId=declId, size=size)
```
jakobandersen: Should not be changed, the original better conveys the semantics and is more readable. Here the transformation is even wrong because code has been moved across the comment!

sphinx/domains/cpp.py
```diff
@@ -2600,9 +2600,11 @@ def _stringify(self, transform: StringifyTransform) -> str:
...
                 res.append(' ')
             res.append('const')
-        if self.const or self.volatile or len(self.attrs) > 0:
-            if self.next.require_space_after_declSpecs():
-                res.append(' ')
+        if (
+            (self.const or self.volatile or len(self.attrs) > 0) and
+            self.next.require_space_after_declSpecs()
+        ):
+            res.append(' ')
```
jakobandersen, line 2607: Should not be changed, the original better conveys the semantics and is more readable.

sphinx/domains/cpp.py
```diff
@@ -2660,9 +2662,11 @@ def _add_anno(signode: TextElement, text: str) -> None:
...
                 signode += addnodes.desc_sig_space()
             _add_anno(signode, 'const')
-        if self.const or self.volatile or len(self.attrs) > 0:
-            if self.next.require_space_after_declSpecs():
-                signode += addnodes.desc_sig_space()
+        if (
+            (self.const or self.volatile or len(self.attrs) > 0) and
+            self.next.require_space_after_declSpecs()
+        ):
+            signode += addnodes.desc_sig_space()
```
jakobandersen, line 2669: Should not be changed, the original better conveys the semantics and is more readable.

sphinx/domains/cpp.py
```diff
@@ -7911,9 +7915,12 @@ def checkType() -> bool:
...
                             addParen += 1
-                        elif typ == 'func':
-                            if title.endswith('()') and not displayName.endswith('()'):
-                                title = title[:-2]
+                        elif (
+                            typ == 'func' and
+                            title.endswith('()') and
+                            not displayName.endswith('()')
+                        ):
+                            title = title[:-2]
```
jakobandersen, line 7923: Should not be changed, the original better conveys the semantics and is more readable. Here it merges nested conditions with top-level if-elif, which is quite bad. It's two different things.

jakobandersen (changes requested):
I have looked at the C and C++ domains, and I find most of the changes disturbing. If the ratio of problematic to good transformations is similar then I don't think SIM102 is good to auto-apply in general.

sphinx/builders/_epub_base.py
```diff
@@ -418,11 +418,10 @@ def copy_image_files_pil(self) -> None:
                     logger.warning(__('cannot copy image file %r: %s'),
                                    path.join(self.srcdir, src), err)
                 continue
-            if self.config.epub_fix_images:
-                if img.mode in ('P',):
-                    # See the Pillow documentation for Image.convert()
-                    # https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.convert
-                    img = img.convert()
+            if self.config.epub_fix_images and img.mode in ('P',):
```
danieleades, line 421: @jakobandersen is there a reason you don't like collapsing the statements [here](https://github.com/sphinx-doc/sphinx/pull/10077/files#r780977852), but don't mind collapsing them at this line? As an outsider, I'm struggling to see the semantic difference. Maybe you can give me a little guidance as to your reasoning?

sphinx/domains/c.py
danieleades: true

sphinx/domains/cpp.py
danieleades: can you explain why you think the original better conveys the semantics?

to me they convey identical semantics, but with one less level of nesting.

not saying you're wrong, just wanting to hear your reasoning.

perhaps the meaning of this code is better conveyed by *naming* the boolean checks here with an assignment before the `if` statement, then by any particular arrangement of the statements?

sphinx/domains/cpp.py
danieleades: > Here it merges nested conditions with top-level if-elif, which is quite bad

can you explain why this is different to the `if` statement directly above it?

the line above-

```python
if typ == 'any' and additional_conditions
   ...
```
the line i've changed-

```python
elif typ == 'func' and additional_conditions
   ...
```

if anything it's now more consistent. right?

tk0miya (changes requested):
LGTM with nits.

sphinx/builders/_epub_base.py
jakobandersen: As mentioned originally: I only reviewed the C and C++ domains as that's the code I know best. The example you linked to is not the most problematic, and I could live with that particular change. I don't know the epub format, so I can't comment on what is reasonable here.

sphinx/domains/cpp.py
jakobandersen: The outer condition is whether there is a local grammatical construct (CV-qualifiers and attributes) which means we should even consider emitting a space. The inner condition is whether a nested grammatical construct (the next declarator) is of a type which necessitates a space.
Stylistically the change means it takes up more lines and introduces parens, which I both find less readable as well for this case.

sphinx/domains/cpp.py
jakobandersen: In the original you have three cases: 1) an ``any`` with ``()``which could be a function, 2) definitely a function, 3) otherwise. Case (2) happens to have nested conditions for when the title should be processed.
With the else-part, is the transformation even the same code? Say for something which is a ``fiunc`` but the title doesn't end with ``()`` then before it would hit case (2) and fail, but with the change it would hit the ``else`` case.
Perhaps the transformation is valid, but it is definitely not obvious to me right now.

danieleades: it's definitely logically identical. 

> Say for something which is a fiunc but the title doesn't end with () then before it would hit case (2) and fail, but with the change it would hit the else case.

exactly. in the original code it would hit the inner clause and fail. with the new code it will hit the outer clause and fail. There are no other nested cases to handle, so these can be collapsed. This is true any time a conditional block contains *only* a single other conditional block.

Worth mentioning there's no semantic difference between an `if` clause and an `elif` clause other than the order in which they execute. you can satisfy yourself this is true by adding changing the `if` clause to an `elif` clause, and adding a new

```python
if False:
    continue
````
before it

jakobandersen: Ah, I visually misaligned the ``else`` afterwards with this conditional. That is, there are only 2 cases, so the transformation is indeed sound. Then I have no strong opinion about this case.

AA-Turner:
Closed, a maintainer is against a number of the changes and we haven't heard from the proponent of the change since January.

A

PR #10078, opened 2022-01-10
address some unused loop control variables

danieleades:
tidy up some loop control variables

### DIFF
--- a/doc/development/tutorials/examples/recipe.py
+++ b/doc/development/tutorials/examples/recipe.py
@@ -87,7 +87,7 @@ class RecipeIndex(Index):
         # first letter of the recipe as a key to group thing
         #
         # name, subtype, docname, anchor, extra, qualifier, description
-        for name, dispname, typ, docname, anchor, _ in recipes:
+        for _name, dispname, typ, docname, anchor, _ in recipes:
             content[dispname[0].lower()].append(
                 (dispname, 0, docname, anchor, docname, '', typ))
 
--- a/doc/development/tutorials/examples/recipe.py
+++ b/doc/development/tutorials/examples/recipe.py
@@ -87,7 +87,7 @@ class RecipeIndex(Index):
         # first letter of the recipe as a key to group thing
         #
         # name, subtype, docname, anchor, extra, qualifier, description
-        for _name, dispname, typ, docname, anchor, _ in recipes:
+        for _name, dispname, typ, docname, anchor, _priority in recipes:
             content[dispname[0].lower()].append(
                 (dispname, 0, docname, anchor, docname, '', typ))
 
--- a/sphinx/writers/texinfo.py
+++ b/sphinx/writers/texinfo.py
@@ -312,7 +312,7 @@ class TexinfoTranslator(SphinxTranslator):
             del node_menus[top['node_name']]
             top['node_name'] = 'Top'
         # handle the indices
-        for name, content in self.indices:
+        for name, _content in self.indices:
             node_menus[name] = []
             node_menus['Top'].append(name)
 
@@ -320,7 +320,7 @@ class TexinfoTranslator(SphinxTranslator):
         """Collect the relative links (next, previous, up) for each "node"."""
         rellinks = self.rellinks
         node_menus = self.node_menus
-        for id, entries in node_menus.items():
+        for id in node_menus.keys():
             rellinks[id] = ['', '', '']
         # up's
         for id, entries in node_menus.items():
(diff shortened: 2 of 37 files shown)


danieleades:
i've not touched `sphinx.util.get_matching_files`, but the function looks a little fishy to me

```python
def get_matching_files(dirname: str,
                       exclude_matchers: Tuple[PathMatcher, ...] = ()) -> Iterable[str]:  # NOQA
    """Get all file names in a directory, recursively.

    Exclude files and dirs matching some matcher in *exclude_matchers*.
    """
    # dirname is a normalized absolute path.
    dirname = path.normpath(path.abspath(dirname))
    dirlen = len(dirname) + 1    # exclude final os.path.sep

    for root, dirs, files in os.walk(dirname, followlinks=True):
        relativeroot = root[dirlen:]

        qdirs = enumerate(path_stabilize(path.join(relativeroot, dn))
                          for dn in dirs)  # type: Iterable[Tuple[int, str]]
        qfiles = enumerate(path_stabilize(path.join(relativeroot, fn))
                           for fn in files)  # type: Iterable[Tuple[int, str]]
        for matcher in exclude_matchers:
            qdirs = [entry for entry in qdirs if not matcher(entry[1])]
            qfiles = [entry for entry in qfiles if not matcher(entry[1])]

        dirs[:] = sorted(dirs[i] for (i, _) in qdirs)

        for i, filename in sorted(qfiles):
            yield filename
```

it doesn't necessarily need to be addressed in this PR, but I wonder if the logic is unnecessarily convoluted.

doc/development/tutorials/examples/recipe.py
```diff
@@ -87,7 +87,7 @@ def generate(self, docnames=None):
         # first letter of the recipe as a key to group thing
         #
         # name, subtype, docname, anchor, extra, qualifier, description
-        for name, dispname, typ, docname, anchor, _ in recipes:
+        for _name, dispname, typ, docname, anchor, _ in recipes:
```
tk0miya: Where this leading underscore convention comes from?

danieleades: While it's not mentioned in PEP8, it's a *very* common convention in Python. It's also a convention in JavaScript, typescript, Haskell, Rust, Ruby, Go, and probably others

tk0miya: In this line, the last element is named as "_". It's also a well-used technique, I think. Which is better? If we'll switch to your proposal, do we have to fix it too?

sphinx/writers/texinfo.py
```diff
@@ -312,15 +312,15 @@ def collect_node_menus(self) -> None:
...
+        for name, _content in self.indices:
             node_menus[name] = []
             node_menus['Top'].append(name)
 
     def collect_rellinks(self) -> None:
         """Collect the relative links (next, previous, up) for each "node"."""
         rellinks = self.rellinks
         node_menus = self.node_menus
-        for id, entries in node_menus.items():
+        for id in node_menus.keys():
```
tk0miya: Is `keys()` call needed?
```suggestion
        for id in node_menus:
```

tests/test_ext_napoleon.py
```diff
@@ -101,9 +101,9 @@ def test_unknown_app_type(self):
     def test_add_config_values(self):
         app = mock.Mock(Sphinx)
         setup(app)
-        for name, (default, rebuild) in Config._config_values.items():
+        for name in Config._config_values.keys():
```
tk0miya: This is also.
```suggestion
        for name in Config._config_values:
```

doc/development/tutorials/examples/recipe.py
tk0miya: Note: the last element of `recipes` is a "priority" value.

tk0miya (commented):
LGTM with nits.

doc/development/tutorials/examples/recipe.py
danieleades: > In this line, the last element is named as "_". It's also a well-used technique, I think. Which is better? If we'll switch to your proposal, do we have to fix it too?

Using a single underscore is also a common convention. Personally I prefer the leading underscore because its more explicit about *what* you're ignoring

sphinx/writers/texinfo.py
danieleades: No, not required. Will update

doc/development/tutorials/examples/recipe.py
tk0miya: +1. Let's change it to `_priority`.

danieleades: done

danieleades:
@tk0miya comments addressed. You may want to squash merge this one

tk0miya:

PR #10080, opened 2022-01-10
simplify some set comparisons (SIM109)

danieleades:


sphinx/util/logging.py
```diff
@@ -375,12 +375,8 @@ def is_suppressed_warning(type: str, subtype: str, suppress_warnings: List[str])
...
             target, subtarget = warning_type, None
 
-        if target == type:
-            if ((subtype is None and subtarget is None) or
-                    subtarget is None or
-                    subtarget == subtype or
-                    subtarget == '*'):
-                return True
+        if target == type and subtarget in [None, subtype, "*"]:
+            return True
```
danieleades: was scratching my head over this for a little while. wouldn't mind an extra pair of eyes or two to check my logic here!

jakobandersen:
Probably mostly curiosity and nitpicking, but I would guess that a tuple was faster to construct than a list, though I have only done a basic test that confirms it.

danieleades:
> Probably mostly curiosity and nitpicking, but I would guess that a tuple was faster to construct than a list, though I have only done a basic test that confirms it.

that's an interesting thought! i found this resource - https://stackoverflow.com/a/25368371/17539451

that appears to indicate there's no difference between `tuple` and `list` in practice here, but that a `set` might be slightly faster. Ultimately, i very much doubt it's going to be a bottleneck, so probably whichever one is most readable is the right move. I'm inclined to think a list (or a tuple for that matter) is slightly easier to grok than a set, simply because they're more common.

we probably talking about nanoseconds here though...

jakobandersen:
> that's an interesting thought! i found this resource - https://stackoverflow.com/a/25368371/17539451

Very nice! And it also explain why I saw a difference in my own example, as I assigned the object.

> that appears to indicate there's no difference between `tuple` and `list` in practice here, but that a `set` might be slightly faster.

But a set requires the elements to be hashable, so in general that transformation would not work.

> Ultimately, i very much doubt it's going to be a bottleneck, so probably whichever one is most readable is the right move. I'm inclined to think a list (or a tuple for that matter) is slightly easier to grok than a set, simply because they're more common.
>
> we probably talking about nanoseconds here though...

Indeed, go for readability, I just got curious :-). And it's nice to know about that peephole optimization.

tk0miya (approved):
LGTM.

IMO, I prefer a bit to use tuples for such purposes. How about replacing lists by tuples if nobody objects?

danieleades:
> LGTM.
> 
> IMO, I prefer a bit to use tuples for such purposes. How about replacing lists by tuples if nobody objects?

done

sphinx/util/logging.py
mgeier: Did you drop the check for `subtype is None` on purpose?

This doesn't look right. Or am I missing something?

danieleades: > Did you drop the check for `subtype is None` on purpose?
> 
> This doesn't look right. Or am I missing something?

I did! here's the original code-

```python
        if target == type:
            if ((subtype is None and subtarget is None) or
                    subtarget is None or
                    subtarget == subtype or
                    subtarget == '*'):
                return True
```

the case you're referring to is `(subtype is None and subtarget is None)`. There's another `or` case which is `subtarget == subtype`. The second case includes the case that they're both `None`, hence `(subtype is None and subtarget is None)` is redundant and can be removed.

mgeier: Yeah, now I'm seeing it, thanks for the clarification!

tk0miya:
Merged. Thank you for your work.

PR #10081, opened 2022-01-10
use 'callable' to check if object is callable (B004)

danieleades:
from the [bugbear](https://pypi.org/project/flake8-bugbear/) docs-

> B004: Using hasattr(x, '__call__') to test if x is callable is unreliable. If x implements custom __getattr__ or its __call__ is itself not callable, you might get misleading results. Use callable(x) for consistent results.

also, it looks better.


tk0miya:
Cool!

PR #10083, opened 2022-01-10
use class-style syntax for 'NamedTuple's

danieleades:
the class-style syntax is more readable, and more consistent with other objects


tk0miya:
Absolutely!

PR #10089, opened 2022-01-11
py domain: Ignore aliases for resolving `:any:` cross-references

### LINKED ISSUE #9577: "More than one target found" when all targets actually resolve to the same point
### Describe the bug

Sphinx emits a "More than one target found" warning when resolving `:any:` crossrefs even if all targets actually point to the same place, which can happen e.g. if documenting a class implemented in a private module but reexported in a public one.

### How to Reproduce

In an empty directory:
```sh
mkdir pkg
echo 'from ._implementation import Foo; __all__ = ["Foo"]' >pkg/__init__.py
echo 'class Foo: """The Foo."""' >pkg/_implementation.py
sphinx-apidoc . -o doc -F -A me -V 0.0
echo $'\nThe `pkg.Foo` is also the `pkg._implementation.Foo` and also the `Foo`.' >>doc/index.rst
(cd doc; PYTHONPATH=.. make O=-Ddefault_role=any html)
```
emits
```
.../doc/index.rst:23: WARNING: more than one target found for 'any' cross-reference 'Foo': could be :py:class:`pkg.Foo` or :py:class:`pkg._implementation.Foo`
```
even though there's only one entry for pkg.Foo in the docs (and pkg._implementation.Foo also points there).

### Expected behavior

No warning.

### Your project

N/A, see script above to generate it.

### Screenshots

_No response_

### OS

linux

### Python version

3.9

### Sphinx version

4.1.2

### Sphinx extensions

autodoc

### Extra tools

_No response_

### Additional context

_No response_

### LINKED ISSUE #10088: Sphinx 4.x + :any: + autodoc + aliasing objects = warning
### Describe the bug

Let's say you have a module:
```python
# mypackage/module.py
class A:
    pass
```
And expose it in your `__init__.py`:
```python
from .module import A
__all__ = ["A"]
```
In your doc, you want to expose the "official" api:
```rst
.. autoclass:: mypackage.A
```
And reference it:
```rst
Use the :any:`A` class
```
Alternatively, in `conf.py` use `default_role = "any"` and in your doc:
```rst
Use the `A` class
```

Then autodoc will output an warning:
```
more than one target found for 'any' cross-reference 'A': could be :py:class:`mypackage.A` or :py:class:`myackage.module.A`
```
And if building with `-W`, this warning will be an error.
I believe there's no way to avoid this warning without abandoning `:any:` altogether.

### How to Reproduce

```
$ git clone git@github.com:ewjoachim/sphinx-bug-any-canonical.git
$ cd sphinx-bug-any-canonical
$ cd docs
$ make SPHINX_OPTS=-W html
<build fails>
```

### Expected behavior

There used to be a way for me to simultaneously:
- Use `:any:`
- Define the API of my project in `__init__.py` while having the real classes live elsewhere
- use `autodoc`
- build with `-W`

without the combination of those things failing the build. I can probably rewrite a lot of documentation to use explicit `py:obj:` classes instead of any but it would be a lot of work, and I'm not sure it would make my doc more readable.

### Your project

https://github.com/ewjoachim/sphinx-bug-any-canonical

### Screenshots

_No response_

### OS

Linux

### Python version

3.8, 3.9, 3.10

### Sphinx version

latest stable: 4.3.2

### Sphinx extensions

sphinx.ext.autodoc

### Extra tools

_No response_

### Additional context

I believe what happens is that:
- `sphinx/ext/autodoc/__init__.py ClassDocumenter.add_directive_header` we have
```python
        canonical_fullname = self.get_canonical_fullname()
        if not self.doc_as_attr and canonical_fullname and self.fullname != canonical_fullname:
            self.add_line('   :canonical: %s' % canonical_fullname, sourcename)
```
This means that the py:class directive that will be generated by autodoc will have a `:canonical:` option

- `sphinx/domains/python.py PyObject.add_target_and_index`
```python
        canonical_name = self.options.get('canonical')
        if canonical_name:
            domain.note_object(canonical_name, self.objtype, node_id, aliased=True,
                               location=signode)
```
If an object has a canonical option, both names will be added to the known objects.

This means that ``:any:`Class` `` will always end up confused, and issue a warning.

I don't know how I can solve this.

ewjoachim:
Subject: closes #10088 (EDIT: already closed as duplicate) and also closes #9577 I believe, since they are duplicates.
With this PR, we can use `:any:` again in Sphinx v4.x with autodoc & python aliases.
I believe that everything that used to work will still work

### Feature or Bugfix

- Bugfix

### Purpose
See #10088 
The issue is that that autodoc adds `canonical` (which is a good thing), and `any` doesn't play well with `aliased` objects (which is the bad thing we're solving here)

### Detail
What I did, it's all in the python domain code:
- If there are more than one matches when resolving any, we remove aliased nodes from matches.
  - either this brings the number of matches to 1 and it's all fine
  - or there's still more than one match and it means the issue is real.
- In the tests with the canonical option, I've added multiple `any` lookups, and the test checks that there is no warning.

Just to be sure, I've run the test before the fix and it's red, and after the fix, it's green.

I don't know if this qualifies for a "hotfix" (not sure).

### Relates
- #10088 #9577
- #9205 was the same issue for normal xref resolutions, while this solves `any` xref resolutions


ewjoachim:
This actually relates to https://github.com/sphinx-doc/sphinx/pull/9207 but for `any` refs

ewjoachim:
Hm, hey, anyone ?

ewjoachim:
@tk0miya if you have any lead as to what I can do to help this PR go forward ?

sphinx/domains/python.py
```diff
@@ -1413,6 +1413,11 @@ def resolve_any_xref(self, env: BuildEnvironment, fromdocname: str, builder: Bui
 
         # always search in "refspecific" mode with the :any: role
         matches = self.find_obj(env, modname, clsname, target, None, 1)
+
+        if len(matches) > 1:
+            # Remove all duplicated matches
+            matches = [(name, obj) for name, obj in matches if not obj.aliased]
+
         for name, obj in matches:
             if obj[2] == 'module':
```
AA-Turner: Skips a list comprehension step
```suggestion
        for name, obj in matches:
            if len(matches) > 1 and obj.aliased:
                # Skip duplicated matches
                continue

            if obj[2] == 'module':
```

A

AA-Turner:
Please update CHANGES for `5.x`

A

sphinx/domains/python.py
ewjoachim: Done (I've stored `len(matches) > 1` in a variable)

ewjoachim:
We should be all good now :)

ewjoachim:
Rebased after 5.0 release :)

ewjoachim:
(is it expected that the bug label has not been set on the PR ?)

AA-Turner (approved):

ewjoachim:
Rebased after 5.0.1. Also, see https://github.com/sphinx-doc/sphinx/pull/10515

AA-Turner (approved):

ewjoachim:
Oh, I see you added this to the 5.1.0 milestone, I'll update CHANGES so that it appears in the 5.1.0 section :)

CHANGES
```diff
@@ -32,6 +32,8 @@ Bugs fixed
 * #10520: HTML Theme: Fix use of sidebar classes in ``agogo.css_t``.
 * #6679: HTML Theme: Fix inclusion of hidden toctrees in the agogo theme.
 * #10566: HTML Theme: Fix enable_search_shortcuts does not work
+* #9577, #10088: Fix warning for duplicate Python references when using :any:
```
```suggestion
* #9577, #10088: py domain: Fix warning for duplicate Python references when
  using :any:
```

CHANGES
```diff
@@ -32,6 +32,9 @@ Bugs fixed
 * #10520: HTML Theme: Fix use of sidebar classes in ``agogo.css_t``.
 * #6679: HTML Theme: Fix inclusion of hidden toctrees in the agogo theme.
 * #10566: HTML Theme: Fix enable_search_shortcuts does not work
+* #9577, #10088: py domain: Fix warning for duplicate Python references when
+  using :any:
+  and autodoc
```
```suggestion
* #9577, #10088: py domain: Fix warning for duplicate Python references when
  using ``:any:`` and autodoc.
```

AA-Turner (approved):

AA-Turner:
Thanks @ewjoachim!

A

PR #10091, opened 2022-01-12
In translated docs, sort glossaries by translated terms

### LINKED ISSUE #9827: Glossary :sorted: directive and internationalization.
### Describe the bug

I have problem when translate python glossary : 
The [french glossary](https://docs.python.org/fr/3/glossary.html) is currently unsorted. Some tests adding `:sorted:` directive in **glossary.rst** apparently doesn't help. The `:sorted:`  directive for `.. :glossary:` seems to be applied before translating file. The result is the glossary apears unsorted in french.

link to original (french) discussion : https://discuss.afpy.org/t/ordre-alphabetique-du-glossaire/594

### How to Reproduce

```
$ git clone https://github.com/python/python-docs-fr
$ cd python-docs-fr
$ make
$ make serve
$ # open http://localhost:8000/glossary.html to check items order 
```

### Expected behavior

_No response_

### Your project

https://github.com/python/python-docs-fr

### Screenshots

_No response_

### OS

Linux Mint

### Python version

3.10

### Sphinx version

3.2.1

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

jeanas:
This is done by moving the sorting from the glossary directive to a
transform operating after the i18n transform.

Closes #9827

### Feature or Bugfix
- Bugfix

### Purpose

Translated glossaries were sorted according to alphabetical order of English terms, which makes no sense to a reader of the native language. An example can be found here: https://docs.python.org/fr/3/glossary.html. This sorts them according to translated terms instead. To this end, the glossary directive no longer sorts the entries itself. Instead, it sets a boolean attribute on the glossary node, and a transform happening after the i18n transform takes care of sorting glossaries with this attribute.

This is my first contribution to Sphinx and I'm still new to the code; please let me know if I've missed something.

### Relates

#9827

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -9,7 +9,6 @@
 """
 
 import re
-import unicodedata
 import warnings
 from copy import copy
 from typing import (TYPE_CHECKING, Any, Callable, Dict, Iterable, Iterator, List, Optional,
@@ -336,6 +335,7 @@ class Glossary(SphinxDirective):
     def run(self) -> List[Node]:
         node = addnodes.glossary()
         node.document = self.state.document
+        node['sorted'] = ('sorted' in self.options)
 
         # This directive implements a custom format of the reST definition list
         # that allows multiple lines of terms before the definition.  This is
@@ -400,9 +400,8 @@ class Glossary(SphinxDirective):
             was_empty = False
 
         # now, parse all the entries into a big definition list
-        items = []
+        items: List[nodes.definition_list_item] = []
         for terms, definition in entries:
-            termtexts: List[str] = []
             termnodes: List[Node] = []
             system_messages: List[Node] = []
             for line, source, lineno in terms:
@@ -416,7 +415,6 @@ class Glossary(SphinxDirective):
                                           node_id=None, document=self.state.document)
                 term.rawsource = line
                 system_messages.extend(sysmsg)
-                termtexts.append(term.astext())
                 termnodes.append(term)
 
             termnodes.extend(system_messages)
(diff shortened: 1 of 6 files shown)


jeanas:
(CC @fipaddict @JulienPalard.)

sphinx/domains/std.py
```diff
@@ -336,6 +335,7 @@ class Glossary(SphinxDirective):
     def run(self) -> List[Node]:
         node = addnodes.glossary()
         node.document = self.state.document
+        node.sorted = ('sorted' in self.options)
```
tk0miya: We usually use node-attributes to store data. How about using it instead?
```suggestion
        node["sorted"] = ('sorted' in self.options)
```

sphinx/transforms/__init__.py
```diff
@@ -404,6 +405,22 @@ def apply(self, **kwargs: Any) -> None:
                 info = r.groupdict()
             node.attributes.update(info)
 
+class GlossarySorter(SphinxTransform):
+    """Sort glossaries that have the ``sorted`` flag."""
+    # Must be done after i18n.
+    default_priority = 30
```
tk0miya: IMO, this transform has too much high priority. How about changing this to the middle priority 500?
https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_transform

tk0miya (changes requested):
LGTM with nits.

jeanas:
Thanks for the review @tk0miya, I'll address your comments soon. Do you have an idea why the CI is failing? All GitHub seems to tell me is just "this check failed" for a bunch of jobs without a log; did I miss it somewhere? Tests pass locally for me.

jeanas:
Oh, and the "Summaries" are

“GitHub Actions has encountered an internal error when running your job.”

tk0miya:
Now I restarted CI jobs manually. I hope it goes well.

jeanas:
Yes, it went a lot better. I've pushed an update addressing your comments and fixing flake8 as well as mypy checks.

tk0miya (approved):

tk0miya:
Thank you for your quick update! I'll merge this after CI passed.

tk0miya:

jeanas:
Thanks a lot for the review!

PR #10094, opened 2022-01-12
test: Follow the new HTML structure of docutils-0.18

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Since docutils-0.18, the HTML structure for citations and footnotes has
been changed.  This modifies our testcase to follow the new HTML structure.
- I suppose this is the last piece of #9777. After merging this, we can unpin the docutils-0.18.
- I'm not sure this is a breaking change or not. I need an advice from RtD team @ericholscher 
- refs: #9777

### DIFF
--- a/tests/test_build_html.py
+++ b/tests/test_build_html.py
@@ -361,8 +361,6 @@ def test_html4_output(app, status, warning):
     'index.html': [
         (".//meta[@name='hc'][@content='hcval']", ''),
         (".//meta[@name='hc_co'][@content='hcval_co']", ''),
-        (".//dt[@class='label']/span[@class='brackets']", r'Ref1'),
-        (".//dt[@class='label']", ''),
         (".//li[@class='toctree-l1']/a", 'Testing various markup'),
         (".//li[@class='toctree-l2']/a", 'Inline markup'),
         (".//title", 'Sphinx <Tests>'),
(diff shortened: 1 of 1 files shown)


ericholscher:
Thanks for the ping, I'll note this on our side and see how it might impact us.

mitya57:
This is changing only tests, how can it be a breaking change?

tk0miya:
The HTML structure of docutils has already changed between docutils 0.17 and 0.18. As a result, upgrading it might cause some CSS not to work. I was pointed out it's breaking changes on the last upgrade (maybe on 0.16 to 0.17).

mitya57:
Ok, so *unpinning docutils* may be a breaking change, but not *this PR*.

But readthedocs theme pins docutils itself, so this should not affect them:
https://github.com/readthedocs/sphinx_rtd_theme/blob/1.0.0/setup.py#L122

tk0miya:
This PR accepts the change of the HTML structure in docutils-0.18. ​But I'll rework if somebody votes this is a breaking and not desirable change.

mitya57:
Ah ok, I understand what you mean now.

agjohnson:
@tk0miya I see this is on the 4.5.0 milestone. When are you expecting a release candidate for this release? We can prioritize some deeper sphinx_rtd_theme testing ahead of that.

We do anticipate the docutils 0.18 change to introduce some backwards incompatible changes to the theme. Last we tested these seemed manageable, but we can do another round of tests. We are hoping to add some deprecations next release so that we can keep the latest theme release aligned with the latest Sphinx release.

tk0miya:
Basically, we'll release a new version in bi-months. So 4.5.0 will be released at the end of February or the beginning of March.

AA-Turner:
Is there anything blocking this PR @tk0miya? It would be nice to upgrade the pin on Docutils :)

A

tk0miya:
Okay, let's go ahead. Nobody objects to merging this.

PR #10108, opened 2022-01-16
Close #9456: html search: Add a config variable; html_show_search_summary

### LINKED ISSUE #9456: r634b05f3a5070fef98 breaks SQLAlchemy's search page, emits dozens of unnecessary HTTP requests
### Describe the bug

upgrading to the 4.1 series, SQLAlchemy's search page now has a bunch of tags that look like `<p class="context">` throughout them with ellipses inside with no functionality and it looks like these all **emit indivudal HTTP requests** that seem to do nothing, slowing down our search page.    

the change is exactly that made in 634b05f3a5070fef98.  If i edit this directly in searchtools.js, the old behavior is restored.    the change appears to now ignore an option that was in place before which explicitly states that the contextual files for this feature are not available.     I have also tried setting html_copy_source to True, hoping to at least see the contextual stuff that I see in other doc pages, but it doesn't work for my site anyway.

i think it's a bad idea to emit an unecessary ajax / network request that has no chance of being useful.    Can the old behavior please be restored?  this change forces unnecessary network overhead for a feature that's not in use and also breaks our layout (which I can work around wtih CSS, but the 95 extra network calls makes this a non -starter).

### How to Reproduce

```
$ git clone https://github.com/sqlalchemy/sqlalchemy
$ cd sqlalchemy/docs/build
$ make html
$ cd output/html
$ python -m http.server  # view docs on webserver
$ # navigate to http://localhost:8000
```

### Expected behavior

_No response_

### Your project

https://docs.sqlalchemy.org/

### Screenshots

good:

good

bad:

bad

### OS

linux

### Python version

3.9

### Sphinx version

4.1.1

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #9494, #9456

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,9 @@ Deprecated
 Features added
 --------------
 
+* #9494, #9456: html search: Add a config variable
+  :confval:`html_show_search_summary` to enable/disable the search summaries
+
 Bugs fixed
 ----------
 
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1351,6 +1351,13 @@ that use Sphinx's HTMLWriter class.
 
    .. versionadded:: 1.0
 
+.. confval:: html_show_search_summary
+
+   If true, the text around the keyword is shown as summary of each search result.
+   Default is ``True``.
+
+   .. versionadded:: 4.5
+
 .. confval:: html_show_sphinx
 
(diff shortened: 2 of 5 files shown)


Meem514 (commented):
How do you fail when it's a mock?

tk0miya:
What do you mean?

PR #10114, opened 2022-01-17
Fix #10110: sphinx-build: Emit builder-finished before shutdown

### LINKED ISSUE #10110: Warning handling for the build-finished event for -W
I'd like to issue warnings in `build-finished`. Currently, they aren't converted to errors with `-W` (ref: felix-hilden/sphinx-codeautolink#100). But my extension deals with the result HTML, so there's nowhere else that I can place the logic.

I'd like for `-W` to raise the warnings in all build steps as expected.

- Implement extension logic elsewhere: surely often possible, but unfortunately not in this case
- Keep not raising: again, it's not often that a cleanup step would warn, but I can see it being potentially useful in other scenarios as well. Although in most of those cases you'd just raise an error.

**Additional context**
I found the relevant source in [build](https://github.com/sphinx-doc/sphinx/blob/f38bd8e9529d50e5cceffe3ca55be4b758529ff7/sphinx/application.py#L326). The logic looks solid, because `build-finished` should be emitted regardless of error state. And if you'd like to have the warning summary messages before that like it is currently without including `build-finished`, it's completely fine by me. I just want them to be converted to errors.

Also, feel free to just close this if you're keen on keeping the current behavior 😄 But I'd like to know why!

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- The error on the builder-finished event has been ignored for the
calculation of the exit code.  This emits the event earilier step to be
calculated correctly.
- refs: #10110


sphinx/application.py
```diff
@@ -370,8 +372,6 @@ def build(self, force_all: bool = False, filenames: List[str] = None) -> None:
                 os.unlink(envfile)
             self.events.emit('build-finished', err)
```
felix-hilden, line 337: Do we want to emit twice in case there was an error? Meaning if `build-finished` emits an error, is it emitted again with the error information?

felix-hilden: Maybe the status/warncount code could be moved outside the try statement instead, so that `build-finished` is emitted only once but errors are still logged afterwards.

tk0miya: I think the following code will not raise an error. So I moved them outside of try-block now.
Thank you for your comment.

felix-hilden (commented):
I think this looks good! Thank you so much for taking on this issue 🙏 Looking forward to seeing it live 👍

Issue #10073, opened 2022-01-09
Support for "data" URIs in ImageConverter.guess_mimetypes()


mgeier:
**Is your feature request related to a problem? Please describe.**

This problem came up when using the `sphinxcontrib.rsvgconverter` extension, which is supposed to automatically convert SVG images to PDF when using the LaTeX builder, which it does with the help of the external program `rsvg-convert`.

When using the HTML builder, the extension is supposed to do nothing.
Previously, there was a problem with this (https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter/issues/8), which was fixed in #8649.

However, there is still one error case remaining:

When using `data` URIs instead of image files, this is still broken.
The data type is not recognized in this case and therefore it is handled as unsupported by the builder, even if the builder actually supports it.

Example reST code:

```rst
[...]

PR #10118, opened 2022-01-18
ImageConverter: ignore '?' image URIs

mgeier:
While #10073 has fixed my minimal example, it wasn't enough to fix my original problem https://github.com/spatialaudio/nbsphinx/pull/559.

It turns out that on top of `data:` URIs, remote URIs have also been automatically handled as "not supported" and therefore unnecessarily invoked the image converters.

This PR ignores all `'?'` images, which includes `data:` URIs, therefore I've reverted the changes from  #10099, which are no longer necessary.

I still don't know what `'?'` means exactly, it would probably be nice to add a comment explaining what that is and why it's ignored.

Example project for experiments: https://github.com/mgeier/rsvgconverter-bug

### DIFF
--- a/sphinx/transforms/post_transforms/images.py
+++ b/sphinx/transforms/post_transforms/images.py
@@ -197,11 +197,10 @@ class ImageConverter(BaseImageConverter):
     def match(self, node: nodes.image) -> bool:
         if not self.app.builder.supported_image_types:
             return False
+        elif '?' in node['candidates']:
+            return False
         elif set(self.guess_mimetypes(node)) & set(self.app.builder.supported_image_types):
             # builder supports the image; no need to convert
-            return False
-        elif node['uri'].startswith('data:'):
-            # all data URI MIME types are assumed to be supported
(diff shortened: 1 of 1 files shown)


tk0miya:
I consider the `ImageConverter` should handle both local and remote images. But, essentially, current implementation mainly supports local images only. So this proposal is a correct approach.

Note: One of my extension, sphinxcontrib-cacoo supports remote images by overriding `match()` method.
https://github.com/sphinx-contrib/cacoo

tk0miya:
> I still don't know what '?' means exactly, it would probably be nice to add a comment explaining what that is and why it's ignored.

`'?'` represents "nonlocal" images.

https://github.com/sphinx-doc/sphinx/blob/1e1f8e218046cbcda69be1f58234324fc48c7db4/sphinx/environment/collectors/asset.py#L39-L42

mgeier:
Thanks for merging, and thanks for the additional information, that's very helpful!

PR #10123, opened 2022-01-20
extlink: Add support for suppressing 'hardcoded link' warnings

progval:
Subject: Add support for suppressing 'hardcoded link' warnings

### Feature or Bugfix

- Bugfix

### Purpose
These warnings were added in v4.4.0, but some projects may prefer to hide them,
eg. if hardcoded links appear in docstrings.

### Relates
- [PR 9800](https://github.com/sphinx-doc/sphinx/pull/9800), which introduced these warnings

### DIFF
--- a/sphinx/ext/extlinks.py
+++ b/sphinx/ext/extlinks.py
@@ -76,7 +76,8 @@ class ExternalLinksChecker(SphinxPostTransform):
                 msg = __('hardcoded link %r could be replaced by an extlink '
                          '(try using %r instead)')
                 replacement = f":{alias}:`{match.groupdict().get('value')}`"
-                logger.warning(msg, uri, replacement, location=refnode)
+                logger.warning(msg, uri, replacement, location=refnode,
+                               type="extlink", subtype="hardcoded")
 
 
 def make_link_role(name: str, base_url: str, caption: str) -> RoleFunction:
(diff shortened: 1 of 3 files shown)


sphinx/ext/extlinks.py
```diff
@@ -76,7 +76,8 @@ def check_uri(self, refnode: nodes.reference) -> None:
                 msg = __('hardcoded link %r could be replaced by an extlink '
                          '(try using %r instead)')
                 replacement = f":{alias}:`{match.groupdict().get('value')}`"
-                logger.warning(msg, uri, replacement, location=refnode)
+                logger.warning(msg, uri, replacement, location=refnode,
+                               type="extlink", subtype="hardcoded")
```
jeanas, line 80: (I'm no maintainer, just someone lurking around.)

I think you also need to update the list of warnings in the documentation of `suppress_warnings` here:

https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings

jeanas: Never mind, I'm blind. Sorry.

tk0miya:
I think it's better to disable the detector instead of suppressing warnings. So I'll adopt #10126 instead of this.

PR #10127, opened 2022-01-22
[WIP] intersphinx, allow an inventory to be a self ref

### LINKED ISSUE #9169: Resolve intersphinx to own project as current build

Motivated from https://github.com/astropy/astropy/pull/11690

Take packages A and B, with package A as the core package and package B as an affiliate package.

When B subclasses a class from package A, docstrings can be inherited from the A class. If the docstring contains a reference, e.g.``:doc:`/path/to` ``, then this link will incorrectly point to the documentation in the B package, not in the A package.

One option is to make the links intersphinx links, e.g. ``:doc:`A:/path/to` ``. While this is useful for the B package, now the link is broken for the A package, or at best refers to an old version of the docs: whatever is specified in the intersphinx mapping.

Allow intersphinx references to the package being built to be "stripped" of the intersphinx portion and be resolved as local links. In the example above, ``:doc:`A:/path/to` `` -> ``:doc:`/path/to` `` when in package A. Now both packages A and B can use the same links. B gets an intersphinx link, A gets a link to the current build.

In https://github.com/astropy/astropy/pull/11690 we implemented a bit of a hack by setting a function with higher priority than intersphinx to change the link types. While this works, a built-in solution would be better.

**Additional context**

- https://github.com/astropy/astropy/pull/11690

jakobandersen:
### Feature or Bugfix
- Feature

### Purpose
A resolution for https://github.com/sphinx-doc/sphinx/issues/9169 where a new config variable ``intersphinx_self_mapping`` can be used to specify an inventory name which will make corresponding ``:external+thatName:domain:role:`` references to fall back as if the user wrote ``:domain:role:``.
This is useful for Python projects where a docstring can be inherited from an external project. That external project would need to write ``:external:`` references in those docstrings and use this self-mapping mechanism.

TODOs:
- Name of config variable?
- Does it need an ``env_version`` bump?
- Documentation
- CHANGES entry

### Relates
Fixes https://github.com/sphinx-doc/sphinx/issues/9169

### DIFF
--- a/sphinx/ext/intersphinx.py
+++ b/sphinx/ext/intersphinx.py
@@ -497,10 +497,14 @@ class IntersphinxRole(SphinxRole):
     def run(self) -> Tuple[List[Node], List[system_message]]:
         assert self.name == self.orig_name.lower()
         inventory, name_suffix = self.get_inventory_and_name_suffix(self.orig_name)
-        if inventory and not inventory_exists(self.env, inventory):
-            logger.warning(__('inventory for external cross-reference not found: %s'),
-                           inventory, location=(self.env.docname, self.lineno))
-            return [], []
+        self_map = self.env.config.intersphinx_self_mapping
+        is_self_ref = self_map is not None and self_map == inventory
+
+        if not is_self_ref:
+            if inventory is not None and not inventory_exists(self.env, inventory):
+                logger.warning(__('inventory for external cross-reference not found: %s'),
+                               inventory, location=(self.env.docname, self.lineno))
+                return [], []
 
         role_name = self.get_role_name(name_suffix)
         if role_name is None:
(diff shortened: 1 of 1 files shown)


tk0miya:
I can understand the case. But I feel very strange to annotate "local" reference as an "external" reference... I'm not sure this should be implemented as a built-in feature.

jakobandersen:
> I can understand the case. But I feel very strange to annotate "local" reference as an "external" reference... I'm not sure this should be implemented as a built-in feature.

It is indeed a bit of a hax. Ideally the root of the problem, that docstrings are copied from external projects, should be addressed directly.

AA-Turner:
Closing in favour of #13291.

A

PR #10128, opened 2022-01-22
[WIP] intersphinx, deprecate old explicit inventory specification

jakobandersen:
### Feature or Bugfix
- Feature?

### Purpose
Now that we have the ``external`` role for intersphinx (https://github.com/sphinx-doc/sphinx/pull/9822) I think we should deprecate the old syntax.
This PR will warn by default, but there is a configuration variable to disable these warnings to aid in the transition.
As an initial suggestion, how about the following deprecation scheme?
- The configuration variable to silence the warnings will be deprecated immediately and removed in v6 or v7 (i.e., 2023).
- The actual old syntax will be removed in a major version or two afterwards.
If I remember correctly then some projects (e.g., the Linux kernel) compiles with the version they ship, which can be rather old, so we need of research to figure out when exactly we can remove the syntax.

@tk0miya, any thoughts?

### Detail

TODOs if this should go forward:
- Name of config variable.
- Documentation
- CHANGES entry.

### DIFF
--- a/doc/internals/release-process.rst
+++ b/doc/internals/release-process.rst
@@ -76,12 +76,12 @@ Sphinx 2.x:
 
 * Sphinx 2.x will contain a backwards-compatible replica of the function
   which will raise a ``RemovedInSphinx40Warning``.
-  This is a subclass of :exc:`python:PendingDeprecationWarning`, i.e. it
-  will not get displayed by default.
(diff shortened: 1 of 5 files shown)


AA-Turner:
Currently I don't think this is worth the disruption, but we could revisit it later.

A

PR #10132, opened 2022-01-23
Fix #10122: make.bat should check the installation before help

### LINKED ISSUE #10122: make.bat template jumps to help block without first checking if sphinx is installed
### Describe the bug

If you were to invoke `make.bat`, before installing sphinx, you would get the following error:
```
'sphinx-build' is not recognized as an internal or external command,
operable program or batch file.
```

The reason why the above error is emitted and not the `echo.The 'sphinx-build' command was not found. Make sure you have Sphinx[...]` message is because `if "%1" == "" goto help` is called before checking if the app is installed.

This error is at line 13, in file sphinx/sphinx/templates/quickstart/make.bat.new_t.

Line 13 should be moved to after the app check, `if errorlevel 9009 ([...])`.

### How to Reproduce

As an end user, I receive a folder with sphinx generated files, e.g. make.bat, conf.py, Makefile, and attempt to run make by calling `make.bat`

### Expected behavior

Error message stating that I should install sphinx

### Your project

NA

### Screenshots

_No response_

### OS

Win

### Python version

3.8.7

### Sphinx version

NA

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #10122

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,8 @@ Bugs fixed
 ----------
 
 * #10133: autodoc: Crashed when mocked module is used for type annotation
+* #10122: sphinx-build: make.bat does not check the installation of sphinx-build
+  command before showing help
 
 Testing
 --------
--- a/sphinx/templates/quickstart/make.bat.new_t
+++ b/sphinx/templates/quickstart/make.bat.new_t
@@ -10,8 +10,6 @@ if "%SPHINXBUILD%" == "" (
 set SOURCEDIR={{ rsrcdir }}
 set BUILDDIR={{ rbuilddir }}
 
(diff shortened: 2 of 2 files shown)


CHANGES
```diff
@@ -15,6 +15,8 @@ Features added
 
 * #9494, #9456: html search: Add a config variable
   :confval:`html_show_search_summary` to enable/disable the search summaries
+* #10122: sphinx-build: make.bat does not check the installation of sphinx-build
```
tk0miya: This must be placed at "bugs fixed" category.

PR #10140, opened 2022-01-28
Docs: Fix typos in autosummary.rst

tcmetzger:
Subject: Fix two small typos I happened to find while reading autosummary.rst

### Feature or Bugfix
- Bugfix

### Purpose
When reading autosummary.rst, I found two typos (a missing word and an extra word). This PR fixes those two typos.

### Detail
* Add "are" to `The following variables are available in the templates:`
* Remove superfluous "here" from `Here, "public" means that the name does not start with an underscore.`


tk0miya:

PR #10147, opened 2022-01-30
Apply `autodoc_typehints_format` to all type hints

zuqq:
In using the new `autodoc_typehints_format` configuration option (thanks a lot for adding it!), I noticed that it doesn't apply to type hints on certain kinds of items (e.g., properties).

This patch changes the behavior of `autodoc_typehints_format`: it now also applies to attributes, data, properties, and type variable bounds. After this change, all calls to `stringify_typehint` in `sphinx/ext/autodoc/__init__.py` take `self.config.autodoc_typehints_format` into account.


tests/roots/test-ext-autodoc/target/typehints.py
```diff
@@ -1,7 +1,13 @@
-from typing import Any, Tuple, Union
+import pathlib
+from typing import Any, Tuple, TypeVar, Union
 
 CONST1: int
+#: docstring
```
zuqq, line 5: I added these docstrings because the items are not picked up by autodoc otherwise.

tk0miya:
Thank you for your proposal. At first glance, this is a good improvement. To go forward this, I'll review this in a few days. Please wait for a moment.

>I added the change to the 4.4.1 section of CHANGES, with the argument that the existing behavior is unintended.

4.4.x releases are only used for bug fixes. So improvements will be released on 4.5.0 as a "new feature".

zuqq:
> 4.4.x releases are only used for bug fixes. So improvements will be released on 4.5.0 as a "new feature".

Ah, I see; I've moved it there.

bryanforbes:
Should this PR include the call to `stringify` in `sphinx.ext.napoleon.docstring`?

zuqq:
> Should this PR include the call to `stringify` in `sphinx.ext.napoleon.docstring`?

Good catch! I don't personally use `sphinx.ext.napoleon`, so I didn't check that part of the code.

It looks like making that change is a little more involved: `GoogleDocstring` and `NumpyDocstring` are not guaranteed to have access to the Sphinx config (because `None` is allowed in `__init__`), but it seems like they do in practice (see `sphinx.ext.napoleon._process_docstring`). So maybe a separate PR?

zuqq:
@tk0miya I rebased; could you please review this PR before the next release?

tests/roots/test-ext-autodoc/target/typehints.py
zuqq: Ah, that's the long-standing #1063.

tk0miya:
Sorry for my late response. Looks very nice. So I'm merging this into the 5.x branch. It will be released as 5.0 soon. Thank you for your great work!

PR #10150, opened 2022-01-30
Close #10146: autodoc: autodoc_default_options does not support `no-value` option

### LINKED ISSUE #10146: Make autodoc abbreviate long dictionaries
Examples in the wild:

| [`babel.core.Locale.negotiate`](https://babel.pocoo.org/en/latest/api/core.html#babel.core.Locale.negotiate) | [`searx.engines.engine_shortcuts`](https://docs.searxng.org/src/searx.engines.html#searx.engines.engine_shortcuts)
|--|--|
|image | image

I think for long dictionaries it would be better if they were automatically abbreviated, or there at least should be an option to do so.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #10146

### DIFF
--- a/doc/usage/extensions/autodoc.rst
+++ b/doc/usage/extensions/autodoc.rst
@@ -528,7 +528,8 @@ There are also config values that you can set:
    The supported options are ``'members'``, ``'member-order'``,
    ``'undoc-members'``, ``'private-members'``, ``'special-members'``,
    ``'inherited-members'``, ``'show-inheritance'``, ``'ignore-module-all'``,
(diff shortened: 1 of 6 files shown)


doc/usage/extensions/autodoc.rst
```diff
@@ -541,6 +542,9 @@ There are also config values that you can set:
    .. versionchanged:: 4.1
       Added ``'class-doc-from'``.
 
+   .. versionchanged:: 4.5
+      Added ``'class-doc-from'``.
```
tk0miya:
```suggestion
      Added ``'no-value'``.
```

Issue #10154, opened 2022-02-01
Characters for section headings are not displaying properly in basics.html 


noahsprent:
### Describe the bug

What was I trying to do?
I wanted to know how to represent different levels of headings. On the basics page, [there is a section](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections) about which characters Python's style guide recommends for representing different levels of headings in .rst.

What did I expect to happen:
In the [.rst file preview on GitHub](https://github.com/sphinx-doc/sphinx/blob/master/doc/usage/restructuredtext/basics.rst#sections) and the [Python style guide page](https://devguide.python.org/documenting/#sections) the characters that it is recommended be used for these levels appear in a box that makes it clear the following comma is not to be included.

What happened:
In the page in the sphinx docs the characters are not surrounded by a box (see screenshot). This might lead people (me) to (perhaps stupidly) assume that the comma is part of the recommendation. e.g. the headings should look like:

```
Subheading
-,-,-,-,-,-,-,-,-,
```

Rather than:
```
Subheading
----------------
```

### How to Reproduce

Visit https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

### Your project

https://www.sphinx-doc.org/

### Screenshots

![Screenshot 2022-02-01 09 05 33](https://user-images.githubusercontent.com/39567418/151941817-e5c0fdec-eed5-41d3-8eb5-940cc639181d.png)

### OS

Chrome OS Version 98.0.4758.67

### Python version

Not relevant

### Sphinx version

Not relevant

jeanas:
Not having a box around inline code is part of the general theme; I'm not sure one would want to change it just for this. I'm a random lurker, not a maintainer, but here is my advice: how about submitting a PR yourself to fix the issue? That will be just as fast as reporting the problem. For example, you could include the characters in quotes (e.g., `"``=``"` in the reST source) so that they are clearly separated.

noahsprent:
Thanks for the information on the theme, I didn't realise that was the case and thought it might be a more complex issue. I'm happy to submit a PR as it's quite a simple issue, but as I am not involved in the project at all have no idea whether there would be preferences from the core team about the way in which to resolve it. As one of the characters is themselves a quote it might be confusing to surround the characters in quotes. I think the easiest might be to simply remove the commas.

jeanas:
[...]

PR #10156, opened 2022-02-02
Remove commas when describing characters for section headings

noahsprent:
Subject: Remove commas when describing characters for section headings in docs

### Feature or Bugfix
- Bugfix

### Purpose
In the .html of the docs the theme does not display inline code and therefore it is a little confusing which characters to use for denoting section headings. I suggest removal of the commas here, but am very happy for this to be rejected in favour of a different solution. Just trying to be helpful by opening the request. 

### Detail
- No more detail necessary I don't think? 

### Relates
- Related to issue #10154

### DIFF
--- a/doc/usage/restructuredtext/basics.rst
+++ b/doc/usage/restructuredtext/basics.rst
@@ -245,10 +245,10 @@ follow:
 
 * ``#`` with overline, for parts
 * ``*`` with overline, for chapters
-* ``=``, for sections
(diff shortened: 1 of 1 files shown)


jeanas (approved):
Sounds good to me.

tk0miya:
Absolutely!

tk0miya:
Thanks!

PR #10169, opened 2022-02-06
Fix #9529: LaTeX: named footnotes are converted to "?"

### LINKED ISSUE #9529: sphinxfootnotemark is not working correctly after update sphinx >=3.5.0
### Describe the bug

After sphinx update to version **3.5.0 or higher**, only first footnote is functional after converting rst > tex > pdf. Other references to the same footnote are translated as question mark.

As little example after .rst files are translated to .tex it looks like this:
```
Some text
\begin{footnote}[1]\sphinxAtStartFootnote
URL: \sphinxurl{https://access.redhat.com/documentation/en-us/red\_hat\_enterprise\_linux/8/html-single/upgrading\_from\_rhel\_7\_to\_rhel\_8/index}.
%
\end{footnote}.

Some other text
\sphinxfootnotemark[1].
```

But when you translate it from latex to .pdf, output looks like:
Some text<sup>1</sup>.
Some other text<sup>?</sup>.

### How to Reproduce

```
$ git clone https://github.com/xstodu07/sphinx-issues
$ cd sphinx-issues
$ pip install tox
$ # tox - Create virtualenv with dependencies in requirements.txt, build latex/*.pdf and html/
$ # OR create own virtualenv, activate it (for example: source .tox/docs/bin/activate) and build latex using 'make pdflatex' from 'docs/' directory
$ tox
$ # check docs/_build/latex/footnotes.pdf
```

### Expected behavior

Some text<sup>1</sup>.
Some other text<sup>1</sup>.

### Your project

https://github.com/xstodu07/sphinx-issues

### Screenshots

_No response_

### OS

Linux RHEL 7.9

### Python version

3.6

### Sphinx version

3.5.0 or higher

### Sphinx extensions

sphinxcontrib.bibtex
 - has no effect, if this extension is used or not, same happened to me when footnotes are used in original way - this extension probably use named labels (see first comment below)

### Additional context

I noticed that both sphinx versions (old and  version >=3.5.0) use different packages for footnotes, but dont understand much the syntax of latex .sty files:
```
# sphinx4.0.3
Package: sphinxpackagefootnote
%% [2021/02/04 v1.1d footnotehyper adapted to sphinx (Sphinx team)]
%% Version: based on footnotehyper.sty 2021/02/04 v1.1d
(...)
# sphinx2.3.0
[2017/10/27 v1.7 hyperref aware footnote.sty for sphinx (JFB)]
%% Package: footnotehyper-sphinx
%% Version: based on footnotehyper.sty 2017/03/07 v1.0
(...)

Footnotes are working correctly only with the older version [2017/10/27 v1.7 hyperref aware footnote.sty for sphinx (JFB)]. Of course, only without this functionality:
* #2616 (fifth item): LaTeX: footnotes from captions are not clickable,
  and for manually numbered footnotes only first one with same number is
  an hyperlink

And also, it looks like the problem is between translating latex files to pdf (sphinx generate bad .sty file for footnotes). Because generated '.tex' files are same in use of both sphinx versions.

### PR DESCRIPTION
**Pull request by tk0miya**
### Feature or Bugfix
- Bugfix

### Purpose
- Named auto numbered footnote (ex. ``[#named]``) that is referred
    multiple times was rendered to a question mark.
-   This calls `\sphinxstepexplicit` for every footnote node that is
    referred multiple times.
- refs: #9529 
- including https://github.com/sphinx-doc/sphinx/pull/10168

### CODE REVIEW

**Review by jfbu — APPROVED** [review]
Indeed you are right that what was needed was to know which footnotes are multiply referred too, so LGTM!

small nits: The macro `\sphinxstepexplicit` is now a bit misnamed, let's however stick with its current name.  But at some point a comment in `texinputs/sphinx.sty` may need update like this:
```diff
index 4d42199a7..6c9f1606e 100644
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -300,9 +300,9 @@
 % Support scopes for footnote numbering
 \newcounter{sphinxscope}
 \newcommand{\sphinxstepscope}{\stepcounter{sphinxscope}}
-% Explicitly numbered footnotes may be referred to, and for this to be
-% clickable we need to have only one target. So we will step this at each
-% explicit footnote and let \thesphinxscope take it into account
+% Some footnotes are multiply referred-to.  For unique hypertarget in pdf,
+% we need an additional counter.  It is called "sphinxexplicit" for legacy
+% reasons as "explicitly" numbered footnotes may be multiply referred-to.
 \newcounter{sphinxexplicit}
 \newcommand{\sphinxstepexplicit}{\stepcounter{sphinxexplicit}}
 % Some babel/polyglossia languages fiddle with \@arabic, so let's be extra
```
This extra patch may be postponed to later occaion.

*edited*: what follows was based on some confused memory, the situation described below with multiple sources may still lead to wrong hyperlinks even with this PR (related #10175).

I have noticed that this PR gives correct hyperlinks even in situation where named footnotes originate in multiple documents, and so may use for example both ``1`` as footnote mark, which in pdf single document output may be confusing if on same page, but at least with this PR the hyperlink is ok. For example 
```
========== ===
Test [#a]_
========== ===
   1       One
========== ===

.. [#a] Footnote one.
```
in document `doc1.rst` and
```
========== ===
Test [#a]_
========== ===
   1       Two
========== ===

.. [#a] Footnote two.
```
in `doc2.rst`, both in toctree of `index.rst`. In future there is potential using the `sphinxscope`  apparatus in `sphinx.sty` to create in PDF a disambiguating footnote mark, but at any rate already hyperlink will point to correct footnote, even if on same page in PDF.

tk0miya:
Thank you for reviewing. Updated the comment now.

Let's discuss the problem with multiple sources in https://github.com/sphinx-doc/sphinx/issues/10175.

Thanks as always!

PR #10170, opened 2022-02-06
The latest docutils converts :pep: role to the "https" URL

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- docutils-0.19 (unreleased) generates "https" URL from :pep: role.
- refs: https://github.com/sphinx-doc/sphinx/runs/5080565975?check_suite_focus=true


AA-Turner (approved):

AA-Turner:
I'd perhaps add a comment that the `?` can be dropped with Docutils 0.18 support, but apart from that looks good.

A

PR #10197, opened 2022-02-15
Reduce `body_min_width` in basic theme

palewire:
The current 450px setting is too large for many small device width. It causes x-overflows that make the page appear broken and introduces difficulty reading on many phones.


sphinx/themes/basic/theme.conf
```diff
@@ -7,7 +7,7 @@ sidebars = localtoc.html, relations.html, sourcelink.html, searchbox.html
 [options]
 nosidebar = false
 sidebarwidth = 230
-body_min_width = 450
+body_min_width = 300
```
tk0miya: Is any reason to choose this value?

I just searched the screensizes of well-used mobile devices. Some reports says 360px is smallest size at present.
https://www.browserstack.com/guide/ideal-screen-sizes-for-responsive-design
https://mediag.com/blog/popular-screen-resolutions-designing-for-all/

palewire: I picked it because it solved the bug. Another value could work fine if it also resolve the issue.

mgeier:
For reference, this has been added in  #4376 by @timhoffm.

tk0miya:
I determined to use 360px for the new default value of `body_min_width` (see reports above).

tk0miya:

PR #10203, opened 2022-02-16
fix Bases: %s msgid for RU lang

MarshalX:
Subject: fix wrapping of bases with blockqoute in RU locale

### Feature or Bugfix
- Bugfix

### Purpose
Sphinx adding blockquite without any reasons as a wrapper for bases. There is extra spaces in gettext locales for RU lang.

### Detail

Screenshots, reproducer and so on here: https://github.com/MarshalX/sphinx-bases-bug


tk0miya:
Thank you for your contribution. The change in transifex.com will be merged by weekly task automatically. So please wait for a moment.

PR #10208, opened 2022-02-19
Close #10160: Remove headings from module docstrings

### LINKED ISSUE #10160: Follow PEP 257 docstring conventions
[PEP 257](https://www.python.org/dev/peps/pep-0257/) establishes conventions for docstrings, in particular:

>Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

Sphinx currently violates that convention by formatting its module docstrings as follows:

```
"""
    sphinx.builders
    ~~~~~~~~~~~~~~~

    Builder superclass for all builders.

    :copyright: Copyright 2007-2022 by the Sphinx team, see AUTHORS.
    :license: BSD, see LICENSE for details.
"""
```

Note that the summary is not in the first line but after a heading and a blank line. This is problematic because when e.g. browsing the Sphinx API via [Pydoctor](https://github.com/twisted/pydoctor), Pydoctor just shows the headings (since it assumes the PEP 257 convention):

image

So it would be great if Sphinx could follow PEP 257 by removing such headings from its module docstrings.

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- So far, our module docstings starts with headings for the module name.
But it's redundant and not helpful. So this removes them all from our
codebase.
- refs: #10160


tk0miya:
I'll apply #10160 instead of this. So withdrawing now.

Issue #10227, opened 2022-03-01
Duplicate search results 


calebchiam:
### Describe the bug

Certain multi-term search queries will generate duplicate sets of search results. 

This is due to how `performObjectSearch` is written, where the criteria for result inclusion is if the `object` term appears in the `fullname` and each of the `otherTerms` appears in the `haystack`. Since both `fullname` and `haystack` contain the `name` as a substring, a two-term search query that has terms that are both substrings of `name` will result in `performObjectSearch` outputting the same set of results for each term. 

As there is no code to guard against duplication of results, the duplicate results show up in the final output.

### How to Reproduce

Search 'builders changes' on the [Sphinx docs](https://www.sphinx-doc.org/en/master/search.html?q=builders+changes). Duplicate results appear at the bottom of the page, see attached screenshot.

In general, inspecting the `objects` index to get an object name and decomposing a name into two terms will result in the same error.

### Expected behavior

Each search result should appear only once.

### Your project

https://github.com/sphinx-doc/sphinx

### Screenshots

**Searching 'builders changes':**
<img width="886" alt="image" src="https://user-images.githubusercontent.com/14286996/156212697-172e9606-2da5-4520-af37-4660066604da.png">

**Inspecting the `objects` index:**
![image](https://user-images.githubusercontent.com/14286996/156212938-9337aa14-72d4-4539-96c4-0ce399629deb.png)

**The outputs from each of the object and terms search method calls:**
![image](https://user-images.githubusercontent.com/14286996/156213110-7c84d0f9-e6b6-46f8-acc8-70b98e94b89c.png)

### OS

Mac

### Python version

3.9

### Sphinx version

4.4.0

### Additional context

Relevant previously closed issue: https://github.com/sphinx-doc/sphinx/pull/3911

PR #3911, opened 2017-07-05
Filter out duplicate search results

adamgreg:
Subject: Filter out duplicate search results

### Feature or Bugfix
Bugfix

### Purpose
When performing a search using the searchtools.js, if the query includes multiple terms then this can lead to duplicate partial results being added by performObjectSearch().

### Detail
This pull request skips duplicate results in displayNextItem() by checking if each result is identical to the last one in the sorted list.

PR #10228, opened 2022-03-01
Fixing duplicate search results bug

calebchiam:
### Feature or Bugfix
- Bugfix

### Purpose
- See #10227 

### Detail

This fix instantiates a hash set to check for whether a result entry has come up previously before including it in the final search result array. Each entry is uniquely represented by a concatenated string of its contents, excluding its score. 

Notice that we reverse the array before doing this check, so that in the case of duplicates, the entry with the highest score is kept. 

### Relates
- #10227 
- #3911

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -238,6 +238,17 @@ var Search = {
       }
     });
 
+    // remove duplicate search results
+    var seen = new Set();
+    results = results.reduce((acc, result) => {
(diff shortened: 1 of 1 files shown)


tk0miya:
Thank you for your contribution. But we fully rewrote searchtools.js in the master branch. So it would be better to change it on the branch instead of 4.x branch. Could you apply this bug fix to it, please?

calebchiam:
Sure, I'll take a look and make a PR soon.

tk0miya:
I decided to fix this in the 5.0 release (https://github.com/sphinx-doc/sphinx/pull/10236). So I closing this.

PR #10236, opened 2022-03-05
Fixing duplicate search results bug (5.0)

calebchiam:
### Feature or Bugfix
- Bugfix

### Purpose
- See #10227 

### Detail

This fix instantiates a hash set to check for whether a result entry has come up previously before including it in the final search result array. Each entry is uniquely represented by a concatenated string of its contents, excluding its score.

Notice that we reverse the array before doing this check, so that in the case of duplicates, the entry with the highest score is kept.

### Relates
- #10227 
- #10228 (fix on 4.x branch)
- #3911

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -291,6 +291,20 @@ const Search = {
       return leftScore > rightScore ? 1 : -1;
     });
 
+    // remove duplicate search results
+    // note the reversing of results, so that in the case of duplicates, the highest-scoring entry is kept
+    let seen = new Set();
+    results = results.reverse().reduce((acc, result) => {
+      let resultStr = result.slice(0, 4).concat([result[5]]).map(v => String(v)).join(',');
+      if (!seen.has(resultStr)) {
+        acc.push(result);
+        seen.add(resultStr);
+      }
+      return acc;
+    }, []);
+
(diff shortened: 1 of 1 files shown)


calebchiam:
Hi @tk0miya, I've applied the fix on the 5.0 branch as requested. 

Is there an estimate for when 5.0 will be released? If it's still some time away, it may be worth releasing the fix on the 4.x branch for the time being, since it's an issue that affects many Sphinx docs.

Some other examples out in the wild:
- [Flask: search 'app context'](https://flask.palletsprojects.com/en/2.0.x/search/?q=app+context)
- [Cleanlab: search 'noisy labels'](https://docs.cleanlab.ai/v1.0.1/search.html?q=noisy+labels&check_keywords=yes&area=default)

Thank you!

tk0miya:
Thanks! I'll merge this after CI passed.

>Is there an estimate for when 5.0 will be released? If it's still some time away, it may be worth releasing the fix on the 4.x branch for the time being, since it's an issue that affects many Sphinx docs.

Now it is planned for the end of April or the beginning of May. I feel it's not far.

calebchiam:
Awesome! Good to hear, and thanks for the hard work.

tk0miya:
Merged. Thank you for your contribution. Please be patient until the 5.0 release (Spring has come soon!).

PR #10239, opened 2022-03-06
Close #10234: autosummary: Add "autosummary" CSS class to summary tables

### LINKED ISSUE #10234: autosummary tables could use an extension-specific CSS class
### Is your feature request related to a problem? Please describe.
We are working on a theme that does not inherit from a builtin theme's CSS. When using autosummary with this theme, the tables inherit a CSS rule not designed for autosummary tables. We'd like a way to identify (via CSS) an autosummary table from other parts of the document(s).

### Describe the solution you'd like
In discussing how to override our theme's rule using a way to only target autosummary tables, we've agreed that autosummary should use an extension-specific CSS class that would distinguish its tables.

### Describe alternatives you've considered
1. Use a special CSS override whose selectors target only autosummary tables. This approach carries some uncertainty because of the problem stated above. In the meantime, we've gone with this approach since the `longtable` CSS class seems to be intended for latex writers.
2. Monkey patch the `autosummary_table_visit_html()` in our theme's source. This seems unnecessarily complex to target only autosummary tables. It also may not be forward compatible.

### Additional context
The theme's css rule that causes problems is 
```css
.md-typeset code {word-break: break-word;}
```
We needed to design a rule that sets `word-break: normal;` as it was presumptuously intended by autosummary.

Our theme's issue was reported in https://github.com/jbms/sphinx-immaterial/issues/35

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- refs: #10234

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@ Deprecated
 Features added
 --------------
 
+* #10234: autosummary: Add "autosummary" CSS class to summary tables
 * #10125: extlinks: Improve suggestion message for a reference having title
 * #9494, #9456: html search: Add a config variable
   :confval:`html_show_search_summary` to enable/disable the search summaries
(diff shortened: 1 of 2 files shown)


2bndy5:
Looks good to me. Thanks for the quick response!

tk0miya:
Thank you for confirming. Merging now. Please wait for the next release (4.5.0).

PR #10243, opened 2022-03-07
Fix ArrowRight shortcut.

marxin:
The regression is introduced with
799385f5558a888d1a143bf703d06b66d6717fe4.


marxin:
A more complex fix is in #10245, closing this one.

PR #10249, opened 2022-03-11
Ensure ObjectDescription signature nodes include line numbers

jbms:
### Feature or Bugfix

- Bugfix

### Purpose
Previously, if any options were specified to the directive, line
numbers were not properly included in the signature nodes.

### DIFF
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -159,12 +159,12 @@ class ObjectDescription(SphinxDirective, Generic[T]):
         node.document = self.state.document
         source, line = self.get_source_info()
         # If any options were specified to the directive,
-        # `self.state.document.current_line` will at this point be set to
-        # `None`.  To ensure nodes created as part of the signature have a line
+        # self.state.document.current_line will at this point be set to
+        # None.  To ensure nodes created as part of the signature have a line
         # number set, set the document's line number correctly.
         #
         # Note that we need to subtract one from the line number since
-        # `note_source` uses 0-based line numbers.
+        # note_source uses 0-based line numbers.
         if line is not None:
             line -= 1
         self.state.document.note_source(source, line)
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -157,6 +157,15 @@ class ObjectDescription(SphinxDirective, Generic[T]):
 
         node = addnodes.desc()
         node.document = self.state.document
+        source, line = self.get_source_info()
+        # If any options were specified to the directive,
+        # `self.state.document.current_line` will at this point be set to
+        # `None`.  To ensure nodes created as part of the signature have a line
+        # number set, set the document's line number correctly.
+        #
+        # Note that we need to subtract one from the line number since
+        # `note_source` uses 0-based line numbers.
+        self.state.document.note_source(source, line - 1 if line is not None else None)
         node['domain'] = self.domain
         # 'desctype' is a backwards compatible attribute
         node['objtype'] = node['desctype'] = self.objtype
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -165,7 +165,9 @@ class ObjectDescription(SphinxDirective, Generic[T]):
         #
         # Note that we need to subtract one from the line number since
         # `note_source` uses 0-based line numbers.
-        self.state.document.note_source(source, line - 1 if line is not None else None)
+        if line is not None:
+            line -= 1
+        self.state.document.note_source(source, line)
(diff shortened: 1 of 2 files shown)


tk0miya:
Could you explain your problem, please? Does it happen only with the directives implemented via ObjectDescription?

jbms:
You can get a sense of the problem from the test case --- basically in:

```rst
.. py:function foo(bar: X)
   :noindexentry:

   Description
```

The signature line won't have correct source line numbers, which means, for example, that if the `X` type can't be resolved the warning message won't show correct line numbers.

Due to the way docutils parses directives, it only happens if there are options to the directive.  At first I thought it may be a bug in docutils, but I think it is actually the responsibility of the directive itself to sort out line numbers anyway, since only the directive knows which source line is actually the source of a given node that it produces.

For the signature, sphinx already attempts to set the line number to match the first line of the directive here:
https://github.com/sphinx-doc/sphinx/blob/b12b39db05d605cfb946f84afd27e0f6cf3c9e13/sphinx/directives/__init__.py#L175

However, in many cases the line numbers are instead obtained from `self.state.document.current_line` as child nodes are added, so setting the source info on the root signature node is not sufficient.

jbms:
@tk0miya Is there anything left to do on this?

jbms:
@tk0miya It would be great to get this fix in.  Do you think you might have chance to review it again soon?

tests/test_domain_py.py
```diff
@@ -1351,3 +1352,16 @@ def test_warn_missing_reference(app, status, warning):
...
             in warning.getvalue())
+
+
+@pytest.mark.sphinx(confoverrides={'nitpicky': True})
+@pytest.mark.parametrize('include_options', (True, False))
+def test_signature_line_number(app, include_options):
+    text = (".. py:function:: foo(bar : string)\n" +
+            ("   :noindexentry:\n" if include_options else ""))
+    doc = restructuredtext.parse(app, text)
+    xrefs = list(doc.traverse(condition=addnodes.pending_xref))
```
AA-Turner: This should be `.findall()`

A

AA-Turner (commented):
I should be able to review over the weekend.

A

tests/test_domain_py.py
jbms: Thanks --- fixed.  I had thought that `traverse` already returned an iterator, but I see that `traverse(...)` is equivalent to `list(findall(...))`

sphinx/directives/__init__.py
```diff
@@ -157,6 +157,15 @@ def run(self) -> List[Node]:
...
         node.document = self.state.document
+        source, line = self.get_source_info()
+        # If any options were specified to the directive,
+        # `self.state.document.current_line` will at this point be set to
+        # `None`.  To ensure nodes created as part of the signature have a line
+        # number set, set the document's line number correctly.
+        #
+        # Note that we need to subtract one from the line number since
+        # `note_source` uses 0-based line numbers.
+        self.state.document.note_source(source, line - 1 if line is not None else None)
```
AA-Turner: The ternary within the function call is hard to parse:
```suggestion
        if line is not None:
            line -= 1
        self.state.document.note_source(source, line)
```

AA-Turner (approved):

sphinx/directives/__init__.py
```diff
@@ -157,6 +157,17 @@ def run(self) -> List[Node]:
 
         node = addnodes.desc()
         node.document = self.state.document
+        source, line = self.get_source_info()
+        # If any options were specified to the directive,
+        # `self.state.document.current_line` will at this point be set to
```
AA-Turner:
```suggestion
        # self.state.document.current_line will at this point be set to
```

sphinx/directives/__init__.py
```diff
@@ -157,6 +157,17 @@ def run(self) -> List[Node]:
...
+        # `None`.  To ensure nodes created as part of the signature have a line
```
AA-Turner:
```suggestion
        # None.  To ensure nodes created as part of the signature have a line
```

sphinx/directives/__init__.py
```diff
@@ -157,6 +157,17 @@ def run(self) -> List[Node]:
...
+        # number set, set the document's line number correctly.
+        #
+        # Note that we need to subtract one from the line number since
+        # `note_source` uses 0-based line numbers.
```
AA-Turner:
```suggestion
        # note_source uses 0-based line numbers.
```

AA-Turner:
The markup within comments was incorrect (single rather than double backtick), but I just removed as it is within comments.

A

PR #10251, opened 2022-03-11
Make code role highlighting consistent with code-block directive

### LINKED ISSUE #5157: Inline code is not syntax highlighted

jbms:
### Feature or Bugfix
- Bugfix

### Purpose
Fixes https://github.com/sphinx-doc/sphinx/issues/5157
    
This is factored out of the sphinx-immaterial theme:
https://github.com/jbms/sphinx-immaterial/blob/1ef121a612d4f5afc2a9ca9c4e3f20fca89065e8/sphinx_immaterial/inlinesyntaxhighlight.py#L1
    
See also: https://github.com/sphinx-doc/sphinx/pull/6916

### DIFF
--- a/sphinx/roles.py
+++ b/sphinx/roles.py
@@ -11,6 +11,9 @@
 import re
 from typing import TYPE_CHECKING, Any, Dict, List, Tuple, Type
 
+import docutils.parsers.rst.directives
+import docutils.parsers.rst.roles
+import docutils.parsers.rst.states
 from docutils import nodes, utils
 from docutils.nodes import Element, Node, TextElement, system_message
 
@@ -341,6 +344,57 @@ class Abbreviation(SphinxRole):
         return [nodes.abbreviation(self.rawtext, text, **options)], []
 
 
+# Sphinx provides the `code-block` directive for highlighting code blocks.
+# Docutils provides the `code` role which in theory can be used similarly by
+# defining a custom role for a given programming language:
+#
+#     .. .. role:: python(code)
+#          :language: python
+#          :class: highlight
+#
+# In practice this does not produce correct highlighting because it uses a
+# separate highlighting mechanism that results in the "long" pygments class
+# names rather than "short" pygments class names produced by the Sphinx
+# `code-block` directive and for which this extension contains CSS rules.
+#
+# In addition, even if that issue is fixed, because the highlighting
+# implementation in docutils, despite being based on pygments, differs from that
+# used by Sphinx, the output does not exactly match that produced by the Sphinx
+# `code-block` directive.
+#
+# This issue is noted here: //github.com/sphinx-doc/sphinx/issues/5157
+#
+# This overrides the docutils `code` role to perform highlighting in the same
+# way as the Sphinx `code-block` directive.
+#
+# TODO: Change to use `SphinxRole` once SphinxRole is fixed to support options.
+def code_role(name: str, rawtext: str, text: str, lineno: int,
+              inliner: docutils.parsers.rst.states.Inliner,
+              options: Dict = {}, content: List[str] = []
+              ) -> Tuple[List[Node], List[system_message]]:
+    options = options.copy()
+    docutils.parsers.rst.roles.set_classes(options)
+    language = options.get('language', '')
+    classes = ['code']
+    if language:
+        classes.append('highlight')
+    if 'classes' in options:
+        classes.extend(options['classes'])
+
+    if language and language not in classes:
+        classes.append(language)
+
+    node = nodes.literal(rawtext, text, classes=classes, language=language)
+
+    return [node], []
+
+
+code_role.options = {  # type: ignore
+    'class': docutils.parsers.rst.directives.class_option,
+    'language': docutils.parsers.rst.directives.unchanged,
+}
+
+
 specific_docroles: Dict[str, RoleFunction] = {
     # links to download references
     'download': XRefRole(nodeclass=addnodes.download_reference),
@@ -367,6 +421,10 @@ def setup(app: "Sphinx") -> Dict[str, Any]:
 
     for rolename, func in specific_docroles.items():
         roles.register_local_role(rolename, func)
+
+    # Since docutils registers it as a canonical role, override it as a
+    # canonical role as well.
+    roles.register_canonical_role('code', code_role)
 
     return {
         'version': 'builtin',
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -511,10 +511,25 @@ class HTMLTranslator(SphinxTranslator, BaseTranslator):
         if 'kbd' in node['classes']:
             self.body.append(self.starttag(node, 'kbd', '',
                                            CLASS='docutils literal notranslate'))
-        else:
+            return
+        lang = node.get("language", None)
+        if 'code' not in node['classes'] or not lang:
             self.body.append(self.starttag(node, 'code', '',
                                            CLASS='docutils literal notranslate'))
             self.protect_literal_text += 1
+            return
+
+        opts = self.config.highlight_options.get(lang, {})
+        highlighted = self.highlighter.highlight_block(
+            node.astext(), lang, opts=opts, location=node, nowrap=True)
+        starttag = self.starttag(
+            node,
+            "code",
+            suffix="",
+            CLASS="docutils literal highlight highlight-%s" % lang,
+        )
+        self.body.append(starttag + highlighted.strip() + "</code>")
+        raise nodes.SkipNode
 
     def depart_literal(self, node: Element) -> None:
         if 'kbd' in node['classes']:
--- a/sphinx/writers/latex.py
+++ b/sphinx/writers/latex.py
@@ -1747,10 +1747,27 @@ class LaTeXTranslator(SphinxTranslator):
     def visit_literal(self, node: Element) -> None:
         if self.in_title:
             self.body.append(r'\sphinxstyleliteralintitle{\sphinxupquote{')
+            return
         elif 'kbd' in node['classes']:
             self.body.append(r'\sphinxkeyboard{\sphinxupquote{')
-        else:
+            return
+        lang = node.get("language", None)
+        if 'code' not in node['classes'] or not lang:
             self.body.append(r'\sphinxcode{\sphinxupquote{')
+            return
+
+        opts = self.config.highlight_options.get(lang, {})
+        hlcode = self.highlighter.highlight_block(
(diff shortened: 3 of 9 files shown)


sphinx/writers/html.py
```diff
@@ -75,9 +75,15 @@ def translate(self) -> None:
...
+if TYPE_CHECKING:
+    _MIXIN_BASE = BaseTranslator
+else:
+    _MIXIN_BASE = object
+
+
+class HTMLTranslatorCommonMixin(SphinxTranslator, _MIXIN_BASE):
     """
-    Our custom HTML translator.
+    Overrides that are common to both HTMLTranslator and HTML5Translator.
```
tk0miya: It seems this is a different topic from this PR. It should be discussed in another one. I feel this makes HTML writers strongly coupled. Then it makes removing the HTML4 writer from core difficult.

tk0miya (commented):
-1: It's better to propose adding a new role to the docutils project first. No reason to add a new Sphinx-originated custom role.

sphinx/writers/html.py
tk0miya: Ah, sorry. I found this as https://github.com/sphinx-doc/sphinx/pull/10250.

jbms:
Docutils already contains a `:code:` role --- it just doesn't highlight correctly in sphinx due to using the wrong CSS class names, as noted in this long-standing bug: https://github.com/sphinx-doc/sphinx/issues/5157  Additionally though, even if you use one of the workarounds to fix that bug, the highlighting is not consistent with the sphinx `code-block` directive, which is confusing.

This PR makes `:code:` just work in the way I think users expect --- the same as the sphinx `code-block` directive.

Per your comment here:
https://github.com/sphinx-doc/sphinx/pull/6916#issuecomment-568158487

You seemed to be supportive to fixing this problem in Sphinx itself.  You had been waiting on your pygments PR to add a nowrap option to the latex highlighter.  However, that PR seems to have stalled, and I was able to implement a fix for latex without needing that option.

AA-Turner:
Long term, given that a `code` role now exists in Docutils, wouldn't it be better to try and remove the custom code from Sphinx? This would mean all reStructuredText users get the same behaviour.

A

jbms:
@AA-Turner I wasn't actually aware that docutils had a `.. code::` directive in addition to a `:code:` role.

This is my understanding of the current situation in sphinx:

- `:code:` role and `.. code::` directive: from docutils, under default configuration produces highlighting with long class names not included in sphinx themes and therefore appears unstyled in sphinx-generated html.  In practice probably rarely used with Sphinx since it does not work by default.
- `.. code-block::` directive: from sphinx, produces styled output with short class names with sphinx-modified styling and is correctly styled by sphinx themes.  Widely used with Sphinx.
- Literal blocks: From docutils, but modified by sphinx to take into account the `.. highlight` directive language setting and optionally perform syntax highlighting.  Produces short class names with sphinx-modified styling and is correctly styled by sphinx themes.  Widely used with sphinx.

It sounds like you are proposing that we eliminate `code-block`, literal blocks, all of the sphinx code that handles highlighting, and fix docutils to output class names that work with sphinx themes?  Or do you have a different proposal for how to fix this?

tk0miya:
Wow, I did not notice reST has already provided the `code` role since v0.9. Okay, let's implement it ASAP.
https://docutils.sphinx-users.jp/docutils/docs/ref/rst/roles.html#code

>:code: role and .. code:: directive: from docutils, under default configuration produces highlighting with long class names not included in sphinx themes and therefore appears unstyled in sphinx-generated html. In practice probably rarely used with Sphinx since it does not work by default.

Could you let me know an example of `code` directive, please? I consider it's supported.

sphinx/roles.py
```diff
@@ -341,6 +344,53 @@ def run(self) -> Tuple[List[Node], List[system_message]]:
...
+# In addition, even if that issue is fixed, because the highlighting
+# implementation in docutils, despite being based on pygments, differs from that
+# used by Sphinx, the output does not exactly match that produced by the Sphinx
+# `code-block` directive.
+#
+# This issue is noted here: //github.com/sphinx-doc/sphinx/issues/5157
+#
+# This overrides the docutils `code` role to perform highlighting in the same
+# way as the Sphinx `code-block` directive.
+def code_role(name: str, rawtext: str, text: str, lineno: int,
```
tk0miya, line 371: Could you use `SphinxRole` class to implement a role, please?

sphinx/writers/html.py
```diff
@@ -511,10 +497,31 @@ def visit_literal(self, node: Element) -> None:
...
+        if 'code' not in node['classes'] or not lang:
             self.body.append(self.starttag(node, 'code', '',
                                            CLASS='docutils literal notranslate'))
             self.protect_literal_text += 1
+            return
+
+        def warner(msg):
+            logger.warning(msg, location=node)
+
+        highlight_args = dict(node.get("highlight_args", {}), nowrap=True)
```
tk0miya: Where this option comes from?

sphinx/writers/html.py
```diff
@@ -511,10 +497,31 @@ def visit_literal(self, node: Element) -> None:
...
-        else:
+            return
+        lang = node.get("language", None)
+        if 'code' not in node['classes'] or not lang:
             self.body.append(self.starttag(node, 'code', '',
                                            CLASS='docutils literal notranslate'))
             self.protect_literal_text += 1
+            return
+
+        def warner(msg):
```
tk0miya: Where is this helper function used? AFAIK, our highlighting module no longer support this argument.

sphinx/writers/html.py
```diff
@@ -511,10 +497,31 @@ def visit_literal(self, node: Element) -> None:
         if 'kbd' in node['classes']:
             self.body.append(self.starttag(node, 'kbd', '',
                                            CLASS='docutils literal notranslate'))
-        else:
+            return
+        lang = node.get("language", None)
+        if 'code' not in node['classes'] or not lang:
```
tk0miya, line 516: Should the node only having `language` attribute be represented as an inline code? I think it would be better to judge it only by `code` attribute.

tk0miya:
I perfectly understand now. The `code` role is mainly used to define a custom role via `role` directive. So far I know the technique, but I did not recognize it's based on the role.

sphinx/writers/latex.py
```diff
@@ -1747,10 +1747,31 @@ def depart_citation_reference(self, node: Element) -> None:
...
+        def warner(msg):
+            logger.warning(msg, location=node)
+
+        highlight_args = dict(node.get("highlight_args", {}), nowrap=True)
+        opts = self.config.highlight_options.get(lang, {})
+        hlcode = self.highlighter.highlight_block(
+            node.astext(), lang, opts=opts,
+            location=node, **highlight_args
+        )
+        hlcode = hlcode.replace(r'\begin{Verbatim}[commandchars=\\\{\}]',
```
tk0miya, line 1764: Note: This macro can be changed via `highlight_args`. But as I commented above, the `code` role does not support it. So we don't need to care the output.

Additionally, it would be better to use `nowrap` option for the LaTeX formatter (refs: https://github.com/pygments/pygments/pull/1343) instead of str.replace(). But it's still not released yet. It would be better to mention the option as a comment here.

tk0miya (changes requested):
LGTM with nits. But merging HTML4 and HTML5 writers is another topic. So it should be separated from this PR. Could you add the change to each writers, please?

sphinx/writers/html.py
jbms: Removed.

sphinx/writers/latex.py
jbms: Done.

sphinx/writers/html.py
jbms: Removed.

sphinx/roles.py
jbms: Done.

sphinx/writers/html.py
jbms: The current logic is to highlight only if `code` is in the classes AND the `language` attribute has been set.  If `language` isn't set, we can't highlight since we don't know what language to use.  If `language` is set but `code` is not in classes (not sure how that would happen), then it also won't highlight.

There is the option to use the default language set by the `highlight` directive if the user just uses the plain `code` directive.  That would make the `code` role consistent with the `code-block` directive.  However, that would affect existing uses of the `code` directive, so I think it is probably better not to do that.

jbms:
Thanks for your feedback -- I have addressed your comments and also updated the documentation.

jbms:
I changed this to use `SphinxRole`, but then changed it back because I realized `SphinxRole` does not work correctly with options.

The problem is that you need to define the option specs via the `options` attribute, but the `__call__` method then sets `self.options` to the actual option values, not the option specs.  In general the `SphinxRole` design seems kind of odd, in that you have a single object representing the role that gets modified each time it is used.  Normally it nonetheless works but for options we run into a problem.  It would be better to have a design where a fresh object is created for each use of the role.  However, we could alternatively fix it by using a different attribute for the option values, e.g. `option_values`.

jbms:
@tk0miya Is there anything left to do on this?

AA-Turner:
I'll review this properly over the next week or so.

A

sphinx/writers/html.py
tk0miya: I undertstand.

tk0miya (approved):
Sorry for the late response. LGTM.

I'll take a look at the design problem of SphinxRole in another time. Thank  you for the info.

tk0miya:
Just merged. Thank you for your great work!

jbms:
Thanks.  Looks like this still needs to be merged into 5.x since my pull request targeted 4.x.

tk0miya:
Thanks! I merged this into 5.x branch manually now.

PR #10258, opened 2022-03-11
Recognize a documented attribute of a module as non-imported.

lokik:
Subject: To recognize better, whether a module attribute is imported or not.

Feature or Bugfix - It changes the behavior of the autosummary, so it is a breaking change,
but this change can or can not (depending on the point of view) be considered as a bugfix

### Purpose
If an object of an imported class is instantiated in a module and it is assigned to a module attribute, 
it has been so far always considered as imported. This results in not documenting some module members.

The behavior comes from the fact, that object.__module__ attribute is inherited from its class
and the comparison of __module__ attribute have been the way how to recognize imported 
stuff.

Since there is probably no simple way how to find out whether an object
was imported or not, this pullrequest solve the issue partially.
If a module attribute is explicitly documented in the source file, 
it is surely non-imported - thus, it is marked so.

### DIFF
--- a/sphinx/ext/autosummary/generate.py
+++ b/sphinx/ext/autosummary/generate.py
@@ -193,8 +193,10 @@ class ModuleScanner:
                            name, exc, type='autosummary')
             return False
 
-    def scan(self, imported_members: bool, attr_docs: Mapping[str,str] = {}) -> List[str]:
+    def scan(self, imported_members: bool) -> List[str]:
         members = []
+        analyzer = ModuleAnalyzer.for_module(self.object.__name__)
+        attr_docs = analyzer.find_attr_docs()
         for name in members_of(self.object, self.app.config):
             try:
                 value = safe_getattr(self.object, name)
@@ -306,6 +308,8 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
         """Find module attributes with docstrings."""
         attrs, public = [], []
         try:
+            analyzer = ModuleAnalyzer.for_module(name)
+            attr_docs = analyzer.find_attr_docs()
             for namespace, attr_name in attr_docs:
                 if namespace == '' and attr_name in members:
                     attrs.append(attr_name)
@@ -334,10 +338,8 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
     ns.update(context)
 
     if doc.objtype == 'module':
-        analyzer = ModuleAnalyzer.for_module(name)
-        attr_docs = analyzer.find_attr_docs()
         scanner = ModuleScanner(app, obj)
-        ns['members'] = scanner.scan(imported_members, attr_docs)
+        ns['members'] = scanner.scan(imported_members)
         ns['functions'], ns['all_functions'] = \
             get_members(obj, {'function'}, imported=imported_members)
         ns['classes'], ns['all_classes'] = \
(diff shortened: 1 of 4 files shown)


sphinx/ext/autosummary/generate.py
```diff
@@ -335,8 +334,10 @@ def get_modules(obj: Any) -> Tuple[List[str], List[str]]:
     ns.update(context)
 
     if doc.objtype == 'module':
+        analyzer = ModuleAnalyzer.for_module(name)
+        attr_docs = analyzer.find_attr_docs()
```
tk0miya: Could you generate `analyzer` inside `scanner.scan()` too (without removing it from `get_module_attrs`)? It's no cost to call it because cached.

I'd like to merge `ModuleScanner` class and `get_module_attrs()` in nearly future. So it's better not to the interface of `scan()` method.

sphinx/ext/autosummary/generate.py
```diff
@@ -208,7 +209,7 @@ def scan(self, imported_members: bool) -> List[str]:
                 if inspect.ismodule(value):
                     imported = True
                 elif safe_getattr(value, '__module__') != self.object.__name__:
-                    imported = True
+                    imported = objtype != 'data' or ('', name) not in attr_docs
```
tk0miya, line 214: Is this condition related to the `__module__`? I think this check should be separated to another condition.

tk0miya (changes requested):
LGTM with nits. Could you add a testcase for this, please?

sphinx/ext/autosummary/generate.py
lokik: I viewed it as "fix" for the the fact, that instances of user defined classes has `__module__` property set to the module, where the class is defined, and not to the one where the object is instantiated. So from the original intent of the commit, yes, it is related, as it is an attempt to fix the false positive of the 'is-the-stuff-imported?' test. 

But you are right, that you can consider it and define, that non-imported module attributes are just these, that have some docstring and thus made the condition separate (before the currently modified one). However, this IMHO would lead to further change in the autosummary behavior: so far the instantiated object without docstrings that has `__module__` set to the module, where they are instantiated (e.g. if their class is defined in the same module) are considered as non-imported - and that would change.

So, this is maybe really a more systematic treating of the problem and thus maybe a better option? However, the resulting criterion would have no connection to "being imported" at all - in this case I would consider an introduction of a new option, something like "autosummary_only_documented_modules_attributes", and to treat the module attributes according to this new option, independently to the autosummary_imported_members option.

sphinx/ext/autosummary/generate.py
lokik: The both have been incorporated.

sphinx/ext/autosummary/generate.py
tk0miya: In autodoc, global variables having doc-comment are documented without `:imported-members:` option even if their `__module__` value are different because they're considered as "locally defined variables".

```
class Foo:
    pass


#: a global variable having different __module__
var1 = int

#: a global variable having same __module__
var2 = Foo
```

So I think it's better to check the member has doc-comment or not first to determine `imported` flag.

```
if ('', name) in attr_docs:
  imported = False
elif inspect.ismodule(value):
  ...
```

tk0miya:
Merged this into the 5.x branch (see #10428).

Thank you for your work!

PR #10260, opened 2022-03-12
Enable FORCE_COLOR and NO_COLOR for terminal colouring

hugovk:
Subject: Allow these widely-used environment variables to improve readability of output

### Feature or Bugfix

- Feature

### Purpose

Colour can help diagnose problems, and a common place for Sphinx and the need to diagnose problems is on a CI.

However, because [GitHub Actions doesn't use a tty](https://github.com/actions/runner/issues/241), output of tools which autodetect the terminal output is usually monochrome.

Normally you only want to enable colour on a tty, because if you're piping data between commands (not a tty) the colour codes can mess up parsing of piped output. But that's not the case here.

That's why many tools and CI setups use the `FORCE_COLOR` environment variable, and/or command-line options to turn it back on.

Other tools for reference:

* pytest [issue](https://github.com/pytest-dev/pytest/issues/7464), [PR](https://github.com/pytest-dev/pytest/pull/7466) and [docs](https://pytest.org/en/7.0.x/reference/reference.html#environment-variables)
* nox [issue](https://github.com/theacodes/nox/issues/502) and [PR](https://github.com/theacodes/nox/pull/524)
* tox [PR](https://github.com/tox-dev/tox/pull/1630)
* pypa/build [PR](https://github.com/pypa/build/pull/335)
* Colorama [WIP PR](https://github.com/tartley/colorama/pull/230)
* Node.js [docs](https://nodejs.org/api/tty.html#writestreamgetcolordepthenv)

### Detail
- Like other tools, `NO_COLOR` takes precedence
- Keep the Windows-specific check before because it initialises `colorama`
- `FORCE_COLOR` next

### Relates
- https://no-color.org/

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -73,6 +73,7 @@ Other contributors, listed alphabetically, are:
 * Pauli Virtanen -- autodoc improvements, autosummary extension
 * Eric N. Vander Weele -- autodoc improvements
 * Stefan van der Walt -- autosummary extension
+* Hugo van Kemenade -- support FORCE_COLOR and NO_COLOR
 * Thomas Waldmann -- apidoc module fixes
 * John Waltman -- Texinfo builder
 * Barry Warsaw -- setup command improvements
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,7 @@ Deprecated
 Features added
 --------------
 
+* #10260: Enable ``FORCE_COLOR`` and ``NO_COLOR`` for terminal colouring
 * #10234: autosummary: Add "autosummary" CSS class to summary tables
 * #10125: extlinks: Improve suggestion message for a reference having title
 * #9494, #9456: html search: Add a config variable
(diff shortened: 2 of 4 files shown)


tk0miya (changes requested):
LGTM. Could you add mentions for these envvars to `doc/man/sphinx-build.rst` too?

hugovk:
Added: https://sphinx--10260.org.readthedocs.build/en/10260/man/sphinx-build.html#environment-variables

ssbarnea (approved):
After digging for few good number of minutes about why sphinx does not use colors i reached this one.

Yes, current cli arguments are not automation/scripting friendly and we are much better reusing existing standard variables for controlling the colors.

ssbarnea:
@tk0miya Can you please have another look, it seems fixed. I am really eager to get this fix in especially as the older tox trick with `{tty:--color}` no longer works, for the same reasons. TTY is not the same as supporting color or not.

tk0miya (approved):

tk0miya:
Thank you for your update. Merging now!

Issue #10253, opened 2022-03-11
`:pep:` directive generates redirected links


arekfu:
### Describe the bug

The `:pep:` directive generates links to `https://www.python.org/dev/peps/pep-$PEP`, but the linkcheck builder warns that these URLs are now redirected to `https://peps.python.org/pep-$PEP/`.

### How to Reproduce

```
$ python3 -m venv ./venv
$ source ./venv/bin/activate
$ pip install sphinx
$ git clone git@github.com:arekfu/sphinx-issue-pep.git && cd sphinx-issue-pep
$ sphinx-build -b linkcheck . _build
Running Sphinx v4.4.0
building [mo]: targets for 0 po files that are out of date
building [linkcheck]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
reading sources... [100%] index
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index

(           index: line    4) redirect  https://www.python.org/dev/peps/pep-0008 - permanently to https://peps.python.org/pep-0008/
build succeeded.

Look for any errors in the above output or in _build/output.txt
```

### Expected behavior

No linkcheck warning should be produced.

### Your project

https://github.com/arekfu/sphinx-issue-pep

### OS

Linux Ubuntu 20.04

### Python version

3.8.10

### Sphinx version

4.4.0

arekfu:
I just realised this is probably a [docutils](https://github.com/docutils-mirror/docutils) bug, isn't it?

arekfu:
Link to the docutils bug report: [here](https://sourceforge.net/p/docutils/bugs/445/)

nijel:
The default value for this is hardcoded in Sphinx as well, https://github.com/sphinx-doc/sphinx/pull/10267 should address this.

PR #10267, opened 2022-03-16
environment: Update default value for pep_base_url

nijel:
Subject: Update default value for pep_base_url

### Feature or Bugfix

- Bugfix

### Purpose
- Avoids Sphinx generating URLs for PEPs which redirects

### Detail
- #10253

### Relates
- https://sourceforge.net/p/docutils/code/9045/
- https://sourceforge.net/p/docutils/bugs/445/


tk0miya:
Thanks!

PR #10268, opened 2022-03-16
URI-escape image filenames

eric-wieser:
Without this change, local images with `#` in their name result in incorrect URLs in HTML builds.

There is already a similar call to `urllib.parse.quote` for file downloads, added in #9670, suggesting this is a sensible approach.

This also adds a hack to make things work for `#` specifically in LaTeX, but more work is likely needed for other weird character there.

### Feature or Bugfix

- Bugfix


tk0miya (changes requested):
Looks good. Could you add a testcase for this, please? Then I'll merge this soon.

eric-wieser:
#9670 didn't have a test-case, so I was hoping I could get away without one here too...

I might have time for a test-case in a week or so, we'll see.

tk0miya:
It seems #9670 contains a test-case :-)

tk0miya:
Hmm... epubcheck and LaTeX build are failed when we use `#` for the filename. So escaping is not a good way to handle such files. We need to rename them on copying to the generated document.

eric-wieser:
It sounds like the latex code needs escaping too

eric-wieser:
Thanks for fixing the broken tests!

eric-wieser:
Is there any way to see the latex error message from CI?

eric-wieser:
I think the CI failure is actually a bug in `epubcheck`

sphinx/builders/_epub_base.py
```diff
@@ -6,6 +6,7 @@
 from os import path
 from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple
 from zipfile import ZIP_DEFLATED, ZIP_STORED, ZipFile
+from urllib.parse import quote
```
AA-Turner:
```suggestion
from urllib.parse import quote
from zipfile import ZIP_DEFLATED, ZIP_STORED, ZipFile
```

AA-Turner:
@eric-wieser it looks like the test failures are pretty simple to solve--are you up for having a look, it would be nice to get this in to 5.3.

A

CHANGES
```diff
@@ -20,6 +20,7 @@ Features added
   ``:option:`--module[=foobar]``` or ``:option:`--module foobar```.
   Patch by Martin Liska.
 * #10881: autosectionlabel: Record the generated section label to the debug log.
+# 10268: Correctly URI-escape image filenames.
```
```suggestion
* #10268: Correctly URI-escape image filenames.
```

AA-Turner:
Thanks @eric-wieser!

A

PR #10285, opened 2022-03-21
Fixed singledispatch documentation

jmp1985:
Subject: Single dispatch functions were not documented due to minor bug which this pull request fixes

### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
- Previously, singledispatch functions never had their variants documented. This bugfix results in the variants being documented. 

### Detail

Here is a minimal example package which shows the error: 
[bug_example.zip](https://github.com/sphinx-doc/sphinx/files/8317999/bug_example.zip)

Here is an old picture of the output:
sphinx_old

Here is new picture of the output which is what I think is intended:
sphinx-new


tk0miya:
Could you add a testcase, please?

jmp1985:
Hi @tk0miya. I would be happy to add a test case but would you be able to point me to an example test or template I could follow?

tk0miya:
Please check these files:

* https://github.com/sphinx-doc/sphinx/blob/5.x/tests/test_ext_autodoc.py#L2063-L2127
* https://github.com/sphinx-doc/sphinx/blob/5.x/tests/roots/test-ext-autodoc/target/singledispatch.py
* https://github.com/sphinx-doc/sphinx/blob/5.x/tests/roots/test-ext-autodoc/target/singledispatchmethod.py

They confirms the result of autodoc for singledispatch functions and methods. I think this change should change (add?) something new to these testcases.

jmp1985:
Thanks @tk0miya 

I have added the test cases. It turned out that with single dispatch if you register the variants as follows everything works fine:

```
@func.register(int)
def func_int(arg):
    pass
```

However, if you use type annotations as follows then sphinx would not work:

```
@func.register
def func_int(arg: int):
    pass
```

This pull request now fixes this behaviour and adds a test case for functions registered using type annotations.

CheddarBobbb2 (approved):

tk0miya (approved):
Thank you for your update. I don't know singledispatch now supports type annotated functions and methods.
https://docs.python.org/3/library/functools.html#functools.singledispatch

LGTM!

tk0miya:

PR #10286, opened 2022-03-22
[C++] Support requires-clause in more places

jbms:
### Feature or Bugfix
- Feature

### Purpose
Previously a C++20 requires-clause was only supported on `function`
declarations.  However, the C++ standard allows a require-clause on
class/union templates, alias templates, and variable templates, and
also allows a requires clause after each template parameter list, not
just the final one.

This moves the requiresClause to be a property of `ASTTemplateParams`
rather than `ASTDeclaration` to better match the C++ grammar and
allows requires clauses in many places that are supported by C++20 but
were not previously allowed by Sphinx, namely:

- On class/union templates, alias templates, and variable templates

- After each template parameter list, not just the last one.

- After the template parameter list in template template parameters.

Additionally:

- This adds support for template parameters on unions.

- This adds support for trailing requires clauses on functions without
  a template prefix.  This is allowed by C++20 for non-template members
  of class templates.

When encoding the id, the requires clause of the last template
parameter list is treated specially in order to preserve compatibility
with existing v4 ids.

### DIFF
--- a/sphinx/domains/cpp.py
+++ b/sphinx/domains/cpp.py
@@ -3659,17 +3659,29 @@ class ASTTemplateParamNonType(ASTTemplateParam):
 
 
 class ASTTemplateParams(ASTBase):
-    def __init__(self, params: List[ASTTemplateParam]) -> None:
+    def __init__(self, params: List[ASTTemplateParam],
+                 requiresClause: Optional["ASTRequiresClause"]) -> None:
         assert params is not None
         self.params = params
+        self.requiresClause = requiresClause
 
-    def get_id(self, version: int) -> str:
+    def get_id(self, version: int, exclude_requires: bool = False) -> str:
+        # Note: For `version==4`, `exclude_requires` is set to `True` when
+        # encoding the id of the last template parameter list of a declaration,
+        # as that requires-clause, if any, is instead encoded by
+        # `ASTDeclaration.get_id` after encoding the template prefix, for
+        # consistency with the existing v4 format used when only a single
+        # requires-clause was supported.
         assert version >= 2
         res = []
         res.append("I")
         for param in self.params:
             res.append(param.get_id(version))
         res.append("E")
+        if not exclude_requires and self.requiresClause:
+            res.append('IQ')
+            res.append(self.requiresClause.expr.get_id(version))
+            res.append('E')
         return ''.join(res)
 
     def _stringify(self, transform: StringifyTransform) -> str:
@@ -3677,6 +3689,9 @@ class ASTTemplateParams(ASTBase):
         res.append("template<")
         res.append(", ".join(transform(a) for a in self.params))
         res.append("> ")
+        if self.requiresClause is not None:
+            res.append(transform(self.requiresClause))
+            res.append(" ")
         return ''.join(res)
 
     def describe_signature(self, signode: TextElement, mode: str,
@@ -3691,6 +3706,9 @@ class ASTTemplateParams(ASTBase):
             first = False
             param.describe_signature(signode, mode, env, symbol)
         signode += addnodes.desc_sig_punctuation('>', '>')
+        if self.requiresClause:
+            signode += addnodes.desc_sig_space()
+            self.requiresClause.describe_signature(signode, mode, env, symbol)
 
     def describe_signature_as_introducer(
             self, parentNode: desc_signature, mode: str, env: "BuildEnvironment",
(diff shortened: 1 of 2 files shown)


jbms:
Note: I made some major changes to this PR to support a lot of additional things.

As it is currently implemented, it does not prohibit requires clauses on concepts even though those are disallowed by C++.  But I don't think that is a problem.

Note: I noticed that there is support for non-standard concept introduction syntax, which I guess was part of an earlier concepts proposal.  Should that be removed since it is not supported by C++?

sphinx/domains/cpp.py
```diff
@@ -3659,24 +3659,33 @@ def describe_signature(self, signode: TextElement, mode: str,
 
 
 class ASTTemplateParams(ASTBase):
-    def __init__(self, params: List[ASTTemplateParam]) -> None:
+    def __init__(self, params: List[ASTTemplateParam],
+                 requiresClause: Optional["ASTRequiresClause"] = None) -> None:
```
jakobandersen:
```suggestion
                 requiresClause: Optional["ASTRequiresClause"]) -> None:
```

sphinx/domains/cpp.py
```diff
@@ -3659,24 +3659,33 @@ def describe_signature(self, signode: TextElement, mode: str,
...
         assert params is not None
         self.params = params
+        self.requiresClause = requiresClause
 
-    def get_id(self, version: int) -> str:
+    def get_id(self, version: int, exclude_requires: bool = False) -> str:
```
jakobandersen, line 3668: What is the reason for not including it in the ID?

jakobandersen (changes requested):
Thanks for the PR. There are a number of structural changes to the code, and you list multiple changes. Please split it up into multiple commits to make it easier to review.

jakobandersen:
> As it is currently implemented, it does not prohibit requires clauses on concepts even though those are disallowed by C++. But I don't think that is a problem.

It would be nice to not allow it, unless it is really cumbersome to disallow.

> Note: I noticed that there is support for non-standard concept introduction syntax, which I guess was part of an earlier concepts proposal. Should that be removed since it is not supported by C++?

They were indeed implemented during the Concepts TS time-frame. If they are in the way of implementing new things we should deprecate them, otherwise I see no harm in letting them be.

sphinx/domains/cpp.py
jbms: Done.

sphinx/domains/cpp.py
jbms: I added a comment to explain.

jbms:
Thanks for reviewing this change.  I have split it up into multiple commits, and added a check to prevent a requires-clause on concepts.

jbms:
@jakobandersen Please let me know if there is anything I can do to help make progress with this and my other C++ PRs.  Thanks!

jakobandersen:
> When encoding the id, the requires clause of the last template
> parameter list is treated specially in order to preserve compatibility
> with existing v4 ids.

Is this really needed? As currently only the clause being ID-encoded is allowed, then all currently allowed declarations will have the same ID afterwards anyway. Or am I missing something?

jbms:
> > When encoding the id, the requires clause of the last template
> > parameter list is treated specially in order to preserve compatibility
> > with existing v4 ids.
> 
> Is this really needed? As currently only the clause being ID-encoded is allowed, then all currently allowed declarations will have the same ID afterwards anyway. Or am I missing something?

Previously, the requires clause was encoded after the template parameter list, so with a requires clause after the last template parameter list we would have:

```
"I" + template_parameter_list_id + "Q" + "IQ" + requires_clause_id + "E"
```

if we didn't special case requires clauses on the last template parameter list, it would instead be:

```
"I" + template_parameter_list_id + "IQ" + requires_clause_id + "E" + "E"
```

jbms:
I didn't see anything in the Itanium ABI regarding requires clauses --- not sure if you just invented a convention for them or if you used a convention defined elsewhere.  With this change, I just used a convention that seemed to make sense.

AA-Turner:
@jakobandersen Should we move this to 5.x, or is it nearly ready to be merged for 5.1?

A

jakobandersen:
> @jakobandersen Should we move this to 5.x, or is it nearly ready to be merged for 5.1?

I won't have time to properly dig back into it before the weekend, so please go ahead with the release and move this to 5.2.

jakobandersen:
> if we didn't special case requires clauses on the last template parameter list ...

Indeed, good catch.

jakobandersen:
> I didn't see anything in the Itanium ABI regarding requires clauses --- not sure if you just invented a convention for them or if you used a convention defined elsewhere. With this change, I just used a convention that seemed to make sense.

Right, to my understanding it hasn't been updated officially yet, so this is just something I cooked up, and your changes are fine.

jakobandersen:
I have rebased and updated this PR and merged it as #10703. Thanks again for your nice work.

PR #10288, opened 2022-03-24
PEP role: Include trailing slash in  URL to avoid 301 redirect

hugovk:
Subject: PEP role: Include trailing / in  URL to avoid 301 redirect

### Feature or Bugfix

- Feature

### Purpose
- Link `:pep:8` to `https://peps.python.org/pep-0008/` instead of `https://peps.python.org/pep-0008`

### Detail
- Without the trailing `/`, there's an HTTP 301 redirect to add the slash, which adds a small but unnecessary delay

#### Without trailing `/`, the 301 may take between 20ms and 335ms (caching off)

image

image

image

#### With the trailing `/`, that request is skipped

image

image

image

- Also update docs and docstrings to use the new form.

### Relates
- https://github.com/sphinx-doc/sphinx/pull/10267


tk0miya:
Thanks!

PR #10296, opened 2022-03-25
[C++] Resolve nested symbols through type aliases and base classes

jbms:
### Feature or Bugfix
- Feature

### Purpose
Previously, symbol resolution always stopped at type aliases and also did not consider base classes.
For example, with the following document:

```rst
.. cpp:class:: Class

   .. cpp::type:: NestedType = int

.. cpp:type:: Alias = Class

:cpp:expr:`Alias::NestedType`
```

`Alias::NestedType` was not able to be resolved.  With this change, it
is correctly resolved to `Class::NestedType`.

Additionally, symbols are also resolved through base classes in a
similar way:

```rst
.. cpp:class:: Base

   .. cpp::type:: NestedType = int

.. cpp:class:: Class : Base

:cpp:expr:`Class::NestedType`
```

With this change, `Class::NestedType` will be correctly resolved to
`Base::NestedType`.


jbms:
@jakobandersen If you have a chance to look at this as well, that would be awesome!

jbms:
@AA-Turner @jakobandersen Since there seems to be renewed activity on the C++ domain, if I rebase this to HEAD would you have time to look at it soon?

AA-Turner:
I've hesitated to review this previously as I have limited C++ experience, and I believe Jakob has limited time to review nowadays.

That being said, I'm happy to review an updated PR, and will trust it on the basis of the tests passing.

A

jakobandersen:
The idea is good, I'll try to review it as soon as possible.

jbms:
@jakobandersen 

> The idea is good, I'll try to review it as soon as possible.

Okay, I rebased it.

AA-Turner:
Ping @jakobandersen if you have time to have a look at this.

A

jakobandersen:
I'll try to get to it within the next weekend or two.

jbms:
@jakobandersen I just rebased this to HEAD --- this has been sitting as a monkey patch in the sphinx-immaterial theme for several years now --- it would be great if we can merge it in.

AA-Turner:
@jakobandersen I would like to get this PR in at some point but I'd prefer to have another review of the C++ resolution changes. If you realistically won't have time I can merge this on trust that @jbms knows what he's doing, but we are pretty lacking in C++ expertise on the team I fear.

A

sphinx/domains/cpp/_symbol.py
```diff
@@ -935,6 +998,54 @@ def direct_lookup(self, key: LookupKey) -> Symbol:
...
+            nested_name = trailing_type_spec.name
+        elif isinstance(declaration.declaration, ASTType):
+            trailing_type_spec = declaration.declaration.declSpecs.trailingTypeSpec
+            if not isinstance(trailing_type_spec, ASTTrailingTypeSpecName):
+                return None
+            nested_name = trailing_type_spec.name
+        else:
+            return None
+        symbols, failReason = self.parent.find_name(
+            nested_name, templateDecls=[], typ='any',
```
jakobandersen: I'm not sure if ``any`` is technically correct here, but I don't think it matters too much right now (as long as it is not ``class``).

jakobandersen (changes requested):
Sorry for the delay, here is an initial round of review.

- The first commit relaxes the lookup, but there are no tests. Can you add one?
- The type alias resolution does not seem to handle templated aliases, which is ok, but what happens with the lookup in that case? Can you add a test so we test the behavior?

Also, for eventually merging, please don't squash the commits in the end. The symbol resolution logic is complicated (because C++ is), and it really helps figuring out what is going on when things are separated in commits. If the commits are squashed it is impossible to untangle what the logic behind the changes were afterwards.

jbms:
Thanks for the review @jakobandersen.  I added a test for the relaxed lookup.  Regarding templated aliases, I believe I already have a test: `test_domain_cpp_resolve_template_through_type_alias_template`  Are you looking for something else?

AA-Turner:
Rebased to take account of the formatting changes in the C domain. @jakobandersen would you have time to look at this following Jeremy's comments?

A

jbms:
I rebased this.

Even though @jakobandersen hasn't had a chance to review this again, I think it would be reasonable to just merge it.

PR #10270, opened 2022-03-18
[C++] Support conditional operator "?"

jbms:
### Feature or Bugfix
- Feature

### Purpose
Adds support to the C++ domain for the conditional operator "?".  This operator is often used in template arguments.

PR #10302, opened 2022-03-27
C++, conditional expression (update of #10270)

jakobandersen:
Completion of #10270.

Note: env bumping was done earlier in 4.x so it's not needed here.


jbms:
Thanks for your revisions --- this looks good to me.

PR #10303, opened 2022-03-28
Fix translation of "alias of"

ultmaster:
Subject: Adding a blank space between "alias of" and python name.

### Feature or Bugfix

- Bugfix

### Purpose

To fix the complaint: `1: WARNING: Inline interpreted text or phrase reference start-string without end-string.`

This is because: ```:py:class:`_ast.Constant`的别名``` has no space between `` ` `` and `的别名`.

### Detail

Adding a blank space.

I used `msgfmt -c sphinx.po -o sphinx.mo` to generate the binary file. Not sure whether this is the correct way to do it.

### Relates

None.


tk0miya:
Thank you for your contribution. I applied this change to our project on transifex.com. It will automatically be applied to this repository soon.

Issue #10310, opened 2022-03-30
Sphinx automodule: class method not documented when decorated with mocked function


GutiCW:
### Describe the bug

I have a class with decorated methods. The function used as a decorator is provided by a module not available when building the documentation, hence this module is mocked in `conf.py` using `autodoc_mock_imports = ['module_with_decorator']`.

```
.. automodule:: my_class
    :members:
```

```
from module_with_decorator import my_decorator


class MyClass:
    """
    My test class.
    """

    @staticmethod
    def my_undecorated_static_method():
        """This is my_undecorated_static_method."""

    @staticmethod
    @my_decorator(param0, param1)
    def my_decorated_static_method():
        """This is my_decorated_static_method."""

    @classmethod
    def my_undecorated_class_method(cls):
        """This is my_undecorated_class_method."""

    @classmethod
    @my_decorator(param0, param1)
    def my_decorated_class_method(cls):
        """This is my_decorated_class_method."""

    def my_undecorated_instance_method(self):
        """This is my_undecorated_instance_method."""

    @my_decorator(param0, param1)
    def my_decorated_instance_method(self):
        """This is my_decorated_instance_method."""
```

The decorated class methods are not shown in the documentation - others are shown correctly.

[![Missing class method decorated with mocked function][1]][1]

Any suggestions regarding this issue?

  [1]: https://i.stack.imgur.com/c8cCe.png

### How to Reproduce

Decorate a class method with a function from a mocked module.
PowerShell:
$builder_name = "html" ; $doc_source = "...\test\doc\tp_testcase_docstring" ; $dest = "$doc_source\_build\$builder_name" ; conda activate $conda_env ; sphinx-build "$doc_source" "$dest" -b $builder_name

### Expected behavior

The class methods should be visible in the documentation.

### Your project

[test.zip](https://github.com/sphinx-doc/sphinx/files/8380987/test.zip)

### OS

Windows

### Python version

3.9.7

### Sphinx version

4.4.0

### Sphinx extensions

sphinx.ext.autodoc

### Additional context

Related to [stackoverflow.com](https://stackoverflow.com/questions/71613847/sphinx-automodule-class-method-not-documented-when-decorated-with-mocked-functi).

tk0miya:
At present, decorators on the mocked module are not working because they return mocked objects instead of functions or methods. I don't have good ideas to generate documents for them. So I marked this as "help wanted".

Please post your idea or send a PR to implement it. Thanks,

PR #10313, opened 2022-03-31
Support mocked decorator for class methods

GutiCW:
Subject: support mocked decorator for class methods

### Feature or Bugfix

- Feature

### Purpose
- Before this PR, class methods decorated with mocked functions were not shown in the generated documentation.

### Detail
- Consider bound methods in `ismock()` to handle this specific case.

### Relates
- #10310

### DIFF
--- a/sphinx/ext/autodoc/mock.py
+++ b/sphinx/ext/autodoc/mock.py
@@ -186,6 +186,7 @@ def ismock(subject: Any) -> bool:
 
     return False
 
+
 def undecorate(subject: _MockObject) -> Any:
     """Unwrap mock if *subject* is decorated by mocked object.
 
--- a/sphinx/ext/autodoc/mock.py
+++ b/sphinx/ext/autodoc/mock.py
@@ -9,7 +9,7 @@ from types import MethodType, ModuleType
 from typing import Any, Generator, Iterator, List, Optional, Sequence, Tuple, Union
 
 from sphinx.util import logging
-from sphinx.util.inspect import safe_getattr
+from sphinx.util.inspect import isboundmethod, safe_getattr
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


sphinx/ext/autodoc/mock.py
```diff
@@ -129,6 +129,14 @@ def invalidate_caches(self) -> None:
             sys.modules.pop(modname, None)
 
 
+def _method_is_bound(method: MethodType) -> bool:
+    """Check if the method is a bound method."""
+    try:
+        return method.__self__ is not None
```
tk0miya: Could you use `safe_getattr()` to access the `__self__` attribute? It prevents autodoc from crash when abnormal object raises an exception unexpectedly.

sphinx/ext/autodoc/mock.py
```diff
@@ -129,6 +129,14 @@ def invalidate_caches(self) -> None:
             sys.modules.pop(modname, None)
 
 
+def _method_is_bound(method: MethodType) -> bool:
```
tk0miya: I feel `isboundmethod()` is better for this. And it's much better to move this to `sphinx.util.inspect` module.

tk0miya (changes requested):
Could you add a testcase to `tests/test_ext_autodoc_mock.py`, please?

sphinx/ext/autodoc/mock.py
GutiCW: Done.

sphinx/ext/autodoc/mock.py
GutiCW: Done.

GutiCW:
> Could you add a testcase to `tests/test_ext_autodoc_mock.py`, please?

Done, if my test extension suffices.

GutiCW:
Fyi, with my last commit (5a293346cf4fecb8629487f14f3dc22838c1a3e6) I removed the recursive call as it is actually not necessary.

tk0miya (approved):

tk0miya:
Thank you for your contribution. Good work!

Issue #10295, opened 2022-03-25
An event after post-transform but before writing to filesystem?


cblegare:
**Is your feature request related to a problem? Please describe.**

I would like to be able to change something in the environment or in the builder *after* cross-reference resolution, but *before* writing the files.  In my case, I wanted to add items (resolved cross-references) to a Domain index.

**Describe the solution you'd like**

I don't know if this could be possible, especially to keep the parallel builds working the same, but maybe an event that is fired not for a specific document, after post-transforms, before write?

**Describe alternatives you've considered**

Wrote a PostTransformer that regenerate the Domain index after xref resolution.  It works, but it need to redo the same job over and over, since it can't either

- Know when it processing the last document
- Know when all cross-references have been resolved.

And there is no later hooks before writing to the filesystem (that I know of).

**Additional context**

- [sphinx-dev post](https://groups.google.com/g/sphinx-dev/c/GaVMIwOA0sg)

tk0miya:
Unfortunately, there is no chance to hook an event during xref resolution (post-transforming) and writing. Sphinx goes these two processes continuously for each document. For example, xref resolution for docB would be invoked after writing for docA. So it's hard to add such an event.

cblegare:
Hello and thank you for this response.

> Unfortunately, there is no chance to hook

Yes, I suspected that

I have found an other way that would make even more sense and solve my issue ultimately, and opened a new issue about it.

PR #10329, opened 2022-04-06
Closes #10299: generate domain index entries later (HTML)

### LINKED ISSUE #10299: Domain index entries could be generated later in the build process

Domain index entries could be generated later in the build process.

Domain indices, at least in HTML-based builders, are written during `Builder.finish` but generated during `Builder.prepare_writing`.  In other words

```
# Abstract builder when applicable, else HTML builder
Builder.build()
├── ...
├── Builder.write()
│   └── Builder.prepare_writing()
│       └── Index.generate()                # Here domain index entries are generated.
└── Builder.finish()
    └── Builder.gen_indices()
        └── Builder.write_domain_indices()  # Here domain index entries are used.
```

During `prepare_writing`, some class attributes from each domain Index are used, but not the generated content.  In other words, from what I have seen, the generated index entries could be generated (`Index.generate()`) directly during `Builder.write_domain_indices()`). In my opinion, it would make perfect sense to wait until the generated domain index entries are needed before generating them.  

Also, this enable domains to provide a wider variety of features along with their Index.

- https://github.com/sphinx-doc/sphinx/issues/10295
- Adding a finish task, but there is no API for this.  I would need to implement a custom builder.

**Additional context**

- [sphinx-dev post](https://groups.google.com/g/sphinx-dev/c/GaVMIwOA0sg)

---

In case you would consider this acceptable, I would probably write a patch since the required changes in the code seams quite minor and safe at first sight

cblegare:
Subject: Generate domain index entries as late as possible for HTML builders

### Feature or Bugfix

- Feature

### Purpose

During `prepare_writing`, some class attributes from each domain Index are used, but not the generated content.  In other words, from what I have seen, the generated index entries could be generated (`Index.generate()`) directly during `Builder.write_domain_indices()`). In my opinion, it would make perfect sense to wait until the generated domain index entries are needed before generating them.  

Also, this enable domains to provide a wider variety of features along with their Index.

### Relates
- https://github.com/sphinx-doc/sphinx/issues/10299
- https://github.com/sphinx-doc/sphinx/issues/10295

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -200,7 +200,7 @@ class StandaloneHTMLBuilder(Builder):
     download_support = True  # enable download role
 
     imgpath: str = None
-    domain_indices: List[Tuple[str, Type[Index], List[Tuple[str, List[IndexEntry]]], bool]] = []  # NOQA
+    domain_indices: List[Tuple[str, Index]] = []  # NOQA
 
     def __init__(self, app: Sphinx) -> None:
         super().__init__(app)
@@ -461,10 +461,7 @@ class StandaloneHTMLBuilder(Builder):
                     if isinstance(indices_config, list):
                         if indexname not in indices_config:
                             continue
-                    content, collapse = indexcls(domain).generate()
-                    if content:
-                        self.domain_indices.append(
-                            (indexname, indexcls, content, collapse))
+                    self.domain_indices.append((indexname, indexcls(domain)))
 
         # format the "last updated on" string, only once is enough since it
         # typically doesn't include the time of day
@@ -490,11 +487,11 @@ class StandaloneHTMLBuilder(Builder):
         rellinks: List[Tuple[str, str, str, str]] = []
         if self.use_index:
             rellinks.append(('genindex', _('General Index'), 'I', _('index')))
-        for indexname, indexcls, _content, _collapse in self.domain_indices:
+        for indexname, index in self.domain_indices:
             # if it has a short name
-            if indexcls.shortname:
-                rellinks.append((indexname, indexcls.localname,
-                                 '', indexcls.shortname))
+            if index.shortname:
+                rellinks.append((indexname, index.localname,
+                                 '', index.shortname))
 
         # back up script_files and css_files to allow adding JS/CSS files to a specific page.
         self._script_files = list(self.script_files)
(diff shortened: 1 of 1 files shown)


cblegare:
Hello again @tk0miya

The above PR implements possible solution number **4 Do not consider domain_indices to be in the API** from the comment in this issue: https://github.com/sphinx-doc/sphinx/issues/10299#issuecomment-1085341741

In case the approach is not suitable, please consider providing me with some guidance on what kind of approach would be, and I'll gladly submit another PR.

sphinx/builders/html/__init__.py
```diff
@@ -490,11 +487,11 @@ def prepare_writing(self, docnames: Set[str]) -> None:
         rellinks: List[Tuple[str, str, str, str]] = []
         if self.use_index:
             rellinks.append(('genindex', _('General Index'), 'I', _('index')))
-        for indexname, indexcls, _content, _collapse in self.domain_indices:
+        for indexname, index in self.domain_indices:
```
tk0miya, line 490: Before this change, `self.domain_indices` contains only domains having contents. On the other hand, it contains all domains. As a result, `rellinks` will contain hyperlinks to not existing pages. What I mentioned at https://github.com/sphinx-doc/sphinx/issues/10299#issuecomment-1084849148 is this. As you said, rellinks never use each index entry itself. But it implicitly checks the existence. So I consider this will break rellinks.

tk0miya (changes requested):

sphinx/builders/html/__init__.py
cblegare: You are right

My suggestion is to clean rellinks later on


https://github.com/sphinx-doc/sphinx/pull/10329/files#diff-be839d87d39d93093003a0a4ce40cf9610d0a9445e5e0aeb1cdc6bab527bfbbcR733

If this is not acceptable, I'll try to find an other way

tk0miya: Modifying rellinks after generating HTML is too late. It does not change the HTML files already written. You need to rewrite them all again. This is why we generate domain index entries before writing.

cblegare: thank you @tk0miya for your help and patience.

I made an error when analysing when what code was called at what time, and I though that it was not too late to remove empty domains from rellinks.  I since understood why I (wrongly) ended up to this conclusion.

I am sorry for wasting your time with this.  This leads me to two improvement ideas I'll open issues for. Thank you again.

PR #10337, opened 2022-04-09
Cache `Publisher` objects to speed up Sphinx

AA-Turner:
I don't think this is a big enough change to justify being a "feature". It does seem to be quite a big speed-up though (at least 20% faster).

Tests with `builddoc.yml` show 55s on `5.x` and 35s on this branch (~36% faster).

Tests on the main CI (Python 3.10, ubuntu) show 211s on `5.x` and 160s on this branch (~24% faster).

A

### Feature or Bugfix
- Bugfix


sphinx/builders/__init__.py
```diff
@@ -90,6 +95,9 @@ def __init__(self, app: "Sphinx") -> None:
         self.tags.add("format_%s" % self.format)
         self.tags.add("builder_%s" % self.name)
 
+        # cache of Docutils Publishers (file type -> publisher object)
+        self._publishers: Dict[str, docutils.core.Publisher] = {}
```
tk0miya: I prefer to store (cached) objects during the build to the registry (`sphinx.registry.SphinxComponentRegistry`). It's used as the common component store. I believe it's the more appropriate place. Could you move this? Then we can  keep using `sphinx.io.read_doc()`.

sphinx/builders/__init__.py
```diff
@@ -464,8 +504,20 @@ def read_doc(self, docname: str) -> None:
...
+        publisher = self._get_publisher(filetype)
+
         with sphinx_domains(self.env), rst.default_role(docname, self.config.default_role):
-            doctree = read_doc(self.app, self.env, self.env.doc2path(docname))
+            # set up error_handler for the target document
+            error_handler = UnicodeDecodeErrorHandler(docname)
+            codecs.register_error('sphinx', error_handler)  # type: ignore
+
+            publisher.set_source(source_path=filename)
+            publisher.publish()
```
tk0miya, line 476: I'm afraid reusing docutils components between multiple documents. At the first glance, the document object is generated at every build. So it's not shared. But reader, parser and writer objects are shared. Do you know these objects do not store any status? It seems nobody resets the status of them on the `publish()` call.

sphinx/builders/__init__.py
AA-Turner: I looked at the component registry but everything there seems to be storing types or functions, not instantiated classes. 

If you'd still prefer to move it I will do (I could add a helper `create_publisher` to the registry too)

A

sphinx/builders/__init__.py
AA-Turner: The reader, parser and writer objects do store state, but it is overwritten on every call to `.publish()`. 

We could explicitly reset the state of the objects after each call?

A

sphinx/builders/__init__.py
tk0miya: My large concern is not to add a new responsibility to the application object. Because it has already had a lot of responsibilities (in other words, methods and attributes). I feel creating a parser object is not an important responsibility of the "application" itself.

At present, the "reading" responsibility and its implementations are gathers to "sphinx.io" module. I believe it should not be moved to the app. This is why the cache should be stored to the registry. It's not important to use registry itself. But it should not be implemented to the app object.

sphinx/builders/__init__.py
tk0miya: Thank you for your explanation. Okay, let's go with this.

sphinx/builders/__init__.py
AA-Turner: I moved the set-up to the registry, how is it now?

A

AA-Turner:
CI passed.

A

sphinx/registry.py
```diff
@@ -125,6 +127,9 @@ def __init__(self) -> None:
         #: additional transforms; list of transforms
         self.transforms: List[Type[Transform]] = []
 
+        # private cache of Docutils Publishers (file type -> publisher object)
```
tk0miya, line 130: Basically, we don't use the `_` prefix for attribute names. From our experience, private markers are ignored by hackers. So we need to take care of both public and private methods.

sphinx/registry.py
```diff
@@ -461,6 +466,41 @@ def get_envversion(self, app: "Sphinx") -> Dict[str, str]:
         envversion['sphinx'] = ENV_VERSION
         return envversion
 
+    def create_publisher(self, app: "Sphinx", filename: str) -> Publisher:
```
tk0miya: I considers registry is a mere storage of components. So it should not have "features" as possible. So I'd like to move this `sphinx.io` back again.

tk0miya (changes requested):

sphinx/registry.py
AA-Turner: Done

sphinx/registry.py
AA-Turner: How's the current?

A

sphinx/io.py
```diff
@@ -153,30 +152,45 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
 
 
-def read_doc(app: "Sphinx", env: BuildEnvironment, filename: str) -> nodes.document:
+def read_doc(publisher: Publisher, docname: str, filename: str) -> nodes.document:
```
tk0miya: I'm not sure about the responsibility of the new `read_doc()`. This only calls the `publish()` method of the given publisher. I feel there is no reason to be separated from the env class. Is it better to merge this into it?

sphinx/io.py
```diff
@@ -153,30 +152,45 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
     # set up error_handler for the target document
-    error_handler = UnicodeDecodeErrorHandler(env.docname)
+    error_handler = UnicodeDecodeErrorHandler(docname)
     codecs.register_error('sphinx', error_handler)  # type: ignore
 
+    publisher.set_source(source_path=filename)
+    publisher.publish()
+
+    doctree = publisher.document
+    # settings get modified in ``write_doctree``; get a local copy
```
tk0miya: This mentions to the behavior of `env.write_doctree()`. It's better to move this to `env.read_doc()`.

BTW, what is the merit of cloning?

sphinx/registry.py
tk0miya: Thank you for your update. I prefer the idea of `create_publisher()`. +1 for this.

tk0miya (approved):
I commented a few, but it should not prevent merging this into the 5.0 release. So I'd like to merge this before feature freeze and refactor it after the beta release.

AA-Turner:
@tk0miya I merged the function into `Builder.read_doc`, I think that's what you meant?

> what is the merit of cloning?

I added a longer comment -- Docutils' publisher reuses the settings object, but in `Builder.write_doctree` we edit it for pickleability. I want to update Docutils to allow more of the behaviour Sphinx has to patch to happen natively, but that will take a little time.

A

AA-Turner:
CI now passing.

A

sphinx/io.py
```diff
@@ -153,30 +152,30 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
 
 
-def read_doc(app: "Sphinx", env: BuildEnvironment, filename: str) -> nodes.document:
```
tk0miya, line 156: I prefer to keep this as is and mark it deprecated. I'll send a change to your branch from now on.

sphinx/builders/__init__.py
```diff
@@ -464,8 +465,21 @@ def read_doc(self, docname: str) -> None:
...
+            codecs.register_error('sphinx', UnicodeDecodeErrorHandler(docname))  # type: ignore
+
+            publisher.set_source(source_path=filename)
+            publisher.publish()
+            doctree = publisher.document
+
+            # The settings object is reused by the Publisher for each document.
+            # Becuase we modify the settings object in ``write_doctree``, we
+            # need to ensure that each doctree has an independent copy.
+            doctree.settings = doctree.settings.copy()
```
tk0miya, line 482: Thank you for updating comment. I understand what you're going. I think it's much better to move this into `write_doctree()`.

tk0miya (approved):
LGTM with nits.

sphinx/io.py
tk0miya: Done.

tk0miya:
Thank you for your work. This must be the best improvement in 5.0!

Issue #10344, opened 2022-04-14
The example in the documentation in section 'Add your own static files to the build assets' is incorrect


Thecarisma:
### Describe the bug

In the [Add your own static files to the build assets](https://www.sphinx-doc.org/en/master/development/theming.html#add-your-own-static-files-to-the-build-assets) documentation example 

```python
from os import path
from sphinx.util.fileutil import copy_asset_file

def copy_custom_files(app, exc):
    if app.builder.format == 'html' and not exc:
        staticdir = path.join(app.builder.outdir, '_static')
        copy_asset_file('path/to/myextension/_static/myjsfile.js', staticdir)

def setup(app):
    app.connect('builder-inited', copy_custom_files)
```

When sphinx-build is executed it will fail with the error message

```python
Handler <function copy_custom_files at 0x00000250277A53F0> for event 'builder-inited' threw an exception (exception: copy_custom_files() missing 1 required positional argument: 'exc')
```

Likely because when the `builder-initiated` event is emitted only 1 positional argument *app* is invoked with the function. But when the event is changed to `build-finished` event, the two positional arguments *app* and *exc* is invoked with the function.

The solution is simply to change the event in the example from `builder-inited` to `build-finished`. or provide the appropriate example for `builder-inited` event.

### How to Reproduce

Simply execute the example in the [Add your own static files to the build assets](https://www.sphinx-doc.org/en/master/development/theming.html#add-your-own-static-files-to-the-build-assets) section of the documentation

### Your project

https://github.com/Thecarisma/themata

### OS

Any

### Python version

Any

### Sphinx version

Any

cblegare:
Hello @Thecarisma 

I am no maintainer here, just trying to help.  It seems to me that you are right, the example is wrong.  It also look superfluous, since files in a theme's `static` folder will all be copied to the built static folder.  See https://www.sphinx-doc.org/en/master/development/theming.html#creating-themes. Have you tried it?

I see that you opened a PR, which is great thank you. I also opened one, removing the section entirely. I suppose @tk0miya can choose whatever is best for Sphinx.

Finally, I've looked into these themes you are working on.  They're great :) thanks for this nice work!

Thecarisma:
Hi @cblegare 

I am aware the files in the static folder get copied to the built _static folder, but my use case is to copy a javascript file from the module folder which is outside the themes static folder to the built static folder.

[...]

PR #10345, opened 2022-04-14
Refractor the example in 'Add your own static files to the build assets' documentation section

Thecarisma:

### Feature or Bugfix

- Refactoring

### Purpose
Fix the incorrect example on the theming [Add your own static files to the build assets](https://www.sphinx-doc.org/en/master/development/theming.html#add-your-own-static-files-to-the-build-assets) documentation page

### Relates
- https://github.com/sphinx-doc/sphinx/issues/10344

### DIFF
--- a/doc/development/theming.rst
+++ b/doc/development/theming.rst
@@ -265,7 +265,7 @@ Here is some sample code to accomplish this:
            copy_asset_file('path/to/myextension/_static/myjsfile.js', staticdir)
 
    def setup(app):
-       app.connect('builder-inited', copy_custom_files)
+       app.connect('build-finished', copy_custom_files)
 
 
 Inject JavaScript based on user configuration


cblegare:
As [discussed](https://github.com/sphinx-doc/sphinx/issues/10344#issuecomment-1099761588), I would like to suggest also adding a comment on why this use-case is interesting, and not superfluous.

before

```rst
If you are packaging your own build assets with an extension
(e.g., a CSS or JavaScript file), you need to ensure that they are placed
in the ``_static/`` folder of HTML outputs.
```

after

```rst
If you are packaging your own build assets with an extension
(e.g., a CSS or JavaScript file), you need to ensure that they are placed
in the ``_static/`` folder of HTML outputs.
This can be necessary when the adding assets using the ``static/`` folder 
of a theme is not adequate, either because your extension is not a theme 
per-se or the static files are dynamically provided, for instance.
```

This is only a suggestion, and I may be wrong here.

tk0miya:
Merged. Thank you for your contribution.

PR #10352, opened 2022-04-16
Don't use deprecated `setup.py build_sphinx`

AA-Turner:
Refactor `tox.ini` to call `sphinx -M html` directly.

A

### Feature or Bugfix
- Refactoring

PR #10354, opened 2022-04-16
Enable the opt-in EncodingWarning

AA-Turner:
Whilst testing for #10352, building Sphinx failed for me on windows as the new `linkify_issues_in_changelog` uses the default encoding, which isn't UTF-8 on windows. This adds the new [opt-in EncodingWarning
](https://docs.python.org/3/library/io.html#io-encoding-warning) to check for such cases in the future.

A

### Feature or Bugfix
- Refactoring

### Relates
- https://docs.python.org/3/library/io.html#io-encoding-warning


Makefile
```diff
@@ -66,11 +66,11 @@ doclinter:
 
 .PHONY: test
 test:
-	@$(PYTHON) -X dev -m pytest -v $(TEST)
+	@$(PYTHON) -X dev -X warn_default_encoding -m pytest -v $(TEST)
```
tk0miya, line 69: Is it okay to pass this option to the old pythons?

tk0miya (approved):
LGTM with nits.

Makefile
AA-Turner: Yes, see https://docs.python.org/3/using/cmdline.html#cmdoption-X

> It also allows passing arbitrary values and retrieving them through the [sys._xoptions](https://docs.python.org/3/library/sys.html#sys._xoptions) dictionary.

So for Python 3.9 and earlier the value is just ignored.

A

tk0miya: Thank you for your explanation. Okay, let's merge :-)

tk0miya:
Note: fixed a conflict.

Makefile
AA-Turner: Thanks!

AA-Turner:
CI passed here too!

tk0miya:
Merged!

PR #10355, opened 2022-04-16
Move `babel` support to a dedicated utility file

AA-Turner:
Running arbitrary commands from `setup.py` [is deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html). This moves the babel and translations support to a decdicated file, `utils/babel_runner.py`, and updates documentation and the transifex workflow to match.

A

### Feature or Bugfix
- Refactoring


tk0miya (approved):
LGTM!

PR #10361, opened 2022-04-17
Only test with coverage on pushes to `sphinx-doc/sphinx`

AA-Turner:
To speed up CI for PRs, this moves the coverage tests to only run on pushes to branches in this repo, rather than on forks and pull requests too.

I also bumped `pytest` to 4.6 and removed some now-redundant code (we had previously always installed `pytest-cov`, which has required `pytest>=4.6` since 2020)

https://github.com/pytest-dev/pytest-cov/blob/694f7fd497b07aaea76d96e484068faf440e8301/setup.py#L125

A

### Feature or Bugfix
- Refactoring

### Relates
- Split from #10164

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,8 +47,7 @@ extras_require = {
         "types-requests",
     ],
     'test': [
-        'pytest',
-        'pytest-cov',
+        'pytest>=4.6',
         'html5lib',
         "typed_ast; python_version < '3.8'",
         'cython',
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,13 +18,10 @@ jobs:
           docutils: du16
         - python: "3.9"
           docutils: du17
-          coverage: "--cov ./ --cov-append --cov-config setup.cfg"
         - python: "3.10"
           docutils: du18
         - python: "3.11-dev"
           docutils: py311
-    env:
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -47,8 +47,7 @@
         "types-requests",
     ],
     'test': [
-        'pytest',
-        'pytest-cov',
+        'pytest>=4.6',
```
tk0miya, line 50: Note: pytest-4.6 has been packaged since Ubuntu 20.04. But I think it's okay because this dependency is for developers, not users.

| Ubuntu LTS | pytest version|
| --- | --- |
| 18.04 | 3.3.2 |
| 20.04 | 4.6.9 |
| 22.04 | 6.2.5 |

refs: https://packages.ubuntu.com/jammy/python3-pytest

tk0miya (approved):
Looks good. But it seems CI are not passed. Could you check them please?

AA-Turner:
> Looks good. But it seems CI are not passed. Could you check them please?

Updated, I forgot that generators aren't reversible. Seems to be passing now.

A

tk0miya:
Thank you for quick update. Merging now.

tk0miya:
@AA-Turner The coverage job was failed just after merging this. Could you check this, please?
https://github.com/sphinx-doc/sphinx/runs/6261200953?check_suite_focus=true

PR #10362, opened 2022-04-18
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:
Subject: add informational Codecov status checks

### Feature or Bugfix
- Refactoring

### Purpose
Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. I know there has been some churn on this file on having status checks at all, and I'd love to understand what has been so painful there.

Let me know if this makes sense or if we can do something that would be helpful.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,7 +3,7 @@ coverage:
   status:
     project:
       default:
-        enabled: no
+        informational: true
     patch:
       default:
-        enabled: no
+        informational: true


tk0miya:
Personally, I don't care about the score of coverage. So feedback is not needed (to me).

Additionally, we have a plan to invoke codecove just after merging PRs (refs: https://github.com/sphinx-doc/sphinx/pull/10361). So I guess this change will not work.

tk0miya:
Personally, I've not checked the coverages at all on review. Please ask to other maintainers. Thanks,

tk0miya:
As I commented above, we determined to invoke codecov after merges. So I'm closing this now. Thank you for your proposal!

tk0miya:
As I commented above, \*I\* 've not checked the coverage. So I don't get anything from codecov.

PR #10366, opened 2022-04-19
Allow emphasising placeholders in `option` directives

### LINKED ISSUE #9965: Allow a variable part in ..option directives

marxin:
Support parsing of "variable part" for option directives similarly
to samp roles.

Fixes: #9965.

This is a follow-up of #10001 which targeted `4.x` branch. I included @jakobandersen which were suggested in:
https://github.com/jakobandersen/sphinx/tree/marxin-option_parse_variable_part.

Apart from that, I have to handle empty `args` in `handle_signature` function.


jakobandersen:
Looks good to me.

> Apart from that, I have to handle empty `args` in `handle_signature` function.

Not exactly sure what you mean here. Do you have an example?

sphinx/domains/std.py
```diff
@@ -170,15 +172,42 @@ def handle_signature(self, sig: str, signode: desc_signature) -> str:
...
+                if args[0] == ' ':
+                    signode += addnodes.desc_sig_space()
+                    args = args.strip()
+                if args[0] == '=':
+                    signode += addnodes.desc_sig_punctuation('=', '=')
+                    args = args[1:]
+                for part in samp_role.parse(args):
+                    if isinstance(part, nodes.Text):
+                        text = part.astext()
+                        signode += nodes.Text(text, text)
```
AA-Turner: Avoid a DeprecationWarning
```suggestion
                        signode += nodes.Text(part.astext())
```

marxin:
> Looks good to me.

Thanks. Can you please mark it with a Github Approval?

> 
> > Apart from that, I have to handle empty `args` in `handle_signature` function.
> 
> Not exactly sure what you mean here. Do you have an example?

Sure, happens for situation like: `.. option:: -Wno-shift-overflow, -Wshift-overflow=n, -Wshift-overflow`. Which is parsed in the following way:
```
optname: "-Wno-shift-overflow", args: ""
optname: "-Wshift-overflow", args: "=n"
optname: "-Wshift-overflow", args: ""
```

doc/usage/restructuredtext/domains.rst
```diff
@@ -1750,6 +1750,9 @@ There is a set of directives allowing documenting command-line programs:
    referenceable by :rst:role:`option` (in the example case, you'd use something
    like ``:option:`dest_dir```, ``:option:`-m```, or ``:option:`--module```).
 
+   Use :confval:`option_detailed_parse` for parsing of
+   "variable part" of a literal text (similarly to ``samp`` directive).
```
AA-Turner:
```suggestion
   "variable part" of a literal text (similarly to the ``samp`` role).
```

doc/usage/configuration.rst
```diff
@@ -706,6 +706,15 @@ General configuration
 
     .. versionadded:: 3.0
 
+.. confval:: option_detailed_parse
+
+   Default is ``False``.
+   Parse option in more detail to provide a better markup, similarly
+   to ``samp`` role, support parsing of "variable part" of a literal text.
+   For example, in ``.. option:: -foption={TYPE}``, the part ``TYPE``
+   would be emphasized.
```
AA-Turner: The documentation is unclear to me -- what do you mean by "a literal text"? This only applies to options, so perhaps something like:

```suggestion
.. confval:: option_detailed_parse

   Default is ``False``.
   When enabled, emphasise placeholders in ``.. option:`` directives.
   To display literal braces, escape with a backslash (``\{``). For example, 
   ``option_detailed_parse=True`` and ``.. option:: -foption={TYPE}`` would 
   render with ``TYPE`` emphasised.
```

AA-Turner (commented):
Some comments on the documentation.

An aside, perhaps the option could be named `option_emphasise_placeholders` or similar? `parse_detailed` doesn't tell me much about what is going on, and it isn't just the parsing that changes with this option, but the output too.

A

marxin:
> An aside, perhaps the option could be named `option_emphasise_placeholders` or similar? `parse_detailed` doesn't tell me much about what is going on, and it isn't just the parsing that changes with this option, but the output too.

Thanks, I welcome the suggested changes, what about @jakobandersen?

marxin:
> An aside, perhaps the option could be named `option_emphasise_placeholders` or similar? `parse_detailed` doesn't tell me much about what is going on, and it isn't just the parsing that changes with this option, but the output too.

Well, I can imagine it can support more syntax for the future, so please let's use the more generic name `option_emphasise_placeholders`.

marxin:
I would really like to see this in `5.x` as I'm in the review process for almost 6 months.
Can you @AA-Turner and @jakobandersen make an approval of it? Thanks.

AA-Turner:
The implementation looks reasonable, although I still don't like 'detailed_parse', as users would likely want to configure any other 'detailed parsing' independently from this.

A

marxin:
> The implementation looks reasonable, although I still don't like 'detailed_parse', as users would likely want to configure any other 'detailed parsing' independently from this.

I don't have a problem with the option name suggestion, thus applied.

AA-Turner (approved):
Thanks @marxin! 

A

AA-Turner:
Thanks for your perseverance @marxin, merged.

A

marxin:
Thank you guys for merging this, really helps me.

Issue #10104, opened 2022-01-15
Message.locations duplicate unnecessary


ghost:
### Describe the bug

When running 

`make clean; make gettext`

there are times the list of locations is duplicated unnecessarily, example:

```
#: ../../manual/render/shader_nodes/vector/vector_rotate.rst:38
#: ../../manual/modeling/hair.rst:0
#: ../../manual/modeling/hair.rst:0
#: ../../manual/modeling/hair.rst:0
#: ../../manual/modeling/metas/properties.rst:92
```

or 

```
#: ../../manual/movie_clip/tracking/clip/toolbar/solve.rst:96
#: ../../manual/physics/dynamic_paint/brush.rst:0
#: ../../manual/physics/dynamic_paint/brush.rst:0
#: ../../manual/physics/dynamic_paint/brush.rst:0
#: ../../manual/physics/dynamic_paint/brush.rst:0
#: ../../manual/physics/dynamic_paint/canvas.rst:0
#: ../../manual/physics/dynamic_paint/canvas.rst:0
#: ../../manual/physics/dynamic_paint/canvas.rst:0
#: ../../manual/physics/dynamic_paint/canvas.rst:0
#: ../../manual/physics/dynamic_paint/canvas.rst:0
#: ../../manual/physics/dynamic_paint/canvas.rst:0
#: ../../manual/physics/fluid/type/domain/cache.rst:0
```
as shown in this screen viewing of the 'pot' file result:
 
<img width="1552" alt="Screenshot 2022-01-15 at 20 41 41" src="https://user-images.githubusercontent.com/16614157/149637271-1797a215-ffbe-410d-9b66-402b75896377.png">

After debugging a little, the problem appeared to be in the file:

[sphinx/builders/gettext.py](https://www.sphinx-doc.org/en/master/_modules/sphinx/builders/gettext.html)

in the '__init__' method.

My simple solution is this:

```
    def __init__(self, text: str, locations: List[Tuple[str, int]], uuids: List[str]):
        self.text = text
        # self.locations = locations
        self.locations = self.uniqueLocation(locations)
        self.uuids = uuids

    def uniqueLocation(self, locations: List[Tuple[str, int]]):
        loc_set = set(locations)
        return list(loc_set)
```
**Note,** _this solution will probably needed to be in the_

`babel.messages.pofile.PoFileParser._process_comment()`

_and in the_ 

`babel.messages.catalog.Message.__init__()`

_as well._

### How to Reproduce

[...]

PR #10379, opened 2022-04-23
Addressing problems raised in sphinx_issues #10104

hoangduytranuk:
Subject: Solving problems raised in issues #10104 
- Breaking changes: master
- Feature

### Purpose
- Allowing message locations to be sorted uniquely in alphabetic order. This will remove duplicates locations and enable easier search and identify locations for developers

### Detail
-  remove duplicates locations 
- enabled easier search and identify locations for developers
- 
### Relates
- [Message.locations duplicate unnecessary](https://github.com/sphinx-doc/sphinx/issues/10104)

### DIFF
--- a/sphinx/builders/gettext.py
+++ b/sphinx/builders/gettext.py
@@ -1,5 +1,5 @@
 """The MessageCatalogBuilder class."""
-
+import os
 from codecs import open
 from collections import OrderedDict, defaultdict
 from datetime import datetime, timedelta, tzinfo
@@ -32,7 +32,7 @@ class Message:
     """An entry of translatable message."""
     def __init__(self, text: str, locations: List[Tuple[str, int]], uuids: List[str]):
         self.text = text
-        self.locations = locations
+        self.locations = list(set(locations))
         self.uuids = uuids
 
 
@@ -57,7 +57,7 @@ class Catalog:
 
     def __iter__(self) -> Generator[Message, None, None]:
         for message in self.messages:
-            positions = [(source, line) for source, line, uuid in self.metadata[message]]
+            positions = [(os.path.relpath(source, start=os.getcwd()), line) for source, line, uuid in self.metadata[message]]
             uuids = [uuid for source, line, uuid in self.metadata[message]]
             yield Message(message, positions, uuids)
 
--- a/sphinx/builders/gettext.py
+++ b/sphinx/builders/gettext.py
@@ -32,13 +32,14 @@ class Message:
     """An entry of translatable message."""
     def __init__(self, text: str, locations: List[Tuple[str, int]], uuids: List[str]):
         self.text = text
-        self.locations = list(set(locations))
+        self.locations = list(sorted(list(set(locations))))     #make list unique and sorted alphabetically
         self.uuids = uuids
 
 
 class Catalog:
     """Catalog of translatable messages."""
-
+    # make a class var so any instance can share this parent class, where all building info can be found, such as srcdir etc..
+    builder = None
     def __init__(self) -> None:
         self.messages: List[str] = []  # retain insertion order, a la OrderedDict
 
@@ -56,8 +57,9 @@ class Catalog:
         self.metadata[msg].append((origin.source, origin.line, origin.uid))  # type: ignore
 
     def __iter__(self) -> Generator[Message, None, None]:
+        src_dir = Catalog.builder.srcdir
         for message in self.messages:
-            positions = [(os.path.relpath(source, start=os.getcwd()), line) for source, line, uuid in self.metadata[message]]
+            positions = [(relpath(source, start=src_dir), line) for source, line, uuid in self.metadata[message]]
             uuids = [uuid for source, line, uuid in self.metadata[message]]
             yield Message(message, positions, uuids)
 
@@ -213,7 +215,8 @@ class MessageCatalogBuilder(I18nBuilder):
     def init(self) -> None:
         super().init()
         self.create_template_bridge()
-        self.templates.init(self)
+        self.templates.init(self)        
+        Catalog.builder = self  # make the builder visible to instances of catalog
 
     def _collect_templates(self) -> Set[str]:
         template_files = set()


sphinx/builders/gettext.py
```diff
@@ -32,7 +32,7 @@ class Message:
     """An entry of translatable message."""
     def __init__(self, text: str, locations: List[Tuple[str, int]], uuids: List[str]):
         self.text = text
-        self.locations = locations
+        self.locations = list(set(locations))
```
AA-Turner: I'd use `sorted` here -- `set` doesn't guarantee any sort order at a language level.

```suggestion
        self.locations = sorted({*locations})
```

sphinx/builders/gettext.py
```diff
@@ -57,7 +57,7 @@ def add(self, msg: str, origin: Union[Element, "MsgOrigin"]) -> None:
 
     def __iter__(self) -> Generator[Message, None, None]:
         for message in self.messages:
-            positions = [(source, line) for source, line, uuid in self.metadata[message]]
+            positions = [(os.path.relpath(source, start=os.getcwd()), line) for source, line, uuid in self.metadata[message]]
```
AA-Turner: This would call `os.getcwd()` for every item in the loop. Does ommitting `start=` still give a usable result? Otherwise you could use cache `os.getcwd()` in a local.

sphinx/builders/gettext.py
```diff
@@ -1,5 +1,5 @@
 """The MessageCatalogBuilder class."""
-
+import os
```
AA-Turner, line 2: Keep a blank line here
```suggestion

import os
```

AA-Turner (changes requested):
A few comments:

A

sphinx/builders/gettext.py
tk0miya: Understandable change.

But this should be moved to the outside of this class because this class does not have appropriate directory information. I think what directory you want is the base directory of the document (a.k.a srcdir). But it seems you're using `os.getcwd()`. But it does not equal to srcdir. It usually equals to the confdir that contains conf.py, but not promised. It sometimes returns a different directory if 3rd party extension changes it.


BTW, please use `sphinx.util.osutil:relpath()` instead of `os.path.relpath()`. It will not be crashed even if `path` and `start` are on different drives on Windows.

hoangduytranuk:
Thank you so much for these feedbacks. I will return to fix the code at a later time and post another commit to the pull request. 

Best Regards,
Hoang Tran

sphinx/builders/gettext.py
hoangduytranuk: Fixed the issue #10104 after the code reviews from Adam Turner and Takeshi Komiya:
- Added the class variable 'builder' to Catalog class, so any instances of Catalog can access the information within the builder, including things such as app, config, srcdir, outdir etc.. see MessageCatalogBuilder for details.
- Initialise this Catalog.builder as the instance of MessageCatalogBuilder is created, ie. in the init()
- As Message class is created, list of locations passing to it will be made singular, using set(), sorted alphabetically for easier searching and identifying origins where the message can be found.
- source dir is used within the __iter__ of the Catalog, using sphinx.util.osutil.relpath, for safer functionality, to work out the relative path for each location entry.

sphinx/builders/gettext.py
```diff
@@ -32,13 +32,14 @@ class Message:
     """An entry of translatable message."""
     def __init__(self, text: str, locations: List[Tuple[str, int]], uuids: List[str]):
         self.text = text
-        self.locations = locations
+        self.locations = list(sorted(list(set(locations))))     #make list unique and sorted alphabetically
```
AA-Turner, line 35: `sorted` returns a list and can take a set as input

```suggestion
        self.locations = sorted(set(locations))  # sort and de-duplicate the locations list
```

sphinx/builders/gettext.py
AA-Turner: > Added the class variable 'builder' to Catalog class

Modifying global state like this is a bad idea, please could you refactor?

A

tk0miya: Yes, adding a class variable is not a good idea. `Catalog` class is a data class that represents message catalog. It is not related to the builder at all. So the path manipulation using builder should be called outside of the catalog class. How about processing just before calling `GettextRenderer`?

tk0miya:
Finally, I reimplemented this as https://github.com/sphinx-doc/sphinx/issues/10104 and merged it. Thank you for your contribution!

hoangduytran:
I would like to add a little comment as I only found out today after using 

`catalog.dump_po(out_path, data, line_width=4069, sort_output=self.is_sorted)
`
I found that user comments are repeated and I had to enter the following block of code to change the behaviour:


    ```
def _write_comment(comment, prefix=''):
        # NEVER wrap comments, this observation: "xgettext always wraps comments even if --no-wrap is passed;" is FALSE. There seemed to be a bug in the xgettext code, because wrapping doesn't always occur
        # Make sure comments are unique and sorted alphabetically so locations can be easily searched and identify
        has_comment = (bool(comment) and len(comment) > 0)
        if not has_comment:
            return

        is_list_type = isinstance(comment, list)
        comment_list = (list(comment) if not is_list_type else comment)
        comment_set_list = list(set(comment_list))
        comment_set_list.sort()

        for line in comment_set_list:
            _write('#%s %s\n' % (prefix, line.strip()))
```

to remove duplicated user comments

**Comment by hoangduytran** [comment]
Sorry the above solution is NOT perfect yet. I have just found another incident where the 

```
message.user_comments 
message.auto_comments
```

are duplicated, by simply read/write this file:

[https://gitlab.com/inkscape/inkscape/-/jobs/2544777172/artifacts/file/po/inkscape.pot](https://gitlab.com/inkscape/inkscape/-/jobs/2544777172/artifacts/file/po/inkscape.pot) 

The line repeated was:

`# TRANSLATORS: "Swatches" -> color samples`

I have to use the PyCharmIDE to debug and found out, so the correction code is this:

```
sort_by = None
    if sort_output:
        sort_by = "message"
    elif sort_by_file:
        sort_by = "location"

    for message in _sort_messages(catalog, sort_by=sort_by):
        if not message.id:  # This is the header "message"
            if omit_header:
                continue
            comment_header = catalog.header_comment
            if width and width > 0:
                lines = []
                for line in comment_header.splitlines():
                    lines += wraptext(line, width=width,
                                      subsequent_indent='# ')
                comment_header = u'\n'.join(lines)
            _write(comment_header + u'\n')

        **comment_list = []
        comment_list.extend(message.user_comments)
        comment_list.extend(message.auto_comments)
        comment_list_unique = list(set(comment_list))
        comment_list_unique.sort()**

        _write_comment(comment_list_unique)
```
the section surrounded by '**". This is an extraction from the file 

`site-packages/babel/messages/pofile.py`

routine:

` def _write_message(message, prefix=''):`

Ok, I'm out.

PR #10389, opened 2022-04-25
Replace doclinter with sphinx-lint

JulienPalard:
Subject: Just ran [sphinx-lint](https://github.com/sphinx-contrib/sphinx-lint) on Sphinx own doc.

Unsurprisingly I found more issues in `sphinx-lint` than in `Sphinx doc` while doing so ;-)

But it found a real one:
```diff
-   For Sphinx version 3.5.1 beta 3 this would be `(3, 5, 1, 'beta', 3)``.
+   For Sphinx version 3.5.1 beta 3 this would be ``(3, 5, 1, 'beta', 3)``.
```

so I'm happy. Other things were trailing whitespaces, and so on.


AA-Turner (approved):
Is `sphinx-lint` at a point where we can add it to CI for core Sphinx?

A

JulienPalard:
> Is `sphinx-lint` at a point where we can add it to CI for core Sphinx?

Yes, multiple projects already does it (cpython, pandas, sympy, ...) but I would recommend not having it mandatory yet for Sphinx, as it's typically here that we can find surprising things like rst-in-rst ;)

Do you want me to add the github workflow in the same PR?

AA-Turner:
> Do you want me to add the github workflow in the same PR?

Please, currently Sphinx uses a [custom linter](https://github.com/sphinx-doc/sphinx/blob/5.x/utils/doclinter.py) via tox:

https://github.com/sphinx-doc/sphinx/blob/5eeeb9ce375e2f35807b54f17cec64696a7828d7/tox.ini#L88-L95

I think using a standard tool would be beneficial.

A

JulienPalard:
Updated the Makefile:

```
$ make doclinter
sphinx-lint CHANGES *.rst doc/
No problems found.
```

Updated tox.ini:

```
$ tox -e docslint
docslint create: /home/mdk/clones/sphinx/.tox/docslint
docslint develop-inst: /home/mdk/clones/sphinx
[...]
docslint run-test-pre: PYTHONHASHSEED='2544686718'
docslint run-test: commands[0] | sphinx-lint CHANGES CONTRIBUTING.rst README.rst doc/
No problems found.
_____________________________________________________________________________________________________ summary _____________________________________________________________________________________________________
  docslint: commands succeeded
  congratulations :)
```

And Github Actions side I just bumped the Python version from 3.6 to 3.7 in the lint workflow, because sphinx-lint requires it.

.github/workflows/lint.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.7
```
AA-Turner: I'd just run on the latest Python, saves updating every year for EOL dates
```suggestion
        python-version: 3
```

AA-Turner (approved):

.github/workflows/lint.yml
JulienPalard: Modified to `>=3.7` just in case there exist runners with 3.6 would be picked by default.

utils/doclinter.py
```diff
@@ -1,77 +0,0 @@
-"""A linter for Sphinx docs"""
-
-import os
-import re
-import sys
-from typing import List
-
-MAX_LINE_LENGTH = 85
```
tk0miya, line 8: It seems `sphinx-lint` does not check the length of lines. So I'd not like to remove this script from our repository. How about using both?

tk0miya (changes requested):

utils/doclinter.py
AA-Turner: Perhaps `sphinx-lint` could add a line-length option?

Otherwise we could simplify to e.g.:

```python
"""Check line length in rst documents."""

import os
import re
import sys

MAX_LINE_LENGTH = 85


def check_line_length(path: str) -> bool:
    with open(path) as f:
        document = f.readlines()

    error = False
    for i, line in enumerate(document):
        if len(line) > MAX_LINE_LENGTH:
            if not (
                re.match(r'^\s*\.\. ', line)  # ignore directives and hyperlink targets
                or re.match(r'^\s*__ ', line)  # ignore anonymous hyperlink targets
                or re.match(r'^\s*``[^`]+``$', line)  # ignore a very long literal string
            ):
                print(f'{path}:{i + 1}: the line is too long ({len(line)} > {MAX_LINE_LENGTH}).')
                error = True

    return error


if __name__ == '__main__':
    error = False
    for path in sys.argv[1:]:
        if os.path.isfile(path):
            error ^= check_line_length(path)
        elif os.path.isdir(path):
            for root, _dirs, files in os.walk(path):
                for filename in files:
                    if filename.endswith('.rst'):
                        path = os.path.join(root, filename)
                        error ^= check_line_length(path)
    raise SystemExit(error + 0)
```

A

tk0miya: I'd not like to reduce our checks for docs. So it's very nice if sphinx-lint support it. Until then, it's also okay to remove duplicated features from doclinter.py.

JulienPalard: Ohhh yes good catch.

We do have this checker, but it's not enabled by default:

https://github.com/sphinx-contrib/sphinx-lint/blob/b25d09c5a0cf15030c575d18078d815724a1e3e0/sphinxlint.py#L316

And if we enable it using `--severity 0` it'll whine about other things that you typically allow like using default roles.

We'll have to implement a way to enable/disable checks by name instead of by severity level for more granularity.

I'll be offline the next week, but there's no rush anyway.

JulienPalard: Fixed using `--enable line-too-long --max-line-length 85`.

JulienPalard:
CI fails due to https://github.com/psf/requests/issues/6140.

AA-Turner (approved):

JulienPalard:
The CI could probably be fixed by changing https://docs.python-requests.org to https://requests.readthedocs.io, I don't know if there's a precedent on how to handle this? Maybe just wait a few hours. I'm following upstream's issue.

nateprewitt:
@JulienPalard I'd advise moving the domain to https://requests.readthedocs.io if that's acceptable. That will be the primary domain going forward as there isn't a clear path to Kenneth retrieving the python-requests.org domain.

AA-Turner:
Thanks Nate, we went ahead and did this in #10484. Thanks for confirming https://requests.readthedocs.io/ will be the main domain going ahead.

A

AA-Turner:
I closed & re-opened but just realised the fix won't be in this branch so it'll most likely fail again! Sorry for the noise.

A

.github/workflows/lint.yml
```diff
@@ -13,9 +13,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v3
```
AA-Turner:
```suggestion
      uses: actions/setup-python@v4
```

AA-Turner (approved):
One note, looks good otherwise.

A

tk0miya (approved):
Looks good. Thank you for developing great tool!

AA-Turner:
@JulienPalard it is failing on the below, which is valid reST ([rendered](https://www.sphinx-doc.org/en/master/templating.html#sphinx_version_tuple)). I think check_missing_space_after_literal needs to be updated to ignore `[\-.,:;!?\\/'")]}>]`, see item 7 of the [inline markup recognition rules](https://www.docutils.org/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules).

```restructuredtext
   The version of Sphinx used to build represented as a tuple of five elements.
   For Sphinx version 3.5.1 beta 3 this would be ``(3, 5, 1, 'beta', 3)``.
   The fourth element can be one of: ``alpha``, ``beta``, ``rc``, ``final``.
   ``final`` always has 0 as the last element.
```

For now I will disable that check and merge if CI passes, as it is a good improvement.

A

tox.ini
```diff
@@ -81,9 +81,9 @@ basepython = python3
 description =
     Lint documentation.
 extras =
-    docs
+    lint
 commands =
-    python utils/doclinter.py CHANGES CONTRIBUTING.rst README.rst doc/
+    sphinx-lint --enable line-too-long --max-line-length 85 CHANGES CONTRIBUTING.rst README.rst doc/
```
AA-Turner:
```suggestion
    sphinx-lint --disable missing-space-after-literal --enable line-too-long --max-line-length 85 CHANGES CONTRIBUTING.rst README.rst doc/
```

JulienPalard:
> @JulienPalard it is failing on the below, which is valid reST ([rendered](https://www.sphinx-doc.org/en/master/templating.html#sphinx_version_tuple)).

Can't reproduce:

https://github.com/sphinx-doc/sphinx/pull/10573

PR #10391, opened 2022-04-27
Deprecate `sphinx.util.jsdump`

AA-Turner:
`json` from the standard library works as a drop-in replacement.

This means less code to maintain and a likely better implementation (the Chinese tests now work loading the index, so there's better language support at the very least).

I'd prefer to remove it entirely with no deprecation period, as it is an internal utility module and there doesn't seem to be [any usage of it publicly](https://cs.github.com/?q=jsdump+language%3APython+NOT+path%3A%2F%5Cbsphinx%5Cb%2F&p=2&pt=ddf1eedd43c11bb4c1934cd707737d1865fa6fb81e7b5f0bc0490d44cd34611c07f18566770b2758ece3cb5db3a56e13db44b52627f90d840ebc33bdbca074efe18591365471a5d2089c9968f1a56701fa9e95b95e0cb2ade64cd770038737ac36c4a7ef3bb96b50f08a117ac59b3199a7b2b0df9d41b4f723fa5971f823811cd93c6ecd8f98c79a024e6941cd2ff76ffc5fb7e0ba8fc80d802602d5a4cefbc61ceb6d55dc4f546c4c74ac248d31f1811b1fba7815a045dafb7d47d6c3b5ab7ba8af724721c0badbf1825dfa5f8e463023d6b933aa315c4019b322dd18407875fcd5e1f9aa17b85dbb6a589523fec2edcc18c37b&scope=&scopeName=All+repos), but I don't know if that is allowed -- @tk0miya I would appreciate your view here if possible? 

A

### Feature or Bugfix
- Refactoring


tk0miya (commented):
+1 for deprecation. Foradays, no reason to have our own JSON implementation.
-0 for removing it on 6.0. No reason to hurry. Let's mark it deprecated and will remove it on a normal pace.

sphinx/util/jsdump.py
```diff
@@ -1,194 +1,21 @@
...
-
-
-def load(f: IO) -> Any:
-    return loads(f.read())
+"""This module implements a simple JavaScript serializer."""
+import json
+
+from sphinx.deprecation import RemovedInSphinx60Warning, deprecated_alias
+
+deprecated_alias(
```
tk0miya: I think it's enough to add a `warnings.warn()` call to the top of the module without rewriting (replacing). Nobody harms by old implementation.

AA-Turner:
Adjusted to 7.0 and restored the `jsdump` implementation, thanks!

A

sphinx/search/__init__.py
```diff
@@ -224,7 +226,7 @@ class IndexBuilder:
     passed to the `feed` method.
     """
     formats = {
-        'jsdump':   jsdump,
+        'json':     json,
```
tk0miya, line 229: I noticed sphinxcontrib-serializinghtml package depends on "jsdump" format. So it should be also updated after v5.0 released.

sphinx/util/jsdump.py
```diff
@@ -192,3 +194,21 @@ def loads(x: str) -> Any:
...
+        'load': load,
+    },
+    RemovedInSphinx70Warning,
+    {
+        'dumps': 'json.dumps',
+        'dump': 'json.dump',
+        'loads': 'json.loads',
+        'load': 'json.load',
+    }
+)
```
tk0miya: Please use `warnings.warn()` instead of `deprecated_alias()` for this case. `deprecated_alias()` lets developers know the specified attributes and functions has been deprecated and will be removed on accessing them.

In this case, we'd like to deprecate the whole of the module. So it's better to warn the deprecation on importing.
```suggestion
warnings.warn('sphinx.util.jsdump has been deprecated. Please use "json" instead.', RemovedInSphinx70Warning)
```

tk0miya (commented):
LGTM with nits.

sphinx/util/jsdump.py
AA-Turner: @tk0miya done

A

tk0miya (approved):
LGTM. Thank you for your work and very quick update!

Issue #10392, opened 2022-04-27
Configure 'sphinx.testing.fixtures' as a global fixture


stephenfin:
pytest allows you to provides fixtures that are [reusable by others](https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html#making-your-plugin-installable-by-others). This simply requires configuring an entry point. We should do this as doing so would make using this module for testing extensions slightly easier.

This is somewhat related to #7008

Issue #7008, opened 2020-01-09
sphinx.testing is undocumented


goerz:
https://github.com/sphinx-doc/sphinx-testing is marked deprecated and points to `sphinx.testing` as a replacement. However, the only documentation for this is https://www.sphinx-doc.org/en/master/devguide.html#unit-testing, which does not provide any details.

My first assumption was that maybe it's sufficient to replace `from sphinx_testing import with_app` with `from sphinx.testing import with_app`, but that does not seem to be the case:
~~~
ImportError: cannot import name 'with_app' from 'sphinx.testing'
~~~

I can sort of figure out the use of `sphinx.testing` by looking at the source code for Sphinx' own tests, but nonetheless there should be proper documentation. The (closed) issue #3458 already pointed out that documentation is still a todo item.

tk0miya:
Absolutely. It must be a task for us!

tk0miya:
As a quick note:

* Please create `tests/conftest.py` with following content:
```
import pytest

from sphinx.testing.path import path

pytest_plugins = 'sphinx.testing.fixtures'


@pytest.fixture(scope='session')
def rootdir():
    return path(__file__).parent.abspath() / 'roots'
```
* It provides absolute path of root directory of sphinx projects for testing as `rootdir()`. With above settings, they are placed under `tests/roots`.
* Please put sphinx projects for testing under `tests/roots/test-*`.  By default, `test-root` is used for testing.
* Please write your test code.
```
import pytest

def test(app):
    # app is a Sphinx application object for default sphinx project (`tests/roots/test-root`).
    app.build()

@pytest.mark.sphinx(buildername='latex')
def test_latex(app):
     # latex builder is chosen here.
     app.build()

@pytest.mark.sphinx(testroot='case1')
def test_case1(app):
    # app is Sphinx application for case1 sphinx project (`tests/roots/test-case1`)
    app.build()

@pytest.mark.sphinx(confoverrides={'master_doc': 'content'})
def test_confoverrides(app):
    # a Sphinx application configured with given setting
    app.build()
```

I'll make a document in detail if I have time...

jnikula:
Proper documentation for `sphinx.testing` would really be appreciated.

I'm banging my head against the wall migrating just this simple `@with_app` usage from `sphinx-testing`:

```
@with_app(confdir=testdir, create_new_srcdir=True, buildername='text')
```

[...]

PR #10393, opened 2022-04-27
Configure Sphinx pytest fixtures as globally accessible

stephenfin:
Makes the fixtures in `sphinx.testing.fixtures` globally accessible whenever Sphinx is installed,
as seen in the [pytest documentation](https://docs.pytest.org/en/7.1.x/how-to/writing_plugins.html#making-your-plugin-installable-by-others).

### Feature or Bugfix

- Feature

### Purpose

Allow users to use `sphinx.testing.fixtures` without having to configure `pytest_plugins` in a `conftest.py` file.

### Detail

- #10392

### Relates

- #7008

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -212,6 +212,7 @@ setup(
         'Framework :: Sphinx',
         'Framework :: Sphinx :: Extension',
         'Framework :: Sphinx :: Theme',
+        'Framework :: Pytest',
         'Topic :: Documentation',
         'Topic :: Documentation :: Sphinx',
         'Topic :: Internet :: WWW/HTTP :: Site Management',
@@ -241,6 +242,9 @@ setup(
         ],
         'distutils.commands': [
             'build_sphinx = sphinx.setup_command:BuildDoc',
+        ],
+        'pytest11': [
+            'sphinx = sphinx.testing.fixtures',
         ],
     },
     python_requires=">=3.6",


stephenfin:
Hmm, we probably want to prefix these fixtures with a `sphinx_` prefix. Going to mark this as draft until I get to do that.

AA-Turner:
https://github.com/thisch/pytest-sphinx seems to use the `sphinx` name as a PyTest plugin.

Would officially adopting `pytest-sphinx` be an appropriate solution, or would it be better to do as proposed here? (*Edit:* it seems that project is actually about running [`sphinx.ext.doctest`](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#module-sphinx.ext.doctest) tests with PyTest rather than exposing Sphinx's own test fixtures, but regardless it uses the `sphinx` name)

A

stephenfin:
> [thisch/pytest-sphinx](https://github.com/thisch/pytest-sphinx) seems to use the sphinx name as a PyTest plugin.
>
> Would officially adopting pytest-sphinx be an appropriate solution, or would it be better to do as proposed here? (Edit: it seems that project is actually about running [sphinx.ext.doctest](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#module-sphinx.ext.doctest) tests with PyTest rather than exposing Sphinx's own test fixtures, but regardless it uses the sphinx name)

~Are pytest namespaces required to be globally unique? I assume the issue here would be that if you installed both `sphinx` and `pytest-sphinx` in the same venv (which you would have to do for the latter to work), one would clobber the other?~ _[edit: rhetorical question, of course: I know they will]_ I wonder if we can work with @thisch to get their fixtures integrated into Sphinx core, assuming they're useful enough (no guaranteed of course: I'm in the process of evaluating the general usefulness of Sphinx's *own* fixtures to decide which of them should be exposed globally). If not, I'd be tempted to ask @thisch to rename their entrypoint and grab `sphinx` on the basis that it is our project name :slightly_smiling_face:

twmr:
I'm totally in favor of integrating the features of pytest-sphinx into sphinx core.  Actually, I have some problems with my plugin, due to the simple way of parsing docstrings/rst files. Using the sphinx infrastructure would definitely help to improve the parsing code.  There is also a problem with mapping the grouping concept of doctest (I mean the `[group]` that can be specified in various directives), because it is IMO impossible to map this concept to pytest equivalents.

I don't yet know what features you fixture(s) have. For me it is also possible to rename pytest-sphinx to e.g. pytest-sphinx-doctest, but I would prefer integrating it into the sphinx ecosystem and exposing it as a pytest plugin.

stephenfin:
@thisch Great to hear. Let me try to work on this over the next week or two and see if I can get something useful out (i.e. a global set of pytest fixtures plus some documentation on how to use it). If I do, adding the features of `pytest-sphinx` here would be great addition :slightly_smiling_face:

setup.py
```diff
@@ -242,6 +243,9 @@ def _run_domain_js(self, domain):
         'distutils.commands': [
             'build_sphinx = sphinx.setup_command:BuildDoc',
         ],
+        'pytest11': [
```
tk0miya, line 246: Note for me: this entrypoint is described in pytest docs.
https://docs.pytest.org/en/6.2.x/writing_plugins.html#making-your-plugin-installable-by-others

tk0miya:
+1 for adding the entrypoint. It must be useful.

+0 for giving a new name. I'm okay to do that. But the sphinx.testing module has already been used widely in many extensions. So we have to be careful to rename it not to break the compatibility of the module.

-1 for merging pytest-sphinx to the core. At a first glance, the goal of sphinx.testing and it. So it's difficult to merge it into the Sphinx core. I believe it's out of the purpose of Sphinx.

PR #10418, opened 2022-05-03
Handle functions wrapping callable objects as functions

### LINKED ISSUE #10414: FunctionDocumenter rejects functions wrapping callable objects
### Describe the bug

Functions that wrap callable objects are not properly documented as functions (i.e. they are missing a function signature, see screenshots below)

The `FunctionDocumenter.can_document_member` ends up calling this `is_function` precidate:

https://github.com/sphinx-doc/sphinx/blob/ef665d37fa84788c6e0e08c1a1dac6f78be0de3d/sphinx/util/inspect.py#L305-L307

The issue is then that that predicate attempts to "unwrap" the function:

https://github.com/sphinx-doc/sphinx/blob/bd4b19451c9697b3deec687b8e4c893c3916ad81/sphinx/util/inspect.py#L53-L70

This `unwrap_all` does not recognize the case when a function wraps callable objects, and as a result ends up using `DataDocumenter` instead. Among other things, this prevents the function signature from being documented properly. 

### How to Reproduce

Reproducing on the https://github.com/nv-legate/cunumeric repo requires a fairly extensive compiler toolchain including CUDA support. An MRE that illustrates that `isfunction` fail to recognized functions that wrap callable objects (directly analogous to the ones in our case)  is given below:

```
In [1]: from sphinx.util.inspect import isfunction

In [2]: from functools import wraps

In [3]: class Functor:
   ...:     def __call__(self, a, b):
   ...:         print(a, b)
   ...: 

In [4]: def decorator(f):
   ...:     @wraps(f)
   ...:     def wrapper(*args, **kw):
   ...:         return f(*args, **kw)
   ...:     return wrapper
   ...: 

In [5]: f = Functor()

In [6]: wrapped_f = decorator(f)

In [7]: isfunction(wrapped_f)
Out[7]: False
```

### Expected behavior

`FunctionDocumenter` should be selected and used for functions that wrap callable objects. 

It's possible this could be resolved by generalizing `sphinx.util.inspect.isfunction` or alternatively, `FunctionDocumenter.can_document_member` could add additional conditions to pass. 

### Your project

https://github.com/nv-legate/cunumeric

### Screenshots

Current (bad) output for a function wrapping a callable object:

image

Desired output:

image

The "fixed" version above was created using a workaround of a `FunctionDocumenter` subclass that can recognize our callable objects, ref: https://github.com/nv-legate/cunumeric/pull/317

### OS

Linux

### Python version

3.7+

### Sphinx version

4.4.0

### Sphinx extensions

   "sphinx.ext.autodoc",     "sphinx.ext.autosummary",     "sphinx.ext.intersphinx",     "sphinx.ext.mathjax",     "sphinx.ext.napoleon",     "sphinx_copybutton",     "sphinx_markdown_tables",     "recommonmark",

### Extra tools

_No response_

### Additional context

_No response_

bryevdv:
- bugfix (fixes #10414)

cc @AA-Turner @tk0miya

### DIFF
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -61,8 +61,17 @@ def unwrap_all(obj: Any, *, stop: Callable = None) -> Any:
         elif ispartial(obj):
             obj = obj.func
         elif inspect.isroutine(obj) and hasattr(obj, '__wrapped__'):
+            # Don't unwrap wrapped callable types:
             if hasattr(obj.__wrapped__, '__call__'):
-                return obj  # Don't unwrap wrapped callable types
+                # From Python 3.10, staticmethod and classmethod have
+                # "__wrapped__" and "__wrapped__.__call__" attributes.
+                # Check that we are in Python 3.9 or earlier, or that the object
+                # isn't a staticmethod or classmethod before returning.
+                # xref: https://docs.python.org/3.10/whatsnew/3.10.html#other-language-changes
+                if sys.version_info[:2] < (3, 10):
+                    return obj
+                if not (isstaticmethod(obj) or isclassmethod(obj)):
+                    return obj
             obj = obj.__wrapped__  # type: ignore
         elif isclassmethod(obj):
             obj = obj.__func__
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -61,6 +61,8 @@ def unwrap_all(obj: Any, *, stop: Callable = None) -> Any:
         elif ispartial(obj):
             obj = obj.func
         elif inspect.isroutine(obj) and hasattr(obj, '__wrapped__'):
+            if hasattr(obj.__wrapped__, '__call__'):
+                return obj  # Don't unwrap wrapped callable types
             obj = obj.__wrapped__  # type: ignore
         elif isclassmethod(obj):
             obj = obj.__func__
--- a/tests/roots/test-ext-autodoc/target/functions.py
+++ b/tests/roots/test-ext-autodoc/target/functions.py
@@ -1,4 +1,4 @@
-from functools import partial
+from functools import partial, wraps
 
 
 def func():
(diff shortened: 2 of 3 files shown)


bryevdv:
@AA-Turner it does seem this change interferes with the handling of coroutines

AA-Turner:
For reference, the failing tests are decorated `async` methods:

https://github.com/sphinx-doc/sphinx/blob/31eba1a76dd485dc633cae48227b46879eda5df4/tests/roots/test-ext-autodoc/target/coroutine.py#L10-L18

A

sphinx/util/inspect.py
```diff
@@ -61,6 +61,8 @@ def unwrap_all(obj: Any, *, stop: Callable = None) -> Any:
         elif ispartial(obj):
             obj = obj.func
         elif inspect.isroutine(obj) and hasattr(obj, '__wrapped__'):
+            if hasattr(obj.__wrapped__, '__call__'):
+                return obj  # Don't unwrap wrapped callable types
```
AA-Turner:
```suggestion
            # Don't unwrap wrapped callable types:
            if hasattr(obj.__wrapped__, '__call__'):
                # From Python 3.10, staticmethod and classmethod have
                # "__wrapped__" and "__wrapped__.__call__" attributes.
                # Check that we are in Python 3.9 or earlier, or that the object
                # isn't a staticmethod or classmethod before returning.
                # xref: https://docs.python.org/3.10/whatsnew/3.10.html#other-language-changes
                if sys.version_info[:2] < (3, 10):
                    return obj
                if not (isstaticmethod(obj) or isclassmethod(obj)):
                    return obj
```

A

bryevdv: So this seems pretty unpleasant and complicated. I guess at this point I would back up and note that after all, Sphinx must be using its own `isfunction` implementation for a reason (i.e. some difference from stdlib `isfunction` must actually be desired). I'm happy to commit this suggestion if you do think making a change is best, but OTOH if the current behavior is defensible then a custom Documenter subclass to override does work fine.

bryevdv:
@AA-Turner I have committed your suggestion. I do think it would also be fine just to close this, given the unexpected complexity and given that a workaround is fairly trivial.

bryevdv:
It looks like there are still non-trivial errors, I won't actually have bandwidth to return to this, so I will go ahead and close.

PR #10443, opened 2022-05-10
Epub: Add WebP mime type

Blendify:
When building docs with WebP images using the EPUB builder results in this warning:

`WARNING: unknown mimetype for _images/<image_name>.webp, ignoring`

This commit fixes this by adding the mapping for webp-files to the correct mime type.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,24 @@
+Release 5.1.0 (in development)
+==============================
+
+Dependencies
+------------
+
+Incompatible changes
+--------------------
+
+Deprecated
+----------
+
+Features added
+--------------
+
+Bugs fixed
+----------
+
+Testing
+--------
+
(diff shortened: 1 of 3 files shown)


AA-Turner:
This would need a test case, ideally.

A

tk0miya (changes requested):
LGTM. To ship this as v5.0.0, could you rebase this into the 5.0.x branch, please? Then I'll merge this soon.

tk0miya:
I think this is a data mapping. So testing is not needed.

AA-Turner:
I was more thinking a simple regression test that WebP is accepted, rather than anything fancier.

A

tk0miya:
+0 for adding a test.

tk0miya:
It seems 5.0.x branch is already broken (it contains a commit for 5.1). So I'll fix it after merging this.

Issue #8922, opened 2021-02-23
Autosummary does not respect autodoc's :meta public:


kratsg:
**Describe the bug**

When trying to document private methods (e.g. those prefixed with `_`) and tell sphinx to treat them as public via `:meta public:` in the docstring - autodoc shows the method but autosummary does not.

Example python class below:

```python
class Foo:
  def public(self):
    """ I'm public. """
    pass

  def private(self):
    """ I'm private. """
    pass

  def private_public(self):
    """ I'm private, but publicly documented.
    :meta public:
    """
    pass
```

**To Reproduce**

```
.. autoclass:: {{ objname }}
   :show-inheritance:
   :members:
```

will include the privately documented method correctly, but

```
.. autoclass:: {{ objname }}
   :show-inheritance:

   {% block methods %}

   {% if methods %}
   .. rubric:: {{ _('Methods') }}

   {% for item in methods %}
   .. automethod:: {{ name }}.{{ item }}
   {%- endfor %}

   {% endif %}
   {% endblock %}
```

will not. Even more confusingly, `autosummary` has no functionality to actually document private methods at all (https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#customizing-templates) in a trivial way.

**Expected behavior**

I naively expected autosummary to respect `:meta XYZ:` that autodoc respects.

**Environment info**
- OS: Mac OS Mojave 10.14.6
- Python version: 3.8.6
- Sphinx version: 3.5.1
- Sphinx extensions:  [e.g. sphinx.ext.autodoc, recommonmark]
- Extra tools: [e.g. Browser, tex or something else]

**Additional context**

A current workaround seems to be to do something like the following instead:

```
   {% for item in members %}
   {% if item not in attributes and item not in inherited_members and not item.startswith('__') %}
   .. automethod:: {{ name }}.{{ item }}
   {% endif %}
   {%- endfor %}
```

but this does seem rather lengthy just to be able to get the private methods of a class (and not the special methods/members) by going through *all* members. Perhaps `autosummary` needs some extra help.

tk0miya:
Thank you for reporting.

Note: this is almost the same as #8700.

kratsg:
> Note: this is almost the same as #8700.

[...]

Issue #8700, opened 2021-01-18
Let autosummary respect `:meta private:`


xuhdev:
**Is your feature request related to a problem? Please describe.**

Currently autodoc allows inserting `:meta private:` as a way to declare a function private: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html

However, autosummary does not do so.

**Describe the solution you'd like**

In autosummary, [`functions`](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#functions) currently:

> List containing names of “public” functions in the module. Here, “public” here means that the name **does not start with an underscore**. Only available for modules.

If the bold part can be changed to **does not start with an underscore and does not have :meta: private**, the issue should be resolved.

**Describe alternatives you've considered**

Customize the autosummary template to exclude each single function that I would like to exclude. Quite tedious though.

jdillard:
I consolidated #8922 into this issue because both have the same underlying cause, which is that autosummary does not inspect docstring metadata when constructing its public template member lists.

The combined scope should cover both directions:

- `:meta private:` excludes an otherwise public member.
- `:meta public:` includes an otherwise private, underscored member. (formally covered in #8922)

A single PR should address both behaviors.

PR #10445, opened 2022-05-11
Fix #8180: Docstring metadata ignored for attributes

### LINKED ISSUE #8180: Docstring metadata ignored for attributes
**Describe the bug**
Metadata in the docstrings (or doc-comments) of attributes is ignored; for example, `:meta public:` does not make an attribute whose name starts with an underscore public.

**To Reproduce**
Apply `autodoc` to a class such as the following:
```
class Foo:
    #: A class attribute whose name starts with an underscore.
    #:
    #: :meta public:
    _bar = 12
```
There will be no entry generated for `_bar`.

**Expected behavior**
The attribute `_bar` above should be listed in the generated documentation.

**Your project**
Here is a tiny project using the example above: [example.zip](https://github.com/sphinx-doc/sphinx/files/5180051/example.zip)

**Environment info**
- OS: macOS 10.14.6
- Python version: 3.8.0
- Sphinx version: 3.3.0 (I tried my patch below on the `3.x` branch)
- Sphinx extensions:  `sphinx.ext.autodoc`

**Additional context**
The cause of the bug is that `autodoc.Documenter.filter_members` does not properly look up the docstring/doc-comments for attributes when searching for metadata. The following patch seems to fix the problem (I haven't made a PR since I'm not at all sure that this is the right way to do it):

<details>

```
diff --git a/sphinx/ext/autodoc/__init__.py b/sphinx/ext/autodoc/__init__.py
index ed02c2c90..0d9dc5f5f 100644
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -647,20 +647,24 @@ class Documenter:
             else:
                 isattr = False
 
-            doc = getdoc(member, self.get_attr, self.env.config.autodoc_inherit_docstrings,
-                         self.parent, self.object_name)
-            if not isinstance(doc, str):
-                # Ignore non-string __doc__
-                doc = None
-
-            # if the member __doc__ is the same as self's __doc__, it's just
-            # inherited and therefore not the member's doc
-            cls = self.get_attr(member, '__class__', None)
-            if cls:
-                cls_doc = self.get_attr(cls, '__doc__', None)
-                if cls_doc == doc:
+            if (namespace, membername) in attr_docs:
+                doc = '\n'.join(attr_docs[(namespace, membername)])
+                has_doc = True
+            else:
+                doc = getdoc(member, self.get_attr, self.env.config.autodoc_inherit_docstrings,
+                             self.parent, self.object_name)
+                if not isinstance(doc, str):
+                    # Ignore non-string __doc__
                     doc = None
-            has_doc = bool(doc)
+
+                # if the member __doc__ is the same as self's __doc__, it's just
+                # inherited and therefore not the member's doc
+                cls = self.get_attr(member, '__class__', None)
+                if cls:
+                    cls_doc = self.get_attr(cls, '__doc__', None)
+                    if cls_doc == doc:
+                        doc = None
+                has_doc = bool(doc)
 
             metadata = extract_metadata(doc)
             if 'private' in metadata:
```
</details>

akruis:
### Feature or Bugfix
- Bugfix

### Purpose
- Add test cases for #8180.
- Fix the bug reported in #8180. The change may also help to implement #8700 and #8922

### Detail
- The change adds two tests in test_ext_autodoc_private_members: test_private_attributes and test_private_attributes_and_private_members. Both are modeled analog to existing test cases for `:meta private:` and `:meta public:` for functions and module variables.
- About the fix: Currently the decision logic in `autodoc.Documenter.filter_members` does not see the docstring of class attributes. Therefore the fix must make the docstring available. For variables of a module ("global variables") the docstring is stored in `ObjectMember.docstring` (see `ModuleDocumenter.get_module_members()`). `autodoc.Documenter.filter_members` retrieves the docstring from the ObjectMember object. For class attributes, the ObjectMember does not contain the docstring, because `autodoc.importer.get_class_members` does not set it, when it creates the ObjectMember object. My fix changes `autodoc.importer.get_class_members` to set the docstring.
- Rejected alternatives: Issue #8180 contains a patch, that modifies `autodoc.Documenter.filter_members`. This patch also fixes the bug, but is makes the information flow more complicated. Currently `autodoc.Documenter.filter_members` uses just two ways to get the docstring of a member: as the result of `sphinx.util.inspect.getdoc` or from ObjectMember.docstring. The patch would add a third way: using `analyzer.find_attr_docs()`.

### Relates
- #8180 fixed by this pull request.
- #8700 and #8922: autosummary also uses `autodoc.importer.get_class_members`. Therefore, if autosummary shall respect metadata, it needs the docstring.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,8 @@ Features added
 Bugs fixed
 ----------
 
+* #8180: autodoc: Docstring metadata ignored for attributes
+
 Testing
 --------
 
--- a/sphinx/ext/autodoc/importer.py
+++ b/sphinx/ext/autodoc/importer.py
@@ -304,12 +304,19 @@ def get_class_members(subject: Any, objpath: List[str], attrgetter: Callable
                     members[name] = ObjectMember(name, INSTANCEATTR, class_=cls,
                                                  docstring=docstring)
 
-            # append instance attributes (cf. self.attr1) if analyzer knows
+            # append or complete instance attributes (cf. self.attr1) if analyzer knows
             if analyzer:
                 for (ns, name), docstring in analyzer.attr_docs.items():
                     if ns == qualname and name not in members:
+                        # otherwise unknown instance attribute
                         members[name] = ObjectMember(name, INSTANCEATTR, class_=cls,
                                                      docstring='\n'.join(docstring))
+                    elif (ns == qualname and docstring and
+                          isinstance(members[name], ObjectMember) and
+                          not members[name].docstring):
+                        # attribute is already known, because dir(subject) enumerates it.
+                        # But it has no docstring yet
+                        members[name].docstring = '\n'.join(docstring)
     except AttributeError:
         pass
 
--- a/tests/roots/test-ext-autodoc/target/private.py
+++ b/tests/roots/test-ext-autodoc/target/private.py
@@ -13,3 +13,15 @@ def _public_function(name):
 
 PRIVATE_CONSTANT = None  #: :meta private:
 _PUBLIC_CONSTANT = None  #: :meta public:
+
+
+class Foo:
+    #: A public class attribute whose name starts with an underscore.
+    #:
+    #: :meta public:
+    _public_attribute = 47
+
+    #: A private class attribute whose name does not start with an underscore.
+    #:
+    #: :meta private:
+    private_attribute = 11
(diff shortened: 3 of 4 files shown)


AA-Turner:
The `4.5.x` branch is closed for anything other than regressions specific to 4.5, please could you rebase to `5.x`?

A

akruis:
As requested I rebased this pull request to branch 5.x. See #10447.

PR #10452, opened 2022-05-15
Fix #10344: docs: Update explanation for copying static files manually

### LINKED ISSUE #10344: The example in the documentation in section 'Add your own static files to the build assets' is incorrect

tk0miya:
### Feature or Bugfix
- Bugfix
- Refactoring

### Purpose
- refs: #10344

### DIFF
--- a/doc/development/theming.rst
+++ b/doc/development/theming.rst
@@ -248,11 +248,11 @@ Now, you will have access to this function in jinja like so:
 Add your own static files to the build assets
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you are packaging your own build assets with an extension
-(e.g., a CSS or JavaScript file), you need to ensure that they are placed
-in the ``_static/`` folder of HTML outputs. To do so, you may copy them directly
-into a build's ``_static/`` folder at build time, generally via an event hook.
-Here is some sample code to accomplish this:
+By default, Sphinx copies static files on the ``static/`` directory of the template
+directory.  However, if your package needs to place static files outside of the
+``static/`` directory for some reasons, you need to copy them to the ``_static/``
+directory of HTML outputs manually at the build via an event hook.  Here is a sample
+code to accomplish this:
 
 .. code-block:: python
 
--- a/doc/development/theming.rst
+++ b/doc/development/theming.rst
@@ -251,8 +251,8 @@ Add your own static files to the build assets
 By default, Sphinx copies static files on the ``static/`` directory of the template
 directory.  However, if your package needs to place static files outside of the
 ``static/`` directory for some reasons, you need to copy them to the ``_static/``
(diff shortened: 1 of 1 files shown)


doc/development/theming.rst
```diff
@@ -248,11 +248,11 @@ Now, you will have access to this function in jinja like so:
...
-If you are packaging your own build assets with an extension
-(e.g., a CSS or JavaScript file), you need to ensure that they are placed
-in the ``_static/`` folder of HTML outputs. To do so, you may copy them directly
-into a build's ``_static/`` folder at build time, generally via an event hook.
-Here is some sample code to accomplish this:
+By default, Sphinx copies static files on the ``static/`` directory of the template
+directory.  However, if your package needs to place static files outside of the
+``static/`` directory for some reasons, you need to copy them to the ``_static/``
+directory of HTML outputs manually at the build via an event hook.  Here is a sample
+code to accomplish this:
```
AA-Turner:
```suggestion
directory of HTML outputs manually at the build via an event hook.  Here is an
example of code to accomplish this:
```

AA-Turner (approved):
One grammar suggestion but otherwise seems OK.

A

doc/development/theming.rst
tk0miya: Thanks!

PR #10460, opened 2022-05-20
logging: always show source locations as absolute paths

dhellmann:
Subject: logging: always show source locations as absolute paths

### Feature or Bugfix
- Bugfix

### Purpose

Nodes attached to the parse tree via the `include` directive have
their `source` attribute set to the relative path of the included
file. Other nodes in the tree use the full absolute path. For
consistent logging, ensure that the node location is always expressed
as an absolute path, when the filename is known.

### Relates

See https://github.com/sphinx-contrib/spelling/issues/153 for an
example of the effect of the original problem.

### DIFF
--- a/sphinx/util/logging.py
+++ b/sphinx/util/logging.py
@@ -12,6 +12,7 @@ from docutils.utils import get_source_line
 
 from sphinx.errors import SphinxWarning
 from sphinx.util.console import colorize
+from sphinx.util.osutil import abspath
 
 if TYPE_CHECKING:
     from sphinx.application import Sphinx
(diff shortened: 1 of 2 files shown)


AA-Turner (approved):

tk0miya (approved):
LGTM. Before merging, I'd like to wait for the opinion of @shimizukawa for a while.

shimizukawa (approved):
LGTM.
I can't think of a case where that would be a problem.

dhellmann:
I added a unit test for the new behavior.

PR #10465, opened 2022-05-22
Allow specifying multiple CSS files in themes

### LINKED ISSUE #10444: Load several css files in theme.conf
We have our own Sphinx theme called otcdocstheme. In this theme we are loading our own (specific) css file: otcdocstheme.css and some unspecific ones like bootstrap.min.css etc. The specific one is mentioned in theme.conf as it should be and the unspecific ones like bootstrap.min.css are loaded in the (layout-) .html files. This leads to the fact that the unspecific css files are loaded after the specific ones and this causes several issues due to the order, for instance overwriting elements (especially with !important flag). We also derive from the theme with some subprojects having their own .conf files which are also loaded after the theme-specific css files, as you can see in the listing below

```
<head>
...
  <link rel="stylesheet" type="text/css" href="_static/css/pygments.css">
  <link rel="stylesheet" type="text/css" href="_static/css/otcdocstheme.css">
  <link rel="stylesheet" type="text/css" href="_static/css/subtheme.css">
  ...
  <link rel="stylesheet" type="text/css" href="_static/css/bootstrap.min.css">
  <link rel="stylesheet" type="text/css" href="_static/css/fontsawesome-all.min.css">
...
</head>
```

It would like to have the option to specify more than one .css file in the theme.conf section to overcome this issue.

The alternative would be to load the unspecific css files in our specific one like bootstrap.min.css in otcdocstheme.css , but this is somehow strange in my opinion.

tk0miya:
### Feature or Bugfix
- Feature

### Purpose
- closes #10444


mgeier:
Thanks @tk0miya, this works great!

Since [html_style](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=html_style#confval-html_style) can be used to overwrite this, wouldn't it make sense to extend `html_style` as well to accept a list of themes (in addition to a single string)?

sphinx/builders/html/__init__.py
```diff
@@ -272,6 +272,15 @@ def _get_style_filename(self) -> str:
...
             return 'default.css'
 
+    def get_style_filenames(self) -> str:
+        if self.config.html_style is not None:
+            return [self.config.html_style]
+        elif self.theme:
+            stylesheet = self.theme.get_config('theme', 'stylesheet')
+            return [s.strip() for s in stylesheet.split(',')]
+        else:
+            return ['default.css']
```
AA-Turner:
```suggestion
    def _get_style_filenames(self) -> Iterator[str]:
        if isinstance(self.config.html_style, str):
            yield self.config.html_style
        elif self.config.html_style is not None:
            yield from self.config.html_style
        elif self.theme:
            stylesheet = self.theme.get_config('theme', 'stylesheet')
            yield from (s.strip() for s in stylesheet.split(', '))
        else:
            yield 'default.css'
```

sphinx/builders/html/__init__.py
```diff
@@ -309,7 +318,9 @@ def init_highlighter(self) -> None:
     def init_css_files(self) -> None:
         self.css_files = []
         self.add_css_file('pygments.css', priority=200)
-        self.add_css_file(self._get_style_filename(), priority=200)
+
+        for filename in self.get_style_filenames():
```
AA-Turner:
```suggestion
        for filename in self._get_style_filenames():
```

CHANGES
```diff
@@ -13,6 +13,9 @@ Deprecated
 Features added
 --------------
 
+* #10444: html theme: Allow to specify multiple CSS files via ``stylesheet``
+  setting on ``theme.conf``
```
```suggestion
* #10444: html theme: Allow specifying multiple CSS files through the ``stylesheet``
  setting in ``theme.conf`` or by setting ``html_style`` to an iterable of strings.
```

AA-Turner (changes requested):
We should also remove the private `_get_style_filename` method if it is no longer used.

A

PR #10468, opened 2022-05-22
Drop python 3.6 support

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Based on our "Python version support policy", python3.6 support will be ended at 2023.04.
- This drops the support from the master branch used for development for v6.0.
- I'll also drop python 3.7 support on the next PR.
- This only drops the code for py36. It would be better to refactor with new syntax, libraries and so on after clean up.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,8 +10,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - python: "3.6"
-          docutils: du14
         - python: "3.7"
           docutils: du15
         - python: "3.8"
--- a/.github/workflows/builddoc.yml
+++ b/.github/workflows/builddoc.yml
@@ -11,7 +11,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.8
     - name: Install dependencies
       run:  |
         sudo apt update
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.8
     - name: Install dependencies
       run:  pip install -U tox
     - name: Run Tox
--- a/CHANGES
+++ b/CHANGES
@@ -4,6 +4,8 @@ Release 6.0.0 (in development)
 Dependencies
 ------------
 
+* Drop python 3.6 support
+
 Incompatible changes
 --------------------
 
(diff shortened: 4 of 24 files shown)


.github/workflows/main.yml
```diff
@@ -10,8 +10,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        - python: "3.6"
-          docutils: du14
```
tk0miya, line 14: Note: Ubuntu-20.04 distributes docutils-0.16. So it will be the new supported base version. I'll work on it after merging this.
https://packages.ubuntu.com/focal/python3-docutils

AA-Turner: Would it be possible to increase the minimum requirement to Docutils 0.18? There are lots of patches in Sphinx for Docutils 0.17 or earlier, and it would be nice to remove those at the same time as we drop end-of-life Python versions.

As far as I can tell the current Sphinx version support policy only makes reference to Ubuntu for Python versions, not for Docutils.

A

AA-Turner (approved):

.github/workflows/main.yml
tk0miya: In my thought, it's simple to refer to Ubuntu LTS to decide the version of dependencies. I feel dropping dependency does not help our work. So -0 for your proposal

tk0miya:
Does this conflict with https://github.com/sphinx-doc/sphinx/pull/10470? May I merge this?

AA-Turner:
> Does this conflict with #10470? May I merge this?

It probably does, but I'd say merge this first and then I'll deal with the conflicts and rebase #10470.

A

.github/workflows/main.yml
AA-Turner: > I feel dropping dependency does not help our work.

We could take a view to support every version of Python / Docutils perpeptually, and just accumulate a lot of patches. I personally find it harder to reason about the code when there are multiple layers of patches etc, especially ones that monkeypatch Docutils, hence looking to bump the version higher and remove code.

I'd argue we should look to make things simpler for ourselves as maintainers where possible - Ubuntu is a re-distributor and will cope with decisions that we make under their policies, but I don't think we should be somewhat arbitrarily bound by them when there are significant code simplifications from the bump to 0.18.

A

danieleades:
you might want to consider running `pyupgrade` over these deprecation PRs

https://pypi.org/project/pyupgrade/

AA-Turner:
I ran `pyupgrade` locally on my larger refactoring branch (https://github.com/AA-Turner/sphinx/commits/drop-py37-refactoring) -- this should probably stay more focussed.

A

AA-Turner:
I will rebase #10470 on master.

A

PR #10470, opened 2022-05-23
Drop Python 3.7 and Docutils 0.17

AA-Turner:
Follows on from #10468, I polished up an experiment from April.

I dropped Docutils 0.16 and 0.17 in different commits. I would advocate for a higher minimum to reduce the matrix of potential Sphinx versions x Docutils versions, and by the time Sphinx 6 is released (April 2023) we may have Docutils 0.19 and 0.20.

I also have a few refactoring steps in a branch (https://github.com/AA-Turner/sphinx/tree/drop-py37-refactoring), although I'll propose those as PRs after we've dropped Python 3.6 / 3.7.

### Feature or Bugfix
- Refactoring

A


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 2.4.0
-envlist = docs,flake8,mypy,twine,py{36,37,38,39,310},du{14,15,16,17,18}
+envlist = docs,flake8,mypy,twine,coverage,py{36,37,38,39,310},du{17,18,19}
```
hugovk:
```suggestion
envlist = docs,flake8,mypy,twine,coverage,py{37,38,39,310,311},du{17,18,19}
```

And below on line 18.

sphinx/util/inspect.py
```diff
@@ -149,7 +144,7 @@ def getslots(obj: Any) -> Optional[Dict]:
 
 def isNewType(obj: Any) -> bool:
     """Check the if object is a kind of NewType."""
-    if sys.version_info >= (3, 10):
+    if sys.version_info[:2] >= (3, 10):
```
hugovk, line 147: I think all these `[:2]` additions are unnecessary for `>`, `>=` and `<`?

Or is it for consistency so we don't need to worry about `<=`?

AA-Turner: I always forget which operator is the special case so personally adding `[:2]` gives me more reassurance, you're likely right that it isn't strictly necessary.

A

tox.ini
AA-Turner: Adding 311 envs should be done on the 5.x branch, I've dropped the outdated envs on this PR.

A

hugovk: Thanks, please see https://github.com/sphinx-doc/sphinx/pull/10871 to add `311` to 5.x.

PR #10471, opened 2022-05-23
Remove deprecated items for Sphinx 6.0

AA-Turner:
Removes items marked with `RemovedInSphinx60Warning`, add `RemovedInSphinx80Warning`.

### Feature or Bugfix
- Refactoring

A


sphinx/domains/c.py
```diff
@@ -12,7 +12,6 @@
 from sphinx.addnodes import pending_xref
 from sphinx.application import Sphinx
 from sphinx.builders import Builder
-from sphinx.deprecation import RemovedInSphinx60Warning
```
jakobandersen, line 15: For now please bump this to ``RemovedInSphinx70Warning``. We need to confirm with the kernel folks which minimum Sphinx version they use.

sphinx/domains/c.py
```diff
@@ -3253,27 +3252,8 @@ def handle_signature(self, sig: str, signode: TextElement) -> ASTDeclaration:
...
-                self.object_type = ast.objectType  # type: ignore
-                if self.env.config['c_warn_on_allowed_pre_v3']:
-                    msg = "{}: Pre-v3 C type directive '.. c:type:: {}' converted to " \
-                          "'.. c:{}:: {}'." \
-                          "\nThe original parsing error was:\n{}"
-                    msg = msg.format(RemovedInSphinx60Warning.__name__,
-                                     sig, ast.objectType, ast, eOrig)
-                    logger.warning(msg, location=signode)
+            ast = self.parse_definition(parser)
+            parser.assert_end()
```
jakobandersen: See above, revert.

sphinx/domains/c.py
```diff
@@ -3675,39 +3655,6 @@ def process_link(self, env: BuildEnvironment, refnode: Element,
...
-            signode = nodes.inline(classes=classes)
-            ast.describe_signature(signode, 'markType', self.env, parentSymbol)
-
-            if self.env.config['c_warn_on_allowed_pre_v3']:
-                msg = "{}: Pre-v3 C type role ':c:type:`{}`' converted to ':c:expr:`{}`'."
-                msg += "\nThe original parsing error was:\n{}"
-                msg = msg.format(RemovedInSphinx60Warning.__name__, text, text, eOrig)
-                logger.warning(msg, location=self.get_location())
-            return [signode], []
-
```
jakobandersen: See above, revert.

sphinx/domains/c.py
```diff
@@ -3917,9 +3864,6 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value("c_extra_keywords", _macroKeywords, 'env')
     app.add_post_transform(AliasTransform)
 
-    app.add_config_value("c_allow_pre_v3", False, 'env')
-    app.add_config_value("c_warn_on_allowed_pre_v3", True, 'env')
-
```
jakobandersen: See above, revert.

jakobandersen (changes requested):

sphinx/domains/c.py
AA-Turner: Are we only keeping this code for the Linux kernel? If so, perhaps a specific opt-in environment variable rather than keeping on moving the depreciation?

Who would be best to talk to on the kernel side to check if they've updated their syntax?

A

AA-Turner: @jnikula You mentioned [here](https://github.com/sphinx-doc/sphinx/issues/7008#issuecomment-817309815) that "I did the bulk of the work to convert the Linux kernel to use Sphinx." -- would you be able to comment from the kernel side on what your needs/wants are, or suggest someone else we could ask please?

A

jnikula: @AA-Turner Thanks for the ping. I was heavily involved in the early days of converting the kernel to use Sphinx, but I haven't closely followed the recent developments. Basically I think I'd also be inclined to push for *much* higher minimum Sphinx version requirements than what the kernel currently has. The minimum at the moment is v1.7.9 (or v2.4.4 for PDF). It's difficult to maintain support for a wide range of Sphinx versions.

Perhaps the best bet would be to mail the kernel documentation list at linux-doc@vger.kernel.org and Cc Jonathan Corbet <corbet@lwn.net> to try to reach an understanding on the recommended minimum version and version ranges that makes sense for both parties to support. HTH.

AA-Turner: Posted at https://lore.kernel.org/linux-doc/LO3P123MB26810D190462B6BBBF1305F6C2A19@LO3P123MB2681.GBRP123.PROD.OUTLOOK.COM/T/#u

A

AA-Turner: The outcome of that posting was that the kernel doesn't require it. Python still does though, so I'll remove it from this PR.

A

PR #10478, opened 2022-05-27
Only create a target for object descriptions

### LINKED ISSUE #9662: Allow creating custom targets in a domain

ghost:
Subject: Only create a target for object descriptions

### Feature or Bugfix
- Feature

### Purpose
Adding feature for https://github.com/sphinx-doc/sphinx/issues/9662 (ability to only create a target for a domain description).
Thanks again to @jakobandersen for the inspiration.

This is #9675 repurposed and rebased onto the current 5.x. In the hope that it now gets more traction.

Open Tasks:
* Bikeshedding:  should the parameter be named `:hidden:`,  `:targetonly:`, `:your-favorite-word-here:`, ...
* If this implementation gets an approval, I'll add the documentation for the parameter before merging.

### Relates
* Closes https://github.com/sphinx-doc/sphinx/issues/9662
* https://github.com/sphinx-doc/sphinx/pull/9671
* https://github.com/sphinx-doc/sphinx/pull/9675


sphinx/directives/__init__.py
```diff
@@ -203,8 +205,30 @@ def run(self) -> List[Node]:
...
         self.after_content()
+
+        if node['hidden']:
+            node = self.replace_node_with_target(node)
+            self.set_source_info(node)
+
         return [self.indexnode, node]
 
+    # TODO: This method does not need access to self, make a (non-class) function out of it?
+    def replace_node_with_target(self, node: nodes.Node) -> nodes.target:
```
AA-Turner:
```suggestion
    @staticmethod
    def replace_node_with_target(node: nodes.Node) -> nodes.target:
```

sphinx/directives/__init__.py
```diff
@@ -203,8 +205,30 @@ def run(self) -> List[Node]:
...
+
+        def collect_ids(node: nodes.Node) -> List[str]:
+            if isinstance(node, nodes.Element):
+                ids: List[str] = node.get('ids', [])
+                for c in node.children:
+                    ids.extend(collect_ids(c))
+                return ids
+            else:
+                return []
```
AA-Turner: Does this function need to be defined as a closure? If not please factor out for speed.

sphinx/directives/__init__.py
```diff
@@ -203,8 +205,30 @@ def run(self) -> List[Node]:
...
+
+        ids: List[str] = collect_ids(node)
+        target_node = nodes.target()
+        target_node['ids'] = ids
+        return target_node
```
AA-Turner: Does this work?
```suggestion
        target_node = nodes.target(ids=collect_ids(node))
        return target_node
```

sphinx/domains/javascript.py
```diff
@@ -37,10 +37,10 @@ class JSObject(ObjectDescription[Tuple[str, str]]):
     #: based on directive nesting
     allow_nesting = False
 
-    option_spec: OptionSpec = {
-        'noindex': directives.flag,
+    option_spec: OptionSpec = ObjectDescription.option_spec.copy()
+    option_spec.update({
```
AA-Turner: Please just copy `:hiden:` to all subclasses, it makes it easier to reason about.

AA-Turner (changes requested):
This needs a lot of tests -- for the new transform and for the behaviour x each domain.

A

AA-Turner:
You also need to fix the existing test failures.

A

sphinx/domains/javascript.py
ghost: > Please just copy :hiden: to all subclasses, it makes it easier to reason about.

Here I followed the examples given py `sphinx/domains/python.py` et al as well as the
docutils classes, which (except for the Tables) copy the `option_spec` of their
super class end enhance it.  This IMHO makes it sense since – like for example
in this PR – we can add an option/functionality to all object description
directives simply by adding it to the super class `ObjectDescription` without
modifying the (potentially many) subclasses.  Isn't that what inheritance is
about?

If you still think it makes sense to verbatim copy the `option_spec` values,
please say so and I'll change it.  In that case: Should I also update all the
current `option_spec.copy()` not touched by this PR, i.e. in
`sphinx/domains/rst.py` and `sphinx/domains/python.py`?

latosha-maltba:
So, I added documentation and unit tests.

I'm still not quiet happy with the naming of the option: `:hidden:` doesn't seem to grasp it, `:notypesetting:` seems more precise but is IMHO clumsy.  Better `:nocontent:` however, the object description is build form signature+content, so this is misleading and there is a term missing for the  union of both signature and content.  Maybe `:astarget:`? (`:onlytarget:` omits that we are also creating an index entry not only a target.) For the consistency with `:noindex:` and `:noindexentry` I tend to lean to a `:noXXX:` word but that is not required.  Naming is hard, so thoughts are welcome.

sphinx/domains/cpp.py
```diff
@@ -7083,10 +7083,12 @@ class CPPObject(ObjectDescription[ASTDeclaration]):
                      can_collapse=True),
     ]
 
-    option_spec: OptionSpec = {
+    option_spec: OptionSpec = ObjectDescription.option_spec.copy()
```
jakobandersen: Can you change the ``CPPAlias`` class similarly to the ``CAlias`` class? I.e., copy this ``option_spec`` and delete  ``notypesetting``.

doc/usage/restructuredtext/domains.rst
```diff
@@ -50,13 +50,18 @@ give the directive option flag ``:noindexentry:``.
 If you want to typeset an object description, without even making it available
 for cross-referencing, you can give the directive option flag ``:noindex:``
 (which implies ``:noindexentry:``).
+If you do not want to typeset anything but create a target and index entry for
+later reference, you can give the directive option flag ``:notypesetting:``.
```
jakobandersen: This may be slightly ambiguous: does ``notypesetting`` imply not ``noindex``/``noindex``? I would assume that it is independent. I.e., if the user specifies ``notypesetting`` and ``noindexentry``/``noindex``, then there will be almost no side-effect of having the directive.
```suggestion
If you do not want to typeset anything you can give the directive option flag ``:notypesetting:``.
This can for example be used to create only a target and index entry for later reference.
```

jakobandersen (changes requested):
Looks good to me. I like ``notypesetting`` as it conveys exactly what it is doing, and it is in line with ``noindex`` and ``noindexentry``.

sphinx/domains/cpp.py
ghost: > Can you change the `CPPAlias` class similarly to the `CAlias` class? I.e., copy this `option_spec` and delete `:notypesetting:`.

Sure! ... However, @AA-Turner requested the opposite in https://github.com/sphinx-doc/sphinx/pull/10478#discussion_r883970880 but that discussion is not yet resolved.  Let's wait for their response.

doc/usage/restructuredtext/domains.rst
ghost: > This may be slightly ambiguous: does notypesetting imply not noindex/noindex? I would assume that it is independent. I.e., if the user specifies notypesetting and noindexentry/noindex, then there will be almost no side-effect of having the directive.

You are correct in your assumption: Currently, `:notypesetting:` and `:noindex:`/`:noindexentry:` are completely independent, neither implies the other and they can freely mixed.
The combination `:notypesetting:` and `:noindex:` results in essentially an empty `<index>` node which results in a noop for the usual outputs. However, some domain out there might provide ids in the nodes it produces even if `:noindex:` is provided. In this case we generate a `<target>` node with those ids so they'll become linkable.
The combination `:notypesetting:` and `:noindexentry:` usually results in an linkable `<target>` node, i.e. can be referenced later but as suggested no entry on the generated index page.

I'll use your wording suggestions once I know what to do with the `option_spec`.

sphinx/domains/javascript.py
AA-Turner: Makes sense, the current method (`.update`) is fine.

A

sphinx/domains/cpp.py
ghost: done

latosha-maltba:
* Rebased this on newest 5.x so I can add a `CHANGES` entry without conflicts.
* Addressed the two issues raised by @jakobandersen 

Thanks for all the feedback!

latosha-maltba:
Is there anything that prevents this from being merged or what else can I do to facilitate the adoption of this?

jakobandersen (approved):

latosha-maltba:
@AA-Turner, @tk0miya anything preventing this from being merged?

AA-Turner (approved):

PR #10482, opened 2022-05-27
Update `requests` links

AA-Turner:
Fixes CI due to the `python-requests` domain no longer working.

@tk0miya please can this go into 5.0.0 to fix CI / for correct docs?

### Feature or Bugfix
- Bugfix

A

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -109,7 +109,6 @@ texinfo_documents = [
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3/', None),
-    'requests': ('https://docs.python-requests.org/en/latest/', None),
     'readthedocs': ('https://docs.readthedocs.io/en/stable', None),
 }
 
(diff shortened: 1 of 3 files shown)


doc/conf.py
```diff
@@ -109,7 +109,6 @@
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3/', None),
-    'requests': ('https://docs.python-requests.org/en/latest/', None),
```
tk0miya, line 112: Why removing this?

AA-Turner: As I made the authentication link inline, and nothing else used interphinx requests.

A

tk0miya: I feel it's better to update this URL by new one (refs: #10484). What do you think?

AA-Turner: I don't mind either way, just important for CI to be fixed.

A

tk0miya: Thanks, I'd like to choose #10484.

PR #10493, opened 2022-05-30
Add `aside.topic` for Docutils 0.18+

AA-Turner:
Found when rebuilding the Python documentation with Sphinx 5.

I'm not sure if this is worth a `5.0.1` release or not.

### Feature or Bugfix
- Bugfix

A

### DIFF
--- a/sphinx/themes/agogo/static/agogo.css_t
+++ b/sphinx/themes/agogo/static/agogo.css_t
@@ -271,7 +271,8 @@ div.document ol {
 
 /* Sidebar */
 
-div.sidebar {
+div.sidebar,
+aside.sidebar {
   width: {{ theme_sidebarwidth|todim }};
   {%- if theme_rightsidebar|tobool %}
   float: right;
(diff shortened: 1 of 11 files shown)


tk0miya (approved):
Thanks. It must be an important bug. So I'm merging this into the 5.0.x branch.

mgeier:
This fix is still missing for Sphinx's own theme `sphinx13`.

https://github.com/sphinx-doc/sphinx/blob/af425c069f6f51ba8fb3c348fd53479f0434a779/doc/_themes/sphinx13/static/sphinx13.css#L375

I think this is a good example for how it is easy to miss fixes in the `sphinx13` theme. That's exactly why I suggested to move it to the rest of the built-in themes: #7733
Maybe this is worth reconsidering?

AA-Turner:
@mgeier -- mea culpa, I searched within the `sphinx/themes` directory. #10521 is a fix for 5.0.2.

A

mgeier:
I wouldn't say it's your fault, because the `sphinx13` theme is very easy to miss, that's why I suggested #7733, which would make such things harder to miss in the future.

PR #10502, opened 2022-05-31
`ifconfig`: Add a meta node to fix iteration

### LINKED ISSUE #10496: unknown node type ifconfig in sphinx 5.0.0
### Describe the bug

My project's CI just switched to sphinx 5.0.0 and I'm unable to generate my documentation that depends heavily on `ifconfig`. In this particular case there seems to be an issue with doing an `.. include::` of a file that uses `ifconfig`.

```
source/overview.rst:46: WARNING: unknown node type: <ifconfig: <comment...><paragraph...><paragraph...><paragraph...>>

Exception occurred:
  File "/home/davidh/miniconda3/envs/polar2grid_py310/lib/python3.10/site-packages/docutils/nodes.py", line 2068, in unknown_departure
    raise NotImplementedError(
NotImplementedError: <class 'sphinx.writers.html5.HTML5Translator'> departing unknown node type: ifconfig
```

### How to Reproduce

```
$ git clone https://github.com/ssec/polar2grid
$ conda env create -n polar2grid -f continuous_integration/environment.yaml
$ conda activate polar2grid
$ cd doc
$ make html
```

### Expected behavior

No error.

### Your project

https://github.com/ssec/polar2grid/

### Screenshots

_No response_

### OS

Ubuntu

### Python version

3.10

### Sphinx version

5.0.0

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

AA-Turner:
Fixes #10496

This issue arose due to the change from `.traverse` to `.findall`, which moved from a precompiled list to a generator.

### Feature or Bugfix

- Bugfix

A

### DIFF
--- a/sphinx/ext/ifconfig.py
+++ b/sphinx/ext/ifconfig.py
@@ -20,6 +20,7 @@ from docutils import nodes
 from docutils.nodes import Node
 
 import sphinx
+from sphinx import addnodes
 from sphinx.application import Sphinx
 from sphinx.util.docutils import SphinxDirective
 from sphinx.util.nodes import nested_parse_with_titles
@@ -64,7 +65,7 @@ def process_ifconfig_nodes(app: Sphinx, doctree: nodes.document, docname: str) -
             node.replace_self(newnode)
         else:
             if not res:
-                node.replace_self([])
+                node.replace_self(addnodes.meta())
             else:
                 node.replace_self(node.children)
 
--- a/sphinx/ext/ifconfig.py
+++ b/sphinx/ext/ifconfig.py
@@ -20,7 +20,6 @@ from docutils import nodes
 from docutils.nodes import Node
 
 import sphinx
-from sphinx import addnodes
 from sphinx.application import Sphinx
 from sphinx.util.docutils import SphinxDirective
 from sphinx.util.nodes import nested_parse_with_titles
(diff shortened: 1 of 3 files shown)


AA-Turner:
cc: @djhoese can you test this please and see if it works?

AA-Turner:
3.11 failing due to https://github.com/pytest-dev/pytest/issues/10008

A

djhoese:
It works! Thanks. I tested it on both my reproducer in the related issue and on my actual project and both build and look correct.

sphinx/ext/ifconfig.py
```diff
@@ -64,7 +65,7 @@ def process_ifconfig_nodes(app: Sphinx, doctree: nodes.document, docname: str) -
             node.replace_self(newnode)
         else:
             if not res:
-                node.replace_self([])
+                node.replace_self(addnodes.meta())
```
tk0miya: Why do you insert a meta node? If my understanding is correct, it's better to convert the result of `findall()` to the list on the top of the loop, instead?

AA-Turner: Converting to a list is the option of last resort as it is slower (double iteration) and takes more memory -- a meta node here has no effect on the output and in effect is a noop node. I could convert to a list if you'd prefer, but this is the reason I didn't.

A

tk0miya: I think doctrees should follow the semantics. So -1 for using meaningful node for the noop purpose. If you'd like to insert a noop node, how about using `Text('')` instead? It's meaningless and noop node. I can accept it.

AA-Turner:
@tk0miya I didn't have time to update this one, please feel free to push to this branch wrapping the findall code in a list, I think that is better than inserting a blank text node.

A

tk0miya:
Now I pushed "wrapping" code. Merging now.

PR #10517, opened 2022-06-03
Support for Glossary Identifiers

### LINKED ISSUE #1399: Better support for multiple glossaries
You can use the glossary directive multiple times in a documentation without problems when you never define the same word twice. 

This makes it possible to have a glossary with subtitles for topics eg.

I wonder if anyone ever had the need to put more than one glossary into the documentation. Maybe domain specific glossaries? One for math terms and one for chemical terms. 

If this is a realistic usecase it would be nice to label the glossaries so that I can specify where a term belongs to. Maybe some kind of indicator as we use with the domains.

Example:

```
#!rst

.. glossary:: math

   term
      (definition)

This is a :math:term:`term`for my math specific glossary
```

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1399
- Originally reported by: Jan Ulrich Hasecke
- Originally created at: 2014-02-19T13:08:39.295

Arsha-M:
### Feature or Bugfix
- Feature

### Purpose
- Currently, Sphinx supports multiple glossaries, but the same term cannot appear twice. This is improved by adding an identifier to glossaries so that the correct glossary can be referenced.

### Detail
- Added the identifier to the glossary directive
- Added documentation
- Added testing

### Relates
-  Closes #1399

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -830,6 +830,25 @@ Glossary
    The whole characters in "key" is used instead of a first character; it is
    used for "Combining Character Sequence" and "Surrogate Pairs" grouping key.
 
+   If you want to specify a "glossary key" to make glossaries unique even if
+   they have terms with the same names, you can put a "key" as
+   ".. glossary:: key". For example::
+
+      .. glossary::
+
+         term 1
+            Definition
+      
+      .. glossary:: second
+
+         term 1
+            Different Definition
+
+   This "glossary key" makes is so that you can properly reference glossaries as::
+      
+      The syntax to reference the first glossary is :term:`term 1` and to reference
+      the second glossary the syntax is :term:`second:term 1`.
+
    In i18n situation, you can specify "localized term : key" even if original
    text only have "term" part. In this case, translated "localized term" will be
    categorized in "key" group.
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -264,8 +264,9 @@ def split_term_classifiers(line: str) -> List[Optional[str]]:
     return parts
 
 
-def make_glossary_term(env: "BuildEnvironment", textnodes: Iterable[Node], key: str, index_key: str,
-                       source: str, lineno: int, node_id: str, document: nodes.document
+def make_glossary_term(env: "BuildEnvironment", textnodes: Iterable[Node], key: str,
+                       index_key: str, source: str, lineno: int, node_id: str,
+                       document: nodes.document
                        ) -> nodes.term:
     # get a text-only representation of the term and register it
     # as a cross-reference target with the proper key.
@@ -313,7 +314,7 @@ class Glossary(SphinxDirective):
             key = self.arguments[0].strip() + ":"
         else:
             key = ''
-        
+
         node = addnodes.glossary()
         node.document = self.state.document
         node['sorted'] = ('sorted' in self.options)
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -264,30 +264,30 @@ def split_term_classifiers(line: str) -> List[Optional[str]]:
     return parts
 
 
-def make_glossary_term(env: "BuildEnvironment", textnodes: Iterable[Node], index_key: str,
+def make_glossary_term(env: "BuildEnvironment", textnodes: Iterable[Node], key: str, index_key: str,
                        source: str, lineno: int, node_id: str, document: nodes.document
                        ) -> nodes.term:
     # get a text-only representation of the term and register it
-    # as a cross-reference target
+    # as a cross-reference target with the proper key.
     term = nodes.term('', '', *textnodes)
     term.source = source
     term.line = lineno
-    termtext = term.astext()
+    term_reference = key + term.astext()
 
     if node_id:
         # node_id is given from outside (mainly i18n module), use it forcedly
         term['ids'].append(node_id)
     else:
-        node_id = make_id(env, document, 'term', termtext)
+        node_id = make_id(env, document, 'term', term_reference)
         term['ids'].append(node_id)
         document.note_explicit_target(term)
 
     std = cast(StandardDomain, env.get_domain('std'))
-    std._note_term(termtext, node_id, location=term)
+    std._note_term(term_reference, node_id, location=term)
 
     # add an index entry too
     indexnode = addnodes.index()
-    indexnode['entries'] = [('single', termtext, node_id, 'main', index_key)]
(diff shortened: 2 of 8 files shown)


Arsha72128:
Hello! I believe that I am ready for the workflows to be run and for a review. :)

I successfully ran the linter, unit tests, and documentation locally. However, I am sure that I made a mistake somewhere as this is my first contribution.

I don't understand how the `i18n.py` module works so in the `make_glossary_term()` call, I passed an empty string as the key. Is this correct?

Also, are you happy with the referencing of the terms? Here is an example of what it looks like:

image

image

The source to produce this is:

image

image

sphinx/transforms/i18n.py
```diff
@@ -220,7 +220,7 @@ def apply(self, **kwargs: Any) -> None:
                     parts = split_term_classifiers(msgstr)
                     patch = publish_msgstr(self.app, parts[0], source,
                                            node.line, self.config, settings)
-                    patch = make_glossary_term(self.env, patch, parts[1],
+                    patch = make_glossary_term(self.env, patch, '', parts[1],
```
Arsha-M, line 223: Is this correct?

Arsha72128:
Hello @AA-Turner, gentle ping. Would I be able to get some feedback on these changes? 🙏

Thanks!

AA-Turner:
I can review the change, but I want to review the discussion on the overall design first.

A

doc/usage/restructuredtext/directives.rst
```diff
@@ -830,6 +830,25 @@ Glossary
    The whole characters in "key" is used instead of a first character; it is
    used for "Combining Character Sequence" and "Surrogate Pairs" grouping key.
 
+   If you want to specify a "glossary key" to make glossaries unique even if
```
tk0miya, line 833: I feel "key" is a bit functional wording. It would be better to use more semantic words (ex. classifier, category, glossary type, and so on).

In my understanding, this PR improves Sphinx to be able to have multiple "kinds" of glossaries. So this document should describe the design as possible. It's an advantage of multiple glossaries. But it's not a main goal.

sphinx/domains/std.py
```diff
@@ -264,30 +264,31 @@ def split_term_classifiers(line: str) -> List[Optional[str]]:
...
         term['ids'].append(node_id)
     else:
-        node_id = make_id(env, document, 'term', termtext)
+        node_id = make_id(env, document, 'term', term_reference)
         term['ids'].append(node_id)
         document.note_explicit_target(term)
 
     std = cast(StandardDomain, env.get_domain('std'))
-    std._note_term(termtext, node_id, location=term)
+    std._note_term(term_reference, node_id, location=term)
```
tk0miya, line 287: It's better to separate the glossary key and term in the domain database. It might cause conflict with a term having colon and a term having a key. No reason to store them in concatenated form.

```
.. glossary::

   key:term

.. glossary:: key

   term
```

sphinx/domains/std.py
```diff
@@ -264,30 +264,31 @@ def split_term_classifiers(line: str) -> List[Optional[str]]:
...
 
     # add an index entry too
     indexnode = addnodes.index()
-    indexnode['entries'] = [('single', termtext, node_id, 'main', index_key)]
+    indexnode['entries'] = [('single', term_reference, node_id, 'main', index_key)]
```
tk0miya, line 291: Do you check how a term having a key is shown on the index? Is it better to prepend the key here?

sphinx/domains/std.py
```diff
@@ -302,13 +303,18 @@ class Glossary(SphinxDirective):
 
     has_content = True
     required_arguments = 0
-    optional_arguments = 0
+    optional_arguments = 1
     final_argument_whitespace = False
```
tk0miya, line 307: Enabling this allows users to use phrases to glossary key.

```
.. glossary:: document writing
```

tk0miya:
I think a title of reference and an index entry should be bare "Term1".

shimizukawa:
Here is a summary of my opinion.

1. Sometime I also want to have a type for the term and allow a term with the same name for each type.
2. In articles, display terms of the same name as they are.
3. In the Index, display terms of the same name with separated entries so that they are indistinguishable from each other.

In the index, for example,
separated entries as:
```
term (type1)
term (type2)
```
or pair style as:
```
term
    type1
    type2
```

doc/usage/restructuredtext/directives.rst
Arsha-M: Thanks for the review! 🙏

I will change the term "key" to "classifier", and improve the document.

sphinx/domains/std.py
Arsha-M: @tk0miya What do you suggest I do here?

Should I allow this behavior?

Or, check for spaces, and if there are spaces, throw an error?

sphinx/domains/std.py
Arsha-M: I don't exactly understand what you mean, could you please clarify? I am confused by the code block that you added, and why you don't want them to be stored in concatenated form.

What I am understanding is you would like me to do the following:

```python
std._note_term(termtext, node_id, location=term)
```

where

```python
termtext = "Term1"
```

Arsha72128:
Thanks for the feedback @tk0miya @shimizukawa 

I would like to summarize the requested changes before I make them:

1. Instead of calling it a glossary "key" change it to "classifier"
2. Make it so that the index shows the terms with the different types in a nicer format. [as suggested](https://github.com/sphinx-doc/sphinx/pull/10517#issuecomment-1173039568) by @shimizukawa 
3. Fix the behavior of the `optional_arguments` in a manner that @tk0miya suggests.
4. Improve the way that the "classifier" and term are stored in the domain database.

Thanks again for your feedback, I greatly appreciate it.

n-peugnet:
@shimizukawa:
> 3. In the Index, display terms of the same name with separated entries so that they are indistinguishable from each other.
> 
> 
> In the index, for example, separated entries as:
> 
> ```
> term (type1)
> term (type2)
> ```
> 
> or pair style as:
> 
> ```
> term
>     type1
>     type2
> ```

IMO the pair style would be perfect. This would also generate these entries:

```
type1
    term
type2
    term
```

But I would like to be able to translate `type1` and `type2` in other languages.

neongreen-sc:
> Whenever you say "title of reference" do you mean when you do :term:`page2:Term2` , you would like it to be rendered as "Term2" instead of "page2:Term2"?

I'd also like to voice support for "Term2" by default instead of "page2:Term2".

My usecase is building documentation for several teams working in the same monorepo. Some terms are used differently by different teams. If terms were rendered with category prefixes by default, it means we'd _always_ have to override the reference text.

PR #10518, opened 2022-06-03
Add `include_patterns` as the opposite of `exclude_patterns`

AA-Turner:
Currently Sphinx walks the source directory recursivley, looking for files. This is useful when documentation sources are in a distinct `Documentation` or etc folder, but for projects like the PEPs where the project *is* the documentation, it is more useful to specify which files we want to include.

This adds an implementation which by default works exactly the same as today, but allows configuring `include_paths`.

### Feature or Bugfix
- Feature

A


sphinx/util/matching.py
```diff
@@ -63,9 +64,10 @@ class Matcher:
           For example, "**/index.rst" matches with "index.rst"
     """
 
-    def __init__(self, patterns: List[str]) -> None:
-        expanded = [pat[3:] for pat in patterns if pat.startswith('**/')]
-        self.patterns = compile_matchers(patterns + expanded)
+    def __init__(self, exclude_patterns: Iterable[str],
+                 include_patterns: Iterable[str] = ()) -> None:
```
tk0miya, line 68: It seems the Matcher class does not use the `include_patterns` argument. Why did you add this here?

sphinx/util/matching.py
```diff
@@ -99,3 +101,63 @@ def patfilter(names: Iterable[str], pat: str) -> List[str]:
         _pat_cache[pat] = re.compile(_translate_pattern(pat))
     match = _pat_cache[pat].match
     return list(filter(match, names))
+
+
+def get_matching_files(
+    dirname: str,
+    exclude_patterns: Iterable[str] = (),
+    include_patterns: Iterable[str] = ("**",)
```
tk0miya, line 109: Is it better to put this argument before `exclude_patterns`? It's a chance for refactoring.

sphinx/directives/other.py
```diff
@@ -84,7 +84,7 @@ def parse_content(self, toctree: addnodes.toctree) -> List[Node]:
         all_docnames.remove(self.env.docname)  # remove current document
 
         ret: List[Node] = []
-        excluded = Matcher(self.config.exclude_patterns)
+        excluded = Matcher(self.config.exclude_patterns, self.config.include_patterns)
```
tk0miya, line 87: This matcher is used to emit "toctree contains reference to excluded document" warning. I think it's a bit better to show another warnings like "toctree contains reference to not included document", right?

doc/usage/configuration.rst
```diff
@@ -219,6 +218,26 @@ General configuration
...
+   Example patterns:
+
+   - ``'**'`` -- all files in the source directory and subdirectories, recursively
+   - ``'library/xml'`` -- just the ``library/xml`` directory
+   - ``'library/xml*'`` -- all files and directories starting with ``library/xml``
+   - ``'**/doc'`` -- all ``doc`` directories (this might be useful if
+     documentation is co-located with source files)
+
+   :confval:`include_patterns` is also consulted when looking for static files
+   in :confval:`html_static_path` and :confval:`html_extra_path`.
```
tk0miya, line 237: Is it helpful to apply `include_patterns` to static files and extra files? If applied, we need to specify asset files here:
```
include_patterns = ['*.rst', '*.css', '*.js', '*.png']
```

Is this expected?

tk0miya:
Sorry for the late review. But it would be better to refine this. +1 for adding this configuration.

PR #10519, opened 2022-06-04
Add Changelog to project_urls

adamchainz:
Make it easy to jump from the PyPI page to what has change.d

### Feature or Bugfix
- Feature

### Purpose
description says it all

### Detail
n/a

### Relates
n/a

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ setup(
     long_description_content_type='text/x-rst',
     project_urls={
         "Code": "https://github.com/sphinx-doc/sphinx",
+        "Changelog": "https://www.sphinx-doc.org/en/master/changes.html",
         "Issue tracker": "https://github.com/sphinx-doc/sphinx/issues",
     },
     zip_safe=False,


AA-Turner:
What does this render as in PyPI? Does it have a special icon?

A

setup.py
```diff
@@ -67,6 +67,7 @@
     long_description_content_type='text/x-rst',
     project_urls={
         "Code": "https://github.com/sphinx-doc/sphinx",
+        "Changelog": "https://www.sphinx-doc.org/en/master/changes.html",
```
AA-Turner, line 70: For consistency with the name of the file.
```suggestion
        "Changes": "https://www.sphinx-doc.org/en/master/changes.html",
```

adamchainz:
Yes it has an icon: https://github.com/pypa/warehouse/blob/3c93c0365ea7afafbbe0bbe5e76f9a6c29559afd/warehouse/templates/packaging/detail.html#L24

See example here: https://pypi.org/project/apig-wsgi/

setup.py
adamchainz: I would stick with "changelog" to be consistent with the page title and the many other projects that have a "changelog" link in their sidebar (I must have contributed this to like a hundred repos now).

AA-Turner (approved):

tk0miya (approved):
LGTM

tk0miya:
Thank you for your contribution. Please wait for the next release, 5.1.0.

adamchainz:
Thank you for maintaining Sphinx 😃

PR #10520, opened 2022-06-04
agogo theme: Wrong use of sidebar* configs

mgeier:
There are two different things called "sidebar":

* a part of the "outer" page layout, CSS class `sphinxsidebar`
* a kind of text box to be used within the page content using the `sidebar` directive, CSS class `sidebar`

The config values `sidebarwidth` and `rightsidebar` are supposed to control the former.

### Feature or Bugfix

- Bugfix


AA-Turner:
`float: right` has been there since the theme was created (03aaed9), and the right hand sidebar code has been there since version 2.3.0 (a950df1).

Has this been wrong for that long?

A

mgeier:
> Has this been wrong for that long?

Apparently.

But I guess the `rightsidebar` option is barely used at all, maybe nobody has ever used it with the agogo theme?
And even if somebody has used it, they might not have used the `sidebar` directive, or they just haven't noticed.

And the wrong sidebar width might also be barely noticeable.

I didn't notice that on a rendered page (does anyone even use `agogo`?), I just happened to notice it when looking at your PR diff: https://github.com/sphinx-doc/sphinx/pull/10493/files.

mgeier:
> `float: right` has been there since the theme was created

That's in `basic` as well, so it has no effect.

AA-Turner (approved):
I agree with your logic, this CSS is currently incorrect. The [theme options page](https://www.sphinx-doc.org/en/master/usage/theming.html#builtin-themes) notes:

> * sidebarwidth (CSS length): Width of the sidebar, default 20em.
> * rightsidebar (true or false): Put the sidebar on the right side. Defaults to True.

It is implicit, but reasonably clear, that it is talking about the gutter sidebar rather than the reST `.. sidebar::` directive.

A

mgeier:
I wrote [above](https://github.com/sphinx-doc/sphinx/pull/10520#issuecomment-1146644014):

> But I guess the `rightsidebar` option is barely used at all, maybe nobody has ever used it with the agogo theme?

I was wrong, because `rightsidebar` is set to `true` by default:

https://github.com/sphinx-doc/sphinx/blob/af425c069f6f51ba8fb3c348fd53479f0434a779/sphinx/themes/agogo/theme.conf#L11

See also https://sphinx-themes.org/sample-sites/default-agogo/.

So the `sidebar` *directive* is also right by default, as it is in all other themes.

And with my suggested change, it stays right, even if `rightsidebar` is set to `False`.

Issue #10528, opened 2022-06-05
Enable Mypy "strict optional"


danieleades:
Mypy has 'strict optional' checking disabled, as per https://github.com/sphinx-doc/sphinx/pull/4933

this is hiding a lot of type errors in the code base, and should be enabled.

Probably makes sense to do this incrementally, as it will require quite a few changes.

danieleades:
the `rebuild` arg of `add_config_value` is consistently used with a value of `None`, whereas the type of the arg is `bool | str`. what should this value be (rather than None)?

```python
def setup(app: Sphinx) -> Dict[str, Any]:
    app.add_builder(TexinfoBuilder)

    app.add_config_value('texinfo_documents', default_texinfo_documents, None)
    app.add_config_value('texinfo_appendices', [], None)
    app.add_config_value('texinfo_elements', {}, None)
    app.add_config_value('texinfo_domain_indices', True, None, [list])
    app.add_config_value('texinfo_show_urls', 'footnote', None)
    app.add_config_value('texinfo_no_detailmenu', False, None)
    app.add_config_value('texinfo_cross_references', True, None)

    return {
        'version': 'builtin',
        'parallel_read_safe': True,
        'parallel_write_safe': True,
    }
```

AA-Turner:
There are very few recent issues that would have been solved by stricter `None` checking -- beyond type purity, I'm not sure what the goal is. I support the general effort, but I'm just not sure on prioritisation.

A

AA-Turner:
A quick test (after fixing some Windows typing issues, which weren't really issues):

```shell
(sphinx) S:\Development\sphinx>mypy sphinx
<... 1399 lines of output elided ...>
Found 916 errors in 92 files (checked 175 source files)
```

A

danieleades:
Id consider it more an aspirational goal, than a high priority.
I think stricter type checking has more value when writing new code, rather than weeding out errors in battle-tested legacy code

tk0miya:
+1: I agree and understand enabling "strict optional" is cool and worthy. But it's very tough work. So we need your help. 

When I introduce annotations to Sphinx, I gave up to use "Optional" for our code and disable the strict check.

danieleades:
> +1: I agree and understand enabling "strict optional" is cool and worthy. But it's very tough work. So we need your help.
> 
> When I introduce annotations to Sphinx, I gave up to use "Optional" for our code and disable the strict check.

[...]

PR #10530, opened 2022-06-05
Increase static typing strictness

AA-Turner:
Goes from 916  to 650 errors with `strict_optional = True`.

### Feature or Bugfix
- Bugfix
- Refactoring

xref: #10528

A

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -295,7 +295,7 @@ class Sphinx:
 
     def _create_fresh_env(self) -> BuildEnvironment:
         env = BuildEnvironment(self)
-        env._fresh_env_used = True  # type: ignore
+        self._fresh_env_used = True
         return env
 
     def _load_existing_env(self, filename: str) -> BuildEnvironment:
@@ -304,16 +304,16 @@ class Sphinx:
                 with open(filename, 'rb') as f:
                     env = pickle.load(f)
                     env.setup(self)
-                    env._fresh_env_used = False
+                    self._fresh_env_used = False
         except Exception as err:
             logger.info(__('failed: %s'), err)
             env = self._create_fresh_env()
         return env
 
     def _post_init_env(self) -> None:
-        if self.env._fresh_env_used:  # type: ignore  # NoQA
+        if self._fresh_env_used:
             self.env.find_files(self.config, self.builder)
-        del self.env._fresh_env_used  # type: ignore  # NoQA
+        del self._fresh_env_used
 
     def preload_builder(self, name: str) -> None:
         self.registry.preload_builder(self, name)
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -326,6 +326,8 @@ class Sphinx:
         return self.registry.create_builder(self, name, self.env)
 
     def _init_builder(self) -> None:
+        if not hasattr(self.builder, "env"):
+            self.builder.set_environment(self.env)
         self.builder.init()
         self.events.emit('builder-inited')
 
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -1003,8 +1003,8 @@ class Sphinx:
 
         self.registry.add_js_file(filename, priority=priority, **kwargs)
         if hasattr(self, 'builder') and hasattr(self.builder, 'add_js_file'):
-            self.builder.add_js_file(filename, priority=priority,
-                                     **kwargs)  # type: ignore[attr-defined]
+            self.builder.add_js_file(filename,  # type: ignore[attr-defined]
+                                     priority=priority, **kwargs)
 
     def add_css_file(self, filename: str, priority: int = 500, **kwargs: Any) -> None:
         """Register a stylesheet to include in the HTML output.
@@ -1065,8 +1065,8 @@ class Sphinx:
         logger.debug('[app] adding stylesheet: %r', filename)
         self.registry.add_css_files(filename, priority=priority, **kwargs)
         if hasattr(self, 'builder') and hasattr(self.builder, 'add_css_file'):
-            self.builder.add_css_file(filename, priority=priority,
-                                      **kwargs)  # type: ignore[attr-defined]
+            self.builder.add_css_file(filename,  # type: ignore[attr-defined]
+                                      priority=priority, **kwargs)
 
     def add_stylesheet(self, filename: str, alternate: bool = False, title: str = None
                        ) -> None:
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -245,10 +245,16 @@ class Sphinx:
 
         # create the project
         self.project = Project(self.srcdir, self.config.source_suffix)
-        # create the builder
-        self.builder = self.create_builder(buildername)
+
         # set up the build environment
         self.env = self._init_env(freshenv)
+
+        # create the builder
+        self.builder = self.create_builder(buildername)
+
+        # build environment post-initialisation, after creating the builder
+        self._post_init_env()
+
         # set up the builder
         self._init_builder()
 
@@ -289,19 +295,25 @@ class Sphinx:
 
     def _create_fresh_env(self) -> BuildEnvironment:
         env = BuildEnvironment(self)
-        env.find_files(self.config, self.builder)
+        env._fresh_env_used = True  # type: ignore
         return env
 
-    def _load_existing_env(self, filename) -> BuildEnvironment:
+    def _load_existing_env(self, filename: str) -> BuildEnvironment:
         try:
             with progress_message(__('loading pickled environment')):
                 with open(filename, 'rb') as f:
                     env = pickle.load(f)
                     env.setup(self)
+                    env._fresh_env_used = False
         except Exception as err:
             logger.info(__('failed: %s'), err)
             env = self._create_fresh_env()
         return env
+
+    def _post_init_env(self) -> None:
+        if self.env._fresh_env_used:  # type: ignore  # NoQA
+            self.env.find_files(self.config, self.builder)
+        del self.env._fresh_env_used  # type: ignore  # NoQA
 
     def preload_builder(self, name: str) -> None:
         self.registry.preload_builder(self, name)
@@ -311,10 +323,9 @@ class Sphinx:
             logger.info(__('No builder selected, using default: html'))
             name = 'html'
 
-        return self.registry.create_builder(self, name)
+        return self.registry.create_builder(self, name, self.env)
 
     def _init_builder(self) -> None:
-        self.builder.set_environment(self.env)
         self.builder.init()
         self.events.emit('builder-inited')
 
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -135,9 +135,6 @@ class Sphinx:
         self.phase = BuildPhase.INITIALIZATION
         self.verbosity = verbosity
         self.extensions: Dict[str, Extension] = {}
-        self.builder: Optional[Builder] = None
-        self.env: Optional[BuildEnvironment] = None
-        self.project: Optional[Project] = None
         self.registry = SphinxComponentRegistry()
 
         # validate provided directories
@@ -251,7 +248,7 @@ class Sphinx:
         # create the builder
         self.builder = self.create_builder(buildername)
         # set up the build environment
-        self._init_env(freshenv)
+        self.env = self._init_env(freshenv)
         # set up the builder
         self._init_builder()
 
@@ -283,20 +280,28 @@ class Sphinx:
             else:
                 logger.info(__('not available for built-in messages'))
 
-    def _init_env(self, freshenv: bool) -> None:
+    def _init_env(self, freshenv: bool) -> BuildEnvironment:
         filename = path.join(self.doctreedir, ENV_PICKLE_FILENAME)
         if freshenv or not os.path.exists(filename):
-            self.env = BuildEnvironment(self)
-            self.env.find_files(self.config, self.builder)
+            return self._create_fresh_env()
         else:
-            try:
-                with progress_message(__('loading pickled environment')):
-                    with open(filename, 'rb') as f:
-                        self.env = pickle.load(f)
-                        self.env.setup(self)
-            except Exception as err:
-                logger.info(__('failed: %s'), err)
-                self._init_env(freshenv=True)
+            return self._load_existing_env(filename)
+
+    def _create_fresh_env(self) -> BuildEnvironment:
+        env = BuildEnvironment(self)
+        env.find_files(self.config, self.builder)
+        return env
+
+    def _load_existing_env(self, filename) -> BuildEnvironment:
+        try:
+            with progress_message(__('loading pickled environment')):
+                with open(filename, 'rb') as f:
+                    env = pickle.load(f)
+                    env.setup(self)
+        except Exception as err:
+            logger.info(__('failed: %s'), err)
+            env = self._create_fresh_env()
+        return env
 
     def preload_builder(self, name: str) -> None:
         self.registry.preload_builder(self, name)
@@ -986,7 +991,7 @@ class Sphinx:
             kwargs['defer'] = 'defer'
 
         self.registry.add_js_file(filename, priority=priority, **kwargs)
-        if hasattr(self.builder, 'add_js_file'):
+        if hasattr(self, 'builder') and hasattr(self.builder, 'add_js_file'):
             self.builder.add_js_file(filename, priority=priority, **kwargs)  # type: ignore
 
     def add_css_file(self, filename: str, priority: int = 500, **kwargs: Any) -> None:
@@ -1047,7 +1052,7 @@ class Sphinx:
         """
         logger.debug('[app] adding stylesheet: %r', filename)
         self.registry.add_css_files(filename, priority=priority, **kwargs)
-        if hasattr(self.builder, 'add_css_file'):
+        if hasattr(self, 'builder') and hasattr(self.builder, 'add_css_file'):
             self.builder.add_css_file(filename, priority=priority, **kwargs)  # type: ignore
 
     def add_stylesheet(self, filename: str, alternate: bool = False, title: str = None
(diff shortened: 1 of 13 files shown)


AA-Turner:
cc: @danieleades 

A

sphinx/util/parallel.py
```diff
@@ -49,7 +55,7 @@ def __init__(self, nproc: int) -> None:
         # task arguments
         self._args: Dict[int, Optional[List[Any]]] = {}
         # list of subprocesses (both started and waiting)
-        self._procs: Dict[int, multiprocessing.context.ForkProcess] = {}
+        self._procs: Dict[int, "ForkProcess"] = {}
```
tk0miya: Is this quotation needed?

sphinx/registry.py
```diff
@@ -153,11 +153,11 @@ def preload_builder(self, app: "Sphinx", name: str) -> None:
 
             self.load_extension(app, entry_point.module)
 
-    def create_builder(self, app: "Sphinx", name: str) -> Builder:
+    def create_builder(self, app: "Sphinx", name: str, env: BuildEnvironment) -> Builder:
         if name not in self.builders:
             raise SphinxError(__('Builder name %s not registered') % name)
 
-        return self.builders[name](app)
+        return self.builders[name](app, env)
```
tk0miya: This is a breaking change if the 3rd party builder overrides the `__init__()` method. We need to handle `TypeError` if it does not take the `env` argument:
```suggestion
        try:
            return self.builders[name](app, env)
        except TypeError:
            warnings.warn("The 'env' argument to Builder will be required from Sphinx 7.")
            builder = self.builders[name](app)
            builder.set_environment(env)
            return builder
```

sphinx/application.py
```diff
@@ -283,20 +286,34 @@ def _init_i18n(self) -> None:
...
+                    env._fresh_env_used = False
+        except Exception as err:
+            logger.info(__('failed: %s'), err)
+            env = self._create_fresh_env()
+        return env
+
+    def _post_init_env(self) -> None:
+        if self.env._fresh_env_used:  # type: ignore  # NoQA
+            self.env.find_files(self.config, self.builder)
+        del self.env._fresh_env_used  # type: ignore  # NoQA
```
tk0miya: How about storing this flag to app object instead?

sphinx/application.py
```diff
@@ -135,9 +135,6 @@ def __init__(self, srcdir: str, confdir: Optional[str], outdir: str, doctreedir:
         self.phase = BuildPhase.INITIALIZATION
         self.verbosity = verbosity
         self.extensions: Dict[str, Extension] = {}
-        self.builder: Optional[Builder] = None
-        self.env: Optional[BuildEnvironment] = None
-        self.project: Optional[Project] = None
```
tk0miya, line 140: Note: Nobody can run their own code until initializing the builder. So this initialization can be removed.

tk0miya (changes requested):

sphinx/util/parallel.py
AA-Turner: Updated locally.

sphinx/registry.py
AA-Turner: As I added the warning code to `Builder.__init__`, I think just making the default argument `None` would preserve backwards compatability here?

A

sphinx/application.py
AA-Turner: Done locally

sphinx/application.py
AA-Turner: Sorry, I don't understand this one -- what are you referring to in "this initialization can be removed"?

A

sphinx/application.py
```diff
@@ -986,7 +1002,7 @@ def add_js_file(self, filename: str, priority: int = 500,
             kwargs['defer'] = 'defer'
 
         self.registry.add_js_file(filename, priority=priority, **kwargs)
-        if hasattr(self.builder, 'add_js_file'):
+        if hasattr(self, 'builder') and hasattr(self.builder, 'add_js_file'):
             self.builder.add_js_file(filename, priority=priority, **kwargs)  # type: ignore
```
danieleades: we should avoid blanket 'ignore' statements' mypy can be configured to show you the specific error code, and you can use that in the ignore statement

`type: ignore[error-code]`

(what is the error here?)

sphinx/application.py
```diff
@@ -1047,7 +1063,7 @@ def add_css_file(self, filename: str, priority: int = 500, **kwargs: Any) -> Non
         """
         logger.debug('[app] adding stylesheet: %r', filename)
         self.registry.add_css_files(filename, priority=priority, **kwargs)
-        if hasattr(self.builder, 'add_css_file'):
+        if hasattr(self, 'builder') and hasattr(self.builder, 'add_css_file'):
             self.builder.add_css_file(filename, priority=priority, **kwargs)  # type: ignore
```
danieleades: no blanket "ignores"

sphinx/builders/html/__init__.py
```diff
@@ -199,8 +200,8 @@ class StandaloneHTMLBuilder(Builder):
     imgpath: str = None
     domain_indices: List[Tuple[str, Type[Index], List[Tuple[str, List[IndexEntry]]], bool]] = []  # NOQA
```
danieleades: no blanket "ignores"/"noqas"

also, this type is horrible. can this be cleaned up with an alias (or two)?

sphinx/registry.py
tk0miya: Does this builder work as before?

```
class MyBuilder(Builder):
    def __init__(self, app):
        # my initialization code here
        ...

        super().__init__(app)
```

danieleades (changes requested):
there's a few blanket warnings suppressions that can be tightened, although these are technically *adjacent* to the changed code

sphinx/registry.py
AA-Turner: Yep, note in `builders/__init__.py`:

```python
class Builder:
    ... # class setup elided

    def __init__(self, app: "Sphinx", env: BuildEnvironment = None) -> None:
        # instance setup elided

        if env is not None:
            self.env: BuildEnvironment = env
            self.env.set_versioning_method(self.versioning_method,
                                           self.versioning_compare)
        else:
            warnings.warn("The 'env' argument to Builder will be required from Sphinx 7.",
                          RemovedInSphinx70Warning, stacklevel=2)
```

So `env` will default to `None` and then raise a warning.

A

sphinx/application.py
tk0miya: Sorry for confusing. I don't have a good word for this. It simply means okay for this removal.

AA-Turner: No worries! Thanks :)

A

sphinx/application.py
AA-Turner:
```
sphinx\application.py: note: In member "add_js_file" of class "Sphinx":
sphinx\application.py:1006:13: error: "Builder" has no attribute "add_js_file"
sphinx\application.py: note: In member "add_css_file" of class "Sphinx":
sphinx\application.py:1067:13: error: "Builder" has no attribute "add_css_file"
```
Even with the literal `hasattr` check mypy still complains?

A

danieleades: you might want to add `show-error-codes = true` to the mypy config in order to get the codes to use in the 'ignore' statement

yeah mypy basically ignores `hasattr`-type statements. Ideally you'd structure things so that they're not required at all, since adding and removing attributes at runtime is basically anathema to static type checking. Obviously that's not possible for a legacy codebase that relies on this (anti)pattern. In this case you can safely ignore the type errors.

Are there any invariants/guarantees that these attributes will always exist here? Or is it genuinely possible for them to not be set?

sphinx/builders/__init__.py
```diff
@@ -75,15 +77,21 @@ class Builder:
     #: The builder supports data URIs or not.
     supported_data_uri_images = False
 
-    def __init__(self, app: "Sphinx") -> None:
+    def __init__(self, app: "Sphinx", env: BuildEnvironment = None) -> None:
```
danieleades, line 80:
```suggestion
    def __init__(self, app: "Sphinx", env: Optional[BuildEnvironment] = None) -> None:
```

sphinx/builders/html/__init__.py
```diff
@@ -199,8 +200,8 @@ class StandaloneHTMLBuilder(Builder):
...
 
-    def __init__(self, app: Sphinx) -> None:
-        super().__init__(app)
+    def __init__(self, app: Sphinx, env: BuildEnvironment = None) -> None:
```
danieleades, line 214:
```suggestion
    def __init__(self, app: Sphinx, env: Optional[BuildEnvironment] = None) -> None:
```

sphinx/application.py
AA-Turner: Well Python is a dynamic language... The `Builder` class does not have the `add_(js|css)_file` method, in core Sphinx only the `StandaloneHTMLBuilder` does. I don't really think this is an antipattern, either.

A

danieleades: > Well Python is a dynamic language... The `Builder` class does not have the `add_(js|css)_file` method, in core Sphinx only the `StandaloneHTMLBuilder` does. I don't really think this is an antipattern, either.
> 
> A

absolutely python is dynamic, and allows for runtime extension of existing objects. That said, this pattern limits the effectiveness of static analysis, IDE code completion, documentation generation, and so on. You can usually solve the same problems (extending an object with additional attributes or methods) in the static case instead, by using inheritance or even better, composition.

I'm emphatically **not** advocating for major refactoring of this pattern in this project, but I'll admit my choice of words has given away my own prejudices on the subject... If pressed, I would point out that moving errors from runtime to 'lint time' has productivity benefits *in the general* case for any codebase at scale. Not a hill I plan to die on today ;)

sphinx/builders/html/__init__.py
AA-Turner: Done

sphinx/builders/__init__.py
AA-Turner: Here I'm using `None` as a sentinel, I could have equally used `...`. I don't want to indicate to users that `None` is a valid type, which I believe using `Optional[]` syntax would do. Hence I'm -1 on both these suggestions, unless you've alternate ideas?

A

sphinx/application.py
AA-Turner: Fair enough, I have found sometimes that a typing-first approach limits the scope of solutions availible, but if we agree on gradualism that's a good thing!

A

sphinx/registry.py
tk0miya: Please try this small custom builder.

On my local, it causes a crash:
```
$ sphinx-build -b custom . _build/custom
```
```
  File "/Users/tkomiya/work/sphinx/sphinx/cmd/build.py", line 272, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
  File "/Users/tkomiya/work/sphinx/sphinx/application.py", line 253, in __init__
    self.builder = self.create_builder(buildername)
  File "/Users/tkomiya/work/sphinx/sphinx/application.py", line 326, in create_builder
    return self.registry.create_builder(self, name, self.env)
  File "/Users/tkomiya/work/sphinx/sphinx/registry.py", line 161, in create_builder
    return self.builders[name](app, env)
TypeError: __init__() takes 2 positional arguments but 3 were given
```

tk0miya: https://github.com/tk0miya/custom_builder

AA-Turner: Ahh, yes I see, sorry. One second.

A

AA-Turner: @tk0miya this should be fixed with the latest commit.

A

sphinx/registry.py
```diff
@@ -153,11 +153,20 @@ def preload_builder(self, app: "Sphinx", name: str) -> None:
...
-        return self.builders[name](app)
+        try:
+            return self.builders[name](app, env)
+        except TypeError:
+            warnings.warn(
+                f"The custom builder {name} defines a custom __init__ method without the "
+                f"'env'argument. Report this bug to the developers of your custom builder, "
+                f"this is likely not a issue with Sphinx. The 'env' argument will be required "
+                f"from Sphinx 7.", RemovedInSphinx70Warning, stacklevel=2)
+            return self.builders[name](app)
```
tk0miya: * The new code never calls `set_environment()`. As a result, custom builders will get crashed because the env object is None.
* This will also cause another warning on `Builder.__init__()`. The users will see warnings twice or more.

Please be careful to changing the API. I believe Sphinx is not only an application but also a framework. So I consider this is a mix of refactoring and API changing. Is it intended?

sphinx/builders/__init__.py
danieleades: that's not how this works. Setting a default value of `None` effectively coerces the type from `Type` to `Optional[Type]`. That is, the former desugars to the latter *implicitly*. The setting that controls this is called [`no-implicit-optional`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional).

It's better to be explicit and use the `Optional`. If using 'strict' mypy checking, failing to use it becomes a hard error.

sphinx/registry.py
AA-Turner: I was concerned the potential to call `set_environment` twice, but it seems that wouldn't be a problem, so updated. I also added a check so that we only issue one warning.

> Please be careful to changing the API. I believe Sphinx is not only an application but also a framework. So I consider this is a mix of refactoring and API changing. Is it intended?

This PR shouldn't have any end-user facing changes. I would say I consider Sphinx more of an application with a plugin mechanism though, rather than a library/framework -- we should be able to make changes that improve things on our end, although of course seeking to keep backwards compatability.

A

sphinx/builders/__init__.py
AA-Turner: I added `no_implicit_optional = True` and everything still passes -- is this what you meant?

A

danieleades: Yep, it should be an error with that config. Not sure why you're not seeing one, could be you're hitting the maximum number of errors, or it could be a false negative. In either case you should be able to convince yourself from the documentation that using the explicit `Optional` is strictly correct.

danieleades: ah i think i see the issue. you have this-

```ini
strict_optional = False
no_implicit_optional = True
```

to see these errors your would need

```ini
strict_optional = True
no_implicit_optional = True
```

if you really want to give yourself nightmares, try

```ini
strict = True
```

danieleades:
@AA-Turner this is already a herculean effort, nice work!

If you're not able to squash *all* of the errors in this PR, you may want to consider whitelisting a subset of files with `strict_optional = False`, so that the CI checks can run on the rest. Then the whitelist can be incrementally shrunk in subsequent PRs. See my comment over here - https://github.com/sphinx-doc/sphinx/issues/10528#issuecomment-1147611888

that would look something like

current:

```ini
[mypy]
python_version = 3.6
disallow_incomplete_defs = True
show_column_numbers = True
show_error_context = True
ignore_missing_imports = True
follow_imports = skip
check_untyped_defs = True
warn_unused_ignores = True
strict_optional = False
```

new:

```ini
[mypy]
python_version = 3.6
disallow_incomplete_defs = True
show_column_numbers = True
show_error_context = True
ignore_missing_imports = True
follow_imports = skip
check_untyped_defs = True
warn_unused_ignores = True

[mypy-some.module]
strict_optional = False

[mypy-some.other.module]
strict_optional = False
```

(i think? i usually use `pyproject.toml` for configuring mypy)

AA-Turner:
I'm not going to do the whitelist approach -- I got to 34 modules and the list is ridiculous (100+ lines long) and there are still 244 errors reported.

My progress is in the details block for posterity, but I don't reccomend adding it to `setup.cfg`.

<details>
<summary>Details</summary>

```cfg
strict_optional = True
no_implicit_optional = True

[mypy-sphinx.application]
strict_optional = False

[mypy-sphinx.writers.latex]
strict_optional = False

[mypy-sphinx.builders.latex]
strict_optional = False

[mypy-sphinx.builders.html]
strict_optional = False

[mypy-sphinx.builders.gettext]
strict_optional = False

[mypy-sphinx.builders.singlehtml]
strict_optional = False

[mypy-sphinx.builders.dirhtml]
strict_optional = False

[mypy-sphinx.builders.latex.transforms]
strict_optional = False

[mypy-sphinx.domains.c]
strict_optional = False

[mypy-sphinx.domains.cpp]
strict_optional = False

[mypy-sphinx.domains.javascript]
strict_optional = False

[mypy-sphinx.domains.python]
strict_optional = False

[mypy-sphinx.domains.changeset]
strict_optional = False

[mypy-sphinx.builders.xml]
strict_optional = False

[mypy-sphinx.builders.manpage]
strict_optional = False

[mypy-sphinx.builders.texinfo]
strict_optional = False

[mypy-sphinx.writers.texinfo]
strict_optional = False

[mypy-sphinx.writers.text]
strict_optional = False

[mypy-sphinx.writers.html5]
strict_optional = False

[mypy-sphinx.writers.html]
strict_optional = False

[mypy-sphinx.environment]
strict_optional = False

[mypy-sphinx.environment.adapters.indexentries]
strict_optional = False

[mypy-sphinx.util.docfields]
strict_optional = False

[mypy-sphinx.directives.code]
strict_optional = False

[mypy-sphinx.transforms.post_transforms.images]
strict_optional = False

[mypy-sphinx.builders.linkcheck]
strict_optional = False

[mypy-sphinx.ext.autosummary]
strict_optional = False

[mypy-sphinx.ext.autodoc]
strict_optional = False

[mypy-sphinx.ext.apidoc]
strict_optional = False

[mypy-sphinx.ext.autosummary.generate]
strict_optional = False

[mypy-sphinx.ext.imgmath]
strict_optional = False

[mypy-sphinx.ext.intersphinx]
strict_optional = False

[mypy-sphinx.ext.mathjax]
strict_optional = False

[mypy-sphinx.ext.graphviz]
strict_optional = False

[mypy-sphinx.ext.inheritance_diagram]
strict_optional = False
```

</details>

A

AA-Turner:
Followup: 

> If you're not able to squash all of the errors in this PR

This PR was never about getting all of them, it started as ensuring MyPy passes on Windows and then I fixed some "easy" targets. C.f. Zen, pragmatism & purity...!

A

PR #10532, opened 2022-06-06
Add ``sphinx.ext.collapse``

AA-Turner:
[``sphinx.ext.collapse``](https://sphinx--10532.org.readthedocs.build/en/10532/usage/extensions/collapse.html) adds support for [collapsible content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) in HTML.

### Feature or Bugfix
- Feature

### Purpose
- There have been several requests for support of `<details>`, and it is a useful feature.

### Relates
- https://discuss.python.org/t/adopting-the-diataxis-framework-for-python-documentation/15072/43?u=aa-turner

cc: @vsajip @EpicWink @ezio-melotti (from the Discuss thread)

([Rendered documentation](https://sphinx--10532.org.readthedocs.build/en/10532/usage/extensions/collapse.html))

A

### DIFF
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -1083,37 +1083,6 @@ Including content based on tags
       This directive is designed to control only content of document.  It could
       not control sections, labels and so on.
 
-.. _collapsible:
-
-Using collapsible content
--------------------------
-
-.. rst:directive:: .. collapsible:: <summary description>
-
-   For HTML builders, place the content of the directive into a HTML `details
-   disclosure`_ element, with the *summary description* text included as the
-   summary for the element. The *summary description* text is parsed as
-   reStructuredText, and can be broken over multiple lines if required. An
-   example and the equivalent output are shown below:
-
-   .. code-block:: reStructuredText
-
-      .. collapsible:: ``literal`` and **bold** content.
-
-         This is the body of the directive.
-
-   .. collapsible:: ``literal`` and **bold** content.
-
-      This is the body of the directive.
-
-   Only the HTML 5 output format supports collapsible content. For other builders,
-   the *summary description* text and the body of the directive are rendered in
-   the document.
-
-   .. versionadded:: 5.1
-
-   .. _details disclosure: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details
-
 .. _table-directives:
 
 Tables
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -1107,8 +1107,8 @@ Using collapsible content
       This is the body of the directive.
 
    Only the HTML 5 output format supports collapsible content. For other builders,
-   the body of the directive is rendered in the document and the *summary
-   description* text is ignored.
+   the *summary description* text and the body of the directive are rendered in
+   the document.
 
    .. versionadded:: 5.1
 
--- a/doc/usage/restructuredtext/directives.rst
+++ b/doc/usage/restructuredtext/directives.rst
@@ -1083,6 +1083,37 @@ Including content based on tags
       This directive is designed to control only content of document.  It could
       not control sections, labels and so on.
 
+.. _collapsible:
+
+Using collapsible content
+-------------------------
+
+.. rst:directive:: .. collapsible:: <summary description>
+
+   For HTML builders, place the content of the directive into a HTML `details
+   disclosure`_ element, with the *summary description* text included as the
+   summary for the element. The *summary description* text is parsed as
+   reStructuredText, and can be broken over multiple lines if required. An
+   example and the equivalent output are shown below:
+
+   .. code-block:: reStructuredText
+
+      .. collapsible:: ``literal`` and **bold** content.
+
+         This is the body of the directive.
+
+   .. collapsible:: ``literal`` and **bold** content.
+
+      This is the body of the directive.
+
+   Only the HTML 5 output format supports collapsible content. For other builders,
+   the body of the directive is rendered in the document and the *summary
+   description* text is ignored.
+
+   .. versionadded:: 5.1
+
+   .. _details disclosure: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details
+
 .. _table-directives:
 
 Tables
--- a/sphinx/ext/collapse.py
+++ b/sphinx/ext/collapse.py
@@ -34,23 +34,23 @@ class summary(nodes.General, nodes.TextElement):
     """
 
 
-def visit_collapsible(self: HTML5Translator, node: nodes.Element) -> None:
+def visit_collapsible(translator: HTML5Translator, node: nodes.Element) -> None:
     if node.get('collapsible_open'):
-        self.body.append(self.starttag(node, 'details', open='open'))
+        translator.body.append(translator.starttag(node, 'details', open='open'))
     else:
-        self.body.append(self.starttag(node, 'details'))
+        translator.body.append(translator.starttag(node, 'details'))
 
 
-def depart_collapsible(self: HTML5Translator, node: nodes.Element) -> None:
-    self.body.append('</details>\n')
+def depart_collapsible(translator: HTML5Translator, node: nodes.Element) -> None:
+    translator.body.append('</details>\n')
 
 
-def visit_summary(self: HTML5Translator, node: nodes.Element) -> None:
-    self.body.append(self.starttag(node, 'summary'))
+def visit_summary(translator: HTML5Translator, node: nodes.Element) -> None:
+    translator.body.append(translator.starttag(node, 'summary'))
 
 
-def depart_summary(self: HTML5Translator, node: nodes.Element) -> None:
-    self.body.append('</summary>\n')
+def depart_summary(translator: HTML5Translator, node: nodes.Element) -> None:
+    translator.body.append('</summary>\n')
 
 
 class Collapsible(SphinxDirective):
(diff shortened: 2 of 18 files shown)


AA-Turner:
Screenshots:

Note that both the summary line and the body are parsed as reStructuredText, and the body can have sections etc.

<details>
<summary>reST source used:</summary>

```rest
Collapsible directive tests
===========================

.. collapsible::

   Default section summary line

.. collapsible:: Custom summary line for the collapsible content

   Collapsible sections can also have custom summary lines

.. collapsible::

   Collapsible sections can have normal reST content such as **bold** and
   *emphasised* text, and also links_!

   .. _links: https://link.example/

.. collapsible::

   Collapsible sections can have sections:

   A Section
   ---------

   Some words within a section, as opposed to outwith the section.

.. collapsible:: Summary text here with **bold** and *em* and a :rfc:`2324`
                 reference! That was a newline in the reST source! We can also
                 have links_ and `more links <https://link2.example/>`__.

   This is some text!
```

</details>

<details>
<summary>Output HTML</summary>

image

</details>

<details>
<summary>Output HTML (expanded)</summary>

image

</details>

A

pradyunsg:
Relevant prior art for this: https://github.com/tk0miya/sphinxcontrib-details-directive

AA-Turner:
That would have saved ... a lot of work! More the fool for not looking if this had been implemented before. I'll look through and see if there's anything @tk0miya implemented that I haven't.

A

vsajip:
> That would have saved ... a lot of work! More the fool for not looking

Not at all, and thanks for looking at this, Adam! I wasn't aware of that extension, either. Perhaps a review should be done of other `sphinxcontrib` extensions like this to see if they would benefit from folding into Sphinx itself, so that everyone benefits without having to know about available extensions etc. (I guess they're not well-enough known to the wider community).

doc/usage/restructuredtext/directives.rst
```diff
@@ -1030,6 +1030,37 @@ Including content based on tags
...
+
+         This is the body of the directive.
+
+   .. collapsible:: ``literal`` and **bold** content.
+
+      This is the body of the directive.
+
+   Only the HTML 5 output format supports collapsible content. For other builders,
+   the body of the directive is rendered in the document and the *summary
+   description* text is ignored.
```
vsajip: Rather than being ignored, perhaps it could be rendered as an ordinary paragraph just before the body.

sphinx/util/nodes.py
```diff
@@ -626,6 +626,25 @@ def process_only_nodes(document: Node, tags: "Tags") -> None:
...
+
+    for node in document.findall(addnodes.collapsible):
+        if builder_name not in HTML_5_BUILDERS:
+            # A comment on the comment() nodes being inserted: replacing by [] would
+            # result in a "Losing ids" exception if there is a target node before
+            # the only node, so we make sure docutils can transfer the id to
+            # something, even if it's just a comment and will lose the id anyway...
+            # Slicing node.children removes the first element, which will always
+            # be a collapsible_summary node.
+            node.replace_self(node.children[1:] or nodes.comment())
```
vsajip: Or perhaps leave it in as an ordinary paragraph, as I said above.

vsajip (commented):
My few comments here are on the feature itself rather than the implementation. Perhaps it would be useful to compare and contrast with the `sphinxcontrib` "details" extension (e.g. incorporating an "open" flag option). Of course, that extension remains useful for older Sphinx versions, which is perhaps why it was developed as such.

doc/usage/restructuredtext/directives.rst
AA-Turner: Done

tk0miya:
I hesitated to add a new directive to Sphinx for some reasons.

* About dialect. This will make the difference in the reST between docutils and Sphinx larger. I know reST allows to enhance its syntax via directives and roles. But I prefer to keep them almost the same as possible. If you think the collapsible content is important, how about proposing this to the docutils?
* About output format. I consider the principle of reST and Sphinx as "single input, multiple outputs". So it's not better to add a notation only for HTML.

This is why I created sphinxcontrib-details as an extension and not created as a built-in feature.

pradyunsg:
FWIW, I also lean slightly toward pushing users to use the extension instead of baking this into Sphinx itself.

EpicWink:
> FWIW, I also lean slightly toward pushing users to use the extension instead of baking this into Sphinx itself.

I agree. I think you're referring to [`sphinx-design`](https://sphinx-design.readthedocs.io/en/latest/dropdowns.html)?

pradyunsg:
No. See https://github.com/sphinx-doc/sphinx/pull/10532#issuecomment-1147845664

AA-Turner:
Ok, if we don't go ahead with this we should at least improve documentation. @tk0miya would you consider moving your extension to @sphinx-contrib to make it more discoverable / 'official'?

A

tk0miya:
What do you mean "official"? I'm okay to move the repo to the sphinx-contrib org. But I don't think it does not mean the extension is an "official" extension.

tk0miya:
Just moved: https://github.com/sphinx-contrib/sphinxcontrib-details-directive

pradyunsg:
Let's close this out. We still would want to fix the extension though, since it's borked on the newest Sphinx versions.

electric-coder:
I've been using [sphinx-design](https://sphinx-design.readthedocs.io/en/latest/dropdowns.html) to get `details` (linked above) and there's also [sphinx-toolbox](https://sphinx-toolbox.readthedocs.io/en/stable/extensions/collapse.html) which also works well.

While it'd be nice to have this integrated in Sphinx core I don't think it's a priority since there are viable alternatives.

AA-Turner:
https://github.com/python/cpython/pull/111743

Though wouldn't be immediately useful for CPython.

sphinx/ext/collapse.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @staticmethod
+    def _parse_content(
+        state: RSTState,
+        node: nodes.Element,
+        content: StringList,
+        offset: int,
+    ) -> list[nodes.Node]:
+        # Same as util.nested_parse_with_titles but try to handle nested
+        # sections which should be raised higher up the doctree.
```
chrisjsewell: I have to say I'm completely against this kind of "hacking" into docutils.
If you want to do something like this, then either it needs to be upstreamed to docutils, or it needs much, much more testing than is in this PR.
simple example: what happens if you put your `collapse` directive inside a `note` directive?

Not only is it already brittle in rst, but also its definitely not supported by myst 😅

salampro6444 (approved):

sphinx/ext/collapse.py
chrisjsewell: I certainly get the desire to have nested parsing with sections (in fact I am trying to achieve it in https://sphinx-rust.readthedocs.io)

but this should be a "core" function within sphinx/docutils, with good testing, rather than having multiple implementations scattered across the code base

sphinx/ext/collapse.py
```diff
@@ -0,0 +1,206 @@
...
+            return
+
+        for summary_node in document.findall(summary):
+            summary_para = nodes.paragraph('', '', *summary_node)
+            summary_node.replace_self(summary_para)
+
+        for collapsible_node in document.findall(collapsible):
+            # A comment on the comment() nodes being inserted: replacing by [] would
+            # result in a "Losing ids" exception if there is a target node before
+            # the only node, so we make sure docutils can transfer the id to
```
chrisjsewell: I guess this was copy-pasted from the `OnlyNodeTransform` 😉 
again I feel this is a horrible hack, that just serves to hide the actual issue rather than address it;
the id should not be lost without a warning, it should most probably be propagated to the next sibling

AA-Turner: > I guess this was copy-pasted from the `OnlyNodeTransform` 😉

Yes, I should update the text...

> again I feel this is a horrible hack, that just serves to hide the actual issue rather than address it;
> the id should not be lost without a warning, it should most probably be propagated to the next sibling

I think it is fine to loose the ID here. This function runs to remove the nodes from the doctree for unsupported builders. Does it make sense to keep an ID for a node that doesn't exist anymore?

As it happens, in every non-pathalogical case, children will exist, as the summary node is guaranteed by the ``.. collapse`` directive, and will always have children. We could therefore remove the ``comment()`` logic, I believe.

A

sphinx/ext/collapse.py
```diff
@@ -0,0 +1,206 @@
...
+class summary(nodes.General, nodes.TextElement):
+    """Node for the description for collapsible content.
+
+    This is used by the :rst:dir:`collapse` directive.
+    """
+
+
+def visit_collapsible(translator: HTML5Translator, node: nodes.Element) -> None:
+    if node.get('collapsible_open'):
+        translator.body.append(translator.starttag(node, 'details', open='open'))
```
chrisjsewell, line 36: It would be good to add a class to this, something like `sphinx-ext-collapse`, so that they can be targeted more specifically in css.

chrisjsewell (changes requested):

chrisjsewell:
Hey all, I have a different solution to propose #12507 that I think would be a lot better 🙃

As the author/maintainer of https://sphinx-design.readthedocs.io/en/furo-theme/dropdowns.html, if this was merged, I would in a heartbeat basically deprecate `dropdown` and push people to use the `collapsible` option on admonitions

electric-coder:
@chrisjsewell from a simplicity POV PR 12506 makes the most sense because having the option integrated under existing admonitions avoids having an additional directive and extension.

AA-Turner:
Closing in favour of either https://github.com/sphinx-contrib/sphinxcontrib-details-directive or #12507, though the drafted words and tests from this PR may be able to be re-used.

A

PR #10548, opened 2022-06-13
Fix minor search summary issues

shiftinv:
Subject: Fix minor search summary issues

### Feature or Bugfix
- Bugfix

### Purpose
I believe all of these are mainly cosmetic minor regressions from #10028.
See [5.x](https://sphinx-test-stuff.readthedocs.io/en/tmp-5.x/search.html?q=gettext) vs [this branch](https://sphinx-test-stuff.readthedocs.io/en/tmp-fix-searchtools/search.html?q=gettext) for comparison (with rtd's server-side search disabled as it's not an issue when that's used).

### Details
- Summary element type should be `<p>` instead of `<div>`, fixes styling (originally changed in 339cab76579ee2b9b0c6ed85cdf3f89afe589067)
- The lowercase text should only be used for context matching, not for the final output
- Removal of header links was broken

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -158,7 +158,7 @@ const Search = {
     const htmlElement = document
       .createRange()
       .createContextualFragment(htmlString);
-    _removeChildren(htmlElement.querySelectorAll(".headerlink"));
+    htmlElement.querySelectorAll(".headerlink").forEach((el) => el.parentNode.removeChild(el));
     const docContent = htmlElement.querySelector('[role="main"]');
     if (docContent !== undefined) return docContent.textContent;
     console.warn(
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -516,7 +516,7 @@ const Search = {
     const top = startWithContext === 0 ? "" : "...";
     const tail = startWithContext + 240 < text.length ? "..." : "";
 
-    let summary = document.createElement("div");
+    let summary = document.createElement("p");
     summary.classList.add("context");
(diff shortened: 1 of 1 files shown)


sphinx/themes/basic/static/searchtools.js
```diff
@@ -158,7 +158,7 @@ const Search = {
     const htmlElement = document
       .createRange()
       .createContextualFragment(htmlString);
-    _removeChildren(htmlElement.querySelectorAll(".headerlink"));
+    htmlElement.querySelectorAll(".headerlink").forEach((el) => el.parentNode.removeChild(el));
```
AA-Turner, line 161: What's the functional change here?

AA-Turner (commented):
Are these severe enough regressions to go in 5.0.2?

A

sphinx/themes/basic/static/searchtools.js
shiftinv: This previously used jQuery's `htmlElement.find('.headerlink').remove()`, which would remove the `.headerlink`s themselves.
As far as I can tell, `_removeChildren` is meant to be a replacement for `.empty()` - it only takes a single element instead of a `NodeList` as returned by `querySelectorAll`, and would only remove child elements instead of the element itself; as it stands right now I'm fairly sure this particular call is effectively a no-op.

The functional change here is just getting rid of the `¶` glyphs in the output as intended.

shiftinv:
> Are these severe enough regressions to go in 5.0.2?

The `div`/`p` change might be, the original switch to `p` was marked as a breaking change in 4.0 so this regression might be considered moderate, not sure how much relies on it.
Let me know if I should split the changes into separate PRs, these initially just seemed small enough to go into one PR.

tk0miya (approved):
LGTM!

PR #10549, opened 2022-06-13
Add type info to version and project config keys

### LINKED ISSUE #10544: inventory.py does not check type of 'version' config var
### Describe the bug

I have a project in which I use the default templated conf.py. In my conf.py, I decided to generate the `release` variable automatically from git by calling a small helper function. I called this helper function `version`, thereby shadowing the `version` config var. This lead to sphinx'es inventory.py choking with the attached stack trace and unhelpful error message.

### How to Reproduce

Just use any conf.py, and set version to something other than a string or bytes object, e.g. `version = lambda x: None`

### Expected behavior

Two things:
1. In the default conf.py template, it would be good if there was a comment that `version` is a reserved variable name, even if it is not used in the template invocation.
2. In inventory.py, it would be good to add a check that `version` is in fact a string, and give a verbose error message in case it is not.

### Your project

https://gitlab.com/gerbolyze/gerbonara/

### Screenshots

https://gitlab.com/gerbolyze/gerbonara/-/jobs/2576333109
Error log: [sphinx-err-_uvliwtf.log](https://github.com/sphinx-doc/sphinx/files/8882249/sphinx-err-_uvliwtf.log)

### OS

linux

### Python version

3.10.4 (cpython)

### Sphinx version

5.0.0

### Sphinx extensions

autodoc

### Extra tools

_No response_

### Additional context

_No response_

jaseg:
Add type information for "version" and "project" configuration keys.

### Feature or Bugfix
- Bugfix

### Purpose
This shows a warning if they are set to something other than a str. In both cases, ```sphinx/util/inventory.py``` will throw a ```TypeError```.

This fixes #10544.

### Relates
- #10544

### DIFF
--- a/sphinx/config.py
+++ b/sphinx/config.py
@@ -82,11 +82,11 @@ class Config:
 
     config_values: Dict[str, Tuple] = {
         # general options
-        'project': ('Python', 'env', []),
+        'project': ('Python', 'env', [str]),
         'author': ('unknown', 'env', []),
         'project_copyright': ('', 'html', [str]),
         'copyright': (lambda c: c.project_copyright, 'html', [str]),
-        'version': ('', 'env', []),
+        'version': ('', 'env', [str]),
         'release': ('', 'env', []),
         'today': ('', 'env', []),
         # the real default is locale-dependent


AA-Turner (approved):
Looks good -- in practice becuase of the `re.sub(r"\s+", " ", string)` call, only strings are accepted here.

A

tk0miya:
I found you were already warned by the current implementation. Please take a look at your build log:

```
WARNING: The config value `version' has type `function', defaults to `str'.
```
https://gitlab.com/gerbolyze/gerbonara/-/jobs/2576333109#L33

PR #10558, opened 2022-06-16
Remove `ru_RU` translation, it is empty and duplicates `ru`

mitya57:
Currently sphinx has both `ru` and `ru_RU` translations. It is the same language, so we don't need two different translations for it. The latter was added recently (in 271c8d7bd3cda684), there are no translated strings and no registered translators on Transifex.

This pull request removes duplicate `ru_RU` translation. Please also remove the language on Transifex so the new import does not pull it again.


tk0miya:
Thank you for your comment. Comments from native speakers are very helpful because I can't judge new language request is correct or not.

tk0miya:
Removed ru_RU on transifex.com also. Thanks!

PR #10562, opened 2022-06-16
Remove more deprecated items in Sphinx 6.0

AA-Turner:
Follows on from #10471

### Feature or Bugfix
- Refactoring

A


sphinx/builders/html/__init__.py
```diff
@@ -1215,18 +1215,14 @@ def setup_resource_paths(app: Sphinx, pagename: str, templatename: str,
     pathto = context.get('pathto')
 
     # favicon_url
-    favicon = context.get('favicon')
+    favicon = context.get('favicon_url')
     if favicon and not isurl(favicon):
         context['favicon_url'] = pathto('_static/' + favicon, resource=True)
```
tk0miya: It's better to call this variable as `favicon_url`.
```suggestion
    favicon_url = context.get('favicon_url')
    if favicon_url and not isurl(favicon_url):
        context['favicon_url'] = pathto('_static/' + favicon_url, resource=True)
```

sphinx/builders/html/__init__.py
```diff
@@ -1215,18 +1215,14 @@ def setup_resource_paths(app: Sphinx, pagename: str, templatename: str,
...
-    else:
-        context['favicon_url'] = favicon
 
     # logo_url
-    logo = context.get('logo')
+    logo = context.get('logo_url')
     if logo and not isurl(logo):
         context['logo_url'] = pathto('_static/' + logo, resource=True)
```
tk0miya: This is also.
```suggestion
    logo_url = context.get('logo_url')
    if logo_url and not isurl(logo_url):
        context['logo_url'] = pathto('_static/' + logo_url, resource=True)
```

sphinx/domains/python.py
```diff
@@ -788,9 +784,6 @@ def get_signature_prefix(self, sig: str) -> List[nodes.Node]:
         if 'classmethod' in self.options:
             prefix.append(nodes.Text('classmethod'))
             prefix.append(addnodes.desc_sig_space())
-        if 'property' in self.options:
```
tk0miya, line 791: I noticed we should emit a warning for this usage to users during 5.x.

sphinx/util/compat.py
```diff
@@ -1,31 +1,12 @@
...
-    if 'sphinx.ext.autosummary' in sys.modules:
-        from sphinx.ext import autosummary
-        if hasattr(autosummary, '_objects'):
-            autosummary._objects['_app'] = app  # type: ignore
-        else:
-            autosummary._app = app  # type: ignore
-
-
 def setup(app: "Sphinx") -> Dict[str, Any]:
-    app.connect('builder-inited', register_application_for_autosummary, priority=100)
```
tk0miya, line 27: It seems all compatiblity helpers are removed from this module. Let's remove whole of this module!

tk0miya (approved):
LGTM with nits.

tk0miya:
Could you check the flake8 warnings, please?

sphinx/domains/python.py
AA-Turner: Opened #10603 to track, we should add the warning in 5.0.3 or as early as possible otherwise.

A

PR #10565, opened 2022-06-17
Fix parsing of options with enabled option_emphasise_placeholders

marxin:
If the option is enabled, the current code wrongly uses

```python
if not args:
  continue
```

which skips adjustement of signode and firstname
at the very of the loop.

Fixes: #10366.


AA-Turner:
Can you add a test to ensure this behaviour doesn't regress?

A

marxin:
> Can you add a test to ensure this behaviour doesn't regress?

Sure, I've just done that.

tk0miya:
It seems conflicted. Could you check it please?

marxin:
> It seems conflicted. Could you check it please?

Sure, the conflicts are resolved now.

PR #10566, opened 2022-06-17
Fix usage of enable_search_shortcuts theme config value.

marxin:
The option value is not properly propagated to Java script.

I've just tested that locally with value unset in config and set to `False` and `True` in `html_theme_options`.


AA-Turner:
Can you add a test to ensure this behaviour doesn't regress?

A

marxin:
> Can you add a test to ensure this behaviour doesn't regress?

Sure, the tests are added.

tk0miya:
LGTM. Thanks!

PR #10569, opened 2022-06-18
Improve static typing strictness

danieleades:
fixes some 'strict' type errors which are hidden by the current mypy config

### DIFF
--- a/sphinx/builders/changes.py
+++ b/sphinx/builders/changes.py
@@ -127,7 +127,7 @@ class ChangesBuilder(Builder):
             with open(targetfn, 'w', encoding='utf-8') as f:
                 text = ''.join(hl(i + 1, line) for (i, line) in enumerate(lines))
                 ctx = {
-                    'filename': self.env.doc2path(docname, None),
+                    'filename': self.env.doc2path(docname, False),
                     'text': text
                 }
                 f.write(self.templates.render('changes/rstsource.html', ctx))
--- a/sphinx/builders/latex/theming.py
+++ b/sphinx/builders/latex/theming.py
@@ -113,12 +113,12 @@ class ThemeFactory:
         if name in self.themes:
             theme = self.themes[name]
         else:
-            theme = self.find_user_theme(name)
+            theme = self.find_theme(name)
 
         theme.update(self.config)
         return theme
 
-    def find_user_theme(self, name: str) -> Theme:
+    def find_theme(self, name: str) -> Theme:
         """Find a theme named as *name* from latex_theme_path."""
         for theme_path in self.theme_paths:
             config_path = path.join(theme_path, name, 'theme.conf')
--- a/sphinx/builders/latex/theming.py
+++ b/sphinx/builders/latex/theming.py
@@ -114,8 +114,6 @@ class ThemeFactory:
             theme = self.themes[name]
         else:
             theme = self.find_user_theme(name)
-            if not theme:
-                theme = Theme(name)
 
         theme.update(self.config)
         return theme
@@ -130,4 +128,4 @@ class ThemeFactory:
                 except ThemeError as exc:
                     logger.warning(exc)
 
-        return None
+        return Theme(name)
--- a/sphinx/builders/latex/theming.py
+++ b/sphinx/builders/latex/theming.py
@@ -113,12 +113,14 @@ class ThemeFactory:
         if name in self.themes:
             theme = self.themes[name]
         else:
-            theme = self.find_theme(name)
+            theme = self.find_user_theme(name)
+            if not theme:
+                theme = Theme(name)
 
         theme.update(self.config)
         return theme
 
-    def find_theme(self, name: str) -> Theme:
+    def find_user_theme(self, name: str) -> Theme:
         """Find a theme named as *name* from latex_theme_path."""
         for theme_path in self.theme_paths:
             config_path = path.join(theme_path, name, 'theme.conf')
@@ -128,4 +130,4 @@ class ThemeFactory:
                 except ThemeError as exc:
                     logger.warning(exc)
 
-        return Theme(name)
+        return None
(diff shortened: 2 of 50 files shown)


sphinx/parsers.py
```diff
@@ -84,7 +84,7 @@ def decorate(self, content: StringList) -> None:
         append_epilog(content, self.config.rst_epilog)
 
 
-def setup(app: "Sphinx") -> Dict[str, Any]:
+def setup(app: Sphinx) -> Dict[str, Any]:
```
tk0miya: Is this okay? Sphinx class is not imported on runtime.

sphinx/util/compat.py
```diff
@@ -7,7 +7,7 @@
     from sphinx.application import Sphinx
 
 
-def register_application_for_autosummary(app: "Sphinx") -> None:
+def register_application_for_autosummary(app: Sphinx) -> None:
```
tk0miya: This is also not imported.

sphinx/util/docutils.py
```diff
@@ -20,6 +21,7 @@
 from docutils.utils import Reporter, unescape
 
 from sphinx.deprecation import RemovedInSphinx70Warning, deprecated_alias
+from sphinx.environment import BuildEnvironment
```
tk0miya: This will cause recursion imports.

sphinx/util/docutils.py
```diff
@@ -239,7 +240,9 @@ def __init__(self, env: "BuildEnvironment") -> None:
         self.env = env
         super().__init__()
 
-    def lookup_domain_element(self, type: str, name: str) -> Any:
+    def lookup_domain_element(
+        self, type: str, name: str
+    ) -> Tuple[Type[Directive], List[system_message]]:
```
tk0miya: The first item of the return value is a directive or role object.

danieleades:
there's still plenty of 'strict' errors remaining, but i think this is sufficient improvement to wrap up the PR. I'll look at making more targeted changes in future PRs, and *hopefully* update the configuration to start applying strict checking on subsets of the codebase.

sphinx/builders/latex/theming.py
```diff
@@ -130,4 +128,4 @@ def find_user_theme(self, name: str) -> Theme:
                 except ThemeError as exc:
                     logger.warning(exc)
 
-        return None
+        return Theme(name)
```
tk0miya: This is not a good change. `Theme(name)` is not a user theme.

sphinx/builders/linkcheck.py
```diff
@@ -276,7 +276,7 @@ def get_request_headers() -> Dict:
 
             for u in candidates:
                 if u in self.config.linkcheck_request_headers:
-                    headers = dict(DEFAULT_REQUEST_HEADERS)
```
tk0miya, line 279: Originally, this line creates a copy of the `DEFAULT_REQUEST_HEADERS`. Could you create a copy, please?

sphinx/ext/doctest.py
```diff
@@ -362,11 +362,11 @@ def get_filename_for_node(self, node: Node, docname: str) -> str:
             filename = relpath(node.source, self.env.srcdir)\
                 .rsplit(':docstring of ', maxsplit=1)[0]
         except Exception:
-            filename = self.env.doc2path(docname, base=None)
+            filename = self.env.doc2path(docname, base=False)
```
tk0miya: The `base` keyword is no longer needed.
```suggestion
            filename = self.env.doc2path(docname, False)
```

tk0miya (commented):
LGTM with nits.

sphinx/builders/linkcheck.py
danieleades: done

sphinx/ext/doctest.py
danieleades: done

sphinx/builders/latex/theming.py
danieleades: does changing the function method from 'find_user_theme' to 'find_theme' address your concerns?

i note that the function is not used anywhere else in the code base than this module.

tk0miya: I don't think so. The purpose of `ThemeFactory#get` is "find_theme". It tries to find a theme from built-in ones. After that, it tries to find a user theme via `find_user_theme`. Finally, it creates an empty Theme object from given name. So this is not "find_theme".

danieleades: What would you like it to be called?

tk0miya: I prefer keeping as is without refactoring (it returns None if not found).

sphinx/builders/changes.py
```diff
@@ -127,7 +127,7 @@ def hl(no: int, line: str) -> str:
             with open(targetfn, 'w', encoding='utf-8') as f:
                 text = ''.join(hl(i + 1, line) for (i, line) in enumerate(lines))
                 ctx = {
-                    'filename': self.env.doc2path(docname, None),
+                    'filename': self.env.doc2path(docname, False),
```
AA-Turner, line 130: Why is the type for `filename` `str | Literal[False]`? `str | None` would make more sense, or just the empty string here if we're wanting a falsy value.

danieleades: i'm not sure I understand the question.

> Why is the type for `filename` `str | Literal[False]`

where are you getting that from? The type for `filename` is `str`, since this is what `doc2path` returns.

Are you referring to the args for `doc2path`? the `base`/`basedir` argument is a boolean flag, not a path to the base directory

AA-Turner: My mistake, thanks. I got mixed up and was thinking of the second argument as the default/fallback.

A

tk0miya (approved):
LGTM.
I'm okay to merge this if @AA-Turner is also okay.

danieleades:
sounds good! it'll need a 'squash' merge, or I can do that myself this evening. whichever you prefer.

AA-Turner (approved):

PR #10571, opened 2022-06-18
Reduce content in the `sphinx-quickstart` `conf.py` file

### LINKED ISSUE #10056: Potential improvements to sphinx-quickstart's default `conf.py_t` template

Sphinx's conf.py template duplicates contents from Sphinx's documentation, which results in redundancy and does not provide any context for potential additional configuration values. The documentation comments also add additional length to the configuration file; making it seem more complex than it actually is.

More importantly, users do not add such comments for configuration values they add or modify, which means that the configuration file ends up with a mix of values-with-comments-describing-them and those without. This makes the configuration file difficult to navigate and to locate the relevant configuration within it.

It also allows for inconsistency between the documentation and the conf.py template to creep in unnoticed. Right now, for example, `language` is in the "General configuration" section in the conf.py template. However, it is not under the "General configuration" header in the documentation: https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration. Instead, it is under the "Options for internationalization" section: https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-internationalization. I wouldn't be surprised if some of the comments in the conf.py file are outdated or missing information that's included in the corresponding documentation page.

- Drop the comments from the configuration file template, that duplicate the documentation for each of the options or provide example usage of the options.
- Add links to each of the "sections" in the configuration file, that correspond to the relevant documentation section.

This has quite a few benefits IMO:

- it helps improve discoverability of not-in-the-file configuration options.
- it makes it easier to get to the "proper" documentation of the options.
- it makes the configuration file more maintainable, by reducing redundancy added by the comments.
- it makes the configuration file less "complicated" [^1].
- it makes it easier to keep the configuration file in sync with the documentation (`language` example above).
- it clearly communicates to documentation authors how to maintain this file (make a section and put the documentation URL in the next line, or add to the existing section).
- it nudges extension authors to publish similar configuration listing pages, since that's an expectation set by the overall style of the configuration file.

- Maintaining status quo.

  This is the path of least effort. I don't think this is a good idea though, because we will continue to push for the mixed-comment + seeming-more-complicated-than-it-actually-is configuration files.

- Not including the URLs, only dropping comments

  This would reduce the percieved complexity of the file, as well as the length of it. I think it's still a worthwhile improvement, but we'd lose the discoverability benefits of having a URL immediately accessible.

**Additional context**

Example configuration file, that incorporates all the suggestions: https://github.com/pradyunsg/installer/blob/48f44b87eb438bc32ecad27c29b5d90fa03f2a37/docs/conf.py
Example configuration file, that drops the comments and does NOT include URLs: https://github.com/pypa/pip/blob/7f8a6844037fb7255cfd0d34ff8e8cf44f2598d4/docs/html/conf.py

FWIW, I think having a "Options for internationalization" section in the conf.py file  when `language` is set, would help make it clearer that the Sphinx supports it out-of-the-box.

[^1]: In addition to the reduction of length, I think the comments are usually a subtle indicator that there's some sort of complexity/subtly -- they usually serve to provide context that might not be obvious to the reader. Most options in Sphinx's configuration file are not subtle in that way, and even if they are, the details of *how* they interact in the context of a single documentation set should be more important than the fact that the verbatim information about how they behave (i.e. let the documentation authors add comments about the things they care about).

pradyunsg:
Subject: (see pull request title)

### Feature or Bugfix

- Feature
- ~Bugfix~
- ~Refactoring~

### Purpose

Simplify the sphinx-quickstart `conf.py` file, reducing the percieved complexity of the `conf.py` file generated by default.

### Detail

- Dropping comments about configuration fields
- Dropping commented code generation
- Ensure the file always has a trailing newline
- Consistent newline separation of various sections
- Include URLs to the relevant documentation pages listing all configuration variables

### Relates

Closes #10056


pradyunsg:
For comparing the difference in the output files generated by this, here's an example based on running:

```
$ sphinx-quickstart /tmp/dev1 -p test -l en -a test --sep -r 1.0 -v 1.0 --ext-intersphinx --ext-todo
$ cat /tmp/dev1/source/conf.py
```

Before this PR:

<details>

```py
#
# This file only contains a selection of the most common options. For a full
# list see the documentation:
# https://www.sphinx-doc.org/en/master/usage/configuration.html

# -- Path setup --------------------------------------------------------------

#
# import os
# import sys


# -- Project information -----------------------------------------------------

project = 'test'
copyright = '2022, test'
author = 'test'

# The short X.Y version
version = '1.0'

# The full version, including alpha/beta/rc tags
release = '1.0'


# -- General configuration ---------------------------------------------------

# ones.
extensions = [
    'sphinx.ext.intersphinx',
    'sphinx.ext.todo',
]

# The language for content autogenerated by Sphinx. Refer to documentation
# for a list of supported languages.
#
# This is also used if you do content translation via gettext catalogs.
# Usually you set "language" from the command line for these cases.
language = 'en'

# This pattern also affects html_static_path and html_extra_path.
exclude_patterns = []


# -- Options for HTML output -------------------------------------------------

# a list of builtin themes.
#
html_theme = 'alabaster'

# -- Extension configuration -------------------------------------------------

# -- Options for intersphinx extension ---------------------------------------

# Example configuration for intersphinx: refer to the Python standard library.
intersphinx_mapping = {
    'python': ('https://docs.python.org/3', None),
}

# -- Options for todo extension ----------------------------------------------

# If true, `todo` and `todoList` produce output, else they produce nothing.
todo_include_todos = True
```

(no trailing newline)

</details>

After this PR:

<details>

```py
#
# For the full list of built-in configuration values, see the documentation:
# https://www.sphinx-doc.org/en/master/usage/configuration.html

# -- Project information -----------------------------------------------------
# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information

project = 'test'
copyright = '2022, test'
author = 'test'

version = '1.0'
release = '1.0'

# -- General configuration ---------------------------------------------------
# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration

extensions = [
    'sphinx.ext.intersphinx',
    'sphinx.ext.todo',
]

exclude_patterns = []

language = 'en'

# -- Options for HTML output -------------------------------------------------
# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output

html_theme = 'alabaster'

# -- Options for intersphinx extension ---------------------------------------
# https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#configuration

intersphinx_mapping = {
    'python': ('https://docs.python.org/3', None),
}

# -- Options for todo extension ----------------------------------------------
# https://www.sphinx-doc.org/en/master/usage/extensions/todo.html#configuration

todo_include_todos = True
```

</details>

sphinx/templates/quickstart/conf.py_t
```diff
@@ -1,119 +1,71 @@
...
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-#
-{% if append_syspath -%}
 import os
 import sys
 sys.path.insert(0, {{ module_path | repr }})
-{% else -%}
-# import os
-# import sys
-{% if module_path -%}
```
tk0miya, line 20: Note: Both`append_syspath` and `module_path` are available on apidoc. And quickstart does not passes both of them. So this if-condition is not needed.

tk0miya (approved):
LGTM

humitos (approved):
💯

sphinx/templates/quickstart/conf.py_t
```diff
@@ -1,119 +1,71 @@
...
-# Add any Sphinx extension module names here, as strings. They can be
-# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
-# ones.
-extensions = [
+extensions = [{% if extensions %}
 {%- for ext in extensions %}
     '{{ ext }}',
 {%- endfor %}
-]
+{% endif %}]
```
AA-Turner, line 36: Not sure I understand the if condition here, what's the difference between it and looping over zero extensions?

A

pradyunsg: Whitespace. :)

pradyunsg: The code currently generates:

```
extensions = []
```

And

```
extensions = [
    'sphinx.ext.intersphinx',
]
```

This is consistent with how empty lists are usually written in Python, by humans.

pradyunsg:
@AA-Turner Added the CHANGES entry. :)

CHANGES
```diff
@@ -20,6 +20,8 @@ Features added
   directives through a new ``option_emphasise_placeholders`` configuration option.
 * #10439: std domain: Use the repr of some variables when displaying warnings,
   making whitespace issues easier to identify.
+* #10571: Reduce content in the ``sphinx-quickstart`` ``conf.py`` file. Patch by
+  Pradyun Gedam.
```
```suggestion
```suggestion
* #10571: quickstart: Reduce content in the generated ``conf.py`` file. Patch by
  Pradyun Gedam.
```

CHANGES
```diff
@@ -20,6 +20,9 @@ Features added
   directives through a new ``option_emphasise_placeholders`` configuration option.
 * #10439: std domain: Use the repr of some variables when displaying warnings,
   making whitespace issues easier to identify.
+```suggestion
```
```suggestion
```

AA-Turner:
Thanks @pradyunsg!

A

PR #10576, opened 2022-06-18
Fixed path bug with links in SVG inheritance diagrams

ayshih:
Subject: Fixes links in SVG inheritance diagrams when those diagrams are in documents not at the base level

### Feature or Bugfix

- Bugfix

### Purpose
Currently, links in SVG inheritance diagrams only work when the documents are at the base level.  That's because the URIs supplied to the diagram are relative to the document (as is appropriate for PNG output), but the SVGs live in the `/_images` folder, so the URIs need to be relative to that location.  There is a hard-coded `'../'` that gets prepended to the URIs, which fixes only base-level documents.  Documents in sub-directories result in SVG links that have one or more extraneous `'../'` that break the links.

### Detail
This PR fixes the URIs to be relative to the image directory prior to making the SVG inheritance diagrams.

### Relates
Fixes #10570

### DIFF
--- a/sphinx/ext/inheritance_diagram.py
+++ b/sphinx/ext/inheritance_diagram.py
@@ -30,6 +30,7 @@ LaTeX.
 
 import builtins
 import inspect
+import pathlib
 import re
 from importlib import import_module
 from typing import Any, Dict, Iterable, List, Tuple, cast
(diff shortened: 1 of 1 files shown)


ayshih:
Note that this PR fixes a bug in Sphinx that is unrelated to any use of third-party extensions (e.g., `sphinx_automodapi`).

Non-exhaustive list of issues that this PR closes: #3176, sunpy/sunpy#2097, astropy/astropy#4935, probabilistic-numerics/probnum#459

Related issues:
- matplotlib/matplotlib#17913: matplotlib is choosing to use PNG inheritance diagrams because the links in SVG inheritance diagrams are broken
- #2484: Astropy's inheritance diagrams are all in subdirectories, so that's why removing the prepended `'../'` fixed all of their inheritance diagrams.  It was erroneously concluded that this was not a bug in Sphinx.
- astropy/photutils#345: The significance of why moving the inheritance diagram from a subdirectory to the base level fixed the SVG links was not recognized at the time.

ayshih:
Closing this PR since #10614 is more encompassing

PR #10577, opened 2022-06-18
LaTeX: fix #8686 Text can fall out of code-block at end of page

### LINKED ISSUE #8686: [LaTeX] Text can fall out of code-block at end of page and leave artifact on next page
**Describe the bug**

When a code block is at the end of a page and the contents makes the block larger than can fit on the page, it is broken into two sections. In particular when the last line was wrapped, the content spills out of the code block. The code block on the next page remains empty.

**To Reproduce**

1. Start a new Sphinx project with default settings (I used version 3.4.0) and use the attached [index.txt](https://github.com/sphinx-doc/sphinx/files/5803257/index.txt) (rename to rst because GitHub does not recognise rst extension).
2. run `make latexpdf`
3. view output

**Expected behavior**
The code block is split across two pages, and the contents of the code block directive is spread between the two pages.

**Actual behavior**
Screenshot 2021-01-12 at 16 57 36

**Environment info**
- OS: macOS 11.1
- Python version: 3.8.2
- Sphinx version: 3.4.0
- Sphinx extensions: none
- Extra tools: MacTeX 2020

jfbu:
This is short but deep hack into `fancyvrb` package which is used for legacy reasons by Sphinx LaTeX. It fixes #8686 issue in a way compatible with emphasized lines and line numbers.  However it currently has breaking change that the line number is repeated identically for each piece of a wrapped line, whereas until now it is indicated only once, at the first line coming out from wrapped one.

edit: the breaking changed about code line numbers has been removed in a later commit. And WIP status was thus removed.

<strike>Keeping this as WIP for the moment as I need to do some more testing. And I am hesitating if the PR should be on master and not 5.x.</strike>

At the LaTeX level, I have made some internal name and interface changes for reasons of coherence but I feel it is very unlikely anyone actually modified or used the internal macros, so I don't consider this a breaking change.

Example of output (from the mwe at #8686):

(this was before removal of repeated code line number)

Capture d’écran 2022-06-18 à 23 30 44

On sees the number 2 is repeated all along. <s>I can add some extras to not repeat but this means some complications.</s> (update: done)

I think it is very unlikely fancyvrb (in view of its recent changes over the last years) will ever change the core stuff we hacked into, so it seems reasonably safe.  Nevertheless it is deep hack and things work in a delicate equilibrium with fancvyvrb internal dealings. 

### Relates
- #8686


sphinx/texinputs/sphinxlatexliterals.sty
```diff
@@ -50,8 +50,8 @@
 % 4- while still allowing expansion of Pygments latex mark-up
 % Other aspects such as framing, caption handling, codeline wrapping are
 % added on top of it. We should stop using fancyvrb and implement
-% 1, 2, 3, 4 by own Sphinx fully native Verbatim. This would allow to solve
-% limitations with wrapped long code line not allowing page break.
+% 1, 2, 3, 4 by own Sphinx fully native Verbatim. This would greatly simply
```
jfbu: s/simply/simplify/

jfbu:
Ready for review.  In case of line wrapping the code line number is printed only once as is currently case.  So this fixes #8686 with no other change to output.

Example output:
Capture d’écran 2022-06-19 à 10 02 23

From this project 
[8686_codeblock.zip](https://github.com/sphinx-doc/sphinx/files/8934960/8686_codeblock.zip)

jfbu:
The squashed commit message was written via the GitHub interface and is not hard-wrapped :-(

jfbu:
due to limitations of LaTeX `\colorbox/\fcolorbox` this required follow-up additions, which were concluded by merge of #10648 for 5.1.0 release

PR #10523, opened 2022-06-04
Fix double brackets

AA-Turner:
Fixes #10522.

### Feature or Bugfix
- Bugfix

A

PR #10595, opened 2022-06-25
Fix #10594: HTML: field term colons are doubled if using Docutils 0.18+

### LINKED ISSUE #10594: Sphinx 5.0+ prints double colons for roles
### Describe the bug

For examle:
:1: Foo
is renderend as
**1:** Foo
with Sphinx 4.5
https://doc.wikimedia.org/pywikibot/stable/library_usage.html

whereas Sphinx 5.0+ renders
**1::** Foo

https://doc.wikimedia.org/pywikibot/master/library_usage.html

### How to Reproduce

Other examples with variables and parameters
Sphinx 4.5 (single quotes after Variables, Parameters, Return type) like: 
**Variables: summary_key** - Edit summary message key
https://doc.wikimedia.org/pywikibot/stable/scripts_ref/scripts.html#module-scripts.basic

Sphinx 5.02  (doible quotes after Variables, Parameters, Return type) like:
**Variables:: summary_key** - Edit summary message key
https://doc.wikimedia.org/pywikibot/master/scripts_ref/scripts.html#module-scripts.basic

### Expected behavior

Single quotes are expected as in Sphinx 4.5

### Your project

See "Show Source" of the given links above.

### Screenshots

Sphinx 4.5
image

Sphinx 5.0.2
image

### OS

Linux

### Python version

Python 3.7

### Sphinx version

4.5.0 vs. 5.0.2

### Sphinx extensions

'sphinx.ext.autodoc', 'sphinx.ext.autosectionlabel', 'sphinx.ext.extlinks',     'sphinx.ext.todo',     'sphinx.ext.viewcode',     'sphinx.ext.autosummary',     'sphinx.ext.napoleon',

### Extra tools

_No response_

### Additional context

_No response_

jfbu:
### Relates
- #10523 

The Docutils commit responsible for the change at 0.18 is [rev8734](https://sourceforge.net/p/docutils/code/8734/)

Nota bene:
- our test suite did not detect the added `<span class="colon">:</span>` at Docutils 0.18. As I am not competent enough to add a test conditional on Docutils version I did not (remember I am only LaTeX guru here).
- I guess the `docinfo` class for `<dl>` is not involved at Sphinx level

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,8 @@ Features added
 Bugs fixed
 ----------
 
+#10594: HTML: field term colons are doubled if using Docutils 0.18+
+
 Testing
 --------
 
--- a/CHANGES
+++ b/CHANGES
@@ -16,7 +16,7 @@ Features added
 Bugs fixed
 ----------
 
-#10594: HTML Theme: field term colons are doubled if using Docutils 0.18+
+* #10594: HTML Theme: field term colons are doubled if using Docutils 0.18+
 
 Testing
 --------
--- a/CHANGES
+++ b/CHANGES
@@ -16,7 +16,7 @@ Features added
 Bugs fixed
 ----------
 
-#10594: HTML: field term colons are doubled if using Docutils 0.18+
+#10594: HTML Theme: field term colons are doubled if using Docutils 0.18+
 
 Testing
 --------
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -686,15 +686,11 @@ dl.field-list > dt {
     padding-right: 5px;
 }
 
-/* Compatibility patch for Docutils 0.18+ */
-dl.field-list > dt .colon {
-    display: none;
-}
-
-/* Docutils 0.17 and older */
+{%- if docutils_version_info[:2] < (0, 18) %}
 dl.field-list > dt:after {
     content: ":";
 }
+{% endif %}
 
 dl.field-list > dd {
     padding-left: 0.5em;
--- a/sphinx/themes/basic/static/basic.css_t
+++ b/sphinx/themes/basic/static/basic.css_t
@@ -686,6 +686,12 @@ dl.field-list > dt {
     padding-right: 5px;
 }
 
+/* Compatibility patch for Docutils 0.18+ */
+dl.field-list > dt .colon {
+    display: none;
+}
+
+/* Docutils 0.17 and older */
 dl.field-list > dt:after {
     content: ":";
 }


sphinx/themes/basic/static/basic.css_t
```diff
@@ -686,6 +686,12 @@ dl.field-list > dt {
...
 
+/* Compatibility patch for Docutils 0.18+ */
+dl.field-list > dt .colon {
+    display: none;
+}
+
+/* Docutils 0.17 and older */
 dl.field-list > dt:after {
     content: ":";
 }
```
AA-Turner: This only inserts the CSS for Docutils 0.17.1 and older:

```suggestion
{%- if docutils_version_info[:2] < (0, 18) %}
dl.field-list > dt:after {
    content: ":";
}
{% endif %}
```

CHANGES
```diff
@@ -16,6 +16,8 @@ Features added
 Bugs fixed
 ----------
 
+#10594: HTML: field term colons are doubled if using Docutils 0.18+
```
```suggestion
#10594: HTML Theme: field term colons are doubled if using Docutils 0.18+
```

AA-Turner (changes requested):

AA-Turner:
Thanks @jfbu! I suggested a couple of changes.

Testing HTML output is a much bigger challenge in general, as it pretty much needs to be a visual comparison -- layered CSS files etc. can affect output with identical HTML content.

A

jfbu:
> Thanks @jfbu! I suggested a couple of changes.

ok thanks, applied.

I see now that you had been defining and using `docutils_version_info` elsewhere in `basic.css_t` already.  And that we require Docutils `>= 0.14` which added `docutils.__version_info__`.

About the change to CHANGES, I add noticed you used `HTML Theme` however I found this terminology a bit unclear.
Your suggestion applied anyhow.

> 
> Testing HTML output is a much bigger challenge in general, as it pretty much needs to be a visual comparison -- layered CSS files etc. can affect output with identical HTML content.
> 


Yes, understood, however I was surprised that you we did not have a check which would have signaled html output (I mean the html files, not how they are rendered, and not even the CSS files) got modified by the upstream Docutils 0.18 changes.

Visual comparison is imho simply not feasible and will always let things pass through.  I apply it from time to time on our own docs `sphinx.pdf` but very superficially, trying more to detect changes in LaTeX compilation log behavior.

AA-Turner:
> About the change to CHANGES, I had noticed you used HTML Theme however I found this terminology a bit unclear. Your suggestion applied anyhow.

I just use it as it seems to be the precedent for CSS changes etc.

> Yes, understood, however I was surprised that we (I initially wrote "you" by mistake ;-) ) did not have a check which would have signaled html output (I mean the html files, not how they are rendered, and not even the CSS files) got modified by the upstream Docutils 0.18 changes.

Hmm, perhaps we can add such a thing. It is difficult to construct a short test file that hits all the various permutations of reST syntax, though.

A

PR #10599, opened 2022-06-25
Use Docutils 0.19 footnote styles

AA-Turner:
This backports the Docutils 0.19 footnote style DOM to Sphinx 5.1.

It is too big to go into 5.0.3 really, I think.

@pradyunsg is this a useful thing from a theme perspective?

### Feature or Bugfix
- Feature
- Bugfix

A


sphinx/util/docutils.py
```diff
@@ -182,10 +183,41 @@ def using_user_docutils_conf(confdir: Optional[str]) -> Generator[None, None, No
             os.environ['DOCUTILSCONFIG'] = docutilsconfig
 
 
+@contextmanager
+def du19_footnotes() -> Generator[None, None, None]:
+    def visit_footnote(self, node):
+        label_style = self.settings.footnote_references
+        if not isinstance(node.previous_sibling(), type(node)):
+            self.body.append(f'<aside class="footnote-list {label_style}">\n')
+        self.body.append(self.starttag(node, 'aside',
```
tk0miya, line 192: Is it okay to output `<aside>` tag forcedly? For example, sphinx_rtd_theme does not support docutils-0.18. And it expects not to use `<aside>` for footnotes. So this will break renderings of the theme.
https://github.com/readthedocs/sphinx_rtd_theme/blob/master/setup.py#L122

AA-Turner: Ahh, yes this should be also gated on DU 18 or 18.1.

A

tk0miya (approved):

sphinx/util/docutils.py
```diff
@@ -182,10 +183,46 @@ def using_user_docutils_conf(confdir: Optional[str]) -> Generator[None, None, No
             os.environ['DOCUTILSCONFIG'] = docutilsconfig
 
 
+@contextmanager
+def du19_footnotes() -> Generator[None, None, None]:
```
tk0miya, line 187: Is there any reason to patch docutils instead of overriding methods in `sphinx/writers/html5.py`? I think overriding is a bit simpler than patching. Either is okay to me.

AA-Turner: Makes it easier to remove the patch when we drop support for DU 18 was my thought.

A

tk0miya: Okay, let's go with this.

PR #10612, opened 2022-06-28
Remove redundant static typing casts

danieleades:
remove redundant `typing.cast`s and updates mypy config to prevent these in future.

type casts should only be used as a last resort, as they can mask errors if refactoring later changes the type of an object.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,6 +32,7 @@ follow_imports = skip
 check_untyped_defs = True
 warn_unused_ignores = True
 strict_optional = False
+no_implicit_optional = True
 warn_redundant_casts = True
 
 [tool:pytest]
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ follow_imports = skip
 check_untyped_defs = True
 warn_unused_ignores = True
 strict_optional = False
-no_implicit_optional = True
+warn_redundant_casts = True
 
 [tool:pytest]
 filterwarnings =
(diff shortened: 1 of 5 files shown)


danieleades:
i don't believe the CI failure is to do with this PR

setup.cfg
```diff
@@ -32,7 +32,7 @@ follow_imports = skip
 check_untyped_defs = True
 warn_unused_ignores = True
 strict_optional = False
-no_implicit_optional = True
```
tk0miya, line 35: Why did you remove this line?

sphinx/domains/c.py
```diff
@@ -3552,7 +3552,6 @@ def _render_symbol(self, s: Symbol, maxdepth: int, skipThis: bool,
 
     def apply(self, **kwargs: Any) -> None:
         for node in self.document.findall(AliasNode):
-            node = cast(AliasNode, node)
```
tk0miya, line 3555: Good point. Our stubs surely provide nice annotation for `findall()`!
https://github.com/tk0miya/docutils-stubs/blob/b14071d065f8ee35399856fcaff6cd608294a342/docutils-stubs/nodes.pyi#L32-L33

tk0miya (commented):
LGTM with nits.

tk0miya:
I manually re-run failed tests :-)

setup.cfg
```diff
@@ -32,7 +32,7 @@ follow_imports = skip
...
+warn_redundant_casts = True
```
AA-Turner, line 36: Please discuss the config change in a different PR/issue.
```suggestion
no_implicit_optional = True
warn_redundant_casts = True
```

AA-Turner (approved):

AA-Turner:
Thanks Dan!

A

Issue #10615, opened 2022-06-29
LaTeX: wrapped long code line has acquired color problems with current 5.x branch if it is emphasized


jfbu:
### Describe the bug

A wrapped long code line will have color issues it if is also emphasized.  Some of input may even acquire the background color hence become invisible.

### How to Reproduce

```
.. code-block:: python
   :linenos:
   :emphasize-lines: 1,3
   
   print("a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line has problems with colors; moreover here the closing parenthesis has same color as background hence is invisible")
   print("a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line with no problems if not emphasized")
   print("syntax highlighting is ok" + " in this short line"); A = B
   print("syntax highlighting is ok" + " in this short line"); A = B

ABC

```

### Your project

N/A

### Screenshots

![Capture d’écran 2022-06-29 à 18 17 46](https://user-images.githubusercontent.com/2589111/176486968-a161cbb2-e967-4580-80c6-fcb2c2831339.png)

### OS

Mac

### Python version

3.10.4

### Sphinx version

This is only for current 5.x after merge of #10577 (40841ae80cc6b81024760201a33a2828414c57f7).

Issue #10610, opened 2022-06-27
LaTeX: problems (frame, color) when long code line wraps over a page limit in some circumstances


jfbu:
### Describe the bug

Unexpected loss of frame on page 1 and color on page 2

### How to Reproduce

```
.. raw:: latex

   \vspace*{18cm}

1. A list

   #) one

   #) two

   #) .. code-block:: python

         astr = "a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line"

```

### Your project

N/A

### Screenshots

![Capture d’écran 2022-06-28 à 00 00 06](https://user-images.githubusercontent.com/2589111/176043472-9fa1ce5b-4883-4633-9d42-37e405b966f1.png)

### OS

Mac

### Python version

3.10.4

### Sphinx version

5.x (v5.1.0+/12e86ff0e)

### Additional context

Relates #8686.

The #10577 works generally, but here something strange happens with a code-block as item in a numbered list.

edit: 

- the issue is actually not at all related to lists, it happens with code-blocks at top level
- it has to do with syntax highlighting,  the effect disappears with "none" as language.

Bug clearly related to the long code line being almost entirely an argument to a `\PYG` macro.  Some problems with "groups" in TeX modifying scopes of certain things, presumably.

edit:

This has to do with Pygmentize mark-up assigning some color (not background color, which would create unbreakable box, here only foreground color) via `\textcolor{<color>}{<long input>}` to a long input which ends up split across pages (prior to #10577 the text would have spilled into bottom margin of first page).  I have noticed that already LaTeX package `framed` has problems in such circumstances (the framing is partially contaminated by the text color, but text on second page has correct color).  Inserting into a Sphinx LaTeX document
```latex
\begin{sphinxVerbatim}[commandchars=\\\{\}]
{{\color{blue} a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line}}
[...]

PR #10616, opened 2022-06-29
LaTeX: partially fix color issues for wrapped code lines with pagebreak

jfbu:
This does not fix entirely #10610 but it does sufficiently for it not to
require reverting #10577 which tried to solve #8686 conundrum.  In
extreme cases, the #8686 problem meant that some contents disappeared
at page bottom, so it is probably better to maintain #10577 which
will avoid any such overflow of code beyond its frame, even though in
some specific cases (a colored entity such as a long string is partly on
both pages), some syntax highlighting gets lost.

There are anyhow other issues with colors for wrapped code lines, even
with no pagebreaks involved, such as #10615.  This patch does not change
the situation there.

Prior to this patch (but after #10577):
Capture d’écran 2022-06-29 à 18 42 36

With this patch
Capture d’écran 2022-06-29 à 18 41 36

This patch does some change to private Sphinx LaTeX macro `\spx@fcolorbox` and `\spx@CustomFBox`.

There is no CHANGES entry because it fixes issues with #10577 which has been merged but not released yet.

test done with
```rest
Test
----


.. raw:: latex

   \vspace*{15cm}%\showoutput

.. code-block:: python

   A = [x for x in range(10)]
   print("a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line a very long line")
   A = [x for x in range(10)]


ABC
```

### DIFF
--- a/sphinx/texinputs/sphinxlatexliterals.sty
+++ b/sphinx/texinputs/sphinxlatexliterals.sty
@@ -1,7 +1,7 @@
 %% LITERAL BLOCKS
 %
 % change this info string if making any custom modification
-\ProvidesFile{sphinxlatexliterals.sty}[2022/06/18 code-blocks and parsed literals]
+\ProvidesFile{sphinxlatexliterals.sty}[2022/06/29 code-blocks and parsed literals]
 
 % Provides support for this output mark-up from Sphinx latex writer:
 %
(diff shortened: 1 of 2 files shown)


jfbu:
complete fix came with #10648

PR #10623, opened 2022-07-01
fix a few 'flake8-simplify' lints

danieleades:


AA-Turner:
please just blacklist deprecated modules from the lint rather than changing them (e.g. jsdump).

A

sphinx/cmd/quickstart.py
```diff
@@ -170,10 +170,7 @@ def _has_custom_template(self, template_name: str) -> bool:
               It will be removed in the future without deprecation period.
         """
         template = path.join(self.templatedir, path.basename(template_name))
-        if self.templatedir and path.exists(template):
-            return True
-        else:
-            return False
+        return bool(self.templatedir) and path.exists(template)
```
tk0miya, line 173: Is this `bool` needed?

sphinx/environment/adapters/toctree.py
```diff
@@ -194,12 +194,11 @@ def _entries_from_toctree(toctreenode: addnodes.toctree, parents: List[str],
                     for subtocnode in list(toc.findall(addnodes.toctree)):
                         if not (subtocnode.get('hidden', False) and
                                 not includehidden):
-                            i = subtocnode.parent.index(subtocnode) + 1
-                            for entry in _entries_from_toctree(
+                            start = subtocnode.parent.index(subtocnode) + 1
+                            for i, entry in enumerate(_entries_from_toctree(
                                     subtocnode, [refdoc] + parents,
-                                    subtree=True):
+                                    subtree=True), start):
```
tk0miya, line 203: I feel this is a bit unreadable. How about storing the list before this loop?
```
entries = _entries_from_toctree(...)
for i, entry in enumerate(entries, start):
    ...
```

sphinx/ext/autodoc/__init__.py
```diff
@@ -2350,10 +2350,7 @@ def isslotsattribute(self) -> bool:
         """Check the subject is an attribute in __slots__."""
         try:
             __slots__ = inspect.getslots(self.parent)
-            if __slots__ and self.objpath[-1] in __slots__:
-                return True
-            else:
-                return False
+            return bool(__slots__) and self.objpath[-1] in __slots__
```
tk0miya, line 2351: Is this bool call needed?

sphinx/cmd/quickstart.py
AA-Turner: Yep, as if templatedir is falsy, the check short circuits and would return the value of templatedir, rather than False.

A

sphinx/ext/apidoc.py
```diff
@@ -177,11 +177,7 @@ def is_skipped_package(dirname: str, opts: Any, excludes: List[str] = []) -> boo
         return True
 
     # Check there is some showable module inside package
```
AA-Turner:
```suggestion
    # Check there is some showable module inside package.
    # If all submodules are excluded, return True.
```

sphinx/ext/autodoc/__init__.py
AA-Turner: Yep, see above.

A

sphinx/transforms/post_transforms/images.py
```diff
@@ -231,10 +228,7 @@ def guess_mimetypes(self, node: nodes.image) -> List[str]:
     def handle(self, node: nodes.image) -> None:
         _from, _to = self.get_conversion_rule(node)
 
-        if _from in node['candidates']:
-            srcpath = node['candidates'][_from]
-        else:
-            srcpath = node['candidates']['*']
+        srcpath = node['candidates'].get(_from, node['candidates']['*'])
```
AA-Turner, line 231: This one feels less clean than the others, but I'm not sure of any better refactoring.

A

sphinx/util/jsdump.py
```diff
@@ -54,22 +54,14 @@ def decode_string(s: str) -> str:
...
-char   finally   null   true
-class   float   package   try
-const   for   private   typeof
-continue   function   protected   var
-debugger   goto   public   void
-default   if   return   volatile
-delete   implements   short   while
-do   import   static   with
-double   in   super""".split())
+reswords = {"abstract", "else", "instanceof", "switch", "boolean", "enum", "int",
```
AA-Turner: As per my earlier comment, please revert this one.

A

AA-Turner (approved):
Looks good barring a few comments.

A

AA-Turner:
Oh actually, this PR should also add `flake8-simplify` to the plugin list, right?

A

sphinx/util/jsdump.py
danieleades: is there some way to determine which modules are 'deprecated'?

should i just look for the warnings at the top of the file?

danieleades:
> Oh actually, this PR should also add `flake8-simplify` to the plugin list, right?
> 
> A

that's my intention. right now it's still generating a lot of errors

sphinx/util/jsdump.py
AA-Turner: I think this is the only instance in this PR, in the general deprecations list also contains deprecated modules though.

A

AA-Turner:
> that's my intention. right now it's still generating a lot of errors

OK. Do you want this PR merged, or do you want to hold off? (It is still marked as 'draft', with no indication as to why).

A

danieleades:
> > that's my intention. right now it's still generating a lot of errors
> 
> OK. Do you want this PR merged, or do you want to hold off? (It is still marked as 'draft', with no indication as to why.)
> 
> A

It's marked as a draft because I'm (ab)using the CI to validate the changes I make incrementally. When it's ready for review I'll remove the "draft" status

sphinx/environment/adapters/toctree.py
danieleades: > I feel this is a bit unreadable. How about storing the list before this loop?
> 
> ```
> entries = _entries_from_toctree(...)
> for i, entry in enumerate(entries, start):
>     ...
> ```

this doesn't work, and i cannot for the life of me figure out why...

as you can see, the second last commit passes CI, and the latest commit has only this change and fails.

one of these methods must be modifying it's arguments or something?

sphinx/cmd/quickstart.py
tk0miya: Understand. Thanks,

sphinx/environment/adapters/toctree.py
tk0miya: Sorry, it's my wrong. Okay, let's revert this change and go with your idea.

tk0miya:
How about merging this after CI passed? It's better to merge step by step even if your work still remained. Large and long life PR is difficult to review, merge and keep maintained (from conflicts).

sphinx/environment/adapters/toctree.py
danieleades: i mean, it's a code smell. The fact that this *can't* be refactored in this way tells me that this *should* be refactored.
Not in this PR though.

sphinx/util/inspect.py
```diff
@@ -357,17 +337,19 @@ def isproperty(obj: Any) -> bool:
...
+        isinstance(obj, typing._GenericAlias)  # type: ignore[attr-defined]
+    ) or (
+        # only for py39+
+        hasattr(types, 'GenericAlias') and
+        isinstance(obj, types.GenericAlias)  # type: ignore[attr-defined]
+    ) or (
+        # for py39+
+        hasattr(typing, '_SpecialGenericAlias') and
+        isinstance(obj, typing._SpecialGenericAlias)  # type: ignore[attr-defined]
+    )
```
tk0miya: Is this change really better? Should we not use the `else` clause for True/False decision, right?

danieleades:
i think i'm going to have to close this. I can address these lints in separate PRs

PR #10634, opened 2022-07-04
Make `-P` (pdb) work better with exceptions triggered from events

jbms:
### Feature or Bugfix

- Bugfix

### Purpose
Previously, if an exception was raised from an event listener, and the
`-P` option was specified, the debugger would be started not for the
original error but for the `ExtensionError` wrapping it that was
raised by `EventManager.emit`.  That made it difficult to debug the
error.

With this change, when `-P` is specified, wrapping of errors in
`ExtensionError` is disabled, which allows pdb to debug the original
error.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -30,6 +30,7 @@ Features added
 Bugs fixed
 ----------
 
+* #10634: Make -P (pdb) option work better with exceptions triggered from events
 * #10031: py domain: Fix spurious whitespace in unparsing various operators (``+``,
   ``-``, ``~``, and ``**``). Patch by Adam Turner.
 * #10460: logging: Always show node source locations as absolute paths.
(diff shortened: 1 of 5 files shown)


AA-Turner:
+1 on the idea.

Is it possible to implement this without adding a new attribute to `Sphinx`? The application object already has a really big API.

This would also need a CHANGES entry.

A

jbms:
I think it is difficult to avoid making it a parameter to `Sphinx.__init__`, since some events are emitted by `Sphinx.__init__` so it would be too late to enable it later.  In principle it could be passed as a separate parameter to `EventManager.__init__` by `Sphinx.__init__`, since currently the variable is only used by `EventManager`, though I'm not sure that is an improvement, and I think it is possible that we will find other code in Sphinx that is currently wrapping exceptions and needs to be changed (and will then need access to this variable).

jbms:
I added a changelog entry.

tk0miya (approved):
LGTM :-)

PR #10639, opened 2022-07-04
LaTeX: optional rounded corners for framing of code-blocks

jfbu:
Rounded corners for the framing of code-blocks, via a new configuration verbatimradius of 'sphinxsetup', which defaults
to 0pt.

If this dimension is non zero, the LaTeX package pict2e will be loaded
to construct frames with rounded corners for code-blocks.  Also, the
default background color will then change from white to a light
gray. And if the LaTeX package pict2e was not found, a LaTeX warning is
issued during PDF build.

### Feature or Bugfix

- Feature

By default, nothing changes: one needs to use the ``verbatimradius``  LaTeX key inside ``'sphinxsetup'``. And if one sets it  to a strictly postitive value, the layout in PDFs should not be modified if `verbartimsep` and `verbatimborder` stay the same.

### Detail

Capture d’écran 2022-07-04 à 23 36 24

another example (the above was with default colors triggered from setting ``verbatimradius>0pt``)

Capture d’écran 2022-07-04 à 23 54 51

This uses the [pict2e](https://ctan.org/pkg/pict2e) which is light weight.  But on Ubuntu it is part of the somewhat big [texlive-pictures](https://packages.ubuntu.com/bionic/texlive-pictures), which contains all of pgf+TikZ...

I also tested with  a [tcolorbox](https://ctan.org/pkg/tcolorbox) based implementation but there is not much interest if the box is not made much fancier.  The impact on build time is very noticeable on big documents if using `\tcbox` in place of our custom own-defined rounded box based on the [pict2e](https://ctan.org/pkg/pict2e) interface to the PDF operators.

Compatible with all engines as far as I know. But not tested for platex (Japanese).

PR #10648, opened 2022-07-07
LaTeX: box padding (t,r,b,l), borders (t,r,b,l), shadows (x,y) or radii

jfbu:
- [x] Implement an enhanced coloured box allowing 4 padding parameters, 4 border widths, a shadow with possibly negative x-offset/y-offset, and possibly inset, and colours for background, border and shadow

- [x] Do also with rounded corners with one radius per corner (circle arcs only) and constant border widths.

- [x] Also do rounded boxes with shadows

- [x] At page breaks, implement "open" style variant.  Done at this stage for code-blocks via a <strike> `verbatimopenframe` LaTeX-key added to `'sphinxsetup'`. </strike> But perhaps this option wll migrate to the envisioned CSS style (see next items). Indeed: Finally using `pre_box-decoration-break=slice` syntax.

- [x] Provide some interface to configure `make latexpdf`  to actually use the above styling possibility

This is a big enhancement of LaTeX PDF.

<hr>
  
<b>TL;DR:</b>
Short story: I have implemented all as ``'sphinxsetup'`` options but with CSS-like names.  In future it is to be hoped some automatic conversion from true CSS will be added.

Long musings:
<i>
My idea was to add perhaps a configuration variable `latex_style` which could be for example a string in simple CSS with styling for `nav.contents`, `aside.topic`, `admonition`, `note`, ..., `warning`, ..., `pre`.  Rather than extending again the `\sphinxsetup`. Via this `latex_style` either a parsing of it by Python (we don't need a full fledged CSS parser, but on the other hand if users see that CSS syntax work, they will want copy paste from HTML CSS then parsing becomes complicated), or to some extent I can parse by TeX, but this will not be good for future maintenance. But this would be feasible for example to support the CSS syntax for color specs (with some difficulties for `rgb(R%,G%,B%)` input, so a Python replace would replace `%` by `\%` for example which can then be used by TeX macro programmer).

I was also thinking one could allow "body" in the `latex_style` for example to set a page background color in PDF, or configure the font. And each of the above selectors could be with `font-weight` etc.., styling too, as well as `color`.

Apart from rounded boxes with varying border width which I have not yet done (the problem is that I recall some PDF viewer prefer border to be drawn latest, else I would fill the path and then overlay the background on a small path; so here I have to do lineto's plus one path to fill for each corner, which is work), the LaTeX is basically done.

The problem is how best to present a user-interface for the 4 paddings, 4 border-widths, 2 shadow sizes, and various colours...

...as said above I thought the most convenient for Sphinx user would be a CSS input.

Any advice? I can't now do this soon. Also, titles for admonitions are a bit problematic for styling because the current LaTeX rendering maps if we were to emulate it via CSS style is something like this
```
p.admonition-title { display: inline; }
p.admonition-title + p  { display: inline; }
p.admonition-title:after { content: ":\0020"; } 
```
(in LaTeX there are 2 space tokens after the colon for legacy reasons).  This is very complicated way here to represent the actual legacy LaTeX behaviour...

</i>

An example of box with varying padding and 4 distinct radii:

Capture d’écran 2022-07-07 à 23 08 01

An example of "openness" at page breaks:

Capture d’écran 2022-07-08 à 10 35 51

### Feature or Bugfix

- Feature: code-blocks can have shadows, can behave in a "slice" manner at page-breaks, topic boxes can have background color and separated customizability of all border widths, rounded corners, warning-type notices can have 4 borderwidths, 4 paddings, rounded corners, shadows... etc...

- Bugfix: #10650, #10651

### Relates

#10639

### DIFF
--- a/doc/latex.rst
+++ b/doc/latex.rst
@@ -8,21 +8,26 @@ LaTeX customization
 
    \begingroup
    \sphinxsetup{%
-         verbatimborder=2pt,
-         verbatimsep=5pt,
-         verbatimradius=5pt,
-         VerbatimColor={named}{OldLace},
-         VerbatimBorderColor={named}{Gold},
-         TitleColor={named}{DarkGoldenrod},
-         hintBorderColor={named}{LightCoral},
-         attentionborder=3pt,
-         attentionBorderColor={named}{Crimson},
-         attentionBgColor={named}{FloralWhite},
-         noteborder=2pt,
-         noteBorderColor={named}{Olive},
-         cautionborder=3pt,
-         cautionBorderColor={named}{Cyan},
-         cautionBgColor={named}{LightCyan}}
+      TitleColor={named}{DarkGoldenrod},
+      pre_border-width=2pt,
+      pre_padding=5pt,
+      pre_border-radius=5pt,
+      pre_background-TeXcolor={named}{OldLace},
+      pre_border-TeXcolor={named}{Gold},
+      div.warning_border-width=3pt,
+      div.warning_padding=6pt,
+      div.warning_padding-right=18pt,
+      div.warning_padding-bottom=18pt,
+      div.warning_border-TeXcolor={named}{DarkCyan},
+      div.warning_background-TeXcolor={named}{LightCyan},
+      div.warning_box-shadow=-12pt -12pt inset,
+      div.warning_box-shadow-TeXcolor={named}{Cyan},
+      attentionborder=3pt,
+      attentionBorderColor={named}{Crimson},
+      attentionBgColor={named}{FloralWhite},
+      noteborder=2pt,
+      noteBorderColor={named}{Olive},
+      hintBorderColor={named}{LightCoral}}
    \relax
 
 Unlike :ref:`the HTML builders <html-themes>`, the ``latex`` builder does not
@@ -65,7 +70,10 @@ The ``latex_elements`` configuration setting
 
 A dictionary that contains LaTeX snippets overriding those Sphinx usually puts
 into the generated ``.tex`` files.  Its ``'sphinxsetup'`` key is described
-:ref:`separately <latexsphinxsetup>`.
+:ref:`separately <latexsphinxsetup>`.  It allows also local configurations
+inserted in generated files, via :rst:dir:`raw` directives.  For example, in
+the PDF documentation this chapter is styled especially, as will be described
+later.
 
 Keys that you may want to override include:
 
@@ -575,35 +583,48 @@ The colors used in the above are provided by the ``svgnames`` option of the
    }
 
 It is possible to insert further uses of the ``\sphinxsetup`` LaTeX macro
-directly into the body of the document, via the help of the :rst:dir:`raw`
+directly into the body of the document, via the help of the ``raw``
 directive. This chapter is styled in the PDF output using the following at the
-start of the chapter::
+start of the chapter (which uses keys described later in :ref:`additionalcss`)::
 
   .. raw:: latex
 
      \begingroup
      \sphinxsetup{%
-           % These were used as defaults for the whole sphinx.pdf
-           % VerbatimColor={RGB}{242,242,242},%
-           % VerbatimBorderColor={RGB}{32,32,32},%
-           % verbatimradius=3pt,%
-           % New definitions for this chapter code-blocks only:
-           VerbatimColor={named}{OldLace},
-           VerbatimBorderColor={named}{Gold},
-           verbatimradius=5pt,
-           verbatimsep=5pt,
-           verbatimborder=2pt,
-           % Other configuration for this chapter only:
-           TitleColor={named}{DarkGoldenrod},
-           hintBorderColor={named}{LightCoral},
-           attentionborder=3pt,
-           attentionBorderColor={named}{Crimson},
-           attentionBgColor={named}{FloralWhite},
-           noteborder=2pt,
-           noteBorderColor={named}{Olive},
-           cautionborder=3pt,
-           cautionBorderColor={named}{Cyan},
-           cautionBgColor={named}{LightCyan}}
+         TitleColor={named}{DarkGoldenrod},
+         % pre_border-width is 5.1.0 alias for verbatimborder
+         pre_border-width=2pt,
+         % pre_padding is 5.1.0 alias for verbatimsep
+         pre_padding=5pt,
+         % rounded boxes are new at 5.1.0
+         pre_border-radius=5pt,
+         % TeXcolor means syntax must be as for LaTeX \definecolor
+         pre_background-TeXcolor={named}{OldLace},
+         pre_border-TeXcolor={named}{Gold},
+         %
+         % 5.1.0 alias for warningborder
+         div.warning_border-width=3pt,
+         div.warning_padding=6pt,
+         div.warning_padding-right=18pt,
+         div.warning_padding-bottom=18pt,
+         div.warning_border-TeXcolor={named}{DarkCyan},
+         div.warning_background-TeXcolor={named}{LightCyan},
+         div.warning_box-shadow=-12pt -12pt inset,
+         div.warning_box-shadow-TeXcolor={named}{Cyan},
+         %
+         % 5.1.0 new name would be div.attention_border-width
+         attentionborder=3pt,
+         % same as div.attention_border-TeXcolor
+         attentionBorderColor={named}{Crimson},
+         % same as div.attention_background-TeXcolor
+         attentionBgColor={named}{FloralWhite},
+         %
+         % no CSS-like names yet at 5.1.0 for note-type admonitions
+         noteborder=2pt,
+         noteBorderColor={named}{Olive},
+         hintBorderColor={named}{LightCoral}%
+     }
+
 
 And this is placed at the end of the chapter source to end the scope of
 the configuration::
@@ -852,32 +873,6 @@ Do not use quotes to enclose values, whether numerical or strings.
 
     Default: ``\fboxrule``
 
-``verbatimradius``
-    The radius of the rounded corners of the frame around :rst:dir:`code-block`\ s.
-
-    Default: ``0pt``
-
-    .. versionadded:: 5.1.0
-
-       If non-zero, it will trigger the loading of LaTeX package pict2e_.
-
-       .. caution::
-
-          The interface will remain experimental during the 5.x cycle: the
-          feature itself will remain, but the interface may evolve to include
-          other display elements such as admonitions.  For the latter it is
-          actually already possible for LaTeX-expert Sphinx users to use
-          functionalities of LaTeX packages such as tcolorbox_, via
-          redefinitions of the ``sphinxheavybox`` and ``sphinxlightbox``
-          `Environments`_.  This would be more difficult for code-blocks, and
-          this key achieves it.  The pict2e_ interface to some basic PDF
-          graphics operations is much more light-weight than the pgf_
-          framework used by tcolorbox_.
-
-    .. _pict2e: https://ctan.org/pkg/pict2e
-    .. _tcolorbox: https://ctan.org/pkg/tcolorbox
-    .. _pgf: https://ctan.org/pkg/pgf
-
 ``shadowsep``
     The separation between contents and frame for :dudir:`contents` and
     :dudir:`topic` boxes.
@@ -893,20 +888,6 @@ Do not use quotes to enclose values, whether numerical or strings.
     The width of the frame around :dudir:`topic` boxes.
 
     Default: ``\fboxrule``
-
-``shadowBorderColor``
-    The color of the frame around :dudir:`topic` boxes.
-
-    Default: ``{rgb}{0,0,0}`` (black)
-
-    .. versionadded:: 5.1.0
-
-``shadowShadowColor``
-    The color of the lateral "shadow" to the right and bottom.
-
-    Default: ``{rgb}{0,0,0}`` (black)
-
-    .. versionadded:: 5.1.0
 
 |notebdcolors|
     The colour for the two horizontal rules used by Sphinx in LaTeX for styling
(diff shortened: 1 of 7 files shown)


jfbu:
I had used ``aside.topic`` influenced by recent addition to our ``basic.css_t`` but ``div.topic`` is better after all because it is still there in ``basic.css_t`` and in PDF output topic boxes are in the vertical flow, they are not really typeset "aside".

doc/latex.rst
```diff
@@ -8,21 +8,21 @@ LaTeX customization
...
-         VerbatimColor={named}{OldLace},
-         VerbatimBorderColor={named}{Gold},
+         pre_border-width=2pt,
+         pre_padding=5pt,
+         pre_border-radius=5pt,
+         pre_background-TeXcolor={named}{OldLace},
+         pre_border-TeXcolor={named}{Gold},
          TitleColor={named}{DarkGoldenrod},
          hintBorderColor={named}{LightCoral},
          attentionborder=3pt,
```
jfbu: I left those to show that old option names still worked but perhaps I should add a comment to that effect. Legacy syntax: ``attentionborder=3pt``, new syntax ``div.attention_border-width=3pt``, or ``attentionBorderColor`` versus ``div.attention_border-TeXcolor``.

doc/latex.rst
```diff
@@ -8,21 +8,21 @@ LaTeX customization
...
          attentionBorderColor={named}{Crimson},
          attentionBgColor={named}{FloralWhite},
          noteborder=2pt,
          noteBorderColor={named}{Olive},
-         cautionborder=3pt,
-         cautionBorderColor={named}{Cyan},
-         cautionBgColor={named}{LightCyan}}
+         div.caution_border-width=3pt,
+         div.caution_border-TeXcolor={named}{Cyan},
+         div.caution_background-TeXcolor={named}{LightCyan}}
```
jfbu: Reason for TeXcolor and not color is due to input syntax one sees on right side.  I could add a parser of HTML color specs but well, that would still extend TeX code and may introduce frailties, better to wait for a more general approach which would allow general CSS input then extract from it stuff that this PR can interpret as options for ``'sphinxsetup'`` with TeX syntax.

doc/latex.rst
```diff
@@ -575,35 +578,30 @@ The colors used in the above are provided by the ``svgnames`` option of the
    }
 
 It is possible to insert further uses of the ``\sphinxsetup`` LaTeX macro
-directly into the body of the document, via the help of the :rst:dir:`raw`
+directly into the body of the document, via the help of the ``raw``
```
jfbu, line 586: It seems ``:rst:dir:`raw` `` did not give functioning link and I had not time (and don't feel now either) to check syntax

doc/latex.rst
```diff
@@ -575,35 +578,30 @@ The colors used in the above are provided by the ``svgnames`` option of the
...
+         pre_border-width=2pt,
+         pre_padding=5pt,
+         pre_border-radius=5pt,
+         pre_background-TeXcolor={named}{OldLace},
+         pre_border-TeXcolor={named}{Gold},
+         TitleColor={named}{DarkGoldenrod},
+         hintBorderColor={named}{LightCoral},
+         attentionborder=3pt,
+         attentionBorderColor={named}{Crimson},
+         attentionBgColor={named}{FloralWhite},
```
jfbu, line 620: refer above comment

doc/latex.rst
```diff
@@ -852,32 +850,6 @@ Do not use quotes to enclose values, whether numerical or strings.
...
-          redefinitions of the ``sphinxheavybox`` and ``sphinxlightbox``
-          `Environments`_.  This would be more difficult for code-blocks, and
-          this key achieves it.  The pict2e_ interface to some basic PDF
-          graphics operations is much more light-weight than the pgf_
-          framework used by tcolorbox_.
-
-    .. _pict2e: https://ctan.org/pkg/pict2e
-    .. _tcolorbox: https://ctan.org/pkg/tcolorbox
-    .. _pgf: https://ctan.org/pkg/pgf
-
```
jfbu, line 880: I removed this because it was not part of a release yet.  So ``verbatimradius`` replaced by  ``pre_border-radius``

doc/latex.rst
```diff
@@ -894,20 +866,6 @@ Do not use quotes to enclose values, whether numerical or strings.
...
-    .. versionadded:: 5.1.0
-
-``shadowShadowColor``
-    The color of the lateral "shadow" to the right and bottom.
-
-    Default: ``{rgb}{0,0,0}`` (black)
-
-    .. versionadded:: 5.1.0
-
```
jfbu, line 910: I removed this as not part yet of a release.  Rather than old style naming this is now ``div.topic_border-TeXcolor`` and ``div.topic_box-shadow-TeXcolor`` and there is also an additional ``div.topic_background-TeXcolor``.

doc/latex.rst
```diff
@@ -982,6 +940,213 @@ Do not use quotes to enclose values, whether numerical or strings.
...
+   Rounded boxes are done using the pict2e_ interface to some basic PDF
+   graphics operations.  If this LaTeX package can not be found the build will
+   proceed and render all boxes with straight corners.
+
+.. _pict2e: https://ctan.org/pkg/pict2e
+
+
+Options for topic boxes:
+
+- ``div.topic_border-top-width``,
```
jfbu: Initially I used ``aside.topic`` but probably ``div.topic`` is better.

sphinx/texinputs/sphinx.sty
```diff
@@ -47,6 +47,11 @@
 }{
     \RequirePackage{color}
 }
+% the \colorlet of xcolor (if at all loaded) is overkill for our use case
+\newcommand{\sphinxcolorlet}[2]
+    {\expandafter\let\csname\@backslashchar color@#1\expandafter\endcsname
+                     \csname\@backslashchar color@#2\endcsname }
+
```
jfbu, line 54: This hunk was originally in sphinxlatexadmonitions.sty but we need it in sphinx.sty now

sphinx/texinputs/sphinx.sty
```diff
@@ -72,36 +77,19 @@
 \AtBeginDocument{\define@key{sphinx}{bookmarksdepth}{\hypersetup{bookmarksdepth=#1}}}
 % \DeclareBoolOption[false]{usespart}% not used
 % dimensions, we declare the \dimen registers here.
-\newdimen\sphinxverbatimsep
+\newdimen\sphinxverbatimsep    % <-- TO BE REMOVED NOT USED ANYMORE AT 5.1.0
```
jfbu, line 80: indeed I removed usage of this dimen register, replaced by macro holder. Probably we can remove it at 6.0.0

sphinx/texinputs/sphinx.sty
```diff
@@ -72,36 +77,19 @@
...
 \newdimen\sphinxverbatimborder
-\newdimen\sphinxverbatimradius
-\newdimen\sphinxshadowsep
-\newdimen\sphinxshadowsize
-\newdimen\sphinxshadowrule
+%
```
jfbu, line 82: The above are simply moved a little further down. But `\sphinxverbatimradius`  was never released and is removed

sphinx/texinputs/sphinx.sty
```diff
@@ -72,36 +77,19 @@
...
+\sphinxverbatimsep=\fboxsep    % <-- TO BE REMOVED NOT USED ANYMORE AT 5.1.0
   \define@key{sphinx}{verbatimsep}{\sphinxverbatimsep\dimexpr #1\relax}
 \sphinxverbatimborder=\fboxrule
   \define@key{sphinx}{verbatimborder}{\sphinxverbatimborder\dimexpr #1\relax}
-\sphinxverbatimradius=0pt
-  \define@key{sphinx}{verbatimradius}{%
-     \sphinxverbatimradius\dimexpr #1\relax
-     \ifdim\sphinxverbatimradius=\z@\else\RequirePackage{sphinxpackageboxes}\fi
-     }
-\AtBeginDocument{\define@key{sphinx}{verbatimradius}{\sphinxverbatimradius\dimexpr #1\relax}}
```
jfbu, line 96: This line and above removed because new names and was never part of a release

sphinx/texinputs/sphinx.sty
```diff
@@ -72,36 +77,19 @@
...
-% topic boxes
-\sphinxshadowsep =5pt
-  \define@key{sphinx}{shadowsep}{\sphinxshadowsep\dimexpr #1\relax}
-\sphinxshadowsize=4pt
-  \define@key{sphinx}{shadowsize}{\sphinxshadowsize\dimexpr #1\relax}
-\sphinxshadowrule=\fboxrule
-  \define@key{sphinx}{shadowrule}{\sphinxshadowrule\dimexpr #1\relax}
-% verbatim
```
jfbu, line 104: above lines kept but moved down

sphinx/texinputs/sphinx.sty
```diff
@@ -161,9 +160,6 @@
 }%
 % Default color chosen to be as in minted.sty LaTeX package!
 \sphinxDeclareSphinxColorOption{VerbatimHighlightColor}{{rgb}{0.878,1,1}}
-% topic and contents boxes
-\sphinxDeclareSphinxColorOption{shadowBorderColor}{{rgb}{0,0,0}}
-\sphinxDeclareSphinxColorOption{shadowShadowColor}{{rgb}{0,0,0}}
```
jfbu, line 166: removed because never released and new names for the colour options

sphinx/texinputs/sphinx.sty
```diff
@@ -180,6 +176,414 @@
 \sphinxDeclareSphinxColorOption{attentionBgColor}{{rgb}{1,1,1}}
 \sphinxDeclareSphinxColorOption{dangerBgColor}{{rgb}{1,1,1}}
 \sphinxDeclareSphinxColorOption{errorBgColor}{{rgb}{1,1,1}}
+%%%%%%%%
+%
+% Additions at 5.1.0
```
jfbu, line 181: These additions are done in a backwards compatible way.  Old option such as ``verbatimborder`` is made an alias for new name ``pre_border-width`` for example.

sphinx/texinputs/sphinx.sty
```diff
@@ -180,6 +176,414 @@
...
+    \spx@topic@withbackgroundcolortrue
+    \definecolor{sphinxTopicBackgroundColor}#1%
+}
+\newif\ifspx@topic@withshadowcolor
+\define@key{sphinx}{div.topic_box-shadow-TeXcolor}{%
+    \spx@topic@withshadowcolortrue
+    \definecolor{sphinxTopicShadowColor}#1%
+}
+% warning, caution, attention, danger, error
+\def\spx@tempc#1{%
```
jfbu: This whole section is a bit hard as it uses TeX coding techniques to handle five admonition types but basically it is a repeat of what was done above explictly for code-blocks and topic boxes.

This is all backwards compatible with one exception: formerly there was no padding parameter for such warning boxes.  The padding was computed dynamically at time of typesetting in a way which made its sum with the (unique) border-width constant equal to 0.6\baselineskip (thus normally 7.2 pt excepr for Japanese documents I guess).

If user did not assign a value to ``warningborder`` for example the document will compile exactly the same. But if another value from default was used, it would be needed now to use the interface for padding if one wanted to maintain exactly same PDF output.  Anyway I don't think this is really important.  People using Sphinx LaTeX probably don't care much if vertical spacing changes a bit.  In the past when I started enhancing rendering I went to extreme measures to maintain exact same LaTeX output and this is actually the cause of the strange behaviour (see #10651) of handling of border-width for warning: when I made this configurable I stupidly decided that the text baseline should remain at same location even if border width was changed.

sphinx/texinputs/sphinx.sty
```diff
@@ -278,12 +682,10 @@
 
 %% FRAMED ENVIRONMENTS
 %
+\RequirePackage{sphinxpackageboxes}
 \input{sphinxlatexadmonitions.sty}
 \input{sphinxlatexliterals.sty}
 \input{sphinxlatexshadowbox.sty}
-\ifdim\sphinxverbatimradius=\z@\else
-  \RequirePackage{sphinxpackageboxes}
-\fi
```
jfbu, line 286: this hunk was from never released code, it is removed from re-factoring

sphinx/texinputs/sphinxlatexadmonitions.sty
```diff
@@ -65,24 +66,117 @@
...
+  \spx@boxes@radius@bottomright \@nameuse{spx@\spx@noticetype @radius@bottomright}%
+  \spx@boxes@radius@bottomleft  \@nameuse{spx@\spx@noticetype @radius@bottomleft}%
+  \relax
+  \iftrue\@nameuse{ifspx@\spx@noticetype @withshadow}%
+     \spx@boxes@withshadowtrue
+     \spx@boxes@shadow@xoffset \@nameuse{spx@\spx@noticetype @box@shadow@xoffset}%
+     \spx@boxes@shadow@yoffset \@nameuse{spx@\spx@noticetype @box@shadow@yoffset}\relax
+  \else
+     \spx@boxes@withshadowfalse
+  \fi\@nameuse{fi}%
```
jfbu, line 92: The strange ``\iftrue`` is because of aspects of  TeX conditionals which I can't explain here ;-)

sphinx/texinputs/sphinxlatexadmonitions.sty
```diff
@@ -65,24 +66,117 @@
...
+   % Those are used by sphinxVerbatim if the \ifspx@inframed boolean is true
+   \setlength{\FrameRule}{0.5\dimexpr\spx@boxes@border@top+\spx@boxes@border@bottom\relax}%
+   % Prior to 5.1.0 \FrameSep served to configure the (unique) padding, via
+   % \fboxsep parameter of \fcolorbox; it was determined as 0.6\baselineskip -
+   % borderwidth, increasing borderwidth did not change text location
+   % vertically as borderwidth+padding remained constant.  But this limited
+   % considerably (at most to 4pt or 5pt) the usual borderwidths.
+   % Now \FrameSep is not used, but is needed still by sphinxVerbatim (see
+   % prior comment above).  We give it some reasonable definition.
+   \setlength{\FrameSep}{0.5\dimexpr\spx@boxes@padding@top+\spx@boxes@padding@bottom\relax}%
```
jfbu, line 137: This is the weird situation with border width I alluded to above

sphinx/texinputs/sphinxlatexadmonitions.sty
```diff
@@ -65,24 +66,117 @@
...
+                \spx@boxes@fcolorbox@rounded{##1}%
+              \fi}%
+             {\spx@boxes@fcolorbox{##1}}%
+        \fi
+  \ifspx@boxes@withshadow
+   \ifspx@boxes@insetshadow\else
+    \ifdim\spx@boxes@shadow@xoffset>\z@\hskip-\spx@boxes@shadow@xoffset\relax
+    \fi
+   \fi
+  \fi
```
jfbu, line 178: in future I wll perhaps refactor as this kind of code is encountered thrice: here, for code-blocks and for topics, and is a bit bulky

sphinx/texinputs/sphinxlatexadmonitions.sty
```diff
@@ -129,11 +223,6 @@
   {\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}
 % or just use package options
 
-% the \colorlet of xcolor (if at all loaded) is overkill for our use case
-\newcommand{\sphinxcolorlet}[2]
-    {\expandafter\let\csname\@backslashchar color@#1\expandafter\endcsname
-                     \csname\@backslashchar color@#2\endcsname }
-
```
jfbu, line 136: hunk was moved to sphinx.sty

sphinx/texinputs/sphinxlatexliterals.sty
```diff
@@ -60,8 +62,8 @@
 % Skip to next page if not enough space at bottom
 \RequirePackage{needspace}
 
-\newbox\spx@verb@tempboxa
-\newbox\spx@verb@tempboxb
+\newbox\spx@tempboxa
+\newbox\spx@tempboxb
```
jfbu: I think I meant those lines to be in sphinxpackageboxes.sty will amend later

sphinx/texinputs/sphinxlatexliterals.sty
```diff
@@ -119,14 +120,30 @@
 % #3 are the contents, and in the context of usage of fancyvrb+framed,
 % it will arrive here already transformed into horizontal boxes,
 % interline penalties and glues.
+%
+% TODO: "open borders" at page breaks via First, Mid, Last "frame commands"
+% passing the information here to set the
+% \spx@boxes@border@{bottom,top} to zero appropriately
```
jfbu: This is obsolete, I have implemented "open border" at pagebreaks since, (for code-blocks only which is what this file is about). To be removed.

sphinx/texinputs/sphinxlatexliterals.sty
```diff
@@ -138,82 +155,148 @@
...
+        % formerly last prior item in vertical list.  TODO: remove this at 6.0.0 ?
         \nointerlineskip
-        % TODO: add some \lineskip glue here, this is all in a \vbox so can't split
-        #2% continuation hint attached below frame
+        #2% continuation hint attached below frame, uses \spx@verb@fcolorbox@width@sp
        }% end of \vbox
 }%
-\def\spx@fcolorbox@put@c#1{% hide width from framed.sty measuring
-    \moveright.5\wd\tw@\hb@xt@\z@{\hss#1\hss}%
+\def\spx@verb@fcolorbox@put@c#1{% hide width from framed.sty measuring
```
jfbu, line 181: this renaming is for internal macro not part of any public API, same for other changes

sphinx/texinputs/sphinxlatexliterals.sty
```diff
@@ -138,82 +155,148 @@
...
+                                \spx@boxes@withbackgroundcolorfalse\fi
+  \sphinxcolorlet{spx@boxes@backgroundcolor}{VerbatimColor}%
+  %
+  \ifspx@pre@withbordercolor\spx@boxes@withbordercolortrue\else
+                            \spx@boxes@withbordercolorfalse\fi
+  \sphinxcolorlet{spx@boxes@bordercolor}{VerbatimBorderColor}%
+  %
+  \ifspx@pre@withshadowcolor\spx@boxes@withshadowcolortrue\else
+                            \spx@boxes@withshadowcolorfalse\fi
+  \sphinxcolorlet{spx@boxes@shadowcolor}{sphinxVerbatimShadowColor}%
```
jfbu, line 274: this one color is new here, formerly code-blocks could not be rendered with  a shadowed box

sphinx/texinputs/sphinxlatexshadowbox.sty
```diff
@@ -16,74 +16,83 @@
...
+     \spx@boxes@withshadowfalse
+  \fi
+  \ifspx@topic@insetshadow\spx@boxes@insetshadowtrue\else
+                          \spx@boxes@insetshadowfalse\fi
+  \ifspx@topic@withshadowcolor\spx@boxes@withshadowcolortrue\else
+                              \spx@boxes@withshadowcolorfalse\fi
+  %
+  \ifspx@topic@withbackgroundcolor\spx@boxes@withbackgroundcolortrue\else
+                                  \spx@boxes@withbackgroundcolorfalse\fi
+  \sphinxcolorlet{spx@boxes@backgroundcolor}{sphinxTopicBackgroundColor}%
```
jfbu, line 53: topic boxes could not have background color nor rounded corners in the past

sphinx/texinputs/sphinxpackageboxes.sty
```diff
@@ -2,21 +2,61 @@
...
+% - pict2e.  Ideally we would need a recent version of this package which
+%   allows dimensional arguments to its \moveto, \lineto, etc...  but we add
+%   ourselves some wrappers equivalent to its v0.4a 2020/08/16 version.
 
-% Currently, this file only provides a replacement to the \spx@verb@fcolorbox
-% of sphinxlatexliterals.sty which will draw boxes with a frame having
-% rounded corners, and a background color.
+% Provides box registers \spx@tempboxa, \spx@tempboxb usable in other places
+\newbox\spx@tempboxa
+\newbox\spx@tempboxb
```
jfbu, line 13: there is currently a duplicate in sphinxlatexliterals.sty which I will remove

sphinx/texinputs/sphinxpackageboxes.sty
```diff
@@ -2,21 +2,61 @@
...
+
+% WIP
```
jfbu: this WIP comment to be removed

sphinx/texinputs/sphinxpackageboxes.sty
```diff
@@ -2,21 +2,61 @@
...
-% - pict2e.  Ideally we would need a recent version of this package which
-%   allows dimensional arguments to its \moveto, \lineto, etc...
-%   but we add ourselves some wrapper to facilitate the usage.
+\newdimen\spx@boxes@shadow@xoffset
+\newdimen\spx@boxes@shadow@yoffset
+%
+\newdimen\spx@boxes@radius@topleft % only circular arcs, x-radius same as y-radius
+\newdimen\spx@boxes@radius@topright
+\newdimen\spx@boxes@radius@bottomright
+\newdimen\spx@boxes@radius@bottomleft
```
jfbu, line 40: here a bunch of \dimen's are defined for ease of usage, but the dimensions specified by options will be stored in macros before being assigned to them to avoid having to allocate dozens of \dimen's

sphinx/texinputs/sphinxpackageboxes.sty
```diff
@@ -38,79 +80,358 @@
...
-% dimensions or dimensional expressions.  The \unitlength will always be 1pt.
-\def\spx@moveto(#1,#2)%
-   {\moveto(\strip@pt\dimexpr#1,\strip@pt\dimexpr#2)}
-\def\spx@lineto(#1,#2)%
-   {\lineto(\strip@pt\dimexpr#1,\strip@pt\dimexpr#2)}
-% attention we use here [2] always; and there are two more mandatory
-% arguments, angles, we don't need to worry about them here.
-\def\spx@circlearc#1#2#3{\circlearc[2]%
-    {\strip@pt\dimexpr#1}{\strip@pt\dimexpr#2}{\strip@pt\dimexpr#3}%
+       and verbatimradius has been ignored}}%
```
jfbu: this line to be edited because ``verbatimradius`` does not exist anymore as option name

sphinx/texinputs/sphinxpackageboxes.sty
```diff
@@ -38,79 +80,358 @@
...
+% \spx@boxes@fcolorbox.  Now this is only a hook, because the initialization
+% will be done by caller rather.
+\let\spx@boxes@fcolorbox@setuphook\@empty
+% The same applies to \spx@boxes@fcolorbox@rounded, also with (same name)
+% \spx@boxes@fcolorbox@setup left to caller configuration.
+%
+% The parameters are interpreted as they would as CSS properties.
+% For the shadow inset variant see separate \spx@boxes@fcolorbox@insetshadow
+\long\def\spx@boxes@fcolorbox#1{%
+  \hbox\bgroup
```
jfbu, line 107: this \hbox is started here and ended in another macro depending whether with inset shadow or not

sphinx/texinputs/sphinxpackageboxes.sty
```diff
@@ -38,79 +80,358 @@
...
-      (\strip@pt\dimexpr-.5\sphinxverbatimborder,\strip@pt\dimexpr-.5\sphinxverbatimborder)%
-              \color{VerbatimColor}% color for background
-              \spx@dopath\fillpath
-              \color{VerbatimBorderColor}% color for border
-\ifspx@opt@verbatimwithframe % even with \sphinxverbatimborder set to 0pt, the
-                             % stroke will produce a visible contour, so we
-                             % must explicitly exclude doing it.
-              \linethickness{\sphinxverbatimborder}%
-              \spx@dopath\strokepath
+% Various radii but only one borderwidth. No shadow
```
jfbu: to be edited because I actually added the shadow

jfbu (commented):
I noticed in the latex code some obsoleted comments which I will update (done).

In this review I make some comments to try to explain aspects of the design of this large PR.  Backwards compatiblity is respected, apart from one minor aspect, see comment about warning type boxes and their border and padding.

tk0miya (approved):
LGTM.

jfbu:
I have ensured legacy implicit computation of padding from the border-width user setting of warning type notices (via ``warningborder``  etc...) will apply, so no change to PDF even for proojects having used ``warningborder`` etc...

Will merge after testing completes.

Issue #10419, opened 2022-05-04
Update the web theme used in Sphinx' documentation


choldgraf:
### Context

_This issue is not a specific proposal, but I wanted to see if others in the community agree with me here, in the hopes that this might lead to a pathway forward, or explicitly settle the issue as not high-priority._

The Sphinx documentation has an excellent amount of information in it, and is a resource that I regularly go back to in order to learn about the project and its APIs. However because there is a lot of information, it becomes particularly important that this information is easy to navigate and discover different topics.

The website theme for Sphinx was added [about 9 years ago](https://github.com/sphinx-doc/sphinx/blame/335bf513e85eb2727322a2be9a731af427584fe0/doc/conf.py#L23). In the past several years, there have been a lot of new developments in both web standards and design/UI expectations, as well as many theme improvements in the Sphinx ecosystem.

For example:

- [Furo](https://pradyunsg.me/furo/quickstart) has become very popular across the Python ecosystem
- [the pydata theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/) is used in many parts of the PyData ecosystem and is designed for more complex docs layouts
- [the sphinx book theme](https://sphinx-book-theme.readthedocs.io/) is a minimalist theme that builds off of the pydata theme
- [lutra](https://pradyunsg.me/lutra/) is a newcomer that is still in development, but also aims to be a high-quality and modern theme for more complex documentation structures.

And [a lot of other nice themes are demoed here](https://sphinx-themes.org/).

### Question to discuss

In my opinion, Sphinx would benefit from having a theme that aligns more with how modern documentation sites tend to be structured. This could either be via improvements to its current theme, or adopting another theme that is in the community.

Do others generally agree or disagree that an overhaul of the theme to be more modern would have a large benefit? Has this been discussed before, and / or do others have ideas of what themes could be explored?

### More specific issues

Since this issue is fairly broad in its scope, and since there's not consensus on just switching to a community-developed theme, I've opened up a few issues for more specific improvements to make to the current Sphinx theme in case that makes things easier to discuss and track separately:

- https://github.com/sphinx-doc/sphinx/issues/10659
- https://github.com/sphinx-doc/sphinx/issues/10658
- https://github.com/sphinx-doc/sphinx/issues/10660
- https://github.com/sphinx-doc/sphinx/issues/10661
- https://github.com/sphinx-doc/sphinx/issues/10662
- https://github.com/sphinx-doc/sphinx/issues/10663
[...]

Issue #10646, opened 2022-07-06
Modernising the `basic` theme


AA-Turner:
The `basic` theme is showing its age -- non-responsive, and not really a great advert for Sphinx out-of-the-box.

Improving it is tricky though, as almost every theme relies on it through Jinja templating.

I see two options:

1) Try and modernise the theme, whilst preserving compatability as much as humanly possible.
2) Start over with basic v2

(1) is more attractive as the benefits flow through, but perhaps a lot more work.

At some point we should also evaluate the stable of default themes and if that needs a refresh, but I think the `basic` work is more pressing.

A

humitos:
I know that @pradyunsg and @choldgraf have had some good discussions about a newer basic theme and they created https://github.com/pradyunsg/sphinx-basic-ng/. It may useful to take those conversations and code as a base to decide the direction of this work in case it can be reused.

tk0miya:
I don't have knowledge of modern HTML. So I'm not sure which option is better. But +1 for improvement if somebody will help us.

>At some point we should also evaluate the stable of default themes and if that needs a refresh, but I think the basic work is more pressing.

All bundled themes are not maintained well. So it might be time to unbundle.

pradyunsg:
I'd be more than happy to help out with this -- see also https://github.com/sphinx-doc/sphinx/issues/10419#issuecomment-1181695008 where I responded to the suggestion of borrowing the new things from basic-ng!

AA-Turner:
Thank you Pradyun!

A

PR #10652, opened 2022-07-09
Refresh the Sphinx13 theme

AA-Turner:
This keeps the essence of the current Sphinx theme, refreshing it to feel a little more 'modern'. The changes also make the theme work a little better on mobile phones / other such smaller viewports.

xref #10419

Rendered preview: https://sphinx--10652.org.readthedocs.build/en/10652/

### Feature or Bugfix
- Feature

A


tk0miya (approved):
Cool! +1 for this change

jfbu (commented):
I sort of liked the detailed contents page as currently at https://www.sphinx-doc.org/en/master/contents.html, would it be possible to keep a version of it with a link to it on index.html inside the items listed in Documentation section? It seems that the "Contents" now on index.html links back to ``index.html#`` itself so we get to see only the main headings.  For people navigation on computers with large viewport, seems to me a detailed contents listing is useful as it allows to get a global view of the project with details.

jfbu (changes requested):
Suggested change:
```diff
diff --git a/doc/conf.py b/doc/conf.py
index de38cb458..c23a297e8 100644
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -56,7 +56,7 @@ epub_use_index = False
 epub_guide = (('toc', 'contents.xhtml', 'Table of Contents'),)
 epub_description = 'Sphinx documentation generator system manual'
 
-latex_documents = [('contents', 'sphinx.tex', 'Sphinx Documentation',
+latex_documents = [('index', 'sphinx.tex', 'Sphinx Documentation',
                     'Georg Brandl', 'manual', 1)]
 latex_logo = '_static/sphinx.png'
 latex_elements = {
```
Indeed on `make latex` I got this
```
make clean latex
Removing everything under '_build'...
Running Sphinx v5.1.0+/bd2f74f68
making output directory... done
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://requests.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://docs.readthedocs.io/en/stable/objects.inv...
building [latex]: all documents
updating environment: [new config] 92 added, 0 changed, 0 removed
reading sources... [100%] usage/theming                                         
pickling environment... done
checking consistency... done
WARNING: "latex_documents" config value references unknown document contents
copying TeX support files... copying TeX support files...
done
build succeeded, 1 warning.

The LaTeX files are in _build/latex.
Run 'make' in that directory to run these through (pdf)latex
(use `make latexpdf' here to do that automatically).
```

Incidentally there is a bug here of Sphinx because the LaTeX build directory actually does not contain any output tex file only the support style files, so the end of the message to user is plain wrong.

AA-Turner:
@jfbu I re-added the contents page; please would you have another look?

A

jfbu (approved):
Thanks for re-enacting contents!

I will be off-line for next two days likely, can you check my comment about `latex_documents` currently `make latexpdf` can not build  ;-) Approving merge in advance.

choldgraf:
I think this is definitely an improvement on what is there now. The mobile view looks better as well.

A couple quick suggestions, though in my opinion none should be blocking and this could be merged as-is as an iterative improvement either way.

1. Make the left sidebar content `position: sticky` so that it acts as an in-page table of contents for the page and is visible when people scroll down (at present, it just becomes empty space).
2. Rearrange the TOCTree slightly so that the introductory material is first in the sidebar.
3. Group the TOCTree items in `caption` groupings so that it is easier to quickly look for what you want to find. Something like `Get Started`, `Topic Guides`, `Reference`, `About`.
4. Rename `Man pages` to something else. I suspect that most people will not know what `Man pages` means.

Longer-term, I think that it would be much better if there were a way to see all of the pages underneath a top-level section in a way that could be expanded/collapsed (e.g., what the ReadTheDocs, Furo, PyData, or Book themes do), but that is probably not necessary in this PR in my opinion.

AA-Turner:
@choldgraf -- 2 through 4 aren't strictly to do with the theme, so I suggest they're taken in followup PRs.

#10646 covers improving the basic theme, which would wash through to all other themes -- your final suggestion I think would be more appropriate there.

A

AA-Turner:
(To be clear, I don't view this PR as an end-state -- it was a few hours' work on updating the styling, and is by no means perfect. But I think better to merge a mild improvement now and build on it rather than blocking it for a long time.)

benjaoming:
I would also like to see an alternative menu for mobile view. It's a bit problematic that this proposal removes the navigation?

image

AA-Turner:
@benjaoming I didn't see your comment before merging --- what would your suggestion here be?

A

benjaoming:
@AA-Turner 

My direct suggestion would be that a new element is created that's only intended for the mobile view. This is a normal pattern on responsive platforms because the sidebar often needs to not even be a sidebar. It transforms into a so-called "burger" menu :)

```
.sphinxsidebar-mobile {
    display: none;
}
@media (max-width: 750px) {
    .sphinxsidebar {
        display: none;
    }
    .sphinxsidebar-mobile {
        display: block;
    }
    div.body {
        border-left: none;
    }
}
```

AA-Turner:
(aside: you don't need to at-mention me every time!)

So `sphinxsidebar-mobile` would be an exact copy of the content of `sphinxsidebar`? The reason I set the sidebar to invisible on smaller viewports was that I didn't find a quick way of it looking good, and I wasn't sure about putting the sidebar content below the master body content, as it also felt odd.

Would you be able to propose a PR please just so I understand in a little more detail?

A

benjaoming:
> The reason I set the sidebar to invisible on smaller viewports was that I didn't find a quick way of it looking good, and I wasn't sure about putting the sidebar content below the master body content, as it also felt odd.

It makes total sense, hiding the current sidebar on small viewports would be the first step.

So there are basically two approaches:

1) Write CSS for the current `.sidebar` to look good on mobile but be hidden and displayed upon clicking a toggler button (the burger) OR
2) Create an entirely different `.sidebar-mobile` and have it displayed when the burger is clicked

By burger menu, I mean for instance what you see here: https://getbootstrap.com/docs/5.1/examples/carousel/

In the above Bootstrap example, there is enough CSS rules to make the same element display completely differently on mobile and desktop.

> Would you be able to propose a PR please just so I understand in a little more detail?

You're looking down a very deep tunnel of re-inventing stuff that other themes already manage much better. So with all respect, I find that it's a bit like tossing efforts in the wrong direction, that's why I made a couple of quick demos in #10491. But on the other hand, if someone has a big drive for creating front-end stuff, it's not impossible that Sphinx ships its own modern responsive theme.

choldgraf:
That makes sense @AA-Turner - I will open up modular issues for other things I've noticed and they can be discussed independently.

I would echo what @benjaoming notes though - modernizing the basic Sphinx themes is going to be a pretty good chunk of work. Though there is a lot of nice prior art now in community themes as well. Will open issues to track some of that as well.

PR #10671, opened 2022-07-16
Alphabetically sort the list of contributors

pradyunsg:
This matches the language used to describe the listing.

This is basically a reordering of the content. :)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,75 +19,75 @@ Other co-maintainers:
 
 Other contributors, listed alphabetically, are:
 
+* Adrián Chaves (Gallaecio) -- coverage builder improvements
 * Alastair Houghton -- Apple Help builder
 * Alexander Todorov -- inheritance_diagram tests and improvements
 * Andi Albrecht -- agogo theme
-* Jakob Lykke Andersen -- Rewritten C++ domain
-* Henrique Bastos -- SVG support for graphviz extension
-* Daniel Bültmann -- todo extension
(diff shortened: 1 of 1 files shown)


tk0miya:
Thanks!

AA-Turner:
@pradyunsg -- it seems the list was previously sorted fairly well by surname, whereas this PR changed it to forename sorting (only the first two entries were incorrect in the 'old' scheme).

If forename sorting is to be the way going forwards, we should update the note. Personally when I see 'sorted alphabetically' with a list of names, I expect surname sorting.

We should also update Jakob to be in the list of co-maintainers.

A

pradyunsg:
> Personally when I see 'sorted alphabetically' with a list of names, I expect surname sorting.

Fascinating! (human culture is weird)

I certainly don't read it that way. I read it as sorted by names, as they're written. That notably doesn't exclude folks who don't have surnames, which is uncommon but does happen.

None the less, if folks reckon there's value in clarifying this, I'd suggest adding "as written" if it helps disambiguation of what the sort order is (and keeping the current sort order :P).

AA-Turner:
> Fascinating! (human culture is weird)

Indeed, and I'm perhaps more influenced by the school registers &c of my childhood. I've written a brief PR to clarify the sorting scheme and incidentally to sort the maintainer list.

A

pradyunsg:
x-ref https://github.com/sphinx-doc/sphinx/pull/10676

PR #10676, opened 2022-07-17
Clarify sorting in AUTHORS; sort and update maintainers list

AA-Turner:
The maintainers list was also a little out of date, so I updated it, adding @jakobandersen @stephenfin and @francoisfreitag.

A

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,21 +1,19 @@
-Sphinx is written and maintained by Georg Brandl <georg@python.org>.
+Sphinx's maintainers, listed alphabetically by forename, are:
 
-Substantial parts of the templates were written by Armin Ronacher
-<armin.ronacher@active-4.com>.
-
-Other co-maintainers:
-
-* Takayuki Shimizukawa <shimizukawa@gmail.com>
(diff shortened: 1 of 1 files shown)


AUTHORS
```diff
@@ -1,23 +1,21 @@
...
 * Takeshi Komiya <@tk0miya>
-* Jean-François Burnol <@jfbu>
-* Yoshiki Shibukawa <@shibu_jp>
-* Timotheus Kampik - <@TimKam>
-* Adam Turner - <@AA-Turner>
+* Timotheus Kampik <@TimKam>
+* Yoshiki Shibukawa <@shibukawa>
 
-Other contributors, listed alphabetically, are:
+Other contributors, listed alphabetically by forename, are:
```
pradyunsg: It's not just the forename, since two people with the same first name will be sorted by their second name. :)

pradyunsg: This isn't hypothetical either, there's multiple folks named Martin in this list. :)

tk0miya (approved):

tk0miya:
Thanks!

PR #10679, opened 2022-07-18
Update bindep.txt for LaTeX pict2e.sty optional dependency

jfbu:
Refs: #10648 

When building PDF via LaTeX, the ``pict2e`` LaTeX package will be loaded:
- if in conf.py the ``'sphinxsetup'`` key of ``latex_elements`` has been used with one of the radii setting CSS-named-alike options added at #10648 
- and if it actually exists on user system,

So it is not a binding dependency as in case of its absence the ``make latexpdf``  falls back to rendering with straight corners.

Nevertheless adding it here.


AA-Turner (approved):

jfbu:
Thanks for review @AA-Turner! now merging...

PR #10680, opened 2022-07-18
Update include_patterns implementation

AA-Turner:
#10518

- Removes extraneous argument to `Matcher`
- Removes `include_patterns` interaction with `html_(static|extra)_path`
- Adds non-included document warning for TocTrees

A


sphinx/directives/other.py
```diff
@@ -130,6 +131,11 @@ def parse_content(self, toctree: addnodes.toctree) -> List[Node]:
                                    location=toctree)
                     self.env.note_reread()
                 else:
+                    if not included(self.env.doc2path(docname, False)):
```
tk0miya: If my understanding is correct, the "not included" documents are all ignored on `Project.discover()`. As a result, they're not added to `env.found_docs`. If so, they will never be reached here because former if-block catches them.

sphinx/directives/other.py
```diff
@@ -130,6 +131,11 @@ def parse_content(self, toctree: addnodes.toctree) -> List[Node]:
...
+                        logger.warning(
+                            __('toctree contains reference to non-included  document %r'),
+                            docname, type='toc', subtype='non_included', location=toctree,
```
tk0miya: Please add description for the new subtype to `doc/usage/configuration.rst`.

sphinx/environment/adapters/toctree.py
```diff
@@ -162,6 +163,10 @@ def _entries_from_toctree(toctreenode: addnodes.toctree, parents: List[str],
                         logger.warning(__('toctree contains reference to document %r that '
                                           'doesn\'t have a title: no link will be generated'),
                                        ref, location=toctreenode)
+                    if not included(self.env.doc2path(ref, False)):
```
tk0miya: I guess this block should be moved into the `except KeyError` block. Is my understanding correct?

PR #10681, opened 2022-07-19
Fix more strict static typing errors

danieleades:
fix more strict typing errors.

the purpose of this PR is to take a few more steps towards using `strict_optional = True` or even `strict = True` in mypy's config


sphinx/cmd/quickstart.py
```diff
@@ -321,10 +323,11 @@ def ask_user(d: Dict[str, Any]) -> None:
     print()
 
 
-def generate(d: Dict, overwrite: bool = True, silent: bool = False, templatedir: str = None
-             ) -> None:
+def generate(
+    d: Dict, overwrite: bool = True, silent: bool = False, templatedir: Optional[str] = None
```
danieleades, line 327: i'd prefer to move the default empty str to the method signature here, but wasn't sure how you'd feel about changing the type signature

sphinx/directives/code.py
```diff
@@ -239,7 +239,7 @@ def read(self, location: Optional[Tuple[str, int]] = None) -> Tuple[str, int]:
 
     def show_diff(self, location: Optional[Tuple[str, int]] = None) -> List[str]:
         new_lines = self.read_file(self.filename)
-        old_filename = self.options.get('diff')
+        old_filename = self.options['diff']
```
danieleades, line 242: doesn't look to me like the `None` case was handled before, so i don't think this is any less safe than before

sphinx/builders/latex/__init__.py
```diff
@@ -510,25 +510,25 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.connect('config-inited', validate_latex_theme_options, priority=800)
     app.connect('builder-inited', install_packages_for_ja)
 
-    app.add_config_value('latex_engine', default_latex_engine, None,
+    app.add_config_value('latex_engine', default_latex_engine, False,
```
tk0miya, line 513: Could you use an empty string for this case here? The docstring of `add_config_value()` tells us to use string instead of boolean.
https://github.com/sphinx-doc/sphinx/blob/89db13d71fcc949efc0a7d223683fe8901997d39/sphinx/application.py#L524-L527

sphinx/writers/texinfo.py
```diff
@@ -765,10 +765,10 @@ def depart_block_quote(self, node: Element) -> None:
         self.ensure_eol()
         self.body.append('@end quotation\n')
 
-    def visit_literal_block(self, node: Element) -> None:
+    def visit_literal_block(self, node: Optional[Element]) -> None:
```
tk0miya: Why do you think `node` is optional?

sphinx/cmd/quickstart.py
tk0miya: I think no developers pass `None` value expectedly. So let's go ahead :-)

sphinx/writers/texinfo.py
danieleades: because that's how it's called in `sphinx.writers.texinfo.TexinfoTranslator.visit_productionlist`

```python
    def visit_productionlist(self, node: Element) -> None:
        self.visit_literal_block(None)
        ...
```

sphinx/builders/latex/__init__.py
danieleades: i haven't changed the type signature of this method, i've only changed the way it's called to *comply* with the pre-existing type signature.
I can absolutely do what you're describing, by changing the signature from

```python
    def add_config_value(self, name: str, default: Any, rebuild: Union[bool, str],
                         types: Any = ()) -> None:
        ...
```

to 

```python
    def add_config_value(self, name: str, default: Any, rebuild: str,
                         types: Any = ()) -> None:
        ...
```

but this is a breaking change, and i think out of the scope of this PR.
Looking at the code, a literal boolean is perfectly acceptable, so it could be that the docstring is out of date.

there's also this line-

```
        .. versionchanged:: 0.6
           Changed *rebuild* from a simple boolean (equivalent to ``''`` or
           ``'env'``) to a string.  However, booleans are still accepted and
           converted internally.
```

so the docstring is consistent with the usage.

Is there a plan to deprecate the boolean arg?

danieleades:
looks like some type errors introduced outside this PR-

```
mypy sphinx
sphinx/util/logging.py: note: In function "convert_serializable":
sphinx/util/logging.py:81:13: error: "LogRecord" has no attribute "location"  [attr-defined]
sphinx/util/logging.py: note: In member "filter" of class "DisableWarningIsErrorFilter":
sphinx/util/logging.py:435:9: error: "LogRecord" has no attribute "skip_warningsiserror"  [attr-defined]
Found 2 errors in 1 file (checked 174 source files)
```

sphinx/cmd/quickstart.py
danieleades: it's called with `None` internally. If strict optionals were fully-implemented it would be easy to detect where this is the case and fix it, but right now it's a complex problem. Will revert this change. To be revisited in future

danieleades:
@tk0miya there's a number of functions with this pattern-

*from `sphinx.builders.latex.transforms`*

```python
    def get_footnote_by_reference(self, node: nodes.footnote_reference) -> nodes.footnote:
        docname = node['docname']
        for footnote in self.footnotes:
            if docname == footnote['docname'] and footnote['ids'][0] == node['refid']:
                return footnote

        return None
```

where the `None` case is never handled. To me, that means there's an assumption that it's unreachable, and the `return None` has just been added to keep mypy happy. mypy is *only* happy with this if `strict_optional` is not enabled.

I think an error should be raised in the `None` case. Pragmatically it could be directly in the function, but the purist in me says this should be delegated to the calling code.

as in,

```python
    def get_footnote_by_reference(self, node: nodes.footnote_reference) -> nodes.footnote:
        docname = node['docname']
        for footnote in self.footnotes:
            if docname == footnote['docname'] and footnote['ids'][0] == node['refid']:
                return footnote

        raise SomeError("Not found...")
```

vs

```python
    def get_footnote_by_reference(self, node: nodes.footnote_reference) -> Optional[nodes.footnote]:
        docname = node['docname']
        for footnote in self.footnotes:
            if docname == footnote['docname'] and footnote['ids'][0] == node['refid']:
                return footnote

        return None
```

what do you think?

there's another example of the pattern at `sphinx.transforms.post_transforms.images.get_conversion_rule`. Probably other examples too.

danieleades:
i'm running out of low-hanging fruit for these typing improvements, which is a good problem to have. Would be useful for someone with more knowledge of the code base to start making some targeted changes to specific modules... @tk0miya, @AA-Turner

AA-Turner:
> Would be useful for someone with more knowledge of the code base to start making some targeted changes to specific modules

I'm happy to support someone who's interested in making typing improvements, but it isn't a massive priority for me personally--I want to focus on performance, code quality generally, and stability--unfortunately I have very limited time so adding more to the list is hard.

A

danieleades:
> code quality generally,

i would place improved type annotations and linting firmly under this category ;)

AA-Turner:
In re raising exceptions vs returning None, I would vote for raising exceptions as the more 'Pythonic' pattern -- it is easy to forget to check for None at the calling site, whereas exceptions fail noisily if unhandled.

A

danieleades:
> In re raising exceptions vs returning None, I would vote for raising exceptions as the more 'Pythonic' pattern -- it is easy to forget to check for None at the calling site, whereas exceptions fail noisily if unhandled.
> 
> A

This is true in the current configuration. Ultimately I hope to get to a point where strict optional is enforced, and the you cannot forget to handle the None case at the call site

danieleades:
> In re raising exceptions vs returning None, I would vote for raising exceptions as the more 'Pythonic' pattern -- it is easy to forget to check for None at the calling site, whereas exceptions fail noisily if unhandled.
> 
> A

are you able to confirm absolutely that these methods MUST always return a value? I'm not familiar enough with the code to say for sure that's an invariant

danieleades:
i'm not planning more changes on this PR beyond those needed to resolve the open comments. I think that's enough typing changes for one PR

sphinx/builders/latex/__init__.py
danieleades: hi @tk0miya how would you like to proceed here? I can change all the uses of `bool` to `str` for this method if you prefer? Are there plans then to deprecate `bool`? If so, what's the most appropriate way to mark this?

something like this-

```python
import warnings
warnings.warn("blah blah blah", DeprecationWarning)
```
?

danieleades: @tk0miya i'm going to leave it for now and not address this in this PR. The reason being is i think it's a bigger change, and requires separate discussion.

Already this method doesn't actually accept any old string, it only accepts string literals of either `env` or `html`. Strictly speaking this should probably use an `Enum`, but the next best choice (retaining backwards compatibility) is to use `typing.Literal`.

danieleades:
@tk0miya, @AA-Turner - have had to make some formatting changes to get CI to pass, but this should now be good to go

AA-Turner (approved):

AA-Turner (approved):

AA-Turner:
Thanks Dan!

A

PR #10691, opened 2022-07-22
autosummary: Support documenting inherited attributes

### LINKED ISSUE #9884: sphinx autosummary: "failed to import Child.model" (inherited instance attribute)
### Describe the bug

I'm having trouble with sphinx and autosummary. Apparently, sphinx is unable to document inherited instance attributes. For some reason, the resulting error is an import error.

python code:
```
class Base:
    """Base class."""

    def __init__(self, **kwargs):
        """Init."""
        self.model = None
        """An attribute."""


class Child(Base):
    """Child class."""

    def __init__(self, **kwargs):
        """Init."""
        super().__init__()
        pass
```

rst documentation file:
```
.. autosummary::
    :toctree: generated

    mylib.Base
    mylib.Child
```
when I run sphinx-build with `make html` I get the following error:
```
.../new_docs/generated/mylib.Child.rst:24: WARNING: autosummary: failed to import Child.model
```

Notice that everything else is imported correctly, so this is not a path issue.
Is this a bug? According to #3257 it should be supported. I'm using sphinx-build 4.3.0.

### How to Reproduce

MWE to reproduce the error: [sphinx-demo.zip](https://github.com/sphinx-doc/sphinx/files/7581235/sphinx-demo.zip)
unzip the archive
```
$ cd docs
$ make html
```

### Expected behavior

inherited instance attributes are documented correctly and without warnings.

### Your project

[sphinx-demo.zip](https://github.com/sphinx-doc/sphinx/files/7581235/sphinx-demo.zip)

### Screenshots

_No response_

### OS

WSL 2 (Ubuntu)

### Python version

3.9.7

### Sphinx version

4.3.0

### Sphinx extensions

sphinx.ext.autosummary

### Extra tools

_No response_

### Additional context

_No response_

jenshnielsen:
For inherited attributes qualname will not match the name of the actual obj that the attribute belongs to

Subject: Autosummary don't filter attributes on qualname

### Feature or Bugfix
- Bugfix

### Purpose
Allow autosummary to document inherited instance attributes. 
The current implementation of import_ivar_by_name filters attributes if the name of the 
object that the attribute belongs to does not match the object being documented. However,
for inherited attributes this is not the case. For my simple project here https://github.com/jenshnielsen/debugproject
simply not filtering on qualname but only on the attribute name seems to resolve the issue. 
It is not clear to me if there are any unwanted sideeffects of this and we should filter on the list of 
qualnames for the object and all its super classes (if any)
 
If the super class is defined in another module the above is not sufficient so, try looking 
up instance attributes on modules in the MRO of the object

### Detail
- If this approach is acceptable, I still need to write tests for this.

### Relates
- Fixes https://github.com/sphinx-doc/sphinx/issues/9884


jenshnielsen:
This unfortunatly only works if the super class is in the same module so it does not make much sense to implement it like this

jenshnielsen:
Mypy failures are caused by https://github.com/python/typeshed/pull/8064 being part of mypy 0.971 now released

jenshnielsen:
I have been using this for a while and other people in #9884 reports that this solves the issue for them too. Would it be possible to have this reviewed?

jenshnielsen:
@AA-Turner @takaakiaoki would it be possible to have this reviewed?

jenshnielsen:
We have been using this to build our own docs since August and it solves a very real problem for me and for others, see comments in https://github.com/sphinx-doc/sphinx/issues/9884  Is there anything I can do to increase the change of this getting reviewed and eventually merged into master?

jjyao:
@AA-Turner @takaakiaoki could someone review this PR? Thanks!

IljaManakov:
Wow, this PR has been sitting here for 3 months. Can this please be merged at some point? The bug is really annoying.

AA-Turner:
Thank you @jenshnielsen!

A

Issue #10706, opened 2022-07-25
Quickstart `make -f` error: `path/to/Makefile is not a directory`


leogermond:
### Describe the bug

When using the quickstart project, calling the Makefile with the `--file` option (https://www.gnu.org/software/make/manual/make.html#index-_002d_002dfile) results in a weird error:
```
make -f /path/to/doc/Makefile -j 1 html

Application error:
Output directory (/path/to/doc/Makefile) is not a directory
```

### How to Reproduce

```
$ mkdir test_sphinx_makefile
$ cd test_sphinx_makefile
$ pip install sphinx
$ sphinx-quickstart
> separate source: n
> project name: test
> project author: test
> project release:
> project language: en
$ cd ..
$ make -f test_sphinx_makefile/Makefile html
```

```
Running Sphinx v5.1.0

Configuration error:
config directory doesn't contain a conf.py file (/home/germond/Labs/repro)
make: *** [test_sphinx_makefile/Makefile:20: test_sphinx_makefile/Makefile] Error 2
```

Error text can vary, the main issue is that the Makefile path is taken as if it were the target.

### Expected behavior

A html build should be run and finish successfully, as it would by calling `make -C test_sphinx_makefile html`

```
make: Entering directory '/home/germond/Labs/repro/test_sphinx_makefile'
Running Sphinx v5.1.0
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
reading sources... [100%] index                                                            
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] index                                                             
generating indices... genindex done
writing additional pages... search done
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in _build/html.
```

### Your project

github.com/adacore/cuda

### OS

Linux

### Python version

3.8.10

### Sphinx version

5.1.0

jfbu:
I think problems are to be expected from `make -f` if the make is not executed from same directory as where Makefile resides, correct way is `make -C` then. (cf discussion at #10707). At my locale `make -f Makefile html` works already.

Can you check if https://github.com/jfbu/sphinx/commit/a9c2c7f24608ec1d3ae689638d67cd9740e1c80f 
If makes (sic)  `make -f RenamedMakefile html` work for me.

[...]

PR #10707, opened 2022-07-25
[Quickstart] Allow calling the Makefile using `make -f path/to/Makefile <target>`

leogermond:
Subject: Allow calling the quickstart Makefile with `make -f`

### Feature or Bugfix

- Bugfix

### Purpose
- This PR modifies the default Makefile so that it can be called using `make -f`. Otherwise it would call it with a target that is the path to the Makefile itself, this would result in a weird "path/to/Makefile is not a directory" error.
- This PR does impacts people using the quickstart setup. It relies on the shell having the `pwd` command.

### Relates
- https://github.com/sphinx-doc/sphinx/issues/10706

### DIFF
--- a/sphinx/templates/quickstart/Makefile.new_t
+++ b/sphinx/templates/quickstart/Makefile.new_t
@@ -12,10 +12,13 @@ BUILDDIR      = {{ rbuilddir }}
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+# The Makefile itself, referenced either by a relative or absolute path
+MAKEFILE = Makefile $(shell pwd)/Makefile
+
+.PHONY: help $(MAKEFILE)
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile
+%: $(MAKEFILE)
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 


jfbu:
I do not understand:
- when moving to source (where `conf.py` and `Makefile` are), `make -f Makefile html` works for me with current Makefile, already
- if I rename Makefile to something else and try to use (in same directory) this new name via `make -f` then it does not work indeed, but this applies both to current Sphinx **and** to your branch, with sphinx-build complaining  about not knowing the builder, as you mention,
- if I issue `make -f` from *parent directory* (cf your #10706) I get the expected error about this directory not containing a `conf.py`.  And this applies whether or not using your branch. This seems to be what you report at #10706 but my understanding of https://www.gnu.org/software/make/manual/make.html#Makefile-Names is that this is to be expected.

Besides as you already mention in #10706, using `make -C` does work fine, from an arbitrary location, if argument of `-C` is the relative or absolute path to source directory.

My make is GNU Make 4.2.1

jfbu:
Hmm it seems this may solve it:
```diff
diff --git a/sphinx/templates/quickstart/Makefile.new_t b/sphinx/templates/quickstart/Makefile.new_t
index 1a527578b..5d5985d3b 100644
--- a/sphinx/templates/quickstart/Makefile.new_t
+++ b/sphinx/templates/quickstart/Makefile.new_t
@@ -12,10 +12,10 @@ BUILDDIR      = {{ rbuilddir }}
 help:
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+.PHONY: help
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
-%: Makefile
+.DEFAULT:
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```

For example renaming this to `MyMakefile` then `make -f MyMakefile html pseudoxml latexpdf`  works fine at my locale.

I got the hint from [this stackoverflow answer](https://stackoverflow.com/a/33018558) but some advice would be needed to make sure we do not break things.

jfbu:
cf https://www.gnu.org/software/make/manual/make.html#Last-Resort

tk0miya:
I guess this change will cause trouble for Windows users. So it's better not to use `pwd` command as possible. Could you check the advice from jfbu, please?

AA-Turner:
We want to move away from Makefile long-term, so I'm closing this. Thank you for your time, though.

A

PR #10708, opened 2022-07-25
incremental adoption of 'strict optional'

danieleades:
This PR adds a mypy 'whitelist' such that strict optional is applied to all modules that currently have no errors. The idea is that this whitelist shrinks over time, never gets any bigger, and can eventually be removed when `strict optional` has been fully-implemented. This means that no new strict optional type errors can be reintroduced to modules that have been completely cleared.

The downside is that it makes the mypy config file quite noisy until it's fully adopted

there's also still quite a lot of globbing in the whitelist. This could get more verbose before it gets less verbose when those globs are broken down further

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,9 +31,75 @@ ignore_missing_imports = True
 follow_imports = skip
 check_untyped_defs = True
 warn_unused_ignores = True
-strict_optional = False
+strict_optional = True
 no_implicit_optional = True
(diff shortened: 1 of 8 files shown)


AA-Turner (approved):

AA-Turner:
Thanks Dan!

tk0miya (approved):
LGTM! Thank you for your support :-)

Issue #10715, opened 2022-07-27
Regression in agogo theme


gvimont:
### Describe the bug

Pull request #10520 introduced a regression in the display of the Table of Content of the `agogo` theme.
The `agogo` sample can be seen here: https://sphinx-themes.org/sample-sites/default-agogo/

### How to Reproduce

https://sphinx-themes.org/sample-sites/default-agogo/

The background's width of the Table of Contents takes the whole page.

### Expected behavior

The background's width of the Table of Contents should be limited to the TOC width.

### Your project

Internal

### OS

Linux

### Python version

3.8

### Sphinx version

5.1.0

gvimont:
This is not a regression but #10520 make visible a bug in `agogo` theme.

mgeier:
I'm the author of #10520 and I must confess that I wasn't aware the `agogo` uses the wrong CSS class for its sidebar (`sidebar` instead of the `sphinxsidebar`) and it doesn't "correctly" style the `sidebar` directive.

So I agree that `agogo` was broken before but it is more broken after my change.

PR #10716, opened 2022-07-27
Fix/agogo

gvimont:

### Feature or Bugfix

- Bugfix

### Purpose
This improves the display of the sidebar of the agogo theme

### Relates
 #10715

### DIFF
--- a/sphinx/themes/agogo/static/agogo.css_t
+++ b/sphinx/themes/agogo/static/agogo.css_t
@@ -274,6 +274,8 @@ div.document ol {
 div.sidebar,
 aside.sidebar {
   font-size: .9em;
+  width: {{ theme_sidebarwidth }};
+  float: right;
 }
 
 div.sidebar a, aside.sidebar a, div.header a {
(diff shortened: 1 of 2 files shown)


tk0miya:
@mgeier Could you check this please?

mgeier:
This doesn't fully restore the CSS from before https://github.com/sphinx-doc/sphinx/pull/10520/files (e.g. the `rightsidebar` option is ignored).

I don't think the `pagewidth = 100%` change is good, because it removes the left/right margin of the page and the text touches the edge of the window.

The `sidebar` directive is still not properly styled, but I guess no `agogo` user uses that.

I'm not sure it's worth trying to fix `agogo`. Maybe it's best to revert #10520 and accept that it is broken?

Just out of curiosity: @gvimont Do you actually use the `agogo` theme? Or how did you notice the problem?

gvimont:
Yes, I'm actually using it (but I could probably use another theme).
I could change `pagewidth` to a better value.
As you said, another solution is to revert #10520, and the theme could be marked as deprecated if it is not maintained anymore.

mgeier:
I've created an alternative PR which reverts my previous change: #10787

AA-Turner:
I'll accept the revert commit.

A

PR #10717, opened 2022-07-28
search: support searching for (sub)titles

### LINKED ISSUE #10689: search: increase priority for full (sub)title match when searching
### Describe the bug

When searching with an exact match of a (sub)title, the result should be shown at the very top.
Example:
https://www.sphinx-doc.org/en/master/search.html?q=More+topics+to+be+covered

As seen the result is shown as 3rd:
Screenshot from 2022-07-21 17-26-35

It's defined in `quickstart.rst` as

```rst
More topics to be covered
-------------------------
```

### How to Reproduce

Use the current docs.

### Expected behavior

_No response_

### Your project

Sphinx

### Screenshots

_No response_

### OS

All

### Python version

any

### Sphinx version

any

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

marxin:
Collect all titles from all pages and utilize a full match
(case insensitive) in Search page.

Fixes: #10689

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,7 @@ Features added
 * #10755: linkcheck: Check the source URL of raw directives that use the ``url``
   option.
 * #10781: Allow :rst:role:`ref` role to be used with definitions and fields.
+* #10717: HTML Search: support searching for (sub)title.s
 
 Bugs fixed
 ----------
--- a/CHANGES
+++ b/CHANGES
@@ -18,7 +18,8 @@ Features added
 * #10755: linkcheck: Check the source URL of raw directives that use the ``url``
   option.
 * #10781: Allow :rst:role:`ref` role to be used with definitions and fields.
-* #10717: HTML Search: support searching for (sub)title.s
+* #10717: HTML Search: Increase priority for full title and 
+  subtitle matches in search results
 
 Bugs fixed
 ----------
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -59,7 +59,7 @@ if docutils.__version_info__[:2] <= (0, 17):
 
 # This is increased every time an environment attribute is added
 # or changed to properly invalidate pickle files.
-ENV_VERSION = 56
+ENV_VERSION = 57
 
 # config status
 CONFIG_OK = 1
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -183,6 +183,7 @@ class WordCollector(nodes.NodeVisitor):
     def __init__(self, document: nodes.document, lang: SearchLanguage) -> None:
         super().__init__(document)
         self.found_words: List[str] = []
+        self.found_titles: List[Tuple[str, str]] = []
         self.found_title_words: List[str] = []
         self.lang = lang
 
@@ -213,7 +214,10 @@ class WordCollector(nodes.NodeVisitor):
         elif isinstance(node, nodes.Text):
             self.found_words.extend(self.lang.split(node.astext()))
         elif isinstance(node, nodes.title):
-            self.found_title_words.extend(self.lang.split(node.astext()))
+            title = node.astext()
+            ids = node.parent['ids']
+            self.found_titles.append((title, ids[0] if ids else None))
+            self.found_title_words.extend(self.lang.split(title))
         elif isinstance(node, Element) and self.is_meta_keywords(node):
             keywords = node['content']
             keywords = [keyword.strip() for keyword in keywords.split(',')]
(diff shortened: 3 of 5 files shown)


AA-Turner:
Does this work for partial matches?

A

marxin:
> Does this work for partial matches?

Yes, I've just added partial match support.

AA-Turner:
Please add tests and a CHANGES entry.

A

marxin:
I've just added CHANGES, but for a working test, I would need an HTML inspection of rendered `search.html` page.
Do we have such tests?

marxin:
As an improvement, it should be possible to emit a link directly to the title in a page.
But I can't find a function that would give me for a `nodes.title` Node an anchor:
```
{'rawsource': 'Demo documentation', 'children': [<#text: 'Demo documentation'>], 'attributes': {'ids': [], 'classes': [], 'names': [], 'dupnames': [], 'backrefs': []}, 'tagname': 'title', 'parent': <section "demo documentation": <title...><compound...><paragraph...><substitution_defin ...>, '_document': <document: <substitution_definition "gol"...><section "demo documen ...>, 'source': '/home/marxin/Programming/texi2rst-generated/sphinx/demo/index.rst', 'line': 2}
```

Can you please help me?

jbms:
I think you may need to look at the parent `section` element to find the ids.

jbms:
Note: In https://github.com/jbms/sphinx-immaterial I have implemented something similar entirely client side, but there are a few differences:

- The sub-sections are parsed from the HTML document itself, while extracting snippets.
- If the search text is found within the page, the result link is to the nearest containing section.

This is the source code of my implementation, for reference:
https://github.com/jbms/sphinx-immaterial/blob/main/src/assets/javascripts/sphinx_search.ts

In general handling this when building the index is probably better, though given that the HTML must be parsed anyway to handle the snippets I'm not sure.

marxin:
> In general handling this when building the index is probably better, though given that the HTML must be parsed anyway to handle the snippets I'm not sure.

Yes, I do prefer the server side implementation and I'm still curious about the title links as mentioned in my previous comment. One should be able to get a link to them.

marxin:
> Something like `node.parent["names"]` should work? The anchor link is on the `docutils.nodes.section` node as I recall.

Yep, that almost works:
```patch
index bbb28c0b9..7916d26f0 100644
@@ -216,6 +216,8 @@ class WordCollector(nodes.NodeVisitor):
         elif isinstance(node, nodes.title):
             title = node.astext()
             self.found_titles.append(title)
+            print('node:', node)
+            print('node.parent[names]:', node.parent['names'])
             self.found_title_words.extend(self.lang.split(title))
         elif isinstance(node, Element) and self.is_meta_keywords(node):
             keywords = node['content']
```

emits something like:
```
node: <title>Comparison of GCC docs in Texinfo and Sphinx</title>
node.parent[names]: ['comparison of gcc docs in texinfo and sphinx']
node: <title>HTML output</title>
node.parent[names]: ['html output']
node: <title>Formatting</title>
node.parent[names]: ['formatting']
```

So the last missing piece is probably an escaping that will emit e.g. `comparison-of-gcc-docs-in-texinfo-and-sphinx`?

AA-Turner:
Ahh, can you try `["ids"]`?

A

marxin:
> Ahh, can you try `["ids"]`?

Works for me, added that.

marxin:
Can you please @AA-Turner review the pull request now?

marxin:
> The PR only shows 5 results, and doesn't highlight the title, whereas the current master shows the title, albeit as the third result.

Yeah, it's a fancy feature of `Read the Docs`, it must be a plug-in that is used for Sphinx docs.

Please compare it with another pull request:
https://sphinx--10807.org.readthedocs.build/en/10807/search.html?q=More+topics+to+be+covered

AA-Turner:
Rebased.

A

CHANGES
```diff
@@ -18,6 +18,7 @@ Features added
 * #10755: linkcheck: Check the source URL of raw directives that use the ``url``
   option.
 * #10781: Allow :rst:role:`ref` role to be used with definitions and fields.
+* #10717: HTML Search: support searching for (sub)title.s
```
```suggestion
* #10717: HTML Search: Increase priority for full title and 
  subtitle matches in search results
```

AA-Turner:
If running an incremental build, `searchindex.js` is only updated after loading it, so we need to bump the environment version.

A

AA-Turner (approved):

AA-Turner:
Thanks @marxin!

A

marxin:
Thanks for merging that!

PR #10718, opened 2022-07-28
Save search result score to HTML element for debugging

marxin:
It is handy for debugging of search results.


AA-Turner:
Rebased.

A

marxin:
@AA-Turner  Can you please merge the current `5.x` branch into `master` so that I can experiment with Sphinx docs search results?

AA-Turner:
Done, but I think `readthedocs`'s server side search is overriding `searchtools.js`.

If you build the docs locally you can run `python -m http.server --directory build/sphinx` and go to `localhost:8000/` to get around CORS, does that help?

A

marxin:
Great, that works fine, thanks!

PR #10725, opened 2022-07-29
use pyupgrade on codebase

danieleades:
this an experimental run of using `pyupgrade` on this code-base.

This will need to be reviewed carefully. I can imagine some whitelisting might be required, for example if a test is deliberately using an older python syntax for a test or similar, then pyupgrade would blindly update it to new syntax and invalidate the test.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -163,7 +163,7 @@ def linkify_issues_in_changelog(app, docname, source):
 
         def linkify(match):
             url = 'https://github.com/sphinx-doc/sphinx/issues/' + match[1]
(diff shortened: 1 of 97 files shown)


danieleades:
note that i've not included CI enforcement of pyupgrade changes, as this would require buying into the pre-commit ecosystem.

AA-Turner:
Thank you for the PR here Dan, though I'm going to defer this right now.

I do intend to run `pyupgrade` or something similar in the future, but I'd prefer more incremental change, and given the dual-long-lived-branch structure Sphinx uses at the moment I worry a little about managing git with this.

A

PR #10726, opened 2022-07-29
Fix #10723: LaTeX verbatimwithframe=false without effect since 5.1.0

### LINKED ISSUE #10723: LaTeX: 5.1.0 has made the 'sphinxsetup' ``verbatimwithframe=false`` become without effect
### Describe the bug

The frame shows despite using ``verbatimwithframe=false``.

### How to Reproduce

```

TEST
====

.. code-block::

   A

```

and 
```
# -- LateX --

latex_elements = {
    'sphinxsetup': 'verbatimwithframe=false',
}
```
in conf.py
the `make latexpdf`.

### Expected behavior

No frame with this option.

### Your project

See above

### Screenshots

Capture d’écran 2022-07-29 à 15 21 03

### OS

Mac

### Python version

3.10.4

### Sphinx version

5.1.0

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

jfbu:
Sorry... there was some internal refactoring during development phase of 5.1.0 LaTeX and this was a collateral.

In this PR I made internal changes which are not strictly necessary (usage of `boxes`  named dimension registers in place of `pre` named macros; this is why some `\relax`  can be removed) but they appeared logical from coding point of view.

The cause of the problem is that the Boolean was configured to set a macro `\spx@pre@border` to zero dimension but the actual drawing code used `\spx@boxes@border` instead.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* #10723: LaTeX: 5.1.0 has made the 'sphinxsetup' ``verbatimwithframe=false``
+  become without effect.
+
 Testing
 --------
 
--- a/sphinx/texinputs/sphinxlatexliterals.sty
+++ b/sphinx/texinputs/sphinxlatexliterals.sty
@@ -1,7 +1,7 @@
 %% LITERAL BLOCKS
 %
 % change this info string if making any custom modification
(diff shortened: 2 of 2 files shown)


jfbu:
CI/windows test_build_linkcheck.py fail for reasons unrelated to this PR.

PR #10257, opened 2022-03-11
[C++] Ensure consistent non-specialization template argument representation

jbms:
Subject: <short purpose of this pull request>

### Feature or Bugfix
- Bugfix

### Purpose
Previously, in certain cases the template arguments of non-specializations were retained, leading to incorrect merging of symbols.

PR #10729, opened 2022-07-29
[C++] Ensure consistent non-specialization template argument representation, rebased and updated

jakobandersen:
Rebased and updated version of #10257.
In addition the parameter pack parsing is fixed for one of the noted cases, while the other case is not valid C++ code anyway.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -21,6 +21,7 @@ Bugs fixed
 
 - #10257: C++, ensure consistent non-specialization template argument
   representation.
+- #10729: C++, fix parsing of certain non-type template parameter packs.
 
 Testing
 --------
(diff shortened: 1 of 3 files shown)


jbms:
Thanks!

jbms:
Per the TODO entry in my original PR you added support for parsing:

'int (X::*)(bool)...'

However, that does not appear to be valid C++ syntax.  I believe the correct syntax is:

`int (X::*... x)(bool)`


https://godbolt.org/z/r34v5r6vK

jakobandersen:
It depends on which location. In the test it was as a non-type template parameter:
```c++
struct X {};

template<int (X::*)(bool)...>
void f() {}
```
i.e., a pack of member function pointers. The addition was to make that parse in Sphinx.
The case of ``int (X::*... x)(bool)`` already works as a function parameter in Sphinx.

jbms:
Ah, thanks for the clarification.

PR #10738, opened 2022-07-31
Fix Python docstring type handling bug

ProGamerGov:
This PR ensures that NumPy style docstrings using  '`of`' are properly handled.

For example:

```
def func(t, x, y):
    """
    Args:

        t (int or float): Variable description.
        x (list of list of int, optional): Variable description.
        y (tensor or tuple of tensors): Variable description.
    """ 
```

According to the regex used here: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/domains/python.py#L381, only some of the variables in the above function's docstring will be properly handled. Anything with '`of`' in it will not be properly separated by the regex code. This PR fixes this issue. 

If this PR is on the wrong branch, let me know!

### Feature or Bugfix
- Bugfix

### Purpose
- Fixes docstring format handling bug.

### Detail

Below are some examples of failures made visible by using Intersphinx. Notably types using '`of`' are not properly hyperlinked.

### Relates
- N/A


AA-Turner:
Is the `of` delimiter endorsed by the google style guide?

A

ProGamerGov:
@AA-Turner Yeah, you can see an example here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html


```
        Args:
            param1 (str): Description of `param1`.
            param2 (:obj:`int`, optional): Description of `param2`. Multiple
                lines are supported.
            param3 (:obj:`list` of :obj:`str`): Description of `param3`.
```


I've noticed the issue in projects like [pytorch/captum](https://github.com/pytorch/captum).

AA-Turner:
The link you provided is from the now-archived `sphinxcontrib.napolean` project. Is there an example from the official google style guide rather than something third-party?

A

AA-Turner:
https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html seems to be the up-to-date copy.

sphinx/domains/python.py
```diff
@@ -380,7 +380,7 @@ def make_xrefs(self, rolename: str, domain: str, target: str,
                    innernode: Type[TextlikeNode] = nodes.emphasis,
                    contnode: Node = None, env: BuildEnvironment = None,
                    inliner: Inliner = None, location: Node = None) -> List[Node]:
-        delims = r'(\s*[\[\]\(\),](?:\s*or\s)?\s*|\s+or\s+|\s*\|\s*|\.\.\.)'
+        delims = r'(\s*[\[\]\(\),](?:\s*or|of\s)?\s*|\s+or|of\s+|\s*\|\s*|\.\.\.)'
```
AA-Turner:
```suggestion
        delims = r'(\s*[\[\]\(\),](?:\s*or|of\s)?\s*|\s+or|of\s+|\s*\|\s*|\.\.\.)'  # NoQA: E275, E501
```

ProGamerGov:
@AA-Turner So, actually I may have confused Google's style guide with NumPy's style guide. NumPy's official style guidelines show `list of int` being used here: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

Google's style guide though doesn't seem to explicitly mention it: https://google.github.io/styleguide/pyguide.html

AA-Turner:
`NumpyDocstring` is a superclass of `GoogleDocstring`. Please may you update the PR to override the regular expression only in `NumpyDocstring`, and add a link back to the style guide with an explanation as to why we are overriding the regular expression.

A

ProGamerGov:
@AA-Turner Yeah, we can do that. Though I'm not 100% sure of the regex works as intended. My usage of the or operator slightly removed the spaces around `or` in the output.

sphinx/domains/python.py
```diff
@@ -380,7 +380,7 @@ def make_xrefs(self, rolename: str, domain: str, target: str,
                    innernode: Type[TextlikeNode] = nodes.emphasis,
                    contnode: Node = None, env: BuildEnvironment = None,
                    inliner: Inliner = None, location: Node = None) -> List[Node]:
-        delims = r'(\s*[\[\]\(\),](?:\s*or\s)?\s*|\s+or\s+|\s*\|\s*|\.\.\.)'
+        delims = r'(\s*[\[\]\(\),](?:\s*or|of\s)?\s*|\s+or|of\s+|\s*\|\s*|\.\.\.)'  # NoQA: E275, E501
```
AA-Turner:
```suggestion
        delims = r'(\s*[\[\]\(\),](?:\s*or|of\s)?\s*|\s+or|of\s+|\s*\|\s*|\.\.\.)'
```

sphinx/domains/python.py
AA-Turner: Does this fix the whitespace problems?
```suggestion
        delims = r'(\s*[\[\]\(\),](?:\s*o[rf]\s)?\s*|\s+o[rf]\s+|\s*\|\s*|\.\.\.)'
```

ProGamerGov: Yeah that appears to solve the issue!

ProGamerGov:
Here's an example of image showing that cases using `type1 of type2` are now being hyperlinked correctly after the changes in this PR:

sphinx_bug_fixed

ProGamerGov:
@AA-Turner We can merge the PR now in its current form as it resolves the issue!

Example:

sphinx_bug_fixed


The `sphinx.ext.napolean` extension doesn't seem to require any changes to fully support this change.

AA-Turner:
Please add tests ensuring the behaviour doesn't regress in the future.

A

AA-Turner:
I have a slight concern over this change in that it encourages non-standard descriptions of typing -- for example if I had an object that was a `list[tuple[date, ...]]`, I think using that format in the docstring is much better than `list of tuple of int`.

I don't know if napoleon supports the type-annotation format though, and cannot test at the moment (on mobile).

A

ProGamerGov:
@AA-Turner Do you happen to know which testing file I should add the tests to?

Also, as far as I'm aware, napoleon handles `list[tuple[date, ...]]` and a quick test verifies this:

sphinx_example_1

AA-Turner:
See tests/test_ext_napoleon_docstring.py, around line 160 for an exemplar.

Thanks for confirming the 'proper' style works. Please also add a CHANGES entry and add yourself to AUTHORS.

A

ProGamerGov:
@AA-Turner Thanks! I've added myself to the AUTHORS file and added a new test for the changes. For the CHANGES file, I wasn't 100% sure what to say for the change, so let me know if I need to change it

ProGamerGov:
So, the tests don't currently seem to fail if the changing in this PR haven't been introduced. I think that part of the existing code may have already been capable of handing cases with '`of`', and thus the tests will always pass.

ProGamerGov:
@AA-Turner Testing now verifies that functionality introduced in this PR remains present. As the changes in this PR were on a single line in `sphinx/domains/python.py`, I added tests to `tests/test_domain_py.py`. These tests actually do fail if someone breaks the regex code changes introduced in this PR. I also added another test for the `or` deliminator as I couldn't find any examples of it being tested.

I think that this PR is finally ready for merging!

CHANGES
```diff
@@ -13,11 +13,17 @@ Deprecated
...
+* #10738: napoleon: Add support for docstring types using 'of', like
+  ``type of type``. Example: ``tuple of int``.
+
 Bugs fixed
 ----------
 
 * #10723: LaTeX: 5.1.0 has made the 'sphinxsetup' ``verbatimwithframe=false``
   become without effect.
+* #10738: py domain: Added testing for docstring types using 'or' in the
+   python domain, like ``type or type``. Previously this functionality wasn't
```
AA-Turner: This second entry isn't needed.

A

AA-Turner (changes requested):

AA-Turner:
Looks good otherwise, please could you also look at the flake8 failure and then ping me.

A

CHANGES
ProGamerGov: Okay, I've removed it

ProGamerGov:
@AA-Turner I've fixed the formatting for flake8

AA-Turner:
Thanks Ben!

A

Issue #8220, opened 2020-09-17
please document the rst-class and/or cssclass directive


zzzeek:
Restructured text has a [class directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#class) that lets you add a css class to an element.   This won't work in Sphinx because it conflicts with the default domain's "class", usually [py:class](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html?highlight=py%3Aclass#directive-py-class).

Sphinx has already found a solution for this, which is that it redefines docutils Class under [cssclass and rst-class](https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/directives/other.py#L390).

I found these by source-code diving and purposely breaking docutils "Class" and following where the import errors led me.   Otherwise, "rst-class" and "cssclass" are not [documented](https://www.sphinx-doc.org/en/master/search.html?q=rst-class) at all except for a passing, unlinked mention of rst-class [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=rst-class#directive-centered).

the "rst-class" directive should be documented at the very least in the [domains](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html?highlight=domains) chapter so that those of us wondering why we can't use RST `.. class::` know where to look for the alternative.

PR #10759, opened 2022-08-08
LaTeX: support for booktabs-style and zebra-striped tables

jfbu:
This transfers here development from #6666 to continue it and merge with #6671.

- [x] revamp initial work of @sephalon in #6666 for compatibility with current Sphinx LaTeX and to address initial issues.  This has been done originally at #6666 PR but I am moving it here and will adapt WIP status until the whole thing is done.
- [x] merge initial work of @sephalon from #6671
- [x] allow some tables to be styled the traditional way in documents globally using the booktabs option
- [x] allow tables to be styled the booktabs way in documents globally not using the booktabs option
- [x] as for booktabs style, think about locally styled tables; LaTeX aspects are not complicated based upon booktabs-styling with #6666 as merged here, the problem to obtain as light-weight mark-up as possible.  Simplest seems to go via classes but then we must used `latex-` prefixed names to dispel user expectation that it also applies to say html build.  <s>Actually using a container directive is possibly better as on LaTeX side it will wrap in a scope limiting environment. Maybe for first installment we can simply advice using some such recipe (container directive plus raw latex) which does not need extra code in Sphinx itself.</s> 

*finally I used classes with no `latex-` prefix even though they also may influence html output, but this sounds actually good*

removed longish paragraph which was in part obsoleted and too complex to edit; in brief all changes are opt-in except that

-  for merged cells in a single row the row colour (or the dedicated merge row colour) is now obeyed, so people who added manual usage of `\rowcolor` to their documents had only white merged cells but now the merged cell in single row (only) will react to row colour.

- for people who added manual usage of `\rowcolors` command to their document, the effect will be overwritten by this PR. Sphinx does not itself use `\rowcolors` due to the latter limitations (and the fact that not all people have xcolor).  Sphinx intervenes via its table templates and own code to influence the row colours, and this will overrule effect of manually inserted `\rowcolors`. I don't know if this means the PR should go to master rather than 5.x.


jfbu:
Memo: https://github.com/sphinx-doc/sphinx/pull/10759/commits/bb859c669679baebd8cc8d10c99382478c0d1647 probably closes #8220 as it makes mention of ``rst-class`` more prominent and links to it.

jfbu:
In order to illustrate the effect of this PR I am adding here snapshots from our own document, using:

```
latex_table_style = ['booktabs', 'colorrows']
```

and the default colours activated by ``'colorrows'``:

first table
Capture d’écran 2022-08-13 à 18 16 10

another
Capture d’écran 2022-08-13 à 18 16 55

Now a long table with a tabularcolumns directive.

Capture d’écran 2022-08-13 à 18 17 34

on next page:

Capture d’écran 2022-08-13 à 18 18 28

The effect was obtained via this 

```latex
>{\sphinxcolorblend{!95!red}\centering\noindent\bfseries\color{red}}\Y{.12}
```

as configuration for the third column.  It works here because my TeX installation has the LaTeX xcolor.sty.  On user devices with xcolor.sty not installed (I don't know actual situation but many years ago this was not automatic in Linux distros), the `\sphinxcolorblend`  would be silently ignored.

I was going to post an example of a complex grid table but it shows there is currently a problem to fix it. Here it is:

Capture d’écran 2022-08-13 à 18 42 54

This table uses a tabularcolumns (with no vertical lines):

```latex
.. tabularcolumns:: >{\sphinxcolorblend{!90!blue}}TTT>{\columncolor{yellow}\sphinxnorowcolor}TT
```

and illustrates various features.  The conf.py is:

```
latex_table_style = ['colorrows']

latex_elements = {
    'sphinxsetup': '''TableRowColorOdd=red!10,
                      TableRowColorEven= blue!10,
                      TableRowColorHeader=   {gray}{0.75},
                      TableMergeColorHeader=green,
                      TableMergeColorOdd=green!20,
                      TableMergeColorEven=green!10''',
}
```

where the colours have been chosen at random (and test the new acceptance of `\colorlet`-like syntax).  One sees in this table:

- merged cells remain with no background colour, except for the case of single-row multi-column cells.  They then use the special Merge color, if set.
- in the header each row uses the same colouring style
- the fifth column displays the normal colours corresponding to ``'sphinxsetup'``
- the fourth column obeys yellow column color thanks to `\sphinxnorowcolor`
- the first column uses `\sphinxcolorblend`, whose effect is ignored in merged cells.

The table has a bug because the cell3-2-par2 should not end up with  a `|`.  This originates in a `latex.py`  `\multicolumn` which has no yet taken into account that this `table.colsep`  is `''`.  I will fix that later.  When using `booktabs` I noticed another problem that there is some offset in the parity of row colouring, which was supposed to be worked around already.  In brief, I have to fix that before I consider my August Sphinx time is over...

```
grid table
----------

.. tabularcolumns:: >{\sphinxcolorblend{!90!blue}}TTT>{\columncolor{yellow}\sphinxnorowcolor}TT

+---------+---------+---------+---+--------+
| header1 | header2-3         | a | normal |
+---------+---------+---------+---+--------+
| header1 | header2-3         | a | normal |
+---------+---------+---------+---+--------+
| header1 | header2-3         | a | normal |
+=========+=========+=========+===+========+
| cell1-1 | cell1-2 | cell1-3     | normal |
+---------+         +---------+---+--------+
| cell2-1 |         | cell2-3     | normal |
+         +---------+---------+---+--------+
|         | cell3-2-par1      | d | normal |
+---------+                   +---+--------+
| cell4-1 | cell3-2-par2      | e | normal |
+---------+---------+---------+---+--------+
| cell5-1 single-row merged   | f | normal |
+---------+---------+---------+---+--------+
```

jfbu:
The latest commits in this PR fix some issues with merged cells which were not yet documented as such: when a tabularcolumns directive uses no `|`,  nevertheless merged cells would sometimes use them; also, horizontal widths was not computed correctly (but the discrepancy was not really visible).  On the other hand if the tabularcolumns directive contains a `|`, nothing changes, i.e. merged cells are bordered too vertically, even though in complex column spec one could imagine only some columns are bordered.  In theory, the LaTeX writer could analyse where the `|` are located in the colspec and use this to influence its handling of merged cells, but this should be topic of another PR: this one only takes care of no `|` at all in the tabularcolumns colspec.

To be honest, merged cells have required already a few years back tremendous work by Sphinx team, but perhaps we should not so much worry about them, as in fact I guess few Sphinx users paint their tables as grid tables, which allow merged cells... We had to write our own LaTeX code because the support by LaTeX package of such situation is dire. I.e. it may look fine once you have spend one hour manually adjusting your various usages of multirow and multicolumn (and in fact don't forget e.g. Sphinx uses tabulary and tabulary's multicolumn does not allow multi-paragraphs.  Similarly for many many years multirow in LaTeX did not allow multiple paragraphs.  In recent years some basic improvements happend, e.g. it seems very recently multirow can cope with multi-paragraphs cells, but I have not tested, and the problem is that these problems should have been solved by LaTeX 20 years ago, but they have not been solved.

jfbu:
again one of those transient test_build_linkcheck.py failures... https://github.com/sphinx-doc/sphinx/runs/7821615851?check_suite_focus=true which are not related to this PR...

jfbu:
There is a quite strong deficiency of LaTeX's `\cline`. It shifts a bit the text upwards for cells underneath it. One can exaggerate the effect by setting `\arrayrulewidth` to a high value like here `3pt`:
Capture d’écran 2022-08-15 à 20 11 04
(see the two leftmost red "circles")

The other "feature" one sees clearly (the red "circles" on the right) is that colour panels cover it.  This is mentioned in `colortbl` documentation, but the recommendation to use `\hhline` seems impractical to use due to the complex syntax of `\hhline` which means complex changes to the Sphinx latex.py code related to tables and the table templates which for me is currently definitely a no-go.  The `\cmidrule` of booktabs do not have this problem and we could let `\cline` act like them, but they look appropriate only in tables with no vertical lines, as they will create gaps in these vertical lines!

You may think such basic problems have been solved years ago in LaTeX... but alas no. Tables in LaTeX are a mess. (perhaps I will suppress that comment, but I need ventilation from time to time).

jfbu:
The problems with LaTeX `\cline` being hidden behing colour panels and reducing vertical space are solved.  The figure uses a high value of `\arrayrulewidth` to better check all is correct.

Capture d’écran 2022-08-16 à 16 05 06

But I have become aware of other problems in certain grid tables with merged cells, which are yet to be analyzed but originate in some LaTeX code in this PR prior to latest commit.  The grid tables used in our testing (the one above is an extension) do not show the problems I have noticed in some other tables.

jfbu:
I have added in my clone at https://github.com/jfbu/sphinx/commit/abab0bf1e43d8ce809bc62452f58930b628731a3 an extra commit in order for booktabs + colorrows to give tables with no "white gaps".  Here are some snapshots of what it produces in our own document:

Capture d’écran 2022-08-17 à 21 20 19


Capture d’écran 2022-08-17 à 21 18 57

This one has a problem because the small gap underneath the top rule is filled with the default colour for headers (slighly darker gray than the one for odd rows).  To fix this is a bit tough on macro side because it will need `\sphinxtoprule` to peek ahead to see if next token is `\endhead`, `\endfirsthead`,  or `\sphinxtableatstartofbodyhook` which is only way to realize there is no header and thus adapt the colour.  Easier would be to add mark-up to the table to signal to LaTeX macros that it has no header.

Capture d’écran 2022-08-17 à 21 17 54

Also this one has a problem, because we see that the column color tint does not appy.  For it to apply we would have to use an altogether different approach than using the `\specialrule` of booktabs.

Capture d’écran 2022-08-17 à 21 17 10

Besides globally on zooming on pictures I think  I am seeing anti-aliasing which might be a sign there is some extra space I did not take into account, andI would need to dig into `\specialrule` code from booktabs which I don't really feel like doing.

So this is bit problematic (when I started writing this comment and took the snapshots I was not aware of all problems).

jfbu:
In one of the longtable snapshots of https://github.com/sphinx-doc/sphinx/pull/10759#issuecomment-1214189134, one sees that header colour is applied to the table continuation hints.

Maybe it is better they do not?  This is achieved by adding `\sphinxnorowcolor` at two locations in the longtable template:
```diff
diff --git a/sphinx/templates/latex/longtable.tex_t b/sphinx/templates/latex/longtable.tex_t
index 9f722a713..906ea81e1 100644
--- a/sphinx/templates/latex/longtable.tex_t
+++ b/sphinx/templates/latex/longtable.tex_t
@@ -45,7 +45,7 @@
 \endfirsthead
 
 \multicolumn{<%= table.colcount %>}{c}%
-{\makebox[0pt]{\sphinxtablecontinued{\tablename\ \thetable{} \textendash{} <%= _('continued from previous page') %>}}}\\
+{\sphinxnorowcolor\makebox[0pt]{\sphinxtablecontinued{\tablename\ \thetable{} \textendash{} <%= _('continued from previous page') %>}}}\\
 \sphinxtoprule
 <%= ''.join(table.header) -%>
 <%- if table.header -%>
@@ -54,7 +54,7 @@
 \endhead
 
 \sphinxbottomrule
-\multicolumn{<%= table.colcount %>}{r}{\makebox[0pt][r]{\sphinxtablecontinued{<%= _('continues on next page') %>}}}\\
+\multicolumn{<%= table.colcount %>}{r}{\sphinxnorowcolor\makebox[0pt][r]{\sphinxtablecontinued{<%= _('continues on next page') %>}}}\\
 \endfoot
 
 \endlastfoot
```
and I could add an option but I would rather not.  I could perhaps replace the added token `\sphinxnorowcolor` by an alias `\sphinxnorowcolorforcontinuationhints` so a user can still redefine it to `\relax` in preamble or via container class to obtain same looks as in https://github.com/sphinx-doc/sphinx/pull/10759#issuecomment-1214189134. 

With the above extra patch on obtains this for top of page:
Capture d’écran 2022-08-19 à 08 31 47

and this for bottom of page:
Capture d’écran 2022-08-19 à 08 33 57

I guess this removal of the header row colour from contination hints should be default?

jfbu:
I have now merged the separate work on colorizing booktabs gaps mentioned in earlier comments so that it is done per default and tables look as in https://github.com/sphinx-doc/sphinx/pull/10759#issuecomment-1218424121.  I consider now it is better indeed to colorize these gaps if booktabs+colorrows is chosen by user conf.py, and the longtable aspects should not be considered blocking.  However I needed to make one extra addition to the longtable template.

This brings to an end (? I was already saying PR was ready 10 days ago...) my work on this... as my Sphinx+LaTeX time will now be more scarce in coming months I wanted to finish a coherent whole.

jfbu:
Two short notes:
- in the Python code I used `colsep` to designate either`''` or `'|'`. The name comes from initial #6666. It looks natural as shorthand for "column separator". But LaTeX uses `\tabcolsep` for the dimension register denoting whitespace (on each side of `|`). So the name is a bit surprising to LaTeX user. I didn't look for a better one.
- recently there is a new LaTeX package https://github.com/lvjr/tabularray which looks promising.  It may in future help to replace all three of tabular, tabulary and longtable perhaps (I notice it uses optionally internally varwidth package as Sphinx does via explicit mark-up). But this can not be until a few years because it is undergoing active development (first release in May 2021, see its [ChangeLog](https://github.com/lvjr/tabularray/wiki/ChangeLog), with some breaking changes). By the way it uses a `colsep` key in its table config with the meaning afaict of interfacing to some equivalent of `\tabcolsep` illustrating the terminology problem mentioned in previous item.

jfbu:
@AA-Turner Although the various comments I added when I was actively working on this may give a frightening impression, this PR is completed and I am not planning any further change... it adds opt-in features only (some are tested in our own document via a slightly modified conf.py).  We can wait to see if @tk0miya has comments.  For 6.0.0 release I would like to make some configuration the default, so that tables look better in LaTeX.

sephalon:
@jfbu Thank you very much for your hard work to bring these features upstream! :-)

PR #10766, opened 2022-08-13
Viewcode: Support import paths that differ from the directory structure

ProGamerGov:
If the module import path used by users does not match the directory / file structure, `import_lib`'s `import_module` function cannot find it.

Below is a basic example of where this failure occurs:

```
package/
    `__init__.py` # import package.subpackage1 as subpackage1
    subpackage1/
        `__init__.py` # from package.subpackage1.subpackage2 import module_1
        subpackge2/
            `__init__.py`
            module_1.py
```

Using the above directory / file structure, users import `module1` via:

```
import package.subpackage1.module1
```

Sphinx is generally able to handle this sort of structure, except for the viewcode extension that uses the `get_full_modname` function from here: https://github.com/sphinx-doc/sphinx/blob/5.x/sphinx/util/__init__.py

That function specifically causes an error with the above example when `importlib`'s `import_module` is run:

```
from importlib import import_module
module = import_module("package.subpackage1.module1") # Fails
# ModuleNotFoundError: No module named 'package.subpackage1.module1'
```

The exact path to the module however has no error:
```
from importlib import import_module
module = import_module("package.subpackage1.subpackage2.module1") # Works
```

### Feature / Bugfix

This PR is my attempt at fixing the above `import_module` error by locating the exact path to the module, which is then used with `import_module`.

Currently this issue can be seen on projects where the viewcode extension is enabled, yet no source code link is created. Example: https://captum.ai/api/utilities.html#visualization

My solution utilizes `importlib`'s `__import__` [function](https://docs.python.org/3/library/importlib.html#importlib.__import__) to find the directory module path to the object.


ProGamerGov:
@AA-Turner, @tk0miya let me know if I need to change anything for this PR to be merged

tests/roots/test-ext-viewcode-find-package/main_package/__init__.py
```diff
@@ -0,0 +1 @@
+import main_package.subpackage as subpackage  # NOQA
```
danieleades: don't use blanket `NOQA`s. I'm guessing this is an F401?

ProGamerGov: @danieleades I was just following the other viewcode tests as examples:


```
from .mod1 import Class1, func1  # NOQA
from .mod2 import Class2, func2  # NOQA
```

https://github.com/sphinx-doc/sphinx/blob/5.x/tests/roots/test-ext-viewcode/spam/__init__.py


```
from .submodule import Class1, func1  # NOQA
```

https://github.com/sphinx-doc/sphinx/blob/5.x/tests/roots/test-ext-viewcode-find/not_a_package/__init__.py


But I can change all of them if you want me to.

danieleades: please. a blanket noqa may inadvertently squash other lints. Lint suppression should be as targeted as possible.

ProGamerGov: @danieleades Okay, I've replaced all 4 instances of `# NOQA` with `# noqa: F401`.

tests/roots/test-ext-viewcode-find-package/main_package/subpackage/__init__.py
```diff
@@ -1,3 +1,3 @@
-from main_package.subpackage._subpackage2 import submodule  # NOQA
+from main_package.subpackage._subpackage2 import submodule  # noqa: F401
 
-__all__ = ["submodule"]
\ No newline at end of file
+__all__ = ["submodule"]
```
danieleades: do you still need noqa: F401 fi you're explicitly exporting names with `__all__ = [...]`? I would hope flake8 would be smart enough to recognise that

tests/roots/test-ext-viewcode-find-package/conf.py
```diff
@@ -0,0 +1,24 @@
...
+import sys
+
+source_dir = os.path.abspath('.')
+if source_dir not in sys.path:
+    sys.path.insert(0, source_dir)
+extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode']
+exclude_patterns = ['_build']
+
+
+if 'test_linkcode' in tags:  # NOQA
```
danieleades: what error are you suppressing here?

tests/roots/test-ext-viewcode-find-package/main_package/subpackage/__init__.py
ProGamerGov: No I don't think so. I'll remove that line

tests/roots/test-ext-viewcode-find-package/conf.py
ProGamerGov: That line was in the other test as well: https://github.com/sphinx-doc/sphinx/blob/5.x/tests/roots/test-ext-viewcode/conf.py

I've removed both instances, and there doesn't appear to be any errors as a result. Maybe it was required at some point in the past, but it doesn't appear to be required now.

ProGamerGov: @danieleades That `# NOQA` was for some old error that is no longer present

ProGamerGov:
@danieleades The conflicts have been resolved, so this PR is ready for merging

ProGamerGov:
I think that I'm going to remake this PR

PR #10770, opened 2022-08-14
Further improve type annotations, reduce mypy whitelist

danieleades:


sphinx/writers/texinfo.py
```diff
@@ -115,7 +115,7 @@ class TexinfoWriter(writers.Writer):
 
     settings_defaults: Dict = {}
 
-    output: str = None
+    output: Optional[str] = None  # type: ignore[assignment]
```
danieleades, line 118: false positive. Probably an error in the associated docutils type stubs (i haven't checked).

sphinx/builders/latex/transforms.py
```diff
@@ -94,13 +94,16 @@ def expand_show_urls(self) -> None:
     def get_docname_for_node(self, node: Node) -> str:
         while node:
             if isinstance(node, nodes.document):
-                return self.env.path2doc(node['source'])
+                doc = self.env.path2doc(node['source'])
+                if doc is None:
+                    raise ValueError("unreachable code")
+                return doc
```
danieleades: @tk0miya can you please check you're happy with this approach here? is `path2doc` genuinely *never* `None` in this situation?

perhaps there's a better way to refactor this. this doesn't feel very 'clean' as is

danieleades:
i'm not planning further changes in this PR @AA-Turner, @tk0miya

this will need careful review, as these are larger changes than previous 'typing' PRs (a good chunk of the codebase now has strict optional checking though!)

AA-Turner:
Thanks @danieleades, as always. At some point please add yourself to AUTHORS, you've now made a series of fairly substantial typing-related improvements.

A

PR #10774, opened 2022-08-17
doc: add note on unsupported include directive for source-read event content modification

QSchulz:
Subject: doc: add note on unsupported include directive for source-read event content modification

### Purpose
- Files included with the `include` directive have their content available
in the source-read event but modifications will not make it to the
output, so let's document it.
### Relates
- #10678 

### Note
Does this need to be sent to master and/or 4.x branch too?


QSchulz:
Or we could add a note stating that there is a limitation for included files until version X.Y.

PR #10798, opened 2022-09-04
Remove `sphinx.builders.gettext` module from whitelist

danieleades:
shrink the mypy 'strict optional' white list


AA-Turner (approved):

AA-Turner:
Thanks Dan!

A

PR #10811, opened 2022-09-08
Autosummary: Extend ``__all__`` members to template rendering

### LINKED ISSUE #10809: Autosummary: ``autosummary_ignore_module_all`` does not work as documented
### Describe the bug

As per documentation :

> Note that if an imported member is listed in `__all__`, it will be documented regardless of the value of autosummary_imported_members.

So the way I understand it is that if I have a small package like this :

```
my_package/
    __init__.py
    my_class.py
    my_module/
        my_submodule.py
```

with the following files :

`my_class.py`
```python
class myClass:
    """" my Class docstring """"
    def my_class_method(self, a):
        """ method docstring """
        return
```

`my_submodule.py`
```python
def f():
    """ f docstring """
    return
```

`__init__.py`
```python
from .my_class import myClass
from . import my_module

__all__ = ["myClass", "my_module"]
```

and a rst file like this:

```rst
An example of recursive module documentation

.. autosummary::
    :toctree: generated
    :recursive:

    my_package
```

assume the option `autosummary_ignore_module_all` is set to `False` in `conf.py`, the autosummary should only consider `myClass` and `my_module` to be part of the package `my_package`. The `generated` folder should normally look like this:

```
generated
    my_package.myClass
    my_packge.my_module
```

But actually, it looks like this :

```
generated
    my_package.my_class
```

Meaning it ignored the imported elements, even though they were in the `__all__` and it did take the `my_module` package even though it was not in the `__all__`

### How to Reproduce

```
$ git clone git@github.com:ClementPinard/sphinx_autosummary_bug_example.git
$ cd sphinx_autosummary_bug_example.gt
$ pip install sphinx
$ sphinx-build -b html docs docs/_build
$ # see docs/generated/ folder
```

### Expected behavior

generated folder should have the files

```
my_package.rst
my_package.my_module.rst
my_patckage.my_module.my_submodule.rst
my_package.myClass.rst
```

### Your project

https://github.com/ClementPinard/sphinx_autosummary_bug_example

### Screenshots

_No response_

### OS

Ubuntu 20.04

### Python version

3.9

### Sphinx version

5.1

### Sphinx extensions

sphinx.ext.autodoc, sphinx.ext.autosummary

ClementPinard:
Fixes #10809

* the option `autosummary_ignore_module_all` when set to `False` adds members to module's members entry that will be used for autodoc, but otherwise it ignores it. As such, if a class is available in the `__all__`, it won't be generated.
* This commit aims at extending the `__all__` handling not only to members, but also to corresponding attribute types (functions, classes, exceptions, modules)
* In short, the `imported_members` option is set to `True` if the object has `__all__` member and autosummary is set to have `autosummary_ignore_module_all` set to False

See an example repo where the behaviour is changed : https://github.com/ClementPinard/sphinx_autosummary_bug_example

this PR is (at least initially) just for illustration and discussion purpose. tests where not checked

### Feature or Bugfix

- Feature
- Bugfix

A mix of both

### Detail
- See issue #10809


ClementPinard:
Hi, is this improvement proposition considered ? I have created this PR as a draft to spark some discussion about changing autosummary behaviour, but otherwise it could be reviewed as is.

I really think the current behaviour is not one users would expect when they use the recursive option.

AA-Turner:
Please may you add a test of the new behaviour, and an entry to the changelog (`CHANGES`)?

A

ClementPinard:
I think I have found a solution that works.

If `__all__` is given , first look for imported modules (possibly renamed) and then look for the rest.

Additionally, there is now a skip list for get_modules, to avoid tryinbg to get documentation from a module that was overwritten in the package namespace

This will cause this kind of code to not crash with sphinx :

(file `package/__init__.py`)

```python
from .a import a
```

autosummary will now know that `package.a` is a function and not a module anymore.

NB: Yes this does happen even though I would personally consider it an antipattern. Some JAVA inspired python code exist out there with only a class in the module and a class the same name of the module and then an import similar to the aforementionned one in the parent package `__init__.py`. Whether we should try to be compliant with this or not is debatable but it it's possible to do that in python, so it will not disappear.

On a more genral note, I have seen that the difference between private members and public members is the leading underscore (see https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/generate.py#L283).

Without a proper `__all__` list it makes sense, but if it is given (and the `autosummary_ignore_module_all` is False), I think it should prevail as it is by definition the public API. It's out of the scope of this PR, but I can add this modification as well if you want.

AA-Turner:
Thank you @ClementPinard!

A

PR #10813, opened 2022-09-09
More precise typing for `BuildEnvironment.domains`

AA-Turner:
@danieleades if you've time for review on this I'd appreciate it--originally I was using imports from `typing` rather than `typing_extensions`, but they refused to work on my computer. I'm not sure if it is my installation that's broken, or if this is a weird quirk.

### Feature or Bugfix
- Refactoring

A

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -79,6 +79,60 @@ versioning_conditions: Dict[str, Union[bool, Callable]] = {
     'text': is_translatable,
 }
 
+if TYPE_CHECKING:
+    from collections.abc import MutableMapping
+
+    from typing_extensions import Literal, overload
+
+    from sphinx.domains.c import CDomain
+    from sphinx.domains.changeset import ChangeSetDomain
+    from sphinx.domains.citation import CitationDomain
+    from sphinx.domains.cpp import CPPDomain
+    from sphinx.domains.index import IndexDomain
+    from sphinx.domains.javascript import JavaScriptDomain
+    from sphinx.domains.math import MathDomain
+    from sphinx.domains.python import PythonDomain
+    from sphinx.domains.rst import ReSTDomain
+    from sphinx.domains.std import StandardDomain
+    from sphinx.ext.duration import DurationDomain
+    from sphinx.ext.todo import TodoDomain
+
+    class DomainsType(MutableMapping[str, Domain]):
+        @overload
+        def __getitem__(self, key: Literal["c"]) -> CDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["cpp"]) -> CPPDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["changeset"]) -> ChangeSetDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["citation"]) -> CitationDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["index"]) -> IndexDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["js"]) -> JavaScriptDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["math"]) -> MathDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["py"]) -> PythonDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["rst"]) -> ReSTDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["std"]) -> StandardDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["duration"]) -> DurationDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["todo"]) -> TodoDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: str) -> Domain: ...  # NoQA: E704
+        def __getitem__(self, key): raise NotImplementedError  # NoQA: E704
+        def __setitem__(self, key, value): raise NotImplementedError  # NoQA: E704
+        def __delitem__(self, key): raise NotImplementedError  # NoQA: E704
+        def __iter__(self): raise NotImplementedError  # NoQA: E704
+        def __len__(self): raise NotImplementedError  # NoQA: E704
+
+else:
+    DomainsType = dict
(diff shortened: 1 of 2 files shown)


sphinx/environment/__init__.py
```diff
@@ -79,6 +79,60 @@
...
+        @overload
+        def __getitem__(self, key: str) -> Domain: ...  # NoQA: E704
+        def __getitem__(self, key): raise NotImplementedError  # NoQA: E704
+        def __setitem__(self, key, value): raise NotImplementedError  # NoQA: E704
+        def __delitem__(self, key): raise NotImplementedError  # NoQA: E704
+        def __iter__(self): raise NotImplementedError  # NoQA: E704
+        def __len__(self): raise NotImplementedError  # NoQA: E704
+
+else:
+    DomainsType = dict
```
danieleades, line 134: type annotations? what type of dict is this?

sphinx/environment/__init__.py
```diff
@@ -79,6 +79,60 @@
...
+        def __getitem__(self, key: Literal["duration"]) -> DurationDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: Literal["todo"]) -> TodoDomain: ...  # NoQA: E704
+        @overload
+        def __getitem__(self, key: str) -> Domain: ...  # NoQA: E704
+        def __getitem__(self, key): raise NotImplementedError  # NoQA: E704
+        def __setitem__(self, key, value): raise NotImplementedError  # NoQA: E704
+        def __delitem__(self, key): raise NotImplementedError  # NoQA: E704
+        def __iter__(self): raise NotImplementedError  # NoQA: E704
+        def __len__(self): raise NotImplementedError  # NoQA: E704
```
danieleades, line 131: this approach doesn't seem right to me. What are you trying to achieve? you might be looking for `TypedDict`, rather than `MutableMapping`?

something like-

```python
class DomainsType(TypedDict):
    c: CDomain
    cpp: CPPDomain
    changeset: ChangeSetDomain
    ...
```

can you give me some more context here?

danieleades (changes requested):

danieleades:
> originally I was using imports from `typing` rather than `typing_extensions`, but they refused to work on my computer. I'm not sure if it is my installation that's broken, or if this is a weird quirk.

what interpreter version are you using locally?

python typing features are generally added to `typing-extensions` *first* and then stabilised into `typing`, so depending on the features you're using and the interpreter version, you might find that the features are available in typing-extensions but not in typing. The differences between build environments are likely down to different interpreter versions.

AA-Turner:
`env.domains` holds references to all loaded domain objects, but each domain has different properties, attributes, etc. I want mypy to recognise that if I type `env.domains['std'].labels` for example that I am accessing the legal `labels` property on the `StandardDomain` object rather than the missing property on the `Domain` type. 

A `TypedDict` won't work as I understand it as users can register their own domains (see `recipie` in the docs or `duration` in `sphinx.ext.duation`)--for these cases I still want the type to be `str->Domain`.

In short--a mapping type of `str->Domain`, with certain magic keys recognised as returning subclasses of `Domain`.

A

sphinx/environment/__init__.py
AA-Turner: This is the `TYPE_CHECKING is False` path, i.e. at runtime--we want to instantiate a normal dict. `mypy` was complaining if I wrote `self.domains = {}` rather than `self.domains = DomainsType()` below.

AA-Turner:
> what interpreter version are you using locally?
> 
> python typing features are generally added to typing-extensions first and then stabilised into typing

Yep, I've contributed (albeit minorly) to TE. Running Python 3.10.6 from conda-forge, mypy 0.971. As I understand it, `mypy` vendors its own copy of typeshed, which is why I thought it might be some glitch as opposed to loading TE dynamically, but `mypy` may have changed its architecture.

A

danieleades:
Are the 'magic keys' always set?

AA-Turner:
From the above, all are always set apart from `duration` and `todo`. They're initialised in this loop:

https://github.com/AA-Turner/sphinx/blob/c215f522e955304ebf50829b329e7d4712eeb91f/sphinx/environment/__init__.py#L269-L271

A

sphinx/environment/__init__.py
danieleades: > This is the `TYPE_CHECKING is False` path, i.e. at runtime--we want to instantiate a normal dict. `mypy` was complaining if I wrote `self.domains = {}` rather than `self.domains = DomainsType()` below.

what i mean is you haven't fully annotated the type. `dict` (or `Dict`) is equivalent to `dict[Any, Any]`, which is not what you want. Unfortunately the type annotations are used this way throughout the codebase...

can you do something like

```python
DomainsType = dict[str, Type[Domain]]
```

?

danieleades: oh i see what you mean about the *runtime* path

sphinx/environment/__init__.py
danieleades: on review i'm not sure there is a better solution than what you're proposing right now. There are feature requests to allow additional keys on `TypedDict`, but there are soundness issues with the proposal.

I guess that what this really highlights is that if this codebase had used strict type annotations from the start, it would probably not be implemented this way. I presume you're avoiding breaking changes here. If you *could* make breaking changes you would just promote the magic keys to concrete properties

AA-Turner: > on review i'm not sure there is a better solution than what you're proposing right now. There are feature requests to allow additional keys on `TypedDict`, but there are soundness issues with the proposal.

Thank you for your review, this is what I feared having tried a `Protocol` and `TypedDict` approach and having gotten nowhere.

> I guess that what this really highlights is that if this codebase had used strict type annotations from the start, it would probably not be implemented this way. 

Challenging given the first commit here was around seven years before [PEP 484](https://peps.python.org/pep-0484/)!

>I presume you're avoiding breaking changes here. If you _could_ make breaking changes you would just promote the magic keys to concrete properties

I'll consider the attributes approach for Sphinx 6.0/7.0, though it has drawbacks in that the user-registered domains will be second-class in some way.

A

danieleades: > the user-registered domains will be second-class in some way.

they *are* second-class, in that they are dynamically registered, instead of statically registered. It makes sense that these would be handled differently, as this is a pretty fundamental difference. I would probably go further and make this explicit-

```python
class DomainsType:
    c: CDomain
    cpp: CPPDomain
    changeset: ChangeSetDomain
    ...
    custom: dict[str, Domain]
```

i think the only way to avoid this difference would be to somehow require users to subclass `DomainsType` to add their own domains

PR #10816, opened 2022-09-12
imgmath: Allow embedding images in HTML as base64

jschueller:
A new flag imgmath_embed allows to include LaTeX output images as inline base64 into the HTML instead of separate png/svg files, using:
 ``
This allows to avoid uploading many small files for continuous integration artifacts.
The latex output file is removed after encoding.

inspired from https://github.com/sphinx-contrib/imageembed


doc/usage/extensions/math.rst
```diff
@@ -133,6 +133,10 @@ are built:
    elements (cf the `dvisvgm FAQ`_). This option is used only when
    ``imgmath_image_format`` is ``'svg'``.
 
+.. confval:: imgmath_embed
+
+   Default: ``False``.  If true, inline LaTeX output as HTML base64 images
+   instead of separate png/svg files.
```
AA-Turner, line 140:
```suggestion

   .. versionadded:: 5.2

```

sphinx/ext/imgmath.py
```diff
@@ -354,5 +370,6 @@ def setup(app: Sphinx) -> Dict[str, Any]:
     app.add_config_value('imgmath_latex_preamble', '', 'html')
     app.add_config_value('imgmath_add_tooltips', True, 'html')
     app.add_config_value('imgmath_font_size', 12, 'html')
+    app.add_config_value('imgmath_embed', False, 'html')
```
AA-Turner:
```suggestion
    app.add_config_value('imgmath_embed', False, 'html', [bool])
```

AA-Turner:
@jschueller please don't force-push, I'll squash on merge.

A

jschueller:
oh, ok

AA-Turner:
Please may you add some tests for the new behaviour, and an entry in CHANGES?

A

AA-Turner:
As a design point, I'd suggest splitting `render_math()` into two functions, at the `# Move generated image on tempdir to build dir` line.

We would always call the function that creates the temporary image, but the flag would control if we called the `move_rendered_image_from_temp_dir_to_build_dir()` function or the `read_rendered_image_and_return_base_64_encoded_string()` function (names to be improved!).

This then means we always know the mime type and don't have to use the guess functions, and don't need to remove any images, as they're in temp directories anyway.

A

jschueller:
@AA-Turner the move is now in a separate function get_img_src, but I dont have as many functions as suggested,
could you give more details on how to split the functions if its not ok for you ?

AA-Turner:
Thanks Julien!

A

jschueller:
thanks!

jschueller:
@AA-Turner it seems it does not work in parallel, any idea why ?
```
File "/usr/lib/python3.10/site-packages/sphinx/util/parallel.py", line 125, in _join_one
    raise SphinxParallelError(*result)
sphinx.errors.SphinxParallelError: FileNotFoundError: [Errno 2] No such file or directory: '/home/devel/project/build/python/src/sphinx_build/html/_images/math/dd25924f1cd0f37c1985116c759fb5c0dbfae2bb.svg'
```

AA-Turner:
Could you open a new issue please, or a PR adding a failing test for parallel building?

A

jschueller:
ok, see #10869

PR #10839, opened 2022-09-20
set up html base class to reduce duplication

danieleades:
WORK IN PROGRESS

there's an enormous amount of duplication between the HTML and HTML5 writers. This PR attempts to remove the duplication (DRY)

### DIFF
--- a/sphinx/writers/_html_base.py
+++ b/sphinx/writers/_html_base.py
@@ -1,30 +1,37 @@
 """docutils writers handling Sphinx' custom nodes."""
 
-import os
-import posixpath
 import re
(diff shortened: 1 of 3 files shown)


danieleades:
before i spend any more time on this, any ideas where i'm introducing the error @AA-Turner, @tk0miya ?

danieleades:
> I'd rather deprecate and remove HTML 4 support and fix this over the long term, see #10843.
> 
> Post deprecation HTML 4 support will be effectively frozen, so we can factor out into a 'do-not-touch' module.
> 
> A

no problem, i'll close this in that case

PR #10840, opened 2022-09-20
extend option directive syntax

marxin:
The motivation of pull request is a documentation syntax we use in GCC docs (working on the porting):

```rst
.. option:: -Wimplicit-fallthrough

  :option:`-Wimplicit-fallthrough` is the same as :option:`-Wimplicit-fallthrough=3`
  and :option:`-Wno-implicit-fallthrough` is the same as
  :option:`-Wimplicit-fallthrough=0`.

.. option:: -Wno-implicit-fallthrough

  Default setting; overrides :option:`-Wimplicit-fallthrough`.

.. option:: -Wimplicit-fallthrough#{n}

  Warn when a switch case falls through.  For example:
  ...

.. option:: -Wimplicit-fallthrough#0

   Disable it altogether.

.. option:: -Wimplicit-fallthrough#1

   Level 1

.. option:: -Wimplicit-fallthrough#2

   Level 2

.. option:: -Wimplicit-fallthrough#3

   Level 3

  ...

Try :option:`-Wimplicit-fallthrough=` or :option:`-Wimplicit-fallthrough` or :option:`-Wimplicit-fallthrough=n`
```

As seen, we basically need to distinguish:
1. `-Woption` and `-Woption=` directives
2. `-Woption=value1`, `-Woption=value2`, ...

Where it would be handy to reference to any of the aforementioned versions. I came up with a new syntax `#` that is a non-breaking version of `=`. Is the suggestion something the community would accept?

Example:
https://splichal.eu/tmp/demo2/html/

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -780,7 +780,9 @@ class StandardDomain(Domain):
             self.labels[name] = docname, labelid, sectname
 
     def add_program_option(self, program: str, name: str, docname: str, labelid: str) -> None:
-        self.progoptions[program, name] = (docname, labelid)
+        # prefer first command option entry
+        if (program, name) not in self.progoptions:
+            self.progoptions[program, name] = (docname, labelid)
 
     def build_reference_node(self, fromdocname: str, builder: "Builder", docname: str,
                              labelid: str, sectname: str, rolename: str, **options: Any
@@ -941,6 +943,10 @@ class StandardDomain(Domain):
         progname = node.get('std:program')
         target = target.strip()
         docname, labelid = self.progoptions.get((progname, target), ('', ''))
+        # for :option:`-foo=bar` search for -foo option directive
+        if not docname and '=' in target:
+            target2 = target[:target.find('=')]
+            docname, labelid = self.progoptions.get((progname, target2), ('', ''))
         if not docname:
             commands = []
             while ws_re.search(target):
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,9 @@ Deprecated
 Features added
 --------------
 
+* #10840: One can cross-reference including an option value: ``:option:`--module=foobar```.
+  Patch by Martin Liska.
+
 Bugs fixed
 ----------
 
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -1775,6 +1775,10 @@ There is a set of directives allowing documenting command-line programs:
    referenceable by :rst:role:`option` (in the example case, you'd use something
    like ``:option:`dest_dir```, ``:option:`-m```, or ``:option:`--module```).
 
+   .. versionchanged:: 5.3
+
+      One can cross-reference including an option value: ``:option:`--module=foobar```.
+
    Use :confval:`option_emphasise_placeholders` for parsing of
    "variable part" of a literal text (similarly to the :rst:role:`samp` role).
 
--- a/tests/roots/test-root/objects.txt
+++ b/tests/roots/test-root/objects.txt
@@ -204,6 +204,19 @@ Link to :option:`hg commit` and :option:`git commit -p`.
 
 Foo bar.
 
+Test repeated option directive.
+
+.. option:: -mapi
+
+  My API.
+
+.. option:: -mapi=secret
+
+  My secret API.
+
+Reference the first option :option:`-mapi=secret`.
+
+
 User markup
 ===========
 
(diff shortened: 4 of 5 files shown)


marxin:
A bit of context: as you likely know I'm working on transition of th GCC compiler documentation from Texinfo to Sphinx and this feature would help you reference `option` much better. As seen, we commonly document option values.

Is the Sphinx community interested in the feature? I can add more tests and document it.
What do you think @AA-Turner, @tk0miya ?

AA-Turner:
My concern is about making the option directive too complicated to maintain for quite a niche usecase.

Is it possible in your conversion work to implement a custom option directive that works for GCC, or a flag to the standard option directive that hooks into your extra processing? If there are constraints that GCC must use only vanilla Sphinx then those routes aren't useable, clearly.

Is the following style usable?


```rest
.. option:: -Wimplicit-fallthrough

   Warn when a switch case falls through. Legal values are:
  
   - 0: Disable it altogether.
   - 1: Level 1
   - 2: Level 2
   - 3: Level 3

   :option:`-Wimplicit-fallthrough` is the same as :option:`-Wimplicit-fallthrough=3`
   and :option:`-Wno-implicit-fallthrough` is the same as
   :option:`-Wimplicit-fallthrough=0`.

.. option:: -Wno-implicit-fallthrough

  Default setting; overrides :option:`-Wimplicit-fallthrough`.

Try :option:`-Wimplicit-fallthrough=` or :option:`-Wimplicit-fallthrough` or :option:`-Wimplicit-fallthrough=n`
```

This would only require changing the option *role* to ignore `=` and anything after `=`.

A

marxin:
> My concern is about making the option directive too complicated to maintain for quite a niche usecase.

I do share the concert about the maintainability of the functionality.

> 
> Is it possible in your conversion work to implement a custom option directive that works for GCC, or a flag to the standard option directive that hooks into your extra processing? If there are constraints that GCC must use only vanilla Sphinx then those routes aren't useable, clearly.

We would like to support only vanilla Sphinx.

> 
> Is the following style usable?
> 
> ```
> .. option:: -Wimplicit-fallthrough
> 
>    Warn when a switch case falls through. Legal values are:
>   
>    - 0: Disable it altogether.
>    - 1: Level 1
>    - 2: Level 2
>    - 3: Level 3
> 
>    :option:`-Wimplicit-fallthrough` is the same as :option:`-Wimplicit-fallthrough=3`
>    and :option:`-Wno-implicit-fallthrough` is the same as
>    :option:`-Wimplicit-fallthrough=0`.
> 
> .. option:: -Wno-implicit-fallthrough
> 
>   Default setting; overrides :option:`-Wimplicit-fallthrough`.
> 
> Try :option:`-Wimplicit-fallthrough=` or :option:`-Wimplicit-fallthrough` or :option:`-Wimplicit-fallthrough=n`
> ```

Yes, that's a feasible solution for us, I think.

> 
> This would only require changing the option _role_ to ignore `=` and anything after `=`.

Yep. That would be a pretty small change, let me push it.

sphinx/domains/std.py
```diff
@@ -780,7 +780,9 @@ def process_doc(self, env: "BuildEnvironment", docname: str, document: nodes.doc
             self.labels[name] = docname, labelid, sectname
 
     def add_program_option(self, program: str, name: str, docname: str, labelid: str) -> None:
-        self.progoptions[program, name] = (docname, labelid)
+        # prefer first command option entry
+        if (program, name) not in self.progoptions:
+            self.progoptions[program, name] = (docname, labelid)
```
AA-Turner, line 785: Why is this bit needed?

marxin: If you a duplicate option directive like:

```rst
.. option:: -Wimplicit-fallthrough

   Warn when a switch case falls through. Legal values are:

.. option:: -Wimplicit-fallthrough=3

   Test
```

Then all cross-refs (e.g. `:option:`-Wimplicit-fallthrough=n` would point to the first directive and not the second one. Hope it's desired behavior.

AA-Turner: Thanks for clarifying

AA-Turner:
Please merge 5.x to fix mypy

AA-Turner:
Please also add a CHANGES entry

marxin:
Both done now.

AA-Turner:
I don't like the GitHub UI where you're not prompted to review the commit message if your last merge was a "rebase and merge"... for posterity I was going to use:

```
Allow cross referencing options with values (#10840)

This change means that text following `=` is ignored when searching for a
corresponding option directive to an option cross reference role.
```

A

marxin:
I like the wording improvement, please commit it.

PR #10842, opened 2022-09-20
Fix `get_terminal_width` on Windows

### LINKED ISSUE #10841: Progress output on Windows
### Describe the bug

Progress output on Windows assumes an 80 columns width terminal, no matter of the actual console width. This causes previous output after column 80 not be overwritten, see screenshot below.

### How to Reproduce

```
$ git clone https://github.com/matplotlib/matplotlib
$ cd matplotlib
$ pip install -r requirements/doc/doc-requirements.txt
$ cd docs
$ make html
```
see above screenshot for output

### Expected behavior

clear rest of line on Windows too

### Your project

https://github.com/matplotlib/matplotlib

### Screenshots

image

### OS

Windows

### Python version

3.9.12

### Sphinx version

5.1.1

### Sphinx extensions

sphinx-gallery

### Extra tools

_No response_

### Additional context

_No response_

StefRe:
### Feature or Bugfix

- Bugfix
- Refactoring

### Purpose
[`shutil.get_terminal_size`](https://docs.python.org/3/library/shutil.html#shutil.get_terminal_size) is available since Python 3.3: it first checks for a COLUMNS environment variable, then queries the terminal and eventually falls back to 80 columns if none of the above works. So it's a drop-in replacement for `get_terminal_width()` that works on Windows too to get the actual terminal width.

### Relates
- closes #10841

### DIFF
--- a/sphinx/util/console.py
+++ b/sphinx/util/console.py
@@ -2,6 +2,7 @@
 
 import os
 import re
+import shutil
 import sys
 from typing import Dict, Pattern
 
@@ -22,21 +23,8 @@ def terminal_safe(s: str) -> str:
 
 
 def get_terminal_width() -> int:
-    """Borrowed from the py lib."""
-    if sys.platform == "win32":
-        # For static typing, as fcntl & termios never exist on Windows.
-        return int(os.environ.get('COLUMNS', 80)) - 1
-    try:
-        import fcntl
-        import struct
-        import termios
-        call = fcntl.ioctl(0, termios.TIOCGWINSZ, struct.pack('hhhh', 0, 0, 0, 0))
-        height, width = struct.unpack('hhhh', call)[:2]
-        terminal_width = width
-    except Exception:
-        # FALLBACK
-        terminal_width = int(os.environ.get('COLUMNS', 80)) - 1
-    return terminal_width
+    """Get number of columns of the terminal."""
+    return shutil.get_terminal_size().columns - 1
 
 
 _tw: int = get_terminal_width()
--- a/sphinx/util/console.py
+++ b/sphinx/util/console.py
@@ -23,7 +23,7 @@ def terminal_safe(s: str) -> str:
 
 
 def get_terminal_width() -> int:
-    """Get number of columns of the terminal."""
+    """Return the width of the terminal in columns."""
     return shutil.get_terminal_size().columns - 1
 
 


sphinx/util/console.py
```diff
@@ -22,21 +23,8 @@ def terminal_safe(s: str) -> str:
...
-        import termios
-        call = fcntl.ioctl(0, termios.TIOCGWINSZ, struct.pack('hhhh', 0, 0, 0, 0))
-        height, width = struct.unpack('hhhh', call)[:2]
-        terminal_width = width
-    except Exception:
-        # FALLBACK
-        terminal_width = int(os.environ.get('COLUMNS', 80)) - 1
-    return terminal_width
+    """Get number of columns of the terminal."""
+    return shutil.get_terminal_size().columns - 1
```
AA-Turner: Previously we only checked COLUMNS if querying the terminal directly failed--perhaps we should maintain that order of operations?:

```suggestion
    """Return the width of the terminal window.

    The width of the terminal connected to sys.__stdout__ is obtained via
    os.get_terminal_size. If the terminal size cannot be determined, the
    environment variable COLUMNS is used if defined. Otherwise, fallback value
    is returned.
    """
    try:
        return os.get_terminal_size(sys.__stdout__.fileno()).columns - 1
    except (AttributeError, ValueError, OSError):
        # FALLBACK
        return int(os.environ.get('COLUMNS', 80)) - 1
```

A

StefRe: Good catch. On the other hand - the order of precedence implemented by `shutil.get_terminal_size` is not without reason. If the user wants (for whatever reason, admittedly I can't think of any right now) to restrict the output width she or he can do it by defining `COLUMNS`. In the current program flow, there is no possibility to do so. So it's a question of backward compatibility for Linux systems (Windows was fixed at 80) versus the principle of least surprise (if someone goes to the trouble to define the environment variable one can assume it's done for a good reason). So I'm neural on it.

sphinx/util/console.py
```diff
@@ -22,21 +23,8 @@ def terminal_safe(s: str) -> str:
...
-        import struct
-        import termios
-        call = fcntl.ioctl(0, termios.TIOCGWINSZ, struct.pack('hhhh', 0, 0, 0, 0))
-        height, width = struct.unpack('hhhh', call)[:2]
-        terminal_width = width
-    except Exception:
-        # FALLBACK
-        terminal_width = int(os.environ.get('COLUMNS', 80)) - 1
-    return terminal_width
+    """Get number of columns of the terminal."""
```
AA-Turner:
```suggestion
    """Return the width of the terminal in columns."""
```

AA-Turner:
Thanks Steffen!

A

PR #10844, opened 2022-09-21
Migrate configs for isort, mypy, and pytest into pyproject.toml

cclauss:
### Feature or Bugfix
- Refactoring

### Purpose
Migrate the configs into `pyproject.toml` using [`ini2toml`](https://pypi.org/project/ini2toml) to do the file conversion and running [`validate-pyproject`](https://github.com/abravalheri/validate-pyproject) to validate the results.  Skip any configuration that is not yet PEP 621 compatible.


AA-Turner (changes requested):
Thanks for the PR @cclauss -- looks OK for merge but please revert all the dependency sorting changes -- I see some merit for a later PR but I'd like to keep this just as configuration migration.

Coping also @danieleades for notice on the mypy configuration migration.

A

AA-Turner (approved):

PR #10824, opened 2022-09-13
Use both title and subtitle name in search results

marxin:
Current status:
https://sphinx--10814.org.readthedocs.build/en/10814/search.html?q=More+topics+to+be+covered

with this pull request:
https://sphinx--10824.org.readthedocs.build/en/10814/search.html?q=More+topics+to+be+covered

PR #10848, opened 2022-09-22
search: use both title and subtitle only if different

marxin:
Fixes: #10824.

Fixes problem seen here:
https://sphinx--10824.org.readthedocs.build/en/10824/search.html?q=getting+started

where one sees entries like 'Getting Started > Getting Started', fixes as:
https://sphinx--10848.org.readthedocs.build/en/10848/search.html?q=getting+started


marxin:
@AA-Turner one more improvement for the search capability, please review

AA-Turner (approved):

PR #10849, opened 2022-09-22
shrink 'any-generics' whitelist for the 'deprecation' module

danieleades:


sphinx/deprecation.py
```diff
@@ -52,10 +52,10 @@ def __getattr__(self, name: str) -> Any:
         return self._objects[name]
 
 
-class DeprecatedDict(dict):
+class DeprecatedDict(Dict[str, Any]):
```
AA-Turner, line 55: Won't this make import time longer / increase the MRO for DeprecatedDict?

A

danieleades: that i do not know.

the other approach to make the typings consistent is to widen the methods to accept values other than `str` as keys

PR #10854, opened 2022-09-23
Move to localStorage for search highlighting

### LINKED ISSUE #10833: Clear ?highlight= query parameter from the URL

The `?highlight=...` query parameter that is added to the URL when performing a search is very annoying. To be clear, the highlighting itself is nice, but it's annoying that it's in the URL, because that means if you copy the URL, the highlight parameter comes along with it. But this is almost never what I want. Consider the following very common workflow:

1. Someone asks a question on some forum, and I know the answer is exists in some (Sphinx) document.
2. I go to the relevant site in question and search for something that will take me to the relevant page.
3. I click through to the page and copy the URL and paste it as an answer to the person.

After doing the above, the URL link as given will include the highlighting from the search. But that highlighting is actually irrelevant to the person reading the URL. It was only there for the original search term I used to find the given page.

For example, I might end up with a URL like this https://docs.sympy.org/latest/modules/simplify/simplify.html?highlight=nsimplify#sympy.simplify.simplify.nsimplify. Note that if I'm not paying close attention I might not even notice the highlighting. Here the search term happens to match the function name, but it could be anything, even something completely irrelevant. Suppose for instance I searched for one function but later decided to copy the link for another.

Of course, if you are aware of this, and also remember to do it, you can fix it by manually removing the `?highlight=` query parameter from the URL. Some Sphinx themes such as Furo and now [pydata sphinx theme](https://github.com/pydata/pydata-sphinx-theme/pull/925) include buttons that you can click to do this for you ("hide search matches" under the search bar in Furo), but it's still an extra step you have to do every time, and you have to actually remember to do it.

To contrast, I can't think of an instance when I have ever intentionally wanted to share a URL with text highlighting included. And even if I did, there is the text fragment anchor for that (see https://github.com/sphinx-doc/sphinx/issues/10201), and anyway I don't see why I would want to share a URL that highlights the text specifically from a search term.

I've actually seen Google search results in the past that included a random `?highlight=` in them, presumably because Google learned about the link from someone copy-pasting a page from a search result. 

I would suggest automatically clearing the `?highlight=` query parameter from the URL. The highlighting should stay there (it would obviously go away after doing a refresh, but I think that's fine). To minimize disruption, I would suggest only clearing the query parameter after submitting a search. Manually loading a URL with a `?highlight=` parameter should leave the URL intact as it is. That way people can continue to manually use the `highlight` parameter to highlight text, in case anyone is actually doing that.

I'm not really a Javascript developer so I don't know for certain whether this is possible, but I'm assuming it is because I've seen similar behaviors with other types of things on other websites. 

Right now the best alternative is the "hide search terms" button on some themes, but this is theme specific, and requires a specific action from the user. It also both clears the parameter and the highlighting, whereas my suggested solution would only clear the parameter and keep the highlighting (until a refresh).

**Additional context**

This was discussed at https://github.com/pydata/pydata-sphinx-theme/pull/925 but it was suggested to open an issue here.

AA-Turner:
### Feature or Bugfix
- Refactoring

Closes #10833

A

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -32,6 +32,8 @@ Features added
 * #6692: HTML Search: Include explicit :rst:dir:`index` directive index entries
   in the search index and search results. Patch by Adam Turner
 * #10816: imgmath: Allow embedding images in HTML as base64
+* #10854: HTML Search: Use browser localstorage for highlight control, stop
+  storing highlight parameters in URL query strings. Patch by Adam Turner.
 
 Bugs fixed
 ----------
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -48,6 +48,7 @@ epub_post_files = [('usage/installation.xhtml', 'Installing Sphinx'),
                    ('develop.xhtml', 'Sphinx development')]
 epub_exclude_files = ['_static/opensearch.xml', '_static/doctools.js',
                       '_static/jquery.js', '_static/searchtools.js',
+                      '_static/sphinx_highlight.js',
                       '_static/underscore.js', '_static/basic.css',
                       '_static/language_data.js',
                       'search.html', '_static/websupport.js']
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -15,8 +15,10 @@ module.exports = function(config) {
 
     // list of files / patterns to load in the browser
     files: [
+      'tests/js/documentation_options.js',
       'sphinx/themes/basic/static/doctools.js',
       'sphinx/themes/basic/static/searchtools.js',
+      'sphinx/themes/basic/static/sphinx_highlight.js',
       'tests/js/*.js'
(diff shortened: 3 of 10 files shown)


mgeier:
In https://github.com/sphinx-doc/sphinx/issues/10833#issuecomment-1249037667, @asmeurer said:

> With my suggestion, you would still be able to manually append `?highlight=` to the end of a URL.

This doesn't seem to be possible with this PR, or am I doing something wrong?

AA-Turner:
https://sphinx--10854.org.readthedocs.build/en/10854/usage/theming.html?highlight=navigation_with_keys#builtin-themes

@mgeier `?highlight=...` support restored, with the query parameter cleared on page load.

A

PR #10866, opened 2022-09-26
shrink 'any-generics' whitelist for the 'locale' module

danieleades:
- remove blanket 'ignore' statements
- add typings for `Callable`

this module is still a bit of a nightmare, as far as typing is concerned. Any comments or input welcome

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -197,7 +197,6 @@ module = [
     "sphinx.ext.*",
     "sphinx.highlighting",
     "sphinx.jinja2glue",
-    "sphinx.locale",
     "sphinx.registry",
     "sphinx.roles",
     "sphinx.search.*",
--- a/sphinx/locale/__init__.py
+++ b/sphinx/locale/__init__.py
@@ -16,7 +16,7 @@ class _TranslationProxy:
     """
     __slots__ = ('_func', '_args')
 
-    def __new__(cls, func: Callable, *args: str) -> '_TranslationProxy':
+    def __new__(cls, func: Callable[..., str], *args: str) -> '_TranslationProxy':
         if not args:
             # not called with "function" and "arguments", but a plain string
             return str(func)  # type: ignore[return-value]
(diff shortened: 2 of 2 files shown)


AA-Turner:
```python
class _TranslationProxy:
    """
    Class for proxy strings from gettext translations. This is a helper for the
    lazy_* functions from this module.

    The proxy implementation attempts to be as complete as possible, so that
    the lazy objects should mostly work as expected, for example for sorting.
    """
    __slots__ = ('_func', '_args')

    def __new__(cls, func: Callable, *args: str) -> "_TranslationProxy":
        if not args:
            # not called with "function" and "arguments", but a plain string
            return str(func)  # type: ignore[return-value]
        return object.__new__(cls)

    def __getnewargs__(self) -> Tuple[str]:
        return (self._func,) + self._args  # type: ignore

    def __init__(self, func: Callable, *args: str) -> None:
        self._func = func
        self._args = args

    def __str__(self) -> str:
        return str(self._func(*self._args))

    def __dir__(self) -> List[str]:
        return dir(str)

    def __getattr__(self, name: str) -> Any:
        return getattr(self.__str__(), name)

    def __getstate__(self) -> Tuple[Callable, Tuple[str, ...]]:
        return self._func, self._args

    def __setstate__(self, tup: Tuple[Callable, Tuple[str]]) -> None:
        self._func, self._args = tup

    def __copy__(self) -> "_TranslationProxy":
        return _TranslationProxy(self._func, *self._args)

    def __repr__(self) -> str:
        try:
            return 'i' + repr(str(self.__str__()))
        except Exception:
            return f'<{self.__class__.__name__} broken>'

    def __add__(self, other: str) -> str:
        return self.__str__() + other

    def __radd__(self, other: str) -> str:
        return other + self.__str__()

    def __mod__(self, other: str) -> str:
        return self.__str__() % other

    def __rmod__(self, other: str) -> str:
        return other % self.__str__()

    def __mul__(self, other: Any) -> str:
        return self.__str__() * other

    def __rmul__(self, other: Any) -> str:
        return other * self.__str__()

    def __hash__(self):
        return hash(self.__str__())

    def __eq__(self, other):
        return self.__str__() == other

    def __lt__(self, string):
        return self.__str__() < string

    def __contains__(self, char):
        return char in self.__str__()

    def __len__(self):
        return len(self.__str__())

    def __getitem__(self, index):
        return self.__str__()[index]
```

This is perhaps better from a typing perspective for `_TranslationProxy`.

A

AA-Turner (approved):

PR #10868, opened 2022-09-26
Shrink 'any-generics' whitelist for the `pycode` module

danieleades:


sphinx/pycode/__init__.py
```diff
@@ -124,7 +124,7 @@ def for_module(cls, modname: str) -> "ModuleAnalyzer":
         cls.cache['module', modname] = obj
         return obj
 
-    def __init__(self, source: IO, modname: str, srcname: str) -> None:
+    def __init__(self, source: IO[Any], modname: str, srcname: str) -> None:
```
AA-Turner: I think this is the better API, and then put the `.read()` calls in the class methods.

```suggestion
    def __init__(self, source: str, modname: str, srcname: str) -> None:
```

danieleades: what would that look like inside the method?

AA-Turner: `self.code = source.read()` becomes `self.code = source`. In the `for_file` and `for_string` methods we would pass a string to the class constructor `cls`.

A

danieleades: oh i see, i thought you meant do the `read` in the `__init__` method. Yep, i agree with you. Will update.

PR #10876, opened 2022-09-28
Use a module level constant as best practice

danieleades:
fixes the following flake8 warnings -

```
./sphinx/locale/__init__.py:165:23: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
./sphinx/locale/__init__.py:165:36: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
```

### DIFF
--- a/sphinx/locale/__init__.py
+++ b/sphinx/locale/__init__.py
@@ -159,11 +159,11 @@ def setlocale(category: int, value: Union[str, Iterable[str], None] = None) -> N
         pass
 
 
-LOCALE_DIR = path.abspath(path.dirname(__file__))
+_LOCALE_DIR = path.abspath(path.dirname(__file__))
 
 
 def init_console(
-    locale_dir: str = LOCALE_DIR,
+    locale_dir: str = _LOCALE_DIR,
     catalog: str = 'sphinx',
 ) -> Tuple[NullTranslations, bool]:
     """Initialize locale for console.
--- a/sphinx/locale/__init__.py
+++ b/sphinx/locale/__init__.py
@@ -4,7 +4,7 @@ import gettext
 import locale
 from collections import UserString, defaultdict
 from gettext import NullTranslations
-from os import path
+from pathlib import Path
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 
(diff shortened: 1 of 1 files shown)


AA-Turner:
The idea here is to reduce import time so I think just `# NoQA: B008` would be better? I don't believe the warning applies as the path won't change at all.

A

danieleades:
> The idea here is to reduce import time so I think just `# NoQA: B008` would be better? I don't believe the warning applies as the path won't change at all.
> 
> A

I'd probably make the argument that the difference in import time is so trivial as to be not worth it, and that defining the 'magic variable' as a module level constant is better practice. These kinds of micro-optimisations are rarely worth it, unless you have benchmarks which show otherwise.

If you MUST have the import time improvement, you could perhaps do this-

```python
def init_console(
    locale_dir: Optional[Path] = None,
    catalog: str = 'sphinx',
) -> Tuple[NullTranslations, bool]:
    """Initialize locale for console.

    .. versionadded:: 1.8
    """
    try:
        # encoding is ignored
        language, _ = locale.getlocale(locale.LC_MESSAGES)
    except AttributeError:
        # LC_MESSAGES is not always defined. Fallback to the default language
        # in case it is not.
        language = None
    locale_directory = locale_dir or Path(__file__).parent.absolute()
    return init([locale_directory], language, catalog, 'console')
```

but then you'll call `Path(__file__).parent.absolute()` every time the function is called.

but honestly i think the optimisation isn't worth it either way, it's a trivial operation

AA-Turner:
OK, can we use a `Final` module level constant but use `os.path` for now?

A

sphinx/locale/__init__.py
```diff
@@ -1,6 +1,7 @@
 """Locale utilities."""
 
 import locale
+from collections import UserString, defaultdict
```
AA-Turner:
```suggestion
```

sphinx/locale/__init__.py
```diff
@@ -159,8 +159,11 @@ def setlocale(category: int, value: Union[str, Iterable[str], None] = None) -> N
         pass
 
 
+LOCALE_DIR = path.abspath(path.dirname(__file__))
```
AA-Turner:
```suggestion
_LOCALE_DIR = path.abspath(path.dirname(__file__))
```

sphinx/locale/__init__.py
```diff
@@ -159,8 +159,11 @@ def setlocale(category: int, value: Union[str, Iterable[str], None] = None) -> N
...
+
+
 def init_console(
-    locale_dir: str = path.abspath(path.dirname(__file__)),  # NoQA: B008
+    locale_dir: str = LOCALE_DIR,
```
AA-Turner:
```suggestion
    locale_dir: str = _LOCALE_DIR,
```

PR #10877, opened 2022-09-28
Add `sphinx._cli`

AA-Turner:
This is the first step towards a new `sphinx` command.

References: #5618, #6938

I took inspiration from @stephenfin's approach but this is a different tack.

### Feature or Bugfix
- Feature

### Sample output

```doscon
(sphinx) I:\Development\sphinx>python -m sphinx._cli.__init__ --version
sphinx 5.3.0+/ac2a54de9

(sphinx) I:\Development\sphinx>python -m sphinx._cli.__init__ --help    
Usage: sphinx [--version] [--help] [--quiet] <command> [<args>]
                                                               
   Manage documentation with Sphinx.                           
                                                               
Options:                                                       
  --version, -V   Show the version and exit.                   
  --help, -h, -?  Show this message and exit.                  
  --quiet, -q     Only print errors and warnings.

For more information, visit <https://www.sphinx-doc.org/en/master/man/>.

(sphinx) I:\Development\sphinx>
```

A

### DIFF
--- a/sphinx/_cli/__init__.py
+++ b/sphinx/_cli/__init__.py
@@ -1,0 +1,296 @@
+"""Base 'sphinx' command.
+
+Subcommands are loaded lazily from the ``_COMMANDS`` table for performance.
+
+All subcommand modules must define three attributes:
+
+- ``parser_description``, a description of the subcommand. The first paragraph
+  is taken as the short description for the command.
+- ``set_up_parser``, a callable taking and returning an ``ArgumentParser``. This
+  function is responsible for adding options and arguments to the subcommand's
+  parser.
+- ``run``, a callable taking parsed arguments and returning an exit code. This
+  function is responsible for running the main body of the subcommand and
+  returning the exit status.
+
+The entire ``sphinx._cli`` namespace is private, only the command line interface
+has backwards-compatability guarantees.
+"""
+
+from __future__ import annotations
+
+import argparse
+import importlib
+import locale
+import sys
+from typing import TYPE_CHECKING
+
+from sphinx._cli.util.colour import (
+    bold,
+    disable_colour,
+    enable_colour,
+    terminal_supports_colour,
+    underline,
+)
+from sphinx.locale import __, init_console
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Collection, Iterable, Iterator, Sequence
+    from typing import NoReturn
+
+    _PARSER_SETUP = Callable[[argparse.ArgumentParser], argparse.ArgumentParser]
+    _RUNNER = Callable[[argparse.Namespace], int]
+
+    from typing import Protocol
+
+    class _SubcommandModule(Protocol):
+        parser_description: str
+        set_up_parser: _PARSER_SETUP  # takes and returns argument parser
+        run: _RUNNER  # takes parsed args, returns exit code
+
+
+# Map of command name to import path.
+_COMMANDS: dict[str, str] = {
+}
+
+
+def _load_subcommand_descriptions() -> Iterator[tuple[str, str]]:
+    for command, module_name in _COMMANDS.items():
+        module: _SubcommandModule = importlib.import_module(module_name)
+        try:
+            description = module.parser_description
+        except AttributeError:
+            # log an error here, but don't fail the full enumeration
+            print(f"Failed to load the description for {command}", file=sys.stderr)
+        else:
+            yield command, description.split('\n\n', 1)[0]
+
+
+class _RootArgumentParser(argparse.ArgumentParser):
+    def format_help(self) -> str:
+        help_fragments: list[str] = [
+            bold(underline(__('Usage:'))),
+            ' ',
+            __('{0} [OPTIONS] <COMMAND> [<ARGS>]').format(bold(self.prog)),
+            '\n',
+            '\n',
+            __('  The Sphinx documentation generator.'),
+            '\n',
+        ]
+
+        if commands := list(_load_subcommand_descriptions()):
+            command_max_length = min(max(map(len, next(zip(*commands), ()))), 22)
+            help_fragments += [
+                '\n',
+                bold(underline(__('Commands:'))),
+                '\n',
+            ]
+            help_fragments += [
+                f'  {command_name: <{command_max_length}}  {command_desc}'
+                for command_name, command_desc in commands
+            ]
+            help_fragments.append('\n')
+
+        # self._action_groups[1] is self._optionals
+        # Uppercase the title of the Optionals group
+        self._optionals.title = __('Options')
+        for argument_group in self._action_groups[1:]:
+            if arguments := [action for action in argument_group._group_actions
+                             if action.help != argparse.SUPPRESS]:
+                help_fragments += self._format_optional_arguments(
+                    arguments,
+                    argument_group.title or '',
+                )
+
+        help_fragments += [
+            '\n',
+            __('For more information, visit https://www.sphinx-doc.org/en/master/man/.'),
+            '\n',
+        ]
+        return ''.join(help_fragments)
+
+    def _format_optional_arguments(
+        self,
+        actions: Iterable[argparse.Action],
+        title: str,
+    ) -> Iterator[str]:
+        yield '\n'
+        yield bold(underline(title + ':'))
+        yield '\n'
+
+        for action in actions:
+            prefix = '    ' * all(o[1] == '-' for o in action.option_strings)
+            opt = prefix + '  ' + ', '.join(map(bold, action.option_strings))
+            if action.nargs != 0:
+                opt += ' ' + self._format_metavar(
+                    action.nargs, action.metavar, action.choices, action.dest,
+                )
+            yield opt
+            yield '\n'
+            if action_help := (action.help or '').strip():
+                yield from (f'        {line}\n' for line in action_help.splitlines())
+
+    @staticmethod
+    def _format_metavar(
+        nargs: int | str | None,
+        metavar: str | tuple[str, ...] | None,
+        choices: Collection[str] | None,
+        dest: str,
+    ) -> str:
+        if metavar is None:
+            if choices is not None:
+                metavar = '{' + ', '.join(sorted(choices)) + '}'
+            else:
+                metavar = dest.upper()
+        if nargs is None:
+            return f'{metavar}'
+        elif nargs == argparse.OPTIONAL:
+            return f'[{metavar}]'
+        elif nargs == argparse.ZERO_OR_MORE:
+            if len(metavar) == 2:
+                return f'[{metavar[0]} [{metavar[1]} ...]]'
+            else:
+                return f'[{metavar} ...]'
+        elif nargs == argparse.ONE_OR_MORE:
+            return f'{metavar} [{metavar} ...]'
+        elif nargs == argparse.REMAINDER:
+            return '...'
+        elif nargs == argparse.PARSER:
+            return f'{metavar} ...'
+        msg = 'invalid nargs value'
+        raise ValueError(msg)
+
+    def error(self, message: str) -> NoReturn:
+        sys.stderr.write(__(
+            '{0}: error: {1}\n'
+            "Run '{0} --help' for information"  # NoQA: COM812
+        ).format(self.prog, message))
+        raise SystemExit(2)
+
+
+def _create_parser() -> _RootArgumentParser:
+    parser = _RootArgumentParser(
+        prog='sphinx',
+        description=__('   Manage documentation with Sphinx.'),
+        epilog=__('For more information, visit https://www.sphinx-doc.org/en/master/man/.'),
+        add_help=False,
+        allow_abbrev=False,
+    )
+    parser.add_argument(
+        '-V', '--version',
+        action='store_true',
+        default=argparse.SUPPRESS,
+        help=__('Show the version and exit.'),
+    )
+    parser.add_argument(
+        '-h', '-?', '--help',
+        action='store_true',
+        default=argparse.SUPPRESS,
+        help=__('Show this message and exit.'),
+    )
+
+    # logging control
+    log_control = parser.add_argument_group(__('Logging'))
+    log_control.add_argument(
+        '-v', '--verbose',
+        action='count',
+        dest='verbosity',
+        default=0,
+        help=__('Increase verbosity (can be repeated)'),
+    )
+    log_control.add_argument(
+        '-q', '--quiet',
+        action='store_const',
+        dest='verbosity',
+        const=-1,
+        help=__('Only print errors and warnings.'),
+    )
+    log_control.add_argument(
+        '--silent',
+        action='store_const',
+        dest='verbosity',
+        const=-2,
+        help=__('No output at all'),
+    )
+
+    parser.add_argument(
+        'COMMAND',
+        nargs=argparse.REMAINDER,
+        metavar=__('<command>'),
+    )
+    return parser
+
+
+def _parse_command(argv: Sequence[str] = ()) -> tuple[str, Sequence[str]]:
+    parser = _create_parser()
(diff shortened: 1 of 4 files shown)


sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+import locale
+import sys
+
+from sphinx._cli.console_utilities import color_terminal, nocolor
+from sphinx.locale import __, init_console
+
+if False:
+    # NoQA
+    from collections.abc import Callable, Iterator, Sequence
+    from typing import Dict, Tuple
```
danieleades: i'm not sure I understand the motivation behind this pattern. Do you have benchmarks that show this makes a significant different to import time (assuming that's what you're optimising for)?

It's not very IDE friendly

AA-Turner: (Responding to your comment https://github.com/sphinx-doc/sphinx/commit/0466fa88ebd7b97e7ddda292450a349f17dde4a7#commitcomment-85236051 here as well so as not to have a discussion on a commit) 

Any typing imports slow down import time by 4% on my computer, and I want the CLI to be as responsive as possible. (Ideally under 100ms to output version/help text.)

On PyCharm this pattern works fine so it may be an issue with a different IDE? `if False` is basically what `if TYPE_CHECKING` does at runtime but it just avoids the import.

A

danieleades: Can't comment on PyCharm, but I believe this will impact all contributors that use VsCode

AA-Turner: OK. I've restored to actual objects. I might bring Sphinx 6 forwards so that we can use `from __future__ import annotations` sooner than April.

A

AA-Turner: `5.x` has been updated.

ashb (commented):
(I came across this issue due to looking for a way to build multiple sphinx builders in a single command/call)

I'm not sure what Sphinx's policy is on pulling in dependencies, but might it be worth considering some of:

- https://click.palletsprojects.com/
- https://pypi.org/project/rich-click/
- https://pypi.org/project/typer/

They handle a lot of the "boilerplate" of dealing with arg parsing, subcommads, output and terminal color/encoding etc. (I'm absolutely not trying to bikeshed here)

stephenfin:
> ### Sample output
> 
> ```
> (sphinx) I:\Development\sphinx>python -m sphinx._cli.__init__ --version
> sphinx 5.3.0+/ac2a54de9
> 
> (sphinx) I:\Development\sphinx>python -m sphinx._cli.__init__ --help    
> Usage: sphinx [--version] [--help] [--quiet] <command> [<args>]
>                                                                
>    Manage documentation with Sphinx.                           
>                                                                
> Options:                                                       
>   --version, -V   Show the version and exit.                   
>   --help, -h, -?  Show this message and exit.                  
>   --quiet, -q     Only print errors and warnings.
> 
> For more information, visit <https://www.sphinx-doc.org/en/master/man/>.
> 
> (sphinx) I:\Development\sphinx>
> ```

Out of curiosity, what's the rational for not installing a `sphinx` executable? I get that `python -m <module-name>` is the preferred pattern for something like `pip` because it helps you understand which version of Python you're running on (per https://github.com/python/cpython/issues/66491), but for a tool like Sphinx the fact that it's written in Python should be irrelevant to anyone bar a Sphinx extension or Sphinx core developer. Surely an unambiguous `sphinx` command would be preferable for normal people? :smile:

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+        class _SubcommandModule(Protocol):
+            parser_description: str
+            set_up_parser: _PARSER_SETUP  # takes and returns argument parser
+            run: _RUNNER  # takes parsed args, returns exit code
+    else:
+        from types import ModuleType as _SubcommandModule
+
+
+# Map of command name to import path.
+_COMMANDS: dict[str, str] = {
```
stephenfin, line 53: Nothing seems to populate this. I assume that's because this isn't complete? Mind marking the PR as a draft until things are closer to shippable?

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+}
+
+
+def _load_subcommand_descriptions() -> Iterator[tuple[str, str]]:
+    import importlib
```
stephenfin: Is there are a cost to importing `importlib` this that necessitates placing it here?

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+
+    for command, module_name in _COMMANDS.items():
+        module: _SubcommandModule = importlib.import_module(module_name)
+        try:
+            description = module.parser_description
+        except AttributeError:
+            continue
```
stephenfin: Shouldn't this be an error, be that a simple message or an exit? Something has registered a "command" but it doesn't follow the format we'd expect

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+        const=-1,
+        help=__('Only print errors and warnings.'),
+    )
+    parser.add_argument(
+        '--silent',
+        action='store_const',
+        dest='verbosity',
+        const=-2,
+        help=__('No output at all'),
+    )
```
stephenfin, line 214: These belong in a group. Passing e.g. `-v -v -q --silent` makes little sense

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+    parser.add_argument(
+        '--colour', '--color',
```
stephenfin: A true patriot would choose either :gb: or :us: (or better :ireland: :wink:).

Kidding aside, I'd just choose one and deal with it. Also, does this need to be tertiary? What would you expect `--color yes` to do if the terminal didn't support it? Perhaps `--color/--no-color` would be better, with the implication that the former only does something if it's supported? Alternatively, if you think we should allow requesting a colour and fail if we can't enable it, perhaps the `auto` behavior could be the default and not explicitly configurable via CLI flags (i.e. you'd still just have `--color/--no-color` and it would default to the former only if supported)

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+        choices=('auto', 'yes', 'no'),
+        default='auto',
+        help=__('Emit coloured output to the terminal, if supported'),
+    )
+
+    parser.add_argument(
+        'COMMAND',
+        nargs='...',
+        metavar=__('<command>'),
```
stephenfin, line 219: I think you'll probably need subparsers for this, tbh. If nothing else, we want a way to list the potential commands available, right?

stephenfin (changes requested):
Couple of comments inline

AA-Turner:
> Out of curiosity, what's the rational for not providing a `sphinx` executable?

This PR is intended to be the first in a series, and just introduces the new CLI infrastructure -- I don't want to provide `sphinx` before at least the main build command is ready.

A

sphinx/_cli/__init__.py
AA-Turner: Around a millisecond (``python -X importtime -Sc 'import importlib'`` vs ``python -X importtime -Sc 'import sys'``). It's not huge, but we want the command line to feel responsive -- ``python -m sphinx._cli.__init__ -V`` currently takes 55ms compared to ``python -V``'s 10ms.

sphinx/_cli/__init__.py
AA-Turner: See comment below re series of PRs -- I intend to merge this one first, and then have follow-ups to implement each individual command.

sphinx/_cli/__init__.py
AA-Turner: Good point, I will change this

sphinx/_cli/__init__.py
AA-Turner: Done

sphinx/_cli/__init__.py
AA-Turner: I've changed my mind on this and I slightly prefer an approach similar to Ruff's ([link](https://beta.ruff.rs/docs/faq/#how-can-i-disable-ruffs-color-output)) -- don't provide a CLI flag, but use environment variables. This makes the behaviour always 'auto', and you can force-disable colour by setting ``NO_COLOR=1``.

If you think we should provide a colour CLI flag I'm happy to implement as you suggest, though.

sphinx/_cli/__init__.py
AA-Turner: See follow-up PRs and ``_create_sub_parser``, this approach doesn't use the argparse subparser machinery.

sphinx/_cli/__init__.py
picnixz: If you are willing to use a "NO_COLOR" as an environment variable, I think you should stick to the same spelling in the functions as well, namely "color.py" instead of "colour.py". One argument I would have for that is that it's one letter less. In addition, "color" is used in other places.

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+    if args.colour == 'yes' or (args.colour == 'auto' and terminal_supports_colour()):
+        enable_colour()
+    else:
+        disable_colour()
+
+    # Handle '--version' or '-V' passed to the main command or any subcommand
+    if 'version' in args or {'-V', '--version'}.intersection(command_argv):
+        from sphinx import __display_version__
+        sys.stderr.write(f'sphinx {__display_version__}\n')
+        raise SystemExit(0)
```
picnixz, line 239: Just wondering, but is there a real difference between raising an exception and using `sys.exit` ? I see that you are catching the exception afterwards and returning the error code (and then reraising it in the main), but I am wondering whether this is the "canonical way" to do it (if so, I would have learned something then).

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+        # hide 'colour'
+        option_strings = [o for o in option_strings if o != '--colour']
+        prefix = '    ' * all(o[1] == '-' for o in option_strings)
+        return prefix + '  ' + ', '.join(map(bold, option_strings))
+
+    @staticmethod
+    def _format_metavar(
+        nargs: int | str | None,
+        metavar: str | tuple[str, ...] | None,
+        choices: Iterable[str] | None | None,
```
picnixz: 1. I think there is a duplicated None.
2. While I understand the fact that it could accept any iterable value, the fact that we call `sorted(choices)` means that the iterable might be exhausted. So perhaps, the correct type should be a Collection[str] (to stress that is sized collection).

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+from sphinx.locale import __, init_console
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable, Iterator, Sequence
+    from typing import NoReturn
+
+    _PARSER_SETUP = Callable[[argparse.ArgumentParser], argparse.ArgumentParser]
+    _RUNNER = Callable[[argparse.Namespace], int]
+
+    if sys.version_info[:2] > (3, 8):
```
picnixz: The Protocol class should be available since Python 3.8, so I don't think you need this import guard. On the other hand, `types.ModuleType` is available since Python 3.4

sphinx/_cli/__init__.py
```diff
@@ -0,0 +1,298 @@
...
+    parser.add_argument(
+        '--colour', '--color',
+        choices=('auto', 'yes', 'no'),
+        default='auto',
+        help=__('Emit coloured output to the terminal, if supported'),
+    )
+
+    parser.add_argument(
+        'COMMAND',
+        nargs='...',
```
picnixz: I think, it would be better to use `argparse.REMAINDER`.

sphinx/_cli/util/colour.py
```diff
@@ -0,0 +1,105 @@
+"""Format coloured console output."""
```
picnixz, line 1: As I said before, I would change 'colour' to 'color' in this file if it is ok. At least, we would be consistent with the rest of the base code.

sphinx/_cli/util/colour.py
```diff
@@ -0,0 +1,105 @@
...
+
+    try:
+        if not sys.stdout.isatty():
+            return False
+    except (AttributeError, ValueError):
+        # Handle cases where .isatty() is not defined, or where e.g.
+        # "ValueError: I/O operation on closed file" is raised
+        return False
+
+    if os.environ.get("TERM", "").lower() in {"dumb", "unknown"}:
```
picnixz: Do we have some preference towards double-quoted or single-quoted strings actually?

sphinx/_cli/util/colour.py
```diff
@@ -0,0 +1,105 @@
...
+    return inner
+
+
+# Wrap escape sequence with ``\1`` and ``\2`` to let readline know
+# that the colour escape codes are non-printable characters
+# [ https://tiswww.case.edu/php/chet/readline/readline.html ]
+#
+# Note: This does not work well in Windows
+# (see https://github.com/sphinx-doc/sphinx/pull/5059)
+if sys.platform == 'win32':
```
picnixz, line 69: Is the function purpose to allow us to create interactive colored strings for future commands ? Also, I think it duplicates what we already have in `sphinx/util/console.py` but is the latter being deprecated soon? Also, this function would have been used for the `colourize` function (in this new module) with `input_mode=True`, but the latter does not accept this keyword anymore.

sphinx/_cli/util/errors.py
```diff
@@ -0,0 +1,165 @@
...
+import sys
+import tempfile
+from typing import TYPE_CHECKING, TextIO
+
+from sphinx.errors import SphinxParallelError
+
+if TYPE_CHECKING:
+    from sphinx.application import Sphinx
+
+_ANSI_CODES: re.Pattern[str] = re.compile('\x1b.*?m')
```
picnixz: I would rename the variable `_ANSI_COLOR_CODES` because this only removes ANSI color sequences and not other escape sequences. If you want a better pattern, you can use `\x1b\[[0-9;]*m` (not that if the string being pruned is colorized in input mode, there will be some dangling `\x01` prefix so maybe we should get rid of it as well?)

sphinx/_cli/util/errors.py
```diff
@@ -0,0 +1,165 @@
...
+
+    last_msgs = exts_list = ''
+    if app is not None:
+        extensions = app.extensions.values()
+        last_msgs = '\n'.join(f'* {strip_colors(s)}' for s in app.messagelog)
+        exts_list = '\n'.join(f'* {ext.name} ({ext.version})' for ext in extensions
+                              if ext.version != 'builtin')
+
+    with tempfile.NamedTemporaryFile(suffix='.log', prefix='sphinx-err-', delete=False) as f:
+        f.write(error_info(last_msgs, exts_list, exc_format).encode('utf-8'))
```
picnixz, line 79: Should we suppress encoding errors that may arise from some exception messages ?

sphinx/_cli/util/errors.py
```diff
@@ -0,0 +1,165 @@
...
+    *,
+    stderr: TextIO = sys.stderr,
+    use_pdb: bool = False,
+    print_traceback: bool = False,
+    app: Sphinx | None = None,
+) -> None:
+    from bdb import BdbQuit
+    from traceback import TracebackException, print_exc
+
+    from docutils.utils import SystemMessage
```
picnixz, line 96: Not sure if it matters, but what about delaying even further the module imports ? (I saw that you used a lot of local imports to reduce the overall loading time).

sphinx/_cli/__init__.py
AA-Turner: sys.exit raises SystemExit internally, this is slightly more explicit.

PR #10878, opened 2022-09-28
imgmath: Fix embed mode

jschueller:
We incorrectly used the destination file instead of the temporary just generated.
The test incorrectly succeeded because of leftover files from previous tests, so moved that test first.

fixes #10816, directed to the 5.2.x branch hoping it would land into the next patch version if any

### DIFF
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -308,7 +308,7 @@ def html_visit_math(self: HTMLTranslator, node: nodes.math) -> None:
         raise nodes.SkipNode from exc
     if self.builder.config.imgmath_embed:
         image_format = self.builder.config.imgmath_image_format.lower()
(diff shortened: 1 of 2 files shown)


AA-Turner (approved):
Should we clean up the test files properly rather than rearranging the order? Approved other than that.

A

jschueller:
yes, I cleaned up the first few output directories instead, but perhaps it could be done more globally in a separate PR

AA-Turner:
I'm trying to get the CircleCI integration to show up...

jschueller:
is there a 5.2.3 planned ?

AA-Turner:
[Sphinx 5.2.3](https://pypi.org/project/Sphinx/5.2.3/) has been released.

A

jschueller:
thanks a lot!

PR #10879, opened 2022-09-28
CI: Install latex utils

jschueller:
This enables extra imgmath and latex tests


AA-Turner:
`LaTeX` tests are meant to be run on CircleCI with [this docker image](https://github.com/sphinx-doc/docker-ci/blob/master/Dockerfile), I think, though the CircleCI integration had failed (I think it is now fixed) 

Could you make a PR against that Dockerfile please?

A

jschueller:
ok, seems circleci has not run in a while

PR #10728, opened 2022-07-29
Fix #10605: autodoc unresolved type annotations

robsdedude:
Constructor annotations were not properly detected in autodoc when
`autodoc_typehints` was `'description'` or `'both'`.

### Feature or Bugfix
- Bugfix

### Relates
- https://github.com/sphinx-doc/sphinx/issues/10605

Issue #10605, opened 2022-06-26
Missing type annotations in generic class description


francois-rozet:
### Describe the bug

With `autodoc_typehints = 'description'`, the parameter type annotations of the constructor (`__init__`) of generic classes are missing from the description.

### How to Reproduce

With `autodoc_typehints = 'description'` in `conf.py`,

```python
from typing import *

#: whatever
T = TypeVar('T')

class Dummy(Generic[T]):
    r"""
    Arguments:
        arg: missing type annotation
    """

    def __init__(self, arg: str):
        pass
```

results in

<img width="275" alt="image" src="https://user-images.githubusercontent.com/37352336/175818496-76c2532f-3b23-461b-8f9d-4d22e6d415ed.png">

in which the type annotation is missing for the `arg` parameter.

### Expected behavior

The type annotations should not be missing.

### Your project

N/A

### OS

Ubuntu 20.04

### Python version

3.8.10

### Sphinx version

5.0.2

### Sphinx extensions

sphinx.ext.autodoc

### Additional context

This issue is probably related to https://github.com/sphinx-doc/sphinx/issues/8219 and https://github.com/sphinx-doc/sphinx/pull/8142.

robsdedude:
I came across a similar issue with `autodoc_typehints = 'description'`. I'm working on a PR that will fix this and my issue.

Basically, the resolution doesn't work for constructors when `autodoc_typehints = 'description'`. So something like

```python
from __future__ import annotations

import typing as t

class Dummy():
    r"""
    Arguments:
        arg: missing type annotation
    """

    def __init__(self, arg: t.Optional[str]):
        pass
```

will also end up with a wrong type hint. In this case, it will be `t.Optional[str]` instead of `Optional[str]` which it'd be for other methods.

francois-rozet:
Hello @robsdedude, thanks for your PR, I hope it gets merged!

> P.S. if you do a lot of `rebase`, avoid using issue identifiers in the commit message as it creates a lot of notifications 😅

robsdedude:
Yeah, I just realized :see_no_evil: I'm sorry.

robsdedude:
So sorry for the pings. I'll change the commit message now for goo.

AA-Turner:
@francois-rozet is this now resolved?

francois-rozet:
Hello @AA-Turner, yes. I tried with a more recent version of Sphinx (6.1.3) and it seems to be resolved.

```python
[...]

Issue #8243, opened 2020-09-26
autodoc_typehints="description" doesn't fully resolve constructor type hints from type comments


godlygeek:
**Describe the bug**
When `autodoc_typehints="description"`, type hint comments for classes aren't fully resolved, even with `autoclass_content="both"`.

**To Reproduce**
Create `module.py` with these contents:
```py
from typing import Optional

Maybe = Optional

class MaybeInt(object):
    """:param value: The value to hold"""
    def __init__(self, value):
        # type: (Maybe[int]) -> None
        self.value = value

    def get_value(self):
        # type: () -> Maybe[int]
        """:return: The held value"""
        return self.value
```

Create `index.rst` with these contents:
```rest
.. automodule:: module
   :members:
```

Generate docs twice in plain text format twice, once with `autodoc_typehints="description"` and once with `autodoc_typehints="signature"`.

```console
python3.8 -msphinx -aE -C -b text -D 'extensions=sphinx.ext.autodoc' -D autoclass_content=both -D autodoc_typehints=description . description
python3.8 -msphinx -aE -C -b text -D 'extensions=sphinx.ext.autodoc' -D autoclass_content=both -D autodoc_typehints=signature . signature
```

And note that `signature/index.txt` has:
```
class module.MaybeInt(value)

   Parameters:
      **value** (*Maybe**[**int**]*) -- The value to hold

   Return type:
      None

   get_value()

      Returns:
         The held value

      Return type:
         Optional[int]
```

In the case of the return type from `get_value`, the `Maybe` was resolved by `typing.get_type_hints` to be `Optional`. In the case of the parameter to the constructor, the `Maybe` was not able to be resolved.

When `autodoc_typehints="signature"`, both were successfully resolved to `Optional`.

**Expected behavior**

Type hints should be fully resolved.

**Environment info**
- OS: Linux
- Python version: 3.8.5
- Sphinx version: 3.2.1
- Sphinx extensions:  sphinx.ext.autodoc

**Additional context**

godlygeek:
And, changing `record_typehints` to store the annotations of `Person.__init__` for `Person` seems to work around the problem, though I'm not sure that's the right fix.

PR #10880, opened 2022-09-28
Fix 10605 constructor annotations in description

robsdedude:
Constructor annotations were not properly detected in autodoc when
`autodoc_typehints` was `'description'` or `'both'`.

### Feature or Bugfix
- Bugfix

### Fixes
- https://github.com/sphinx-doc/sphinx/issues/8243
- https://github.com/sphinx-doc/sphinx/issues/10605
- https://github.com/sphinx-doc/sphinx/issues/11674

### Rebase of
 - https://github.com/sphinx-doc/sphinx/pull/10728


tests/test_ext_autodoc.py
```diff
@@ -2345,7 +2345,11 @@ def test_autodoc(app, status, warning):
 my_name
 
 alias of Foo"""
-    assert warning.getvalue() == ''
+    warning_lines = [line for line in warning.getvalue().split('\n') if line]
+    # This might be a bug. We're not picking up the TypeVar T.
```
robsdedude, line 2882: To clarify: the test file `tests/roots/test-ext-autodoc/target/typehints_lazy.py` causes this warning with or without my changes. Not sure if I did something wrong in the test example or if this needs fixing in autodoc.

robsdedude:
Potentially pending: updating the docs with the new event. But I'd first like to get feedback if it's even worth putting further effort into this approach.

AA-Turner:
By constructor annotations, do you mean only `__init__`, or something else? 

I'm wary about adding a new event to autodoc if this can be solved another way, I suppose. 

Please may you also be a little more specific in your tests--why is `LazyGeneric` important? Why does the laziness matter? 

A

AA-Turner:
Please also don't force-push if possible, I squash-merge PRs and GitHub makes reviewing force-pushes much harder than the merge workflow.

A

robsdedude:
Thanks for having a look.

 > Please also don't force-push if possible.
Sorry. I only do this before the review process has begun. As I know how hard it is to figure out what changed between reviews when force pushed.

> By constructor annotations, do you mean only `__init__`, or something else?
`__init__`, yes. And maybe `__new__`, I'm unsure. I'm talking about this feature where `autodoc` combines the docstring of the class with the one of its constructor.

> I'm wary about adding a new event to autodoc if this can be solved another way, I suppose.
I'm not sure this is possible without a big rewrite. Right now, iirc (this work a few weeks old already), the code that puts together the types for the combined documentation is a processor (`sphinx/ext/autodoc/typehints.py`) that only post-processes the documentation nodes. So it doesn't even have access to the raw Python code. And there is no event (at least that I could find) that correctly emits the type hints from the analysis stage (`sphinx/ext/autodoc/__init__.py`). If you have an idea for an alternative approach, please share.

> Please may you also be a little more specific in your tests--why is `LazyGeneric` important? Why does the laziness matter?
Because `inspect.signature` only gives you unresolved strings instead of the actual classes/types when annotations are lazily evaluated (iirc, default since Python 3.10).

robsdedude:
I'm also battling the tests right now. The issue is that `from __future__ import annotations` was only introduced in Python3.7. So I don't see a clean way of only rendering the `target.typehints_lazy` module when `Python >= 3.7` I've tried putting it in its own file together with
```python
if sys.version_info < (3, 7):
    exclude_patterns = ['index_py37.rst']
```
But then I'm still stuck with `index_py37.rst: WARNING: document isn't included in any toctree`. I sure could expect that warning in the tests, that's what I ended up doing for now, but I'm really not fond of the solution.

AA-Turner:
I am delaying this to 6.0 or 6.1, which will be the next releases after 5.3. It will also solve the test issues as 3.6/3.7 support has been dropped.

A

robsdedude:
@AA-Turner 6.1 is out already and this fix sadly still didn't land. It'd really be a great deal if we could merge this or find a different solution. Unresolved type aliases reduce the usability of the docs significantly.

PR #10882, opened 2022-09-29
Address inconsistently localized entries in objects.inv

### LINKED ISSUE #9778: LANGUAGE environment variable inconsistently affects output of objects.inv
### Describe the bug

Hi,

Not entirely sure where the bug is here, but it seems like there is something up with language handling and generating the objects.inv file. The context to all this is that I'm working on Reproducible Builds, and some update has suddenly rendered a lot of packages that use Sphinx unreproducible - that is, generating different output regardless of the surrounding environment.

In particular, I discovered this by comparing two builds: the first with `LANGUAGE="en_GB:en"` and the second with `LANGUAGE="et_EE:et"` environment variable. What happens is that all of the documentation is identical *except* that a single entry in the `objects.inv` file appears to be translated. This is despite the output including the following logging message in both builds:

    dumping search index in English (code: en)... done
    dumping object inventory... done

(NB. `code: en` here in both builds)

Decoding this zlib-encoded file, I can see that the difference is a translation one:

    # Sphinx inventory version 2
    # Project: OpenDrop
    # Version: 

    developers/index std:doc -1 developers/index.html Developer notes
    -genindex std:label -1 genindex.html Index
    +genindex std:label -1 genindex.html Indeks
    getting_started/index std:doc -1 getting_started/index.html Getting Started
    index std:doc -1 index.html Overview
    usage/conan std:doc -1 usage/conan.html Contact Angle
    usage/ift std:doc -1 usage/ift.html Interfacial Tension

... and, indeed, "Indeks" is in the Estonian .po file:

    #: sphinx/builders/latex/__init__.py:194 sphinx/domains/std.py:604
    #: sphinx/templates/latex/latex.tex_t:97
    #: sphinx/themes/basic/genindex-single.html:30
    #: sphinx/themes/basic/genindex-single.html:55
    #: sphinx/themes/basic/genindex-split.html:11
    #: sphinx/themes/basic/genindex-split.html:14
    #: sphinx/themes/basic/genindex.html:11 sphinx/themes/basic/genindex.html:34
    #: sphix/themes/basic/genindex.html:67 sphinx/themes/basic/layout.html:147
    #: sphinx/writers/texinfo.py:498
    msgid "Index"
    msgstr "Indeks"

This is just confusing though because why isn't "Module Index" translated as well? "Mooduli indeks" is also there in the Estonian .po:

    #: sphinx/domains/std.py:605
    msgid "Module Index"
    msgstr "Mooduli indeks"

... so I suppose the bug here is either that "Index" gets translated whilst  "Module Index" is not... or the other way around. This why I use "inconsistency" in the title of this issue.

Playing around with the code, I am pretty certain that the translated entry is in `sphinx/domains/std.py` — could it be that the data in `initial_data` is being prematurely translated? Either way, though, I was expecting that the documentation and entries are all identical, regardless of the `LANGUAGE` environment variable.

### How to Reproduce

Compare the builds between exporting the `LANGUAGE="en_GB:en` and `LANGUAGE="et_EE:et"` environment variable, specifically the `objects.inv` file.

### Expected behavior

_No response_

### Your project

I'm using opendrop, but this will occur with any package

### Screenshots

_No response_

### OS

Linux

### Python version

3.9

### Sphinx version

4.2.0

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

jayaddison:
Subject: Address inconsistently localized entries in `objects.inv` by removing translation lazy-loader proxy class

### Feature or Bugfix
- Bugfix

### Purpose
Numerous packages within Debian are identified as non-reproducible during parallel builds that are performed side-by-side using builds with English-language and Estonian-language locale configurations (among other environmental differences).

One recurring cause for these appears to be that the `objects.inv` file generated by Sphinx contains some (often only one, or a small number) localized display names for entries.

This behaviour appears to be inconsistent as documented in #9778.

Removing the lazy-loader `_TranslationProxy` class mechanism appears to help based on some local testing - although I'm yet to figure out why that is precisely.

### Detail
- example `diffoscope` link for differences from a `python-psutil` package build (at the time-of-writing, this displays an example `objects.inv` localization issue) - https://tests.reproducible-builds.org/debian/dbd/unstable/amd64/python-psutil_5.9.2-1.diffoscope.html#python-psutil-doc_-.-.---_all.deb---data.tar.xz---data.tar---.-usr-share-doc-python-psutil-doc-html-objects.inv
- I've confirmed by performing a documentation build for `language=et` configured via `sphinx/config.py` (perhaps not the correct place...) that the resulting HTML documentation build _does_ continue to be localized as expected (naive regression test)

### Relates
May resolve #9778.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -19,6 +19,8 @@ Features added
 Bugs fixed
 ----------
 
+* #9778: Address inconsistently localized entries in ``objects.inv`` by removing translation lazy-loader proxy class.
+
 Testing
 --------
 
--- a/sphinx/locale/__init__.py
+++ b/sphinx/locale/__init__.py
@@ -6,90 +6,6 @@ from os import path
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 
-class _TranslationProxy:
-    """
-    Class for proxy strings from gettext translations. This is a helper for the
-    lazy_* functions from this module.
-
-    The proxy implementation attempts to be as complete as possible, so that
-    the lazy objects should mostly work as expected, for example for sorting.
-    """
-    __slots__ = ('_func', '_args')
-
-    def __new__(cls, func: Callable, *args: str) -> '_TranslationProxy':
-        if not args:
-            # not called with "function" and "arguments", but a plain string
-            return str(func)  # type: ignore[return-value]
-        return object.__new__(cls)
-
-    def __getnewargs__(self) -> Tuple[str]:
-        return (self._func,) + self._args  # type: ignore
-
-    def __init__(self, func: Callable, *args: str) -> None:
-        self._func = func
-        self._args = args
-
-    def __str__(self) -> str:
-        return str(self._func(*self._args))
-
-    def __dir__(self) -> List[str]:
-        return dir(str)
-
-    def __getattr__(self, name: str) -> Any:
-        return getattr(self.__str__(), name)
-
-    def __getstate__(self) -> Tuple[Callable, Tuple[str, ...]]:
-        return self._func, self._args
-
-    def __setstate__(self, tup: Tuple[Callable, Tuple[str]]) -> None:
-        self._func, self._args = tup
-
-    def __copy__(self) -> '_TranslationProxy':
-        return _TranslationProxy(self._func, *self._args)
-
-    def __repr__(self) -> str:
-        try:
-            return 'i' + repr(str(self.__str__()))
-        except Exception:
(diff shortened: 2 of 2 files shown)


jayaddison:
After more thought, I don't think that this approach is correct.  The `objects.inv` file _should_ be (and is) localized.  Removing the translation proxy causes builds in different languages to produce the same inventory file, but I don't believe that that's what we want (because the inventory file may be published in a native language for cross-reference from other documentation sites).

PR #10883, opened 2022-09-29
Extend cross referencing options with values

marxin:
This supports a commonly used format when it comes to optional argument: `-fauto-profile[=path]`, where `path` is optional part. Plus it supports a format when space is used: `-fauto-profile path`.

Extends: #10840

### DIFF
--- a/sphinx/domains/std.py
+++ b/sphinx/domains/std.py
@@ -943,10 +943,17 @@ class StandardDomain(Domain):
         progname = node.get('std:program')
         target = target.strip()
         docname, labelid = self.progoptions.get((progname, target), ('', ''))
-        # for :option:`-foo=bar` search for -foo option directive
-        if not docname and '=' in target:
-            target2 = target[:target.find('=')]
-            docname, labelid = self.progoptions.get((progname, target2), ('', ''))
+        if not docname:
+            # Support also reference that contain an option value:
+            # * :option:`-foo=bar`
+            # * :option:`-foo[=bar]`
+            # * :option:`-foo bar`
+            for needle in {'=', '[=', ' '}:
+                if needle in target:
+                    stem, _, _ = target.partition(needle)
+                    docname, labelid = self.progoptions.get((progname, stem), ('', ''))
+                    if docname:
+                        break
         if not docname:
             commands = []
             while ws_re.search(target):
--- a/tests/roots/test-root/objects.txt
+++ b/tests/roots/test-root/objects.txt
@@ -214,7 +214,8 @@ Test repeated option directive.
 
   My secret API.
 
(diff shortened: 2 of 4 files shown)


marxin:
What do you think @AA-Turner?

sphinx/domains/std.py
```diff
@@ -943,10 +943,17 @@ def _resolve_option_xref(self, env: "BuildEnvironment", fromdocname: str,
...
-        # for :option:`-foo=bar` search for -foo option directive
-        if not docname and '=' in target:
-            target2 = target[:target.find('=')]
-            docname, labelid = self.progoptions.get((progname, target2), ('', ''))
+        if not docname:
+            # Support also reference that contain an option value:
+            # * :option:`-foo=bar`
+            # * :option:`-foo[=bar]`
+            # * :option:`-foo bar`
+            for needle in ('=', '[=', ' '):
```
AA-Turner:
```suggestion
            for needle in {'=', '[=', ' '}:
```

sphinx/domains/std.py
```diff
@@ -943,10 +943,17 @@ def _resolve_option_xref(self, env: "BuildEnvironment", fromdocname: str,
...
+                idx = target.find(needle)
+                if idx != -1:
+                    docname, labelid = self.progoptions.get((progname, target[:idx]), ('', ''))
```
AA-Turner: More readable / 'pythonic', I think:
```suggestion
                if needle in target:
                    stem, _, _ = target.partition(needle)
                    docname, labelid = self.progoptions.get((progname, stem), ('', ''))
```

tests/roots/test-root/objects.txt
```diff
@@ -214,7 +214,7 @@ Test repeated option directive.
 
   My secret API.
 
-Reference the first option :option:`-mapi=secret`.
+Reference the first option :option:`-mapi=secret` or :option:`-mapi[=xxx]`.
```
AA-Turner: Can you test for the single space case too?

AA-Turner (changes requested):
+1

AA-Turner:
@marxin please don't force push, it makes review much more difficult after leaving comments.

A

marxin:
All right, so you prefer leaving there the "Suggested change" commits, right?

AA-Turner:
Yes, I squash merge on commit.

A

AA-Turner (approved):

AA-Turner:
Thanks!

A

marxin:
> Yes, I squash merge on commit.

Makes sense, will do that next time.

marxin:
> Thanks!

I thank you for a nice negotiation we made.

PR #10884, opened 2022-09-29
CI: Move LaTeX job to GitHub Actions

jschueller:
moves the latex job to github action instead of circleci


jschueller:
seems like circleci must be unregistered after merge

.github/workflows/latex.yml
```diff
@@ -0,0 +1,22 @@
...
+    env:
+      DO_EPUBCHECK: 1
+    steps:
+    - uses: actions/checkout@v3
+    - name: Use LaTeX 
+      run: |
+        /python3.8/bin/pip install -U pip setuptools
+        /python3.8/bin/pip install -U .[test]
+        mkdir -p test-reports/pytest
+        make test PYTHON=/python3.8/bin/python TEST="--junitxml=test-reports/pytest/results.xml -vv"
```
AA-Turner:
```suggestion
    - name: Check Python version
      run: python --version
    - name: Install dependencies
      run: python -m pip install -U pip tox
    - name: Run Tox
      run: tox -e py -- -vv
```

AA-Turner:
I also wonder what the time difference is between using the docker image vs your orignal suggestion of installing the extra packages.

A

jschueller:
i think it was ~15s to install the extra 400mb packages

.github/workflows/latex.yml
```diff
@@ -0,0 +1,23 @@
+name: CI (LaTeX)
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
AA-Turner:
```suggestion
    runs-on: ubuntu-18.04
    name: Python 3.8
```

AA-Turner:
Thanks @jschueller!

A

AA-Turner:
The CI image is defined in the dockerfile (https://github.com/sphinx-doc/docker/blob/master/ci/Dockerfile) and contains several TeX related packages which aren't in the default `ubuntu-XX.04` runner (https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md). Suggestions on changing the extra packages we use are very welcome!

A

jfbu:
@AA-Turner Thanks! I knew I was missing something obvious and checking more carefully I see for example from [a tox log of a recent  Python3.9+du17 testing](https://github.com/sphinx-doc/sphinx/actions/runs/3238188943/jobs/5306125461#step:8:630):
```
tests/test_build_latex.py::test_build_latex_doc[pdflatex-howto] SKIPPED (not running latex, the required styles do not seem to be installed) [ 32%]
tests/test_build_latex.py::test_build_latex_doc[pdflatex-manual] SKIPPED (not running latex, the required styles do not seem to be installed) [ 32%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-howto] SKIPPED (not running latex, the required styles do not seem to be installed) [ 32%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-manual] SKIPPED (not running latex, the required styles do not seem to be installed) [ 32%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-howto] SKIPPED (not running latex, the required styles do not seem to be installed) [ 32%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-manual] SKIPPED (not running latex, the required styles do not seem to be installed) [ 32%]
```
whereas those pdf building tests (the only ones done to my knowledge) are indeed executed on the latex-tagged CI docker image.

Only building a pdf can trigger check of latex environment; all other tests are on the output of the Sphinx LaTeX writer and do not involve the actual LaTeX installation, and on ̀`sphinx-build -b latex`  Sphinx LaTeX does no actual check of what is the user configuration and the produced files do not depend upon it; there are some conditional branches inside those produced latex source files, but they get exectued only during the pdf build: for example usage of `xcolor`, or of  `pict2e` LaTeX packages  is dynamical, if they don't exist on user system, the pdf build will proceed without them and it avoids letting LaTeX complain and causing its failure.

About 

> Side remark: at some point we were building our document on RTD as part of testing of major branch which was useful, but this is gone so it seems.

have we dropped that or I miss something again?

AA-Turner:
Every PR has a test build on read-the-docs, if you click on the 'checks' bit at the bottom of the PR it should be there.

A

jfbu:
> Every PR has a test build on read-the-docs, if you click on the 'checks' bit at the bottom of the PR it should be there.

I had been looking for them in the checks done when merging into 5.x, but indeed it was for PR checks I had seen them earlier...

PR #10886, opened 2022-09-29
Finer grained control over domain ToC entries

AA-Turner:
Follow-up from #10807 and #6316.

### Feature or Bugfix
- Feature
- Bugfix

cc @jakobandersen @josephmure @asmeurer 

A

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -678,6 +678,11 @@ General configuration
    :term:`object` names (for object types where a "module" of some kind is
    defined), e.g. for :rst:dir:`py:function` directives.  Default is ``True``.
 
+.. confval:: toc_object_entries
+
+  Create table of contents entries for domain objects (e.g. functions, classes,
+  attributes, etc.). Default is ``True``.
+
 .. confval:: toc_object_entries_show_parents
 
    A string that determines how domain objects (e.g. functions, classes,
--- a/doc/usage/restructuredtext/domains.rst
+++ b/doc/usage/restructuredtext/domains.rst
@@ -1180,23 +1180,27 @@ visibility statement (``public``, ``private`` or ``protected``).
    The example are rendered as follows.
 
    .. cpp:type:: std::vector<int> MyList
-      :noindex:
+      :nocontentsentry:
+      :noindexentry:
 
       A typedef-like declaration of a type.
 
    .. cpp:type:: MyContainer::const_iterator
-      :noindex:
+      :nocontentsentry:
+      :noindexentry:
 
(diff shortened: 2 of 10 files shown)


asmeurer:
I can confirm the SymPy docs still build fine with this PR.

josephmure:
The OpenTURNS doc builds correctly with this PR. Thanks!

jakobandersen:
Looks good to me, except I suggest having the direction option be called ``notocentry``, which I think provides a more direct association to the ``toctree`` directive that it in a sense modifies.

Somehow the changelog should gain an "incompatible changes" entry, where the new configuration option is mentioned to restore the old behaviour, but I guess it's not feasible to retroactively change the 5.2 entries.

Thanks for the quick response.

AA-Turner:
my concern with `notocentry` is parsing the phrase for readers unfamilar--as the others (no index, no index entry) use `smushedwords` I hesitate to use e.g. `no-toc-entry`.

What would you prefer of either `nocontentsentry` or `no-toc-entry`?

A

AA-Turner:
> The OpenTURNS doc builds correctly with this PR. Thanks!

Thank you for testing. I would ask that you consider for a little while keeping the ToC entries enabled in your documentation -- I have found it very useful in *reference* documentation to be able to jump directly to the object I want, and your readers may do too. I admit to significant bias here, though, as I implemented the feature, and it would be nice for it to be used rather than disabled.

A

jakobandersen:
> my concern with `notocentry` is parsing the phrase for readers unfamilar--as the others (no index, no index entry) use `smushedwords` I hesitate to use e.g. `no-toc-entry`.
> 
> What would you prefer of either `nocontentsentry` or `no-toc-entry`?

I'm not sure ``notocentry`` is that bad, but it's not too important to me. I'd rather go with the current than using dashes.

I think it's a great feature to have the automatic toc entries. The point of the option is that users can enable them when they are ready, and still get the benefit of newer Sphinx versions in general.

AA-Turner:
[Sphinx 5.2.3](https://pypi.org/project/Sphinx/5.2.3/) has been released.

A

josephmure:
> > The OpenTURNS doc builds correctly with this PR. Thanks!
> 
> Thank you for testing. I would ask that you consider for a little while keeping the ToC entries enabled in your documentation -- I have found it very useful in _reference_ documentation to be able to jump directly to the object I want, and your readers may do too. I admit to significant bias here, though, as I implemented the feature, and it would be nice for it to be used rather than disabled.
> 
> A

We'll certainly test the feature, but I think we'll need a comprehensive redesign of our document to use it effectively.

PR #10888, opened 2022-10-01
Fix performance regression for ``imgmath`` embedding

jschueller:
We need the generated image files to be in one common folder to avoid having to re-generate several times the same image (the sha1 naming of the file acts as a cache to just read the depth)

The location should be possibly common to all workers, that's why I choosed to just keep the original destination folder, ie the file is just moved from tempdir to the build folder before.

This significantly drops my build time from 84min to 26min, but obviously leaves the unused image files in the output folder.

Another solution would be to use another global temporary folder shared across all the workers, or to add a delete pass after all workers have joined, but there does not seem to be an api to do that.

Bugfix: fix imgmath performance with embed=True
Follows  #10816  #10878

### DIFF
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -294,6 +294,20 @@ def cleanup_tempdir(app: Sphinx, exc: Exception) -> None:
         pass
 
 
+def cleanup_mathdir(app: Sphinx, exc: Exception) -> None:
+    if exc:
+        return
+    if not app.builder.config.imgmath_embed:
+        return
+    # in embed mode, the images are still generated in the math output dir
+    # to be shared across workers, but are not useful to the final document
+    mathdir = path.join(app.builder.outdir, app.builder.imagedir, 'math')
+    try:
+        shutil.rmtree(mathdir)
+    except Exception:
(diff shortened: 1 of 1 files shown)


AA-Turner:
> This significantly drops my build time from 84min to 26min, but obviously leaves the unused image files in the output folder.

Is your project open source? 26 minutes is incredibly long, I'd like to see if there are any other optimisations that can be made here.

A

jschueller:
it's not that long considering there are almost 6k math formulas to generate
https://github.com/openturns/openturns

An idea might be generate images in batches to avoid running latex once per image

jschueller:
Also I tried compilation with mathjax, which drops compilation to 14min but validity of formulas is not checked, which is a no-go

jschueller:
@AA-Turner please let me know if I should do anything else, I'd like to submit that for 5.3.0

jschueller:
cool, thanks for taking over

AA-Turner:
@jschueller if you could sanity check that your builds work with my changes it'd be appreciated!

jschueller:
yes, I just tested it and its ok

PR #10892, opened 2022-10-02
Update CHANGES for `:option:` reference format changes

marxin:
Documents: #10883


marxin:
@AA-Turner please include this in `5.3`.

AA-Turner (approved):

PR #2192, opened 2015-12-23
Imgmath (pngmath with svg support)

xuhdev:
As discussed in #2179, should solve #1570 

I've finished the code, updated the tests, and updated the doc. The only thing left is to deprecate the old `pngmath`. Could you tell me how would I do it in a way which fits in Sphinx's policy?

PR #10895, opened 2022-10-04
Fix `cleanup_tempdir` in imgmath

jschueller:
bugfix: seems the attribute name was wrong since #2192


jschueller:
@AA-Turner a good candidate for 5.3.0

AA-Turner (approved):

PR #10921, opened 2022-10-16
Simplify development policy

AA-Turner:
This simplifies the development policy to only have one long-lived branch, and decouples Sphinx's development cadence from Ubuntu.

A


jakobandersen:
Maybe I missed it somewhere, but what is now the branch model. Where should bugfixes, features, and breaking changes go?

PR #10922, opened 2022-10-16
Revert ``html_codeblock_linenos_style`` removal

### LINKED ISSUE #10265: Why deprecate the `html_codeblock_linenos_style`  option?
We've been porting the mkdocs-material theme to sphinx-immaterial, and found the the mkdocs-material theme uses tables to represent literal code blocks with line numbers. With Sphinx's default setting
```py
html_codeblock_linenos_style = "inline"
```
We found a few problems:
1. The line numbers are copied to the clipboard when using clipboard.js.
2. The CSS inherited from mkdocs-material cannot distinguish line numbers from the rest of the literal code block. Thus no intended CSS gets used.
3. Line numbers scroll horizontally with the rest of the literal code block. Ideally, we expect the line numbers to be statically positioned while horizontally scrolling the rest of the literal code block.

Please don't deprecate the `html_codeblock_linenos_style` config option in Sphinx v6.0.

For now, we're thinking of having the sphinx-immaterial theme assert the option to use
```py
html_codeblock_linenos_style = "table"
```
When/If Sphinx v6.0 removes the option, we're going to have to overwrite the `visit_literal_block()` to restore this functionality.

**Additional context**
> I talked with @tk0miya at the Sphinx(jp) hack-a-thon event, and we discussed that there is no need to leave table output in the future.

I've read https://github.com/sphinx-doc/sphinx/pull/7879#issuecomment-650717211, but I don't see any good rationality to deem the option deprecated. The sphinx builtin themes aren't the only themes that supply their own CSS (we don't inherit any CSS from sphinx builtin themes). If writing CSS for code blocks with line numbers (using `"table"`) is too complex, then leave it to the theme designers to choose how to set the option - don't deprive them of the option all together.

We're not as flexible as we'd like to be concerning CSS changes. This is because we need to minimize the possible merge conflicts when merging in updates from the mkdocs-material theme. Same reason goes for using clipboard.js instead of executablebooks/sphinx-copy-button extension (though this isn't as strict a requirement as keeping the CSS unaltered).

Original issue for the sphinx-immaterial theme is https://github.com/jbms/sphinx-immaterial/issues/26

AA-Turner:
The removal would break ``sphinx-immaterial``. Until their upstream source is fixed, or they find a workaround, revert the removal.

Closes #10265

cc: @2bndy5

A

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1065,6 +1065,20 @@ that use Sphinx's HTMLWriter class.
 
    .. versionadded:: 1.8
 
+.. confval:: html_codeblock_linenos_style
+
+   The style of line numbers for code-blocks.
+
+   * ``'table'`` -- display line numbers using ``<table>`` tag
+   * ``'inline'`` -- display line numbers using ``<span>`` tag (default)
+
+   .. versionadded:: 3.2
+   .. versionchanged:: 4.0
+
+      It defaults to ``'inline'``.
+
+   .. deprecated:: 4.0
+
 .. confval:: html_context
 
    A dictionary of values to pass into the template engine's context for all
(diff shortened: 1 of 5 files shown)


2bndy5:
Thank you for this!

Issue #10920, opened 2022-10-16
6.0.0 release plan


AA-Turner:
I intend to release the first beta of Sphinx 6.0.0 today. This will allow extensions to beta-test against the new version of Sphinx. 6.0.0 final will be released no earlier than 01/12/22.

Issues marked against 6.0.0: [sphinx-doc/sphinx/milestone/125](https://github.com/sphinx-doc/sphinx/milestone/125)

A

AA-Turner:
[Sphinx 6.0.0b1](https://pypi.org/project/Sphinx/6.0.0b1/) has been released. From now, no breaking changes should be merged.

A

jfbu:
@AA-Turner Is there a way you could ping from such release announcement? For some reason I do not configure my github account to receive email notifications on all issue creations. So I missed this.  Else I would have pushed some PR to make Sphinx by default use for pdf via latex modified rendering of code-blocks and tables (refer in particular #10620). Now, this would be considered breaking in LaTeX world but may not be so in Sphinx.  It is my estimate that only a minority of Sphinx users actually care about the pdf looks. But if such change (for example code-blocks always with some light background and rounded corners) is breaking I can now not merge it until probably 7.0.0...

AA-Turner:
@jfbu sorry; I will ping in future. If you can push your PR before Monday I will release beta 2 with the change, I think it will still be fine as we have 2 months until final release.

A

jfbu:
@AA-Turner ah ok good.  Even if you had pinged me I would have missed the beta 1 as I had some other tasks this week-end such as biking...

AA-Turner:
Do let me know if you make the PR (ping/etc), I think a cosmetic change (rendering) is not breaking for Sphinx on further thought, as it does not affect our APIs or other contracts. That means it could go in 6.0, 6.1, etc.

A

jfbu:
@AA-Turner I have done a PR #10924, but now I am not sure about how to modify CHANGES, in view of 
>  I think a cosmetic change (rendering) is not breaking for Sphinx on further thought,
it looks as if "Incompatible changes" section is not appropriate.

hugovk:
Tip: you can do Watch > Custom > Releases to just get notifications for releases:

<img width="1363" alt="image" src="https://user-images.githubusercontent.com/1324225/196117388-6a8e708c-0517-4e4e-aa0b-9689a98be8cc.png">

jfbu:
[...]

Issue #10620, opened 2022-07-01
LaTeX: change defaults for admonition or code-block colors from white or black as currently used


jfbu:
I have tested some configuration, vaguely motivated to have about same colors for PDF rendering as in the default HTML theme, Alabaster.

My main motivation is for code-blocks where I feel the framing is distractive, and I feel a light grey background colour would be better.  But I have not tested what it gives on printed paper.

Here is the configuration I tried (with the proposal to make it the default at 6.0.0).  Mainly colors are changed and a few widths.

```rest
latex_elements = {
    'sphinxsetup': """%
  VerbatimColor={RGB}{230,230,230},%
  verbatimwithframe=false,
  verbatimsep=5pt,
%       shadowBorderColor={RGB}{102,102,102},% requires #10619
%       shadowShadowColor={RGB}{187,187,187},%
  shadowrule=2pt,%
  shadowsize=6pt,%
  noteBorderColor={RGB}{204,204,204},%
  hintBorderColor={RGB}{204,204,204},%
  importantBorderColor={RGB}{204,204,204},%
  tipBorderColor={RGB}{204,204,204},%
  noteborder=5pt,%
  hintborder=5pt,%
  importantborder=5pt,%
  tipborder=5pt,%
  warningBorderColor={RGB}{255,119,119},%
  cautionBorderColor={RGB}{255,119,119},%
  attentionBorderColor={RGB}{255,119,119},%
  dangerBorderColor={RGB}{255,119,119},%
  errorBorderColor={RGB}{255,119,119},%
  warningBgColor={RGB}{255,238,238},%
  cautionBgColor={RGB}{255,238,238},%
  attentionBgColor={RGB}{255,238,238},%
  dangerBgColor={RGB}{255,238,238},%
  errorBgColor={RGB}{255,238,238},%
  % LaTeX code of heavyboxes draws border "inwards" so
  % it must be not too big
  warningborder=3pt,%
  cautionborder=3pt,%
  attentionborder=3pt,%
  dangerborder=3pt,%
  errorborder=3pt,%
""",
}
```

*edit: on second thoughts perhaps keeping verbatimsep=3pt which is current default (it is set to `\fboxsep` and the latter default to `3pt` in LaTeX) is better than the 5pt tried out here*

It renders like this (with #10619 merged):

![foo-5](https://user-images.githubusercontent.com/2589111/176874832-0e0740ea-d542-4817-a3ca-76fd7763d324.svg)

[...]

PR #10924, opened 2022-10-16
LaTeX: switch to some nicer defaults for tables and code-blocks

jfbu:
### Relates
- #10620, #10920

This addresses only partially #10620 as rendering defaults for admonitions are left unchanged by this PR.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -43,6 +43,11 @@ Deprecated
 Features added
 --------------
 
+* #10924: LaTeX: adopt better looking defaults for tables and code-blocks.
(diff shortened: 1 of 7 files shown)


jfbu:
@AA-Turner I am now squash-merging this for beta2 release (once testing completes).  CHANGES mentions this as a new feature and gives pointers on how to revert it.  If beta testing turns out to reveal this change causes problems we can revert it.  (I am a bit unsure about how people will react to tables being with "zebra-striped" coloured rows per default).

PR #10930, opened 2022-10-23
Highlight all search terms on the search results page

### LINKED ISSUE #10864: highlightSearchWords() is not called after fragments are loaded via AJAX
### Describe the bug

When I build Sphinx' own documentation locally, and search for `example`, search results look like this:
Screenshot 1

Only if I go to browser console and call `SphinxHighlight.highlightSearchWords()`, it starts looking like this:
Screenshot 2

`highlightSearchWords()` is called on DOMContentLoaded event, but page fragments are loaded separately after that event.

### How to Reproduce

```bash
$ git clone https://github.com/sphinx-doc/sphinx.git
$ cd sphinx
$ python3 ./sphinx/cmd/build.py -T doc build/html/
$ cd build/html
$ python3 -m http.server
$ # open http://127.0.0.1:8000/search.html?q=example
```

### Expected behavior

Search terms are highlighted after all page fragments are loaded.

### Your project

Sphinx itself

### Screenshots

_No response_

### OS

Linux

### Python version

3.10.7

### Sphinx version

5.2.0.post1 and master

### Sphinx extensions

_No response_

### Extra tools

_No response_

### Additional context

_No response_

mitya57:
Subject: Highlight all search terms on search results page

### Feature or Bugfix
- Bugfix

### Purpose
`SphinxHighlight.highlightSearchWords()` is called on DOMContentLoaded event, and it will not highlight page fragments loaded asynchronously after that event. Also, when doing two searches in a row, that function uses search term from the _previous_ search query because it was saved in localstorage.

So we should highlight all such fragments manually.

(At first I tried to rehighlight the whole page by calling `SphinxHighlight.highlightSearchWords()` after each fragment is loaded. It did not work because the first invocation deleted `sphinx_highlight_terms` from localstorage, so all subsequent invocations did not work.)

### Relates
- Fixes #10864.


mitya57:
I have pushed a new version of this PR with two improvements:
- Now it highlights all occurrences of search term in a fragment, not just the first one.
- `highlightSearchWords()` is no longer called on the search page at all. First, it's no longer needed after my changes. Second, it was wrong because it uses search terms from the previous search query, not from the current one.

Now our [WebKit-based tests][1] in Debian finally pass in a reproducible way (previously they have been failing from time to time because of race condition between `highlightSearchWords()` and my new code).

[1]: https://salsa.debian.org/python-team/packages/sphinx/-/tree/debian/master/debian/jstest

@AA-Turner Sorry for direct ping, but I don't want this PR to be buried under a bunch of more recent ones, so whenever you have time, please take a look.

AA-Turner:
Thanks @mitya57!

A

Issue #788, opened 2015-01-02
Creating multi-language HTML builds


shimizukawa:
With the translation machinery released in version 1.1 it is a natural idea to generate multiple builds using different languages and create a HTML tree including different languages with the outer pages allowing to switch between them.

It would also be interesting to investigate if this makes any sense with output formats other than HTML.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/788
- Originally reported by: Dinu Gherman
- Originally created at: 2011-10-11T18:57:14.846

aruhier:
Hello,
It would be a great idea to generate builds by language !

SamB:
I ran into some documentatin that does this - it's not great that they've hardcoded the language list into layout.html, but this does demonstrate that people want to do this, and shows a UI they think works well.

Anyway, the relevant bits of doc source can be found from:

-  <https://github.com/search?q=repo%3Acryptpad%2Fdocumentation+language-selector-container+AND+NOT+path%3Awww%2F&type=code>

The language switcher itself starts at:
-  <https://github.com/cryptpad/documentation/blob/d478435047a6402113568bdde043303e0ae1709e/_themes/theme/layout.html#L105>

and the rendered docs are at: <https://docs.cryptpad.org>

jayaddison:
Note also: the use of `Accept-Language` header on the webserver in the approach above is entirely optional; a simple static webserver will still be able to host the combined multilingual HTML documentation absolutely fine, and users will be able to choose their preferred language.  However, the initial/default language they see will be the project default instead of a dynamically-negotiated one determined from the browser settings.

PR #10949, opened 2022-10-31
Disable localisation when ``SOURCE_DATE_EPOCH`` is set

### LINKED ISSUE #9778: LANGUAGE environment variable inconsistently affects output of objects.inv

jayaddison:
Subject: disables localization when reproducible builds are requested, as determined by a non-empty [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) environment variable.

### Feature or Bugfix
- Bugfix

### Purpose
The [Reproducible Builds](https://www.reproducible-builds.org/) project aims to provide confidence to consumers of packaged software that the artefacts they're downloading and installing have not been altered by the environment they were built in, and can be replicated at a later date if required.

The project performs continuous reproducibility testing of many different software packages - many of which fail reproducibility tests for various reasons.  The project then identifies ways to fix the causes of differences, or filter out irrelevant differences that should not affect the software's practical behaviour.

Builds of localized documentation using `sphinx` currently account for a large [category](https://tests.reproducible-builds.org/debian/issues/unstable/sphinxdoc_translations_issue.html) of reproducible build testing failures, because the builders (intentionally) use varying environment locales at build-time.  This can affect the contents of the `objects.inv` file.

During investigation, it turned out that many `gettext`-localized values -- particularly in Python modules under `sphinx.domains` -- were being translated at module-load-time (in other words, when each Python module was `import`'ed) -- and would not subsequently be re-localized.

This creates two unusual effects:

1. Attempting to write a test case to build the same application in two different languages was not initially possible -- the first-loaded translation catalog (as found in the `sphinx.locale.translators` global variable) -- would remain in-use for subsequent application builds under different locales
2. Localization of strings could vary depending on whether the relevant modules were loaded before or after the resource catalogs were populated

A possible approach to a fix here is a naive one: all translations are performed lazily so that module imports can occur in any order and localization of inventory entries should occur only when translations of those items are requested (during the `objects.inv` dump, in the context of the investigation).

~~With that change in place, it's relatively straightforward to then disable localization (set `languages = None`) when a reproducible build is requested.~~

Edit: description correction: the branch has been updated to configure `languages = ["und"]` instead of `languages = None`.  `und` is a three-character ISO-639-3 code for 'undetermined' language.  This is used to _prevent_ Python's stdlib `gettext` from attempting to determine the host's locale from various environment variables (including the `LANGUAGE` variable).

### Relates
- May resolve #9778


jayaddison:
Hmm.. the original issue report was specifically about the `LANGUAGE` environment variable, and that isn't exercised here at the moment.  Perhaps I'm on the wrong track here.

tests/test_util_inventory.py
```diff
@@ -83,3 +84,58 @@ def test_read_inventory_v2_not_having_version():
...
+        # Build the same app using English (US) locale
+        srcdir_en = _write_appconfig(tempdir, "en")
+        reproducible_inventory_en = _build_inventory(srcdir_en)
+
+    # Also build the app using Estonian (EE) locale without build reproducibility enabled
+    srcdir_et = _write_appconfig(tempdir, "et", prefix="localized")
+    localized_inventory_et = _build_inventory(srcdir_et)
+
+    # Ensure that the reproducible inventory contents are identical
+    assert reproducible_inventory_et.read_bytes() == reproducible_inventory_en.read_bytes()
```
jayaddison, line 138: :warning: This test assertion _fails_ when tests are run like so:

```sh
$ LANGUAGE=et_EE:et tox -e py tests/test_util_inventory.py
```

...so the environment variable continues to affect the inventory output in that case.

jayaddison: Explanation: this was due to `gettext.translation` and the way that it checks environment variables for locales when the `languages` parameter is empty-list-or-none.

https://docs.python.org/3/library/gettext.html#gettext.translation

sphinx/locale/__init__.py
```diff
@@ -112,9 +112,21 @@ def init(
...
+        # See https://reproducible-builds.org/docs/source-date-epoch/
+        #
+        # Note: Providing an empty/none value to gettext.translation causes
+        # it to consult various language-related environment variables to find
+        # locale(s).  We don't want that during a reproducible build; we want
+        # to run through the same code path, but to return NullTranslations.
+        #
+        # To achieve that, specify the ISO-639-3 'undetermined' language code,
+        # which should not match any translation catalogs.
+        languages: Optional[List[str]] = ['und']
```
jayaddison, line 125: This may seem overly complicated compared to, for example:

```python
if getenv('SOURCE_DATE_EPOCH') is not None:
    return NullTranslations(), False
```

...which would achieve similar results.

My thinking here is that the presence/absence of `SOURCE_DATE_EPOCH` should, ideally, minimize the effect on the control flow of the program and build.

If the environment variable results in completely different control flows, then arguably the program being built _is_ different.

My sense is that the ideal roadmap for `SOURCE_DATE_EPOCH` is to introduce it as-required to achieve reproducible builds, but also to keep in mind that it should be removed in future when no-longer-required.

(and to be precise in the case of `sphinx`: a potential time for removal would be when it's possible to perform deterministic documentation builds in all-available-locales -- meaning that the software would be packaged for universal applicability and should build into an identical artefact regardless of the build host's locale)

sphinx/locale/__init__.py
```diff
@@ -222,15 +234,7 @@ def setup(app):
...
     def gettext(message: str, *args: Any) -> str:
-        if not is_translator_registered(catalog, namespace):
-            # not initialized yet
-            return _TranslationProxy(_lazy_translate, catalog, namespace, message)  # type: ignore[return-value]  # NOQA
-        else:
-            translator = get_translator(catalog, namespace)
-            if len(args) <= 1:
-                return translator.gettext(message)
-            else:  # support pluralization
-                return translator.ngettext(message, args[0], args[1])
```
jayaddison: This seems like it's dropping some functionality; `ngettext` isn't referenced elsewhere in the codebase.  Needs more investigation..

jayaddison: Ok, yep: `ngettext` isn't used elsewhere, and plural forms aren't used for any localized strings at the moment (nor do they have test coverage).

When used they do function correctly, and that should be maintained by the migration to the `_lazy_translate` method in this pull request.  Test coverage is probably a good idea, though -- and maybe this change should be unpicked into a separate PR.

jayaddison:
> Hmm.. the original issue report was specifically about the `LANGUAGE` environment variable, and that isn't exercised here at the moment.  Perhaps I'm on the wrong track here.

To confirm: exporting `LANGUAGE` environment values before running `tox` to run unit tests did uncover an unhandled situation here.  Essentially, as noted [here](https://github.com/sphinx-doc/sphinx/pull/10949#discussion_r1009458144), `gettext.translation` consults a number of environment variables to auto-detect a language when the `languages` parameter is empty.

That means that we can't simply configure an empty language list when running in reproducible build mode; `gettext.translation` would again consult those other env vars.  The solution currently in-place is to use a special language tag: `und` (undetermined).  `gettext` shouldn't find any translated resources for that language and will return a `NullTranslations` translator, regardless of the build machine's configured locale(s).

Maddy2317 (approved):

lamby:
> […] we can't simply configure an empty language list when running in reproducible build mode; gettext.translation would again consult those other env vars. The solution currently in-place is to use a special language tag: und (undetermined). gettext shouldn't find any translated resources for that language and will return a NullTranslations translator, regardless of the build machine's configured locale(s).

Thank you for the summary of what is going on here as well as an overview of the proposed solution. It would be great to see this land in Sphinx.

jayaddison:
After looking back at the changes, I think my use of `str` typecasts to patch up a few test failure situations is potentially fragile and/or wasteful.

I'm planning to push a few commits soon to move those typecasts earlier in the code flow - to the declaration of the relevant localized string variables.

Even so: the fact that this typecasting was necessary was discovered by unit test coverage, and that makes me concerned that there could be other locations in the code that are affected too (without test coverage to highlight that).  I don't have a sense for the level of test coverage that's in place here, though.

sphinx/locale/__init__.py
jayaddison: Potential cleanup for this opened in https://github.com/sphinx-doc/sphinx/pull/10979; resolving this thread.

jayaddison: (to clarify for any email notifications: closing this thread related to `ngettext` cleanup -- not closing the entire pull request discussion thread)

sphinx/locale/__init__.py
AA-Turner: > (and to be precise in the case of `sphinx`: a potential time for removal would be when it's possible to perform deterministic documentation builds in all-available-locales -- meaning that the software would be packaged for universal applicability and should build into an identical artefact regardless of the build host's locale)

If it is realistic to achive this, I think it's a good goal -- would you be happy to open a new issue to track the 'locale independent' workstream?

A

AA-Turner:
@jayaddison this is only missing a CHANGES entry, then happy to merge!

A

AA-Turner:
Commit message draft, for later:

```rest
This commit disables Sphinx's localisation features when reproducible
builds are requested, as determined by a non-empty SOURCE_DATE_EPOCH_
environment variable.

The `Reproducible Builds`_ project aims to provide confidence to
consumers of packaged software that the artefacts they're downloading
and installing have not been altered by the environment they were
built in, and can be replicated at a later date if required.

Builds of localised documentation using Sphinx currently account for
a large category of reproducible build testing failures, because the
builders intentionally use varying environment locales at build-time.
This can affect the contents of the ``objects.inv`` file.

During investigation, it turned out that many ``gettext``-localised
values (particularly in Python modules under `sphinx.domains`) were
being translated at module-load-time and would not subsequently be
re-localised.

This creates two unusual effects:

1. Attempting to write a test case to build the same application in
   two different languages was not initially possible, as the
   first-loaded translation catalogue (as found in the 
   ``sphinx.locale.translators`` global variable) would remain in-use
   for subsequent application builds under different locales.

2. Localisation of strings could vary depending on whether the
   relevant modules were loaded before or after the resource
   catalogues were populated.

We fix this by performing all translations lazily so that module
imports can occur in any order and localisation of inventory entries
should occur only when translations of those items are requested.

Localisation can then be disabled by configuring the ``gettext``
language to the ISO-639-3 'undetermined' code (``'und'``), as this
should not have an associated translation catalogue. We also want to
prevent ``gettext`` from  attempting to determine the host's locale
from environment variables (including ``LANGUAGE``).

.. _SOURCE_DATE_EPOCH: https://reproducible-builds.org/docs/source-date-epoch/
.. _Reproducible Builds: https://www.reproducible-builds.org/
```

sphinx/locale/__init__.py
jayaddison: From searching around before doing that, I think that this would be approximately the same as #788 (although that is specific to HTML, currently).

I'll spend some more time reading before deciding whether to add more comments there or to open a separate ticket.

CHANGES
```diff
@@ -83,6 +83,7 @@ Bugs fixed
 * #11192: Restore correct parallel search index building.
   Patch by Jeremy Maitin-Shepard
 * Use the new Transifex ``tx`` client
+* #9778: Disable localisation of the ``objects.inv`` file during Reproducible Builds
```
```suggestion
* #9778: Disable localisation when the ``SOURCE_DATE_EPOCH`` environment
  variable is set, to assist with 'reproducible builds'. Patch by James Addison
```

AA-Turner: @jayaddison I didn't know if you'd like to be credited as Jay (per your username) or James (also per your username!).

jayaddison: Good question - I'm rarely sure myself.  Mostly in commit messages I go by James, so that probably makes sense here.  Thanks for checking :)

AA-Turner:
Thank you @jayaddison!

A

sphinx/locale/__init__.py
AA-Turner: Thank you, I appreciate your help here!

A

jayaddison:
You're welcome - thank you too for the review and merge!  I'll keep a lookout for any issue reports related to this in future, especially in the weeks/months after it's released.

jayaddison:
After further investigation of the original bug #9778 today, I believe that the fix here was unnecessary; see commentary at https://github.com/sphinx-doc/sphinx/issues/9778#issuecomment-1501172176 for details.

It may make sense to scavenge some of the test coverage, but I think that this change should be reverted (and #11306 does that).

PR #10962, opened 2022-11-07
Fix lookup table for multi-word key names (``kbd`` role)

textshell:
It seems the original PR adding multi word key support forgot to add commas and python helpfully just concatenated the strings instead of building the required tuples.


textshell:
CI failure seems unrelated, not sure what is happening there.

textshell:
rebased to current master to clear CI troubles.

Can i do anything else to help getting this merged?

PR #10965, opened 2022-11-10
imgmath: Fix relative file path

### LINKED ISSUE #10944: imgmath failed to resolve image path for files in nested folder
### Describe the bug

Hi. I am using the sphinx.ext.imgmath plugin to generate images for inline math. The images (of math) are correctly generated in `build/html/_images`, but the generated HTML files do not contain correct link to them.

When there are nested files, instead of generating the link (to the image) relative to the file location, the generated links are instead relative to `build/html`. For example, say I have a page `source/foo/index.rst` that contains math. Instead of generating an image tag as `<img src="../_images/math/..."`, Sphinx generate ``<img src="_images/math/..."`` As a result, images of math are not displayed correctly.

This problem seems to be introduced in Sphinx [v5.3.0](https://github.com/sphinx-doc/sphinx/releases/tag/v5.3.0). [v5.2.3](https://github.com/sphinx-doc/sphinx/releases/tag/v5.2.3) does not have this behavior.

### How to Reproduce

Create the following 3 files:

index.rst
```rst
.. toctree::
   :maxdepth: 2
   :caption: Contents:
   :numbered:

   foo/index
```

foo/index.rst
```rst
.. math::
  1 + 1
```

conf.py
```
project = 'Test'
copyright = '2022, Lesley Lai'
author = 'Lesley Lai'

extensions = ['sphinx.ext.imgmath']

exclude_patterns = []

html_theme = 'alabaster'
```

Then run the following commands:
```bash
$ sphinx-build -M html . _build
$ # open _build/html/foo and see
```

### Environment Information

```text
Platform:              win32; (Windows-10-10.0.22623-SP0)
Python version:        3.10.0 (tags/v3.10.0:b494f59, Oct  4 2021, 19:00:18) [MSC v.1929 64 bit (AMD64)])
Sphinx version:        5.3.0
Docutils version:      0.16
Jinja2 version:        3.1.2
```

### Sphinx extensions

sphinx.ext.imgmath

### Additional context

To make `sphinx.ext.imgmath` work correctly, latex needs to be installed in the system.

jschueller:
Closes #10944

### DIFF
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -317,7 +317,8 @@ def html_visit_math(self: HTMLTranslator, node: nodes.math) -> None:
             image_format = self.builder.config.imgmath_image_format.lower()
             img_src = render_maths_to_base64(image_format, rendered_path)
         else:
-            relative_path = path.relpath(rendered_path, self.builder.outdir)
+            bname = path.basename(rendered_path)
+            relative_path = path.join(self.builder.imgpath, 'math', bname)
             img_src = relative_path.replace(path.sep, '/')
         c = f'<img class="math" src="{img_src}"' + get_tooltip(self, node)
         if depth is not None:
@@ -357,7 +358,8 @@ def html_visit_displaymath(self: HTMLTranslator, node: nodes.math_block) -> None
             image_format = self.builder.config.imgmath_image_format.lower()
             img_src = render_maths_to_base64(image_format, rendered_path)
         else:
-            relative_path = path.relpath(rendered_path, self.builder.outdir)
+            bname = path.basename(rendered_path)
+            relative_path = path.join(self.builder.imgpath, 'math', bname)
             img_src = relative_path.replace(path.sep, '/')
         self.body.append(f'<img src="{img_src}"' + get_tooltip(self, node) +
                          '/></p>\n</div>')
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -207,10 +207,9 @@ def render_math(
     """Render the LaTeX math expression *math* using latex and dvipng or
     dvisvgm.
 
-    Return the filename relative to the built document and the "depth",
+    Return the image absolute filename and the "depth",
     that is, the distance of image bottom and baseline in pixels, if the
     option to use preview_latex is switched on.
-    Also return the temporary and destination files.
 
     Error handling may seem strange, but follows a pattern: if LaTeX or dvipng
     (dvisvgm) aren't available, only a warning is generated (since that enables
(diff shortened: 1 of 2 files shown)


jschueller:
@return42 and now ?

return42:
Perfect .. works like a charm:

grafik

-----

IDK why the CI reports an [flake8 issue](https://github.com/sphinx-doc/sphinx/actions/runs/3600688863/jobs/6065620995#step:5:16) not related to this patch.

jschueller:
lets hope it can be merged soon

jschueller:
maybe @marxin or @jfbu ?

jfbu:
@jschueller I rebased your work on current master tip so that we have fresh linting.

@AA-Turner Should we consider fixing the #10944 as done here for 6.0.1 rather than 6.1.0?

jfbu (commented):
In reviewing this I realized #10888 modified the returned path from ``render_math()`` to be absolute rather than relative.  But the docstring still says https://github.com/sphinx-doc/sphinx/blob/a9b0f2708b67a355c5f4969ccbbe9bd695518805/sphinx/ext/imgmath.py#L212
Could the #10888 have been achieved while keeping the behaviour of ``render_math()`` as formerly? And at same time avoid #10944? If not the docstring of ``render_math()`` should be updated.

jfbu (changes requested):
Actually #10888 also modified the returned type of ``render_math()`` but did not update accordingly the docstring, which still says that it also returns the temporary filepath as well as the absolute path of destination image file. I see no mention of this change in CHANGES or in our document. I think it is important that it should be documented somewhere (I don't know to what extent or if at all ``render_math()`` is public API. Edit: surely not public API, but my point is that docstrings should be kept up-to-date with code...).

jschueller:
ok, I updated the docstring

sphinx/ext/imgmath.py
```diff
@@ -209,10 +209,9 @@ def render_math(
     """Render the LaTeX math expression *math* using latex and dvipng or
     dvisvgm.
 
-    Return the filename relative to the built document and the "depth",
+    Return the absolute filename to the built document and the "depth",
```
jfbu: I think you meant to remove "to the built document" part as it referred not to the image file but to the html one.

jschueller: yes

jfbu (approved):
LGTM, thanks
ping @AA-Turner about perhaps adding some brief summary of overall architecture at start of module as the #10888 discussion was somewhat terse and it is hard to follow it a posteriori.  In  particular at 4cb857d5a65c07a49e1b8ece116d9880fae38152 the extension internal dynamics were somewhat refactored and it would be nice to have some brief overview in plain terms of what this all does.

PR #10973, opened 2022-11-14
Remove guides sections ToCs captions from index page

silopolis:
They are redundant with section titles on this page and look weird alone between descriptions and ToCs themselves.


AA-Turner:
The captions are needed for the side-bar headings, unfortunately. I agree though on the duplication.

A

PR #10981, opened 2022-11-17
apidoc: Compile 'excludes' regexps only once

progval:
### Feature or Bugfix
- Bugfix

### Purpose

is_excluded() called fnmatch.fnmatch on each input string, which caused fnmatch.translate and re.compile for each of them.

This is usually fine because both have their own internal LRU cache, but when there are more patterns than the cache's size, this causes every pattern to be recompiled on every call to is_excluded().

### Detail

On a project with a pathological number of excludes (`*` was accidentally expanded before passing to apidoc), this patch drops the total runtime from 80 to 1.6 seconds.

I also considered changing it to a single pattern:

```diff
diff --git a/sphinx/ext/apidoc.py b/sphinx/ext/apidoc.py
index 665b098f6..4f238bd76 100644
--- a/sphinx/ext/apidoc.py
+++ b/sphinx/ext/apidoc.py
@@ -418,8 +418,12 @@ def main(argv: List[str] = sys.argv[1:]) -> int:
     if not args.dryrun:
         ensuredir(args.destdir)
     excludes = [
-        re.compile(fnmatch.translate(path.abspath(exclude)))
-        for exclude in args.exclude_pattern
+        re.compile(
+            "(" + "|".join(
+                fnmatch.translate(path.abspath(exclude))
+                for exclude in args.exclude_pattern
+            ) + ")"
+        )
     ]
     modules = recurse_tree(rootpath, excludes, args, args.templatedir)
 
```

but this does not seem to bring any speed-up


AA-Turner:
Thanks @progval!

A

sphinx/ext/apidoc.py
```diff
@@ -429,7 +429,7 @@ def main(argv: list[str] = sys.argv[1:]) -> int:
         ensuredir(args.destdir)
     excludes = tuple(
         re.compile(fnmatch.translate(path.abspath(exclude)))
-        for exclude in args.exclude_pattern
+        for exclude in dict.fromkeys(args.exclude_pattern)
```
progval, line 432: Out of curiosity, why `dict.fromkeys()` instead of `set()`?

AA-Turner: Probably overthinking this -- I wanted to de-duplicate patterns, but also e.g. you might want to order patterns in case there is a speed boost for more commonly excluded items. ``set`` would loose order, ``dict`` retains order. But this is fairly inconsequential!

Issue #4576, opened 2018-02-07
slow running sphinx in large project


abchenson:
Subject: slow running sphinx in large project
### Problem

We have a sphinx project with about 300 rst files ( about 200 generated stubs, e.g. for each member of each class etc. comparable to pandas docs, 60k lines in the non generated pages).  The build takes over 12mins and each file takes over a second to write, using up most of that time.  This single file writing time is the concern as it seems excessive, and to have been getting worse, so we're worrying about scaling, which might affect our choice of tools.  The amount of code being documented is not that huge present, and this part of the docs may double in size, at which point we'll really be in trouble.  

I know about the -j option already but the scaling is still a worry here if there project e.g. doubled in size.  I've also had a look at https://github.com/sphinx-doc/sphinx/issues/4230 but this seems unrelated as 1.5 was not that much better for us.  Here I have used sphinx 1.6.6.  I tried 1.7.0b2 but this causes the code documentation to fail (see https://github.com/sphinx-doc/sphinx/issues/4573 ) so it's difficult to compare but doesn't seem to help a lot.

The question is really where the bottleneck is here and if anything could be done to significantly increase the speed (either at our end our yours). Any help would be much appreciated!

#### Procedure to reproduce the problem

This is very hard to reproduce with a simple mocked-up example.  Below there is a version of our project with code (and notebooks) replaced with mockups and the rst "scrambled" (alphanumerics randomised apart from navigation directives and roles). The scrambling results in a lot of warnings, but this does not seem to affect the speed issue so I'm hoping it has not effected the cause of the problem.

Unpack into one directory, have the speed_test_requirements/requirements.txt pip installed, and then use the following command run in the docs directory to see the slow running
```
make html
```

[scrambled_slow_example.tar.gz](https://github.com/sphinx-doc/sphinx/files/1703979/scrambled_slow_example.tar.gz)

#### Error logs / results

This mock/scrambled version takes over 10 minutes, a bit quicker than the real version.

### Environment info
- OS: Linux
- Python version: 3.5
- Sphinx version: 1.6.6

marcinpraczko:
Hi, Is there any work on this? I have the same issue - which to be honest slow down my work.
Simple RST update and then times required to rebuild HTML pages:

```
make: Entering directory '/code/docs'
Running Sphinx v1.8.5
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 37 source files that are out of date
[...]

PR #10988, opened 2022-11-23
Add faster ``TocTree._toctree_copy`` method

hofmandl1:
This was pair-programmed with @ax-lothas

As in the standalone html builder the navigation is flattened out for every single html page, the code needs to create a specialized toctree for every html page.

Previously this was done by deep-copying the complete navigation toctree and then stripping out the parts not needed on the particular page.

With this change the code only (deep)-copies the needed parts of the toctree avoiding unnecessary copying+throwing-away

The performance improvements seems to be smaller for smaller page counts and get bigger the more pages are involved:

```
+----------------------------------+-----------+-----------+-------------------+
|         830 Source Files         |   5.3.0   |  5.3.0mod |       Ratio       |
+----------------------------------+-----------+-----------+-------------------+
|          SingleThreaded          |  49.137 s |  43.517 s | 0.885625903087287 |
| Parallel (auto, 16 logical CPUs) |  13.334 s |  12.849 s | 0.963626818659067 |
+----------------------------------+-----------+-----------+-------------------+

+----------------------------------+-----------+-----------+-------------------+
|        6166 Source Files         |   5.3.0   |  5.3.0mod |       Ratio       |
+----------------------------------+-----------+-----------+-------------------+
|          SingleThreaded          | 907.497 s | 521.945 s | 0.575147906825036 |
| Parallel (auto, 16 logical CPUs) | 195.400 s | 150.139 s | 0.768367451381781 |
+----------------------------------+-----------+-----------+-------------------+
```

### Relates
- #4576


sphinx/environment/adapters/toctree.py
```diff
@@ -281,9 +283,41 @@ def get_toctree_ancestors(self, docname: str) -> List[str]:
             d = parent[d]
         return ancestors
 
+    ET = TypeVar('ET', bound=Element)
+
+    def _toctree_copy(self, node: ET, depth: int, maxdepth: int, collapse: bool = False) -> ET:
+        """Utility: deep-copy a TOC but omit things with the semantics of ._toctree_prune.
+
+        !!! MUST be kept consistent with ._toctree_prune !!!
```
AA-Turner: What do you mean by 'consistent' here?

A

hofmandl1: While `_toctree_prune` deletes unneeded stuff from a (previously done) deep copy of the toctree, `_toctree_copy` does the deep copy itself but only copies the parts, that `_toctree_prune` would not delete. So assuming, that `_toctree_prune` would be adjusted in order to e.g. not delete a particular node type, then `_toctree_copy` would also need to be adjusted to copy that particular node type.

TBH we didn't fully understand the different node types involved here. It would probably be best, to completely get rid of `_toctree_prune` in order not to have this consistency problem any more (maybe it is not that big of a problem if there aren't any adjustments to that code done in the foreseeable future). We tried getting rid of `_toctree_prune` but found that it is used in another place and didn't feel confident enough to do the necessary changes.

AA-Turner:
Thanks!

A

PR #10993, opened 2022-11-27
Fix #10984 [DOC] document latex_additional_files behavior with .tex

### LINKED ISSUE #10984: Document latex_additional_files behavior for files with .tex extension.
### Describe the bug

As discussed in https://github.com/sphinx-doc/sphinx/issues/4090, adding files to `latex_additional_files` that end in `.tex` causes them to be appended to the `latexmk` command so that they get compiled automatically.  This can cause problems for files that are intended to be `\input{}` as discussed in https://github.com/sphinx-doc/sphinx/issues/4090.  The resolution there was to append `.tex.txt` to the filename in the [documentation] example, however, the [documentation] does not actually document the behaviour for files ending in `.tex`.

Could the actual intended behaviour please be precisely specified in the  so that users don't need to guess and check?  (Are there any other special cases?)

[documentation]: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-latex_additional_files

### How to Reproduce

The complete behaviour is described in https://github.com/sphinx-doc/sphinx/issues/4090.  Just include an incomplete file with a name ending in `.tex` in `latex_additional_files`.  This files gets appended to the `latexmk` command line, which then attempts to compile it as if it were a complete standalone file.

This is probably the intended behaviour, but if so, please document this.

### Environment Information

N/A

### Sphinx extensions

_No response_

### Additional context

_No response_

jfbu:


jfbu:
Merging this so that this is out of the way, even though we are currently undergoing mypy 0.991 failted tests (cf #10995)

Issue #10992, opened 2022-11-26
TODO tutorial extension fails if 'todo' content contains links


danieleades:
### Describe the bug

The 'todo' directives in the 'todo extension' tutorial don't support content that contains a glossary 'term'. While this is maybe not a show-stopper for this tutorial, it also breaks an entire class of extensions based on this example.

I have an extension (https://github.com/danieleades/sphinx-graph/issues/42) that uses the structure outlined in this tutorial. It allows you to define dependencies between paragraphs of text, and to generate links between them. These paragraphs of text can in principle contain arbitrary content, including glossary terms. Right now, the build breaks as soon as you add glossary term nested inside the content of the directive.

I've created a minimal reproduction using the 'todo' tutorial here - https://github.com/danieleades/todo-extension-bug

I've created a stackoverflow question for my own library [here](https://stackoverflow.com/questions/74575915/custom-sphinx-directive-fails-when-the-content-contains-a-nested-glossary-term), but since this also impacts the 'todo' extension I think this is a genuine bug (or limitation) for extensions of this type. 

### How to Reproduce

```
git clone https://github.com/danieleades/todo-extension-bug.git
cd todo-extension-bug
poetry run make latexpdf
```

you should see the following output-

```
.
.
.
Exception occurred:
  File ".../site-packages/docutils/nodes.py", line 624, in __getitem__
    return self.attributes[key]
KeyError: 'refdocname'
.
.
.
```

### Environment Information

```text Platform:              linux; (Linux-6.0.6-76060006-generic-x86_64-with-glibc2.35) Python version:        3.10.6 (main, Nov  2 2022, 18:53:38) [GCC 11.3.0]) Python implementation: CPython Sphinx version:        5.3.0 Docutils version:      0.19 Jinja2 version:        3.1.2 ```

### Sphinx extensions

None

### Additional context

None

danieleades:
@tk0miya , @AA-Turner i've created a minimal reproduction of this bug in my own extension here- https://github.com/danieleades/sphinx-graph/tree/bug/42/nested-links

I'm sure you're both very busy, but completely stumped, and I'd really appreciate some help! I'm sure it's something super simple.

i'm passing the content of the directive and storing it global state like this-

```python
class Directive(SphinxDirective):
    has_content = True
    required_arguments = 1

    def run(self) -> Sequence[nodes.Node]:
        """Run the directive and return a Vertex node."""

        content_node = nodes.Element()
[...]

PR #10995, opened 2022-11-27
fix flake8 warnings

danieleades:
Looks like there's a few flake8 warnings on master. This PR aims to address those.

The errors stem from the use of type comments. These comments are non-standard syntax, though they are supported by the PyCharm IDE.
Unfortunately pylance, mypy, and flake8 don't like them.
*without* type comments, it's not possible to type annotate 'for' loops on untyped iterators, so in these instances I've suppressed warnings rather than removing the type comments. Possibly there's a better long term solution, but hopefully this gets the CI working again and starts putting some nice green ticks next to the pull requests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+  - repo: https://github.com/asottile/yesqa
+    rev: v1.4.0
+    hooks:
+      - id: yesqa
+        additional_dependencies:
+          - flake8-comprehensions
+          - flake8-bugbear
+          - flake8-simplify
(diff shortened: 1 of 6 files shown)


danieleades:
unfortunately it looks like these fixes introduce new mypy errors...

danieleades:
will trade CI fixes for issue reviews ;)

- #11001 
- #10992

jfbu:
I merged your commit to master via #11035, unfortunately it got squashed with my commit deactivating Python 3.12-dev testing and fixing `tox 4` compatibility rather than keeping separate.  This was due to a mistake of mine on the github web interface.

jfbu:
Sorry I can not force-push to master to fix my squashing mistake

danieleades:
No problem at all!

Issue #1001, opened 2015-01-03
comma after leading optional argument in function signature


shimizukawa:
If an optional argument is at the beginning of the parameter list of a function or method, the comma renders after the closing bracket as opposed to before it.  Here is an example from the Python documentation (notice also that the comma for the final optional argument, in contrast, renders correctly inside the brackets):

range([start], stop[, step])

(from http://docs.python.org/dev/library/functions.html#func-range )

This should instead be--

range([start,] stop[, step]) (or perhaps range([start, ]stop[, step]) )

This occurs even though the corresponding reST documentation is correct:

.. function:: range([start,] stop[, step])

I poked around in the Sphinx code, and it is possible that this is a relevant area (the sphinx.domains.python._pseudo_parse_arglist() function):

https://github.com/sphinx-doc/testing/blob/d5f4e1592a35cd92c16b36c667579647d3d4e854/sphinx/domains/python.py#L37

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1001
- Originally reported by: Chris Jerdonek
- Originally created at: 2012-09-01T19:28:21.873

shimizukawa:
_From Chris Jerdonek on 2012-09-01 17:35:59+00:00_

See also this Python issue (what sparked creating this Sphinx issue):

http://bugs.python.org/issue15831

shimizukawa:
_From Chris Jerdonek on 2012-09-02 03:21:21+00:00_

Given Sphinx's testing structure, what would be the proper way to add tests for changes to the _pseudo_parse_arglist() function that I referenced in the original comment?  Would there be a way for me to add unit tests to test this individual function (or a related function) in isolation?

shimizukawa:
_From Chris Jerdonek on 2012-09-02 16:27:49+00:00_

I attached a first attempt at a patch.  It changes only the visit/depart logic in sphinx/writers/html.py.

With the patch, all of the following functions for example render correctly (as shown below):

{{{
#!python
foo(a)
foo(a, b)
foo([a])
foo(a[, b])
foo([a, ]b)
foo([a, ]b, c)
foo([a][, b])
foo([a, ]b[, c])
foo([a, ]b, c[, d])
foo([a, ]b, c, [d, ]e)
foo([a, ]b, c, [d, ]e[, f])
foo([a, b])
foo([a, b, ]c)
foo(a[, b, c])
foo(a[, b[, c]])
foo(a, [b, [c, ]]d)
[...]

PR #11011, opened 2022-12-03
ENH: enable multiline signatures when too long.

TLouf:
Subject:  enable multiline signatures when they are too long.

### Feature or Bugfix

- Feature

### Purpose
Improve the readability of API references by allowing to set a number of characters a (for instance) Python function's signature cannot exceed. When it would, display each parameter on its own indented line, as should be done in code following style guides like PEP 8.

### Detail
- The proposed implementation introduces a configuration option `python_maximum_signature_line_length`, an integer representing the maximum number of characters that cannot be exceeded by a signature. When negative (the default), there is no maximum, no line break will be introduced no matter how long the signature.
- A `:singlelinesig:` directive still allows to disable this behavior on a per-object basis.

See for the example, with `python_maximum_signature_line_length = 88` set in `conf.py`, the rst below:
```rst
.. function:: fun_with_short_sig(a, b, c)

.. function:: fun_with_long_sig(this_is_one_very_long_argument_name, what_about_this_one='and_its_default')

.. function:: another_fun_with_long_sig(this_is_one_very_long_argument_name, what_about_this_one='and_its_default')
   :singlelinesig:
```
would build to the following HTML:

image

- This is still very much a work in progress, as I've only looked at Python and html, the CSS is surely not perfect and I haven't introduced new tests or modified the docs. In its current state see it more as a proof of concept.

### Relates
#1514 https://github.com/pandas-dev/pandas/issues/49945


emezh:
Hope this will work for C too (and not just for Py). Thanks in advance.

TLouf:
Thanks for showing your interest :) I've implemented the functionality similarly for C and C++.

If a maintainer finds the time, would be good to hear from one of them whether this seems like the way to go, or if there's something I'm not aware of that makes this implementation flawed.

pradyunsg:
@TLouf Thanks for this PR! I'm excited to see this moving forward! ^>^

In case you missed it, this pull request is currently failing continous integration checks since `mypy`'s type checking does not like some of the changes in `_parse_arglist`.

jakobandersen:
From a very brief look:
- the ``addnodes.desc_content()`` should not be used here, as it is reserved for the content of a directive (i.e., the indented stuff with description of an object).
- it would be great if the option ``singlelinesig``  would be more general, e.g., ``param-line-spec`` where one could specify which parameters should be grouped on lines. The specification language probably requires a bit of discussion iteration. Also, see the next comment. Some simple cases could for example be
  - ``1-3,4-``: "let parameter 1 through 3 be on the line with the function name, and the rest on the next line"
  - ``,1-``: "after the opening ``(``, put the rest on the next line"
- is a "max length" character constraint good for HTML? When people are using different screens, different zoom levels, etc. it feels like the line breaking should be dynamical.
  Once upon a time I tried to sketch how the HTML structure and CSS could be set up, but never found time to get into a real implementation: https://github.com/sphinx-doc/sphinx/issues/1514#issuecomment-798458599
- For the C and C++ domains (perhaps Python as well), remember to increment the environment version in the bottom when the ASTs are modified.

pradyunsg:
> is a "max length" character constraint good for HTML?

Yes.

It's the same discussion as why we should have a line length limit and most of the text/prose arguments for websites made in https://mediawiki.org/wiki/Reading/Web/Desktop_Improvements/Features/Limiting_content_width and we've had a similar discussion in the issue about limiting content width in the default Sphinx themes (sorry for not finding the issue number; it's 1am and a quick search didn't surface it).

TLouf:
> the addnodes.desc_content() should not be used here, as it is reserved for the content of a directive (i.e., the indented stuff with description of an object).

Fixed, I introduced a new `desc_parameterline` class to hold parameter lines.

> it would be great if the option singlelinesig would be more general, e.g., param-line-spec

This would be significantly more complex to implement I think, especially because what you describe is a completely different style: namely one with a block indented at the level of the signature's opening parenthesis. What I had in mind is a [`Black`-like style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines), which has every parameter on its own line, indented by one level more than the object's name, which is straightforward to implement in HTML/CSS and other output formats with some kind of list environment. So I would maybe reserve that idea for another PR.

Otherwise, I'm a bit stuck on the LaTeX implementation now, especially for C and C++.  I was thinking of putting the arguments in a `fulllineitems` environment, have the function name and each argument in separate `pysigline`s. But I can't figure out how to wrap the return type and function name of C functions in a `pysigline`.

sphinx/addnodes.py
```diff
@@ -247,6 +247,10 @@ def astext(self):
         return f'({super().astext()})'
 
 
+class desc_parameterline(nodes.General, nodes.Element):
+    """Node for a parameter line."""
```
AA-Turner: Please expand on this description, I can forsee forgetting this PR in the future and being confused as to what a "parameter line" is.

sphinx/domains/c.py
```diff
@@ -686,9 +686,15 @@ def describe_signature(self, signode: Any, mode: str,
 
 
 class ASTParameters(ASTBase):
-    def __init__(self, args: List[ASTFunctionParameter], attrs: ASTAttributeList) -> None:
+    def __init__(
+        self,
+        args: List[ASTFunctionParameter],
+        attrs: ASTAttributeList,
+        multiline: bool = False
```
AA-Turner: "multiline" -> "multi_line" please (only commented on one of them here)
```suggestion
        multi_line: bool = False
```

sphinx/domains/c.py
```diff
@@ -3126,6 +3137,7 @@ class CObject(ObjectDescription[ASTDeclaration]):
     option_spec: OptionSpec = {
         'noindexentry': directives.flag,
         'nocontentsentry': directives.flag,
+        'singlelinesig': directives.flag,
```
AA-Turner: This needs to be solved in a better way, but I'd prefer not to compound the readability problem by adding another flag in the same style. 
```suggestion
        'single-line-signature': directives.flag,
```

AA-Turner (commented):
Brief review. Please add tests and documentation for the proposed feature.

A

doc/usage/configuration.rst
```diff
@@ -2945,6 +2969,18 @@ Options for the Python domain
...
+
+   An integer representing the maximum number of characters that cannot be exceeded
+   by a Python object's signature. When negative (the default), there is no maximum,
+   no line break will be introduced no matter how long the signature. When positive,
+   all objects whose signature exceed the given character limit will have each of
+   their arguments displayed on a separate, indented line. The directive
+   :rst:dir:`single-line-signature` allows to disable this behavior on specific
+   objects.
+
+   .. versionadded:: 6.x
```
AA-Turner: I've mixed feelings on this, how would you feel about just always breaking to one line per argument and having a global (boolean) option to preserve single lines?

A

cjw296: As a potential user of this, I'd really like to be able to specify a threshold (and have Sphinx ship with a sensible default...)
I don't want simple functions with one or two args having their signatures split across lines, and I have an even enough mix that a boolean is going to cause an annoying amount of work no matter which way it is set....

pradyunsg: I tend to agree with @cjw296 -- something like https://installer.pypa.io/en/stable/api/__init__/ should stay on all-entries-on-one-line (instead of breaking that up) because they fit, while something like https://testfixtures.readthedocs.io/en/latest/api.html#testfixtures.compare should get split across multiple lines.

Using the number of characters in the signature is the natural piece for enabling that behaviour with a clear semantic; and the explicit opt-outs on a per-project and per-signature level are also useful to get the exact semantics that an author might want.

To answer the how I'd feel question: As a user, I'd be... annoyed/frustrated with a change as proposed by @AA-Turner; *especially* if the `installer` API example would get broken up into multiple lines after a Sphinx minor/patch update. As a theme author, I'd... implore you to, should you decide to change the default behaviour like that, to handle it like a backwards-incompatible change to a public Python API of this project. :)

emezh: re "When negative (the default), there is no maximum, no line break will be introduced no matter how long the signature".
I assume that **if** the html theme sets a maximum body width (either through a configuration parameter or hardcoded), then the negative default value will not override this, i.e. the signature will be wrapped based on html theme. Perhaps that could be noted/documented in the description of the `python_maximum_signature_line_length` parameter.
For example, see `basic` html theme's `body_max_width` [parameter](https://www.sphinx-doc.org/en/master/usage/theming.html).

doc/usage/configuration.rst
```diff
@@ -2903,6 +2903,18 @@ Options for the C domain
 
   .. versionadded:: 4.0.3
 
+.. confval:: c_maximum_signature_line_length
```
emezh, line 2934: Thanks for putting in support for C and C++ in addition to Py.
Just a thought about individual `<language>_maximum_signature_line_length` variables.
In the project I'm working on, there are both C and Py APIs and the HTML body has the same width. Would it make sense to have a "global" `maximum_signature_line_length` that applies to all languages and if a user wants a different setting for a specific language, then the language-specific `<language>_maximum_signature_line_length` variable can be used to override the default?
This improvement can be done later. I'll be grateful to start without the global one :)

doc/usage/configuration.rst
TLouf: @emezh yes indeed, you can see the line breaks introduced by `python_maximum_signature_line_length` as hard-wrapping, while the soft-wrapping in the view (here the HTML render) can wrap lines independently. I did not make this clear at all in the doc though, so I clarified this point in my last commit ([0bc3b68](https://github.com/sphinx-doc/sphinx/pull/11011/commits/0bc3b68d34c4d49f7cc71413a6d83dd74ef12f2d))

doc/usage/configuration.rst
TLouf: I had thought about that as well, I think it makes sense. It requires very little effort to implement, so since you asked, I went ahead and did so.

TLouf:
I think I now have a solution for the LaTeX implementation, at least I get the kind of result I expect:

image

I cannot claim I understand everything that's going on in the LaTeX style files though, so it probably needs a review from someone with a proper understanding of those.

Also let me know if there's anything else I forgot that should be added for this PR to move forward.

cjw296:
FWIW, if Latex is the blocker for this, any way we could just short cut that to be "do whatever currently happens for Latex"?
(I only ask as I suspect the majority of people who are after this aren't rendering latex docs, I don't think any of my projects are, for example).

sphinx/texinputs/sphinxlatexobjects.sty
```diff
@@ -146,6 +146,22 @@
...
+  \settowidth{\py@argswidth}{#1\sphinxcode{(}}%
+\let\label\spx@label
+  \py@argswidth=\dimexpr\linewidth+\labelwidth-\py@argswidth\relax\relax
+  \item[#1\sphinxcode{(}]
+
+  \@for\param:=#2\do{
+    \param\sphinxcode{,}
+
+  }
+  \item[\sphinxcode{)}#3\strut]
```
jfbu:
```suggestion
  % render each argument on its own line
  \item[#1\sphinxcode{(}\strut]

  \@for\param:=#2\do{
    \param\sphinxcode{,}

  }
  \itemsep\z@skip
  \item[\sphinxcode{)}#3\strut]
  % alternative to above two lines, which aligns the closing parenthesis
  % with the arguments:
  % \noindent\sphinxcode{)}{#3}
```

jfbu (commented):
I suggested a change to the latex code, as some stuff was not needed because it serves only when the ``\py@sigparams`` is used.  The change includes a commented-out alternative which puts the closing parenthesis not at start of line, but to match the listed parameters.

Note : so far "multiline" has been used to refer to a cpp domain feature of sharing object signatures with common description, but here the added ``\pysigmultilinewithargsret`` is to render each argument on its own line... hence some potential confusion with the ``\ifsphinxsigismultiline`` boolean also used in the LaTeX code.

I saw very briefly that writers/latex.py had some change/reuse of ``multiline`` into ``multi_line``, but did not check if any impact on the cpp thing.

TLouf:
> Note : so far "multiline" has been used to refer to a cpp domain feature of sharing object signatures with common description, but here the added \pysigmultilinewithargsret is to render each argument on its own line... hence some potential confusion with the \ifsphinxsigismultiline boolean also used in the LaTeX code.

I renamed that to `pysigwithmultilineargsret`, should avoid confusion.

> I saw very briefly that writers/latex.py had some change/reuse of multiline into multi_line, but did not check if any impact on the cpp thing.

I renamed all `multiline` instances to `multi_line` hence the small change, but the `is_multi_line` property of `desc_signature` nodes is independent of the one that I added for `desc_parameterlist` ones. So it should not impact the cpp domain feature. 

> I noticed some extra white line above closing parenthesis in html rendering (Alabaster 0.7.13):

Thanks for pointing it out, I was using the previous release to make my tests, there must have been a change to `dl` tags margins in the last one. I added a few lines in the basic CSS to force zero-margins on the `dl`s holding parameter lists.

jfbu (commented):
> I renamed that to `\pysigwithmultilineargsret`, should avoid confusion.

ok, what about `\pysiglinesoneperargs` ? Perhaps `\pysigwitharglines` or `\pysigwithargslines`. Unfortunately we can't use underscore character in LaTeX command names for more readable names. 

`\pysigwithonelineperarg` may be best.

sphinx/texinputs/sphinxlatexobjects.sty
```diff
@@ -146,6 +146,21 @@
   \item[{#1\sphinxcode{(}\py@sigparams{#2}{#3}\strut}]
   \pysigadjustitemsep
 }
+\newcommand{\pysigwithmultilineargsret}[3]{%
+  % render each argument on its own line
+  \item[#1\sphinxcode{(}\strut]
+
```
jfbu:
```suggestion
  \leavevmode\par\nopagebreak
  % this relies on \pysigstartsignatures having set \parskip to zero
```

To avoid a page break occurring right above first function argument.

sphinx/texinputs/sphinxlatexobjects.sty
```diff
@@ -146,6 +146,21 @@
...
+  \@for\param:=#2\do{
+    \param\sphinxcode{,}
+
+  }
```
jfbu:
```suggestion
  \@for\param:=#2\do{\param\sphinxcode{,}\par}
```

To match coding style in LaTeX macro files

sphinx/texinputs/sphinxlatexobjects.sty
```diff
@@ -146,6 +146,21 @@
...
+  \itemsep\z@skip
+  \item[\sphinxcode{)}#3\strut]
+  % alternative to above two lines, which aligns the closing parenthesis
+  % with the arguments:
+  % \noindent\sphinxcode{)}{#3}
```
jfbu: The rendering of signatures with multiple arguments  (`\pysiglinewithwithargsret`) does not allow pagebreaks in-between arguments.  With this alternative no `\parbox` is used so a page break is possible.  Above I inhibited it after `(`, but if we insert `)` like here via an `\item` (which is a complicated delayed action LaTeX thing), it is nay impossible (besides core hacks) to prevent a page break occurring before it.  This is why I would recommend considering the alternative described in a code comment above, with an extra `\nopagebreak` as it does prevent a page break before the closing `)`.  However the `)` will be aligned with start of arguments on prior lines, it will not be at start of line.

Capture d’écran 2023-01-23 à 20 35 35

```suggestion
  \nopagebreak\noindent\sphinxcode{)}{#3}
```
EDIT there is a better suggestion in a later https://github.com/sphinx-doc/sphinx/pull/11011#discussion_r1085081511

jfbu: Now it is matter of taste if an isolated `)` at top of page before object description is or not a bad thing.  As far as I understand regarding the location of `)` either at start of line or underneath first character of last argument the [PEP8](https://peps.python.org/pep-0008/) considers both OK. I prefer the latter, which would be the LaTeX output if the above suggested change is applied (which also due to the `\nopagebreak` prevents a page break; in contrast to a `\nopagebreak` before an `\item` which does not work).

jfbu: One could also put the `)` on same line as last parameter.  However we then would like to add a `\nopagebreak` before that last parameter line.  But LaTeX has no simple variant of its `\@for` macro to loop *and* know when we are handling the last item.  This is only provided by some external rather extensive frameworks, or we must assume we have a recent LaTeX and can then tap into its "LaTeX3" programming layer, which however would make maintenance more complex to future maintainers as it raises the level of  LaTeX knowledge needed.
<sub>I can always add devilish TeX macros doing this but it would make maintenance even more impossible to others. EDIT: it is also possible to insert `\nopagebreak\par` before each `\param` of the `\@for` loop rather than after it, so we can locate the closing parenthesis on same line as last argument and prevent pagebreaks all the way.  But it is more complex to avoid page break only at the last one.</sup>

TLouf: Hmmmm I don' t know, I initially put the closing parenthesis dedented (in LaTeX but also in other formats) to follow Black's style. To quote from its [doc:](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines) "Also, having the closing bracket dedented provides a clear delimiter between two distinct sections of the code that otherwise share the same indentation level". So their argument is that by dedenting the parenthesis you see more clearly the start of the docstring. But well in documentation where most of the time the arguments and the docstring have a different style anyway, I don't know how relevant that is... And then there's this whole caveat about avoiding page breaks that you mention. So personally I don't mind accepting the tradeoff offered by your first suggestion above. If anyone has a strong opinion on this, let us know.

jfbu: > If anyone has a strong opinion on this, let us know.

I don't have a strong opinion but a slight preference for the closing brace aligned with first characters of arguments...

It is feasible to have it dedentet and at same time to prevent a page break but I have to plunge a bit into bowels of LaTeX's lists, o refresh my memory about what are the right dimensional parameters to use here (will do this).

jfbu:
```suggestion
  % fulllineitems sets \labelwidth to be like \leftmargin
  \nopagebreak\noindent\kern-\labelwidth\sphinxcode{)}{#3}
```

This dedents the `)`, and prevents a page break right above it.  As the `)` is alone on its line, and somewhat to the left, the vertical space to object description looks a bit large but it is exactly the same as with `single-line-signature`, as one can check experimenting with using for example `)xxx....xxxx` in the LaTeX code.  Also tested if object description starts itself by a list.

jfbu: > the vertical space to object description looks a bit large

But this effect goes away with presence of type annotation.  Here is an example with our own documentation (the `python_maximum_signature_line_length` was set to very small to force one line per argument rendering)
Capture d’écran 2023-01-24 à 11 34 18

(the annoying footnote markers in PDF are explained by the ``latex_show_urls = 'footnote'``)

jfbu: another example showing how very long type annotations wrap
Capture d’écran 2023-01-24 à 11 40 47

I forgot to say that parameters are left-aligned with description. 

I am fine with applying this which keeps in sync LaTeX and HTLM rendering.  The above suggested change of not using `\item`  for the closing `)` avoids a page break.

Regarding LaTeX, this PR looks good to me.

If we merge soon, the doc ``versionadded`` directive arguments in ``doc/usage/configuration.rst`` should be ``6.2.0``, not ``6.x``.

jfbu: Example with nesting:
Capture d’écran 2023-01-24 à 11 55 41
Due to the way LaTeX is configured by default for nested lists, the indent from left compared to object name decreases the deeper we are in a sub-list.

doc/usage/configuration.rst
```diff
@@ -666,6 +666,23 @@ General configuration
...
+   respectively, :confval:`python_maximum_signature_line_length`,
+   :confval:`c_maximum_signature_line_length` and
+   :confval:`cpp_maximum_signature_line_length`). When negative (the default), there
+   is no maximum, no line break will be introduced no matter how long the signature.
+   When positive, all objects whose signature exceed the given character limit will
+   have each of their arguments displayed on a separate, indented line. This
+   introduces hard line breaks, however, it does not override a potential soft
+   wrapping introduced in the format in which the documentation is built. The
+   directive :rst:dir:`single-line-signature` allows to disable this behavior on
+   specific objects.
```
jfbu:
```suggestion
   directive option :ref:`single-line-signature <pyfuncsinglelinesig>`
   (here for the example of the :rst:dir:`py:function` directive) allows to disable
   this behavior on specific objects.
```

doc/usage/restructuredtext/domains.rst
```diff
@@ -231,6 +231,14 @@ The following directives are provided for module and class contents:
       Describe the location where the object is defined.  The default value is
       the module specified by :rst:dir:`py:currentmodule`.
 
+
```
jfbu:
```suggestion
   .. _pyfuncsinglelinesig:

```

jfbu (commented):
Suggested edit to fix linking.  But it appears I do not know how to link to an option of a ReST directive via a suitable role (I should read the docs...) so I used `:ref:`.

jfbu (approved):
LaTeX aspects fine for me. Thanks for contribution!

doc/usage/restructuredtext/domains.rst
TLouf: Found it! Can be linked with 
```
:rst:dir:`py:function:single-line-signature`
```
The text of the link is then **py:function:single-line-signature**

jfbu: 🥇

sphinx/texinputs/sphinxlatexobjects.sty
TLouf: That's a great solution, thanks a lot for digging into that!

(and it's always good to be reminded of the extent of your ignorance hahaha)

jfbu: @TLouf thanks to stimulation from your PR, I have updated master at 55170d01d to replace a hard-coded `\emph` by `\sphinxparam`.  This means once your PR is merged, it is even possible (to LaTeX geeks) to redefine `\sphinxparam` to do extra customization such as shifting horizontally the parameter on its line (styling inside is done by `\DUrole` mark-up).  But... as was mentioned elsewhere many projects do not bother about building the PDF docs... anyway let's be thorough... I like the way our own Sphinx doc renders to PDF with your PR and all signatures rendered one param per line.  The presence of type annotations avoids this visual artefact of increased vertical distance which I experience when a single `)` is all the way to the left on its own.

emezh:
Just wondering what else needs to be done before this is merged?

cjw296:
Looks like it's approved, just some conflicts to resolve and then a merge. Looks like that'll have to be @TLouf or a Sphinx maintainer such as @jfbu ...

TLouf:
I was seeing more PRs being merged after @jfbu's approval, so I've been waiting for someone to give me the go-ahead before resolving conflicts (so that I wouldn't have to do it several times).

@AA-Turner does this look good to you?

AA-Turner:
@TLouf if you’ve time could you update this branch please?

I will try and look at it before the end of April (likely either before Saturday or after the 24th).

A

TLouf:
Ok great! I've just updated the branch.

AA-Turner:
Hi @TLouf,

I have reviewed and pushed the above series of commits.

In brief, the changes are:

1. Don't change the HTML 4 writer (it is frozen pending removal)
2. Make the type of ``*maximum_signature_line_length`` ``int | None``, I think ``None`` is more natural to signify 'no limit' than a negative number.
3. Note in documentation that the only action taken is splitting the parameter list, i.e. if one had a limit of 80 characters and a 90 character function name, we would not split the function name into parts.
4. Keep ``is_multiline`` (for ``desc_signature`` nodes) the same, as the attribute is used by e.g. Breathe. It would be nice to change to ``is_multiline_signature`` or etc, but that can be considered later.
5. **larger change**: rename ``single-line-signature`` to domain-specific names: for Python, ``single-line-argument-list``, and for C/C++ ``single-line-parameter-list``. I did this to reflect common verbiage which differs between various languages, though there is a compelling argument for full unification.
6. Refactored the tests to split up the various cases.
7. Use the ``multi_line_parameter_list`` attribute on signature nodes rather than passing around an argument through various functions.
8. **larger change**: I removed the ``desc_parameter_line`` node type that you added, as it is not strictly needed, and complicated the domains a little further. The trade-off here is that either domains need to emit and handle the additional node type (and therefore also all writers), or writers use an amount of internal state to track multi-line parameter lists.
  \
  Personally, I think that the latter is the better of the two options, but a compelling argument could be made for reverting that commit and keeping the new ``desc_parameter_line``.

I have added a CHANGES entry, as that was missing.

The last outstanding item is considering if support can be added for the built-in Javascript domain, which is the only one missing at the moment -- @TLouf would you have time to investigate this?

Thanks,
Adam

jfbu:
I will push some LaTeX test.  The situation with the code currently in `sphinxlatexobjects.sty` is that it has remained much more stable since 2016 than other LaTeX support, extensions/improvements were mostly done not by modified mark-up but by internal change to how LaTeX behaves when the PDF is actually build.  This is hard to test; but here this PR adds a new LaTeX writer mark-up so enters into easy enough testing.  For legacy reasons `sphinxlatexobjects.sty` uses control sequences with `\pysig...` prefix but they are not exclusive to Python domains, they are used for all domains as they are generated by the LaTeX writer from the docutils nodes.

jfbu:
> I will push some LaTeX test.

I have been a bit confounded by the following problem: https://github.com/sphinx-doc/sphinx/blob/master/tests/roots/test-root/objects.txt starts with
```
.. function:: func_without_module(a, b, *c[, d])

   Does something.
```

The syntax appears to match https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-signatures explanation about 

> For functions with optional parameters that don’t have default values (typically functions implemented in C extension modules without keyword argument support), you can use brackets to specify the optional parts:

But this input syntax appears to inhibit this PR mechanism.

```rest
.. function:: fun_with_long_sig_B(this_is_one_very_long_argument_name,\
              what_about_this_one='and_its_default', *c)

   The description

.. function:: fun_with_long_sig_C(this_is_one_very_long_argument_name,\
              what_about_this_one='and_its_default', *c[, d])

   The description

```
and ``maximum_signature_line_length = 10`` in `conf.py` then `make html` and I get
Capture d’écran 2023-04-08 à 11 17 25

which shows no linewrap in second function, the one with the strange signature in the  `index.rst` source.

Now, will the answer simply be that the ``fun_with_long_sig_C`` is simply wrong mark-up, but then why do we use it in our `objects.txt` test file... please pardon the Python noob here

TLouf:
@AA-Turner  Thanks a lot for such a thorough review! I agree with most of the changes, just about your 5th point:

- >  **larger change**: rename `single-line-signature` to domain-specific names: for Python, `single-line-argument-list`, and for C/C++ `single-line-parameter-list`. I did this to reflect common verbiage which differs between various languages, though there is a compelling argument for full unification.

  I would argue for unification in favour of `single-line-parameter-list`. To quote [Python's docs](https://docs.python.org/3/faq/programming.html#faq-argument-vs-parameter): 

  > [Parameters](https://docs.python.org/3/glossary.html#term-parameter) are defined by the names that appear in a function definition, whereas [arguments](https://docs.python.org/3/glossary.html#term-argument) are the values actually passed to a function when calling it"

  So I would talk about parameters when documenting a function's signature. And NumPy's docstring style agrees with that. Anyway, leaving arguments of authority aside, since most people use the words parameters / arguments interchangeably, no user will hate Sphinx for not using their preferred terminology. On the other hand, people writing docs for both C/C++ and Python will definitely hate Sphinx when they have to go back to the docs of the other language to understand why the option works for a language but not for the other.

- > The last outstanding item is considering if support can be added for the built-in Javascript domain, which is the only one missing at the moment -- @TLouf would you have time to investigate this?

  My latest commit implements it, I just need to write the tests now.

- In the CHANGES entry I'd add a mention of @jfbu 's work on the LaTeX part. I just did a rudimentary implementation but the code in the end is mostly his.

- My first new commit fixes the hyperlink to the directive option, there's actually a way to make it work, even though it does not render great.

TLouf:
@jfbu yep the current implementation does not cover that case. However,

```python
def compile(source[, filename[, symbol]]):
    ...
```
is not correct python code,it raises a syntax error. So no auto-documented functions (which are the vast majority in API references) will use this syntax. Personally the only place I remember seeing functions documented this way is the documentation of Python's standard library. And they seem to be moving away from it (see [Python's 2 docs](https://docs.python.org/2/library/functions.html?highlight=open#open) for `open` and [Python 3's](https://docs.python.org/3/library/functions.html?highlight=open#open)), although I could not find a discussion of the Python team where this is mentioned. I guess @AA-Turner knows more about this than I do.

At the moment, if I simply pass `multi_line_parameter_list` to `_pseudo_parse_arglist` in the python domain, the above function will be rendered as:
```
def compile(
    source
    [
    , filename
    [
    , symbol
]]
)
```
which looks completely off. If we end up implementing it, what would be the desired format? Something like
```
def compile(
    source[,
    filename[,
    symbol
]]):
```
?

AA-Turner:
> is not correct python code,it raises a syntax error

The syntax is used for optional arguments, and has existed for a long while (inspired by C?) —- look in the CPython documentation for many examples and at the desc_optional (from memory) node for information.

Please go ahead and unify on single-line-parameter-list; thank you for your reasoning. (Please also push a change adding Jean-François to the pitch authors, thank you for pointing this out!)

I’ll next have time for a thorough review towards the end of the month, to set expectations & provide advanced warning on timelines.


Thanks,
Adam

jfbu:
@AA-Turner and @TLouf.  I added at https://github.com/jfbu/sphinx/tree/wip-sig-multiline an additional commit but did not push it here as I am not happy.  This commit causes the `test_build_latex.py::test_build_latex_doc()` which executes 6 (!) PDF builds (pdflatex, lualatex, xelatex each with manual or howto) to all use `python_maximum_signature_line_length = 1`. 
```diff
diff --git a/tests/test_build_latex.py b/tests/test_build_latex.py
index 5dce900e2..e82940143 100644
--- a/tests/test_build_latex.py
+++ b/tests/test_build_latex.py
@@ -95,7 +95,10 @@ def skip_if_stylefiles_notfound(testfunc):
     "engine,docclass",
     product(LATEX_ENGINES, DOCCLASSES),
 )
-@pytest.mark.sphinx('latex')
+@pytest.mark.sphinx('latex',
+                    confoverrides={
+                        'python_maximum_signature_line_length': 1,
+                    })
 def test_build_latex_doc(app, status, warning, engine, docclass):
     app.config.intersphinx_mapping = {
         'sphinx': ('https://www.sphinx-doc.org/en/master/', None),
```

 In this way, its is checked the new LaTeX mark-up causes no PDF build failures.  I fought for a while with adding directly `app.config.python_maximum_signature_line_length = 1` in a `'xelatex'` conditional branch only so that both with and without situations are tested, but this works only if I reduce ``product(LATEX_ENGINES, DOCCLASSES)`` to only use ``'xelatex'``.  Obviously it is time I cease worrying about building the circa 700 PDF pages of our document and do read them at last.  I thought about importing myself the ``setup()`` of the Python domain, but dropped the idea.  I did not yet add a test on output mark-up as I was not happy with this situation.  Feel free to cherry-pick that commit or do something else.  Don't worry about credits line, as I am listed among maintainers (despite I would fail Python 101 class in college), it is not needed.

jfbu:
@TLouf About the handling of this bracket syntax, your proposal for the format, if we implement the support, looks good to me.

TLouf:
I managed to write something that works for the HTML builder, however it complicates the code quite significantly and I think it will even more for the LaTeX builder, so I'm starting to doubt it's actually worth implementing the support. I The problem is, nested optionals span multiple parameters, so multiple lines in multi-line mode, and they imply different rules for the placement of the comma. See for instance the example below

```rst
.. function:: foo([this_is_one_very_long_argument_name[, c]], b, what_about_this_one[, and_this_other_one])
```
which is built as follows with HTML in my latest commits:
```
foo(
    [this_is_one_very_long_argument_name[,
    c,]]
    b,
    what_about_this_one[,
    and_this_other_one,]
)
````
(in the end I decided not to put the closing square bracket with the closing bracket of the signature as in the example given above, since there can be required parameters after optional ones in this syntax, so that's why I decided to always close at the end of a parameter line)
The first line in the parameter list has to be started in the `desc_optional`, the second in the `desc_parameter`, and you have both cases too for ending the line. In LaTeX it's even more problematic because you have nested `\sphinxoptional`, inside which we'd need to insert line breaks. The solution used at the moment to loop over a list of parameters separated by commas does not work, as this way we cannot break inside a `\sphinxoptional`, as far as I know. So implementing the support would mean a significant rewrite for LaTeX as far as I can tell.  And this is (very much) a edge case -- at least for Python, maybe in Javascript docs people use the syntax more.

So let me know what you think, and @jfbu please let me know if you know a way to make it work without too much additional code.

**Comment by jfbu** [comment]
@TLouf I will push a LaTeX commit which renders like this, which is your current HTML rendering:
Capture d’écran 2023-04-11 à 10 55 22

Your latest commits are currently breaking the tests you added for the new feature, so don't blame LaTeX for all red crosses after my imminent push :wink:

**Comment by jfbu** [comment]
Ah wait, I notice that in my latest version as shown in screenshot the nested optional `[` is not before but after the comma.  This is ironical because initially I started by having a version doing it `[ ,`  not `, [` but then at later stage I forgot entirely; the issue I had is about using `\sphinxcode{,}` which typesets the comma in a special way; it is to obtain that consistently that I modifed my first attempt and reached the current one.  Well, this will have to rest a while, will come back later.

**Comment by jfbu** [comment]
@TLouf 

master PDF:
Capture d’écran 2023-04-11 à 12 12 14

HTML with current master (should be same as 6.1.3) differs. Closing brackets are better positioned, but nested opening brackets less so:

Capture d’écran 2023-04-11 à 12 29 51

with this PR, HTML is not modified and PDF now looks alike HTML for those single-line signatures with optional arguments:

PDF:

Capture d’écran 2023-04-11 à 12 12 47

HTML:

Capture d’écran 2023-04-11 à 12 33 34

I used this input:

```rest
.. function:: foobar1([a, ]b, c[, d])

.. function:: foo3([a[, c[, d]]], b, e[, f])
```

For nested optionals in **foo3()**, the commas are now before the brackets in PDF rendering, which does match HTML but is a change which may be interpreted as regression, as  the former PDF behavior for opening brackets looked better than HTML, or as a fix, as PDF differed from HTML and now looks the same. If such nested optional arguments make any sense to start with!

Probably  I can transfer my latest commit hack from `\pysigwithonelineperarg` to `\pysiglinewithargsret` as it was precisely about this but I would rather not and get both case of signatures on single line and signatures with one line per parameter be handled at sphinx.writers.latex level (I.e. the commas should be placed at their final positions already and no core TeX hack should be used to exchange their positions with the opening square brackets).  But this would make PDF differ again from HTML so would be sort of demanding an HTML change as well.



Regarding one line per param, the new LaTeX code obtains the output from a `\sphinxparam` modification not from handling the commas as delimiters, as was the case in initial versions of this PR which evolved since.  It is not an issue if the commas are hidden in braces now.

jfbu:
@TLouf thanks for all this, CI passes.  Sadly I decided to merge master in this branch as rebasing on master caused conflict but I should have expected that this would not let the rebase on master be conflict free yet.  As it stands there is this conflict while rebasing in sphinx/domains/python.py for commit 82f1b9894
```
def setup(app: Sphinx) -> dict[str, Any]:
    app.setup_extension('sphinx.directives')

    app.add_domain(PythonDomain)
    app.add_config_value('python_use_unqualified_type_names', False, 'env')
<<<<<<< HEAD
    app.add_config_value('python_display_short_literal_types', False, 'env')
=======
    app.add_config_value('python_maximum_signature_line_length', -1, 'env')
>>>>>>> 82f1b9894 (ENH: enable multiline signatures when too long.)
    app.connect('object-description-transform', filter_meta_fields)
    app.connect('missing-reference', builtin_resolver, priority=900)
```
In the end we will rebase on master for merge and probably do a squash.  But that will probably not be prior to Adam return.  Don't hesitate force-pushing to https://github.com/sphinx-doc/sphinx/pull/11011/commits/9ce3e3470302548c67c35643c2e3dd9f44a2048f  if you feel my merge of master caused more harm than none.

picnixz:
> The building nested blocks could be [A [, B[, C [, D [, E]]]],] F, G, H. But if this stuff is at very end, then no strange comma before last ]. Now I will go check in "Python for first grade dummies" a reference which should address and elucidate what is reasonable in real life syntax with such optional optional optional optional optional params.

Consider the following example: if `filter_map` is a function that accepts 1) an iterable, 2) a predicate and an iterable, or 3) a predicate, a transformation function and an iterable, we would document it as:

`filter_map([pred [, func],] iterable)`

in order to support:

- `filter_map(iterable)` -- same as `filter(None, iterable)`
- `filter_map(pred, iterable)` -- same as `filter(pred, iterable)`
- `filter_map(pred, func, iterable)` -- same as `filter(pred, map(func, iterable))`.

In particular, `func` can only be supplied if `pred` is supplied and in order to keep a syntax similar to `filter` or `map`, we need them in front. 

By the way, if there are a lot of parameters with small names, but one parameter with a very long name, won't it be better to have them on two lines ? (this is actually what I wanted more). Would it make sense to actually use the indentation used in `.rst` document instead and break at the locations the user requested ? (I don't think a user will actually write a function with very long parameter names and not break them in the document itself).

jfbu:
> Would it make sense to actually use the indentation used in `.rst` document instead and break at the locations the user requested ?

@picnixz It seems one would need for that to intervene very early on in https://github.com/sphinx-doc/sphinx/blob/b6e6805f80ad530231cf841e689498005cf2bb96/sphinx/directives/__init__.py#L85-L95 as it removes the backslashes from user mark-up.  Using
```diff
diff --git a/sphinx/directives/__init__.py b/sphinx/directives/__init__.py
index 517a52900..ded000ce5 100644
--- a/sphinx/directives/__init__.py
+++ b/sphinx/directives/__init__.py
@@ -87,6 +87,7 @@ class ObjectDescription(SphinxDirective, Generic[ObjDescT]):
         Retrieve the signatures to document from the directive arguments.  By
         default, signatures are given as arguments, one per line.
         """
+        print(self.arguments)
         lines = nl_escape_re.sub('', self.arguments[0]).split('\n')
         if self.config.strip_signature_backslash:
             # remove backslashes to support (dummy) escapes; helps Vim highlighting
@@ -224,6 +225,7 @@ class ObjectDescription(SphinxDirective, Generic[ObjDescT]):
 
         self.names: list[ObjDescT] = []
         signatures = self.get_signatures()
+        print(signatures)
         for sig in signatures:
             # add a signature node for each signature in the current unit
             # and add a reference target for it
```
and trying `make html` with
```
.. function:: foo3([a[,\
              c[,\
              d]]], b, e[, f])
```
I got
```
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
['foo3([a[,\\\nc[,\\\nd]]], b, e[, f])']                                       
['foo3([a[,c[,d]]], b, e[, f])']
```

jfbu:
> Consider the following example: if `filter_map` is a function that accepts 1) an iterable, 2) a predicate and an iterable, or 3) a predicate, a transformation function and an iterable, we would document it as:

Thanks @picnixz for real-life illustration!

sphinx/texinputs/sphinxlatexobjects.sty
jfbu, line 1: You can now remove lines L#150--158 (only  line 159 should stay `\def\sphinxoptionalextraspace{0.5mm}`)

(update, see my other comment, I will do it myself as part of other refactoring)

jfbu: There are some other related needed changes with  `\sphinxparamcommaoneperline` business (which now as it stands has no effect) and `\sphinxparamcomma`.  Perhaps now the `,` is not needed in mark-up.  As I very soon will be here very sparsely I will look at this right now and push directly.

jfbu (commented):
Thanks, the LaTeX is much easier now, I will update the branch to follow-up the refactoring as said in my comments <s>above</s> below.

TLouf:
Just pushed some commits that enable the text writer to handle the optional syntax correctly and that complement tests. The latest addresses the `FIXME` you left @jfbu, now the comma placement relative to closing brackets is consistent with other output formats, and is handled by the writer.

The choice I made for this placement is the following:
```
foo(
  a[,
  b[,
  c,]]
)
```
instead of placing the last comma at the end of the line, because I wanted it to be consistent with
```
foo(
  [a[,
  b,]]
  c,
)
```
which is the correct syntax, according to #1001  (where they give the example of `foo(a, [b, [c, ]]d)`).

The only thing I see now as an arguable choice is the trailing comma on the last parameter. It's just personal preference, I like to see a comma on each parameter line, but I won't mind if others wish to remove it, and the change is straightforward to make.

@picnixz 
> By the way, if there are a lot of parameters with small names, but one parameter with a very long name, won't it be better to have them on two lines ? (this is actually what I wanted more). 

Doesn't the current release of Sphinx enable that? I mean, for an HTML output for instance, if the width of the signature's container is limited, this parameter with a very long name will be on a single line and the others wrapped to the next one, right?

> Would it make sense to actually use the indentation used in .rst document instead and break at the locations the user requested ? (I don't think a user will actually write a function with very long parameter names and not break them in the document itself).

They will if they use `autodoc`. But I think your idea makes sense. I'd suggest opening a new issue to discuss it tough, since it'd imply a very different approach from this PR's, as @jfbu pointed out.

sphinx/texinputs/sphinxlatexstyletext.sty
```diff
@@ -113,11 +113,10 @@
 
 % Special characters
 %
-% Modified internally when rendering signatures with one line per param
-% If redefining this, mind that it is followed by explicit "," in input.
-\let\sphinxparamcomma\empty
-% The "," in mark-up will have been swallowed if this is executed.
-\def\sphinxparamcommaoneperline{\sphinxcode{,}}
+\def\sphinxparamcomma{, }% by default separae parameters with comma + space
```
jfbu: typo here s/separae/separate/ needed

sphinx/texinputs/sphinxlatexstyletext.sty
```diff
@@ -113,11 +113,10 @@
...
+% if signature is rendered with one line per param, this is used instead
+% (the \texttt makes the comma slightly more distinctive)
+\def\sphinxparamcommaoneperline{\texttt{,}}
```
jfbu, line 119: Now that the comma does not play a crucial role in the LaTeX aspects, we could actually revert to the former hardcoded comma (plus space, as anyhow TeX will ignore it if it is at end of line in case of one param-per-line).

But well, this gives complete customizability.  I replaced the formerly used  `\sphinxcode{,}` in favor of `\texttt{,}` as there is no reason that a change to `\sphinxcode` should impact the commas in function signatures.

sphinx/writers/latex.py
```diff
@@ -806,10 +806,8 @@ def visit_desc_parameterlist(self, node: Element) -> None:
                                        for c in node.children]
         # How many required parameters are left.
         self.required_params_left = sum(self.list_is_required_param)
-        self.param_separator = r'\sphinxparamcomma, '
+        self.param_separator = r'\sphinxparamcomma '
```
jfbu, line 800: The space token here is gobbled by TeX when gathering the `sphinxparamcomma`-named control sequence and it serves to avoid a letter following immediately if any (but that should not happen as it will be a priori either a `}` or a  `\sphinxparam`) to create an undefined LaTeX control sequence.

sphinx/writers/latex.py
```diff
@@ -806,10 +806,8 @@ def visit_desc_parameterlist(self, node: Element) -> None:
...
         self.multi_line_parameter_list = node.get('multi_line_parameter_list', False)
-        if self.multi_line_parameter_list:
-            self.param_separator = self.param_separator.rstrip()
```
jfbu: I removed this, as it became unneeded (see above comments).

sphinx/texinputs/sphinxlatexstyletext.sty
jfbu: ah well I relaunch the Windows CI which had this occasional error we sometimes encounter but now I recall I must push another commit... sorry for inconvenience. Will wait though for Windows CI to complete.

jfbu:
> The only thing I see now as an arguable choice is the trailing comma on the last parameter. It's just personal preference, I like to see a comma on each parameter line, but I won't mind if others wish to remove it, and the change is straightforward to make.

Same opinion here.

picnixz:
@TLouf 

> Doesn't the current release of Sphinx enable that? I mean, for an HTML output for instance, if the width of the signature's container is limited, this parameter with a very long name will be on a single line and the others wrapped to the next one, right?

Err, actually I didn't check myself. I think it might also depend on the theme you are using. I remember having some weird rendering where the line-break occurred after the type annotation but I don't remember which version of Sphinx I was using or if I had some additional extensions. I'll need to check that later.

> They will if they use autodoc. But I think your idea makes sense. I'd suggest opening a new issue to discuss it tough, since it'd imply a very different approach from this PR's, as @jfbu pointed out.

What I meant is that you use the indentation not of the .rst only but of the source code itself. So, whether they use autodoc or not, it does not matter. People likely won't have a single python function written on a single line spanning 280 characters (at least, they will break it, or they have very large screen which is not my case since I do a hard wrap at 79 characters as recommended by PEP8). 

I'll probably open an issue concerning this idea next week or when time permits.

TLouf:
Ok now the output for multiline parameter lists with optional parameters should conform to the syntaxes shown in #1001. 

@picnixz 
> I remember having some weird rendering where the line-break occurred after the type annotation

Ah that may very well be the case, didn't think of that.

> What I meant is that you use the indentation not of the .rst only but of the source code itself.

Ok that makes perfect sense then!

TLouf:
@jfbu let me know if I can be of any help for the LaTeX tests, I'm not sure I understood the problem you mentioned in a previous comment. It sure would be good to have tests for this feature, but seeing how few tests there are for LaTeX outputs, I haven't ventured into writing them myself.

AA-Turner:
@TLouf is there anything specificly you'd like me to review since I last looked at this PR? Anything I can help you with?

A

TLouf:
Then apart from the changes I highlighted in review comments, let me know if more tests are needed, in particular for LaTeX and text, and if you agree with the multiline syntax for optional parameters, and in general with leaving a trailing comma after the last parameter.

jfbu:
@TLouf 


> ....I'm not sure I understood the problem you mentioned in a previous comment. It sure would be good to have tests for this feature, but seeing how few tests there are for LaTeX outputs, I haven't ventured into writing them myself.

Testing LaTeX is primarily testing the output of the latex builder, and I can add this at later date (I am very busy until at least mid-May).

Then, there are the tests actually doing a PDF build which check the LaTeX macros do not create failure. Currently only ``test_build_latex.py::test_build_latex_doc``.  It does only one pass, so it is not really a complete test of our PDF build (LaTeX requires at least three passes, even more sometimes due to cross-references, indices, table of contents), but it is already very slow, particularly because we test `xelatex` and `lualatex`, and we test both `’howto’` and `’manual’`.  My earlier remark is that I wanted to configure this process to use app.config.python_maximum_signature_line_length = 1 only for example with ‘xelatex’ but at that time I could not adapt ``test_build_latex.py::test_build_latex_doc`` and its product(LATEX_ENGINES, DOCCLASSES) to achieve that, due to the when this configuration is actually activated.  I don’t have time to look at this closely, but at that time, if not separating this into two tests I could either not test the maximum signature length setting or test it but then with all engines.  I wanted to test it only with `’xelatex’` so that builds are tested both with and without the setting.

(by reply-via-email so sorry for possible formatting issues in the end)

TLouf:
@jfbu 
Thanks for the clarification! I see two ways to achieve that. The first is by modifying the `parametrize` list, as done in 30041a6. The other is to keep the product the way it is and skip inside the test function, as follows:
```python
if python_maximum_signature_line_length is not None:
    if engine == 'xelatex':
        app.config.python_maximum_signature_line_length = python_maximum_signature_line_length
    else:
        pytest.skip("only testing for xelatex to reduce testing time.")
```
Personally I prefer the first solution since it entails fewer changes and no skipped test.

Regarding the latex writer tests, I can write them myself if you're too busy, I don't think it'll pose any problem.

jfbu:
@TLouf 
I checked quickly  and I have a problem at my locale with https://github.com/sphinx-doc/sphinx/pull/11011/commits/30041a6ae70c7662e7b1af5a8e29f711124abf5d. I have no time now but maybe tonight (CET). I don't understand why it succeeded in our  CI as it fails at my locale:
```
tests/test_build_latex.py ......F.  
```
I have some idea why perhaps it failed at my locale. It is `test_build_latex_doc[xelatex-howto-1]` which causes build failure
```
(xelatex/sphinxtests.toc

! LaTeX Error: Something's wrong--perhaps a missing \item.
```
and it seemingly fails while inputting the `.toc` file which must be a remnant of the previous xelatex run with manual class.  The LaTeX details are fuzzy to me as I don't have time to examine this more closely now. A chapter entry in the toc file can cause problems if run with article class but why precisely this shows as missing item is not yet elucidated.

- recently on master branch we have added a patch b6e6805f8 which would help debug this, but might make the merge more complex
- my intent was not to run xelatex twice, once with your added feature and once without. The point is that potential failure will not come from engine, it is nay impossible for basic LaTeX as the one of this PR to cause failure with one engine and not another, so I wanted to use xelatex to test the feature, but not increase from 6 to 8 build jobs: they are lengthy enough.  If I had wanted to increase from 6 to 8 I would have picked pdflatex. I wanted to use xelatex to be run with the config variable set, and only with the config variable set.

> +\# Only running test with `python_maximum_signature_line_length` not None with xelatex to
+\# reduce testing time.
 
it only reduces compared to doing 6 times 2... I wanted to keep at most 6 build runs.

Besides I am not sure your test actually works. There are no `\pysigwithonelineperarg` in the  sphinxtests.tex file in the xelatex output dir which should be the one from the last build job.  It will be easier to clarify this with b6e6805f8 which needs adjustment to match the idea of your 30041a6ae70. 

The problem with
```
    if engine == 'xelatex':
        app.config.python_maximum_signature_line_length = python_maximum_signature_line_length
    else:
        pytest.skip("only testing for xelatex to reduce testing time.")
```
is as I reported (but I was not clear) that setting app.config.python_maximum_signature_line_length at this location does not work in my experience.  I had done this kind of conditional branch and observed that the config variable was not obeyed.


I am intrigued why at my locale it seems the third xelatex run encountered an already existing .toc file which causes it to fail, but not on our CI, and I don't even yet have a clear understanding as I simply do not have to the time to check why this failure happened at all.

I suggest you (edit) *either* force-push to remove  30041a6ae70 (edit) *or* <s>and then</s> merge master so that we can have easier investigate this. Well I tried now merging master, there are no conflicts and I checked the merge of `tests/test_build_latex.py` looks ok.  After merging master the test succeeds at my locale:

```console
tests/test_build_latex.py ........                                                                                                                               [100%]

========================================================================= slowest 25 durations =========================================================================
8.32s call     tests/test_build_latex.py::test_build_latex_doc[lualatex-manual-None]
8.00s call     tests/test_build_latex.py::test_build_latex_doc[lualatex-howto-None]
7.96s call     tests/test_build_latex.py::test_build_latex_doc[xelatex-manual-None]
7.55s call     tests/test_build_latex.py::test_build_latex_doc[xelatex-manual-1]
7.25s call     tests/test_build_latex.py::test_build_latex_doc[xelatex-howto-None]
7.18s call     tests/test_build_latex.py::test_build_latex_doc[xelatex-howto-1]
6.58s call     tests/test_build_latex.py::test_build_latex_doc[pdflatex-manual-None]
4.84s call     tests/test_build_latex.py::test_build_latex_doc[pdflatex-howto-None]
...[setup durations removed]...
```

There are no `\pysigwithonelineperarg`  to be found. But I am not sure if test is run with "None" or "1" last in your modified pytest.mark.parametrize.

Running the test with make not tox I see that the None are executed before the "1".
```
tests/test_build_latex.py::test_build_latex_doc[pdflatex-manual-None] PASSED                                                                                     [ 12%]
tests/test_build_latex.py::test_build_latex_doc[pdflatex-howto-None] PASSED                                                                                      [ 25%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-manual-None] PASSED                                                                                     [ 37%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-howto-None] PASSED                                                                                      [ 50%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-manual-None] PASSED                                                                                      [ 62%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-howto-None] PASSED                                                                                       [ 75%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-manual-1] PASSED                                                                                         [ 87%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-howto-1] PASSED                                                                                          [100%]
```
So we seem to have instance of the problem I had already encountered about the timing of setting the python domain config variable.

TLouf:
Ok, indeed there is something I don't understand either with the configuration. The most probable culprit is the `pytest.mark.sphinx` decorator in my opinion. Setting it as I've done in 8037a5b -- crucially with `freshenv=True` -- fixes the issue according to my local tests, please confirm on your side.

jfbu:
> @jfbu Ok, indeed there is something I don't understand either with the configuration. The most probable culprit is the `pytest.mark.sphinx` decorator in my opinion. Setting it as I've done in [8037a5b](https://github.com/sphinx-doc/sphinx/commit/8037a5b605810293c8ae0724d48a7fc027994ae7) -- crucially with `freshenv=True` -- fixes the issue according to my local tests, please confirm on your side.

Yes, confirmed :+1:


```
platform linux -- Python 3.10.8, pytest-7.2.2, pluggy-1.0.0 -- /home/bla/bla/sphinx/.tox/py310/bin/python
cachedir: .tox/py310/.pytest_cache
libraries: Sphinx-6.2.0+/8037a5b60, docutils-0.19
base tempdir: /tmp/pytest-of-someone/pytest-39
rootdir: /home/bla/bla/sphinx, configfile: pyproject.toml
collected 6 items                                                                                                                                      

tests/test_build_latex.py::test_build_latex_doc[pdflatex-howto] PASSED                                                                           [ 16%]
tests/test_build_latex.py::test_build_latex_doc[pdflatex-manual] PASSED                                                                          [ 33%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-howto] PASSED                                                                           [ 50%]
tests/test_build_latex.py::test_build_latex_doc[lualatex-manual] PASSED                                                                          [ 66%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-howto] PASSED                                                                            [ 83%]
tests/test_build_latex.py::test_build_latex_doc[xelatex-manual] PASSED                               
```

and

```
$ grep -rl pysigwithoneline /tmp/pytest-of-someone/pytest-39/
/tmp/pytest-of-someone/pytest-39/root/_build/latex/pdflatex/sphinxtests.tex
/tmp/pytest-of-someone/pytest-39/root/_build/latex/sphinxlatexobjects.sty
```

and after merging-in master:

```
$ grep -rl pysigwithoneline /tmp/pytest-of-someone/pytest-40/
/tmp/pytest-of-someone/pytest-40/root/_build/latex/pdflatexhowto/sphinxtests.tex
/tmp/pytest-of-someone/pytest-40/root/_build/latex/pdflatexmanual/sphinxtests.tex
/tmp/pytest-of-someone/pytest-40/root/_build/latex/sphinxlatexobjects.sty
```

> 
> EDIT: it can actually work with something similar to what I added in [30041a6](https://github.com/sphinx-doc/sphinx/commit/30041a6ae70c7662e7b1af5a8e29f711124abf5d) by simply setting `freshenv=True` in the `pytest.mark.sphinx` marker. It's cleaner so I think I'll do it that way.

As you prefer, the way of https://github.com/sphinx-doc/sphinx/pull/11011/commits/8037a5b605810293c8ae0724d48a7fc027994ae7 looks good to me (as I can understand the syntax!).

One could imagine variants like testing this PR configuration only with `howto` class and all three engines.  But as it stands it is ok with me.

I do have to review our test suite, e.g. we are not testing here build of Japanese documents which go via yet another LaTeX engine (`platex`)

ping @AA-Turner could we do
```diff
diff --git a/tox.ini b/tox.ini
index 512aa5f74..f49ea91c1 100644
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ setenv =
     PYTHONWARNINGS = error
     PYTEST_ADDOPTS = {env:PYTEST_ADDOPTS:} --color yes
 commands=
-    python -X dev -X warn_default_encoding -m pytest --durations 25 {posargs}
+    python -X dev -X warn_default_encoding -m pytest -vv --durations 25 {posargs}
 ```

For locale testing with `tox` most of the time we will test only one feature hence use a specific unit test not launch the full test suite (as it is done by CI anyhow).  Then the enhanced verbosity of pytest is helpful for example to get names of individual tests in case the unit test as here launches multiple ones (another example is with latex tests relative to tables). Related: https://github.com/sphinx-doc/sphinx/pull/11333#discussion_r1169797535

**Comment by AA-Turner** [comment]
*[I'm starting to work towards preparing the 6.2.0 release this weekend -- setting expectations that this PR won't be in 6.2]*

A

**Comment by jfbu** [comment]
Thanks for this ping @AA-Turner! As far as LaTeX is concerned it is all good for me to merge into 6.2.0, I hope my numerous comments did not create impression to contrary.  Testing is fine as @TLouf  has solved the issue I was not able to address properly a few weeks (or days?) ago, I will now add an additional check of mark-up, and push an update to latex.rst to mention the `\sphinxparam`, `\sphinxparamcomma` and `\sphinxparamcommaoneperline` that we added to help customize signatures. For LaTeX all looks good to  me. I am ok with having a comma after the final parameter of a function, which is an aspect @TLouf was submitting for further discussion.

File: doc/latex.rst [hunk 16]
```diff
@@ -1477,7 +1477,12 @@ Macros
      ``\sphinxkeyboard``
 
   .. versionadded:: 6.2.0
-     ``\sphinxparam``, ``\sphinxsamedocref``
+
```
jfbu, line 1481: arrgh, this empty line is a left-over from re-flowing paragraph.  It does not matter for output, nevertheless I will push a commit to remove. Sorry.

TLouf:
@AA-Turner Just pushed tests for text output, they're very similar to the ones for HTML output so it was straightforward. From my point of view this PR is ready for a final review. If you have time to find the time to squeeze this one into 6.2.0, it would be a really nice addition I think :)

jfbu (approved):

jfbu:
Thanks for updates and merge @AA-Turner!  I am a bit lost again in our release policy. This was merged to master and shows in CHANGES as being part of upcoming `7.0.1`.   Btw there is no `7.0.x` branch indeed.  In the past we used patch number bump only for urgent hotfixes, which this is not.

Related to my query:  in [doc/latex.rst](https://github.com/sphinx-doc/sphinx/blob/master/doc/latex.rst) I documented on line 1482 the change to appear in `7.1.0`.  (sorry but in the past I never understood how to link directly to a specific line of a `.rst` file on github interface and I have not tried again since)

AA-Turner:
> This was merged to master and shows in CHANGES as being part of upcoming `7.0.1`. Btw there is no 7.0.x branch indeed.

Thank you for pointing this out, I hadn't noticed that typo (7.0.1 vs 7.1.0)! You are of course right that the unreleased section of CHANGES should read ``7.1.0``. 

I didn't create a ``7.0.x`` branch, on the rationale that if we needed a hot-fix it would be very simple to create the branch at a later date.

``7.1.0`` was also the version I bumped all the other changes to, e.g.: https://github.com/sphinx-doc/sphinx/blob/53b9b28366ab129ab3640b3d69a761888daeac05/doc/usage/configuration.rst?plain=1#L682-L699

A

jfbu:
@AA-Turner ah it was a typo! I would not have imagined that :smile:,  I thought to have monopoly on such things... 

Sorry I had missed you bumped to `7.1` elsewhere in the doc.  Quite clear now why there is no `7.0.x` branch.

And finally,  thanks for showing me the `?plain=1#L682-L699` syntax for `rst` files! That's a huge leap forward in my github mastery...

AA-Turner:
> And finally, thanks for showing me the `?plain=1#L682-L699` syntax for `rst` files! That's a huge leap forward in my github mastery...

Appending ``?plain=1`` to the URL makes the file appear in the GitHub viewer like a standard code file, and then one can use the standard tools when clicking on a line number -- <kbd>Copy permalink</kbd> is then the one I use to have the line numbers automatically appended to the URL.

If you use the [refined github](https://github.com/refined-github/refined-github/)
 browser extension, there's a button added automatically to view as 'code' or 'preview', e.g.:

image

A

PR #11024, opened 2022-12-10
Prevent warning filters from being discarded

### LINKED ISSUE #11022: Runtime warnings filters are ignored
### Describe the bug

Runtime warnings which are [usually filtered out by Numpy](https://github.com/numpy/numpy/blob/main/numpy/__init__.py#L232) still appear when building documentation using the `autodoc` extension, because [a warning context manager is used when importing modules](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autodoc/importer.py#L50).
This means that warnings filters which are set by imported modules are not persistent across consecutive imports, thus displaying warnings which should be hidden.

### How to Reproduce

```bash
python -m venv venv
source venv/bin/activate
pip install sphinx
pip install numpy==1.23.5
pip install astropy==5.1.1
```

```python
import numpy
import astropy.table
# Expected behaviour, no warning is displayed.
```

```python
from sphinx.ext.autodoc import import_module
import_module("numpy")
import_module("astropy.table")
# Unexpected behaviour, displays "<frozen importlib._bootstrap>:228: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 80 from C header, got 88 from PyObject".
```

### Environment Information

```text 
Platform:              linux; (Linux-6.0.6-arch1-1-x86_64-with-glibc2.36) 
Python version:        3.10.8 (main, Oct 13 2022, 21:13:48) [GCC 12.2.0]) 
Sphinx version:        5.3.0 
Docutils version:      0.19 
Jinja2 version:        3.1.2
```

### Sphinx extensions

sphinx.ext.autodoc

### Additional context

_No response_

teobouvard:
Using a context manager to ignore `ImportWarning` during module imports also prevents libraries setting warning filters ([such as NumPy](https://github.com/numpy/numpy/blob/4ca8204c5c13b3a5c9482772813e67184f3f47c8/numpy/__init__.py#L231)) from being persistent, since the filters they set are discarded when exiting the context manager.

I might be missing something, but it seems the usage of this context manager to ignore such warnings is redundant with the [default warning filters](https://docs.python.org/3/library/warnings.html#default-warning-filter) installed by Python.

Closes #11022


AA-Turner:
Thanks @teobouvard!

A

PR #11025, opened 2022-12-11
Migrate coveragepy config into pyproject.toml

skirpichev:
### Feature or Bugfix
- Refactoring

### Notes

So far, only flake8 config now in the setup.cfg.  In principle, you can workaround PyCQA/flake8#234 with [flake518](https://github.com/carstencodes/flake518).


pyproject.toml
```diff
@@ -215,3 +215,17 @@ markers = [
...
+parallel = true
+source = ['sphinx']
+
+[tool.coverage.report]
+exclude_lines = [# Have to re-enable the standard pragma
+                 'pragma: no cover',
+                 # Don't complain if tests don't hit defensive assertion code:
+                 'raise NotImplementedError',
+                 # Don't complain if non-runnable code isn't run:
+                 'if __name__ == .__main__.:']
```
AA-Turner:
```suggestion
exclude_lines = [
    # Have to re-enable the standard pragma
    'pragma: no cover',
    # Don't complain if tests don't hit defensive assertion code:
    'raise NotImplementedError',
    # Don't complain if non-runnable code isn't run:
    'if __name__ == .__main__.:',
]
```

PR #11033, opened 2022-12-15
No blanket lint suppression

danieleades:
removes the use of 'blanket' noqas.

These statements can accidentally obscure other issues. In general, lint suppression scope should be as small as possible

i've also added a pre-commit lint and associated github workflow action in a separate commit to ensure no blanket noqas can sneak back in...

PR is currently stacked on #10995


jfbu:
Thanks for all your work. Now merging (via a rebase...)

PR #11037, opened 2022-12-19
Replace the nondetermistic default locale_dir kwarg with fixed value for reproducibility

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort I noticed that sphinx could not be built reproducibly.

This is because the `locale_dir` keyword argument for the `init_console` method defaults to the path that method is implemented in (using Python's `__file__` mechanism), and then that (non-deterministic) default argument is included in Sphinx's own documentation. This makes the build unreproducible as it varies depending on the directory you create the documentation from.

This patch uses the typical `fn(arg1=None) […] if arg1 is None: arg1=defaultval` pattern to avoid this, and was first submitted by me to Debian in [bug #1025801](https://bugs.debian.org/1025801).


sphinx/locale/__init__.py
```diff
@@ -144,13 +144,15 @@ def init(
 
 
 def init_console(
-    locale_dir: str = _LOCALE_DIR,
+    locale_dir: Optional[str] = None,
```
AA-Turner:
```suggestion
    locale_dir: str | None = None,
```

PR #11044, opened 2022-12-21
Document autosummary template variable "objtype"

timhoffm:
The variable was missing in the documentation, but it clearly exists:

https://github.com/sphinx-doc/sphinx/blob/984416247370a87ad947d89f334a8ce2d8ea6558/sphinx/ext/autosummary/generate.py#L346

Note: I've not listed the objtype values used by the deprecated Documenters in `sphinx/ext/autodoc/depreacted.py`. Since they are deprecated, I assume users should not use them anymore.


AA-Turner:
Thanks!

PR #11045, opened 2022-12-22
Create a 'search field' component for themes

### LINKED ISSUE #10791: Remove heading from search box in sphinx13 style
The heading takes extra space and adds a bit of clutter. Since the only element in the section is the input field we can as well only show the input field and use the placeholder to spell out "Search".

Current (left) / suggestion (right):

grafik grafik

Since the search box is a component (`searchbox.html`), which may need to stay as is for compatibility, I suggest to introduce a new reduced component (e.g. `searchfield.html`) and use that for sphinx13.

Let me know what you think. I can come up with a PR if there is interest for it.

timhoffm:
### Feature or Bugfix
- Feature. Closes #10791.

### Purpose
searchfield.html is similar to the existing searchbox.html, but does not have the heading "Quick Search". Instead, it uses "Search" as a placeholder text. This gives a cleaner and less cluttered appearance.

Before:
image

After:
image

### Detail
The searchfield component is added to the basic theme and thus can be used in any derivative theme. It is not used by default for backward compatibility. Also searchbox.html was deliberately not modified for backward compatibility.

The searchfield component is made the default in the new sphinx13 theme.

CSS adaptions ensure that the top of the search field is aligned with the top of the main content.

### DIFF
--- a/doc/_themes/sphinx13/layout.html
+++ b/doc/_themes/sphinx13/layout.html
@@ -32,7 +32,7 @@
 {%- block content %}
 <div class="document">
   <div class="sphinxsidebar" role="navigation" aria-label="main navigation">
-    {%- include "searchbox.html" %}
+    {%- include "searchfield.html" %}
     <div class="sphinxsidebar-navigation__contents">
       <h3>{{ _('On this page') }}</h3>
       {{ toc }}
--- a/doc/_themes/sphinx13/static/sphinx13.css
+++ b/doc/_themes/sphinx13/static/sphinx13.css
@@ -68,6 +68,7 @@ div.sphinxsidebar {
     overflow-wrap: break-word;
     margin: 0;
     padding-right: 15px;
+    padding-top: 0.5em;
     font-size: 1em;
 }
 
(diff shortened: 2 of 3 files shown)


mgeier:
Please note that the `searchbox` also has a "Hide Search Matches" feature:

https://github.com/sphinx-doc/sphinx/blob/cc314f13e8a98393ab018d83d8957a724a6f338a/sphinx/themes/basic/static/sphinx_highlight.js#L99-L110

You might want to duplicate this for your `searchfield`.

Alternatively, you might be able to just re-use the `#searchbox` ID, which would give you this feature for free.

timhoffm:
Thanks for the feedback. I've gone back to `#searchbox`, which seems more robust compared to duplicating style and logic.

mgeier:
Yes, thanks, my comment has been fully addressed.

However, I'm not the right person to approve or reject this PR as a whole.

timhoffm:
ping @AA-Turner. You approved the fundamental idea in https://github.com/sphinx-doc/sphinx/issues/10791#issuecomment-1229479591. Would you be willing to review? Thanks.

AA-Turner (approved):

AA-Turner:
Thanks!

A

PR #11056, opened 2022-12-31
add SIM401 lint

danieleades:
I'm personally less convinced by this lint. I think there's a net benefit, but a small one. Coming from Rust, I do like the functional approach, and i like that the variable only gets bound in one place.

More a matter of taste than a clear benefit. I vote for merging, but I'm open to being convinced otherwise


sphinx/builders/latex/theming.py
```diff
@@ -111,10 +111,10 @@ def load_builtin_themes(self, config: Config) -> None:
...
     def get(self, name: str) -> Theme:
         """Get a theme for given *name*."""
-        if name in self.themes:
-            theme = self.themes[name]
-        else:
-            theme = self.find_user_theme(name) or Theme(name)
+        theme = self.themes.get(
+            name,
+            self.find_user_theme(name) or Theme(name)
+        )
```
AA-Turner, line 117: Most are fine, I don't like this one though, due to the `or` in the `default` argument.

AA-Turner:
Closing out-of-date PR (feel free to update & re-open).

A

PR #11057, opened 2022-12-31
Add SIM113 lint

danieleades:


sphinx/environment/adapters/toctree.py
```diff
@@ -209,9 +209,10 @@ def _entries_from_toctree(toctreenode: addnodes.toctree, parents: List[str],
                         if not (subtocnode.get('hidden', False) and
                                 not includehidden):
                             i = subtocnode.parent.index(subtocnode) + 1
-                            for entry in _entries_from_toctree(
-                                    subtocnode, [refdoc] + parents,
-                                    subtree=True):
+                            entries = _entries_from_toctree(
+                                subtocnode, [refdoc] + parents, subtree=True
+                            )
+                            for entry in entries:
```
danieleades: @AA-Turner this one had me stumped last time i tried to introduce this lint.

Only reason I can think the behaviour changes here is that the list that's being iterated over is modified inside the loop. That's far from ideal, so I think this is crying out for a refactor.

Do you have any insights here?

PR #11060, opened 2022-12-31
improve type annotations

danieleades:
this is sisyphean, but i feel like i'm slowly reducing the local entropy in the universe.

There's definitely a chronic overuse of argument defaults in this repo.... if only there was a way to lint for functions with default arguments, where the arguments are always fully-specified in usage. I think there'd be a lot of examples in this repo


danieleades:
Fyi https://github.com/MartinThoma/flake8-simplify/issues/165

danieleades:
closing. this will be a tricky rebase, and I'd rather address the changes in smaller, more organised commits

PR #11064, opened 2023-01-01
De-glob mypy whitelist for 'sphinx.domains.*'

danieleades:


danieleades:
@AA-Turner, looks like CI is failing. Don't think this is anything to do with my changes

sphinx/domains/__init__.py
```diff
@@ -276,7 +278,7 @@ def directive(self, name: str) -> Callable | None:
         fullname = f'{self.name}:{name}'
         BaseDirective = self.directives[name]
 
-        class DirectiveAdapter(BaseDirective):  # type: ignore
+        class DirectiveAdapter(BaseDirective):  # type: ignore[valid-type,misc]
```
danieleades, line 283: i think this error can be removed, rather than suppressed. Just need to know the type of `self.directives`. Any ideas?

sphinx/domains/c.py
```diff
@@ -1547,8 +1547,14 @@ def __setattr__(self, key: str, value: Any) -> None:
         else:
             return super().__setattr__(key, value)
 
-    def __init__(self, parent: Symbol, ident: ASTIdentifier,
-                 declaration: ASTDeclaration, docname: str, line: int) -> None:
+    def __init__(
+        self,
+        parent: "Symbol",
```
AA-Turner:
```suggestion
        parent: Symbol,
```

sphinx/domains/c.py
```diff
@@ -1743,10 +1749,15 @@ def candidates() -> Generator[Symbol, None, None]:
...
             Symbol.debug_indent -= 2
 
-    def _symbol_lookup(self, nestedName: ASTNestedName,
-                       onMissingQualifiedSymbol: Callable[[Symbol, ASTIdentifier], Symbol],
-                       ancestorLookupType: str, matchSelf: bool,
-                       recurseInAnon: bool, searchInSiblings: bool) -> SymbolLookupResult:
+    def _symbol_lookup(
+        self,
+        nestedName: ASTNestedName,
+        onMissingQualifiedSymbol: Callable[["Symbol", ASTIdentifier], Symbol | None],
```
AA-Turner:
```suggestion
        onMissingQualifiedSymbol: Callable[[Symbol, ASTIdentifier], Symbol | None],
```

sphinx/domains/math.py
```diff
@@ -93,8 +93,9 @@ def resolve_xref(self, env: BuildEnvironment, fromdocname: str, builder: Builder
                      typ: str, target: str, node: pending_xref, contnode: Element
                      ) -> Element | None:
         assert typ in ('eq', 'numref')
-        docname, number = self.equations.get(target, (None, None))
-        if docname:
+        result = self.equations.get(target)
+        if result:
+            docname, number = result
```
AA-Turner, line 98: can we keep the tuple unpacking here?

sphinx/domains/rst.py
```diff
@@ -267,8 +267,9 @@ def resolve_xref(self, env: BuildEnvironment, fromdocname: str, builder: Builder
                      ) -> Element | None:
         objtypes = self.objtypes_for_role(typ)
         for objtype in objtypes:
-            todocname, node_id = self.objects.get((objtype, target), (None, None))
-            if todocname:
+            result = self.objects.get((objtype, target))
+            if result:
+                todocname, node_id = result
```
AA-Turner, line 272: can we keep the tuple unpacking here?

sphinx/domains/rst.py
```diff
@@ -278,8 +279,9 @@ def resolve_any_xref(self, env: BuildEnvironment, fromdocname: str, builder: Bui
                          ) -> list[tuple[str, Element]]:
         results: list[tuple[str, Element]] = []
         for objtype in self.object_types:
-            todocname, node_id = self.objects.get((objtype, target), (None, None))
-            if todocname:
+            result = self.objects.get((objtype, target))
+            if result:
+                todocname, node_id = result
```
AA-Turner, line 284: can we keep the tuple unpacking here?

danieleades: No, not really. If you unpack the tuple before checking the `get` call returned anything, there's no way to know (from a type safety perspective) that just because `todocname` is set, `node_id` must also be set. By promoting the 'Optional' to the level of the tuple, and moving the tuple unpacking inside the if statement, then everything is correctly typed

danieleades: I *could* use the walrus operator here to shave off a line. Would still need to unpack the tuple inside the if block

sphinx/domains/math.py
danieleades: no, the tuple unpacking has to be inside the `if` statement for this to be correctly typed.

I could potentially use the walrus operator here, which would look something like-

```python
if (result := self.equations.get(target)):
    docname, number = result
```

sphinx/domains/__init__.py
AA-Turner: `dict[str, type[docutils.parsers.rst.Directive]]`

danieleades: scratch that. Doesn't solve this error. Will leave it as is for now

PR #11075, opened 2023-01-02
improve typing in 'sphinx.highlighting'

danieleades:


AA-Turner:
py38 doesn't support runtime subscripting (CI failing)

A

danieleades:
> py38 doesn't support runtime subscripting (CI failing)
> 
> A

rats. will park this for now

AA-Turner:
> rats. will park this for now

quoting probably works, or just use `typing.Type`.

A

PR #11078, opened 2023-01-03
Fix relative references in SVGs generated by ``sphinx.ext.graphviz``

### LINKED ISSUE #11077: Relative references in graphviz svgs have the wrong root
### Describe the bug

When including a graphviz graph that itself has a reference to a file by relative path (embedded image or hyperlink), the link target is wrong in the rendered docs when using svg format.

This is because the rendered graphviz file is put in the `_images` folder in the built documentation, irrespective of where the original documentation file was, hence any relative path (e.g.`../../_stativ/image.svg`) is now invalid since the root of the graph changed.

I have a PR that would fix this by rewriting relative URLs in generated SVGs

### How to Reproduce

index.rst:
```
graphviz
========

.. graphviz::
   :align: center

   digraph test {
     foo [label="foo", URL="#graphviz", target="_parent"]
     bar [label="bar", image="./_static/images/test.svg"]
     baz [label="baz", URL="./_static/images/test.svg"]
     foo -> bar -> baz
   }
```

conf.py:
```
extensions = ['sphinx.ext.graphviz']
exclude_patterns = ['_build']
html_static_path = ["_static"]
```

The rendered svg still points to e.g. `./_static/images/test.svg`, even though since it's now in `_images/`, it should be `../_static/images/test.svg`

### Environment Information

```text
Platform:              linux; (Linux-6.1.1-arch1-1-x86_64-with-glibc2.36)
Python version:        3.9.15 (main, Nov 18 2022, 14:40:14)
[GCC 12.2.0])
Sphinx version:        6.0.0
Docutils version:      0.19
Jinja2 version:        3.0.3
```

### Sphinx extensions

```python
`['sphinx.ext.graphviz']`
```

### Additional context

_No response_

Panaetius:
closes #11077


grahamwaters:
could this help?
https://github.com/grahamwaters/ReferenceMedic

Panaetius:
> could this help? https://github.com/grahamwaters/ReferenceMedic

I don't think it does. It doesn't support svg files at the moment. And it seems to just fix paths for a very specific folder layout, whereas in this case we know exactly from which source document to which target folder relative reference need to be changed in sphinx, and what type of attributes can be affected, so searching svg (XML) with a regex sounds like a bad idea when we can instead specifically look for e.g. `href` attributes to update and not accidentally update something that's not relevant. Especially since ONLY links/references need to be updated, paths in plain text do no need to be changed.
Also that library only deals with file paths, but we need to fix URLs in general (Like `#my-anchor` needs to be changed to `../#my-anchor`). The targets can be anywhere in the file structure, essentially, depending on where the source *.rst file was, not in a fixed directory structure.

AA-Turner:
Thanks @Panaetius!

A

PR #11080, opened 2023-01-03
Fix #6744: support for seealso directive should be via an environment

### LINKED ISSUE #6744: LaTeX: support for seealso directive should be via an environment to allow styling
The LaTeX writer support for the "seealso" directive produces this kind of mark-up:

```latex
Text before


\sphinxstrong{See also:}


This is contents of seealso directive



Text after
```

from

```rest
Text before

.. seealso::

   This is contents of seealso directive

Text after
```

This makes impossible styling on the LaTeX side the way the contents are rendered (for example use some background color).

This is in contrast to most other elements of the LaTeX writer output, hence is a bug.

Admittedly [our docs](https://www.sphinx-doc.org/en/1.5/markup/para.html#directive-seealso) says

> The content of the seealso directive should be a reST definition list. Example:

Still, a ``sphinxseealso`` environment would be useful to customize the included definition list. Besides there does not seem much rationale for restricting the contents of "seealso" directive to be definition lists only. Our doc says

>  For the HTML output, it is shown boxed off from the main flow of the text.

but without a LaTeX "sphinxseealso" environment, similar effect or even non similar is impossible for PDF output.

**Environment info**
- Sphinx version: 2.1.x and earlier

**Additional context**
- motivated from a [discussion of the Sphinx user group](https://groups.google.com/d/msgid/sphinx-users/qoelng%247t62%241%40blaine.gmane.org)

jfbu:
See #6744 for description

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -16,6 +16,9 @@ Features added
 Bugs fixed
 ----------
 
+* #6744: LaTeX: support for seealso directive should be via an environment
+  to allow styling.
+
 Testing
 --------
 
--- a/doc/latex.rst
+++ b/doc/latex.rst
@@ -1431,6 +1431,14 @@ Environments
      parameters, such as ``noteBorderColor``, ``noteborder``,
      ``warningBgColor``, ``warningBorderColor``, ``warningborder``, ...
 
+- Environment for the :rst:dir:`seealso` directive: ``sphinxseealso``.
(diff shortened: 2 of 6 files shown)


jfbu:
Linting via Ruff fails as is the case currently for master at 08ca934ff for unrelated reasons. I will merge now because this code is a priori not impacting PDF output, only LaTeX mark-up.

PR #11085, opened 2023-01-05
Remove flake8 plugins in favour of Ruff

danieleades:
I don't think we need both, since ruff fully replaces these plugins


danieleades:
I'd like to remove isort as well, however with the default configuration, it will re-sort the imports. If i get time i'll mess about and see if i can't match the behaviour with a little configuration. That's a job for a different PY maybe.

@charliermarsh

pyproject.toml
```diff
@@ -81,8 +81,6 @@ docs = [
 ]
 lint = [
     "flake8>=3.5.0",
-    "flake8-comprehensions",
-    "flake8-bugbear",
     "flake8-simplify",
     "isort",
```
AA-Turner, line 85:
```suggestion
    "isort",
    "ruff",
```

pyproject.toml
```diff
@@ -81,10 +81,9 @@ docs = [
...
+    "ruff",
```
danieleades, line 86: @AA-Turner 

we should pin this, as ruff is moving very fast and PRs will break because of changes not related to the PRs (eg. #11086)

PR #11092, opened 2023-01-05
Suppress ``ValueError`` in ``apply_source_workaround``

AA-Turner:
cc @gaborbernat 

Does this fix your issue?

A

### DIFF
--- a/sphinx/util/nodes.py
+++ b/sphinx/util/nodes.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import re
 import unicodedata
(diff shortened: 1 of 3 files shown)


sphinx/util/nodes.py
```diff
@@ -152,7 +153,8 @@ def apply_source_workaround(node: Element) -> None:
 
     # workaround: literal_block under bullet list (#4913)
     if isinstance(node, nodes.literal_block) and node.source is None:
-        node.source = get_node_source(node)
```
gaborbernat, line 155: It does 👍

gaborbernat:
Thanks!

PR #11106, opened 2023-01-07
Apply ruff 0.0.213 linting advices on c40f4f904f3b053

jfbu:
ping @AA-Turner 

is this how we are supposed to do this?

should it go to 6.1.x branch rather?

I applied `ruff` at my locale and applied the patch, this should match the one from https://github.com/sphinx-doc/sphinx/actions/runs/3861663841/jobs/6582772116

But it creates mypy failures:
```
sphinx/pycode/parser.py: note: In member "visit_Expr" of class "VariableCommentPicker":
sphinx/pycode/parser.py:409:84: error: "str" has no attribute "decode"; maybe "encode"?  [attr-defined]
sphinx/util/inspect.py:702: error: Unused "type: ignore" comment
sphinx/domains/std.py: note: In function "warn_missing_reference":
sphinx/domains/std.py:1116:58: error: "Domain" has no attribute "anonlabels"  [attr-defined]
sphinx/builders/changes.py: note: In member "write" of class "ChangesBuilder":
sphinx/builders/changes.py:58:24: error: Value of type "Optional[str]" is not indexable  [index]
sphinx/setup_command.py:149: error: Unused "type: ignore" comment
Found 5 errors in 5 files (checked 175 source files)
```

and also a whole bunch of ``E501 line too long`` flake8 errors (see https://github.com/jfbu/sphinx/actions/runs/3862012540/jobs/6583332177#step:5:23), which should be reported when I now create this PR which will activate flake8.

Thus creating this PR only as a testimonial.

### DIFF
--- a/sphinx/application.py
+++ b/sphinx/application.py
@@ -362,10 +362,7 @@ class Sphinx:
                   else __('finished with problems'))
         if self._warncount:
             if self.warningiserror:
-                if self._warncount == 1:
-                    msg = __('build %s, %s warning (with warnings treated as errors).')
-                else:
-                    msg = __('build %s, %s warnings (with warnings treated as errors).')
+                msg = __("build %s, %s warning (with warnings treated as errors).") if self._warncount == 1 else __("build %s, %s warnings (with warnings treated as errors).")
             else:
                 if self._warncount == 1:
                     msg = __('build %s, %s warning.')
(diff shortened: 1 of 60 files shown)


AA-Turner:
I applied 5008291b instead, allowing a more gradual approach in applying the lint checks.

A

jfbu:
> I applied [5008291](https://github.com/sphinx-doc/sphinx/commit/5008291b3d2ab92309f1dbc087b0c0b1c7f4e29f) instead, allowing a more gradual approach in applying the lint checks.

ok, cool

PR #11113, opened 2023-01-08
Consider duplicate names when using image post transforms

### LINKED ISSUE #11110: Figures go missing from latex pdf if their files have the same base name and they use a  post transform
### Describe the bug

I'm trying to build a pdf from a document that contains some svg images. The images have identical names, but reside in different directories. I use `sphinxcontrib.rsvgconverter` from [sphinxcontrib-svg2pdfconverter](https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter) to convert them from svg to pdf.

When multiple svg images with the same base name are passed to `rsvgconverter.convert`, the destination path for both images is identical, therefore all but the last image is discarded when it is overwritten by a subsequent transformation.

The result is that the final pdf will have the last converted svg repeated several times instead of each svg appearing when it is supposed to.

### How to Reproduce

[Here](https://github.com/aaron-cooper/sphinx-issue-demo) is a repo containing some code that reproduces the issue.

The file [foo_expected.pdf](https://github.com/aaron-cooper/sphinx-issue-demo/blob/main/foo_expected.pdf) (github doesn't show the full content of the pdf, remember to click "more pages" I guess...) shows what I would expect the output to be, and [foo_actual.pdf](https://github.com/aaron-cooper/sphinx-issue-demo/blob/main/foo_actual.pdf) shows what it actually is. Notice that the two identically named png files are shown correctly in foo_actual.pdf, but for the two svg's a red circle is shown twice, instead of a black circle and then a red circle.

### Environment Information

```text
Please paste all output below into the bug report template



Platform:              linux; (Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.35)
Python version:        3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0])
Sphinx version:        6.1.2
Docutils version:      0.19
Jinja2 version:        3.0.3
Pygments version:      2.13.0
```

### Sphinx extensions

_No response_

### Additional context

_No response_

aaron-cooper:
Closes: #11110

### DIFF
--- a/sphinx/transforms/post_transforms/images.py
+++ b/sphinx/transforms/post_transforms/images.py
@@ -236,7 +236,8 @@ class ImageConverter(BaseImageConverter):
         else:
             srcpath = node['candidates']['*']
 
-        filename = get_filename_for(srcpath, _to)
+        filename = self.env.images[srcpath][1]
+        filename = get_filename_for(filename, _to)
         ensuredir(self.imagedir)
         destpath = os.path.join(self.imagedir, filename)
 


jfbu (approved):
Thanks for catching this longstanding bug, any reason for assigning to filename first and not do it only one step?

LGTM, but I am only the LaTeX guy here so I hope for confirmation by other maintainers...

aaron-cooper:
> any reason for assigning to filename first and not do it only one step?

Just prefer to do 1-ish thing per line I guess? I didn't do it with a huge amount of purpose or anything...

AA-Turner (approved):

aaron-cooper:
👀

jfbu:
Thanks for bug report and fix.

PR #11129, opened 2023-01-14
Use the "toctree-wrapper" class for the sphinx13 theme

mgeier:
Subject: Use the same styling for the separate TOC page as for all other TOCs.

### Feature or Bugfix
- Bugfix

### Detail

It looks like currently there is no styling, so it doesn't make any difference.

But if somebody decides to style the TOCs, this will also apply to the separate TOC page.


jfbu:
merging latest master update into this

AA-Turner:
Thanks @mgeier!

A

PR #11131, opened 2023-01-14
Don't use attribute type annotation for constructor parameters

hoodmane:
Fix a bug where if a class has an attribute and a constructor parameter with the same name, napoleon will use the type annotation from the attribute for the constructor argument.

### Feature or Bugfix
- Bugfix

### Detail
Suppose we have a class that looks like 
```py
class Example:
    """
    A Class

    Parameters
    ----------
    blah:
        Description of parameter blah
    """
    def __init__(self, blah: int):
        pass

    blah: bool
```
The processed docstring looks like:
```rst
A Class
       
:param blah: Description of parameter blah
:type blah: bool
```
Note that the parameter has type `int` but Napoleon took the type from the attribute which has type `bool`. 
A more practical reason this would happen is that the parameter has type `Optional[dict]` and this is stored into an attribute of type `dict` where if `None` is passed a default dictionary is created.

I added a new parameter `is_attribute` to `_lookup_annotation` and `_consume_field`(s) which defaults to `False`. If `is_attribute` is `False` and `what` is `"class"` then we look up the type from the constructor arguments instead. With these changes it produces the correct output:
```rst
:param blah: Description of parameter blah
:type blah: int
```

### DIFF
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -278,8 +278,8 @@ class GoogleDocstring:
             line = self._lines.get(0)
         return lines
 
-    def _consume_field(self, parse_type: bool = True, prefer_type: bool = False
-                       ) -> tuple[str, str, list[str]]:
+    def _consume_field(self, parse_type: bool = True, prefer_type: bool = False,
+                       is_attribute: bool = False, ) -> tuple[str, str, list[str]]:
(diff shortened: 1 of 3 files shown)


hoodmane:
CI errors look unrelated to my changes.

AA-Turner:
See #11143

A

PR #11139, opened 2023-01-17
TST: change PYTHONWARNINGS from "all" to "error"

mgeier:
This is just an experiment, I don't know if that actually works.

At https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS, the value `all` is not mentioned.

Setting it to `error` normally makes sense in a CI setting, but there might be too many warnings, so we might have to add a few exceptions.


jfbu:
merging latest master update into this branch

PR #11144, opened 2023-01-20
Doc of html_file_suffix (closes #11141)

### LINKED ISSUE #11141: html_file_suffix defaults to None not ".html"?
### Describe the bug

Unless I'm misunderstanding [this line](https://github.com/sphinx-doc/sphinx/blob/e17f39e7b535f4849754d46f2cd2e411df8aa535/sphinx/builders/html/__init__.py#L1362), it looks like [html_file_suffix](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_file_suffix) defaults to `None` and not `.html` like the docs say, but instead looks like it picks up `.html` [from here](https://github.com/sphinx-doc/sphinx/blob/e17f39e7b535f4849754d46f2cd2e411df8aa535/sphinx/builders/html/__init__.py#L198). It seems like the default should be fixed, or if that isn't possible then update the docs.

### How to Reproduce

I'm not sure the best way to reproduce, but it came up in this PR: https://github.com/jdillard/sphinx-sitemap/pull/57

### Environment Information

```text
Platform:              linux; (Linux-5.4.0-135-generic-x86_64-with-glibc2.31)
Python version:        3.9.16 (main, Dec 25 2022, 18:26:15) 
[GCC 9.4.0])
Sphinx version:        6.1.3
Docutils version:      0.19
Jinja2 version:        3.1.2
Pygments version:      2.14.0
```

### Sphinx extensions

_No response_

### Additional context

_No response_

jfbu:
closes #11141

### DIFF
--- a/doc/usage/configuration.rst
+++ b/doc/usage/configuration.rst
@@ -1369,8 +1369,8 @@ that use Sphinx's HTMLWriter class.
 
 .. confval:: html_file_suffix
 
-   This is the file name suffix for generated HTML files.  The default is
-   ``".html"``.
+   This is the file name suffix for generated HTML files, if set to a :obj:`str`
+   value.  If left to the default ``None``, the suffix will be ``".html"``.
 
    .. versionadded:: 0.4
 


jfbu:
@tk0miya Good to see you! CI acquired a problem with ``tests/test_config.py::test_needs_sphinx`` which I could reproduce at my locale after force reinstall of tox (and any Sphinx version) but I have lost the older list of packages versions installed by tox, so I have difficulties identifying which version bump of dependencies trigger the problem (and I am only LaTeX guy here).  Executing the test via ``make test`` passes (with some EncodingWarning from pytest).

jfbu:
@tk0miya I am really silly it is simply the bump of Alabaster to ``0.7.13`` which triggers the problem. In my environment for `make test` I still had ``0.7.12``.  With ``0.7.13`` it fails too.

Issue #10887, opened 2022-09-30
section heading in function doc string triggers error: Unexpected section heading


Phillip-M-Feldman:
### Describe the bug

The presence of a section heading in a doc string triggers the error "unexpected doc string".  Is this something that is supposed to work?  When a function's doc string is very long, the ability to divide it into sections is essential.

### How to Reproduce

Put a section heading in the doc string of a function or method.

### Your project

N/A

### OS

Windows 10

### Python version

3.8.8

### Sphinx version

5.2.2

AA-Turner:
Hi Phillip, please provide a minimal reproducer with Sphinx 5.2.3 (ideally a single .rst file and a minimal conf.py)

A

Phillip-M-Feldman:
A minimal reproducer is attached.
[Sphinx_bug_demo.zip](https://github.com/sphinx-doc/sphinx/files/9700884/Sphinx_bug_demo.zip)

AA-Turner:
Minimal reproducer:

```python
import shutil
from pathlib import Path

from sphinx.cmd.make_mode import run_make_mode

Path("conf.py").write_text('''\
''', encoding="utf-8")

Path("index.rst").write_text('''\
.. py:function:: func()

   Overview
   --------
   
   Lorem ipsum dolar sit amet
''', encoding="utf-8")

shutil.rmtree("_build", ignore_errors=True)
run_make_mode(["html", ".", "_build", "-anETW"])
```

A

Issue #11116, opened 2023-01-09
Linux documentation build memory hog


marxin:
### Describe the bug

As mentioned here:
https://lore.kernel.org/all/cb916eae-fdf6-504f-8f38-7928e0fa8344@gmail.com/

the Linux kernel documentation builds https://docs.kernel.org/doc-guide/sphinx.html?highlight=sphinx#sphinx-build gets slower and consumer much more memory.

I bisected that to 080517ce55119a6358337b5f939b3942e4de63b4.
@AA-Turner: Can you please take a look?

### How to Reproduce

Clone Linux repo and apply:
https://lore.kernel.org/all/87wn629ggg.fsf@meer.lwn.net/

`make htmldocs`

### Environment Information

```text
Aforementioned Sphinx version.
```

AA-Turner:
@AA-Turner: Can you please take a look?

I'm taking a break due to other work until Easter; apologies for this. I'd advise pinning to an earlier version if a root cause can't be found.

A

marxin:
Do you think it would make sense to bisect `docutils` in order to find what caused that?

marxin:
So the issue is caused by the following removed hunk:
https://github.com/sphinx-doc/sphinx/blob/4660b62de0e83f42b6e2ede7a1a19e861e4d34e4/sphinx/util/nodes.py#L630-L643

which now relies on `docutils` `copy function which also copies `_document` property:
https://github.com/docutils/docutils/blob/bc9bf5fc2c7f905507a49e8e37c12368c2f4a0cc/docutils/docutils/nodes.py#L1056-L1061

If I add copying of `_document` to the `_new_copy` function, I get the very same memory hog.

AA-Turner:
I have released v6.1.3 which should fix this.

A

jfbu:
I will close the issue for time being but we can always reopen at later date.

The fix happened at dcb4429abacda015f98875fe9489ced696155bd7 and was merged into 6.1.3.

The original removal of the monkeypatch of DocUtils happened at 080517ce55119a6358337b5f939b3942e4de63b4 from #10470 which was part of the `6.0.0` release.

PR #11147, opened 2023-01-22
Add ``content_offset`` parameter to ``nested_parse_with_titles``

jbms:
### Feature or Bugfix

- Bugfix

### Purpose
Previously, `nested_parse_with_titles` always passed `0` as the input offset when invoking `nested_parse`.  When parsing the content of a directive, as is a common use case for `nested_parse_with_titles`, this leads to incorrect source file/line number information, as it does not take into account the directive's `content_offset`, which is always non-zero.

This issue affects *all* object descriptions due to #10887.  It also affects the `ifconfig` extension.

The `py:module` and `js:module` directives employed a workaround for this issue, by wrapping the calls to `nested_parse_with_title` with `switch_source_input`.  That worked, but was more complicated (and likely less efficient) than necessary.

This commit adds an optional `content_offset` parameter to `nested_parse_with_titles`, and fixes callers to pass the appropriate content offset when needed.

This commit eliminates the now-unnecessary calls to `switch_source_input` and instead specifies the correct `content_offset`.


jbms:
Note: An alternative fix would be to modify `nested_parse_with_titles` to always invoke `switch_source_input`.  That would avoid the need for the `content_offset` parameter and any changes to callers of this function.  But it would introduce at least some unnecessary added overhead from constructing a new StateMachine object.

jfbu:
merging latest master update into this, for CI checks

jbms:
After the merge, CI seems to pass.

jbms:
@jfbu It would be great if this fix can be merged soon.  What do you think?

jfbu:
> @jfbu It would be great if this fix can be merged soon. What do you think?

@jbms it looks good to me but I am the LaTeX contractor here so pinging @AA-Turner

jbms:
> > @jfbu It would be great if this fix can be merged soon. What do you think?
> 
> @jbms it looks good to me but I am the LaTeX contractor here so pinging @AA-Turner

Oh, I see.  Unfortunately he said he was out until Easter: https://github.com/sphinx-doc/sphinx/issues/11116#issuecomment-1375688482

PR #11150, opened 2023-01-23
Include source in highlighting warnings

jbms:
### Feature or Bugfix
- Feature

### Purpose
If pygments fails to lex a source string as the specified highlight language, Sphinx prints a warning.  Previously, that warning did not include the actual source text, although it does include location information.

However, in some cases the location information may be missing, there may be multiple highlighted literals on the same line, or the rST is automatically generated somehow.  In such cases, it can be difficult to determine the source text that led to the error.

With this change, the source text is included in the warning.


jbms:
> Isn't this problematic in connection with the `literalinclude` directive? If using it, the source code might be bulky...

I suppose I could try to add some logic to exclude that case.  Even if the warning is very long that doesn't seem like necessarily a problem, though.

Issue #9173, opened 2021-05-05
Docs of LaTeX templating


Lennonka:
**Is your feature request related to a problem? Please describe.**

I'm always frustrated when I want to change composition of latex output. I've noticed in the docs that LaTeX templating is possible and available as an experimental feature, but the hint is not sufficient to start with it.

**Describe the solution you'd like**

I'd love info about LaTeX templating syntax in the docs, or at least an example of what is supposed to be inside the files `_templates/latex.tex_t` etc.

**Describe alternatives you've considered**

Studying the source code, but I don't have really time for this, so some docs would be greatly appreciated! :)

**Additional context**

- https://www.sphinx-doc.org/en/master/latex.html#miscellany

grubert:
Did you have a look at the docutils documentation https://docutils.sourceforge.io/docs/user/latex.html ?
this most likely is version depending

PR #11165, opened 2023-01-31
Support and prefer ``.jinja`` to ``_t`` for static templates

jayaddison:
### Feature or Bugfix
- Refactoring

### Purpose
This is a pedantic change that relates to nedbat/coveragepy#1488, where `coveragepy` emits some warnings because it's unclear based on the file suffix whether some template files that contain Python code (and are processed by Jinja) indeed are Python code.

The suggested changes here probably only make sense if the corresponding templates are only expected to be used internally by Sphinx itself.  If the template filenames are part of any kind of implicit interface for external libraries/applications, then this change might be more disruptive than the small benefit that it could provide.

### Detail
- Where templates named with a `_t` suffix exist that contain Jinja templates, replace the suffix with `.jinja2`
- `.j2` was considered as an alternative suffix - `.jinja2` seems less ambiguous
  - 2023-02-19: Also, the Jinja2 project has been renamed to Jinja

### Relates
- nedbat/coveragepy#1488


mgeier:
This would break all extensions that use the `_t` convention and `copy_asset_file()` (as described in the docs: https://www.sphinx-doc.org/en/master/development/theming.html#add-your-own-static-files-to-the-build-assets).

So please don't do that.

jayaddison:
Thanks @mgeier - I wasn't aware of that.

I think that results in two possible paths to follow: either I'll abandon these changes, or I'll include compatibility with the existing suffix for extensions that rely on that.

(I'm leaning towards the former at the moment, but will spend some more time thinking about it)

AA-Turner:
> I think that results in two possible paths to follow: either I'll abandon these changes, or I'll include compatibility with the existing suffix for extensions that rely on that.

I support the change but we'd have to include a deprecation period of perhaps a year or longer.

A

sphinx/util/fileutil.py
```diff
@@ -4,17 +4,33 @@
...
 
+from sphinx.deprecation import RemovedInSphinx80Warning
 from sphinx.util.osutil import copyfile, ensuredir
 from sphinx.util.typing import PathMatcher
 
 if TYPE_CHECKING:
     from sphinx.util.template import BaseRenderer
 
 
+def _template_basename(filename: str) -> Optional[str]:
```
jayaddison: `Optional` is used here to support Python 3.8.  Ruff wants to suggest `str | None`, but doesn't appear to limit that suggestion based on [configured Python version](https://github.com/sphinx-doc/sphinx/blob/ecfd08d3254587972299c23cd596660f5796099c/pyproject.toml#L140) ([union type syntax is available from Python 3.10](https://docs.python.org/3.10/whatsnew/3.10.html#pep-604-new-type-union-operator)) - I'll report that to `ruff`.

jayaddison: Nope, my mistake; that suggestion is [valid when `from __future__ import annotations` is in use](https://github.com/charliermarsh/ruff/issues/2382#issuecomment-1409939445).  Today I learned.

jayaddison:
Although this is ready-for-review, I'm also trying to gauge my level of commitment in terms of supporting and migrating existing extensions.

Generally I think the best approach would be to try to identify as many extensions as possible that are reliant on the existing template suffix format and to offer pull requests with relevant migrations for them (hopefully in most cases those would be straightforward).  I might need to spend a bit of time to recharge before starting that analysis, though.

jayaddison:
I've run some (admittedly legacy) [GitHub code searches](https://github.com/search?p=2&q=filename%3A%2A.py+sphinx+copy_asset_file+-%22the+Sphinx+team%22&type=Code) to find potentially-affected extensions.  Very few use the `_t` suffix on templates, and they do not appear to be widely-used extensions.  A few other extensions _could_ be configured or read from input files (or files from directories) that contain the template suffix.

In cases that are `popular OR definitely affected` (regardless of maintenance status) I'll offer pull requests to adjust the behaviour.

So far I don't think that there are many valid use cases for dynamic content rendering during asset file copying, but there may be valid use cases I'm yet to learn about.

jayaddison:
> So far I don't think that there are many valid use cases for dynamic content rendering during asset file copying, but there may be valid use cases I'm yet to learn about.

Following on from that: I'm still unaware of many good use cases for this copy-and-apply-templating in the Sphinx `copy_asset_file` method.  There could be some out there (and I wonder about non-free/public code in particular), but I haven't yet figured out what they could be.  Even if valid use cases do exist, I wonder whether custom code to apply a Jinja template and then use Sphinx code to perform the copy could be clearer and more explicit about the intent of the code, without any loss of functionality.

That's making me wonder whether instead of suggesting a `.jinja` template suffix as an alternative, it could make more sense to remove the templating functionality from `copy_asset_file` and only allow it to perform static file copying (probably a major-api-version change required if so, though?).

mgeier:
> I've run some (admittedly legacy) [GitHub code searches](https://github.com/search?p=2&q=filename%3A%2A.py+sphinx+copy_asset_file+-%22the+Sphinx+team%22&type=Code) to find potentially-affected extensions. Very few use the `_t` suffix on templates, and they do not appear to be widely-used extensions.

The suggested change here will not only affect extensions that use `copy_asset_file()`!

It will even not only affect extensions, but also themes and plain old Sphinx users who want to overwrite an existing file in `_static` (or `_templates`) with a file that uses Jinja markup.

Here are examples for the (at least) three different cases:

* Extension using `copy_asset()` to copy files from https://github.com/spatialaudio/nbsphinx/tree/master/src/nbsphinx/_static
* Theme just providing a few static files (no Python code involved at all!): https://github.com/mgeier/insipid-sphinx-theme/tree/master/src/insipid_sphinx_theme/insipid/static
* Normal Sphinx documentation, just overwriting a template that's normally provided by Sphinx (no Python code used except for defining `templates_path` in `conf.py`): https://github.com/spatialaudio/nbsphinx/pull/719/files#diff-fcf9d762319bf115a8c6cdad49abf75fd39d76d4d1a2d44ed7f114653968a6d4

Those are just examples where I myself have used the `_t` suffix. I'm sure there are many more examples on Github (but maybe hard to search for) and even more in non-public repositories.

I was wondering though: since it seems to be reasonably simple to support both `_t` and `.jinja` for a deprecation period, why not support both indefinitely and never deprecate them?

jayaddison:
Thank you for looking into this further @mgeier.  I became sidetracked by one particular detail, as I often do, and forgot to look at the rest of the picture.

> I was wondering though: since it seems to be reasonably simple to support both _t and .jinja for a deprecation period, why not support both indefinitely and never deprecate them?

That seems like a good approach to me :+1:

sphinx/util/fileutil.py
```diff
@@ -4,17 +4,33 @@
...
+from sphinx.deprecation import RemovedInSphinx80Warning
 from sphinx.util.osutil import copyfile, ensuredir
 from sphinx.util.typing import PathMatcher
 
 if TYPE_CHECKING:
     from sphinx.util.template import BaseRenderer
 
 
+def _template_basename(filename: str) -> str | None:
+    # TODO: remove this method
```
AA-Turner: Remove when?

sphinx/util/fileutil.py
```diff
@@ -4,17 +4,33 @@
...
+    if filename.lower().endswith('_t'):
+        warnings.warn(
+            f"{filename!r}: filename suffix '_t' for templates is deprecated. "
+            "If the file is a Jinja2 template, use the suffix '.jinja' instead.",
+            RemovedInSphinx80Warning,
```
AA-Turner: Perhaps a generic ``PendingDeprecationWarning``, with a comment to re-visit not earlier than (e.g.) 1 Jan 2025?

sphinx/util/fileutil.py
```diff
@@ -4,17 +4,33 @@
...
+        )
+        return filename[:-2]
+    elif filename.lower().endswith(".jinja"):
+        return filename[:-6]
+    return None
```
AA-Turner, line 29:
```suggestion
    return filename
```

sphinx/util/fileutil.py
```diff
@@ -35,14 +51,13 @@ def copy_asset_file(source: str, destination: str,
         # Use source filename if destination points a directory
         destination = os.path.join(destination, os.path.basename(source))
 
-    if source.lower().endswith('_t') and context is not None:
+    if _template_basename(source) and context is not None:
```
AA-Turner, line 52: (given earlier suggestion to make ``_template_basename`` always return *something*)
```suggestion
    if _template_basename(source) != source and context is not None:
```

sphinx/util/fileutil.py
```diff
@@ -35,14 +51,13 @@ def copy_asset_file(source: str, destination: str,
...
         if renderer is None:
             from sphinx.util.template import SphinxRenderer
             renderer = SphinxRenderer()
 
         with open(source, encoding='utf-8') as fsrc:
-            if destination.lower().endswith('_t'):
-                destination = destination[:-2]
+            destination = _template_basename(destination) or destination
```
AA-Turner, line 58: (given earlier suggestion to make ``_template_basename`` always return *something*)
```suggestion
            destination = _template_basename(destination)
```

tests/test_util_fileutil.py
```diff
@@ -101,3 +101,14 @@ def excluded(path):
...
+
+
+def test_template_basename():
+    assert not _template_basename("asset.txt")
+    assert _template_basename("asset.txt.jinja") == "asset.txt"
+    assert _template_basename("sidebar.html.jinja") == "sidebar.html"
+
+
+def test_legacy_template_basename():
+    # TODO: remove with sphinx8
```
AA-Turner: See earlier comment about 1 Jan 2025 or later

AA-Turner:
Looks good overall, thank you @jayaddison!

Two additional requests:

1. Please add an entry to ``CHANGES`` and to ``doc/extdev/deprecated.rst``
2. (if you have time) Would it be possible to test this PR with themes that use the `_t` feature, or override Sphinx-provided files, or etc? 


Thanks,
Adam

jayaddison:
Thanks @AA-Turner - I'll apply a few of those feedback suggestions soon.

Also: yep, I should have time to test this with an existing theme.  I'll try to do that before pushing the code review feedback commits.

jayaddison:
Ok: the [`sphinx-carousel` theme](https://sphinx-carousel.readthedocs.io/en/1.2.0/) uses the `_t` suffix for some custom CSS content.  Mostly this is used to fix CSS styling that conflicts with other Sphinx extensions/themes, but it can also apply CSS styling to Sphinx content in the absence of other extensions.

Such a sample is provided below, confirming that the custom CSS rules are applied: the `color: revert` CSS styling is sourced from the (originally-`_t`-suffixed, templated) `carousel-custom.css` file.

image

This sample app was built using the local development branch of Sphinx that I have locally (`pip install <path-to-sphinx-working-directory>`).

sphinx/util/fileutil.py
jayaddison: Updated with a maybe-arbitrary suggested removal timeline of 2 years from today.

jayaddison:
> Please add an entry to CHANGES and to doc/extdev/deprecated.rst

Please note: I've updated the `CHANGES`, but not yet the `deprecated.rst` file.  Although the behaviour of the method may break for some after removal-time, the method itself does continue to exist in the API interface, and that file appears to document removals from the interface.

(I'd be OK with documenting it there anyway as an exception - maybe this is a semantic argument - but figured it'd be worth checking that first)

CHANGES
```diff
@@ -22,7 +22,14 @@ Incompatible changes
 Deprecated
 ----------
 
+* LaTeX: some internals of the ``sphinxadmonition`` environment have been
+  marked for removal at 7.0.0 in the :file:`sphinxlatexadmonitions.sty`.
+  Custom re-definitions of ``sphinxlightbox`` or ``sphinxheavybox`` will have
+  to be updated if they depended upon them.
```
jayaddison: @jfbu I accidentally didn't resolve this merge conflict properly - was this text intended to be removed as part of a7949d42903efa64bc5f8f870d39d811e272ecda?

jfbu: @jayaddison Yes it should be removed, sorry about this but I indeed reverted my mind during this period when I was active on the LaTeX things. So I added it and later removed it in https://github.com/sphinx-doc/sphinx/commit/a7949d42903efa64bc5f8f870d39d811e272ecda as you found out.

edit: now that I think about it I remember that hunk removal ended up in wrong commit during a rebase at my locale, but it is definitely intended that this paragraph in CHANGES should be removed indeed.

jayaddison: No problem, thanks for checking :+1:

jfbu (commented):
Thanks for all this.  Could you please also update the "hint" near very bottom of [doc/latex.rst](https://github.com/sphinx-doc/sphinx/blob/master/doc/latex.rst) ?

CHANGES
jfbu: Thanks for conscienciousness ;-) :+1:  or conscientiousness rather I am not sure but try to be conscientious !

jayaddison:
Thanks!  Yes, I hadn't found that.  I'll update that in a few moments.

jfbu (commented):
It  seems a ``_templates/latex.tex_t``  is now ignored by ``make latex``, without raising a deprecation warning (but a ``_templates/latex.tex.jinja`` is obeyed as expected).

To test I simply created an empty project, copied over ``latex.tex.jinja`` into the project ``_templates/`` directory as  file with basename ``latex.tex_t`` (old suffix), then added therein some `13pt` as `\documentclass` option and executed `make latex` with your branch checked out.  There was no `13pt` found in `_build/latex/<foo>.tex` but no warning either.

jfbu:
> It seems a `_templates/latex.tex_t` is now ignored by `make latex`, without raising a deprecation warning (but a `_templates/latex.tex.jinja` is obeyed as expected).
> 

This happens in writers.latex.astext() which makes no check.

jfbu:
> Thanks @jfbu - I think that's more of a breaking change than is intended at the moment. Roughly speaking, there should be a recommendation (deprecation warning) for people to migrate, but existing workflows should remain unchanged.

+1
> 
> This is also making me wonder how widespread the `make latex` usage could be: quite large, is my guess. Perhaps even two years isn't a sufficient migration duration for that.

Well I don't think the use of the LaTeX templates is that widespread, because it requires LaTeX know-how (I should check some projects first, but I don't expect I will find many, if any ...).  But the documentation does mention it (I don't think it mentions the `imgmath`  related templates, will check, in contrast to the `make latex` templates).  A period of 2 years looks to me as too prudent. I would be ok with making old templates suffixes work during next major release and raise warnings during that period, and be removed at the one after that i.e. at 8.0.  I am not sure about our major releases cycle duration, perhaps 8.0 is for 2024?
> 
> I'll make sure that the latex-related workflow is unaffected and also check why a deprecation warning didn't appear.

The ``render()`` method of  ``LaTeXTranslator`` might need attention.

jayaddison:
Ok, thank you.  I'm a bit slow to catch up on the details, but I think I'm getting there.

The approach I'm looking at currently is to update `LaTeXTranslator` to check for multiple template file paths -- with an emitted warning if a `_t` suffix template is found -- and I'm planning to add corresponding test coverage for that.

jayaddison:
Ok, there's one behaviour in particular that I think is useful to understand while migrating the LaTeX rendering code.

It's the fallback case that applies within this line range, and it occurs when no template is found matching the requested name (such as `latex.tex_t`):

https://github.com/sphinx-doc/sphinx/blob/9299003d40c775150c53366dcfdd4d944d0da1af/sphinx/writers/latex.py#L517-L523

The default behaviour allows the renderer to refer to Sphinx library's built-in templates instead of the locally-copied templates present in the project directory.  That's an important distinction because those templates are not copied by default.

This also explains/relates-to your comment @jfbu:

> To test I simply created an empty project, copied over latex.tex.jinja into the project _templates/ directory ...

(I discovered this while reverting some of the `.tex.jinja` suffix logic back to `.tex_t` temporarily to try to provide clear diffs and to understand the problem)

doc/latex.rst
```diff
@@ -1811,12 +1811,12 @@ Miscellany
...
-   LaTeX source if you have a file named ``_templates/latex.tex_t`` in your
+   LaTeX source if you have a file named ``_templates/latex.tex.jinja`` in your
    project.
 
-   Additional files ``longtable.tex_t``, ``tabulary.tex_t`` and
-   ``tabular.tex_t`` can be added to ``_templates/`` to configure some aspects
-   of table rendering (such as the caption position).
+   Additional files ``longtable.tex.jinja``, ``tabulary.tex.jinja`` and
+   ``tabular.tex.jinja`` can be added to ``_templates/`` to configure some
+   aspects of table rendering (such as the caption position).
```
jayaddison, line 1819: I think it's worth extending this documentation to mention the [non-default syntax](https://github.com/sphinx-doc/sphinx/blob/9299003d40c775150c53366dcfdd4d944d0da1af/sphinx/util/template.py#L77-L84) for the Jinja2 files in this context.

jayaddison: Relates-to: #9173

jayaddison:
Something isn't quite right yet; I tried to _remove_ the `_t` suffix support from the LaTeX writer, to check what'd happen if this is merged and then the time-to-removal elapses.

The `test_latex_table_custom_template_caseA` test case began failing unexpectedly.  I'm trying to figure out why that happened.

jayaddison:
> Something isn't quite right yet; I tried to _remove_ the `_t` suffix support from the LaTeX writer, to check what'd happen if this is merged and then the time-to-removal elapses.
> 
> The `test_latex_table_custom_template_caseA` test case began failing unexpectedly. I'm trying to figure out why that happened.

Ah, silly me:

It's because I modified:

```diff
-            for template_suffix in ('_t', '.jinja'):
+            for template_suffix in ('.jinja'):
```

... and that's wrong (`template_suffix` iterates over the characters within the string `.jinja`, since it is no longer contained in a `tuple`).

The following works as expected:

```diff
-            # TODO: remove "_t" template suffix support after 2025-04-06
-            for template_suffix in ('_t', '.jinja'):
-                template = path.join(self.builder.confdir, template_dir,
-                                     template_name + template_suffix)
-                if path.exists(template):
-                    if template_suffix == '_t':
-                        warnings.warn(
-                            f"{template!r}: filename suffix '_t' for templates is deprecated. "
-                            "If the file is a Jinja2 template, use the suffix '.jinja' "
-                            "instead.",
-                            PendingDeprecationWarning,
-                        )
-                    return template
+            template = path.join(self.builder.confdir, template_dir, f"{template_name}.jinja")
+            if path.exists(template):
+                return template
```

sphinx/ext/imgmath.py
```diff
@@ -95,9 +95,9 @@ def generate_latex_macro(image_format: str,
     }
 
     if config.imgmath_use_preview:
-        template_name = 'preview.tex_t'
+        template_name = 'preview.tex.jinja'
     else:
-        template_name = 'template.tex_t'
```
jayaddison: There is a backwards-compatibility issue here, I think - in the case where `imgmath_use_preview` is enabled and also users have chosen to make a local copy of the relevant `*.tex_t` template.

Perhaps a rare case, but given the potential long timespan for the change, I think it'd be worth fixing this.

jayaddison: Resolved, albeit with some code duplication.

sphinx/util/fileutil.py
```diff
@@ -35,14 +50,13 @@ def copy_asset_file(source: str, destination: str,
         # Use source filename if destination points a directory
         destination = os.path.join(destination, os.path.basename(source))
 
-    if source.lower().endswith('_t') and context is not None:
+    if _asset_destination(source) != source and context is not None:
```
jayaddison: A self-nitpick: `destination` is an overloaded variable/method name here.  It should be possible (and may be safer) to improve the clarity of what is being referred to.

sphinx/util/fileutil.py
jayaddison: This seemed to make sense to me earlier, but after looking again, there seem to be some subtleties around directories-vs-filenames as destination that the previous implementation was handling.

The fact that that wasn't obvious is not ideal, so a better fix would make the reasoning more transparent (either via comments, or preferably by restructuring the logic).

sphinx/writers/latex.py
```diff
@@ -516,7 +516,7 @@ def generate(content: list[tuple[str, list[IndexEntry]]], collapsed: bool) -> No
     def _find_template(self, template_name: str) -> str:
         for template_dir in self.config.templates_path:
             # TODO: remove "_t" template suffix support after 2025-04-06
-            for template_suffix in ('_t', '.jinja'):
+            for template_suffix in ['_t', '.jinja']:
```
AA-Turner:
```suggestion
            for template_suffix in ('_t', '.jinja'):
```

Would a tuple work here for a (tiny) memory saving?

sphinx/ext/imgmath.py
```diff
@@ -95,16 +96,25 @@ def generate_latex_macro(image_format: str,
...
+        template_name = 'template.tex'
 
     for template_dir in config.templates_path:
-        template = path.join(confdir, template_dir, template_name)
-        if path.exists(template):
-            return LaTeXRenderer().render(template, variables)
-
-    return LaTeXRenderer(templates_path).render(template_name, variables)
+        # TODO: remove "_t" template suffix support after 2025-04-06
+        for template_suffix in ['_t', '.jinja']:
```
AA-Turner:
```suggestion
        for template_suffix in ('_t', '.jinja'):
```
Same tuple comment

AA-Turner:
@jayaddison is there anything else outstanding here?

A

jayaddison:
Thanks @AA-Turner - no further changes planned, but I'd like to read it through again myself one more time before merge.

In particular, learning more about the LaTeX templates was something of a surprise, so I'd like to re-evaluate and make sure I'm comfortable with it before it's accepted (because I'd like to be around when it's removed, too).

sphinx/writers/latex.py
jayaddison: Yep, that works and does seem reasonable :+1: 

(I chose the `list` type after making a [mistake](https://github.com/sphinx-doc/sphinx/pull/11165#issuecomment-1499322285) while practicing removal of `_t` support; test coverage does catch that though, even if someone else makes the same mistake later)

jayaddison:
A pedantic observation:

- The `copy_asset` methods check for filenames ending in `_t` / `.jinja` in a _case-insensitive_ way
- The code to apply templating during output generation (for example, LaTeX templating) check for existence of templates using _case-sensitive_ inputs

I don't think that's likely to cause huge problems; on case-sensitive filesystems I doubt that people would prefer to use `_T` or `.JINJA` as file suffixes.  Reality being what it is, there are probably some rare cases where that would occur, though.

sphinx/util/fileutil.py
jayaddison: I think that adding a docstring to the `_template_basename` method would be a decent approach to resolve this (many IDEs would display that when developers visit the callsites).

jayaddison: (`_template_basename` method docstring added)

jayaddison:
> I don't think that's likely to cause huge problems; on case-sensitive filesystems I doubt that people would prefer to use `_T` or `.JINJA` as file suffixes. Reality being what it is, there are probably some rare cases where that would occur, though.

I don't think it's worth the additional filesystem existence checks to look for these.  If someone does encounter a problem due to uppercase file extensions on a case-sensitive system, then hopefully they can resolve that locally and harmlessly by renaming the files.  If not, then with an issue report we can consider adjusting the code appropriately.

jayaddison:
Ok, thanks for bearing with me while I hold tedious public conversations with myself @AA-Turner.  I think I'm now comfortable that this changeset is ready to merge :+1:

doc/latex.rst
jayaddison: (I'll try to find time to offer some updated documentation related to this in a separate pull request)

CHANGES
```diff
@@ -23,6 +23,9 @@ Deprecated
 ----------
 
 * #11247: Deprecate the legacy ``intersphinx_mapping`` format
+* #11165: The ``_t`` suffix for templating (``sphinx.util.copy_asset()``,
+  ``sphinx.util.copy_asset_file()`` and built-in templates) is now deprecated.
```
mgeier: I would prefer if they would not yet be called "deprecated".

I think consensus was to mark them as "pending deprecation"?

sphinx/ext/imgmath.py
```diff
@@ -95,16 +96,25 @@ def generate_latex_macro(image_format: str,
...
-        template_name = 'template.tex_t'
+        template_name = 'template.tex'
 
     for template_dir in config.templates_path:
-        template = path.join(confdir, template_dir, template_name)
-        if path.exists(template):
-            return LaTeXRenderer().render(template, variables)
-
-    return LaTeXRenderer(templates_path).render(template_name, variables)
+        # TODO: remove "_t" template suffix support after 2025-04-06
```
mgeier: I think support should not be *removed* at that time, but the deprecation period should start at that time!

At the given date, a `DeprecationWarning` should be created, and after another year or two support for `_t` can be removed.

sphinx/util/fileutil.py
```diff
@@ -15,6 +16,23 @@
     from sphinx.util.template import BaseRenderer
 
 
+def _template_basename(filename: str) -> str | None:
+    """Given an input filename:
+    If the input looks like a template, then return the filename output should
+    be written to.  Otherwise, return no result (None)."""
+    # TODO: remove "_t" template suffix support after 2025-04-06
```
mgeier: See above, the deprecation period should *start* at that date.

sphinx/util/fileutil.py
```diff
@@ -15,6 +16,23 @@
...
+    if filename.lower().endswith('_t'):
+        warnings.warn(
+            f"{filename!r}: filename suffix '_t' for templates is deprecated. "
```
mgeier:
```suggestion
            f"{filename!r}: filename suffix '_t' for templates will become deprecated. "
```

sphinx/writers/latex.py
```diff
@@ -512,14 +513,27 @@ def generate(content: list[tuple[str, list[IndexEntry]]], collapsed: bool) -> No
...
 
-    def render(self, template_name: str, variables: dict[str, Any]) -> str:
-        renderer = LaTeXRenderer(latex_engine=self.config.latex_engine)
+    def _find_template(self, template_name: str) -> str:
         for template_dir in self.config.templates_path:
-            template = path.join(self.builder.confdir, template_dir,
-                                 template_name)
-            if path.exists(template):
-                return renderer.render(template, variables)
+            # TODO: remove "_t" template suffix support after 2025-04-06
```
mgeier: As above, this should be the start of the deprecation period.

tests/test_build_latex.py
```diff
@@ -1379,6 +1379,10 @@ def test_latex_table_custom_template_caseA(app, status, warning):
     result = (app.outdir / 'python.tex').read_text(encoding='utf8')
     assert 'SALUT LES COPAINS' in result
 
+    # # TODO: remove "_t" template suffix support after 2025-04-06
```
mgeier: same as above

tests/test_util_fileutil.py
```diff
@@ -101,3 +101,14 @@ def excluded(path):
...
+
+
+def test_template_basename():
+    assert not _template_basename("asset.txt")
+    assert _template_basename("asset.txt.jinja") == "asset.txt"
+    assert _template_basename("sidebar.html.jinja") == "sidebar.html"
+
+
+def test_legacy_template_basename():
+    # TODO: remove "_t" template suffix support after 2025-04-06
```
mgeier: same as above

mgeier:
Is it really necessary to ever remove support for `_t`?

It looks like minimal additional code, and the whole deprecation mechanism sure seems much more complicated than simply keeping both `_t` and `.jinja`, doesn't it?

doc/latex.rst
jfbu: > I think it's worth extending this documentation to mention the [non-default syntax](https://github.com/sphinx-doc/sphinx/blob/9299003d40c775150c53366dcfdd4d944d0da1af/sphinx/util/template.py#L77-L84) for the Jinja2 files in this context.

Thanks for all your hard work @jayaddison.  Perhaps we can document the [non-default syntax](https://github.com/sphinx-doc/sphinx/blob/9299003d40c775150c53366dcfdd4d944d0da1af/sphinx/util/template.py#L77-L84) as you rightly insisted upon  as a comment inside the template files themselves?  People intending to replace them by custom templates will probably start at looking how the originals are done.  Perhaps
```diff
diff --git a/sphinx/templates/latex/latex.tex_t b/sphinx/templates/latex/latex.tex_t
index deb030504..aae912a09 100644
--- a/sphinx/templates/latex/latex.tex_t
+++ b/sphinx/templates/latex/latex.tex_t
@@ -1,3 +1,4 @@
+<# Use JSP/eRuby like tagging as curly braces are frequent in LaTeX -#>
 %% Generated by Sphinx.
 \def\sphinxdocclass{<%= docclass %>}
 <% if latex_engine == 'lualatex' -%>
```
or a better one in this and the other latex templates.

CHANGES
jayaddison: Ok, yep: I'll update the `CHANGES` entry (and this pull request title) to say "pending deprecation" instead.

Sorta-related: reading back through some of the previous commentary, it looks like I misread one of your previous messages:

>> I was wondering though: since it seems to be reasonably simple to support both _t and .jinja for a deprecation period, why not support both indefinitely and never deprecate them?
>
> That seems like a good approach to me :+1: 

Somehow I didn't read/understand the `never deprecate them` part of that.

To me, the goal here is to make the purpose and content of the template files less ambiguous.  Currently `_t` indicates that they're templates, but doesn't say much about what kind of templating that is.  `.jinja` indicates both template, and a templating language that's in use.

(so, correcting myself: supporting both sounds good, but I would like to suggest deprecating `_t` at some point)

jayaddison: If we supported both suffixes `_t` and `.jinja` equally: would there be any situations where we'd recommend using `_t` instead of `.jinja`?

AA-Turner:
I've centralised the 'deprecation' calls to ``sphinx.deprecation``, and added documentation on how themes can silence the warning, which is important while Sphinx 6.1 and earlier are still supported by said themes (as you'd need to maintain two identical templates, or do some dynamic renaming, or etc -- which static templates are designed to avoid).

I've also updated the documentation in (what will become) https://www.sphinx-doc.org/en/master/development/theming.html.

A

AA-Turner:
Thanks @jayaddison!

A

jayaddison:
Thanks again @AA-Turner - I'll stay around to watch for issues related to this (and should probably re-read the deprecation policy a few times during the process).

jayaddison:
I've begun thinking about an edge-case today that I haven't yet found in practice, but it does still worry me:

What if some Sphinx sites/extensions contain static files that intentionally already use a `.jinja` suffix that _aren't_ intended to be evaluated during file copying?

With this change, those files would be evaluated and the results written to a file without the extension, possibly a breaking change in those cases.

jayaddison:
Even though I'd be slightly annoyed with myself for reverting this change (for clearly not planning it thoroughly), perhaps this should _not_ be included in 6.2.0 - with my [previous comment](https://github.com/sphinx-doc/sphinx/pull/11165#issuecomment-1509862003) in mind, I think a better way to phase it in would be:

1. In release `A`, add a warning during copying of any files that have a `.jinja` suffix 
1. In release `B` some time later, begin supporting `.jinja` as a template suffix
1. In release `C`, some time later again, drop support for the legacy `_t` template

Step one would allow time for people who do use `.jinja` static files for any reason to think about a migration path, and ideally to let us know what the reasons for that are.

Reasoning:

Perhaps I'm overthinking it, but since this is something I'm contributing in personal time (I do make contributions that I consider work-related here too, but this isn't one of them), and would continue to complete this in personal time, I'd prefer to choose a path that's likely to be as minimally-stressful for users (and therefore myself) as possible, even if that takes longer.

If it is included in 6.2.0 anyway though, I'll make time available to support these changes as best possible.

jfbu:
> Even though I'd be slightly annoyed with myself for reverting this change (for clearly not planning it thoroughly), perhaps this should _not_ be included in 6.2.0

@jayaddison Thanks for all your work on this.  I understand your concern and maybe indeed we can delay it so I will add a LGTM to your reverting PR, and wait for @AA-Turner opinion. My presence here will be very sparse for about  4 weeks.

jayaddison:
Thank you, @jfbu.

PR #11169, opened 2023-02-01
Make --parallel a synonym for -j

ferdnyc:
Subject: Add `--parallel` to the sphinx command-line options

### Feature or Bugfix

- Feature

### Purpose
- Add the (relatively self-documenting) command-line option `--parallel`, as a synonym for the (comparatively-inscrutable) `-j` flag.

### Detail
- It was pointed out that `-j auto` (or any other use of `-j`) isn't particularly self-documenting, as command-line arguments go.
- Also appears in the `--help` summary generated by `argparse`:
  ```console
  $ PYTHONPATH=`pwd` python3 -m sphinx --help
  usage: __main__.py [OPTIONS] SOURCEDIR OUTPUTDIR [FILENAMES...]

  [...]
  
  general options:
    -b BUILDER          builder to use (default: html)
    -a                  write all files (default: only write new and changed
                        files)
    -E                  don't use a saved environment, always read all files
    -d PATH             path for the cached environment and doctree files
                        (default: OUTPUTDIR/.doctrees)
    -j N, --parallel N  build in parallel with N processes where possible
                        (special value "auto" will set N to cpu-count)
  ```

### Relates
- https://github.com/python/devguide/pull/1038#discussion_r1088601323

### DIFF
--- a/sphinx/cmd/build.py
+++ b/sphinx/cmd/build.py
@@ -142,7 +142,8 @@ files can be built by specifying individual filenames.
     group.add_argument('-d', metavar='PATH', dest='doctreedir',
                        help=__('path for the cached environment and doctree '
                                'files (default: OUTPUTDIR/.doctrees)'))
-    group.add_argument('-j', metavar='N', default=1, type=jobs_argument, dest='jobs',
+    group.add_argument('-j', '--parallel', metavar='N', default=1,
+                       type=jobs_argument, dest='jobs',
                        help=__('build in parallel with N processes where '
                                'possible (special value "auto" will set N to cpu-count)'))
     group = parser.add_argument_group('build configuration options')


hugovk:
`-j` is usually a shortcut for `--jobs`. From `make --help`:
```
  -j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
```

ferdnyc:
@hugovk 

> `-j` is usually a shortcut for `--jobs`. From `make --help`:
> 
> ```
>   -j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
> ```

I guess my concern with that is, `--jobs` isn't particularly more self-documenting than `-j`. (The fact that said jobs run _in parallel_ is, at **best**, only _implied_.) The parallelism is the crucial aspect of the argument, and `--parallel` makes that explicit.

I'll see your `make` and raise you a `cmake --build`:
```console
$ cmake --build
Usage: cmake --build <dir>             [options] [-- [native-options]]

Options:
  --parallel [<jobs>], -j [<jobs>]
                 = Build in parallel using the given number of jobs. 
                   If <jobs> is omitted the native build tool's 
                   default number is used.
                   The CMAKE_BUILD_PARALLEL_LEVEL environment variable
                   specifies a default parallel level when this option
                   is not given.
```

ferdnyc:
Other tools on my system that use `--parallel`, based on a quick, and likely insufficient, `zgrep -i '\-parallel' /usr/share/man/man1/*` scan (which is a terrible metric, anyway):

* `curl` (though the short synonym is a confounding `-Z`)
* `iperf` (though the short synonym is a sensible `-P`)
* `django-admin test` (with no short synonym)

The list of tools that use `--jobs` is, I'll freely admit, much longer, and includes at least:
* `buildah-build`
* `bundle-{install,update}`
* `cargo-*`
* `dpkg-buildpkg`
* `git-{clone,fetch,pull,submodule}`
* `parallel` (hah! :rofl:)
* `prove`
* `trial`

CAM-Gerlach:
FWIW, at least for me as a user that rarely uses this argument in other tools (thus without the benefit of the familiarity argument for `-j`), just based off the names I find ``--jobs`` as easy to understand as ``--parallel`` in terms of its overall purpose, and actually substantially easy viz. it's usage, as to me I would tend to assume `--parallel` is a one-off flag while `--jobs`` takes an argument (as this does), with an obvious purpose.

And while we're at it, we could really use synonyms for the other args too, e.g. `--nitpicky`. In fact, @AA-Turner himself included synonyms for all of those included in the Python PEP repo's [build.py wrapper script](https://github.com/python/peps/blob/main/build.py), so it would be nice to upstream that idea.

ferdnyc:
@CAM-Gerlach 

> FWIW, at least for me as a user that rarely uses this argument in other tools (thus without the benefit of the familiarity argument for `-j`), just based off the names I find `--jobs` as easy to understand as `--parallel` in terms of its overall purpose, and actually substantially easy viz. it's usage, as to me I would tend to assume `--parallel` is a one-off flag while `--jobs`` takes an argument (as this does), with an obvious purpose.

Well, the original context this came up in was a lament/concern (expressed _by **you**_!! :wink: ) regarding self-documenting command lines.

So, it isn't so much about "understanding how to use the option" (that's what `--help` is for), it's more a question of "intuiting how this option, observed _in situ_, affects the command".

In other words, when you see something like (the original example from the other thread):

```sh
python -m sphinx -b html . build/html --jobs auto
```

...Does that `--jobs auto` do an appreciably better job than `-j auto`, in immediately and obviously indicating to the majority of users/readers, "this will be done in parallel, automatically scaling the number of processes/threads to the current CPU core count"?

Vs. this command line, same question:

```sh
python -m sphinx -b html . build/html --parallel auto
```

If the answer to the former is "yes", then `--jobs` is fine.

I'm not _personally_ convinced that, approached from that direction, `--jobs` is really expressive enough. (And I bet if I went back and looked, I'd find old discussion among the CMake devs, where they reached a similar conclusion before deciding to break with `make` regarding the long form of this flag.)

...But, I may simply be wrong.

ferdnyc:
> as to me I would tend to assume `--parallel` is a one-off flag while `--jobs`` takes an argument (as this does), with an obvious purpose.

CMake actually has you covered, on that point: _their_ `--parallel` (and `-j`) takes an **optional** integer (only) argument, to specify an explicit count. If it's used **without** an argument, it enables auto-scaling.

So... I guess it's either the best or worst of both worlds, depending on your perspective?

(`make`, OTOH, also makes the integer argument to `-j` / `--jobs` optional. But the documented effect is "infinite jobs with no arg.", which doesn't sound particularly useful.)

CAM-Gerlach:
> CMake actually has you covered, on that point: _their_ `--parallel` (and `-j`) takes an **optional** integer (only) argument, to specify an explicit count. If it's used **without** an argument, it enables auto-scaling.

`--parallel` is certainly nicer/clearer than `-j auto`, for sure. Only issue is consistency with the existing `-j` and other tools, particularly considering Sphinx is commonly invoked via `make`, and at least up until recently auto-created makefiles/Windows `make.bat` scripts in newly created projects by default (though I believe the intention is to move away from that and to `python -m sphinx` as the standard invocation).

sphinx/cmd/build.py
```diff
@@ -142,7 +142,8 @@ def get_parser() -> argparse.ArgumentParser:
     group.add_argument('-d', metavar='PATH', dest='doctreedir',
                        help=__('path for the cached environment and doctree '
                                'files (default: OUTPUTDIR/.doctrees)'))
-    group.add_argument('-j', metavar='N', default=1, type=jobs_argument, dest='jobs',
+    group.add_argument('-j', '--parallel', metavar='N', default=1,
```
cclauss, line 145: Why not `--jobs` given the `dest`, etc.?

CAM-Gerlach: (N.B., see also the extensive discussion above on that point)

hugovk:
I've opened https://github.com/sphinx-doc/sphinx/pull/11210 as an alternative using `--jobs`.

PR #11177, opened 2023-02-04
Cleanup: remove 'noqa' comment from a statement that is never visited

jayaddison:
### Feature or Bugfix
- Refactoring

### Purpose
- Removes a `noqa` (linter ignore statement) from a Python statement that is never visited at runtime

### Detail
- Versions of `ruff` prior to charliermarsh/ruff#2408 suggested the `noqa` statement, and more recent versions no longer suggest it (in fact, they suggest removing it)

### Relates
- https://github.com/charliermarsh/ruff/issues/2408

### DIFF
--- a/sphinx/util/display.py
+++ b/sphinx/util/display.py
@@ -8,7 +8,7 @@ from sphinx.util import logging
 from sphinx.util.console import bold, colorize, term_width_line  # type: ignore
 
 if False:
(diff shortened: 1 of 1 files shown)


jayaddison:
(already-resolved-by 4c4745af9cc788f94568bb3457f887b8c124f73f)

PR #11204, opened 2023-02-19
add support for line numbering in the viewcode extention

### LINKED ISSUE #10589: display the line numbers using sphinx.ext.viewcode?
I'm using the [sphinx.ext.viewcode](https://www.sphinx-doc.org/en/master/usage/extensions/viewcode.html) to document my API. Is it possible to display the line numbers on the produced code-block and if yes how ?

I've tried to find an `html_option` in the docs but I didn't find anything. As this question is on stale for a month on SO, I assumed it was not available in Sphinx so I'm moving it to a feature request.

12rambau:
add support for line numbering in the viewcode extention

### Feature or Bugfix
Feature

### Detail
- I added a `viewcode_show_lineos` variale (default to False) to control wether or not line number shoud be added to the viewcode pages. 
- This change the highlighter behavior in `collect_pages` method. 
- The highliter will create a table with all the line number in one coloum and all the code in a second one. As viewcode is customizeing the first and last line of each tag, I created an offset variable that search for the second column tag and simply add it to point to correct lines when modifying the content. 
- The new variable is documented and a short test check for the existence of the table in the output file.
- I decided not to change the display of the line numbers.
- for testing I checked using an extra build if the items are still modified in the table the same way.

I have 1 questions: 
- what's the use this line: https://github.com/sphinx-doc/sphinx/blob/93226ca91acdba7fdf0c7efde27ccc9c28cf7ece/sphinx/ext/viewcode.py#L256 It's 14 years old and when I check the output of line I only see 1 `<pre>` tag meaning that this is not changing anything. 

### Relates
Fix #10589


AA-Turner:
I believe everything in this PR was contained in the older #6319 -- if there are any additional tests or etc we should add, please re-open!

A

12rambau:
Ah sorry to have created a duplication. I checked before starting coding but didn't find out

PR #11210, opened 2023-02-25
Make ``--jobs`` a synonym for ``-j``

hugovk:
Subject: Add `--jobs` to the Sphinx command-line options

### Feature or Bugfix

- Feature

### Purpose
- Add the command-line option `--jobs` as a synonym for the `-j` flag.

### Detail
- Add the usual long option `--jobs`. Compare with `make --help`:

```
 -j [N], --jobs[=N]          Allow N jobs at once; infinite jobs with no arg.
```

## Before

```console
$ sphinx-build --help
...
  -j N              build in parallel with N processes where possible (special
                    value "auto" will set N to cpu-count)
...
```

## After

```console
$ sphinx-build --help
...
  -j N, --jobs N    build in parallel with N processes where possible (special
                    value "auto" will set N to cpu-count)
...
```

### Relates
- Alternative to and closes https://github.com/sphinx-doc/sphinx/pull/11169
- https://github.com/python/devguide/pull/1038#discussion_r1088601323

### DIFF
--- a/doc/man/sphinx-build.rst
+++ b/doc/man/sphinx-build.rst
@@ -138,7 +138,7 @@ Options
    the build directory; with this option you can select a different cache
    directory (the doctrees can be shared between all builders).
 
-.. option:: -j N
+.. option:: -j N, --jobs N
 
    Distribute the build over *N* processes in parallel, to make building on
    multiprocessor machines more effective.  Note that not all parts and not all
@@ -150,6 +150,9 @@ Options
 
    .. versionchanged:: 1.7
       Support ``auto`` argument.
+
+   .. versionchanged:: 6.2
+      Add ``--jobs`` long option.
 
 .. option:: -c path
 
--- a/sphinx/cmd/build.py
+++ b/sphinx/cmd/build.py
@@ -142,7 +142,8 @@ files can be built by specifying individual filenames.
     group.add_argument('-d', metavar='PATH', dest='doctreedir',
                        help=__('path for the cached environment and doctree '
                                'files (default: OUTPUTDIR/.doctrees)'))
-    group.add_argument('-j', metavar='N', default=1, type=jobs_argument, dest='jobs',
+    group.add_argument('-j', '--jobs', metavar='N', default=1, type=jobs_argument,
+                       dest='jobs',
                        help=__('build in parallel with N processes where '
                                'possible (special value "auto" will set N to cpu-count)'))
     group = parser.add_argument_group('build configuration options')


doc/man/sphinx-build.rst
```diff
@@ -151,6 +151,9 @@ Options
    .. versionchanged:: 1.7
       Support ``auto`` argument.
 
+   .. versionchanged:: 6.2
```
hugovk, line 154: versionadded or versionchanged?

6.1.4 or 6.2?

cclauss (approved):
Nice!

doc/man/sphinx-build.rst
CAM-Gerlach: On the former point, the [Sphinx docs themselves](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) are rather ambiguous, as for `versionadded` it says

> This directive documents the version of the project which added the described feature to the library or C API. When this applies to an entire module, it should be placed at the top of the module section before any prose.

which implies that the directive should be an immediate child of the thing added. However, the example given is for a parameter, which doesn't have it's own scope and is instead manually called out in the descriptive text:

> ```rst
> .. versionadded:: 2.5
>   The *spam* parameter.
> ```

And yet, the `versionchanged` description right below it includes "new parameters" as the first example of something a `versionchanged` directive would be used for instead of a `versionadded` (emphasis mine):

> Similar to [versionadded](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded), but describes when and what changed in the named feature in some way (***new parameters***, changed side effects, etc.).

As for my take, I used to favor `.. versionadded:: X.Y The <sub-thing> <sub-thing-type>` as it was a little more concise to read and write, but my thinking has evolved to recommend `.. versionchanged`` instead as it is clearer and more semantically correct, and allows tools and Sphinx extensions (including one I'm building now) parsing the reST source, doctrees, etc to be able to determine unambiguously when a Sphinx-defined construct (function, class, method, struct member, option, etc) was added [^1], It's also the usage I've generally seen in practice, at least in the Python docs and elsewhere.

[^1]: Alternatively, for the latter you could rely on heuristics, e.g. don't assume the whole construct was added if the `versionadded` has descriptive text, and potentially also parsing out parameters added from the description, but it's a lot less reliable, and a custom directive for added parameters (e.g. `.. paramadded:: 3.12 spam` is a much cleaner solution for that).

CAM-Gerlach (approved):
LGTM from me, FWIW

doc/man/sphinx-build.rst
ferdnyc: It is, to a great extent, up to the documenters. Even going by the Python documentation itself (as the most extensive example available), there's little consistency. For example, in the `python` command-line flags documentation, the [documentation for the `-X` family of options](https://docs.python.org/3/using/cmdline.html#cmdoption-X) documents the addition of each new <code>-X <var>arg</var></code> arg as a `versionadded`... but the addition of the `-X` flag _itself_ is documented as a `versionchanged`!! :confused: 

> #### `-X`
> Reserved for various implementation-specific options. CPython currently defines the following possible values:
> 
> `-X faulthandler` to enable [faulthandler](https://docs.python.org/3/library/faulthandler.html#module-faulthandler);
> 
> `-X showrefcount` to output the total reference count and number of used memory blocks when the program finishes or after each statement in the interactive interpreter. This only works on [debug builds](https://docs.python.org/3/using/configure.html#debug-build).
> 
> `-X tracemalloc` to start tracing Python memory allocations using the [tracemalloc](https://docs.python.org/3/library/tracemalloc.html#module-tracemalloc) module. By default, only the most recent frame is stored in a traceback of a trace. Use -X tracemalloc=NFRAME to start tracing with a traceback limit of NFRAME frames. See the [tracemalloc.start()](https://docs.python.org/3/library/tracemalloc.html#tracemalloc.start) for more information.
> 
> `-X int_max_str_digits` configures the [integer string conversion length limitation](https://docs.python.org/3/library/stdtypes.html#int-max-str-digits). See also [PYTHONINTMAXSTRDIGITS](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONINTMAXSTRDIGITS).
> 
> `-X importtime` to show how long each import takes. It shows module name, cumulative time (including nested imports) and self time (excluding nested imports). Note that its output may be broken in multi-threaded application. Typical usage is python3 -X importtime -c 'import asyncio'. See also [PYTHONPROFILEIMPORTTIME](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPROFILEIMPORTT`IME).
> 
> `-X dev`: enable [Python Development Mode](https://docs.python.org/3/library/devmode.html#devmode), introducing additional runtime checks that are too expensive to be enabled by default.
> 
> `-X utf8` enables the [Python UTF-8 Mode](https://docs.python.org/3/library/os.html#utf8-mode). -X utf8=0 explicitly disables [Python UTF-8 Mode](https://docs.python.org/3/library/os.html#utf8-mode) (even when it would otherwise activate automatically).
> 
> `-X pycache_prefix=PATH` enables writing .pyc files to a parallel tree rooted at the given directory instead of to the code tree. See also [PYTHONPYCACHEPREFIX](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPYCACHEPREFIX).
> 
> `-X warn_default_encoding` issues a [EncodingWarning](https://docs.python.org/3/library/exceptions.html#EncodingWarning) when the locale-specific default encoding is used for opening files. See also [PYTHONWARNDEFAULTENCODING](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNDEFAULTENCODING).
> 
> `-X no_debug_ranges` disables the inclusion of the tables mapping extra location information (end line, start column offset and end column offset) to every instruction in code objects. This is useful when smaller code objects and pyc files are desired as well as suppressing the extra visual location indicators when the interpreter displays tracebacks. See also [PYTHONNODEBUGRANGES](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONNODEBUGRANGES).
> 
> `-X frozen_modules` determines whether or not frozen modules are ignored by the import machinery. A value of “on” means they get imported and “off” means they are ignored. The default is “on” if this is an installed Python (the normal case). If it’s under development (running from the source tree) then the default is “off”. Note that the “importlib_bootstrap” and “importlib_bootstrap_external” frozen modules are always used, even if this flag is set to “off”.
> 
> It also allows passing arbitrary values and retrieving them through the [sys._xoptions](https://docs.python.org/3/library/sys.html#sys._xoptions) dictionary.
> 
> _Changed in version 3.2_: The [-X](https://docs.python.org/3/using/cmdline.html#cmdoption-X) option was added.
> 
> _New in version 3.3_: The -X faulthandler option.
> 
> _New in version 3.4_: The -X showrefcount and -X tracemalloc options.
> 
> _New in version 3.6_: The -X showalloccount option.
> 
> _New in version 3.7_: The -X importtime, -X dev and -X utf8 options.
> 
> _New in version 3.8_: The -X pycache_prefix option. The -X dev option now logs close() exceptions in [io.IOBase](https://docs.python.org/3/library/io.html#io.IOBase) destructor.
> 
> _Changed in version 3.9_: Using -X dev option, check encoding and errors arguments on string encoding and decoding operations.
> 
> The -X showalloccount option has been removed.
> 
> _New in version 3.10_: The -X warn_default_encoding option.
> 
> _Deprecated since version 3.9, removed in version 3.10_: The -X oldparser option.
> 
> _New in version 3.11_: The -X no_debug_ranges option.
> 
> _New in version 3.11_: The -X frozen_modules option.
> 
> _New in version 3.11_: The -X int_max_str_digits option.
> 

My take is, since only an alias is added, not warranting a new documentation section, then that only represents a _change_ to the existing documentation for `-j`. (Now, `-j`/ `--jobs`.) So, yeah, +1 for `versionchanged` from me.

CAM-Gerlach: > Even going by the Python documentation itself (as the most extensive example available), there's little consistency.

Yeah, the Python docs are famously inconsistent on stuff like that. We're slowly working on it (though at least a little of it is my fault, since I used to use `versionadded` for that in the past).

PR #11221, opened 2023-03-02
Added TOC object entry support for C++

Rouslan:
Subject: Add TOC object entry support for C++

### Feature or Bugfix
- Feature

### Purpose
- Add TOC object entry support for C++

### Detail
- object entries from the cpp domain are added to the table of contents when toc_object_entries = True in the config
- when toc_object_entries_show_parents = "domain" in the config, the TOC entries omit the namespaces added by the cpp:namespace and cpp:namespace-push directives.


AA-Turner:
Thanks @Rouslan!

A

PR #11222, opened 2023-03-04
Bugfix: Make attribute/"thing" selector much more lax

### LINKED ISSUE #10322: Autodoc fails on enums containing special characters
### Describe the bug

Consider the following enum:
```python3
Colorspace = Enum("Colorspace", "RGB RGBA L LA 1 CMYK CMYK;I P PA")
```

Autodoc'ing this enum will yield the following error:
```
WARNING: invalid signature for autoattribute ('Colorspace.CMYK;I')
WARNING: don't know which module to import for autodocumenting 'Colorspace.CMYK;I' (try placing a "module" or "currentmodule" directive in the document, or giving an explicit module name)
```

It looks like autodoc may be unable to handle enum attributes containing the `;` character.
(In Python, the attribute can be accessed by indexing: `Colorspace['CMYK;I']`)

### How to Reproduce

(see the description)

### Expected behavior

It should be possible to document enum attributes containing a semicolon character (`;`). Respectively any character that is not compatible with dotted access but valid in other notations (e.g. getitem indexing).

### Your project

private, sorry

### Screenshots

_No response_

### OS

Linux Ubuntu 20.04

### Python version

3.8.10

### Sphinx version

4.5.0

### Sphinx extensions

sphinx.ext.autodoc

### Extra tools

_No response_

### Additional context

_No response_

mjsir911:
Previous was expecting a regex "word", corresponding to python variable token rules but with some things like `enum.Enum`, `typing.TypedDict` & `django.models.TextChoices`, passed strings are converted to attributes and so those attributes can have arbitrary naming contents

Fixes #10322

Of note is that I've implemented this by accepting almost anything, but per the bug we could just add a few common breakage characters like `-` or `;`: `[\w;-]`, but this requires more foresight as to what might come up

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -60,7 +60,7 @@ MethodDescriptorType = type(type.__subclasses__)
 py_ext_sig_re = re.compile(
     r'''^ ([\w.]+::)?            # explicit module name
           ([\w.]+\.)?            # module and/or class name(s)
-          (\w+)  \s*             # thing name
+          ([^.()]+)  \s*         # thing name
           (?: \((.*)\)           # optional: arguments
            (?:\s* -> \s* (.*))?  #           return annotation
           )? $                   # and nothing more
--- a/tests/roots/test-ext-autodoc/target/enums.py
+++ b/tests/roots/test-ext-autodoc/target/enums.py
@@ -23,4 +23,10 @@ class EnumCls(enum.Enum):
         pass
 
 
-EnumClsWithSpecialCharAttrs = enum.Enum("EnumClsWithSpecialCharAttrs", "da-sh normal semi;colon")
+EnumClsWithSpecialCharAttrs = enum.Enum("EnumClsWithSpecialCharAttrs", [
+    "da-sh",
+    "normal",
+    "semi;colon",
+    "()",
+    "a()",
+])
--- a/tests/roots/test-ext-autodoc/target/enums.py
+++ b/tests/roots/test-ext-autodoc/target/enums.py
@@ -21,3 +21,6 @@ class EnumCls(enum.Enum):
     def say_goodbye(cls):
         """a classmethod says good-bye to you."""
         pass
+
+
+EnumClsWithSpecialCharAttrs = enum.Enum("EnumClsWithSpecialCharAttrs", "da-sh normal semi;colon")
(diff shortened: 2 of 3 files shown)


mjsir911:
Both the ruff issue & the test issues with `test_format_date` isn't due to my changes.

mjsir911:
> * How do we document a special key containing `()`?

I'm not sure if the current regex can allow ambiguity between a field called `()` & optional arguments list coming after that field, looking into how to regex this one out.

> * Can we change the pattern so that escaped characters are gobbled in the attribute name? (and let autodoc escape the field name when needed).

Confused about this one too, could you elaborate?

picnixz:
> I'm not sure if the current regex can allow ambiguity between a field called () & optional arguments list coming after that field, looking into how to regex this one out.

It would be good to test then and include more pathological field namnes.

> Confused about this one too, could you elaborate?

Yeah, now that I reread myself it is confusing. What I meant is that:

```rst

.. py:attribute:: MyEnum.a\(\)
   
   The ``a()`` member.
```

would be parsed correctly because of the backslashes, so you won't end up with having just `a`.

dimaqq:
Kind poke @AA-Turner maybe?

I was hoping to see this land in Sphinx 8.

I'm not sure about all of the code, but we've cherry-picked the regexp change alone into our conf.py to have the TypedDict keys documentable.

Here's our code:
- Sphinx 6: https://github.com/canonical/operator/blob/828b542d521930c8e225c313eac6995ad7c6d0ea/docs/custom_conf.py#L26-L38
- Sphinx 8: https://github.com/canonical/operator/blob/233aee2e3da109f523c81f4014ded63fe53792db/docs/custom_conf.py#L26-L39 (one line diff wrt. the above)

sphinx/ext/autodoc/__init__.py
```diff
@@ -60,7 +60,7 @@
 py_ext_sig_re = re.compile(
     r'''^ ([\w.]+::)?            # explicit module name
           ([\w.]+\.)?            # module and/or class name(s)
-          (\w+)  \s*             # thing name
+          ([^.()]+)  \s*         # thing name
```
AA-Turner, line 63: I'd prefer to be more conservative here as you suggest - perhaps just add '-', ' ', ';' or etc to begin with.

AA-Turner (commented):
Sorry for the delay. Please could you resolve conflicts and add a CHANGES entry?

A

Issue #11213, opened 2023-02-27
Use a box in the pdf output for guilabel


12rambau:
currently guilabel are displayed as little boxes in the html outputs in most of the the theme I know like rdt:

![image](https://user-images.githubusercontent.com/12596392/221668721-80423121-7c19-4c81-8161-0b0086b8c35a.png)

In the latex build it's only represented as a Italic word. what if instead we use a little box around the text to display it as an actual guilabel as this example: 

![image](https://user-images.githubusercontent.com/12596392/221669041-0ee980e6-ca36-4757-8e29-18edcc3237cf.png)

to build this I simply added the [tcolorbox](https://www.ctan.org/pkg/tcolorbox) package and created a custom command: 

```latex
\newtcbox{\sphinxbtn}[1][]{box align=base, nobeforeafter, size=small, boxsep=2pt, #1}
```

My proposal here is to replace the `\guilabel` one. What do you think ?

jfbu:
The mark-up is via LaTeX macro `\sphinxguilabel`.  Its default definition
```
\protected\def\sphinxguilabel#1{\emph{#1}}
```
is indeed a bit dull.  But you can replace it by
```
\protected\def\sphinxguilabel#1{\mybtn{#1}}
```
with `\mybtn` defined via `\newtcbox`. The `\protected\def` rather than `\renewcommand` is for reasons of the thing ending up in titles, which LaTeX always has trouble with ("moving arguments", "fragile macros", "token not allowed in a PDF string", this kind of things).

Two remarks
1. we will not load `tcolorbox` per default.  It has tremendous LaTeX dependencies and if one uses it extensively slows down the build significantly.  I tested this with using it systematically for code boxes, and it was an order of magnitude slow-down compared to the `pict2e` based approach which has been added to Sphinx at 5.1.0 and activated per default at 6.0.0.
2. the latex support code available in Sphinx since 5.1.0 for rounded boxes with background color and more has its user interface via `'sphinxsetup'` options, but no such interface exists for using directly for other purposes, as in your guilabel proposal, with a `tcolorbox`-like key=value syntax.  Surely this could be added to our support file `sphinxpackageboxes.sty` and adding this as feature request.

Once such a friendly user interface is added, you could have a `\sphinxbox[bgcolor=.., width=..., radius=..., ]{contents}` kind of thing.  The support code is in Sphinx since 5.1.0, but is lacking high-level access.

jfbu:
[...]

PR #11224, opened 2023-03-07
LaTeX: add ``\sphinxbox`` command

jfbu:
Subject: add `\sphinxbox` user interface for inline boxes with possibly rounded corners, background color, even a shadow.

Here is for example what one obtains with the code at time of creation of the PR and this user extra in conf.py:

```
latex_elements = {
    'preamble': r'''
\protected\def\sphinxguilabel#1{\sphinxbox[box_background-TeXcolor=yellow!20,%
                                           boxborder=1pt,%
                                           boxsep=3pt,%
                                           box_border-radius=3pt]{#1}}''',
}
```
(edit: see next comment for final syntax, no more ``boxborder``, ``boxsep``)

for customized rendering in PDF via LaTeX of the `:guilabel:` role:

2023-03-07 screenshot

Code is almost done but still WIP due to desire to proceed further with the refactoring of support for CSS-like `sphinxsetup` options (already in master branch) as there is still potential for sharing some almost identical TeX code which is currently repeated at various places, and got again repeated in the definition of `\sphinxbox`.

Also perhaps add some `\newsphinxbox` in the `tcolorbox` spirit.

And also hesitation about defaults.

(we can discuss separately the specific #11213 question about `:guilabel:`, whether to modify its default or not).

Expect some force pushes, WIP.

Question about a "display" variant will not be part of this PR, which is only about (unbreakable) "horizontal" boxing, i.e. of inline elements not display elements.

### Feature or Bugfix

- Feature

### Related
- #11213

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -18,6 +18,10 @@ Deprecated
 Features added
 --------------
 
+* LaTeX: a command ``\sphinxbox`` for styling text elements with a (possibly
+  rounded) box, optional background color and shadow, has been added.
+  See :ref:`sphinxbox`. (refs #11224)
+
 Bugs fixed
 ----------
 
--- a/doc/latex.rst
+++ b/doc/latex.rst
@@ -35,7 +35,7 @@ benefit from prepared themes. The :ref:`latex-options`, and particularly the
 :ref:`latex_elements <latex_elements_confval>` variable, provides much of the
 interface for customization. For example:
 
-.. code-block:: python
+.. code-block:: latex
 
    # inside conf.py
    latex_engine = 'xelatex'
@@ -233,7 +233,7 @@ Keys that you may want to override include:
 ``'extrapackages'``
    Additional LaTeX packages.  For example:
 
-   .. code-block:: python
+   .. code-block:: latex
 
        latex_elements = {
            'extrapackages': r'\usepackage{isodate}'
@@ -1248,7 +1248,7 @@ customizable:
 
 Here is a random example (not especially recommended!):
 
-.. code-block::
+.. code-block:: latex
 
    latex_elements = {
        'sphinxsetup': """%
@@ -1445,6 +1445,65 @@ Macros
   .. versionadded:: 1.8.3
 
 - ``\sphinxcite``: A wrapper of standard ``\cite`` for citation references.
+
+
+.. _sphinxbox:
+
+The ``\sphinxbox`` command
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 6.2.0
+
+It can be used to "box" inline text elements with all the customizability which
+has been described in :ref:`additionalcss`.  It is a LaTeX command with one
+optional argument, which is a comma-separated list of key=value pairs, as for
+:ref:`latexsphinxsetup`.  Here is the complete list of keys:
+
+- ``border-width``,
+- ``border-top-width``, ``border-right-width``, ``border-bottom-width``,
+  ``border-left-width``,
+- ``box-decoration-break``,
+- ``padding``,
+- ``padding-top``, ``padding-right``, ``padding-bottom``, ``padding-left``,
+- ``border-radius``,
+- ``border-top-left-radius``, ``border-top-right-radius``,
+  ``border-bottom-right-radius``, ``border-bottom-left-radius``,
+- ``box-shadow``,
+- ``border-TeXcolor``, ``background-TeXcolor``, ``box-shadow-TeXcolor``.
+
+Refer to :ref:`additionalcss` for important syntax information.  The default
+configuration uses no shadow, a border-width of ``\fboxrule``, a padding of
+``\fboxsep``, rounded corners (with radius ``\fboxsep``) and background and
+border colors as for the default rendering of code-blocks.  One can modify
+these defaults via using ``\sphinxboxsetup[key=value,...]`` or also via
+``\sphinxsetup`` but all key names must then be prefixed with ``box_``.
+
+A utility ``\newsphinxbox`` if provided to create an alternative,
+e.g. ``\foo`` to ``\sphinxbox`` with a different set of defaults, but which
+otherwise behaves the same.  Here is a random example of use:
+
+.. code-block:: latex
+
+   latex_elements = {
+       'preamble': r'''
+   % define a sphinxbox with some defaults:
+   \newsphinxbox[border-width=4pt,%
+                 border-radius=4pt,%
+                 background-TeXcolor=yellow!20]{\foo}
+   % use this \foo to redefine rendering of some text elements:
+   \protected\def\sphinxguilabel#1{\foo{#1}}
+   \protected\def\sphinxmenuselection#1{\foo[background-TeXcolor=green!20,
+                                             border-width=1pt,
+                                             box-shadow=3pt 3pt,
+                                             box-shadow-TeXcolor=gray]{#1}}
+   % and this one will use \sphinxbox directly
+   % one can also add options within square brackets as in usage of \foo above
+   \protected\def\sphinxkeyboard#1{\sphinxbox{\sphinxcode{#1}}}
+   ''',
+   }
+
+In the above example, you can probably use ``\renewcommand`` syntax if you
+prefer (with ``[1]`` in  place of ``#1`` then).
 
 Environments
 ~~~~~~~~~~~~
--- a/sphinx/texinputs/sphinx.sty
+++ b/sphinx/texinputs/sphinx.sty
@@ -311,8 +311,6 @@ will be set to white}%
 % Except that \sphinxbox is for boxing inline elements (no linebreak, no
 % pagebreak).
 %
-% (this addition happens in a descendant commit of this one)
-%
 % MEMO: the 6.2.0 refactoring induced a change that the default \fboxrule used
 % by verbatimborder and shadowrule is now not expanded to its value immediately in
 % preamble (no assignment to a \dimen register is done, only storage in a
@@ -361,6 +359,8 @@ will be set to white}%
 \spx@tempa{spx@attention@}{div.attention_}{attentionborder}{1pt}
 \spx@tempa{spx@danger@}   {div.danger_}   {dangerborder}   {1pt}
 \spx@tempa{spx@error@}    {div.error_}    {errorborder}    {1pt}
+% this one new at 6.2.0: (we do not create a "legacy name" for it)
+\spx@tempa{spx@box@}      {box_}          {box_border-width}\fboxrule
 % Set default box-decoration-break style for codeblocks to slice
 \spx@pre@border@opentrue % new default at 6.0.0: slice, not clone
 
@@ -401,17 +401,11 @@ will be set to white}%
 \spx@tempa{spx@attention@}{div.attention_}  {\dimexpr.6\baselineskip-\spx@attention@border\relax}
 \spx@tempa{spx@danger@}   {div.danger_}     {\dimexpr.6\baselineskip-\spx@danger@border\relax}
 \spx@tempa{spx@error@}    {div.error_}      {\dimexpr.6\baselineskip-\spx@error@border\relax}
+\spx@tempa{spx@box@}      {box_}            \fboxsep % new at 6.2.0
 % define legacy verbatimsep key as alias of pre_padding key
 \expandafter\let\expandafter\KV@sphinx@verbatimsep\csname KV@sphinx@pre_padding\endcsname
 % define legacy shadowsep key as alias of div.topic_padding key
 \expandafter\let\expandafter\KV@sphinx@shadowsep\csname KV@sphinx@div.topic_padding\endcsname
-% (temporary explanation: this next comment and code line belong to a descendant
-%  commit down the line for \sphinxbox support, left-overs from a rebase)
-% also define a boxsep key as alias of box_padding key, because we defined boxborder
-% boxborder was defined only as collateral of batch definitions so some hesitation
-% we do keep these boxborder and boxsep as they configure the probably two most
-% important parameters; still I feel not definitive about what is best to do
-\expandafter\let\expandafter\KV@sphinx@boxsep\csname KV@sphinx@box_padding\endcsname
 
 % Corner radii keys
 %
@@ -438,6 +432,7 @@ will be set to white}%
 \spx@tempa{spx@attention@}{div.attention_}  \z@
 \spx@tempa{spx@danger@}   {div.danger_}     \z@
 \spx@tempa{spx@error@}    {div.error_}      \z@
+\spx@tempa{spx@box@}      {box_}            {3pt}% new and tentative default at 6.2.0
 
 % Shadow keys
 %
@@ -455,6 +450,7 @@ will be set to white}%
 \spx@tempa{spx@attention@}
 \spx@tempa{spx@danger@}
 \spx@tempa{spx@error@}
+\spx@tempa{spx@box@}% new at 6.2.0
 %
 \def\spx@tempa#1{% #1 = macro prefix
   \expandafter\spx@tempb
@@ -495,6 +491,7 @@ will be set to white}%
 \spx@tempa{spx@attention@}{div.attention_}
 \spx@tempa{spx@danger@}   {div.danger_}
 \spx@tempa{spx@error@}    {div.error_}
+\spx@tempa{spx@box@}      {box_}% new at 6.2.0
 
 % Support for legacy shadowsize
 % This definition was broken at 5.1.0 and fixed at 6.1.2
@@ -531,6 +528,7 @@ will be set to white}%
 \spx@tempa{spx@attention@}
 \spx@tempa{spx@danger@}
 \spx@tempa{spx@error@}
+\spx@tempa{spx@box@}% new at 6.2.0
 %
 \def\spx@tempa#1{% #1 = macro prefix
   \expandafter\spx@tempb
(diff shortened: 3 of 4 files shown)


jfbu:
In the final version, the key names have changed a bit from what is in my initial message on top, no more a  ``boxborder`` or ``boxsep`` but exclusively the CSS-like syntax from 5.1.0.

Here is an example rendering:
Capture d’écran 2023-03-07 à 12 37 11

With this in conf.py
```
latex_elements = {
    'preamble': r'''
% define a sphinxbox with some defaults:
\newsphinxbox[border-width=4pt,%
              border-radius=4pt,%
              background-TeXcolor=yellow!20]{\foo}
% use this \foo to redefine rendering of some text elements:
\protected\def\sphinxguilabel#1{\foo{#1}}
\protected\def\sphinxmenuselection#1{\foo[background-TeXcolor=green!20,
                                          border-width=1pt,
                                          box-shadow=3pt 3pt,
                                          box-shadow-TeXcolor=gray]{#1}}
% and this one will use \sphinxbox directly
% one can also add options within square brackets as in usage of \foo above
\protected\def\sphinxkeyboard#1{\sphinxbox{\sphinxcode{#1}}}
''',
}
```

and

```rest
Some text before :guilabel:`fa-folder` some text :menuselection:`Edit` some
text :kbd:`Ctrl-x Ctrl-f`
```

as input.

jfbu:
Failure of  ``test_gettext_dont_rebuild_mo`` on Windows does not look to me as being related. So I will merge as I can't be sure of when I can be active here.

n-peugnet:
Excellent! I made my own with tcolorbox as this package makes it easy. But one thing it cannot do is allowing inline boxes to break between words (even less so inside words).

Are these boxes allowed to break? Can it be an option?
Thank you for your work on Sphinx's $\LaTeX$ rendering!

jfbu:
@n-peugnet Thanks for comment!

About:

> Are these boxes allowed to break? Can it be an option?

Sadly, no.  I added a comment in the source:
https://github.com/sphinx-doc/sphinx/blob/ee3c720cfa6d8fcbfbe01c52bbe78a6ba0e8ed49/sphinx/texinputs/sphinx.sty#L807-L813

With a few more details: I am not aware of a LaTeX package offering fully functional (i.e. with no or very few restrictions on input) breakable horizontal boxes.
- basically, one has to box individually each character adding invisible glue in-between to allow a line break,
- but the LaTeX programmer can not assume the input is only composed of character and space tokens,
- it can be composed of quite a lot of LaTeX commands nested one within the other, or primitive TeX boxing operations, or switch to math mode, etc...
- even if you insist that input is only stream of characters, a good boxing should not interfere with TeX ligatures and intra-character kerning.

There is a well-known 20+ years old package called [soul](https://ctan.org/pkg/soul) which attempts to add underlining or other effets, and in particuler background highlighting via its macro `\hl`.  It is unmaintained and for many many years was not compatible with UTF-8 in pdfLaTeX, and very recently, a few weeks ago, LaTeX maintainers merged it with companion "soulutf8" package, which addresses that.  Underlining and other effects can also be achieved with the even older [ulem](https://ctan.org/pkg/ulem) package but with any braced content such as a ``\emph{some words}`` that part will not break across lines.  Here is some example output using the more able  [soul](https://ctan.org/pkg/soul) :
Capture d’écran 2023-03-14 à 10 18 31
where I inserted a nested ``\emph`` inside the `\hl` command of the package: one can see it does obey the line break.

For simply adding a background color, the [soul](https://ctan.org/pkg/soul)  ``\hl`` used inside a redefinition of ``\sphinxguilabel`` for example should work.  However I looked at the package documentation and unfortunately if you want to do anything such as controlling the size of the colored area, you will have to enter somewhat deep into the package code and identify ``@`` letter macros to modify.  And I have no idea if it is feasible for example to treat the first and last characters differently.

In the past I have implemented for Sphinx the line-breaking of long code lines highlighted by Pygments: the ``verbatimwrapslines`` and ``verbatimforcewraps`` options of ``'sphinxsetup'``.  The former was of a nature very different from the issue we discuss here, it was a question of interaction with the LaTeX package used by Sphinx called ``fancyvrb``  [1]_.   But for the ``verbartimforcewraps`` (off by default) it is very similar to the issue here (once the ``verbatimwrapslines`` thing has been solved to start with).  To implement it I had to hijack the Pygments mark-up to grab character per character.

As long as you control fully the input mark-up you can probably do something.  In the case of highlight code, a very important thing is that the ``\PYG`` macro is never nested.  But for Sphinx LaTeX mark-up of roles, probably some nesting can happen, and it would require time to overview all possibilties.  And in pdflatex case, UTF-8 characters are special, one has to identify their "active" first byte, which was the more complex part in my implementation of ``verbatimforcewraps`` (notice that in the LaTeX world at large, package authors -- for example for glossary or indexing or bibliography related packages -- have often requested the LaTeX team a robust way to grab the first character of a word, which as far as I know, started having official LaTeX-supported solutions only quite recently, for package authors using the LaTeX3 programming interface).  And for doing this I had to support only ``utf8`` I could not support the ``utf8x`` which in recent years has become more and more incompatible with upstream LaTeX evolving support of Unicode input.

Let's say you have solved all these problems, there is still some work to do for boxing with decorations such as a border or a shadow or rounded corners as provided by ``\sphinxbox`` : rounded corners for example have to apply only for start and finish... so as said above one has to identify the first and last characters and treat them differently from others and as I indicate at end of footnote below, I am running out of time available for Sphinx right now, and for surely at least 4 or 5 months.  If it was only a matter to handle a stream of characters it is feasible somewhat quickly, only difficulty is the business of utf-8 in pdflatex.  In lualatex or xelatex it is trivial as each Unicode character looks as only one token to TeX.

Thus if you use ``\sphinxbox``, there is currently and for the foreseeable future no breaking of contents across successive lines.  If implemented one day, it will have to be activated by some option, as it will go along quite complex extra TeX coding. For a simple background highlighting or underlining or overstriking, you can try the macros such as `\hl` from package [soul](https://ctan.org/pkg/soul) however I have no idea how this will work if its input contains nested `\sphinx...` macros which is not to exclude.

Many years ago I had applied some highlighting of my own in the documentation of a LaTeX package I authored.  But all the little colorboxes ended up causing some stack overflow problem (not sure if I was using pdflatex or latex+dvipdfmx though, the latter behaving differently in matters of colors, it can not have multiple color stacks).

Probably the LuaTeX is the more-or-less sole hope that one can do something general: it offers hooks into the node lists created by TeX.  Perhaps, probably one can use that.  However this would be heavy-duty LuaTeX thing and I am simply lacking both the knowledge and the will to invest time into it.  If it was a matter of a one-hour effort perhaps, but this looks more like weeks...  and I searched a bit and did not identify a LuaTeX-specific add-on offering this, perhaps it exists nevertheless. (maybe [this on overleaf](https://www.overleaf.com/learn/latex/Articles/Pandora%E2%80%99s_%5Chbox%3A_Using_LuaTeX_to_Lift_the_Lid_of_TeX_Boxes) can be a LuaTeX ouverture).

(perhaps if you ask on a LuaTeX forum they will tell you of course it is well-known and you only have to do this or that...)

.. [1]: digression: it is a historical legacy from prehistory of package that Sphinx uses ``fancyvrb``  for code-blocks, the principal reason being to actually force input lines to give one-to-one output lines, i.e. deactivate the paragraph reflowing ; in retrospect there was *much simpler* way to do this than employ ``fancyvrb`` which also is designed to escape all special-to LaTeX characters -- but Pygments has already done that for us! I considered many times getting rid of ``fancyvrb`` entirely, but for a project maintained on my free time and in view of potential complications, this has low priority -- and for next few months I will surely not have time, I ended right now a period where I could do such things.

n-peugnet:
@jfbu Thank you for this very detailed answer. Sorry I could have checked the code myself before asking. I did some research of my own back when I made my version of this (that will probably be replaced by yours once I get to upgrade Sphinx), and I quickly understood that it was very complicated to do in LaTeX. The [only beautiful solution I found](https://tex.stackexchange.com/a/429498) was indeed based on soulutf8 package combined with Tikz and the code is still quite complicated. .

jfbu:
@n-peugnet ah well of course I had forgotten the more modern stuff which uses TikZ ability and multiple passes to put absolute positional marks on the page and then add graphical layers on next run. I am not sure this works also with platex, and I saw problem reports with xetex from 2016 circa, perhaps solved since (pdftex has built-in such things `\pdfsavepos/\pdflastxpos/\pdflastypos` and one could use that for simple-minded tasks, not sure if TikZ pdftex graphics driver builds upon that, and I do not know what it does for dvipdfmx driver). 

The [tikzmark](https://ctan.org/pkg/tikzmark) package has a *highlighting* library, documented starting on page 17 of its pdf documentation, which seems to implement exactly the naive idea one could have about positioning  a start and an end, and interpolating.  However, you already point out to [a more advanced method](https://tex.stackexchange.com/a/429498) indeed based upon [tikzmark](https://ctan.org/pkg/tikzmark) but in combination with [soul](https://ctan.org/pkg/soul).  Why they have to use [soul](https://ctan.org/pkg/soul) may have to do with allowing contents of varying height, or allowing pagebreaks. But using  `soul` adds its intrinsic limitations.

About LuaTeX, I should have digged a bit more, there is the [lua-hl](https://ctan.org/pkg/lua-ul) package whose description is:
> This package provides underlining, strikethough, and highlighting using features in LuaLaTeX which avoid the restrictions imposed by other methods. In particular, kerning is not affected, the underlined text can use arbitrary commands, hyphenation works etc. 
> The package requires LuaTeX version ≥ 1.12.0. 

The important part here is: *the underlined text can use arbitrary commands*.

However I can not engage into LuaLaTeX only code here at Sphinx. (assuming it is possible to achieve what we want with it).

I think though, I can provide an experimental  `\sphinxbreakablebox` which will simply attempt to grab character per character and insert breaking points in-between.  This can also add data to display some special sign at a line break as for codelines.  As I said above, similar sthings are already done by Sphinx in the bowels of ``sphinxlatexliterals.sty``. I can try to support nested mark-up by making simplifying assumptions about the encountered commands, not sure if I can invest the time in reviewing all possible Sphinx produced mark-up from e.g. LaTeX escapes.

The last two days I completed and finished my on-going refactoring of Sphinx LaTeX internals with adding support for `box-decoration-break=slice` to admonitions and topic directives.  Doing this horizontally is only a variant... *once*  the character per character split is achieved. In the "vertical" case, this is much different as package ``framed`` has used TeX page building behind the scenes to break the input into large rectangular rendered boxes, and the Sphinx code only has to wrap these boxes in suitable decoration; for the horizontal case, the big problem is to know where TeX will horizontally break, but at the stage when this gets done, material is split into horizontal lines and it is about impossible to dissect them at macro level; the `soul` package innards force TeX to apply its line-breaking algorithm in a vertical box of minuscule 1/65536 TeX pt width, thus TeX applies fully all possibe hyphenation, then `soul` dissects this vertical box into its horizontal components which hopefully more or less represent each one character,  but there are a lot of complications to recover kerning etc...

jfbu:
@n-peugnet I have implemented an experimental breakable box.  Maybe I will include this as something to use at your own risk.  From previous experience with using many small boxes one next to the other, there may be problems with long documents of pdf color stack.
- I have to add a strut to each character to equalize the heights
- Unfortunately I can not show the result for rounded boxes, due to an issue with pict2e when stroking a zero width path, there is still some visible alias on screen; I had already seen this problem and I am not sure what is its origin. So, inhibiting the right border of first stuff does not work if the other corners are rounded, and same for last character.  In-between stuff use the "rectangle" routine anyhow.
- I added support for UTF-8 and even for input with some macros inside. To test I inserted manually an extra `\textit` to get the second screenshot below. Supported macros are only of the `\text...̀  type, i.e. those which modifies font properties.  Any macro doing some examination of its argument will be utterly broken.
- Shadow is not supported if on the left; but ok if on right, as the screenshots demonstrate.  A priori the code caters for inset shadow but as I did not test that part, it has probably some bug.
- we can see some issues because there is no inter-small-boxes stretches, so it protrudes a bit unevenly in right margin
- This is home made stuff with a few dozen lines, can not compete with approaches using the full TikZ apparatus or LuaLaTeX engine features...
- If included in Sphinx, it will be left as "use at your own risk".
- There may be some problems with some PDF viewers with antialiasing effetc of adjacent boxes.  The shadows here insert some "glueing" which may help counteract such a problem. I used evince on Ubuntu 20.04 for this.

Example of output:

Capture d’écran du 2023-03-14 19-59-19

Capture d’écran du 2023-03-14 20-06-18

index.rst:
```rest
TEST
----

aaaa bbbb some text :guilabel:`some very œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù
éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù
éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù
œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç
éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù
éèàù éèàù éèàù long guilabel` This was a very very long guitlabel. Did it wrap correctly?
```

conf.py:
```
latex_elements = {
    'preamble': r"""
  \protected\def\sphinxguilabel#1{%
    \sphinxbreakablebox[% commented-out options are the default from sphinx.sty
                        % border-width=\fboxrule,
                        % padding=\fboxsep,
                        % border-radius=\fboxsep,
                        % border-TeXcolor={RGB}{32,32,32},%  default VerbatimBorderColor
                        % background-TeXcolor={gray}{0.95},% default VerbatimColor
                        % Now some extra settings:
                        border-radius=0pt,%  (needed currently due to an issue with pict2e;
                                          %   even setting width to 0pt, the lateral border
                                          %   is seen. So the start and end, if using the
                                          %   "rounded" routine will have an annoying extra border
                                          %   right where we don't want it)
                        border-width=1.5pt,
                        box-shadow=2pt 2pt,
                        box-shadow-TeXcolor=gray!50]{#1}}
"""
}
```

The LaTeX with an extra `\textit`, only as proof of concept.  But the way it is implemented is very fragile. The main problem is that the 2-argument macro `\sphinxhref` will cause outrageous breakage.  Not sure if it can end up there anyhow.

```latex
\sphinxAtStartPar
aaaa bbbb some text \sphinxguilabel{some very œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù
éèàù éèàù \textit{œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù
éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù
œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç
éèàù éèàù éèàù éèàù} œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù éèàù éèàù éèàù œîôêç éèàù
éèàù éèàù éèàù long guilabel} This was a very very long guitlabel. Did it wrap correctly?
```

I tested only UTF-8 with some diacritics useful in my mother tongue.

If you want to test it at your locale you should checkout the master branch at my forka at commit https://github.com/jfbu/sphinx/commit/e8efd18e04051181ad13c4dbefd17c7358cdfebc

Hmm, I will create a branch so that this commit does not disappear as soon as I update my master branch. Done at  https://github.com/jfbu/sphinx/pull/new/wip_breakable_sphinxbox

jfbu:
Merge of master into my workbranch [wip_breakable_sphinxbox](https://github.com/jfbu/sphinx/tree/wip_breakable_sphinxbox) has allowed "breakablebox" to support rounded corners.

Capture d’écran du 2023-03-15 18-35-07

But the -one-box-per-character things induces aliasing artifacts at box junctures at least in my evince viewer at some zoom levels.  One sees them above.  I am not sure if they come from some minuscule differences in coordinates, but many many years ago I experimented with such problems in a context where one was sure all coordinates computed by TeX where perfectly identical from one rectangle to the adjacent one.  Then perhaps something happens when converting to PostScript or PDF internals. Typically one sees some fine border but when zooming it sometimes disappear or reappear but never changes size on screen. 

Personnaly I would prefer occasional problem with a non-breaking box than this.  After some polishing I might merge it though, but will not consider this as a feature to be maintained at all costs.

There is a big difference between the user point of view who wants things done and the maintainer point of view: already Sphinx for many years allows a lot of LaTeX customization such as using the full TikZ for such `\sphinxguilabel` thing (it is much more difficult to plug-in externalities into for example code-blocks support code).  And user does not care if that adds immense dependency and multiplies par 10 the build time and perhaps limit to only LuaLaTeX.  But for maintainer it is an altogether different business...

Merge of master allows also varying border width also for rounded boxes:
Capture d’écran du 2023-03-15 18-52-10

n-peugnet:
@jfbu:

This is very impressive! I didn't expect you to do it when I posted my previous comment. I hope you didn't feel obliged to do it because of me and that it was a fun experiment!

> Personnaly I would prefer occasional problem with a non-breaking box than this. After some polishing I might merge it though, but will not consider this as a feature to be maintained at all costs.

I completely agree with you on this. I was more suggesting the idea and linking to my findings to get your opinion and maybe give you some ideas. :slightly_smiling_face:

jfbu:
@n-peugnet I did some final wrap-up and my effort is at https://github.com/jfbu/sphinx/tree/wip_breakable_sphinxbox

To use, grab the file `sphinxbreakablebox.sty` and either inject it via `latex_additional_files` in your projects or put it in some place where LaTeX can find it, and in both cases add `\usepackage{sphinxbreakablebox}` to the `'preamble'` entry of `latex_elements`.
But you must be using upcoming Sphinx 6.2.0 or, at this time, the current tip of master here.

More comments inside the file.  This will not get merged in Sphinx core for time being, too hacky, too much workload to test it thoroughly in Sphinx context, and the problems with the PDF viewers artifacts.

PR #11226, opened 2023-03-08
Display examples of roles in the documentation

hugovk:
Subject: Show examples of roles in the docs so we can see how they look

### Feature or Bugfix

- Feature

### Purpose
- Showing how roles are displayed helps us decide whether to use them

### Detail
- Display examples of the roles in the docs

### Before
- https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html

### After
- https://sphinx--11226.org.readthedocs.build/en/11226/usage/restructuredtext/roles.html


AA-Turner:
Thanks as always @hugovk!

PR #11081, opened 2023-01-03
Fix #11074: Can't change sphinxnote to use sphinxheavybox

jfbu:
I have made an entry in Bugfixes, but as I commened in #11074 I did not consider the changed behavior at 5.1.0 as a bug (it would have been a regression), as it was never said explicitly one could use `sphinxheavybox` without precautions directly.

Memo: this allows to let `sphinxnote` etc... use `sphinxheavybox` but do not deduce one can wrap arbitrarily this way for a new environment unrelated to admonitions: under the hood there is `sphinxadmonition` dispatch which does some preparatory steps before the `sphinxnote` etc... are actually executed.

Fixes #11074

Issue #11074, opened 2023-01-02
Can't change sphinxnote to use sphinxheavybox starting with 5.1.0


therealprof:
### Describe the bug

I used to redefine the "light" box environments to use heavybox instead:

```
\renewenvironment{sphinxnote}[1] {\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}
\renewenvironment{sphinxhint}[1] {\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}
\renewenvironment{sphinximportant}[1] {\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}
\renewenvironment{sphinxtip}[1] {\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}
```

However, starting with 5.1.0 I get the following error(s) when running LaTeX on the generated code:

```
Chapter 1.
(/usr/local/texlive/2022/texmf-dist/tex/latex/base/ts1cmss.fd)
! Missing number, treated as zero.
<to be read again>
                   \spx@note@border@top
l.181 \begin{sphinxadmonition}{note}{Note:}
```

with previous versions this worked fine and as intended. I haven't been able to find a workaround but I suppose this might be related to the introduction of the new configuration attributes in #10648 which doesn't exist for notes?

### How to Reproduce

Add this to your `conf.py` and try to render a document which uses notes:
```
latex_elements = {
  'preamble':
      r'''\renewenvironment{sphinxnote}[1] {\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}'''
}
```

### Environment Information

```text
Platform:              darwin; (macOS-10.16-x86_64-i386-64bit)
Python version:        3.9.15 (main, Oct 11 2022, 22:25:13)
[Clang 12.0.0 (clang-1200.0.32.29)])
Python implementation: CPython
Sphinx version:        6.0.0
Docutils version:      0.18.1
Jinja2 version:        3.1.2
```

### Additional context

5.0.2 works just fine, every later version I tried after that is broken.

jfbu:
Yes since #10648 the heavybox environment executes some LaTeX set-up which depends on the name of the notice (warning, attention, etc...) and configures the rendering to match either the default or  the `conf.py`  setting of the relevant `'sphinxsetup'` keys.
[...]

PR #11234, opened 2023-03-11
LaTeX: add support for styling ``note`` (``hint``, ...) similarly as ``warning`` ( ``danger``, ...)

jfbu:
Subject: build on recent refactoring to add possibility to render note-type admonitions the same way as warning-type, i.e. use for them ``sphinxheavybox`` and have all the customizability via CSS-named alike options.

### Feature or Bugfix

- Feature

Relates #11074 in so far as it is careful to keep working the possibility for user to do things such as 
```
\renewenvironment{sphinxnote}[1]{\begin{sphinxheavybox}\sphinxstrong{#1} }{\end{sphinxheavybox}}
```
which was restored at #11081.

But it is not even needed to add such code to preamble now, as ``sphinxheavybox`` will be used automatically for ``note`` if one of the CSS-named alike options for it added by this PR is used, such as ``div.note_border-width``.  

Also using the new ``noteBgColor`` option (its CSS name alias is ``div.note_background-TeXcolor``) to configure a background color, will automatically trigger usage of ``sphinxheavybox``.

The above ``\renewenvironment`` by itself allowed only a very limited usage of ``sphinxheavybox``, now  the full range of its customizability is accessible.

### DIFF
--- a/sphinx/texinputs/sphinxlatexadmonitions.sty
+++ b/sphinx/texinputs/sphinxlatexadmonitions.sty
@@ -11,6 +11,7 @@
 %   This is a dispatch supporting
 %
 %   - note, hint, important, tip                 (via sphinxlightbox)
+%                    (also optionally via sphinxheavybox since 6.2.0)
 %   - warning, caution, attention, danger, error (via sphinxheavybox)
 %
 %   Each sphinx<notice name> environment can be redefined by user.
@@ -35,7 +36,14 @@
 % Some are quite plain
 \newenvironment{sphinxseealso}[1]{\sphinxstrong{#1:}\par\nopagebreak}{}
 
-% the spx@notice@bordercolor etc are set in the sphinxadmonition environment
+% This \dimen register is a legacy relic from Sphinx 1.5, as well as
+% spx@notice@border@color (used in sphinxlightbox environment and set in
+% sphinxadmonition environment) which one could dispose of, using \spx@notice
+% which is defined by sphinxadmonition environment.  Let's simply keep as used
+% since 1.5.
+\newdimen\spx@notice@border
+
+% spx@notice@bordercolor and \spx@notice@border are set in sphinxadmonition environment
 \newenvironment{sphinxlightbox}{%
   \par
   \noindent{\color{spx@notice@bordercolor}%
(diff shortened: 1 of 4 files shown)


jfbu:
I will squash and merge after testing completes.

Refer then to https://www.sphinx-doc.org/en/master/latex.html#additional-css-like-sphinxsetup-keys for the documentation of the new features.

sphinx/texinputs/sphinxlatexadmonitions.sty
```diff
@@ -51,78 +59,59 @@
...
+%
+% Then at 6.2.0, note, hint, important, tip can now access full customization
+% of sphinxheavybox: CSS-named alike options (if used) trigger sphinxheavybox.
+% The \renewenvironment method still works but then only common border width,
+% non-configurable padding, straight corners and no shadow are all implied,
+% as long as the related CSS-named options are not used.
 \newenvironment{sphinxnote}[1]
-  {\begin{sphinxlightbox}\sphinxstrong{#1} }{\end{sphinxlightbox}}
+  {\edef\spx@env{sphinx\ifspx@opt@heavynote heavy\else light\fi box}%
+   \expanded{\noexpand\begin{\spx@env}}\sphinxstrong{#1} }
```
jfbu, line 87: This ``\expanded`` requires a pdflatex from 2019 at most.  I will need to revert to older LaTeX syntax.

sphinx/texinputs/sphinxlatexadmonitions.sty
```diff
@@ -194,19 +193,18 @@
...
   \sphinxcolorlet{spx@notice@bordercolor}{sphinx#1BorderColor}%
   \sphinxcolorlet{spx@notice@bgcolor}{sphinx#1BgColor}%
   \spx@notice@border \dimexpr\csname spx@#1@border\endcsname\relax
-  % start specific environment, passing the heading as argument
+  % trigger the sphinx<type> environment, #2=heading is passed as argument
   \begin{sphinx#1}{#2}}
-  % workaround some LaTeX "feature" of \end command
- {\edef\spx@temp{\noexpand\end{sphinx\spx@noticetype}}\spx@temp}
+  % workaround some LaTeX "feature" of \end command (can't use "sphinx#1" here)
+ {\expanded{\noexpand\end{sphinx\spx@noticetype}}}
```
jfbu, line 208: I should not have use `\expanded` here, as it is not available prior to 2019 (except with lualatex).

PR #11243, opened 2023-03-15
LaTeX: let rounded boxes support up to 4 distinct border-widths

jfbu:
Since 5.1.0, rounded boxes (which only got started being used by default for code-blocks at 6.0.0) mandated a constant border width.  Lift this restriction.

In passing, fix a regression at ad3235421, which broke `pre_box-shadow-TeXcolor` option.

### Feature or Bugfix
- Feature
- Bugfix

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -27,6 +27,7 @@ Features added
   optional shadow.  See :ref:`additionalcss`. (refs: #11234)
 * LaTeX: admonitions and :dudir:`topic` (and contents_) directives,
   and not only :rst:dir:`code-block`, support ``box-decoration-break=slice``.
+* LaTeX: let rounded boxes support up to 4 distinct border-widths (refs: #11243)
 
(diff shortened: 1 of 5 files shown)


jfbu:
This gives a minimal example of use, where only the border-right-width was changed. I added for this chapter in our document a shadow to code-blocks in addition to the cool colors chosen already some months back and which are awe-inspiring.  

Capture d’écran du 2023-03-15 17-48-22

And here is a more varied example: (here with `.. topic::` directive)

Capture d’écran du 2023-03-15 17-52-55

not sure what to do with bottom shadow for `box-decoration-break=slice` simpler to simply leave it.

Inset shadow are not compatible with rounded boxes.  Formerly they would force rectangular box, now to the contrary rounded corners force ignoring inset shadow.  I deemed this change not breaking as the two together never have been compatible anyhow.

PR #11258, opened 2023-03-21
Add missing quotes in html 4 deprecation text

rffontenelle:
Subject: Just a missing closing double-quote in html 4 deprecation warning that I spotted when translating.


jfbu:
Thanks! As I merged this via a web interface rebase, this changed the SHA of the commit...

PR #11276, opened 2023-03-31
Fix #11274: external links are not properly escaped for ``\sphinxupquote``   compatibility

### LINKED ISSUE #11274: External links are not properly escaped in LaTeX
### Describe the bug

When compiling a LaTeX document (independently of the underlying engine), external links containing ligatures are incorrectly escaped by `\sphinxhref` if they are inside a `\sphinxupquote` block.

The issue lies in the definition of `\sphinxupquote` itself:

https://github.com/sphinx-doc/sphinx/blob/c951f5f6393e4e3b4d820eb7d2e243fdc17b2cc1/sphinx/texinputs/sphinxlatexliterals.sty#L963-L994

As we can see, 

https://github.com/sphinx-doc/sphinx/blob/c951f5f6393e4e3b4d820eb7d2e243fdc17b2cc1/sphinx/texinputs/sphinxlatexliterals.sty#L993

defines `-` as a "special" character which when put inside a `\sphinxupquote` may be handled differently. The tex content being generated contains the following block:

```tex
\begin{fulllineitems}
\phantomsection\label{\detokenize{index:MyClass.config}}
\pysigstartsignatures
\pysigline{%
  \sphinxbfcode{\sphinxupquote{config}}%
  \sphinxbfcode{%
    \sphinxupquote{% <- the '-' will be interpreted differently inside 
      \DUrole{p}{:}\DUrole{w}{  }%
      \sphinxhref%
        {https://www.sphinx-doc.org/en/master/extdev/appapi.html\#sphinx.config.Config}%
        {sphinx.config.Config}}}}
\pysigstopsignatures\begin{quote}

\sphinxAtStartPar
A configuration object.
\end{quote}

\end{fulllineitems}
```

While the content being generated is correct, the actual target (i.e., what is really inside the resulting `\href`) is `https://www.sphinx\unhbox \voidb@x \kern \z@ \ETC. [...]`. This can be seen by adding after

https://github.com/sphinx-doc/sphinx/blob/c951f5f6393e4e3b4d820eb7d2e243fdc17b2cc1/sphinx/texinputs/sphinxlatexliterals.sty#L935

something like `\PackageWarning{}{\scantokens{\href{#1}}}` (note that the compilation *WILL* fail because the commands inside are still interpreted, but this is only for inspection). For instance, clicking on the link from a PDF viewer (e.g., Okular) shows the following error:

```
Malformed URL
Invalid hostname (character '%' not permitted); source was "x %5Ckern %5Cz@ %5Cchar %60%5Cprotect %5Cdiscretionary %7B%5Cchar %5Cdefaulthyphenchar %7D%7B%7D%7B%7D doc.org"; scheme = "https", userinfo = "www.sphinx\unhbox \voidb", host = "", path = "/en/master/extdev/appapi.html", fragment = "sphinx.config.Config"
``` 

# Simple Fix

Here is a hotfix using a custom extension:

```
_SPHINX_LITERAL_NOLIG = frozenset('`<>\'-')  
_is_unsafe = _SPHINX_LITERAL_NOLIG.__contains__

def quote_uri(refuri):
    r"""
    Safe escape for LaTeX links.

    This escapes all non-alphanumeric characters, which is stronger than simply
    escaping the URI using :func:`urllib.parse.quote`.

    :param refuri: The link to escape.
    :type refuri: str
    :return: A link that can be used in LaTeX.
    :rtype: str

    This patches the following code::

        \sphinxupquote{       # <- exclude some tokens that may appear in URI
            ...
            \sphinxhref{URI}  # <- fail if URI has non-alphanumeric characters
        }
    """

    def transform(c):
        return f'%{ord(c):02X}' if _is_unsafe(c) else c
    return ''.join(map(transform, refuri))


class CustomLaTeXTranslator(LaTeXTranslator):
    def visit_reference(self, node):
        if node.hasattr('refuri'):
            node['refuri'] = quote_uri(node['refuri'])
        super().visit_reference(node)

def setup(app):
    app.set_translator('latex', CustomLaTeXTranslator, override=True)
    return {'parallel_read_safe': True, 'parallel_write_safe': True}
```

### index.rst

```rst
Foo
===

.. py:class:: MyClass

   .. attribute:: config
      :type: sphinx.config.Config

       A configuration object.
```

### conf.py

```py
#
# This file only contains a selection of the most common options. For a full
# list see the documentation:
# https://www.sphinx-doc.org/en/master/usage/configuration.html

# -- Path setup --------------------------------------------------------------

#
# import os
# import sys


# -- Project information -----------------------------------------------------

project = 'Foo'
copyright = '2023, picnixz'
author = 'picnixz'


# -- General configuration ---------------------------------------------------

# ones.
extensions = ['sphinx.ext.intersphinx']

# The master toctree document.
master_doc = 'index'

# This pattern also affects html_static_path and html_extra_path.


# -- Options for HTML output -------------------------------------------------

# a list of builtin themes.
#
html_theme = 'alabaster'

intersphinx_mapping = {
    'sphinx': ('https://www.sphinx-doc.org/en/master/', None)
}
```

### Environment Information

```text
Sphinx version:        6.1.3
Docutils version:      0.18.1
Jinja2 version:        3.1.2
Pygments version:      2.14.0
```

### Sphinx extensions

```python
['sphinx.ext.intersphinx']
```

jfbu:
Fix #11274

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -64,6 +64,7 @@ Bugs fixed
   leak to the shadow and border at a page break
 * #11264: LaTeX: missing space before colon after "Voir aussi" for :rst:dir:`seealso`
   directive in French
+* #11274: LaTeX: external links may break PDF build if under ``\sphinxupquote``
 * #11147: Fix source file/line number info in object description content and in
   other uses of ``nested_parse_with_titles``.  Patch by Jeremy Maitin-Shepard.
 * #11192: Restore correct parallel search index building.
--- a/sphinx/texinputs/sphinxlatexliterals.sty
+++ b/sphinx/texinputs/sphinxlatexliterals.sty
@@ -1,7 +1,7 @@
 %% LITERAL BLOCKS
 %
 % change this info string if making any custom modification
-\ProvidesFile{sphinxlatexliterals.sty}[2023/03/19 code-blocks and parsed literals]
+\ProvidesFile{sphinxlatexliterals.sty}[2023/04/01 code-blocks and parsed literals]
 
 % Provides support for this output mark-up from Sphinx latex writer:
 %
@@ -264,7 +264,8 @@
 \def\sphinx@verbatim@nolig@list {\do \`}%
 % Some characters . , ; ? ! / are neither pygmentized nor "tex-escaped".
 % This macro makes them "active" and they will insert potential linebreaks.
-% Not compatible with math mode (cf \sphinxunactivateextras).
+% Not compatible with math mode (cf \sphinxunactivateextras, which uses
+% these lists to make sure activated characters get de-activated).
 \newcommand*\sphinxbreaksbeforeactivelist {}% none
 \newcommand*\sphinxbreaksafteractivelist  {\do\.\do\,\do\;\do\?\do\!\do\/}
 \newcommand*\sphinxbreaksviaactive {%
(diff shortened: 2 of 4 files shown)


jfbu:
I merged this because the Windows CI failure looked (to me) unrelated.

jfbu:
I had seen already the Windows CI failure and it disappeared after pushing dummy commit 1397cb90f.

If I had merged this via the web interface which would have created a new commit perhaps I would not have needed to push again.

ping @AA-Turner I can not understand the Windows CI issue with test_gettext_dont_rebuild_mo but I have seen it now quite a number of times.

PR #11280, opened 2023-04-01
Add ability to "skip" adding marked documentation to the search index

wlach:
### Purpose

Sometimes when writing documentation that are sections that you do not want to be searchable (e.g. at our company we autogenerate some documentation with a header section which is quite generic and not useful when it comes up in search resuls)

### Detail

The following is an example of use:

````rst
.. tip::
   :class: noindex

   Fizzbuzz: should not appear in search index

This text *should* appear in the search index.
````

Works great. It seems in keeping with the general spirit of [Docutils classes](https://docutils.sourceforge.io/docs/ref/doctree.html#classes) from what I can tell.


AA-Turner:
@wlach this seems reasonable. I've taken the liberty to rename to `no-search` to match the file-wide metadata option.

Please could you add a CHANGES entry and a test? Ideally we would also document this somewhere.

A

wlach:
Thanks @AA-Turner, I'll try to find some time to update this in the next couple weeks.

AA-Turner:
@wlach quick ping in case you have time to look at this before 8.2.

A

wlach:
> @wlach quick ping in case you have time to look at this before 8.2.
> 
> A

Thanks for the ping. I did what you suggested. I thought of maybe adding something to the docs too, but I couldn't find an obvious place to put it. Should we maybe add a section on "search" at some point?

xkhelm (approved):
thanks

AA-Turner (approved):

AA-Turner:
> Should we maybe add a section on "search" at some point?

I now realise we have no documentation for the HTML search feature! We should indeed add such a page.

A

Issue #6279, opened 2019-04-12
Not possible to produce math as SVG for HTML via xelatex


jfbu:
I tried to use `xelatex --no-pdf` as latex command for handling math equations, as [dvisvgm](https://dvisvgm.de/) can convert the `.xdv` file into SVG (similarly as it does with `.dvi` file from a `latex` run).

But with

```
extensions = [
    'sphinx.ext.imgmath',
]

imgmath_image_format = 'svg'

imgmath_latex = 'xelatex'

imgmath_latex_args = ['--no-pdf']
```

I get on make html the error 

```
Exception occurred:
  File "/opt/miniconda3/lib/python3.6/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/4y/h4wblhdx08q0j910d0rkq3m80000gn/T/tmp356x6ljh/math.svg'
```

This is caused (surely) by the fact that `sphinx.ext.imgmath` only supports `.dvi` not `.xdv` as intermediate file extension from latex run.

But `xelatex` is useful to use OpenType Math fonts (although there are few of them). This could help for example use same fonts as MathJaX does.

Note: even after adding support for `.xdv` one should think about correct header for xelatex preamble.

**Environment info**
- OS: Mac
- Python version: 3.6.8
- Sphinx version: 2.0.0
- Sphinx extensions:  sphinx.ext.imgmath
- Extra tools: TL2019, dvisvgm 2.6.2

**Additional context**

- #5301, #6278

Note that using ``imgmath_latex = 'dvilualatex'`` does work. **edit** but then dvisvgm may not succeed into embedding the math fonts (see #6280)

tk0miya:
I think following patch will help xelatex + SVG case.
```
diff --git a/sphinx/ext/imgmath.py b/sphinx/ext/imgmath.py
index eb0d35c25..5bff8def3 100644
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -134,7 +134,10 @@ def compile_math(latex, builder):

     try:
         subprocess.run(command, stdout=PIPE, stderr=PIPE, cwd=tempdir, check=True)
+        if builder.config.imgmath_latex == 'xelatex':
+            return path.join(tempdir, 'math.xdv')
+        else:
-        return path.join(tempdir, 'math.dvi')
+            return path.join(tempdir, 'math.dvi')
[...]

Issue #6280, opened 2019-04-12
For math into SVG, allow PDF as intermediate format (as dvisvgm can handle it)


jfbu:
**UPDATE on April 2, 2023**:

With my current installation (TeXLive 2023, dvisvgm 3.0.3 built with ghostscript support) the problems mentioned below with executing `dvilualatex + dvisvgm` with LaTeX file configured to use `unicode-math` and XITS or other OpenType Math font seem to have disappeared.

So it is even less urgent to support PDF as intermediate image type.

<hr>
<hr>

This is related to #5301 and #6279.

I tried using `lualatex` with `XITS` fonts. I tested with this latex file `temp.tex`

```
\documentclass{article}
\usepackage{unicode-math}\setmainfont{XITS}\setmathfont{XITS Math}
\begin{document}
$F^{\mu\nu} = \partial^\mu A^\nu - \partial^\nu A^\mu$
\end{document}
```

and executed `dvilualatex` and then `dvisvgm` (with or without `--no-fonts`). But the `dvisvgm` currently fails:

```
kpathsea:make_tex: Invalid filename `XITSMath:mode=base;script=math;language=DFLT;+ssty=0;.mf', contains ':'
WARNING: font file 'XITSMath:mode=base;script=math;language=DFLT;+ssty=0;.mf' not found
kpathsea:make_tex: Invalid filename `XITSMath:mode=base;script=math;language=DFLT;.mf', contains ':'
WARNING: font file 'XITSMath:mode=base;script=math;language=DFLT;.mf' not found
kpathsea:make_tex: Invalid filename `XITS:mode=node;script=latn;language=DFLT;+tlig;.mf', contains ':'
WARNING: font file 'XITS:mode=node;script=latn;language=DFLT;+tlig;.mf' not found
pre-processing DVI file (format version 2)
processing page 1
  graphic size: 164.198365pt x 580.858197pt (57.709125mm x 204.148308mm)
  WARNING: can't embed font 'XITS:mode=node;script=latn;language=DFLT;+tlig;'
  WARNING: can't embed font 'XITSMath:mode=base;script=math;language=DFLT;+ssty=0;'
  WARNING: can't embed font 'XITSMath:mode=base;script=math;language=DFLT;'
  output written to temp.svg
1 of 1 page converted in 0.244583 seconds
```

and produces this svg file

```
<?xml version='1.0' encoding='UTF-8'?>
<!-- This file was generated by dvisvgm 2.6.2 -->
[...]

PR #11281, opened 2023-04-01
support for imgmath_latex=tectonic

drdv:
Subject: support use of tectonic in imgmath

### Feature or Bugfix
- Feature

### Purpose
- support use of tectonic and xelatex in imgmath

### Detail
Currently, using
```
imgmath_latex = "tectonic"
imgmath_latex_args = ["--outfmt", "xdv"]
```
is not possible. This PR is related to #6279

### DIFF
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -130,9 +130,10 @@ def compile_math(latex: str, builder: Builder) -> str:
     # build latex command; old versions of latex don't have the
     # --output-directory option, so we have to manually chdir to the
     # temp dir to run it.
+    command = [builder.config.imgmath_latex]
     # assume that executable is on the PATH (i.e., don't handle /path/to/tectonic)
     if builder.config.imgmath_latex not in ['tectonic']:
-        command = [builder.config.imgmath_latex, '--interaction=nonstopmode']
+        command.append('--interaction=nonstopmode')
     # add custom args from the config file
     command.extend(builder.config.imgmath_latex_args)
     command.append('math.tex')
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -237,9 +237,7 @@ def render_math(
     latex = generate_latex_macro(image_format,
                                  math,
                                  self.builder.config,
-                                 self.builder.confdir,
-                                 self.builder.config.imgmath_latex in ['xelatex',
-                                                                       'tectonic'])
+                                 self.builder.confdir)
 
     filename = f"{sha1(latex.encode()).hexdigest()}.{image_format}"
     generated_path = path.join(self.builder.outdir, self.builder.imagedir, 'math', filename)
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -127,12 +127,13 @@ def compile_math(latex: str, builder: Builder) -> str:
     with open(filename, 'w', encoding='utf-8') as f:
         f.write(latex)
 
+    imgmath_latex_name = path.basename(builder.config.imgmath_latex)
+
     # build latex command; old versions of latex don't have the
     # --output-directory option, so we have to manually chdir to the
     # temp dir to run it.
     command = [builder.config.imgmath_latex]
-    # assume that executable is on the PATH (i.e., don't handle /path/to/tectonic)
-    if builder.config.imgmath_latex not in ['tectonic']:
+    if imgmath_latex_name not in ['tectonic']:
         command.append('--interaction=nonstopmode')
     # add custom args from the config file
     command.extend(builder.config.imgmath_latex_args)
@@ -141,8 +142,7 @@ def compile_math(latex: str, builder: Builder) -> str:
     try:
         subprocess.run(command, capture_output=True, cwd=tempdir, check=True,
                        encoding='ascii')
-        # FIXME: maybe it would be better to add a param: imgmath_dvi_extension?
-        if builder.config.imgmath_latex in ['xelatex', 'tectonic']:
+        if imgmath_latex_name in ['xelatex', 'tectonic']:
             return path.join(tempdir, 'math.xdv')
         else:
             return path.join(tempdir, 'math.dvi')
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -80,13 +80,17 @@ def write_svg_depth(filename: str, depth: int) -> None:
 
 
 def generate_latex_macro(image_format: str,
-                         math: str, config: Config, confdir: str = '') -> str:
+                         math: str,
+                         config: Config,
+                         confdir: str = '',
+                         add_dvips_option: bool = False) -> str:
     """Generate LaTeX macro."""
     variables = {
         'fontsize': config.imgmath_font_size,
         'baselineskip': int(round(config.imgmath_font_size * 1.2)),
         'preamble': config.imgmath_latex_preamble,
         'tightpage': '' if image_format == 'png' else ',tightpage',
+        'dvips': '' if not add_dvips_option else ',dvips',
         'math': math,
     }
 
@@ -126,7 +130,9 @@ def compile_math(latex: str, builder: Builder) -> str:
     # build latex command; old versions of latex don't have the
     # --output-directory option, so we have to manually chdir to the
     # temp dir to run it.
-    command = [builder.config.imgmath_latex, '--interaction=nonstopmode']
+    # assume that executable is on the PATH (i.e., don't handle /path/to/tectonic)
+    if builder.config.imgmath_latex not in ['tectonic']:
+        command = [builder.config.imgmath_latex, '--interaction=nonstopmode']
     # add custom args from the config file
     command.extend(builder.config.imgmath_latex_args)
     command.append('math.tex')
@@ -134,7 +140,11 @@ def compile_math(latex: str, builder: Builder) -> str:
     try:
         subprocess.run(command, capture_output=True, cwd=tempdir, check=True,
                        encoding='ascii')
-        return path.join(tempdir, 'math.dvi')
+        # FIXME: maybe it would be better to add a param: imgmath_dvi_extension?
+        if builder.config.imgmath_latex in ['xelatex', 'tectonic']:
+            return path.join(tempdir, 'math.xdv')
+        else:
+            return path.join(tempdir, 'math.dvi')
     except OSError as exc:
         logger.warning(__('LaTeX command %r cannot be run (needed for math '
                           'display), check the imgmath_latex setting'),
@@ -226,7 +236,9 @@ def render_math(
     latex = generate_latex_macro(image_format,
                                  math,
                                  self.builder.config,
-                                 self.builder.confdir)
+                                 self.builder.confdir,
+                                 self.builder.config.imgmath_latex in ['xelatex',
+                                                                       'tectonic'])
 
     filename = f"{sha1(latex.encode()).hexdigest()}.{image_format}"
     generated_path = path.join(self.builder.outdir, self.builder.imagedir, 'math', filename)
--- a/sphinx/ext/imgmath.py
+++ b/sphinx/ext/imgmath.py
@@ -82,15 +82,15 @@ def write_svg_depth(filename: str, depth: int) -> None:
 def generate_latex_macro(image_format: str,
                          math: str,
                          config: Config,
-                         confdir: str = '',
-                         add_dvips_option: bool = False) -> str:
+                         confdir: str = '') -> str:
     """Generate LaTeX macro."""
     variables = {
         'fontsize': config.imgmath_font_size,
         'baselineskip': int(round(config.imgmath_font_size * 1.2)),
         'preamble': config.imgmath_latex_preamble,
-        'tightpage': '' if image_format == 'png' else ',tightpage',
-        'dvips': '' if not add_dvips_option else ',dvips',
+        # the dvips option is important when imgmath_latex in ["xelatex", "tectonic"],
+        # it has no impact when imgmath_latex="latex"
+        'tightpage': '' if image_format == 'png' else ',dvips,tightpage',
         'math': math,
     }
 
--- a/sphinx/templates/imgmath/preview.tex_t
+++ b/sphinx/templates/imgmath/preview.tex_t
@@ -9,7 +9,7 @@
 \pagestyle{empty}
 <%= preamble %>
 
-\usepackage[active<%= tightpage %><%= dvips %>]{preview}
+\usepackage[active<%= tightpage %>]{preview}
 
 \begin{document}
 \begin{preview}
--- a/sphinx/templates/imgmath/preview.tex_t
+++ b/sphinx/templates/imgmath/preview.tex_t
@@ -9,7 +9,7 @@
 \pagestyle{empty}
 <%= preamble %>
 
-\usepackage[active<%= tightpage %>]{preview}
+\usepackage[active<%= tightpage %><%= dvips %>]{preview}
 
 \begin{document}
 \begin{preview}


jfbu (changes requested):
Can you explain in particular why he addition of `dvips` option to `preview` latex package call?  Apart from that your PR indeed will allow using `xelatex` but extra configuration in conf.py must be added

```
imgmath_latex = 'xelatex'
imgmath_latex_args = ['-no-pdf']
imgmath_image_format = 'svg'
imgmath_use_preview = True # to get correct alignment with surrounding inline text
```

thus documentation needs to be updated.  I guess motivation for `xelatex` is to use `unicode-math` and unicode fonts with a Math table?
But one can already use `lualatex` for that and it will produce `dvi` file so does not need update to `imgmath.py` and works already.  But I don't object to support `xelatex/tectonic`.

I don't have [`tectonic`](https://tectonic-typesetting.github.io/en-US/) but it seems it is based on XeTeX engine so perhaps it has same option `-no-pdf` to produce `xdv` file? can you please point to where one finds this in its documentation? (I supposed you tested your patch so it must be working...)

Why added preview package `dvips` option?  The preview package documentation says:
```
dvips If this option is specified as a class option or to other packages, several packages pass things like page size information to Dvips, or cause crop marks or draft messages written on pages. This seriously hampers the usability of previews. If this option is specified, the changes will be undone if possible.
```
i.e. this is for the `dvips + ps2pdf` tool chain or similar postprocessing.  

Actually I looked at the code and this option triggers insertion in the `.dvi` of PostScript code.  I am not sure (and doubt a lot) whether `dvipng` or `dvisvgm` do anything with it.

Thus this setting appears irrelevant and even possibly prejudiciable although my limited testing showed no effect.  Looking more at `preview` code, I see that this option combined with `tightpage` (triggered from using  `svg` which is only possiblity with `xelatex` because `dvipng` will not digest `.xdv` but only `.dvi`) prevents it from executing the code devoted to the `xelatex` case (`preview` has an option `xetex` but it is executed automatically normally when used with `xelatex` engine).

My testing was very limited and I found no real ill-effect of this `dvips`, but it seems to be not motivated, please explain why you want to add this. `Tectonic` appears to be based on `XeTeX` code combined with other things at engine level, so I don't expect any reason why `dvips` should be employed with it.

`preview` originates in 2001 and was last updated in 2017, and this `dvips` option must have been coded surely close to 20 years ago.  As I said its effect is to insert PostScript code in the output, and it also prevents some special handling of XeTeX engine by `preview`.

drdv:
Thanks for the comments.

1. My main motivation is to be able to use `tectonic`. I added support for `xelatex` as well simply because `tectonic` is based on it and it seems logical to allow using it directly. 
2. `conf.py` should include
```python
imgmath_latex = "tectonic"
imgmath_latex_args = ["--outfmt", "xdv"]

# imgmath_latex = "xelatex"
# imgmath_latex_args = ["-no-pdf"]

imgmath_image_format = "svg"
imgmath_use_preview = True
```
One of the options in `tectonic -h` is
```
--outfmt <format> The kind of output to generate [default: pdf]  
    [possible values: pdf, html, xdv, aux, fmt] 
```
3. The reason for the `dvips` option of the `preview` package is explained at https://github.com/mgieseki/dvisvgm/issues/236#issuecomment-1493119335 (as you can see in my question in this issue, I have already tried using `\usepackage[active,tightpage,xetex]{preview}` but it doesn't work). To summarize: this is necessary in order for `dvisvgm` to have access to `depth` so that vertical alignment becomes possible. 
4. Indeed we need to update the documentation (if this PR seems acceptable). The current version is somewhat a minimal demonstration of the feature because I wasnt's sure whether it goes in the right direction (and it didn't make much sense to start changing many things). In addition to documentation, a unit test has to be added, also probably there is a better way to organize this feature. Namely, probably an additional configuration variable is needed (`imgmath_dvi_extension` - see `FIXME` in the PR) - but I was not sure whether this is acceptable. If all this sounds good, I can modify the code. 
5. I have tested the feature on `fedora 37` and it requires installing `dvisvgm`, `tectonic` and `texlive-fontspec`. The latter seems kind of redundant as required fonts are surely downloaded by `tectonic`, but for the moment I don't see an alternative (see: https://github.com/mgieseki/dvisvgm/issues/237#issuecomment-1493284068)

jfbu:
> The reason for the dvips option of the preview package is explained at xetex XDV depth information  mgieseki/dvisvgm#236 (comment) (as you can see in my question in this issue, I have already tried using \usepackage[active,tightpage,xetex]{preview} but it doesn't work). To summarize: this is necessary in order for dvisvgm to have access to depth so that vertical alignment becomes possible.

Thanks! However Sphinx has special built-in code for this and thus the advice by `dvisvgm` author did not know about this.

https://github.com/sphinx-doc/sphinx/blob/6809ca7d04da47f8780647296961934cdc07e14a/sphinx/ext/imgmath.py#L62-L79


 https://github.com/sphinx-doc/sphinx/blob/6809ca7d04da47f8780647296961934cdc07e14a/sphinx/ext/imgmath.py#L184-L202

In my brief testing the alignment of my inlined math snippets with surrounding text in paragraph looked ok, but I spend most of my time checking other things related to `dvips` option in `preview` source code.

I take this opportunity to add the following information with `lualatex`, although not related to your goal to support `tectonic` toolchain.  In my #6280 from four years ago I reported that using `dvilualatex` produces a `dvi` which `dvisvgm` could not handle properly when OpenType Math fonts were used.  But on testing now with `dvisvgm 3.0.3` it seems to work ok. So I updated #6280, it is possible to use
```
imgmath_latex = 'dvilualatex'
imgmath_latex_preamble = r'''
   \usepackage{unicode-math}
   \setmainfont{XITS}
   \setmathfont{XITS Math}
'''
imgmath_image_format = 'svg'
imgmath_use_preview = True
```
It is quite slow though.

The PR is acceptable of course, but please provide Sphinx example of some input which will cause output not to have correct alignment and how `dvips` fixed that.  I am  100% sure that we should not use `dvips`.

jfbu:
Here is an example where I used your PR but removed `,dvips`:
Capture d’écran 2023-04-02 à 13 18 29

using in conf.py
```
extensions = ['sphinx.ext.imgmath']

imgmath_latex = 'xelatex'
imgmath_latex_args = ['-no-pdf']
# imgmath_latex = 'dvilualatex'
imgmath_latex_preamble = r'''
   \usepackage{unicode-math}
   \setmainfont{XITS}
   \setmathfont{XITS Math}
'''
imgmath_image_format = 'svg'
imgmath_use_preview = True
```
to test OpenType Math fonts.  This is very slow though.  And `index.rst` contains
```rest
q :math:`E = pq^2` p Hello :math:`(p^n + q^n = y^n)` math  ppp :math:`F^{\mu\nu} = \partial^\mu A^\nu - \partial^\nu A^\mu`
```

The aligment with surrounding text looks ok.  Of course there is cleary here a fontsize problem...

jfbu:
@drdv I will examine that later better. I possibly have gotten confused at some point and maybe you are right about `dvips`.

jfbu (approved):

jfbu:
@drdv Sorry about waste of  your time with the `dvips` thing. To remove my request for changes I needed to "approve", but I will need to make a proper review later.

As per documentation I can take care of it at later stage.  For example can take this opportunuty to add the `dvilualatex` info.  Some years ago, it did not work.  I don't know since when `dvisvgm` is able to handle the `.dvi` using references to system OpenType fonts, but it was not in 2019 when I last checked.  Now it seems to be ok.

sphinx/ext/imgmath.py
```diff
@@ -126,15 +130,22 @@ def compile_math(latex: str, builder: Builder) -> str:
     # build latex command; old versions of latex don't have the
     # --output-directory option, so we have to manually chdir to the
     # temp dir to run it.
-    command = [builder.config.imgmath_latex, '--interaction=nonstopmode']
+    command = [builder.config.imgmath_latex]
+    # assume that executable is on the PATH (i.e., don't handle /path/to/tectonic)
+    if builder.config.imgmath_latex not in ['tectonic']:
```
jfbu: maybe use `path.basename(build.config.imgmath_latex)`.  As this is again used below maybe we need a variable to store the user command. The `imgmath_latex` is [documented](https://www.sphinx-doc.org/en/master/usage/extensions/math.html#confval-imgmath_latex) to have to be either a plain command name (without options) or a full path.

sphinx/ext/imgmath.py
```diff
@@ -126,15 +130,22 @@ def compile_math(latex: str, builder: Builder) -> str:
...
+        command.append('--interaction=nonstopmode')
     # add custom args from the config file
     command.extend(builder.config.imgmath_latex_args)
     command.append('math.tex')
 
     try:
         subprocess.run(command, capture_output=True, cwd=tempdir, check=True,
                        encoding='ascii')
-        return path.join(tempdir, 'math.dvi')
+        # FIXME: maybe it would be better to add a param: imgmath_dvi_extension?
```
jfbu: About adding a param, it would have made sense if we were to allow `pdf`.  Indeed in the past we tested `dvisvgm --pdf` as alternative.  But here it is only a matter of `dvi` versus `xdv` and no engine is compatible with both: the latex command forces that either one of the two  works.  So, I don't think it needed to add a parameter.

sphinx/ext/imgmath.py
```diff
@@ -126,15 +130,22 @@ def compile_math(latex: str, builder: Builder) -> str:
...
+        if builder.config.imgmath_latex in ['xelatex', 'tectonic']:
```
jfbu: See above about `path.basename()`.

sphinx/ext/imgmath.py
```diff
@@ -226,7 +237,9 @@ def render_math(
     latex = generate_latex_macro(image_format,
                                  math,
                                  self.builder.config,
-                                 self.builder.confdir)
+                                 self.builder.confdir,
```
jfbu: As `dvips` is the default state of `preview` when it detects that a `dvi` file is to be produced, and as it appears that `dvips` is needed work-around for usage with `xelatex` to let `preview` add PostScript information into the `dvi/xdv` that `dvisvgm` relies upon to recover the box dimension information, in particular the `depth`, I believe we can actually use always `dvips` here.  At same time, if we do that we make more difficult supporting a `pdf` intermediate file type, but for reasons at #6280 and for matters of evolving `dvisvgm` support of PDF, supporting only `dvi` and `xdv` looks reasonable.

sphinx/templates/imgmath/preview.tex_t
```diff
@@ -9,7 +9,7 @@
 \pagestyle{empty}
 <%= preamble %>
 
-\usepackage[active<%= tightpage %>]{preview}
+\usepackage[active<%= tightpage %><%= dvips %>]{preview}
```
jfbu: It could be ``\usepackage[active,dvips<%=tightpage %>]{preview}`` as argued above.

sphinx/ext/imgmath.py
jfbu: > As `dvips` is the default state of `preview` when it detects that a `dvi` file is to be produced,

Under the `tightpage` option.  This comes from this code in `prtightpage.def` file:
```
\ifnum\pr@graphicstype=\z@
  \ifcase
    \ifx\XeTeXversion\@undefined
      \ifx\pdfoutput\@undefined \@ne\fi
      \ifx\pdfoutput\relax \@ne\fi
      \ifnum\pdfoutput>\z@ \tw@\fi \@ne
    \else \thr@@\fi
  \or \ExecuteOptions{dvips}\relax
  \or \ExecuteOptions{pdftex}\relax
  \or \ExecuteOptions{xetex}\relax\fi\fi
```
I do need to check that `dvips` is not detrimental to `dvipng`.

jfbu: We do need to change the regex for depth with `png` if using `dvips` option.

Without:
```
$ dvipng -o ceaa1dfa0f274a525b952117e67d0cfd2de7d5ed.png -T tight -z9 -gamma 1.5 -D 110 -bg Transparent --depth math.dvi
This is dvipng 1.17 Copyright 2002-2015, 2019 Jan-Ake Larsson
[1 depth=1] 
```


With:
```
tmp0tx70t_4$ dvipng -o ceaa1dfa0f274a525b952117e67d0cfd2de7d5ed.png -T tight -z9 -gamma 1.5 -D 110 -bg Transparent --depth math.dvi
This is dvipng 1.17 Copyright 2002-2015, 2019 Jan-Ake Larsson
[1 (preview-latex version 13.1) depth=1] 
```

```diff
-depth_re = re.compile(r'\[\d+ depth=(-?\d+)\]')
+depth_re = re.compile(r'\[.* depth=(-?\d+)\]')
```

allows the depth information to be recovered if `dvipng` and `preview` was used with `dvips` option.

jfbu: In my testing the `png` files are identical whether or not using `dvips` option.  The only difference is that console output of `dvipng` is slightly changed as above.

So I think it is ok we use `dvips` always with `preview`.

I understand this is peripheral to your `tectonic` support objective, and that it may not be good style to introduce such a change on this occasion, but this looks to me like simplfying measure.

jfbu: Maybe we can use slightly more strict regex
```diff
-depth_re = re.compile(r'\[\d+ depth=(-?\d+)\]')
+depth_re = re.compile(r'\[\d.* depth=(-?\d+)\]')
```
but this does not look fragile anyhow.

sphinx/templates/imgmath/preview.tex_t
jfbu: I have made quite some testing, and `dvips` option seems to have nil impact on the `dvipng` output.  So we can have it here always because:

- it does not (in my testing) change anything when image format is `'png'`.
- it is already implied by `tightpage` option with standard `latex`,
- with `xelatex`, it is needed in addition to `tightpage` else `dvisvgm` can not report the found depth

The `tightpage` is needed when format is `'svg'` else `dvisvgm` can not report an image depth.  With standard `latex`(not `pdflatex` or `xelatex`)  the `tightpage` option of `preview` automatically activates the `dvips` option, and `dvisvgm` can extra depth information.

So we could almost use always `dvips,tightpage` except that in my testing `tightpage` with `'png'` image format causes slight modification compared to not using it:  the depth is different (it seems by 1pt) and the picture width also it seems, slightly.

Thus I propose we keep `tightpage` conditional on image format being `'svg'` but we can use `dvips` always: does not change anything to the `.png` output from `dvipng`, is already tacitly implied if image format is `'svg'` and using `latex`, is needed in this PR if using `xelatex/tectonic`. 

What we can do as alternative to hardcode  `dvips` always here, is to include it in the  `tightpage` variable. This also will avoid adding a  `dvips` variable.

See [comment](https://github.com/sphinx-doc/sphinx/pull/11281#discussion_r1155338756)

sphinx/ext/imgmath.py
jfbu:
```diff
-        'tightpage': '' if image_format == 'png' else ',tightpage',
+        'tightpage': '' if image_format == 'png' else ',dvips,tightpage',
```
and don't add a `'dvips'` variable.

So we keep in this PR the `'png'` case unchanged although I tested that using `dvips` with it changed nothing, but it would be needed to modify `depth_re` regex as above.  Better to not introduce such a change in this unrelated PR.

sphinx/ext/imgmath.py
drdv: Done. Personally I would have preferred to use 
```python
from pathlib import Path
imgmath_latex_name = Path(builder.config.imgmath_latex).name
# ...
```
but I prefer to be consistent with the rest of the code.

sphinx/ext/imgmath.py
drdv: Indeed.

sphinx/ext/imgmath.py
drdv: Done.

sphinx/ext/imgmath.py
drdv: I am a bit confused. My general impression is that you advocate to always include `dvips` as an option to the `preview` package, but in your last comment above, it seems to me that, you argue against it (maybe I am missing something). 

My two cents: I have verified empirically that your conclusions about the relation between the `dvips` option and `png` files is correct. I don't see the change of `depth_re` as very intrusive (even though it is not directly related to adding `tectonic`). But there is something else that bothers me - it is the intent. If we always add `dvips` as an option, then we have to complement this with a ton of comments (useful in such and such case, not useful in other cases, ...). While if we are explicit in the code and add it only where we use it, then it would be mostly clear for a person reading the code. 

Another thing. In the documentation of the `dvips` option (in the `preview` package) I read: "... This seriously hampers the usability of previews ...". While I am not entirely sure what this means, it seems to me that it is safer to omit this option whenever necessary (but I might be wrong). 

So I didn't change this in the current commit. But if you think that we should hard-code it after having read my comments above, then I would agree. In such a case, instead of

```python
\usepackage[active,dvips<%= tightpage %>]{preview}
```
I would opt for 
```python
\usepackage[active,<%= tightpage %>]{preview}
```
and adding `dvips,tightpage` upon instantiation of the template (as you suggest towards the end of your remarks). I prefer the latter option because, at least it excludes the `PNG` case.

sphinx/templates/imgmath/preview.tex_t
drdv: I have already addressed these issues above. Here I summarize the cases:

# Cases

Note: `xelatex` covers the case of `tectonic` as well (with only one difference: the `--interaction` parameter).

## PNG

### latex

+ `tightpage` is not used
+ `dvips` option has no impact on PNG files (empirically verified)
+ `dvips` option results in a different `stdout` from `dvipng` (with the standard options) and requires a slight tweaking of `depth_re`

### xelatex

+ cannot be used
+ `xelatex` outputs XDV file (i.e., DVI version 7) and `dvipng` cannot use it

## SVG

### latex

+ `tightpage` is used
+ `dvips` has no impact on SVG files (empirically verified - I didn't understand why it is implied by `tightpage` in the code snippet from `prtightpage.def`: my ability to read such latex code is limited)

### xelatex

+ `tightpage` is used
+ `dvips` is used (otherwise depth information is not displayed by `dvisvgm`)

drdv: @jfbu : You mentioned `luatex` and some other considerations about `pdf` etc. Could we maybe draw a diagram (e.g., with graphviz dot) of all the cases of interest? This would be good as documentation and might help reorganize the code (in the future), for example having e.g., a `class ImgmathProcessingBase` which is extended later: `class ImgmathLatex: ImgmathProcessingBase`, `class ImgmathXelatex: ImgmathProcessingBase` ...

sphinx/ext/imgmath.py
jfbu: > I am a bit confused. My general impression is that you advocate to always include `dvips` as an option to the `preview` package, but in your last comment above, it seems to me that, you argue against it (maybe I am missing something).

Sorry about my very verbose comments.  I was initially vehemenly against `dvips` then after having understood it was required for `dvisvgm` to communicate to Sphinx the depth (as `dvisvgm` only looks for postscript specials, which `preview` writes only under `dvips` option), I started considering using it always for simplicity sake. The PostScript specials are ignored by `dvipng` (empirical).  With `latex`, the `tightpage` option automatically also triggers `dvips` option.  This is confirmed by this extract of the `preview` documentation on top of page 5:
> If neither dvips or pdftex have been specified, the corresponding option will get autodetected and invoked.

We must force it with `xetex` because `preview` in that case will not do `dvips` option.

About
> Another thing. In the documentation of the dvips option (in the preview package) I read: "... This seriously hampers the usability of previews ...". While I am not entirely sure what this means, it seems to me that it is safer to omit this option whenever necessary (but I might be wrong).

`preview` author refers to `dvips` used as class option or as option of some packages adding crop marks for example: the point is that `previews`'s `dvips` option is configured to counteract.  In LaTeX, a class option is also presented to packages.  So if `dvips` is used as class option, `preview` receives it too, and it does its counter-measures. I don't read PostScript so I can't be sure but it seems to be related to page dimensions. 

The whole point is that `dvisvgm` has a specific mechanism tied to recover the PostScript which `preview` adds to `dvi` (or `xdv`) under `tightpage` option.  But this mechanism is by default inihibited by `preview` when it is used with XeTeX (as you discovered, or with PDFTeX if used in PDF mode -- "latex" command is PDFTeX used in DVI mode, "pdflatex" Is PDFTeX used in PDF mode).

> I prefer the latter option because, at least it excludes the PNG case.

I agree.  Although `dvips` option seems not have any effect on `png` case, it is better not to introduce any change in a patch  which is about allowing `xelatex` as command and only regards `svg` case.

sphinx/templates/imgmath/preview.tex_t
jfbu: Your summary appears correct to me; and as we have discussed `dvips` is automatically triggered by `preview` if it has received `tightpage` option and is used with `latex`.  So using `dvips,tightpage` for `'svg'` is no change at all to `latex`.  It would be a change to `pdflatex` but `imgmath` has not so far supported `'pdf'` as intermediate image format.  And `dvips,tightpage` is required if we want `xelatex` to work here.

As per the `luatex`.  What happens is that I considered so far main reason for perhaps not using `latex` but one of `xelatex/lualatex` is for matters of math fonts.  Thanks to you, I realized today that one can actually use already `dvilualatex` as a replacement of `latex` for the `imgmath_latex` command, also with `'svg'`.  It was not the case some years ago if one wanted `'svg'` for image format.  It works today with current `dvisvgm`.  So the user can, using `dvilualatex` add a custom usage of OpenMath fonts, which `latex` can not use, only `lualatex` or `xelatex`.

So here I saw only need for perhaps some addition to document, rather than adding classes.  The existing code seemed generic enough (apart from not handling `xdv` but this is a bit normal as `dvipng` can not be used with it).  Also I am a bit prejudiced against XeTeX in matters of TeX math: it is known to have a number of bugs in the math rendering which have not been fixed for years.  Now if in future the `tectonic` project achieves a rewrite of XeTeX in Rust perhaps it will become attractive for speed... 

The priority-one thing would be to find a way to execute latex or xelatex or dvilualatex only once, on one file containing all math snippets.  Both of `dvipng` and `dvisvgm` can handle multiple pages dvi.  The speed gain would be tremendous. I did a test with only 100 equations and it takes (on my slow old laptop, divide by 5 to 10 on current ones) about 50 seconds.  [Caching the LaTeX preamble](https://github.com/jfbu/sphinx/tree/imgmath_cache_latexfmt) I could achieve about 10% gain.  This is not enough gain. And anyhow caching the latex preamble in a dumped format is impossible with `lualatex/xelatex` if OpenType fonts are used. 

I have some ideas about how to proceed but now I will not have the time for it in next few weeks probably (I said that here already close to a month ago, but now this is for real...)

jfbu (approved):
LGTM with nits: as per our discusion I suggest you suppress the added argument to `generate_latex_macro`, do not  modify the `preview.tex_t` and simply do 

```diff
-        'tightpage': '' if image_format == 'png' else ',tightpage',
+        'tightpage': '' if image_format == 'png' else ',dvips,tightpage',
```

drdv:
Done. I added as well a comment about the usefulness of `',dvips'`.

jfbu (commented):
LGTM,.

Only follow-up I see here is that some complicated math may need multipass, and our document says about this:
> Some fancy LaTeX mark-up (an example was reported which used TikZ to add various decorations to the equation) require multiple runs of the LaTeX executable. To handle this, set this configuration setting to 'latexmk' (or a full path to it) as this Perl script reliably chooses dynamically how many latex runs are needed.

So I considered perhaps at some point we want to support a `latexmk + xelatex` workflow.  But, although I know `latexmk` already, I spend fifteen minutes trying to let it work to stop at `.xdv` file and all my attempts such as ``latexmk -xdv -dvi- -latex="xelatex -no-pdf" foo`` fail; this one because  ``the extensions differ between the actual output file 'foo.xdv', and the expected output 'foo.dvi'``.  I will have to investigate more. This is ironical because `latexmk` does all its passes with `.xdv` format as output for `xelatex`.  The only thing is that I can't get it to stop before producing the `pdf`.

But `latexmk -xelatex` will produce anyhow the `.xdv` (and the `pdf` which is the step I wanted to kill, but let's forget that).

Our test for selecting `.xdv` extension can not accomodate command being `latexmk -xelatex`.  We could imagine looking for `-xelatex` in the command arguments but there are (quite a few) additional latexmk syntaxes possible.

So indeed perhaps at some point adding a `imgmath_latex_output_extension` as you considered under another name could be considered.

For time being I will simply merge this.

Thanks for contribution.  I will add documentation later.

jfbu:
> 2\. add unit tests: I have to see how to use tectonic (probably in a docker) ... suggestions are welcome

Our test suite is already too much time-consuming,  in particular on Windows, and LaTeX builds are part of this.
The `tectonic` thing looks like a more long term subject, we have not received yet PRs or feature requests of supportin git as a new engine.  I wonder how you run it on your Sphinx projects generally speaking.  Overriding the Makefile? There is no objection of principle, but personnaly for example I would consider higher priority to support  `latex+dvipdfmx` in place of `pdflatex` because it produces significantly smaller files (like `xelatex` which also uses `(x)dvipdfmx` as pdf-producer, which compresses fonts more than `pdflatex`.)

drdv:
Thanks for updating the documentation (and for the discussion/help).

> I wonder how you run it on your Sphinx projects generally speaking. Overriding the Makefile?

Currently I use `tectonic` only in the context of `imgmath` so there is no need to do anything other than setting some `imgmath` variables. Using `tectonic` as an engine for generating `pdf` instead of `html` is something different and I haven't had the need to do it. 

I was not aware of `latexmk` (I will try to test a bit).

PR #11283, opened 2023-04-03
Fix translator class check in ``SigElementFallbackTransform``

### LINKED ISSUE #11272: [BUG] SigElementFallbackTransform incorrectly checks builder's translator class.
### Describe the bug

In the `SigElementFallbackTransform` class, the following lines are wrong:

https://github.com/sphinx-doc/sphinx/blob/ba080286b06cb9e0cadec59a6cf1f96aa11aef5a/sphinx/transforms/post_transforms/__init__.py#L239-L242

I would expect the use of `issubclass` instead of `isinstance` since `translator` is a class and not a translator instance. Additionally, the comment says that subclasses of `SphinxTranslator` support `desc_sig_*` nodes but what they actually do is supporting the `docutils.nodes.inline` node because `SigElementFallbackTransform` converts those `desc_sig_*` nodes to `docutils.nodes.inline`. 

Due to the incorrect check on the custom translator, I am not allowed to have a custom translator handling `desc_sig_*` nodes. In addition, due to:

https://github.com/sphinx-doc/sphinx/blob/ba080286b06cb9e0cadec59a6cf1f96aa11aef5a/sphinx/transforms/post_transforms/__init__.py#L244-L246

custom translator classes *must* implement *all* handlers for `desc_sig_*` nodes explicitly. For instance, I cannot just handle `desc_sig_space` differently from the other `desc_sig_*` using a visitor pattern and must instead test *inside* `visit_inline` whether the node "looks" like a `desc_sig_*` (yes, because we cannot even know that `desc_sig_*` was transformed into a `docutils.nodes.inline` and that it was not a "true" inline node).

I would suggest adding an information in the inline node generated by `SigElementFallbackTransform.fallback` in order to know that this inline node was actually a `desc_sig_*` node (same for `desc_inline` being transformed into an inline node).

### index.rst

```rst
Foo
===

.. py:function:: foo(a: int, b: int)

   The foo function.
```

### conf.py

```python
project = 'Foo'
copyright = '2023, picnixz'
author = 'picnixz'

extensions = []

master_doc = 'index'

html_theme = 'alabaster'

##############################################

from sphinx.writers.html import HTML5Translator

class CustomHTMLTranslator(HTML5Translator):
    def visit_desc_space(self, node):
        raise Exception()

    def visit_inline(self, node):
        if node['classes'] == ['w']:
            print('desc_sig_space as inline')
        super().visit_inline(node)

def setup(app):
    app.set_translator('html', CustomHTMLTranslator, override=True)
```

With the above example, the exception is never raised and `'desc_sig_space as inline'` will be printed twice.

### Environment Information

```text
Sphinx version:        6.1.3
Docutils version:      0.18.1
Jinja2 version:        3.1.2
Pygments version:      2.14.0
```

picnixz:
Fix #11272 

If a more precise `fallback` method must be implemented as I suggested in #11272, I can add it to that PR directly or make a separate one.


AA-Turner:
> If a more precise `fallback` method must be implemented as I suggested in #11272, I can add it to that PR directly or make a separate one.

I would suggest a new PR with your proposal, please. I'm unlikely to have significant time to review until late April though, sorry.

A

PR #11288, opened 2023-04-04
Ensure arguments to ``PIL.Image.resize()`` are integers

### LINKED ISSUE #11286: Using epub_max_image_width triggers error because _epub_base.py may submit a float value to resize
### Describe the bug

Line 432 of _epub_base.py does a "/" instead of "//" to create a value that is passed to resize. Since this is not always an integer - it causes a failure.

The fix is simple:
change
'nh = (height * nw) / width
to use // instead of /

### How to Reproduce

set epub_max_image_width
use an image that is resized down but is not resulting in an integer for width when that happens

### Environment Information

```text
n/a
```

### Sphinx extensions

```python
n/a
```

### Additional context

_No response_

jamesreinders:
Updates _epub_base.py so that the computation of nh is always an integer - otherwise some calls to resize fail (floats are not allowed)

Subject: Close issue 11286

### Feature or Bugfix
- Bugfix

### Purpose
Close issue 11286

### Detail
Close issue 11286

### Relates
https://github.com/sphinx-doc/sphinx/issues/11286

### DIFF
--- a/sphinx/builders/_epub_base.py
+++ b/sphinx/builders/_epub_base.py
@@ -424,7 +424,7 @@ class EpubBuilder(StandaloneHTMLBuilder):
                 (width, height) = img.size
                 nw = self.config.epub_max_image_width
                 if width > nw:
-                    nh = (height * nw) / width
+                    nh = (height * nw) // width
                     img = img.resize((nw, nh), Image.BICUBIC)
             try:
                 img.save(path.join(self.outdir, self.imagedir, dest))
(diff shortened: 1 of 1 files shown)


krutziseries (approved):

sphinx/builders/_epub_base.py
```diff
@@ -424,7 +424,7 @@ def copy_image_files_pil(self) -> None:
                 (width, height) = img.size
                 nw = self.config.epub_max_image_width
                 if width > nw:
-                    nh = (height * nw) / width
+                    nh = (height * nw) // width
```
AA-Turner:
```suggestion
                    nh = round((height * nw) / width)
```

AA-Turner:
Ideally we'd have a test but I don't think this hurts.

A

AA-Turner (approved):

AA-Turner:
Thank you James!

A

Issue #11124, opened 2023-01-12
Memory hog for Linux documentation with `-j1`


marxin:
### Describe the bug

As shown here:
https://lore.kernel.org/all/bdd04f2e-64a4-e3ef-eb4c-3681c6f52c2f@gmail.com/

```
 ======  ===================================  ===============================
                     elapsed time                       maxresident
         -----------------------------------  -------------------------------
 Sphinx     -j1     -j2      -j4      -j6       -j1     -j2     -j4     -j6
 ======  ======== ======== ======== ========  ======= ======= ======= =======
  6.1.3  15:03.83 11:31.99  9:35.15  8:49.01  2949056 1059516  978232  967400
  6.1.2  15:11.74 18:06.89 16:39.93      OOM  2961524 5548344 5255372      --
  5.3.0  14:13.04 10:16.81  8:22.37  8:09.74   711532  937660  846016  800340
 ======  ===================================  ===============================
```

The Linux documentation build takes much more memory with `-j1` compared to `5.3.0` release.

I bisected that to 463a69664c2b7f51562eb9d15597987e6e6784cd (one needs cherry-pick dcb4429abacda015f98875fe9489ced696155bd7)

### How to Reproduce

`make SPHINXOPTS="-q -j1" htmldocs`

### Environment Information

```text
Sphinx `6.1.3`
```

marxin:
@AA-Turner

marxin:
I'm attaching CPU and memory graphs for `linux` documentation (with `-j1`) with the current caching:

![linux-sphinx-cache](https://user-images.githubusercontent.com/2658545/229782597-03ccabd5-81dc-4345-82d7-1f3bedbf565b.svg)

and if I reverted the mentioned patch:

![linux-sphinx-no-cache](https://user-images.githubusercontent.com/2658545/229782677-1650d69a-18a5-4674-9505-ab5b81fded7e.svg)

So overall, the caching scheme speeds up the documentation build by **8%**.

However, I noticed that with `-j auto` (`j > 2`), there is **no** cache hit for ` _write_doc_doctree_cache` that is likely caused by the usage of a process pool where there is not overlap in between read and write phase.

marxin:
So the situation with my patch should be back to when multiple jobs are used. When it comes to `-j1`, yes, the memory usage is higher, but the build finishes sooner. If it's really a problem, I can imagine a `sphinx-build` option for this optimization. Feel free to re-open the issue if it's the case.

[...]

PR #11290, opened 2023-04-05
Only use ``_write_doc_doctree_cache`` in serial mode

marxin:
As shown in #11124, if the reading operation is using multiple processes, then the content of self.env._write_doc_doctree_cache is not synchronized with the main process. Thus use it only in serial mode.

For cpython documentation I get to the following numbers before this revision:
-j1 + caching of _write_doc_doctree_cache: **95 s**
-j1 + not caching of _write_doc_doctree_cache: **86 s**
-j32 + caching of _write_doc_doctree_cache: **57 s**
-j32 + no caching of _write_doc_doctree_cache: **48 s**

Related: #11124


AA-Turner:
@marxin How's this?

A

Issue #11285, opened 2023-04-04
[BUG] Tests have global side-effects.


picnixz:
### Describe the bug

While implementing some feature, I wanted to test the latter by selectively running some tests but I found that the tests are not executed in an isolated context, namely they have undesirable side-effects.

A simple way to see that is by *manually* running the following tests:

```sh
python -m tox -e py310 \
  -- -s --no-header --no-summary \
  tests/test_ext_autodoc.py::test_autodoc_typed_instance_variables
```

```sh
python -m tox -e py310 \
  -- -s --no-header --no-summary \
  tests/test_ext_autodoc_autoattribute.py::test_autoattribute_typed_variable_in_alias \
  tests/test_ext_autodoc.py::test_autodoc_typed_instance_variables
```

The first test succeeds but the second one fails with the following assertion error:

```
AssertionError: assert ['', '.. py:m...ed_vars', ...] == ['', '.. py:m...ed_vars', ...]
  At index 62 diff: '   .. py:attribute:: Derived.attr2' != '   .. py:attribute::   Derived.attr7'
  Left contains 31 more items, first extra item: '      :module: target.typed_vars'
  Use -v to get more diff
```

The reason is that, in the first test, the documenter being used is a `ModuleDocumenter` for [tests/roots/test-ext-autodoc/target/typed_vars.py](https://github.com/sphinx-doc/sphinx/blob/29d7166af203b33958f7f6c261ae3813e27b32cc/tests/roots/test-ext-autodoc/target/typed_vars.py). In particular, when the documenter documents its members, the annotations of the `Alias` class is *NOT* updated using the annotations of the `Derived` and `Class` class. By adding

```py
if self.fullname == 'target.typed_vars.Alias.attr2':
    logger.warning([self.fullname, self.parent.__annotations__])
elif self.fullname == 'target.typed_vars.Alias':
    logger.warning([self.object, self.object.__annotations__])
``` 

after 

https://github.com/sphinx-doc/sphinx/blob/669bcc0a190ce9921451bad28431886fbaad170b/sphinx/ext/autodoc/__init__.py#L877-L880

then the output of the first command is:

```
# warning:
WARNING: [<class 'target.typed_vars.Derived'>, {'attr7': <class 'int'>}]
```

**However**, if we do the same thing when running the second command, we get the following output:

```
# warning: (this is for test_autoattribute_typed_variable_in_alias)
[...]

PR #11293, opened 2023-04-06
Make test_attrgetter_using test dependant.

marxin:
One cannot run the mentioned test individually and thus it brings troubles with pytest -n auto.

@AA-Turner


picnixz:
Since your seem to fix those kind of tests, maybe you can gather them under the issue I created (which also includes a dependent test) https://github.com/sphinx-doc/sphinx/issues/11285 or make a separate "summary" issue ? (so that we know which are the ones to fix / that were already fixed)

marxin:
Sure, good point. Do you have any other tests that have a global context dependency?

picnixz:
Well, just running `python -m tox -e py310 -- tests/test_ext_autodoc*` would yield more failures but I don't know which test is dependent on which. I found the dependencies by explicitly testing different pairs (in the order they were specified when running `python -m tox -e py310 -- tests/test_ext_autodoc*`).

marxin:
I see, I can confirm that as well with `pytest tests/test_ext_autodoc.py --random-order`. Can you please work on that :P ?

picnixz:
> I see, I can confirm that as well with pytest tests/test_ext_autodoc.py --random-order. Can you please work on that :P ?

Well, I wonder actually about the way to patch this thing. Either we patch it by "cleanly" resetting things up, or we modify the flow of `do_autodoc` in the tests so that every autodoc test is independent. That's why I created #11285 in order to know which direction would be the best. Because I highly suspect that this is not the only test that has dependencies and that the whole testing flow should be revised.

AA-Turner:
> Either we patch it by "cleanly" resetting things up, or we modify the flow of `do_autodoc` in the tests so that every autodoc test is independent.

I'm not entirely what the two options are here, but I think philisophically we should try and eliminate implicit state between tests. I think every test *should* be independent, but if it makes the tests take much longer, perhaps we make any state that is needed explicit via module-level setup or fixtures etc?

A

PR #11294, opened 2023-04-06
Use a shared file-system lock in ``create_server``

marxin:
With a parallel run of tests, one gets "Address already in use" errors as all tests attempt to bind to the same port. Fix it with a shared file-system lock.

@AA-Turner

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -9,11 +9,11 @@ import filelock
 # Generated with:
 # $ openssl req -new -x509 -days 3650 -nodes -out cert.pem \
 #     -keyout cert.pem -addext "subjectAltName = DNS:localhost"
-FILE_DIR = pathlib.Path(__file__).parent
-CERT_FILE = str(FILE_DIR / "certs" / "cert.pem")
+TESTS_ROOT = pathlib.Path(__file__).parent
+CERT_FILE = str(TESTS_ROOT / "certs" / "cert.pem")
 
 # File lock for tests
-LOCK_PATH = str(FILE_DIR / 'test-server.lock')
+LOCK_PATH = str(TESTS_ROOT / 'test-server.lock')
 
 
 class HttpServerThread(threading.Thread):
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -4,10 +4,16 @@ import pathlib
 import threading
 from ssl import PROTOCOL_TLS_SERVER, SSLContext
 
+import filelock
+
 # Generated with:
 # $ openssl req -new -x509 -days 3650 -nodes -out cert.pem \
 #     -keyout cert.pem -addext "subjectAltName = DNS:localhost"
-CERT_FILE = str(pathlib.Path(__file__).parent / "certs" / "cert.pem")
+FILE_DIR = pathlib.Path(__file__).parent
+CERT_FILE = str(FILE_DIR / "certs" / "cert.pem")
+
+# File lock for tests
+LOCK_PATH = str(FILE_DIR / 'test-server.lock')
 
 
 class HttpServerThread(threading.Thread):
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -1,13 +1,20 @@
 import contextlib
 import http.server
+import os
 import pathlib
 import threading
 from ssl import PROTOCOL_TLS_SERVER, SSLContext
 
+import filelock
```
picnixz, line 7: This one should be marked as an external dependency

marxin: Can you help me how to do that, please?

picnixz: I think you can simply add a dependency in pyproject.toml under `test = [...]` as `test = [..., "filelock"]`. Afterwards, tox would install that dependency when running tests. If there are other locations to modify, let's wait for @AA-Turner to confirm.

tests/utils.py
```diff
@@ -34,12 +41,15 @@ def __init__(self, handler, *args, **kwargs):
...
-            server_thread.terminate()
+        lock = filelock.FileLock(LOCK_PATH)
+        with lock:
+            server_thread = thread_class(handler, daemon=True)
+            server_thread.start()
+            try:
+                yield server_thread
+            finally:
+                server_thread.terminate()
+            os.remove(LOCK_PATH)
```
picnixz: If the file fails to be removed, is there a possibility for the remaining tests to indefinitely hang ?

marxin: I don't think so, there will be an exception raised and the locked scoped will be left.

marxin: But the removal is problematic for Windows:
```
>           os.remove(LOCK_PATH)
E           PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'D:\\a\\sphinx\\sphinx\\tests\\test-server.lock'
```

so we should likely leave the file there.

picnixz: Yes, that's what I thought, but I am wondering how the `filelock` API *acquires* its lock in the first place. Anyway, assuming that it works, are you sure that the port will be freed at the same time that the lock is released ?

marxin: Yes, I made an intensive parallel testing `pytest tests/test_build_linkcheck.py -n 32` and it works fine now.

picnixz: > Yes, I made an intensive parallel testing `pytest tests/test_build_linkcheck.py -n 32` and it works fine now.

What I actually fear is that the behaviour is different on Linux and Windows builds. Therefore, I am wondering whether we could add a specific "optional" CI checks that would do that check for us on those builds.

marxin: What type of check do you speak about?

picnixz: This one: `pytest tests/test_build_linkcheck.py -n 32`. Like checking separately the tests that lead to race conditions.

marxin: Sure, let me write it on my TODO-list. Yes, we can do that once we'll enable parallel tests generally.

tests/utils.py
```diff
@@ -4,10 +4,16 @@
...
+
 # Generated with:
 # $ openssl req -new -x509 -days 3650 -nodes -out cert.pem \
 #     -keyout cert.pem -addext "subjectAltName = DNS:localhost"
-CERT_FILE = str(pathlib.Path(__file__).parent / "certs" / "cert.pem")
+FILE_DIR = pathlib.Path(__file__).parent
+CERT_FILE = str(FILE_DIR / "certs" / "cert.pem")
+
+# File lock for tests
+LOCK_PATH = str(FILE_DIR / 'test-server.lock')
```
AA-Turner:
```suggestion
TESTS_ROOT = pathlib.Path(__file__).parent
CERT_FILE = str(TESTS_ROOT / "certs" / "cert.pem")

# File lock for tests
LOCK_PATH = str(TESTS_ROOT / 'test-server.lock')
```

PR #11297, opened 2023-04-06
Fix instability in ``test_ext_viewcode``

marxin:
Use `freshenv` argument and remove `outdir` for one tests as the previous one can leave an unexpected leftover.

Related: #11285.

### DIFF
--- a/tests/test_ext_viewcode.py
+++ b/tests/test_ext_viewcode.py
@@ -1,12 +1,13 @@
 """Test sphinx.ext.viewcode extension."""
 
 import re
+import shutil
 
 import pygments
 import pytest
 
 
-@pytest.mark.sphinx(testroot='ext-viewcode')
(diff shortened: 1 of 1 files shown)


tests/test_ext_viewcode.py
```diff
@@ -52,6 +53,7 @@ def test_viewcode(app, status, warning):
 
 @pytest.mark.sphinx('epub', testroot='ext-viewcode')
 def test_viewcode_epub_default(app, status, warning):
+    shutil.rmtree(app.outdir)
```
picnixz, line 56: Is the `app.outdir` always distinct for distinct tests ? (for multi-processes executions, I don't know whether this could be an issue or not and I am not familiar with the current testing setup/tearDown flow of Sphinx tests).

marxin: For `pytest -n auto`, each `gw` (worker) has its own folder, so it's fine.

PR #11298, opened 2023-04-06
Support type comments in ``PropertyDocumenter``

### LINKED ISSUE #11277: [BUG] PropertyDocumenter does not add annotation from type comment
### Describe the bug

Consider the following code:

```py
class MyClass:
    """My class."""

    @property
    def foo(self) -> int:
      """The foo property."""
      return 0

    @property
    def bar(self):
      # type: () -> int
      """The bar property."""
      return 0
``` 

Currently, the `sphinx.ext.autodoc.PropertyDocumenter` does not fire the `autodoc-before-process-signature` event, and thus the `sphinx.ext.autodoc.type_comments.update_annotations_using_type_comments` event handler is never called for properties. Instead, this documenter solely relies on an existing return annotation, which may or may not exist

https://github.com/sphinx-doc/sphinx/blob/669bcc0a190ce9921451bad28431886fbaad170b/sphinx/ext/autodoc/__init__.py#L2731-L2747

In particular, the `bar` property does not have a return annotation `int` because the type comment is never parsed. Note that `PropertyDocumenter.add_directive_header` is given as input the stringified signature of the object, but for properties, the signature does not include the return annotation. More precisely, when entering `PropertyDocumenter.add_directive_header`, 

https://github.com/sphinx-doc/sphinx/blob/669bcc0a190ce9921451bad28431886fbaad170b/sphinx/ext/autodoc/__init__.py#L2716-L2730

the *real* function is extracted from the descriptor *after* the signature string has been computed. In particular, the signature being computed at

https://github.com/sphinx-doc/sphinx/blob/669bcc0a190ce9921451bad28431886fbaad170b/sphinx/ext/autodoc/__init__.py#L2731-L2735

is now the *correct* signature and is used to extract return annotation for the property type. However, this signature is never given the chance to be updated using type comments since the `autodoc-before-process-signature` event is never fired.

### How to Fix

This issue can be easily fixed by one of the following method:

* Fire the `autodoc-before-process-signature` event directly in `PropertyDocumenter.import_object` method in order to update the signature of the real object. I fix this in my project as follows:

```python
class CustomPropertyDocumenter(PropertyDocumenter):
    option_spec: OptionSpec = PropertyDocumenter.option_spec.copy()

    def document_members(self, *args, **kwargs):
        return  # do not support members

    def import_object(self, raiseerror=None):
        rv = super().import_object(raiseerror)
        if rv:
            func = self._get_function()
            # update underlying property's annotations if needed
            self.app.emit('autodoc-before-process-signature', func, False)
        return rv

    def _get_function(self):
        if safe_getattr(self.object, 'fget', None):  # property
            return self.object.fget

        if safe_getattr(self.object, 'func', None):  # cached_property
            return self.object.func

        return None
```
  
  I think it makes sense to fire this event at the earliest possible stage since the directive options (which are added early during the generation stage) depend on the signature of the member being documented. In particular, `autodoc-before-process-signature` should be indeed fired before calling `Documenter.format_signature`, namely inbetween `Documenter.import_object`[^1] and `Documenter.format_signature`[^2]:

-----

[^1]: https://github.com/sphinx-doc/sphinx/blob/669bcc0a190ce9921451bad28431886fbaad170b/sphinx/ext/autodoc/__init__.py#L878

[^2]: https://github.com/sphinx-doc/sphinx/blob/669bcc0a190ce9921451bad28431886fbaad170b/sphinx/ext/autodoc/__init__.py#L930

### class.py

```py
class MyClass:
    """My class."""

    @property
    def foo(self) -> int:
      """The foo property."""
      return 0

    @property
    def bar(self):
      # type: () -> int
      """The bar property."""
      return 0
```

### index.rst

```rst
Foo
===

.. autoclass:: class.MyClass
   :members:
```

### conf.py

```py
import os
import sys

sys.path.insert(0, os.path.abspath('.'))

project = 'Foo'
copyright = '2023, picnixz'
author = 'picnixz'

extensions = ['sphinx.ext.autodoc']

include_patterns = ['index.rst', 'class.py']

master_doc = 'index'

html_theme = 'alabaster'
``` 

### Environment Information

```text
Sphinx version:        6.1.3
Docutils version:      0.18.1
Jinja2 version:        3.1.2
Pygments version:      2.14.0
```

### Sphinx extensions

```python
['sphinx.ext.autodoc']
```

### Additional context

Solving this issue may serve in solving #10396

picnixz:
Fix #11277


AA-Turner:
Thank you for the very comprehensive report and fix!

A

PR #11303, opened 2023-04-08
Re-allow localisation when SOURCE_DATE_EPOCH is set (follow-up to #10949)

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- I may have take an entirely incorrect (basically, inverted) approach to solving #9778
- The approach - disabling localisation entirely for reproducible builds (#10949) is likely not a good idea
- I'd really appreciate sanity-checking on this

### Detail
- #10949 fixed an unusual behaviour around lazy-loading of localised resources
  - > During investigation, it turned out that many gettext-localized values -- particularly in Python modules under sphinx.domains -- were being translated at module-load-time (in other words, when each Python module was import'ed) -- and would not subsequently be re-localized.
- However, #10949 also completely disabled localisation during reproducible builds (detected by a non-empty `SOURCE_DATE_EPOCH` environment variable)
- Disabling localisation may not have been necessary: as mentioned in the [original bug description](https://github.com/sphinx-doc/sphinx/issues/9778#issue-1036427985), the `objects.inv` file -- where reproducibility problems were occurring -- was already being requested in a single language (English-only); so localisation of the rest of the package-under-build should be irrelevant

### Relates
- Original bug report: #9778
- Previous fix attempt: #10949

### DIFF
--- a/sphinx/locale/__init__.py
+++ b/sphinx/locale/__init__.py
@@ -105,22 +105,10 @@ def init(
     if translator.__class__ is NullTranslations:
         translator = None
 
-    if getenv('SOURCE_DATE_EPOCH') is not None:
-        # Disable localization during reproducible source builds
-        # See https://reproducible-builds.org/docs/source-date-epoch/
-        #
-        # Note: Providing an empty/none value to gettext.translation causes
-        # it to consult various language-related environment variables to find
-        # locale(s).  We don't want that during a reproducible build; we want
-        # to run through the same code path, but to return NullTranslations.
-        #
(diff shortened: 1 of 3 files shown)


jayaddison:
Sorry @AA-Turner @lamby; I don't currently feel confident that #10949 was the correct fix -- and nor am I confident that this patch makes it correct.

I'm going to revert #10949 completely until I can confirm repro and fix of the originally-reported issue.  Maybe these changes do resolve the problem, but for a large widely-used project, I think that a 'first, do no harm' principle makes sense.

jayaddison:
Revert opened as #11306.

PR #11305, opened 2023-04-09
feat: initial Sphinx Java11+ support

### LINKED ISSUE #11304: [FEATURE] Sphinx Java 11+ support
I'm trying to figure out how to also use Sphinx for Java documentation.

At hight level it will be possible to to create Java documentation at the same way that current Python users do, for example:

```
.. testcode::

    import java.util.Arrays
    import java.util.List

    List<String> result = Arrays.asList("Python", "Java");
    System.out.println(result);

.. testoutput::

    [Python, Java]
```

Testing by `make javadoctest`

This is the current high level flow for Sphinx Python testing documentation process validation:
- Python:
1. Register Builder(Runs test snippets in the documentation)
2. Flow: User (create documentation) -> Sphinx Directive to Test Code -> `Builder Compile the Code` (DocTestBuilder - Compile source into a code object that can be executed by exec() or eval()) -> Sphinx Directive to Test Output

- **Proposal for Java**:

1. Register New Builder `JavaDocTestBuilder extends from DocTestBuilder` (Runs test snippets in the documentation)
2. Flow: User (create documentation) -> Sphinx Directive to Test Code -> `Builder Compile the Code` (**create** `JavaDocTestBuilder extends from DocTestBuilder`) ->  Sphinx Directive to Test Output
3. Java side will **override `compile` method** for running examples in docstrings via **Java [Shell](https://docs.oracle.com/en/java/javase/11/jshell/introduction-jshell.html)**

**Additional context**

These Java integration will be offer flavors to document Java code:
- Java native code: To test documentation for projects that consume only Java native libraries will be needed to define `conf.py` with only flavor `java`.
- Java code that consume third party libraries: In this case will be needed to define by `conf.py` a flavor with `java_with_maven`, and also define where is your maven project that contains third dependencies thru an absolute `path` (also consider option to integrate with Gradle definition).

davisusanibar:
Subject: Enable Java11+ snippest documentation support on Sphinx

### Feature or Bugfix
- Feature

### Purpose
1. At hight level it will be possible to to create Java documentation at the same way that current Python users do, for example:

```
.. testcode::

    import java.util.Arrays
    import java.util.List

    List<String> result = Arrays.asList("Python", "Java");
    System.out.println(result);

.. testoutput::

    [Python, Java]
```

Testing by `make javadoctest`

2. These Java integration will be offer flavors to document Java code:
- Java native code: To test documentation for projects that consume only Java native libraries will be needed to define `conf.py` with only flavor `java`.
- Java code that consume third party libraries: In this case will be needed to define by `conf.py` a flavor with `java_with_maven`, and also define where is your maven project that contains third dependencies thru an absolute `path` (also consider option to integrate with Gradle definition).

3. Java side will **override `compile` method** for running examples in docstrings via **Java [Shell](https://docs.oracle.com/en/java/javase/11/jshell/introduction-jshell.html)**

### Detail

- [x] [Feature] Sphinx Java11+ support for Java Native Code
- [x] [Feature] Sphinx Java11+ support for Java 3rd Party Dependencies thru Maven
- [x] [Test] Testing locally
- [x] [Test] Testing CI Unix
- [ ] [Test] Testing CI Windows
- [x] [Fix] Fix code review issues
- [x] [Documentation] Update documentation

### Relates
- To close #11304


sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+            # get list of all Java jar dependencies
+            with open(path_java_project / ".cp.tmp", encoding='utf-8') as f:
+                stdout_dependency = f.read()
+            if not stdout_dependency:
+                raise RuntimeError(
+                    __("Invalid process to collect JShell dependencies library"),
+                )
+        else:
+            raise RuntimeError(
+                __("Invalid Java flavor to be analyzed"),
```
picnixz: It would be good to indicate the flavour that is not known and either use an ExtensionError or create a specific exception class for those things.

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+            # go to project that contains all your Java maven dependencies
+            path_java_project = self.config.javadoc_options['path']
+            # create list of all Java jar dependencies
+            subprocess.check_call(
+                [
+                    "mvn",
+                    "-q",
+                    "dependency:build-classpath",
+                    "-DincludeTypes=jar",
+                    "-Dmdep.outputFile=.cp.tmp",
```
picnixz: What happens if this file already exists on the filesystem for some obscure reason and that it should not be overwritten ? (I'd suggest using a temporary file instead).

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+            if not (path_java_project / ".cp.tmp").exists():
+                raise RuntimeError(
+                    __("Invalid process to create JShell dependencies library"),
+                )
+            # get list of all Java jar dependencies
+            with open(path_java_project / ".cp.tmp", encoding='utf-8') as f:
+                stdout_dependency = f.read()
+            if not stdout_dependency:
+                raise RuntimeError(
```
picnixz: Maybe another type of exception (e.g., derived from ExtensionError)?

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+                    "dependency:build-classpath",
+                    "-DincludeTypes=jar",
+                    "-Dmdep.outputFile=.cp.tmp",
+                    f"-Dversion={self.env.config.version}",
+                ],
+                cwd=path_java_project,
+                text=True,
+            )
+            if not (path_java_project / ".cp.tmp").exists():
+                raise RuntimeError(
```
picnixz: Maybe another type of exception (e.g., derived from ExtensionError)?

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+        # continue with Python logic code to do Java output validation
+        output = f"print('''{self.clean_output(out_java)}''')"
+
+        # continue with Sphinx default logic
+        return compile(output, name, self.type, flags, dont_inherit)
+
+    def clean_output(self, output: str) -> str:
+        if output[-3:] == '-> ':
+            output = output[:-3]
+        if output[-1:] == '\n':
```
picnixz: You can replace this by `output.endswith('\n')` I think. Correct me if I am wrong, but do you want to remove the *last* newline character or *all* the trailing newline characters (in which case, I'd suggest using `output.rstrip('\n')`)?

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+            output = output[:-1]
+        output = (4 * ' ').join(output.split('\t'))
+        return output
+
+
+def setup(app: Sphinx) -> Dict[str, Any]:
```
picnixz: Use `dict[str, Any]` instead in accordance with PEP 585 (do not forget to use `from __future__ import annotations`).

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+    app.add_directive("testcode", JavaTestcodeDirective)
```
picnixz: This kind of directive is a bit too generic if we can only test for Java code. What about creating a domain for Java and adding that directive to the Java domain ?

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+    app.add_directive("testoutput", TestoutputDirective)
+    app.add_builder(JavaDocTestBuilder)
+    # this config value adds to sys.path
+    app.add_config_value("doctest_path", [], False)
```
picnixz: I would suggest prefixing all configuration values by `java` if they are only intended for Java code.

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+    DocTestBuilder,
+    TestcodeDirective,
+    TestoutputDirective,
+    doctest,
+    sphinx,
+)
+from sphinx.locale import __
+
+
+class JavaTestcodeDirective(TestcodeDirective):
```
picnixz: Maybe `JavaTestcodeDirective` -> `JavaTestCodeDirective`? 

In addition, the directive could support different versions of Java instead of only supporting Java 11. This could be configured via `:version: 11` for instance.

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+    def run(self):
+        node_list = super().run()
+        print("check_node_list")
```
picnixz: I know that this is a draft, but remember to remove the `print` (maybe by adding some todo comment).

picnixz (commented):
I know that this is a draft, but here are some comments.

sphinx/ext/javadoctest.py
davisusanibar: > Maybe `JavaTestcodeDirective` -> `JavaTestCodeDirective`?
> 
> In addition, the directive could support different versions of Java instead of only supporting Java 11. This could be configured via `:version: 11` for instance.

Currently it is tested with ny Java version >= 11. it depende about what version was configured intn you JAVA_HOME.

picnixz: I agree with that, but since Java 11 is not yet deprecated, people might write code that is *not* compatible with some Java version (e.g., Java 12 adds more methods to the `String` interface that are not present in Java 11). That way, if the current Java version is not installed on the OS running the tested code, then the snippet is not executed at all and there are not errors (I would suggest emitting a warning instead).

sphinx/ext/javadoctest.py
davisusanibar: When we invoke Jshell to run the code at `out_java, err_java = proc_jshell_process.communicate(code)`, for some reason the response message add `->\n`

Expected:
`[Python, Java]`

But JShell return:
`[Python, Java]
    -> `

Just changing to use `output.endswith('\n')`

sphinx/ext/javadoctest.py
davisusanibar: Changed

sphinx/ext/javadoctest.py
davisusanibar: Thanks, changed

sphinx/ext/javadoctest.py
davisusanibar: Changed

sphinx/ext/javadoctest.py
davisusanibar: Changed

sphinx/ext/javadoctest.py
davisusanibar: Changed

sphinx/ext/javadoctest.py
davisusanibar: For current proposal we are only working testcode / testoutput. For next feature that will be doctest in that escenario there will be needed more changes and I'll be create that module. Let see if feature is used/needed to add more improvements.

sphinx/ext/javadoctest.py
davisusanibar: This part of variables are part of main variables.

The new variable added is `java_doctest_config` and containt prefix java recommended.

sphinx/ext/javadoctest.py
davisusanibar: Just added logic to validate Java/Maven needed by flavors. There is also pending to add a new feature thru `doctest` to handle `javaversion` as the same that [pyversion](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#:~:text=pyversion%2C%20a%20string%20option%2C%20can%20be%20used%20to%20specify%20the%20required%20Python%20version%20for%20the%20example%20to%20be%20tested.%20For%20instance%2C%20in%20the%20following%20case%20the%20example%20will%20be%20tested%20only%20for%20Python%20versions%20greater%20than%203.3%3A) is working, but this will be consider as new feature in another PR

sphinx/ext/javadoctest.py
davisusanibar: Deleted

sphinx/ext/javadoctest.py
picnixz: What I meant is that `testcode` may not be the best name for that. It does not highlight the fact that the code being tested is Java code. Also, if third-party extension use that name (and they should not use such generic name IMO), I suggest using `java_testcode` and `java_testoutput` at least (or something like that). Or adding using `testcode` and `testoutput` with an additional `language` directive option specifying the language to use.

tests/test_ext_javadoctest.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+
+@pytest.mark.sphinx('javadoctest', testroot='ext-javadoctest')
+def test_build(app, status, warning):
+    app.builder.build_all()
+    if app.statuscode != 0:
```
picnixz: Testing should be done via assertion and not by raising assertions. Use: 

`assert app.statuscode == 0`

It would be good to test the output as well (like, what is *really* the output, line per line, if there are any). Just ensuring that `app.statuscode` does not fail is insufficient I think.

sphinx/ext/javadoctest.py
picnixz: If the response message is `->\n`, then you should remove the whole suffix in one go and not only `->` then `\n`.

sphinx/ext/javadoctest.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # continue with Sphinx default logic
+        return compile(output, name, self.type, flags, dont_inherit)
+
+    def clean_output(self, output: str) -> str:
+        if output[-3:] == '-> ':
+            output = output[:-3]
+        if output[-1:] == '\n':
+            output = output[:-1]
+        output = (4 * ' ').join(output.split('\t'))
```
picnixz, line 394: If you only want to indent the output, you can use `textwrap.indent` together with `tabsize=4` instead. It will take care of normalizing tabs. If your output has some `\t` that are *inside* a string, that may break things (and you should also test this).

tests/roots/test-ext-javadoctest-maven/index.rst
```diff
@@ -0,0 +1,27 @@
...
+dependency management suchs as Maven/Gradle, in this case initially we are offering support
+for Maven.
+
+In this case is needed to define by `conf.py` a flavor with `java_with_maven`, and also define
+where is your maven project that contains third dependencies thru an absolute `path`.
+
+Java Guava Example
+------------------
+
+.. testcode::
```
picnixz: We should have more tests:

- We need failing tests to test against the exceptions that are raised (like, a bad input, a bad output, bad input and output, failure to parse the input, failure to parse the output, etc).
- We need passing tests testing complicated strings (with `\t` for instance) in the output. 
- May we have a smaller dependency than `google` ? I am not very happy to have this installed everytime we run the tests. 
- We need a test assuming that 1) `java` is not configured or 2) `java` has an unsupported version.

tests/roots/test-ext-javadoctest-maven/index.rst
```diff
@@ -0,0 +1,27 @@
+Test Java With Third Dependencies Handled by Maven
+==================================================
+
+In case we need to test documentation for projects that consume Java native libraries and
+Third party Java dependencies, then, is needed to configure/add these dependencies thru a
```
picnixz: - In case -> if
- Third party -> third-party (it is used as an adjective) 
- thru -> through (thru is informal)
- then, is needed -> then these dependencies need to be configured or added through a [..]
- dependency management -> dependency management tool 
- suchs as -> such as Maven or Gradle. 
- in this case innitially -> Currently, only Maven is supported. 

This file should not be inside the testroot but should be part of the Sphinx documentation itself. For tests, only the example is needed (but the example must also be included in the documentation).

tests/roots/test-ext-javadoctest-maven/index.rst
```diff
@@ -0,0 +1,27 @@
...
+dependency management suchs as Maven/Gradle, in this case initially we are offering support
+for Maven.
+
+In this case is needed to define by `conf.py` a flavor with `java_with_maven`, and also define
```
picnixz, line 8: This paragraph needs to be more precise: You should explain the new options using the `.. confval::` directive (e.g., what is `javadoc_options`, etc). Again this paragraph should be moved to the documentation and not in the testroot (I know that this is the last milestone of your tasks but keep in mind this comment).

tests/roots/test-ext-javadoctest-maven/example/pom.xml
```diff
@@ -0,0 +1,24 @@
...
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
+        </dependency>
+    </dependencies>
+</project>
```
picnixz: I think a newline is missing here.

davisusanibar:
Sorry for the long time under draft implementation, @picnixz / @AA-Turner PR is ready for review.

In addition, I would appreciate if you could help me with the render documentation errors, since I didn't catch what the cause is.

Here are my concerns about current implementation:

- Initially, PR [re-used current directives testcode & testoutput](https://github.com/sphinx-doc/sphinx/blob/4fe49639c5373285998301d8816e76679a996b1f/sphinx/ext/javadoctest.py#L105:L106), then minimal code was needed to extend current doctest.py methods to support Java testing documentation.
- A comment on the PR suggested adding a custom Java directive, and this PR contains all the changes, but there is a lot of duplicate code in doctest.py to support new `JavatestCode` or `JavatestOutput`.
- As a starting point, I suggest we keep things simple, then iterate, and improve as necessary (the main drawback is the end user will be confused with the same name directive for different languages).

I would appreciate any comments or ideas you may have.

davisusanibar:
Please consider, as an additional limitation, if the end-user is using a dockerized environment, then the first test will fail with this error message. For this reason, in the first test, I just skip it:

```
.. javatestcode::

    System.out.println("docker");

.. javatestoutput::

    docker
```

The error will be:

```
GOT:

    Sep 03, 2023 1:05:54 AM java.util.prefs.FileSystemPreferences$1 run
    INFO: Created user preferences directory.
    docker

EXPECTED:

    docker
```

davisusanibar:
> Please consider, as an additional limitation, if the end-user is using a dockerized environment, then the first test will fail with this error message. For this reason, in the first test, I just skip it:
> 
> ```
> .. javatestcode::
> 
>     System.out.println("docker");
> 
> .. javatestoutput::
> 
>     docker
> ```
> 
> The error will be:
> 
> ```
> GOT:
> 
>     Sep 03, 2023 1:05:54 AM java.util.prefs.FileSystemPreferences$1 run
>     INFO: Created user preferences directory.
>     docker
> 
> EXPECTED:
> 
>     docker
> ```

The custom directive extension caused this error, which was not present in the original proposal.

picnixz:
- Docs: search for warnings that are emitted (look in the CI/CD logs). I only have access to my phone but I think the rst file you write has some issue somewhere. 

Alternatively, you could try to generate the Sphinx docs locally to catch the errors in a less verbose output.

For the rest, I'll only review it upon my return (mid October).

PR #11306, opened 2023-04-09
Revert "Disable localisation when ``SOURCE_DATE_EPOCH`` is set (#10949)"

jayaddison:
Reverts pull request #10949 and follow-up commit 2c83af0aab7080e0b78d4a16981eed878b2cac4c.

(post-merge of that PR, I haven't been able to re-confirm the before-and-after behaviour of the fix, and I'd prefer not to risk it being included in a `sphinx` release and causing issues for users until we're fairly-highly confident that it fixes the problem, only the problem, and doesn't introduce unintended side-effects)


jayaddison:
From further investigation: I do now believe that #10949 was unnecessary; see commentary at https://github.com/sphinx-doc/sphinx/issues/9778#issuecomment-1501172176 for details.

sphinx/locale/__init__.py
```diff
@@ -203,7 +203,12 @@ def setup(app):
     .. versionadded:: 1.8
     """
     def gettext(message: str) -> str:
-        return _TranslationProxy(catalog, namespace, message)  # type: ignore[return-value]
```
jayaddison, line 206: This seemed like a nice cleanup -- and unlocks other simplifications -- so if it's possible to re-apply later in a separate change, then I'd like to do that.

tests/test_locale.py
```diff
@@ -55,44 +55,3 @@ def test_add_message_catalog(app, rootdir):
...
-def test_init_environment_language(rootdir, monkeypatch):
-    with monkeypatch.context() as m:
-        m.setenv("LANGUAGE", "en_US:en")
-        _ = _empty_language_translation(rootdir)
-        assert _('Hello world') == 'HELLO WORLD'
-
-    with monkeypatch.context() as m:
-        m.setenv("LANGUAGE", "et_EE:et")
-        _ = _empty_language_translation(rootdir)
-        assert _('Hello world') == 'Tere maailm'
```
jayaddison, line 76: This test case may also be worth salvaging separately, because it doesn't rely on `SOURCE_DATE_EPOCH`.

sphinx/locale/__init__.py
AA-Turner: Which bit were you talking about here? Open to keeping it if my partial revert didn't.

A

tests/test_locale.py
AA-Turner: Kept!

sphinx/locale/__init__.py
jayaddison: In particular: switching everything over to lazy-loading (so that the `gettext` method doesn't require any conditional logic) - and as a side-effect, allowing content to be localised into more than one language during a unit test.

Referring back to some previous [notes](https://github.com/sphinx-doc/sphinx/pull/10949#issue-1429065601):
> Attempting to write a test case to build the same application in two different languages was not initially possible -- the first-loaded translation catalog (as found in the sphinx.locale.translators global variable) -- would remain in-use for subsequent application builds under different locales

This can wait until after 6.2.x, unless you're particularly keen to dig into it.  I'd plan to take another look at it fresh in a few days' time.

jayaddison: In particular: switching everything over to lazy-loading (so that the `gettext` method doesn't require any conditional logic) - and as a side-effect, allowing content to be localised into more than one language during a unit test.

Referring back to some previous [notes](https://github.com/sphinx-doc/sphinx/pull/10949#issue-1429065601):
> Attempting to write a test case to build the same application in two different languages was not initially possible -- the first-loaded translation catalog (as found in the sphinx.locale.translators global variable) -- would remain in-use for subsequent application builds under different locales

This can wait until after 6.2.x, unless you're particularly keen to dig into it.  I'd plan to take another look at it fresh in a few days' time.

PR #11318, opened 2023-04-11
linkcheck builder: close streamed HTTP response objects when no content reads are required

### LINKED ISSUE #11317: linkcheck builder: streamed HTTP requests do not close their local socket
### Describe the bug

Running `pytest tests/test_build_linkcheck.py` emits `ResourceWarning`s related to unclosed sockets, such as:

```
tests/test_build_linkcheck.py::test_invalid_ssl
  /usr/lib/python3.11/logging/__init__.py:1572: ResourceWarning: unclosed <socket.socket [closed] fd=15, family=2, type=1, proto=6>
    next_f = f.f_back
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
```

This occurs when HTTP requests are made using the `requests` library with `stream=True` enabled (approximate behaviour: do not begin consuming content from the connection immediately, and provide it iteratively as it is received).

This occurs in a couple of places currently:

- https://github.com/sphinx-doc/sphinx/blob/776d01eb6229f5d5384da17fc15df9d50fa8e87a/sphinx/builders/linkcheck.py#L319-L320
- https://github.com/sphinx-doc/sphinx/blob/776d01eb6229f5d5384da17fc15df9d50fa8e87a/sphinx/builders/linkcheck.py#L341-L343

I'm not certain whether this implies a bug/problem.  The upstream `requests` project mentions that connection pooling and the cost of recreating sockets (as opposed to re-using existing ones) can be significant.

I've found two possible resolutions locally (either: make requests using a context-manager, so that the `__exit__` implicitly calls `.close`, or: call `.close` on the response object explicitly), and will offer a pull request with one of those.

### How to Reproduce

Simple repro case:
```sh
$ pytest tests/test_build_linkcheck.py`
```

To trace where the resource allocations originally occurred:

```sh
$ python3 -X tracemalloc=20 -m pytest tests/test_build_linkcheck.py
```

### Environment Information

```text
Platform:              linux; (Linux-6.1.0-7-amd64-x86_64-with-glibc2.36)
Python version:        3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0])
Sphinx version:        6.2.0+/2c83af0aa
Docutils version:      0.19
Jinja2 version:        3.1.2
Pygments version:      2.15.0
```

### Sphinx extensions

```python
None
```

### Additional context

Discovered while investigating #11299, although does not appear directly related.

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Intended to reduce warning noise (`ResourceWarning` output) during linkcheck build tests

### Detail
- The linkchecker makes HTTP requests in a couple of places in streaming mode to minimize the amount of traffic/data required
- Since `requests` doesn't know whether the caller may intend to read content from a streamed HTTP response object, it holds the socket open
- We can ~~explicitly close~~ invoke each request using a [context manager](https://docs.python.org/3/reference/datamodel.html#context-managers), ~~since we know for certain that we are not going to read data from it in these cases~~ to allow the [corresponding `__exit__` method](https://github.com/psf/requests/blob/61c324da43dd8b775d3930d76265538b3ca27bc1/requests/models.py#L709-L710) to perform cleanup

### Relates
- Resolves #11317

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ env:
   FORCE_COLOR: "1"
   PYTHONDEVMODE: "1"  # -X dev
   PYTHONWARNDEFAULTENCODING: "1"  # -X warn_default_encoding
-  PYTHONWARNINGS: "error,always:unclosed:ResourceWarning::"  # default: all warnings as errors, except ResourceWarnings about unclosed items
+  PYTHONWARNINGS: "error"
 
 jobs:
   ubuntu:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ env:
   FORCE_COLOR: "1"
   PYTHONDEVMODE: "1"  # -X dev
   PYTHONWARNDEFAULTENCODING: "1"  # -X warn_default_encoding
-  PYTHONWARNINGS: "error"
+  PYTHONWARNINGS: "error"  # default: all warnings as errors
 
 jobs:
   ubuntu:
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -316,21 +316,20 @@ class HyperlinkAvailabilityCheckWorker(Thread):
             try:
                 if anchor and self.config.linkcheck_anchors:
                     # Read the whole document and see if #anchor exists
-                    response = requests.get(req_url, stream=True, config=self.config,
-                                            auth=auth_info, **kwargs)
-                    response.raise_for_status()
-                    found = check_anchor(response, unquote(anchor))
+                    with requests.get(req_url, stream=True, config=self.config, auth=auth_info,
+                                      **kwargs) as response:
+                        response.raise_for_status()
+                        found = check_anchor(response, unquote(anchor))
 
-                    if not found:
-                        raise Exception(__("Anchor '%s' not found") % anchor)
+                        if not found:
+                            raise Exception(__("Anchor '%s' not found") % anchor)
                 else:
                     try:
                         # try a HEAD request first, which should be easier on
                         # the server and the network
-                        response = requests.head(req_url, allow_redirects=True,
-                                                 config=self.config, auth=auth_info,
-                                                 **kwargs)
-                        response.raise_for_status()
+                        with requests.head(req_url, allow_redirects=True, config=self.config,
+                                           auth=auth_info, **kwargs) as response:
+                            response.raise_for_status()
                     # Servers drop the connection on HEAD requests, causing
                     # ConnectionError.
                     except (ConnectionError, HTTPError, TooManyRedirects) as err:
@@ -338,10 +337,9 @@ class HyperlinkAvailabilityCheckWorker(Thread):
                             raise
                         # retry with GET request if that fails, some servers
                         # don't like HEAD requests.
-                        response = requests.get(req_url, stream=True,
-                                                config=self.config,
-                                                auth=auth_info, **kwargs)
-                        response.raise_for_status()
+                        with requests.get(req_url, stream=True, config=self.config,
+                                          auth=auth_info, **kwargs) as response:
+                            response.raise_for_status()
             except HTTPError as err:
                 if err.response.status_code == 401:
                     # We'll take "Unauthorized" as working.
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -321,8 +321,8 @@ class HyperlinkAvailabilityCheckWorker(Thread):
                         response.raise_for_status()
                         found = check_anchor(response, unquote(anchor))
 
-                        if not found:
-                            raise Exception(__("Anchor '%s' not found") % anchor)
+                    if not found:
+                        raise Exception(__("Anchor '%s' not found") % anchor)
                 else:
                     try:
                         # try a HEAD request first, which should be easier on
--- a/tests/test_build_linkcheck.py
+++ b/tests/test_build_linkcheck.py
@@ -31,6 +31,9 @@ class DefaultsHandler(http.server.BaseHTTPRequestHandler):
         if self.path[1:].rstrip() == "":
             self.send_response(200, "OK")
             self.end_headers()
+        elif self.path[1:].rstrip() == "anchor.html":
+            self.send_response(200, "OK")
+            self.end_headers()
         else:
             self.send_response(404, "Not Found")
             self.end_headers()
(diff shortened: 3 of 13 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -318,6 +318,7 @@ def check_uri() -> tuple[str, str, int]:
                     # Read the whole document and see if #anchor exists
                     response = requests.get(req_url, stream=True, config=self.config,
                                             auth=auth_info, **kwargs)
+                    response.close()  # no HTTP body reads required; close the response
```
jayaddison: Reading the content _is_ required here, as part of `check_anchor`.

(so a context-manager approach might be more appropriate here)

jayaddison:
Converting to draft status while learning more about the possible effects on local socket re-use (the linkchecker is likely to contact a host multiple times while checking URL/anchor validity, and so it'd be worthwhile to be cautious about changes here).

sphinx/builders/linkcheck.py
jayaddison: A better explanation, because I forgot what was going on here:

In the general case, we want to ensure that `response.raise_for_status` and `response.close` are both called after receiving a response from the server.  They provide error-checking and resource cleanup, respectively.

Most of the time, we can call `close` as soon as the HTTP status line (200 OK, etc) is received.

The anchor-checking case (looking for a specific HTML ID within a documentation webpage) is different though: it doesn't seem to make sense to anchor-check on error responses (HTTP 4xx), even though theoretically the server _could_ include HTML and the anchors we're looking for in the response.  Documentation references shouldn't be served from pages that produce 4xx responses.

So in that case we want to call `response.raise_for_status` first.

It _might_ be possible to move the `response.close` call into the `finally` clause of the `try...except` handler.

The context manager approach provided by the `requests` library calls `response.close` when the `response` context-manager exits, though: that's convenient and could be easier than writing a `finally` clause that has to handle multiple different arrival paths.

tests/test_build_linkcheck.py
```diff
@@ -95,6 +101,15 @@ def test_defaults(app):
...
     assert "Not Found for url: http://localhost:7777/image.png" in rowsby["http://localhost:7777/image.png"]["info"]
+    # anchor should be found
+    assert rowsby['http://localhost:7777/anchor.html#found'] == {
+        'filename': 'links.rst',
+        'lineno': 14,
+        'status': 'working',
+        'code': 0,
+        'uri': 'http://localhost:7777/anchor.html#found',
+        'info': '',
+    }
```
francoisfreitag, line 112: Isn’t that case covered by https://github.com/sphinx-doc/sphinx/blob/30f347226fa4ccf49b3f7ef860fd962af5bdf4f0/tests/roots/test-linkcheck/links.txt#L5

francoisfreitag (dismissed):
Otherwise LGTM.

Side note, if we want to speedup linkcheck tests, we may replace links going to servers on the internet with going to the local HTTP server, its latency should be lower and very likely faster.

tests/test_build_linkcheck.py
jayaddison: Thanks for the review, @francoisfreitag.

> Isn’t that case covered by
> 
> https://github.com/sphinx-doc/sphinx/blob/30f347226fa4ccf49b3f7ef860fd962af5bdf4f0/tests/roots/test-linkcheck/links.txt#L5

Although the linkchecker should look for the anchor element when checking that URL, I don't think that there's test coverage on it.

There are a couple of [`test_defaults*` test cases](https://github.com/sphinx-doc/sphinx/blob/30f347226fa4ccf49b3f7ef860fd962af5bdf4f0/tests/test_build_linkcheck.py#L27-L104) -- but they don't (didn't?) check any positive-found anchor cases.

> Side note, if we want to speedup linkcheck tests, we may replace links going to servers on the internet with going to the local HTTP server, its latency should be lower and very likely faster.

Yep, a good idea.  Commit 97f07ca83c70bba55fe1c1cb25993e5240e4187d should handle that.  It does look like there are still [two (broken link) public URLs](https://github.com/sphinx-doc/sphinx/tree/97f07ca83c70bba55fe1c1cb25993e5240e4187d/tests/roots/test-linkcheck-documents_exclude) in the test data -- but those [should be `ignored`](https://github.com/sphinx-doc/sphinx/blob/97f07ca83c70bba55fe1c1cb25993e5240e4187d/tests/test_build_linkcheck.py#L696-L713) according to the relevant linkchecker test.

francoisfreitag: Hum... The reason why https://www.sphinx-doc.org/en/master/usage/installation.html#overview is not verified seems to be malformed RST:
`/tmp/pytest-of-freitafr/pytest-12/linkcheck/links.txt:5: ERROR: Unknown target name: "https://www.sphinx-doc.org/en/master/usage/installation.html#overview".`

I believe it should either be a [standalone hyperlink](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#standalone-hyperlinks) or a name should be added, e.g.
``` 
`install overview <https://www.sphinx-doc.org/en/master/usage/installation.html#overview>`_
```

jayaddison: Well spotted :+1: 

Since the Internet URLs have been replaced by `localhost` in the development branch recently, it feels like it may be easiest to add equivalent anchor-reference coverage using a `localhost` URL.  What do you think?

francoisfreitag: Works for me, thanks!

jayaddison: Thank you :+1: 

I think I've stacked too many tasks up for myself recently, so I'm worried about creating more distractions for myself: but... this conversation has me wondering whether `sphinx-lint` could detect malformed URLs like that (if so: then we could add it to some of our continuous integration checks).

francoisfreitag: I believe this would be a great addition. Don’t feel like you should take it all by yourself, adding an issue is a good way to keep track of ideas for improvement, and opens the door for others to tackle the issue.

jayaddison: Filed as sphinx-contrib/sphinx-lint#62.

jayaddison:
@AA-Turner this branch should resolve the linkcheck test flakiness short-term, until I can figure out and clear up the mess I'm making of #11340.

.github/workflows/main.yml
```diff
@@ -13,7 +13,7 @@ env:
   FORCE_COLOR: "1"
   PYTHONDEVMODE: "1"  # -X dev
   PYTHONWARNDEFAULTENCODING: "1"  # -X warn_default_encoding
-  PYTHONWARNINGS: "error,always:unclosed:ResourceWarning::"  # default: all warnings as errors, except ResourceWarnings about unclosed items
+  PYTHONWARNINGS: "error"
```
AA-Turner:
```suggestion
  PYTHONWARNINGS: "error"  # default: all warnings as errors
```

AA-Turner:
Thanks!

jayaddison:
Thank you, @AA-Turner!

PR #11320, opened 2023-04-12
Use ``overwrite_file`` context manager in ``test_ext_autodoc_configs``

marxin:
The tests modify source files (index.rst), that is not restored and thus another test can reach a modified source file and it leads to unexpected test results.

Related: #11285 

@AA-Turner

### DIFF
--- a/tests/test_ext_autodoc_configs.py
+++ b/tests/test_ext_autodoc_configs.py
@@ -2,6 +2,7 @@
 
 import platform
 import sys
+from contextlib import contextmanager
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


AA-Turner:
Perhaps if ``docutilsconf`` is set we could also clean it up in ``SphinxTestApp.cleanup()``?

A

AA-Turner:
Thanks!

marxin:
> Perhaps if `docutilsconf` is set we could also clean it up in `SphinxTestApp.cleanup()`?

Sure, done in #11345.

Issue #11299, opened 2023-04-07
Flaky unit tests: linkcheck builder (tests/test_build_linkcheck.py)


jayaddison:
### Describe the bug

The [`test_linkcheck_allowed_redirects`](https://github.com/sphinx-doc/sphinx/blob/4659fc21a92c3ff8c7019b56bf2e519bb5583ce5/tests/test_build_linkcheck.py#L340-L365) test appears to be failing rarely and intermittently:

- Sometimes an unexpected `broken` result is received instead of `working`:
  - https://github.com/sphinx-doc/sphinx/actions/runs/4630653760/jobs/8192510588#step:10:2378
  - https://github.com/sphinx-doc/sphinx/actions/runs/4636993886/jobs/8205438908#step:10:1897
- Sometimes a timeout occurs:
  - https://github.com/sphinx-doc/sphinx/actions/runs/4628948378/jobs/8188651117#step:10:2393

Other tests within the same [`linkcheck` builder test module](https://github.com/sphinx-doc/sphinx/blob/4659fc21a92c3ff8c7019b56bf2e519bb5583ce5/tests/test_build_linkcheck.py) also appear to be affected:

- `test_too_many_requests_retry_after_HTTP_date` https://github.com/sphinx-doc/sphinx/actions/runs/4659078989/jobs/8245535037?pr=11312 (thanks, @picnixz)
- `test_raw_node` https://github.com/jayaddison/sphinx/actions/runs/4697985339/jobs/8329708670?pr=1#step:10:2447

It's possible that they're slightly separate issues, but are likely to be related (and probably can be fixed at the same time).

### How to Reproduce

Run the relevant [`main.yml` continuous integration `pytest` tests](https://github.com/sphinx-doc/sphinx/blob/4659fc21a92c3ff8c7019b56bf2e519bb5583ce5/.github/workflows/main.yml#L19-L68) on a repeated basis.  This is demonstrated by jayaddison/sphinx#1.

### Environment Information

```text
GitHub Actions workflows (examples are linked in the bug description section)
```

### Sphinx extensions

```python
N/A
```

picnixz:
First, it seems that it only affects Python 3.11 and Python 3.12 builds. One possibility that comes up to me is we *might* have some race conditions or some deadlock. This may be due to the new implementation of the HTTP server with filelocks but I am not entirely sure about it. 

[...]

PR #11323, opened 2023-04-13
test_linkcheck_allowed_redirects: enable HTTP HEAD request support

jayaddison:
This is intended to reduce the number of HTTP requests within this particular test to closer-to-the-norm compared to other tests within the [`test_build_linkcheck.py` module](https://github.com/sphinx-doc/sphinx/blob/509cc4533c92adae92997d63c8a7b11295f9e207/tests/test_build_linkcheck.py); this test has been [failing intermittently](https://github.com/sphinx-doc/sphinx/issues/11299) and it seems possible that timeouts due to HTTP roundtrip traffic are a contributing factor.

### Feature or Bugfix
- Bugfix (maybe)

### Purpose
- Intended to reduce flakiness in the `test_linkcheck_allowed_redirects` test -- or to rule out another potential source of that problem
- Also intended to avoid increasing the `linkcheck` request timeout thresholds for this (and other) tests

### Detail
- Should reduce the number of requests made by the unit test to the test HTTP server by half.  The server is tiny and basic, and I don't exactly understand why reducing the total request count should have any effect on the duration of the individual requests - but it's an idea to explore, and the logic expressed by the test should remain valid.

### Relates
- Relates to #11299

### DIFF
--- a/tests/test_build_linkcheck.py
+++ b/tests/test_build_linkcheck.py
@@ -342,7 +342,7 @@ def test_follows_redirects_on_GET(app, capsys, warning):
                         'linkcheck_allowed_redirects': {'http://localhost:7777/.*1': '.*'},
                     })
 def test_linkcheck_allowed_redirects(app, warning):
-    with http_server(make_redirect_handler(support_head=False)):
+    with http_server(make_redirect_handler(support_head=True)):
         app.build()
 
     with open(app.outdir / 'output.json', encoding='utf-8') as fp:


francoisfreitag (approved):

jayaddison:
Thanks for the review @francoisfreitag.  After doing more testing, I'm not really convinced that this would fix the test flakiness.

I'd be OK with it being merged.. but if there's no particular reason for it, then I'd probably tend towards closing it unmerged, since otherwise it'll pollute the commit history.

Either way, I'll run a few more tests in jayaddison/sphinx#1 and perhaps that'll discover more -- or perhaps not :)  That has been my Friday, 2023-04-14..

jayaddison:
(the approach I prefer now, since I couldn't find any particular reason for a performance regression, is simply to increase the timeout thresholds slightly further in #11326)

francoisfreitag:
My understanding of this change : allowing `HEAD` saves the linkcheck behavior of retrying with `GET` after `HEAD` failed, which makes the test faster. Since the method is not relevant for the test, might as well use `HEAD`.

jayaddison:
That is true that this change should provide a very, very small runtime improvement for the tests, yes.

jayaddison:
Nothing about the experimental evaluation data collected in jayaddison/sphinx#1 _disproved_ this fix -- that is, no test failures occurred in the `linkcheck` tests with this change applied -- but I still don't think that the value (reduced test runtime -- extremely marginal) is worth the additional investigation and history overhead.

I'll close this pull request -- if someone later feels that this change is particularly valuable (perhaps it turns out that actually it _does_ fix the problem comprehensively, somehow), then as long as there's supporting evidence for that, I've no problem with it being merged.

Roughly speaking: I don't mind being wrong, but I'd like to understand why.  In this case I feel like I'm wrong, but the data hasn't yet confirmed it.  So I'll choose not to proceed here.

PR #11331, opened 2023-04-15
continuous integration: resolve docutils installation step build failures

### LINKED ISSUE #11330: build failure: DeprecationWarning during docutils dependency install
### Describe the bug

The continuous integration unit test workflow uses an environment variable to treat [most warnings as errors](https://github.com/sphinx-doc/sphinx/blob/b6e6805f80ad530231cf841e689498005cf2bb96/.github/workflows/main.yml#L16) by default.

Recently, `pip` has begun emitting a [`DeprecationWarning` during the installation of `docutils`](https://github.com/sphinx-doc/sphinx/actions/runs/4708875860/jobs/8351594861#step:8:98) on versions of Python below 3.12, resulting in build failures.

I think that this may be related to https://github.com/pypa/pip/pull/11916, where the version of `setuptools` vendored by `pip` was updated (and the [relevant warning](https://github.com/pypa/pip/pull/11916/files#diff-d97f638babc116bdf28d8da57c10706aa71436891eecb56e48108f1e3a461a82R121) was introduced).

Overriding that particular warning message (while continuing to treat other warnings as errors) may be the safest and simplest way forward; I'll provide a pull request to do that.

### How to Reproduce

I think that any continuous integration builds for this repository that utilise `pip` 23.1 are likely to be affected.

When opening a pull request to include the fix, I'll begin with something like a no-op commit to demonstrate the expected build failure, and after that a subsequent commit containing the fix (and hopefully illustrating that the build passes with that in place).

### Environment Information

```text
N/A
```

### Sphinx extensions

```python
N/A
```

### Additional context

_No response_

jayaddison:
### Feature or Bugfix
- Maintenance

### Purpose
- Restore continuous integration test success.

### Detail
- `pip`'s vendored libraries have begun warning about the deprecation of `pkg_resources`
- Our build stages intentionally treat warnings as errors by default
- This change temporarily permits ~~this specific warning~~ deprecation warnings to occur and be reported-but-not-considered-failure.
- The first commit on this branch will attempt to replicate the problem, and the second commit will attempt to resolve it.

### Relates
- Resolves #11330.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,7 +60,7 @@ jobs:
       run: python -m pip install --upgrade "docutils==${{ matrix.docutils }}.*"
       if: "!endsWith(matrix.python, '-dev')"
       env:
-        PYTHONWARNINGS: "error,default:pkg_resources is deprecated:DeprecationWarning"
+        PYTHONWARNINGS: "error,default::DeprecationWarning"
     - name: Install Docutils ${{ matrix.docutils }} (ignore warnings)
       run: python -m pip install --upgrade "docutils==${{ matrix.docutils }}.*"
       if: "endsWith(matrix.python, '-dev')"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,7 +60,7 @@ jobs:
       run: python -m pip install --upgrade "docutils==${{ matrix.docutils }}.*"
       if: "!endsWith(matrix.python, '-dev')"
       env:
-        PYTHONWARNINGS: "error,default:pkg_resources is deprecated:DeprecationWarning"
+        PYTHONWARNINGS: "error,default:pkg_resources is deprecated:DeprecationWarning,default:pkg_resources.declare_namespace:DeprecationWarning"
     - name: Install Docutils ${{ matrix.docutils }} (ignore warnings)
       run: python -m pip install --upgrade "docutils==${{ matrix.docutils }}.*"
       if: "endsWith(matrix.python, '-dev')"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,7 +60,7 @@ jobs:
       run: python -m pip install --upgrade "docutils==${{ matrix.docutils }}.*"
       if: "!endsWith(matrix.python, '-dev')"
       env:
-        PYTHONWARNINGS: "error,default:pkg_resources is deprecated:DeprecationWarning::,default:deprecated call.*pkg_resources.declare_namespace:DeprecationWarning::"
+        PYTHONWARNINGS: "error,default:pkg_resources is deprecated:DeprecationWarning::,default:Deprecated call to `pkg_resources.declare_namespace:DeprecationWarning::"
     - name: Install Docutils ${{ matrix.docutils }} (ignore warnings)
       run: python -m pip install --upgrade "docutils==${{ matrix.docutils }}.*"
       if: "endsWith(matrix.python, '-dev')"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,7 +60,7 @@ jobs:
       run: python -m pip install --upgrade "docutils==${{ matrix.docutils }}.*"
       if: "!endsWith(matrix.python, '-dev')"
       env:
-        PYTHONWARNINGS: "error,default:pkg_resources is deprecated:DeprecationWarning::,default:pkg_resources.declare_namespace:DeprecationWarning::"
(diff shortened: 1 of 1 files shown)


jayaddison:
Lowering the severity of the specific `pkg_resources is deprecated as an API` message was insufficient to allow the (non-development) `docutils` package install to succeed with warnings-as-errors; the third commit adds further lenience, allowing any deprecation warnings during (again, non-dev-version) `docutils` package installation to succeed.

picnixz:
Instead of suppressing some warnings as errors, why not:

- freeze the pip version, or
- filter the exact DeprecationWarning causing the issue instead of silencing all DeprecationWarning exceptions ? (if possible)

jayaddison:
Thanks @picnixz - reasonable suggestions.  Replying inline:

> Instead of suppressing some warnings as errors, why not:
> 
>     * freeze the pip version, or

This doesn't appear to be a bug/problem with `pip`, it's a message for our information, so I don't think we need to pause upgrades to the `pip` version we're building with.

>     * filter the exact DeprecationWarning causing the issue instead of silencing all DeprecationWarning exceptions ? (if possible)

That's worth more consideration, yep.  I attempted a precise, single-message filter in 671bc0edb3cff80ebc7bd490f5926489990c4f85, but it turns out that there was a [second deprecation warning](https://github.com/sphinx-doc/sphinx/actions/runs/4709867363/jobs/8353190335#step:8:96) (also `pkg_resources`-related):

```
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

Perhaps it would be better to provide filters for both of those messages.

I did briefly consider a message filter on the single-word string `pkg_resources`, to catch both - but it felt wrong somehow to use a single message filter to attempt to handle two not-directly-related warnings.

jayaddison:
Ah, an important detail from the documentation explaining why the message-filtering isn't working as-expected as of 09984f4c89ccbfa208ff9925e5747d8594983970:

> In [-W](https://docs.python.org/3/using/cmdline.html#cmdoption-W) and [PYTHONWARNINGS](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS), message is a literal string that the start of the warning message must contain (case-insensitively), ignoring any whitespace at the start or end of message.

- :heavy_check_mark: `pkg_resources is deprecated` matches from the start of `pkg_resources is deprecated as an API`
- :negative_squared_cross_mark: `pkg_resources.declare_namespace` _does not_ match from the start of <code>Deprecated call to `pkg_resources.declare_namespace('sphinxcontrib')`</code>

jfbu:
@jayaddison trying out at https://github.com/jfbu/sphinx/commit/b95fc0e88b6ea9f869a7cf014e80b335d6e05b55 (it worked in my local testing).

jfbu:
@jayaddison Can you cherry-pick https://github.com/jfbu/sphinx/commit/6bbe3f62a1ba946e96531226df130ac84348f115 it lets CI pass

jayaddison:
Ah, thanks @jfbu!  I was experimenting with something similar.  Since yours is proven working, let's go with that :)

jayaddison:
And for completeness to explain why 40e547adc87c77c8255b7367a136be6a11b5d714 _didn't_ work:

The documentation mentions that regular expressions are supported during some warning filtering:

> message is a string containing a regular expression that the start of the warning message must match, case-insensitively.

... but that excludes `PYTHONWARNINGS` environment variables (and I should have noticed that in the text I quoted previously) - there the message filter is evaluated as a literal string only:

> In [-W](https://docs.python.org/3/using/cmdline.html#cmdoption-W) and [PYTHONWARNINGS](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS), message is a literal string that the start of the warning message must contain (case-insensitively), ignoring any whitespace at the start or end of message.

jfbu (approved):
LGTM, thanks!

I hesitated about the added comment as we have very few in such files and I am not sure about our policy here about long lines, but well, this does not undergo linting! And it can always be adjusted later.

jfbu:
Thanks for contribution, I will merge this now because this helps other contributed PRs not end up receiving unjust and unfair red crosses :smile:

jayaddison:
Thanks again @jfbu.  About that [comment line](https://github.com/sphinx-doc/sphinx/blob/e5e115d72bbd0a768ab59d094a2ea6f18793a6c8/.github/workflows/main.yml#L16):

My hope is that we can resolve the `ResourceWarning` warnings (see #11317), and then to simplify that line and remove the comment at the same time.

jfbu:
Thanks for the documenting of the rationale behing the comment and other choices.  I streamlined the commit message and added hard linebreaks out of habit. Hope ok with you.

PR #11333, opened 2023-04-17
LaTeX: Fix  'multiply-defined references' PDF build warning when one or more reST labels directly precede a ``py:module`` or ``automodule`` directive

### LINKED ISSUE #11093: LaTeX: reST label before a ``module`` (or ``automodule``) sphinx.ext.autodoc directive causes pdf build to report multiply defined reference
### Describe the bug

As said in the title, the PDF build reports multiply-defined references.

### How to Reproduce

Use 
```
import os
import sys
sys.path.insert(0, os.path.abspath('.'))

extensions = ['sphinx.ext.autodoc']
```

in `conf.py`

file `index.rst`:
```rest

.. toctree::

   testfile
```

file `testfile.rst`:
```rest
.. _somelabel:

.. module:: testmodule

Some title
==========

Some text.

.. autofunction:: foo
```

file `testmodule.py`
```python
"""not much here
"""

def foo():
    "I do not do much and simply return ``None``"

    return None

```
(the  autofunction part is not part of the problem).

Then run `make latexpdf`.

Console output ends with
```
Latexmk: ====List of undefined refs and citations:
  Label `testfile:somelabel' multiply defined
```

This is not a serious problem (meaning it does not cause build failure).

But it happens in real life, see https://github.com/numpy/numpy/issues/22930#issuecomment-1371989684 and the 
```
LaTeX Warning: Label `reference/maskedarray.generic:maskedarray-generic' multip
ly defined.
```
is an instance of the present issue.  I have not yet checked but expect the others reported there as well.

### Environment Information

```text
Platform:              darwin
Python version:        3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)])
Sphinx version:        6.2.0+/6e6b9ad45
Docutils version:      0.19
Jinja2 version:        3.1.2
Pygments version:      2.14.0
```

### Sphinx extensions

```python
``["sphinx.ext.autodoc"]``
```

Edit: all NumPy instances from building their `numpy-ref.pdf` are of this nature: https://github.com/numpy/numpy/issues/22930#issuecomment-1372520387

picnixz:
Fix #11093

### Discussion

In the future, a different approach could be envisaged, similar to the one taken by the `texinfo` builder. For now, I decided implementing the minimal number of required changes.

@jfbu (for the ping)


CHANGES
```diff
@@ -73,6 +73,8 @@ Features added
 Bugs fixed
 ----------
 
+* #11093: LaTeX: fix build warnings when one or more reST labels directly preceed an
+  :rst:dir:`py:module` or :rst:dir:`automodule` directive. Patch by Bénédikt Tran (picnixz)
```
jfbu: Can you please change this to reflect the edited title of the pull request and reflow to not exceed 85 chars per line.

Also, please change commit message to "Update CHANGES for PR #11333"

tests/test_build_latex.py
```diff
@@ -1708,3 +1708,30 @@ def test_copy_images(app, status, warning):
...
+    def count_label(name):
+        text = r'\phantomsection\label{\detokenize{%s}}' % name
+        return content.count(text)
+
+    pattern = r'\\phantomsection\\label\{\\detokenize\{index:label-(?:auto-)?\d+[a-z]*}}'
+    expect_labels = {match.group() for match in re.finditer(pattern, content)}
+    result_labels = set()
+
+    # iterate over the (explicit) labels in the corresponding index.rst
+    for rst_label_name in {
```
jfbu: I would rather use a list here.  Else the order is not specified and imagine test fails at some point. Then you apply a wrong fix.  Then test fails for another duplicate label, so you think you have fixed a first issue but perhaps not another one, whereas in fact you have fixed nothing. Looks to me simpler if failures are triggered in the listed order.

tests/test_build_latex.py
```diff
@@ -1708,3 +1708,30 @@ def test_copy_images(app, status, warning):
...
+        'label_1a', 'label_1b', 'label_2', 'label_3',
+        'label_auto_1a', 'label_auto_1b', 'label_auto_2', 'label_auto_3',
+    }:
+        tex_label_name = 'index:' + rst_label_name.replace('_', '-')
+        tex_label_code = r'\phantomsection\label{\detokenize{%s}}' % tex_label_name
+        assert content.count(tex_label_code) == 1, f'duplicated label: {tex_label_name!r}'
+        result_labels.add(tex_label_code)
+
+    # sort the labels for a better visual diff, if any
+    assert sorted(result_labels) == sorted(expect_labels)
```
jfbu: In my usage of tox which is not quite same as our CI, this does not help much. Why not simply test  with len()? As the for loop succeeded we know our labels are there without duplicate, so only thing possibly wrong is if mysteriously there are other labels in output matching the regex.

tests/test_build_latex.py
```diff
@@ -1708,3 +1708,30 @@ def test_copy_images(app, status, warning):
...
+    app.build()
+    content: str = (app.outdir / 'python.tex').read_text()
+
+    def count_label(name):
+        text = r'\phantomsection\label{\detokenize{%s}}' % name
+        return content.count(text)
+
+    pattern = r'\\phantomsection\\label\{\\detokenize\{index:label-(?:auto-)?\d+[a-z]*}}'
+    expect_labels = {match.group() for match in re.finditer(pattern, content)}
+    result_labels = set()
```
jfbu: The naming is counterintuitive to me. I would rename `expect_labels` to `output_labels`(or even `result_labels`!) and `result_labels` to `input_labels`.  Or perhaps `labels_in_output` for the former and `labels_from_input` for the latter.

jfbu (commented):
LGTM with nits

The test is very thorough... see my comments about some details.

Thanks for contribution!

CHANGES
jfbu: And move this after next one so that issue numbers are ordered

tests/test_build_latex.py
picnixz: Oh that's my bad. I totally swapped the two names ! To follow your advice I used `output_labels` and `tested_labels`.

tests/test_build_latex.py
picnixz: Actually, that's what I wanted to do, namely checking that all labels have been correctly checked. 

The reason why I used `sorted` is to have a diff which tells me which value is in the LHS but not in the RHS (or vice-versa). But this *may* actually be a feature for PyCharm only. I'll change that into a `len` check then.

picnixz: I confirmed that this is *NOT* a PyCharm feature but a tox feature.

__**With `len` checks**__

```text
>       assert len(tested_labels) == len(output_labels)
E       AssertionError: assert 7 == 8
E        +  where 7 = len({'\\phantomsection\\label{\\detokenize{index:label-1a}}', '\\phantomsection\\label{\\detokenize{index:label-1b}}', '\\phantomsection\\label{\\detokenize{index:label-2}}', '\\phantomsection\\label{\\detokenize{index:label-auto-1a}}', '\\phantomsection\\label{\\detokenize{index:label-auto-1b}}', '\\phantomsection\\label{\\detokenize{index:label-auto-2}}', ...})
E        +  and   8 = len(frozenset({'\\phantomsection\\label{\\detokenize{index:label-1a}}', '\\phantomsection\\label{\\detokenize{index:label-1b}}', '\\phantomsection\\label{\\detokenize{index:label-2}}', '\\phantomsection\\label{\\detokenize{index:label-3}}', '\\phantomsection\\label{\\detokenize{index:label-auto-1a}}', '\\phantomsection\\label{\\detokenize{index:label-auto-1b}}', ...}))

tests/test_build_latex.py:1738: AssertionError
```

__**With `sorted` checks**__

```txt

E         Full diff:
E           [
E            '\\phantomsection\\label{\\detokenize{index:label-1a}}',
E            '\\phantomsection\\label{\\detokenize{index:label-1b}}',
E            '\\phantomsection\\label{\\detokenize{index:label-2}}',
E         -  '\\phantomsection\\label{\\detokenize{index:label-3}}',
E            '\\phantomsection\\label{\\detokenize{index:label-auto-1a}}',
E            '\\phantomsection\\label{\\detokenize{index:label-auto-1b}}',
E            '\\phantomsection\\label{\\detokenize{index:label-auto-2}}',
E            '\\phantomsection\\label{\\detokenize{index:label-auto-3}}',
E           ]
```

CHANGES
picnixz: Actually, I have a question about this. It appears that some LaTeX issues are sorted, but where should I put fixes related to other components ? (for previous versions, it appears that the order is arbitrary, whatever the issue is related to)

tests/test_build_latex.py
jfbu: Indeed you are right, I badly tested your test here.  And to reproduce the output I had to modify tox.ini to use pytest -vv.

This final assert tests a rather exotic situation which is not the original one about duplicated labels; as a rule many tests are rather minimal in our test suite, and this one really goes to some length to cover all possibilities ;-) ok let's go with it.

picnixz:
Oh there is a typo in the CHANGES (preceed instead of precede). Let me fix this before merging

CHANGES
jfbu: Yes the order used to be simply that entries where added chronologically, then at some point we started grouping together entries according to builder, and then more recently still I started ordering the LaTeX entries, and only them, as at some point a few weeks ago I did lots of LaTeX work in view of 6.2.0, this felt good.

jfbu:
> Oh there is a typo in the CHANGES (preceed instead of precede). Let me fix this before merging

ah you had not copied pasted the edited PR title then... tsss...

CHANGES
```diff
@@ -75,6 +75,9 @@ Bugs fixed
 
 * #11079: LaTeX: figures with align attribute may disappear and strangely impact
   following lists
+* #11093: LaTeX: fix "multiply-defined references" PDF build warnings when one or
+  more reST labels directly precede an :rst:dir:`py:module` or :rst:dir:`automodule`
```
jfbu, line 79: I am not native speaker so I don't know if we should use singular `precedes` or plural as here

jfbu (approved):
Thanks for contribution.

I will merge soon.

CHANGES
picnixz: Well, according to https://english.stackexchange.com/questions/13284/which-is-correct-one-or-more-is-or-one-or-more-are, the plural form seems more popular (and I am not a native speaker neither).

jfbu: ok at least we have some argument now! I will merge after testing completes thanks in particular for very thorough added test on this occasion

CHANGES
jfbu: > Yes the order used to be simply that entries where added chronologically,

Sorry for added post-merge noise but to clarify I meant here that new entries used to be added below older ones so the changes are predominantly listed from earliest to latest from top to bottom at each release.

Issue #11282, opened 2023-04-03
Cross-reference take most of time when building large package with a probable bug in make_xref


junshi356:
### Describe the bug

Building large packages takes a lot of time, most of which is cost on resolving cross-references. 

When building a package of 70MB and 3399 files ([azure-mgmt-containerservice](https://pypi.org/project/azure-mgmt-containerservice/)), it cost 3615 seconds and 2806 seconds are cost on resolving cross reference.
![image](https://user-images.githubusercontent.com/40595015/229457468-7755765b-b72c-4255-a690-399f7e8c075d.png)

"Fuzzy" searching mode of [PythonDomain.find_obj](https://github.com/sphinx-doc/sphinx/blob/39fa276b360dc36953b3ba62e19d6a978cd5109c/sphinx/domains/python.py#L1346) method is iterating all loaded objects in the memory, when the package has a lot of objects and xref to resolve, the build process is very slow.

The `searchmode` parameter is designed to control whether using Fuzzy mode to resolve cross references, but I think it's not working under most cases (e.g. if I cross-reference the built-in `:class: str`, Sphinx will use fuzzy mode to match it). 

I think the problem is in [PyXrefMixin.make_xref](https://github.com/sphinx-doc/sphinx/blob/39fa276b360dc36953b3ba62e19d6a978cd5109c/sphinx/domains/python.py#L377), attribute 'refspecific' is set to True, and won't be changed whatever the reftarget string is. Is it the expected behavior? Update the node attributes according to "refspecific" in return value back seems more reasonable after [parsing reftarget](https://github.com/sphinx-doc/sphinx/blob/39fa276b360dc36953b3ba62e19d6a978cd5109c/sphinx/domains/python.py#L381).

And can we have a switch in config files to only find exact match and don't fuzzy search when building a package?

Screenshot:
Full iterating for fuzzy mode:
![image](https://user-images.githubusercontent.com/40595015/229457756-1d0c1aa5-a931-4864-bd26-1f8b4fdd33c1.png)

Seems should update the node attributes "refspecific" after `parse_reftarget`.
![image](https://user-images.githubusercontent.com/40595015/229469075-d20d50a7-88f8-4968-9300-109972b20858.png)

### How to Reproduce

Reproduce python script:
```
from sphinx.application import Sphinx

app = Sphinx(
        srcdir= './rst',
        confdir='./conf',
        outdir= './output',
        buildername='html',
        doctreedir='./doctrees'
    )
app.build()
```

[...]

PR #11338, opened 2023-04-18
Fix a ``MemoryError`` in ``sphinx.ext.intersphinx`` when using ``None`` or ``typing.*`` as inline type references.

### LINKED ISSUE #11337: [BUG] ``MemoryError`` in ``sphinx.ext.intersphinx`` and undesirable side-effects in domains.
### Describe the bug

As mentioned in https://github.com/sphinx-doc/sphinx/issues/11282#issuecomment-1511900129, I found that `sphinx.ext.intersphinx` has a serious bug when parsing a document with the following structure:

```rst

.. py:function:: foo(x)
   
   The foo function.
   
   :param x: Some x.
   :type x: typing.Any

.. py:function:: bar(a)
   
   The bar function.
   
   :param a: Some a.
   :type a: typing.Any
```

As I said, here what happens during the resolution phase:

https://github.com/sphinx-doc/sphinx/blob/39fa276b360dc36953b3ba62e19d6a978cd5109c/sphinx/domains/python.py#L1328-L1331

In particular, when encountering `typing.*` or ``None`` in an inline reference such as `:type NAME: TYPE`, the latter will be considered as an `obj` reference (and not a `class` reference). This is due to

https://github.com/sphinx-doc/sphinx/blob/39fa276b360dc36953b3ba62e19d6a978cd5109c/sphinx/domains/python.py#L96-L100

Now, if `sphinx.ext.intersphinx` is enabled, we have:

https://github.com/sphinx-doc/sphinx/blob/5a6b2b16b6b588cdd4dab1b43b7873153da10baa/sphinx/ext/intersphinx.py#L355-L370

I found the following issues:

- If the role is `py:obj`, the corresponding object types are all possible object types. In particular, `'attribute'` and `'method'` coexist and `'method'` will be duplicated. This can lead to a ``MemoryError`` (I had a medium-size project where `object_types` grow more than of a hundred times so I can imagine when there are VERY large projects).
- In https://github.com/sphinx-doc/sphinx/blob/5a6b2b16b6b588cdd4dab1b43b7873153da10baa/sphinx/ext/intersphinx.py#L422-L429 the `objtypes` list is `domain._role2type[typ]`. This means that any modification to the former propagate to the latter, thereby causing an undesirable side-effect. Note that this **currently** seems to only affect the `py:obj` role *but* the side-effect can also be seen for other property-like roles.

With my above example, the domain's role dispatcher map *grows* as follows:

```python
# domain._role2type['obj'] when parsing foo()
['function', 'data', 'class', 'exception', 'method', 'classmethod', 'staticmethod', 'attribute', 'property', 'module']

# domain._role2type['obj'] when parsing bar()
['function', 'data', 'class', 'exception', 'method', 'classmethod', 'staticmethod', 'attribute', 'property', 'module', 'method']
```

### conf.py

```python
extensions = ['sphinx.ext.intersphinx']
intersphinx_mapping = {'python': ('https://docs.python.org/3',)}
``` 

### index.rst

```rst

.. py:function:: foo(x)
   
   The foo function.
   
   :param x: Some x.
   :type x: typing.Any

.. py:function:: bar(a)
   
   The bar function.
   
   :param a: Some a.
   :type a: typing.Any
```

### Environment Information

```text
Sphinx version:        6.2.0+/b04fe6f89
Docutils version:      0.18.1
Jinja2 version:        3.1.2
Pygments version:      2.15.0
```

### Sphinx extensions

```python
['sphinx.ext.intersphinx']
```

### Additional context

- #11282

picnixz:
Fix #11337 

### Discussion

I resisted making `domain.objtypes_for_role` return a `tuple` instead of a list to ensure that everything is read-only (it appears that occurrences of `objtypes` are only used as read-only, except in intersphinx) and expose a setter to update the corresponding object types given a specific role but I decided not to.

### Relates
- #11282


sphinx/ext/intersphinx.py
```diff
@@ -358,16 +358,21 @@ def _resolve_reference_in_domain(env: BuildEnvironment,
...
+    if domain.name == 'py' and 'attribute' in objtypes and 'method' not in objtypes:
         # properties are stored as py:method since Sphinx 2.1
-        objtypes.append('method')
+        extra_objtypes.append('method')
 
     # the inventory contains domain:type as objtype
-    objtypes = [f"{domain.name}:{t}" for t in objtypes]
+    domain_name = domain.name
+    objtypes = [f"{domain_name}:{t}" for t in objtypes]
+    objtypes.extend([f'{domain_name}:{t}' for t in extra_objtypes])
```
AA-Turner:
```suggestion
    objtypes = [f"{domain_name}:{typ}" for typ in objtypes + extra_objtypes]
```

AA-Turner:
I've updated to use dict-based set operations, to preserve order (I can't work out if order preservation is important to intersphinx---I don't think so, but I'm not confident).

Let me know thoughts on this approach!

A

sphinx/ext/intersphinx.py
```diff
@@ -410,30 +412,30 @@ def _resolve_reference(env: BuildEnvironment, inv_name: str | None, inventory: I
...
+
+    domain_name = node.get('refdomain')
+    if not domain_name:
+        # only objects in domains are in the inventory
+        return None
+    if honor_disabled_refs \
+            and (domain_name + ":*") in env.config.intersphinx_disabled_reftypes:
+        return None
+    domain = env.get_domain(domain_name)
+    objtypes = {}.fromkeys(domain.objtypes_for_role(typ)).keys()
```
picnixz: We could defer the creation of the dictionary:

```python
_objtypes = domain.objtypes_for_role(typ)
if not _objtypes:
    return None
objtypes = {}.fromkeys(_objtypes).keys()
```

AA-Turner: This was included in https://github.com/sphinx-doc/sphinx/pull/11338/commits/6a9e617a49041b042d1666ba49a726f06aa46601, if I understand you correctly!

A

PR #11343, opened 2023-04-20
Partially revert "Disable localisation when SOURCE_DATE_EPOCH is set (#10949)"

AA-Turner:
Alternative to @jayaddison's #11306.

The comment at https://github.com/sphinx-doc/sphinx/issues/9778#issuecomment-1501172176 posits that the reproducability issue is already fixed, but the ``objects.inv`` reproducability test is failing -- @jayaddison any suggestions?

A


jayaddison:
@AA-Turner I think that I miswrote the test, based on incorrect assumptions.

It says: the [EN and ET inventories should be identical in a reproducible build](https://github.com/sphinx-doc/sphinx/blob/f82c3c99126e644125d243ba0d0788197401e416/tests/test_util_inventory.py#L137-L138), but what that means is: check that localisation is disabled.

That's not a viable fix, in my opinion.  Ideally we want localisation to continue to be supported: we can't rule-out the system build locale as a variance factor in the output until there is a way to configure and build for multiple locales (essentially: ignore the system build locale, and build for all/a configured list of locales, producing identical results each time).

In short: I don't feel comfortable disabling localisation in reproducible-builds builds entirely - it seems like the wrong approach.

jayaddison:
Thanks @AA-Turner!

PR #11350, opened 2023-04-22
linkcheck tests: use thread-safe collections.deque for client header recording

### LINKED ISSUE #11348: Flaky unit test: test_auth_header_uses_first_match
### Describe the bug

The `test_auth_header_uses_first_match` test seems to fail intermittently, particularly when threaded test HTTP servers are in use (as in #11340).

### How to Reproduce

Visible in continuous build output in jayaddison/sphinx#2.

### Environment Information

```text
N/A
```

### Sphinx extensions

```python
N/A
```

### Additional context

_No response_

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- When threaded test HTTP servers are in use, as in #11340, then client header related tests have been failing intermittently.
- This pull request attempts to address that by using a thread-safe collection that is shared between the unit test (object owner) and server handler (object recipient/producer thread).

### Detail
- [`collections.deque`](https://docs.python.org/3/library/collections.html#collections.deque) is thread-safe and supports `append` and iteration, similar to the `list` object currently in-use.
- Tested as part of jayaddison/sphinx#2.

### Relates
- Resolves #11348.

### DIFF
--- a/tests/test_build_linkcheck.py
+++ b/tests/test_build_linkcheck.py
@@ -3,13 +3,13 @@
 from __future__ import annotations
 
 import base64
-from collections import deque
 import http.server
 import json
 import re
 import textwrap
 import time
 import wsgiref.handlers
+from collections import deque
 from datetime import datetime
 from os import path
 from queue import Queue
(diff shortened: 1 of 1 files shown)


jayaddison:
(please hold-off on merging until after the 6.2.0 release)

jayaddison:
This does _not_ seem like it'll provide a fix for #11348; see discussion in that issue for details.

PR #11363, opened 2023-04-26
Remove setuptools support

AA-Turner:


sebastinas:
@mitya57 Just FYI, this change breaks the build of around 40 packages in Debian. See https://codesearch.debian.net/search?q=sphinx.setup_command&literal=1

mitya57:
Thanks for the heads up. I usually ask Lucas for a mass rebuild and then he files bugs for all failing packages.

AA-Turner:
`setup.py` itself has now been deprecated for a while and should be avoided, from Sphinx's perspective it's irresponsible to continue to encourage use of the file. 

A

PR #11368, opened 2023-04-27
Add support for sub-modules in autosummary

gregory-shklover:
Subject: Adds support for "modules" template argument so one can generate recursive documentation. 

### Feature or Bugfix
- Feature

### Purpose
Adds support for generating recursive module documentation.
For each module, documents imported sub-modules and adds corresponding "modules" template parameter.
The parameter can be used in a module template like this:
  ...
  .. automodule:: {{ fullname }}
  
     {% block modules %}
     {% if modules %}
     .. rubric:: Modules
  
     .. autosummary::
        :toctree: {{ name }}
     {% for item in modules %}
        {{ item }}
     {%- endfor %}
     {% endif %}
     {% endblock %}
  ...


sphinx/ext/autosummary/generate.py
```diff
@@ -111,6 +112,19 @@ def _underline(title: str, line: str = '=') -> str:
...
+    """Test if specified module name is 'parent' or sub-module of 'parent'"""
+    return module is not None and (parent == module or module.startswith(parent + "."))
+
+
+def _obj_module(o: Any) -> str:
+    """Returns object module name"""
+    if isinstance(o, _types.ModuleType):
+        return o.__name__
+
+    return getattr(o, '__module__', None)
```
picnixz: I think it's safer to use `inspect.getmodule`. For instance, Python built-in singletons such as `Ellipsis` have no ``__module__`` but are still declared in the `builtins` module.

tests/test_ext_autosummary.py
```diff
@@ -306,8 +306,12 @@ def test_autosummary_generate_content_for_module_imported_members(app):
                                   'quuz', 'qux']
     assert context['functions'] == ['bar']
     assert context['all_functions'] == ['_quux', 'bar']
-    assert context['classes'] == ['Class', 'Foo']
-    assert context['all_classes'] == ['Class', 'Foo', '_Baz']
+    # changed the meaning of automosummary_imported flag to include only sub-modules
```
picnixz, line 309: It would be better to have a distinct test root where you automatically include submodules or when you don't. Additionally, you probably need to document the new template variable ``modules``, and give some examples (and probably add a CHANGES entry). 

It is said "changed the meaning of autosummary_imported flag", but what was changed and why was it changed ? Also why was the 'Class' detected before and not any more after your addition ?

gregory-shklover:
Converted to "draft" to resolve differences vs ":recursive:" feature in autosummary.

PR #11374, opened 2023-04-27
tests: linkcheck: add connection-count measurement

jayaddison:
### Feature or Bugfix
- Feature

### Purpose
- To help reviewing the acceptability of any implementations for #11324, we will want to know how many distinct linkchecker connections are being constructed before-and-after a change.

### Detail
- Adds an instrumentation class, `ConnectionMeasurement`, to the linkchecker unit test module that counts uniquely-constructed `urllib3` connection objects

### Relates
- Supporting feature for implementation of #11324


jayaddison:
Thank you!

Issue #11205, opened 2023-02-19
automation of linting test using pre-commit


12rambau:
I was working on a PR and I was surprised to see many linting errors.
I didn't see any documentation on which test I should run from my side (ruff, flak8 isort others). 

**Describe the solution you'd like**
Use pre-commits to run all these linting checks automatically and use the same pre-commits in the github tests for consistency

hugovk:
Another benefit is pinning the lint tools, so you don't get hit by sudden failures when linters release with add new checks.

onerandomusername:
> Another benefit is pinning the lint tools, so you don't get hit by sudden failures when linters release with add new checks.

Yeah, I noticed that ruff is unpinned and will install whatever the latest version is. That's going to cause a lot of errors with how often ruff has a release with new features.

There's over 100 lint failures on master right now.

12rambau:
I started to work on it with the combo: black/prettier/ruff and it works quite ok.
Problem is taht lots of tests are now failing because of the linting. That make the test non robust to lint which is (according to me) an issue, I'll see if adding "pytest-regressions" and "beatifulsoup" could save us

jdillard:
> Then we can update the versions and fix new warnings in a controlled manner in followups.

Is there a "best practice" for doing this? I wasn't sure if it is just setting a reminder or maybe a cron job that runs the latest version so you can monitor the number of new warnings and fix them before they get out of hand.

12rambau:
You can but do you really want to ? `ruff` and `black` are sending reminder in the pre-commit commands but to be fair once you set up something that works and more importantly that is consistent, you don't want to change it every month (unless you identify an error or want to add a new feature)

hugovk:
I use https://pre-commit.ci - it autofixes PRs, and sends autoupdate PRs every month (by default, but I set it to quarterly with this at the end of `.pre-commit-config.yaml`:

```yaml
ci:
  autoupdate_schedule: quarterly
```

Also doesn't stop me running `pre-commit autoupdate`, which I might do if I'm also updating other stuff in the config file.

12rambau:
I'll go step by step as all the linting operation will crash. step one black + pre-commit + github actions

hugovk:
This project uses tox, which is very well suited for adding an optional `tox -e lint` to run pre-commit.

For example:

[...]

PR #11375, opened 2023-04-27
apply ruff using pre-commit hook

12rambau:
I'm proposing to execute linting operation using pre-commit to make sure that people don't get bad surprises when creating a PR. 

### Feature or Bugfix
- Feature

### Relates
related to #11205 

Instead of getting hard stop as for blakc I would like to itterate on this one. First thing first here is a proposal on running ruff within pre-commits to make sure it's run everytime. 

proposal following commits:
- [ ] use a pre-commit action to run ruff in the lint CI 
- [ ] drop ruff from requirements (as it's living in the pre-commit env)
- [ ] drop isort and flake8 as they are meant to be replaced by ruff

@AA-Turner let me know which one you want to see and which you don't and I'll commit accordingly

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,28 +13,14 @@ env:
   FORCE_COLOR: "1"
 
 jobs:
-  ruff:
+  pre-commits:
     runs-on: ubuntu-latest
-
     steps:
-    - uses: actions/checkout@v3
-    - name: Set up Python
-      uses: actions/setup-python@v4
-      with:
-        python-version: 3
-    - name: Install pip
-      run: python -m pip install --upgrade pip
-
-    - name: Install known good Ruff
-      run: python -m pip install ruff==0.0.261
-    - name: Lint with known good Ruff
-      run: ruff . --diff --format github
-
-    - name: Install latest Ruff
-      run: python -m pip install --upgrade ruff
-    - name: Lint with Ruff
-      continue-on-error: true
-      run: ruff . --diff --format github
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - uses: pre-commit/action@v3.0.0
 
   flake8:
     runs-on: ubuntu-latest
@@ -101,7 +87,7 @@ jobs:
       run: >
         sphinx-lint
         --enable line-too-long
-        --max-line-length 85 
+        --max-line-length 85
         CHANGES
         CONTRIBUTING.rst
         README.rst
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+default_install_hook_types: [pre-commit]
+
+repos:
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: "v0.0.263"
+    hooks:
+      - id: ruff
+        stages: [commit]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,4 +5,3 @@ repos:
     rev: "v0.0.263"
     hooks:
       - id: ruff
-        stages: [commit]
--- a/doc/internals/contributing.rst
+++ b/doc/internals/contributing.rst
@@ -105,7 +105,7 @@ These are the basic steps needed to start developing on Sphinx.
        . ~/.venv/bin/activate
        pip install -e .
 
-#. install `pre-commit hooks <https://pre-commit.com/>`__::
+#. Install `pre-commit hooks <https://pre-commit.com/>`__::
 
        pre-commit install
 
--- a/doc/internals/contributing.rst
+++ b/doc/internals/contributing.rst
@@ -105,6 +105,10 @@ These are the basic steps needed to start developing on Sphinx.
        . ~/.venv/bin/activate
        pip install -e .
 
+#. install `pre-commit hooks <https://pre-commit.com/>`__::
+
+       pre-commit install
+
 #. Create a new working branch. Choose any name you like. ::
 
        git checkout -b feature-xyz
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,10 +236,10 @@ ignore = [
     "Q000",  # double quotes found but single quotes preferred
     "Q001",  # single quote docstring found but double quotes preferred
     # flake8-return
-    "RET503",  # Missing explicit `return` at the end of function able to return value
+    "RET503",  # missing explicit `return` at the end of function able to return value
     "RET504",  # unnecessary variable assignment before `return` statement
     "RET505",  # unnecessary `else` after `return` statement
-    "RET506",  # Unnecessary `else` after `raise` statement
+    "RET506",  # unnecessary `else` after `raise` statement
     # flake8-raise
     "RSE102",
     # Ruff-specific rules
@@ -254,14 +254,14 @@ ignore = [
     "S301",  # 'pickle' unsafe when loading untrusted data
     "S324",  # probable use of insecure hash functions
     "S603",  # `subprocess` call: check for execution of untrusted input
-    "S607",  # Starting a process with a partial executable path
+    "S607",  # starting a process with a partial executable path
     "S701",  # use autoescape=True for Jinja
     # flake8-simplify
     "SIM102", # nested 'if' statements
     "SIM103", # return condition directly
     "SIM105", # use contextlib.suppress
     "SIM108", # use ternary operator
-    "SIM114", # Combine `if` branches using logical `or` operator
+    "SIM114", # combine `if` branches using logical `or` operator
     "SIM115", # use context handler for opening files
     "SIM117", # use single 'with' statement
     # flake8-self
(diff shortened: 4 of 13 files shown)


pyproject.toml
```diff
@@ -235,8 +236,10 @@ ignore = [
     "Q000",  # double quotes found but single quotes preferred
     "Q001",  # single quote docstring found but double quotes preferred
     # flake8-return
+    "RET503",  # Missing explicit `return` at the end of function able to return value
     "RET504",  # unnecessary variable assignment before `return` statement
     "RET505",  # unnecessary `else` after `return` statement
+    "RET506",  # Unnecessary `else` after `raise` statement
```
danieleades: should this PR not just focus on the introduction of pre-commit, rather than also changing the lints?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,8 @@
+default_install_hook_types: [pre-commit]
+
+repos:
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: "v0.0.263"
+    hooks:
+      - id: ruff
+        stages: [commit]
```
danieleades: missing newline. might as well add the 'end of file fixer' hook?

danieleades: do you need to configure the 'stages'?

danieleades:
i'm for this change, but this PR would need to modify CI to use pre-commit. You'd also need to update the documentation

.pre-commit-config.yaml
12rambau: not compulsory if it's the default value. I can remove it if you want

pyproject.toml
12rambau: My objective is to limit the impact to the chore code to the minimum so instead of modifying sphinx sources I decided to add extra ignores. I think they could then be dropped 1 by 1 in follow-up PR when someone feel motivated enough. Again for stability reasons.

.pre-commit-config.yaml
12rambau: when it comes to non-python files I prefer to use prettier which include all the pre-commit linter hooks in one. I think it could deserve it's own PR as it leads to many changes. let me edit the file

12rambau:
@danieleades I think I adressed all your comments. I agree that isort and flake8 fate should be decided in a follow-up PR. Also note that 3.12 tests are failing and I'm not sure this PR have changed anything regarding them. let me know if I should change anything.

pyproject.toml
danieleades: is this necessary because pre-commit is installing a different version of ruff? why are extra ignores required because of the way ruff is being executed?

12rambau: 2 reasons: 
1. because ruff is adding and removing rules on new release and there was a very loose pin in the initial CI (that's also one of the issue reported in #11205), pinning the version in pre-commit is ensuring this will not happen again 
2. The code base has changed in main since I started this PR and some new code is making this one fail. reason why I added these 2 extra ignores. Note that I don't know how they were merged in the first place as SIM114 was manually ignored on many lines.

danieleades: As an aside, I think Sphinx's approach of applying all lints and selectively disabling them after breakages is pretty reckless. Should really be inverted.

12rambau: even worth when you know that some of them are not compatible (D212 - D213)

danieleades: Thoughts @AA-Turner ?

12rambau: Note that any changes to linter will break the the current PR and branches. My take on this one is to use several PRs:
1. set up the structure (pre-commit, pinned linter version) without changing the code
2. sanitize ruff rules (one by one to avoid breaking everything)
3. integrate mypy to tox ? 
4. see if we can apply other rules (black, prettier)

This PR is dedicated to 1.

pyproject.toml
```diff
@@ -250,12 +253,15 @@ ignore = [
     "S113",  # probable use of requests call without timeout
     "S301",  # 'pickle' unsafe when loading untrusted data
     "S324",  # probable use of insecure hash functions
+    "S603",  # `subprocess` call: check for execution of untrusted input
+    "S607",  # Starting a process with a partial executable path
```
picnixz: To be consistent in the comments, I would remove the capital "S" and make it lowercase (check the other ignored rules that you added as well).

.github/workflows/lint.yml
```diff
@@ -13,28 +13,14 @@ env:
...
-
-    - name: Install latest Ruff
-      run: python -m pip install --upgrade ruff
-    - name: Lint with Ruff
-      continue-on-error: true
-      run: ruff . --diff --format github
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
picnixz, line 22: What is the reason of (explicitly) requesting Python 3.10 ? is "python-version: 3" not sufficient ?

doc/internals/contributing.rst
```diff
@@ -105,6 +105,10 @@ These are the basic steps needed to start developing on Sphinx.
        . ~/.venv/bin/activate
        pip install -e .
 
+#. install `pre-commit hooks <https://pre-commit.com/>`__::
```
picnixz: "install" -> "Install"

Also, I think there should be only one "_" and not two.

pyproject.toml
picnixz: > Note that I don't know how they were merged in the first place as SIM114 was manually ignored on many lines.

I think that's because the ruff check only checks the diff and not old places. So if you add correct stuff, it won't show up other errors. To that end, we could add `args: ['--diff', '--format=github']` to the pre-commit config. If you do that, this would however mean that old codebased won't be ruffed but only new code will (and I think you won't need extra ignores).

.github/workflows/lint.yml
12rambau: I'm following the guidelines of the github actions: https://github.com/actions/setup-pytho

> The python-version input is optional. If not supplied, the action will try to resolve the version from the default .python-version file. If the .python-version file doesn't exist Python or PyPy version from the PATH will be used. The default version of Python or PyPy in PATH varies between runners and can be changed unexpectedly so we recommend always setting Python version explicitly using the python-version or python-version-file inputs.

3.10 seems like a reasonable choice to avoid bad surprises. Anything you would me to change ?

doc/internals/contributing.rst
12rambau: it's an anonymous target to avoid raising warnings during the build, any reason you want me to use a reusable target ?

12rambau:
>If you do that, this would however mean that old codebased won't be ruffed but only new code will (and I think you won't need extra ignores).

I prefer to test the whole codebase in the check and catch errors instead of adding extra args do you agree ?

.github/workflows/lint.yml
picnixz: Well, prior to this, we used `python-version: 3`, and I think using the same thing as before is good. Also, we used python because we wanted to install ruff using pip. But do we still need it ? You are not running any Python command in the pre-commit step so..

doc/internals/contributing.rst
picnixz: Oh sorry, that's right. Alternatively, you can put the link afterwards (like it is done for the rest of the document):

```rst
blabla `pre-commit`__::
  
    blablabla

.. __: https://pre-commit.com
```

(personally, I would leave it inline because the line is short enough).

picnixz:
> I prefer to test the whole codebase in the check and catch errors instead of adding extra args do you agree ?

I honestly think that this is will be more deterimental in a first phase. It would be better to first enforce new code to be pushed correctly. In a second time, we could review the whole codebase and check whether the linters pass or not. At least this would avoid introducing additional ignore (for me the "S60*" errors should not be suppressed so it may be good to (one day) suppress them locally one by one).

.github/workflows/lint.yml
12rambau: you need python to run the pre-commit action : https://github.com/pre-commit/action.
i I only set it to python 3 it will not be consistent on all runners (from what I understand) so I'll let you decide.

picnixz: With `python-version: 3`, runners use the latest cached version, usually the latest released version from Python (currently, Python 3.11). You won't be consistent *only* if you are omit the `python-version` field. But, pinning it to "latest major" (i.e. "3") is fine.

12rambau:
Is there anything I should change here to make this PR move forward ?

AA-Turner:
I don't use pre-commit, and don't see a substantial benefit from adopting it. What's the rationale here?

The most I would probably add is the configuration file, so that those who wish to use pre-commit may. I wouldn't change CI configuration to add indirection via pre-commit.

A

12rambau:
The rational is adding some rules that will be run automatically before commit so that lazy developers (like me) stop sending code to the distributed repository with linting issues. the pre-commits libs are run on isolated environment so don't collide with one another and you can remove linting lib from you or deps as they are dealt with by the pre-commit config file. 

Also for newcomers, there is literally nothing to understand as opposed to `tox` that needs a bit more work to use to its full potential. My personal experience is that I didn't managed in my other PR to create the correct env to lint the sphinx correctly (surely because I'm not good enough) but I find it relevant for an open-source project to be as welcoming as possible for beginner contributors.

Finally I find that there were some traction on my issue (https://github.com/sphinx-doc/sphinx/issues/11205) which is surprising for a packaging problem showing both the excitement of the community toward `pre-commit` and maybe the need for contributors to be helped with tools they know.

PR #11384, opened 2023-04-28
Remove alabaster default sidebars logic

AA-Turner:


AA-Turner:
I'm going to skip this removal for Sphinx 7.

A

PR #11349, opened 2023-04-22
linkcheck builder: ensure error response is removed-from-scope before fallback request 

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Ensures that the connection from the initial HTTP HEAD request may be re-used for the HTTP GET in the case of failures.

### Detail
- Initially discovered during #11299; temporarily worked-around by increased timeouts, but re-appeared during development of #11340.
- Tested as part of jayaddison/sphinx#2.

### Relates
- Resolves #11299.

PR #11392, opened 2023-05-03
linkcheck tests: test webservers: enable HTTP/1.1 protocol

jayaddison:
### Feature or Bugfix
- Refactoring

### Purpose
- In preparation for a fix for #11324 (use of session-based HTTP requests from the `linkcheck` builder), this change upgrades the communication protocol of the test webservers, allowing multiple requests to occur within a single HTTP connection.

### Detail
- To ensure that the ~~single-threaded~~ test webserver does not become blocked by the `linkcheck` client, this change includes the resource-cleanup (connection close) behaviour from #11318.
  - edit: outdated context; multi-threaded webserving has been enabled

### Relates
- Prepares for #11324.
- Builds upon #11318.


jayaddison:
I'm having difficulty replicating the unit test failures here (mostly timeouts) on Python 3.11.2.

To try to reduce the differences between my local env and the GitHub workflow, I've used `python -m pytest -vv --durations 25`, and have `graphviz` installed.

jayaddison:
After installing a 2.x version of `urllib3` ([not yet supported](https://github.com/psf/requests/issues/6432) by `requests`, but installed [during the tests](https://github.com/sphinx-doc/sphinx/actions/runs/4928871201/jobs/8807781660?pr=11392#step:7:125)), I'm able to replicate the timeout failures that occur during these tests.

Given the unsupported-mix between `requests` 2.x and `urllib3` 2.x, perhaps it would make sense to pin `urllib3<2.0.0`.

jayaddison:
(note: b89450cbea91b856d3aea753a15707061c6a6d01 is included here largely to help validate a compatibility issue described in #11419.  if it or a similar change is worth applying, then it'd probably be better to open that as a separate pull request with an independent changelog entry)

tests/utils.py
```diff
@@ -19,7 +19,7 @@
 class HttpServerThread(threading.Thread):
     def __init__(self, handler, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.server = http.server.HTTPServer(("localhost", 7777), handler)
+        self.server = http.server.ThreadingHTTPServer(("localhost", 7777), handler)
```
jayaddison, line 22: Note for reviewers: without threaded webservers, HTTP1.1's ability to leave connections open causes difficulty for the Sphinx unit tests in combination with `urllib3` commit https://github.com/urllib3/urllib3/commit/a80c248c34a77e106551bf997e726457b03f42b5 - a connection-pool thread safety fix, included from v2.0.0 of that library onwards.

AA-Turner (approved):

AA-Turner:
Looks like timeout errors on CI:

```
test_auth_header_uses_first_match

>       assert f"Authorization: Basic {encoded_auth}\n" in stdout
E       AssertionError: assert 'Authorization: Basic dXNlcjE6cGFzc3dvcmQ=\n' in ''

(           index: line    1) broken    http://localhost:7777/ - HTTPConnectionPool(host='localhost', port=7777): Read timed out. (read timeout=0.05)
```

jayaddison:
Can you link to an example build log?  That test is definitely troublesome, I'm not certain I've figured out exactly why it fails though (so any more context could be useful).

jayaddison:
One thing to try, based on conversation in #11348 (a thread about that exact same test case in combination with multithreaded HTTP1.1 webservers) -- setting `daemon_threads` to `False` for the webservers.

jayaddison:
If the `daemon_threads` adjustment didn't make a difference, then there is one other idea that I had, based on a sense that sharing the `records` variable between the unit test and webserver handler could be causing problems:

We could migrate the relevant tests to print captured headers to `stdout`/`stderr`, and then use the `pytest` `capsys` fixture to capture the output from there (a pipe) instead of using a shared variable.

The resulting test code would be a bit more concise and simpler (I drafted an approach to this locally yesterday).  The only thing I don't love about it is that printing an authorization header -- even in test code -- doesn't seem ideal (I'm being way-overcautious perhaps, but sometimes code gets copy-pasted from places it really shouldn't be.  perhaps creating a variable alias for `stderr` so that the code would be extremely unlikely to work as-is when pasted would somewhat mitigate that...).

jayaddison:
Another way, I suppose, would be to implement some authentication logic within the test handlers, and then to check the response status code during the test (no header-capturing required, and no secondary communication channel required).

jayaddison:
> Another way, I suppose, would be to implement some authentication logic within the test handlers, and then to check the response status code during the test (no header-capturing required, and no secondary communication channel required).

That could work easily for some of the tests - such as [`test_auth_header_uses_first_match`](https://github.com/sphinx-doc/sphinx/blob/d3c91f951255c6729a53e38c895ddc0af036b5b9/tests/test_build_linkcheck.py#L209-L223) - because only a single item of information is being communicated (authentication header is received as expected) -- but note that there are other tests (such as [`test_linkcheck_request_headers_default`](https://github.com/sphinx-doc/sphinx/blob/d3c91f951255c6729a53e38c895ddc0af036b5b9/tests/test_build_linkcheck.py#L276-L289)) that inspect other properties of the request.

I guess another way to phrase that is: in the tests that inspect request/header information, there is currently custom logic applied on a per-test basis.

Staying with the short-term goal here of moving the logical evaluation into the webserver handler, and communicating success/failure via HTTP status code, we could either:

* Build a mega-test-handler that can inspect any of the fields referenced from any of the tests, and is constructed with relevant constructor arguments to configure it appropriately for each test.
OR
* Accept a lambda or other callable into an adaptable-test-handler, that invokes the (optional?) callable to evaluate success/failure.

In the latter case, I think we should still implement authentication outside of the lambda/callable approach.  Mixing authentication and customizable request handling logic feels like a non-great pattern (and even though this is test code, we should try to structure things in a way that matches good practices).

I'll begin experimenting with some of that later today.

To recap the theory/goal of all this: I think that the flaky `test_auth_header_uses_first_match` results are something to do with the fact that the `records` list variable is being used in two separate thread contexts (within the HTTP test server, and within the unit test code).  Moving the evaluation logic that refers to `records` to purely within the test handler, and using the HTTP status code as a synchronization boundary and communication channel, I think should be more reliable.

jayaddison:
After thinking about the previous few comments ([one](https://github.com/sphinx-doc/sphinx/pull/11392#issuecomment-1549584575), [two](https://github.com/sphinx-doc/sphinx/pull/11392#issuecomment-1549736831), [three](https://github.com/sphinx-doc/sphinx/pull/11392#issuecomment-1551010147)) (apologies for the stream-of-consciousness) related to `test_auth_header_uses_first_match`, I've written up #11426 as a possible changeset.

The [test failure in `test_too_many_requests_user_timeout`](https://github.com/sphinx-doc/sphinx/actions/runs/5003267380/jobs/8964280355#step:9:1908) remains a puzzle to me; I hadn't seen that before.  Perhaps it was exposed when 0817ade877bdbc7a8bb213fa177e3b522788a03d was added?  Or is it unrelated?

Increasing the timeouts again (as in d8f15c7361c4766667db65252b685f19d4a08b56) could work around it short-term - but if possible I'd like to spend time figuring out whether there's a deeper problem and solution, because if there is, then I think the time spent investigating should be recovered during the application's usage on aggregate.

jayaddison:
> The [test failure in `test_too_many_requests_user_timeout`](https://github.com/sphinx-doc/sphinx/actions/runs/5003267380/jobs/8964280355#step:9:1908) remains a puzzle to me; I hadn't seen that before. Perhaps it was exposed when https://github.com/sphinx-doc/sphinx/commit/0817ade877bdbc7a8bb213fa177e3b522788a03d was added? Or is it unrelated?

Facts known:

 - The error reported is timeout (ref: [build logs here](https://github.com/sphinx-doc/sphinx/actions/runs/5003267380/jobs/8964280355#step:9:1940)).
 - The linkcheck client issued a single HTTP HEAD request, and the test server responded with a single HTTP status 429 response (ref: [build logs here](https://github.com/sphinx-doc/sphinx/actions/runs/5003267380/jobs/8964280355#step:9:1949)).
 - When the linkcheck client receives an HTTP 429 response, it [re-raises the exception](https://github.com/sphinx-doc/sphinx/blob/0817ade877bdbc7a8bb213fa177e3b522788a03d/sphinx/builders/linkcheck.py#L336-L337) and [subsequently applies retry handling logic](https://github.com/sphinx-doc/sphinx/blob/0817ade877bdbc7a8bb213fa177e3b522788a03d/sphinx/builders/linkcheck.py#L347-L348) (using the error response as an argument).

Thoughts:

I've puzzled over and been suspicious about object scoping in relation to the linkchecker previously - as in #11349.

The `response` object has a kind of code smell about it in this case - it seems to exist for longer than is necessary (yes, we read some information from it later -- but those items of information could be retrieved up-front, and then the code that requires them could be refactored to accept parameters).  While it's held, I don't think that the connection resources associated with it can be released -- because the server-side of the connection is open, this is HTTP1.1.

(that shouldn't really result in a _timeout_ exception, though - we got all of the the response data we needed from the request, since it was HTTP HEAD.. am I off-base somehow with this theory?  the server does call [`self.end_headers()`](https://docs.python.org/3/library/http.server.html#http.server.BaseHTTPRequestHandler.end_headers), and that in turn calls `self.flush_headers()` - so the client should have received a complete response)

It's usually nice and possible to forget about garbage collection and the implications of it while writing Python code - in this case though, I think that it may be relevant.

This all probably results in another path that I can follow and ramble on about for a while, before maybe returning with some useful findings, or possibly without discovering anything at all and having simply wasted keyboard (and reader) time.

jayaddison:
Nope, I'm stuck re: [`test_too_many_requests_user_timeout`](https://github.com/sphinx-doc/sphinx/actions/runs/5003267380/jobs/8964280355#step:9:1908).

I'm going to remove the `daemon_threads` modification, moving this branch back to 94644fdc205b8506aae77a5c395c2952582e18c5, and will run the tests a few more times from there to see if the failure (or similar ones) continue to occur.

jayaddison:
I probably shouldn't really add emotions to descriptions about technical stuff; I tend to think it's unprofessional and not worthwhile.  But I am frustrated and a bit angry that I can't seem to track down the remaining timeout issues during these tests.

_Maybe_ the `test_auth_header_uses_first_match` test fixup (now testing in jayaddison/sphinx#3) is resolved using #11426 - I'm hopeful about that.

But another _seemingly unrelated_ [timeout failure](https://github.com/jayaddison/sphinx/actions/runs/5025371154/jobs/9012240574#step:5:1971) occurred on that branch.

Those problems aren't enough for me to throw in the towel yet, because I think this can be fixed.  But HTTP1.1 is technology from 1997 or so, and this shouldn't really be a highly complex use-case - it's a simple client and server communicating over a couple of threads.  So I feel like I should be experienced enough to figure that out.  But it feels like it's taking a while for me to figure it out, to say the least, and that's frustrating.

jayaddison:
I could probably be convinced that not using HTTP1.1 nor threading is in fact sensible after all.  I don't know, though - I do think that those are realistic for what the linkchecker will experience on the web (at least until it's possible to migrate to some other, improved and ideally widely consensus-approved HTTP version).

jayaddison:
> Taking a break from this until next week. So far from https://github.com/jayaddison/sphinx/pull/3 I'm hopeful that https://github.com/sphinx-doc/sphinx/pull/11426 resolves the test flakiness observed here, but also note that there was at least one unrelated timeout that occurred.

An update: I do think that the refactor from #11426 eliminates the `test_auth_header_uses_first_match` test flakiness.

jayaddison:
@AA-Turner @francoisfreitag I'd like to apologise for my extremely verbose communication while developing a bunch of the changes offered for Sphinx.  I think that was quite verbose and distracting.

When you have time, could you let me know whether the pull requests I currently have open are worth progressing, and whether further changes by me would be considered?  If not, that's OK - I'd prefer not to consume your and my own time if further changes from me are unlikely to be accepted.

Thanks you in advance - and I'll try to improve my communication style and approach; I'd be glad for any feedback.  You've both been supportive collaboratively and helpful technically in terms of finding problems in the pull requests so far.

francoisfreitag (approved):
HTTP/1.1 is probably used more often than 1.0, so might as well use that in tests. :+1:

jayaddison:
Thanks again @francoisfreitag :+1: 

My perspective is that HTTP 1.1 is a more widely-used protocol than HTTP 1.0 nowadays, although I fully admit that that could be biased - and I didn't find statistics comparing those two protocol versions on the [HTTP Archive website](https://httparchive.org/) when I looked a few moments ago.

In some ways HTTP 1.0 seems to me to be more robust against unusual failure modes than HTTP 1.1 -- but I think that the motivating factors for it - namely more efficient use of network resources, particlarly for websites that receive high volumes of traffic - were reasonable.

AA-Turner:
@jayaddison please could you resolve conflicts from #11426?

A

jayaddison:
> @jayaddison please could you resolve conflicts from #11426?

Done; it looks like there may be one or two more `Content-Length` response headers that I missed during the conflict resolution, though; checking..

jayaddison:
> > @jayaddison please could you resolve conflicts from #11426?
> 
> Done; it looks like there may be one or two more `Content-Length` response headers that I missed during the conflict resolution, though; checking..

Ok, the missed content-length header has also been resolved.  There was only one codesite that required updating.  The other that I had been considering was [this `send_error` call](https://github.com/sphinx-doc/sphinx/blob/835152a9db9bf9fac8d4c299eef5dad7a4174e53/tests/test_build_linkcheck.py#L203) -- it turns out that for versions of Python 3.4 onwards, [calling `send_error` will cause a content-length header to be transmitted to the client](https://docs.python.org/3.5/library/http.server.html?highlight=http%20server#http.server.BaseHTTPRequestHandler.send_error).

AA-Turner:
Thanks @jayaddison!

A

jayaddison:
Thank you!

PR #11395, opened 2023-05-04
shrink mypy whitelist for 'sphinx.application'

danieleades:


sphinx/registry.py
```diff
@@ -369,15 +372,17 @@ def add_autodoc_attrgetter(self, typ: type,
...
         logger.debug('[app] adding js_file: %r, %r', filename, attributes)
         self.js_files.append((filename, attributes))
 
     def has_latex_package(self, name: str) -> bool:
         packages = self.latex_packages + self.latex_packages_after_hyperref
         return bool([x for x in packages if x[0] == name])
 
-    def add_latex_package(self, name: str, options: str, after_hyperref: bool = False) -> None:
+    def add_latex_package(
+        self, name: str, options: str | None = None, after_hyperref: bool = False,
```
AA-Turner: Why the new default?

danieleades: i think that was a mistake!

PR #11396, opened 2023-05-04
shrink mypy whitelist for 'sphinx.builders'

danieleades:


sphinx/builders/__init__.py
```diff
@@ -296,7 +292,10 @@ def build_update(self) -> None:
                        len(to_build))
 
     def build(
-        self, docnames: Iterable[str], summary: str | None = None, method: str = 'update',
+        self,
+        docnames: Iterable[str] | None,
```
AA-Turner, line 296: We should change callers here to pass an empty iterable, rather than changing the signature (do we currently pass `None`)?

A

danieleades: I'll have to check properly but from memory I did this because we do already pass `None`. There's also a truthiness check in the method body, which would work on an empty list or tuple, but would it work on a generic iterable? I'm not so sure

AA-Turner: ``bool(iter(()))`` is the simplest example

danieleades: i would have to also modify the `build_all` method, which uses `None`. Happy with that?

```python
    def build_all(self) -> None:
        """Build all source files."""
        self.compile_all_catalogs()

        self.build(None, summary=__('all source files'), method='all')
```

would become

```python
    def build_all(self) -> None:
        """Build all source files."""
        self.compile_all_catalogs()

        self.build((), summary=__('all source files'), method='all')
```

or similar

danieleades: you'd also have to change the 'write' method

```python
def write(
        self,
        build_docnames: Iterable[str] | None,
        updated_docnames: Sequence[str],
        method: str = 'update',
    ) -> None:
        if build_docnames is None or build_docnames == ['__all__']:
            # build_all
            build_docnames = self.env.found_docs
        if method == 'update':
            # build updated ones as well
            docnames = set(build_docnames) | set(updated_docnames)
        else:
            docnames = set(build_docnames)
        logger.debug(__('docnames to write: %s'), ', '.join(sorted(docnames)))

        # add all toctree-containing files that may have changed
        for docname in list(docnames):
            for tocdocname in self.env.files_to_rebuild.get(docname, set()):
                if tocdocname in self.env.found_docs:
                    docnames.add(tocdocname)
        docnames.add(self.config.root_doc)

        with progress_message(__('preparing documents')):
            self.prepare_writing(docnames)

        with progress_message(__('copying assets')):
            self.copy_assets()

        if self.parallel_ok:
            # number of subprocesses is parallel-1 because the main process
            # is busy loading doctrees and doing write_doc_serialized()
            self._write_parallel(sorted(docnames),
                                 nproc=self.app.parallel - 1)
        else:
            self._write_serial(sorted(docnames))
```

it's a bigger change than i'd like to make in this PR. I'm really just focusing on documenting the *current* behaviour rather than making wholesale changes

PR #11400, opened 2023-05-05
shrink whitelist for 'sphinx.ext.autodoc.type_comment'

danieleades:


sphinx/ext/autodoc/type_comment.py
```diff
@@ -88,14 +88,19 @@ def get_type_comment(obj: Any, bound_method: bool = False) -> Signature | None:
...
         else:
             module = ast.parse(source, type_comments=True)
             subject = cast(ast.FunctionDef, module.body[0])
 
-        if getattr(subject, "type_comment", None):
-            function = ast.parse(subject.type_comment, mode='func_type', type_comments=True)
-            return signature_from_ast(subject, bound_method, function)  # type: ignore
+        type_comment = getattr(subject, "type_comment", None)
+        if type_comment:
+            function = ast.parse(type_comment, mode='func_type', type_comments=True)
```
AA-Turner, line 100: Why introduce an assignment?

danieleades: It allows mypy to infer that within the 'if' block, type_comment  exists and is not None.

PR #11404, opened 2023-05-06
shrink mypy whitelist for 'sphinx.ext.duration'

danieleades:


sphinx/ext/duration.py
```diff
@@ -57,7 +57,7 @@ def on_source_read(app: Sphinx, docname: str, content: list[str]) -> None:
 
 def on_doctree_read(app: Sphinx, doctree: nodes.document) -> None:
     """Record a reading duration."""
-    started_at = app.env.temp_data.get('started_at')
+    started_at = app.env.temp_data['started_at']
```
danieleades, line 60: i think this is justified since the `None` case is already not handled in the existing code

PR #11408, opened 2023-05-08
Add option to skip sphinx logger setup

flynneva:
Subject: Add the ability to skip sphinx logging formatted so third-party libs that wrap sphinx can use their logger instead

### Feature or Bugfix

- Feature

### Purpose

- Add the ability to skip sphinx logging formatted so third-party libs that wrap sphinx can use their logger instead

### Detail

- today sphinx removes any existing python loggers when `build_main` is called
- this is extremely undesirable since it limits 3rd parties from wrapping the sphinx build command with their own logger

### Relates
- <URL or Ticket>


AA-Turner:
Sphinx is intended for use as an application; please could you open an issue for this feature with more discussion of the rationale?

A

PR #11418, opened 2023-05-12
Remove setup_command from pyproject.toml

14mRh4X0r:
### Feature or Bugfix
- Bugfix

### Purpose
- pyproject.toml is still referencing `setup_command` in places, including as entry point for `distutils.commands`. This breaks any invocation of setuptools involving getting information from entry points.

### Detail
- Running e.g. `python setup.py --help-commands` will raise `ModuleNotFoundError: No module named 'sphinx.setup_command'`

### Relates
- colcon/colcon-core#556
- #11363

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -20,6 +20,9 @@ Features added
 Bugs fixed
 ----------
 
+* #11418: Remove distutils entry point and any other references to
+  ``setup_command`` from pyproject.toml
+
 Testing
 --------
 
--- a/CHANGES
+++ b/CHANGES
@@ -20,8 +20,9 @@ Features added
 Bugs fixed
 ----------
 
-* #11418: Remove distutils entry point and any other references to
-  ``setup_command`` from pyproject.toml
+* #11418: Clean up remaining references to ``sphinx.setup_command``
+  following the removal of support for setuptools.
+  Patch by Willem Mulder.
 
 Testing
 --------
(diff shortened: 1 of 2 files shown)


AA-Turner:
``setup.py`` is considered deprecated generally, does [``importlib.metadata``](https://docs.python.org/3.11/library/importlib.metadata.html#entry-points) provide the same error? E.g. ``importlib.metadata.entry_points(group='distutils.commands')``.

A

CHANGES
```diff
@@ -39,6 +39,9 @@ Features added
 Bugs fixed
 ----------
 
+* #11418: Remove distutils entry point and any other references to
+  ``setup_command`` from pyproject.toml
```
AA-Turner, line 43:
```suggestion
* #11418: Clean up remaining references to ``sphinx.setup_command``
  following the removal of support for setuptools.
```

CHANGES
```diff
@@ -39,6 +39,10 @@ Features added
 Bugs fixed
 ----------
 
+* #11418: Clean up remaining references to ``sphinx.setup_command``
+  following the removal of support for setuptools.
+  Patch by Willem.
```
AA-Turner, line 44: @14mRh4X0r please do update with your surname if you'd like, I only saw your first name on your profile.

AA-Turner:
I've released [Sphinx 7.0.1](https://pypi.org/project/Sphinx/7.0.1/) with this fix.

A

14mRh4X0r:
> `setup.py` is considered deprecated generally, does [`importlib.metadata`](https://docs.python.org/3.11/library/importlib.metadata.html#entry-points) provide the same error? E.g. `importlib.metadata.entry_points(group='distutils.commands')`.

Just for completeness' sake: it does, that's [what setuptools calls internally](https://github.com/pypa/setuptools/blob/main/setuptools/dist.py#L994).

PR #11422, opened 2023-05-14
autofix 'strict optional'

danieleades:
this is an experiment in autofixing strict optional errors using https://github.com/hauntsaninja/no_implicit_optional

gee whizz this would pair well with autoformatting wouldn't it?


danieleades:
@AA-Turner do you mind re-triggering the jobs on this PR?

PR #11426, opened 2023-05-17
tests: linkcheck: refactor to remove variable sharing between tests and test webservers

### LINKED ISSUE #11348: Flaky unit test: test_auth_header_uses_first_match

jayaddison:
### Feature or Bugfix
- Refactoring

### Purpose
- Tests for #11324 will benefit from the use of threaded webservers during unit tests.
  - The `test_auth_header_uses_first_match` test appears flaky when threaded test webservers are used (see #11348).
    - My working theory is that sharing of the `records` variable -- used to capture request header values within the server, and then assert on their values in the unit test -- across threads is somehow unsafe/unreliable (and could be the cause of the test flakiness).

### Detail
- Refactor to remove the sharing of the `records` variable between unit tests and their corresponding webserver request handler threads.
- Allow each test to provide a callable that declares the success criteria, and have the webserver thread evaluate that and communicate a resulting success/failure code using HTTP status codes.

### Relates
- May resolve #11348.
- Could be tested in combination with #11392.

### DIFF
--- a/tests/test_build_linkcheck.py
+++ b/tests/test_build_linkcheck.py
@@ -194,7 +194,7 @@ def test_raises_for_invalid_status(app):
     )
 
 
-def custom_handler(valid_credentials=None, success_criteria=lambda _: True):
+def custom_handler(valid_credentials=(), success_criteria=lambda _: True):
     """
     Returns an HTTP request handler that authenticates the client and then determines
     an appropriate HTTP response code, based on caller-provided credentials and optional
@@ -225,8 +225,10 @@ def custom_handler(valid_credentials=None, success_criteria=lambda _: True):
 
         @authenticated
         def do_GET(self):
-            response = (200, "OK") if success_criteria(self) else (400, "Bad Request")
-            self.send_response(*response)
+            if success_criteria(self):
+                self.send_response(200, "OK")
+            else:
+                self.send_response(400, "Bad Request")
             self.end_headers()
 
     return CustomHandler
--- a/tests/test_build_linkcheck.py
+++ b/tests/test_build_linkcheck.py
@@ -299,8 +299,6 @@ def test_linkcheck_request_headers_no_slash(app):
     def check_headers(self):
         if "X-Secret" in self.headers:
             return False
-        if "sesami" in self.headers.as_string():
-            return False
         if self.headers["Accept"] != "application/json":
             return False
         return True
--- a/tests/test_build_linkcheck.py
+++ b/tests/test_build_linkcheck.py
@@ -2,13 +2,13 @@
 
 from __future__ import annotations
 
-import base64
 import http.server
 import json
 import re
 import textwrap
 import time
 import wsgiref.handlers
+from base64 import b64encode
 from datetime import datetime
 from os import path
 from queue import Queue
@@ -194,16 +194,42 @@ def test_raises_for_invalid_status(app):
     )
 
 
-def capture_headers_handler(records):
-    class HeadersDumperHandler(http.server.BaseHTTPRequestHandler):
+def custom_handler(valid_credentials=None, success_criteria=lambda _: True):
+    """
+    Returns an HTTP request handler that authenticates the client and then determines
+    an appropriate HTTP response code, based on caller-provided credentials and optional
+    success criteria, respectively.
+    """
+    expected_token = None
+    if valid_credentials:
+        assert len(valid_credentials) == 2, "expected a pair of strings as credentials"
+        expected_token = b64encode(":".join(valid_credentials).encode()).decode("utf-8")
+        del valid_credentials
+
+    class CustomHandler(http.server.BaseHTTPRequestHandler):
+        def authenticated(method):
+            def method_if_authenticated(self):
+                if expected_token is None:
+                    return method(self)
+                elif self.headers["Authorization"] == f"Basic {expected_token}":
+                    return method(self)
+                else:
+                    self.send_response(403, "Forbidden")
+                    self.end_headers()
+
+            return method_if_authenticated
+
+        @authenticated
         def do_HEAD(self):
             self.do_GET()
 
+        @authenticated
         def do_GET(self):
-            self.send_response(200, "OK")
+            response = (200, "OK") if success_criteria(self) else (400, "Bad Request")
+            self.send_response(*response)
             self.end_headers()
-            records.append(self.headers.as_string())
-    return HeadersDumperHandler
+
+    return CustomHandler
 
 
 @pytest.mark.sphinx(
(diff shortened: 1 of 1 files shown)


jayaddison:
(I suppose technically #11392 should be merged first, or this should be merged into #11392 -- because threads aren't in use in the mainline codebase yet, so the pull request title is inaccurate at the moment, and would be in the commit history too)

jayaddison:
I'm sorta unsatisfied and not entirely convinced by this explanation, but it does seem that the algorithm used by `headers.as_string` method is non-trivial complexity-wise; from the [relevant documentation](https://docs.python.org/3/library/http.server.html#http.server.BaseHTTPRequestHandler.headers) it's essentially an `email.message.Message`-like object, and the latter has [had a few bugs](https://github.com/python/cpython/pulls?q=is%3Apr+email+folding+is%3Aclosed) and [has been rewritten](https://github.com/python/cpython/pull/4693) at least once.

Would that complexity be enough to put some of the unit tests near the client timeout threshold? ... I don't know, [0.05s](https://github.com/sphinx-doc/sphinx/blob/fff7d120f0e24869b7923b18c4e97b916215eed9/tests/roots/test-linkcheck-localserver/conf.py#L2) seems like it should be a lot of compute-time budget.  :shrug: 

I've removed a call to the method in 81ce0f0dcdf03788d19666e295f9ed88b94327a5 and on a sample size of one, it allowed the two instances of `test_linkcheck_request_headers` on Py3.8 (two because we run the tests against two versions of `docutils`) to pass when they both failed on a previous version of the code (although I did add a merge-latest between the two).

jayaddison:
Further suggestions and feedback welcome; otherwise I think that this should be ready for review (no further changes planned at the moment).

tests/test_build_linkcheck.py
```diff
@@ -214,25 +240,26 @@ def do_GET(self):
...
-    with http_server(capture_headers_handler(records)):
+    with http_server(custom_handler(valid_credentials=("user1", "password"))):
         app.build()
 
-    stdout = "\n".join(records)
-    assert "Authorization" not in stdout
+    with open(app.outdir / "output.json", encoding="utf-8") as fp:
+        content = json.load(fp)
+
+    assert content["status"] == "broken"
```
francoisfreitag, line 267: It would be great to verify why the link is considered broken (got a 403).

francoisfreitag (approved):
Looking good!

tests/test_build_linkcheck.py
jayaddison: It's possible that the response in this test case may later be updated to HTTP 401 (the client is not providing any authentication details, so that response would be valid, I think) - but I agree, it would be helpful to confirm that this is due to HTTP 403 currently.

jayaddison:
Thank you @francoisfreitag!

tests/test_build_linkcheck.py
```diff
@@ -194,16 +194,42 @@ def do_GET(self):
...
+
+        @authenticated
         def do_HEAD(self):
             self.do_GET()
 
+        @authenticated
         def do_GET(self):
-            self.send_response(200, "OK")
+            response = (200, "OK") if success_criteria(self) else (400, "Bad Request")
+            self.send_response(*response)
```
AA-Turner:
```suggestion
            if success_criteria(self):
                self.send_response(200, "OK")
            else:
                self.send_response(400, "Bad Request")
```

tests/test_build_linkcheck.py
```diff
@@ -194,16 +194,42 @@ def do_GET(self):
     )
 
 
-def capture_headers_handler(records):
-    class HeadersDumperHandler(http.server.BaseHTTPRequestHandler):
+def custom_handler(valid_credentials=None, success_criteria=lambda _: True):
```
AA-Turner:
```suggestion
def custom_handler(valid_credentials=(), success_criteria=lambda _: True):
```

AA-Turner:
More timeouts:

> ```
> FAILED tests/test_build_linkcheck.py::test_defaults - assert 6 == 5
>  +  where 6 = len(["links.rst:4: [broken] http://localhost:7777/#!bar: HTTPConnectionPool(host='localhost', port=7777): Read timed out. (read timeout=0.05)", "links.rst:7: [broken] http://localhost:7777#does-not-exist: Anchor 'does-not-exist' not found", 'links.rst:9: [broken] path/to/notfound: ', "links.rst:6: [broken] http://localhost:7777/#top: Anchor 'top' not found", 'links.rst:11: [broken] http://localhost:7777/image.png: 404 Client Error: Not Found for url: http://localhost:7777/image.png', 'links.rst:13: [broken] http://localhost:7777/image2.png: 404 Client Error: Not Found for url: http://localhost:7777/image2.png'])
>  +    where ["links.rst:4: [broken] http://localhost:7777/#!bar: HTTPConnectionPool(host='localhost', port=7777): Read timed out. (read timeout=0.05)", "links.rst:7: [broken] http://localhost:7777#does-not-exist: Anchor 'does-not-exist' not found", 'links.rst:9: [broken] path/to/notfound: ', "links.rst:6: [broken] http://localhost:7777/#top: Anchor 'top' not found", 'links.rst:11: [broken] http://localhost:7777/image.png: 404 Client Error: Not Found for url: http://localhost:7777/image.png', 'links.rst:13: [broken] http://localhost:7777/image2.png: 404 Client Error: Not Found for url: http://localhost:7777/image2.png'] = <built-in method splitlines of str object at 0x55aa021f40d0>()
>  +      where <built-in method splitlines of str object at 0x55aa021f40d0> = "links.rst:4: [broken] http://localhost:7777/#!bar: HTTPConnectionPool(host='localhost', port=7777): Read timed out. (read timeout=0.05)\nlinks.rst:7: [broken] http://localhost:7777#does-not-exist: Anchor 'does-not-exist' not found\nlinks.rst:9: [broken] path/to/notfound: \nlinks.rst:6: [broken] http://localhost:7777/#top: Anchor 'top' not found\nlinks.rst:11: [broken] http://localhost:7777/image.png: 404 Client Error: Not Found for url: http://localhost:7777/image.png\nlinks.rst:13: [broken] http://localhost:7777/image2.png: 404 Client Error: Not Found for url: http://localhost:7777/image2.png\n".splitlines
> ```

jayaddison:
Thanks again @AA-Turner.  Related to this and #11432: do you have any idea why it might be that the LaTeX-related continuous integration tests seem to be more susceptible to timeout failures?

PR #11431, opened 2023-05-21
linkcheck builder: don't retry with HTTP GET after SSLError on HTTP HEAD

### LINKED ISSUE #9592: Linkcheck shouldn't retry on SSL errors
### Describe the bug

When linkcheck sees `SSLError` it still retries the HTTP request [hoping for different results](https://quoteinvestigator.com/2017/03/23/same/). It shouldn't do this and it's actually a great place for optimization.

```console
(guides/supporting-windows-using-appveyor: line  152) broken    http://tox.testrun.org - HTTPSConnectionPool(host='tox.testrun.org', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError("hostname 'tox.testrun.org' doesn't match 'codespeak.net'")))
```

### How to Reproduce

```console
$ git clone https://github.com/pypa/packaging.python.org
$ cd packaging.python.org
$ pip install nox virtualenv -U
$ nox -s linkcheck
# After this, see the error cited above
```

### Expected behavior

Linkcheck should mark links as broken, the first time it sees TLS errors related to fetching the respective URLs.
Moreover, linkcheck should memorize hostnames that produced have broken TLS configuration and skip checking any further links having the same domain — they'll result in the same issue so it's reasonable to mark them all broken right away saving resources and making the check fail fast.

### Your project

https://github.com/pypa/packaging.python.org

### Screenshots

_No response_

### OS

GHA CI / Ubuntu 20.04

### Python version

CPython v3.9.6

### Sphinx version

v4.0.3

### Sphinx extensions

sphinx.ext.extlinks, sphinx.ext.intersphinx, sphinx.ext.todo, sphinx_inline_tabs, sphinx_copybutton

### Extra tools

_No response_

### Additional context

https://github.com/pypa/packaging.python.org/pull/968/checks#step:7:116

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- If an `SSLError` occurs following an HTTP HEAD operation, it seems unlikely that a subsequent HTTP GET will succeed, so don't attempt one.

### Detail
- The `SSLError` exception that can be raised by `requests` [is a subclass of `ConnectionError`](https://github.com/psf/requests/blob/2ad18e0e10e7d7ecd5384c378f25ec8821a10a29/requests/exceptions.py#L57-L58), so [our logic to handle HTTP HEAD request failures has allowed an HTTP GET to occur as a fallback](https://github.com/sphinx-doc/sphinx/blob/d3c91f951255c6729a53e38c895ddc0af036b5b9/sphinx/builders/linkcheck.py#L335-L341).
- My understanding is that [TLS operates at a protocol layer below HTTP](https://en.wikipedia.org/wiki/HTTPS#Network_layers), and so if there is a TLS-related error from a host, it seems unlikely that retrying with a different higher-layer protocol request (HTTP GET instead of HTTP HEAD) could succeed.
- We shouldn't make additional HTTP requests that we do not believe will succeed.

### Relates
- Partially addresses #9592.
- See also #9306 (could invalidate this approach?).

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -334,6 +334,7 @@ class HyperlinkAvailabilityCheckWorker(Thread):
                     break
 
                 except SSLError as err:
+                    # SSL failure; report that the link is broken.
                     return 'broken', str(err), 0
 
                 except (ConnectionError, TooManyRedirects) as err:
@@ -364,7 +365,7 @@ class HyperlinkAvailabilityCheckWorker(Thread):
                     continue
 
                 except Exception as err:
-                    # Unhandled exception (intermittent or permanent); report that the
+                    # Unhandled exception (intermittent or permanent); report that
                     # the link is broken.
                     return 'broken', str(err), 0
 
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -18,7 +18,7 @@ from urllib.parse import unquote, urlparse, urlunparse
 
 from docutils import nodes
 from requests import Response
-from requests.exceptions import ConnectionError, HTTPError, TooManyRedirects
+from requests.exceptions import ConnectionError, HTTPError, SSLError, TooManyRedirects
 
 from sphinx.application import Sphinx
 from sphinx.builders.dummy import DummyBuilder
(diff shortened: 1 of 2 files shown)


francoisfreitag (approved):
:+1:

jayaddison:
Thank you, @francoisfreitag!

jayaddison:
Thanks, @AA-Turner!

Issue #1687, opened 2015-01-19
Allow linkcheck to treat URLs that return 401 to be considered working.


tomprince:
401 is authorization required. For example,

https://console.aws.amazon.com/ec2/v2/home?region=us-west-1#LaunchInstanceWizard:ami=ami-f8f1c8bd

is a reasonable thing to link to that linkcheck doesn't like.

birkenfeld:
Agreed.

PR #11432, opened 2023-05-21
Refactor: linkcheck builder: reduce the lifetime of the 'response' variable

jayaddison:
### Feature or Bugfix
- Refactoring

### Purpose
- Reduce the runtime duration during which the `response` variable that holds the results of HTTP request results should be considered live and is non-garbage-collectable. 

### Detail
- I think that delayed resource-collection of the `response` variable may be a contributing factor to timeouts that occur during our unit tests.

### Relates
- #11299.
- https://github.com/sphinx-doc/sphinx/pull/11392#issuecomment-1554462310
  - > The `response` object has a kind of code smell about it in this case - it seems to exist for longer than is necessary (yes, we read some information from it later -- but those items of information could be retrieved up-front, and then the code that requires them could be refactored to accept parameters). While it's held, I don't think that the connection resources associated with it can be released -- because the server-side of the connection is open, this is HTTP1.1.

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -87,11 +87,10 @@ class AnchorCheckParser(HTMLParser):
                 break
 
 
-def check_anchor(response: requests.requests.Response, anchor: str) -> bool:
-    """Reads HTML data from a response object `response` searching for `anchor`.
-    Returns True if anchor was found, False otherwise.
+def contains_anchor(response: Response, anchor: str) -> bool:
+    """Determines whether an anchor is contained within an HTTP response
     """
-    parser = AnchorCheckParser(anchor)
+    parser = AnchorCheckParser(unquote(anchor))
     # Read file in chunks. If we find a matching anchor, we break
     # the loop early in hopes not to have to download the whole thing.
     for chunk in response.iter_content(chunk_size=4096, decode_unicode=True):
@@ -286,16 +285,11 @@ class HyperlinkAvailabilityCheckWorker(Thread):
             return {}
 
         def check_uri() -> tuple[str, str, int]:
-            # split off anchor
-            if '#' in uri:
-                req_url, anchor = uri.split('#', 1)
-                for rex in self.anchors_ignore:
-                    if rex.match(anchor):
-                        anchor = None
-                        break
-            else:
-                req_url = uri
-                anchor = None
+            req_url, delimiter, anchor = uri.partition('#')
+            for rex in self.anchors_ignore if delimiter and anchor else []:
+                if rex.match(anchor):
+                    anchor = None
+                    break
 
             # handle non-ASCII URIs
             try:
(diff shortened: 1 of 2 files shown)


sphinx/builders/linkcheck.py
```diff
@@ -313,71 +307,97 @@ def check_uri() -> tuple[str, str, int]:
...
-                else:
+                    error_message = str(err)
+                    continue
+
+                except HTTPError as err:
+                    error_message = str(err)
+
+                    # Unauthorized: the reference probably exists
+                    if status_code == 401:
+                        return 'working', 'unauthorized', 0
```
jayaddison: I'm not completely convinced that this logic is sound.  It was proposed in https://github.com/sphinx-doc/sphinx/issues/1687 - and for many webservers, yes perhaps HTTP 401 could indicate that a resource exists but is unavailable due to a lack of authentication - but my expectation would be that a decently large number of webservers apply that kind of response code across entire URL path ranges, in a way that does not confirm to the client whether a specific individual URL within that path range exists.

jayaddison: Filed #11433 to discuss and track this.

AA-Turner:
Thanks @jayaddison!

A

AA-Turner:
https://github.com/sphinx-doc/sphinx/actions/runs/5615478788/job/15215837543 -- LaTeX timed out

jayaddison:
> https://github.com/sphinx-doc/sphinx/actions/runs/5615478788/job/15215837543 -- LaTeX timed out

See also a similar test failure at: https://github.com/sphinx-doc/sphinx/actions/runs/5615865316/job/15217324918

sphinx/builders/linkcheck.py
```diff
@@ -313,71 +307,83 @@ def check_uri() -> tuple[str, str, int]:
...
+            status_code = -1
+            response_url = retry_after = ''
+            for retrieval_method, retrieval_kwargs in _retrieval_methods(
+                    self.config.linkcheck_anchors, anchor,
+            ):
+                try:
+                    with retrieval_method(url=req_url, auth=auth_info, config=self.config,
+                                          **retrieval_kwargs, **kwargs) as response:
+                        if response.ok and anchor and not contains_anchor(response, anchor):
+                            raise Exception(__(f'Anchor {anchor!r} not found'))
```
jayaddison, line 325: It looks like I introduced a regression with the rewrite of these lines - and the regression is reported here: https://github.com/sphinx-doc/sphinx/issues/11532#issuecomment-1658326326

Basically: when anchor checking is not enabled -- that is, `self.linkcheck_anchors` is `False` -- the first retrieval method returned is an HTTP HEAD request.  It can have `response.ok`, and the `anchor` value may be non-empty -- because we continue to extract (partition) the anchor from the `uri` regardless of whether anchor linkchecking will be performed.  The HTTP HEAD response doesn't contain a content body, so `contains_anchor` will return false, and the exception is raised.

I'll prepare a fixup/patch for this and some test coverage to cover this situation within the next few days.

cc @adamtheturtle

PR #11435, opened 2023-05-21
test_intl: resolve test flakiness by using integer nanosecond timestamps during file modification detection

### LINKED ISSUE #11232: Seemingly random CI failures of ``tests/test_intl.py::test_gettext_dont_rebuild_mo`` with Windows
### Describe the bug

I am (edit: *not*) competent to understand seemingly occasional:

```
FAILED tests/test_intl.py::test_gettext_dont_rebuild_mo - AssertionError: assert 1 == 0
 +  where 1 = <function test_gettext_dont_rebuild_mo.<locals>.get_number_of_update_targets at 0x0000022947B2A8E0>(<SphinxTestApp buildername='dummy'>)
= 1 failed, 1787 passed, 23 skipped, 5 xfailed, 28 xpassed, 6 warnings in 271.50s (0:04:31) =
Error: Process completed with exit code 1.
```

which I have observed a number of times recently either on PRs or direct pushes to master.

### How to Reproduce

See https://github.com/sphinx-doc/sphinx/actions/runs/4383723023/jobs/7674313285 or earlier failed test of #11224 prior to merge to master: https://github.com/sphinx-doc/sphinx/actions/runs/4354037120/jobs/7608843307

As this seems random, I have no recipe for reproducing.

### Environment Information

```text
not relevant
```

### Sphinx extensions

_No response_

### Additional context

_No response_

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- The intent here is to improve the reliability of the tests within the `test_intl` module, and in particular to resolve #11232.

### Detail
- The flaky test assertions relate to the detection of modified files during application builds; and it seems to be message catalog files (`.mo`) that are relevant.
- Timestamp detection had been using floating-point seconds-since-the-epoch.
- Based on a small sample size of testing from jayaddison/sphinx#3, I believe that integer-based nanoseconds-since-the-epoch may be more reliable (but I'm not 100% confident).
- At the same time, revert commit 718993afc8e0e61c689a7681aa27f0f6dfbbcb7e that I believe should be unnecessary -- and maybe-cause-related -- simplifying the code and retrieving the current-time as nanoseconds-since-epoch.

### Relates
- May resolve #11232.

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -496,7 +496,7 @@ class Builder:
             doctree = publisher.document
 
         # store time of reading, for outdated files detection
-        self.env.all_docs[docname] = time.time_ns()
+        self.env.all_docs[docname] = time.time_ns() // 1_000
 
         # cleanup
         self.env.temp_data.clear()
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -496,7 +496,7 @@ class Builder:
             doctree = publisher.document
 
         # store time of reading, for outdated files detection
-        self.env.all_docs[docname] = time.time()
+        self.env.all_docs[docname] = time.time_ns()
 
         # cleanup
         self.env.temp_data.clear()
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -496,11 +496,7 @@ class Builder:
             doctree = publisher.document
 
         # store time of reading, for outdated files detection
-        # (Some filesystems have coarse timestamp resolution;
-        # therefore time.time() can be older than filesystem's timestamp.
-        # For example, FAT32 has 2sec timestamp resolution.)
-        self.env.all_docs[docname] = max(time.time(),
-                                         path.getmtime(self.env.doc2path(docname)))
+        self.env.all_docs[docname] = time.time()
 
         # cleanup
         self.env.temp_data.clear()
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -7,7 +7,7 @@ import os
 import pickle
 from collections import defaultdict
 from copy import copy
-from datetime import datetime
+from datetime import datetime, timezone
 from os import path
 from typing import TYPE_CHECKING, Any, Callable, Generator, Iterator
 
@@ -166,9 +166,9 @@ class BuildEnvironment:
         # All "docnames" here are /-separated and relative and exclude
         # the source suffix.
 
-        # docname -> mtime at the time of reading
+        # docname -> time of reading (in integer nanoseconds)
         # contains all read docnames
-        self.all_docs: dict[str, float] = {}
+        self.all_docs: dict[str, int] = {}
         # docname -> set of dependent file
         # names, relative to documentation root
         self.dependencies: dict[str, set[str]] = defaultdict(set)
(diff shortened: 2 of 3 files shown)


picnixz:
It appears the Pycode parser fails for Python 3.12 (seems that the last "empty" line is not detected or something like that), or seems that there is some issue with the end line number of a statement. Weird that it happens with your implementation but not with other PRs (???)

jayaddison:
Thanks!  I guess an updated `python3.12-dev` version has appeared recently that somehow affects those tests.  Are the numbers that have changed line numbers or column numbers?

jayaddison:
Would you like to file a bugreport for those test failure or should I do that?

I've been taking a look at it - it does seem like the f-string changes could be relevant, although we don't use f-strings directly in the affected unit tests or parsing code (but there could be some common syntax tree / other parsing logic that has been modified and affects the test results).

jayaddison:
I don't notice any cases where more more than one empty line appears contiguously in the sample code containing `some_function`, and that makes me wonder how the resulting `end_pos` line test assertion could be off by two.

(I don't have a nightly build of Py3.12 available currently so I'm trying to reason about what the cause could be, as part of determining the cause.  I think that it may be worth reporting upstream, but would like to collect more details before doing that)

picnixz:
I think it is essentially this: 

```python
end_pos = self.current.end[0] - 1 
while emptyline_re.match(self.get_line(end_pos)): 
    end_pos -= 1 
```

If you want to investigate, you probably need to look at the differences of the `tokenize` module in Python 3.11 and 3.12.

jayaddison:
> There was a PEP-isation for f-strings which may have changed the way that the AST parser acts and deduce line numbers. I didn't bisect the culprit but it's probably the work that has been done upstream concerning f-strings and the AST parser.

Yep, this did seem to be the cause of the Python3.12-dev test failures here.  In particular, it introduced a change in last-line DEDENT token emission; as tracked (and now resolved) by https://github.com/python/cpython/issues/104976

(so I _think_ it should be possible to run the tests on this PR again and they should pass with the latest Py3.12-dev)

sphinx/environment/__init__.py
AA-Turner, line 1:
```diff
         # docname -> mtime at the time of reading
         # contains all read docnames
-        self.all_docs: dict[str, float] = {}
+        self.all_docs: dict[str, int] = {}
```

sphinx/builders/__init__.py
```diff
@@ -496,11 +496,7 @@ def read_doc(self, docname: str, *, _cache: bool = True) -> None:
             doctree = publisher.document
 
         # store time of reading, for outdated files detection
-        # (Some filesystems have coarse timestamp resolution;
-        # therefore time.time() can be older than filesystem's timestamp.
-        # For example, FAT32 has 2sec timestamp resolution.)
-        self.env.all_docs[docname] = max(time.time(),
-                                         path.getmtime(self.env.doc2path(docname)))
+        self.env.all_docs[docname] = time.time_ns()
```
AA-Turner, line 499:
```suggestion
        self.env.all_docs[docname] = time.monotonic_ns()
```

AA-Turner:
Thanks @jayaddison!

A

AA-Turner:
https://github.com/sphinx-doc/sphinx/actions/runs/5615685910/job/15216466568

A test failure on ``test_gettext_dont_rebuild_mo``...

> ```
> FAILED tests/test_intl.py::test_gettext_dont_rebuild_mo - AssertionError: assert 1 == 0
>  +  where 1 = <function test_gettext_dont_rebuild_mo.<locals>.get_number_of_update_targets at 0x0000020DF2DD83C0>(<SphinxTestApp buildername='dummy'>)
> = 1 failed, 1821 passed, 23 skipped, 5 xfailed, 28 xpassed in 212.44s (0:03:32) =
> ```

PR #11440, opened 2023-05-23
Compatibility: pycode parser: adjust dedent-token handling to maintain py3.12+ compatibility

### LINKED ISSUE #11436: Test failures: Python 3.12-dev: pycode.parser.DefinitionFinder line numbering assertion failures
### Describe the bug

With recent development builds of Python 3.12, a few parsing-related tests have begun failing.  These are built upon [Python's `tokenizer` module](https://docs.python.org/3/library/tokenize.html) and are implemented in the [`pycode.parser.DefinitionFinder` class](https://github.com/sphinx-doc/sphinx/blob/d48cc783b6e48b99cbbcf5462f44c59a4ed78e6a/sphinx/pycode/parser.py#L461-L533).

The failures appear to relate to the [logic that determines the ending line number for definitions](https://github.com/sphinx-doc/sphinx/blob/d48cc783b6e48b99cbbcf5462f44c59a4ed78e6a/sphinx/pycode/parser.py#L528-L530).

Example build logs:

- https://github.com/sphinx-doc/sphinx/actions/runs/4974642742/jobs/8901224179?pr=11423#step:9:1915
- https://github.com/sphinx-doc/sphinx/actions/runs/5044422489/jobs/9047456446#step:9:1915

cc @picnixz 

### How to Reproduce

Minimal repro steps pending; in the meantime, this appears reproducible for most pushes/pull requests that initiate GitHub Actions unit test workflows from this repository.

### Environment Information

```text
Python 3.12.0a7+

(sourced from: https://github.com/sphinx-doc/sphinx/actions/runs/4974642742/jobs/8901224179?pr=11423#step:5:10)
```

### Sphinx extensions

```python
N/A
```

### Additional context

Some discussion from today (2023-05-22) in #11435.

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Resolves test failures in Python code parsing (particularly for named definitions within indented code blocks).

### Detail
- From version 3.12.0a7 of Python, the line positions included on DEDENT tokens (encountered during Sphinx's Python code parsing logic, to determine the line-positioning of definitions) are always within the bounds of the number of lines in the source file.
- This means that we should not decrement that line position when we encounter it at the end-of-file for version 3.12+ of Python.

### Relates
- Resolves #11436.

Thank you, @mgmacias95 for guidance on the fix.

### DIFF
--- a/sphinx/pycode/parser.py
+++ b/sphinx/pycode/parser.py
@@ -528,11 +528,9 @@ class DefinitionFinder(TokenProcessor):
             end_pos, end_col = self.current.end
 
             def is_implicit_dedent():
-                # EOF
                 if end_pos == len(self.buffers) + 1 and end_col == 0:
                     return True
-                # Non-empty code line
-                if self.get_line(end_pos):
+                if end_col < len(self.get_line(end_pos)):
                     return True
                 return False
 
--- a/sphinx/pycode/parser.py
+++ b/sphinx/pycode/parser.py
@@ -532,6 +532,7 @@ class DefinitionFinder(TokenProcessor):
                     return True
                 if end_col < len(self.get_line(end_pos)):
                     return True
+                return False
 
             # Remove empty lines starting from the line _before_ implicit dedents
             # refs: https://github.com/sphinx-doc/sphinx/issues/11436
--- a/sphinx/pycode/parser.py
+++ b/sphinx/pycode/parser.py
@@ -532,7 +532,7 @@ class DefinitionFinder(TokenProcessor):
                 if end_pos == len(self.buffers) + 1 and end_col == 0:
                     return True
                 # Non-empty code line
-                if self.get_line(end_pos):
+                if not emptyline_re.match(self.get_line(end_pos)):
                     return True
                 return False
 
--- a/sphinx/pycode/parser.py
+++ b/sphinx/pycode/parser.py
@@ -528,9 +528,11 @@ class DefinitionFinder(TokenProcessor):
             end_pos, end_col = self.current.end
 
             def is_implicit_dedent():
+                # EOF
                 if end_pos == len(self.buffers) + 1 and end_col == 0:
                     return True
-                if end_col < len(self.get_line(end_pos)):
+                # Non-empty code line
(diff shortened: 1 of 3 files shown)


sphinx/pycode/parser.py
```diff
@@ -525,7 +525,21 @@ def finalize_block(self) -> None:
...
         if definition[0] != 'other':
             typ, funcname, start_pos = definition
-            end_pos = self.current.end[0] - 1
+            end_pos, end_col = self.current.end
+
+            def is_implicit_dedent():
+                if end_pos == len(self.buffers) + 1 and end_col == 0:
+                    return True
+                if end_col < len(self.get_line(end_pos)):
+                    return True
```
jayaddison: The first of these two conditional `if` blocks handles implicit dedent tokens that occur when the end-of-file is reached (because any open code blocks should be closed at that point).

I'm less confident in explaining what the second conditional is intended to handle - and it may be possible to simplify the logic used for it.

jayaddison: The second condition catches dedent tokens that are contained on the same line as some code.  For example, similar to the 'compact' test code coverage added:

```
class A:
    def b():
        pass
def c():
    pass
```

... we want the definition of `b` to omit the line that `c` is on.  It's a non-empty line, and it emits a dedent token, and it contains a name token.

picnixz:
I think the issue was solved with https://github.com/python/cpython/pull/105030. At least my latest PR does not fail on 3.12.

jayaddison:
Thanks @picnixz, glad to hear that the tests are passing again.

I had a read of that thread, was slightly puzzled because it affected a different platform to our test failures, and then through a linked `pycodestyle` discussion, landed at python/cpython#104980 - which I _think_ was likely the resolution (I could be mistaken.. but from the description and the way that relates to my understanding of the problem here, it fits).

Either way; closing this as unnecessary.

picnixz:
Do you want to leave the issue open? (maybe for confirmation?)

jayaddison:
I think we can close it, unless it begins occurring again.  Asking for confirmation seems like a good sanity-check though :+1:

PR #11494, opened 2023-07-19
Fix HTML specification mandates value for attribute defer

mvorisek:
### Purpose

fixes xmllint warning:
```
Build finished. The HTML pages are in _build/html.
./search.html:20: parser error : Specification mandates value for attribute defer
  <script src="searchindex.js" defer></script>
                                    ^
-:1: parser error : Document is empty
```

there are more `defer` attributes across the repo, but they have values, thus the fixed file/line is the only one which needs a fix

let me know if the PR should be retargetted as it is bugfix

### DIFF
--- a/sphinx/themes/basic/search.html
+++ b/sphinx/themes/basic/search.html
@@ -15,7 +15,7 @@
     <script src="{{ pathto('_static/language_data.js', 1) }}"></script>
 {%- endblock %}
 {% block extrahead %}
-  <script src="{{ pathto('searchindex.js', 1) }}" defer></script>
+  <script src="{{ pathto('searchindex.js', 1) }}" defer="defer"></script>
   {{ super() }}
 {% endblock %}
 {% block body %}


AA-Turner:
I can't find a resource stating ``defer`` must have a value. See e.g. https://html.spec.whatwg.org/#attr-script-defer -- this states:

> The ``async`` and ``defer`` attributes are [boolean attributes](https://html.spec.whatwg.org/#boolean-attribute) that indicate how the script should be evaluated

A

mvorisek:
I do not think the issue is with `defer` as such, but any html attribute without value. Boolean html values, to be true, should have the attribute name as value. When you lookup the other "defer" occurences in this repo, you will see it is defined on more places and there are in the `defer="defer"` form.

AA-Turner:
Note: https://html.spec.whatwg.org/#boolean-attribute

I see you are using ``xmllint``, which may not support HTML 5. To quote from the specification:

> The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

[W3C's validator](https://github.com/validator/validator/tree/main#usage) looks to support HTML 5, see [their website](https://validator.github.io/validator/#usage).

A

mvorisek:
@AA-Turner I agree with you, in HTML 5, a boolean attribute may omit the value. As I saw all other boolean attributes in sphinx, at least a lot of them. are defined with value, it might be good to review them. The full list of boolean html attributes is here: https://meiert.com/en/blog/boolean-attributes-of-html/

PR #11499, opened 2023-07-22
Re-structure the ``linkcheck`` builder

AA-Turner:
cc: @jayaddison

This PR re-organises and re-structures the ``linkcheck`` builder.

Of note:

- All functions defined within functions are factored out into top-level functions or class methods
- Classes and methods have been re-arranged (Builder, PostTransform, Checker, Worker)
- TLS verification on ``sphinx.util.requests`` has been changed to not pass the ``Config`` object all the way down
- The ``Hyperlink`` object now stores the document path
- ``BuildEnvironment`` and ``Config`` objects are used to extract properties and are not stored as class attributes

A

### DIFF
--- a/sphinx/builders/linkcheck.py
+++ b/sphinx/builders/linkcheck.py
@@ -12,22 +12,27 @@ from html.parser import HTMLParser
 from os import path
 from queue import PriorityQueue, Queue
 from threading import Thread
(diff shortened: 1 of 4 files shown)


jayaddison:
@AA-Turner is the removal of the `requests` version from the Sphinx user-agent string an intentional part of this changeset?

Everything else looks good to me; I have one or two minor suggestions that I'll open as a separate pull request.

Issue #11393, opened 2023-05-03
Additional hashlib usedforsecurity support


eseglem:
**Is your feature request related to a problem? Please describe.**
Due to needing to be FIPS compliant, we had been monkey patching md5 within sphinx for a while. And more recently using the older fix. Supporting it is greatly appreciated, but a recent update to it broke us. 

This commit: https://github.com/sphinx-doc/sphinx/commit/13a87d807089b3eebb8127136fac58b6cd65a68d switched from a try / except to an if statement. Unfortunately, it doesn't work when you are stuck on CentOS 7 with python3.8. And `usedforsecurity` does exist in 3.8 on CentOS / RHEL. 

While digging in to what broke I figured out an alternative way to implement it which could broaden support a little and shouldn't break anything else. 

**Describe the solution you'd like**
Use `inspect` rather than check against the python version. And potentially `functools.partial`. 

```python
import functools
import hashlib
import inspect

hashlib_kwargs = {"usedforsecurity": False} if "usedforsecurity" in inspect.getfullargspec(hashlib.md5).kwonlyargs else {}
md5 = functools.partial(hashlib.md5, **hashlib_kwargs)
sha1 = functools.partial(hashlib.sha1, **hashlib_kwargs)
```

This way it checks availability of the kwarg once and uses it from then on. 

**Describe alternatives you've considered**
- Pin to earlier version until we can upgrade OS / Python.
- Write a new monkey patch for it.

I know its a niche use case and wouldn't fault anyone for not wanting to switch things up more. But I figured I would pass on the idea if nothing else. 

Edit: I initially thought this was more of a feature request than anything, but upon further digging and finding the above commit, I think it is a bit of a bug / regression.

AA-Turner:
`usedforsecurity=False` will be set by default in Sphinx 7.2, which drops support for Python 3.8.

A

PR #11512, opened 2023-07-25
Deprecate ``md5`` and ``sha1`` wrappers in ``sphinx.util``

AA-Turner:


picnixz:
Does this resolve https://github.com/sphinx-doc/sphinx/issues/11393 ? if so, we could perhaps close this issue ?

PR #11517, opened 2023-07-25
Fix rendering of type annotations with ``Literal`` enumeration values

### LINKED ISSUE #11473: Enum values inside `typing.Literal` render weirdly
### Describe the bug

While `typing.Literal` is picked up properly by autodoc when it just contains a string, when the contained value is an enum variant, the output is very messy.

For some reason, it shows `~typing.Literal`, rather than just `Literal`, and the enum variant seems to just be the output of  `repr(SomeEnum.X)`, which includes the value of that enum, which isn't something we'd generally need to see. It also doesn't produce a clickable reference to the specific variant (or at least to the enum class).

While using `repr` works well when the Literal just contains strings, it looks very weird for enum variants, and clearly seems to be a bug.

Here's how the documentation would look from the minimal reproducable example below:

image

<details>
<summary>
And here's another screenshot of how this might look from an actual project, to demonstrate how "cluttery" this can get:
</summary>

image
</details>

### How to Reproduce

```bash
python -m venv .venv
emulate bash -c '. .venv/bin/activate'
pip install sphinx
mkdir src
cat > src/example.py <<EOF
from typing import Literal
from enum import Enum

def works_great(x: Literal["foo"]) -> None:
    """Does stuff"""

class SomeEnum(str, Enum):
    X = "x"
    Y = "y"
    Z = "z"

def problematic_func(x: Literal[SomeEnum.X]) -> None:
    """Also does stuff"""
EOF
cat > src/conf.py <<EOF
import os, sys

sys.path.insert(0, os.path.abspath("."))

extensions = ["sphinx.ext.autodoc"]
EOF
cat > src/index.rst <<EOF
API reference
=============

.. automodule:: example
    :members:
    :undoc-members:
EOF
sphinx-build -b html src _build
xdg-open _build/index.html
```

### Environment Information

```text
Platform:              linux; (Linux-6.3.7-arch1-1-x86_64-with-glibc2.37)
Python version:        3.11.3 (main, Jun  5 2023, 09:32:32) [GCC 13.1.1 20230429])
Sphinx version:        7.0.1
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.15.1
```

### Sphinx extensions

```python
["sphinx.ext.autodoc"]
```

### Additional context

_No response_

picnixz:
Fix #11473.

I don't know whether there are more places to be updated so feel free to comment.


picnixz:
I'm converting into a draft because I don't know which whould be the best course of action. 

I'll present the different possibilities on the issue (let's discuss there).

AA-Turner:
Thanks!

PR #11536, opened 2023-07-29
Do not hardcode expected JS/CSS checksum in tests

mitya57:
Subject: Do not hardcode expected JS/CSS checksum in tests

### Feature or Bugfix
- Bugfix

### Purpose
This way the tests won't break with a different alabaster version. Also they won't need updating every time we change something in our JS/CSS files.

### DIFF
--- a/tests/test_build_html.py
+++ b/tests/test_build_html.py
@@ -1188,29 +1188,29 @@ def test_assets_order(app):
     # css_files
     expected = [
         '_static/early.css',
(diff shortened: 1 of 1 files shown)


AA-Turner:
Thanks for the PR, but I do want to expect the hard-coded values -- whilst a minor hassle to update if we change the file contents, it validates that the hashes work for what we expect them to.

A

PR #11558, opened 2023-08-07
Update ``pygments`` CSS files checksums in tests.

picnixz:
Make the tests pass again.

@AA-Turner I know that you want them to be hardcoded, but should we actually compute them dynamically using our custom checksum to handle a (user) local installed version of `pygments` that may lead to different checksums ? Or should we freeze the `pygments` version? (currently it's only lowerbounded by `2.13`).

EDIT: Here it's for Pygments 2.16.1 that was released yesterday.

### DIFF
--- a/tests/test_build_html.py
+++ b/tests/test_build_html.py
@@ -1188,7 +1188,7 @@ def test_assets_order(app):
     # css_files
     expected = [
         '_static/early.css',
-        '_static/pygments.css?v=b3523f8e',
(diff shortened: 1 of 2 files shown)


AA-Turner:
On reflection, I've updated the tests to ignore the hashes and added a new test for ``_file_checksum`` specifically -- I think this is better than my previous position of testing indirectly via ``test_assets_order`` and ``test_dark_style``.

A

PR #11565, opened 2023-08-08
Refactor ``sphinx.environment.adapters.TocTree``

AA-Turner:
This speeds up TocTree resolution by 3-6x, in local benchmarks.

A

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -969,7 +969,7 @@ class StandaloneHTMLBuilder(Builder):
             kwargs['includehidden'] = False
         if kwargs.get('maxdepth') == '':
             kwargs.pop('maxdepth')
-        toctree = global_toctree_for_doc(self.env, docname, self, collapse, **kwargs)
+        toctree = global_toctree_for_doc(self.env, docname, self, collapse=collapse, **kwargs)
         return self.render_partial(toctree)['fragment']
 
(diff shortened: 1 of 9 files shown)


mgeier:
This change broke the generation of galleries in `nbsphinx` (which are based on `toctree`): https://github.com/spatialaudio/nbsphinx/pull/755

I know that I'm doing some dubious monkey patching (https://github.com/spatialaudio/nbsphinx/blob/master/src/nbsphinx/__init__.py#L1485-L1506), but I'm not sure if the document representation is supposed to have changed in this PR?

The API change didn't make any problems, but the document structure seems to have changed. I still have to look into the details of what has changed here, but if it is not too much of a problem to restore the previous stucture, this would be great!

AA-Turner:
The actual resolution algorithm shouldn't've changed. Please could you open a new issue with how the new API (function based) differs from the old (class based)?

A

mgeier:
OK, I looked into it and it turns out that I was wrong. The document structure is not different, but my monkey-patching is broken.

In https://github.com/spatialaudio/nbsphinx/pull/756 I'm trying to monkey-patch `_resolve_toctree` instead of `TocTree.resolve` (depending on Sphinx version), but it still doesn't work.

It seems like setting the property `toctree['nbsphinx_gallery']` doesn't work anymore, do you have an idea why this might be the case?

PR #11500, opened 2023-07-22
Add more information for the ``test_gettext_dont_rebuild_mo`` test

AA-Turner:

PR #11572, opened 2023-08-09
Always say why something is out of date

larsoner:
Adds logging in two branches that was missing from the `logger.debug`. Gives output like:
```
[build target] outdated 'auto_examples/future/plot_future_imports' from dependency '/home/larsoner/python/sphinx-gallery/sphinx_gallery/tests/tinybuild/doc/auto_examples/future/images/thumb/sphx_glr_plot_future_imports_broken_thumb.png': 2023-08-09 18:36:25.911416 -> 2023-08-09 18:36:34.238679
```
Perhaps it's a bit verbose but it at least indicates that this file is marked outdated because a dependency of it was created after the file that uses it.

*BUT* this brings up an issue -- the dependency itself was *not* changed the two calls to `make html`. The conditional `if depmtime > mtime:` around line 501 checks if the `mtime` of the *file during previous build* is older than that of the *dependency in this build*, when it seems like it should check if the `mtime` of the *dependency during previous build* is older than that of the *dependency in this build*. In other words, if I create `index.rst`, then create `whatever.inc`, and `.. include: whatever.inc`, *`index.rst` will always show up as out of date* just because of the order in which I created `index` and `whatever.inc`.

Should I try to fix that here as well? It seems like a potentially tough thing to fix because it would mean adding all dependency files to `self.all_docs` (where `self` is the `BuildEnvironment`) -- or maybe better adding a `self.all_deps` that is a str->float mtime mapping for all dependencies.


larsoner:
> BUT this brings up an issue -- the dependency itself was not changed the two calls to make html

Looks like this has actually been fixed on `master` sometime so updating fixed this part for me!

sphinx/environment/__init__.py
```diff
@@ -502,10 +502,35 @@ def get_outdated_files(self, config_changed: bool) -> tuple[set[str], set[str],
...
+                                'missing dependency %r',
+                                docname,
+                                deppath)
                             changed.add(docname)
                             break
                         depmtime = _last_modified_time(deppath)
                         if depmtime > mtime:
+                            try:
+                                depmtime_dt = datetime.utcfromtimestamp(depmtime)
+                            except ValueError:  # e.g., year 53606865 is out of range
```
picnixz: Honestly, I think this is unnecessary because it's likely that when we will hit this year, `datetime.utcfromtimestamp` would handle it correctly. Also, it complicates the flow and ~~unnecessarily introduces a `try-and-catch` block which is costly in general.~~

picnixz: In addition, I think that we report the time in UTC (somewhere else in the file).

larsoner: I only added the try/except because CIs complained. I'll revert it so we can see the failure again (I rebased so can't find it easily anymore)

sphinx/environment/__init__.py
```diff
@@ -502,10 +502,27 @@ def get_outdated_files(self, config_changed: bool) -> tuple[set[str], set[str],
...
+                                docname,
+                                deppath)
                             changed.add(docname)
                             break
                         depmtime = _last_modified_time(deppath)
                         if depmtime > mtime:
+                            depmtime_dt = datetime.utcfromtimestamp(depmtime)
+                            depmtime_str = str(depmtime_dt)
+                            mtime_dt = datetime.utcfromtimestamp(mtime)
+                            mtime_str = str(mtime_dt)
```
picnixz: The conversion to `str` is unneeded since you are afterwards using `%s`.

sphinx/environment/__init__.py
```diff
@@ -502,10 +502,27 @@ def get_outdated_files(self, config_changed: bool) -> tuple[set[str], set[str],
...
+                                '[build target] changed %r '
+                                'missing dependency %r',
+                                docname,
+                                deppath)
                             changed.add(docname)
                             break
                         depmtime = _last_modified_time(deppath)
                         if depmtime > mtime:
+                            depmtime_dt = datetime.utcfromtimestamp(depmtime)
+                            depmtime_str = str(depmtime_dt)
```
picnixz: Same here.

sphinx/environment/__init__.py
picnixz: Wait, what do you mean by CIs complained? did it complained because of `utcfromtimestamp`?

larsoner:
https://github.com/sphinx-doc/sphinx/actions/runs/5821558864/job/15784236771?pr=11572#step:9:2434
```
                        if depmtime > mtime:
>                           depmtime_dt = datetime.utcfromtimestamp(depmtime)
E                           ValueError: year 53608964 is out of range
...
=========================== short test summary info ============================
FAILED tests/test_intl.py::test_gettext_dont_rebuild_mo - ValueError: year 53608965 is out of range
FAILED tests/test_intl.py::test_html_rebuild_mo - ValueError: year 53608964 is out of range
FAILED tests/test_intl.py::test_xml_footnotes - ValueError: year 53608964 is out of range
====== 3 failed, 1957 passed, 22 skipped, 2 warnings in 192.82s (0:03:12) ======
```

sphinx/environment/__init__.py
larsoner: > Wait, what do you mean by CIs complained? 

See the main comment thread, I accidentally replied there to this bit.

> try-and-catch block which is costly in general.

Can you explain whaht you mean? I haven't experienced slowdowns at least. `timeit` seems to think it's a negligible cost:
```
$ python -m timeit -s "import datetime" "datetime.datetime.fromtimestamp(1000.)"
500000 loops, best of 5: 414 nsec per loop
$ python -m timeit -s "import datetime" "try:" "    datetime.datetime.fromtimestamp(1000.)" "except:" "    pass"
500000 loops, best of 5: 417 nsec per loop
```

larsoner:
> we divide the timestamp by 1_000_000 before

Ahhh okay this is indeed probably the problem -- I'll simplify the code and push soon!

sphinx/environment/__init__.py
picnixz: Ok, maybe my knowledge was a bit too old. Since Python 3.11, there are 0-cost exceptions so it should be fine. It's just that prior to 3.11, a try-catch would make the frame larger. As such, I'm not sure how important is the effect when Sphinx is used for *very* large projects (when there are possibly million of files to be processed). 

Also, prior to 3.11, I think that Python may have already have a somewhat effective try-catch in case there is no exception being thrown. 

TL;DR: I was wrong!

larsoner: > Also, prior to 3.11, I think that Python may have already have a somewhat effective try-catch in case there is no exception being thrown.

At least 3.10 seems effective because that's what I was profiling on above

larsoner:
Okay finally green!

AA-Turner:
Please may you add an entry to CHANGES?

A

larsoner:
Done @AA-Turner !

larsoner:
Windows failure looks probably unrelated... is it a flakey test?

AA-Turner:
> is it a flakey test?

Yes, see #11232; #11500.

A

AA-Turner (approved):

AA-Turner:
Thank you @larsoner!

A

Issue #11584, opened 2023-08-12
code-block/sourcecode :class: option documentation doesn't make sense


smheidrich:
### Describe the bug

The [documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-code-block-class) for the `:class:` option of the `code-block`/`sourcecode` directive currently reads

>The class name of the graph.

Maybe it's just me but this doesn't tell me anything. What graph? What effects does setting a class name have?

Through guessing and experimentation I figured out that the classes set via this option are added to the code block's HTML `class` attribute, but it's still unclear what (if any) effect this has for other build targets, for example.

### How to Reproduce

Look at the [`code-block`/`sourcecode` `:class:` option documentation](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-code-block-class) and fail to understand it.

### Environment Information

```text
Current Sphinx master docs
```

### Sphinx extensions

```python
n/a
```

### Additional context

- The `:class:` option was first documented in https://github.com/sphinx-doc/sphinx/pull/10464 after it was pointed out that it was undocumented in https://github.com/sphinx-doc/sphinx/issues/10461.
- The current description saying "The class name of the graph" is also present in the [Graphviz extension docs](https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html). There it makes perfect sense, because [Graphviz graphs do have a well-defined `class` attribute](https://graphviz.org/docs/attrs/class/) and anyone who knows Graphviz will understand. So the explanation was most likely copied from there when a completely new one was warranted.

PR #11585, opened 2023-08-12
Improve code-block/sourcecode class option docs

smheidrich:
Subject: Improve `code-block`/`sourcecode` class option docs.

### Feature or Bugfix
- Bugfix

### Purpose
Makes clear what the `:class:` option of `code-block`/`sourcecode` actually does.

### Detail
n/a

### Relates
- #11584


doc/usage/restructuredtext/directives.rst
```diff
@@ -597,7 +597,8 @@ __ https://pygments.org/docs/lexers
    .. rst:directive:option:: class: class names
       :type: a list of class names separated by spaces
 
-      The class name of the graph.
+      Class names to add to the code block's HTML ``class`` attribute when
+      building as HTML.
```
smheidrich: As I mentioned in #11584, it's still not clear to me what this does for other build targets like LaTeX... From experimentation it seems like "nothing" is the answer, but is that true for all of them?

AA-Turner: cc: @jfbu re LaTeX.

Issue #11590, opened 2023-08-15
C coverage support in `sphinx.ext.coverage` is non-functional


stephenfin:
### Describe the bug

The [`sphinx.ext.coverage`](https://www.sphinx-doc.org/en/master/usage/extensions/coverage.html) extension supports checking coverage for both the Python and C domain. However, the latter does not appear to work. The extension checks the [`self.env.domaindata['c']['objects']`](https://github.com/sphinx-doc/sphinx/blob/6cb783c0024a873722952a67ebb9f41771c8eb6d/sphinx/ext/coverage.py#L118) but unlike in the Python domain, this attribute is never set. As a result, every API will show as undocumented. My git splunking suggests this has been the case for many years and there's a chance this has never worked, which probably explains why the only usage of the various C-specific configuration options I can find on GitHub are in CPython (an attempted user), Sphinx (the definition) or a fork of one of these two projects.

### How to Reproduce

To reproduce this you need a C application. I have packaged a minimal example, which you can find here:

https://github.com/stephenfin/sphinx_ext_coverage_test

Instructions are provided there on how to use this repo to reproduce.

### Environment Information

```text
Platform:              linux; (Linux-6.4.8-200.fc38.x86_64-x86_64-with-glibc2.37)
Python version:        3.11.4 (main, Jun  7 2023, 00:00:00) [GCC 13.1.1 20230511 (Red Hat 13.1.1-2)])
Python implementation: CPython
Sphinx version:        7.2.0+/2f9d39ce8
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Sphinx extensions

```python
extensions = [
    'sphinx.ext.coverage',
]
```

stephenfin:
[...]

PR #11591, opened 2023-08-15
Fix C coverage support for `sphinx.ext.coverage`

stephenfin:
### Feature or Bugfix

- Bugfix

### Purpose

We were using the `objects` attribute of the C domain to pull all C objects that we were aware of. However, this is never set to any value nor does it ever appear to have been. Instead, we should have been using and parsing the return value from `get_objects`.

As an aside, this would suggest code coverage for C APIs has never worked or at least has not worked for a *very* long time. This explains why every GitHub hit for the various configuration options comes from Sphinx (for the implementation), Python (for the sole user), or a fork of these. This might be a good time to deprecate and remove this aspect of the coverage extension. We can discuss that elsewhere.

### Detail

- #11590

### Relates

(none)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -80,6 +80,8 @@ Bugs fixed
 * #11473: Type annotations containing :py:data:`~typing.Literal` enumeration
   values now render correctly.
   Patch by Bénédikt Tran.
+* #11591: Fix support for C coverage in ``sphinx.ext.coverage`` extension.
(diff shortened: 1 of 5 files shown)


stephenfin:
> @stephenfin please could you add a note to CHANGES? I'd like to include this in 7.2.

Done :+1:

AA-Turner:
Thanks!

Issue #1288, opened 2015-01-03
CoverageBuilder: improve docs


shimizukawa:
http://sphinx-doc.org/ext/coverage.html

Please improve the documentation of the above page.

What kind of coverage is used here? The coverage of unittests, or coverage of the docs (how many methods have docstrings ...).

Please put on the "I am a newbee" glasses before writing the docs. Thank you.

---
- Bitbucket: https://bitbucket.org/birkenfeld/sphinx/issue/1288
- Originally reported by: Thomas Güttler
- Originally created at: 2013-10-07T16:00:18.114

techtonik:
The link is gone.

techtonik:
New link https://www.sphinx-doc.org/en/master/usage/extensions/coverage.html

Still unclear what exactly this extension does. Would be nice to see the output for the Sphinx itself.

cdruck:
I concur ... documentation is lacking, don't know how to use, don't know what to expect

PR #11592, opened 2023-08-15
coverage: Specify modules to run coverage for

stephenfin:
### Feature or Bugfix

Feature

### Purpose

Currently, the coverage builder lets you check for partially documented modules, but there is no mechanism to identify totally undocumented modules. Resolve this by introducing a new `coverage_modules` config option. This is a list of modules that should be documented somewhere within the documentation tree. Any modules that are specified in the configuration value but are not documented anywhere will result in a warning. Likewise, any modules that are not in the config option but are documented somewhere will result in a warning.

> **Note**
> I have a WIP series locally that will complete the documentation for this module and close out #1288, but I'm keeping that out of this to avoid making it any more complicated than necessary.

### Detail

(none)

### Relates

(none)


stephenfin:
I'll wait for the test to re-run, but it looks like the CI failures are flakes and this is otherwise passing as expected.

@AA-Turner Who's the best person other than me to review this and try get it into 7.3/8.0?

stephenfin:
fwiw, I came up with this after spotting the gap while trying to use coverage for a project I contribute to. [After testing](https://review.opendev.org/c/openstack/openstacksdk/+/891690), it seemed to do what I expected and highlighted a whole load of missing documentation.

doc/usage/extensions/coverage.rst
```diff
@@ -6,15 +6,58 @@
 
 This extension features one additional builder, the :class:`CoverageBuilder`.
 
+.. todo:: Write this section.
```
picnixz, line 9: Didn't you just write it :) ?

sphinx/ext/coverage.py
```diff
@@ -65,6 +66,45 @@ def _add_row(col_widths: list[int], columns: list[str], separator: str) -> Itera
...
+    # handle this
+    mod = import_module(mod_name)
+
+    modules.add(mod_name)
+
+    for sub_mod_info in pkgutil.iter_modules(mod.__path__):
+        if sub_mod_info.name == '__main__':
+            continue
+
+        # TODO: Do we want to skip private modules (note: different from private
```
picnixz: I think we can skip them by using a regex instead so let's pick them by default.

sphinx/ext/coverage.py
```diff
@@ -65,6 +66,45 @@ def _add_row(col_widths: list[int], columns: list[str], separator: str) -> Itera
...
+        # objects)?
+        # if sub_mod_info.name.startswith('_'):
+        #     continue
+
+        if sub_mod_info.ispkg:
+            modules |= load_modules(f'{mod_name}.{sub_mod_info.name}', ignored_module_exps)
+        else:
+            modules.add(f'{mod_name}.{sub_mod_info.name}')
```
picnixz: Don't you want to exclude a submodule if the latter matches a regexpr?

sphinx/ext/coverage.py
```diff
@@ -90,6 +130,7 @@ def init(self) -> None:
         for (name, exps) in self.config.coverage_ignore_c_items.items():
             self.c_ignorexps[name] = compile_regex_list('coverage_ignore_c_items',
                                                         exps)
+        self.modules = self.config.coverage_modules
```
picnixz: I think module_names would be better.

sphinx/ext/coverage.py
```diff
@@ -167,11 +208,43 @@ def ignore_pyobj(self, full_name: str) -> bool:
...
+        #   missing modules and missing objects but it requires manual configuration.
+        if not self.modules:
+            modules = set(seen_modules)
+        else:
+            modules = set()
+            for mod_name in self.modules:
+                try:
+                    modules |= load_modules(mod_name, self.mod_ignorexps)
+                except ImportError as err:
+                    logger.warning(__('module %s could not be imported: %s'), mod_name, err)
```
picnixz: Since people like to suppress warnings, could you also decide for a type/subtype for the log record and add it to the existing ones.

picnixz (commented):
Do we need some extra tests?

sphinx/ext/coverage.py
stephenfin: wfm

sphinx/ext/coverage.py
stephenfin: Good point. Done.

sphinx/ext/coverage.py
stephenfin: Agreed. Done.

sphinx/ext/coverage.py
stephenfin: My one disagreement: none of the other `logger.warning` calls in here do this, so I think we want to do it for all or none. Can I do this as a follow-up? I've added a TODO.

picnixz: Sure! Overall, we need to add those type/subtype more or less everywhere where warnings could be annoying because people request the ability to suppress them (and to turn them as non-errors), so you could have a separate issue for warnings regarding coverage-related things.

stephenfin:
> Do we need some extra tests?

Yes, we do. Added now

sphinx/ext/coverage.py
```diff
@@ -65,6 +66,44 @@ def _add_row(col_widths: list[int], columns: list[str], separator: str) -> Itera
     yield _add_line(col_widths, separator)
 
 
+def load_modules(mod_name: str, ignored_module_exps: list[re.Pattern[str]]) -> set[str]:
```
picnixz: Actually, do you want the function to be part of the public API? or could we rename it `_load_modules`?

stephenfin: Fair point. Done.

sphinx/ext/coverage.py
```diff
@@ -282,19 +354,23 @@ def _write_py_statistics(self, op: TextIO) -> None:
...
                 value = 100.0
 
             table.append([module, '%.2f%%' % value, '%d' % len(self.py_undocumented[module])])
-        table.append([
-            'TOTAL',
-            f'{100 * len(all_documented_objects) / len(all_objects):.2f}%',
-            f'{len(all_objects) - len(all_documented_objects)}',
-        ])
+
+        if len(all_objects):
```
picnixz:
```suggestion
        if all_objects:
```

sphinx/ext/coverage.py
```diff
@@ -271,8 +344,7 @@ def build_py_coverage(self) -> None:
 
     def _write_py_statistics(self, op: TextIO) -> None:
         """ Outputs the table of ``op``."""
-        all_modules = set(self.py_documented.keys()).union(
-            set(self.py_undocumented.keys()))
+        all_modules = set(self.py_documented.keys()).union(set(self.py_undocumented.keys()))
```
picnixz:
```suggestion
        all_modules = frozenset(self.py_documented.keys() | self.py_undocumented.keys())
```

tests/roots/test-ext-coverage/grog/coverage_missing.py
```diff
@@ -0,0 +1,8 @@
+"""This module is intentionally not documented."""
+
+class Missing:
+    """An undocumented class."""
+
+    def missing_a(self):
+        """An undocumented method."""
+        pass
```
picnixz:
```suggestion
```

tests/test_ext_coverage.py
```diff
@@ -10,8 +10,10 @@ def test_build(app, status, warning):
     app.builder.build_all()
 
     py_undoc = (app.outdir / 'python.txt').read_text(encoding='utf8')
-    assert py_undoc.startswith('Undocumented Python objects\n'
```
picnixz: Any reason for this change?

picnixz (commented):
Just so that you know, I didn't review the doc content.

sphinx/ext/coverage.py
stephenfin: I was pretty sure this was done for a reason. Something like this is what had been done elsewhere. I can't remember now though so...

tests/test_ext_coverage.py
stephenfin: I think this is an artefact from an earlier revision of the patch. It's unnecessary but also harmless. I'll take it out if I need to respin.

stephenfin:
> Just so that you know, I didn't review the doc content.

Thanks. I think I have addressed everything now?

picnixz:
I think so. At least I don't have anything more in mind, except the todo for the doc but as you mentioned, it'll be another PR

sphinx/ext/coverage.py
```diff
@@ -65,6 +66,86 @@ def _add_row(col_widths: list[int], columns: list[str], separator: str) -> Itera
...
+            modules |= _load_modules(mod_name, ignored_module_exps)
+        except ImportError as err:
+            # TODO(stephenfin): Define a subtype for all logs in this module
+            logger.warning(__('module %s could not be imported: %s'), mod_name, err)
+            py_undoc[mod_name] = {'error': err}
+            continue
+
+    # if there are additional modules then we warn (but still scan)
+    additional_modules = set(seen_modules) - modules
+    if additional_modules:
```
AA-Turner: I think this warning only captures half of what you noted -- modules that are in ``coverage_modules`` but not in ``seen_modules`` don't seem to get a warning.

```suggestion
    # if there are additional modules then we warn (but still scan)
    if additional_modules := set(seen_modules) - modules:
```

sphinx/ext/coverage.py
```diff
@@ -65,6 +66,86 @@ def _add_row(col_widths: list[int], columns: list[str], separator: str) -> Itera
     yield _add_line(col_widths, separator)
 
 
+def _load_modules(mod_name: str, ignored_module_exps: list[re.Pattern[str]]) -> set[str]:
```
AA-Turner: I wonder if we'd be able to use ``importlib`` (e.g. [``importlib.util.find_spec``](https://docs.python.org/dev/library/importlib.html#importlib.util.find_spec); [``importlib.util.spec_from_file_location``](https://docs.python.org/dev/library/importlib.html#importlib.util.spec_from_file_location)) to avoid importing the modules, given we only want to find their names and subpackages.

AA-Turner (approved):
Looks good, two notes

A

sphinx/ext/coverage.py
stephenfin: Does it matter? We won't import modules that we don't want to document as we check the module name first, and we import all the other modules anyway via later calls to `import_module`.

sphinx/ext/coverage.py
stephenfin: They weren't getting a warning but they were getting captured in the report. I've added the warning nonetheless.

stephenfin:
@AA-Turner this is done now

doc/usage/extensions/coverage.rst
AA-Turner, line 9: Will this TODO remain?

doc/usage/extensions/coverage.rst
```diff
@@ -6,15 +6,58 @@
...
+   list as well as all sub-packages and sub-modules found in each. When this is
+   not provided, Sphinx will only provide coverage for Python packages and
+   modules that it is aware of: that is, any modules documented using the
+   :rst:dir:`py:module` directive provided in the :doc:`Python domain
+   </usage/domains/python>` or the :rst:dir:`automodule` directive provided by
+   the :mod:`~sphinx.ext.autodoc` extension.
 
-Several configuration values can be used to specify what the builder
-should check:
+   .. versionadded: 7.2
```
AA-Turner:
```suggestion
   .. versionadded:: 7.2
```

doc/usage/extensions/coverage.rst
stephenfin: Not long-term. I have a separate series but GitHub doesn't let me stack PRs. I'll push once this is merged.

AA-Turner:
Thanks!

A

PR #11593, opened 2023-08-15
sphinx.ext.coverage: Remove errant print

stephenfin:
### Feature or Bugfix

Bugfix

### Purpose

Remove an errant print. I suspect this was left over when the coverage report was introduced recently. This is noisy and we already have the capability to log missing modules to stdout as-is.

### Detail

(none)

### Relates

(none)


stephenfin:
This is also done in #11592 but I'm aware that might take longer to review/discuss so I'd like to get this in before we release the coverage report feature.

picnixz (approved):
Good catch! Are there other errant prints in the project? (perhaps in tests or something like that). If so, we could remove the as well.

PR #11594, opened 2023-08-15
Fix layout issues in agogo theme on smaller devices

lukengda:
Subject: Fix layout issues in agogo theme on smaller devices

### Feature or Bugfix
- Feature

### Purpose
- This pull request aims to improve some layout issues in the agogo theme for devices with less screen space.
- I provided before and after screenshots to show the effects
- The changes can be shown on the agogo theme demo <https://sphinx-themes.org/sample-sites/default-agogo/> by generating the demos or by patching the browser css.

### Detail
- fix: background colors breaking at horizontal scrolling on smaller devices
  The body now grows to fit all the content, which makes the background colors of the theme no longer break as seen below:
  sphinx-agogo-scrolling-fix-current sphinx-agogo-scrolling-fix-current
- feature: add whitespace around the web page frame to ease the eyes on smaller screens
  The content text is very hard to read without any margin to the browser window. Modern browsers eliminated all margins from the browser side, which results in a painful reading experience for some users. The margin to the side does not affect the presentation on bigger screens where the content has auto-growing margins left and right. To make the changes appealing, the header was grown equal parts:
  sphinx-agogo-scrolling-fix-current sphinx-agogo-scrolling-fix-current

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -86,6 +86,9 @@ Bugs fixed
   Patch by Bénédikt Tran.
 * #11591: Fix support for C coverage in ``sphinx.ext.coverage`` extension.
   Patch by Stephen Finucane.
+* #11594: HTML Theme: Enhancements to horizontal scrolling on smaller
+  devices in the ``agogo`` theme.
+  Patch by Lukas Engelter.
 
 Testing
 -------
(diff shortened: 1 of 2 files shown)


AA-Turner:
Please add a note to CHANGES

A

lukengda:
@AA-Turner, hope the changelog is okay like this. Thanks for taking care so quickly!

PR #11596, opened 2023-08-15
Fix rendering of enumerations inheriting from mixin and data types.

### LINKED ISSUE #11353: Bad computation of enum superclass in `sphinx.ext.autodoc.importer.get_class_members()`
### Describe the bug

The computation for the super-class in [`sphinx.ext.autodoc.importer.get_class_members()`](https://github.com/sphinx-doc/sphinx/blob/v6.1.3/sphinx/ext/autodoc/importer.py#L227) is:
```python
superclass = subject.__mro__[1]
```

The version above does not check the super-class taken into account is actually the expected base enum type.

As a consequence, when the super-class is wrong:
- the method does not filter out a couple of members it's used to filter-out for enums: `_generate_next_value_`, `__weakref__`, `_member_names_`, `_member_map_`, `_member_type_`, `_value2member_map_`
- the method filters out other members it should not: `__str__()` for instance.

The computation below fixes the thing:
```python
superclass = list(filter(sphinx.util.inspect.isenumclass, subject.__mro__))[1]
```

### How to Reproduce

```bash
git clone https://github.com/alxroyer/scenario/
cd scenario/
python -m sphinx -T -E -b html -D language=en ./tools/conf/sphinx/ ./doc/html/
```

The problem occurs for the [`StrEnum`](https://github.com/alxroyer/scenario/blob/v0.2.2/src/scenario/enumutils.py#L25) class.

### Environment Information

```text
Sphinx version in use: 5.3.0, but the wrong line is still the same with 6.1.3.
```

### Sphinx extensions

```python
Base `sphinx.ext.autodoc` should be enough.
```

### Additional context

_No response_

picnixz:
Fix #11353.

I'm not sure that everything is covered so feel free to argue.


sphinx/ext/autodoc/importer.py
```diff
@@ -198,7 +198,8 @@ def get_object_members(
             if name not in members:
                 members[name] = Attribute(name, True, value)
 
-        superclass = subject.__mro__[1]
+        # enumerations are created as `EnumName([mixin_type, ...] [data_type,] enum_type)`
+        superclass = subject.__mro__[-2]
```
AA-Turner: Perhaps also note that the superclass that we want is the data_type class. Can there be multiple data_type classes?
```suggestion
        # enumerations are created as ``EnumName([mixin_type, ...] [data_type,] enum_type)``
        superclass = subject.__mro__[-2]
```

picnixz: According to Python, no: [https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing](https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing):

> A new [Enum](https://docs.python.org/3/library/enum.html#enum.Enum) class must have one base enum class, up to one concrete data type, and as many [object](https://docs.python.org/3/library/functions.html#object)-based mixin classes as needed. 

The order of these base classes is:

```python
class EnumName([mix-in, ...,] [data-type,] base-enum):
    pass
```

However, thank you for pointing it since it also told me that it's possible to have *no* data-type class. In this case, the code still won't work properly. According to [https://docs.python.org/3/howto/enum.html#others](https://docs.python.org/3/howto/enum.html#others), a *data-type* class is a mixin class which defines a `__new__`, so I should technically check if this is the case or not which I will do this week-end.

picnixz: Actually, after re-reading the issue, what we wanted is not the data-type but the base enumeration type. The reason is that we want to exclude enum-related methods and keep only those from the mixin and possibly the data-type class.

picnixz: After some thoughts, here is what I decided to do:

- The original issue stems from the fact that some enumeration methods are not filtered out, like `_generate_next_value` (or smth like that).  As such, the functionality that we want is to be able to remove those private methods. 
- The way we achieved it before is to take `superclass = cls.__mro__[1]`, because we assumed that the class is of the form `class MyEnum(Enum)` leading to mro of `[MyEnum, Enum, object]`. 
- This caused the following issue: if we declare `class MyEnum(str, Enum)`, we end up with filtering out methods from `str` and not `Enum`. 

Now, if we use `cls.__mro__[-2]`, we will always target `Enum`. Even when subclassing another enum, e.g., `class Other(MyEnum)`, the MRO would still be `[Other, MyEnum, Enum, object]`. What we want is to also be able to keep methods from `MyEnum` in `Other`, so we don't want to actually exclude `MyEnum` methods, only those coming from `Enum`.

So we can actually directly use `Enum`. Now, this PR also solves the other issue that arose from the original issue. The other issue is that `dir(MyEnum)` does *not* return public methods or special ones, so, some methods are not rendered when inheriting from a mixin + enum type. However, that's another question. Some are exposed because they are indeed no in `Enum.__dict__` and that's fine, but at the same time, those that have same names as in `Enum.__dict__` such as special methods are still *not* rendered. Now, if `special-members` option is used, the special members will be included correctly (before, even with special-members option, they were not included because they were never stored).

picnixz:
@alxroyer Could you check if this patch would fix your issue?

picnixz:
Thanks to the new tests, I observed that there were issues. So it's still WIP

chrisjsewell:
> Thanks to the new tests, I observed that there were issues. So it's still WIP

Maybe convert to draft then 😄 ?

picnixz:
> Maybe convert to draft then 😄 ?

I fixed it actually, now I'm just cleaning it up.

picnixz:
So there is something with docutils but I don't know what.

@jayaddison can you investigate this one please?

picnixz:
There's definitely something. Maybe it's my fix that is causing the issue because it's really weird.

chrisjsewell:
> There's definitely something wrong with docutils HEAD. Maybe it's my fix that is causing the issue because it's really weird.

Lokks like `tests/test_builders/test_build_linkcheck.py::test_defaults` is flaky, because I had the same failure on my PR (just re-run and it dissapeared)

picnixz:
> Lokks like tests/test_builders/test_build_linkcheck.py::test_defaults is flaky, because I had the same failure on my PR (just re-run and it dissapeared)

It's funny because every time I rerun the test, it failed. So I am quite unlucky :(

picnixz:
I am really unlucky. I clearly don't know why this one became flaky now...

tests/test_extensions/test_ext_autodoc.py
```diff
@@ -1407,76 +1413,320 @@ def test_slots(app):
...
+    ) -> list[str]:
+        rst_options = dict.fromkeys(flags, '')
+        return self.entry(name, doc, role='method', args=args, indent=indent, **rst_options)
+
+    def member(self, name: str, value: Any, doc: str, *, indent: int = 3) -> list[str]:
+        rst_options = {'value': repr(value)}
+        return self.entry(name, doc, role='attribute', indent=indent, **rst_options)
+
+
+@pytest.fixture(scope='module')
```
chrisjsewell: seems like it would be safer to just have this as a function scope (given its returning a mutable)

sphinx/ext/autodoc/importer.py
```diff
@@ -20,16 +21,86 @@
...
     from sphinx.ext.autodoc import ObjectMember
 
 logger = logging.getLogger(__name__)
 
 
+def _filter_enum_dict(
+    enum_class: type[Enum],
+    attrgetter: Callable[[Any, str, Any], Any],
+    enum_class_dict: Mapping[str, object],
+) -> Generator[tuple[str, type, Any], None, None]:
```
chrisjsewell, line 41: Can you just add a doc string, briefly explaining the function and what it yields

chrisjsewell (changes requested):
Thanks, the general logic looks reasonable, and the test cases capture a good range of possibilities.
Just some nitpicks.

sphinx/ext/autodoc/importer.py
picnixz: Sure,. I'll do it in a few moments

tests/test_extensions/test_ext_autodoc.py
picnixz: Seems reasonable.

sphinx/ext/autodoc/importer.py
```diff
@@ -20,16 +21,86 @@
...
+    # enumerations are created as ``EnumName([mixin_type, ...] [member_type,] enum_type)``
+
+    # attributes that were found on a mixin type or the data type
+    candidate_in_mro: set[str] = set()
+    # see: https://docs.python.org/3/howto/enum.html#supported-dunder-names
+    sunder_names = {'_name_', '_value_', '_missing_', '_order_', '_generate_next_value_'}
+    # sunder names that were picked up (and thereby allowed to be redefined)
```
picnixz: For myself: move this comment one line above

picnixz (commented):
I'm considering improving the logic of checking autodoc lines. For now I'll leave the current formatter as a base implementation but I'll review it in another PR so that we can easily match autodoc related content instead of having to write those lines with strings, indents and empty lines that are sometimes hard to remember.

tests/test_extensions/test_ext_autodoc.py
```diff
@@ -1407,76 +1413,411 @@ def test_slots(app):
...
+
+    actual = do_autodoc(app, 'class', fmt.target, options)
+    assert list(actual) == [
+        *fmt.brief('this is enum class'),
+        *fmt.method('say_goodbye', 'a classmethod says good-bye to you.', 'classmethod'),
+        *fmt.method('say_hello', 'a method says hello to you.'),
+        *fmt.member('val1', 12, 'doc for val1'),
+        *fmt.member('val2', 23, 'doc for val2'),
+        *fmt.member('val3', 34, 'doc for val3'),
+        *fmt.member('val4', 34, ''),  # val4 is alias of val3
```
jayaddison, line 1513: This seems odd to me - the values of `val3` and `val4` could diverge, couldn't they?  (they're both initialized to `34`, but could change?)

picnixz: I purposedly have an alias. And no the can't change because it's an enumeration member.

jayaddison: Ah, I missed that it was an enum member, that explains it.  Thanks!

jayaddison (approved):

chrisjsewell (approved):
Looks good thanks

PR #11597, opened 2023-08-15
Fix MathJax lazy loading.

### LINKED ISSUE #9686: Current mathjax loading on per-page basis breaks math in table of contents
### Describe the bug

When a page doesn't use mathjax (:math:) it doesn't load mathjax. This breaks titles in the table of contents that do use math. 

### How to Reproduce

- perform a sphinx-quickstart
- add 'sphinx.ext.mathjax' to extensions
- create otherpage.rst
```
:math:`1+1=\alpha`
==================

Test page with math in title.
```

add otherpage to toctree of index.rst.
- make html

### Expected behavior

Math is properly rendered in table of contents of index.html

### Your project

n/a

### Screenshots

2021-09-29-133959_1084x394_scrot

### OS

Linux

### Python version

3.9.1

### Sphinx version

3.5.1 and also 4.2.0 (probably also newer)

### Sphinx extensions

sphinx.ext.mathjax

### Extra tools

_No response_

### Additional context

See implementation on per-page loading https://github.com/sphinx-doc/sphinx/issues/6241

### LINKED ISSUE #10192: There is no way to enforce MathJax when maths only appear in TOC trees.
This is almost a bug report.

Sphinx behaves such that if there is no math in an rst document, MathJax javascript is not included in the generated html file. This is to save loading time. But it frequently happens that maths only appear in the TOC trees of the rst document, in which case maths are not rendered properly because MathJax is not activated. For example, see

https://doc.sagemath.org/html/en/reference/schemes/index.html

It seems that there is no mechanism to enforce MathJax in this situation. It would be nice if we can enable this enforcement through the conf file.

### LINKED ISSUE #11483: Mathjax not working with singlehtml
### Describe the bug

This is probably related to #9686.

When using the mathjax extension, math is working in the html output but not in the singlehtml output.

Manually adding the line

    <script async="async" src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>

which is missing in the singlehtml output but present in the html output makes math work well.

### How to Reproduce

Run sphinx with mathjax extension on a documentation containing some math with the singlehtml target. The actual test case that was used uses markdown files and math is not present in the main rst file but in one of the later included files (in case that makes a difference, which may be the case reading #9686).

### Environment Information

```text
This is running in a linux CI environment. I can reproduce it locally with sphinx 7.0.1.
```

### Sphinx extensions

```python
myst_parser, sphinx_math_dollar, sphinx.ext.mathjax
```

### Additional context

_No response_

picnixz:
Fix #9686 (html output).
Fix #10192 (math in toctree).
Fix #11483 (singlehtml output).

- html: MathJax is not loaded when the math equation is in a title and the document is included in a toctree. At least, this does not require to always include HTML assets (which may be an overkill when you have a lot of pages but only one containing math equations). 
- singlehtml: MathJax is only loaded depending on whether `index.rst` contains math equations or not. Previously, the hack was to use `set_html_assets_policy('always')`, but this may have side-effects in the future.


AA-Turner (approved):

picnixz:
I'll update the CHANGES as well.

CHANGES.rst
```diff
@@ -23,6 +23,11 @@ Bugs fixed
 * #11617: ANSI control sequences are stripped from the output when writing to
   a warnings file with :option:`-w <sphinx-build -w>`.
   Patch by Bénédikt Tran.
+* #9686: html: Fix MathJax lazy loading when equations appear in titles.
```
```suggestion
* #9686: html builder: Fix MathJax lazy loading when equations appear in titles.
```

sphinx/domains/math.py
```diff
@@ -134,11 +134,14 @@ def get_objects(self) -> Iterable[tuple[str, str, str, str, str, int]]:
...
     def has_equations(self, docname: str | None = None) -> bool:
-        if docname:
-            return self.data['has_equations'].get(docname, False)
-        else:
+        if not docname:
             return any(self.data['has_equations'].values())
 
+        return (
+            self.data['has_equations'].get(docname, False)
+            or any(self.has_equations(d) for d in self.env.toctree_includes.get(docname, ()))
```
AA-Turner: Would save line length, but feel free to reject.
```suggestion
            or any(map(self.has_equations, self.env.toctree_includes.get(docname, ())))
```

PR #11602, opened 2023-08-16
Add tox ``lint`` target

### LINKED ISSUE #11599: tox targets for non-tests
> @AA-Turner Could I ask what was the rationale for this change? How are contributors expected to run e.g. linters on their changes without the tox targets?

_Originally posted by @stephenfin in https://github.com/sphinx-doc/sphinx/commit/ff852bc7c31f48e66100e4647749fc199d92ca79#commitcomment-124548969_

stephenfin:
### Feature or Bugfix

Refactoring

### Purpose

Re-add a tox environment for linters. This environment uses the 'lint' extras
and runs the same commands as the GitHub Workflows. A note is added reminding
people to keep the two in-sync.

### Detail

This resolves #11599. There is an (arguably superior but more involved) alternative at #11603.

### Relates

(none)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,6 @@
 [tox]
-minversion = 2.4.0
+minversion = 4.2.0
 envlist = py{39,310,311,312,313}
-isolated_build = True
 
 [testenv]
 usedevelop = True
@@ -30,7 +29,6 @@ commands=
     python -X dev -X warn_default_encoding -m pytest --durations 25 {posargs}
 
 [testenv:docs]
-basepython = python3
 description =
     Build documentation.
 extras =
@@ -40,7 +38,6 @@ commands =
     sphinx-build -M {env:BUILDER:html} ./doc ./build/sphinx -nW --keep-going {posargs}
 
 [testenv:docs-live]
-basepython = python3
 description =
     Build documentation.
 extras =
--- a/tox.ini
+++ b/tox.ini
@@ -28,6 +28,19 @@ setenv =
 commands=
     python -X dev -X warn_default_encoding -m pytest --durations 25 {posargs}
 
+[testenv:lint]
+description =
+    Run linters.
+extras =
+    lint
+# If you update any of these commands, don't forget to update the equivalent
+# GitHub Workflow step
+commands =
+    ruff . --diff --format github
+    flake8 .
+    isort --check-only --diff .
+    mypy sphinx/
+
 [testenv:docs]
 description =
     Build documentation.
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -26,8 +25,20 @@ setenv =
...
+    Run linters.
+extras =
+    lint
+# If you update any of these commands, don't forget to update the equivalent
+# GitHub Workflow step
+commands =
+    ruff . --diff --format github
+    flake8 .
+    isort --check-only --diff .
+    mypy sphinx/
```
picnixz, line 42: Don't we also apply mypy to tests?

tox.ini
```diff
@@ -1,7 +1,6 @@
 [tox]
-minversion = 2.4.0
+minversion = 4.2.0
 envlist = py{39,310,311,312}
-isolated_build = True
```
picnixz, line 4: Why did you remove the isolated build? (I don't really use tox personally and I run the tests from CLI using specific Python version)

stephenfin: All the context is in the commit messages. From [`f2f2e79` (#11602)](https://github.com/sphinx-doc/sphinx/pull/11602/commits/f2f2e79a9d5a221c50f3eeb08dc0dfce48ac5d08):

> `isolated_build` is no longer required since builds are isolated by default in tox 4.
> 
> `basepython` is no longer required since tox 4 only supports Python 3 and 'basepython' defaults to the version of Python that tox is installed with.

picnixz: Oh sorry, I didn't see the commit message itself. My bad !

tox.ini
stephenfin: Apparently not: I simply copied these commands from `.github/workflows/lint.yml`

picnixz: Oh good to know. @AA-Turner any reason why we are not applying mypy to tests actually ? (is it perhaps because there would be too many false positive due to monkey patching or other weird constructions?)

stephenfin: > (is it perhaps because there would be too many false positive due to monkey patching or other weird constructions?)

Not @AA-Turner but that's usually why I don't run mypy over tests, yes.

electric-coder:
@picnixz 

> Oh good to know. @AA-Turner any reason why we are not applying mypy to tests actually? 

Most likely because it can be a lot of extra work and maintainers didn't prioritize it. However, it would improve code quality.

@stephenfin

> Not @AA-Turner but that's usually why I don't run mypy over tests, yes.

Absolutely worth it even if you have to use `# ignore` on occasion.

PR #11251, opened 2023-03-19
use black pre-commit

12rambau:
Subject: lint the lib with pre-commits

### Feature or Bugfix

Refactoring

### Purpose
- As mentione in the associated issue if one want's to participate in the development of Sphinx, the linter is only executed on Github side. using pre-commits, we could get a more consistent linting test (as everyone would run the same version on the same generated envorionment) and make sure anything that make its way to Github is nicely shaped. 

### Detail
- [x] add pre-commits 
- [ ] execute pre-commits in the linting github action
- [x] use black (https://github.com/psf/black)
- [ ] use ruff (https://github.com/charliermarsh/ruff)

### Relates
Fix #11205

PR #11603, opened 2023-08-16
Integrate pre-commit

### LINKED ISSUE #11205: automation of linting test using pre-commit

stephenfin:
### Feature or Bugfix

Refactoring

### Purpose

Integrate `pre-commit` for linting. This is an exceptionally useful tool that provides a mechanism to automate all sorts of linters, including all of those that we currently support and rely on. It can also be used as a standalone tool, as demonstrated by the GitHub Workflows used within.

This change also adds a `lint` tox environment, which is simply a wrapper around `pre-commit` (for those used to this workflow), and migrates the various linter-related GitHub Action steps to use `pre-commit`. The latter is important since without this, configurations will drift. Developers who wish to configure their environment manually or in a different manner can continue to do so, but the vast majority can benefit from automated linting and fixing.

As a future change, we may wish to integrate [pre-commit.ci](https://pre-commit.ci/) since that will automate the fixing of PRs using pre-commit (currently we just check if they would require fixes) but that's a separate issue.

> **Note**
> The `pre-commit` tool utilises the `pre-commit` client-side hook. Client-side hooks are not included in clones and must be enabled manually by the user. In addition, the `pre-commit` tool itself must be installed and enabled on the `sphinx` repo to enable the pre-commit check workflow. As such, this is entirely optional and can be ignored by contributors with different workflows or those who do not wish to use `pre-commit`.

### Detail

This is an alternative to #11602 and potentially #11375. It resolves #11205.

### Relates

- #11602
- #11375
- #11205

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,41 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.284
+    hooks:
+      - id: ruff
+        args: ['--diff', '--format', 'github']
+
+  - repo: https://github.com/pycqa/flake8
+    rev: 6.1.0
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          - flake8-simplify
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.12.0
+    hooks:
+      - id: isort
+        name: isort (python)
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.5.0
+    hooks:
+      - id: mypy
+        additional_dependencies:
+          - docutils-stubs
+          - types-requests
+        files: ^sphinx/
+
+  - repo: https://github.com/sphinx-contrib/sphinx-lint
+    rev: v0.6.8
+    hooks:
+      - id: sphinx-lint
+        args: ['--enable', 'line-too-long', '--max-line-length', '85']
+        files: |
+          (?x)^(
+            CHANGES
+            | CONTRIBUTING.rst
+            | README.rst
+            | doc/.*
+          )
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,12 @@
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-json
+      - id: check-yaml
+      - id: debug-statements
+        exclude: '^sphinx/cmd/build.py$'
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.0.284
     hooks:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,19 +25,9 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3"
-    - name: Install pip
-      run: python -m pip install --upgrade pip
-
-    - name: Install known good Ruff
-      run: python -m pip install ruff==0.0.284
-    - name: Lint with known good Ruff
-      run: ruff . --format github
-
-    - name: Install latest Ruff
-      run: python -m pip install --upgrade ruff
-    - name: Lint with latest Ruff
-      continue-on-error: true
-      run: ruff . --format github
+    - uses: pre-commit/action@v3.0.0
+      with:
+        extra_args: ruff --all-files
 
   flake8:
     runs-on: ubuntu-latest
@@ -48,12 +38,9 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3"
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install --upgrade "flake8>=3.5.0" "flake8-simplify"
-    - name: Lint with flake8
-      run: flake8 .
+    - uses: pre-commit/action@v3.0.0
+      with:
+        extra_args: flake8 --all-files
 
   isort:
     runs-on: ubuntu-latest
@@ -64,12 +51,9 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3"
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install --upgrade isort
-    - name: Lint with isort
-      run: isort --check-only --diff .
+    - uses: pre-commit/action@v3.0.0
+      with:
+        extra_args: isort --all-files
 
   mypy:
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


stephenfin:
Looks like [`545cce2` (#11603)](https://github.com/sphinx-doc/sphinx/pull/11603/commits/545cce20b30e8ed51ee787cdbd713f51fe24f31c) had knock-on implications for other tests. Will address.

12rambau:
I've already tried twice to propose PR to integrate pre-commits and they were all refused because involving "too much changes" my last one (that you linked) is creating the file itself but only runs ruff. Can someone provides review there ? 

That's already the second time that I try to contribute to sphinx and that my PR is closed because overridden by someone else. Is it common practice here ? if yes I'll stop trying and will simply open well documented issues.

stephenfin:
> I've already tried twice to propose PR to integrate pre-commits and they were all refused because involving "too much changes" my last one (that you linked) is creating the file itself but only runs ruff. Can someone provides review there ?
> 
> That's already the second time that I try to contribute to sphinx and that my PR is closed because overridden by someone else. Is it common practice here ? if yes I'll stop trying and will simply open well documented issues.

It's not common practice. However, you've overloaded the PRs in all cases. Rather than just focusing on simply integrating a new tool, the PRs have also reformatted the entire code base with a different tool (#11251) or changed the rules for an existing linter (#11375). I suspect you'd have more luck if you tackled these topics in separate PRs.

stephenfin:
@AA-Turner I don't want to merge this without getting your agreement on this, given the imbalance in number of contributions of late. As such, could I get your ack on this? fwict, you're using a local virtual environment to (presumably) enable IDE integration? If so, this should have no significant impact for you. I suspect the only real change would be whenever we want to bump a linter version we would now update the pacakge version in `.pre-commit-config.yaml` rather than in `.github/workflows/lint.yaml` (so a different YAML file). From my end, this would make my own workflow *significantly* easier since it aligns with how things work in virtually ever project I contribute to (I work in vim and rely on pre-commit to catch lint failures as I commit).

AA-Turner:
> fwict, you're using a local virtual environment to (presumably) enable IDE integration?

I use a single conda environment (through the `mamba` reimplementation) for development of Sphinx that contains dependencies and associated tools (e.g. ``sphinx-lint``, ``ruff``). I then run these tools from the command line, though I use PyCharm as an editor. My main objection continues to be that I don't understand the point/benefits of pre-commit -- every tool is faster when run on its own, and we don't have a suite of thousands of linting tools. Personally, I run ``ruff`` every so often, and ``python -m sphinx build -M html doc build -anETW --keep-going`` if I've changed the documentation. ``isort`` and ``flake8`` could probably be removed at this point, but I'm not 100% sure of the Ruff implementation so they're kept for surety.

> this should have no significant impact for you

If I make no changes, yes, though I feel that contributors ought adopt the guidelines set out by the project, so I'd feel obliged to change to using ``pre-commit``, which would slow down my workflow -- I tend to make very frequent small commits that I later extensively rebase into a final PR. If each of these had a 5-15 second delay it would be rather annoying.

The only reason I don't use ``tox`` is that I use Windows & I have had various problems with it -- I'd prefer to remove it entirely, but I recognise that it is used by various people.

A

jayaddison:
Although I thinking linting and enforcement of code style using continuous integration, I often feel a bit conflicted about `pre-commit`.  Linting sets a policy for code style but doesn't restrict the tools that developers can use, or have to use, to meet that policy.  `pre-commit` goes slightly further and channels developers through a certain set of workflow tools.  From a diversity and choice perspective that makes me question it.

(I tend to use `vim` only, and manually run lint-checkers and tests from the commandline where possible.  I don't doubt that it makes me less productive under many circumstances, and perhaps the resulting corrective commit noise is annoying sometimes, but it reassures me that I'm contributing work that I've developed -- and hopefully understand and remember -- by myself)

To try to make that feedback more relevant here: I'd personally prefer if this were opt-in by default rather than opt-out.

12rambau:
> Linting sets a policy for code style but doesn't restrict the tools that developers can use
At the end of the day for a PR to be validated, you need to comply with the ruff checks, so what tool choices do you have ? pre-commit only brings to your local development the linting operation that are/should be tested in the CI.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+        args: ['--diff', '--format', 'github']
+
+  - repo: https://github.com/pycqa/flake8
+    rev: 6.1.0
+    hooks:
+      - id: flake8
+        additional_dependencies:
+          - flake8-simplify
+
+  - repo: https://github.com/pycqa/isort
```
12rambau, line 23: isn't `isort` covered by `ruff`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - id: debug-statements
+        exclude: '^sphinx/cmd/build.py$'
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.284
+    hooks:
+      - id: ruff
+        args: ['--diff', '--format', 'github']
+
+  - repo: https://github.com/pycqa/flake8
```
12rambau, line 16: isn't `flake8` covered by `ruff`?

jayaddison:
I'm probably being too stubborn - it was the mention of "(pass -n to `git commit`)" that made me think that this would require opt-out.  I'm not hugely familiar with `git` hooks, but I seem to remember that they can be activated-by-default in some repos.

As for alternatives: there are many linters, but there's only one `pre-commit` -- it seems to be a widely-used de-fact standard.  I'm sort-of-ok with that, but personally it feels slightly more dynamic (in terms of downloading binaries from `git` repositories that run on local developer machines) than I'm comfortable with.  I'm not aware of any problems with it, but it's the kind of project that I'd be terrified to maintain (due to the level of usage and effect it can have, and the potential levels of responsibility and pressure as a result), and that makes me cautious too.

And to be honest: sometimes I do just code a fix (or equally often, poor attempt at one) and then lean on continuous integration to catch problems.  That's more likely in cases where I don't totally trust the install processes for the continuous integration components.

stephenfin:
To be clear, I want to re-add the ability to install and run the correct version of all our linters with one command. As far as I'm aware, our three options here are pre-commit, tox, and nox. I have a preference for pre-commit since it's can be automatic (if installed) and is easily wrapped by tox as I've done here, but I will settle for any of the three. However, whatever tool we settle on should be used by CI to prevent it going stale.

AA-Turner:
I'm keeping an eye on [this thread](https://discuss.python.org/t/67194), as introducing a standardised command mechanism would meet the goal of running a single command and mean that using pre-commit is still optional. We may also get this 'for free' if we switch to `uv`, `rye`, or a similar tool for workflow management (as a different PR proposes). However for now closing this one -- if no progress is made then we can revisit this.

A

PR #11610, opened 2023-08-17
Fix signature of docutils include_source monkeypatch

ksunden:
The docutils `insert_input` method's second argument is called 'source', not 'path'

### Feature or Bugfix

- Bugfix

### Purpose

The monkeypatching of docutils broke the matplotlib `plot` directive, which calls `insert_input` and passes the `source` parameter by name.
This PR simply matches the signature of the monkeypatch to that of the upstream method such that calling methods are not affected.

### Relates
- Introduced by #11510


AA-Turner:
This will be in 7.2.2

PR #11624, opened 2023-08-20
Remove ANSI colors when writing to external warnings file.

### LINKED ISSUE #11574: Log file stream is not closed when calling sphinx through api
### Describe the bug

Hi,

we have built a wrapper around sphinx for internal use in our company.
Therefore, we do not call sphinx through cli, but through its api.
I've noticed, that the log-file that I am passing to the sphinx instance is not being closed after the build has finished.
Thus, I can't move or remove it within the same process. as it is still opened.

I dont know if this behavior is intended or maybe sphinx does not care about it, as the tool is supposed to be called through its cli.
If thats the case, feel free to just close this issue. Otherwise, maybe I can help implementing a fix for this.
We would be very happy to be able to use the logfile inside the same process after the build process.

BR
Valentin

### How to Reproduce

Minimal working example:

- Any working sphinx project inside a folder (in this case "test_proj").
- A simple test script:

```test.py
import os
from sphinx.cmd.build import build_main

args = ["-b", "html", "test_proj", "test_proj/_build", "-w", "log_file.txt"]
build_main(args)

os.remove("log_file.txt")
```

### Environment Information

```text
Platform:              win32; (Windows-10-10.0.22000-SP0)
Python version:        3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)])
Sphinx version:        7.1.2
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Sphinx extensions

_No response_

### Additional context

_No response_

### LINKED ISSUE #11617: Don't add ANSI colour codes to Sphinx -w warnings file
### Describe the bug

If the terminal supports colour, Sphinx outputs in colour. This is good, it improves readability.

But if outputting to a file, it also adds ANSI colour codes, where is shouldn't.

### How to Reproduce

`index.rst` with an error, in this case, a typo in the role:
```rst
:peep:`8`
```

And an empty `conf.py` will do.

1. Build with warnings and output to a file: `sphinx-build -b html . _build -n -w sphinx-warnings.txt`

image

3. View the file in an editor:

Actual result:

```
[31m/private/tmp/bug-report/index.rst:1: ERROR: Unknown interpreted text role "peep".[39;49;00m
```

Because Sphinx is correctly detecting my terminal can do colour, it's correctly adding ANSI colour codes to the terminal output.

But it's also incorrectly adding them to the text file. 

Expected result:

Colour in the terminal but no ANSI colour codes in the text file:

```
/private/tmp/bug-report/index.rst:1: ERROR: Unknown interpreted text role "peep".
```

### Environment Information

```text
Platform:              darwin; (macOS-13.5-arm64-arm-64bit)
Python version:        3.11.4 (v3.11.4:d2340ef257, Jun  6 2023, 19:15:51) [Clang 13.0.0 (clang-1300.0.29.30)])
Sphinx version:        7.2.2
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Sphinx extensions

```python
None.
```

### Additional context

_No response_

picnixz:
Fix #11617.
Fix #11574.

In addition, I needed to fix #11574 so that testing can be done properly (otherwise the file descriptor is kept opened in write mode and it is not possible to correctly read it).

@AA-Turner Can you bump the development version 7.2.3 (or 7.3) so that I can add a CHANGES entry please?


mgeier:
Is the `--keep-colors` option really necessary?

I can't imagine anyone wanting those escape sequences in their warnings file.

More importantly, I think adding this option would cause confusion with the already existing `--color`  and `--no-color` options.

picnixz:
Err.. I don't really know. Like sometimes you may want to keep the colors when you print them using some weird commands. Or for filtering. 

It was just to ensure that if there was someone out there doing it, then they don't need to change much (although I would recommend using `tee` instead then).

picnixz:
On a second though, I think not having that option would be the best. People that want to keep colora should simply tee the standard output I think.

picnixz:
So, I removed `--keep-colors` since it's indeed not a good idea to keep (and it simplified the flow).

sphinx/cmd/build.py
```diff
@@ -307,8 +311,11 @@ def _parse_confoverrides(
     return confoverrides
 
 
-def build_main(argv: Sequence[str]) -> int:
-    """Sphinx build "main" command-line entry."""
+def build_main(argv: Sequence[str], *, closefd: bool = False) -> int:
+    """Sphinx build "main" command-line entry.
+
+    If *closefd* is true, close the file descriptors were opened by Sphinx.
+    """
```
AA-Turner: Thoughts on removing the ``closefd`` option and simply always closing?

sphinx/cmd/build.py
```diff
@@ -330,10 +338,17 @@ def build_main(argv: Sequence[str]) -> int:
...
         handle_exception(app, args, exc, args.error)
-        return 2
+        retcode = 2
+
+    if warnfp is not None and closefd:
+        # close a file descriptor opened by Sphinx internally
+        # (only "second" stream of the Tee object can be safely
+        # closed since the first one may be sys.stderr).
+        warnfp.close()
+    return retcode
```
AA-Turner: I think it's cleaner to keep return and use finally here.

A

AA-Turner (changes requested):

sphinx/cmd/build.py
picnixz: I'm pretty sure I did this for a reason, but now I don't remember why. And now, I think it's not good to leave the fd opened since it isn't leaked outside of the function anyway. So yes, I'll do that.

sphinx/cmd/build.py
picnixz: Ah yes, thank you! the `finally` clause is always something that I forget to use 50% of the time.

picnixz:
I'll make the changes today or tomorrow and add a CHANGES entry as well.

CHANGES.rst
```diff
@@ -20,6 +20,9 @@ Bugs fixed
   Patch by Vinay Sajip.
 * #11622: Ensure that the order of keys in ``searchindex.js`` is deterministic.
   Patch by Pietro Albini.
+* #11617: ANSI control sequences are stripped from the output when writing to
+  an external file.
```
```suggestion
* #11617: ANSI control sequences are stripped from the output when writing to
  a warnings file with :option:`-w <sphinx-build -w>`.
```

AA-Turner:
Thanks @picnixz!

PR #11628, opened 2023-08-21
variable referenced before assignment fix.

jchanvfx:
Subject: bugfix to some variables being referenced before assignment.

### Feature or Bugfix
- Bugfix

```
/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/autodocsumm/__init__.py:223: RemovedInSphinx80Warning: The tuple interface of ObjectMember is deprecated. Use (obj.__name__, obj.object) instead.
  members = [(membername, member) for (membername, member) in members

Exception occurred:
  File "/opt/hostedtoolcache/Python/3.9.17/x64/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py", line 779, in filter_members
    self.name, membername, member, exc, type='autodoc')
UnboundLocalError: local variable 'membername' referenced before assignment
```

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -675,12 +675,12 @@ class Documenter:
 
         # process members and determine which to skip
         for obj in members:
-            try:
-                membername = obj.__name__
-                member = obj.object
-                # if isattr is True, the member is documented as an attribute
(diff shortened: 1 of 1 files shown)


AA-Turner:
Please may you add a test?

A

AA-Turner:
Fixed by #11651 and in Sphinx 7.2.4. Thanks for the PR!

A

PR #11634, opened 2023-08-22
Fixed two more bugs with SVG inheritance-diagram links

ayshih:
Subject: Fixed two more bugs with SVG inheritance-diagram links

### Feature or Bugfix

- Bugfix

### Purpose
This PR is a follow-on to fixes to inheritance-diagram links (#10614), because I discovered two bugs when trying to re-enable SVG inheritance diagrams on matplotlib (see matplotlib/matplotlib#26567).  One of these bugs is actually in the graphviz writer, so affects use cases beyond just inheritance diagrams.

### Detail
- The bug in the graphviz writer (see #11078) is that it uses `self.builder.outdir` (e.g., `build/html`) as the base directory for relative URLs rather than the directory of the current document (e.g., `build/html/api` for `build/html/api/doc1.rst`).  This happens to still work fine for many cases because the base directory can "cancel out" when determining the correct relative path from the image directory (e.g., `build/html/_images`).  However, when the relative URL doesn't need to go all the way down to `self.builder.outdir` (e.g., `build/html/api/doc1.rst` pointing to `build/html/api/doc2.rst` means that the relative URL is simply `doc2.rst`), parts of the provided base directory get pulled into the generated relative path, and so passing in the wrong base directory results in an incorrect relative path (e.g., `../html/doc2.rst` instead of `../api/doc2.rst`).
- When the inheritance diagram is in the same document as the class description, #10614 accidentally kept the relative path from `self.builder.outdir` instead of using just the filename.

### Relates
- matplotlib/matplotlib#26567


AA-Turner:
Looking at 7.2.5.

A

sphinx/ext/graphviz.py
```diff
@@ -230,16 +231,15 @@ def fix_svg_relative_paths(self: SphinxTranslator, filepath: str) -> None:
...
 
-        old_path = path.join(self.builder.outdir, url)
-        new_path = path.relpath(
-            old_path,
-            start=path.join(self.builder.outdir, self.builder.imgpath),
-        )
+        doc_dir = (self.builder.outdir / self.builder.current_docname).parent
+        old_path = doc_dir / rel_uri
+        img_path = doc_dir / self.builder.imgpath
+        new_path = path.relpath(old_path, start=img_path)
```
AA-Turner, line 247: What's the difference here with:

```suggestion
        new_path = path.relpath(rel_uri, start=self.builder.imgpath)
```

CHANGES
```diff
@@ -16,6 +16,10 @@ Features added
 Bugs fixed
 ----------
 
+* #11634: Fixed two more bugs in inheritance diagrams that resulted in
+  broken links.
```
AA-Turner: Perhaps? (I think the changelog should at least describe the type of bug/roughly what was fixed)
```suggestion
* #11634: Fixed inheritance diagram relative link resolution
  for sibling files in a subdirectory.
```

AA-Turner:
@ayshih would you be able to look over the changes I've made here?

A

ayshih:
Hmm, one of your changes has caused this PR to no longer fix the matplotlib docs, so I'll track down why (and also expand the test accordingly).

AA-Turner:
[953f2f6](https://github.com/sphinx-doc/sphinx/pull/11634/commits/953f2f67c2d413786fadc73932ced5b62919649c) probably fixes matplotlib, but it reintroduces the problem I wanted to solve (``current_docname`` is not a universal attribute).

A

ayshih:
I have verified that the matplotlib docs build correctly with the most recent changes.  I have also expanded the unit test to cover what was being missed (linking between files in the same subdirectory).

PR #11635, opened 2023-08-22
 Support positional-only parameters in classmethods 

AA-Turner:
cc: @randolf-scholz -- please could you check that this works for you?

A

### DIFF
--- a/sphinx/ext/autodoc/preserve_defaults.py
+++ b/sphinx/ext/autodoc/preserve_defaults.py
@@ -180,9 +180,9 @@ def update_defvalue(app: Sphinx, obj: Any, bound_method: bool) -> None:
             # __signature__ can't be set directly on bound methods.
             obj.__dict__['__signature__'] = sig
     except (AttributeError, TypeError):
-        # Failed to update signature (e.g. built-in or extension types):
-        # "obj" might not have __dict__, e.g. when decorated with a class
-        # that defines __slots__.
+        # Failed to update signature (e.g. built-in or extension types).
+        # For user-defined functions, "obj" may not have __dict__,
+        # e.g. when decorated with a class that defines __slots__.
         # In this case, we can't set __signature__.
         return
     except NotImplementedError as exc:  # failed to ast_unparse()
--- a/sphinx/ext/autodoc/preserve_defaults.py
+++ b/sphinx/ext/autodoc/preserve_defaults.py
@@ -142,8 +142,8 @@ def update_defvalue(app: Sphinx, obj: Any, bound_method: bool) -> None:
             return
 
         if args.defaults or args.kw_defaults:
-            is_classmethod = bound_method and inspect.ismethod(obj)
-            if is_classmethod and hasattr(obj, '__func__'):
+            is_bound_method = bound_method and inspect.ismethod(obj)
+            if is_bound_method and hasattr(obj, '__func__'):
                 sig = inspect.signature(obj.__func__)
             else:
                 sig = inspect.signature(obj)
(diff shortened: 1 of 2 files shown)


sphinx/ext/autodoc/preserve_defaults.py
```diff
@@ -142,7 +142,11 @@ def update_defvalue(app: Sphinx, obj: Any, bound_method: bool) -> None:
             return
 
         if args.defaults or args.kw_defaults:
-            sig = inspect.signature(obj)
+            is_classmethod = bound_method and inspect.ismethod(obj)
```
randolf-scholz: Is it clear that this catches `classmethod` and `classmethod` only? (and not any otherwise decorated things?)
Possibly it would be better to make a helper function.

sphinx/ext/autodoc/preserve_defaults.py
```diff
@@ -142,7 +142,11 @@ def update_defvalue(app: Sphinx, obj: Any, bound_method: bool) -> None:
...
+            if is_classmethod and hasattr(obj, '__func__'):
+                sig = inspect.signature(obj.__func__)
```
randolf-scholz: If the `is_classmethod` is correct, then `__func__` should always exist, so I'd rather do

```python
if is_classmethod:
    assert hasattr(obj, '__func__')
    sig = inspect.signature(obj.__func__)
```

Which would help to find bugs if there are any in `is_classmethod`.

randolf-scholz:
@AA-Turner The patch works well for the projects I use sphinx in :+1:, but I added some comments to code sections that could be problematic in some edge cases regardless.

AA-Turner:
Hi @randolf-scholz -- thank you, but I can't see any comments -- please may you check?

A

randolf-scholz:
@AA-Turner These don't show up for you?

image

AA-Turner:
Ahh, they're pending comments -- you need to go in to the Files tab & then finish the review (Review Changes -> Comment)

image

sphinx/ext/autodoc/preserve_defaults.py
```diff
@@ -165,13 +169,8 @@ def update_defvalue(app: Sphinx, obj: Any, bound_method: bool) -> None:
...
-            if bound_method and inspect.ismethod(obj):
-                # classmethods
-                cls = inspect.Parameter('cls', inspect.Parameter.POSITIONAL_OR_KEYWORD)
-                parameters.insert(0, cls)
-
             sig = sig.replace(parameters=parameters)
-            if bound_method and inspect.ismethod(obj):
+            if is_classmethod:
                 # classmethods can't be assigned __signature__ attribute.
                 obj.__dict__['__signature__'] = sig
```
randolf-scholz: This is problematic since `obj` might not have `__dict__`. This can happen if it is decorated with a class that uses `__slots__`. It's a pretty cursed example/edge-case...

```python
from functools import wraps


def make_cursed_decorator(func):
    class CursedDecorator:
        __slots__ = ()

        @staticmethod
        @wraps(func)
        def __call__(*args, **kwargs):
            return func(*args, **kwargs)

    return CursedDecorator()


class A:
    @classmethod
    @make_cursed_decorator
    def foo(cls, x: int, /, *, y: int = 0) -> int:
        return x + y


A.foo(1, y=2)
A.foo.__dict__  # AttributeError
```

sphinx/ext/autodoc/preserve_defaults.py
AA-Turner: Good point, this captures all bound methods.

sphinx/ext/autodoc/preserve_defaults.py
AA-Turner: As I'd like this to go in to 7.2.3 I don't want to introduce a new assert.

sphinx/ext/autodoc/preserve_defaults.py
AA-Turner: The outer ``except (AttributeError, TypeError):`` would catch this, but it isn't very clear -- I've added a note to capture this case.

randolf-scholz (commented):
LGTM

PR #11638, opened 2023-08-23
Allow relative-to-root toctree dirs in autosummary

### LINKED ISSUE #11637: Allow prefix `/` in autosummary's `toctree` to refer to base directory

At the moment, Autosummary's `toctree` directive is always relative to the current file.  This means that the generated files always move with the file that contains an toctree-creating `autosummary` directive. In our case, we'd like to universally put the generated files into one separate folder at the documentation root, regardless of where they were generated from.  This is particularly a problem when dealing with recursive autosummary generations, and at the moment is effectively limiting us to our manual documentation all having a flat folder structure.

Perhaps a leading `os.sep` on the `toctree` directory could be interpreted as relative to the base documentation directory?

Nothing much short of leaving all our documentation in a flat structure / moving all the generated files to be next to the file that generated them.

**Additional context**

None.

(*edit*: I'd be happy to do this myself if it's an acceptable direction)

jakelishman:
Subject: Allow relative-to-root toctree dirs in autosummary

### Feature or Bugfix
- Feature

### Purpose

At the moment, Autosummary's `toctree` directory is always relative to the current file. This means that the generated files always move with the file that contains an toctree-creating `autosummary` directive. In our case, we'd like to universally put the generated files into one separate folder at the documentation root, regardless of where they were generated from. This is particularly a problem when dealing with recursive autosummary generations, and at the moment is effectively limiting us to our manual documentation all having a flat folder structure.

### Detail

This treats a leading `/` in the directory of a `toctree` option in an `autosummary` block as requesting a directory relative to the source documentation root, rather than relative to the current directory.  Previously, prefixing a path with `/` would typically cause build failures on Unix-likes because

```python
posixpath.join("/prefix/path", "/other") == "/other"
```

and Sphinx (hopefully!) didn't have permission to create directories in the filesystem root.  This means that the change should not conflict with any existing practical working documentation builds.

Autosummary currently uses `posixpath` for most operations even on Windows, so the root indicator is only permitted as `/`, not `os.sep`.

### Relates
- Fix #11637

### DIFF
--- a/doc/usage/extensions/autosummary.rst
+++ b/doc/usage/extensions/autosummary.rst
@@ -81,15 +81,16 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
      directory. If no argument is given, output is placed in the same directory
      as the file that contains the directive.  The directory is interpreted as
      relative to the directory of the source file, unless it begins with the
-     path separator ``/``, in which case it is relative to the documentation
-     root.  All paths should use the separator ``/`` regardless of the
-     operating system used for the build.
+     ``/``, in which case it is relative to the documentation source directory.
+     All paths should use the separator ``/`` regardless of the operating system
+     used for the build.
 
      You can also use ``caption`` option to give a caption to the toctree.
 
      .. versionadded:: 7.3
 
-        allowed setting a directory relative to the documentation root.
+        Directories may now be set relative to the documentation source
+        directory.
 
      .. versionadded:: 3.1
 
--- a/doc/usage/extensions/autosummary.rst
+++ b/doc/usage/extensions/autosummary.rst
@@ -81,12 +81,13 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
      directory. If no argument is given, output is placed in the same directory
      as the file that contains the directive.  The directory is interpreted as
      relative to the directory of the source file, unless it begins with the
-     system path separator (``/`` for Unix-likes), in which case it is relative
-     to the documentation root.
+     path separator ``/``, in which case it is relative to the documentation
+     root.  All paths should use the separator ``/`` regardless of the
+     operating system used for the build.
 
      You can also use ``caption`` option to give a caption to the toctree.
 
-     .. versionchanged:: 7.3
+     .. versionadded:: 7.3
 
         allowed setting a directory relative to the documentation root.
 
--- a/doc/usage/extensions/autosummary.rst
+++ b/doc/usage/extensions/autosummary.rst
@@ -74,12 +74,21 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
 
      The ``toctree`` option also signals to the :program:`sphinx-autogen` script
      that stub pages should be generated for the entries listed in this
-     directive.  The option accepts a directory name as an argument;
+     directive.
+
+     The option accepts a directory name as an argument;
      :program:`sphinx-autogen` will by default place its output in this
      directory. If no argument is given, output is placed in the same directory
-     as the file that contains the directive.
+     as the file that contains the directive.  The directory is interpreted as
+     relative to the directory of the source file, unless it begins with the
+     system path separator (``/`` for Unix-likes), in which case it is relative
+     to the documentation root.
 
      You can also use ``caption`` option to give a caption to the toctree.
+
+     .. versionchanged:: 7.3
+
+        allowed setting a directory relative to the documentation root.
 
      .. versionadded:: 3.1
 
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -235,7 +235,7 @@ class Autosummary(SphinxDirective):
 
         if 'toctree' in self.options:
             tree_prefix = self.options['toctree'].strip()
-            if posixpath.isabs(tree_prefix):
+            if tree_prefix.startswith("/"):
                 tree_prefix = posixpath.relpath(tree_prefix, "/")
                 dirname = ""
             else:
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -75,7 +75,7 @@ from sphinx.locale import __
 from sphinx.project import Project
 from sphinx.pycode import ModuleAnalyzer, PycodeError
 from sphinx.registry import SphinxComponentRegistry
-from sphinx.util import logging, rst
+from sphinx.util import SEP, logging, rst
 from sphinx.util.docutils import (
     NullReporter,
     SphinxDirective,
@@ -235,8 +235,8 @@ class Autosummary(SphinxDirective):
 
         if 'toctree' in self.options:
             tree_prefix = self.options['toctree'].strip()
-            if tree_prefix.startswith("/"):
-                tree_prefix = posixpath.relpath(tree_prefix, "/")
+            if tree_prefix.startswith(SEP):
+                tree_prefix = posixpath.relpath(tree_prefix, SEP)
                 dirname = ""
             else:
                 dirname = posixpath.dirname(self.env.docname)
--- a/sphinx/ext/autosummary/__init__.py
+++ b/sphinx/ext/autosummary/__init__.py
@@ -234,9 +234,12 @@ class Autosummary(SphinxDirective):
         nodes = self.get_table(items)
 
         if 'toctree' in self.options:
-            dirname = posixpath.dirname(self.env.docname)
-
             tree_prefix = self.options['toctree'].strip()
+            if posixpath.isabs(tree_prefix):
+                tree_prefix = posixpath.relpath(tree_prefix, "/")
+                dirname = ""
+            else:
+                dirname = posixpath.dirname(self.env.docname)
             docnames = []
             excluded = Matcher(self.config.exclude_patterns)
             filename_map = self.config.autosummary_filename_map
--- a/sphinx/ext/autosummary/generate.py
+++ b/sphinx/ext/autosummary/generate.py
@@ -640,7 +640,7 @@ def find_autosummary_in_lines(lines: list[str],
             m = toctree_arg_re.match(line)
             if m:
                 toctree = m.group(1)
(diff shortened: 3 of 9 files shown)


doc/usage/extensions/autosummary.rst
```diff
@@ -74,13 +74,22 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
...
      directory. If no argument is given, output is placed in the same directory
-     as the file that contains the directive.
+     as the file that contains the directive.  The directory is interpreted as
+     relative to the directory of the source file, unless it begins with the
+     system path separator (``/`` for Unix-likes), in which case it is relative
+     to the documentation root.
 
      You can also use ``caption`` option to give a caption to the toctree.
 
+     .. versionchanged:: 7.3
```
picnixz: Since it's a feature, I think it's more a ``versionadded`` that is needed rather than a ``versionchanged``.

doc/usage/extensions/autosummary.rst
```diff
@@ -74,13 +74,22 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
...
-     directive.  The option accepts a directory name as an argument;
+     directive.
+
+     The option accepts a directory name as an argument;
      :program:`sphinx-autogen` will by default place its output in this
      directory. If no argument is given, output is placed in the same directory
-     as the file that contains the directive.
+     as the file that contains the directive.  The directory is interpreted as
+     relative to the directory of the source file, unless it begins with the
+     system path separator (``/`` for Unix-likes), in which case it is relative
```
picnixz: Does it mean that for Windows, it should be `\` ? You shouldn't say "system path separator" otherwise. Also you are using `posixpath` so I am assuming that you only want to treat `/`.

doc/usage/extensions/autosummary.rst
jakelishman: Sure, I wasn't sure how Sphinx preferred these things.  Done in 62b27c495.

doc/usage/extensions/autosummary.rst
jakelishman: Oh thanks for the catch - it's out-of-date documentation.  Originally I had written it to use `os.sep`, but then I realised that the whole directory handling in `autosummary` at the moment requires Unix-like paths and that it'd just make everything inconsistent for Windows builds to use different relative separators.

I've updated this documentation in 62b27c495.

doc/usage/extensions/autosummary.rst
picnixz: Ah no worry. Actually there isn't any strong convention. It's just that I think "versionchanged" should be reserved to "orthogonal" things changing wrt to the previous implementation without being deprecated. But anyone is free to argue !

jakelishman:
Gentle bump on this: is this PR going in a direction that the Sphinx project would be willing to merge?

picnixz:
PRs tend to be stalled for a long time (principally because we are only a small team and also because only 2-3 people have write access (I don't have it)). Also, while this feature is simple and probably useful, it may be of lower priority or have workarounds. I personally don't mind having it merged but it's up to @AA-Turner to decide when to merge it (or whether to accept it). 

On another note, I think we could actually do something with this `/` also for toctree itself or other directives, so maybe the PR will be cherry-picked to be integrated in a larger PR.

jakelishman:
That's fine by me - I know it can take a while to get to PRs, which is why I gave it a month.  With your suggestion of merging into a larger PR: that's largely what I was asking about - I'm happy to do more to merge it into something larger if that would be a preferred direction.

I don't mean to be pushy, I just know from my own maintainerships that it's easy for small PRs with no activity to fall under the radar accidentally.

picnixz:
> I don't mean to be pushy, I just know from my own maintainerships that it's easy for small PRs with no activity to fall under the radar accidentally.

If you want to be reassured, I often look at PRs and how they are updated (ultimately, I'd like to have 0 issues and PRs but that's a looong dream). Since we have only 80+ PRs, it's easy to track them IMO (and generally, if small PRs fall under the radar, that means that they were either implicitly rejected (but never closed), have conflicts, or are blocked by some technical aspects). 

There are some issues where people complain about the fact that sphinx works with an environment approach, namely `/` does not mean the absolute root of your filesystem but only the project root, especially when it comes to including files from somewhere else. Currently, I don't think there's any plan of doing this and I don't think we have something similar for the toctree (not even sure whether it's already supported or not btw), so I wouldn't get my hopes up for now. 

Currently, the dev-version is 7.3.0 and I think we'll release 8.0 next year. I think, for Sphinx 8.x, we should focus on autodoc/autosummary because this component is old, has a lot of issues because of the evolution of the typing system on Python's side, and most of the issues are autodoc ones in general; like "why is my type not rendered, why isn't there a linkable type here, why isn't my method picked, attribute, etc" so maybe we'll come back to autosummary-related issues later (and at this point your PR will be probably considered).

doc/usage/extensions/autosummary.rst
```diff
@@ -74,13 +74,23 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
...
+
+     The option accepts a directory name as an argument;
      :program:`sphinx-autogen` will by default place its output in this
      directory. If no argument is given, output is placed in the same directory
-     as the file that contains the directive.
+     as the file that contains the directive.  The directory is interpreted as
+     relative to the directory of the source file, unless it begins with the
+     path separator ``/``, in which case it is relative to the documentation
+     root.  All paths should use the separator ``/`` regardless of the
+     operating system used for the build.
```
AA-Turner: Currently inconsistent, but I'd prefer to say "source directory" rather than "root", as there are at least 4 direcories that users may configure. 
```suggestion
     as the file that contains the directive.
     The directory is interpreted as relative to the directory of the source file,
     unless it begins with ``/``, in which case it is interpreted as
     relative to the documentation source directory.
```

doc/usage/extensions/autosummary.rst
```diff
@@ -74,13 +74,23 @@ The :mod:`sphinx.ext.autosummary` extension does this in two parts:
...
 
      You can also use ``caption`` option to give a caption to the toctree.
 
+     .. versionadded:: 7.3
+
+        allowed setting a directory relative to the documentation root.
```
AA-Turner:
```suggestion
        Directories may now be set relative to the documentation source directory.
```

sphinx/ext/autosummary/__init__.py
```diff
@@ -234,9 +234,12 @@ def run(self) -> list[Node]:
         nodes = self.get_table(items)
 
         if 'toctree' in self.options:
-            dirname = posixpath.dirname(self.env.docname)
-
             tree_prefix = self.options['toctree'].strip()
+            if posixpath.isabs(tree_prefix):
```
AA-Turner: Maybe consider using SEP rather than '/'. I'd take inpiration from how the ``:doc:`` role is implemented
```suggestion
            if tree_prefix.startswith('/'):
```

sphinx/ext/autosummary/generate.py
```diff
@@ -629,7 +640,9 @@ def find_autosummary_in_lines(
             m = toctree_arg_re.match(line)
             if m:
                 toctree = m.group(1)
-                if filename:
+                if base_path is not None and posixpath.isabs(toctree):
+                    toctree = os.path.join(base_path, posixpath.relpath(toctree, "/"))
```
AA-Turner: Similarly I'd prefer to be explicit that we're looking for a prefix of ``'/'``; semantically this is slightly different from normal path manipulations

doc/usage/extensions/autosummary.rst
jakelishman: Done in e4dd14fa7

doc/usage/extensions/autosummary.rst
jakelishman: Done in e4dd14fa7.

sphinx/ext/autosummary/__init__.py
jakelishman: Done in cdca8ed26.

I didn't use `os.sep` because it would cause a little bit of inconsistency with the rest of Autosummary on Windows, which uses `posixpath` for manipulations regardless of root platform, and `os.sep` on Windows is `\`.  There's `os.altsep` that's `/` on Windows, but I didn't immediately want to go that root, because I was a bit concerned that there'd be knock-on effects from the rest of the module still using `posixpath` and thought that changing the path handling more completely might want to be a different PR if there was a reason to do it.

sphinx/ext/autosummary/generate.py
jakelishman: Done in cdca8ed26.

sphinx/ext/autosummary/__init__.py
AA-Turner: Sorry; SEP is a constant that we define -- see ``osutil``.

A

jakelishman: Ah sorry, I misunderstood.  That seems much preferable - the magic `"/"` constant didn't feel great. Changed in ~879b6b259~ 57fcb86e3 (because I forgot to run lint...).

AA-Turner: (Aside: nice to see a fellow Leodensian!)

By way of an apology, autosummary has a lot of accumulated cruft and I tend to avoid looking at it out of habit. I'll try and look at this soon, but do feel free to ping. I'm currently working on the Python 3.12 documentation ahead of release next week, so as hopefully will get to this no later than mid-October.

Thanks,
Adam

jakelishman: No worries at all, thanks - I'm in no huge rush.  If the only thing holding this up now is just your workload, I can just put it aside for a month and give it a bump if needed then.

I'm actually separately pushing to reduce our codebase's dependence on `autosummary` for similar reasons, and because (imo) it can be easy to fall into not-so-great documentation-structure habits if it's overused, which we certainly do.

(Yeah, it's not so common to see northerners around!  I grew up in York, went down south for undergrad and PhD - looks like you're similar - and now that's all done so I'm back.)

PR #11649, opened 2023-08-26
linkcheck: Use correct function to convert from UTC time to UNIX epoch

mitya57:
Subject: linkcheck: Use correct function to convert from UTC time to UNIX epoch

### Feature or Bugfix
- Bugfix

### Purpose
After subtracting the offset, we get time in UTC. But `time.mktime()` expects a time struct in local time, not in UTC. Quoting the [Python documentation](https://docs.python.org/3/library/time.html):

> Use the following functions to convert between time representations:
>
> From | To | Use
> -- | -- | --
> seconds since the epoch | struct_time in UTC | gmtime()
> seconds since the epoch | struct_time in local time | localtime()
> struct_time in UTC | seconds since the epoch | calendar.timegm()
> struct_time in local time | seconds since the epoch | mktime()

So in our case, the correct function to use is `calendar.timegm()`.

This fixes hanging tests when the local timezone is to the west of UTC. One could test this using:

    TZ=America/New_York python3 -m pytest -k test_too_many_requests_retry_after_HTTP_date

After applying this change, the test passes with timezones from both western and eastern hemispheres.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,7 +6,8 @@ Bugs fixed
 
 * #11618: Fix a regression in the MoveModuleTargets transform,
   introduced in #10478 (#9662).
-* #11649: Use :py:func:`calendar.timegm` to convert from UTC time to UNIX epoch.
+* #11649: linkcheck: Fix conversions from UTC to UNIX time
+  for timezones west of London.
 
 Release 7.2.3 (released Aug 23, 2023)
 =====================================
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@ Bugs fixed
 
 * #11618: Fix a regression in the MoveModuleTargets transform,
   introduced in #10478 (#9662).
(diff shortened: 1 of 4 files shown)


AA-Turner:
Should we add a test for this? Please could you also add an entry to CHANGES, this will go into 7.2.3.

A

mitya57:
Done.

CHANGES
```diff
@@ -6,6 +6,7 @@ Bugs fixed
 
 * #11618: Fix a regression in the MoveModuleTargets transform,
   introduced in #10478 (#9662).
+* #11649: Use :py:func:`calendar.timegm` to convert from UTC time to UNIX epoch.
```
```suggestion
* #11649: linkcheck: Fix conversions from UTC to UNIX time
  for timezones west of London.
```

mitya57: The conversion was wrong for timezones east of London, too. Just the test didn't hang there.

PR #11650, opened 2023-08-26
Make test_include_source_read_event pass when run from /tmp

mitya57:
Subject: Make test_include_source_read_event pass when run from /tmp

### Feature or Bugfix
- Bugfix

### Purpose
Because pytest's base tmp_path is also in /tmp, `source` here is relative path, and thus is not properly converted to doc name.

So the test fails because `sources_reported` dict has a key like `../../../pytest-of-debci/pytest-0/directive-include/baz/baz`, and `assert "baz/baz" in sources_reported` fails.

Example failures in Debian infrastructure can be seen here:
- https://salsa.debian.org/python-team/packages/sphinx/-/jobs/4579414
- https://salsa.debian.org/python-team/packages/sphinx/-/jobs/4612253

You can reproduce locally by copying Sphinx source tree to `/tmp` and running tests from there.

### DIFF
--- a/sphinx/directives/other.py
+++ b/sphinx/directives/other.py
@@ -18,7 +18,7 @@ from sphinx.util import docname_join, logging, url_re
 from sphinx.util.docutils import SphinxDirective
 from sphinx.util.matching import Matcher, patfilter
 from sphinx.util.nodes import explicit_title_re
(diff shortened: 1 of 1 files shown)


AA-Turner:
Thanks!

PR #11655, opened 2023-08-29
fix typos in sphinx.util._DEPRECATED_OBJECTS

kartben:
Some typos are causing erroneous suggestions when using deprecated APIs.


kartben:
That was fast -- thanks!

Issue #11643, opened 2023-08-25
source-read for included files is called with a file path, not a docname


lmoureaux:
### Describe the bug

Since #10678, the `source-read` event is emitted for included files. The first argument passed to the function in this case is a file path, not a docname. This breaks hooks trying to access the document, except when building from the source folder. Extensions such as [`sphinx-last-updated-by-git`](https://github.com/mgeier/sphinx-last-updated-by-git/tree/master) (also currently broken for a different reason) would need to work around the present behavior.

### How to Reproduce

Put the following in a folder called `docs`:

docs/conf.py:
```py
import sys
import os

sys.path.insert(0, os.path.abspath('.'))

extensions = [
        'my-extension'
]
```
docs/index.rst:
```rst
This is a test
==============

.. include:: something-to-include.rst
```
docs/something-to-include.rst:
```rst
Testing
=======
```
docs/my-extension.py:
```py
#!/usr/bin/env python3

import os

from sphinx.application import Sphinx


__version__ = '1.0'


def check_doc_file(app: Sphinx, docname, source):
    path = app.env.doc2path(docname, True)
    print(docname, path)
    assert os.path.isfile(path)


def setup(app: Sphinx):

    app.connect('source-read', check_doc_file)

    return dict(
        version=__version__,
        parallel_read_safe=True,
        parallel_write_safe=True
    )
```

Then execute:
```sh
sphinx-build docs build # fails
cd docs; sphinx-build . ../build # works
```

### Environment Information

```text
Platform:              linux; (Linux-6.2.0-26-generic-x86_64-with-glibc2.35)
Python version:        3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0])
Python implementation: CPython
Sphinx version:        7.2.3
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Sphinx extensions

```python
See minimal example above.
```

### Additional context

Debugging https://github.com/longturn/freeciv21/issues/2023

AA-Turner:
I think this has been fixed by #11650 (the example in the body of the issue no longer fails)

A

mgeier:
This is still a bit fishy.

If you use a path outside the source directory, e.g.

```rst
.. include:: ../../some/path/included-file.rst
```

... AND if the first element of `source_suffix` doesn't match the suffix, e.g.

```python
source_suffix = ['.nonsense', '.rst']
```

[...]

Issue #11620, opened 2023-08-18
``BuildEnvironment.docname`` can now be `None`


AA-Turner:
> It seems that `BuildEnvironment.docname` can now be `None`:

https://github.com/sphinx-doc/sphinx/blob/ed84d63e6f2c4fd43b97fc43ee8be4156a13af9e/sphinx/environment/__init__.py#L545

> or at least `env.docname` in `read-source` event can now be `None`:

https://github.com/sphinx-doc/sphinx/blob/da9f8a5c33ad5eeef05dd780f2988f3ff7351ef3/sphinx/io.py#L114

Ref: https://github.com/sphinx-gallery/sphinx-gallery/issues/1175

_Originally posted by @lucyleeow in https://github.com/sphinx-doc/sphinx/issues/11605#issuecomment-1683232145_

AA-Turner:
@lucyleeow thanks for this report, do you have a reproducer for this at all please?

Thanks,
Adam

AA-Turner:
May be related to #10678 & #11510

AA-Turner:
Not a bug in Sphinx, will report upstream (https://github.com/sphinx-gallery/sphinx-gallery/issues/1175)

lucyleeow:
Thank you @AA-Turner for looking into this!

Would it be appropriate to update the docs so `docname` return type is `str` or `None`.? Though I do note that `path2doc` has always had return type str or `None`, so this is not a new change.

mgeier:
I seem to have the same problem in my `sphinx-last-updated-by-git` extension. I can work around it with https://github.com/mgeier/sphinx-last-updated-by-git/pull/59, but I'm not sure if that's the right thing to do.

I think at least the `source-read` docs should be updated to explain in which situations `docname` can be `None`.

mgeier:
I have added a reproducer: #11648

We can continue discussion there.

AA-Turner:
Closing as fixed by #11657.

PR #11657, opened 2023-08-30
Add an ``'include-read'`` event

### LINKED ISSUE #11648: Documentation of source-read event: "docname" can be None
### Describe the bug

When including a file with the `include` directive, the `docname` argument can be `None` (since Sphinx 7.2).

The docs don't mention that: https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-source-read

### How to Reproduce

`conf.py`:

```python
def source_read(app, docname, source):
    print('#####')
    print('docname:', docname)


def setup(app):
    app.connect('source-read', source_read)
```

`index.rst`:

```rst
.. include:: another-file.txt
```

`another-file.txt`:

```
Some text.
```

Run this:

    python -m sphinx . _build

... and check the terminal output.

### Environment Information

```text
Python version:        3.11.4 (main, Jun  7 2023, 10:13:09) [GCC 12.2.0])
Sphinx version:        7.2.0
Docutils version:      0.18.1
Jinja2 version:        3.1.2
Pygments version:      2.15.1
```

### Sphinx extensions

none, but this behavior is relevant for extensions that want to use the `source-read` event, like in my case https://github.com/mgeier/sphinx-last-updated-by-git

### Additional context

Maybe related: #11643

Possible work-around: https://github.com/mgeier/sphinx-last-updated-by-git/pull/59

AA-Turner:
closes #11648, #11643, #11620

cc: @mgeier @lmoureaux

The easy option would be to set the docname to ``self.env.docname`` (the current document being read), but this might throw off extensions that expect to only have ``"source-read"`` called exactly once per document.

Please test / let me know thoughts!

A


AA-Turner:
I'm leaning towards the generated docname approach -- would appreciate your opinion @mgeier, as this solves the 'docname can be None' issue, but introduces docnames that don't work with ``doc2path()``.

A

mgeier:
I think inventing a string (like `'include-from-mysource'`) is much worse than returning `None`.

If a string is returned, I would expect that string to be the document name.

I'm fine with returning `None`, but the documentation should be updated, as I mentioned in #11648.

AA-Turner:
A good point -- I've also re-visited the original issue (https://github.com/sphinx-doc/sphinx/issues/10678) and have thusly changed my mind -- this PR now introduces a new ``'include-read'`` event, which no longer abuses the ``'source-read'`` event for non-documents.

I assume this would work for your purposes?

A

mgeier:
That sounds much cleaner!

> I assume this would work for your purposes?

Yes, sure, an additional event will not break anything if I don't use it, right?

My extension `sphinx-last-updated-by-git` doesn't need an event for the `include` directive, it only needs to check the `env.dependencies` of every source file, which should contain the included files.

mgeier:
I don't know if anyone needs this, but you could pass the parent document name to the event handler?

AA-Turner:
Good shout!

lucyleeow:
Thanks @AA-Turner, this is a good solution!

PR #11665, opened 2023-09-01
Fix ordering inside searchindex.js not being deterministic

### LINKED ISSUE #11622: sphinx.search._JavaScriptIndex: non-determinite searchindex.js output
### Describe the bug

While looking at some recent [Reproducible Build](https://reproducible-builds.org/) test failures, I noticed that one of the remaining [differences](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/diffoscope-results/quodlibet.html) between builds of [`quodlibet`](https://quodlibet.readthedocs.io/) is due to sorting differences in `searchindex.js`, a file produced to enable client-side content search in web browsers.

Looking at the [relevant code in the `sphinx.search._JavaScriptIndex` class](https://github.com/sphinx-doc/sphinx/blob/c1f2a52d5d97f874d84e4d5421c99a566242af6b/sphinx/search/__init__.py#L150-L173), the code to output the `searchindex.js` content is a fairly straightforward [Python built-in `json.dumps`](https://docs.python.org/3/library/json.html#json.dumps) call.

The failures mentioned occurred with Sphinx v5.3.0 -- I believe they would occur when using the latest release v7.2.2 also (the [index export code appears unchanged](https://github.com/sphinx-doc/sphinx/blob/ed84d63e6f2c4fd43b97fc43ee8be4156a13af9e/sphinx/search/__init__.py#L155-L178)).

I think that a fix may be as straightforward as sending a `sort_keys=True` argument to the `json.dumps` call.  Writing test coverage to provide varying inputs that sort unequally otherwise could be the tricky part.

cc @lamby 

### How to Reproduce

I do not yet have a local reproducer for this bug, but the differences in output can be observed in the diffoscope results for `quodlibet` on Sphinx v5.3.0 here: https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/diffoscope-results/quodlibet.html

### Environment Information
Since the reproducer case I'm reporting from was a hosted build, I can't easily run the environment-reporting commands, but some relevant details are:

```text
Sphinx: v5.3.0
Operating System: Debian Linux (unstable / trixie)
Documentation-package-under-test: quodlibet v4.5.0
```

### Sphinx extensions

```python
extensions = ['sphinx.ext.autodoc', 'sphinx.ext.extlinks', 'contributors']
```

### Additional context

_No response_

emilyalbini:
This PR changes `searchindex.js`'s sorting of dictionary keys to be deterministic.

### Feature or Bugfix

- Bugfix

### Purpose

At my day job we have a need to digitally sign the HTML output of Sphinx, and for those signatures to stay valid as we rebuild the documentation (if all dependencies and source code do not change).

Unfortunately, the contents of `searchindex.js` are not deterministic, as the keys are inserted in random order. This breaks our digital signatures as soon as a rebuild is done.

### Detail

This PR changes the sorting of dumped dictionary keys to be deterministic, and adds a test verifying all dictionary keys in `searchindex.js` are sorted.

### Relates

- Fixes #11622

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -304,3 +304,23 @@ def test_parallel(app):
     app.build()
     index = load_searchindex(app.outdir / 'searchindex.js')
     assert index['docnames'] == ['index', 'nosearch', 'tocitem']
+
+
+@pytest.mark.sphinx(testroot='search')
+def test_search_index_is_deterministic(app):
+    app.builder.build_all()
+    index = load_searchindex(app.outdir / 'searchindex.js')
+
+    def check_if_dicts_are_sorted(item):
+        if isinstance(item, dict):
+            assert list(item.keys()) == sorted(item.keys())
+            for child in item.values():
+                check_if_dicts_are_sorted(child)
+        # Lists in the search index cannot be sorted: for some lists, their
+        # position inside the list is referenced elsewhere in the index, so if
+        # we were to sort lists, the search index would break.
+        elif isinstance(item, list):
+            for child in item:
+                check_if_dicts_are_sorted(child)
+
+    check_if_dicts_are_sorted(index)
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -308,19 +308,35 @@ def test_parallel(app):
 
 @pytest.mark.sphinx(testroot='search')
 def test_search_index_is_deterministic(app):
+    LISTS_NOT_TO_SORT = [
+        # Each element of .titles is related to the element of .docnames in the same position.
+        # The ordering is deterministic because .docnames is sorted.
+        '.titles',
+        # Each element of .filenames is related to the element of .docnames in the same position.
+        # The ordering is deterministic because .docnames is sorted.
+        '.filenames',
+    ]
+
+    # In the search index, titles inside .alltitles are stored as a tuple of
+    # (document_idx, title_anchor). Tuples are represented as lists in JSON,
+    # but their contents must not be sorted. We cannot sort them anyway, as
+    # document_idx is an int and title_anchor is a str.
+    def is_title_tuple_type(item):
+        return len(item) == 2 and isinstance(item[0], int) and isinstance(item[1], str)
+
+    def assert_is_sorted(item, path):
+        err_path = path if path else '<root>'
+        if isinstance(item, dict):
+            assert list(item.keys()) == sorted(item.keys()), f'{err_path} is not sorted'
+            for key, value in item.items():
+                assert_is_sorted(value, f'{path}.{key}')
+        elif isinstance(item, list):
+            if not is_title_tuple_type(item) and path not in LISTS_NOT_TO_SORT:
+                assert item == sorted(item), f'{err_path} is not sorted'
+            for i, child in enumerate(item):
+                assert_is_sorted(child, f'{path}[{i}]')
+
     app.builder.build_all()
     index = load_searchindex(app.outdir / 'searchindex.js')
-
-    def check_if_dicts_are_sorted(item):
-        if isinstance(item, dict):
-            assert list(item.keys()) == sorted(item.keys())
-            for child in item.values():
-                check_if_dicts_are_sorted(child)
-        # Lists in the search index cannot be sorted: for some lists, their
-        # position inside the list is referenced elsewhere in the index, so if
-        # we were to sort lists, the search index would break.
-        elif isinstance(item, list):
-            for child in item:
-                check_if_dicts_are_sorted(child)
-
-    check_if_dicts_are_sorted(index)
+    print(f'index contents:\n{json.dumps(index, indent=2)}')  # Pretty print the index.
+    assert_is_sorted(index, '')
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -308,14 +308,14 @@ def test_parallel(app):
 
 @pytest.mark.sphinx(testroot='search')
 def test_search_index_is_deterministic(app):
-    LISTS_NOT_TO_SORT = [
+    lists_not_to_sort = {
         # Each element of .titles is related to the element of .docnames in the same position.
         # The ordering is deterministic because .docnames is sorted.
         '.titles',
         # Each element of .filenames is related to the element of .docnames in the same position.
         # The ordering is deterministic because .docnames is sorted.
         '.filenames',
-    ]
+    }
 
     # In the search index, titles inside .alltitles are stored as a tuple of
     # (document_idx, title_anchor). Tuples are represented as lists in JSON,
@@ -331,12 +331,13 @@ def test_search_index_is_deterministic(app):
             for key, value in item.items():
                 assert_is_sorted(value, f'{path}.{key}')
         elif isinstance(item, list):
-            if not is_title_tuple_type(item) and path not in LISTS_NOT_TO_SORT:
+            if not is_title_tuple_type(item) and path not in lists_not_to_sort:
                 assert item == sorted(item), f'{err_path} is not sorted'
             for i, child in enumerate(item):
                 assert_is_sorted(child, f'{path}[{i}]')
 
     app.builder.build_all()
     index = load_searchindex(app.outdir / 'searchindex.js')
-    print(f'index contents:\n{json.dumps(index, indent=2)}')  # Pretty print the index.
+    # Pretty print the index. Only shown by pytest on failure.
+    print(f'searchindex.js contents:\n\n{json.dumps(index, indent=2)}')
     assert_is_sorted(index, '')
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@ Bugs fixed
 
 * #11668: Raise a useful error when ``theme.conf`` is missing.
   Patch by Vinay Sajip.
+* #11622: Fix ordering inside searchindex.js not being deterministic.
 
 Testing
 -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,8 @@ Bugs fixed
 
 * #11668: Raise a useful error when ``theme.conf`` is missing.
   Patch by Vinay Sajip.
-* #11622: Fix ordering inside searchindex.js not being deterministic.
+* #11622: Ensure that the order of keys in ``searchindex.js`` is deterministic.
+  Patch by Pietro Albini.
 
 Testing
 -------
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -162,7 +162,7 @@ class _JavaScriptIndex:
     SUFFIX = ')'
 
     def dumps(self, data: Any) -> str:
-        return self.PREFIX + json.dumps(data) + self.SUFFIX
+        return self.PREFIX + json.dumps(data, sort_keys=True) + self.SUFFIX
 
     def loads(self, s: str) -> Any:
         data = s[len(self.PREFIX):-len(self.SUFFIX)]


pietroalbini:
Is the Windows failure a spurious one? I don't think I changed anything that would influence translations.

AA-Turner:
Yes, there's an issue about it somewhere (@jayaddison had a go at fixing it to no avail)

A

tests/test_search.py
```diff
@@ -304,3 +304,27 @@ def test_parallel(app):
...
+@pytest.mark.sphinx(testroot='search')
+def test_search_index_is_deterministic(app):
+    app.builder.build_all()
+    index = load_searchindex(app.outdir / 'searchindex.js')
+
+    def check_if_dicts_are_sorted(item):
+        if type(item) is dict:
+            keys = list(item.keys())
+            if not all(keys[i] <= keys[i + 1] for i in range(len(keys) - 1)):
+                msg = f'keys are not sorted: {item!r}'
```
picnixz: I'd prefer having an assertion to check rather a RuntimeError (pytest should catch it I think?). And have the assert message directly (no msg variable)

tests/test_search.py
```diff
@@ -304,3 +304,27 @@ def test_parallel(app):
...
+
+@pytest.mark.sphinx(testroot='search')
+def test_search_index_is_deterministic(app):
+    app.builder.build_all()
+    index = load_searchindex(app.outdir / 'searchindex.js')
+
+    def check_if_dicts_are_sorted(item):
+        if type(item) is dict:
+            keys = list(item.keys())
+            if not all(keys[i] <= keys[i + 1] for i in range(len(keys) - 1)):
```
picnixz: Isn't it easier to check that `list(...) == list(sorted(...))` ?

tests/test_search.py
```diff
@@ -304,3 +304,27 @@ def test_parallel(app):
...
     assert index['docnames'] == ['index', 'nosearch', 'tocitem']
+
+
+@pytest.mark.sphinx(testroot='search')
+def test_search_index_is_deterministic(app):
+    app.builder.build_all()
+    index = load_searchindex(app.outdir / 'searchindex.js')
+
+    def check_if_dicts_are_sorted(item):
+        if type(item) is dict:
```
picnixz: Is there some reason why it wouldn't work for subclasses?

tests/test_search.py
```diff
@@ -304,3 +304,27 @@ def test_parallel(app):
...
+            if not all(keys[i] <= keys[i + 1] for i in range(len(keys) - 1)):
+                msg = f'keys are not sorted: {item!r}'
+                raise RuntimeError(msg)
+
+            for child in item.values():
+                check_if_dicts_are_sorted(child)
+        # Lists in the search index cannot be sorted: for some lists, their
+        # position inside the list is referenced elsewhere in the index, so if
+        # we were to sort lists, the search index would break.
+        elif type(item) is list:
```
picnixz: Same question as for the dict.

tests/test_search.py
emilyalbini: The msg variable was [a complaint ruff made](https://github.com/sphinx-doc/sphinx/actions/runs/6047168044/job/16410100144), but I'll change the code soon to use an assertion.

tests/test_search.py
```diff
@@ -304,3 +304,23 @@ def test_parallel(app):
...
+    index = load_searchindex(app.outdir / 'searchindex.js')
+
+    def check_if_dicts_are_sorted(item):
+        if isinstance(item, dict):
+            assert list(item.keys()) == sorted(item.keys())
+            for child in item.values():
+                check_if_dicts_are_sorted(child)
+        # Lists in the search index cannot be sorted: for some lists, their
+        # position inside the list is referenced elsewhere in the index, so if
+        # we were to sort lists, the search index would break.
```
AA-Turner: This seems like a problem we should solve?

emilyalbini: I don't see how it's possible to sort all lists without changing the format of the search index itself, which feels outside the scope of this PR. Those orderings are deterministic, just not sortable.

There are two cases today where items are not sorted inside of lists:

* In the `.titles` and `.filenames` lists, each item corresponds to the item at the same position in the (sorted) `.docnames` list. Sorting `.titles` and `.filenames` would break that relationship.

* In the `.alltitles.*` lists, the lists are actually an `(int, str)` tuple, which is represented as a list in JSON. Sorting them wouldn't make much sense, and won't work anyway due to the different types.

Still, I updated the test to check that everything but those two exceptions are sorted.

jayaddison: > Those orderings are deterministic

As long as that's true, this seems reasonable to me.  For a moment I was worried that if `.titles` / `.filenames` were _not_ deterministic, then we'd be solving 95% of cases while leaving a more difficult-to-resolve case deeper in the datastructure.

Could I check why/how confident we are that those fields are deterministically-generated, to get a sense for how robust or fragile that is?

jayaddison:
> @.pietroalbini Is the Windows failure a spurious one? I don't think I changed anything that would influence translations.

> @.AA-Turner Yes, there's an issue about it somewhere (@jayaddison had a go at fixing it to no avail)

Ah, yep - that's #11232.  I think we've gotten fairly close to the cause - my latest guess was [these lines here](https://github.com/sphinx-doc/sphinx/blob/fcc38997f1d9b728bb4ffc64fc362c7763a4ee25/tests/test_intl.py#L54-L56).  It's something to do with whether an internationalized byte-order-mark file is rewritten.

tests/test_search.py
AA-Turner: > Those orderings are deterministic

Ahh, this wasn't clear in the comment -- please would you be able to update it just to clarify?

A

emilyalbini: > Ahh, this wasn't clear in the comment -- please would you be able to update it just to clarify?

Updated the comments to clarify why they are deterministic!

> Could I check why/how confident we are that those fields are deterministically-generated, to get a sense for how robust or fragile that is?

The code seems pretty robust:

https://github.com/sphinx-doc/sphinx/blob/fcc38997f1d9b728bb4ffc64fc362c7763a4ee25/sphinx/search/__init__.py#L385-L386

Also, there is another test that verifies the search index content is the expected one, which (implicitly) tests the relationship between docnames and titles/filenames:

https://github.com/sphinx-doc/sphinx/blob/fcc38997f1d9b728bb4ffc64fc362c7763a4ee25/tests/test_search.py#L179-L197

tests/test_search.py
```diff
@@ -304,3 +304,39 @@ def test_parallel(app):
...
+
+    # In the search index, titles inside .alltitles are stored as a tuple of
+    # (document_idx, title_anchor). Tuples are represented as lists in JSON,
+    # but their contents must not be sorted. We cannot sort them anyway, as
+    # document_idx is an int and title_anchor is a str.
+    def is_title_tuple_type(item):
+        return len(item) == 2 and isinstance(item[0], int) and isinstance(item[1], str)
+
+    def assert_is_sorted(item, path):
+        print('visiting', path if path else '<root>')
```
picnixz: Are those prints really needed? You could include the path in assert messages instead (unless the linter complains about having messages inside the assert, but in that case I still think that we shouldn't print anything).

tests/test_search.py
```diff
@@ -304,3 +304,39 @@ def test_parallel(app):
...
+                assert_is_sorted(value, f'{path}.{key}')
+        elif isinstance(item, list):
+            if not is_title_tuple_type(item) and path not in LISTS_NOT_TO_SORT:
+                assert item == sorted(item), f'{err_path} is not sorted'
+            for i, child in enumerate(item):
+                assert_is_sorted(child, f'{path}[{i}]')
+
+    app.builder.build_all()
+    index = load_searchindex(app.outdir / 'searchindex.js')
+    print('"index contents:\n{json.dumps(index, indent=2)}')  # Pretty print the index.
```
picnixz: I still don't think it's a good idea to pprint. If the test fails in the future, then it is our job to know why it failed and only then would we likely debug the index. But otherwise, I prefer not having print messages in tests that are only meant for debugging.

emilyalbini: I can remove it, but I think this print doesn't add noise to the code or the test passing output, while providing helpful context when debugging a failure (for example, if it fails in CI you can quickly glance what's wrong, rather than having to push a new commit adding prints or trying to reproduce locally). Still, it's your project, so if you'd rather not have the print I'll remove it :slightly_smiling_face:

picnixz: Well first of all it's not my project xD In general, we expect contributors to test locally by running tox or anything else (at least I think it's faster to do it locally first because failures are likely to occur both locally and on CI/CD if any). 

Still, now that you mention it, I don't remember exactly whether the output is shown during the traceback or it its captured somewhere else (I know that pytest captures stdout and stdin when using capsys fixtures but I don't remember if, when testing locally my stuff, I could end up seeing those "print"). If nothing is shown locally, then I think it's fine but in general I tend to avoid leaving print messages whether it's in the src or the test code.

Nevertheless, I think that most of the tests do not print things (though I know that there are some that print messages but I think we didn't create new print messages for a while, meaning the former are some legacy from a long time ago). 

Anyway, let's leave it like that unless there is a real local visual pollution when running the tests (we could always fix later though).

emilyalbini: > Still, now that you mention it, I don't remember exactly whether the output is shown during the traceback or it its captured somewhere else (I know that pytest captures stdout and stdin when using capsys fixtures but I don't remember if, when testing locally my stuff, I could end up seeing those "print"). If nothing is shown locally, then I think it's fine but in general I tend to avoid leaving print messages whether it's in the src or the test code.

Nothing is printed when the output is successful, it's only shown during failure. [I captured both the success and failure output in a gist](https://gist.github.com/pietroalbini/2d9f83925db243f4014d2ad36c3fef99) :slightly_smiling_face:

picnixz (approved):

tests/test_search.py
```diff
@@ -304,3 +304,39 @@ def test_parallel(app):
     app.build()
     index = load_searchindex(app.outdir / 'searchindex.js')
     assert index['docnames'] == ['index', 'nosearch', 'tocitem']
+
+
+@pytest.mark.sphinx(testroot='search')
+def test_search_index_is_deterministic(app):
+    LISTS_NOT_TO_SORT = [
```
AA-Turner:
```suggestion
    lists_not_to_sort = {
```

tests/test_search.py
```diff
@@ -304,3 +304,39 @@ def test_parallel(app):
...
+        # Each element of .titles is related to the element of .docnames in the same position.
+        # The ordering is deterministic because .docnames is sorted.
+        '.titles',
+        # Each element of .filenames is related to the element of .docnames in the same position.
+        # The ordering is deterministic because .docnames is sorted.
+        '.filenames',
+    ]
```
AA-Turner:
```suggestion
    }
```

tests/test_search.py
```diff
@@ -304,3 +304,39 @@ def test_parallel(app):
...
+        return len(item) == 2 and isinstance(item[0], int) and isinstance(item[1], str)
+
+    def assert_is_sorted(item, path):
+        err_path = path if path else '<root>'
+        if isinstance(item, dict):
+            assert list(item.keys()) == sorted(item.keys()), f'{err_path} is not sorted'
+            for key, value in item.items():
+                assert_is_sorted(value, f'{path}.{key}')
+        elif isinstance(item, list):
+            if not is_title_tuple_type(item) and path not in LISTS_NOT_TO_SORT:
```
AA-Turner:
```suggestion
            if not is_title_tuple_type(item) and path not in lists_not_to_sort:
```

tests/test_search.py
```diff
@@ -304,3 +304,39 @@ def test_parallel(app):
...
+                assert item == sorted(item), f'{err_path} is not sorted'
+            for i, child in enumerate(item):
+                assert_is_sorted(child, f'{path}[{i}]')
+
+    app.builder.build_all()
+    index = load_searchindex(app.outdir / 'searchindex.js')
+    print(f'index contents:\n{json.dumps(index, indent=2)}')  # Pretty print the index.
```
AA-Turner:
```suggestion
    # Pretty print the index. Only shown by pytest on failure.
    print(f'searchindex.js contents:\n\n{json.dumps(index, indent=2)}')
```

CHANGES
```diff
@@ -1,3 +1,23 @@
...
+Deprecated
+----------
+
+Features added
+--------------
+
+Bugs fixed
+----------
+
+* #11622: Fix ordering inside searchindex.js not being deterministic.
```
```suggestion
* #11622: Ensure that the order of keys in ``searchindex.js` is deterministic.
  Patch by Pietro Albini.
```

AA-Turner:
@pietroalbini it looks like I can't push to this branch / apply review comments -- please could you check if the tick-box is ticked for allowing me to do this?

Thanks,
Adam

pietroalbini:
Applied all suggestions and rebased on top of master.

> it looks like I can't push to this branch / apply review comments -- please could you check if the tick-box is ticked for allowing me to do this?

Unfortunately the organization I created the fork on doesn't support that :slightly_frowning_face: (it's an enterprise org with SSO).

AA-Turner:
Thanks @pietroalbini!

A

PR #11671, opened 2023-09-02
Fix #11668: Add check for non-existent theme.conf and fail with a mor…

### LINKED ISSUE #11668: Unhelpful error message when theme.conf is missing
### Describe the bug

If the `theme.conf` file is missing, the code here:
https://github.com/sphinx-doc/sphinx/blob/fcc38997f1d9b728bb4ffc64fc362c7763a4ee25/sphinx/theming.py#L72
creates an empty configuration and fails a few lines later at
https://github.com/sphinx-doc/sphinx/blob/fcc38997f1d9b728bb4ffc64fc362c7763a4ee25/sphinx/theming.py#L77
because of course there's no `[theme]` section, as the file doesn't exist!

It would be more helpful to raise an error saying that the `theme.conf` file is missing.

### How to Reproduce

Just remove `theme.conf ` from a theme and try to build some documentation with that theme.

### Environment Information

```text
Platform:              linux; (Linux-4.15.0-213-generic-x86_64-with-glibc2.27)
Python version:        3.10.8 (main, Nov 24 2022, 07:33:30) [GCC 7.5.0])
Sphinx version:        7.2.5
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Sphinx extensions

```python
None
```

### Additional context

_No response_

vsajip:
…e informative error message.

Subject: Fix #11668: Add check for non-existent theme.conf and fail with a more informative error message.

### Feature or Bugfix

- Enhancement

### Purpose
- To fix #11668.
- No particular environment dependencies.

### Detail
- Change to `Theme.__init__()` to add check.
- Added test `test_no_theme_conf()`.

### Relates
- #11668


tests/test_theming.py
```diff
@@ -62,6 +62,19 @@ def test_theme_api(app, status, warning):
     assert not os.path.exists(themedir)
 
 
+def test_no_theme_conf():
```
picnixz: Instead of creating temporary files like this, there is a fixture doing it for you (don't remember its name now but it's something like tmp_path or tmp_something, just look at some other tests or in conftest or in sphinx/testing.py).

Also, testroots should not be modified like that because they might affect other tests. 

I would suggest doing the following instead:

- use a dummy application to create a theme factory object (otherwise mypy may complain)
- create a temporary empty directory
- call `Theme(name, tmpdir, factory)`.
- catch the exception.

That way, you have no side effects and you don't need to import THEMECONF.

Btw I see that the ThemeFactory checks that THEMECONF exists when loading the external themes (lines 215 and 222 of theming.py) so I am actually wondering: when are you getting your error? I don't much about themes but does the application actually calls the Theme constructor or is it via a factory that it actually creates themes? if the Theme class is actually part of the private API, we might have assumed that the THEMECONF file actually exists when calling the constructor.

vsajip: The error occurred when I changed the `MANIFEST.in` in my theme and as a result inadvertently excluded `theme.conf` from the built artifacts (tarball, wheel). Then a documentation build using the theme failed on `sphinx-themes.org` because of the missing file. When I issued a new release after fixing my `MANIFEST.in`, the build now works. The [stack trace](https://github.com/vsajip/sphinx_sizzle_theme/issues/19) showed me where the problem was in `Theme.__init__().`

picnixz: Ah yes, I see. Then you can directly use the Theme class in the tests (but still, change the way that it is tested)

vsajip: OK, done.

vsajip: There are some test errors in the GitHub Actions but they don't appear to be related to my changes.

picnixz: Those are flaky tests, don't worry. ~~Btw, can you assert the exception message as well?~~ Forget the exception since it's a localized error.

picnixz (approved):

tests/test_theming.py
vsajip: Actually, I didn't assert the exception message because I wasn't sure of the i18n/l10n approach to be used - by which I mean, the exception message text is potentially dependent on the environment's locale.

AA-Turner:
Thanks @vsajip!

A

Issue #11003, opened 2022-11-30
Add warning type and subtype for serial read warnings


jdillard:
**Is your feature request related to a problem? Please describe.**

When running Sphinx with the [-W](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-W) flag enabled in a complex build environment (where it adds even more complexity to calculate whether a build should run with parallel flags or not), it would be nice to be able to utilize the [suppress_warnings](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings) confval when installing an extension that is knowingly not marked `parallel_safe` for read/write. Although, currently serial warnings lack a type/subtype, so that isn't possible.

If the concept has a chance of getting approved, I can (attempt to) make the PR.

**Describe the solution you'd like**

Given that `suppress_warnings` basically lets the user choose the comprise that works best for them, it seems it wouldn't hurt to allow a type/subtype for the [serial warnings](https://github.com/sphinx-doc/sphinx/blob/cd3f2e435000835dd98a11497440bc16a79ec31c/sphinx/application.py#L1279-L1284). I don't know what the most future proof types/substypes are, but I propose the following:

1. `serial.read` and `serial.write` (`type='serial', subtype=typ`) for [L1280](https://github.com/sphinx-doc/sphinx/blob/cd3f2e435000835dd98a11497440bc16a79ec31c/sphinx/application.py#L1280) and [L1284](https://github.com/sphinx-doc/sphinx/blob/cd3f2e435000835dd98a11497440bc16a79ec31c/sphinx/application.py#L1284). I could see the argument for using `parallel` instead of `serial` for the type here.

2. `ext.{extension_name}` (`type='ext', subtype=ext.name`) for [L1279](https://github.com/sphinx-doc/sphinx/blob/cd3f2e435000835dd98a11497440bc16a79ec31c/sphinx/application.py#L1279) and [L1283](https://github.com/sphinx-doc/sphinx/blob/cd3f2e435000835dd98a11497440bc16a79ec31c/sphinx/application.py#L1283). If there are other places that throw warnings using the extension's name, this convention could be used there as well. Being able to supress warnings for an extension that are thrown outside the extension's code seems useful.

Also some notes on the documentation for [suppress_warnings](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings):

[...]

PR #9732, opened 2021-10-12
Allow users to suppress autodoc.mocked_object warnings

CyrilRoelandteNovance:

PR #11677, opened 2023-09-05
Add more warning types and sub-types

Tiger3018:

### Bugfix and Refactoring

- Bugfix： bdea05e8 7f5ab050

  > Fix warning type `toc`: in directive, env and util
  > Fix warning type `i18n` and `index`: 3 new subtype
  > Make ``toc``, ``i18n`` and ``index`` warning types cover all relative warnings.

- Refactoring(proposal): #11003 a07f4391

### Purpose

Please see commit message and its TODO.

### Detail

(Refactoring) From commit message

- [x] Remove: Since the `application.py` has the warning of write or read ahead of "doing serial %s", it is safe to remove the second warning.

- [x] Documentation: Replace "should be considered experimental" with proposal typename `ext_*`

- [ ] `util/logging.py`: Change the regex match of warning type.
- [ ] Test: 
- [ ] `application.py`: 

### Relates

- blocks #9732


doc/usage/configuration.rst
```diff
@@ -354,14 +357,19 @@ General configuration
...
+   * ``toc.no_title``
+   * ``toc.not_included``
    * ``toc.not_readable``
    * ``toc.secnum``
 
    You can choose from these types.  You can also give only the first
    component to exclude all warnings attached to it.
 
-   Now, this option should be considered *experimental*.
+   It's recommended to use ``ext_*`` type for community extensions.
```
picnixz: I would say that it's perhaps even better to actually use `ext/NAME` for instance. I am not a fan of something like `ext_my.ext` and would prefer seeing something like `ext/my.ext` or `contrib/my.ext`. 

What do other people think?

doc/usage/configuration.rst
```diff
@@ -393,9 +401,19 @@ General configuration
...
+   .. versionadded:: 7.1
+
+      Added ``index`` warning type.
+
+   .. versionchanged:: 7.2
+
+      Make ``toc``, ``i18n`` and ``index`` warning types cover all relative
+      warnings.
+
+      Added ``i18n.babel``, ``i18n.not_readable``, ``i18n.not_writeable``,
```
picnixz: What would those warnings mean?

Actually we should maybe have a follow-up issue which documents where and when warnings of such type/subtype are emitted because it's not always clear.

sphinx/environment/adapters/toctree.py
```diff
@@ -315,19 +316,23 @@ def _toctree_entry(
...
             message = __('toctree contains reference to non-included document %r')
+            subtype = 'not_included'
         else:
             message = __('toctree contains reference to nonexisting document %r')
+            subtype = 'not_readable'
 
-        logger.warning(message, ref, location=toctreenode)
-        raise
+        logger.warning(message, ref, location=toctreenode, type='toc', subtype=subtype)
+        raise DocumentError(message % ref) from err
```
picnixz: Why this change? We should probably keep the KeyError because I have no idea if it is caught somewhere else differently

sphinx/application.py
```diff
@@ -1304,11 +1304,9 @@ def is_parallel_allowed(self, typ: str) -> bool:
             allowed = getattr(ext, attrname, None)
             if allowed is None:
                 logger.warning(message_not_declared, ext.name)
-                logger.warning(__('doing serial %s'), typ)
```
picnixz: Now, I don't know it is the case, but if suppressed warnings are teated as errors when enabling `-W`, then I would do the following:

- we should allow warnings to be treated as non-errors *if* they are suppressed *and* if a (new) configuration value specifies it. 
- Use `suppressed_warnings_errors = False` to indicate that suppressed warning are to be ignored but have a `True` value by default.

If this is not the case, namely, if suppressed warnings are never treated as errors *even with* `-W`, then no need for a new configuration value as above.

sphinx/environment/__init__.py
```diff
@@ -725,7 +725,7 @@ def check_consistency(self) -> None:
                 if 'orphan' in self.metadata[docname]:
                     continue
                 logger.warning(__("document isn't included in any toctree"),
-                               location=docname)
+                               location=docname, type='toc', subtype='not_included')
```
picnixz: I would prefer `orphan` but I don't know if it could confuse people (like the warning is to warn that a document is orphan but it wasn't mark as such).

sphinx/environment/adapters/toctree.py
```diff
@@ -315,19 +316,23 @@ def _toctree_entry(
...
-    except KeyError:
-        # this is raised if the included file does not exist
+                           ref, location=toctreenode, type='toc', subtype='no_title')
+    except KeyError as err:
+        # It may happen only after `directives/other.py` parsing when
+        # the included document changed/disappeared. Thus, it is a not correctable warning.
         ref_path = env.doc2path(ref, False)
         if excluded(ref_path):
             message = __('toctree contains reference to excluded document %r')
+            subtype = 'excluded'
```
picnixz, line 361: I see you are using this subtype but where is the toc.excluded actually used since it wasn't introduced in this PR?

sphinx/util/nodes.py
```diff
@@ -424,7 +424,8 @@ def inline_all_toctrees(builder: Builder, docnameset: set[str], docname: str,
                     docnameset.add(includefile)
                 except Exception:
                     logger.warning(__('toctree contains ref to nonexisting file %r'),
-                                   includefile, location=docname)
+                                   includefile, location=docname,
+                                   type='toc', subtype='not_readable')
```
picnixz: Same question as for `excluded`.

doc/usage/configuration.rst
```diff
@@ -393,9 +401,19 @@ General configuration
...
       Added ``i18n.inconsistent_references``
 
-    .. versionadded:: 7.1
+   .. versionadded:: 7.1
+
+      Added ``index`` warning type.
+
+   .. versionchanged:: 7.2
+
+      Make ``toc``, ``i18n`` and ``index`` warning types cover all relative
```
picnixz: I think this is not needed (only a changelog entry is needed). The `versionchanged` generally concerns changes at the core level of the object but you didn't change anything for `suppress_warnings` itself, right?

(Well I see that contributors actually hesitated a lot between aded/changed when introducing a new category haha).

As such, I would keep `versionchanged` for all `added ...` and remove the `Make ...`

doc/usage/configuration.rst
```diff
@@ -393,9 +401,19 @@ General configuration
...
+      warnings.
+
+      Added ``i18n.babel``, ``i18n.not_readable``, ``i18n.not_writeable``,
+      ``toc.duplicated``, ``toc.glob``, ``toc.no_title``, ``toc.no_included``.
 
-        Added ``index`` warning type.
+      Towards :doc:`extdev/index:ext-metadata`, Added TODO.
```
picnixz: I think you will fill this up later.

sphinx/util/i18n.py
```diff
@@ -180,10 +182,13 @@ def babel_format_date(date: datetime, format: str, locale: str,
         return formatter(date, format, locale=locale)
     except (ValueError, babel.core.UnknownLocaleError):
         # fallback to English
+        logger.warning(__('Invalid locale for babel. Fallback locale to `en`.'),
```
picnixz: Perhaps you could indicate the bad locale? because such warming is not necessarily helpful otherwise.

sphinx/environment/adapters/toctree.py
Tiger3018: As the comment(L321) indicates, [/sphinx/directives/other.py#L141](/sphinx-doc/sphinx/blame/v7.2.5/sphinx/directives/other.py#L141)

doc/usage/configuration.rst
Tiger3018: Yeah, you are right. Going to move them in next commit.

Introducing some new category can be a headache for newbies without the knowledge of project structure. :joy:

sphinx/environment/adapters/toctree.py
picnixz: Ah sorry! actually I am reviewing everything on phone so it's quite hard to switch within the code (and then I forgot what I was looking at).

sphinx/environment/__init__.py
Tiger3018: Maybe we can change the warning log to `"document isn't included in any toctree, yet not marked as orphan in metadata."`

picnixz: I prefer a shorter message. I think the subtype could still be `orphan`, but then,  is there any possibility that `toc.orphan` could mean something else for users? I would say "ok that means there was an issue with an orphan doc, be it a doc marked explicitly with `:orphan:` or a doc detected as orphan but not marked as such". Would it make sense or would it ne too confusing? also, we could document each type/subtype and its meaning.

Otherwise we can keep `not_included`.

sphinx/environment/adapters/toctree.py
Tiger3018: Only in [environment/__init__.py#L455](/sphinx-doc/sphinx/blame/v7.2.5/sphinx/environment/__init__.py#L455) when there is an OSError. 

`title = ref` and `nodes.reference(..., *[nodes.Text(title)])` in following lines might be only KeyError source. I think this DocumentError could be outstanding for users to check original document status or filesystem status.

sphinx/environment/__init__.py
Tiger3018: `orphan` type is better.

I think a new page/file for warning message types and solutions in this patch is needed. This patch may only include toc/i18n/index types. This file might be `Resolve Warnings<doc/usage/catch.rst>` in <https://www.sphinx-doc.org/en/master/usage/index.html>

doc/usage/configuration.rst
Tiger3018: Personally I like underline and slash both, but there is **only** slash in the existing subtype.

Edit: or change `type.subtype` to `namespace/type.subtype`?
A lot of refactoring though lol... Maybe a new patch

picnixz: For me, it would make more sense but it's only a matter of preferences. I tent to use namespaces with / because it reflects for instance apt sources, or can be interpreted as "main entrypoint is contrib for contributors, or ext for extensions". Having a `_` generally limits me to using words with underscores everywhere (like, for extensions that are packaged in some domain, having `ext/pkg.ext1` and `ext/plg.ext2` feels nicer than `ext_pkg.ext1` as I would understand `ext_pkg` as a single component.

But again, I prefer having a feedback from other people.

sphinx/environment/adapters/toctree.py
picnixz: For now, I would keep the exception type because they might catch it or assume its type. Changing the exception type could lead to incompatible changes and this requires a major version.

cale-vdw:
Hi, is this PR still in progress to be merged soon @Tiger3018 @picnixz?

picnixz:
I personally am not involved in this PR and won't be taking over so it's up to the author to decide when it's ready. I'll nonetheless review the PR when it's updated.

Issue #11662, opened 2023-08-30
Sphinx 7.2.5 breaks with PyO3: "modules may only be initialized once per interpreter process"


pbybee:
### Describe the bug

New error message with this patch release. 7.2.4 succeeds in building the same documents.
``PyO3 modules may only be initialized once per interpreter process``

```
Full traceback
Traceback (most recent call last):
  File ".venv/lib/python3.10/site-packages/sphinx/cmd/build.py", line 298, in build_main
    app.build(args.force_all, args.filenames)
  File ".venv/lib/python3.10/site-packages/sphinx/application.py", line 351, in build
    self.builder.build_all()
  File ".venv/lib/python3.10/site-packages/sphinx/builders/__init__.py", line 252, in build_all
    self.build(None, summary=__('all source files'), method='all')
  File ".venv/lib/python3.10/site-packages/sphinx/builders/__init__.py", line 312, in build
    with logging.pending_warnings():
  File "/usr/lib/python3.10/contextlib.py", line 142, in __exit__
    next(self.gen)
  File ".venv/lib/python3.10/site-packages/sphinx/util/logging.py", line 222, in pending_warnings
    memhandler.flushTo(logger)
  File ".venv/lib/python3.10/site-packages/sphinx/util/logging.py", line 187, in flushTo
    logger.handle(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1634, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1696, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 964, in handle
    rv = self.filter(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 821, in filter
    result = f.filter(record)
  File ".venv/lib/python3.10/site-packages/sphinx/util/logging.py", line 427, in filter
    raise exc
sphinx.errors.SphinxWarning: autodoc: failed to import module 'test' from module 'test'; the following exception was raised:
PyO3 modules may only be initialized once per interpreter process
```

### How to Reproduce

install latest sphinx and run
`sphinx-build -WaE docs/project docs/project/_build/html`

### Environment Information

```text
Platform:              linux; (Linux-5.4.0-149-generic-x86_64-with-glibc2.31)
Python version:        3.10.12 (main, Jun  7 2023, 12:45:35) [GCC 9.4.0])
Python implementation: CPython
Sphinx version:        7.2.5
[...]

PR #11679, opened 2023-09-07
Leverage `importlib.reload` for reloading modules

godlygeek:
### Feature or Bugfix
- Bugfix

### Purpose
Rather than directly manipulating `sys.modules` ourselves, rely on `importlib.reload` to do it, as it works in more cases (particularly for Rust extension modules built with PyO3).

Also, import modules with `typing.TYPE_CHECKING` set to `False` first, and reload them with it set to true. This way, if the reload fails, things are likely to still be in a valid state from the first, successful import.

### Relates
- #11662


sphinx/ext/autodoc/importer.py
```diff
@@ -84,22 +85,19 @@ def import_object(modname: str, objpath: list[str], objtype: str = '',
...
-                    # if that fails (e.g. circular import), retry with
-                    # ``typing.TYPE_CHECKING == False`` after reverting
-                    # changes made to ``sys.modules`` by the failed try
-                    for m in [m for m in sys.modules if m not in orig_modules]:
-                        sys.modules.pop(m)
-
-                    typing.TYPE_CHECKING = False
-                    module = import_module(modname, warningiserror=warningiserror)
+                    # Ignore failures; we've already successfully loaded these modules.
+                    with contextlib.suppress(Exception):
```
AA-Turner: Can we have a narrower exception type than Exception?

godlygeek: I don't think so -- at least, not without sacrificing correctness. If I change this to just catch `ImportError`, then doing:
```rest
.. automodule:: pydantic
```
fails with::
```text
>       _PartialClsOrStaticMethod: TypeAlias = Union[classmethod[Any, Any, Any], staticmethod[Any, Any], partialmethod[Any]]
E       TypeError: type 'classmethod' is not subscriptable
```

Let me flip the question around on you: given that the module was already imported successfully with `typing.TYPE_CHECKING == False`, which sorts of exceptions encountered while reloading the module with `typing.TYPE_CHECKING == True` should cause `import_object` to fail instead of using the already successfully imported module?

AA-Turner: Good point. I take it from your testing that e.g. if ``pydantic`` is successfully imported, and then fails on the reload, the initial imported module is still usable, coherent, etc?

godlygeek:
I'm gonna mark this as draft for now, and investigate the report in https://github.com/sphinx-doc/sphinx/issues/11662#issuecomment-1710733894 that this still isn't enough for some modules...

sphinx/ext/autodoc/importer.py
godlygeek: That's my expectation, and it seems to be the case for pydantic, at least up to the point where a basic `.. automodule:: pydantic` succeeds. I'm not sure how to test it more thoroughly, though. I do suspect that this is still going to be fragile in ways that we can't predict, the hope is just that it works for more cases than both what's in #11511 and what's in #11645

AA-Turner:
Matt -- I've added a basic escape hatch through ``SPHINX_AUTODOC_DO_NOT_RELOAD_MODULES``.

AA-Turner:
I'd like to merge this as it is an improvement over the status quo, even if not perfect -- @godlygeek did you make any progress in your investigations, or any other observations that would halt merging this?

A

godlygeek:
Nothing beyond what I noted in https://github.com/sphinx-doc/sphinx/issues/11662#issuecomment-1711017608

I still think this is probably _more_ correct than both what shipped in 7.2.0 and what shipped in 7.2.5, but I also think that this entire approach of setting `typing.TYPE_CHECKING = True` is fundamentally unwise. I think adding an escape hatch is a very good idea, but given that we know this still doesn't work successfully for some modules (even some that worked with 7.2.4!), I really think that this should be opt-in rather than opt-out.

PR #11683, opened 2023-09-14
Docs: Use Python 3 print() functions in examples

hugovk:
### Feature or Bugfix

- Bugfix?

### Purpose

Let's use Python 3 `print()` functions in the examples:

* https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-samp
* https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-code-block-emphasize-lines


AA-Turner (approved):
Only mildly mortifying...

Thanks Hugo!

A

PR #11684, opened 2023-09-14
linkcheck builder: optionally allow HTTP 401 status code hyperlinks to be reported broken

### LINKED ISSUE #11433: linkchecker: update handling of HTTP 401 unauthorized responses
### Describe the bug

In Sphinx releases since v1.3 the `linkcheck` builder has [considered an HTTP 401 (unauthorized) response to be acceptable](https://github.com/sphinx-doc/sphinx/blob/b87ce32e7dc09773f9e71305e66e8d6aead53dd1/sphinx/builders/linkcheck.py#L186-L189) as confirmation that a hyperlinked reference is available and working (see #1687 for context).

That made sense at the time -- support for authentication in the `linkcheck` builder didn't arrive until Sphinx release v2.3.0 (see #6816) -- but I think that it might be an improvement to update the linkchecker to report references that return HTTP 401 as either 'ignored' or 'broken'.

Basically I think the ideal is to communicate to documentation authors and reviewers -- the likely users of the `linkcheck` builder -- where there are gaps or problems in the links that downstream readers may experience.

Sure, in some environments downstream readers may be expected to provide authentication information to access documents -- but for many projects the documentation is provided public and without requiring such auth.  In either case, if an unauthorized response is received, the linkcheck user should be notified that something may not be accessible for their readers, and this could hint to them that they may need to update their config to add authentication to match user expectations.

### How to Reproduce

I don't have a repro case to offer for this at the moment, but may add one later.

### Environment Information

```text
Affects sphinx v1.3 onwards.
```

### Sphinx extensions

```python
N/A
```

### Additional context

Discovered during refactoring in #11432.

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- For [legacy reasons](https://github.com/sphinx-doc/sphinx/issues/11433#issue-1718443556), HTTP 401 statuses had been considered acceptable as success during linkchecking in the past; this pull request allows deployments to change that behaviour.  This is supported by the fact that [Sphinx can now provide authentication credentials to servers during linkchecking](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_auth).

### Detail
- The HTTP 401 status code indicates that access to a web resource requires authentication, and that the client has not provided that authentication.
- Webservers often configure authentication on a per-directory or per-path basis, so client receipt of an authentication-required status code _does not_ necessarily confirm that a URL is valid.
- :spiral_notepad: With this changeset, the ability to treat HTTP 401 responses from hyperlinks as `broken` is _opt-in_ - deployments must set the `linkcheck_allow_unauthorized` config option to `False` to consider unauthorized links as unavailable during linkchecking.  The default reporting status for unauthorized links will change to `broken` as part of the major-version upgrade to Sphinx 8.0 in future.

### Relates
- Resolves #11433.


sphinx/builders/linkcheck.py
```diff
@@ -437,9 +437,8 @@ def _check_uri(self, uri: str, hyperlink: Hyperlink) -> tuple[str, str, int]:
             except HTTPError as err:
                 error_message = str(err)
 
-                # Unauthorised: the reference probably exists
                 if status_code == 401:
-                    return 'working', 'unauthorized', 0
+                    return 'broken', 'unauthorized', 0
```
picnixz: * I think that the 401 still makes sense if people do not want to expose credentials in their configuration and only assume that something exists (and could return a 401). 

  As such, I think we could keep the `'working'` but change a bit the way that `linkcheck_auth` is handled:
  
  * If we specify "fake" credentials by `None`, we keep `'working'` and do not emit warnings.
  * If no fake credential for the corresponding URL is specified, we keep the `'working'`, but we emit a warning saying that the content couldn't be accessed and that users should specify credentials (fake or not). 

That way, this also ensures that credentials are not exposed unintentionally and that the contents with 401 errors are detected correctly.

* It's more of a follow-up, but what about using an enumeration (or at least mark the possible string statuses we returm using `Literal`) for the `status` instead of using `'working'`, `'broken'`, etc?

  In `process_result`, we use if-case but   using `match` instead + enumerations may be more elegant and clearer (also we wouldn't have an arbitrary status code being emitted).

jayaddison: > I think that the 401 still makes sense if people do not want to expose credentials in their configuration and only assume that something exists (and could return a 401).

Hmm.. point understood, although I think that even when an URL has no linkcheck credentials configured (the "no fake credential" case), an HTTP 401 response should be considered a broken link, because it means that the link is unchecked.  This is probably the core of the disruptive/controversial angle I have on this.

If for some reason a user is unable to provide credentials for some hyperlinks included in a documentation set, but wants a success report from linkchecking despite that, I'd argue they should use ~~`linkcheck_anchors_ignore`~~ `linkcheck_ignore` to skip the relevant websites.

I'm still thinking about the first point -- using `None` as a special marker for intentionally-empty credentials.  It seems a bit fragile and too-clever; I like simple and hard-to-misunderstand things, because I misunderstand a lot.

> It's more of a follow-up, but what about using an enumeration (or at least mark the possible string statuses we returm using `Literal`) for the `status` instead of using `'working'`, `'broken'`, etc?

:+1: 

> In `process_result`, we use if-case but using `match` instead + enumerations may be more elegant and clearer (also we wouldn't have an arbitrary status code being emitted).

Mostly agree, although we still support Py3.9 at the moment - so this could be a future enhancement?

picnixz: Fine by me using `linkcheck_anchors_ignore` although it's a bit counterintuitive in this case (like I don't really want to ignore).

> using None as a special marker for intentionally-empty credentials

It was an example. We could have another configuration value like `linkcheck_auth_bypass` and put what links we expect to have 401.

Concerning the match, I forgot it was introduced in 3.10.

---

Actually, when writing my comment, I thought about a much more elegant solution, namely, you could specify the HTTP response code to expect for specific links and treat them as "working".

jayaddison: > Fine by me using `linkcheck_anchors_ignore` although it's a bit counterintuitive in this case (like I don't really want to ignore).

Let me try to make a convicing argument that it makes sense :)  (I could be wrong!  in which case that'll help too)

The HTTP 401 response is zero-information in terms of hyperlink validity.  It does confirm that the client should use auth if it wants to gain more-than-zero information, but that's the only feedback it provides -- and we hide that from users at the moment.

So we're making network requests that fail, and we're not informing the user about it.  I think we should start by informing the user -- and then they either choose to find auth to gain greater-than-zero info, or they ignore the links and reduce the outbound traffic (and linkchecking time).

picnixz: >  I think we should start by informing the user 

Agreed. Actually, by counterintuitive, I meant that using the `linkcheck_anchors_ignore` was weird to me since I would have used it to suppress errors related to anchors rather than to HTTP response codes.

So, we could keep what you suggested.

jayaddison: Agh, sorry.  I should have written `linkcheck_ignore` instead.  I completely failed to notice/parse the word `anchor` in there.

picnixz: Now it's ok for me !

jayaddison: :)  thank you!

jayaddison: > It's more of a follow-up, but what about using an enumeration (or at least mark the possible string statuses we returm using `Literal`) for the `status` instead of using `'working'`, `'broken'`, etc?
> In `process_result`, we use if-case but   using `match` instead + enumerations may be more elegant and clearer (also we wouldn't have an arbitrary status code being emitted).

I'm working on some refactoring for this at the moment; the `CheckResult` class relies upon being JSON-serializable at the moment, and I don't feel like introducing custom serialization code, so I've opted to use the `StrEnum` type -- `Enum` with string values isn't serializable by default.

Taking that approach adds a dependency on Py3.11, so I'm experimenting with adding `match` statements at the same time.

I did read up a little bit about using `Literal` - it looks like it's mostly a type-checking aid?  I _think_ I'd prefer the enum route, even if it may take a while before it could land in the codebase.  Maybe I'll change my mind as I learn more, though..

AA-Turner:
For this to go into 7.3, it should be opt-in; we can flip to opt-out in 8.0.

A

jayaddison:
Ok, thanks @AA-Turner.

I could wait until 8.0 to merge this, if that's simpler -- or I'll add a config option called `linkcheck_permit_unauthorized_responses`, default `True` (existing behaviour).

More concise and/or better config option name ideas welcome.. I'd like to communicate the fact that it's server-communicated auth status that is the condition, and for the way those cases are handled to be fairly clear from the option name.

picnixz:
An alternative: `linkcheck_ignore_error_codes` containing 401 by default. 

Or an alternative name: `linkcheck_ignore_unauthorized` set to true.

jayaddison:
Hmm.. the error-code list approach could be a neat solution, but gives me a sense of too much configurability.  Roughly speaking, I think HTTP status codes are well-enough defined to determine success and failure in the case of linkchecking.

The `ignore` keyword would be nice because it'd visually align with other configuration options and re-uses existing terminology.. but the use case is different here, I think.  We _do_ check these unauthorized URLs, so they're not ignored.

`strict` could also be OK; my concern there is that it infers that there could be a category of checks -- but since this is intended to be a temporary flag until 8.0, I'd prefer to pick something that can't accidentally become repurposed for other things prior to removal.

To shoot down one of my own other alternative ideas as well: `linkcheck_allow_unauthorized` - concise, re-uses the word allow -- but could be confusing because there is client `auth`, and it's perfectly fine for the linkchecker to make unauthorized requests in most cases.  What we care about is whether the server is saying the request was not authorized.

So.. I suppose `linkcheck_allow_unauthorized_responses` could be OK.  `linkcheck_tolerate_unauthorized_responses`?

picnixz:
I'm not fond of long configuration value names. And I would prefer using standard terminology (i.e., not "tolerate"). I think we should pick something from: allow, ignore, skip or strict.

We could also have something like `linkcheck_ignore_errors = ['unauthorized']` (meaning, you do check but ignore specific errors, and not the link itself). Since we also have a bunch of other errors that turn into warnings, we could also suppress them like that.

jayaddison:
> I'm not fond of long configuration value names. And I would prefer using standard terminology (i.e., not "tolerate"). I think we should pick something from: allow, ignore, skip or strict.

Ok, agreed on both of those.
 
> We could also have something like `linkcheck_ignore_errors = ['unauthorized']` (meaning, you do check but ignore specific errors, and not the link itself). Since we also have a bunch of other errors that turn into warnings, we could also suppress them like that.

I'd like to keep the scope of this change (and the config option) small, and the management of log-level for some of those errors/warnings seems a bit unrelated.

However: it does make me wonder whether it'd be OK to add a warning message even when HTTP 401 is treated as success?

Regarding the option name: probably enough bikeshedding, so `linkcheck_allow_unauthorized` is it, I think.  I resolved my own concern about the ambiguity with client `auth` because I suppose technically that refers to authentication and not authorization.

picnixz:
> whether it'd be OK to add a warning message even when HTTP 401 is treated as success?

Not good if people are turning warnings into errors.

CHANGES.rst
```diff
@@ -304,6 +304,9 @@ Features added
 
 Bugs fixed
 ----------
+* #11433: The ``linkcheck_allow_unauthorized`` configuration option (default
+  value ``true``) has been added to allow treatment of HTTP 401 (unauthorized)
+  server responses as broken hyperlinks.
```
jayaddison: I've a pending commit to clarify this wording to:

```
* #11433: The ``linkcheck_allow_unauthorized`` configuration option (default
  value ``true``) has been added.  Set this option to ``false`` to handle
  HTTP 401 (unauthorized) server responses as broken hyperlinks.
```

(I don't like pushing multiple commits while continuous integration is pending, when possible)

jayaddison: ...and I need to update the Sphinx docs to document this option.  I'll do that now.

sphinx/builders/linkcheck.py
```diff
@@ -437,9 +438,10 @@ def _check_uri(self, uri: str, hyperlink: Hyperlink) -> tuple[str, str, int]:
             except HTTPError as err:
                 error_message = str(err)
 
-                # Unauthorised: the reference probably exists
+                # Unauthorized: the client did not provide required credentials
                 if status_code == 401:
-                    return 'working', 'unauthorized', 0
+                    status = 'working' if self.allow_unauthorized else 'broken'
```
jayaddison: I suppose we'll also want a deprecation warning when the `linkcheck_allow_unauthorized` option is configured to `False`, to indicate that it will be removed.

I'm pausing work at the moment, but I'll look into that within the next day or two.

jayaddison: > I suppose we'll also want a deprecation warning when the `linkcheck_allow_unauthorized` option is configured to `False`, to indicate that it will be removed.

Self-correction (implemented): if we're removing the setting, we'll want to issue a warning if _any_ value is configured for it.

jayaddison:
I don't have any further changes/commits planned here at the moment and would be glad for more review when possible.

doc/usage/configuration.rst
```diff
@@ -2915,6 +2915,14 @@ Options for the linkcheck builder
 
    .. versionadded:: 4.4
 
+.. confval:: linkcheck_allow_unauthorized
```
picnixz, line 2918: To prepare for 8.0, perhaps you could say that it will be removed in 8.0?

sphinx/builders/linkcheck.py
```diff
@@ -283,13 +285,23 @@ def __init__(self, config: Config,
         self.allowed_redirects = config.linkcheck_allowed_redirects
         self.retries: int = config.linkcheck_retries
         self.rate_limit_timeout = config.linkcheck_rate_limit_timeout
+        self.allow_unauthorized = config.linkcheck_allow_unauthorized
```
picnixz: Since we know that it will be removed, maybe we could make it a private variable?

```suggestion
        self._allow_unauthorized = config.linkcheck_allow_unauthorized
```

sphinx/builders/linkcheck.py
```diff
@@ -625,6 +638,7 @@ def setup(app: Sphinx) -> dict[str, Any]:
     app.add_config_value('linkcheck_anchors_ignore', ['^!'], False)
     app.add_config_value('linkcheck_anchors_ignore_for_url', (), False, (tuple, list))
     app.add_config_value('linkcheck_rate_limit_timeout', 300.0, False)
+    app.add_config_value('linkcheck_allow_unauthorized', None, False)
```
picnixz:
```suggestion
    app.add_config_value('linkcheck_allow_unauthorized', True, False)
```

tests/test_build_linkcheck.py
```diff
@@ -396,10 +416,9 @@ def test_auth_header_no_match(app):
     with open(app.outdir / "output.json", encoding="utf-8") as fp:
         content = json.load(fp)
 
-    # TODO: should this test's webserver return HTTP 401 here?
-    # https://github.com/sphinx-doc/sphinx/issues/11433
-    assert content["info"] == "403 Client Error: Forbidden for url: http://localhost:7777/"
-    assert content["status"] == "broken"
+    # This link is considered working based on the default linkcheck_allow_unauthorized=true
```
picnixz, line 424: It's a bit of a lie since you actually have a default set to `None`.

sphinx/builders/linkcheck.py
```diff
@@ -283,13 +285,23 @@ def __init__(self, config: Config,
...
         self.rate_limit_timeout = config.linkcheck_rate_limit_timeout
+        self.allow_unauthorized = config.linkcheck_allow_unauthorized
 
         self.user_agent = config.user_agent
         self.tls_verify = config.tls_verify
         self.tls_cacerts = config.tls_cacerts
 
         self._session = requests._Session()
 
+        if config.linkcheck_allow_unauthorized is not None:
```
picnixz: I think it's better to just keep `linkcheck_allow_unauthorized = True` by default. At least you don't have the `None` case, but you could still emit a warning if `linkcheck_allow_unauthorized` is `False`.

```suggestion
        if not config.linkcheck_allow_unauthorized:
```

I would suggest emitting the warning before opening the session also (namely the sooner the better).

sphinx/builders/linkcheck.py
```diff
@@ -437,9 +449,10 @@ def _check_uri(self, uri: str, hyperlink: Hyperlink) -> tuple[str, str, int]:
             except HTTPError as err:
                 error_message = str(err)
 
-                # Unauthorised: the reference probably exists
+                # Unauthorized: the client did not provide required credentials
                 if status_code == 401:
-                    return 'working', 'unauthorized', 0
+                    status = 'broken' if self.allow_unauthorized is False else 'working'
```
picnixz:
```suggestion
                    status = 'working' if self._allow_unauthorized else 'broken'
```

tests/test_build_linkcheck.py
```diff
@@ -386,6 +391,21 @@ def test_auth_header_uses_first_match(app):
     assert content["status"] == "working"
 
 
+@pytest.mark.sphinx(
+    'linkcheck', testroot='linkcheck-localserver', freshenv=True,
+    confoverrides={'linkcheck_allow_unauthorized': False})
+def test_unauthorized_broken(app):
+    warnings.filterwarnings('ignore', category=DeprecationWarning)
```
picnixz:
```suggestion
    warnings.filterwarnings('ignore', category=RemovedInSphinx80Warning)
```

AA-Turner:
> I could wait until 8.0 to merge this

For X.0 releases, I prefer to have a 'clean' release with only removals/changes to defaults -- a new feature I think should be merged in a point release, with an opt-in.

sphinx/builders/linkcheck.py
jayaddison: Ok, that seems fair.  That was my original plan (to leave the default as `True`) - I changed my mind [here](https://github.com/sphinx-doc/sphinx/pull/11684#discussion_r1334501185) because I figured that we should warn about settings that will be removed regardless of what they are configured to...

...but I suppose it's also true that a deployment that has opted-in (configured `False` explicitly) doesn't really need to know that the setting will be removed, so long as it becomes the default at removal-time.

Basically: I'm looking for reassurance / a double-check that a default of `True` without a warning is OK.  I think it is, and will begin to make that adjustment.

Edit: fixup for boolean value in middle sentence.  the first and third sentences are correct, I believe.

jayaddison: Sorry, I made a mistake in that middle sentence:

> ...but I suppose it's also true that a deployment that has opted-in (configured `True` explicitly) doesn't really need to know that the setting will be removed, so long as it becomes the default at removal-time.

That should read: "opted in (configured `False` explicitly)".

Booleans and opt-in/out breaking my brain.

jayaddison: No, I'm still having trouble figuring out what to do here.  I believe the following are facts:

  * We don't want to add warnings where deployments wouldn't currently see them, because warnings are often treated as errors and we don't want unnecessary breakage.
  * We want the default behaviour to be the existing "HTTP 401 is acceptable" logic.
  * We want to warn about the configuration setting.
  * The configuration setting only has an effect if it is configured to `False` (change to the planned future behaviour).

If a deployment has opted-in to the future behaviour, do they really need a deprecation warning about the setting?

And if a deployment as not opted-in, how do we warn people without introducing possible warning-as-error breakage?

jayaddison: > We want to warn about the configuration setting.

Maybe that's an incorrect assumption here?

(I was the one who [decided that a deprecation warning could make sense](https://github.com/sphinx-doc/sphinx/pull/11684#discussion_r1333134731).  maybe I just created a lot of confusion and extra work for myself)

jayaddison:
Ok, I've reached another review-pause now (no further commits planned).

The main question I have open / to resolve is how and when we should emit deprecation warnings -- if at all -- and whether my [assumptions in this subthread](https://github.com/sphinx-doc/sphinx/pull/11684#discussion_r1337404066) are correct.

sphinx/builders/linkcheck.py
picnixz: > We don't want to add warnings where deployments wouldn't currently see them, because warnings are often treated as errors and we don't want unnecessary breakage.

Indeed.

> We want the default behaviour to be the existing "HTTP 401 is acceptable" logic.

Indeed.

> We want to warn about the configuration setting.

Only if the user changes it, namely only if the default value is changed. In particular, the default value should reflect "HTTP 401 is acceptable". As such, `linkcheck_allow_unauthorized` must be set to `True` by default and can be set to `False` to reflect "401 is not ok".

> The configuration setting only has an effect if it is configured to False (change to the planned future behaviour).

Indeed.

> If a deployment has opted-in to the future behaviour, do they really need a deprecation warning about the setting?

I think so. Btw, I think that deprecation warnings (emitted using the `warnings` module and not using a logger) are *not* treated as errors even with `-W`.

> And if a deployment [h]as not opted-in, how do we warn people without introducing possible warning-as-error breakage?

We don't warn them, namely they will keep their current behaviour. If they don't use the feature, they don't need to know that it will be removed, right? (~~the less they know, the better~~) If warnings are treated as errors, a deprecation warning will still not be considered as an error, so we can emit one when `linkcheck_allow_unauthorized` is set manually. Then, if a 401 occurs, the logger's warning will be treated as an error and we're good.

jayaddison: Perfect, thank you @picnixz - that should allow me to apply the remaining code review suggestions.  I'll push a commit or two with those changes soon.

tests/test_build_linkcheck.py
jayaddison: Resolved coincidentally by the default changing back to `True`.

jayaddison:
Ok, reading back through the commentary, there's one detail I missed:

> For this to go into 7.3, it should be opt-in; we can flip to opt-out in 8.0.

So we should switch the default, not remove the configuration value entirely.  I'll update the deprecation notice about this correspondingly -- and as I understand from the other comments, deprecation warnings don't become errors, so that should be innocuous, and we can remove the warning from 8.0 onwards.

doc/usage/configuration.rst
jayaddison: (resolved with slight adjustment: as suggested [here](https://github.com/sphinx-doc/sphinx/pull/11684#issuecomment-1728675463), a flip to opt-out by default in 8.0 would seem to make sense, allowing more time for deployments to adjust)

sphinx/builders/linkcheck.py
```diff
@@ -66,6 +68,15 @@ def init(self) -> None:
         # set a timeout for non-responding servers
         socket.setdefaulttimeout(5.0)
 
+        if not self.config.linkcheck_allow_unauthorized:
+            deprecation_msg = (
+                "The default value for 'linkcheck_allow_unauthorized' will change "
+                "from `True` in Sphinx 7.3+ to `False`, meaning that HTTP 401 "
+                "unauthorized responses will be reported as broken by default. "
+                "See sphinx-doc/sphinx#11433 for details."
```
jayaddison: I know I probably sound like a broken record with this, but I'm still puzzling over exactly the audience we want to warn with this message.

At the moment it's warning people who've opted-in, who presumably have already read the changelog/description.

I'd like to be 99.9% sure about this so that the relevant audiences are informed.

picnixz: Yes, you want to warn people that a feature will be soon deprecated. In other words, you only want to warn people using it (others don't care since they may not be aware of). If you want to warn people reading the changelog but not using the feature, it may be good to add a small comment in the changelog telling them "hey, it's only for 7.3 - 7.x that this feature is supported, so it's really only for a transition".

Now, what you could do for people to make them the transition is actually warn when you encounter a 401 *even if* `linkcheck_allow_unauthorized` is true but using `warnings` and not the logger! that way, people who don't know about the feature will know that in 8.x they need to be careful about this (this is one thing I forgot to mention actually, so my bad).

sphinx/builders/linkcheck.py
jayaddison: > I think so. Btw, I think that deprecation warnings (emitted using the warnings module and not using a logger) are not treated as errors even with -W.

I've been double-checking this, and it looks like `DeprecationWarning` (and subclasses, like `RemovedInSphinx80Warning`) _are_ still converted to exceptions by `-Werror` (and equivalents)).

picnixz: Wait, how come? I've tried locally and using sphinx option `-W` (not Python `-W`!).

jayaddison: Agh!  Ok.  I had been reading the Python `warnings` documentation and incorrectly assumed that was the `-W` you were referencing; I didn't check the `sphinx-build` options until now.  Ok.

After checking the `sphinx-build` `-W` option: you're absolutely correct, yep - warnings can be emitted with that flag enabled, and (as expected) the program continues.

sphinx/builders/linkcheck.py
jayaddison: That makes sense, yep.  So for people who already have HTTP 401s occurring during linkchecking, we can inform them that a change-in-behaviour is on the horizon.

Please let me know if that interpretation is correct; I'll prepare that additional warning code locally.  And thanks for being patient, for some reason I've found this warning logic tricky to reason about, and want to get it correct.

sphinx/builders/linkcheck.py
```diff
@@ -437,9 +449,30 @@ def _check_uri(self, uri: str, hyperlink: Hyperlink) -> tuple[str, str, int]:
...
 
-                # Unauthorised: the reference probably exists
+                # Unauthorized: the client did not provide required credentials
                 if status_code == 401:
-                    return 'working', 'unauthorized', 0
+                    if self._allow_unauthorized:
+                        deprecation_msg = (
+                            "\n---\n"
+                            "The linkcheck builder encountered an HTTP 401 "
+                            f"(unauthorized) response for the URI {uri}, and "
```
jayaddison: Adding the `uri` here could make this spammy for some deployments, if they are encountering large numbers of HTTP 401s.  I would guess (?) that the typical case is likely a few/rare cases, though - in which case the message is more actionable.  I could be convinced to remove the dynamic URI, though.

jayaddison: (to explain that: the [default Python warning logic](https://docs.python.org/3/library/warnings.html#the-warnings-filter) is to print each _unique_ warning -- based on the message, module and line number -- once.   having a dynamic URI means that this message would appear once per URI)

picnixz: Ok, here should be the final draft (dw I also want to be sure about what we are doing).

----

Currently, a broken link is reported either with `logger.info` in red or with `logger.warn` when `-W` is used. In the future, we want all 401 to be "broken". To make the transition as smooth as possible, I suggest the following:

#### Case 1 

If the configuration value is not changed, 401 are treated as working and reported by `logger.info` in green but this message will contain the link and the fact that it was not authorized.

However, a depreciation warning needs to be emitted. Since the user didn't touch the configuration, this depr. warn.  needs to be emitted when encountering  the first 401. Users will know which links to change since (working) 401 are actually printed.

Note that it doesn't matter whether `-W` is set since 401 are considered as working anyway.

#### Case 2

If the configuration value is changed by the user, then 401 are considered as broken. Since they opted in for this feature in 7.3.x, this means that they want these 401 to be turned into real warnings in 8.x. Note that treating warnings as errors is not problematic since users asked for 401 to be explicitly broken. In particular, with `-W`, they expect them to be warnings (hence errors), and the build to fail. Without `-W`, they still expect a red message. So we simply need to report 401 as broken.

In addition, since they touched the configuration value, they will be warned by a deprecation warning in the constructor. Again, this warning won't cause any failure with `-W`.

jayaddison: Both of those descriptions match my understanding of the intended behaviour, with one small detail/confirmation to check.  Thank you very much @picnixz for writing this up :+1: 

My detail check is in response to this statement in case 1:

> this depr. warn. needs to be emitted when encountering the first 401.

I think this implies that we should _not_ include the `uri` inside the deprecation warning logged in this scenario.

picnixz: Indeed. Because they are already printed separately so I don't think it's good to rewrite them once again.

CHANGES.rst
```diff
@@ -34,6 +34,10 @@ Bugs fixed
 * #11483: singlehtml builder: Fix MathJax lazy loading when the index does not
   contain any math equations.
   Patch by Bénédikt Tran.
+* #11433: Added the ``linkcheck_allow_unauthorized`` configuration option.
+  Set this option to ``false`` to report HTTP 401 (unauthorized) server
```
```suggestion
  Set this option to ``False`` to report HTTP 401 (unauthorized) server
```

sphinx/builders/linkcheck.py
```diff
@@ -437,9 +449,30 @@ def _check_uri(self, uri: str, hyperlink: Hyperlink) -> tuple[str, str, int]:
...
+                            "\n"
+                            "To explicitly continue treating unauthorized "
+                            "hyperlink responses as 'working', set the "
+                            "'linkcheck_allow_unauthorized' config option to "
+                            "``True``."
+                            "\n"
+                            "See sphinx-doc/sphinx#11433 for details."
+                            "\n---"
+                        )
+                        warnings.warn(deprecation_msg, RemovedInSphinx80Warning, stacklevel=1)
```
picnixz, line 473: Since the warning is quite long, should we emit it only the first time? (I don't know whether those warnings are only emitted once or not by default or if we are filtering them at the very top level).

jayaddison: By default, warnings are de-duplicated based on message, source module and source line.

So for the following code:

```py
import warnings

def warning(msg):
    warnings.warn(msg)

warning('hello warning 1')
warning('hello warning 1')
warning('hello warning 1')
warning('hello warning 2')
warning('hello warning 3')
warning('hello warning 3')
```

...we'll see the following output:

```
/home/jka/foo.py:4: UserWarning: hello warning 1
  warnings.warn(msg)
/home/jka/foo.py:4: UserWarning: hello warning 2
  warnings.warn(msg)
/home/jka/foo.py:4: UserWarning: hello warning 3
  warnings.warn(msg)
```

(six method calls being de-duplicated to three distinct warning outputs)

However if we run the code with `PYTHONWARNINGS=always`, it'll instead display all of the six emitted warnings - no de-duplication.

(all of the above is generic Python behaviour - none of this is specific to, or overridden by, Sphinx)

picnixz: Oh then it's fine!

jayaddison: Ok, thanks - I did double-check this again a few moments ago to make sure; and yep, when using a `sphinx-build -b linkcheck doc <output-dir>`, a `warnings.warn` message (with a static/constant `message` value) was only emitted once despite the code being reached multiple times.

picnixz: I think it's because the default warning's filter is set to "once" for Python (or something like that).

jayaddison: I suppose it's _possible_ that someone has a large number of HTTP 401s ~~_and_ wants to consider them as `working`~~ _and_ wants to set `PYTHONWARNINGS=always` -- and for them I suppose it's true that this will produce a large amount of log output.

Hopefully it's reasonable to allow them to file a bugreport / contact us if so?  It would be possible to adjust fairly easily but I feel like it might be over-adjusting for a case that I don't know exists.

jayaddison: > I suppose it's _possible_ that someone has a large number of HTTP 401s _and_ wants to consider them as `working` _and_ wants to set `PYTHONWARNINGS=always` -- and for them I suppose it's true that this will produce a large amount of log output.

Self-correction: that middle condition is the default with this change (HTTP 401 continues to mean `working` by default), so that should be omitted because it would be true for most deployments.

picnixz: Yeah. Let's stick to the simple case. If there set "always", then they more or less know what they are doing also.

jayaddison:
I think this is ready - no further commits/changes planned.

jayaddison:
@AA-Turner @picnixz any further feedback for these changes?  I don't know if it'll have much effect in practice, but if the number of people who have broken hyperlinks and don't realize it yet is greater than zero, I'd like to help with that.

picnixz:
Personally I don't have anything else in mind. However note that I willl be extremely busy for the next months as I am leaning towards the end of my thesis.

sphinx/builders/linkcheck.py
AA-Turner:
```suggestion
                "See https://github.com/sphinx-doc/sphinx/issues/11433 for details."
```

jayaddison:
Thank you @AA-Turner! (and again @picnixz for your review of this)

I'll offer the [GitHub issue reference suggestion](https://github.com/sphinx-doc/sphinx/pull/11684#discussion_r1446034411) in a small follow-up PR.

PR #11693, opened 2023-09-21
Deprecate old-style Makefiles in sphinx-quickstart

AA-Turner:


AA-Turner (commented):
Needs a CHANGES entry

PR #11700, opened 2023-10-01
Add 'noindex' directive for web crawlers to basic theme search results

### LINKED ISSUE #11697: HTML search should set noindex
``search.html`` should set the ``<meta name="robots" content="noindex">`` tag.

* xref: https://github.com/python/docsbuild-scripts/issues/53

A

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Web crawlers should generally prefer to crawl the contents of the documentation and build their own index of those; the `search.html` page implements client-side search, so there's no value to web crawlers in indexing it, and we should communicate to them that they avoid doing that.

### Detail
- Add an HTML `meta` element `robots` with `noindex` configured.
- See also:
  - https://en.wikipedia.org/wiki/Noindex
  - https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag
  - https://moz.com/learn/seo/robots-meta-directives

### Relates
- Resolves #11697.

Edit: update the description to match my understanding that all search results in `search.html` are from client-side searches.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,8 @@ Bugs fixed
 * #11483: singlehtml builder: Fix MathJax lazy loading when the index does not
   contain any math equations.
   Patch by Bénédikt Tran.
+* #11697: html basic theme: add 'noindex' meta robots tag.
+  Patch by James Addison.
 
 Testing
 -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,7 +34,7 @@ Bugs fixed
 * #11483: singlehtml builder: Fix MathJax lazy loading when the index does not
   contain any math equations.
   Patch by Bénédikt Tran.
-* #11697: html basic theme: add 'noindex' meta robots tag.
+* #11697: HTML Search: add 'noindex' meta robots tag.
   Patch by James Addison.
 
 Testing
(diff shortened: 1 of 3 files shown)


sphinx/themes/basic/search.html
```diff
@@ -15,8 +15,9 @@
     <script src="{{ pathto('_static/language_data.js', 1) }}"></script>
 {%- endblock %}
 {% block extrahead %}
-  <script src="{{ pathto('searchindex.js', 1) }}" defer></script>
-  {{ super() }}
+    <script src="{{ pathto('searchindex.js', 1) }}" defer></script>
```
AA-Turner, line 18:
```suggestion
    <script src="{{ pathto('searchindex.js', resource=True) }}" defer></script>
```

AA-Turner:
Could we add a CHANGES entry? I think tests can be skipped if you want, though.

A

sphinx/themes/basic/search.html
AA-Turner: Actually let's hold off.

AA-Turner (approved):

CHANGES.rst
```diff
@@ -34,6 +34,8 @@ Bugs fixed
 * #11483: singlehtml builder: Fix MathJax lazy loading when the index does not
   contain any math equations.
   Patch by Bénédikt Tran.
+* #11697: html basic theme: add 'noindex' meta robots tag.
```
```suggestion
* #11697: HTML Search: add 'noindex' meta robots tag.
```

PR #11705, opened 2023-10-04
Add tests for ``:manpage:`` inside a title

### LINKED ISSUE #11673: heading with :manpage: role results in failed build
### Describe the bug

When using `:manpage:` in a heading in `reStructuredText`, the manpage properties for the node does not get parsed.

### How to Reproduce

I created a repository with a minimal example: https://github.com/dviererbe/sphinx-manpage-heading-bug.

Just run `make install && make run`.

### Environment Information

```text
Platform:              linux; (Linux-6.2.0-31-generic-x86_64-with-glibc2.35)
Python version:        3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0])
Sphinx version:        5.3.0
Docutils version:      0.19
Jinja2 version:        3.1.2
```

### Sphinx extensions

_No response_

### Additional context

Attaching a debugger showed that the manpage node attributes were not parsed:

```
> /path/to/project/docs/.sphinx/venv/lib/python3.10/site-packages/sphinx/writers/html5.py(751)visit_manpage()
-> node['refuri'] = self.manpages_url.format(**node.attributes)
(Pdb) node
<manpage: <#text: 'ls(1)'>>
(Pdb) node.attributes
{'ids': [], 'classes': ['manpage'], 'names': [], 'dupnames': [], 'backrefs': [], 'xml:space': 'preserve'}
```

picnixz:
Fix #11673.

By fixing this, I observed a weird behaviour. In order to determine the HTML title of the page, we store a real title node and not just a string. Then, we create a fake document and render it. However, by doing so, the Sphinx transformations are *never* called on that fake document, but the writer *is* called! That means that there are some visitor methods that assume that a transformation was called, but this is not the case for those partial documents.

I think we need to rethink about this in order to decouple the interaction. I don't know if there are other methods that could fail because of that.


AA-Turner:
Could you open an issue for the title rendering piece?

A

sphinx/builders/html/__init__.py
```diff
@@ -425,7 +425,7 @@ def render_partial(self, node: Node | None) -> dict[str, str]:
         if node is None:
             return {'fragment': ''}
 
-        doc = new_document('<partial node>')
+        doc = new_partial_document()
```
AA-Turner: Perhaps just setting an attribute?

```suggestion
        doc = new_document('<partial node>')
        doc['is_partial'] = True
```

and then ``doc.get('is_partial', False)`` to test.

picnixz: I wanted to keep the old behaviour in case there was someone somewhere who relied on that. Also, it's to remind us that we have possible partial documents somewhere and that we always need to be aware of. 

On the other hand, using an attribute *may* lead false positive where a user would define this specific attribute. When using the source + special name we may avoid these cases (is_partial may be a "simple" attribute name that may be used in some projects).

AA-Turner: I'm averse to introducing new infrastructure around this if it isn't something we want to keep long-term. So I would prefer the attribute approach (keeping the document name the same), we could use ``_is_partial`` or some mangled name if we need to.

A

picnixz: > we could use _is_partial or some mangled name if we need to.

That's fine by me then! Actually I don't know whether we'll keep it long-term or not since I don't know if we will be able to fix this issue on partial documents. 

I'll do the update next week (I'me quite busy these days).

picnixz:
So, color me surprised but #11825 actually fixed this PR. So I'm only adding some tests now to ensure that the behaviour is as expected.

PR #11708, opened 2023-10-05
Fix detection of ANSI control sequence support

### LINKED ISSUE #11675: Regression in terminal color auto-detection?
### Describe the bug

I'm not sure if this is caused by Sphinx or readthedocs.org, but when using Sphinx 7.1 and 7.2 on RTD, the log (which doesn't support color) contains some partial ANSI escape sequences like `[2K`.

See https://github.com/readthedocs/readthedocs.org/issues/10702 for details.

### How to Reproduce

Push to RTD using Sphinx >= 7.1 and look at the logs.

### Environment Information

```text
readthedocs.org
```

### Sphinx extensions

_No response_

### Additional context

_No response_

picnixz:
Fix #11675.

@mgeier or @humitos, could you confirm that this patch works as intended?

cc @AA-Turner


mgeier:
> could you confirm that this patch works as intended?

Yes, works like a charm!
See https://readthedocs.org/projects/nbsphinx/builds/22136182/

humitos:
Thanks @mgeier for testing this! I'm happy it's working as expected 🚀 Thanks @picnixz too for fixing the issue 😉

PR #11710, opened 2023-10-05
HTML writer: update filtering logic for meta tags merged from docutils

### LINKED ISSUE #11699: Duplicate emission of HTML meta tags generated by docutils
### Describe the bug

It looks like some HTML `meta` tags are being duplicated in the HTML builder's output, at least when using the Sphinx basic theme.

For example, in the [`jinja` v3.1 documentation here](https://jinja.palletsprojects.com/en/3.1.x/), I see this when using 'view source' on the page:

```html
<!DOCTYPE html>

<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
  <meta name="viewport" content="width=device-width, initial-scale=1">
  ...
```

I think that the reason is that Sphinx combines meta tags from `docutils` and its' own.  Here's where the `docutils` meta tags are filtered, to remove a few other items that would otherwise cause duplicates:

https://github.com/sphinx-doc/sphinx/blob/cf7d2759af0852d67288e58d823d51fe860749ca/sphinx/writers/html.py#L44

The `[2:]` list index filtering looked a bit odd -- and was written in Y2008 (quite a while ago) so I think it's something to do with the problem.

`docutils` [added the meta `viewport` tag](https://github.com/docutils/docutils/commit/cde97fa658d03b8eccb420f33d2df1d66b94f3cd) more recently, in Y2021 - so I think that the Sphinx code doesn't factor this into the meta tag filtering.

One approach could be to filter based on the properties of each tag rather than their position in the visited elements.  I think that the three items that currently make sense to filter are:

  * Any `meta` tag with a `charset` attribute.
  * Any `meta` tag with a `viewport` attribute.
  * Any `meta` tag with a `generator` attribute (this is because `docutils` emits a tag to indicate that it generated documentation; I _think_ that it's reasonable to remove this because the HTML output from Sphinx is fairly far removed from what `docutils` itself would output standalone).

### How to Reproduce

This can be replicated using Sphinx's own documentation set as an input and running the HTML builder.

For example, by running:

```sh
$ sphinx-build -W -b html doc doc_output
```

### Environment Information

```text
Platform:              linux; (Linux-6.5.0-1-rt-amd64-x86_64-with-glibc2.37)
Python version:        3.11.5 (main, Aug 29 2023, 15:31:31) [GCC 13.2.0])
Sphinx version:        7.3.0+/b9c85984d
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Sphinx extensions

```python
N/A
```

### Additional context

N/A

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Reduces duplication of `viewport` HTML `meta` tags on Sphinx documentation pages.

### Detail
- Updates the existing filtering logic to be less list-index-position-dependent.

### Relates
- Resolves #11699.

### DIFF
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -43,10 +43,9 @@ class HTMLWriter(Writer):
             setattr(self, attr, getattr(visitor, attr, None))
         self.clean_meta = ''.join([
             meta_tag for meta_tag in self.visitor.meta
-            if not any (skip_tag in meta_tag for skip_tag in (
-                    'charset="utf-8"',
-                    'name="generator"',
-                    'name="viewport"',
-                )
-            )
+            if not any(skip_tag in meta_tag for skip_tag in (
+                'charset="utf-8"',
+                'name="generator"',
+                'name="viewport"',
+            ))
         ])
--- a/sphinx/writers/html.py
+++ b/sphinx/writers/html.py
@@ -41,4 +41,12 @@ class HTMLWriter(Writer):
                      'footer', 'html_prolog', 'html_head', 'html_title',
                      'html_subtitle', 'html_body'):
             setattr(self, attr, getattr(visitor, attr, None))
-        self.clean_meta = ''.join(self.visitor.meta[2:])
+        self.clean_meta = ''.join([
+            meta_tag for meta_tag in self.visitor.meta
+            if not any (skip_tag in meta_tag for skip_tag in (
+                    'charset="utf-8"',
+                    'name="generator"',
+                    'name="viewport"',
+                )
+            )
+        ])
--- a/tests/test_build_html.py
+++ b/tests/test_build_html.py
@@ -79,7 +79,7 @@ def tail_check(check):
     return checker
 
 
-def check_xpath(etree, fname, path, check, be_found=True):
+def check_xpath(etree, fname, path, check, be_found=True, limit=None):
     nodes = list(etree.findall(path))
     if check is None:
         assert nodes == [], ('found any nodes matching xpath '
@@ -89,10 +89,14 @@ def check_xpath(etree, fname, path, check, be_found=True):
         assert nodes != [], ('did not find any node matching xpath '
                              '%r in file %s' % (path, fname))
     if callable(check):
-        check(nodes)
+        if check(nodes):
+            return
     elif not check:
         # only check for node presence
-        pass
+        if limit is None:
+            return
+        if len(nodes) <= limit:
+            return
     else:
         def get_text(node):
             if node.text is not None:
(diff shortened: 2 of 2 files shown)


sphinx/writers/html.py
```diff
@@ -41,4 +41,11 @@ def translate(self) -> None:
                      'footer', 'html_prolog', 'html_head', 'html_title',
                      'html_subtitle', 'html_body'):
             setattr(self, attr, getattr(visitor, attr, None))
-        self.clean_meta = ''.join(self.visitor.meta[2:])
+        self.clean_meta = ''.join([
+            meta_tag for meta_tag in self.visitor.meta
+            if not any(skip_tag in meta_tag for skip_tag in (
+                'charset="utf-8"',
+                'name="generator"',
+                'name="viewport"',
```
jayaddison, line 49: The tag values are of type `str`, so for now I'm using string-contains comparisons here.  An alternative could be to parse them as HTML markup and then check the names/attributes..

tests/test_build_html.py
```diff
@@ -105,14 +109,17 @@ def get_text(node):
...
         else:
             if all(not rex.search(get_text(node)) for node in nodes):
                 return
 
-        raise AssertionError('%r not found in any node matching '
-                             'path %s in %s: %r' % (check, path, fname,
-                                                    [node.text for node in nodes]))
+    raise AssertionError('%r not found in any node matching '
+                         'path %s in %s: %r' % (check, path, fname,
+                                                [node.text for node in nodes]))
```
jayaddison, line 122: This is a test utility method, but even so this message is now potentially misleading; it can be emitted if the `limit` check fails.

The reason I de-dented it is that a couple of the `if` branches earlier in the method _seemed_ like they should raise an exception if their checks fail.  Fortunately we did not have any true negative (test that passed despite a check failing) cases.

sphinx/writers/html.py
picnixz: Is it possible for the quoted values to be quoted using single quotes?

jayaddison: That would be valid, yep.  I'm usually reluctant to use regular expressions, and I felt that using `charset`, `viewport` and `generator` instead could risk some false positives.. and that parsing the HTML is kinda heavyweight for what's required here, so I went with a simple approach intending for exactly this kind of discussion.

jayaddison:
I'm not keen on the approach I took to attempting a fix here.  It focused on one small detail without figuring out the interactions between Sphinx, themes and `docutils` in terms of HTML general.  I'm going to close this; I'll extract the test-code indentation fixup and offer that as a separate unrelated pull request, though.

tests/test_build_html.py
jayaddison: It looks like I was confused here; it seems intentional that the `if callable` clause and `elif not check` clause allow logic flow to exit the method without an assertion.

In particular: for custom `check` methods -- [`tail_check`](https://github.com/sphinx-doc/sphinx/blob/35965903177c6ed9a6afb62ccd33243a746a3fc0/tests/test_build_html.py#L70-L79) -- they're responsible for raising an exception themselves on failure.

If I were to be extremely nitpicky, I'd suggest that even though the use of a [`pass` statement](https://github.com/sphinx-doc/sphinx/blob/35965903177c6ed9a6afb62ccd33243a746a3fc0/tests/test_build_html.py#L95) _could_ kinda indicate "yep, everything OK in this case, allow control-flow to continue", it's slightly inconsistent with the use of `return` statements to early-exit in other cases.  I think that's a pedantic detail though.

PR #11712, opened 2023-10-06
Autosummary additional links for inherited methods and attributes

satr-cowi:
Adds additional variables that can be used in autosummary templates for while also then splitting out methods and attributes if this is the preferred way of using.

### Feature or Bugfix
- Feature

### Purpose

The "path" is collected for the class from which this member is inherited, and then the name of the inherited method or attribute is appended to this depending on which it is. 

Can then be used in a template like so:
```
.. rubric:: {{ _('Inherited Methods') }}
.. autosummary::
	{% for item in inherited_methods %}
	{{item}}
	{%- endfor %}
```
Links then appear to where the method/attribute is actually defined.

Snapshot from the end result of my docs which use this looks like so:
image

This is my first time contributing to a major opensource package, so do let me know if I've missed something or anything looks off.


doc/usage/extensions/autosummary.rst
```diff
@@ -303,6 +303,27 @@ The following variables are available in the templates:
 
    .. versionadded:: 1.8.0
 
+.. data:: inherited_methods
+
+   List containing names of only inherited methods.  Only available for
+   classes.
+
+   .. versionadded:: 7.2.7
```
picnixz: I think it'll be probably in 7.3 or 8.x (maybe you could leave a todo and we'll update when the PR is accepted).

doc/usage/extensions/autosummary.rst
```diff
@@ -303,6 +303,27 @@ The following variables are available in the templates:
...
+
+.. data:: inherited_attributes
+
+   List containing names of only inherited attributes.  Only available for
+   classes.
+
+   .. versionadded:: 7.2.7
+
+.. data:: inherited_paths
```
picnixz:
```suggestion
.. data:: inherited_qualname
```

We usually refer this path as the fully qualified name or sometimes the canonical name. We use `:canonical:` in Sphinx, so you could also have `canonical_inherited_names`.

doc/usage/extensions/autosummary.rst
```diff
@@ -303,6 +303,27 @@ The following variables are available in the templates:
...
+
+   List containing the full path inherited members. Will only return the closest
```
picnixz: Maybe you should add an example to distinguish between this and the simple name.

sphinx/ext/autosummary/generate.py
```diff
@@ -320,11 +320,34 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
         ns['members'] = dir(obj)
         ns['inherited_members'] = \
             set(dir(obj)) - set(obj.__dict__.keys())
+
         ns['methods'], ns['all_methods'] = \
             _get_members(doc, app, obj, {'method'}, include_public={'__init__'})
         ns['attributes'], ns['all_attributes'] = \
             _get_members(doc, app, obj, {'attribute', 'property'})
 
+        method_string = [m.split(".")[-1] for m in ns['methods']]
```
picnixz: Perhaps it's better to have use `ns['all_methods']` since otherwise you'll only include public ones.

sphinx/ext/autosummary/generate.py
```diff
@@ -320,11 +320,34 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
...
+        attr_string = [m.split(".")[-1] for m in ns['attributes']]
+
+        # Find the first link for this attribute in higher classes
+        ns['inherited_paths'] = []
+        ns['inherited_methods'] = []
+        ns['inherited_attributes'] = []
+
+        for i in (set(dir(obj)) - set(obj.__dict__.keys())):
```
picnixz: Not all classes have a `__dict__` (and sometimes there are weird ones). So I think you should look at the implementation of `_get_members` to see how members are looked-up.

sphinx/ext/autosummary/generate.py
```diff
@@ -320,11 +320,34 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
...
+            for cl in obj.__mro__:
+                local = set(cl.__dict__.keys())
+                if i in local:
+                    cl_str = cl.__module__ + "." + cl.__name__ + "." + i
```
picnixz: Use f-strings instead.

sphinx/ext/autosummary/generate.py
```diff
@@ -320,11 +320,34 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
...
+                    if i[0] != "_":
```
picnixz: I would use `startswith` instead. Also, what about special methods or those named `__foo__` for instance? because they would be excluded although they could be public.

picnixz (commented):
- You'll need a CHANGES entry when the PR is ready.
- You'll need to add tests.

doc/usage/extensions/autosummary.rst
satr-cowi: Have put 7.3.0 for now; Will add a comment below to check this before merging

doc/usage/extensions/autosummary.rst
satr-cowi: Have adopeted qualnames

doc/usage/extensions/autosummary.rst
satr-cowi: Have added in a simple example to help with the various parameters

sphinx/ext/autosummary/generate.py
satr-cowi: Done

sphinx/ext/autosummary/generate.py
satr-cowi: Have removed this line, I now use the fact that "methods" and "attributes" are already filtered. "qualnames" now includes all the private/special methods too.

satr-cowi:
Hi @picnixz, thanks for your help, the comments were really useful. I've made an example that I used in the updated documentation and can turn into a unit test but want to make sure you are happy with the following first:

I think that `inherited_methods` and `inherited_attributes` should reflect the same as `methods` and `attributes` by only returning "puplic" methods/attributes for consistency. We could add in `inherited_all_methods` and `inherited_all_attributes` too pretty simply, but I'm not sure how big the need is, with the fact that `all_methods` isn't currently documented (a user can always just filter qualnames themselves using these too)?

The __dict__ part has been directly taken from the current implementation of `inherited_members`, so therefore this will not be a breaking change in any way (as this bit wouldn't currently work for classes without a __dict__ anyway). I couldn't figure out how to use _get_members() to only get the local members of a particular class, so have left it how it is for now. Feel free to update if you know a better way but my knowledge of Python inheritance isn't that deep!

picnixz:
I won't have time this week for any in depth review (especially on autodoc-related topics) so it needs to wait until next week or the week after.

satr-cowi:
Hi @picnixz, I believe this is ready to go, I've taken on board the feedback and tried to make better use of get_class_members to collect the methods and attributes.

I added a unit test to one of the existing setups, there is also a simple example in the documentation showing the differences between the various variables that can be accessed.

picnixz:
Hi, I am sorry  ut I don't have time to do anything these weeks. Also I think Adam is also very busy and it doesn't seem to me that we'll push 7.3 before some time. At least the repo is stalled for the past two weeks. 

Don't worry about your PR being forgotten since I still watch the repo. However, for the moment being, I won't be able to review your PR :/

satr-cowi:
Ok, no worries!

satr-cowi:
Hi, I'm aware that the versions/changelog etc. need updating, but do you think we could try and get this in a release at some point?

sphinx/ext/autosummary/generate.py
picnixz: @satr-cowi Have you addressed this issue?

tests/test_extensions/test_ext_autosummary.py
```diff
@@ -356,6 +356,49 @@ def test_autosummary_generate_content_for_module_imported_members_inherited_modu
     assert context['objtype'] == 'module'
 
 
+@pytest.mark.sphinx(testroot='ext-autosummary')
+def test_autosummary_generate_content_for_module_imported_members_inherited_class(app):
+    import autosummary_dummy_inherited_module
+    template = Mock()
+
+    generate_autosummary_content('autosummary_dummy_inherited_module.InheritedAttrClass',
```
picnixz: For this one, I would like to have a much wider test coverage. More precisely, I want to be able to test the behaviour of private names, mangled names, inherited methods that, static methods. For that, I would usggest creating a separate dummy module with multiple classes inside. I also want to see how multiple inheritance is handled and how shadowing names are done.

Most of the time, our tests cover simple cases which are fine but everytime there is an issue with autodoc, it's always because of some weird constructions. So I'd like to at least cover non-trivial cases (without trying really pathological cases).

tests/test_extensions/test_ext_autosummary.py
```diff
@@ -356,6 +356,49 @@ def test_autosummary_generate_content_for_module_imported_members_inherited_modu
...
+    assert context['inherited_methods'] == ['autosummary_dummy_module.Foo.bar']
+    assert context['inherited_attributes'] == ['autosummary_dummy_module.Foo.CONSTANT3',
+                                               'autosummary_dummy_module.Foo.CONSTANT4',
+                                               'autosummary_dummy_module.Foo.baz',
+                                               'autosummary_dummy_module.Foo.value']
+    assert context['fullname'] == 'autosummary_dummy_inherited_module.InheritedAttrClass'
+    assert context['module'] == 'autosummary_dummy_inherited_module'
+    assert context['objname'] == 'InheritedAttrClass'
+    assert context['name'] == 'InheritedAttrClass'
+    assert context['objtype'] == 'class'
```
picnixz, line 596: Fully qualified names should also be tested (I don't see them but maybe I missed them?)

picnixz:
> do you think we could try and get this in a release at some point?

We can definitely make it in a release because this is something I think that is worthwhile. Many people (including me) asked for having a finer control on autosummary. As for "when", I don't know.

sphinx/ext/autosummary/generate.py
satr-cowi: I think my preference is for ns['inherited_methods'] would be to only return the public methods (and therefore be consistent with ns['methods'])

We could introduce ns['all_inherited_methods'] as well, but the documentation doesn't include anything about ns['all_methods'] at the moment so we should probably include that too if we do this.

tests/test_extensions/test_ext_autosummary.py
satr-cowi: The full qualified name of this class is included as 'fullname', and the fully qualified name of the inherited members are in 'inherited_qualnames' (lines 383-389)

tests/test_extensions/test_ext_autosummary.py
satr-cowi: Sure thing, I might need to lean a bit about these more niche cases but happy to give it a first go!

sphinx/ext/autosummary/generate.py
picnixz: Personally, I want us to be consistent without breaking previous changes.

AFAIU the current behaviour is "ns[C] returns public members for category C" and "ns[all_C] returns all members for category C (including private ones)". I don't mind not documenting "all_*" for now, but I would like the documentation for ns["members"] & co. to be updated then. Currently it says (for "members"):

> List containing names of all members of the module or class. Only available
  for modules and classes.

In addition, your "inherited_qualnames" documentation tells me about: 

>    * ``members`` returns a list with many built in methods/attributes and
     ``['__init__', '_i_am_private', 'do_something', 'do_something_else',
     'foo_attr']``

So now your members also include private members.

doc/usage/extensions/autosummary.rst
```diff
@@ -293,16 +293,79 @@ The following variables are available in the templates:
...
+              print("Foo something")
+
+          def _i_am_private(self):
+              print("I'm a private method")
+
+      class bar(foo):
+          def do_something_else(self):
+              print("Bar something")
+
+   Some available parameters for the autosummary of class ``bar`` in the above
```
picnixz, line 382: Also add an example for how to use it in a template.

tests/test_extensions/test_ext_autosummary.py
picnixz: Oh my bad ! it didn't see it. Ideally I would prefer an exact match because that way you know whether we correctly picked up inherited members and how shadowed names were done, but I assume that there are too many things in the list.

satr-cowi: I have tried including all the members in the new unit test, but you then get failures depending on the varying builtin methods included in different versions of Python (e.g. passes for 3.11 and 3.12, but fails on 3.10 and 3.13 for different reasons). Any thoughts on if there is a better way to deal with this rather than what I did above (i.e. don't worry about these additional items in the test)?

tests/test_extensions/test_ext_autosummary.py
satr-cowi: Have made a module where I have tried to test some of these features. Let me know if there is anything specific/additional that you think is missing.

sphinx/ext/autosummary/generate.py
satr-cowi: Current behaviour (as in the example) is:

- methods and attributes returns just public methods/attributes
- members and inherited members return all members including private
- all_methods and all_attributes returns all methods/attributes including private, but these are not documented so most users won't know they exist

I've updated the "members" documentation to make it clear that this and "inherited members" do include private members. It was already clear that methods/attributes were only public.

The proposed additions are just that:

- inherited_methods and inherited_attributes returns just public inherited methods/attributes
- inherited_qualnames is the same as inherited_members (includes private), but just with the fully qualified names

Think this is pretty clear from the documentation now

doc/usage/extensions/autosummary.rst
satr-cowi: Have added an example

satr-cowi:
Hi @picnixz, finally got round to making the requested updates. See my comment above on the failing tests, and do let me know your thoughts on if there is a better way to test including the varying in-built members too for each version of Python.

doc/usage/extensions/autosummary.rst
```diff
@@ -293,16 +293,106 @@ The following variables are available in the templates:
...
 .. data:: inherited_members
 
-   List containing names of all inherited members of class.  Only available for
-   classes.
+   List containing names of all inherited members of class including private ones.
+   Only available for classes.
 
    .. versionadded:: 1.8.0
 
+.. data:: inherited_qualnames
```
picnixz, line 356: This should specify whether this includes public or private attributes.

doc/usage/extensions/autosummary.rst
```diff
@@ -293,16 +293,106 @@ The following variables are available in the templates:
 
 .. data:: members
 
-   List containing names of all members of the module or class.  Only available
-   for modules and classes.
+   List containing names of all members of the module or class including private
```
picnixz, line 346: Just reconfirm, but this was already the case that it returned private attributes right? since we are doing `members = dir(self)`.

picnixz (commented):
Ok, after having digested this PR I have one design concern. Currently `members` returns all attributes, whether they are private or not. In other words, I'd expect `inherited_XXX` to return all XXX whether they are private or not (because of the naming). I know that we already discussed the point with the methods not returning everything, but I honestly think we need to make the smallest breaking change as well as the simplest one and the most usable one.

For now, **do not change your implementation** (I'm just asking around).

In order to only have the public ones, I think it's better to have `public_inherited_XXX` and `inherited_XXX` where the former returns public lookalike attributes while the latter returns all XXX attributes, whether they are private or not.

What do you think? I'm sorry for my late review but I've been working more on CPython these past months, hence limiting my contribution to Sphinx itself.

---

cc @AA-Turner @chrisjsewell

picnixz:
For instance, since we are *adding* new features, we could make name them more precisely, namely `inherited_public_methods` instead of `inherited_methods`. Or we can continue doing what we were doing with `methods` and go (in the future) for the `all_inherited_methods` route.

I like having the `public` in the finer-grained feature because you don't forget about it. On the other hand, not having the `all` is not necessarily bothersome.

doc/usage/extensions/autosummary.rst
satr-cowi: Yes, the existing attributes members and private members already included private ones. Therefore there shouldn't be any change here, just clarifying this in the docs.

doc/usage/extensions/autosummary.rst
satr-cowi: Agreed, will clarify in the description that this does include private members (meaning methods and attributes)

satr-cowi:
> For instance, since we are _adding_ new features, we could make name them more precisely, namely `inherited_public_methods` instead of `inherited_methods`. Or we can continue doing what we were doing with `methods` and go (in the future) for the `all_inherited_methods` route.
> 
> I like having the `public` in the finer-grained feature because you don't forget about it. On the other hand, not having the `all` is not necessarily bothersome.

I also like the clarity of stating 'public', but would like to keep consistency. Ideally we would also update 'methods' to be 'public_methods' too. Could perhaps add this as an equivalent to 'methods', but leave in the original so it is not breaking? Can state it as deprecated in the docs?

If we are doing this, perhaps we should start to document 'all_methods'/'all_inherited_methods' etc. too?

> Finally, can you fix the tests as well please? thank you in advance!

Might not have commented properly on this one, but I did have a question here. Different Python versions have different in built methods. You asked above to include these in the tests, rather than just checking that the expected values are in the result. 

Do you have a good recommendation for testing against each of the expected inbuilt methods for different python versions?

picnixz:
> Do you have a good recommendation for testing against each of the expected inbuilt methods for different python versions?

Somehting like this:

```py
if sys.version_info[:2] >= (3, 14):
	expected_builtins = [...]
elif sys.version_info[:2] >= (3, 13):
	expected_builtins = [...]
elif sys.version_info[:2] >= (3, 12):
	expected_builtins = [...]
```

possibly with finer-grained checks depending on whether a built-in has been added or not (up to you for creating those ifs so that it's the most readable and simple)

tests/test_extensions/test_ext_autosummary.py
satr-cowi: Have sorted using if statements depending on Python version as recommended. Implemented in a way where only small tweaks should be needed for future versions.

satr-cowi:
Tests are sorted to include the expected builtins for multiple Python versions. 

Let me know once we have a conclusion on how we want to implement the naming of public/private members and can change as necessary. 

Will need to update the version numbers I've put above, and add to the changelog, but can do this when ready to merge.

sphinx/ext/autosummary/generate.py
```diff
@@ -398,9 +427,12 @@ def _skip_member(app: Sphinx, obj: Any, name: str, objtype: str) -> bool:
         return False
 
 
-def _get_class_members(obj: Any) -> dict[str, Any]:
+def _get_class_members(obj: Any, return_object: bool = True) -> dict[str, Any]:
```
picnixz: I would actually flip the boolean and use `return_types=False` by default. What do you think?

And whether we change the boolean or not, I think it would be good to make it keyword-only just so that we know what we do when we call the function (we don't need to go to its implementation).

tests/roots/test-ext-autosummary/autosummary_dummy_complex_inheritance_module.py
```diff
@@ -0,0 +1,112 @@
...
+        self.name = "Andrew"
+        self.age = 35
+
+    def get_name(self):
+        return self.name
+
+    def get_age(self):
+        return f"Parent class -  {self.age}"
+
+    # Test mangled name behaviour
```
picnixz, line 21: Can we have a normal private attribute/method (no mangled, just simple underscore)?

tests/roots/test-ext-autosummary/autosummary_dummy_complex_inheritance_module.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self):
+        self.name = "Jerry"
+        self.age = 25
+
+
+__all__ = ["Parent", "Child", "Baby", "Job", "Architect", "Jerry"]
+
+# The following is used to process the expected builtin members for different
+# versions of Python. The base list is for v3.11 (the latest testing when added)
+# and new builtin attributes/methods in later versions can be appended below.
```
picnixz: Can we have those values declared in the test instead? the test roots should be only contains the code we "test" (so this file should be like a real one on which you run autodoc).

tests/test_extensions/test_ext_autosummary.py
```diff
@@ -488,6 +488,376 @@ def test_autosummary_generate_content_for_module_imported_members_inherited_modu
...
+        '__lt__',
+        '__ne__',
+        '__new__',
+        '__reduce__',
+        '__reduce_ex__',
+        '__repr__',
+        '__setattr__',
+        '__sizeof__',
+        '__str__',
+        '__subclasshook__',
```
picnixz: Aren't those methods inherited from `object`? can't they be automatically detected like this?

tests/test_extensions/test_ext_autosummary.py
```diff
@@ -488,6 +488,376 @@ def test_autosummary_generate_content_for_module_imported_members_inherited_modu
...
+        'builtins.object.__ne__',
+        'builtins.object.__new__',
+        'builtins.object.__reduce__',
+        'builtins.object.__reduce_ex__',
+        'builtins.object.__repr__',
+        'builtins.object.__setattr__',
+        'builtins.object.__sizeof__',
+        'builtins.object.__str__',
+        'builtins.object.__subclasshook__',
+    ]
```
picnixz: Can't we generate the `builtins.object.*` methods using `[f'builtins.object.{name}' for name in dir(object)]`?

tests/roots/test-ext-autosummary/autosummary_dummy_complex_inheritance_module.py
satr-cowi: Have added

tests/test_extensions/test_ext_autosummary.py
satr-cowi: Couldn't use object directly as it didn't have some of the expected attributes/methods, but a using a empty class did the trick.

tests/test_extensions/test_ext_autosummary.py
satr-cowi: Similarly, have simplified this into a generated list with some tweaks where necessary

tests/roots/test-ext-autosummary/autosummary_dummy_complex_inheritance_module.py
satr-cowi: Moved to test function

sphinx/ext/autosummary/generate.py
satr-cowi: No longer relevant, a different function is used instead as there is a new implementation of get class members

satr-cowi:
Hi @picnixz, had some time to update this and address your comments.

All the tests pass, and the expected results have been simplified a bit (the only version specific stuff added manually was to do with annotation where it was a bit harder to do manually)

satr-cowi:
Hi @picnixz, have just used this for my package again and thought I'd take that as a chance to follow up and backmerge.

The failing CI runs seem to be unrelated to these changes (although not fully sure about this ty one) 

The version numbers will need updating (and changelog?) but didn't want to act on this until happy that there aren't further changes needed. 

Thanks

Issue #11715, opened 2023-10-09
tls_verify and tls_cacerts config options are not applied to requests in ImageDownloader


ntouran:
### Describe the bug

The `tls_verify` and `tls_cacerts` configuration options are not applied to the `requests` call involved in the `ImageDownloader`. Thus, any configuration you set with those settings does not apply as expected during a build that involves downloading images. An example is if you're doing a `make latex` for source that refers to external images. 

I've identified that the problem in the code is here: https://github.com/sphinx-doc/sphinx/blob/bb74aec2b6aa1179868d83134013450c9ff9d4d6/sphinx/transforms/post_transforms/images.py#L78

c.f. where they are applied here:

https://github.com/sphinx-doc/sphinx/blob/bb74aec2b6aa1179868d83134013450c9ff9d4d6/sphinx/ext/intersphinx.py#L138-L140

I have a pull request in the works that fixes this. 

This is, in essence, a needed follow-up to https://github.com/sphinx-doc/sphinx/pull/3156.

### How to Reproduce

* set `tls_verify = False` in `conf.py`
* Self-sign a certificate to a server
* Refer to an image on that server by https:// URL in ReST
* run `make latex`

Observe error: 

```WARNING: Could not fetch remote image: https://self-signed.exampleserver.tld/user/174/files/978909e4-d1eb-422e-8be2-daf9c2fa9280 [HTTPSConnectionPool(host='self-signed.exampleserver.tld', port=443): Max retries exceeded with url: /user/174/files/978909e4-d1eb-422e-8be2-daf9c2fa9280 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1129)')))]```

### Environment Information

```text
Platform:              win32; (Windows-10-10.0.19042-SP0)
Python version:        3.11.4 (tags/v3.11.4:d2340ef, Jun  7 2023, 05:45:37) [MSC v.1934 64 bit (AMD64)])
Python implementation: CPython
Sphinx version:        7.3.0+/bb74aec2b
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Additional context

My particular use case where this is coming up is that I have some `.md` files with image Markdown in them pointing to external URLs, and when those come via `myst_parser` into sphinx and out to LaTeX, I'm seeing unexpected certificate errors.

PR #11716, opened 2023-10-09
Apply tls_verify and tls_cacerts config to ImageDownloader (#11715)

ntouran:
Subject: Apply tls_verify and tls_cacerts config to ImageDownloader

### Feature or Bugfix

- Bugfix

### Purpose
- Apply existing requests configuration to a request where it is needed in some cases.
- To fully test, you need a self-signed certificate to put remote images on via URL.

### Detail
- Fixes the issue defined in #11715

### Relates
- #11715


picnixz:
If possible, can you check where other `tls_verify` and `tls_cacerets` values are not applied? Like, are there other places where the same issue could happen?

ntouran:
> If possible, can you check where other `tls_verify` and `tls_cacerets` values are not applied? Like, are there other places where the same issue could happen?

I looked around to see if there were any more obvious cases at first but will double check. Good idea. 

Ok:

```bash
$ git grep 'from sphinx.util import' | grep requests
sphinx/builders/linkcheck.py:from sphinx.util import encode_uri, logging, requests
sphinx/ext/intersphinx.py:from sphinx.util import logging, requests
sphinx/transforms/post_transforms/images.py:from sphinx.util import logging, requests
tests/test_build_linkcheck.py:from sphinx.util import requests
```

The 3 clients of the requests util here are:

- [x] linkcheck   -- yes, does use this config correctly in HEAD and GET requests
- [x] intersphinx -- yes, does use proper config already
- [x] ImageDownloader -- now does correct usage applied by this PR

So I think we're good!

picnixz:
Good then! Could you add a CHANGES entry please? 

@AA-Turner I think we're good unless you want to wrap small PRs for the day when 7.3.0 is planned to be released.

ntouran:
> Could you add a CHANGES entry please?

Done!

picnixz:
Perfect! In the future, if you want to contribute again, don't force push commits since it makes incremental review harder (we squash them at the end anyway).

ntouran:
> Perfect! In the future, if you want to contribute again, don't force push commits since it makes incremental review harder (we squash them at the end anyway).

Ok sorry. The 2nd commit on the followup had the wrong username so I only force pushed an update to the username/email to that one, but I see what you're saying, and can understand that.

PR #11719, opened 2023-10-11
[low priority] cleanup: resolve lint warnings reported by ruff

jayaddison:
### Feature or Bugfix
- Linter fixups

### Purpose
- Reduce noise from GitHub beta feature that displays "unchanged files with check annotations" on "Files changed" pull request tab.
- No functional changes intended or desired.

image

### Relates
N/A


sphinx/testing/fixtures.py
```diff
@@ -211,7 +211,7 @@ def if_graphviz_found(app: SphinxTestApp) -> None:  # NoQA: PT004
     graphviz_dot = getattr(app.config, 'graphviz_dot', '')
     try:
         if graphviz_dot:
-            subprocess.run([graphviz_dot, '-V'], capture_output=True)  # show version
+            subprocess.run([graphviz_dot, '-V'], capture_output=True, check=False)
```
picnixz, line 215: Maybe you could keep the comment? (like, put it on the line above)

sphinx/testing/path.py
```diff
@@ -150,7 +150,7 @@ def utime(self, arg: Any) -> None:
         os.utime(self, arg)
 
     def open(self, mode: str = 'r', **kwargs: Any) -> IO:
-        return open(self, mode, **kwargs)
+        return open(self, mode, **kwargs)  # NoQA: SIM115
```
picnixz, line 153: Do we use "NoQA" or "noqa"?

tests/test_pycode_ast.py
```diff
@@ -34,7 +34,6 @@
     ("a % b", "a % b"),                         # Mod
     ("a * b", "a * b"),                         # Mult
     ("sys", "sys"),                             # Name, NameConstant
-    ("1234", "1234"),                           # Num
```
picnixz, line 37: You should add "Num, Constant" undeer the entry of the first ("1234", "1234") so that we know which AST nodes we are testing.

sphinx/testing/path.py
jayaddison: Both at the moment, although `NoQA` appears more frequently in the codebase.  Both have been added in newlines in recent history, so I don't think there's a clear preference emerging.

I've seen `noqa` more often in the past, but `NoQA` does seem slightly clearer in terms of communicating that it's a directive to disable certain quality assurance checks.

```sh
$ grep -rw noqa sphinx tests | wc -l
64
$ grep -rw NoQA sphinx tests | wc -l
84
```

(I used `git log --patch` to try to find recent instances of adding either)

picnixz: I see. Well, I personally prefer noqa because I don't need to hold my SHIFT key :smile: Anyway, I think it's up to you then

picnixz (approved):
Looks good to me.

jayaddison:
Thank you, @picnixz

Issue #11448, opened 2023-06-01
Restructuring multiprocessing


danwos:
**Is your feature request related to a problem? Please describe.**
The current implementation of the multiprocessing architecture has some problems, especially with bigger projects (500 - 25.000 pages):.

* The memory consumption can be extremely high and may need 32 GB or more
* The created processes are not running in parallel for some time, which may cost ~30% of build time

**I would like to work on this**, but need some feedback about concepts and ideas, as the needed changes will affect some part of the architecture.

From some comments and older issues it looks like the multiprocessing part was mainly created to deal with IO waiting times and make the writing more efficient.
It looks like parallelizing the computation itself was not one of the main goals.

The reasons for the above problems are:

### 1. Processes are created to often
Sphinx is calculating chunks based on the number of documents and cores/`-j X`.

Each chunk gets its own process, which is started if a prio-process is done and terminated (simply said).
The amount of chunks is calculated in a way, that for bigger projects you get much more chunks as you have configured via `-j`.
But this also means, that several new processes are "forked"  over the time and get a copy of the raising ENV of the main processes. Process creation costs time.

**Solution idea**: The amount of chunks should match the amount of `-j X`, so that we create long running processes only once (Drawback: The implemented log-collection needs an update, as collecting the logs when a process is done leads to no output to the user for several minutes.)

### 2. Processes are "forked", not "spawned"
Each forked process gets a copy of the main process, which contains the complete ENV including all document information from already calculated docs.

This means if the main process is using 4 GB of RAM and you are working on an 8 core system (`-j 8`), Sphinx will create 7 parallel processes and all of them get a copy of the 4 GB => 4GB + 7*4 GB = 32 GB free RAM needed.
The only solution is to reduce the amount of cores e.g. `-j 4`, but that may cost you up to 50% of build performance.

**Solution idea**:  Less process creation. And if possible "spawing" instead of "forking".
"Spawing" will not make a memory copy of the main process, so data needs to be given to its child-processes via a pipe.
This would be a huge conceptional change.
I also found a PR, where "forking" was added for Mac OS X support.

### 3. Serialized vs Parallel tasks
Before Sphinx starts the parallel tasks, it calculates serial tasks in the main process. 
The serial calculation of a document is taking often longer as the parallel calculation part for the document.

Also, a process gets started after Sphinx has done all the serial calculations for a number of docs (a chunk.)
[...]

PR #11733, opened 2023-10-23
Parallel post transform

ubmarco:
Subject: Enable parallel post transformations including ``doctree-resolved``

### Feature or Bugfix
- Feature

### Purpose
The current parallel writing approach is quite slow for compute intensive extensions in `doctree-resolved`.
The event is part of the post transformation call chain. This PR shifts that workload to the parallel subprocesses.

### Detail
For extensions like [Sphinx-Needs](https://sphinx-needs.readthedocs.io) this speeds up the build time by a factor
of 2.5 to 4 for big projects. The new implementation also leads to more processes running in parallel as the serial post transformation of chunks blocks their execution, so a bottleneck gets resolved.
**The feature is enabled by default.**
The default-enabling of the feature is necessary as Sphinx` direct dependencies
```
sphinxcontrib-applehelp
sphinxcontrib-devhelp
sphinxcontrib-jsmath
sphinxcontrib-htmlhelp>=2.0.0
sphinxcontrib-serializinghtml>=1.1.9
sphinxcontrib-qthelp
alabaster>=0.7,<0.8
```
are also extensions and don't support the new feature yet. They would block the availability.
Other extensions that cannot live with the new implementation, can disable it using a new flag in the `setup()` return value called `parallel_post_transform_safe`. I already activated the flag for all Sphinx internal extensions.

Moving post-transformation into subprocesses can lead to problems with builders or extensions that store data on the builder or the environment. For the builder a merge feature was added that allows builders to define a method to merge certain builder attributes. The implementation works like `merge_info_from` after the parallel reading phase.
Within Sphinx, only the `CheckExternalLinksBuilder` seems to be affected due to the `hyperlinks` attribute that is consumed when the build finishes. This is fixed in this PR.

### Relates
- https://github.com/sphinx-doc/sphinx/issues/11448

### DIFF
--- a/doc/extdev/index.rst
+++ b/doc/extdev/index.rst
@@ -191,6 +191,25 @@ as metadata of the extension.  Metadata keys currently recognized are:
             * The core logic of the extension is parallelly executable during
               the writing phase.
 
+* ``'parallel_post_transform_safe'``: a boolean that specifies if parallel
+  post-processing of doctrees can be used. This is only relevant in case
+  ``parallel_write_safe`` is also active. Post transformation also contains
+  the ``doctree-resolved`` event. This can greatly improve the performance
+  for extensions that do compute intensive tasks in that event.
+  Since extensions usually don't negatively influence the process, this
+  defaults to ``True``. Technically, the post transform and
+  ``doctree-resolved`` is shifted to the parallel writing worker.
+
+  .. note:: The *parallel-post-transform-safe* extension must satisfy the
+            following conditions:
+
+            * The core logic of the extension is parallelly executable during
+              the writing phase (precondition from ``parallel_write_safe``).
+            * The extension does not write data to the environment during
+              post-transform or ``doctree-resolved`` that is expected later
+              in the build to be available, e.g. for the ``build-finished``
+              event. The environment of each subprocess will be discarded
+              after finish writing.
 
 APIs used for writing extensions
 --------------------------------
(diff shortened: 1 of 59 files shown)


twodrops:
We need this badly in a project with more than 80k rST files and 1000+ users depending on this!

twodrops:
The current parallel processing lib used by Sphinx creates parallel processes, but cores keep waiting for all processes to finish until the next set of processes are allocated. In a way, this is somehow not really multi-processing making use of all cores. Cores simply keep waiting.

Is this issue fixed as part of this PR?

danwos:
No, the way Sphinx is dealing with multi-processing during `parallel write` is not touched by this PR, but another PR is under discussion for this.
However, the benefit of executing `post-transform` in a parallel way should bring more performance instead of optimizing the multiprocessing itself. Also, the current PR needs fewer code changes.

Ohh and as the `post-transform` functions got moved out of the `serial execution`, the problems with the bottleneck "blocking serial execution during multiprocessing" should be much less dramatic.

ubmarco:
Test approach for this PR:
In `sphinx/testing/util.py` -> `SphinxTestApp` -> `def __init__()` set the default parameter `parallel` from `0` to `4`.
That way all test cases, that don't specify parallel execution are running in parallel.
This is needed to effectively find issues with the new parallelization implementation.
This revealed that, on master, 3 test cases fail when executed in parallel. Those still fail, but should be investigated in a future PR. Without the modification, all tests are ok.

ubmarco:
Closing in favor of https://github.com/sphinx-doc/sphinx/pull/11746

PR #11736, opened 2023-10-24
Used sys.version_info for Python version checks.

smithdc1:
Subject: Improve type hints for imports dependent upon Python version. 

### Feature or Bugfix
- Refactoring

### Purpose
Mypy understands the use of sys.version_info within if/elif/else statements. See https://mypy.readthedocs.io/en/stable/common_issues.html#python-version-and-system-platform-checks

This allows for the removal of some # type: ignore comments.


AA-Turner:
Thanks David!

A

Issue #10779, opened 2022-08-19
Oprimize writing step of parallel builds


ezio-melotti:
**Is your feature request related to a problem? Please describe.**
During the writing step of the build process, even while using `-j auto`, there are some time-consuming operations that are performed serially.  It might be possible to parallelize them and further improve performances.

**Describe the solution you'd like**
Last night I went down a rabbit hole and tried to optimize `_write_parallel`:
https://github.com/sphinx-doc/sphinx/blob/8c4865c30d5fa847d727fea16519d7afce627932/sphinx/builders/__init__.py#L571-L607

I tried to naively apply the following changes:
```diff
diff --git a/sphinx/builders/__init__.py b/sphinx/builders/__init__.py
index 2aede5c24..d27df5fdd 100644
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -569,9 +569,11 @@ class Builder:
                 self.write_doc(docname, doctree)
 
     def _write_parallel(self, docnames: Sequence[str], nproc: int) -> None:
-        def write_process(docs: List[Tuple[str, nodes.document]]) -> None:
-            self.app.phase = BuildPhase.WRITING
-            for docname, doctree in docs:
+        def write_process(docs: List[str]) -> None:
+            for docname in docs:
+                doctree = self.env.get_and_resolve_doctree(docname, self)
+                self.app.phase = BuildPhase.WRITING
+                self.write_doc_serialized(docname, doctree)
                 self.write_doc(docname, doctree)
 
         # warm up caches/compile templates using the first document
@@ -595,12 +597,7 @@ class Builder:
 
         self.app.phase = BuildPhase.RESOLVING
         for chunk in chunks:
-            arg = []
-            for docname in chunk:
-                doctree = self.env.get_and_resolve_doctree(docname, self)
-                self.write_doc_serialized(docname, doctree)
-                arg.append((docname, doctree))
-            tasks.add_task(write_process, arg, on_chunk_done)
+            tasks.add_task(write_process, chunk, on_chunk_done)
 
         # make sure all threads have finished
         tasks.join()
```
This resulted in a build time reduction of ~25%.

Without patch, using Sphinx v4.5.0 to build the python/cpython docs:
```
real    2m13,386s
user    4m35,046s
sys     0m7,205s
```
With patch:
```
real    1m36,804s
user    4m42,314s
sys     0m5,909s
```

---

[...]

Issue #2946, opened 2016-09-10
Race conditions when running multiple Sphinx instances using the same doctree


mgorny:
Projects using Sphinx (and Makefiles) often define multiple targets for different doc builders. For example, the LLVM project defines a target for manpages and for HTML docs. Since those targets are independent, make is allowed to run them in parallel (and does that). Since they use the same source tree, they use the same doctree root as well.

Lately, I've noticed that the two parallel Sphinx instances 'fight' over the doctree. Basically, it seems that they both ignore each other and create the doctrees independently, i.e. the same files are first written by one Sphinx instance, and afterwards overwritten by the second. While normally this doesn't cause problems (except for being awfully inefficient), I've actually seen a race condition causing major failure (sorry, don't have any logs for it). Basically, one Sphinx instance has opened one of the files for reading while the other was writing it.

We've attempted to find a good solution on the build system side but could fine none satisfactory. If we switched to separate doctrees, that would be a waste of a good caching opportunity. Getting two independent targets serialized is not trivial in CMake — and LLVM people aren't happy with me adding fake dependencies to enforce serialization. But even if we did that, there's still the general problem that other people will be unaware of the issue and will be hitting it in the future.

Therefore, I think it would be best to fix it on Sphinx side. The simplest solution would be to lock some file in the doctree directory. Considering the specific design, the environment pickle should be the best file to lock. However, that would require redesigning the application to keep the file open rather than closing it immediately after reading/writing (which would also be beneficial to avoiding race conditions).

mgorny:
Well, I've tried to give it a shot… and I can't find a sane way of doing it with the current design.

The big problem is, the use of pickles is done quite inconsistently. The pickled environment is first read in `Sphinx._init_env()` which is called by `Sphinx.__init__()`, i.e. pretty early. However, it is not updated until `Builder.build()`, and there it is pickled afterwards. So to get the results I need, I would have to lock the environment from `Sphinx._init_env()` to `Builder.build()`.

[...]

PR #11746, opened 2023-11-03
Parallel post transform including write_doc_serialized

ubmarco:
Subject: Enable parallel post transformations including `doctree-resolved` **and** `write_doc_serialized`.

This is an extended version of PR https://github.com/sphinx-doc/sphinx/pull/11733.
It has basically the same purpose. Sorry for opening a second PR, however I need to run the CI and also want to ask for maintainers' opinion on the work. Maybe there is a show stopper that I do not see yet.

### Feature or Bugfix
- Feature

### Purpose
The current parallel writing approach is quite slow for compute intensive extensions in `doctree-resolved`.
The event is part of the post transformation call chain. This PR shifts that workload to the parallel subprocesses.
To correctly deal with updated doctrees in post-transform, also the step `write_doc_serialized` runs in parallel.

### Detail
For extensions like [Sphinx-Needs](https://sphinx-needs.readthedocs.io) this speeds up the build time by at least a factor of 2 for big projects. The new implementation also leads to more processes running in parallel as the serial post transformation of chunks in the main process blocks their parallel execution, so a bottleneck gets resolved.
All processes defined by `--jobs` (minus 1, for the main process) now start at once when writing parallel.

The main difference between https://github.com/sphinx-doc/sphinx/pull/11733 and this PR is how `write_doc_serialized` is called in case all extensions support parallel post transforming. Here is a summary for `master` branch, the other PR and this PR:

- `master`

  1. main process: post-transform including doctree-resolved
  2. main process: write_doc_serialized
  3. sub processes: write_doc

- https://github.com/sphinx-doc/sphinx/pull/11733

  1. main process: write_doc_serialized
  2. sub process: post-transform including doctree-resolved
  3. sub process: write_doc

  After my investigation, I now think this approach is broken as it changes the logic. write_doc_serialized does 2 things for the html builder: image handling and building the search index. Both need information generated in post-transform.

- this PR

  1. sub process: post-transform including doctree-resolved
  2. sub process: write_doc_serialized
  3. sub processes: write_doc
  4. sub process: return relevant information from write_doc_serialized
  5. main process: merge back relevant information from write_doc_serialized

  I think this is the correct approach. The function name write_doc_serialized is not optimal anymore, as it may run in parallel. On the long run I think Sphinx should change the contract with extensions so all of them *must* support parallel reading and writing.

All Sphinx test cases are green on my local machine. When activating parallel builds for **all** test cases, only 2 fail:
```
FAILED tests/test_domain_std.py::test_productionlist - AssertionError: assert 1 == 2
FAILED tests/test_ext_inheritance_diagram.py::test_inheritance_diagram - KeyError: 'basic_diagram'
```
But they also fail on `master`, so I think the problems existed before.
I also ran all test cases of a new [Sphinx-Needs branch](https://github.com/useblocks/sphinx-needs/tree/mh-sphinx-7.2.5-agg) that supports the latest Sphinx version against this PR. All are successful, also when run in parallel mode.

**Performance**
In terms of performance, this PR improves build times for bigger Sphinx projects that have expensive `doctree-resolved` functions significantly by at least a factor of 2.
For projects that have cheap `doctree-resolved` functions, the performance maybe a bit lower than `master` branch, due to back-merging of artifacts of `write_doc_serialized`. I measured a runtime increase of 2% for a 1000 pages project.

**Memory**
I haven't done a larger analysis besides my local XFCE task manager. Due to the fact that more processes run in parallel at a certain point in time, the memory consumption with the same core count can be higher than for the `master` branch.

**Risk evaluation**
The new parallel post-transformation is activated using a new global configuration variable `enable_parallel_post_transform` which is `False` by default. The feature is also marked as experimental.
Only projects/builds that know all extensions support parallel post transformation for all extensions should activate it.
As the behavior for `enable_parallel_post_transform=False` is not changed, the risk of this PR is low.

**Conclusion**
As a conclusion, I think this PR is a candidate for merging as it behaves correctly in terms of application logic while https://github.com/sphinx-doc/sphinx/pull/11733 should be closed unmerged after confirmation of my assessment.

Any feedback is very welcome!

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -79,11 +79,10 @@ class Builder:
     #: post transformation, to be merged to the main builder in
     #: :py:class:`~sphinx.builders.Builder.merge_builder_post_transform`.
     #: Attributes in the list must be pickleable.
-    #: The approach improves performance when
-    #: pickling and sending data over pipes because only a
-    #: subset of the builder attributes are commonly needed for merging
-    #: to the main process builder instance.
-    post_transform_merge_attr: list[str] = []
+    #: The approach improves performance when pickling and sending data
+    #: over pipes because only a subset of the builder attributes
+    #: are commonly needed for merging to the main process builder instance.
+    post_transform_merge_attr: tuple[str, ...] = ()
 
     def __init__(self, app: Sphinx, env: BuildEnvironment) -> None:
         self.srcdir = app.srcdir
@@ -135,19 +134,21 @@ class Builder:
         pass
 
     def merge_builder_post_transform(self, new_attrs: dict[str, Any]) -> None:
-        """Give builders the option to merge any post-transform information
-        coming from a parallel sub-process back to the main process builder.
-        This can be useful for extensions that consume that information
-        in the build-finish phase.
+        """Merge post-transform data into the master builder.
+
+        This method allows builders to merge any post-transform information
+        coming from parallel subprocesses back into the builder in
+        the main process. This can be useful for extensions that consume
+        that information in the build-finish phase.
         The function is called once for each finished subprocess.
-        Builders that implement this function must also define
-        the class attribute
-        :py:attr:`~sphinx.builders.Builder.post_transform_merge_attr` as it defines
-        which builder attributes shall be returned to the main process for merging.
+        Builders that implement this function must also define the class
+        attribute :py:attr:`~sphinx.builders.Builder.post_transform_merge_attr`
+        as it defines which builder attributes shall be returned to
+        the main process for merging.
 
         The default implementation does nothing.
 
-        :param new_attrs: the attributes from the parallel subprocess to be
+        :param new_attrs: The attributes from the parallel subprocess to be
                           updated in the main builder
         """
         pass
@@ -591,7 +592,8 @@ class Builder:
 
         if self.parallel_ok:
             # number of subprocesses is parallel-1 because the main process
-            # is busy loading and post-transforming doctrees and doing write_doc_serialized();
+            # is busy loading and post-transforming doctrees and doing
+            # write_doc_serialized();
             # in case the global configuration enable_parallel_post_transform
             # is active the main process does nothing
             self._write_parallel(sorted(docnames),
@@ -610,6 +612,34 @@ class Builder:
                 self.write_doc(docname, doctree)
 
     def _write_parallel(self, docnames: Sequence[str], nproc: int) -> None:
+        def write_process_serial_post_transform(
+            docs: list[tuple[str, nodes.document]],
+        ) -> None:
+            self.app.phase = BuildPhase.WRITING
+            # The doctree has been post-transformed (incl. write_doc_serialized)
+            # in the main process, only write_doc() is needed here
+            for docname, doctree in docs:
+                self.write_doc(docname, doctree)
+
+        def write_process_parallel_post_transform(docs: list[str]) -> bytes:
+            assert self.env.config.enable_parallel_post_transform
+            self.app.phase = BuildPhase.WRITING
+            # run post-transform, doctree-resolved and write_doc_serialized in parallel
+            for docname in docs:
+                doctree = self.env.get_and_resolve_doctree(docname, self)
+                # write_doc_serialized is assumed to be safe for all Sphinx
+                # internal builders. Some builders merge information from post-transform
+                # and write_doc_serialized back to the main process using
+                # Builder.post_transform_merge_attr and
+                # Builder.merge_builder_post_transform
+                self.write_doc_serialized(docname, doctree)
+                self.write_doc(docname, doctree)
+            merge_attr = {
+                attr: getattr(self, attr, None)
+                for attr in self.post_transform_merge_attr
+            }
+            return pickle.dumps(merge_attr, pickle.HIGHEST_PROTOCOL)
+
         # warm up caches/compile templates using the first document
         firstname, docnames = docnames[0], docnames[1:]
         self.app.phase = BuildPhase.RESOLVING
@@ -617,30 +647,6 @@ class Builder:
         self.app.phase = BuildPhase.WRITING
         self.write_doc_serialized(firstname, doctree)
         self.write_doc(firstname, doctree)
-
-        def write_process(docs: list[tuple[str, nodes.document]]) -> bytes | None:
-            self.app.phase = BuildPhase.WRITING
-            if self.env.config.enable_parallel_post_transform:
-                # run post-transform, doctree-resolved and write_doc_serialized in parallel
-                for docname, _ in docs:
-                    doctree = self.env.get_and_resolve_doctree(docname, self)
-                    # write_doc_serialized is assumed to be safe for all Sphinx
-                    # internal builders. Some builders merge information from post-transform
-                    # and write_doc_serialized back to the main process using
-                    # Builder.post_transform_merge_attr and
-                    # Builder.merge_builder_post_transform
-                    self.write_doc_serialized(docname, doctree)
-                    self.write_doc(docname, doctree)
-                merge_attr = {
-                    attr: getattr(self, attr, None)
-                    for attr in self.post_transform_merge_attr
-                }
-                return pickle.dumps(merge_attr, pickle.HIGHEST_PROTOCOL)
-            for docname, doctree in docs:
-                # doctree has been post-transformed (incl. write_doc_serialized)
-                # in the main process, only write_doc is needed here
-                self.write_doc(docname, doctree)
-            return None
 
         tasks = ParallelTasks(nproc)
         chunks = make_chunks(docnames, nproc)
@@ -650,23 +656,30 @@ class Builder:
         progress = status_iterator(chunks, __('writing output... '), "darkgreen",
                                    len(chunks), self.app.verbosity)
 
-        def merge_builder(args: list[tuple[str, NoneType]], new_attrs_pickle: bytes) -> None:
-            if self.env.config.enable_parallel_post_transform:
-                new_attrs: dict[str, Any] = pickle.loads(new_attrs_pickle)
-                self.merge_builder_post_transform(new_attrs)
+        def on_chunk_done(args: list[tuple[str, NoneType]], result: NoneType) -> None:
+            next(progress)
+
+        def merge_builder(
+            args: list[tuple[str, NoneType]], pickled_new_attrs: bytes, /,
+        ) -> None:
+            assert self.env.config.enable_parallel_post_transform
+            new_attrs: dict[str, Any] = pickle.loads(pickled_new_attrs)
+            self.merge_builder_post_transform(new_attrs)
             next(progress)
 
         self.app.phase = BuildPhase.RESOLVING
         for chunk in chunks:
-            arg: list[tuple[str, nodes.document | None]] = []
-            for docname in chunk:
-                if not self.env.config.enable_parallel_post_transform:
+            if self.env.config.enable_parallel_post_transform:
+                tasks.add_task(write_process_parallel_post_transform,
+                               chunk, merge_builder)
+            else:
+                arg = []
+                for docname in chunk:
                     doctree = self.env.get_and_resolve_doctree(docname, self)
                     self.write_doc_serialized(docname, doctree)
                     arg.append((docname, doctree))
-                else:
-                    arg.append((docname, None))
-            tasks.add_task(write_process, arg, merge_builder)
+                tasks.add_task(write_process_serial_post_transform,
+                               arg, on_chunk_done)
 
         # make sure all threads have finished
         tasks.join()
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -75,6 +75,15 @@ class Builder:
     supported_remote_images = False
     #: The builder supports data URIs or not.
     supported_data_uri_images = False
+    #: Builder attributes that should be returned from parallel
+    #: post transformation, to be merged to the main builder in
+    #: :py:class:`~sphinx.builders.Builder.merge_builder_post_transform`.
+    #: Attributes in the list must be pickleable.
+    #: The approach improves performance when
+    #: pickling and sending data over pipes because only a
+    #: subset of the builder attributes are commonly needed for merging
+    #: to the main process builder instance.
+    post_transform_merge_attr: list[str] = []
 
     def __init__(self, app: Sphinx, env: BuildEnvironment) -> None:
         self.srcdir = app.srcdir
@@ -122,6 +131,24 @@ class Builder:
     def init(self) -> None:
         """Load necessary templates and perform initialization.  The default
         implementation does nothing.
+        """
+        pass
+
+    def merge_builder_post_transform(self, new_attrs: dict[str, Any]) -> None:
+        """Give builders the option to merge any post-transform information
+        coming from a parallel sub-process back to the main process builder.
+        This can be useful for extensions that consume that information
+        in the build-finish phase.
+        The function is called once for each finished subprocess.
+        Builders that implement this function must also define
+        the class attribute
+        :py:attr:`~sphinx.builders.Builder.post_transform_merge_attr` as it defines
+        which builder attributes shall be returned to the main process for merging.
+
+        The default implementation does nothing.
+
+        :param new_attrs: the attributes from the parallel subprocess to be
+                          updated in the main builder
         """
         pass
 
@@ -564,7 +591,9 @@ class Builder:
 
         if self.parallel_ok:
             # number of subprocesses is parallel-1 because the main process
-            # is busy loading doctrees and doing write_doc_serialized()
+            # is busy loading and post-transforming doctrees and doing write_doc_serialized();
+            # in case the global configuration enable_parallel_post_transform
+            # is active the main process does nothing
             self._write_parallel(sorted(docnames),
                                  nproc=self.app.parallel - 1)
         else:
@@ -581,11 +610,6 @@ class Builder:
                 self.write_doc(docname, doctree)
 
     def _write_parallel(self, docnames: Sequence[str], nproc: int) -> None:
-        def write_process(docs: list[tuple[str, nodes.document]]) -> None:
-            self.app.phase = BuildPhase.WRITING
-            for docname, doctree in docs:
-                self.write_doc(docname, doctree)
-
         # warm up caches/compile templates using the first document
         firstname, docnames = docnames[0], docnames[1:]
         self.app.phase = BuildPhase.RESOLVING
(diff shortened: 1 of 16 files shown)


picnixz:
Quick comment (still busy I am):

- I would prefer to have this feature disabled by default and possibly enabled in Sphinx 9.x (have at least 1 major version for upgrading).
- With the feature disabled by default, the PR could be smaller. And there would be less risks (I clearly don't want to risk breaking projects with the latest Sphinx version).

I have no time to contribute to Sphinx until February so I can only comment things here and there. 

I think it's a good approach though. However we need to carefully review what happens and mark the feature as experimental imo.

danwos:
I like the PR and also understand the concerns of @picnixz.

For me, Sphinx should get a configuration variable, like `parallel_post_transform`. 
If this is `True` the parallel execution of this PR gets activated.
By default it is currently `False`, with the plan to set it to `True` with version `9.x`

With Sphinx `8.x` maybe a Deprecation warning can be shown, that with version `9.x` an extension needs to provide `parallel_post_transform_safe`. If this is not the case, it will be set to `True` for the specific version (from `9.x` on.)

But I would keep the already provided changes for the internal extensions, as @ubmarco has already tested all of them and he knows that setting `parallel_post_transform_safe` to `True` is safe for them.
Please keep in mind that 54 of the 61 files just got this line as change `'parallel_post_transform_safe': True,`.
So the PR touches 2-3 files with more complex stuff only.

picnixz:
> For me, Sphinx should get a configuration variable, like parallel_post_transform.
If this is True the parallel execution of this PR gets activated.
By default it is currently False, with the plan to set it to True with version 9.x

That's a good idea.

> has already tested all of them and he knows that setting parallel_post_transform_safe to True is safe for them.

While tests show indeed that the built-in extensions are safe, I don't think we have a 100% coverage so there might be corner cases where we could be surprised. Don't get me wrong here, I think it's fine if we enable this for the default extensions. However, I cannot assure you that we won't break any existing projects that we are not aware of (especially those that automatically update to the latest version of their dependencies).

With the configuration value, we would also reduce the lines that this PR touches (although it's not much).

ubmarco:
Relevant issues: https://github.com/sphinx-doc/sphinx/issues/10779 and https://github.com/sphinx-doc/sphinx/issues/11448

ubmarco:
@picnixz I refactored the PR as discussed.
- There is a new global configuration parameter `enable_parallel_post_transform` which is `False` by default
- The touched files went down
- Docs were added
- CI is still happy

The only thing missing is a test case from my point of view.

bmwiedemann:
Please ensure that this does not make race-conditions worse - see issue #2946

danwos:
#2946 is for me not fulfilling the definition of "race -conditions" because Sphinx gets called twice from external, with the same config of doctree-location. Even MS Word and others have the same problem if the tools get called twice for the same file. Instance "A" is normally not informed by instance "B" about a documentation change. And "A" would overwrite the results from "B", in most cases.
For me, the ticket asks more for a database behavior, so to support concurrent requests. Nothing that Sphinx needs to provide (imho). 
Also, the workaround is simple, without losing performance: Let the first builder use `-j auto` to get the build done on 100% CPU usage. Then invoke the second builder, which will reuse the doctress from the first run and nearly calculates the new output format only.

Internally race-conditions should not be a thing in Sphinx, as long as the code dealing with one file is called serial. 
What is still the case here. Only merging data back from concurrent doctree-handlers to the main ENV object is something you need to care about and avoid "race-conditions", but this is also handled by Sphinx and some extra small voodoo in this PR.

However, in my understanding, this PR does not make it worse, as it is already. :)

bmwiedemann:
There are also issues with `sphinx-build -j auto` (e.g. in qemu, kernel-docs)

picnixz:
> I refactored the PR as discussed.

Thanks for that. Now I'm only available for a few days before going offline until mid February. Today I cleaned up what I needed to do but there are a lot of ongoing issues that we need to address (those related to test failures for instance). 

Also, on the organisational side we are doing something internally so we won't be doing anything until everything is arranged and done (probably early January). As such, you'll need to wait again (sorry for that).

sphinx/builders/linkcheck.py
```diff
@@ -60,12 +60,25 @@ class CheckExternalLinksBuilder(DummyBuilder):
...
         self.hyperlinks: dict[str, Hyperlink] = {}
         # set a timeout for non-responding servers
         socket.setdefaulttimeout(5.0)
 
+    def merge_builder_post_transform(self, new_attrs: dict[str, Any]) -> None:
+        """Merge hyperlinks back to the main builder after parallel
+        post-transformation.
+
+        param new_attrs: the attributes from the parallel subprocess to be
+                         udpated in the main builder (self)
```
AA-Turner:
```suggestion
        :param new_attrs: the attributes from the parallel subprocess to be
                          udpated in the main builder (self)
```

AA-Turner:
@ubmarco I don't seem to be able to push to this branch, would you be able to check please?

https://github.com/AA-Turner/sphinx/tree/mh-parallel-post-transform-parallel-write-doc-serial

A

ubmarco:
> @ubmarco I don't seem to be able to push to this branch, would you be able to check please?
> 
> https://github.com/AA-Turner/sphinx/tree/mh-parallel-post-transform-parallel-write-doc-serial
> 
> A

I think the problem is that I created the fork on our organization and not on my user. According to the [docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) it only works for forks in personal accounts. Will keep that in mind for the future.
I rebased my branch on current master (prefer this for cleaner history) and added your `style` commit.

sphinx/builders/linkcheck.py
ubmarco: fixed by the incorporated `style` commit

AA-Turner:
I've added a new commit (https://github.com/AA-Turner/sphinx/commit/14899132a113be09c166acde7440a1ef4c69888a) in https://github.com/AA-Turner/sphinx/tree/mh-parallel-post-transform-parallel-write-doc-serial.

I'm hesitant about adding new public API (`merge_builder_post_transform` / `post_transform_merge_attr`), as we'll need to support it for the long term. But I don't have any proposals on an alternate design at the moment, so it might be the best option.

A

ubmarco:
I integrated your `Refactoring` commit.

arwedus:
I've tried the sphinx version of this PR in large projects (3000 - 30.000 rst files), and it really speeds up things. I'd be happy to see this in the next sphinx release!

picnixz:
@arwedus Could you also give me the runtimes when

- this PR + #11939 are enabled,
- this PR is enabled,
- #11939 is enabled,
- none of the two PRs are enabled?

I don't have a huge project in production to test against and thus I'd like some feedback from projects as yours.

tox.ini
```diff
@@ -24,7 +24,7 @@ setenv =
     PYTHONWARNINGS = error
     PYTEST_ADDOPTS = {env:PYTEST_ADDOPTS:} --color yes
 commands=
-    python -X dev -X warn_default_encoding -m pytest --durations 25 {posargs}
+    python -X dev -X warn_default_encoding -m pytest -n 4 --durations 25 {posargs}
```
picnixz: Using xdist is not possible until I fixed the side effects.

sphinx/testing/util.py
```diff
@@ -91,8 +91,8 @@ def __init__(
         status: IO | None = None,
         warning: IO | None = None,
         tags: list[str] | None = None,
-        docutils_conf: str | None = None,
-        parallel: int = 0,
+        docutilsconf: str | None = None,
```
picnixz:
```suggestion
        docutils_conf: str | None = None,
```

This class can still be public API so we shouldn't change this name.

sphinx/testing/util.py
```diff
@@ -91,8 +91,8 @@ def __init__(
...
+        parallel: int = 4,
```
picnixz:
```suggestion
        parallel: int = 0,
```

This default value is also public so don't change it.

tests/test_versioning.py
```diff
@@ -17,7 +17,10 @@ def _setup_module(rootdir, sphinx_test_tempdir):
     srcdir = sphinx_test_tempdir / 'test-versioning'
     if not srcdir.exists():
         shutil.copytree(rootdir / 'test-versioning', srcdir)
-    app = SphinxTestApp(srcdir=srcdir)
+    # parallelisation is not supported by this test case
+    # as the global variable 'doctrees' is not preserved
+    # when subprocesses finish
+    app = SphinxTestApp(srcdir=srcdir, parallel=0)
```
picnixz, line 23: Revert it.

picnixz (changes requested):
Various comments for the testing component

ubmarco:
Oh my, these latest changes are not supposed to be pushed to this PR, it's for my local testing. Sorry, will revert them.

tests/test_versioning.py
ubmarco: I want to keep this change as this specific test case really does not support parallel mode.
Setting this explicitly avoids issues when running all test cases in parallel mode (which I did a lot of testing this PR).

sphinx/testing/util.py
ubmarco: Reverted this

sphinx/testing/util.py
ubmarco: I reverted this. It was a mistake.

tox.ini
ubmarco: I reverted this. It was a mistake.

ubmarco:
I reverted the unintended changes of my last commit and rebased the PR on latest master.

sphinx/builders/__init__.py
```diff
@@ -581,11 +612,34 @@ def _write_serial(self, docnames: Sequence[str]) -> None:
...
+            for docname in docs:
+                doctree = self.env.get_and_resolve_doctree(docname, self)
+                # write_doc_serialized is assumed to be safe for all Sphinx
+                # internal builders. Some builders merge information from post-transform
+                # and write_doc_serialized back to the main process using
+                # Builder.post_transform_merge_attr and
+                # Builder.merge_builder_post_transform
+                self.write_doc_serialized(docname, doctree)
+                self.write_doc(docname, doctree)
+            merge_attr = {
```
picnixz, line 637: Could you add (in the docstring or somewhere) that the attributes defined by `post_transform_merge_attr` should be picklable?

sphinx/builders/__init__.py
```diff
@@ -125,6 +133,26 @@ def init(self) -> None:
...
+        that information in the build-finish phase.
+        The function is called once for each finished subprocess.
+        Builders that implement this function must also define the class
+        attribute :py:attr:`~sphinx.builders.Builder.post_transform_merge_attr`
+        as it defines which builder attributes shall be returned to
+        the main process for merging.
+
+        The default implementation does nothing.
+
+        :param new_attrs: The attributes from the parallel subprocess to be
```
picnixz, line 151: Should it be a dict or could you live with a `Mapping` (so that we really show that we do not act on the inputs).

sphinx/builders/__init__.py
```diff
@@ -581,11 +612,34 @@ def _write_serial(self, docnames: Sequence[str]) -> None:
...
+                doctree = self.env.get_and_resolve_doctree(docname, self)
+                # write_doc_serialized is assumed to be safe for all Sphinx
+                # internal builders. Some builders merge information from post-transform
+                # and write_doc_serialized back to the main process using
+                # Builder.post_transform_merge_attr and
+                # Builder.merge_builder_post_transform
+                self.write_doc_serialized(docname, doctree)
+                self.write_doc(docname, doctree)
+            merge_attr = {
+                attr: getattr(self, attr, None)
```
picnixz, line 638: Should we really return `None` if the attribute does not exist? shouldn't we raise an exception otherwise?

sphinx/builders/__init__.py
```diff
@@ -598,21 +652,34 @@ def write_process(docs: list[tuple[str, nodes.document]]) -> None:
...
-        # (see: ``on_chunk_done()`` function)
+        # (see: ``merge_builder()`` function)
         progress = status_iterator(chunks, __('writing output... '), "darkgreen",
                                    len(chunks), self.app.verbosity)
 
         def on_chunk_done(args: list[tuple[str, NoneType]], result: NoneType) -> None:
             next(progress)
 
+        def merge_builder(
+            args: list[tuple[str, NoneType]], pickled_new_attrs: bytes, /,
```
picnixz, line 663: What is the *args* ? Why is the second item always ``None``?

sphinx/builders/_epub_base.py
```diff
@@ -167,6 +168,17 @@ def init(self) -> None:
         self.use_index = self.get_builder_config('use_index', 'epub')
         self.refnodes: list[dict[str, Any]] = []
 
+    def merge_builder_post_transform(self, new_attrs: dict[str, Any]) -> None:
```
picnixz, line 171: AFAICT, I think *new_attrs* are not meant to be writable so a mapping should be fine. 

By the way, "new_attrs" is a bit confusing for me. Maybe "context" or "extras" could be a better name.

sphinx/builders/html/__init__.py
```diff
@@ -240,6 +242,32 @@ def init(self) -> None:
...
+        :param new_attrs: the attributes from the parallel subprocess to be
+                          udpated in the main builder (self)
+        """
+        # handle indexer
+        if self.indexer is None:
+            lang = self.config.html_search_language or self.config.language
+            self.indexer = IndexBuilder(self.env, lang,
+                                        self.config.html_search_options,
+                                        self.config.html_search_scorer)
+        indexer_data = new_attrs['indexer']
```
picnixz, line 258: With this, I highly think we should use "context" instead of "new_attrs". Because it's more like an "update" rather than replacing one attribute with another.

sphinx/builders/__init__.py
chrisjsewell, line 637: This is not a robust/flexible enough mechanism for merging data generated during post-processing:

- It only accounts for data saved on to the `Builder`, not on to the `Environment`
- It doesn't give any hook for extensions to merge data

chrisjsewell (changes requested):

sphinx/config.py
```diff
@@ -254,6 +254,7 @@ class Config:
         'smartquotes_excludes': _Opt(
             {'languages': ['ja'], 'builders': ['man', 'text']}, 'env', ()),
         'option_emphasise_placeholders': _Opt(False, 'env', ()),
+        'enable_parallel_post_transform': _Opt(False, 'html', ()),
```
picnixz, line 257: Should we use 'html' or 'env' ? (I always forget how it works).

sphinx/search/__init__.py
```diff
@@ -553,3 +553,12 @@ def get_js_stemmer_code(self) -> str:
                     (base_js, language_js, self.lang.language_name))
         else:
             return self.lang.js_stemmer_code
+
+    def __getstate__(self):
```
picnixz, line 557:
```suggestion
    def __getstate__(self) -> dict[str, Any]:
```

(and import Any, if needed)

sphinx/search/__init__.py
```diff
@@ -553,3 +553,12 @@ def get_js_stemmer_code(self) -> str:
...
+        """Get the object's state.
+
+        Return a copy of self.__dict__ (which contains all instance attributes),
+        to avoid modifying the original state.
+        """
+        # remove env for performance reasons - it is not not needed by consumers
```
picnixz, line 563:
```suggestion
        # remove env for performance reasons - it is not needed by consumers
```

tests/test_versioning.py
picnixz: Ah sorry for my "harsh" comment. Actually, it was because you had the 'parallel=4' before (but now it's fine).

sphinx/builders/__init__.py
chrisjsewell: There should be two clear "hook" points, that "anyone" (builders, extensions, etc) can use:

1. A hook function/method where they can write logic to add data to the "entity" that will be pickled and returned from a single process
2. A hook function/method where they can write logic to "merge" the pickled data back into the core process

sphinx/config.py
chrisjsewell: Currently yeh it should be `env`, ideally it would a value to signify that all output formats should be rebuild (whilst still using the cached doctrees and env) but this does not currently exist

arwedus:
> @arwedus

Unfortunately, I can't test this right now, because if I install sphinx with

    pip install

I get this error by sphinx:

```
sphinx.errors.SphinxWarning: Failed to convert typing.Any to a set or tuple
```

I've tried with verbose but the log is not super useful:

```
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/config.py", line 514, in _validate_valid_types
    valid_types = tuple(valid_types)
  File "/usr/lib/python3.10/typing.py", line 312, in inner
    return func(*args, **kwds)
  File "/usr/lib/python3.10/typing.py", line 403, in __getitem__
    return self._getitem(self, parameters)
  File "/usr/lib/python3.10/typing.py", line 425, in Any
    raise TypeError(f"{self} is not subscriptable")
TypeError: typing.Any is not subscriptable

During handling of the above exception, another exception occurred:

  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/cmd/build.py", line 337, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/application.py", line 229, in __init__
    self.setup_extension(extension)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/application.py", line 402, in setup_extension
    self.registry.load_extension(self, extname)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/registry.py", line 466, in load_extension
    metadata = setup(app)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/myst_parser/__init__.py", line 12, in setup
    setup_sphinx(app, load_parser=True)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/myst_parser/sphinx_ext/main.py", line 59, in setup_sphinx
    app.add_config_value(f"myst_{name}", default, "env", types=Any)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/application.py", line 541, in add_config_value
    self.config.add(name, default, rebuild, types)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/config.py", line 434, in add
    valid_types = _validate_valid_types(types)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/config.py", line 516, in _validate_valid_types
    logger.warning(__('Failed to convert %r to a set or tuple'), valid_types)
  File "/usr/lib/python3.10/logging/__init__.py", line 1847, in warning
    self.log(WARNING, msg, *args, **kwargs)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/util/logging.py", line 131, in log
    super().log(level, msg, *args, **kwargs)
  File "/usr/lib/python3.10/logging/__init__.py", line 1879, in log
    self.logger.log(level, msg, *args, **kwargs)
  File "/usr/lib/python3.10/logging/__init__.py", line 1547, in log
    self._log(level, msg, args, **kwargs)
  File "/usr/lib/python3.10/logging/__init__.py", line 1624, in _log
    self.handle(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1634, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 1696, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 964, in handle
    rv = self.filter(record)
  File "/usr/lib/python3.10/logging/__init__.py", line 821, in filter
    result = f.filter(record)
  File "/home/.../.sphinx-venv/lib/python3.10/site-packages/sphinx/util/logging.py", line 432, in filter
    raise exc
sphinx.errors.SphinxWarning: Failed to convert typing.Any to a set or tuple
```

arwedus:
@picnixz : With my patches to both myst-parser and breathe, I managed to run the build on my rather big and complicated sphinx project.

It stands at 721 source files, contains lots of sphinx-needs, intersphinx links, plantUML, dynamic functions, sphinx-autoapi and a dozen other extensions.

Build times for clean build (deleted build director):

sphinx 7.3, parallel_post_transform = true, python 3.10:

real    4m53,530s
user    21m12,846s
sys     1m15,931s


sphinx 7.2.6, python 3.10:

real    7m54,836s
user    14m34,006s
sys     1m20,964s


I tested on a beefy machine, where any improvement in parallelism should bring down the wall time: 36-core Intel(R) Core(TM) i9-9980XE CPU @ 3.00GHz, 128 GB RAM.

Consecutive runs gave similar results. The build with sphinx 7.2.6 threw some unexplainable errors, so the numbers for user time might be a bit off. The results clearly show improved parallelism. I wonder about the increase in total CPU time ("user"), might be due to issues with the sphinx 7.2.6 build.

picnixz:
@arwedus Thank you very much for the timings. If possible, could you also time your project without `parallel_post_transform` (i.e., HEAD vs 7.2.6 release) for an *incremental* build time (an perhaps give me the numbers for a fresh build as well). I would like to know the direct consequences of https://github.com/sphinx-doc/sphinx/pull/11939.

sphinx/builders/__init__.py
ubmarco: I understand the approach is not super flexible. There are however some reasons why I went for it:

1. Pickling big objects can be quite expensive.
2. The environment is written to disc earlier in the process iirc. So what's the point of giving extensions an option to modify it?
3. Iirc the built-in builders have attributes that are not picklable. Custom builders can easily have more of these problems.

In general I think the whole idea of sending huge objects through pipes for flexibility has too many downsides. It's slow, can break, is difficult to debug and hard to make right. I'm also worrying about the implications when moving to spawn as child start method (or going for libraries such as mpire).

I just went for the slim variant. Going forward it would make sense to offer an alternative approach to the whole multiprocessing mechanism. E.g sqlite as data store - for surgical modifications by extensions instead of ever growing Python objects.

sphinx/builders/_epub_base.py
```diff
@@ -167,6 +168,17 @@ def init(self) -> None:
...
+        """Merge images back to the main builder after parallel
+        post-transformation.
+
+        :param new_attrs: the attributes from the parallel subprocess to be
+                          udpated in the main builder (self)
```
cmarqu, line 176:
```suggestion
                          updated in the main builder (self)
```

sphinx/builders/html/__init__.py
```diff
@@ -240,6 +242,32 @@ def init(self) -> None:
 
         self.use_index = self.get_builder_config('use_index', 'html')
 
+    def merge_builder_post_transform(self, new_attrs: dict[str, Any]) -> None:
+        """Merge images and search indexer back to the main builder after parallel
+        post-transformation.
+
+        :param new_attrs: the attributes from the parallel subprocess to be
+                          udpated in the main builder (self)
```
cmarqu, line 250:
```suggestion
                          updated in the main builder (self)
```

sphinx/builders/linkcheck.py
```diff
@@ -80,6 +82,17 @@ def init(self) -> None:
             )
             warnings.warn(deprecation_msg, RemovedInSphinx80Warning, stacklevel=1)
 
+    def merge_builder_post_transform(self, new_attrs: dict[str, Any]) -> None:
+        """Merge hyperlinks back to the main builder after parallel
+        post-transformation.
+
+        :param new_attrs: the attributes from the parallel subprocess to be
+                          udpated in the main builder (self)
```
cmarqu, line 90:
```suggestion
                          updated in the main builder (self)
```

cmarqu (commented):
Small typofixes.

Issue #11567, opened 2023-08-09
Circular dependencies between ``Sphinx`` and ``sphinxcontrib-*``


bruno-digitbio:
### Describe the bug

I'm not sure if this is something that the Sphinx team even is interested in explicitly supporting, but the latest release of the `sphinxcontrib` tooling cannot be used with `rules_python` (i.e. by users that use Bazel to install Sphinx).

We build our monorepo's docs at work with Sphinx, and today when running some automated tooling to bump package versions, I encountered issues like the following:

```bash
$ bazel build //...
ERROR: /home/bruno/.cache/bazel/_bazel_bruno/8f2b6fc37214e29f3e0fada1717a9254/external/pip_sphinx/BUILD.bazel:22:11: in py_library rule @pip_sphinx//:pkg: cycle in dependency graph:
    //building/python/sphinx:build_wrapper (ebf05eacb2537ac9057c50f99af724bc4fbe8ed0ce57ceeb546c8f679dec84e5)
.-> @pip_sphinx//:pkg (ebf05eacb2537ac9057c50f99af724bc4fbe8ed0ce57ceeb546c8f679dec84e5)
|   @pip_sphinxcontrib_devhelp//:pkg (ebf05eacb2537ac9057c50f99af724bc4fbe8ed0ce57ceeb546c8f679dec84e5)
`-- @pip_sphinx//:pkg (ebf05eacb2537ac9057c50f99af724bc4fbe8ed0ce57ceeb546c8f679dec84e5)
```

These errors can easily be solved by pinning the older versions at one patch number below the current one:
```
sphinxcontrib-applehelp==1.0.4
sphinxcontrib-devhelp==1.0.2
sphinxcontrib-htmlhelp==2.0.1
sphinxcontrib-qthelp==1.0.3
sphinxcontrib-serializinghtml==1.1.5
```

[...]

PR #11747, opened 2023-11-06
Make the first party extensions optional, add [extensions] extra

hroncok:
This is my attempt to make all the sphinxcontrib requirements optional.

### Feature or Bugfix

- Feature

### Purpose

We package sphinx in Fedora Linux and we don't really have a strong reason to pull in packages like applehelp there.
I decided to see if we could get rid of them.

### Detail

<del>Note that this is a very early draft. I am yet to test this in Fedora and see if this approach is feasible.</del> It works for us.

If it is, this obviously requires some documentation changes and possible deprecations. I am however unable to drive that forward right now. I would be interested to hear from the Sphinx maintainers whether this approach makes sense or not at all.

### Relates
- https://github.com/sphinx-doc/sphinx/issues/11567

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,12 +55,6 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
-    "sphinxcontrib-applehelp",
-    "sphinxcontrib-devhelp",
-    "sphinxcontrib-jsmath",
-    "sphinxcontrib-htmlhelp>=2.0.0",
-    "sphinxcontrib-serializinghtml>=1.1.9",
-    "sphinxcontrib-qthelp",
     "Jinja2>=3.0",
     "Pygments>=2.14",
     "docutils>=0.18.1,<0.21",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -76,6 +70,24 @@ dependencies = [
...
+    "sphinxcontrib-htmlhelp>=2.0.0",
+]
+serializinghtml = [
+    "sphinxcontrib-serializinghtml>=1.1.9",
+]
+qthelp = [
+    "sphinxcontrib-qthelp",
+]
 docs = [
     "sphinxcontrib-websupport",
```
hroncok, line 100: TODO maybe add the dependencies here?

hroncok:
I have identified 5 Fedora packages that are impacted by this change. Most of them explicitly build the docs in some of the removed formats and are fixed by declaring a dependency on the particular sphinxcontrib package. One of them was a bit more interesting, as the tests only (accidentally) succeeded with `sphinxcontrib_jsmath-1.0.1-py3.12-nspkg.pth` insatlled: https://github.com/fastavro/fastavro/pull/730

Long story short, we would like to have this change in Fedora. Please let me know if we should carry this downstream-only or if you are eventually interested as well.

hroncok:
The flake8 CI fail in a file I have not changed.

AA-Turner:
This would be too painful a change in terms of deprecations at the moment, so I've closed this PR.

A

PR #11770, opened 2023-11-24
HTML search: Fix exact matching on titles with extra whitespace

wlach:
Subject: Fix exact matching on title when there are leading/trailing spaces

### Feature or Bugfix

- Bugfix

### Purpose

When there is a search term with whitespace (e.g. ` Superman`), the override to boost entries that match on the title (e.g. `Superman`) doesn't work properly. This PR fixes that.

### Detail

N/A

### Relates

No ticket filed for this issue, just noticed it.

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -288,9 +288,9 @@ const Search = {
     let results = [];
     _removeChildren(document.getElementById("search-progress"));
 
-    const queryLower = query.toLowerCase();
+    const queryLower = query.toLowerCase().trim();
     for (const [title, foundTitles] of Object.entries(allTitles)) {
-      if (title.toLowerCase().includes(queryLower) && (queryLower.length >= title.length/2)) {
+      if (title.toLowerCase().trim().includes(queryLower) && (queryLower.length >= title.length/2)) {
         for (const [file, id] of foundTitles) {
           let score = Math.round(100 * queryLower.length / title.length)
           results.push([


wlach:
@AA-Turner sorry for the direct ping, but would you mind reviewing this? I see you've reviewed and written code with this file before. It's a very simple fix and I think low risk.

picnixz:
I don't think there is anything wrong with that but to be sure, do we actually assume that spaces are important? Or is it possible to keep the spaces when we use quotes? the search we provide is not that good because there are many results that are not always the best ones but I always forget whether we support quotes or not...

Btw, Adam is preetty busy and me as well. The project has been stalled for some time but we expect some changes after New Year's Eve. I personally won't be available until mid Feb so don't expect too much from me.

wlach:
> I don't think there is anything wrong with that but to be sure, do we actually assume that spaces are important? Or is it possible to keep the spaces when we use quotes? the search we provide is not that good because there are many results that are not always the best ones but I always forget whether we support quotes or not...

I don't see any sign the search code handles quoted exact matches. 

The existing code trims forward/trailing spaces when not matching the title and tokenizing the terms passed in, so this is just making the title exact match subcase (which was fairly recently added) match its behaviour: https://github.com/sphinx-doc/sphinx/blob/451de8eb8dcfaf85016f0282b9c686da59b4c6d1/sphinx/themes/basic/static/searchtools.js#L257

> Btw, Adam is preetty busy and me as well. The project has been stalled for some time but we expect some changes after New Year's Eve. I personally won't be available until mid Feb so don't expect too much from me.

No problem, thanks for the heads up!

PR #11774, opened 2023-11-28
Adjust the expected string to match Python 3.11+ changed output

befeleme:
Subject: With Python 3.13.0a2 'test_autodoc_default_options' started failing. It's because Python 3.13.0a2 has changed the `__doc__` and `__weakref__` docstrings, see: https://github.com/python/cpython/issues/112266

### Feature or Bugfix
- future bugfix

### DIFF
--- a/tests/test_ext_autodoc_configs.py
+++ b/tests/test_ext_autodoc_configs.py
@@ -11,6 +11,7 @@ from sphinx.testing import restructuredtext
 from .test_ext_autodoc import do_autodoc
 
 IS_PYPY = platform.python_implementation() == 'PyPy'
+IS_PY311_AND_LATER = sys.version_info >= (3, 11)
 
 
 @contextmanager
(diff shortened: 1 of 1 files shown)


hroncok:
Is it necessary to differentiate the asserted contents? What if we ditched the ` (if defined)` part (always)?

befeleme:
Since `actual` is a list of strings, the change of even a single character in the expected output will result in the false-y assertion. The below could work but its readability is quite worse:
```
assert any('      list of weak references to the object' in output_part for output_part in actual)
```

hroncok:
Oh, I thought `actual` is a string and we are checking a substring here. Never mind me then.

befeleme:
...or I could join `actual` to achieve a single string and use `in` operator without more hassle.

hugovk:
Looks good!

I didn't spot this PR and opened a fairly-similar one this week: https://github.com/sphinx-doc/sphinx/pull/11793

mitya57:
In fact #11793 is better, because in the latest version the condition is improved (0b55aaec30d9b117fa824c631d7203ec3509a776).

AA-Turner:
Closed by #11793.

A

PR #11776, opened 2023-11-28
Add full-length versions of CLI flags

hugovk:
### Feature or Bugfix

- Feature

### Purpose
- Improve usability of `sphinx-build`

### Detail
- I find it very difficult to remember what the single letter command-line arguments are for `sphinx-build` and often have to re-look them up via `--help`
- I usually use `sphinx-build` via a `Makfile`, and long options are just fine there, and better at self-documenting
- Long options are a CLI best practice, see [Command Line Interface Guidelines](https://clig.dev/#arguments-and-flags)

> **Have full-length versions of all flags.** For example, have both `-h` and `--help`. Having the full version is useful in scripts where you want to be verbose and descriptive, and you don’t have to look up the meaning of flags everywhere.

- And [GNU Coding Standards](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html):

> Please define long-named options that are equivalent to the single-letter Unix-style options. We hope to make GNU more user friendly this way. 

---

I didn't add any for these two because I wasn't sure what their long versions could be:

> * `-D setting=value`
> * `-A name=value`

### Relates
- https://github.com/sphinx-doc/sphinx/pull/11210 added `--jobs`


doc/man/sphinx-build.rst
```diff
@@ -174,56 +198,77 @@ Options
...
+   .. versionchanged:: 7.3
+      Add ``--traceback`` long option.
+
+.. option:: -P, --pdb
 
    (Useful for debugging only.)  Run the Python debugger, :mod:`pdb`, if an
    unhandled exception occurs while building.
 
+   .. versionchanged:: 7.3
+      Add ``--pdf`` long option.
```
ezio-melotti:
```suggestion
      Add ``--pdb`` long option.
```

doc/man/sphinx-build.rst
```diff
@@ -174,56 +198,77 @@ Options
...
 
    .. versionchanged:: 7.3
 
       ANSI control sequences are stripped when writing to *file*.
 
-.. option:: -W
+   .. versionchanged:: 7.3
+      Add ``--write`` long option.
+
+.. option:: -W, --errors
```
ezio-melotti: Would `--fail-fast` be a better name than `--errors`?
I don't find `--errors` particularly descriptive.

doc/man/sphinx-build.rst
hugovk: Oops :)

doc/man/sphinx-build.rst
hugovk: Hmm, this option is often used in combination with:

```
  --keep-going          with -W, keep going when getting warnings
```

My thinking was these might go together:

```
--errors --keep-going
```

And `--fail-fast` would seem to contradict it:

```
--fail-fast --keep-going
```

Open to other suggestions!

doc/man/sphinx-build.rst
```diff
@@ -64,40 +64,55 @@ Options
...
 
    If given, always write all output files. The default is to only write output
    files for new and changed source files. (This may not apply to all
    builders.)
 
-.. option:: -E
+   .. versionchanged:: 7.3
+      Add ``--all`` long option.
+
+.. option:: -E, --no-env
```
picnixz: Maybe `--isolate` or `--no-incremental` to mimick Python/mypy options (I think `--no-incremental` is more correct since `--isolate` may assume not looking for the configuration file).

doc/man/sphinx-build.rst
```diff
@@ -155,12 +176,15 @@ Options
 
    .. versionadded:: 0.5
 
-.. option:: -n
+.. option:: -n, --nit-picky
```
picnixz:
```suggestion
.. option:: -n, --nitpicky
```

doc/man/sphinx-build.rst
```diff
@@ -174,56 +198,77 @@ Options
...
+.. option:: -q, --quiet
 
    Do not output anything on standard output, only write warnings and errors to
    standard error.
 
-.. option:: -Q
+   .. versionchanged:: 7.3
+      Add ``--quiet`` long option.
+
+.. option:: -Q, --really-quiet
```
picnixz:
```suggestion
.. option:: -Q, --silent
```

doc/man/sphinx-build.rst
picnixz: If possible, I would have suggested `--fatal-warnings` because it's exactly what it does for `gcc`. Or we can go for `--warning-is-error` or `--warn-as-error` (but I would prefer the former).

doc/man/sphinx-build.rst
hugovk: Ruff has `--isolated` to ignore config:

```
      --isolated
          Ignore all configuration files
```

Mypy has `--no-incremental` to disable the cacge:

```
Incremental mode:
  Adjust how mypy incrementally type checks and caches modules. Mypy caches type information about modules into a cache to let you speed up future
  invocations of mypy. Also see mypy's daemon mode: mypy.readthedocs.io/en/stable/mypy_daemon.html#mypy-daemon

  --no-incremental          Disable module cache (inverse: --incremental)
```

I couldn't find similar Python options.

Mypy's `--no-incremental` seems closer, will update to use that 👍

doc/man/sphinx-build.rst
```diff
@@ -155,12 +176,15 @@ Options
...
 
-.. option:: -n
+.. option:: -n, --nit-picky
 
    Run in nit-picky mode.  Currently, this generates warnings for all missing
    references.  See the config value :confval:`nitpick_ignore` for a way to
    exclude some references as "known missing".
 
+   .. versionchanged:: 7.3
+      Add ``--nit-picky`` long option.
```
hugovk:
```suggestion
      Add ``--nitpicky`` long option.
```

doc/man/sphinx-build.rst
```diff
@@ -174,56 +198,77 @@ Options
...
+      Add ``--quiet`` long option.
+
+.. option:: -Q, --really-quiet
 
    Do not output anything on standard output, also suppress warnings.  Only
    errors are written to standard error.
 
-.. option:: -w file
+   .. versionchanged:: 7.3
+      Add ``--really-quiet`` long option.
```
hugovk:
```suggestion
      Add ``--silent`` long option.
```

doc/man/sphinx-build.rst
```diff
@@ -174,56 +198,77 @@ Options
...
 
    .. versionchanged:: 7.3
 
       ANSI control sequences are stripped when writing to *file*.
 
-.. option:: -W
+   .. versionchanged:: 7.3
+      Add ``--write`` long option.
+
+.. option:: -W, --errors
```
hugovk:
```suggestion
.. option:: -W, --fatal-warnings
```

doc/man/sphinx-build.rst
```diff
@@ -174,56 +198,77 @@ Options
...
 
    Turn warnings into errors.  This means that the build stops at the first
    warning and ``sphinx-build`` exits with exit status 1.
 
+   .. versionchanged:: 7.3
+      Add ``--errors`` long option.
```
hugovk:
```suggestion
      Add ``--fatal-warnings`` long option.
```

doc/man/sphinx-build.rst
picnixz: Fot Python it's `-I` for "isolated".

picnixz:
Btw for the `-D` it's simply `--define` and for the `-A` I think it should be something like `--var=...` or `--html-var` or `--jinja-var` since it's for templates only iirc.

doc/man/sphinx-build.rst
AA-Turner: I prefer ``--fail-on-warning`` here, it's more descriptive, if a little wordier.

AA-Turner:
An alternate suggestion for ``-D`` would be ``--config-override``, to be more descriptive that the option is intended to override settings in the configuration file.

A

doc/man/sphinx-build.rst
```diff
@@ -241,7 +245,7 @@ Options
    .. versionchanged:: 7.3
       Add ``--write`` long option.
```
hugovk:
```suggestion
      Add ``--warning-file`` long option.
```

Similarly needed for the other two after this.

doc/man/sphinx-build.rst
picnixz: I usually try to come with flags that are used by other well-known tools because it makes less options to remember overall, but I think yours is a bit betetr when used with `--keep-going`

PR #11778, opened 2023-11-30
Add possibility to use custom Project class

eliemichel:
Add possibility to use custom Project class

### Feature or Bugfix

Feature

### Purpose

Using a custom Project class is meant to enable for instance rewriting rules in `path2doc`, to address [this issue](https://stackoverflow.com/questions/77512795/use-a-different-docname-in-build-tree-than-in-source-tree).

**NB** it is still unclear how to apply similar rules to linked images and documents.

**Context** I need this to manage the translations of my [LearnWebGPU programming guide](https://eliemichel.github.io/LearnWebGPU/). I cannot use an entierly different sphinx project for the different translations, because I want translations to live in the same doc tree (so that I can have extensions that help keeping track of what is translated or not -- which is done at the page level rather than using gettext).

### Detail

Upon configuration of the app (either in `conf.py` or in extensions), one can customize the project class:

```python
class TranslatedProject(Project):
    pass # [...]

app.project_class = TranslatedProject
```

### Relates
- [StackOverflow](https://stackoverflow.com/questions/77512795/use-a-different-docname-in-build-tree-than-in-source-tree)


picnixz:
Could you add a CHANGES entry and  a documentation entry? (or we can postpone the doc but create a new issue for that instead)

sphinx/application.py
```diff
@@ -147,6 +147,7 @@ def __init__(self, srcdir: str | os.PathLike[str], confdir: str | os.PathLike[st
         self.verbosity = verbosity
         self.extensions: dict[str, Extension] = {}
         self.registry = SphinxComponentRegistry()
+        self.project_class = Project  # can be modified by extensions in setup
```
chrisjsewell, line 150: Ughh, I feel this should be a method, like `set_project_class`, with some form of `override = False` argument, as we do for other methods, to guard against multiple extensions trying to override this

otherwise we are on shaky ground, if we start endorsing extensions to be directly changing attributes  of `Sphinx` (which in general will completely break sphinx)
ideally all these attributes would already be readonly, but such is life

chrisjsewell (changes requested):

sphinx/application.py
picnixz: I'd suggest adding a documentation entry telling that this feature is to be used at your own risk and that extensions should mention that they change the Project class in their documentation.

And yes, I'd agree that we could have a method instead with this `override` argument (instead of a property setter) and make the project class read-only but retrievable via a property.

eliemichel:
Maybe this PR can be seen as a Proof of Concept that does a little more than just asking for a new feature but since I am not familiar with this repo and its practice, it will take 10 times less time to project maintainers to do the polishing and copy paste the elements of documentation that I put in the original post.

Sorry, I'm not saying I am not going do eventually do this, but it is not on the top-most part of my priorities. I just updated CHANGES but could not find where the documentation entry is supposed to go. I agree that directly modifying `self.project_class` is not ideal, don't have time to change this right now.

picnixz:
No worries! Keep the PR opened and we can (when we also have time) do the changes then. Or you can come back later.

We have other tasks on the backlog so it might be stalled for a while (possibly for years since I don't want to promise anything). However, I'd be willing to review any related PR though.

Here, it's more important to vhange the logic of accessing and modifying the project class rather than the documentation entry itself but since it adds a non-trivial layer, the doc entry should be included in the same PR (contrary to my previous comment).

AA-Turner:
The linked SO question has been removed. The motivating example about translations would be better served by other feature requests to build multiple translations at once.

eliemichel:
FTR, I will continue to use my fork because I could not find another workaround. The original question was removed automatically because nobody could find an answer, meaning that there is indeed a need for a solution.

PR #11780, opened 2023-11-30
Add PyPI role

hugovk:
### Feature or Bugfix

- Feature

### Purpose
- Add a `pypi` role to turn ``` :pypi:`Sphinx#history` ``` into https://pypi.org/project/Sphinx/
- It's a [quite common](https://github.com/search?q=extlinks+https%3A%2F%2Fpypi.org%2Fproject%2F%25s+path%3A**%2Fconf.py&type=code) `extlinks` entry, so would be useful upstream in Sphinx

### Detail
- Add `pypi` role, similar to existing `pep` and `rfc` roles

### Relates
- https://github.com/sphinx-doc/sphinx/pull/11781

### DIFF
--- a/doc/usage/restructuredtext/roles.rst
+++ b/doc/usage/restructuredtext/roles.rst
@@ -251,6 +251,17 @@ The following roles generate external links:
 
    For example: :pep:`8`
 
+.. rst:role:: pypi
+
+   A reference to a PyPI project.  This generates appropriate index entries.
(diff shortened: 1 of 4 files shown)


AA-Turner:
This sounds like https://pypi.org/project/sphinxcontrib-cheeseshop/, which I've just spring-cleaned a little. Perhaps we could advertise that better?

A

PR #11785, opened 2023-12-04
Display ``typing.Annotated`` metadata in the Python domain

### LINKED ISSUE #11773: Better rendering of Annotated types
For classes that have a `__class_getitem__()` method that returns an `Annotated` type, the extra metadata stored on the Annotated object is not rendered in the type hints in sphinx docs.

Extra metadata is included in rendered type hints.

**Additional context**
Minimal example that builds is at https://github.com/dstansby/quantity-typing-docs/tree/c69c2a8c86314ef672d4484072067888c2627ac7 The code and rendered docs are below:
```python
from typing import Annotated

__all__ = ["deg2rad"]

class A:
    def __class_getitem__(cls, t: type):
        return Annotated[cls, t]


def deg2rad(arg: A[str]) -> A[bytes]:
    return A()
```
Screenshot 2023-11-28 at 13 30 51

Note how only the class `A` is shown. Ideally I would like the type hint to read A[str].

Possibly related to https://github.com/sphinx-doc/sphinx/pull/9590

dstansby:
Subject: Add metadata to Annotated types

### Feature or Bugfix

- Feature

### Purpose
This adds metadata to `Annotated` types.

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
Fixes https://github.com/sphinx-doc/sphinx/issues/11773


sphinx/util/typing.py
```diff
@@ -365,7 +365,8 @@ def format_literal_arg(arg):
             args = ', '.join(map(format_literal_arg, annotation_args))
             return f'{module_prefix}Literal[{args}]'
         elif str(annotation).startswith('typing.Annotated'):  # for py39+
-            return stringify_annotation(annotation_args[0], mode)
+            meta = [stringify_annotation(meta, mode) for meta in annotation.__metadata__]
```
picnixz: Metadata is actually arbitrary so I'm not sure what should be done here. Here, you assume that your metadata are annotations but they can be non-annotations (like real values). So what should we do? 

Actually, I want to delay whatever rendering issues we have currently because we might change the logic of how rendering is done, e.g., by allowing custom rendering for each type of components.

tests/test_util_typing.py
```diff
@@ -360,8 +360,8 @@ def test_stringify_type_hints_pep_585():
 
 def test_stringify_Annotated():
     from typing import Annotated  # type: ignore[attr-defined]
-    assert stringify_annotation(Annotated[str, "foo", "bar"], 'fully-qualified-except-typing') == "str"
-    assert stringify_annotation(Annotated[str, "foo", "bar"], "smart") == "str"
+    assert stringify_annotation(Annotated[str, "foo", "bar"], 'fully-qualified-except-typing') == "str[foo, bar]"
+    assert stringify_annotation(Annotated[str, "foo", "bar"], "smart") == "str[foo, bar]"
```
picnixz: Can you add some tests where the metadata is more than just strings but like real instances of objects and see what happens?  See https://peps.python.org/pep-0593/#combining-runtime-and-static-uses-of-annotations for some examples.

Zac-HD (changes requested):
Consider the parameter `x: Annotated[int, Ge(0), Lt(10)]` ([using `annotated-types`](https://github.com/annotated-types/annotated-types#annotated-types)).  Rendering this as `x: int[Ge(0), Lt(10)]` would be absurd.

I therefore think that we should render `typing.Annotated` types in documentation with `f"Annotated[{stringify_annotation(T)}, {', '.join(repr(x) for x in metadata)}]`.

tests/test_util/test_util_typing.py
```diff
@@ -184,6 +187,11 @@ def test_restify_type_hints_containers():
                                            "[:py:obj:`None`]")
 
 
+def test_restify_Annotated():
+    assert restify(Annotated[str, "foo", "bar"]) == ':py:class:`~typing.Annotated`\\ [:py:class:`str`]'
+    assert restify(Annotated[str, "foo", "bar"], 'smart') == ':py:class:`~typing.Annotated`\\ [:py:class:`str`]'
```
Zac-HD: As above, it'd be really valuable for users of e.g. [annotated-types](https://pypi.org/project/annotated-types/) to show the repr of the annotations too.

AA-Turner: I haven't used annotated types, could you provide a sketch of what would currently be output and what you'd prefer? Just so we can capture that in a test.

Zac-HD:
```python
from typing import Annotated
from annotated_types import Gt, Lt, Len

def coinflip(
    p_heads: Annotated[float, Gt(0.0), Lt(1.0)],
    label: Annotated[str, Len(1, 10)] | None = None,
) -> bool:
    ...
```

would ideally render the signature as something like
> coinflip(p_heads: [Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated)[[float](https://docs.python.org/3/library/functions.html#float), Gt(gt=0.0), Lt(lt=1.0)], label: [Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated)[[str](https://docs.python.org/3/library/functions.html#func-str), Gt(gt=0.0), Lt(lt=1.0)]) -> [bool](https://docs.python.org/3/library/functions.html#bool)

using the `repr` of the dataclasses.

tests/test_util/test_util_typing.py
```diff
@@ -1,5 +1,5 @@
 """Tests util.typing functions."""
-
+import dataclasses
```
AA-Turner, line 3:
```suggestion

import dataclasses
```

AA-Turner:
Thanks all!

A

2bndy5:
I think I found a regression about this. We're using the `Annotated` class + pydantic to validate types of a user-defined config in conf.py...

```py
CSSClassType = Annotated[str, pydantic.AfterValidator(nodes.make_id)]

class CustomAdmonitionConfig(pydantic.BaseModel):
    title: Annotated[Optional[str], pydantic.Field(validate_default=True)] = None
    classes: List[CSSClassType] = []
```
But autodoc-ing `CustomAdmonitionConfig`'s `title` and `classes` attributes results in errors trying to reference the metadata:
> /opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/site-packages/sphinx_immaterial/custom_admonitions.py:docstring of sphinx_immaterial.custom_admonitions.CustomAdmonitionConfig.title:1: WARNING: py:obj reference target not found: typing.Annotated[str | None, FieldInfo(annotation=NoneType, required=True, validate_default=True)]


> /opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/site-packages/sphinx_immaterial/custom_admonitions.py:docstring of sphinx_immaterial.custom_admonitions.CustomAdmonitionConfig.classes:1: WARNING: py:obj reference target not found: typing.List[~typing.Annotated[str, AfterValidator(func=<function make_id at 0x7fd2a7b5cae0>)]]

I think lesson here is that not all metadata is compatible with this. Is there a way to disable it?

2bndy5:
I found the culprit. So, `sphinx_immaterial.apidoc.python.attribute_style`) monkey-patches the `handle_signature()` function of `sphinx.domains.python.PyAttribute` (and `PyProperty`) to convert the signature's type into xrefs using a monkey-patched `sphinx.domains.python._parse_annotation()`.

I'm not yet sure where to go from here. I'd prefer to skip displaying the `typing.Annotated` metadata all together because they are implementation details. If we do decide to start supporting `Annotated` metadata, then we would need to way to allow ignoring unresolved xref warnings in the signature's type.

PR #11789, opened 2023-12-10
remove redundant linting

danieleades:
ruff replaces flake8 and isort, so these can be removed


picnixz:
Is the isort integration complete? if we don't use features from isort not support by ruff, we could consider using ruff only.

danieleades:
converting to draft. I'll revisit this when #11815 has been addressed

AA-Turner:
Ruff doesn't yet seem to implement E704 or SIM113, so Flake8 is still required. I've removed isort, though.

A

danieleades:
@AA-Turner would you mind reopening this?

SIM113 is supported, as of ruff 0.1.14
E704 was [added in february last year](https://github.com/astral-sh/ruff/pull/2680), not sure why it's not listed in the docs ([i've asked the question](https://github.com/astral-sh/ruff/pull/3689#issuecomment-1902534188))

danieleades:
scratch that, created a new one - #11902

PR #11790, opened 2023-12-10
use ruff to autoformat

danieleades:
Ruff now also has the ability to auto-format code.

Creating this PR for comments. This would need to be integrated into CI before it could be merged

## Advantages

- consistent format style across codebase
- ruff formatting is optimised to reduce noise in diffs
- never think about formatting style again
- **never talk about formatting style in a PR again**
- consistent with Black code style, which is *very* widely used across python ecosystem
- it just looks so much nicer
- [ruff formatter is configurable](https://docs.astral.sh/ruff/settings/#format) (though i highly recommend using the defaults)

## disadvantages

- one-off pain/churn for existing PRs
- ruff formatter is configurable


picnixz:
I quicky looked at some files and some parts should not be reformatted I think (like the dictionaries being aligned, sometimes you want to keep that) so we need to add some on/off formatting for ruff. 

Otherwise, I think it's fine. And if we ever change our mind about how to format things, we can just add a custom formatter I think?

danieleades:
I'll revisit this when #11815 has been addressed

danieleades:
> I quicky looked at some files and some parts should not be reformatted I think (like the dictionaries being aligned, sometimes you want to keep that) so we need to add some on/off formatting for ruff.
> 
> Otherwise, I think it's fine. And if we ever change our mind about how to format things, we can just add a custom formatter I think?

can you point to some specific examples of formatting you have concerns about?

full disclosure, i'm an autoformatting die-hard. I happen to believe that having consistent formatting is more important than having formatting that you *like*, particularly in codebases which are large, or have many contributors. It just so happens that i also like the formatting that ruff provides, though i consider that a secondary benefit.

picnixz:
I personally don't mind the autoformat but I found a file where the dictionaries had aligned keys and sometimes it's better for visual purposes (it's probably the reason why it was like this). 

But since it's a lot of things that were touched, it's probably better to just let ruff do the formatting.

AA-Turner:
My view remains as in #9955. I'll close this PR as it will be continuously out-of-sync.

A

PR #11791, opened 2023-12-10
Fix some type errors in the C and C++ domains

danieleades:
type annotations added based on usage


AA-Turner:
Flake8 failing

danieleades:
> Flake8 failing

should be fixed now

PR #11796, opened 2023-12-12
Python domain: Correctly match when signature simultaneously has annotations for both the function and the return type

arashbm:
### Feature or Bugfix
- Bugfix

### Purpose
- Fixes python signature regex to successfully parse simultaneous annotations of function and return type.

### Detail
Python domain fails to match functions where both function name and return type have type annotation, e.g.
```python
hello[int](a, b, c) -> world[int]
````

The fix comes down to an overly eager `.*` in the regex that matches everything from the first `[` to the last `]` in one fell swoop.

I'm not 100% sure if the two new test cases are in the right place but if not, let me know and I'll move them.

### DIFF
--- a/sphinx/domains/python.py
+++ b/sphinx/domains/python.py
@@ -51,7 +51,7 @@ logger = logging.getLogger(__name__)
 py_sig_re = re.compile(
     r'''^ ([\w.]*\.)?            # class name(s)
           (\w+)  \s*             # thing name
-          (?: \[\s*(.*)\s*])?    # optional: type parameters list
+          (?: \[\s*(.*?)\s*\])?  # optional: type parameters list
           (?: \(\s*(.*)\s*\)     # optional: arguments
            (?:\s* -> \s* (.*))?  #           return annotation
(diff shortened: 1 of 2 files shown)


picnixz:
Ok, this one is propably my bad. Before reviewing it-, I want you to complete the tests that I wrote for PEP 695 (https://github.com/sphinx-doc/sphinx/blob/35965903177c6ed9a6afb62ccd33243a746a3fc0/tests/test_domain_py.py#L1849 and later) and check whether they still pass with a return value because I totally missed that case.

I really don't like regexes like that because we always miss some cases... 

Note: I'm only quickly available for a week or so before going off until mid february.

AA-Turner:
This seems important to support.

AA-Turner:
@picnixz do you want to take over this PR?

picnixz:
I'm travelling for the next two weeks but I can take over it once I'm back!

picnixz:
Superseded by https://github.com/sphinx-doc/sphinx/pull/13328.

PR #11804, opened 2023-12-19
Add :notranslate: metadata feature

### LINKED ISSUE #11787: Allow skipping documents (or paragraphs) in i18n

On some projects, the amount of documentation is large, but only part of that documentation is nice to translate, while part of it should not be translated.

Two examples:

- Python documentation (https://docs.python.org) — The C API documentation is bulky, unlikely to be useful for people who don't understand English, and very technical. In the python-docs-fr repo (CC @JulienPalard), the README documents that one should not translate it.
- Python Packaging User Guide (https://packaging.python.org) — The [tutorials](https://packaging.python.org/en/latest/tutorials) are nice to have translated, but the [PyPA specifications](https://packaging.python.org/en/latest/specifications) should not be.

In such cases, it would be very helpful to be able to entirely skip generating catalog messages for a given reST/Markdown file or even directly. This would obviate the need to document what should/shouldn't be translated, and make it easier to have an accurate representation of the translation progress. (The [Potodo](https://pypi.org/project/potodo) tool has an option to exclude given PO files from the statistics, but that won't work if you're using a single POT/PO file for everything (`gettext_compact = "mydoc"`), and it doesn't help translators on platforms like Weblate.)

This should also be reflected in the `|translation progress|` substitution.

jeanas:
Allow Gettext message extraction to be skipped on selected documents using a `:notranslate:` metadata field.

### Feature or Bugfix

Feature

### Relates

Fixes #11787

### DIFF
--- a/sphinx/transforms/i18n.py
+++ b/sphinx/transforms/i18n.py
@@ -342,6 +342,14 @@ class Locale(SphinxTransform):
     default_priority = 20
 
     def apply(self, **kwargs: Any) -> None:
+        # TODO/FIXME: If this document is marked :notranslate:, it might
+        # be better to skip translation here. Some paragraphs might
+        # still get translated in rare cases, e.g., if gettext_compact=True
+        # and there is an identical paragraph in a document that is not
+        # marked :notranslate:. This might be tricky, since self.env.metadata
+        # is only populated in the doctree-read event, which runs after
+        # this transform.
+
         settings, source = self.document.settings, self.document['source']
         msgstr = ''
 
--- a/doc/usage/advanced/intl.rst
+++ b/doc/usage/advanced/intl.rst
@@ -85,6 +85,10 @@ because code blocks do not contain references anyway.)
 .. versionadded:: 4.5
    The ``#noqa`` mechanism.
 
+If certain parts of the documentation should not be translated, e.g., you want
+to translate tutorials but not technical reference material, you can use the
+``:notranslate:`` metadata field. See :ref:`special-metadata-fields`.
+
 
 Translating with sphinx-intl
 ----------------------------
--- a/doc/usage/restructuredtext/field-lists.rst
+++ b/doc/usage/restructuredtext/field-lists.rst
@@ -34,6 +34,8 @@ document metadata, and is not displayed in the output.
    *docinfo* as normal and will be displayed in the output.
 
 
+.. _special-metadata-fields:
+
 Special metadata fields
 -----------------------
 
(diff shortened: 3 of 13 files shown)


sphinx/transforms/i18n.py
```diff
@@ -342,6 +342,14 @@ class Locale(SphinxTransform):
     default_priority = 20
 
     def apply(self, **kwargs: Any) -> None:
+        # TODO/FIXME: If this document is marked :notranslate:, it might
+        # be better to skip translation here. Some paragraphs might
+        # still get translated in rare cases, e.g., if gettext_compact=True
+        # and there is an identical paragraph in a document that is not
+        # marked :notranslate:. This might be tricky, since self.env.metadata
+        # is only populated in the doctree-read event, which runs after
+        # this transform.
```
jeanas, line 351: Oh dear. I just realized that this will also make the `translation_progress` variable wrong. It should not account for `:notranslate:` documents.

@AA-Turner Can you help here? Should I make the environment metadata collector run earlier somehow? Should I traverse the document tree directly to search for metadata in `is_translatable_document`?

picnixz: Correct me if I'm wrong but your issue is:

- You want to know whether the document should be translated or not.
- However this information is gathered by the collector which at the point when the transformation is executed is not yet called.
- So you end up translating everything even if it was marked `:notranslate:` (but was not yet detected).
- When you count the number of mistranslations, then you are also accounting those that were marked as `:notranslate:` since they were gathered during the `Locale` transform.

I don't know if running the metadata collector before is fine or not since it's triggered by the 'doctree-read' event. 

Maybe we could add a dedicated `MetadataCollectorTransform` which does the same but only for `:notranslate:` and run that transformation just before `Locale` by adding the information to the *doctree* itself and not to the environment?

jeanas: Sorry for the long hiatus. Yes, you've understood the problem correctly.

> Maybe we could add a dedicated MetadataCollectorTransform which does the same but only for :notranslate: and run that transformation just before Locale by adding the information to the doctree itself and not to the environment?

I don't know whether special-casing `:notranslate:` is the proper fix. I imagine that another solution is to just run the metadata collector earlier than `doctree-read`, but this might puzzle extension transforms that add metadata to the doctree in order to have it collected.

picnixz: Yes this is what I fear hence the special casing. We are already doing special casing for some stuff so I think it's fine but we might want to think of another solution. I don't have time for that now but maybe others have an idea?

jeanas:
I've lost interest to work on this, sorry. If anyone wants to pick it up, feel free.

PR #11815, opened 2023-12-25
Pin linting

danieleades:
The versions of linters used should be pinned so that all contributors see the same warnings/errors, and new warnings and errors are only introduced when the pinned versions are bumped.

This will help prevent the consistent breakage of PRs (and also local dev) that is preventing the CI in this project from giving useful feedback to maintainers and contributors. CI is *only* useful if it's generally passing, and any new warnings/errors are directly related to the changes made in a given PR.

i've not bothered pinning flake8 and isort because i think these should be removed entirely anyway (see #11789)

I've also not addressed the myriad warnings in this PR


AA-Turner:
I think now addressed.

A

picnixz:
By addressed you mean we pin ruff in the other PR? because it's not the case currently I think

PR #11819, opened 2023-12-26
inline SIM115 whitelist

danieleades:
- simplifies config
- SIM115 whitelist will be naturally removed when `sphinx.testing.path` is removed
- supports my never-ending quest to remove more lines than i add ;)


AA-Turner:
Thanks Daniel!

A

PR #11820, opened 2023-12-27
Enable the entire pycodestyle category

danieleades:


sphinx/ext/autodoc/__init__.py
```diff
@@ -2169,7 +2169,7 @@ def format_args(self, **kwargs: Any) -> str:
             kwargs.setdefault('unqualified_typehints', True)
 
         try:
-            if self.object == object.__init__ and self.parent != object:
+            if self.object == object.__init__ and not isinstance(self.parent, object):
```
picnixz: This one must be left as is  Because this check is always false. I think you need to manually check the rules for the isinstance/is/== checks (as well in the other huge PR).

picnixz (dismissed):

PR #11824, opened 2023-12-29
Allow extensions to define the keys returned by linkcode ext

n-peugnet:
Subject: Allow extensions to define the keys the linkcode extension should return for a domain.

### Feature or Bugfix

- Feature

### Purpose
- The linkcode extension currently only works for some [pre-defined domains](https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html#confval-linkcode_resolve).
  By allowing extensions to define the keys associated with a domain, it opens up the linkcode extension to a lot more new domains.
  For instance, in the sphinxcontrib-phpdomain, one could simply add the following in `setup(app)` function to make linkcode work for the `php` domain:

  ```py
  app.add_linkcode_domain('php', ['namespace', 'class', 'fullname'])
  ```

### Detail
- I am not sure that it is the best way to do it. I took inspiration from the methods above that allow to interact with autodoc extension.

### Relates
- Did not find any related issue.

P.S. Ah sorry, I made this commit on an old tag for my tests, I will rebase my branch.


picnixz:
It appears to be quite similar to what we do for search languages so I'd say it's fine?

n-peugnet:
@picnixz I just pushed commits that add a test and a changelog entry. Don't hesitate to edit it and my change in the docs as well if you find better ways to describe it.

picnixz:
Windows tests for PO files are flaky, don't worry about it

picnixz:
By the way, in the future, it's better not to force push commits. We squash them before merging and force-pushing makes review harder (I also used to rebase/force push)

doc/usage/extensions/linkcode.rst
```diff
@@ -38,6 +38,9 @@ Configuration
    - ``javascript``: ``object`` (name of the object), ``fullname``
      (name of the item)
 
+   Support for other domains can be added with
+   :py:meth:`.Sphinx.add_linkcode_domain()`.
```
picnixz: Can you provide an example of its usage?

sphinx/application.py
```diff
@@ -1178,6 +1178,15 @@ def add_autodoc_attrgetter(self, typ: type, getter: Callable[[Any, str, Any], An
...
         self.registry.add_autodoc_attrgetter(typ, getter)
 
+    def add_linkcode_domain(self, domain: str, keys: list[str]) -> None:
+        """Register a new list of keys to use for a domain by the
+        :mod:`sphinx.ext.linkcode` extension.
+
+        .. versionadded:: 7.3
+        """
+        from sphinx.ext.linkcode import domain_keys
+        domain_keys[domain] = keys
```
picnixz: Just wondering but what happens if two extensions do the same? should extend the keys and removing duplicates? 

Also, you should probably say (in the doc) that the keys an extension defines overrides whatever keys were previously defined if we go with the current implementation.

picnixz (commented):
Personally I don't mind the feature and I think it can land in 7.3. When we'll do the release and it's not included, you can ask for it be included (assuming you addressed the small comments, sorry if I didn't put them before).

doc/usage/extensions/linkcode.rst
n-peugnet: I will add one in a new commit.

sphinx/application.py
n-peugnet: > Just wondering but what happens if two extensions do the same? should extend the keys and removing duplicates?

I don't think so, the idea is that an extension that provides support for a new domain, for instance `sphinxcontrib-phpdomain`, can now add the `php` domain to the list of domains supported by `linkcode`. I don't think that anybody uses two extensions adding support for the same domain, so I don't see any chances of conflict here.

Another possibility I see would be to completely forbid overriding an existing key, or maybe add a warning when a key is overridden. But I thought overriding the existing keys could also be useful in a way.

Maybe we should wait for another opinion? I'm not in a hurry for this feature.

> Also, you should probably say (in the doc) that the keys an extension defines overrides whatever keys were previously defined if we go with the current implementation.

I will add a more descriptive explanation when we settle on the preferred behaviour.

picnixz: I would like the opinion of another maintainer/contributor. Since we are in a holiday period, it might take a while though. When we release new versions, we create an issue and you can ask at that moment if we didn't decide on anything.

AA-Turner:
@jdknight you were looking at linkcode recently - any thoughts on this?

jdknight:
> @jdknight you were looking at linkcode recently - any thoughts on this?

Adding support for custom domains looks like a good idea to me.

AA-Turner:
@n-peugnet I am not sure that this should be a method on `app`/`Sphinx`, as the first-party extensions should try and act like normal extensions as much as possible. As such, I've pushed a proposal to make this a function within `sphinx.ext.linkcode` -- please let me know your thoughts.

A

n-peugnet:
> I am not sure that this should be a method on `app`/`Sphinx`, as the first-party extensions should try and act like normal extensions as much as possible. As such, I've pushed a proposal to make this a function within `sphinx.ext.linkcode` -- please let me know your thoughts.

If I remember correctly, the reason I chose to add it in `app`/`Sphinx` was to imitate [`add_autodocumenter()`](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_autodocumenter), also described in [Developing autodoc extensions](https://www.sphinx-doc.org/en/master/development/tutorials/autodoc_ext.html). But I completely understand, and I agree that core Sphinx should avoid having special code for extensions, even if they are first-party.

So this is fine by me.

PR #2224, opened 2016-01-08
Fix #2060: ``:manpage:`` role uses incorrect style (node)

tk0miya:
It should be use bold style to reference man pages in man format.
On other formats, it is still represented as emphasis.

To keep style of other format, I added new node; `sphinx.addnodes.manpage`.
It makes doctree a little complicated, but it separates styles and data structure.

PR #11825, opened 2023-12-29
Allow custom target in manpage role

n-peugnet:
Subject: Allow custom target inside `<>` in `manpage` role, like in hyperlinks.

### Feature or Bugfix
- Feature

### Purpose
- Some manpages exist in multiple versions. For instance, the [`mailx.1` manpage](https://manpages.debian.org/mailx.1) has 3 conflicting versions. I would like to specify which version to link to (for instance `bsd-mailx/mailx.1`) without changing the displayed text (so keeping only `mailx.1` displayed). By making the manpage node extend node.reference and by applying the XRefRole treatment, we can now specify the target of a manpage role inside `<>` like in an hyperlink.

### Detail
- For now I only tested the rendering in LaTeX PDF and HTML builders I didn't find any issue. Except that in HTML, the FixedTextElement style is now respected (which IMO is a good thing) so the link appears like this (with RTD theme):
  **Before**:
  2023-12-29-153158_575x53_scrot

  **After**:
  2023-12-29-153007_577x60_scrot

### Relates
- Did not find any related issue.


sphinx/addnodes.py
```diff
@@ -549,7 +549,7 @@ class literal_strong(nodes.strong, not_smartquotable):
     """
 
 
-class manpage(nodes.Inline, nodes.FixedTextElement):
+class manpage(nodes.reference, nodes.FixedTextElement):
```
n-peugnet: I am not sure if this change is really needed.

n-peugnet:
> I think changing the base class of `manpage` is not the way to go. Concerning the formatting, I think we shouldn't change it for now.

I was not sure if it was required, but it seemed to make sense as the rest of the "reference" nodes are based on this class.

> You should also check with the `manpage` format itself because it creates manpage nodes which are also visited.

I will try when I have some time. From what I saw there are tests that cover this usage.

> And you need to see what happens when you set the `manpages_url` configuration value.

I already checked this, as it is my main use case. It is working the same way as before.

sphinx/addnodes.py
n-peugnet: > I think changing the base class of `manpage` is not the way to go. Concerning the formatting, I think we shouldn't change it for now.

@picnixz: I tried reverting this change but I get the same formatting results in HTML. It must come from something else.
We can also wait for another opinion on the formatting changes, and if it is OK I can modify the expected results in the tests, and add a changelog entry etc.

Otherwise I don't really know where to look for this formatting change in the code. I will continue to try some things but I don't know if I will manage to keep the existing HTML formatting.

picnixz: Would it be ok for you to postpone it for a few weeks/months? because I won't have time for in-depth review before march/april. 

If anyone else has time to dig or wants to discuss feel free to ignore my opinion as I don't mind both formats actually (just that I want to keep changes minimal if possible, especially if it concerns base classes)

n-peugnet:
> > You should also check with the `manpage` format itself because it creates manpage nodes which are also visited.
> 
> I will try when I have some time. From what I saw there are tests that cover this usage.

Just for the record, here is the test that checks this role in `manpage` output: https://github.com/sphinx-doc/sphinx/blob/68295a5a78933d4613a923a11cd0590be6c448d4/tests/test_build_manpage.py#L17

Which has been added in https://github.com/sphinx-doc/sphinx/pull/2224

doc/usage/restructuredtext/roles.rst
```diff
@@ -175,6 +175,9 @@ different style:
    ``:manpage:`ls(1)``` displays :manpage:`ls(1)`. Creates a hyperlink to an
    external site rendering the manpage if :confval:`manpages_url` is defined.
 
+   .. versionchanged:: 7.3
+      Allowed to specify a custom target inside ``<>`` like hyperlinks.
```
AA-Turner:
```suggestion
      Allow specifying a target with ``<>``, like hyperlinks.
      For example, ``:manpage:`blah <ls(1)>```displays :manpage:`blah <ls(1)>`.
```

AA-Turner:
Thanks Nicolas!

A

n-peugnet:
Thank you @AA-Turner, but it is still missing the changelog entry that I didn't add.

PR #11828, opened 2023-12-31
Enable MyPy error codes

DimitriPapadopoulos:
Subject: Enable MyPy error codes

### Feature or Bugfix

- Feature

### Purpose
Enable MyPy error codes as suggested by [Repo-Review](https://learn.scientific-python.org/development/guides/repo-review/?repo=sphinx-doc%2Fsphinx&branch=master).

### Detail
* [MY104](https://learn.scientific-python.org/development/guides/style#MY104): MyPy enables ignore-without-code
  Must have `"ignore-without-code"` in `enable_error_code = [...]`. This will force all skips in your project to include the error code, which makes them more readable, and avoids skipping something unintended.
  ```ini
  [tool.mypy]
  enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
* [MY105](https://learn.scientific-python.org/development/guides/style#MY105): MyPy enables redundant-expr
  Must have `"redundant-expr"` in `enable_error_code = [...]`. This helps catch useless lines of code, like checking the same condition twice.
  ```ini
  [tool.mypy]
  enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
  ```
* [MY106](https://learn.scientific-python.org/development/guides/style#MY106): MyPy enables truthy-bool
  Must have `"truthy-bool"` in `enable_error_code = []`. This catches mistakes in using a value as truthy if it cannot be falsey.
  ```ini
  [tool.mypy]
  enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
  ```

### Relates
- [Repo-Review](https://learn.scientific-python.org/development/guides/repo-review/?repo=sphinx-doc%2Fsphinx&branch=master)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -221,6 +221,7 @@ enable_error_code = [
     "redundant-self",
     "truthy-iterable",
     "ignore-without-code",
+    "redundant-expr",
     "unused-awaitable",
 ]
 disable_error_code = [
--- a/sphinx/domains/python/_annotations.py
+++ b/sphinx/domains/python/_annotations.py
@@ -232,13 +232,14 @@ class _TypeParameterListParser(TokenProcessor):
         while current := self.fetch_token():
             if current == token.NAME:
                 tp_name = current.value.strip()
+                tp_kind: Parameter.kind
                 if self.previous and self.previous.match([token.OP, '*'], [token.OP, '**']):
                     if self.previous == [token.OP, '*']:
                         tp_kind = Parameter.VAR_POSITIONAL
                     else:
-                        tp_kind = Parameter.VAR_KEYWORD  # type: ignore[assignment]
+                        tp_kind = Parameter.VAR_KEYWORD
                 else:
-                    tp_kind = Parameter.POSITIONAL_OR_KEYWORD  # type: ignore[assignment]
+                    tp_kind = Parameter.POSITIONAL_OR_KEYWORD
 
                 tp_ann: Any = Parameter.empty
                 tp_default: Any = Parameter.empty


picnixz:
That's interesting. Could we perhaps investigate each failure separately? or as a common issue with multiple tasks? maybe there are some bugs we are mot aware of.

DimitriPapadopoulos:
I can try, but I would need input is some cases. For example, when MyPy documents the type of `s` to be `str`, the Python code might add `if s is not None` just in case or cast with `str(s)`. Should we:
1. Keep the defensive code and add a `# type: ignore`?
2. Remove the defensive code and rely on static typing?
3. Change the documented type of s from `str` to `str | None`?

picnixz:
First, let's not break anything, namely if you are not sure, leave it as it is (ie, make mypy fail). 

Afterwards, if we have some redundant check, just ignore the error for now (we probably need a dedicated investigation to see whether mypy is right or not). 

I don't know how much are easy to fix (ie the false positives) but I think it's worth to know what may be wrong or not and whether we can safely ignore the error or not.

DimitriPapadopoulos:
I think there are no false positives in this PR, unlike #11831.

picnixz:
The thing is that we probably don't want a subclass of `str` hence the redundant cast. I didn't look where the errors occur but there are some classes that behaved like strings for compatibility reasons without inheriting from it (all those pathlike classes).

DimitriPapadopoulos:
I see. But then, would MyPy static typing qualify pathlike classes as `str`?

picnixz:
That I don't know actually since we might have only put the str annotation. But it's worth checking whether the cast is redundant or not

danieleades:
here be dragons.

I'm not sure we're in a position to *fully* trust mypy's resolution of types in this repo. That's not to disparage mypy, but there are a lot of ignore error codes and `# type: ignore`s in this repo which will confound the resolution.

I had a quick skim of the errors, and they certainly don't *look* like false positives.

What's the best way forward- split the fixes into a handful of diffs for review?

DimitriPapadopoulos:
How to split? By error type ([MY104](https://learn.scientific-python.org/development/guides/style#MY104), [MY105](https://learn.scientific-python.org/development/guides/style#MY105), [MY106](https://learn.scientific-python.org/development/guides/style#MY106))?

danieleades:
> How to split? By error type ([MY104](https://learn.scientific-python.org/development/guides/style#MY104), [MY105](https://learn.scientific-python.org/development/guides/style#MY105), [MY106](https://learn.scientific-python.org/development/guides/style#MY106))?

Seems logical enough

danieleades:
> [0a76199](https://github.com/sphinx-doc/sphinx/commit/0a76199f994ea07ad8429fab3e1798fe01cdde80) enabled "ignore-without-code" ([MY104](https://learn.scientific-python.org/development/guides/style#MY104)) last week.
> 
> Let's try "redundant-expr" ([MY105](https://learn.scientific-python.org/development/guides/style#MY105)).

are you also intending to apply fixes in this PR?

danieleades:
You can resolve the following one by adding a type declaration (and removing the 'type ignore' statements)

```python
tp_kind: Parameter.VAR_POSITIONAL | Parameter.VAR_KEYWORD | Parameter.POSITIONAL_OR_KEYWORD  # noqa: E501
```

> sphinx/domains/python/_annotations.py: note: In member "parse" of class "_TypeParameterListParser":
> sphinx/domains/python/_annotations.py:256:20: error: Left operand of "and" is always true  [redundant-expr]

haven't looked at these next few yet

> sphinx/ext/autodoc/__init__.py: note: In member "filter_members" of class "Documenter":
> sphinx/ext/autodoc/__init__.py:717:20: error: Left operand of "and" is always true  [redundant-expr]
> sphinx/ext/autodoc/__init__.py:780:20: error: Left operand of "and" is always true  [redundant-expr]
> sphinx/builders/html/__init__.py: note: In function "handle_page":
> sphinx/builders/html/__init__.py:1052:25: error: If condition in comprehension is always true  [redundant-expr]
> sphinx/builders/html/__init__.py: note: In function "css_tag":
> sphinx/builders/html/__init__.py:1052:25: error: If condition in comprehension is always true  [redundant-expr]
> sphinx/domains/cpp/_parser.py: note: In member "_parse_decl_specs_simple" of class "DefinitionParser":
> sphinx/domains/cpp/_parser.py:1205:[16](https://github.com/sphinx-doc/sphinx/actions/runs/7602533873/job/20703187797?pr=11828#step:5:17): error: Left operand of "and" is always true  [redundant-expr]

these next 3 are false positives. the left operand is not a constant, it's reassigned on each loop

> sphinx/domains/cpp/_parser.py:1209:16: error: Left operand of "and" is always true  [redundant-expr]
> sphinx/domains/cpp/_parser.py:1228:16: error: Left operand of "and" is always true  [redundant-expr]
> sphinx/domains/cpp/_parser.py:1232:16: error: Left operand of "and" is always true  [redundant-expr]
> sphinx/domains/c/_parser.py: note: In member "_parse_decl_specs_simple" of class "DefinitionParser":

these next 3 are false positives. the left operand is not a constant, it's reassigned on each loop

> sphinx/domains/c/_parser.py:627:16: error: Left operand of "and" is always true  [redundant-expr]
> sphinx/domains/c/_parser.py:631:16: error: Left operand of "and" is always true  [redundant-expr]
> sphinx/domains/c/_parser.py:635:16: error: Left operand of "and" is always true  [redundant-expr]

looks like they're mostly false positives

danieleades:
> these next 3 are false positives. the left operand is not a constant, it's reassigned on each loop

you should be able to address these by changing the initial assignment of the variables from

```python
some_type = None
```
to
```python
some_type: SomeType | None = None
```

the false positive is caused by the implicit optional. You can see from the mypy config in `pyproject.toml` that implicit optional is (temporarily) allowed in those files, which confounds the type resolution for the new error codes you're adding

DimitriPapadopoulos:
I don't have much experience with MyPy. The type of `tp_kind` should be the [`Parameter.kind.description`](https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind.description) as far as I can tell, but I don't know how to express that.

danieleades:
might be a better use of energy to focus on shrinking the existing mypy whitelist (especially the implicit optional whitelist) before adding these additional error codes. Although they're useful for shining a light on places where the type inference breaks down because of the implicit optional.
Thoughts @AA-Turner?

I've been chipping away at this whitelist, but frankly don't have a good enough understanding of the cxx modules to make big inroads

AA-Turner:
Closing for now -- but we can return to this later once all of the other typing problems are resolved!

PR #11833, opened 2024-01-01
Add recent Ruff rules

DimitriPapadopoulos:
Subject: Add recent ruff rules

### Feature or Bugfix

- Feature

### Purpose
- Try new ruff rules from.

### Detail
- Depending on the CI results, we can comment them out or fix the reported issues, whatever makes more sense.

### Relates
- https://docs.astral.sh/ruff/rules/

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -313,7 +313,7 @@ select = [
     "TCH003",  # Move standard library import `{}` into a type-checking block
     "TCH004",  # Move import `{qualified_name}` out of type-checking block. Import is used for more than type hinting.
     "TCH005",  # Found empty type-checking block
-    "TCH006",  # nvalid string member in `X | Y`-style union type
+    "TCH006",  # Invalid string member in `X | Y`-style union type
     # flake8-todos ('TD')
 #    "TD001",  # Invalid TODO tag: `{tag}`
 #    "TD003",  # Missing issue link on the line following this TODO
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -188,6 +188,7 @@ select = [
 #    "Q001",  # Single quote multiline found but double quotes preferred
     "Q002",  # Single quote docstring found but double quotes preferred
     "Q003",  # Change outer quotes to avoid escaping inner quotes
+    "Q004",  # Unnecessary escape on inner quote character
     # flake8-return ('RET')
     "RET501",  # Do not explicitly `return None` in function if it is the only possible return value
     "RET502",  # Do not implicitly `return None` in function able to return non-`None` value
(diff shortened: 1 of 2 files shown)


.ruff.toml
```diff
@@ -265,6 +266,7 @@ select = [
     "S609",  # Possible wildcard injection in call due to `*` usage
     "S612",  # Use of insecure `logging.config.listen` detected
 #    "S701",  # Using jinja2 templates with `autoescape=False` is dangerous and can lead to XSS. Ensure `autoescape=True` or use the `select_autoescape` function.
+    "S702",  # Mako templates allow HTML and JavaScript rendering by default and are inherently open to XSS attacks
```
picnixz: I don't think we use Mako templates (unless you wanted to say "Make")

.ruff.toml
```diff
@@ -293,6 +295,7 @@ select = [
     "SIM300",  # Yoda conditions are discouraged, use `{suggestion}` instead
     "SIM401",  # Use `{contents}` instead of an `if` block
     "SIM910",  # Use `{expected}` instead of `{original}`
+
```
picnixz: why this space?

.ruff.toml
```diff
@@ -310,6 +313,7 @@ select = [
     "TCH003",  # Move standard library import `{}` into a type-checking block
     "TCH004",  # Move import `{qualified_name}` out of type-checking block. Import is used for more than type hinting.
     "TCH005",  # Found empty type-checking block
+    "TCH006",  # nvalid string member in `X | Y`-style union type
```
picnixz:
```suggestion
    "TCH006",  # Invalid string member in `X | Y`-style union type
```

.ruff.toml
```diff
@@ -361,6 +366,7 @@ select = [
     "UP038",  # Use `X | Y` in `{}` call instead of `(X, Y)`
     "UP039",  # Unnecessary parentheses after class definition
     "UP040",  # Type alias `{name}` uses `TypeAlias` annotation instead of the `type` keyword
+    "UP041",  # Replace aliased errors with `TimeoutError`
```
picnixz, line 368: I don't think it only affects this kind of exception, or does it?

.ruff.toml
DimitriPapadopoulos: This file `.ruff.toml` contains a complete list of rules for each category of rules. The rules in this PR are the only missing rules, and since they all are new rules, I presume they should be added for completeness and consistency. Unwanted or unused rules are commented out.

Are you suggesting commenting this rule out?

picnixz: I think we can comment it out? or maybe add a comment saying it's unused but keep it in case we use it but forget about enabling it?

.ruff.toml
DimitriPapadopoulos: That's just a copy from the ruff [Rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up) page:
> | Code  | Name                                                                            | Message                                    |       | 
> | :---  | :---                                                                            | :---                                       | :---  | 
> | UP041 | [`timeout-error-alias`](https://docs.astral.sh/ruff/rules/timeout-error-alias/) | Replace aliased errors with `TimeoutError` | 🛠️ 🧪 |

.ruff.toml
DimitriPapadopoulos: Removed

.ruff.toml
DimitriPapadopoulos: I have commented it just like the above rule "S701".

Issue #11834, opened 2024-01-01
ruff: Add new refurb (FURB) rules


DimitriPapadopoulos:
**Is your feature request related to a problem? Please describe.**

New [refurb](https://github.com/dosisod/refurb) (FURB) rules have been added to ruff, add them to file [`.ruff.toml`](https://github.com/sphinx-doc/sphinx/blob/master/.ruff.toml) in Sphinx.

**Describe the solution you'd like**

Add all rules, review the errors, fix those that make sense, suppress the rest:
<details>
<summary>541 refurb issues (ruff 0.1.7)</summary>

```
sphinx/builders/gettext.py:244:17 [FURB101]: Replace `with open(x, ...) as f: y = f.read()` with `y = Path(x).read_text(...)`
sphinx/cmd/quickstart.py:385:5 [FURB101]: Replace `with open(x, ...) as f: y = f.read()` with `y = Path(x).read_text(...)`
sphinx/environment/__init__.py:603:13 [FURB101]: Replace `with open(x, ...) as f: y = f.read()` with `y = Path(x).read_bytes()`
sphinx/ext/autosummary/generate.py:531:13 [FURB101]: Replace `with open(x, ...) as f: y = f.read()` with `y = Path(x).read_text(...)`
sphinx/ext/graphviz.py:139:17 [FURB101]: Replace `with open(x, ...) as f: y = f.read()` with `y = Path(x).read_text(...)`
sphinx/search/__init__.py:549:13 [FURB101]: Replace `with open(x, ...) as f: y = f.read()` with `y = Path(x).read_text(...)`
sphinx/search/__init__.py:551:13 [FURB101]: Replace `with open(x, ...) as f: y = f.read()` with `y = Path(x).read_text(...)`
sphinx/theming.py:47:13 [FURB103]: Replace `with open(x, ...) as f: f.write(y)` with `Path(x).write_bytes(y)`
sphinx/transforms/post_transforms/images.py:86:21 [FURB103]: Replace `with open(x, ...) as f: f.write(y)` with `Path(x).write_bytes(y)`
sphinx/transforms/post_transforms/images.py:135:9 [FURB103]: Replace `with open(x, ...) as f: f.write(y)` with `Path(x).write_bytes(y)`
sphinx/environment/collectors/dependencies.py:35:15 [FURB104]: Replace `os.getcwd()` with `Path.cwd()`
[...]

PR #11847, opened 2024-01-02
Apply refurb/ruff rule FURB140

DimitriPapadopoulos:
Subject: Apply refurb/ruff rule FURB140

### Feature or Bugfix

- Refactoring

### Purpose
Fix these issues:
```
[FURB140]: Replace `[f(...) for ... in x]` with `list(starmap(f, x))`
```
```
[FURB140]: Replace `f(...) for ... in x` with `starmap(f, x)`
```

### Relates

### DIFF
--- a/sphinx/ext/napoleon/docstring.py
+++ b/sphinx/ext/napoleon/docstring.py
@@ -7,6 +7,7 @@ import contextlib
 import inspect
 import re
 from functools import partial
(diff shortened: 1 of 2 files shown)


AA-Turner:
Although perhaps more verbose, I think I prefer the non-`starmap` versions here.

A

PR #11848, opened 2024-01-02
Apply refurb/ruff rule FURB145

DimitriPapadopoulos:
Subject: Apply refurb/ruff rule FURB145

### Feature or Bugfix

- Refactoring

### Purpose
Fix this issue:
```
[FURB145]: Replace `x[:]` with `x.copy()`
```

### Relates

### DIFF
--- a/sphinx/domains/changeset.py
+++ b/sphinx/domains/changeset.py
@@ -126,7 +126,7 @@ class ChangeSetDomain(Domain):
 
     def clear_doc(self, docname: str) -> None:
         for changes in self.changesets.values():
-            for changeset in changes[:]:
+            for changeset in changes.copy():
                 if changeset.docname == docname:
                     changes.remove(changeset)
 
--- a/sphinx/events.py
+++ b/sphinx/events.py
@@ -77,7 +77,7 @@ class EventManager:
     def disconnect(self, listener_id: int) -> None:
         """Disconnect a handler."""
         for listeners in self.listeners.values():
-            for listener in listeners[:]:
+            for listener in listeners.copy():
                 if listener.id == listener_id:
                     listeners.remove(listener)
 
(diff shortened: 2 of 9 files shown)


picnixz:
I personally prefer the `[:]` syntax because it's more concise.

DimitriPapadopoulos:
The purpose of the code is to perform shallow copy, so using `copy()` instead of `[:]` makes the intent clear.

Besides, that's what the Python documentation suggests in the first place in [`copy` — Shallow and deep copy operations](https://docs.python.org/3/library/copy.html#module-copy).

If you think about it, Van Rossum based Python on basic English keywords that provide clear syntax and uncluttered code, not concision. If you prefer concision over readability and uncluttered code, try Perl :smile: Years ago, I used to program scripts in Perl. After six months, I tried to modify one of my own scripts. Although the code was heavily commented, it used lots of the concise Perl expressions. This exercise convinced me that Perl should not be used for but for short one-shot scripts, and that choosing concision over readability and uncluttered code is not a good choice it in the long term.

picnixz:
Well, we could always stick to that but even CPython does not enforce it (I think there are more occurrences of `[:]` for list copy rather than `copy()`) (personally, I've seen more the `[:]` to shallow copy lists rather than `.copy()`).

Anyway, it'd be good if another maintainer gives its opinion on this one.

DimitriPapadopoulos:
Habits die hard, I guess. The `copy()` function is a Python 3 addition. The mere fact that it has been added, although you could already use `[:]` or `list()`, tells us something...

See for example this blog: [Python: How to Copy a List? (The Idiomatic Way)](https://www.afternerd.com/blog/python-copy-list/)

electric-coder:
> `copy()`

builtin function, unambiguous.

> `[:]`

ok, I think it's a slice but what does it do?!

AA-Turner (approved):

PR #11849, opened 2024-01-02
Apply refurb/ruff rule FURB149

DimitriPapadopoulos:
Subject: Apply refurb/ruff rule FURB149

### Feature or Bugfix

- Refactoring

### Purpose
Fix these issues:
```
[FURB149]: Replace `x is True` with `x`
```
```
[FURB149]: Replace `x is False` with `not x`
```

### Relates

### DIFF
--- a/sphinx/directives/code.py
+++ b/sphinx/directives/code.py
@@ -319,7 +319,7 @@ class LiteralIncludeReader:
 
                         return lines[lineno:]
 
-            if inclusive is True:
+            if inclusive:
                 raise ValueError('start-after pattern not found: %s' % start)
             else:
                 raise ValueError('start-at pattern not found: %s' % start)
@@ -348,7 +348,7 @@ class LiteralIncludeReader:
                             pass  # end-before ignores first line
                         else:
                             return lines[:lineno]
-            if inclusive is True:
+            if inclusive:
                 raise ValueError('end-at pattern not found: %s' % end)
             else:
                 raise ValueError('end-before pattern not found: %s' % end)
--- a/sphinx/ext/autosummary/generate.py
+++ b/sphinx/ext/autosummary/generate.py
@@ -237,7 +237,7 @@ class ModuleScanner:
                 # list all members up
                 imported_members
                 # list not-imported members
-                or imported is False
+                or not imported
                 # list members that have __all__ set
                 or (respect_module_all and '__all__' in dir(self.object))
             ):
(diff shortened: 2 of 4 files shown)


DimitriPapadopoulos:
I knew this wouldn't work as soon as I noticed constructs such as:
```python
if x is True:
    ...
elif x is False:
    ...
else:
    ...
```
It's probably a bad idea to have such objects or tests, but fixing such issues would require more substantial work. I'll try to keep only the straightforward cases and discard other bad fixes for now.

AA-Turner (approved):
I've updated the ``skipped`` type to be `True | False | None` -- other than that all good.

A

chrisjsewell (approved):
Heya, should the rule not also activated in our ruff configuration?

DimitriPapadopoulos:
Yes, I'm applying the rules for now. It's just a couple PRs and then I can add FURB. I don't feel like adding FURB rules one by one in pyproject.toml and eventually replace them with a single FURB.

chrisjsewell:
> I'm just applying the rules for now. Just a couple PRs away from adding FURB to pyproject.toml. I don't feel like adding FURB rules one by one in pyproject.toml and eventually replace them with a single FURB.

`FURB` is already selected though: https://github.com/sphinx-doc/sphinx/blob/cb8a28dd7e11a7a0a5bc9694163fc164433e2f36/.ruff.toml#L126

From I can see from https://docs.astral.sh/ruff/rules/#refurb-furb, there is not actually a `FURB149` implemented in ruff 😅 

In fact, it was specifically not added: https://github.com/astral-sh/ruff/pull/9539

I would prefer we only fix these things, at the same time the rule is actually added to ruff, otherwise it is not clear if it is a correct change or not 😬

DimitriPapadopoulos:
Actually `FURB149` is available in preview mode:

```console
$ ruff check --select FURB142 --preview
sphinx/builders/__init__.py:317:9: FURB142 [*] Use of `set.add()` in a for loop
    |
315 |           doccount = len(updated_docnames)
316 |           logger.info(bold(__('looking for now-outdated files... ')), nonl=True)
317 |           for docname in self.env.check_dependents(self.app, updated_docnames):
    |  _________^
318 | |             updated_docnames.add(docname)
    | |_________________________________________^ FURB142
319 |           outdated = len(updated_docnames) - doccount
320 |           if outdated:
    |
    = help: Replace with `.update()`

Found 1 error.
[*] 1 fixable with the `--fix` option.
$ 
```

It should land in standard mode in upcoming versions of ruff. I am applying the rule so that the issue doesn't arise in the future. Additionally, the fix makes total sense.

DimitriPapadopoulos:
You're right about this one. I guess I had run refurb instead of ruff at some point. Additionally, these fixes are not really helpful.

PR #11856, opened 2024-01-04
Apply refurb/ruff rule FURB113

DimitriPapadopoulos:
Subject: Apply refurb/ruff rule FURB113

### Feature or Bugfix

- Refactoring

### Purpose
Fix some of these issues:
```
[FURB113]: Use `x.extend(...)` instead of repeatedly calling `x.append()`
```
Leave out these files for now, as most occurrences are probably more readable in their current state:
* `sphinx/domains/c.py`
* `sphinx/domains/cpp.py`

### Details

* Not sure about these changes. Sure, the number of function calls is reduced. However, readability is not always improved. Unless, the function call happens in a tight loop, the optimisation might not be worth it, unless perhaps in specific cases. Thoughts?

### Relates
https://github.com/sphinx-doc/sphinx/issues/11834

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -465,7 +465,7 @@ select = [
 # allow print() in the tutorial
 "doc/development/tutorials/examples/recipe.py" = ["T201"]
 "sphinx/domains/**" = ["FURB113"]
-"tests/test_domain_cpp.py" = ["FURB113"]
+"tests/test_domains/test_domain_cpp.py" = ["FURB113"]
 
 # from .flake8
 "sphinx/*" = ["E241"]
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -125,7 +125,7 @@ select = [
     # refurb ('FURB')
 #    "FURB101",  # `open` and `read` should be replaced by `Path({filename}).{suggestion}`
 #    "FURB105",  # Unnecessary empty string passed to `print`
-#    "FURB113",  # Use `{suggestion}` instead of repeatedly calling `{name}.append()`
+    "FURB113",  # Use `{suggestion}` instead of repeatedly calling `{name}.append()`
     "FURB118",  # Use `operator.{operator}` instead of defining a function
     "FURB131",  # Prefer `clear` over deleting a full slice
     "FURB132",  # Use `{suggestion}` instead of check and `remove`
@@ -464,6 +464,8 @@ select = [
 "doc/development/tutorials/examples/*" = ["INP001"]
 # allow print() in the tutorial
 "doc/development/tutorials/examples/recipe.py" = ["T201"]
+"sphinx/domains/**" = ["FURB113"]
+"tests/test_domain_cpp.py" = ["FURB113"]
 
 # from .flake8
 "sphinx/*" = ["E241"]
--- a/sphinx/builders/latex/__init__.py
+++ b/sphinx/builders/latex/__init__.py
@@ -375,9 +375,11 @@ class LaTeXBuilder(Builder):
             for subdir, title in self.titles:
                 if docname.startswith(subdir):
                     newnodes.extend(
-                        nodes.Text(_(' (in ')),
-                        nodes.emphasis(title, title),
-                        nodes.Text(')'),
+                        (
+                            nodes.Text(_(' (in ')),
+                            nodes.emphasis(title, title),
+                            nodes.Text(')'),
+                        )
                     )
                     break
             else:
--- a/sphinx/builders/latex/__init__.py
+++ b/sphinx/builders/latex/__init__.py
@@ -374,13 +374,11 @@ class LaTeXBuilder(Builder):
             newnodes: list[Node] = [nodes.emphasis(sectname, sectname)]
             for subdir, title in self.titles:
                 if docname.startswith(subdir):
-                    newnodes.extend(
-                        (
-                            nodes.Text(_(' (in ')),
-                            nodes.emphasis(title, title),
-                            nodes.Text(')'),
-                        )
-                    )
+                    newnodes.extend((
+                        nodes.Text(_(' (in ')),
+                        nodes.emphasis(title, title),
+                        nodes.Text(')'),
+                    ))
                     break
             else:
                 pass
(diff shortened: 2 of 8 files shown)


sphinx/search/ja.py
```diff
@@ -420,15 +420,10 @@ def split(self, input: str) -> list[str]:
...
-        seg.append('E1')
-        seg.append('E2')
-        seg.append('E3')
-        ctype.append('O')
-        ctype.append('O')
-        ctype.append('O')
+        seg.extend(input)
+        ctype.extend(self.ctype_(t) for t in input)
+        seg.extend(('E1', 'E2', 'E3'))
+        ctype.extend(('O', 'O', 'O'))
```
danieleades: can you do better here?

```python
seg = ['B3', 'B2', 'B1', **input, 'E1', 'E2', 'E3']
ctype = ['O', 'O', 'O', **(self.ctype_(t) for t in input), 'O', 'O', 'O']
```
```

Comment (DimitriPapadopoulos) [reply sphinx/search/ja.py]: Done, thanks for the tip.

File: sphinx/search/ja.py [hunk 2]
```diff
@@ -418,17 +418,8 @@ def split(self, input: str) -> list[str]:
...
-            seg.append(t)
-            ctype.append(self.ctype_(t))
-        seg.append('E1')
-        seg.append('E2')
-        seg.append('E3')
-        ctype.append('O')
-        ctype.append('O')
-        ctype.append('O')
+        seg = ['B3', 'B2', 'B1', *input, 'E1', 'E2', 'E3']
+        ctype = ['O', 'O', 'O', *(self.ctype_(t) for t in input), 'O', 'O', 'O']
```
AA-Turner:
```suggestion
        ctype = ['O', 'O', 'O', *map(self.ctype_, input), 'O', 'O', 'O']
```

sphinx/builders/texinfo.py
```diff
@@ -165,9 +165,13 @@ def assemble_doctree(
...
-                    newnodes.append(nodes.Text(_(' (in ')))
-                    newnodes.append(nodes.emphasis(title, title))
-                    newnodes.append(nodes.Text(')'))
+                    newnodes.extend(
+                        (
+                            nodes.Text(_(' (in ')),
+                            nodes.emphasis(title, title),
+                            nodes.Text(')'),
+                        )
+                    )
```
AA-Turner:
```suggestion
                    newnodes.extend((
                        nodes.Text(_(' (in ')),
                        nodes.emphasis(title, title),
                        nodes.Text(')'),
                    ))
```

sphinx/builders/latex/__init__.py
```diff
@@ -374,9 +374,13 @@ def assemble_doctree(
...
-                    newnodes.append(nodes.Text(_(' (in ')))
-                    newnodes.append(nodes.emphasis(title, title))
-                    newnodes.append(nodes.Text(')'))
+                    newnodes.extend(
+                        (
+                            nodes.Text(_(' (in ')),
+                            nodes.emphasis(title, title),
+                            nodes.Text(')'),
+                        )
+                    )
```
AA-Turner:
```suggestion
                    newnodes.extend((
                        nodes.Text(_(' (in ')),
                        nodes.emphasis(title, title),
                        nodes.Text(')'),
                    ))
```

sphinx/ext/inheritance_diagram.py
```diff
@@ -306,9 +306,10 @@ def generate_dot(self, name: str, urls: dict[str, str] | None = None,
             n_attrs.update(env.config.inheritance_node_attrs)
             e_attrs.update(env.config.inheritance_edge_attrs)
 
-        res: list[str] = []
-        res.append('digraph %s {\n' % name)
-        res.append(self._format_graph_attrs(g_attrs))
+        res: list[str] = [
+            'digraph %s {\n' % name,
```
AA-Turner:
```suggestion
            f'digraph {name} {{\n',
```

DimitriPapadopoulos: I was thinking of changing residual traditional string interpolation and format() to f-strings in a different PR, but perhaps there aren't so many of them.

sphinx/search/ja.py
DimitriPapadopoulos: Even better, thanks!

danieleades (changes requested):
should the rule not be added to `ruff.toml`?

DimitriPapadopoulos:
Ideally, yes. However, I am not certain this rule should be applied to all files because I find the result less readable for some files, and I haven't applied it to these files:
* sphinx/domains/c.py
* sphinx/domains/cpp.py

I have found how to ignore a specific rule in specific files, I'll give it a try:
https://docs.astral.sh/ruff/configuration/#configuring-ruff

danieleades:
> Ideally, yes. However, I am not certain this rule should be applied to all files because I find the result less readable for some files, and I haven't applied it to these files:
> 
> * sphinx/domains/c.py
> * sphinx/domains/cpp.py
> 
> I have found how to ignore a specific rule in specific files, I'll give it a try: https://docs.astral.sh/ruff/configuration/#configuring-ruff

I don't want to speak for the maintainers, but here's an approach that I would find useful-

- add the rule in this PR and whitelist the files where it results in less readable code
- create a new PR removing the whitelist and applying the "less readable" changes to the other files for discussion

it could be that where this rule results in less readable code, a larger refactor can result in something better (as per some of the examples in this PR)

AA-Turner (approved):

DimitriPapadopoulos:
Actually the changes suggested by refurb are not always as bad as I thought they were. Let me give your method a try though, I find it handy because there have been extensive changes `sphinx/domains` that would require to rebase this branch before proceeding.

.ruff.toml
```diff
@@ -464,6 +464,8 @@ select = [
 "doc/development/tutorials/examples/*" = ["INP001"]
 # allow print() in the tutorial
 "doc/development/tutorials/examples/recipe.py" = ["T201"]
+"sphinx/domains/**" = ["FURB113"]
+"tests/test_domain_cpp.py" = ["FURB113"]
```
danieleades:
```suggestion
"tests/test_domains/test_domain_cpp.py" = ["FURB113"]
```
?

danieleades (approved):

PR #11865, opened 2024-01-09
Draft: linkcheck builder: refactor status codes into an enum

jayaddison:
### Feature or Bugfix
- Refactoring

### Purpose
- Use `StrEnum` values to provide type-checkable linkcheck status codes.

### Detail
- Requires Python 3.11, for [`StrEnum` support](https://docs.python.org/3.11/whatsnew/3.11.html#enum).
- Removes an apparently unused status codes (`local`, `old`).

### Relates
- N/A


AA-Turner:
We can't adopt this until 2025 (https://www.sphinx-doc.org/en/master/internals/release-process.html#python-version-support-policy). Would just using `class StrEnum(str, Enum): ...` work?

https://github.com/python/cpython/blob/3.11/Lib/enum.py#L1273-L1303

A

sphinx/builders/linkcheck.py
```diff
@@ -99,61 +110,59 @@ def process_result(self, result: CheckResult) -> None:
                     'info': result.message}
         self.write_linkstat(linkstat)
 
-        if result.status == 'unchecked':
-            return
-        if result.status == 'working' and result.message == 'old':
-            return
+        match result.status:
```
AA-Turner, line 122: We can't adopt `match` blocks as we still support Python 3.9 until at least April.

jayaddison:
> We can't adopt this until 2025 (https://www.sphinx-doc.org/en/master/internals/release-process.html#python-version-support-policy). Would just using `class StrEnum(str, Enum): ...` work?
> 
> https://github.com/python/cpython/blob/3.11/Lib/enum.py#L1273-L1303

Thanks - from some testing against both py3.9 and py3.11: that does seem workable, yep.

Doing so introduces a minor forwards-compatibility unit test failure that can be solved by referring to `result.status.value` instead of `result.status`; that's probably a good idea anyway so that the intended output is less ambiguous.

I'll likely push that change and some small related fixups/cleanups tomorrow.

sphinx/builders/linkcheck.py
```diff
@@ -98,74 +115,75 @@ def process_result(self, result: CheckResult) -> None:
...
-                logger.info(color('redirect  ') + result.uri +
-                            color(' - ' + text + ' to ' + result.message))
-            self.write_entry('redirected ' + text, result.docname, filename,
-                             result.lineno, result.uri + ' to ' + result.message)
-        else:
-            raise ValueError('Unknown status %s.' % result.status)
+
+        match result.status:
+            case LinkStatus.IGNORED:
+                msg = darkgray('-ignored- ') + result.uri + f': {result.message}' if result.message else ''  # NoQA: E501
```
jayaddison, line 131: As hinted by the appearance of `E501`: some of these output message constructions are a bit lengthy.  I'm considering extracting a small helper function to attempt to wrap the intended behaviour.

sphinx/builders/linkcheck.py
jayaddison: Do you think it's worth merging the `StrEnum` usage without the `match` statements?  I'm on the fence about it, so I'd be OK waiting until Py3.9 is our baseline.

electric-coder: > Do you think it's worth merging the `StrEnum` usage without the `match` statements?

`Enum` with `match` is always worth it. You can get the same result by using an `Enum` that inherits from `str`, see [this example by Ethan Furman][1] and the [canonical SO question][2].

[1]: https://stackoverflow.com/a/69457785
[2]: https://stackoverflow.com/a/58608362

jayaddison:
Closing this to reduce the number of pull requests open (both for Sphinx, and also in my own backlog).  I'll revisit this next year.

PR #11878, opened 2024-01-13
Enable some doc comment lints (D*)

danieleades:


AA-Turner:
@danieleades -- please can you avoid force-pushing if possible, it makes reviewing more challenging.

danieleades:
> @danieleades -- please can you avoid force-pushing if possible, it makes reviewing more challenging.

sure thing

AA-Turner:
Dan -- out of interest would you be interested in having commit permissions on Sphinx? The work you've done on mypy/ruff is really appreciated (even if we disagree on autoformatters)

A

danieleades:
> Dan -- out of interest would you be interested in having commit permissions on Sphinx?

I'm flattered that you're offering. I'm inclined to decline however- i'm enjoying tinkering away, but the ability to commit changes myself is not really a blocker for me, since i wouldn't want to commit any changes that hadn't passed consensus *anyway*. Thank you though.

> even if we disagree on autoformatters!

nothing wrong with a bit of healthy, constructive disagreement!

AA-Turner:
Fair enough; do let me know if you reconsider in the future, though. Thanks again for your support for Sphinx!

A

PR #11880, opened 2024-01-14
run coverage check on PRs

danieleades:
run coverage checks on PRs so codecov can comment on the coverage change

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,31 @@
+name: Coverage
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+jobs:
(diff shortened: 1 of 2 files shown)


danieleades:
@AA-Turner i find it strange that coverage is not configured to run on PRs since that prevents codecov from reporting the coverage diff. Is there a reason its set up this way?

Also strange that the tests are passing but the coverage job is not passing in this PR... any ideas?

AA-Turner:
We disabled this on PRs due to time and that coverage information isn't used by anyone currently.

danieleades:
> coverage information isn't used by anyone currently.

Well of course it isn't used by anyone, it isn't generated!

PR #11885, opened 2024-01-16
Closes #11884 fix searchtools.js: docContent.textContent

### LINKED ISSUE #11884: Error in searchtools.js: Uncaught (in promise) TypeError: Cannot read properties of null (reading 'textContent')
### Error:

There is an error when using the search if there is no `[role="main"]` element on the page:
> searchtools.js:159: Uncaught (in promise) TypeError: Cannot read properties of null (reading 'textContent')

### Expected:

No error, but a console warning:
> "Content block not found. Sphinx search tries to obtain it via '[role=main]'. Could you check your theme or template."

### Details:

In [searchtools.js](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/static/searchtools.js#L166), if `htmlElement.querySelector('[role="main"]');` doesn't find anything, then the [`querySelector`](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector) returns `null`. This causes the error in the next line `if (docContent !== undefined) return docContent.textContent;`

I guess this error was introduced in [commit 3b01fbe2ad](https://github.com/sphinx-doc/sphinx/commit/3b01fbe2adf3077cad5c2cf345c6d000d429d7ac#diff-52f537c2146ef4388e77c5e7616aecdabb7fc09454d03694ae238551b9fc1286R148), before which `htmlElement.find()` was used, which would return `undefined` if nothing found.

### How to Reproduce

(Sorry, I found this error in a project that is not mine, so I can't debug it properly, but I think the error is obvious enough in the code.)

- Have no `[role="main"]` element on the page
- search something using the search input box on the page

### Environment Information

```text
I don't know the environment
```

### Sphinx extensions

_No response_

### Additional context

_No response_

kkntl:
Subject: fix bug #11884 (searchtools.js, `docContent` would be `null` instead of `undefined`)

### Feature or Bugfix
- Bugfix

### Purpose
- Check `docContent`, returned by `document.querySelector`, for "truthy" instead of `!== undefined`.

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -164,7 +164,7 @@ const Search = {
     const htmlElement = new DOMParser().parseFromString(htmlString, 'text/html');
     htmlElement.querySelectorAll(".headerlink").forEach((el) => { el.remove() });
     const docContent = htmlElement.querySelector('[role="main"]');
-    if (docContent !== undefined) return docContent.textContent;
+    if (docContent) return docContent.textContent;
     console.warn(
       "Content block not found. Sphinx search tries to obtain it via '[role=main]'. Could you check your theme or template."
     );


AA-Turner:
Is it possible to add a test for this?

sphinx/themes/basic/static/searchtools.js
```diff
@@ -164,7 +164,7 @@ const Search = {
     const htmlElement = new DOMParser().parseFromString(htmlString, 'text/html');
     htmlElement.querySelectorAll(".headerlink").forEach((el) => { el.remove() });
     const docContent = htmlElement.querySelector('[role="main"]');
-    if (docContent !== undefined) return docContent.textContent;
+    if (docContent) return docContent.textContent;
```
jayaddison, line 167: This looks good to me, and seems consistent with our validation of `querySelector` results in `doctools.js` here: https://github.com/sphinx-doc/sphinx/blob/80d5396f9d434d7c71d6f7ac5fa727a972b69584/sphinx/themes/basic/static/doctools.js#L124-L125

PR #11891, opened 2024-01-18
Load themes through iteration rather than recursion

AA-Turner:
To prepare for allowing TOML-based theme definitions.

### DIFF
--- a/sphinx/builders/html/__init__.py
+++ b/sphinx/builders/html/__init__.py
@@ -884,7 +884,7 @@ class StandaloneHTMLBuilder(Builder):
     def cleanup(self) -> None:
         # clean up theme stuff
         if self.theme:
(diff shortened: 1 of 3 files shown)


tacaswell:
This seems to have broken themes that inherent from other third-party themes:

```
Theme error:
The 'mpl_sphinx_theme' theme inherits from 'pydata_sphinx_theme', which is not a loaded theme. Loaded themes are: agogo, alabaster, basic, bizstyle, classic, default, epub, haiku, mpl_sphinx_theme, nature, nonav, pyramid, scrolls, sphinxdoc, traditional
```

From the comment I assume this means that the default branch is currently in a bit of flux.   Is this something that you expected to break for now (and will work again in the future) or something that will require changes on the user side?

PR #11894, opened 2024-01-18
Do not add checksums to css files if building using the htmlhelp builder

### LINKED ISSUE #11889: HTMLHELP Relative URL strings must not have a query component
### Describe the bug

With the implementation of #11415 checksums are added to css file links in the html output. But in the htmlhelp output format the windows help viewer fails to load the stylesheets. If I remove the query component before compiling the chm file it works.

### How to Reproduce

1. Use any sphinx project and run it with the output format htmlhelp.
2. Compile the .hpp file with the Microsoft HTML HELP Workshop.
3. Open the chm file and there are no styles.

### Environment Information

```text
Platform:              win32; (Windows-11-10.0.22631-SP0)
Python version:        3.12.1 (tags/v3.12.1:2305ca5, Dec  7 2023, 22:03:25) [MSC v.1937 64 bit (AMD64)])
Sphinx version:        7.2.6
Docutils version:      0.20.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

### Additional context
It's a similar issue to #11598 where it does not work with epub.
I can provide a pull request to fix this if you are interested.

mkay-sh:
Subject: The windows help compiler requires that css links don't have a query component. This PR disables the checksum for css files if the htmlhelp builder is used.

### Feature or Bugfix
- Bugfix

### Relates
-  Fixes #11889


AA-Turner:
Please could you add an entry to CHANGES?

jayaddison:
This looks like it affects Sphinx 7.1.0 onwards currently, and I do think that this would solve the problem.

However: it's not great to introduce builder-specific fixes in Sphinx itself, and I'd suggest that maybe adding a `crc32_checksum` attribute (default `True`, overridden as `False` in the [`sphinxcontrib-htmlhelp`](https://github.com/sphinx-doc/sphinxcontrib-htmlhelp/) builder) to the base HTML builder -- since that's the top-level at which this feature is relevant -- could be an alternative.  (I considered a generic `checksum` attribute name -- but I think that there are other checksum/integrity methods we might want to consider in future, hence trying to avoid a future naming conflict)

That said: that would require some release co-ordination (addition + release in Sphinx, then a corresponding usage + baseline Sphinx dependency bump + release of `sphinxcontrib-htmlhelp`) - so so it's possible that I'm suggesting more work than is practically worthwhile.

jayaddison:
Ok: now I understand that as of recently, many of the `sphinxcontrib-*` extensions no longer have Python package dependency references back to Sphinx itself.  So in the short term, my alternative suggestion (implement an attribute in Sphinx, release, update the baseline dep in the extension, then release that) is not valid.

I'll try to learn a bit more about the reasoning behind the dependency versioning change (#11567); to my mind, the ability to list supported Sphinx versions from each extensions seems quite important.

jayaddison:
I think my previous comments here were too hasty and cautious about this - I now think it's a reasonable solution to a genuine problem, and that unless another approach arrives (that I could and should offer myself if I find time), on balance it's better to fix the existing issue.  My apologies for delay caused.

mkay-sh:
> Please could you add an entry to CHANGES?

I added an entry to the changes. I hope it is okay like this.

picnixz:
Thank you! I'll merge it when the tests are done

picnixz:
Thank you!

PR #11898, opened 2024-01-20
shrink mypy whitelist

danieleades:


sphinx/util/inspect.py
```diff
@@ -598,15 +599,19 @@ def signature(subject: Callable, bound_method: bool = False, type_aliases: dict
...
 
-def evaluate_signature(sig: inspect.Signature, globalns: dict | None = None,
-                       localns: dict | None = None,
+def evaluate_signature(sig: inspect.Signature, globalns: dict[str, object] | None = None,
+                       localns: dict[str, object] | None = None,
                        ) -> inspect.Signature:
     """Evaluate unresolved type annotations in a signature object."""
-    def evaluate_forwardref(ref: ForwardRef, globalns: dict, localns: dict) -> Any:
+    def evaluate_forwardref(
+        ref: ForwardRef, globalns: dict[str, object] | None, localns: dict[str, object] | None,
```
jayaddison: This cluster of three methods (`evaluate_signature`, `evaluate_forwardref`, `evaluate`) is the only part I'm uncertain about so far, because `ForwardRef._evaluate` seems to be an undocumented method.

jayaddison (approved):
All looks OK to me - I didn't find any mismatched signatures, although the `ForwardRef` (`localns`, `globalns`) arguments I wasn't sure about.

sphinx/util/inspect.py
danieleades: yeah it's a weird one.

here's the signature of `ForwardRef._evaluate`

```python
    def _evaluate(self, globalns, localns, recursive_guard):
        if self.__forward_arg__ in recursive_guard:
            return self
        if not self.__forward_evaluated__ or localns is not globalns:
            if globalns is None and localns is None:
                globalns = localns = {}
            elif globalns is None:
                globalns = localns
            elif localns is None:
                localns = globalns
            if self.__forward_module__ is not None:
                globalns = getattr(
                    sys.modules.get(self.__forward_module__, None), '__dict__', globalns
                )
            type_ = _type_check(
                eval(self.__forward_code__, globalns, localns),
                "Forward references must evaluate to types.",
                is_argument=self.__forward_is_argument__,
                allow_special_forms=self.__forward_is_class__,
            )
            self.__forward_value__ = _eval_type(
                type_, globalns, localns, recursive_guard | {self.__forward_arg__}
            )
            self.__forward_evaluated__ = True
        return self.__forward_value__
```

and here's the type stub of the `eval` method where those args get used:

```python
def eval(
    __source: str | ReadableBuffer | CodeType,
    __globals: dict[str, Any] | None = None,
    __locals: Mapping[str, object] | None = None,
) -> Any: ...
```

so my type signature (`globalns: dict[str, object] | None, localns: dict[str, object] | None`) is consistent with the usage in `eval`

sphinx/util/inventory.py
```diff
@@ -77,7 +77,9 @@ def read_compressed_lines(self) -> Iterator[str]:
 
 class InventoryFile:
     @classmethod
-    def load(cls: type[InventoryFile], stream: IO, uri: str, joinfunc: Callable) -> Inventory:
+    def load(
+        cls: type[InventoryFile], stream: IO[str], uri: str, joinfunc: Callable,
```
picnixz:
```suggestion
        cls: type[InventoryFile], stream: IO[str], uri: str, joinfunc: Callable[[str, str], str],
```

A similar type annotation for the load functions should be done (I cannot comment on an unchanged part of the code).

sphinx/util/inspect.py
picnixz: I'd like to suggest changing `object` to `Any` since the latest typeshed uses `dict[str, Any]`: https://github.com/python/typeshed/blob/49b1a1a96a90946ff8885792eec30acb5cf39af0/stdlib/typing.pyi#L929

picnixz (approved):

sphinx/util/inventory.py
danieleades: done!

sphinx/util/inspect.py
danieleades: > I'd like to suggest changing `object` to `Any` since the latest typeshed uses `dict[str, Any]`: https://github.com/python/typeshed/blob/49b1a1a96a90946ff8885792eec30acb5cf39af0/stdlib/typing.pyi#L929

done!

sphinx/util/inventory.py
```diff
@@ -78,7 +78,10 @@ def read_compressed_lines(self) -> Iterator[str]:
 class InventoryFile:
     @classmethod
     def load(
-        cls: type[InventoryFile], stream: IO[str], uri: str, joinfunc: Callable,
+        cls: type[InventoryFile],
+        stream: IO[bytes],
+        uri: str,
+        joinfunc: Callable[[str, str], Any],
```
picnixz:
```suggestion
        joinfunc: Callable[[str, str], str],
```

sphinx/util/inventory.py
```diff
@@ -91,7 +94,10 @@ def load(
 
     @classmethod
     def load_v1(
-        cls: type[InventoryFile], stream: InventoryFileReader, uri: str, join: Callable,
+        cls: type[InventoryFile],
+        stream: InventoryFileReader,
+        uri: str,
+        join: Callable[[str, str], Any],
```
picnixz:
```suggestion
        join: Callable[[str, str], str],
```

sphinx/util/inventory.py
```diff
@@ -111,7 +117,10 @@ def load_v1(
 
     @classmethod
     def load_v2(
-        cls: type[InventoryFile], stream: InventoryFileReader, uri: str, join: Callable,
+        cls: type[InventoryFile],
+        stream: InventoryFileReader,
+        uri: str,
+        join: Callable[[str, str], Any],
```
picnixz:
```suggestion
        join: Callable[[str, str], str],
```

sphinx/util/inventory.py
```diff
@@ -4,7 +4,7 @@
 import os
 import re
 import zlib
-from typing import IO, TYPE_CHECKING, Callable
+from typing import IO, TYPE_CHECKING, Any, Callable
```
picnixz:
```suggestion
from typing import IO, TYPE_CHECKING, Callable
```

picnixz:
Thank you!

PR #11899, opened 2024-01-20
Give a more plausible example for a combination of tags

pcgaustad:
Minor emendation. The tags "latex" and "html" are seldom true at the same time.


AA-Turner:
Thanks!

PR #11901, opened 2024-01-21
Use Ruff's ``extend-exclude``

danieleades:
ruff's docs list 'extend exclude' as the preferred option.
that reduces the boilerplate, and allows the config to only list the *differences* from standard python config


danieleades:
now that's interesting, the `[lint]` block doesn't support the `extend-exclude` config, but the top level does.
That doesn't matter for sphinx right now, but *if* sphinx were to use ruff for formatting, and wanted to use ruff to format those files but *not* lint them, this would have to be reverted

AA-Turner:
> now that's interesting, the `[lint]` block doesn't support the `extend-exclude` config, but the top level does.
> 
> That doesn't matter for sphinx right now, but *if* sphinx were to use ruff for formatting, and wanted to use ruff to format those files but *not* lint them, this would have to be reverted

Should we report this upstream?

danieleades:
> Member

we could. it might also be deliberate. If you have a top-level extend exclude and then child extend excludes, then suddenly you need some kind of inheritance strategy. or rules about only having one or the other

picnixz:
>  That doesn't matter for sphinx right now, but if sphinx were to use ruff for formatting, and wanted to use ruff to format those files but not lint them, this would have to be reverted

Is there a way for enable the formatter and disable the linter using a top-level comment ?

PR #11902, opened 2024-01-21
Deprecate redundant linting

danieleades:
removes deprecated linting.

note that this is missing one lint - E704. I feel like it's overkill keeping an entire flake8 toolchain around for just one lint.
Especially since that lint has been comprehensively disabled inline with `noqas`


danieleades:
perhaps i should remove all of the `noqa: f704`s? i'll add another separate commit with those removed and we can discuss

AA-Turner:
Would the current Ruff settings raise the flake8 failure from https://github.com/sphinx-doc/sphinx/pull/11791#issuecomment-1870882620? I'll try and find some other recent examples where flake8 failed but Ruff didn't.

A

danieleades:
> Would the current Ruff settings raise the flake8 failure from [#11791 (comment)](https://github.com/sphinx-doc/sphinx/pull/11791#issuecomment-1870882620)? I'll try and find some other recent examples where flake8 failed but Ruff didn't.
> 
> A

it *should*. From memory that was just a formatting issue related to some indents. In principle, the `E*` lints should pick that up, right?

If only there was some kind of tool, perhaps built into ruff, that would automatically resolve issues with formatting and indentation ;)

AA-Turner:
Can we keep the 'external' entry and the inline suppressions? I'd be happier to merge then.

A

danieleades:
> Can we keep the 'external' entry and the inline E704 suppressions? I'd be happier to merge then.
> 
> A

the `external` entry doesn't seem to be required any more. Perhaps as a result of the adding and subsequent removing of E704 from ruff?

AA-Turner:
Having looked a little closer, see:

* https://github.com/astral-sh/ruff/issues/2402
* https://github.com/astral-sh/ruff/issues/4666

There are several rules that flake8 does implement that Ruff simply doesn't, and we aren't at a point where we'll adopt an automatic formatter. So I think I'm happy to drop flake8-simplify, but we still need to keep flake8 proper, it seems.

A

AA-Turner:
See #11906

PR #11903, opened 2024-01-21
use consistent capitalisation for 'noqa' statements

danieleades:
all lowercase is the format used in the docs of both flake8 and ruff

right now there's a mixture of the two

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -128,7 +128,7 @@ locale_dirs = ['locale/']
 gettext_compact = False
 
 nitpick_ignore = {
-    ('cpp:class', 'template<typename TOuter> template<typename TInner> Wrapper::Outer<TOuter>::Inner'),  # noqa: E501
+    ('cpp:class', 'template<typename TOuter> template<typename TInner> Wrapper::Outer<TOuter>::Inner'),  # NoQA: E501
     ('cpp:identifier', 'MyContainer'),
     ('js:func', 'SomeError'),
     ('js:func', 'number'),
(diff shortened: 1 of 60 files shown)


AA-Turner:
If we're going to standardise, please can we use NoQA. It's much more obvious that it means 'no quality assurance'.

A

danieleades:
> If we're going to standardise, please can we use NoQA. It's much more obvious that it means 'no quality assurance'.
> 
> A

just to play devils advocate here, why does it matter what it stands for?

I think sphinx is the only repo i've seen that capitalises the statement this way. Surely community consistency is more important?

that said, it's not a hill i plan to die on, and I would prefer `NoQA` to a mixture of the two

electric-coder:
> If we're going to standardise, please can we use NoQA.
> I think sphinx is the only repo i've seen that capitalises the statement this way.

Not even the Flake8 documentation is completely consistent as there are at least [two capitalized examples mixed with the prevalent lowercase][1].

However, since the majority of uses is lowercase I'd opine for trying to be consistent with the trend of the Flake8 repository.

A good [SO overview thread][2] for reference.

[1]: https://flake8.pycqa.org/en/latest/search.html?q=noqa&check_keywords=yes&area=default
[2]: https://stackoverflow.com/q/45346575

picnixz:
Just my 2 cent before vanishing again: I personally prefer noqa because 1) that's what I see the most 2) I don't have to capitalise anything (so less keys to type).

jayaddison:
I'm fairly ambivalent about it, but I do prefer `NoQA` in Sphinx, following the same reasoning that it's marginally more self-documenting.  I'm sympathetic to the fewer-shift-keypresses argument too though.

PR #11909, opened 2024-01-22
builders: fix handling of string-based outdir  in _file_checksum

2maz:
Subject:  fix handling of string-based outdir in _file_checksum

### Feature or Bugfix
- Bugfix

### Purpose
- ensure robust path handling in _file_checksum

### Detail
Use sphinx in combination with jupyter-book and encountered this error: "AttributeError: 'str' object has no attribute 'joinpath'"
Sphinx.application.outdir is set to str by juypter-book (https://github.com/executablebooks/jupyter-book/blob/9e4804de11a2edf44ff9b56c5b02728acc3aa37b/jupyter_book/sphinx.py#L131C17-L131C23)

### DIFF
--- a/sphinx/builders/html/_assets.py
+++ b/sphinx/builders/html/_assets.py
@@ -8,8 +8,7 @@ from typing import TYPE_CHECKING, Any, NoReturn
 from sphinx.deprecation import RemovedInSphinx90Warning
 from sphinx.errors import ThemeError
 
-if TYPE_CHECKING:
-    from pathlib import Path
+from pathlib import Path
 
 
 class _CascadingStyleSheet:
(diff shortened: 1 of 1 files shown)


AA-Turner:
This is a bug in juypter-book, not Sphinx. Modifying the paths is unsupported, and falls outside of the contract. The reason why it now fails is as we've started moving towards ``pathlib.Path`` objects.

Please encourage juypter-book to update their usages: if they want something backwards compatible, perhaps:

``app.srcdir = type(app.srcdir)(Path(app.srcdir).as_posix())``

Though, I strongly advise just removing the manipulation.

A

2maz:
@AA-Turner
1. If the paths should not be modified, why not making them read-only? 
2. I see that the interface was outdir: str | os.PathLike[str] before, so the 'bug' has been created by sphinx by changing the interface. This backward compatibility issue can be easily solved here?

2maz:
@AA-Turner this was exposed by local setup (pip resolver) issue with a too recent sphinx version 7.X.
Juypter-book dependency is actually <6, where this issue will not be seen.

PR #11910, opened 2024-01-22
Maintenance: apply pyupgrade suggestions for py3.9

jayaddison:
### Feature or Bugfix
- Refactoring

### Purpose
- Some suggestions provided by [`pyupgrade`](https://github.com/asottile/pyupgrade) to modernise the codebase to the [Python 3.9 baseline](https://github.com/sphinx-doc/sphinx/blob/fa290049515c38e68edda7e8c17be69b8793bb84/pyproject.toml#L16).

### Detail
- For application code: `find sphinx -name '*.py' -exec pyupgrade --py39 {} \;`
- For the test suite: began by running a similar command, but that resulted in many false positives.  The test suite makes use of various pre-py3.9 patterns that we should likely retain support for for a while.

### Relates
- N/A

### DIFF
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -83,7 +83,7 @@ if TYPE_CHECKING:
     from collections.abc import MutableMapping
     from typing import Literal
 
-    from typing_extensions import overload
+    from typing import overload
 
     from sphinx.domains.c import CDomain
     from sphinx.domains.changeset import ChangeSetDomain
--- a/sphinx/environment/__init__.py
+++ b/sphinx/environment/__init__.py
@@ -81,9 +81,7 @@ versioning_conditions: dict[str, bool | Callable] = {
 
 if TYPE_CHECKING:
     from collections.abc import MutableMapping
-    from typing import Literal
-
-    from typing import overload
+    from typing import Literal, overload
 
     from sphinx.domains.c import CDomain
     from sphinx.domains.changeset import ChangeSetDomain
--- a/tests/test_util/test_util_typing.py
+++ b/tests/test_util/test_util_typing.py
@@ -62,7 +62,7 @@ T = TypeVar('T')
 MyInt = NewType('MyInt', int)
 
 
-class MyList(List[T]):
+class MyList(list[T]):
     pass
 
 
--- a/tests/test_util/test_util_typing.py
+++ b/tests/test_util/test_util_typing.py
@@ -62,7 +62,7 @@ T = TypeVar('T')
 MyInt = NewType('MyInt', int)
 
 
-class MyList(list[T]):
+class MyList(List[T]):
     pass
 
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,7 +29,7 @@ jobs:
       run: python -m pip install --upgrade pip
 
     - name: Install Ruff
-      run: python -m pip install "ruff==0.1.14"
+      run: python -m pip install "ruff==0.2.0"
 
     - name: Lint with Ruff
       run: ruff . --output-format github
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,7 +1,7 @@
 target-version = "py39"  # Pin Ruff to Python 3.9
 line-length = 95
-required-version = "0.1.14"
-show-source = true
+required-version = "0.2.0"
+output-format = "full"
 
 [lint]
 preview = true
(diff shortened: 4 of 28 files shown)


jayaddison:
Some caution here: it looks like I used the wrong commandline invocation.  That _should_ have been:

```sh
$ find sphinx -name '*.py' -exec pyupgrade --py39-plus {} \;
```

The results seem no different, but I'm double-checking it anyway.

jayaddison:
> Some caution here: it looks like I used the wrong commandline invocation.

In this case, this was a non-issue, thanks to the [`argparse` setting `allow_abrev` having a default of `True`](https://docs.python.org/3.11/library/argparse.html#allow-abbrev) - unambiguous long-name flags/options are expanded.

As it turns out, all of the changes here seem acceptable for `py3-only` too.  But I'll keep the pull request description and title as-is, based on the intent.

sphinx/environment/__init__.py
```diff
@@ -81,9 +81,7 @@
 
 if TYPE_CHECKING:
     from collections.abc import MutableMapping
-    from typing import Literal
-
-    from typing_extensions import overload
+    from typing import Literal, overload
```
ferdnyc, line 84: I don't get how this was working _now_. Nothing imports `typing-extensions` into the project, that I can see. It's not listed in any of the dependencies lists, and a `pip install -e '.[test]'` from the project root doesn't pull it into the current venv, even when running Python 3.9.

Even more confusing, it _never has_ been imported into the project, AFAICT. A `git log -G 'typing.extensions'` turns up nothing but two edits to **this file** back in 2022.

ferdnyc: Does Mypy automatically import it as needed, or something?

tests/test_util/test_util_typing.py
```diff
@@ -62,7 +62,7 @@ class MyEnum(Enum):
 MyInt = NewType('MyInt', int)
 
 
-class MyList(List[T]):
+class MyList(list[T]):
```
ferdnyc: This seems like a strange change, in isolation. The entire rest of the file is still using the capitalized `typing` imports. Why change only this _one_ class?

sphinx/environment/__init__.py
ferdnyc: Never mind, `mypy` does indeed have a dependency on `typing_extensions`. I hadn't realized _Mypy_ wasn't part of the `sphinx[test]` dependencies either.

tests/test_util/test_util_typing.py
jayaddison: That's a good question; I think the most truthful response is that it's a cautious approach, and that I was concerned that since the test expectations have `typing.List` within string literals, I didn't feel confident that they would be testing the same behaviour after a migration.

A naive evaluation of `type(list)` and `type(typing.List)` locally indicates that they remain non-identical.. so a more cautious approach could be to omit this change.

To attempt to extract a grander principle from all of the above, perhaps it's that automated code reformatting is fine, but that I sense some amount of risk in automatically rewriting test cases (because, ultimately, that could mean that a behaviour-change is obscured by what appears to be a stylistic modification).

jayaddison: If the test-behaviour-change issue is a concern, and given commit `A` prior to the test suite change, and commit `B` subsequent to the change, then I suppose it could make sense to run both test suites from `A` and `B` against the application/library code from commit `B`, to assert that the changes in `B` did indeed not have any behavioural effect.

It feels like it may be insufficient to provide complete confirmation of no-behaviour-change, though.  It could simply be that the test coverage was insufficient to detect a change.

sphinx/environment/__init__.py
jayaddison: Well spotted - I scanned that line a few times but didn't think to check the origin of the module.  I suppose it's somewhat-hidden by the fact that the `TYPE_CHECKING` paths are evaluated under different conditions (such as when using `mypy` as you mention).

FWIW: That does appear to be the _only_ reference to `typing_extensions` in the source.  There are a few non-`sphinx`-internal imports within `TYPE_CHECKING` guards; mostly to `docutils` and `jinja` (relatively safe, stable dependencies), and some Python internal deps (again OK until such time as those interfaces change).

ferdnyc: \*nod\* I noticed that it was the only instance of `typing_extensions` in the codebase, and (with these changes) the only surviving import from it is the `@overload` decorator.

Which actually existed in Python 3.9 `typing`, even, so it seems like there's already no reason to use `typing_extensions` for anything.

(Runtime introspection features like `get_overloads()` weren't added to `typing` until Python 3.11, but Sphinx doesn't use those.)

tests/test_util/test_util_typing.py
picnixz: Yes, this one needs to be like that. While `typing.List` is an alias of `list`, it's more a special alias. There are runtime differences between `typing.List` and `list` and those differences may matter for the tests (note that in the tests, I use `list` *and* `typing.List` to ensure that objects in the `builtins` and `typing` modules are correctly handled even though they are semantically equivalent).

sphinx/environment/__init__.py
picnixz: We don't use `get_overloads` for now but we might support it in the future (there is an open issue somewhere I think but I forgot where)

picnixz:
Thank you !

PR #11916, opened 2024-01-25
Static templates: also support ``.jinja`` files

jayaddison:
### Feature or Bugfix
- Feature

### Purpose
- Allow themes and other extensions that utilise the static template functionality to use the [suggested](https://jinja.palletsprojects.com/en/3.1.x/templates/#template-file-extension) (note: not strictly required) `.jinja` extension for templates.
- The originating motivation I have for this feature is to eventually remove a catch-all missing-suffix expectation in `coveragepy` that can cause warnings to be emitted when non-Python files are treated as potentially containing Python bytecode for coverage evaluation (ref https://github.com/nedbat/coveragepy/issues/1488).

### Detail
- When considering whether to render a file using Jinja, check for _either_ the existing already-supported `_t` suffix _or_ secondly for the additionally-supported `.jinja` suffix.
- These changes currently target version 7.3 of Sphinx, but I'd be fine with waiting until a future version to introduce them.

### Relates
- Re-attempts previous pull request #11165 - this time without deprecation notices or a deprecation schedule.


AA-Turner:
Can we split adding support for ``.jinja`` and renaming the internal template files to two PRs?

jayaddison:
> Can we split adding support for `.jinja` and renaming the internal template files to two PRs?

That seems reasonable, yep.  I've restored the original template file paths in this branch.

This could be a case where I'd consider rebasing the commits here.  It doesn't seem to make a practical difference given that the changes will be squashed, but somehow it feels odd to have two changes that combine into a no-op in a merged PR.

sphinx/writers/latex.py
```diff
@@ -517,12 +517,13 @@ def generate(content: list[tuple[str, list[IndexEntry]]], collapsed: bool) -> No
...
-            if path.exists(template):
-                return renderer.render(template, variables)
+            for template_suffix in ('_t', '.jinja'):
+                template = path.join(self.builder.confdir, template_dir,
+                                     template_name + template_suffix)
+                if path.exists(template):
+                    return renderer.render(template, variables)
 
-        return renderer.render(template_name, variables)
+        return renderer.render(template_name + '_t', variables)
```
jayaddison: Something about this line feels wrong to me, in the sense of introducing an unintended behaviour change / incompatibility.

jayaddison: I think it's safest to undo this and further reduce the size of the diff.

chrisjsewell:
Definitely agree with this by the way, I was just thinking the other day that I wanted these templates to be compatible with https://marketplace.visualstudio.com/items?itemName=samuelcolvin.jinjahtml
(Haven't reviewed the actual code yet though)

CHANGES.rst
```diff
@@ -75,6 +75,10 @@ Deprecated
 
 Features added
 --------------
+* #11165: Support the `officially recommended`_ ``.jinja`` suffix for template
+  files. Patch by James Addison and Adam Turner
```
```suggestion
  files. 
  Patch by James Addison and Adam Turner
```

sphinx/util/fileutil.py
```diff
@@ -23,7 +23,7 @@ def _template_basename(filename: str | os.PathLike[str]) -> str | None:
     basename = os.path.basename(filename)
     if basename.lower().endswith('_t'):
         return str(filename)[:-2]
-    elif basename.lower().endswith(".jinja"):
```
jayaddison: I'm not sure about the removal of `.lower()` here; what's the thinking behind that?

jayaddison: @AA-Turner Although it might seem non-ideal in some ways to allow people to create template files named arbitrary combinations of `[Jj][Ii][Nn][Jj][Aa]`, initially I still feel that it's better to compare the file suffix case-insensitively.

Ref: https://www.hanselman.com/blog/git-is-casesensitive-and-your-filesystem-may-not-be-weird-folder-merging-on-windows for example.

AA-Turner (approved):

jayaddison:
Thank you @AA-Turner!

PR #11919, opened 2024-01-29
#11917 - Careful access of annotations on class objects/fix incorrect display of inherited members 

### LINKED ISSUE #11917: Sphinx accesses inherited annotations incorrectly in python 3.9 and earlier
### Describe the bug

In python 3.10, a bug was fixed where an attempt to read the annotations on a class object could accidentally return the annotations on a superclass instead.  For python 3.9 and earlier, the is the suggested workaround:

https://docs.python.org/3.10/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older

It appears that this needs to be implemented in sphinx/util/inspect.py line 88

In my test case, class A1 is NOT using inherited-members and class A2 is using inherited-members.
The inherited attribute `inherit_me` shows up in both classes when sphinx is run in a python 3.9 environment but only on class A2 if you build and install in python 3.10

### How to Reproduce

index.rst 
```rst
My Docs
=======

.. currentmodule:: my_module

.. autoclass:: A1
   :members:

.. autoclass:: A2
   :members:
   :inherited-members:
```

source code
```python
class B:
    inherit_me: int
    """Inherit me from baseclass B"""

class A1(B):
    """inherit test #1  - members but not inherited-memebers"""

    my_a1 = 1
    """my_a1"""


class A2(B):
    """inherit test #2. -- members and inherited-members"""

    my_a2 = 2
    """my_a2"""
```

### Environment Information

Python 3.9 environment:

```text
Platform:              darwin; (macOS-14.2.1-arm64-arm-64bit)
Python version:        3.9.18 (main, Aug 24 2023, 18:16:58)
[Clang 15.0.0 (clang-1500.1.0.2.5)])
Sphinx version:        7.2.6
Docutils version:      0.20.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

Python 3.10 environment:
```text
Platform:              darwin; (macOS-14.2.1-arm64-arm-64bit)
Python version:        3.10.13 (main, Aug 24 2023, 12:59:26) [Clang 15.0.0 (clang-1500.1.0.2.5)])
Sphinx version:        7.2.6
Docutils version:      0.20.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

### Sphinx extensions

```python
extensions = [ "sphinx.ext.autodoc" ]
source_suffix = {
  '.rst': 'restructuredtext'
}
```

### Additional context

This is possibly a duplicate or related to a number of other issues referencing inherited annotations... but I've debugged this particular test case

j-carson:
Subject: Fixes a bug in the access of class __annotations__ dictionary in python 3.9 and earlier, causing superclass members to be mistaken for members of the class being documented

### Feature or Bugfix
- Bugfix

### Purpose
If the __annotations__ dictionary on a class object is not accessed correctly, then Sphinx autoclass can get the superclass annotations rather than the annotations of the object currently being documented. If "inherited-members" is not requested, then these superclass annotations would result in incorrect output.

This problem only happens in python 3.9 and earlier, and I implemented the suggested work-around per the docs

https://docs.python.org/3.10/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older

### Detail
Closes #11917

### Relates
#11917

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,6 +72,8 @@ Bugs fixed
   Patch by Colin Marquardt.
 * #11598: Do not use query components in URLs for assets in EPUB rendering.
   Patch by David Runge.
+* #11917: Fixes some corner-cases involving inherited class members on python 3.9 and earlier
+  Patch by Janet Carson.
 
 Testing
 -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,8 +72,8 @@ Bugs fixed
   Patch by Colin Marquardt.
 * #11598: Do not use query components in URLs for assets in EPUB rendering.
   Patch by David Runge.
-* #11917: Fixes some corner-cases involving inherited class members on python 3.9
-  and earlier. Patch by Janet Carson.
+* #11917: Fix rendering of annotated inherited members for Python 3.9.
+  Patch by Janet Carson.
 
 Testing
 -------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,8 +72,8 @@ Bugs fixed
   Patch by Colin Marquardt.
 * #11598: Do not use query components in URLs for assets in EPUB rendering.
   Patch by David Runge.
-* #11917: Fixes some corner-cases involving inherited class members on python 3.9 and earlier
-  Patch by Janet Carson.
+* #11917: Fixes some corner-cases involving inherited class members on python 3.9
+  and earlier. Patch by Janet Carson.
 
 Testing
 -------
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -87,7 +87,7 @@ def getall(obj: Any) -> Sequence[str] | None:
 
 def getannotations(obj: Any) -> Mapping[str, Any]:
     """Get __annotations__ from given *obj* safely."""
-    if (sys.version_info >= (3, 10, 0)) or (not isinstance(obj, type)):
+    if sys.version_info >= (3, 10, 0) or not isinstance(obj, type):
         __annotations__ = safe_getattr(obj, '__annotations__', None)
     else:
         # Workaround for bugfix not available until python 3.10 as recommended by docs
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -92,7 +92,8 @@ def getannotations(obj: Any) -> Mapping[str, Any]:
     else:
         # Workaround for bugfix not available until python 3.10 as recommended by docs
         # https://docs.python.org/3.10/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older
-        __annotations__ = obj.__dict__.get('__annotations__', None)
+        __dict__ = safe_getattr(obj, '__dict__', {})
+        __annotations__ = __dict__.get('__annotations__', None)
     if isinstance(__annotations__, Mapping):
         return __annotations__
     else:
--- a/sphinx/util/inspect.py
+++ b/sphinx/util/inspect.py
@@ -87,7 +87,12 @@ def getall(obj: Any) -> Sequence[str] | None:
 
 def getannotations(obj: Any) -> Mapping[str, Any]:
     """Get __annotations__ from given *obj* safely."""
-    __annotations__ = safe_getattr(obj, '__annotations__', None)
+    if (sys.version_info >= (3, 10, 0)) or (not isinstance(obj, type)):
+        __annotations__ = safe_getattr(obj, '__annotations__', None)
+    else:
+        # Workaround for bugfix not available until python 3.10 as recommended by docs
+        # https://docs.python.org/3.10/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older
+        __annotations__ = obj.__dict__.get('__annotations__', None)
     if isinstance(__annotations__, Mapping):
         return __annotations__
     else:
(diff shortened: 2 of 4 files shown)


sphinx/util/inspect.py
```diff
@@ -87,7 +87,12 @@ def getall(obj: Any) -> Sequence[str] | None:
 
 def getannotations(obj: Any) -> Mapping[str, Any]:
     """Get __annotations__ from given *obj* safely."""
-    __annotations__ = safe_getattr(obj, '__annotations__', None)
+    if (sys.version_info >= (3, 10, 0)) or (not isinstance(obj, type)):
+        __annotations__ = safe_getattr(obj, '__annotations__', None)
+    else:
+        # Workaround for bugfix not available until python 3.10 as recommended by docs
+        # https://docs.python.org/3.10/howto/annotations.html#accessing-the-annotations-dict-of-an-object-in-python-3-9-and-older
+        __annotations__ = obj.__dict__.get('__annotations__', None)
```
picnixz: Not all objects have a `__dict__` (some ill-formed types may not necessarily have one) so I'd suggest yousing `getattr(obj, '__dict__', {})` for this one (or use `safe_getattr` which *should* be fine).

tests/roots/test-ext-autodoc/target/inherited_annotations.py
```diff
@@ -0,0 +1,16 @@
...
+    inherit_me: int
+    """Inherited"""
+
+class NoTypeAnnotation(HasTypeAnnotatedMember):
+    a = 1
+    """Local"""
+
+class NoTypeAnnotation2(HasTypeAnnotatedMember):
+    a = 1
+    """Local"""
```
picnixz: I think this file is missing one LF at the end (at least it appears so on github but I don't remember whether the other testroots files have one or not actually).

CHANGES.rst
```diff
@@ -72,6 +72,8 @@ Bugs fixed
   Patch by Colin Marquardt.
 * #11598: Do not use query components in URLs for assets in EPUB rendering.
   Patch by David Runge.
+* #11917: Fixes some corner-cases involving inherited class members on python 3.9
+  and earlier. Patch by Janet Carson.
```
```suggestion
* #11917: Fix rendering of annotated inherited members for Python 3.9.
  Patch by Janet Carson.
```

We do not support 3.8 and below so there is no need to mention it.

sphinx/util/inspect.py
```diff
@@ -87,7 +87,12 @@ def getall(obj: Any) -> Sequence[str] | None:
 
 def getannotations(obj: Any) -> Mapping[str, Any]:
     """Get __annotations__ from given *obj* safely."""
-    __annotations__ = safe_getattr(obj, '__annotations__', None)
+    if (sys.version_info >= (3, 10, 0)) or (not isinstance(obj, type)):
```
picnixz:
```suggestion
    if sys.version_info >= (3, 10, 0) or not isinstance(obj, type):
```

tests/test_extensions/test_ext_autodoc_autoclass.py
```diff
@@ -515,3 +515,55 @@ def test_autoattribute_TypeVar_module_level(app):
         "   alias of TypeVar('T1')",
         '',
     ]
+
+
+@pytest.mark.sphinx('html', testroot='ext-autodoc')
+def test_inherited_instance_variable_with_annotations(app):
```
picnixz, line 521: No need to comment what the test does I think. Most of the tests don't have a docstring and it should be pretty clear what we are actually doing because of the name imo.

picnixz (changes requested):
I don't think there should be any issue with that one (except for ill-formed types) so once changes have been applied, it should be fine to merge (someone else might do it since I'm only here for a few hours (I'll be back around mid February)).

CHANGES.rst
j-carson: added

sphinx/util/inspect.py
j-carson: Will look more at this -- I did use the exact code recommended by the python contributor team though

tests/test_extensions/test_ext_autodoc_autoclass.py
j-carson: removed

tests/roots/test-ext-autodoc/target/inherited_annotations.py
j-carson: fixed

sphinx/util/inspect.py
j-carson: I want to be careful because the whole point of this PR is a corner case where safe_getattr definitely does not work, and I'm not sure how to test a malformed type?

picnixz: The official doc also mentions 

> Note that some exotic or malformed type objects may not have a __dict__ attribute, so for extra safety you may also wish to use [getattr()](https://docs.python.org/3.10/library/functions.html#getattr) to access __dict__.

Now for actual testing, we can just ignore it. I am not aware of any construction in plain Python that makes the `__dict__` of a class disappear, even with metaclass hacks. It usually happens when a class is written in C (mypyc native classes may not have a `__dict__` iirc, but I'm not sure that this portion of the code could be triggered with them).

Anyway, if you want to be on the safe side, using `getattr(obj, '__dict__', {})` is sufficient.

j-carson: So, use getattr to get the dict, but not add a test case for the corner case of the corner case?

picnixz (approved):
Everything looks fine to me now. Once you fixed the merge conflicts and the tests are fine, we can merge it.

picnixz:
Thank you !

PR #11926, opened 2024-02-01
Extension registry: add sphinxprettysearchresults to blacklist

### LINKED ISSUE #11925: sphinx-pretty-searchresults is now redundant
(`sphinx-pretty-searchresults` provides a workaround for bug #1618 in Sphinx, and that bug was fixed within Sphinx itself from v2.0.0 and beyond)

We should add jt to the list of extensions incorporated into the core.

_Originally posted by @jayaddison in https://github.com/sphinx-doc/sphinx/issues/11921#issuecomment-1918945841_

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Discourage use of the now-fairly-long-redundant `sphinxprettysearchresults` extension.

### Detail
- Adds `sphinxprettysearchresults` to the list of blacklisted extensions, with the corresponding Sphinx version where the functionality became available in Sphinx core (v2.0.0).

### Relates
- Resolves #11925.
- Relates to #1618, #4022.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -72,6 +72,9 @@ Bugs fixed
   Patch by Colin Marquardt.
 * #11598: Do not use query components in URLs for assets in EPUB rendering.
   Patch by David Runge.
+* #11925: Blacklist the ``sphinxprettysearchresults`` extension; the functionality
+  it provides was merged into Sphinx v2.0.0.
+  Patch by James Addison.
 
 Testing
 -------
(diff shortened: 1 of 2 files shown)


jayaddison:
Note for anyone curious / reviewing: adding an extension to the blacklist doesn't prevent a project from building, but it does cause a warning message to be emitted, and the extension will not be loaded.

Relevant code snippet: https://github.com/sphinx-doc/sphinx/blob/aff95789ac252149e5319794f818c35831259f4d/sphinx/registry.py#L438-L446

PR #11927, opened 2024-02-01
Upgrade Ruff configuration for v0.2.0

charliermarsh:
## Summary

Hopefully a welcome PR, but this upgrades Sphinx's Ruff configuration to reflect some of the changes we're making in v0.2.0. We use Sphinx in our ecosystem checks, and I wanted to do an example migration anyway, both for illustrative purposes and for my own testing. (This PR won't pass until we do the actual v0.2.0 release.)

Here's a summary of the relevant changes for Sphinx:

- `show-source = true` was replaced by `output-format = "full"`. (This is a deprecation, not a removal, so changing it here merely removes the deprecation warning.)
- Rules `ANN101` and `ANN102` were deprecated. (Selecting them with preview enabled is now an error; selecting them outside of preview is a warning.)
- `PGH001` was redirected to `S307` (I enabled `S307` and changed the `# NoQA` assertions, which would've continued to work, but with a deprecation warning).
- `RUF011` was redirected to `B035` (which you already have enabled, so I just removed `RUF011`, which would've warned).
- `TCH006` was redirected to `TCH010` to match the upstream plugin.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,7 @@ docs = [
 ]
 lint = [
     "flake8>=3.5.0",
-    "ruff==0.1.14",
+    "ruff==0.2.0",
     "mypy==1.8.0",
     "sphinx-lint",
     "docutils-stubs",
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,7 +29,7 @@ jobs:
       run: python -m pip install --upgrade pip
 
     - name: Install Ruff
-      run: python -m pip install "ruff==0.1.14"
+      run: python -m pip install "ruff==0.2.0"
 
     - name: Lint with Ruff
       run: ruff . --output-format github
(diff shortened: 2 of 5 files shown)


charliermarsh:
For completeness, these are the warnings that would've been raised:

```
warning: The `show-source` option has been deprecated in favor of `output-format`'s "full" and "concise" variants. Please update your configuration to use `output-format = <full|concise>` instead.
warning: `RUF011` has been remapped to `B035`.
warning: `TCH006` has been remapped to `TCH010`.
ruff failed
  Cause: Selection of deprecated rule `ANN102` is not allowed when preview is enabled.
```

AA-Turner:
Thanks Charlie!

pyproject.toml
```diff
@@ -82,7 +82,7 @@ docs = [
 ]
 lint = [
     "flake8>=3.5.0",
-    "ruff==0.1.14",
+    "ruff==0.2.0",
     "mypy==1.8.0",
```
HOLYGHOST1111, line 86: me

PR #11932, opened 2024-02-03
[8.x] [intersphinx] support for arbitrary title names

### LINKED ISSUE #11711: Intersphinx cannot link to some Headers containing digits
### Describe the bug

I recently found an issue where I could not link to headers containing digits.
The link targets were automatically generated with autosectionlabels and work fine locally.

Using sphobjinv I can also verify that all link targets are generated in the objects.inv file.
However, I cannot access  headers via interspinx that contain more than one number separated by spaces.

I presume that there is an issue with intersphinx' conversion/evaluation of the objects.inv file.

### How to Reproduce

I created a small example tthat reproduces the error here:
https://github.com/Thosse/Playground_GH_Pages

The corresponding Github page can be found here:
https://thosse.github.io/Playground_GH_Pages/

### Environment Information

```text
This was reproduced on a manjaro arch linux and Windows 10
```

### Sphinx extensions

```python
['sphinx.ext.autosectionlabel']
```

### Additional context

_No response_

picnixz:
Fix #11711.

Since the logic is different, this required a complete update on how intersphinx inventories are represented. Before, each line in an inventory was of the form

```
name [SPACE] domain [COLON] objtype [SPACE] priority [SPACE] URI [SPACE] display name
```

However, if *name* contains a whitespace (which is the case in the linked issue), then it becomes extremely hard to have a "good" regex. Now, we represent records as follows:

```
priority ([COLON] name_length) name [SPACE] domain [COLON] objtype [SPACE] URI [SPACE] display name
```

where the name's length (and the colon) is only included *if* the name contains a space. In particular, the inventory's size is *not* changed if the name has no space inside. Otherwise, we increase a bit the inventory size with a 64-bit integer for those entries.


sphinx/util/inventory.py
```diff
@@ -172,7 +243,16 @@ def escape(string: str) -> str:
...
-                             (name, domainname, typ, prio, uri, dispname))
+
+                    # For names with spaces, we need to know exactly where
+                    # the ref-type string starts. Technically, we should not
+                    # have ':' inside domain or role names, but extensions
+                    # may have some weird role names and they could handle
+                    # them internally to be docutils compatible. As such,
+                    # we encode the length of the name after the priority.
+                    slen = f':{len(name)}' if ' ' in name else ''
+                    entry = '%s%s %s %s:%s %s %s\n' % (
```
picnixz, line 268: One alternative that I had in mind is to use a special character for the end of the name, but this means that we assume that this character is not used in the name (e.g., `\x00`). The size of the inventory won't be much bigger but we need to check that the name does not contain such character and raise an exception if this is the case during the writing phase.

picnixz:
Quick update after discussing with someone: 

- If a project builds its documentation with the new format, then any other project using that project needs to be built using the latest sphinx version otherwise they won't be able to read the new inventory format. 
- I think this bug already exists for inventories built using v1 vs v2 so it's not really a problem I'd say. However, we could possibly have a configuration value for the format of the inventory we are actually expecting.

bskinn:
I don't think this change to the inventory semantics is necessary.

The [most recent regex] I implemented for `sphobjinv`, which incorporated a couple of tweaks from https://github.com/bskinn/sphobjinv/issues/181 and https://github.com/bskinn/sphobjinv/issues/256, has been working in the wild for [two years now].

It successfully imports all objects from the minimal example provided by OP at #11711. After a `pip install sphobjinv` in a virtualenv:

```python
>>> import sphobjinv as soi
>>> inv = soi.Inventory(url='https://thosse.github.io/Playground_GH_Pages/objects.inv')
>>> len(inv.objects)
18
>>> from pathlib import Path
>>> len(Path("objects.txt").read_text().splitlines()) - 4  # Minus 4 b/c the header lines aren't objects
18
```

That decoded inventory, for reference:

```none
# Sphinx inventory version 2
# Project: Pages Playground
# Version: 
genindex std:label -1 genindex.html Index
index std:doc -1 index.html Intersphinx Minimal Example
index:1 2 3 fails std:label -1 index.html#fails 1 2 3 Fails
index:1 fails 1 std:label -1 index.html#fails-1 1 Fails 1
index:1 works std:label -1 index.html#works 1 Works
index:123 works std:label -1 index.html#id1 123 Works
index:fails 1 2 3 std:label -1 index.html#fails-1-2-3 Fails 1 2 3
index:fails 1 2 fails std:label -1 index.html#fails-1-2-fails Fails 1 2 Fails
index:fails 1 fails 2 std:label -1 index.html#fails-1-fails-2 Fails 1 Fails 2
index:fails fails 1 std:label -1 index.html#fails-fails-1 Fails Fails 1
index:fails fails 2 fails fails std:label -1 index.html#fails-fails-2-fails-fails Fails Fails 2 Fails Fails
index:headers without digits work std:label -1 index.html#headers-without-digits-work Headers without Digits work
index:intersphinx minimal example std:label -1 index.html#intersphinx-minimal-example Intersphinx Minimal Example
index:works 1 std:label -1 index.html#works-1 Works 1
index:works 1 works std:label -1 index.html#works-1-works Works 1 Works
```

It's of course possible that my regex is still missing some key features. But I've spent a _lot_ of time working on this reverse-engineering, and I think it's pretty solid based on the lack of tickets in the `sphobjinv` tracker, which is lately seeing [~40K downloads per month][dls].

There are a few more issues in `sphobjinv` related to the semantics of the object-reference lines in the v2 `objects.inv`. I'd be happy to collect those together if it would be helpful.


[most recent regex]: https://github.com/bskinn/sphobjinv/blob/258f10d6fea37c6990e8e675e29c81e180d4bd85/src/sphobjinv/re.py#L71-L85

[two years now]: https://github.com/bskinn/sphobjinv/blame/258f10d6fea37c6990e8e675e29c81e180d4bd85/src/sphobjinv/re.py#L71-L85

[dls]: https://www.pepy.tech/projects/sphobjinv

picnixz:
@bskinn Thank you for your help but there are some titles such as ``word word1:word2 integer`` that would fail your regex since `word1:word2` would be incorrectly matched as `domain:reftype`. While I don't have a specific example of such title, I cannot exclude that possibility (that's the reason why I gave up on using regexes). However, would you be willing to check that this regex could also extend your regex (and handle my weird case)?

```python
pattern = re.compile(r'''^
	(?P<title>(?:.|[^\s:]+:\S+\s+-?\d+\s*)+)?
	\s+
	(?P<domain>[^\s:]+):(?P<reftype>\S+)
	\s+
	(?P<prio>-?\d+)
	\s+?
	(?P<uri>\S*)
	\s+
	(?P<dispname>.+?)
	\r?
$''', re.M | re.X)
```

I would have wanted to use negative lookbehinds but only .NET regex parser supports variable-width patterns.

> There are a few more issues in sphobjinv related to the semantics of the object-reference lines in the v2 objects.inv. I'd be happy to collect those together if it would be helpful.

I'd be glad to have it. It could help a lot for #12204.

bskinn:
> but there are some titles such as `word word1:word2 integer` that would fail your regex since `word1:word2` would be incorrectly matched as `domain:reftype`.

Indeed, that is a point of failure.

I can take a look at your regex sometime soon, sure, but---I had an idea. What if we changed the paradigm of the format a bit further?

The key hangup as far as I can tell is that the single-line object format doesn't allow for an unambiguous separator character sequence, given the possible domain (not `domain`) of `name`s and `dispname`s.

What if an uncompressed v3 inventory were something like this:

```none
# Sphinx inventory version 3
# Project: MyProject
# Version: 0.1.0
# Object count: 4
# Uncompressed inventory MD5: abcdef123789654
word1:word2 1
  std
  label
  1
  foo/bar.html
  -

word1 word2:word3 2
  std
  label
  1
  foo/bar.html
  -

word1 1 word2:word3 2
  std
  label
  1
  foo/bar.html
  -

foobar
  rst
  directive:option
  1
  baz/quux.html
  -
```

This format would provide `<NEWLINE><SPACE><SPACE>` as a delimiter between fields, and `<NEWLINE><NEWLINE>` as a delimiter between objects. (The ability to use `<NEWLINE>` as part of the delimiter is the key feature here, as none of the fields for a given object reference permit newlines, as best I know.

This format would take a lot more vertical space when decompressed, but ... so what? In nearly all uses, humans won't be reading it, and it'll be zlib-compressed anyways. Even in the few cases where someone _is_ looking directly at it, it still seems quite readable to me.

What do you think?

And, even though there are extra characters in these delimiters, the zlib compression should work quite effectively on them.

bskinn:
Definitely true, including the name length as you propose is probably about the most minimal change achievable, and wouldn't dramatically change the flow of line parsing.

---

Quick size analysis, switching to my newline-delimited proposal would increase the current Python 3.12 compressed inventory by about 2%:

```python
>>> import sphobjinv as soi
>>> inv = soi.Inventory(url="https://docs.python.org/3/objects.inv")
>>> import zlib
>>> ( len_v2 := len(zlib.compress(inv.data_file(), 9)) )
136182
>>> def newline_fmt(do):
...   return f"{do.name}\n  {do.domain}\n  {do.role}\n  {do.priority}\n  {do.uri}\n  {do.dispname}"
...
>>> file_newlines = b'\n'.join(inv.data_file().splitlines()[:4]) + b'\n' + ('\n\n'.join(newline_fmt(o) for o in inv.objects)).encode()
>>> print(file_newlines.decode()[:245])
# Sphinx inventory version 2
# Project: Python
# Version: 3.12
CO_FUTURE_DIVISION
  c
  member
  1
  c-api/veryhigh.html#c.$
  -

METH_CLASS
  c
  macro
  1
  c-api/structures.html#c.$
  -

>>> ( len_nls := len(zlib.compress(file_newlines, 9)) )
138880
>>> len_nls - len_v2
2698
>>> len_nls/len_v2
1.019811722547767
```

Impact is likely greater for smaller inventories, due to the smaller number of repeats of the `\n  ` and `\n\n` motifs. But, for smaller inventories, the size is already not a problem.

picnixz:
> Is that trailing ? supposed to be inside that final )? Where it's located now, I read it as making the entire title group optional.

Ah, yes thank you, it should be (I did my regex this morning and I already forgot about it)

picnixz:
> Quick size analysis, switching to my newline-delimited proposal would increase the current Python 3.12 compressed inventory by about 2%:

I see. If we include the title length, I think we'll probably have less (in the end, it's as if you had a longer title with at most 3 characters (because, who in their right mind would have a title of more than 1000 characters)).

bskinn:
> If we include the title length, I think we'll probably have less

Yes, agreed, yours is much more concise.

> because, who in their right mind would have a title of more than 1000 characters

Yessss, goodness.

I like your proposal far better than continuing work to find a more robust regex. (I rather suspect that it might _always_ be possible to find a pathological example for any regex we might devise, and any time spent attempting to refine the regex is likely better spent on something else.)

bskinn:
> I'd be glad to have it. It could help a lot for [#12204](https://github.com/orgs/sphinx-doc/discussions/12204)

Ok, here's what I turned up, in no particular order:

- Possible colon in role:
  - https://github.com/bskinn/sphobjinv/issues/256
  - https://github.com/bskinn/sphobjinv/pull/258
- Empty dispname is NOT possible
  - https://github.com/bskinn/sphobjinv/issues/190
  - https://github.com/bskinn/sphobjinv/pull/189
    - This PR also closes the following three issues, 181/182/183
- Priority must be an integer
  - https://github.com/bskinn/sphobjinv/issues/182
- URI can be absent
  - https://github.com/bskinn/sphobjinv/issues/183
- Names can have spaces (this is not news 😅)
  - https://github.com/bskinn/sphobjinv/issues/181
- Encodings
  - https://github.com/bskinn/sphobjinv/issues/104
  - https://github.com/bskinn/sphobjinv/issues/103#issuecomment-628690714
- In the header, it's possible for each of `project` and `version` to be blank
  - https://github.com/bskinn/sphobjinv/issues/85
- Why `sphobjinv` doesn't support v1 `objects.inv` files
  - https://github.com/bskinn/sphobjinv/issues/69
- Discussion of automatic directive -> role transformation in `suggest` output
  - https://github.com/bskinn/sphobjinv/issues/234


I also have an entire [test module] in `sphobjinv` dedicated to testing values that do/don't survive passage through `sphinx.util.inventory.InventoryFile`. It's very likely that there are errors here, but they're thorough/correct enough that they handle much of what's out in the wild.


[test module]: https://github.com/bskinn/sphobjinv/blob/main/tests/test_valid_objects.py

PR #11933, opened 2024-02-03
[autodoc] allow undocumented attributes in `__init__` to be picked up by `:undoc-members:`

### LINKED ISSUE #10193: `:undoc-members:` option doesn't work for instance attributes
### Describe the bug

The `:undoc-members:` option does not include instance attributes without docstrings when used with the `.. autoclass::` directive.

### How to Reproduce

**example.py**
```py
class Example:
    """A simple example."""

    def __init__(self, one_arg: str = "value"):

        self.instance_var1 = 123
        self.instance_var2 = one_arg
```

**example.rst**
```rst
example
*******

.. autoclass:: example.Example
    :members:
    :undoc-members:
```

**conf.py**
```py
import os
import sys
sys.path.insert(0, os.path.abspath(os.path.join('..', '..')))

html_theme = 'sphinx_rtd_theme'

extensions = [
    'sphinx.ext.autodoc',
    'sphinx_rtd_theme'
]

autoclass_content = 'class'
autodoc_class_signature = "mixed"
autodoc_typehints = 'signature'
autodoc_preserve_defaults = True
add_module_names = False
add_function_parentheses = True
autodoc_typehints_description_target="all"
```

### Expected behavior

Instance attributes without docstrings should be included by autodoc in the documentation if the `:undoc-members:` option is specified.

### Your project

Personal project

### Screenshots

undoc_members

### OS

Windows 10 Pro

### Python version

3.9.0

### Sphinx version

4.4.0

### Sphinx extensions

autodoc

### Extra tools

None

### Additional context

_No response_

picnixz:
Fix #10193.

Due to various side-effects, the test suite must currently be executed in its own (see #11285).


sphinx/pycode/parser.py
```diff
@@ -396,6 +396,15 @@ def visit_Assign(self, node: ast.Assign) -> None:
 
         # not commented (record deforders only)
         for varname in varnames:
+            if (
+                # TODO(picnixz): decide whether to include special names or not
```
picnixz, line 400: I am considering possibly picking attributes defined in `__new__` as well but I think it should be another PR. I don't remember why I used "special" names though...

PR #11935, opened 2024-02-03
C domain: Fix namespace-pop context

ferdnyc:
The ref_context key is `c:parent_key`, not `cp:parent_key`.

Subject: Correct a `ref_context` key in `CNamespacePop`

### Feature or Bugfix

- Bugfix

### Purpose
- Correct a `ref_context` item named `cp:parent_key`, which the `CNamespacePop` directive was seetting when restoring the context.

### Detail
- The erroneous key here was the only instance in the entire codebase of that name being used. All others are `c:parent_key`.
- This appears to have likely been a typo, made when converting the code from the C++ domain and editing the key down from `cpp:parent_key`.


ferdnyc:
A Windows failure, once again, in the unrelated test `tests/test_intl/test_intl.py::test_gettext_dont_rebuild_mo`.

PR #11936, opened 2024-02-04
Fix typos

tk0miya:
### Feature or Bugfix
- Refactoring

### Purpose
- Fix typo

Hello. I'm back from a long break.


sphinx/registry.py
```diff
@@ -59,7 +59,7 @@
 class SphinxComponentRegistry:
     def __init__(self) -> None:
         #: special attrgetter for autodoc; class object -> attrgetter
-        self.autodoc_attrgettrs: dict[type, Callable[[Any, str, Any], Any]] = {}
+        self.autodoc_attrgetters: dict[type, Callable[[Any, str, Any], Any]] = {}
```
picnixz, line 58: This will break important packages such as SAGE math:

https://github.com/sagemath/sage/blob/0c390a06f725cfeef11fb7f8430a0f451250af6d/src/sage_docbuild/ext/sage_autodoc.py#L2893-L2899

I would suggest keeping `autodoc_attrgettrs` as a (writable and readable) property but emit a RemovedInSphinx90Warning whenever it is accessed.

picnixz (dismissed):
Welcome back ! here are some comments

sphinx/registry.py
tk0miya: Oops. You're right. I'll add proxy methods.

sphinx/registry.py
```diff
@@ -132,6 +132,10 @@ def __init__(self) -> None:
         # private cache of Docutils Publishers (file type -> publisher object)
         self.publishers: dict[str, Publisher] = {}
 
+    @property
+    def autodoc_attrgettrs(self) -> dict[type, Callable[[Any, str, Any], Any]]:
+        return self.autodoc_attrgetters
```
picnixz, line 137:
```suggestion
		"""Backward-compatible alias to 'autodoc_attrgetters'."""
        return self.autodoc_attrgetters
```

Do you want to add a warning?

AA-Turner: I'd merge a PR to do so but I'm just trying to clear the backlog of 'easy' PRs currently.

PR #11937, opened 2024-02-04
Fix #10322: autodoc: Fail on enums containing special characters

### LINKED ISSUE #10322: Autodoc fails on enums containing special characters

tk0miya:
### Feature or Bugfix
- Bugfix

### Purpose
- refs: #10322 
- Filter members having invalid characters in python (like `CMYK;I`).

### DIFF
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -696,6 +696,11 @@ class Documenter:
                     RemovedInSphinx80Warning, stacklevel=2,
                 )
 
+            # filter a member having invalid name
+            match = py_ext_sig_re.match(membername)
+            if not match:
+                break
+
             # if isattr is True, the member is documented as an attribute
             isattr = member is INSTANCEATTR or (namespace, membername) in attr_docs
 
--- a/tests/roots/test-ext-autodoc/target/enums.py
+++ b/tests/roots/test-ext-autodoc/target/enums.py
@@ -21,3 +21,6 @@ class EnumCls(enum.Enum):
     def say_goodbye(cls):
         """a classmethod says good-bye to you."""
         pass
+
+
+ColorSpace = enum.Enum("ColorSpace", "RGB RGBA CMYK CMYK;I")
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -77,6 +77,7 @@ Bugs fixed
 * #11925: Blacklist the ``sphinxprettysearchresults`` extension; the functionality
   it provides was merged into Sphinx v2.0.0.
   Patch by James Addison.
+* #10322: autodoc: Fail on enums containing special characters
 
 Testing
 -------
--- a/tests/test_extensions/test_ext_autodoc.py
+++ b/tests/test_extensions/test_ext_autodoc.py
@@ -1463,6 +1463,33 @@ def test_enum_class(app):
     ]
 
     # checks for an attribute of EnumClass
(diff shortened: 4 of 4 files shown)


tests/roots/test-ext-autodoc/target/enums.py
```diff
@@ -21,3 +21,6 @@ def say_hello(self):
     def say_goodbye(cls):
         """a classmethod says good-bye to you."""
         pass
+
+
+ColorSpace = enum.Enum("ColorSpace", "RGB RGBA CMYK CMYK;I")
```
picnixz, line 26: I'd suggest adding a docstring for this one since enumeration classes have different docstrings depending on the Python's version (or you can just hardcode the default docstring as we did for the arguments list).

sphinx/ext/autodoc/__init__.py
```diff
@@ -696,6 +696,11 @@ def is_filtered_inherited_member(name: str, obj: Any) -> bool:
                     RemovedInSphinx80Warning, stacklevel=2,
                 )
 
+            # filter a member having invalid name
```
picnixz, line 699: Should we really filter out the 'invalid' names? because, they are still valid in Python although we need to access them using `getattr` and not via dots. 

Should we perhaps add some escape sequences and modify the parsers accordingly? (in which case the PR will likely be much bigger). If we go with your approach, I'll suggest adding some documentation saying that exotic member names will not be rendered (for now).

tk0miya: I feel showing `Foo.invalid;attr` in API docs is not good because users can't access such an invalid attribute. As you said, users can access it in other ways. But it's better to mention the better approach in the document, I think.

So I prefer to add a note to the doc of autodoc. What do you think?

I noticed this would be a breaking change. So it would be better to merge this into the next major version. Please let me know your thoughts.

picnixz: Yes, let us say that in the docs, namely, we don't support rendering those weird attributes. 

> I noticed this would be a breaking change. So it would be better to merge this into the next major version. Please let me know your thoughts.

Currently, do we even manage to render those weird fields? If so, I think we don't beautifully render them so I don't know whether it's really a "breaking" change or not. Neverthless, people may perhaps have some hacky approaches so to be on the safe side, we could merge this PR in 8.0 and then maybe work for a better rendering.

PR #11939, opened 2024-02-04
Fix pickled doctrees cache.

### LINKED ISSUE #11474: Latex file is not updated when using the option numfig = True
### Describe the bug

Latex file is not updated when using the option numfig = True.

### How to Reproduce

The bug can be reproduced with the following bash code:

```bash
# generate a default project
sphinx-quickstart --no-sep -p "myproject" -a test -r "" -l en
# add numfig = True option to conf.py
echo -e "\nnumfig=True" >> conf.py
# build latex & print main latex file modification time
make latex && ls _build/latex/myproject.tex --full-time
# now add a change to index.rst
echo -e "\n*new change*" >> index.rst
# build latex & print main latex file modification time again
make latex && ls _build/latex/myproject.tex --full-time
```

In the code above, after updating index.rst,  when running `make latex && ls _build/latex/myproject.tex --full-time`  the file `myproject.tex` is not updated.

The output of `make latex` in the terminal seems normal (I see no difference with respect to running without setting numfig = True, just that the tex file is not updated), 

The workaround for this is running sphinx-build the -E option (don't use a saved environment): `sphinx-build -E -b latex . _build/latex/`

### Environment Information

```text
Platform:              linux; (Linux-6.2.14-300.fc38.x86_64-x86_64-with-glibc2.37)
Sphinx version:        7.0.1
Docutils version:      0.19
Jinja2 version:        3.1.2
Pygments version:      2.14.0
```

### Sphinx extensions

_No response_

### Additional context

_No response_

### LINKED ISSUE #11908: no rebuild after file changes are detected
### Describe the bug

I currently update out company template from 5.3.0 to 7.2.6 (7.3.0 is planned as final version). The I noticed that documents are not rebuild. Sphinx do detect changes but somehow latexmk thinks there is nothing to do. 
It is possible to do a make clean and then a rebuild. But as we often build multiple documents changes last 3 minutes instead 10 seconds. During documentation development this sums up to a lot wait time.

I tracked the point down to the 6.0.1 -> 6.1.0 update.

With 6.0.1 a change is detected and the document is rebuiild.
sphinx_6_0_1_rebuildAfterChange

Wit 6.1.0 the change is also detected but no rebuild.
sphinx_6_1_0_noRebuildAfterChange

I found the bug #11474 which describes basically the same behavior but during my tests changing the numfig setting do not show any affect.
I also tested Sphinx 7.1.0 due to the mentioned bug and the latest version 7.2.6.  The behavior is the same as 6.1.0.

### How to Reproduce

I have no small example.  Can do it if needed.

```
make venv
make latexpdf
change one or multiple files
make latexpdf
```

### Environment Information

```text
Platform:              linux; (Linux-4.19.0-26-amd64-x86_64-with-glibc2.35)
Python version:        3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0])
Sphinx version:        6.0.1
Docutils version:      0.18.1
Jinja2 version:        3.1.3
```

```text 
Platform:              linux; (Linux-4.19.0-26-amd64-x86_64-with-glibc2.35)
Python version:        3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0])
Sphinx version:        6.1.0
Docutils version:      0.18.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

```text 
Platform:              linux; (Linux-4.19.0-26-amd64-x86_64-with-glibc2.35)
Python version:        3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0])
Sphinx version:        7.1.0
Docutils version:      0.20.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

```text 
Platform:              linux; (Linux-4.19.0-26-amd64-x86_64-with-glibc2.35)
Python version:        3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0])
Sphinx version:        7.2.6
Docutils version:      0.20.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

### Sphinx extensions

**requirements.txt for test with Sphinx 6.0.1 and 6.1.0**
```python
docutils==0.18.1
sphinx_rtd_theme==2.0.0
sphinx-autobuild==2021.3.14
sphinxcontrib-bibtex==2.5.0
sphinxcontrib-globalsubs==0.1.1
sphinxcontrib-scalebybuilder==0.1.1
sphinxcontrib-svg2pdfconverter==1.2.2
sphinxcontrib-wavedrom==3.0.4
```

**requirements.txt for test with Sphinx 7.1.0 and 7.2.6**
```python
sphinx_rtd_theme==2.0.0
sphinx-autobuild==2021.3.14
sphinxcontrib-bibtex==2.6.2
sphinxcontrib-globalsubs==0.1.1
sphinxcontrib-scalebybuilder==0.1.1
sphinxcontrib-svg2pdfconverter==1.2.2
sphinxcontrib-wavedrom==3.0.4
```

**Extension**
```python
extensions = [
    'sphinx_rtd_theme',
    'sphinx.ext.intersphinx',
    'sphinx.ext.autodoc',
    'sphinx.ext.mathjax',
    'sphinx.ext.viewcode',
    'sphinx.ext.todo',
    'sphinxcontrib.bibtex',
    'sphinxcontrib.inkscapeconverter',
    'sphinxcontrib.scalebybuilder',
    'sphinxcontrib.globalsubs',
    'sphinxcontrib.wavedrom'
]
```

### Additional context

_No response_

picnixz:
Closes #11474.
Closes #11908.

Note that this fix may increased build time for very large projects since we need to revert back to reading at least twice the pickled files. 

Alternatively, we could add the possibility to mark which document should be 'refreshed' from its file if needed? though I don't know which ones we should actually target. Maybe @AA-Turner has some solution for this?


picnixz:
For now, I'll merge this since I don't see any other simple solution. It will increase the build time of all projects, yet it will not cause any issues with rebuilds. However, I assume that it should not be too stressing to read 70ks pickled doctrees on modern systems (I could have been concerned 30 years ago but I think the performance drop is marginal).

aakhavanQC:
Is there an ETA on the release for this? TIA

aakhavanQC:
> Err... technically, it's fixed on master. However, I don't know when we'll be able to release 7.3 so.. I cannot give an ETA on the 7.3 release, sorry :(

No worries. Do you think I could patch this in to 7.2.6 without any other issues? Seems simple enough...

picnixz:
> No worries. Do you think I could patch this in to 7.2.6 without any other issues? Seems simple enough...

Mmh. I don't think it would affect other stuff in the code. At least it shouldn't. So technically, you could patch your local installation. However, I'd recommend just installing Sphinx 7.3 using `pip` with `git+` syntax (in a virtual environment for instance or in a local folder, not system-widely otherwise you could mess up your system installation).

aakhavanQC:
> > No worries. Do you think I could patch this in to 7.2.6 without any other issues? Seems simple enough...
> 
> Mmh. I don't think it would affect other stuff in the code. At least it shouldn't. So technically, you could patch your local installation. However, I'd recommend just installing Sphinx 7.3 using `pip` with `git+` syntax (in a virtual environment for instance or in a local folder, not system-widely otherwise you could mess up your system installation).

Unfortunately it breaks one of my extensions (autodoc_pydantic), because it looks like it's trying to import `py_sig_re` from `sphinx.domains.python` but it's not in the module anymore because it was refactored: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/domains/python/__init__.py

Is this a deliberate API change? I can report it upstream or make a new issue here.

```
Extension error:
Could not import extension sphinxcontrib.autodoc_pydantic (exception: cannot import name 'py_sig_re' from 'sphinx.domains.python' (/prj/iceng/pdk/work_aakhavan/devicedata/.venv/lib/python3.11/site-packages/sphinx/domains/python/__init__.py))
```

Used here: https://github.com/mansenfranzen/autodoc_pydantic/blob/c07d1f2223897d2fa5983423bbe73ac73a624fee/sphinxcontrib/autodoc_pydantic/directives/directives.py#L14

picnixz:
Ah, yes. I see that we really need to be on the same page on what is public API and what's not. So here's an example of something that we think as private API (non-capitalized module-level constant) however users as using it as if it were public (although not documented in the RST documentation).

Most of those public/private API issues stem from 10y.o. code base where people had different conventions and where we did not expect things to be used if not explicitly documented. Personally, I don't expect users to use non-capitalized module-level constants as if they were public API. 

(long story short, the rationale behind the refactorization is that the Python domain was too big of a file so many things were moved into 'private' API because we did not expect them to be used directly by extensions but it appears we are breaking things, so maybe we should re-export them normally but deprecate them).

I'm not sure you should report this upstream since it's our fault I would say. We did not clarify what is public/private so we should somewhat fix that before releasing 7.3

chrisjsewell:
> I don't expect users to use non-capitalized module-level constants as if they were public API.

Not heard that one before 😅 
as far as I was aware, the only globally agreed on convention is that if it starts with `_` then it's private

picnixz:
>  as far as I was aware, the only globally agreed on convention is that if it starts with _ then it's private

Well.. technically, PEP 8 tells you that:

> Documented interfaces are considered public, unless the documentation explicitly declares them to be provisional or internal interfaces exempt from the usual backwards compatibility guarantees. All undocumented interfaces should be assumed to be internal.

However, they also say that you should use underscored names. For constants, they recommend you to use capital letters. So I would assume that non-capitalized constants are not meant to be publicly used. Anyway, I think we should create a separate issue to make anything that could have introduced breaking changes and make them explicitly deprecated (and after that, we'll be good and won't have problems anymore I think).

riccardomanfrin:
Affected as well, have to `rm -rf` the latex `_build` and wait for the full regeneration of the doc

PR #11940, opened 2024-02-04
[gettext] fix flaky test on windows

picnixz:
This is a PR for triggering the CI/CD. I'll try to see if it changes anything (and to check if a special statement caused the regular issues). So I'll likely leave that PR open and try to retrigger CI/CD a few times to make it fail.


jayaddison:
Something I've been wondering about: is the `.mo` file genuinely a source dependency of the project during rebuilds?

It _seems_ like the failing test is trying to say that "no, modifying the .mo file should _not_ result in detection of files-to-update".

But I think that some of the logic in the codebase (notably [here](https://github.com/sphinx-doc/sphinx/blob/ceb3b2a904cb14f959917873e8be91fb194ca393/sphinx/environment/__init__.py#L442-L456)) _does_ seem to consider `mo` files as dependencies (at least when the builder is configured accordingly).

picnixz:
Well the issue is that it sometimes works and not on Windows. I suspect that it's because when we are querying the last modification time, something happens. So I put some debug prints here and there to check. 

As for the dependency, I don't know but I think yes.

picnixz:
Ok, so I know what happens:

- When we read the documents, we put some timestamp to indicate when it was read. We use `time.time_ns() // 1000` which gives us a *lower-bound* on the real time since EPOCH in microseconds. In particular, the time of the document is either the one that we stored or something a bit *later*. 

- When we want to test for the last modification time, we extract the closest microsecond of the last modification time as given by the OS in nanoseconds. So, we do `-(ns // -1000)`. 

Now, the failure that I could extract has a `time.time_ns()` equal to `T = 1707062261168486400`. So, we will store `T // 1000 = 1707062261168486`. The exact value would be `1707062261168486.4`. Now, if we compare it to a document for which `os.stat(file).st_mtime_ns` also returns `T`, we actually compute `T' = -(T // -1000) = 1707062261168487` and here is the little issue. 

It should be fine to err for the release build because better be safe than sorry. But for the tests, since those failures appear quite often, we always get a notification on the fact that the workflow fails. Also, people might get confused because they don't know whether what they did affected this component or not sometimes. So I suggest that, for the tests only, we could mock `time.time_ns` for the specific tests and only for Windows platforms.

Actually the whole issue stems from the fact that `time.time_ns()` has a resolution of [318 microseconds on Windows](https://peps.python.org/pep-0564/#windows) vs [84ns on Unix-based systems](https://peps.python.org/pep-0564/#linux). That's likely the reason why my two files end up having the same `st_mtime_ns` value but not on Unix (since they are effectively created at different points in time). If Windows had a good `time.time_ns()` resolution, then this issue would likely disappear.

picnixz:
Closing since I'm done with my investigation.

picnixz:
So let's work again on that one.

picnixz:
I think it now works?

tests/test_intl/test_intl.py
```diff
@@ -681,50 +687,73 @@ def test_translation_progress_classes_true(app):
...
-    # Even if the timestamp of the mo file is updated, the number of documents
-    # to be updated is 0. gettext builder does not rebuild because of mo update.
-    os.utime(app0.srcdir / 'xx' / 'LC_MESSAGES' / 'bom.mo', (mtime + 10, mtime + 10))
-    update_targets = get_update_targets(app)
-    assert update_targets[1] == set(), update_targets
+    with monkeypatch.context() as m:
+        if os.name != 'posix':
+            offset = time.time_ns()
+            t0 = time.perf_counter_ns()
+            monkeypatch.setattr(time, 'time_ns', lambda: offset + time.perf_counter_ns() - t0)
```
jayaddison: > Actually the whole issue stems from the fact that time.time_ns() has a resolution of [318 microseconds on Windows](https://peps.python.org/pep-0564/#windows) vs [84ns on Unix-based systems](https://peps.python.org/pep-0564/#linux). That's likely the reason why my two files end up having the same st_mtime_ns value but not on Unix (since they are effectively created at different points in time). If Windows had a good time.time_ns() resolution, then this issue would likely disappear.

During this monkeypatching, does `time.perf_counter_ns` act as the improved-resolution timer that you mention?

And if so: why is it necessary to _also_ increase the interval size for the modified-timestamps (from `+ 5` to `+ 1000` for example)?

picnixz: I'm trying something now so don't worry about the state of that PR. I'm doing a totally different approach and trying to see what really happens actually (what I pushed does not work...)

picnixz:
I think I fixed the test. But now, there is another issue with monkeypatch + pytest which is probably due to the fact that I'm tempering with `time`.

picnixz:
... I think the fix is correct. I'll eat and cleanup everything. The issue with pytest was due to the teardown phase which is probably messed up because of the monkey patches (or because too many files are written at the same time). At least, the culprit appears to be the `print` which may cause some colorization issues. This is probably a separate issue (I also needed to unroll all the tests to be sure that all my executions are isolated since it's impossible to specify an isolated testroot for now).

picnixz:
I can safely say that the fix now works !

picnixz:
Merging this one should technically (and I hope so) fix the future issues on Windows that we had.

tests/test_intl/test_intl.py
```diff
@@ -681,50 +682,188 @@ def test_translation_progress_classes_true(app):
...
+    # to ensure that calls to time.time_ns() remain consistent.
+    clock.sleep(0.1 if os.name == 'posix' else 1)
+
+    # check that the source files were not modified
+    assert bom_rst.stat().st_mtime_ns == bom_rst_time
+    assert index_rst.stat().st_mtime_ns == index_rst_time
+    # check that the 'bom' document is discovered after the .mo
+    # file has been written on the disk (i.e., read_doc() is called
+    # after the creation of the .mo files)
+    assert app.env.all_docs['bom'] > mo_path.stat().st_mtime
```
jayaddison:
```suggestion
    assert app.env.all_docs['bom'] > mo_path.stat().st_mtime_ns / 1_000
```

tests/test_intl/test_intl.py
```diff
@@ -681,50 +682,188 @@ def test_translation_progress_classes_true(app):
...
+    catalog_write_mo = CatalogInfo.write_mo
+
+    def mock_write_mo(self, locale, use_fuzzy=False):
+        catalog_write_mo(self, locale, use_fuzzy)
+        # ensure that the .mo file being written has a correct fake timestamp
+        _set_mtime_ns(self.mo_path, time.time_ns())
+
+    # see: https://github.com/pytest-dev/pytest/issues/363
+    with pytest.MonkeyPatch.context() as mock:
+        if os.name == 'posix':
```
jayaddison, line 747: Is this preferable to checking `sys.platform`?

picnixz: That's how we skip tests so I think it's fine.

jayaddison:
FYI: I've got a branch where I've attempted to keep all the 'fix' parts of this, while filtering-out the refactor-like changes.

That helped me to understand that the key parts are both the time-mocking and also the very-specific modified-time patch _after_ the BOM file has been written but _before_ it is re-read.

It's currently visible for comparison at: https://github.com/sphinx-doc/sphinx/compare/master...jayaddison:sphinx:issue-11941/pr-11940-review-experimentation#files_bucket (although I'll probably remove that branch at some point)

I did also make one or two changes during that - nothing too significant.  I'll try to remember to provide the relevant things as code-review feedback.

tests/test_intl/test_intl.py
```diff
@@ -19,10 +18,14 @@
 from sphinx.testing.util import assert_node, etree_parse, strip_escseq
 from sphinx.util.nodes import NodeMatcher
 
+_CATALOG_LOCALE = 'xx'
+
+py_time_sleep = time.sleep
+
```
jayaddison:
```suggestion
```

(a small cleanup)

tests/test_intl/test_intl.py
```diff
@@ -681,50 +682,188 @@ def test_translation_progress_classes_true(app):
...
-    # erase doctreedir before gettext build.
-    shutil.rmtree(app0.doctreedir)
+    # remove all sources for the next test
+    shutil.rmtree(app.srcdir, ignore_errors=True)
+    time.sleep(0.1 if os.name == 'posix' else 0.5)  # real sleep
+
+
+@sphinx_intl
+@pytest.mark.sphinx('gettext', testroot='builder-gettext-dont-rebuild-mo', freshenv=True)
+def test_gettext_dont_rebuild_mo(mock_time_and_i18n, app):
```
jayaddison, line 824: In some ways it makes sense to re-separate these tests; however I think the original intent was to confirm that after using one builder to create a `.mo` file, the second builder _does not_ rebuild that exact same `.mo` file.

tests/test_intl/test_intl.py
```diff
@@ -875,15 +1014,16 @@ def test_html_rebuild_mo(app):
...
+    assert updated == set()
 
-    mtime = (app.srcdir / 'xx' / 'LC_MESSAGES' / 'bom.mo').stat().st_mtime
-    os.utime(app.srcdir / 'xx' / 'LC_MESSAGES' / 'bom.mo', (mtime + 5, mtime + 5))
-    app.env.find_files(app.config, app.builder)
-    _, updated, _ = app.env.get_outdated_files(config_changed=False)
-    assert len(updated) == 1
+    _, bom_file = _get_bom_intl_path(app)
+    old_mtime = bom_file.stat().st_mtime
+    new_mtime = old_mtime + (dt := 1000)
```
jayaddison:
```suggestion
    new_mtime = old_mtime + (dt := 5)
```

tests/test_intl/test_intl.py
```diff
@@ -681,50 +682,188 @@ def test_translation_progress_classes_true(app):
...
+
+    # check that the source files were not modified
+    assert bom_rst.stat().st_mtime_ns == bom_rst_time
+    assert index_rst.stat().st_mtime_ns == index_rst_time
+    # check that the 'bom' document is discovered after the .mo
+    # file has been written on the disk (i.e., read_doc() is called
+    # after the creation of the .mo files)
+    assert app.env.all_docs['bom'] > mo_path.stat().st_mtime
 
-    # phase1: build document with non-gettext builder and generate mo file in srcdir
```
jayaddison, line 696: Let's either remove both the `phase1` and `phase2` comments, or neither of them (I'd prefer: keep them both).

tests/test_intl/test_intl.py
picnixz: We are actually deleting the doctrees by the first one. We use the dummy builder to actually write the po files but that could be done separately. 

Previously, we checked that the dummy builder did indeed rebuild the files and that the gettext does not. But since we only care about the file but not its content nor from where it comes, it shouldn't matter.

tests/test_intl/test_intl.py
picnixz: Ah, I think I forgot to write them again so I'll do it tomorrow

tests/test_intl/test_intl.py
```diff
@@ -681,50 +680,188 @@ def test_translation_progress_classes_true(app):
...
+) -> tuple[pytest.MonkeyPatch, _MockClock]:
+    from sphinx.util.i18n import CatalogInfo
+
+    # save the 'original' definition
+    catalog_write_mo = CatalogInfo.write_mo
+
+    def mock_write_mo(self, locale, use_fuzzy=False):
+        catalog_write_mo(self, locale, use_fuzzy)
+        # ensure that the .mo file being written has a correct fake timestamp
+        _set_mtime_ns(self.mo_path, time.time_ns())
```
jayaddison, line 743: @picnixz I was thinking about this part of the PR again - this is where most of the repair happens, if I understand correctly.

It does two things:

  * Retrieves a mocked, monotonic timestamp by calling `time.time_ns` - this is guaranteed to be less than the next value retrieved.
  * Ensures that the timestamp is the one written to the modified-time on the `.mo` file on the filesystem.

Instead of patching over this in a test case where it occurs -- I wonder whether we should move the platform-specific-fix (even though it is ugly) into the Sphinx application code?

I tried that with a patch that includes:

```diff
diff --git a/sphinx/util/i18n.py b/sphinx/util/i18n.py
index 004b5d85f..dd0ff1cc5 100644
--- a/sphinx/util/i18n.py
+++ b/sphinx/util/i18n.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
--- a/sphinx/util/i18n.py
+++ b/sphinx/util/i18n.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import os
 import re
+import time
 from datetime import datetime, timezone
 from os import path
 from typing import TYPE_CHECKING, Callable, NamedTuple
@@ -68,6 +69,16 @@ class CatalogInfo(LocaleFileInfoBase):
                 write_mo(file_mo, po, use_fuzzy)
             except Exception as exc:
                 logger.warning(__('writing error: %s, %s'), self.mo_path, exc)
+                return
+
+        # Allow 318us of time to pass so that when the env.all_docs dependency
+        # graph is updated, Windows always records a timestamp for the catalog
+        # that is strictly greater-than the written .mo file's mtime_ns.
+        #
+        # Ref: https://github.com/sphinx-doc/sphinx/issues/11941
+        # Ref: https://peps.python.org/pep-0564/#windows
+        if os.name != 'posix':
+            time.sleep(3.18e-7)
 
 
 class CatalogRepository:
```

The intended benefit of doing that would be twofold: it would allow us to avoid the mocking/patching in the tests (simpler), and could (very rarely, I expect) prevent the same issue happening if for some reason projects are rapidly rebuilt on Windows hosts.

(I wouldn't have been able to figure out how to do that without experimenting with the patched test code - and I'm still not 100% certain that it'd be sensible, but it's an idea)

picnixz: I prefer not touching the code base at all. For release builds, the issue is that we detect documents as outdated while the are not necessarily really outdated. So it doesn't hurt to rebuild the docs. 

However, for tests, it's annoying that they fail since we run them quite often. 

Concerning the mock itself, every builder (even the dummy one) create those po/mo files. However, since they write something on the disk, the timestamps are given by the OS directly and the 'time' is given by the OS. However, I want the time always be given by this monotonic fake clock (what I essentially do is forcing 'time' informations that are not thread-relevant to be simulated by my fake clock).

We could have added a sleep after writing the po/mo files but I don't know how much files would be added. I think that there are MO files for *every* different RST file so for projects with thousands of files, this small sleep might be relevant. Note that `time.sleep` might come with a slight overhead and it's not guaranteed that you're actually sleeping what you are requesting (it may be less or more, it depends on the platform). Thus, to be on the safe side, we would probably have something like `time.sleep(0.001)` (and I don't think we parallelize this task).

For what it's worth, I think we shouldn't put delays in the release if they are only there to 'patch' tests. Also, note that we explicitly choose to have microsecond resolution and also allow for rounding errors (see `sphinx.environment._last_modified_time`).

jayaddison: Ok, yep - I agree with that reasoning.

If in future someone _does_ happen to encounter this issue in practice on a Windows-based host, then ideally they can provide a bugreport and test coverage to demonstrate that, and _then_ it could be fixed in-application (as opposed to the other way round, like here, where we have an intermittent problem unrelated to the purpose of the test, and changing the app to cater for that does seem odd).

jayaddison (approved):

picnixz:
I think I'll merge this one next since it's not hard to revert it if there are issues. It's not blocking per se but at least it would reduce the number of random CI/CD failures.

jayaddison:
Nice work @picnixz!

PR #11944, opened 2024-02-04
HTML Search: Use anchor for search preview

### LINKED ISSUE #11943: HTML search results: identical summaries displayed for different link targets
### Describe the bug

My understanding (that could be mistaken) about the [`html_show_search_summary`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_show_search_summary) feature (enabled by default) is that each search result displayed to the user should have a relevant snippet of the content displayed alongside, to aid the user's determination of which result(s) are most relevant.

When the feature is enabled, we make an HTTP request to each result and display a portion of that using JavaScript.

### How to Reproduce

Build the Sphinx documentation locally and then run a basic webserver to host it:

```sh
sphinx.git $ sphinx-build -b html doc _build 
sphinx.git $ cd _build
sphinx.git $ python -m http.server -b 127.0.0.1
```

Open `http://127.0.0.1:8000` in a web browser, and perform a search for the query term `test`.

Observe that multiple results with separate HTML anchor targets appear in the search results.  However: they display the same search snippet (not useful - this is the bug).

image

### Environment Information

```text
Sphinx version:        7.2.6
Docutils version:      0.20.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

### Sphinx extensions

_No response_

### Additional context

Discovered during investigation of #11942.

wlach:
Subject: Use anchor for generating search previews

### Feature or Bugfix

- Bugfix

### Purpose

Attempts to use anchor for generating HTML preview, if available. This ensures that the preview is based on the anchor in the document, for cases where we matched on that title.

### Detail

You can reproduce this bug using the procedure @jayaddison mentioned in #11943

```
sphinx.git $ sphinx-build -b html doc _build 
sphinx.git $ cd _build
sphinx.git $ python -m http.server -b 127.0.0.1
```

Then search for something like "test"

Before:

image

After:

image

Note how the content in the "after" image actually reflects what you'd be jumping to!

### Relates

Closes #11943
Discovered when discussing #11942

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -99,7 +99,7 @@ const _displayItem = (item, searchTerms, highlightTerms) => {
       .then((data) => {
         if (data)
           listItem.appendChild(
-            Search.makeSearchSummary(data, searchTerms)
+            Search.makeSearchSummary(data, searchTerms, anchor)
           );
         // highlight search terms in the summary
         if (SPHINX_HIGHLIGHT_ENABLED)  // set in sphinx_highlight.js
@@ -160,10 +160,15 @@ const Search = {
   _queued_query: null,
   _pulse_status: -1,
 
-  htmlToText: (htmlString) => {
+  htmlToText: (htmlString, anchor) => {
     const htmlElement = new DOMParser().parseFromString(htmlString, 'text/html');
     htmlElement.querySelectorAll(".headerlink").forEach((el) => { el.remove() });
-    const docContent = htmlElement.querySelector('[role="main"]');
+    let docContent = undefined;
+    if (anchor && htmlElement.querySelector(anchor)) {
+      docContent = htmlElement.querySelector(anchor);
+    } else {
+      docContent = htmlElement.querySelector('[role="main"]');
+    }
     if (docContent) return docContent.textContent;
     console.warn(
       "Content block not found. Sphinx search tries to obtain it via '[role=main]'. Could you check your theme or template."
@@ -549,8 +554,8 @@ const Search = {
    * search summary for a given text. keywords is a list
    * of stemmed words.
    */
-  makeSearchSummary: (htmlText, keywords) => {
-    const text = Search.htmlToText(htmlText);
+  makeSearchSummary: (htmlText, keywords, anchor) => {
+    const text = Search.htmlToText(htmlText, anchor);
     if (text === "") return null;
 
     const textLower = text.toLowerCase();
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -163,13 +163,19 @@ const Search = {
   htmlToText: (htmlString, anchor) => {
     const htmlElement = new DOMParser().parseFromString(htmlString, 'text/html');
     htmlElement.querySelectorAll(".headerlink").forEach((el) => { el.remove() });
-    let docContent = undefined;
-    if (anchor && htmlElement.querySelector(anchor)) {
-      docContent = htmlElement.querySelector(anchor);
-    } else {
-      docContent = htmlElement.querySelector('[role="main"]');
+    if (anchor) {
+      const anchorContent = htmlElement.querySelector(anchor);
+      if (anchorContent) return anchorContent.textContent;
+
+      console.warn(
+        `Anchor block not found. Sphinx search tries to obtain it via '${anchor}'. Check your theme or template.`
+      );
     }
+
+    // if anchor not specified or not found, fall back to main content
+    const docContent = htmlElement.querySelector('[role="main"]');
     if (docContent) return docContent.textContent;
+
     console.warn(
       "Content block not found. Sphinx search tries to obtain it via '[role=main]'. Could you check your theme or template."
     );
--- a/tests/js/searchtools.js
+++ b/tests/js/searchtools.js
@@ -32,32 +32,34 @@ describe('Basic html theme search', function() {
 });
 
 describe("htmlToText", function() {
+
+  const testHTML = `<html>
+  <div class="body" role="main">
+    <section id="getting-started">
+      <h1>Getting Started</h1>
+      <p>Some text</p>
+    </section>
+    <section id="other-section">
+      <h1>Other Section</h1>
+      <p>Other text</p>
+    </section>
+    <section id="yet-another-section">
+      <h1>Yet Another Section</h1>
+      <p>More text</p>
+    </section>
+  </div>
+  </html>`;
+
   it("basic case", () => {
-    expect(Search.htmlToText(`
-    <html>
-    <div class="body" role="main">
-      <section id="getting-started">
-        <h1>Getting Started</h1>
-        <p>Some text</p>
-      </section>
-    </div>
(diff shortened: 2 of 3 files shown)


sphinx/themes/basic/static/searchtools.js
```diff
@@ -160,10 +160,15 @@ const Search = {
...
   _pulse_status: -1,
 
-  htmlToText: (htmlString) => {
+  htmlToText: (htmlString, anchor) => {
     const htmlElement = new DOMParser().parseFromString(htmlString, 'text/html');
     htmlElement.querySelectorAll(".headerlink").forEach((el) => { el.remove() });
-    const docContent = htmlElement.querySelector('[role="main"]');
+    let docContent = undefined;
+    if (anchor && htmlElement.querySelector(anchor)) {
+      docContent = htmlElement.querySelector(anchor);
```
jayaddison: I have a concern here: I think this is going to retrieve the text content within the anchor element?

In practice we may want to retrieve some of the following HTML sibling element text as well.

We do have some [test coverage on `searchtools.js`](https://github.com/sphinx-doc/sphinx/blob/ceb3b2a904cb14f959917873e8be91fb194ca393/tests/js/searchtools.js) and I wonder whether we could use that to make some assertions about what we retrieve and summarize (and also safeguard the feature against regressions).

wlach: I don't think we want the sibling element. The anchor defines a section within the documentation. I don't think we want (or need) any text outside of it: it's (1) likely to be out of context and (2) probably outside of the [text we'd take for the snippit](https://github.com/wlach/sphinx/blob/27f4d3605989ea2216c5f370b07833a501de7267/sphinx/themes/basic/static/searchtools.js#L563) except in edge cases.

Here's the section for the 2to3 example you gave earlier:

image

jayaddison: Ok, brilliant.  I'm doing a few more tests locally, and getting around to agreeing with all this.

sphinx/themes/basic/static/searchtools.js
```diff
@@ -160,10 +160,15 @@ const Search = {
...
+    } else {
+      docContent = htmlElement.querySelector('[role="main"]');
+    }
     if (docContent) return docContent.textContent;
     console.warn(
       "Content block not found. Sphinx search tries to obtain it via '[role=main]'. Could you check your theme or template."
```
jayaddison, line 180: Could you update this warning message to make sure it's not misleading in cases where an anchor is provided?

wlach: We only match on the anchor if it's (1) there and (2) querySelector actually matches some content, otherwise we always fall back to the main document. So I think this warning message is still valid as-is.

jayaddison: If we display a result where the hyperlink contains an anchor, and neither the anchor nor the `[role=main]` content are found on the page, then I think it could be confusing to tell a developer/user that `[role=main]` content was missing.

It's partially true, but it doesn't necessarily indicate how to completely resolve the (two-fold) problem.

wlach:
@jayaddison Ready for another look whenever you are. Added some tests while I was here

tests/js/searchtools.js
```diff
@@ -31,6 +31,36 @@ describe('Basic html theme search', function() {
...
+      <section id="getting-started">
+        <h1>Getting Started</h1>
+        <p>Some text</p>
+      </section>
+      <section id="other-section">
+        <h1>Other Section</h1>
+        <p>Other text</p>
+      </section>
+    </div>
+    </html>`, '#other-section').trim().split(/\s+/)).toEqual(['Other', 'Section', 'Other', 'text']);
```
jayaddison: Could you use this single (larger) HTML content as the input for both test cases?  I think that'll make it clearer that selecting a summary with-or-without an anchor produces different results (in fact, ideally we might want to test without an anchor, with one anchor, and then with a second anchor -- and confirm that cases two and three produce different summaries).

tests/js/searchtools.js
```diff
@@ -31,6 +31,38 @@ describe('Basic html theme search', function() {
...
+    </section>
+    <section id="yet-another-section">
+      <h1>Yet Another Section</h1>
+      <p>More text</p>
+    </section>
+  </div>
+  </html>`;
+
+  it("basic case", () => {
+    expect(Search.htmlToText(testHTML).trim().split(/\s+/)).toEqual([
```
jayaddison, line 54: I have some ideas about the reason, but just to check they match reality: why are we testing `htmlToText` and not `makeSearchSummary`?

tests/js/searchtools.js
```diff
@@ -31,6 +31,38 @@ describe('Basic html theme search', function() {
...
+      'Getting', 'Started', 'Some', 'text', 
+      'Other', 'Section', 'Other', 'text', 
+      'Yet', 'Another', 'Section', 'More', 'text'
+    ]);
+  });
+
+  it("will start reading from the anchor", () => {
+    expect(Search.htmlToText(testHTML, '#other-section').trim().split(/\s+/)).toEqual(['Other', 'Section', 'Other', 'text']);
```
jayaddison, line 62: Is the `trim...split` behaviour here to avoid being too-precise about whitespace in the output?

If so, an alternative could be to make a less-restrictive assertion; something like `retrievedText.startsWith("Other Section")`.

tests/js/searchtools.js
wlach: I think it's easier to test the lower-level construct to verify that it doesn't have any bugs. makeSearchSummary has a bunch of magic numbers and heuristics in it which seems much more subject to change.

tests/js/searchtools.js
wlach: That's right, just trying not to test whitespace. I think I prefer to check for all the tokens in the output, feels more comprehensive to me.

jayaddison (approved):
Looks good to me - thank you @wlach!

picnixz (approved):
Seems fine to me as well.

picnixz:
@wlach Could you add a changelog entry btw?

wlach:
> @wlach Could you add a changelog entry btw?

Done.

picnixz:
Thank you!

sphinx/themes/basic/static/searchtools.js
```diff
@@ -160,11 +160,22 @@ const Search = {
   _queued_query: null,
   _pulse_status: -1,
 
-  htmlToText: (htmlString) => {
+  htmlToText: (htmlString, anchor) => {
     const htmlElement = new DOMParser().parseFromString(htmlString, 'text/html');
     htmlElement.querySelectorAll(".headerlink").forEach((el) => { el.remove() });
+    if (anchor) {
+      const anchorContent = htmlElement.querySelector(anchor);
```
jayaddison, line 167: Hi @wlach - apologies for not thinking of this during code review: do you think it would make sense to constrain the anchor query further, to the main document content?

```suggestion
      const anchorContent = htmlElement.querySelector(`[role="main"] ${anchor}`);
```

(I thought of this while writing a [comment](https://github.com/sphinx-doc/sphinx/issues/12052#issuecomment-1986935319) about another issue)

jayaddison: Handled by #12105.

Issue #11941, opened 2024-02-04
[CI/CD] Improve `test_gettext_dont_rebuild_mo` on Windows platforms.


picnixz:
People contributing to Sphinx have the surprise that their PRs fail due to some Windows failures. 

- When we read documents, we put some timestamp to indicate when it was read. We use `time.time_ns() // 1000` which gives us a *lower-bound* on the real time since EPOCH in microseconds. In particular, the time of the document is either the one that we stored or something a bit *later*. 

- When we want to test for the last modification time, we extract the closest microsecond of the last modification time as given by the OS in nanoseconds. So, we do `-(ns // -1000)`. 

Now, by playing in #11940, at some point I had `time.time_ns()` equal to `T = 1707062261168486400`. So, we store `T // 1000 = 1707062261168486` (the exact microsecond value could be `1707062261168486.4` but we ignore the fractional parts). Now, if we compare it to a document for which `os.stat(file).st_mtime_ns` also returns `T`, we actually compute `T' = -(T // -1000) = 1707062261168487`. Usually, we don't compare the file with time `T` to a file with time `T' = T` but this entirely boils down to whether `T` and `T'` have a sufficiently good precision or not.

On Linux, it doesn't happen because the documents we are comparing our files to are "older" (the dependencies are created before those on which they depend). However, on Windows, both the dependencies and the dependent have the same `time.time_ns()` value. This is due to the resolution of `time.time_ns()` which is [318 microseconds on Windows](https://peps.python.org/pep-0564/#windows) and [84ns on Unix-based systems](https://peps.python.org/pep-0564/#linux). And I assume that `os.stat(...).st_mtime_ns` should have the same precision on Windows (namely, a not that good). 

It should be fine to err for the release build because better be safe than sorry. But for the tests, since those failures appear quite often, we always get a notification on the fact that the workflow fails. Also, people might get confused because they don't know whether what they did affected this component or not sometimes. So I suggest that, for the tests only, we could mock `time.time_ns` for the specific tests and only for Windows platforms so that it has a better resolution. I'll try something in 10 days or so, unless someone wants to do it.

_Reformulation of https://github.com/sphinx-doc/sphinx/issues/11940#issuecomment-1925822663_

PR #11954, opened 2024-02-05
tests: cleanup: remove unused _setup_intl autouse fixture

### LINKED ISSUE #11953: [tests] _setup_intl autouse fixture writes mo files to unused paths
### Describe the bug

While reading #11940 I was looking through some of the `test_intl.py` code and noticed the [`_setup_intl` `pytest` fixture](https://github.com/sphinx-doc/sphinx/blob/ceb3b2a904cb14f959917873e8be91fb194ca393/tests/test_intl/test_intl.py#L41-L56), which has `autouse=True` (meaning that the fixture is evaluated for all tests in the module - I've checked that that's the case).

The fixture appears to write `.mo` translation catalog files for applications when they have `.po` resource files and where either no `.mo` file exists, or it is out-of-date.

However: the file path that is checked for an `.mo` file -- and subsequently the file path that an `.mo` file is written to - contain a duplicate path structure:

```
/tmp/pytest-of-user/pytest-28/builder-gettext-dont-rebuild-mo/xx/LC_MESSAGES/xx/LC_MESSAGES/bom.mo
```

Because these paths seem incorrect, I think it may be valid to remove the `_setup_intl` fixture entirely; it's performing file I/O but not to paths that are used by other tests.

I've confirmed that unit tests continue to pass locally with the fixture removed.

### How to Reproduce

Apply the following patch to ceb3b2a904cb14f959917873e8be91fb194ca393 (or any relatively recent commit):

```diff
diff --git a/tests/test_intl/test_intl.py b/tests/test_intl/test_intl.py
index 4287182fa..b4109c021 100644
--- a/tests/test_intl/test_intl.py
+++ b/tests/test_intl/test_intl.py
@@ -52,6 +52,10 @@ def _setup_intl(app_params):
                 mo.parent.mkdir(parents=True, exist_ok=True)
 
             if not mo.exists() or os.stat(mo).st_mtime < os.stat(po).st_mtime:
+                print(app_params.kwargs['srcdir'])
+                print(dirpath / f)
+                print(mo)
+                raise
                 # compile .mo file only if needed
                 write_mo(mo, read_po(po))
 
```

And then run a test from the `test_intl.py` module:

```sh
$ pytest tests/test_intl/test_intl.py -k test_gettext_dont_rebuild_mo
```

The test should fail due to the raised exception during the autouse fixture, and the unexpected `.mo` file path will be printed.

### Environment Information

```text
Sphinx version:        7.3.0+/ceb3b2a90
Docutils version:      0.20.1
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

### Sphinx extensions

```python
N/A
```

### Additional context

N/A

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Removes an unused test fixture.

### Detail
- This can wait until after a fix for #11941 is merged, if we'd prefer to ensure that the two are unrelated (I believe they are, but there is no harm in waiting).
- See #11953 for details.

### Relates
- Resolves #11953.

### DIFF
--- a/tests/test_intl/test_intl.py
+++ b/tests/test_intl/test_intl.py
@@ -33,29 +33,6 @@ def read_po(pathname):
         return pofile.read_po(f)
 
 
-def write_mo(pathname, po):
-    with open(pathname, 'wb') as f:
-        return mofile.write_mo(f, po)
-
-
-@pytest.fixture(autouse=True)
-def _setup_intl(app_params):
-    assert isinstance(app_params.kwargs['srcdir'], Path)
-    srcdir = app_params.kwargs['srcdir']
-    for dirpath, _dirs, files in os.walk(srcdir):
-        dirpath = Path(dirpath)
-        for f in [f for f in files if f.endswith('.po')]:
-            po = str(dirpath / f)
-            mo = srcdir / 'xx' / 'LC_MESSAGES' / (
-                os.path.relpath(po[:-3], srcdir) + '.mo')
-            if not mo.parent.exists():
-                mo.parent.mkdir(parents=True, exist_ok=True)
-
-            if not mo.exists() or os.stat(mo).st_mtime < os.stat(po).st_mtime:
-                # compile .mo file only if needed
-                write_mo(mo, read_po(po))
-
-
 @pytest.fixture(autouse=True)
 def _info(app):
     yield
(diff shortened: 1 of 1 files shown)


picnixz:
We can remove the `_setup` (which I did in #11940) but I need the `write_mo` for the patch. We'll wait to see whether the patch I suggested is fine for Adam or not.

PR #11958, opened 2024-02-06
HTML Search: Fix partial matches overwriting full matches

### LINKED ISSUE #11957: HTML Search: Partial matches can overwrite main matches in certain rare cases
### Describe the bug

In the case of the first document in a Sphinx index, partial matches can overwrite the main one. This causes the rank to be artificially lowered.

### How to Reproduce

This is difficult to reproduce manually, I will reproduce it with a unit test when I file a PR.

### Environment Information

### Sphinx extensions

_No response_

### Additional context

_No response_

wlach:
Subject: Fix partial matches overwriting full matches for the 0th document

### Feature or Bugfix
- Bugfix

### Purpose

Fixes matches being artificially lower than they should be. 

### Detail

Due to a small bug, we would overwrite full matches with the partial match score for the first document. This is a trivial bug, but makes testing more difficult.

### Relates
- Closes #11957


sphinx/themes/basic/static/searchtools.js
```diff
@@ -466,14 +466,18 @@ const Search = {
       // add support for partial matches
       if (word.length > 2) {
         const escapedWord = _escapeRegExp(word);
-        Object.keys(terms).forEach((term) => {
-          if (term.match(escapedWord) && !terms[word])
```
wlach, line 481: `terms[word]` will be falsey (`0`) for the first document.

sphinx/themes/basic/static/searchtools.js
```diff
@@ -466,14 +466,18 @@ const Search = {
...
-            arr.push({ files: terms[term], score: Scorer.partialTerm });
-        });
-        Object.keys(titleTerms).forEach((term) => {
-          if (term.match(escapedWord) && !titleTerms[word])
-            arr.push({ files: titleTerms[word], score: Scorer.partialTitle });
-        });
+        if (!terms.hasOwnProperty(word)) {
```
wlach, line 480: It is marginally more efficient to do this check outside the loop too.

tests/js/searchtools.js
```diff
@@ -21,7 +21,7 @@ describe('Basic html theme search', function() {
         "&lt;no title&gt;",
         "",
         null,
-        2,
+        5,
```
wlach, line 24: With this change, the score gets boosted to 5 (what it should be).

sphinx/themes/basic/static/searchtools.js
jayaddison: Unless the performance improvement is significant, I think it'd be better to leave the conditional check after the `&&` clause.

(that's based on considering that if the first condition fails, then the clause after the `&&` isn't evaluated.  if moving the property-check eliminates a lot of regex matches, then fair enough, but this seems intended to be primarily a correctness fixup, so I'd lean towards making the logical diff as small as possible)

wlach: I'm going to push back on this a bit: I think this small improvement is in scope (since we're changing the logic anyway) and the diff here is relatively small and easy to understand. Moving the property check should indeed reduce the number of regex matches a fair bit.

jayaddison: Ok, no problem.  I think we'll have to agree to disagree on that.

jayaddison:
@wlach could you merge the latest changes from the `master` branch into this one?  That should resolve the Windows test failures.  From what I remember of the plan, this is to be merged before #11942.

sphinx/themes/basic/static/searchtools.js
jayaddison: I'd prefer if you could avoid the extraction of an outer `if` statement here.  When viewing the diff, the larger set of line changes obscure what the bugfix is; and that's based on my assessment today, after having understood the fix a couple of weeks ago.  It'd be trickier for someone who wasn't familiar with the original pull request, or looking at it again after a longer duration of time.

wlach: I stand by my original point. If you've convinced yourself that the code works, can we not just move on? We're talking about a <20 line diff here.

wlach:
> @wlach could you merge the latest changes from the `master` branch into this one? That should resolve the Windows test failures. From what I remember of the plan, this is to be merged before #11942.

Done

sphinx/themes/basic/static/searchtools.js
jayaddison: I spend a decent amount of time reading old code changes and commit history, often related to small changes or fixes like this one, and it's usually much easier and less time-consuming to reason about minimal changes (with helpful commit messages ideally).

Narrowing the size of the diff _generally_ allows for increased review/inspection focus under those circumstances.  In this case the _behavioural_ diff isn't huge, but the code diff is larger than required to solve the bug.

In other words, when authoring, minimizing diffs can be a way to demonstrate that you value the time and attention of other developers (even ones who you may never meet or who may not exist yet!).

wlach: Yes, I understood your point, I just don't agree with it in this case. I have more to say on this, but I think we've gone out of scope of a review. If you want to talk about it more, feel free to email me.

jayaddison: Thanks for explaining.  From my perspective, bug report, code review and related discussion are often useful information when trying to understand a commit/change, so I'd prefer to keep the discussion public; in this case I do think that the size of the line diff is relevant to code review.

Would you be OK with me drafting an alternative PR?  I don't know if it'd be more likely to be accepted for merge, and it's OK if it isn't, but I'd like to present it as an alternative.

wlach: > Thanks for explaining. From my perspective, bug report, code review and related discussion are often useful information when trying to understand a commit/change, so I'd prefer to keep the discussion public; in this case I do think that the size of the line diff is relevant to code review.

I don't think I have more to say on the technical front. But just to state it one more time so it's clear: I think this (small) PR as it stands fixes the issue in a more fundamental way and improves performance so it's my strong recommendation that it be the final state of the code.  I don't feel the benefit of breaking it up into two PRs is worth the hassle and churn (leaving aside that this discussion has now occupied more effort than that would entail...).

> Would you be OK with me drafting an alternative PR? I don't know if it'd be more likely to be accepted for merge, and it's OK if it isn't, but I'd like to present it as an alternative.

It's not clear to me when someone who can actually merge the code will look at these changes. I can't stop you from proposing an alternative, but I think the best thing is probably just to wait for a maintainer to break our impasse. I am willing to make the change you suggest if it will actually get this merged and my PR is open to updates from the owners of this repository.

jayaddison:
I think that the problem here is due to an incorrect test fixture in our test suite; I've added one or two more details in the linked bugreport (#11957) and as a result am closing this too.  Please re-open with supporting details if I'm mistaken.

wlach:
> I think that the problem here is due to an incorrect test fixture in our test suite; I've added one or two more details in the linked bugreport (https://github.com/sphinx-doc/sphinx/issues/11957) and as a result am closing this too. Please re-open with supporting details if I'm mistaken.

@jayaddison It looks to me like the index generated by Sphinx only uses an array for the term entries if there are multiple terms. See this example:

https://repo-parser-demo.netlify.app/searchindex.js

And note the output here:

```
{ ..., "terms": {"thi": [0, 3, 4, 5, 10, 12], "i": [0, 6, 10, 12], "an": [0, 7], "type": 0, "you": [0, 3, 11], "might": 0, "abl": 0, ... }
```


I haven't tested, but it looks to me that the code that writes it out is here:

https://github.com/sphinx-doc/sphinx/blob/35965903177c6ed9a6afb62ccd33243a746a3fc0/sphinx/search/__init__.py#L371

I presume it's done this way as a space-saving strategy.

Unfortunately I don't have the rights to re-open this PR and issue, could you please do it for me?

jayaddison:
Yikes - OK.  Thank you @wlach.  That's a cleverer/more-adaptive format than I'd expected, and I should have checked for single-document terms.  Yep, I'll re-open these.

sphinx/themes/basic/static/searchtools.js
jayaddison: @wlach based on index-format findings (and test coverage), should this be rephrased to:

> `terms[word]` will be falsey (`0`) for a word that only exists in document zero

(kinda wordy, but captures the fact that `[0, 1]`, for example, is _not_ falsy)

jayaddison: (edited with a correction and for conciseness)

sphinx/themes/basic/static/searchtools.js
```diff
@@ -477,14 +477,18 @@ const Search = {
...
+          Object.keys(terms).forEach((term) => {
+            if (term.match(escapedWord))
+              arr.push({ files: terms[term], score: Scorer.partialTerm });
+          });
+        }
+        if (!titleTerms.hasOwnProperty(word)) {
+          Object.keys(titleTerms).forEach((term) => {
+            if (term.match(escapedWord))
+              arr.push({ files: titleTerms[word], score: Scorer.partialTitle });
```
jayaddison, line 489: Is the access to `titleTerms[word]` here a (separate) bug?

In both this and the previous loop, we're checking whether `term` matches.  And in the previous loop we access `terms[term]` if it does.

wlach: > Is the access to titleTerms[word] here a (separate) bug?

I think #11957 as described is broad enough to describe both. It's the same error: a partial match overwriting a full one.

jayaddison, line 489: Ok.  If we want to fix it here too then:

(for context: see description of #11958)

Edit: suggestion removed; it relates to a separate bug.

wlach: Oh wait, I see what you mean now given https://github.com/sphinx-doc/sphinx/pull/11958/files#r1509999419. Yeah, this looks like a separate bug strictly speaking. This is such a rabbit hole. 😭 

I'd be inclined to fix it in a new PR with its own test (maybe depending on this one).

jayaddison: Yep, no problem, there's a lot going on here :)  As mentioned though, I feel good about the effect on quality that all this will have.

Agreed; I'll create a separate bugreport for this, and then request review for #12037 after this one is merged.

Could you clarify in the description of #11957 that it relates solely to terms that only occur in the zeroth document in the index?  (if my understanding of that is true?)

Edit: fix pr/issues references

jayaddison: Arg, sorry - those linked issues were wrong; I meant to request a change for the description of #11957 - to make the nature of the bug more precise.

jayaddison: Reported separately as #12040.

sphinx/themes/basic/static/searchtools.js
jayaddison: Ok, after finding more time to understand the code, I concede that the performance improvements here are important.  I've opened #12045 with a description of what I believe the problem is.

Also, compared to my suggested alternative #12040, I think that the use of the JavaScript `hasOwnProperty` function is probably better/safer code style, making this pull request preferable.

@wlach could you edit the description of this pull request (#11958) to add 'Resolves #12045'?

(and thanks for your patience!)

jayaddison:
Note: if reviewing this pull request using the GitHub web viewer, then it may be easier to inspect by requesting the diff with whitespace-changes ignored: https://github.com/sphinx-doc/sphinx/pull/11958/files?w=1 (the `w=1` query-string param)

sphinx/themes/basic/static/searchtools.js
picnixz: I wouldn't bother much about this. Here, it's more important to use `hasOwnProperty` instead of `!terms[word]` which could lead to incorrect boolean evaluations (and personally if we can have an easy-to-write improvement because we can isolate a condition from outside a loop, it's fine for me).

picnixz (approved):

sphinx/themes/basic/static/searchtools.js
picnixz: I wouldn't bother much about this. Here, it's more important to use `hasOwnProperty` instead of `!terms[word]` which could lead to incorrect boolean evaluations (and personally if we can have an easy-to-write improvement because we can isolate a condition from outside a loop, it's fine for me).

picnixz:
Thank you!

jayaddison:
Thank you both @wlach @picnixz!

wlach:
Such a relief to finally get this in, thanks all!

PR #11960, opened 2024-02-06
HTML Search: Fix multiple term matching edge case

### LINKED ISSUE #11959: HTML Search: Multiple term matching sometimes breaks
### Describe the bug

Due to a small bug in term matching, multiple term matching on document content doesn't work when the same word appears in both the title and the document.

### How to Reproduce

It's a bit fiddly to reproduce with a real example, since whether you encounter the bug depends on the order that search and title terms are processed. However it's quite easy to reproduce with a unit test, which I'll attach to the fix.

### Environment Information

### Sphinx extensions

_No response_

### Additional context

_No response_

wlach:
Subject: Fix multiple term matching edge case

### Feature or Bugfix

- Bugfix

### Purpose

Fixes multiple matches getting accidentally overwritten in certain rare cases

### Detail

In certain rare cases, term matches could get overwritten due to a logic bug in the search code. Fix this and add a test.

### Relates

Closes #11959

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -511,11 +511,8 @@ const Search = {
 
       // create the mapping
       files.forEach((file) => {
-        if (!fileMap.has(file)) {
-          fileMap.set(file, [word]);
-        } else if (fileMap.get(file).indexOf(word) === -1) {
-          fileMap.get(file).push(word);
-        }
+        if (!fileMap.has(file)) fileMap.set(file, [word]);
+        else if (fileMap.get(file).indexOf(word) === -1) fileMap.get(file).push(word);
       });
     });
 
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -511,9 +511,13 @@ const Search = {
 
       // create the mapping
       files.forEach((file) => {
-        if (fileMap.has(file) && fileMap.get(file).indexOf(word) === -1)
-          fileMap.get(file).push(word);
-        else fileMap.set(file, [word]);
+        if (fileMap.has(file)) {
+          if (fileMap.get(file).indexOf(word) === -1) {
+            fileMap.get(file).push(word);
+          }
+        } else {
+          fileMap.set(file, [word]);
+        }
       });
     });
 
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -511,12 +511,10 @@ const Search = {
 
       // create the mapping
       files.forEach((file) => {
-        if (fileMap.has(file)) {
-          if (fileMap.get(file).indexOf(word) === -1) {
-            fileMap.get(file).push(word);
-          }
-        } else {
+        if (!fileMap.has(file)) {
           fileMap.set(file, [word]);
+        } else if (fileMap.get(file).indexOf(word) === -1) {
+          fileMap.get(file).push(word);
         }
       });
     });
--- a/tests/js/searchtools.js
+++ b/tests/js/searchtools.js
@@ -34,15 +34,12 @@ describe('Basic html theme search', function() {
         },
         docnames:["index"],
         filenames:["index.rst"],
-        indexentries:{},
-        objects:{},
-        objtypes: {},
-        objnames: {},
         terms:{main:0, page:0},
         titles:["Main Page"],
         titleterms:{ main:0, page:0 }
       }
       Search.setIndex(index);
+
       searchterms = ['main', 'page'];
       excluded = [];
       terms = index.terms;
--- a/tests/js/searchtools.js
+++ b/tests/js/searchtools.js
@@ -49,7 +49,7 @@ describe('Basic html theme search', function() {
         'Main Page',
         '',
         null,
-        7,
+        15,
         'index.rst']];
       expect(Search.performTermsSearch(searchterms, excluded, terms, titleterms)).toEqual(hits);
     });
--- a/tests/js/searchtools.js
+++ b/tests/js/searchtools.js
@@ -52,7 +52,7 @@ describe('Basic html theme search', function() {
         'Main Page',
         '',
         null,
-        15,
+        7,
         'index.rst']];
       expect(Search.performTermsSearch(searchterms, excluded, terms, titleterms)).toEqual(hits);
     });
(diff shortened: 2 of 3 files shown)


tests/js/searchtools.js
```diff
@@ -27,6 +27,36 @@ describe('Basic html theme search', function() {
...
+      searchterms = ['main', 'page'];
+      excluded = [];
+      terms = index.terms;
+      titleterms = index.titleterms;
+      hits = [[
+        'index',
+        'Main Page',
+        '',
+        null,
+        7,
```
wlach: Should be changed to 15 once #11958 lands

```suggestion
        15,
```

sphinx/themes/basic/static/searchtools.js
```diff
@@ -496,9 +496,13 @@ const Search = {
...
-        if (fileMap.has(file) && fileMap.get(file).indexOf(word) === -1)
-          fileMap.get(file).push(word);
-        else fileMap.set(file, [word]);
+        if (fileMap.has(file)) {
+          if (fileMap.get(file).indexOf(word) === -1) {
+            fileMap.get(file).push(word);
+          }
+        } else {
+          fileMap.set(file, [word]);
+        }
```
jayaddison: Please check my logic on this; would the following be equivalent?

```
if (!fileMap.has(file)) {
  fileMap.set(file, []);
}
if (fileMap.get(file).indexOf(word) === -1) {
  fileMap.get(file).push(word);
}
```

I think I translate the logic as: ensure that each `file, word` combination encountered is created uniquely in the filemap.

wlach: Yes, I think that should work and would be less confusing!

wlach: Actually, I think this is a little better:

```suggestion
        if (!fileMap.has(file)) {
          fileMap.set(file, [word]);
        } else if (fileMap.get(file).indexOf(word) === -1) {
          fileMap.get(file).push(word);
        }
```

jayaddison: Yep, that seems good to me.  One less comparison in the not-yet-existing-file case, and still fairly straightforward to reason about :+1:

jayaddison:
Two nitpicks with the test case:

- Could you omit the `objtypes` and any other non-essential properties of the test fixture?
- Adding a newline between the setup phase of the test (creating and setting the index) and the assertion/expectation phase (creating the expected `hits` list and running `expect`) could provide a useful hint to readers about how the test is structured.

wlach:
> Two nitpicks with the test case:
> 
>     * Could you omit the `objtypes` and any other non-essential properties of the test fixture?
> 
>     * Adding a newline between the setup phase of the test (creating and setting the index) and the assertion/expectation phase (creating the expected `hits` list and running `expect`) could provide a useful hint to readers about how the test is structured.

Done! For future reference if you have suggestions using GitHub's "suggest" feature in a comment on the diff is a little easier for me to apply -- see step 6 here:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request

jayaddison (approved):
This looks good to me :+1:

jayaddison:
I've been fairly cautious about reviewing these search changes because I'd like to understand them in some detail.  This one still looks good to me.

One minor suggestion that I have -- frustratingly, GitHub doesn't allow suggestions on line ranges that include unmodified lines, it seems -- would be to aim for consistency with the [creation of `scoreMap`](https://github.com/wlach/sphinx/blob/2528b8024b1753127eb75a6331256249adc6a4a5/sphinx/themes/basic/static/searchtools.js#L492-L493) when building `fileMap` here.  That would be something like:

```javascript
        if (!fileMap.has(file)) fileMap.set(file, [word]);
        else if (fileMap.get(file).indexOf(word) === -1) fileMap.get(file).push(word);
```

...however: that's optional, and again I think this pull request is in good shape.

wlach:
> One minor suggestion that I have -- frustratingly, GitHub doesn't allow suggestions on line ranges that include unmodified lines, it seems -- would be to aim for consistency with the [creation of scoreMap](https://github.com/wlach/sphinx/blob/2528b8024b1753127eb75a6331256249adc6a4a5/sphinx/themes/basic/static/searchtools.js#L492-L493) when building fileMap here. That would be something like:
> ...

In JavaScript it's generally considered best practice to always use curly braces in a conditional clause: https://eslint.org/docs/latest/rules/curly -- especially when there's an else involved. 

That said, this would be consistent with the style elsewhere in this file. I guess I'll change it, this PR needs a changelog entry anyway.

jayaddison (approved):
Thanks again @wlach :+1:

tests/js/searchtools.js
```diff
@@ -27,6 +27,33 @@ describe('Basic html theme search', function() {
       expect(Search.performTermsSearch(searchterms, excluded, terms, titleterms)).toEqual(hits);
     });
 
+    it('should be able to search for multiple terms', function() {
+      index = {
+        alltitles: {
```
picnixz, line 32: That's for another PR but the formatting style differs in this file so we should just make it the same (like, there are spaces after colons sometimes and sometimes not) and sometimes there are semicolons and sometimes not.

tests/js/searchtools.js
picnixz: Should we wait instead for that PR to land then? If not, add a TODO note.

tests/js/searchtools.js
wlach: IMO we should run JavaScript through [prettier](https://prettier.io/). Definitely out of scope here though.

tests/js/searchtools.js
wlach: Yeah, my preference would be to land #11958, then rebase this one on top of that (with the above change applied). I'll update the PR description with a note.

wlach:
@jayaddison @picnixz Ok I rebased this on top of master / #11958 so I think it's ready to go in too.

picnixz:
I think it's fine. It'll be the last PR that I'll review for the next 10 days (I'm travelling).

Thank you. If there is some critical things to do next week, ping me or chrisjsewell or jfbu which have write permissions.

wlach:
> I think it's fine. It'll be the last PR that I'll review for the next 10 days (I'm travelling).
> 
> Thank you. If there is some critical things to do next week, ping me or chrisjsewell or jfbu which have write permissions.

No problem, I'll probably just wait for you/others to take a look at the remaining work on their own schedule. If it's been a while I will ping.

Issue #11707, opened 2023-10-05
[follow-up] Generating `<title>` in HTML builds.


picnixz:
This is a follow-up of #11705.

In order to generate the content of the `<title>` element for each document, Sphinx saves instances of `docutils.nodes.title` in the environment and uses them during the writing phase. More precisely, for each document, we either extract the content of ``:title:`` or the title of the first *section*. Then, during the writing phase, we render it by creating a *fake* document. 

In both cases, what is stored is a `docutils.nodes.title` element which will **not** be transformed by Sphinx transformations (or post-transformations). In particular, when the fake document is translated into HTML, it may lack information added by those transformations. If a visitor method expects a node to be transformed by Sphinx transformations or post-transformations, then it would not work for partial documents. 

A simple workaround (that I implemented in #11705) is to detect whether the document being processed is partial or not and apply the visitor if it is not. This however requires to know in advance which transformation is here or not. In particular, some extensions may need a more careful check for that. 

The tasks to address are the following:

* [ ] Check that all partial documents are detected correctly during the writing phase and that the information stored in those partial documents is sufficient. 
* [ ] Change the way that titles are stored in the environment, either by keeping a *real* reference to them (i.e., the node being stored is automatically transformed by the transformation or post-transformation since they are modifying the document in-place) or by applying the transformations and post-transformations to them when they are used in a partial documents. Note that applying transformations and post-transformations to partial documents may be tricky because these are created during the writing phase and not the build phase IIRC.

### Details concerning `<title>` 

I am only aware of using partial documents to render the `<title>` element. Currently, if we have a doctree whose first `<section>` element (detected as the title of the document) contains a node that could be transformed as well and whose transformation adds information needed by the translator (e.g., a `<manpage>` node as of 7.2.6), the title being stored will contain that manpage node. However, during the translation phase, we will first render the `<title>` using a fake document and apply the translator on that specific document. Upon encountering the `<manpage>`, it tries to format `manpages_url` but since the manpage transformation was never applied to that node, the latter lacks the required formatting arguments, leading to #11673.

cc @AA-Turner

PR #11966, opened 2024-02-08
[HTML search] enable discovery of named caption ids

### LINKED ISSUE #11965: [HTML search] named captions have empty targets (anchors) in the titles index
### Describe the bug

As noted in #11961, we have some duplicate search results that appear in the frontend at the moment.

_Some_ of this is due to lack of JavaScript de-duplication of results, but it looks to me like we _also_ have duplication of entries in the `alltitles` part of the search index data that is built by the Python-side HTML/search-index builder code.

**Correction 2024-02-08**: it turns out that these entries are _not_ in fact duplicates.  By coincidence there is a table-of-contents (`toctree`) with a `caption` of `Get started` that emits an empty ID (because it has no `name` option configured).  I think we can improve the situation when a `name` _is_ configured - and #11966 is intended to do that.

### How to Reproduce

After self-building the Sphinx documentation from e976059fd6ab49b1c2784445d7095c9287897724 or a nearby commit, serve the contents of the built documentation via `python -m http.server -b 127.0.0.1` and then run a [query for `get started`](http://127.0.0.1:8000/search.html?q=get+started).

You should see two near-duplicate results at the top of the page; both link to the `index.html` page, one without an anchor, and one with the anchor/fragment `#get-started`.

image

Open the developer tools and inspect the contents of `Search._index.alltitles["Get started"]` in a browser JavaScript console.

image

Observe that the title heading `Get started` has two entries, both for the same file ID of `30`.  This seems to be a bug, and can contribute towards the presence of duplicate titles entries processed by the client.

### Environment Information

```text
Sphinx version:        7.2.6
Docutils version:      0.19
Jinja2 version:        3.1.3
Pygments version:      2.17.2
```

### Sphinx extensions

```python
N/A
```

### Additional context

Relates-to / discovered during investigation of #11961.

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- It seems that under some circumstances, items in the `alltitles` key of the JavaScript search index contain near-duplicates, where some items have an `id` (corresponding to an HTML anchor) and some do not.
- Given that all titles should (I think?) be hyperlinkable references, this changeset prevents title entries without IDs from being included in the search index.

### Detail
- ~~Skip `title` index entries when no IDs are present on the `title` node.~~
- ~~TBD: Work out why an equivalent entry _with_ an ID exists in the doctree.~~
- Allow the ID references for `caption` title nodes to be discovered from the parent of the `compact_paragraph` node that they are contained with.

### Relates
- May resolve #11965.

cc @wlach

### DIFF
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -504,7 +504,8 @@ class IndexBuilder:
             elif isinstance(node, nodes.title):
                 title = node.astext()
                 ids = node.parent['ids']
-                word_store.titles.append((title, ids[0] if ids else None))
+                if ids:
+                    word_store.titles.append((title, ids[0] if ids else None))
                 word_store.title_words.extend(split(title))
             for child in node.children:
                 _visit_nodes(child)
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -503,10 +503,8 @@ class IndexBuilder:
                 word_store.words.extend(split(node.astext()))
             elif isinstance(node, nodes.title):
                 title = node.astext()
-                ids = node.parent['ids']
-                if not ids:  # for example, a caption directive without a name ref
-                    return
-                word_store.titles.append((title, ids[0]))
+                ids = node.parent['ids'] or node.parent.parent['ids']
+                word_store.titles.append((title, ids[0] if ids else None))
                 word_store.title_words.extend(split(title))
             for child in node.children:
                 _visit_nodes(child)
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -504,6 +504,8 @@ class IndexBuilder:
             elif isinstance(node, nodes.title):
                 title = node.astext()
                 ids = node.parent['ids'] or node.parent.parent['ids']
+                if not ids:
+                    return
                 word_store.titles.append((title, ids[0] if ids else None))
                 word_store.title_words.extend(split(title))
             for child in node.children:
--- a/sphinx/search/__init__.py
+++ b/sphinx/search/__init__.py
@@ -504,8 +504,9 @@ class IndexBuilder:
             elif isinstance(node, nodes.title):
                 title = node.astext()
                 ids = node.parent['ids']
-                if ids:
-                    word_store.titles.append((title, ids[0] if ids else None))
+                if not ids:  # for example, a caption directive without a name ref
+                    return
+                word_store.titles.append((title, ids[0]))
                 word_store.title_words.extend(split(title))
             for child in node.children:
                 _visit_nodes(child)
--- a/tests/roots/test-search/index.rst
+++ b/tests/roots/test-search/index.rst
@@ -18,6 +18,8 @@ textinheading
 International
 
 .. toctree::
+   :caption: table-of-contents
+   :name: index-toc
 
    tocitem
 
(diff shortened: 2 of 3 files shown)


jayaddison:
> do not currently contain IDs

(unless the `:name:` option is specified, I think)

picnixz:
Ah ! before doing anything with the search engine, I'm wondering whether #11707 is relevant or not.

jayaddison:
> Ah ! before doing anything with the search engine, I'm wondering whether #11707 is relevant or not.

Thanks @picnixz - I'm not sure, but I don't _think_ that that's relevant in this case.  The problem here seems to be something to do with doctree nodes of type `nodes.title`, distinct from the singleton HTML `<title>`.

I'm narrowing in on the behaviour of the `:caption:` and `:name:` options on the `toctree` directive as the problem here, but I don't yet fully understand what's going on.

jayaddison:
For example: applying the following patch breaks one of the tests, and that seems unexpected.

```diff
diff --git a/tests/roots/test-search/index.rst b/tests/roots/test-search/index.rst
index fc2298b5a..ccaed7c67 100644
--- a/tests/roots/test-search/index.rst
+++ b/tests/roots/test-search/index.rst
@@ -18,6 +18,7 @@ textinheading
 International
 
 .. toctree::
+   :caption: table-of-contents
 
    tocitem
 

```

jayaddison:
Back to first principles: write a test case to illustrate the problem(s).

The problem here is that when the `caption` and `name` options are applied to an object, the resulting index entry has an empty target/ref, because the code that inspects the `nodes.title` object doesn't find the relevant ID  (the ID _does_ get created; it appears in the rendered HTML and can be `:ref:`'d.  so the problem appears to be mainly related to the `IndexBuilder`).

jayaddison:
Figured it out, I think.  `caption` elements are a bit of a special-case; they do emit a `nodes.title` element, but that is nested within a [`compact_paragraph` element](https://github.com/sphinx-doc/sphinx/blob/e976059fd6ab49b1c2784445d7095c9287897724/sphinx/environment/adapters/toctree.py#L161).  So a lookup to find the `ids` won't succeed unless we check one level above in the node tree.

sphinx/search/__init__.py
```diff
@@ -503,7 +503,9 @@ def _visit_nodes(node):
                 word_store.words.extend(split(node.astext()))
             elif isinstance(node, nodes.title):
                 title = node.astext()
-                ids = node.parent['ids']
+                ids = node.parent['ids'] or node.parent.parent['ids']
+                if not ids:
+                    return
```
jayaddison, line 508: Note: there is also a (legacy?) equivalent [code block in `WordCollector`](https://github.com/sphinx-doc/sphinx/blob/e976059fd6ab49b1c2784445d7095c9287897724/sphinx/search/__init__.py#L232-L236).

jayaddison:
Todo:

- [ ] Check (and test cover) whether `code-block` directive captions are also handled correctly.

jayaddison:
> Todo:
> 
>     * [ ]  Check (and test cover) whether `code-block` directive captions are also handled correctly.

They're not - in fact, code snippets with named captions are omitted from the `title` index entirely at the moment.  I think that implies that it would be a separate feature request to add them / make them searchable (I've drafted some code to do that, but it seems worth keeping that separate from the bugfix here).

jayaddison:
The bugreport that this pull request is linked to is confusing and has changed from its' initially-stated problem.  If it seems confusing to me -- I wrote it -- then I don't think it's likely to be understandable for future readers.

I'll close both and re-open a more precisely-defined bugreport after re-assessing the problem.

PR #11766, opened 2023-11-23
Do not use query components in URLs for assets in EPUB rendering

dvzrv:
Subject: Do not use query components in URLs for assets in EPUB rendering

### Feature or Bugfix
- Bugfix

### Purpose
If the EPUB builder is detected, do not add checksums in a query component to the URL of css and js assets.

### Detail
The XHTML in EPUBs does not allow the use of query components in relative URLs.
The checksumming of assets, introduced in
ae206694e68bea074aca633ea0d32e9ed882a95f unconditionally adds a query component for the checksum, which breaks rendering on many enduser devices (tested on e.g. Kindle Paperwhite 3, Kobo Aura H2O Edition 2).

### Relates
Fixes #11598

Issue #11598, opened 2023-08-15
EPUB: Relative URL strings must not have a query component


vimalkvn:
### Describe the bug

With the addition of checksums to static files ( #11415 ), EPUBs created using `make epub`, fail to validate

Here is a sample output from EPUBCheck:

```bash
epubcheck build/epub/SphinxEPUBTest.epub 

Running epubcheck with arguments: build/epub/SphinxEPUBTest.epub

Validating using EPUB version 3.3 rules.
ERROR(RSC-033): build/epub/SphinxEPUBTest.epub/genindex.xhtml(8,85): \
Relative URL strings must not have a query component, \
but found one in "_static/pygments.css?v=8e8a900e".
ERROR(RSC-033): build/epub/SphinxEPUBTest.epub/genindex.xhtml(9,81): \
Relative URL strings must not have a query component, \
but found one in "_static/epub.css?v=0a53256b".
```

### index.rst

```rst
Welcome to Sphinx EPUB Test's documentation!
============================================

.. toctree::
   :maxdepth: 2
   :caption: Contents:

```

### conf.py

```python
project = 'Sphinx EPUB Test'
copyright = '2023, Vimalkvn'
author = 'Vimalkvn'
release = '1'
extensions = []

templates_path = ['_templates']
exclude_patterns = []

html_theme = 'alabaster'
html_static_path = ['_static']
```

### Environment Information

```text
Platform:              linux; (Linux-5.4.0-155-generic-x86_64-with-glibc2.29)
Python version:        3.8.10 (default, May 26 2023, 14:05:08) 
[GCC 9.4.0])
Python implementation: CPython
Sphinx version:        7.1.2
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

picnixz:
Unfortunately, we want to include them in the HTML output. So, should we strip them for EPUB files, possibly renaming the files when writing the output?

@AA-Turner any thoughts on that one?

vimalkvn:
Not sure if I'm missing something, but believe the checksums are not useful even while previewing the docs locally. Still need to use  `sphinx-build -E ...` so CSS changes are picked up. Perhaps, not including checksums for local non-http(s) URLs might cover both these cases.

picnixz:
I am a bit rusty with HTML specs but here are some thoughts:

[...]

PR #11967, opened 2024-02-08
html builder omit query component when builder name start with 'epub'

turky:

### Feature or Bugfix
- Bugfix

### Purpose
PR #11766 removed query component from URLs in EPUB.  But if we want enhance epub builder for their own needs, it doesn't work. This PR aim to solve this.

### Detail
compare builder name start with 'epub'  not fixed literal. we can use our own builder name with prefix 'epub'.

### Relates
#11598


turky:
Thank you for your response. I've thinked about your question, any choice will be embeding negative effects into the code base. withdrawing my PR is the best one.

picnixz:
Oh, you did not need to do that actually. I don't think it would have been a bad idea to have a configuration value. Anyway, we can always reopen this PR if the needs arise !

PR #11972, opened 2024-02-11
Modernize test suite

### LINKED ISSUE #11285: [BUG] Tests have global side-effects.

picnixz:
Closes #11285.

This branch is based on #11940 so that the gettext patch is included and failures on randomly ordered tests are only affected by that order.

I'm only keeping a PR alive so that I can use the CI/CD.


picnixz:
A quick update:

- The PR is still not ready (I didn't bother about lint for now).
- I disabled `shared_result` because it's what's mostly causing the errors (but it seems that I forgot some places or I don't know what happened).
- I enabled random orders + parallel jobs (so that I can have more samples of 'bad' tests). 

Currently, most of the tests are executed in an isolated build, namely isolated source directory. The reason is that some tests incorrectly use `shared_result` and the error does not occur because they are executed before others that may be polluting it. 

The i18n and server-related tests are not parallelized because it's causing too many failures due to busy ports or whatever so I disabled their suite for parallel execution (and it should be reflected in the workflows/Makefile/tox files).

For the currently failing tests, they are likely related to the change of names (I didn't check them yet). Anyway, I'll investigate next week but at least, I have a good view of what is the issue. I think I mostly resolved all modules-related pollution (when we inject attributes or whatever) so it remains to fix the filesystem-related pollution.

picnixz:
So, now I started implementing another approach. The way we are using `shared_result` is incorrect since it assumes test ordering. This forbids us to use `pytest-xdist` so I'm trying to fix all the test suite.

picnixz:
So I implemented multiple things in order to fix this issue. For now, here is a summary of what I did (the commits are a mess btw):

- I documented sphinx.testing. I needed to do this because in the future, there are way too many possibilities of incorrectly using `shared_result` or *srcdir*. I saw that many tests are using them but I don't know whether this was intended or not. For instance, there are tests that use *srcdir* because they are parametrized and that way, the parametrized tests only build once. But inside the test sometimes I see `app.build(force_all=True)` or `freshenv=True`, but this essentially means that the test is meant to be isolated from others. 
- I refactored `sphinx.testing` but kept backwards compatibility. There are some subtle assumptions that the public API is doing but those assumptions are wrong according to how the private behaves. For instance, the `status` and `warning` fixtures assume that `app._status` and `app._warning` are `io.StringIO` objects but that's not necessarily true with how `make_app` exposes the `SphinxTestApp` constructor. 
- I also added some extra fixtures that can help a lot such as defining the default testroot inside a testfile or how tests should be isolated. All of this is documented in `doc/development/testing.rst`
- For now, I only fixed some tests and not everything, so the CI/CD will fail when I'll push my changes in a few minutes. 

I could have done multiple PRs but honestly, everything is due to the fact that either tests are incorrectly using `shared_result` or because nothing is documented. I know that the PR is huge so we can even delay it for 8.x if needed. But I really think that the changes are needed because after everything is fixed, we can enable the `pytest-xdist` (except for the tests with networking or subprocesses since it's more OS-tricky) and the time for the tests should drastically reduce.

picnixz:
So, not everything is entirely sure to be fixed. However, on my machine, using 'xdist' seems to work correctly (and the whole test suite needs 10 seconds, but I have an quite powerful setup). In a sequential mode, the test suite requires around 1 minute but all tests pass (although I'm sure that there are still side-effects that I couldn't catch).

picnixz:
So I think I'm more or less done with this one. I'll do a summary of what I did (and what I needed to do):

- First of all, since I documented `sphinx.testing`, I wanted to use type annotations as much as possible so that PyCharm can easily help in autocompletion for fixtures. However, I observed that we use `follow_imports="skip"` for mypy which implies that some types are not correctly deduced. I think it's a mistake because we should take into account the imports, possibly in `TYPE_CHECKING` guarded blocks. However, I don't why, but the CI/CD does not seem to pick the new configuration (locally, mypy passes without any issue so I don't know...)

- The whole logic of how tests should be written (at least on our side) is a bit changed. In almost *all* tests, there are some `force_all=True` being used but those are actually used because they were just C/C of other tests. From now on, we should be much more careful on when we use that flag.

- I enabled the xdist plugin making tests 4x times faster. However, not all tests are meant to be executed in parallel, especially those that are known to be tricky such as network-related tests. Those tests are automatically marked as 'serial' and are never executed in parallel (see the workflows + conftest.py). If you think a test is not parallel friendly, add them here or use `@pytest.mark.serial()` to mark them as such. I added the `@pytest.mark.parallel()` which *forces* a test to be considered parallel friendly, but use it only if you are sure about it !

- Now, for the new "isolation" logic. One could wonder why I did such complicated thing but here is the rationale:

  * We could achieve perfect isolation by creating a temporary directory for *every* test (see https://github.com/sphinx-doc/sphinx/issues/11285#issuecomment-1499788021). However, this comes at a 3.7x cost. 
  * The slowdown reason is that *parametrized* tests would *also* be using unique files every time ! So, we could actually use `shared_result`, but again, this would require people to *know* that there is not another shared_result in the module somewhere else that is not affecting the build ! 
  * Uniqueness of the sources directory should also depend on the configuration being used. Sometimes, two tests are using different configurations but the same test root. This means that they are not meant to be the same ! and they cannot be executed in parallel either if they use the same sources directory since the build directory is a subdirectory thereof (meaning that the environment would be messed up). 

So, instead of knowing which test to isolate, and without isolating every of them, I implemented 3 isolation modes: "minimal", "grouped" and "always". They work as follows:

* Tests are *always* isolated according to their module and their class where they are defined. While this might slow the tests, it is a good tradeoff between tests without any isolation and tests that can be using the same testroot but located in two different modules (and this is hard to guess). As such, if we have two modules M1 and M2 with tests, each of their tests will have their sources directory in `/tmp/NS-FOR-M1/...` and `/tmp/NS-FOR-M2/...` respectively.

* In addition to this "NS-FOR-##" isolation, tests with the same keyword arguments for the `sphinx()` marker will be using the same sources directory by default. The rationale behind is that most of the time, we only check the output and do not mess the inputs. If we want to mess the inputs, either we should create a really different testroot or we should isolate this specific test. Thus, 

  ```python
	
  @pytest.mark.sphinx('html', testroot='abc', confoverrides={'a': 1})
  def test_1(app): ...

  @pytest.mark.sphinx('html', testroot='abc', confoverrides={'a': 1}) 
  def test_2(app): ...
  ```

will use the same sources directory ``/tmp/NS/CHECKSUM-FOR-CONFIGURATION/SRCDIR``. As you can see, there is the "namespace" part and the "checksum" part which identifies the sphinx parameters. Note that we can achieve the same manually by using `shared_result` but again the ID should well-chosen.

* If the tests have in addition some parametrized values, one should use ``@pytest.mark.isolate('grouped')`` otherwise you'll end up with many many different isolated values (I'll implement tests for testing this logic). 

* If you are not sure about what to do, just use `@pytest.mark.isolate()` and you'll be sure that your test will be isolated.

By the way, the namespace ID is constructed from the module/class *and* the process and thread IDs because I don't know if xdist really isolate everything properly to avoid race conditions. The docs are quite sparse and there are not much explanations for this. 

* Apart from all this, I gave more fixtures to the end-user (and for us) and gave also the possibility to parametrize fixtures a bit better. Currently, we use the *HUGE* "test-root" as the default test root but it's not needed at all ! there are so many files inside that the build time increases everytime we use it. So now, the default test root is "test-minimal" which has almost no files. 

* There were some places where I refactored and modernized the tests because they were using global variables and carried states across tests. Instead, I use the pytest stash which is meant for that (there's also a global cache attached to a pytest configuration object but the stash is more explicit and safe). And I also needed some utils to ease my life so I gave them as well. If you want to keep them private, feel free to argue since they are quite orthogonal to Sphinx itself.

### TL;DR

- Tests can be parallelized and execute 4x faster.
- Testing is now documented.
- If you have questions, just ask.

### Conclusion

I will **NOT** merge this one until I've got feedback. The isolation policy logic *is* complicated but I don't see how I can simplify it without changing the fixtures themselves (and thus breaking whatever public API we were providing). At least, if we want to change the way we isolate tests, the logic can be easily re-implemented without affecting the public API.

sphinx/templates/htmlhelp/project.stp
picnixz, line 1: Mmmh I don't know why I have modified this file.

danieleades:
> * First of all, since I documented `sphinx.testing`, I wanted to use type annotations as much as possible so that PyCharm can easily help in autocompletion for fixtures. However, I observed that we use `follow_imports="skip"` for mypy which implies that some types are not correctly deduced. I think it's a mistake because we should take into account the imports, possibly in `TYPE_CHECKING` guarded blocks. However, I don't why, but the CI/CD does not seem to pick the new configuration (locally, mypy passes without any issue so I don't know...)

the mypy CI job doesn't run on the tests directory *at all* at the moment, it only runs on the `sphinx` directory. I would suggest updating that in a separate PR and then rebasing.
You might find that adding this (and also `follow_imports`) in this PR is biting off a lot...

picnixz:
> the mypy CI job doesn't run on the tests directory at all at the moment, it only runs on the sphinx directory

That I know, but the `sphinx.testing` directory is not part of the test directory. What I don't understand is why sphinx.testing._fixtures does not type-check properly even though I've added the 'follow-import'. Without the follow-import thing everything becomes "Any" and this is *really* a bad idea. 

But yes, I'll probably split this huge PR into multiple ones.

jayaddison:
Thanks, @picnixz - I'd been gradually building up the courage to review this; smaller PRs will help for that :)

picnixz:
Yeah, I had a look at the files and I was like "mmh, ok, maybe it'd be too much for 1 huge PR". The thing is that some changes are so small (like just removing one flag) and since I passed through the tests one by one, it's really hard to make a "clear" commit history. Like, you change one test, and you observe that there are some side-effects but you don't know which one it is from so you try to fix them one by one and at the end a single commit doesn't make sense.

picnixz:
Closing it in favor of #12017, #12018 and #12019

PR #11975, opened 2024-02-12
fix typo in getting-started.rst

smontanaro:
Subject: Caught a simple typo

### Feature or Bugfix
- Bugfix

### Purpose
- fix trivial typo in `getting-started.rst`

### DIFF
--- a/doc/tutorial/getting-started.rst
+++ b/doc/tutorial/getting-started.rst
@@ -40,7 +40,7 @@ line tools available.  You can do a basic verification running this command:
    (.venv) $ sphinx-build --version
    sphinx-build 4.0.2
 
-If you see a similar output, you are on the right path!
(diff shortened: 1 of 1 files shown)


picnixz:
Personally, I prefer the current wording (which I think is gramatically correct though?). As a non-native speaker, could you explain why the current wording is incorrect?

smontanaro:
The change reads more naturally to this native American English speaker. Maybe British English is different? I've been surprised many times by differences which pop up. If you're happy with it, we can just close this.

picnixz:
Since it is gramatically correct, I will keep the current wording (unless someone proves me wrong, and I would be very surprised since my life would have been a lie all along). But thank you for the work! If you spot other typos, feel free to open other PRs.

Issue #10817, opened 2022-09-12
Support syntax def myfunc(arg: "float64[:,:]"):


paugier:
I get a lot of warnings like 
```
`WARNING: could not parse arglist ('arg: float64[:,:]'): Unable to parse Slice object`
```
as soon as I use [Transonic](https://transonic.readthedocs.io) / [Pythran](https://pythran.readthedocs.io) syntax in annotations, i.e. for code like:

```python
Ac = "float64[:,:]"

def myfunc(arg: Ac):
   ...
```

This is a correct syntax for code annotations, so it would be nice to at least have no warning in this case.

Note that the rendered documentation is correct for such case. The problem is really the warnings.

AA-Turner:
A PR is welcome.

A

PR #11981, opened 2024-02-14
PyCode "Unparse": Added support for slice objects in subscriptions

MatrixEditor:
Subject: Fix documentation issue when slice objects are used in subscriptions

### Feature or Bugfix
- Feature

### Purpose
- Added function to handle parsed slice objects

### Detail
- For instance, consider the following class definition:
  ```python
  class Foo: ...

  class Bar(BaseType):
      #: doc comment
      field1: int
      field2: Foo[1:2:3]
      """doc comment"""
  ```
  Internally, sphinx will raise an exception while parsing this file:
  ```
  ...
  File "venv-3.12/lib/python3.12/site-packages/sphinx/pycode/ast.py", line 170, in visit_Subscript
    return f"{self.visit(node.value)}[{self.visit(node.slice)}]"
                                       ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/ast.py", line 407, in visit
    return visitor(node)
           ^^^^^^^^^^^^^
  File "venv-3.12/lib/python3.12/site-packages/sphinx/pycode/ast.py", line 188, in generic_visit
    raise NotImplementedError('Unable to parse %s object' % type(node).__name__)
  NotImplementedError: Unable to parse Slice object
  ```
- *The implemented function does not alter existing code*

### Relates

- #10817

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,8 @@ Features added
 * #11892: Improved performance when resolving cross references in cpp domain.
   Patch by Rouslan Korneychuk.
 
+* #11981: pycode "UnparseVisitor": Added support for slice objects in subscriptions
+
 Bugs fixed
 ----------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,8 @@ Features added
 * #11892: Improved performance when resolving cross references in cpp domain.
   Patch by Rouslan Korneychuk.
 
-* #11981: pycode "UnparseVisitor": Added support for slice objects in subscriptions
+* #11981: Improve rendering of signatures using ``slice`` syntax,
+  e.g., ``def foo(arg: np.float64[:,:]) -> None: ...``.
 
 Bugs fixed
 ----------
--- a/sphinx/pycode/ast.py
+++ b/sphinx/pycode/ast.py
@@ -156,6 +156,20 @@ class _UnparseVisitor(ast.NodeVisitor):
     def visit_Set(self, node: ast.Set) -> str:
         return "{" + ", ".join(self.visit(e) for e in node.elts) + "}"
 
+    def visit_Slice(self, node: ast.Slice) -> str:
+        if not node.lower and not node.upper and not node.step:
+            # Empty slice with default values -> [:]
+            return ":"
+
+        start = self.visit(node.lower) if node.lower else ""
+        stop = self.visit(node.upper) if node.upper else ""
+        if not node.step:
+            # Default step size -> [start:stop]
+            return f"{start}:{stop}"
+
+        step = self.visit(node.step) if node.step else ""
+        return f"{start}:{stop}:{step}"
+
     def visit_Subscript(self, node: ast.Subscript) -> str:
         def is_simple_tuple(value: ast.expr) -> bool:
             return (
@@ -188,16 +202,4 @@ class _UnparseVisitor(ast.NodeVisitor):
     def generic_visit(self, node: ast.AST) -> NoReturn:
         raise NotImplementedError('Unable to parse %s object' % type(node).__name__)
 
-    def visit_Slice(self, node: ast.Slice) -> str:
-        if not node.lower and not node.upper and not node.step:
-            # Empty slice with default values -> [:]
-            return ":"
 
-        start = self.visit(node.lower) if node.lower else ""
-        stop = self.visit(node.upper) if node.upper else ""
-        if not node.step:
-            # Default step size -> [start:stop]
-            return f"{start}:{stop}"
-
-        step = self.visit(node.step) if node.step else ""
-        return f"{start}:{stop}:{step}"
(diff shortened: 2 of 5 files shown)


picnixz:
Can you add tests and a CHANGES entry?

picnixz (changes requested):
Can you add tests and a CHANGES please?

sphinx/pycode/ast.py
```diff
@@ -187,3 +187,9 @@ def visit_Tuple(self, node: ast.Tuple) -> str:
 
     def generic_visit(self, node: ast.AST) -> NoReturn:
         raise NotImplementedError('Unable to parse %s object' % type(node).__name__)
+
+    def visit_Slice(self, node: ast.Slice) -> str:
```
picnixz: Can you put the method above in alphabetical order ? (I think the `visit_*` methods are alphabetically ordered).

tests/test_pycode/test_pycode_ast.py
```diff
@@ -51,6 +51,8 @@
      "lambda x=0, /, y=1, *args, z, **kwargs: ..."),  # posonlyargs
     ("0x1234", "0x1234"),                       # Constant
     ("1_000_000", "1_000_000"),                 # Constant
+    ("Tuple[:,:]", "Tuple[::, ::]"),            # Index, Subscript, Slice
```
picnixz: `[::,::]` feel weird and incorrect to me. For the entirely empty slice, I think `:` should be used instead (i.e., when node.lower, node.upper and node.step are not set).

MatrixEditor: You're right - fixing that in a bit.

tests/test_pycode/test_pycode_ast.py
```diff
@@ -51,6 +51,9 @@
      "lambda x=0, /, y=1, *args, z, **kwargs: ..."),  # posonlyargs
     ("0x1234", "0x1234"),                       # Constant
     ("1_000_000", "1_000_000"),                 # Constant
+    ("Tuple[:,:]", "Tuple[:, :]"),              # Index, Subscript, 2x Slice
```
picnixz, line 54: I'm a bit nitpicky, but do you think it's prefereable to have a space after `,` ? Also, could you perhaps add some tests using numpy extended syntax where you have commas + slices?

MatrixEditor: No problem. Maybe the example isn't the best one, but I still think that space is okay (the space would be added automatically when reformatting a document e.g. using black). *The space is not added by the `visit_Slice`* function.

picnixz: Good for me then !

tests/test_pycode/test_pycode_ast.py
```diff
@@ -51,6 +51,13 @@
...
+    ("Tuple[1:2]", "Tuple[1:2]"),               # Index, Subscript, Slice(no-step)
+    ("Tuple[1:2:3]", "Tuple[1:2:3]"),           # Index, Subscript, Slice
+    ("x[:, np.newaxis, :, :]",
+     "x[:, np.newaxis, :, :]"),                 # Index, Subscript, numpy extended syntax
+    ("y[:, 1:3][np.array([0, 2, 4]), :]",
+     "y[:, 1:3][np.array([0, 2, 4]), :]"),       # Index, 2x Subscript, numpy extended syntax
```
picnixz, line 60: Just by looking at all those identity checks, I think we should have used a sentinel value, say ``IDENTICAL``, to indicate that the output is the same as the input.  This is an easy follow-up PR though.

picnixz (approved):

CHANGES.rst
```diff
@@ -27,6 +27,8 @@ Features added
 * #11892: Improved performance when resolving cross references in cpp domain.
   Patch by Rouslan Korneychuk.
 
+* #11981: pycode "UnparseVisitor": Added support for slice objects in subscriptions
```
```suggestion
* #11981: Improve rendering of signatures using ``slice`` syntax,
  e.g., ``def foo(arg: np.float64[:,:]) -> None: ...``.
```

picnixz (commented):
Actually, can you check that https://github.com/sphinx-doc/sphinx/issues/10817 is indeed solved?

MatrixEditor:
Yes, but that will take me a few moments.

picnixz:
Don't worry, you're fast enough (it also allows me to work on something else in the meantime)

MatrixEditor:
Yes, #10817 is indeed fixed:

temp/\_\_init\_\_.py
```python
Ac = "float64[:, :]"


def myfunc(arg: Ac):
    """Some docs"""
    ...
```

index.rst
```rst
Welcome to temp's documentation!
================================

.. autofunction:: temp.myfunc


Indices and tables
==================

* :ref:`genindex`
* :ref:`modindex`
* :ref:`search`
```

Results in:
image

picnixz:
Is it possible to include this one as a test as well?

picnixz:
Thank you!

PR #11985, opened 2024-02-17
Fix count of default substitutions

DelazJ:
Subject: With the addition of |translation progress| substitution, sphinx has now more than three default substitutions. Fixing the issue.

### Feature or Bugfix
- Bugfix


picnixz:
I like this one.

Issue #10968, opened 2022-11-11
`sphinx-build -M html -c .` sets the config dir to be `./html`


gwerbin:
### Describe the bug

```text
$ sphinx-build -M html -c . ./source ./build -W

Running Sphinx v5.3.0

Configuration error:
config directory doesn't contain a conf.py file (<elided>/my-project/docs/html)
```

This seems like a bug. I was under the impression that `-c` set the configuration directory, and that the source and build directories would be set relative to the config directory. Why is `sphinx-build` looking in the nonexistent `./html` directory for my config file, when I asked it to look in `.`?

### How to Reproduce

```text
$ mv source/conf.py ./conf.py
$ sphinx-build -M html -c . ./source ./build -W
```

### Environment Information

```text
Platform:              darwin; (macOS-12.4-arm64-arm-64bit)
Python version:        3.10.7 (main, Nov  2 2022, 03:06:32) [Clang 13.1.6 (clang-1316.0.21.2.5)])
Python implementation: CPython Sphinx version:        5.3.0
Docutils version:      0.19
Jinja2 version:        3.1.2
```

### Sphinx extensions

N/A

### Additional context

I apologize if this is due to my misunderstanding of how `-c` works. However, if that's the case, the docs are very unclear, and I offer that this is probably still a "docs bug".

gwerbin:
@AA-Turner I just saw that this was closed. Is it for lack of contributions? I can try to clarify the documentation if that's all that's needed.

AA-Turner:
Hi Greg,

That would be really appreciated, thank you! The issue here is that `-M` requires the source and output directories to be passed before any options.

A

jdillard:
Looks like this is the section of the docs that needs to be updated: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-M

jdillard:
The docs were updated in https://github.com/sphinx-doc/sphinx/pull/11992

PR #11992, opened 2024-02-20
Clarify the proper 'make-mode' usage in the docs

jdillard:
Subject: Clarify the proper 'make-mode' usage in the docs.

### Feature or Bugfix
- Bugfix documentation

### Purpose
- The fact that source and output directories need to come first isn't mentioned, which can lead to usage issues.

### Relates
- Fixes the issue mentioned here: https://github.com/sphinx-doc/sphinx/issues/10968#issuecomment-1894998203

### DIFF
--- a/doc/man/sphinx-build.rst
+++ b/doc/man/sphinx-build.rst
@@ -39,8 +39,11 @@ Options
    Extensions can add their own builders.
 
    .. important::
-      Sphinx only recognizes the ``-M`` option if it is used first. The source
-      and output directories must also be passed before any options.
+      Sphinx only recognizes the ``-M`` option if it is used first, along with
+      the source and output directories, before any other options are passed.
+      For example::
+
+          sphinx-build -M html ./source ./build -W --keep-going
 
    The *make-mode* provides the same build functionality as
    a default :ref:`Makefile or Make.bat <makefile_options>`,
--- a/doc/man/sphinx-build.rst
+++ b/doc/man/sphinx-build.rst
@@ -39,7 +39,8 @@ Options
    Extensions can add their own builders.
 
    .. important::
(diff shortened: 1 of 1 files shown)


doc/man/sphinx-build.rst
```diff
@@ -39,7 +39,8 @@ Options
    Extensions can add their own builders.
 
    .. important::
-      Sphinx only recognizes the ``-M`` option if it is used first.
+      Sphinx only recognizes the ``-M`` option if it is used first. The source
+      and output directories must also be passed before any options.
```
kaycebasques: So `-M` must come first, then source dir, then output dir, then all other options? Current phrasing is still not crystal clear IMO

jdillard: Correct. I wasn't sure the best wording either, what about:
```
Sphinx only recognizes the ``-M`` option if it is used first, along with
the source and output directories, before any other options are passed.
```

picnixz: Actually, wouldn't it be better to just include an example of what should be done and what should not ?

kaycebasques: +1 to a working example as a simple and effective way of squashing ambiguity

jdillard: How about something like:

```
Sphinx only recognizes the ``-M`` option if it is used first, along with
the source and output directories, before any other options are passed. For example::

    sphinx-build -M html ./source ./build -W --keep-going
```

kaycebasques: Looks good to me

picnixz: +1

jdillard: Sweet. Updated in 02793bc0b489027d16edd82dc1fabb4ea0b313fd

PR #11996, opened 2024-02-21
Pass autosummary options to recursive template

### LINKED ISSUE #11994: Recursive autosummary does not pass all options
### Describe the bug

When using autosummary with recursion and option `:nosignatures:`, the option is only applied to the top level and not passed on to submodules. This is a problem, because Sphinx extensions can treat the presence of a signature significantly different than just displaying humble parentheses `(…)`.

For example, with `"sphinx_autodoc_typehints"`, I get this (which I want to disable in my summaries):
image

#### Example

Assume a project of the following structure:
```shell
my_module
├── __init__.py
└── submodule.py
```
where `submodule.py` contains a more detailed implementation:
```python
def function_with_complex_signature(
    arg1: str,
    arg2: tuple[float, float],
    arg3: dict,
) -> tuple[list[tuple[float, float]], dict, str]:
    pass
    return [(0, 0)], {}, ""
```

Sphinx with extension autosummary:
```python
extensions = ["sphinx.ext.autosummary"]
autosummary_generate = True
```

and index.rst with `:recursive:` and `:nosignatures:`:
```reStructuredText
.. autosummary::
   :toctree: generated/
   :recursive:
   :nosignatures:

   my_module
```

#### Actual behavior

The html page `my_module.submodule.html` lists the function with signature.
image

#### Expected behavior

- The signatures of functions in nested modules should also be hidden.
- With recursion, options should apply to all submodules, since there is no other way to configure them separately (other than no recursion and listing the nested entries explicitly in the directive).

### Environment Information

```text
Platform:              linux; (Linux-5.15.0-91-generic-x86_64-with-glibc2.35)
Python version:        3.9.17 (main, Jul  5 2023, 20:41:20) 
[GCC 11.2.0])
Sphinx version:        7.2.6
Docutils version:      0.20.1
Jinja2 version:        3.1.2
Pygments version:      2.16.1
```

### Sphinx extensions

```python
["sphinx.ext.autosummary"]
```

### Additional context

_No response_

aeisenbarth:
Subject: Pass autosummary options to recursive template

### Feature or Bugfix
- Bugfix

### Purpose
- When using autosummary with `:recursive:`, options like `:nosignatures:` were not passed, so that nested modules would not have them set.
- This PR makes nested autosummaries inherit options from the root autosummary.

### Detail
- Changed `find_autosummary_in_lines` to parse `:nosignatures:` and add it to `AutosummaryEntry`
- Pass those `AutosummaryEntry` attributes which are actual options to `generate_autosummary_content`. The expression to convert the entry to a dictionary while excluding some is a bit complex. `NamedTuple._asdict` is [actually meant for public use](https://docs.python.org/3/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields) (just underscored to avoid conflicts). An alternative would be to have an options dictionary (including `template` and `recursive`) in AutosummaryEntry.
  I could have merged them into `context`, but decided against because it has a different purpose.
- Changed the template to include the options. Here, I also included `:recursive:` instead explicitly excluding it from the options dictionary. The for loop is maybe a bit too generalized. The idea is that options with boolean or falsy (`""`) values don't need to be printed, True options only include the name and for the rest name and value.
- The Python whole file could benefit from reformatting with e.g. black.

### Relates
- Resolves https://github.com/sphinx-doc/sphinx/issues/11994

### DIFF
--- a/sphinx/ext/autosummary/generate.py
+++ b/sphinx/ext/autosummary/generate.py
@@ -82,6 +82,7 @@ class DummyApplication:
 class AutosummaryEntry(NamedTuple):
     name: str
     path: str | None
+    nosignatures: bool
     template: str
     recursive: bool
 
@@ -260,12 +261,14 @@ def generate_autosummary_content(name: str, obj: Any, parent: Any,
                                  template: AutosummaryRenderer, template_name: str,
                                  imported_members: bool, app: Any,
                                  recursive: bool, context: dict,
+                                 options: dict | None = None,
                                  modname: str | None = None,
                                  qualname: str | None = None) -> str:
     doc = get_documenter(app, obj, parent)
 
     ns: dict[str, Any] = {}
     ns.update(context)
+    ns['options'] = options if options is not None else {}
 
     if doc.objtype == 'module':
         scanner = ModuleScanner(app, obj)
@@ -521,9 +524,14 @@ def generate_autosummary_docs(sources: list[str],
         if app:
             context.update(app.config.autosummary_context)
 
+        options = {
+            option: value
+            for option, value in entry._asdict().items()
+            if option not in ['name', 'path']
+        }
         content = generate_autosummary_content(name, obj, parent, template, entry.template,
                                                imported_members, app, entry.recursive, context,
-                                               modname, qualname)
+                                               options, modname, qualname)
 
         filename = os.path.join(path, filename_map.get(name, name) + suffix)
         if os.path.isfile(filename):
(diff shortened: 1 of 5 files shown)


jayaddison:
@aeisenbarth could you edit the 'related bugs' section to use the phrasing 'Resolves #11994'?  (this is an [GitHub linked-issue indicator](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) that auto-closes the referenced bug if/when a pull request is merged)

Issue #11873, opened 2024-01-12
Provide a better error message if the master document is not included in project documents


AA-Turner:
If `index.rst` does not appear in the pattern of a custom `{include,exclude}_patterns`, the error message is less than helpful. Instead, we should check that the master doc appears in the Project documents, and provide a better error if it doesn’t.

_Inspiration from https://github.com/sphinx-doc/sphinx/issues/11722#issuecomment-1886157588_

Lynn-Hu-D:
Hi! Could anyone give me some idea about what do master doc and Project documents refer to?

Lynn-Hu-D:
If my` config.py` file locates inside the **docs** folder and I have set `include_patterns = ["/docs/**"]` . In this situation, should the `index.rst` be in the include_patterns?

aflip:
I could take this up, but it is unclear to me how I would do ", we should check that the master doc appears in the Project documents, and provide a better error if it doesn’t."

picnixz:
See #12011 for an implementation and some directions I suggested. I wanted to add tests on that PR but the author closed it before I had the time to (I forgot about the PR and since they closed it, I assumed they did not want to work on it anymore).

picnixz:
So, the PR #12011 is now back on track (and thus the task is taken).

popsiclelmlm:
Small triage note: #12011 appears to implement this and was merged on 2024-07-14. It updates the builder/config path for the missing master document cases described here, but it didn't auto-close this issue because it wasn't linked with a closing keyword. Is this ready to close?

picnixz:
Oh thanks for that, and indeed the task is done.

PR #12011, opened 2024-02-26
Provide a better error message if the master document is not included in project documents #11873

Lynn-Hu-D:
Subject: Provide a better error message if the master document is not included in project documents [#11873]

### Feature or Bugfix

- Bugfix
- Refactoring

### Purpose
- Provide a better error message if the master document is not included in project documents

### Detail
Three cases if the master document (index.rst) is not found.
- Case 1: The master document is included in the exclude_patterns
- Case 2: The include_patterns is set, but the master document is not included in the include_patterns
- Case 3: The master document does not exist in the project t document.

### DIFF
--- a/sphinx/builders/__init__.py
+++ b/sphinx/builders/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import codecs
 import pickle
+import re
 import time
 from os import path
 from typing import TYPE_CHECKING, Any, Literal, final
@@ -21,7 +22,7 @@ from sphinx.util.console import bold
 from sphinx.util.display import progress_message, status_iterator
 from sphinx.util.docutils import sphinx_domains
 from sphinx.util.i18n import CatalogInfo, CatalogRepository, docname_to_domain
-from sphinx.util.osutil import SEP, ensuredir, relative_uri, relpath
+from sphinx.util.osutil import SEP, canon_path, ensuredir, relative_uri, relpath
 from sphinx.util.parallel import ParallelTasks, SerialTasks, make_chunks, parallel_available
 
 # side effect: registers roles and directives
@@ -423,9 +424,40 @@ class Builder:
         else:
             self._read_serial(docnames)
 
-        if self.config.root_doc not in self.env.all_docs:
-            raise SphinxError('root file %s not found' %
-                              self.env.doc2path(self.config.root_doc))
+        if self.config.master_doc not in self.env.all_docs:
+            from sphinx.project import EXCLUDE_PATHS
+            from sphinx.util.matching import _translate_pattern
+
+            master_doc_path = self.env.doc2path(self.config.master_doc)
+            master_doc_canon = canon_path(master_doc_path)
+            for pat in EXCLUDE_PATHS:
+                if not re.match(_translate_pattern(pat), master_doc_canon):
+                    continue
+                msg = __('Sphinx is unable to load the master document (%s) '
+                         'because it matches a built-in exclude pattern %r. '
+                         'Please move your master document to a different location.')
+                raise SphinxError(msg % (master_doc_path, pat))
+            for pat in self.config.exclude_patterns:
+                if not re.match(_translate_pattern(pat), master_doc_canon):
+                    continue
+                msg = __('Sphinx is unable to load the master document (%s) '
+                         'because it matches an exclude pattern specified '
+                         'in conf.py, %r. '
+                         'Please remove this pattern from conf.py.')
+                raise SphinxError(msg % (master_doc_path, pat))
+            if set(self.config.include_patterns) != {'**'} and not any(
+                re.match(_translate_pattern(pat), master_doc_canon)
+                for pat in self.config.include_patterns
+            ):
+                msg = __('Sphinx is unable to load the master document (%s) '
+                         'because it is not included in the custom include_patterns = %r. '
+                         'Ensure that a pattern in include_patterns matches the '
+                         'master document.')
+                raise SphinxError(msg % (master_doc_path, self.config.include_patterns))
+            msg = __('Sphinx is unable to load the master document (%s). '
+                     'The master document must be within the source directory '
+                     'or a subdirectory of it.')
+            raise SphinxError(msg % master_doc_path)
 
         for retval in self.events.emit('env-updated', self.env):
             if retval is not None:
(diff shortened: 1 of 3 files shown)


sphinx/builders/__init__.py
```diff
@@ -420,8 +420,18 @@ def read(self) -> list[str]:
             self._read_serial(docnames)
 
         if self.config.root_doc not in self.env.all_docs:
-            raise SphinxError('root file %s not found' %
-                              self.env.doc2path(self.config.root_doc))
+            if ("**" in self.config.exclude_patterns or
```
picnixz: This fails if you have an empty `exclude_patterns` but `include_patterns = ['a.rst', 'b.rst']` only. 

Instead, check whether `self.config.root_doc` is matched by `include_patterns`. Since the include/exclude patterns follow fnmatch-style patterns, you can check whether the root doc is matched (or not) by those patterns using [fnmatch.fnmatch](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch).

A similar logic can be applied for `exclude_patterns` where you check whether `self.config.root_doc` is *not* matched by `exclude_patterns`. You might also want to have a look at [source_suffix](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-source_suffix)

picnixz (changes requested):

picnixz:
Please add some tests (or I'll do it myself this w-e since this might be a bit difficult).

sphinx/builders/__init__.py
Lynn-Hu-D: Thank you for sharing details about `fnmatch.fnmatch `and` source_suffix`. I will focus on refining the code further.

Regarding the issue where the process fails with an empty` exclude_patterns` and `include_patterns = ['a.rst', 'b.rst']`, my understanding is that this failure occurs because the root document `index.rst` is not listed in include_patterns. Could there be a misconception on my part regarding the root document? Is it possible for any file ending with `.rst` to serve as the root document?

picnixz: Yes, it can be configured with [root_doc](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-root_doc)

Lynn-Hu-D: I tested the following
Case A
root file name = index.rst
root_doc = default
`exclude_patterns=[] `and `include_patterns = ['a.rst', 'b.rst']`
output  an error message saying `index.rst` not in `include_patterns`

Case B
root file name = index.rst
root_doc = default
`exclude_patterns=[]` and `include_patterns = ['index.rst', 'b.rst']`
no error, build successfully

Case C
root file name = a.rst
root_doc = default
`exclude_patterns=[]` and `include_patterns = ['a.rst', 'b.rst']`
output  an error message saying `index.rst` not in `include_patterns`

Case D
root file name = a.rst
root_doc = 'a'
`exclude_patterns=[]` and `include_patterns = ['a.rst', 'b.rst']`
no error, build successfully

Are these cases intended behavior, or which one should I modify the error message?

picnixz: Can you also check with the `source_suffix` configuration where you don't use the .rst extension but use .txt extensions?

picnixz: (Ah and sorry for not answering question, but they are correct).

You should also test with some exclusion patterns and inclusion patterns where both contain the index (I don't know which one has the priority)

Lynn-Hu-D: No worries. I got your point that I need to take all souce_suffix (.rst, .txt, .md) into account.
From the documentation, the exclude_patterns has a higher priority.

sphinx/builders/__init__.py
```diff
@@ -420,8 +421,22 @@ def read(self) -> list[str]:
             self._read_serial(docnames)
 
         if self.config.root_doc not in self.env.all_docs:
-            raise SphinxError('root file %s not found' %
-                              self.env.doc2path(self.config.root_doc))
+            root_doc_path = self.env.doc2path(self.config.root_doc)
+            if ("**" in self.config.exclude_patterns or
```
picnixz: I think using `fnmatch` is sufficient. No need to have the `**` check IMO. I'll add tests (in the `tests/` folder) this w-e directly on your branch and will merge it afterwards.

picnixz:
Sorry, I totally forgot about that PR and since I am travelling I won't be able to add the tests. I'll do it in 10 days or so but thank you for your work for now!

picnixz (approved):

Lynn-Hu-D:
> Sorry, I totally forgot about that PR and since I am travelling I won't be able to add the tests. I'll do it in 10 days or so but thank you for your work for now!

Lynn-Hu-D:
Hello,
I didn't intend to close the issue; it was an accident. What steps should I take now to have this PR merged?

picnixz:
>  I didn't intend to close the issue; it was an accident. What steps should I take now to have this PR merged?

You should add a test for that. Do you want some help? (I will be busy until... the end of the week but ping me next week if I forgot about it).

sphinx/builders/__init__.py
```diff
@@ -420,8 +421,20 @@ def read(self) -> list[str]:
             self._read_serial(docnames)
 
         if self.config.root_doc not in self.env.all_docs:
-            raise SphinxError('root file %s not found' %
-                              self.env.doc2path(self.config.root_doc))
+            root_doc_path = self.env.doc2path(self.config.root_doc)
+            if any(fnmatch(root_doc_path, pat) or root_doc_path.endswith(pat)
```
picnixz: I don't think we want `root_doc_path.endswith(pat)`, otherwise if you have `root_doc_path = 'index.rst'` and one of your pattern is simply `rst`, then it doesn't match the pattern (the correct pattern would be `*.rst`). 

Instead, you should check that `root_doc_path != pat` (see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-exclude_patterns for examples).

sphinx/builders/__init__.py
```diff
@@ -420,8 +421,20 @@ def read(self) -> list[str]:
...
+                   for pat in self.config.exclude_patterns):
+                raise SphinxError('customized exclude_patterns is set ' +
+                                  'and root file %s is in the exclude_patterns' %
+                                  root_doc_path)
+            elif not any((fnmatch(root_doc_path, pat) or root_doc_path.endswith(pat))
```
picnixz: Same as above.

picnixz (changes requested):

sphinx/builders/__init__.py
picnixz: I'm sorry I didn't check it before and I approved the PR.

PR #12012, opened 2024-02-26
use 'types-docutils' instead of 'docutils-stubs' for docutils type annotations

danieleades:
closes #12007

I expect this to be a long-running PR which will help identify gaps in the typeshed docutils libraries.

Once all the issues are squashed upstream this can be merged.

PR #12013, opened 2024-02-26
update mypy CI job to use local config

danieleades:
the mypy CI job currently installs its dependencies independently of the `pyproject.toml` file. This PR updates this so that the same dependencies are used for both local dev and in CI

see https://github.com/sphinx-doc/sphinx/pull/12012#issuecomment-1963645656

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,6 +130,7 @@ exclude = [
 ]
 
 [tool.mypy]
+files = "sphinx"
 check_untyped_defs = true
 disallow_incomplete_defs = true
 follow_imports = "skip"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -130,6 +130,7 @@ exclude = [
 ]
 
 [tool.mypy]
+files = "sphinx"
```
picnixz:
```suggestion
files = ["sphinx"]
```

picnixz: (I find it weird having a string when the config value is a plural form)

danieleades:
> Ah so that's also the reason why my mypy checks fail in #11972?

yeah at the moment mypy isn't run on the `tests/` directory at all in CI

picnixz (approved):

PR #12014, opened 2024-02-26
[WIP] run mypy on tests/ directory

danieleades:
mypy is currently not run on the tests directory.

This PR addresses that.

There's a *massive* number of errors- too many to introduce in one PR. Hence I'm whitelisting the existing issues to be incrementally improved.


picnixz:
Actually, I didn't want to be run on tests/ for now. I just wanted to have the --follow-imports=normal so let's do this first I think

PR #12017, opened 2024-02-26
[lint] change ``mypy`` imports policy.

picnixz:
See https://github.com/sphinx-doc/sphinx/pull/11972#issuecomment-1963881026


sphinx/ext/autosummary/generate.py
```diff
@@ -133,7 +133,7 @@ def __init__(self, app: Sphinx) -> None:
 
         if app.translator:
             self.env.add_extension("jinja2.ext.i18n")
-            self.env.install_gettext_translations(app.translator)
+            self.env.install_gettext_translations(app.translator)  # type: ignore[attr-defined]
```
picnixz, line 137: The attribute exists if the extension is loaded.

sphinx/jinja2glue.py
```diff
@@ -17,7 +17,7 @@
 try:
     from jinja2.utils import pass_context
 except ImportError:
-    from jinja2 import contextfunction as pass_context
+    from jinja2 import contextfunction as pass_context  # type: ignore[attr-defined, no-redef]
```
picnixz: That's because of the try-except (and it depends on the jinja version I think)

sphinx/testing/fixtures.py
```diff
@@ -82,7 +82,7 @@ def app_params(request: Any, test_params: dict, shared_result: SharedResult,
     if test_params['shared_result']:
         if 'srcdir' in kwargs:
             msg = 'You can not specify shared_result and srcdir in same time.'
-            raise pytest.Exception(msg)
+            pytest.fail(msg)
```
picnixz, line 85: As you can see, using 'skip' was a bad idea since pytest.Exception never existed in the first place !

sphinx/jinja2glue.py
danieleades: is sphinx actually still supporting multiple jinja2 versions? `pyproject.toml` defines jinja2 >= 3.0, which is the current major version

looking at [the docs](https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.pass_context), i would expect something like the following:

```python
from jinja2 import pass_context
```

does that work?

sphinx/util/i18n.py
```diff
@@ -207,6 +207,7 @@ def format_date(
             # Check if we have to use a different babel formatter then
             # format_datetime, because we only want to format a date
             # or a time.
+            function: Callable[..., str]
```
danieleades: what args does the callable take?

sphinx/util/rst.py
```diff
@@ -54,7 +54,7 @@ def charwidth(char: str, widechars: str) -> int:
 def heading(env: Environment, text: str, level: int = 1) -> str:
     """Create a heading for *level*."""
     assert level <= 3
-    width = textwidth(text, WIDECHARS[env.language])
+    width = textwidth(text, WIDECHARS[env.language])  # type: ignore[attr-defined]
```
danieleades, line 58: is this due to a gap in the type stubs for `Environment`?

sphinx/util/tags.py
```diff
@@ -67,7 +67,7 @@ def eval_condition(self, condition: str) -> bool:
             msg = 'chunk after expression'
             raise ValueError(msg)
 
-        def eval_node(node: Node) -> bool:
+        def eval_node(node: Node | None) -> bool:
```
danieleades, line 70: is this signature right? the code doesn't appear to handle the `None` case

sphinx/ext/autosummary/generate.py
danieleades: where you've assumed that attributes exist that can't be detected by mypy, it might be nice to add comments explaining why you are certain it exists

pyproject.toml
```diff
@@ -133,7 +133,7 @@ exclude = [
 files = ["sphinx"]
 check_untyped_defs = true
 disallow_incomplete_defs = true
-follow_imports = "skip"
+follow_imports = "normal"
```
danieleades: this is the default value, so i think it can just be removed

pyproject.toml
```diff
@@ -133,7 +133,7 @@ exclude = [
...
 ignore_missing_imports = true
 no_implicit_optional = true
```
danieleades: `no_implicit_optional = true` is also the default value now, come to think of it.

pyproject.toml
```diff
@@ -133,7 +133,7 @@ exclude = [
 files = ["sphinx"]
 check_untyped_defs = true
 disallow_incomplete_defs = true
-follow_imports = "skip"
+follow_imports = "normal"
 ignore_missing_imports = true
```
danieleades: do we really want to ignore missing imports here?

danieleades (changes requested):

pyproject.toml
picnixz: I'm not really sure, I haven't tested it but I can check what happens if I remove it.

pyproject.toml
picnixz, line 136:
```suggestion

```

pyproject.toml
picnixz, line 138:
```suggestion
```

sphinx/util/tags.py
picnixz: Wait, I'm wondering why I added it actually.

pyproject.toml
picnixz, line 136:
```suggestion
```

sphinx/ext/autosummary/generate.py
picnixz, line 136:
```suggestion
            # see: https://jinja.palletsprojects.com/en/latest/extensions/#i18n-extension
            env.install_gettext_translations(app.translator)  # type: ignore[attr-defined]
```

sphinx/jinja2glue.py
picnixz: Mmh, if the requirements are jinja2 >= 3.0, then it should be fine.

sphinx/jinja2glue.py
```diff
@@ -194,7 +194,8 @@ def init(
         self.environment.globals['accesskey'] = pass_context(accesskey)
         self.environment.globals['idgen'] = idgen
         if use_i18n:
-            self.environment.install_gettext_translations(builder.app.translator)
+            self.environment.install_gettext_translations(  # type: ignore[attr-defined]
```
picnixz, line 197:
```suggestion 
            # see: https://jinja.palletsprojects.com/en/latest/extensions/#i18n-extension
            self.environment.install_gettext_translations(  # type: ignore[attr-defined]
```

sphinx/util/i18n.py
picnixz: It depends on whether we use `format_date`, `format_time` or `format_datetime`. The `babel_format_date` function itself assumes just a `Callable` because it makes some hacky injection (which I prefer not to change in this PR).

sphinx/util/rst.py
picnixz: This is because of 

https://github.com/sphinx-doc/sphinx/blob/fde43ed8b4367d6f32cbeb7917e76f4be238a190/sphinx/util/template.py#L94-L105

where `language` is an extended attribute. But fair enough, it's better if I double check this one.

picnixz:
Ok, so my comments got somewhat duplicated and placed a bit weirdly. I'll just give up on GH and directly commit using my IDE (wait a bit)

sphinx/jinja2glue.py
picnixz: Yes it should

sphinx/util/i18n.py
picnixz: I'll detail this.

sphinx/util/rst.py
picnixz: It's due to 

https://github.com/sphinx-doc/sphinx/blob/fde43ed8b4367d6f32cbeb7917e76f4be238a190/sphinx/util/template.py#L94-L105

which automatically injects the attribute. I'll change it for something safer.

sphinx/util/tags.py
picnixz: It's because of `node.expr2` which can be None.

(Technically speaking, it should not be as such but mypy complains and I prefer not to suppress that specific error)

pyproject.toml
```diff
@@ -152,6 +149,9 @@ enable_error_code = [
     "ignore-without-code",
     "unused-awaitable",
 ]
+disable_error_code = [
+    "import-untyped",
```
picnixz, line 154: It's to avoid having to install extra type modules for now (not all of them have typing and I think we can live with it for now)

danieleades: the alternative would be to suppress this warning on a per-module basis

sphinx/registry.py
```diff
@@ -11,7 +11,7 @@
 if sys.version_info >= (3, 10):
     from importlib.metadata import entry_points
 else:
-    from importlib_metadata import entry_points
+    from importlib_metadata import entry_points  # type: ignore[import-not-found]
```
danieleades, line 14: this shouldn't be required. `mypy` understands `if sys.version_info ...` calls
if this is throwing an error, maybe `(3, 10)` is the wrong version?

pyproject.toml
picnixz: Mmh, maybe it's better that way actually. Technically, this error code should not overlap with something where an import fail but let's make it explicit

sphinx/search/ja.py
```diff
@@ -13,16 +13,16 @@
...
-    import MeCab
+    import MeCab  # type: ignore[import-not-found]
     native_module = True
 except ImportError:
     native_module = False
 
 try:
-    import janome.tokenizer
+    import janome.tokenizer  # type: ignore[import-not-found]
     janome_module = True
```
danieleades, line 26: the capitalisation of `*_module` variables is inconsistent. Since these are constants they should probably be all-caps

sphinx/theming.py
```diff
@@ -22,7 +22,7 @@
 if sys.version_info >= (3, 10):
     from importlib.metadata import entry_points
 else:
-    from importlib_metadata import entry_points
+    from importlib_metadata import entry_points  # type: ignore[import-not-found]
```
danieleades, line 25: again, this shouldn't be required

sphinx/writers/latex.py
```diff
@@ -29,7 +29,7 @@
     from docutils.utils.roman import toRoman
```
danieleades, line 29: should this be behind an `if sys.version ...` check, rather than catching the `ImportError`?

danieleades (changes requested):

sphinx/theming.py
picnixz: Well, if I didn't put it, mypy complains =/

sphinx/writers/latex.py
picnixz: No because it's OS-dependent and not really a version dependent. And this is something we have no real control over. It's been years that it worked like that so...

sphinx/search/ja.py
picnixz: I don't think this should be part of this PR.

pyproject.toml
picnixz: ~~Ok, installing `types-Pygments` also installs `types-docutils` which in turns makes everything a mess.~~

Ok turning on this one makes the docutils objects type aware and there are 180+ errors because of that. I'll just switch it off for now until the docutils stubs are good.

sphinx/theming.py
danieleades: that makes me think the import guard is wrong

sounds like `if sys.version_info >= (3, 10):` is the wrong python version to guard against

sphinx/writers/latex.py
danieleades: > No because it's OS-dependent and not really a version dependent

ah ok

> It's been years that it worked like that 

and it's definitely still the case?

pyproject.toml
picnixz: Great, now I have more errors even without those packages. I don't know what mypy is doing or what stubs were installed :(

picnixz: Ok, so in a virtual environment it seems it's ok

sphinx/registry.py
picnixz: No, it's the correct one. `importlib.metadata` was experimental even though it was present in the standard library. In addition, entry_points were different:

> The “selectable” entry points were introduced in importlib_metadata 3.6 and Python 3.10. Prior to those changes, entry_points accepted no parameters and always returned a dictionary of entry points, keyed by group. With importlib_metadata 5.0 and Python 3.12, entry_points always returns an EntryPoints object. See [backports.entry_points_selectable](https://pypi.org/project/backports.entry-points-selectable) for compatibility options.


(see "Compatibility note" of https://docs.python.org/3/library/importlib.metadata.html)

picnixz:
We will need some follow-up on that PR:

- By adding 'follow_imports', we would likely want to also add the Pillow and Pygments typing. However, I observed that using them installs something that appears to be docutils typing which seems different from the `docutils-stubs`.

- With `types-Pillow` and `types-Pygments`, there are a *LOT* of typing errors. And the goal is not to update them in this PR but maybe in another one. This PR was only there to make #12018 type aware but I don't to fix everything else.

danieleades:
> We will need some follow-up on that PR:
> 
> * By adding 'follow_imports', we would likely want to also add the Pillow and Pygments typing. However, I observed that using them installs something that appears to be docutils typing which seems different from the `docutils-stubs`.
> * With `types-Pillow` and `types-Pygments`, there are a _LOT_ of typing errors. And the goal is not to update them in this PR but maybe in another one. This PR was only there to make [[test] Re-implementation of the test suite. #12018](https://github.com/sphinx-doc/sphinx/pull/12018) type aware but I don't to fix everything else.

makes sense. Don't want to boil the ocean in this PR.
Is it worth creating an issue to log the future improvements so they don't get forgotten?

picnixz:
>  Is it worth creating an issue to log the future improvements so they don't get forgotten?

Yes

danieleades (approved):
LGTM!

PR #12023, opened 2024-02-28
linkcheck builder: add 'linkcheck_report_timeouts_as_broken' config option

jayaddison:
### Feature or Bugfix
- Feature (adjustment)

### Purpose
- Makes the distinct `linkcheck` status code for `timeout` added in #11876 opt-in instead of always-enabled.
- I've been worrying about potential downstream linkchecker-reporting problems if we enable this silently; this provides a migration path.

### Detail
- Adds a default-enabled `linkcheck_report_timeouts_as_broken` configuration option.

### Relates
- Follow-up to #11876 and #11868.


sphinx/builders/linkcheck.py
```diff
@@ -674,6 +684,7 @@ def setup(app: Sphinx) -> dict[str, Any]:
     app.add_config_value('linkcheck_anchors_ignore_for_url', (), '', (tuple, list))
     app.add_config_value('linkcheck_rate_limit_timeout', 300.0, '')
     app.add_config_value('linkcheck_allow_unauthorized', True, '')
+    app.add_config_value('linkcheck_distinguish_timeouts', False, '')
```
picnixz, line 687:
```suggestion
    app.add_config_value('linkcheck_distinguish_timeouts', False, bool)
```

picnixz, line 687:
```suggestion
    app.add_config_value('linkcheck_distinguish_timeouts', False, '', bool)
```

picnixz:
```suggestion
    app.add_config_value('linkcheck_distinguish_timeouts', False, '', bool)
```

picnixz (commented):
I'm not fan of the configuration name but I don't have a better one

jayaddison:
> I'm not fan of the configuration name but I don't have a better one

Yep, it took me a while to decide, and doesn't seem perfect to me either.  I think `linkcheck_report_timeouts` was one of the other similar-quality candidates (but could kind-of imply that we don't report timed-out hyperlinks as problems when it's disabled..).

jayaddison:
I think that merging this would hide the problem reported in #12038 -- so I think we should try to figure out the problem there beforehand.

jayaddison:
> I think that merging this would hide the problem reported in #12038 -- so I think we should try to figure out the problem there beforehand.

...with one exception to that rule: if we have not resolved that problem by the time that a 7.3.0 release is planned, then we should merge this to retain backwards-compatible behaviour by default.

jayaddison:
No, sorry, I'm talking nonsense.  The affected test doesn't _only_ check the link status, it _also_ checks for a specific SSL-related failure message.  False alarm.

picnixz:
> if we have not resolved that problem by the time that a 7.3.0 release is planned,

Problem before that is that I don't know when 7.3 is planned. I'm not sure I can do it myself since I only have perms on github directly.

picnixz:
> I think that merging this would hide the problem reported in #12038 -- so I think we should try to figure out the problem there beforehand.

Since I have time in the train, here are some alternatives:

* `linkcheck_discriminate_timeout` (defaults False)
* `linkcheck_explicit_timeout` (defaults False)
* `linkcheck_explicit_errors` (defaults False) which could be used in the future if we want to have different report status based on HTTP error codes or something else, instead of gobbling them in the "broken" status. I kinda like this one because it's not just for timeout but also if the host is offline for instance.

picnixz:
Btw, maybe add some doc for that?

CHANGES.rst
```diff
@@ -93,7 +93,8 @@ Bugs fixed
   Set this option to ``False`` to report HTTP 401 (unauthorized) server
   responses as broken.
   Patch by James Addison.
-* #11868: linkcheck: added a distinct ``timeout`` reporting status code.
+* #11868: linkcheck: added a distinct ``timeout`` reporting status code,
+  optionally enabled by setting ``linkcheck_discriminate_timeout`` to ``True``.
```
```suggestion
  optionally enabled by setting :confval:`linkcheck_discriminate_timeout` to ``True``.
```

sphinx/builders/linkcheck.py
```diff
@@ -308,6 +317,7 @@ def __init__(self, config: Config,
         self.retries: int = config.linkcheck_retries
         self.rate_limit_timeout = config.linkcheck_rate_limit_timeout
         self._allow_unauthorized = config.linkcheck_allow_unauthorized
+        self._discriminate_timeouts = config.linkcheck_discriminate_timeout
```
picnixz: It's really a nitpick, but do you want:

- `_discriminate_timeouts` (with s) and `linkcheck_discriminate_timeout` (no s)
- `_discriminate_timeouts` and `linkcheck_discriminate_timeouts` (both s)
- `_discriminate_timeout` and `linkcheck_discriminate_timeout` (no s)

I'm in favor of the last one (less to type, still explanatory and timeout is mostly singular in the other configuration values).

picnixz (commented):
Small nitpick.

CHANGES.rst
jayaddison: Ah, thank you.  I should fix `linkcheck_allow_unauthorized` in `CHANGES.rst` too.

CHANGES.rst
```diff
@@ -94,7 +94,8 @@ Bugs fixed
   responses as broken.
   Patch by James Addison.
 * #11868: linkcheck: added a distinct ``timeout`` reporting status code,
```
picnixz: You need a period here.

sphinx/builders/linkcheck.py
jayaddison: Hmm.  In context when the code is making an individual HTTP request, `timeout` - for example as it relates to `rate_limit_timeout` is appropriate.

But in the context of the behaviour of the `linkcheck` builder for an entire session, the configuration relates to handling of each and every timeout encountered -- not only the first timeout, and not the abstract notion of timeout as "a duration before software stops waiting" -- because these are actual timeouts that occurred contacting network hosts.

So, long story short: `config.linkcheck_discriminate_timeout` should become `config.linkcheck_discriminate_timeouts`, and I'll apply that change.  Thank you.

jayaddison: (and my apologies for the extra typing introduced by that)

AA-Turner:
I've renamed the option to ``linkcheck_report_timeouts_as_broken``, which is clearer, though a logical inversion. The question is if this is removed in Sphinx 8, or simply has the default changed to `False`.

A

AA-Turner (approved):

PR #12024, opened 2024-02-28
Fix typos in `usage/configuration.rst`

afranke:
I’m guessing this section was copied over from the previous one and these bits got forgotten when updating.

I’m also wondering whether there is a missing “at an” before “earlier or lazier step”. I can amend this PR if that’s confirmed.


picnixz:
I think "at an" is indeed needed. Could you also fix this in the CSS section please?

picnixz:
Anyway I think everything is fine. Thank you for your work !

PR #12029, opened 2024-03-01
[HTML search] tests: bugfix: correction for test index fixture format

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Make the `tests/js/searchtools.js` test file more representative of the search index format used by the Sphinx client-side JS code.

### Detail
- Ranking of search results relies on scoring, and Sphinx exact-match queries on terms (arbitrary words) are intended to score `5`, while partial-matches score `2` per term ([ref](https://github.com/sphinx-doc/sphinx/blob/9a30ca7da1aeffe0c21b81fc7722eb1aed1512da/sphinx/themes/basic/static/searchtools.js#L44-L46)).
- Multi-word searches such as `datetime timedelta` are supported by Sphinx; internally the JavaScript on the page tokenizes these into multi-term queries (`['datetim', 'timedelta']`).
- The search index itself only contains _single-word_ terms.
- Each term in the index is mapped to a list of documents where the term is found, and these document-ids are zero-indexed (`0, 1, 2, ...`).
- :bug: The `tests/js/searchtools.js` file has an incorrect representation of the `terms` entry in the index, instead mapping a search term to a single document-id.

### Relates
- Resolves #12028.

PR #12031, opened 2024-03-01
Tests: expand search coverage for searchtools.js

jayaddison:
### Feature or Bugfix
- Feature (tests)

### Purpose
- Improve the test coverage of the `searchtools.js` search index/querying functionality, to aid maintenance and development.

### Detail
- [ ] Object-name matching.
  - [ ] Fully-qualified path.
  - [ ] Local-lexical-scope names.
  - [ ] Priorities.
- [ ] Document title matching.
- [x] Exact-matching.
- [x] Multi-word document term matching.
- [ ] Partial document term matching.
- [ ] Single-document and multi-document terms in the index (when a term only exists within a single document, it appears as `{term: 1}` in the index, instead of within an `Array` such as `{term: [0, 1, 2, ...]}` in the multi-document case.

### Relates
- Follows-up / inspired-by #12029.


jayaddison:
Todo: revisit this after other search-related pull requests are merged.  Each of them tends to include individual test coverage.

PR #12034, opened 2024-03-02
improve type safety when using NodeMatcher

danieleades:
Improve type safety by refactoring the `NodeMatcher`.

Mypy is unable to infer the return type of the `Node.find_all` method when `NodeMatcher` is used. By moving the helper method into the `NodeMatcher` class and adding some generics it's possible for type checkers to correctly infer the return type of the iterator.

I've noticed that a matcher rarely gets reused, so an alternative approach would be to get rid of the helper class entirely and turn it into a plain function.

### DIFF
--- a/sphinx/builders/html/transforms.py
+++ b/sphinx/builders/html/transforms.py
@@ -47,7 +47,7 @@ class KeyboardTransform(SphinxPostTransform):
         matcher = NodeMatcher(nodes.literal, classes=["kbd"])
         # this list must be pre-created as during iteration new nodes
         # are added which match the condition in the NodeMatcher.
-        for node in list(self.document.findall(matcher)):  # type: nodes.literal
+        for node in list(matcher.findall_in(self.document)):
             parts = self.pattern.split(node[-1].astext())
             if len(parts) == 1 or self.is_multiwords_key(parts):
                 continue
--- a/sphinx/builders/html/transforms.py
+++ b/sphinx/builders/html/transforms.py
@@ -47,7 +47,7 @@ class KeyboardTransform(SphinxPostTransform):
         matcher = NodeMatcher(nodes.literal, classes=["kbd"])
         # this list must be pre-created as during iteration new nodes
         # are added which match the condition in the NodeMatcher.
-        for node in list(matcher.findall_in(self.document)):
+        for node in list(matcher.findall(self.document)):
             parts = self.pattern.split(node[-1].astext())
             if len(parts) == 1 or self.is_multiwords_key(parts):
                 continue
--- a/sphinx/util/nodes.py
+++ b/sphinx/util/nodes.py
@@ -62,7 +62,7 @@ class NodeMatcher(Generic[N]):
         self.classes = node_classes
         self.attrs = attrs
 
-    def _match(self, node: Node) -> bool:
+    def match(self, node: Node) -> bool:
         try:
             if self.classes and not isinstance(node, self.classes):
                 return False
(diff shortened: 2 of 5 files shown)


sphinx/builders/html/transforms.py
```diff
@@ -47,7 +47,7 @@ def run(self, **kwargs: Any) -> None:
         matcher = NodeMatcher(nodes.literal, classes=["kbd"])
         # this list must be pre-created as during iteration new nodes
         # are added which match the condition in the NodeMatcher.
-        for node in list(self.document.findall(matcher)):  # type: nodes.literal
+        for node in list(matcher.find_all_in(self.document)):
```
danieleades: i'm baffled as to why this needs to be wrapped in `list(...)` but it seems to break without it

???

sphinx/util/nodes.py
```diff
@@ -83,7 +85,15 @@ def match(self, node: Node) -> bool:
             return False
 
     def __call__(self, node: Node) -> bool:
-        return self.match(node)
+        return self._match(node)
+
+    def findall_in(self, node: Node) -> Iterator[T]:
```
picnixz: I prefer having something like

```
for node in matcher.findall(...)
```

rather than 

```
for node in matcher.findall_in(...)
```

Main reason is that it's somewhat similar to `re` where you have

```
pattern = re.compile(...)
m = pattern.findall(...)
```

sphinx/util/nodes.py
```diff
@@ -43,24 +47,22 @@ class NodeMatcher:
...
         # => [<reference ...>, <reference ...>, ...]
     """
 
-    def __init__(self, *node_classes: type[Node], **attrs: Any) -> None:
+    def __init__(self, *node_classes: type[T], **attrs: Any) -> None:
         self.classes = node_classes
         self.attrs = attrs
 
-    def match(self, node: Node) -> bool:
+    def _match(self, node: Node) -> bool:
```
picnixz: I think we should keep it public since people might rely on this class. The class is not really private IMO and extensions could potentially use it.

sphinx/util/nodes.py
```diff
@@ -33,7 +34,10 @@
 caption_ref_re = explicit_title_re  # b/w compat alias
 
 
-class NodeMatcher:
+T = TypeVar("T", bound=Node)
```
picnixz: Maybe use `N` or `NT` for "NodeType" instead of `T`?

jayaddison:
Are any of the affected callsites recursive methods?

What I'm thinking: an additional frame/entry on the stack (for the intermediate method call) is usually likely to have negligible effect for a call-and-return, but within a deeply recursive method it could have a significant effect on memory usage.

sphinx/util/nodes.py
danieleades: done

sphinx/util/nodes.py
danieleades: done

sphinx/util/nodes.py
danieleades: done

picnixz (commented):
Seems fine to me but can you confirm that `self.document` is of correct type actually ? (what is the revealed type?) because I remember that mypy was not happy sometimes because of the fact that `self.document` was not typed at all (and did not exist on `SphinxTransform`).

danieleades:
> Seems fine to me but can you confirm that `self.document` is of correct type actually ? (what is the revealed type?) because I remember that mypy was not happy sometimes because of the fact that `self.document` was not typed at all (and did not exist on `SphinxTransform`).

i'm making some changes to the type stubs upstream in typeshed that should help with this

picnixz:
Thank you!

jayaddison:
> Are any of the affected callsites recursive methods?

From a relatively quick check of the code: no, none of the affected methods are recursive.

PR #12035, opened 2024-03-02
Don't use a dash if there's no description to introduce

nedbat:
Subject: In parameter lists, an en-dash is added even if there's no description following it.

### Feature or Bugfix

- Bugfix

### Purpose
The coverage.py docs currently have this:
image

My docstrings aren't structured in a way to extract parameters, so there are no descriptions for them.  But there's a dash leading to nothing.

With this fix, the same parameter list looks like this, which makes more sense:
image


sphinx/util/docfields.py
```diff
@@ -236,7 +236,7 @@ def make_field(
         inliner: Inliner | None = None,
         location: Element | None = None,
     ) -> nodes.field:
-        def handle_item(fieldarg: str, content: str) -> nodes.paragraph:
```
picnixz, line 239: Is the typing really a list of node or just a single Node? (note that in the case of a single node, content[0] is the first node's child).

sphinx/util/docfields.py
```diff
@@ -254,8 +254,12 @@ def handle_item(fieldarg: str, content: str) -> nodes.paragraph:
                 else:
                     par += fieldtype
                 par += nodes.Text(')')
-            par += nodes.Text(' -- ')
-            par += content
+            empty = len(content) == 0
```
picnixz: If "content" is not a string nor a list of nodes but a real `Node`, you can write:

```
if content.astext().strip():
    par += nodes.Text(' -- ')
    par += content
```

If content is a list of string or a list of node, I don't think it's right to just check the first node element being added since the next nodes in the list might be non-empty text.

sphinx/util/docfields.py
nedbat: The data I was seeing said it was a `list[Node]`.  It's definitely not `str`.  Other type hints indicate it's `list[Node]`, though when I tried to update more of them to be precise, mypy gave me paradoxical errors.

picnixz: > The data I was seeing said it was a list[Node].

Do you mean the runtime type? or the statically-deduced type? (it would be good to know what happens actually)

nedbat: I mean the type as reported by `print(f"{content = }")`, and similar runtime debugging.  It's a `list[Node]`.

picnixz: Ok, then I think you should rather check that the concatenation of all `astext()` of all those nodes is empty and not just the first one (I don't know from where this list is actually constructed).

nedbat: I am checking if the list has only one element, and then checking the contents of the only node.  You are right that a more general kind of emptiness could be a longer list that has no text, but this handles the case of literally no description.

picnixz: Actually, I am worried about the fact that extensions might pass more than one item (maybe without noticing it). But I think we can ignore this case maybe?

nedbat: OK, I can make it more general.

picnixz: Thank you! (actually, I don't know why this is a list of nodes and not a single node).

picnixz (approved):
Thank you. I'll do a little check tomorrow so I'll only merge it after that but consider your PR approved.

picnixz:
Thank you! I don't think a changelog entry is needed, but feel free to create one in a separate PR if you want!

PR #12037, opened 2024-03-02
[HTML search] correction for scoring of search terms that only appear in document indexed with id zero.

### LINKED ISSUE #11957: HTML Search: Partial matches can overwrite main matches in certain rare cases

### LINKED ISSUE #12045: [HTML search] optimization: don't loop over all document terms and title terms during partial-matching.
There seems to be a potentially-large inefficiency in the Sphinx JavaScript search code: for any query word greater than length two (likely to be >90% of them, I'd guess!), we iterate through _all_ document terms and all title terms in the client's search index to check for partial matches.

~~However, if an exact-match was already found on the query word, then we won't add any of those candidates, even when they do match.  That means that we spend JavaScript compute resources iterating through items that are unused.~~ Since Sphinx 7.3.0, this is no longer true thanks to #11958 -- if an _exact-match_ on a stemmed term is found, we skip checks for partial matches for that query term.

The relevant code is found here: https://github.com/sphinx-doc/sphinx/blob/cf7d2759af0852d67288e58d823d51fe860749ca/sphinx/themes/basic/static/searchtools.js#L469-L476

~~I don't have any stats on the performance impact of this, but it feels like it may be significant, especially for large documentation projects.~~ [Initial results](https://github.com/sphinx-doc/sphinx/issues/12045#issuecomment-2228901785) appear to indicate a **9ms** vs **48ms** difference on a local test machine.

- ~~If we've already found exact-matches for a document term, then do not iterate over all document terms to check for partial matches.~~
- ~~If we've already found exact-matches for a document title term, then do not iterate over all document title terms to check for partial matches.~~
- Find a non-brute-force algorithm for partial substring matching on terms and titles.

**Additional context**
- Extracted from #11957.

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Restores expected scoring for exact-match query terms that only appear in document zero in the client-side JavaScript search index.
- Includes a performance optimization: skips partial-matching entirely when exact-matches exist.

### Detail
- Words (terms) from both document content and document titles are stored in the JavaScript search index on the client -- this results in a mapping from individual terms to a list of the document-ids of the documents where they appear.
- For terms/title-terms that appear in multiple documents, the document-ids are stored as a [(sorted) list/`Array`](https://github.com/sphinx-doc/sphinx/blob/bea8b6bf4db0992a987d3645029b990610f5212b/sphinx/search/__init__.py#L380).  In cases where the term only appears in a single document, only a [single integer value is stored](https://github.com/sphinx-doc/sphinx/blob/bea8b6bf4db0992a987d3645029b990610f5212b/sphinx/search/__init__.py#L375-L378).  (credit to [@wlach for explaining](https://github.com/sphinx-doc/sphinx/pull/11958#issuecomment-1974071072))
- The document-ids are zero-indexed (`0, 1, 2, ...`) so there is a zeroth document, and we need to be careful when evaluating that in boolean expressions because zero is falsy in JavaScript.
- :bug: A term that only appears in the document with ID zero will incorrectly be scored as a partial-match.

### Relates
- Resolves #11957.
- Resolves #12045.
- Alternative approach to #11958.

cc @wlach

### DIFF
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -483,7 +483,7 @@ const Search = {
         });
         Object.keys(titleTerms).forEach((term) => {
           if (term.match(escapedWord) && titleTerms[word] === undefined)
-            arr.push({ files: titleTerms[term], score: Scorer.partialTitle });
+            arr.push({ files: titleTerms[word], score: Scorer.partialTitle });
         });
       }
 
--- a/sphinx/themes/basic/static/searchtools.js
+++ b/sphinx/themes/basic/static/searchtools.js
@@ -478,12 +478,12 @@ const Search = {
       if (word.length > 2) {
         const escapedWord = _escapeRegExp(word);
         Object.keys(terms).forEach((term) => {
-          if (term.match(escapedWord) && !terms[word])
+          if (term.match(escapedWord) && terms[word] === undefined)
             arr.push({ files: terms[term], score: Scorer.partialTerm });
         });
         Object.keys(titleTerms).forEach((term) => {
-          if (term.match(escapedWord) && !titleTerms[word])
-            arr.push({ files: titleTerms[word], score: Scorer.partialTitle });
+          if (term.match(escapedWord) && titleTerms[word] === undefined)
+            arr.push({ files: titleTerms[term], score: Scorer.partialTitle });
         });
       }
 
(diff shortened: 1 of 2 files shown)


jayaddison:
Replaced by / became-identical-to #11957.

PR #12039, opened 2024-03-02
github template config: fixup: use updated issue label names

jayaddison:
### Feature or Bugfix
- Bugfix (GitHub issue template config)

### Purpose
- Pre-populate labels on feature requests and bugs as expected.

### Detail
- We use `type:bug` and `type:enhancement` instead of the default `bug` and `enhancement` labels for our GitHub issues.

### Relates
- N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,7 @@
 name: Feature request
 about: Suggest an idea for this project
 title: '<short description for the feature>'
-labels: 'type:enhancement'
+labels: 'type:proposal'
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,7 @@
 name: Feature request
 about: Suggest an idea for this project
 title: '<short description for the feature>'
-labels: 'enhancement'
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -2,7 +2,7 @@
 name: Feature request
 about: Suggest an idea for this project
 title: '<short description for the feature>'
-labels: 'enhancement'
+labels: 'type:enhancement'
```
picnixz: I'm conflicted between the enhancement and the proposal actually.

jayaddison: Good point.  Every open issue with label `type:enhancement` currently _also_ has `type:proposal` (enhancement implies proposal).  The reverse is not true (some open proposals are enhancements, but not all).

picnixz: As such I think we should use proposal instead and we will change to enhancement if we agree about it.

picnixz:
```suggestion
labels: 'type:proposal'
```

picnixz:
Thank you!

.github/ISSUE_TEMPLATE/feature_request.md
jayaddison: Agreed - thank you!

PR #12042, opened 2024-03-03
fix annotation for directives.other.Include.run

danieleades:
the return type here should be `Sequence` rather than `list` since it uses the super class method which is typed as `Sequence`


danieleades:
the mypy error here is wrong, but it's wrong because the docutils-stubs type is wrong. Meaning it can't be fixed until after migrating to the typeshed stubs.

setting as draft

picnixz:
Could we just ignore this one with an override instead?

danieleades:
> Could we just ignore this one with an override instead? Or do you think it will be fixed upstream fast enough?

hopefully fixed upstream soon enough!

AA-Turner:
Covered in #12012

PR #12043, opened 2024-03-03
Fix a warning in #12036

picnixz:
Internal hotfix.


picnixz:
Good to know that the auto-squash surprised me (I expected the tests to be required as well)

PR #12050, opened 2024-03-07
tests: implement readiness check before yielding local-http(s) test servers

### LINKED ISSUE #12038: flaky unit test: test_build_linkcheck.test_connect_to_selfsigned_fails
### Describe the bug

The [`test_connect_to_selfsigned_fails` test case in the `test_build_linkcheck` module](https://github.com/sphinx-doc/sphinx/blob/9a30ca7da1aeffe0c21b81fc7722eb1aed1512da/tests/test_builders/test_build_linkcheck.py#L628-L639) appears to fail occasionally on Windows:

```
================================== FAILURES ===================================

app = <SphinxTestApp buildername='linkcheck'>

    @pytest.mark.sphinx('linkcheck', testroot='linkcheck-localserver-https', freshenv=True)
    def test_connect_to_selfsigned_fails(app):
        with https_server(OKHandler):
            app.build()
    
        with open(app.outdir / 'output.json', encoding='utf-8') as fp:
            content = json.load(fp)
        assert content["status"] == "broken"
        assert content["filename"] == "index.rst"
        assert content["lineno"] == 1
        assert content["uri"] == "https://localhost:7777/"
>       assert "[SSL: CERTIFICATE_VERIFY_FAILED]" in content["info"]
E       assert '[SSL: CERTIFICATE_VERIFY_FAILED]' in "HTTPSConnectionPool(host='localhost', port=7777): Read timed out. (read timeout=0.05)"

tests\test_builders\test_build_linkcheck.py:639: AssertionError
```

### How to Reproduce

See https://github.com/sphinx-doc/sphinx/actions/runs/8080885896/job/22078226984 for an example failure during a pull request -- seemingly without any failure-related changes -- from common base commit 9a30ca7da1aeffe0c21b81fc7722eb1aed1512da of the `master` branch.

### Environment Information

```text
N/A
```

### Sphinx extensions

```python
N/A
```

### Additional context

A [timeout of 0.05s](https://github.com/sphinx-doc/sphinx/blob/9a30ca7da1aeffe0c21b81fc7722eb1aed1512da/tests/roots/test-linkcheck-localserver-https/conf.py#L2) seems plenty to make a connection to a local HTTPS server that uses a self-signed certificate.  However from prior experience of timeouts like this in similar `linkcheck` tests, there can be some amount of interleaving and conflict between the individual test cases.

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Intended to prevent a flaky test failure from causing false-positive failure notices in Sphinx's continuous integration tests on GitHub Actions.

### Detail
- Implements a [`socket`-based](https://docs.python.org/3/library/socket.html#socket.create_connection) readiness check on each test HTTP(S) server created for use during unit tests before `yield`ing it for use by the corresponding test case.

### Relates
- Resolves #12038.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,6 @@ jobs:
           docutils: "0.19"
 
     steps:
-    if: false
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python }}
       uses: actions/setup-python@v5
@@ -75,38 +74,20 @@ jobs:
 
   windows:
     runs-on: windows-2019
-    name: Windows ${{ matrix.python }} (Docutils ${{ matrix.docutils }})
-    strategy:
-      fail-fast: false
-      matrix:
-        python:
-        - "3.9"
-        - "3.10"
-        - "3.11"
-        - "3.12"
-        - "3.13-dev"
-        docutils:
-        - "0.18"
-        - "0.20"
-        include:
-        # test every supported Docutils version for the latest supported Python
-        - python: "3.12"
-          docutils: "0.19"
+    name: Windows
 
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python ${{ matrix.python }}
+    - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: ${{ matrix.python }}
+        python-version: "3"
     - name: Check Python version
       run: python --version
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         python -m pip install .[test]
-    - name: Install Docutils ${{ matrix.docutils }}
-      run: python -m pip install --upgrade "docutils~=${{ matrix.docutils }}.0"
     - name: Test with pytest
       run: python -m pytest -vv --durations 25
       env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,6 +46,7 @@ jobs:
           docutils: "0.19"
 
     steps:
+    if: false
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python }}
       uses: actions/setup-python@v5
@@ -74,20 +75,38 @@ jobs:
 
   windows:
     runs-on: windows-2019
-    name: Windows
+    name: Windows ${{ matrix.python }} (Docutils ${{ matrix.docutils }})
+    strategy:
+      fail-fast: false
+      matrix:
+        python:
+        - "3.9"
+        - "3.10"
+        - "3.11"
+        - "3.12"
+        - "3.13-dev"
+        docutils:
+        - "0.18"
+        - "0.20"
+        include:
+        # test every supported Docutils version for the latest supported Python
+        - python: "3.12"
+          docutils: "0.19"
 
     steps:
     - uses: actions/checkout@v4
-    - name: Set up Python
+    - name: Set up Python ${{ matrix.python }}
       uses: actions/setup-python@v5
       with:
-        python-version: "3"
+        python-version: ${{ matrix.python }}
     - name: Check Python version
       run: python --version
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         python -m pip install .[test]
+    - name: Install Docutils ${{ matrix.docutils }}
+      run: python -m pip install --upgrade "docutils~=${{ matrix.docutils }}.0"
     - name: Test with pytest
       run: python -m pytest -vv --durations 25
       env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,7 @@ env:
 
 jobs:
   ubuntu:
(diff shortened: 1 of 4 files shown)


jayaddison:
I'm going to push a few more commits for this momentarily, including a changelog entry.  But this code also needs to be evaluated on a large-batch of Windows CI unit test builds, to check that it's reliable.

.github/workflows/main.yml
```diff
@@ -26,6 +26,7 @@ env:
 
 jobs:
   ubuntu:
+    if: false
```
picnixz: is it to force Windows jobs?

tests/utils.py
```diff
@@ -15,11 +16,15 @@
 # File lock for tests
 LOCK_PATH = str(TESTS_ROOT / 'test-server.lock')
 
+HOST_NAME = "localhost"
+HOST_PORT = 7777
+HOST = (HOST_NAME, HOST_PORT)
```
picnixz:
```suggestion
ADDRESS = (HOST_NAME, HOST_PORT)
```

tests/utils.py
```diff
@@ -45,6 +50,8 @@ def server(handler):
             server_thread = thread_class(handler, daemon=True)
             server_thread.start()
             try:
+                # await test server connectivity before yielding a result
+                socket.create_connection(HOST, timeout=0.5).close()
```
picnixz: Use a context manager instead (at least it's clearer IMO)

.github/workflows/main.yml
jayaddison: The CI changes are temporary, to try to prove whether the flakiness is resolved on Windows; since that doesn't require running the `ubuntu` tests, they're disabled here (to reduce resource usage).

picnixz:
If you want to check whether it's reliable or not and you don't have access to Windows, just commit a space diff, then commit without the space and again just to trigger the CI/CD

tests/utils.py
jayaddison: I've pushed a change with that, although am not 100% sure whether it's an improvement in this case (in most cases I'd agree that using a context manager for socket/file/other closable resources makes sense).

picnixz: Yeah after looking at it.. both have their disadvantages, like the first version is a bit tricky in the sense that you see "create" then "close" afterwards but the alternative has a `pass` ... which is not better actually.

I'll just revise my opinion and let you keep whichever version you feel is better for you.

jayaddison: Thanks, yep, neither is perfect.  I'll revert it to gain some more continuous integration confidence.. but might revert it back again (or think of other ideas... ?).

jayaddison: One thing that worries me about the `pass` pattern is that someone might later think "hey, this is a bug in the test - the `yield` should be inside the context-manager" -- because that's a typical pattern -- and if they make that change, tests would continue to succeed _most_ of the time (as they do before this branch), until the test flakiness reappeared again in future.

I can't think how to guard against that other than comments though, really.

picnixz: Your argument is sound, let's keep what you did originally (sorry for bothering you)

tests/utils.py
```diff
@@ -45,7 +50,8 @@ def server(handler):
             server_thread = thread_class(handler, daemon=True)
             server_thread.start()
             try:
-                yield server_thread
+                socket.create_connection(ADDRESS, timeout=0.5).close()  # attempt connection.
```
picnixz: If you want to end your sentences with periods you should also start with a capital letter (I think it's fine not to have a period for non-capitalized small comments in general)

tests/utils.py
jayaddison: No problem.  I'm sorry for all the pings when you've been busy.  I think this is ready now, and I'll remove the 'do not merge flag' in a few moments.

tests/utils.py
jayaddison: Ok, that's reasonable :)

The full-stop was partly to reduce the chance that one or other of the lines could be deleted, leaving the other one parse as valid but without the relevant context.

picnixz:
Is it fine to merge? I don't have anything else to say

jayaddison:
> Is it fine to merge? I don't have anything else to say

I count 50 Python test builds on Windows without seeing the failure occur again - that gives me reasonable (although not complete) confidence that it's OK to merge.  Even so, I'll monitor the test results on mainline and PRs for the next few weeks.

picnixz:
Thank you!

jayaddison:
And thank you for the review and merge!

jayaddison:
> I count 50 Python test builds on Windows without seeing the failure occur again - that gives me reasonable (although not complete) confidence that it's OK to merge. Even so, I'll monitor the test results on mainline and PRs for the next few weeks.

The same [error message appeared again](https://github.com/sphinx-doc/sphinx/actions/runs/8251553945/job/22568877990?pr=12065#step:6:2150) in #12065.

Issue #12049, opened 2024-03-06
Can't extend builtin theme if it doesn't have explicit layout.html


guyer:
### Describe the bug

A theme that customizes `layout.html` cannot inherit from a theme that has an inherited `layout.html` and does not *explicitly* have its own, e.g., [default](https://github.com/sphinx-doc/sphinx/tree/master/sphinx/themes/default), [nature](https://github.com/sphinx-doc/sphinx/tree/master/sphinx/themes/nature), [sphinxdoc](https://github.com/sphinx-doc/sphinx/tree/master/sphinx/themes/sphinxdoc), or [traditional](https://github.com/sphinx-doc/sphinx/tree/master/sphinx/themes/traditional).

### How to Reproduce

If I add a custom theme to [lumache](https://www.sphinx-doc.org/en/master/tutorial/getting-started.html) that inherits from a builtin theme:
```diff
diff --git a/docs/source/_themes/lumache/layout.html b/docs/source/_themes/lumache/layout.html
new file mode 100644
index 0000000..7170ae5
--- /dev/null
+++ b/docs/source/_themes/lumache/layout.html
@@ -0,0 +1 @@
+{% extends "!layout.html" %}
diff --git a/docs/source/_themes/lumache/theme.conf b/docs/source/_themes/lumache/theme.conf
new file mode 100644
index 0000000..ffeaaf1
--- /dev/null
+++ b/docs/source/_themes/lumache/theme.conf
@@ -0,0 +1,2 @@
+[theme]
+inherit = sphinxdoc
diff --git a/docs/source/conf.py b/docs/source/conf.py
index 4ea7454..7e23856 100644
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -24,5 +24,6 @@ exclude_patterns = []
 # -- Options for HTML output -------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
 
-html_theme = 'alabaster'
+html_theme = 'lumache'
+html_theme_path = ['_themes']
 html_static_path = ['_static']
```
I get 
<details>

<summary>

```
Theme error:
An error happened in rendering the page index.
Reason: RecursionError('maximum recursion depth exceeded')
```

</summary>

```
Running Sphinx v7.1.2
building [mo]: targets for 0 po files that are out of date
writing output... 
building [html]: targets for 1 source files that are out of date
updating environment: [new config] 1 added, 0 changed, 0 removed
[...]

PR #12056, opened 2024-03-08
docs: add reference to templating from theme development.

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- Clearly link the mentioned "exclamation mark" syntax from the theme guide to the explanation of it in the templating primer.

### Detail
- (strictly speaking it is kinda hinted by reference to the previous paragraph that it's the place to look; I think we could make it more direct though)

### Relates
- Discovered during investigation of #12049.

### DIFF
--- a/doc/development/templating.rst
+++ b/doc/development/templating.rst
@@ -30,6 +30,7 @@ No.  You have several other options:
   produces pickle files with the page contents, and postprocess them using a
   custom tool, or use them in your Web application.
 
+.. _templating-primer:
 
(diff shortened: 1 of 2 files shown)


chrisjsewell (approved):
LGTM thanks 👍 want me to merge?

jayaddison:
Thank you @chrisjsewell!  Sure thing re: merge; I don't have (and not entirely sure whether I want..) write permissions.

PR #12061, opened 2024-03-12
Fix: autodoc: can't detect overloads for functions defined in other file

### LINKED ISSUE #11410: autodoc can't detect overloads for functions defined in other file
### Describe the bug

Related to #7786, overloads for functions defined in a "private" (or other) module with overloads but exposed in a public-facing `__init__.py` are not detected

### How to Reproduce

```python
# demo/__init__.py
from ._private import foobar

__all__ = ['foobar']
```

```python
# demo/_private.py
from __future__ import annotations

from typing import overload

from typing_extensions import Literal


@overload
def foobar(a: Literal[True]) -> Literal['a']:
    ...


@overload
def foobar(a: Literal[False]) -> Literal['b']:
    ...


@overload
def foobar(a: bool) -> Literal['a', 'b']:
    ...


def foobar(a: bool) -> Literal['a', 'b']:
    """return either a or b"""
    if a:
        return 'a'
    return 'b'
```

```rst
# index.rst

.. automodule:: demo
   :members:
   :undoc-members:

```

Expected output:
```text
demo.foobar(a: Literal[True]) -> Literal['a']
demo.foobar(a: Literal[False]) -> Literal['b']
demo.foobar(a: bool) -> Literal['a', 'b']

   return either a or b
```

### Environment Information

```text
Platform:              darwin; (macOS-13.3.1-x86_64-i386-64bit)
Python version:        3.10.9 (main, Mar 30 2023, 13:44:33) [Clang 14.0.0 (clang-1400.0.29.202)])
Sphinx version:        6.1.3
Docutils version:      0.19
Jinja2 version:        3.1.2
Pygments version:      2.14.0
```

### Sphinx extensions

```python
["sphinx.ext.autodoc"]
```

### Additional context

_No response_

akpircher:
Subject: fix autodoc not detecting overloads defined in other module, but exported in *this* module.

### Feature or Bugfix
- Bugfix

### Purpose
- Document overloads for a function defined in a private module but exposed publicly. 

### Detail
This PR fix is based on the solution used in #8283. Currently, the Documenter class gets overloads defined in the same file, and records the real/source module as a dependency, and then nothing else.

The fix provided here adds overloads defined in the "real" module to the public module.

I'm not entirely sure if this is the "right" way to go about this, but it's the simplest solution without a refactor.

### Relates
- Fixes #11410

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -93,6 +93,8 @@ Bugs fixed
 * #12038: Resolve ``linkcheck`` unit test timeouts on Windows by adding a readiness
   check to the test HTTP(S) server setup code.
   Patch by James Addison.
+* #11410: Fix autodoc can't detect overloads for functions defined in other file.
+  Patch by Amy Pircher.
 
 Testing
 -------
--- a/sphinx/ext/autodoc/__init__.py
+++ b/sphinx/ext/autodoc/__init__.py
@@ -925,6 +925,11 @@ class Documenter:
             try:
                 analyzer = ModuleAnalyzer.for_module(guess_modname)
                 self.directive.record_dependencies.add(analyzer.srcname)
+                if self.analyzer:
+                    analyzer.find_attr_docs()
+                    for name, signatures in analyzer.overloads.items():
+                        if name not in self.analyzer.overloads:
+                            self.analyzer.overloads[name] = signatures
             except PycodeError:
                 pass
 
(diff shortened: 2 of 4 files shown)


picnixz:
> I'm not entirely sure if this is the "right" way to go about this, but it's the simplest solution without a refactor.

I think it's better if we lookup for the correct target and fetch its ovreloading rather than polluting the analyzer of the module importing an overloaded definition. I would like a larger coverage where:

- The overloaded function is imported but should not be rendered (e.g., you ask to only document things in `__all__` but you are not re-exporting the overloaded function): in this case, the analyzer should not be called *at all*.
- The overloaded function is imported is aliased and should be rendered: render it with the aliased name.
- The overloaded function is imported is aliased and should not be rendered: same as case 1.

By the way, why does the class documenter works (i.e., the overloaded definitions seem to be here for the constructor) but not for functions?

ktbarrett:
@picnixz Any chance you could have a look at this? You seem to have a better grasp on how this needs to be done.

PR #12065, opened 2024-03-12
bump ruff version

danieleades:


.github/workflows/lint.yml
```diff
@@ -29,10 +29,12 @@ jobs:
       run: python -m pip install --upgrade pip
 
     - name: Install Ruff
-      run: python -m pip install "ruff==0.2.0"
+      run: |
```
picnixz, line 32: Just wondering, but if, by any change, we unpin the version, could you perhaps make the job fail? (so that we know that we had some error).

danieleades: i'm not sure what you mean?

the version of ruff is still pinned (it's set in `pyproject.toml`)
the `required-version` config prevents accidentally misaligning the github workflow with `pyproject.toml`, but
since the version is no longer set in the workflow with this change, it's no longer possible for them to become misaligned and the check is not necessary

picnixz: Ah no, it's not what I meant.

What I meant is that you extract the required-version from the project dependencies but what happens if the awk command fails to extract? could you perhaps just check that the extracted version is a non-empty string (and if not, echo the configuration file directly?)

I don't know how the workflow would look if it fails because no required version can be extracted so it was more a question from someone who doesn't work with workflows in general.

danieleades: ah ok. If the `sed` command fails to parse a valid version identifier from `pyproject.toml` the following command will fail:

```
python -m pip install "ruff==${ruff_version}"
```

with `ERROR: No matching distribution found for ruff=={WHATEVER_SED_RETURNED}`

that will certainly fail if sed returns an empty string

picnixz:
Thank you!

PR #12070, opened 2024-03-12
add PERF401 lint

danieleades:


picnixz:
Could you make an issue gathering all the new lints you want to add? it's just to track them and that way I know which are opened or not (PRs are sorted by update time on my side so it's a bit unordered...)

picnixz:
Thank you!

PR #12075, opened 2024-03-13
Simplify pyupgrade lint group configuration

danieleades:
simplifies config
also means future pyupgrade lints will be opt out, not opt in (but still only introduced on explicit bump of ruff version)


danieleades:
closing in favour of #12086

PR #12080, opened 2024-03-14
Add config variable html_link_path to Link paths or files into the Outdir

berlin2123:
Subject:  
Add config variable html_static_link_path to **Link** paths or files into the Output _static path instead of Copying them.

### Feature or Bugfix
- Feature

### Purpose
-  **Linking instead of copying** will dramatically reduce hard disk usage and speed  up the build process, especially when the directories or files are  particularly large. 

### Relates
- https://github.com/orgs/sphinx-doc/discussions/12051

### Testing
- just tested inside Linux, not yet Windows or MacOS


picnixz:
(I'm just cancelling the WFs for now, since otherwise the workers are busy and I want to first merge the small lint PRs)

berlin2123:
do not Merging right now.

I thinking about not only link into _static

picnixz:
For safety, I'll convert your PR into a draft

berlin2123:
For security reasons, it is no longer prepared to link into any other directory than `_static`. 

That to say, if someone set the `link-dst` inside a dynamic output directory, it may seems works ok in the beginning.
However, while the changes of source files, especially adding new source files, the buld proccess, may go into the `link-dst` to change files of some `specific file names related the new added source`, which may cause files within `link-src` to be modified directly.

berlin2123:
- suport put link-name into other folder instead of the default. 
- added warnings in docs
- could not set 'into other DST' to work, if someone don't read the docs carefully.
- wiil warning in build log, if some don't use the default DST.

jdillard:
I'm not a fan of this being introduced.

1) For the security reasons stated
2) The added testing burden (and the fact it is untested due to said burden)
3) No numbers stating the degree to which it sped up the build process

berlin2123:
For security: 
- the default dir is named as `_DoNotEditHerein` , will alert usr don‘t edit the files inside it.
- warning in document, and do not allow them setting to other DST, if someone `do not using the default dir in outdir` and `not read the document carefully`.
- no matter set the DST to work or not, if one set the DST is not inside the default, will warning in build logs.

For speed up: 
- If your project using GBs or TBs data of vidoes, copy may using dozens of minutes or even hours. however, link will not using time ( <millisecond or faster ).  
- Saving storage space is similar to speed up build process.

jayaddison:
Hi @berlin2123 - what are the goal(s) of this pull request?  To reduce disk space used by the output of Sphinx?

Sphinx generally isn't responsible for de-duplication of content written to disk -- although that can be achieved automatically by some filesystems (ISO9660, ZFS, ...).

berlin2123:
@jayaddison just used to share files into DocumentRoot(`build/html`) of all sphinx projects instead of copy files into DocumentRoot.

could be used to share local installed mathjax, images or vidoes into DocumentRoot.

berlin2123:
@jayaddison just used to share files into DocumentRoot(`build/html`) of all sphinx projects instead of copy files into DocumentRoot.

could be used to share local installed mathjax, images or vidoes into DocumentRoot.

doc/usage/configuration.rst
```diff
@@ -1262,6 +1262,79 @@ that use Sphinx's HTMLWriter class.
...
+      :file:`_static`. Such as, setting
+      :file:`html_link_path = ['../../mathjax']` will create a
+      symlink :file:`_static/_DoNotEditHerein/mathjax` inside OutDir, not
+      link files inside :file:`../../mathjax` into OutDir.
+
+   .. note::
+
+      It will work in Windows if you make sure python run with
+      :file:`os.symlink`, seeing `the Python3 docs
+      <https://docs.python.org/3/library/os.html#os.symlink>`_.
```
jayaddison, line 1334: Using [`Path.hardlink_to`](https://docs.python.org/3.10/library/pathlib.html#pathlib.Path.hardlink_to) could be a better experience, because that would not require [administrator privileges when creating these files on Windows](https://docs.python.org/3.10/library/os.html#os.symlink).

However: that is only available from Python3.10+.

berlin2123: hardlink do not support across different file systems. I do not recommend to use hardlink.

jayaddison (changes requested):
This is a large and potentially risky solution for a small problem, and I don't think we should accept that.

However, a feature request to enable symlinking of an individual file may be acceptable.

I would recommend starting with `MathJax` for that -- because that should resolve the [stated problem](https://github.com/orgs/sphinx-doc/discussions/12051#discussion-6337888).

Symlinking build files should only be enabled when an `integrity` value is provided for them -- this is a safety check to confirm three things:

  * The configuration manager has specified a particular file and content.
  * The file to symlink was correct at the time of build (we should check it during the build).
  * The web browser can check the content it retrieves (the value is a [SubResource Integrity](https://www.w3.org/TR/2016/REC-SRI-20160623/) checksum and is [included in the HTML](https://github.com/sphinx-doc/sphinx/blob/cf7d2759af0852d67288e58d823d51fe860749ca/tests/test_ext_math.py#L95-L97), so that the browser can check it).  This is necessary because the local file may be modified -- for example, by system upgrades.

Therefore: I suggest an alternative design, for one file to begin with, but that can be expanded later:

  * Check that `mathjax_path` is a local file.
  * Check that `mathjax_options['integrity']` is configured.
  * If *both* the above are true, then add the `maxjax_options['integrity']` value to a new internal dictionary, `HTMLBuilder.self._content_integrities`.
  * During the HTML build, if a `filename` (either JS or CSS) is found in `app.config.html_link_path`, *and* if `_integrity_checksum(filename) == self._content_integrities[filename]` (it is really the configured content), *then* create a symlink for it instead of copying it.

What do you think?

berlin2123:
@jayaddison,
Although I'm not quite sure how this solution of yours is implemented, for when only linking specific mathjax files is needed, this solution of yours looks like a better one.

What I actually need, however, is also to link muti video files, not copy them. For example, using sphinx to create a `documentation site with videos` or `educational course site with videos`. The video files can easily be dozens of gigabytes. Copying them is time consuming and space consuming (ext4, xfs, etc.)

jayaddison:
@berlin2123 I won't accept a pull request that handles multiple files or multiple filetypes yet.

I would accept a change that enables support for the single mathjax JS file, and then after some months of that being deployed and waiting for bug / success reports, might accept a change to expand support to other filetypes.

A faster (weeks rather than months) timescale might be possible if the test coverage for the single-file case is convincing to me and other reviewers -- but I cannot guarantee that.

berlin2123:
Ok. fine. 

 I've since found, without changing sphinx and sphinx-autobuild, a way to achieve the same effect as link. That is to configure alias when doing reverse proxy.
```
# Test environment: Ubuntu 22.4, 
# Nginx (proxy), sphinx project, sphinx-autobuild, and <The PATH to alias>, both at the same host node.

## install Nginx 
apt install nginx

## add a proxy configure file for nginx
# cat /etc/nginx/sites-available/sphinx_<project_name>
server {
    listen 32280;                 # any port, as your need.

    # your_domain_name ;    # commit just for ip only test

    location /_static/myalias_videos {
        alias /data/sphinx/static_data/videos;
        autoindex on;
    }

    location /_static/myalias_mathjax {
        alias /data/sphinx/static_data/mathjax;
        autoindex on;
    }

    location / {
        proxy_pass http://localhost:8000$request_uri;     # the port is the port when start sphinx-autobuild
    }
}

## enable nginx configure file.
ln -s /etc/nginx/sites-available/sphinx_<project_name> /etc/nginx/sites-enabled/
nginx -t    # Detect any obvious errors in the configuration file.
systemctl restart nginx.service

## test in browser
# now you can test it by access http://<localhost or IP or DomanName>:32280/_static/myalias_mathjax
```

PR #12083, opened 2024-03-14
Remove support for the Sphinx 0.5 ``intersphinx_mapping`` format

picnixz:
Extracted from #11706.

cc @jayaddison


sphinx/ext/intersphinx.py
```diff
@@ -649,36 +680,76 @@ def install_dispatcher(app: Sphinx, docname: str, source: list[str]) -> None:
...
+                name, uri, type='intersphinx', subtype='config',
+            )
+            del config.intersphinx_mapping[name]
+            continue
+
+        if (name_for_uri := seen.setdefault(uri, name)) != name:
+            logger.warning(
+                __('intersphinx_mapping[%r]: URI %r shadows URI from intersphinx_mapping[%r]; '
+                   'ignoring.'), name, uri, name_for_uri, type='intersphinx', subtype='config',
+            )
```
jayaddison:
```suggestion
        if uri in seen and seen[uri] != name:
            logger.warning(
                __('intersphinx_mapping[%r]: URI %r shadows URI from intersphinx_mapping[%r]; '
                   'ignoring.'), name, uri, seen[uri], type='intersphinx', subtype='config',
            )
```

sphinx/ext/intersphinx.py
```diff
@@ -291,9 +314,11 @@ def load_mappings(app: Sphinx) -> None:
                 inventories.main_inventory.setdefault(type, {}).update(objects)
 
 
-def _create_element_from_result(domain: Domain, inv_name: str | None,
-                                data: InventoryItem,
-                                node: pending_xref, contnode: TextElement) -> nodes.reference:
+def _create_element_from_result(
+    domain: Domain, inv_name: InventoryName | None,
+    data: InventoryItem,
+    node: pending_xref, contnode: TextElement,
```
jayaddison: (this comment applies to a few method/function signatures)

Nitpick / preference: when the argument list begins on a separate line to the method/function name and there are type hints, I'd suggest using one-line-per-argument:

```suggestion
    domain: Domain,
    inv_name: InventoryName | None,
    data: InventoryItem,
    node: pending_xref,
    contnode: TextElement,
```

sphinx/ext/intersphinx.py
```diff
@@ -649,36 +680,76 @@ def install_dispatcher(app: Sphinx, docname: str, source: list[str]) -> None:
...
         try:
-            if isinstance(value, (list, tuple)):
-                # new format
-                name, (uri, inv) = key, value
-                if not isinstance(name, str):
-                    logger.warning(__('intersphinx identifier %r is not string. Ignored'),
-                                   name)
-                    config.intersphinx_mapping.pop(key)
-                    continue
+            uri, inv = value
```
jayaddison:
```suggestion
            uri, inv = value
            inv = inv if isinstance(inv, (tuple, list)) else (inv,)
```

(to be combined with simplification of L742)

sphinx/ext/intersphinx.py
```diff
@@ -649,36 +680,76 @@ def install_dispatcher(app: Sphinx, docname: str, source: list[str]) -> None:
...
-                logger.warning(msg)
 
-            if not isinstance(inv, tuple):
-                config.intersphinx_mapping[key] = (name, (uri, (inv,)))
-            else:
-                config.intersphinx_mapping[key] = (name, (uri, inv))
-        except Exception as exc:
-            logger.warning(__('Failed to read intersphinx_mapping[%s], ignored: %r'), key, exc)
-            config.intersphinx_mapping.pop(key)
+        config.intersphinx_mapping[name] = (name, (uri, tuple(targets)))
```
jayaddison: So a mapping entry for `name` begins with `name` again in the tuple (this matches the type hint).  Is this the normalized format?

sphinx/ext/intersphinx.py
picnixz: Originally, it was not one a single line and they were grouped as such. I just put them separetely since otherwise I would have hit the limit. But perhaps it's better if I just chop down the whole list.

sphinx/ext/intersphinx.py
picnixz: It was always like that and I technically prefer keeping this one as such (otherwise tests will be probably affected and I want to minimize changes in this PR).

sphinx/ext/intersphinx.py
picnixz: I want to isolate the unpack statement from the assignment itself and I want to first check the type of *uri* before doing anything on the *inv* part.

sphinx/ext/intersphinx.py
picnixz: Without `setdefault`, you'll never populate the `seen` map.

AA-Turner:
Have rebased this on the now-split intersphinx package.

A

sphinx/ext/intersphinx/_load.py
```diff
@@ -21,55 +21,100 @@
...
-                    config.intersphinx_mapping.pop(key)
-                    continue
+            uri, inv = value
+        except Exception as exc:
+            LOGGER.warning(
+                __('Failed to read intersphinx_mapping[%s], ignored: %r'),
+                name, exc, type='intersphinx', subtype='config',
+            )
+            del config.intersphinx_mapping[name]
+            continue
```
AA-Turner: I think we should either (a) raise a ConfigError immediately, or (b) collect all errors and then raise a ConfigError with all the errors.

AA-Turner (commented):
Looks good -- main high-level feedback is that we should now be stricter on rejecting invalid formats (and add appropriate tests that `ConfigError` is raised)

A

sphinx/ext/intersphinx/_load.py
picnixz: Mmh. This works for me. I'm going to eat and will switch to gaming mode so I won't work on this anymore (but I'll work on it tomorrow). 

- raising the error immediately: less code and check in tests, but incremental debugging may be needed (multiple rebuilds if you have multiple errors).
- deferring the exception: current flow can be kept (just adding some `.append` and a final join in the final message), and you see all errors in one go.

I think deferring the exception is fine.

AA-Turner: I've pushed an initial implementation of this -- an alternative would be to instead log the errors at the test/validate site and increment an error counter, which would allow for more specific error messages. I also haven't tested with ``-W`` yet. I'm away most of today but should be able to look again this evening.

A

sphinx/ext/intersphinx/_load.py
```diff
@@ -37,70 +38,66 @@ def normalize_intersphinx_mapping(app: Sphinx, config: Config) -> None:
...
-                    name, target, type='intersphinx', subtype='config',
-                )
+                msg = __('inventory location must be a non-empty string or None')
+                errors.append((msg, name, target))
+                del config.intersphinx_mapping[name]
+                continue
 
         config.intersphinx_mapping[name] = (name, (uri, tuple(targets)))
 
+    if len(errors) > 0:
```
picnixz:
```suggestion
    if errors:
```

sphinx/ext/intersphinx/_load.py
```diff
@@ -37,70 +38,66 @@ def normalize_intersphinx_mapping(app: Sphinx, config: Config) -> None:
...
+        for (msg, name, value) in errors:
```
picnixz:
```suggestion
        for msg, name, value in errors:
```

sphinx/ext/intersphinx/_load.py
```diff
@@ -37,70 +38,66 @@ def normalize_intersphinx_mapping(app: Sphinx, config: Config) -> None:
...
+            error_msg = __('Invalid value %r in intersphinx_mapping[%r]: %s')
+            LOGGER.error(error_msg % (value, name, msg))
+        msg = __('Invalid `intersphinx_mapping` configuration (%s errors).')
```
picnixz, line 125:
```suggestion
        msg = __('Invalid `intersphinx_mapping` configuration (%d error(s)).')
```

tests/test_extensions/test_ext_inheritance_diagram.py
```diff
@@ -154,7 +154,7 @@ def test_inheritance_diagram_png_html(tmp_path, app):
     inv_file = tmp_path / 'inventory'
     inv_file.write_bytes(external_inventory)
     app.config.intersphinx_mapping = {
-        'https://example.org': str(inv_file),
+        'example': ("https://example.org", str(inv_file)),
```
picnixz:
```suggestion
        'example': ('https://example.org', str(inv_file)),
```

sphinx/ext/intersphinx/_load.py
```diff
@@ -21,55 +22,96 @@
...
+    # different project names (and thus, the build are no more reproducible)
+    # depending on which one is inserted last in the cache.
+    seen: dict[InventoryURI, InventoryName] = {}
+
+    errors = []
+    for name, value in config.intersphinx_mapping.copy().items():
+        # ensure that intersphinx projects are always named
+        if not isinstance(name, str):
+            msg = __('project identifier must be a string')
+            errors.append((msg, name, name))
```
jayaddison: Nitpicking I think, but: in these cases where we supply `name` as the third argument (`value`, logically), the resulting error messages could be a bit confusing.

picnixz: Errr... I didn't get this one. Could you explain with an example please?

picnixz:
> Do we need to update ENV_VERSION with this change?

I don't know but to be sure, let's do it :)

sphinx/ext/intersphinx/_load.py
AA-Turner: > Nitpicking I think, but: in these cases where we supply `name` as the third argument (`value`, logically), the resulting error messages could be a bit confusing.

I agree. See [`8ada5cd` (#12083)](https://github.com/sphinx-doc/sphinx/pull/12083/commits/8ada5cd299b29208a149ef9372e7963b29b2d3b8) for an improvement...

AA-Turner:
> Do we need to update `ENV_VERSION` with this change?

We shouldn't need to, as the resulting data that gets stored in the environment is unchanged -- we haven't changed how we do the normalisation.

A

picnixz:
Actually, by doing so, it would at least invalidate the cache (which could contain old intersphinx references)

AA-Turner:
> just tell me whether it's you or me that need to work on the branch now

I'm done for now, don't worry

jayaddison:
@jayaddison 
> > Do we need to update `ENV_VERSION` with this change?

@AA-Turner  
> We shouldn't need to, as the resulting data that gets stored in the environment is unchanged -- we haven't changed how we do the normalisation.

@picnixz 
> Actually, by doing so, it would at least invalidate the cache (which could contain old intersphinx references). But if you want, you can revert the last commit (just tell me whether it's you or me that need to work on the branch now; I don't want race issues)

Ok.  Perhaps reverting it here, but adding it in #12087 could make sense?  Reasoning: it could be unexpected to have resolved non-deterministic inventories, and yet not rewrite existing environment files affected by that after an upgrade to 8.x takes place.

picnixz:
> Ok. Perhaps reverting it here

Actually, by invalidating the environment, you somewhat fix the bugs in #12087. I prefer erring on the safe side at the cost of a single rebuild (also, it's kind of a major change where now we issue errors instead).

Also, we are just storing less information so technically I would say we should still create a new environment (from a typing PoV, it's no more possible to have a None name, now it should always be a string).

sphinx/ext/intersphinx/_load.py
```diff
@@ -130,7 +199,7 @@ def fetch_inventory_group(
                     return True
         return False
     finally:
-        if failures == []:
+        if not failures:
```
jayaddison, line 202: This shouldn't cause any problems, right?  (`failures` is _definitely_ a `list`, and even if this causes bytecode cache invalidation, those caches are per-module anyway, and other changes in this file would cause that anyway)

AA-Turner (approved):

AA-Turner:
@picnixz do you have any other changes, or is this ready to merge from your perspective?

A

sphinx/ext/intersphinx/_load.py
AA-Turner: This is fine, it's a style point (I sometimes prefer doing checks like this or like ``len(obj) == 0`` as it is a quick reminder what *kind* of truth testing you're doing.

jayaddison (approved):

picnixz:
> do you have any other changes, or is this ready to merge from your perspective?

I don't have anything else! I expect people to be surprised if they build very old projects but I think the test coverage is sufficient. I'm just unsure whether the typing definitions I put should be kept in shared or not but this is something we can decide later (for now, let's keep them like that because the other PR is based on this one and I don't want to again solve conflicts). By the way, thank you for the CHANGES entry.

AA-Turner:
Thanks all!

A

EwoutH:
Hi!

I just stumbled onto this PR as the root cause of our docs breaking. While I looked back and there was a deprecation warning cast by Sphinx 7.0 (which is nice, thanks!), the error message in 8.0 could be a bit clearer, pointing directly to some docs or migration guide on how to update it.
```
ERROR: Invalid value `None` in intersphinx_mapping['http://docs.python.org/']. Expected a two-element tuple or list.
```
For anyone also stumbling on this PR, the our fix was replacing:
```Python
intersphinx_mapping = {"http://docs.python.org/": None}
```
with 
```Python
intersphinx_mapping = {'python': ('https://docs.python.org/3', None)}
```
As following [these docs](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping).

I was a bit amazed to see a 10 year old line of code breaking our docs configuration. Doesn't happen often!

PR #12086, opened 2024-03-14
Ruff simplify config

danieleades:
simplify ruff config


picnixz:
Should I understand that, in the future, if ruff introduces new rules, they will be enabled but if any of them introduce an error, the dependabot will say "hey I cannot upgrade to that ruff"?

danieleades:
> Should I understand that, in the future, if ruff introduces new rules, they will be enabled but if any of them introduce an error, the dependabot will say "hey I cannot upgrade to that ruff"?

If a ruff update introduces new lints that result in new errors, that will cause the CI jobs to fail on the dependabot PR. That's a signal that manual changes are needed before ruff can be updated

PR #12088, opened 2024-03-14
extend FURB lints

danieleades:


picnixz:
I think the tests failed because of the _StrPath that we needed to keep for backward compatibility.

picnixz:
I don't understand. It works locally but not remotely??

sphinx/ext/napoleon/docstring.py
```diff
@@ -1340,10 +1341,7 @@ def translate(
             return []
 
         # apply type aliases
-        items = [
-            translate(func, description, role)
-            for func, description, role in items
-        ]
+        items = list(starmap(translate, items))
```
picnixz, line 1344: I think Adam had a discussion about this one in a previous FURB-PR where they preferred an explicit call over `starmap`. Personally I don't mind the starmap because I'm used to parse "complex" calls and my IDE helps me a lot if I need to inspect the types, but it's maybe better to avoid using `starmap` in general.

sphinx/transforms/post_transforms/__init__.py
```diff
@@ -155,7 +156,7 @@ def resolve_anyref(
             def stringify(name: str, node: Element) -> str:
                 reftitle = node.get('reftitle', node.astext())
                 return f':{name}:`{reftitle}`'
-            candidates = ' or '.join(stringify(name, role) for name, role in results)
+            candidates = ' or '.join(starmap(stringify, results))
```
picnixz, line 159: Same comment as above.

tests/test_extensions/test_ext_apidoc.py
```diff
@@ -1,23 +1,29 @@
...
 
 
+class Apidoc(NamedTuple):
+    coderoot: Path
+    outdir: Path
+
+
 @pytest.fixture()
-def apidoc(rootdir, tmp_path, apidoc_params):
+def apidoc(rootdir, tmp_path, apidoc_params) -> Apidoc:
```
picnixz:
```suggestion
def apidoc(
    rootdir: Path, 
    tmp_path: Path, 
    apidoc_params: tuple[Sequence[Any], Mapping[str, Any]],
) -> Apidoc:
```

+ add imports + ``from __future__ import annotations``.

While I don't really care about typing inside a test function, I think it's good to type the fixture definitions themselves.

tests/test_extensions/test_ext_apidoc.py
```diff
@@ -56,19 +62,16 @@ def test_pep_0420_enabled(make_app, apidoc):
     assert (outdir / 'a.b.e.rst').is_file()
     assert (outdir / 'a.b.x.rst').is_file()
 
-    with open(outdir / 'a.b.c.rst', encoding='utf-8') as f:
-        rst = f.read()
-        assert "automodule:: a.b.c.d\n" in rst
-        assert "automodule:: a.b.c\n" in rst
+    rst = (outdir / 'a.b.c.rst').read_text(encoding='utf-8')
```
picnixz: How come this was done if you ignore FURB101?

tests/test_extensions/test_ext_apidoc.py
```diff
@@ -356,10 +347,9 @@ def test_toc_all_references_should_exist_pep420_disabled(make_app, apidoc):
         'TOC:\n{}'.format(", ".join(missing_files), toc)
 
 
-def extract_toc(path):
+def extract_toc(path: Path):
```
picnixz:
```suggestion
def extract_toc(path: Path) -> None:
```

(You know I don't like half-typed defs)

utils/bump_version.py
```diff
@@ -21,7 +22,7 @@ def stringify_version(version_info, in_develop=True):
     return version
 
 
-def bump_version(path, version_info, in_develop=True) -> None:
+def bump_version(path: Path, version_info, in_develop=True) -> None:
```
picnixz: This one will be updated after fixing the conflicts right?

tests/test_extensions/test_ext_apidoc.py
danieleades: unsure...
but it's definitely an improvement

sphinx/ext/napoleon/docstring.py
danieleades: maybe because i'm increasingly spending my time with more functional languages, but starmap feels very natural and succinct to me

tests/test_extensions/test_ext_apidoc.py
danieleades: i think i started applying FURB101 and decided that changing from str to Path needs its own PR.
but i think these changes stand on their own and should be kept

sphinx/ext/napoleon/docstring.py
picnixz: For me it's fine, but there was a recent PR where we preferred keeping the non-starmap version. Mmh..., ok the `translate` function is close to the call and it's typed. Let's keep it.

picnixz (approved):

picnixz:
Thank you!

PR #12089, opened 2024-03-14
[tests] add public properties ``SphinxTestApp.status`` and ``SphinxTestApp.warning``

picnixz:
This is a first step to the test suite modernization. I can extract those changes from my current implementation without changing anything (which is good from a review PoV).

cc @jayaddison 

btw, the name of the branch is totally unrelated (at first, I just wanted to type things but then I saw that to type them I need this small change).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -106,7 +106,7 @@ Bugs fixed
 
 Testing
 -------
-* #11285: :func:`!pytest.mark.sphinx` and :class:`sphinx.testing.util.SphinxTestApp`
+* #11285: :func:`!pytest.mark.sphinx` and :class:`!sphinx.testing.util.SphinxTestApp`
   accept *warningiserror*, *keep_going* and *verbosity* as keyword arguments.
   Patch by Bénédikt Tran.
 * #11285: :class:`!sphinx.testing.util.SphinxTestApp` *status* and *warning*
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,11 @@ Deprecated
 * #11693: Support for old-style :file:`Makefile` and :file:`make.bat` output
   in :program:`sphinx-quickstart`, and the associated options :option:`!-M`,
   :option:`!-m`, :option:`!--no-use-make-mode`, and :option:`!--use-make-mode`.
+* #11285: Direct access to :attr:`!sphinx.testing.util.SphinxTestApp._status`
+  or :attr:`!sphinx.testing.util.SphinxTestApp._warning` is deprecated. Use
+  the public properties :attr:`!sphinx.testing.util.SphinxTestApp.status`
+  and :attr:`!sphinx.testing.util.SphinxTestApp.warning` instead.
+  Patch by Bénédikt Tran.
 
 Features added
 --------------
@@ -99,6 +104,19 @@ Bugs fixed
 
 Testing
 -------
+* #11285: :func:`!pytest.mark.sphinx` requires keyword arguments, except for
+  the builder name which can still be given as the first positional argument.
+  Patch by Bénédikt Tran.
+* #11285: :func:`!pytest.mark.sphinx` accepts *warningiserror*, *keep_going*
+  and *verbosity* as additional keyword arguments.
+  Patch by Bénédikt Tran.
+* #11285: :class:`!sphinx.testing.util.SphinxTestApp` *srcdir* argument is
+  now mandatory (previously, this was checked with an assertion).
+  Patch by Bénédikt Tran.
+* #11285: :class:`!sphinx.testing.util.SphinxTestApp` *status* and *warning*
+  arguments are checked to be :class:`io.StringIO` objects (the public API
+  incorrectly assumed this without checking it).
+  Patch by Bénédikt Tran.
 
 Release 7.2.6 (released Sep 13, 2023)
 =====================================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -106,14 +106,8 @@ Bugs fixed
 
 Testing
 -------
-* #11285: :func:`!pytest.mark.sphinx` requires keyword arguments, except for
-  the builder name which can still be given as the first positional argument.
-  Patch by Bénédikt Tran.
-* #11285: :func:`!pytest.mark.sphinx` accepts *warningiserror*, *keep_going*
-  and *verbosity* as additional keyword arguments.
-  Patch by Bénédikt Tran.
-* #11285: :class:`!sphinx.testing.util.SphinxTestApp` *srcdir* argument is
-  now mandatory (previously, this was checked with an assertion).
+* #11285: :func:`!pytest.mark.sphinx` and :class:`sphinx.testing.util.SphinxTestApp`
+  accept *warningiserror*, *keep_going* and *verbosity* as keyword arguments.
   Patch by Bénédikt Tran.
 * #11285: :class:`!sphinx.testing.util.SphinxTestApp` *status* and *warning*
   arguments are checked to be :class:`io.StringIO` objects (the public API
--- a/sphinx/testing/util.py
+++ b/sphinx/testing/util.py
@@ -180,7 +180,7 @@ class SphinxTestApp(sphinx.application.Sphinx):
 
     @property
     def status(self) -> StringIO:
-        """The in-memory I/O for the application status messages."""
+        """The in-memory text I/O for the application status messages."""
         # sphinx.application.Sphinx uses StringIO for a quiet stream
         assert isinstance(self._status, StringIO)
         return self._status
--- a/sphinx/testing/util.py
+++ b/sphinx/testing/util.py
@@ -98,33 +98,31 @@ class SphinxTestApp(sphinx.application.Sphinx):
     directory, whereas in the latter, the user must provide it themselves.
     """
 
-    # Allow the builder name to be passed as a keyword argument
-    # but only make it positional-only for ``pytest.mark.sphinx``
-    # so that an exception can be raised if the constructor is
-    # directly called and multiple values for the builder name
-    # are given.
+    # see https://github.com/sphinx-doc/sphinx/pull/12089 for the
+    # discussion on how the signature of this class should be used
 
     def __init__(
         self,
-        /,
+        /,  # to allow 'self' as an extras
         buildername: str = 'html',
-        *,
-        srcdir: Path,
+        srcdir: Path | None = None,
+        builddir: Path | None = None,  # extra constructor argument
+        freshenv: bool = False,  # argument is not in the same order as in the superclass
         confoverrides: dict[str, Any] | None = None,
         status: StringIO | None = None,
         warning: StringIO | None = None,
-        freshenv: bool = False,
-        warningiserror: bool = False,
         tags: list[str] | None = None,
+        docutils_conf: str | None = None,  # extra constructor argument
         verbosity: int = 0,
         parallel: int = 0,
+        # additional arguments at the end to keep the signature
         keep_going: bool = False,
-        # extra constructor arguments
-        builddir: Path | None = None,
-        docutils_conf: str | None = None,
+        warningiserror: bool = False,  # argument is not in the same order as in the superclass
         # unknown keyword arguments
         **extras: Any,
     ) -> None:
+        assert srcdir is not None
+
         if verbosity == -1:
             quiet = True
             verbosity = 0
--- a/sphinx/testing/util.py
+++ b/sphinx/testing/util.py
@@ -1,4 +1,5 @@
 """Sphinx test suite utilities"""
+
 from __future__ import annotations
 
 import contextlib
@@ -6,7 +7,9 @@ import os
 import re
 import sys
 import warnings
-from typing import IO, TYPE_CHECKING, Any
+from io import StringIO
+from types import MappingProxyType
+from typing import TYPE_CHECKING
 from xml.etree import ElementTree
 
 from docutils import nodes
@@ -18,8 +21,9 @@ import sphinx.pycode
 from sphinx.util.docutils import additional_nodes
 
 if TYPE_CHECKING:
-    from io import StringIO
+    from collections.abc import Mapping
     from pathlib import Path
+    from typing import Any
 
     from docutils.nodes import Node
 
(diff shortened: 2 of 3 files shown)


sphinx/testing/util.py
```diff
@@ -73,28 +77,75 @@ def etree_parse(path: str) -> Any:
...
+        verbosity: int = 0,
         parallel: int = 0,
+        keep_going: bool = False,
+        # extra constructor arguments
+        builddir: Path | None = None,
+        docutils_conf: str | None = None,
+        # unknown keyword arguments
+        **extras: Any,
     ) -> None:
-        assert srcdir is not None
```
picnixz, line 97: This assert was really unnecessary since it means that you must specify *srcdir* yourself. But with the previous signature, this means that you don't have a "default" srcdir... In particular, I highly doubt that people write something like `SphinxTestApp('html', srcdir)` and had `srcdir=...`. 

Anyway, the way the markers were parsed, people would have anyway written `srcdir=...` in the marker and thus it would have been a keyword argument. Also, I don't think this class was ever meant to be public in the first place since we never documented it properly and instead provided the `make_app` fixture.

sphinx/testing/util.py
```diff
@@ -138,12 +207,6 @@ def build(self, force_all: bool = False, filenames: list[str] | None = None) ->
 
 
 class SphinxTestAppWrapperForSkipBuilding:
-    """A wrapper for SphinxTestApp.
```
picnixz: Ah, I removed the comment here. Will revert.

CHANGES.rst
```diff
@@ -13,6 +13,11 @@ Deprecated
 * #11693: Support for old-style :file:`Makefile` and :file:`make.bat` output
   in :program:`sphinx-quickstart`, and the associated options :option:`!-M`,
   :option:`!-m`, :option:`!--no-use-make-mode`, and :option:`!--use-make-mode`.
+* #11285: Direct access to :attr:`!sphinx.testing.util.SphinxTestApp._status`
+  or :attr:`!sphinx.testing.util.SphinxTestApp._warning` is deprecated. Use
+  the public properties :attr:`!sphinx.testing.util.SphinxTestApp.status`
+  and :attr:`!sphinx.testing.util.SphinxTestApp.warning` instead.
```
jayaddison, line 19: I don't see this mentioned in the linked issue thread - what's the reason to change these?

It also diverges a bit from the [main `Sphinx` class `._warning` attribute](https://github.com/sphinx-doc/sphinx/blob/23dd754f6375afb62e8af1e9e3b17d20b113fcc0/sphinx/application.py#L177-L180) - is that OK?

picnixz: The reason is that the fixture actually returns `app._status` but it's return type is `StringIO`. Ah perhaps I should just remove the linked issue because the whole implementation I'm doing is fixing things here and there that were here for years but were actually wrong (and no one complained at all). 

The reason why I changed this is because in the original signature of the SphinxTestApp constructor, status and warning were of IO types. However, the annotation was StringIO... Also, the current implementation only works if you are using a stringIO (it's because of the shared_result thing), so... even if you forcibly specifyied in your marker a non-StringIO status object... + use shared_result, it would miserably fail.

Since it's a private attribute, it shouldn't even be exposed normally so I used a public property with a correct return type. By the way, there won't be any deprecation warning, it's just that it's better if you use the public property anyway.

picnixz:
I wanted to have two implementation of two plugins side-by-side and use the new plugin only for some tests but it's not possible in pytest... So I'll need to merge both the new and legacy implementation together and maintain them so that we can incrementally change the tests.

chrisjsewell:
> This is a first step to the test suite modernization. 

Heya, can you explain what this means, as obviously its quite vague:

- What are the current issues with the test suite?
- and how does this PR (and further ones) seek to address these issues?

sphinx/testing/util.py
```diff
@@ -112,17 +163,35 @@ def __init__(
...
+        assert isinstance(self._status, StringIO)
+        return self._status
+
+    @property
+    def warning(self) -> StringIO:
+        """The in-memory text I/O for the application warning messages."""
+        # sphinx.application.Sphinx uses StringIO for a quiet stream
+        assert isinstance(self._warning, StringIO)
+        return self._warning
+
```
chrisjsewell, line 192: this all seems reasonable 👍

picnixz:
> What are the current issues with the test suite?

There are many many side-effects which makes it impossible to use with `xdist` (see #11285) (I forgot to mention that PR actually). By the way, the test suite takes roughly 8 minutes to finish. With xdist, it should boil down to 2-3 minutes I think (and locally it's around 1 minute in general).

> and how does this PR (and further ones) seek to address these issues?

Each test should be isolated in a way that it does not affect other tests. Some tests use the same test-root but they sometimes change the sources in-place. Other tests import modules but sometimes they do not unload them and thus annotations are kept (and mess other tests). We should never have side-effects like that in general. Also many tests are actually C/C of others and some of them use `app.build(force_all=True)` but not all tests should do this ! IMO, `force_all=True` should only be done *inside* a test if more than *one* build is done. If the test is parametrized, then the parametrization is semantically equivalent to a big for loop so you just need a single `app.build()` with the same test directory for those parametrized tests. 

This PR is actually just one thing I could extract from the huge PR with all the modifications. Also, there are various typing issues that were never resolved or assumptions that were wrong (like, assuming that `app._status` is a StringIO is actually incorrect because nothing is checked.

sphinx/testing/util.py
```diff
@@ -73,28 +77,75 @@ def etree_parse(path: str) -> Any:
...
-        docutils_conf: str | None = None,
+        verbosity: int = 0,
         parallel: int = 0,
+        keep_going: bool = False,
+        # extra constructor arguments
+        builddir: Path | None = None,
+        docutils_conf: str | None = None,
+        # unknown keyword arguments
+        **extras: Any,
     ) -> None:
```
chrisjsewell: changing this signature will defintely break a lot of sphinx extension test suites, is there really a need to do this?

Using `/` and `*` is definitely recommended for **new** APIs, to make them easier to change later without breaking the public API.
I feel adding them to existing APIs is essentially doing the inverse of what they are inteded for.

picnixz: If I were to play the Devil's advocate I'd say "this was never documented and was still marked as experimental since 1.6". I can find a way to make it find without breaking changes though. The issue is that I honestly don't kwow who would actually pass the arguments without using keywords. 

Having something like that would force people to understand which parameter they are using. Note that the class inherits from Sphinx but did not even expose every other constructor arguments that the super class did (like verbosity, which I think *should* be exposed because the marker is really many to be used as a "constructor configuration" IMO). 

But let's avoid breaking changes, and I'll just parse the things differently.

chrisjsewell:
> Each test should be isolated in a way that it does not affect other tests.

Completely agree, this should be a goal of the test suite 👍 
I don't think this really has anything to do with "modernization", its just fixing something that is currently broken 😅

picnixz:
>  I don't think this really has anything to do with "modernization", its just fixing something that is currently broken 😅

Ah yes, actually it's only some parts of the test suite that need to be modernized (like, using monkeypatch instead of manual patching).

picnixz:
I'll patch the signature this w-e.

picnixz:
Here's my thought:

- Currently, the signature of SphinxTestApp is

```python
def __init__(self, buildername='html', srcdir=None, builddir=..., ...)
```

and inside the constructor we have `assert srcdir is not None`. In other words, you cannot write:

```python
app = SphinxTestApp()
app = SphinxTestApp('html')
app = SphinxTestApp('html', builddir=...)
app = SphinxTestApp(builddir=...)

srcdir = None
app = SphinxTestApp('html', srcdir)
```

In addition, we have

```python
kwargs['srcdir'] = srcdir = sphinx_test_tempdir / kwargs.get('srcdir', testroot)
```

In other words, there will ALWAYS be a `srcdir` keyword argument when using the marker. Thus, you cannot have both the positional and the keyword argument at the same time (meaning that you will not be able to use the positional one argument in the marker!). In particular, when you use `make_app` instead of `SphinxTestApp`, you will call the constructor with a keyword argument. 

Finally, the `make_app` docstring says

> Provides make_app function to initialize SphinxTestApp instance.
> if you want to initialize 'app' in your test function. please use this
> instead of using SphinxTestApp class [directly].

In other words, the SphinxTestApp class is NOT meant to be part of the public API. As such, you should only be able to construct a SphinxTestApp instance using `srcdir` as a keyword argument. Therefore, the signature I suggest here where *srcdir* (and whatever follows) is a keyword argument is actually the one that we expected for years.

Finally, the sphinx marker never included the `srcdir` as a marker keyword so... it was probably not meant to be supported from a public PoV (and only meant for us). 

Note that in Sphinx 4.2, we added the *builddir* argument but just after the *srcdir* one. So, we already changed the signature at that time ! and any codebase using positional arguments would likely have been updated since then. I think there shouldn't be a lot of codebase which does not use keyword arguments for constructing a SphinxTestApp unless they really want to rely on a fragile signature.

As such, I will still do this kind of breaking changes because 1) the SphinxTestApp class was never meant to be public 2) SphinxTestApp is not meant to be public and only constructed via `make_app` whose implementation we gave until now couldn't work if we did not use a keyword argument for *srcdir* 3) the keyword argument is mandatory, hence everything after should be specified as keyword argument as well (from a syntax PoV).

sphinx/testing/util.py
picnixz: - The '/' is not really an issue (it just allows you that 'self' as an extra keyword).
- The '*' is needed (see explanation below)

CHANGES.rst
jayaddison: I'm not worried about breaking downstream applications, but sometimes I build the projects in the `tests/roots` externally by calling `sphinx-build` directly on them so that I can more easily inspect/compare how the constructed content changes.  So that makes me cautious about any changes that might mean that `SphinxTestApp` behaves differently to `sphinx.application.Sphinx`.

(it's OK though - it's a small detail and I think review/understanding of the other aspects is more important)

picnixz: > I build the projects in the tests/roots externally by calling sphinx-build

The command-line is not affected by that. However, it's real that the SphinxTestApp (current) implementation differs from its parent class (which I really found it weird).

picnixz:
Since the plugin will be a bit long to implement, I'll use separate branch. That way, I can rebase and you can see the diffs of the 3 PRs instead of having every chunk at each PR.

chrisjsewell:
> In other words, the SphinxTestApp class is NOT meant to be part of the public API.

whether it "meant" to be or not, unfotunately it was not named `_SphinxTestApp` and so it IS part of the public API,
an arbitrary comment (not even on `SphinxTestApp` itself) does not change that 😬

picnixz:
> whether it "meant" to be or not, unfotunately it was not named _SphinxTestApp and so it IS part of the public API,

Actually, this one was not enforced during the years (and there are a lot of internals that are named without `_`). Anyway, now that I think about it, I can remove '*' from the signature and do it as before (though there will be some ugly "srcdir: Path | None = None") but I can do something at the level of the marker instead.

chrisjsewell:
> Actually, this one was not enforced during the years (and there are a lot of internals that are named without `_`).

Well they are not internals then, they are part of the public API 😉 

but yeh thanks @picnixz, I don't wanna block you from moving things forward, and obviously its annoying when things have maybe not have been handled correctly in the past (one of my loves of Rust is that everything is private and immutable by default, so its more diffucult to make these mistakes).

But we should be very careful if breaking the "public API"; that it is really needed and, if so, then it is clearly documented

chrisjsewell (approved):
cheers!

I would commit it with a title like "Add `SphinxTestApp.status` and `SphinxTestApp.warning` public properties" though,
since the current PR title/description is not very explanatory 😅

jayaddison:
I'm still not sure about this.  I'm OK with the changes to make the `__init__` signature of `SphinxTestApp` more similar to `Sphinx` itself -- the test class should tend towards an accurate representation of the way a normal Sphinx build would behave.

But I don't understand the reasoning for making methods public on `SphinxTestApp` that aren't public on `Sphinx`.  Do we plan to make those public on `Sphinx` in future?

chrisjsewell:
> But I don't understand the reasoning for making methods public on `SphinxTestApp` that aren't public on `Sphinx`. Do we plan to make those public on `Sphinx` in future?

Checking the output, in particular of warnings, is highly common in sphinx extension test suites.

You just have to look at https://github.com/search?q=app._warning&type=code, to see the many test suites currently having to use the private attribute for this

jayaddison:
Ok, thank you.  People do tend to find and copy code in test cases into non-test code, so I worry that making the attributes public on `SphinxTestApp` only could cause confusion at some point.  I understand that people are already using it, and that extra type validation is useful, but I'm not sure about the visibility modification yet.

picnixz:
It's something that we must somewhat expose because there is no other way if you don't use the app fixture + status/warning fixture. If you use make_app, you need to access the status/warnings like that.

jayaddison:
I _think_ my concerns are unfounded, and so please go ahead.  Thank you both for explaining.

So that it's communicated, I'll attempt to explain what I was worrying about:

* We push this change.
* Code that uses `SphinxTestApp` begins using the public accessor method (looks nicer, more acceptable, type-checked - a good idea).
* Other code copies stuff from that test code and attempts it in production and it breaks.

Why I think my concerns are unfounded:

* It should fairly clearly be test code, so it should be rare for people to copy it.
* If they do move it into production it'll break quickly because the relevant attribute names don't exist on the `Sphinx` class.

...and also I can't think of a suggestion to type-check an attribute value on an object that wouldn't require introducing other dependencies or doing unnecessarily complicated things, and I don't think we should do those.

So, longwinded explanation, but: I withdraw my concerns.

picnixz:
No worries! I'll meege tomorrow (from my computer rather than my phone).

picnixz:
@chrisjsewell Fun fact, I forgot about the commit message ! But since I want to push a hotfix, I'll just amend the commit message in this hotfix.

jayaddison:
@chrisjsewell @picnixz some trivia from archaeological code curiosity: the way that test code used to be provided with the `status` and `warning` output from a Sphinx test app build was through that the fixture (as available using a [`@with_app` decorator](https://github.com/sphinx-doc/sphinx/blob/5dd51a7fa490d96425962fd6e942d6ae6ca74cd2/tests/util.py#L348)) would:

* Create fresh `StringIO` objects to hold the results.
* Call `SphinxTestApp` with those as warning/status attribute initializers.
* Call the underlying test case with the `StringIO` objects as arguments, so that the test can inspect them.

That approach was deprecated during the migration to `pytest` as far as I can tell.  It seems like it was quite a neat approach.

jayaddison:
> @chrisjsewell @picnixz some trivia from archaeological code curiosity: the way that test code used to be provided with the `status` and `warning` output from a Sphinx test app build was that the fixture (as available using a [`@with_app` decorator](https://github.com/sphinx-doc/sphinx/blob/5dd51a7fa490d96425962fd6e942d6ae6ca74cd2/tests/util.py#L348)) would:
> 
> * Create fresh `StringIO` objects to hold the results.
> * Call `SphinxTestApp` with those as warning/status attribute initializers.
> * Call the underlying test case with the `StringIO` objects as arguments, so that the test can inspect them.
> 
> That approach was deprecated during the migration to `pytest` as far as I can tell. It seems like it was quite a neat approach.

The legacy of this remains in the current code, in a `with_text_app` decorator: https://github.com/sphinx-doc/sphinx/blob/bf0bec3b4b8c019acad4a77a9e228de4e65b538b/tests/test_builders/test_build_text.py#L9-L15 that remains in use.

picnixz:
It's still being used when using the shared result by the way. You can see it in the `app_params()` fixture.

jayaddison:
The way that the parameters were added as (optional, I think?) test arguments is what I think was particularly nice (although they were still `StringIO` objects, where I guess most test authors would want to compare against `str` values directly).

https://github.com/sphinx-doc/sphinx/blob/5dd51a7fa490d96425962fd6e942d6ae6ca74cd2/tests/test_build.py#L82

picnixz:
Actually, that's what I would have expected as well because most of the time you do `status.getvalue()` when you request the fixture... so... I think we can (in the future) have a new fixture which simply returns `status.getvalue()` directly (something like `app_stdout` with normal messages and `app_stderr` with warnings).

chrisjsewell:
> when you request the fixture

personally I feel that having separate fixtures is over-complicating things 😬  

but why we are on the subject, actually one of the most annoying things is having to deal with colored output of warnings;
in general, when you want to check warnings, you don't want the to have console color codes in.
Unless I am missing something, there is no clear way to temporarily turn off color codes

picnixz:
Mmh, you should technically be able to turn off colors by adding `NO_COLOR` to the environment.

chrisjsewell:
> EDIT 2: ok it's easy to make it work. I'll do it tomorrow.

I would encorage you to have a look at #12130 first, lets have a think about this from a top-level perspective 😄

picnixz:
> personally I feel that having separate fixtures is over-complicating things 😬

Well, I agree that I would prefer using `app.status` instead of requesting `status`, but I think that having a fixture which gives you a view on the lines without colors (or possibly with colors) is better (even better, we should return an object that can be compared to list of strings + handle regex, a bit similar to `_pytest.pytester.LineMatcher`).

> I would encorage you to have a look at https://github.com/sphinx-doc/sphinx/issues/12130 first, lets have a think about this from a top-level perspective 😄

I've looked at it but to me it's something unrelated. Actually, the main issue is that the `ColorizeFormatter` of the logger does not care about whether colorization should be done or not.

chrisjsewell:
> Actually, the main issue is that the `ColorizeFormatter` of the logger does not care about whether colorization should be done or not.

As I've commented  in https://github.com/sphinx-doc/sphinx/pull/12134#issuecomment-2006460300, it does already obey the colorization setting

PR #12090, opened 2024-03-14
run mypy on the utils/ directory

danieleades:


utils/bump_version.py
```diff
@@ -87,7 +91,7 @@ def processing(message):
 
 
 class Changes:
-    def __init__(self, path):
+    def __init__(self, path: Path):
```
picnixz: I personally don't like half-typed methods.

```suggestion
    def __init__(self, path: Path) -> None:
```

utils/bump_version.py
```diff
@@ -120,7 +124,7 @@ def finalize_release_date(self) -> None:
             f.write('=' * len(heading) + '\n')
             f.write(self.filter_empty_sections(body))
 
-    def add_release(self, version_info) -> None:
+    def add_release(self, version_info) -> None:  # type: ignore[no-untyped-def]
```
picnixz: I think we can have:

```python
VersionInfoType = Union[
    tuple[int, int, int, str], 
    tuple[int, int, int, str, int]
]
```

utils/babel_runner.py
```diff
@@ -233,16 +233,13 @@ def _get_logger():
 
     os.chdir(ROOT)
     if action == "extract":
-        raise SystemExit(run_extract())
+        run_extract()
```
picnixz, line 236: I kind of like the raise SystemExit(func()). And I prefer having the correct exit code every time.

danieleades: the trouble is, the `func()` doesn't return anything. Did they return error codes at some point in the past?

picnixz: Oh. I didn't know. But probably it returned 0 or 1 according to the code. At one time, I had asked Adam why using SystemExit instead of sys.exit() and I think they told me it's better that way. Now, if nothing is being returned..., then let's leave it like that.

Can you, however, change the if-if-if into if-elif-elif (it makes sure we don't have an if-duplicate, though it wouldn't happen, but better safe than sorry)?

utils/bump_version.py
danieleades: i've added this alias, however there's a couple of places where the code assumes that version_info[4] exists. I've added ignore statements for these for now, but it will need a more careful look

picnixz: > where the code assumes that version_info[4] exists.

Oh, then you can maybe remove the union (actually, I incorrectly counted my elements..., I don't think we should have the first `tuple[int, int, int, str]`)

utils/babel_runner.py
danieleades: have a look and see if you're happy with the approach

utils/bump_version.py
```diff
@@ -1,27 +1,41 @@
...
-package_dir = os.path.abspath(os.path.join(script_dir, '..'))
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+script_dir = Path(__file__).parent
+package_dir = script_dir.parent
 
 RELEASE_TYPE = {'a': 'alpha', 'b': 'beta'}
 
+VersionInfo: TypeAlias = Union[tuple[int, int, int, str], tuple[int, int, int, str, int]]
```
picnixz: From what I see, the `tuple[int, int, int, str]` form is never used, so we can just have `tuple[int, int, int, str, int]`.

utils/bump_version.py
```diff
@@ -42,7 +56,7 @@ def bump_version(path, version_info, in_develop=True) -> None:
         f.write('\n'.join(lines) + '\n')
 
 
-def parse_version(version):
+def parse_version(version: str) -> VersionInfo:
```
picnixz, line 59: The construction normalizes the output so that it has 5 elements, so yes, let's remove the union in the type.

utils/babel_runner.py
picnixz: I am very happy!

picnixz (approved):

picnixz:
Thank you!

PR #12091, opened 2024-03-14
[tests] skip tests that need remote content if host is offline

picnixz:
Don't forget people who work offline!

### DIFF
--- a/sphinx/testing/fixtures.py
+++ b/sphinx/testing/fixtures.py
@@ -257,7 +257,7 @@ def sphinx_remote_query_address() -> tuple[str, int]:
 @pytest.fixture(scope='session')
 def if_online(  # NoQA: PT004
     request: pytest.FixtureRequest,
-    sphinx_remote_query_address,
+    sphinx_remote_query_address: tuple[str, int],
 ) -> None:
     """Skip the test if the host has no connection.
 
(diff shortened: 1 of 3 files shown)


jayaddison:
@picnixz `test_run_epubcheck` unexpectedly fails offline too, and could be worth adding to the list here.

The reason for that is quite obscure: it seems that we write a slightly different output `.epub` file containing 'remote references' if we're unable to download the content locally during the build.  When `epubcheck` checks the (different) resulting output, it isn't happy about those remote references and complains.

I don't know of a way to run GitHub Actions workflows in a way that prevents them from making outbound network connectivity, but such a thing might exist.

picnixz:
I actually don't run it because I don't have epubcheck.. so... I never saw the error.

jayaddison:
Yep, neither had I previously - I'm trying to make the tests behave more similarly offline and online after seeing this change.  In particular: replacing some of the remote image links with content served by our good friend `localhost:7777`.

When I did that, I tried to get as many skipped tests to run as possible (in case I'd missed any other cases), and found that `test_run_epubcheck` failed.. but only when offline!  Weird.

jayaddison:
Filed the details as #12094 - there are some easy-ish small improvements, and it may be low priority, but I'll try to suggest one or two adjustments soon.

PR #12093, opened 2024-03-14
[tests] Implementation of the new testing plugin (backward-compatible) [part 2]

picnixz:
Follows #12089.

This is the smallest possible set of changes that could be done in order to introduce the new plugin since pytest does not offer the possibilty to enable plugins per conftest file (i.e., I cannot test both the new plugin and keep the old one for the other tests).

As such, I need to merge both plugins (legacy and new implementations) in one plugin. I could like.. ignore the test part of the new plugin but it wouldn't make sense since no one would be able to see the plugin in action otherwise (and just pushing the utility functions does not seem right either).

Once #12089 is merged, this one can follow as it is based on top of the latter. After that, we need to fix each test individually so that they use the new features (for now, the plugin is still backwards compatible, though some parts can be removed when we're done with the cleanup).

cc @jayaddison


picnixz:
So, for those interested in reviewing:

- Be aware that ``xdist`` is probably the most capricious framework I've worked with for pytest. Note that I added it as a dependency but we do not use it (for now). I also pinned the dependency because I need to hack into it at some point. 
- Ignore the changes in the *.rst files since they are the same as #12089.

There are two major parts in this PR. The first part is the implementation of the plugin itself. Most of the work is done in the `internal` package where no magic occurs. Markers are processed and arguments are derived according to some strategy that is explained in the comments.

The second part consists in testing the plugin itself. And there's a bit of magic there. As I said before, xdist is capricious. One major drawback is the lack of output capturing. When using pytester, you essentially run an instance of pytest in the process and you can check things directly in the tests you want. Now, imagine you have two tests and you want to be sure that test1 and test2 use different files. The issue is that you cannot test that in test1, nor can you test that in test2 since they need to be executed simultaneously (I want that). So, you would usually print inside the tests, capture the output and compare whatever you want.

The issue is that xdist does not allow you to capture ANYTHING you print using "print". The only way to capture something is to write to the disk (which I don't want to) or use the report sections of pytest + the '-rA' flags. That way, when test1 and test2 are done, the pytest you were running will show the report sections (even if you had used xdist to run test1 and test2 in parallel). And NOW, you can parse that output. This is the job of the "magico" plugin. 

Feel free to ask any questions.

chrisjsewell:
Heya, as per #12089, could you change your initial comment, to be more specific about what issues this PR proposes to fix, and how it proposes to fix them

picnixz:
Also, the SharedResult class was using a weird construction (like, you use instances for fixtures but use a class variable to hold whatever you want for the whole module). It's fine but it's actually cleaner to use the stash object that pytest suggests instead. Also, having the fixture name clash with the marker argument without being of the same type was confusing for me so I renamed it (shared_result -> module_cache) but kept the old one for backwards compat.

picnixz:
Aaaand, finally, I've introduced a better lazy app which does what you think it should do. Like, if you use `freshenv=True` but use a shared result, it should not be accepted. Same as if you use "force_all" (I explained the rationale at the code level).

sphinx/testing/internal/cache.py
```diff
@@ -0,0 +1,61 @@
...
+
+from io import StringIO
+from typing import TYPE_CHECKING, TypedDict
+
+if TYPE_CHECKING:
+    from sphinx.testing.util import SphinxTestApp
+
+
+class _CacheEntry(TypedDict):
+    """Cached entry in a :class:`SharedResult`."""
```
picnixz:
```suggestion
    """Cached entry in a :class:`ModuleCache`."""
```

(for me for later)

sphinx/testing/internal/markers.py
```diff
@@ -0,0 +1,306 @@
...
+    from typing_extensions import Required
+
+    from sphinx.testing.internal.isolation import NormalizableIsolation
+    from sphinx.testing.internal.pytest_util import TestRootFinder
+
+
+class SphinxMarkEnviron(TypedDict, total=False):
+    """Typed dictionary for the arguments of :func:`pytest.mark.sphinx`.
+
+    Note that this class differs from :class:`SphinxInitKwargs` since it
```
picnixz: I'll need to update this comment.

sphinx/testing/internal/markers.py
```diff
@@ -0,0 +1,306 @@
...
+class TestParams(TypedDict):
+    """A view on the arguments of :func:`pytest.mark.test_params`."""
+
+    shared_result: str | None
+
+
+def _get_sphinx_environ(node: PytestNode, default_builder: str) -> SphinxMarkEnviron:
+    args, kwargs = get_mark_parameters(node, 'sphinx')
+
+    if len(args) > 1:
```
picnixz: For the marker, I think we should stick to positional builder name + keyword arguments for the rest.

sphinx/testing/internal/markers.py
```diff
@@ -0,0 +1,306 @@
...
+    check_mark_str_args('test_params', shared_result=shared_result)
+
+    if shared_result is not None:
+        # include the testroot id for visual purposes (unless it is
+        # not specified, which only occurs when there is no rootdir)
+        return f'{testroot}-{shared_result}' if testroot else shared_result
+
+    if testroot is None:
+        # neither an explicit nor the default testroot ID is given
+        pytest.fail('missing %r or %r parameter' % ('testroot', 'srcdir'))
```
picnixz: Note that without explicit keyword arguments, it's a bit harder to detect this. Also, the current implementation of the plugin assumed that 'srcdir' was a keyword argument so...

cc @chrisjsewell

tests/conftest.py
```diff
@@ -23,30 +40,166 @@ def _init_console(locale_dir=sphinx.locale._LOCALE_DIR, catalog='sphinx'):
...
-collect_ignore = ['roots']
+# Exclude resource directories for pytest test collector
+collect_ignore = ['certs', 'roots']
 
 os.environ['SPHINX_AUTODOC_RELOAD_MODULES'] = '1'
 
 
+def pytest_configure(config: Config) -> None:
+    config.addinivalue_line('markers', 'serial(): mark a test as non-xdist friendly')
+    config.addinivalue_line('markers', 'unload(*pattern): unload matching modules')
```
picnixz: For now, they are unused but will be used for autodoc tests.

picnixz:
After looking around on github, I think I need to keep a complete backwards compatibility of the plugin, even with how the fixtures and objects are processed. I cannot assume some things so instead, I found a way to switch between implementations according to whether we use the legacy implementation or not.

From now, on the legacy implementation will be kept frozen and won't be maintained anymore (unless I've introduced a bug). Once we are in 9.x, we will remove this switch and continue onwards with the new implementation.

picnixz:
I've removed most of the code that is needed after and only left a minimal set of functions (although I'm giving the full set of utils, I think it's fine since it will be used later).

sphinx/testing/fixtures.py
```diff
@@ -2,216 +2,470 @@
...
+
+    from sphinx.testing.internal.isolation import IsolationPolicy
+    from sphinx.testing.internal.markers import TestParams
+
+    AnySphinxTestApp = Union[SphinxTestApp, SphinxTestAppWrapperForSkipBuilding]
+    AnyAppParams = Union[AppParams, AppLegacyParams]
+
+DEFAULT_ENABLED_MARKERS: Final[list[str]] = [
     (
-        'sphinx(builder, testroot=None, freshenv=False, confoverrides=None, tags=None, '
```
picnixz: Note that the current implementation (on master) assumes that 'testroot' is a keyword argument (it nevers checks for the 1st positional argument).

sphinx/testing/_internal/markers.py
```diff
@@ -0,0 +1,340 @@
...
+    # 1. process pytest.mark.sphinx
+    env = _get_sphinx_environ(node, default_builder)
+    # 1.1a. deduce the isolation policy from freshenv if possible
+    freshenv: bool | None = env.pop('freshenv', None)
+    if freshenv is not None:
+        if 'isolate' in env:
+            err = '%r and %r are mutually exclusive' % ('freshenv', 'isolate')
+            pytest.fail(format_mark_failure('sphinx', err))
+
+        isolation = env['isolate'] = Isolation.always if freshenv else default_isolation
```
jayaddison, line 270: Should `isolate=True` also imply `freshenv=True`?  (and if not, why not?)

sphinx/testing/_internal/util.py
```diff
@@ -0,0 +1,106 @@
...
+    r"""Generate a random unique identifier prefixed by *prefix*.
+
+    :param prefix: An optional prefix to prepend to the unique identifier.
+    :return: A random unique identifier.
+    """
+    suffix = os.urandom(UID_BUFLEN).hex()
+    return '-'.join((os.fsdecode(prefix), suffix)) if prefix else suffix
+
+
+def get_environ_checksum(*args: Any) -> int:
```
jayaddison: Could/should this accept a single (type-hinted) `env` object as a parameter instead?

picnixz: Mmh. Actually the function is very generic, but I can make two versions, where one is typed and expects some explicit parameters and another just takes the kwargs. I don't know if we'll need a more generic thing in the future, that's why I made it like that.

sphinx/testing/_internal/markers.py
picnixz: Yes and no. 

Since 'freshenv=True' means "you wipe out the environment", it's as if you were fully isolated (that's what you want when you do some `-E`, you want to be clean from whatever cached thing), thus having `isolate=True` is kind of equivalent to having `freshenv=True`. 

The reason why I did not consider `freshenv=True` equivalent to `isolate=True` is that once you've isolated your sources, you assume you are... isolated. You wouldn't use `-E` the very first time you build your doc right? so that's the same idea.

tests/roots/test-minimal/conf.py
```diff
@@ -0,0 +1,3 @@
+# minimal test root
+include_patterns = ['index.rst']
+exclude_patterns = ['_build']
```
jayaddison, line 3: Are these two configuration options required?

tests/test_extensions/test_ext_autodoc.py
```diff
@@ -2163,7 +2164,9 @@ def test_singledispatchmethod_classmethod_automethod(app):
 @pytest.mark.skipif(sys.version_info[:2] >= (3, 13),
                     reason='Cython does not support Python 3.13 yet.')
 @pytest.mark.skipif(pyximport is None, reason='cython is not installed')
-@pytest.mark.sphinx('html', testroot='ext-autodoc')
+# use an explicit 'srcdir' to make the path smaller on Windows platforms
+# so that cython can correctly compile the files
+@pytest.mark.sphinx('html', srcdir=uuid.uuid4().hex, testroot='ext-autodoc')
```
jayaddison, line 2696: What's the failure that occurs on Windows here?  I can't find any entries in GitHub issues for it.

If the `uuid` usage here is a workaround then perhaps it'd be good to keep a note that we can remove that if+when the underlying problem is solved.

sphinx/testing/util.py
```diff
@@ -204,14 +207,65 @@ def build(self, force_all: bool = False, filenames: list[str] | None = None) ->
         super().build(force_all, filenames)
 
 
+class SphinxTestAppLazyBuild(SphinxTestApp):
```
jayaddison, line 225: This seems logically similar to `SphinxTestAppWrapperForSkipBuilding` - is there a reason we can't use that instead?

tests/test_extensions/test_ext_autodoc.py
picnixz: The issue is about the path length. Because of the way the tests are isolated, the path of the file on Windows ends up being very long. And when Cython compiles the files, it appears to compile a file that has a path too long. So by using an explicit srcdir, the isolated path is smaller and cython won't use a too large path either.

jayaddison:
Ok, I'm gradually beginning to understand this; I'll probably have a lot of questions.  One of the next ones is: why do we need `TestRootFinder`?  It seems to add a decent amount of code and test coverage, and I'm unsure how it relates to the other changes.  There's a chance I'll figure that out as I read/examine the changes more, but I thought I'd ask here too.

sphinx/testing/util.py
picnixz: There are real differences. First is that the other class is a proxy class so you don't have a direct inheritance. The second is that it does not check the usage of the class and I don't want to make tests using the old plugin fail because of that.

tests/roots/test-minimal/conf.py
picnixz: I'm not sure that the exclude_patterns include the '_build' by default. I think the default exclude patterns are `[]`. It's just to speed-things up when finding iterating the files.

tests/test_extensions/test_ext_autodoc.py
jayaddison: Do we need to use a unique `srcdir` name for that, or could we pick a short fixed name?

sphinx/testing/util.py
jayaddison: Should it be added to `__all__`?  (I don't use that feature often but I do notice that `SphinxTestAppWrapperForSkipBuilding` is in there)

picnixz:
Actually, if I remove it, I need to always process the three fixtures (default testroot, rootdir and prefix (though this one could be assumed to always be 'test-')) separately. In particular, you need to decide how empty strings and None objects are handled if by any change you use that. I'll think a bit about this today again but it's possible to remove it (though I personnally think that having something centralized for this logic is fine, especially for checking that the user did not choose a bad fixture value).

tests/test_extensions/test_ext_autodoc.py
picnixz: We can pick a short fixed name but then you need to be sure that it won't clash with another test in the future. Honestly, using a UUID in general is fine and it's not much work (I could even just write os.urandom(16).hex() instead).

tests/roots/test-minimal/conf.py
jayaddison: Ok, thanks.  This is really a question in case anyone later thinks "ah, this is the absolute minimal Sphinx project" (based on the dirname).  A `conf.py` file does seem to be required for Sphinx to build a project, but it can be empty.

If we're not outputting files into this test-root during the tests (and I don't think we should be?), then we could remove the `exclude` setting at least.  Probably the `include_patterns` too.

sphinx/testing/util.py
picnixz: I personally think that this class is an implementation detail and not part of the public scope but I don't want to put an `_` beforehand since it could mean private for the module itself (but it's only private for outside of Sphinx itself). So I will not export the new ones (though I will leave those that are already exported in the public scope).

It's more of a protected class (public for the `sphinx.testing`, private outside, except for our tests).

jayaddison: Sounds good.  Thank you :)

tests/test_extensions/test_ext_autodoc.py
jayaddison: > I could even just write os.urandom(16).hex() instead

Ok, and if I understand, then that's [what `isolate=True` does](https://github.com/picnixz/sphinx/blob/f7d7b738602d2ba8675880172587a3ff6f7bf78f/sphinx/testing/_internal/markers.py#L277).  Could we reuse that here instead of a different random ID scheme?

picnixz: Actually, with `isolate=True`, you have long paths. If you explicitly set *srcdir*, you don't have long paths, namely:

- isolate=True => the srcdir being generated will be located at `TMPDIR/SOME-RANDOM-ID-BASED-ON-TEST-LOCATION/SOME-UNIQUE-ID-BASED-ON-CONFIGURATION/TESTROOT-NAME`
- explicit srcdir => location will be `TMPDIR/-/0/EXPLICIT-SRCDIR-POSSIBLY-WITH-SOME-UNIQUE-ID-AT-THE-END`.

Since I want to keep the path as small as possible for this specific test (other tests seem fine), I don't event want the unique suffix that would be added if I were to combine srcdir and isolate.

sphinx/testing/_internal/util.py
```diff
@@ -0,0 +1,115 @@
...
+    container = '@'.join(filter(None, names)) or node.nodeid
+    return unique_object_id(container)
+
+
+def get_location_id(location: TestNodeLocation) -> str:
+    """Get a unique hexadecimal identifier out of a test location.
+
+    The ID is based on the physical node location (file and line number).
+    """
+    fspath, lineno = location
```
jayaddison, line 116: Perhaps it doesn't matter in this case, but I noticed that sometimes we add `1` to the line number, and sometimes not.

https://github.com/picnixz/sphinx/blob/f7d7b738602d2ba8675880172587a3ff6f7bf78f/sphinx/testing/_internal/pytest_util.py#L363

Is there a potential for off-by-one in some cases?  For an ID generation scheme, I guess it probably usually doesn't matter so long as the entire set of output values retains distinct values regardless of offset.

picnixz: It doesn't really matter. The +1 is when you need a visual confirmation (since the location is 0-based) and is really when you know that the line number is correct (i.e., >= 0). Here, this function can *technically* be called with a negative number, like '-1' in some rare cases which is why I did not include the offset.

The problem is that pytest does not tell me when I'm guaranteed to have a correct test location... so I can't really know about this. Now, I should perhaps change the type and not use a test node location but just a `tuple[str, int]` since the line number could be '-1' if I'm using directly `item.location[:2]`.

Ok, long story short: 

- you can call this function with `(fspath, lineno)` when the location was successfully retrieved using `get_node_location` (because if lineno is bad, that function returns None).
- you can also call this function with `(fspath, -1)` when you are dealing with a `pytest.Item` for which `item.location[:2]` *may* return `(fspath, None)`. So, what I do when I'm in this case is replace the None by -1.

Note: in #12106, I did not correctly replace None by -1 so I'll fix this in a moment.
Note 2: done

jayaddison:
Big-idea question: is the testing plugin also intended for use by other projects (for example, themes) to test their builds, or is it mainly/entirely for use within Sphinx?

picnixz:
> Big-idea question: is the testing plugin also intended for use by other projects (for example, themes) to test their builds, or is it mainly/entirely for use within Sphinx?

It's only meant to be used as a plugin for testing your Sphinx extension (as a third-party). It's not meant to be used as a "regular" Python module, though people can still do whatever they want but for this I will say they shouldn't be surprised in case some things don't work they expect.

jayaddison:
> > Big-idea question: is the testing plugin also intended for use by other projects (for example, themes) to test their builds, or is it mainly/entirely for use within Sphinx?
> 
> It's only meant to be used as a plugin for testing your Sphinx extension (as a third-party), in Python, and when you use `pytest`. It's not meant to be used as a "regular" Python module, though people can still do whatever they want but for this I will say they shouldn't be surprised in case some things don't work they expect.

Ok, thank you.  In that case, I have a question/suggestion: would renaming `TestRootFinder` to `TestProjectFinder` make sense?  We place our test projects in a `test-roots` dir, but that may not be the case elsewhere.

picnixz:
Err... why not. It's just that `testroot` is a common name when taking about tests and their static content. It wasn't for the testroot `test-root`.

jayaddison:
> Err... why not. It's just that `testroot` is a common name when taking about tests and their static content. It wasn't for the testroot `test-root`.

OK, I didn't realize the `testroot` name had become a term of reference.  To me, the phrase 'test project' is self-descriptive and I think accurate, but if `testroot` is well-known then that's good too.

chrisjsewell (changes requested):
Heya, before I look properly into the semantics, please can you add in the user documentation, an explanation for extension developers on how they can use the testing tools in their package

jayaddison:
FYI @picnixz: I'm planning to take a more detailed code-review/local-exploration of these changes later today.

picnixz:
> Heya, before I look properly into the semantics, please can you add in the user documentation, an explanation for extension developers on how they can use the testing tools in their package

Mmh, I was waiting for separate PR for that actually. Technically speaking, the `_internals` module is only for us (so not even extension developers, only for Sphinx) and the rest did not change much (from an extension developer, there is no changes to what is currently served since by default, the plugin uses the legacy implementation).

sphinx/testing/_internal/pytest_util.py
```diff
@@ -0,0 +1,415 @@
...
+class TestRootFinder:
+    """Object responsible for finding the testroot files in *rootdir*.
+
+    For instance::
+
+        finder = TestRootFinder('/foo/bar', 'test-', 'default')
+
+    describes a testroot root directory at ``/foo/bar/roots``. The name of the
+    directories in ``/foo/bar/roots`` consist of a *prefix* and an *ID* (in
+    this case, the prefix is ``test-`` and the default *ID* is ``default``).
```
jayaddison:
```suggestion
        finder = TestRootFinder('/foo/bar', 'test-', 'default')

    describes a testroot root directory at ``/foo/bar/``. The name of the
    directories in ``/foo/bar/`` consist of a *prefix* and an *ID* (in this
    case, the prefix is ``test-`` and the default *ID* is ``default``).
```

sphinx/testing/fixtures.py
```diff
@@ -2,224 +2,476 @@
...
+) -> AppParams | AppLegacyParams:
+    """Parameters that are specified by ``pytest.mark.sphinx``.
+
+    See :class:`sphinx.testing.util.SphinxTestApp` for the allowed parameters.
     """
-    Test parameters that are specified by 'pytest.mark.test_params'
+    if sphinx_use_legacy_plugin:
+        msg = ('legacy implementation of sphinx.testing.fixtures is '
+               'deprecated; consider redefining sphinx_legacy_plugin() '
+               'in conftest.py to return False.')
```
jayaddison:
```suggestion
        msg = ('legacy implementation of sphinx.testing.fixtures is '
               'deprecated; consider redefining sphinx_use_legacy_plugin() '
               'in conftest.py to return False.')
```

sphinx/testing/fixtures.py
```diff
@@ -2,224 +2,476 @@
...
+        return __app_params_fixture_legacy(
+            request, test_params, shared_result,
+            sphinx_test_tempdir, testroot_finder.path,
+        )
 
-    :param Union[str] shared_result:
-       If the value is provided, app._status and app._warning objects will be
-       shared in the parametrized test functions and/or test functions that
-       have same 'shared_result' value.
-       **NOTE**: You can not specify both shared_result and srcdir.
```
jayaddison, line 129: Perhaps worth keeping this on `__app_params_fixture_legacy`?

sphinx/testing/_internal/markers.py
```diff
@@ -0,0 +1,364 @@
...
+) -> str:
+    """Deduce the sources directory from the given arguments.
+
+    :param srcdir: An optional explicit source directory name.
+    :param testroot: An optional testroot ID to use.
+    :param shared_result: An optional shared result ID.
+    :return: The sources directory name *srcdir* (non-empty string).
+    """
+    check_mark_str_args('sphinx', srcdir=srcdir, testroot=testroot)
+    check_mark_str_args('test_params', shared_result=shared_result)
```
jayaddison, line 178: I'm not sure that it's worth extracting a utility method for these checks; I think it may be simpler to perform the checks inline.

(this method is used twice here and in part-3, and similarly `check_mark_keywords` is used three times)

sphinx/testing/_internal/markers.py
```diff
@@ -0,0 +1,364 @@
...
+        namespace, checksum = '-', 0
+    else:
+        namespace = get_container_id(node)
+        # Do a somewhat hash on configuration values to give a minimal protection
+        # against side-effects (two tests with the same configuration should have
+        # the same output; if they mess up with their sources directory, then they
+        # should be isolated accordingly). If there is a bug in the test suite, we
+        # can reduce the number of tests that can have dependencies by adding some
+        # isolation safeguards.
+        checksum = _get_common_config_checksum(env['buildername'], **env)
```
jayaddison, line 308: Since checksumming selected elements from the Sphinx `env` should be relatively cheap, I think we could do this for unique `srcdir` values too.

However: perhaps it would be better to merge this feature separately before the plugin-related parts of the changeset?

jayaddison (commented):
I could be mistaken, but I don't think that this is a simplest / most minimal approach towards introducing `xdist` / parallelsm.

I don't yet understand all the context, but to me it seems that running our own tests in a parallelizable way is worth decoupling from providing that same functionality to third-party test suites -- unless we have high confidence that the change is small and safe.  In this case based on review so far I do not have a sufficient level of confidence.

jayaddison:
In the back of my mind I am thinking: the cost and time savings from this change could be significant - so is it a risk to delay/block it?  It could be.

However, when I think about what a pull request to enable parallelism _should_ contain, I would ideally expect that it should be no more than 50 lines of code - an additional dependency, some config changes, some CI workflow changes.

Instead it seems we're adding more code _within_ the codebase to work around the fact that the application is not parallelizable in some fundamental ways.  One of those is the in-process mutation of objects/imports that I'm aware `autodoc` and others do -- but even so my sense is that solving the problem by working backwards towards the origins would be long-term much more effective, because otherwise we're adding yet more layers that become difficult to unpick (that being the history of the test suite and plugin becoming more complex instead of less complex).

The annoying part is that, again, I don't have better suggestions, and so I think my vote review should be considered a +0 (no impact positive/negative on whether we merge this).

sphinx/testing/_internal/markers.py
picnixz: The reason is mostly because to have always the same message and to avoid possible mistakes.

picnixz:
> I don't yet understand all the context, but to me it seems that running our own tests in a parallelizable way is worth decoupling from providing that same functionality to third-party test suites

Then this means not using our own plugin at all and having a plugin at the `tests/` level. I can do it (and then sphinx/testing will not be changed at all) without too much work though.

sphinx/testing/fixtures.py
picnixz: Yes

picnixz:
> the cost and time savings from this change could be significant - so is it a risk to delay/block it? It could be.

Not really. It's not that hard for me to move pieces around or refactor things. But in light of your previous comment, I will move everything to our local tests so that I reduce the risk downstream as much as possible. It's also more effective that way and I'm confident that I could fix any future issue (like, if there's is an issue, we can just switch back to the old plugin implementation and voilà).

> However, when I think about what a pull request to enable parallelism should contain, I would ideally expect that it should be no more than 50 lines of code - an additional dependency, some config changes, some CI workflow changes.

Yes, that's what I would have expected but xdist is not friendly with the output so there are some things that *must* be done once you've enabled this plugin (like, every part that has a print will *NOT* be rendered). Now, if you enable `xdist`, you need to fix the side-effects in a single PR (otherwise the CI/CD will not work at all!), whereas with the approach where I have a default isolation strategy, you can fix side-effects in multiple PRs.

> Instead it seems we're adding more code within the codebase to work around the fact that the application is not parallelizable in some fundamental ways

The application is parallelizable but it's not friendly with pytest + xdist at some point.

>  One of those is the in-process mutation of objects/imports that I'm aware autodoc and others do -- but even so my sense is that solving the problem by working backwards towards the origins would be long-term much more effective, because otherwise we're adding yet more layers that become difficult to unpick (that being the history of the test suite and plugin becoming more complex instead of less complex).

Half of the side-effects bugs come from autodoc, the other half comes from the modifications on sources that are then kept inbetween tests (and thus, making them fail if they are executed in an arbitrary order). Technically speaking, my PR (and the following) will not directly solve that since it's a Python-related and not a filesystem related issue. 

There are hundreds of autodoc tests and it's difficult to locate which one has a dependency on the other since Sphinx changes somtimes the module's annotations and some modules are used more than once by the autodoc tests (sometimes only a part of those modules are used). I have a planned fixture to fix those tests though but it's not in this PR yet and you cannot really see it in effect unless you enable the parallelization (you could enable the plugin for this subset of tests but because of the pytest integration, parametrized tests will need special casing in order to patch the workers instantiations).

---

There is ONE possibility to have an isolation strategy that could be achieved by pytest so I'll try this one. This is something Adam tried in https://github.com/sphinx-doc/sphinx/issues/11285#issuecomment-1499788021 but comes at the cost of 4x slowdown. The slowdown is probably (I think) due to the parametrized tests and is not compatible with the `shared_result` approach. So I *could* try something that makes it compatible but I'm not sure how good it will be.

jayaddison:
> > the cost and time savings from this change could be significant - so is it a risk to delay/block it? It could be.
> 
> Not really. It's not that hard for me to move pieces around or refactor things. But in light of your previous comment, I will move everything to our local tests so that I reduce the risk downstream as much as possible. It's also more effective that way and I'm confident that I could fix any future issue (like, if there's is an issue, we can just switch back to the old plugin implementation and voilà).

I do believe that you'd be able to maintain/support/improve this, yep.  However: I'd prefer (selfishly!) for your time for that to be available for core Sphinx issues.

Also selfishly, I don't know how much of my own time I'd be able (or want) to offer to fixing test plugin issues.  Creating a certan amount of 'testing gravity' could help us make sure plugins/themes/etc are remaining compatible - and if so perhaps I'd end up feeling satisfied that we're improving ecosystem consistency and _would_ in fact start to help -- but we only have a few active maintainers/committers so I think there's a risk of DDoS'ing ourselves.

(hard to quantify the risks, especially without data, but these are the things I'm thinking about)

picnixz:
Here are some plans that I can suggest (honestly, it was fun on my side to play with pytest so I don't mind the time loss).


### Plan 1

- Continue with what I did, but only do it for Sphinx itself. We might change in the future the original plugin implementation to make it more robust but we will not make it "xdist" friendly (and that will be left to downstream extensions).
- I move the new implementation of the fixtures in our local `tests/`. Maybe there are things I don't need so I'll cleanup some utils (like, the `find_context` function is fine and helpful, but perhaps it's an overkill).
- We fix the tests by using the isolate approach (which is cleaner than using your own unique ID).

### Plan 2

- Abandon the new plugin implementation and fix our tests bit by bit. We can modify a bit our fixtures so that *srcdir* is betterly chosen internally and make it xdist compatible for our needs but no refactorization on the markers themselves.
- It will be harder to fix the tests one by one because it's hard to see which tests have dependencies or not. While this is true for Plan 1, in this plan, we don't have this automatic isolation according to the test module + configuration and then there are probably more tests that need to be isolated manually.
- Some helper functions need to be refactored (as mentioned in https://github.com/sphinx-doc/sphinx/pull/12118#issuecomment-2003327289 and https://github.com/sphinx-doc/sphinx/pull/12118#issuecomment-2003376532). But we need to figure out which ones need to be refactored...

sphinx/testing/_internal/markers.py
picnixz: > I think we could do this for unique srcdir values too.

No because otherwise you can have very long paths. It doesn't really make sense to make something unique even more unique IMO (there is still a call to SHA-1 behind).

sphinx/testing/_internal/pytest_util.py
picnixz: Good catch (it was something I did not touch since weeks ago).

sphinx/testing/_internal/cache.py
```diff
@@ -0,0 +1,155 @@
...
+            (f'{name:{kw}s} {strvalue}' for name, strvalue in config),
+            [f'{" messages ":-^{tw}}', text] if (text := self._messages) else (),
+            [f'{" warnings ":-^{tw}}', text] if (text := self._warnings) else (),
+            ['=' * tw],
+        )
+        return '\n'.join(lines)
+
+
+# XXX: RemovedInSphinx90Warning
+class LegacyModuleCache:  # kept for legacy purposes
```
jayaddison, line 133: Is the caching functionality here (either the legacy caching, or the more modern alternative) used in the codebase yet?  I couldn't find references to it.

picnixz: It should be used somewhere unless I messed up the conflicts resolution.... I'll check it upon my return.

AA-Turner:
@picnixz is this new testing plug-in needed, given we've managed to solve most independence and parallisation issues? If so, I would only feel comfortable merging in smaller increments -- I don't have sufficient time to review one +2,300/-200 line PR.

A

picnixz:
There are too many conflicts that I need to resolve, so I think I'll need to rewrite it. I'm fine with closing this one and the related PRs. If we can now safely enable the xdist plugin then my plugin is no more needed but I don't know if it's possible.

AA-Turner:
Ok. Closing this PR, then -- thank you for what is clearly quite a lot of work that went into it!

A

Issue #12094, opened 2024-03-14
tests: network state (online/offline) affects epubcheck unit test outcome.


jayaddison:
### Describe the bug

The [`test_run_epubcheck` unit test case](https://github.com/sphinx-doc/sphinx/blob/98d0cc8602e254218cdd01212366d5304e5fba48/tests/test_builders/test_build_epub.py#L375-L390) passes as expected when:

* Java is [considered runnable](https://github.com/sphinx-doc/sphinx/blob/98d0cc8602e254218cdd01212366d5304e5fba48/tests/test_builders/test_build_epub.py#L382).
* The `epubcheck[.jar]` program [is found](https://github.com/sphinx-doc/sphinx/blob/98d0cc8602e254218cdd01212366d5304e5fba48/tests/test_builders/test_build_epub.py#L381-L382).
* The `DO_EPUBCHECK` environment variable [is set](https://github.com/sphinx-doc/sphinx/blob/98d0cc8602e254218cdd01212366d5304e5fba48/tests/test_builders/test_build_epub.py#L375-L376) (the value does not matter).
* It can download a [remote image contained in the `test-root/images.txt`](https://github.com/sphinx-doc/sphinx/blob/98d0cc8602e254218cdd01212366d5304e5fba48/tests/roots/test-root/images.txt#L15-L16) file.

(note: we _do_ enable `DO_EPUBCHECK` in our continous integration [in the `latex` workflow](https://github.com/sphinx-doc/sphinx/blob/98d0cc8602e254218cdd01212366d5304e5fba48/.github/workflows/main.yml#L121-L143) -- and `epubcheck` is installed into the relevant `sphinx-ci` container image)

There are a few quirks here: I think that the test behaving differently offline isn't desirable though (or to phrase it another way: connectivity shouldn't be a requirement for most dependable software to test/build, because it's hard to predict when/where it will be rebuilt).

### How to Reproduce

Ensure that `epubcheck` and the local development and test dependencies for `sphinx` are installed, and then disable network connectivity on the unit test host.

From the `sphinx.git` checkout directory:
```
$ DO_EPUBCHECK=1 pytest tests/test_builders/test_build_epub.py -k epubcheck
```

### Environment Information

```text
Platform:              linux; (Linux-6.6.15-rt-amd64-x86_64-with-glibc2.37)
Python version:        3.11.8 (main, Feb  7 2024, 21:52:08) [GCC 13.2.0])
Python implementation: CPython
[...]

PR #12098, opened 2024-03-15
[tests] epub: skip the epubcheck test if Java or epubcheck itself are not found

jayaddison:
### Feature or Bugfix
- Bugfix

### Purpose
- The `test_run_epubcheck` test has to be enabled explicitly by the environment setting the `DO_EPUBCHECK` variable.  For the test to pass without indicating any problems when the test did not exercise any behaviour or check any results seems wrong - it could result in test reports that produce a false sense of confidence.

### Detail
- Instead of reporting `success` when the `test_run_epubcheck` does nothing due to the apparent absense of either Java or `epubcheck.jar` on the host, use `pytest` to indicate a `skip` result.

### Relates
- Relates to / discovered during investigation of #12094 (but does not fix it).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -102,6 +102,9 @@ Bugs fixed
 Testing
 -------
 
+* pytest: report the result of ``test_run_epubcheck`` as ``skipped`` instead of
+  ``success`` when Java and/or the ``epubcheck.jar`` code are not available.
+
 Release 7.2.6 (released Sep 13, 2023)
 =====================================
 
--- a/tests/test_builders/test_build_epub.py
+++ b/tests/test_builders/test_build_epub.py
@@ -378,16 +378,21 @@ def test_duplicated_toctree_entry(app, status, warning):
 def test_run_epubcheck(app):
     app.build()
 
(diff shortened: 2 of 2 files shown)


jayaddison:
Output when `java` is found not runnable on the host:

```
tests/test_builders/test_build_epub.py ..........s..                                                   [100%]

========================================== short test summary info ===========================================
SKIPPED [1] tests/test_builders/test_build_epub.py:382: Unable to run Java; skipping test
======================================= 12 passed, 1 skipped in 3.73s ========================================
```

Output when `epubcheck.jar` is found unavailable on the host:

```
tests/test_builders/test_build_epub.py ..........s..                                                   [100%]

========================================== short test summary info ===========================================
SKIPPED [1] tests/test_builders/test_build_epub.py:386: Could not find epubcheck; skipping test
======================================= 12 passed, 1 skipped in 3.85s ========================================
```

picnixz:
Ah ! thank you ! I had this change in my huge re-implementation but forgot about it. So thanks for that.

picnixz:
Thank you!

jayaddison:
Thanks!

